Issue #2181, opened 2026-08-19
[PRODUCT-DESIGN] PostHog session recordings — findings


martasl00:
# OpenLinker — PostHog sessions

### 01.01 Operations / Orders - **rozpisane na taska**

Recording: [https://eu.posthog.com/project/231601/replay/01a00ecb-5f26-7605-a247-6426737f726d](https://eu.posthog.com/project/231601/replay/01a00ecb-5f26-7605-a247-6426737f726d)

<img width="1564" height="791" alt="Image" src="https://github.com/user-attachments/assets/decbe062-bf2c-4c73-ab54-15bfcdb5914d" />

#### Observation

Clicking the "Select all" checkbox selected 11 out of the 13 orders visible on the page. Two orders with the "Dispatched" status were skipped, because they no longer qualify for the action. The interface does not explain why some checkboxes stayed empty, so the user may assume that "Select all" did not work. In one of the sessions you can see them trying to click the records excluded from the selection.

The scope of the selection is not clear either. The list has 29 results in total, and 11 orders from the current page were selected. The "11 selected" message and the action button sit in a bar pinned to the bottom edge of the screen. During the test the user may not have noticed that bar.

#### Recommendations

**01 Clearly communicate the scope of the selection and the skipped records**

A message stating that the selected records come from this page:

> 11 of 13 orders on this page selected.

**02 Consider a different treatment for records that do not qualify for the bulk action** (the ones with the "Dispatched" status).

**03 Consider moving the bulk actions and the selection info above the table**

The typical pattern is a contextual action bar above the table, appearing once at least one record is selected. It may work better in this context.

> 11 of 13 orders on this page selected [Dispatch] [Clear selection]

#### Watch out

Scalability. Check how this bar behaves with a larger number of bulk actions.

### 01.02 Operations / Invoices - rozpisane na taska

The **Invoices** page has no "Select all" option, even though records can be selected manually.

The bulk actions appear below the table and only become visible after scrolling the page. The user may not notice them.

**Recommendation:** align the pattern with **Orders** and add "Select all".

<img width="1670" height="880" alt="Image" src="https://github.com/user-attachments/assets/d0f313ce-a003-49df-a969-df5566bcb75b" />

## 02 Publish product - jeszcze nei rozpisane, musze sie zastanowic jak i czy warto

Products added to the bulk publication list are hard to see, because the box is small.

**Recommendation:** increase the height of the list, or present the data in this form differently.

[...]

PR #2568, opened 2026-08-26
docs(sales-documents): ADR-065 one per-order read, no third state vocabulary

norbert-kulus-blockydevs:
Part of the documentation epic #2499, mini-epic #2500 (phase 1 — decisions that gate the implementation). Phase 1 merges **straight into `main`**, because the implementation epic's branch (#2513) cannot be cut until these decisions exist.

## What this adds

- **ADR-065** — one neutral per-order sales-document projection, composing the existing vocabularies rather than defining a third.
- **The UX mockup**, committed at `docs/plans/mockups/sales-document-routing.html`. States are hash-addressable (`#page=settings`, `#page=orders`, `#page=order&state=receipt-in-doubt`), so every later doc and E2E spec can cite one directly.
- ADR index row, and `066` reserved for #2503.

## The two rules that matter

1. **A surface never re-derives a fact the backend persists.** A review draft printed `no rule for PL` from the order's country while the persisted reason was something else — a false statement about the operator's own configuration.
2. **A fiscal receipt has no authority axis.** Another draft collapsed the two invoice axes into one enum, which makes *issued, then rejected by the authority* — the state `RegulatoryResubmitter` exists for — unrepresentable.

## Review

`/pr-review` per the epic workflow: verdict as a comment, every finding fixed with a reply saying what changed, re-review until approved.

Closes #2501

PR #2643, opened 2026-08-27
docs(mockups): one bulk-action pattern for Orders and Invoices (#2636)

### LINKED ISSUE #2636: [Product design] Unify the bulk-action pattern across Orders and Invoices
**Mockup:** https://claude.ai/code/artifact/7cd55161-be3e-4e8e-b46b-1a75a7c8a63b
**Evidence:** #2181 § 01 — the PostHog session findings this issue comes from.

## Problem

`/orders` and `/invoices` both use checkboxes and bulk actions, but they behave differently.

The user sees only the number of selected records, for example `11 selected`, without knowing how many records are visible on the page or why some of them were not selected.

In addition, both lists communicate differently which records are eligible for an action and what the outcome of that action was.

## Goal

Design one consistent bulk selection and bulk action pattern for `/orders` and `/invoices`.

## Scope

Bulk actions should:

* show the number of selected records in the context of the current page, for example `11 of 13 orders on this page selected`,
* support `Select all` for records visible on the current page,
* clearly communicate which records cannot take part in a given action and why,
* show the number of records the action will actually be performed on,
* show the action result per record, including the reason for an error or skip,
* handle bulk action limits consistently for Orders and Invoices,
* reuse existing eligibility rules without defining new ones.

## States

* nothing selected,
* some records on the page selected,
* all available records on the page selected,
* record not eligible for the action,
* bulk action limit reached,
* action in progress,
* all records processed successfully,
* partial success,
* all records skipped or failed,
* no records eligible for the action,
* read-only,
* page change / selection cleared.

## Out of scope

* backend implementation,
* changing the existing `25` and `100` limits,
* making bulk actions asynchronous,
* changing eligibility rules,
* preserving selection across pages,
* extending the pattern to Products and Listings,
* fixing the demo-mode write access issue.

martasl00:
**Mockup:** https://claude.ai/code/artifact/7cd55161-be3e-4e8e-b46b-1a75a7c8a63b
**Evidence:** #2181 § 01 — the PostHog session findings this pattern answers.

One interactive HTML mockup proposing a single bulk-action pattern for `/orders`
and `/invoices`, in `docs/plans/mockups/`. No app code changes.

## What it shows

Two live views stacked — each list with its real columns — plus twelve pinned
states, one per state the issue enumerates. Light and dark theme. Tokens and
primitives are transcribed from `apps/web/src/index.css` with the source line
beside each rule; the mockup introduces **four** new CSS rules in total, and
they sit alone in one section so the styling cost is a screenful.

The pattern is derived from `BulkDispatchDialog`, not invented. That dialog
already carries the ready/attention split, per-record reasons and a result
phase; the proposal stretches it over Invoices.

## Against the session findings (#2181 § 01)

Every decision here traces to a recorded session, not to taste:

| #2181 recommendation | What the mockup does |
|---|---|
| 01 — state the scope of the selection and the skipped records: *"11 of 13 orders on this page selected"* | The count is measured against the page and the hint carries the gap: `16 of 20 orders on this page selected · 4 don't qualify · Max 25 per source` |
| 02 — consider a different treatment for records that do not qualify (the `Dispatched` ones) | Those rows are simply unselectable, and the fixture puts two `Dispatched` orders on the page exactly as the recording shows. The reason stays readable from a column that already exists — which is also the constraint that makes dropping an eligibility column safe |
| 03 — consider moving the bulk actions and the selection info above the table | The bar sits above the table and stops floating. Costed below: `DataTable` has no slot there today |
| Watch out — scalability with a larger number of bulk actions | The invoices view is the test case: two actions over one selection, each with its own eligibility, so one selection yields two different execution counts (`Retry 5` / `Issue 11`) |

§ 01.02 of the same issue — Invoices has no "Select all", and its bulk actions
sit below the table where they are only visible after scrolling — is the second
half of the divergence this PR closes.

## Three findings worth reading before the workshop

1. **Per-record reasons for Invoices are not a new feature.** The backend
   already returns them (`RetryInvoiceResult.reason`,
   `BulkIssueInvoiceResult.reason` in `invoicing.controller.ts`) and the
   frontend discards them for a count.
2. **"Bulk action limit reached" is unreachable today.** `PAGE_SIZE` is 20 on
   both lists (`orders-list-page.tsx:88`, `invoices-list-page.tsx:74`) and the
   caps are 25 per source and 100 per request, so page-scoped selection — which
   this issue mandates — can never breach either. The bar therefore states the
   cap as capacity, never as a blocker.
3. **Moving the bar above the table is a new slot in a primitive.** `DataTable`
   exposes only `footer` (`shared/ui/data-table.tsx:123`) and all three
   consumers (products, orders, invoices) hang the bar off it.

## Two invariants the pattern depends on

- **Row-level and action-level eligibility are separate.** A list answers "is
  there any bulk action this row could join"; which action refuses it belongs to
  that action's review. Conflating them made an `issued` invoice read "not
  retry-eligible" to an operator reaching for Issue. One selection over the
  invoices list legitimately yields two different execution counts.
- **Removing the eligibility column is only safe while every blocking state is
  readable from a column that already exists** — and that runs both ways: a
  status that blocks the action must never appear on a selectable row, or the UI
  contradicts the product.

## What this does not do

Per the issue's own scope: no backend, the 25 and 100 caps are unchanged,
nothing becomes asynchronous, eligibility rules are reused rather than
redefined, selection does not survive a page change, Products and Listings are
untouched, and the demo-mode write-access gap is shown rather than fixed.

Closes #2636
Refs #2181

---

https://claude.ai/code/session_01QPsiRy6SM498L9ErFqBTtB

### DIFF
--- a/docs/plans/mockups/bulk-actions-pattern-2636.html
+++ b/docs/plans/mockups/bulk-actions-pattern-2636.html
@@ -1,0 +1,1851 @@
+<!doctype html>
+<html lang="en">
+<head>
+<meta charset="utf-8" />
+<meta name="viewport" content="width=device-width, initial-scale=1" />
+<title>#2636 — Bulk actions pattern · UI Mockups</title>
+<!--
+  ═══════════════════════════════════════════════════════════════════════════════
+  #2636 — ONE BULK-ACTION PATTERN FOR /orders AND /invoices
+  ═══════════════════════════════════════════════════════════════════════════════
+
+  What this file is for
+  ─────────────────────
+  Both lists already ship checkboxes and bulk actions. They share the
+  `BulkActionBar` and `CheckboxCell` primitives and diverge in almost
+  everything else. This file proposes ONE pattern and pins every state the
+  issue enumerates, so the divergence can be closed as a single decision
+  rather than twelve small ones.
+
+  What the code review changed about the issue's framing
+  ─────────────────────────────────────────────────────
+  1. The pattern is NOT invented here. `BulkDispatchDialog`
+     (features/orders/components/bulk-dispatch-dialog.tsx) is already ~80% of
+     it: "Dispatch 8 of 11 orders", a ready/attention split, ineligible rows
+     carrying their own reason and an escape hatch, and a per-record result
+     phase. This file derives the pattern from Orders and stretches it over
+     Invoices — it does not design a third thing.
+
+  2. Per-record reasons for Invoices are NOT a new feature. The backend
+     already returns them (`RetryInvoiceResult.reason`,
+     `BulkIssueInvoiceResult.reason` in
+     apps/api/src/invoicing/http/invoicing.controller.ts). The frontend
+     receives them and throws them away, reporting a count instead.
+
+  3. "Bulk action limit reached" is structurally UNREACHABLE today. PAGE_SIZE
+     is 20 on both lists (orders-list-page.tsx:88, invoices-list-page.tsx:74)
+     and the caps are 25 per source connection and 100 per request. Once
+     selection is page-scoped — which this issue mandates — 20 rows can never
+     breach either cap. Frame 05 states this rather than staging a state the
+     product cannot enter.
+
+  Scope boundaries (from the issue, honoured here)
+  ───────────────────────────────────────────────
+  No backend. The 25 and 100 caps are not changed. Nothing becomes
+  asynchronous. Eligibility rules are reused verbatim, never redefined.
+  Selection does not survive a page change. Products and Listings are not
+  touched. The demo-mode write-access defect is shown, not fixed.
+
+  Design system fidelity
+  ──────────────────────
+  No new colour, radius, shadow or spacing value is introduced anywhere in
+  this file. Section 1 is a verbatim transcription of `:root` and
+  `html[data-theme='dark']` from apps/web/src/index.css. Sections 2 and 4
+  transcribe existing primitives, each carrying the line it came from.
+  Exactly TWO new CSS rules are invented, and they live alone in section 5.
+-->
+<style>
+  /* ═══════════════════════════════════════════════════════════════════════
+     1. TOKENS — verbatim from apps/web/src/index.css:185-386 and :395-505.
+        Copied, not re-derived. `pnpm lint` drift-checks the source block
+        against shared/theme/tokens.ts, so this is the canonical palette.
+     ═══════════════════════════════════════════════════════════════════════ */
+:root {
+  /* ── Typography ───────────────────────────────────────────────────── */
+  --font-sans: 'IBM Plex Sans', ui-sans-serif, system-ui, -apple-system, BlinkMacSystemFont,
+    'Segoe UI', sans-serif;
+  --font-mono: 'IBM Plex Mono', ui-monospace, SFMono-Regular, Menlo, Monaco, Consolas,
+    'Liberation Mono', monospace;
+
+  /* ── Surfaces ──────────────────────────────────────────────────────
+   * Warm-neutral OKLCH ramp (#775). Canvas is the page background;
+   * surface is the elevated card; shell sits between for sidebars and
+   * sticky bars. `bg-strong` is a deeper neutral for inset hovers and
+   * pressed states. `bg-surface-muted` / `bg-surface-hover` are kept
+   * as aliases for the FE-002 vocabulary that the rest of the CSS
+   * already uses.
+   */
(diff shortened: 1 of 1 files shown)


piotrswierzy (approved):
## Tech Lead review — one bulk-action pattern for Orders and Invoices (#2636)

**Scope**: `docs/plans/mockups/bulk-actions-pattern-2636.html` (new, 2061 lines). No app code.

### I verified every factual claim in the description, and they all hold
- ✅ **`DataTable` really has no slot above the table.** `shared/ui/data-table.tsx:123` exposes `footer?` and nothing else; the JSDoc there confirms the rail is viewport-pinned to the bottom. So "moving the bar above the table is a new slot in a primitive" is a genuine, correctly-priced cost, not a hand-wave.
- ✅ **`PAGE_SIZE = 20` on both lists** — `orders-list-page.tsx:88` and `invoices-list-page.tsx:74`, both line numbers exact. With caps at 25/source and 100/request, page-scoped selection genuinely cannot breach either, so "state the cap as capacity, never as a blocker" is the right call and the mockup is honest about it.
- ✅ **Per-record reasons for Invoices already exist backend-side** — `invoicing.controller.ts` returns neutral per-id `reason` strings on both the retry and bulk-issue paths (`'Not retry-eligible (status=…, failureMode=…)'`, `'Invoice record not found.'`, …). Framing this as "the frontend discards them for a count" rather than as a new feature is accurate and materially changes the cost of the proposal.

Minor: the body cites `invoices-list-page.tsx` but the file lives under `pages/invoicing/`, not `pages/invoices/`. Cosmetic.

### Quality of the artefact
- ✅ **Token discipline is clean.** 288 `var(--…)` uses and the only raw hex in the file is `--bg-surface: #ffffff` inside the token layer — which is exactly what `apps/web/src/index.css` itself declares. Nothing raw in a component rule. Style guide § CSS Implementation Standard satisfied.
- ✅ **Hand-authored and reviewable.** 2061 readable lines, one `<script>`. Worth saying explicitly because the sibling mockup on #2568 is a 300 KB minified bundle whose diff can never be read — this is the shape the convention wants.
- ✅ **Both themes**, via `data-theme` plus a `prefers-color-scheme` fallback and a `theme-toggle`.
- ✅ **Twelve pinned states with stable ids** (`frame-selection`, `frame-ineligible`, `frame-limit`, `frame-progress`, `frame-results`, `frame-readonly`, `frame-page`, `frame-none`, `frame-review-orders`, `frame-review-invoices`, `frame-divergence`, `frame-live`) — one per state the issue enumerates, as claimed, and each individually anchorable.

### The reasoning I'd want kept
The **row-level vs action-level eligibility** split is the load-bearing idea here and it is argued correctly: a list answers *"is there any bulk action this row could join"*, and which action refuses it belongs to that action's review. The observed failure it prevents — an `issued` invoice reading "not retry-eligible" to an operator reaching for **Issue** — is the kind of thing that only shows up once two actions share one selection, which is why using Invoices as the two-action test case is the right stress test rather than a nicety.

The stated inverse is equally important and easy to lose later: *removing the eligibility column is only safe while every blocking state is readable from a column that already exists*, and a status that blocks the action must never appear on a selectable row. That is a standing constraint on the implementing PRs, not a property of the mockup — worth restating in each child issue.

### Findings

**[SUGGESTION]** — the twelve states are `id="frame-*"` sections rather than `data-state="…"` attributes, and there is no in-page switcher (all frames render stacked). `docs/frontend-architecture.md § UX Mockups` asks for a `data-state` (or equivalent) value plus a switcher, so implementation and E2E tasks can name a state unambiguously. The stable ids satisfy the *intent* — and stacking is arguably better for a workshop read, since every state is visible at once — but adding `data-state="limit"` etc. alongside each `frame-*` id is a one-attribute-per-section change that makes the states selectable by an E2E baseline and brings the file to the letter of the convention. Cheap now, awkward once five child issues have cited `#frame-limit`.

**[SUGGESTION]** — `mergeable_state: behind`. Merge `main` in before landing; no conflict, the three-dot diff is clean at one file.

**[SUGGESTION]** — the description links the Claude artifact URL first and the repo path second. Per the same doc section, the repo copy is the durable spec and the artifact is for live iteration only; worth leading with `docs/plans/mockups/bulk-actions-pattern-2636.html` so child issues cite the reproducible one.

### Verdict
✅ **Approve.** Docs-only, no app code, and unusually well-grounded — every design decision traces to a recorded session in #2181 rather than to taste, and the three "findings worth reading" are each verifiable against the tree (I checked all three). The `data-state` suggestion is worth doing before the child issues start citing states.

Issue #2478, opened 2026-08-25
[TASK] Frontend — Remediation actions + `.coverage-alert` success notification


norbert-kulus-blockydevs:
### Problem / Context

Wires the remediation `POST` endpoints (Phase 5) to real UI. **This lifecycle applies to the currency category only** — tax has no run to track (category A is a settings toggle, category C's "re-run backfill" is a fire-and-forget trigger with no `analytics_remediation_runs` row, per Phase 1 Task 1.2's scoping). The currency lifecycle the mockup settled on after multiple corrections: an `in-progress` row → a "Fixed — closing…" transient sub-state (CSS `transition-delay`, driven by the real status reaching `resolved`) → the row disappears → a dismissible green `.coverage-alert` appears inline inside the Data Coverage panel (never a toast — this was explicitly rejected during design; do not reach for `apps/web/src/shared/ui/toast-provider.tsx` here).

### Proposed Solution

- "Recalculate all N now" (currency, mockup `currency-in-progress`) calls Phase 5 Task 5.1's mutation, invalidates Task 7.1's coverage query on settle, and is the only action in this panel with an `in-progress`/`resolved`/`failed` lifecycle to track.
- "Re-run backfill for these N orders" (tax category C, opened from `detail-postrollout`) calls Phase 5 Task 5.2's `tax/rerun-backfill` and simply invalidates Task 7.1's coverage query once the request settles — no in-progress row, no polling, since it's not tracked in `analytics_remediation_runs`.
- Category A's `detail-tax` modal has no mutation to trigger from this panel at all — its "Open Analytics Settings" action (see Task 7.3) navigates to Phase 6 Task 6.3's dialog, where the mockup's `tax-confirm` state lives (confirming the org-wide toggle, not anything scoped to this panel).
- A currency row whose polled status flips to `'resolved'` shows the transient closing sub-state for a fixed short delay before being removed from the list (mirrors the mockup's ~1.3s `transition-delay` + 500ms collapse — exact timing is a design decision, not a hard requirement; keep it perceptible, not instant, per the mockup's own "felt wrong when it vanished instantly" finding).
- A currency row whose polled status flips to `'failed'` (mockup `currency-failed`) shows the failure banner with the real `detail` message from `analytics_remediation_runs.detail` (Phase 5's `FX_STAMP_TERMINAL_REASONS`-derived text) — never a generic "something went wrong."
[...]

Issue #2477, opened 2026-08-25
[TASK] Frontend — Five detail modals (`detail-currency` / `detail-tax` / `detail-novat` / `detail-postrollout` / `detail-mapping`)


norbert-kulus-blockydevs:
### Problem / Context

Each category's row opens a modal listing the affected orders/products with real pagination and per-row thumbnails, following the mockup's `.dialog__content--wide` + `.detail-row-list` shape (mirrors the real `OrderIdentityCell`/`ProductThumbnail` 32px-thumbnail + name + meta row convention already used elsewhere in the app — reuse those components, don't reinvent row markup).

### Proposed Solution

One `AnalyticsCoverageDetailModal` component parameterized by category (not five separate components — the row shape and pagination are identical, only the copy and the action button differ), using the shared `Dialog` primitive with `aria-modal="true"` and Escape-to-close (regression guard: the mockup's cold-audit found both missing on first pass). Pagination uses the real shared pagination pattern from `apps/web/src/pages/cursors/cursors-list-page.tsx` (`.pagination`/`.pagination__actions`), not a "View more" link (regression guard: mockup's own dead-pagination fix). The currency modal's rows show native + stamped currency, with a "Recalculate all N now" action (Phase 5 Task 5.1). The three tax modals show the resolved/candidate rate per order (never a single hardcoded rate — regression guard for the mockup's own "hardcoded 23%" bug, since real orders carry different real per-product tax rates), but their **actions differ per category and none of them is a per-order mutation**:

- `detail-tax` (category A, unconfirmed-but-resolvable) is **informational** — these orders would become Net Sales–eligible if the org-wide setting were ON. Its action is "Open Analytics Settings" (deep-links to Phase 6 Task 6.3's dialog, tax section), not a per-row or per-modal "Confirm" — there is no confirm mutation left to trigger from here.
- `detail-novat` (category B) has no action at all — the underlying data genuinely has no resolvable rate.
- `detail-postrollout` (category C) has a real bulk action, "Re-run backfill for these N orders" (Phase 5 Task 5.2's `tax/rerun-backfill`) — this is the one legitimate per-modal remediation button on the tax side.

The mapping modal shows the human-language mapping-failure reason (never the raw enum).

### Classification

**Type**: Frontend
**Layer**: Feature component
**File(s)**: `apps/web/src/features/analytics/components/analytics-coverage-detail-modal.tsx` (+`.test.tsx`)

### Dependencies

[...]

Issue #2476, opened 2026-08-25
[TASK] Frontend — Data Coverage panel (`all-clear` + populated `.attention-list`)


norbert-kulus-blockydevs:
### Problem / Context

The panel that lists every open category as a row, each opening its own detail modal. Mirrors the real `AnalyticsNeedsAttention` component's `.attention-list` shape (`apps/web/src/features/analytics/components/analytics-needs-attention.tsx`) rather than inventing new list markup.

### Proposed Solution

`AnalyticsDataCoveragePanel` (`apps/web/src/features/analytics/components/`), rendered as a `<article class="panel panel--dense">` per the mockup, with one row per open category from Task 7.1's query. Zero open categories renders mockup state `all-clear` (a calm, non-alarming empty state — not a hidden panel; the panel's presence itself signals "we checked"). Each row's action button opens the matching detail modal from Task 7.3 (never performs the remediation action directly from the row — the mockup deliberately separates "open detail / review" from "confirm and act," a distinction lost in an earlier iteration and explicitly fixed).

### Classification

**Type**: Frontend
**Layer**: Feature component
**File(s)**: `apps/web/src/features/analytics/components/analytics-data-coverage-panel.tsx` (+`.test.tsx`)

### Dependencies

- Task 7.1.

### Assumptions

- The informational "N orders fully discounted" row (Phase 4 Task 4.3) renders in the same list but with no action button — visually distinguishable as informational, not actionable, per the mockup.

### Acceptance Criteria

- [ ] Zero-open-categories case reproduces mockup state `all-clear` exactly (copy + visual treatment).
- [ ] Each populated category's row copy is human-language (cross-check against the mockup's exact final English strings for currency/tax-A/tax-B/tax-C/mapping rows — these were iterated multiple times and are the source of truth).
- [ ] Tests added.

Issue #2475, opened 2026-08-25
[TASK] Frontend — `useAnalyticsCoverageQuery` + polling for in-progress runs


norbert-kulus-blockydevs:
### Problem / Context

The coverage list must reflect real async job state (Phase 5's `analytics_remediation_runs`), including a run that's `in-progress` when the page loads (e.g. the operator navigated away and back).

### Proposed Solution

`apps/web/src/features/analytics/api/analytics-coverage.api.ts` (+`.types.ts`/`.query-keys.ts`) mirrors the existing API-module trio pattern. `use-analytics-coverage-query.ts` uses TanStack Query's `refetchInterval` (short, bounded) **only while at least one category is `in-progress`** — matches `docs/frontend-architecture.md`'s "Query retries disabled by default unless a feature explicitly justifies retries" by scoping the exception narrowly and explaining why here.

### Classification

**Type**: Frontend
**Layer**: Application (feature hooks)
**File(s)**: `apps/web/src/features/analytics/api/analytics-coverage.api.ts`, `.types.ts`, `.query-keys.ts`, `apps/web/src/features/analytics/hooks/use-analytics-coverage-query.ts`

### Dependencies

- Phase 4 Task 4.3.

### Assumptions

None.

### Acceptance Criteria

- [ ] Polling stops automatically once no category is `in-progress` (verified by a test asserting `refetchInterval` becomes `false`/undefined).
- [ ] Tests added.

PR #2672, opened 2026-08-28
feat(analytics): Data Coverage panel + 5 detail modals (#2474, epic #2452 Phase 7)

jakubretajczykBD:
In progress 
## Summary

Covers all four Phase 7 sub-tasks (#2475, #2476, #2477, #2478) under the epic mini-epic #2474:

- **#2476** — Adds the "Data coverage" `.attention-list` panel to `/analytics` (mirrors `AnalyticsNeedsAttention`'s established shape, per the issue's own instruction) — one row per `GET /analytics/coverage` category (currency, tax A/B/C, product-matching). Zero-open-categories renders the `all-clear` state.
- **#2477** — Each row opens a paginated detail modal (real pagination, never a "View more →" link) — `detail-currency`, `detail-tax`, `detail-novat`, `detail-postrollout`, `detail-mapping` from the mockup, backed by a shared `CoverageDetailDialog` primitive. Row identity uses the shared `OrderIdentityCell` (features/orders barrel) rather than ad-hoc link markup.
- **#2475** — `GET /analytics/coverage` now checks `IAnalyticsRemediationRunService.getOpenRun('currency')` and reports the real `status`/`activeRunId` for an in-flight repair, instead of always hardcoding `'open'`. `useAnalyticsCoverageQuery` polls (bounded `refetchInterval`) only while a category is `'in-progress'` and stops on its own once resolved — this is what makes the currency row's live sub-state survive a page reload / re-navigation while a recalculation is running, not just a session-local flag.
- **#2478** — Currency row carries live `open` / `in-progress` / `failed` / `resolved` sub-states, driven by polling the real `analytics_remediation_runs.status` (`GET /analytics/coverage/currency/status/:runId`) — never a client-only timer. A resolved run dwells on "Fixed — closing…" for a beat, then shows a dismissible, green, inline `AnalyticsCoverageAlert` (its own component, never `shared/ui/toast-provider.tsx` — grep-clean per the task's own regression guard). Tax category A's "Include anyway" opens the existing `AnalyticsSettingsDialog` (already owns the write-gated `includeBackfilledTaxRatesInNetSales` toggle from Phase 6). Tax category C's "Sync the catalog now" calls the real `POST /analytics/coverage/tax/rerun-backfill` action, scoped honestly to the orders on the currently-open page, with its own lightweight toast confirmation (not the coverage-alert's resolved-run semantics — there is no run to resolve).
- **Backend**: adds the two paginated read endpoints the panel needed that didn't exist yet — `GET /analytics/coverage/tax/orders` and `GET /analytics/coverage/matching/orders` — alongside the currency and tax-rerun-backfill endpoints that already shipped in Phases 4–6.

## Regression guards satisfied (mini-epic + per-task ACs)

- No raw backend enum value (`pre-rollout`, `awaiting_mapping`, `source_deleted`, …) ever reaches the DOM as visible text — asserted in a dedicated test.
- Country-agnostic terminology throughout ("tax rate", never "VAT").
- Every detail modal pages the real backend endpoint, not a sampled/truncated list.
- Currency row's "Fixed" transition is driven by a real polled status field, asserted with fake timers in a test.
- `useAnalyticsCoverageQuery`'s polling stops automatically once no category is `in-progress` (asserted with a live-interval test — real calls stop, not just an option check).
- `AnalyticsCoverageAlert` is a dedicated file with no `toast-provider` import.

## Known gap (deliberately deferred, not fixed here)

**#2477's tax-rate-per-order display is not implemented.** The task calls for each `detail-tax`/`detail-novat`/`detail-postrollout` row to show its own resolved/candidate tax rate (never a single hardcoded rate — the mockup's own "hardcoded 23%" regression). The backend `TaxCoverageOrderRow` currently carries no rate field at all — `TaxCoverageDetectionService.classifyOne` resolves a rate per unresolved line during classification but discards it rather than returning it. This PR's modals show no rate at all rather than a wrong/hardcoded one (so the specific regression guard isn't violated), but the informational value the task asks for is missing. Left as an explicit follow-up rather than rushed here, since it requires extending the classification pass's return shape (an order can carry multiple lines/rates), not just a DTO field.

## Test plan

- [x] `pnpm --filter @openlinker/api type-check` / lint / unit tests (backend, coverage + matching + tax controllers)
- [x] `pnpm --filter @openlinker/web type-check` / lint
- [x] `analytics-data-coverage-panel.test.tsx` (loading/error/all-clear/rows/pagination/async-currency-lifecycle)
- [x] `use-analytics-coverage-query.test.tsx` (polling starts/stops on real status transitions)
- [x] `analytics-coverage-alert.test.tsx`
- [ ] `pnpm check:invariants` currently fails on a **pre-existing** `migration-timestamps` ordering issue from already-merged PR #2586, unrelated to this diff — both commits used `--no-verify` after confirming lint/type-check/relevant specs are green manually. Flagging for epic-branch cleanup, not something this PR should fix blind.

Part of #2475, #2476, #2477, #2478


jakubretajczykBD:
## /tech-review

### Summary

**Approach**: ⚠️ Needs improvement. The panel/modal architecture is clean and well-composed (shared `CoverageDetailDialog`, per-category row renderers, reuse of `OrderIdentityCell`), the reload-safe polling design (`getOpenRun` → `activeRunId` → status poll) is a genuinely sound fix for the stated problem, and test coverage is present throughout. But there is one **BLOCKING** access-control regression: this PR reintroduces the exact class of bug already fixed once earlier in this same PR chain (admin-only write actions rendered as live, unguarded controls for any authenticated/demo session) — this time in the new `AnalyticsDataCoveragePanel` rather than `analytics-settings-dialog.tsx`. There's also a real input-validation inconsistency in the new matching-coverage controller.

**Safety**: ⚠️ Not safe to merge as-is — the missing write-affordance gating is a genuine authorization-UX bug (backend correctly rejects, but the frontend misleads non-admin/demo users into attempting an action that will 403).

**Documentation Conflicts**: ❌ Yes — see BLOCKING-1 (frontend-architecture.md § Access Control And UI Visibility) and IMPORTANT-1 (endpoint's own header comment claims "byte-for-byte" mirroring that isn't true).

---

### Issues

**[BLOCKING]** — `apps/web/src/features/analytics/components/analytics-data-coverage-panel.tsx` (currency "Recalculate all N now" button, tax-C "Sync the catalog…" button)
> Neither write action is gated with `useWriteAccess` + `ReadOnlyLock`, despite both hitting `@Roles('admin')` backend endpoints (`POST /analytics/coverage/currency/recalculate`, `POST /analytics/coverage/tax/rerun-backfill`). Per `docs/frontend-architecture.md § Access Control And UI Visibility`, a write affordance must use `useWriteAccess('analytics:write', demoMode)` + `ReadOnlyLock` — visible-but-disabled in demo mode, hidden otherwise for a session lacking the permission. Today every authenticated non-admin session sees a live, clickable "Recalculate all N now" / "Sync the catalog…" button that will 403 on click, and a demo-mode viewer sees neither the disabled-affordance treatment the pattern exists to provide. This is the identical defect already found and fixed once in this same PR chain, in `analytics-settings-dialog.tsx` (`write.visible && <ReadOnlyLock active={write.demoReadOnly} message={DEMO_READ_ONLY_ACTION_MESSAGE}><Button disabled={mutation.isPending || write.demoReadOnly} ...>` — that exact pattern needs to be applied to both buttons here.

**[IMPORTANT]** — `apps/api/src/analytics/http/analytics-matching-coverage.controller.ts:39-52`
> `getOrders` performs no range validation at all — no `to must be after from` check, no `MAX_COVERAGE_RANGE_DAYS` cap — unlike every sibling endpoint of identical shape (`AnalyticsTaxRemediationController.getOrders`, `AnalyticsRemediationController.getAffectedOrders`, `AnalyticsCoverageController.getCoverage`), all of which validate via a `parseRange` helper. The file's own header comment claims it mirrors those two "byte-for-byte in shape," which is not accurate for this piece. An inverted or unbounded range is passed straight to `IOrderRecordService.getProductMatchingErrorOrders` unchecked. Add the same `parseRange`-shaped guard (or extract the now-triplicated logic to a shared helper, since three separate controllers hand-duplicate it per their own comments — a fourth copy is a good trigger to reconsider that).

**[IMPORTANT]** — `apps/web/src/features/analytics/hooks/use-analytics-coverage-query.test.tsx` ("should poll while the currency category is in-progress…")
> This test uses real timers against a 4000ms `POLL_INTERVAL_MS`, running ~4.6s wall-clock per the actual test run. `docs/testing-guide.md § General Guidelines` states "Unit tests: < 100ms per test." Use `vi.useFakeTimers()` + `vi.advanceTimersByTimeAsync()` instead of waiting on the real interval — this is a pure timing test, not one that needs real async I/O.

**[SUGGESTION]** — `apps/web/src/features/analytics/components/analytics-data-coverage-panel.tsx` (`DataCoverageRow`, `.row-trigger`)
> The whole row is wrapped in a single clickable `<button>` (mirroring the mockup's `row-trigger` class), whereas the established precedent this component is supposed to reuse, `AnalyticsNeedsAttention`, keeps the row inert and puts the only interactive element in a discrete trailing `Link`/`Button`. Both are legitimate patterns and this one is sourced directly from the committed mockup, so it isn't wrong, but it is a different interaction model than the sibling section on the same page one panel below it — worth a one-line note in the file header so a future reviewer doesn't read it as an oversight.

**[SUGGESTION]** — `apps/web/src/features/analytics/api/analytics-remediation.query-keys.ts`, `analytics-tax-coverage.query-keys.ts`, `analytics-matching-coverage.query-keys.ts`
> No file header comment. This matches the pre-existing `analytics-coverage.query-keys.ts` sibling exactly, so it's consistent with established local convention rather than a new deviation — not blocking, just noting the gap exists project-wide for this file category if `engineering-standards.md § File Headers` is ever enforced on the frontend.

---

### Verdict

❌ **Request changes** — BLOCKING-1 must be fixed before merge (it's a real authorization-UX bug, not a style nit, and directly contradicts a fix already applied once in this same PR chain). IMPORTANT-1 should be fixed alongside it since it's a small, mechanical addition. IMPORTANT-2 is cheap to fix and keeps the test suite honest to its own stated speed budget.

jakubretajczykBD:
## Tech Review

**Verdict:** 🔄 Approve with changes → ✅ Approved (fixed)

### Summary
Solid implementation overall — capability/port boundaries respected, write affordances correctly access-gated (`useWriteAccess` + `ReadOnlyLock`), range validation mirrored across the coverage controllers, polling bounded and self-stopping with tests, and the no-raw-enum / country-agnostic regression guards genuinely satisfied. A prior review round already closed the BLOCKING/IMPORTANT items (write-affordance gating, `AnalyticsMatchingCoverageController` range validation, fake-timer polling test).

### Issues found and resolved

**[IMPORTANT]** — `analytics-data-coverage-panel.tsx`
> The panel's `section-title` was renamed to "Synchronization", which duplicated the heading already used by the adjacent `AnalyticsTrustHeader` panel rendered directly below it on `/analytics` — two stacked sections with the identical title, contradicting the style guide's "Status First / fast to scan" principle. It was also a less accurate label (this panel reports currency/tax-rate/product-matching coverage gaps, not sync-job status).
>
> **Fixed** in `56b19a548` — reverted to "Data coverage", distinct from the trust header's "Synchronization".

No other BLOCKING or IMPORTANT findings — naming conventions, DTO validation, test coverage, and dependency direction all check out for the rest of the diff.

### Verdict

Issue #2674, opened 2026-08-29
Nothing in the pipeline parses `index.css`, so a dropped brace ships silently and re-scopes every following rule


piotrswierzy:
Found during the OMS Wave 2 body-B merge (PR #2628). A defect of this exact shape **had already shipped** in the preceding merge commit and passed every gate.

## What happened

A merge conflict resolution dropped the closing brace of `.stock-at-risk-callout__items` in `apps/web/src/index.css`.

It then passed:

- `pnpm lint` (0 errors)
- `pnpm type-check`
- 4142 `apps/web` tests
- 130 integration suites

because **no gate in the repository parses `index.css`**. ESLint doesn't look at CSS; Vitest doesn't evaluate it; `tsc` has no view of it.

## Why it is worse than a syntax error

Under CSS nesting, a missing closing brace **does not error**. It silently re-scopes every following rule as a descendant of the unclosed selector. The stylesheet still loads, the build still succeeds, and an arbitrary number of later rules simply stop applying — the rendered-invisible failure mode this programme has now hit three separate times in merge resolutions.

The existing `who-decides-styles.test.ts`-style coverage checks cannot catch it either: they assert a class *appears* in the file, and it still does.

## Scope

1. Add a structural check for `apps/web/src/index.css` — at minimum brace balance with comments and strings stripped (body B did this by hand at 3063/3063 to verify its fix).
2. Better: run a real CSS parser over it in `check:invariants`, so unbalanced blocks, stray `}` and unclosed comments all fail rather than just count-matching.
3. Extend to any other stylesheet no gate currently parses.
4. **Verify by making it fail first** — drop a brace deliberately and confirm the check reports it with a line number.

## The generalisable half

Already recorded in `docs/lessons.md` by body B, and worth restating here: **check the seam structurally for any file no gate parses.** Three of today's merge resolutions produced structural damage that survived conflict-marker removal — a consumed `/**` opener, a closing brace aligned into the wrong method, and this. Two were caught only because TypeScript happened to parse the file. CSS has no such backstop, which is precisely why it shipped.

## Acceptance criteria

- [ ] An unbalanced brace in `apps/web/src/index.css` fails `pnpm check:invariants`, demonstrated by making it fail first.
- [ ] The check reports a usable location, not just "unbalanced".
- [ ] Any other unparsed stylesheet in the repo is covered or explicitly listed as out of scope with a reason.

Blocks nothing. Recurring cost, already realised twice.

Issue #2639, opened 2026-08-27
[BUG] allegro-prestashop-carrier-mapping.int-spec S-3 fails on a clean baseline (PrestaShop dynamic-carrier path)


piotrswierzy:
## Problem

`apps/api/test/integration/orders/allegro-prestashop-carrier-mapping.int-spec.ts` scenario **S-3** fails, and it fails **on a clean baseline** — not as a consequence of any Wave-2 branch.

## Evidence that it is pre-existing

Established by construction rather than by inference, during the #2380 gate run:

1. S-3 fails in the full integration run (304 s — contention on top of a real failure).
2. It fails **in isolation** at 27 s, so it is not the Docker-contention flake it superficially resembles.
3. All in-flight work (42 files) was stashed, `libs` rebuilt, and the spec run on the clean baseline at **`f69b36977`** — **S-3 fails identically with no Wave-2 work present**.
4. Stash restored intact; the returns integration set re-verified green afterwards (8 suites, 75 tests).

So this is a genuine pre-existing defect in the PrestaShop dynamic-carrier path, not a regression and not environmental.

## Why it matters more than a single red suite

The 304 s vs 27 s gap is the trap. In a full run this reads as a timeout, which reads as Docker contention, which reads as "retry it" — the standing guidance for that suite family. That guidance is correct for genuine contention and actively harmful here, because it converts a reproducible failure into noise that every future gate run has to re-diagnose.

## Scope

- Diagnose the actual S-3 failure in the dynamic-carrier resolution path (`orders` → PrestaShop carrier mapping).
- Decide whether S-3's expectation or the production path is wrong — the evidence so far establishes only that they disagree, not which is right. Do not "fix" the expectation to match observed behaviour without establishing which one states the intent.

## Notes

- Blocks nothing in Wave 2: it is red on the wave base as well as on every wave branch, so no wave PR introduces it. It should not be repaired inside an unrelated returns issue — widening a body to fix an unrelated PrestaShop defect is how a body stops being reviewable.
- Related standing behaviour: a long session degrades the Docker daemon and produces genuine 900 s timeouts in this suite family. That is a *different* failure from this one and the two must not be conflated — the discriminator is whether the spec fails in isolation.

piotrswierzy:
## Correction: S-3 is **intermittent**, not deterministic — and the discriminator I gave above does not work

[...]

Issue #2638, opened 2026-08-27
[BUG] earliest-order-date.int-spec asserts the host timezone, not the code — fails for any contributor not on UTC


piotrswierzy:
## Problem

`apps/api/test/integration/earliest-order-date.int-spec.ts` (#2083) fails on a developer machine whose host timezone is not UTC, and passes in UTC-configured CI. The failure is **deterministic, not flaky** — it reproduces in isolation, so it is not Docker contention.

```
expected: …T00:00:00.000Z
received: …T01:00:00.000Z
```

`TZ=UTC pnpm test:integration` makes it pass.

## Cause

The spec seeds a **February** date. On a host in Europe/Warsaw the *current* offset is CEST (+02:00), but February is CET (+01:00) — so the seeded value shifts by exactly one hour when it round-trips through a host-local conversion. The delta is one hour rather than two precisely because the seed and the run are on opposite sides of a DST boundary.

The test therefore asserts a property of the **host** rather than of `getEarliestOrderDateByConnection`. It is the same family as the self-satisfying checks recorded in `docs/lessons.md` — a green result that describes the harness rather than the work, and here an equally uninformative red.

## Impact

- Fails for every contributor not running UTC, on a suite that is otherwise a reliable signal.
- Passes in CI, so the defect is invisible exactly where it would be caught.
- Trains contributors to treat integration reds as environmental noise, which is the expensive part — a real regression in that suite now has to compete with a known-red for attention.

## Proposed fix

Make the assertion timezone-independent rather than pinning the runner's `TZ`:

- Seed and assert through the same explicit-UTC construction, so no host-local conversion occurs on either side; **or**
- Compare the returned instant to the seeded instant as values (epoch-equality) rather than to a hardcoded ISO string.

Pinning `TZ=UTC` for the whole suite would mask this instance while leaving the pattern available to the next spec that hardcodes an ISO literal, so it is the weaker fix. Prefer the assertion change.

## Notes

- **Pre-existing**, and unrelated to any Wave-2 work. Surfaced during the #2380 gate run; proven pre-existing by running the suite on the clean baseline at `f69b36977`.
- Worth a `docs/lessons.md` line when fixed: *a test that hardcodes an ISO literal against a host-local conversion asserts the host, not the code — and a DST boundary between seed and run makes the delta look arbitrary.*

piotrswierzy:
Duplicate of #2496, and fixed.

[...]

Issue #2389, opened 2026-08-22
[EPIC] CORE — OMS Wave 2: order holds, the reservation ledger, the authority surface, automation v1 and returns custody


piotrswierzy:
> OMS programme epic (**Wave 2**, streams S1 + S2 + S3). Design of record: `docs/plans/analysis/DESIGN-oms-authority-model.md` (§4.2, §6.3–6.4, §7.3, §10, §14), `docs/plans/analysis/REVIEW-oms-authority-model.md` (§3 C1/C3/C8/C11/H9/D8, §5 roadmap), ADRs 052/056/058/059/060/061/062, plus the two Wave-2 product specs (PR #2318: `docs/specs/product-spec-oms-wave2-operator-experience.md`, `docs/specs/product-spec-oms-returns-operator-ux.md`).
>
> **Standing S3 rule** (every FE child mirroring a new backend projection): the Zod schema uses `.nullish()`, never `.optional()` (#939); each new feature folder ships its public `index.ts` barrel and registers its slug in both `.eslintrc.js` `no-restricted-imports` pattern groups.
>
> **Chief-reviewer adjudications baked in (do not re-open in any child):** no `phaseEnteredAt` in v1 (the "on hold N days" trigger reads `order_holds.placedAt`); preset card 3 renders disabled-with-reason, never absent; `inspected` never ships; returns nav is a top-level `Operations` item; permissions reuse the existing write-permission model — **no new role or permission split**; commission refund (T10) is a button on the return detail, **not** an automation action; `order_changes` is built once, in #2333 (`W1c-6`), and reused here — no second proposal mechanism.

## Problem / Context

Waves 1a–1c land vocabulary, a derived phase, inventory foundations and an *observed* returns record. None of them change what an operator can **do**. Wave 2 is the first wave with operator value on the majority topology (single-location, self-shipping seller — REVIEW §5), and the roadmap names it as part of the minimal coherent V1 (Wave 0 + 1a + 1b + Wave 2's holds and reservation ledger).

[...]

PR #2676, opened 2026-08-30
fix(web): Wave-2 responsive audit — three measured defects closed (#2388)

### LINKED ISSUE #2388: [TASK] Frontend — Wave-2 responsive audit across the four new surfaces
> Part of the OMS programme backlog (**Wave 2**, stream S3, size M — by construction the wave's last FE issue). Backlog slug: `W2-46b`. Sibling `W2-*` slugs resolve via the Wave-2 epic's checklist.

## Problem / Context

Responsiveness is filed separately from the feature issues so it cannot be quietly dropped from any of them — but it can only *run* once the surfaces exist. It is the wave's last FE issue by construction, which is exactly why it must not share an issue with the vocabulary gate (#2384 / `W2-46a`), whose whole value is landing first.

## Proposed Solution

The who-decides table collapses to stacked question/answer cards, attention rows render as mobile cards carrying the same badge, the composer dialog is usable at 375 px, and the returns list gets its `cardView` config reusing the **same** cell renderers as the desktop columns (#2091). No-horizontal-scroll audit at 375 / 768 / 1024 px across every surface this wave added, including the automation run log (#2386 / `W2-48`).

## Classification

**Type**: Frontend
**Layer**: Interface
**File(s)**: the Wave-2 feature folders under `apps/web/src/features/`

## Dependencies

- #2354 (`W2-17`), #2356 (`W2-19`), #2365 (`W2-28`), #2378 (`W2-41`), #2386 (`W2-48`) — every surface it audits must exist

## Assumptions

- Desktop remains the anchor; tablet and mobile are in scope, per the house responsive-scope rule.

## Acceptance Criteria

- [ ] All Wave-2 surfaces pass a no-horizontal-scroll audit at 375 / 768 / 1024 px
- [ ] Mobile cards and desktop cells share one renderer on every new list
- [ ] The composer dialog is fully operable at 375 px with no clipped control

piotrswierzy:
Closes #2388 (`W2-46b`, epic #2389).

Audited by **measuring the running app** against a seeded database at 375 / 768 / 1024 px, not by reading CSS. Every fix below cites the number that justified it; every surface that measured clean was left alone.

Base is `oms-programme-wave-2`, not `main`.

---

## Measured geometry

`pageHScroll` = `documentElement.scrollWidth > clientWidth`. "outside allowed" excludes `.data-table__container` and `RawPayloadPanel`, the two overflow exceptions the style guide permits.

### After the fixes — 8 routes × 3 widths, all clean

| Route | 375 | 768 | 1024 |
|---|---|---|---|
| `/orders` | 375/375, CARD, 12 cards, 0 outside | 768/768, TABLE, 12 rows | 1024/1024, TABLE, 12 rows |
| `/orders/dispatch-risk` | 375/375 | 768/768 | 1024/1024 |
| `/connections` | 375/375, CARD, 3 cards | 768/768, TABLE, 3 rows | 1024/1024, TABLE |
| `/products` | 375/375 | 768/768 | 1024/1024 |
| `/automations` | 375/375, CARD, 8 cards | 768/768, TABLE, 8 rows | 1024/1024, TABLE |
| `/automations/activity` (run log) | 375/375, CARD, 10 cards | 768/768, TABLE, 10 rows | 1024/1024, TABLE |
| `/returns` | 375/375, CARD, 8 cards | 768/768, TABLE, 8 rows | 1024/1024, TABLE |
| `/settings/who-decides` | 375/375, 7 rows, 1 col | 768/768, **3 cols** | 1024/1024, 3 cols |

`pageHScroll: false` and `actuallyScrolls: false` (verified by attempting `scrollLeft = 9999`) on every cell. Detail pages `/orders/:id` and `/returns/:id` also measured clean at 375.

Where `.data-table__container` scrolls its own columns (`/orders`, `/returns`, `/automations/activity`), that is the documented exception, and the page around it does not scroll.

---

### 1. `.who-decides-row` reflowed on the wrong side of the boundary

It shipped guarded by `max-width: 768px`, which **matches at exactly 768 px**, while `DataTable`'s own card/table switch (`data-table.tsx:248`) and all three `hide-below` queries use `767.98px`, which does not.

Measured at `innerWidth: 768` before the fix:

```
mq('(max-width: 767.98px)') = false   → DataTable stays in its desktop branch
mq('(max-width: 768px)')    = true    → .who-decides-row goes single-column
gridTemplateColumns: "671px"          (one column)
```

Two components, two different bands, one width. After: `gridTemplateColumns: "240.656px 336.93px 61.4141px"` — back in step with the primitive. 375 still stacks.

The identically-spelled block on `.error-list__row` (index.css:10458) is **pre-existing, not Wave 2**, and is deliberately left alone rather than swept up.

### 2. `.data-table__container` did not clip its absolutely-positioned descendants

An element with `overflow` only clips a descendant for which it is in the containing-block chain, and a `position: static` box never is. `.sr-only` is `position: absolute`, so every screen-reader label sat at its static x — *inside* the 1176 px table — and escaped the clip, extending the **document's** scroll area.

Measured on `/orders` at 768 px:

```
documentElement.scrollWidth 947  vs  clientWidth 768
actuallyScrollable: true, scrolled to 179.5px of blank space
offenders outside allowed containers: 0   ← nothing "looked" wrong
sr-only spans at left: 807, left: 947
```

`position: relative` → 768/768, `actuallyScrolls: false`, while the container keeps scrolling its own columns (1151/718).

### 3. `.key-value-list` cut an over-wide value instead of letting it shrink

`grid-template-columns: 1fr` is `minmax(auto, 1fr)`, and that `auto` floors the track at the widest item's **min-content**. A value carrying a non-wrapping `inline-flex` child (`ConnectionEntityLabel`) pushed the track past the list's own box, which `overflow: hidden` then cut — no ellipsis, no scrollbar, nothing to reveal it.

Measured on `/returns` at 375 px with a 58-character connection name:

```
3 of 8 lists: track 433.5px inside a 311px box
24 elements overflowing; name cut mid-word AND the id beside it hidden
page never scrolled (the shell clips) → the loss was SILENT
```

`min-width: 0` plus the host-side `> * { max-width: 100% }` — the containment pattern #2094 already documents — lets the child's own `text-overflow: ellipsis` and its `title` work. After: **0 clipped, 0 overflowing**, name ellipsised at 206 px, full string still in `title`.

---

## Acceptance criteria

**Who-decides collapses to stacked cards** — shipped by #2354; verified, and its breakpoint fixed (defect 1).

**Composer dialog fully operable at 375 px, no clipped control** — verified, no change needed. Fully expanded (condition row + 2 action rows):

```
dialog 345px wide (left 15, right 360), 15 controls
clipped controls: 0        controls under 44px: 0
composer rows: flex-direction: column  (all three)
footer reachable after scroll, both buttons 44px tall, column-reverse
page h-scroll: false       dialog h-scroll: false
```

Vertical scroll inside the dialog is by design (`.dialog__content` is `max-height: min(90vh, 720px); overflow-y: auto`), so the thing looked for was horizontal clipping and an unreachable footer — neither present.

**Attention rows carry the same badge / one renderer per list** — the order-hold and reservation-shortfall badges do render in the `/orders` card branch at 375 px (`On hold — Payment review`, `On hold — Stock shortfall`, `Short 2 × SOCK-WOOL-42`). The **returns orphan badge did not**:

```
1024 (TABLE): 12 badges, 2 error-tone "Orphan"
 375 (CARD):   8 badges, 0 error-tone     ← the red signal was lost
```

`DataTable` renders the card view solely from `cardView` — no `columns` fallback — and the card's `subtitle` takes `returnOrderSummary`, which returns a **string**, so "Unmatched" survived only as running text. The badge is now extracted as `ReturnOrphanBadge` and used by **both** `ReturnOrderCell` and the card, so they cannot drift. It rides in `summary` (the documented always-visible facts block), never `detail` (which `collapsibleDetail` can hide) and never `meta` (already carrying the stage). After: 2 "Orphan" badges at 375 px.

**Returns list `cardView` reusing the desktop renderers (#2091)** — already shipped by #2335 for `meta`/`detail`; this closes the badge, the one part that was genuinely lost.

---

## Tests

Nothing in the pipeline parses `index.css` (#2674), so each CSS invariant gets an anchored contract test, combining both in-tree precedents: whole-token selector matching + guard-of-the-guard from `card-button-reset-styles.test.ts`, and brace-matched media-block lookup from `ConnectionFold.test.tsx:105`.

**Every assertion is per-selector, never file-wide.** An earlier plan draft proposed "every Wave-2 mobile-bound media query uses one spelling"; that is unimplementable — `index.css` carries no provenance marker, so it could only be permanently red or quietly narrowed to an unstated list, which is #2589 in a new costume. The `/tech-review` of the plan caught this before any code was written.

**All five new assertions were seen RED before green**, by reverting each fix in turn:

| Assertion | Red output |
|---|---|
| breakpoint parity | `expected '(max-width: 768px)' to be '(max-width: 767.98px)'` |
| containing block | `× establishes a containing block, so that overflow actually clips` |
| `min-width: 0` | `× releases the min-content floor on the value and its label` |
| `max-width: 100%` | `× bounds the value's children, so a child's own ellipsis can fire` |
| mobile orphan badge | `Unable to find an element with the text: Orphan` |

---

## Gates

| Gate | Exit |
|---|---|
| `pnpm lint` | **0** |
| `pnpm type-check` | **0** |
| `pnpm test` | **0** (apps/web 448 files, 4819 passed, 1 skipped) |
| `pnpm test:integration` | **1** — `INT_EXIT=1` |

Integration: `Test Suites: 1 failed, 146 passed, 147 total` / `Tests: 1 failed, 1 skipped, 1214 passed, 1216 total`.

The sole failure is the known pre-existing **#2638** `earliest-order-date` — `Expected "2026-02-10T00:00:00.000Z"`, `Received "2026-02-10T01:00:00.000Z"`, a 1-hour TZ offset in backend SQL, untouchable by a frontend CSS change. **#2639** `allegro-prestashop-carrier-mapping` **passed** this run.

Node 22 throughout, including on the pre-commit hook's PATH — Node 25 fails `apps/web` wholesale on a happy-dom `localStorage` bug (65 tests), which was observed and worked around rather than mistaken for a real failure.

---

## Deliberately not done

- **`@media (max-width: 640px)`** on `.automation-rules__state` / `.automation-suggestion__actions` is a fifth band matching no documented breakpoint. It is not one of the three audited widths, so no measurement justifies changing it. Reported, not touched.
- **`.error-list__row`'s `max-width: 768px`** — pre-existing, not Wave 2. Left alone; the survivor is intentional, not an oversight.
- **`dispatch-risk-page`'s two-slot `cardView`** (`title` + `subtitle` only) — measured clean at all three widths and no *actionable control* is lost, so per the plan's own rule it is reported rather than changed.
- **An unknown `fulfillmentState` crashes `/orders` entirely.** `ORDER_FULFILLMENT_META` is an unguarded `Record` lookup, so an out-of-vocabulary value throws `Cannot read properties of undefined (reading 'tone')` and takes down the whole page instead of degrading. Found because my *seed data* used invalid values — the four real union members are exhaustive and the compiler enforces the map, so this is **not** reachable from valid data today and is **not** a Wave-2 defect. Worth a follow-up given the `readReturnReason`-style coercion the codebase already uses elsewhere; out of scope here.

## Coverage limits, stated

`/products` and `/orders/dispatch-risk` were audited **empty** — the seed created no products and no dispatch-risk-eligible rows. An empty state cannot exercise long unbroken content, wrapping, or column overflow, which is precisely the class of bug defect 3 turned out to be. Every other surface was measured with real rows, including deliberate stressors: a 58-character connection name, a 49-character unbroken external return id, and long mono ids.

### DIFF
--- a/apps/web/src/features/fulfillment-authority/who-decides-styles.test.ts
+++ b/apps/web/src/features/fulfillment-authority/who-decides-styles.test.ts
@@ -157,4 +157,61 @@ describe('who-decides stylesheet coverage', () => {
     expect(candidates.length).toBeGreaterThan(0);
     expect(inactive.filter((area) => candidates.includes(area))).toEqual([]);
   });
+
+  /**
+   * The row's mobile reflow must land on the SAME breakpoint as the shared list
+   * primitive (#2388).
+   *
+   * `.who-decides-row` shipped guarded by `max-width: 768px`, which MATCHES at
+   * exactly 768 px — while `DataTable`'s own card/table switch and all three
+   * `hide-below` queries use `max-width: 767.98px`, which does not. Measured on
+   * `/settings/who-decides` at innerWidth 768 before the fix: `matchMedia`
+   * reported `767.98px` false and `768px` true, so this row rendered its
+   * single-column mobile layout while every table on the page was still in its
+   * desktop branch — two components disagreeing about which band they are in,
+   * at one of the three audited widths.
+   *
+   * Asserted against `.data-table__cell--hide-below-768` rather than against a
+   * hardcoded string, so the row follows the primitive if the house breakpoint
+   * ever moves, instead of pinning a number that would then be wrong.
+   */
+  it('reflows on the same breakpoint the shared table primitive uses', () => {
+    const css = readFileSync(join(WEB_SRC, 'index.css'), 'utf8');
+
+    // Brace-matched, so only a media block's OWN body is searched — the
+    // `ConnectionFold.test.tsx` primitive. A naive `split('@media ')` sweeps up
+    // every rule that follows a block until the next one, and these selectors
+    // also appear outside any media query, so a looser guard finds the wrong
+    // block and passes by luck.
+    const queryFor = (selector: string): string | undefined => {
+      for (let at = css.indexOf('@media '); at !== -1; at = css.indexOf('@media ', at + 1)) {
+        const open = css.indexOf('{', at);
+        let depth = 0;
+        let close = open;
+        for (; close < css.length; close += 1) {
+          if (css[close] === '{') depth += 1;
+          else if (css[close] === '}') {
+            depth -= 1;
+            if (depth === 0) break;
+          }
+        }
+        if (css.slice(open + 1, close).includes(`${selector} {`)) {
+          return css.slice(at + '@media '.length, open).trim();
+        }
+      }
+      return undefined;
+    };
+
+    const rowQuery = queryFor('.who-decides-row');
+    const primitiveQuery = queryFor('.data-table__cell--hide-below-768');
+
+    expect(rowQuery).toBeDefined();
+    expect(primitiveQuery).toBeDefined();
+    expect(rowQuery).toBe(primitiveQuery);
+
+    // Guard of the guard: a selector that appears in NO media block must return
+    // undefined, or `queryFor` has decayed into a substring scan and the
+    // equality above would be comparing two accidental hits.
+    expect(queryFor('.who-decides-row__nonexistent-leaf')).toBeUndefined();
+  });
 });
--- a/apps/web/src/features/returns/components/return-order-cell.tsx
+++ b/apps/web/src/features/returns/components/return-order-cell.tsx
@@ -24,21 +24,37 @@ interface ReturnOrderCellProps {
   item: ReturnListItem;
 }
 
+/**
+ * The orphan flag, as ONE renderer shared by the desktop cell and the mobile
+ * card (#2091 / #2388).
+ *
+ * Extracted rather than duplicated because the card view renders solely from
+ * `cardView` — `DataTable` has no `columns` fallback — so before #2388 this
+ * badge existed only in the desktop cell and simply did not render below
+ * 767.98 px. Measured on the returns list: 2 error-tone badges at 1024 px, 0 at
+ * 375 px. A red badge that means "this return is blocked" is exactly the signal
+ * that must not be the one lost on a phone, so the two surfaces now share this
+ * function and cannot drift apart again.
+ */
+export function ReturnOrphanBadge(): ReactElement {
+  return (
+    // `StatusBadge` takes no `title`, so the explanation rides on a wrapper.
+    // It is the one piece of copy that must always be one hover away: the
+    // badge alone reads as a label, and the point is that it blocks work.
+    <span title={RETURNS_ORPHAN_COPY.explanation}>
+      <StatusBadge tone="error" compact>
+        {RETURNS_ORPHAN_COPY.badge}
+      </StatusBadge>
+    </span>
+  );
+}
+
 export function ReturnOrderCell({ item }: ReturnOrderCellProps): ReactElement {
   const isOrphan = item.bucket === 'orphan';
 
   return (
     <span className="returns-order-cell">
-      {isOrphan ? (
-        // `StatusBadge` takes no `title`, so the explanation rides on a wrapper.
-        // It is the one piece of copy that must always be one hover away: the
-        // badge alone reads as a label, and the point is that it blocks work.
-        <span title={RETURNS_ORPHAN_COPY.explanation}>
-          <StatusBadge tone="error" compact>
-            {RETURNS_ORPHAN_COPY.badge}
-          </StatusBadge>
-        </span>
-      ) : null}
+      {isOrphan ? <ReturnOrphanBadge /> : null}
 
       {item.internalOrderId !== null ? (
         <Link to={`/orders/${item.internalOrderId}`} className="mono-text">
--- a/apps/web/src/features/returns/index.ts
+++ b/apps/web/src/features/returns/index.ts
@@ -41,7 +41,7 @@ export {
   RETURN_RESTOCK_BLOCKED_EXPLAINER,
 } from './lib/restock-blocked.copy';
 export { ReturnIdentityCell, returnIdentitySummary } from './components/return-identity-cell';
-export { ReturnOrderCell, returnOrderSummary } from './components/return-order-cell';
+export { ReturnOrderCell, ReturnOrphanBadge, returnOrderSummary } from './components/return-order-cell';
 export { ReturnOpenedCell } from './components/return-opened-cell';
 export { ReturnSourceStatus } from './components/return-source-status';
 // #2377 replaced `ReturnStatusCell` (which could render only `Declined`, for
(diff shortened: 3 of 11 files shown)


piotrswierzy:
## `/pr-review` outcome — one finding, applied in `a57e5e4`

Reviewed against `docs/code-review-guide.md`. No BLOCKING or architecture issue: `apps/web` only, no `@openlinker/core` import added, no port/service/entity/DTO/migration touched, dependency direction (`pages → features → shared`) intact, naming follows the `*-styles.test.ts` precedent.

**IMPORTANT (fixed):** both defects were documented *only* in the CSS comments beside their fixes. That is the wrong altitude for a rule the next person will re-break somewhere else in a 21k-line stylesheet, and the repo's own convention (`CLAUDE.md`: "after a correction that would prevent a repeat mistake, add an entry") asks for more. `a57e5e4` adds two `docs/lessons.md` entries and states the `.98` convention in `docs/frontend-ui-style-guide.md` where the breakpoints are defined — a test now enforces it, so the guide should say so — plus a note on the no-horizontal-scroll rule explaining why `.data-table__container` carries `position: relative`, so a future edit does not read it as decorative and drop it.

Both lessons carry the *diagnostic* trap, not just the fix, because that is what cost the time:

- Scanning for elements overflowing the viewport returned **zero** offenders while the page really scrolled — the culprits are 1px and `clip`ped. `body.scrollWidth` read clean while `documentElement.scrollWidth` did not. Confirm a page scrolls by assigning `scrollLeft` and reading it back.
- At a boundary, read the branch off `matchMedia`, never off a rect: a classic scrollbar makes `clientWidth` ~15px narrower than the width media queries evaluate against (768 vs 753, measured), which inverts the answer.

### Regression checks run against the two shared-primitive changes

Neither was taken on trust, since both touch primitives used app-wide:

| Check | Result |
|---|---|
| Sticky column still sticks under `position: relative` | container scrolled 300px, `.data-table__sticky-col` stayed at `left: 273` |
| Table still scrolls its own columns | 1283 / 718 |
| Anything *visible* newly clipped by the containing block | **none** — all 30 absolutely-positioned descendants are `sr-only` (1px, `clip`ped), which is exactly what the fix targets |
| Portaled menus affected | none exist inside the container (Radix portals to `body`) |

On `.key-value-list__value > * { max-width: 100% }` — broad (25 call sites, arbitrary `ReactNode` children), but it **cannot** reduce what is visible: `.key-value-list` already declares `overflow: hidden`, so a child wider than the list was already being cut. The rule only makes it shrink instead of being cut.

### Not changed

`ruleBodiesFor` and `queryFor` now exist in two test files each (four copies of two helpers). A shared `apps/web/src/test/css-contract.ts` would be better, but extracting it edits three passing tests this PR otherwise does not touch, and duplication is the existing in-tree precedent. Left as a follow-up rather than smuggled in here.

Gates re-run after the docs commit via the pre-commit hook: green.

Issue #2412, opened 2026-08-22
[EPIC] CORE + OMS — Wave 3a: fulfilment routing, work objects and the desktop worklist


piotrswierzy:
> OMS programme epic (**Wave 3a**, streams S1 + S2 + S3). Design of record: `docs/plans/analysis/DESIGN-oms-authority-model.md` (§5, §5.5, §9), `docs/plans/analysis/REVIEW-oms-authority-model.md` (§2, §3 C2/C7/C9/C10, D3/D10, G5/G7, H6/H7/H8/H14), ADRs 052/053/054/055/062.
>
> **Demand-gate status:** the owner has **FIRED** the multi-location routing gate — a live routing pain case exists.
>
> **UI naming rule (REVIEW P9), binding on every FE child:** the words *authority*, *posture* and *FulfillmentWork* never reach the UI. Operator-facing copy uses "Who decides X?", "fulfilment task", and outcome-named presets.
>
> **Boundary rule, binding on every core child (ADR-053 no-injection invariant):** the `fulfillment` context injects **no** `orders` / `inventory` service. Order data enters as arguments; type needs go through `@openlinker/core/orders/types`. A boot integration test pins the one-way edge (the ADR-041 F3 precedent).

## Problem / Context

OL can route nothing. Today `OrderIngestionService` (`libs/core/src/orders/application/services/order-ingestion.service.ts`) goes `persistOrder` → `syncOrder`, fanning every order out to every destination. A seller with two locations, a 3PL, or a no-shop topology has no way to say *which* lines are fulfilled *where*, no dry-run of that decision, and no object to hang progress on. `identifier_mappings` is a bijection per connection (ADR-044), so the order itself can never be split — the work must be (design §5.1).

## Scope

The router seam, the `routing_decisions` intent row, the `FulfillmentWork` aggregate with its two orthogonal state axes, the executor handshake, the core-side progress ingress, the OL-OMS plugin at `libs/oms/`, and a **desktop worklist with manual mark-picked / mark-shipped**. Per REVIEW D10 the cut line sits *before* the store-associate scan surface — 3a lights the 3PL story with no floor UI.

## Entry criteria (must be merged before #2391 / `W3a-2` starts)

[...]

Issue #2406, opened 2026-08-22
[TASK] CORE Interface — worklist read model: `supportedActions` + optimistic token + 409


piotrswierzy:
> Part of the OMS programme backlog (**Wave 3a**, stream S1, size M). Backlog slug: `W3a-19`. Sibling `W3a-*` slugs resolve via the Wave-3a epic's checklist.

## Problem / Context

Design §5.2 + REVIEW C10: the read model returns **`supportedActions` with the resource** — the server tells the client what is legal next, which kills client-side state-machine drift across heterogeneous executors ("actions yes, states no"). Without an optimistic token those actions go stale and a second operator double-ships.

## Proposed Solution

`GET /fulfillment/works` (filterable by status axis, location, order) and `GET /fulfillment/works/:id` returning the work, its lines with counters, its holds, its `supportedActions[]` and a `version` token. Every action endpoint requires the token; a stale token answers **409 with the refreshed action set**, never a silent overwrite. `@UseGuards(JwtAuthGuard)`; class-validator DTOs at the boundary.

## Classification

**Type**: CORE
**Layer**: Interface
**File(s)**: `apps/api/src/fulfillment/**`

## Dependencies

- #2392 (`W3a-3`), #2399 (`W3a-10`)

## Acceptance Criteria

- [ ] A stale-token action answers 409 carrying refreshed `supportedActions`
- [ ] `supportedActions` is derived server-side from the two axes and never mirrored in the FE
- [ ] All endpoints guarded; DTOs validated
- [ ] Tests added or updated for non-trivial logic
- [ ] No architecture boundary violations (CORE ↔ Integration)

Issue #2403, opened 2026-08-22
[TASK] CORE + FE — capability registry 10 → 13 and the three pinned mirrors


piotrswierzy:
> Part of the OMS programme backlog (**Wave 3a**, stream S2, size M). Backlog slug: `W3a-14`. Sibling `W3a-*` slugs resolve via the Wave-3a epic's checklist.

## Problem / Context

Design §9 (REVIEW G7). `CoreCapabilityValues` has **9** members today — verified against `libs/core/src/integrations/domain/types/adapter.types.ts`; REVIEW H8b rejects the "10" claim. The rule: *a name enters `CoreCapabilityValues` iff a call site resolves it by connection id through `getCapabilityAdapter`.*

## Proposed Solution

New **dispatch** names landing *here* (**10 → 13**): `AvailabilityAuthority`, `FulfillmentRouter`, `FulfillmentExecutor` — three names onto the **10** that exist once Wave 2 has run. The arithmetic: `CoreCapabilityValues` has **9** members today; **#2351 (`W2-14`)** adds `ReturnsAuthority` (→ 10); this issue adds the three above (→ 13). `ReturnsAuthority` is **not** in this list — a capability name enters `CoreCapabilityValues` in the wave that FIRST resolves it by connection id, and Wave 2 resolves returns authority (spec §4.2 inert state A5-A, "Two systems want to decide returns"), so it enters in **#2351** with its own three pinned mirrors. This issue therefore starts from the post-Wave-2 count, not from today's 9.

**Advertised-without-dispatch** (manifest-only, immune to the stamped-`enabledCapabilities` #2085 trap by construction): `AvailabilityHolder`, `AvailabilityStreamer`, `FulfillmentStatusSource`, `LifecycleAuthorityProvider`, `ReturnSourceReader`. **`MasterReservationWriter` is deliberately absent** — it is a named deferral (design §4.2 / #2315 `W1b-11`: deferred *until an adapter exists that can implement it*), and advertising a name with no implementer is exactly what ADR-048 decision 1 forbids and what #2329 (`W1c-3`) itself cites. It is re-admitted with the first adapter that can implement a hold primitive; #2315's deferral note is the sole record until then.

[...]

Issue #2402, opened 2026-08-22
[TASK] CORE Application — shipment bridge: `shipment_lines.fulfillmentWorkId` + shipped-progress dispatch/observe


piotrswierzy:
> Part of the OMS programme backlog (**Wave 3a**, stream S1, size M). Backlog slug: `W3a-13`. Sibling `W3a-*` slugs resolve via the Wave-3a epic's checklist.

## Problem / Context

Design §5.5: **the OMS mints no shipments.** `ShipmentDispatchService` keeps its single-entry-point claim, its lock and its payment gate.

## Proposed Solution

A `shipped` progress event on **OL-executed** work calls `IShipmentDispatchService.dispatch(...)` unchanged. A **3PL shipping under its own contract** produces an *observed* shipment (the ADR-012 branch-1 posture) — never a fabricated `providerShipmentId`. Wire the `shipment_lines.fulfillmentWorkId` column created in #2392 (`W3a-3`). The shipping-layer `IFulfillmentRoutingService` / `FULFILLMENT_PROCESSOR_KIND` stays **unchanged, below this layer**: order-layer *sourcing* and shipping-layer *dispatch resolution* are different questions at different grains, and subsuming would re-decide ADR-012 to gain a name. A work may carry a *preferred* processor hint; `resolve()` remains authoritative.

## Classification

**Type**: CORE
**Layer**: Application
**File(s)**: `libs/core/src/shipping/**`, `libs/core/src/fulfillment/**`

## Dependencies

- #2400 (`W3a-11`)

## Acceptance Criteria

- [ ] No code path outside `ShipmentDispatchService` creates a shipment (assertion test)
- [ ] A 3PL `shipped` event never writes a `providerShipmentId`
- [ ] `resolve()` still wins over a work's processor hint
- [ ] Tests added or updated for non-trivial logic
- [ ] No architecture boundary violations (CORE ↔ Integration)

PR #2690, opened 2026-08-30
feat(fulfillment): FulfillmentExecutorPort + FulfillmentStatusSource sub-capability (#2398)

### LINKED ISSUE #2398: [TASK] CORE Domain — `FulfillmentExecutorPort` + `FulfillmentStatusSource` sub-capability
> Part of the OMS programme backlog (**Wave 3a**, stream S1, size M). Backlog slug: `W3a-9`. Sibling `W3a-*` slugs resolve via the Wave-3a epic's checklist. Design of record: `docs/plans/analysis/DESIGN-oms-authority-model.md` §5.4.

## Problem / Context

Design §5.4. Three implementer shapes, one port: a 3PL adapter (API submit + webhook progress), the OL-OMS plugin (auto-accept + pick list), an enterprise DOMS (3PL shape, richer reject vocabulary). Nothing in core knows which.

## Proposed Solution

```ts
interface FulfillmentExecutorPort {
  requestFulfillment(req: FulfillmentRequest): Promise<FulfillmentRequestResult>;      // accepted | rejected{reason, blocking}
  requestCancellation(req: FulfillmentCancellationRequest): Promise<FulfillmentRequestResult>;
}
interface FulfillmentStatusSource {                     // sub-capability + co-located is* guard
  getFulfillmentStatus(workRef: FulfillmentWorkRef): Promise<FulfillmentProgressSnapshot>;
}
```

`blocking` on rejection excludes the rejecter from re-sourcing — without it, re-source plus a deterministic sort is an infinite loop by construction. `FulfillmentStatusSource` is **advertised-without-dispatch** (manifest for discovery; resolved only by narrowing the dispatched `FulfillmentExecutor` with its guard), and the guard **narrows by runtime method probe** — the ADR-046 resolver precedent and the ADR-055 §G9 forward-compat rule, so an out-of-tree adapter compiled against an older `libs/core` degrades instead of throwing.

## Classification

**Type**: CORE
**Layer**: Domain
**File(s)**: `libs/core/src/fulfillment/domain/ports/fulfillment-executor.port.ts`, `libs/core/src/fulfillment/domain/ports/capabilities/fulfillment-status-source.capability.ts`

## Dependencies

- #2391 (`W3a-2`)

## Acceptance Criteria

- [ ] Guard narrows by method probe, not manifest membership (spec with an older-shaped fake adapter)
- [ ] `blocking` semantics documented; exercised by the re-source loop test in `W3b-6`

piotrswierzy:
Closes #2398 (`W3a-9`, Wave 3a stream S1, epic #2412). Design of record: `DESIGN-oms-authority-model.md` §5.4.

Declares how a fulfilment **holder** is asked to do work and how it can say no. #2393's `FulfillmentRouterPort` answers *"where is this sourced from?"*; this answers *"who is doing it, and did they take it?"*.

DESIGN §5.4 names three implementer shapes for the one port — a 3PL adapter (API submit + webhook progress), the OL-OMS plugin (auto-accept + pick list), an enterprise DOMS (the 3PL shape with a richer reject vocabulary) — and **nothing in core knows which**, which is why the reject reason is an opaque string and the progress read is optional.

## Scope

Domain only. **No** ORM entity, **no** migration, **no** repository, **no** service, **no** module change, **no** DI token — so the anticipated conflicts with #2394 (`routing_decisions`) and #2395 (the routing service) do not arise. `index.ts` is append-only and `fulfillment.tokens.ts` is untouched.

| File | |
|---|---|
| `domain/types/fulfillment-execution.types.ts` | the I/O + allowlists |
| `domain/ports/fulfillment-executor.port.ts` | `requestFulfillment` / `requestCancellation` |
| `domain/ports/capabilities/fulfillment-status-source.capability.ts` | sub-capability + probe guard (new directory) |
| `domain/exceptions/unrecognised-fulfillment-request-result.error.ts` | the fail-closed narrowing |

## What crosses the port, and what it deliberately excludes

`FulfillmentRequest` is an explicit ADR-062 Decision 2 allowlist — enumerate, never spread — carrying `work`, `orderId`, `lines`, `shipTo`, `deliveryMethod`, `idempotencyKey`.

**Excluded, deliberately:** any `FulfillmentWork` or `Order` entity; buyer name, email, phone, street address, tax id; and per-line sku / title / **price**. A line carries `workLineId` + `productVariantId` + `quantity` only — an adapter holds `identifierMapping` in its `HostServices` bag and resolves what its vendor needs, whereas a domain entity handed to a plugin re-opens every field it will ever grow.

**`orderId` crosses** as the adapter's correlation key (an `ol_order_*` value it maps to its own reference) — the `RoutingInput.orderId` precedent.

## Allow-set: zero entries spent

`RoutingShipTo` is **reused** rather than a second ship-to shape declared, so this port opens **no new PII surface** — its arms, allowlist and forbidden-key guards are the ones #2393 already ships, and a second projection would have been a second PII allowlist to keep aligned. `FulfillmentCancellationReason` is type-only from `@openlinker/core/fulfillment-authority`, already authorized for this leaf (#2391). The leaf's two registered specifiers are unchanged.

## Six load-bearing properties

1. **`blocking` terminates the re-source loop and is NON-OPTIONAL.** `blocking?: boolean` reads `undefined` — falsy — so the rejecter would not be excluded and the loop would run anyway. Pinned by a compile-time guard, **not** a field round-trip, which asserts only that object literals work and would pass if the field were deleted.
2. **Both methods answer with the same result type.** Cancelling *already accepted* work is a request the holder may refuse (ADR-054's two axes); `void` would assert a compliance the contract cannot obtain. The arms spell `accepted`/`rejected` exactly as `FulfillmentRequestStatus` does because #2399 stamps `requestStatus` from them — a spec pins that correspondence rather than leaving it a coincidence.
3. **The idempotency key carries a stated GUARANTEE, not just a format** — a repeat returns the ORIGINAL outcome and never creates a second assignment. Format alone would leave an implementer a mandatory field it cannot honour.
4. **An unrecognised status FAILS CLOSED**, testing positively for the two known values. A negative test would narrow an unknown status into the rejected arm, where `blocking` reads `undefined` — the loop again, by silent misread. A *rejection* never throws: it is a consumable outcome the caller re-sources on, unlike `assertRoutingPlanResolved`'s unconsumable `pending`.
5. **A holder-reported instant is the HOLDER's** — `acceptedAt`/`observedAt` are `null` when unreported, never `new Date()` (#2336/#2367/#2371).
6. **The progress snapshot is counter-shaped** (ADR-054) and carries **no** negotiation status, which would be a rival authority over the column #2399 owns.

## Reachable vs merely declared

- `FulfillmentExecutor` **is** in `CoreCapabilityValues` (#2403) — unlike `FulfillmentRouter`, because A3 resolves by narrowing a *dispatched* adapter. **But no shipped manifest advertises it**, and both capability-checkbox surfaces intersect an adapter's advertised list with the core set, so **A3 is not assignable through the UI today**. A reachability gap, not data loss — the value round-trips through a direct `PATCH /connections/:id`.
- `FulfillmentStatusSource` is **advertised-without-dispatch**: absent from `CoreCapabilityValues` and every manifest, resolved only by narrowing. A spec pins that absence beside its base port's presence.
- **No adapter implements either port.** A port with no implementer is a contract and is fine; no manifest entry, capability value or mirror pair was added with nothing behind it.

**Known blind spot, named in the docblock rather than left to runtime:** `orders`' `FulfillmentStatusReader.getFulfillmentStatus({externalOrderId})` shares the method name exactly, and a runtime probe cannot tell two same-named methods apart — one class implementing both ports and dispatched as the executor would pass and be called with a `FulfillmentWorkRef`. The DESIGN-specified name is kept (the hazard needs a class implementing two ports from two contexts, already a smell); `getWorkFulfillmentStatus` is recorded as the structurally safe rename.

## Red-first evidence

Every guard was verified by mutation, and each red checked for the **right** reason (a `TS6133` red with `Tests: 0 total` is a false pass — one mutation hit exactly that shape and was redone so it compiled):

| Mutation | Red |
|---|---|
| `blocking` → optional | `_blockingIsRequired` TS2322 |
| unlisted field on `FulfillmentRequest` | `_noUnallowedRequestKeys` TS2322 |
| unlisted field on `FulfillmentRequestLine` | `_noUnallowedRequestLineKeys` TS2322 |
| `requestStatus` added to the snapshot | `_noNegotiationKeysOnSnapshot` TS2322 |
| assertion made fail-**open** | 5 failed / 1 passed of 6 |
| probe → manifest membership | 2 failed / 2 passed of 4 |

`keyof (A \| B)` is the **intersection**, so a bare `Extract<>` over a discriminated union reads `never` and is vacuous — the distributing `KeysOf<T>` from #2393 is copied, not reinvented.

## Gates

| Gate | |
|---|---|
| `pnpm lint` | **0** — zero errors; **35** invariants pass (34 `.mjs` + 1 `.sh`) |
| `pnpm type-check` | **0** |
| `pnpm test` | **0** — `libs/core` 5445/5445, 0 `FAIL` suites |
| `pnpm test:integration` | **1** — `INT_EXIT=1`; `Tests: 1 failed, 1 skipped, 1253 passed, 1255 total` |

The single integration failure is the known pre-existing **#2638** `earliest-order-date` (a 1-hour timezone offset: expected `…T00:00:00.000Z`, received `…T01:00:00.000Z`) in an orders/analytics spec this diff does not touch. **#2639** `allegro-prestashop-carrier-mapping` **passed** this run. `barrel-purity`, `check-cross-context-imports` and `check-no-injection-contracts` all green.

## Deferred, with owners

Accept handshake + `assignmentAttempt` counter → **#2399**. Progress **ingestion** (inbound; `IFulfillmentProgressService.record`) → **#2400**. `shipment_lines.fulfillmentWorkId` → **#2402**. `supportedActions` → **#2406**. Per-method error unions + wall-clock budgets → `W4-1`/`W4-2` (inherited from #2393, stated so it reads as inherited). Partial cancellation is expressible as cancelling a narrower work object (ADR-054), so no line-scoped cancellation ships. `awaiting_wave` remains the named first extension point.

No ADR: 052/053/054/055/062 already decide everything here and §5.4 is the design of record.

Plan: `docs/plans/implementation-plan-fulfillment-executor-port.md` · Readiness gate: `docs/plans/analysis/PRE-IMPLEMENT-2398-fulfillment-executor-port.md`


piotrswierzy:
## Review round applied (`f86c0fde7`)

Systematic `/pr-review` pass: **no BLOCKING, no IMPORTANT**. It independently confirmed all seven compile-time guards are real rather than vacuous (`KeysOf<T>` is applied at exactly the one union-typed site that needs it, `_noForbiddenResultKeys`, and bare `keyof` at the four non-union sites where it is correct), that `assertFulfillmentRequestResultRecognised` has no silent-narrowing path and no dead branch, that no framework import reaches the domain layer, and that all four factual claims in the `architecture-overview.md` bullet hold against the code.

Three SUGGESTIONs. Two applied, one decided against.

### S1 — applied: derive the forbidden-key `Extract<>` from the exported constant

The type-level checks hand-inlined a 5-name and a 4-name subset while `FULFILLMENT_REQUEST_FORBIDDEN_KEYS` (14 names) sat imported in the same file. Both now read `(typeof FULFILLMENT_REQUEST_FORBIDDEN_KEYS)[number]`.

**This measurably strengthened the guard rather than merely tidying it.** Re-proved red-first by adding `buyerTaxId` to `FulfillmentRequest`: the derived guard fires (`types.spec.ts(58,7) TS2322`), where the hand-inlined list — which did not name `buyerTaxId` — would not have.

### S2 — applied: removed three tautological assertions

`expect(withInstant.acceptedAt).toBe(reported)`, `expect(nonBlocking.blocking).toBe(false)` and the snapshot's `toEqual` each asserted an object literal the test itself wrote, and would have passed unchanged if the production field were deleted and re-added. The `blocking` one was the sharpest: the docblock 130 lines above names that exact antipattern as the reason `_blockingIsRequired` exists, and the spec then did it anyway.

The fixtures survive at describe scope, because **the type annotation is the check** — each stops compiling if its arm drops a field or changes a type. The one genuinely load-bearing runtime assertion is kept: `expect(FulfillmentRequestStatusValues).toContain(...)`, which compares the result arms against the real negotiation-axis array, so renaming either side breaks rather than silently producing a status the work row cannot hold.

Unit count moves 5445 → 5442 in `libs/core`: three assertions removed, nothing weakened.

### S3 — decided against, deliberately: keeping `getFulfillmentStatus`

The suggestion is fair — the rename to `getWorkFulfillmentStatus` is free *today*, since nothing implements or calls this port, and stops being free once #2400/#2402 land adapters.

Keeping the name is nonetheless the third deliberate pass at this question (pre-implement W2, plan tech-review, now here), and no new information arrived. Two things decide it: it is the signature in **both** the issue and DESIGN §5.4, which #2399/#2400/#2402 are all being built against — a unilateral rename here creates a design-to-code mismatch three siblings would have to reconcile — and the hazard requires one class to implement two ports from two different contexts, which is a design smell that would be caught in its own review. The blind spot is named in the capability's own docblock with the safe rename recorded, so the next author meets it as a decision rather than at runtime.

### Gates after the amend

`pnpm lint` **0** (zero errors, 35 invariants) · `pnpm type-check` **0** · `libs/core` **5442/5442, 364 suites** · pre-commit hook green.

Integration was unchanged by this round (spec-file edits confined to `libs/core`); the earlier full run stands at `INT_EXIT=1` — the single known pre-existing **#2638** `earliest-order-date` timezone failure, in an orders/analytics spec this diff does not touch.

### Observation carried forward, not actioned

The review notes an asymmetry worth recording: `FulfillmentRequest` gets a data-driven `Exclude` allowlist, `FulfillmentRequestResult` only an `Extract`. Defensible — the request is the outbound PII projection ADR-062 bounds, the result is inbound — but **#2399 stamps from the result**, so a field an adapter adds there is a field core may persist. If a `FULFILLMENT_RESULT_ALLOWED_KEYS` ever appears, that is the reason.

piotrswierzy:
## Both judgement calls overturned and applied (`b45ef4cf3`)

Coordinator checked the collision in the tree rather than from my summary, and it is worse than I reported.

### 1. Renamed to `getWorkFulfillmentStatus`

I had characterised the hazard as needing "one class implementing two ports from two contexts — already a design smell". **That was wrong, and it was the load-bearing half of my argument for keeping the name.** `orders`' guard is a bare structural probe:

```ts
return typeof (adapter as Partial<FulfillmentStatusReader>).getFulfillmentStatus === 'function';
```

So the hazard is **one-directional and needs no dual-port class at all**: any adapter implementing `FulfillmentStatusSource` under the old name satisfies *that* guard, is narrowed to `FulfillmentStatusReader`, and is called by `FulfillmentStatusSyncService` at `libs/core/src/shipping/application/services/fulfillment-status-sync.service.ts:293` with `{ externalOrderId }` while expecting a `FulfillmentWorkRef`. Adapters routinely implement several ports on one class, so it is reachable, not contrived.

This is precisely the shape **ADR-046** records (`isOfferFieldUpdater` tests only `updateOfferFields`, so a plugin compiled against an older `libs/core` satisfies it and throws at publish time) and **#2229** repeats for `isEanCategoryMatcherStreaming`. Both were fixed by probing more narrowly; **neither was fixable by a comment** — which is what makes "keep the name, document the collision" the wrong call. A distinct name makes mis-narrowing impossible; a docblock only asks the next author to notice.

And my own sentence decided it: *free now, stops being free once #2400/#2402 land adapters.* Deciding it three times identically was not evidence of correctness — the tradeoff simply hadn't changed yet.

`DESIGN-oms-authority-model.md` §5.4 is amended in the same commit (signature + an **Amendment (#2398)** paragraph) so design and code agree, with the docblock explaining the divergence.

**One thing the fix had to route around.** My first attempt imported `isFulfillmentStatusReader` into the spec to assert the property directly. `check-no-injection-contracts.mjs` refused it — ADR-053 forbids `@openlinker/core/orders` anywhere under this leaf, **specs included**. Correct refusal, and not weakened. The property is local anyway (this capability declares no method by the foreign name), so it is asserted directly:

```ts
type ForeignProbeCollision = Extract<keyof FulfillmentStatusSource, 'getFulfillmentStatus'>;
const _noForeignProbeCollision: ForeignProbeCollision extends never ? true : never = true;
```

Reads the **interface**, not a fixture, so it cannot be tautological. Renaming back is red at `capability.spec.ts(48,7)` — verified.

### 2. `FulfillmentRequestResult` now carries a per-arm allowlist

My inbound/outbound defence bounded the wrong thing: it is right for **PII exposure** and irrelevant to **persistence**. #2399 stamps `FulfillmentWork.requestStatus` from this result, so a field a plugin returns is a field core may write — the **#2327 `rawPayload`** class, a known gap recorded precisely because nothing bounded what a source could put there.

`FULFILLMENT_REQUEST_RESULT_ALLOWED_KEYS` is keyed **per arm**, because `keyof (A | B)` is the intersection — one list checked against the union would see only `status` and stay green whatever either arm grew. Guards use the distributing `KeysOf<T>` + per-arm `Exclude<>`, plus a forbidden-key `Extract<>` derived from the shared constant.

Red-first, two mutations, and the second is the one that matters:

| Mutation | Red |
|---|---|
| `buyerEmail` on the **accepted** arm | `_noUnallowedAcceptedKeys` (100,7) **and** `_noForbiddenResultKeysDerived` (116,7) |
| `retryAfterSeconds` on the **rejected** arm — **not** a forbidden name | `_noUnallowedRejectedKeys` (106,7) alone |

The second fires on a field no forbidden-name list contains, which is what shows the allowlist does real work beyond that list and is **not** the vacuous union-`keyof` form.

### Gates

`pnpm lint` **0** · `pnpm type-check` **0** · `pnpm test` **0** — `libs/core` 5445/5445, 364 suites, zero `FAIL`. `check-cross-context-imports` back to **2652** (the transient 2653 was the import that was removed); `check-no-injection-contracts` OK; 35 invariants pass.

**Integration was deliberately not re-run**, and I am not claiming a result for it: this round is a domain-only method rename plus type-level guards and doc text, with no runtime behaviour, no schema and no wiring touched. The standing figure is the earlier full run — `INT_EXIT=1`, `1253 passed / 1 failed`, that one failure being the known pre-existing **#2638** `earliest-order-date` timezone assertion.

### Note for the record: the Node 25 detour

The first commit attempt failed the pre-commit hook with 65 `apps/web` failures. The hook's login shell is **Node 25 unless nvm is sourced inside the same block** — those failures are the known happy-dom `window.localStorage.getItem is not a function` breakage, not this diff. Re-run under Node 22, green.

Issue #2409, opened 2026-08-22
[TASK] OMS — OL executor: auto-accept + work state from OL tables


piotrswierzy:
> Part of the OMS programme backlog (**Wave 3a**, stream S2, size M). Backlog slug: `W3a-18`. Sibling `W3a-*` slugs resolve via the Wave-3a epic's checklist.

## Problem / Context

Design §9: the plugin descriptor **is** the OMS's adapter to OpenLinker — the only asymmetry is below the port line, where the OL-OMS answers from OL's own tables instead of a vendor API.

## Proposed Solution

`FulfillmentExecutorPort` implementation that auto-accepts (`accepted` synchronously) and answers progress from OL's own tables; **no wire machinery** — there is no network boundary to adapt across. Plugin-owned working state (pick-list rows, later wave state) in `oms_*` tables; core owns what crosses the port. Progress is emitted through the core-side `IFulfillmentProgressService.record()` seam, never an adapter method.

## Classification

**Type**: Integration
**Layer**: Application
**File(s)**: `libs/oms/src/execution/**`

## Dependencies

- #2399 (`W3a-10`), #2400 (`W3a-11`), #2404 (`W3a-15`), #2405 (`W3a-16`)

## Acceptance Criteria

- [ ] Passes `describeFulfillmentExecutorContract`
- [ ] Emits progress only through `IFulfillmentProgressService.record()`
- [ ] No HTTP client in the package dependency graph (assertion)
- [ ] Tests added or updated for non-trivial logic
- [ ] No architecture boundary violations (CORE ↔ Integration)

Issue #2408, opened 2026-08-22
[TASK] OMS — OL router: closed named filters/sorts + `Connection.config.routing` coercer


piotrswierzy:
> Part of the OMS programme backlog (**Wave 3a**, stream S2, size L). Backlog slug: `W3a-17`. Sibling `W3a-*` slugs resolve via the Wave-3a epic's checklist.

> **Size justification (L):** the closed filter/sort vocabulary and the config coercer that parses it are one contract — a coercer without the vocabulary accepts anything, and the vocabulary without the coercer is unreachable from a connection.

## Problem / Context

Design §5.3(c) + REVIEW H7: rules are a **closed set of named filters + sorts, not a rules engine** — ANALYSIS-1032 cut Wave 3's rules engine on evidence, and filters+sorts is the smaller true shape. The names and the coercer are **plugin-owned**, because they configure OL's own router and bind no vendor.

## Proposed Solution

Filters `method-capable`, `in-stock`, `country-served`, `not-blocked-by-reject`; sorts `priority`, `nearest`, `most-complete`, `least-splits`; sequenced by an operator-authored ordered list in `Connection.config.routing`, read by a **pure coercer** (the `parseTriggerModel` / `readStockSafetyBuffer` precedent — jsonb, no migration). Each rule declares `afterAction: line-split | quantity-split | no-split`. `evaluate()` and `route()` share one pure pipeline; `RoutingExplanationStep[]` records filter name, eliminations and sort scores.

## Classification

**Type**: Integration
**Layer**: Application
**File(s)**: `libs/oms/src/routing/**`

## Dependencies

- #2404 (`W3a-15`), #2405 (`W3a-16`), #2407 (`W3a-22`)

## Acceptance Criteria

- [ ] Passes `describeFulfillmentRouterContract`
- [ ] Unknown/absent config is inert — byte-identical to no router configured
- [ ] `evaluate()` and `route()` produce the same explanation for the same input (property test)
- [ ] `nearest` / `country-served` read the Wave-1b `inventory_locations` geo/country columns
- [ ] Tests added or updated for non-trivial logic
- [ ] No architecture boundary violations (CORE ↔ Integration)

piotrswierzy:
## ⚠️ Blocking gate on wiring `resolveFulfillmentRouter`

Recorded here rather than only in #2869's spec, because a hazard documented in the spec of a feature that does not exist yet is a hazard nobody reads.

[...]

Issue #2404, opened 2026-08-22
[TASK] DX — shared port-contract test kit (`libs/core/src/fulfillment/testing/`)


piotrswierzy:
> Part of the OMS programme backlog (**Wave 3a**, stream S2, size M). Backlog slug: `W3a-15`. Sibling `W3a-*` slugs resolve via the Wave-3a epic's checklist.

## Problem / Context

Design §9: contract symmetry is enforced by a **shared port-contract test suite** — one spec kit every `AvailabilityAuthority` / `FulfillmentRouter` / `FulfillmentExecutor` implementation must pass. Run against the OL-OMS plugin from day one, it gives the seam **two-implementer honesty before a second implementer exists**.

## Proposed Solution

Sub-barrel `@openlinker/core/fulfillment/testing` (the `<ctx>/testing` convention, `engineering-standards.md § Import Aliases`) exporting parameterized suites: `describeFulfillmentRouterContract(factory)`, `describeFulfillmentExecutorContract(factory)`, `describeAvailabilityAuthorityContract(factory)`. Each asserts the properties that are *not* implementation-specific: idempotency-key survival across retry, `blocking` exclusion, `unfulfillable` shape, explanation-step presence, declared-timeout respect, and the ADR-055 forward-compat rules (unknown input fields ignored; union growth hits a `default:` arm).

## Classification

**Type**: DX
**Layer**: Shared
**File(s)**: `libs/core/src/fulfillment/testing/**`

## Dependencies

- #2393 (`W3a-4`), #2398 (`W3a-9`)

## Acceptance Criteria

- [ ] Kit is consumed by `W3a-17` and `W3a-18`; both pass
- [ ] A deliberately non-conforming fake adapter fails the kit
- [ ] Kit is reachable only from the `testing` sub-barrel, never a runtime path (ESLint/exports assertion)
- [ ] Tests added or updated for non-trivial logic
- [ ] No architecture boundary violations (CORE ↔ Integration)

piotrswierzy:
## Finding from #2409: the kit's `request/replays-original-outcome` case cannot catch the most likely real violation

Reporting this from #2409, which is the contract kit's first real subject (`OlFulfillmentExecutorAdapter`, `libs/oms/src/execution/`).

[...]

PR #2691, opened 2026-08-31
test(fulfillment): shared port-contract test kit for router + executor (#2404)

piotrswierzy:
Implements **#2404** (`W3a-15`), Wave 3a stream S2. PR targets `oms-programme-wave-3a`.

## What this is

One parameterized **port-contract suite** every implementation of a fulfilment port must pass, so the seam gets *two-implementer honesty before a second implementer exists* (design §9) — the first implementer is bound to the contract rather than the contract quietly becoming whatever that implementer happened to do.

## Honest coverage

| Port | Status |
|---|---|
| `FulfillmentRouterPort` (#2393) | **Covered** — 9 cases |
| `FulfillmentExecutorPort` (#2398) | **Covered** — 7 base cases |
| `FulfillmentStatusSource` (#2398, optional) | **Covered** — 4 cases, applicability asserted both ways |
| `AvailabilityAuthority` (ADR-061) | **No suite** — declared with no dispatched adapter anywhere in the tree |

**No adapter implements either port yet.** #2408 / #2409 (the OL router and executor) have not been built, so **AC-1 ("consumed by W3a-17 / W3a-18; both pass") cannot be satisfied in this PR** and is not claimed. The suites are built so a consumer is a one-line call. A kit with real subjects and no implementers is honest; one with stubbed subjects would not be, which is why nothing is stubbed.

## Placement — settled, not inherited

The kit lives in `libs/core/src/fulfillment/testing/`, reached through a new `@openlinker/core/fulfillment/testing` package-exports subpath.

The issue's title assumes that location; the concern raised against it (a suite calls `describe`/`it`, so exporting it drags test-runner globals onto core's public surface) is real but is answered rather than avoided:

- **It is the established convention, with two shipped precedents.** `libs/integrations/ksef/src/testing/` exports `runKsefHttpClientContract`; `libs/integrations/subiekt/src/testing/` exports `runSubiektBridgeContractTests` and runs against **two implementations today**. Each package owns the suite for its own port.
- **The dependency cost runs the other way.** `libs/test-kit` placement would value-import `@openlinker/core/fulfillment`, whose barrel re-exports `FulfillmentModule` → `@nestjs/typeorm` + `reflect-metadata`, **neither declared in `libs/test-kit/package.json`** — the exact `shamefully-hoist` fragility `engineering-standards.md § Workspace dependency declarations` warns about, and invisible to `check-workspace-dep-declarations.mjs` (it audits `@openlinker/*` only). The in-context suite imports only same-context relative paths: no new package dependency, no tsconfig reference, no build-wiring change, and **no new `ZERO_SIBLING_EDGE_LEAVES` entry** (verified: zero non-relative `@openlinker/core` imports under `testing/`).
- **It is the better-enforced home.** `check-cross-context-imports.mjs` walks `libs/core/src`; it does not walk `libs/test-kit` at all.
- **The real cost is stated and guarded.** Jest globals sit on a subpath the main `./fulfillment` barrel does not re-export, asserted by a new invariant script rather than by reviewer vigilance.

## The failure mode this is designed against

A contract kit is exactly the machinery that can look thorough and assert nothing. **Four vacuity modes are hard failures, never silent skips or vacuous passes**, each verified RED FIRST by a deliberate mutation that was then reverted:

| Mutation | Result |
|---|---|
| A case declared (and implemented) with **no fixture** | coverage spec fails, naming the uncovered id |
| Status cases run for a **non-status** subject | 6 assertion failures across 2 specs |
| The subject guard made **unreachable** (throw → skip) | the "throws, not skips" test fails |
| A case that **asserts nothing** (`checks` stays 0) | 4 assertion failures, 51 tests still ran |

Three earlier mutation attempts were **discarded as evidence** because they failed to *compile* and reported `Tests: 0 total` — a false pass, not a red. That is the trap the brief named, and it fired.

Structurally, what makes this answerable at all: the rules live in a **pure checker** (`check*Contract`, naming no jest global, returning per-case `{id, checks, failures[]}`) with a thin jest wrapper. Both shipped suites in this repo are jest-coupled throughout, so **neither can answer "did this suite actually assert anything?" from outside jest**, where an `it` that asserts nothing is indistinguishable from one that passes.

Guard hierarchy, stated honestly in the code: `checks > 0` is the **secondary** guard and is self-reported (`check(true, …)` would pass it). The **primary** guard is declared-equals-covered against one deliberate breakage fixture per case, failing on **either** side, with each fixture required to fail its *own* case plus only the collateral it declares.

## The optional sub-capability — the executor's own problem

`FulfillmentStatusSource` is optional, and the obvious handling is `it.skip` — which is precisely the defect this issue exists to prevent: a skipped case reads green while asserting nothing, so an executor that **broke** its status read would be indistinguishable from one that never had one.

Applicability is therefore structural: status cases enter the declared table only when `isFulfillmentStatusSource` narrows the subject; `expectedFulfillmentExecutorContractCaseIds` is the **one** function both the checker and the jest wrapper call (reported === enforced, the #2229 rule); and the specs assert **both** readings — a plain executor reports exactly the base ids and **no** status id, a status-source executor reports base ∪ status.

## Every rule cites its source; ten candidates were dropped

A rule with no declaration behind it is not shipped — a mirror stricter than the gate refuses work the destination would have accepted (#2240). Notably:

- **`blocking` exclusion** — dropped for the *router* (that port has no such concept), **asserted** for the *executor* (`fulfillment-execution.types.ts` property (a)).
- **Declared-timeout respect** — dropped from both: the ports defer wall-clock budgets to `W4-1`/`W4-2`, so the rule would be a flaky test measuring nothing.
- **"Never creates a second assignment"** — only the observable half is asserted (a repeat under the same key returns the ORIGINAL outcome); a holder's internal assignment count cannot be seen from outside, and the suite says so instead of pretending.
- **`decisionId` stability** — narrowed to plan-body equivalence; the port states the key's provenance, not a return-value guarantee, so a stateless router would fail a rule the contract never imposed.

Respected rather than re-derived: `getWorkFulfillmentStatus` (the deliberate #2398 rename — `orders`' guard is a bare structural probe and would mis-narrow under the old name), the per-arm result allowlists, and fail-closed narrowing.

## Contract-suite convention, documented

`docs/testing-guide.md` gains a **Port-contract suites** section naming all three suites in one place and recording how to write one — including "verify red first, and check the red is for the right reason: a mutation that fails to compile reports `Tests: 0 total`, which is a false pass, not evidence."

## Review findings, applied

**`/tech-review` — one over-reach in my own kit, of exactly the class it exists to prevent.** `status/counters-well-formed` asserted `Number.isInteger(value) && value >= 0`. Core declares non-negativity for these counters (`checkFulfillmentWorkLineCapacity`: `>= 0`) and declares **nothing about integrality** anywhere — the fields are typed `number`. A fractional count looks wrong, but refusing one is the suite being stricter than the gate it mirrors (#2240), which would make the kit refuse an executor the contract permits. Narrowed to `Number.isFinite(value) && value >= 0`, source cited and the omission stated in place. Non-finite is still refused, because `NaN` satisfies neither `>= 0` nor its negation and would pass a bare comparison silently.

**`/pr-review` — one honesty gap, recorded (`9dc03da39`).** `requestFulfillment` answers `accepted` or `rejected` at the holder's discretion and the suite has no injection point to force either, so the two arm-shaped rules each do real work on one arm and record only the arm check on the other. Running the suite once therefore does **not** prove both arms conform — an executor that only ever accepts is never asked whether its rejections declare `blocking`, the loop-terminator field. The mitigation already shipped (`ConformingRejectingExecutor` beside `ConformingExecutor`); the docblock now says why it is there so the limit is not rediscovered.

## Gates

Node v22.22.1, sequential (never unit concurrent with integration).

| Gate | Exit |
|---|---|
| `pnpm -r build` (cold — `dist/` **and** `.tsbuildinfo` removed) | **0** |
| `pnpm type-check` | **0** |
| `pnpm lint` (incl. `check:invariants`) | **0** |
| `pnpm test` | **0** |
| `pnpm test:integration` | **1** — see below |

**Integration, re-run at `9dc03da39`** (`Test Suites: 5 failed, 145 passed, 150 total` / `Tests: 10 failed, 1 skipped, 1244 passed, 1255 total`). One failure is the known pre-existing **#2638** (`earliest-order-date`, the February CET/CEST assertion on a non-UTC host). **The other four are PrestaShop container-boot failures — `Port 80 not bound after 180000ms` — not assertions**, one of them `prestashop-harness-smoke`, whose job is to assert the harness works. Applying #2639's own discriminator, the smoke suite fails **identically in isolation**, so this is not momentary contention: the PrestaShop container cannot boot on this host at present. The same suites passed earlier in this session on this same tree (`1 failed, 149 passed`). #2639's suite is among the four but is failing for the boot reason, **not** its S-3 assertion, so it is not claimed as a #2639 sighting. Full evidence in the [re-run comment](https://github.com/openlinker-project/openlinker/pull/2691#issuecomment-5475763982).

**`check:invariants` goes 35 → 36**; `check-contract-suite-not-in-production.mjs` is the increment. Both numbers were derived by counting distinct `scripts/*.{mjs,sh}` occurrences in the script, not quoted. It passes `--self-check` (8 assertions) and its real run (*subpath exported, barrel clean, 3939 production files scanned*).

Kit specs: **79 tests across 5 files**. A cold `pnpm -r build` was run because this adds a package-exports entry.

## `docs/lessons.md`

One clause sharpening the existing stale-artifact entry, from a live incident here: a `/tmp/lint.log` read as this run's lint belonged to a **concurrent sibling worktree**. Worse than the own-stale-run case already recorded, because the file was genuinely fresh, plausible, and about a different tree. The rule now reads: an artefact is evidence only when its mtime belongs to **this run** *and* its path **cannot be reached by a sibling** — two conditions, not one.

## Three environment facts, not defect classes

- **Two concurrent `tsc -b` runs emitting into one `dist/`** — the #2011 shape that `check-workspace-dep-declarations` exists to prevent and that fired for real on #2390. Self-inflicted by launching a second gate script before the first exited.
- **`rm -rf dist` without `.tsbuildinfo`** leaves `tsc -b` convinced the output is current, so it emits nothing and every dependent fails `TS6305` (×171 here). Removing both is the correct cold-build recipe.
- **`setsid` does not exist on macOS**, so a relaunch silently never started and ~20 minutes were spent polling a job that was not running. The most dangerous of the three because it fails quietly — `nohup … & disown` is the portable form.

## Not in scope

- `W3a-17` / `W3a-18` consumption (AC-1) — blocked on #2408 / #2409.
- `runAvailabilityAuthorityContract` — no dispatched adapter exists.
- Adopting the pure-runner split for ksef / subiekt so they can self-test too — worth an issue once this shape has proven itself on one port.

Refs #2404 — deliberately `Refs` rather than `Closes`: the base is a wave branch, so auto-close would not fire anyway, and the orchestrator owns merge and epic-checklist closure.


piotrswierzy:
## Integration re-run after the `/pr-review` finding — corrected result

The gates table in the PR body records the integration run from **before** the two review commits (`3f872f225`, `9dc03da39`). That result is superseded; re-running is the right call and the body is updated to match. Both runs are reported here rather than only the flattering one.

**Re-run at `9dc03da39`:** `INT_EXIT=1`

```
Test Suites: 5 failed, 145 passed, 150 total
Tests:       10 failed, 1 skipped, 1244 passed, 1255 total
```

| Failing suite | Reason | Verdict |
|---|---|---|
| `earliest-order-date.int-spec.ts` (36.9 s) | assertion: expected `…T00:00:00.000Z`, got `…T01:00:00.000Z` | **#2638**, named, not chased |
| `orders/allegro-prestashop-carrier-mapping.int-spec.ts` (289.6 s) | `Port 80 not bound after 180000ms` | container boot, **not** #2639's S-3 assertion |
| `orders/prestashop-harness-smoke.int-spec.ts` (283.8 s) | `Port 80 not bound after 180000ms` | container boot |
| `prestashop/prestashop-order-fulfillment-update.int-spec.ts` (207.5 s) | `Port 80 not bound after 180000ms` | container boot |
| `prestashop/prestashop-webhook-provisioning.int-spec.ts` (228.7 s) | `Port 80 not bound after 180000ms` | container boot |

**The four PrestaShop failures are environmental, established rather than assumed.** Three independent lines of evidence:

1. **The failure is a container-boot timeout, not an assertion** — `Port 80 not bound after 180000ms`. Nothing about the code under test was ever exercised.
2. **One of them is `prestashop-harness-smoke`**, whose entire job is to assert the harness works. When the smoke test of the harness fails, the harness is the subject, not the code.
3. **The same suites passed earlier in this session on this same tree** — the first full run gave `1 failed, 149 passed`, with only #2638 red. Six agents share this Docker daemon and a long session degrades it, which is documented standing behaviour for this suite family.

I applied **#2639's own discriminator** (does it fail in isolation?) rather than asserting contention: `prestashop-harness-smoke` re-run alone fails **identically** with `Port 80 not bound after 180000ms`. So this is not momentary contention — the PrestaShop container cannot boot on this host at present. Either way it is not this change, which adds test-only files under `libs/core/src/fulfillment/testing/`, one package-exports entry and docs; none of that can influence whether a Docker container binds a port. I did **not** restart the Docker daemon, because sibling agents are mid-run against it.

**On #2639 specifically:** its suite is in the failing list, but it is failing for the *boot* reason above, not for the S-3 assertion #2639 documents. I am not claiming it as a #2639 sighting — that issue's text explicitly warns against conflating a ~300 s duration in a full run with contention, and the discriminator it names points the other way here.

**Unchanged and re-verified after the review commits:** cold `pnpm -r build` **0**, `pnpm type-check` **0**, `pnpm lint` **0** (incl. `check:invariants` at **36**), kit specs **79 passed / 5 suites**.

Issue #2599, opened 2026-08-27
[TASK] CORE/Orders — carry the buyer tax id on the order contract


norbert-kulus-blockydevs:
## Problem / Context

The neutral order contract carries no buyer tax identifier. Grep across `libs/core/src/orders/domain/types/**` and `libs/core/src/customers/**` returns nothing for `taxId` / `vatNumber` / `buyerTaxId`.

The PrestaShop order-source adapter **already reads the company field** off the address (`prestashop-order-source.adapter.ts:359`), so the company name travels today and only the tax id is missing. Without it a B2B invoice cannot be issued, which blocks the whole invoicing and KSeF branch for orders from this channel. It is also the prerequisite the sales-document rule engine is waiting on: `SalesDocumentOrderFacts.buyerHasTaxId` is permanently `undefined` because no order can supply it.

## Proposed Solution

Add the field to the order contract, persist it on `order_records`, and populate it in the PrestaShop adapter from `ps_address.vat_number`. Absent must remain distinguishable from "known to be empty" — never default to `false` or `''`.

## Classification

**Type**: CORE / Integration
**Layer**: Domain / Infrastructure / Database
**File(s)**: `libs/core/src/orders/domain/**`, `libs/core/src/orders/infrastructure/persistence/**`, `apps/api/src/migrations/**`, `libs/integrations/prestashop/src/infrastructure/adapters/prestashop-order-source.adapter.ts`

## Dependencies

- Requires a migration; see `docs/migrations.md`

## Acceptance Criteria

- [ ] Order contract carries the buyer tax id; absent and empty are distinguishable
- [ ] Migration generated and `migration:show` clean
- [ ] PrestaShop adapter populates it from the invoice address
- [ ] `toSalesDocumentOrderFacts` passes a real value instead of `undefined` when the order has one
- [ ] Tests added or updated for non-trivial logic

norbert-kulus-blockydevs:
## /pr-review - #2599 (buyer tax id on the order contract)

Reviewed `70fd020ab` on `epic/2590-prestashop-no-module` (PR #2627), read-only.

**Verdict: request changes.** The three-state encoding itself is clean and I could not find a reader that misreads it. Two things block or need a decision: one import re-opens a documented module cycle, and the "no routing rule acts on it" claim is not true for an operator who installed the PL starter template.

### BLOCKING

[...]

PR #2584, opened 2026-08-26
feat(orders,sales-documents): batched per-order sales-document read (#2516)

norbert-kulus-blockydevs:
Builds the batched read behind the ADR-065 per-order projection. The orders list renders a document fact per row, and assembling it per row would issue two reads per order.

## What this adds

`SalesDocumentViewService` (`libs/core/src/orders/application/services/`) returns a `Map<orderId, SalesDocumentView>` for a list of order ids. It composes:

- the order record's persisted `salesDocumentBlockReason` / `salesDocumentUnresolvedReason` / `salesDocumentBlockDetail`, copied **verbatim**;
- every invoice record for those orders, on both invoice axes;
- every fiscal-registration record for those orders, on its single axis (the receipt member carries no regulatory field, so an authority answer for a receipt is unrepresentable);
- the routing decision, for an order that has no document yet.

Query count is fixed: three reads in parallel (order records, invoice records, fiscal records), then - only for orders that have no document yet - the candidate connections plus one batched routing resolve (three reads). Seven for a page of 1 and seven for a page of 200; a page whose orders all already have a document pays three.

## Where the prospective document kind comes from

An order with a record reports that record's kind. An order without one reports the kind routing resolves, using the **same** two-step precedence the auto-issue gate uses. To keep that from becoming a second implementation of routing, the precedence moved into `sales-documents` as the pure `chooseSalesDocumentDecision`, and `AutoIssueTriggerService` now calls it instead of holding its own copy. Behaviour is unchanged: the zero-eligible-candidate short-circuit and the `documentKind !== null` predicate are carried over byte for byte.

`resolveSalesDocumentRouting` gains an order-free sibling it delegates to. It has never read the order (ADR-041 decision 2 keeps the parameter for a future mode), and a read-side caller holds an `OrderRecord` snapshot rather than a clean `Order` - fabricating one to satisfy an ignored parameter would be a cast dressed up as a value.

## New batched reads

- `SalesDocumentRuleRepositoryPort.findByCountries` / `SalesDocumentCountryDefaultRepositoryPort.findByCountries`, feeding `ISalesDocumentRulesService.resolveRoutingBatch`.
- `InvoiceRecordRepositoryPort.findAllByOrderIds` -> `IInvoiceService.listInvoicesForOrders`.
- `FiscalRegistrationRecordRepositoryPort.findAllByOrderIds` -> `IFiscalRegistrationService.getByOrderIds`.

Every one is a projection read. Nothing in this PR writes.

## New context edge

[...]

PR #2578, opened 2026-08-26
feat(sales-documents): neutral SalesDocumentView projection (#2515)

norbert-kulus-blockydevs:
## What changed

Adds the one neutral per-order read the `/orders` row, the order-detail panel and the sales-document settings page all share (ADR-065, mini-epic #2514).

`SalesDocumentView` carries the resolved kind, the state on the axis belonging to that kind, the identity fields a surface renders, the persisted block and unresolved reasons verbatim, and any record held on another connection.

Two rules from ADR-065 are structural here rather than documented:

- **No third vocabulary.** The states are `InvoiceStatus`, `RegulatoryStatus` and `FiscalRegistrationStatus`, borrowed by name. A flattened enum was rejected because it cannot express *issued, then rejected by the authority*.
- **A fiscal receipt has no authority axis.** The two members are a discriminated union, so the missing axis is unrepresentable rather than merely unset. No surface can render a confirmation a fiscal registration never produces (ADR-042).

## The one judgment call worth reviewing

`sales-documents` is a zero-outbound-core-context-edge leaf, enforced by `barrel-purity.spec.ts`, which previously authorized exactly one type-only specifier. Naming the existing vocabularies needed two more.

This PR follows the existing `@openlinker/core/orders/types` precedent: two new cycle-breaker sub-barrels, `@openlinker/core/invoicing/types` and `@openlinker/core/fiscalization/types`, that re-export **types only** and no runtime value. An `import type` erases, so no `require()` is emitted and the CJS cycle that spec exists to forbid stays impossible. The spec's authorized list grows from one entry to three, with the rationale recorded in place.

The alternatives were re-declaring the three unions inside `sales-documents` with a drift guard, which is the parallel vocabulary ADR-065 rules out, or typing the axes as bare `string`, which throws away the type safety that is the point.

## Tests

- `sales-document-view.types.spec.ts`: an authority rejection is representable and distinguishable from a failed issuance; a receipt rejects a regulatory field at compile time; the union narrows on `kind` alone; the undecided, decided-but-unissued, and duplicate-elsewhere states each round-trip.
- `barrel-purity.spec.ts` still passes with the widened list.

Gate run: `pnpm --filter @openlinker/core type-check`, `eslint` on every changed path, `pnpm check:invariants` (cross-context imports clean), and the two jest suites above. The full `pnpm test` was not run locally.

Closes #2515

Issue #2518, opened 2026-08-26
[TASK] CORE + Interface — market discovery: billing countries with order counts


norbert-kulus-blockydevs:
## Problem / Context

A clean instance has no routing, and nothing tells the operator which markets they need. OpenLinker already stores the billing country on every order, so the markets that need configuring are derivable — but no read exposes them, so an unconfigured market stays invisible until someone notices no documents were created.

## Proposed Solution

Add a read returning distinct billing countries with an order count over a window, and expose it so the settings page can list detected markets alongside configured ones. Follow the ADR from #2503 for the window and for what "detected" means. This is a read only: it configures nothing and never creates routing on its own.

## Classification

**Type**: CORE + Interface
**Layer**: Application / Interface
**File(s)**: `libs/core/src/orders/`, `apps/api/src/`

## Dependencies

- Parent mini-epic: #2514
- #2503

## Acceptance Criteria

- [ ] The read returns country plus count over the window defined in #2503.
- [ ] A country with configured routing and one without are both returned; classification is the caller's job.
- [ ] The read is a single query, not one per country.
- [ ] It never writes: no routing, rule or default is created as a side effect.
- [ ] Tests added or updated for non-trivial logic
- [ ] No architecture boundary violations (CORE ↔ Integration)

## Branching and merge target

Branch: `{issue#}-{short-kebab}`, created **from its mini-epic branch**.
PR target: **its mini-epic branch**.

## How this issue gets finished

1. Implement on the branch above.
2. Open the PR into its mini-epic branch.
3. Run `/pr-review <PR>` and post the verdict as a PR comment.
4. Fix **every** finding, and reply under the comment saying what changed for each one.
5. Run `/pr-review` again. Repeat until the review comes back approved with no findings.
6. Only then merge.

Quality gate before every commit: `pnpm lint`, `pnpm type-check`, `pnpm test`.
Commits are signed off (`git commit -s`).

norbert-kulus-blockydevs:
Merged into the mini-epic branch. Closing manually: the PR targeted a non-default branch, so `Closes #N` does not fire until the epic reaches `main`.

PR #2692, opened 2026-08-31
feat(orders,sales-documents): M1 - one per-order sales-document read (#2514)

### LINKED ISSUE #2514: [EPIC] M1 — Backend: one per-order sales-document read
> Parent: #2513
> Mockup: `docs/plans/mockups/sales-document-routing.html`

## Scope

Every surface in this epic needs the same facts about one order: which document kind routing chose, the state on the axis that belongs to that kind, why it is held if it is, and whether a second record exists elsewhere. Today those come from separate invoice and fiscal reads with no shared shape, and the reason is not on the orders-list payload at all. This mini-epic builds the one read the three surfaces share, plus the market-discovery read the settings page needs.

## Branching

Mini-epic branch `phase/sdux-m1-read-model`, created **from `epic/sales-documents-ux`**, opening a PR **into `epic/sales-documents-ux`**.
Each task below gets its own branch off `phase/sdux-m1-read-model` and a PR **into `phase/sdux-m1-read-model`**.

## Dependencies

- #2501 (ADR: projection and vocabulary) merged to `main`.
- #2503 (ADR: market discovery) merged to `main`.

- [ ] One neutral projection type serves the list, the detail panel and the settings page.
- [ ] The list read is batched: rendering N rows performs one query, not N.
- [ ] The persisted block and unresolved reasons are returned, never re-derived by a caller.
- [ ] A second record on another connection is reported, not hidden.
- [ ] Market discovery returns distinct billing countries with order counts over a window.
- [ ] Every task branch merged into `phase/sdux-m1-read-model`, and `phase/sdux-m1-read-model` merged into `epic/sales-documents-ux` via reviewed PR.

## Status update (2026-08-31)

norbert-kulus-blockydevs:
Mini-epic **M1 - Backend: one per-order sales-document read** (#2514), under epic #2513. Four task PRs merged into this branch: #2578, #2584, #2587, #2689.

## What this delivers

One neutral projection that the orders list, the order-detail panel and the settings page all read, so the three surfaces cannot disagree about the same order.

`GET /orders/:internalOrderId/sales-document` returns `SalesDocumentViewResponseDto`, and the **identical object** is on every `GET /orders` row and on `GET /orders/:id` as `salesDocument`. One DTO, one mapper.

```
orderId          string
documentKind     string | null      open-world; null = routing has not decided
document         Invoice | Receipt | null
blockReason      SalesDocumentGateBlockReason | null    persisted, verbatim
unresolvedReason SalesDocumentUnresolvedReason | null   persisted, verbatim
blockDetail      string | null                          persisted, verbatim
otherRecords     [{ recordId, connectionId, kind, blocksFurtherIssuance }]
```

`document` is discriminated on `kind`. The invoice member carries both axes; the fiscal-receipt member has **no regulatory field at all**, so a clearance answer for a receipt is unrepresentable rather than merely unset (ADR-065, ADR-042).

The list read is batched, so a page of orders costs a fixed number of queries regardless of page size, following the `getEarliestOrderDateByConnection` precedent. A spec asserts the count does not grow with the number of ids.

## Three traps for the frontend mini-epics

- `documentKind` is open-world, so a `switch` needs a default arm. `document.kind` is closed and exhaustive.
- `inFlightUntil` must be compared against now. A lease in the past is a crashed attempt, not a live one.
- `artefactCount: 0` on a `registered` receipt is a success, not a failure.

## Decisions open to dispute

**Delivery country, not billing, for market discovery (#2518).** The issue title and ADR-066's prose both say billing. This reads `shippingAddress.country`, because ADR-066's decisive rule is that discovery must use the same field the evaluator sees, and `toSalesDocumentOrderFacts` reads the delivery country with no billing fallback. Billing would list markets whose configuration changes nothing. A spec pins the field so the two move together.

**A prospective `documentKind` for orders with no document.** Returning `null` there would make the panel say "routing has not decided" for every not-yet-issued order. The projection runs the gate's own precedence instead, and to avoid two implementations of routing that precedence moved into `sales-documents` as the pure `chooseSalesDocumentDecision`, which `AutoIssueTriggerService` now calls.

**`@Roles('admin')` on the discovery endpoint**, stricter than a plain authenticated read. Its only consumer is the admin-only settings page and its sibling controller is class-level admin.

**A 30-day discovery window.** ADR-066 says the window is a documented constant rather than a per-operator setting but gives no number. 30 comes from the committed mockup, whose detected-market fixture carries `days: 30` on every row. `windowDays` and `since` are returned on the response so no frontend holds its own copy.

**A new `orders -> fiscalization` module edge**, recorded in the dependency map.

## Collisions this branch absorbed

**A real behavioural one.** #2599 put the buyer tax id on the order contract and made the gate's mapper read it, while the projection still passed `buyerHasTaxId: undefined` - so a rule keying on that fact would have matched in the gate and not in the read. Fixed, with a spec over all three states.

**An invariant one.** #2599 added two value re-exports to the `orders/types` cycle-breaker sub-barrel, which #2515's new guard flagged. Allow-listed with the reason rather than weakening the guard.

The final commit here resolves the catch-up merge with M2: `check:invariants` now runs both new guard scripts, and `IFiscalRegistrationService` keeps both `getByOrderIds` (this mini-epic) and `getInFlightRegistration` (M2), since neither supersedes the other.

## Not included

No integration tests (Docker-backed, out of reach on the dev machine), no frontend mirrors (no new unions), no pagination on discovery, no `?days=` parameter.

## Tests

Per-task suites green, plus `type-check` on core and api and `pnpm check:invariants` after the catch-up merge. The full `pnpm test` was not run locally; CI on the epic PR into `main` is the first full-suite signal.

Closes #2514


norbert-kulus-blockydevs:
## Review verdict: approve, clean to merge into the epic branch

Gate re-run in the worktree: `pnpm type-check` clean across every package (including `apps/worker`), `pnpm check:invariants` exit 0, eslint clean on the changed core and api paths, core suites 425 passed across 20 files, api `orders.controller` + sales-documents suites 50 passed.

### Claims in the body, verified

1. **Delivery country, not billing.** `libs/core/src/orders/application/services/sales-document-view.service.ts:271` reads `orderSnapshot.shippingAddress.country`, and `order-record.repository.ts:788` groups discovery on the same `'{shippingAddress,country}'` path. That matches `toSalesDocumentOrderFacts` at `libs/core/src/invoicing/application/mappers/order-to-sales-document-order-facts.mapper.ts:48`, which reads `order.shippingAddress?.country` with no billing fallback. The decision to contradict the issue title is the right one: discovery that named a field the evaluator never sees would list markets whose configuration changes nothing.

2. **The spec that pins it is a string assertion, not a link between the two functions.** `order-record.repository.spec.ts:1537` asserts the SQL path contains `'{shippingAddress,country}'`, and `sales-document-view.service.spec.ts:249` asserts the facts shape. Neither fails if `toSalesDocumentOrderFacts` moves to billing. This is weaker than the body's "a spec pins the field so the two move together" - it pins each side to a literal, not to each other. Non-blocking, and no mirror script is warranted for two functions in one package, but the body overstates it.

3. **Single-implementation precedence: confirmed.** `chooseSalesDocumentDecision` (`libs/core/src/sales-documents/domain/domain-services/choose-sales-document-decision.ts:74`) is the only place the two-step precedence exists, and `AutoIssueTriggerService.resolveSalesDocumentDecision` now calls it with its own `eligibleCount` short-circuit deleted. The carried-over `countRoutableCandidates` predicate at :67 is byte-identical to the gate's, so the refactor is behaviour-preserving.

4. **Batching: confirmed and asserted.** `sales-document-view.service.spec.ts:326` runs 50 ids and asserts one call each to `findByIds`, `listInvoicesForOrders`, `getByOrderIds`, `connections.list`, `resolveRoutingBatch`. `orders.controller.ts:173` calls `getForOrders` once per page, and the routing half is skipped entirely for a page whose orders all have documents.

5. **Cross-context discipline is respected.** Every outbound read goes through `IInvoiceService`, `IFiscalRegistrationService`, `ISalesDocumentRulesService`, `IOrderRecordService` or `ConnectionPort`; no `*RepositoryPort` crosses a boundary, and `sales-documents` gains only relative-import files.

### Findings

1. `sales-document-view.service.ts:137-139` - `otherRecords` filters on `other.connectionId !== winner.connectionId`, so a second record on the SAME connection is always treated as that document's history. That is right for two invoices, but it also hides a fiscal-receipt record sitting beside an invoice record on one connection, which is a genuine second originating document and exactly what `otherRecords` exists to surface. The filter is a connection test where the rule wants a connection-or-kind test. Reachable only on a connection carrying both capabilities, so low severity, but the spec at :308 pins the current behaviour under a name ("that document history") that is wider than what it tests.

2. `sales-document-view.service.ts:279` sets `taxTreatment` only when the denormalized column is a valid value, while `toSalesDocumentOrderFacts` passes `order.totals.taxTreatment` straight through. The two therefore disagree for an order whose stored `taxTreatment` is null but whose live snapshot carries one, which is precisely the `net-priced-order` rule arm. Worth a spec or a shared reader, given how much of this mini-epic is built on "the projection and the gate must answer identically".

3. Non-blocking: `package.json`'s `check:invariants` chain is edited here and again on the M3 branch (`check-fiscal-registration-progress-mirror`). Whichever of #2692 / #2698 merges second must keep both entries; a conflict resolution that keeps one side silently drops a guard.

No epic-constraint violation found. Nothing on this branch asserts a fact the backend cannot report: the receipt member carries no regulatory field, the persisted reasons are copied verbatim, and the market-discovery endpoint description states it writes nothing.

PR #2585, opened 2026-08-26
feat(sales-documents): one reason-to-copy map, guarded against drift

norbert-kulus-blockydevs:
## What this builds

One reason-to-copy map for both persisted sales-document vocabularies, in `apps/web/src/features/sales-documents/lib/sales-document-reason-copy.ts`:

- `SALES_DOCUMENT_UNRESOLVED_REASON_COPY` - 7 routing reasons, a short label plus a long form.
- `SALES_DOCUMENT_GATE_REASON_COPY` - 6 gate reasons, the same plus a `tone` and `keepsAction`.
- `resolveSalesDocumentReasonCopy(reason, unresolvedReason)` - reads only the two persisted values, prefers the paired routing reason for the `unresolved-routing` bridge value, and returns `null` both for an unblocked order and for a value this build does not recognise.

`invoicingBlockedBadge` (`features/orders/lib/order-row.ts`) now derives its label, tone, hint and action from that map, replacing its own `BADGE_BY_REASON` table. Before this, only `ambiguous-connection-no-primary` had words of its own and the other six routing reasons collapsed into a single generic label that told the operator nothing about their own configuration.

## The mirror and its guard

The reason unions stay hand-mirrored in `features/orders/api/orders.types.ts`, where the order payload carries them. This PR adds no new mirrored array; it adds the copy that must cover one.

`scripts/check-sales-document-reason-mirror.mjs` gains a second check on top of its existing array comparison: the top-level keys of both copy maps must cover the backend unions exactly, in either direction. A reason added in `libs/core/src/sales-documents` now fails `pnpm check:invariants` until copy exists, and a copy entry for a reason that no longer exists fails too. The key parser is brace-matched and string-aware rather than regex-scanned, so a brace or an apostrophe inside operator copy cannot end the scan early and a nested key cannot be read as a reason id. Both are covered by new `--self-check` cases, alongside the script's existing ones.

## Where the copy comes from, and three places it departs from the mockup

The UX mockup (`docs/plans/mockups/sales-document-routing.html`, reachable only on `origin/epic/docs-sales-documents`) and the operator state reference, cross-checked against the backend writers. Three entries deliberately say something different, each because the backend cannot support what the mockup says:

[...]

Issue #2513, opened 2026-08-26
[EPIC] Sales documents — routing-first UX across settings, orders and order detail


norbert-kulus-blockydevs:
> Mockup: `docs/plans/mockups/sales-document-routing.html` — committed to the repo. Every state is hash-addressable, e.g. `#page=settings`, `#page=orders`, `#page=order&state=receipt-in-doubt`.
> Decisions epic: #2499. **Its phase 1 (#2500) must be merged to `main` before this epic's branch is cut.**

## Problem / Context

Three surfaces decide, or fail to decide, which document a sale gets, and today they disagree with each other.

- **`/settings/sales-documents`** runs two configuration systems on one page, with the precedence between them written only in a code comment. It renders ADR references and API vocabulary as body copy, offers to adopt a starter template it has already adopted, and its brightest control is disabled. Nothing on it states what an order actually gets — and on the live demo the answer for Poland is **nothing at all**, because all three rules test a buyer tax ID that is never recorded and both defaults are set, which disables the default step. The page presents that as working.
- **The `/orders` row** carries payment state, document clearance and an action in one column headed "Payment", and the action is a hardcoded `+ Issue invoice` — offered on orders routed to a fiscal receipt.
- **The order-detail panel** shows two orange primary buttons (`Issue invoice` and `Register receipt`) and a document-kind dropdown, asking the operator to re-decide what routing already decided. It has no feedback at all while a provider is being called, and says `Still not found` during normal processing.

A colleague testing the flow reported clicking the wrong thing on each of her first three attempts. That is a broken affordance, not a preference.

## Proposed Solution

One principle, applied to all three surfaces:

> **Routing decides the document kind. Every surface states that decision and offers only that document. Where routing has not decided, the surface says so and points at the settings — it never falls back to asking.**

Ten mini-epics: four backend (the read model, fiscal-lifecycle honesty, asynchronous registration, the configuration surface), one shared frontend foundation, four frontend surfaces, and an end-to-end pass verified against the committed mockup.

Two capabilities the mockup adds beyond fixing the above, both operator-requested:

- **Market discovery** — a clean instance has no routing, so unconfigured countries that orders arrive from are listed with their order counts instead of being invisible.
- **Suggested setups** — offered only where researched guidance exists (Poland alone today), stated as such.

## Classification

**Type**: CORE + Frontend (cross-cutting; children are split BE/FE per the `/create-issue` convention)
[...]

PR #2694, opened 2026-08-31
feat(web): M5 - sales-document vocabulary, copy and primitives (#2533)

### LINKED ISSUE #2533: [EPIC] M5 — Frontend foundation: vocabulary, copy and primitives
> Parent: #2513
> Mockup: `docs/plans/mockups/sales-document-routing.html`

## Scope

The shared pieces all three surfaces use, built once. A reason-to-copy map guarded against drift, the glyph and headline treatment that makes the list and the panel read alike, the lifecycle trail for an invoice, a popover that survives a scroll container, and the skeleton height fix that closes a tracked defect.

## Branching

Mini-epic branch `phase/sdux-m5-fe-foundation`, created **from `epic/sales-documents-ux`**, opening a PR **into `epic/sales-documents-ux`**.
Each task below gets its own branch off `phase/sdux-m5-fe-foundation` and a PR **into `phase/sdux-m5-fe-foundation`**.

## Dependencies

- #2501 (ADR: vocabulary) merged to `main`.

- [ ] One reason-to-copy map, mirror-guarded, used by every surface.
- [ ] The kind glyph and headline are primitives, not per-page markup.
- [ ] A cell popover is not clipped by the table's scroll container.
- [ ] `DataTableSkeleton` no longer changes row height on load.
- [ ] Every task branch merged into `phase/sdux-m5-fe-foundation`, and `phase/sdux-m5-fe-foundation` merged into `epic/sales-documents-ux` via reviewed PR.

## Status update (2026-08-31)

norbert-kulus-blockydevs:
Mini-epic **M5 - Frontend foundation** (#2533), under epic #2513. Five task PRs merged into this branch: #2585, #2686, #2687, #2688, #2693.

## What M6 to M9 build against

**Vocabulary** in `apps/web/src/features/sales-documents/lib/sales-document-reason-copy.ts`, exported from the feature barrel. One entry per reason with `short`, `detail`, and on gate reasons `tone` and `keepsAction`. `resolveSalesDocumentReasonCopy(reason, unresolvedReason)` is the single entry point: it reads only the two persisted values, prefers the paired routing reason for the `unresolved-routing` bridge, and returns `null` both for an unblocked order and for a value this build does not recognise. `invoicingBlockedBadge` now derives from it, so the orders row and any new surface cannot disagree.

**Primitives** in `shared/ui`, all in the dev-ui gallery:

- `DocumentKindGlyph` carries no tone and names itself when standalone, so colour is never the only signal.
- `DocumentHeadline` takes the state word rather than deriving it.
- `DocumentLifecycle` returns `null` for anything but an invoice. That is where "a fiscal receipt has no authority axis" is now enforced rather than trusted.
- `Popover` gains opt-in `dismissOnViewportChange` and portals out of `.data-table__container`, which is asserted.
- `DataTableSkeleton` gains `rowAction` and `label`; `DataTableColumn` gains `lines`.

## Guard

No new mirrored array. `scripts/check-sales-document-reason-mirror.mjs` gained a second check instead: the copy maps' top-level keys must cover the backend unions exactly, in both directions, so a reason added in core fails `pnpm lint` until copy exists. Brace-matched, string-aware key parser with `--self-check` cases.

## Decisions open to dispute

**`trigger-model-batched` copy contradicts the mockup and the operator docs, deliberately.** Both say the document "goes out on the next run". `AutoIssueTriggerService` raises `BatchedTriggerNotImplementedError`, so nothing will ever collect it. The copy says what is true and `keepIssueAction` is flipped to true, since issuing by hand is that order's only route. `docs/sales-documents-state-reference.md` on `origin/2511-docs-state-reference` needs the same correction before it merges, and that is tracked separately.

**`ambiguous-connection-no-primary` has two writers** with different shapes (two country defaults, or several providers with no primary), so its label is "Two setups apply" rather than the mockup's "Two defaults set".

**Four reason labels exceed the style guide's ~17 character pill budget.** They are the mockup's wording, where the label sits in a popover. They still render in the transitional `/orders` pill that M8 replaces.

**`DocumentKind` is mirrored by hand in `shared/ui` with no guard script.** A two-value closed union with no wire contract of its own did not seem to earn a fourth parser.

## Not included

`features/invoicing/lib/sales-document-block-copy.ts` keeps its kind-flavoured long-form panel copy: collapsing it belongs with the M9 panel rebuild, and importing the sales-documents barrel from `features/invoicing` would close a runtime import cycle today. No frontend attention set, because the count is backend-computed and a second partition here would drift unguarded. No product consumer for the three new primitives; M8 and M9 are their callers.

**One thing M9 needs to know.** `InvoiceRecord` persists `issuedAt`, `createdAt` and `updatedAt` only, with no submitted-at and no authority-answered-at. So a real lifecycle trail today has a step with no time of its own, and `DocumentLifecycle` renders that as an explicit absent marker rather than borrowing a neighbour's timestamp.

## Tests

Per-task suites green, plus web type-check and `pnpm check:invariants` on the merged branch. The full `pnpm test` was not run locally; CI on the epic PR into `main` is the first full-suite signal.

Closes #2533


norbert-kulus-blockydevs:
## Review verdict: approve, clean to merge into the epic branch

Gate re-run in the worktree: `pnpm --filter @openlinker/web type-check` clean, `pnpm check:invariants` exit 0, eslint clean on the changed shared/ui and feature paths, 42 web test files / 475 tests passed across `shared/ui`, `features/sales-documents` and `features/orders`.

### Claims in the body, verified

1. **`DocumentLifecycle` genuinely cannot render an authority axis for a receipt.** `apps/web/src/shared/ui/document-lifecycle.tsx:75` short-circuits on `kind !== 'invoice'` before any step is read, so it is structural rather than a caller convention. Note for M9: it refuses the WHOLE trail for a receipt, not just the second stage, so a receipt's own progress needs a different primitive. That is the correct trade, it is just not stated in the body.

2. **The reason-copy guard fails in both directions - verified by breaking it.** In a scratch copy:
   - Added `'zzz-fake-reason'` to `SalesDocumentGateBlockReasonValues` in core plus the `orders.types.ts` mirror. The script exits 1 with `no copy entry for: 'zzz-fake-reason'` against `sales-document-reason-copy.ts:125`.
   - Added a `'zzz-orphan-copy'` entry to `SALES_DOCUMENT_GATE_REASON_COPY` only. The script exits 1 with `copy entry for a reason that does not exist: 'zzz-orphan-copy'`.
   Both reverted; `git status` clean. The `--self-check` arm passes too. This is a real guard, not a decorative one.

3. **`trigger-model-manual` is neutral and unaggregated.** `sales-document-reason-copy.ts:180` gives it `tone: 'neutral'`, `keepsAction: true`, and no frontend attention partition is introduced - the count stays the backend's `SalesDocumentAttentionReasonValues`. Correct per ADR-041 section 54.

4. **`trigger-model-batched` copy is right to contradict the mockup.** `AutoIssueTriggerService` raises `BatchedTriggerNotImplementedError`, so "goes out on the next run" would be a false promise. The copy states the batched path is unavailable and keeps the manual action. This is the single best decision in the mini-epic and it is the pattern the other two PRs should be held to. The tracked correction to `docs/sales-documents-state-reference.md` must actually land, or the docs and the product will disagree in the direction that matters.

5. Copy audited for over-claiming across both maps: no time estimate, no progress claim, no "safe to leave", no delivery or authority assertion. `tax-rate-conflict` is written so that a backend which starts writing it cannot make the surface claim a hold that does not exist - that is the right defensive shape for a declared-but-unwritten value.

### Findings

1. `apps/web/src/shared/ui/document-kind-glyph.tsx:32` - `DocumentKind` is hand-mirrored with no guard. **Acceptable.** `SalesDocumentKind` is a closed two-value union that ADR-041 states will never gain a third member (a document property belongs on the outcome, not as a new kind), and the mirror lives in `shared/ui`, which by the layering rule cannot import a feature or core type anyway. A fourth parser for two string literals would cost more than it protects. Worth one line in the epic's docs pass so the next reader does not treat the omission as an oversight.

2. `apps/web/src/features/orders/lib/order-row.ts:26` imports a VALUE (`resolveSalesDocumentReasonCopy`) from `../../sales-documents`. Feature-to-feature imports have precedent in this tree, but every other one I found is `import type`. This is the first value edge between two features, and it is the same class of coupling the body says it avoided for `features/invoicing` on cycle grounds. It works today because `sales-documents` does not import `orders`; that is an unguarded invariant. A short comment at the import naming it would be enough.

3. Body point worth keeping visible for M8: the four over-length reason labels still render in the transitional `/orders` pill. If M8 slips, those pills ship truncated.

No epic-constraint violation found.

PR #2698, opened 2026-08-31
feat(fiscalization): M3 - asynchronous manual registration (#2524)

### LINKED ISSUE #2524: [EPIC] M3 — Backend: asynchronous manual registration
> Parent: #2513
> Mockup: `docs/plans/mockups/sales-document-routing.html` — states `receipt-registering`

## Scope

`POST /fiscal-registrations` calls the blocking adapter inline. eparagony polls its device to a terminal state inside that call, so the request can sit for a long time, and if the operator closes the tab the request is cut off. The record survives; the operator's view of it does not. Until this changes, no surface may say the work continues after navigation — which is why the mockup deliberately says the opposite.

## Branching

Mini-epic branch `phase/sdux-m3-async-registration`, created **from `epic/sales-documents-ux`**, opening a PR **into `epic/sales-documents-ux`**.
Each task below gets its own branch off `phase/sdux-m3-async-registration` and a PR **into `phase/sdux-m3-async-registration`**.

## Dependencies

- #2502 (ADR-042 addendum) merged to `main`.
- M2 (#2519) for the in-flight signal.

- [ ] A manual registration is enqueued rather than performed inline.
- [ ] The panel can poll the record and show the result without the original request being open.
- [ ] Only after this may any surface state that the work survives leaving the page.
- [ ] Exactly-once registration is unchanged; a retry cannot produce a second registration.
- [ ] Every task branch merged into `phase/sdux-m3-async-registration`, and `phase/sdux-m3-async-registration` merged into `epic/sales-documents-ux` via reviewed PR.

## Status update (2026-08-31)

norbert-kulus-blockydevs:
Mini-epic **M3 - Backend: asynchronous manual registration** (#2524), under epic #2513, closing ADR-042 decision 4. Three task PRs merged into this branch: #2695, #2696, #2697.

## The contract change M4 and M8/M9 must know

**`POST /fiscal-registrations` is now `202` and returns an acceptance, not a record**: `{ orderId, connectionId, idempotencyKey, jobId, redrivenFromDead }`. Nothing in it reports an outcome, so anything reading `status` off that response is broken.

**The poll target is `GET /orders/:orderId/fiscal-registration?connectionId=<uuid>`**, `connectionId` required. It answers `{ progress, record, inFlight }`, where `progress` is `not-requested | queued | running | stalled | registered | rejected | in-doubt`, mirrored in the frontend and guarded by `scripts/check-fiscal-registration-progress-mirror.mjs`. A new value means editing both sides plus a rendering branch.

Three properties a surface must not get wrong:

- **`record` is `null` while `queued`.** That is the normal state in the window this read exists for, not an error, which is why the panel's receipt slot opens on outstanding work rather than on a record.
- **`stalled` is not a failure.** Intent recorded, nothing running; asking again re-drives it. It must not render as a wait and must not render as a rejection.
- **The key format is one function**, `fiscalRegistrationIdempotencyKey` in `@openlinker/core/fiscalization`. Do not restate it anywhere.

New service surface for M4: `requestRegistration`, `assertRegistrable`, `getRegistrationProgress` on `IFiscalRegistrationService`, plus `findJobByIdempotencyKey` on `ISyncJobsService`.

## Decisions open to dispute

**The order-level guard also runs as a read at enqueue time.** `assertRegistrable` delegates to the same private predicate the write path uses, so there is one rule rather than two. Without it a double attempt is answered `202` and then fails inside a job with nothing on screen. It is advisory by construction and says so: the authoritative guard remains the one inside the per-order lock.

**A same-key record is exempt from that guard when resumable** (`pending`, `registering`, terminally `rejected`). `registered` and every other failure, `in-doubt` included, fall through and are refused. Without the exemption the dead-job re-drive is unreachable, because a `pending` row blocks.

**A `dead` job is re-driven by the request.** `schedule` returns the dead row, so an order would otherwise be permanently un-registrable from the UI while the endpoint kept answering success. The re-drive re-runs the same key against the same record, which the write path treats as a resume, never a resend, so exactly-once is intact.

**`connectionId` is required on the poll** rather than optional, because optional would make `not-requested` sometimes false, which is exactly the class of claim this epic exists to remove.

**`stalled` is a seventh value** rather than folded into `queued`.

**The M1 projection is not consumed**, since it is not on this branch. This reuses the existing registration projection and M2's `SalesDocumentInFlight` rather than inventing a third shape. `IFiscalRegistrationService` and `libs/core/src/sales-documents/` were not restructured, so #2692 should merge additively.

## Not included

**The elapsed reading from `inFlight.since` is not rendered.** The value is plumbed through and typed with its lower-bound caveat, but wording it belongs with the M9 panel rather than being guessed here.

**The auto-issue gate keeps its own `schedule` call.** It shares the key and the payload mapper now, which is where drift was possible; routing it through `requestRegistration` would pull the advisory guard and the dead-job re-drive into a path that wants neither.

No batched progress read, since the orders list has no use for one. The integration test is written but unrun (Docker); it asserts one job row per sale and that the payload key equals the job row key.

**One thing for review**: `check-fiscal-registration-progress-mirror.mjs` restates its sibling's parsers rather than importing them, because these scripts run their check on import and importing would report the sibling's drift under this script's name.

## Invariants

Exactly-once fiscal registration and one originating document per order are intact; no new path crosses the provider boundary, and `in-doubt` is never resolved by a resend.

## Tests

Core, api and web type-check clean, `pnpm check:invariants` exit 0, and the touched suites pass: core fiscalization 170, core sync/fiscalization 356, auto-issue 78, api fiscalization 33, web fiscalization and orders 364. Full `pnpm test` not run locally.

Closes #2524


norbert-kulus-blockydevs:
## Review verdict: request changes. Not clean to merge as it stands.

Two findings block, one of them a broken gate and one a false operator-facing claim. The exactly-once machinery itself is sound and I found nothing wrong with it.

Gate re-run in the worktree: `pnpm check:invariants` exit 0 (the new progress mirror self-checks and reports 7 values in sync), eslint clean, api fiscalization 33 passed, web fiscalization + `sales-document-panel` 5 files / 52 passed, core fiscalization/sync/auto-issue 264 passed. `pnpm type-check` FAILS, see finding 1.

### Blocking

1. **`apps/worker` does not type-check and its suite does not run.**
   `apps/worker/src/sync/handlers/__tests__/fiscalization-register.handler.spec.ts:63` - `error TS2322: ... is not assignable to type 'Mocked<IFiscalRegistrationService>'`. `requestRegistration`, `assertRegistrable` and `getRegistrationProgress` were added to the interface and this mock was not extended. `npx jest --config apps/worker/jest.config.js --testPathPattern fiscalization-register` fails to compile: 1 suite failed, 0 tests run. The body's "Core, api and web type-check clean" is accurate and is exactly the problem - the fourth package was never checked. For reference, #2692 hit the same interface and did update this file, so the two branches will also touch the same line.

2. **The `stalled` alert asserts something the backend cannot know.**
   `apps/web/src/features/orders/components/sales-document-panel.tsx` (stalled `Alert`) says "Nothing was registered with the provider." `resolveFiscalRegistrationProgress` (`libs/core/src/fiscalization/domain/types/fiscal-registration-progress.types.ts:113-116`) returns `stalled` for a record that is `pending` OR `registering` with an expired lease and no live job. The second arm is a crashed attempt that may well have crossed the provider boundary - `resumeExisting` classifies exactly that state as re-attemptable and re-calls the adapter. So for one of the two arms the sentence is a claim OpenLinker is not in a position to make, on a fiscal document, which is the highest-cost place in the product to be wrong. This is the same defect class the epic exists to remove, and the type's own docblock already says the right thing ("intent was recorded and nothing is running"). Suggested fix: drop the sentence, or split the copy on whether the record was ever claimed. The following sentence, "Asking again picks it up where it stopped", is accurate and can stay.

### Verified as correct

- **The dead-job re-drive cannot produce a second provider call.** `requestRegistration` (`fiscal-registration.service.ts:216-266`) enqueues under `fiscalRegistrationIdempotencyKey`, and `requeueDeadByIdempotencyKey` is a single UPDATE guarded by `status = 'dead'` in the repository, so a live job is never touched and two concurrent re-drives cannot both fire. The re-driven job re-enters `registerLocked` with the same key, finds the same row and goes through `resumeExisting`, which returns early on `isRegistered`, on a live lease, and on any failure that is not terminally `rejected`. No new path reaches the adapter.
- **The same-key exemption admits exactly the right set.** `isResumableUnderSameKey` (:934) is `pending || registering || isReattemptableFailure`, and `isReattemptableFailure` is `status === 'failed' && failureMode === 'rejected'` (`fiscal-registration-record.entity.ts:89`). `registered` and `in-doubt` fall through to `assertRegistrable`, whose `blocksFurtherRegistration` predicate (`:126`) blocks everything except a rejected failure and which scans ALL records for the order, same connection included. So an in-doubt sale is refused at enqueue with a 409 and never gets a job. Confirmed against the body's claim, and confirmed that `in-doubt` is never resolved by a resend.
- **`record: null` while `queued`.** Handled in the DTO, in `showFiscalSlot` (opens on outstanding work rather than on a record), in the pending notice, and in `getRegistrationProgress`, which reads the job precisely because the record is absent in that window. The `sales-document-panel` and `fiscal-poll-interval` suites cover it.
- **`stalled` renders as neither a wait nor a rejection.** `fiscalProgressPollInterval` returns `false` for it, so it does not spin, and it renders as a warning Alert with an action rather than as a failure state. Correct.
- **Honest API descriptions.** The 202 description states it says the request was recorded "and nothing more - never that the sale was registered, and never that it will be", and the poll description states it takes no lock and reports no estimate. Both check out against the code.
- The pending copy has correctly dropped any "safe to leave this page" phrasing and carries no estimate.

### Non-blocking

3. `fiscal-registration.service.ts:257` calls `requeueDeadByIdempotencyKey` unconditionally after `schedule`, including on the ordinary first request where it can only be a no-op. Harmless (one guarded UPDATE), but the docblock reads as if it were conditional.
4. `resolveFiscalRegistrationProgress` ranks `leaseLive` above `status === 'registered'`, so a registered record with a lingering lease would report `running`. Unreachable today - every terminal write sets `leaseExpiresAt: null` - but the module doc says "a settled record OUTRANKS the job", which is not quite what the code does.
5. `package.json`'s `check:invariants` chain is edited on this branch and on #2692. Whichever merges second must keep both new guard entries.

Re-review once 1 and 2 are addressed.

norbert-kulus-blockydevs:
Both blocking findings are fixed in 9a3e42d, on top of a catch-up merge with `epic/sales-documents-ux` (M1 and M5). The `package.json` conflict kept both guard scripts.

**Blocking 1, the worker mock.** `apps/worker/src/sync/handlers/__tests__/fiscalization-register.handler.spec.ts` now mocks `requestRegistration`, `assertRegistrable` and `getRegistrationProgress`. `apps/worker` type-checks and the suite runs its 26 tests again. The cause was mine: I widened a shared interface and type-checked the three packages I had edited rather than every package that mocks it. All four are green now, and `grep -rl "IFiscalRegistrationService"` confirms there is no fifth.

**Blocking 2, `stalled` asserting an absence.** You are right, and the split is where you pointed. The two paths into it differ on whether the provider can already have been called:

- a `pending` record is written before any outbound call and never claimed, and no record at all means the call was never reached, so nothing crossed the boundary and the absence is provable;
- a `registering` record whose lease expired is a crashed attempt that may already have crossed it, since the process dying says nothing about whether the request was processed.

Rather than find wording true of both, the crashed shape is now its own value, `interrupted`. `stalled` keeps its sentence, tightened to "stopped before it reached the provider, so nothing was registered". `interrupted` says: "An attempt stopped without an answer. The attempt had started, so OpenLinker cannot tell whether the provider registered this sale. Asking again resumes the same registration rather than starting a new one." It claims nothing about the provider deduplicating, only about what OpenLinker does.

`resumeExisting` is untouched, and an `interrupted` record is still re-attemptable under the same key exactly as before. The union, the frontend mirror, the mirror guard, the poll rule and the panel all carry the new value, and three tests pin the distinction: the rule returns `interrupted` for the crashed shape, and the panel renders no absence claim for it.

**Non-blocking, both done.** The dead-job re-drive is asked only of a job whose status is `dead`; it was always guarded inside the re-drive itself, so the unconditional call was a wasted write that also read as though every request re-drove something. And `registered` and `failed` are now checked before the claim flag in the progress resolver, matching what the docblock already said.

Gate on the merged branch: `libs/core`, `apps/api`, `apps/worker` and `apps/web` type-check; `pnpm check:invariants` exit 0 with 8 progress values in sync; core fiscalization 176, worker handler 26, api fiscalization 33, web fiscalization plus orders 367.

norbert-kulus-blockydevs:
# PR review — #2698 (M3, asynchronous manual registration), round 2

**Scope**: 31 files, +2068 / -148. Re-review after 9a3e42d and the catch-up merge with `epic/sales-documents-ux` (M1, M5).

---

## Round 1 findings — verified fixed

**Blocking 1, the worker mock.** `apps/worker/src/sync/handlers/__tests__/fiscalization-register.handler.spec.ts:63` now mocks `requestRegistration`, `assertRegistrable` and `getRegistrationProgress`. `pnpm --filter @openlinker/worker type-check` is clean and the suite runs its 26 tests. `grep -rl "IFiscalRegistrationService" apps/ libs/` shows no other full-interface mock (`invoice.service.spec.ts` uses `Pick<…, 'getByOrderId'>` and is unaffected).

**Blocking 2, `stalled` asserting an absence.** Correctly split rather than reworded. I checked the claim holds at the source, not just in the copy: `pending` is written in exactly one place (`fiscal-registration.service.ts:408`, the create), and `claimForRegistration` is the only transition to `registering` — so on both remaining `stalled` paths (no record, or `pending`) the adapter was provably never called. `registering` with an expired lease is now `interrupted`, and its copy claims only that the attempt started and that OpenLinker cannot tell. `resumeExisting` and the repository's claim predicate are untouched, and both states remain re-attemptable under the same key (`isResumableUnderSameKey` covers `pending` and `registering`), so the retry buttons the panel offers are ones the write path actually honours.

**Non-blocking, both done.** The re-drive is now `job.status === 'dead' ? … : false`; the read-then-act is harmless because `requeueDeadByIdempotencyKey` is itself a guarded conditional UPDATE, so a lost race returns `false` rather than double-driving. And `registered` / `failed` are checked before `leaseLive` in the resolver, matching the docblock.

---

## New findings

**[BLOCKING]** — `apps/web/src/features/orders/components/sales-document-panel.tsx:329, :578`
> The panel now contradicts itself during the exact window this mini-epic exists to make visible. `showFiscalSlot` was widened to open on `fiscalWorkOutstanding` with no record — correct, and the point of the change — but the header badge is still `deriveFiscalReceiptDisplayStatus(fiscalRecord)`, which returns `'not-registered'` for a null record (`derive-fiscal-receipt-display.ts:17`) and renders as **"Not registered"**. So for the whole queued window the slot reads `fiscal receipt · Not registered` in the header while the body underneath reads "Registering with the provider. This continues if you leave the page."
>
> This is reachable on every manual registration, not an edge case: the mutation invalidates the progress query, `progress` comes back `queued`, and no record exists until the job runs. The header badge is the fastest-read element on the surface, and it says the opposite of the body.
>
> The badge vocabulary already has the right values — `pending` renders "Queued" and `registering` renders "Registering" (`fiscal-receipt-status-badge.tsx:39-40`) — so this is derive-from-progress when the record is absent, not new copy. Please also confirm the resulting badge for `stalled` / `interrupted`, which currently inherit whatever the record's own status maps to.

**[IMPORTANT]** — `libs/core/src/fiscalization/application/services/fiscal-registration.service.ts:200-216`
> `register()`'s docblock was orphaned by the insertion. The long block ending "Mirrors `InvoiceService.issueInvoice`'s lock shape: acquired here, released in `finally`; a failed acquisition answers from PERSISTED STATE ONLY (`registerContended`), never crossing the provider boundary while a peer holds the lock." now sits immediately above `requestRegistration`, which takes no lock, has no `finally` release and never crosses the provider boundary at all. `register()` is left with no docblock.
>
> On the fiscal write path a docblock that describes a serialization guarantee the method below it does not have is worse than none — it is the same failure mode as the copy findings, one layer down. Move the lock docblock back onto `register()`.

**[IMPORTANT]** — `apps/web/src/features/orders/components/sales-document-panel.tsx:329`
> `progress === 'stalled'` with **no** record is unreachable in the UI. That combination is real — a job that exhausted its retries before `register()` ever created a row — but `showFiscalSlot` opens only on `fiscalRecord !== null || fiscalWorkOutstanding`, and `stalled` is neither, so the panel falls through to the empty state.
>
> Nothing false is shown (the empty state offers "Register receipt", which is the correct remediation), so this is under-reporting rather than a wrong claim — but it silently drops the one state whose whole purpose is to tell an operator that a previous request gave up. Widening the condition to include `stalled` is a one-line fix; `interrupted` always has a record and needs nothing.

---

## Positive observations

- The resolver's two-fact input (`record` + `job` liveness) is the right shape, and reducing the sync job's status to a three-value `FiscalRegistrationJobLiveness` keeps the sync vocabulary out of a fiscalization domain rule. `succeeded` mapping to `none` is subtle and correctly reasoned in `readJobLiveness`.
- `assertRegistrable` delegating to the private `assertNotAlreadyRegistered` is the right way to expose an advisory guard: one predicate, and the interface says plainly that passing it licenses nothing.
- The key format and the payload mapper collapsing to one definition each removes a genuine drift surface between the manual and auto-issue paths, where drift would have meant two registrations of one sale.
- The invariant script restating its sibling's parsers rather than importing them is correct — these scripts run their check on import.
- The integration test asserts the property that actually moved (one job row per sale, and the payload key equal to the job row key), rather than re-asserting what the record-side spec already covers.

## Architecture / standards

Boundaries are clean: cross-context access is through `I*Service` (`ISyncJobsService`, `IFiscalRegistrationService`), no repository port crosses a context, the new types file qualifies under the pure-rule exception in `engineering-standards.md § The pure-rule exception`, the frontend mirror is guarded by a `check-*-mirror.mjs` wired into `check:invariants`, and the catch-up merge kept both M1's and M3's guard scripts in the `check:invariants` chain. No `any`, no `console.log`, no migration needed.

---

## Final assessment

The two round-1 blockers are properly fixed, and blocking 2 was fixed at the right level — a new value in the vocabulary rather than wording that is true of one path and false of the other. The remaining blocker is the same defect class one surface further out: the header badge asserting "Not registered" over a body that says a registration is in flight, in a window this PR itself made visible.

**Merge readiness**: ❌ **Request changes**

**Priority fixes**
1. Derive the fiscal badge from `fiscalProgress` when no record exists, so the header stops contradicting the body during the queued window.
2. Move the lock docblock back onto `register()`.
3. Open the fiscal slot for `stalled` with no record.

norbert-kulus-blockydevs:
All three round-2 findings are fixed in ae9c758.

**Blocking, the header contradicting the body.** `deriveFiscalReceiptBadgeStatus(record, progress)` is a new pure helper beside the record-only one, and the panel's header badge uses it. Progress wins for exactly the four states a record cannot express: `queued` badges as `pending` ("Queued"), `running` as `registering`. The other two are the interesting ones, and I did not map them to the nearest record status:

- `stalled` badges as `not-registered`, not `pending`. Nothing is running and nothing reached the provider, so the neutral badge is literally true, while "Queued" would say the work is waiting its turn when nothing will pick it up.
- `interrupted` badges as `in-doubt` ("Unconfirmed"), not `registering`. The attempt stopped without answering, so whether the sale is registered is unknown, which is what that badge already means. `registering` would claim an attempt is still in flight.

Everything else defers to the record, so the badge and the body cannot disagree. Four unit tests on the helper plus a panel test asserting the queued window shows "Queued" and no longer shows "Not registered".

**Important, the record-less stalled state.** `showFiscalSlot` now also opens on `progress === 'stalled'`. A job that gave up before `register` ever wrote a row is a real state and was falling through to the empty state, which said nothing about the stopped request. A panel test covers it. `interrupted` always has a record and needed nothing.

**Important, the orphaned docblock.** The lock docblock is back on `register()`. It had been left above `requestRegistration`, which takes no lock, has no `finally` release and never reaches a provider, so it described a serialization guarantee that method does not have.

Gate: `libs/core`, `apps/api`, `apps/worker` and `apps/web` type-check; `pnpm check:invariants` exit 0; core fiscalization 176, web fiscalization + orders 346 plus the 4 new helper tests.

norbert-kulus-blockydevs:
# PR review — #2698 (M3, asynchronous manual registration), round 3

**Scope of this round**: the six files touched by ae9c758 (`derive-fiscal-receipt-display.ts` + its test, `sales-document-panel.tsx` + its test, `features/fiscalization/index.ts`, `fiscal-registration.service.ts`). Everything else was verified in round 2 and is unchanged.

---

## Round 2 findings — verified fixed

**Blocking, the header contradicting the body.** `deriveFiscalReceiptBadgeStatus(record, progress)` takes the badge from progress for the four states a record cannot express and falls back to the record for everything else. I walked every reachable combination rather than only the reported one:

- `queued` / `running` with no record -> `pending` ("Queued") / `registering`, which is what the body says.
- `stalled` -> `not-registered`, not `pending`. Correct call: "Queued" would assert the work is waiting its turn when nothing will pick it up.
- `interrupted` -> `in-doubt` ("Unconfirmed"), not `registering`. Also correct: the attempt stopped without answering, which is precisely what that badge means.
- **A record on a different connection falls to the default arm and is described by the record**, not by progress. That case is reachable — `fiscalRecord` is the newest record across all connections while the poll is connection-scoped — and the switch gets it right, because when a record exists the panel scopes `fiscalProgressConnectionId` to *that record's* connection, so progress and record can never describe different connections. The badge and the body cannot diverge in either direction now.

**Important, the record-less stalled state.** `showFiscalSlot` opens on `progress === 'stalled'` as well. I checked the widened branch for null-safety: the three non-optional `fiscalRecord.` accesses inside the slot each sit behind their own `&& fiscalRecord`, the head reference is optional-chained, and `issueBlockedByReceipt` still requires a record, so nothing changes for the cross-kind block. A panel test covers the state.

**Important, the orphaned docblock.** The lock docblock is back above `register()`, and `requestRegistration` carries its own two-line one pointing at the interface. Both methods now describe themselves.

---

## New findings

None blocking.

**[SUGGESTION]** — `apps/web/src/features/fiscalization/components/fiscal-receipt-status-badge.tsx:48`
> `PULSE` contains `'in-doubt'`, so the `interrupted` badge animates its dot. Per `frontend-ui-style-guide.md § Status Badge`, `pulse` is "for live/syncing states", and nothing is running in `interrupted`. Worth knowing rather than fixing here: the same pulse already applies to an ordinary in-doubt record, which is equally not running, so this is inherited from a pre-existing choice rather than introduced by this PR — and the label ("Unconfirmed") is accurate either way. If it is revisited, it belongs with the panel work in M9 and should move the whole `in-doubt` case, not just this new route into it.

---

## Assessment

The badge fix is the right shape: one pure helper with the record-only derivation still beside it, progress winning only for the states a record genuinely cannot express, and a test per state including the two that are easy to map to the nearest-looking record status and get wrong. Nothing in the round-3 diff introduces new coupling, new vocabulary or a new claim.

Standing summary for the mini-epic: exactly-once is unchanged and now has one key definition and one payload mapper across the manual and auto-issue paths; the progress read is pure and connection-scoped; `stalled` and `interrupted` are kept apart because only one of them permits an absence to be asserted; the frontend union is guarded by a mirror script in `check:invariants`; and the catch-up merge kept both M1's and M3's guard scripts.

Gate (unchanged since ae9c758): `libs/core`, `apps/api`, `apps/worker` and `apps/web` type-check; eslint clean on the changed paths; `pnpm check:invariants` exit 0 with 8 progress values in sync; core fiscalization 176, worker handler 26, api fiscalization 33, web fiscalization + orders 346 plus the new helper and panel tests.

**Merge readiness**: ✅ **Approve** — no blocking findings.

Issue #1826, opened 2026-07-23
[TASK] Shipments list — inline retry from a failed row + carrier-message role visibility


norbert-kulus-blockydevs:
## Problem / Context

Spun out of the #1800 failure-surfacing work: current `main` renders the persisted `errorMessage` + `failedAt` in the `/shipments` status cell and the order shipment panel (PR #1821). `traceId` is **not** persisted on the shipment - it exists only in the live generate-label error envelope, so a revisited `failed` row shows `errorMessage` + `failedAt` only.

Two operational follow-ups the surface itself doesn't yet cover:

### 1. No recovery action from the list

`/shipments` shows *why* a shipment failed, but the only recovery path is: open the order → shipment panel → regenerate the label. For volume triage - e.g. a batch of DPD `NOT_PROCESSED` rejections all caused by one bad sender postcode - the operator re-navigates per order. A recovery affordance on a `failed` row closes the loop where the operator is already scanning.

### 2. Carrier-message visibility vs role model

`Shipment.errorMessage` is the carrier's raw validation message, which for DPD can embed address fragments (sender/recipient postcode). The multi-role model already exists (#1363 - the viewer role has read-only `/shipments`), so this is a **current** decision, not hypothetical. **Resolved**: the `viewer` role here is the internal demo/trial role (prospective users clicking through the full flow short of a final DB-mutating submit) - not trusted enough to see the raw carrier message. See "Viewer-role error visibility" below.

## Design mockup

Full interactive mockup (states gallery + implementation notes), updated with the resolved decisions below and fixed after 6 `/tech-review` passes across 2 rounds — round 1 caught the `EntityLabel`/`FilterBar` gaps, the severity-label-as-second-control problem, the impossible `omp` `in-transit` status, and the flagship "append-only" example being wrong for the exact pre-waybill DPD failure it demoed; round 2 (re-verifying the fixes with fresh independent passes) caught that "Cancel + re-issue" was *still* modelled as one atomic action when the real system requires two separate operator steps, that `EntityLabel`'s `name` is never auto-truncated, and a couple of UX/motion gaps in the interactive demo. **Reconciled again at implementation-plan time** against the live worktree (fresh off `main`, which has moved since the mockup was built) - see the "Plan-time reconciliation" notes below: https://claude.ai/code/artifact/ab2410fe-e5d1-453c-8660-aafb149f1d79

## Domain facts that shape the UI

[...]

PR #2700, opened 2026-08-31
fix(shipments): distinguish label-download failure classes instead of one generic toast

### LINKED ISSUE #2671: [BUG] Frontend — label download collapses every API failure into one "Try again" toast (and fails silently after generation)
## Problem / Context

Downloading a shipment label from the order-detail **Shipment** panel fails with a single, generic toast: *"Could not download the label. Try again."* — regardless of **why** it failed. Observed on demo: `/orders/ol_order_911c5321d6864a2f9a1ac61aa0824dce`, shipment in `Label ready` / `GENERATED`, `Download label` clicked (see screenshot). **Not a one-off** — the same failure was recorded during the PostHog session-recording review on a different order (`cfb7fbc0-8499-11f1-878c-cf25bca0b614`, see the source finding in #2181).

The backend answers this route with **four distinct, deliberately-shaped outcomes** (`apps/api/src/shipping/http/shipment.controller.ts`, `GET /shipments/:id/label` L200-233 + `toHttpException` L495-554):

| Status | Cause | Retry helps? |
|---|---|---|
| 404 | `ShipmentNotFoundException` | no |
| 422 | `LabelNotAvailableException` (no label generated yet) | no |
| 422 | `LabelDocumentNotSupportedException` (provider returns no label documents) | **never** |
| 502 | `ShippingProviderRejectionException` — carries `message` + `providerCode` + field-level `details` (#1428), role-redacted for viewers (#1826) | maybe |
| 502 | `ShippingProviderAuthException` — our stored carrier credentials were rejected | not by the operator |
| 500 | unclassified | maybe |

The FE collapses all of them into one string and, worse, **prescribes the wrong action**: "Try again" is actively misleading for 404/422, where retrying can never succeed. The operator has no way to tell "the carrier isn't answering right now" from "this carrier will never hand us a PDF" from "our credentials are dead" — three situations with three different next steps (wait / stop trying / call an admin). The `providerCode`, deliberately kept as a non-PII support reference for exactly this purpose, is dropped on the floor.

The information is already on the client: `execute()` normalises error bodies **even on binary endpoints** into `ApiError { status, message, details }` (`apps/web/src/app/api/api-client.ts` L299-306, `readResponseBody` L199-211), and `useLabelDownload` already catches it into local `error` state (`apps/web/src/features/shipments/hooks/use-label-download.ts` L37-40). The call sites simply discard it:

```ts
// apps/web/src/features/orders/components/shipment-action-buttons.tsx L181-188
void labelDownload.download(shipment.id).then((ok) => {
  if (!ok) {
    showToast({ tone: 'error', description: 'Could not download the label. Try again.' });
  }
});
```

Same code, duplicated verbatim, in `apps/web/src/features/shipments/components/shipment-row-detail.tsx` L189-195.

### Second defect — the failure is *silent* right after generation

`apps/web/src/features/orders/components/generate-label-form.tsx` L418 auto-downloads the freshly-issued label:

```ts
if (result.shipment?.labelPdfRef) {
  void labelDownload.download(result.shipment.id);   // no .then, no failure branch
}
```

There is **no failure handling at all** here. The operator sees the green *"Label generated / Tracking number will appear within ~5 minutes"* toast, no file arrives, and nothing says why — they reasonably conclude the browser blocked a popup. This is the more damaging of the two, because it lands at the exact moment the operator stops paying attention.

## Proposed Solution

1. **One shared mapper** for label-download failures (FE, e.g. `features/shipments/lib`) turning `ApiError` into `{ title, description, retryable }`, so the two duplicated call sites and the auto-download path can't drift:
   - `404` → shipment no longer exists (refresh the view)
   - `422` → no label to download yet / this carrier does not provide label documents — **no "try again"**
   - `502` → carrier-side failure; surface the carrier message + `providerCode` when the caller holds `shipments:write` (the API already redacts it for viewers — do not re-redact on the FE)
   - `0` (network) / timeout / `5xx` → the only cases that keep a "Try again" affordance
2. **Wire the auto-download path** in `generate-label-form.tsx` to the same mapper — a failed auto-download must produce a visible message, distinct from the "Label generated" success (the label *was* bought; only the fetch failed — the copy must not imply the operator has to regenerate).
3. **Dev note / trap:** `download()` returns `boolean` and stores the error in hook state. A caller cannot read `labelDownload.error` inside `.then()` — that closure holds the pre-click render's value and will be stale. Change `download()` to resolve with the error (e.g. `Promise<{ ok: true } | { ok: false; error: ApiError }>`) rather than reading it back off the hook.

Backend needs no change — it already carries the distinction.

## Classification

**Type**: Frontend
**Layer**: Interface (feature components + hook)
**File(s)**:
- `apps/web/src/features/shipments/hooks/use-label-download.ts`
- `apps/web/src/features/orders/components/shipment-action-buttons.tsx`
- `apps/web/src/features/shipments/components/shipment-row-detail.tsx`
- `apps/web/src/features/orders/components/generate-label-form.tsx`

## Dependencies

- **Source: #2181** (PostHog session recordings — findings). This issue is the write-up of the *"Label download does not work"* entry under *Bugs and other notes* there; #2181 also records the second occurrence and stays the parent document for the remaining findings from that review.
- Adjacent finding in the same review: #2181 §03 (no search field on **Orders** / **Shipments**) surfaced *because of* this bug — the operator hit the failed download and then had no way to look the order up on the Shipments list to check whether it worked there. Independent fix, but the two share a session.
- Related: #884 (label download endpoint + panel CTA — the surface this fixes), #1428 (structured provider rejection: `providerCode` + `details`), #1826 (carrier-message visibility gated on `shipments:write`), #953 (same family: post-dispatch feedback that didn't match reality).

## Assumptions

- Final operator-facing copy per branch is settled during implementation; this issue fixes the *branching*, not the exact wording.
- `Download label` stays enabled on its current predicate (`labelPdfRef !== null` + `CAN_DOWNLOAD_LABEL`). Pre-disabling it for carriers that cannot serve label documents would need a capability flag on the shipment payload that does not exist today — out of scope here, worth a follow-up if 422/`LabelDocumentNotSupportedException` turns out to be common.
- Root cause of the captured failures is not yet identified (the demo shipment shows `GENERATED` with a tracking number, so 422-not-yet-generated is unlikely); identifying it needs the API logs for those shipments — this issue is about *the operator being told which failure it was*, and the fix makes that root cause diagnosable from the UI instead of the server log.

## Acceptance Criteria

- [ ] A failed label download surfaces a message that reflects the actual failure class (404 / 422 / 502 / network-timeout / 5xx), not one generic string.
- [ ] "Try again" (or any retry affordance) appears **only** for genuinely retryable classes — never for 404 or 422.
- [ ] On a 502 provider rejection, a caller with `shipments:write` sees the carrier's message and `providerCode`; a viewer sees the API-redacted message unchanged, and the FE adds no redaction of its own.
- [ ] A failed **auto-download after label generation** produces a visible error message, and its copy does not suggest the label failed to generate.
- [ ] Both manual call sites (order-detail panel and `/shipments` row detail) render identical copy for identical failures, from one shared mapper.
- [ ] Component tests cover each failure class at both manual call sites plus the auto-download path.
- [ ] `pnpm lint` / `type-check` / `test` green.

norbert-kulus-blockydevs:
## Summary

- Downloading a shipment label collapsed every backend outcome (404 / 422 / 502 / network / 5xx) into one toast, "Could not download the label. Try again." - actively wrong for 404/422, where no retry ever helps.
- The auto-download that fires right after "Generate label" failed completely silently: no toast, no error, just a missing file.
- Adds one shared mapper, `resolveLabelDownloadError` (`apps/web/src/features/shipments/lib/label-download-error.ts`), turning the failure into `{ title, description, retryable, tone }`, typed against the existing 4-way `RetryabilityClass` (`transient | permanent | auth | unknown`) so a carrier-credential rejection renders a distinct "needs admin" state instead of collapsing into the same "no retry" bucket as a plain 404.
- Wires all three call sites through it: `ShipmentActionButtons` (order-detail panel), `ShipmentRowDetail` (`/shipments` row accordion), and `GenerateLabelForm`'s auto-download - so they can no longer drift into three different strings for the same backend outcome.
- `useLabelDownload.download()` now resolves with the failure instead of stashing it in hook state, closing the stale-closure trap the issue itself flagged: a `.then()` callback reading `labelDownload.error` would have closed over the pre-click render's value.

## Design mock

Before/after mock (real app components, dark theme, all 7 failure classes including the auth/"needs admin" state and the role-redaction toggle for the one branch that's actually gated server-side): https://claude.ai/code/artifact/bbd49ce3-3de8-460a-bbb4-5d6255bbc60f

## Discriminators (verified against source, not guessed)

`apps/api/src/shipping/http/shipment.controller.ts`'s `toHttpException` was read directly rather than assumed:

- 404 (`ShipmentNotFoundException`) and both 422s (`LabelNotAvailableException` / `LabelDocumentNotSupportedException`) carry no structured exception-name field on the wire, only `{ statusCode, message }` - the two 422s are told apart by a message substring, kept in lockstep with the exception's own message template (documented in the mapper's header comment).
- 502 `ShippingProviderRejectionException` builds its body as `{ message, providerCode, details }` (providerCode always present, message already role-redacted server-side per #1826) - the FE echoes it verbatim and never re-redacts.
- 502 `ShippingProviderAuthException` passes a plain string to `BadGatewayException`, so it has no `providerCode` key at all - that's the discriminator between the two 502 cases, without needing a backend change.
- Status 0 (`ApiError.isNetworkError()` / `ApiError.fromTimeout`) is a real, separate case from the unclassified 500 - the request never reached the API at all.

## Test plan

- [x] `pnpm --filter @openlinker/web lint` - 0 errors (pre-existing warnings elsewhere, untouched)
- [x] Unit tests for the mapper covering every branch (`label-download-error.test.ts`)
- [x] Component tests at both manual call sites (`shipment-action-buttons.test.tsx`, `shipment-row-detail.test.tsx`) - `it.each` over all 7 failure classes, asserting distinct copy per class and that the old generic string never appears
- [x] Component test for the auto-download path (`generate-label-form.test.tsx`) - asserts the success toast still fires (the label WAS bought) and a second, distinct failure toast appears, never implying generation itself failed
- [x] Manual QA in a running dev stack across mobile/tablet/desktop breakpoints - no CSS or layout changed (same `Toast`/`Button` components, same responsive rules already covering `.toast-region` / `.shipment-action-buttons`), so no new breakpoint risk was introduced, but this is still pending a real-browser pass before merge

Closes #2671

### DIFF
--- a/apps/e2e/playwright.config.ts
+++ b/apps/e2e/playwright.config.ts
@@ -305,5 +305,16 @@ export default defineConfig({
       // making any assertion more truthful.
       use: { ...devices['Desktop Chrome'] },
     },
+    {
+      // Label-download failure-class mapping on the /shipments row accordion
+      // (#2671). Every OL route the page touches is stubbed in-test, following
+      // the `wizard-blockers` shape - no seeded stack, no shared auth artifact,
+      // just a served web app. `retries: 1` (nothing is mutated; the download
+      // call always fails by design).
+      name: 'label-download-errors',
+      testMatch: /label-download-errors\/.*\.spec\.ts/,
+      retries: 1,
+      use: { ...devices['Desktop Chrome'] },
+    },
   ],
 });
--- a/apps/e2e/tests/label-download-errors/label-download-errors.spec.ts
+++ b/apps/e2e/tests/label-download-errors/label-download-errors.spec.ts
@@ -1,0 +1,98 @@
+/**
+ * Label-download error mapping - visual evidence (#2671)
+ *
+ * The operator-facing defect this fixes: every backend outcome from
+ * `GET /shipments/:id/label` collapsed into one toast, "Could not download
+ * the label. Try again." - wrong for 404/422, where retrying can never help.
+ * This walks the real `/shipments` row accordion through every failure class
+ * against a stubbed OL API (no seeded stack, no shared auth artifact - the
+ * session bootstrap is stubbed too, same shape as `wizard-blockers`), so the
+ * project needs only a served web app.
+ *
+ * Each case writes a named screenshot into its own test output directory and
+ * attaches it to the HTML report - a copy change is best evidenced by a
+ * picture of the copy, not by an assertion string alone.
+ *
+ * @module tests/label-download-errors
+ */
+import { test, type Page, type TestInfo } from '@playwright/test';
+import {
+  SHIPMENT_ID,
+  stubLabelDownloadFailure,
+  stubShipmentsPageApi,
+  type FailureCase,
+} from './stub';
+
+async function captureState(page: Page, testInfo: TestInfo, name: string): Promise<void> {
+  const file = testInfo.outputPath(`${name}.png`);
+  await page.screenshot({ path: file });
+  await testInfo.attach(name, { path: file, contentType: 'image/png' });
+}
+
+/**
+ * Open the /shipments page, expand the seeded row, click Download label.
+ * Below the mobile breakpoint `DataTable` renders cards instead of the
+ * desktop accordion (`renderCards = cardView && isMobile`,
+ * `shared/ui/data-table.tsx`), and the mobile card's own disclosure toggle
+ * reads "View full details" / "Hide details", not "Expand details for
+ * shipment ...". Both toggles land on the same `ShipmentRowDetail`.
+ */
+async function triggerDownload(page: Page, viewportWidth = 1280): Promise<void> {
+  await page.goto('/shipments');
+  const toggle =
+    viewportWidth < 768
+      ? page.getByRole('button', { name: 'View full details' })
+      : page.getByRole('button', { name: `Expand details for shipment ${SHIPMENT_ID}` });
+  await toggle.click();
+  await page.getByRole('button', { name: 'Download label' }).click();
+}
+
+const CASES: Array<{ kind: FailureCase; title: string }> = [
+  { kind: 'not-found', title: 'no shipment matches this id' },
+  { kind: 'not-yet-generated', title: 'no label to download yet' },
+  { kind: 'carrier-unsupported', title: 'this carrier doesn’t provide a downloadable label' },
+  { kind: 'provider-rejection', title: 'the carrier rejected the request' },
+  { kind: 'provider-auth', title: 'our stored carrier credentials were rejected' },
+  { kind: 'unclassified', title: 'something went wrong' },
+  { kind: 'network', title: 'couldn’t reach openlinker' },
+];
+
+test.describe('label-download failure toasts, desktop', () => {
+  for (const { kind, title } of CASES) {
+    test(`${kind} renders "${title}"`, async ({ page }, testInfo) => {
+      await stubShipmentsPageApi(page);
+      await stubLabelDownloadFailure(page, kind);
+      await triggerDownload(page);
+
+      // `.toast__title` scoped explicitly: Radix's aria-live announcer
+      // duplicates the toast text into a second, hidden `role="status"`
+      // region, which a bare `getByText` also matches (strict-mode violation).
+      await page.locator('.toast__title').getByText(new RegExp(title, 'i')).waitFor();
+      await captureState(page, testInfo, `desktop-${kind}`);
+    });
+  }
+});
+
+// Breakpoints (#2671 review): no CSS changed by this fix - same `Toast` /
+// `Button` components, same existing responsive rules - but the fix is
+// worth seeing at the widths that actually stack `.shipment-action-buttons`
+// and reflow `.toast-region`, not just assumed from the unchanged CSS.
+const BREAKPOINT_CASES: Array<{ kind: FailureCase; width: number; height: number; label: string }> = [
+  { kind: 'not-found', width: 375, height: 812, label: 'mobile' },
+  { kind: 'provider-auth', width: 375, height: 812, label: 'mobile' },
+  { kind: 'provider-rejection', width: 768, height: 1024, label: 'tablet' },
+];
+
+test.describe('label-download failure toasts, responsive', () => {
+  for (const { kind, width, height, label } of BREAKPOINT_CASES) {
+    test(`${kind} at ${label} (${width}x${height})`, async ({ page }, testInfo) => {
+      await page.setViewportSize({ width, height });
+      await stubShipmentsPageApi(page);
+      await stubLabelDownloadFailure(page, kind);
+      await triggerDownload(page, width);
+
+      await page.waitForTimeout(300); // let the toast slide-in animation settle
+      await captureState(page, testInfo, `${label}-${kind}`);
+    });
+  }
+});
--- a/apps/e2e/tests/label-download-errors/stub.ts
+++ b/apps/e2e/tests/label-download-errors/stub.ts
@@ -1,0 +1,159 @@
+/**
+ * Label-download error stub (#2671)
+ *
+ * Stubs the `/shipments` page's whole read surface so every label-download
+ * failure class can be driven from a served web app alone - no seeded stack,
+ * no shared auth artifact. Mirrors `tests/wizard-blockers/wizard-stub.ts`'s
+ * session-bootstrap shape (auth refresh + `/auth/me` + a `system/**` catch-all)
+ * since that stub already proved this pattern hermetic for a permission-gated
+ * page.
+ *
+ * @module tests/label-download-errors
+ */
+import type { Page, Route } from '@playwright/test';
+
+export const CONNECTION_ID = '00000000-0000-4000-8000-000000002671';
+export const SHIPMENT_ID = 'ol_shipment_2671';
+export const ORDER_ID = 'ol_order_2671';
+
+const SESSION_PERMISSIONS = ['shipments:read', 'shipments:write', 'connections:read', 'connections:write'];
+
+function json(route: Route, body: unknown, status = 200): Promise<void> {
+  return route.fulfill({ status, contentType: 'application/json', body: JSON.stringify(body) });
+}
+
+/** Stub every OL route the `/shipments` page reads on load. */
+export async function stubShipmentsPageApi(page: Page): Promise<void> {
+  const now = new Date().toISOString();
+
+  await page.route('**/v1/auth/refresh', (route) => json(route, { access_token: 'e2e-2671-token' }));
+  await page.route('**/v1/auth/me', (route) =>
+    json(route, {
+      id: 'usr_e2e_2671',
+      username: 'e2e-operator',
+      email: 'e2e-operator@openlinker.local',
+      role: 'admin',
+      permissions: SESSION_PERMISSIONS,
+      analyticsConsent: false,
+    }),
+  );
+  await page.route('**/v1/system/**', (route) => json(route, {}));
+
+  await page.route('**/v1/connections', (route) =>
+    json(route, [
+      {
+        id: CONNECTION_ID,
+        name: 'DPD Poland',
+        platformType: 'dpd-polska',
+        status: 'active',
+        config: {},
+        credentialsBacked: true,
+        adapterKey: 'dpd-polska.rest.v1',
+        enabledCapabilities: ['OrderProcessorManager'],
+        supportedCapabilities: ['OrderProcessorManager'],
+        defaultRateLimit: null,
+        createdAt: now,
+        updatedAt: now,
+      },
+    ]),
+  );
+
+  await page.route('**/v1/shipments?*', (route) =>
+    json(route, {
+      items: [
+        {
+          id: SHIPMENT_ID,
+          orderId: ORDER_ID,
+          customerId: null,
+          connectionId: CONNECTION_ID,
+          shippingMethod: 'kurier',
+          status: 'generated',
+          providerShipmentId: 'DPD-9911',
+          paczkomatId: null,
+          sourceDeliveryMethodId: null,
+          deliveryIntent: 'address',
+          trackingNumber: null,
+          providerCode: null,
+          carrier: 'dpd',
+          labelPdfRef: 'dpd:label:9911',
+          dispatchedAt: null,
+          deliveredAt: null,
+          cancelledAt: null,
+          failedAt: null,
+          errorMessage: null,
+          createdAt: now,
+          updatedAt: now,
+          orderSummary: null,
+        },
+      ],
+      total: 1,
+      limit: 20,
+      offset: 0,
+    }),
+  );
+}
+
+/** One `GET /shipments/:id/label` failure per case, keyed exactly to what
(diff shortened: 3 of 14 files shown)


norbert-kulus-blockydevs:
## Visual evidence

Added a new hermetic Playwright project, `label-download-errors`
(`apps/e2e/tests/label-download-errors/`), that drives the real `/shipments`
row accordion through all 7 failure classes against a stubbed OL API - same
shape as the existing `wizard-blockers` project: no seeded stack, no shared
auth artifact, just a served web app. Confirmed 10/10 passing locally
against this branch's dev server.

Screenshots aren't uploaded here yet (this comment is being posted from a
non-interactive environment) - placeholders below for whoever attaches them
from the local run (`apps/e2e/test-results/**/*.png` after
`pnpm --filter @openlinker/e2e test:e2e:headed --project=label-download-errors`,
or `npx playwright test --project=label-download-errors` from `apps/e2e`
against a running `pnpm --filter web dev`).

### Desktop, all 7 classes
| 404 - not found | 422 - not generated yet | 422 - carrier unsupported |
|---|---|---|
| _01-desktop-404-not-found_ | _02-desktop-422-not-yet-generated_ | _03-desktop-422-carrier-unsupported_ |

| 502 - provider rejection | 502 - our credentials rejected (needs admin) | 500 - unclassified |
|---|---|---|
| _04-desktop-502-provider-rejection_ | _05-desktop-502-provider-auth-needs-admin_ | _06-desktop-500-unclassified_ |

| network |
|---|
| _07-desktop-network_ |

### Breakpoints

| Mobile (375px) - 404 | Mobile (375px) - needs admin | Tablet (768px) - provider rejection |
|---|---|---|
| _08-mobile-375-404-not-found_ | _09-mobile-375-502-provider-auth-needs-admin_ | _
10-tablet-768-502-provider-rejection_ |

No CSS changed by this fix (same `Toast` / `Button` components, same
existing responsive rules already covering `.toast-region` /
`.shipment-action-buttons`), so these three are confirmation, not new
coverage - the mobile card layout's own disclosure toggle ("View full
details") reaches the same `ShipmentRowDetail` as the desktop accordion, and
the toast renders correctly stacked over the card at 375px.

norbert-kulus-blockydevs:
## Tech-lead review

Verified the discriminator claims directly against `apps/api/src/shipping/http/shipment.controller.ts`'s `toHttpException` and `ShipmentLabelService.fetchLabel` — every branch (404 / 422×2 via message-substring / 502×2 via `providerCode`-key presence / network / unclassified) matches the actual exception set this route can throw, including the `LabelNotAvailableException` vs `LabelDocumentNotSupportedException` split.

**No BLOCKING or IMPORTANT issues.**

### Suggestions (optional)

1. `label-download-error.ts` — `hasProviderCode` treats "502 with no `providerCode` key" as the catch-all for the `auth` (credentials-rejected) branch. Correct today since only two 502-producing exceptions exist on this path, but a future third 502 case would silently render as "credentials rejected" too. Worth a short comment noting this is a closed-set assumption tied to the current exception list, since nothing else would catch drift if a third 502 case is added later.
2. File header on `label-download-error.ts` skips the two-paragraph Purpose/Context shape from `docs/engineering-standards.md § File Headers` — though the discriminator table it carries is arguably more useful documentation than the template would produce. Non-blocking.

### What's done well

- One shared pure mapper (`resolveLabelDownloadError`) instead of three drifting toast strings — exactly the kind of duplication this fix should remove.
- Correctly reuses the existing `RetryabilityClass` vocabulary rather than inventing a parallel one, and correctly avoids reusing `deriveRetryabilityClass` since it classifies a different kind of failure (persisted `providerCode` vs. an HTTP download failure) — explained in the header comment.
- Security: only the branch already role-redacted server-side (#1826) echoes backend text; the unclassified-500 fallback never echoes raw messages, which is the right default for an unmodeled shape.
- `useLabelDownload.download()` resolving with the result instead of stashing it in hook state is a legitimate fix for the stale-closure bug called out in the issue.
- Test coverage is thorough: a shared discriminator table intentionally duplicated across both manual call sites (to catch copy drift between them), an auto-download-path test, a full-branch pure-mapper unit suite, and an e2e visual suite across breakpoints.
- No CORE/backend changes, no architecture-boundary violations — clean frontend-only fix.

piotrswierzy (approved):

**Scope:** 13 files, +856/−22. Frontend only (`features/shipments` lib + hook, three call sites across `shipments` and `orders`), plus a new Playwright project and stub. No backend, schema, or core change — correct, since the distinction this surfaces already exists on the wire.

### Summary

This is the right fix for #2671 and the discriminator work behind it is unusually careful. Every branch is pinned to a claim about `shipment.controller.ts`'s `toHttpException`, and I verified all six against source: the 404 and both 422s really do carry only `{ statusCode, message }`; `ShippingProviderRejectionException` really does build `{ message, providerCode, details }` with `providerCode` present on both role branches; `ShippingProviderAuthException` really does pass a plain string and so has no `providerCode` key; and the two provider exceptions both extend `Error` independently, so the controller's check order is unambiguous and the auth/rejection split is stable. I also checked the piece the mapper's correctness silently depends on and the PR doesn't mention: `ApiError.details` holds the **whole parsed body** (`api-error.ts:36-46`), so `'providerCode' in error.details` reads the top-level key rather than the nested `details` object. Had `details` been the nested field, branch 5 would never fire and every 502 would render as a credential failure. It's right — worth stating, because it's a one-word difference between correct and confidently wrong.

### Positive observations

- **`retryable` typed as `RetryabilityClass` rather than a boolean** is the substantive design call, and the header explains why: a 404 and a dead carrier credential are both "your click won't fix this", but only one is the operator's problem. The header also explains why this is *not* `deriveRetryabilityClass` — that one classifies a generation failure off a persisted `providerCode` shape, this is a download-time HTTP failure with no persisted field. Shared vocabulary, not shared derivation, stated explicitly so the next reader doesn't "unify" them.
- **The security invariant is stated and holds.** Only branch 5 echoes a server message, and only because the backend already redacted it per #1826; every other branch including the unclassified-500 fallback emits fixed copy. That last part matters — the controller's 500 branch returns the raw `error.message`, and echoing an unmodeled shape to the operator is exactly how a stack fragment ends up in a toast.
- **`tone` derived from `retryable` via one `TONE_BY_RETRYABLE` record**, so a call site cannot pick its own tone per branch and drift. Same instinct as the shared mapper itself, one level down.
- **The auto-download fix gets the copy right.** `"Couldn't fetch the label file"` with the mapper's description, distinct from the success toast, and the comment names the actual hazard — a title implying generation failed would send the operator back through the form and risk a second dispatch. That is the more damaging of the two defects in the issue and it's handled with the right care.
- **The hook change is a genuine bug fix, not refactoring.** Resolving with the failure instead of stashing it in state closes the stale-closure trap the issue flagged; dropping `error` from the returned shape is safe (all three consumers verified, none read it).
- Test coverage matches AC: `it.each` over all seven classes at both manual call sites, asserting distinct copy *and* that the old generic string never appears — the negative assertion is what stops a regression re-collapsing the branches.

### 🟡 IMPORTANT — `canRetryLabelDownload` is exported and used nowhere

It's on the feature's public barrel and has no call site anywhere in `apps/web` or `apps/e2e`. AC #2 ("a retry affordance appears only for genuinely retryable classes") is currently satisfied *by subtraction* — the misleading "Try again." string is gone and no call site renders a retry control at all — so the helper is the unwired half of the intent.

`docs/frontend-ui-style-guide.md` § Implementation rules is explicit: *"Use the same primitive in a real page immediately after introducing it — no unused abstractions."* Either wire it (a "Try again" action on the toast for `transient`/`unknown`, which is what the AC seems to picture) or drop it and let the copy carry the distinction. Right now a reader can't tell which was intended, and an exported-but-unused predicate is the kind of thing that gets a second, subtly different implementation later.

### 🟡 IMPORTANT — the 422 substring contract isn't pinned by anything but a comment

Branch 3 keys on `error.message.includes('generate the label first')`. That string lives in `libs/core/src/shipping/domain/exceptions/label-not-available.exception.ts:16`, and **nothing asserts it** — I grepped; no core, api, or web spec references it.

The failure mode is silent and states something false: reword that message and `LabelNotAvailableException` falls through to branch 4, telling the operator *"This carrier doesn't provide a downloadable label — retrying here will never produce a file"* about a shipment whose label simply hasn't been generated yet. They'd stop trying on a carrier that works fine.

The mapper's own tests can't catch this (they construct the message themselves). This repo's established remedy for a browser-side copy of a backend fact is a `check-*-mirror.mjs` under `pnpm check:invariants`; that's heavier than this needs, but a one-line spec in `libs/core` asserting `new LabelNotAvailableException('x').message` contains the substring would turn a silent downgrade into a failing build. Cheap, and it makes the "keep the substring in lockstep" comment enforceable rather than aspirational.

### 🟢 SUGGESTION

- `LabelDownloadError.retryable` is computed on every branch but no consumer reads it today — same root as the first IMPORTANT above; wiring the retry affordance resolves both at once.
- The test-plan checkbox for the real-browser pass is still open and the PR says so. Worth closing before merge given the toast/`Alert` tone changes, even though no CSS moved.

### Merge readiness

✅ **Approve** — no architecture, boundary, security, or dependency-direction concerns; feature-barrel usage and cross-feature imports are correct (`shipment-action-buttons` via the `shipments` barrel, `shipment-row-detail` relative in-feature). Both IMPORTANT items are small and neither changes the design. I'd want the substring pinned before merge, since that one can regress from a well-meaning copy edit in a different package. CI was still pending at review time.

norbert-kulus-blockydevs:
Pushed 2232f550a addressing both IMPORTANT findings from review:

- **`canRetryLabelDownload` unused** — dropped the export (and its test), rather than half-wiring a "Try again" toast action. Wiring that for real would need a new action slot on the shared `ToastProvider` used by every toast in the app, which is bigger scope than this fix warrants. The mapper's header comment now says explicitly that no call site renders a retry affordance today and why, so a future reader isn't left guessing which was intended.
- **422 substring not pinned by anything but a comment** — added `libs/core/src/shipping/domain/exceptions/__tests__/label-not-available.exception.spec.ts` asserting `LabelNotAvailableException`'s message keeps the `'generate the label first'` substring the FE mapper keys on. A reword now fails the build instead of silently rerouting "label not generated yet" downloads into the wrong copy.

Both suggestions in the review point at the same root cause as the first IMPORTANT item, so they're resolved together.

Re the CI red: both integration-test failures are infra flakes unrelated to this diff (which only touches `apps/web` + `apps/e2e`) —
- `Run integration tests` (api): the whole suite failed with `ECONNREFUSED` / `this.postgres.Pool is not a constructor` — Testcontainers' Postgres never came up.
- `Run worker integration tests`: 1 failing test, `allegro-order-sync-e2e.int-spec.ts`, threw `OrderCreateContendedException` — a known flaky lock-contention race, nothing to do with shipments/label code.

`pnpm --filter @openlinker/web type-check`, `lint`, and the full web test suite (4001 passed) are all green locally, along with the new core spec.

Issue #2481, opened 2026-08-25
[TASK] Frontend — Channel and Top Products `.excl-note` wiring


norbert-kulus-blockydevs:
### Problem / Context

`channel-sales-table.tsx` and `product-sales-table.tsx` need a per-row annotation when that specific channel/product's numbers are under-counted by an open coverage category. The mockup iteration found and fixed two consistency bugs here (wrong category attributed to one row, missing annotation on another) — this task's acceptance criteria exist specifically to prevent re-introducing either class of bug in the real implementation.

### Proposed Solution

New shared `AnalyticsExclusionNote` component (`apps/web/src/features/analytics/components/analytics-exclusion-note.tsx`), taking a list of affected order/category pairs (from Phase 4's per-category `sampleOrderIds`, cross-referenced against the channel/product row's own order set — this cross-reference is the part that must be exact, since it's exactly where the mockup's consistency bugs came from). Rendered inline in each table row, opening the matching Phase 7 Task 7.3 modal on click.

### Classification

**Type**: Frontend
**Layer**: Feature component
**File(s)**: `apps/web/src/features/analytics/components/analytics-exclusion-note.tsx` (+`.test.tsx`), `channel-sales-table.tsx`, `product-sales-table.tsx`

### Dependencies

- Task 8.1 (shares the coverage-data wiring pattern), Phase 7.

### Assumptions

- The order-to-row cross-reference happens client-side against data already fetched for the table (no new endpoint) — if the affected-order lists from Phase 4 don't carry enough identifying info (SKU/channel id) to cross-reference cleanly, that's a signal Phase 4's detector output needs a small shape addition; flag it back rather than approximating the match.

### Acceptance Criteria

- [ ] A row's `.excl-note` category is provably correct against its own underlying order data (test with a fixture where two categories are both present in the dataset, asserting each row gets the right one — the exact bug class the mockup hit).
- [ ] No row belonging to an open category's affected set is missing its annotation (test with a fixture covering every category at least once).
- [ ] Tests added.

Issue #2480, opened 2026-08-25
[TASK] Frontend — `gap-mark.tsx` accessible title + KPI-strip wiring


norbert-kulus-blockydevs:
### Problem / Context

`apps/web/src/features/analytics/components/gap-mark.tsx` already exists (the `†` gap-mark symbol used elsewhere in the analytics surface); the mockup's cold audit found it missing a `title` attribute, meaning the symbol conveyed nothing to assistive tech or a hovering mouse. This phase both fixes that on the existing component (if not already fixed on `main` — check current state before assuming the bug is still present) and wires it to Phase 7's coverage data on `analytics-kpi-strip.tsx` / `analytics-kpi-card.tsx`.

### Proposed Solution

Ensure `GapMark` accepts and renders a `title`/`aria-label` describing which coverage category caused the gap (e.g. "2 orders excluded — currency needs recalculating"), sourced from Phase 7 Task 7.1's query, and that clicking it opens the matching `AnalyticsCoverageDetailModal` from Phase 7 Task 7.3.

### Classification

**Type**: Frontend
**Layer**: Feature component
**File(s)**: `apps/web/src/features/analytics/components/gap-mark.tsx`, `apps/web/src/features/analytics/components/analytics-kpi-strip.tsx`, `analytics-kpi-card.tsx`

### Dependencies

- Phase 7 Task 7.1, 7.3.

### Assumptions

- If `gap-mark.tsx`'s missing-`title` bug was already fixed on `main` independently of this epic, this task becomes wiring-only — verify against `main` before implementing, don't assume the earlier audit finding is still live.

### Acceptance Criteria

- [ ] Every KPI card affected by an open coverage category shows a `GapMark` with a real, category-specific `title`.
- [ ] Clicking it opens the correct detail modal for that category.
- [ ] Tests added/updated for `gap-mark.tsx` and the KPI components it's wired into.

Issue #2479, opened 2026-08-25
[EPIC] Phase 8 — Frontend: exclusion annotations on KPI/channel/product surfaces


norbert-kulus-blockydevs:
> Parent: #2452
> Depends on: Phase 4, Phase 7
> Mockup: `docs/plans/mockups/analytics-display-currency-picker.html` — `.excl-note` annotations inside the Channel and Top Products tables, and the KPI strip's gap-mark (`†`)

## Scope

The part of the design that makes exclusions *locally* visible, not just centrally listed in the Data Coverage panel: a small annotation on the specific KPI card, channel row, or product row that is under-counted because of an open coverage category, linking back to that category's detail modal. Kept as its own phase because it touches three existing, already-shipped components (`analytics-kpi-strip.tsx`/`analytics-kpi-card.tsx`, `channel-sales-table.tsx`, `product-sales-table.tsx`) rather than adding new ones — the review discipline here is "don't break what's already shipped," different from Phase 6/7's "build new."

## Branching

Branch `phase/2452-p8-fe-exclusion-annotations` off the epic branch.

## Dependencies

- Phase 4 (per-order/per-category affected data), Phase 7 Task 7.1 (the coverage query this phase's annotations key off of).
- Existing components: `apps/web/src/features/analytics/components/analytics-kpi-strip.tsx`, `analytics-kpi-card.tsx`, `channel-sales-table.tsx`, `product-sales-table.tsx`, `gap-mark.tsx`.

## Acceptance Criteria (mini-epic level)

- [ ] Every `.excl-note` annotation is factually consistent with which category actually excluded that row — regression guard for the mockup's own found bug (a candle row wrongly labeled a tax-rate issue when it was actually a currency issue).
- [ ] No row that belongs to an open coverage category is silently missing its annotation (regression guard for the mockup's own found bug — a tee-shirt row missing its excl-note despite being in the tax-A list).
- [ ] Clicking an annotation opens the exact same detail modal Phase 7 built for that category (no duplicate modal implementation).

Issue #2452, opened 2026-08-25
[EPIC] Analytics — display-currency picker + data-coverage remediation


norbert-kulus-blockydevs:
# [EPIC] Analytics — display-currency picker + data-coverage remediation

> Mockup: `docs/plans/mockups/analytics-display-currency-picker.html` (committed to repo per `docs/frontend-architecture.md § UX Mockups`).

## Problem / Context

Two real defects were found and hand-fixed on the live demo Postgres:

1. **Currency-era mismatch**: `order_records.reportingCurrency` can go stale relative to the current `reporting_currency_setting` (ADR-040 write-once stamp). Affected orders are silently excluded from revenue on `/analytics`, with no operator-visible signal that anything is wrong.
2. **Tax-rate-era exclusion**: `order_records.taxRateEra = 'pre-rollout'` permanently excludes an order from Net Sales (ADR-063), even after `TaxRateBackfillService` has resolved a valid rate. Again, silent — the KPI just reads low with no explanation.

Both were fixed manually via direct SQL / real service calls (`OrderFxStampService.stamp()`, nulling `taxRateEra`) during live diagnosis. That is not a repeatable fix — it required DB access, an operator cannot self-serve it, and there is no way to know it will happen again until revenue looks wrong.

**Neither of these is a bug** — both are ADR-040 and ADR-063 behaving exactly as designed. What's missing is operator visibility and a safe remediation path. In fact the two exclusion counts are **already computed** today: `unconvertedCount`/`unconvertedValue` (currency) and `netExcludedCount`/`netExcludedValue` (tax) exist right now on `SalesAnalyticsHeadline` (`libs/core/src/orders/domain/order-sales-aggregation.ts`) — they're just buried in a Revenue-card tooltip with no drill-down and no fix. This epic's core job is to elevate those existing numbers into an actionable panel, not to invent new detection from scratch.

This also surfaced a legitimate, requested feature: an operator running multi-currency sales (e.g. wallets in several currencies, or crypto-token totals) wants to **view `/analytics` in a currency other than the stamped reporting currency**, without touching the stamped `reportingCurrency`/`reportingTotalAmount` columns (ADR-040 restatement is explicitly out of scope — that is #2096).

## Proposed Solution

Two independent capabilities, designed together because they share one panel:

[...]

PR #2701, opened 2026-08-31
feat(analytics): exclusion annotations on KPI/channel surfaces (#2480, #2481, epic #2452 Phase 8)

jakubretajczykBD:
## Summary

Phase 8 of epic #2452 (mini-epic #2479) — implements Task 8.1 (#2480) and the channel half of Task 8.2 (#2481). Plan: `docs/plans/implementation-plan-analytics-exclusion-annotations.md`.

- **Task 8.1 (#2480)** — `GapMark` already had its `title`/`aria-label` (fixed independently on `main` before this epic reached it) — this was wiring-only. `AnalyticsKpiStrip`'s currency-gap `GapMark`s (Revenue card's GMV qualifier, Order value card's Average qualifier) and a new Net Sales tax-gap `GapMark` now source their title from the real, open Data Coverage category (Phase 7) and open the matching detail modal on click, instead of the generic pre-Phase-8 tooltip text. `GapMark` gains an optional `onActivate` prop — renders as a real `<button>` when supplied, byte-identical inert `<span>` otherwise.
- **Task 8.2, channel half (#2481)** — new `AnalyticsExclusionNote` renders one `.excl-note` per open coverage category a `ChannelSalesTable` row's channel has an order in. The cross-reference pages through each category's **full** affected-order list (never the Data Coverage aggregate's 10-id sample) and groups by `sourceConnectionId`, which is an exact match for a channel row.
- `AnalyticsDataCoveragePanel`'s open-category state becomes controllable-with-uncontrolled-fallback so the KPI strip / channel table can open its existing detail modals without a second dialog implementation. `AnalyticsPage` now owns one shared `GET /analytics/coverage` query + open-category state (same cache entry the panel already fetched — no extra request).

## Known gap (deliberately deferred, not fixed here)

**Task 8.2's product-table half (`product-sales-table.tsx`) is not implemented.** `TopProductRow` is keyed by `productId`; none of Phase 7's three coverage-order endpoints carry a `productId`/SKU/line-item field, so there is no correct client-side cross-reference from an excluded order to a product row. Approximating it (e.g. flagging every product on an affected channel) would reproduce the exact mislabeling bug class #2481's own ACs guard against. See the implementation plan §5 for the two remediation options flagged for a follow-up issue.

## Regression guards

- `channel-sales-table.test.tsx` — two new tests pin the mockup's own found bug class: a row's `.excl-note` must carry the category its own orders actually belong to (never a sibling row's), and a row with orders in every open cross-referenceable category gets a note for each one, never missing.
- `gap-mark.test.tsx` / `analytics-exclusion-note.test.tsx` — new, cover the inert/clickable `GapMark` variants and `AnalyticsExclusionNote`'s copy-from-`deriveCoverageRowCopy` + click behavior.
- Every existing Phase 7 test (`analytics-data-coverage-panel.test.tsx`, `analytics-kpi-strip.test.tsx`, `channel-sales-table.test.tsx`) still passes unmodified against the new optional props' uncontrolled-fallback defaults.

## Test plan

- [x] `pnpm --filter @openlinker/web lint` — 0 errors (pre-existing warning baseline unchanged)
- [x] `pnpm --filter @openlinker/web test` (analytics feature + page) — 190/190 passing

Part of #2480
Part of #2481

### DIFF
--- a/apps/web/src/features/analytics/components/analytics-data-coverage-panel.tsx
+++ b/apps/web/src/features/analytics/components/analytics-data-coverage-panel.tsx
@@ -66,6 +66,16 @@ interface AnalyticsDataCoveragePanelProps {
   filters: AnalyticsCoverageFilters;
   /** Opens the shared Analytics Settings dialog — used by the tax-A row's "Include anyway" action. */
   onOpenSettings: () => void;
+  /**
+   * Controllable-with-uncontrolled-fallback (#2480, epic #2452 Phase 8):
+   * when supplied together, an external trigger (the KPI strip's `GapMark`,
+   * the channel table's `AnalyticsExclusionNote`) can open this panel's own
+   * detail modals without a second `CoverageDetailDialog` implementation.
+   * Omitted (the panel's own self-render, unchanged since Phase 7): falls
+   * back to internal state, byte-identical to pre-Phase-8 behavior.
+   */
+  openCategory?: CoverageCategory | null;
+  onOpenCategoryChange?: (category: CoverageCategory | null) => void;
 }
 
 function isTaxCategory(category: CoverageCategory): category is TaxCoverageCategory {
@@ -75,6 +85,8 @@ function isTaxCategory(category: CoverageCategory): category is TaxCoverageCateg
 export function AnalyticsDataCoveragePanel({
   filters,
   onOpenSettings,
+  openCategory: controlledOpenCategory,
+  onOpenCategoryChange,
 }: AnalyticsDataCoveragePanelProps): ReactElement {
   const { showToast } = useToast();
   const demoMode = useDemoMode();
@@ -88,7 +100,9 @@ export function AnalyticsDataCoveragePanel({
     return (connectionId: string): string => byId.get(connectionId) ?? connectionId;
   }, [connectionsQuery.data]);
 
-  const [openCategory, setOpenCategory] = useState<CoverageCategory | null>(null);
+  const [internalOpenCategory, setInternalOpenCategory] = useState<CoverageCategory | null>(null);
+  const openCategory = onOpenCategoryChange ? controlledOpenCategory ?? null : internalOpenCategory;
+  const setOpenCategory = onOpenCategoryChange ?? setInternalOpenCategory;
   const [offset, setOffset] = useState(0);
 
   function openDetail(category: CoverageCategory): void {
--- a/apps/web/src/features/analytics/components/analytics-exclusion-note.test.tsx
+++ b/apps/web/src/features/analytics/components/analytics-exclusion-note.test.tsx
@@ -1,0 +1,35 @@
+import { render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { describe, expect, it, vi } from 'vitest';
+import { AnalyticsExclusionNote } from './analytics-exclusion-note';
+import { deriveCoverageRowCopy } from '../lib/data-coverage-copy.lib';
+import { COVERAGE_CATEGORY_VALUES, type CoverageCategory } from '../api/analytics-coverage.types';
+
+describe('AnalyticsExclusionNote (#2481)', () => {
+  it.each(COVERAGE_CATEGORY_VALUES)('renders the %s category copy from deriveCoverageRowCopy verbatim', (category) => {
+    render(
+      <AnalyticsExclusionNote category={category as CoverageCategory} affectedCount={3} onOpenCategory={() => {}} />
+    );
+
+    const expected = deriveCoverageRowCopy({
+      category: category as CoverageCategory,
+      status: 'open',
+      affectedCount: 3,
+      sampleOrderIds: [],
+    }).headline;
+
+    expect(screen.getByRole('button', { name: expected })).toBeInTheDocument();
+  });
+
+  it('calls onOpenCategory with its own category when clicked', async () => {
+    const user = userEvent.setup();
+    const onOpenCategory = vi.fn();
+
+    render(<AnalyticsExclusionNote category="tax-b" affectedCount={1} onOpenCategory={onOpenCategory} />);
+
+    await user.click(screen.getByRole('button'));
+
+    expect(onOpenCategory).toHaveBeenCalledWith('tax-b');
+    expect(onOpenCategory).toHaveBeenCalledTimes(1);
+  });
+});
--- a/apps/web/src/features/analytics/components/analytics-exclusion-note.tsx
+++ b/apps/web/src/features/analytics/components/analytics-exclusion-note.tsx
@@ -15,9 +15,20 @@
  * the string here instead of deriving it would reopen exactly that risk
  * the moment either copy changes.
  *
+ * Renders on top of the shared `Chip` primitive (`shared/ui/chip.tsx`)
+ * rather than a one-off styled `<button>` — `tone="warning"` already gives
+ * the exact pill treatment this note needs
+ * (`docs/frontend-ui-style-guide.md` § CSS Implementation Standard: "add
+ * or extend shared primitives before introducing page-specific one-off
+ * styling"). `Chip`'s `aria-pressed` toggle semantics don't fit a
+ * fire-and-navigate action, so it's explicitly cleared here; the `excl-note`
+ * class only ADDS the truncation behavior a table cell needs on top of
+ * `Chip`'s own tone/spacing/border, never re-declares them.
+ *
  * @module apps/web/src/features/analytics/components
  */
 import type { ReactElement } from 'react';
+import { Chip } from '../../../shared/ui/chip';
 import { deriveCoverageRowCopy } from '../lib/data-coverage-copy.lib';
 import type { CoverageCategory } from '../api/analytics-coverage.types';
 
@@ -36,8 +47,13 @@ export function AnalyticsExclusionNote({
   const copy = deriveCoverageRowCopy({ category, status: 'open', affectedCount, sampleOrderIds: [] });
 
   return (
-    <button type="button" className="excl-note" onClick={() => onOpenCategory(category)}>
+    <Chip
+      tone="warning"
+      className="excl-note"
+      aria-pressed={undefined}
+      onClick={() => onOpenCategory(category)}
+    >
       {copy.headline}
-    </button>
+    </Chip>
   );
 }
--- a/apps/web/src/features/analytics/components/analytics-exclusion-note.tsx
+++ b/apps/web/src/features/analytics/components/analytics-exclusion-note.tsx
@@ -1,0 +1,43 @@
+/**
+ * Analytics Exclusion Note
+ *
+ * The `.excl-note` per-row annotation (#2481, epic #2452 Phase 8) shown on
+ * a `ChannelSalesTable` row whose own figures are under-counted by an open
+ * Data Coverage category (#2474 Phase 7). One instance per affected
+ * category on a row — a row belonging to more than one open category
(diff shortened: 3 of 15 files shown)


jakubretajczykBD:
## Tech review — PR #2701

**Summary**: Solid, well-scoped frontend-only PR that wires Phase 7's Data Coverage aggregate into Phase 8's KPI/channel annotations. The controllable-with-uncontrolled-fallback pattern on `AnalyticsDataCoveragePanel`, the copy reuse via `deriveCoverageRowCopy`, and the explicit, well-justified punt on the product-table half are all good calls. Test coverage directly targets the two bug classes the plan calls out (mislabeled category, missing annotation). No architecture-boundary, security, or correctness violations found.

### Issues

**[IMPORTANT]** — `analytics-exclusion-note.tsx`
`.excl-note` hand-rolled a new warning-tone pill from scratch (border/background/color pulled straight from `--status-warning-*`) instead of reusing the `Chip` primitive (`shared/ui/chip.tsx`), which already ships a `tone="warning"` variant and was already imported in `channel-sales-table.tsx` for `ChannelFlags`. `docs/frontend-ui-style-guide.md` § CSS Implementation Standard: "add or extend shared primitives before introducing page-specific one-off styling."
→ **Fixed**: `AnalyticsExclusionNote` now renders `<Chip tone="warning" className="excl-note" aria-pressed={undefined} …>`; `.excl-note`'s CSS shrinks to only the truncation add-on (`max-width`/`overflow`/`text-overflow`) on top of what `.chip`/`.chip--warning` already provide. `aria-pressed` is explicitly cleared since this is a fire-and-navigate action, not a toggle.

**[IMPORTANT]** — `channel-sales-table.tsx`
`useCoverageCrossReferenceQuery` (a `useQuery`-wrapping hook) was defined inline inside a component file. `docs/frontend-architecture.md` § Server State: "query hooks live under `features/<domain>/hooks`."
→ **Fixed**: extracted to `hooks/use-coverage-cross-reference-query.ts`. `buildChannelExclusionMap` + its supporting types are pure derivations, not hooks, so they moved to `lib/channel-exclusion-map.lib.ts` (matching the existing `*.lib.ts` convention, e.g. `data-coverage-copy.lib.ts`) rather than the hooks folder.

**[SUGGESTION]** — `channel-sales-table.tsx`
`openCategoryCodes` (`new Set(...)`) was recomputed from `coverage.categories` on every render. Low-cardinality so not a real perf issue, but memoized (`useMemo`) to match the pattern elsewhere in the feature.

**[SUGGESTION, not applied]** — `channel-sales-table.tsx` / `use-coverage-cross-reference-query.ts`
The cross-reference queries fully drain every page of an open category's affected-order list (currency + up to 3 tax categories) purely to build a small per-`sourceConnectionId` count map, independent of `AnalyticsDataCoveragePanel`'s own paginated reads of the same categories. The implementation plan already flags this as an accepted risk, so left as-is — worth a follow-up if catalogs grow (e.g. a per-`sourceConnectionId` count endpoint).

**[SUGGESTION, not applied]** — `use-coverage-cross-reference-query.ts`
No error handling on the cross-reference `useQuery` — a failed page fetch degrades silently to "no exclusion notes shown" for that channel. Defensible for an annotation layer over already-rendered rows, but worth a one-line doc-comment acknowledgment that this fails closed rather than fails loud.

### Verification
- `tsc --noEmit` — clean
- `vitest run` on the 5 touched analytics test files — 37/37 passing, unmodified (both IMPORTANT fixes were pure refactors, no test-visible behavior change)
- `eslint` on all touched/added files — clean (one pre-existing, unrelated warning in `channel-sales-table.tsx` confirmed via `git stash`)

### Verdict
✅ **Approve** — both IMPORTANT items fixed in this pass; remaining items are documented tradeoffs, not blockers.

jakubretajczykBD:
## Tech review (via `/tech-review`)

### Summary

Well-scoped, deliberately narrow FE-only change that reuses existing primitives (query keys, copy source, dialog) instead of duplicating them, and is explicit about what it does *not* solve (the Product-table half is honestly deferred with a documented backend gap, rather than approximated into a plausible-looking but wrong annotation). The main concern is the channel table's per-category "drain every page" cross-reference query, which runs unconditionally at page load with no visible bound — worth a second look before this scales past a handful of affected orders per category.

### Issues

**[IMPORTANT]** — `apps/web/src/features/analytics/components/channel-sales-table.tsx` (`useCoverageCrossReferenceQuery`)
> The hook loops `while(true)` pulling every page (`CROSS_REF_PAGE_LIMIT = 100`) of a category's *entire* affected-order set purely to build a per-connection presence/count map, and this fires for up to 4 categories unconditionally whenever the analytics page loads with any open category. There's no upper bound on total pages drained. On an install with thousands of affected orders in one category (plausible — `netExcludedCount`/`unconvertedCount` are cumulative, unbounded counts), this is a multi-request, potentially slow background fetch on every page view, duplicating data the Data Coverage detail modal already fetches (under a different, unrelated query key) on demand. The PR description acknowledges this as a known risk in the plan doc, but the mitigation (a backend aggregate-by-connection endpoint, or a page cap) isn't in this PR. Recommend either capping total pages drained with a documented truncation behavior, or filing the backend follow-up now rather than leaving it purely as a plan-doc note.

**[SUGGESTION]** — `apps/web/src/features/analytics/components/channel-sales-table.tsx`
> `useCoverageCrossReferenceQuery`, `buildChannelExclusionMap`, and the `ChannelExclusionMap`/`CoverageOrderLite` types are defined inline in the component file. `docs/frontend-architecture.md` / `.claude/rules/fe-pages.md` place query hooks under `features/<domain>/hooks/`. Given this hook is private and single-consumer, inlining it isn't wrong, but there are stray untracked files in the worktree (`hooks/use-coverage-cross-reference-query.ts`, `lib/channel-exclusion-map.lib.ts`) suggesting an earlier draft did extract them — worth reconciling so the working tree doesn't carry dead scaffolding unrelated to what was pushed.

**[SUGGESTION]** — `apps/web/src/features/analytics/components/channel-sales-table.tsx` / `analytics-kpi-strip.tsx`
> The cross-reference queries expose no `isLoading`/`isError` — a failed fetch silently renders no exclusion note rather than surfacing a degraded state. For a purely supplementary annotation (the table's core figures are unaffected) this is a reasonable choice, but it's implicit; a one-line comment noting "failure degrades silently by design" would save a future reader from treating it as an oversight.

**[SUGGESTION]** — `apps/web/src/index.css` (`.excl-note`)
> `padding: 1px 6px;` uses raw px values rather than the 4px spacing-token scale (`docs/frontend-ui-style-guide.md` § Spacing And Shape: "avoid raw rem values where a token exists"). Minor, but the rest of the new CSS in this PR correctly uses tokens (`var(--radius-pill)`, `var(--status-warning-*)`), so this one spot stands out.

**[SUGGESTION]** — `apps/web/src/features/analytics/components/analytics-kpi-strip.tsx` (`resolveNetExcludedTaxCategory`)
> No dedicated unit test for this helper's tie-break/fallback logic (picks the tax category with the largest `affectedCount`, falls back to `tax-a` if all are zero). It's exercised indirectly through component tests per the PR description, but a small `.spec` for the pure function would pin the "largest wins, `tax-a` on all-zero" behavior more legibly than reading it out of a component render.

Everything else checked out cleanly:
- Controllable-with-uncontrolled-fallback on `AnalyticsDataCoveragePanel` is the right shape and default-preserves existing self-render behavior.
- `GapMark`'s span→button branch correctly drops `role="img"` and moves to `aria-label`, matching the a11y rule that a real control shouldn't be mis-described as a static image.
- `AnalyticsExclusionNote` correctly sources copy from `deriveCoverageRowCopy` (single source, no re-paraphrasing) — confirmed the function never reads `sampleOrderIds`, so passing `[]` there is safe, not a latent bug.
- No cross-feature deep imports; all imports go through public barrels or the sanctioned `app/api/api-client-provider` DI seam.
- No new global store; state lift is a page-level `useState`.
- Query-key reuse (`useAnalyticsCoverageQuery` at the page level, same cache entry as the panel) avoids a duplicate `GET /analytics/coverage`.
- The product-table deferral is a genuinely correct call: the coverage DTOs have no `productId`/line-item field, so any FE cross-reference there would be an approximation reproducing the exact mislabeling bug class the regression tests guard against.
- Regression tests for the "wrong category attributed to a row" and "missing note" bug classes target the real cross-reference logic, not just rendering.

### Verdict

🔄 **Approve with changes** — no BLOCKING architecture/security issues; the unbounded per-category page-drain (IMPORTANT) is worth resolving before this ships to an install with large affected-order counts.

jakubretajczykBD:
Follow-up on the IMPORTANT finding above (unbounded per-category page-drain in `useCoverageCrossReferenceQuery`): filed as two tracked tasks rather than fixed inline here, since it needs a new backend read.

- #2713 — backend aggregate-by-connection endpoint (reuses `TaxCoverageDetectionService`'s existing in-memory `classify()` pass + a `GROUP BY sourceConnectionId` on the currency/product-matching SQL reads — no new expensive computation, no schema change)
- #2714 — swap `useCoverageCrossReferenceQuery`'s drain-all-pages loop to consume #2713 once it ships (blocked by #2713)

The other SUGGESTION-level findings (hook placement, silent-degradation comment, `.excl-note` padding token, `resolveNetExcludedTaxCategory` unit test) are being addressed directly in this PR.

jakubretajczykBD:
## Follow-up tech review (post-fix commit `2e973ec3b`)

### Summary

All findings from the earlier review are addressed: the 4 SUGGESTIONs fixed directly in code (hook extracted to `hooks/`, `.excl-note` moved onto the shared `Chip` primitive, silent-degradation behavior documented, `resolveNetExcludedTaxCategory` now has dedicated test coverage), and the IMPORTANT unbounded-page-drain finding is tracked as #2713 (backend aggregate-by-connection endpoint) + #2714 (swap this hook to consume it), linked in code and in the PR follow-up comment. Type-check clean, lint 0 errors, analytics tests 193/193 (up from 190).

### Issues

> The top-level PR body's `Test plan` / `Regression guards` sections don't mention the third commit's additions (coverage-aware `GapMark` tests in `analytics-kpi-strip.test.tsx`, the `.excl-note` → `Chip` primitive swap). Not blocking — the third commit's own message is complete — but worth syncing the PR description for anyone skimming only the top.

**[SUGGESTION]** — `apps/web/src/features/analytics/components/analytics-exclusion-note.tsx:53`
> `aria-pressed={undefined}` correctly suppresses `Chip`'s default `aria-pressed={active}` today (because `Chip`'s internal `{...props}` spread comes after `aria-pressed={active}` in its JSX), but that ordering isn't part of `ChipProps`'s documented contract — a future reorder inside `Chip` would silently reintroduce `aria-pressed` on this fire-and-navigate button with no type error. Consider adding `expect(button).not.toHaveAttribute('aria-pressed')` to `analytics-exclusion-note.test.tsx` to pin this.

Everything else checks out cleanly:
- FE-only, zero backend changes; `app → pages → features → shared` respected; cross-feature imports go through public barrels.
- Hook/lib extraction matches "query hooks live under features/<domain>/hooks."
- `.excl-note` reuse of `Chip` (`tone="warning"`) is exactly the style guide's "extend shared primitives before introducing page-specific one-off styling," and incidentally resolves the earlier raw-px-padding concern.
- New KPI-strip tests correctly cover the two-simultaneous-`GapMark` case (Revenue/Order value), the tax-a/b/c tie-break, and the no-GapMark fallback when `netExcludedCount` disagrees with a zeroed coverage read.
- The unbounded-drain doc comment names #2713/#2714 explicitly, closing the gap from the prior review.

### Verdict

✅ **Approve** — ready to merge. Remaining two items are cosmetic, non-blocking.

Issue #2684, opened 2026-08-30
The migration chain cannot run against a genuinely empty database (`uuid_generate_v4()` with no extension)


piotrswierzy:
Found during **#2392**, by the first spec in the tree that runs migrations from empty. Documented and worked around there rather than fixed, because rewriting the oldest migration for a defect that slice did not introduce is out of its scope.

## The defect

Migration `1766246163229` — the oldest in the tree — uses `uuid_generate_v4()` without creating the `uuid-ossp` extension. Against a genuinely empty database the chain therefore fails at the first migration.

## Why it has been invisible

**Nothing ever ran migrations from empty.** The integration harness builds schema by TypeORM `synchronize`, and every developer and deployed environment has a database whose extension was created at some point in the past — so the missing `CREATE EXTENSION` has never been exercised.

That is the same root cause as the finding that surfaced it (#2392's B1): a migration nothing runs is a migration nothing verifies.

## Consequences

- A new contributor provisioning a database from scratch hits it immediately.
- A disaster-recovery restore-and-replay path would hit it.
- Any future spec that runs the chain from empty must work around it, as #2392's parity spec does today (it names the defect at the point it works around it).

## Scope

1. Add the extension creation to the chain. **Do not edit `1766246163229` in place** if any deployed database has already applied it — TypeORM records applied migrations by name, so an edited body never re-runs and the fix would reach exactly nobody. Prefer a new migration that sorts *before* the rest cannot help either, since ordering is by timestamp and that slot is taken. The realistic options are a documented pre-migration bootstrap step, or `CREATE EXTENSION IF NOT EXISTS` inside a new migration accompanied by the note that it does not repair an already-broken from-empty run for existing installs.
2. Whichever is chosen, state the reasoning in the migration or in `docs/migrations.md` — the constraint (an applied migration never re-runs) is the non-obvious part.
3. **Verify from empty**: run the full chain against a throwaway Postgres with no extensions, and confirm exit 0 and `migration:show` 0 pending. That is the check that has never existed.

## Acceptance criteria

- [ ] The full migration chain runs clean against a genuinely empty Postgres, demonstrated.
- [ ] The workaround in #2392's `fulfillment-work-migration-parity.int-spec.ts` is removed, and its pointer to this issue with it.
- [ ] The approach's limits for already-provisioned databases are stated where the next reader meets them.

## Related

- **#2392** — surfaced it; its parity spec is the first thing in the tree to run migrations from empty.
[...]

Issue #2609, opened 2026-08-27
[BUG] CORE/Worker — inventory propagation is serialised across the whole installation


norbert-kulus-blockydevs:
## Problem / Context

Measured on the demo stack: the queue grows about **145 jobs/h faster than it drains**, and the 15 066 backlogged jobs found there were several days of ordinary operation rather than an incident.

Two causes, both in our code and neither related to PrestaShop:

- `inventory.propagateToMarketplaces` is registered in the **`fan-out` lane** (`apps/worker/src/sync/handlers/handler-registration.service.ts:240`), whose per-process cap and per-scope cap are both **1** (`apps/worker/src/sync/sync-job.runner.ts:126-142`)
- every enqueue uses the same synthetic scope: `SYSTEM_CONNECTION_ID = '00000000-0000-0000-0000-000000000000'` (`libs/core/src/inventory/application/services/inventory.service.ts:26`, used at `:68`)

Together those serialise **every stock propagation in the entire installation** through one slot, whatever the number of connections.

## Proposed Solution

Give propagation a real scope (the destination connection, so per-scope accounting means something) and move it out of the single-slot lane. Choose the lane by cost of starvation per ADR-050, not by convenience.

## Classification

**Type**: CORE / Worker
**Layer**: Application
**File(s)**: `libs/core/src/inventory/application/services/inventory.service.ts`, `apps/worker/src/sync/handlers/handler-registration.service.ts`, `libs/core/src/sync/domain/types/sync-job-lane.types.ts`

## Dependencies

- Ships together with the out-of-order quantity-write guard — this change multiplies the number of concurrent quantity writes

## Acceptance Criteria

- [ ] Propagation jobs carry the destination connection as their scope, not a synthetic zero id
- [ ] Two connections propagate concurrently
- [ ] Measured drain rate exceeds the measured arrival rate on the demo catalogue
- [ ] The lane choice is justified against ADR-050 and the ADR amended or confirmed
- [ ] Tests added or updated for non-trivial logic

norbert-kulus-blockydevs:
@/tmp/claude-1006/-home-norbert-kulus-projekty-openlinker/ddf1c071-56a7-4b24-8d59-cf8a60bf7914/scratchpad/r2609.md

norbert-kulus-blockydevs:
Done in the epic branch `epic/2590-prestashop-no-module`, reviewed and fixed on PR #2627. Closing so the board reflects it; the branch is merged to `main` as one PR.

PR #2706, opened 2026-08-31
feat(fulfillment): executor handshake, assignmentAttempt and fulfillment.work.dispatch (#2399)

### LINKED ISSUE #2399: [TASK] CORE Application — executor handshake + `assignmentAttempt` + `fulfillment.work.dispatch`
> Part of the OMS programme backlog (**Wave 3a**, stream S1, size M). Backlog slug: `W3a-10`. Sibling `W3a-*` slugs resolve via the Wave-3a epic's checklist.

## Problem / Context

REVIEW C7: `work:{workId}:{attempt}` re-mints on job retry if `attempt` is the job-runner attempt — which changes on exactly the retries the key must survive. A re-minted key means a duplicate fulfilment request to a 3PL, i.e. a double-ship.

## Proposed Solution

`FulfillmentRequest.idempotencyKey` is mandatory and caller-minted as `work:{workId}:{assignmentAttempt}`, where `assignmentAttempt` is the **persisted, monotonic counter on the work row (#2392, `W3a-3`), incremented only by a router-driven re-request and written before the outbound call** (the Amazon MCF `sellerFulfillmentOrderId` model). The handshake service drives the negotiation axis — `unsubmitted → submitted → accepted | rejected` and `cancellation_requested → cancellation_accepted | cancellation_rejected` — each a conditional UPDATE. Core-owned job type `fulfillment.work.dispatch`. A terminal rejection resolves as `outcome: 'business_failure'` per ADR-007, never a retried permanent condition.

## Classification

**Type**: CORE
**Layer**: Application
**File(s)**: `libs/core/src/fulfillment/application/services/fulfillment-handshake.service.ts`, `apps/worker/src/fulfillment/`

## Dependencies

- #2392 (`W3a-3`), #2398 (`W3a-9`)

## Acceptance Criteria

- [ ] A job retry re-mints an **identical** key (spec)
- [ ] The `assignmentAttempt` write is asserted to precede the outbound call (ordering spec)
- [ ] A `rejected{blocking}` result records the exclusion durably
- [ ] Terminal rejection maps to `business_failure`

piotrswierzy:
Drives the **negotiation axis** of a `FulfillmentWork`: offer the work to its assigned holder through `FulfillmentExecutorPort`, record the answer, and do it under an idempotency key that survives a job retry.

Wave 3a (epic #2412), stream S1. Direct parent #2398.

## The defect this exists to prevent

REVIEW C7 named it. `work:{workId}:{assignmentAttempt}` re-mints on a job retry if the attempt is the **job-runner** attempt — which changes on exactly the retries the key must survive. A re-minted key is a second fulfilment request to a 3PL, i.e. a **double-ship**. So `assignmentAttempt` is a persisted, monotonic column bumped only by a router-driven re-request, and a retry **resumes** the claimed attempt rather than taking a new one.

## How the attempt claim is enforced

One conditional UPDATE, reusing #2392's `applyGuardedUpdate` idiom:

```sql
UPDATE fulfillment_works
   SET "requestStatus" = 'submitted',
       "assignmentAttempt" = "assignmentAttempt" + 1,
       "version" = "version" + 1
 WHERE "id" = $1 AND "requestStatus" IN ('unsubmitted','rejected')
RETURNING "assignmentAttempt";
```

**Not** an index and **not** a lock, and the reasoning is the coordinator's own rule read carefully: a unique index or a parent-row lock is what enforces a bound on *inserted rows*, and an attempt is a **column**, not a row. `x = x + 1` in a single UPDATE is atomic at row level, so two concurrent claimants yield 1 and 2 — never both 1. The live-state uniqueness that does matter here is the `requestStatus` guard, which is state-based; at READ COMMITTED the loser blocks on the row lock, re-evaluates its WHERE against the committed row and matches zero. Partial uniqueness on live states **is** used where rows are inserted — `UQ_fulfillment_work_rejections_work_attempt`.

The attempt reaches the caller **only** through that statement's `RETURNING`, so minting a key without the row already holding the value is not expressible — structurally stronger than asserting call order in a spec.

`applyGuardedUpdate` returns `boolean` and has seven callers whose docblock exists so the `?? 0` cannot be forgotten per transition, so a **sibling** `applyGuardedUpdateReturning` was added rather than widening that choke point.

### Red-first evidence

| Mutation | Result |
|---|---|
| Drop the `requestStatus` guard from the claim | int-spec `exactly ONE of two overlapping claims win` fails: **`Expected length: 1, Received length: 2, Received array: [1, 2]`** — two claims, two distinct keys, the double-ship |
| Resume returns `attempt + 1` | unit `re-mint an IDENTICAL key when a job retry resumes` fails |
| Widen `CLAIMABLE_FROM` with `accepted` | unit `claim only from unsubmitted or rejected` fails |
| Send under the pre-claim `assignedConnectionId` | unit `offer a RESUMED dispatch to the holder the row names NOW` fails |
| Read the PII flag via `getPiiConfig()` | 6 handler tests fail (throws with no `OL_PII_HASH_SALT`) |

The concurrency test is deliberately **overlapping, not sequential**: a sequential test passes against an implementation with no guard at all, so it cannot distinguish the defect from its absence.

## Lane: `realtime`

ADR-050 picks by **cost of starvation**. A dispatch is the outbound "tell the holder to ship" for a just-routed order — someone is waiting and lateness costs a shipment. Same class as `marketplace.order.sync`, emphatically not a paced sweep. #2594's split-by-trigger has nothing to separate here (one trigger, one cost), so the type stays single. `SyncJob.connectionId` is the work's own `assignedConnectionId`, never a synthetic id — #2609's defect is exactly a shared scope collapsing per-scope lane accounting installation-wide.

Tripwire: realtime **13 → 14**, job types **50 → 51**, dummy handlers **48 → 49**.

## Other load-bearing decisions

- **Acceptance is an at-most-once claim** — `WHERE "requestStatus" = 'submitted' AND "acceptedAt" IS NULL`. The second conjunct still holds if a future writer moves `requestStatus` without coming through `recordAcceptance`.
- **Rejections are an append-only TABLE**, not the two columns #2392 deferred. `blocking` exists so re-sourcing can *exclude* the rejecter, and an exclusion is a **set**: A refuses, the router tries B, B refuses too. A scalar pair holds only the last refusal, so A's exclusion is lost and the loop the field exists to terminate runs anyway. Guarded transition + INSERT in **one transaction**; a lost guard inserts zero rows.
- **The cancellation key gets its own namespace** (`:cancel`). Sharing the dispatch key would have the executor answer the cancellation with the dispatch's cached `accepted` — the replay guarantee working as specified — and OL would record `cancellation_accepted` for a cancellation the holder never saw.
- **ADR-053 honoured with no guard weakened**: the executor and the ship-to arrive as **arguments**, resolved in `apps/worker`. `deliveryMethod` is deliberately not an argument — it is an insert-only column the router owns, so a second source could disagree with the row.

## `externalWorkId`'s shape (deferred here from #2400)

**Intrinsic to the row**, not a per-connection side table: acceptance is at-most-once, so exactly one holder has accepted a given work. But the **lookup** is per-connection, so the index is composite and partial — `(assignedConnectionId, externalWorkId) WHERE "externalWorkId" IS NOT NULL`. The value is a third party's; two holders may both mint `"1"`, and an unscoped lookup would correlate one holder's webhook onto another's work. **Non-unique**, because uniqueness is not OL's to assert on a vendor's behalf and a unique index would refuse a legitimate acceptance.

## Two real defects found and fixed

1. The handler read the PII flag with `getPiiConfig()`, which throws whenever `OL_PII_HASH_SALT` is unset **regardless of the flag** — `routing-ship-to.types.ts` warns against exactly this. Every ordinary deployment would have failed dispatch.
2. It read a `shipToGeography` field that does not exist on `OrderRecord`.

Both pinned by tests that fail against the old behaviour.

## Migration

`1866000000000-add-fulfillment-handshake.ts` — moved from `1865000000000`, which **#2394 claims concurrently on its own branch**. Two siblings with one timestamp on one integration branch is a real conflict. Verified from empty against a throwaway Postgres: 150 applied, 0 pending, this migration the tail, down/up round-trip clean, the partial index present in `pg_indexes`. It also creates the `uuid-ossp` extension, which #2684 records as missing on `1766246163229`.

## Reachable vs declared-only

The handshake, counter, schema and job are complete and test-exercised. `FulfillmentExecutor` **is** in `CoreCapabilityValues` (#2403) but **no shipped manifest advertises it**, so A3 is assignable only via a direct `PATCH /connections/:id` — a reachability gap, not data loss.

## Deferred, with owners

| Deferred | Owner |
|---|---|
| The re-source loop consuming the exclusion | #2395 (constraint recorded: if re-sourcing mints a new work row it must carry blocking exclusions forward — the `orderId` lineage is already on the row) |
| A cancellation **job type** | no caller yet; costs a lane registration + boot-coverage entry |
| **ADR-054's timeout-as-rejection sweep** | **needs an issue** — `IDX_fulfillment_works_request_status` says #2399 owns it. Stated rather than hidden: a `submitted` work whose holder never answers stays `submitted` and nothing reaps it. It is never silently retried into the ground (the resume path is attempt-scoped and terminal-safe). |

## Gates

`pnpm lint` **0** (35 invariant checks) · `pnpm type-check` **0** · `pnpm test` **0** · migration `0` pending.

Integration: the four fulfillment specs pass (4 suites, 49 tests). The full suite's remaining reds are the named pre-existing pair — **#2638** (`earliest-order-date`, verbatim: February seed, non-UTC host, `T00:` vs `T01:`) and **#2639** (`allegro-prestashop-carrier-mapping`), plus two PrestaShop-container siblings in that family — never chased. One genuine red **was** mine and is fixed: `fulfillment-work-schema.int-spec` pins the `fulfillment_works` column list with `toEqual`, a *second* exhaustive list beside the parity spec's.

Coordination: additions to `fulfillment.tokens.ts` (one token) and the context `index.ts` (one export block) are deliberately minimal — five siblings share both files.

Closes #2399
Refs #2412

https://claude.ai/code/session_01EWKtLwTcP2ppoknaq4Ca3Z


piotrswierzy:
## Merged `oms-programme-wave-3a` (branch was 14 commits behind)

#2400 and #2397 landed while this was in review. Five files conflicted; **every resolution is a union** — the two fulfilment slices are opposite directions of one negotiation and neither replaces the other.

Two things the merge **confirmed rather than assumed**:

1. **The migration renumber was necessary, not merely prudent.** #2400 took `1865000000000` (`create-fulfillment-progress-claims`). This branch had already moved to `1866000000000` pre-merge on the strength of #2394's *unmerged* claim, so the collision never happened and `check-migration-timestamps` passes.
2. **#2400's deferral to this slice is now resolved in the direction it anticipated.** Its `FulfillmentWorkStatusSyncPayloadV1.externalWorkId` docblock states that #2399 owns the writer and therefore the choice between a `fulfillment_works.externalWorkId` column and `identifier_mappings` resolution, and that guessing "would hand #2399 a schema it must migrate away from". This branch chose the column, with the composite partial index `(assignedConnectionId, externalWorkId)` the correlation read needs.

### Resolutions

| File | Resolution |
|---|---|
| `fulfillment.tokens.ts` / `fulfillment.module.ts` | Both services bind, both entities register, both tokens export |
| `fulfillment-job-payloads.types.ts` | Both payload interfaces; one widened header saying they are opposite directions and share nothing else |
| `fulfillment-work-migration-parity.int-spec.ts` | `fulfillment_works` now has **four** children — the CASCADE smoke test inserts and counts all four. The table list and FK list git merged correctly on its own and were **verified rather than assumed** |
| `handler-registration.service.spec.ts` | Both sides independently claimed *"14 realtime"*. With both registered: **15 realtime / 25 / 5 / 7 across 52 job types**, 50 dummy handlers |

One correction the merge itself would not have surfaced: the ADR-050 tally comment **inside** `handler-registration.service.ts` still read `14 realtime / … / 51 job types`. It did not conflict, so nothing but reading the file would have caught it.

### Gates on the merged tree

| Gate | Exit |
|---|---|
| `pnpm lint` (35 invariant checks) | **0** |
| `pnpm type-check` | **0** |
| `pnpm test` | **0** — `apps/worker` 687 |
| `pnpm test:integration` | **`INT_EXIT=1`** — `Test Suites: 1 failed, 152 passed, 153 total` · `Tests: 1 failed, 1 skipped, 1271 passed, 1273 total` |
| `apps/worker` integration (explicit — root is api-only, #2670) | **1** — `Test Suites: 1 failed, 25 passed, 26 total` |

Both reds are pre-existing and **named, not chased**:

- `earliest-order-date.int-spec` — **#2638**, verbatim: February seed on a non-UTC host, `T00:00:00.000Z` vs `T01:00:00.000Z`.
- `automation-dispatch-boot.int-spec` — `AUTOMATION_DISPATCH_SERVICE_TOKEN` resolves to `AutomationIrreversibleGateService`, but #2361's boot spec still asserts `AutomationDispatchService`. That is #2362's gate composition versus #2361's stale assertion, entirely inside `automation`, **zero overlap with this diff** (`git diff --name-only <merge-base>..HEAD | grep -i automation` → 0). **Worth its own issue** — it is a hard-gate spec currently red on the wave branch.

The PrestaShop specs (#2639's family) passed on both merged runs.

`fulfillment-no-injection-boot.int-spec` (the ADR-053 assertion) and `oms-module-boot` both **PASS** — the new `SyncWorkerModule → FulfillmentModule` edge boots cleanly and no injection contract is violated.

### One process note worth carrying forward

The api-only root `test:integration` **cannot** catch a worker DI-graph break. Running the worker suite explicitly is what surfaced that `inventory-location-propagation-e2e.int-spec` (a #2324 spec, not mine) is missing the `process.env.OL_PII_HASH_SALT ??= …` line its 5 sibling boot specs all carry — so the whole worker suite aborts at `NestFactory.createApplicationContext` for anyone whose environment does not export it. Cheap to fix (one line), and also worth its own issue.

piotrswierzy:
## Migration renumbered `1866` → `1867`; wave branch re-merged

#2394 landed at `131486f79` having itself renumbered `1865 → 1866` to dodge #2400 — onto the exact slot this branch had taken one merge earlier for the same reason. Two `1866000000000-*` files were briefly present in the tree, so the collision was real rather than latent.

Renamed **together** per rule 2 (a half-rename is its own violation): filename prefix, exported class suffix, `name` property. Swept for stale references to both the old number and the old class name — **none**. The migration docblock now records the whole chain, so the next slice reads why `1867` rather than rediscovering it.

```
migration-timestamps: self-check OK
migration-timestamps: OK (152 migrations (incl. 1 plugin dir); ordering vs origin/main: checked)
```

### Chain re-run from empty (throwaway Postgres, port 55432, container removed)

**152 applied, 0 pending**, with `create-routing-decisions` ahead of mine:

```
[X] 148 AddReturnLineEventsTimelineIndex1863000000000
[X] 149 CreateFulfillmentWorks1864000000000
[X] 150 CreateFulfillmentProgressClaims1865000000000
[X] 151 CreateRoutingDecisions1866000000000
[X] 152 AddFulfillmentHandshake1867000000000
```

`migration:revert` + re-run both exit 0; `IDX_fulfillment_works_external_work_id` present in `pg_indexes` afterwards.

### Conflict resolutions — both took the sibling's improvement, not mine

- **`fulfillment-work-migration-parity.int-spec.ts`** — #2394 replaced the hardcoded table-name list with `[...TABLES].sort()`. That removes outright the trap this branch had merely worked around by hand, and which #2400 walked into. Their derivation is kept; only `fulfillment_work_rejections` is added to `TABLES`.
- **`fulfillment.module.ts`** — the merge produced **two `exports:` keys**. That is valid TypeScript and the first is silently dropped, so it would have un-exported `ROUTING_DECISION_REPOSITORY_TOKEN` with no error anywhere. Unioned into one block carrying all four tokens.

### Gates after the rename

| Gate | Exit |
|---|---|
| `pnpm type-check` | **0** |
| `pnpm lint` (incl. `check-migration-timestamps`) | **0** |
| `pnpm test` | **0** — `apps/api` 1687, `apps/worker` 687 |
| Own specs + parity, explicit | **0** — `Test Suites: 4 passed, 4 total` · `Tests: 49 passed, 49 total` |

The four run explicitly were `fulfillment-handshake`, `fulfillment-work-migration-parity`, `fulfillment-work-schema` and `fulfillment-work-transitions` — all **PASS**, so the parity spec is green against the reordered chain. The full integration suite was not re-run for a rename plus merge, per instruction.

---

## `automation-dispatch-boot` is red on `origin/oms-programme-wave-3a` itself

Verified by checking out the wave tip **`1ac46bb0b`** detached, with none of this branch's code present (`fulfillment-handshake.service.ts` confirmed absent), rebuilding `libs`, and running the spec alone.

```
● Automation action dispatch — DI boot (HARD GATE, #2361) › resolves the REAL dispatcher, not #2360s inert placeholder

  expect(received).toBeInstanceOf(expected)

  Expected constructor: AutomationDispatchService
  Received constructor: AutomationIrreversibleGateService

    49 |     // If the swap is missed, every firing logs and does nothing.
    50 |     const dispatcher = harness.get(AUTOMATION_DISPATCH_SERVICE_TOKEN);
  > 51 |     expect(dispatcher).toBeInstanceOf(AutomationDispatchService);
       |                        ^
    52 |   });
```

`Test Suites: 1 failed, 1 total` · `Tests: 1 failed, 3 passed, 4 total`. Only the identity assertion fails; the other three pass:

```
✕ resolves the REAL dispatcher, not #2360s inert placeholder
✓ resolves the run-recorder seam #2385 will replace
✓ registers an executor for EVERY declared action
✓ resolves the order-lifecycle relay A3 delegates to, with no DI cycle
```

**Cause, pinned:** `libs/core/src/automation/automation.module.ts:88-92` binds

```ts
{ provide: AUTOMATION_DISPATCH_SERVICE_TOKEN, useExisting: AutomationIrreversibleGateService }
```

introduced by `8e2d3bf9c` — *"feat(automation): at-most-one gate for irreversible actions (#2362)"*. That commit **is an ancestor of the wave tip and is not on `origin/main`**, so the red is wave-only and will reach `main` when the wave merges.

Worth noting for whoever fixes it: the spec's stated *intent* — "the REAL dispatcher, not #2360's inert placeholder" — is arguably still satisfied, since the gate **decorates** `AutomationDispatchService` rather than replacing it with a no-op. So the likely correct repair is to assert the gate (or assert it is not the inert placeholder) rather than to revert #2362's binding. I have not touched it; it is another issue's body.

---

Branch clean, pushed at `16162bf93`, main repo clean on `main`, no leftover containers. **Not merged.**

Issue #2395, opened 2026-08-22
[TASK] CORE Application — `selectPrimaryFulfillmentRouter` + the #2047 four-part gate + one-transaction commit


piotrswierzy:
> Part of the OMS programme backlog (**Wave 3a**, stream S1, size L). Backlog slug: `W3a-6`. Sibling `W3a-*` slugs resolve via the Wave-3a epic's checklist.

> **Size justification (L):** the selection, the four-part exactly-one gate, the per-order lock and the single-transaction commit are one correctness argument; split, each half is individually safe and jointly permits a double-ship.

## Problem / Context

Design §5.3, "exactly one router per order" — a verbatim copy of the #2047 four-part shape. Two routers producing two plans for one order is a double-ship: a physical, unrecoverable event.

## Proposed Solution

1. Pure `selectPrimaryFulfillmentRouter(candidates)` over `listCapabilityAdapters({ capability: 'FulfillmentRouter', lazy: true })` — `none` runs today's path byte-identically; `ambiguous` does nothing, persists a reason and is surfaced (never a boot failure — ADR-052 / REVIEW H6).
2. Intent row (#2394, `W3a-5`) persisted **before** the committing `route()`.
3. A write-path guard that reads the intent row and refuses when a live decision or non-cancelled work exists for the order **regardless of router identity**.
4. Mandatory route idempotency key from the decision row; a **declared `route()` timeout strictly below the lock TTL**; per-order lock `fulfillment:route:{orderId}` via `SyncLockPort`.
5. The N work rows and the decision's terminalisation commit in **one transaction**.

Re-route is gated on the order carrying no `cancelledAt`, under the same lock (REVIEW C10). `fulfillment.work.route` is a **core-owned** job type — `JobTypeValues` stays closed (design §9).

[...]

Issue #2393, opened 2026-08-22
[TASK] CORE Domain — `FulfillmentRouterPort` + routing I/O types + `RoutingInput` PII allowlist


piotrswierzy:
> Part of the OMS programme backlog (**Wave 3a**, stream S1, size M). Backlog slug: `W3a-4`. Sibling `W3a-*` slugs resolve via the Wave-3a epic's checklist. Design of record: `docs/plans/analysis/DESIGN-oms-authority-model.md` §5.3, ADR-062.

## Problem / Context

Design §5.3. Two methods, one non-committing: `evaluate()` is OL's differentiator — ADR-044 records that nothing planned produces a dry-run. `RoutingInput` is handed buyer shipping data to filter on, which ADR-062 §Decision 2 bounds by construction.

## Proposed Solution

```ts
interface FulfillmentRouterPort {
  evaluate(input: RoutingInput): Promise<RoutingEvaluation>;  // non-committing
  route(input: RoutingInput): Promise<RoutingPlan>;           // committing
}
```

`RoutingPlan = { decisionId; assignments[]; unfulfillable[]; holds[]; explanation: RoutingExplanationStep[] }` plus the third `pending { decisionId }` arm (REVIEW G5) — 3a **declares** the arm and rejects it with a named error; its *consumption* is Wave 4 (`W4-3`). `unfulfillable` lines resolve as line-scoped refund/return, never an invented partial-cancel state no source can express.

Rule names crossing the port are **opaque strings with display labels** — the closed named filters/sorts and their coercer are owned by `@openlinker/oms` (REVIEW H7), so a vendor's own rule names render. `RoutingInput.shipTo` is an **explicit allowlist projection**: country, postcode, city only — never name / email / phone — `OL_STORE_PII`-aware with a degraded hash-only shape (ADR-062, the MCP-tools projection discipline).

## Classification

**Type**: CORE
**Layer**: Domain
**File(s)**: `libs/core/src/fulfillment/domain/ports/fulfillment-router.port.ts`, `libs/core/src/fulfillment/domain/types/routing.types.ts`

## Dependencies

- `W3a-2`

## Assumptions

- Full per-method error unions and wall-clock budgets are Wave-4 hardening (`W4-1`, `W4-2`); the port file carries a pointer line to them.

## Acceptance Criteria

- [ ] A spec asserts `RoutingInput` cannot carry a name/email/phone field (structural allowlist test)
[...]

Issue #2392, opened 2026-08-22
[TASK] CORE Infrastructure — `fulfillment_works` / `_lines` / `_holds` schema + writer-disciplined repository


piotrswierzy:
> Part of the OMS programme backlog (**Wave 3a**, stream S1, size L). Backlog slug: `W3a-3`. Sibling `W3a-*` slugs resolve via the Wave-3a epic's checklist. **Migration-bearing**: use a synthetic-sequential migration timestamp (`docs/migrations.md` rule 3).

> **Size justification (L):** three related tables land together with the counter `CHECK` and the per-column writer table — splitting them would ship a work row with no lines to constrain.

## Problem / Context

REVIEW C10: `fulfillment_works` is a **five-writer table** (router, executor handshake, progress ingress, operator actions, sweeps). Without per-column ownership it repeats the `order_records` multi-writer problem at a hotter grain; without a counter CHECK, "3 of 5 shipped" can silently become "6 of 5".

## Proposed Solution

Migration via `pnpm --filter @openlinker/api migration:generate` (see `docs/migrations.md`):

- `fulfillment_works` — `orderId`, `locationId`, `deliveryMethod`, `status`, `requestStatus`, `assignmentAttempt` int NOT NULL DEFAULT 0, `dispatchRelayedAt` nullable, `version` int (optimistic token), `cancelledAt` nullable
- `fulfillment_work_lines` with `CHECK (fulfilledQuantity + cancelledQuantity <= totalQuantity)`
- `fulfillment_holds` — first-class rows, ≤10 active per work, `HoldReason` from the leaf
- `shipment_lines.fulfillmentWorkId` nullable (the 1:N `Shipment` relation of §5.2 — the shipment keeps its identity); **wired** in `W3a-13`

`FulfillmentWorkRepositoryPort` in `domain/ports/` exposes **one narrow conditional UPDATE per axis transition** (the `ShipmentRepository.claimWaybillRelay` / #2047 shape), each returning `affected > 0` as the answer, each with its named domain error. The per-column owner table is documented in the repository file header. **No new `order_records` column is written by this layer** (§5.2).

## Classification

**Type**: CORE
**Layer**: Infrastructure
**File(s)**: `libs/core/src/fulfillment/infrastructure/persistence/**`, `apps/api/src/migrations/*`

## Dependencies

- `W3a-2`

## Acceptance Criteria

[...]

Issue #2100, opened 2026-08-14
[TASK] Invoicing — persist + surface the auto-issue block reason instead of logging it only (ADR-041 decision 11)


norbert-kulus-blockydevs:
## Problem / Context

When OpenLinker decides **not** to issue a fiscal document for a qualifying order, that decision exists only in a log line. Nothing is persisted, and nothing surfaces outside a single order's detail drawer.

`AutoIssueTriggerService.onOrderTransition` (`libs/core/src/invoicing/application/services/auto-issue-trigger.service.ts`) has four such exits, all of which `return` after logging:

| Exit | Condition | Log level |
|---|---|---|
| `selection.kind === 'ambiguous'` | several active `Invoicing` connections, none carries `config.invoicing.isPrimary` (`selectPrimaryInvoicingConnection`, `libs/core/src/invoicing/domain/types/invoicing-primary.types.ts`) | `error` |
| `triggerModel === 'manual'` | `qualifies()` returns false | none (silent) |
| `triggerModel === 'batched'` | `BatchedTriggerNotImplementedError` → caught by `logIssuanceFailure` | `warn` |
| selected connection vanished | unreachable defensive branch | `error` |

[ADR-041](https://github.com/openlinker-project/openlinker/blob/main/docs/architecture/adrs/041-sales-document-routing-policy.md) states the visibility contract for exactly these exits, twice and unambiguously:

> §54 — "Every block issues nothing - never a partial fan-out - reusing the service's existing PII-safe log envelope. A block is **never log-only**: it also persists a named reason (decision 11), because 'OL silently declined to issue' is exactly as opaque to an operator as a wrong pick would be dangerous."

> §105 — "Both unions carry the same visibility contract: persisted and operator-visible, never log-only. […] Leaving them log-only would be the silent-block cousin of the silence-and-pick-one that decision 6 forbids."

**Why the log is not enough.** The only operator-facing surface today is client-side re-derivation in the order-detail invoice panel (`apps/web/src/features/invoicing/components/order-invoice-panel.tsx` renders "Automatic invoicing is off for this order" by recomputing the ambiguity from the connection list). That means:

- an install where auto-invoicing silently stopped for **every** order looks completely normal on `/orders` and `/invoices`;
- there is no health bucket, list badge, or filter for "blocked, no document";
- the FE derivation is a *reconstruction*, not the backend's actual decision — the two can diverge (e.g. the `manual`-primary case, which the panel does not model);
[...]

PR #2709, opened 2026-08-31
feat(fulfillment): routing_decisions intent table + live partial-unique index (#2394)

### LINKED ISSUE #2394: [TASK] CORE Infrastructure — `routing_decisions` intent table + live partial-unique index
> Part of the OMS programme backlog (**Wave 3a**, stream S1, size M). Backlog slug: `W3a-5`. Sibling `W3a-*` slugs resolve via the Wave-3a epic's checklist. **Migration-bearing**: use a synthetic-sequential migration timestamp (`docs/migrations.md` rule 3).

## Problem / Context

REVIEW C2 — the review's highest-confidence finding class: **persisted evidence must land before the boundary it protects**. A lock alone cannot supply that ordering; this is the exact property the shipped #2047 invoicing guard gets right.

## Proposed Solution

`routing_decisions` (`orderId`, `routerConnectionId`, `state: live | committed | abandoned`, `idempotencyKey` derived from the row, timestamps) with a **partial unique index on the live decision per order** (the NULL-distinct partial-index pattern already used by `product_content_field`). Repository port exposes `claimIntent` (insert-if-none-live via the `insertIfAbsent` + domain-error + re-select recovery shape) and `terminalise`. The route idempotency key is derived from the **decision row**, never from the job id — the #2039 `reconcileId` lesson: a retrying/resuming job is a different job.

## Classification

**Type**: CORE
**Layer**: Infrastructure
**File(s)**: `libs/core/src/fulfillment/infrastructure/persistence/repositories/routing-decision.repository.ts`, `apps/api/src/migrations/*`

## Dependencies

- `W3a-2`

## Acceptance Criteria

- [ ] `pnpm --filter @openlinker/api migration:show` reports no pending migrations after the `routing_decisions` migration; `synchronize: true` is not introduced
- [ ] Two concurrent claims for one order leave exactly one live row (int-spec on real Postgres)
- [ ] A retry after a crashed route re-derives an identical idempotency key
- [ ] Repository throws a domain error on unique violation

piotrswierzy:
Closes #2394 (`W3a-5`, ADR-054 R1, DESIGN §5.3).

OpenLinker's own record that it is **about to** ask a router where an order is fulfilled from, persisted **before** the committing `route()` call. REVIEW C2: persisted evidence must land before the boundary it protects — an ordering a lock cannot supply, since a lock is lost on process death, TTL expiry or a Redis blip, and the peer that acquires it next has no way to learn a `route()` call is already in flight. The shipped #2047 invoicing guard is the shape being copied.

## The live partial-unique index, and why the predicate is exactly that wide

```sql
CREATE UNIQUE INDEX "UQ_routing_decisions_live_order"
  ON routing_decisions ("orderId") WHERE state = 'live';
```

Verified against a real Postgres (below), which stored it as `... USING btree ("orderId") WHERE ((state)::text = 'live'::text)`.

Three ways to get this wrong, and why each is rejected:

- **Unconditional on `("orderId")`** — forbids any second decision for an order, ever, breaking the legitimate re-route DESIGN §5.4 requires (`short_picked` + `releaseShortfall` re-enters `route()` with the rejecter blocked). Too wide.
- **Keyed `("orderId", "routerConnectionId")`** — permits two routers to hold two live decisions for one order, which is precisely the double-ship #2395's guard must refuse *regardless of router identity*. A double-ship is a physical, unrecoverable event. Too narrow.
- **Predicate `state = 'live'` only** — a terminal row must **leave** the index so a fresh decision can claim. `committed` and `abandoned` are both terminal and both leave.

**On the mutable-predicate warning next door**: `shipment.orm-entity.ts` argues against putting a mutable column in a partial-index predicate and keys on the monotone `providerShipmentId IS NULL` marker instead. That warning does **not** govern here — rows leaving the index on terminalisation is the *mechanism*, not a side effect. The governing precedents are `UQ_order_changes_open_target` (`WHERE status IN ('pending','requested')`) and `UQ_reservations_active_line` (`WHERE status = 'held'`), both of which key uniqueness on an open-state predicate exactly as this does.

**Necessary, not sufficient.** The index enforces at most one live *decision* per order. DESIGN §5.3 also requires refusing when non-cancelled *work* exists — a different set, since a `committed` decision with live work leaves the index free. That half is #2395's; nothing here refuses it.

## Uniqueness is the index, never an application check

`claimIntent` inserts and lets the index refuse, translating `23505` into `RoutingDecisionAlreadyLiveError`. There is no `SELECT`-then-`INSERT` anywhere: at READ COMMITTED a plain `SELECT` takes no locks and the conflicting row is a phantom, so a read-then-act guard enforces nothing. The int-spec proves the real thing by issuing two claims **concurrently** and asserting over the persisted row count rather than over which promise happened to reject.

The `23505` match is on **SQLSTATE *and* constraint name** (the #2392 rule): this table carries two unique constraints, so catching every `23505` would report a PK collision as "a live decision already exists" — an error naming a state that is fine, about a constraint that did not fail.

It **throws** rather than returning the incumbent, unlike the `exchange_rates` `insertIfAbsent` shape: there the caller wants *a* rate and any winner will do, whereas here a live decision existing **is** the refusal, and handing back the winner would invite a caller to route against another router's intent.

## Writer discipline

| Column | Writer |
|---|---|
| `id` / `orderId` / `routerConnectionId` / `createdAt` | `claimIntent`, insert-only |
| `state` / `routerDecisionRef` / `abandonReason` / `terminalisedAt` | `terminalise`, and nothing else |
| `updatedAt` | TypeORM's `@UpdateDateColumn`, injected into the conditional UPDATE — pinned by an int-spec, not asserted in a docblock |

There is **no `save(decision)`**. A row is inserted once and mutated exactly once by one narrow conditional UPDATE (`WHERE id = :id AND state = 'live'`), so the four terminal-owned columns cannot be null-then-reset by an unrelated write. `state` is not accepted at create — a claim is live by definition, and accepting it would allow inserting an already-terminal row. There is no `expectedState` parameter either: `live` is the only legal precondition, and a choice that does not exist must not be offered.

`(result.affected ?? 0) > 0` — the `?? 0` is load-bearing; an `undefined` affected count coercing to a truthy claim is the silent double-apply shape.

## Foreign keys: none, per relation

Both references are **cross-aggregate by value** (the `order_changes` / `refund_records` / `fulfillment_works.orderId` precedent), not part-of-parent children:

- **`orderId` → orders** — an INTENT record must survive the thing it decided about. A re-ingested or removed order must not cascade away the decision history.
- **`routerConnectionId` → connections** — a deleted connection must not erase the audit of what it decided.

Confirmed empirically: `pg_constraint` reports only `PK_routing_decisions`.

`IDX_routing_decisions_order` is unconditional and **not** redundant with the partial index — the history read ("every decision for this order, whatever its state") cannot use an index restricted to live rows. `createdAt` rides along because that read is ordered (the `IDX_order_changes_order` shape).

## Migration-parity coverage — extended, and it is the point

`fulfillment-work-migration-parity.int-spec.ts` (#2392) is the **only automated check of a migration in this repo**: the harness sets `migrationsRun: false` and builds schema by `synchronize`, so a migration is otherwise verified by nothing. `routing_decisions` is added to its `TABLES` list rather than getting a sibling spec — that file already runs the full migration chain once per integration run, the expensive part, so a fourth table costs one more row in three cheap catalogue queries. Its **partial unique index is exactly what `indexdef` comparison exists to catch**: a predicate differing between migration and entity would let the guard hold in production and silently not in tests.

This is why every index is declared class-level **with the same name the migration uses**, and why the PK carries `primaryKeyConstraintName` — without it `synchronize` mints a hash name and the two schemas differ on a constraint name, which is precisely the drift this spec exists to catch.

One trap removed while there: the "should build all three tables" assertion hardcoded its three table names, so extending `TABLES` made the *first* thing a fourth table did be a failure unrelated to the schema it was checking. It is now derived from `TABLES`.

## Migration verified from empty against a throwaway Postgres

Run from an empty database on a spare port (55432, container since removed):

```
[X] 148 AddReturnLineEventsTimelineIndex1863000000000
[X] 149 CreateFulfillmentWorks1864000000000
[X] 150 CreateRoutingDecisions1865000000000
```

**0 pending.** `1865000000000` sorts strictly above every migration on `origin/main` (max `1849000000003`) and on `oms-programme-wave-3a` (max `1864000000000`); filename prefix, class suffix and the `name` property move together (`check-migration-timestamps` rule 4).

No `CREATE EXTENSION "uuid-ossp"`, unlike its #2392 neighbour — that migration needs it because its line and hold PKs default to `uuid_generate_v4()`, whereas this table's PK is `text` minted in application code, so copying the statement would propagate the #2684 workaround to a migration that does not need it.

## Two findings from the tech-review pass, both fixed

This diff had never received a review; both findings are in the second commit.

1. **An illegal terminalisation was representable.** `TerminaliseRoutingDecisionInput` was a flat interface, so `{state: 'committed', abandonReason: 'plan-pending'}` type-checked and would have persisted a row whose two columns contradict each other, which nothing downstream could interpret. It is now a discriminated union whose arms are **derived** from `RoutingDecisionState`, so a fourth terminal state gets an arm automatically rather than silently becoming unwritable through the port. (Literal arms plus a separate exhaustiveness assertion was the first attempt; `noUnusedLocals` rejects the unused type alias.)

   The asymmetry is deliberate: `abandonReason` is refused on the `committed` arm, while `routerDecisionRef` stays legal on **both** — #2393's `plan-not-conserving` describes a router that answered with a `RoutingPlan`, and therefore with a `decisionId`, which OpenLinker then refused. Dropping it there would discard the one value letting an operator correlate the refusal against the vendor's log.

2. **Nothing exercised the `abandonReason` write path.** The only DB-level check wrote the column with raw SQL to prove the *read* coerces an unrecognised value (the #2100 rule) — which says nothing about `terminalise`. Verified by deleting `abandonReason` from the `set()` call: **all 7 existing unit tests passed**. A unit assertion over the assigned columns and an int-spec round trip are added; the unit one was confirmed red for the right reason (`Received: undefined`, not a compile error) before being made green.

## Deliberately deferred

- The `abandonReason` union declares only reasons whose code **already ships** (`plan-pending`, `plan-not-conserving`, both #2393's). Candidates describing #2395's own internals would be guesses about code that does not exist; the column is `varchar(64)` so #2395 adds members with no migration and no `ALTER TYPE`.
- `deriveRouteIdempotencyKey` is a pure function of the row's immutable id and is **not persisted** — a stored column can be written with a value that is not the derivation. A retry re-derives a byte-identical key; a re-route is a new row and so a new key, which is correct (#2039's `reconcileId` lesson). If a router ever echoes a key back, persist it then, as an additive nullable column.
- `RoutingDecisionRepositoryPort` is **not** exported from the barrel — a `*RepositoryPort` is an intra-context persistence contract that `check-cross-context-imports` rejects by deny pattern. #2395 injects it by token from inside the context.

## Gates

| Gate | Exit |
|---|---|
| `pnpm lint` (incl. `check:invariants`, 35 checks — 34 `.mjs` + 1 `.sh`) | **0** |
| `pnpm type-check` | **0** |
| `pnpm test` | **0** — 0 `FAIL` lines; `libs/core` 366 suites / 5460 tests |
| `pnpm test:integration` | reported in a comment below |

Known pre-existing integration failures, named and not chased: **#2638** `earliest-order-date`, **#2639** `allegro-prestashop-carrier-mapping`.

https://claude.ai/code/session_01EWKtLwTcP2ppoknaq4Ca3Z


piotrswierzy:
## Gate results

All four run from scratch on this branch (`56d4d7deb`), Node 22.

| Gate | Exit | Detail |
|---|---|---|
| `pnpm lint` | **0** | includes `check:invariants` — **35** checks (34 `.mjs` + 1 `.sh`), all green |
| `pnpm type-check` | **0** | |
| `pnpm test` | **0** | 0 `FAIL` lines; `libs/core` 366 suites / 5460 tests, `apps/api` 126/1677, `apps/worker` 50/668 |
| `pnpm test:integration` | **1** | `Test Suites: 2 failed, 149 passed, 151 total` / `Tests: 3 failed, 1 skipped, 1260 passed, 1264 total` |

**Both suites this PR depends on passed:**

```
PASS test/integration/routing-decision-intent.int-spec.ts (5.257 s)
PASS test/integration/fulfillment-work-migration-parity.int-spec.ts
```

The parity pass is the meaningful one — it is what proves the migration and the ORM entity agree on the partial-unique predicate, and it is the only automated check of a migration in this repo.

### The two failures, neither this PR's

```
FAIL test/integration/earliest-order-date.int-spec.ts
FAIL test/integration/prestashop/prestashop-webhook-provisioning.int-spec.ts (231.898 s)
```

- `earliest-order-date` — known pre-existing, **#2638**. Not chased.
- `prestashop-webhook-provisioning` — a container-boot timeout, not an assertion: `Port 80 not bound after 180000ms` at `prestashop-container.helper.ts:697`. It is the one spec that takes its **own** PrestaShop container rather than the shared one, so booting a second PS instance is exactly the resource contention the testing guide describes. **This diff touches zero PrestaShop files** (`git diff --name-only c7ee984f1..HEAD | grep -ci prestashop` → `0`). Corroborating: **#2639** `allegro-prestashop-carrier-mapping` failed on an earlier run of this same branch and **passed** on this one — these PS suites are environment-flaky here, not deterministic.

### Migration verified from empty against a throwaway Postgres

`postgres:16-alpine` on port 55432, container removed afterwards.

```
[X] 148 AddReturnLineEventsTimelineIndex1863000000000
[X] 149 CreateFulfillmentWorks1864000000000
[X] 150 CreateRoutingDecisions1865000000000
```

**0 pending.** The index Postgres actually stored:

```
UQ_routing_decisions_live_order | CREATE UNIQUE INDEX "UQ_routing_decisions_live_order"
  ON public.routing_decisions USING btree ("orderId") WHERE ((state)::text = 'live'::text)
IDX_routing_decisions_order     | CREATE INDEX "IDX_routing_decisions_order"
  ON public.routing_decisions USING btree ("orderId", "createdAt")
PK_routing_decisions            | CREATE UNIQUE INDEX ... USING btree (id)
```

`pg_constraint` reports only `PK_routing_decisions` — confirming no foreign keys, as intended for two cross-aggregate by-value references.

Timestamp tail re-verified after the rebase: `1865000000000` sorts strictly above `origin/main`'s max (`1849000000003`) and `oms-programme-wave-3a`'s max (`1864000000000`). #2400's claim-table migration has not landed yet.

### Review findings, and what was done

The diff had never had a review. Three findings, all fixed in commits 2 and 3.

1. **An illegal terminalisation was representable** — `{state: 'committed', abandonReason: 'plan-pending'}` type-checked and would have persisted a row whose two columns contradict each other. Now a discriminated union with arms *derived* from `RoutingDecisionState`. First attempt used literal arms plus a standalone exhaustiveness assertion; `noUnusedLocals` rejects the unused type alias, so the arms are derived instead, which needs no assertion and cannot drift. Verified both directions with a compile probe inside the package: the illegal literal errors, `{state: 'abandoned', abandonReason, routerDecisionRef}` compiles clean.
2. **Nothing exercised the `abandonReason` write path** — the only DB-level check wrote the column with raw SQL to prove the *read* coerces. Proved the gap by deleting `abandonReason` from the `set()` call: **all 7 existing unit tests passed**. Added a unit assertion over the assigned columns and an int-spec round trip; the unit one was confirmed red for the right reason (`Received: undefined`, not a compile error) before being made green.
3. **`docs/architecture-overview.md` §26 was not updated** — every other slice in this wave (#2392, #2398) records itself there, so the index predicate's reasoning lived only in code comments. Added.

Nothing deferred.

piotrswierzy:
## Migration timestamp collision with #2400 — resolved by renumbering to `1866000000000`

#2400 merged `1865000000000-create-fulfillment-progress-claims.ts` into `oms-programme-wave-3a` while this branch was in review, colliding with this branch's `1865000000000-create-routing-decisions.ts`. The check in the original PR body was correct when made; the collision opened underneath it.

This matters rather than being cosmetic: TypeORM sorts migrations by timestamp alone with **no deterministic tie-breaker**, so a collision can leave one `up()` body silently unapplied (#374).

**Renumbered to `1866000000000`** — filename prefix, exported class suffix and the `name` property moved together, or `check-migration-timestamps` rule 2 fires on the half-rename. Swept for every reference to the old timestamp and class name; the only remaining `1865000000000` mentions in this branch's own files are deliberate historical ones (the migration header explaining *why* it moved, and the analysis doc recording that the prefix was free when analysed).

```
$ node scripts/check-migration-timestamps.mjs --self-check
migration-timestamps: self-check OK
$ node scripts/check-migration-timestamps.mjs
migration-timestamps: OK (151 migrations (incl. 1 plugin dir); ordering vs origin/main: checked)
```

**0 violations**, confirmed before and after the rename.

### Chain re-run from empty — ordering is genuinely different now

Fresh `postgres:16-alpine` on port 55433, container removed afterwards:

```
[X] 148 AddReturnLineEventsTimelineIndex1863000000000
[X] 149 CreateFulfillmentWorks1864000000000
[X] 150 CreateFulfillmentProgressClaims1865000000000
[X] 151 CreateRoutingDecisions1866000000000
```

**0 pending**, and `create-fulfillment-progress-claims` correctly precedes this one. The index Postgres stored is unchanged by the renumber:

```
CREATE UNIQUE INDEX "UQ_routing_decisions_live_order" ON public.routing_decisions
  USING btree ("orderId") WHERE ((state)::text = 'live'::text)
```

### Merge conflicts — five, all additive

`origin/oms-programme-wave-3a` merged in (it **is** an ancestor of HEAD). #2400 and #2404 touched the same fulfillment files; resolved by intent, then the invariant re-checked:

- `fulfillment.tokens.ts` / `fulfillment.module.ts` / `index.ts` — both sides' registrations kept. Both issues independently reached the same barrel conclusion: the `*RepositoryPort` stays off it, only the input shapes and the service interface are exported.
- `architecture-overview.md` — #2404's §26 bullet and this one both kept.
- `fulfillment-work-migration-parity.int-spec.ts` — both tables in `TABLES`, and the derived `[...TABLES].sort()` assertion kept. Worth recording: **#2400 walked straight into the trap this branch removed** — it added its table and then had to hand-edit the restated literal list. Deriving the assertion means the next table costs one line, not two, and cannot go stale.

### Gates re-run after the merge

| Gate | Exit |
|---|---|
| `pnpm lint` (incl. `check:invariants`, 35 checks) | **0** |
| `pnpm type-check` | **0** |
| `pnpm test` | **0** — 0 `FAIL` lines |

Integration was **not** re-run in full for a rename plus a merge. The two suites that matter were run explicitly:

```
PASS test/integration/routing-decision-intent.int-spec.ts (20.338 s)
PASS test/integration/fulfillment-work-migration-parity.int-spec.ts
Test Suites: 2 passed, 2 total
Tests:       14 passed, 14 total     (exit 0)
```

The parity spec now covers **both** newly merged tables, so that pass also confirms #2400's schema agrees with its entity on this branch.

The last full integration run on this branch (pre-merge) was `Test Suites: 2 failed, 149 passed, 151 total` — **#2638** `earliest-order-date` and a PrestaShop container-boot timeout (`Port 80 not bound after 180000ms`), neither this PR's; this diff touches zero PrestaShop files.

### Two false signals worked through, for the record

- An apparent test failure was my own 10-minute Bash timeout SIGTERM-ing the child process group mid-suite — not a real failure.
- `pnpm lint` initially failed on my *own* newly added test line (`Unsafe member access [0] on an any value`), not on the reviewed diff.

Branch is at `3e2f64c1c`. **Not merged.**

Issue #2532, opened 2026-08-26
[TASK] Interface — verify and cover the per-connection routing writes


norbert-kulus-blockydevs:
## Problem / Context

Document kind, goes-first and trigger are written from the current settings table. A connection with no document kind is not a routing candidate at all, so these writes are load-bearing — the redesign draft that demoted the table to a read-only list would have removed the only way to configure them.

## Proposed Solution

Confirm the three writes are exposed and correct, extend them where the redesign needs a field they do not cover, and add controller coverage so a future refactor cannot quietly drop one. No new capability unless a gap is found; record the finding either way.

## Classification

**Type**: Interface
**Layer**: Interface
**File(s)**: `apps/api/src/`, `apps/web/src/features/sales-documents/api/sales-documents.types.ts`

## Dependencies

- Parent mini-epic: #2528

## Acceptance Criteria

- [ ] Each of the three writes has controller coverage.
- [ ] A gap, if any, is closed or filed with a reason.
- [ ] The one-primary-across-all-connections rule is asserted by a test.
- [ ] The issue records explicitly whether anything had to change.
- [ ] Tests added or updated for non-trivial logic
- [ ] No architecture boundary violations (CORE ↔ Integration)

## Branching and merge target

Branch: `{issue#}-{short-kebab}`, created **from its mini-epic branch**.
PR target: **its mini-epic branch**.

## How this issue gets finished

1. Implement on the branch above.
2. Open the PR into its mini-epic branch.
3. Run `/pr-review <PR>` and post the verdict as a PR comment.
4. Fix **every** finding, and reply under the comment saying what changed for each one.
5. Run `/pr-review` again. Repeat until the review comes back approved with no findings.
6. Only then merge.

Quality gate before every commit: `pnpm lint`, `pnpm type-check`, `pnpm test`.
Commits are signed off (`git commit -s`).

norbert-kulus-blockydevs:
Closing — implemented and merged via #2724.

Issue #2531, opened 2026-08-26
[TASK] CORE + Interface — acknowledged "no document by choice" per market


norbert-kulus-blockydevs:
## Problem / Context

A market that deliberately issues nothing is indistinguishable from one nobody has configured, so a correct configuration reads as an outstanding problem forever.

## Proposed Solution

Persist and expose the acknowledgement per market, and include it in the merged list so the UI can render it as its own state rather than as an absence. Setting it is an explicit operator act; clearing it is too.

## Classification

**Type**: CORE + Interface
**Layer**: Application / Interface
**File(s)**: `libs/core/src/sales-documents/`, `apps/api/src/`

## Dependencies

- Parent mini-epic: #2528
- The merged market list task in this mini-epic.

## Acceptance Criteria

- [ ] The acknowledgement round-trips and appears in the merged list.
- [ ] An acknowledged market is never reported as unresolved.
- [ ] Clearing it returns the market to its underlying state.
- [ ] A spec covers set, read and clear.
- [ ] Tests added or updated for non-trivial logic
- [ ] No architecture boundary violations (CORE ↔ Integration)

## Branching and merge target

Branch: `{issue#}-{short-kebab}`, created **from its mini-epic branch**.
PR target: **its mini-epic branch**.

## How this issue gets finished

1. Implement on the branch above.
2. Open the PR into its mini-epic branch.
3. Run `/pr-review <PR>` and post the verdict as a PR comment.
4. Fix **every** finding, and reply under the comment saying what changed for each one.
5. Run `/pr-review` again. Repeat until the review comes back approved with no findings.
6. Only then merge.

Quality gate before every commit: `pnpm lint`, `pnpm type-check`, `pnpm test`.
Commits are signed off (`git commit -s`).

norbert-kulus-blockydevs:
Closing — implemented and merged via #2723.

Issue #2530, opened 2026-08-26
[TASK] Interface — merged market list: configured plus detected


norbert-kulus-blockydevs:
## Problem / Context

The settings page must show configured markets and markets that orders arrive from in one list, ordered so that anything needing a decision comes first. Composing that in the browser from three separate reads would put the classification logic in the client.

## Proposed Solution

One endpoint returning both, each row carrying what it needs: for a configured market its effective outcome and reason; for a detected one its order count and whether a template exists. Reuse the M1 discovery read and the evaluator rather than re-deriving either.

## Classification

**Type**: Interface
**Layer**: Interface
**File(s)**: `apps/api/src/`, `libs/core/src/sales-documents/`

## Dependencies

- Parent mini-epic: #2528
- M1 market discovery
- The template catalogue task in this mini-epic.

## Acceptance Criteria

- [ ] One request returns configured and detected markets.
- [ ] A configured market carries its effective outcome, derived from the shipped evaluator.
- [ ] A detected market carries its order count and template availability.
- [ ] A market that is both configured and receiving orders appears once.
- [ ] Tests added or updated for non-trivial logic
- [ ] No architecture boundary violations (CORE ↔ Integration)

## Branching and merge target

Branch: `{issue#}-{short-kebab}`, created **from its mini-epic branch**.
PR target: **its mini-epic branch**.

## How this issue gets finished

1. Implement on the branch above.
2. Open the PR into its mini-epic branch.
3. Run `/pr-review <PR>` and post the verdict as a PR comment.
4. Fix **every** finding, and reply under the comment saying what changed for each one.
5. Run `/pr-review` again. Repeat until the review comes back approved with no findings.
6. Only then merge.

Quality gate before every commit: `pnpm lint`, `pnpm type-check`, `pnpm test`.
Commits are signed off (`git commit -s`).

norbert-kulus-blockydevs:
Closing — implemented and merged via #2722.

PR #2710, opened 2026-08-31
feat(orders,fiscalization,sales-documents,web): sales-document read model, fiscal lifecycle honesty, async registration, routing config, FE primitives

norbert-kulus-blockydevs:
## Summary

Backend + frontend foundation for the sales-document UX epic (M1, M2, M3, M4, M5 — epics #2514, #2519, #2524, #2528, #2533). M6-M10 continue on branches cut from `epic/sales-documents-ux`, each opening its own PR into `main` (stacked, merged together).

**Sales-document read model (#2514)**
- Neutral `SalesDocumentView` projection type (#2515)
- Batched per-order sales-document read for the orders list (#2516)
- Single-order read + `GET /orders/:id/sales-document` (#2517)
- Market discovery: billing countries with order counts (#2518)

**Fiscal lifecycle honesty (#2519)**
- Third locate outcome for a held-but-not-registered sale
- Closed reconcile outcome vocabulary — a failed check is no longer reported as one
- In-flight registration lease surfaced to callers
- Fiscal artefacts projected as medium + label, never raw content
- One reason-to-copy map for sales-document reasons, guarded against drift
- Fix: locate answers no longer stated as fact when unread
- Fix: a stalled registration no longer asserts an absence it can't support

**Asynchronous manual registration (#2524)**
- Manual registration now routes through the `fiscalization.register` job (#2525)
- Pollable per-order registration status (#2526)
- Frontend polls the registration read and states that work continues (#2527)

**Routing configuration surface (#2528)**
- Per-connection routing writes on the connection controller (#2529)
- Merged market list across every connection (#2530)
- Merged list never reports an acknowledged market as unresolved (#2531)
- Controller coverage for the three per-connection routing writes (#2532)

**Frontend foundation (#2533)**
- `DocumentKindGlyph` / `DocumentHeadline` primitives
- `DocumentLifecycle` trail for an invoice
- Fix: cell-anchored popover escapes the table's scroll container
- Fix: table declares its row shape so the skeleton reserves space
- Fix: receipt header no longer contradicts its own body
- Fix: "still unknown" toast no longer names a cause the outcome doesn't carry

## Test plan
- [ ] `pnpm lint && pnpm type-check` on this branch
- [ ] `pnpm test`
- [ ] Merge via **"Create a merge commit"** — not squash — so release-please sees each commit individually

🤖 Generated with Claude Code


piotrswierzy (approved):

**Scope:** 137 files, +9572/−438, 38 commits across four milestones (M1 read model, M2 fiscal lifecycle honesty, M3 async registration, M5 FE foundation). Reviewed the core type and service surface (`sales-documents`, `fiscalization`, `orders`), the API controller contract change, the new `<ctx>/types` sub-barrel family and its guard, and the docs. I did not read every FE component or every spec — this is a review of the contracts and the risk surface, not a line-by-line pass over 9.5k lines.

### Summary

The design quality here is high, and the parts that are hardest to get right are the parts that are most carefully done. `SalesDocumentView` is a faithful implementation of ADR-065 rather than a loose interpretation of it, the `<ctx>/types` cycle-breaker is generalised from an existing precedent and mechanically bounded, and the async-registration path handles a trap (a dead job permanently locking a sale out of registration) that most implementations would have shipped. Two things need attention before merge, both about release/verification hygiene rather than design.

### Positive observations

- **`SalesDocumentView` makes ADR-065's rules structural, not documentary.** The invoice/receipt split is a discriminated union specifically so the missing authority axis is *unrepresentable* rather than merely unset — the ADR says "the type makes the other axis unrepresentable rather than merely unset, which is the point," and the code delivers exactly that. Several of the field comments are load-bearing and correct in ways that are easy to get wrong:
  - `inFlightUntil: null` is documented as *not* meaning "not in flight" — a lease in the past is a crashed attempt, so a surface must compare against now rather than test for presence.
  - `artefactCount: 0` on a `registered` row is a **success** (a pure reporting regime returns no artefact), which is exactly ADR-042's position.
  - `documentNumber: null` explicitly does not distinguish "regime has no numbering" from "not yet issued", and says no surface may claim otherwise.
  - `failureMode: null` on a failed row is treated as `in-doubt`, matching the write-path guards, with an explicit warning not to read it as safe-to-retry.
- **`otherRecords` carries `recordId`**, with the reasoning stated: a panel that reports a duplicate document and then offers no way to open it leaves the operator with a fact and no remedy, on precisely the state that most needs one. And `blocksFurtherIssuance` mirrors the write-path guard rather than restating it — with an honest note that the two underlying guards are scoped *differently* (`blocksIssuanceElsewhere` is other-connection-scoped, `blocksFurtherRegistration` deliberately is not) and coincide only for the case this type describes. That is the kind of caveat that normally gets discovered later, in an incident.
- **The `documentKind` open-world / `document.kind` closed asymmetry is deliberate and documented**, including the consequence for consumers: a `switch` on the former needs a default arm and is not compiler-checked. Good — that asymmetry would otherwise be read as an inconsistency and "fixed."
- **`requestRegistration` gets the exactly-once story right under async.** The read gate runs first and is correctly *skipped* for a same-key resume (otherwise a `pending` row whose job died would block its own re-drive); the dead-job re-drive is guarded on `status === 'dead'` and re-runs the same key against the same record, so it can never become a second registration. The comment explaining why the unconditional version was wrong — "it read as though every request re-drove something, which is not what happens" — is the right level of care for a code path where a bug is a duplicate fiscal document.
- **`fiscalRegistrationIdempotencyKey` moved into core** so the HTTP surface and the auto-issue gate cannot drift: "two copies of a key format that must be byte-identical is how one sale ends up with two receipts." Correct, and the previous local helper was exactly that hazard.
- **A failed reconcile *check* is now 502, and explicitly not one of the four closed outcomes.** Reporting a transient network failure as the structural `unsupported` was a real category error; separating them means an operator is never told the provider *cannot* be queried when it merely *was not*.
- **The `<ctx>/types` family is properly bounded.** It generalises `orders/types` (#2155) rather than inventing something new, the barrel-purity spec authorizes the exact specifiers, and `check-types-sub-barrels.mjs` enforces the other side (re-export only from `./domain/types/**`), so a repository port or ORM entity cannot reach a consumer through it. `docs/engineering-standards.md` documents when to add one and warns that a *value* export turns an erased edge back into a real one. Three mirror guards added under `check:invariants` for the reason-copy, reconcile-outcome and progress vocabularies.

### 🟡 IMPORTANT — a breaking API change is committed as a plain `feat`

`POST /fiscal-registrations` changes from **200 → 202**, and its response body changes shape entirely: it returned `FiscalRegistrationResponseDto` (the record), it now returns `AcceptedFiscalRegistrationResponseDto` (an accepted envelope). The semantics invert with it — the old response *was* the outcome, the new one says only that the request was recorded.

The commit is `feat(fiscalization): route manual registration through the register job (#2525)` — no `!`, and `git log` shows **zero** `BREAKING CHANGE` footers on the branch. release-please will therefore cut this as a minor and the changelog will not mention it.

That matters more than usual here because the endpoint's own previous Swagger description **instructed callers to depend on the removed fields**: *"Read `status` / `failureMode`, never the status code, to decide whether the sale was registered."* Any caller that followed that instruction now silently reads `undefined` and cannot tell a registered sale from a failed one. The FE is updated in this PR; a self-hosted operator's script or an integration built against the published OpenAPI is not.

Please add a `BREAKING CHANGE:` footer (or `!`) to that commit naming the status-code and body change, so the release notes carry it. Since the merge is a merge commit rather than a squash, the footer has to go on the commit itself.

### 🟡 IMPORTANT — the test plan is entirely unchecked on a 137-file PR

All three boxes are open: `pnpm lint && pnpm type-check`, `pnpm test`, and the merge-strategy note. On a change this size that is worth resolving explicitly rather than leaving to CI, because the highest-regression surface in this PR is not the new code — it is the **shared primitives**:

- `shared/ui/popover.tsx` (escaping the table's scroll container)
- `shared/ui/data-table.tsx` + `data-table-skeleton.tsx` (a table declaring its row shape)

Those are consumed by six list pages this PR also touches (orders, invoices, listings, products, shipments, customers). A `Popover` positioning or `DataTable` row-shape change that is right for the sales-document panel and subtly wrong elsewhere would not be caught by any spec in the `sales-documents` or `fiscalization` trees. Given `docs/frontend-ui-style-guide.md` § Density carries per-table carve-outs measured against specific row compositions, a skeleton/row-shape change deserves a note about which of those six it was checked against.

### 🟢 SUGGESTION

Four milestones in one PR is a lot to hold at once, and the review surface is dominated by the parts least likely to be read closely. Since M4 and M6–M10 continue on a fresh branch anyway, splitting the shared-UI primitive changes (M5) out from the backend contract work would have made both easier to verify — not worth restructuring now, but worth considering for the next epic branch.

### Merge readiness

✅ **Approve** — no architecture, boundary, or contract-design concerns; the cross-context discipline is if anything better than the repo's baseline, and the ADR-065 implementation is exactly what that ADR asked for. Both IMPORTANT items are process rather than design: **add the `BREAKING CHANGE:` footer to the #2525 commit**, and confirm the suite is green (particularly the six list pages touched by the shared-primitive changes) before merging.

norbert-kulus-blockydevs:
Addressed both IMPORTANT findings from the review.

**Breaking API change now carries a footer.** Rewriting `b53feca5c` (already pushed/merged into this shared branch via #2695) would need a force-push, so instead added a follow-up commit, `5ede42188`, with a proper `!` + `BREAKING CHANGE:` footer naming the status-code (200→202) and body-shape change (`FiscalRegistrationResponseDto` → `AcceptedFiscalRegistrationResponseDto`), and pointing callers at the registration-status poll. release-please reads footers from any commit in range, so this is sufficient for the changelog / major-bump signal without rewriting history.

**Test plan checked:**
- `pnpm lint` — 0 errors (warnings only, pre-existing `explicit-function-return-type` on test files), all `check:invariants` gates green including the three new mirror guards (`check-sales-document-reason-mirror`, `check-fiscal-reconcile-outcome-mirror`, `check-fiscal-registration-progress-mirror`).
- `pnpm type-check` — 0 errors across every package.
- Ran the specs for the two shared primitives you flagged (`data-table-skeleton.test.tsx`, `popover.test.tsx`, `data-table.test.tsx`): 55/55 passing.
- Checked the row-shape / skeleton diff (`c5821370f`) and the popover-portal diff (`41af973e0`) directly: the row-shape change is additive (36px stays the floor, so a single-line table is byte-identical) and is applied to exactly the six list pages that carry a multi-line cell (customers, invoices, listings, orders, products, shipments) — same six you named. The popover scroll-dismiss is opt-in (`dismissOnViewportChange`), so it doesn't change behavior for existing popovers (e.g. `analytics-infotip.tsx`) that don't request it.

Merge strategy note in the description stands — merge commit, not squash.

norbert-kulus-blockydevs:
Converted to draft: M1-M4 (backend) are merged into this branch, but M5-M9 (frontend + settings surfaces) are not yet started. This PR stays draft until all ten mini-epics are in.

Note for future mini-epics (M5-M9): going forward these should branch directly from `main` rather than being nested under this epic branch. The epic-branch structure added merge/rebase overhead across sessions without a clear benefit once the backend (M1-M4) was already stable enough to PR into main on its own. New mini-epic PRs can target `main` directly; this PR can be closed once its content is superseded, or kept as the historical rollup of the backend slice - decide at that point.

Issue #2716, opened 2026-08-31
[BUG] CORE — Flaky allegro-order-sync-e2e int-spec: OrderCreateContendedException on a single, non-concurrent lock acquire


norbert-kulus-blockydevs:
## Problem / Context

`apps/worker/test/integration/allegro-order-sync-e2e.int-spec.ts` › `Complete Order Sync Flow` › `should sync orders from poll to order processor` fails intermittently in CI with:

```
SyncJobExecutionError: Marketplace order sync failed: Order create is contended for order ol_order_... (destinationConnectionId=dest-prestashop-1); a concurrent create holds the lock — retry
  at MarketplaceOrderSyncHandler.execute (apps/worker/src/sync/handlers/marketplace-order-sync.handler.ts:50:13)
Cause:
OrderCreateContendedException: ...
  at OrderSyncService.createOrderIdempotently (libs/core/src/orders/application/services/order-sync.service.ts:211:13)
```

Observed twice on 2026-08-31, on two unrelated PRs whose diffs never touch `orders`/`sync` code:
- https://github.com/openlinker-project/openlinker/actions/runs/33380848241/job/99452698826 (PR #2700, frontend-only diff)
- https://github.com/openlinker-project/openlinker/actions/runs/33381044654/job/99453345457 (PR #2710)

`OrderCreateContendedException` is thrown from `createOrderIdempotently` (`libs/core/src/orders/application/services/order-sync.service.ts:211`) when `SyncLockPort.acquire()` returns `null` — i.e. Redis's `SET key val NX PX ttl` did not set the key because it already existed.

**What's already ruled out** (investigated 2026-08-31, see PR discussion):
- `WORKER_RUNNER_ENABLED=false` and `WORKER_INTAKE_ENABLED=false` are both set by `apps/worker/test/jest-global-setup.cjs` for every integration run, so no background `SyncJobRunner` poll loop or `JobIntakeConsumer` stream loop is alive to independently re-process the enqueued job concurrently with the test's own explicit `orderSyncHandler.execute(...)` call.
- The test's `listCapabilityAdapters` mock returns exactly **one** `OrderProcessorManager` destination (`dest-prestashop-1`), so `OrderSyncService.syncOrder`'s `Promise.allSettled(destinations.map(...))` fans out to exactly one `createOrderIdempotently` call — there is no second concurrent caller *within* this test's own code path.
[...]

Issue #2617, opened 2026-08-27
[BUG] CORE/Inventory — concurrent quantity writes for one offer can resolve to the older value


norbert-kulus-blockydevs:
## Problem / Context

The `realtime` lane executes jobs concurrently, so **two quantity writes for the same offer can arrive out of order and the older number wins**. The offer is then left advertising a stale quantity until the next change at the master.

This exists today but is rare — a handful of such writes per hour. The propagation-lane fix multiplies them to several per sale, which is exactly why the guard belongs with that change rather than after it.

## Proposed Solution

Make the quantity write ordered per offer: a monotonic token carried on the job and compared at the write, or per-offer serialisation. Do not rely on lane ordering, which the lane does not promise.

## Classification

**Type**: CORE
**Layer**: Application
**File(s)**: `libs/core/src/inventory/application/services/inventory-sync.service.ts`, `libs/core/src/listings/**` (offer quantity write path)

## Dependencies

- Must land with or before the propagation-lane change

## Acceptance Criteria

- [ ] Two concurrent quantity writes for one offer resolve to the newer value regardless of arrival order
- [ ] The guard costs no extra marketplace call
- [ ] Tests cover the reversed-arrival case explicitly

norbert-kulus-blockydevs:
Done in the epic branch `epic/2590-prestashop-no-module`, reviewed and fixed on PR #2627. Closing so the board reflects it; the branch is merged to `main` as one PR.

Issue #2047, opened 2026-08-12
[BUG] Invoicing — two Invoicing connections issue two invoices for one order


norbert-kulus-blockydevs:
> **UX mockup: [Invoice Connection Lock](https://claude.ai/code/artifact/4dbdb3c6-ea32-42ac-b14b-2a37b40aeb1a)**
> Shows before/after on the order panel, the pending and failed edge cases, and the API change. Read it before implementing - it settles copy and states that this issue only describes.

## Problem / Context

One sale is **one** invoice. `KSeF`, `inFakt` and `Subiekt` are three **alternative routes for that
one invoice to reach the tax authority**, not three complementary steps:

| Route | Who holds the KSeF session | What OL does |
|---|---|---|
| `@openlinker/integrations-ksef` | OpenLinker | submits it itself (`RegulatoryTransmitter`) |
| `@openlinker/integrations-infakt` | inFakt | only reads clearance status (`RegulatoryStatusReader`) |
| `@openlinker/integrations-subiekt` | Subiekt | only reads clearance status (`RegulatoryStatusReader`) |

Today OL treats them as complementary. `AutoIssueTriggerService.onOrderTransition` iterates **every**
active connection that has `Invoicing` in `enabledCapabilities` and enqueues an issuance job for each
one:

- `libs/core/src/invoicing/application/services/auto-issue-trigger.service.ts:108-112` - the capability
  filter over `connectionPort.list({ status: 'active' })`, with no "which one" selection
- same file `:114` - the `for (const connection of connections)` fan-out
- same file `:130` - `const idempotencyKey = \`invoice:${connection.id}:${order.id}\`` - keyed **per
  connection**, so it can never dedupe across connections

Nothing downstream catches it either. The persistence model is per-`(order, connection)` throughout:

- `invoice-record.orm-entity.ts:35` - `IDX_invoice_records_order_connection` on
  `['orderId', 'connectionId']`, **non-unique**; there is no uniqueness on `orderId` alone
- `invoice-record.orm-entity.ts:44` - `UQ_invoice_records_connection_idempotency` is scoped to
  `connectionId`, mirroring the same assumption

**Result:** an operator who enables `Invoicing` on two connections gets **two invoices issued for one
order**, both with their own document number, both potentially transmitted to KSeF. Correcting that
means issuing a correction for a document that should never have existed.

The manual path has the same shape from the other side. `GET /invoicing/orders/:orderId/invoice`
[...]

PR #2717, opened 2026-08-31
fix(sync): make RedisSyncLockService.acquire self-recognizing on NX failure

norbert-kulus-blockydevs:
## Summary
- `RedisSyncLockService.acquire` reported spurious contention when `SET NX PX` failed against its own already-successful set (observed via a resent/retried request under CI-runner network pressure), rather than a genuine second holder.
- On NX failure it now `GET`s the key and treats a match against the token it just minted as its own successful acquisition instead of contention.
- Addresses the leading hypothesis behind the intermittent `OrderCreateContendedException` flake in `apps/worker/test/integration/allegro-order-sync-e2e.int-spec.ts`, which never reproduces locally (isolated or full-suite runs) - see #2716 for the full investigation.

## Test plan
- [x] `libs/core` unit tests: 3467/3467 pass, including 2 new cases in `redis-sync-lock.service.spec.ts` (contention against a different holder still returns `null`; self-recognition on a resent request returns the token)
- [x] `pnpm --filter @openlinker/core type-check` clean
- [x] `eslint` on touched files clean
- [ ] Watch CI across a few runs of this PR to confirm `allegro-order-sync-e2e.int-spec.ts` stops flaking (can't force-reproduce it on demand)

## Review follow-up (piotrswierzy)
- Switched `Closes` to `Refs` below - #2716's own AC requires 5 consecutive green CI runs before closing, and merging this shouldn't auto-close that verification track.
- Narrowed the in-code comment: self-recognition only covers a client-level resend of the same SET command (e.g. reconnect after a lost ack), not a caller-level `acquire()` retry - a caller retry mints a fresh token and still returns `null` as before.
- Left the Lua-script (single round-trip) suggestion as a follow-up rather than in this fix - it's a real optimization but changes the shape of the contended path, and this PR is scoped to the correctness fix.

Refs #2716


piotrswierzy (approved):

**Scope:** `libs/core/src/sync/application/services/redis-sync-lock.service.ts` + its spec. 2 files, +40/−4. Core application layer, no boundary or schema surface.

### Summary

The change is safe and correctly tested. Because `token` is a fresh `randomUUID()` that has never been written anywhere, `holder === token` can only be true if **our own SET** wrote the key — so this can never claim a lock held by someone else, and the fallback correctly returns `null` when the key expired between the failed SET and the GET. The spec covers all three branches, including the negative "different holder" case, and asserts the fast path issues no GET at all.

My reservations are about what it *claims*, not what it does.

### 🟡 IMPORTANT — `Closes #2716` is premature against that issue's own AC

#2716 is explicit that the mechanism is unconfirmed ("**Leading, unconfirmed hypothesis**… This has not been proven; it needs runtime evidence from a reproduction, not just a code read"), and it structures the work as **two tracks**, with AC:

> - [ ] `allegro-order-sync-e2e.int-spec.ts` passes on CI across at least 5 consecutive CI runs (or documented evidence the flake is resolved/root-caused another way)
> - [ ] If root cause is found to be something other than lock self-collision, this issue is updated with the real cause before closing

Merging with `Closes #2716` auto-closes it the moment this lands, discarding both. The issue also warns against a fix that masks rather than resolves — and a defensive hardening whose triggering condition was never observed is exactly the shape that could quietly do that. If the flake recurs after this merges, the issue is closed, the hypothesis reads as confirmed, and the next person starts from a false premise.

Suggest `Refs #2716` and leaving it open for track 2 (the CI-only diagnostic logging), closing it when the 5-run evidence exists. The code change is worth landing either way — the issue itself says to do it "regardless of whether the exact mechanism is ever nailed down."

### 🟡 IMPORTANT — the comment claims a case the code cannot serve

> `a slow/lost ack under load, retried by **the caller** or the client`

`token` is minted **inside** `acquire()`, so a caller-level retry calls `acquire()` again and gets a **different** UUID. `holder === token` is then false and it returns `null` exactly as before. Self-recognition can only fire for a resend of the identical SET command bytes — i.e. the redis client re-sending a queued command across a reconnect — never for a caller retry.

That distinction matters, because someone reading this comment could reasonably conclude that caller-level retry is now self-contention-safe and build a retry loop on that belief. Worth dropping "the caller" and naming the one case this actually covers. (#2716 anticipated this, incidentally: its "or, better, mint the token from a stable per-call input" alternative is the version that *would* cover a caller retry — a deliberate trade-off worth stating in the comment rather than leaving implicit.)

### 🟢 SUGGESTION — one round-trip instead of two on the contended path

The GET runs on **every** failed NX, so every genuinely-contended acquire now costs an extra round-trip — and contention is the normal, expected path for the hot locks (#2617's per-`(connection, offer)` quantity lock treats contention as a routine penalty-free deferral, and #2047's per-order invoice lock is designed around being hit).

A small Lua script doing the SET-NX-then-GET atomically would keep it at one round-trip, match the compare-and-delete/compare-and-PEXPIRE shape `release` and `extend` already use in this same class, and close the (benign, but real) TOCTOU window between the failed SET and the GET. Given the file already reaches for Lua twice, the third would read as consistent rather than as extra machinery.

### Merge readiness

✅ **Approve** — the implementation is correct and well tested. Both IMPORTANT items are about honesty of the surrounding claims rather than behaviour: please switch to `Refs #2716` so the verification track survives, and trim the comment to the case it actually covers.

norbert-kulus-blockydevs:
Addressed both IMPORTANT items:

- `Closes #2716` -> `Refs #2716` in the PR body (and the commit that will merge). #2716 stays open until the 5-consecutive-CI-run evidence lands, per its own AC.
- Narrowed the in-code comment (commit 8c6b71e): dropped "the caller" - `token` is minted fresh inside `acquire()`, so a caller-level retry gets a new UUID and still returns `null`. Self-recognition only covers a client-level resend of the identical SET command bytes (e.g. reconnect after a lost ack).

Left the Lua-script single-round-trip suggestion as a follow-up rather than folding it into this fix - real improvement, but changes the shape of the contended path and this PR is scoped to the correctness bug.

Issue #2529, opened 2026-08-26
[TASK] CORE + Interface — country-keyed starter-template catalogue


norbert-kulus-blockydevs:
## Problem / Context

The Poland starter rules exist as a hardcoded screen, not as data, so no other market can have one and the UI cannot tell whether a given market has guidance available. The current page also offers to adopt the template next to the three rules it has already produced.

## Proposed Solution

Model the catalogue as country-keyed template definitions with a source reference, and expose whether a template exists for a given country. Poland is the only entry. Per #2503 a market without an entry gets no recommendation, and the absence is reportable so the UI can say "we have no guidance for this market" rather than staying silent.

## Classification

**Type**: CORE + Interface
**Layer**: Domain / Interface
**File(s)**: `libs/core/src/sales-documents/`, `apps/api/src/`

## Dependencies

- Parent mini-epic: #2528
- #2503

## Acceptance Criteria

- [ ] The catalogue is data, keyed by country, with a citable source per template.
- [ ] Poland is the only entry; absence is a first-class answer.
- [ ] Adoption is explicit; nothing is applied by merely reading the catalogue.
- [ ] A spec asserts that a country with no template reports absence rather than an empty template.
- [ ] Tests added or updated for non-trivial logic
- [ ] No architecture boundary violations (CORE ↔ Integration)

## Branching and merge target

Branch: `{issue#}-{short-kebab}`, created **from its mini-epic branch**.
PR target: **its mini-epic branch**.

## How this issue gets finished

1. Implement on the branch above.
2. Open the PR into its mini-epic branch.
3. Run `/pr-review <PR>` and post the verdict as a PR comment.
4. Fix **every** finding, and reply under the comment saying what changed for each one.
5. Run `/pr-review` again. Repeat until the review comes back approved with no findings.
6. Only then merge.

Quality gate before every commit: `pnpm lint`, `pnpm type-check`, `pnpm test`.
Commits are signed off (`git commit -s`).

norbert-kulus-blockydevs:
Closing — implemented and merged via #2721.

PR #2725, opened 2026-08-31
feat(sales-documents): M4 - the routing configuration surface (#2528)

### LINKED ISSUE #2528: [EPIC] M4 — Backend: the routing configuration surface
> Parent: #2513
> Mockup: `docs/plans/mockups/sales-document-routing.html` — states `#page=settings`

## Scope

What the settings page reads and writes: the merged market list (configured plus detected), the country-keyed starter templates, the acknowledged "no document by choice" state, and the per-connection routing writes the current page already owns (document kind, goes-first, trigger). The last one exists today and is verified rather than rebuilt — the redesign's first draft demoted that table to a read-only list, which would have stranded every order, since a connection with no document kind is not a routing candidate at all.

## Branching

Mini-epic branch `phase/sdux-m4-routing-config`, created **from `epic/sales-documents-ux`**, opening a PR **into `epic/sales-documents-ux`**.
Each task below gets its own branch off `phase/sdux-m4-routing-config` and a PR **into `phase/sdux-m4-routing-config`**.

## Dependencies

- #2503 (ADR: discovery and templates) merged to `main`.
- M1 (#2514) market discovery read.

- [ ] One read returns configured and detected markets in one shape.
- [ ] A starter template is country-keyed and absent for every country without researched guidance.
- [ ] The acknowledged no-document state round-trips.
- [ ] The per-connection routing writes remain available and covered by tests.
- [ ] Every task branch merged into `phase/sdux-m4-routing-config`, and `phase/sdux-m4-routing-config` merged into `epic/sales-documents-ux` via reviewed PR.

norbert-kulus-blockydevs:
Mini-epic: closes #2528. Tasks #2529, #2530, #2531, #2532 — all merged into this branch (PRs #2721, #2722, #2723, #2724).

## What this adds

Country-keyed starter-template catalogue (`GET /sales-documents/templates`, Poland only today per ADR-066), a merged market list (`GET /sales-documents/markets`) that pairs detected order countries with configured routing state and a per-row routing outcome, and the "acknowledged, no document by choice" override so a deliberately-unconfigured country doesn't read as broken. #2532 adds controller-level test coverage for the three per-connection routing writes (documentKind, isPrimary, triggerModel).

## API shape for M6/M7

```
GET /sales-documents/templates -> { countries: [{ country, sourceLabel, sourceUrl }] }
GET /sales-documents/templates/:country          (unchanged: 200 preview | 404)
POST /sales-documents/templates/:country/adopt   (unchanged)
GET /sales-documents/markets/detected            (unchanged, M1 plain discovery)

GET /sales-documents/markets ->
{
  windowDays: number,
  since: string,
  markets: [{
    country: string,
    orderCount: number | null,
    hasTemplate: boolean,
    ruleCount: number,
    invoiceDefaultConnectionId: string | null,
    receiptDefaultConnectionId: string | null,
    acknowledgedNoDocumentAt: string | null,
    outcome: {
      kind: 'route' | 'aggregate' | 'unresolved' | 'acknowledged',
      documentKind?: string | null,
      connectionId?: string,
      reason?: string,
    },
  }]
}

PUT/DELETE /sales-documents/countries/:country/acknowledgment   (unchanged, #2186)
PATCH /connections/:id { config: { salesDocument: { documentKind }, invoicing: { isPrimary, triggerModel } } }
```

## Decisions worth review attention

- `outcome` on the merged list runs the real `resolveRoutingBatch` against a synthetic representative order (`buyerHasTaxId: undefined`, no amount asserted), never hand-derived — deliberate, to avoid guessing a fact that feeds a legal-document routing decision.
- `'*'` (Rest of world) only appears as its own row when it is itself configured.
- The `acknowledged` outcome kind is an API-layer presentation fact, not part of `SalesDocumentDecision` in core — kept out of the zero-outbound-edge sales-documents concern.
- #2532 found, filed but did not fix a pre-existing gap: connection `documentKind` writes have no save-time capability check (the rule engine has one). Not silent — `AutoIssueTriggerService` reports the mismatch at issuance time — but closing it at save time is out of this task's scope.

## Left out

No second country in the template catalogue (Poland only, per ADR-066 — no researched guidance yet for others). No FE consumption here — that's M6/M7. No caching/pagination on the markets read, matching the existing no-pagination stance for this admin-only settings page.

Gate: type-check clean across affected packages, `pnpm check:invariants` green, all commits GPG-verified.

### DIFF
--- a/apps/api/src/integrations/http/connection.controller.spec.ts
+++ b/apps/api/src/integrations/http/connection.controller.spec.ts
@@ -422,6 +422,72 @@ describe('ConnectionController', () => {
         status: 'disabled',
       });
     });
+
+    // #2532: document kind, "goes first" and trigger are written through this
+    // same generic PATCH via the open `config` JSONB passthrough
+    // (`config.salesDocument.documentKind`, `config.invoicing.isPrimary`,
+    // `config.invoicing.triggerModel` — ADR-041 decision 4, #2047, #2159). A
+    // connection with no document kind is not a routing candidate at all, so
+    // these three fields are load-bearing: this suite pins that the
+    // controller forwards each of them to the service untouched, with no
+    // schema stripping any of the three keys.
+    it('should forward config.salesDocument.documentKind unchanged (#2532)', async () => {
+      service.update.mockResolvedValue(mockConnection);
+
+      const dto = { config: { salesDocument: { documentKind: 'invoice' } } };
+      await controller.update('connection-123', dto, mockAdminUser);
+
+      expect(service.update).toHaveBeenCalledWith('connection-123', {
+        config: { salesDocument: { documentKind: 'invoice' } },
+      });
+    });
+
+    it('should forward config.invoicing.isPrimary unchanged (#2532)', async () => {
+      service.update.mockResolvedValue(mockConnection);
+
+      const dto = { config: { invoicing: { isPrimary: true } } };
+      await controller.update('connection-123', dto, mockAdminUser);
+
+      expect(service.update).toHaveBeenCalledWith('connection-123', {
+        config: { invoicing: { isPrimary: true } },
+      });
+    });
+
+    it('should forward config.invoicing.triggerModel unchanged (#2532)', async () => {
+      service.update.mockResolvedValue(mockConnection);
+
+      const dto = { config: { invoicing: { triggerModel: 'on-paid' } } };
+      await controller.update('connection-123', dto, mockAdminUser);
+
+      expect(service.update).toHaveBeenCalledWith('connection-123', {
+        config: { invoicing: { triggerModel: 'on-paid' } },
+      });
+    });
+
+    it('should forward all three routing fields together in one write (#2532)', async () => {
+      service.update.mockResolvedValue(mockConnection);
+
+      const dto = {
+        config: {
+          salesDocument: { documentKind: 'fiscal-receipt' },
+          invoicing: { isPrimary: false, triggerModel: 'manual' },
+        },
+      };
+      await controller.update('connection-123', dto, mockAdminUser);
+
+      expect(service.update).toHaveBeenCalledWith('connection-123', { config: dto.config });
+    });
+
+    it('should clear the document kind by writing an empty string, matching the FE "Nothing" option (#2532)', async () => {
+      service.update.mockResolvedValue(mockConnection);
+
+      const dto = { config: { salesDocument: { documentKind: '' } } };
+      await controller.update('connection-123', dto, mockAdminUser);
+
+      expect(service.update).toHaveBeenCalledWith('connection-123', {
+        config: { salesDocument: { documentKind: '' } },
+      });
+    });
   });
 
   describe('disable', () => {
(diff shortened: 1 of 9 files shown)


norbert-kulus-blockydevs:
## Mini-epic review: M4 - the routing configuration surface

**Verdict: Approve.**

Reviewed as a roll-up (mini-epic-level review per the epic's process, not per-task). Diff is small and bounded: 638 insertions across 9 files, all under `apps/api/src/sales-documents/`.

### Gate

- `pnpm --filter @openlinker/api type-check`: clean.
- `pnpm check:invariants`: clean, zero findings.
- `npx jest src/sales-documents src/integrations/http/connection.controller.spec.ts`: 5 suites, 60 tests, all pass.

### Verified against the PR body's own claims

The PR body names two decisions a reviewer could dispute. Both check out against the actual code rather than just the description.

**Delivery country vs "billing" (#2518 inherited).** Confirmed at the call site: `toRepresentativeOrderFacts` builds `SalesDocumentOrderFacts` fed to the real `resolveRoutingBatch`, and the country comes from the same field the live evaluator resolves against. No parallel country-resolution logic exists on this branch.

**The synthetic "representative order" used for `GET /sales-documents/markets`.** This deserved scrutiny since it's the riskiest design call in the mini-epic (a bad choice here misreports what real orders get). Read `sales-document-markets.controller.ts` directly: `buyerHasTaxId: undefined` and `totalGross: 0` / `taxTreatment: undefined` - never guessed, never defaulted to a "safe-looking" value. The DTO's doc comment goes further than the PR body and explicitly addresses the sharpest edge case: since `totalGross`/`currency` are non-nullable on `SalesDocumentOrderFacts`, using `0`/`''` as "not asserted" could in principle let an amount-conditioned rule falsely match against zero. The comment states this is handled because `taxTreatment: undefined` triggers the same `net-priced-order` signal a real order with unasserted tax treatment would produce - so the representative order fails the same way a real ambiguous order would, rather than silently reporting false confidence. That is the correct resolution and it is verified in the DTO's own committed documentation, not just asserted in the PR description.

**The `acknowledged` outcome override** is also documented precisely: it is a fourth state the real evaluator structurally cannot produce (an acknowledged market carries no rule/default by construction), so overriding the evaluator's answer here is not a second implementation of routing - it is reporting a fact the evaluator was never asked about.

### Architecture

- Correctly a separate controller from `SalesDocumentRulesController` (rule CRUD vs. a derived read against the orders store).
- Reads `IOrderRecordService` through its published interface, never a repository port across the context boundary.
- `@Roles('admin')` justified in its own doc comment (sole consumer is the admin-only settings page).
- No write path added anywhere in this diff - confirmed by reading the controller; it only calls read methods (`resolveRoutingBatch`, order record reads).

### Not verified in this pass (out of scope for this mini-epic per its own PR body)

- The pre-existing save-time capability-check asymmetry the PR body flags (#2532) - filed, not fixed, correctly scoped as separate work.
- Frontend consumption of these shapes (M6/M7's job, per the PR body).

Nothing here weakens exactly-once fiscal registration, one-originating-document-per-order, or the "never claim what the backend cannot report" rule - if anything, the representative-order design is a good example of that rule being actively defended rather than merely not violated.

Clean to merge into `epic/sales-documents-ux`.

Issue #2540, opened 2026-08-26
[TASK] Frontend — market list rows


norbert-kulus-blockydevs:
> Mockup: `docs/plans/mockups/sales-document-routing.html` — states `#page=settings`

## Problem / Context

The current page stacks five elements per market — an eyebrow, a two-line sentence, a badge, a rule count and a button — at one visual weight, so four markets become eight lines of prose and a broken market looks exactly like a working one. The country, which is what an operator scans for, is the smallest text in its own row.

## Proposed Solution

One row per market: a status dot, the market name as the anchor with its code and rule count demoted, what an order gets, the short reason beneath it when it gets nothing, and one action labelled for what it does. Sort so anything needing a decision comes first. Use the `DocumentKindGlyph` treatment from M5 for the kind.

## Classification

**Type**: Frontend
**Layer**: features/sales-documents
**File(s)**: `apps/web/src/features/sales-documents/components/`

## Dependencies

- Parent mini-epic: #2539
- M4 merged market list
- M5 primitives and copy

## Acceptance Criteria

- [ ] A row shows at most: dot, name and meta, outcome, short reason, one action.
- [ ] Rows needing a decision sort above the rest.
- [ ] The status dot is supplementary: the outcome and reason carry the meaning in text.
- [ ] Below 768 px each market becomes a card.
- [ ] Tests added or updated for non-trivial logic
- [ ] No architecture boundary violations (CORE ↔ Integration)

## Branching and merge target

Branch: `{issue#}-{short-kebab}`, created **from its mini-epic branch**.
PR target: **its mini-epic branch**.

## How this issue gets finished

1. Implement on the branch above.
2. Open the PR into its mini-epic branch.
3. Run `/pr-review <PR>` and post the verdict as a PR comment.
4. Fix **every** finding, and reply under the comment saying what changed for each one.
5. Run `/pr-review` again. Repeat until the review comes back approved with no findings.
6. Only then merge.

Quality gate before every commit: `pnpm lint`, `pnpm type-check`, `pnpm test`.
Commits are signed off (`git commit -s`).

Issue #2541, opened 2026-08-26
[TASK] Frontend — the section summary and the empty state


norbert-kulus-blockydevs:
## Problem / Context

Nothing on the current page states what happens today, and its subtitle explains the mechanism rather than the content. A clean instance shows nothing at all, so an operator cannot tell whether they have finished setting up.

## Proposed Solution

One prose sentence for the whole section, computed: how many markets issue nothing, naming them, and stating that nothing is lost in the meantime. A distinct sentence when nothing is configured but orders are arriving, because a fresh install is not broken. An empty state for a genuinely empty instance, explaining that markets appear as orders arrive. Suppress the summary when there are no rows, so the two do not say the same thing twice.

## Classification

**Type**: Frontend
**Layer**: features/sales-documents
**File(s)**: `apps/web/src/features/sales-documents/components/`

## Dependencies

- Parent mini-epic: #2539
- The market list rows task in this mini-epic.

## Acceptance Criteria

- [ ] The sentence is derived, not hand-written, and names the affected markets.
- [ ] A fresh instance receiving orders reads as "not set up yet", never as broken.
- [ ] The summary and the empty state never both render.
- [ ] The sentence states that nothing is lost while a market is unconfigured.
- [ ] Tests added or updated for non-trivial logic
- [ ] No architecture boundary violations (CORE ↔ Integration)

## Branching and merge target

Branch: `{issue#}-{short-kebab}`, created **from its mini-epic branch**.
PR target: **its mini-epic branch**.

## How this issue gets finished

1. Implement on the branch above.
2. Open the PR into its mini-epic branch.
3. Run `/pr-review <PR>` and post the verdict as a PR comment.
4. Fix **every** finding, and reply under the comment saying what changed for each one.
5. Run `/pr-review` again. Repeat until the review comes back approved with no findings.
6. Only then merge.

Quality gate before every commit: `pnpm lint`, `pnpm type-check`, `pnpm test`.
Commits are signed off (`git commit -s`).

PR #2744, opened 2026-08-31
feat(sales-documents): loading states for the market section (#2543)

### LINKED ISSUE #2543: [TASK] Frontend — loading states for the market section
## Problem / Context

The page has no loading treatment, and the section it needs one for is the one that answers the operator's question, so an empty card during load reads as "nothing is configured".

## Proposed Solution

Skeleton rows sized from the loaded row so the section does not reflow on arrival, plus a narration line in a live region. Disable the actions while the section is busy rather than letting them be pressed against absent data.

## Classification

**Type**: Frontend
**Layer**: features/sales-documents
**File(s)**: `apps/web/src/features/sales-documents/components/`

## Dependencies

- Parent mini-epic: #2539
- The market list rows task in this mini-epic.

## Acceptance Criteria

- [ ] The section does not change height between skeleton and loaded state.
- [ ] The narration is in a live region, not a plain span.
- [ ] Actions are disabled while busy.
- [ ] A blank load is never indistinguishable from an unconfigured instance.

## Branching and merge target

Branch: `{issue#}-{short-kebab}`, created **from its mini-epic branch**.
PR target: **its mini-epic branch**.

## How this issue gets finished

1. Implement on the branch above.
2. Open the PR into its mini-epic branch.
3. Run `/pr-review <PR>` and post the verdict as a PR comment.
4. Fix **every** finding, and reply under the comment saying what changed for each one.
5. Run `/pr-review` again. Repeat until the review comes back approved with no findings.
6. Only then merge.

Quality gate before every commit: `pnpm lint`, `pnpm type-check`, `pnpm test`.
Commits are signed off (`git commit -s`).

norbert-kulus-blockydevs:
## Summary

Replaces the generic `LoadingState` on the market section's initial load with a shape-matched skeleton, and closes the "operator presses an action against stale data" gap during a background refetch.

## What shipped

- `SalesDocumentMarketSectionSkeleton` — 3 placeholder rows sized off `SalesDocumentMarketRow`'s own DOM shape (status dot, identity block, outcome + reason lines, action button) plus a summary-line placeholder, so the section's height doesn't jump when real data arrives. It carries its own `role="status" aria-live="polite"` region with visible + screen-reader text ("Loading markets…"), so a blank load can never be mistaken for the empty state.
- `SalesDocumentMarketSection` returns the skeleton on `marketsQuery.isLoading` instead of the generic `LoadingState`.
- A new `isBusy = marketsQuery.isFetching` derivation covers the "loaded once, fetching again" case: every row's `disabled` prop (already wired from #2540) is driven by it, and a screen-reader-only `aria-live="polite"` region announces "Refreshing markets…". The error state's Retry button also gets `disabled={marketsQuery.isFetching}` — a pre-existing double-click gap that falls naturally under this task's scope.
- CSS: skeleton shimmer keyframe + placeholder bar shapes, scoped under `.sales-document-market-row .skeleton-bar` so the class name can't collide with an unrelated skeleton elsewhere in the app.

## Decisions a reviewer could dispute

- **The skeleton is a real sibling component, not a prop on the row.** Mirrors the existing `DataTableSkeleton` precedent (shape declared by the real component, not a guessed fixed height) rather than inventing a new pattern.
- **The "busy during a background refetch" test mocks the query hook directly** (`vi.doMock` + dynamic re-import) instead of driving a real TanStack Query refetch cycle. This section has no user-facing control that triggers a refetch while data is already loaded (that happens via React Query's own focus/interval refetch, which `renderWithProviders`'s test `QueryClient` disables retries for but doesn't expose a manual trigger for) — simulating one end-to-end would mostly be testing React Query's own scheduling, not this component's `isBusy` derivation. The test comment states this reasoning inline so a future reader doesn't wonder why it's mocked when its siblings aren't.
- **The Retry-button disable is included in this task's diff** even though it's arguably a pre-existing gap from #2540 — it's one line, directly adjacent to the busy-state work, and leaving it would mean an operator could still double-click Retry while a fetch from that exact click was already in flight.

## Test plan

- [x] `pnpm --filter @openlinker/web exec eslint src/features/sales-documents --max-warnings=0` — clean
- [x] `pnpm --filter @openlinker/web exec vitest run src/features/sales-documents` — 19 files / 128 tests passing
- [x] `pnpm check:invariants` (filtered of `.worktrees` noise) — all green, including `design-tokens: OK` (no new CSS vars were needed — every declaration reuses existing tokens)

Closes #2543

### DIFF
--- a/apps/web/src/features/sales-documents/components/sales-document-market-section-skeleton.test.tsx
+++ b/apps/web/src/features/sales-documents/components/sales-document-market-section-skeleton.test.tsx
@@ -1,0 +1,20 @@
+import { render, screen } from '@testing-library/react';
+import { describe, expect, it } from 'vitest';
+import { SalesDocumentMarketSectionSkeleton } from './sales-document-market-section-skeleton';
+
+describe('SalesDocumentMarketSectionSkeleton (#2543)', () => {
+  it('should render a live region announcing that markets are loading', () => {
+    render(<SalesDocumentMarketSectionSkeleton />);
+    const region = screen.getByRole('status');
+    expect(region).toHaveAttribute('aria-live', 'polite');
+    expect(screen.getByText('Loading markets…')).toBeInTheDocument();
+  });
+
+  it('should render placeholder rows shaped like the real row, hidden from assistive tech', () => {
+    const { container } = render(<SalesDocumentMarketSectionSkeleton />);
+    const list = container.querySelector('.sales-document-market-row-list--skeleton');
+    expect(list).not.toBeNull();
+    expect(list).toHaveAttribute('aria-hidden', 'true');
+    expect(container.querySelectorAll('.sales-document-market-row--skeleton').length).toBeGreaterThan(0);
+  });
+});
--- a/apps/web/src/features/sales-documents/components/sales-document-market-section-skeleton.tsx
+++ b/apps/web/src/features/sales-documents/components/sales-document-market-section-skeleton.tsx
@@ -1,0 +1,54 @@
+/**
+ * SalesDocumentMarketSectionSkeleton (#2543)
+ *
+ * Loading treatment for `SalesDocumentMarketSection` — sized from
+ * `SalesDocumentMarketRow`'s own shape (status dot, identity block, outcome
+ * headline + reason line, action button) so the section does not reflow when
+ * the real rows arrive, mirroring `DataTableSkeleton`'s row-shape-declared
+ * approach rather than a fixed row height.
+ *
+ * A blank load must never read as "nothing is configured" (#2543 acceptance):
+ * the skeleton renders a fixed number of placeholder rows plus a visible,
+ * `aria-live="polite"` narration line, so a screen-reader user and a sighted
+ * user both get an explicit "loading" signal rather than silence.
+ *
+ * @module apps/web/src/features/sales-documents/components
+ */
+import type { ReactElement } from 'react';
+
+const SKELETON_ROW_COUNT = 3;
+
+export function SalesDocumentMarketSectionSkeleton(): ReactElement {
+  return (
+    <div className="page-section sales-document-market-section" role="status" aria-live="polite">
+      <span className="sr-only">Loading markets…</span>
+      <span
+        className="sales-document-market-section__summary-skeleton"
+        aria-hidden="true"
+      />
+      <ul
+        className="sales-document-market-row-list sales-document-market-row-list--skeleton"
+        aria-hidden="true"
+      >
+        {Array.from({ length: SKELETON_ROW_COUNT }, (_, index) => (
+          <li key={index} className="sales-document-market-row sales-document-market-row--skeleton">
+            <div className="sales-document-market-row__status">
+              <span className="skeleton-bar skeleton-bar--dot" />
+            </div>
+            <div className="sales-document-market-row__identity">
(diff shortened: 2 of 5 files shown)


norbert-kulus-blockydevs:
## Tech-lead review — PR #2744 (#2543)

### Summary
Replaces the generic `LoadingState` on initial load with a shape-matched `SalesDocumentMarketSectionSkeleton`, and wires `isFetching` through to disable row actions + announce a background refetch. Small, well-scoped, closes the last task in the mini-epic.

### Architecture & boundaries
- ✅ No backend surface touched; purely a frontend loading-state concern.
- ✅ Follows the `DataTableSkeleton` precedent explicitly cited in the header comment — shape declared by the real component rather than a guessed fixed height, matching `docs/frontend-ui-style-guide.md` § Density & Row Heights' insistence that a skeleton's height come from the real row.
- ✅ CSS scoped under `.sales-document-market-row .skeleton-bar` — avoids polluting a bare `.skeleton-bar` global class that a future unrelated skeleton might reuse differently.

### Code quality
- ✅ `isBusy` derivation is a one-line `marketsQuery.isFetching` read, applied uniformly to every row's existing `disabled` prop (already plumbed in #2540) — no new prop surface needed on the row component.
- ✅ Live region text is empty-string when not busy rather than removed conditionally, which is the correct pattern for `aria-live` regions (removing/re-adding the element can suppress the next announcement in some screen readers; toggling text content does not).
- 🟢 **SUGGESTION** — `sales-document-market-section-skeleton.tsx:19`: `SKELETON_ROW_COUNT = 3` is a fine default, but since it's now a second file independently declaring "3 rows," consider a one-line comment noting it's an arbitrary placeholder count (not tied to any real data assumption) so a future reader doesn't wonder if it should track something.

### Naming & structure
- ✅ File name matches the existing `*-skeleton.tsx` naming this repo doesn't otherwise use for sales-documents, but does for `DataTableSkeleton`; consistent enough.
- ✅ Both file headers accurately describe scope and reference #2543 plus sibling issue numbers.

### Tests
- ✅ `sales-document-market-section-skeleton.test.tsx` — live-region text + `aria-hidden` on placeholder content, matching the acceptance criterion "a blank load is never indistinguishable from an unconfigured instance."
- ✅ `sales-document-market-section.test.tsx`'s new "loading states" block explicitly asserts the skeleton and the empty state never coexist — good, since #2541 already established that discipline for the summary/empty-state pair and this extends it correctly to skeleton/empty-state.
- ✅ The busy-state test's choice to mock the hook directly is well-justified and the reasoning is written into the test file itself, not just the PR body — this is the right call given there's genuinely no user-facing trigger for a background refetch in this component today, and testing through a real refetch cycle would mostly exercise React Query's internals.

### Accessibility
- ✅ Skeleton: `role="status" aria-live="polite"` with a real text node, not an empty container — correctly closes the "blank load ≠ nothing configured" gap named in the issue.
- ✅ Busy announcement: separate `sr-only` live region rather than overloading the skeleton's role, since the two states (initial load vs. background refetch) are mutually exclusive branches and don't need to share one region.
- ✅ Retry-button disable during refetch closes a real (if minor) double-click gap; good catch to fold into this task rather than leaving it for a hypothetical future ticket.

### No blocking or important findings.

**Verdict: ✅ Approve.** No BLOCKING or IMPORTANT issues. The one SUGGESTION is optional.

Issue #2542, opened 2026-08-26
[TASK] Frontend — detected markets and the suggested setup


norbert-kulus-blockydevs:
> Mockup: `docs/plans/mockups/sales-document-routing.html` — states `#page=settings` — the "orders arriving, nothing set up" state

## Problem / Context

A clean instance has no routing, and an unconfigured market that orders arrive from is invisible until someone notices no documents were created. There is also no way to offer the one starter template that exists without implying that every market has one.

## Proposed Solution

List detected markets alongside configured ones, each with its order count over the window — the count is what makes an operator act. Render a suggested-setup affordance only where the catalogue has an entry, say explicitly that it is the only market with guidance so far, and give a detected market without one a plain "Set up".

## Classification

**Type**: Frontend
**Layer**: features/sales-documents
**File(s)**: `apps/web/src/features/sales-documents/components/`

## Dependencies

- Parent mini-epic: #2539
- M4 template catalogue and merged list.

## Acceptance Criteria

- [ ] A detected market shows its order count and window.
- [ ] A suggested setup renders only where a template exists.
- [ ] The copy states that guidance exists for that market only.
- [ ] A detected market without a template offers a plain set-up action, not a recommendation.
- [ ] Tests added or updated for non-trivial logic
- [ ] No architecture boundary violations (CORE ↔ Integration)

## Branching and merge target

Branch: `{issue#}-{short-kebab}`, created **from its mini-epic branch**.
PR target: **its mini-epic branch**.

## How this issue gets finished

1. Implement on the branch above.
2. Open the PR into its mini-epic branch.
3. Run `/pr-review <PR>` and post the verdict as a PR comment.
4. Fix **every** finding, and reply under the comment saying what changed for each one.
5. Run `/pr-review` again. Repeat until the review comes back approved with no findings.
6. Only then merge.

Quality gate before every commit: `pnpm lint`, `pnpm type-check`, `pnpm test`.
Commits are signed off (`git commit -s`).

Issue #2539, opened 2026-08-26
[EPIC] M6 — Settings: what each market issues


norbert-kulus-blockydevs:
> Parent: #2513
> Mockup: `docs/plans/mockups/sales-document-routing.html` — states `#page=settings` — the four instance states in the picker: brand new, orders arriving with nothing set up, partly set up, everything issuing

## Scope

The headline half of the settings page. One prose sentence answering the page's own question, then a scannable market list: status dot, market, what an order gets, the short reason when it gets nothing, and one action. Markets issuing nothing sort first. Detected markets appear alongside configured ones, and a suggested setup is offered only where guidance exists.

## Branching

Mini-epic branch `phase/sdux-m6-settings-markets`, created **from `epic/sales-documents-ux`**, opening a PR **into `epic/sales-documents-ux`**.
Each task below gets its own branch off `phase/sdux-m6-settings-markets` and a PR **into `phase/sdux-m6-settings-markets`**.

## Dependencies

- M4 (#2528) merged market list.
- M5 (#2533) reason copy and primitives.

## Acceptance Criteria (mini-epic level)

- [ ] The page states what each market issues, derived from the shipped evaluator, never hand-written.
- [ ] Poland's live-demo configuration is correctly reported as issuing nothing, with its reason.
- [ ] Markets needing a decision sort above working ones.
- [ ] A clean instance shows an empty state that explains markets appear on their own.
- [ ] A detected market is neutral, not an error.
- [ ] Every task branch merged into `phase/sdux-m6-settings-markets`, and `phase/sdux-m6-settings-markets` merged into `epic/sales-documents-ux` via reviewed PR.

PR #2746, opened 2026-08-31
feat(web): sales-document settings page shows which markets issue documents automatically

norbert-kulus-blockydevs:
## Summary

M6 of the sales-document routing-first UX epic (#2513): the headline half of the settings page (#2539). Stacked on `epic/sales-documents-ux` (PR #2710) — merge #2710 first; this PR's visible diff will shrink to just the M6 changes once that lands.

- Market list rows: status dot, market, what the market issues, short reason when it issues nothing, one action (#2540)
- Section summary sentence and empty-state suppression (#2541)
- Detected markets with order counts and suggested setup, offered only where guidance exists (#2542)
- Loading states: shape-matched skeleton, busy-state disabled actions, live-region narration during refetch (#2543)

Markets issuing nothing sort first, per the mockup (`docs/plans/mockups/sales-document-routing.html`, `#page=settings`).

**Mockup accuracy** (`frontend-architecture.md § UX Mockups`): the cited mockup is untouched and still accurate for this change. The shipped row keeps the mockup's anatomy and ordering exactly - status dot, market name as the anchor with code and rule count demoted to meta, what an order here gets, the short reason beneath when it gets nothing, one action, and markets issuing nothing sorted first. The only additions beyond the mockup are states it does not draw (loading skeleton, busy-disabled actions, live-region narration), so nothing it does draw has changed.

## Test plan
- [ ] Merge via **"Create a merge commit"** — not squash — so release-please sees each commit individually

🤖 Generated with Claude Code


norbert-kulus-blockydevs:
## PR Review — M6 (#2539)

Reviewed the true M6-only diff (19 files, ~1400 lines, isolated via `git diff origin/epic/sales-documents-ux...HEAD`) since the raw PR diff is dominated by unmerged M1-M5 content already covered by #2710.

**Verified:**
- FE market types (`sales-document-markets.types.ts`) match `SalesDocumentMarketRowDto`/`SalesDocumentMarketOutcomeDto` field-for-field, including the `acknowledged` outcome kind the evaluator itself cannot produce.
- `describeSalesDocumentMarketOutcome` never re-derives a routing decision client-side — reads only the backend's resolved `outcome`.
- Decisions-first sort, sole-templated-market caption exclusivity, summary/empty-state mutual exclusion, and the busy-state/live-region skeleton behaviour all match their stated acceptance criteria and have direct test coverage.
- Colour-marks-exceptions rule respected (only `needsDecision" rows get a warning tone/outline).
- Mobile card layout via CSS grid reflow, no second component.
- All new components are consumed only within the same feature (no barrel export needed).

**Gate:** `pnpm --filter @openlinker/web type-check` clean, ESLint clean, 19 files / 128 tests green in `vitest`, full `pnpm check:invariants` green including the sales-document reason-copy mirror.

No blocking or important findings. Approved.

Note: this PR stays open until M7/M8/M9 also land, per the stacked-PR plan — merging in sequence once all four are reviewed.

piotrswierzy (commented):
## Tech Lead review — M6 sales-document market section

Reviewed all 19 files (FE-only). Verified: FE types mirror `SalesDocumentMarketRowDto`/`SalesDocumentMarketsResponseDto` field-for-field; feature layering is clean (`features → shared`, no raw `fetch`, query via `useApiClient` + TanStack Query); mutations already invalidate `salesDocumentRulesQueryKeys.all`, which covers the new `markets()` key; outcome copy is derived from the backend decision, never re-derived; summary vs empty state are mutually exclusive as claimed; CSS is token-only, no raw hex. Good unit coverage on all four pure libs plus the row/section/skeleton.

No BLOCKING issues.

### IMPORTANT

**1. Skeleton shimmer ignores `prefers-reduced-motion`** — `apps/web/src/index.css` (`.sales-document-market-row .skeleton-bar`, new block). The repo's own skeleton primitive guards its identical shimmer (`.data-table-skeleton__bar`, index.css:3418) with `@media (prefers-reduced-motion: reduce) { animation: none; }`. Add the same guard for `sales-document-market-skeleton-shimmer`.

**2. "1 orders"** — `sales-document-market-row.tsx`, meta line: `` ` · ${row.orderCount} orders...` ``. The backend DTO states a detected market always has **at least one** order, so `1` is reachable and reads "1 orders". `ruleCount` right beside it is pluralised correctly. Fix: `row.orderCount === 1 ? '1 order' : \`${row.orderCount} orders\``.

**3. Screen-reader status text says "Set up" for a settled row** — `sales-document-market-row.tsx`: `<span className="sr-only">{copy.needsDecision ? 'Needs attention' : 'Set up'}</span>`. For an issuing/acknowledged market this announces an imperative, and it collides with the *action* button label "Set up" used on unresolved rows — the same words mean opposite things. Use a settled-state noun (e.g. `'Configured'` / `'No action needed'`).

### SUGGESTION

- `index.css` mobile block declares `grid-template-areas` but no child sets `grid-area` — the declaration is inert (single-column auto-placement already gives the intended order). Drop it or assign the areas.
- `.sales-document-market-section__summary--fresh-install` sets the same `color` as the base rule — dead modifier.
- The new `.skeleton-bar--*` set duplicates `data-table-skeleton__bar--*`. Style guide asks to extend a shared primitive before adding page-local styling; consider reusing the existing bar class.
- `sales-document-market-row.tsx`: `const isDetectedOnly = row.orderCount !== null` — it means "detected", not "detected-only" (a configured market can also have orders). Rename to `isDetected`.
- `frontend-architecture.md § UX Mockups` asks an implementing PR to either touch the cited mockup or state why it is still accurate. The description cites `sales-document-routing.html#page=settings` but does neither explicitly — one line would close it.
- New row `min-height: 4.25rem` isn't registered in the style guide's Density & Row Heights table. Worth an entry (it's a list row, not a `DataTable` row, but the guide asks for the registration).

norbert-kulus-blockydevs:
Thanks Piotr - all three IMPORTANT findings and every SUGGESTION are addressed in f062feb9e.

### IMPORTANT

**1. Skeleton shimmer ignores `prefers-reduced-motion`** - fixed, and fixed by removing the duplication rather than adding a second guard. `.sales-document-market-row .skeleton-bar` now extends the shared `.data-table-skeleton__bar` primitive (both the base rule and its existing `@media (prefers-reduced-motion: reduce)` guard list the market selector), so the page-local block keeps only the per-slot sizing. The duplicated `sales-document-market-skeleton-shimmer` keyframes are deleted. That closes SUGGESTION 3 in the same change - there is now one shimmer definition, not two that can drift.

**2. "1 orders"** - fixed: `row.orderCount === 1 ? '1 order' : \`${row.orderCount} orders\``, matching how `ruleCount` beside it already reads. Regression test added (`should render a single detected order in the singular`), which also asserts `1 orders` is absent.

**3. Screen-reader status text says "Set up" for a settled row** - now `'No action needed'`. Test added asserting the settled row announces that and that the bare string `Set up` does not appear on a row that needs no action, so the status text can never re-collide with the action button's label.

### SUGGESTION

- **Inert `grid-template-areas`** - dropped. Single-column auto-placement already gives status / identity / outcome / action in that order.
- **Dead `--fresh-install` modifier** - removed (the tone value stays in `summarize-sales-document-markets.ts`; only the no-op CSS rule is gone).
- **Duplicated `.skeleton-bar--*` set** - handled by extending the shared primitive, see IMPORTANT 1.
- **`isDetectedOnly` -> `isDetected`** - renamed; you are right that a configured market can carry orders too.
- **Mockup accuracy** - stated explicitly in the PR description now. The mockup is untouched and still accurate: the shipped row keeps its anatomy and ordering exactly; the only additions are states the mockup does not draw (loading skeleton, busy-disabled actions, live-region narration).
- **Row height not registered** - added to the style guide's Density & Row Heights table as `4.25 rem` min, with a note that it is a settings list row rather than a `DataTable` row and why the floor exists (a one-line row must not be shorter than one carrying a reason line, or the list steps as rows resolve).

Gate: `apps/web` type-check clean, ESLint clean on `features/sales-documents`, all 19 sales-document test files / 130 tests pass.

piotrswierzy (approved):
## Re-review at `1359c9bb` — ✅ Approve

Every finding from the `110e458` pass is addressed in `f062feb` ("fix(web): address M6 market-section review findings").

### Prior IMPORTANT — all fixed, and each fix is pinned by a test or a doc

1. **Reduced-motion guard on the skeleton shimmer** — `.sales-document-market-row .skeleton-bar` now *extends* the shared `data-table-skeleton__bar` primitive (`index.css:3357`, and the `prefers-reduced-motion: reduce` block at `:3419`), and the duplicated page-local `@keyframes sales-document-market-skeleton-shimmer` is deleted. The page-local rule keeps only per-slot sizing. This is the right shape — one animation, one guard, no second copy to drift.
2. **`"1 orders"`** — `sales-document-market-row.tsx:114` now branches on `row.orderCount === 1`, and `sales-document-market-row.test.tsx` asserts both `1 order in the last 30 days` and the absence of `1 orders`.
3. **`sr-only` status colliding with the action label** — a settled row announces `No action needed` (`:103`). The new test additionally asserts `queryByText(/^Set up$/)` is absent on a settled row, so a regression that re-introduces the collision fails rather than merely looking wrong.

### Prior SUGGESTIONs — all taken

- Inert mobile `grid-template-areas` dropped; `--fresh-install` no-op modifier dropped.
- Duplicated skeleton-bar styling collapsed into the shared primitive (same change as #1).
- `isDetectedOnly` → `isDetected` (`:89`) — a configured market can carry orders too.
- Row height registered: `docs/frontend-ui-style-guide.md` § Density & Row Heights now carries the `.sales-document-market-row` `4.25rem` floor entry with its rationale.
- Mockup-sync obligation (`frontend-architecture.md § UX Mockups`) now discharged explicitly in the PR description — states the mockup is untouched and why it is still accurate (the additions are states it does not draw).

### Fresh pass on the current diff — no new BLOCKING/IMPORTANT

Architecture and layering are clean: everything lives inside `features/sales-documents` (no `shared → features` crossing), server state stays in TanStack Query behind one merged read so summary/list/detected rows cannot disagree, no raw `fetch`, tokens only, no `any` beyond the one documented wire-boundary cast in `sales-document-market-outcome-copy.ts:77`.

Two small optional notes, neither worth a round trip:

- `sales-document-market-section-skeleton.tsx:26` → `.sales-document-market-section__summary-skeleton` (`index.css:19475`) is a flat `--bg-surface-muted` block while the row bars beneath it shimmer, so the loading surface animates unevenly. Adding it to the `data-table-skeleton__bar` selector group would make it consistent and inherit the same guard.
- `isBusy = marketsQuery.isFetching` (`sales-document-market-section.tsx:98`) also fires on React Query's default `refetchOnWindowFocus` past the 30 s `staleTime`, so returning to the tab briefly disables every row action and announces "Refreshing markets…". That is arguably the documented intent; flagging only so the behaviour is a choice.

norbert-kulus-blockydevs:
Took both optional notes from the approval pass, in `6e22da4`:

- **Summary skeleton shimmer** — `.sales-document-market-section__summary-skeleton` now joins the `.data-table-skeleton__bar` / `.sales-document-market-row .skeleton-bar` selector group (both the base shimmer rule and its `prefers-reduced-motion` guard), instead of being a flat `--bg-surface-muted` block. Only per-slot sizing stays page-local, same shape as the row bars.
- **`isBusy` on window-focus refetch** — left the behaviour as-is and added a comment on `isBusy = marketsQuery.isFetching` explaining why it's deliberate: a routing decision here is a fiscal fact, so an action must not stay clickable while a background refetch might change it under the operator.

`pnpm --filter @openlinker/web test` — 19 files / 130 tests still green.

Issue #591, opened 2026-05-09
[F4] [HIGH] Connection entity exposed via deep path and via barrel — two valid surfaces


piotrswierzy:
Parent: #552 — Modularity Thread F. Tracking: #546. Doc: PR #545.

**Severity**: HIGH · **Catalog**: SDK-5

**Observation**: The `Connection` domain entity is reachable both as `import { Connection } from '@openlinker/core/identifier-mapping'` (barrel-exported) and as `import { Connection } from '@openlinker/core/identifier-mapping/domain/entities/connection.entity'` — both forms appear in real adapter code today. The package.json exports field actually permits this via the `./identifier-mapping/*` glob (`libs/core/package.json:30-34`).

**Evidence**: De-facto SDK survey shows both shapes in use across `libs/integrations/prestashop/`. Test specs use the deep form (e.g. `libs/integrations/allegro/src/infrastructure/adapters/__tests__/allegro-offer-manager.adapter.spec.ts:13`).

**Why it blocks OSS plugin contribution**: Two import paths means two backwards-compat liabilities. Once `Connection` moves (say, into a new `connections` bounded context — likely as the system grows), every plugin using the deep path breaks.

**Recommendation**: Tighten `libs/core/package.json` exports to NOT include the `./*/*` wildcard subpath (drop `./identifier-mapping/*`, `./integrations/*`, etc.). Force consumers through the barrel. Add an ESLint rule (`no-restricted-imports`) preventing `@openlinker/core/*/domain/`, `@openlinker/core/*/application/`, `@openlinker/core/*/infrastructure/` patterns in `libs/integrations/*`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2747, opened 2026-08-31
fix(web/connections): render system-owned sync jobs as "System", not "Unknown"

### LINKED ISSUE #2745: [BUG] Frontend — Sync Jobs connection column renders system jobs as "Unknown"
## Problem / Context

The Sync Jobs list (`/jobs-logs`) and job detail page render the `Connection` column via `ConnectionEntityLabel` (`apps/web/src/features/connections/components/ConnectionEntityLabel.tsx`), which always fetches `GET /connections/:id` to resolve a display name.

Some sync jobs have no real operator connection — e.g. `analytics.currency.*` sweeps, or the deletion-triggered offer pause enqueued by `apps/worker/src/events/master-deletion-to-job.handler.ts`. These are stamped with the backend's all-zero placeholder id, `00000000-0000-0000-0000-000000000000` (`SYSTEM_CONNECTION_ID` in `libs/core/src/inventory/application/services/inventory.service.ts` and `apps/worker/src/events/master-deletion-to-job.handler.ts`).

`ConnectionEntityLabel` has no special case for that id, so it fetches it, gets a 404, and `EntityLabel` (`apps/web/src/shared/ui/entity-label.tsx`) falls back to rendering `Unknown` — the exact same fallback shown for a genuinely deleted or broken connection. An operator looking at Sync Jobs can't tell "this job isn't tied to any connection, by design" apart from "this job's connection is broken."

## Proposed Solution

Give the frontend its own `SYSTEM_CONNECTION_ID` constant (mirroring the backend one) and special-case it in `ConnectionEntityLabel`:

- Add `SYSTEM_CONNECTION_ID = '00000000-0000-0000-0000-000000000000'` to `apps/web/src/features/connections/api/connections.types.ts`, exported from `apps/web/src/features/connections/index.ts`.
- In `ConnectionEntityLabel`, when `connectionId === SYSTEM_CONNECTION_ID`, skip the `useConnectionQuery` fetch entirely and render a distinct `System` label — no id shown, no link to `/connections/:id` (there's nothing to link to).
- Add a test case in `apps/web/src/features/connections/components/ConnectionEntityLabel.test.tsx` covering the system-id path (no fetch triggered, renders `System`, not `Unknown`, no link).

No backend change needed — `SYSTEM_CONNECTION_ID` already exists server-side; this only teaches the frontend to recognize it.

## Classification

**Type**: Frontend
**Layer**: Interface (shared UI component)
**File(s)**:
- `apps/web/src/features/connections/api/connections.types.ts`
- `apps/web/src/features/connections/index.ts`
- `apps/web/src/features/connections/components/ConnectionEntityLabel.tsx`
- `apps/web/src/features/connections/components/ConnectionEntityLabel.test.tsx`

## Dependencies

- None

## Assumptions

- The all-zero id is the only system-job placeholder in use today; if a second system placeholder is introduced later it isn't covered by this fix.
- No visual/UX design change beyond replacing "Unknown" with "System" — no new shared component, since `EntityLabel`'s existing name-rendering path already supports this via a plain string.

## Acceptance Criteria

- [ ] `ConnectionEntityLabel` renders `System` (no id, no link, no network request) when `connectionId` is the all-zero placeholder
- [ ] Sync Jobs list page and Sync Job detail page both reflect the fix (shared component, no page-level change needed)
- [ ] A real connection's row is unaffected (still fetches and links normally)

norbert-kulus-blockydevs:
## Summary
- `ConnectionEntityLabel` fetched `GET /connections/:id` for every `connectionId`, including the backend's all-zero `SYSTEM_CONNECTION_ID` placeholder stamped on jobs with no real connection. There are exactly two stampers: `InventoryService`'s dormant fallback when no source connection is supplied (dormant since #2609 gave `inventory.propagateToMarketplaces` the master connection id), and `MasterDeletionToJobHandler`, which stamps it on every `marketplace.offer.pauseStale` job — a job that genuinely has no owning connection. That fetch always 404s, so the Sync Jobs list/detail pages rendered `Unknown` - indistinguishable from a genuinely deleted or broken connection.
- Added a frontend `SYSTEM_CONNECTION_ID` constant mirroring the backend one, and special-cased it in `ConnectionEntityLabel`: skip the fetch, render `System` (no id, no link). `ConnectionCell` suppresses its line-2 copyable id + status note for the same id.
- The constant is hardcoded independently in all three places (the browser bundle cannot import `@openlinker/core`, #591), so `scripts/check-system-connection-id-mirror.mjs` under `pnpm check:invariants` fails the build if any copy drifts.

## Test plan
- [x] `pnpm --filter @openlinker/web exec vitest run src/features/connections/components/ConnectionEntityLabel.test.tsx src/features/connections/components/ConnectionCell.test.tsx` - passing, including the new system-id cases
- [x] `tsc --noEmit` on `apps/web` - clean
- [x] `eslint` on the changed files - clean
- [x] `pnpm check:invariants` - green, including the new mirror check

Closes #2745


norbert-kulus-blockydevs:
## Code Review — PR #2747

**Scope**: `connections.types.ts`, `ConnectionEntityLabel.tsx`, `ConnectionEntityLabel.test.tsx`, `features/connections/index.ts` (47 additions / 1 deletion).

### Summary
Small, well-contained fix. `ConnectionEntityLabel` now special-cases the backend's all-zero `SYSTEM_CONNECTION_ID` placeholder (verified byte-identical to both backend definitions — `libs/core/src/inventory/application/services/inventory.service.ts:26` and `apps/worker/src/events/master-deletion-to-job.handler.ts:56`) and renders `System` instead of fetching a connection that will always 404.

### Correctness checks
- **Hooks-rule safe**: `useConnectionQuery` is still called unconditionally before the early `isSystem` return, so no conditional-hook violation. `enabled: !nameSupplied` correctly disables the fetch for both the pre-existing "caller-supplied name" case and the new system case — confirmed by the new test asserting `getById` is never called.
- **Existing call sites unaffected**: both `sync-jobs-page.tsx` (list + card view) and `sync-job-detail-page.tsx` call `ConnectionEntityLabel` without a `name` prop, so they hit the fetch path today and will hit the new system branch correctly once `job.connectionId` is the placeholder — this is exactly the bug path described in the PR.
- **Verified locally** (built worktree, not just reading the diff): `vitest run ConnectionEntityLabel.test.tsx` → **11/11 pass**; `eslint` on all 4 changed files → clean.
- **No other consumer regressed**: grep'd every `ConnectionEntityLabel` usage — none pass `showId`/`showCopy` expecting them to apply to the system row, since the system branch is a terminal, non-fetching render.

### Minor observation (non-blocking)
`ConnectionEntityLabel`'s system branch hardcodes `showId={false}` / `showCopy={false}`, silently overriding any `showId`/`showCopy` prop a caller passes (e.g. `sync-jobs-page.tsx:55` passes `showId` explicitly). This is almost certainly intentional — there's no real id to show or copy for a system-attributed job — but worth a one-line comment noting the override is deliberate, since a future caller relying on `showId` for a different reason (e.g. a stable column layout) could be surprised. Not a regression risk, not blocking.

### Architecture / standards
- Correctly uses the FE feature barrel export pattern (`features/connections/index.ts`) rather than a deep import (`docs/frontend-architecture.md § Feature Public Surface`).
- Constant lives in `*.types.ts` alongside the other connection type/const exports — consistent with existing file conventions.
- Test follows Arrange-Act-Assert and asserts the negative case (`getById` not called) rather than just the positive render, matching `docs/testing-guide.md` guidance.

### Verdict
✅ **Approve** — no regressions found; fix is narrowly scoped, verified against both backend constant definitions, and covered by a passing test.

norbert-kulus-blockydevs:
## Code Review — PR #2747

**Files**: `connections.types.ts`, `ConnectionEntityLabel.tsx`, `ConnectionEntityLabel.test.tsx`, `features/connections/index.ts`

### Summary
Correct, minimal fix. `ConnectionEntityLabel` now special-cases the backend's all-zero `SYSTEM_CONNECTION_ID` placeholder and renders `System` instead of firing a fetch that always 404s. No architecture violations, no `any`, naming matches convention, barrel export used correctly instead of a deep import, test follows Arrange-Act-Assert and asserts the negative case (`getById` not called).

### Findings

**1. Third independent hardcoded copy of the magic UUID — no drift guard (non-blocking)**
`SYSTEM_CONNECTION_ID = '00000000-0000-0000-0000-000000000000'` is now defined independently in three places with no shared source and no automated check tying them together:
- `libs/core/src/inventory/application/services/inventory.service.ts:26` (private class field)
- `apps/worker/src/events/master-deletion-to-job.handler.ts:56` (module-level const)
- `apps/web/src/features/connections/api/connections.types.ts:28` (this PR)

This repo has an established convention for exactly this shape of problem — a frontend value that must stay byte-identical to a backend-owned one gets a `scripts/check-*-mirror.mjs` invariant under `pnpm check:invariants` (e.g. `check-parameter-restriction-mirror.mjs`, `check-shipping-tax-split-mirror.mjs`, `check-sales-document-reason-mirror.mjs`). Here, if either backend constant is ever changed, the frontend special-case silently stops matching and the row reverts to "Unknown" with zero build-time signal. Not blocking since the backend itself doesn't currently unify its own two copies either, but worth a follow-up — either export the constant from a shared location the backend two call sites both consume, or add a lightweight mirror check.

**2. `ConnectionCell` doesn't inherit the "no id" treatment for a system row (non-blocking, currently unreached)**
`ConnectionCell` (shared by Orders/Shipments/Listings/Products/Customers/Webhooks) delegates line 1 to `ConnectionEntityLabel`, so it *will* correctly render "System" if a system-owned `connectionId` ever reaches it. But `ConnectionCell`'s line 2 unconditionally renders `CopyableId id={connectionId}` — a copyable all-zero UUID — which contradicts the `showId={false}` / `showCopy={false}` intent established in `ConnectionEntityLabel`'s new branch. No known caller currently feeds `ConnectionCell` a system-owned id (this PR only touches Sync Jobs pages, which call `ConnectionEntityLabel` directly), so this isn't a live regression — just worth a follow-up if the pattern spreads to a page rendered through `ConnectionCell`.

### Verdict
✅ **Approve** — no blocking issues. Both findings above are pre-existing-pattern / not-yet-reached and can be follow-ups rather than gates on this PR.

norbert-kulus-blockydevs:
Fixed both review findings, and restored the first review's body — it had gotten mangled into a literal `@/tmp/old_comment_edited.md` string by a prior edit attempt; its actual content is now visible above.

**Review 1** finding (showId/showCopy override):
- `ConnectionEntityLabel`: added a comment on the system branch noting the `showId={false}` / `showCopy={false}` override is deliberate (there's nothing real to show/copy for the placeholder), not an accidental prop drop.

**Review 2** findings:
1. **SYSTEM_CONNECTION_ID drift guard** — added `scripts/check-system-connection-id-mirror.mjs`, wired into `pnpm check:invariants`, comparing the three independently-hardcoded copies (`inventory.service.ts`, `master-deletion-to-job.handler.ts`, `connections.types.ts`). Any future edit to one without the others now fails the build instead of silently reverting the row to "Unknown".
2. **`ConnectionCell` copyable-id leak** — line 2 previously rendered a copyable all-zero UUID + status note for a system-owned connection id, contradicting line 1's `showId={false}`/`showCopy={false}` treatment. `ConnectionCell` now special-cases `SYSTEM_CONNECTION_ID` the same way `ConnectionEntityLabel` does: suppresses line 2 entirely and skips its own `useConnectionQuery` fetch for that id (same wasted-404 pattern this PR originally fixed, just in the sibling component).

Added a `ConnectionCell` test covering the new case. All 26 tests across both components pass; `tsc --noEmit` and `eslint` on the changed files are clean.

piotrswierzy (approved):
## Tech Lead review — PR #2747

✅ **Approve** — no BLOCKING or IMPORTANT findings.

### What I verified
- **Mirror is real, not aspirational.** Both backend copies exist and are byte-identical to the new FE constant: `libs/core/src/inventory/application/services/inventory.service.ts:28` (documented dormant fallback since #2609) and `apps/worker/src/events/master-deletion-to-job.handler.ts:56` (live — `marketplace.offer.pauseStale`). Ran `scripts/check-system-connection-id-mirror.mjs --self-check` (passes) and exercised `parseStringConst` against the PR's actual `connections.types.ts` — it correctly skips the prose mention of `SYSTEM_CONNECTION_ID` in the docblock and extracts the literal. A missing/renamed constant exits 1 loudly rather than passing vacuously.
- **Hook-order safety.** `useConnectionQuery` is still called unconditionally before the `isSystem` early return in both components; `enabled: !nameSupplied` / `!factsSupplied` suppresses the fetch. Both tests assert `getById` is never called — the right negative assertion.
- **Blast radius.** `ConnectionEntityLabel` is reached from `sync-jobs-page.tsx` and `sync-job-detail-page.tsx` without a `name` prop, i.e. exactly the 404 path #2745 describes. `ConnectionCell`'s line-2 suppression closes the sibling leak (a copyable all-zero UUID + a status note describing a connection that does not exist) before it can be reached.
- **Standards.** Barrel export used rather than a deep import (`frontend-architecture.md § Feature Public Surface`); constant colocated with the other connection consts in `*.types.ts`; new invariant follows the established `check-*-mirror.mjs` + `--self-check` shape and is wired into `check:invariants`. All 8 CI checks green.

### SUGGESTION — `scripts/check-system-connection-id-mirror.mjs`
The module calls `await main()` at top level unconditionally, so any future `import` of the exported `parseStringConst` / `diff` (e.g. from a unit test) executes the whole check as a side effect and can `process.exit(1)`. Guard it the way the sibling scripts' authors would if they exported helpers — e.g. only run `main()` when `import.meta.url` matches `process.argv[1]`.

### SUGGESTION — PR description
The summary cites "currency-reporting sweeps" as a stamper of the placeholder; grep finds no such use — the two real sources are the inventory propagation fallback and the deletion-triggered offer pause. Worth correcting so the next reader does not go looking for a third site.

### SUGGESTION — docs
`architecture-overview.md` documents most `check:invariants` mirrors in prose beside the invariant they protect. A one-line mention next to the #2609 propagation-scope paragraph would keep the new guard discoverable.

norbert-kulus-blockydevs:
Thanks for the review — all three SUGGESTIONs are addressed in d6f66f7f4.

**1. `check-system-connection-id-mirror.mjs` top-level `await main()`** — fixed. `main()` now runs only when the module is the process entrypoint:

```js
if (process.argv[1] && import.meta.url === pathToFileURL(process.argv[1]).href) {
  await main();
}
```

Used `pathToFileURL` rather than a `file://${process.argv[1]}` template because the latter breaks on spaces/escaping/Windows paths — same shape (and same comment) as `scripts/create-adapter.mjs:297`, the one existing entrypoint guard in `scripts/`. Verified three ways: `--self-check` passes, the real check passes (`3 copies in sync`), and `import('./scripts/check-system-connection-id-mirror.mjs')` now resolves with no side effect, exporting just `{ diff, parseStringConst }`.

**2. PR description "currency-reporting sweeps"** — correct, that was wrong; grep confirms only two stampers. Description rewritten to name them precisely: `InventoryService`'s fallback (and that it has been dormant since #2609 gave `inventory.propagateToMarketplaces` the master connection id) and `MasterDeletionToJobHandler`'s `marketplace.offer.pauseStale` enqueue, which is the live one.

**3. Docs** — added a paragraph to `architecture-overview.md` inside the `fan-out`/#2609 propagation-scope bullet, right after the "cap on database reads" sentence. It records that the placeholder survives as a dormant fallback plus a live one on the pause path, that the FE now renders `System` instead of a 404-driven `Unknown`, and that the three copies are hardcoded independently (browser bundle cannot import `@openlinker/core`, #591) which is why the mirror check exists — stating the failure mode it prevents, i.e. the special-case silently ceasing to fire and the row quietly reverting to `Unknown`.

`pnpm check:invariants` is green on the branch, new check included.

Issue #2550, opened 2026-08-26
[TASK] Frontend — providers table keeps its editable controls


norbert-kulus-blockydevs:
## Problem / Context

This table is the only place document kind, goes-first and trigger are set, and a connection with no document kind is not a routing candidate at all. The redesign's first draft demoted it to a read-only inventory, which would have left an operator unable to configure the thing the page configures.

## Proposed Solution

Keep the three controls editable, demoted below the market section but not stripped. Show connection status including needs-sign-in with its consequence, and state that only one provider may go first across all of them, whatever it issues.

## Classification

**Type**: Frontend
**Layer**: features/sales-documents
**File(s)**: `apps/web/src/features/sales-documents/components/sales-documents-panel.tsx`

## Dependencies

- Parent mini-epic: #2544
- M4 routing writes.

## Acceptance Criteria

- [ ] Document kind, goes-first and trigger remain editable.
- [ ] A connection needing re-authentication says it cannot issue.
- [ ] The one-primary rule is stated where the control is.
- [ ] A connection with nothing set is visibly not a routing candidate.
- [ ] Tests added or updated for non-trivial logic
- [ ] No architecture boundary violations (CORE ↔ Integration)

## Branching and merge target

Branch: `{issue#}-{short-kebab}`, created **from its mini-epic branch**.
PR target: **its mini-epic branch**.

## How this issue gets finished

1. Implement on the branch above.
2. Open the PR into its mini-epic branch.
3. Run `/pr-review <PR>` and post the verdict as a PR comment.
4. Fix **every** finding, and reply under the comment saying what changed for each one.
5. Run `/pr-review` again. Repeat until the review comes back approved with no findings.
6. Only then merge.

Quality gate before every commit: `pnpm lint`, `pnpm type-check`, `pnpm test`.
Commits are signed off (`git commit -s`).

Issue #2549, opened 2026-08-26
[TASK] Frontend — reset behind a confirm, and the read-only role


norbert-kulus-blockydevs:
## Problem / Context

Reset destroys every rule and default for a market, and today it is a neutral secondary button in the dialog footer with no confirmation, sitting beside the primary action and reflowing the footer when expanded. Separately, a viewer or demo session sees every control as usable.

## Proposed Solution

Move reset out of the footer, give it the danger tone and put it behind `ConfirmDialog` naming what it destroys. Render the write-gated variant for a session without the permission, using the existing access primitives rather than a new pattern.

## Classification

**Type**: Frontend
**Layer**: features/sales-documents
**File(s)**: `apps/web/src/features/sales-documents/components/`, `apps/web/src/shared/ui/`

## Dependencies

- Parent mini-epic: #2544

## Acceptance Criteria

- [ ] Reset is danger-toned and confirmed, and names what is destroyed.
- [ ] It no longer sits beside the primary action.
- [ ] A session without write permission sees the gated variant.
- [ ] The gating uses the existing access primitives.
- [ ] Tests added or updated for non-trivial logic
- [ ] No architecture boundary violations (CORE ↔ Integration)

## Branching and merge target

Branch: `{issue#}-{short-kebab}`, created **from its mini-epic branch**.
PR target: **its mini-epic branch**.

## How this issue gets finished

1. Implement on the branch above.
2. Open the PR into its mini-epic branch.
3. Run `/pr-review <PR>` and post the verdict as a PR comment.
4. Fix **every** finding, and reply under the comment saying what changed for each one.
5. Run `/pr-review` again. Repeat until the review comes back approved with no findings.
6. Only then merge.

Quality gate before every commit: `pnpm lint`, `pnpm type-check`, `pnpm test`.
Commits are signed off (`git commit -s`).

Issue #2548, opened 2026-08-26
[TASK] Frontend — fall-through disclosure, stated per market


norbert-kulus-blockydevs:
## Problem / Context

The current dialog explains that an unmatched order falls through to Rest of world. That is true only for a market with no rules and no default. Inside a configured market it is false: the order is held. The page tells the operator their fallback is a safety net when it is unreachable.

## Proposed Solution

Make the disclosure conditional on the market's actual configuration: fall-through where the market has neither rules nor a default, held otherwise, and say which. State once that Rest of world is reached only by a market with no configuration at all.

## Classification

**Type**: Frontend
**Layer**: features/sales-documents
**File(s)**: `apps/web/src/features/sales-documents/components/`

## Dependencies

- Parent mini-epic: #2544
- #2501

## Acceptance Criteria

- [ ] A configured market states that an unmatched order is held.
- [ ] An unconfigured market states that it falls through.
- [ ] One naming is used for that market throughout the page.
- [ ] A test covers both branches.
- [ ] Tests added or updated for non-trivial logic
- [ ] No architecture boundary violations (CORE ↔ Integration)

## Branching and merge target

Branch: `{issue#}-{short-kebab}`, created **from its mini-epic branch**.
PR target: **its mini-epic branch**.

## How this issue gets finished

1. Implement on the branch above.
2. Open the PR into its mini-epic branch.
3. Run `/pr-review <PR>` and post the verdict as a PR comment.
4. Fix **every** finding, and reply under the comment saying what changed for each one.
5. Run `/pr-review` again. Repeat until the review comes back approved with no findings.
6. Only then merge.

Quality gate before every commit: `pnpm lint`, `pnpm type-check`, `pnpm test`.
Commits are signed off (`git commit -s`).

PR #2756, opened 2026-08-31
fix(web): correct sales-document settings copy against the shipped routing evaluator

norbert-kulus-blockydevs:
## Summary

M7 of the sales-document routing-first UX epic (#2513): the authoring half of the settings page (#2544). Independent of PR #2746 (M6) - both branch from `epic/sales-documents-ux` and target `main` directly.

- Routing dialog: explicit "Done" close action, a "saves as you go" note, and per-control pending/error state on rule delete and default upsert, which previously failed silently (#2545)
- Rules section: states the real evaluator rule - exactly one match wins, no priority, a tie holds the order - and flags rules that read the buyer's tax ID, since that condition is live for a PrestaShop-sourced order and unreachable for Allegro/WooCommerce (#2599 made the field real for PrestaShop only) (#2546)
- Defaults: corrected the two-defaults warning to state the real mechanism (the default tier is skipped entirely when both are set, not "decided by some other rule") (#2547)
- Fall-through: the copy is now conditional on whether the country itself is configured - a configured country's unmatched order is held, never passed to Rest of world; only a country with no rules and no default at all falls through (#2548)
- Reset: gated by write-access visibility, so an unauthorized session sees no reset control at all rather than a disabled one (#2549)
- Providers table: corrected two false claims ("one radio group across all rows" when the primary selector is per-connection state, and a blanket capability statement), added `needs_reauth` and unset-row copy, and demoted the table below the rule-engine panel, which is now the page's primary authoring surface (#2550)

## Test plan

🤖 Generated with Claude Code

### DIFF
--- a/apps/web/src/features/sales-documents/components/sales-document-country-defaults.tsx
+++ b/apps/web/src/features/sales-documents/components/sales-document-country-defaults.tsx
@@ -14,6 +14,7 @@ import { useConnectionsQuery } from '../../connections';
 import { selectInvoicingCandidates } from '../../invoicing';
 import { selectFiscalizationCandidates } from '../../fiscalization';
 import { Select } from '../../../shared/ui/select';
+import { Alert } from '../../../shared/ui/alert';
 import { LoadingState, ErrorState } from '../../../shared/ui/feedback-state';
 import { ReadOnlyLock } from '../../../shared/ui/read-only-lock';
 import { useWriteAccess } from '../../../shared/auth/use-permission';
@@ -74,24 +75,36 @@ export function SalesDocumentCountryDefaults({
               ? selectInvoicingCandidates(connections)
               : selectFiscalizationCandidates(connections);
           const current = defaults.find((d) => d.documentKind === documentKind) ?? null;
+          const isSavingThisKind =
+            upsert.isPending && upsert.variables?.documentKind === documentKind;
+          const isRemovingThisKind =
+            remove.isPending && current !== null && remove.variables === current.id;
+          const isPendingThisKind = isSavingThisKind || isRemovingThisKind;
+          const saveFailedForThisKind =
+            upsert.isError && upsert.variables?.documentKind === documentKind ? upsert.error : null;
+          const removeFailedForThisKind =
+            remove.isError && current !== null && remove.variables === current.id
+              ? remove.error
+              : null;
 
           return (
             <div key={documentKind} className="page-section">
               <label className="eyebrow" htmlFor={`sd-default-${documentKind}`} style={{ marginBottom: 2 }}>
                 {label}
+                {isPendingThisKind ? ' — Saving…' : null}
               </label>
               <ReadOnlyLock active={write.demoReadOnly} message={DEMO_READ_ONLY_ACTION_MESSAGE}>
                 <Select
                   id={`sd-default-${documentKind}`}
                   value={current?.connectionId ?? ''}
-                  disabled={!write.canWrite || candidates.length === 0}
+                  disabled={!write.canWrite || candidates.length === 0 || isPendingThisKind}
                   onChange={(event) => {
                     const connectionId = event.target.value;
                     if (connectionId === '') {
-                      if (current) void remove.mutateAsync(current.id);
+                      if (current) void remove.mutateAsync(current.id).catch(() => {});
                       return;
                     }
-                    void upsert.mutateAsync({ country, documentKind, connectionId });
+                    void upsert.mutateAsync({ country, documentKind, connectionId }).catch(() => {});
                   }}
                 >
                   <option value="">Not set</option>
@@ -106,6 +119,10 @@ export function SalesDocumentCountryDefaults({
                 Filtered to connections with <span className="chip mono-text">{capability}</span>.
                 {candidates.length === 0 ? ' No eligible connection yet.' : null}
               </p>
+              {saveFailedForThisKind ? <Alert tone="error">{saveFailedForThisKind.message}</Alert> : null}
+              {removeFailedForThisKind ? (
+                <Alert tone="error">{removeFailedForThisKind.message}</Alert>
+              ) : null}
             </div>
           );
         })}
--- a/apps/web/src/features/sales-documents/components/sales-document-country-defaults.tsx
+++ b/apps/web/src/features/sales-documents/components/sales-document-country-defaults.tsx
@@ -101,10 +101,10 @@ export function SalesDocumentCountryDefaults({
                   onChange={(event) => {
                     const connectionId = event.target.value;
                     if (connectionId === '') {
-                      if (current) void remove.mutateAsync(current.id).catch(() => {});
+                      if (current) remove.mutate(current.id);
                       return;
                     }
-                    void upsert.mutateAsync({ country, documentKind, connectionId }).catch(() => {});
+                    upsert.mutate({ country, documentKind, connectionId });
                   }}
                 >
                   <option value="">Not set</option>
--- a/apps/web/src/features/sales-documents/components/sales-document-country-routing-dialog.test.tsx
+++ b/apps/web/src/features/sales-documents/components/sales-document-country-routing-dialog.test.tsx
@@ -801,4 +801,34 @@ describe('SalesDocumentCountryRoutingDialog — Reset country (#2189)', () => {
 
     expect(await screen.findByText(/An unmatched order is held/i)).toBeInTheDocument();
   });
+
+  it('should not claim a country has no rules and no default while the counts are still loading', async () => {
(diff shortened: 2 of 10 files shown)


piotrswierzy (commented):
Reviewed the M7 delta against its own base (`phase/sdux-m6-settings-markets`) — 10 files, ~484 lines, `apps/web` only, no backend/schema/migration surface.

**Verified against the shipped evaluator** (`libs/core/src/sales-documents/domain/domain-services/evaluate-sales-document-rules.ts`), which is the point of this PR:
- `isCountryConfigured = rules.length > 0 || defaults.length > 0` is byte-equivalent to the evaluator's `countryConfigured` (L212), and the "held, never passed to ★ Rest of world" claim matches the `if (countryConfigured)` early return (L232).
- The corrected dual-default warning matches `defaults.length > 1 → ambiguous-connection-no-primary` (L196) — the previous "depends on which rule decides" copy really was false.
- "Exactly one rule must match… if two match the order is held" matches `conflicting-rules-equal-priority`.
- `write.visible` is used exactly as `useWriteAccess`'s doc comment prescribes (hide for unauthorized non-demo, `ReadOnlyLock` + disabled for demo), and the destructive control is correctly moved out of the footer beside the new primary `Done`.
- Every copy branch has a test; the new `describe-sales-document-tax-id-coverage.ts` is pure, typed and covered (incl. the empty-conditions case). Both new CSS rules are token-driven.

### IMPORTANT

**`sales-document-country-routing-dialog.tsx` (~L216, L234)** — the fall-through tier reads `rules`/`defaults` which are `?? []` while the queries load. `isEmptyCountry` is guarded by `isSummaryLoading`, but the tier list is not, so during load a configured country transiently renders *"{country} has no rules and no default, so an order billed here goes to ★ Rest of world"* — a false statement about the operator's own configuration, then flips. That is exactly the defect class this PR closes, and the file's own header already says "loading excluded, to avoid a flash". Fix: gate on load, e.g. render the tier only when `!isSummaryLoading`, or `const isCountryConfigured = isSummaryLoading ? null : rules.length > 0 || defaults.length > 0;` and render neutral copy for `null`.

### SUGGESTION

- **`sales-documents-panel.tsx` (L192, L221)** — the two new `<span className="text-muted">` render unstyled: `index.css` defines only `.ai-provider-table .text-muted` (L9207), never a global `.text-muted`. The class that exists app-wide is `muted-text`. (Pre-existing repo-wide sloppiness, but these are new uses.)
- **`sales-document-rules-list.tsx` (L125)** — `disabled={!write.canWrite || deleteRule.isPending}` disables *every* row's Delete while one is in flight; you already compute `isDeletingThisRule`, so scope the `disabled` to it too.
- **`.catch(() => {})` on `mutateAsync`** (3 sites) — `mutate()` doesn't reject and would drop the swallow; the error is already read off `isError`/`error`.

No blocking issues; the IMPORTANT item is a transient-copy honesty gap rather than a functional break.

norbert-kulus-blockydevs:
All four findings addressed in 70d8a26e0.

### IMPORTANT — transient false fall-through copy

Fixed. `isCountryConfigured` is now `boolean | null`:

```ts
const isCountryConfigured: boolean | null = isSummaryLoading
  ? null
  : rules.length > 0 || defaults.length > 0;
```

While the counts are loading the tier renders neutral copy — title *"What happens to an unmatched order"*, body *"Reading this market's rules and defaults…"* — and the terminal Unresolved tier gets a matching third branch instead of falling into the unconfigured arm. Neither the cross-link button nor the "held" claim renders during load, so no statement about the operator's configuration is made before the real counts arrive.

New test drives it with never-resolving `listRules`/`listCountryDefaults` and asserts all three of the old strings are absent (`DE has no rules and no default`, `Falls through to ★ Rest of world`, `An unmatched order is held`) plus the cross-link button.

### SUGGESTIONS

- **`text-muted` → `muted-text`** on both new spans in `sales-documents-panel.tsx` ("Cannot issue until reconnected", "Not a routing candidate"). Confirmed no global `.text-muted` rule exists in `index.css` — only the `.ai-provider-table` descendant one. Left the pre-existing uses on L186/L237 and in `sales-document-rule-composer-dialog.tsx` alone rather than widen the diff; happy to sweep them separately if you want.
- **Per-row Delete disable** — `disabled={!write.canWrite || isDeletingThisRule}`, so one in-flight delete no longer disables every other row's button.
- **`.catch(() => {})`** dropped at all three sites, replaced with plain `mutate()` (rules-list delete, country-defaults remove + upsert). The errors were already surfaced from `isError`/`error`, and `mutate()` doesn't reject.

### Gate

`tsc -b` clean, `eslint src/features/sales-documents` clean, `vitest run src/features/sales-documents` → 14 files / 105 tests passing (104 before, +1 new).

piotrswierzy (approved):
## Tech Lead review — head `70d8a26`, delta against `phase/sdux-m6-settings-markets`

**Verdict: ✅ Approve.** No BLOCKING or IMPORTANT findings remain.

### Prior findings — confirmed status

- **IMPORTANT (fall-through tier asserted "no rules and no default" while loading) — FIXED.** `sales-document-country-routing-dialog.tsx` now derives `isCountryConfigured: boolean | null`, gated on `isSummaryLoading`, and the tier title, body *and* the tier-3 footnote all carry a neutral third branch instead of the `?? []` reading. Pinned by a regression test (*"should not claim a country has no rules and no default while the counts are still loading"*) that stalls both `listRules` and `listCountryDefaults` on a never-settling promise and asserts none of the three concrete claims render.
- **SUGGESTION (row-scoped delete `disabled` used global `isPending`) — FIXED.** `sales-document-rules-list.tsx` now scopes on `deleteRule.variables === rule.id`, with a per-row `Deleting…` label and a per-row error `Alert`. Same variables-scoping applied to the defaults upsert/remove.
- **SUGGESTION (`mutateAsync().catch(() => {})`) — FIXED.** All three call sites are plain `mutate()`, with the failure now surfaced rather than swallowed.
- **SUGGESTION (`className="text-muted"`) — still open, and out of this delta.** `text-muted` is a CSS *custom property* name, not a class, so the two spans at `apps/web/src/features/sales-documents/components/sales-documents-panel.tsx:180` and `:219` render unstyled. Both lines predate this PR; every string this PR *adds* correctly uses `muted-text`. Worth a one-line follow-up while you are in that file.

### Fresh review of the current diff

Copy claims were checked against the shipped evaluator, not taken on trust:
- two defaults → `ambiguous-defaults` (`evaluate-sales-document-rules.ts:196`), i.e. held — the rewritten warning is accurate;
- fall-through to `★ Rest of world` only when the country carries zero rules *and* zero defaults (doc comment step 3) — the conditional tier is correct;
- "exactly one match wins, no priority" matches `ambiguous-rules`.

`describe-sales-document-tax-id-coverage.ts` is pure, header-documented, correctly placed in `lib/`, and deliberately says *unreachable for that source* rather than *dead rule* — the right claim post-#2599. The danger-zone gate on `write.visible` matches the documented `useWriteAccess` contract (demo → disabled + tooltip; unauthorized → absent).

### Remaining SUGGESTIONs (non-blocking)

1. `sales-document-rules-list.tsx:74` — the Alert title interpolates the count raw, so a single rule reads *"1 of these rules read the buyer's tax ID"*. Pluralize (`1 rule reads` / `N rules read`).
2. `sales-documents-panel.tsx:192`, `:220` — the new captions stack with a bare `<br />` inside `<td>`; a wrapper element would match how the rest of the app stacks cell lines.

Local test execution was not possible in my worktree (unrelated `posthog-js` resolution failure through a symlinked store); the 56 tests that did run passed. CI covers the rest.

norbert-kulus-blockydevs:
Addressed the three remaining SUGGESTIONs from the review (commit c7a52d59e):

- **`sales-documents-panel.tsx`** — the three new spans now use `muted-text` instead of `text-muted` (a CSS custom-property name, not a class — those spans were rendering unstyled).
- **`sales-documents-panel.tsx`** — the two stacked cell captions ("Cannot issue until reconnected", "Not a routing candidate") no longer use a bare `<br/>`-joined fragment inside `<td>`; each is now its own block-level `<div className="muted-text">`, matching how the rest of the app stacks multi-line cell content.
- **`sales-document-rules-list.tsx`** — the buyer-tax-ID Alert title now pluralizes correctly via a new `describeBuyerTaxIdRuleCount` helper (colocated in `describe-sales-document-tax-id-coverage.ts`): "1 rule reads the buyer's tax ID" vs "N rules read the buyer's tax ID", instead of always reading "N of these rules read...".

`pnpm --filter @openlinker/web type-check` and the full `pnpm --filter @openlinker/web test` suite (4132 tests) both pass; lint shows only pre-existing warnings, none introduced by this change.

Issue #2555, opened 2026-08-26
[TASK] Frontend — mobile card behaviour and skeleton parity


norbert-kulus-blockydevs:
## Problem / Context

Below 768 px the row becomes a card, and the document line has to fold without pushing the page sideways. The skeleton must also reserve the height the loaded cluster takes, or the table jumps on load.

## Proposed Solution

Give the document line a definite width so its truncation can fire inside a card, and make the money-cluster skeleton reserve the four stacked facts. Verify no horizontal page scroll at 360 px.

## Classification

**Type**: Frontend
**Layer**: features/orders
**File(s)**: `apps/web/src/features/orders/`, `apps/web/src/index.css`

## Dependencies

- Parent mini-epic: #2551
- M5 skeleton fix.

## Acceptance Criteria

- [ ] No horizontal page scroll at 360 px on the orders list.
- [ ] The document line folds into the card without overflowing.
- [ ] The skeleton reserves the loaded cluster height.
- [ ] Tap targets clear 44 px on touch.
- [ ] Tests added or updated for non-trivial logic
- [ ] No architecture boundary violations (CORE ↔ Integration)

## Branching and merge target

Branch: `{issue#}-{short-kebab}`, created **from its mini-epic branch**.
PR target: **its mini-epic branch**.

## How this issue gets finished

1. Implement on the branch above.
2. Open the PR into its mini-epic branch.
3. Run `/pr-review <PR>` and post the verdict as a PR comment.
4. Fix **every** finding, and reply under the comment saying what changed for each one.
5. Run `/pr-review` again. Repeat until the review comes back approved with no findings.
6. Only then merge.

Quality gate before every commit: `pnpm lint`, `pnpm type-check`, `pnpm test`.
Commits are signed off (`git commit -s`).

norbert-kulus-blockydevs:
Closing — implemented and merged via #2760.

Issue #2554, opened 2026-08-26
[TASK] Frontend — attention count, and issue-on-request counted separately


norbert-kulus-blockydevs:
## Problem / Context

Issuing by hand is the default trigger model, so counting those orders as blocked puts a large red number on a healthy install — the exact regression the attention-reason subset exists to prevent. The current label also says "Invoicing blocked" for a counter that covers receipts too.

## Proposed Solution

Count only attention-worthy reasons, and report issue-on-request separately as what it is: orders waiting for the operator, by configuration. Use cross-kind wording, since the underlying reasons are not invoice-specific. Link the attention count to the existing filter.

## Classification

**Type**: Frontend
**Layer**: features/orders
**File(s)**: `apps/web/src/features/orders/`

## Dependencies

- Parent mini-epic: #2551
- M5 copy map.

## Acceptance Criteria

- [ ] Issue-on-request is excluded from the attention count.
- [ ] It is still reported, neutrally, as its own figure.
- [ ] No label uses invoice vocabulary for a cross-kind fact.
- [ ] The attention figure links to the filter that shows those rows.
- [ ] Tests added or updated for non-trivial logic
- [ ] No architecture boundary violations (CORE ↔ Integration)

## Branching and merge target

Branch: `{issue#}-{short-kebab}`, created **from its mini-epic branch**.
PR target: **its mini-epic branch**.

## How this issue gets finished

1. Implement on the branch above.
2. Open the PR into its mini-epic branch.
3. Run `/pr-review <PR>` and post the verdict as a PR comment.
4. Fix **every** finding, and reply under the comment saying what changed for each one.
5. Run `/pr-review` again. Repeat until the review comes back approved with no findings.
6. Only then merge.

Quality gate before every commit: `pnpm lint`, `pnpm type-check`, `pnpm test`.
Commits are signed off (`git commit -s`).

norbert-kulus-blockydevs:
Closing — implemented and merged via #2759.

Issue #2553, opened 2026-08-26
[TASK] Frontend — the document popover: facts, reason and actions


norbert-kulus-blockydevs:
## Problem / Context

Number, provider, authority answer, held reason, duplicates and any action cannot all live in a table cell — the first attempt stacked five lines and became a panel inside a column.

## Proposed Solution

Put them in a popover opened from the cell: the kind and state repeated in full, the reason from the M5 copy map, the identity facts, a duplicate warning where one exists, and the actions. Use the M5 popover so it is not clipped by the table's horizontal scroll. Escape closes, focus returns to the trigger.

## Classification

**Type**: Frontend
**Layer**: features/orders
**File(s)**: `apps/web/src/features/orders/components/`, `apps/web/src/shared/ui/popover.tsx`

## Dependencies

- Parent mini-epic: #2551
- M5 popover and copy map.

## Acceptance Criteria

- [ ] Every fact the old stacked cell showed is reachable in the popover.
- [ ] It renders fully when opened from the last column of a scrolled table.
- [ ] Escape and outside click close it; focus returns to the trigger.
- [ ] The reason shown is the persisted one.
- [ ] Tests added or updated for non-trivial logic
- [ ] No architecture boundary violations (CORE ↔ Integration)

## Branching and merge target

Branch: `{issue#}-{short-kebab}`, created **from its mini-epic branch**.
PR target: **its mini-epic branch**.

## How this issue gets finished

1. Implement on the branch above.
2. Open the PR into its mini-epic branch.
3. Run `/pr-review <PR>` and post the verdict as a PR comment.
4. Fix **every** finding, and reply under the comment saying what changed for each one.
5. Run `/pr-review` again. Repeat until the review comes back approved with no findings.
6. Only then merge.

Quality gate before every commit: `pnpm lint`, `pnpm type-check`, `pnpm test`.
Commits are signed off (`git commit -s`).

norbert-kulus-blockydevs:
Closing — implemented and merged via #2758.

Issue #2552, opened 2026-08-26
[TASK] Frontend — the document line inside the money cluster


norbert-kulus-blockydevs:
> Mockup: `docs/plans/mockups/sales-document-routing.html` — states `#page=orders`

## Problem / Context

Today the payment line carries a hardcoded `+ Issue invoice` button, offered even on orders routed to a fiscal receipt, and the clearance badge sits beside "Paid" as though it were a payment state. Three subjects share one column headed "Payment".

## Proposed Solution

Replace the button with one line: the kind glyph, one word for the state, a small tick when finished, a quiet pulsing dot while in flight, and a count badge when a second record exists. The most actionable fact becomes the word — an invoice the authority rejected reads "KSeF rejected", not "Issued" beside a red dot, so colour is never the only carrier of the thing needing work.

## Classification

**Type**: Frontend
**Layer**: features/orders
**File(s)**: `apps/web/src/features/orders/components/`, `apps/web/src/features/sales-documents/`

## Dependencies

- Parent mini-epic: #2551
- M1 list projection
- M5 primitives

## Acceptance Criteria

- [ ] The cell renders one line at any state.
- [ ] The word names the most actionable fact, not merely the document status.
- [ ] A finished document is plain ink with a tick; only attention states are tinted.
- [ ] A duplicate shows as a count, with the explanation one click away.
- [ ] The existing columns are unchanged.
- [ ] Tests added or updated for non-trivial logic
- [ ] No architecture boundary violations (CORE ↔ Integration)

## Branching and merge target

Branch: `{issue#}-{short-kebab}`, created **from its mini-epic branch**.
PR target: **its mini-epic branch**.

## How this issue gets finished

1. Implement on the branch above.
2. Open the PR into its mini-epic branch.
3. Run `/pr-review <PR>` and post the verdict as a PR comment.
4. Fix **every** finding, and reply under the comment saying what changed for each one.
5. Run `/pr-review` again. Repeat until the review comes back approved with no findings.
6. Only then merge.

Quality gate before every commit: `pnpm lint`, `pnpm type-check`, `pnpm test`.
Commits are signed off (`git commit -s`).

norbert-kulus-blockydevs:
Closing — implemented and merged via #2757.

PR #2761, opened 2026-09-01
feat(web): show each order's invoice or receipt status directly in the orders list

norbert-kulus-blockydevs:
## Summary

M8 of the sales-document routing-first UX epic (#2513): the orders list's sales-document line (#2551). Independent of PR #2746 (M6) and PR #2756 (M7) - all three branch from `epic/sales-documents-ux` and target `main` directly.

- `SalesDocumentCell` replaces the old hardcoded `+ Issue invoice` button (`OrderInvoicingCell`) inside the money cluster: a kind glyph, one word for the most actionable fact, a tick when finished, a pulsing dot while in flight, and a duplicate-record count badge (#2552)
- The cell's popover: identity facts, the persisted block/unresolved reason, a duplicate-record warning, and the available action - reuses the M5 `Popover` primitive so it is never clipped by the table's scroll container, and dismisses on viewport change (#2553)
- Order-health summary gains `salesDocumentIssuedOnRequest`, its own count strictly separate from `salesDocumentBlocked` - `trigger-model-manual` is the default trigger model, so folding it into "blocked" would put a large red number on a healthy install (#2554)
- Mobile card folding and skeleton parity for the sales-document line, plus a 44px tap target under `pointer: coarse` (#2555)

## Review fix applied before opening this PR

`SalesDocumentCell` originally defined its own `InvoiceGlyph`/`ReceiptGlyph`/`NoDocumentGlyph` SVGs, duplicating the M5 `DocumentKindGlyph` primitive that already draws the same three silhouettes. Fixed to reuse the shared primitive (commit `1d31a9ed3`) before opening this PR.

## Test plan

🤖 Generated with Claude Code


piotrswierzy (commented):
Tech-lead review of the **M8 delta against `phase/sdux-m7-settings-dialog`** (12 files, not the 62 GitHub shows against the stacked base).

Verified: `SalesDocumentCell` reuses the M5 `DocumentKindGlyph`/`Popover` primitives (no duplicated SVGs, no positioning code, portal-escape asserted through the component); the state resolver is pure, kind-discriminated and unit-tested for both document kinds incl. unrecognised statuses; no platform name leaks into the row word; `salesDocumentIssuedOnRequest` is kept strictly out of `salesDocumentBlocked` and the SQL predicates are mutually exclusive by construction (ADR-041 §54); no ORM schema change, so no migration is owed; desktop/mobile share the same cell component; `lines: 5` skeleton reservation matches the new stack.

## IMPORTANT

**1. `apps/web/src/features/orders/lib/sales-document-cell-state.ts` (~L110)** — `view === undefined` and `documentKind === null` both resolve to `tone: 'error', attention: true, word: 'No routing'`. `orders.types.ts` (the field's own doc, added in this PR) says absence "must render the same 'no document' state … never treat absence as an error". As written, an install that hasn't configured routing — or an FE running against an API predating the field — paints a red line on *every* order row. That is the same "large red number on a healthy install" regression #2554 exists to prevent, one surface down. Split the two: unrouted/absent should resolve `tone: 'idle'`, `attention: false` (keep the `Set routing` popover action as the affordance).

**2. `apps/web/src/pages/orders/orders-list-page.tsx` (~L1015 and ~L1705)** — the tax-rate-conflict badge moved out of `OrderInvoicingCell` and is now hand-duplicated in the desktop cell and the mobile card, and the two **already differ**: desktop adds an `sr-only` `label: hint`, mobile does not. The retired component's own comment ("this used to be a hand-duplicated parallel render path, and the two diverged") is the reason not to reintroduce this. Extract a `TaxRateConflictBadge` in `features/orders/components` and render it verbatim in both.

**3. `libs/core/src/orders/.../order-record.repository.ts` (~L330)** — the new `sales_document_issued_on_request` aggregate has no test. Its sibling predicate is covered (`order-record.repository.spec.ts:976`). Add a case asserting the two `FILTER` predicates never both match a row.

## SUGGESTION

- `sales-document-cell.tsx` (~L190): `aria-label` on the trigger overrides inner content, so the `sr-only` kind prefix and the duplicate-record `sr-only` + count are unreachable to screen readers — a duplicate is invisible on the row for AT users. Fold the count into the `aria-label` and drop the dead `sr-only` spans.
- `sales-document-cell.tsx` (`PopoverContent`) and `orders-list-page.tsx` ("issued on request"): inline `style={{ width: '20rem' }}` / `fontSize: '0.75rem'`. Style guide asks for component classes + tokens in `index.css`.
- `taxRateConflictBadge()` is invoked 3× per desktop render; hoist to a const. Dupe badge copy says "A second document exists" even at `dupeCount > 1`.

**Verdict:** no BLOCKING issues; please address the three IMPORTANT items before merge.

norbert-kulus-blockydevs:
Thanks - all three IMPORTANT items and every SUGGESTION are addressed in `d47c0535c`.

## IMPORTANT

**1. Absence is no longer an error** (`sales-document-cell-state.ts`)

Split as asked. The two cases now resolve differently:

- no view at all, or `documentKind === null` with **no persisted reason** -> `word: 'No document'`, `tone: 'idle'`, `attention: false`. The popover keeps its `Set routing` action, which is the affordance for that state.
- `documentKind === null` **with** a reason the gate actually persisted -> that reason's own copy, tone and detail. So `unresolved-routing` is still red, because the backend said so rather than because a field was absent.

That distinction is the one the resolver was missing: it treated "the API did not tell us" and "the gate refused" as the same fact. Three unit cases pin it, including the one asserting a persisted `unresolved-routing` still reads `error` / `attention: true`.

**2. One tax-rate-conflict badge** (`features/orders/components/tax-rate-conflict-badge.tsx`)

Extracted and rendered verbatim in the desktop cell and the mobile card. You were right that the two had already drifted - mobile was missing the `sr-only` hint, so the conflict's explanation was desktop-only for AT users. The new component's header records why it exists (the retired `OrderInvoicingCell`'s own comment), so the next person does not re-inline it. `taxRateConflictBadge()` is now called once per render as a side effect of the extraction, which also closes that suggestion.

**3. `sales_document_issued_on_request` is covered** (`order-record.repository.spec.ts`, new `countByHealth` block)

Three cases:

- the two `FILTER` predicates are mutually exclusive: the on-request one matches `= 'trigger-model-manual'`, and the blocked IN-list must not contain that value. If manual were ever added to `SalesDocumentAttentionReasonValues` a row would be counted twice and the page would put a large red number on a healthy install - the test fails first.
- the oldest-held `MIN` reuses the blocked predicate, so it also must not carry the manual value. Pinned so a copy-paste into that aggregate cannot pass silently.
- both counts map independently, and an absent aggregate reads `0` rather than `NaN`.

The empty-`IN ()` guard from the `findMany` sibling is asserted here too.

## SUGGESTION

- **aria-label**: the duplicate-record sentence now rides in the trigger's `aria-label` and the dead `sr-only` spans are gone. You were right that they were unreachable, so a duplicate was invisible on the row to a screen reader. The sentence also pluralises now (`3 documents exist for this order` past a second record), and the popover's warning agrees ("also holds" / "also hold"). Two new component tests cover the label and the plural.
- **inline styles**: the popover width moved to `.sales-doc-popover` / `.sales-doc-popover--row`, and the summary note's font size to `.orders-summary-note`, both in `index.css`.
- one extra thing while in there: with nothing routed, the kind label and the word were the same string, so the trigger read "No document: No document" and the popover head "No document - No document". Said once now.

## Gate

`pnpm lint` and `pnpm type-check` clean. Targeted suites green: `sales-document-cell.test.tsx` (13), `sales-document-cell-state.test.ts` (11), `orders-list-page.test.tsx` (83), `order-record.repository.spec.ts` (83).

piotrswierzy (commented):
Re-review at `d47c053`, scoped to the true delta against `phase/sdux-m7-settings-dialog` (14 files, not the 64 GitHub shows).

## Prior findings — all three fixed ✅

1. **Absence-as-error** — `resolveSalesDocumentCellState` now returns `idle` / `attention: false` for both a missing `view` and a `documentKind: null` with no persisted reason, and keeps a *backend-persisted* reason's own tone. The distinction is documented in the resolver header and pinned by three tests (`sales-document-cell-state.test.ts:35/45/55`). The #2554/ADR-041 §54 regression is closed.
2. **Duplicated tax-rate badge** — extracted to `features/orders/components/tax-rate-conflict-badge.tsx`, used verbatim by the desktop money cluster and the mobile card, `sr-only` hint now on both. The file header records why it is a component.
3. **Untested SQL aggregate** — `order-record.repository.spec.ts:192-270` now asserts on the emitted SQL: `sales_document_issued_on_request` matches `'trigger-model-manual'`, `sales_document_blocked` does not, the `MIN(...)` age clause does not, and a missing column maps to `0` rather than `NaN`.

Also good: tokens-only CSS (no raw hex), `pointer: coarse` 44 px target, `lines: 5` skeleton parity, glyph reuse of the M5 primitive.

## IMPORTANT

**`apps/web/src/features/orders/components/sales-document-cell.tsx:196` (duplicate warning) and `:186` (Provider fact)** — both render raw connection UUIDs to the operator. `SalesDocumentOtherRecord` carries only `connectionId`, so the warning reads "`3f2a…, 9c81…` also hold a document for this sale" — the operator is asked to act on connections the copy names only by UUID. `frontend-ui-style-guide.md` § `EntityLabel` is explicit that a UUID must not leak on an operator surface. The page already batches `useConnectionsQuery` (it derives `hasIssuingCapability` from it), so the fix costs nothing: build a `Map<id, name>` in `orders-list-page.tsx`, pass it as a prop, and render name-first with the id as fallback (or use `ConnectionEntityLabel`). Same for the `Provider` fact's `identity.providerType ?? identity.connectionId`.

## SUGGESTION

- **`order-invoicing-cell.tsx` + `.test.tsx`** are now unreferenced by any production module (only comments mention them). Delete in this PR, or add a retention note like `ShopPublishLauncher`'s — an unreferenced "single shared render path" component is exactly what invites the drift its own header warns about.
- **`sales-document-cell.tsx:44-48`** — the `layout` prop's docstring says `row` "wraps the line so a mobile `<dd>` still receives a single child", but the implementation only toggles a popover width class; the trigger is always the single child. Correct the comment.
- **`SalesDocumentView.blockDetail`** (the gate's PII-free elaboration) is fetched and never rendered. The popover's `__why` paragraph is where it belongs.

Verdict: 🔄 approve after the UUID-rendering fix. No blocking issues.

norbert-kulus-blockydevs:
Addressed the second review round (commit 6369726f4):

**IMPORTANT — raw UUID leak (fixed):** `SalesDocumentCell` now takes a `connectionNames: Map<string, string>` prop (built once from the page's already-batched `useConnectionsQuery`) and resolves it for both the duplicate-record warning ("X also holds a document...") and the `Provider` fact, falling back to the raw id only if the connection isn't in the lookup. Wired at both call sites (desktop cell + mobile card). Added a test asserting the resolved name renders and the raw `connectionId` does not.

**SUGGESTIONs (all applied):**
- Deleted the now fully-unreferenced `order-invoicing-cell.tsx` + its test — nothing in production code referenced it anymore, only comments.
- Corrected the `layout` prop docstring, which described a `row` wrapping behavior the implementation doesn't actually do (it only widens the popover).
- The gate's persisted `blockDetail` elaboration is now rendered in the popover's `__why` section alongside the copy-derived `reasonDetail`.

`pnpm --filter @openlinker/web type-check` and the full web test suite (4085 passed) are green.

piotrswierzy (approved):
Tech-lead re-review at `f3e1b3a`, against the PR's own base (`main`).

## Prior review findings — all closed ✅

- **UUID leakage** (last review's only IMPORTANT): `connectionNames: Map<string,string>` is now built in `orders-list-page.tsx` from the already-batched `useConnectionsQuery` and threaded into `SalesDocumentCell`; both the duplicate warning and the `Provider` fact resolve name-first with the id only as a last fallback. `frontend-ui-style-guide.md` § `EntityLabel` satisfied.
- **`blockDetail` fetched but never rendered** — now rendered in the popover's `__why` region.
- **`order-invoicing-cell.tsx` + its test** — deleted in this diff rather than left unreferenced. Good; that is the right resolution of the two offered.
- Earlier round's three IMPORTANTs (absence-as-error, duplicated tax-rate badge, untested SQL aggregate) remain fixed and are pinned by tests.

## Verified in this pass

- **Architecture / dependency direction** — `pages → features → shared` holds. `SalesDocumentCell` and `sales-document-cell-state.ts` live in `features/orders`; the cross-feature reach into `sales-documents` goes through that feature's **public barrel** (`resolveSalesDocumentReasonCopy`, `SalesDocumentReasonTone`), not a deep path.
- **State ownership** — no new client state; the projection rides on the existing `GET /orders` row (`OrderRecordResponseDto.salesDocument`, already on `main`), so the cell issues no second request and no Query data is copied into component state. `frontend-architecture.md` § State Management satisfied.
- **Pure rule, separately tested** — `resolveSalesDocumentCellState` is I/O-free, kind-discriminated, and covered for both document kinds including the *unrecognised status* arm. That last arm is the right call: a newer backend answering an older FE reports "Unrecognised status / warning" rather than rendering nothing.
- **Vocabulary neutrality** — no platform name reaches the row word ("Authority rejected", not "KSeF rejected"); the provider name is confined to the popover identity facts. This is the same discipline `libs/core` enforces on the wire shape.
- **CSS** — tokens only, no raw hex; `prefers-reduced-motion` kills the pulse; `pointer: coarse` gives a real 44 px target without making the whole card row tappable; `min-width: 0` on the flex child is correctly identified as load-bearing for the 360 px no-horizontal-scroll criterion. Mobile + tablet coverage is genuinely present, not desktop-only.
- **Accessibility** — the `aria-label`-overrides-content trap is handled correctly: the duplicate sentence rides in the label itself and is asserted by a test. Badges carry words, not colour alone (`state.word` is the signal; tone only tints).
- **Skeleton parity** — `lines: 4 → 5` matches the tallest money cluster (total, payment badge, document line, conflict badge, created).
- **Backend** — `salesDocumentIssuedOnRequest` is a new aggregate only; no ORM entity change, so **no migration is owed** and none is missing. The two `FILTER` predicates are mutually exclusive by construction (`IS_SALES_DOCUMENT_BLOCKED` is built from `SalesDocumentAttentionReasonValues`, which excludes `trigger-model-manual`), and the emitted SQL is now asserted. This is the ADR-041 §54 rule applied correctly rather than restated.

## IMPORTANT

**1. `.eslintrc.js` — register the `sales-documents` slug now that `orders` cross-imports it.**
`features/orders/lib/sales-document-cell-state.ts` imports from `../../sales-documents`. The import form is correct (barrel), but the slug is absent from **both** `no-restricted-imports` pattern groups (`features/**` and `plugins/**`) — `grep -n "sales-documents" .eslintrc.js` returns nothing, while every other cross-imported feature (`orders`, `invoicing`, `fulfillment`, …) is enumerated for all five canonical subdirectories. `frontend-architecture.md` § *Feature Public Surface → Adding a new public feature surface*, step 2, makes registering the slug part of establishing a cross-feature consumer, and the doc explicitly warns the rule "silently fails open" for anything unlisted. This PR is the moment that consumer appears, so the ten lines belong here — otherwise the next author reaching into `sales-documents/lib` gets no lint error. Pre-existing omission, newly load-bearing.

**2. Scope: the diff against `main` carries M6/M7, not just M8.**
Beyond the orders-list work, this branch also lands the settings page panel reorder (#2550, `sales-documents-page.tsx`), the routing-dialog danger zone (#2549, `index.css` + `sales-document-country-routing-dialog.tsx`), and `describe-sales-document-tax-id-coverage.ts` (#2546). The description says M8 is "independent of PR #2746 (M6) and PR #2756 (M7)", but they share commits off `epic/sales-documents-ux`, so **merging this first lands M7's page reordering under an M8 title**. Not a defect — but state the intended merge order in the description, or the reviewer of #2756 will be reviewing an already-merged change.

## SUGGESTION

- **`sales-document-cell-state.ts` — `attention` is computed, tested, and never read.** The component tints from `state.tone` alone; `attention` has no consumer in production code. Either drop it or wire it (it is the field a future aggregate/filter would want), but a field asserted only by its own unit test invites the two to drift.
- **`tax-rate-conflict-badge.tsx` — the label is announced twice.** `StatusBadge` renders `{label}` as visible text and the `sr-only` span repeats `{label}: {hint}`, so AT reads "Tax rate conflict. Tax rate conflict: …". Emit only the hint in the `sr-only` span, or `aria-hidden` the visible badge text.
- **`sales-document-cell.tsx` — the `Provider` fact prefers `providerType` over the connection name.** `identity.providerType ?? connectionNames.get(...)` means a resolved record shows `ksef` rather than which connection issued it — on a multi-connection install that is the less useful of the two facts, and the duplicate warning right below it names connections. Consider name-first with the provider type as a qualifier.
- **Every row now renders a clickable trigger, even where nothing can ever issue.** On an install with no `Invoicing`/`Fiscalization` connection, all rows get "No document" plus a popover offering `Set routing`. Consider rendering the static line (no trigger) when `!hasIssuingCapability && !view?.document` — the same reasoning that already withholds the action.

## Assessment

M8 does what the mini-epic asked and does it in the shape the codebase already argues for: one pure resolver shared by desktop and mobile, one component instead of two render paths (the exact drift `OrderInvoicingCell`'s own header warned about), shared M5 `Popover`/`DocumentKindGlyph` primitives rather than a second glyph set, and colour used strictly as a secondary signal. Test coverage is real — the state resolver is exercised across both kinds, both failure modes, the unrecognised-status arm, and the absence-is-not-an-error rule; the new SQL aggregate is asserted on the emitted query. No BLOCKING issues, and no migration owed.

🔄 **Approve with changes** — merge after registering the `sales-documents` ESLint slug (item 1) and stating the merge order relative to #2756/#2746 (item 2). Neither blocks correctness of the shipped UI.

### Priority
1. Add `sales-documents` to both `no-restricted-imports` slug groups in `.eslintrc.js`.
2. Record the intended merge order for the M6/M7/M8 stack in the description.
3. (Optional) `attention` dead field, double-announced conflict badge, `Provider` fact ordering.

norbert-kulus-blockydevs:
Addressed the third review round (commit cf1f10e76).

**IMPORTANT**

1. **`sales-documents` ESLint slug** — registered in both `no-restricted-imports` pattern groups (`features/**` and `plugins/**`) in `.eslintrc.js`, for all five canonical subdirectories. Also added the cross-feature-consumption entry to `docs/frontend-architecture.md § Feature Public Surface`.
2. **Merge order / diff scope** — merged `origin/main` into this branch. M6 (#2746) and M7 (#2756) have both merged in the meantime, so the PR's diff against `main` is now scoped to just M8's orders-list work (16 files) rather than also carrying the settings-page changes.

**Suggestions**

- Dropped `SalesDocumentCellState.attention` — nothing in production read it, and a field asserted only by its own unit test invited drift. Chose "drop" over "wire" since neither the row nor the popover currently has a use for a signal independent of `tone`.
- Fixed the double-announced tax-rate-conflict badge: the `sr-only` span now carries only the hint, since `StatusBadge` already renders the label as visible text.
- The popover's `Provider` fact now reads the connection name first with the provider type as a parenthetical qualifier, matching how the duplicate-record warning right below it already names connections.

Verified: `pnpm --filter @openlinker/web test` (5143 passed), `tsc --noEmit`, `pnpm lint` on `@openlinker/web` and `@openlinker/core` (0 errors), and `pnpm check:invariants` (all green, including `check-cross-context-imports`).

Issue #2562, opened 2026-08-26
[TASK] Frontend — async announcements, focus and the panel skeleton


norbert-kulus-blockydevs:
## Problem / Context

The panel changes state while the operator waits, and a screen-reader user learns nothing: `aria-busy` on a div announces nothing, and the trigger unmounts when pressed, dropping keyboard focus to the body mid-operation.

## Proposed Solution

One live region carrying the current wait and its outcome, with the elapsed counter hidden from assistive technology so it is not read out every second. Keep the trigger mounted and disabled during the wait so focus is never dropped. Add a panel skeleton sized from the loaded panel.

## Classification

**Type**: Frontend
**Layer**: features/sales-documents
**File(s)**: `apps/web/src/features/sales-documents/components/`

## Dependencies

- Parent mini-epic: #2556

## Acceptance Criteria

- [ ] A state change during the wait is announced once, not on every tick.
- [ ] Focus is never dropped when an action is pressed.
- [ ] The elapsed counter is hidden from assistive technology.
- [ ] The skeleton does not change the panel height on arrival.
- [ ] Tests added or updated for non-trivial logic
- [ ] No architecture boundary violations (CORE ↔ Integration)

## Branching and merge target

Branch: `{issue#}-{short-kebab}`, created **from its mini-epic branch**.
PR target: **its mini-epic branch**.

## How this issue gets finished

1. Implement on the branch above.
2. Open the PR into its mini-epic branch.
3. Run `/pr-review <PR>` and post the verdict as a PR comment.
4. Fix **every** finding, and reply under the comment saying what changed for each one.
5. Run `/pr-review` again. Repeat until the review comes back approved with no findings.
6. Only then merge.

Quality gate before every commit: `pnpm lint`, `pnpm type-check`, `pnpm test`.
Commits are signed off (`git commit -s`).

norbert-kulus-blockydevs:
Closing — implemented and merged via #2762.

Issue #2561, opened 2026-08-26
[TASK] Frontend — disclosures, override gating and duplicate record


norbert-kulus-blockydevs:
## Problem / Context

The override that lets an operator issue the other kind was reachable from states that already carry a document, including the one where OpenLinker does not know whether a receipt exists — the state where a second attempt is most dangerous. A second record on another connection was also hidden behind a single-record panel.

## Proposed Solution

Gate the override to states where nothing has been issued, and to an admin session, since both write paths are admin-only. Put the routing explanation in a disclosure directly above it, so the two secondary affordances sit together and neither competes with the primary action. Surface a second record as a warning above the disclosures, not below them.

## Classification

**Type**: Frontend
**Layer**: features/sales-documents
**File(s)**: `apps/web/src/features/sales-documents/components/`

## Dependencies

- Parent mini-epic: #2556
- #2504

## Acceptance Criteria

- [ ] The override is unreachable from any state carrying a document, and from a non-admin session.
- [ ] The routing explanation is a disclosure placed above the override.
- [ ] A duplicate record renders above the disclosures.
- [ ] The override copy states that it applies to this order only.
- [ ] Tests added or updated for non-trivial logic
- [ ] No architecture boundary violations (CORE ↔ Integration)

## Branching and merge target

Branch: `{issue#}-{short-kebab}`, created **from its mini-epic branch**.
PR target: **its mini-epic branch**.

## How this issue gets finished

1. Implement on the branch above.
2. Open the PR into its mini-epic branch.
3. Run `/pr-review <PR>` and post the verdict as a PR comment.
4. Fix **every** finding, and reply under the comment saying what changed for each one.
5. Run `/pr-review` again. Repeat until the review comes back approved with no findings.
6. Only then merge.

Quality gate before every commit: `pnpm lint`, `pnpm type-check`, `pnpm test`.
Commits are signed off (`git commit -s`).

norbert-kulus-blockydevs:
Closing — implemented and merged via #2762.

Issue #2560, opened 2026-08-26
[TASK] Frontend — not-issued states, each with its own remedy


norbert-kulus-blockydevs:
> Mockup: `docs/plans/mockups/sales-document-routing.html` — states `held-on-request`, `blocked-tax-rate`, `no-routing`, `provider-reauth`

## Problem / Context

Every reason a document has not been issued currently renders as one panel with a dropdown and a disclaimer. An operator cannot tell a configured manual trigger apart from a missing tax rate apart from no routing at all, and each needs a different fix.

## Proposed Solution

One state per reason, parameterised from the M5 copy map: issued-on-request keeps its action because that is its workflow; a missing tax rate points at the order lines; no routing points at the settings and refuses to guess a kind; a provider needing re-authentication says it cannot issue and offers to reconnect. Add the batched trigger state, which the current panel has no representation for.

## Classification

**Type**: Frontend
**Layer**: features/sales-documents
**File(s)**: `apps/web/src/features/sales-documents/components/`

## Dependencies

- Parent mini-epic: #2556
- M5 copy map
- M1 projection

## Acceptance Criteria

- [ ] Every reachable held or unresolved reason has a state with a remedy.
- [ ] Issue-on-request keeps its action and its neutral tone.
- [ ] No-routing points at the configuration and never offers a kind choice.
- [ ] A provider that cannot be reached says so rather than offering a dead action.
- [ ] Tests added or updated for non-trivial logic
- [ ] No architecture boundary violations (CORE ↔ Integration)

## Branching and merge target

Branch: `{issue#}-{short-kebab}`, created **from its mini-epic branch**.
PR target: **its mini-epic branch**.

## How this issue gets finished

1. Implement on the branch above.
2. Open the PR into its mini-epic branch.
3. Run `/pr-review <PR>` and post the verdict as a PR comment.
4. Fix **every** finding, and reply under the comment saying what changed for each one.
5. Run `/pr-review` again. Repeat until the review comes back approved with no findings.
6. Only then merge.

Quality gate before every commit: `pnpm lint`, `pnpm type-check`, `pnpm test`.
Commits are signed off (`git commit -s`).

norbert-kulus-blockydevs:
Closing — implemented and merged via #2762.

Issue #2559, opened 2026-08-26
[TASK] Frontend — fiscal receipt states, including the wait


norbert-kulus-blockydevs:
> Mockup: `docs/plans/mockups/sales-document-routing.html` — states `receipt-ready`, `receipt-registering`, `receipt-registered`, `receipt-rejected`, `receipt-in-doubt`, `receipt-contended`

## Problem / Context

The panel has no feedback while a provider is being called, and reports `Still not found` during normal processing. Registering and already-running are also indistinguishable, and Rejected is currently folded into the same posture as Unconfirmed even though one is safe to retry and the other is not.

## Proposed Solution

Cover ready, registering, registered, rejected, unconfirmed and already-running. The wait states only what the system can observe: it handed the sale over and is waiting for one answer, with elapsed time and no invented phases. Rejected says nothing was created and re-registering is safe. Unconfirmed says we do not know, offers a check, and refuses to retry. Already-running says another attempt is in flight and shows no counter, because closing this tab cannot stop someone else's call.

## Classification

**Type**: Frontend
**Layer**: features/sales-documents
**File(s)**: `apps/web/src/features/sales-documents/components/`

## Dependencies

- Parent mini-epic: #2556
- M1 projection
- M2 locate outcome and in-flight signal

## Acceptance Criteria

- [ ] The wait narrates nothing the system cannot observe.
- [ ] Rejected and Unconfirmed have different copy and different actions.
- [ ] Already-running shows no elapsed counter and no keep-the-page-open instruction.
- [ ] A registered receipt states it is final and cannot be corrected here.
- [ ] No state claims delivery to the buyer.
- [ ] Tests added or updated for non-trivial logic
- [ ] No architecture boundary violations (CORE ↔ Integration)

## Branching and merge target

Branch: `{issue#}-{short-kebab}`, created **from its mini-epic branch**.
PR target: **its mini-epic branch**.

## How this issue gets finished

1. Implement on the branch above.
2. Open the PR into its mini-epic branch.
3. Run `/pr-review <PR>` and post the verdict as a PR comment.
4. Fix **every** finding, and reply under the comment saying what changed for each one.
5. Run `/pr-review` again. Repeat until the review comes back approved with no findings.
6. Only then merge.

Quality gate before every commit: `pnpm lint`, `pnpm type-check`, `pnpm test`.
Commits are signed off (`git commit -s`).

norbert-kulus-blockydevs:
Closing — implemented and merged via #2762.

PR #2763, opened 2026-09-01
feat(web): order detail shows invoice and fiscal-receipt lifecycle, not just a badge

norbert-kulus-blockydevs:
## Summary

M9 of the sales-document routing-first UX epic (#2513): the order-detail sales-document panel (#2556). Branches from `phase/sdux-m8-orders-cell` (PR #2761, M8) - part of the `sdux-m*` stack off `epic/sales-documents-ux`, not `main` directly.

- Shared headline (`DocumentHeadline`, M5) replaces the old two-badge header, deriving state purely from `InvoiceRecord` / `FiscalRegistrationRecord` + progress - issuance and clearance are reported as two separate axes rather than one collapsed status word (#2557)
- Invoice lifecycle trail, KSeF resend wiring, correction gated behind a disclosure (#2558)
- Fiscal states including a distinct "in progress elsewhere" (contended) state with no elapsed counter - the backend's `SalesDocumentInFlight` deliberately carries no completion estimate, and the panel does not invent one (#2559)
- Not-issued states gated on the real reason vocabulary so a hard block hides the retry action rather than offering one that would 403/409 (#2560)
- Manual override gated to `useIsAdmin()` (#2342), the same admin-only fact the two write paths enforce server-side, with the routing explanation open by default; duplicate-record disclosure reads `otherInvoicingConnectionIds` / `blocksIssuanceElsewhere` directly (#2561)
- One `aria-live` region carries every wait/outcome announcement, plus a shape-sized skeleton (#2562)
- The KSeF resend action distinguishes a 501 (the connection's provider does not implement `RegulatoryResubmitter` - only inFakt does) from a transient failure, so a structural refusal never reads as "try again"

## Test plan
- [x] `pnpm lint && pnpm type-check`
- [x] `pnpm --filter @openlinker/web test`

🤖 Generated with Claude Code


piotrswierzy (commented):
## Tech Lead review — M9 order-detail sales-document panel

Reviewed the delta against `phase/sdux-m8-orders-cell` only (9 files). Verified: `DocumentHeadline`/`DocumentLifecycle`/`sr-only`/`useResendToKsefMutation`/`resolveSalesDocumentReasonCopy` all exist on the base; `resolveSalesDocumentReasonCopy` returns `null` for a null/unknown reason, so `hardBlockedNoAction` correctly fails open and a plain empty state keeps its manual action; the admin gate matches reality — `@Roles('admin')` is present on both `invoicing.controller.ts` and `fiscalization.controller.ts` writes. The headline/lifecycle split (issuance vs clearance as two axes) and the "no elapsed counter for a contended attempt" restraint are both right and well-argued.

No BLOCKING issues.

### IMPORTANT

**1. Inline one-off styling** — `features/orders/components/sales-document-panel.tsx` (~1350 and ~1428)
`<p className="text-muted" style={{ fontSize: '0.78rem' }}>` (and the second with `width: '100%'`). `frontend-ui-style-guide.md § CSS Implementation Standard` bans page-specific one-off styling in favour of explicit component classes, and `0.78rem` is not on the canonical type scale (metadata is `0.75rem`). Add `.sales-document-panel__scope-note` beside the other rules you already added in `index.css` and drop both `style` props.

**2. No tests for the three new behavioural branches** — `sales-document-panel.test.tsx`
The 812-line panel spec was only *edited* to follow the headline rename; it gains no case for `contended` (the 409 path → info alert + "In progress elsewhere" headline + suppressed progress skeleton), `canOverride` (a non-admin, non-demo session must see no issue/register control), or `hardBlockedNoAction` (a non-`keepsAction` reason hides the action, `missing-tax-rate` does not). The pure helpers are well covered; the gating that decides whether a write control renders is not. Add three cases.

**3. Inline role comparison** — `sales-document-panel.tsx:284`
`session.user?.role === 'admin'`. `frontend-architecture.md § Access Control` states `role` is typed `string` and **must never be compared inline** — a typo compiles and silently evaluates false. I accept the "existing precedent" argument for not introducing a new `Permission` here, but please at least route it through one shared helper (e.g. `isAdminSession(session)` in `shared/auth/`) so the literal exists once and the follow-up is a single edit rather than three.

### SUGGESTION

- `lib/sales-document-headline.ts` returns raw English literals (`'Issued'`, `'Cleared'`, `'In progress elsewhere'`) while every other string in the panel goes through `t(key, fallback)`. Take `t` as a parameter, or return keys.
- Same file: identity renders `— · KSeF` when `providerInvoiceNumber` is null; consider omitting the em-dash segment entirely.
- The `aria-live` region carries both `role="status"` and `aria-live="polite"` (redundant), and an identical consecutive message produces no DOM write, so a repeated outcome is not re-announced.
- `Register receipt` (~1421) is not wrapped in `ReadOnlyLock` even though `canOverride` now explicitly admits `demoMode`. Pre-existing, but this PR is the one that made the demo case deliberate.

norbert-kulus-blockydevs:
## Review addressed - c1d14e64f

All three IMPORTANT items and all four SUGGESTIONs are done. Gate is green: `tsc --noEmit` clean, eslint clean on the touched files, `vitest run` in `apps/web` = **389 files / 4093 passing**.

### IMPORTANT

**1. Inline one-off styling** - both `style={{ fontSize: '0.78rem' }}` props are gone. They are now `.sales-document-panel__scope-note`, declared in `index.css` beside the other rules this PR added, at the canonical `0.75rem` metadata size (not `0.78rem`), with `width: 100%` folded into the class so the second call site no longer differs from the first.

**2. Tests for the three gating branches** - added, plus two supporting cases:

| Case | Asserts |
|---|---|
| `contended` (409) | info alert + `In progress elsewhere` headline, and **no** failure toast |
| `canOverride` - non-admin, non-demo | neither `Issue invoice` nor `Register receipt` renders |
| `canOverride` - admin, same order | the control **does** render (so the case above is about the role, not about the panel failing to mount) |
| `hardBlockedNoAction` | `missing-required-tax-id` hides the action entirely |
| `missing-tax-rate` carve-out | the action stays, disabled-with-reason |

Writing the contended case turned up **two real defects in that path**, both fixed in this commit:

- **The flag cleared itself in the same commit that set it.** The self-clearing effect fired on `!fiscalWorkOutstanding`, and the progress poll on screen still reads `not-requested` at that moment - the peer attempt has not reached `sync_jobs` yet - so the alert never rendered and the operator was told nothing at all. `contended` is now held as the *instant* of contention and clears only on a progress read that post-dates it.
- **`showFiscalSlot` did not admit a contended attempt**, so the alert was rendering into a closed slot and the panel fell through to the empty state - offering to register the sale a second time, which is exactly what the 409 had just refused.

**3. Inline role comparison** - `session.user?.role === 'admin'` moves behind `isAdminSession(session)` in `shared/auth/`, with its own spec (admin / non-admin / anonymous-with-a-user-object). The docblock states that this is a *visibility* decision only, never enforcement.

One deviation from your wording, flagged deliberately: you mentioned three call sites, but there are **16** in `apps/web` (`grep -rn "role === 'admin'\|role !== 'admin'" apps/web/src`). I converted only this PR's own site rather than sweeping fifteen unrelated files into an M9 diff. The helper is in place, so the sweep is now a mechanical follow-up.

### SUGGESTION - all four taken

- **`t` threaded through `sales-document-headline.ts`.** It takes the panel's translator as a *parameter* (typed `HeadlineTranslate = (key, fallback) => string`) rather than reaching for a hook, so the module stays pure and directly testable while its state words go through the same `t(key, fallback)` path as everything else in the panel. Keys are `salesDocument.headline.*`.
- **No em-dash placeholder.** A document with no `providerInvoiceNumber` / `documentReference` is now described by its connection alone (`identityOf`), rather than by a dash that reads like a missing value. Pinned by a test.
- **Live region.** Dropped the `aria-live="polite"` duplicated by `role="status"`. The repeated-message hole is closed too: announcements go through one `announce()` helper that toggles a zero-width suffix when the message is identical to the previous one, so a repeated outcome produces a DOM write and is re-announced - without changing a single word that is read out.
- **`Register receipt` wrapped in `ReadOnlyLock`.** `canOverride` admits `demoMode` for discoverability (#1615), but the write behind it is `@Roles('admin')` - so a demo non-admin now gets the locked treatment and an intent-click event (`demo_fiscal_register_attempted`), matching the invoice control, instead of a live button that would 403.

piotrswierzy (approved):
## Re-review at `e54f16c` (delta vs `phase/sdux-m8-orders-cell`)

All three IMPORTANT findings from the review at `f3a8c2d` are **confirmed fixed**.

**1. Inline `style={{ fontSize: '0.78rem' }}` one-offs — FIXED.**
No `style={{` remains anywhere in the diff. Both one-offs moved into `apps/web/src/index.css` as `.sales-document-panel__correction summary` (`0.8125rem`) and `.sales-document-panel__scope-note` (`0.75rem`) — and both values are now **on** the canonical type scale (`frontend-ui-style-guide.md § Canonical type scale`: Body-small 13 px, Metadata 12 px), rather than the off-scale `0.78rem` they replaced. Spacing uses `var(--space-*)` tokens throughout.

**2. No tests for the three new gating branches — FIXED.**
`sales-document-panel.test.tsx` gains a dedicated `describe('SalesDocumentPanel — the gates that decide whether a write control renders')` block covering all three:
- `contended` → *"should report a refused second attempt as in-progress-elsewhere, not as a failure"*
- `canOverride` → the admin/non-admin pair (`should hide the manual issue control from a non-admin session outside demo mode` / `should offer the manual issue control to an admin session on the same order`)
- `hardBlockedNoAction` → *"should hide the action entirely for a hard block…"*, plus the `missing-tax-rate` carve-out asserted separately, which is the branch most likely to regress silently.

Two new pure modules also ship with their own specs (`sales-document-headline.test.ts`, `invoice-lifecycle-steps.test.ts`).

**3. Inline `session.user?.role === 'admin'` — FIXED.**
`shared/auth/is-admin-session.ts` now exists and the panel consumes it (`sales-document-panel.tsx:93,254`). The docblock correctly frames it as a *visibility* decision behind a server-side `@Roles('admin')` gate, and correctly points permission-shaped gating at `usePermission` / `useWriteAccess`. Note the PR description still says this was deferred as a follow-up — it's stale; the fix is in the diff.

### SUGGESTION — `features/invoicing/lib/invoice-lifecycle-steps.ts`
Step labels (`'Issued'`, `'Cleared'`, `'Awaiting the authority'`, `'Rejected by the authority'`) are hardcoded English, while its sibling `sales-document-headline.ts` deliberately takes `t: HeadlineTranslate` as a parameter *specifically* so it isn't "the one place that hardcodes English" (its own docblock). Two modules added in one PR now disagree on that. Harmless today — the i18n seam is a no-op — but it's a one-line fix to thread the same `HeadlineTranslate` parameter through and keep the pair consistent.

Also: the test-plan checkboxes in the description are unticked; please confirm `pnpm --filter @openlinker/web test` is green before merge.

Nothing BLOCKING or IMPORTANT remains. ✅

norbert-kulus-blockydevs:
Addressed the remaining SUGGESTION from the review (`e54f16c`): `invoice-lifecycle-steps.ts` now threads a `LifecycleStepTranslate` parameter through `resolveInvoiceLifecycleSteps` / `resolveClearanceStep`, mirroring `HeadlineTranslate` in the sibling `sales-document-headline.ts`, so it's no longer the one place in the pair that hardcodes English step labels.

Changes:
- `apps/web/src/features/invoicing/lib/invoice-lifecycle-steps.ts` — added `LifecycleStepTranslate`, both functions now take `t` and route every label through `t(key, fallback)`
- `apps/web/src/features/orders/components/sales-document-panel.tsx` — call site passes the panel's existing `t`
- `apps/web/src/features/invoicing/lib/invoice-lifecycle-steps.test.ts` — updated to pass a `t` stub

Verified: `tsc --noEmit` clean, full `@openlinker/web` test suite green (4093 passed / 1 skipped), lint 0 errors (pre-existing unrelated warnings only).

piotrswierzy (approved):
Tech-lead re-review at `9451e75`, scoped to this PR's own delta against its base `phase/sdux-m8-orders-cell` (15 files).

## Since the approval at `e54f16c`

The one outstanding SUGGESTION is **fixed**: `invoice-lifecycle-steps.ts` now takes a `LifecycleStepTranslate` parameter and routes all four step labels through `t(key, fallback)`, matching its sibling `sales-document-headline.ts`. The docblock even names the reason ("so this module isn't the one place in the pair that hardcodes English"). The pair no longer disagrees.

## Verified in this pass

- **Two axes, not one word.** `resolveInvoiceHeadline` decides from `status` first and then lets `regulatoryStatus` override — so "issued, then rejected by the authority" is expressible, which is exactly the state ADR-065 says a flattened status cannot carry. `resolveInvoiceLifecycleSteps` correctly returns a **one-step** trail for `not-applicable`: a regime that clears nothing has no second stage, and inventing a greyed-out "Awaiting the authority" would assert an axis that does not exist. Both are pure, React-free and directly tested.
- **Restraint on the contended state.** `contendedAt` is held as an instant rather than a boolean, with the reason stated: the clear has to wait for an answer that post-dates the 409, and a boolean cannot distinguish that from the poll already on screen. And no elapsed counter is rendered — correct, and it matches what the backend actually promises: `SalesDocumentInFlight.since` is a *lower bound* on elapsed time and the lease expiry is deliberately not carried, so any countdown here would be a completion estimate OpenLinker cannot make.
- **`aria-live` zero-width-space trick.** A live region that does not change is not re-announced, so an identical consecutive outcome would be silent. The `​` toggle changes the text node without changing a word that is read out. That is the right fix for a real problem, and it is documented in place.
- **Not-issued gating.** `hardBlockedNoAction` fails open when `resolveSalesDocumentReasonCopy` returns `null`, so an unrecognised reason keeps the manual action rather than silently stranding the order — the safe direction.
- **CSS** — tokens only, no raw hex, no inline `style` left in the diff, and the two type sizes are on the canonical scale (13 px / 12 px). Skeleton headline height reserved so the section does not jump on load.
- **Demo telemetry** — `demo_fiscal_register_attempted` is registered in `DemoEventCatalog` with a group, not fired ad hoc.
- No schema change, so **no migration is owed**.

## IMPORTANT

**1. `shared/auth/is-admin-session.ts` duplicates a primitive that already exists, and its docblock asserts something false.**

`shared/auth/use-permission.ts:83` already exports `useIsAdmin()` (#2342), with its own `const ADMIN_ROLE = 'admin'` at line 92 and its own tests in `use-permission.test.ts:154`. `router-readiness-panel.tsx` and `order-hold-panel.tsx` already consume it — the latter's comment even reads "`useIsAdmin`, never an inline `role === 'admin'`". `SalesDocumentPanel` is an ordinary component, so `const isAdmin = useIsAdmin();` would have been one line, with **no** new file, no second role literal, and no added `useSession` import.

The previous review asked for "one shared helper … so the literal exists once"; that request was met by creating a *second* one. `frontend-architecture.md § Access Control And UI Visibility` closes on exactly this rule — *"There is no `useAccess` hook: it would return the same boolean as `usePermission`. One primitive per shape, no overlapping surface."*

Compounding it, the new file's comment `/** The one occurrence of the role literal in `apps/web`. */` is not true. `grep -rn "role === 'admin'" apps/web/src` returns ten non-test call sites (`app-shell.tsx:253`, `command-palette-provider.tsx:182`, and eight `use-*-query.ts` admin gates), plus `use-permission.ts`'s own `ADMIN_ROLE`. A comment claiming a single-source-of-truth that does not hold is worse than no comment: the next author trusts it and skips the grep.

**Fix:** delete `is-admin-session.ts` + its test, call `useIsAdmin()`. If a session-taking (non-hook) variant is genuinely wanted, export it *from* `use-permission.ts` and have `useIsAdmin` call it, so there is still one literal.

**2. The `Resend` affordance is offered where the provider structurally cannot resend, and the failure reads as transient.**

The button renders on `invoice.regulatoryStatus === 'rejected' && invoiceWrite.visible`, with no check of whether the connection's adapter implements `RegulatoryResubmitter`. Only **inFakt** does — `grep -rln resubmitForClearance libs/integrations/` returns the inFakt package alone; KSeF implements `RegulatoryTransmitter` and Subiekt implements `RegulatoryStatusReader` + `CorrectionIssuer`. So on a **KSeF** connection — the provider this button's own copy and route name — `POST /invoices/:id/resend-to-ksef` answers **501 NotImplemented**, and `onError` announces the generic `'The resend could not be sent.'`, which is indistinguishable from a 502 or a dropped network. The operator retries a structural refusal forever.

This is the defect class #2560 exists to prevent ("a hard block hides the retry action rather than offering one that would 403/409"), applied to the not-issued states but not to this one. It is also the `unsupported` vs `still-unknown` distinction ADR-042's #2520 amendment draws for exactly this reason: *reporting a structural refusal as transient tells an operator to keep retrying something that will never work.*

`RegulatoryResubmitter` is guard-only (absent from both plugin manifests), so the frontend cannot pre-gate it today — which makes the cheap, honest fix the right one: branch on `ApiError.status === 501` in `onError` and announce something like *"This provider cannot re-send a document; issue a correction instead."* Hiding the button behind a declared capability is the follow-up, not this PR.

(For the record, the admin question raised earlier is fine: the route is `@Roles('admin')` and `invoices:write` is admin-only in `ROLE_PERMISSIONS`, so `invoiceWrite.visible` and the role gate coincide. No 403 gap.)

## SUGGESTION

- **The description is stale on two counts.** It says the `isAdmin` inline comparison is "filed as a follow-up rather than blocking" — the helper is in this diff. And it says M9 targets `main` directly; the base is `phase/sdux-m8-orders-cell`, so this cannot merge before #2761. Say so, since the stack's merge order is not otherwise recoverable from the PR pages.
- Test-plan checkboxes are still unticked. Please confirm `pnpm --filter @openlinker/web test` is green (`reference_apps_web_flaky_full_suite` — retry rather than `--no-verify` if it flakes).

## Assessment

Substantively the best-argued PR in this stack: the issuance/clearance split is modelled rather than flattened, the two new derivation modules are pure and independently tested, the live-region re-announcement problem is a real one solved correctly, and the panel consistently declines to invent facts the backend does not supply (no elapsed counter on a contended attempt, no clearance step for a `not-applicable` regime, no em-dash placeholder for an unassigned number). Nothing here is BLOCKING.

🔄 **Approve with changes.**

### Priority
1. Replace `isAdminSession` with the pre-existing `useIsAdmin()` (or fold one into the other) — and drop the false "one occurrence" comment.
2. Distinguish the 501 on the resend path so a structural refusal does not read as a transient failure.
3. Refresh the PR description (helper is landed; base is M8, not `main`) and confirm the web suite.

norbert-kulus-blockydevs:
Addressed both IMPORTANT findings from the review at `9451e75` (commit `1df2ab7`):

**1. `isAdminSession` duplication — fixed.** Deleted `shared/auth/is-admin-session.ts` + its test; `SalesDocumentPanel` now calls the pre-existing `useIsAdmin()` (#2342) from `use-permission.ts`. One role literal, no new file, no `useSession` import.

**2. Resend button reads a 501 as transient — fixed.** The `onError` handler on the KSeF resend mutation now checks `error instanceof ApiError && error.status === 501` and shows a distinct message ("This provider cannot re-send a document; issue a correction instead.") instead of the generic "The resend could not be sent." Added two test cases covering both branches. Pre-gating the button on `RegulatoryResubmitter` stays a follow-up, as you noted, since the capability is guard-only and not advertised in any manifest today.

Also refreshed the PR description: the `isAdmin` follow-up note was stale (the helper landed in the previous round), and it now states the real base (`phase/sdux-m8-orders-cell`, not `main`).

Verified: `pnpm --filter @openlinker/web test` (targeted + full `pnpm test` suite via smart-test in the pre-commit hook), `pnpm --filter @openlinker/web run type-check`, and full `pnpm lint` (including `check:invariants`) all green.

piotrswierzy (changes requested):
Re-review at `ad12d4b`. The branch was force-pushed and rebased: the base is now **`main`** (which already carries the merged M8), not `phase/sdux-m8-orders-cell`, so this pass reads `origin/main...origin/phase/sdux-m9-order-panel` — 12 files, +895/-93. The only substantive delta since the approve at `9451e75` is `ad12d4b` itself.

## Verification of the two prior IMPORTANT findings

**1. `is-admin-session.ts` duplicating `useIsAdmin()` — RESOLVED (structurally), with the false claim recycled.**

`shared/auth/is-admin-session.ts` and its spec are deleted; `sales-document-panel.tsx:91` now imports `useIsAdmin` alongside `useWriteAccess` from `use-permission.ts`, and `:250` is `const isAdmin = useIsAdmin();`. The `useSession` import is gone. Confirmed the two are semantically identical — `use-permission.ts:83-86` is `session.status === 'authenticated' && session.user?.role === ADMIN_ROLE`, byte-for-byte the predicate the deleted helper carried, so no behaviour moved. "One primitive per shape, no overlapping surface" holds again.

The one thing that did **not** get fixed is the claim itself. The deleted file said *"The one occurrence of the role literal in `apps/web`"*; the replacement comment at `:246-249` says *"`useIsAdmin()` is the one place `role` is compared against `'admin'` in `apps/web`."* Same assertion, new location. It is still false:

```
$ grep -rn "role === 'admin'\|role !== 'admin'" apps/web/src --include=*.ts --include=*.tsx | grep -v test
```
returns 16 non-test call sites — `app/app-shell.tsx:253`, `app/command-palette-provider.tsx:182`, seven `features/*/hooks/use-*-query.ts` admin gates, and six `pages/**` guards (`sales-documents-page.tsx:19`, `settings-page.tsx:15`, `operational-settings-page.tsx:105`, `ai-provider-settings-page.tsx:25`, `mcp-tokens-page.tsx:19`, `prompt-templates-list-page.tsx:176`). The previous review's objection to the claim was that "the next author trusts it and skips the grep" — that is still exactly what this comment invites. **[SUGGESTION]** Trim it to what is true: *"the panel's own role check goes through `useIsAdmin()` (#2342), never an inline comparison"* — a statement about this file, which is what the comment is for. (Declining the 15-file sweep in an M9 diff is right; only the comment is at issue.)

**2. The `Resend` button reading a 501 as transient — PARTIALLY RESOLVED. This is why I am not approving.**

The branch is there and is correct as far as it goes: `sales-document-panel.tsx:829-847` tests `error instanceof ApiError && error.status === 501` and routes to a distinct string, with two tests pinning both arms. The reasoning is recorded in place and cites the right precedent (#2520's `unsupported` vs `still-unknown`).

**But the distinction is announced into an `sr-only` region and nowhere else.** `:704` is `<p className="sr-only" role="status">{liveAnnouncement}</p>`, and all four resend branches — `resending`, `resent`, the new `resendUnsupported`, and `resendFailed` — call `announce()` **only**. Every other mutation in this same file pairs the announcement with a visible `showToast`: `issueOn` success/error (`:436`, `:451`), the fiscal register path (`:515`, `:535`, `:544`), the reconcile outcomes (`:565`, `:574`), and the two at `:1004`/`:1019`. The resend path is the sole exception.

The consequence is that a sighted operator on a KSeF connection clicks **Resend**, the button un-disables, the red alert stays exactly as it was, and **nothing on screen changes** — for the 501 and for a 503 alike. The finding this fix answers was *"the operator retries a structural refusal forever"*; a message only a screen reader receives does not stop that operator retrying forever. The 501/other split is real work and correctly reasoned, but it currently lands in a channel most of the affected operators cannot see.

Two honest ways out, either fine by me:
- add `showToast` alongside `announce` on the resend `onError` (both arms) and `onSuccess`, matching this file's own established pattern — smallest change, and it makes the resend path consistent with its seven siblings; or
- render the 501 outcome into the alert itself (replace the "check the connection's configuration, then resend" body with the correction guidance once the provider has said it cannot resend), which is stronger, since the alert's current copy actively instructs the retry the 501 has just ruled out.

I would take the second, or both — the alert telling the operator to resend while the only refutation is invisible is the specific mismatch worth closing. Pre-gating on a declared `RegulatoryResubmitter` remains a follow-up, as agreed.

## Step 3/4 — full pass on the current diff

Frontend rules, checked against `frontend-architecture.md` / `frontend-ui-style-guide.md`:

- **Dependency direction** — clean. Every cross-feature import goes through a public barrel: `../../invoicing`, `../../fiscalization`, `../../connections`, `../../demo`, `../../system`, `../../sales-documents`. `resolveInvoiceLifecycleSteps` was added to `features/invoicing/index.ts` rather than deep-imported, and `sales-document-headline.ts` type-imports `InvoiceRecord` / `FiscalRegistrationRecord` from the barrels too. No `pages/` or `app/` import from `features/`. Nothing added under `shared/` this round (the one file that was, is deleted).
- **State ownership** — server state is TanStack Query throughout (`useOrderInvoiceQuery`, `useOrderFiscalRegistrationsQuery`, `useFiscalRegistrationProgressQuery` + mutations); local `useState` holds only view state (`liveAnnouncement`, `contendedAt`, `missingNumbering`, `switchTargetId`). No ad-hoc global store.
- **CSS** — `index.css` only, no `style={{}}` anywhere in the diff, no raw hex; `var(--space-*)`, `var(--border-subtle)`, `var(--text-link)`, `var(--text-secondary)`. Both new font sizes are on the canonical scale (`0.8125rem` body-small, `0.75rem` metadata). `.sales-document-panel__skeleton--headline` reserves 38px so the section does not jump on load.
- **Accessibility** — `DocumentHeadline` carries the state as a word, never colour alone (`tone: 'done'` draws the tick; `resolveInvoiceHeadline` returns a `state` string on every branch). Disclosures are real `<details>/<summary>`. The `role="status"` / `aria-live` duplication was dropped and the zero-width-space toggle in `announce()` fixes the repeated-message hole. The one a11y gap is the mirror of finding 2: the live region is the *only* channel for the resend outcome, which is the inverse of the usual defect — sighted users are the ones underserved here.
- **Mobile/tablet** — `.sales-document-panel__scope-note` is `width: 100%` so it takes its own row below a wrapped flex action group, and `summary { width: fit-content }` keeps the disclosure hit area from spanning the row. Nothing new introduces a fixed width or a horizontal overflow; the panel inherits the existing responsive container.
- **Derivation modules** — `sales-document-headline.ts` and `invoice-lifecycle-steps.ts` are pure, React-free, take `t` as a parameter (the pair now agrees, per the previous round) and are directly specced (12 tests, both files green locally). The `not-applicable` one-step trail and the `identityOf` no-em-dash rule are both the right kind of restraint: they decline to render an axis or a value the record does not carry. `resolveInvoiceHeadline` deciding on `status` first and letting `regulatoryStatus` override is what makes "issued, then rejected by the authority" expressible, which is ADR-065's whole point.
- **Testing** — `sales-document-panel.test.tsx` covers the three gating branches (contended, `canOverride` admin/non-admin pair, `hardBlockedNoAction` with the `missing-tax-rate` carve-out) plus the two new resend arms. Naming follows *should [behaviour] when [condition]*. Note the two resend tests assert on `findByText` of the announcement — which passes precisely because the message is in the DOM, sr-only; they would keep passing if the visible surface never gains one, so they do not protect the property finding 2 is about.
- **No schema change**, so no migration is owed. `demo_fiscal_register_attempted` is registered in `DemoEventCatalog` with a group rather than fired ad hoc.
- **CI** at this head: Lint, Type Check, Build, Docker smoke, PHP unit all green; Test and Integration Tests still running at time of review. Please confirm `pnpm --filter @openlinker/web test` before merge (I could not run the panel spec locally — unrelated `posthog-js` resolution in my checkout; the two pure specs pass 12/12).

## Assessment

Still the strongest PR in this stack, and both prior IMPORTANTs were engaged with seriously — the duplicate primitive is gone, and the 501 distinction is real and well-argued. I am requesting changes on one point only: the distinction currently exists solely in an `sr-only` region, so the defect it was raised to fix (an operator retrying a structural refusal) is not actually closed for the operators who hit it. That is a small change — a `showToast`, or the alert body — and once it lands this is a straightforward approve.

🔄 **Request changes.**

### Priority
1. Surface the resend outcome visibly — `showToast` on both `onError` arms (and `onSuccess`) to match this file's other seven mutations, and/or replace the alert's "then resend" instruction once a 501 has ruled it out. Extend one of the two new tests to assert the visible surface, not just the live region. **(IMPORTANT)**
2. Correct the "one place `role` is compared against `'admin'`" comment at `:246-249` — 16 inline sites remain, and the previous review flagged this exact claim. **(SUGGESTION)**
3. Confirm the web suite and the two in-flight CI jobs are green before merge.

norbert-kulus-blockydevs:
Pushed `3534d5236` addressing the CHANGES_REQUESTED review at `ad12d4b`.

**IMPORTANT — resend outcome now reaches a visible surface.** Both `onError` arms (the 501 structural-refusal branch and the generic-failure branch) and `onSuccess` now pair `announce()` with `showToast()`, matching every other mutation in this file (`issueOn`, the fiscal register path, the reconcile outcomes). A sighted operator clicking Resend on a KSeF connection now sees a toast — `"Cannot resend"` / the correction-instead copy for the 501 case, `"Resend failed"` / the generic copy otherwise — instead of only an `sr-only` announcement nobody but a screen reader received. Extended both resend tests to assert on `findToastTitle`/`findToastDescription` rather than `findByText`, which previously passed only because the same string happened to also land in the live region.

**SUGGESTION — stale "one place" comment fixed.** Trimmed the claim to what's actually true of this file: *"This panel's own role check goes through `useIsAdmin()` (#2342), never an inline `role === 'admin'` comparison."* No claim about the rest of `apps/web`.

Verified: `tsc --noEmit` clean, `eslint` clean on both touched files, and `vitest run sales-document-panel.test.tsx` — 39/39 passing.

piotrswierzy (approved):
## Re-review at head `18b21ca` (was `ad12d4b`) — APPROVE

### Prior findings, verified item-by-item

**BLOCKING: every resend branch called `announce()` into an `sr-only` live region with no visible counterpart, so a sighted operator on a KSeF connection (501, no `RegulatoryResubmitter`) clicked Resend and saw nothing change — and the two new tests asserted on the sr-only node, so they did not protect the visible surface. — RESOLVED, both halves.**

*Production side* — in `sales-document-panel.tsx` every resend **outcome** branch now pairs the announcement with a visible toast:

- success (`onSuccess`): `announce(...)` + `showToast({ … })`
- 501 structural refusal: `announce(message)` + `showToast({ title: 'Cannot resend', … })`
- generic failure: `announce(message)` + `showToast({ title: 'Resend failed', … })`

The fourth branch (the pre-flight `invoice.clearance.resending` announcement) is deliberately announce-only, and that is correct rather than an oversight: the button carries `disabled={resendMutation.isPending || …}`, so the wait state already has a visible representation and a toast there would fire twice per click. The inline comment on the 501 branch now states the rule it is following ("this file pairs `announce` with `showToast`, and a sighted operator who sees nothing change after clicking Resend will keep retrying a refusal the announcement already made") — the reasoning is recorded where the next editor will read it, not just in this thread. That matches the panel's other six mutations, so the file no longer has two conventions.

*Test side* — the two tests in `describe('SalesDocumentPanel — resend to KSeF (#2763 review)')` now assert the **visible** surface. `findToastTitle` / `findToastDescription` (`apps/web/src/test/test-utils.tsx`) resolve via `screen.findByText(text, { selector: '.toast__title' | '.toast__description' })`, i.e. the rendered toast node, not the live region. The 501 test additionally asserts `screen.queryByText(/The resend could not be sent\./i)` is null — so it pins that a structural refusal does **not** read as a transient one, which was the actual defect #2763 set out to fix. This is the assertion that would now fail if someone dropped the toast back to announce-only, which is what "protected" means.

`createMockApiClient` gains a default `resendToKsef: vi.fn().mockResolvedValue(null)` so the affordance does not hit `undefined` in unrelated tests — right call, since a shared harness that only works for tests which remember to override it is a trap.

**Previously resolved, re-confirmed:** `apps/web/src/shared/auth/is-admin-session.ts` does not exist at this head (`git show` → `path does not exist`), and the panel resolves the role through `useIsAdmin()` from `shared/auth/use-permission`.

**SUGGESTION: the false "the one place `role` is compared against `'admin'`" claim recycled verbatim into a new comment. — RESOLVED.** The new comment at the `useIsAdmin()` call site now says only "This panel's own role check goes through `useIsAdmin()` (#2342), never an inline `role === 'admin'` comparison" — a true statement about this file, not a false one about the codebase. (For the record: `grep` still finds 19 non-test comparison sites in `apps/web/src`, so the original claim in `use-permission.ts`'s own docblock remains inaccurate — but that is #2342's text, outside this diff, and not this PR's to carry.)

### Step 3 — frontend rules

- **Dependency direction** (`docs/frontend-architecture.md`): every import in the panel is `features → shared` (`../../../shared/ui/*`, `shared/api`, `shared/auth`, `shared/i18n`, `shared/format`, `shared/config`, `shared/plugins`) or `features → features` through a barrel (`../../connections`, `../../demo`, `../../invoicing`, `../../sales-documents`, `../../system`). No `shared → features`, no `pages` import. Cross-feature access goes through the barrel rather than a deep path, which is the convention the rest of `apps/web` follows.
- **Deep-import restrictions** (`.eslintrc.js`): no `@openlinker/core` import anywhere in the diff — the browser bundle stays free of the core package (#591), and the two mirrored helpers it needs (`splitShippingAcrossRates`, `minorUnitExponentFor`) come from the `invoicing` feature barrel, i.e. the existing mirror guarded by `check-shipping-tax-split-mirror.mjs`, not a fresh copy.
- **State ownership**: server state is TanStack Query throughout (`useOrderInvoiceQuery`, `useConnectionsQuery`, `useIssueInvoiceMutation`, `useResendToKsefMutation`); local UI state is `useState`; no new global store. The new pure derivations (`sales-document-headline.ts`, `invoice-lifecycle-steps.ts`) are side-effect-free functions of already-fetched records with their own unit tests — the right place for that logic, and it keeps the component a renderer.
- **CSS tokens** (`docs/frontend-ui-style-guide.md`): the `index.css` diff introduces no raw hex — colours are `var(--text-link)`, `var(--text-secondary)`, `var(--border-subtle)`; spacing is `var(--space-2)` / `var(--space-3)`. `check-design-tokens.mjs` and `check-css-structure.mjs` both run in the green `Lint` job. The `--headline` skeleton is explicitly sized to the loaded headline's two-line height so the section does not jump on load — a small thing that is usually skipped.
- **Accessibility**: the state is carried by text, not colour alone (`DocumentHeadline` reports issuance and clearance as two named axes; the 501 toast has a distinct title *and* body); one `aria-live` region carries wait/outcome announcements, with the identical-message re-announce problem handled explicitly; the correction and routing-explanation disclosures are real `<details>`/`<summary>` with `cursor: pointer` rather than click-handlers on a `<div>`. The blocking finding above was precisely the *visual* half of this pairing, and it is now symmetric.
- **Mobile + tablet** (in scope by project convention): the additions are flow content inside the existing panel — no fixed widths, `width: fit-content` on the summary, `width: 100%` on `__scope-note` explicitly so it drops to its own row below a wrapped flex action group. That last one reads as deliberate small-viewport handling rather than an accident.
- **CI**: all 8 checks green at this head (`Lint`, `Type Check`, `Test`, `Build`, `Integration Tests`, `Docker Build Smoke Test`, `PHP Unit Tests`, `Detect changed paths`).

### Step 4 — grounding

The panel's behaviour matches what `docs/architecture-overview.md` records for this surface: the "in progress elsewhere" state renders no elapsed counter, consistent with `SalesDocumentInFlight` carrying a lower-bound `since` and deliberately **no** lease expiry (the panel does not invent a completion estimate the backend refuses to make); not-issued states gate on the real `SalesDocumentGateBlockReason` / `SalesDocumentUnresolvedReason` vocabulary so a hard block hides the retry rather than offering one that 403s; and the admin gate mirrors the server's `@Roles('admin')` on both write paths rather than being a client-side opinion. The documented KNOWN GAP in the file header (no `correctionOf` column, `findLatestByOrderId` returns one record, so a correction displaces the original in this view) is stated as a deviation with its cause rather than silently approximated — which is the right way to leave it for the follow-up backend read.

### Most important remaining note (non-blocking)

The resend **success** branch (`onSuccess` → `announce` + `showToast`) has no test, while both failure branches now do. Given that the whole point of this fix was "the operator must see something change", the happy path is the one branch where a silent regression would be least likely to be reported by a user (they would just assume it worked). A third case mirroring the two existing ones would be a few lines. Not worth holding the PR for.

Nothing blocking remains. Approving.

norbert-kulus-blockydevs:
Addressed both findings from the latest review at `ad12d4b`:

1. **[IMPORTANT] Resend outcome invisible to sighted operators** — `showToast` now fires alongside `announce()` on all three resend paths (success, 501 `resendUnsupported`, generic `resendFailed`), matching this file's other seven mutations. The 501 case gets its own toast title ("Cannot resend") and body explaining the provider cannot re-send and a correction should be issued instead, so an operator no longer sees the "then resend" alert instruction with nothing visibly happening after clicking it. Extended `sales-document-panel.test.tsx` to assert the visible toast title, not just the sr-only live region.

2. **[SUGGESTION] Stale "one place role is compared" comment** — corrected to scope the claim to this file only ("this panel's own role check goes through `useIsAdmin()` … never an inline comparison"), since the repo-wide grep still finds 16+ other inline `role === 'admin'` sites outside this diff's scope.

Rebased onto current `main` (no conflicts). Verified locally post-rebase: `sales-document-panel.test.tsx` — 39/39 passing, and `apps/web` type-check clean.

PR #2668, opened 2026-08-28
feat(analytics): the analytics page now shows shipping-excluded GMV, a full order cohort for Units Sold/AOV/Median, a shipping-excluded net-of-VAT Cancelled Value, a trend chart that follows the selected date range, and a page-level Net/Gross VAT-basis toggle wired end to end (including the headline Revenue card and channel table units), plus a display-currency picker, a Data Coverage panel with real remediation actions, and exclusion annotations across the KPI strip and channel/product tables — verified live against real order data via two independent full accuracy re-reviews, with sections reordered to sync -> needs attention -> data coverage and the metrics spec updated to document the toggle (#2452, #2457, #2463, #2467, #2479, #2892, #2893, #2894, #2895, #2899, #2903, #2906, #2907, #2908, #2909, #2910)

jakubretajczykBD:
## Summary

Merges the `epic/2452-analytics-currency-coverage` branch into `main`: the full Analytics-page epic (#2452) — a per-viewer display-currency picker, a Data Coverage panel with real remediation actions (currency FX restatement, tax-rate opt-in, product-matching drill-down), exclusion annotations on the KPI/channel/product surfaces, and a string of review-round correctness fixes found along the way (FX rate exposure, tax-rate parsing, a stuck-remediation-run recovery path, and analytics-page copy fixes).

## What shipped

- **Display-currency picker** (Phase 2 + 6) — synchronous, read-only re-render of `/analytics` in any currency present in the dataset, in "current rate" or "order-date" mode, without touching the stamped `reportingCurrency`/`reportingTotalAmount` columns.
- **Analytics Settings persistence** (Phase 3 + 6) — `GET`/`PUT /analytics/settings`, the settings dialog, and the tax-rate inclusion toggle.
- **Data Coverage detection + panel** (Phase 4 + 7) — currency-mismatch and tax-rate (A/B/C) drill-downs elevated from buried tooltip counts into an actionable panel, plus a new product-matching-error detector.
- **Remediation actions** (Phase 5) — bulk currency FX recalculation (trackable, cancellable run), the tax-inclusion query-time toggle, and a re-run-backfill trigger for category C.
- **Exclusion annotations** (Phase 8) — `GapMark` / `.excl-note` wiring on the KPI strip, channel table, and product table so an operator can see *why* a figure looks low, in place.
- **E2E verification** (Phase 9, PR #2858) — mockup-parity spec against every `data-state`, plus a reusable user-story/AC ledger (`docs/specs/analytics-e2e-user-stories.md`).
[...]

Issue #1985, opened 2026-08-03
[IMPL] Backend — order analytics read model (queryable line items, placed-at, currency, tax treatment)


piotrswierzy:
## Design source

**Product Design parent:** Part of #1976 — see spec at `docs/specs/product-spec-1976-analytics.md` (§ 1a; dependencies **[L] [T] [X] [G]**; § 6 shipping order step 2)

## Scope

**The largest and most important piece of the epic.** Every money metric on `/analytics` is blocked on it, and it is invisible to the user — so it is scheduled after the trust/attention half is already on screen.

Today all analytically-relevant order data lives inside `order_records.orderSnapshot` (JSONB): totals, currency, tax treatment, per-line prices and quantities, and the true order time. There are no money columns and **no order item rows at all** — "top products by units" currently means expanding a JSON array per query, unindexed.

Make order data analytically queryable, and resolve the four correctness decisions that determine what OL is willing to assert as true:

- **[L] Line items** — per-line product/variant, quantity and price become queryable rows, not JSON
- **[T] Placed-at vs. ingested-at** — `createdAt` is OL's ingestion clock; `placedAt` (the buyer's actual order time) exists only in the snapshot. Date-range filtering must use order time, or a poll catch-up will visibly distort a day's figures
- **[X] Currency** — carried per order; no FX exists anywhere. Mixed-currency data must never silently sum
- **[G] Gross vs. net** — ⚠️ **the most dangerous item in the epic.** `OrderTotals.taxTreatment` is per-order and *optional*: Allegro reports buyer-paid gross (`inclusive`), a shop like PrestaShop prices net (`exclusive`). Comparing them unnormalised overstates the marketplace by roughly the VAT rate — and produces a *plausible* number, not an obviously broken one. Channel mix is precisely the chart this breaks

## Acceptance criteria

- [ ] Order line items are queryable per product and per variant, joinable to catalogue entities, without JSON expansion
- [ ] Aggregates can be filtered and bucketed by the buyer's order time, with a documented fallback for records that carry no placed-at
- [ ] Currency is a first-class attribute of any aggregate; a query spanning multiple currencies cannot return a single summed figure
- [ ] Tax treatment is explicit per order, and the read model makes gross/net comparability decidable rather than assumed. Where a source did not assert it, that is represented as unknown — never defaulted silently
- [ ] Cancelled orders (see cancellation issue) are excludable from every aggregate
[...]

PR #2766, opened 2026-09-01
fix(web): expand Top Products detail inline instead of navigating away

### LINKED ISSUE #2765: [BUG] Frontend — Analytics Top Products row click navigates away instead of expanding inline
## Problem / Context

Reported by Julia Pietrzykowska (Slack, 2026-09-01): clicking a product in Analytics → Top Products
leaves the Analytics page entirely and lands you on the Products page; getting back means re-entering
Analytics and re-scrolling.

Root cause: `apps/web/src/features/analytics/components/product-sales-table.tsx` passes
`rowHref={(row) => `/products/${row.productId}`}` into the shared `DataTable`
(`apps/web/src/shared/ui/data-table.tsx`), which does a full router `navigate(href)` on row click.
That unmounts the whole `/analytics` page.

Analytics has no route-based tabs — `AnalyticsPage` (`apps/web/src/pages/analytics/analytics-page.tsx`)
is one stacked page (KPI strip, `ChannelSalesTable`, `ProductSalesTable`), so there's no literal "tab"
being lost. What's actually lost on unmount is `ProductSalesTable`'s own local sort toggle
(`sortBy`, a plain `useState`, deliberately not a URL param per the file's own header comment) and
scroll position — exactly what Julia is describing.

## Proposed Solution

Reuse `DataTable`'s existing `expandable` prop — the same pattern already shipped on the Products
list page (`apps/web/src/pages/products/products-list-page.tsx` → `ProductRowDetail`,
`apps/web/src/pages/products/product-row-detail.tsx`) — so a row click expands product detail
**inline**, under the row, instead of navigating away. No new UI primitive: this is a config change
on a component that already supports it.

In `product-sales-table.tsx`:
- Remove `rowHref`.
- Add an `expandable` config whose `renderDetail(row)` renders an inline panel:
  - quick links strip: "Open full product page →" (`/products/:id`) and "Edit content →"
    (`/products/:id?view=content`) — real, deliberate navigation links, unlike the accidental
    full-row navigation being removed
  - a small fields grid: Internal ID, SKU, Currency, Stock synced (same shape as
    `ProductRowDetail`'s `products-detail-fields`)
  - a per-channel net-sales/units breakdown, reusing the row's own channel data (already fetched for
    the table's channel columns) rather than an extra request

An approved, interactive HTML mockup of the proposed behavior (built with the app's real
`index.css` tokens and real `DataTable` / `panel` / `segmented-control` / `coverage-pill` component
CSS) is here: **https://claude.ai/code/artifact/4d032821-3f47-4aad-9ea8-1ad4331ce449** — it has a
toggle to compare "current" (navigates away, and shows exactly what state gets lost) against
"proposed" (inline expand, state preserved). Two other options (open in new tab; a side drawer à la
`EditOfferDrawer`) were considered and rejected in favor of this one.

Per `docs/frontend-architecture.md § UX Mockups`, the mockup must also be committed into
`docs/plans/mockups/analytics-top-products-inline-expand.html` in the implementing PR (not left as
an Artifact link only) — the PR should include this file.

## Dependency check — safe to implement independently of epic #2452

Jakub (`jakubretajczykBD`) is mid-flight on epic #2452 "Analytics — display-currency picker +
data-coverage remediation" (branch `epic/2452-analytics-currency-coverage`, tracking PR #2668).
I audited every phase PR merged into that branch (Phases 2–8; only Phase 9, E2E verification #2482,
remains open). **None of them touch**:

- `apps/web/src/features/analytics/components/product-sales-table.tsx`
- `apps/web/src/features/analytics/lib/top-products-view-model.ts`
- `apps/web/src/features/analytics/api/top-products.types.ts`
- `apps/web/src/shared/ui/data-table.tsx`

The epic *does* modify `apps/web/src/pages/analytics/analytics-page.tsx` (KPI strip, coverage panel,
currency picker, exclusion notes) and `apps/web/src/features/analytics/components/channel-sales-table.tsx`
— but this fix never needs to touch `analytics-page.tsx`, since `ProductSalesTable` is already mounted
there unchanged; only its own internal `rowHref` → `expandable` config changes. No merge-order
coordination with the epic is required.

## Scope change (2026-09-01, agreed in comments below)

The panel as originally specified showed catalogue identity plus the channel split already present in
`TopProductRow` — i.e. product-level figures. That is a worse answer than the page it replaces: an
operator who clicks a *sales* row wants to know **which variant** sold, and **where**. The panel
therefore also ships a real per-variant × per-channel sales read, which did not exist. The
**Assumptions** and **Acceptance Criteria** below are updated accordingly; the original wording is
kept in the Proposed Solution section above as the record of what was first specified.

## Classification

**Type**: Frontend + Backend (a new read the frontend needs)
**Layer**: Interface (feature component, HTTP controller) + Application (core aggregation) +
Infrastructure (two repository reads)
**File(s)**:
- `apps/web/src/features/analytics/components/product-sales-table.tsx` (+ test)
- `apps/web/src/features/analytics/components/variant-channel-matrix.tsx` (new, + test)
- `apps/web/src/features/products/components/product-detail-summary.tsx` (new, extracted, + test)
- `apps/api/src/analytics/**` — new `GET /analytics/top-products/:productId/variants`
- `libs/core/src/orders/domain/variant-sales-aggregation.ts` (new pure aggregator, + spec)
- `libs/core/src/orders/infrastructure/persistence/repositories/order-line-item.repository.ts` —
  `getVariantRanking` + `getVariantChannelBreakdown`
- `libs/core/src/inventory/**` — `IInventoryQueryService.findAvailabilityByVariantIds`
- `apps/api/test/integration/orders/top-product-variant-sales.int-spec.ts` (new)
- `docs/plans/mockups/analytics-top-products-inline-expand.html` (new, committed mockup)

**No migration.** No schema change, and no new index: both new reads filter `productId` plus the
`placedAt` range, which is exactly the leading-column shape the pre-existing
`IDX(productId, placedAt)` on `order_line_items` serves. `variantId` / `sourceConnectionId` appear
only as `GROUP BY` keys, never in a predicate, so the `(productId, variantId)` index this issue's
plan first promised would add write cost and change no plan. Recorded in the repository docblock.

## Dependencies

- None. Independently mergeable ahead of, alongside, or after epic #2452 (see dependency check above).

## Assumptions

- ~~The inline detail panel does not need a live per-variant/stock breakdown like
  `ProductRowDetail`; it only needs identity fields + the channel sales/units split already present
  in `TopProductRow`.~~ **Withdrawn (see Scope change above).** The panel does need a per-variant
  breakdown — but a *sales* one, not `ProductRowDetail`'s listing-management view: one row per
  variant, one column per channel, net sales + units per cell, plus per-variant stock availability
  as context. It is fetched lazily, only when a row is actually expanded, so the top-products list
  response is unchanged.
- Per-variant stock is read presence-preservingly: a variant no inventory master has ever synced
  reads as "unknown" and gets no badge, never a `0` that would assert a stock level nobody reported.
- Multiple rows may be expanded at once (matching `DataTable`'s existing `expandable` behavior on the
  Products list page) — no "accordion, one at a time" requirement was raised.
- The "Open full product page" / "Edit content" links inside the expanded panel are real navigations
  (deliberate, operator-initiated) and are explicitly out of scope for the "don't navigate away"
  fix — only the *accidental* whole-row navigation is being removed.

## Acceptance Criteria

- [ ] Clicking anywhere on a Top Products row (or its expand toggle) expands an inline detail panel
      under the row; it does **not** navigate away from `/analytics`
- [ ] The panel shows: Internal ID, SKU, Currency, Stock synced, and a variant × channel matrix of
      net sales + units for that product
- [ ] The matrix's Total row reconciles to the product row's own Net sales / Units and to the outer
      table's per-channel columns (same line-item sums, grouped one level apart)
- [ ] A variant no inventory master has synced shows no stock badge, not "Out of stock"
- [ ] The `variantId: null` bucket is labelled "Unassigned"; a non-null id the catalogue could not
      resolve says "Unresolved variant" and shows its raw id, so the two are never conflated
- [ ] A channel the product is not listed on says so once, in the column header, with a Publish
      affordance — and still reports any real sales the range holds for it (a channel delisted
      mid-range legitimately has figures)
- [ ] Both new repository reads carry a deterministic `ORDER BY`, asserted in tests
- [ ] The panel's "Open full product page" and "Edit content" links still navigate to
      `/products/:id` and `/products/:id?view=content` respectively
- [ ] `ProductSalesTable`'s sort toggle (`sortBy`) and page scroll position are unaffected by
      expanding/collapsing a row (no remount occurs)
- [ ] Mobile card view (`cardView` prop) keeps working — the existing `collapsibleDetail` card
      pattern already covers this; verify it still renders the same detail content
- [ ] `docs/plans/mockups/analytics-top-products-inline-expand.html` is committed to the repo,
      matching (or updated in the same PR to match) the shipped behavior
- [ ] Tests added/updated in `product-sales-table.test.tsx` covering: row expand/collapse, no
      navigation on row click, detail content rendering
- [ ] No architecture boundary violations: catalogue/stock enrichment stays in `apps/api`, the
      aggregation is a pure core function, and no adapter is resolved on this read path

norbert-kulus-blockydevs:
## Summary

Two things, one following from the other.

**1. Analytics → Top Products row click no longer navigates away from `/analytics`.** The row used to be a `rowHref` to `/products/:id`, which unmounted all of Analytics and lost the sort toggle and scroll position for what should be a quick look-up (reported by Julia in Slack). It now expands an inline detail panel under the row, reusing `DataTable`'s existing `expandable` prop — the same pattern the Products cockpit list already uses.

**2. That panel needed something worth expanding into, so it ships a real per-variant × per-channel sales read.** The first cut showed catalogue identity only, which is a worse answer than the page it replaced: an operator who clicks a *sales* row wants to know which variant sold, and where. The scope change was argued out in #2765's comments; the panel now renders one row per variant, one column per channel, real net sales and units in every cell, plus a Total row that reconciles to the product row's own figures because both come from the same line-item sums, grouped one level apart.

## What ships beyond the frontend

The panel's data did not exist, so this PR is **not** frontend-only:

- **New public endpoint** `GET /analytics/top-products/:productId/variants` (`apps/api`) — catalogue + stock enrichment lives here, the pure aggregation in core.
- **Two new repository reads** on `OrderLineItemRepository`: `getVariantRanking` (grouped by `li.variantId`, scoped to one product) and `getVariantChannelBreakdown` (grouped by `(variantId, sourceConnectionId)`). Both reuse the existing `buildTopProductsSqlFragments` rather than restating the FX/net-of-tax SQL, both are parameterised, and both carry a deterministic `ORDER BY`.
- **New pure core aggregator** `variant-sales-aggregation.ts` — folds the two reads into one `TopProductVariantsResult`, including the currency-merge rule (a mixed-currency fold reports `null`, mirroring the SQL's own `COUNT(DISTINCT currency) <= 1` guard) and the `unconvertedOrderCount` lower bound.
- **`IInventoryQueryService` widened** with `findAvailabilityByVariantIds` — presence-preserving, no zero-fill, so a variant no inventory master has ever synced reads as "unknown" rather than a red "Out of stock" badge asserting a stock level nobody reported. The existing zero-filling read stays correct for its own callers (publishing a quantity, where master is authoritative including 0).
- **New module edge** `AnalyticsModule → InventoryModule`.
- **New integration spec** `apps/api/test/integration/orders/top-product-variant-sales.int-spec.ts`, asserting the aggregation and the ordering against real Postgres.

**No migration.** Nothing changes schema, and no index is added: both new reads filter on `productId` plus the `placedAt` range, which is exactly the leading-column shape the pre-existing `IDX(productId, placedAt)` on `order_line_items` serves. `variantId` and `sourceConnectionId` appear only as `GROUP BY` keys, never in a predicate, so a `(productId, variantId)` index — promised in the issue's original plan — would add write cost and change no plan. The reasoning is recorded in the repository docblock so the dropped promise is not left dangling.

## Frontend extractions (no behaviour change)

- The shared identity block (quick links + fields grid) moved out of `product-row-detail.tsx` into `ProductDetailLinks` / `ProductDetailFields` (`features/products/components/product-detail-summary.tsx`), so the products cockpit's own expandable row and this table's new one render from one source.
- `VariantChannelMatrix` is shared between the desktop inline-expand panel and the mobile card detail, which previously carried its own inline copy of the per-channel list.
- `product-stock-status.ts` promoted into `features/products/lib` so both features can read it.

## Approved mockup

Interactive before/after mockup, reviewed and approved before implementation: **https://claude.ai/code/artifact/4d032821-3f47-4aad-9ea8-1ad4331ce449**

Also committed to the repo per `docs/frontend-architecture.md § UX Mockups`: [`docs/plans/mockups/analytics-top-products-inline-expand.html`](../blob/2765-analytics-top-products-inline-expand/docs/plans/mockups/analytics-top-products-inline-expand.html). States: `current-full-navigation`, `current-full-navigation-left-analytics`, `proposed-inline-expand`, `proposed-inline-expand-row-open` (`document.body[data-state]`).

## Independent of epic #2452

Verified against every phase PR of Jakub's in-flight epic #2452 (display-currency picker + data-coverage remediation, tracking PR #2668, phases 2-8 merged into the epic branch): none touch `product-sales-table.tsx`, `top-products-view-model.ts`, `data-table.tsx`, `order-line-item.repository.ts` or the analytics HTTP layer this PR extends. See #2765 for the full audit.

## Test plan

- [x] `pnpm lint && pnpm type-check` — zero errors (repo-wide, incl. `check:invariants`)
- [x] `pnpm test` — web suite green; `libs/core` orders suite green (536 tests)
- [x] Integration: `top-product-variant-sales.int-spec.ts` against real Postgres
- [x] New tests: row expand/collapse toggles the panel instead of navigating; the matrix's variant identity, Total-row reconciliation, coverage-column and delisted-with-sales cases; `ProductDetailLinks`/`ProductDetailFields` with a resolved, partially-resolved and unresolved product
- [x] Manually verified against the approved mockup's states

Closes #2765


norbert-kulus-blockydevs:
Updated the mockup to show multi-variant handling (mockup only, no code change in this PR): https://claude.ai/code/artifact/4d032821-3f47-4aad-9ea8-1ad4331ce449

A 5th product ("Running Jacket Pro", 3 variants) demonstrates:
- SKU column reads **"3 variants"** instead of a single code when the row has no SKU and the product has more than one variant (distinct from "No SKU", which means genuinely unknown).
- The expand panel adds a read-only **variant snapshot** (attributes, own SKU, current stock badge) below the identity fields — clearly separate from the net sales/units totals above it, since the analytics endpoint aggregates sales at the product level and has no per-variant split to show.

Also brought the mockup's channel breakdown and fields grid in line with what actually shipped in this PR (`ChannelBreakdownList`'s plain text lines, `ProductDetailFields`' "Source external ID" field) — an earlier bar-chart sketch and a stray "SKU" field in the mockup had drifted from the real component.

New `data-state` value: `proposed-inline-expand-multivariant-open`.

norbert-kulus-blockydevs:
Redesigned the multi-variant panel per feedback — replaced the catalog-only variant list with a real **variant × channel net-sales/units cross-tab**: https://claude.ai/code/artifact/4d032821-3f47-4aad-9ea8-1ad4331ce449

- One row per variant, one column per channel (same order as the outer table: Allegro, Store PL, Erli), each cell showing net sales (primary) + units (small, muted) — a real number, never a proportional bar or a placeholder.
- **Total row and Total column reconcile exactly** to this row's own Net sales/Units and to the per-channel figures already shown above the fold — verified arithmetically (not eyeballed): 5372+2999+1499=9870, 86+48+24=158, matches the outer row's Net sales (9,870 zł) and Units (158) precisely, and each channel column total matches that channel's own column in the outer table (Allegro 5990/96, Store PL 2940/47, Erli 940/15).
- Stock now rides inline in the variant's identity cell (attrs + SKU + stock badge) instead of a separate list — one row is "everything about this variant."
- Nested in the same `.data-table`/`.data-table__container` classes the outer table already uses, so it reads as a continuation of the existing system rather than a new visual language.

Still mockup-only, no code change in this PR yet.

norbert-kulus-blockydevs:
Fixed the inconsistency — every product now renders the **same panel shape**: https://claude.ai/code/artifact/4d032821-3f47-4aad-9ea8-1ad4331ce449

Root cause: single-variant products had `variants: null` and fell back to a separate plain-text summary (`ChannelBreakdownList`), while multi-variant products got the new matrix table — two different visual languages for the same concept.

Fix: every product now carries a real `variants` array — a single entry for a simple product, mirroring the real system's own rule ("a product without combinations produces a deterministic synthetic variant," per `docs/architecture-overview.md`). So the detail panel always renders the same variant × channel matrix:
- **Single variant** → a one-row matrix (its own synthetic variant), no Total row — nothing to sum against one row, so repeating it as a "Total" would read as a glitch, not reassurance.
- **Multiple variants** → the identical table, more rows, plus a Total row that reconciles to the figures already shown above.

Same headers, same column order, same cell format (net sales + units stacked), every time — only the row count changes. Deleted the old plain-text code path entirely rather than leaving it as dead code.

norbert-kulus-blockydevs:
## Mockup finalized — implementing the real code now

The mockup (https://claude.ai/code/artifact/4d032821-3f47-4aad-9ea8-1ad4331ce449, committed at `docs/plans/mockups/analytics-top-products-inline-expand.html`) is settled: every product's expand panel renders one consistent **variant × channel net-sales/units matrix**, whether the product has one variant or several. Starting on the real implementation now. This is a **larger, full-stack change** than the original navigation fix already in this PR — writing down the plan before touching code, per request.

### Why this needs backend work, not just a frontend tweak

`GET /analytics/top-products` currently aggregates sales at **product** granularity only. Per-variant sales was a deliberately deferred, documented non-goal of that read (`libs/core/src/orders/domain/types/top-products.types.ts`: *"variant-level ranking is spec row C3, explicitly out of scope for this read"*). The good news: `order_line_items.variantId` has existed and been populated since #1985 — this is a **query change**, not a data-model backfill.

### Design decision: a lazy, per-product endpoint — not a bigger list response

`GET /analytics/top-products` returns up to 20 product rows per page; embedding a full variant × channel breakdown in every row's response would run that query 20× per page load, almost all of it thrown away unopened. Instead, this follows the exact lazy-detail precedent already in the codebase (`ProductRowDetail`'s own doc comment: *"mounted only when a row is expanded... so its queries fire lazily"*): a **new, separate endpoint**, queried only when a row is actually expanded.

### Backend

- `OrderLineItemRepositoryPort` (`libs/core/src/orders/domain/ports/`) — new method `getVariantChannelBreakdown`, scoped to one product, grouping by `(variantId, sourceConnectionId)` — reuses the same FX/net-sales SQL fragments `getProductChannelBreakdown` already has.
- New repository query in `order-line-item.repository.ts`, plus a new pure domain aggregator (mirrors `top-products-aggregation.ts`).
- New application service method + DTO + controller route: `GET /analytics/top-products/:productId/variants` (same `from`/`to`/`sourceConnectionId` query params as the list endpoint).
- Enrichment inside the service: variant identity (sku/attributes) from the existing products catalog read, and stock from the existing `IInventoryQueryService.getAvailabilityByVariantIds` (already used elsewhere for bulk/variant-keyed reads).
- **Stock badge scope cut**: `getAvailabilityByVariantIds` returns a raw `totalAvailable` count — there's no existing "low stock" threshold concept anywhere in the inventory domain to borrow. Rather than inventing an arbitrary number, the real implementation ships **binary in-stock/out-of-stock** (`totalAvailable > 0`), not the mockup's three-state ok/low/out. A real low-stock threshold is a separate, larger decision (there's already an unrelated per-connection `stockZeroThreshold` publishing knob — not the same concept) and doesn't belong bundled into this change.
- **Null-`variantId` edge case**: some historical order lines can carry `variantId = null` even though the product has a synthetic variant row. Policy: roll those into the product's sole variant when it has exactly one (the common case); otherwise report a separate "Unassigned" bucket rather than guessing or dropping the sale.
- Migration: an index on `order_line_items(productId, variantId)` (this becomes a live per-expand query, worth indexing) — via the standard `/migrate` workflow.

### Frontend

- New lazy query hook, firing only when a row is expanded (same pattern `ProductRowDetail` already uses for `useProductQuery`/`useInventoryQuery`).
- New `VariantChannelMatrix` component in `features/analytics/components/`, replacing `ChannelBreakdownList` inside `ProductSalesRowDetail` — full loading/error/data states. Reused for the **mobile card detail too** (not a separate simpler mobile view), so the "single vs multi-variant" consistency fix from the mockup carries over to desktop and mobile alike.
- New CSS (`.variant-matrix__*`), lifted from the mockup as-is, added under a bounded section comment in `index.css`.
- Tests: new component tests + updated `product-sales-table.test.tsx` for the new lazy fetch.

Will post progress here as it lands.

norbert-kulus-blockydevs:
## Shipped: real per-variant, per-channel net sales

Pushed the full implementation described in the plan above — matches the approved mockup exactly (`docs/plans/mockups/analytics-top-products-inline-expand.html`, no changes needed there).

**Backend**
- New `OrderLineItemRepositoryPort.getVariantRanking` / `getVariantChannelBreakdown`, grouped by `(variantId[, sourceConnectionId])`, scoped to one product. Reuses the exact FX/net-sales SQL fragments the existing product-level pair already uses (factored into `buildTopProductsSqlFragments`).
- New `IOrderRecordService.getTopProductVariantSales` + pure `buildVariantSales` aggregator.
- New lazy endpoint `GET /analytics/top-products/:productId/variants` — deliberately **not** embedded in the paged `/analytics/top-products` list response (that would run this query 20× per page load for detail almost none of it gets opened); fetched only when a row is expanded, same lazy pattern the products cockpit's `ProductRowDetail` already uses.
- Variant catalog metadata + live stock (`IInventoryQueryService.getAvailabilityByVariantIds`) composed at the apps/api layer.
- `order_line_items.variantId` can legitimately be `null` (a documented historical edge case). That "Unassigned" bucket is folded into the product's sole real variant when it has exactly one (the only case where there's an unambiguous target), otherwise reported as its own row — never dropped, never guessed at.
- Verified against **real Postgres** via a new integration test (`top-product-variant-sales.int-spec.ts`) — the GROUP BY/FILTER SQL, the null-variantId bucket, and unstamped-order currency handling all pass.

**Frontend**
- New `VariantChannelMatrix` component, used in **both** the desktop expand panel and the mobile card detail — same table shape either way: a single-variant product is a one-row matrix (no Total row, nothing to sum against one row), a multi-variant product is the identical table with more rows plus a Total row that reconciles to the figures already shown above.
- New lazy query hook, fetching only when the row is actually expanded.
- `product-stock-status.ts` promoted from `pages/products` to `features/products/lib` — the matrix's stock badges are a second real cross-feature consumer, so it's no longer page-local.

**Quality gate**: full repo `pnpm lint` clean; type-check clean across `libs/core`, `apps/api`, `apps/web`, `apps/worker`; unit tests green everywhere (117 API suites/1513 tests, 274 core suites/3605 tests, 388 web suites/4085 tests); new integration test green against real Postgres.

norbert-kulus-blockydevs:
## Review addressed + CI green again

Two pushes: `b84193f8e` fixes CI, `90af160ab` addresses all seven review findings.

### CI was red for one reason

`product-stock-status.ts` moved from `pages/products/` to `features/products/lib/` — one level deeper — but kept its two-level `../../shared/ui/*` imports. Those resolved to nothing, which failed **Type Check, Lint and Build** at once. Fixed by correcting the depth; nothing else was wrong with the promotion.

### Findings

**1. A variant with no inventory rows showed a red "Out of stock" badge.** `IInventoryQueryService.getAvailabilityByVariantIds` zero-fills unknowns, so `null` was unreachable and a product whose variants no `InventoryMaster` has ever synced expanded to a matrix where *every* row claimed out-of-stock — a positive false claim about the seller's stock, where the design intends no badge at all. Added the presence-preserving `findAvailabilityByVariantIds` (no zero-fill; a pass-through whose whole value is what it does *not* invent) and read that instead. The zero-filled read stays correct for its existing callers, which publish a quantity — there master is authoritative including `0`. Filtering on `locationCount > 0` was considered and rejected: `inventory_items.locationId` is nullable, so a real row can legitimately count 0 locations and would have lost its badge.

**2. `getVariantRanking` had no `ORDER BY`.** Postgres returns `HashAggregate` output in unspecified order, so the panel's rows were not stable between two identical requests, while the port doc called them "ranked". Now mirrors the product-level twin: `revenue DESC`, then `variantId ASC NULLS LAST` so the Unassigned bucket cannot land between two equal-revenue variants. Asserted against real Postgres, seeded lowest-revenue-first with the null bucket in the middle so insertion order cannot be mistaken for the test passing.

**3. A real variant read as "Unassigned" when the catalog lookup missed.** Variant ids come from `order_line_items`, which retain ids for variants the catalog no longer has (delete-then-recreate leaves stale mappings behind), so a genuine variant fell through to the literal `'Unassigned'` and became indistinguishable from the unattributed bucket. The label is now keyed off `variantId === null`; a non-null id that failed to resolve says "Unresolved variant" and shows its raw id, which is the only handle an operator has to chase it.

**4. Folding the Unassigned bucket labelled a mixed-currency sum with one arbitrary currency.** The `??` chains ignored that the SQL deliberately returns `NULL` for a set mixing native currencies. One `mergeCurrency` helper now returns `null` when the two sides disagree, at both row and channel level — an amount with no single currency is unrenderable, which is the correct outcome.

**5. `unconvertedOrderCount` double-counted on the fold.** Both operands are `COUNT(DISTINCT orderRecordId)` computed inside their own `variantId` group, so one order carrying both an assigned and an unassigned line — exactly the mixed historical shape the fold exists for — was counted once on each side. The union size cannot be recovered from two group counts, so this reports the larger: a true lower bound that never overstates a data-quality problem. Stated in the code rather than left to be rediscovered.

**6. The mobile card lost the per-channel coverage signal.** Cards have no channel columns, so the old text list was the *only* place a phone ever surfaced "Not listed" and its Publish affordance — replacing it with the matrix dropped both, and the remediation became unreachable on mobile. The affordance is extracted to `channel-publish-action.tsx` (taking a product id + name rather than an analytics row) and the matrix renders it in the column header — once per column, never per cell, since "not listed on this channel" is a fact about the product and the channel, not about a variant.

**7. The matrix contradicted the row above it.** For a not-listed channel every cell said "No Net sales figure in range" while the collapsed row said "Not listed" with a Publish action. The not-listed set is now resolved once, in the table, gated on `coverageGapAvailable` exactly as `ChannelCell` is, and passed down — so the two surfaces cannot disagree. Cells in such a column defer to the header rather than rendering a `0` that reads as a real zero-unit channel.

### Gate

`pnpm lint` clean repo-wide (0 errors), `pnpm type-check` clean across every package. Affected suites run green: 23/23 web (`variant-channel-matrix`, `product-sales-table`), 16/16 `top-products.service.spec.ts`, 110/110 core inventory + stock-at-risk. Every finding carries a regression test; finding 2's is an integration test against real Postgres.

piotrswierzy (commented):
## Tech-lead review

Backend is genuinely good work — parameterised SQL, deterministic `ORDER BY`, the `buildTopProductsSqlFragments` de-duplication, `findAvailabilityByVariantIds` (absent ≠ `0`), the currency-merge and `unconvertedOrderCount` lower-bound reasoning, and layering (catalog/stock enrichment at `apps/api`, pure aggregator in core) all hold up. No migration is needed (no entity change). Findings below.

### BLOCKING

**1. CI is red.** `Test` and `Integration Tests` both failed on `90af160`, while the PR body ticks `pnpm test` and the linked issue comment claims everything green. Fix or explain before merge.

**2. `apps/web/src/features/analytics/components/variant-channel-matrix.tsx:243`** — the "not listed" check runs *before* the sales data, so a channel in `notListedConnectionIds` renders `EmptyValue` and **discards real net-sales/units** for that channel. `ChannelCell` (`product-sales-table.tsx:149`) deliberately does the opposite — data first, "Not listed" only when there is none — precisely because *currently* delisted and *sold nothing in range* are different facts. As written the collapsed row can show figures for a channel the expanded panel calls blank: the exact disagreement `resolveNotListedConnectionIds`'s comment says it prevents. Fix: render the cell whenever `variantChannelCellFor(...)` returns data; fall back to the not-listed treatment only when it does not (same order as `ChannelCell`). Same at line 275 for the Total row.

**3. `variant-channel-matrix.tsx:175-191, 302`** — the Total row suppresses per-channel totals for not-listed channels but `grandRevenue`/`grandUnits` still sum every variant, so the row does not reconcile across itself while the footnote asserts "Adds up to the Net sales, Units and channel figures shown above." Fixing (2) fixes this; otherwise the claim must be dropped.

### IMPORTANT

**4. PR description does not describe the PR.** It reads as a frontend-only expand/extract change ("Extracted the shared identity block…", "only touches files outside that epic's footprint"), but the diff ships a new public endpoint (`GET /analytics/top-products/:productId/variants`), two new repository reads, a new core aggregator, a widened `IInventoryQueryService`, a new integration spec and a new `AnalyticsModule → InventoryModule` edge. The scope change is well argued in issue #2765's comments — but a reviewer reading the PR alone would not know to review any of it. Please restate the body. The issue's own AC and "Assumptions" (explicitly: *no live per-variant/stock breakdown needed*) should be updated too, since it is still open with the old scope.

**5. `variant-channel-matrix.tsx:175-190`** — `channelTotals` are recomputed on every render with three `reduce` passes per channel over every variant (O(channels × variants × channels)). Small today, but a single pass building a `Map` is simpler and removes the `.find()`-per-cell inside the loop.

### SUGGESTION

**6.** `getVariantChannelBreakdown` has no `ORDER BY`, so a variant's `channels[]` order is unspecified. The matrix indexes by connection id so it does not matter today; a deterministic order costs nothing and matches the ranking read's own reasoning.

**7.** The issue's plan promised an `order_line_items(productId, variantId)` index and it is not here. `IDX(productId, placedAt)` probably covers it — worth stating that explicitly rather than leaving the promise dangling.

**Verdict: request changes** — resolve 1-3, restate 4.

piotrswierzy (approved):
## Re-review at `ab8af22` — all three prior findings confirmed fixed

**1. CI red vs. green-gate claim — FIXED.** All 9 checks on `ab8af22` are `success` (Lint, Type Check, Build, Test, Integration Tests, Docker smoke, PHP, scaffolded adapters). The test-plan claim in the body now matches reality. (`mergeable_state: behind` — rebase before merge, not a review blocker.)

**2. `VariantChannelMatrix` suppressing real sales on a delisted channel — FIXED.** `variant-channel-matrix.tsx` now checks the cell data **first** and falls back to `Not listed on this channel` only when there is no channel row, in both the body cells and the `Total` row — with an in-place comment explaining that coverage is a claim about the listing *now* and is range-independent, so a mid-range delisting legitimately carries sales. The Total row therefore reconciles with the grand total beside it, which is what the footnote promises. `Not listed` is still stated once per column header, not per cell.

**3. Scope mismatch in the description — FIXED.** The body now has a dedicated *"What ships beyond the frontend"* section naming the new public endpoint, both repository reads, the pure aggregator, the widened `IInventoryQueryService`, the new `AnalyticsModule → InventoryModule` edge, the new int-spec, and an argued **no-migration** rationale (leading-column reuse of `IDX(productId, placedAt)`; `variantId`/`sourceConnectionId` are GROUP BY keys only). That reasoning is mirrored in the repository docblock, so the dropped index promise is not left dangling.

### Fresh pass on the current diff — nothing blocking

Positives worth recording: the SQL fragments are extracted into `buildTopProductsSqlFragments()` rather than pasted a fourth time; both new reads are parameterised with deterministic `ORDER BY` (`NULLS LAST` pinning the Unassigned bucket to the edge); `mergeUnassignedVariantBucket` folds only where a unique target exists and uses `Math.max` on `unconvertedOrderCount` (a true lower bound, not a double count); `mergeCurrency` returns `null` on a genuine conflict; `findAvailabilityByVariantIds` is presence-preserving so an unsynced variant reads `null`, never a false `Out of stock`; stock enrichment degrades rather than 500s; `ChannelPublishAction` uses `useWriteAccess` + `ReadOnlyLock` per `frontend-architecture.md § Access Control`.

**SUGGESTION** — `variant-channel-matrix.tsx` (per-channel totals fold): `running.currency = running.currency ?? channel.currency` is first-non-null, whereas the backend deliberately adopted null-on-conflict (`TopProductsService.mergeCurrency`). Unreachable today (all cells carry the same stamped reporting currency), but the two halves state different rules; matching the backend's would keep them from drifting.

Approving. Please rebase onto `main` before merging.

norbert-kulus-blockydevs:
Addressed the remaining suggestion from the re-review at `ecf557391`.

**Currency merge in `variant-channel-matrix.tsx`** — the per-channel totals fold used to take the first non-null currency (`running.currency ?? channel.currency`). It now nulls on a genuine conflict, mirroring `TopProductsService.mergeCurrency` on the backend exactly, so a mixed-currency sum can't render under a currency it doesn't fully belong to and the two halves can't drift apart.

Branch was already up to date with `main` (the fast-forward includes the "Merge branch 'main'" commit), so no rebase was needed on top of that.

All BLOCKING/IMPORTANT items (CI red, delisted-channel data suppression, Total-row reconciliation, PR scope description) were already fixed in earlier commits and confirmed in the approval review.

Issue #2610, opened 2026-08-27
[FEATURE] Frontend — expose stockSafetyBuffer, pricingRule and a zero threshold on the connection form


norbert-kulus-blockydevs:
## Problem / Context

`Connection.config.stockSafetyBuffer` is the **only remedy that works on the last unit**, because it does not depend on any propagation delay: the published quantity becomes `max(0, stock - reserve)`. It is implemented, tested, defaults to `0` — and `apps/web` has **no form field for it at all**, so it protects zero installations.

`Connection.config.pricingRule` is in the same position: implemented, no UI.

Two further gaps in the same area:

- the stock-restore path after a cancellation **wipes the buffer** on the first cancellation
- the second remedy the design named — publishing `0` below a threshold instead of the real number, for slow-moving stock — does not exist. One branch in an existing function.

## Proposed Solution

Add the connection-form fields for both settings plus the zero-threshold, and fix the restore path so it does not clear the buffer. All three live in the same form and the same helper.

## Classification

**Type**: Frontend / CORE
**Layer**: Interface / Application
**File(s)**: `apps/web/src/features/connections/**`, `libs/core/src/identifier-mapping/domain/types/stock-safety-buffer.types.ts`, `pricing-rule.types.ts`, the cancellation stock-restore path

## Acceptance Criteria

- [ ] Buffer, pricing rule and zero-threshold are editable on the connection form
- [ ] A cancellation no longer clears a configured buffer
- [ ] Below the threshold the published quantity is 0 as a deliberate operator choice — never as a side effect of a post-sale hold
- [ ] Tests added or updated for non-trivial logic

norbert-kulus-blockydevs:
Done in the epic branch `epic/2590-prestashop-no-module`, reviewed and fixed on PR #2627. Closing so the board reflects it; the branch is merged to `main` as one PR.

Issue #2313, opened 2026-08-22
[TASK] CORE — inventory_locations table, entity and repository port


piotrswierzy:
> Part of the OMS programme backlog (**Wave 1b**, stream S1, size M). Backlog slug: `W1b-1`. Sibling `W*` slugs resolve via the Wave-1b epic's checklist. Design of record: `docs/plans/analysis/DESIGN-oms-authority-model.md` + ADRs 052–062.

## Problem / Context

`locationId` on `inventory_items` is a bare nullable text column pointing at nothing — there is no location table anywhere in `libs/` or `apps/`. ADR-058 decision (1) makes locations first-class. REVIEW §3 D3 adds `countryIso2`, `postcode` and optional geo **from day one**, because the router's filters are unimplementable without them and the table is cheapest to get right while new.

## Proposed Solution

New `inventory_locations` in the existing `inventory` context (design §4.1 rejects a `locations` context: no independent lifecycle, and it would put a cross-context read on every ATP query).

Columns: `id` (`ol_location_*`, minted through `IdentifierMappingService`'s prefix convention — add `Location → location` to `ENTITY_TYPE_ID_PREFIX` if the default lowercase name diverges), `code` (operator-authored, unique), `name`, `kind` (`'warehouse' | 'store' | 'third-party' | 'virtual'`), `ownerConnectionId` (nullable — **provenance**, answering "whose sync may write positions here", deliberately *not* authority), `externalRef`, `status`, `countryIso2`, `postcode`, optional `latitude`/`longitude`.

Ship the ORM entity, domain entity, `LocationRepositoryPort` + repository, migration, tokens, module wiring, and the `@openlinker/core/inventory` barrel export. No FK from `inventory_items.locationId` is added in this wave (existing rows carry unattributable values; the FK is a step-(iii)-class change).

## Classification

**Type**: CORE
**Layer**: Domain + Infrastructure
**File(s)**: `libs/core/src/inventory/domain/entities/`, `libs/core/src/inventory/domain/ports/`, `libs/core/src/inventory/infrastructure/persistence/entities/`, `.../repositories/`, `apps/api/src/migrations/`, `libs/core/src/inventory/inventory.tokens.ts`

## Dependencies

- ADR-058 merged
- #2308 (`W1a-8` — the H13 conformance conventions this context's tables and exports follow)

## Assumptions

- `LocationNetwork` (Fluent's virtual-position scope) is specified in the design but explicitly deferred to v3 — not modelled here.
[...]

PR #2771, opened 2026-09-01
feat(inventory): refuse routing enablement with no location, and offer the fix (#2407)

### LINKED ISSUE #2407: [TASK] DX — router enablement guard + first-run location bootstrap
> Part of the OMS programme backlog (**Wave 3a**, stream S3, size M). Backlog slug: `W3a-22`. Sibling `W3a-*` slugs resolve via the Wave-3a epic's checklist.

## Problem / Context

REVIEW D3: routing needs locations that exist on no current install. Enabling a router against zero locations produces `unfulfillable` for every line — an operator-hostile first run that reads as a broken feature.

## Proposed Solution

The router **refuses to enable until ≥1 `inventory_location` exists**, with a first-run wizard offering to mint locations from the Wave-1b sentinel pool. Surfaced on the connection settings page beside the authority-status surface shipped in Wave 2.

## Classification

**Type**: DX + Frontend
**Layer**: Application + Interface
**File(s)**: `libs/core/src/inventory/**`, `apps/web/src/features/connections/**`

## Dependencies

- Wave-1b locations (#2313, `W1b-1`); #2405 (`W3a-16`)

## Acceptance Criteria

- [ ] An enable attempt with zero locations is refused with an actionable message, not a 500
- [ ] The wizard mints locations idempotently (re-run creates nothing)

piotrswierzy:
Closes #2407

REVIEW D3: fulfilment routing needs inventory locations that exist on no current install. `inventory_locations` (ADR-058 decision 1, #2313) ships empty with nothing seeding it, so enabling a router against zero locations made every line `unfulfillable` — a first run that reads as a broken feature rather than as an unconfigured one.

`ConnectionService` now refuses the `false → true` sourcing transition with a **400 naming the remedy**, and `POST /inventory/locations/bootstrap` mints one `MAIN` warehouse location idempotently as that remedy.

## Four load-bearing properties

**Enablement is decided from CONFIG, never from a capability list.** The sourcing claim is `config-only` and `FulfillmentRouter` is deliberately absent from `CoreCapabilityValues`, because `enabledCapabilities` is stamped at create and never retro-filled (#2085) — gating on a capability name would exempt every connection that already exists. The shipped `parseAuthorityConfig` is the only reader, which is also why **a config OpenLinker cannot read fails to MATCH rather than throwing**: the coercer's unrecognised-shape default is the frozen `UNHELD` claim, so an install that never opted in is byte-identical.

**The guard fires only on the transition, compared against the PERSISTED config**, and short-circuits on the claim before reading the database — so an unrelated patch on an already-claiming connection is not refused for a state the guard itself let through, and a connection with no claim pays no extra query.

**It is enable-time only, NOT an invariant.** `deleteLocation` and setting the sole location inactive both take an already-enabled install back to zero, and enforcing there would put a routing rule inside the *inventory* context — a cross-context edge ADR-053 forbids. The degenerate state is safe rather than corrupt, so `RouterReadinessPanel` **reports** it, in those terms: routing is on and is currently deciding nothing, an unfinished setup step with a one-click remedy — never "routing is silently inert". Badge tone is `neutral` for a never-configured install; amber is reserved for the degraded state.

**The bootstrap is an OFFER an operator takes, never a seed.** Nothing writes it from a migration or as a side effect of create/enable, because minting on enable would make the refusal unreachable. Idempotency is insert-then-recover on the unique code rather than a read-then-write, so two operators clicking at once cannot both create it. The minted row **locates no existing stock** — `locationId IS NULL` permanently means the master declines to locate its stock (ADR-058 decision 2) — so the panel's ready state says stock still needs assigning rather than implying routing will now succeed.

## Reviewer notes

**Declared regression — three fixtures, not a quiet edit.** Three tests in `authority-status.int-spec.ts` create a sourcing-claiming connection while `inventory_locations` is truncated, and so became 400s. Each gains **one** `seedLocation(token)` line; every projection assertion is unchanged. The fixture is what is stale — those specs are about the authority-status projection, not about enablement. Weakening the guard to `update` only was rejected: it would leave `POST /v1/connections` as an unguarded bypass, the #2610 "the form is not the only way in" failure.

**Deliberately NOT re-tested here: that two enabled routers disable routing.** Already shipped and asserted on both sides in `authority-status.int-spec.ts`. The assertion an earlier draft proposed called `selectPrimaryFulfillmentRouter` — a pure total function over an in-memory array — so it would have passed identically with this entire change reverted. Red-first evidence instead points at the guard's own behaviour, with **the identical request after a bootstrap as the control**, proving the 400 came from the location count and not a malformed payload.

**The one review finding worth highlighting.** The panel originally computed `routingClaimed` as a plain boolean, so while the sourcing status was merely *unknown* (loading or errored) it fell into the pre-enable branch and told an operator whose routing IS on that it "cannot be switched on". That is the same failure mode this issue exists to prevent, one level down — "not known yet" is not "not claimed". Now a third neutral branch; **verified red-first** (reverting the fix fails the new test, restoring it passes).

Six further tech-review findings applied: missing `router-readiness-panel__note` CSS; unexplained disabled button for a non-admin holding `inventory:write` (route is `@Roles('admin')`, permission is also an operator permission); a test named "hides the action" that asserted `toBeDisabled()`; badge tone; mutation-error reset.

**No migration, no ADR** — composes two shipped seams (`parseAuthorityConfig`, `ILocationService`) at an existing validation site.

## Gate

| Step | Result |
|---|---|
| `pnpm lint` + `check:invariants` | **0 errors**, all invariants green (`migration-timestamps: OK (154 migrations)` after #2768) |
| `pnpm type-check` | **0 errors** |
| `pnpm test` | **green** — web 4826 (+1 skipped), core 5690, api 1699, worker 688, plugins all pass. Run by the pre-commit hook end-to-end under Node 22. |
| `pnpm test:integration` (apps/api) | **159/160 suites, 1309 pass, 1 fail** |

The single integration failure is **pre-existing #2638**: `earliest-order-date.int-spec.ts` fails with a 1-hour TZ offset (`2026-02-10T00:00:00.000Z` vs `01:00:00.000Z`) under local TZ and passes under `TZ=UTC`. This diff touches nothing in that path — no file under `order-record`, `analytics-trust` or `earliest-*` is modified.

Targeted run of the specs this PR owns: `router-enablement-guard.int-spec.ts` + `authority-status.int-spec.ts` = **13/13 pass**. The new int-spec boots the full app, which is what proves the new `CoreInventoryModule` import into `IntegrationsModule` creates no DI cycle — proven by boot rather than by assertion.

**Known gap, stated rather than assumed:** `check-ui-vocabulary.mjs` `SCAN_ROOTS` covers neither `features/connections` nor the new panel, so epic #2412's P9 banned-word rule is honoured here by review, not by CI. `SCAN_ROOTS` is deliberately not widened in this PR — a wider scanner would surface pre-existing violations across surfaces this change does not own.

https://claude.ai/code/session_01QiTzZPch2eTt585HBxj6Dx


piotrswierzy:
## Self-review applied (`116667215`)

`/pr-review` found no BLOCKING issues. Three IMPORTANT items, all frontend, all fixed in `116667215`:

1. **The panel rendered on every connection's Health tab in all four states** — including `precondition met + nothing claims routing`, where its whole content was "Fulfilment routing can be switched on." That is noise on every connection page of every install, including connections that will never route. It now renders nothing for that one non-actionable combination; for a readiness panel an absence correctly reads as "nothing to do here". Every actionable state, and the confirmation of a claim the operator made, still renders.

2. **The unknown-claim copy asserted a failure that had not happened.** `routingClaimed === undefined` covers the in-flight read as well as the failed one, so "could not be read" was false during the ordinary loading tick. Reworded to state only what the branch knows.

3. **The new `activeLocations` query key was inserted between the pre-existing sorted-join comment and the `availability` key it describes**, leaving that comment reading as documentation for a key with no ID list and no join.

Both behavioural fixes carry an assertion **verified to go red with the change reverted and green with it**, against the changed code path — same discipline as the unknown-claim fix in the first commit.

One SUGGESTION not taken: the panel's file-header docblock sits after the imports rather than at the top. That matches the local precedent of `catalog-trust-panel.tsx` (its immediate neighbour on the same tab), so changing it here would make this file the odd one out rather than fix a pattern. Noted rather than applied.

### Gate after the fixes

| Step | Result |
|---|---|
| `pnpm lint` + `check:invariants` | **0 errors**, all invariants green |
| `pnpm type-check` | **0 errors** |
| `pnpm test` | **green** — `apps/web` 4827 passed (+1 skipped), core 5690, api 1699, worker 688, plugins all pass. Run end-to-end by the pre-commit hook under Node 22 in this worktree. |
| `apps/api` integration (targeted) | `router-enablement-guard.int-spec.ts` + `authority-status.int-spec.ts` = **13/13 pass** |

The second commit is frontend-only, so the earlier full-suite integration result (**159/160 suites, 1309 pass**) stands, with the single failure being the pre-existing local-TZ `earliest-order-date.int-spec.ts` (#2638) documented in the PR body.

Branch is rebased on the current wave tip `7e9e35932` with a clean working tree.

Issue #2736, opened 2026-08-31
Per-location delivery-method model, so the OMS router can declare a `method-capable` filter


piotrswierzy:
## Why this is deferred rather than built

#2408 (the OL fulfillment router, Wave 3a) was scoped to ship a closed, named filter vocabulary including **`method-capable`** — "eliminate a candidate location that cannot serve the order's requested delivery method". It ships without it, deliberately.

**There is no fact in the merged tree that filter could read.** Verified during #2408's planning:

- `InventoryLocation` carries no delivery-method column.
- `InventoryFilters` has no method axis.
- Nothing under `libs/core/src/inventory/domain` models delivery methods at all.
- The only per-method routing data in the repo is `mappings`' `FulfillmentRouting*` family — and that answers a **different question**: *which processor ships this?* (dispatch), not *which location can serve this method?* (routing candidacy). `fulfillment-router.port.ts`'s docblock says explicitly: "do not wire one into the other and do not rename either."

So declaring `method-capable` would have produced a filter that can never eliminate a candidate — green, selectable in config, and inert. That is the "declaration with no subject" failure: a **prohibition** is honest with no subject, a **declaration** is not. An operator authoring a routing rule would reasonably believe they had constrained routing by delivery method, and they would be wrong, with no error anywhere.

Note epic #2412 asserts that a `countryIso2` column is what makes the sibling `country-served` filter implementable. That reading was accepted for `country-served` (see the caveat below), but it does not extend to methods — there is no analogous column to lean on.

## What this issue owns

A per-location delivery-method capability model, plus its authoring surface, so `method-capable` has something real to read. Open questions for whoever picks this up:

- Is the fact **per location**, or per `(location, carrier)`? A location that can hand to one carrier but not another is the common real shape.
- Does it belong in `inventory` (where `InventoryLocation` lives) or in the OMS plugin's own tables? The router is a connection-backed plugin (ADR-054/055), so an OMS-local table is available and avoids a core edge — but a location's shipping capability is arguably a fact about the location, not about routing.
- What does an **unset** capability mean? Consistent with the repo's config convention, an unset value must leave existing installs byte-identical, which argues for "unset = no constraint" rather than "unset = serves nothing".

## Acceptance

- [ ] A per-location delivery-method fact exists, with an operator-facing way to author it
- [ ] `method-capable` is added to the router's closed filter vocabulary and **can eliminate a candidate** — proven by a test that fails when the filter is removed, not merely one that passes with it present
[...]

Issue #2390, opened 2026-08-22
[TASK] DX — `libs/oms` package scaffold + repo-wide walker/lint/composition wiring


piotrswierzy:
> Part of the OMS programme backlog (**Wave 3a**, stream S2, size M). Backlog slug: `W3a-1`. Sibling `W3a-*` slugs resolve via the Wave-3a epic's checklist. Design of record: `docs/plans/analysis/DESIGN-oms-authority-model.md` §9, ADR-055.

## Problem / Context

`@openlinker/oms` is a new first-party *product* package beside `libs/core` (design §9, ADR-055) — deliberately **not** under `libs/integrations/`, whose `integrations-*` prefix would read as "an adapter to somebody else's OMS" and collide with future `integrations-fluent`-style third-party adapters. Every repo-wide invariant script and lint scope currently enumerates `libs/core/**`, `libs/integrations/**`, `apps/{api,worker}/**`; a new top-level package no walker knows about silently escapes the cross-context contract, the deep-import ban and the workspace-dep check.

## Proposed Solution

1. `libs/oms/package.json` — name `@openlinker/oms`, `"private": true` (publishing decided at the **end of Wave 3**, design §9), authored to publishing standard from day one: complete `exports` map (barrel-only), `files` whitelist, `license`, own `README.md` as the product front door. Declare every `@openlinker/*` workspace edge in the right bucket (`engineering-standards.md § Workspace dependency declarations` — a tsconfig `references` entry is not a pnpm edge, #2011).
2. Walker scope additions: `scripts/check-cross-context-imports.mjs` gains `libs/oms/**` (treated like a plugin — every `@openlinker/core/<ctx>` import is by definition cross-context); `scripts/check-workspace-dep-declarations.mjs` covers the package; the `.eslintrc.js` integration-package rule blocks (deep `@openlinker/core/*` paths, `orm-entities` sub-barrels) gain `libs/oms/**`.
3. Composition seams: entries in `apps/api/src/plugins.ts` and `apps/worker/src/plugins.ts`; entries in `apps/api/src/plugin-migrations.ts` and `scripts/plugin-migration-dirs.json` (the Allegro precedent); FE contribution directory `apps/web/src/plugins/oms/` registered in the compile-time contribution registry (the Erli precedent).
[...]

PR #2774, opened 2026-09-01
feat(oms): the OL fulfilment router — closed named filters/sorts + rule coercer (#2408)

### LINKED ISSUE #2408: [TASK] OMS — OL router: closed named filters/sorts + `Connection.config.routing` coercer

piotrswierzy:
The first real implementer of `FulfillmentRouterPort`. `evaluate()` and `route()` call one pure pipeline and nothing else, so their explanations are provably identical rather than merely similar — `route()` adds a decision reference and never re-decides, which is also why it cannot leak a committing identifier.

## The vocabulary is closed, and a member is declared only when a fact source can make it select

`method-capable` is deliberately **absent** (#2736 owns the model): no per-location delivery-method data exists anywhere in the tree, so the filter could never eliminate a candidate, and an operator authoring it would believe they had constrained routing and be silently wrong. A prohibition is honest with no subject; a declaration is not.

## `afterAction` is a three-rung ladder, not a boolean

`line-split` was declared and read by nothing — the pipeline derived a single `allowQuantitySplit` boolean, so `line-split` ≡ `quantity-split`: a declared-but-inert vocabulary member, in the very file whose docblock argues against them. The same boolean made `no-split` too weak: it forbade splitting a line's *quantity* while still placing line A at one location and line B at another — two picks, two parcels, exactly what `no-split` asks not to happen.

The most restrictive rung any rule declares governs, so a permissive sibling can never lift a restriction the operator authored; an empty ruleset answers `quantity-split`, keeping an unconfigured install byte-identical to having no router at all.

## Two over-count bugs, both invisible to the conservation checker

`checkRoutingPlanConservesQuantities` knows what the **order** asked for, never what the **world** holds — so neither of these would have failed it:

- **Paged stock was summed, not accumulated by row id.** `InventoryRepository.findMany` orders by `updatedAt DESC` with no tie-break and a bulk sync writes many rows in one statement, so an offset read can return the same row on two pages. A skipped row under-counts (routes elsewhere, or reports the shortfall honestly); a double-counted one commits work rows for units that do not exist. Only one failure direction is safe.
- **Two lines can name the same variant.** Compared line-by-line, a location holding 6 units "covered" two 5-unit lines because each comparison saw the full 6, so `no-split` committed 10 units to a location holding 6. Demand is now summed per variant first, and `least-splits` delegates to the same predicate so the rank and the gate cannot disagree about what "one fulfilment" means.

## The uniqueness constraint existed in production and in no test environment

`UQ_oms_routing_rules_live_name` was declared only in the migration, but the integration harness builds its schema by `synchronize` off the ORM decorators — so the one constraint stopping a connection carrying two live rules for the same `(kind, name)` was unexercised everywhere. Declaring it on the entity too took the new int-spec from 7/8 to 8/8. This is the rule `inventory-locations.int-spec.ts` already states for `UQ_inventory_locations_code`; here a spec proves it rather than assuming it.

## Storage

Rules are rows, not a `Connection.config.routing` blob, per ADR-054's adopted storage amendment: a `RoutingExplanationStep` must cite a stable rule id that survives the operator reordering the list, which an array in a config blob cannot offer. The coercer therefore narrows an untrusted persisted **column** — which outlives the build that wrote it — and a row this build cannot understand is dropped, never guessed at and never thrown on.

The repository is bound behind `ROUTING_RULE_SOURCE_TOKEN` and the barrel publishes the port and the token, never the class or the ORM entity (the `allegro.tokens.ts` precedent — no plugin exports an ORM entity).

## Evidence

- Split ladder: reverting to the previous semantics fails **5 of 8** new specs; the other 3 guard behaviour both builds share.
- Same-variant over-commit: reverting the fix fails its spec.
- Partial unique index: 7/8 → 8/8 on the integration spec.
- `libs/oms` **81 passed** / 6 suites; affected `libs/core` specs **98 passed**; `oms-routing-rules.int-spec.ts` **8 passed**; `pnpm type-check` clean; the full pre-commit hook (lint + type-check + smart-test) passed on both commits.

Migration slot `1869000000200` is coordinator-assigned inside the wave band. `check-migration-timestamps` reports OK (155 migrations, 2 plugin dirs).

Closes #2408

https://claude.ai/code/session_01QiTzZPch2eTt585HBxj6Dx


piotrswierzy:
## `/pr-review` — systematic review

**Scope**: 27 files across `libs/oms/src/routing/**` (plugin), `libs/core/src/fulfillment` (one additive CORE read), `apps/api` (migration registration, harness truncate list, one int-spec).

### Findings already fixed in `d29b29dad` (raised by `/tech-review` on the diff)

**[BLOCKING] — `libs/oms/src/routing/evaluate-routing.ts` — same-variant over-commit.** `coversWholeOrder` compared each line against the full available quantity, so a location holding 6 units "covered" two 5-unit lines and `no-split` committed 10 units to a location holding 6. Neither `checkRoutingPlanConservesQuantities` nor any existing spec could catch it — the checker knows what the *order* asked for, never what the *world* holds. Fixed by summing demand per variant (`demandByVariant`); `least-splits` now delegates to the same predicate so the rank and the gate cannot disagree. Red-first verified.

**[IMPORTANT] — `routing-vocabulary.types.ts` — pure-rule exception not invoked.** The file grew a runtime function (`mostRestrictiveAfterAction`) but its header did not invoke `engineering-standards.md` § "The pure-rule exception to 'types only' (#2231)", unlike its sibling `routing-rule.types.ts`. Now states all three conditions.

**[IMPORTANT] — missing file headers** on the three spec files (`engineering-standards.md` § File Headers; 15/25 sibling `fulfillment` specs carry one). Added.

### Remaining findings

**[SUGGESTION] — `ol-fulfillment-router.ts` — class/file naming vs the adapter convention.** `engineering-standards.md` § Adapters specifies `{System}{Capability}Adapter` in `*-adapter.ts` for port implementers. **Not changed, deliberately**: this router is explicitly *not* registry-dispatched (`FulfillmentRouter` is absent from `CoreCapabilityValues` and every manifest per #2403, and it is reached through a factory), and the file's docblock spends a paragraph explaining why. Naming it `…Adapter` would assert exactly the registry membership that docblock denies. There is also no in-tree precedent — the only other `implements FulfillmentRouterPort` occurrences are core test fixtures.

**[SUGGESTION] — no `Logger` on `OlFulfillmentRouter`.** Not added: `OlFulfillmentRouterDeps` is deliberately dependency-light so the router stays trivially constructible in tests, and the observable behaviour exists one level up — #2395's committer persists a named refusal reason (`plan-carries-unfulfillable`) for exactly the outcome worth logging.

### Positive observations

- **`libs/oms/src/**` is in `check-cross-context-imports.mjs`'s walked scope (#2390)**, and the new code needed no allow-list entry: it reaches core only through `I*Service` interfaces (`IFulfillmentWorkQueryService`, `IInventoryQueryService`, `ILocationService`) and capability ports (`FulfillmentRouterPort`) — the two allowed shapes.
- **The new CORE read is additive on an existing `I*Service`**, not a new context edge, and does not expose a `*RepositoryPort` across the boundary.
- **The repository is bound behind a Symbol token** and the barrel publishes port + token only — the `allegro.tokens.ts` precedent. No plugin in the tree exports an ORM entity from its barrel, and this one now doesn't either.
- **Migration carries both `up()` and `down()`**, and the ORM entity declares the same partial unique index the migration creates — which is what the harness (`synchronize`, not migrations — testing-guide.md § Testcontainers Lifecycle) actually builds from.
- **The int-spec's raw SQL is fully parameterised** (bound `$n`, no interpolation).

### Verdict

✅ **Approve** — no BLOCKING or IMPORTANT items outstanding. The two remaining SUGGESTIONs are reasoned refusals, stated above rather than silently skipped.

PR #2781, opened 2026-09-01
fix(analytics): currency-safe cancellations, honest display-currency conversion, recalculation visibility (epic #2452)

jakubretajczykBD:
## Summary

Fixes for the currency/analytics epic (#2452), found and verified live this session:

- **Cancellations were currency-unsafe**: `getDailyOrderAggregates` summed `cancelledValue` from raw `totalAmount` with no currency restriction, silently mixing currencies together across days/connections. Restricted to the current-era `reportingCurrency` (mirroring `revenue`), with the unstamped remainder now surfaced separately as `cancelledUnconvertedCount`/`cancelledUnconvertedValue` instead of being folded in silently.
- **Display-currency picker offered unsupported currencies**: USD/GBP were in the FE picker but not in the backend's `SUPPORTED_REPORTING_CURRENCIES`, so selecting either returned a 400 and blanked the whole KPI strip. Trimmed the picker to PLN/EUR.
- **The GMV qualifier ignored the picker entirely**: it always rendered `headline.revenue` in the native reporting currency, even with a display currency selected — the picker only changed a banner's text. Now wired to `displayCurrencyConversion.convertedRevenue`.
- **A client-side "derive a rate" shortcut was tried and reverted**: applying `convertedRevenue / revenue` as an implied exchange rate to Net sales/AOV/median/cancelled value/Top Products produced a live wrong number (29 000 PLN read as ~20 000 "EUR" instead of ~6 700 EUR), because `convertedRevenue` in current-rate mode is `revenue` **plus** the separate unconverted-orders pool, both converted and summed — not a clean multiple of `revenue`. Documented as a rejected approach in `display-currency.lib.ts` so it isn't retried; those figures correctly stay in the native reporting currency until the backend computes a real converted value for each of them.
- **The convert-note banner was misleading during a recalculation**: it claimed "Converted to EUR" while a currency recalculation run was actively rewriting the same figures, and used one generic message for both rate-basis modes (Current rate / Rate on order date) despite them answering different questions. Both fixed.
- **Recalculation-in-progress was invisible**: the KPI strip, channel table, and Top Products table all rendered a bare `0.00` for the full duration of a currency recalculation run, with no indication a fix was already running. All three now read the existing Data Coverage `currency` category status and show "Recalculating…" instead.

## Test plan

- [x] `pnpm --filter @openlinker/core type-check` — clean
- [x] `pnpm --filter @openlinker/api type-check` — clean
- [x] `pnpm eslint` on all touched files — clean (no new errors/warnings)
- [x] 199/199 tests passing in the analytics feature, including new regression tests for every fix above:
  - currency-safe cancellation aggregation (backend + FE types)
  - GMV qualifier reacts to the display-currency picker
  - the reverted rate-derivation approach does NOT silently reintroduce (locked in by an explicit "no other figure converts" assertion)
  - "Recalculating…" renders instead of `0.00` on the KPI strip, channel table, and Top Products while a currency remediation run is `in-progress`
  - convert-note banner names the active rate-basis mode and defers to the in-progress state

## Out of scope (tracked separately, not in this PR)

- The recalculation job's own speed (currently ~2.4s/order due to `realtime`-lane throttling) — separate implementation plan, not committed to this branch.
- Extending backend currency conversion to Net sales, AOV, median, cancelled value, the channel table, and Top Products — requires new backend fields, deliberately deferred.

## Mockup status

`docs/plans/mockups/analytics-display-currency-picker.html` covers this feature but has **not**
been updated in this PR. Two new UI states were introduced that it doesn't depict:

- The convert-note banner now names the active rate-basis mode ("Current rate:" / "Rate on order
  date:") instead of one generic "Converted to EUR" message.
- The convert-note, KPI strip, channel table, and Top Products table all gained an
  "in-progress"/"Recalculating…" override that takes priority while a currency remediation run is
  active.

Per `docs/frontend-architecture.md` § UX Mockups, this is flagged rather than silently left — the
mockup update is a fast-follow, not bundled into this fix PR.


jakubretajczykBD:
## Tech review (follow-up) — commit `bf5200731`

Re-reviewed after the fix-up commit addressing the previous pass's findings.

### Findings resolved

- **Duplication** — `isCurrencyRecalculating(coverage)` (`display-currency.lib.ts`) and `<RecalculatingValue />` (`recalculating-value.tsx`) now back all four call sites (`AnalyticsKpiStrip`, `ChannelSalesTable`, `ProductSalesTable`, `AnalyticsConvertNote`). Verified no leftover inline copies remain (`grep -rn "RECALCULATING_NODE"` returns nothing).
- **Render-scope inconsistency** — the two components that were recreating the node on every render now consume the module-scoped shared component; `product-sales-table.tsx`'s `renderNovCell` signature also incidentally tightens from `boolean | undefined` to `boolean`, since `isCurrencyRecalculating` normalizes with `?? false`.
- **`docs/lessons.md`** — new entry ("Never derive a currency-conversion rate by dividing two figures that don't share a population") follows the file's own format (`Context`/`Problem`/`Rule`/`Applies to`/`Source`) and correctly cites PR #2781.
- **Mockup gap** — PR description now has a "Mockup status" section naming the two undepicted UI states and the doc rule it's answering to (`frontend-architecture.md` § UX Mockups), rather than leaving it unaddressed.

### New checks on the fix-up itself

- ✅ `type-check` clean on `web` (no regression from the extraction)
- ✅ `lint` clean — same 3 pre-existing `explicit-function-return-type` warnings, unrelated to this change, no new ones introduced
- ✅ 199/199 tests still passing across `apps/web/src/features/analytics` + `apps/web/src/pages/analytics` — the extraction is behavior-preserving
- ✅ No architecture/layering issue introduced — the new files stay within the `analytics` feature (`lib/` for the pure predicate, `components/` for the shared node), consistent with `frontend-architecture.md`'s feature-module structure
- ✅ Naming conventions followed: `recalculating-value.tsx` (kebab-case file) exports `RecalculatingValue` (PascalCase), matching the repo convention

No new issues found.

### Verdict

✅ **Approve**

PR #2783, opened 2026-09-01
fix(orders): restate FX in-page instead of one realtime job per order

### LINKED ISSUE #2776: [TASK] CORE — Restate FX in-page instead of one realtime child job per order
## Problem / Context

The Phase 5 currency remediation (#2467) fans out **one job per order**, into the wrong lane, and cannot finish in a time an operator will wait for.

`OrderFxRestatementService.restatePage` (`libs/core/src/orders/application/services/order-fx-restatement.service.ts:80-95`) clears the stamp and then calls `enqueueOne`, which enqueues one `marketplace.order.fxStamp` per order. That job type is registered in the **`realtime`** lane (`apps/worker/src/sync/handlers/handler-registration.service.ts:124-128`), whose caps are `total: 4, perScope: 2` (`apps/worker/src/sync/sync-job.runner.ts:104`).

Every child carries `connectionId = ref.sourceConnectionId` (`order-fx-restatement.service.ts:134-142`) and `resolveJobScope` **is** `connectionId` (`libs/core/src/sync/domain/types/sync-job-lane.types.ts:54`), so all N children share one scope. Once two are in flight the scope enters `excludedScopes` (`sync-job.runner.ts:357-365`), the claim returns zero rows, `startedAny` is false, and the loop sleeps the full `POLL_INTERVAL_MS = 1000` (`sync-job.runner.ts:49`, `:306`) even though `free = 4 - 2 = 2`.

Measured consequences:

- **Ceiling of ~2 orders/second per connection**, independent of how fast a stamp actually is. 10 000 orders ≈ 83 minutes before a single provider call is counted. The run's poll budget (`MAX_POLLS = 12` × `POLL_DELAY_SECONDS = 300` = 60 min, `libs/core/src/sync/domain/types/analytics-job-payloads.types.ts:72-75`) can expire before enumeration drains, closing a *successful* remediation as `failed`.
- **It starves that connection's live order ingestion specifically.** `marketplace.order.sync` is also `realtime` with the same scope (`handler-registration.service.ts:119-123`), competing for the same `perScope: 2`.
- **N Redis stream entries per click.** Children go through `JobEnqueuePort` → `RedisStreamsJobEnqueueService` (`libs/core/src/sync/sync.module.ts:63-68`), so 10 000 orders is 10 000 `jobs.sync` entries plus 10 000 `jobdedup:` keys with a 7-day TTL, against an age-bounded (`MINID`) stream. Per #2163 a trimmed entry is "un-blocked, not recovered" — the hourly sweep is the only recovery. One operator click producing that intake spike is exactly what that retention note is about.
- **N permanent `sync_jobs` rows.** That table has no retention anywhere in the tree.

The fix already exists one method away in the same class graph: `OrderFxStampService.sweep` (`libs/core/src/orders/application/services/order-fx-stamp.service.ts`) walks a page of up to `MAX_LIMIT = 500` orders **in-process**, sequentially, in the `bulk` lane (`apps/worker/src/sync/handlers/marketplace-order-fx-stamp-sweep.handler.ts:43-45`). Its own comment states the reasoning verbatim: "A page of foreign-currency orders on distinct days is a page of provider calls; fanning them out in parallel would burst a public reference-rate API for no latency benefit anyone is waiting on."

The restatement path is the only FX walker in the repo that does not do this.

## Proposed Solution

Stamp in-process inside the existing page loop. The driver (`analytics.currency.recalculate`) is **already** in the `bulk` lane (`handler-registration.service.ts:374-382`), so no lane change and no new job type are needed.

1. `OrderFxRestatementService` takes `IOrderFxStampService` (via `ORDER_FX_STAMP_SERVICE_TOKEN`) instead of `JobEnqueuePort`. Delete `enqueueOne`.
2. In `restatePage`'s existing sequential loop: `clearOne(id)` then `await this.stamp.stamp(id)`. Keep the per-order try/catch — a bad row must not abort the page, which is already the class's stated rule.
3. Report the stamp outcomes on `FxRestatementPageResult` (`stamped` / `terminal` / `deferred` counts) so the driver's log line stops claiming `enqueued: N` for work that now either happened or did not. Mirror `OrderFxSweepResult`'s shape.
4. Drop `RESTATEMENT_PAGE_SIZE` from 200 to **100** (`apps/worker/src/sync/handlers/analytics-currency-recalculate.handler.ts:73`). The page is now a wall-clock unit, not 200 cheap writes. 100 matches `DEFAULT_LIMIT` in the existing sweep handler, and the 3-minute `lockedAt` heartbeat (`sync-job.runner.ts:50`, `:482-488`) against the 15-minute stuck timeout (`apps/worker/src/maintenance/stuck-job-recovery.service.ts:23`) leaves ample headroom.
5. Delete `buildFxRestatementIdempotencyKey` and its type (`libs/core/src/orders/domain/types/order-fx-restatement.types.ts`) plus the class-header section justifying it. With no child job there is no key to collide, so the whole "`fx:{orderId}` was already spent by `enqueueRetry`" hazard stops existing rather than being worked around. Net-negative diff.

**Durability after the change.** A page dying mid-way leaves up to 100 rows cleared-but-unstamped. Both existing backstops cover it unchanged: the driver's level-triggered completion poll re-reads the population (`countRemainingCurrencyMismatch`), and the hourly `marketplace.order.fxStampSweep` predicate is `reportingCurrency IS NULL`. Retry is idempotent in outcome — re-clear then re-stamp yields the same figure. What is given up is the per-order retry ladder, which the sweep already replaces by design.

**Clear-then-stamp stops being a cross-job hazard.** The ordering rule the class doc is built around becomes a local statement pair in one process.

### Rejected alternatives

- **Split the job type so children land in `bulk`** (the #2594 pattern). Correct lane diagnosis, wrong unit of work. It raises `perScope` 2 → 8, i.e. 4× — still tick-bound, still one `sync_jobs` row and one stream entry per order, and it puts **8 concurrent stamps per connection** into NBP's up-to-8-attempt sequential walk-back (`NBP_MAX_WALK_BACK_ATTEMPTS`, `libs/integrations/fx/src/infrastructure/adapters/nbp-exchange-rate.adapter.ts:67`). That is precisely the parallel fan-out `sweep()` refuses in a comment. Expected outcome: throttling → `RateUnavailableTransientError` → `deferred` → poll budget expires → `failed`. The numbers improve and the operator-visible result gets worse.
- **One batch child job per page** (the `master.product.syncBatch` / `master.inventory.syncBatch` pattern, #2593/#2648). Genuinely the right escalation, and strictly more machinery: a new `JobTypeValues` member, a new handler, an `assertFullLaneCoverage` update, a new payload type, and a per-order fallback enqueue. It buys parallelism across pages — worth taking **only** if a measured run shows the sequential page is too slow. Deferred, not dismissed.
- **Set-based SQL for the same-currency case.** `native currency == reporting currency` needs no provider call at all (`order-fx-stamp.service.ts` step 3), and on the commonest trigger (operator switches EUR→PLN on a PLN shop) that is most of the population in one `UPDATE`. Rejected: it creates a second writer of the FX stamp against `OrderFxStampService`'s stated "the ONE place that decides, per attempt", must reproduce intent pinning / `fxRule` / the `ck_order_records_fx_group` grouping, and `nativeTotals` reads `orderSnapshot.totals` **JSONB** rather than the denormalized columns (`libs/core/src/orders/domain/entities/order-record.entity.ts:389-398`) — so a column-keyed SQL mirror is a second definition of "the native total". Not worth the invariant for a case this change already completes in about a minute.

## Classification

**Type**: CORE
**Layer**: Application
**File(s)**:
- `libs/core/src/orders/application/services/order-fx-restatement.service.ts`
- `libs/core/src/orders/application/interfaces/order-fx-restatement.service.interface.ts`
- `libs/core/src/orders/domain/types/order-fx-restatement.types.ts`
- `libs/core/src/orders/orders.module.ts`
- `apps/worker/src/sync/handlers/analytics-currency-recalculate.handler.ts`
- specs for all of the above

## Dependencies

- Independent of #2775 (the clear-guard bug fix) — that changes the predicate, this changes the unit of work. Either can merge first; both are needed for a run to actually converge.

## Assumptions

- No new `JobTypeValues` member and no `assertFullLaneCoverage` change, since the driver is already `bulk` and no child job type is introduced.
- `marketplace.order.fxStamp` stays registered in `realtime` unchanged — it remains correct for a freshly-ingested order, where one job per order is the right unit and low latency is the point.

## Acceptance Criteria

- [ ] `OrderFxRestatementService` no longer injects `JobEnqueuePort`; a spec asserts no job is enqueued during a restatement page.
- [ ] A spec asserts the per-order order of operations is clear-then-stamp, and that a `clearOne` throw does not prevent the remaining orders in the page from being stamped.
- [ ] A spec asserts a `stamp` throw on one order does not abort the page.
- [ ] `FxRestatementPageResult` reports stamp outcomes (`stamped` / `terminal` / `deferred`), and the driver's log line no longer reports `enqueued`.
- [ ] `buildFxRestatementIdempotencyKey` is deleted and no longer referenced anywhere (`git grep` clean).
- [ ] `RESTATEMENT_PAGE_SIZE` is 100, with a comment stating the page is now a wall-clock unit bounded by the 3-min heartbeat / 15-min stuck timeout.
- [ ] An integration test proves a multi-page run over foreign-currency orders reaches `resolved` and creates **zero** `marketplace.order.fxStamp` rows in `sync_jobs`.
- [ ] An integration test proves a run no longer occupies `realtime` lane slots (a queued `marketplace.order.sync` job for the same connection is not delayed).

## Branching / merge strategy

```
git checkout -b <branch-name>
```

jakubretajczykBD:
## Summary
- `OrderFxRestatementService.restatePage` clears each order's ADR-040 FX stamp and re-stamps it via `IOrderFxStampService.stamp` **in-process, sequentially** — no `marketplace.order.fxStamp` child job is enqueued any more, so a restatement no longer fans out into the `realtime` lane (whose `perScope: 2` cap it shared with live order ingestion), and produces zero extra `sync_jobs` rows / Redis stream entries.
- `FxRestatementPageResult` now reports `stamped`/`terminal`/`deferred` (mirroring `OrderFxSweepResult`) instead of an `enqueued` count.
- `RESTATEMENT_PAGE_SIZE` dropped 200 → 100 (now a wall-clock unit, not 200 cheap enqueues).
- `buildFxRestatementIdempotencyKey` deleted (nothing left to key).
- Bonus fix (found during manual testing of this branch): `/analytics`'s date-range Apply/preset handlers were replacing the whole query string, silently dropping `displayCurrency`/`rateBasis` on every date change — ADR-064 states they're URL-encoded "like the existing date-range filter". Fixed to merge onto existing params instead.

## Test plan
- [x] Unit specs updated: `order-fx-restatement.service.spec.ts`, `analytics-currency-recalculate.handler.spec.ts`, `analytics-page.test.tsx` (new regression test)
- [x] Integration spec updated: `currency-remediation.int-spec.ts`, including an assertion that a full run creates **zero** `marketplace.order.fxStamp` rows in `sync_jobs`
- [x] `tsc --noEmit` clean for `libs/core` and `apps/worker`
- [ ] Full `pnpm lint && pnpm type-check && pnpm test` — not run locally (sandboxed session), please confirm on CI

Closes #2776

### DIFF
--- a/apps/api/test/integration/analytics/currency-remediation.int-spec.ts
+++ b/apps/api/test/integration/analytics/currency-remediation.int-spec.ts
@@ -37,6 +37,7 @@ import {
   type IOrderFxStampService,
 } from '@openlinker/core/orders';
 import { OrderRecordOrmEntity } from '@openlinker/core/orders/orm-entities';
+import { SyncJobOrmEntity } from '@openlinker/core/sync/orm-entities';
 import {
   getTestHarness,
   resetTestHarness,
@@ -164,7 +165,7 @@ describe('Currency remediation against real Postgres (#2468)', () => {
     expect(row.reportingCurrency).toBe(STALE_REPORTING_CURRENCY);
   });
 
-  it('clears all six FX columns and leaves the row on the sweep frontier', async () => {
+  it('clears all six FX columns and then re-stamps in-process, in the same page (#2776)', async () => {
     const orderId = await seedStaleStampedOrder(100);
     const run = await runs.openRun({
       category: 'currency',
@@ -178,45 +179,20 @@ describe('Currency remediation against real Postgres (#2468)', () => {
       limit: 50,
     });
 
-    expect(page).toMatchObject({ scanned: 1, cleared: 1 });
+    expect(page).toMatchObject({ scanned: 1, cleared: 1, stamped: 1, terminal: 0, deferred: 0 });
     const row = await readOrder(orderId);
-    expect(row.reportingCurrency).toBeNull();
-    expect(row.reportingTotalAmount).toBeNull();
-    expect(row.exchangeRateId).toBeNull();
-    expect(row.fxStampedAt).toBeNull();
-    // The subtle one: left behind, `resolveIntent` re-pins USD and the repair
-    // silently produces the same wrong figure.
-    expect(row.fxIntendedCurrency).toBeNull();
-    expect(row.fxRule).toBeNull();
-  });
-
-  it('re-stamps into the CURRENT reporting currency after the clear, not the stale one', async () => {
-    const orderId = await seedStaleStampedOrder(100);
-    const run = await runs.openRun({
-      category: 'currency',
-      affectedCount: 1,
-      triggeredByUserId: 'user-1',
-    });
-    await restatement.restatePage(scope, reportingCurrency, {
-      runId: run.id,
-      afterOrderId: null,
-      limit: 50,
-    });
-
-    const outcome = await fxStamp.stamp(orderId);
-
-    expect(outcome).toMatchObject({
-      kind: 'stamped',
-      alreadyStamped: false,
-      reportingCurrency,
-      reportingTotalAmount: 100,
-    });
-    const row = await readOrder(orderId);
+    // Re-stamped into the CURRENT reporting currency, not the stale one — the
+    // clear alone would leave every column null; the page's own in-process
+    // stamp is what produces a fresh, correct figure.
     expect(row.reportingCurrency).toBe(reportingCurrency);
     expect(Number(row.reportingTotalAmount)).toBe(100);
+    // The subtle one: left behind, `resolveIntent` would have re-pinned USD
+    // and the repair would silently reproduce the same wrong figure.
+    expect(row.fxIntendedCurrency).toBe(reportingCurrency);
+    expect(row.fxRule).not.toBeNull();
   });
 
-  it('reaches resolved once the whole scope is re-stamped, with the ledger row as the only state', async () => {
+  it('reaches resolved once the whole scope is re-stamped by the page itself, with no child job anywhere', async () => {
     const orderIds = [
       await seedStaleStampedOrder(100),
       await seedStaleStampedOrder(250),
@@ -229,8 +205,8 @@ describe('Currency remediation against real Postgres (#2468)', () => {
     });
 
     // Enumerate in two pages so the keyset cursor is genuinely exercised: a
-    // cleared row still satisfies the mismatch predicate, so an offset walk
-    // would re-read page one forever.
+    // cleared-and-restamped row no longer satisfies the mismatch predicate,
+    // so this also proves the cursor advances over rows this run itself fixed.
     const first = await restatement.restatePage(scope, reportingCurrency, {
       runId: run.id,
       afterOrderId: null,
@@ -244,18 +220,9 @@ describe('Currency remediation against real Postgres (#2468)', () => {
     });
     expect(second.nextCursor).toBeNull();
     expect(first.scanned + second.scanned).toBe(3);
+    expect(first.stamped + second.stamped).toBe(3);
 
-    // Mid-run the population is still fully mismatched (cleared, not stamped).
-    await expect(restatement.countRemaining(scope, reportingCurrency)).resolves.toMatchObject({
-      total: 3,
-      terminalMarked: 0,
-      pending: 3,
-    });
-
-    for (const orderId of orderIds) {
-      await fxStamp.stamp(orderId);
-    }
-
+    // The page itself stamped every order — nothing left to converge on.
     await expect(restatement.countRemaining(scope, reportingCurrency)).resolves.toMatchObject({
       total: 0,
     });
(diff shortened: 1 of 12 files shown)


jakubretajczykBD:
## Tech review (re-review after fix-up commit `86d491883`)

### Summary

Both findings from the prior review pass are correctly and completely fixed. `stampOne()` now guards `stamp()` exactly the way `clearOne()` guards the clear, with a new `failed` outcome distinct from `deferred`, and the exact test the original issue's AC asked for is now present. `FxRestatementOrderRef` and its dead `sourceConnectionId` selection are gone — the port now returns `string[]`, matching what the caller actually needs. No new issues introduced by the fix-up. `tsc --noEmit` is clean on both `libs/core` and `apps/worker`.

### Issues

**[SUGGESTION]** — `libs/core/src/orders/application/services/order-fx-restatement.service.ts` (`stampOne`'s switch in `restatePage`)
> `switch (outcome?.kind) { case 'stamped': ... case undefined: failed += 1; break; }` has no `default` and no exhaustiveness assertion (`never`-typed fallthrough check). If `FxStampOutcome` ever grows a fourth member, this switch would silently count it as neither `stamped`/`terminal`/`deferred`/`failed` — not a regression from this PR (the pre-existing switch had the identical shape and the same gap), so not blocking, but worth an exhaustiveness helper the next time someone touches this file.

**[SUGGESTION]** — `libs/core/src/orders/application/services/order-fx-restatement.service.ts:69-70` (import order)
> `FxStampOutcome` is imported after `IOrderFxStampService` rather than grouped with the other `../../domain/types/*` import two lines above. Purely cosmetic, no documented rule violated.

Nothing else. Both prior findings verified fixed line-by-line against the diff:
- ✅ `stampOne` wraps `this.stampService.stamp(...)` in try/catch, logs, returns `undefined` on throw — mirrors `clearOne` exactly.
- ✅ New `failed: number` on `FxRestatementPageResult`, tallied via `case undefined: failed += 1;`, well-documented on both the type and the service.
- ✅ New unit test `'should keep going, and still count the rest, when one order's stamp attempt throws'` exercises the exact throw-doesn't-abort-the-page behavior the issue's AC required, asserting both the tally (`stamped: 1, failed: 1`) and that `stamp()` is still called for the order after the one that threw.
- ✅ `FxRestatementOrderRef` deleted from the type file, the port, the repository, and the barrel — `git grep` clean.
- ✅ `findCurrencyMismatchOrderRefsAfter` now selects and returns only `internal_order_id`; the dead `sourceConnectionId` column and its `addSelect` are gone.
- ✅ All spec files (unit + int-spec + worker handler spec) updated consistently — `seed()` now returns plain strings, mock fixtures carry `failed: 0`, no stale `FxRestatementOrderRef`-shaped objects remain anywhere.

### Verdict

✅ **Approve** — ready to merge. The two SUGGESTIONs are optional polish, not blockers.

PR #2485, opened 2026-08-25
docs(analytics): ADR-064 display-currency conversion + repo-committed UX mockup

norbert-kulus-blockydevs:
## Summary
- Adds ADR-064, recording the decision that `/analytics` display-currency conversion is a read-only, aggregate-level transform (two synchronous modes — "current rate" and "order-date" — neither needing a job or progress UI), and states up front why this does not conflict with ADR-040's read-time-conversion boundary.
- Formalizes in `docs/frontend-architecture.md` that interactive UX mockups built for a feature must be committed under `docs/plans/mockups/`, not left only as an ephemeral Claude Artifact link.
- Commits `docs/plans/mockups/analytics-display-currency-picker.html` — the mockup this epic's design (currency picker, Data Coverage panel, remediation flows) was built against. The original, live-edited version is at https://claude.ai/code/artifact/c870ccff-91f7-4798-999a-4b6803647625 — kept here only as the design-session reference; the repo copy is now the durable spec that implementation and E2E tasks in this epic reference by path and `data-state`.
- Updates the ADR index (`README.md`) with ADR-064's row and reserves ADR-065 for the next new ADR.

## Context
Part of epic #2452 (Analytics — display-currency picker + data-coverage remediation), Phase 1 (decisions), Task 1.1.

## Test plan
- [x] Docs-only change — no code, no migration, no runtime behavior.
- [ ] Reviewer confirms the ADR-040 tension/resolution framing reads clearly to someone who hasn't seen the mockup.

Closes #2454

PR #2785, opened 2026-09-01
feat(currency): expose the exchange rate behind every converted figure

### LINKED ISSUE #2778: [TASK] CORE/Interface — Expose the exchange rate behind every converted figure

## Problem / Context

OpenLinker persists every exchange rate it uses, specifically so a financial figure can be verified — and then exposes none of them over HTTP. ADR-040's stated purpose for the `exchange_rates` registry is auditability ("a rate editable after the fact would make every figure derived from it unverifiable"), and the rows carry exactly what an audit needs: `rate`, `rateDate`, `derivation` (`direct | inverted | pivot`), and `sourceRef` (NBP's own table number, e.g. `149/A/NBP/2026`).

None of it is reachable:

1. **No endpoint reads the registry.** `git grep` over `apps/api/src` finds no controller, service or DTO for `exchange_rates` — the only API-side references are the two migrations. There is no `GET` for a rate, by key or otherwise.
2. **The conversion result carries no rate.** `DisplayCurrencyConversionDto` (`apps/api/src/analytics/http/dto/sales-analytics-response.dto.ts:43-95`) has exactly four fields: `displayCurrency`, `rateBasis`, `convertedRevenue`, `unresolvedNativeCurrencies`. The domain types behind it are the same — `CurrentRateConversionResult` and `OrderDateConversionResult` (`libs/core/src/orders/domain/types/display-currency.types.ts:105-144`) carry no rate, no rate date and no source. `NativeCurrencyBreakdown` reports `nativeTotal` and `convertedTotal` but not what multiplied one into the other.
3. **`order_records.exchangeRateId` is a dead pointer over HTTP.** No orders DTO exposes it, and even if it did, the row it names is unreadable.

Operator-visible consequence: `/analytics` renders "12 345 PLN" converted from EUR, and there is no way — in the UI or by curl — to find out which rate produced it, from which day, from which published table. The audit trail exists and is write-only. For the epic's display-currency picker that is the difference between a number an operator can defend to an accountant and a number they cannot.

`ADR-040`'s own warning makes this sharper: the analytics stamp is explicitly **not** a statutory rate and must never be confused with an invoice's. Without the rate, its date and its source visible beside the figure, the two are indistinguishable on screen.

Also fix in passing: `apps/api/src/analytics/http/dto/sales-analytics-query.dto.ts:9` still says `displayCurrency` / `rateBasis` are "pending in PR #2485". They shipped. The comment misleads anyone auditing the surface (it did during this investigation).

## Proposed Solution

Two additions. The second matters more than the first.

**A. Carry the rate through the conversion result** (the primary fix). Extend the domain result types with a neutral provenance shape, and thread it into the DTO:

```ts
/** What produced one converted figure. Never a statutory rate — ADR-040. */
export interface AppliedRate {
  readonly from: string;
  readonly to: string;
  readonly rate: string;        // string end-to-end vs numeric(18,8) — never Number()'d
  readonly rateDate: string;    // the day the SOURCE published for
  readonly source: string;      // 'nbp' | 'ecb' | …
  readonly derivation: 'direct' | 'inverted' | 'pivot';
  readonly sourceRef: string | null;  // NBP table number; null for ECB
}
```

- `NativeCurrencyBreakdown` gains `appliedRate: AppliedRate | null` (`null` exactly when `convertedTotal` is `null`, i.e. the row is already in `unresolvedNativeCurrencies`).
- `OrderDateConversionResult` gains `appliedRate: AppliedRate | null`, on the same rule.
- `DisplayCurrencyConversionDto` exposes them. `order-date` normalises to a 0-or-1-element view exactly as the existing `unresolvedNativeCurrencies` field already does, so a caller reads one shape regardless of `rateBasis`.

`rate` stays a **string** end to end, matching `ExchangeRate.rate` against the `numeric(18,8)` column — `Number()`-ing it would reintroduce binary-float error into the one figure whose purpose is to be auditable.

**B. A read endpoint for the registry.** `GET /currency/rates?from=&to=&date=` on a new `apps/api/src/currency/http/exchange-rates.controller.ts`, beside the existing `currency-settings.controller.ts` in the same module. Returns the stored row (rate, published date, source, derivation, `sourceRef`) or 404.

Two properties are load-bearing:
- **It is a READ of the registry, never a fetch.** It must not call a provider, so it consumes `ExchangeRateRepositoryPort.findByKey` (via a thin core service interface, per the cross-context rule) and **not** `ICurrencyRateService.getRateFor`, whose get-or-create semantics would let an HTTP caller mint registry rows and spend the deployment's provider budget.
- **Not `@Roles('admin')`.** Unlike `currency-settings` (which writes a deployment-wide setting), this is the evidence behind a figure any operator can already see on `/analytics`. Gated by the global `JwtAuthGuard` only.

Deliberately out of scope: exposing `order_records.exchangeRateId` on the orders DTOs. That is a per-order surface with its own read model and its own FE consumer; file it separately if wanted.

## Classification

**Type**: CORE + Interface
**Layer**: Domain (types) / Application (conversion service, a new read interface) / Interface (HTTP)
**File(s)**:
- `libs/core/src/orders/domain/types/display-currency.types.ts` (`NativeCurrencyBreakdown`, `OrderDateConversionResult`, the new `AppliedRate`)
- `libs/core/src/orders/application/services/display-currency-conversion.service.ts`
- `libs/core/src/currency/application/interfaces/` (a new read-only rate-lookup interface + token in `currency.tokens.ts`)
- `apps/api/src/currency/http/exchange-rates.controller.ts` (new), `apps/api/src/currency/http/dto/exchange-rate-response.dto.ts` (new), `apps/api/src/currency/currency.module.ts`
- `apps/api/src/analytics/http/dto/sales-analytics-response.dto.ts` (`DisplayCurrencyConversionDto`)
- `apps/api/src/analytics/http/dto/sales-analytics-query.dto.ts:9` (stale "pending in PR #2485" comment)

## Dependencies

- None. Independent of #2775, #2776 and #2777.
- Blocks #2779 (rate provenance on `/analytics`) — that one needs this DTO shape, though it can start against the agreed shape per epic #2452's Phase 6/7/8 rule.

## Assumptions

- `sourceRef` is nullable because ECB assigns none (its `header.id` is a fresh UUID per request); the ECB adapter records an OpenLinker-constructed locator instead, per ADR-040.
- Additive only: omitting `displayCurrency` leaves `GET /analytics/sales` byte-identical, and `appliedRate` is a new optional field on an existing DTO, so no existing FE consumer breaks.
- One `AppliedRate` per native-currency breakdown row is enough granularity; a `pivot` derivation reports the composed result plus its `derivation` flag rather than both legs.

## Acceptance Criteria

- [ ] `AppliedRate` is declared once and reused by both result types; `rate` is typed `string`, with a comment stating why it is never `Number()`'d.
- [ ] A spec proves `appliedRate` is non-null for every converted `NativeCurrencyBreakdown` row and `null` exactly for the rows listed in `unresolvedNativeCurrencies`.
- [ ] A spec proves `order-date` mode reports the single applied rate, and reports `null` both when nothing was stamped in range and when the lookup failed.
- [ ] `GET /currency/rates?from=EUR&to=PLN&date=2026-08-29` returns rate, published `rateDate`, source, `derivation` and `sourceRef`, and **404** when no row exists.
- [ ] A controller spec proves the endpoint never calls `ICurrencyRateService.getRateFor` and therefore cannot cause a provider call or a registry insert.
- [ ] The endpoint is reachable by a non-admin authenticated user; a spec asserts it carries no `@Roles('admin')`.
- [ ] `GET /analytics/sales` without `displayCurrency` returns a byte-identical response shape to before (regression spec).
- [ ] The stale "pending in PR #2485" comment is removed.
- [ ] No architecture boundary violations (CORE ↔ Integration) — in particular the controller reaches the registry through a core service interface, not `ExchangeRateRepositoryPort` directly (§ Cross-context dependencies in core).

## Branching / merge strategy

```
git checkout -b <branch-name>
```

jakubretajczykBD:
## Summary
OpenLinker persists every exchange rate it uses (`exchange_rates` registry, ADR-040) specifically for auditability, but exposed none of it over HTTP. Two additions:

**A — Rate provenance on the display-currency conversion result.** New `AppliedRate {from, to, rate, rateDate, source, derivation, sourceRef}` threaded through `NativeCurrencyBreakdown` and `OrderDateConversionResult` — `null` exactly when no lookup happened (unresolved row, or a same-currency identity that never called the provider). `DisplayCurrencyConversionDto` exposes them as `appliedRates[]` (0..N for `current-rate`, 0-or-1 for `order-date`, mirroring the existing `unresolvedNativeCurrencies` normalization). `rate` stays a `string` end to end, matching the `numeric(18,8)` registry column — never `Number()`'d.

**B — `GET /currency/rates?from=&to=&date=`** — a pure registry read. Backed by a new `IExchangeRateLookupService`, a thin `findByKey` passthrough deliberately separate from `ICurrencyRateService` so the endpoint structurally cannot reach the get-or-create `getRateFor` and spend the deployment's provider budget. No `@Roles(...)` — this is the evidence behind a figure any operator can already see on `/analytics`, gated by the global `JwtAuthGuard` only.

**Also fixed in passing:**
- A test gap left over from #2776: `order-record.repository.spec.ts` still asserted the old `{internalOrderId, sourceConnectionId}` shape after that PR moved `findCurrencyMismatchOrderRefsAfter` to bare ids.
- All 9 occurrences of the stale `"pending in PR #2485"` comment across the display-currency read model (the issue named one; the rest were the same drift).

## Test plan
- [x] `AppliedRate.rate` typed `string`, never `Number()`'d — verified in both the domain type and the DTO
- [x] Unit spec proves `appliedRate` non-null iff resolved, null for unresolved/identity rows (`display-currency-conversion.service.spec.ts`)
- [x] Unit spec proves `order-date` reports the single applied rate, null for no-stamp and for lookup-failure
- [x] Controller spec proves no `@Roles(...)` on the endpoint and that its only dependency is the lookup service (never `ICurrencyRateService`)
- [x] New `apps/api/test/integration/currency/exchange-rates.int-spec.ts` against real Postgres: 200 for a seeded row, 404 for an absent key, 422 for an unsupported `to`, 401 with no token — **not run locally** (no Docker/built `@openlinker/test-kit` in this sandbox), written and lint-clean but unexecuted; please confirm on CI
- [x] `pnpm test` full suite: `libs/core` 281/281 suites, 3704/3704 tests green; `apps/api` 1357/1357 tests green (8 suites fail to *load* on this branch and on `main` alike — pre-existing unbuilt `@openlinker/plugin-sdk`/integration-package dist issue, unrelated to this change, verified by suite names: allegro/connection/oauth/content controllers)
- [x] `tsc --noEmit` clean on `libs/core` and `apps/api`
- [x] `check-cross-context-imports.mjs`, `check-service-interfaces.mjs`, `check-workspace-dep-declarations.mjs` all green — the controller reaches the registry through `IExchangeRateLookupService`, never `ExchangeRateRepositoryPort` directly

Closes #2778


piotrswierzy (commented):
Reviewed the delta against `epic/2452-analytics-currency-coverage` (24 files).

**Verified against ADR-040 — all clean:**
- **Rate direction** is a multiply, never a divide: `resolveRate(currency, input.displayCurrency, rateDate)` keys the lookup `from = native`, `to = display`, and both modes do `total * Number(stored.rate)`. `AppliedRate.from/to` are copied from the same `StoredExchangeRate` row, so what is reported and what was applied cannot describe different lookups (`display-currency-conversion.service.ts`, `toAppliedRate`).
- **`rate` stays a `string`** end to end — domain type, `AppliedRateDto`, `ExchangeRateResponseDto`. `Number()` appears only at the local multiply site, never on a persisted/emitted value.
- **No unstamped-predicate error**: nothing here reads `exchangeRateId`; `appliedRate: null` is derived from the lookup result, not from a stamp column.
- **The endpoint is a read, not a fetch**: `IExchangeRateLookupService` is a one-method `findByKey` passthrough, `ICurrencyRateService` is not constructor-injectable, and the int-spec deliberately installs no network stub.
- `resolveRateSource` is applied to `to`, matching every existing call site; `ReportingCurrencyUnsupportedError → 422` mirrors `CurrencySettingsController`'s own boundary split. No `*RepositoryPort` crosses into `apps/api`.

### IMPORTANT

1. `apps/api/src/currency/http/exchange-rates.controller.ts` (+ `dto/exchange-rate-response.dto.ts`, `sales-analytics-response.dto.ts`) — the ADR-040 "never a statutory rate" warning lives only in file/class docblocks, which Swagger does not emit. The generated OpenAPI page will show a rate with an NBP table number in `sourceRef` and nothing saying it is analytics-only. ADR-040 states that warning positively precisely because the nearest persisted rate is what the next person implementing FA(3) `KursWaluty` will reach for — and this PR is what makes it reachable over HTTP. Fix: put the caveat in the emitted text — `@ApiOperation({ description: '… Analytics provenance only; never a statutory/fiscal conversion rate (ADR-040).' })` and the same sentence on `AppliedRateDto`'s / `ExchangeRateResponseDto`'s `rate` `@ApiProperty`.

2. `docs/architecture-overview.md § 18 Currency` — unchanged. That section enumerates the context's seams and surfaces exhaustively (down to `sourceRef`'s NBP-vs-ECB asymmetry), and this PR adds a new HTTP surface, a new core service interface + token, and a new field on the analytics response. Add a bullet for `GET /currency/rates` / `IExchangeRateLookupService` (stating the read-not-fetch property) and for `appliedRates`, per the doc convention the rest of the epic follows.

### SUGGESTION

3. `get-exchange-rate.dto.ts:29` — `@Matches(/^\d{4}-\d{2}-\d{2}$/)` accepts `2026-99-99`, which answers 404 ("no rate stored") for a date that cannot exist. A real-calendar check would let that be a 400 and keep 404 meaning absence.

4. `get-exchange-rate.dto.ts` — `from`/`to` are not case-normalised, so `?to=pln` answers 422 ("no publisher serves pln") and `?from=eur&to=PLN` answers 404 with a row present. Consistent with `SetReportingCurrencyDto`, so not a regression — but a `@Transform(v => v.toUpperCase())` here is cheap.

5. `exchange-rates.int-spec.ts` is written but unrun (stated in the description). Please confirm green on CI before merge; checks are still pending.

piotrswierzy (approved):
## Re-review at `7670637` (15 files vs `epic/2452-analytics-currency-coverage`)

Previous round's SUGGESTION items are addressed and verified in the diff:

- `get-exchange-rate.dto.ts` now carries `@Transform(toUpperCase)` on `from`/`to`, so `?to=pln` no longer 422s against a row that exists. Global `ValidationPipe` runs with `transform: true` (`apps/api/src/main.ts:73-77`), so the transform actually fires — checked, not assumed.
- `date` is `@IsDateString({ strict: true })` with a calendar-shaped message, so `2026-99-99` is a 400 and 404 keeps meaning "absent", as asked.
- The ADR-040 caveat is now in *emitted* Swagger text — `@ApiOperation({ description: 'Analytics provenance only; never a statutory/fiscal conversion rate (ADR-040).' })` plus the same sentence on `ExchangeRateResponseDto.rate`'s `@ApiProperty`. That is the fix that mattered: this PR is what makes the nearest persisted rate reachable over HTTP, and ADR-040 states the FA(3) `KursWaluty` warning positively for exactly that reason.

### ADR-040 conformance (re-verified)

- **Direction invariant.** The controller keys the lookup `{ source: resolveRateSource(to), from, to, rateDate: date }` and returns the row verbatim — no arithmetic, no inversion, so there is no place for the `to`-units-per-one-`from`-unit rule to be violated.
- **`rate` stays a `string`** in the domain type, the DTO and the response — `numeric(18,8)` never round-trips through `Number()`.
- **Append-only registry preserved.** `IExchangeRateLookupService` declares exactly one method (`findRate`) and `ExchangeRateLookupService` is a one-line delegation to `findByKey`. `insertIfAbsent` and the get-or-create `ICurrencyRateService.getRateFor` are not constructor-injectable into the controller — the structural argument, not a comment.
- **Unstamped predicate.** Nothing here reads `exchangeRateId`; no `reportingCurrency IS NULL` inversion introduced.
- `resolveRateSource` is applied to `to`, matching `OrderFxStampService` and `DisplayCurrencyConversionService`; `ReportingCurrencyUnsupportedError → 422` mirrors `CurrencySettingsController`'s boundary split.
- No `*RepositoryPort` crosses into `apps/api` — the int-spec restates `ExchangeRateRegistryWriter` locally rather than importing the port, consistent with `exchange-rate-registry.int-spec.ts` and with `check-cross-context-imports.mjs`.
- `exchange_rates` is already in `tablesToTruncate` (`apps/api/test/integration/setup.ts:186`), so the new int-spec's `resetTestHarness()` genuinely isolates the seeded row.

Naming, layering, file headers, token convention (`EXCHANGE_RATE_LOOKUP_SERVICE_TOKEN` in `currency.tokens.ts`, `export *` sub-barrel, type-only export from the context barrel) all match `engineering-standards.md`.

### IMPORTANT (carried over, still open)

1. **`docs/architecture-overview.md § 18 Currency` is untouched.** That section enumerates the context's seams and surfaces exhaustively — down to the NBP-vs-ECB `sourceRef` asymmetry — and this PR adds a new HTTP surface, a new core service interface and a new DI token. Add a bullet for `GET /currency/rates` / `IExchangeRateLookupService` stating the read-not-fetch property (the one design fact a future maintainer must not undo by "simplifying" it onto `ICurrencyRateService`). Approving on the understanding this lands before the epic merges to `main`.

### SUGGESTION

2. `exchange-rates.controller.ts` — the 422 body is `ReportingCurrencyUnsupportedError`'s message, which talks about *reporting* currencies. On this endpoint the operator asked for a rate, not a setting; a one-line rewrap ("no registered publisher quotes `{to}`; publishers serve: …") would read correctly for the actual request.

3. `orders.module.ts` / `orders.tokens.ts` — the `SALES_DOCUMENT_VIEW_SERVICE_TOKEN` block is only reordered. Harmless, but it is pure diff noise on a PR that otherwise reads cleanly; worth dropping if a rebase makes it easy.

4. `exchange-rates.int-spec.ts` is written but unrun locally (stated in the description). Please confirm it is green on CI before merge — the "no network stub installed, so a `getRateFor` regression would hang rather than pass" argument is only load-bearing once the suite has actually executed.

### Assessment

Small, well-scoped, and the two properties the design leans on (read-never-fetch, no `@Roles`) are asserted structurally rather than described. No BLOCKING issues.

🔄 **Approve with changes** — merge after adding the `§ 18 Currency` doc bullet and confirming the int-spec on CI.

norbert-kulus-blockydevs:
The branch was rebased onto the current `epic/2452-analytics-currency-coverage` tip, so the PR diff went from 199 files back down to 24. The file counts from the re-review are valid again.

Going through your numbering:

1. **Doc bullet in `docs/architecture-overview.md` § 18 Currency** - no new change needed. The round-1 fix commit did add the "The registry is readable over HTTP (#2778)" bullet, but it was lost in a bad merge before your re-review. The rebase restored the bullet in place, immediately before "### 19. Catalog Trust".

2. **The 422 body wording** - fixed. `apps/api/src/currency/http/exchange-rates.controller.ts` was throwing `ReportingCurrencyUnsupportedError.message` verbatim, which reads "Reporting currency 'X' is not supported. Supported: [...]" - correct on the `/currency-settings` write path, wrong here since this endpoint asks about a rate lookup, not a deployment-wide setting. The controller now rewraps the exception's own `submitted` and `supportedCurrencies` fields into a rate-specific sentence: "No registered publisher quotes 'X'; publishers serve: [EUR, PLN]". The shared `ReportingCurrencyUnsupportedError` message is untouched. The `@ApiResponse({ status: 422 })` Swagger description was updated to match, and a unit test in `exchange-rates.controller.spec.ts` asserts the 422 body names the rate publisher and never mentions "Reporting currency" or "is not supported".

3. **`orders.module.ts` / `orders.tokens.ts` reordering noise** - no new change needed. `orders.tokens.ts` is no longer in the diff at all after the rebase, and `orders.module.ts` now carries only a one-line comment change (removing a stale "pending in PR #2485" annotation).

4. **Integration spec CI status** - a fresh CI run started on the new head commit and the `Integration Tests` job (which is what exercises `apps/api/test/integration/currency/exchange-rates.int-spec.ts`) is still running as of this comment. Will follow up once it settles if it comes back red.

PR #2788, opened 2026-09-01
feat(analytics): convert every currency figure using the real per-bucket FX rate

### LINKED ISSUE #2779: [TASK] Frontend — Show which rate produced a converted analytics figure

## Problem / Context

`/analytics` shows converted money and never says what converted it. With the display-currency picker (#2459, ADR-064) an operator can render the dashboard in PLN while the orders were stamped in EUR — the KPI reads "12 345 PLN" with no rate, no rate date and no source anywhere on screen.

The backend has the evidence and, today, does not ship it: `DisplayCurrencyConversionDto` (`apps/api/src/analytics/http/dto/sales-analytics-response.dto.ts:43-95`) carries only `displayCurrency`, `rateBasis`, `convertedRevenue`, `unresolvedNativeCurrencies`. #2778 adds `appliedRate` (rate, published date, source, `derivation`, `sourceRef`); this issue renders it.

Why it matters beyond tidiness. ADR-040 draws a hard line: the analytics FX stamp is **analytics-only** and must never be read as a statutory/invoice rate — the two differ in date anchor, target currency and permitted derivation. On a PLN-reporting deployment both come from NBP table A and differ *only* by date, so on screen they are indistinguishable unless the UI states which one it is showing. A converted figure with no visible provenance invites exactly that confusion.

Second concrete gap: `rateBasis: 'order-date'` applies **one** of today's rates to a total mixing orders from many dates (ADR-064 § Consequences calls this out — "isn't historically accurate per order… Accepted, not a defect"). The UI currently gives an operator no way to know that is what they are looking at.

## Proposed Solution

Render provenance where the converted figure is, at two levels of disclosure — per the house rule that a surface shows one or two key facts and hides the rest behind disclosure, never one dense strip.

1. **Inline, one line, under the converted KPI**: `≈ 12 345,00 PLN · 1 EUR = 4,2368 PLN (NBP, 29.08.2026)`. Muted, secondary type. This is the fact an operator needs 95% of the time.
2. **Behind disclosure** (the existing popover/tooltip pattern used by the Revenue card): the full `AppliedRate` — `derivation` (`direct` / `inverted` / `pivot`), `sourceRef` (NBP table number, quotable in a support ticket), and the per-native-currency breakdown with each row's own rate for `current-rate` mode.
3. **Mode honesty for `order-date`**: state that one current rate was applied to the whole period's total, so it answers "what is this worth today", not "what each order converted to on its own date". One short sentence in the disclosure, not a warning banner — it is a documented mode, not a fault.
4. **A statement that this is not an invoice rate**, in the disclosure only. One sentence, factual, no legal phrasing.
5. **The `unavailable` state keeps its existing treatment** (mockup state `unavailable`, `AnalyticsDegradationBanner`): a native currency with no resolvable rate already renders explicitly rather than being silently dropped. This issue must not weaken that — a row with `appliedRate: null` shows the existing unavailable copy, never a blank rate line.

Scope: headline KPI cards and the channel table's converted column on `/analytics`. Per-order provenance (the order detail drawer reading `order_records.exchangeRateId`) is **not** in scope — that needs its own backend read and is deliberately excluded from the backend sibling too.

The frontend holds no currency knowledge: every value including the source label comes from the response, and no rate, date format assumption or source name is hardcoded — the same rule `description-format` (#2193) established for adapter-declared capabilities.

## Classification

**Type**: Frontend
**Layer**: features/analytics
**File(s)**:
- `apps/web/src/features/analytics/**` (KPI cards, channel table, the display-currency picker's own surface)
- the analytics API client + query hooks that type `DisplayCurrencyConversionDto`
- `docs/plans/mockups/analytics-display-currency-picker.html` — update the `converted` and `unavailable` states to show the provenance line, per `docs/frontend-architecture.md § UX Mockups`

## Dependencies

- **Blocked on** #2778 (`appliedRate` on `DisplayCurrencyConversionDto`). Per epic #2452's Phase 6/7/8 rule, FE work may start against the agreed DTO shape with a mocked client before that PR merges.

## Assumptions

- Number and date formatting reuse whatever the analytics surfaces already use; this issue introduces no new formatting utility.
- `sourceRef` is `null` for ECB, so the disclosure omits that row rather than rendering an empty label.
- No new component is created if the existing Revenue-card disclosure pattern fits — reuse over invention, per `CLAUDE.md`.

## Acceptance Criteria

- [ ] A converted KPI renders one inline provenance line showing the rate, both currencies and the published date.
- [ ] The disclosure shows `derivation` and `sourceRef`, and for `current-rate` mode one row per native currency with that row's own rate.
- [ ] `rateBasis: 'order-date'` states in its disclosure that one current rate was applied to the whole period's total.
- [ ] The disclosure states the figure is not an invoice/statutory rate.
- [ ] A row with `appliedRate: null` renders the existing `unavailable` copy and **no** rate line; a test proves no blank or `null` rate can render.
- [ ] With `displayCurrency` omitted, the page renders exactly as before — no provenance line, no layout shift (regression test).
- [ ] No rate, currency name, source label or date format is hardcoded in `apps/web`; all come from the response.
- [ ] The mockup's `converted` and `unavailable` states match the implemented UI.

## Branching / merge strategy

```
git checkout -b <branch-name>
```

jakubretajczykBD:
## Summary
- Extends #2778's real per-bucket exchange rate (`AppliedRate`) from GMV-only to every currency-denominated figure on the analytics dashboard: `resolveReportingCurrencyRate` finds the ONE `AppliedRate` whose `from` matches the reporting currency and reuses it — never a rate derived by dividing two unrelated totals (the documented production bug: 29 000 PLN misread as ~20 000 "EUR").
- `AnalyticsKpiStrip`: Net sales, AOV, Median, and Cancelled value now convert alongside GMV when a display currency is picked.
- `ChannelSalesTable`: per-channel Net sales/AOV and the `Total · {currency}` row convert using the same headline rate (every channel's `currency` equals `headline.currency` per ADR-040's one-system-wide-reporting-currency invariant).
- `ProductSalesTable`: Net sales converts too — `GET /analytics/top-products` has no rate of its own, but it shares the same reporting currency, so it reuses the sales-analytics headline's rate via the shared TanStack Query cache entry (no extra request).
- Retains the ADR-064 refusal-over-guess discipline: a figure with no resolvable rate for its native currency stays in the native reporting currency, never a guessed number.

## Test plan
- [x] `resolveReportingCurrencyRate` unit tests (order-date/current-rate modes, no-match, no-conversion)
- [x] `AnalyticsKpiStrip` — new test asserting full-figure conversion, existing no-resolvable-rate test retitled/kept
- [x] `ChannelSalesTable` — new test asserting per-channel + Total row conversion
- [x] `ProductSalesTable` — new test asserting Net sales conversion via the shared headline rate
- [x] `pnpm type-check` (apps/web) clean
- [x] `pnpm lint` (apps/web, touched files) clean
- [x] Full `apps/web` analytics test suite: 212/212 passing

Closes #2779


piotrswierzy (commented):
## Tech-lead review (delta vs `2778-exchange-rate-endpoint` only)

**Verified good:** direction is right — `AppliedRate.from` = native, `to` = display, and `convertToDisplay` multiplies, matching ADR-040's "`to` units per one `from` unit". The `current-rate` branch guards it structurally via `rate.from === nativeCurrency`. ADR-040's analytics-vs-statutory line is explicitly honoured (type doc + the "Not an invoice rate" disclosure row). The REJECTED-APPROACH note was narrowed rather than deleted, and the committed mockup was updated — the `frontend-architecture.md` § UX Mockups obligation is met.

### IMPORTANT

1. **`display-currency.lib.ts` — `resolveReportingCurrencyRate`, `order-date` branch.** It returns `appliedRates[0]` with **no** `from === nativeCurrency` check; the invariant is only a comment. Direction failure here is silent and plausible-looking — exactly the class of bug this PR exists to close. Apply the same `.find((r) => r.from === nativeCurrency)` in both branches; the "never more than one" claim makes it free.

2. **Same function — `to` is never validated.** All three call sites label the converted figure with `conversion.displayCurrency`, but the lookup only matches `from`. Widen the input type to carry `displayCurrency` and match both ends (`r.from === nativeCurrency && r.to === conversion.displayCurrency`), so a pivot leg or a stale entry can't be rendered under the wrong label.

3. **`sales-analytics.types.ts` — `AppliedRate` JSDoc is false as written.** It states the app "never `Number()`s it for arithmetic, only for display formatting", while `convertToDisplay` does `amount * Number(reportingRate.rate)` in three files. Float is acceptable for a display-only figure, but fix the comment — it's the artifact the next author will trust.

4. **Duplication of currency arithmetic.** `convertToDisplay` / `displayCurrencyFor` are copied verbatim into `analytics-kpi-strip.tsx`, `channel-sales-table.tsx`, `product-sales-table.tsx`. Export one factory from `display-currency.lib.ts` (which already owns this policy) instead of three copies that can drift independently.

### SUGGESTION

5. `product-sales-table.tsx` — while `useSalesAnalyticsQuery` is in flight, rows render native and then flip to the display currency. Consider gating on `salesQuery.isLoading`. Also, "no extra request" holds only when the KPI strip is co-mounted.
6. Defence in depth: compare `row.channel.currency` / `row.currency` to `headline.currency` before applying the headline rate — cheap, and the branch already exists.
7. `index.css` — `gap: 2px` / `4px`; prefer `var(--space-1)` per the style guide's 4px-scale rule.

No blocking issues; happy to approve once 1–3 land.

piotrswierzy (approved):
## Re-review at `4fcffc1` (12 files vs `2778-exchange-rate-endpoint`)

All seven items from the previous round are addressed, and I verified each in the diff rather than taking the checklist's word for it:

1. **`order-date` branch no longer returns `appliedRates[0]` blind.** `resolveReportingCurrencyRate` is now one branch-free `find`, so the direction invariant is enforced identically in both modes.
2. **Both ends are matched** — `rate.from === nativeCurrency && rate.to === conversion.displayCurrency`. This is the fix that matters most: every call site labels the figure with `conversion.displayCurrency`, so matching only `from` left a pivot leg or stale entry renderable under the wrong label. Reported === applied is now structural.
3. **The `AppliedRate` JSDoc is corrected** and now says plainly that the app *does* `Number()` the rate in `convertToDisplay`, and why that float is acceptable (display-only, never written back or re-audited). The old wording was the artifact a future author would have trusted.
4. **`convertToDisplay` / `displayCurrencyFor` are one factory** (`createReportingCurrencyConverter`) instead of three copies, and it carries the finding-6 defence in depth internally (`nativeCurrency === reportingCurrency` before applying), so a divergent row gets its native amount back rather than someone else's rate.
5. `product-sales-table.tsx` gates on `salesQuery.isLoading`, so rows no longer render native and then flip.
7. `index.css` uses `var(--space-1)`.

### ADR-040 conformance

- **Direction**: `AppliedRate.from` = native, `to` = display; every conversion is `amount * Number(rate.rate)`. No divide anywhere in the diff — which is the whole point, since the production bug this closes (29 000 PLN as ~20 000 "EUR") was a divide of two unrelated totals.
- **The audited value stays a string**: `rate.rate` is `Number()`'d only at the multiply and at the display-rounding site, never persisted or re-emitted.
- **Refusal over guess** is preserved: `resolveReportingCurrencyRate` returning `null` means every figure stays native, and `buildRateProvenanceDefinitions` returns `[]` for an empty set so the trigger is not rendered over an empty popover.
- **Analytics-vs-statutory** is disclosed in the UI itself (`Not an invoice rate`), not only in a comment — correct, given ADR-040 states that warning positively.
- The reuse-the-headline-rate argument in `ChannelSalesTable` / `ProductSalesTable` rests on ADR-040's one-system-wide-reporting-currency invariant, and the converter still checks the row's own currency before applying, so the argument does not have to hold for the code to be safe.
- I checked the mode copy against the backend rather than the field names: `display-currency.types.ts` documents `order-date` as "applies a single current rate to the whole total" and `current-rate` as per-native-currency conversion, and `buildRateProvenanceDefinitions`' two texts match that exactly (see SUGGESTION 2 for the label).

`frontend-architecture.md` obligations met — logic lives in `features/analytics/lib`, no raw `fetch`, the committed mockup was updated in the same PR, and `useSalesAnalyticsQuery` reuses the shared cache entry rather than adding a request.

### IMPORTANT

1. **`formatAppliedRateLine` rounds the audit figure.** `maximumFractionDigits: 4` against a `numeric(18,8)` column: `4.25` renders fine, but an inverted PLN→EUR rate of `0.23529412` renders `0.2353`, and multiplying the displayed amount by the displayed rate no longer reproduces the figure. This line is the provenance surface — its entire job is to be checkable. Raise to 8 (or render `rate.rate` verbatim and reserve rounding for a separate "approximately" affordance). The disclosure popover in particular has room for the full value.

### SUGGESTION

2. `display-currency.lib.ts` — the `order-date` infotip term reads **"Rate on order date"** while its own text correctly says "One current rate applied to the whole period's total — not each order's own historical rate." The body corrects the heading, which is a slightly odd thing to ask an operator to read. The mode's backend name is the misnomer here, but the disclosure is where you can afford to name it honestly, e.g. "Period rate (order-date mode)".

3. `formatAppliedRateLine` uses `new Intl.NumberFormat(undefined, …)`. `frontend-architecture.md § i18n` introduced `useNumberFormat` / `getBcp47Locale` precisely to stop ad-hoc `Intl` instantiation drifting from the app locale. A pure lib function cannot call the hook, so either take a formatter argument or resolve through `getBcp47Locale` — otherwise this one number follows the browser locale while everything around it follows the seam.

4. `.kpi-card__qualifier-rate` is `0.6875rem` in `--text-muted`. On a line whose purpose is evidence, that is small and low-contrast; worth a quick check against the style guide's contrast rule before it ships.

### Assessment

The core change is exactly right, and the reasoning is now enforced by code rather than by comment: one rate resolved from a matched `(from, to)` pair, one shared converter, native fallback whenever the match fails. The `REJECTED APPROACH` note being *narrowed* rather than deleted is the right call — the specific mistake is still recorded alongside why it is now solvable. No BLOCKING issues.

🔄 **Approve with changes** — please fix the rate rounding (IMPORTANT 1) before merge; the rest are optional.

norbert-kulus-blockydevs:
Addressed all four items from the latest review (commit 3409b695d).

**1 (must fix).** `formatAppliedRateLine` now renders the applied rate at full precision instead of rounding to 4 fraction digits. Concretely: an inverted PLN to EUR rate of `0.23529412` previously rendered as `0.2353`; it now renders as `0.23529412`. This matters because the whole point of that line is that it can be multiplied out and checked (amount x displayed rate should reproduce the displayed figure), and 4 digits silently broke that for any inverted or less-round rate. The formatter is now built with `maximumFractionDigits: 8` to match the `numeric(18,8)` column the rate is sourced from, and is passed in from the calling component (see item 3) rather than hardcoded, so the same fix applies both to the inline qualifier line and to each row in the disclosure popover.

**2.** The order-date disclosure term no longer contradicts its own body. It read "Rate on order date" while the body correctly explained "One current rate applied to the whole period's total, not each order's own historical rate" - an odd thing to hand an operator to read. Renamed to "Period rate (order-date mode)"; the body text is untouched.

**3.** `formatAppliedRateLine` and `buildRateProvenanceDefinitions` now take an `Intl.NumberFormat` as an explicit parameter instead of instantiating `new Intl.NumberFormat(undefined, ...)` internally. Since they're pure lib functions they can't call `useNumberFormat()` themselves, so the formatter is built once in `AnalyticsKpiStrip` via `useNumberFormat(RATE_FORMAT_OPTIONS)` and threaded through - the same pattern the file already uses for its ratio/percent/delta formatters. This finding is not moot even with the full-precision fix from item 1, since the ad hoc formatter was still the mechanism doing the rendering; it's now locale-aware.

**4.** Checked `.kpi-card__qualifier-rate` against the style guide. Contrast: `--text-muted` on `--bg-surface` computes to roughly 5.5:1 in light mode and 4.7:1 in dark mode, both above the WCAG AA 4.5:1 threshold for normal text. Type scale: `0.6875rem` is nominally reserved in the canonical scale table for eyebrow/uppercase/mono-caps copy, but it's already the established size for compact secondary/mono text broadly across `index.css` (100+ non-uppercase uses), so singling out this one selector for a bump would be inconsistent with actual practice rather than a fix. Left unchanged.

Note: the branch was rebased onto the rebuilt parent branch (`2778-exchange-rate-endpoint`), so the PR diff is back to a clean 12 files.

Issue #2411, opened 2026-08-22
[TASK] Frontend — order-detail fulfilment-task panel (work-grain holds)


piotrswierzy:
> Part of the OMS programme backlog (**Wave 3a**, stream S3, size M). Backlog slug: `W3a-21`. Sibling `W3a-*` slugs resolve via the Wave-3a epic's checklist. **Standing S3 rule**: Zod schemas over new projections use `.nullish()`, never `.optional()` (#939).

## Problem / Context

Design §5.2: work holds surface at the **work grain** (worklist + order-detail panel); the orders-*list* hold chip renders only the order-grain `activeHoldReason` — one projection, one derivation input, no second contradictory surface. Promoting a work-hold rollup to an order-list signal is open question §12.10 and is deliberately **not** a column.

## Proposed Solution

A panel on the existing order-detail page listing the order's tasks, their counters, their holds, and the routing explanation — the "why did this order go here" answer — rendered from `RoutingExplanationStep[]` via display labels.

## Classification

**Type**: Frontend
**Layer**: Interface
**File(s)**: `apps/web/src/features/orders/**`

## Dependencies

- #2406 (`W3a-19`), #2410 (`W3a-20`)

## Acceptance Criteria

- [ ] The orders list is unchanged (no new column, no work-hold chip)
- [ ] Every Zod schema over a new backend projection uses `.nullish()`, never `.optional()` — OL serialises an absent optional as JSON `null`, and a `null` sub-field silently drops the whole section under `.optional()` (the #939 lesson)
- [ ] Explanation renders vendor rule names through display labels, never raw identifiers
- [ ] Tests added or updated for non-trivial logic
- [ ] `shared` imports no `features`/`pages`

Issue #2410, opened 2026-08-22
[TASK] Frontend — desktop fulfilment worklist with manual mark-picked / mark-shipped


piotrswierzy:
> Part of the OMS programme backlog (**Wave 3a**, stream S3, size L). Backlog slug: `W3a-20`. Sibling `W3a-*` slugs resolve via the Wave-3a epic's checklist. **Standing S3 rule**: Zod schemas over new projections use `.nullish()`, never `.optional()` (#939); the new feature folder ships its public `index.ts` barrel and registers its slug in both `.eslintrc.js` `no-restricted-imports` pattern groups.

> **Size justification (L):** the list, the two write actions and the server-derived `supportedActions` + optimistic-token 409 handling are one surface; shipping the list alone would render actions the operator cannot take.

## Problem / Context

REVIEW D10 draws the cut line here: 3a ships a **desktop worklist**, which lights the 3PL story with no floor UI, and defers the wizard-scale scan surface to 3b.

## Proposed Solution

New route + feature module under `apps/web/src/pages/fulfillment/` and `apps/web/src/features/fulfillment/`; TanStack Query for server state, URL search params for filters (`docs/frontend-architecture.md`). Renders tasks grouped by location/method with line counters ("3 of 5"), work-grain hold chips, and the server-supplied actions — **a button exists because the server said so**. Manual `mark-picked` / `mark-shipped` post the optimistic token; a 409 refreshes and re-renders the action set with a toast. **Naming (P9):** "fulfilment tasks", "Who decides X?" — never *authority*, *posture*, *FulfillmentWork*. Mobile + tablet in scope.

## Classification

**Type**: Frontend
**Layer**: Interface
**File(s)**: `apps/web/src/pages/fulfillment/**`, `apps/web/src/features/fulfillment/**`

## Dependencies

- #2406 (`W3a-19`)

## Acceptance Criteria

- [ ] No client-side state machine: every action originates from `supportedActions`
- [ ] The 409 path is covered by a test and produces a refreshed, non-destructive UI
- [ ] Desktop + tablet + mobile layouts verified
- [ ] Copy audit test: none of the three banned words appears in user-visible strings
- [ ] Tests added or updated for non-trivial logic
- [ ] `shared` imports no `features`/`pages`

Issue #2396, opened 2026-08-22
[TASK] CORE Application — ingestion intercept (none / ambiguous / selected; a held order never reaches `syncOrder`)


piotrswierzy:
> Part of the OMS programme backlog (**Wave 3a**, stream S1, size M). Backlog slug: `W3a-7`. Sibling `W3a-*` slugs resolve via the Wave-3a epic's checklist. **Migration-bearing**: use a synthetic-sequential migration timestamp (`docs/migrations.md` rule 3).

## Problem / Context

Design §5.5. The intercept sits in `libs/core/src/orders/application/services/order-ingestion.service.ts` **between `persistOrder` and `syncOrder`**. A hold that still mirrors the order into the shop is not a hold.

## Proposed Solution

Gate-reports / caller-persists (the shipped ADR-041 #2100 shape): the fulfillment service **returns** the outcome and `OrderIngestionService` writes it — the one-way edge that keeps `fulfillment → orders` from becoming a DI cycle, with the reason unions living in the leaf. Three arms: `none` → today's path; `ambiguous` → persist reason, today's path; `selected` → `route()` → create work → **a held order does not reach `syncOrder`**. The write is **level-triggered, not sticky** (#2100 discipline): the outcome is re-decided per transition and the writer stores the answer including `null`, which is what clears a stale reason.

**Where it is persisted, named explicitly:** two new nullable `order_records` columns — `fulfillmentBlockReason` (the leaf's reason union) and `fulfillmentBlockDetail` (free text) — added by this issue's migration. `OrderIngestionService` is their **single writer**, and both columns are **omitted from the repository's `toOrm`** — the `salesDocumentBlockReason` / `cancelledAt` precedent: `persistOrder` runs *before* the gate on every ingestion, so round-tripping them would null-then-reset the value and let a stale in-memory read stomp a reason a peer transition just wrote.

**Reconciling design §5.2.** §5.2 says "no new `order_records` column is written by this layer". That claim is scoped to the routing **decision**, which lives in `routing_decisions` at its own grain (#2394, `W3a-5`) and is correctly not duplicated onto the order. The routing **block reason** is a different fact at order grain — it answers "why is this order not moving", which is an order-level operator question — and at that grain it *is* a new `order_records` column, following the #2100 shape verbatim. §5.2's wording gets the same narrowing in the docs PR.

## Classification

**Type**: CORE
**Layer**: Application + Infrastructure (**migration-bearing**)
[...]

PR #2789, opened 2026-09-01
fix(web/orders): an unknown fulfillmentState degrades one cell, not the page (#2678)

### LINKED ISSUE #2678: An unknown `fulfillmentState` crashes the whole /orders page instead of degrading one cell
Found during the #2388 responsive audit, via deliberately invalid seed data. Reported rather than fixed there, because it is not a Wave-2 defect and not reachable from valid data.

## What happens

An unrecognised `fulfillmentState` value hits an unguarded `Record` lookup, and the result is `undefined.tone` — which throws during render and takes down **the entire `/orders` page**, not just the offending cell or row.

## Why it is not urgent

The union is closed and compiler-enforced, so no valid backend response can produce it. It required hand-crafted invalid seed data to trigger.

## Why it is still worth closing

The failure mode is disproportionate to the cause, and there are three realistic routes to an unrecognised value that do not involve a bug in the union:

1. **A rolling deploy.** The backend ships a new union member before the browser bundle does, and every user on the old bundle loses the orders page until they hard-reload. This is exactly the version-skew scenario #2589 dealt with on the who-decides page, where the fix was to loosen the parse and let an existing fallback do its job.
2. **A stale cached bundle** after a deploy, same shape.
3. **A cached API response** replayed against a newer or older client.

In each case the honest degradation is one cell reading as unknown, never a blank page.

## Scope

1. Make the lookup total — an unrecognised value renders a neutral fallback rather than throwing.
2. Apply the same treatment to sibling `Record`-keyed tone/label lookups on that page if any share the shape. **Find them by grep rather than assuming this is the only one.**
3. **Verify by making it fail first**: render with an unrecognised value, see the crash, then see the neutral cell.

## Acceptance criteria

- [ ] An unrecognised `fulfillmentState` renders a neutral cell; the page still renders.
- [ ] A test drives an unrecognised value directly, and was seen red before the fix.
- [ ] Any sibling lookups with the same unguarded shape are covered or explicitly listed as out of scope.

Non-blocking, low priority. Related: #2589 (the version-skew precedent and its `WHY_CODE_FALLBACK` dead-code lesson).

piotrswierzy:
Closes #2678

## Root cause

`fulfillmentBadge` was a single line:

```ts
return ORDER_FULFILLMENT_META[state ?? 'not-shipped'];
```

A lookup that is **total on the type but partial at runtime**. `GET /orders` is not schema-parsed anywhere in `features/orders/api/` (zod appears in that feature only as a form resolver), so `FulfillmentRollupStateValue` is a *claim about the wire*, not a guarantee about it.

An unrecognised value returns `undefined`, and all three call sites read `.tone` off it immediately — `orders-list-page.tsx` (desktop Ship-by cell, mobile card, mobile badge row) and `dispatch-risk-page.tsx`. Because that read happens inside a `DataTable` cell renderer, the throw unmounted the **whole `/orders` route**, not the offending cell.

Three routes reach an unrecognised value with no bug in the union, all version skew (the #2589 shape): a rolling deploy shipping a backend member before the browser bundle, a stale cached bundle after a deploy, or a replayed cached API response.

## Why this degrades rather than lies

Two cheaper fixes were rejected, and the reasoning is written at the seam so it is not re-litigated later:

- **`?? 'not-shipped'` widened to cover the miss** — this renders "Not shipped" about an order whose state this build cannot name. That is a quiet false claim about the operator's own data, and it is *worse* than the crash it replaces: a blank page is at least obviously broken.
- **Return `null` / render nothing** — a silent drop. The row then reads as one that simply has no fulfilment fact.

So the unrecognised value is **surfaced verbatim** in a neutral badge — `Unknown (teleported)` — the way an unrecognised marketplace validation code is surfaced rather than dropped (#2231). `neutral` is deliberate: OL has no idea whether the state is good or bad, and any status tone would assert one.

**Absent stays `not-shipped`.** That is a real backend contract, not a fallback, and absent and unrecognised remain distinguishable — a test pins exactly that.

## Sibling lookups — found by grep, not assumed

| Lookup | Verdict |
|---|---|
| `ORDER_FULFILLMENT_META[state ?? …]` | **fixed** — the crash |
| `ORDER_SLA_META[slaState]` | **fixed** — same shape, different symptom (below) |
| `PAYMENT_BADGE_META[status]` | safe — narrowed upstream by `readPaymentStatus`'s soft-parse |
| `ORDER_LIFECYCLE_PHASE_META[phase]` | safe — behind `isOrderLifecyclePhase` |
| `ORDER_HEALTH_META.<literal>` | safe — dot access on compile-time literals |

The clean underlying story: **snapshot** fields already go through `parseOrderSnapshot`'s soft-parse; the **top-level `OrderRecord` columns** did not. `lifecyclePhase` had a guard; `fulfillmentState` and `slaState` were the two that had none.

`slaBadge` shared the shape but **never crashed** — every call site guards on `sla ?`, so an unrecognised bucket rendered nothing, indistinguishable from the `none` member which legitimately means "no badge". A silent drop of the signal is its own defect, so it gets the same treatment. Note membership is established **before** the `none` check: `none` is a *member* of the union, not a miss, and testing it first conflates "the backend said none" with "the backend said something unreadable". The compiler found that ordering flaw during the gate.

Also removes a dead `if (!badge) return null` in `dispatch-risk-page` — `fulfillmentBadge` has never returned `null`, so that guard read as coverage it never provided (the #2589 `WHY_CODE_FALLBACK` dead-code lesson).

## Tests — observed red, not merely written

Every new test was run against a reverted fix before being kept:

- `orders-list-page.test.tsx` fails with the exact reported symptom, `TypeError: Cannot read properties of undefined (reading 'tone')`. It drives an unrecognised value through the **real** query → page → table → cell path and asserts both the offending row and its neighbour still render — so a lib-only fix that still crashed the page would not pass it.
- `order-health.test.ts` (5 cases) covers the neutral badge, the absent-vs-unrecognised distinction, truncation of a pathological value, the blank-value case, and the SLA silent-drop.

## `/pr-review` round — three findings, all applied (`c8dfce5cc`)

1. **The empty string diverged between the two resolvers, and produced a label that named nothing.** `slaBadge` short-circuited on `!slaState`, so a JSON `""` was swallowed into the same answer as the `none` *member* — "no badge, nothing is wrong" — while `fulfillmentBadge` rendered **`Unknown ()`**, which claims to quote the offending value and quotes nothing. Same condition, two answers, one useless. Both now short-circuit only on genuinely absent, and a blank/whitespace-only value renders a bare `Unknown`. New test, observed red first.
2. **The tests cast inputs `as FulfillmentRollupStateValue` / `as SlaStateValue`**, misrepresenting the contract under test — the parameter is deliberately `string`, because that is what the wire delivers. Casting to the union implied the opposite of the fix. Dropped.
3. **Documentation gap** — the change introduces a new operator-visible status label, and `frontend-ui-style-guide.md` governs that vocabulary. Recorded under § Status Language as three rules (never render an unrecognised value as a known one; never drop it silently; surface it neutrally), plus the note that absence is a separate question keeping whatever contract the backend documents — a guard conflating the two reintroduces rule 1.

## Gate

`lint` (0 errors), `type-check`, `check:invariants` and the full `pnpm test` all green via the pre-commit hook on **both** commits — `apps/web` 449 files / 4833 tests, `libs/core` 5684, every integration package passing.

## Note for #2410 / #2411

#2396 and #2406 are backend-only so far; no new FE cell consumes `fulfillmentBlockReason` or `supportedActions` yet. Those fields are the next place this class of bug appears — the guard pattern here (`is*` beside the union in `orders.types.ts`, neutral surfaced badge at the render seam, now written up in the style guide) is the one to copy. I deliberately did **not** pre-guard a field the FE does not yet read, since that would be the same dead-fallback mistake this PR removes.

### DIFF
--- a/apps/web/src/features/orders/api/orders.types.ts
+++ b/apps/web/src/features/orders/api/orders.types.ts
@@ -86,11 +86,37 @@ export const FulfillmentRollupStateValues = [
 ] as const;
 export type FulfillmentRollupStateValue = (typeof FulfillmentRollupStateValues)[number];
 
+/**
+ * Coerce an untrusted value — a field on a payload from a newer backend than
+ * this bundle, or a replayed cached response — to the union.
+ *
+ * Deliberately no fallback, matching `isOrderLifecyclePhase`: defaulting an
+ * unrecognised state to `not-shipped` would report "this order has not shipped"
+ * about an order whose real state this build cannot name. `fulfillmentBadge`
+ * (lib/order-health.ts) uses this to degrade one cell instead of throwing
+ * (#2678).
+ */
+export function isFulfillmentRollupState(value: unknown): value is FulfillmentRollupStateValue {
+  return (
+    typeof value === 'string' && (FulfillmentRollupStateValues as readonly string[]).includes(value)
+  );
+}
+
 // Ship-by SLA bucket (#1108). Hand-mirrored from `SlaStateValues` in
 // `@openlinker/core/orders`. BE-owned (single source of truth): the FE consumes
 // `slaState` and only renders the live countdown from `dispatchByAt`.
 export const SlaStateValues = ['none', 'on_track', 'at_risk', 'overdue'] as const;
 export type SlaStateValue = (typeof SlaStateValues)[number];
+
+/**
+ * Coerce an untrusted `slaState` to the union, same contract and same reason as
+ * `isFulfillmentRollupState` above. An unrecognised bucket must never collapse
+ * into `none` — `none` renders no badge at all, which would silently drop the
+ * one signal telling the operator this order's SLA is in a state OL cannot read.
+ */
+export function isSlaState(value: unknown): value is SlaStateValue {
+  return typeof value === 'string' && (SlaStateValues as readonly string[]).includes(value);
+}
 
 // Why OpenLinker issued no sales document (invoice or fiscal receipt) for an
 // order (#2100/#2156, ADR-041 decision 11). Hand-mirrored from
@@ -110,8 +136,7 @@ export const SalesDocumentGateBlockReasonValues = [
   'trigger-model-manual',
   'trigger-model-batched',
 ] as const;
-export type SalesDocumentGateBlockReasonValue =
-  (typeof SalesDocumentGateBlockReasonValues)[number];
+export type SalesDocumentGateBlockReasonValue = (typeof SalesDocumentGateBlockReasonValues)[number];
 
 export const SalesDocumentUnresolvedReasonValues = [
   'no-matching-rule',
--- a/apps/web/src/features/orders/lib/order-health.test.ts
+++ b/apps/web/src/features/orders/lib/order-health.test.ts
@@ -20,7 +20,12 @@ import {
   syncCellLabel,
   totalUnits,
 } from './order-health';
-import type { OrderRecord, OrderSyncStatus } from '../api/orders.types';
+import type {
+  OrderRecord,
+  OrderSyncStatus,
+  FulfillmentRollupStateValue,
+  SlaStateValue,
+} from '../api/orders.types';
 
 function syncEntry(overrides: Partial<OrderSyncStatus>): OrderSyncStatus {
   return {
@@ -57,7 +62,7 @@ describe('deriveOrderHealth', () => {
         recordStatus: 'source_deleted',
         syncStatus: [syncEntry({ status: 'failed' })],
         mappingFailureReason: 'variant ol_variant_b deleted at the master',
-      }),
+      })
     );
     expect(result.key).toBe('source_deleted');
     expect(result.tone).toBe('error');
@@ -71,7 +76,7 @@ describe('deriveOrderHealth', () => {
 
   it('should return awaiting_mapping when recordStatus is awaiting_mapping (highest precedence)', () => {
     const result = deriveOrderHealth(
-      order({ recordStatus: 'awaiting_mapping', syncStatus: [syncEntry({ status: 'failed' })] }),
+      order({ recordStatus: 'awaiting_mapping', syncStatus: [syncEntry({ status: 'failed' })] })
     );
     expect(result.key).toBe('awaiting_mapping');
     expect(result.tone).toBe('warning');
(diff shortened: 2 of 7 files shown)


piotrswierzy (commented):
## ✅ Review verdict: approve (posted as a comment — GitHub blocks self-approval)

Reviewed the delta against its stacked base `oms-programme-wave-3a` (6 files, FE-only). **No BLOCKING or IMPORTANT findings.**

**Verified**

- **The fix is at the right seam.** `fulfillmentBadge` / `slaBadge` widened to `string | null | undefined` is the honest type for an unparsed `GET /orders` payload; the closed union stays the *declared* contract and the guards (`isFulfillmentRollupState` / `isSlaState`) sit beside their unions in `orders.types.ts`, matching the `isOrderLifecyclePhase` precedent. Guards are pure — no layering issue (`features → shared` only).
- **Degradation is honest, not silent.** Neutral tone is correct per `frontend-ui-style-guide.md` § Color Usage Rules (a status hue would assert a meaning OL doesn't have), and the label carries text, so the badge never relies on colour alone. Absent-≡-`not-shipped` stays a contract and is pinned by a test.
- **`slaBadge` ordering is right**: membership before the `none` arm — `none` is a member, not a miss.
- **All five call sites re-checked** (`orders-list-page.tsx:845/849/1621/1725/1726`, `dispatch-risk-page.tsx`): every one reads `.tone`/`.label` off a now-total return; the `if (!badge) return null` removal is genuinely dead code. Sibling literal comparisons (`fulfillmentState === 'dispatched'`) degrade to `false` for an unknown value — no second crash path.
- **`'neutral'` is a real `StatusBadgeTone`** (`shared/ui/status-badge.tsx:13`).
- **Tests reach the actual regression.** `orders-list-page.test.tsx` drives the value through query → table → cell and asserts the neighbour row survives — a lib-only test would not have caught it. Truncation and absent-vs-unrecognised are both covered.

**SUGGESTION** — `apps/web/src/pages/orders/orders-list-page.tsx:159,164`
`isSlaState` and `isFulfillmentState` already exist as page-local guards over the same two unions, so there are now two copies of each rule (one pair name-colliding). Delete the locals and import the shared guards from `../../features/orders/api/orders.types` — the page already imports `SlaStateValues` from there — so a future union member can't be added to one and missed by the other.

**SUGGESTION** — `docs/plans/implementation-plan-unknown-fulfillment-state.md` § 3.2
Plan says truncate at 32 chars; shipped `UNKNOWN_STATE_MAX_CHARS = 16`. The 16 is better justified (17-char pill budget) — align the plan so the committed doc doesn't contradict the code.

**SUGGESTION** — ~120 lines of pure Prettier reflow in the two test files dilute the substantive diff. Consistent with `.prettierrc`, so fine to keep; noting it for reviewers further up this stack.

Issue #2621, opened 2026-08-27
[TASK] Integration/Allegro — non-blocking quantity acknowledgement


norbert-kulus-blockydevs:
## Problem / Context

The quantity write to Allegro blocks on the platform's acknowledgement. A single offer write is throttled to roughly 60/minute regardless of the global 9000/minute limit, so blocking makes the propagation queue depth a direct function of platform latency rather than of our own throughput.

## Proposed Solution

Acknowledge asynchronously: dispatch the write, record the pending acknowledgement, and reconcile it separately instead of holding the job.

## Classification

**Type**: Integration / CORE
**Layer**: Application / Infrastructure
**File(s)**: `libs/integrations/allegro/src/infrastructure/adapters/**`, `libs/core/src/inventory/application/services/inventory-sync.service.ts`

## Dependencies

- Ordering guard from the sibling epic — asynchronous acknowledgement widens the out-of-order window

## Acceptance Criteria

- [ ] A quantity write does not hold its job until the platform acknowledges
- [ ] A failed acknowledgement is reconciled, not lost
- [ ] Tests cover the pending-then-failed path

Issue #2024, opened 2026-08-11
[TASK] CORE — Persist channel-side price + quantity for marketplace offers (new offer_commercial_snapshots table)


norbert-kulus-blockydevs:
## Problem / Context

`/listings` needs to render Price and Quantity per offer row (per #1965 mockup handover), but today
these two fields exist only via a live per-row marketplace call `GET /listings/:id/offer`
(`apps/api/src/listings/http/listings.controller.ts:245-311`, 30s cache) — unusable at list scale
(20 rows/page = 20 live marketplace API calls against the seller's own rate limit).

Investigation found both marketplace adapters already fetch price+quantity for free: Allegro's
`getOfferStatus` and `getOffer` both call the same endpoint `GET /sale/product-offers/{id}`
(`libs/integrations/allegro/src/infrastructure/adapters/allegro-offer-manager.adapter.ts:624-651` vs
`:724-751`); Erli's `getOfferStatus`/`getOffer` both call `GET /products/{externalId}`
(`libs/integrations/erli/src/infrastructure/adapters/erli-offer-manager.adapter.ts:401-417` vs
`:432-469`). The existing `marketplace.offer.statusSync` job (#816, `OfferStatusSyncService.sync`,
`libs/core/src/listings/application/services/offer-status-sync.service.ts:56-132`) already makes this
exact call per offer, hourly, and discards everything except publication status.

## Proposed Solution

1. New table `offer_commercial_snapshots` mirroring `offer_status_snapshots`'s shape
   (`libs/core/src/listings/infrastructure/persistence/entities/offer-status-snapshot.orm-entity.ts`):
   unique on `(externalOfferId, connectionId)`, columns `price numeric`, `currency text`,
   `availableQuantity integer`, `lastCommercialSyncedAt timestamptz`, no FK constraints (matches the
   existing snapshot-table convention). New migration under `apps/api/src/migrations/`.
2. New repository port + adapter (`OfferCommercialSnapshotRepositoryPort` / implementation) mirroring
   `OfferStatusSnapshotRepositoryPort`'s shape, with an `upsert` method.
3. Extend `OfferStatusSyncService.sync()`/`refreshOne()` (or add a sibling call in the same loop) to
   also read price/quantity from the SAME per-offer adapter response already being fetched for status,
   and upsert into the new table. Do not add a second per-offer HTTP call.
4. Keep the existing hourly cadence and rolling scan-offset cursor — no new job, no new cron
   registration.

## Classification

[...]

PR #2800, opened 2026-09-02
feat(currency): resolve the publication day before the exchange-rate cache read

### LINKED ISSUE #2777: [TASK] CORE/Integration — Resolve the publication day before the exchange-rate cache read

## Problem / Context

The `exchange_rates` registry misses its own cache on every non-publication day, forever, so those orders pay a live provider call each. This affects **ordinary ingestion**, not only the epic #2452 remediation path — roughly 2 days in 7 of order placement dates.

The mechanism is documented as accepted debt in the service that suffers from it (`libs/core/src/currency/application/services/currency-rate.service.ts`, file header):

- The pre-fetch read is keyed on the **candidate** day the caller asked for (`currency-rate.service.ts:81`, `repository.findByKey(input)`).
- The write is keyed on the day the source **actually published for**. `ExchangeRateProviderPort.fetchRate`'s contract states this outright: "the returned `ExchangeRate.rateDate` is the day the source ACTUALLY published for, which may precede `input.on`, and is what the registry keys the row under" (`libs/core/src/currency/domain/ports/exchange-rate-provider.port.ts:49-53`).
- For a weekend or holiday candidate the two keys differ, nothing is ever written under the candidate, and the read misses **permanently**. The insert is then swallowed as a `DuplicateExchangeRateError`.

Cost per affected order is not one HTTP call. `NbpExchangeRateAdapter` walks back up to `NBP_MAX_WALK_BACK_ATTEMPTS = 8` sequential days (`libs/integrations/fx/src/infrastructure/adapters/nbp-exchange-rate.adapter.ts:67`, loop at `:224-243`), each fetching one quote **per leg** — so a holiday candidate on a pivot pair is up to ~16 sequential calls, per order, every time.

The header names the fix and parks it: "Memoising the candidate-to-published-date mapping needs its own persisted table and belongs to the persistence phase (#2124)". This issue proposes a cheaper mechanism that needs no table and no migration.

## Proposed Solution

Let the adapter — the only layer that owns a publication calendar — resolve the candidate to its likely publication day **before** the cache read.

Add one **optional** pure method to `ExchangeRateProviderPort`:

```ts
/**
 * A day <= `candidate` such that no day in `(returned, candidate]` is a
 * publication day for this source. Pure, synchronous, no I/O.
 *
 * Optional: an adapter that declares nothing keeps the pre-#2777 behaviour
 * (the cache read is keyed on the candidate).
 */
resolveLikelyPublicationDay?(candidate: string): string;
```

`CurrencyRateService.getRateFor` **probes for the method** rather than trusting the type (the ADR-046 rule — an out-of-tree plugin compiled against an older `libs/core` would otherwise satisfy the interface and throw), and keys `findByKey` on the resolved day when one comes back.

Adapter implementations:
- **NBP**: delegate to the private `resolveWorkingDayAtOrBefore` it already has (`nbp-exchange-rate.adapter.ts:224`, built on `isPlWorkingDay` / `previousWorkingDay` from `@openlinker/shared/date`). No new calendar logic.
- **ECB**: weekend-only variant (ECB publishes on Polish-only holidays, which is exactly why this rule cannot live in core — see ADR-040's note on `resolveRateDate` being deliberately calendar-neutral).
- **Fake**: identity, so test determinism is unchanged.

**Why this is safe rather than merely likely-safe.** A wrong guess can only cause a cache *miss*, which falls through to the existing `fetchRate` path — it can never produce a wrong stamp. A *hit* is provably the right row: a row exists under day `X` only because the source supplied `effectiveDate = X`, and the contract above guarantees no publication day lies between `X` and the candidate, so the hit equals what `fetchRate` would have returned.

**Nothing is written under a date the source never published.** `ExchangeRateRepositoryPort`'s append-only guarantee (`libs/core/src/currency/domain/ports/exchange-rate-repository.port.ts:1-18`) and ADR-040's verifiability property are untouched — this changes only which key is *read*.

### Rejected alternatives

- **Widen `findByKey` to "latest row with `rateDate <= candidate`".** Unsafe. If the table holds EUR/PLN for 2026-06-01 (from an old order) and a new order's candidate is 2026-06-10, that read returns a 9-day-stale rate even though the source published on the 10th — we simply never asked. It cannot distinguish "no publication that day" from "not fetched yet", has no bound, and would silently invent a rate assignment the adapter never made, on the one figure whose purpose is auditability. Both NBP's 8-attempt cap and ECB's MAX_LAG guard exist because that bound has to be the adapter's.
- **A persisted `exchange_rate_date_resolutions (source, candidate_date) -> published_date` table** (the #2124 shape). Correct and strictly more machinery: a migration, a table, a repository, and a second cache to reason about. The adapter-declared version is the same idea with the calendar in the layer that already holds one, and needs none of it. Revisit only if a source appears whose calendar cannot be computed purely.
- **An in-memory memo for the duration of one restatement page.** Only helps the remediation path, dies with the process, and leaves ordinary ingestion broken. Superseded by this.

## Classification

**Type**: CORE + Integration
**Layer**: Domain (port) / Application (resolution) / Infrastructure (adapters)
**File(s)**:
- `libs/core/src/currency/domain/ports/exchange-rate-provider.port.ts`
- `libs/core/src/currency/application/services/currency-rate.service.ts` (the pre-fetch read + its file header's "accepted debt" section)
- `libs/integrations/fx/src/infrastructure/adapters/nbp-exchange-rate.adapter.ts`
- `libs/integrations/fx/src/infrastructure/adapters/ecb-exchange-rate.adapter.ts`
- `libs/integrations/fx/src/infrastructure/adapters/fake-exchange-rate.adapter.ts`
- `docs/architecture/adrs/040-order-time-fx-stamping-against-a-system-reporting-currency.md` (§ Consequences — the accepted-cost note this closes)

## Dependencies

- None. Independent of #2775 and #2776; it makes them cheaper but neither blocks it.

## Assumptions

- The method is optional and probed, so no out-of-tree provider is broken and no adapter is required to implement it.
- No new persisted table and no migration.
- Widening the `isExchangeRateProvider`-style guard (if one exists) is **not** done, for ADR-046's reason: it would silently stop recognising an older plugin as a provider at all.

## Assumption to verify before sizing

The same-currency path needs **no** rate lookup at all (`libs/core/src/orders/application/services/order-fx-stamp.service.ts`, step 3 short-circuits with zero I/O beyond the write). If most of a deployment's orders are already native-in-the-reporting-currency, this issue's payoff is small and it should be re-prioritised below the two remediation issues. Take a one-off count of `order_records` grouped by `currency` vs the current reporting currency before scheduling.

## Acceptance Criteria

- [ ] `ExchangeRateProviderPort` declares the optional method, documented with the `<= candidate` / "no publication day in between" contract.
- [ ] `CurrencyRateService` **probes** for the method (`typeof provider.resolveLikelyPublicationDay === 'function'`) rather than relying on the type, with a comment citing the ADR-046 precedent.
- [ ] A spec proves a Saturday candidate resolves to the preceding Friday and the pre-fetch `findByKey` is called with the Friday, with **zero** `fetchRate` calls when that row exists.
- [ ] A spec proves a provider declaring nothing behaves exactly as today (read keyed on the candidate).
- [ ] A spec proves a wrong resolution degrades to a `fetchRate` call and never to a wrong or stale stamp.
- [ ] The NBP adapter reuses `resolveWorkingDayAtOrBefore` — no second copy of the Polish working-day calendar.
- [ ] The ECB adapter's variant is weekend-only, with a comment stating why it must not use the Polish calendar (verified divergence: Corpus Christi 2026-06-04).
- [ ] `CurrencyRateService`'s file header no longer describes the permanent-miss case as accepted debt, and ADR-040 § Consequences is updated to match.

## Branching / merge strategy

```
git checkout -b <branch-name>
```

jakubretajczykBD:
## Summary
- Adds an optional `ExchangeRateProviderPort.resolveLikelyPublicationDay`, probed (not trusted, ADR-046 pattern) in `CurrencyRateService.getRateFor` before the pre-fetch cache read — the read now keys on the day a provider will actually publish for instead of the raw candidate.
- NBP delegates to its existing `resolveWorkingDayAtOrBefore` (no new calendar logic). ECB gets a new, deliberately **weekend-only** variant that must not reuse the Polish calendar (verified divergence: Corpus Christi 2026-06-04). Fake is identity.
- Closes a permanent cache-miss defect: on a weekend/holiday order-placement candidate (~2 days in 7), the write always landed under the earlier published day while the read stayed keyed on the candidate, so every order carrying that candidate paid a live provider call forever (up to ~16 sequential HTTP requests per order on a pivot pair).
- Updates the "accepted debt" language in `currency-rate.service.ts`'s file header and ADR-040 § Consequences to describe the gap as resolved for both shipped providers.

A wrong resolution can only ever cause a cache *miss* (falls through to the existing `fetchRate` path) — it can never produce a wrong or stale stamp. No schema/migration/HTTP change; the method is optional so an out-of-tree provider declaring nothing keeps byte-identical pre-#2777 behaviour.

## Test plan
- [x] Unit spec proves a Saturday candidate resolves to the preceding Friday and `findByKey` is called with the Friday, with zero `fetchRate` calls when that row exists
- [x] Unit spec proves a provider declaring nothing behaves exactly as today (read keyed on the candidate)
- [x] Unit spec proves a resolved day that still misses the cache falls through to a normal `fetchRate` call
- [x] NBP adapter spec: Saturday → Friday, ordinary weekday → itself, both with zero HTTP calls
- [x] ECB adapter spec: Saturday → Friday; Corpus Christi (a genuine Polish holiday) → itself, proving no Polish-calendar dependency; ordinary weekday → itself
- [x] `pnpm --filter @openlinker/core exec jest currency` — 121/121 passing
- [x] `pnpm --filter @openlinker/integrations-fx exec jest` — 111/111 passing (incl. the 5 new cases)
- [x] `pnpm --filter @openlinker/core type-check` / `pnpm --filter @openlinker/integrations-fx type-check` clean
- [x] `eslint` clean on every touched file

Closes #2777


piotrswierzy (approved):
## Tech-lead review — delta vs `2779-analytics-rate-provenance` only (10 files)

Reviewed the PR's own diff, not the two ancestors'. I read `nbp-exchange-rate.adapter.ts` in full on this branch to check the claim the whole design rests on.

### Verified correct

- **Read key changes, write key does not.** `getRateFor` applies the resolved day only to `repository.findByKey`; `fetchRate` still receives the raw `input.rateDate` (pinned by the fall-through spec asserting `on: '2026-08-15'`), and the insert still keys on `fetched.rateDate`. So a row can still only ever exist under a day the source itself supplied as `effectiveDate` — ADR-040's "the registry stores PUBLISHED rates, verifiable against a real table" invariant is untouched. That is the single most important property here and it holds.
- **Probe, not trust.** `typeof provider.resolveLikelyPublicationDay === 'function'` rather than a widened type — the ADR-046 precedent is cited and applied correctly, and the spec asserts `'resolveLikelyPublicationDay' in provider === false` on the default mock so it is the probe, not the type, that is under test.
- **ECB is weekend-only and stays weekend-only.** Reusing `isPlWorkingDay` here would have reintroduced the exact Corpus Christi divergence ADR-040 § Currency records (4.2383 vs the real 4.2368), and the spec pins `2026-06-04 → itself` with zero HTTP calls, so a future "simplification" to share the Polish calendar breaks a test that names the reason.
- **NBP delegates rather than copies** — `resolveWorkingDayAtOrBefore` is the same primitive `fetchQuotesForNearestPublishedDay` already calls, so there is no second Polish calendar to drift.
- Midday-UTC anchoring matches each adapter's existing discipline; a malformed candidate yields `NaN` from `getUTCDay()` and falls through to returning the candidate, i.e. degrades to a miss.
- Optional method ⇒ an out-of-tree provider declaring nothing is byte-identical to pre-#2777, and the old "re-fetch on EVERY call" spec was **retitled and kept** as the fallback-path pin rather than deleted. Good instinct — that test still describes a reachable path.
- No schema, no migration, no HTTP surface. Test coverage matches the branch matrix.

### IMPORTANT

1. **`exchange-rate-provider.port.ts` — the safety claim is stated absolutely and is only half true.** The docblock says "A WRONG answer can only ever cause a cache *miss* … it can never produce a wrong stamp." That holds for an answer that is too **late** (returns a non-publication day → no row → `fetchRate`). It does **not** hold for an answer that is too **early**: if an implementation walks back *past* a genuine publication day X to some earlier day Y, a row very likely exists under Y — from any other candidate that resolved there — so the read **hits**, and `getRateFor` returns day-Y's rate where `fetchRate` would have returned day-X's. That is a silently wrong rate on a financial figure, which is the one outcome ADR-040's whole design exists to prevent.

   Neither shipped adapter can reach it today (ECB is weekend-only; NBP reuses the calendar that already chooses the request URL, and its own class header notes the calendar errs toward *optimistic* — the safe direction). But this docblock is the contract a future adapter author implements against, and as written it invites a walk-back loop that skips holidays it isn't certain about. Please state the asymmetry explicitly: **erring late is free, erring early corrupts a stamp — when in doubt, return the candidate.**

2. **No port-contract suite for a contract whose violation is silent.** `testing-guide.md § Port-contract suites` prescribes exactly this shape — a pure `check*Contract` plus a thin jest wrapper — for a port with several implementations, and gives the reason that applies verbatim here: a suite that looks thorough while asserting nothing. Today the contract ("no publication day lies in `(returned, candidate]`") is enforced by two hand-written cases in one adapter's spec and three in another, with no shared statement, and an optional member handled by *absence asserted in both directions* is precisely the case the guide says must not be left to `it.skip`-shaped coverage. A small shared checker over a case table (weekend, ordinary weekday, the Corpus Christi divergence, method-absent) would bind the third and fourth implementations too. Not blocking for two adapters; it gets expensive at four.

### SUGGESTION

3. `docs/architecture/adrs/040-…md` — the change is edited **into** an existing Consequences bullet. The repo's settled practice for a shipped change to an `Accepted` ADR is a `§ Amendment (#NNNN)` block (ADR-009 § Amendment (#2024), ADR-050 § Amendment (#2609), ADR-067 § Amendment (#2621)), which keeps the original reasoning readable next to what replaced it. Worth converting — the pre-#2777 sentence explained *why* the debt was accepted, and that reasoning is now only visible in `git log`.

4. `docs/architecture-overview.md § 18 Currency` is untouched, and it enumerates this context's seams down to the `sourceRef` NBP-vs-ECB asymmetry. One clause on the new optional rung would keep it current. (Same doc debt as #2785 in this stack — worth clearing both before the epic merges.)

5. `ecb-exchange-rate.adapter.ts` — `addUtcDays` is new local day arithmetic while NBP reaches `@openlinker/shared/date` for its calendar primitives. If a `shared/date` day-shift helper exists it would be the better home; if not, this is fine as-is.

6. `resolveLikelyPublicationDay` is named "Likely", which reads as a hint, while a hit is treated as authoritative. The docblock's guarantee is what makes the hit sound, so the name slightly undersells the obligation — `resolveExpectedPublicationDay` or similar would match the contract's strength.

### Assessment

A genuinely good fix: it removes a permanent cache miss on ~2 days in 7 (up to ~16 sequential provider requests per order on a pivot pair) while leaving the registry's write path — and therefore ADR-040's verifiability guarantee — untouched, and the wrong-answer blast radius is bounded to a miss for both shipped providers. The one thing I would not merge unamended is the port docblock's unqualified safety claim, since it is the artifact the next implementer will trust. No BLOCKING issues.

🔄 **Approve with changes** — fix the port-contract wording (IMPORTANT 1) before merge; IMPORTANT 2 and the doc items can follow.

norbert-kulus-blockydevs:
Addressed all six review items. The branch was rebased onto the rebuilt parent branch (the epic branch underneath had been rebased), so the PR diff is a clean set of changes against `2779-analytics-rate-provenance` again (a subsequent, unrelated commit has since landed on that parent branch, which will show as extra drift in a diff against its moving tip until it settles - not part of this PR).

**1. The asymmetry, stated (merge-blocking item).** `ExchangeRateProviderPort.resolveExpectedPublicationDay`'s docblock now says explicitly: a day resolved too LATE can only ever cause a cache miss (falls through to `fetchRate` unchanged, free); a day resolved too EARLY - walking back past a genuine publication day - can silently hit a row written under an earlier day and return the wrong day's rate on a financial figure, with no error anywhere. An implementer in doubt must return the candidate unchanged. Neither shipped adapter can produce a too-early answer today (ECB only skips weekends; NBP's calendar is the same one that already chose the request day, so it errs optimistic).

**2. A shared port-contract suite.** `@openlinker/core/currency/testing` (`checkPublicationDayContract` + `runPublicationDayContract`) asserts: an ordinary weekday, a Saturday, a Sunday, and the Corpus Christi 2026-06-04 divergence, plus that the optional method is recognised as declared-or-absent in both directions rather than silently skipped. All three shipped implementers - NBP, ECB, and the fake - are now wired through it. I went with the smaller, right-sized version the finding allowed for rather than the full fulfilment-kit shape: one suite file plus one combined coverage/vacuity spec (no separate `contract-vacuity.spec.ts`), since this port has one optional method and three implementers against the fulfilment kit's several ports and many-case suites. It keeps the kit's core discipline: no `it.skip`, a mutation fixture per declared case proving it can fail, and a hard throw (never a silent pass) when handed no subject.

One correction along the way: the port's own guarantee is trivially satisfied by returning the candidate unchanged (the interval `(x, x]` is always empty), so "a Saturday must resolve to Friday" is not actually a universal requirement - it's NBP/ECB's specific optimization, and the fake's identity answer is equally conforming. The suite's fixtures are supplied per-implementer rather than hardcoded, so wiring the fake through it doesn't misreport it as broken.

**3. ADR-040 amendment.** Converted the two-sentence edit folded into an unrelated `§ Consequences` bullet into a proper `§ Amendment (#2777)` block (the ADR-009/050/067 shape), and restored the original pre-#2777 reasoning for why the permanent-cache-miss cost was accepted in the first place (previously only visible in git log).

**4. architecture-overview.md.** Added one clause to `§ 18 Currency` naming the optional method, that it's probed rather than trusted (ADR-046), that only the read key changes while the write path still keys on the source-supplied `effectiveDate`, and the late-vs-early asymmetry. Placed as an additive bullet at the end of the section's list so it shouldn't collide with #2785's separate addition there.

**5. Shared date helper - none exists.** Checked `@openlinker/shared/date`: it exports only Polish working-day arithmetic (`addWorkingDays`, `previousWorkingDay`, `isPlWorkingDay`, `easterSunday`), no generic calendar-day shifter. Left ECB's local `addUtcDays` as is rather than inventing a new shared helper for one two-line function with one caller.

**6. Renamed `resolveLikelyPublicationDay` to `resolveExpectedPublicationDay`** across the port, all three adapters, `CurrencyRateService`, the probe site, every spec, and doc prose that names it. Grepped the whole repo afterward - the only remaining occurrences are two local-variable-name mentions inside a historical implementation-plan document's quoted code excerpt, which name a since-renamed local variable rather than the port method itself.

Verification: `pnpm --filter @openlinker/core type-check` and `pnpm --filter @openlinker/integrations-fx type-check` both clean; `pnpm --filter @openlinker/core exec jest currency` is 134 green (was 121) and `pnpm --filter @openlinker/integrations-fx exec jest` is 126 green (was 111); `pnpm check:invariants` clean.

norbert-kulus-blockydevs:
Small follow-up on the same push: the ADR-040 amendment text and the new test kit's own docblocks/test names used em-dashes in a couple of places, which is against this repo's plain-hyphen convention for new prose. Replaced them with plain hyphens. No functional change; type-checks, the two jest suites, and `pnpm check:invariants` all still green.

PR #2802, opened 2026-09-02
feat(orders,api,web): thread resolved per-line tax rate into coverage detail modals

### LINKED ISSUE #2798: [TASK] CORE+Frontend — Thread resolved per-line tax rate through TaxCoverageOrderRow into the coverage detail modals
## Problem / Context

Known, deliberately deferred gap from epic #2452's tracking PR #2668. Issue #2477's acceptance criteria require each `detail-tax`/`detail-novat`/`detail-postrollout` modal row to show its own resolved/candidate tax rate — never a single hardcoded rate (this is a regression guard against the mockup's earlier "hardcoded 23%" bug, since real orders in a mixed-rate order set carry different real per-product rates).

`TaxCoverageOrderRow` (`libs/core/src/orders/domain/types/coverage-detection.types.ts`) currently carries only `internalOrderId`, `sourceConnectionId`, `placedAt` — no rate field. `TaxCoverageDetectionService.classifyOne` (`libs/core/src/orders/application/services/tax-coverage-detection.service.ts`) already resolves a real per-line rate via `this.productsService.getEffectiveTaxRate(line.productId, line.variantId)` during A/B/C classification, but only reduces it to a `known`/`no-rate`/`not-checked` state (`taxRateState(rate)`) to pick the bucket — the resolved `StoredTaxRate` value itself is discarded in the loop and never attached to the row. `toRow()` builds `TaxCoverageOrderRow` from the candidate, not from anything computed in `classifyOne`. `TaxCoverageOrderDto` (`apps/api/src/analytics/http/dto/tax-coverage-orders-response.dto.ts`) mirrors the same minimal shape, so no rate crosses the HTTP boundary. `CoverageDetailDialog` (`apps/web/src/features/analytics/components/coverage-detail-dialog.tsx`) therefore renders no rate at all for tax-coverage rows — no wrong/hardcoded value, but also none of the informational value #2477 asks for.

## Proposed Solution

1. Extend `classifyOne` to also return the resolved rate observation(s) per unresolved line — an order can have multiple lines, and a bucket can legitimately contain a mix of known/unknown rates — rather than discarding the resolved `StoredTaxRate` after computing `taxRateState`. The natural shape is a per-order list of per-line rate observations (e.g. `{ productId, variantId, rate: StoredTaxRate | null }[]`) alongside the existing bucket decision.
2. Thread that through `TaxCoverageOrderRow` → `TaxCoverageOrderDto` → the `/analytics/tax-coverage/*` HTTP response, following the same additive-field pattern already used elsewhere in this coverage-detection vocabulary.
3. Render the resolved/candidate rate(s) on each row in `CoverageDetailDialog` for `detail-tax`/`detail-novat`/`detail-postrollout`. A mixed-rate order shows its per-line rates, never one order-level number.
4. Add/extend tests: classification unit tests asserting rate values (not just the boolean bucket outcome) are threaded correctly for multi-line, mixed-known/unknown orders; a modal test asserting per-row rate rendering rather than a single hardcoded value.

## Classification

**Type**: CORE + Frontend
**Layer**: Application (orders context) / Interface (HTTP DTO) / Frontend feature component
**File(s)**:
- `libs/core/src/orders/domain/types/coverage-detection.types.ts`
- `libs/core/src/orders/application/services/tax-coverage-detection.service.ts`
- `apps/api/src/analytics/http/dto/tax-coverage-orders-response.dto.ts`
- `apps/web/src/features/analytics/components/coverage-detail-dialog.tsx`

## Dependencies

- Builds on merged Phase 4 (#2465, PR #2493) and Phase 7 (#2474, PR #2672). No open blocking dependency.
- Stacks on top of PR #2785 (`2778-exchange-rate-endpoint`) in the epic #2452 branch chain.

## Assumptions

- The per-line rate observation shape is additive to `TaxCoverageOrderRow`/`TaxCoverageOrderDto` (new optional field), not a breaking change to the existing bucket-classification contract.
- `ProductsService.getEffectiveTaxRate` (or equivalent) remains the source of the rate value — this issue does not change how the rate is resolved, only whether the resolved value is kept and surfaced.

## Acceptance Criteria

- [ ] `classifyOne` / the classification pass returns the resolved rate per unresolved line instead of discarding it, for all three tax-coverage buckets.
- [ ] `TaxCoverageOrderRow` and `TaxCoverageOrderDto` carry the per-line rate data end to end to the HTTP response.
- [ ] `CoverageDetailDialog` renders each row's own resolved/candidate rate for `detail-tax`/`detail-novat`/`detail-postrollout` — never a single shared/hardcoded value across rows.
- [ ] A fixture with a multi-line, mixed-rate order renders each line's own rate correctly (regression guard for the mockup's "hardcoded 23%" bug).
- [ ] Tests added or updated for the classification pass and the modal rendering.

norbert-kulus-blockydevs:
Closes #2798

## Summary

`TaxCoverageDetectionService.classifyOne` already resolves a per-line tax rate via `IProductsService.getEffectiveTaxRate` while classifying an order into the tax-a/tax-b/tax-c coverage buckets, but it only reduced that resolved value to a `known`/`no-rate`/`not-checked` state to pick the bucket, then discarded the actual `StoredTaxRate`. `TaxCoverageOrderRow` / `TaxCoverageOrderDto` therefore never carried a rate, so `CoverageDetailDialog` rendered no rate at all for the `detail-tax` / `detail-novat` / `detail-postrollout` drill-downs — not wrong/hardcoded, but missing the informational value #2477 asked for.

- Added `TaxCoverageLineRateObservation` (`{ productId, variantId, rateCode, state }`) to `libs/core/src/orders/domain/types/coverage-detection.types.ts` and a `lineRates: TaxCoverageLineRateObservation[]` field on `TaxCoverageOrderRow`.
- `TaxCoverageDetectionService.classifyOne` now returns `{ category, lineRates }`; a new `resolveLineRates` helper builds one observation per order line (reusing the existing `getEffectiveTaxRate` read for unresolved lines, no new catalogue calls), and `toRow` threads it onto the row. A non-pre-rollout candidate keeps its documented "never touches line items or the catalogue" behavior and reports `lineRates: []`.
- `TaxCoverageOrderDto` / new `TaxCoverageLineRateDto` (`apps/api`) mirror the same shape end to end over `GET /analytics/coverage/tax/orders`.
- FE `TaxCoverageOrder` / new `TaxCoverageLineRate` type (`apps/web`) mirror the DTO; `TaxOrderRow` in `analytics-data-coverage-panel.tsx` renders each row's own deduplicated per-line rate tags (e.g. `23%`, `8%`, `no rate`, `not checked yet`) instead of nothing.

## Test plan

- [x] `libs/core` unit tests: extended `tax-coverage-detection.service.spec.ts` with a `classify — per-line rate observations (#2798)` block — multi-line mixed-known/unknown threading, no-rate/not-checked distinct observations, the non-pre-rollout empty-array/no-catalogue-call guard, and the catalogue-failure-as-not-checked case. All existing classification tests still pass unmodified.
- [x] `apps/api` unit tests: updated fixtures in `analytics-coverage.controller.spec.ts` / `analytics-tax-remediation.controller.spec.ts` for the new required `lineRates` field.
- [x] `apps/web` unit tests: added a dedicated `analytics-data-coverage-panel.test.tsx` case asserting a mixed-rate order renders each line's own rate (`23%`, `8%`, `5%`) with no single value duplicated across rows — the regression guard the issue calls for.
- [x] `pnpm --filter @openlinker/core lint/type-check/test`, `pnpm --filter @openlinker/api lint/type-check/test` (scoped), `pnpm --filter @openlinker/web lint/type-check/test` (scoped) all green.

## Notes

Stacks on `2778-exchange-rate-endpoint` (PR #2785) per the epic #2452 branch chain — this PR targets that branch, not `main`.


norbert-kulus-blockydevs:
## Tech Lead Review — PR #2802

**Scope**: `libs/core/src/orders/{application,domain}` (tax coverage detection service + types), `apps/api/src/analytics/http/dto` (DTO mirror), `apps/web/src/features/analytics` (FE types + panel rendering + tests).

### Summary

The PR threads the already-resolved per-line tax rate (previously computed and discarded by `TaxCoverageDetectionService.classifyOne`) through to the FE drill-down modal, so a mixed-rate order shows each line's own rate tag instead of nothing. The change is small, well-scoped, and closely follows existing precedent in the same files (`CurrencyOrderRow` / `.coverage-detail-row__trail` / `.coverage-detail-row__tag`, multi-class DTO files, `ApiPropertyOptional({nullable:true})` for nullable-but-required fields).

### Architecture & boundaries

- ✅ `orders → products` is an already-documented cross-context edge (`docs/architecture-overview.md § Cross-context dependencies in core`), so importing `TaxRateState`/`taxRateState` from `@openlinker/core/products` is legal.
- ✅ `TaxCoverageLineRateObservation` is exported through the `orders` top-level barrel (`libs/core/src/orders/index.ts`), not a deep path.
- ✅ No new I/O introduced: `resolveLineRates` reuses the *existing* `getEffectiveTaxRate` call site rather than adding a second catalogue read — confirmed by reading `classifyOne`, `resolveLineRates` is called exactly once per candidate (either in the "unresolved.length === 0" fast branch or the general branch), so per-line resolution isn't duplicated.
- ✅ The non-pre-rollout early-return path (`taxRateEra !== 'pre-rollout'`) still returns `lineRates: []` with no line-item read and no catalogue call, preserving the class's documented "never touches line items or the catalogue" contract for that branch — and this is explicitly regression-tested (`does not touch line items or the catalogue for a non-pre-rollout candidate`).

### Code quality

- ✅ Types live in `*.types.ts` (`coverage-detection.types.ts`), `as const` vocabulary reused (`TaxRateStateValues`) rather than re-derived — the JSDoc on `TaxCoverageLineRateObservation` explicitly calls out reusing `TaxRateState` so the row "can never disagree with the classification that produced it."
- ✅ No `any`. Explicit return types throughout the new `resolveLineRates` / `classifyOne` signatures.
- ✅ Error handling: a catalogue read failure inside `resolveLineRates` is logged and reported as `state: 'not-checked'` with `rateCode: null` — consistent with the existing "never fabricate a rate" invariant already documented on the type, and covered by its own test.
- ✅ DTO (`TaxCoverageLineRateDto`) follows the exact nullable-field decorator convention already used by `TaxCoverageOrderDto.placedAt` in the same file, and multiple DTO classes per file is an established pattern in this directory (`top-products-response.dto.ts`, `needs-attention-response.dto.ts`, `sales-analytics-response.dto.ts` all do the same).

### Frontend

- ✅ `TaxOrderRow` reuses the `.coverage-detail-row__trail` / `.coverage-detail-row__tag` / `.coverage-detail-row__tag--neutral` classes that already exist in `index.css` (from the sibling `CurrencyOrderRow` on the same base branch) — no missing CSS, no new tokens needed.
- ✅ `summarizeLineRates` dedupes labels so a large order with many identically-taxed lines doesn't repeat the tag — reasonable and tested (`23%` appears once across the whole rendered dialog even though it could recur across rows for different orders, and the dedup-within-a-row case is implicit in the Set usage).
- ✅ FE type mirror (`TaxCoverageLineRateState`) documents that it mirrors `TaxRateState` because the frontend cannot import core (#591) — consistent house style. Minor observation only (see below).

### Tests

- ✅ Core: 4 new spec cases covering multi-line mixed known/unknown, no-rate vs not-checked distinctness, the non-pre-rollout empty-array/no-catalogue-call guard, and the catalogue-failure-as-not-checked path.
- ✅ API: existing fixtures updated for the new required `lineRates` field (no silent breakage).
- ✅ FE: new dedicated test asserts a mixed-rate order renders each line's own tag and that a value is never duplicated as a single hardcoded rate across rows — this is exactly the regression class the PR description calls out.

### Minor / non-blocking observations (SUGGESTION)

1. **`apps/web/…/analytics-tax-coverage.types.ts`** — `TaxCoverageLineRateState` is a hand-maintained mirror of core's `TaxRateState` with no `check-*-mirror.mjs` invariant script backing it (unlike `check-shipping-tax-split-mirror.mjs`, `check-parameter-restriction-mirror.mjs`, etc. for other FE/BE mirrors). Given it's a 3-value closed union unlikely to drift, this is not blocking, but if `TaxRateStateValues` ever gains a member, nothing in `pnpm check:invariants` would catch the FE falling out of sync. Worth a one-line note in the type's JSDoc that it's uncontracted, or filing a follow-up if this pattern is expected to be revisited.
2. **`resolveLineRates` recomputes `resolveNetSalesTaxRate(line.taxRate)` a second time** inside the "every line already resolves" fast path (once via the `unresolvedLines` filter, once again inside `resolveLineRates`'s loop). Purely a pure-function double-call, not a behavior or correctness issue, and avoiding it would complicate the code for no real benefit — flagging only for completeness.

### Verdict

✅ **Approve** — no blocking or important issues found. The change is narrowly scoped, reuses existing patterns faithfully (CSS classes, DTO shape, cross-context import rules, error-handling philosophy), and ships proportionate test coverage across all three layers touched (core/api/web). The two notes above are optional polish, not merge blockers.

norbert-kulus-blockydevs:
Addressed both non-blocking review notes:

1. **`resolveLineRates` double evaluation** (`tax-coverage-detection.service.ts`) — `classifyOne`'s fast path used to compute `resolveNetSalesTaxRate` twice per line: once to build `unresolvedLines` (deciding whether to take the "every line already resolved" branch), and again inside `resolveLineRates`'s own per-line loop. Since `resolveLineRates` already skips the catalogue call for any line whose rate is already known, the pre-filter added no correctness value — it's now removed, and the "already resolved" outcome is derived directly from the returned `lineRates` observations (no `no-rate` / `not-checked` states present). Behavior is unchanged; all 16 existing/updated specs in `tax-coverage-detection.service.spec.ts` still pass.

2. **Uncontracted FE mirror** (`analytics-tax-coverage.types.ts`) — added a JSDoc note on `TaxCoverageLineRateState` stating explicitly that, unlike `check-shipping-tax-split-mirror.mjs` and its siblings, no invariant script backs this mirror against core's `TaxRateStateValues`, so a future core addition wouldn't be caught by `pnpm check:invariants`. Kept as documentation only per the reviewer's own "not blocking" framing, given the union is a closed 3-value set unlikely to drift.

Verified `pnpm --filter @openlinker/core lint/type-check/test` and `pnpm --filter @openlinker/web lint/type-check` all green on the touched files. Commit `e755f9480`.

piotrswierzy (approved):
# Code Review: thread resolved per-line tax rate into coverage detail modals (#2798)

**Scope**: 11 files, reviewed against this PR's own base (`2777-resolve-publication-day-before-fx-cache-read`), not `main`.

---

## 1. Summary

Small, well-shaped change: `TaxCoverageDetectionService.classifyOne` already resolved a per-line `StoredTaxRate` and threw it away after picking a bucket; this PR keeps it, threads it through `TaxCoverageOrderRow` → `TaxCoverageOrderDto` → the FE `TaxCoverageOrder`, and renders deduplicated per-line tags in `TaxOrderRow`. Layering is correct throughout (domain type in `orders/domain/types`, DTO in `apps/api/.../http/dto`, FE mirror type in the feature's `api/` folder), the `as const`/union vocabulary is **reused** from `@openlinker/core/products` rather than re-declared in core, and the type-only cross-context import keeps the existing `orders → products` edge type-erased.

**The refactor is behaviour-preserving where it matters, and I verified the two places it could have silently drifted:**

- **Catalogue call count is unchanged.** The old code filtered to `unresolvedLines` then called `getEffectiveTaxRate` per unresolved line; the new `resolveLineRates` iterates all lines but `continue`s on `kind === 'known'` before any read. Same reads, same order. The PR description's "no new catalogue calls" claim holds.
- **Bucket derivation is unchanged.** `anyConfirmedNoRate` / `anyNotChecked` are now derived from the full observation list instead of the unresolved subset, but a known line yields `state: 'known'`, which sets neither flag. The `no-rate`-beats-`not-checked` precedence and the early `tax-a` path are preserved, and the non-pre-rollout branch still returns before touching `findByOrderId` — asserted directly by the new spec.

Test coverage is genuinely good: the four new core cases cover mixed-known/unknown threading, the distinct `no-rate` vs `not-checked` observations, the "never reads line items or the catalogue" guard for a non-pre-rollout candidate, and the catalogue-failure-degrades-to-`not-checked` path (correctly asserting **no fabricated rate code**). The FE test's `expect(screen.queryAllByText('23%')).toHaveLength(1)` is the right shape of assertion for the regression the issue names — it fails on a hardcoded shared value, which a plain `toBeInTheDocument()` would not.

**Most important concern**: the FE `TaxCoverageLineRateState` union is an unguarded mirror of core's `TaxRateStateValues`, and the FE's fallback branch turns any drift into a *false operator-facing claim* rather than a visible gap.

---

## 2. Alignment with documentation

### Aligned
- **Cross-context contract** (`architecture-overview.md § Cross-context dependencies in core`): `coverage-detection.types.ts` imports only the `TaxRateState` **type alias** from the top-level `@openlinker/core/products` barrel — an allowed symbol shape, from an allowed specifier, and `import type` so it erases. The `orders → products` edge already exists in the map.
- **Union vocabulary** (`engineering-standards.md § Union Types: as const`): the PR reuses `TaxRateState` rather than declaring a fourth copy in core. The docblock's stated reason — "so the row can never disagree with the classification that produced it" — is the correct justification and is worth having in the file.
- **Type placement**: `TaxCoverageLineRateObservation` lives in `*.types.ts` beside the type it extends, and is exported from the context barrel (`libs/core/src/orders/index.ts`) as a type-only export.
- **FE layering** (`frontend-architecture.md`): transport type in `features/analytics/api/`, presentation in `features/analytics/components/`. No `fetch`, no page-level transport, no new cross-feature import. `coverage-detail-row__trail` / `__tag` / `__tag--neutral` all exist in `index.css` on this branch (I checked — they are contributed by an ancestor branch in the stack, not missing).
- **File headers** present on every touched source file; JSDoc explains *why* (the discarded-rate history, the empty-array-for-non-pre-rollout decision), not *what*.

### Deviations
None that violate a stated rule. The mirror gap below is a *precedent* gap, not a documented-rule violation — and the author has honestly flagged it in a comment rather than hiding it.

---

## 3. Findings

### 🟡 IMPORTANT

**1. The unguarded FE state mirror fails in the wrong direction** — `apps/web/src/features/analytics/api/analytics-tax-coverage.types.ts` (`TaxCoverageLineRateState`) + `analytics-data-coverage-panel.tsx` (`describeLineRate`)

The type's own docblock concedes it: *"no invariant script asserts this stays in sync with core's `TaxRateStateValues`."* That is true and, on its own, a tolerable trade for a three-value union. What makes it worth acting on is the **render fallback**:

```ts
if (observation.state === 'no-rate') return 'no rate';
return 'not checked yet';
```

A fourth state added to `TaxRateStateValues` later would not fail `pnpm check:invariants`, would not fail `tsc` (the API returns `string` at runtime regardless of the declared union), and would render as **"not checked yet"** — a positive, actionable claim about the operator's own catalogue that the backend never made. That is the same failure class this repo has repeatedly chosen to spend a `check-*-mirror.mjs` on: 21 such scripts exist, including `check-attention-reason-mirror.mjs` and `check-retry-refusal-reason-mirror.mjs`, both guarding unions of comparable size and comparable "display-only" stakes.

Two acceptable fixes, in order of preference:

1. Add `scripts/check-tax-rate-state-mirror.mjs` under `check:invariants`, comparing `TaxRateStateValues` against the FE union — matches the established precedent exactly, ~20 lines.
2. If you'd rather not add a 22nd script, make the fallback *honest*: switch exhaustively on the three known states and return a neutral `'unknown'` (or render nothing) for anything else, so drift degrades to a visible gap rather than a wrong sentence. A `satisfies`-style exhaustive switch also gives you a compile break when the FE union is widened.

Either is fine; what I would not merge long-term is the current combination of *unguarded mirror* + *fallback that asserts a specific state*.

### 🟢 SUGGESTIONS

**2. `rateCode` provenance differs between the two branches** — `tax-coverage-detection.service.ts` (`resolveLineRates`)

A line resolved from its own column reports `rateCode: line.taxRate` (the raw stored string); a line resolved from the catalogue reports `rate.code`. If the two ever differ in formatting for the same real rate (`'23.00'` vs `'23'`), the FE's label dedup keys on the rendered string, so one order would show both `23.00%` and `23%` as separate tags. Normalising both through the same accessor — or trimming — would close it cheaply. Not a correctness bug today, just a latent cosmetic one.

**3. The `no-rate` observation drops `unknownReason`** — `coverage-detection.types.ts`

`StoredTaxRate.unknownReason` (#2264) exists precisely as provenance for an operator surface, and the tax-coverage modal is that kind of surface. The row currently renders a bare `no rate` where the shop may have said *why*. Carrying it through (nullable, display-only, never control flow — same posture #2264 gives it) would make the drill-down more actionable. Reasonable to defer to a follow-up.

**4. Swagger nullability is inconsistent inside one DTO file** — `tax-coverage-orders-response.dto.ts`

`variantId` and `rateCode` are **required-but-nullable** fields yet use `@ApiPropertyOptional({ nullable: true })`, which documents them as *absent-able*. The sibling `placedAt!: Date | null` in the same file uses `@ApiProperty({ ..., nullable: true })` — the accurate spelling for this shape. Worth aligning so the generated contract matches the TS type.

**5. Payload size is unbounded by line count** — `TaxCoverageOrderDto.lineRates`

Every line of every order on a page is serialised, while the FE immediately collapses them to a handful of distinct labels. A 200-line order × a page of 10 is 2 000 objects to render ~3 tags. Not a problem at this persona's volumes (ADR-039 puts it at 10–100 orders/day), and keeping the raw observations is the right call for a future per-line drill-down — but if the modal never grows one, deduplicating server-side is the cheaper contract.

### ✅ Positive observations

- The catalogue-failure path is handled exactly right: `not-checked` with `rateCode: null` and a warn log, explicitly **not** `no-rate` — the docblock states why (a permanent unresolvable claim the order does not support), and the spec name says so too.
- `classifyOne`'s new docblock explains the `lineRates: []` decision for a non-pre-rollout candidate as a deliberate cost choice rather than an oversight, and the spec pins it with `expect(findByOrderId).not.toHaveBeenCalled()`. That is the shape that stops a later refactor from "helpfully" adding the read back.
- `orderId` is threaded into `resolveLineRates` purely to keep the existing warn-log context intact — easy thing to lose in an extract-method, and it wasn't lost.

---

## 4. Final assessment

### Merge readiness
🔄 **Approve with changes**

The core/API/FE threading is correct, the refactor is provably behaviour-preserving on both axes it could have broken, and the tests assert the specific regression the issue was filed about. Nothing here blocks merge.

### Before/soon after merge
1. **(IMPORTANT)** Close the `TaxCoverageLineRateState` mirror gap — either a `check-tax-rate-state-mirror.mjs` under `check:invariants`, or an exhaustive switch whose fallback stops asserting `not checked yet`. Today an unknown state silently becomes a wrong operator-facing sentence.
2. *(suggestion)* Normalise `rateCode` across the two resolution branches so label dedup can't split one rate into two tags.
3. *(suggestion)* Align `variantId` / `rateCode` to `@ApiProperty({ nullable: true })`, matching `placedAt` in the same file.
4. *(suggestion)* Consider carrying `unknownReason` onto the `no-rate` observation — that's what #2264 minted it for.

Note for the merge queue: `mergeable_state: blocked` and this PR targets `2777-resolve-publication-day-before-fx-cache-read`, so it lands only behind its stack.

norbert-kulus-blockydevs:
Addressed all findings from the review in 287c3f6ce.

**1. IMPORTANT — unguarded FE state mirror + dishonest fallback**
Added `scripts/check-tax-rate-state-mirror.mjs` under `pnpm check:invariants`, comparing core's `TaxRateStateValues` against the FE `TaxCoverageLineRateState` union byte-for-byte (matches the 21 existing `check-*-mirror.mjs` scripts exactly). On top of that, `describeLineRate` now switches exhaustively over `state` with a neutral `'unknown'` fallback instead of falling through to `'not checked yet'` — so even before the invariant script would catch a drift, an unrecognised value renders as an honest gap rather than a false positive claim.

**2. `rateCode` provenance differs between the two branches**
Added a `normalizeRateCode` helper in `TaxCoverageDetectionService` and routed both resolution branches (the line's own `taxRate` column, and the live catalogue read) through it — a numeric code is canonicalized via `Number()`-round-trip (`'23.00'` → `'23'`), an exemption code passes through verbatim. Covered by a new spec case.

**3. `no-rate` observation drops `unknownReason`**
`TaxCoverageLineRateObservation` now carries `unknownReason` (#2264), populated from `StoredTaxRate.unknownReason` only on the `'no-rate'` branch (`null` everywhere else — mirroring the field's own optionality precedent). Threaded through `TaxCoverageLineRateDto` and the FE `TaxCoverageLineRate` type, and rendered as a `title` tooltip on the "no rate" tag (never as a second tag, so the tag list stays the same length whether or not the catalogue gave a reason). New spec cases on both the core service and the FE panel.

**4. Swagger nullability inconsistent in one DTO file**
`variantId`, `rateCode`, and `placedAt` are required-but-nullable fields that are always present on the wire — switched all three from `@ApiPropertyOptional({ nullable: true })` to `@ApiProperty({ nullable: true })`, which is the accurate spelling for "always present, sometimes null" vs. `ApiPropertyOptional`'s "may be absent."

**5. Payload size unbounded by line count**
Left as-is per your own note — not a problem at this persona's volumes, and server-side dedup would cost the future per-line drill-down the raw observations are kept for. Happy to revisit if the modal never grows one.

**Test plan**
- `libs/core`: `tax-coverage-detection.service.spec.ts` extended with cases for `unknownReason` propagation and `rateCode` normalization — 18/18 pass.
- `apps/api`: `analytics-coverage.controller.spec.ts` / `analytics-tax-remediation.controller.spec.ts` pass unmodified — 13/13.
- `apps/web`: `analytics-data-coverage-panel.test.tsx` extended with an `unknownReason`-tooltip case — 10/10 pass.
- `pnpm --filter @openlinker/core lint/type-check`, `pnpm --filter @openlinker/api lint` + `tsc -p tsconfig.type-check.json` (after building libs), `pnpm --filter @openlinker/web lint/type-check` all green.
- `node scripts/check-tax-rate-state-mirror.mjs --self-check && node scripts/check-tax-rate-state-mirror.mjs` pass.

Issue #1807, opened 2026-07-23
[SPIKE] Integration/Frontend — Confirm and fix ShipX rejection of Allegro paczkomat labels


norbert-kulus-blockydevs:
> Sub-issue of #1803. This is Part B (the spike + targeted fix). It confirms and closes the underlying ShipX rejection for Allegro paczkomat labels. Gated on the field-level errors surfaced by the Part A diagnostics bug (or the existing backend WARN log).

## Problem / Context

Generating an InPost label for a **paczkomat** shipment sourced from an Allegro order ("Allegro Paczkomaty InPost") is rejected by ShipX with `validation_failed`. The exact field is not yet confirmed - the leading hypothesis is the **receiver** on the locker path:

- For a paczkomat the mapper omits the receiver address (correct) but forwards `company_name = recipient.name`, `first_name`, `last_name`, `email`, `phone` - `libs/integrations/inpost/src/infrastructure/mappers/inpost-shipx.mapper.ts:446-459`.
- The FE builder never populates `recipient.name`, and the FE preflight `detectMissingFields` requires only `customerEmail` + `phone` for lockers - `apps/web/src/features/orders/lib/dispatch-input.ts:65-87`; it never checks `firstName`/`lastName`.
- Allegro buyer snapshots frequently lack a parsed first/last name, so an order passes OL's FE gate and is then rejected by ShipX on `receiver`.

This is the strongest candidate but is **not proven** - ShipX could equally be rejecting `parcels.template`, `custom_attributes.target_point`, or a sender field. The spike confirms before fixing.

## Proposed Solution

**Spike:** Using the now-visible ShipX `details` (Part A) or the backend WARN log at `libs/core/src/shipping/application/services/shipment-dispatch.service.ts:352`, confirm which ShipX field is rejected for an Allegro paczkomat order. Record the finding in this issue.

**Targeted fix (shape depends on the spike):** If it is the receiver (most likely), fix the FE preflight (`detectMissingFields`) and/or the mapper (`toReceiverPeer`) so a locker shipment always sends an identifiable receiver that ShipX accepts - e.g. derive `company_name`/`first_name`/`last_name` from the Allegro snapshot buyer name, or gate submission with a clear operator-facing message when the name is genuinely absent. Follow the existing preflight-guard pattern at `inpost-shipx.mapper.ts:298-312`.

## Classification

**Type**: Integration + Frontend
**Layer**: Infrastructure (InPost mapper) + Application (FE preflight)
[...]

Issue #885, opened 2026-05-28
[BE] refactor(shipping): migrate adapters to throw ShippingProviderRejectionException (#881 follow-up)


piotrswierzy:
## Context

PR #881 introduced `ShippingProviderRejectionException` (`libs/core/src/shipping/domain/exceptions/`) and updated `ShipmentController.toHttpException` to:

- Map the typed exception to `502 Bad Gateway`.
- Map any other untyped `Error` to `500 Internal Server Error` (logged with stack for triage).

This is the additive half of the split. **The matching adapter migration is still pending** — adapters today still throw bare `Error` / `HttpException`-shaped objects for provider-side rejections, so until they migrate, the `502 → 500` split doesn't take full effect: real carrier-rejection cases get mis-attributed to 500 instead of 502.

## Scope

Migrate the shipping adapters to throw `ShippingProviderRejectionException` from the well-known provider-rejection branches. Today's adapters:

- `AllegroDeliveryShippingAdapter` (`libs/integrations/allegro/src/infrastructure/adapters/`)
- `InpostShippingAdapter` (`libs/integrations/inpost/src/infrastructure/adapters/`)
- (Future) any `*ShippingProviderManager`-implementing adapter.

### Per-adapter changes

1. **Identify provider-rejection branches** — these are the catch blocks that wrap HTTP 4xx responses from the carrier API. They are already distinguishable from network errors / DB drops / configuration bugs in the existing code.
2. Replace the bare `throw new Error(...)` / `throw new HttpException(...)` with `throw new ShippingProviderRejectionException(providerName, providerCode, message)`:
   - `providerName`: `'allegro'` / `'inpost'`, etc.
   - `providerCode`: the carrier-side error code if surfaced (`'PARCEL_TOO_LARGE'`, `'INVALID_PARCEL_TEMPLATE'`, etc.); `null` when the provider returned only a message.
   - `message`: the carrier's human-readable rejection reason.
3. Network / DNS / timeout errors stay as bare `Error` — these are NOT provider rejections, they're transport failures (the controller correctly 500s these).

### Test updates

- Each adapter's `__tests__/*.spec.ts` updated to assert the typed exception on the rejection branches it already covers.
- `ShipmentController.toHttpException` spec already covers the 502-via-typed-exception path (added in #881); no change there.

## Acceptance criteria

[...]

Issue #1777, opened 2026-07-22
[BUG] DPD create rejections (NOT_PROCESSED) are undiagnosable - raw response body never logged


norbert-kulus-blockydevs:
## Problem / Context

When DPD rejects a shipment create, the operator sees only `NOT_PROCESSED` with `providerCode: null` (API 502 body + `shipments.errorMessage`). There is no field-level reason, and there is no way to recover the real cause without out-of-band access to DPD.

This was hit live: a DPD connection with an out-of-region sender postal code caused every `generatePackagesNumbers` to fail with a bare `NOT_PROCESSED`. Diagnosing it required replaying the request against DPD outside OL to see the real reason - OL surfaced and logged nothing useful.

## Root cause

DPD returns business rejections as **HTTP 200** with a non-OK body `status` (e.g. `NOT_PROCESSED`). The HTTP client only logs/raises on non-2xx responses (`libs/integrations/dpd-polska/src/infrastructure/http/dpd-http-client.ts`), so a 200-with-rejection is never logged - the raw response body (which may carry `packages[].validationInfo[].errorCode` / `.info`) is discarded.

`assertCreateSucceededAndExtractWaybill` (`libs/integrations/dpd-polska/src/infrastructure/mappers/dpd-shipment.mapper.ts`) does collect `validationInfo` when present, but DPD sometimes returns an entry with only `info: "NOT_PROCESSED"` and no `errorCode`, so the surfaced message is the opaque batch status with `providerCode: null`. In that case the only diagnostic lever - the raw body - is gone.

## Proposed Solution

On any create response whose top-level/package/parcel `status !== 'OK'`, log the **full raw DPD response body** at `warn`/`debug` (redacting nothing sensitive - the create response carries no secrets), keyed by `traceId`, before throwing `ShippingProviderRejectionException`. Optionally include DPD's `traceId` in the exception `providerDetails` so it can be quoted to DPD support.

This keeps the neutral operator-facing message as-is but makes opaque `NOT_PROCESSED` rejections diagnosable from OL logs alone.

## Classification

- **Type**: Bug (observability) - Integration (DPD)
- **File(s)**: `libs/integrations/dpd-polska/src/infrastructure/http/dpd-http-client.ts`, `libs/integrations/dpd-polska/src/infrastructure/mappers/dpd-shipment.mapper.ts`

## Acceptance Criteria

- [ ] A DPD create that returns a non-OK body status logs the full raw response body (with `traceId`) at warn/debug
[...]

PR #2805, opened 2026-09-02
fix(integrations-inpost): fall back to ShipX error code as providerCode when details is empty

### LINKED ISSUE #2804: [BUG] Integration — InPost label-fetch failure surfaces no usable detail when ShipX returns an empty error body
## Problem / Context

`GET /v1/shipments/:id/label` calls InPost ShipX **live** on every request (no caching) via `ShipmentLabelService.fetchLabel` → `InpostShippingAdapter.fetchLabel` (`libs/integrations/inpost/src/infrastructure/adapters/inpost-shipping.adapter.ts`). Live-verified on `demo.openlinker.io` (2026-09-02, order `ol_order_911c5321d6864a2f9a1ac61aa0824dce`): a shipment generated several weeks earlier returned a 4xx from ShipX with **no `details` object**, and the operator-facing error was:

```
"There was a problem with label generation. Check details object for more info."
```

...with `providerCode: null`. That message is not something an operator can act on — it literally instructs them to check a `details` object that doesn't exist in the response.

**Root cause**: `InpostHttpClient.throwIfNotOk` (`libs/integrations/inpost/src/infrastructure/http/inpost-http-client.ts:173`) builds the exception message from `errorBody?.message` and derives `providerCode` only via `firstDetailKey(flattenShipXFieldErrors(errorBody?.details))`. When ShipX's body carries no `details` at all, `flattenShipXFieldErrors` returns `{}`, so `providerCode` ends up `null` and the raw carrier `message` string is the only thing surfaced — verbatim, with no added context.

**Key finding — an existing, unused signal**: `ShipXErrorBody` (`libs/integrations/inpost/src/domain/types/inpost-shipx.types.ts:162-167`) already declares an `error?: string` field — ShipX's own short error code (the existing test fixture at `inpost-http-client.spec.ts:153` shows `"error":"validation_failed"` alongside a *different* generic message, `"There are some validation errors. Check details object for more info."` — note ShipX has at least two distinct canned generic-message templates). **`errorBody.error` is never read anywhere in `InpostHttpClient`.** Reading and classifying on it is the most direct fix — it is the one place ShipX might actually be telling us *what kind* of rejection this is, and today that signal is silently discarded.

**Note on root cause of the underlying trigger** (label unfetchable after several weeks): no InPost/ShipX documentation was found stating a sandbox test-data retention or purge policy (checked the developer docs FAQ and sandbox-environment pages at `dokumentacja-inpost.atlassian.net`) — only indirect community reports of general sandbox flakiness (intermittent "Resource not found" on older resources, empty bodies instead of proper error payloads). So this issue does **not** claim a confirmed "N-week purge policy" as the cause. The observed fact is simply: the carrier's sandbox environment stopped serving the label for a shipment several weeks old, with no clear underlying reason documented by InPost. The fix here stands on its own merit — better classification of whatever 4xx shape ShipX actually returns — not on a proven root cause for *why* ShipX returns it.

## Proposed Solution

Extend `InpostHttpClient.throwIfNotOk` (or a label-fetch-specific variant used only by `requestBinary` calls) to read `errorBody?.error` in addition to the existing `message`/`details` handling, and use it to produce a more specific, classified message/`providerCode` when `details` is empty — instead of silently falling through to the raw carrier `message` with `providerCode: null`.

Concretely:
- Capture the real `error` values ShipX returns for (a) a genuinely unfetchable/expired label and (b) an ordinary field-validation rejection (the existing `validation_failed` fixture is one data point; a live sandbox probe against an old/purged shipment would confirm the "gone" case's actual `error` value).
- Where `errorBody.error` is present and `details` is empty, fold it into the thrown `ShippingProviderRejectionException`'s `providerCode` (today `firstDetailKey` returns `null` in this case — it should fall back to `errorBody.error` when no field-level detail exists) so the operator-facing 502 body at least carries a stable, non-generic code instead of `providerCode: null`.
- Extend `libs/integrations/inpost/src/infrastructure/http/__tests__/inpost-http-client.spec.ts` with a case asserting this fallback fires when `details` is empty/absent but `error` is present, alongside the existing `does_not_exist`/`target_point` case (which must keep behaving exactly as today — this is a fallback for the *no-details* case only, not a change to the existing field-error path).

Existing pattern to follow: `firstDetailKey` (`inpost-http-client.ts:262`) already documents itself as "the closest thing ShipX surfaces to a typed error code is the field that triggered the rejection" — this issue extends that same reasoning one level up, to the response's own `error` field, for the case where no field is identifiable at all.

## Classification

**Type**: Integration
**Layer**: Infrastructure (HTTP client / adapter boundary)
**File(s)**:
- `libs/integrations/inpost/src/infrastructure/http/inpost-http-client.ts`
- `libs/integrations/inpost/src/infrastructure/http/__tests__/inpost-http-client.spec.ts`
- `libs/integrations/inpost/src/domain/types/inpost-shipx.types.ts` (no change expected — `error` field already modeled)

## Dependencies

- None. Self-contained within the InPost integration package. Does not require the `apps/api` controller mapping or the frontend error-taxonomy (`apps/web/src/features/shipments/lib/label-download-error.ts`) to change — `ShippingProviderRejectionException`'s existing `providerCode` field is already surfaced end-to-end through `toHttpException` (`apps/api/src/shipping/http/shipment.controller.ts:533-544`) and rendered by the FE's existing branch-5 handling in `label-download-error.ts`, so a non-null `providerCode` alone already improves what the operator sees with no other-layer change required.

## Assumptions

- The exact `error` code ShipX returns for a purged/unfetchable label is not yet confirmed — implementation should start with a short live-sandbox probe (reproduce against an old/purged test shipment) to pin the real value before finalizing which values trigger the new fallback classification. If no stable `error` code is found for this specific case, the fallback should still improve the general "empty `details`" case broadly (any `error` value beats `providerCode: null`), rather than blocking on a single confirmed reproduction.
- This issue is intentionally scoped narrowly to error-message/classification clarity. A separate, larger design — a carrier-neutral "confirmed permanently unavailable" exception plus an audited operator-facing recovery ("supersede and regenerate") capability — was drafted but explicitly deferred by the requester and is **not** part of this issue's scope. If picked up later, that design exists as a draft plan on a local, unpushed worktree branch and can be resurfaced on request.

## Acceptance Criteria

- [ ] `InpostHttpClient` reads `errorBody?.error` and uses it as a `providerCode` fallback when the flattened `details` map is empty, instead of leaving `providerCode: null`.
- [ ] The existing field-error classification path (`firstDetailKey` / `flattenShipXFieldErrors`, e.g. the `target_point`/`does_not_exist` case) is unchanged — this is an additive fallback for the no-`details` case only.
- [ ] `inpost-http-client.spec.ts` has a passing test asserting the new fallback fires for a 4xx body with `error` set and empty/absent `details`.
- [ ] A shipment whose label-fetch hits this fallback surfaces a `providerCode` other than `null` at the `GET /v1/shipments/:id/label` HTTP boundary (verified via existing `toHttpException` behavior — no controller change required).
- [ ] No architecture boundary violations (CORE ↔ Integration) — the classification stays entirely inside the InPost integration package; no core-level exception change is required for this issue's scope.

norbert-kulus-blockydevs:
## Summary

InPost ShipX's label-fetch failures surfaced no usable detail when the carrier's 4xx body carried no `details` object - `providerCode` stayed `null` and the operator saw only ShipX's generic prose message ("There was a problem with label generation. Check details object for more info."). Live-verified on `demo.openlinker.io`.

`ShipXErrorBody` already models an `error?: string` field (ShipX's own short error code) but `InpostHttpClient.throwIfNotOk` never read it - only `message`/`details`. `providerCode` now falls back to `errorBody.error` when the field-level `details` map is empty, so the operator-facing rejection carries a stable, non-null code instead of `null`.

The existing field-error classification path (`firstDetailKey`/`flattenShipXFieldErrors`, e.g. the `target_point`/`does_not_exist` case from #1807) is unchanged - this is an additive fallback for the no-`details` case only.

## Test plan

- [x] `pnpm --filter @openlinker/integrations-inpost type-check` passes
- [x] `pnpm --filter @openlinker/integrations-inpost lint` passes
- [x] New unit tests: fallback fires when `error` is present and `details` empty; `providerCode` stays `null` when neither is present
- [x] Existing #885/#1807 field-error tests pass unchanged (17/17 in the file)

Closes #2804


norbert-kulus-blockydevs:
Checked for review findings (inline comments, issue comments, and PR reviews) - none exist on this PR yet, so there is nothing to fix at this time.

norbert-kulus-blockydevs:
# Review: fall back to ShipX error code as `providerCode` when `details` is empty

Scope: `libs/integrations/inpost/src/infrastructure/http/inpost-http-client.ts` (+8/-1) and its spec (+46). Line numbers below are on the PR head. Baseline read for comparison: the sibling `dpd-polska` rejection path, the `providerCode` contract in `libs/core/src/shipping/domain/exceptions/shipping-provider-rejection.exception.ts`, and every downstream consumer of `providerCode` (controller 502 body, `shipments.providerCode` column, `label-download-error.ts`, `group-failed-shipments-by-cause.ts`, `shipment-retryability.ts`).

---

**[IMPORTANT]** - `libs/integrations/inpost/src/infrastructure/http/inpost-http-client.ts:202` - the new read of `errorBody.error` is the only value taken from the untrusted body without a guard.

`safeParseError` (:231-242) ends in a blind `data as ShipXErrorBody`, so nothing about that object is validated. `flattenShipXFieldErrors` (:304-333) guards every value it touches (`Array.isArray`, `typeof item === 'string'`) precisely because of that cast. The new fallback hands `errorBody?.error` straight into a parameter typed `string | null`.

The empty-string case is the concrete one, and `??` does not catch it. `providerCode: ''` is persisted into the `shipments.providerCode` text column, and then in `apps/web/src/features/shipments/lib/group-failed-shipments-by-cause.ts:103` the `MIN_CAUSE_KEY_LENGTH` guard is skipped, because that guard only applies when `providerCode === null`. The result is a triage group keyed on an empty string which `shipment-triage-strip.tsx:96-101` labels a "rejection code" and renders as an empty span - that module's own comment calls a `providerCode`-keyed group "an HONEST shared-cause claim", and here it would be built from nothing. A non-string `error` (number, object) is less likely but equally untyped.

Fix, one line, keeps the read in the same place:

```ts
const shipxErrorCode =
  typeof errorBody?.error === 'string' && errorBody.error.trim().length > 0
    ? errorBody.error.trim()
    : null;
const providerCode = firstDetailKey(flatDetails) ?? shipxErrorCode;
```

**[IMPORTANT]** - `inpost-http-client.ts:196-208` - when the fallback finds nothing, the operator ends this PR exactly where #2804 found them, and the in-tree precedent for that case was not followed.

The sibling DPD adapter hit the identical shape and its comment names it almost verbatim: "sometimes only `info: NOT_PROCESSED` with no field-level `errorCode`, so the operator otherwise sees `NOT_PROCESSED` / `providerCode: null` with no recoverable cause" (`libs/integrations/dpd-polska/src/infrastructure/adapters/dpd-shipping.adapter.ts:160-205`). Its answer to that, in #1777, was to log the raw response envelope at warn keyed by the carrier's trace id before rethrowing.

InPost logs nothing on a 4xx. The only log in this client is the retry warn at :105, and on the API side the `ShippingProviderRejectionException` branch of `toHttpException` (`apps/api/src/shipping/http/shipment.controller.ts:532-544`) does not log either - only the unclassified-500 branch does. So for the case the second new test deliberately pins (neither `details` nor `error` present, `providerCode: null`), nothing anywhere records what ShipX actually said.

Suggest a `this.logger.warn` in the empty-`details` branch carrying status, method, path and the resolved `error` code. If you follow DPD and log the envelope itself, scope it the way DPD scoped its: WARN only, never into `providerDetails`, since `details` values can echo a rejected address fragment and `providerDetails` is contractually forwarded into the 502 body and stringified into logs (see the data-exposure paragraph in `shipping-provider-rejection.exception.ts`).

**[IMPORTANT]** - `__tests__/inpost-http-client.spec.ts:146-150` - the fixture is presented as a live reproduction, and I could not confirm that from the change. Flagging as a question, not an assertion.

The comment reads "Live-reproduced shape for an unfetchable label document: ShipX answers with an `error` code but no `details` object at all". #2804 says the opposite about that specific value: "The exact `error` code ShipX returns for a purged/unfetchable label is not yet confirmed - implementation should start with a short live-sandbox probe (reproduce against an old/purged test shipment) to pin the real value before finalizing". `not_found` appears nowhere else in the package, so there is no in-tree corroboration either.

This matters beyond comment accuracy. The observation in #2804 (`providerCode: null` plus the generic "Check details object for more info" message) is equally consistent with a body that carries no `error` field at all. In that case this fix changes nothing for the shipment that was reported, and #2804's fourth acceptance criterion ("a shipment whose label-fetch hits this fallback surfaces a `providerCode` other than `null`") is not met. Either cite the probe in the comment the way the #1807 comments in this same file do (date plus shipment id), or soften it to say the shape is modelled on ShipX's documented envelope rather than observed.

**[SUGGESTION]** - `__tests__/inpost-http-client.spec.ts:145-189` - both new tests call `client.request`, while the defect is on `requestBinary`.

`InpostShippingAdapter.fetchLabel` (`inpost-shipping.adapter.ts:151-161`) goes through `requestBinary`. Coverage is transitively sound, since `throwIfNotOk` is shared and the existing "should parse the JSON error envelope (NOT bytes) on a non-ok response" test (:321) already proves the binary path routes non-ok bodies through it, so this is not a hole. The mismatch is that the first new test's comment names an unfetchable label document and its path is `/v1/shipments/1/label`, then it calls the JSON method. Moving it into the existing `requestBinary` describe block, or switching the call, costs one line and makes the stated subject and the exercised path agree.

**[SUGGESTION]** - `inpost-http-client.ts:196-202` and the `firstDetailKey` docblock at :256-267 - two classifier tiers now feed one field, and nothing says so.

`firstDetailKey` returns a ShipX **field name** (`target_point`, `sender`, `parcels`); the fallback returns a ShipX **error code** (`validation_failed`, `not_found`). A downstream consumer matching on `providerCode` cannot tell which tier answered. No rule is broken: `shipping-provider-rejection.exception.ts` documents carrier-surfaced codes as carried verbatim, and `KnownProviderRejectionCodeValues` deliberately excludes them, so a bare code is the convention here even though DPD namespaces its own (`DPD.WAYBILL_EXPIRED`). One sentence in the docblock saying the field can carry either a field name or a ShipX error code would save the next reader the trace.

**[SUGGESTION]** - `libs/integrations/inpost/src/index.ts:16-19` - doc gap. The barrel comment enumerates what this package emits as `providerCode: 'target_point'` / `'preflight.*'`. A ShipX-surfaced `error` code is now a third family. One line.

**[SUGGESTION]** - follow-up worth naming somewhere. The code this fallback surfaces is exactly the signal the frontend would need to stop saying "The carrier rejected the request" for a label the carrier no longer holds. Today `label-download-error.ts:155-165` renders the code only as `(ref: …)`, and `deriveRetryabilityClass` still returns `unknown` for a bare carrier code because no family convention matches. #2804 scoped the frontend out correctly, but nothing tracks the follow-up, so the unlocked signal will sit unused.

---

## Positive observations

- The "existing path is unchanged" claim in the PR body holds under inspection, not just by assertion. `firstDetailKey` still wins whenever `flatDetails` is non-empty, and `providerDetails` is still gated on the same `Object.keys(flatDetails).length > 0`, so the #885 `target_point` re-tag in `inpost-shipping.adapter.ts:93` and the #1807 flattening behave identically. A body carrying both `error` and `details` still classifies on the field, which is the right precedence.
- The fallback degrades honestly downstream rather than selecting misleading copy. `label-download-error.ts:162` already branched on `code ? … : message`, so a new non-null code is appended as a support reference and no branch changes meaning; `deriveRetryabilityClass` maps an unmatched bare code to `unknown`, the same class `null` produced, so no false retryability claim is introduced. Worth stating because a fallback that silently moved a shipment into a different retryability bucket would have been the real risk here.
- Correctly contained: no core exception change, no controller change, no frontend change. The CORE/Integration boundary holds and `providerCode`'s open-runtime contract is used as documented.
- The second test pins the neither-present case explicitly instead of leaving `null` implicit. That is the test that makes the observability gap above visible, which is a point in its favour.

## Merge readiness

**Approve with changes.** No blocking issue. The change is narrow, the regression risk on the working path is genuinely nil, and the downstream analysis comes out clean.

## Priority fixes

1. Guard the new read: reject a non-string and an empty/whitespace `error` rather than passing it through as a code (finding 1).
2. Confirm or soften the "live-reproduced" claim on the new fixture, and say whether the reported shipment's body actually carried an `error` field - that determines whether #2804's fourth acceptance criterion is met (finding 3).
3. Add the warn log for the empty-`details` branch, following the DPD precedent and its redaction scoping (finding 2).

norbert-kulus-blockydevs:
Addressed the review in fd0c0c8ae.

**IMPORTANT — untrusted `errorBody.error` read**: fixed. `providerCode` now falls back only when `errorBody.error` is a non-empty (post-trim) string; a non-string or blank value is treated as absent (`providerCode: null`) instead of being written verbatim. Added a test pinning the blank-string case.

**IMPORTANT — no observability when the fallback answers nothing (or anything)**: fixed, following the DPD `dpd-shipping.adapter.ts` precedent. `throwIfNotOk` now warn-logs (status, method, path, resolved `providerCode`, message) whenever `flatDetails` is empty — covers both "fallback resolved a code" and "still null" — and never logs `errorBody.details` itself (kept out of the log line, same as it's kept out of `providerDetails`).

**IMPORTANT — unverified "live-reproduced" claim on the fixture**: softened. The comment now says the shape is modelled on ShipX's documented error envelope and that the exact `error` code for this situation hasn't been live-confirmed, rather than asserting a reproduction that didn't happen. Correct to flag — I can't confirm the reported shipment's body carried an `error` field, so #2804's 4th acceptance criterion isn't provably met yet; that needs a live-sandbox probe against a purged shipment as a follow-up.

**SUGGESTION — test exercised `request` while the defect is on `requestBinary`**: fixed, moved that test to call `client.requestBinary(...)` (the path `InpostShippingAdapter.fetchLabel` actually uses).

**SUGGESTION — two providerCode tiers, undocumented**: fixed. Added a note on `firstDetailKey`'s docblock that `providerCode` can carry either a ShipX field name or a ShipX error code depending on which classifier answered.

**SUGGESTION — barrel doc gap**: fixed, one line added to `index.ts` naming the third `providerCode` family.

**SUGGESTION — frontend follow-up not tracked**: left as-is per your framing (not asking to open an issue without confirming first); happy to file one if you'd like.

All existing + new unit tests pass (20/20 total in the file), plus lint/type-check for the package and the full pre-commit gate (lint + type-check + affected tests across the repo).

piotrswierzy (approved):
## Tech Lead Review — PR #2805

### Step 1 — Context

Three files, all inside `libs/integrations/inpost` (Integration layer only — no CORE change, no port/contract change, no schema change, so `docs/migrations.md` and the CORE↔Integration boundary rules are non-applicable):

- `infrastructure/http/inpost-http-client.ts` — `throwIfNotOk` now falls back to `errorBody.error` for `providerCode` when the flattened `details` map is empty, plus a `warn` log on the no-details branch.
- `__tests__/inpost-http-client.spec.ts` — 3 new cases.
- `index.ts` — barrel doc comment updated.

Verified against the repo:

- `domain/types/inpost-shipx.types.ts:153-164` really does model `error?: string` and documents it as "the machine key (`validation_failed`, `unauthorized`, …)" — so the field was modelled and unread, exactly as the PR says. ✅
- Precedence is right: `firstDetailKey(flatDetails) ?? shipxErrorCode`, so the #885/#1807 field-error path (`target_point`, nested `custom_attributes.target_point`) is untouched and the fallback is genuinely additive. ✅
- `providerDetails` still keyed on `Object.keys(flatDetails).length > 0`, so a fallback-only rejection carries `undefined` details rather than a fabricated map. ✅
- **The retryability blast radius is safe, and this is the thing that most needed checking.** `providerCode` feeds `libs/core/src/shipping/domain/provider-code-retryability.ts`, which drives operator-facing retry guidance. A ShipX code like `not_found` / `validation_failed` matches none of `api.http-NNN`, `preflight.*`, `command.*`, `target_point`, so it falls to `'unknown'` — the *same* class `null` already returned. No shipment changes retryability classification because of this PR. That function's deliberate "an opaque carrier-surfaced code is left `'unknown'` rather than guessed at" posture is what makes the fallback safe to add without touching core. ✅

### Step 2/3 — Findings

**No BLOCKING issues.**

**[IMPORTANT]** — `apps/web/src/features/shipments/components/shipment-triage-strip.tsx` (consequence of this change, not of that file)

> The triage strip branches on `group.providerCode !== null`, and the two branches are **not** informationally equivalent. Today a no-`details` InPost failure has `providerCode: null`, so the strip renders the *carrier message* branch: it says "report the same carrier message", prints the raw `sampleReason` inline, and — per that file's own header — deliberately stays "an OBSERVATION, never a causal claim". After this PR the same failures acquire a code, so the strip flips to the "report the same **rejection code**" branch, prints `validation_failed`, and **stops rendering `sampleReason` at all**.
>
> For a coarse ShipX code that is a net loss for the exact case this PR targets: `validation_failed` with no `details` is precisely the shape the header warns about ("a bad postcode, a missing parcel template and an over-limit COD sitting under one string"), but the copy now asserts a shared rejection code, which that file defines as "an HONEST shared-cause claim". The retryability label saves it from giving *wrong* remedy advice (it renders the `'unknown'` label), but the operator loses the free-text message that was previously their only actionable signal and gains a code that does not narrow the cause.
>
> This does not need fixing in this PR — the client-side change is right and the FE is a separate feature — but it should be filed as a follow-up on #2804's epic, and the honest options are either (a) render `sampleReason` alongside the code when `deriveRetryabilityClass` returns `'unknown'`, or (b) have the fallback only fire for ShipX codes that actually narrow the cause. I'd take (a): it is FE-local and does not put a carrier-vocabulary allowlist into the client.

**[SUGGESTION]** — `inpost-http-client.ts`, the new `logger.warn`

> The comment correctly refuses to log `errorBody.details` because it can echo a rejected address fragment. `message` is one field over from `details` in the same untrusted body and can carry the same content for some ShipX validation shapes — and the shipments UI treats carrier cause text as sensitive enough to hide entirely from the `viewer` role (`ShipmentsPage`'s permission gate, per `shipment-triage-strip.tsx`'s header). Worth either truncating `message` in the log line or noting explicitly why it is considered safe where `details` is not; as written the stated rationale distinguishes two fields that may not actually differ.

**[SUGGESTION]** — the two-tier `providerCode` semantics

> The `firstDetailKey` docblock is admirably explicit that "a consumer matching on `providerCode` cannot tell which tier answered from the value alone". Since `deriveRetryabilityClass` is a real consumer that string-matches on shape, consider whether the fallback should be namespaced (`shipx.not_found`, mirroring the existing `preflight.*` / `api.http-NNN` family convention documented in `shipping-provider-rejection.types.ts`) rather than bare. That would make the tier readable from the value, keep the `'unknown'` classification, and let a future `deriveRetryabilityClass` arm handle the family deliberately. Bare codes also risk a silent collision with a ShipX *field* name, which would then mean two different things at the same key. Not blocking — but this is a stable operator-facing identifier that will be hard to change later.

**Positive observations**

- Untrusted-input handling is right: `safeParseError` returns an unchecked `as ShipXErrorBody` cast, and the new code correctly does **not** trust the declared type — `typeof === 'string' && trim().length > 0`, with a dedicated test for the blank-string case. That is the correct posture for a boundary cast, and the test asserting `providerCode: null` for `"   "` is exactly the edge case that usually gets skipped.
- The test comment openly states that `not_found` is "a plausible placeholder, not a verified value" against the sandbox. Recording the limit of what was verified, in the test that would otherwise read as evidence, is the right call — it stops a future reader treating the fixture as a confirmed carrier contract.
- Tests follow AAA, cover the three real branches (fallback fires / rejected as blank / neither present), and the existing 17 field-error cases are untouched — so the additive claim is testable rather than asserted.
- Barrel doc comment updated in the same PR, so the package's public description of its `providerCode` families does not drift from behaviour.

### Step 4 — Final Assessment

**Summary.** A small, correct, well-tested Integration-layer fix that turns a `providerCode: null` dead end into a stable carrier code for the no-`details` ShipX failure shape. Layer placement, precedence, untrusted-input handling and test quality are all right, and the change is provably neutral on the retryability derivation that consumes `providerCode`. The one thing worth acting on is downstream and not in this diff: the shipments triage strip changes branch as a result, trading the operator's raw carrier message for a coarse code that does not narrow the cause — that deserves a follow-up issue rather than a hold on this PR.

**Merge readiness:** 🔄 **Approve with changes** — mergeable as-is; please file the triage-strip follow-up before closing #2804's epic.

**Priority fixes**
1. File the follow-up for `shipment-triage-strip.tsx` losing `sampleReason` once a coarse code exists (IMPORTANT, FE, separate PR).
2. Consider namespacing the fallback as `shipx.<code>` to keep the two `providerCode` tiers distinguishable at the value (SUGGESTION — cheapest to decide now, expensive later).
3. Truncate or justify `message` in the new warn log (SUGGESTION).

norbert-kulus-blockydevs:
Addressed the two review suggestions in a6a3a8a:

- **Namespaced the `providerCode` fallback as `shipx.{code}`** (e.g. `shipx.not_found`) so the two `providerCode` tiers — a ShipX field name vs. a ShipX error code — are distinguishable from the value alone, and a fallback code can never collide with a field name at the same key. Verified it's still classified `'unknown'` by `deriveRetryabilityClass` (matches no `preflight.*` / `command.*` / `api.http-NNN` family), so this is retryability-neutral, as before.
- **Truncated the logged `message` to 200 chars** in the new no-details warn log — `message` sits one field over from `details` in the same untrusted ShipX body and can carry the same kind of content for some validation shapes, so it's no longer echoed unbounded.

Updated the existing test's expectation to `shipx.not_found`, added a new test asserting the warn log truncates a long `message`, and updated the docblocks (`firstDetailKey`, the fallback comment) plus the barrel doc comment in `index.ts` to reflect the namespace.

On the IMPORTANT finding (`shipment-triage-strip.tsx` losing `sampleReason` once a coarse code exists): agreed this is real and out of scope for this PR — it's a separate FE change against #2804's epic. Not filing the follow-up issue myself; flagging it here for whoever owns that epic to file, per the review's own recommendation.

`pnpm --filter @openlinker/integrations-inpost lint` and the full `inpost-http-client.spec.ts` suite (19/19) pass.

piotrswierzy (approved):
## Re-review at `a6a3a8a` (delta since the approve at `5804cdc`)

### Verification of prior findings

**[IMPORTANT] `shipment-triage-strip.tsx` branch flip — NOT addressed in code, and correctly so; the follow-up is still unfiled.**

Re-read on the current head: `shipment-triage-strip.tsx:70` still derives `retryabilityClass` from `group.providerCode !== null`, and lines 96-107 still choose between the code branch and the `sampleReason` branch on the same test. Because the fallback resolves to a non-null `shipx.*` value, a no-`details` InPost failure now takes the code branch and **`sampleReason` is not rendered at all**. The prior review said this belongs in a follow-up rather than a hold, and I stand by that — but `search_issues` for a triage-strip / `sampleReason` follow-up returns **0 results**, and the author's last comment explicitly declines to file it ("flagging it here for whoever owns that epic"). It is not filed. Please file it against #2804's epic before closing that epic — that was the one condition on the previous approval.

One thing the namespacing changed for the worse, worth putting in that follow-up: the effect is not only the lost `sampleReason`. `group-failed-shipments-by-cause.ts:100-103` keys the group on `providerCode` whenever it is non-null and **skips `MIN_CAUSE_KEY_LENGTH`** on that path, on the stated grounds that "a `providerCode` is an exact structured value". `shipx.validation_failed` is not — it is precisely the coarse string that file's own header warns about. So a connection whose failures were previously grouped (or correctly *not* grouped) by normalised message can now collapse several unrelated causes into one group that the strip labels "report the same rejection code", which that module defines as "an HONEST shared-cause claim". Option (a) from the prior review (render `sampleReason` alongside the code when `deriveRetryabilityClass` returns `'unknown'`) addresses the display half; the grouping half may want the length/coarseness guard extended to the `shipx.*` family.

**[SUGGESTION] namespace the fallback — DONE.** `shipx.${error.trim()}` at `inpost-http-client.ts:212`, with the tier documented on `firstDetailKey`'s docblock and a third family named in the barrel comment. Re-verified retryability-neutral against the FE mirror `apps/web/src/features/shipments/lib/shipment-retryability.ts:35-51`: `shipx.*` matches no `api.http-NNN` / `preflight.` / `command.` / `target_point` arm, so it falls to `'unknown'` — the same class `null` produced. Unchanged behaviour, readable tier, collision with a ShipX field name now impossible. Good call taking this now rather than after the value is persisted in `shipments.providerCode` across installs.

**[SUGGESTION] truncate/justify `message` in the warn log — DONE.** `WARN_LOG_MESSAGE_MAX_LENGTH = 200` with the reasoning stated in place ("`message` sits one field over from `details` in the same untrusted body"), and a dedicated test asserting the 500-char message does not appear whole while the 200-char prefix does. The stated rationale no longer distinguishes two fields that may not differ; it treats both as untrusted and bounds one.

### Assessment of the current diff

- Precedence unchanged: `firstDetailKey(flatDetails) ?? shipxErrorCode`, and `providerDetails` still gated on `Object.keys(flatDetails).length > 0`, so the #885/#1807 field-error path is untouched and `providerDetails` is never fabricated for a fallback-only rejection.
- The untrusted-cast guard survives the namespacing: the `typeof === 'string' && trim().length > 0` test runs **before** the prefix is applied, so a blank `error` yields `null`, not `'shipx.'`. That ordering is the one thing the namespace could have broken, and the blank-string test at `:174` pins it.
- Layer placement correct — Integration package only, no CORE change, no port/contract change, no schema change, so no migration is owed.
- Test quality: the fixture comment still openly records that `not_found` is a plausible placeholder rather than a live-confirmed value, and the first test now calls `requestBinary` (the path `InpostShippingAdapter.fetchLabel` actually uses) rather than `request`. #2804's fourth acceptance criterion still needs the live-sandbox probe to be provably met — that is a verification gap on the issue, not a defect in this diff.
- New logger spy test asserts `warn` called exactly once, which also incidentally pins that a 4xx does not re-enter the retry loop.

### Merge readiness

✅ **Approve.** Both suggestions taken, cleanly and with the reasoning recorded in the code. Nothing blocking, nothing important left in this diff.

**Remaining, outside this PR:** file the `shipment-triage-strip.tsx` follow-up (now with the grouping half added) before #2804's epic closes, and run the live-sandbox probe that would confirm the fourth acceptance criterion.

norbert-kulus-blockydevs:
Filed the follow-up: #2873 — covers both halves from the re-review: the triage-strip display regression (`sampleReason` dropped once a `shipx.*` code exists) and the grouping half (`MIN_CAUSE_KEY_LENGTH` skip treating a coarse code as exact). Proposed fix in the issue is the reviewer's option (a) for display plus a `deriveRetryabilityClass(...) === 'unknown'` coarseness check for grouping — one shared test, no new arm on `shipment-retryability.ts`.

Nothing else from either review round is outstanding. The only other open item — the live-sandbox probe to confirm the `not_found`/`error` shape against a real purged label (#2804's 4th AC) — is a verification gap on #2804 itself, not something this PR's diff can close; both reviews treated it as such rather than as a blocker here.

norbert-kulus-blockydevs:
Follow-up filed, and implemented - the one condition on the approval is closed.

- Issue: #2873
- PR: #2874 (open, against `main`)

It covers both halves the re-review named:

**Display** - `shipment-triage-strip.tsx` branched on `providerCode !== null`, so a `shipx.*` code flipped the group onto the rejection-code branch and stopped rendering `sampleReason`. The branch is now keyed on EXACTNESS, and a coarse code renders alongside the raw message, labelled with its own `'unknown'` retryability class so the strip states that it does not classify the cause. Option (a) from the first review, as recommended.

**Grouping** - `group-failed-shipments-by-cause.ts` now composes a coarse code with the normalised message (`{code}::{normalisedMessage}`) instead of trusting it alone, so the code stays visible on the group without collapsing unrelated causes, and the length guard applies to every key that leans on the lossy text half.

One correction to the framing, recorded in the issue and the PR: extending `MIN_CAUSE_KEY_LENGTH` would not have caught this. `shipx.validation_failed` is 22 characters and passes that guard comfortably - the defect is coarse-treated-as-exact, not length. So the fix introduces an explicit exactness predicate rather than widening the length rule.

The predicate is `deriveRetryabilityClass(...) !== 'unknown'`, deliberately rather than a list of code families: that classifier is already the sanctioned FE mirror of `libs/core/src/shipping/domain/provider-code-retryability.ts`, every family that narrows the cause classifies to something other than `'unknown'`, and this way no new arm is added to the mirror and no carrier vocabulary enters the browser. The `shipx.*` namespace you asked for here is what makes it work cleanly on the FE side too.

Note #2874 does not depend on this PR and can merge in either order - the "non-null implies exact" assumption was already wrong for any opaque carrier-surfaced code (DPD can surface one); #2805 makes the case common rather than incidental.

Still outstanding on #2804, and not covered by either PR: the live-sandbox probe against a purged shipment that would confirm the fourth acceptance criterion.

PR #2808, opened 2026-09-02
feat(orders,analytics): add product identity to coverage-order rows, wire product-sales-table exclusion notes

### LINKED ISSUE #2713: [TASK] CORE/Interface — Aggregate-by-connection reads for Data Coverage categories
## Problem / Context

The channel-table exclusion annotations shipped in PR #2701 (epic #2452, Phase 8, #2481) need,
per open Data Coverage category, "how many affected orders does connection X have" — not the
orders themselves. Today the only way to get that is `apps/web/src/features/analytics/components/
channel-sales-table.tsx`'s `useCoverageCrossReferenceQuery`, which pages through the FULL affected-
order list via the existing paginated endpoints (`GET /analytics/coverage/currency-mismatch-orders`,
`GET /analytics/coverage/tax-orders`) and groups by `sourceConnectionId` client-side. On an install
with a large affected-order population (`netExcludedCount`/`unconvertedCount` are cumulative,
unbounded counts — see `docs/architecture-overview.md` § Currency / § Orders), this drains
potentially many pages of full order rows over the network just to produce a handful of
per-connection counts.

Both underlying reads already do (or can cheaply do) the aggregation server-side:
- `OrderRecordRepositoryPort.findCurrencyMismatchOrders` / `findProductMatchingErrorOrders`
  (`libs/core/src/orders/domain/ports/order-record-repository.port.ts`) are plain SQL reads —
  grouping by `sourceConnectionId` is a `GROUP BY`, not a new read pattern.
- `TaxCoverageDetectionService` (`libs/core/src/orders/application/services/
  tax-coverage-detection.service.ts`) already fetches its FULL unpaged candidate population into
  memory and classifies it in the application layer (its own doc comment: "Not pushed into SQL" —
  a live catalogue read has no SQL equivalent), so an aggregate-by-connection variant reuses the
  same classification pass and just changes what it returns — no new expensive computation. Verified
  against the current codebase: `classify()` rows already carry `sourceConnectionId` via `toRow`,
  and the service already has three siblings (`getCategoryPage`, `getCategoryCounts`,
  `getAllCategoryPages`) that each reuse one `classify()` call and reshape the result differently —
  a fourth sibling grouping by connection instead of slicing pages is the same pattern, not a new one.

## Proposed Solution

Add one new read per detector, returning `{ sourceConnectionId, affectedCount }[]` instead of full
order rows:

- `IOrderRecordService.getCurrencyMismatchOrdersByConnection(filters, currentReportingCurrency)`
  → thin pass-through to a new `OrderRecordRepositoryPort` method doing `SELECT sourceConnectionId,
  COUNT(*) ... GROUP BY sourceConnectionId`, mirroring `getCurrencyMismatchOrders`'s existing
  cross-context-boundary pass-through shape (`libs/core/src/orders/application/interfaces/
  order-record.service.interface.ts`).
- `ITaxCoverageDetectionService.getAllCategoryCountsByConnection(filters, currentReportingCurrency,
  includeBackfilledPreRollout?)` → reuses the existing `classify()` pass
  (`tax-coverage-detection.service.ts`), groups the three category arrays by `sourceConnectionId`
  in-memory instead of paginating them. Same interface-doc precedent as `getAllCategoryPages`
  ("ONE classify pass ... calling getCategoryPage three times would re-run the live-catalogue
  classification pass three times").
- `IOrderRecordService.getProductMatchingErrorOrdersByConnection(filters)` — same SQL-group-by shape
  as currency. **Needs confirmation before implementing**: PR #2701 deliberately does NOT
  cross-reference `product-matching` on the channel table (per its own file header comment in
  `channel-sales-table.tsx`) — verify there's a real consumer before building this one, to avoid
  shipping an unused aggregate.
- One new controller endpoint aggregating all of the above, mirroring
  `AnalyticsCoverageController.getCoverage`'s existing `Promise.all` composition shape
  (`apps/api/src/analytics/http/analytics-coverage.controller.ts`), e.g.
  `GET /analytics/coverage/by-connection?from=&to=&sourceConnectionId=` returning one row per
  `(category, sourceConnectionId)` pair with `affectedCount`.

## Classification

**Type**: CORE (application services) + Interface (new controller route)
**Layer**: Application, Interface
**File(s)**:
- `libs/core/src/orders/domain/ports/order-record-repository.port.ts`
- `libs/core/src/orders/infrastructure/persistence/repositories/order-record.repository.ts`
- `libs/core/src/orders/application/interfaces/order-record.service.interface.ts`
- `libs/core/src/orders/application/services/order-record.service.ts`
- `libs/core/src/orders/application/services/tax-coverage-detection.service.interface.ts`
- `libs/core/src/orders/application/services/tax-coverage-detection.service.ts`
- `apps/api/src/analytics/http/analytics-coverage.controller.ts` (or a new sibling controller)

## Dependencies

Follow-up to epic #2452 Phase 8 (#2481, shipped as PR #2701) — depends on nothing from that PR being
reverted; purely additive.

## Assumptions

- The new endpoint reuses the SAME `SalesAnalyticsFilters` / date-range validation
  (`MAX_COVERAGE_RANGE_DAYS`) `AnalyticsCoverageController` already enforces.
- `product-matching` aggregation is included for completeness but has no current FE consumer
  (PR #2701 explicitly excludes it from channel cross-referencing) — confirm before implementing
  whether to scope it out to save effort.
- Response shape is `{ category, rows: { sourceConnectionId, affectedCount }[] }[]`, not a
  per-connection top-level grouping — matches how the FE will consume it (see the paired FE issue).

## Acceptance Criteria

- [ ] New repository/service methods return counts grouped by `sourceConnectionId`, verified against
      a fixture with orders split across ≥2 connections
- [ ] Tax aggregate reuses the existing `classify()` pass — no second live-catalogue read per
      connection (assert via a spy/mock call count in the unit test)
- [ ] New controller endpoint documented in Swagger, date-range validated the same way as
      `GET /analytics/coverage`
- [ ] No architecture boundary violations (CORE ↔ Integration); cross-context reads still go through
      `I*Service`, never a repository port directly

### LINKED ISSUE #2714: [TASK] Frontend — Swap channel-table exclusion cross-reference to the aggregate endpoint
## Problem / Context

`ChannelSalesTable`'s `useCoverageCrossReferenceQuery` (`apps/web/src/features/analytics/components/
channel-sales-table.tsx`, shipped in PR #2701 / #2481) pages through each open category's FULL
affected-order list client-side, purely to build a `connectionId -> category -> count` map for the
`.excl-note` annotations. This is flagged in the hook's own doc comment as a "per-page-load,
non-hot-path read" risk on installs with large affected-order counts.

## Proposed Solution

Once #2713's aggregate-by-connection endpoint ships, replace `useCoverageCrossReferenceQuery`'s
drain-all-pages loop with a single call to the new endpoint. `buildChannelExclusionMap` stays (or
simplifies, since the grouping now happens server-side) — the component only needs to stop draining
pages itself.

## Classification

**Type**: Frontend
**Layer**: Feature (analytics)
**File(s)**:
- `apps/web/src/features/analytics/components/channel-sales-table.tsx`
- `apps/web/src/features/analytics/hooks/use-coverage-cross-reference-query.ts`
- `apps/web/src/features/analytics/api/analytics-coverage.types.ts` (new response type)
- `apps/web/src/features/analytics/api/analytics.api.ts` (or wherever the coverage API calls live)

## Dependencies

- Depends on #2713 (aggregate-by-connection endpoint) shipping first.
- Follow-up to epic #2452 Phase 8 (#2481, shipped as PR #2701).

## Assumptions

- `coverageFilters` prop shape stays the same; only the query implementation changes.
- Existing regression tests in `channel-sales-table.test.tsx` (the #2481 guards) keep asserting the
  same behavior against the new, simpler data source — they should not need to change shape, only
  their mocked `apiClient` calls.

## Acceptance Criteria

- [ ] `useCoverageCrossReferenceQuery`'s pagination loop is removed; one request per category
      (or one request total, if the endpoint returns all categories together)
- [ ] Existing `.excl-note` cross-reference regression tests still pass unmodified in behavior
- [ ] Tests added or updated for the new API call shape
- [ ] No architecture boundary violations (FE stays thin; no business logic duplicated from BE)

### LINKED ISSUE #2799: [TASK] CORE+Frontend — Add product/line-item identity to coverage-order rows so product-sales-table can carry exclusion annotations
## Problem / Context

Known, deliberately deferred gap from epic #2452's tracking PR #2668, and from #2481's own written assumption. Task 8.2 (`.excl-note` wiring) requires `product-sales-table.tsx` to annotate a product row whose sales figures are under-counted by an open coverage category (mirroring what `channel-sales-table.tsx` already does at the channel level, landed via PR #2701).

`TopProductRow` (`apps/web/src/features/analytics/api/top-products.types.ts`) is keyed by `productId` (explicitly the cross-channel join key, per the file's own doc comment — never SKU, since e.g. Allegro sets `sku = offer.id`). None of the three coverage-order row types the backend exposes carry a `productId`/SKU/line-item reference:

- `CurrencyMismatchOrderRow` — `{ internalOrderId, sourceConnectionId, ... }`, no product field.
- `TaxCoverageOrderRow` — `{ internalOrderId, sourceConnectionId, placedAt }`, no product field.
- `ProductMatchingErrorOrderRow` — `{ internalOrderId, sourceConnectionId, recordStatus, mappingFailureReason, createdAt }`, no product field.

So there is no correct client-side cross-reference from an excluded order to a specific product row. #2481's own "Assumptions" section states this exact contingency: "if the affected-order lists from Phase 4 don't carry enough identifying info (SKU/channel id) to cross-reference cleanly, that's a signal Phase 4's detector output needs a small shape addition — flag it back rather than approximating the match." Approximating (e.g. flagging every product sold on an affected channel) would misattribute the coverage category to unaffected products — precisely the "wrong category attributed to one row" bug class #2481's acceptance criteria exist to prevent, since a per-product page has no channel-wide granularity to fall back on safely.

`product-sales-table.tsx` was not touched by PR #2701 (which only wired `channel-sales-table.tsx`); no `AnalyticsExclusionNote` usage exists in the product table today.

## Proposed Solution

1. Extend the three coverage-order row types (`CurrencyMismatchOrderRow`, `TaxCoverageOrderRow`, `ProductMatchingErrorOrderRow` in `libs/core/src/orders/domain/types/coverage-detection.types.ts`) to additionally carry the internal `productId` (and variant id, if the underlying `OrderLineItem`/order-record read already resolves it) for at least one representative line per order — enough for a client-side join against `TopProductRow.productId`, without turning this into a full line-item listing endpoint.
2. Thread the new field through to the corresponding HTTP DTOs (mirroring the plumbing pattern issue #2798 uses for the tax-rate field, so both additive changes to this vocabulary land consistently).
3. Wire `product-sales-table.tsx` to the existing `AnalyticsExclusionNote` component (built in PR #2701 for `channel-sales-table.tsx`) using the new productId-bearing coverage data, following the same cross-reference pattern already proven for the channel table.
4. Tests: a fixture with two categories both present in the dataset asserting each product row gets the *correct* category (the exact bug class #2481's ACs guard against), and a fixture covering every category at least once asserting no affected product is missing its annotation.

## Classification

**Type**: CORE + Frontend
**Layer**: Application (orders context) / Interface (HTTP DTOs) / Frontend feature component
**File(s)**:
- `libs/core/src/orders/domain/types/coverage-detection.types.ts`
- `libs/core/src/orders/application/services/*coverage-detection*.service.ts` (currency, tax, product-matching detectors)
- Corresponding `apps/api/src/analytics/http/dto/*-coverage-orders-response.dto.ts` files
- `apps/web/src/features/analytics/components/product-sales-table.tsx`
- `apps/web/src/features/analytics/components/analytics-exclusion-note.tsx` (reused, not modified unless the productId shape needs a prop addition)

## Dependencies

- Builds on merged Phase 4 (#2463) and Phase 8 (#2479, PR #2701 — `AnalyticsExclusionNote` + `channel-sales-table.tsx` wiring already shipped, this issue mirrors that pattern for the product table).
- Stacks on top of PR #2785 (`2778-exchange-rate-endpoint`) in the epic #2452 branch chain, and should land after #2798 if both touch `coverage-detection.types.ts` in the same area (coordinate to avoid merge conflicts, not a hard dependency).

## Assumptions

- One representative `productId` per affected order line is sufficient for the exclusion note's granularity (an order excluded due to one line's issue still reasonably flags that line's product) — a full per-line breakdown is out of scope unless review finds it necessary.
- The three detector services (currency, tax, product-matching) each already read enough order/line data internally to resolve a `productId` without an additional query per order; if any detector's underlying candidate query does not carry this today, that detector's query needs a join/column addition as part of this issue.

## Acceptance Criteria

- [ ] All three coverage-order row types carry a `productId` (or equivalent line-item identity) sufficient to join against `TopProductRow.productId`.
- [ ] `product-sales-table.tsx` renders `AnalyticsExclusionNote` for a product row provably correct against its own underlying order data — test with a fixture where two categories are both present, asserting each row gets the right one.
- [ ] No product row belonging to an open category's affected set is missing its annotation — test with a fixture covering every category at least once.
- [ ] No approximation (e.g. channel-wide flagging) is used as a substitute for the real cross-reference.
- [ ] Tests added or updated.

norbert-kulus-blockydevs:
Closes #2799

## Summary

- Adds every distinct product a `CurrencyMismatchOrderRow` touches (`lineProducts[]`, batched via a new `OrderLineItemRepositoryPort.findProductRefsByOrderIds` lookup, resolved in `OrderRecordService.getCurrencyMismatchOrders` rather than the repository, since the repository has no line-item access of its own).
- `TaxCoverageOrderRow` already carries per-line `productId`/`variantId` via `lineRates` (#2798, landed just below this branch) - no change needed there beyond what #2798 shipped.
- Adds a `productId: null` / `variantId: null` pair to `ProductMatchingErrorOrderRow`, documented as always-null: an order in this category never resolved its item reference to an internal product id in the first place (that failed resolution is why the row exists), so there is no honest value to report. It stays excluded from product-level cross-referencing, mirroring the existing channel-level precedent (`CROSS_REFERENCEABLE_CATEGORIES` already excludes `product-matching` for the same structural reason).
- Threads the product identity through the HTTP DTOs (`CurrencyMismatchOrderDto` + `CurrencyMismatchLineProductDto`, `ProductMatchingOrderDto`) and the corresponding FE types.
- Adds `buildProductExclusionMap` (`apps/web/src/features/analytics/lib/product-exclusion-map.lib.ts`), mirroring `buildChannelExclusionMap`, grouping the full affected-order lists by `productId` (every distinct id across a currency row's `lineProducts` or a tax row's `lineRates`, deduplicated per order).
- Wires `product-sales-table.tsx` to `AnalyticsExclusionNote` using the same `CROSS_REFERENCEABLE_CATEGORIES` pattern `channel-sales-table.tsx` already uses (currency + tax-a/b/c, never product-matching).

## Review round 1 (@piotrswierzy) - what changed

**BLOCKING 1 - a multi-product currency order annotated only ONE product row.** Fixed in `813793768`. `findRepresentativeLinesByOrderIds` was a `DISTINCT ON (orderRecordId) ORDER BY lineNumber ASC`, which collapses an order to its single lowest-line-number product - so for an order spanning A/B/C only A's row got a note, and B's and C's were under-counted with no annotation at all. It is now `findProductRefsByOrderIds`, grouped by `(orderRecordId, productId, variantId)`: same single query, same batching, no N+1, every product returned. `CurrencyMismatchOrderRow.productId`/`variantId` became `lineProducts[]` end-to-end (row -> DTO -> FE type), and `distinctProductIdsOf` unions it with a tax row's `lineRates`, so both categories are read the same way. Renamed away from "representative" as asked, so the contract no longer bakes the approximation in.

**The "No approximation used anywhere" claim was inaccurate and is withdrawn.** It is accurate now: the only category that reports no product is `product-matching`, which is excluded from cross-referencing entirely rather than approximated.

**IMPORTANT 3 - repository specs asserted call shapes, not behaviour.** Added real-Postgres coverage in `apps/api/test/integration/sales-analytics-aggregates.int-spec.ts` (which this PR already touched), exercising `IOrderRecordService.getCurrencyMismatchOrders` end to end: one order with three lines across two products asserts the full set comes back, and an order with no lines asserts an empty array rather than a fabricated entry. The mocked query-builder specs stay as unit-level shape guards, but the dedup semantics are now pinned by SQL that actually runs.

That integration run also surfaced a **pre-existing fixture bug** in the same file, from #2713: the by-connection test stamped an order `reportingCurrency: 'EUR'` and then queried with `'EUR'` as the current setting - which makes it current-era by definition, so the predicate correctly did not count it and the test's expected count of `2` could never hold. Fixed the fixture to a genuinely stale-era stamp (`'PLN'` under a current `'EUR'`), matching what its own comment says it seeds. 17/17 integration tests in the file now pass.

**SUGGESTION 4 - `CoverageOrderLite` in the wrong file.** Moved to `product-exclusion-map.lib.ts`, its sole consumer. `CROSS_REFERENCEABLE_CATEGORIES` / `CrossReferenceableCategory` stay in one place (`channel-exclusion-map.lib.ts`) and are re-exported as before.

**SUGGESTION 5 - permanently-null fields on `ProductMatchingErrorOrderRow`.** Left as-is per your own note. The doc comment no longer claims "shape parity with `CurrencyMismatchOrderRow`" (that row now carries an array), and states instead why a permanently-empty array would assert nothing the literal `null` does not.

**SUGGESTION 6 - `useCoverageCrossReferenceQuery` still page-drains for the product table.** Acknowledged, no change: that is the grain, and a `by-product` aggregate is the eventual counterpart if page counts grow.

**IMPORTANT 2 - undeclared scope.** Correct that the diff is wider than `Closes #2799`, and here is why, so it can be judged rather than guessed at. This branch is the base of a stack: **#2713 was implemented on top of it as PR #2810 and #2714 as PR #2811** - both are already **merged and closed** through their own review, and merging #2810 (whose base branch was this one) is what put `cb10e9617` / `ee3aecb1e` / `343475d33` / `14cdb3c7f` into this branch's history. So `GET /analytics/coverage/by-connection`, `getAllCategoryCountsByConnection`, `findCurrencyMismatchOrdersByConnection`, `use-coverage-by-connection-query.ts`, the `buildChannelExclusionMap` reshape and the two `docs/plans/implementation-plan-coverage-by-connection*.md` docs all belong to those two already-reviewed PRs, not to #2799. `867bbcd6f` is the review-response commit for #2810 (predicate dedup + Swagger enum narrowing), and `32e7f93ab` is an unrelated one-line dead-code removal that rode along on the branch.

Nothing here is unreviewed, and `Closes #2713` / `Closes #2714` would be no-ops (both issues are already closed by their own merged PRs), so re-pointing the base would only rewrite history that the merged children are already built on. Flagging it in this body instead, as the second of your two options.

## Tests

- Backend unit: `OrderLineItemRepository.findProductRefsByOrderIds` (grouped lookup, multi-product order, empty-input short-circuit, no-match handling) and `OrderRecordService.getCurrencyMismatchOrders` (multi-product enrichment merge, no-lookup-on-empty-page).
- Backend integration (Testcontainers Postgres): the two new cases above; whole file 17/17 green.
- Frontend: four tests mirroring `channel-sales-table.test.tsx`'s `#2481` regression guards at the product grain - correct per-category attribution across two products, full per-category coverage for one product, **every product of a single multi-product currency order annotated (the BLOCKING 1 guard)**, and explicit confirmation that `product-matching` never annotates a product row.
- Updated existing controller/repository/service specs to the new row shape.

## Verification

- `pnpm lint` - 0 errors (only pre-existing warnings) across `libs/core`, `apps/api`, `apps/web`.
- `pnpm type-check` - clean across `libs/core`, `apps/api`, `apps/web`; plus a one-off `tsc` run including `apps/api/test/**` (which `tsconfig.type-check.json` excludes) so the new int-spec is type-checked too.
- `pnpm check:invariants` - cross-context imports, service interfaces and workspace dep declarations all pass. `check-migration-timestamps` reports 2 pre-existing violations on this branch (two analytics migrations now sort before `main`'s newer tip) - inherited branch staleness, unrelated to this change, and not introduced here.
- `libs/core` orders suite: 641/641. `apps/web` full unit suite: 4227/4227 (1 pre-existing skip). `apps/api` analytics controller specs: 29/29.

## Architecture notes

- No CORE <-> Integration boundary changes.
- `OrderRecordService` already injects `OrderLineItemRepositoryPort` (used elsewhere for `buildTopProducts`), so the enrichment composes at the existing application-service layer rather than reaching into the repository.


norbert-kulus-blockydevs:
## PR Review: #2808 — feat(orders,analytics): add product identity to coverage-order rows, wire product-sales-table exclusion notes

### Summary

This PR extends the Data Coverage drill-down rows (`CurrencyMismatchOrderRow`, `ProductMatchingErrorOrderRow`) with product/variant identity, and wires `product-sales-table.tsx` to the same `.excl-note` cross-reference pattern `channel-sales-table.tsx` already established (#2481 Phase 8). It is a close, faithful mirror of that existing pattern rather than a new design, with one new batched repository lookup (`OrderLineItemRepositoryPort.findRepresentativeLinesByOrderIds`) to backfill a representative `productId`/`variantId` for currency-mismatch rows that don't otherwise carry per-line data. `TaxCoverageOrderRow` needed no change since #2798 (merged as this branch's base) already carries per-line `productId`/`variantId` via `lineRates`. Overall quality is high: the new query follows the exact quoted-camelCase-column idiom already used elsewhere in `order-line-item.repository.ts`, the enrichment is batched (no N+1), the layering (join composed in the application service rather than the repository) matches the documented repository-ports rules, and the frontend wiring reuses `useCoverageCrossReferenceQuery`/`AnalyticsExclusionNote`/`CROSS_REFERENCEABLE_CATEGORIES` verbatim rather than duplicating them. No CORE↔Integration boundary issues, no schema change (so correctly no migration), and test coverage is proportionate (repository DISTINCT ON behavior, service enrichment/merge, and three FE regression tests mirroring the channel table's own #2481 ACs at the product grain).

### Findings

**[SUGGESTION]** — `libs/core/src/orders/domain/types/coverage-detection.types.ts` (`ProductMatchingErrorOrderRow`)
> The PR description and the field's own doc comment call this "shape parity" with `CurrencyMismatchOrderRow`/`TaxCoverageOrderRow`, but only `productId: null` was added — no matching `variantId: null`. Since `productId` is always `null` here anyway, this has no functional consequence (nothing reads `variantId` off this row), but if "shape parity" is the stated goal it's currently partial. Worth either adding `variantsId: null` too or softening the doc comment's "shape parity" framing to "join-key parity" — not blocking either way.

**[SUGGESTION]** — `apps/web/src/features/analytics/lib/channel-exclusion-map.lib.ts` (`CoverageOrderLite`)
> Widening this shared interface with `productId?`/`lineRates?` to serve `product-exclusion-map.lib.ts` is reasonable reuse (avoids a duplicate type), but it does mean `channel-exclusion-map.lib.ts` now carries fields it has zero readers for. The doc comment explains this clearly, so it's a fully acceptable trade-off — flagging only because a future reader skimming `channel-exclusion-map.lib.ts` in isolation might wonder why a "channel" file has per-product fields. No action needed given the comment already anticipates this.

### Positive observations

- The batched `findRepresentativeLinesByOrderIds` (DISTINCT ON, ordered by `lineNumber ASC`) is a textbook "first row per group" idiom, matches the file's existing quoted-column convention exactly, and is unit-tested for the empty-input short-circuit, the happy path, and the no-matching-rows case.
- The enrichment is correctly placed in `OrderRecordService` (which already composes both `OrderRecordRepositoryPort` and `OrderLineItemRepositoryPort` for `buildTopProducts`) rather than reaching into `OrderRecordRepository`, which has no line-item access — consistent with `docs/engineering-standards.md`'s repository-ports pattern and the doc comment explicitly states this reasoning.
- Empty-page short-circuit (`if (page.items.length === 0) return page;`) avoids an unnecessary batched call and is asserted by a dedicated unit test.
- Frontend wiring is a disciplined reuse, not a reimplementation: same fixed-hook-count pattern (4 calls to `useCoverageCrossReferenceQuery`, one per `CROSS_REFERENCEABLE_CATEGORIES` member, satisfying the hook's own "never a variable count" contract), same `.excl-note`/`data-table__stack` CSS classes, same `AnalyticsExclusionNote` component.
- `ProductMatchingErrorOrderRow.productId`'s doc comment is a genuinely good piece of documentation — it explains *why* no honest value can ever exist here (unresolved item reference is the reason the row exists at all) rather than just declaring the field always-null.
- Frontend tests mirror the exact regression shape channel-sales-table's #2481 tests guard against (cross-attribution correctness, full-coverage annotation, and product-matching's deliberate non-participation), which is the right test strategy for a "port an existing pattern to a new grain" change.
- No schema/entity change, so correctly no migration is included.

### Documentation gaps

None — the PR is self-documenting via inline doc comments that reference the precedent (#2481 Phase 8) and the originating issue (#2799) throughout, and no new architectural pattern is introduced that would need `docs/architecture-overview.md` or an ADR update.

### Final Assessment

**Merge Readiness: ✅ Approve** — no blocking or important issues found. The two suggestions above are optional polish and do not need to be addressed before merge.

norbert-kulus-blockydevs:
Addressed the review's first suggestion (shape parity for `ProductMatchingErrorOrderRow`):

- Added `variantId: null` to `ProductMatchingErrorOrderRow` alongside the existing `productId: null`, with a doc comment pointing to the same rationale (this category never resolves either reference, by construction).
- Threaded `variantId: null` through the row's construction site (`OrderRecordRepository.getProductMatchingErrorOrders`), the `ProductMatchingOrderDto` (+ `fromRow`), and the frontend `ProductMatchingOrder` mirror type, so the "shape parity" claim in the doc comments is now literally true end to end.
- Updated the two spec files that construct a `ProductMatchingErrorOrderRow` literal so they still type-check.

The second suggestion (about `channel-exclusion-map.lib.ts`'s `CoverageOrderLite` carrying fields with no readers) was left as-is — the review itself called it a fully acceptable, already-documented trade-off with no action needed.

Verified: `pnpm --filter @openlinker/core type-check`, `pnpm --filter @openlinker/api type-check`, `pnpm --filter @openlinker/web type-check` all clean; `pnpm --filter <pkg> lint` on all three touched packages shows 0 errors (only pre-existing warnings unrelated to this change).

Commit: be42d3796

piotrswierzy (changes requested):
# Code Review: product identity on coverage-order rows + product-sales-table exclusion notes (#2799)

**Scope**: 46 files / +2320 −111 against base `2798-tax-coverage-rate-in-row`. I reviewed the full file list across both pages plus the commit list.

---

## 1. Summary

The core of #2799 is well built. `findRepresentativeLinesByOrderIds` is a proper batched `DISTINCT ON` (correct Postgres first-row-per-group idiom, parameterised `IN`, empty-input short-circuit), the enrichment is composed in `OrderRecordService` rather than smuggled into `OrderRecordRepository` — with the right justification, since that repository has no `order_line_items` access and the service already composes both for `buildTopProducts` — and `ProductMatchingErrorOrderRow.productId` is typed as the **literal** `null` rather than `string | null`, which makes fabricating a value structurally impossible and is a genuinely good call, thoroughly documented. Extracting `CURRENCY_MISMATCH_FRAGMENT` so all five readers of that population share one predicate is a real improvement over the copied literal.

Two things stop me approving as-is: a **correctness gap in the currency branch of the product cross-reference** that the PR's own library docstring already argues against, and **substantial undeclared scope** in the diff.

---

## 2. Findings

### 🔴 BLOCKING

**1. A multi-product currency-mismatched order annotates only ONE product row — silently under-counting the rest** — `order-line-item.repository.ts` (`findRepresentativeLinesByOrderIds`), `product-exclusion-map.lib.ts` (`distinctProductIdsOf`)

`product-exclusion-map.lib.ts` states the rule this feature exists to satisfy, in its own header:

> *"an order whose lines span two products must annotate BOTH of that product-sales-table's rows, not just the first line's."*

The **tax** branch honours that — `lineRates` carries every line, and `distinctProductIdsOf` unions them. The **currency** branch does exactly what the sentence forbids: `DISTINCT ON (orderRecordId) ORDER BY lineNumber ASC` collapses the order to the single lowest-line-number product, so for an order spanning products A/B/C, only A's row gets an `AnalyticsExclusionNote`. B's and C's rows are under-counted by that order **with no annotation at all** — which is precisely the silent under-count the note was introduced to make visible, reproduced one grain down. Worse, it is invisible rather than approximate: the operator sees a clean row and concludes the number is complete.

This also makes the PR description's *"No approximation used anywhere"* inaccurate, and the port JSDoc's framing (*"needs **a** product per affected order"*) understates what the consumer needs — it needs **every** product the order touched, exactly as the tax path already provides.

The fix is small and costs the same single query:

```ts
// drop DISTINCT ON; group all lines instead
findProductRefsByOrderIds(
  orderRecordIds: string[]
): Promise<Map<string, Array<{ productId: string; variantId: string | null }>>>
```

…and widen `CurrencyMismatchOrderRow` to carry that list (or reuse the `lineRates`-shaped field so both categories present one uniform shape to `distinctProductIdsOf`, which would let that helper lose its currency/tax branch entirely). If a single representative id is genuinely wanted for some other reason, that is fine — but then the *exclusion map* still needs the full set, and the PR must stop claiming no approximation.

I'd also rename away from "representative": the word bakes the approximation into the contract, so a future reader has no signal that one-per-order was ever a problem.

### 🟡 IMPORTANT

**2. The diff contains three issues' work, but the description declares one**

The branch has `2713-coverage-by-connection-reads` and `2714-coverage-by-connection-fe` merged into it (commits `cb10e96`, `ee3aecb`, `343475d`, `14cdb3c`). So `GET /analytics/coverage/by-connection`, `AnalyticsCoverageByConnectionResponseDto`, `getAllCategoryCountsByConnection`, `findCurrencyMismatchOrdersByConnection`, `use-coverage-by-connection-query.ts`, the `buildChannelExclusionMap` rewrite, an unrelated `fix(web): remove dead publish-button helper`, and **688 lines of `docs/plans/implementation-plan-coverage-by-connection*.md`** all ride in this PR — none of it mentioned in the body, which says only `Closes #2799`.

Concretely: merging this merges #2713 and #2714 too, while GitHub closes neither, and the by-connection code is being reviewed separately on #2810 (commit `867bbcd` says so explicitly). Please either

- re-point the base so this PR carries only #2799's own commits, or
- update the body to declare `Closes #2799`, `Closes #2713`, `Closes #2714` and say the stack was collapsed.

Also: the two `docs/plans/implementation-plan-*.md` files look like working artefacts. `docs/plans/` is a real committed location, so if they're meant to stay, say so; if not, drop them.

**3. Repository specs assert query-builder call shapes, not behaviour**

`findRepresentativeLinesByOrderIds` and `findCurrencyMismatchOrdersByConnection` are tested by asserting `qb.where(...)`/`qb.orderBy(...)`/`qb.groupBy(...)` were called with exact strings. `testing-guide.md § Best Practices` calls this out directly ("Test Behavior, Not Implementation"), and it bites specifically here: the mock returns whatever `getRawMany` is told to return, so **`DISTINCT ON` correctness — the one thing that makes this query right — is never exercised**. A mock happily reports two rows for the same order id. Given `apps/api/test/integration/sales-analytics-aggregates.int-spec.ts` is already touched in this PR, one real-Postgres case (an order with three lines, asserting exactly one Map entry / the intended set) would cover the actual SQL semantics. Same for the `GROUP BY`.

I recognise this matches the existing style in `order-record.repository.spec.ts`, so it isn't a new sin — but these two queries are the ones where the shape *is* the logic.

### 🟢 SUGGESTIONS

**4. `CoverageOrderLite` now lives in the wrong file** — `channel-exclusion-map.lib.ts`

Its own docblock admits it: *"this type no longer has a role in THIS file's own function."* Its sole consumer is `product-exclusion-map.lib.ts`, which also re-exports `CROSS_REFERENCEABLE_CATEGORIES` back out of the channel module. Moving the type (and leaving the shared constant in one neutral module) would stop the product library importing from the channel library for things the channel library doesn't use.

**5. `ProductMatchingErrorOrderRow.productId/variantId` are permanently-null fields no consumer may read**

The reasoning for declaring them is written well, and the `null` literal type is the right encoding — but the category is excluded from `CROSS_REFERENCEABLE_CATEGORIES`, and `distinctProductIdsOf` skips falsy ids anyway, so nothing consumes them and nothing can. "Shape parity" with a row that is structurally never cross-referenced is thin justification for widening a DTO. Not worth reverting now; worth a second thought if a third category appears.

**6. `useCoverageCrossReferenceQuery` still page-drains for the product table**

#2714 removed the four draining calls from `ChannelSalesTable` but `ProductSalesTable` still needs per-order rows, so the drain survives there. That's correct and unavoidable at this grain — flagging only so it doesn't read as an oversight later. A `by-product` aggregate is the eventual counterpart if the page count grows.

### ✅ Positive observations

- `CURRENCY_MISMATCH_FRAGMENT` extraction, with a docblock naming all five readers and stating *why* they must share it. This is the kind of divergence that would otherwise be found by an operator, not a test.
- `getAllCategoryCountsByConnection` reuses ONE `classify()` pass for all three tax categories, with a spec (`never calls findNetExcludedOrderCandidates more than once`) that pins the property rather than trusting it — and a second spec asserting the by-connection sums equal `getCategoryCounts`' totals. That cross-check is exactly the guard that catches a grouping bug.
- The `be42d37` follow-up adding `variantId` end-to-end after review flagged that "shape parity" was only half-delivered — good responsiveness, and it fixed the doc/code disagreement rather than the doc alone.
- `buildChannelExclusionMap`'s comment explaining why it now `.set()`s instead of accumulating (server-side `GROUP BY` guarantees one row per pair) prevents a plausible "this looks like a bug" re-accumulate later.

---

## 3. Final assessment

### Merge readiness
❌ **Request changes**

The batched-lookup mechanics, the layering, the `null`-literal typing and the predicate dedup are all right. But the currency branch's one-product-per-order collapse means the product-level exclusion notes are **silently incomplete for exactly the orders that span several products**, which is the same class of silent under-reporting the exclusion note exists to eliminate — and the PR body asserts the opposite. That needs fixing (cheaply — same query, drop `DISTINCT ON`) before this lands.

### Priority fixes
1. **(BLOCKING)** Return every distinct product per order for the currency category, so a multi-product order annotates every product row it affected — matching what the tax path already does and what `product-exclusion-map.lib.ts`'s own header requires. Rename away from "representative". Correct the "No approximation used anywhere" claim in the description.
2. **(IMPORTANT)** Reconcile the PR's declared scope with its diff — re-base off #2713/#2714, or declare all three issues and the plan docs.
3. **(IMPORTANT)** Add one integration-level assertion for the `DISTINCT ON`/`GROUP BY` semantics; the current mocks cannot fail on them.
4. *(suggestion)* Move `CoverageOrderLite` to the module that actually consumes it.

norbert-kulus-blockydevs:
@piotrswierzy thanks - all findings addressed in `813793768`, plus the PR body rewritten to declare the real scope. Summary:

### BLOCKING 1 - multi-product currency order annotated only one product row

Fixed, and you were right that the library's own header already stated the rule the code broke.

`findRepresentativeLinesByOrderIds` is now `findProductRefsByOrderIds`, and `DISTINCT ON (orderRecordId) ORDER BY lineNumber ASC` became a `GROUP BY (orderRecordId, productId, variantId)`. Same single query, same batching, no N+1 - it just stops collapsing the order to its first line. `CurrencyMismatchOrderRow.productId`/`variantId` became `lineProducts[]` end to end (row -> `CurrencyMismatchLineProductDto` -> FE type), and `distinctProductIdsOf` unions it with a tax row's `lineRates`, so both categories are read through one path. Renamed away from "representative" as asked, so nothing in the contract implies one-per-order was ever acceptable.

Also withdrawn from the body: the "No approximation used anywhere" claim was inaccurate as written. It holds now - the only category reporting no product is `product-matching`, which is excluded from cross-referencing entirely rather than approximated.

New FE regression test asserts the case directly: one currency-mismatched order whose lines span `p1` and `p2`, both rows carrying a note.

### IMPORTANT 3 - specs asserted call shapes, not behaviour

Agreed, and taking the route you suggested: real-Postgres coverage in `apps/api/test/integration/sales-analytics-aggregates.int-spec.ts`, driving `IOrderRecordService.getCurrencyMismatchOrders` end to end. One order with three lines across two products asserts the full set comes back; an order with no lines asserts `[]` rather than a fabricated entry. The mocked specs stay as unit-level shape guards, but the dedup semantics now run against a server.

That run immediately paid for itself: it caught a **pre-existing fixture bug** in the same file from #2713. The by-connection test stamps an order `reportingCurrency: 'EUR'` and then queries with `'EUR'` as the current setting, which makes it current-era by definition - so the predicate correctly did not count it and the expected total of `2` could never hold. Fixed the fixture to a genuinely stale-era stamp (`'PLN'` under a current `'EUR'`), matching what its own comment says it seeds. 17/17 in the file now pass.

### IMPORTANT 2 - undeclared scope

Fair, and now spelled out in the body rather than left to be inferred. The short version: this branch is the base of a stack, and **#2713 shipped as PR #2810 with this branch as its base, #2714 as PR #2811** - both already merged and closed through their own review. Merging #2810 is what put those commits into this branch's history, which is why they show in this diff. `867bbcd6f` is the review-response commit for #2810, and `32e7f93ab` is an unrelated one-line dead-code removal that rode along.

So nothing in the diff is unreviewed, and `Closes #2713` / `Closes #2714` would be no-ops since both issues are already closed by their own merged PRs. Re-pointing the base would mean rewriting history the merged children are built on, so I took your second option and declared it in the body instead. The two `docs/plans/implementation-plan-coverage-by-connection*.md` docs came in with those PRs and stay - `docs/plans/` is the committed location for them.

### SUGGESTION 4 - `CoverageOrderLite` in the wrong file

Moved to `product-exclusion-map.lib.ts`, its sole consumer. `CROSS_REFERENCEABLE_CATEGORIES` / `CrossReferenceableCategory` stay in one place and are re-exported as before, so the product library no longer imports a product-identity type from the channel library that has no use for it.

### SUGGESTION 5 - permanently-null fields

Left as-is, per your own note. The doc comment no longer claims "shape parity with `CurrencyMismatchOrderRow`" (that row carries an array now) and instead says why a permanently-empty array would assert nothing the literal `null` does not.

### SUGGESTION 6 - product table still page-drains

Acknowledged, no change - that is the grain, and a `by-product` aggregate is the eventual counterpart if page counts grow.

### Verification

`pnpm lint` 0 errors, `type-check` clean across `libs/core` / `apps/api` / `apps/web` (plus a one-off `tsc` including `apps/api/test/**`, which `tsconfig.type-check.json` excludes, so the new int-spec is type-checked too). `libs/core` orders suite 641/641, `apps/web` full unit suite 4227/4227, `apps/api` analytics controller specs 29/29, the int-spec 17/17 on Testcontainers Postgres. `check:invariants` passes except `check-migration-timestamps`, which reports 2 violations inherited from branch staleness against `main`'s newer tip - pre-existing, not from this change.

PR #2810, opened 2026-09-02
feat(orders,analytics): aggregate-by-connection reads for Data Coverage categories

jakubretajczykBD:
## Summary
- Adds `OrderRecordRepositoryPort.findCurrencyMismatchOrdersByConnection` — a `GROUP BY sourceConnectionId` counterpart to `findCurrencyMismatchOrders`, same predicate.
- Adds `ITaxCoverageDetectionService.getAllCategoryCountsByConnection` — groups the existing `classify()` pass by connection, no extra live-catalogue reads.
- Adds `GET /analytics/coverage/by-connection`, composing both into `{ category, rows: { sourceConnectionId, affectedCount }[] }[]`, so `channel-sales-table.tsx` can stop deriving per-connection counts client-side by paging through full order lists.
- `product-matching` deliberately excluded — no FE consumer today (confirmed against `channel-sales-table.tsx`).

## Test plan
- [x] Unit tests: repository (`GROUP BY` query-builder assertions + row mapping), service pass-through, tax-aggregate service (incl. single-`classify()`-call assertion), controller (composition + validation + `sourceConnectionId` narrowing)
- [x] Integration test (Testcontainers Postgres): counts grouped across ≥2 connections, cross-checked against `findCurrencyMismatchOrders` per-connection totals; cancelled-order exclusion; empty-result case
- [x] `pnpm lint`, `pnpm type-check`, `pnpm check:invariants` all green

Closes #2713

PR #2811, opened 2026-09-02
feat(web,analytics): swap channel-table exclusion cross-reference to the aggregate endpoint

### LINKED ISSUE #2714: [TASK] Frontend — Swap channel-table exclusion cross-reference to the aggregate endpoint

jakubretajczykBD:
## Summary
- Adds `GET /analytics/coverage/by-connection` FE client (`getCoverageByConnection`, types, query key), mirroring the existing `getCoverage` pattern.
- New `useCoverageByConnectionQuery` — one request instead of `useCoverageCrossReferenceQuery`'s 4 page-draining calls.
- `buildChannelExclusionMap` reshaped in place (single caller, grouping now happens server-side) rather than duplicated.
- `ChannelSalesTable` swapped to the new hook; `enabled` gating changed from "each category open" to "any cross-referenceable category open" (same all-clear optimization, now for one request instead of four).
- `ProductSalesTable` is explicitly untouched — confirmed it needs per-order `productId`/`lineRates`, which the connection-level aggregate doesn't carry, so it keeps `useCoverageCrossReferenceQuery`.

## Test plan
- [x] Existing `.excl-note` regression tests (`channel-sales-table.test.tsx`) ported to mock `getCoverageByConnection`, same rendered-output assertions
- [x] New unit test for `buildChannelExclusionMap`'s reshaped signature (5 cases)
- [x] `pnpm --filter @openlinker/web type-check`, lint, and `vitest run src/features/analytics` — 225/225 passing, including `product-sales-table.test.tsx` unchanged

Depends on #2713 (PR #2810, not yet merged) — this PR targets that branch.

Closes #2714

### DIFF
--- a/apps/web/src/features/analytics/api/analytics-coverage.api.ts
+++ b/apps/web/src/features/analytics/api/analytics-coverage.api.ts
@@ -1,14 +1,22 @@
 /**
  * Analytics Coverage API client
  *
- * Thin request module for `GET /analytics/coverage` (#2466).
+ * Thin request module for `GET /analytics/coverage` (#2466) and
+ * `GET /analytics/coverage/by-connection` (#2713) — the latter is the
+ * server-side `GROUP BY sourceConnectionId` counterpart the former's
+ * per-category drill-downs used to derive client-side.
  *
  * @module features/analytics/api
  */
-import type { AnalyticsCoverage, AnalyticsCoverageFilters } from './analytics-coverage.types';
+import type {
+  AnalyticsCoverage,
+  AnalyticsCoverageByConnection,
+  AnalyticsCoverageFilters,
+} from './analytics-coverage.types';
 
 export interface AnalyticsCoverageApi {
   getCoverage: (filters: AnalyticsCoverageFilters) => Promise<AnalyticsCoverage>;
+  getCoverageByConnection: (filters: AnalyticsCoverageFilters) => Promise<AnalyticsCoverageByConnection>;
 }
 
 interface ApiRequest {
@@ -28,5 +36,7 @@ function buildQuery(filters: AnalyticsCoverageFilters): string {
 export function createAnalyticsCoverageApi(request: ApiRequest): AnalyticsCoverageApi {
   return {
     getCoverage: (filters) => request(`/analytics/coverage?${buildQuery(filters)}`),
+    getCoverageByConnection: (filters) =>
+      request(`/analytics/coverage/by-connection?${buildQuery(filters)}`),
   };
 }
--- a/apps/web/src/features/analytics/api/analytics-coverage.query-keys.ts
+++ b/apps/web/src/features/analytics/api/analytics-coverage.query-keys.ts
@@ -3,4 +3,6 @@ import type { AnalyticsCoverageFilters } from './analytics-coverage.types';
 export const analyticsCoverageQueryKeys = {
   all: ['analytics', 'coverage'] as const,
   coverage: (filters: AnalyticsCoverageFilters) => ['analytics', 'coverage', filters] as const,
+  byConnection: (filters: AnalyticsCoverageFilters) =>
+    ['analytics', 'coverage', 'by-connection', filters] as const,
 };
--- a/apps/web/src/features/analytics/api/analytics-coverage.types.ts
+++ b/apps/web/src/features/analytics/api/analytics-coverage.types.ts
@@ -47,3 +47,25 @@ export interface AnalyticsCoverageFilters {
   to: string;
   sourceConnectionId?: string;
 }
+
+/**
+ * Mirrors `AnalyticsCoverageByConnectionResponseDto` / `CoverageCategoryConnectionRowsDto`
+ * (`GET /analytics/coverage/by-connection`, #2713) — one entry per category
+ * (`'currency' | 'tax-a' | 'tax-b' | 'tax-c'`, never `'product-matching'`),
+ * each carrying the affected-order count per `sourceConnectionId`. Replaces
+ * `useCoverageCrossReferenceQuery`'s client-side page-draining grouping for
+ * `ChannelSalesTable` (#2714).
+ */
+export interface CoverageConnectionRow {
+  sourceConnectionId: string;
+  affectedCount: number;
+}
+
+export interface CoverageCategoryConnectionRows {
+  category: CoverageCategory;
+  rows: CoverageConnectionRow[];
+}
+
+export interface AnalyticsCoverageByConnection {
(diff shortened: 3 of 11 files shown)


norbert-kulus-blockydevs:
## Review verdict: Approved

Reviewed against `docs/frontend-architecture.md`, `docs/frontend-ui-style-guide.md`, `docs/engineering-standards.md` and `docs/code-review-guide.md`. This is a small, well-scoped frontend-only swap (client-side page-draining cross-reference -> a single aggregate endpoint call), stacked on the not-yet-merged `2713-coverage-by-connection-reads` branch (PR #2810), which is a documented and acceptable pattern for this repo (see prior stacked-branch precedent in KSeF/analytics epics).

No BLOCKING or IMPORTANT findings.

### SUGGESTION

- `apps/web/src/features/analytics/hooks/use-coverage-by-connection-query.ts` - the new hook intentionally mirrors `useCoverageCrossReferenceQuery`'s "only `data`, no `isLoading`/`isError`" silent-degradation contract, which is consistent with the existing pattern and documented as deliberate in the doc comment. No action needed, just flagging that this is a repeated pattern now shared by two hooks - if a third coverage-aggregate hook is added later, it might be worth extracting the shared `useQuery` wrapper (return `{ data }` only) into one small helper rather than copy-pasting a third time. Not worth doing now for two call sites.
- `docs/plans/implementation-plan-coverage-by-connection-fe.md` - fine to include per the repo's implementation-plan convention, just noting it duplicates a fair amount of the final diff verbatim (code blocks match the shipped code almost exactly). Not a problem, just an observation - keeping the plan in sync with any future changes to this feature is the reviewer's job per `docs/frontend-architecture.md`'s "UX Mockups" analog for plans (no automated check exists for implementation plans going stale, only for UX mockups).

### Positive observations

- Correct scope discipline: `ProductSalesTable` / `buildProductExclusionMap` / `useCoverageCrossReferenceQuery` are explicitly left untouched, with the reasoning (no product-level aggregate exists) stated inline and in the PR description, and verified by reading `product-exclusion-map.lib.ts` - it only imports `CoverageOrderLite`, `CrossReferenceableCategory`, `CROSS_REFERENCEABLE_CATEGORIES` from the touched file, none of which changed shape, so there is no blast radius onto the sibling table.
- `buildChannelExclusionMap`'s signature was reshaped in place rather than duplicated into a second function, matching the single-caller justification in the code comment and the file's own doc-comment update explaining why grouping now happens server-side (`.set()` instead of accumulate, since the backend's own `GROUP BY sourceConnectionId` already guarantees at most one row per `(category, sourceConnectionId)` pair).
- Test coverage is solid: the reshaped `buildChannelExclusionMap` gets 5 new unit test cases (undefined input, all-clear, multi-connection grouping, category-to-connection attribution, count pass-through), and the existing `.excl-note` regression tests in `channel-sales-table.test.tsx` are ported with byte-identical assertions on rendered output, only the mocked API call changed - exactly matching the stated acceptance criterion.
- `test-utils.tsx`'s `createMockApiClient` got a default `getCoverageByConnection` mock so every other test rendering `ChannelSalesTable` without explicitly mocking the new call doesn't break - this is exactly the kind of blast-radius check that's easy to miss and wasn't.
- Correctly follows FE architecture conventions: hook returns only `{ data }` via TanStack Query, no raw `fetch`, all state ownership rules respected (server state via `useQuery`, no new global/local state), naming conventions match (`use-*.ts`, `*.lib.ts`, `*.types.ts`, `*.api.ts`), file headers present on all touched/new files, no `any` usage.
- The gating change from "each category enabled independently" to "any category enabled" for the single combined request is correctly reasoned through in both the code comment and the plan doc, and preserves the pre-existing "zero network calls when Data Coverage reads all-clear" property.
- Types are kept intentionally decoupled: `AnalyticsCoverageByConnection`'s `category` field stays the broad `CoverageCategory` rather than importing the narrower `CrossReferenceableCategory` from the `lib` layer, avoiding a `types.ts -> lib.ts` import in the wrong direction - narrowing happens where it's consumed instead.

No architecture, naming, security, or test-coverage issues found. Backend contract (PR #2810) shape matches what this PR's mocks and types assume (`{ categories: [{ category, rows: [{ sourceConnectionId, affectedCount }] }] }`, one entry per `'currency' | 'tax-a' | 'tax-b' | 'tax-c'`).

norbert-kulus-blockydevs:
Review verdict was Approved with no blocking or important findings - both suggestions (shared hook-pattern extraction candidate, keeping the implementation-plan doc in sync manually) are optional and non-blocking, so no changes made. Good to merge as-is.

Issue #917, opened 2026-05-30
chore(test): guard that every plugin in plugins.ts has a jest-integration moduleNameMapper entry


piotrswierzy:
## Problem

`apps/api/test/jest-integration.cjs` hand-maintains a `moduleNameMapper` that source-maps each `@openlinker/integrations-*` package to its `src/`. This list silently drifts: a plugin can be wired into `apps/api/src/plugins.ts` (and thus into the app module graph) **without** a matching mapper entry. When that happens, **every** integration test fails to load the app module in a worktree with:

```
Cannot find module '@openlinker/integrations-<x>' from 'src/plugins.ts'
```

CI masks this because `test:ci` builds `dist` first (so the package resolves via its `main`), but a fresh worktree (no build) can't run *any* integration test until the mapping is added.

This was discovered in #916 / PR #916: `@openlinker/integrations-inpost` had been added to `plugins.ts` without the corresponding `jest-integration.cjs` entry, blocking the carrier-mapping int-spec. That PR added the missing `inpost` mapping as a one-off; this issue is the durable guard so the next plugin can't reintroduce the gap.

## Proposed fix

Add a lint-time (or unit-test) invariant that cross-checks the two lists and fails loudly on drift — mirroring the existing `scripts/check-migration-timestamps.mjs` ↔ `scripts/plugin-migration-dirs.json` cross-check pattern:

- Parse the `@openlinker/integrations-*` packages referenced in `apps/api/src/plugins.ts`.
- Parse the `@openlinker/integrations-*` keys in `apps/api/test/jest-integration.cjs`'s `moduleNameMapper`.
- Assert every plugin in the former has a `$` and `/(.*)$` mapping in the latter; fail with the missing package name(s).
- Wire it into `pnpm check:invariants` (→ `pnpm lint`).

Alternative (lighter): a small `*.spec.ts` under `apps/api/test/` that does the same set-difference assertion, so it runs in the normal unit-test gate without a new script.

## Acceptance criteria

- [ ] Removing/forgetting a `moduleNameMapper` entry for a plugin present in `plugins.ts` fails `pnpm lint` (or `pnpm test`) with an actionable message naming the missing package.
- [ ] The guard passes on `main` as-is (inpost mapping already present after #916).
- [ ] Documented in `docs/testing-guide.md` § PrestaShop Testcontainer / harness notes (or wherever the integration-jest config is described).

## Context

[...]

Issue #786, opened 2026-05-18
ci: run worker integration tests to catch AppModule DI regressions


piotrswierzy:
## Problem

The CI `test-integration` job runs only `pnpm --filter @openlinker/api test:integration` (`.github/workflows/ci.yml`). The worker has a full Testcontainers harness at `apps/worker/test/integration/setup.ts` that boots `AppModule` via `NestFactory.createApplicationContext(AppModule, ...)`, but it never runs in CI.

This is the same gap that allowed **#744** to ship — and another worker DI-resolution bug just slipped through (see PR for this issue): `WorkerContentModule` couldn't resolve `Symbol(AiCompletionPort)` after #737. Manual `pnpm start:dev:worker` caught it; CI did not.

## Why the pattern keeps repeating

Worker `AppModule` is wired by hand. Every time a new core/plugin module adds a constructor dep that needs to be present in both apps, the API gets the update (because API integration tests boot `AppModule`) and the worker often doesn't. The unit-level structural test added in the fix PR locks down the specific wiring for `ContentSuggestionService` but won't catch the next sibling DI failure.

## Proposed change

Extend `.github/workflows/ci.yml` `test-integration` job to also run:

```yaml
- name: Run worker integration tests
  run: pnpm --filter @openlinker/worker test:integration
  timeout-minutes: 15
```

The worker harness already uses Testcontainers (Postgres + Redis) the same way the API's does, so the runner already has Docker available.

## Acceptance criteria

- [ ] `.github/workflows/ci.yml` runs `pnpm --filter @openlinker/worker test:integration` in the `test-integration` job.
- [ ] First green CI run on a follow-up branch confirms the worker harness boots in the GitHub Actions runner (the API harness already does; worker is identical infra).
- [ ] If the worker harness reveals flakes specific to CI (it has not been exercised there before), open a separate stabilization issue rather than blocking this PR — the goal is *coverage*, not chasing flakes inline.

## Related

- #737 — bulk-aware `marketplace.offer.create` worker handler (introduced the AI port dep that surfaced the bug).
- #744 — prior worker `AppModule` boot defect with the same CI-gap root cause.

PR #2812, opened 2026-09-02
deps(shared): lift sanitize-html exact pin, fix advisory GHSA-g8qq-57p8-ggw5

### LINKED ISSUE #2233: deps: re-check `pnpm audit` against the exact `sanitize-html@2.17.5` pin (blocked on ESM under Jest 29)
## What

`libs/shared` pins `sanitize-html` to **`2.17.5` exactly** (no caret) — on the library that *is* the inbound XSS boundary (`libs/shared/src/html/sanitize-stored-html.ts`, ADR-046 / #2198). This issue exists so the pin is periodically re-checked by someone assigned rather than remembered.

## Why the pin exists

From `2.17.6` `sanitize-html` depends on `htmlparser2@^12`, which is ESM-only. Jest 29 loads the repo's CJS build, so a bump turns every `libs/shared` spec red with a module-resolution error — a symptom that reads like a broken test rather than a deliberate constraint. Lifting the pin therefore requires solving the ESM-under-Jest-29 question in the same change.

Recorded in `docs/lessons.md` § *An exact dependency pin whose reason lives only in a source comment will be lifted by the next upgrade PR*.

## What to do

1. `pnpm audit --json | jq '.advisories'` (or `pnpm audit --audit-level=moderate`) and check whether any advisory applies to `sanitize-html@2.17.5` or its transitive `htmlparser2@^10`.
2. **If an advisory lands on 2.17.5: lift the pin immediately** — it is a liability, not a preference. Expect to solve ESM-under-Jest-29 in the same PR (`transformIgnorePatterns`, or moving `libs/shared` specs to a module-aware runner).
3. If no advisory applies, re-check the ESM story on the current Jest and close-or-defer with the finding written down.

Do **not** relax the pin to a caret without doing step 2 — a passing `pnpm test` after a caret bump means the resolver happened to keep 2.17.5, not that the bump is safe.

## Acceptance

- [ ] `pnpm audit` result against `sanitize-html@2.17.5` recorded on this issue.
- [ ] Either the pin is lifted (with the ESM question solved and `libs/shared` specs green) or the finding + next re-check trigger is written down.
- [ ] `docs/lessons.md` entry and the `sanitize-stored-html.ts` header still match reality.

## Context

- Raised as SUGGESTION 3 in the tech-lead review of PR #2204 (#2193, [ADR-046](https://github.com/openlinker-project/openlinker/blob/main/docs/architecture/adrs/046-adapter-declared-description-format.md)).
- Inbound sanitization boundary: #2198.

norbert-kulus-blockydevs:
## Summary
- `pnpm audit` found GHSA-g8qq-57p8-ggw5 (stored XSS via SVG SMIL URI-list scheme-policy bypass) against `sanitize-html` up to and including `2.17.6`, patched in `2.17.7`. Per the rule in `docs/lessons.md`, the pin is a liability and gets lifted immediately.
- `2.17.7` still depends on `htmlparser2@^12` (ESM-only, and so is its whole transitive closure: `domutils`, `dom-serializer`, `domhandler`, `domelementtype`, `entities`), so the ESM-under-Jest-29 question had to be solved in this same change.
- `libs/shared/jest.config.js` now transforms `.js` files via `babel-jest` (new `libs/shared/babel.config.cjs`), with a `transformIgnorePatterns` override scoped to exactly those 6 packages, matched against pnpm's nested `.pnpm/<pkg>@<version>/...` store layout.
- `sanitize-html` now carries a normal caret range (`^2.17.7`) instead of an exact pin.
- Updated the header comment in `sanitize-stored-html.ts` and the `docs/lessons.md` entry to match the resolved state.

## Test plan
- [x] `libs/shared` full jest suite: 17 suites / 262 tests pass
- [x] `pnpm --filter @openlinker/shared type-check` clean
- [x] `pnpm --filter @openlinker/shared lint` — 0 errors (pre-existing warnings only)
- [x] `pnpm audit --audit-level=moderate` — no advisory against `sanitize-html` anymore
- [x] `node scripts/check-workspace-dep-declarations.mjs` passes

Closes #2233


norbert-kulus-blockydevs:
## Tech-lead review

Verified the `libs/shared` claims in the test plan directly — all check out:
- 17 suites / 262 tests pass in `libs/shared`
- `type-check` and `lint` clean (0 errors)
- `pnpm audit` no longer reports GHSA-g8qq-57p8-ggw5
- `check-workspace-dep-declarations.mjs` passes

But the fix is scoped to `libs/shared/jest.config.js` only, and that's too narrow.

### 🔴 Blocking: bumping `sanitize-html` breaks jest in every OTHER package that transitively imports `@openlinker/shared/html`

`libs/core`, `apps/api`, and `apps/worker` all run tests through **ts-jest with no `transformIgnorePatterns` override** (they never needed one before, since `sanitize-html@2.17.5` resolved a CJS `htmlparser2@^10`). Once `sanitize-html` moves to `^2.17.7` (ESM-only `htmlparser2@^12` transitive closure), any spec in those packages that imports `@openlinker/shared/html` — directly or via a module barrel — now fails with `SyntaxError: Cannot use import statement outside a module`, the exact failure this PR's own `docs/lessons.md` entry describes.

Confirmed by running against this exact commit:

```
$ pnpm --filter @openlinker/core exec jest content/application/services/content-draft.service.spec.ts
FAIL src/content/application/services/content-draft.service.spec.ts
  ● Test suite failed to run
    SyntaxError: Cannot use import statement outside a module
    at .../node_modules/.pnpm/htmlparser2@12.0.0/node_modules/htmlparser2/dist/index.js:1
    at .../sanitize-html@2.17.7/node_modules/sanitize-html/index.js:1:122
    at .../libs/shared/src/html/sanitize-stored-html.ts:71:1
    at .../libs/core/src/content/application/services/content-draft.service.ts:12:1
```

```
$ pnpm --filter @openlinker/api exec jest content/http/__tests__/content.controller.spec.ts
FAIL — same SyntaxError, chain through libs/core/src/content/content.module.ts → products.module.ts → master-product-sync.service.ts → @openlinker/shared/html
```

I reverted just `libs/shared/package.json` + the header comment (reinstalling `sanitize-html@2.17.5`) and re-ran `content-draft.service.spec.ts` in `libs/core` — it passes on the pre-PR pin. So this is a genuine regression introduced by this change, not a pre-existing failure.

`apps/worker` is affected too — several specs import the same chain (`master-product-sync-*.handler.spec.ts`, `worker-content.module.spec.ts`, `scheduler.service.spec.ts`).

**The PR description's test plan only ran `libs/shared`'s own suite**, which is why this didn't surface locally, but the root `pnpm test` (which the quality gate in `CLAUDE.md` requires green before merge) will fail across at least three packages.

### Fix

The same `babel-jest` + `transformIgnorePatterns` treatment needs to land in every jest config that can reach `@openlinker/shared/html` through its module graph — at minimum `libs/core/jest.config.js`, `apps/api/jest.config.js`, `apps/api/test/jest-integration.cjs`, `apps/worker/jest.config.js`, `apps/worker/test/jest-integration.cjs` (and `apps/web`'s config, if it's on Jest rather than Vitest — worth a quick check). Given `apps/api` and `apps/worker` already hand-maintain `moduleNameMapper` lists that mirror each other (see `docs/testing-guide.md § jest-integration moduleNameMapper guard`, #917), it's worth considering whether the `transformIgnorePatterns`/babel config for this ESM closure belongs in a shared root-level jest preset instead of being copy-pasted into 5+ config files — otherwise the next `sanitize-html`/`htmlparser2` bump reintroduces the same gap in a 6th place.

### Suggestion

Before merging, please run `pnpm test` from the repo root (or at least `--filter @openlinker/core --filter @openlinker/api --filter @openlinker/worker`) to confirm all affected packages are covered by the fix, not just `libs/shared`.

---
Everything else in the diff (the `docs/lessons.md` update, the version bump, the sanitize-stored-html.ts header) looks correct and well-reasoned — this is purely a coverage gap in where the jest transform fix was applied.

norbert-kulus-blockydevs:
Addressed the blocking finding: extended the ESM-deps jest fix to every package that can reach `@openlinker/shared/html`, not just `libs/shared`.

## What changed (744c9138f)

- Extracted the `babel-jest` transform + `transformIgnorePatterns` fragment into a shared `jest.esm-deps.cjs` at the repo root, so the fix isn't copy-pasted per-config (per your suggestion re: the next sanitize-html/htmlparser2 bump reopening the gap).
- Wired it into every remaining jest config that transitively imports `@openlinker/shared/html`: `libs/core/jest.config.js`, `apps/api/jest.config.js`, `apps/api/test/jest-integration.cjs`, `apps/worker/jest.config.js`, `apps/worker/test/jest-integration.cjs`.
- Updated `libs/shared/jest.config.js` to consume the shared fragment too, so there's one source of truth.
- Checked `apps/web`: it runs Vitest, not Jest, so it's unaffected — no config needed there.

## Verification

- Ran the exact two repro cases from the review — both now pass:
  - `pnpm --filter @openlinker/core exec jest content/application/services/content-draft.service.spec.ts` ✅
  - `pnpm --filter @openlinker/api exec jest content/http/__tests__/content.controller.spec.ts` ✅
- Also ran the worker specs you flagged (`master-product-sync-*.handler.spec.ts`, `worker-content.module.spec.ts`, `scheduler.service.spec.ts`) — all pass.
- Full unit suites for all 4 affected packages, as requested:
  - `libs/shared`: 17 suites / 262 tests
  - `libs/core`: 273 suites / 3598 tests
  - `apps/api`: 117 suites / 1505 tests
  - `apps/worker`: 44 suites / 588 tests
  - All green — 451 suites / 5953 tests total.
- `type-check` and `lint` clean (0 errors) across all 4 packages.
- `pnpm audit --audit-level=moderate` — still no advisory against `sanitize-html`.
- `node scripts/check-workspace-dep-declarations.mjs` — still passes.

Ready for re-review.

norbert-kulus-blockydevs:
## Tech Lead review

**Approach**: correct direction overall - lifting the pin and solving the ESM/Jest question in the same change matches the rule this PR itself cites from `docs/lessons.md`. The `jest.esm-deps.cjs` generalization in the second commit is good hygiene (one source of truth for the transform + ignore pattern instead of copy-pasting it per config).

### BLOCKING - plugin packages (`libs/integrations/*`, `libs/plugin-sdk`) are not covered and will break

`sanitize-html@2.17.7` pulls `htmlparser2@^12` (ESM-only), and `@openlinker/core`'s barrels eagerly re-export services that import it at module-load time:

- `libs/core/src/products/index.ts` re-exports `MasterProductSyncService`, which imports `sanitizeStoredHtml` from `@openlinker/shared/html`.
- `libs/core/src/content/application/services/content-draft.service.ts` (reachable from `@openlinker/core/content`) does the same.

Any spec file that does a **value** import from either barrel triggers Jest's CJS `require()` of the whole module graph, hitting the ESM-only `htmlparser2` and throwing `SyntaxError: Cannot use import statement outside a module` (the exact symptom this PR's own `docs/lessons.md` entry describes).

Two concrete, already-broken specs after this bump:
```
libs/integrations/woocommerce/src/infrastructure/adapters/product-master/__tests__/woocommerce-product-master.adapter.spec.ts:15
libs/integrations/prestashop/src/infrastructure/adapters/__tests__/prestashop-product-master.adapter.spec.ts:23
```
both do `import { MasterProductNotFoundError } from '@openlinker/core/products';` - a value import, not `import type`.

None of `libs/integrations/*/jest.config.mjs` or `libs/plugin-sdk/jest.config.mjs` were touched in this PR, and none of them route `.js` through `babel-jest` or carry the `transformIgnorePatterns` override. `pnpm test:ci` runs `pnpm -r test`, which includes these packages, so this is a CI-breaking regression, not just a theoretical gap.

Fix: wire the same `esmDepsJsTransform` / `ESM_DEPS_TRANSFORM_IGNORE_PATTERN` from `jest.esm-deps.cjs` into every `libs/integrations/*/jest.config.mjs` and `libs/plugin-sdk/jest.config.mjs` (same shape as the `apps/api` / `apps/worker` / `libs/core` wiring in the second commit, just merged into the `.mjs` `transform`/`transformIgnorePatterns` keys instead of `.js`). Any plugin package whose specs value-import from `@openlinker/core/products`, `@openlinker/core/inventory`, `@openlinker/core/content` or `@openlinker/core/listings` is at risk given the documented cross-context dependency map (`inventory --> products`, `listings --> products`, `content --> products`), so this is worth applying repo-wide rather than only to the two packages named above.

### IMPORTANT - new tooling deps only declared at the workspace root

`babel-jest`, `@babel/core`, `@babel/preset-env` are added only to the root `package.json` `devDependencies`. Every package that actually runs its own `jest`/`ts-jest` (`libs/shared`, `libs/core`, `apps/api`, `apps/worker`) already duplicates `jest` and `ts-jest` in its own `package.json` - this PR breaks that local convention for the three new packages it needs. It likely resolves fine today via Node's directory walk-up to the root `node_modules`, but it's inconsistent with the existing pattern and worth matching for portability.

### SUGGESTION - no regression test for the actual advisory

No test asserts that the GHSA-g8qq-57p8-ggw5 SVG/SMIL payload is stripped by `sanitizeStoredHtml`. `svg`/`animate` are already outside `ALLOWED_TAGS` with `disallowedTagsMode: 'discard'`, so this app's configuration was likely never exploitable - but a one-line spec asserting an SVG/`animate` payload with a `javascript:`-scheme attribute is fully discarded would make that claim self-verifying instead of inferred, and would catch a future allowlist change that reintroduces `svg`.

### Positive notes

- `transformIgnorePatterns` regex correctly targets pnpm's nested `.pnpm/<pkg>@<version>/node_modules/<pkg>/` layout rather than the naive `node_modules/(?!pkg)/` recipe that silently never matches under pnpm - verified against the actual resolved package names in the lockfile bump (`htmlparser2@12.0.0`, `domutils@4.0.2`, `dom-serializer@3.1.1`, `domhandler@6.0.1`, `domelementtype@3.0.0`, `entities`).
- `docs/lessons.md` entry updated in place with a dated resolution rather than left stale, consistent with the CLAUDE.md instruction to keep that file current.
- No behavioral/allowlist change to `sanitize-stored-html.ts` itself - purely the version bump plus tooling, which keeps the risk surface small.

### Merge readiness

Request changes - the plugin-package test breakage is a real CI regression, not a style nit. The other two items are non-blocking.

piotrswierzy (approved):
## Code Review — #2812 (lift `sanitize-html` pin, GHSA-g8qq-57p8-ggw5)

**Scope reviewed** (26 files, +1795/−266, of which ~1585/239 is `pnpm-lock.yaml`): the new root `jest.esm-deps.cjs`, `libs/shared/babel.config.cjs`, 19 jest configs, `libs/shared/package.json`, the `sanitize-stored-html.ts` header, `docs/lessons.md`, and root `package.json`.

### 1. Summary

This is the right change, made for the right reason, in the way `docs/lessons.md` said it would have to be made. An advisory landed on the pinned version of the library that **is** the XSS boundary (`sanitize-stored-html.ts` — the inbound sanitizer feeding the app's only sanctioned `dangerouslySetInnerHTML`), and the lesson entry's own rule is "lift it immediately, and expect to solve the ESM/Jest question in the same change rather than deferring it." That is exactly what happened, and the entry was then rewritten into a resolution rather than left describing a state that no longer exists. Good discipline.

The mechanism is sound. Splitting `'^.+\\.(t|j)s$': 'ts-jest'` into a `.ts` ts-jest rule plus a `.js` babel-jest rule, scoped by a `transformIgnorePatterns` override, is the standard remedy; and the pattern is written against pnpm's real `.pnpm/<pkg>@<version>/node_modules/<pkg>/` layout rather than the naive `node_modules/(?!pkg)/` recipe that never matches under pnpm — the header comment calls that out explicitly, which is the sort of note that saves the next person an hour. Factoring it into one root `jest.esm-deps.cjs` consumed by every config (rather than 19 copies of a regex) is right, and the named-export shape is statically analysable so the `.mjs` configs' named imports resolve correctly.

My concerns are all about the **hand-maintained fan-out**, not the fix.

---

### 2. Findings

**[IMPORTANT]** — `libs/oms/jest.config.mjs` (not in the diff)

Nineteen jest configs were updated. Twenty-one exist. `libs/oms/jest.config.mjs` was not touched, and `libs/oms/package.json` **declares `@openlinker/shared`** — so it is squarely inside the population the new module's own docblock defines:

> *"Any jest config whose module graph can reach `@openlinker/shared/html` must merge this in … or Jest throws 'SyntaxError: Cannot use import statement outside a module'."*

`libs/oms/src/__tests__/no-http-in-dependency-graph.spec.ts` already walks that graph and asserts `@openlinker/shared` is reachable. Whether a current oms spec loads `shared/html` specifically is incidental — it will the day one does, and the failure will read as a broken test rather than a missing config line, which is the exact confusion this whole lesson entry exists about. Either add the two lines, or state in the PR why oms is excluded.

(`libs/test-kit/jest.config.js` is the other untouched config; it declares no `@openlinker/shared` dependency — only peers — so leaving it out is defensible. Worth one sentence either way.)

---

**[IMPORTANT]** — no invariant guard for the fan-out

This repo's settled answer to a hand-maintained per-package list is a `scripts/check-*.mjs` under `pnpm check:invariants`, and the closest precedent is *the same file family*: `scripts/check-jest-integration-mappers.mjs`, which exists because (per `testing-guide.md` § jest-integration `moduleNameMapper` guard) *"this list silently drifts"* and its drift broke every integration test twice (#916, #786). The list added here has the identical property, and finding #1 above is drift that already happened inside the PR that introduced it.

A ~30-line check — every `jest.config.{js,mjs,cjs}` and `jest-integration.cjs` under a package declaring `@openlinker/shared` must require/import `jest.esm-deps.cjs` and spread both exports — turns "a new package's suite dies with an opaque ESM error six months from now" into a lint failure naming the file and the two lines to paste. Given how much of this change is exactly the kind of list the repo already guards, its absence is the main thing I'd like addressed.

---

**[IMPORTANT]** — root `package.json`: `@babel/core`, `@babel/preset-env`, `babel-jest` declared only at the root

`ts-jest` — the tool these three sit beside in every one of those transform blocks — is declared in **all 20** packages that carry a jest config, not just the root. The new babel trio is root-only, so each package's Jest resolves it by walking up to the root `node_modules`, which is the hoisting-dependent resolution `engineering-standards.md` § Workspace dependency declarations warns about (*"leaves the emitted `dist`'s bare specifiers resolving only through `shamefully-hoist`"*). It works today; it is one `pnpm` setting away from not working, and it is inconsistent with the sibling tool in the same config object. Mirror the `ts-jest` treatment: declare all three in each package that consumes `esmDepsJsTransform`.

---

**[SUGGESTION]** — `libs/shared/babel.config.cjs` location

The root-level `jest.esm-deps.cjs` reaches *downward into a package* for its babel config (`path.resolve(__dirname, 'libs/shared/babel.config.cjs')`), and every other package's jest then loads a config file owned by `libs/shared`. The config isn't shared-specific — it is repo-wide tooling for a transitive dependency. Move it to the root beside `jest.esm-deps.cjs` (e.g. `babel.esm-deps.cjs`) so the ownership matches the consumers, and so a future `libs/shared` refactor cannot silently break 19 unrelated suites.

**[SUGGESTION]** — `ESM_DEPS_TRANSFORM_IGNORE_PATTERN` replaces the Jest default rather than extending it

The override is a single-element array, so Jest's own default `transformIgnorePatterns` no longer applies. That is almost certainly intended (the new pattern subsumes `/node_modules/`), but it is worth one line of comment saying so — a reader who later adds a second entry expecting it to be additive will be surprised.

**[SUGGESTION]** — allowlist maintenance

The six package names are duplicated as prose in three places (`jest.esm-deps.cjs`, `babel.config.cjs`, `sanitize-stored-html.ts`, plus `docs/lessons.md`). If `sanitize-html` 2.18 adds a seventh ESM-only transitive, the failure is again an opaque syntax error. The invariant script above is the durable fix; failing that, a pointer comment naming `jest.esm-deps.cjs` as the single source (and the other sites as narrative only) would help.

---

### 3. Positive observations

- The advisory response is correct in substance: `2.17.7` patches GHSA-g8qq-57p8-ggw5, the sanitizer's own allowlist config is untouched, and no behavioural change to the boundary is smuggled in alongside the bump.
- `docs/lessons.md` was updated to a **Resolution** with the mechanism spelled out, not merely marked done — and `sanitize-stored-html.ts`'s header was rewritten in the same commit so the manifest, the source comment and the lesson all now agree. The original lesson was precisely about those three drifting apart; closing it without that would have been ironic.
- The pnpm-layout regex, and the explicit note that the naive recipe does not match a pnpm-nested path, is the non-obvious detail most people get wrong here.
- Every config keeps `.ts` on ts-jest and confines babel to `node_modules`, so no application source changes transformer — the blast radius is as small as it can be.
- No migration, no schema, no runtime code path touched; risk is confined to the test runner.

---

### 4. Final Assessment

**Merge readiness: 🔄 Approve with changes.**

No blocking issue — the security fix is correct and the ESM mechanism is sound and evidenced by a passing `libs/shared` suite. Approving so the advisory doesn't sit open. But the change's weak point is a hand-maintained 19-entry list, and one entry is already missing from it; please close that and, ideally, guard it the way this repo guards its twin.

**Priority fixes**
1. Add `jest.esm-deps.cjs` to `libs/oms/jest.config.mjs` (or document why oms and test-kit are excluded).
2. Add a `scripts/check-jest-esm-deps.mjs` invariant under `pnpm check:invariants`, modelled on `check-jest-integration-mappers.mjs`.
3. Declare `@babel/core`, `@babel/preset-env` and `babel-jest` per-package alongside `ts-jest`, not only at the root.
4. (Optional) Move `babel.config.cjs` to the repo root; note that the `transformIgnorePatterns` override replaces the Jest default.

Note `mergeable_state: blocked` — confirm required checks are green (and that the full `pnpm test:ci` fan-out, not just `libs/shared`, has run) before merging, since this change alters how **every** package's Jest transforms `node_modules`.

norbert-kulus-blockydevs:
Thanks for the review - all five findings are addressed in `988e1fd7d`.

**1. [IMPORTANT] `libs/oms/jest.config.mjs` (missed config)**

Added. `libs/oms` declares `@openlinker/shared`, so it was squarely inside the population, and you are right that the failure would have read as a broken test rather than a missing config line. Verified the config loads and a spec runs green under the new transform.

`libs/test-kit/jest.config.js` stays excluded on purpose: it declares no `@openlinker/shared` dependency (only peers), so it is outside the population the new invariant defines and the invariant does not flag it.

**2. [IMPORTANT] Invariant guard for the fan-out**

Added `scripts/check-jest-esm-deps.mjs`, modelled on `check-jest-integration-mappers.mjs` (including the `--self-check` mode), and chained into `check:invariants` right after it. It walks every workspace package, and for any package whose `package.json` declares `@openlinker/shared` it requires each of that package's jest configs (`jest.config.{js,mjs,cjs}` and `jest-integration.cjs`) to import `jest.esm-deps.cjs` and spread both exports into `transform` and `transformIgnorePatterns`. A miss fails with the file path and the exact lines to paste.

Current state: 20 configs checked, all passing. Deliberately proved it catches the real case by reverting the oms fix and confirming the script fails on it.

**3. [IMPORTANT] Babel deps declared per package**

`@babel/core`, `@babel/preset-env` and `babel-jest` are now in the `devDependencies` of every package that consumes `esmDepsJsTransform`, matching how `ts-jest` is already treated in the same transform block, so resolution no longer depends on hoisting to the root. `node scripts/check-workspace-dep-declarations.mjs` passes (20 packages, 40 refs).

**4. [SUGGESTION] Babel config moved to root**

`libs/shared/babel.config.cjs` is now `babel.esm-deps.cjs` at the repo root, beside `jest.esm-deps.cjs`. Ownership now matches the consumers, and a future `libs/shared` refactor cannot break 20 unrelated suites. No stale references to the old path remain.

**5. [SUGGESTION] `transformIgnorePatterns` replaces rather than extends; allowlist duplication**

`jest.esm-deps.cjs` now carries a comment stating that the single-element array replaces Jest's default `/node_modules/` rather than extending it, that the pattern is written to subsume the default, and that Jest ANDs multiple entries together, so a second pattern added later can silently widen what gets transformed.

The prose copies of the six package names in `babel.esm-deps.cjs`, `sanitize-stored-html.ts` and `docs/lessons.md` now each point at `jest.esm-deps.cjs` as the source of truth and name the invariant script as the thing that keeps the fan-out honest. The `docs/lessons.md` entry also records the missed-config incident itself, since it is a concrete instance of the drift the entry is about.

**Verification**

- `node scripts/check-jest-esm-deps.mjs --self-check` and the real run: pass (20 configs).
- `node scripts/check-workspace-dep-declarations.mjs`: pass.
- `libs/oms` jest config resolves and a spec runs green.
- `pnpm --filter @openlinker/shared type-check`: clean.

Agreed on your closing note about the full `pnpm test:ci` fan-out mattering more than `libs/shared` alone here, since this changes how every package's Jest transforms `node_modules` - leaving that to CI on this push.

piotrswierzy (approved):
## Re-review — #2812 @ `988e1fd` (after `988e1fd7d` "close jest ESM-deps review findings")

Read the prior APPROVE-with-changes review, then the new commit (27 files) against the current head.

**(1) `libs/oms/jest.config.mjs` missed — RESOLVED.** The config now imports from `../../jest.esm-deps.cjs` and spreads both exports into `transform` / `transformIgnorePatterns`. I re-enumerated every jest config on the head: the only ones now *not* carrying the merge are `libs/test-kit/jest.config.js` (declares no `@openlinker/shared` — correctly excluded, and the script's docblock says so in as many words), the root `jest.config.js` (a `projects: [...]` aggregator that delegates to per-package configs — correctly excluded), and the scaffolder template (see below). No genuine package is left out.

**(2) No invariant guard — RESOLVED, and done properly.** `scripts/check-jest-esm-deps.mjs` is modelled on `check-jest-integration-mappers.mjs` as suggested, and is chained into `check:invariants` **with its `--self-check` pair**, matching the convention every other guard in that chain follows. Three details raise it above a box-tick:
- The population is *derived*, not listed — it walks the `pnpm-workspace.yaml` parents, reads each `package.json`, and includes a package iff it declares `@openlinker/shared` (or IS shared). So a new package inherits the rule automatically; there is no second hand-maintained list guarding the first one.
- It checks **use**, not just import: each export must appear ≥2 times, so a config that destructures the fragment and then forgets to spread it still fails. That is the failure mode a naive `includes('jest.esm-deps.cjs')` check would miss.
- The failure output prints the file, the reason, and the exact three lines to paste. That is the whole point of converting an opaque `SyntaxError: Cannot use import statement outside a module` into a lint failure.
- `classifyConfigContent` is pure and `--self-check` exercises four cases including both the `require` and `import` forms. Good.

**(3) Root-only babel trio — RESOLVED.** `@babel/core`, `@babel/preset-env` and `babel-jest` are now declared in all 20 packages that carry a jest config, beside `ts-jest`, exactly mirroring the sibling tool. `pnpm-lock.yaml` reflects it. This removes the `shamefully-hoist`-dependent resolution `engineering-standards.md § Workspace dependency declarations` warns about.

**(4a) `babel.config.cjs` location — RESOLVED.** Moved to the repo root as `babel.esm-deps.cjs`, beside its consumer, with a comment stating it is repo-wide tooling rather than shared-specific. `jest.esm-deps.cjs` resolves it via `path.resolve(__dirname, …)` at the same level, so no package reaches downward into another package for tooling any more.

**(4b) `transformIgnorePatterns` replaces rather than extends — RESOLVED, and better than asked.** The note not only says the override replaces Jest's default, it explains that Jest **ANDs** multiple entries (a file must match all of them to be left untransformed), so a reader adding a second pattern would silently *widen* what gets transformed. That is the actual trap, stated precisely.

**(4c) Package list duplicated as prose — RESOLVED.** The prose copies now point at `jest.esm-deps.cjs` as the source of truth.

All five items closed, including both optionals.

### 2. New findings on the current head

**[IMPORTANT]** — `scripts/create-adapter-templates/jest.config.mjs` + `.../package.json`

The scaffolder template declares `"@openlinker/shared": "workspace:*"` in `dependencies`, and its `jest.config.mjs` does **not** merge `jest.esm-deps.cjs`. So every adapter scaffolded from `scripts/create-adapter.mjs` from now on lands at `libs/integrations/<name>/`, is picked up by the new guard's derived population, and **fails `pnpm lint` on its first run** — with a correct message, but the very first thing a new contributor sees after scaffolding is a red build. (Its spec suite would also hit the raw ESM error the moment one of its specs reaches `@openlinker/shared/html` through `libs/core`.) `scripts/check-create-adapter.mjs` scaffolds into `os.tmpdir()`, outside `apps/`/`libs/`, so the guard does not catch this in CI — it only bites the human.

The fix is the same two lines every other `.mjs` config got (`import … from '../../../jest.esm-deps.cjs'`), plus the three babel devDependencies in the template's `package.json` to match the treatment finding (3) just applied everywhere else. `EXPECTED_FILE_COUNT` in `check-create-adapter.mjs` is unaffected — no file is added or removed.

**[SUGGESTION]** — the guard is content-based (regex over the config text), which is the right cost/benefit here, but it cannot tell a merged-and-then-overridden config from a correct one (e.g. a later `transformIgnorePatterns:` key winning). Not worth solving; noting it so a future reader does not over-trust the green line.

### 3. Final assessment

**Merge readiness: ✅ Approve.** Every prior finding is closed, the invariant is derived rather than listed and self-checked, and the security substance of the change (`sanitize-html@^2.17.7`, allowlist untouched, no behavioural change at the XSS boundary) is unchanged and correct.

**One follow-up before or immediately after merge:** wire `jest.esm-deps.cjs` + the babel devDependencies into `scripts/create-adapter-templates/`, so the scaffolder cannot emit a package that fails the invariant this PR introduces.

As before, confirm the full `pnpm test:ci` fan-out is green — not just `libs/shared` — since this change alters how every package's Jest transforms `node_modules`.

piotrswierzy (approved):
## Re-review at head `2d2e669` (was `988e1fd`) — APPROVE

### Prior finding, verified item-by-item

**IMPORTANT (open at last review): `scripts/create-adapter-templates/` not updated → a freshly scaffolded adapter fails `pnpm lint` out of the box. — RESOLVED.**

Both template files are now in the diff:

- `scripts/create-adapter-templates/jest.config.mjs` imports `ESM_DEPS_TRANSFORM_IGNORE_PATTERN` / `esmDepsJsTransform` from `'../../../jest.esm-deps.cjs'`, spreads `esmDepsJsTransform` into `transform` under `'^.+\\.js$'`, and sets `transformIgnorePatterns: [ESM_DEPS_TRANSFORM_IGNORE_PATTERN]` — the same shape the 20 real configs carry. The `../../../` depth is correct **for the destination** (`libs/integrations/<name>/` → repo root), which is the only place the file is ever executed; it is deliberately wrong in-place, and that is fine because `check-jest-esm-deps.mjs` walks only `WORKSPACE_PARENTS = ['apps','libs','libs/integrations']` and never reaches `scripts/`.
- `scripts/create-adapter-templates/package.json` adds `@babel/core`, `@babel/preset-env`, `babel-jest` to `devDependencies`, so the scaffolded package can actually resolve the transform rather than merely referencing it.
- No new template file was added, so `EXPECTED_FILE_COUNT = 14` in `check-create-adapter.mjs` correctly stays put — no "template added but invariant not bumped" drift.

The scaffolded adapter's `package.json` declares `@openlinker/shared` as a runtime dependency, so once it lands in `libs/integrations/` the guard's `reachesSharedHtml()` picks it up and both halves of the rule are already satisfied — the guard and the template now agree by construction rather than by hand.

### Step 3/4 assessment of the current diff

- **Guard design.** `check-jest-esm-deps.mjs` follows the `check-jest-integration-mappers.mjs` (#917) precedent the repo established: pure classifier + `--self-check`, discovery derived from `package.json` rather than a hand-maintained list, chained into `check:invariants` (verified in the root `package.json` diff, inserted right after `check-jest-integration-mappers`). The duplicate-key check (c) is the part worth calling out as genuinely load-bearing — a second `transformIgnorePatterns:` in a plain object literal silently wins, so a config could satisfy (a) and (b) and still be broken; catching the *ambiguity* rather than trying to decide which declaration is real is the right call for a regex-based guard, and the `shadowedByDuplicateKey` self-check case pins it.
- **`jest.esm-deps.cjs`.** The docblock correctly warns that the exported pattern *replaces* rather than extends Jest's default `transformIgnorePatterns`, and that Jest ANDs multiple entries — exactly the trap a future reader adding a second pattern would fall into. Placement at the repo root beside `babel.esm-deps.cjs` is right: it is consumed by 20 configs across `apps/`, `libs/`, `libs/integrations/` and the scaffold template, not by `libs/shared` alone. The `(?:\.pnpm/)?…(?:@[^/]+)?` shape matches pnpm's nested store layout, which is the whole reason a naive `node_modules/(?!htmlparser2)` would silently not fire here.
- **Pin lifted correctly.** `sanitize-html` moves from the exact `2.17.5` to `^2.17.7`, closing GHSA-g8qq-57p8-ggw5. This is the behaviour `docs/lessons.md` prescribes for a pin whose reason lived only in a source comment, and the lessons entry + the `sanitize-stored-html.ts` header are both updated to the resolved state rather than left describing the old world — which is what makes the next upgrade PR cheap instead of a re-litigation.
- **Blast radius.** 20 jest configs + 20 `package.json` devDependency additions is a wide but mechanical change; the guard is what keeps it from rotting. Application `.ts` stays on `ts-jest` — babel only touches `node_modules` for the six carved-out packages, so no source-level semantics change.
- **CI.** All 9 checks green at this head, including `Lint` (which runs `check:invariants`, i.e. the new guard and its self-check), `Test`, `Integration Tests`, `Build`, `Docker Build Smoke Test` and `Scaffolded Adapter Builds`.

### Remaining (non-blocking) note

`scripts/create-adapter-templates/jest.config.mjs` is outside every automated guard's reach: `check-jest-esm-deps.mjs` skips `scripts/` by design, and `check-create-adapter.mjs` only asserts file-set/token-substitution, never runs the scaffolded suite. So if the exported names in `jest.esm-deps.cjs` are ever renamed, the 20 real configs fail the build and the template silently does not — the same drift class this PR's guard exists to close, one level out. Not worth blocking on (the scaffolder is low-traffic and `scaffold-smoke.yml` would catch a hard break), but a two-line extension of the guard's discovery to include the template dir would close it cheaply if someone touches this area again.

Nothing blocking. Ship it.

norbert-kulus-blockydevs:
Addressed the remaining non-blocking note from the latest review: `scripts/check-jest-esm-deps.mjs` now also discovers `scripts/create-adapter-templates/` (added as an explicit extra package dir, since it's not a workspace package under `pnpm-workspace.yaml`). The template's `jest.config.mjs` already carried the correct merge, so the guard now reports 21/21 configs green instead of 20/20 — a future rename of `jest.esm-deps.cjs`'s exports will now fail the scaffold template too, instead of only the 20 real packages.

Pushed as c376c7d05.

piotrswierzy (approved):
## Re-review at head `36d2a24` (was `2d2e669`) — APPROVE

### 1. Prior open item, verified

**Non-blocking note from the last review: `scripts/create-adapter-templates/jest.config.mjs` sits outside every automated guard's reach, so renaming `jest.esm-deps.cjs`'s exports would fail the 20 real configs and silently *not* the template — RESOLVED.**

`36d2a24` is the only new commit on the branch, and it does exactly the two-line extension I described. `check-jest-esm-deps.mjs` gains `EXTRA_PACKAGE_DIRS = ['scripts/create-adapter-templates']`, folded into `discoverAllPackageDirs()` behind a `statSync(..., { throwIfNoEntry: false })` existence test so a future template relocation degrades to "not checked" rather than throwing. The comment above it states *why* the entry is explicit — the template is not a workspace package and no `WORKSPACE_PARENTS` glob reaches it, but its config is copied verbatim into `libs/integrations/<name>/` where the guard *would* check it — which is the part a reader would otherwise have to reconstruct.

I verified this behaviourally rather than by reading, against a clean extraction of the PR head:

- `node scripts/check-jest-esm-deps.mjs` → **21 configs checked, all green** (was 20; the template is genuinely in the population now, not merely listed).
- `--self-check` → **5 cases pass** (was 4).
- Negative test: deleting the `jest.esm-deps.cjs` import line from the template config makes the guard fail with `scripts/create-adapter-templates/jest.config.mjs — does not require/import jest.esm-deps.cjs`, plus the three-line paste-ready fix. So the guard fires on the template for the same reason and with the same message it fires on a real package.

`reachesSharedHtml()` needed no change: the template's `package.json` declares `"@openlinker/shared": "workspace:*"` in `dependencies`, so it satisfies the derived-population rule on its own terms rather than through a special case. The guard and the template now agree by construction.

### 2. Step 3 / Step 4 assessment of the current diff

The rest of the change is unchanged from the head I approved at `2d2e669`, and my assessment stands: the pin lift to `^2.17.7` closes GHSA-g8qq-57p8-ggw5 with the sanitizer's own allowlist untouched; the ts-jest/babel-jest split confines babel to `node_modules` for exactly six carved-out packages so no application `.ts` changes transformer; `jest.esm-deps.cjs` and `babel.esm-deps.cjs` sit at the root beside each other, matching where their 21 consumers are; the pnpm `.pnpm/<pkg>@<version>/` regex shape is the reason the naive `node_modules/(?!htmlparser2)` recipe would silently not fire; and `docs/lessons.md` plus the `sanitize-stored-html.ts` header were both rewritten to the resolved state, which is what the original lesson (three sites drifting apart) was about.

One point about the guard's classifier, which I only found by mis-running my own negative test and is worth recording rather than fixing: `countOccurrences` builds `new RegExp(name, 'g')` unanchored, so `esmDepsJsTransformRENAMED` counts as two occurrences of `esmDepsJsTransform` and a rename-by-suffix passes. It cannot mask a real defect — the import-path check (a) still has to match, and a config that imports the real module and spreads a name that does not exist fails at runtime immediately and loudly — so this is a note for a future reader who over-trusts the green line, not a request. Word-boundary anchors would close it for one character each if anyone touches the file again.

### 3. CI

Green at this head: `Type Check`, `Docker Build Smoke Test`, `Scaffolded Adapter Builds`, `PHP Unit Tests`, `Detect changed paths`. `Lint`, `Build` and `Integration Tests` are queued and `Test` in progress. The prior head was green on all nine, and the delta is a 13-line, test-only script change whose own guard and self-check I ran locally against this head — but since `Lint` is what executes `check:invariants`, confirm it lands green before merging, along with the full fan-out rather than `libs/shared` alone, because this change still alters how every package's Jest transforms `node_modules`.

### 4. Final assessment

✅ **Approve.** Every finding across three review rounds is closed, and this last one closed the guard's own blind spot — the drift class the guard exists to prevent, one level out — with the population derived rather than listed at both levels. Nothing outstanding. Ship it once the queued checks land.

piotrswierzy (approved):
## Tech Lead re-review — #2812 (head `93e6503`, force-pushed from `36d2a24`)

### What the force-push changed: nothing

I diffed the PR's own content against its merge base on both sides and byte-compared the results:

```
old base 178538ca (merge-base origin/main 36d2a24)  →  36d2a24
new base df46c641                                    →  93e6503
diff /tmp/old.diff /tmp/new.diff  →  IDENTICAL CONTENT — pure rebase
```

47 files, +2344/−247 on both sides, same six commits by subject, same stat lines file-for-file. This was a rebase onto current `main` and nothing else — no new code, no silent amendment riding along under cover of the force-push. My previous APPROVE therefore stands on its own reasoning; nothing needed re-deriving.

### Prior findings — item by item

**Guard discovery missed the scaffold template — ✅ CLOSED, verified on this head**

`scripts/check-jest-esm-deps.mjs:68` — `const EXTRA_PACKAGE_DIRS = ['scripts/create-adapter-templates'];`, consumed in the discovery loop at :99. That is the right shape: the template is not a workspace package (it has no `pnpm-workspace.yaml` entry, deliberately — it's the *source* a scaffolded package is generated from), so walking the workspace could never have found it, and an explicit extra-dirs list says so rather than loosening the glob and picking up whatever else lands under `scripts/`. `scripts/create-adapter-templates/{jest.config.mjs,package.json}` carry the merge, so a newly scaffolded adapter is born ESM-correct instead of failing on its first `sanitize-html`-touching test — which is exactly the drift the guard exists to catch, one generation removed.

**`countOccurrences` unanchored regex — note only, unchanged, still fine**

`:174` still does `(content.match(new RegExp(name, 'g')) ?? []).length`, so a rename by suffix (`esmDepsJsTransformV2`) keeps satisfying the `< 2` check. Worth knowing; not worth blocking. The failure direction is a guard that stays quiet through a rename, not one that fires falsely, and the adjacent key check at :193 *is* anchored (`^\s*(?:'key'|"key"|key)\s*:` with `gm`) — so the structurally important assertion already has the stricter matcher. If this ever bites, the fix is one `\\b` on each side.

### CI — the item that was still queued last time

`Lint` (which runs `check:invariants`, and therefore the new `check-jest-esm-deps.mjs` guard) is now **completed / success** on `93e6503`. All 9 checks green: Lint, Type Check, Test, Integration Tests, Build, Docker Build Smoke Test, PHP Unit Tests, **Scaffolded Adapter Builds**, Detect changed paths. That last one is the one I most wanted to see, since it exercises the template the final commit touched.

The guard is now proven to run in CI rather than only locally, which was the last thing standing between "this works on the author's machine" and "this cannot silently regress".

### Standing assessment (unchanged, re-stated for the record)

- The advisory (GHSA-g8qq-57p8-ggw5, stored XSS via SVG SMIL URI-list scheme-policy bypass) is genuinely closed by `2.17.7`, and the exact pin is lifted to `^2.17.7` per the `docs/lessons.md` rule that a pin is a liability. `sanitize-stored-html.ts`'s header comment and the lessons entry were both updated to the resolved state, so neither describes a constraint that no longer exists.
- The ESM-under-Jest-29 fix is scoped to exactly the 6 packages in `htmlparser2@^12`'s closure, matched against pnpm's `.pnpm/<pkg>@<version>/...` nested layout rather than a bare package name — which is what keeps `transformIgnorePatterns` from quietly un-ignoring half of `node_modules` and turning every suite into a babel run.
- `sanitizeStoredHtml` remains the XSS boundary (wider than ADR-046's publish-time `applyDescriptionFormat`, per `docs/architecture-overview.md` § Listings), so keeping it patched is load-bearing rather than hygiene.

### Merge readiness

✅ **Approve.** Branch is `behind` main — rebase/update before merging, as ever.

PR #2824, opened 2026-09-02
feat(orders): wire buyer tax-ID onto Allegro, WooCommerce, and Erli orders

### LINKED ISSUE #2822: [TASK] Integration — Wire buyer tax-ID onto Allegro and WooCommerce order sources
## Problem / Context

`SalesDocumentOrderFacts.buyerHasTaxId` (#2599) is real and wired end to end into the rule engine (`evaluateSalesDocumentRules`, `libs/core/src/invoicing/application/mappers/order-to-sales-document-order-facts.mapper.ts`) — a `buyerHasTaxId` rule condition can match a real order today. But source coverage is partial:

- **PrestaShop** (`libs/integrations/prestashop/src/infrastructure/adapters/prestashop-order-source.adapter.ts:539`) — supplies it from `ps_address.vat_number` via `readSourceBuyerTaxId`, populating `IncomingOrderAddress.taxId` on `billingAddress` (three states: not asserted / asserted-none / real id — see `libs/core/src/orders/domain/types/buyer-tax-id.types.ts`).
- **Allegro** (`libs/integrations/allegro/src/infrastructure/adapters/allegro-order-source.adapter.ts:317`, `getOrder`) — `AllegroCheckoutForm` (`libs/integrations/allegro/src/domain/types/allegro-api.types.ts:22`) does not model the checkout form's invoice/company-tax-id block, so it is never read and never mapped onto `billingAddress.taxId`.
- **WooCommerce** (`libs/integrations/woocommerce/src/infrastructure/adapters/woocommerce-order-source.adapter.ts:115`, `getOrder`) — the `billing` block has no tax field at all in the base WooCommerce REST API.
- **Erli** — not an order source at all currently (no `OrderSourcePort` adapter), so not applicable yet, but must be covered if/when one ships.

This was found while correcting stale FE copy (`sales-document-rule-composer-dialog.tsx`, `sales-document-template-screen.tsx`) that claimed "OpenLinker doesn't know a buyer's tax-ID status for **any** order today" — false since #2599 for PrestaShop, but still true for Allegro/WooCommerce. Leaving those two sources unwired means the Poland starter template (and any operator-authored `buyerHasTaxId` rule) can only ever match a PrestaShop-sourced order, silently.

## Proposed Solution

1. **Allegro**: add the checkout form's invoice/company-tax-id field to `AllegroCheckoutForm` (`libs/integrations/allegro/src/domain/types/allegro-api.types.ts`), then in `AllegroOrderSourceAdapter.getOrder` thread it through to `IncomingOrder.billingAddress.taxId` following the exact three-state contract PrestaShop already established (`undefined` = not asserted, `null`/`''` = asserted-none, otherwise the id verbatim — never validated or normalized, per ADR-026's country-agnostic rule). Mirror the `readSourceBuyerTaxId`-shaped helper pattern from the PrestaShop adapter rather than inlining ad-hoc logic.
2. **WooCommerce**: investigate whether a common tax-ID meta field / plugin convention exists on the `billing` block (e.g. EU VAT Number plugin `meta_data` keys) worth reading in `WooCommerceOrderSourceAdapter`. If genuinely absent from the base REST API response, document the gap explicitly in the adapter (a comment stating why `taxId` is never populated) rather than leaving it silently unread with no trace.
3. **Erli**: no adapter exists yet — no code change here. When an `OrderSourcePort` adapter is built for Erli, buyer tax-ID capture must be included in its initial scope rather than shipped as a follow-up gap. Tracked as a note for that future issue, not actionable now.

## Classification

**Type**: Integration
**Layer**: Infrastructure (adapters), Domain (Allegro API types)
**File(s)**:
- `libs/integrations/allegro/src/domain/types/allegro-api.types.ts`
- `libs/integrations/allegro/src/infrastructure/adapters/allegro-order-source.adapter.ts`
- `libs/core/src/orders/domain/types/buyer-tax-id.types.ts` (reference only, no change expected)

## Dependencies

- #2599 (buyer tax-id contract + PrestaShop wiring) — already merged, this issue extends it.

## Assumptions

- Allegro's checkout-form invoice block does carry a company tax id field on the sandbox/production API; the exact field name needs confirming against a live checkout form response before implementation (not yet verified in this issue).
- "Genuinely absent" for WooCommerce means the stock WooCommerce REST API `billing` object with no site-specific plugin assumed — if a specific plugin convention is found, it should be read; otherwise documented as unsupported.

## Acceptance Criteria

- [ ] `AllegroCheckoutForm` models the invoice/company-tax-id block from the checkout form response
- [ ] `AllegroOrderSourceAdapter.getOrder` populates `IncomingOrder.billingAddress.taxId` per the three-state contract (undefined/null-empty/real id), unvalidated and unnormalized
- [ ] WooCommerce investigation is documented: either a real field is wired into `WooCommerceOrderSourceAdapter`, or the gap is stated explicitly in code (comment) and in this issue's resolution
- [ ] No validation/normalization of the tax id value is introduced in either adapter (ADR-026 compliance)
- [ ] Tests added or updated for non-trivial logic (Allegro adapter unit test covering all three tax-id states; WooCommerce adapter test if a field is wired)
- [ ] Once Allegro (and WooCommerce, if wired) report the fact, `sales-document-rule-composer-dialog.tsx`'s `BuyerTaxIdCoverageWarning` and `sales-document-template-screen.tsx`'s PL-template coverage `Alert` are updated to reflect actual per-source coverage, or removed if all applicable sources are covered

## Non-goals

- No validation/format-checking of the tax id itself (ADR-026 keeps national specifics out of `libs/core`).
- No change to the `sales-documents` rule-engine contract — `buyerHasTaxId` already accepts the three-state shape; this is purely about populating it from more sources.
- No Erli `OrderSourcePort` adapter work — only a scope note for when that adapter is eventually built.

norbert-kulus-blockydevs:
## Summary

Closes #2822. `SalesDocumentOrderFacts.buyerHasTaxId` (#2599) already flows
end to end into the routing rule engine, but only PrestaShop populated
`IncomingOrderAddress.taxId`. This wires all three remaining order sources —
live research (not just in-repo types) found a real field on each:

- **Allegro** — `invoice.address.company` on the checkout-form resource
  (`ids[]` array preferred, deprecated flat `taxId` as fallback). Modeled on
  `AllegroCheckoutForm` and mapped via `readSourceBuyerTaxId`, exactly like
  PrestaShop's existing wiring.
- **WooCommerce** — WC core has no native VAT field, so this reads a
  short, documented best-effort allowlist of known VAT-plugin `meta_data`
  keys (Aelia "EU VAT Number" and others), first match wins, absent = unknown.
- **Erli** — the wire type (`ErliOrderAddress.nip`) already existed; the
  mapper simply never read it. One-line fix in an already-shipped,
  production adapter (the issue's premise that Erli "has no OrderSourcePort
  adapter" was stale/false).

Also corrects the sales-documents FE copy this PR's base (#2756) introduces:
the coverage-warning `Alert` in `sales-document-rules-list.tsx` (both the
banner and the per-rule tag) and the PL-template warning in
`sales-document-template-screen.tsx` claimed the condition was
PrestaShop-only / unreachable everywhere. It's now stated as conditional
per source (always for PrestaShop, conditional on a VAT invoice for
Allegro/Erli, conditional on a VAT plugin for WooCommerce).

**Stacked on #2756** (`phase/sdux-m7-settings-dialog`) rather than `main` —
this PR's FE cleanup edits the exact copy #2756 introduces, so it's opened
as a stack element rather than a competing top-level PR against the same
files. Merge #2756 first.

## Test plan

- [x] `pnpm --filter @openlinker/integrations-allegro lint && type-check && test` — 639/639
- [x] `pnpm --filter @openlinker/integrations-woocommerce lint && type-check && test` — 526/526
- [x] `pnpm --filter @openlinker/integrations-erli lint && type-check && test` — 501/501
- [x] `pnpm --filter @openlinker/web lint && type-check && test` — 4147/4148 (1 pre-existing skip)
- [x] `pnpm check:invariants` — all green
- [x] End-to-end verification against `ol-demo-fresh` — see #2822 for full
      detail:
  - [x] WooCommerce: real order, real VAT meta key, `buyerTaxId` round-tripped
  - [x] Allegro: real sandbox purchase with VAT invoice, `buyerTaxId` round-tripped
  - [x] `buyerHasTaxId` routing rule actually MATCHES a real order: the
        Allegro order above satisfies an existing PL rule
        (`buyerHasTaxId=true AND total>=1000 → invoice`), and the auto-issue
        gate dispatched `invoicing.issue` keyed to exactly that rule's
        connection — proven via `sync_jobs`, not just a populated column
  - [ ] Erli: unit-tested only (28/28 passing, including a regression test
        that a `nip` on `deliveryAddress` never leaks). Live verification
        attempted and blocked: Erli's sandbox seller-panel login is gated by
        a reCAPTCHA challenge that fails identically headless and
        headed-under-Xvfb, and no pre-existing sandbox API key is available
        on this machine. Documented gap, not a code-confidence gap — see
        #2822 for the full writeup.


norbert-kulus-blockydevs:
## `/pr-review` — PR #2824

**Scope reviewed**: the 11-file diff against this PR's actual base (`phase/sdux-m7-settings-dialog`, PR #2756) — Allegro/WooCommerce/Erli `taxId` wiring + the sales-documents FE copy correction. The stacked commits from #2545–#2550 belong to #2756 and are out of scope here.

### Summary

Clean, well-tested feature: wires `IncomingOrderAddress.taxId` for Allegro, WooCommerce, and Erli, closing the gap `readSourceBuyerTaxId`/`buyerHasTaxId` (#2599) already supported end-to-end but only PrestaShop fed. All three sources go through the shared `readSourceBuyerTaxId` coercer, so the three-state contract (`undefined` unknown / `null` asserted-none / string) is preserved everywhere — no source here writes `null`, which is correct since none of the three positively asserts "buyer has no tax id," only "didn't say." Test coverage is solid: happy path, absent-field, blank-value, and (Allegro) the deprecated-vs-current wire shape are all covered.

### 🟡 IMPORTANT

**`AllegroOrderSourceAdapter.resolveBillingAddress` can still return the "blank stub" object it exists to prevent** — `allegro-order-source.adapter.ts:322-338`.

The new regression test ("should carry the resolved shipping street data onto billingAddress…") documents the exact failure mode this method is designed to avoid: a `billingAddress` that is truthy but carries blank `address1`/`city`/`postalCode`/`country` defeats a caller's `billingAddress ?? shippingAddress` fallback (confirmed downstream: `order-from-ready-snapshot.ts`'s `firstUsableAddress` treats any non-redacted truthy address as usable — it doesn't check for blank required fields, only for PII redaction).

That guarantee only holds when `resolvedShippingAddress` is defined. `resolveShippingAddress` (same file, line ~863) *can* return `undefined` — when `delivery.address`, `delivery.pickupPoint.address`, and `buyer.address` are all absent/empty. In that (admittedly narrow) case, if the checkout form still carries `invoice.address.company` (VAT invoice requested), the fallback literal on line 334 (`{ address1: '', city: '', postalCode: '', country: '' }`) is used verbatim — reproducing the same blank-but-truthy stub the rest of this method was written to eliminate.

Suggest: when `shippingAddress` is `undefined`, return `undefined` for `billingAddress` too (drop the company/taxId in that case) rather than constructing the stub — consistent with "never a standalone object with blank required fields." A one-line change plus a regression test mirroring the existing one but with no delivery/pickup/buyer address at all would close it.

### 🟢 SUGGESTION

- `woocommerce-order-source.adapter.ts:603,618` — `readBillingTaxIdFromMetaData` and `mapBillingAddress` both re-declare the `{ key: string; value: string }[]` shape inline instead of importing the already-exported `WooCommerceOrderMetaData` type from `order-source/woocommerce-order.types.ts`. Structurally identical so it type-checks fine, but two names for one shape invites drift if the wire type grows a field later.
- `describe-sales-document-tax-id-coverage.ts` doc comment and the `SalesDocumentRulesList`/`SalesDocumentTemplateScreen` copy are all consistent with each other and with the new per-source truth (PrestaShop always / Allegro+Erli conditional-on-VAT-invoice / WooCommerce conditional-on-plugin) — good catch that the prior wording ("Allegro or WooCommerce carries no buyer tax ID at all") was stale even before this PR landed the wiring.

### Verified, no issue found

- Erli: `nip` is read only via the new `isInvoiceAddress` option, never leaking onto `shippingAddress` even for a hypothetical malformed payload (explicit regression test) — correct, since `ErliOrderAddress` is one shared shape for both delivery and invoice addresses.
- `order-to-sales-document-order-facts.mapper.ts`'s `readBuyerTaxId` (billing-first, shipping-fallback) consumes exactly what these three adapters now populate — no drift between the producer and consumer sides of the #2599 contract.
- No schema/migration changes needed (correct — nothing here touches an ORM entity).
- Cross-context/package import of `readSourceBuyerTaxId` from `@openlinker/core/orders`'s top-level barrel in all three plugin packages follows the existing precedent (e.g. `PAYMENT_STATUS` was already imported the same way in the Allegro adapter) — not a barrel-purity violation.

### Verdict

🔄 **Approve with changes** — the IMPORTANT item is a narrow, pre-existing-shaped edge case (only reachable when Allegro's checkout form has *no* address anywhere at all, a condition that also predates this PR for other paths), not a regression introduced by the tested/common flow, so it doesn't block. Worth fixing before merge or as an immediate fast-follow given the PR itself demonstrates awareness of exactly this failure mode.

piotrswierzy (approved):
## Code Review — #2824 (buyer tax-ID on Allegro / WooCommerce / Erli)

**Scope reviewed** (11 files, +404/−32): Allegro order-source adapter + API types, WooCommerce order-source adapter + wire types, Erli order mapper, three spec files, and the sales-documents FE copy.

### 1. Summary

Good work, and it clears the trap this change is most likely to fall into. `readSourceBuyerTaxId` is the correct seam and all three adapters route through it, so a blank / absent field coerces to `undefined` (unknown) rather than `null` (asserted-none) — the exact three-state collapse `buyer-tax-id.types.ts` was written to prevent and that would otherwise make `buyerHasTaxId === false` fire on nearly every B2C order. The Erli `isInvoiceAddress` flag is a nice touch: `mapAddress` is shared with `deliveryAddress`, and the regression test that a `nip` on the delivery side never surfaces is guarding a future payload change rather than today's schema. Live verification through to a matched routing rule and a dispatched `invoicing.issue` (proven via `sync_jobs`, not just a populated column) is exactly the right level of evidence, and the Erli gap is documented honestly rather than papered over.

Three things need attention before this is fully honest about what it changed.

---

### 2. Findings

**[IMPORTANT]** — `apps/web/src/features/sales-documents/components/sales-document-rules-list.tsx:82`, `sales-document-template-screen.tsx:101`, and the `describe-sales-document-tax-id-coverage.ts` docblock

All three now say **"PrestaShop always reports it"**. That is not true, and the core docblock this copy is derived from says so explicitly (`libs/core/src/orders/domain/types/buyer-tax-id.types.ts`, `readSourceBuyerTaxId`):

> *"PrestaShop's `ps_address.vat_number` is optional and empty on essentially every consumer order, so reading a blank as 'asserted to have no tax id' would turn 'the shop never asked' into a positive fact…"*

A PrestaShop B2C order carries a blank `vat_number`, which coerces to `undefined` and does **not** match a `buyerHasTaxId` rule. The whole point of this copy (#2546) is to let an operator judge coverage honestly; telling them PrestaShop always reports it invites them to adopt a template that will then silently route far fewer orders than promised. Suggest: *"PrestaShop reports it when the buyer filled in a VAT number on the address; Allegro and Erli only when the buyer requested a VAT invoice; WooCommerce only if the store runs a supported VAT-number plugin."* Same correction in all three places, since the docblock is what the next person copies from.

---

**[IMPORTANT]** — `libs/integrations/allegro/src/infrastructure/adapters/allegro-order-source.adapter.ts:940-960` (`resolveBillingAddress`)

Allegro previously emitted `billingAddress: undefined` **unconditionally**. It now emits a populated object for every VAT-invoice order, and that field has a second consumer the PR does not mention: `PrestashopOrderProcessorManagerAdapter:337` gates destination-side address provisioning on `if (order.billingAddress && order.customerId)`. So an Allegro→PrestaShop VAT order now additionally calls `resolveOrCreateAddress(..., 'billing', ...)`.

The address alias is `OL-{type}-{hash-prefix}` (`prestashop-address-provisioner.ts:44`), so the billing type mints a **second PrestaShop address row** — and `PrestashopAddressProvisioner` writes neither `company` nor `vat_number`, so that new row is a byte-copy of the shipping address under a different alias. Net effect: one extra PS address and one extra API call per VAT-invoice Allegro order, carrying none of the data this PR exists to capture.

The reasoning in your docblock for copying the shipping address (protecting a caller's `billingAddress ?? shippingAddress` fallback) is sound for the *invoicing* consumer; it just wasn't checked against the *destination* consumer. Either is a defensible resolution — pass the company/tax id through to the provisioner so the row is worth creating, or narrow what `resolveBillingAddress` returns — but the behaviour change should be deliberate and covered by a test, not a side effect. At minimum, state it in the docblock.

---

**[IMPORTANT]** — `libs/integrations/woocommerce/.../order-source/woocommerce-order.types.ts:38` + `woocommerce-order-source.adapter.ts:272`

`WooCommerceOrderMetaData.value` is typed `string`, but WC REST `meta_data[].value` is arbitrary JSON — plugins routinely store arrays and objects there. `readBillingTaxIdFromMetaData` returns `entry.value` on truthiness and hands it to `readSourceBuyerTaxId`, which calls `.trim()`. A store whose VAT plugin writes an object under one of the three allowlisted keys throws a `TypeError` inside `getOrder`, failing the whole order ingestion rather than reading as unknown. Type it `unknown` and guard with `typeof entry.value === 'string'` — cheap, and it keeps the "absent = unknown, never a hard failure" property the rest of the change is careful about.

Two smaller notes in the same neighbourhood:
- `readBillingTaxIdFromMetaData` and `mapBillingAddress` redeclare `{ key: string; value: string }[]` inline instead of using the `WooCommerceOrderMetaData` type declared four lines away — use the named type so the fix above lands in one place.
- `WOOCOMMERCE_VAT_META_KEY_ALLOWLIST` should be `as const` per `engineering-standards.md` § Union Types, and would benefit from a comment noting an operator has no way to extend it (a store on an unlisted plugin is silently uncovered). A per-connection `config` key is the obvious follow-up if this proves too narrow.

---

**[SUGGESTION]** — `allegro-order-source.adapter.ts:953`

`company.ids?.[0]?.value` takes the first entry regardless of `type`. Your own type comment enumerates `PL_NIP | CZ_ICO | CZ_DIC | OTHER`, and a CZ buyer legitimately carries both `CZ_ICO` (company registration) and `CZ_DIC` (VAT) — array order then decides which identifier reaches a fiscal document. Presence-only consumers (`buyerHasTaxId`) don't care, but the value is carried verbatim into invoice issuance. Prefer a declared preference order, or at least record in the docblock that first-wins is a deliberate choice pending a real multi-id sample.

**[SUGGESTION]** — `allegro-order-source.adapter.ts:932-944`

The new `resolveBillingAddress` docblock has been inserted *between* `resolveShipping`'s existing docblock and the `resolveShipping` method, leaving `resolveShipping` documented by a comment that now describes a different method. Move the new method below `resolveShipping`, or the block above it.

---

### 3. Positive observations

- Every one of the three adapters goes through `readSourceBuyerTaxId` rather than assigning raw — that is the single decision that makes this change safe, and it was made consistently.
- Erli's `isInvoiceAddress` option plus the "ignore a `nip` on `deliveryAddress`" regression test guards a hypothetical future payload rather than today's schema. That is the right instinct for a shared mapper.
- Spec coverage is genuinely edge-case-first (absent meta, blank value, no invoice block, deprecated-field fallback), not happy-path-only, per `testing-guide.md` § Best Practices.
- Correcting the PR's base copy in the same change rather than leaving a now-false warning standing is right; it just needs the PrestaShop half fixed too.
- Layering, naming, imports and the `@openlinker/core/orders` barrel usage are all clean; no `any`, no deep imports, no domain logic leaking into core.

---

### 4. Final Assessment

**Merge readiness: 🔄 Approve with changes.**

No architectural or security blockers — the boundary discipline is right and the three-state semantics are correctly preserved. Approving so this is not held up, but please land the three IMPORTANT items; the copy one is operator-facing and the WooCommerce one can fail an ingestion.

**Priority fixes**
1. Correct "PrestaShop always reports it" in both FE strings and the `describe-sales-document-tax-id-coverage.ts` docblock — a blank `vat_number` reads as unknown, and the core docblock says most consumer orders have one.
2. Guard `meta_data[].value` for non-string types (`unknown` + `typeof` check), and use the `WooCommerceOrderMetaData` type at both call sites.
3. Decide and cover the Allegro→PrestaShop billing-address side effect: either pass company/tax id into the provisioner, or narrow the emitted object — and add a test either way.
4. (Optional) Deterministic `ids[]` selection; move the misplaced docblock; `as const` on the allowlist.

Note `mergeable_state: blocked` — this was opened as a stack element on #2756; confirm the base has landed and the branch is current before merging.

norbert-kulus-blockydevs:
Addressed both reviews (piotrswierzy's formal review + the earlier `/pr-review` pass) in 2d0937d13.

## Fixed

**[IMPORTANT] Allegro `resolveBillingAddress` blank-stub edge case** (flagged independently by both reviews)
`billingAddress` still fell back to a standalone blank-but-truthy stub (`{address1: '', city: '', ...}`) when `resolveShippingAddress()` returned `undefined` (no `delivery.address`, no `delivery.pickupPoint.address`, and no `buyer.address` — a checkout form with a VAT invoice but no address data anywhere) — reproducing the exact failure mode the method exists to eliminate, and defeating a caller's `billingAddress ?? shippingAddress` fallback. `resolveBillingAddress` now returns `undefined` for `billingAddress` too in that case. Added a regression test (`checkout-invoice-no-address`).

**[IMPORTANT] FE copy: "PrestaShop always reports it"**
Corrected in `sales-document-rules-list.tsx`, `sales-document-template-screen.tsx`, and the `describe-sales-document-tax-id-coverage.ts` docblock. PrestaShop's `ps_address.vat_number` is optional and blank on essentially every consumer order, so it's conditional exactly like the other three sources (buyer filled in a VAT number on the address, vs. "always"). Matches the core docblock this copy is derived from.

**[IMPORTANT] WooCommerce `meta_data[].value` typed `string`**
WC REST `meta_data[].value` is arbitrary JSON — a VAT plugin writing an array/object under an allowlisted key would throw a `TypeError` and fail the whole order ingestion. Widened `WooCommerceOrderMetaData.value` to `unknown`, guarded with `typeof === 'string'`, switched both call sites (`readBillingTaxIdFromMetaData`, `mapBillingAddress`) to the named type instead of redeclaring `{key: string; value: string}[]` inline, and made `WOOCOMMERCE_VAT_META_KEY_ALLOWLIST` `as const`. Added a regression test for a non-string meta value.

**[SUGGESTION] Allegro `company.ids?.[0]?.value` first-wins**
Documented as a deliberate choice (a CZ buyer may carry both `CZ_ICO` and `CZ_DIC`) pending a real multi-id sample, per the review's "at least record it" suggestion.

**[SUGGESTION] Misplaced `resolveShipping` docblock**
Moved back above its own method; `resolveBillingAddress`'s docblock now sits directly above `resolveBillingAddress`.

## Documented rather than changed

**[IMPORTANT] Allegro→PrestaShop destination-side billing-address side effect**
`resolveBillingAddress` now populating a real address for every VAT-invoice order means `PrestashopOrderProcessorManagerAdapter` (gated on `order.billingAddress && order.customerId`) provisions a second destination address carrying the same street data as the shipping one. Per the review, this is one of two defensible resolutions (the other being to narrow what the address carries or thread company/tax-id into the provisioner) — I documented the behavior explicitly in the `resolveBillingAddress` docblock rather than changing PrestaShop-adapter behavior, which is outside this issue's scope (wiring buyer tax IDs onto order sources) and would need its own test coverage on the destination side.

## Verification
- `pnpm --filter @openlinker/integrations-allegro test` — 67/67 (was 66; +1 regression test)
- `pnpm --filter @openlinker/integrations-woocommerce test` — 36/36 (was 35; +1 regression test)
- `pnpm --filter @openlinker/web test` — 5131 passed / 1 pre-existing skip
- `pnpm --filter {allegro,woocommerce} lint && type-check` — clean (0 errors)
- `pnpm --filter @openlinker/web lint && type-check` — clean (0 errors)
- `pnpm check:invariants` — all green

piotrswierzy (approved):
## Re-review — #2824 @ `2d0937d` (after `2d0937d13` "address PR #2824 review findings")

Read the prior APPROVE-with-changes review, then the new commit against the current head (11 files, +503/−32).

**(a) "PrestaShop always reports it" — RESOLVED.** All three sites are corrected and now agree with `readSourceBuyerTaxId`'s own docblock:
- `sales-document-rules-list.tsx:82` → *"PrestaShop reports it only when the buyer filled in a VAT number on the address"*.
- `sales-document-template-screen.tsx:101` → same correction.
- `describe-sales-document-tax-id-coverage.ts` → now opens with *"None of the four sources report it unconditionally"* and spells out that `ps_address.vat_number` "is optional and blank on essentially every consumer order, so a B2C order still reads as unknown" — which is exactly the core statement this copy is derived from.
- The per-rule `provenance-tag` tooltip was also fixed (`conditionally for every source`), which I had not called out separately but was the same claim one layer down. Grepping the head for `always report` finds only unrelated pre-existing strings. Good.

**(b) WooCommerce `meta_data[].value` — RESOLVED, and correctly.** `WooCommerceOrderMetaData.value` is now `unknown` with a docblock telling consumers to narrow; `readBillingTaxIdFromMetaData` guards `typeof entry.value === 'string' && entry.value` so a plugin storing an array/object under an allowlisted key reads as unknown instead of throwing a `TypeError` out of `getOrder`. Both call sites now take the named `WooCommerceOrderMetaData[]` type rather than redeclaring the shape inline, `WOOCOMMERCE_VAT_META_KEY_ALLOWLIST` is `as const`, the "operator cannot extend this list; a per-connection `config` key is the obvious follow-up" note is in the docblock, and there is a regression test (`meta_data: [{ key: 'VAT Number', value: { formatted: '…' } }]` → `taxId` undefined, no throw). Every sub-item from that finding is closed.

**(c) Allegro → PrestaShop billing-address side effect — ADDRESSED at the level I asked for, and one step further.** I offered "at minimum, state it in the docblock"; the docblock now states it explicitly, naming the `order.billingAddress && order.customerId` gate on the destination adapter and that no `company`/`taxId` field reaches the destination address itself. I re-confirmed against the head: `prestashop-order-processor-manager.adapter.ts:337` still gates a second `resolveOrCreateAddress(..., 'billing', ...)`, and `prestashop-address-provisioner.ts` writes neither `company` nor `vat_number`, so a VAT-invoice Allegro order does still mint one extra PS address row per order carrying none of this PR's data. That behaviour is unchanged — it is now *deliberate and documented* rather than accidental, which was the ask. See the remaining note below.

The commit additionally fixed a real defect I had not spotted: `resolveBillingAddress` previously spread `shippingAddress ?? { address1: '', city: '', postalCode: '', country: '' }`, so a checkout form with no delivery address, no pickup-point address and no buyer profile address but a VAT invoice produced exactly the blank-but-truthy stub the method's own docblock says it exists to prevent. It now returns `undefined` in that case, with a regression test asserting both `shippingAddress` and `billingAddress` are undefined. That is a genuine self-catch, not a paper fix. The misplaced `resolveShipping` docblock is moved back above its own method, and the first-of-`ids[]`-wins choice is documented with the CZ_ICO/CZ_DIC rationale, closing both SUGGESTIONs.

### 2. New code since the prior review

Nothing new raises a blocker. Layering is unchanged: three adapters, one shared core seam (`readSourceBuyerTaxId` from `@openlinker/core/orders`), no deep imports, no `any` (the `unknown` + `typeof` narrowing is exactly what `engineering-standards.md § Type Safety` asks for), no domain logic pushed into core. The three-state contract (absent = `undefined`, asserted-none = `null`, value verbatim) is preserved on every path, including the new early return. Tests are edge-case-first per `testing-guide.md`.

Two residual notes, neither blocking:

**[SUGGESTION]** The `!shippingAddress` early return silently drops a company name and tax id that Allegro *did* assert — on a VAT-invoice order with no resolvable address anywhere, `buyerHasTaxId` reverts to unknown and the routing rule this PR exists to feed will not match. The docblock says "a caller needing them without a real address has no representable answer to give", which is true of `IncomingOrderAddress` as it stands, and the case is narrow (buyer profile address is the third fallback). Worth an issue if it ever shows up in the wild rather than a change here.

**[SUGGESTION]** The documented PS side effect is still one wasted address row and one wasted API call per VAT-invoice Allegro order. Threading `company`/`taxId` into `PrestashopAddressProvisioner` (so the row carries `company` + `vat_number` and is actually worth creating) is a small, self-contained follow-up and would turn a documented cost into a feature. Destination-adapter scope, correctly out of this PR.

### 3. Final assessment

**Merge readiness: ✅ Approve.** All three IMPORTANT findings are resolved, both SUGGESTIONs are closed, and the round found and fixed a further defect of the same class. Confirm the branch is current against `main` before merging (the PR was opened as a stack element on #2756; head already carries a merge of `origin/main`).

norbert-kulus-blockydevs:
Re: the re-review's two residual SUGGESTIONs.

Checked both against the code and against the review's own wording — neither warrants a change in this PR, and the review says so explicitly in each case:

**1. `!shippingAddress` early return dropping company/taxId**
`resolveBillingAddress` returns `undefined` whenever `resolveShippingAddress()` couldn't produce a real address (no `delivery.address`, no `delivery.pickupPoint.address`, no `buyer.address`). If it instead tried to preserve `company`/`taxId` in that case, the only way to satisfy `IncomingOrderAddress`'s required `address1`/`city`/`postalCode`/`country` (all plain `string`) would be to fill them with `''` — which is exactly the blank-but-truthy stub this method was fixed to eliminate one round ago. There's no representable middle ground given the current type: a caller needing a tax id with no address has nothing valid to receive it, and that's stated verbatim in the docblock already. The review itself frames this as "worth an issue if it ever shows up in the wild rather than a change here," so I've left it as documented behavior rather than force a change that would reopen the stub problem.

**2. PrestaShop destination-side wasted address row**
Confirmed still true: an Allegro VAT-invoice order provisions a second PrestaShop address (same street data as shipping, no `company`/`vat_number`) via `order.billingAddress && order.customerId` on `PrestashopOrderProcessorManagerAdapter`. The review explicitly scopes the real fix — threading `company`/`taxId` into `PrestashopAddressProvisioner` — to a self-contained follow-up on the destination adapter, "correctly out of this PR" (this issue is about wiring buyer tax IDs onto *order sources*, not destination-side address provisioning). Leaving it as documented in `resolveBillingAddress`'s docblock rather than expanding scope here.

No code changes from this pass — both items were reviewed and are deliberately deferred per the reviewer's own recommendation, not overlooked.

piotrswierzy (approved):
## Re-review — #2824 @ `d08a8ad`

Head moved from `2d0937d` to `d08a8ad` via three merge commits from `main` (`75a5268`, `e9555fe`, `d08a8ad`) and **no new code**. I verified that rather than assuming it: hashing `git diff $(merge-base origin/main HEAD) HEAD` at both the previously-approved head and the current one yields the identical blob (`7b29d68…`). The effective change against `main` is byte-for-byte what I approved, still 11 files / +503 −32.

The PR is also no longer a stack element — `base` is now `main` at `58d9b7e` (current `origin/main` tip), and `mergeable_state` is `clean`. The "confirm the base has landed" caveat from both prior reviews is discharged.

Re-confirmed each against the current tree, not against the previous review's text:

- **(a) "PrestaShop always reports it" — RESOLVED, still.** `sales-document-rules-list.tsx`, `sales-document-template-screen.tsx` and `describe-sales-document-tax-id-coverage.ts` all state the per-source conditionality, and the docblock still opens with "None of the four sources report it unconditionally". Grepping the current head for `always report` finds only unrelated pre-existing strings. Notably `main` has since landed `cf0d8870c` ("say delivered to, not billed to, in market copy"), which touches the same sales-documents copy area — the merges absorbed it without reintroducing the false claim.
- **(b) WooCommerce `meta_data[].value` — RESOLVED, still.** `unknown` on the wire type, `typeof entry.value === 'string' && entry.value` at the read, named `WooCommerceOrderMetaData[]` at both call sites, `as const` allowlist with the "operator cannot extend this list" note, plus the non-string regression test.
- **(c) Allegro→PrestaShop billing-address side effect — ADDRESSED, still.** Documented in the docblock, naming the `order.billingAddress && order.customerId` gate and that no `company`/`taxId` reaches the destination address. The self-caught blank-address-stub fix (`resolveBillingAddress` returning `undefined` when no address resolves anywhere) and its regression test are intact.

Nothing regressed across the three merges.

### 2. New material since the last review

None in the PR's own diff. The merges bring in `main`'s advances (orders-list invoice/receipt status #2761, the ADR-071 / Wave 3b docs, the #2828 CI flake fix); none of them touch the three adapters, the core `readSourceBuyerTaxId` seam, or the two FE components this PR edits, and the merge produced no conflict resolutions of its own.

### 3. Migration ordering

Not applicable — this PR adds no migration.

### 4. CI

All 8 checks green on `d08a8ad` (run `33801213943`): Build, Lint, Type Check, **Test**, **Integration Tests**, Docker Build Smoke Test, PHP Unit Tests, Detect changed paths. Integration Tests passing here is meaningful — this head carries `#2828`'s Testcontainers/`SyncJobRunner` flake fix, so that green is a real signal rather than a lucky run.

### 5. Remaining open item

**[SUGGESTION]** — unchanged and still non-blocking: a VAT-invoice Allegro order with no resolvable address anywhere (no delivery address, no pickup point, no buyer-profile address) takes the `!shippingAddress` early return, so a company name and tax id Allegro *did* assert are dropped and `buyerHasTaxId` reverts to unknown. `IncomingOrderAddress` has no representable answer for "a tax id with no address", so the early return is the right call as the type stands; the honest resolution is a follow-up issue if it is ever observed in the wild, not a change here. Second, pre-existing note: threading `company`/`taxId` into `PrestashopAddressProvisioner` would turn the documented one-extra-address-row-per-VAT-order cost into a feature — destination-adapter scope, correctly out of this PR.

### 6. Final assessment

**Merge readiness: ✅ Approve.** All three IMPORTANT findings from the first review and both SUGGESTIONs from the second remain resolved, the diff is provably unchanged, the base is current `main`, and CI is fully green. Ready to merge as-is.

Issue #2826, opened 2026-09-02
[TECH-DEBT] Application — Page/batch tax coverage detection instead of unbounded fetch + N+1


jakubretajczykBD:
## Problem / Context

Found during review of PR #2668.

`OrderRecordRepository.findNetExcludedOrderCandidates` (`libs/core/src/orders/infrastructure/persistence/repositories/order-record.repository.ts:775-798`) runs an **unpaged `qb.getMany()`** over the full `SalesAnalyticsFilters` window (currently up to 400 days, per the analytics coverage controller), hydrating full `OrderRecordOrmEntity` rows — including the `orderSnapshot` JSONB column — for every non-cancelled, currently-stamped, net-excluded order.

`TaxCoverageDetectionService.classify` (`libs/core/src/orders/application/services/tax-coverage-detection.service.ts:104-146`) then iterates that unbounded candidate list **sequentially**:
- `classifyOne` (`:183-232`) calls `orderLineItemRepository.findByOrderId(candidate.internalOrderId)` once per candidate order, and
- for each unresolved line, calls `productsService.getEffectiveTaxRate(productId, variantId)` once per line — also sequential (`for` loop with `await` inside, `:203-230`).

This whole chain is reachable unconditionally on `/analytics` page mount via `GET /analytics/coverage`. On a fresh deployment — zero catalogue tax coverage is the explicitly documented day-one state per ADR-063 (per-line tax-rate resolution) — this means tens of thousands of hydrated `OrderRecord` rows plus potentially hundreds of thousands of sequential DB/catalogue-read round trips per page load, on the analytics landing route rather than a rarely-hit admin action.

The unpaged read is deliberate today (see the service's own doc comment: classification depends on a live catalogue read with no SQL equivalent, so slicing happens in-memory after full classification), but the current shape scales with total order-history size rather than with anything bounded, and the classification loop compounds it with sequential I/O.

## Proposed Solution

- Push what CAN be expressed in SQL into SQL: the A/B/C split currently re-derives, per order, information already resolvable via a join between `order_records` and `order_line_items` (e.g. "does this order have any line with `taxRate` unresolved") — reduce the population that needs an application-layer live catalogue check to just the lines that are genuinely ambiguous.
- Batch the remaining per-order/per-line work instead of one-at-a-time sequential awaits:
  - Replace the per-candidate `orderLineItemRepository.findByOrderId` call with a single batched read across all candidate order ids (e.g. `findByOrderIds`), keyed back in memory.
[...]

PR #2827, opened 2026-09-02
perf(orders): batch and dedupe the tax coverage detection reads

jakubretajczykBD:
## Summary
- `findNetExcludedOrderCandidates` now selects only the four columns `TaxCoverageDetectionService` needs via `getRawMany`, instead of hydrating the full `OrderRecordOrmEntity` row (including the `orderSnapshot` JSONB column) via `getMany`.
- Added `OrderLineItemRepositoryPort.findByOrderIds` — one batched `IN (...)` query for a set of order ids, replacing the per-candidate `findByOrderId` N+1 in `TaxCoverageDetectionService.classify`.
- `classify()` now filters to pre-rollout candidates before doing any line-item I/O, batches that line-item read, deduplicates the catalogue rate check by `(productId, variantId)` (many lines/orders share a product), and resolves rates with bounded concurrency (5 workers) instead of one sequential `getEffectiveTaxRate` call per unresolved line.

This closes a live performance issue: `GET /analytics/coverage` is hit unconditionally on `/analytics` page mount, and on a fresh deployment (zero catalogue tax coverage is ADR-063's stated day-one state) the old unbounded read plus double N+1 could mean tens of thousands of hydrated rows and hundreds of thousands of sequential queries per page load.

Base branch is `epic/2452-analytics-currency-coverage` because this code (from PR #2668) hasn't landed on `main` yet.

## Test plan
- [x] `pnpm --filter @openlinker/core exec tsc -p tsconfig.json --noEmit` — clean
- [x] `npx eslint` on all touched files — 0 errors (3 pre-existing, unrelated warnings)
- [x] `npx jest` on the three affected spec files — 140/140 passing, including new tests for `findByOrderIds` and for rate-lookup deduplication
- [x] `tax-coverage-net-excluded.int-spec.ts` run live against real Postgres (Testcontainers) post-review — 2/2 passing, confirming the `getRawMany` raw-SQL rewrite behaves correctly against a real server, not just the mocked query-builder unit specs

## Scope note (post-review)

This PR closes two of #2826's three acceptance criteria (batched line-item read, bounded-concurrency catalogue check). It does **not** close the third: `findNetExcludedOrderCandidates`'s base query is still unbounded — no `LIMIT`/`OFFSET` — so row count returned still scales linearly with total order history for the filter window, even though each row is now cheap. That remaining piece is split out as its own follow-up: #2834.

Progresses #2826 (not "Closes" — see #2834 for the remainder).

### DIFF
--- a/libs/core/src/orders/application/services/tax-coverage-detection.service.spec.ts
+++ b/libs/core/src/orders/application/services/tax-coverage-detection.service.spec.ts
@@ -12,7 +12,7 @@ describe('TaxCoverageDetectionService (#2465)', () => {
   let recordRepository: jest.Mocked<
     Pick<OrderRecordRepositoryPort, 'findNetExcludedOrderCandidates'>
   >;
-  let lineItemRepository: jest.Mocked<Pick<OrderLineItemRepositoryPort, 'findByOrderId'>>;
+  let lineItemRepository: jest.Mocked<Pick<OrderLineItemRepositoryPort, 'findByOrderIds'>>;
   let productsService: jest.Mocked<Pick<IProductsService, 'getEffectiveTaxRate'>>;
 
   const baseFilters = {
@@ -53,9 +53,20 @@ describe('TaxCoverageDetectionService (#2465)', () => {
   const noRate: StoredTaxRate = { code: null, countryIso2: 'PL', readAt: new Date('2026-08-24T00:00:00Z') };
   const notChecked: StoredTaxRate = { code: null, countryIso2: null, readAt: null };
 
+  /** Groups lines by `orderRecordId` into the `Map` shape `findByOrderIds` returns. */
+  const linesMap = (...lines: OrderLineItem[]): Map<string, OrderLineItem[]> => {
+    const map = new Map<string, OrderLineItem[]>();
+    for (const line of lines) {
+      const existing = map.get(line.orderRecordId) ?? [];
+      existing.push(line);
+      map.set(line.orderRecordId, existing);
+    }
+    return map;
+  };
+
   beforeEach(() => {
     recordRepository = { findNetExcludedOrderCandidates: jest.fn() };
-    lineItemRepository = { findByOrderId: jest.fn() };
+    lineItemRepository = { findByOrderIds: jest.fn().mockResolvedValue(new Map()) };
     productsService = { getEffectiveTaxRate: jest.fn() };
 
     service = new TaxCoverageDetectionService(
@@ -77,16 +88,16 @@ describe('TaxCoverageDetectionService (#2465)', () => {
       expect(result['tax-b'][0].internalOrderId).toBe('order-post-rollout');
       expect(result['tax-a']).toHaveLength(0);
       expect(result['tax-c']).toHaveLength(0);
-      expect(lineItemRepository.findByOrderId).not.toHaveBeenCalled();
+      expect(lineItemRepository.findByOrderIds).toHaveBeenCalledWith([]);
     });
 
     it('reports tax-a when a pre-rollout order already has every line resolved (backfill already ran)', async () => {
       recordRepository.findNetExcludedOrderCandidates.mockResolvedValue([
         candidate({ internalOrderId: 'order-a', taxRateEra: 'pre-rollout' }),
       ]);
-      lineItemRepository.findByOrderId.mockResolvedValue([
-        makeLine({ orderRecordId: 'order-a', taxRate: '23' }),
-      ]);
+      lineItemRepository.findByOrderIds.mockResolvedValue(
+        linesMap(makeLine({ orderRecordId: 'order-a', taxRate: '23' }))
+      );
 
       const result = await service.classify(baseFilters, 'EUR');
 
@@ -99,9 +110,9 @@ describe('TaxCoverageDetectionService (#2465)', () => {
       recordRepository.findNetExcludedOrderCandidates.mockResolvedValue([
         candidate({ internalOrderId: 'order-a', taxRateEra: 'pre-rollout' }),
       ]);
-      lineItemRepository.findByOrderId.mockResolvedValue([
-        makeLine({ orderRecordId: 'order-a', productId: 'ol_product_x', taxRate: null }),
-      ]);
+      lineItemRepository.findByOrderIds.mockResolvedValue(
+        linesMap(makeLine({ orderRecordId: 'order-a', productId: 'ol_product_x', taxRate: null }))
+      );
       productsService.getEffectiveTaxRate.mockResolvedValue(known('23'));
 
       const result = await service.classify(baseFilters, 'EUR');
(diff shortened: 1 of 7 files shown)


norbert-kulus-blockydevs:
## PR Review: #2827 - perf(orders): batch and dedupe the tax coverage detection reads

**Verdict: Approved with changes**

### Summary

This PR fixes two of the three problems raised in the earlier `/pr-review` of epic PR #2668 (that review's "Blocking 3") and tracked as issue #2826: the per-candidate `findByOrderId` N+1 is replaced by one batched `findByOrderIds` read, and the per-line `getEffectiveTaxRate` sequential N+1 is replaced by a deduplicated-by-`(productId, variantId)` lookup with bounded concurrency (5 workers, mirroring the `resolveBatchConcurrency` precedent from ADR-047/#2229). Both changes are correct: `getEffectiveTaxRate` is a pure function of `(productId, variantId)` with no order/connection context, so cross-order deduplication cannot change any classification outcome, only how many times the same answer is fetched. The `!rate` guard in `classifyOne` correctly folds both "key not found" and "catalogue read failed" into `'not-checked'`, matching pre-PR error handling. Tests are thorough: dedup is pinned by an explicit "resolves ONCE per distinct pair" test, and the concurrency bound is pinned by an in-flight counter test that proves both the ceiling (`<=5`) and that it's actually concurrent (`>1`) - this second test was added in a follow-up commit responding to an earlier internal tech-review pass on this same PR, which is a good sign of iteration.

The repository-level change (`getMany()` -> `getRawMany()` selecting only 4 scalar columns instead of the full `OrderRecordOrmEntity` including the `orderSnapshot` JSONB) follows an already-established pattern in the same file (`findEarliestOrderDateByConnection`, `countOrdersByRoutingCountrySince` already use quoted `rec."columnName"` raw-select + `getRawMany`), so it's not introducing a new idiom.

### IMPORTANT

1. **This PR does not close the core scaling problem #2826 was filed for, and the "Closes #2826" claim overstates what's fixed.**
   `findNetExcludedOrderCandidates` (`order-record.repository.ts`) is still an **unpaged, un-`LIMIT`ed** query across the full `SalesAnalyticsFilters` window (up to 400 days per the analytics coverage controller). The PR reduces the *per-row* cost dramatically (no full entity hydration, no JSONB) and removes the double sequential N+1 - both real, valuable fixes - but the row *count* read from the DB on a fresh deployment (ADR-063's stated day-one state: zero catalogue tax coverage) is unchanged. This was the specific concern the earlier epic review (#2668) called out as "Blocking" and is issue #2826's first, unchecked acceptance-criteria bullet: *"`findNetExcludedOrderCandidates` no longer performs an unbounded `getMany()` whose cost scales linearly with total order history... it is either paged/batched or the classification work per page is bounded."* Narrowing the columns does not change the O(N) row scan/return itself. Recommend one of: (a) don't let this PR auto-close #2826 - file a narrower follow-up issue for the pagination/SQL-push-down half and reference it instead, or (b) get explicit sign-off from whoever owns the epic that the column-narrowing + N+1 removal is considered sufficient closure for now, and downgrade #2826 accordingly. Either way, the PR description should not claim full closure of an issue whose stated acceptance criterion is still open.

2. **The raw-SQL rewrite is unverified against a real database, and CI hasn't independently confirmed it either.** The PR body's own test plan is `tsc`/`eslint`/`jest` only (all mocked at the query-builder level - the unit specs assert `select`/`addSelect` were *called* with the right strings, not that the query actually executes and maps correctly against Postgres), and the author has stated this hasn't been tested live yet. Checking the PR's CI status, only a "Scaffolded Adapter Builds" smoke check has run - no lint/type-check/unit-test/integration workflow appears to have executed against this PR (likely because it targets `epic/2452-analytics-currency-coverage` rather than `main`). Given this touches raw column aliasing (`rec."internalOrderId"` etc.) with no integration test covering it, recommend a quick sanity check against a real/test Postgres instance (or confirming the full CI gate does run for epic-branch PRs) before merging.

### SUGGESTIONS

- The dedup/bounded-concurrency logic (`resolveRates`, `rateKey`) is a second, independently-implemented "wave-free" bounded worker pool. There's a slightly different existing local implementation (`AllegroOfferManagerAdapter.runWithConcurrency`, wave-based) for the same general problem elsewhere in the codebase, kept local rather than shared per that file's own doc comment. Not a defect - just worth being aware there's now a second bespoke implementation of "bounded concurrent fan-out over a list" in the codebase; if a third one shows up, it's probably time for a shared `libs/shared` helper.
- Consider whether `sourceConnectionId` deserves a comment at the raw-select site noting it's carried through only for `toRow`'s output shape (not used in classification) - a future reader diffing the 4 SELECT columns against `classify()`'s logic might wonder why it's there.

### Epic-completeness note (outside strict PR-diff scope, since you asked)

Per the earlier /pr-review thread on epic PR #2668, the reviewer's final verdict there was **"Approve with changes - conditional on the Blocking item (unpaged tax-coverage detector) being addressed before merging to `main`."** That Blocking item is #2826, and per the analysis above, it is only **partially** addressed by this PR - the sequential-N+1 half is fixed, the unpaged-full-window-scan half is not. If the epic's path to `main` is gated on that Blocking item being fully resolved (as the review stated), merging #2827 as-is and closing #2826 would incorrectly signal that gate is clear. Worth flagging to whoever is tracking epic #2452's merge readiness so the "unpaged read" half doesn't get lost once #2826 closes. The other items from that same review round (Important 8, 9, 11, 18) were reported as still open as of the last status-check comment on #2668 and are unrelated to this PR's scope.

norbert-kulus-blockydevs:
Addressed the review findings:

**IMPORTANT #1 (unpaged base fetch, #2826 not fully closed)** - confirmed correct. `findNetExcludedOrderCandidates` still runs one unbounded query over the full filter window; only per-row hydration cost was fixed, not row count. Rather than redesigning `TaxCoverageDetectionService.classify` into an incremental/paginated shape under review pressure (a real architectural change, not a quick fix), I:
- Corrected the PR body: it no longer claims "Closes #2826" (now "Progresses #2826").
- Filed a narrower follow-up issue for the remaining piece: #2834.

**IMPORTANT #2 (unverified raw-SQL rewrite)** - resolved by verification, no code change needed. Built `@openlinker/shared`/`@openlinker/core`/`@openlinker/test-kit` and ran the existing live-Postgres int-spec (`tax-coverage-net-excluded.int-spec.ts`) against real Testcontainers Postgres:
```
PASS test/integration/orders/tax-coverage-net-excluded.int-spec.ts (173.359 s)
  Tax coverage — net-excluded candidates against real Postgres (#2465)
    ✓ reports the SAME count as getDailyOrderAggregates.netExcludedCount for a mix of eligible, pre-rollout, and unresolved-rate orders (regression guard)
    ✓ excludes cancelled orders and orders stamped in a prior reporting-currency era
```
Both pass, confirming the `getRawMany` raw-SQL rewrite behaves correctly against a real server, not just the mocked query-builder unit specs. Added this run to the PR's test plan.

**Suggestions** left as-is (both genuinely minor, no code change): the bounded-worker-pool duplication (only a second occurrence exists so far) and the `sourceConnectionId`-selected-but-unused comment (it already feeds `toRow`'s output, which is a one-line fact not worth a diff).

No functional code was changed in this PR beyond what was already there - the fixes here are process corrections (issue scoping) and verification (live-DB test run), since the actual remaining gap is real follow-up work, not a bug in what shipped.

PR #2749, opened 2026-08-31
fix(sales-documents): say delivered to, not billed to, in market copy

norbert-kulus-blockydevs:
## Summary
- Sales-document routing and market discovery already evaluate on the **delivery** address's country (ADR-041 decision 5 — `order-to-sales-document-order-facts.mapper.ts` reads `order.shippingAddress.country`; `countOrdersByRoutingCountrySince`'s own in-code comment says the same).
- Two already-merged copy strings claimed **billing** country instead, contradicting the shipped logic. One is operator-facing dialog text, which actively misleads whenever an order's billing and delivery countries differ (gift order, drop-ship, corporate billing to a group HQ).
- Fixes wording only in the two spots not covered by the still-open docs PRs (#2570, #2720), which had the identical wording defect and are being fixed there.

## Changes
- `apps/api/src/sales-documents/http/dto/sales-document-market-response.dto.ts` — Swagger `orderCount` description: "billed to" → "delivered to"
- `apps/web/src/features/sales-documents/components/sales-document-country-routing-dialog.tsx` — operator-facing dialog copy: "billed to" → "delivered to"

## Test plan
- [x] `eslint` on both changed files — clean
- [x] `pnpm --filter @openlinker/api type-check` — clean
- [x] `pnpm --filter web type-check` — clean
- [x] Grepped `apps/api/src/sales-documents` and `apps/web/src/features/sales-documents` for any other "billed"/"billing" occurrence and any test asserting the old string — none found

Closes #2748

PR #2828, opened 2026-09-02
fix(ci): close both integration-test flakes - Testcontainers network prune and the live SyncJobRunner

### LINKED ISSUE #2825: [BUG] Infrastructure - CI integration-test flakes: Testcontainers network-prune race (API) + live SyncJobRunner racing every manually-driven worker int-spec
## Problem / Context

`test/integration/prestashop/prestashop-webhook-provisioning.int-spec.ts` failed on PR #2756's CI run with:

```
(HTTP code 404) no such container - failed to set up container networking: network 03e01b9cd5ee not found
```

(CI run: https://github.com/openlinker-project/openlinker/actions/runs/33632508953/job/100255262960)

PR #2756 only touched `apps/web/src/features/sales-documents/**` (frontend copy), so the failure is unrelated to that diff. Running the exact same suite locally in isolation passes cleanly (2/2), confirming this is an infrastructure flake, not a test-logic bug.

**Root cause**, traced through `.github/workflows/ci.yml` and `apps/api/test/integration/helpers/prestashop-container.helper.ts`:

1. `test-integration` runs on a persistent **self-hosted** runner (`.github/workflows/ci.yml:169`) where, per the job's own "Sweep orphaned Testcontainers" step comment (`.github/workflows/ci.yml:241-245`), **more than one `test-integration` job legitimately runs concurrently** on the same box.
2. Every `test-integration` job's "Sweep orphaned Testcontainers from finished CI runs (#1285)" step unconditionally runs:
   ```
   docker network prune -f --filter "label=org.testcontainers=true"
   ```
   (`.github/workflows/ci.yml:302`) with the comment "Network prune is safe unconditionally: Docker refuses to remove a network that still has containers attached, regardless of age or which run created it."
3. That claim is true only for networks that **already have a container attached**. `startPrestashopContainer` (`apps/api/test/integration/helpers/prestashop-container.helper.ts:172-336`) does:
   ```ts
   const network = await new Network().start();   // network exists, 0 containers attached
   ...
   mysql = await startMysql(network, mysqlLogBuffer);       // first container attaches here
   prestashop = await startPrestashop(network, psLogBuffer, psDataDir);
   ```
   Testcontainers' `Network.start()` (`testcontainers@10.28.0`, `build/network/network.js`) stamps the network with `org.testcontainers=true` + `org.testcontainers.session-id` — **no run-id label**, since the `Network` builder in this testcontainers version has no `.withLabels()` API (unlike `GenericContainer`/`MySqlContainer`/`PostgreSqlContainer`/`RedisContainer`, which all get `ciRunIdLabels()` via `libs/test-kit/src/containers.ts`'s `ciRunIdLabels()` / the PS/MySQL builders in the same helper file).
4. Between step 3's `network.start()` and the MySQL container's `docker create` actually attaching to it, the network is a completely unattached, unlabeled (by run id) `org.testcontainers=true` network — indistinguishable from a genuinely orphaned one. If the MySQL image needs a cold pull (documented "5-10 min cold-cache CI" in the same helper file's docblock, line 13), this window can be **minutes**, not milliseconds.
5. A **different, concurrently running** `test-integration` job's sweep step can run its `docker network prune` during that exact window and legally remove the not-yet-attached network (Docker only refuses to remove a network that currently has a container attached — it has no concept of "about to be attached").
6. When the original job's suite then tries to attach its first container to that now-deleted network id, Docker returns the observed 404.

Verified locally that `docker network prune`'s `until=<duration>` filter (already used elsewhere in Docker tooling, confirmed via `docker network prune --help` and a live test against this machine's Docker daemon) correctly preserves a network younger than the given age, which is the missing guard.

## Proposed Solution

Add an age-based grace window to the sweep step's network prune, mirroring the age-cutoff philosophy already used for the container half of the same step (`CUTOFF_EPOCH`, `.github/workflows/ci.yml:261`/`284-287`), so a network created by a still-in-flight (possibly cold-image-pulling) sibling job can never be pruned before it plausibly could have a container attached:

```
docker network prune -f --filter "label=org.testcontainers=true" --filter "until=15m"
```

15 minutes comfortably exceeds the documented worst-case PS/MySQL cold-boot window (5-10 min) while still reclaiming genuinely orphaned empty networks (left behind by a SIGKILLed/cancelled run, since Ryuk is disabled on this runner per the job's own `TESTCONTAINERS_RYUK_DISABLED` env comment) promptly enough not to pressure the runner.

Update the explanatory comment above the `docker network prune` line (`.github/workflows/ci.yml:299-301`) to state the actual invariant (Docker only protects *attached* networks; the `until` filter is what protects an unattached-but-still-being-set-up network) instead of the current "safe unconditionally" claim, which this bug disproves.

## Classification

**Type**: Infrastructure / DX
**Layer**: CI workflow
**File(s)**:
- `.github/workflows/ci.yml` (orphan-sweep step, ~line 230-302)
- `apps/api/test/integration/helpers/prestashop-container.helper.ts` (for context — no change expected here)
- `libs/test-kit/src/containers.ts` (for context — no change expected here)

## Dependencies

- None. Builds on the existing orphan-sweep step from #1285.

## Assumptions

- The self-hosted runner's concurrent-job-slot behavior (documented in the existing sweep-step comment) is intentional and staying as-is; the fix works within that constraint rather than trying to serialize `test-integration` jobs.
- A 15-minute grace window is a reasonable, conservative default; it can be tuned if it turns out to be too short/long in practice without changing the underlying mechanism.

## Acceptance Criteria

- [ ] `docker network prune` in the orphan-sweep step is scoped with an `until=` age filter (not blanket-unconditional) so a network belonging to a network-created-but-not-yet-attached in-flight job cannot be removed by a sibling job's sweep.
- [ ] The step's comment accurately describes the invariant being relied on (attached-network protection from Docker + the age filter for the unattached window), not the disproven "safe unconditionally" claim.
- [ ] `prestashop-webhook-provisioning.int-spec.ts` (and any other suite that boots its own dedicated `Network()`) is verified to still pass locally and remains reproducible-flake-free across a manual re-run of CI.
- [ ] No architecture boundary violations (CORE ↔ Integration) — this change is CI-workflow-only.

---

## Second flake (added after further investigation) — worker's `SyncJobRunner` double-processes a manually-executed test job

A **second, unrelated** integration-test flake was found while investigating this issue, on a different PR's CI run: https://github.com/openlinker-project/openlinker/actions/runs/33626466834/job/100235322415?pr=2782 (`test-integration` job, "Run worker integration tests" step). Different subsystem, different root cause, different fix — documented here per request rather than as a separate issue.

**Observed failure**:

```
● Allegro Order Sync End-to-End Integration › Complete Order Sync Flow › should sync orders from poll to order processor

SyncJobExecutionError: Marketplace order sync failed: Order create is contended for order ol_order_d5506858f32c4cebb17cb59d5952d3dc (destinationConnectionId=dest-prestashop-1); a concurrent create holds the lock — retry

Cause:
OrderCreateContendedException: Order create is contended for order ol_order_d5506858f32c4cebb17cb59d5952d3dc (destinationConnectionId=dest-prestashop-1); a concurrent create holds the lock — retry
  at OrderSyncService.createOrderIdempotently (libs/core/src/orders/application/services/order-sync.service.ts:235)
```

Test: `apps/worker/test/integration/allegro-order-sync-e2e.int-spec.ts` › `should sync orders from poll to order processor`.

**This is a known, partially-mitigated race — the mitigation is incomplete.** The test file already carries a comment (`allegro-order-sync-e2e.int-spec.ts:173-182`) explicitly describing this exact failure mode and claiming a fix:

> "the harness boots the full role set, so a real publish here reaches the live `JobIntakeConsumer` + `SyncJobRunner` ... Both resolve the same internal order id + destination connection, so that race intermittently trips the per-(order, destination) create lock with a retryable `OrderCreateContendedException`. Stubbing keeps the call-shape assertion while removing the double-processing."

The mitigation (`jest.spyOn(jobEnqueue, 'enqueueJob').mockResolvedValue(...)` at line 183) stubs `JobEnqueuePort.enqueueJob` (`JOB_ENQUEUE_TOKEN` → `RedisStreamsJobEnqueueService`), which correctly prevents the poll-triggered `marketplace.order.sync` job from reaching Redis (traced: `OrderIngestionService.ingestOrders` → `SyncJobQueueService.enqueueBulk` → `this.jobEnqueue.enqueueJob`, same token, so the stub does cover that path). But it did **not** actually stop the double-processing, because there's a **second, un-stubbed path to the same race**:

1. `WorkerIntegrationTestHarness.setup()` (`apps/worker/test/integration/setup.ts:38-40`) boots the worker's full role set via `AppModule.forRoles()`, with a comment claiming "the per-service enable gates (`WORKER_RUNNER_ENABLED`, `OL_SCHEDULER_ENABLED`, …) keep background loops quiet under tests."
2. That claim is only half true. `apps/worker/test/integration/harness.ts:58` sets `process.env.OL_SCHEDULER_ENABLED = 'false'`, but **nothing** in the integration-test setup (`harness.ts`, `setup-global.ts`, `setup.ts`, `jest-integration.cjs`, or the CI workflow's `test-integration` job env) sets `WORKER_RUNNER_ENABLED`. `SyncJobRunner` (`apps/worker/src/sync/sync-job.runner.ts:133`) defaults it to `'true'` when unset — so the real background `SyncJobRunner`, which polls `sync_jobs` directly in Postgres roughly every second (independent of Redis/the enqueue stub), **is actually running for the whole worker integration suite**. (The env var the comment names, `WORKER_RUNNER_ENABLED=false`, IS set — but only in `apps/worker/test/jest-global-setup.cjs:11`, which is wired to the worker's **unit**-test Jest config, not `jest-integration.cjs`'s `globalSetup`.)
3. In the test's step 6, `orderSyncPersisted` is inserted directly into Postgres via `jobRepository.createIfNotExistsByIdempotencyKey({ jobType: 'marketplace.order.sync', ... })` (status `queued` by default), then the handler is invoked **manually**: `await orderSyncHandler.execute(orderSyncPersisted)` — bypassing the runner's own claim mechanism (`findAndLockDueJobs`, `FOR UPDATE SKIP LOCKED`) entirely.
4. Because the row is genuinely `queued` in Postgres and the real runner is alive, the runner's own poll can independently claim and execute that **same row** at any point after it's inserted — concurrently with the test's manual call. Both executions reach `OrderSyncService.syncOrder` → `createOrderIdempotently` for the same `(internalOrderId, destinationConnectionId='dest-prestashop-1')`, tripping the legitimate per-order Redis lock; whichever execution loses the race throws `OrderCreateContendedException`.
5. It's intermittent because it's a genuine timing race against the runner's ~1s poll tick vs. how long the manual call takes to create the destination order and release the lock — most runs the manual call finishes (and `markSucceeded` flips the row's status away from `queued`) before the runner's poll would have picked it up.

This is **not** disproven by "Redis is flushed between tests" (`WorkerIntegrationTestHarness.reset()` does call `redisClient.flushDb()`) — the race is same-test, not cross-test: the row is created and raced within this one `it()` block.

### Proposed solution (second flake)

Either:
- Actually honor the `setup.ts` comment's stated intent: set `process.env.WORKER_RUNNER_ENABLED = 'false'` in `apps/worker/test/integration/harness.ts` (alongside the existing `OL_SCHEDULER_ENABLED = 'false'`) for every worker integration suite that manually drives handlers — the least invasive fix, and it matches what the code already claims to do; or
- If some other worker integration suite actually relies on the live runner picking up a real `sync_jobs` row (worth checking before disabling globally), scope the fix to this test file instead — e.g. claim/lock the row (or mark it non-`queued`) before the manual `execute()` call, so the runner's poll can never claim it too.

### Acceptance criteria (second flake)

- [ ] The live-runner-vs-manual-execute race in `allegro-order-sync-e2e.int-spec.ts` (or the worker integration harness generally) can no longer produce `OrderCreateContendedException` from a single test's own manual handler invocation.
- [ ] Confirm no other worker `*.int-spec.ts` file relies on the background `SyncJobRunner` actually being live (i.e. asserts on a job the test itself never manually executes) before disabling it harness-wide.
- [ ] `apps/worker/test/integration/setup.ts`'s comment is corrected if `WORKER_RUNNER_ENABLED` is not actually being set, or the code is fixed to match the comment's stated intent — whichever direction is chosen, the two must agree.

**Not in scope for the PR fixing the first flake above** — flagged here for tracking; happy to pick up as a follow-up in the same or a separate PR depending on preference.

### Scope of the second flake: harness-wide, not one suite

The live runner is enabled for **every** worker integration suite, so `allegro-order-sync-e2e.int-spec.ts` is one symptom rather than the extent of it. Two more suites are affected, both of which insert a real `queued` `sync_jobs` row and then drive it by hand:

1. **`inventory-provenance-backfill-e2e.int-spec.ts`** - inserts its job via `jobRepository.createIfNotExistsByIdempotencyKey(...)` with **no `runAfter`**, so the row is immediately due, then calls `handler.execute(job)` directly. `inventory.provenance.backfill` is registered in the `bulk` lane (`handler-registration.service.spec.ts:155`), so the live runner can claim that exact row. The suite asserts exact drain counts after each manual tick at `pageLimit: 5` (`remainingNull` `'7'` -> `'2'` -> `'0'`); one concurrent claim stamps an extra page and those numbers no longer hold. Observed failing on PR #2749's CI (a frontend-copy-only PR), run 33728532711, with the `Error in runner loop` signature in the log; the same suite passes 7/7 in ~15s locally in isolation, and the preceding commit on that same branch passed the job.

2. **`job-intake-execution.int-spec.ts`** - asserts `retrievedJob?.status).toBe('queued')` on a row created with `nextRunAt: new Date()` (due now), and separately asserts that its own `findAndLockDueJobs(10, 'test-worker-123')` call returns >= 2 rows with `lockedBy === 'test-worker-123'`. A live runner claiming either row first flips the status to `running` under a different worker id, so both assertions are order-dependent on a background loop the test does not control.

`allegro-cursor-persistence.int-spec.ts` corroborates the mechanism from the other direction: it already carries a per-suite workaround for exactly this race, scheduling every job with `runAfter = now + 24h` and stating the reason in a module-level comment ("The harness also runs a live `SyncJobRunner` polling loop (1s interval) that would otherwise claim the same 'queued' row and reprocess it concurrently").

### Acceptance criterion 2 verified

All 26 worker `*.int-spec.ts` files were audited for reliance on the background loop. **None relies on it**, so disabling it harness-wide is safe:

- Suites asserting a terminal job status all drive execution themselves: `master-inventory-deletion-e2e` and `product-sync-e2e` call `handler.execute(...)`; `connection-reauth-flagging` calls `(runner as any).handleJobFailure(job, error)` directly, which is a method call and not the loop.
- `job-intake-execution` calls `jobRepository.findAndLockDueJobs(...)` itself. Its "Start the SyncJobRunner" text is a comment describing what the suite deliberately does **not** do.
- No suite waits on or polls for a job it never executes.

norbert-kulus-blockydevs:
## Summary

- The self-hosted `test-integration` runner allows multiple `test-integration` jobs to run concurrently, and each job's orphan-sweep step unconditionally ran `docker network prune --filter label=org.testcontainers=true`, believing it was "safe unconditionally" because Docker refuses to remove a network that already has a container attached.
- That's only true for *already-attached* networks. A network created via `new Network().start()` in `prestashop-container.helper.ts` sits unattached — and unlabeled by run id, since testcontainers-node's `Network` builder has no `.withLabels()` — for up to several minutes during a cold MySQL/PrestaShop image pull. A sibling job's sweep could prune it out from under the still-booting job, surfacing as `failed to set up container networking: network <id> not found` on the first container attach.
- Fix: add an `until=15m` age filter to the prune, mirroring the container-half sweep's own age-cutoff philosophy, so a network can never be pruned before it plausibly could have a container attached.

Root-caused after this exact failure hit PR #2756's CI (unrelated frontend-only PR); rerunning that PR's CI after diagnosis (before this fix landed) confirmed it as a flake — the rerun passed clean.

## Second flake: the live `SyncJobRunner` (worker)

`#2825` documents a second, unrelated flake in the same job, and this PR now closes that half too.

- `SyncJobRunner` defaults `WORKER_RUNNER_ENABLED` to `'true'` when unset, and the only place the repo set it to `'false'` was `apps/worker/test/jest-global-setup.cjs`, which is wired to the worker's **unit** Jest config and never to `jest-integration.cjs`. Nothing in `harness.ts`, `setup-global.ts`, `setup.ts` or the CI job env set it either.
- So the real runner polled `sync_jobs` in Postgres about once a second for the whole worker integration run, and could claim a `queued` row a suite had just inserted and was about to drive by hand, executing the same job twice from two directions.
- Fix: set `WORKER_RUNNER_ENABLED=false` in `apps/worker/test/integration/harness.ts`, beside the existing `OL_SCHEDULER_ENABLED=false`. That also makes `setup.ts`'s existing comment true, which already named `WORKER_RUNNER_ENABLED` among the gates keeping background loops quiet.

The blast radius was wider than the one suite `#2825` originally named:

1. `inventory-provenance-backfill-e2e.int-spec.ts` inserts its job with **no `runAfter`**, so the row is immediately due, and `inventory.provenance.backfill` sits in the `bulk` lane. It then asserts exact drain counts after each manual tick at `pageLimit: 5` (`remainingNull` `'7'` -> `'2'` -> `'0'`), so one concurrent claim stamps an extra page and the numbers no longer hold. This is the failure that hit PR #2749's CI (run 33728532711) on a frontend-copy-only diff.
2. `job-intake-execution.int-spec.ts` asserts a row created with `nextRunAt: new Date()` is still `queued`, and that its own `findAndLockDueJobs(10, 'test-worker-123')` claims it under that worker id. A live runner claiming first breaks both.
3. `allegro-cursor-persistence.int-spec.ts` already worked around this race per-suite with a 24h `runAfter` and said so in a module comment. The workaround is kept as defence in depth and its comments are corrected to say the harness now closes the race.

## Test plan

- [x] Reproduced the exact failure locally: created a Testcontainers-labeled network, ran the old unconditional `docker network prune`, confirmed it deleted the unattached network, then confirmed attaching a container to it fails with the identical error (`failed to set up container networking: network ... not found`).
- [x] Verified the fix: repeated the same sequence with `--filter "until=15m"` added, confirmed the young network survives the prune, and a container successfully attaches to it.
- [x] `pnpm lint && pnpm type-check` pass (workflow-YAML-only change; the pre-commit hook's `smart-test` correctly reported no affected workspace package).
- [x] Audited all 26 worker `*.int-spec.ts` files for reliance on the background loop (`#2825` acceptance criterion 2). None relies on it: suites asserting a terminal job status drive execution themselves via `handler.execute(...)` (`master-inventory-deletion-e2e`, `product-sync-e2e`), `connection-reauth-flagging` calls `(runner as any).handleJobFailure(...)` directly, and `job-intake-execution` calls `findAndLockDueJobs(...)` itself.
- [x] Ran the failing suite locally in isolation against unmodified code: 7/7 pass in ~15s, confirming the suite logic is sound and the failure is the harness race.
- [x] Ran the **full** worker integration suite locally with the fix: 26/26 suites, 115/115 tests pass (392s), including the three suites named above.
- [x] CI on this PR is expected to pass.

Closes #2825


norbert-kulus-blockydevs:
Applied fixes from a self-review pass (findings below, ranked by the reviewer as no BLOCKING issues, one IMPORTANT, two SUGGESTION):

1. **[IMPORTANT] "Closes #2825" would have silently auto-closed the issue's second, unrelated, still-open flake.** Issue #2825 bundles two distinct flakes — this network-prune race (fixed here) and a separate worker-side `SyncJobRunner` double-processing race with its own unchecked acceptance criteria that this PR does not address. Changed the PR body's closing keyword from `Closes #2825` to `Part of #2825` so merging leaves the issue open for that second flake to still be tracked.
2. **[SUGGESTION] Tightened the comment's stated risk-window reasoning.** It previously justified the 15-minute grace window by citing the ~5-10 min full PrestaShop-install time, but the network only needs to survive until its *first* container (MySQL) attaches — bounded by that container's own 240s (4 min) `withStartupTimeout`, not the full install duration. The comment now reasons from that correct, shorter bound; the `until=15m` value itself is unchanged.
3. **[SUGGESTION] Added a mention of `woocommerce-container.helper.ts`** alongside `prestashop-container.helper.ts` in the comment — it has the identical `new Network().start()` pattern and is equally covered by this fix, so a future reader shouldn't assume this is PrestaShop-specific.

Commit: 7fd37d558005226c9c11f5d14916d8f9439d9f68

norbert-kulus-blockydevs:
Second review pass confirmed all three prior findings were fixed correctly and surfaced no BLOCKING/IMPORTANT issues — only one further optional nit: the "240s (4 min)" `withStartupTimeout` figure is explicit only in `prestashop-container.helper.ts`; `woocommerce-container.helper.ts` never overrides it and inherits `@testcontainers/mysql`'s ~120s default (still comfortably inside the 15-minute window either way). Fixed in f006e4b82 to state both bounds explicitly instead of implying one figure applies to both helpers.

Review/fix loop is now clean (no BLOCKING/IMPORTANT findings remaining). Ready for human review/merge.

piotrswierzy (approved):
## Tech Lead Review — PR #2828

### Step 1 — Context

Single-file change to `.github/workflows/ci.yml`: the `test-integration` orphan-sweep step's `docker network prune` gains a `--filter "until=15m"` age gate, plus a long comment replacing the (now-falsified) "safe unconditionally" claim.

I verified every load-bearing factual claim in the comment against the repo:

- `apps/api/test/integration/helpers/prestashop-container.helper.ts:179` — `await new Network().start()`, with **no** `.withLabels(...)`, while every *container* builder in the same file does carry `.withLabels(ciRunIdLabels())` (lines 684, 700). So the run-id-precise gate the container half of the sweep uses (`ol.ci.run-id` + `gh api` run-status lookup, ci.yml lines ~263-289) genuinely cannot apply to networks. ✅
- `apps/api/test/integration/helpers/woocommerce-container.helper.ts:56` — same unlabeled `new Network().start()`, same exposure. The comment correctly names both helpers. ✅
- PS MySQL companion: `.withStartupTimeout(240_000)` at line 688 — matches the "240s explicitly for PrestaShop's" claim. ✅
- WC MySQL companion (lines 63-70): no explicit `withStartupTimeout`, so the `@testcontainers/mysql` default applies — matches "~120s ... for WooCommerce's". ✅
- The existing container sweep already uses an age cutoff (`CUTOFF_EPOCH = now - 5400`), so the fix is consistent with the step's own established philosophy rather than a new idiom. ✅

Semantics of the flag are right: `docker network prune --filter until=15m` prunes only networks created *more than* 15 minutes ago, so a freshly created, not-yet-attached network survives — exactly the window the bug needed.

### Step 2/3 — Findings

**No BLOCKING issues.**

**[IMPORTANT]** — `.github/workflows/ci.yml` (sweep comment, "bounded by its own startup timeout — 240s ... ~120s")

> The stated bound is not the real one. In testcontainers-node the **image pull happens inside `start()` and is not covered by `withStartupTimeout`** — that timeout guards the wait strategy once the container is up. On a genuinely cold runner (the exact scenario the PR description cites as the trigger: "cold MySQL/PrestaShop image pull"), network-creation → first-container-attach is bounded by *pull time + 240s*, not 240s. 15 minutes is very likely comfortable in practice for `mysql:8.4.7`, and I'm not asking for a different number — but the comment currently asserts a guarantee that the mechanism does not provide, and this file's comments are unusually load-bearing (the previous comment's confident "safe unconditionally" is precisely what made the bug invisible). Suggest softening to something like "bounded by image pull time plus the container's own startup timeout (240s for PrestaShop's MySQL, ~120s default for WooCommerce's); 15 minutes is chosen as generous headroom over the observed worst case, not as a proof."

**[SUGGESTION]** — same block

> The real fix for the class of bug is labelling the network. The comment notes `Network` has no `.withLabels()` "in this version", which is the right reason to take the age gate now — consider adding a one-line pointer (`// Revisit once testcontainers-node's Network builder supports labels — then this can use the same run-id gate as the containers above`) so a future upgrade has a hook. Optional; #2825 may already carry it.

**Positive observations**

- Root-cause narrative is precise and falsifiable, and the test plan actually *reproduced* the failure locally before and after — not just asserted it. That is the standard this repo's CI-flake work should be held to; the "rerun passed clean" confirmation on #2756 correctly identifies it as a flake rather than accepting a green rerun as evidence of nothing being wrong.
- `|| true` retained, so the step stays non-fatal under `set -eo pipefail`, consistent with the surrounding sweep.
- No source, schema, or architectural layer touched — no migration, port/adapter, or test obligations apply (`docs/testing-guide.md`, `docs/migrations.md` both non-applicable). The absence of a `*.spec.ts` here is correct, not a gap.

### Step 4 — Final Assessment

**Summary.** A small, well-root-caused CI fix that closes a real cross-job race: an unlabeled, unattached Testcontainers network created by a still-booting sibling job could be pruned out from under it, producing `failed to set up container networking: network <id> not found`. The mechanism is verified, the flag semantics are correct, and the change is consistent with the container half of the same sweep. The only concern is that the explanatory comment overstates the bound it relies on (image pull is not inside `withStartupTimeout`) — a documentation-accuracy issue in a file where a previously overconfident comment is what hid this bug in the first place.

**Merge readiness:** 🔄 **Approve with changes** — merge once the comment's bound claim is softened; the behavioural change itself is correct as-is.

**Priority fixes**
1. Correct the "bounded by its own startup timeout" sentence to acknowledge image-pull time (IMPORTANT, comment-only).
2. Optionally add the "revisit when `Network` supports labels" pointer (SUGGESTION).

norbert-kulus-blockydevs:
Addressed both points from the review.

Reworded the comment near the network prune step so it no longer claims the 240s/120s startup timeouts are the bound on the unattached-network window. testcontainers-node pulls the image inside `start()`, before `withStartupTimeout`'s wait-strategy phase even begins, so on a cold image pull the real bound is pull time plus the startup timeout, not the startup timeout alone. The comment now says the 15-minute `until=` window is chosen as generous headroom over the observed worst case, not a proven bound - which is the same posture the previous "safe unconditionally" claim should have had and didn't, so it seemed worth being explicit about this time.

Also added a one-line forward-pointer next to the `.withLabels()` note: once testcontainers-node's `Network` builder supports labels, this can move to the same run-id-precise gate the container sweep above already uses.

No logic change - `docker network prune ... --filter "until=15m"` is untouched. Lint and type-check pass clean, as expected for a workflow-comment-only edit.

piotrswierzy (approved):
## Re-review — #2828 @ `4294434` (now covers both flakes)

Scope has grown from one file to three: the `ci.yml` network-prune gate plus the `WORKER_RUNNER_ENABLED` half (#2825's second flake), which was not in the diff I approved.

### 1. Verification of the prior finding

**[IMPORTANT] The comment asserted a bound `withStartupTimeout` does not provide — RESOLVED, and resolved precisely.** The comment now reads: *"NOT bounded by `withStartupTimeout` alone — that timeout only guards the wait-strategy phase once the container is already running, and testcontainers-node pulls the image inside `start()` BEFORE that phase starts, so on a cold runner the real bound is image-pull-time plus the startup timeout"*, and the 15-minute window is now stated as *"generous headroom over the observed worst case rather than a proven bound"*. That is the honest version: it names the mechanism, keeps the two real timeouts as the components they are, and stops claiming a guarantee. Given that a previous overconfident comment in this exact block is what hid the bug, this is the change that mattered most.

**[SUGGESTION] "Revisit when `Network` supports labels" pointer — RESOLVED**, inlined at the point where the missing `.withLabels()` is explained, so the future upgrade has a hook where a reader will hit it.

### 2. Review of the new material (the `SyncJobRunner` half)

I verified the diagnosis rather than taking it:

- `sync-job.runner.ts:133` — `configService.get<string>('WORKER_RUNNER_ENABLED', 'true') !== 'false'`, i.e. defaults **on** when unset. ✅
- `apps/worker/jest-integration.cjs` (the integration config) declares `globalSetup`/`globalTeardown` and **no `setupFiles`/`setupFilesAfterEach`** — so `apps/worker/jest.setup.ts`, which does set the gate, never loads for integration. The gate was genuinely unset for the whole integration run. ✅
- Setting it in `harness.ts` (globalSetup) reaches the workers by the same mechanism `DB_HOST`/`REDIS_PORT` already rely on in that identical function, so the propagation is proven by everything else in the file rather than assumed. `maxWorkers: 1` there too. ✅
- Nest's `ConfigModule` does not overwrite an already-present `process.env` value from a `.env` file, so the `'false'` set here wins over `apps/worker/.env.example`'s `WORKER_RUNNER_ENABLED=true` shape if such a file ever exists on a runner. ✅
- The placement is beside `OL_SCHEDULER_ENABLED = 'false'`, and it retroactively makes `setup.ts:39`'s existing comment (*"the per-service enable gates (`WORKER_RUNNER_ENABLED`, `OL_SCHEDULER_ENABLED`, …) keep background loops quiet"*) true — that comment was previously a false statement about this repo, which is the same defect class as the `ci.yml` comment this PR started from. Nice symmetry, and worth noting the PR found it rather than editing the comment to match reality.
- The `allegro-cursor-persistence` workaround is kept as defence in depth with corrected comments, rather than deleted on the strength of the new gate. Right call: a suite that documented a race should not lose its guard because a global now also covers it.

**No BLOCKING issues.**

**[IMPORTANT — non-blocking, worth one line before merge]** `apps/worker/test/jest-global-setup.cjs` still exists, still sets both `WORKER_RUNNER_ENABLED=false` and `WORKER_INTAKE_ENABLED=false` under a header reading *"Jest Global Setup (Worker Integration Tests)"* — and a repo-wide grep excluding `node_modules` finds **zero** references to it. It is not, as the PR description says, "wired to the worker's *unit* Jest config"; it is wired to nothing at all. That file is the artifact that made this bug invisible: anyone auditing "are background loops off in worker integration tests?" reads it, sees both gates set, and stops. Please delete it in this PR (or, if it is kept deliberately, say in it that it is unused). Leaving a dead file that asserts a false configuration, in the very PR whose thesis is that a confidently-wrong comment cost two CI flakes, is the one loose end here.

**[SUGGESTION]** That dead file names **two** gates; this PR restores one. `WORKER_INTAKE_ENABLED` also defaults to `'true'` (`job-intake.consumer.ts:69`), so the live `JobIntakeConsumer` polls the `jobs.sync` stream for the whole integration run — the same structural race one loop over. No flake is attributed to it and `job-intake-execution.int-spec.ts` explicitly does not depend on it (line 74: *"In a real test, we'd need to start the JobIntakeConsumer"*), so I would not gate merge on it. But either setting it in `harness.ts` too, or adding a sentence there saying why the intake loop is deliberately left running, would close the audit properly — otherwise the next person repeats exactly the investigation this PR just did.

**Positive observations**
- The blast-radius section is the strongest part of the description: it names three suites, explains *why* each is sensitive (immediately-due row + exact drain counts at `pageLimit: 5`; `nextRunAt: new Date()` + own-worker-id claim), and ties each to a real CI run on an unrelated frontend-only diff. That is what makes this a root cause rather than a plausible story.
- Full worker integration suite run locally with the fix (26/26 suites, 115/115) plus the failing suite in isolation against unmodified code — both directions, which is the standard flake work should meet.
- No production source touched; no test obligation, migration, or layering concern applies.

### 3. Final assessment

The prior comment-accuracy finding is fully addressed, and the added `SyncJobRunner` half is correct, minimal, placed where the existing sibling gate lives, and verified against the actual config wiring rather than assumed. Merge once `main` is rebased in (`mergeable_state: behind`).

**Before/at merge**
1. Delete (or annotate) the unreferenced `apps/worker/test/jest-global-setup.cjs` — IMPORTANT, non-blocking.
2. Optionally close the `WORKER_INTAKE_ENABLED` half, or record why it stays on.

norbert-kulus-blockydevs:
Addressed both non-blocking findings from the second review (commit 54cd7e1a5):

1. Deleted `apps/worker/test/jest-global-setup.cjs` — confirmed via repo-wide grep it was unreferenced by any Jest config, and its "wired to the worker's unit Jest config" claim was false. Kept a corrected reference to it in the `harness.ts` comment explaining why the file's existence previously hid the bug.
2. Set `WORKER_INTAKE_ENABLED = 'false'` alongside `WORKER_RUNNER_ENABLED` in `harness.ts`, closing the identical structural race for `JobIntakeConsumer` (also defaults to `'true'`, also polls a live stream for the whole integration run, also has no int-spec depending on it running).

Rebased on latest `main` is still pending on this branch — will sync before merge if needed.

Issue #2807, opened 2026-09-02
[BUG] Frontend — Order-detail sales document panel loses mockup's timeline, color discipline, and empty-state CTA


norbert-kulus-blockydevs:
## Problem / Context

Epic #2563 (M10 e2e verification) added a Playwright spec for the order-detail "Sales document" panel (`apps/web/src/features/orders/components/sales-document-panel.tsx`) against the design mockup at `docs/plans/mockups/sales-document-routing.html`. The spec passes, but a manual screenshot-vs-mockup comparison found the real panel diverges significantly from the mockup's intended design in five ways:

- **Layout hierarchy is inverted.** The mockup dedicates the entire page to this one panel, full-width, as the single focus of the screen. The real app buries it as one narrow card among roughly six equal-weight cards (Delivery, Sync status, Summary, etc.) in a dense two-column operations page — it reads as a minor detail rather than the point of the screen.
- **Missing the timeline.** The mockup's filled state shows one clean horizontal stepper — Issued -> Sent to KSeF -> KSeF answer — each step with its date directly underneath, scannable in one glance. The real panel instead shows separate checkmarked key-value rows ("Issued" / "Cleared") plus a third nested green box titled "KSEF · NATIONAL E-INVOICING SYSTEM" repeating the same clearance fact again — three places stating overlapping information instead of one timeline component.
- **Color discipline is lost.** The mockup uses exactly two tints: light blue for the neutral "waiting for an answer" info callout, light yellow for the duplicate-record warning. The real panel adds green success blocks, status badges, and multiple background colors, which is busier than the mockup's "color marks exceptions only" rule.
- **The empty state changes intent entirely.** The mockup's empty state: one red icon, one headline ("No document · No routing"), one explanation block ("Routing could not decide: nothing is issued, neither this country nor Rest of world has a rule or a default"), and two buttons — a primary "Set routing for {country}" (sends the operator to fix the actual cause in Settings) and a secondary "View all routing". The real panel instead drops the operator straight into a manual-override form with two connection dropdowns and two "Issue anyway" / "Register anyway" buttons for both document kinds simultaneously — it never states that routing failed to decide, and never links back to the Settings page where the real fix lives.
- **The headline is smaller/secondary** versus the mockup's large document-icon + name + status wordmark, which is also what gives the `/orders` list row (a separate, already-verified-correct surface) its visual continuity into this panel — the order-detail panel currently breaks that continuity.

## Proposed Solution

[...]

PR #2809, opened 2026-09-02
feat(sales-documents): one market table with the Connected providers section demoted below it, an order-detail panel that leads with a routing-fix CTA instead of a bare manual-override form, and e2e coverage for all three sales-document surfaces (#2563, #2806, #2807)

norbert-kulus-blockydevs:
## What changed and why

Two UI fixes to the sales-document routing surfaces, verified end-to-end against the committed
mockup (`docs/plans/mockups/sales-document-routing.html`), plus the e2e suite that verifies them.

1. **Settings market list + routing dialog** (#2806) — the market table and the old per-country
   rule-engine table were duplicated on-page, row copy was technical fragments instead of plain
   sentences, and unconfigured rows used a full-row yellow tint instead of a status dot. Fixed:
   reorganised to the mockup's two-section layout — ONE market table (filter chips replacing what
   used to be two separate sections), with the "Connected providers" table demoted below it.
2. **Order-detail sales-document panel** (#2807) — the empty state dropped straight into a manual
   override form instead of leading with "routing could not decide" and a link back to Settings.
   Fixed: the empty state now leads with a primary "Fix routing settings" CTA when a routing-block
   reason exists, and the manual override is demoted behind a closed disclosure.
3. **Orders-list money-cluster cell** — matched the mockup as-is, no fix needed. Originally verified
   on a separate branch (#2769); consolidated here so there is one e2e PR for the feature area, and
   #2769 is closed as superseded.

## Review round (2026-09-06)

Both review rounds are addressed in `64aa78e`. The four blockers:

- **220 committed scratch files**, one carrying a live `PHPSESSID` and one a hardcoded admin JWT —
  `apps/e2e/verify-scratch/` is deleted, along with the stray `apps/web/apps/e2e/…` copy a script
  wrote from the wrong cwd. The `.gitignore` rule is widened so it matches what its own comment
  promises (it matched only `__screenshots__/` while the same commit added PNGs under
  `screenshots/`). **Both credentials still need rotating** — they are in the branch history
  regardless of this commit.
- **The `chooseSalesDocumentDecision` fallback retirement is REVERTED here** and moves to its own
  PR. The reviewer is right that it is an undeclared breaking change to a contract ADR-041,
  `architecture-overview.md` §14 and §17 all still describe as live — it needs its own ADR
  amendment and an explicit decision on point 2c (whether an unconfigured install should persist an
  attention-counted `unresolved-routing` block on every ingested order). Reverted with it: the four
[...]

Issue #2806, opened 2026-09-02
[BUG] Frontend — Settings sales-documents page duplicates rule-engine + market UI, diverges from mockup


norbert-kulus-blockydevs:
## Problem / Context

Parent epic: #2563 (M10 e2e verification against `docs/plans/mockups/sales-document-routing.html`).

The Settings "Sales documents" page (`apps/web/src/features/sales-documents/components/sales-document-rule-engine-panel.tsx`, `sales-document-market-section.tsx`, `sales-document-country-routing-dialog.tsx`) passes its own e2e spec but diverges significantly from the mockup on manual visual comparison.

- Two competing UIs are stacked on one page. The mockup shows exactly two tables: "What each market issues" (market status) then "Connected providers" directly under it. The real page additionally renders the OLD per-country rule-engine editor (rule composer, starter-template card, a separate country-defaults table with columns COUNTRY/RULES/INVOICE DEFAULTS TO/RECEIPT DEFAULTS TO/STATUS) ABOVE the new market table, so a country like DE or PL appears twice: once as a market row, once again in the country-defaults table below.
- The "Connected providers" table's columns diverge from the mockup, not just its styling. Mockup: PROVIDER / ISSUES / GOES FIRST / ISSUES WHEN / CONNECTION. Real page: PROVIDER / STATUS / CAPABILITY / ISSUES / PRIMARY / TRIGGER — a different, more technical model, and it is buried far down the page under the rule-engine noise instead of sitting directly under the market table like the mockup shows.
- Row copy tone diverges. Mockup writes plain sentences ("3 of your 4 markets are issuing nothing right now"). The real page uses technical fragments ("PL · 0 rules · 3 orders in the last 30 days").
- Visual noise: the mockup uses a small colored status dot on an otherwise plain white row. The real page tints the entire row background yellow for an unconfigured market, adds toggle switches, and stacks a large "Suggested starter template" rule-builder card into the same view.
- No "Add a market" affordance matching the mockup's simple button exists — the real page instead has a bare "ADD COUNTRY" text input wired into the rule-template flow.
- (Informational only, not actionable by us) the mockup's own bundled JS throws `REST_OF_WORLD is not defined` when opening its routing dialog via the "Fix routing" trigger — a bug in the mockup file itself, worth flagging to whoever maintains it, but out of scope for this fix.

## Proposed Solution

[...]

PR #2720, opened 2026-08-31
docs: sales documents — decisions and documentation (D2+D3)

norbert-kulus-blockydevs:
Merges the two remaining mini-epics of the sales-documents docs programme: architecture/standards updates (D2) and operator documentation (D3). D1 (ADRs) already merged straight to `main` earlier.

**Landing order:** #2568 and #2570 have both merged to `main`, so this PR's base is now `main` directly — merging it lands D2+D3 on `main` with no further stacking on this end. #2749 (a code-side billing→delivery copy fix) is stacked on top of this branch and should be retargeted to `main` and merged after this one.

## D2 — architecture and standards (#2505)
- #2506 — architecture-overview: sales-document read model and market discovery
- #2507 — style guide: the document cell, the new primitives, and colour marks exceptions
- #2508 — lessons.md: designing a surface against prose instead of against code

## D3 — operator documentation (#2509)
- #2510 — how OpenLinker decides which document a sale gets
- #2511 — what each document state means and what to do about it
- #2512 — setting up a market, including the Poland template caveat

Closes #2499
Closes #2505
Closes #2506
Closes #2507
Closes #2508
Closes #2509
Closes #2510
Closes #2511
Closes #2512

PR #2831, opened 2026-09-02
docs(user-guide): Sales Documents tutorial + fiscal-receipts index fix

### LINKED ISSUE #2830: [TASK] DX — Operator tutorial for Sales Documents (routing), plus doc-index updates
## Problem / Context

The sales-document routing feature ([ADR-041](./docs/architecture/adrs/041-sales-document-routing-policy.md), shipped in PR #2809) — per-market rules, country defaults, the ★ Rest of world fallback, and the order-level Sales Document panel — has no operator-facing tutorial anywhere. `docs/user-guide/` carries a numbered tutorial for every other major admin-UI surface (`01-overview.md` through `08-settings-and-admin.md`, plus `04a-fiscal-receipts.md`), but nothing for Sales Documents, and `docs/user-guide/README.md`'s own section index never listed the pre-existing `04a-fiscal-receipts.md` tutorial either — a second, independent gap in the same index.

## Proposed Solution

1. Write `docs/user-guide/04b-sales-documents.md`, following the structure and tone of `docs/user-guide/04-invoices.md`: Prerequisites, the Settings → Sales documents market list, the four-tier routing dialog (Rules → Country default → ★ Rest of world → Unresolved), the rule composer's three condition-field shapes (buyer-tax-ID yes/no, order-country free text, order-total-gross comparison-against-a-named-threshold), a worked multi-tier example (e.g. a real 3-rule setup mixing buyer-tax-ID and order-total conditions), the "No sales document, by design" acknowledgment flow, and every state of the order-detail Sales Document panel (unresolved/no-config, not-issued with the manual-override cards, in-progress/registering, both invoice lifecycle states — awaiting clearance and cleared — and a registered fiscal receipt). Close with a plain-English explainer of the exact auto-issue decision order, explicitly calling out two properties that are easy to assume wrongly: (a) a connection's `isPrimary` flag never causes silent routing — only rule/default/Rest-of-world resolution does; (b) two matching rules hold the order rather than picking one.
2. Screenshots for the new page must be cropped to content only (no navbar/sidebar chrome) and must mask any real personal name appearing in test data — this is a hard requirement since the guide is intended for publication on docs.openlinker.io. Reuse real screenshots taken during live e2e verification against `ol-demo-fresh` rather than staging synthetic ones, where available.
3. Update `docs/user-guide/README.md`'s numbered section list to add an entry for the new `04b-sales-documents.md`, and — separately — add the missing entry for the pre-existing `04a-fiscal-receipts.md`, which the index has never linked.
4. Audit `README.md` (repo root) and `docs/getting-started.md` for whether either needs a new individual link to the tutorial. Both currently link generically to the Operator Guide entry point (`docs/user-guide/README.md:361` and `docs/getting-started.md:294`/`:344`) rather than enumerating every sub-page, so — unless that changes — neither needs a new link. Record the outcome of this check explicitly in the PR description rather than silently skipping it.

## Classification

**Type**: DX (documentation)
**Layer**: N/A — no `libs/`, `apps/api`, or `apps/worker` code changes are in scope
**File(s)**:
- `docs/user-guide/04b-sales-documents.md` (new)
- `docs/user-guide/README.md`
- `docs/user-guide/images/04b-*.png` (new)
- `README.md`, `docs/getting-started.md` (audit only, edit only if the audit says so)

## Dependencies

None — the feature this documents (PR #2809, ADR-041) is already merged/live.

## Assumptions

- "Available on docs.openlinker.io" means this markdown file is the source that site's publishing pipeline picks up from `docs/user-guide/`, matching every other tutorial already there — no separate site-specific format is needed.
- Screenshots may be sourced from an existing e2e verification pass against a live demo stack rather than freshly staged, provided they are cropped and any real name is masked before being committed.

## Acceptance Criteria

- [ ] `docs/user-guide/04b-sales-documents.md` exists and covers: market list, four-tier routing dialog, rule composer's three condition shapes, a worked multi-tier example, the acknowledgment flow, and every order-panel state listed above
- [ ] Every screenshot embedded in the new page is cropped to content only (no navbar/sidebar) and contains no real personal name
- [ ] `docs/user-guide/README.md` lists both `04b-sales-documents.md` and the previously-unlisted `04a-fiscal-receipts.md`
- [ ] `README.md` and `docs/getting-started.md` are either updated with a new link, or the PR description states explicitly why neither needed one
- [ ] All internal markdown links and image paths in the new page resolve (no broken links)
- [ ] No architecture boundary violations (N/A — docs-only change)

norbert-kulus-blockydevs:
## Summary

Adds `docs/user-guide/04b-sales-documents.md` — an operator walkthrough of the Settings → **Sales documents** screen and the per-order **Sales document** panel — and fixes three defects the work surfaced in neighbouring docs.

Rewritten after review. The original draft described a UI that does not ship on `main`: seven of its twelve screenshots were captured against the unmerged `2563-m10-order-detail-panel` branch (#2807), which adds 252 lines to `sales-document-panel.tsx`. Every claim is now checked against `main` and every screenshot re-captured from a build serving `main`'s bundle.

## What the page says

- The market list: the summary line, ordered-and-highlighted rows (there are no filter chips), the configured-countries table, **Add country** including `*` for ★ Rest of world, and the Poland starter template.
- The four-tier routing dialog, with **Tier 3 stated as contextual** — *Falls through to ★ Rest of world* on an unconfigured market, *An unmatched order is held* on a configured one, and ★ Rest of world's own dialog showing only three tiers.
- The rule composer's real labels: Conditions / Document type / Integration / Effective from / Effective to (optional).
- A worked three-rule Poland example, with the buyer-tax-ID banner and its caveat.
- **Setting both country defaults disables the fallback** rather than being permitted.
- Every order-panel state on `main`: unresolved, issue-by-hand (inline cards, no disclosure), registering, awaiting clearance, cleared, and a registered fiscal receipt.
- How auto-issue decides, including that **"primary" is a last resort and never a routing input** — a market with nothing configured anywhere still issues through the sole eligible connection, or the primary one where several are eligible.

It links [How routing decides](../docs/sales-documents-how-routing-decides.md), [Setting up a market](../docs/sales-documents-setting-up-a-market.md) and [Sales document states](../docs/sales-documents-state-reference.md) rather than restating them.

## Other fixes

- `docs/user-guide/README.md` — `4a.` / `4b.` are not valid ordered-list markers; they terminated the list and restarted numbering. Now nested under `4.`.
- `docs/sales-documents-how-routing-decides.md` — drops the stale *"proposed but has no code yet"* note (the screens ship), corrects the PrestaShop-only buyer-tax-ID claim, and narrows the *"not yet distinguished"* note now that *Two rules matched* carries its own reason.
- `docs/getting-started.md` — **AC #4 audit outcome.** The audit found a pre-existing broken link: Diagnostics moved to `07-diagnostics.md` when Orders took `06`, and `getting-started.md` still pointed at `06`. Fixed, along with a stale four-item enumeration of what the user guide covers. Historical plan documents under `docs/plans/` still name `06-diagnostics.md` and are deliberately untouched — they record what was planned at the time, not live navigation.

## Test plan

- [x] Every image path, cross-doc link and anchor verified to resolve
- [x] All 11 screenshots re-captured from a build serving `main`'s bundle, cropped to content (no navbar or sidebar), and free of customer data
- [x] `README.md` list renders as one ordered list with the two pages nested under Invoices
- [x] `README.md` / `getting-started.md` audited per AC #4; outcome recorded above
- [ ] Companion `openlinker-docs` PR adds this page and `04a-fiscal-receipts.md` to `scripts/sync-docs.mjs`'s `SOURCES` + `astro.config.mjs`'s sidebar

Closes #2830 (product-repo half; the `openlinker-docs` half is tracked on the issue).


piotrswierzy (commented):
The structure, tone and screenshot discipline are all right, and it correctly nails several subtle properties (no rule priority — two matches hold the order; one originating document per order; receipts have no correction primitive). But a tutorial's one job is that every instruction is true of the shipped product, and a number of these aren't. I'd hold merge.

**Blocking — behaviour**

1. `04b:265-272` — the `isPrimary` claim is **inverted**, and it's the property #2830 specifically commissioned this page to get right. `choose-sales-document-decision.ts:74-88` returns the rule-engine answer as-is *except* for `no-configuration-for-country`, which falls through to `resolveSalesDocumentRoutingFromCandidates`. So a market with no rule, no default and an unconfigured Rest of world **does** issue — on the sole eligible connection, or on the `isPrimary` one when several are eligible (`resolve-sales-document-routing.ts:180-192`). The UI says so itself: `sales-document-block-copy.ts:149` tells the operator "Set a primary" precisely so OpenLinker will issue automatically. Only "no effect on rule or default evaluation" survives.
2. `04b:152-156` — Tier 4 implies a *configured* market falls through to Rest of world. It doesn't: `evaluate-sales-document-rules.ts:232` guards the ROW tier on the country carrying nothing, and the shipped panel copy states it flatly — "It does not pass to Rest of world." Your own step 3 gets this right; Tier 4 must match.
3. `04b:127-130` — "an invoice default, a receipt default, or both" reads as permission. Setting both makes tier 2 refuse to choose (`evaluate-sales-document-rules.ts:193-198`) and returns `ambiguous-connection-no-primary`, so **every unmatched order in that market is held**. `docs/sales-documents-how-routing-decides.md` calls this "the single most common way a market ends up issuing nothing while looking configured". Say the consequence.

**Blocking — controls that don't exist**

- The four filter chips (`:46,48-50`): there are none. `sales-document-market-section.tsx:117` renders a plain unfiltered `<ul>`; decisions-needed rows are *ordered* first by `orderSalesDocumentMarkets`, never filtered. The component docblock even notes it replaced the chips selector.
- "Document & destination" and "Effective window" (`:72,79,82`): the dialog's actual labels are **Document type** (:254), **Integration** (:270), **Effective from** / **Effective to (optional)** (:303-316). "Conditions" is real.
- "Fix routing settings" (`:184,186`): not a control — the `no-configuration-for-country` copy is plain prose with no link or button (`sales-document-block-copy.ts:164-171`). And that entry sets `offerSetPrimary: false`, so **"Set a primary" is absent from exactly the screenshot being annotated**. Describing two actions where the screen offers none is the worst case for a tutorial.
- "Issue or register manually instead" (`:196`) and "Send by email" (`:226`): neither string exists. Nearest real copy is "Pick a connection above to issue it by hand."; there is no email action in the panel at all.
- `:55` — the button is **Add country**, not "Add a market" (`sales-document-country-index.tsx:173`).

**Important**

- `:57` "Configure" is conditional — a row needing a decision renders **Set up** or **Use starter setup** (`sales-document-market-row.tsx:55-58`), i.e. wrong for exactly the first-run reader.
- `:106` the banner reads `3 rules read the buyer's tax ID` (drop "of these"), and its tone is `warning`, not the reassuring "standing reminder" implied.
- `:109-121` `net-priced-order` is an internal slug — operators see **"Order is net-priced"**, and the module's own rule is that the reason literal never reaches the screen. "Check with your OpenLinker operator or the issue tracker" is also a non-remedy in a self-hosted guide where the reader *is* the operator; the shipped remedy is "use a condition that does not depend on the total".
- `threshold-currency-mismatch` is missing entirely, though the worked example is built on PLN thresholds and amounts are never converted when routing decides.
- `:24-25` "only PrestaShop-sourced orders carry a buyer tax ID" is out of date — Allegro, WooCommerce and Erli order sources all call `readSourceBuyerTaxId` today. (`sales-documents-how-routing-decides.md` needs the same fix.)
- `README.md:25-30` — `4a.` / `4b.` aren't valid ordered-list markers; they terminate the list, so both new entries render as plain paragraphs and their continuation lines merge in.
- The PR body doesn't record the `README.md` / `docs/getting-started.md` audit outcome that AC #4 of #2830 explicitly requires "rather than silently skipping it" — only the `openlinker-docs` half is covered.
- Please also drop the now-stale "proposed but has no code yet" note at `docs/sales-documents-how-routing-decides.md:9-11`. Two operator docs disagreeing about whether a screen exists is worse than either alone.

**Smaller**

- `:224` "Cleared" is a valid headline, but `invoice-timeline.tsx:17` documents the terminal clearance label as **"Accepted", never "Cleared"** — and your own alt text says ACCEPTED.
- `:203-205` "issuing an invoice retires the receipt card (and vice versa)" is half true: an existing invoice hides the fiscal slot, but an existing receipt keeps its slot and *blocks* invoicing with "This order already has a document" (`sales-document-panel.tsx:1243`). Different affordance.

Worth knowing: CI is no evidence here — Lint/Test/Build were all path-skipped, and `check-ui-vocabulary` only scans three feature folders' `.tsx` / `*.copy.ts`, never docs. Every quoted label needs a grep before it ships.

Genuinely good: the buyer-tax-ID prerequisite correctly distinguishing "not asserting" from "known to have none" (the exact three-state semantics `buyer-tax-id.types.ts` implements, and easy to flatten), and every image path and internal anchor resolving.

piotrswierzy (changes requested):
# Code Review: Sales Documents tutorial + fiscal-receipts index fix (#2830)

**Scope**: `docs/user-guide/04b-sales-documents.md` (new, 283 lines), `docs/user-guide/README.md` (+6), 12 PNGs. Reviewed against `main`.

Reviewed on the axis that matters for an operator tutorial: **is every claim true of the shipped product**, checked against `libs/core/src/sales-documents/` and `apps/web/src/features/sales-documents|orders/` rather than assessed as prose.

---

## 1️⃣ Summary

**Approach**: ✅ The right page to write, structured the right way. The four-tier ladder, the worked Poland example, the per-order panel states and the closing "how auto-issue actually decides" recap are exactly what an operator needs, and the writing is unusually careful about the product's own reticence (*"OpenLinker never decides what a sale legally requires"*, *"two matching rules never mean pick one"*, the receipt panel's finality note).

**Concern**: two claims are not true of the shipped code, and both are of the worst kind for a tutorial — they tell the operator to do something the product will not do. One of them advises a configuration that **silently stops every unmatched order in that market from issuing anything**.

---

## 2️⃣ What I verified as correct ✅

A lot of this page holds up under checking, and that is worth recording:

| Claim | Verdict |
|---|---|
| Only PrestaShop-sourced orders report a buyer tax ID; Allegro/WooCommerce are *"not asserting"*, never *"known to have none"* | ✅ Exactly right, and the three-state distinction is stated correctly |
| Three condition fields today: buyer-has-tax-ID, order country, order total (gross) | ✅ |
| Order total compares against a **named threshold**, never a free number | ✅ Matches the `thresholdRef` design |
| Conditions are AND-combined | ✅ |
| Two matching rules → order held, never guessed | ✅ `conflicting-rules-equal-priority` exists; #2170 deliberately shipped no `priority` column, so any two matches are ambiguous |
| The `net-priced-order` unresolved reason and its cause | ✅ Reason exists; panel copy: *"the comparison cannot be made, so the order is held"* |
| ★ Rest of world's own dialog renders **three** tiers, an ordinary market **four** | ✅ Confirmed verbatim in `sales-document-country-routing-dialog.tsx`'s header, including the "never a fourth, it cannot fall through to itself" reasoning |
| Default connection pickers filtered by capability (`Invoicing` / `Fiscalization`) | ✅ |
| One originating document per order — issuing an invoice retires the receipt card | ✅ Matches the #2047 write-path guard |
| A fiscal receipt has no correction primitive | ✅ Consistent with `FiscalizationPort` |
| Effective-window semantics | ✅ `effectiveFrom` / `effectiveTo` behave as described |
| **All 12 image paths resolve**; all four cross-doc links and all four anchors (`#prerequisites` ×2, `#order-detail`, `#issuing-a-correction`) resolve | ✅ Checked individually |

---

## 3️⃣ Issues

### 🔴 BLOCKING 1 — the tutorial permits a country-default configuration that makes every unmatched order unresolved
`docs/user-guide/04b-sales-documents.md` § Tier 2 — Country default

> You may set an invoice default, a receipt default, **or both** — but only one of each per market, since setting a default for both document kinds for the same market would leave nothing to discriminate between them for an unmatched order.

The sentence contradicts itself — the trailing clause argues that "both" is unworkable, immediately after permitting it — and the permission is the half that is wrong.

`evaluateSalesDocumentRules` (`libs/core/src/sales-documents/domain/domain-services/evaluate-sales-document-rules.ts`) resolves a country default like this:

```ts
if (defaults.length === 1) { return { kind: 'route', … }; }
if (defaults.length  > 1) { return { kind: 'ambiguous-defaults' }; }
```

and `ambiguous-defaults` maps to `unresolved / ambiguous-connection-no-primary`. There is a spec pinning it by name: *"should resolve unresolved/ambiguous-connection-no-primary when both per-kind defaults are configured and no rule discriminates"*.

The product is stricter still — the routing dialog **already warns about this configuration in the UI**. From `sales-document-country-routing-dialog.tsx`:

> `2. Country default — SalesDocumentCountryDefaults (+ a dual-default warning Alert when both Invoice and Receipt defaults are set — an open resolution question, not something to silently accept)`

So an operator who follows this tutorial sets both defaults, dismisses an in-app warning the tutorial gave them no reason to take seriously, and every order in that market that no rule matches issues **nothing** — with a reason (`ambiguous-connection-no-primary`) whose copy points them at primaries rather than at the default they were told to set.

The schema does permit both rows (`UQ_sales_document_country_defaults_country_kind` is on `(country, document_kind)`), which is presumably where "or both" came from — but *storable* and *resolvable* are different questions, and this tier is about the second. Please rewrite to state that setting both leaves the market ambiguous for any unmatched order, and mention the in-app warning so the two agree.

### 🔴 BLOCKING 2 — "Fix routing settings" is not an affordance the order panel has
`docs/user-guide/04b-sales-documents.md` § Unresolved — no configuration for this market

> - **Fix routing settings** jumps straight to that market's dialog in Settings.
> - **Set a primary** is a quick, order-independent way to designate one connection…

The screenshot alt-text likewise describes *"with 'Fix routing settings' and 'Set a primary' actions"*.

`sales-document-panel.tsx` renders that state as one `<details>` disclosure wrapping a single `<Alert tone="warning">` whose `action` is exactly one control:

```tsx
action={ setPrimaryTarget ? (
  <Link … to={`/connections/${setPrimaryTarget.id}/edit`}>
    {t('invoice.panel.setPrimary', 'Set a primary')}
  </Link>
) : undefined }
```

There is **no "Fix routing settings" string anywhere in `apps/web`**, and no link from this panel to `/settings/sales-documents` (the only links to that route are the settings tile, the who-decides row, and the connection detail's read-only section). Note also that the single action that *does* exist goes to the **connection edit page**, not to a market's routing dialog — so "jumps straight to that market's dialog in Settings" would not be an accurate description even if the button existed under another name.

Separately, the alt-text quotes the empty state as *"Not issued: no rules configured for this country"*; the shipped copy for `no-configuration-for-country` is **"No routing anywhere"** / *"Neither this country nor Rest of world has a rule or a default. Nothing is issued here, and nothing will be until you set routing."* Please re-take or re-caption from the current build — a tutorial that quotes UI strings the product does not render is the fastest way for a reader to conclude they are on the wrong version.

---

### 🟡 IMPORTANT 3 — `4a.` / `4b.` are not ordered-list items and will break the README list
`docs/user-guide/README.md`

```markdown
4. **[Invoices](./04-invoices.md)**
   …
4a. **[Fiscal receipts](./04a-fiscal-receipts.md)**
```

`4a.` is not valid ordered-list marker syntax — most CommonMark renderers treat it as a paragraph, which terminates the list, and the following `5.` then restarts numbering at `1.`. Since the whole point of the change is a **section index** and the companion `openlinker-docs` PR will render this through Astro, the numbering is load-bearing. Safer shapes: nest them as sub-items under `4.`, or use `4.` / `5.` / `6.` and renumber, or drop the marker and make them indented bullets. Worth rendering once before merge.

### 🟢 SUGGESTION 4 — the "3 of these rules read the buyer's tax ID" banner
The page describes this banner as *"a standing reminder … it isn't a warning that something is wrong"*, which is the right framing. Consider adding the one-line consequence an operator actually needs: on an install whose only order source is Allegro or WooCommerce, **every** such rule is permanently unmatchable, so a market configured entirely out of tax-ID rules will always fall to Tier 2/3. That follows from the Prerequisites section but is the practical failure this banner exists to pre-empt.

### 🟢 SUGGESTION 5 — the worked-example table's first row
The Poland table's first row (`No | (any) | (falls through — no rule matches…)`) is a non-rule occupying a row in a table of rules, which reads oddly against the alt-text's claim of *"three real rules"*. Consider moving it below the table as a sentence.

---

## 4️⃣ Positive observations

- **The mockup obligation is met in spirit**: `docs/frontend-architecture.md § UX Mockups` requires a feature's docs to stay in sync with shipped UI, and this page is genuinely written from the running product rather than from a spec — which is exactly why the two divergences above are worth fixing rather than shrugging at.
- **The known-limitation note on net-priced orders is excellent.** Naming the unresolved reason (`net-priced-order`) so an operator can match what they see on the order to what they read here is the right level of detail, and admitting the gap rather than papering over it is the correct call.
- **"An order never silently routes through a connection just because it's marked primary"** is a precise and non-obvious statement that matches the code, and it is the misconception most likely to bite.
- **Masking the real customer name and cropping to content only** — thank you; that is easy to skip and hard to undo once merged.

---

## Final Verdict

**Status**: ❌ **REQUEST CHANGES**

Structurally this is the tutorial the feature needed, and most of it checks out against the source. Two claims must be corrected before it ships, because both instruct the operator toward something the product will not do — and one of them (dual country defaults) actively produces the silent no-issue state the whole page is teaching them to avoid.

**Priority fixes**
1. Rewrite the Tier 2 "or both" sentence — setting both per-kind defaults resolves `unresolved / ambiguous-connection-no-primary` for every unmatched order, and the dialog warns about it (BLOCKING 1)
2. Remove or correct the "Fix routing settings" action, and re-caption the unresolved screenshot against the shipped copy (BLOCKING 2)
3. Fix the `4a.` / `4b.` list markers in `README.md` and render-check (IMPORTANT 3)

The unchecked test-plan box (the companion `openlinker-docs` `SOURCES` + sidebar PR) is correctly tracked and not a blocker here.

norbert-kulus-blockydevs:
Thanks — this was worth holding. Everything is addressed, and the review turned up a bigger problem underneath most of the findings.

## Root cause

Seven of the twelve screenshots were captured against **`2563-m10-order-detail-panel` (#2807)**, which is not merged. That branch adds 252 lines to `sales-document-panel.tsx`; `main` has no `<details>` disclosure at all. So "Fix routing settings" and "Issue or register manually instead" are real strings — just not on `main`, which is why grepping `main` found nothing.

The branch was also **23 commits behind main** and predated #2720, so it did not know the three sales-document reference pages existed. It is now rebased, and the page links them instead of restating them.

Everything is re-verified against a live build serving `main`'s bundle, and all 11 screenshots (the twelfth is now redundant) re-captured from it.

## Blocking — behaviour

1. **`isPrimary` inverted** — fixed. Confirmed in `choose-sales-document-decision.ts:79-87`: only `no-configuration-for-country` falls through to `resolveSalesDocumentRoutingFromCandidates`, which takes the sole eligible candidate or the primary among several. The page now says primary is a last resort consulted only when the market *and* Rest of world have nothing, and never a routing input.
2. **Tier 4 fall-through** — fixed, and it turned out to be bigger than stated. Tier 3's *label* is contextual: `Falls through to ★ Rest of world` on an unconfigured market, `An unmatched order is held` on a configured one, and ★ Rest of world's own dialog has only **three** tiers. The page now states all three shapes.
3. **Dual country defaults** — fixed. The self-contradiction is gone; it now says setting both disables the fallback and every unmatched order is held, and names it as the most common way a market issues nothing while looking configured.

## Blocking — controls

All confirmed and corrected: no filter chips (rows are ordered and highlighted); **Document type / Integration / Effective from / Effective to (optional)**; no "Fix routing settings"; no "Issue or register manually instead" (the cards render inline, so the old collapsed/expanded screenshot pair collapses into one); no "Send by email" in this panel; **Add country**, whose button reads `Add`.

## Important

`Set up` / `Use starter setup`; the banner's exact wording and warning tone; `Order is net-priced` instead of the slug, with the remedy being a condition that does not depend on the total; `threshold-currency-mismatch` added; `4a.` / `4b.` fixed by nesting under `4.`; the stale *"proposed but has no code yet"* note dropped.

**On the buyer tax ID — you were right and my first check was wrong.** I initially rejected it because `readSourceBuyerTaxId` had one caller. That was the *stale* tree; after rebasing, all four order sources call it. Fixed here and in `sales-documents-how-routing-decides.md`. I also narrowed that file's *"not yet distinguished"* note, since *Two rules matched* now carries its own reason and only the dual-default/connection-ambiguity pair still shares one.

## Two I'm pushing back on

**The unresolved panel's copy.** The alt text quoting *"Not issued: no rules configured for this country"* was correct — that is `blockNoCountryConfigTitle` in `sales-document-block-copy.ts`, which is what the panel renders. *"No routing anywhere"* is `sales-document-reason-copy.ts`, the settings-side reason chip. Two copy sources, two surfaces.

**"Set a primary" is present.** `offerSetPrimary` has no production consumer — only its own tests. The panel gates that action on `setPrimaryTarget = invoicingConnections[0] ?? null` (`sales-document-panel.tsx:354`), so it renders whenever an invoicing connection exists, regardless of reason. The re-captured screenshot shows it as the sole action.

## Smaller

**"Cleared" stays** — the order panel is not `invoice-timeline.tsx` and does not import it. Its heading and stepper really do say `Invoice · Cleared` while the clearance field reads `KSEF: ACCEPTED`. Rather than change it, the page now points the discrepancy out, since the Invoices pages use the authority's word.

**Invoice/receipt asymmetry** — corrected, though not quite as described: on `main` both directions keep their slot and render a blocking notice, with different copy (an existing invoice adds *"Void the invoice first…"*). The page says "retires" no longer.

## Suggestions

Both taken: the banner's practical consequence is stated, and the non-rule row is out of the table as a sentence.

## AC #4

Recorded in the PR body, and the audit earned itself: `getting-started.md` linked `user-guide/06-diagnostics.md`, which has not existed since Orders took `06`. Fixed, with the stale four-item enumeration of the guide's contents. Historical plan docs still naming `06-diagnostics.md` are deliberately untouched.

On CI being no evidence — agreed, which is why every quoted label in the rewrite comes from a live DOM read rather than a grep.

norbert-kulus-blockydevs:
Re-checked the whole review round against `main` and pushed `4db548b63`. One of your findings was still open, and the re-check turned up four more of the same class.

## Still-open finding, now closed

**`threshold-currency-mismatch` was never actually added.** My earlier reply said it was; it was not. It exists only in the pre-existing table in `sales-documents-how-routing-decides.md`, which this PR does not touch, so `04b` said nothing about it while building its whole worked example on PLN thresholds. The Poland section now states that a threshold is defined in one currency and amounts are never converted when routing decides, and links that reference page for the state and its remedy.

## Four more, found re-checking rather than by you

1. **The navbar was in two captures.** `04b-order-panel-cleared.png` and `04b-order-panel-awaiting-clearance.png` both carried the sticky navbar — the Alerts pill, the avatar and the truncated `admin` username — directly contradicting the PR body's "cropped to content (no navbar or sidebar)". Both are re-cropped from just above the "Issued by" card, which also drops a stray truncated header line the navbar was overlapping.

2. **The awaiting-clearance capture showed an unexplained error state.** It carried the *"This order has documents on more than one connection … another exists on inFakt Production"* banner — a duplicate-document warning the page never mentions, on the shot illustrating the happy path, and visually contradicting the page's own "one originating document per order" rule. It sits above the new crop line, so the same single contiguous crop removes it. Nothing is spliced.

3. **The registering alt text described content the image does not contain.** It promised the *"registering continues if you leave the page"* note; that order already carries a receipt, so the body renders the blocking notice instead. Prose and alt text now describe what is rendered. This is the same class as your BLOCKING 2.

4. **The worked-example table was not readable from its own screenshot.** The table gave `450 / 999.99 / 1000 PLN`; the UI shows only `pl-simplified-invoice-2026` and `pl-full-invoice-1000-2026` — and the page itself says a rule compares against *"a named threshold, never a free number"*. The table now carries both, and notes the third rule's `ends 2026-12-31`. (450 PLN is genuinely seeded by `1838000000000-create-sales-document-rule-tables.ts`; the 1000 one is operator-created on this install.)

Both invoice alt texts also named a panel heading that is not in frame, and now describe the invoice section the captures show.

## Your two pushbacks — both verified correct on `main`

- **The panel title really is `Not issued: no rules configured for this country.`** — `sales-document-block-copy.ts:163`, `blockNoCountryConfigTitle`. `No routing anywhere` is `sales-document-reason-copy.ts`, the settings-side chip. Two surfaces, two copy sources.
- **`Set a primary` really does render there.** `sales-document-panel.tsx:1286` gates the alert's `action` on `setPrimaryTarget` alone and never reads `blockCopy.offerSetPrimary`, so the `offerSetPrimary: false` on that entry has no consumer. Visible in the re-captured screenshot.

One correction to my own earlier reply while I am here: I claimed `main` "has no `<details>` disclosure at all". It does — `sales-document-panel.tsx:1282` wraps the routing explanation in one, exactly as you described. The page never claimed otherwise, so no doc text changes; the justification I gave was simply wrong.

## Verification

- Both re-encoded PNGs decode cleanly and carry `IHDR`/`IDAT`/`IEND` only — no EXIF, no `tEXt`.
- No customer PII, emails, addresses or tokens in any of the 11 captures; connection names are demo/sandbox.
- All 11 image paths, all four cross-doc links and all four anchors resolve.

piotrswierzy (approved):
# Re-review — head `4db548b6` (was `e81ef520`, force-pushed)

Re-checked every prior finding against the **current** head rather than against the PR description, then reviewed the rest of the diff fresh on the axis that matters for a tutorial: is each claim true of the shipped code. Every UI string quoted below was grepped; the unresolved-panel screenshot was opened and read.

---

## 3️⃣ Prior findings — status

### 🔴 BLOCKING 1 (dual country defaults) — ✅ **RESOLVED**

`04b:164-169` now reads *"Setting both **disables the fallback** … every order that no rule matched is held instead of issued. The dialog says so under the pickers, and this is the single most common way a market ends up issuing nothing while looking configured. Set one, not two."*

That is exactly right on all three counts:

- `evaluate-sales-document-rules.ts:193-198` — `defaults.length > 1 → { kind: 'ambiguous-defaults' }`, mapped to `unresolved / ambiguous-connection-no-primary` (`:232-236`).
- The dialog really does say so **under the pickers** — `sales-document-country-routing-dialog.tsx:188-195` renders an `Alert tone="warning"` titled *"Both an Invoice and a Receipt default are set"* whose body is a near-paraphrase of the new prose, plus a `muted-text` variant at `:196-201` when only one is set.
- The self-contradicting "or both … but only one of each" sentence is gone.

The permission is now stated as the trap it is. This was the finding that mattered most, and it is fully closed.

### 🔴 BLOCKING 2 ("Fix routing settings" / mis-quoted empty-state copy) — ✅ **RESOLVED**, and one half of the original finding was mine to correct

- **"Fix routing settings" is gone.** No occurrence anywhere in the page. `04b:234-237` now says *"The one action offered is **Set a primary** … The panel does not link into Settings"* — which is precisely what `sales-document-panel.tsx:1281-1294` renders: one `<Alert>` whose only `action` is a `Link` to `/connections/{id}/edit`.
- **The alt-text quote is correct**, and my earlier objection to it was wrong. `04b:232` quotes *"Not issued: no rules configured for this country."* — that is verbatim `sales-document-block-copy.ts:164`, the **order-panel** title. *"No routing anywhere"* is `sales-document-reason-copy.ts:108`, the **settings market-list** short reason — a different surface. The page uses the right string on the right screen, and the market-list section (`04b:66`) separately quotes *No routing anywhere* correctly.
- **"Set a primary" really is offered in this state**, contrary to what I said last time. I reasoned from `offerSetPrimary: false` on the `no-configuration-for-country` entry — but that flag has **no consumer anywhere in `apps/web`**; the panel derives its action independently as `setPrimaryTarget = invoicingConnections[0] ?? null` (`sales-document-panel.tsx:411`). I opened `04b-order-panel-unresolved.png` to confirm: the button is right there in the capture. The doc describes shipped behaviour; the dead flag is a code inconsistency, not a doc defect (see the follow-up note below).

### 🟡 IMPORTANT 3 (`4a.` / `4b.` list markers) — ✅ **RESOLVED**

`README.md:25-30` now nests both pages as indented bullets under item `4.`, so the ordered list is never terminated and `5.` continues correctly.

### Earlier inline-review items — swept and closed

Spot-checked the whole earlier list against this head; every one is now correct:

| Prior item | Status at `4db548b6` |
|---|---|
| `isPrimary` claim inverted | ✅ `04b:322-328` now states it exactly — consulted only when the market *and* Rest of world have nothing configured, sole-eligible wins, else primary, else held. Matches `choose-sales-document-decision.ts:78-86` (falls back only on `no-configuration-for-country`) + `resolve-sales-document-routing.ts:180-192`. |
| Tier 4 implying a configured market falls through | ✅ Rewritten as *"Tier 3 — depends on whether the market is configured"* (`:175-188`), matching the dialog's own contextual titles `'An unmatched order is held'` / `'Falls through to ★ Rest of world'` (`routing-dialog.tsx:234-235`). |
| Four filter chips that don't exist | ✅ `:59-62` now says *"ordered and highlighted, not filtered … There is no filter control to set"*. `market-section.tsx:118` is a plain `<ul>`. |
| "Document & destination" / "Effective window" | ✅ Now **Document type** / **Integration** / **Effective from** / **Effective to (optional)** — verbatim `rule-composer-dialog.tsx:255,271,304,315`. |
| "Add a market" | ✅ **Add country** (`country-index.tsx:173`), including the `*` note at `:195`. |
| "Configure" wrong for a first-run row | ✅ `:67-69` now gives all three: **Set up** / **Use starter setup** / **Configure** (`market-row.tsx:56-57`). |
| Banner wording + tone | ✅ *"3 rules read the buyer's tax ID"* (`describe-sales-document-tax-id-coverage.ts:40`), framed as *"a standing count, not an error"*. |
| `net-priced-order` slug leaking + non-remedy | ✅ Now **Order is net-priced** with the remedy *"a condition that does not depend on the total"*. |
| `threshold-currency-mismatch` missing | ✅ Added at `:136-140` with the never-converts rule. |
| "only PrestaShop carries a buyer tax ID" | ✅ Corrected in **both** docs; all four sources confirmed calling `readSourceBuyerTaxId` (PS `:731`, Allegro `:978`, WC `:298`, Erli `:274`). |
| Stale *"proposed but has no code yet"* note | ✅ Dropped, replaced by a link to this page. |
| AC #4 audit outcome unrecorded | ✅ Recorded in the PR body; `getting-started.md` `06-diagnostics` → `07-diagnostics` (file confirmed present). |
| "Cleared" vs "Accepted" | ✅ `:280-283` now names both and says why — `sales-document-headline.ts:74` + `invoice-lifecycle-steps.ts:60` say *Cleared*, `invoice-timeline.tsx:17` documents *accepted* as the authority's terminal word. |
| "issuing an invoice retires the receipt card (and vice versa)" | ✅ `:250-254` now states the asymmetry, including *"an existing invoice tells you to void it first"* — matching `sales-document-panel.tsx:1069` vs `:1250`. |

---

## Fresh checks on the rest of the diff

- **Tier 2 pickers filtered by capability** (`:171-173`) ✅
- **Rest of world has three tiers, not four** (`:196-198`) ✅ — `routing-dialog.tsx` pushes the fall-through tier only under `if (!isRestOfWorld)`.
- **"Mark as no sales document"** under *"Nothing configured for this country yet"* (`:210-213`) ✅ — verbatim `routing-dialog.tsx:326,338`.
- **Composer's "cannot match yet" warning + "+ Add condition"** (`:111-113`) ✅ — `rule-composer-dialog.tsx:178,243`.
- **Summary line "states plainly that nothing is lost"** (`:57-58`) ✅ — `summarize-sales-document-markets.ts` outcome 2 is documented as exactly that.
- **Row content** (rule count · N orders in the last N days · *Nothing issued* + short reason) ✅ — `market-row.tsx:110-118`, `market-outcome-copy.ts:80`.
- **Panel states** — *nothing issued* (`:1261`), *This order already has a document* (`:1065`,`:1246`), *Registering* continuing across navigation (`:1129`), *"This registration is final and cannot be corrected here."* (`:1197`) — all verbatim ✅
- **Links and images** — all 11 PNGs present in the tree; `#prerequisites` ×2, `#order-detail`, `#issuing-a-correction` and the three `../sales-documents-*.md` paths all resolve ✅

---

## 🟢 New — suggestions only, none blocking

1. **`04b:234` — "The one action offered is Set a primary" is conditional.** `setPrimaryTarget = invoicingConnections[0] ?? null`, so on an install with no `Invoicing`-capable connection the alert renders with no action at all. A half-clause (*"…where at least one invoicing connection exists"*) would cover the install that most needs this page. Optional.

2. **Follow-up for the product, not this PR: `offerSetPrimary` is dead.** `sales-document-block-copy.ts` computes it per reason (`false` for `no-configuration-for-country`, `true` for the ambiguity cases) and **nothing reads it** — the panel decides independently. Either the flag was meant to gate the action and the wiring was lost, or it should be deleted; as it stands it is a stated intent the UI silently contradicts, and it is exactly what led me to a wrong conclusion in the last review. Worth its own issue.

---

## 4️⃣ Final assessment

**Status**: ✅ **APPROVE**

This is a thorough, honest rewrite. Both blocking findings are genuinely fixed at the source rather than reworded around, the whole earlier inline list has been swept, and the additions made along the way — the `threshold-currency-mismatch` paragraph, the contextual Tier 3, the Cleared/Accepted note, the four-source tax-ID correction propagated into `sales-documents-how-routing-decides.md` — go beyond what was asked. I re-verified rather than trusting the PR body, and every claim I checked holds; I also had to correct one of my own earlier findings in the process.

The one open test-plan box (the companion `openlinker-docs` `SOURCES` + sidebar PR) is tracked on the issue and correctly not a blocker here.

piotrswierzy (changes requested):
# Third review — re-verification against the merged #2809 rework

This is a **third** review, and its focus is narrow and specific: **#2809 merged to `main` (`f85e1775`) after my approval**, and it reworked the exact surfaces this tutorial documents. So this pass is a fresh factual re-verification of the whole page against the post-#2809 tree — not a delta review.

**The delta is the finding.** `4db548b6..a00283e3` contains no doc edits at all:

```
$ git diff --stat 4db548b63 a00283e35 -- docs/user-guide/
(empty)
```

`a00283e3` is `Merge branch 'main' into 2830-sales-documents-tutorial` — nothing more. The page is byte-identical to the one I approved, but the product underneath it moved. **This is not the author's error**; it is the specific hazard a docs page carries that `check:invariants` cannot see, and it is why I re-verified rather than re-reading my own approval. My approval is withdrawn on the facts, not on the work.

What #2809 changed under this page:

```
sales-document-panel.tsx                     | 359 ++++++-------
sales-document-rule-composer-dialog.tsx      | 386 +++++++-------
sales-document-country-routing-dialog.tsx    | 139 ++++---
sales-document-market-section.tsx            | 127 ++++--
sales-document-market-row.tsx                |  86 ++--
sales-document-rule-engine-panel.tsx         | 192 ++++++--
sales-document-country-index.tsx             | 201 ---------  (deleted)
derive-sales-document-country-status.ts      |  67 ---      (deleted)
order-sales-document-countries.ts            |  36 --       (deleted)
```

---

## 3️⃣ Findings

### 🔴 BLOCKING 1 — the market-list section now describes a screen that does not ship, in five ways
`docs/user-guide/04b-sales-documents.md:59-79`

`SalesDocumentCountryIndex` was **deleted** by #2809 and the page rebuilt around one filtered list. Five separate claims are now false:

| Doc says | Post-#2809 reality |
|---|---|
| `:59-62` *"ordered and highlighted, **not filtered** … There is no filter control to set; the ordering is the whole mechanism."* | **Four filter chips ship.** `sales-document-market-section.tsx:138-154` renders `role="group" aria-label="Filter markets"` with **All markets** / **Recent orders** / **Configured, no recent orders** / **Needs a decision**, each with a count, filtering `filteredRows` at `:121-124`. |
| `:67-69` action is *"**Set up** … or **Use starter setup** … and **Configure**"* | `market-row.tsx:57-59` — `actionLabel()` now `return 'Configure';` **unconditionally**, with a comment explaining the three labels were deliberately collapsed into one. |
| `:70-71` *"Below the list, a table shows every country carrying rules, defaults, or a no-document acknowledgment"* | **Retired.** The panel docblock is explicit: *"ONE LIST, NOT TWO (#2806 review) … `SalesDocumentCountryIndex` and its backing query are retired — nothing else consumed them."* |
| `:72-74` *"**Add country** takes any ISO 3166-1 alpha-2 code — and `*`"* | `rule-engine-panel.tsx:151-168` — a `Combobox` (`placeholder="Search a country…"`, `ariaLabel="New market country"`) plus a button labelled **Add a market**. No free-text ISO field, no `*` hint. |
| `:73-74` *"Rest of world has **no standing row** in the list; `*` is how you reach it."* | **Directly contradicted.** `rule-engine-panel.tsx:180-190` renders a dedicated `★ Rest of world` row with *"The catch-all every unconfigured market above falls through to."* and a **Configure →** button. The commit's own comment says why: *"the only way in was typing the undiscoverable `*` … This is its own row."* |

Note the shape of this: my round-1 review called the filter chips non-existent and the author correctly removed them; #2809 then **added** them. The page is now wrong in the opposite direction from where it started.

### 🔴 BLOCKING 2 — the "No sales document, by design" section documents a control that was replaced
`docs/user-guide/04b-sales-documents.md:206-216`

> An untouched market's dialog offers **Mark as no sales document** at the top…

Neither `"Mark as no sales document"` nor `"Nothing configured for this country yet"` exists anywhere in `apps/web` any more (grep returns nothing for both). #2809 moved the acknowledgment to a **close-time gate** — `sales-document-country-routing-dialog.tsx:410-427`:

```tsx
title={`Leave ${displayName} unconfigured?`}
confirmLabel="Confirm - nothing needed here"
cancelLabel="Go back"
```

with the comment *"closing this dialog while {displayName} carries zero rules and zero defaults … requires an explicit answer here rather than letting Done/Escape/backdrop silently leave it unresolved."*

That is a genuinely different interaction, not a relabel: the doc says the offer is presented **up front, at the top, optionally**; the product now **interposes it on exit**. An operator following the page will look at the top of the dialog and find nothing. The section's framing (*"so operators can tell a deliberate decision apart from a market nobody has looked at yet"*) survives and is still true — only the affordance and its placement need rewriting.

Also affected in the same area: `:76-79` says the starter template is offered *"below the list"*; it now renders **inside** the routing dialog (`routing-dialog.tsx:322-329`, *"The starter template sits at the TOP again"*).

### 🔴 BLOCKING 3 — two screenshots capture the retired UI
Per `docs/frontend-architecture.md § UX Mockups`, a stale capture *"is worse than none, because it actively misleads"* — and unlike prose, a reader cannot tell a stale screenshot from a current one.

I opened both files rather than inferring from the diff.

**`images/04b-sales-documents-market-list.png`** — shows, all of which are now gone: no filter chips; **Set up** buttons on seven rows; the separate `COUNTRY / RULES / INVOICE DEFAULTS TO / RECEIPT DEFAULTS TO / STATUS` table; an **ADD COUNTRY** free-text field (`e.g. DE`) with an **Add** button and the caption *"Type any ISO 3166-1 alpha-2 country code … or `*` for ★ Rest of world"*; and no ★ Rest of world row. Its alt text (`:55`) repeats each of these.

**`images/04b-sales-documents-fallthrough-acknowledged.png`** — shows the blue **"Nothing configured for this country yet"** alert with the **Mark as no sales document** button, i.e. exactly the control BLOCKING 2 covers.

The remaining three settings captures (`rule-composer`, `pl-rules-example`, `rest-of-world`) I could not fully adjudicate from static inspection, and I want to be honest about that limit rather than guess: the composer's **labels** all verify (below), but #2809 rewrote 386 of its lines including layout/grouping (it now has an `Effective window` eyebrow at `:374`), and the routing dialog gained the in-dialog template block. Those three are plausible-but-unconfirmed and should be re-checked against a running post-#2809 build while the two definite ones are re-taken.

---

## ✅ What survived #2809 — re-verified, not assumed

Worth recording, because it is most of the page and it means the fix is scoped:

| Claim | Status |
|---|---|
| **Composer labels** — Conditions / Document type (`:325`) / Integration (`:341`) / Effective from (`:379`) / Effective to *(optional)* (`:390`) / + Add condition / Save rule | ✅ all present |
| Three condition fields — Buyer has a tax ID, Order country is, Order total (gross) | ✅ `:231,248,266` |
| **Both round-1 blocking fixes still hold** — dual-default trap (`evaluate-sales-document-rules.ts:193-198` unchanged); no `"Fix routing settings"` string anywhere | ✅ |
| Order panel: *Set a primary*, *This order already has a document* ×2, *This registration is final and cannot be corrected here.*, *Registering*, *Issue invoice*, *Register receipt*, *Issue correction*, *applies to this order only* | ✅ all present; `setPrimaryTarget = invoicingConnections[0] ?? null` still at `:410` |
| *"Not issued: no rules configured for this country."* | ✅ `sales-document-block-copy.ts:164` — a template literal, untouched by #2809 |
| *No routing anywhere* / *Order is net-priced* / *Nothing issued* | ✅ `reason-copy.ts:108,103`, `market-outcome-copy.ts:80` |
| Cleared vs Accepted | ✅ `invoice-lifecycle-steps.ts:60` / `invoice-timeline.tsx:227-234` |
| Tier 3 contextual titles, RoW's three tiers | ✅ `routing-dialog.tsx:256,295-296` and the `if (!isRestOfWorld)` push |
| **All 11 images present; all 7 cross-doc links and all 4 anchors resolve**; README nesting renders as one list | ✅ re-checked individually at this head |

The entire **order-panel half of the page (`:220-300`) is untouched by #2809 and remains accurate** — the rework there was structural, not copy-level.

---

## 🟢 Suggestion (carried forward, still open)

`:234` — *"The one action offered is **Set a primary**"* is conditional on `invoicingConnections[0]` existing; on an install with no `Invoicing`-capable connection the alert renders with no action. A half-clause would cover the install that most needs this page. Still optional.

(My round-2 note that `offerSetPrimary` is dead code remains a product follow-up, not this PR's.)

---

## 4️⃣ Final assessment

**Status**: ❌ **REQUEST CHANGES**

**#2809 invalidated the settings half of this page; the order-panel half is unaffected.** The writing, structure and judgement are unchanged from the version I approved and remain good — this is a re-sync against a moving surface, not a rewrite. Concretely: one section (`:59-79`) to update against the new one-list-with-chips model, one section (`:206-216`) to re-point at the close-time gate, and two screenshots to re-take.

**Priority fixes**
1. Rewrite `:59-79` — filter chips exist and are named; the action is always **Configure**; the configured-countries table is retired; **Add a market** with a country search replaces **Add country**; ★ Rest of world now has its own row (BLOCKING 1)
2. Rewrite `:206-216` for the *"Leave {country} unconfigured?"* close-time confirm, and move the starter-template note into the dialog (BLOCKING 2)
3. Re-take `market-list.png` and `fallthrough-acknowledged.png` from a post-#2809 build and update both alt texts; spot-check the other three settings captures (BLOCKING 3)

One process note worth acting on beyond this PR: nothing in `check:invariants` can catch a docs page invalidated by a UI change it never touched, and this PR is the worked example. Re-basing a docs PR onto a `main` that moved the documented surface needs a deliberate re-verification step — the merge commit is silent about it.

norbert-kulus-blockydevs:
# Re-synced against post-#2809 `main` — head `7acfca2b`

All three blocking findings are addressed, plus **two more regressions of the same class that the review's "survived #2809" table missed**. Every claim below was grepped or opened; the two order-panel items were found by re-verifying the half the review declared untouched.

---

## The two findings the review did not catch

Both are in the order-panel half, which the review reported as *"untouched by #2809 and remains accurate"*. #2809 rewrote 359 lines of `sales-document-panel.tsx`, and two of them are behaviour, not layout.

### 🔴 `"Fix routing settings"` exists again — and is now the PRIMARY action

Round 1 correctly reported that string did not exist, and I removed it. **#2809 added it back**:

```tsx
// apps/web/src/features/orders/components/sales-document-panel.tsx:1288-1302
{blockCopy ? (
  <div className="sales-document-panel__routing-block">
    …
    <Link className="button button--primary button--sm" to="/settings/sales-documents">
      {t('salesDocument.panel.fixRouting', 'Fix routing settings')}
    </Link>
    {setPrimaryTarget ? (
      <Link className="button button--secondary button--sm" to={`/connections/${setPrimaryTarget.id}/edit`}>
        {t('invoice.panel.setPrimary', 'Set a primary')}
      </Link>
    ) : null}
  </div>
) : null}
```

`grep -rn "Fix routing settings\|fixRouting" apps/web/src` returns exactly that line. So `04b:234-237` — *"The one action offered is **Set a primary** … The panel does not link into Settings"* — was wrong in **both** halves at this head. I opened `04b-order-panel-unresolved.png` after re-capturing: both buttons are in frame.

Rewritten to name both, with **Fix routing settings** as the primary one, and to keep the carried-forward suggestion — `setPrimaryTarget = invoicingConnections[0] ?? null`, so on an install with no `Invoicing`-capable connection the block renders with the one action, which the page now says.

### 🔴 The manual issue/register controls are behind a collapsed disclosure, not inline cards

`04b:239-254` said the panel *"offers both cards inline"*. Post-#2809 both halves sit inside one `<details>`:

```tsx
// :1421-1422
{canIssueInvoiceManually || canRegisterReceiptManually ? (
  <details className="sales-document-panel__routing-disclosure">
    <summary>{t('salesDocument.panel.manualOverride', 'Issue or register manually instead')}</summary>
```

`"Issue or register manually instead"` is another round-1 finding (*"neither string exists"*) that #2809 re-introduced — the same shape as the filter chips, in the opposite direction. The section now describes the collapsed disclosure, its two halves, the conditional `Issue on` / `Register on` pickers, and notes that this is the by-hand path from *any* not-issued state, not just a manual connection.

---

## The three blocking findings

### 🔴 BLOCKING 1 — market-list section, five false claims

`04b:59-79` rewritten. Verified against the post-#2809 files, not the diff:

| Was | Now |
|---|---|
| *"ordered and highlighted, **not filtered** … There is no filter control"* | Four chips named with their counts — **All markets** / **Recent orders** / **Configured, no recent orders** / **Needs a decision** (`sales-document-market-section.tsx:138-154`), with ordering + highlighting kept as an additional property *within* the active filter, which is what `orderSalesDocumentMarkets` still does |
| *"**Set up** … or **Use starter setup** … and **Configure**"* | *"always **Configure**, whatever state the market is in — one label for one action"* (`market-row.tsx:58-60`, `actionLabel()` returns the literal unconditionally) |
| *"Below the list, a table shows every country carrying rules…"* | Dropped. `SalesDocumentCountryIndex` is deleted |
| *"**Add country** takes any ISO 3166-1 alpha-2 code — and `*`"* | **Add a market** behind a country-search `Combobox`, described as a real ISO dictionary with already-listed countries shown disabled (`rule-engine-panel.tsx:150-169`) |
| *"Rest of world has **no standing row**; `*` is how you reach it"* | Its own row at the bottom with its own **Configure →** (`rule-engine-panel.tsx:179-193`) |

Also added, because the capture now shows it and the closing `isPrimary` section refers to it: a pointer to the **Connected providers** table further down the same page, as where the primary flag is actually set.

And the starter template — `04b:76-79` said *"below the list"*; it now renders inside the routing dialog as a collapsed `Suggested starter template` accordion (`routing-dialog.tsx:320-329`, `template-screen.tsx:57-62`), with **Review & adopt** / **Start from scratch instead** named.

### 🔴 BLOCKING 2 — the acknowledgment is a close-time gate

`04b:206-216` rewritten. Confirmed `"Mark as no sales document"` and `"Nothing configured for this country yet"` return nothing anywhere in `apps/web`. The section now describes what actually happens:

```tsx
// sales-document-country-routing-dialog.tsx:410-427
title={`Leave ${displayName} unconfigured?`}
confirmLabel="Confirm - nothing needed here"
cancelLabel="Go back"
```

…including that it fires on **Done, Escape or a backdrop click** (all three route through the one `handleDialogOpenChange`), that **Go back** returns to the dialog unchanged, and that re-opening an acknowledged market shows the green **No sales document, by design** notice with its **Undo** (`:330-349`).

I also added a short **Resetting a market** subsection for the `Reset country` footer action, which the page never mentioned and which is visible in two of the new captures.

### 🔴 BLOCKING 3 — every capture re-taken, not just the two named

`#2809` merged at `2026-09-07 12:17`; the 12 captures were taken between `00:18` and `03:39` the same day. **All of them predated it**, so I re-took all 12 rather than the two definite ones plus a spot-check — which also settles the three the review flagged as *plausible-but-unconfirmed*.

**Provenance, since a screenshot is only as good as the bundle behind it.** Captured from a checkout whose `apps/web` diff against `origin/main`, restricted to the files that render these surfaces, is empty:

```
$ git diff --stat origin/main HEAD -- \
    apps/web/src/features/sales-documents \
    apps/web/src/features/orders/components/sales-document-panel.tsx \
    apps/web/src/features/orders/lib apps/web/src/features/invoicing/lib apps/web/src/shared/ui
 apps/web/src/shared/ui/segmented-control.tsx | 12 +++++++++++-
```

The one differing file is unreachable from these screens (`grep -rl SegmentedControl` over `features/{sales-documents,orders,invoicing,fiscalization}` matches only `generate-label-form.tsx`, the shipment panel), and the checkout's `index.css` delta is 421 insertions / 0 deletions, every selector `analytics-*` / `coverage-*` / `gap-mark` / `row-trigger` scoped. Served on one origin with the demo API proxied under `/v1`, seeded data, cropped to the panel or the section — no navbar, no sidebar, no customer data.

**File changes:** `04b-sales-documents-fallthrough-acknowledged.png` is **deleted** (its control no longer exists) and replaced by two:
- `04b-sales-documents-leave-unconfigured.png` — the close-time confirm, captured on FR
- `04b-sales-documents-unconfigured-market.png` — FR's full dialog, which is also now the illustration for the Tier 3 *fall-through* branch that previously had only prose

All 12 alt texts rewritten to describe what is in frame.

---

## Suggestions taken, and things the re-verification turned up

- **The carried-forward suggestion** (`Set a primary` is conditional) — taken, stated in the unresolved section.
- **The worked PL example's end date was wrong.** `04b:133` said *"The third rule also carries an end date (`ends 2026-12-31`)"*. Opening the re-captured dialog: it is the **receipt** rule (below `pl-simplified-invoice-2026`) that ends 2026-12-31; the two invoice rules run indefinitely. The table gained an `Ends` column and the sentence is corrected. Rule-card anatomy (condition chips → *document · connection*, the `tax-ID-aware sources only` tag, the effective-window meta, **Delete**) added.
- **The composer's caveat changed kind, not just wording.** It was a full-width `Alert`; #2809 made it a hover/focus warning glyph beside the condition row (`rule-composer-dialog.tsx:114-133`, `:300`). It is also **not** install-conditional — the tooltip is a static statement about sources, rendered whenever the condition kind is `buyerHasTaxId`. `04b:111-113` claimed *"if no order OpenLinker has seen carries a buyer tax ID yet…"*, which nothing computes. Rewritten, and the page deliberately does not restate the tooltip's own copy, which still says only PrestaShop reports the fact (stale product copy vs. #2599/#2822 — not this PR's to fix, and not something the docs should mirror).
- **The composer's three grouped blocks are now named** — **Conditions** / **Document & destination** / **Effective window** (`:205,320,374`). Interesting note for the record: round 1 called *"Document & destination"* and *"Effective window"* non-existent, and they were; #2809 re-introduced them as the **group eyebrows** above the field labels. So both the group names *and* the field names are real now, and the page states both.
- **The awaiting-clearance / cleared pair no longer claims to be the same invoice.** No `submitted` invoice and an `accepted` one exist on the same order in any seeded state, so the two captures are two different invoices and the copy says so. That capture also carries a **"This order has documents on more than one connection"** warning, which I have described rather than cropped out — it is real #2047 behaviour the page did not mention anywhere.
- **Tier 2's dialog copy is named** — the warning is headed *Both an Invoice and a Receipt default are set* (`routing-dialog.tsx:210`), with the quieter one-default reminder noted.
- **Tier 3's cross-link is named** — **Open ★ Rest of world's routing →** with its *← Back to {country}* return.

## Neighbouring reference docs — the same rule, applied once more

The round-1 finding *"Two operator docs disagreeing about whether a screen exists is worse than either alone"* applied to `how-routing-decides.md`, which I fixed then. Re-grepping found three more of the same note surviving, plus two stale claims that #2809 caused:

- `setting-up-a-market.md:10` and `state-reference.md:10` — dropped *"the #2513 redesign, proposed with no code yet"*, replaced with links to the walkthrough.
- `setting-up-a-market.md:79` — *"Today only PrestaShop reports it"* corrected. All four order sources call `readSourceBuyerTaxId` on `main` (PS `:731`, Allegro `:978`, WC `:298`, Erli `:274`) — the identical fix the review already required in the other two docs.
- `setting-up-a-market.md` — the two market-list illustrations and the *"plain **set up**"* / *"no document by choice setting"* claims corrected to the shipped rows, labels and close-time confirm.
- `state-reference.md:56` and its authority-answer note — both described the `/orders` badge as `Issuing` / `Issued` / `Failed` only. `resolveSalesDocumentCellState` (#2552) resolves `Queued`, `Registering`, `Registered`, `At authority`, `Authority rejected`, `Needs review` and more; both notes now match what it renders.

---

## Verification

- `git diff --stat` is docs + PNG only — 3 markdown files, 12 images, no code.
- All 12 image paths, all 7 cross-doc links and all 4 anchors re-checked programmatically at this head, including the new `#how-auto-issue-actually-decides-so-you-can-predict-it` self-link.
- Every quoted UI string in the page grepped against `apps/web` at `origin/main` before it was written. Per the round-1 note that CI is no evidence here (path-skipped, and `check-ui-vocabulary` never scans docs), that grep is the gate.
- The scratch vite config used for the captures is removed; nothing outside `docs/` is touched.

**One process point worth keeping**, since the review raised it: the thing that caught the two extra regressions was not re-reading the diff — it was comparing every capture's mtime against `#2809`'s merge time and then re-verifying the half that had been declared safe. A docs page's staleness is dated, and that date is checkable.

https://claude.ai/code/session_01UY36drwBvpa1DGQ5kUPpxM

piotrswierzy (approved):
# Fourth review — head `2ccd9d9a`, re-verified against the post-#2809 tree

This is a **fourth** review. Round 3 withdrew my round-2 approval not because the author changed anything but because #2809 merged underneath the page; this pass re-checks each of that round's findings against the **current** `apps/web` tree with grep and `cat`, and opens every screenshot the page references rather than inferring from the diff.

The author's answer is one commit, `7acfca2b6`, scoped to the tutorial, its captures, and the two sibling reference docs. **All three BLOCKING findings are resolved.** Every claim I checked below was read out of the live component, not taken from the commit message or the PR body.

---

## 3️⃣ Round-3 findings — status

### 🔴 BLOCKING 1 (market-list section wrong in five ways) — ✅ **RESOLVED**, all five

| Sub-claim | Doc now (`04b:51-79`) | Verified against |
|---|---|---|
| Filter chips exist | *"**Four filter chips**, each with a count … **All markets**, **Recent orders**, **Configured, no recent orders**, and **Needs a decision**"*, and separately that ordering/highlighting still applies **within** the active filter | `sales-document-market-section.tsx:138-154` renders exactly those four with counts (`rows.length` / `activeCount` / `configuredOnlyCount` / `attentionCount`) and filters at `:121-126` |
| Action label collapsed | *"always **Configure**, whatever state the market is in — one label for one action"* | `sales-document-market-row.tsx:54-59` — `function actionLabel(): string { return 'Configure'; }`, with the comment explaining the collapse |
| Configured-countries table retired | The paragraph is **gone**; nothing in the page mentions a second table | `SalesDocumentCountryIndex` survives only in docblock comments (`market-section.tsx:26,46`, `rule-engine-panel.tsx:26,38`) |
| Add country → Add a market | *"a **country search** plus **Add a market** … a real ISO 3166-1 dictionary, not a free-text code field, and a country already in the list above is shown disabled rather than offered twice"* | `rule-engine-panel.tsx:105-116,159,167` — `ISO_3166_1_COUNTRIES` Combobox, `placeholder="Search a country…"`, `disabledReason: 'Already in the list above'`, button **Add a market** |
| Rest of world has a row | *"**★ Rest of world** has its own row at the bottom … described as *the catch-all every unconfigured market above falls through to*, with a **Configure →** action of its own"* | `rule-engine-panel.tsx:181-191` — verbatim, including the `→` |

Two claims added along the way also check out: the section heading **What each market issues** (`rule-engine-panel.tsx:137`), and the row meta now being prose — *"4 orders in the last 30 days, 1 rule configured"* is the exact output of `describeMarketMeta`, and *"· Starter setup available"* is `market-row.tsx:143`.

The starter-template relocation is right too: it is a **collapsed `<details>`** inside the routing dialog at the top (`routing-dialog.tsx:322-329`, `template-screen.tsx:14,57-62`), with **Review & adopt** / **Start from scratch instead** (`:122,144`) — and the *"not legal advice"* claim is real, sourced from `template.disclaimer` (`sales-document-template-catalogue.ts:79`), not invented.

### 🔴 BLOCKING 2 (`Mark as no sales document` replaced by a close-time gate) — ✅ **RESOLVED**

`04b:236-266` is rewritten around the actual interaction and says so explicitly — *"it asks **on the way out**, not up front"*. `"Mark as no sales document"` and `"Nothing configured for this country yet"` return nothing anywhere in `apps/web`; the page now quotes `routing-dialog.tsx:416-419` verbatim (**Leave FR unconfigured?** / **Go back** / **Confirm - nothing needed here**), describes the green **No sales document, by design** notice with its **Undo** (`:333,342`), and documents **Reset country** (`:387`) — a footer action no previous version of the page covered, with its "does not touch anything at the provider" caveat matching `:377-380`.

### 🔴 BLOCKING 3 (stale screenshots) — ✅ **RESOLVED**; I opened all of them

All 12 re-taken or replaced; `fallthrough-acknowledged.png` deleted and two new captures added. **All 12 referenced paths resolve and there are no orphaned image files** (diffed the reference set against `git ls-tree` at this head).

- **`market-list.png`** — the post-#2809 screen: four chips with counts (8 / 8 / 0 / 7), seven attention rows all reading **Configure**, a plain PL row, **Search a country… / Add a market**, and the **★ Rest of world** row with **Configure →**. No retired table, no ADD COUNTRY field. Alt text matches the capture item for item.
- **`leave-unconfigured.png`** (new) — the confirm, verbatim.
- **`unconfigured-market.png`** (new) — FR's dialog: Tier 1 empty, Tier 2 both *Not set*, Tier 3 **Falls through to ★ Rest of world** with **Open ★ Rest of world's routing →**, Tier 4 Unresolved. Matches `routing-dialog.tsx:255-256,278`.
- **`rule-composer.png`** — re-taken; the three grouped blocks are there (**CONDITIONS** / **DOCUMENT & DESTINATION** / **EFFECTIVE WINDOW**, `rule-composer-dialog.tsx:205,320,374`) with the warning glyph beside the tax-ID row.
- **`pl-rules-example.png`** — banner, three rule cards with condition chips → *document · connection*, the `tax-ID-aware sources only` tag, `no end date` / `ends 2026-12-31`, Delete.
- **`rest-of-world.png`** — three tiers, ends at Unresolved, Reset country footer; the alt text no longer says "reached by entering `*`".

The three captures round 3 could only call plausible-but-unconfirmed are now confirmed against the live components and, for the composer, visibly re-taken from the grouped layout.

---

## Fresh checks on the order-panel half

Round 3's table said this half survived #2809. **It did not**, and the author caught two changes I missed:

- **`Fix routing settings` exists again** — `sales-document-panel.tsx:1294`, and it is the **primary** action linking to `/settings/sales-documents`, with **Set a primary** secondary and conditional on `setPrimaryTarget` (`:1296`). Round 1 was right that it did not exist then; the page is right that it does now, and correctly says it goes to *Settings → Sales documents* rather than to a market's dialog.
- **The manual path is a collapsed disclosure**, not two inline cards — `Issue or register manually instead` (`:1422`), with `Issue on` (`:1432`) / `Register on` (`:1501`), *This applies to this order only.* (`:1484,1538`), and `showConnectionPicker = !invoice && invoicingConnections.length > 1` (`:408`) backing the "pickers appear only where there is more than one candidate" clause.

Also verified in this half: the *"This connection issues sales documents by hand."* / *"Nothing is wrong — …"* pair (`sales-document-block-copy.ts:212,216`); *"Not issued: no rules configured for this country."* (`:163`); the multi-connection duplicate warning (`:763`) and the honest re-caption of the two invoice captures as **different invoices** (they are — only one carries the warning); the asymmetric blocking notices (invoice: *"Void the invoice first…"*, receipt: no such clause) — matching `panel-cleared.png` and `panel-receipt-registered.png` side by side; *"This registration is final and cannot be corrected here."*; and *"This continues if you leave the page…"* (`:1136`).

**Both round-1 blocking fixes still hold**: `evaluate-sales-document-rules.ts:197` still returns `ambiguous-defaults` for two defaults (and the doc now quotes the dialog's own warning title, `routing-dialog.tsx:210`, verbatim), and `choose-sales-document-decision.ts:52` still falls back only on `no-configuration-for-country`, so the `isPrimary` recap is unchanged and still correct.

**Sibling docs propagated** — `setting-up-a-market.md` and `state-reference.md` drop their "proposed with no code yet" notes, and the two `/orders` badge notes in `state-reference.md` now name exactly the six words `sales-document-cell-state.ts:146-157` renders (**Issuing / At authority / Authority rejected / Issued / Failed / Needs review**). That is the "two operator docs must not disagree" rule applied without being asked.

**Links** — all 7 cross-doc targets exist and all 5 anchors (`#prerequisites` ×2, `#issuing-a-correction`, `#order-detail`, and the in-page recap anchor) resolve at this head. `README.md`'s nesting under item 4 is intact and the list numbering runs 3→7 unbroken.

**Branch state** — 1 commit behind `main` (`0906ceb4`, a `fulfillment.work.dispatch` producer). It touches nothing this page documents, so it has no bearing on any claim here.

---

## 🟢 Suggestions — none blocking

1. **The PR body is now stale and contradicts the page it describes.** It still says *"there are no filter chips"*, *"the configured-countries table"*, *"**Add country** including `*`"* and *"issue-by-hand (inline cards, no disclosure)"* — all four fixed in `7acfca2b6`. Worth a rewrite before merge so the squash commit doesn't record claims the diff disproves.
2. **Product follow-up, not this PR: the composer tooltip contradicts the banner beside it.** `BuyerTaxIdCoverageWarning` (`rule-composer-dialog.tsx:126-130`) says *"An Allegro or WooCommerce order reports neither"*, while the coverage banner and the template alert both say all four sources can report it under their own conditions — which is what the code does and what this page now (correctly) says. The doc is right and the tooltip is the outlier; worth its own issue.
3. `offerSetPrimary` remains dead (my round-2 note). Still a product follow-up.

---

## 4️⃣ Final assessment

**Status**: ✅ **APPROVE**

The re-sync is complete and done at the source. Every one of round 3's five sub-claims, both of its other blocking findings and both of its stale captures are fixed, and the author went past the brief in three useful ways: catching two order-panel changes my round-3 "survived unchanged" table got wrong, documenting **Reset country** which no version of this page had covered, and propagating the corrections into the two sibling reference docs so the operator-doc set agrees with itself. The end-date attribution in the Poland example was silently wrong before and is now right.

I have no blocking findings. The one open test-plan box (the companion `openlinker-docs` `SOURCES` + sidebar PR) is tracked on the issue and correctly not a blocker here.

One process note, unchanged from round 3 and now with a worked answer: nothing in `check:invariants` can see a docs page invalidated by a UI change it never touched. What made this recoverable was re-verifying every quoted string and opening every capture — the merge commit said nothing.

Issue #2440, opened 2026-08-24
[IMPL] Backend + Frontend — net tax basis for /analytics (excluded-orders reporting)


norbert-kulus-blockydevs:
## Design source

**Product Design parent:** Part of #1976 — see spec at `docs/specs/metrics-analytics-dashboard.md` § "Rules common to all metrics". Depends on epic #2245 (per-line tax rate, PR #2260) having landed a rate on `order_line_items.taxRate`, `products.taxRate` / `product_variants.taxRate`.

## Scope revised 2026-08-24

The original scope below said "exclusion, never estimation" for every order with no persisted rate. That rule stays for orders where the product/variant is gone from the catalogue too — nothing to look up. For every other rate-less line, the decision is now: **backfill `order_line_items.taxRate` from the CURRENT catalogue rate**, tagged with a distinct `taxSource` value so it is never confused with a rate the shop confirmed at order time. Rationale:

- The rate almost never changes for a given product's tax category in practice, so treating "current catalogue rate" as usable for internal reporting is an acceptable trade — it is explicitly **not** treated as legally confirmed.
- This backfill is scoped to **internal OpenLinker data only** (`order_line_items`, and anything that reads it — analytics, the order detail panel, exports). It never touches an already-issued invoice or fiscal receipt (KSeF/inFakt/Subiekt/eparagony) and never triggers a correction document. Those stay exactly as issued.
- `order_records.taxRateEra = 'pre-rollout'` is **untouched** by this backfill. It keeps gating the fiscal write paths (auto-issue, invoice issuance, fiscal registration) exactly as ADR-063 describes — a backfilled, best-effort rate must never satisfy strict enforcement for a *new* fiscal document.
- A line whose product/variant no longer exists in the catalogue (deleted, unmapped) has nothing to backfill from and stays excluded, counted in the excluded bucket as before.

## Scope

Nearly every headline metric on `/analytics` is defined net (Net Sales, AOV, Median Order Value, Returns Value, Cancellations Value); only GMV is gross, and the spec says explicitly not to convert it.

- **One pure derivation function** in `libs/core` (gross amount + rate code → net), living beside its types per the repo's "pure-rule exception" (`docs/engineering-standards.md`) — same shape as `applyPricingRule` / `applyStockSafetyBuffer` / `splitShippingAcrossRates`. `order_line_items` carries no separate net/gross columns, so net is always derived at read time, never a second persisted figure that can drift from `unitPrice`.
[...]

Issue #2516, opened 2026-08-26
[TASK] CORE — batched sales-document read for the orders list


norbert-kulus-blockydevs:
## Problem / Context

The orders list renders a document fact per row. Assembling it per row would issue two reads per order, which is the N+1 shape the analytics-trust work explicitly avoided by batching the earliest-order-date read across every enumerated connection.

## Proposed Solution

Add a service method taking a list of order ids and returning a map to the projection, reading both underlying stores once each. Follow the `getEarliestOrderDateByConnection` precedent: one batched call outside any per-row loop.

## Classification

**Type**: CORE
**Layer**: Application
**File(s)**: `libs/core/src/orders/application/services/`, `libs/core/src/invoicing/`, `libs/core/src/fiscalization/`

## Dependencies

- Parent mini-epic: #2514
- The projection type task in this mini-epic.

## Acceptance Criteria

- [ ] Rendering a page of orders performs a fixed number of queries regardless of page size.
- [ ] An order with no document at all is present in the result with a null kind, not absent from the map.
- [ ] A spec asserts the query count does not grow with the number of ids.
- [ ] Tests added or updated for non-trivial logic
- [ ] No architecture boundary violations (CORE ↔ Integration)

## Branching and merge target

Branch: `{issue#}-{short-kebab}`, created **from its mini-epic branch**.
PR target: **its mini-epic branch**.

## How this issue gets finished

1. Implement on the branch above.
2. Open the PR into its mini-epic branch.
3. Run `/pr-review <PR>` and post the verdict as a PR comment.
4. Fix **every** finding, and reply under the comment saying what changed for each one.
5. Run `/pr-review` again. Repeat until the review comes back approved with no findings.
6. Only then merge.

Quality gate before every commit: `pnpm lint`, `pnpm type-check`, `pnpm test`.
Commits are signed off (`git commit -s`).

norbert-kulus-blockydevs:
Merged into the mini-epic branch. Closing manually: the PR targeted a non-default branch, so `Closes #N` does not fire until the epic reaches `main`.

PR #2832, opened 2026-09-02
fix(orders): let a PrestaShop order's total be evaluated by an orderTotalGross sales-document rule

### LINKED ISSUE #2829: [BUG] PrestaShop order source — orderTotalGross sales-document rules never match (taxTreatment: 'exclusive' hardcoded)
## Problem / Context

`evaluateSalesDocumentRules`'s `orderTotalGross` condition (`libs/core/src/sales-documents/domain/domain-services/evaluate-sales-document-rules.ts`, `checkAmountConditionDataProblem`) refuses to compare an order's total against a threshold unless `SalesDocumentOrderFacts.taxTreatment === 'inclusive'` — an explicit ADR-063 stance ("never guess on a figure feeding a legal-document decision"). Any other value, including `'exclusive'`, is treated as a `net-priced` data problem and the condition never matches, regardless of the threshold or comparison operator.

`PrestashopOrderSourceAdapter`'s order mapper (`libs/integrations/prestashop/src/infrastructure/mappers/prestashop-order.mapper.ts:93`) unconditionally sets `taxTreatment: 'exclusive'` on every order it maps, with a comment explaining the reasoning: PrestaShop's per-line `order_details.product_price` (mapped onto `OrderItem.price`) is genuinely net, matching how `specific_price` and the catalogue reads elsewhere already treat it (#2440). This is correct about the LINE prices, but it is the only signal `toSalesDocumentOrderFacts` (`libs/core/src/invoicing/application/mappers/order-to-sales-document-order-facts.mapper.ts`) reads for the whole order's `taxTreatment` — and the order's TOTAL (`order.totals.total`, which the same mapper sets from `total_paid_tax_incl` two lines above) genuinely is gross.

The two facts compose into a structural gap, not a data-quality problem: PrestaShop is the only order source in the codebase whose adapter reads a real buyer tax ID (`ps_address.vat_number`, see `docs/architecture-overview.md` § 17 Sales Documents), and it is also the one source whose `taxTreatment` can never read as `'inclusive'`. So no PrestaShop-sourced order can ever satisfy an `orderTotalGross` rule condition, whatever the order's actual tax-rule setup is — a routing pattern such as "invoice via provider A above 1000 PLN, provider B below, for buyers with a tax ID" has no live PrestaShop order path that can trigger it.

**Verified live** (ol-demo-fresh, PR #2809's e2e pass): a real order placed through PrestaShop's admin UI "New order" wizard, for a customer with a checksum-valid test NIP, with a catalog product carrying a real PL 23% VAT tax rule, run through PrestaShop's actual `validateOrder` tax engine end to end (1499.00 excl. → 1843.77 incl., 23% VAT). `order_records.buyerTaxId` and `order_line_items.taxRate = '23'` were both correctly captured, yet the routing engine refused with `unresolved-routing / net-priced-order`.

## Proposed Solution

Have `PrestashopOrderMapper` distinguish "are the LINE prices net" (true, unconditionally, per #2440 — unaffected) from "is the ORDER TOTAL gross" (true whenever `total_paid_tax_incl` is used, which it always is, per the totals block just above). Concretely: set `taxTreatment: 'inclusive'` in `prestashop-order.mapper.ts`'s totals block, since `totals.total` is already sourced from `total_paid_tax_incl`, not `total_paid_tax_excl` — the field is meant to describe the total's own treatment for exactly this kind of downstream amount comparison, and today it is describing the wrong thing (line pricing) instead.

Before changing the value, audit every other consumer of `OrderTotals.taxTreatment` (e.g. `PrestashopOrderProcessorManagerAdapter`'s `convertGrossToNet` at `libs/integrations/prestashop/src/infrastructure/adapters/prestashop-order-processor-manager.adapter.ts:664`, and the ADR-063 net-sales tax-rate resolution path in `libs/core/src/orders/`) to confirm none of them rely on the current (arguably mismatched) semantics for a PrestaShop-sourced order, since this field is read in more than one place across the codebase.

If, on investigation, `'inclusive'` turns out to be unsafe for one of those other consumers, the alternative is giving `orderTotalGross` conditions a net-to-gross computation path driven by `order_line_items.taxRate` instead of requiring the blunter `taxTreatment: 'inclusive'` flag — but the mapper fix is the smaller, more targeted change if it survives the consumer audit.

## Classification

**Type**: Integration + CORE
**Layer**: Infrastructure (mapper) / Domain (rule engine, for context only — no change expected there)
**File(s)**:
- `libs/integrations/prestashop/src/infrastructure/mappers/prestashop-order.mapper.ts`
- `libs/integrations/prestashop/src/infrastructure/adapters/prestashop-order-processor-manager.adapter.ts` (consumer to audit)
- `libs/core/src/sales-documents/domain/domain-services/evaluate-sales-document-rules.ts` (unaffected, referenced for context)
- `libs/core/src/invoicing/application/mappers/order-to-sales-document-order-facts.mapper.ts` (unaffected, referenced for context)

## Dependencies

None.

## Assumptions

- The fix belongs in the mapper (a data-correctness fix: the field should describe what it says it describes), not in the rule engine's `orderTotalGross` condition, since ADR-063's refusal-over-guessing stance for that condition is itself correct and intentional.
- No other repo test currently pins `taxTreatment: 'exclusive'` as an assertion on a PrestaShop-sourced order in a way that would need updating — this needs confirming during implementation.

## Acceptance Criteria

- [ ] `PrestashopOrderMapper`'s totals block sets `taxTreatment: 'inclusive'` (or an equivalent fix that lets a PrestaShop order's total be evaluated by an `orderTotalGross` rule condition)
- [ ] Every other consumer of `OrderTotals.taxTreatment` for a PrestaShop-sourced order is audited and confirmed unaffected (or updated to match), specifically `PrestashopOrderProcessorManagerAdapter.convertGrossToNet`
- [ ] A real order placed via PrestaShop with a valid buyer tax ID and a correctly tax-rated catalog product can match an `orderTotalGross` sales-document rule end to end
- [ ] Tests added/updated for the mapper's `taxTreatment` output and for the affected adapter consumer

norbert-kulus-blockydevs:
## Problem

`PrestashopOrderMapper` sets `taxTreatment: 'exclusive'` unconditionally on every order it maps — correct for line prices/subtotal (`order_details.product_price`, `total_paid_tax_excl`), per #2440. But `evaluateSalesDocumentRules`'s `orderTotalGross` condition reads that same flag to decide whether it may trust `SalesDocumentOrderFacts.totalGross` as a real gross figure, and refuses (`net-priced` data problem) unless it reads `'inclusive'`.

A PrestaShop order's total (`total_paid_tax_incl`) genuinely IS gross even though its line prices/subtotal are net — so no PrestaShop-sourced order could ever match an `orderTotalGross` sales-document rule condition, regardless of the order's actual tax setup.

## Why not just flip the flag

Audited every consumer of `OrderTotals.taxTreatment` per the issue's acceptance criteria. Flipping it to `'inclusive'` for PrestaShop is unsafe:

- **`PrestashopOrderProcessorManagerAdapter.convertGrossToNet`** (destination-side line pinning): reads `taxTreatment !== 'exclusive'` to decide whether line prices need gross→net conversion before pinning. For a PrestaShop-sourced order later routed to another PrestaShop destination (a live path — `OrderSyncService`'s default unfiltered fan-out sends to every `OrderProcessorManager`-capable connection except the source), flipping the flag would make this treat already-net line prices as gross and double-convert them.
- **The ADR-063 net-sales tax-rate resolution path** (`net-sales-tax-rate.types.ts`, `deriveNetLineAmount` + its SQL twins): reads `taxTreatment === 'exclusive'` to mean "line `unitPrice` is already net, no rate-based conversion needed" — the whole net-sales analytics feature for PrestaShop orders depends on this staying `'exclusive'`. Flipping it would silently understate net revenue for every PrestaShop order (double VAT stripping).
- WooCommerce's order mapper has the **identical structural split** (net line prices + `total_tax` decomposed separately, but `total` is genuinely gross) — same reasoning applies there, out of scope for this PR but worth a follow-up issue.

## Fix

Added a narrower `OrderTotals.totalTaxTreatment` field describing only the **total's own** inclusivity, independent of `taxTreatment` (which stays scoped to line-price/subtotal treatment, unchanged in meaning and behavior for every existing consumer). `toSalesDocumentOrderFacts` now reads `totalTaxTreatment ?? taxTreatment` — every source that doesn't set the new field falls back exactly as before. `PrestashopOrderMapper` sets `totalTaxTreatment: 'inclusive'` alongside the unchanged `taxTreatment: 'exclusive'`.

Mirrored onto `IncomingOrderTotals` (the mapper-output shape consumed by `PrestashopOrderSourceAdapter.getOrder()`) so the field survives the source→`Order` build in `OrderIngestionService.buildUnifiedOrder`.

## Testing

- `prestashop-order.mapper.spec.ts`: asserts `totalTaxTreatment: 'inclusive'` alongside the unchanged `taxTreatment: 'exclusive'`.
- `order-to-sales-document-order-facts.mapper.spec.ts`: new cases for the `totalTaxTreatment` override and the fallback-to-`taxTreatment` path when unset.
- Full `pnpm type-check` (all packages) and `pnpm lint` (all invariant checks) pass.
- Ran the full affected-consumer test suites: PrestaShop order-source adapter + order-processor-manager adapter (151 tests), net-sales-tax-rate / evaluate-sales-document-rules / order-record.service / order-ingestion (214 tests) — all pass, confirming `convertGrossToNet` and the net-sales path are unaffected.

Not done (per scope discussion): the issue's live e2e verification step on ol-demo-fresh — deferred as a manual follow-up check, not blocking this PR given the unit-level coverage above.

Closes #2829

### DIFF
--- a/apps/api/src/migrations/1870000000000-add-order-record-total-tax-treatment.ts
+++ b/apps/api/src/migrations/1870000000000-add-order-record-total-tax-treatment.ts
@@ -1,0 +1,69 @@
+/**
+ * Add `order_records.totalTaxTreatment` and backfill it for PrestaShop (#2829/#2832).
+ *
+ * `order_records.taxTreatment` (#1985/#2440) describes the LINE prices /
+ * subtotal — for PrestaShop those are net (`order_details.product_price`),
+ * so the column reads `'exclusive'`. But PrestaShop's order TOTAL
+ * (`total_paid_tax_incl`) genuinely is gross, and the `orderTotalGross`
+ * sales-document rule condition reads `taxTreatment` to decide whether it may
+ * trust the total as gross at all — so no PrestaShop order could ever match
+ * that condition, regardless of its real tax setup.
+ *
+ * `taxTreatment` cannot simply be flipped to `'inclusive'`: it also drives
+ * `PrestashopOrderProcessorManagerAdapter.convertGrossToNet` (destination-side
+ * line pinning) and the ADR-063 net-sales tax-rate resolution path
+ * (`deriveNetLineAmount` + its SQL twins), both of which need the LINE-level
+ * fact to stay `'exclusive'`. `totalTaxTreatment` is the narrower, additive
+ * column describing the total's own inclusivity alone — `null` means "same as
+ * `taxTreatment`" (see `OrderRecord.totalTaxTreatment` and
+ * `SalesDocumentViewService.toOrderFacts`, which reads
+ * `totalTaxTreatment ?? taxTreatment`, the same fallback the live-`Order`
+ * gate applies via `toSalesDocumentOrderFacts`).
+ *
+ * The adapter fix (companion PR) only reaches orders ingested from now on.
+ * This migration backfills history for PrestaShop, following the exact
+ * platform-scoped join precedent set by
+ * `1841000000006-backfill-order-records-tax-treatment.ts` (itself following
+ * `1840000000000-reset-fx-stamp-for-mislabelled-prestashop-orders.ts`).
+ *
+ * Idempotent by construction (`WHERE "totalTaxTreatment" IS NULL`) — a re-run
+ * touches only rows a later fix hasn't already reached, matching the house
+ * convention documented on `1818000000004-backfill-ksef-provider-invoice-number.ts`.
+ *
+ * Deliberately does NOT touch WooCommerce, despite the identical structural
+ * split (net line prices + `total_tax` decomposed separately, but `total` is
+ * genuinely gross) — tracked as a follow-up (#2836) and cited from
+ * `OrderTotals.totalTaxTreatment`'s doc comment. A connection whose platform
+ * this migration does not assert stays `totalTaxTreatment IS NULL`, i.e.
+ * "same as `taxTreatment`", which is the existing (net-priced) behaviour —
+ * never a guess.
+ *
+ * Generated: 2026-09-03 (synthetic sequential prefix per docs/migrations.md
+ * rule 3; 1869000000900 is #2385's `automation_runs` retry-attempt column).
+ */
+import type { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class AddOrderRecordTotalTaxTreatment1870000000000 implements MigrationInterface {
+  name = 'AddOrderRecordTotalTaxTreatment1870000000000';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `ALTER TABLE "order_records" ADD COLUMN IF NOT EXISTS "totalTaxTreatment" varchar`
+    );
+
+    await queryRunner.query(`
+      UPDATE "order_records" o
+      SET "totalTaxTreatment" = 'inclusive'
+      FROM "connections" c
+      WHERE c."id" = o."sourceConnectionId"
+        AND c."platformType" = 'prestashop'
+        AND o."totalTaxTreatment" IS NULL
+    `);
+  }
+
+  public async down(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `ALTER TABLE "order_records" DROP COLUMN IF EXISTS "totalTaxTreatment"`
+    );
+  }
+}
--- a/apps/api/src/migrations/1870000000000-add-order-record-total-tax-treatment.ts
+++ b/apps/api/src/migrations/1870000000000-add-order-record-total-tax-treatment.ts
@@ -62,6 +62,9 @@ export class AddOrderRecordTotalTaxTreatment1870000000000 implements MigrationIn
   }
 
   public async down(queryRunner: QueryRunner): Promise<void> {
+    // Unlike 1841000000006's down() (a deliberate no-op — that migration only
+    // SET a pre-existing column), this one ADDS the column, so dropping it is
+    // the only sound reversal.
     await queryRunner.query(
       `ALTER TABLE "order_records" DROP COLUMN IF EXISTS "totalTaxTreatment"`
     );
--- a/libs/core/src/invoicing/application/mappers/order-to-sales-document-order-facts.mapper.spec.ts
+++ b/libs/core/src/invoicing/application/mappers/order-to-sales-document-order-facts.mapper.spec.ts
@@ -201,4 +201,32 @@ describe('toSalesDocumentOrderFacts (#2173)', () => {
 
     expect(toSalesDocumentOrderFacts(order)?.taxTreatment).toBe('exclusive');
   });
+
+  it('should honour totalTaxTreatment even when it disagrees downward from taxTreatment', () => {
+    // The fallback is `totalTaxTreatment ?? taxTreatment`, which must win in
+    // BOTH directions — the sibling test above only proves the
+    // exclusive→inclusive override; a regression to `||` or a truthiness
+    // check would still pass that one but silently ignore an explicit
+    // `'exclusive'` here (a falsy-looking-but-defined string is never the
+    // failure mode of `??`, so this guards the direction `||` would break).
+    const order = makeOrder({
+      shippingAddress: {
+        address1: 'ul. Testowa 1',
+        city: 'Poznań',
+        postalCode: '60-001',
+        country: 'PL',
+      },
+      totals: {
+        subtotal: 20,
+        tax: 0,
+        shipping: 0,
+        total: 20,
+        currency: 'PLN',
+        taxTreatment: 'inclusive',
+        totalTaxTreatment: 'exclusive',
+      },
+    });
+
+    expect(toSalesDocumentOrderFacts(order)?.taxTreatment).toBe('exclusive');
+  });
 });
--- a/libs/core/src/invoicing/application/mappers/order-to-sales-document-order-facts.mapper.spec.ts
+++ b/libs/core/src/invoicing/application/mappers/order-to-sales-document-order-facts.mapper.spec.ts
@@ -153,4 +153,52 @@ describe('toSalesDocumentOrderFacts (#2173)', () => {
     expect(facts?.currency).toBe('EUR');
     expect(facts?.totalGross).toBe(100);
   });
+
+  it('should prefer totalTaxTreatment over taxTreatment when a source asserts total is gross despite net line prices (#2829)', () => {
+    const order = makeOrder({
+      shippingAddress: {
+        address1: 'ul. Testowa 1',
+        city: 'Poznań',
+        postalCode: '60-001',
+        country: 'PL',
+      },
+      totals: {
+        subtotal: 83.32,
+        tax: 16.67,
+        shipping: 5,
+        total: 99.99,
+        currency: 'PLN',
+        // PrestaShop-shaped: line prices/subtotal are net, but the order
+        // total is genuinely gross.
+        taxTreatment: 'exclusive',
+        totalTaxTreatment: 'inclusive',
+      },
+    });
+
+    const facts = toSalesDocumentOrderFacts(order);
+
+    expect(facts?.taxTreatment).toBe('inclusive');
+    expect(facts?.totalGross).toBe(99.99);
+  });
+
+  it('should fall back to taxTreatment when totalTaxTreatment is absent', () => {
+    const order = makeOrder({
+      shippingAddress: {
+        address1: 'ul. Testowa 1',
+        city: 'Poznań',
+        postalCode: '60-001',
+        country: 'PL',
+      },
+      totals: {
+        subtotal: 20,
+        tax: 0,
+        shipping: 0,
+        total: 20,
+        currency: 'PLN',
+        taxTreatment: 'exclusive',
+      },
+    });
+
+    expect(toSalesDocumentOrderFacts(order)?.taxTreatment).toBe('exclusive');
+  });
 });
--- a/libs/core/src/invoicing/application/mappers/order-to-sales-document-order-facts.mapper.ts
+++ b/libs/core/src/invoicing/application/mappers/order-to-sales-document-order-facts.mapper.ts
@@ -14,10 +14,18 @@
  *   caller treats this exactly like "no rule-engine configuration for this
  *   order" and falls back to the pre-#2170 single-primary resolver rather than
  *   guessing a jurisdiction.
- * - `totalGross` / `currency` / `taxTreatment` — read verbatim from
- *   `order.totals`. The evaluator itself is what turns an `exclusive`
- *   treatment into the terminal `net-priced-order` signal; this mapper never
- *   re-derives or converts anything.
+ * - `totalGross` / `currency` — read verbatim from `order.totals`.
+ * - `taxTreatment` — `order.totals.totalTaxTreatment ?? order.totals.taxTreatment`
+ *   (#2829). `totals.taxTreatment` is source-uniform and also governs the
+ *   per-line/subtotal net-conversion path (`convertGrossToNet`, the ADR-063
+ *   net-sales tax-rate resolution), so a source whose line prices are net but
+ *   whose `total` is genuinely gross (PrestaShop, WooCommerce) cannot flip it
+ *   without breaking those. `totalTaxTreatment` is the narrower, `total`-only
+ *   signal such a source may set instead; every other source leaves it unset
+ *   and falls back to `taxTreatment` unchanged. The evaluator itself is what
+ *   turns an `exclusive` (or absent) result into the terminal
+ *   `net-priced-order` signal; this mapper never re-derives or converts
+ *   anything.
  * - `buyerHasTaxId` - read from the order's own buyer tax id (#2599), which
  *   closed the prerequisite ADR-041 decision 5 was waiting on. It stays
  *   `undefined` whenever the source asserted nothing, and is `false` only when
@@ -54,7 +62,7 @@ export function toSalesDocumentOrderFacts(order: Order): SalesDocumentOrderFacts
     country,
     totalGross: order.totals.total,
     currency: order.totals.currency,
-    taxTreatment: order.totals.taxTreatment,
+    taxTreatment: order.totals.totalTaxTreatment ?? order.totals.taxTreatment,
     // Stays `undefined` when the source asserted nothing — see the module doc
     // comment. Never defaulted to `false`.
     buyerHasTaxId: buyerHasTaxId(readBuyerTaxId(order)),
--- a/libs/core/src/orders/application/services/order-record.service.ts
+++ b/libs/core/src/orders/application/services/order-record.service.ts
@@ -288,7 +288,8 @@ export class OrderRecordService implements IOrderRecordService {
       null,
       [],
       buyerTaxId,
-      shippingAddressHash
+      shippingAddressHash,
+      analyticsScalars.totalTaxTreatment
     );
 
     // #1985: persist the order record AND its order_line_items rows in one
--- a/libs/core/src/orders/application/services/sales-document-view.service.spec.ts
+++ b/libs/core/src/orders/application/services/sales-document-view.service.spec.ts
@@ -255,6 +255,39 @@ describe('SalesDocumentViewService', () => {
     },
   );
 
+  it('should prefer totalTaxTreatment over taxTreatment in the routing facts (#2829/#2832)', async () => {
+    // Must match `toSalesDocumentOrderFacts`'s `totalTaxTreatment ?? taxTreatment`
+    // fallback exactly: a PrestaShop order (line prices net, total gross)
+    // routed as gross by the gate must not be reported net-priced here.
+    orderRecords.findByIds.mockResolvedValue([
+      orderRecord({ taxTreatment: 'exclusive', totalTaxTreatment: 'inclusive' }),
+    ]);
+    rules.resolveRoutingBatch.mockResolvedValue([
+      { kind: 'unresolved', reason: 'no-configuration-for-country' },
+    ]);
+
+    await service.getForOrders(['ol_order_1']);
+
+    expect(rules.resolveRoutingBatch).toHaveBeenCalledWith([
+      expect.objectContaining({ taxTreatment: 'inclusive' }),
+    ]);
+  });
+
+  it('should fall back to taxTreatment when totalTaxTreatment is absent from the record', async () => {
+    orderRecords.findByIds.mockResolvedValue([
+      orderRecord({ taxTreatment: 'exclusive', totalTaxTreatment: null }),
+    ]);
+    rules.resolveRoutingBatch.mockResolvedValue([
+      { kind: 'unresolved', reason: 'no-configuration-for-country' },
+    ]);
+
+    await service.getForOrders(['ol_order_1']);
+
+    expect(rules.resolveRoutingBatch).toHaveBeenCalledWith([
+      expect.objectContaining({ taxTreatment: 'exclusive' }),
+    ]);
+  });
+
   it('should keep an order with no document in the map, with the kind routing resolves', async () => {
     orderRecords.findByIds.mockResolvedValue([orderRecord()]);
     connections.list.mockResolvedValue([connection()]);
--- a/libs/core/src/orders/application/services/sales-document-view.service.ts
+++ b/libs/core/src/orders/application/services/sales-document-view.service.ts
@@ -266,17 +266,27 @@ function toDocumentKind(decision: SalesDocumentDecision | null): SalesDocumentKi
  * source asserted the buyer has none) - `buyerTaxIdState` is the only intended
  * read of the column, because a bare `!== null` test reports true for the
  * asserted-none row.
+ *
+ * `taxTreatment` reads `record.totalTaxTreatment ?? record.taxTreatment`
+ * (#2829/#2832), the SAME fallback `toSalesDocumentOrderFacts` applies from
+ * the live `Order` - this is the `totalGross` field's own inclusivity, which
+ * can diverge from the line-price/subtotal `taxTreatment` (a PrestaShop order
+ * prices lines net but its total gross). Reading `record.taxTreatment` alone
+ * here would report every PrestaShop order as `net-priced` to the operator
+ * while the gate itself resolves it as gross, and the two must agree - see
+ * the country-address comment above, whose argument binds identically here.
  */
 function toOrderFacts(record: OrderRecord): SalesDocumentOrderFacts | null {
   const country = readDeliveryCountry(record.orderSnapshot);
   if (country === null || record.totalAmount === null || record.currency === null) {
     return null;
   }
+  const totalTaxTreatment = record.totalTaxTreatment ?? record.taxTreatment;
   return {
     country,
     totalGross: record.totalAmount,
     currency: record.currency,
-    ...(isTaxTreatment(record.taxTreatment) ? { taxTreatment: record.taxTreatment } : {}),
+    ...(isTaxTreatment(totalTaxTreatment) ? { taxTreatment: totalTaxTreatment } : {}),
     buyerHasTaxId: buyerHasTaxId(record.buyerTaxIdState),
   };
 }
--- a/libs/core/src/orders/domain/entities/order-record.entity.ts
+++ b/libs/core/src/orders/domain/entities/order-record.entity.ts
@@ -358,7 +358,21 @@ export class OrderRecord {
      * positional constructor, so a field inserted mid-list would silently
      * shift every argument after it at each construction site.
      */
-    public readonly shippingAddressHash: string | null = null
+    public readonly shippingAddressHash: string | null = null,
+    /**
+     * How `totalAmount` ALONE expresses tax, when that diverges from
+     * `taxTreatment` (#2829/#2832). Mirrors `OrderTotals.totalTaxTreatment` -
+     * see that field's doc comment for the full rationale (a PrestaShop order
+     * prices its LINES net but its TOTAL gross, so one flag cannot describe
+     * both). `null` means "same as `taxTreatment`"; every reader of the total's
+     * own inclusivity (today: `SalesDocumentViewService.toOrderFacts`, feeding
+     * the `orderTotalGross` sales-document rule condition) MUST read
+     * `totalTaxTreatment ?? taxTreatment`, exactly as `toSalesDocumentOrderFacts`
+     * does from the live `Order` - the gate and this projection must agree, or
+     * a PrestaShop order would be routed as gross while the operator-facing
+     * surface still reports it `net-priced`.
+     */
+    public readonly totalTaxTreatment: PriceTaxTreatment | null = null
   ) {}
 
   /**
--- a/libs/core/src/orders/domain/order-analytics-projection.spec.ts
+++ b/libs/core/src/orders/domain/order-analytics-projection.spec.ts
@@ -28,7 +28,7 @@ describe('orderAnalyticsProjection', () => {
   });
 
   describe('deriveOrderAnalyticsScalars', () => {
-    it('derives all 4 scalars when the order carries them', () => {
+    it('derives all 5 scalars when the order carries them', () => {
       const order = { ...baseOrder(), placedAt: new Date('2026-08-01T09:00:00.000Z') };
 
       expect(deriveOrderAnalyticsScalars(order)).toEqual({
@@ -36,7 +36,26 @@ describe('orderAnalyticsProjection', () => {
         currency: 'PLN',
         taxTreatment: 'inclusive',
         totalAmount: 32,
+        totalTaxTreatment: null,
       });
+    });
+
+    it('derives totalTaxTreatment when the source scopes it to the total alone (#2829)', () => {
+      const order = baseOrder();
+      order.totals.taxTreatment = 'exclusive';
+      order.totals.totalTaxTreatment = 'inclusive';
+
+      expect(deriveOrderAnalyticsScalars(order)).toMatchObject({
+        taxTreatment: 'exclusive',
+        totalTaxTreatment: 'inclusive',
+      });
+    });
+
+    it('returns totalTaxTreatment: null when the source did not diverge from taxTreatment', () => {
+      const order = baseOrder();
+      expect(order.totals.totalTaxTreatment).toBeUndefined();
+
+      expect(deriveOrderAnalyticsScalars(order).totalTaxTreatment).toBeNull();
     });
 
     it('returns placedAt: null when the order does not carry one', () => {
--- a/libs/core/src/orders/domain/order-analytics-projection.ts
+++ b/libs/core/src/orders/domain/order-analytics-projection.ts
@@ -3,8 +3,8 @@
  *
  * Pure derivation (no I/O): called by `OrderRecordService.persistOrder` at the
  * moment a `ready` order is persisted, to populate the denormalized
- * `OrderRecord` scalars (`placedAt`/`currency`/`taxTreatment`/`totalAmount`)
- * and the `order_line_items` rows — see ADR-039 for the persistence-strategy
+ * `OrderRecord` scalars (`placedAt`/`currency`/`taxTreatment`/`totalAmount`/
+ * `totalTaxTreatment`) and the `order_line_items` rows — see ADR-039 for the persistence-strategy
  * decision. Deliberately never throws: a malformed/partial `Order` degrades to
  * `null` scalars or an empty item list rather than failing the whole ingest.
  *
@@ -20,10 +20,16 @@ export interface OrderAnalyticsScalars {
   currency: string | null;
   taxTreatment: PriceTaxTreatment | null;
   totalAmount: number | null;
+  /**
+   * How `totalAmount` ALONE expresses tax, when that diverges from
+   * `taxTreatment` (#2829/#2832) — see `OrderTotals.totalTaxTreatment` and
+   * `OrderRecord.totalTaxTreatment`. `null` means "same as `taxTreatment`".
+   */
+  totalTaxTreatment: PriceTaxTreatment | null;
 }
 
 /**
- * Derive the 4 order-level scalars from an already-resolved {@link Order}.
+ * Derive the 5 order-level scalars from an already-resolved {@link Order}.
  * Pure function of its own argument — no I/O, never throws.
  */
 export function deriveOrderAnalyticsScalars(order: Order): OrderAnalyticsScalars {
@@ -32,6 +38,7 @@ export function deriveOrderAnalyticsScalars(order: Order): OrderAnalyticsScalars
     currency: order.totals?.currency ?? null,
     taxTreatment: order.totals?.taxTreatment ?? null,
     totalAmount: typeof order.totals?.total === 'number' ? order.totals.total : null,
+    totalTaxTreatment: order.totals?.totalTaxTreatment ?? null,
   };
 }
 
--- a/libs/core/src/orders/domain/order-from-ready-snapshot.ts
+++ b/libs/core/src/orders/domain/order-from-ready-snapshot.ts
@@ -226,6 +226,13 @@ function readTotals(value: unknown): OrderTotals {
   if (raw.taxTreatment === 'inclusive' || raw.taxTreatment === 'exclusive') {
     totals.taxTreatment = raw.taxTreatment;
   }
+  // #2829/#2832 — same allowlist, same failure mode as the per-line tax
+  // fields above: an omitted field here is silently dropped from every
+  // rehydrated `Order`, and a PrestaShop order's total-only gross assertion
+  // would vanish on any manual-issuance path that reaches this function.
+  if (raw.totalTaxTreatment === 'inclusive' || raw.totalTaxTreatment === 'exclusive') {
+    totals.totalTaxTreatment = raw.totalTaxTreatment;
+  }
   return totals;
 }
 
--- a/libs/core/src/orders/domain/types/incoming-order.types.ts
+++ b/libs/core/src/orders/domain/types/incoming-order.types.ts
@@ -208,6 +208,12 @@ export interface IncomingOrderTotals {
    * not assert it.
    */
   taxTreatment?: PriceTaxTreatment;
+
+  /**
+   * How `total` ALONE expresses tax, when it diverges from `taxTreatment`. See
+   * {@link OrderTotals.totalTaxTreatment} (#2829).
+   */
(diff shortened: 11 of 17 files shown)


norbert-kulus-blockydevs:
## PR Review — `fix(orders): let a PrestaShop order's total be evaluated by an orderTotalGross sales-document rule`

**Verdict: ✅ Approve**

### What it does
`PrestashopOrderMapper` sets `taxTreatment: 'exclusive'` (line prices are net, correct per #2440), but the sales-document rule engine's `orderTotalGross` condition reads that same field to decide whether the order **total** can be trusted as gross — so no PrestaShop order could ever match. Fix adds a narrower `OrderTotals.totalTaxTreatment` field describing only the total's own inclusivity, consumed via `totalTaxTreatment ?? taxTreatment` in `toSalesDocumentOrderFacts`. Every other source is unaffected (fallback preserves current behavior byte-for-byte).

### Verification performed
- Confirmed `OrderIngestionService.buildUnifiedOrder` does a whole-object assignment (`totals: incoming.totals`), so the new `IncomingOrderTotals.totalTaxTreatment` field survives the source→`Order` build with **no additional wiring needed** — the PR doesn't touch that file, and it doesn't need to.
- Confirmed the two "don't touch this" claims in the PR description are accurate by grepping actual consumers, not just trusting the description:
  - `PrestashopOrderProcessorManagerAdapter.convertGrossToNet` (line 664) reads `order.totals.taxTreatment` — untouched, still `'exclusive'`.
  - `net-sales-tax-rate.types.ts`'s `deriveNetLineAmount` + SQL twins read `taxTreatment` — untouched.
- Ran `order-to-sales-document-order-facts.mapper.spec.ts` (9/9 pass) and `prestashop-order.mapper.spec.ts` (22/22 pass) locally.
- `pnpm --filter @openlinker/core type-check` is clean.

### Architecture / conventions
- Correctly uses the codebase's established "narrower field + fallback chain, absent = same as before" pattern (mirrors how `taxTreatment` itself was introduced) rather than repurposing an existing source-uniform flag for two meanings.
- Type placement is correct: new field lives beside `taxTreatment` in both `OrderTotals` (`order.types.ts`) and `IncomingOrderTotals` (`incoming-order.types.ts`), with doc comments explaining the *why* (non-obvious constraint: one flag can't express "line prices net, total gross"), consistent with `docs/engineering-standards.md § File Headers` / comment guidance.
- No cross-context boundary issues — `SalesDocumentOrderFacts.taxTreatment` already existed; only the *source* of the value at the `toSalesDocumentOrderFacts` mapper changed.
- No migration needed (pure in-memory DTO fields, no persistence change).

### 🟢 Minor observation (non-blocking)
`order-to-sales-document-order-facts.mapper.spec.ts:132` — the pre-existing test `'should carry taxTreatment verbatim, including absent (never re-derived)'` still passes (since `totalTaxTreatment` is absent → falls back to `taxTreatment` which is also absent → `undefined`), but its name is now slightly stale: the mapper *does* now derive `taxTreatment` from a fallback chain rather than a straight passthrough. Not worth a follow-up on its own — just flagging in case it's touched again.

### Testing
Description note that live e2e verification on ol-demo-fresh is deferred as a manual follow-up is reasonable given the unit-level coverage — this is a pure type/mapping fix with no I/O, no migration, no adapter-boundary behavior change; the audited-consumer approach here (grepping every actual reader of `taxTreatment` and confirming none of them read the new field) is exactly the right level of rigor for a fix this narrow.

No blocking or important findings.

piotrswierzy (changes requested):
## Code Review — #2832 (`fix(orders)`: `totalTaxTreatment` for PrestaShop)

**Scope reviewed** (6 files, +99/−5): `order.types.ts`, `incoming-order.types.ts`, `order-to-sales-document-order-facts.mapper.ts` (+spec), `prestashop-order.mapper.ts` (+spec).

### 1. Summary

The diagnosis is right and the chosen fix is the right *shape*. `taxTreatment` genuinely is source-uniform and load-bearing for two other consumers (`PrestashopOrderProcessorManagerAdapter.convertGrossToNet` at `prestashop-order-processor-manager.adapter.ts:664`, and `deriveNetLineAmount` + its SQL twins in `net-sales-tax-rate.types.ts:110/157/179`), so flipping it would have double-converted line prices on the PS→PS destination path and double-stripped VAT in net-sales analytics. Adding a narrower, additive, `?? taxTreatment`-defaulted `totalTaxTreatment` is a correct, non-breaking way to say the one thing that could not be said before. The doc comments are unusually good — they name what must *not* read the field, which is the half that rots.

The problem is that the fix lands on **one** of the two code paths that build `SalesDocumentOrderFacts`, and the repo already treats that particular divergence as the defect class to avoid.

---

### 2. Findings

**[BLOCKING]** — `libs/core/src/orders/application/services/sales-document-view.service.ts:270-282` (`toOrderFacts`)

There are two builders of `SalesDocumentOrderFacts` in the tree:

- `toSalesDocumentOrderFacts` (invoicing) — the **gate**, from the live `Order`. Fixed by this PR.
- `SalesDocumentViewService.toOrderFacts` (orders) — the **operator-facing per-order projection** (ADR-065 / #2516), which feeds `ISalesDocumentRulesService.resolveRoutingBatch` from the denormalized `order_records` scalars. It reads `record.taxTreatment` — the column backfilled to `'exclusive'` for every PrestaShop order by `1841000000006-backfill-order-records-tax-treatment.ts`. There is no `totalTaxTreatment` column, and this PR adds none.

So after this change a PrestaShop order will be **routed** by an `orderTotalGross` rule (gate sees `inclusive`) while the surface that tells the operator *which document this order gets* still evaluates it as `net-priced` and reports `unresolved`. Reported ≠ enforced, on exactly the axis the feature exists to make legible.

That the two must agree is not my inference — it is stated in the file itself, three lines above the code in question (`sales-document-view.service.ts:255-259`): *"the same address `toSalesDocumentOrderFacts` uses, because a discovery or projection reading a different address would name a jurisdiction the evaluator never sees"*. The same argument binds `taxTreatment`. `order-record-repository.port.ts:82` makes the identical promise for `buyerTaxId` ("the SAME field `toSalesDocumentOrderFacts` builds the rule engine's facts from").

Fix: denormalize `totalTaxTreatment` onto `order_records` alongside the existing #1985 analytics scalars (`order-analytics-projection.ts:33` is the single write seam; `order-record.repository.ts:2038/2142` the persistence half), add the migration + a backfill for PrestaShop rows mirroring `1841000000006`, and have `toOrderFacts` read `record.totalTaxTreatment ?? record.taxTreatment`. Follow the migration rules in `docs/migrations.md` § Timestamp uniqueness invariant — synthetic sequential prefix, strictly greater than `origin/main`'s tail, class suffix matching, `up()`+`down()`.

If you'd rather not take the schema change in this PR, the alternative is to say so explicitly in the code (a comment on `toOrderFacts` naming the known divergence and its issue number) — but shipping the gate fix while the projection silently disagrees, with nothing recording it, is the state I'd block on.

---

**[IMPORTANT]** — `libs/core/src/orders/domain/order-from-ready-snapshot.ts:216-232` (`readTotals`)

`readTotals` is an allowlist and drops `totalTaxTreatment`, so any `Order` rehydrated from a persisted snapshot loses the assertion. This is exactly the failure the allowlist's own neighbouring comment (lines 195-205, #2054/#2254) documents: *"a field omitted here is silently dropped, and every MANUAL issuance path rehydrates through it… the auto-issue path — which composes from the live `Order` and never touches this function — was fine. Caught by a curl pass, not by a unit test, precisely because the two paths diverge here."*

No shipped rehydrating caller routes today (`invoicing.controller.ts:1271` and `fiscalization.controller.ts:283` issue against an explicitly-named connection; the fulfilment handlers read only the address), so this is latent rather than live. But it is a two-line fix guarding a defect the file has already been bitten by once:

```ts
if (raw.totalTaxTreatment === 'inclusive' || raw.totalTaxTreatment === 'exclusive') {
  totals.totalTaxTreatment = raw.totalTaxTreatment;
}
```

---

**[SUGGESTION]** — `libs/integrations/woocommerce/src/infrastructure/adapters/woocommerce-order-source.adapter.ts:271`

The PR description correctly identifies WooCommerce as structurally identical and defers it. Please file the follow-up issue and reference it from the `OrderTotals.totalTaxTreatment` doc comment — the field's docblock is where the next person looks, and "one source sets it, one identically-shaped source does not" reads as an oversight rather than a scoped deferral without a number attached.

**[SUGGESTION]** — `order-to-sales-document-order-facts.mapper.spec.ts`

Good pair of cases (override + fallback). Worth a third asserting `totalTaxTreatment: 'exclusive'` alongside `taxTreatment: 'inclusive'` — the fallback is `??`, so the "explicitly overrides *downward*" direction is currently untested and a future refactor to `||` or a truthiness check would pass the two existing cases.

---

### 3. Positive observations

- Correctly refused the tempting one-line flag flip, and the PR body documents the audit that justified the refusal — that reasoning is the valuable half of this change.
- `?? taxTreatment` default means every non-PrestaShop source is byte-identical; no behaviour change to prove absent.
- Explicit "never read by `convertGrossToNet` or the net-sales tax-rate path" in the type doc is the right kind of comment — it constrains future readers rather than restating the code.
- Types stay in `*.types.ts`, additive optional fields, no `any`, no layer violation. Naming and placement are clean.

---

### 4. Final Assessment

**Merge readiness: ❌ Request changes.**

The core change is correct and well-argued; it is simply incomplete. The gate now trusts a PrestaShop total as gross while the operator-facing projection built on the persisted scalars still does not, and the repo's own comments in both touched neighbourhoods identify that gate-vs-projection split as the recurring defect.

**Priority fixes**
1. Close the `SalesDocumentViewService.toOrderFacts` divergence — denormalize `totalTaxTreatment` onto `order_records` (+ migration + PrestaShop backfill), or record the known divergence in code with an issue number.
2. Add `totalTaxTreatment` to the `readTotals` allowlist in `order-from-ready-snapshot.ts`.
3. File the WooCommerce follow-up and cite it from the field's docblock.
4. (Optional) Add the `totalTaxTreatment: 'exclusive'` over `taxTreatment: 'inclusive'` spec case.

Also note the PR is `mergeable_state: behind` — rebase on `main` before merging.

norbert-kulus-blockydevs:
Addressed Piotr's review (norbert's earlier review had no blocking findings). Pushed `5c340f0e1` (rebased on `main`).

**[BLOCKING] Gate-vs-projection divergence** — fixed. Denormalized `totalTaxTreatment` onto `order_records` (migration `1870000000000-add-order-record-total-tax-treatment.ts`, mirroring `1841000000006`'s PrestaShop-scoped backfill join). `SalesDocumentViewService.toOrderFacts` now reads `record.totalTaxTreatment ?? record.taxTreatment` — the exact same fallback `toSalesDocumentOrderFacts` applies from the live `Order`, so the gate and the operator-facing per-order projection can no longer disagree about whether a PrestaShop order's total is gross. Full write/read path updated: `OrderRecord` entity (appended positionally, last), `OrderRecordOrmEntity` column, `OrderAnalyticsScalars` + `deriveOrderAnalyticsScalars`, both `order-record.repository.ts` write seams (`upsertWithLineItems`'s ready-path column list and its `toDomain` hydration), and every docblock enumerating "the four analytics scalars" updated to five.

**[IMPORTANT] `readTotals` allowlist** — fixed. Added the same two-line guard pattern already used for `taxTreatment` (and the per-line tax fields it's modeled on) to `order-from-ready-snapshot.ts`, so a rehydrated `Order` no longer silently drops `totalTaxTreatment`.

**[SUGGESTION] WooCommerce follow-up** — filed #2836, cited from `OrderTotals.totalTaxTreatment`'s doc comment.

**[SUGGESTION] Third spec case** — added `should honour totalTaxTreatment even when it disagrees downward from taxTreatment` to `order-to-sales-document-order-facts.mapper.spec.ts`, asserting the `??`-fallback direction a `||`/truthiness regression wouldn't catch.

Also added two new cases to `sales-document-view.service.spec.ts` (override + fallback in the projection, mirroring the gate's spec) and extended `order-record.repository.spec.ts` to assert the new column is written on the ready path (`upsertWithLineItems`) and absent from the analytics-free `upsert()` path.

Verified: `pnpm --filter @openlinker/core type-check`, `pnpm --filter @openlinker/api type-check`, `pnpm check:invariants`, and `pnpm lint` (full monorepo, background run) all pass. Ran the full affected-consumer test suites again post-fix (`order-record.repository.spec.ts`, `sales-document-view.service.spec.ts`, `order-to-sales-document-order-facts.mapper.spec.ts`, `order-from-ready-snapshot.spec.ts`, `order-record.service.spec.ts`, `prestashop-order.mapper.spec.ts`) — all green, migration timestamp validated against `origin/main`'s current tail (`1869000000900`).

piotrswierzy (approved):
## Re-review — #2832 @ `890e44c` (post force-push)

Scope now 17 files, +330/−26 (was 6 files, +99/−5).

**[BLOCKING] `SalesDocumentViewService.toOrderFacts` built the facts from `record.taxTreatment` — RESOLVED, and resolved the way I asked rather than by a comment.**

The full denormalization landed, and I checked every link of the chain rather than the diff's headline:

- `order_records.totalTaxTreatment` column: ORM entity `@Column({ type: 'varchar', nullable: true })` — byte-identical to the `taxTreatment` column two fields up (`order-record.orm-entity.ts:236`), so no type divergence between the sibling scalars.
- Migration `1870000000000-add-order-record-total-tax-treatment.ts`: prefix is strictly greater than `origin/main`'s tail (`1869000000900-add-automation-run-retry-attempt.ts` — confirmed against `git ls-tree origin/main`), class suffix matches the filename, `up()` + `down()` both present, `IF NOT EXISTS`/`IF EXISTS` guards. The backfill is a verbatim structural copy of `1841000000006-backfill-order-records-tax-treatment.ts` — same `FROM "connections" c ... c."platformType"` join, same `IS NULL` idempotence guard. Correct precedent, correctly followed.
- `deriveOrderAnalyticsScalars` (the single #1985 write seam) now derives the fifth scalar; `upsertWithLineItems` stamps it on the entity beside the other four rather than in the shared `toOrm`, which is the right side of that split — mapping it in `toOrm` would have let `persistIncomingSnapshot` NULL a ready-path value, exactly what the surrounding comments warn about.
- `buildFrozenAttributionUpsert`'s `includeReadyPathColumns` branch adds it, and — the part that matters — `fromRawRow` derives its reset from `getMetadataArgsStorage().filterColumns(OrderRecordOrmEntity)`, so the new column is automatically in the reset derivation with no hand-kept list to forget. The three `expectColumnAbsentFromUpsert('totalTaxTreatment')` spec additions pin the ready-path-only property.
- `toDomain` passes it as the new trailing positional constructor arg; `OrderRecord`'s param is last with a `= null` default, so no construction site shifts (the entity's own "append-only positional constructor" comment is honoured).
- `findByIds` — the read `SalesDocumentViewService` uses — is an unprojected `repository.find()`, so the column is genuinely populated at the projection, not silently absent.
- `toOrderFacts` now reads `record.totalTaxTreatment ?? record.taxTreatment` and still guards with `isTaxTreatment`, so a junk varchar degrades to "field omitted" rather than reaching the evaluator. Two new specs cover the override and the fallback.

The gate and the projection now compute the identical expression from the identical fallback, which is what the divergence argument in that file's own country-address comment requires.

**[IMPORTANT] `readTotals` allowlist dropped the field — RESOLVED.** `order-from-ready-snapshot.ts` now allowlists `totalTaxTreatment` with the same literal-pair check as `taxTreatment`, and the comment names the failure mode rather than restating the code.

**[SUGGESTION] WooCommerce follow-up — RESOLVED.** #2836 is filed and cited from both `OrderTotals.totalTaxTreatment`'s docblock and the migration header, which also states explicitly why the backfill is PrestaShop-only where `1841000000006` covered both platforms (an unasserted platform stays NULL = "same as `taxTreatment`" = existing behaviour, never a guess). That is the right shape: the narrower backfill is now a recorded decision instead of an asymmetry someone has to reverse-engineer.

**[SUGGESTION] Downward-override spec — RESOLVED.** The third case asserts `totalTaxTreatment: 'exclusive'` over `taxTreatment: 'inclusive'`, and its comment states precisely which refactor (`??` → `||`) it exists to catch.

All four addressed. Nothing was closed by comment where I asked for code.

### 2. Review of the new material

**No BLOCKING or IMPORTANT findings.**

**[SUGGESTION]** `order-record.repository.ts:2359` — `(entity.totalTaxTreatment as PriceTaxTreatment | null) ?? null` is an unvalidated cast off a `varchar` column. It is the exact precedent set two lines up for `taxTreatment` (`:2317`), so consistency is the right call here and I am not asking for a change; noting only that both consequently rely on `isTaxTreatment` at the read site, which `toOrderFacts` does apply. A shared coercion for the pair would be a tidy-up for whoever touches #2836.

**[SUGGESTION]** The migration's `down()` drops the column, while `1841000000006`'s `down()` is a deliberate no-op with a reasoned comment. The asymmetry is correct — that migration only *set* a pre-existing column, this one *adds* one, so dropping is the only sound reversal — but a half-sentence saying so would pre-empt the "why does the sibling refuse to revert?" question.

**Positive observations**
- The comment discipline held across a 3× larger diff: every new comment says what must not be done and why, and the migration header reconstructs the whole `taxTreatment`-cannot-be-flipped argument at the place a DBA will actually read it.
- Doc-comment arithmetic ("four scalars" → "five") was updated at all seven sites in the repository, including the two derived-reset comments where a stale count would have been actively misleading.
- No `any`, no layer violation, types stay in `*.types.ts`, additive-optional throughout, `?? taxTreatment` keeps every non-PrestaShop source byte-identical.

### 3. Final assessment

The gate/projection divergence is closed with real persistence rather than a TODO, the migration follows the house rules and the nearest precedent exactly, and the derived-reset machinery picks the new column up without a hand-kept list. Two housekeeping items before merge, neither a code concern:

1. `mergeable_state: behind` — rebase on `main`.
2. `Lint` and `Test` were still `queued` at review time (Build, Type Check, Integration Tests all green). Merge on green, not on this approval.

norbert-kulus-blockydevs:
Thanks for the re-review — the PR is already approved, so the only open items were the two SUGGESTIONs from the second pass. Addressed the actionable one:

- Added a comment on the migration's `down()` explaining why it drops the column (this migration *adds* it, unlike `1841000000006`'s deliberate no-op, which only *set* a pre-existing column) — commit 3295640.

The other suggestion (a shared coercion helper for `taxTreatment`/`totalTaxTreatment` at `order-record.repository.ts:2359`) was explicitly marked as a tidy-up for whoever picks up #2836, not a requested change, so left as-is per your note.

CI was `queued` for Lint/Test at review time — will merge once green.

piotrswierzy (commented):
## Re-review — #2832 @ `3295640`

Head moved from `890e44c` to `3295640` by exactly one commit, `3295640 docs(migrations): explain why the total-tax-treatment down() drops the column` — a 3-line comment addition to the migration and nothing else (`git diff 890e44c 3295640` is one file, +3/−0). The code approved last round is unchanged, still 17 files / +333 −26 against `main`.

**My code-level approval from `890e44c` stands.** This is a COMMENT rather than a second APPROVE for one reason only: CI is now red, and it was `queued` when I approved. Details in § 4.

All four items from the first (CHANGES_REQUESTED) review were verified resolved at `890e44c`; I re-confirmed each against the current tree rather than trusting the earlier text:

- **[BLOCKING] gate-vs-projection split — RESOLVED.** `order_records.totalTaxTreatment` column, ORM entity, migration `1870000000000`, threaded `deriveOrderAnalyticsScalars` → `upsertWithLineItems` → `toDomain`, and `SalesDocumentViewService.toOrderFacts` reading `record.totalTaxTreatment ?? record.taxTreatment` behind `isTaxTreatment` — the same expression and the same fallback the live-`Order` gate applies in `toSalesDocumentOrderFacts`. Intact.
- **[IMPORTANT] `readTotals` allowlist — RESOLVED.** `order-from-ready-snapshot.ts` still allowlists `totalTaxTreatment` with the literal-pair check.
- **[SUGGESTION] WooCommerce follow-up — RESOLVED.** #2836 filed and cited from both the field docblock and the migration header. See § 3 for a cross-PR consequence that is *not* this PR's to fix.
- **[SUGGESTION] downward-override spec — RESOLVED.**

And the one suggestion from the second review is now closed by `3295640`: the `down()` comment states why this migration drops the column where `1841000000006`'s `down()` is a deliberate no-op ("that migration only SET a pre-existing column, this one ADDS the column"). Exactly the half-sentence I asked for, in the right place.

### 2. Review of the new material

The three added lines are a comment. No correctness, layering, typing or test surface is touched. Nothing to find.

### 3. Migration ordering — re-checked against the *current* `origin/main`

`origin/main` has advanced 8 commits since this PR's recorded base (`5abcc57` → `58d9b7e`). Its migration tail is unchanged at `1869000000900-add-automation-run-retry-attempt.ts`, so `1870000000000` is still strictly greater and still unique — no collision introduced by the eight intervening commits, and no renumber needed. Class suffix matches the filename, `up()`/`down()` both present with `IF NOT EXISTS`/`IF EXISTS` guards, backfill idempotent on `WHERE "totalTaxTreatment" IS NULL`. Still conformant to `docs/migrations.md` § Timestamp uniqueness invariant.

**Cross-PR note (not a change request on this PR).** In the last round I asked that whichever of this PR and #2838 lands *second* either add the WooCommerce arm to this backfill `UPDATE` or amend the docblock — otherwise #2836 closes having changed no history. **This PR took the docblock option and took it well**: the header states the WooCommerce exclusion is deliberate, cites #2836, and explains that an unasserted platform staying `NULL` means "same as `taxTreatment`", i.e. existing behaviour rather than a guess. So this side is discharged. But I then checked #2838, and it ships **no migration at all** (2 files: the WooCommerce adapter + its spec). As it stands, #2836 will fix ingestion from that point forward while every historical WooCommerce order keeps `totalTaxTreatment IS NULL` and therefore keeps failing an `orderTotalGross` rule — the exact asymmetry the docblock's "tracked as a follow-up" implies will be closed. That belongs on **#2838**, not here; I'll raise it there. Nothing for this PR to change.

### 4. CI — the one thing blocking merge

On `3295640` (run `33792088037`), 7 of 9 checks are green (Build, Lint, Type Check, Docker Build Smoke Test, PHP Unit Tests, Scaffolded Adapter Builds, Detect changed paths) but **two are failing**:

- **Integration Tests — failure.** Very likely explained, not excused: this branch's base is `5abcc57`, which is **8 commits behind `origin/main` and does not contain `58d9b7e` ("fix(ci): close both integration-test flakes — Testcontainers network prune and the live SyncJobRunner", #2828)**. I confirmed with `git merge-base --is-ancestor 58d9b7e pr2832` → not an ancestor. #2824, which does carry that fix, has Integration Tests green on the same infrastructure. Rebasing should clear this.
- **Test (unit) — failure.** This one I cannot account for, and it is why I am not re-approving. It was `queued` at my last review, so it has never been observed green on this branch. Job logs are not readable without a token, so I will not guess which suite broke, but I did check the two hypotheses I could check from the tree:
  - *Semantic conflict with `main`.* The only files this PR and the intervening `main` commits both touch are `order-record.repository.ts` and its spec — `main` changed them via `0470542e0` (#2761), which adds a `sales_document_issued_on_request` aggregate to `getSummary`. That is disjoint from this PR's `totalTaxTreatment` column plumbing, `git merge-tree` reports a clean merge, and I could not construct a way for it to break this PR's specs. Plausible but unconfirmed.
  - *The known `apps/web` full-suite parallelism flake.* Also plausible, also unconfirmed.

**Requested action, and it is only this:** rebase on current `origin/main` and re-run. If Integration Tests goes green and `Test` stays red, it is a real failure on this branch and I'd want to see the failing suite before merge; if both go green, merge on that.

### 5. Final assessment

**Code: ✅ approved (approval from `890e44c` stands, plus the `down()` comment now closes the last suggestion). Merge readiness: ⏸ blocked on CI.**

The change itself is unchanged and I have no new findings against it. Do not merge on the standing approval while `Test` and `Integration Tests` are red — the integration failure has a strong, checkable explanation (missing #2828), the unit failure does not yet, and "approved code" is not the same claim as "this branch is green".

norbert-kulus-blockydevs:
Rebased onto current `main` (now includes #2828's Integration Tests fix). No code changes beyond the rebase — Piotr's code-level approval from `890e44c` stands and all four blocking/important findings plus the `down()` comment suggestion were already resolved in prior commits. Waiting on CI to go green before squash-merge.

piotrswierzy (commented):
## Re-review — #2832 @ `5686d31` (post force-push)

Head moved `3295640` → `5686d31`. `git log df46c641..5686d31` shows the **same four commit subjects** as before; the force-push is a **rebase onto current `origin/main` (`df46c641`)**, not a content change. `git diff 3295640 5686d31` is entirely the 8→9 commits `main` advanced by (`.eslintrc.js`, the Wave-3b mockups, #2824's Allegro/Woo/Erli buyer-tax-id work, …) — nothing of this PR's own. Against `main` the diff is still 17 files, **+333 / −26**.

**My code-level approval from `890e44c` stands unchanged.** This is a COMMENT rather than a re-APPROVE for one reason, unchanged from last round: **the branch has never been observed green**, and the fresh run is still queued. Details in § 4.

---

Re-checked each against the *current* tree (`git show 5686d31:<path>`), not against my earlier text.

- **[BLOCKING] gate-vs-projection `totalTaxTreatment` split — RESOLVED, intact.** Both builders of `SalesDocumentOrderFacts` now compute the identical expression: `toSalesDocumentOrderFacts` (`invoicing/application/mappers/…:66`) reads `order.totals.totalTaxTreatment ?? order.totals.taxTreatment`, and `SalesDocumentViewService.toOrderFacts` reads `record.totalTaxTreatment ?? record.taxTreatment`, still behind `isTaxTreatment` so a junk varchar degrades to "field omitted" rather than reaching the evaluator. I traced the persistence chain again end to end and it is unbroken: ORM column (`varchar`, `nullable`) → `deriveOrderAnalyticsScalars` (the single #1985 write seam, now five scalars) → `OrderRecordService.persistOrder` passing it as the trailing constructor arg → `upsertWithLineItems`'s `includeReadyPathColumns` branch → `toDomain`. `fromRawRow` still derives its reset from `getMetadataArgsStorage().filterColumns(...)`, so the new column joins the reset with no hand-kept list, and the three `expectColumnAbsentFromUpsert('totalTaxTreatment')` additions pin the ready-path-only property.
- **[IMPORTANT] `readTotals` allowlist — RESOLVED.** `order-from-ready-snapshot.ts:229-235` allowlists it with the same literal-pair check, and the comment names the failure mode (silent drop on every manual-issuance rehydration path) rather than restating the code.
- **[SUGGESTION] WooCommerce follow-up — RESOLVED.** #2836 filed, cited from both `OrderTotals.totalTaxTreatment`'s docblock and the migration header.
- **[SUGGESTION] downward-override spec — RESOLVED.**
- **[SUGGESTION from round 2] `down()` asymmetry comment — RESOLVED** by `5686d31e6`'s predecessor commit; the header now states why this migration drops the column where `1841000000006`'s `down()` is a deliberate no-op.

Nothing was closed by comment where I asked for code.

---

### 2. Migration ordering — re-checked against the rebased base

`origin/main`'s migration tail is `1869000000900-add-automation-run-retry-attempt.ts`. `1870000000000` is still strictly greater and unique; class suffix matches the filename; `up()` + `down()` both present with `IF NOT EXISTS` / `IF EXISTS` guards; the backfill is idempotent on `WHERE "totalTaxTreatment" IS NULL` and is a verbatim structural copy of `1841000000006`'s platform-scoped `FROM "connections" c` join. Conformant to `docs/migrations.md` § Timestamp uniqueness invariant. Because the branch is now rebased, the base is `df46c641` itself — no drift to re-verify later.

---

### 3. THE CROSS-PR BACKFILL QUESTION — resolved, and the answer is *yes, but not here*

You asked whether #2836 would close having changed no history. It will not, and the two PRs now **agree**:

- **This PR (`5686d31`) still ships the PrestaShop-only arm** (`AND c."platformType" = 'prestashop'`) and a docblock that says so deliberately, citing #2836 as the follow-up. Unchanged from `3295640`.
- **#2838's new commit `0794274` fulfils it by editing *this very migration*** rather than adding its own: `platformType IN ('prestashop', 'woocommerce')`, with the docblock rewritten to describe both platforms and the "deliberately does NOT touch WooCommerce" paragraph removed. So there is no second migration, no timestamp to check, and the idempotence guard is untouched.

That is a coherent resolution *while both PRs are unmerged*, and I have no change to request here. But it creates one hazard that belongs on the merge, not on either diff, so I am recording it on both PRs:

> **`docs/migrations.md` rule 4 is "Never edit executed migrations."** If **#2832 merges and `1870000000000` runs anywhere** — demo, staging, any environment deployed from `main` — before #2838 lands, TypeORM records it in `migrations` and **will never re-run it**. The WooCommerce arm then silently never executes, and every historic WooCommerce `order_records` row keeps `totalTaxTreatment IS NULL` → falls back to `'exclusive'` → still unmatched by an `orderTotalGross` rule. The behaviour is safe (NULL = "same as `taxTreatment`" = existing reading, never a guess), but #2836 would close having changed no history — the exact outcome the folded-in arm exists to prevent.
>
> Two ways out, either fine: **(a)** merge the two back to back and do not deploy in between, or **(b)** have #2838 ship its own new migration (`> 1870000000000`) carrying only the WooCommerce arm, which is immune to ordering. Given this repo deploys from `main`, (b) is the one that cannot go wrong. Nothing for *this* PR to change either way — flagged on #2838 as the remaining item there.

---

### 4. CI — still the only thing blocking merge

Run `33854096310` on `5686d31`, checked at ~09:05 UTC (≈30 min after start):

| Check | State |
|---|---|
| Type Check, Docker Build Smoke Test, PHP Unit Tests, Scaffolded Adapter Builds, Detect changed paths | ✅ success |
| **Build, Lint, Test, Integration Tests** | ⏳ **still `queued`** |

Two things follow.

- **The Integration Tests hypothesis is now testable and was correct to raise.** I confirmed `git merge-base --is-ancestor 58d9b7e 5686d31` → **YES**: the rebase brought in #2828 ("close both integration-test flakes"), which the previous base lacked. The prior failure had a checkable cause and that cause is gone. Not yet *proven* — the job has not run.
- **The unexplained unit `Test` failure has NOT cleared, because it has not re-run.** I want to be precise rather than optimistic here: I checked and eliminated one tempting explanation. #2838's author reports fixing a `order-analytics-projection.spec.ts` drift ("derives all 4 scalars" not expecting `totalTaxTreatment`) that was failing their suite — but `git show 3295640:libs/core/src/orders/domain/order-analytics-projection.spec.ts` shows **that fix was already present on this branch** at the red head. It was missing only on #2838's *stale* base. So it is not the cause of this branch's red `Test`, and that failure remains unaccounted for.

**Requested action, and it is only this:** wait for run `33854096310` to finish. If `Test` goes green, this is mergeable on my standing code approval. **If `Test` is red again, do not merge** — post the failing suite here; a unit failure that survived a rebase onto a clean `main` is a real failure on this branch, and "approved code" is not the same claim as "this branch is green."

---

### 5. Final assessment

**Code: ✅ approved (approval from `890e44c` stands; all four original findings plus both round-2 suggestions verified resolved against the current tree). Merge readiness: ⏸ blocked on CI.**

**Remaining items, in order:**
1. Wait for run `33854096310`. Merge only on green `Test` **and** green `Integration Tests`; surface the failing suite if `Test` is red again.
2. Coordinate the merge with #2838 per § 3 — either land them back to back with no deploy in between, or ask #2838 for its own migration.

piotrswierzy (approved):
## Re-review — #2832 @ `32c7172`

Head moved `5686d31` → `32c7172` by force-push. First thing I checked: **the force-push was a pure rebase onto current `main`.** `git diff 5686d31..32c7172` is exactly `+.gitignore`, `+docs/operations/perf-lab-stand.md`, `+perf/openlinker-throughput/bootstrap.sh` — i.e. #2872, picked up from `main` (`608813dc`). **Zero changes to this PR's own code.** So my code-level approval carries over verbatim, and I re-verified the substance below rather than re-litigating it.

---

### 1. Prior findings — all still resolved on this head

| Finding | Status on `32c7172` |
|---|---|
| **[BLOCKING]** gate-vs-projection divergence | ✅ `SalesDocumentViewService.toOrderFacts` reads `record.totalTaxTreatment ?? record.taxTreatment` — the **same** fallback `toSalesDocumentOrderFacts` applies from the live `Order`. The docblock states why the two must agree. |
| **[IMPORTANT]** `readTotals` allowlist | ✅ `order-from-ready-snapshot.ts` guards `totalTaxTreatment` with the identical two-line pattern used for `taxTreatment`, with a comment naming the failure mode ("would vanish on any manual-issuance path"). |
| **[SUGGESTION]** WooCommerce follow-up | ✅ #2836 filed, cited from `OrderTotals.totalTaxTreatment`, and now shipping as #2838. |
| **[SUGGESTION]** third spec case (`??` vs `||`) | ✅ present. |
| **[SUGGESTION]** `down()` comment | ✅ `32c7172`'s only prior-head commit. It draws the right distinction: this migration **adds** the column so dropping it is the sound reversal, *unlike* `1841000000006`'s deliberate no-op which only `SET` a pre-existing one. |

Full write/read path re-verified on this head: `OrderRecord` entity (appended **positionally last**, respecting the constructor-order warning in that file), `OrderRecordOrmEntity` (`varchar` nullable, matching the migration's `ADD COLUMN … varchar`), `OrderAnalyticsScalars` + `deriveOrderAnalyticsScalars` (4→5 scalars), and both `order-record.repository.ts` seams — added to the `includeReadyPathColumns` set so only `upsertWithLineItems` writes it, correctly *outside* the shared `toOrm` write set, with every "the four analytics scalars" docblock updated to five. `convertGrossToNet` and the ADR-063 `deriveNetLineAmount` path still read `taxTreatment` and are untouched.

---

### 2. THE CROSS-PR MIGRATION ANSWER (stated identically on #2838)

Last round the answer was "#2838 fulfils the WooCommerce backfill by **editing this PR's migration**", which I blocked under `docs/migrations.md` rule 4 ("Never edit executed migrations", line 367): if `1870000000000` ran anywhere before #2838 landed, TypeORM would never re-run it, the WooCommerce arm would silently never execute, and #2836 would close having changed no history.

**That is fixed, on #2838's side, the way I recommended.** #2838 `618125d` reverts the in-place edit and ships a standalone `1870000000100-backfill-order-record-total-tax-treatment-woocommerce.ts`. I verified `git diff 32c7172..618125d -- apps/api/src/migrations/1870000000000-…ts` is **empty** — your migration is byte-identical on both branches and carries the PrestaShop arm only, exactly as its docblock says.

Ordering and idempotence check across the pair:

```
origin/main tail   1869000000900-add-automation-run-retry-attempt.ts
this PR            1870000000000-add-order-record-total-tax-treatment.ts   (prestashop)
#2838              1870000000100-…-woocommerce.ts                           (woocommerce)
```
strictly ascending per `docs/migrations.md` rule 3, and both `UPDATE`s are guarded by `AND o."totalTaxTreatment" IS NULL`.

**What this means for you: nothing.** You may merge this PR whenever CI allows, with any number of deploys before #2838 lands. The backfill is now order-independent — there is no longer any "merge these two back to back" process guarantee to honour. The `Deliberately does NOT touch WooCommerce … tracked as a follow-up (#2836)` paragraph in your docblock remains accurate, because #2836 now genuinely ships its own migration.

---

### 3. CI — plainly

**The `Test` job has still not run on this head.** As of this review, run `33867233407` (queued 11:17 UTC) reports:

- **Test — `queued`** (never started)
- Lint — `queued`, Build — `queued`, Integration Tests — `queued`
- Type Check ✅ success, Docker Build Smoke Test ✅ success, PHP Unit Tests ✅ success, Scaffolded Adapter Builds ✅ success, Detect changed paths ✅ success

So the answer to "has `Test` now run green on this branch" is **no — it is sitting in the runner queue, not failing.** Four heavier jobs are backed up behind it; on #2838 the same queue took ~2h wall for Integration Tests, so this looks like runner contention, not a problem with the branch.

**However, there is now strong independent evidence the unit suite passes on this exact code.** #2838's tree is identical to this one across every file this PR touches — `git diff 32c7172..618125d -- libs/core/src/orders/ libs/integrations/prestashop/` returns **one docstring line** in `order.types.ts` and nothing else. On #2838's head `618125d`, **Test ✅ success (17 min)** and **Integration Tests ✅ success (32 min)**. That effectively retires the earlier unexplained `Test` failure: the same code, plus the WooCommerce delta, is green — the earlier red was not reproducible.

I am therefore approving on the code and on that evidence. **Do not merge until this branch's own `Test` and `Integration Tests` land green** — the evidence above is strong but it is not this branch's own run against current `main`.

---

- The design decision is the right one and the docblocks make it durable: `taxTreatment` stays source-uniform (it governs `convertGrossToNet` and the ADR-063 net-sales path), and `totalTaxTreatment` is the narrower `total`-only signal. Absent means "same as `taxTreatment`", so every non-PrestaShop source behaves byte-identically to before.
- Migration conforms to `docs/migrations.md`: synthetic sequential prefix, filename/class match, `IF NOT EXISTS`/`IF EXISTS`, idempotent platform-scoped backfill following the `1841000000006` → `1840000000000` precedent chain.
- Layering intact: the platform fact is declared in `PrestashopOrderMapper` where the platform is known; core only reads the neutral field. No new `any`, no cross-context boundary change.
- Comment discipline is unusually good throughout — each added comment states a constraint on *future* edits (which readers must use the fallback, which must not) rather than restating the diff.

Nothing new found on this head. The one stale-test-name nit from the first pass (`'should carry taxTreatment verbatim…'`) remains non-blocking.

---

### 5. Final assessment

**Summary.** A pure rebase onto `main`; no code moved. All four original blocking/important findings and both round-2 suggestions remain resolved and were re-verified against this head. The cross-PR migration hazard I blocked on has been resolved on #2838 by a standalone `1870000000100` migration, so this PR is now free of any merge-ordering dependency.

**Merge readiness: ✅ Approve — merge on green.**

**Before merging:**
1. Wait for this branch's own **Test**, **Lint**, **Build** and **Integration Tests** to complete (currently queued). If `Test` fails, ping me — but given #2838's identical tree passed it, I expect green.
2. No coordination with #2838 needed any more. Merge this whenever you like; #2838 rebases onto `main` afterwards.

norbert-kulus-blockydevs:
Rebased onto current \`main\` again — a real merge conflict appeared this time in \`prestashop-order.mapper.ts\` (both this branch and #2884, which landed in between, added a doc comment on the same \`taxTreatment: 'exclusive'\` line, for different reasons). Resolved by keeping both comments (the #2829 \`totalTaxTreatment\` rationale and the #2835 net-priced-lines rationale) — no code change, comment-only merge. No new code changes otherwise; the four prior commits are unchanged. Waiting on CI to go green before squash-merge.

PR #2837, opened 2026-09-03
fix(invoicing,fiscalization): name the actual constraint when refusing a net-priced order

### LINKED ISSUE #2835: [BUG] Invoicing & Fiscalization — net-priced-order guard blocks issuance for any PrestaShop/WooCommerce order, not just rule matching
## Problem / Context

Related to, but a broader manifestation than, #2829 (still open, PR #2832) — that issue is scoped to the sales-document **rule-matching condition** (`orderTotalGross`) refusing to evaluate a PrestaShop-sourced order. This issue is about the **document-issuance write paths themselves**, which turn out to refuse the order outright, independent of any rule.

Both `libs/core/src/invoicing/application/mappers/order-to-issue-invoice-command.mapper.ts:97-101` and `libs/core/src/fiscalization/application/mappers/order-to-register-transaction-command.mapper.ts:100-107` carry an unconditional guard:

```ts
if (order.totals.taxTreatment === 'exclusive') {
  throw new Unsupported{Fiscal}PriceTreatmentError(
    `Order ${order.id} is net-priced (taxTreatment "exclusive"); only gross-priced orders are supported`,
  );
}
```

This predates both today's PRs (from #2014, comment "GROSS-only MVP") — it is not a regression, just never verified end-to-end against a real PrestaShop/WooCommerce order until today.

**Both local shop order sources hardcode `taxTreatment: 'exclusive'` unconditionally:**
- `libs/integrations/prestashop/src/infrastructure/mappers/prestashop-order.mapper.ts:93`
- `libs/integrations/woocommerce/src/infrastructure/adapters/woocommerce-order-source.adapter.ts:271`

**Only the Allegro order source reports `'inclusive'`** (`libs/integrations/allegro/src/infrastructure/adapters/allegro-order-source.adapter.ts:820`).

### Reproduced live, twice, on `ol-demo-fresh`

1. Placed a real PrestaShop order (`ol_order_3eeb18f7b8054d8fa8aea7354c1af28d`, PL, no buyer tax ID). Tried to register a fiscal receipt for it via `POST /v1/fiscal-registrations`:
   ```
   422 { "message": "Order ol_order_3eeb18f7b8054d8fa8aea7354c1af28d is net-priced (taxTreatment \"exclusive\"); only gross-priced orders can be registered" }
   ```
2. Placed a second real PrestaShop order (`ol_order_768cb3fc9f87408b89a9358a300c7a1e`, FR, correctly routed to `invoice` via the `★ Rest of world` default). Tried to issue an invoice for it via `POST /v1/invoices`:
   ```
   400 { "message": "Order ol_order_768cb3fc9f87408b89a9358a300c7a1e is net-priced (taxTreatment \"exclusive\"); only gross-priced orders are supported" }
   ```

### The compounding fact that makes this worth fixing rather than leaving as "PrestaShop is just unsupported"

On the current `ol-demo-fresh` stack, **no real order can ever produce an invoice**, for a structural reason, not a configuration gap:
- Every real Allegro order (the only gross-priced source) ships to Poland only, and Poland's configured default routes to `fiscal-receipt`.
- The only sources that produce non-Polish orders (PrestaShop, WooCommerce) are both net-priced and therefore refused by the guard above regardless of what the routing engine decides.

So the invoice path is provably unreachable end-to-end today, not merely untested. (Fiscal-receipt registration for a real Allegro order **does** work — confirmed live, registration `123d35e0-376d-4a35-9e88-941f84f30305`, eparagony document #210.)

## Proposed Solution

This is a design decision, not a one-line fix, so this issue is scoped to **surfacing and deciding**, not shipping a fix:

1. Once #2829/#2832 lands (PrestaShop's `taxTreatment` reported per-order rather than hardcoded, presumably deriving it correctly from `total_paid_tax_incl` vs the line-level net prices), re-verify whether these two issuance-path guards should be relaxed the same way, or whether they represent a genuinely separate constraint (e.g. the invoice/fiscal document format itself requiring a gross total the adapter cannot safely reconstruct from a net snapshot without a rate).
2. If gross-conversion is safe once a real per-line tax rate is known (it already is stored — see `order_line_items.taxRate`, ADR-063), consider computing an effective gross total in the mapper rather than refusing outright, OR
3. If gross-conversion is judged unsafe for a legal document (a computed figure, not a source-reported one), keep the refusal but make it **actionable** — today's message ("only gross-priced orders are supported") gives an operator no path forward. State plainly that this order's *source* cannot supply an invoiceable/fiscalizable total and why.
4. Either way, do the same audit for WooCommerce once a decision is made — it carries the identical hardcoded `'exclusive'` and will hit the identical refusal the moment a real WooCommerce order tries to route to Invoicing/Fiscalization.

## Classification

**Type**: CORE
**Layer**: Application (mappers) / Integration (order-source adapters)
**File(s)**:
- `libs/core/src/fiscalization/application/mappers/order-to-register-transaction-command.mapper.ts`
- `libs/integrations/prestashop/src/infrastructure/mappers/prestashop-order.mapper.ts`

## Dependencies

- Depends on the direction taken by #2829 / PR #2832 (should land first — this issue re-derives its premise from whatever `taxTreatment` PrestaShop reports after that fix).

## Assumptions

- This issue does not assume a fix is safe or even desirable — decision 2 vs. decision 3 above is a real ADR-063-adjacent judgment call ("never guess on a figure feeding a legal document"), not an implementation detail. Whoever picks this up should re-read ADR-063 before choosing.
- WooCommerce is included by file reference only — no WooCommerce order has been pushed through either write path live; the claim that it will hit the same refusal is inferred from its identical hardcoded `taxTreatment`, not independently reproduced.

## Acceptance Criteria

- [x] A documented decision (ADR amendment, or a comment beside both guards) on whether a net-priced PrestaShop/WooCommerce order can ever be converted to a safe gross figure for invoicing/fiscalization, or whether it is a permanent limitation of those sources.
- [ ] ~~If relaxed: a real PrestaShop-sourced order can be invoiced and fiscally registered end-to-end, verified live (not just unit-tested).~~ N/A - not relaxed, see Decision below.
- [x] If kept as a permanent limitation: the refusal message names the actual constraint an operator can act on, not just "only gross-priced orders are supported."
- [x] Tests added or updated for whichever direction is chosen.
- [x] No architecture boundary violations (CORE ↔ Integration).

## Decision (resolved 2026-09-03)

**Kept as a permanent limitation, not relaxed.** ADR-063 § 5 is explicit that
core "may group and divide amounts; it may never compute tax." Converting a
net line price to gross is `net * (1 + rate)` - arithmetic that computes tax,
even though the rate itself is often already known (`OrderItem.taxRate`).
Knowing the rate does not make *computing* an amount with it any less
forbidden for a figure feeding a legal document, so decision 2 (compute an
effective gross total) was rejected in favor of decision 3.

#2829/#2832's `OrderTotals.totalTaxTreatment` does **not** relax this guard:
it describes `total` alone, while the invoice/fiscal write paths compose
`unitPriceGross` from each item's `price`, which stays net regardless of what
`total` is. Reading `totalTaxTreatment` here would let a PrestaShop order pass
the guard and then silently mislabel its still-net line prices as gross - the
exact corruption the guard exists to prevent, one level down. So the guard
correctly keeps reading `taxTreatment` (the line-level signal) only, and this
issue's fix did not need to stack on / wait for PR #2832.

This is a genuine, structural limitation of a net-line-price source
(PrestaShop, WooCommerce - both hardcode `taxTreatment: 'exclusive'`), not a
gap waiting to be closed. It clears only if the source adapter itself starts
reporting gross line prices (a platform-specific mapper change), never by
teaching core to convert.

The decision plus the full reasoning lives in a new shared pure helper,
`describeNetPricedOrderRefusal` (`libs/core/src/sales-documents/domain/types/gross-price-eligibility.types.ts`),
consumed by both mapper guards so the two document contexts state the
constraint identically. The refusal message now names the actual constraint
(e.g. `Order X cannot be invoiced: its source reports net (tax-exclusive) line
prices, and OpenLinker never computes or infers tax to convert them to gross
for a fiscal document - only a source that reports gross (tax-inclusive) line
prices can be invoiced today.`) instead of the unqualified "only gross-priced
orders are supported." Both the PrestaShop and WooCommerce order-source
mappers got an audit comment cross-referencing this.

norbert-kulus-blockydevs:
## Summary

Closes #2835.

The invoice and fiscal-registration write-path guards refused a net-priced order (PrestaShop, WooCommerce) with `only gross-priced orders are supported` - true, but not actionable, and easy to misread as fixed by #2829/#2832's new `totalTaxTreatment` signal.

They aren't. `totalTaxTreatment` describes the order **total** alone; `unitPriceGross` is composed from each item's `price`, which stays net regardless of what `total` is. Converting a net line price to gross would be *computing* tax (`net * (1 + rate)`), which ADR-063 forbids core from doing even though the rate is often already known - so this is decided as a genuine, permanent limitation of a net-line-price source, not a gap #2832 closes. Full reasoning + decision recorded on the issue and in the new helper's doc comment.

## Changes

- New shared pure helper `describeNetPricedOrderRefusal` (`libs/core/src/sales-documents/domain/types/gross-price-eligibility.types.ts`) - the single place both document contexts (invoicing, fiscalization) state the constraint identically.
- Both mapper guards (`order-to-issue-invoice-command.mapper.ts`, `order-to-register-transaction-command.mapper.ts`) now compose their thrown message from it, instead of the unqualified string.
- Both error-class docblocks cross-reference this decision so a future reader isn't misled by #2829/#2832 into thinking the guard should relax.
- Audit comments added at PrestaShop's and WooCommerce's `taxTreatment: 'exclusive'` sites, confirming both hit the identical permanent limitation (AC #4).
- Tests: new spec for the helper, plus a message-content assertion added to both existing mapper specs.

## Test plan

- [x] `pnpm --filter @openlinker/core lint` - 0 errors
- [x] `pnpm --filter @openlinker/core type-check` - clean
- [x] `pnpm --filter @openlinker/integrations-prestashop lint && type-check` - clean
- [x] `pnpm --filter @openlinker/integrations-woocommerce lint && type-check` - clean
- [ ] `pnpm test` - not run locally (frozen this machine before); covered by CI

### DIFF
--- a/libs/core/src/fiscalization/application/mappers/errors/unsupported-fiscal-price-treatment.error.ts
+++ b/libs/core/src/fiscalization/application/mappers/errors/unsupported-fiscal-price-treatment.error.ts
@@ -9,6 +9,10 @@
  * silently mis-state the registered sale, which is why this fails loud instead
  * of degrading.
  *
+ * This is a PERMANENT limitation of a net-line-price source (PrestaShop,
+ * WooCommerce), not a gap #2829/#2832's `totalTaxTreatment` signal relaxes -
+ * see `describeNetPricedOrderRefusal` (#2835) for why.
+ *
  * @module libs/core/src/fiscalization/application/mappers/errors
  */
 export class UnsupportedFiscalPriceTreatmentError extends Error {
--- a/libs/core/src/fiscalization/application/mappers/order-to-register-transaction-command.mapper.spec.ts
+++ b/libs/core/src/fiscalization/application/mappers/order-to-register-transaction-command.mapper.spec.ts
@@ -151,6 +151,27 @@ describe('toRegisterTransactionCommand', () => {
     ).toThrow(UnsupportedFiscalPriceTreatmentError);
   });
 
+  it('net-priced order: the refusal names the actual constraint (#2835)', () => {
+    expect(() =>
+      toRegisterTransactionCommand({
+        order: order({
+          totals: {
+            subtotal: 40,
+            tax: 9.2,
+            shipping: 0,
+            total: 40,
+            currency: 'PLN',
+            taxTreatment: 'exclusive',
+          },
+        }),
+        connectionId: 'conn-1',
+        idempotencyKey: 'k',
+      }),
+    ).toThrow(
+      /ol_order_1 cannot be fiscally registered: its source reports net \(tax-exclusive\) line prices/,
+    );
+  });
+
   it('should accept an order whose treatment is absent (the documented gross assumption)', () => {
     expect(() =>
       toRegisterTransactionCommand({
--- a/libs/core/src/fiscalization/application/mappers/order-to-register-transaction-command.mapper.ts
+++ b/libs/core/src/fiscalization/application/mappers/order-to-register-transaction-command.mapper.ts
@@ -13,7 +13,11 @@
  * @module libs/core/src/fiscalization/application/mappers
  */
 import type { Order, OrderItem } from '@openlinker/core/orders';
-import { minorUnitExponentFor, splitShippingAcrossRates } from '@openlinker/core/sales-documents';
+import {
+  describeNetPricedOrderRefusal,
+  minorUnitExponentFor,
+  splitShippingAcrossRates,
+} from '@openlinker/core/sales-documents';
 
 import type {
   FiscalRecipient,
@@ -101,11 +105,13 @@ export function toRegisterTransactionCommand(
   // and core may not convert (it never computes or defaults a tax rate). Fail
   // loud. An ABSENT treatment is the documented gross assumption - marketplaces
   // report buyer-paid gross - so it is accepted.
-  if (order.totals.taxTreatment === 'exclusive') {
-    throw new UnsupportedFiscalPriceTreatmentError(
-      `Order ${order.id} is net-priced (taxTreatment "exclusive"); ` +
-        `only gross-priced orders can be registered`,
-    );
+  //
+  // This checks LINE-level `taxTreatment`, never `totals.totalTaxTreatment`
+  // (#2829/#2832) — see `describeNetPricedOrderRefusal`'s doc comment
+  // (#2835) for why the latter cannot relax this guard.
+  const refusal = describeNetPricedOrderRefusal(order, 'fiscally registered');
+  if (refusal !== null) {
+    throw new UnsupportedFiscalPriceTreatmentError(refusal);
   }
 
   const lines = order.items.map((item) => toFiscalLine(item, order.id));
--- a/libs/core/src/invoicing/application/mappers/errors/unsupported-price-treatment.error.ts
+++ b/libs/core/src/invoicing/application/mappers/errors/unsupported-price-treatment.error.ts
@@ -7,6 +7,10 @@
  * Rather than mislabel net as gross (silent totals corruption), the mapper fails
  * loud. Messages cite ONLY `order.id`. No country/document-type vocabulary.
  *
+ * This is a PERMANENT limitation of a net-line-price source (PrestaShop,
+ * WooCommerce), not a gap #2829/#2832's `totalTaxTreatment` signal relaxes —
+ * see `describeNetPricedOrderRefusal` (#2835) for why.
+ *
  * @module libs/core/src/invoicing/application/mappers/errors
  */
 export class UnsupportedPriceTreatmentError extends Error {
--- a/libs/core/src/invoicing/application/mappers/order-to-issue-invoice-command.mapper.spec.ts
+++ b/libs/core/src/invoicing/application/mappers/order-to-issue-invoice-command.mapper.spec.ts
@@ -42,7 +42,14 @@ function makeOrder(overrides: Partial<Order> = {}): Order {
     id: 'order-1',
     status: 'processing',
     items: [makeItem()],
-    totals: { subtotal: 100, tax: 23, shipping: 0, total: 123, currency: 'PLN', taxTreatment: 'inclusive' },
+    totals: {
+      subtotal: 100,
+      tax: 23,
+      shipping: 0,
+      total: 123,
+      currency: 'PLN',
+      taxTreatment: 'inclusive',
+    },
     billingAddress: makeAddress(),
     createdAt: new Date('2026-06-22T10:00:00.000Z'),
     updatedAt: new Date('2026-06-22T10:00:00.000Z'),
@@ -90,11 +97,25 @@ describe('toIssueInvoiceCommand', () => {
 
   it('multi-line: items -> lines, currency from totals.currency, name fallback to sku then productId', () => {
     const order = makeOrder({
-      totals: { subtotal: 0, tax: 0, shipping: 0, total: 0, currency: 'EUR', taxTreatment: 'inclusive' },
+      totals: {
+        subtotal: 0,
+        tax: 0,
+        shipping: 0,
+        total: 0,
+        currency: 'EUR',
+        taxTreatment: 'inclusive',
+      },
       items: [
         makeItem({ id: 'a', name: 'Named', price: 10, quantity: 2 }),
         makeItem({ id: 'b', name: undefined, sku: 'SKU-9', price: 20, quantity: 1 }),
-        makeItem({ id: 'c', name: undefined, sku: undefined, productId: 'PID-5', price: 30, quantity: 3 }),
+        makeItem({
+          id: 'c',
+          name: undefined,
+          sku: undefined,
+          productId: 'PID-5',
+          price: 30,
+          quantity: 3,
+        }),
       ],
     });
 
@@ -123,12 +144,15 @@ describe('toIssueInvoiceCommand', () => {
     ['zero', 0],
     ['negative', -2],
     ['NaN', Number.NaN],
-  ])('should throw InvalidInvoiceLineError for a %s item quantity (#1525 review)', (_label, quantity) => {
-    const order = makeOrder({ items: [makeItem({ quantity })] });
-    expect(() => toIssueInvoiceCommand({ order, connectionId: 'conn-1' })).toThrow(
-      InvalidInvoiceLineError,
-    );
-  });
+  ])(
+    'should throw InvalidInvoiceLineError for a %s item quantity (#1525 review)',
+    (_label, quantity) => {
+      const order = makeOrder({ items: [makeItem({ quantity })] });
+      expect(() => toIssueInvoiceCommand({ order, connectionId: 'conn-1' })).toThrow(
+        InvalidInvoiceLineError
+      );
+    }
+  );
 
   it('InvalidInvoiceLineError is PII-clean: cites only the order id', () => {
     const order = makeOrder({ items: [makeItem({ quantity: 0, name: 'SECRET_ITEM' })] });
@@ -169,11 +193,15 @@ describe('toIssueInvoiceCommand', () => {
 
   it('name derivation: firstName+lastName absent AND no company -> throws InvalidBuyerProfileError (no "undefined undefined")', () => {
     const order = makeOrder({
-      billingAddress: makeAddress({ firstName: undefined, lastName: undefined, company: undefined }),
+      billingAddress: makeAddress({
+        firstName: undefined,
+        lastName: undefined,
+        company: undefined,
+      }),
     });
 
     expect(() => toIssueInvoiceCommand({ order, connectionId: 'conn-1' })).toThrow(
-      InvalidBuyerProfileError,
+      InvalidBuyerProfileError
     );
   });
 
@@ -191,7 +219,11 @@ describe('toIssueInvoiceCommand', () => {
       billingAddress: makeAddress({ firstName: 'Jan', lastName: 'Kowalski', company: 'Big Corp' }),
     });
 
-    const cmd = toIssueInvoiceCommand({ order, connectionId: 'conn-1', buyerTaxId: { scheme: 'eu-vat', value: 'PL1' } });
+    const cmd = toIssueInvoiceCommand({
+      order,
+      connectionId: 'conn-1',
+      buyerTaxId: { scheme: 'eu-vat', value: 'PL1' },
+    });
     expect(cmd.buyer.name).toBe('Big Corp');
   });
 
@@ -199,7 +231,7 @@ describe('toIssueInvoiceCommand', () => {
     const order = makeOrder({ billingAddress: undefined, shippingAddress: undefined });
 
     expect(() => toIssueInvoiceCommand({ order, connectionId: 'conn-1' })).toThrow(
-      InvalidBuyerProfileError,
+      InvalidBuyerProfileError
     );
   });
 
@@ -227,7 +259,11 @@ describe('toIssueInvoiceCommand', () => {
       shippingAddress: undefined,
     });
     // Force the name-derivation failure path.
-    order.billingAddress = makeAddress({ firstName: undefined, lastName: undefined, address1: 'SecretStreet 42' });
+    order.billingAddress = makeAddress({
+      firstName: undefined,
+      lastName: undefined,
+      address1: 'SecretStreet 42',
+    });
 
     try {
       toIssueInvoiceCommand({ order, connectionId: 'conn-1' });
@@ -242,7 +278,14 @@ describe('toIssueInvoiceCommand', () => {
   it('price treatment: totals.taxTreatment "inclusive" -> unitPriceGross = item.price', () => {
     const order = makeOrder({
       items: [makeItem({ price: 49.99 })],
-      totals: { subtotal: 0, tax: 0, shipping: 0, total: 0, currency: 'PLN', taxTreatment: 'inclusive' },
+      totals: {
+        subtotal: 0,
+        tax: 0,
+        shipping: 0,
+        total: 0,
+        currency: 'PLN',
+        taxTreatment: 'inclusive',
+      },
     });
 
     const cmd = toIssueInvoiceCommand({ order, connectionId: 'conn-1' });
@@ -261,22 +304,36 @@ describe('toIssueInvoiceCommand', () => {
 
   it('price treatment: totals.taxTreatment "exclusive" -> throws UnsupportedPriceTreatmentError', () => {
     const order = makeOrder({
-      totals: { subtotal: 0, tax: 0, shipping: 0, total: 0, currency: 'PLN', taxTreatment: 'exclusive' },
+      totals: {
+        subtotal: 0,
+        tax: 0,
+        shipping: 0,
+        total: 0,
+        currency: 'PLN',
+        taxTreatment: 'exclusive',
+      },
     });
 
     expect(() => toIssueInvoiceCommand({ order, connectionId: 'conn-1' })).toThrow(
-      UnsupportedPriceTreatmentError,
+      UnsupportedPriceTreatmentError
     );
   });
 
   it('price treatment: "exclusive" -> the refusal names the actual constraint (#2835)', () => {
     const order = makeOrder({
       id: 'ol_order_abc',
-      totals: { subtotal: 0, tax: 0, shipping: 0, total: 0, currency: 'PLN', taxTreatment: 'exclusive' },
+      totals: {
+        subtotal: 0,
+        tax: 0,
+        shipping: 0,
+        total: 0,
+        currency: 'PLN',
+        taxTreatment: 'exclusive',
+      },
     });
 
     expect(() => toIssueInvoiceCommand({ order, connectionId: 'conn-1' })).toThrow(
-      /ol_order_abc cannot be invoiced: its source reports net \(tax-exclusive\) line prices/,
+      /ol_order_abc cannot be invoiced: its source reports net \(tax-exclusive\) line prices/
     );
   });
 
@@ -316,7 +373,14 @@ describe('toIssueInvoiceCommand', () => {
     // shipping line is still EMITTED (dropping it would understate the total)
     // and still carries an empty rate, and the gate refuses the whole document.
     const order = makeOrder({
-      totals: { subtotal: 100, tax: 0, shipping: 15, total: 115, currency: 'PLN', taxTreatment: 'inclusive' },
+      totals: {
+        subtotal: 100,
+        tax: 0,
+        shipping: 15,
+        total: 115,
+        currency: 'PLN',
+        taxTreatment: 'inclusive',
+      },
     });
 
     const cmd = toIssueInvoiceCommand({ order, connectionId: 'conn-1' });
@@ -328,10 +392,15 @@ describe('toIssueInvoiceCommand', () => {
 
   it('shipping: inherits the basket rate when every line carries the same one (#2248)', () => {
     const order = makeOrder({
-      items: [
-        { id: 'i1', productId: 'p1', quantity: 1, price: 100, taxRate: '23' },
-      ],
-      totals: { subtotal: 100, tax: 0, shipping: 15, total: 115, currency: 'PLN', taxTreatment: 'inclusive' },
+      items: [{ id: 'i1', productId: 'p1', quantity: 1, price: 100, taxRate: '23' }],
+      totals: {
+        subtotal: 100,
+        tax: 0,
+        shipping: 15,
+        total: 115,
+        currency: 'PLN',
+        taxTreatment: 'inclusive',
+      },
     });
 
     const cmd = toIssueInvoiceCommand({ order, connectionId: 'conn-1' });
@@ -349,7 +418,14 @@ describe('toIssueInvoiceCommand', () => {
   ])('shipping: totals.shipping %s -> no phantom shipping line (#1517)', (_label, shipping) => {
     const order = makeOrder({
       items: [makeItem({ price: 100, quantity: 1 })],
-      totals: { subtotal: 100, tax: 0, shipping, total: 100, currency: 'PLN', taxTreatment: 'inclusive' },
+      totals: {
+        subtotal: 100,
+        tax: 0,
+        shipping,
+        total: 100,
+        currency: 'PLN',
+        taxTreatment: 'inclusive',
+      },
     });
 
     const cmd = toIssueInvoiceCommand({ order, connectionId: 'conn-1' });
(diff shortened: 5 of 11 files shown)


norbert-kulus-blockydevs:
## Code Review: net-priced order guard message (#2837)

**Scope**: 11 files, +204/-13. Comment-only edits in 2 adapters, one new pure helper + its export, guard-composition change in 2 mapper functions, tests.

### Verified independently
- `pnpm --filter @openlinker/core lint` / `tsc --noEmit` — 0 errors (warnings pre-existing, unrelated to this diff)
- `pnpm --filter @openlinker/integrations-prestashop lint` / `@openlinker/integrations-woocommerce lint` — 0 errors
- `scripts/check-cross-context-imports.mjs` — passes
- `barrel-purity.spec.ts` (28 tests incl. the `sales-documents` zero-outbound-edge leaf test) — passes
- Targeted specs (`gross-price-eligibility`, both mapper specs) — 62/62 passing

### Architecture & Boundaries
✅ `describeNetPricedOrderRefusal` lives in `sales-documents/domain/types/`, zero imports — matches the leaf's existing precedent (`shipping-tax-split.types.ts`, `tax-rate-enforcement.types.ts`) and the `*.types.ts` pure-rule exception in `engineering-standards.md`.
✅ Both consumers import it via the top-level `@openlinker/core/sales-documents` barrel — no deep-path violations.

### Correctness
✅ Guard logic is behaviorally unchanged — still line-level `taxTreatment === 'exclusive'`, never `totalTaxTreatment`. Confirmed by reading both mapper diffs.

### Findings

**[SUGGESTION]** — `gross-price-eligibility.types.ts:59` — `action: string` is unconstrained. Fine for the two current call sites (`'invoiced'` / `'fiscally registered'`); if a third document kind adopts this, consider narrowing to a closed union for consistent wording. Not blocking.

**[SUGGESTION]** — the two adapter "audit comments" (PrestaShop, WooCommerce) are non-functional and restate reasoning already in the helper's docblock. Acceptable given the PR's stated goal (AC #4) of cross-referencing the decision at both net-pricing sites so a future reader doesn't re-litigate it.

### Test coverage
✅ New helper has a dedicated spec (gross accepted / absent accepted / exclusive refused with exact message / action-text threading). Both existing mapper specs got one assertion each for the new message shape — proportionate for a pure message-composition change.

### Final Assessment

Narrow, well-reasoned fix. Replaces an unqualified, misleading refusal message with one naming the actual permanent constraint (net line prices, not something #2829/#2832 relaxes). Correctly placed in the `sales-documents` leaf with zero new imports.

**Merge Readiness**: ✅ Approve — no priority fixes required.

norbert-kulus-blockydevs:
Addressed the one actionable suggestion from the review (8a71ded).

- `describeNetPricedOrderRefusal`'s `action` parameter narrowed from `string` to a closed union `NetPricedOrderRefusalAction = 'invoiced' | 'fiscally registered'` (`NetPricedOrderRefusalActionValues` as const, per the repo's union-type convention). Both existing call sites (`invoicing`, `fiscalization`) already pass exactly these literals, so this is source-compatible; a future third document kind now gets a compile error instead of silent wording drift.

The second suggestion (the two non-functional adapter audit comments in PrestaShop/WooCommerce) was called out in the review as acceptable given AC #4's cross-reference goal, so left as-is.

Verified: `tsc --noEmit` clean, `eslint` clean, targeted specs (`gross-price-eligibility.types.spec.ts`, both mapper specs) all passing.

piotrswierzy (approved):
## Code Review — #2837 (name the actual constraint when refusing a net-priced order)

11 files, +213/−13, against `main`. Pure message/documentation change plus one new shared pure helper; no behavioural change to what is refused.

### 1. Context

`toIssueInvoiceCommand` and `toRegisterTransactionCommand` each composed a document whose lines are gross by contract (`InvoiceLine.unitPriceGross` / `FiscalTransactionLine.unitPriceGross`) from `OrderItem.price`, and refused `taxTreatment === 'exclusive'` with "only gross-priced orders are supported". This PR routes both through one helper in the `sales-documents` leaf, replaces the message, and adds cross-references at the two adapter sites and both error classes.

### 2. Findings

**Positive — the refusal predicate is byte-identical, and that matters here.** `describeNetPricedOrderRefusal` returns `null` for anything that is not exactly `'exclusive'`, which reproduces both call sites' prior `=== 'exclusive'` test including the documented absent-treatment acceptance. A change framed as "better wording" is exactly where a guard quietly widens or narrows; this one does not, and both mapper specs still carry their original throw-type assertions beside the new message assertions.

**Positive — the central claim is correct, and it is the one a future reader would get wrong.** The docblock's argument that #2829/#2832's `totalTaxTreatment` cannot relax this guard is right and non-obvious: `totalTaxTreatment` describes `total` alone, while both mappers compose `unitPriceGross` per item, so reading it here would let a PrestaShop order past the guard and then mislabel still-net line prices as gross — the same corruption one level down. Given #2832/#2838 are in flight *right now* adding precisely that signal, recording this at the guard rather than only on the issue is the difference between a decision and a trap. The refusal to convert is also correctly grounded (ADR-063 § 5 — core may group and divide, never compute tax), including the sharp part: knowing `OrderItem.taxRate` does not license computing with it.

**Positive — placement follows the leaf's own rules.** `sales-documents` is the zero-outbound-core-edge sink; the new file imports nothing, so `libs/core/src/__tests__/barrel-purity.spec.ts` stays satisfied and both contexts can value-import it. The "a fiscal receipt is not an invoice, so neither context could own it for the other" reasoning is the same one `shipping-tax-split.types` and `tax-rate-enforcement.types` already stand on. A runtime function in a `*.types.ts` is the documented pure-rule exception in `engineering-standards.md` — pure, it *is* the rule for the type beside it, and both change together. All three conditions hold.

**Positive — `NetPricedOrderRefusalAction` is a closed `as const` union**, not a bare `string`, so a third caller is a compile-time event rather than silent wording drift, and the action reads correctly in both clauses ("cannot be {action}" / "can be {action} today"). The spec's fourth case pins exactly that dual use. Message carries no country or document-type vocabulary — ADR-026 § country-agnostic holds.

**[IMPORTANT]** — test plan: `pnpm test` was not run

The PR states `pnpm test` was not run locally. `CLAUDE.md` § Quality gate makes it non-optional before commit, and it bites specifically here: three of the four new/changed assertions are string/regex matches against a template literal, so a single-character drift in the message body is invisible to `lint` and `type-check` (both of which are reported clean) and only `test` catches it. The change is low-risk and CI covers it, so this is not blocking — but please confirm the suite green before merge rather than merging on CI-will-tell.

**[IMPORTANT]** — `gross-price-eligibility.types.ts:53-58` — `taxTreatment?: string` is wider than the domain type

`GrossPriceEligibilityOrder.totals.taxTreatment` is typed `string`, while the value it models is `PriceTaxTreatment` (`libs/core/src/orders/domain/types/order.types.ts:319-320`). I understand why: importing it would add a fourth type-only exception to the `sales-documents` barrel-purity allowlist, and the leaf property is worth more than the narrowing. Two consequences worth stating rather than leaving implicit:

- The guard **fails open** on any unrecognised value. A hypothetical `'excl'` returns `null` and the order is composed as gross. Today that cannot happen — every call site passes a real `Order` whose field is `PriceTaxTreatment`, and structural typing keeps them honest — so this is latent, not live.
- Because it is latent, the mitigation should be a comment, not a type change. The interface's current comment says what it is ("the narrow order shape this check needs") but not why it is `string` rather than the domain union, so the next reader may "tighten" it and reintroduce the cross-context import the leaf exists to avoid. One sentence naming the barrel-purity constraint would close that.

Alternatively, if the fourth `@openlinker/core/orders/types` type-only exception is acceptable to the maintainers (the spec already authorizes `Order` from that exact sub-barrel, and a type-only import erases), importing `PriceTaxTreatment` makes the guard closed and is arguably the better end state. Either resolution is fine; leaving it undocumented is the thing to avoid.

**[SUGGESTION]** — Prettier width in the two spec files

Several added lines exceed the repo's `printWidth: 100` — `order-to-issue-invoice-command.mapper.spec.ts:274` (the inline `totals: {…}` object) and `gross-price-eligibility.types.spec.ts:10-12` / `:17-19` (the single-line `describeNetPricedOrderRefusal(…)` calls). `pnpm lint` does not check formatting, so these pass the reported gate and get reflowed by the pre-commit hook or the next person who touches the file, producing diff noise. Run `prettier --write` on the changed files only (`docs/lessons.md` / repo practice: never a broad glob here).

**[SUGGESTION]** — message says "today", docblock says PERMANENT

The operator-facing sentence ends "…can be invoiced today", which reads as a temporary MVP limit, while the helper's docblock is emphatic that this is permanent under the current architecture and clears only if the *source adapter* starts reporting gross line prices. Both are defensible, and the docblock's own escape hatch justifies "today" — but since the whole point of #2835 is that the old message was misread, consider whether "today" invites the same misreading (an operator waiting for a release that will never come, rather than looking at their source platform). Naming the source as the locus — the docblock already does — would make the sentence self-consistent with the decision it encodes.

### 3. Standards / architecture

- Layer placement correct throughout: helper in the leaf's `domain/types/`, guards stay in each context's `application/mappers/`, adapter comments in `libs/integrations/**`. No domain logic moves into an integration and no platform vocabulary enters `libs/core`.
- Naming, `as const` union pattern, file header with `@module` + `@see`, no `any`, colocated `*.spec.ts` — all per `engineering-standards.md` / `docs/testing-guide.md`.
- No schema change, so no migration expected; no security or idempotency surface.
- The two adapter comments satisfy the issue's audit AC and are `why`-comments, not restatements.

### 4. Final Assessment

**Summary.** A small, well-argued diagnostics fix whose real value is the recorded decision: it forecloses the specific wrong conclusion that #2829/#2832's `totalTaxTreatment` should relax this guard, at the exact moment that signal is being introduced. Behaviour is unchanged, placement respects the `sales-documents` leaf invariant, and the pure-rule `*.types.ts` exception is correctly claimed. Nothing blocking.

**Merge readiness:** 🔄 **Approve with changes** — merge once the two IMPORTANT items are addressed.

**Priority fixes:**
1. Run `pnpm test` and confirm green (the assertions this PR adds are string matches; `lint` + `type-check` cannot see a template drift).
2. Either document *why* `GrossPriceEligibilityOrder.totals.taxTreatment` is `string` rather than `PriceTaxTreatment` (barrel-purity/leaf constraint), or import the union as a fourth authorized type-only exception.
3. (Optional) `prettier --write` the two spec files; reconsider "today" in the operator-facing sentence.

norbert-kulus-blockydevs:
Addressed both IMPORTANT findings and both suggestions:

1. **`pnpm test` confirmed green** — ran the two changed spec files directly (37 tests passed) plus the full `libs/core` suite locally.
2. **Documented why `GrossPriceEligibilityOrder.totals.taxTreatment` is `string`** rather than `PriceTaxTreatment` — added a docblock explaining the barrel-purity tradeoff and the latent (not live) fail-open behavior, so a future reader doesn't "tighten" it and reintroduce a cross-context import.
3. Ran `prettier --write` on the two changed spec files.
4. Dropped "today" from the operator-facing refusal message — the constraint is permanent under the current architecture; the docblock and message now both point at the source platform as the thing that would need to change.

Commit: 01f741f99

piotrswierzy (approved):
## Re-review — #2837 @ `d840e54` (was `8a71ded`)

Re-reviewed against the current head after the force-push. All three items from my previous APPROVE-with-changes are resolved; two of them beyond what I asked for.

### 1. Per-item verification of prior findings

**✅ RESOLVED — the guard predicate is still byte-identical, and the central argument still holds.**

`describeNetPricedOrderRefusal` returns `null` for anything that is not exactly `'exclusive'` (`gross-price-eligibility.types.ts:96-98`), reproducing both call sites' prior `=== 'exclusive'` test including the documented absent-treatment acceptance. Both mappers still throw their original error classes (`UnsupportedPriceTreatmentError` / `UnsupportedFiscalPriceTreatmentError`) and both specs retain their original throw-type assertions beside the new message assertions. Nothing widened or narrowed across the rewrite.

The load-bearing claim — that `totalTaxTreatment` (#2829/#2832) must not relax a LINE-level guard — survives intact in the module docblock, and the force-push actually strengthened its placement: both mapper call sites now carry an inline `This checks LINE-level taxTreatment, never totals.totalTaxTreatment (#2829/#2832)` comment directly above the call. That is the right altitude — a reader touching the guard sees it without navigating to the helper. Verified the old unqualified string is gone from live code at this head (`git grep "only gross-priced orders" d840e548` returns only the docblock quoting it as the thing being replaced).

**✅ RESOLVED — `pnpm test` is green on this head.**

`get_check_runs` on `d840e54` reports **Test: success** (run 33847607447, job 100942941834, 07:28→07:45), alongside Lint, Type Check, Build, Docker Build Smoke Test, PHP Unit Tests and Scaffolded Adapter Builds — all success. Integration Tests still in progress at review time, but this diff touches no persistence, no job, no adapter behaviour, so it has no int-spec surface. This was the finding that mattered: three of the four new assertions are string/regex matches against a template literal, invisible to `lint` and `type-check`. They are now actually exercised.

**✅ RESOLVED — and better than either option I offered.**

`GrossPriceEligibilityOrder.totals.taxTreatment` stays `string`, but now carries a 10-line docblock (`:53-64`) that names the barrel-purity tradeoff explicitly, states the fail-open behaviour is *latent, not live* and why (every real caller passes an `Order` whose field is already `PriceTaxTreatment`; structural typing keeps them honest), and closes with `Do not "tighten" this to close the latent gap without first re-reading the leaf-purity tradeoff above.` That last sentence is the whole point — it addresses the future reader directly rather than leaving them to infer the constraint. Keeping `string` over adding a fourth allowlist exception is the right call for a one-field narrowing that buys nothing today.

**✅ Both suggestions also taken.** `prettier --write` ran on the two spec files, and `"today"` is gone from the operator-facing sentence — the message now ends `only a source that reports gross (tax-inclusive) line prices can be {action}`, which points at the source platform, consistent with the docblock's own escape hatch. The helper's own docblock explains *why* "today" was dropped, so the decision does not get quietly reverted.

### 2. New code since `8a71ded`

The prettier pass produced ~18 unrelated reformat hunks in `order-to-issue-invoice-command.mapper.spec.ts`, most of them stripping trailing commas (`InvalidInvoiceLineError,` → `InvalidInvoiceLineError`). I checked whether this fights the repo: `.prettierrc` declares `trailingComma: "es5"` (Prettier 3 defaults to `"all"`), and the leaf sibling `shipping-tax-split.types.ts` on `main` is already committed in es5 style (`minorUnitExponent: number = DEFAULT_MINOR_UNIT_EXPONENT` with no trailing comma). So these lines now match the declared config and their nearest neighbours — **this is correct, not churn**, and CI Lint agrees. Noting it only because the diff looks alarming at a glance and because `main` is genuinely mixed, so the two spec files in this PR now differ from each other in style (the fiscalization spec was not reformatted). Harmless.

The two adapter audit comments are unchanged and still satisfy AC #4 as `why`-comments.

### 3. `sales-documents` leaf invariant

Re-verified at this head rather than assuming: `gross-price-eligibility.types.ts` has **zero import statements**, so the zero-outbound-core-edge property that `libs/core/src/__tests__/barrel-purity.spec.ts` guards is untouched — the allow-set for `sales-documents` stays at the three authorized type-only `<ctx>/types` specifiers (`orders/types`, `invoicing/types`, `fiscalization/types`, spec `:207-209`) and this PR adds no fourth. Both consumers import through the top-level `@openlinker/core/sales-documents` barrel; no deep paths. Export added to `index.ts` with a `why` comment matching the file's neighbours.

- Runtime function in a `*.types.ts` is the documented pure-rule exception (`engineering-standards.md` § "The pure-rule exception"): pure, it *is* the rule for the type beside it, both change together. All three conditions hold, and it joins the four cited precedents.
- `NetPricedOrderRefusalAction` is a closed `as const` union per the union-type convention; the spec's fourth case pins its dual use in both clauses.
- File header with `@module` + `@see`, colocated `*.spec.ts`, no `any`, layer placement correct (helper in the leaf's `domain/types/`, guards in each context's `application/mappers/`, comments in `libs/integrations/**`). No platform vocabulary in `libs/core`.
- No schema change → no migration. No security, idempotency or concurrency surface.

### 5. Remaining finding

**[SUGGESTION]** — `unsupported-price-treatment.error.ts:6` still says "MVP", contradicting the paragraph this PR added three lines below it.

The first paragraph reads `the MVP maps gross-priced orders only`, while the new paragraph at `:10` says `This is a PERMANENT limitation … not a gap`. The PR deliberately fixed exactly this word in the mapper (`GROSS-only MVP:` → `GROSS-only:` in `order-to-issue-invoice-command.mapper.ts:100`) but not in the error class it also edited — so the one file a reader lands on from a stack trace now says both "MVP" and "PERMANENT" about the same constraint. One-word fix; it is the last place the misreading #2835 exists to close can still be picked up.

Same docblock, same line: `No country/document-type vocabulary` is now true in spirit but not quite in letter — the message says `for a fiscal document` and threads `invoiced` / `fiscally registered`. Those are core's own neutral `SalesDocumentKind` vocabulary, not the country-specific vocabulary ADR-026 forbids, so the constraint is honoured; the sentence just over-claims. Worth a clause if you are touching the line anyway.

Neither is blocking.

### Final Assessment

**Summary.** Every prior finding resolved, two of them beyond the ask — the `string`-typing docblock now instructs the future reader rather than merely explaining, and the inline `never totals.totalTaxTreatment` comments put the decision where it will actually be read. Behaviour unchanged, leaf invariant verified intact at this head, and the full unit suite is green on this exact commit, which is what the earlier review was waiting for.

**Merge readiness:** ✅ **Approve** — no priority fixes. Branch is `behind` `main`; update before merging.

piotrswierzy (approved):
## Re-review — #2837 @ `37d577f`

Re-verified every prior finding against the current head rather than carrying them forward.

**First, what the force-push actually changed.** `git diff d840e548 37d577fb` touches only `apps/web/**` (12 files, the `sales-document-panel` / `invoice-lifecycle-steps` work from #2763 landing on `main`). The PR's own 11-file diff against `origin/main` is byte-identical to what I reviewed at `d840e54`: this was a **rebase onto a newer base, not a content revision**. So the prior verification carries — but I re-checked each item at this SHA anyway, because a rebase is exactly where a hand-resolved conflict silently reverts a docblock.

### 1. Per-item verification of prior findings

**✅ HOLDS — guard predicate byte-identical.** `gross-price-eligibility.types.ts:96-98` — `if (order.totals.taxTreatment !== 'exclusive') return null;`. Reproduces both call sites' prior `=== 'exclusive'` test exactly, including the documented absent-treatment acceptance. Both mappers still throw their original classes (`UnsupportedPriceTreatmentError` / `UnsupportedFiscalPriceTreatmentError`, confirmed in both diffs), and both specs retain their original throw-type assertions (`mapper.spec.ts:318` and `:151`) beside the new message assertions. Nothing widened or narrowed across the rebase.

The inline call-site comments survive intact in both mappers — `This checks LINE-level taxTreatment, never totals.totalTaxTreatment (#2829/#2832)` — which remains the single most valuable line in the diff, since #2832/#2838 are in flight right now introducing exactly the signal a future reader would reach for. The unqualified old string is gone from live code (`git grep "only gross-priced orders" 37d577fb -- libs/core libs/integrations` returns one hit, the docblock quoting it as the thing replaced).

**✅ HOLDS — CI green on this exact head.** `get_check_runs` on `37d577f`: **Test success** (job 100976654892, 11:02→11:19), and this time **Integration Tests success** too (job 100976654926, 10:37→11:12) — the one job still in flight at my last review. Lint, Type Check, Build, Docker Build Smoke, PHP Unit Tests, Scaffolded Adapter Builds all success. 9/9 green. The finding that mattered — three of four new assertions are string matches invisible to `lint`/`type-check` — is now closed on the merge candidate itself.

**✅ HOLDS — the `string`-typing docblock survived the rebase verbatim.** `:53-64` still names the barrel-purity tradeoff, still states the fail-open is *latent, not live* and why (every real caller passes an `Order` whose field is already `PriceTaxTreatment`; structural typing keeps them honest), and still closes with `Do not "tighten" this to close the latent gap without first re-reading the leaf-purity tradeoff above.` Addressing the future reader directly, rather than merely explaining, is what makes this better than either option I originally offered.

### 2. `sales-documents` leaf invariant — re-verified, not assumed

`gross-price-eligibility.types.ts` has **zero import statements** (`grep -n "^import\|from '"` → empty). The zero-outbound-core-edge property `libs/core/src/__tests__/barrel-purity.spec.ts` guards is untouched: the `sales-documents` `authorizedTypeOnlySpecifiers` array at `:206-210` still holds exactly the three authorized entries (`orders/types`, `invoicing/types`, `fiscalization/types`), and this PR adds no fourth. Both consumers import through the top-level `@openlinker/core/sales-documents` barrel — no deep paths — and the `index.ts` export carries a `why` comment matching its neighbours (`shipping-tax-split`, `tax-rate-enforcement`, `sales-document-in-flight`).

### 3. Standards / architecture

- Runtime function in `*.types.ts` is the documented pure-rule exception (`engineering-standards.md` § "The pure-rule exception"): pure, it *is* the rule for the type beside it, both change together. All three conditions hold; it joins the four cited precedents.
- `NetPricedOrderRefusalAction` is a closed `as const` union per the union-type convention, and reads naturally in both clauses ("cannot be {action}" / "can be {action}").
- File header with `@module` + `@see`, colocated `*.spec.ts`, no `any`, layer placement correct (helper in the leaf's `domain/types/`, guards in each context's `application/mappers/`, audit comments in `libs/integrations/**` as `why`-comments satisfying AC #4). No platform vocabulary in `libs/core`.
- No schema change → no migration. No security, idempotency or concurrency surface.

### 4. The one open item — **still open**

**[SUGGESTION]** — `unsupported-price-treatment.error.ts:6` still says `the MVP maps gross-priced orders only`, three lines above the paragraph this PR added at `:10` saying `This is a PERMANENT limitation … not a gap`.

Unchanged by the force-push, as expected given it was a rebase. The PR deliberately fixed this exact word in the mapper (`GROSS-only MVP:` → `GROSS-only:`) but not in the error class it also edits — so the **one file a reader lands on from a stack trace** asserts both "MVP" and "PERMANENT" about the same constraint. Note the fiscal sibling has no such contradiction (its docblock never said MVP), so this is a single-word asymmetry in one file.

Same docblock, same line: `No country/document-type vocabulary` is true in spirit but over-claims in letter — the message now threads `invoiced` / `fiscally registered` and says `for a fiscal document`. Those are core's own neutral `SalesDocumentKind` vocabulary, not the country-specific vocabulary ADR-026 forbids, so the constraint is honoured; the sentence just no longer describes the message accurately. Worth a clause if you touch the line anyway.

Non-blocking, and I would not hold a merge for it — but it is the last place the misreading #2835 exists to close can still be picked up, and it is one word.

### Final Assessment

**Summary.** A rebase-only update: the reviewed content is unchanged and every prior finding independently re-verified at this SHA. What is new and material is that **Integration Tests now report success on this exact commit**, closing the last in-flight job from my previous review — the full gate is 9/9 green on the merge candidate rather than on an ancestor. Leaf invariant re-checked from the allowlist array itself, not from memory.

**Merge readiness:** ✅ **Approve** — no priority fixes. Branch is `behind` `main` (based on `df46c641`, `main` is at `608813dc`); update before merging.

Issue #4, opened 2025-12-09
Epic: Orders & Routing


piotrswierzy:
**US-14 – Import Allegro orders**

> As a merchant, I want OpenLinker to import new Allegro orders automatically so that I don’t have to manually copy them into Presta.

**US-15 – Create Presta orders**

> As a merchant, I want OpenLinker to create standard Presta orders for each Allegro order so that my invoicing and shipping modules continue to work without changes.

**US-16 – Batch order creation**

> As a merchant, I want to select multiple Allegro orders in OpenLinker and create their corresponding Presta orders in bulk so that I can fix past integration issues or import historical orders efficiently.

**US-17 – Failed order diagnostics**

> As a merchant, I want to see which Allegro orders failed to create in Presta and why so that I can fix problems (e.g. missing mapping) and retry.

**US-18 – Status mapping**

> As a merchant, I want Allegro order/payment statuses to be correctly reflected in Presta orders and in OpenLinker so that I always see the real payment and fulfillment state.

piotrswierzy:
Closing as completed — every user story has shipped:

- **US-14 Import Allegro orders** — `AllegroOrderSourceAdapter` (`OrderSourcePort`), event journal via `GET /sale/order-events`.
- **US-15 Create Presta orders** — `PrestashopOrderProcessorAdapter` (`OrderProcessorManagerPort`).
- **US-16 Batch order creation** — Failed-jobs grouping on the dashboard with `useRetryGroupedSyncJobsMutation` covers the bulk-retry case. (If a richer "select N failed orders → bulk reprocess" UX is wanted, file a focused issue.)
- **US-17 Failed order diagnostics** — `/orders?syncStatus=awaiting_mapping` failed-orders page + per-order activity timeline.
- **US-18 Status mapping** — status mappings page on the connection.

If a specific gap surfaces, file a focused issue.

PR #2838, opened 2026-09-03
fix(woocommerce): let a WooCommerce order's total be evaluated by an orderTotalGross sales-document rule

### LINKED ISSUE #2836: [BUG] CORE — WooCommerce order total is gross but taxTreatment is line-scoped (same split as #2829)
## Problem / Context

`#2829`/`#2832` fixed this for PrestaShop: `PrestashopOrderMapper` sets `taxTreatment: 'exclusive'` because line prices (`order_details.product_price`) are net, but the order's own **total** (`total_paid_tax_incl`) is genuinely gross — one flag couldn't describe both, so no PrestaShop order could ever match the `orderTotalGross` sales-document rule condition (`evaluateSalesDocumentRules`, consumed via `toSalesDocumentOrderFacts` and `SalesDocumentViewService.toOrderFacts`), regardless of the order's actual tax setup.

WooCommerce has the **identical structural split** in `libs/integrations/woocommerce/src/infrastructure/adapters/woocommerce-order-source.adapter.ts:271`: `line_items[].price` is net and `total_tax` is reported as a separate decomposed amount, but the order's `total` is genuinely gross — same reasoning, same fix needed. This was noted as scoped-out in #2832's PR description and flagged as a SUGGESTION in review.

## Proposed Solution

`OrderTotals.totalTaxTreatment` / `IncomingOrderTotals.totalTaxTreatment` already exist (`libs/core/src/orders/domain/types/order.types.ts`, `incoming-order.types.ts`, added by #2832) precisely for this split. Set `totalTaxTreatment: 'inclusive'` alongside the existing `taxTreatment: 'exclusive'` in the WooCommerce totals mapper, mirroring `PrestashopOrderMapper`'s pattern.

If a database backfill is wanted for WooCommerce order history, follow `apps/api/src/migrations/1870000000000-add-order-record-total-tax-treatment.ts`'s pattern — same idempotent `WHERE "totalTaxTreatment" IS NULL` shape, scoped by `connections.platformType = 'woocommerce'` instead of `'prestashop'`.

## Classification

**Type**: Integration
**Layer**: Infrastructure (adapter mapper)
**File(s)**:
- `libs/integrations/woocommerce/src/infrastructure/adapters/__tests__/woocommerce-order-source.adapter.spec.ts`
- (optional) a new `apps/api/src/migrations/*-backfill-order-record-total-tax-treatment-woocommerce.ts`

## Dependencies

- #2832 must be merged first (it adds the `totalTaxTreatment` field this issue populates for WooCommerce).

## Assumptions

- No downstream WooCommerce-specific consumer reads `taxTreatment` in a way a `totalTaxTreatment` addition could disturb — this must be re-verified with the same consumer audit #2832's PR description performed for PrestaShop (grep `convertGrossToNet`-equivalent WooCommerce destination logic and the ADR-063 net-sales tax-rate path for any `taxTreatment` read).
- History backfill is optional/discretionary — the field is additive and falls back to `taxTreatment` when unset, so existing WooCommerce orders keep their current (net-priced) `orderTotalGross` evaluation until backfilled.

## Acceptance Criteria

- [ ] WooCommerce order-source adapter sets `totalTaxTreatment: 'inclusive'` alongside unchanged `taxTreatment: 'exclusive'`
- [ ] Mapper spec asserts `totalTaxTreatment: 'inclusive'` alongside the unchanged `taxTreatment: 'exclusive'` case
- [ ] Consumer audit confirms no WooCommerce-side reader of `taxTreatment` is affected by the additive field

norbert-kulus-blockydevs:
## Summary

- `line_items[].price` (mapped onto `OrderItem.price`) is net, but WooCommerce's `order.total` is genuinely gross — the identical structural split #2829/#2832 fixed for PrestaShop.
- Sets `totalTaxTreatment: 'inclusive'` alongside the unchanged `taxTreatment: 'exclusive'` in `mapTotals`, so the `orderTotalGross` sales-document rule condition (`evaluateSalesDocumentRules`) can correctly evaluate WooCommerce orders.

**Stacked on #2832**: this branches off `2829-prestashop-tax-treatment` because the `OrderTotals.totalTaxTreatment` / `IncomingOrderTotals.totalTaxTreatment` field this issue populates was added there. The diff will narrow to just the WooCommerce change once #2832 merges and this PR is retargeted to `main`.

**Consumer audit** (per the issue's acceptance criteria): grepped `libs/integrations/woocommerce/src/` for `taxTreatment` — the only reference is the one this PR adds plus its accompanying test. No WooCommerce-specific reader of `taxTreatment` exists that this additive field could disturb.

**History backfill**: left out per the issue's own note that it's optional/discretionary — the field is additive and falls back to `taxTreatment` when unset.

## Test plan

- [x] `pnpm --filter @openlinker/integrations-woocommerce test -- woocommerce-order-source.adapter.spec.ts` — 28/28 pass
- [x] `pnpm --filter @openlinker/integrations-woocommerce lint` — 0 errors (2 pre-existing unrelated warnings)
- [x] `pnpm --filter @openlinker/integrations-woocommerce type-check` — clean

Closes #2836

### DIFF
--- a/apps/api/src/migrations/1870000000000-add-order-record-total-tax-treatment.ts
+++ b/apps/api/src/migrations/1870000000000-add-order-record-total-tax-treatment.ts
@@ -1,13 +1,16 @@
 /**
- * Add `order_records.totalTaxTreatment` and backfill it for PrestaShop (#2829/#2832).
+ * Add `order_records.totalTaxTreatment` and backfill it for PrestaShop and
+ * WooCommerce (#2829/#2832/#2836).
  *
  * `order_records.taxTreatment` (#1985/#2440) describes the LINE prices /
- * subtotal — for PrestaShop those are net (`order_details.product_price`),
- * so the column reads `'exclusive'`. But PrestaShop's order TOTAL
- * (`total_paid_tax_incl`) genuinely is gross, and the `orderTotalGross`
- * sales-document rule condition reads `taxTreatment` to decide whether it may
- * trust the total as gross at all — so no PrestaShop order could ever match
- * that condition, regardless of its real tax setup.
+ * subtotal — for both PrestaShop and WooCommerce those are net
+ * (`order_details.product_price`, `line_items[].price`), so the column reads
+ * `'exclusive'`. But each platform's order TOTAL (PrestaShop's
+ * `total_paid_tax_incl`, WooCommerce's `order.total`) genuinely is gross, and
+ * the `orderTotalGross` sales-document rule condition reads `taxTreatment` to
+ * decide whether it may trust the total as gross at all — so no order from
+ * either platform could ever match that condition, regardless of its real
+ * tax setup.
  *
  * `taxTreatment` cannot simply be flipped to `'inclusive'`: it also drives
  * `PrestashopOrderProcessorManagerAdapter.convertGrossToNet` (destination-side
@@ -20,23 +23,23 @@
  * `totalTaxTreatment ?? taxTreatment`, the same fallback the live-`Order`
  * gate applies via `toSalesDocumentOrderFacts`).
  *
- * The adapter fix (companion PR) only reaches orders ingested from now on.
- * This migration backfills history for PrestaShop, following the exact
- * platform-scoped join precedent set by
+ * The adapter fixes (companion PRs #2832/#2838) only reach orders ingested
+ * from now on. This migration backfills history for BOTH platforms in one
+ * pass, following the exact platform-scoped join precedent set by
  * `1841000000006-backfill-order-records-tax-treatment.ts` (itself following
- * `1840000000000-reset-fx-stamp-for-mislabelled-prestashop-orders.ts`).
+ * `1840000000000-reset-fx-stamp-for-mislabelled-prestashop-orders.ts`). The
+ * WooCommerce arm was originally deferred to #2836 as a follow-up (per
+ * review on #2838); folded in here instead, since the `UPDATE` is already
+ * platform-scoped and idempotent, so a second migration to add one more
+ * platform value would be pure ceremony.
  *
  * Idempotent by construction (`WHERE "totalTaxTreatment" IS NULL`) — a re-run
  * touches only rows a later fix hasn't already reached, matching the house
  * convention documented on `1818000000004-backfill-ksef-provider-invoice-number.ts`.
  *
- * Deliberately does NOT touch WooCommerce, despite the identical structural
- * split (net line prices + `total_tax` decomposed separately, but `total` is
- * genuinely gross) — tracked as a follow-up (#2836) and cited from
- * `OrderTotals.totalTaxTreatment`'s doc comment. A connection whose platform
- * this migration does not assert stays `totalTaxTreatment IS NULL`, i.e.
- * "same as `taxTreatment`", which is the existing (net-priced) behaviour —
- * never a guess.
+ * A connection whose platform is neither of these two stays
+ * `totalTaxTreatment IS NULL`, i.e. "same as `taxTreatment`", which is the
+ * existing (net-priced) behaviour — never a guess.
  *
  * Generated: 2026-09-03 (synthetic sequential prefix per docs/migrations.md
  * rule 3; 1869000000900 is #2385's `automation_runs` retry-attempt column).
@@ -56,7 +59,7 @@ export class AddOrderRecordTotalTaxTreatment1870000000000 implements MigrationIn
       SET "totalTaxTreatment" = 'inclusive'
       FROM "connections" c
       WHERE c."id" = o."sourceConnectionId"
-        AND c."platformType" = 'prestashop'
+        AND c."platformType" IN ('prestashop', 'woocommerce')
         AND o."totalTaxTreatment" IS NULL
     `);
   }
--- a/apps/api/src/migrations/1870000000000-add-order-record-total-tax-treatment.ts
+++ b/apps/api/src/migrations/1870000000000-add-order-record-total-tax-treatment.ts
@@ -1,16 +1,13 @@
 /**
- * Add `order_records.totalTaxTreatment` and backfill it for PrestaShop and
- * WooCommerce (#2829/#2832/#2836).
+ * Add `order_records.totalTaxTreatment` and backfill it for PrestaShop (#2829/#2832).
  *
  * `order_records.taxTreatment` (#1985/#2440) describes the LINE prices /
- * subtotal — for both PrestaShop and WooCommerce those are net
- * (`order_details.product_price`, `line_items[].price`), so the column reads
- * `'exclusive'`. But each platform's order TOTAL (PrestaShop's
- * `total_paid_tax_incl`, WooCommerce's `order.total`) genuinely is gross, and
- * the `orderTotalGross` sales-document rule condition reads `taxTreatment` to
- * decide whether it may trust the total as gross at all — so no order from
- * either platform could ever match that condition, regardless of its real
- * tax setup.
+ * subtotal — for PrestaShop those are net (`order_details.product_price`),
+ * so the column reads `'exclusive'`. But PrestaShop's order TOTAL
+ * (`total_paid_tax_incl`) genuinely is gross, and the `orderTotalGross`
+ * sales-document rule condition reads `taxTreatment` to decide whether it may
+ * trust the total as gross at all — so no PrestaShop order could ever match
+ * that condition, regardless of its real tax setup.
  *
  * `taxTreatment` cannot simply be flipped to `'inclusive'`: it also drives
  * `PrestashopOrderProcessorManagerAdapter.convertGrossToNet` (destination-side
@@ -23,23 +20,23 @@
  * `totalTaxTreatment ?? taxTreatment`, the same fallback the live-`Order`
  * gate applies via `toSalesDocumentOrderFacts`).
  *
- * The adapter fixes (companion PRs #2832/#2838) only reach orders ingested
- * from now on. This migration backfills history for BOTH platforms in one
- * pass, following the exact platform-scoped join precedent set by
+ * The adapter fix (companion PR) only reaches orders ingested from now on.
+ * This migration backfills history for PrestaShop, following the exact
+ * platform-scoped join precedent set by
  * `1841000000006-backfill-order-records-tax-treatment.ts` (itself following
- * `1840000000000-reset-fx-stamp-for-mislabelled-prestashop-orders.ts`). The
- * WooCommerce arm was originally deferred to #2836 as a follow-up (per
- * review on #2838); folded in here instead, since the `UPDATE` is already
- * platform-scoped and idempotent, so a second migration to add one more
- * platform value would be pure ceremony.
+ * `1840000000000-reset-fx-stamp-for-mislabelled-prestashop-orders.ts`).
  *
  * Idempotent by construction (`WHERE "totalTaxTreatment" IS NULL`) — a re-run
  * touches only rows a later fix hasn't already reached, matching the house
  * convention documented on `1818000000004-backfill-ksef-provider-invoice-number.ts`.
  *
- * A connection whose platform is neither of these two stays
- * `totalTaxTreatment IS NULL`, i.e. "same as `taxTreatment`", which is the
- * existing (net-priced) behaviour — never a guess.
+ * Deliberately does NOT touch WooCommerce, despite the identical structural
+ * split (net line prices + `total_tax` decomposed separately, but `total` is
+ * genuinely gross) — tracked as a follow-up (#2836) and cited from
+ * `OrderTotals.totalTaxTreatment`'s doc comment. A connection whose platform
+ * this migration does not assert stays `totalTaxTreatment IS NULL`, i.e.
+ * "same as `taxTreatment`", which is the existing (net-priced) behaviour —
+ * never a guess.
  *
  * Generated: 2026-09-03 (synthetic sequential prefix per docs/migrations.md
  * rule 3; 1869000000900 is #2385's `automation_runs` retry-attempt column).
@@ -59,7 +56,7 @@ export class AddOrderRecordTotalTaxTreatment1870000000000 implements MigrationIn
       SET "totalTaxTreatment" = 'inclusive'
       FROM "connections" c
       WHERE c."id" = o."sourceConnectionId"
-        AND c."platformType" IN ('prestashop', 'woocommerce')
+        AND c."platformType" = 'prestashop'
         AND o."totalTaxTreatment" IS NULL
     `);
   }
--- a/apps/api/src/migrations/1870000000100-backfill-order-record-total-tax-treatment-woocommerce.ts
+++ b/apps/api/src/migrations/1870000000100-backfill-order-record-total-tax-treatment-woocommerce.ts
@@ -1,0 +1,62 @@
+/**
+ * Backfill `order_records.totalTaxTreatment` for WooCommerce (#2836).
+ *
+ * `1870000000000-add-order-record-total-tax-treatment.ts` (#2829/#2832) added
+ * the column and backfilled it for PrestaShop only, explicitly deferring
+ * WooCommerce to this issue — see that migration's own doc comment and
+ * `OrderTotals.totalTaxTreatment`'s.
+ *
+ * WooCommerce has the identical structural split PrestaShop does: line
+ * prices (`line_items[].price`) are net, so `taxTreatment` stays
+ * `'exclusive'`, but the order TOTAL (`order.total`) genuinely is gross —
+ * see `woocommerce-order-source.adapter.ts`'s `mapTotals`, the adapter-side
+ * half of #2836.
+ *
+ * This is a SEPARATE migration rather than an edit to 1870000000000, and
+ * that is deliberate rather than the ceremony it looks like
+ * (`docs/migrations.md` rule 4, "never edit an already-executed migration"):
+ * 1870000000000 ships in a sibling PR (#2832) that may merge — and run in a
+ * live environment — before this one lands. Editing that file in place would
+ * make the WooCommerce backfill's effect depend on merge order: if
+ * 1870000000000 already executed anywhere, TypeORM's migrations table would
+ * never re-run it, and an in-place edit adding the WooCommerce arm would
+ * silently never apply. A standalone, order-independent migration has no
+ * such dependency — it runs (and backfills WooCommerce) whenever it is
+ * deployed, regardless of when 1870000000000 ran.
+ *
+ * Idempotent by construction (`WHERE "totalTaxTreatment" IS NULL`), matching
+ * 1870000000000's own guard and the house convention documented on
+ * `1818000000004-backfill-ksef-provider-invoice-number.ts` — a re-run (or a
+ * run before/after 1870000000000, in either order) touches only rows a
+ * platform-scoped fix hasn't already reached.
+ *
+ * Generated: 2026-09-04 (synthetic sequential prefix per docs/migrations.md
+ * rule 3; 1870000000000 is #2829/#2832's `totalTaxTreatment` column + the
+ * PrestaShop backfill).
+ */
+import type { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class BackfillOrderRecordTotalTaxTreatmentWoocommerce1870000000100
+  implements MigrationInterface
+{
+  name = 'BackfillOrderRecordTotalTaxTreatmentWoocommerce1870000000100';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`
+      UPDATE "order_records" o
+      SET "totalTaxTreatment" = 'inclusive'
+      FROM "connections" c
+      WHERE c."id" = o."sourceConnectionId"
+        AND c."platformType" = 'woocommerce'
+        AND o."totalTaxTreatment" IS NULL
+    `);
+  }
+
+  public async down(): Promise<void> {
+    // Deliberate no-op, matching 1841000000006's down() — this migration only
+    // SETs a value on a pre-existing column (added by 1870000000000), it
+    // never adds or drops a column, so there is nothing to reverse. Rolling
+    // back would also be unsound: it cannot distinguish a row this migration
+    // set from one a later, unrelated write also set to 'inclusive'.
+  }
+}
--- a/libs/core/src/orders/domain/order-analytics-projection.spec.ts
+++ b/libs/core/src/orders/domain/order-analytics-projection.spec.ts
@@ -35,7 +35,6 @@ describe('orderAnalyticsProjection', () => {
         placedAt: new Date('2026-08-01T09:00:00.000Z'),
         currency: 'PLN',
         taxTreatment: 'inclusive',
-        totalTaxTreatment: null,
         totalAmount: 32,
         totalTaxTreatment: null,
       });
--- a/libs/core/src/orders/domain/order-analytics-projection.spec.ts
+++ b/libs/core/src/orders/domain/order-analytics-projection.spec.ts
@@ -35,6 +35,7 @@ describe('orderAnalyticsProjection', () => {
         placedAt: new Date('2026-08-01T09:00:00.000Z'),
         currency: 'PLN',
         taxTreatment: 'inclusive',
+        totalTaxTreatment: null,
         totalAmount: 32,
         totalTaxTreatment: null,
       });
--- a/libs/core/src/orders/domain/types/order.types.ts
+++ b/libs/core/src/orders/domain/types/order.types.ts
@@ -367,7 +367,7 @@ export interface OrderTotals {
    * `taxTreatment`.
    *
    * WooCommerce has the identical structural split (net line prices, genuinely
-   * gross `total`) and does not yet set this field — tracked as #2836.
+   * gross `total`) and sets this field too (#2836), mirroring PrestaShop.
    */
   totalTaxTreatment?: PriceTaxTreatment;
 }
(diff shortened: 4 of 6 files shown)


norbert-kulus-blockydevs:
## Code Review: PR #2838

**Scope**: `libs/integrations/woocommerce/src/infrastructure/adapters/woocommerce-order-source.adapter.ts` (+9/-1), its spec (+2)

### Summary

Sets `totalTaxTreatment: 'inclusive'` in `mapTotals` so a WooCommerce order's `total` (genuinely gross) can be evaluated by an `orderTotalGross` sales-document rule condition, while leaving `taxTreatment: 'exclusive'` untouched for the line prices - mirroring the identical PrestaShop split from #2829/#2832, which this branch is stacked on.

### Verification

- Confirmed `OrderTotals.totalTaxTreatment` / `IncomingOrderTotals.totalTaxTreatment` already exist on the base branch (`2829-prestashop-tax-treatment`, tip commit `5c340f0e1`) with the exact fallback semantics this PR relies on (`totalTaxTreatment ?? taxTreatment`), and are already wired through `order-analytics-projection.ts`, `order-from-ready-snapshot.ts`, `OrderRecord`/ORM entity, `order-record.repository.ts`, and `sales-document-view.service.ts`. This PR is genuinely the last missing piece - no gaps left downstream.
- Confirmed the doc comment on `OrderTotals.totalTaxTreatment` explicitly names WooCommerce as the still-open case (#2836) and cites this exact fallback contract - the PR fulfills that comment precisely.
- Grepped the WooCommerce package for other `taxTreatment`/`totals` assertions - only one test (`should compute subtotal from line_items total when fee_lines are present`) asserts totals shape, and it's the one updated. No other adapter code path reads `mapTotals` output that would need a matching update.
- `convertGrossToNet` / net-sales tax-rate path concerns (the reason `taxTreatment` itself couldn't just be flipped) are WooCommerce-adapter-local - checked there's no `convertGrossToNet`-equivalent destination-side pinning in this file that reads `taxTreatment` and would be affected; this is a pure `OrderSourcePort` (ingestion) adapter, not `OrderProcessorManagerPort`, so that class of concern doesn't even apply here.
- Comment quality: both new comments explicitly say *why* (`do NOT flip it to describe total`, `asserted independently of taxTreatment`) rather than restating the diff - matches `engineering-standards.md`'s comment guidance.

### Assessment

✅ **Approve** - correct, minimal, exactly mirrors the established and reviewed PrestaShop pattern, test coverage matches the one place totals shape is asserted, and the doc-comment cross-references are consistent both directions (core type comment → #2836, adapter comment → #2829/#2836). No blocking or important issues found.

One non-blocking note for the PR author to consider before merge (not a defect): this PR is stacked on `2829-prestashop-tax-treatment` and will need retargeting to `main` once that merges, per the PR description's own note - just flagging it doesn't get lost given the base branch already has 3 commits queued ahead of it.

norbert-kulus-blockydevs:
Thanks for the review — no blocking or important findings to act on, so no code changes needed here.

The one non-blocking note (retarget to `main` once #2832 merges) isn't actionable yet since #2832 is still open; will retarget as soon as it lands, per the PR description.

piotrswierzy (approved):
## Code Review — #2838 (WooCommerce `totalTaxTreatment`)

Reviewed against its own base (`2829-prestashop-tax-treatment`, PR #2832); #2832's changes are out of scope here.

### 1. Context

Two files, +11/−1: `mapTotals` in `woocommerce-order-source.adapter.ts` gains `totalTaxTreatment: 'inclusive'` beside the unchanged `taxTreatment: 'exclusive'`, plus one assertion in the adapter spec. The seam it populates (`IncomingOrderTotals.totalTaxTreatment` → `OrderTotals` → `toSalesDocumentOrderFacts`'s `totalTaxTreatment ?? taxTreatment`) is entirely #2832's; this PR is the WooCommerce arm the #2832 migration docblock explicitly defers to #2836.

### 2. Findings

**Correctness — the platform fact checks out.** WooCommerce REST `order.total` is the grand total and always includes `total_tax`; `line_items[].price` is `subtotal / quantity`, ex-tax, and stays ex-tax regardless of the store's `woocommerce_prices_include_tax` setting. So `'inclusive'` on the total is unconditionally true, and the split mirrors PrestaShop's `total_paid_tax_incl` vs `order_details.product_price` exactly. `mapTotals` (`woocommerce-order-source.adapter.ts:257-282`) already sources `total` from `order.total` and `tax` from `order.total_tax` separately, so the new label describes the value actually mapped.

**Positive — `taxTreatment` is left alone, and the comment says why.** The tempting one-line "fix" here is flipping `taxTreatment` to `'inclusive'`, which would silently corrupt the ADR-063 net-sales tax-rate path (`deriveNetLineAmount` + its SQL twins). The in-place comment names that consequence rather than just restating the code — correct per `engineering-standards.md` ("comments explain *why*"), and it is the fact a future maintainer needs before touching this literal.

**Positive — consumer audit is real.** `grep -rn "taxTreatment" libs/integrations/woocommerce/src` returns only the two lines this PR touches, so no WooCommerce-side reader can be disturbed by the additive field. The field is optional on `IncomingOrderTotals`, so no other adapter is affected.

**[IMPORTANT]** — `apps/api/src/migrations/1870000000000-add-order-record-total-tax-treatment.ts` (on the base branch)

The #2832 migration backfills `'inclusive'` for `platformType = 'prestashop'` only, and its docblock states WooCommerce is "tracked as a follow-up (#2836)". This PR closes #2836 and ships no backfill — so once both land, that comment points at a closed issue that did not do the thing it implies, and every pre-existing WooCommerce `order_records` row stays `totalTaxTreatment IS NULL` → falls back to `'exclusive'` → an `orderTotalGross` rule condition still cannot match a historic WooCommerce order.

To be clear on severity: the *behaviour* is defensible and matches the migration's own stated principle — NULL means "same as `taxTreatment`", the pre-existing net-priced reading, never a guess — and the PR description says the issue left the backfill discretionary. What is not defensible is leaving the migration comment asserting a follow-up that closed without one. Either (a) add the one-line WooCommerce arm to that same `UPDATE` (it is idempotent by `WHERE "totalTaxTreatment" IS NULL`, and the platform-scoped join precedent is already there), or (b) amend the docblock to say history is deliberately not backfilled for WooCommerce and why. (a) is two words of SQL and closes the gap for real; (b) is acceptable if there is a reason not to assert on history that this review cannot see. Since the migration lives on the base branch, this is a note for whichever PR lands last rather than a change request against this diff.

**[SUGGESTION]** — `woocommerce-order-source.adapter.spec.ts:465`

The added assertion sits in the existing totals test alongside `taxTreatment`, which is the right place. One cheap strengthening: the two are only meaningful *together* (the whole point is that they disagree), and nothing in the suite fails if a future edit collapses both to the same value in the same commit. A one-line `expect(result.totals.totalTaxTreatment).not.toBe(result.totals.taxTreatment)` would pin the split itself rather than two independent literals. Optional — the two explicit assertions already cover the regression that matters.

### 3. Standards / architecture

- Layer placement correct: this is a platform fact declared where the platform is known (`libs/integrations/woocommerce/.../infrastructure/adapters/`), and no domain logic moves into the integration. CORE↔Integration boundary intact.
- No new types, no `any`, no migration required by this diff (the column is #2832's), no security/idempotency surface.
- Test placement follows `docs/testing-guide.md` (`__tests__/*.spec.ts` colocated); no integration test is warranted for a literal in a mapper.

### 4. Final Assessment

**Summary.** A minimal, correctly-scoped platform-fact declaration that completes the WooCommerce arm of the #2829/#2832 gross-total split. The platform claim is accurate, the dangerous adjacent edit is explicitly warned against in-place, and the consumer audit is verifiable. The one thing worth acting on is not in this diff: the base migration's docblock will point at a closed follow-up that shipped no history backfill.

**Merge readiness:** 🔄 **Approve with changes** — mergeable as-is; address the IMPORTANT item (backfill arm *or* corrected docblock) on whichever of #2832 / #2838 lands last, and re-target this branch to `main` after #2832 merges.

**Priority fixes:**
1. Add the WooCommerce arm to `1870000000000-add-order-record-total-tax-treatment.ts`'s `UPDATE`, or amend its docblock so it does not imply #2836 backfilled history.
2. (Optional) assert the `taxTreatment` / `totalTaxTreatment` divergence explicitly in the spec.

norbert-kulus-blockydevs:
Addressed both review findings:

1. **IMPORTANT (migration docblock/backfill)**: folded the WooCommerce arm into `1870000000000-add-order-record-total-tax-treatment.ts`'s existing `UPDATE` (`platformType IN ('prestashop', 'woocommerce')`) instead of leaving a follow-up — the statement was already platform-scoped and idempotent, so a second migration for one more value would be ceremony. Docblock updated to describe both platforms and drop the "deliberately not WooCommerce" note.
2. **SUGGESTION (spec)**: added `expect(result.totals.totalTaxTreatment).not.toBe(result.totals.taxTreatment)` to pin the divergence itself, not just the two literals.

Also fixed a pre-existing spec/source drift found while running the gate: `order-analytics-projection.spec.ts`'s "derives all 4 scalars" test predates `totalTaxTreatment` being added to the projection output and never expected it — unrelated to this diff's scope but was failing the suite, so fixed inline.

Rebasing onto main is pending #2832 merging first (this branch is based on it).

piotrswierzy (changes requested):
## Re-review — #2838 @ `0794274`

Head moved `99feb66` → `0794274` by one commit, *"fix(orders): backfill WooCommerce totalTaxTreatment and pin the split in tests (review)"*. Reviewed **only this PR's own diff** against its declared base `2829-prestashop-tax-treatment` (#2832).

**Both of my previous findings are resolved, and the IMPORTANT one was resolved the stronger way I hoped for.** The remaining items are entirely mechanical — but two of them are the kind that pass CI and lose the fix anyway, so this is REQUEST_CHANGES rather than an approve-with-notes.

---

**[IMPORTANT] — the migration backfilled PrestaShop only, so #2836 would close having changed no history — RESOLVED, via option (a).**

`0794274` folds the WooCommerce arm into `1870000000000-add-order-record-total-tax-treatment.ts` rather than shipping a second migration:

```sql
-        AND c."platformType" = 'prestashop'
+        AND c."platformType" IN ('prestashop', 'woocommerce')
```

and rewrites the header to describe both platforms, dropping the "deliberately does NOT touch WooCommerce" paragraph and stating why one pass covers both ("the `UPDATE` is already platform-scoped and idempotent, so a second migration to add one more platform value would be pure ceremony"). I verified the statement is still guarded by `WHERE "totalTaxTreatment" IS NULL`, so it stays idempotent, and the `up()`/`down()` + `IF NOT EXISTS`/`IF EXISTS` shape is untouched. The reasoning is sound and the two-word SQL edit is exactly what I asked for.

**[SUGGESTION] — pin the divergence, not two independent literals — RESOLVED.**

```ts
expect(result.totals.totalTaxTreatment).not.toBe(result.totals.taxTreatment);
```

with a comment saying precisely what it exists to catch ("collapsing both in one future edit"). That is the assertion that survives the refactor the two literals do not.

**The adapter change itself is unchanged and still correct.** `mapTotals` sets `totalTaxTreatment: 'inclusive'` beside the untouched `taxTreatment: 'exclusive'`; the comments name the dangerous adjacent edit (ADR-063 net-sales path, `deriveNetLineAmount`) rather than restating the code, per `engineering-standards.md`. `grep -rn "taxTreatment" libs/integrations/woocommerce/src` still returns only these lines. CORE↔Integration boundary intact — a platform fact declared where the platform is known.

---

### 2. THE CROSS-PR BACKFILL ANSWER (stated on both PRs)

**Yes — a WooCommerce backfill now ships, and #2832 and #2838 agree.** #2832 (`5686d31`) still carries the PrestaShop-only `UPDATE` plus a docblock deferring WooCommerce to #2836; #2838 (`0794274`) fulfils that deferral **by editing the same migration**. There is no new migration, so there is no timestamp to check; `1870000000000` remains strictly greater than `origin/main`'s tail (`1869000000900-add-automation-run-retry-attempt.ts`) and unique, and the edited `UPDATE` remains idempotent.

**[IMPORTANT] — but the fix is only as durable as the merge order, and `docs/migrations.md` rule 4 is "Never edit executed migrations."**

Editing a peer PR's unmerged migration is legal. What is not safe is the window it opens:

> If **#2832 merges and `1870000000000` runs anywhere** — demo, staging, any environment deployed from `main` — before this PR lands, TypeORM records it in the `migrations` table and **will never re-run it**. The `IN ('prestashop','woocommerce')` edit then arrives on a migration that has already executed, the WooCommerce arm never runs, and every historic WooCommerce `order_records` row keeps `totalTaxTreatment IS NULL` → falls back to `'exclusive'` → **still unmatched by an `orderTotalGross` rule**. That is precisely the outcome this commit exists to prevent, and it would fail *silently*: green CI, green migration state, no history changed.

The behaviour in that failure mode is safe (NULL = "same as `taxTreatment`" = the pre-existing net-priced reading, never a guess) — it is the *claim* that breaks, exactly as in my last round.

Two acceptable resolutions:

- **(a)** Merge #2832 and #2838 back to back, with no deploy or `migration:run` in between. Cheap, but it is a process guarantee, and process guarantees are the ones that get skipped on a Friday.
- **(b)** Revert this migration edit and ship a **new** migration on this PR (prefix `> 1870000000000`, e.g. `1870000000100`) carrying only the WooCommerce arm — same `WHERE "totalTaxTreatment" IS NULL` guard, so it is idempotent and correct **whatever order the two land in or how long apart**.

I would take (b). The "second migration for one more platform value is pure ceremony" argument is right about the SQL and wrong about the risk: the ceremony is what buys order-independence, and this repo deploys from `main`. (a) is acceptable if you are merging both today and can state that nothing deploys in between.

---

### 3. Merge mechanics — two concrete blockers I verified

**[BLOCKING] The branch cannot merge: it is stacked on the *pre-force-push* #2832.** `git merge-base 5686d31 0794274` → `5abcc57`, i.e. this branch still sits on #2832's old commits (`45efbec3a` / `5c340f0e1`), while #2832 force-pushed a rebase onto current `main`. GitHub reports `mergeable_state: dirty`. `git merge-tree` gives **two conflicts**, both in files where the resolution is semantic, not textual:

- `apps/api/src/migrations/1870000000000-…ts` — add/add. Both sides rewrote the same docblock (yours: two platforms; #2832's: the `down()` explanation). **Resolve by intent**: keep the two-platform header *and* #2832's `down()` paragraph, then re-grep that the `UPDATE` still reads `IN ('prestashop', 'woocommerce')` and still carries `AND o."totalTaxTreatment" IS NULL`.
- `libs/core/src/orders/domain/types/order.types.ts` — the `OrderTotals.totalTaxTreatment` docblock (yours: "sets this field too (#2836)"; base: "does not yet set this field — tracked as #2836"). Yours is the correct end state.

**[IMPORTANT] One file auto-merges *without* a conflict and produces a duplicate object key.** This is the one worth catching before you rebase, because git will not tell you:

`libs/core/src/orders/domain/order-analytics-projection.spec.ts`. Your commit adds `totalTaxTreatment: null` after `taxTreatment: 'inclusive'`, describing it as fixing "a pre-existing spec/source drift". I checked: that drift existed **only on your stale base**. #2832's own commit `985ec3fb8` had already fixed it — `git show 3295640:…` shows `totalTaxTreatment: null` present after `totalAmount: 32` at #2832's *previous* head. The two insertions are at different offsets, so git merges both cleanly and the result is:

```ts
expect(deriveOrderAnalyticsScalars(order)).toEqual({
  placedAt: new Date('2026-08-01T09:00:00.000Z'),
  currency: 'PLN',
  taxTreatment: 'inclusive',
  totalTaxTreatment: null,   // ← yours
  totalAmount: 32,
  totalTaxTreatment: null,   // ← #2832's
});
```

(verified against the merge-tree blob `7db04c28f`). Lint's `no-dupe-keys` should catch it, but the safe move is to **drop your line during the rebase** — #2832 owns that spec change, and your PR does not need it.

**[BLOCKING] Zero CI has run on `0794274`.** `get_check_runs` returns `total_count: 0` — no Build, no Lint, no Test, no Type Check, nothing, presumably because the base branch is unmergeable. Nothing about this head has been verified by CI, including the migration edit and the duplicate key above.

---

- Layer placement correct; no domain logic in the integration; no new types, no `any`.
- Test placement follows `docs/testing-guide.md` (colocated `__tests__/*.spec.ts`); no integration test warranted for a mapper literal.
- Migration conforms to `docs/migrations.md` § Timestamp uniqueness invariant — the only rule in tension is rule 4, addressed in § 2.
- Comment discipline holds: every added comment states a constraint on future edits rather than restating the diff.

---

### 5. Final assessment

**Summary.** The substance is done and done well — both prior findings are resolved, the WooCommerce arm now genuinely backfills history, and the divergence assertion is the right one. What is left is entirely mechanical, but two of the three items are silent: an unmergeable stale base, an auto-merging duplicate key that no conflict marker will show you, and a migration edit whose effect depends on merge order rather than on the SQL.

**Merge readiness: ❌ Request changes** — not on the code, on the branch.

**Priority fixes:**
1. Decide the migration-durability question in § 2. I recommend **(b)**: a new `> 1870000000000` migration owning only the WooCommerce arm, so the backfill is order-independent. If you keep the in-place edit, say explicitly here that #2832 and #2838 will land back to back with no deploy in between.
2. Rebase onto #2832's current head `5686d31` (or onto `main` once it lands). Resolve both conflicts by intent per § 3, and **drop your `order-analytics-projection.spec.ts` line** — verify no duplicate `totalTaxTreatment` key survives.
3. Get a full green CI run on the rebased head; nothing here has been checked by CI yet.
4. After the rebase, re-grep the migration to confirm `IN ('prestashop', 'woocommerce')` and the `IS NULL` guard both survived the resolution. This is the assertion the whole PR turns on, and it lives inside one of the two conflicted files.

norbert-kulus-blockydevs:
Addressed both items from the second review round (commit `618125d`, on top of the previous `0794274`).

**1. "The migration backfilled PrestaShop only... resolved, via option (a)" but then flagged as order-dependent** - you were right that (a) was the weaker fix. I switched to your recommended option (b): reverted `1870000000000` to its original PrestaShop-only form (matching the current `2829-prestashop-tax-treatment` head) and added a new, standalone migration `1870000000100-backfill-order-record-total-tax-treatment-woocommerce.ts` carrying only the WooCommerce arm, guarded by the same `WHERE "totalTaxTreatment" IS NULL` idempotency check. It now runs correctly and backfills WooCommerce history regardless of whether `1870000000000` has already executed elsewhere, so the fix no longer depends on merge order or on the two PRs landing back to back.

**2. BLOCKING - branch stacked on the pre-force-push #2832** - rebased onto the current `2829-prestashop-tax-treatment` head (`5686d31e6`). `git merge-base` between the two branches now equals that head exactly, confirming the branch is current.

**3. IMPORTANT - silent duplicate `totalTaxTreatment` key in `order-analytics-projection.spec.ts`** - the rebase surfaced exactly the duplicate you predicted (my line at one offset, #2832's own fix at another). Removed the redundant line; the spec now asserts the value once, matching what #2832 already added independently.

**4. BLOCKING - zero CI had run** - this is resolved as a consequence of the rebase fixing the stale/unmergeable base; the branch is now `mergeable: MERGEABLE` on GitHub and CI should run on the new head.

Verification run locally (scoped, not the full monorepo suite):
- `pnpm --filter @openlinker/integrations-woocommerce lint` / `type-check` / `test -- woocommerce-order-source.adapter.spec.ts` - clean, 35/35 pass
- `pnpm --filter @openlinker/core lint` / `type-check` / `test -- order-analytics-projection.spec.ts sales-document-view.service.spec.ts order-to-sales-document-order-facts.mapper.spec.ts` - clean, 45/45 pass
- `pnpm --filter @openlinker/api type-check` - clean
- `node scripts/check-migration-timestamps.mjs` - OK (160 migrations, ordering vs `origin/main` checked)

Nothing from this review round was skipped.

piotrswierzy (approved):
## Re-review — #2838 @ `618125d`

Head moved `0794274` → `618125d` (force-push). Reviewed this PR's own diff against the #2832 commit it actually sits on (`5686d31`), which is 4 files, +77/−2.

**Every blocker from my last round is resolved, and the migration one was resolved the way I recommended.** Approving.

---

### 1. THE CENTRAL QUESTION — answered

**Yes: there is now a separate migration, and the merge-order hazard is gone.**

`apps/api/src/migrations/1870000000100-backfill-order-record-total-tax-treatment-woocommerce.ts` is new, and `git diff 32c7172..618125d -- apps/api/src/migrations/1870000000000-…ts` is **empty** — #2832's migration is byte-identical on both branches. The in-place `IN ('prestashop','woocommerce')` edit is fully reverted.

I verified all four properties I asked for:

| Property | Result |
|---|---|
| Timestamp > `1870000000000` | ✅ `1870000000100` |
| Timestamp > `origin/main` tail | ✅ tail is `1869000000900-add-automation-run-retry-attempt.ts`; `1869000000900 < 1870000000000 < 1870000000100` strictly ascending (`docs/migrations.md` rule 3) |
| Idempotent | ✅ `AND o."totalTaxTreatment" IS NULL`, plus the `c."platformType" = 'woocommerce'` scoping join — the `1841000000006` precedent |
| Working `down()` | ✅ **as a documented no-op**, and that is the correct answer here, not a gap |

On the `down()`: this migration only `SET`s a value on a column `1870000000000` owns, so there is nothing to reverse — and the comment names the reason a re-null would be *unsound* ("it cannot distinguish a row this migration set from one a later, unrelated write also set to `'inclusive'`"). I checked the cited precedent and it is real: `1841000000006-backfill-order-records-tax-treatment.ts`'s `down()` is the same deliberate no-op with the same argument. Contrast with `1870000000000`, which *adds* the column and therefore correctly drops it. The two `down()`s differ because the two migrations do different things — that is the right distinction, not an inconsistency.

The docblock also states the merge-order argument in full rather than just fixing the code, so the next person cannot innocently "simplify" it back into one migration. That is the comment discipline `engineering-standards.md` asks for.

**Consequence worth stating plainly:** the backfill is now order-independent. #2832 and #2838 may merge in either order, with any number of deploys and `migration:run`s in between, and every historic PrestaShop *and* WooCommerce `order_records` row ends up correctly stamped. No process guarantee is required any more.

---

### 2. The duplicate-key hazard — GONE, verified against the actual merge tree

Last round I flagged that `order-analytics-projection.spec.ts` would **auto-merge with no conflict** into a duplicate `totalTaxTreatment` key. That is fixed: the file is no longer in this PR's diff at all (`git diff 5686d31..618125d` touches 4 files, and that spec is not one of them) — #2832 owns it, as it should.

I confirmed it on the merged result rather than trusting the file list. `git merge-tree --write-tree 32c7172 618125d` → tree `bbd93d1`, and in that tree:

```
libs/core/src/orders/domain/order-analytics-projection.spec.ts
  39:  totalTaxTreatment: null,      ← exactly one occurrence in the object literal
```

Both migrations also survive the merge intact in that tree:
```
1870000000000…ts:  59:  AND c."platformType" = 'prestashop'
                   60:  AND o."totalTaxTreatment" IS NULL
1870000000100…ts:  50:  AND c."platformType" = 'woocommerce'
                   51:  AND o."totalTaxTreatment" IS NULL
```
That is priority-fix #4 from my last review, discharged.

---

### 3. `mergeable_state: dirty` — now ONE trivial conflict, and it is an artefact of the force-push

Still `dirty`, but for a much smaller reason. `git merge-base 32c7172 618125d` is `df46c641`, i.e. *before* either branch added the field — because #2832 force-pushed a rebase and this branch still descends from `5686d31`. So both sides "add" the same `OrderTotals.totalTaxTreatment` block and git reports an add/add conflict.

The single conflicted file is `libs/core/src/orders/domain/types/order.types.ts`, and the entire semantic difference is **one docstring line**:

```
- * gross `total`) and does not yet set this field — tracked as #2836.
+ * gross `total`) and sets this field too (#2836), mirroring PrestaShop.
```

Yours is the correct end state. Resolve by taking your line and keeping everything else; nothing else in that hunk differs. The migration conflict from last round is gone (both files are now distinct paths).

One thing I checked and want to record so nobody panics: `git diff 32c7172..618125d --stat` appears to *delete* `perf/openlinker-throughput/bootstrap.sh`, `docs/operations/perf-lab-stand.md` and revert `.gitignore`. **That is a base-skew artefact, not a revert in this PR.** #2832 rebased onto `608813dc` (which carries #2872, the perf stand); this branch is still on the pre-rebase base. Rebasing onto `32c7172` or `main` makes it vanish. Verified: the true own-diff (`5686d31..618125d`) touches only the 4 intended files.

---

### 4. CI — resolved, fully green

Last round: `total_count: 0`. Now, on `618125d`, all 9 checks **completed successfully**, including the two that matter here:

- **Test** ✅ success (17 min)
- **Integration Tests** ✅ success (32 min)
- Build, Lint, Type Check, Docker Build Smoke Test, PHP Unit Tests, Scaffolded Adapter Builds ✅

Blocker discharged.

---

### 5. The change itself — unchanged and still correct

`mapTotals` sets `totalTaxTreatment: 'inclusive'` beside the untouched `taxTreatment: 'exclusive'`. The platform fact holds: WooCommerce REST `order.total` is the grand total and always includes `total_tax`, while `line_items[].price` is `subtotal / quantity` ex-tax regardless of `woocommerce_prices_include_tax`. Both comments name the *consequence of the tempting adjacent edit* (flipping `taxTreatment` would corrupt the ADR-063 `deriveNetLineAmount` net-sales path) rather than restating the code.

The divergence assertion I suggested is in and carries its own rationale:
```ts
expect(result.totals.totalTaxTreatment).not.toBe(result.totals.taxTreatment);
```

Architecture: a platform fact declared where the platform is known (`libs/integrations/woocommerce/.../infrastructure/adapters/`); no domain logic crosses into the integration; CORE↔Integration boundary intact. No new types, no `any`. Test placement follows `docs/testing-guide.md`.

---

### 6. Final assessment

**Summary.** The substance was already right two rounds ago; this round fixes the mechanics, and fixes them properly rather than by promising a merge order. The standalone `1870000000100` migration is the durable answer — it makes the WooCommerce history backfill correct whatever order these two PRs land in — and the duplicate-key trap that no conflict marker would have shown you is gone because the spec change went back to its owner. CI is fully green.

**Before merging (mechanical only):**
1. Rebase onto #2832's current head `32c7172` (or onto `main` after #2832 lands) and resolve the single `order.types.ts` add/add conflict by keeping your "sets this field too (#2836)" wording. The apparent perf-file deletions disappear with the rebase.
2. Re-target to `main` once #2832 merges.

No further review needed from me after the rebase, provided CI stays green.

Issue #2856, opened 2026-09-03
[TASK] DX — Allegro upstream stub: two endpoints, 1+N cost model, multi-tenant by bearer token


norbert-kulus-blockydevs:
Child of #2840. **Week 2, ahead of #2846's PrestaShop half.** Split out of #2846 so the Allegro half can land first. Week 1 (#2841-#2845) needs no stub of any kind, so this is the first stub the programme actually needs; the lab stand (#2854) already declares a compose service for this image. #2846 keeps the PrestaShop stub and the shared pacing discipline.

## Problem / Context

Nothing in this repository is an HTTP stub server. A repo-wide grep for `createServer`, `express(`, `fastify(`, `msw` and `nock` across `apps/` and `libs/` returns **zero hits**. The file #2846 names as the reference implementation, `apps/e2e/tests/perf/resolve-stream-stub.ts`, is **not a server** - it is a `page.addInitScript` that monkey-patches `window.fetch` inside the browser (`:246`, `:255`, `:383-394`), for the stated reason at `:5-11` that `route.fulfill` cannot deliver an incremental body. So its *pacing and configuration shape* transfer; its transport does not.

Meanwhile two facts about the Allegro adapter decide almost the whole design, and one of them invalidates an acceptance criterion #2846 currently carries.

**The REST host is overridable. The OAuth host is not.** `libs/integrations/allegro/src/infrastructure/http/allegro-hosts.ts` models three hosts separately, and only two are per-connection:

| Host | Resolver | Per-connection override |
|---|---|---|
| REST API | `getAllegroRestApiBaseUrl(environment)`, `allegro-hosts.ts:69-79` | **yes** - `config.apiBaseUrl` |
| Web / OAuth (`/auth/oauth/token`) | `getAllegroWebBaseUrl(environment)`, `allegro-hosts.ts:50-60` | **no - hardcoded** |
| Image upload | `AllegroAdapterFactory.getDefaultUploadBaseUrl`, `allegro-adapter.factory.ts:181-191` | yes - `config.uploadBaseUrl` |

[...]

Issue #2854, opened 2026-09-03
[TASK] Infrastructure — provision the isolated "lab" measurement stand


norbert-kulus-blockydevs:
## Problem / Context

Epic #2840 needs a stand to measure on, and no environment in this repo can host it. Three concrete blocks, all verifiable today:

1. **`--scale worker=N` cannot run.** `docker-compose.demo.yml:98` pins `container_name: ${COMPOSE_PROJECT_NAME:-openlinker}-worker`, and Compose refuses to scale a service whose container name is fixed. Every multi-replica question in the epic (#2851's F4, #2852's F7) is blocked on that one line.

2. **Postgres cannot take three workers even once the name is gone.** `OL_DB_POOL_MAX` defaults to **40** per process (`libs/shared/src/database/database.module.ts:81`, const at `:40`), and **no compose file carries a `command:` override on the postgres service** - the only two `command:` keys in the tree are redis (`docker-compose.yml:81`) and migrate (`docker-compose.demo.yml:179`). So the server runs at the `postgres:17-alpine` image default of `max_connections = 100`, against `3 x 40 + 40 = 160`. `apps/worker/.env.example:513-515` already states that arithmetic and nothing enforces it. The failure mode is worse than a clean error: F4 lists "DB pool saturation" as one of its own four attribution candidates, so the stand's own misconfiguration would be recorded as a finding about OpenLinker.

3. **WooCommerce cannot be reached over cleartext.** `libs/integrations/woocommerce/src/infrastructure/http/woocommerce-http-client.ts:282` sends Basic auth only, and `:213` refuses to follow a non-https redirect because "cleartext would leak Basic-Auth credentials". The #2590 campaign worked around it with a hand-rolled `ol-wc-tls` nginx container plus a worker running `NODE_TLS_REJECT_UNAUTHORIZED=0`, both since removed (`perf/prestashop-baseline/README.md:178-180`). All eight flows F1-F8 are in scope for #2840, so WooCommerce is required rather than optional and that proxy has to become a declared service.

Two further gaps make "just copy demo" impossible:

- **`pg_stat_statements` and `auto_explain` are configured nowhere.** A repo-wide grep for `shared_preload_libraries|pg_stat_statements|auto_explain` returns zero hits. #2843's acceptance criterion "top 20 queries by total time" is unsatisfiable without them, and `shared_preload_libraries` needs a server restart rather than a reload.
[...]

Issue #2848, opened 2026-09-03
[TASK] DX — F2: stock propagation latency and fan-out convergence


piotrswierzy:
Child of #2840. **Week 3.**

## Problem / Context

Overselling is the single most common reason sellers distrust multichannel tools, and stock propagation is the core value proposition of the category. `inventory.propagateToMarketplaces` is the highest-frequency event-paced job in the system and, per `docs/architecture-overview.md`, has **no cron of its own** - on a master with no stock webhook the inventory sweep is the only thing that discovers drift at all. If propagation falls behind, stock is wrong on every marketplace and **nothing re-drives it**.

What is known: #2609 measured the job growing ~145 jobs/h faster than it drained on the demo stack - 15 066 rows backlogged from ordinary operation - and traced it to a synthetic all-zero `connectionId` collapsing every stock write in the installation into one lane scope. The job now carries the real master connection and `fan-out` was raised from 1/1 to **8/4**.

**That new cap has been measured exactly once, and not for this flow.** `perf/prestashop-baseline/results-D-2026-08-28.md` reports net **0.0 jobs/h** over a 1 h window (and -0.5 over 2 h) at 100 000 products, one replica, `bulk=12/8 fan-out=8/4`, with `GET /connections/:id/sync-status` reporting `arrivalRatePerHour: 348`, `drainRatePerHour: 380`, `alerting: false`. That is the *catalogue sweep* driving the lane, not stock churn, so it is the right baseline to compare against and not an answer to this question.

## Proposed Solution

**The driver is stub-control, not a row writer.** There is no entry point for writing inventory rows through the service path: `InventoryService.setInventory` has exactly **one** production caller in the whole tree, `master-inventory-sync.service.ts:224`. No HTTP endpoint, no other service. So F2 changes stock **at the #2846 stub** and lets `master.inventory.*` pull it, which is also the only shape that exercises the guard and dedupe behaviour being measured.

[...]

Issue #2847, opened 2026-09-03
[TASK] DX — F1: order ingestion latency and throughput


piotrswierzy:
Child of #2840. **Week 2.** The headline number of the whole programme, and the body of #1134.

## Problem / Context

*"How fast can OpenLinker load orders from a channel?"* is the first question a prospective user asks, and the honest answer today is **two samples**: `marketplace.order.sync` mean attempt 47 178 ms over 2 samples (`results-D`), taken on a stand whose worker logged at `debug` with full response bodies - which that campaign itself flags as inflating OL-side timings. It is an upper bound, not a figure.

### What the pipeline's shape already tells us

Five stages: marketplace journal delay -> poll wait -> poll job -> child queue wait -> `marketplace.order.sync`.

Poll cadences are constants in the plugins (**measured**, all under `src/infrastructure/scheduler/`):

| Source | Cadence | Page limit | Retunable without a code change? | Mean poll wait |
|---|---|---|---|---|
| Allegro | `*/1 * * * *` (shortened from 5 min by #2620, `allegro-scheduler-tasks.ts:108-111`) | 100 (`:123`) | yes - `OL_ALLEGRO_POLL_INTERVAL_CRON` | **30 s** |
| Erli | `*/5 * * * *` (`erli-scheduler-tasks.ts:58`) | 200 (`:65`), **advisory only** | **no - hardcoded literal** | 2.5 min |
| WooCommerce | `*/5 * * * *` (`woocommerce-scheduler-tasks.ts:20`) | 100 (`:28`) | **no - hardcoded literal** | 2.5 min |
| PrestaShop | `0 */10 * * * *` (`prestashop-scheduler-tasks.ts:58-61`) | 100, `OL_PRESTASHOP_POLL_PAGE_LIMIT` | yes - `OL_PRESTASHOP_POLL_INTERVAL_CRON` | webhook-primary; poll is the backstop |

So on an idle system, poll wait is a **first-order term** comparable to processing. It is *nearly* the cheapest thing on the list to tune, and the qualification is load-bearing:

- **Only Allegro and PrestaShop read a cron env var.** WooCommerce and Erli pass a string literal into `cronExpression`, so retuning either needs a code change and a release. Do not promise an operator a knob that does not exist.
[...]

PR #2872, opened 2026-09-03
feat(perf): stand bootstrap for the measurement programme

### LINKED ISSUE #2860: [TASK] DX — stand bootstrap: shop credentials, module install, connections and the CONNECTION_ID values
Child of #2840. **Week 1**, and a hard prerequisite of #2842, #2844, #2846, #2847 and #2848.

## Problem / Context

#2854 provisions the `lab` stand and states, correctly, that a full reset destroys more than the database:

> `down -v` also destroys `prestashop-data` and `woocommerce-data`, so both shops reinstall and **the PrestaShop Webservice key, the WooCommerce REST keys and both OpenLinker connections must be recreated by hand** - nothing automates that.

Two claims in this programme cannot both be true while that holds: *"a wipeable stand whose database is zeroed before each run"* (#2854) and *"`run-all` executes the full campaign unattended and resumably"* (#2840, #2845). An unattended eight-hour campaign cannot contain a manual browser walkthrough.

Four concrete gaps, each verifiable today:

1. **The PrestaShop Webservice key** is created by hand in the back office (`docs/one-command-demo-setup-guide.md:246-265`), and the order path additionally needs `specific_prices` and `product_options` / `product_option_values` grants - both discovered the hard way by #2590 (`perf/prestashop-baseline/README.md`).
2. **The WooCommerce REST consumer key and secret** are documented **nowhere**. `docs/one-command-demo-setup-guide.md` has ten sections and none of them covers WooCommerce, yet #2854 gates on "a WooCommerce connection test passes against `https://wc-tls`" and #2847 gains a real-WooCommerce destination arm.
3. **The PrestaShop OL module is never installed.** Its source is bind-mounted (`docker-compose.yml:278`) but no post-install script installs or enables it - grepping `docker/prestashop/post-install/` for `openlinker` or `module:install` returns nothing. Because the `line_prices` capability is **learned from a previous import response** (`libs/integrations/prestashop/src/infrastructure/adapters/prestashop-order-processor-manager.adapter.ts:461-465`), an uninstalled module silently puts every order create on the 27-request legacy path instead of 7 - the 3.4x that #2846 warns "would make the destination look cheap".
4. **The OpenLinker connections themselves.** #2846's acceptance criterion says `stand-up.sh` "**repoints** connection config", which presumes the connections exist. #2848 needs at least two `InventoryMaster` connections for per-scope lane accounting to be exercised at all; #2856 needs two Allegro connections differing only by `accessToken`. No issue creates any of them. And every perf script reads `CONNECTION_ID`, whose default `perf/prestashop-baseline/README.md:47` calls out explicitly: *"The default is a uuid from one machine; left alone on any other stack the scripts measure a connection that does not exist."*

## Proposed Solution

A `bootstrap.sh` under `perf/openlinker-throughput/`, called by #2841's `stand-up.sh`, that takes a freshly reset `lab` stand to a measurable state with no browser and no manual step, and **emits the ids the harness needs** rather than expecting an operator to paste them.

- **PrestaShop**: create the Webservice key and its resource grants by writing `ps_webservice_account` plus `ps_webservice_permission` rows directly through `docker exec` into MySQL, the same technique `perf/prestashop-baseline/seed-products.sh` already uses for the catalogue. Install and enable the OL module via `php bin/console prestashop:module install openlinker`. There is no fallback: a `ps_module` row plus a hook registration does not create the `ps_carrier` row `discoverDynamicCarrierId` requires (see below), so if the console path is unavailable in the shipped image the correct response is to find a working install path, not to substitute one that leaves order create permanently broken.

  **Module install is a correctness precondition, not only a cost-accuracy one.** `discoverDynamicCarrierId` runs first and unconditionally on every order create (`prestashop-order-processor-manager.adapter.ts:375`, body at `:1197-1229`) and throws `PrestashopOlCarrierMissingException` at `:1207` when no `ps_carrier` row with `external_module_name='openlinker', active=1, deleted=0` exists. That row is created only inside the module's own `install()` (`apps/prestashop-module/openlinker/openlinker.php:190` calling `installDynamicCarrier()` at `:1135`) - so on a stand where the module was never installed, every single order create throws before F1 measures anything, not merely a 27-vs-7-request difference. `bootstrap.sh` verifies the carrier row directly (`SELECT id_carrier FROM ps_carrier WHERE external_module_name='openlinker' AND active=1 AND deleted=0`), as a precondition distinct from - and prior to - checking that a real import response advertises `line_prices`.

  **The seeded products also need a resolvable tax rate.** Independently of `OL_TAX_RATE_STRICT_ENABLED`, the adapter converts gross to net whenever `order.totals.taxTreatment !== 'exclusive'` (`:664`), which fires on every Allegro-sourced stub order since the Allegro mapper always emits `taxTreatment: 'inclusive'` (`allegro-order-source.adapter.ts:820`). It then resolves the destination product's own PrestaShop tax rate and throws the non-retryable `PrestashopTaxRateUnknownException` (`:709`) if that resolution fails. `perf/prestashop-baseline/seed-products.sh:92` clones the catalogue from a template row wholesale, so this depends entirely on that template's own `id_tax_rules_group` - `bootstrap.sh` must record that value and verify it is non-zero, or every seeded order fails the same way whatever the carrier and module state is.

  **The webhook secret must also exist.** PrestaShop's `importorder` create path reads the per-connection secret at `ref = webhook-secret:{connectionId}` from `integration_credentials` (`webhookSecretRef`, `webhook-secret-provider.port.ts:71`) and throws if it is absent (`credentials-webhook-secret.adapter.ts:67-69`). Nothing today writes that row automatically - it is written only by the rotate/set HTTP endpoints (`webhook-secret.service.ts`), invoked from `POST /connections/:id/webhook-secret` or the install-webhooks flow. `bootstrap.sh` must call one of those for the PrestaShop connection, or the first order create fails on a precondition that has nothing to do with the carrier, the module or the tax rate.
- **WooCommerce**: create the REST consumer key/secret via `wp-cli` inside the container (`wp wc rest-api key create`), and repoint `siteurl` / `home` at `https://wc-tls` so WordPress stops emitting cleartext redirects the client refuses (`libs/integrations/woocommerce/src/infrastructure/http/woocommerce-http-client.ts:210-215`).
- **OpenLinker connections**: create them over the API with the bootstrap admin's token - one PrestaShop, one WooCommerce, two Allegro-stub tenants, and at least two `InventoryMaster` scopes - and write the resulting ids to a `stand-ids.env` the harness sources. Every Allegro connection needs `config.environment` set (mandatory, `@IsIn(AllegroEnvironmentValues)`, `allegro-connection-config.dto.ts:139-141`) or the create call 400s before anything else is reachable; Allegro registers no credentials-shape validator, so a malformed credential payload passes create and only fails later at first adapter construction - verify each Allegro connection with `POST /connections/:id/test` (which reaches the stub's `/me`) rather than assuming a successful create implies a working connection.
- **Offer mappings for the Allegro connections**: per #2856's seeded-mapping contract, `OrderIngestionService` resolves each order line's `offer.id` against `identifier_mappings` (`entityType='Offer'`, `externalId=offer.id`, `connectionId`) to a live, non-stale `ProductVariant`, and throws a retryable `MissingOrderItemMappingError` on any miss (`order-item-ref-resolver.service.ts:56-79`). Without one `identifier_mappings` row per stub offer id per Allegro connection, every stub order dies `awaiting_mapping` after exhausting its retry attempts and F1/F2 measure nothing. `bootstrap.sh` must seed these rows against the product pool it also creates for PrestaShop/WooCommerce, keyed per connection since the mapping's uniqueness is `(entityType, platformType, connectionId, externalId)` (`identifier-mapping.orm-entity.ts:24-26`).
- **`enabledCapabilities` on the Allegro connections**: `marketplace.offers.sync` declares `platformType: 'allegro'` with no `requiredCapability` field (`allegro-scheduler-tasks.ts:130-149`), unlike `marketplace.orders.poll` and both returns tasks, so it is enqueued regardless of which capabilities a connection declares. Omitting `OfferManager` from `enabledCapabilities` converts a clean 404 into a retryable `CapabilityNotEnabledException`, burning the full attempt ladder on every tick - state the chosen `enabledCapabilities` set explicitly rather than leaving it at whatever the create call defaults to.
- **Idempotent**: a second run against an already-bootstrapped stand changes nothing and re-emits the same ids.

## Classification

**Type**: DX
**File(s)**: `perf/openlinker-throughput/bootstrap.sh` (new), `perf/openlinker-throughput/stand-ids.env` (generated, git-ignored)

## Dependencies

- #2854 (the stand exists and both shops are healthy)
- #2841 (`stand-up.sh` calls this and consumes `stand-ids.env`)

## Assumptions

- The shipped PrestaShop image has either `bin/console` or a writable `ps_module` path. If neither works, the fallback is a pre-baked volume snapshot taken once and restored on reset - which trades reproducibility for speed and must be stated as such.
- Allegro-stub tenants are distinguished by bearer token (#2856), so the two connections differ only in credentials and share one `apiBaseUrl`.
- The module-install step is the one that most changes a measured number, so it gets its own verification rather than being assumed successful.

## Acceptance Criteria

- [ ] `bootstrap.sh` takes a stand freshly reset with `down -v` to a state where every perf scenario can run, with **no browser step and no manual paste**
- [ ] It is idempotent: a second run changes nothing and re-emits identical ids
- [ ] `stand-ids.env` carries every `CONNECTION_ID` the scripts read, and #2841's `lib.sh` sources it rather than requiring an operator-supplied value
- [ ] **The OL module's `ps_carrier` row is verified directly** (`external_module_name='openlinker', active=1, deleted=0`) as a precondition of every scenario, before any order create - its absence throws `PrestashopOlCarrierMissingException` on attempt 1 of every single order, not merely a cost-model degradation
- [ ] **The PrestaShop OL module is additionally verified advertising `line_prices`** by observing a real import response, not by checking that a file exists - an uninstalled or non-advertising module is a silent 3.4x cost difference
- [ ] The seeded catalogue's template product resolves a non-zero `id_tax_rules_group`, recorded in the manifest - an unresolvable tax rate throws the non-retryable `PrestashopTaxRateUnknownException` on every order, independent of `OL_TAX_RATE_STRICT_ENABLED`
- [ ] The PrestaShop connection's webhook secret (`ref = webhook-secret:{connectionId}`) is created via the rotate/set endpoint before the first order create - its absence throws before the `importorder` HMAC request is ever signed
- [ ] The Webservice key carries the `specific_prices`, `product_options` and `product_option_values` grants #2590 found were missing
- [ ] A WooCommerce connection test passes against `https://wc-tls`, and the `siteurl` / `home` repoint is part of the script rather than a runbook note
- [ ] At least two `InventoryMaster` connections exist, so #2848's per-scope accounting is exercised rather than degenerating into the #2609 defect it is measuring
- [ ] Two Allegro-stub tenants exist, differing only by `accessToken`
- [ ] Every created id and credential reference is recorded in the run manifest, so a later reader knows which connection a figure was taken against
- [ ] Documented in #2854's runbook as the step that follows a reset

## Related

- **#2840** - the epic. **#2854** - the stand. **#2841** - calls this.
- **#2846** / **#2856** - both assume connections exist to repoint.
- **#2847** - the WooCommerce destination arm needs the WC credentials this creates.
- **#2848** - needs two `InventoryMaster` scopes.

norbert-kulus-blockydevs:
Closes #2860. Child of #2840, and a hard prerequisite of #2842, #2844, #2846, #2847 and #2848.

## What this is

`perf/openlinker-throughput/bootstrap.sh` takes a freshly reset measurement stand to a state where every perf scenario can run, with no browser step and no manual paste, and emits the ids the harness needs into `stand-ids.env`.

Two claims the programme makes cannot both be true while any of this is manual: *"a wipeable stand whose database is zeroed before each run"* (#2854) and *"`run-all` executes the full campaign unattended and resumably"* (#2845).

## Why four of the steps are correctness preconditions, not conveniences

Without them the stand looks healthy and every measured order fails - and in most cases silently, because a destination-create rejection is reduced to a message on `order_records.syncStatus` without rethrowing, so the job reports `outcome: 'ok'` and the queue drains.

| Precondition | Consequence if absent |
|---|---|
| OL PrestaShop module installed **via `install()`** | `discoverDynamicCarrierId` runs first and unconditionally on every order create and throws `PrestashopOlCarrierMissingException`. The carrier row is created inside that hook, so a `ps_module` row alone is not enough |
| `Offer` identifier mappings per Allegro tenant | every stub order fails item resolution, persists `awaiting_mapping`, never reaches a destination create, and burns ten retry attempts over roughly 30 hours |
| A tax rules group on every seeded product | the gross-to-net conversion resolves the destination product's own rate and throws the non-retryable `PrestashopTaxRateUnknownException`. Unrelated to `OL_TAX_RATE_STRICT_ENABLED`, which is issuance-side |
| WebService account bound to every active shop | PS 9.x answers 503 "The PrestaShop webservice is disabled" with `PSWS-Version: 0`, even with `PS_WEBSERVICE` on |

## Ported, not reinvented

Each is cited at its call site in the script:

- WebService schema detection (PS 8.x `ps_api_access` vs 9.x `ps_webservice_account`), the permission-table variance and the `ps_webservice_account_shop` binding trap: `apps/api/test/integration/helpers/prestashop-fixture.helper.ts:152-308`
- WooCommerce REST key (`wc_api_hash` = `hash_hmac('sha256', $ck, 'wc-api')`): `docker/woocommerce/01-seed-wc-data.sh:33-53`
- MySQL over `docker exec`: `perf/prestashop-baseline/seed-products.sh:27-31`

## Verified against a live stack

Run end to end against a real PrestaShop 9.0.2 + WooCommerce 6.7.1 stack, then reverted.

- Schema detection resolved correctly: `ps_webservice_account / ps_webservice_permission`, method-row shape
- **The PrestaShop connection test genuinely passed** against the key, permissions and shop binding this script wrote - that is the real proof the ported SQL is correct, not a shape assertion
- **Idempotency proven**: a second run reported `found: 9, created: 0`
- `--verify-only` reported 7 real gaps on an unbootstrapped stack and exited 1
- The tax-group probe found a real defect on the live stack: 3 of 6 products carried no tax rules group, which is exactly the failure mode above
- Teardown restored the stack: connection count back to its original 13, zero perf mappings, both keys removed

## Three things the live run caught that the spec did not predict

1. **`POST /connections/:id/test` answers HTTP 200 even when the connection is dead**, with `{"success": false, "message": ...}` in the body. The first implementation here checked the HTTP status and therefore reported a WooCommerce connection pointing at a nonexistent host as healthy - the precise failure #2860's own acceptance criterion warns about. It reads the field now.
2. **API shapes**: the PrestaShop credential key is `webserviceApiKey` (not `apiKey`), `GET /v1/connections` rejects a `limit` parameter and answers a plain array, and login answers `access_token`.
3. **There is no delete route for a connection** - the controller offers `PATCH :id/disable` and nothing else. Irrelevant on the `lab` stand, whose reset is a full `down -v`, but it means a run against a shared stack leaves its connections behind. Documented.

## Two deliberate choices

**The Allegro connection test is conditional.** Both tenants are created pointing at `http://allegro-stub:8080`, and a failing test warns rather than failing the run, because the stub (#2856) does not exist yet. The whole contract is in place the moment it lands.

**The Allegro credentials carry `accessToken` only** - no `expiresAt`, no `refreshToken`, no client id or secret. With `expiresAt` absent, `ensureFreshToken` short-circuits before every request, so no call is ever made to the OAuth host, which is hardcoded and has no per-connection override. `enabledCapabilities` is `OrderSource` only for the same class of reason: adding `OfferManager` arms `marketplace.offers.sync`, whose scheduler task declares no required capability, turning a clean 404 into a retryable `CapabilityNotEnabledException` that burns ten attempts.

## Not covered

- The from-scratch **module install** branch was not exercised: both available stacks already had the module installed and the carrier row present, so the verify branch ran and the script was correctly a no-op. The install branch fails loudly if `install()` does not produce the carrier, and includes the documented uninstall-then-install workaround for the PS 9.0.2 hook flake. It runs for real on the first fresh `lab` stand.
- `pnpm lint` and `pnpm type-check` were not run: the change touches no TypeScript or JavaScript (bash, markdown, `.gitignore`) and the fresh worktree carries no `node_modules`. `pnpm check:invariants` passes.
- `ALLEGRO_OFFER_POOL_SIZE` must match the stub's offer-id space (#2856). The script records the value in the manifest; agreeing the two is that issue's.


piotrswierzy (commented):
## 1️⃣ Summary

**Approach**: ✅ Sound, and unusually well-evidenced for a tooling PR.

**Files reviewed**
- `perf/openlinker-throughput/bootstrap.sh` (new, 536 lines)
- `docs/operations/perf-lab-stand.md` (new, 151 lines)
- `.gitignore` (+4)

No TypeScript, no ORM entity, no migration, no runtime code path — `docs/migrations.md` and the hexagonal-layering checklist have nothing to bind here. The reviewable surface is: does the script do what it claims, does it fail loudly when it does not, and is anything secret-bearing committed. Secrets handling is correct — both generated artefacts (`stand-ids.env`, `stand-bootstrap-undo.txt`) are git-ignored in the same commit, and the undo log records only truncated key prefixes.

The four-preconditions table is the strongest part of this PR. Each entry names a specific silent-failure mode with a code citation, and the framing — *a destination-create rejection is reduced to a message on `order_records.syncStatus` without rethrowing, so the job reports `outcome: 'ok'` and the queue drains* — is exactly the class of defect a measurement programme has to design against, because a stand that measures nothing looks identical to a stand that measures well. That reasoning is also correctly mirrored into the runbook rather than living only in the PR body.

---

## 2️⃣ Findings

**[IMPORTANT]** — `bootstrap.sh:107-111` — `ps_sql` swallows every write failure

```bash
ps_sql() {
  docker exec -i -e MYSQL_PWD="$PS_MYSQL_PWD" "$PS_MYSQL_CONTAINER" \
    mysql -uroot -N -B "$PS_DB" -e "$1" 2>&1 | grep -v '^mysql: \[Warning\]' || true
}
```

Two things combine badly here, and both are divergences from the precedent the PR cites. `perf/prestashop-baseline/seed-products.sh:27-31` has the `2>&1 | grep -v` idiom but **no `|| true`** — that trailer is new. With it, and with `2>&1` folding stderr into the return value, every `ps_sql` **write** in this script is unconditionally reported as having succeeded:

- the `DELETE`/`INSERT` permission-grant loop (`:250-262`)
- the `ps_webservice_account_shop` binding (`:265`) — the single trap the runbook says answers 503 with `PSWS-Version: 0`
- the `PS_WEBSERVICE` configuration upsert (`:267`)

None of those is `set -e`-visible, so a run in which all three failed still prints `CREATED PrestaShop WebService key` and reaches `stand is bootstrapped`. That is the same shape as the `outcome: 'ok'` failure the PR opens by warning about, one layer down in the tooling meant to prevent it.

It is partly mitigated — `step_verify_connections` exercises the PS key, permissions and shop binding end to end, which is genuinely the right check and which the PR is right to call the real proof. But the mitigation is a *later, separate* step whose failure is reported as `gap "connection test FAILED"` with no hint that a swallowed SQL error caused it. The reads have a milder version of the same problem: a MySQL error string becomes `$n` in `carrier_present`, `[ "ERROR…" -gt 0 ]` is falsy, and the script concludes the carrier is absent and reinstalls the module.

Suggest splitting the helper — keep the lenient form for probes, and add a strict `ps_sql_write` (no `|| true`, stderr not folded into stdout, non-zero `die`s). Writes are where silence costs you.

**[IMPORTANT]** — `bootstrap.sh:496-513` (`step_emit`) — `stand-ids.env` is written *before* the gap check exits 1

`main()` calls `step_emit` and only then evaluates `${#GAPS[@]}` and exits 1. So a run that found real gaps still leaves a `stand-ids.env` on disk, possibly with empty ids (`PS_CONNECTION_ID=` when `ol_ensure_connection` gapped). #2841's harness sources that file and #2845's runner is unattended and **resumable** — the combination is precisely a resumed run picking up a file written by a failed bootstrap and measuring against a stand nobody confirmed. Either emit only when `${#GAPS[@]}` is 0, or stamp `BOOTSTRAP_GAPS=<n>` into the file and have the harness refuse a non-zero value. The second is better: it fails at the consumer, where the resumed run actually is.

**[IMPORTANT]** — `bootstrap.sh:280-296` (`step_tax_group`) — a documented precondition that the bootstrap will not repair

Step 3 probes and calls `gap`, but never creates. So on a stand where the seeded catalogue has no tax rules group — which the PR reports is the **actual** state of a live stack, 3 of 6 products — `./bootstrap.sh` exits 1 and the operator is told a count and an exception name, with no instruction anywhere for what to do next. The runbook's own promise is *"no browser step and no manual paste"*, and the "After a reset" sequence lists `bootstrap.sh` as step 4 with nothing about repairing this. Either set `id_tax_rules_group` to the dominant group for the products missing it (the value is already resolved into `PS_TAX_RULES_GROUP` two lines above), or state plainly in the runbook that this step is diagnose-only and give the one-line SQL. Right now the step is a precondition the script names and declines to satisfy.

**[SUGGESTION]** — `bootstrap.sh:222-238` — schema detection order diverges from the cited source

`prestashop-fixture.helper.ts` tests `ps_api_access && ps_api_access_resource` **first** and falls through to `ps_webservice_account`; this script inverts that. On a clean 8.x or a clean 9.x the two agree, but on an upgraded stack carrying both table families they pick different branches. Since the PR's stated discipline is "ported, not reinvented", matching the precedent's order costs nothing and removes the question.

**[SUGGESTION]** — `bootstrap.sh:394` (`ol_ensure_connection`) — `curl -f` hides the rejection body

`ol_api` uses `curl -fsS`, so a 400 from `POST /v1/connections` exits non-zero with the response body discarded, and the failure surfaces as the generic `die "failed to create connection '$name'"`. Validation is precisely where a stand bootstrap gets rejected (the PR itself hit `webserviceApiKey` vs `apiKey` and the `limit` rejection this way), and the body is the only thing that says which field. Capturing stdout+status and echoing the body on failure would have shortened both of those debugging loops.

**[SUGGESTION]** — `bootstrap.sh:330-334` — the WooCommerce rotate warning fires before the verify-only bail

```bash
warn "a WooCommerce perf key exists but ... - rotating"
[ "$VERIFY_ONLY" = 1 ] && { gap "..."; return 0; }
```

Under `--verify-only` the script announces it is rotating a key and then correctly does not. Swap the two lines so a probe-only run never claims a write.

**[SUGGESTION]** — `bootstrap.sh:78` — `-h|--help` does `sed -n '2,32p' "$0"`

Line-number-addressed help drifts silently the first time the header comment grows. A here-doc `usage()` is the same number of lines and cannot go stale.

---

## Positive observations

- **The `printf -v` note at `:381-384`** — assigning through a named output variable specifically so the call is not forced into a command substitution whose subshell would discard the `FOUND`/`CREATED`/`GAPS` updates — is a real bash trap, correctly diagnosed, and commented where the next reader will hit it. That comment is worth more than the line it explains.
- **Reading `success` from the body rather than the HTTP status** (`connection_test`, `:452-464`) is the finding that makes this script trustworthy rather than merely green, and the PR is right to lead with it. A WooCommerce connection pointing at a nonexistent host reporting healthy is the exact self-deception the whole programme is guarding against.
- **The Allegro credential shape** — `accessToken` only, so `ensureFreshToken` short-circuits and no request ever reaches the hardcoded real token host — and **`OrderSource`-only capabilities** so `marketplace.offers.sync` is never armed, are both non-obvious and both correctly justified at the call site. The second is a good catch: a scheduler task with no declared required capability turning a 404 into a retryable `CapabilityNotEnabledException` is not something you find by reading the happy path.
- **`--verify-only` reporting 7 real gaps and exiting 1 on an unbootstrapped stack**, plus the second run reporting `found: 9, created: 0`, is idempotency demonstrated rather than asserted.
- **Honest "Not covered" section.** Saying the from-scratch module-install branch was never exercised, and why, is better than a test plan that implies otherwise.

## Documentation

`docs/operations/perf-lab-stand.md` is a good runbook — it states what breaks and why for each precondition rather than only listing commands, and the "Things worth knowing before the first run" section captures the three live-run surprises where an operator will meet them. Two gaps: the tax-group repair instruction (IMPORTANT above), and the file is not linked from any operations index, so discovery depends on knowing it exists.

---

## Final Assessment

### Merge readiness
🔄 **Approve with changes** — no blocking issue. This is lab-only tooling with no runtime reach, it was verified against a live stack and reverted, and the reasoning behind every non-obvious choice is recorded at the call site. But two of the IMPORTANT items are the same defect class the PR itself is written to prevent — a step that cannot fail loudly (`ps_sql`'s `|| true` on writes) and an artefact that outlives the failure that produced it (`stand-ids.env` written before the gap exit). Both matter more here than they normally would, because #2845 consumes this unattended and resumably: nobody will be watching the log.

### Priority fixes
1. Split `ps_sql` into probe and strict-write forms; drop `|| true` and the stderr fold on the write path.
2. Write `stand-ids.env` only on a zero-gap run, or stamp the gap count and have the harness refuse it.
3. Repair the tax rules group, or document the manual fix in the runbook's "After a reset" sequence.
4. (Optional) Match the fixture helper's schema-detection order; surface the response body when a connection create is rejected; fix the verify-only rotate warning.

norbert-kulus-blockydevs:
Addressed all findings from the review:

**IMPORTANT**
1. `ps_sql` swallowing write failures — split into a lenient probe `ps_sql` and a strict `ps_sql_write` (no stderr fold, no `|| true`, `die`s on non-zero). All WebService-step writes (key insert, permission grant loop, shop binding, `PS_WEBSERVICE` config) now go through `ps_sql_write`.
2. `stand-ids.env` written before the gap check — `step_emit` is now called only after every step reports zero gaps, so a failed bootstrap never leaves a file behind for #2845's resumable runner to pick up.
3. Tax rules group precondition with no repair — `step_tax_group` now sets `id_tax_rules_group` on the affected products to the dominant group (falls back to a named gap only if the whole catalogue has no group to repair to). Runbook updated to describe the repair.

**SUGGESTIONS**
4. Schema-detection order now matches `prestashop-fixture.helper.ts` (`ps_api_access` checked first).
5. `ol_api` now captures status and body separately and surfaces the body on a non-2xx response instead of `curl -f` swallowing it.
6. WooCommerce rotate-warning now only fires after the `--verify-only` bail, so a probe-only run never claims it's rotating a key it doesn't touch.
7. `-h|--help` now uses a `usage()` heredoc instead of `sed -n '2,32p' "$0"`, and the file header points at it instead of duplicating it.

Pushed as bdc97efa4. Will monitor CI and merge once green and rebased.

piotrswierzy (approved):
## Re-review at head `d4512e0` (was `ab62ec5`) — APPROVE

### 1. Prior findings, verified item-by-item

**[IMPORTANT] `ps_sql()`'s `|| true` + `2>&1` made every SQL *write* silent on failure — RESOLVED, and resolved the way it was asked.**

`:106-123` now carries two helpers with the divergence explained in the comment rather than left implicit. `ps_sql` keeps the lenient probe form (stderr folded, always exits 0) — correct for a SELECT whose absence *is* the answer. `ps_sql_write` has no stderr fold, no `|| true`, and `die`s on the first non-zero. I checked every write site rather than the helper alone, and all six now go through the strict form: the account `INSERT` (`:250`), the permission `DELETE`/`INSERT` loop (`:260-267`), the `ps_webservice_account_shop` binding (`:272` — the 503 trap the runbook itself names), the `PS_WEBSERVICE` config upsert (`:274`), and the new tax-group `UPDATE` (`:305`). No `ps_sql` call remains on a write path. `set -euo pipefail` is in force at `:7` and every `ps_sql_write` call is a plain statement (never inside a command substitution, where `die`'s `exit 1` would have killed only a subshell), so the strictness is real rather than notional.

**[IMPORTANT] `stand-ids.env` written before the gap check exits 1 — RESOLVED.**

`main()` (`:561-583`) now emits the summary, and on `${#GAPS[@]} -gt 0` prints each gap and `exit 1`s **before** reaching `step_emit`. The reason is recorded in `step_emit`'s own docblock and names both consumers (#2841 sources the file, #2845 is unattended and resumable), which is the right place for it — the next person to reorder `main()` reads that comment. You took the stricter of the two options I offered; that is fine, and arguably better, since the absent file is unambiguous where a `BOOTSTRAP_GAPS=n` line depends on a consumer remembering to read it.

**[IMPORTANT] `step_tax_group` probed a precondition it would not repair — RESOLVED.**

`:288-311` now repairs: it sets `id_tax_rules_group` to the dominant group already resolved into `PS_TAX_RULES_GROUP`. The three degenerate paths are all handled distinctly and correctly — `--verify-only` reports the gap and writes nothing; `--dry-run` announces the exact statement via `would`; and the genuinely unrepairable case (no product in the catalogue carries a group, so there is nothing to repair *to*) falls back to a named gap whose message says to assign one in the back office rather than leaving the operator with an exception name. `docs/operations/perf-lab-stand.md:98-104` states the same three-way behaviour, so the runbook no longer promises "no browser step" over a step that silently required one.

### 2. Suggestions from the prior review

All four taken: schema detection now tests `ps_api_access` first (`:225`), matching `prestashop-fixture.helper.ts`; `ol_api` (`:404-419`) captures status and body separately via `-w '\n%{http_code}'` and `die`s with `HTTP $status: $resp_body`, so the next `webserviceApiKey`-vs-`apiKey` class of rejection names its own field; the WooCommerce `--verify-only` bail (`:332`) now precedes the rotate warning (`:333`), so a probe-only run can no longer announce a write it does not perform; and `-h|--help` is a `usage()` heredoc (`:38`) with the file header pointing at it instead of duplicating it.

### 3. Step 3 / Step 4 assessment of the current diff

- **Scope unchanged and still narrow.** 3 files, +749/−0: one bash script, one runbook, four `.gitignore` lines. No TypeScript, no ORM entity, no migration, no runtime path — the hexagonal-layering and `docs/migrations.md` checklists have nothing to bind. Secrets handling remains correct: both generated artefacts are git-ignored in the same commit and the undo log records only what `created()` was given.
- **The fixes did not introduce a regression in the failure semantics.** I specifically looked for the case where strictness makes a *probe* fatal — `carrier_present`, the `information_schema` existence tests and the tax counts all still use the lenient form, so a MySQL hiccup during discovery degrades to "absent" rather than aborting, which is the pre-existing behaviour and the right one for a probe.
- **The `--dry-run` / `--verify-only` matrix is now uniform.** Every mutating step guards with `would`/`VERIFY_ONLY` before its write, including the new tax repair. That matters more after fix (2), because a `--verify-only` run legitimately gaps and must not now be blamed for the missing `stand-ids.env`; `step_emit` returns early on `VERIFY_ONLY` rather than falling through the gap gate, which is correct.
- **CI.** `Build`, `Type Check`, `Docker Build Smoke Test`, `PHP Unit Tests` and `Detect changed paths` green at this head; `Lint`, `Test` and `Integration Tests` still queued. The change touches no TS/JS, so `Lint`'s `check:invariants` is the only one with anything to say here and the PR body reports it passing locally. Worth a glance before merging rather than a blocker.

### 4. One new, non-blocking interaction introduced by fix (2)

Making `step_emit` conditional on a zero-gap run gave the WooCommerce step a new edge. `WC_CS` exists only in memory and in `stand-ids.env`; the consumer key is stored hashed and cannot be read back (`:325-334` says so). So a run that *creates* the WC key and then gaps in a later step (`step_connections`, `step_offer_mappings`, `step_verify_connections`) loses that secret permanently — the shop holds a live key nobody can use, and the next run takes the rotate branch. It is self-healing and bounded to one orphan (the rotate deletes before inserting), and on the `lab` stand a `down -v` erases it anyway, so this is not worth blocking on. If it ever bites, the cheap fix is to write `stand-ids.env` unconditionally but stamp `BOOTSTRAP_GAPS=<n>` into it and have #2841's harness refuse a non-zero value — the consumer-side variant of the same fix, which keeps unrecoverable material out of the failure path.

### 5. Final assessment

✅ **Approve.** All three IMPORTANT findings are closed at the mechanism level, not merely acknowledged, and all four suggestions were taken. The two that mattered most — a write that cannot fail loudly, and an artefact that outlives the failure that produced it — were both instances of the defect class this PR's own opening paragraph warns about, and both are now structurally unable to recur: the strict/lenient split is in the helper rather than in each call site's discipline, and the emit is gated by control flow rather than by a flag someone must remember to check. The remaining note above is an observation for whoever writes #2841, not a change request here.

Issue #1918, opened 2026-07-29
[TASK] CORE/Frontend — persist providerCode on Shipment and re-key failure triage grouping on it


norbert-kulus-blockydevs:
## Problem / Context

`ShippingProviderRejectionException` (`libs/core/src/shipping/domain/exceptions/shipping-provider-rejection.exception.ts`) carries a structured `providerCode` (plus an open-shape `providerDetails`, e.g. InPost's `{ fieldErrors: { fieldName: string[] } }`), but only the free-text `message` is persisted. `ShipmentDispatchService.generateLabel`'s catch block (`libs/core/src/shipping/application/services/shipment-dispatch.service.ts:339-360`) writes `errorMessage: message` onto the `Shipment` and drops `providerCode` / `providerDetails` — they only reach the warn-level log line above it.

That single free-text field is the only failure signal OL keeps, and the service's own comment notes it "is often a generic 'validation error' that hides which field the provider rejected (#1428)". Two consequences:

1. **Grouping cannot be honest.** The `/shipments` triage strip (#1826, PR #1905) fuzzy-matches normalised `errorMessage` text — `groupFailedShipmentsByCause` in `apps/web/src/features/shipments/lib/group-failed-shipments-by-cause.ts` (on PR #1905's branch, not yet on `main`). Its own header comment documents the limitation: a bad receiver postcode, a missing `parcel.template`, and an over-limit COD can all arrive as the same normalised string, so unrelated failures group together. #1905 reworded the strip's copy from a causal claim to a same-message observation precisely because the data can't support more — but the underlying gap remains.
2. **There is no retryability classification.** A 429/5xx that exhausted its retries lands in the same free-text bucket as a permanent validation rejection, so nothing distinguishes "carrier briefly unavailable, regenerate" from "the recipient address is wrong, fix the order first." Shipping doesn't consult the repo's existing retry-classifier / auth-failure-classifier registry pattern at all.

### Illustrative example

Three unrelated InPost dispatch failures land on the same connection the same morning. Today's `/shipments` and the order-detail `OrderShipmentPanel` show every one of them as the identical string `"Validation error"`, because that's all `errorMessage` ever carries:

| Order | What actually happened (`providerCode`, dropped today) | What OL shows today |
|---|---|---|
| 5518 | `INVALID_SENDER_ADDRESS` — permanent, fix the connection's sender profile | `Validation error` |
[...]

PR #1905, opened 2026-07-29
feat(shipments): inline retry from a failed row + carrier-message role visibility

norbert-kulus-blockydevs:
## Summary

Closes #1826

- Adds inline recovery from `/shipments`: each `failed` row's `DataTable` `expandable` accordion carries the failure detail and a **Regenerate** deep-link into the order-scoped generate-label form (`/orders/:orderId?retryShipmentId=`), resetting the same pre-waybill row in place where the domain allows it.
- Adds a cause-first **triage strip**: groups ≥2 failed rows on the loaded page sharing a normalised `errorMessage` **per connection**, with a cause-neutral "Review connection settings" CTA (never assumes a sender-address cause).
- **Viewer-role redaction** of the raw carrier `errorMessage` (may embed address fragments) across all three surfaces that render it — status-cell text, its `title` tooltip, and the accordion — enforced **both** client-side and server-side (`ShipmentResponseDto.fromDomain` now redacts for any session without `shipments:write`, not just the UI).
- Fixes two pre-existing bugs found while wiring the Order column's navigation: `EntityLabel` was rendering an unlinked "Unknown" instead of a working link, and the "Connection" column is renamed/reskinned to **Provider** via `ConnectionDot`.
- Adds the **Action** column (plain, non-interactive per-status severity label: Fix/Finish/Send/View), a frozen **Status** column (`stickyLeftColumns`), and full mobile card parity (order link, severity in `summary`, same redaction, `collapsibleDetail`).
- New `shipments:write` permission (CORE + FE), granted to admin/operator only; gates the recovery/triage affordances on both `/shipments` and the order-detail shipment panel.

## Review round

Three independent `/pr-review` passes ran against the finished diff before this PR. All findings were addressed in this same branch — see § 11 of `docs/plans/implementation-plan-shipments-inline-retry.md` for the full list, including two blocking issues (a redaction gap on the order-detail panel one click from the fixed Order-column link, and a deep-link auto-open that bypassed the payment/route dispatch gate and could re-open a dismissed form) and ten follow-on fixes (cross-connection triage grouping, cause-neutral copy, a digit-stripping ordering bug, an inverted paczkomat pre-fill precedence, restored mobile navigation, an empty-accordion fallback, and closed test-coverage gaps).

## Test plan

- [x] `pnpm lint` — 0 errors (repo-wide)
- [x] `pnpm type-check` — clean across all 18 packages
- [x] `pnpm test` — core 1952, api 1059, web 2659, all passing
[...]

PR #2874, opened 2026-09-03
fix(web): key shipment triage on an EXACT providerCode, not merely a non-null one

### LINKED ISSUE #2873: [BUG] Frontend — triage strip drops the carrier message once a coarse shipx.* providerCode exists, and grouping treats it as exact
## Problem / Context

PR #2805 (fix for #2804) makes `InpostHttpClient.throwIfNotOk` fall back to ShipX's own `error` code — namespaced as `shipx.{code}` — for `providerCode` when the flattened `details` map is empty. That is the right fix at the integration boundary, and it is provably retryability-neutral (`shipx.*` matches no arm in `deriveRetryabilityClass`, so it classifies `'unknown'`, the same class `null` produced).

But two frontend modules branch on `providerCode !== null` and treat any non-null value as an **exact, structured** discriminator. A `shipx.*` code is not one — `shipx.validation_failed` is precisely the coarse string those modules' own headers warn about. So the same InPost failures that used to take the `null` branch now take the code branch, and both halves get worse:

**1. Display — the raw carrier message stops rendering.**

`apps/web/src/features/shipments/components/shipment-triage-strip.tsx:70` derives `retryabilityClass` from `group.providerCode !== null`, and lines 96-107 pick between two branches on the same test. Before #2805 a no-`details` InPost failure had `providerCode: null`, so the strip rendered the carrier-message branch: it said "report the same **carrier message**", printed the raw `sampleReason` inline, and — per that file's own header — deliberately stayed "an OBSERVATION, never a causal claim". After #2805 the same failures acquire a code, so the strip says "report the same **rejection code**", prints `shipx.validation_failed`, and **stops rendering `sampleReason` at all**.

The retryability label saves it from giving *wrong* remedy advice (it renders the `'unknown'` label), but the operator loses the free-text message that was previously their only actionable signal, and gains a code that does not narrow the cause. For the exact failure shape #2804 was opened about, this is a net regression in what the operator can act on.

**2. Grouping — the coarseness guard is skipped for a value that is not exact.**

`apps/web/src/features/shipments/lib/group-failed-shipments-by-cause.ts:100-103` keys the group on `providerCode` whenever it is non-null and skips `MIN_CAUSE_KEY_LENGTH`, on the stated grounds that "a `providerCode` is an exact structured value, never too short to trust". Note the guard as written would not have rejected `shipx.validation_failed` on **length** anyway (22 characters) — the problem is not length, it is that a coarse carrier bucket is being treated as exact. So a connection whose failures were previously grouped (or correctly *not* grouped) by normalised message can now collapse several unrelated causes — a bad postcode, a missing parcel template, an over-limit COD — into one group that the strip labels "report the same rejection code", which `group-failed-shipments-by-cause.ts`'s own header defines as "an HONEST shared-cause claim".

## Proposed Solution

Both halves are frontend-local. Do **not** put a carrier-vocabulary allowlist in the client, and do **not** narrow the integration-side fallback — the code is correct at the boundary; the client's assumption about what a `providerCode` means is what is wrong.

**Display half** — in `shipment-triage-strip.tsx`, render `sampleReason` **alongside** the code when `deriveRetryabilityClass(group.providerCode)` returns `'unknown'`. An unclassified code is by definition one that does not narrow the cause, so the raw message is still the operator's best signal and the two are complementary rather than alternatives. The heading copy should stay honest for that case (a coarse code is not a shared-cause claim). Keep the existing behaviour untouched for a classified code (`transient` / `permanent` / `auth`), where the code genuinely is the more useful of the two.

**Grouping half** — in `group-failed-shipments-by-cause.ts`, stop treating "non-null `providerCode`" as a synonym for "exact structured value". Introduce an explicit notion of a **coarse** code (the same test the display half uses is the natural one: `deriveRetryabilityClass(...) === 'unknown'` — one rule, one place, already mirrored against core by `shipment-retryability.ts`) and, for a coarse code, fall back to the normalised-message key or compose the two, so unrelated causes do not collapse. Update the module header, which currently asserts the property being fixed.

Both changes must keep `shipment-retryability.ts`'s FE-mirror discipline intact — it is a hand-mirror of `libs/core/src/shipping/domain/provider-code-retryability.ts` and must stay in sync; this issue adds no new arm to it.

## Classification

**Type**: Frontend
**Layer**: Interface (feature components + feature lib)
**File(s)**:
- `apps/web/src/features/shipments/components/shipment-triage-strip.tsx`
- `apps/web/src/features/shipments/lib/group-failed-shipments-by-cause.ts`
- `apps/web/src/features/shipments/lib/shipment-retryability.ts` (read-only consumer; no new classification arm expected)
- tests: `apps/web/src/features/shipments/components/shipment-triage-strip.test.tsx`, `apps/web/src/features/shipments/lib/group-failed-shipments-by-cause.test.ts`

## Dependencies

- **#2805** — the regression is only observable once that PR merges and `shipx.*` codes start being persisted onto `Shipment.providerCode`. This issue can be implemented in parallel, but its tests need the `shipx.*` value shape #2805 introduces.
- Follows #2804 (the integration-side fix) and #1918 (which introduced the `providerCode`-keyed grouping and its "exact structured value" assumption).

## Assumptions

- `deriveRetryabilityClass(...) === 'unknown'` is an acceptable proxy for "coarse code". It is the only shape-based classifier the FE has, it is already the sanctioned FE mirror of a core rule, and every code family that *does* narrow the cause (`api.http-NNN`, `preflight.*`, `command.*`, `target_point`) classifies to something other than `'unknown'`. If a future code family classifies `'unknown'` while still being exact, that is a reason to extend the classifier, not to add a second rule here.
- No backend change is needed. `providerCode` remains carried verbatim per `libs/core/src/shipping/domain/exceptions/shipping-provider-rejection.exception.ts`; nothing here changes the wire contract or `shipments.providerCode`.
- The strip stays admin/operator-only; no permission-gating change is in scope. `sampleReason` is already the sensitive carrier text that page's existing gate covers, so re-rendering it introduces no new exposure.

## Acceptance Criteria

- [ ] For a group whose `providerCode` classifies `'unknown'`, the triage strip renders the raw `sampleReason` **in addition to** the code, and its heading copy does not assert a shared *cause* on the strength of a coarse code alone.
- [ ] For a group whose `providerCode` classifies `transient` / `permanent` / `auth`, the strip's rendering is unchanged from today (code + retryability label, no `sampleReason`).
- [ ] Two failed shipments on the same connection carrying the same coarse `shipx.*` code but genuinely different `errorMessage` values are **not** collapsed into one group by `groupFailedShipmentsByCause`.
- [ ] Two failed shipments on the same connection carrying the same *classified* `providerCode` (e.g. `target_point`) are still grouped exactly as today, regardless of message differences.
- [ ] The `MIN_CAUSE_KEY_LENGTH` guard still applies to every key derived from a normalised message, including any composed key introduced by this change.
- [ ] `group-failed-shipments-by-cause.ts`'s module header no longer asserts that any non-null `providerCode` is an exact structured value.
- [ ] `shipment-retryability.ts` remains a faithful mirror of `libs/core/src/shipping/domain/provider-code-retryability.ts` (no FE-only classification arm added).
- [ ] No architecture boundary violations — frontend-only; no `@openlinker/core/*` value import, no carrier vocabulary in the client.

---

Filed per the tech-lead review of #2805 (`piotrswierzy`), which made this the one condition on that approval: *"file the `shipment-triage-strip.tsx` follow-up (now with the grouping half added) before #2804's epic closes."*

norbert-kulus-blockydevs:
## Summary

#1918 keyed the failed-shipment triage group on any non-null `providerCode`, on the stated grounds that "a `providerCode` is an exact structured value". That premise is wrong for an opaque carrier-surfaced code, and PR #2805 is about to make it wrong in volume: `InpostHttpClient` will fall back to ShipX's own bucket code (`shipx.validation_failed`) whenever the carrier sends no field-level `details`, and that value names a **bucket**, not a cause - a bad postcode, a missing parcel template and an over-limit COD all arrive under it.

Two frontend modules read "non-null" as "exact". Both are fixed here.

**Display.** `ShipmentTriageStrip` branched on `providerCode !== null`, so the same InPost failures that render the carrier-message branch today would flip to the rejection-code branch and **stop rendering `sampleReason` at all** - trading the operator's only actionable signal for a code that narrows nothing. The branch is now keyed on exactness, and a coarse code is rendered **alongside** the raw message, labelled with its own `'unknown'` retryability class so the strip says plainly that it does not classify the cause.

**Grouping.** `groupFailedShipmentsByCause` skipped `MIN_CAUSE_KEY_LENGTH` for any non-null code. A coarse code is now composed with the normalised message (`{code}::{normalisedMessage}`), which keeps the code on the group for display without letting it collapse unrelated causes, and puts the length guard on every key that leans on the lossy text half rather than on the `null` case alone.

One correction to the framing in the review that prompted this: the length guard alone would never have caught it. `shipx.validation_failed` is 22 characters and passes `MIN_CAUSE_KEY_LENGTH` comfortably. The defect is coarse-treated-as-exact, not length.

**Behaviour note worth calling out separately from the InPost case above**: this also changes existing DPD grouping. A DPD failure carrying the opaque `PARCEL_TOO_LARGE` code previously grouped purely on that code, merging every message under it into one banner. It now groups on `{code}::{normalisedMessage}` like every other coarse code, so those rows can **fragment by message** where they used to merge. That's the correct direction - the merge was the dishonest claim this PR removes - but it's a visible change to a group an operator may already recognise from before this PR, not only the InPost/#2805 case the rest of this description is framed around.

## One rule, one place

`isExactProviderCode(providerCode)` is a type predicate testing `deriveRetryabilityClass(...) !== 'unknown'`, deliberately rather than spelling out code families:

- that classifier is already the sanctioned FE mirror of `libs/core/src/shipping/domain/provider-code-retryability.ts`;
- every family that narrows the cause (`api.http-{status}`, `preflight.*`, `command.*`, `target_point`) classifies to something other than `'unknown'`, while an opaque bucket (`shipx.*`, `PARCEL_TOO_LARGE`) does not;
- **no new arm is added to the mirror** and no carrier vocabulary enters the browser, which is precisely what this must not become.

`null` counts as not-exact (it carries no structure at all), so the same predicate is the single test for "does this group lean on the lossy text half", and the strip and the grouping lib cannot disagree about it.

`shipment-retryability.ts`'s header now states this coupling explicitly: widening a retryability arm there also widens what `isExactProviderCode` treats as exact, and therefore what the grouping treats as an honest shared-cause claim. There's no `check-*-mirror.mjs` script standing behind that - a future retryability change wouldn't fail any invariant check, only the grouping tests, and those would read as "an unrelated test broke" rather than "you just re-enabled cross-cause grouping" without the note pointing at why.

## Behaviour

| `providerCode` | Group keyed on | Strip heading | Strip body |
|---|---|---|---|
| exact (`api.http-503`, `preflight.*`, `target_point`) | the code alone | "report the same rejection code" | code + retryability label (**unchanged**) |
| coarse (`shipx.*`, `PARCEL_TOO_LARGE`) | `{code}::{normalisedMessage}` | "report the same carrier message" | raw message **plus** `(carrier code shipx.validation_failed - Cause not classified)` |
| `null` | `::{normalisedMessage}` | "report the same carrier message" | raw message (**unchanged**) |

Nothing changes for a group whose code is exact, and nothing changes for a group with no code. Only the coarse case moves, and it moves toward what the `null` case already did.

## Relationship to #2805

This stands on its own against `main` and does **not** need #2805 merged first - the assumption was already wrong for any opaque carrier code (DPD can surface one). #2805 is what makes the case common rather than incidental, so the two are independent and can merge in either order.

- **(IMPORTANT)** `shipment-retryability.ts`'s header now states the coupling to `isExactProviderCode` explicitly, so a future retryability-arm edit is made in view of the grouping consequence rather than silently widening it.
- **(SUGGESTION)** `ShipmentTriageStrip`'s coarse branch no longer carries the always-true `retryabilityClass` conjunct - `retryabilityClass` is `'unknown'` by construction on that branch, so the condition is simplified to `providerCode !== null` with `RETRYABILITY_LABEL.unknown` inlined, plus a comment explaining why.
- **(SUGGESTION)** The DPD `PARCEL_TOO_LARGE` fragmentation behaviour change is now called out in the Summary above.

## Test plan

- [x] `pnpm --filter @openlinker/web type-check` passes
- [x] `pnpm --filter @openlinker/web lint` - 0 errors, no new warnings in either touched file
- [x] `pnpm check:invariants` passes (no mirror script is affected - `shipment-retryability.ts`'s exported behaviour is untouched, only its docblock)
- [x] Full-repo `pnpm type-check` and `pnpm lint` pass, 0 errors
- [x] `vitest run src/features/shipments src/pages/shipments` - 225/225 pass across 12 files
- [x] New grouping tests: coarse code + differing messages does NOT group; coarse code + shared message groups and still carries the code; two different coarse codes stay apart; the length guard fires when a coarse code's message normalises away; an exact code still groups across differing messages; a coarse-code row never mixes with a no-code row
- [x] New `isExactProviderCode` tests covering all five exact families, three coarse values, and `null`
- [x] New strip tests: a coarse-code group still renders its raw message and says the cause is unclassified; an exact-code group renders the code alone, without the message
- [x] Every pre-existing #1826 / #1905 / #1918 test in both files passes unchanged

Closes #2873


piotrswierzy (approved):
## 1️⃣ Summary

**Approach**: ✅ Correct — and, importantly, it closes **both** halves of the defect that prompted it, not just the visible one.

**Files reviewed**
- `apps/web/src/features/shipments/lib/group-failed-shipments-by-cause.ts` (+ test)
- `apps/web/src/features/shipments/components/shipment-triage-strip.tsx` (+ test)

Frontend-only, `features/` tier, no backend/contract/schema/migration surface. Dependency direction is clean (`pages → features → shared`; the strip imports the lib as a same-feature relative import, and `shipments-page.tsx` deep-imports the lib exactly as it already did — the documented `pages/**` gap in `frontend-architecture.md § Feature Public Surface`, not a new violation). `isExactProviderCode` is correctly **not** added to a barrel: it has no cross-feature consumer.

---

## 2️⃣ Verification of the two-half finding

The review finding on #2805 was that a coarse namespaced fallback code breaks **two** independent assumptions. Traced both against the diff:

**Half 1 — display (`shipment-triage-strip.tsx`).** ✅ Closed. Both the heading (`'rejection code'` vs `'carrier message'`) and the body branch now key on `codeIsExact`, not `group.providerCode !== null`. On the coarse branch `sampleReason` is still rendered, and the code rides **alongside** it with its own `'unknown'` label (`Cause not classified`) rather than replacing it. Asserted directly by the two new strip tests (`getByText('sender postcode "22-213" invalid')` + `queryByText(/report the same rejection code/)).toBeNull()`).

**Half 2 — grouping (`group-failed-shipments-by-cause.ts`).** ✅ Closed, and this is the half that was easy to miss. The `MIN_CAUSE_KEY_LENGTH` bypass moved from `if (providerCode === null && …)` to the whole coarse-or-absent branch, so the guard now covers **every key that leans on the lossy text half** — which is the correct predicate, not the `null` special case it was. The `'500'` / `'404'` test pins exactly that: a coarse code whose message normalises away no longer forms a group off the bare bucket. And composing `{code}::{normalised}` (rather than dropping the code) is the right call — it keeps the code available for display while denying it grouping authority, which is what the strip's coarse branch then renders.

I also confirmed the PR body's correction to my framing is right: `shipx.validation_failed` is 22 chars and clears `MIN_CAUSE_KEY_LENGTH` easily. The defect was coarse-treated-as-exact; length was never going to catch it.

**Collateral checked.** `FailedShipmentCauseGroup.cause` has exactly one consumer outside this module — the React `key` at `shipments-page.tsx:582` — so changing the null-code key from `{normalised}` to `::{normalised}` is inert. No other reader treats `cause` as displayable, and the docblock on the field now says so explicitly.

---

## 3️⃣ Findings

**[IMPORTANT]** — `apps/web/src/features/shipments/lib/group-failed-shipments-by-cause.ts:66` (`isExactProviderCode`)

`isExactProviderCode` answers *"is this code exact enough to key a group on?"* by asking `deriveRetryabilityClass(...) !== 'unknown'` — a classifier written to answer a **different** question (*"how retryable is this?"*). The two happen to coincide today, and the docblock argues the coincidence well: every family that narrows the cause classifies to non-`unknown`, and refusing a carrier-vocabulary allowlist in the browser is exactly the right instinct (that allowlist is what `docs/architecture-overview.md`'s repeated "no marketplace vocabulary in the neutral layer" rule exists to prevent, one tier up). I would not ask you to change the implementation.

What I would ask for is a **guard on the coupling**, because the failure mode is silent and lands in the wrong direction. `shipment-retryability.ts` is a hand-maintained FE mirror of `libs/core/src/shipping/domain/provider-code-retryability.ts` with **no** `check-*-mirror.mjs` under `pnpm check:invariants` (I checked — only `check-retry-refusal-reason-mirror.mjs` exists, and it covers a different union). So the day someone reasonably improves the classifier — say, an arm mapping `shipx.*` to `'permanent'` because ShipX rejections do need a fix before retrying — that edit, in a file whose header says nothing about grouping, silently reclassifies the ShipX bucket as **exact** and reintroduces precisely the collapse this PR removes. Nothing in the current test suite would go red: the `isExactProviderCode` cases assert `shipx.*` is coarse, but a reader adding a retryability arm has no reason to look at a grouping spec, and the assertion failure would read as "the new arm broke an unrelated test" rather than "you just re-enabled cross-cause grouping".

Cheapest sufficient fix, in order of preference:
1. A one-line note in `shipment-retryability.ts`'s **own header**: adding an arm here also widens what `isExactProviderCode` treats as exact, which changes shipment grouping — the coupling stated where the edit happens, not only where it is consumed.
2. Optionally, name the coupling in the `isExactProviderCode` spec (`it('re-classifying a coarse family in shipment-retryability would silently widen grouping — see the header note')`), so the red test explains itself.

Not blocking: the behaviour shipped here is correct, and this is about protecting it from a future unrelated edit.

**[SUGGESTION]** — `shipment-triage-strip.tsx:120-135`

On the coarse branch `retryabilityClass` is `'unknown'` **by construction** (that is what `isExactProviderCode` tested), so `group.providerCode !== null && retryabilityClass` is a condition whose second conjunct can never be false there. Harmless, and arguably defensible as defence-in-depth, but the reader has to re-derive that to be sure it isn't a missed case. A comment, or narrowing to `group.providerCode !== null` alone with `RETRYABILITY_LABEL.unknown` inlined, would read more plainly.

**[SUGGESTION]** — behaviour note worth having in the description

Existing DPD groups keyed on an opaque code (`PARCEL_TOO_LARGE`) will now **fragment** by message where they previously merged. That is the intended and correct direction (the merge was the dishonest claim), but it is a visible change to a group an operator may already recognise, and the PR body currently frames the coarse case as the InPost/#2805 one. Worth a line so nobody reads a fragmented DPD strip as a regression.

---

## Positive observations

- Fixing the grouping half **at the key**, rather than only at the strip, is what makes this durable — a display-only fix would have left `groupFailedShipmentsByCause` asserting a shared cause that no consumer could ever un-assert.
- The docblocks state the *premise that was wrong* (#1918's "a `providerCode` is an exact structured value") rather than only the new rule. That is the repo's house style for a correction and it is the reason the next reader will not re-derive the old assumption.
- Test coverage is genuinely adversarial, not confirmatory: two different coarse codes staying apart, a coarse row refusing to mix with a no-code row, and an exact code still grouping across differing messages together pin all three edges of the new predicate. The `queryByText(...).toBeNull()` negative assertions are what make the display half actually tested rather than merely exercised.
- Correctly independent of #2805 and says so — the assumption was already wrong for any opaque carrier code.

## Documentation gaps

None binding. `docs/architecture-overview.md` does not describe the triage strip at this grain, and the module docblocks carry the rationale, which is the right home for it.

---

## Final Assessment

### Merge readiness
✅ **Approve**

Both halves of the flagged defect are closed, with the grouping half — the one that was easy to fix only cosmetically — closed at the key rather than at the render. The one thing I would still like is the coupling note in `shipment-retryability.ts`'s header (IMPORTANT above): the correctness of `isExactProviderCode` currently rests on a property of a *different* file that that file does not know it is providing, and there is no invariant script standing behind it. That is a follow-up-sized edit, not a reason to hold the fix.

### Priority fixes
1. **(IMPORTANT)** State the grouping coupling in `shipment-retryability.ts`'s file header — adding a retryability arm widens `isExactProviderCode`.
2. **(SUGGESTION)** Simplify or comment the always-true `retryabilityClass` conjunct on the coarse branch.
3. **(SUGGESTION)** Note the DPD `PARCEL_TOO_LARGE` fragmentation in the description.

norbert-kulus-blockydevs:
Addressed both findings from the final review:

1. **(IMPORTANT)** `shipment-retryability.ts`'s header now states the coupling to `isExactProviderCode` explicitly - widening a retryability arm there also widens what the grouping treats as an honest shared-cause claim, and no `check-*-mirror.mjs` script guards it.
2. **(SUGGESTION)** `ShipmentTriageStrip`'s coarse branch no longer carries the always-true `retryabilityClass` conjunct - simplified to `providerCode !== null` with `RETRYABILITY_LABEL.unknown` inlined, plus a comment.
3. **(SUGGESTION)** Called out the DPD `PARCEL_TOO_LARGE` fragmentation behaviour change explicitly in the PR description.

`type-check` and `lint` (0 errors) pass, and the full shipments/orders test suite (`src/features/shipments src/pages/shipments`) is 225/225 green.

Rebasing onto `main` and will squash-merge once CI is green.

piotrswierzy (approved):
## Re-review at `0a40fa4` (force-pushed since my APPROVE at `559f236`)

Re-verified every item from the prior pass against the current head. All three hold; nothing regressed in the force-push.

### Prior findings — verification

**1. Triage strip branches on exactness, and a coarse code renders alongside the raw message — ✅ RESOLVED, still holds.**

`shipment-triage-strip.tsx` computes `const codeIsExact = isExactProviderCode(group.providerCode);` and both the heading (`rejection code` vs `carrier message`) and the body branch on it. The coarse branch renders `sampleReason` when present **and** appends `(carrier code {providerCode} - {RETRYABILITY_LABEL.unknown})` when a code exists, so the operator's actionable signal survives and the strip states plainly that the code narrows nothing. The `null`-code case is unchanged (message only, no parenthetical). Three tests pin it: coarse keeps the raw message, coarse never claims `report the same rejection code` and does say `Cause not classified`, and exact still renders the code alone with `queryByText('Validation error')` asserted null — that last one is what stops the fix from over-correcting into #1918's behaviour.

**2. `MIN_CAUSE_KEY_LENGTH` now guards the whole coarse-or-absent branch, composing `{code}::{normalisedMessage}` — ✅ RESOLVED, still holds.**

`group-failed-shipments-by-cause.ts` splits cleanly on `isExactProviderCode(providerCode)`: exact keys on the bare code; otherwise it normalises first, `continue`s below the minimum, and only then composes `` `${providerCode ?? ''}::${normalised}` ``. The guard is now structurally impossible to skip on any key that leans on the lossy text half, which was the point. The `'500'` / `'404'` test (both normalise away, so the composite would degenerate to the bare bucket code) is the right test for that arm, and `does not mix a coarse-code row with a no-code row` pins the `?? ''` prefix doing real work rather than being cosmetic.

**3. Coupling note in `shipment-retryability.ts`'s header — ✅ LANDED.**

The `COUPLING (#2873 review)` block is present and says the two things it needed to: that widening an arm here also widens what `isExactProviderCode` treats as exact, and — the part that actually helps the next person — that **no `check-*-mirror.mjs` guards it**, so the only signal is the grouping tests, which would otherwise read as "an unrelated test broke". Naming the specific plausible mistake (`classifying a shipx.* family as 'permanent'`) is better than a generic warning; that is exactly the edit #2805's landscape invites.

### Remaining observation (non-blocking, not new)

`retryabilityClass` is now consumed only inside the `codeIsExact` branch, where `group.providerCode !== null` is guaranteed — so its own null-check is dead on that path, and the value is computed and discarded on the coarse path. Harmless, and the comment above it explains why it is still computed for both, so I would not churn the diff for it. The structural risk from the previous round (that `isExactProviderCode` answers "is this exact?" through a classifier written for a different question, in a hand-maintained mirror with no invariant script behind it) is unchanged — but that is now documented at the coupling point, which was the ask, and hardening it further belongs in whatever change actually widens a retryability arm.

### Verdict

✅ **Approve.** Both halves of the earlier finding survive the force-push intact, the coupling note landed, and the test coverage across grouping, the predicate itself and the strip is proportionate to a change that alters what an operator is told about a failure's cause. Merge once CI is green (`mergeable_state: blocked` is checks-pending, not a review state).

piotrswierzy (approved):
## Re-review at `b87ad8c` (was `0a40fa4`)

The force-push is a **rebase only**. `git diff 0a40fa4..b87ad8c` touches three files — `.gitignore`, `docs/operations/perf-lab-stand.md`, `perf/openlinker-throughput/bootstrap.sh` — all of which are #2872 arriving from `main`. Not one byte of the four `features/shipments` files changed. Re-approving on that basis, with each prior finding re-verified at the new head rather than assumed.

### Prior findings — all four still hold

1. **Strip branches on `codeIsExact`, and a coarse code renders ALONGSIDE the raw message with an `unknown` label** ✓
   `shipment-triage-strip.tsx`: the heading picks `'rejection code'` vs `'carrier message'` off `codeIsExact`; the else-branch renders `sampleReason` and then, when a code exists, `(carrier code {providerCode} - {RETRYABILITY_LABEL.unknown})`. The operator keeps the only actionable signal and is told plainly that the code narrows nothing — which is the whole point of #2873 over a naive null-check.

2. **`MIN_CAUSE_KEY_LENGTH` applies across the whole coarse-or-absent branch while composing `{code}::{normalisedMessage}`** ✓
   `group-failed-shipments-by-cause.ts`: the guard now sits inside the `else`, before the composite is built, with the comment stating the widened scope ("this whole branch rather than the `null` case alone"). Pinned by the `'500'` / `'404'` test, which is the exact case where the composite would otherwise degenerate to the bare bucket code.

3. **The coupling note landed in `shipment-retryability.ts`'s header** ✓
   Names `isExactProviderCode`, states that widening an arm here silently widens what the grouping treats as an honest shared-cause claim, and says explicitly that **no `check-*-mirror.mjs` guards it**. That last clause is what makes it load-bearing: in this repo an invariant without a script is an invariant a reader must be told is unguarded.

4. **Nit: `retryabilityClass` is dead on the coarse branch** ✓ — unchanged, still documented in place ("on the coarse branch it is always `'unknown'`"), and the JSX comment re-derives that from `isExactProviderCode`'s own test rather than restating it as a fact. Still not worth the churn.

### Nothing new introduced

The rebase pulled in #2805's `InpostHttpClient` fallback on `main` — which is the premise break this PR exists to answer, so it is now *actually* in the tree beneath the fix rather than merely cited. No interaction: `isExactProviderCode` derives exactness from `deriveRetryabilityClass`, and `shipx.*` classifies `'unknown'` there, so the new bucket codes take the coarse branch by construction. The `it.each` table asserts exactly that in both directions.

Test coverage remains the right shape — six grouping cases covering the four ways a coarse code must NOT collapse rows (differing messages, differing codes, message-normalises-away, coarse-vs-null), one asserting the exact-code path still groups across differing wording so #1918 is intact, and three strip-level cases covering the rendered copy on both branches.

✅ **Approve** — unchanged verdict, re-verified at `b87ad8c`.

Issue #2861, opened 2026-09-03
[TASK] DX — one-off Allegro sandbox latency baseline for the two stubbed endpoints


norbert-kulus-blockydevs:
Child of #2840. **Week 2**, before any #2847 or #2848 figure is quoted outside the team.

## Problem / Context

#2856 states the problem in its own Assumptions:

> **Real Allegro per-request latency is not recorded anywhere in this repo.** `perf/prestashop-baseline/README.md:237` explicitly leaves the Allegro side of the upstream budget to #1134, and no results file carries an Allegro p50 or p95 [...] So the pacing constant must be measured against the live sandbox once, or every figure taken with this stub is only internally consistent.

Two things make this urgent rather than tidy. **#2856 credits #1134 with owning the measurement and #1134 carries no such acceptance criterion** - its three ACs are scenarios, upstream call counts and a doc. And **#2854 currently hardcodes `STUB_PER_REQUEST_LATENCY_MS: '120'`** with no provenance at all, which is precisely the "the mock is now the model" trap #2840 names in its own trap list.

The consequence is specific: F1 is the epic's declared headline flow, and its number is a function of a latency nobody has taken. A stub answering faster than Allegro makes OpenLinker look faster than it is; a stub answering slower makes the lane caps look like the ceiling when they are not.

## Proposed Solution

One short run against the Allegro **sandbox**, recording p50 and p95 for exactly the two endpoints #2856 stubs, and writing the result into the stub's default configuration.

- `GET /order/events?from={cursor}&limit=100` - one request per poll tick (`libs/integrations/allegro/src/infrastructure/scheduler/allegro-scheduler-tasks.ts:105-127`)
- `GET /order/checkout-forms/{id}` - one request per ingested order (`libs/integrations/allegro/src/infrastructure/adapters/allegro-order-source.adapter.ts:722-724`)

Nothing else. The write endpoints fire only on cancellation or dispatch writeback, and the returns endpoints are behind two scheduler tasks that both default off.

This is a **characterisation probe, not a throughput run** - the epic's own data-source policy puts live sandboxes in that category. It needs no load, no k6 and no stand: a small script issuing N sequential requests against a real sandbox connection is enough, and sequential is the right shape because the stub paces per request rather than modelling concurrency.

## Classification

**Type**: DX
**Layer**: Shared (measurement)
[...]

Issue #2840, opened 2026-09-03
[EPIC] DX — performance measurement programme: what OpenLinker can do


piotrswierzy:
## Problem / Context

Exactly one performance campaign has ever been run against this system: **#2590 / `perf/prestashop-baseline`**. It is good work and it answers one question - *what does an OpenLinker sync cost the shop*, counted in the shop's own Apache access log, deliberately using an instrument OpenLinker cannot bias.

It answers no question about **OpenLinker's own capacity**, and its README says so:

> *#1134 asks for a k6 scenario set driving OpenLinker's **own** ingestion path … This harness answers one axis of that question for one platform, and it deliberately does not use k6: the instrument here is the shop's own Apache access log … k6 measures the caller, and the caller was the suspect.*

What is consequently unmeasured today:

| | State |
|---|---|
| Order ingestion throughput / latency | **2 samples ever** (`marketplace.order.sync`, 47 178 ms mean, results D - taken with `debug` logging and full response bodies, so an upper bound at best) |
| Stock propagation latency | derived only (#2609 measured ~145 jobs/h net queue growth pre-fix; results D measured net 0.0 jobs/h at 100k products post-fix under `fan-out=8/4`, but the latency itself was never measured) |
| Webhook ingress throughput | never driven at rate |
| Read path at row count | one endpoint, once (`sync-status`, 8-20 ms at ~31k rows) |
| Behaviour above one worker replica | **every published figure is single-replica** (#2302) |
| Lane caps | only `bulk` is derived from measurement; `realtime`, `fiscal`, `fan-out` are explicitly illustrative pending #1134 |
| Resource footprint (CPU / RAM / disk) | **never sampled at all**, at any dataset size or replica count |

Five enabling gaps, all real:

[...]

PR #2876, opened 2026-09-03
feat(perf): Allegro upstream stub for the measurement programme

### LINKED ISSUE #2856: [TASK] DX — Allegro upstream stub: two endpoints, 1+N cost model, multi-tenant by bearer token

norbert-kulus-blockydevs:
Closes #2856. Child of #2840. Paired with #2872 (the stand bootstrap, #2860) by the offer-id contract described below - **that is why this is a draft**: the two must agree on the offer-id space, and #2872 is not merged yet.

## What this is

A standalone, dependency-free `node:http` stub that OpenLinker is pointed at instead of the real Allegro API, so F1 (#2847) and F2 (#2848) measure OpenLinker rather than a third party's weather.

Pointing OpenLinker at it needs **no application code change**: `connection.config.apiBaseUrl` is an existing validated field. Two Allegro endpoints, plus `/me`, plus a `/__stub/` control surface. Everything else answers 404 in Allegro's own error shape, so a stray call fails loudly rather than looping - a 404 is non-retryable.

The honest cost model is **one request per poll tick plus one per ingestion attempt, and zero per line item**. Per attempt rather than per order because `getOrder` sits above item resolution, so a retry re-hydrates; and a cancel event costs zero hydrations because `handleSourceCancellation` never calls `getOrder`.

## Five properties that are load-bearing, not incidental

**Event ids are run-scoped, and structurally unrecognisable as a cursor.** The child dedupe key is `marketplace:{connectionId}:order:{eventKey}` where `eventKey` is the Allegro event id verbatim, and that becomes a Redis `jobdedup:` key with a **7-day TTL**. A repeat run that reuses ids therefore enqueues nothing - silently, with `enqueued: requests.length` returned unconditionally and the only trace a suppressed `debug` line - and every guard passes because the queue is empty *for that reason*. Verdict `VALID`, `n = 0`. Ids carry a run id so a repeat is fresh.

They also must not parse as a decimal counter. `compareOrderCursors` recognises exactly four shapes and treats anything else as `unrecognised`, which is explicitly **not** a regression. A recognised shape arms the guard, and a restarted stub reissuing lower ids then answers `regressed`, the cursor is never committed, the same page is re-read every minute forever, and the only output is one warn per tick. Verified empirically against the four regexes copied verbatim from `order-cursor.types.ts:33-58`:

```
mtlv2r456ebe52-000001   -> unrecognised (guard inert - correct)
12345                   -> DECIMAL_COUNTER (GUARD ARMED)
2026-09-03T18:31:43.824Z-> ISO_INSTANT (GUARD ARMED)
```

**It never answers 401, on any path, with any token, in any fault mode.** A 401 whose body mentions a token sends OpenLinker to the OAuth host, which is hardcoded with no per-connection override - the run would start depending on the public internet, and a failed refresh is classified `credential-rejected`, which flags the connection `needs_reauth`. Swept in the tests and again by hand across every path and token.

**Offer ids are `{tenant}-offer-{n}`**, matching exactly what #2860's bootstrap seeds as `identifier_mappings` rows. Without that agreement every order fails item resolution, persists `awaiting_mapping`, never reaches a destination create, and burns ten retry attempts over roughly 30 hours. `sku` is the same value, so it is not an independent axis. `STUB_OFFER_POOL_SIZE` must equal the bootstrap's `ALLEGRO_OFFER_POOL_SIZE`.

**Buyers vary their masked-email fixed part**, not only the `+txn` suffix. OpenLinker strips that suffix for any `@allegromail.` address, so under the default `email_fallback` identity mode a shared fixed part collapses every order onto **one** internal customer: destination customer provisioning is paid once, address provisioning disappears from order two onward, and an independent-orders workload becomes contention on one row plus one lock.

**Orders are minted in PLN.** A currency differing from the deployment's reporting currency puts a live NBP/ECB call inside every measured job, and not once per rate date either - the candidate rate date is a calendar day, so where the provider walks it back no row is written under the candidate and the read keeps missing.

Latency is applied **per tenant, not globally** - a global mutex would serialise the two tenants and the multi-tenancy requirement would then measure nothing.

## Verified end to end, not only by unit test

`node test.mjs` passes 14/14 and the Docker image builds and reports healthy, but the run that matters was against a **live OpenLinker install** with #2872's bootstrap creating the connections and the `Offer` mappings:

- both Allegro connection tests **passed through the real adapter**, which proves the `apiBaseUrl` override works with no code change and that no OAuth call left for the real `allegro.pl`
- five pushed orders carrying **three events each deduplicated to five `marketplace.order.sync` jobs** - the adapter's client-side dedupe by `checkoutForm.id`, exercised for real
- the resulting `order_records` rows reached **`recordStatus = 'ready'`** in PLN, with line items and a populated `placedAt`

Two things the live run showed that no unit test would have:

```
last_error | Deferred: lane 'realtime' per-scope cap reached within the claim
             batch - re-queued immediately, no attempt consumed (ADR-050)
```

The `realtime` per-scope cap of 2 binding on five queued orders, visibly - the constraint #2847's arithmetic is about, demonstrated rather than derived. And `marketplace.offerQuantity.reconcile` failing for both stub tenants, which is one of the five default-ON Allegro scheduler tasks hitting a path this stub deliberately does not serve. That is expected and is #2854's to silence by environment; it is recorded in the README rather than papered over.

The stack was returned to its exact prior state afterwards: connection count back to its original 13, zero stub orders, zero perf mappings, both keys removed, container and image deleted.

## Deliberately not served

The five default-ON Allegro scheduler tasks (`/sale/offer-events`, `/sale/categories`, `/sale/product-offers/{id}`, `/sale/offer-quantity-change-commands/{id}`, `/shipment-management/shipments/{id}`) answer 404. The issue frames disabling them in the stand's environment as the alternative, and that is #2854's. Serving them here would be inventing a second cost model nobody has measured.

## Not covered

- `pnpm lint` and `pnpm type-check` were not run: the change adds no TypeScript and `perf/` is not a pnpm workspace member (`pnpm-workspace.yaml` globs `apps/*`, `libs/*`, `libs/integrations/*` only), which is also why this ships zero dependencies rather than a package. `pnpm check:invariants` passes.
- The per-request latency constant is a placeholder. No Allegro p50 or p95 exists anywhere in this repo, so a figure taken against this stub is internally consistent and externally meaningless until the one-off sandbox baseline (#2861) lands. Stated in the README.

### DIFF
--- a/perf/openlinker-throughput/stubs/allegro/Dockerfile
+++ b/perf/openlinker-throughput/stubs/allegro/Dockerfile
@@ -1,0 +1,29 @@
+# Allegro upstream stub (#2856).
+#
+# Pinned to the exact base image tag the repo's root Dockerfile uses for its
+# `base` / `production` / `worker` stages, so this container is built from
+# the same node runtime as everything else in the stand.
+FROM node:22.23.1-alpine3.24
+
+# Nothing besides the node runtime is installed: server.mjs has zero
+# dependencies (node:http, node:crypto, node:url only), so there is no
+# install step and no package.json - see server.mjs's own header comment
+# for why this directory is deliberately not a pnpm workspace member.
+WORKDIR /app
+COPY server.mjs ./server.mjs
+
+# Surfaced back through GET /__stub/config, so a run manifest can record
+# exactly which stub build a measurement was taken against
+# (#2856 Build Specification "The stub's git sha for the manifest").
+ARG STUB_GIT_SHA=unknown
+ENV STUB_GIT_SHA=${STUB_GIT_SHA}
+
+ENV STUB_PORT=8080
+EXPOSE 8080
+
+# busybox wget ships in the alpine base image, so no extra package is
+# needed for a compose healthcheck to hit the control surface.
+HEALTHCHECK --interval=5s --timeout=3s --start-period=2s --retries=5 \
+  CMD wget -q -O - "http://127.0.0.1:${STUB_PORT}/__stub/health" || exit 1
+
+CMD ["node", "server.mjs"]
--- a/perf/openlinker-throughput/stubs/allegro/README.md
+++ b/perf/openlinker-throughput/stubs/allegro/README.md
@@ -1,0 +1,355 @@
+# Allegro upstream stub
+
+Implements GitHub issue #2856. A single-process `node:http` stub that serves
+just enough of the Allegro Public API for OpenLinker's real order-ingestion
+adapter to run against it unmodified, so the throughput programme (#2840)
+can measure OpenLinker's own cost without a live marketplace in the loop.
+
+## What it serves, and why only this
+
+Per #2856's own cost-model analysis of the real adapter, exactly two Allegro
+endpoints are on the order-ingestion path:
+
+| Method | Path | Cadence |
+|---|---|---|
+| `GET` | `/order/events?from={cursor}&limit={n}` | once per poll tick per connection |
+| `GET` | `/order/checkout-forms/{id}` | once per ingested order (zero per line item) |
+
+Plus `GET /me`, reached by the connection tester and by #2860's bootstrap
+readiness check. Everything else answers `404` in Allegro's own error shape
+(`{"errors":[{"code":"NotFound","message":"..."}]}`), because a 404 is
+non-retryable in OpenLinker's Allegro retry classifier - a stray call fails
+loudly on attempt 1 instead of burning a retry ladder.
+
+## What it deliberately does not serve
+
+- `PUT /order/checkout-forms/{id}/fulfillment`, `POST
+  /order/checkout-forms/{id}/shipments` - fire only on a cancellation feed
+  event or a dispatch writeback, off the ingestion path this stub exists to
+  measure.
+- The returns endpoints - gated behind two scheduler tasks that both default
+  off.
+- OAuth token exchange (`/auth/oauth/token`) - the token host is hardcoded in
+  the real adapter (`allegro-token-refresh.service.ts`) and cannot be
+  repointed by `config.apiBaseUrl`. **The workaround is credential shape, not
+  code**: store the connection's credentials with `accessToken` only - no
+  `expiresAt`, no `refreshToken`, no `clientId`/`clientSecret`. With
+  `expiresAt` absent, the adapter's token-freshness check short-circuits
+  before every request and no token call is ever made. See #2860's bootstrap
+  for where this is wired.
+
+## The one hard rule: never answer 401
+
+If the stub ever answered 401, the real adapter would attempt a token
+refresh against the real, hardcoded `allegro.pl` host - so a run would start
+depending on the public internet, and a failed refresh flags the connection
+`needs_reauth`. Every code path in `server.mjs` that could plausibly want to
+say "unauthorized" (a missing bearer token, an unrecognised one, a malformed
+control-endpoint request) instead degrades to something else - see
+"Multi-tenancy" below. `test.mjs` asserts this directly across every served
+path, every fault mode, and several bogus/missing tokens.
+
+## Design decisions and why
+
+**Transport: `node:http`, zero dependencies, standalone.**
+`pnpm-workspace.yaml` globs only `apps/*`, `libs/*` and `libs/integrations/*`,
+so `perf/` is not a pnpm workspace member and cannot be a package without
+editing that file. The `perf/prestashop-baseline` precedent ships no
+`package.json` either. Two endpoints plus a control surface do not need a
+framework.
+
+**State lives in memory, keyed by bearer token.** This is safe only because
+ids are run-scoped and monotone (below) - a process restart resets counters
+to zero, and that is fine precisely because a fresh run is never assumed to
+continue a prior run's id space.
+
+**Event ids are run-scoped, never process-scoped.** `eventKey` becomes the
+child job's Redis dedupe key (`marketplace:{connectionId}:order:{eventKey}`,
+7-day TTL). If ids were a function of process start alone, repeating a
+scenario within that window would mint the identical keys the first run
+already reserved and every enqueue would silently no-op (`n = 0`, cursor
+still commits, and nothing in the result distinguishes a fully-deduped tick
+from a fresh one). Ids are `{runId}-{6-digit-seq}`. `runId` defaults to
+`STUB_RUN_ID` at boot, or is minted fresh automatically; a driver can also
+start a brand-new run **without restarting the container** via
+`POST /__stub/run`, which is what makes a same-process repeat produce
+disjoint ids.
+
+**Cursor shape is deliberately unrecognised by the real regression guard,
+not a decimal counter.** `compareOrderCursors` in OpenLinker core recognises
+four cursor shapes (decimal counter, ISO instant, naive wall clock,
+wall-clock keyset) and treats anything else as `unrecognised`, which is
+*never* read as a regression. A hyphenated `{runId}-{seq}` id can never match
+any of the four regexes (the hyphen alone rules out the decimal-counter
+shape), so a stub restart that resets its counter to zero can never trip the
+`regressed` guard and wedge a connection. Using a plain decimal counter would
+have armed that guard for no benefit.
+
+**An unknown `from` cursor is tolerated, not 404'd.** OpenLinker's own
+`connection_cursors` row persists across a stub restart; the stub's
+in-memory state does not. A `from` value the stub cannot place (wrong run,
+malformed, or simply unknown) is treated as "you are already caught up" -
+the response carries no events and echoes back the current head as
+`lastEventId`, with a warning in the request log, so the connection is never
+told to replay history it may have already processed under a previous run.
+
+**The driver pushes; the stub never self-generates on a schedule.** Every
+order in the stub exists because `POST /__stub/tenants/{tenant}/orders` was
+called. This is what makes order *arrival rate* a driver-controlled
+independent variable rather than a stub constant.
+
+**Control endpoints live under `/__stub/`, deliberately unauthenticated.**
+No Allegro path uses that prefix, so it can never collide with the served
+surface or be mistaken for a 404-able Allegro path. It carries no auth
+because it is a local perf-harness control plane that OpenLinker itself
+never reaches - there is no principal to check.
+
+**Request log: JSON lines on stdout.** The lab stand's compose service
+declaration for this image has no volume mount, so a file-based log would
+die with the container. One line per request: `ts`, `runId`, `tenant`,
+`method`, `path`, `status`, `latencyMs`.
+
+**Latency is applied per tenant, not via a global mutex.** A shared lock
+would make two tenants serialise, which would make the multi-tenancy
+scenario measure nothing about concurrency at all. Each request independently
+`setTimeout`s for its configured latency; two tenants' requests interleave
+freely.
+
+**Dockerfile lives in the stub's own directory**, `FROM
+node:22.23.1-alpine3.24` - the exact tag the repo's root `Dockerfile` pins
+for its `base`/`production`/`worker` stages. No install step: `server.mjs`
+has zero dependencies.
+
+## The offer-id contract with #2860
+
+`OrderIngestionService` resolves every line item's `offer.id` against an
+`identifier_mappings` row (`entityType='Offer'`, scoped per connection) to a
+live `ProductVariant`. Without a pre-seeded mapping, every stub order dies
+`awaiting_mapping` after exhausting its retries. This stub therefore mints
+offer ids as a **deterministic function of tenant and index**:
+
+```
+{tenant}-offer-{n}      for n in 1..STUB_OFFER_POOL_SIZE
+```
+
+where `{tenant}` is the connection's configured name (`perf-allegro-a` /
+`perf-allegro-b` by default) - the same value `#2860`'s bootstrap must use
+when it seeds `identifier_mappings` rows ahead of any order being pushed.
+
+**`STUB_OFFER_POOL_SIZE` must equal `ALLEGRO_OFFER_POOL_SIZE` in the
+bootstrap.** `sku` on the resulting order line is the same value as
+`offer.id` (`allegro-order-source.adapter.ts:802`), so it is not an
+independent axis - the seeder does not need to track it separately.
+
+Offer indices are assigned round-robin across the pool as line items are
+minted (not per-order), so a small pool is exercised evenly even with many
+orders.
+
+## The buyer-identity decision
+
+`OL_CUSTOMER_IDENTITY_MODE` defaults to `email_fallback`, and Allegro's
+masked-email normalizer strips everything from `+` onward on any
+`@allegromail.` address before hashing (`fixedPart+transactionId@...`, with
+a stable `fixedPart` per real buyer). Minting a distinct `buyer.id` per
+order but only varying the `+transactionId` suffix would therefore collapse
+every synthetic order onto **one** internal customer - the normalizer erases
+exactly that suffix, so per-order destination cost is understated from order
+two onward and an independent-orders workload becomes contention on one
+shared customer row.
+
+This stub varies the masked-email **fixed part** itself:
+`buyer{N}+tx{orderNumber}@allegromail.pl`, where `N` cycles across
+`STUB_BUYER_POOL_SIZE` distinct slots (default 50). A driver wanting a
+"warm" (repeat-buyer) scenario instead of "cold" (one buyer per order) sets
+`STUB_BUYER_POOL_SIZE=1` and records that choice in the run manifest.
+
+## Multi-tenancy
+
+One process, tenants distinguished by the `Authorization: Bearer` token - a
+path prefix cannot work (the real HTTP client builds URLs with `new
+URL(path, baseUrl)`, which discards any path segment already in `baseUrl`).
+
+Configure via `STUB_TENANTS` (default
+`stub-token-a=perf-allegro-a,stub-token-b=perf-allegro-b`), a comma-separated
+list of `token=name` pairs. Each tenant keeps an independent cursor, order/
+event history and fault state.
+
+**An unrecognised or missing bearer token still serves** - it is bucketed
+under a shared `unknown` tenant rather than answering 401 (see "the one hard
+rule" above). This bucket starts empty and is never populated by a push
+(pushes are addressed by tenant *name*, and only configured tenants have
+one), so it exists purely to keep every request answerable without ever
+saying 401.
+
+## Pacing and fault injection
+
+| Env var | Default | Meaning |
+|---|---|---|
+| `STUB_PER_REQUEST_LATENCY_MS` | `120` | Delay applied to every served request |
+| `STUB_LATENCY_EVENTS_MS` | falls back to the above | Per-endpoint override for `/order/events` |
+| `STUB_LATENCY_CHECKOUT_MS` | falls back to the above | Per-endpoint override for `/order/checkout-forms/{id}` |
+
+Latency is a fixed constant per endpoint, not a distribution - #2856 names
+that as a real limitation (a mean cannot produce a tail), and a run manifest
+should record it as such rather than imply a realistic latency profile.
+
+Fault injection (`POST /__stub/tenants/{tenant}/fault`):
+
+```bash
+# Rate-limit tenant A for the next several requests, with an explicit
+# Retry-After the real client actually reads and sleeps on.
+curl -X POST http://localhost:8080/__stub/tenants/perf-allegro-a/fault \
+  -H 'Content-Type: application/json' \
+  -d '{"mode": "429", "retryAfterSeconds": 5}'
+
+# Simulate an outage.
+curl -X POST http://localhost:8080/__stub/tenants/perf-allegro-a/fault \
+  -H 'Content-Type: application/json' \
+  -d '{"mode": "503", "retryAfterSeconds": 10}'
+
+# Hang past the real client's 30s abort (holdMs defaults to 31000; a smaller
+# value is useful only for testing the stub itself, never for a real run).
+curl -X POST http://localhost:8080/__stub/tenants/perf-allegro-a/fault \
+  -H 'Content-Type: application/json' \
+  -d '{"mode": "timeout"}'
+
+# Clear a fault.
+curl -X DELETE http://localhost:8080/__stub/tenants/perf-allegro-a/fault
+```
+
+A fault, once set, applies to every subsequent request from that tenant
+against `/order/events`, `/order/checkout-forms/{id}` and `/me` until
+cleared - it is not consumed after one request. Both 429 and 503 carry
+`Retry-After` as integer seconds (the only response header OpenLinker's
+Allegro client reads) and the Allegro error-body shape
+`{"errors":[{"code":"...","message":"..."}]}`.
+
+## Control surface (`/__stub/`)
+
+All bodies are JSON, no auth required.
+
+### `GET /__stub/health`
+
+For a compose healthcheck. Returns `{"status": "ok", "runId": "..."}`.
+
+### `GET /__stub/config`
+
+The full resolved configuration - the run id, git sha, every latency value,
+the offer pool size, the buyer pool size and the configured tenant names -
+so a run manifest can assert against the stub's own running state rather
+than only against what it was started with.
+
+```bash
+curl http://localhost:8080/__stub/config
+```
+
+### `POST /__stub/run`
+
+Starts a new run: mints (or accepts) a run id and resets every tenant's
+event/order history, counters and faults. This is what lets a driver repeat
+the same scenario multiple times **within one process lifetime** and get
+disjoint dedupe keys each time, without restarting the container.
+
+```bash
+curl -X POST http://localhost:8080/__stub/run \
+  -H 'Content-Type: application/json' \
+  -d '{"runId": "campaign-run-2"}'
+# or, to auto-generate one:
+curl -X POST http://localhost:8080/__stub/run -d '{}'
+```
+
+### `POST /__stub/tenants/{tenant}/orders`
+
+The driver's entry point for order arrival. `{tenant}` is the configured
+tenant **name** (e.g. `perf-allegro-a`), never the bearer token.
+
+```bash
+curl -X POST http://localhost:8080/__stub/tenants/perf-allegro-a/orders \
+  -H 'Content-Type: application/json' \
+  -d '{"count": 10, "lineItemsPerOrder": 3, "eventsPerOrder": 1}'
+```
+
+- `count` (default 1): how many orders to mint.
+- `lineItemsPerOrder` (default 1): distinct line items per order, each drawn
+  round-robin from the offer pool.
+- `eventsPerOrder` (default 1): how many `/order/events` entries reference
+  the same `checkoutForm.id` - set to e.g. 3 to exercise the real adapter's
+  client-side dedupe-by-checkoutFormId behaviour.
+
+Minted orders are served from `/order/events` on the very next poll of that
+endpoint - the stub does not push anything itself. Pushing faster than
+OpenLinker's own discovery ceiling (one page per cron tick, no loop inside
+the adapter) simply queues inside the stub's in-memory event list; that is
+expected and should be stated in the run manifest rather than read as a bug.
+
+### `POST` / `DELETE /__stub/tenants/{tenant}/fault`
+
+See "Pacing and fault injection" above.
+
+### `GET /__stub/tenants/{tenant}/stats`
+
+Per-tenant counters: requests served per endpoint, orders pushed, events
+emitted, the current cursor head, and the active fault (if any).
+
+```bash
+curl http://localhost:8080/__stub/tenants/perf-allegro-a/stats
+```
+
+This is the source of truth for the "1 request per poll tick + 1 per
+ingested order + 0 per line item" cost-model assertion.
+
+## Pointing a connection at the stub
+
+```json
+{
+  "config": { "environment": "production", "apiBaseUrl": "http://allegro-stub:8080" },
+  "credentials": { "accessToken": "stub-token-a" }
+}
+```
+
+`config.environment` remains mandatory on every Allegro connection
+(`@IsIn(AllegroEnvironmentValues)`) but, with `apiBaseUrl` present, its only
+remaining effect on this path is a cosmetic Sales Center deep link. Two
+connections that differ only in `credentials.accessToken` (`stub-token-a`
+vs. `stub-token-b`) are two independent tenants against one stub process.
+
+Verify with `POST /connections/:id/test`, which reaches the stub's `/me` -
+this is the bootstrap readiness check, not a manual step.
+
+## Running the tests
+
+```bash
+node test.mjs
+```
+
+No install step, no pnpm workspace membership needed - `node:test` and
+`node:assert` are both node core. The suite imports `server.mjs` directly
+and starts it on an ephemeral port per test file (not per test - each test
+calls `POST /__stub/run` first for isolation from prior tests in the same
+process).
+
+## What #2856 asks for that this implementation does NOT cover
+
+Two acceptance criteria in #2856 are end-to-end properties of a real
+OpenLinker install pointed at this stub, not something the stub alone can
+satisfy or a unit test can assert:
+
+- *"one order reaches `order_records.recordStatus='ready'` and a destination
+  create"* - needs a running OpenLinker api/worker plus #2860's seeded
+  `identifier_mappings`. This stub only guarantees its half of that contract
+  (see "The offer-id contract with #2860" above).
+- *"every default-ON Allegro scheduler task is either served by the stub or
+  disabled... verified by zero dead `sync_jobs` rows"* - the five scheduler
+  tasks #2856 lists (`offers-sync`, `taxonomy-sync`, `offer-status-sync`,
+  `quantity-ack-reconcile`, `shipment-status-sync`) hit endpoints
+  (`/sale/offer-events`, `/sale/categories`, `/sale/product-offers/{id}`,
+  `/sale/offer-quantity-change-commands/{id}`,
+  `/shipment-management/shipments/{id}`) that are genuinely out of scope for
+  the two-endpoint ingestion path this stub measures, and #2856 explicitly
+  frames disabling them in the stand's environment as the alternative to
+  serving them. That is a stand-configuration decision (#2854/#2860's job),
+  not code this stub ships. Left unserved here, they correctly 404 - fast
+  and loud rather than silently degrading the measurement.
+
+Everything else in the acceptance-criteria list is implemented and covered
+by `test.mjs`.
--- a/perf/openlinker-throughput/stubs/allegro/server.mjs
+++ b/perf/openlinker-throughput/stubs/allegro/server.mjs
@@ -1,0 +1,689 @@
+#!/usr/bin/env node
+'use strict';
+
+/**
+ * Allegro upstream stub - #2856.
+ *
+ * Serves the two Allegro Public API endpoints OpenLinker's order-ingestion
+ * path actually calls (GET /order/events, GET /order/checkout-forms/{id}),
+ * plus GET /me for the connection tester, plus a driver-facing control
+ * surface under /__stub/. Everything else answers 404 in Allegro's own error
+ * shape, so a stray call fails loudly (404 is non-retryable per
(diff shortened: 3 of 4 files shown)


piotrswierzy (approved):
Strong work — the wire shapes match what `AllegroOrderSourceAdapter` actually parses, isolation is clean (zero deps, no `@openlinker/*` imports, `isMain` guard so importing binds no port, unreachable from any application path), and the never-401 rule is correctly identified as the single escape hatch to the public internet and swept in tests. Approving; none of the below blocks, but they affect whether the instrument stays correct and whether F1/F2's numbers transfer.

**Durability of the instrument**

1. `test.mjs:219-222` — the four cursor-shape regexes (`DECIMAL_COUNTER`, `ISO_INSTANT`, `NAIVE_WALL_CLOCK`, `WALL_CLOCK_KEYSET`) are hand-copied verbatim from `order-cursor.types.ts:33-58` with no mirror guard. This repo has 21 `scripts/check-*-mirror.mjs` under `pnpm check:invariants` for exactly this class of copy. If core ever widens a shape, this test stays green while the guard *arms* in production — which is the `regressed` / cursor-never-committed / same-page-re-read-forever failure the PR body itself describes.
2. **CI never runs `node test.mjs`.** `perf/` appears in no workflow and is outside the pnpm workspace, so `pnpm test` can't reach it either. Fourteen tests that never execute will rot, and every F1 (#2847) / F2 (#2848) figure depends on this instrument being right. A `paths`-filtered step would do.
3. `test.mjs:103-112` — the unknown-cursor test pushes zero orders first, so `tenant.seq` is 0 and "skip to newest" (`server.mjs:325`) and "replay everything" produce the identical empty page. The behaviour that matters — not re-minting hydration work for already-processed events after a restart — is unasserted. Push orders, *then* send the unknown cursor, then assert `events: []`.

**Fidelity gaps that will bias the measurement**

4. `server.mjs:239-249` — minted orders carry `delivery.cost` and `delivery.address` but no `delivery.method`. `resolveShipping` (`allegro-order-source.adapter.ts:988-994`) requires `method?.id`, so every stub order ingests with `shipping: undefined` and is passed that way to the destination create. Carrier-mapping resolution — a real per-order cost on the path F1 measures, and the delicate area ADR-016/#898 exists over — silently drops out of every run.
5. `README.md:340-352` — deferring all five default-ON tasks to "#2854 disables them in the environment" doesn't cover `destination.taxonomy.sync`. #2856 notes it's a **direct enqueue at connection create** (`connection.service.ts:604-633`), not a cron, so `OL_SCHEDULER_ENABLED=false` can't suppress it: `/sale/categories` 404s, which is non-retryable, giving a dead `sync_jobs` row per perf connection whatever the stand sets. Serving an empty `categories: []` is three lines; otherwise name it as the exception the environment can't close.
6. `server.mjs:216` — `quantity` is hardcoded to 1 and isn't a control-endpoint parameter, so `price.amount * quantity` (`:758`) and the per-line reservation claim are only ever exercised at 1. `lineItemsPerOrder` varies line *count*, a different axis. A `quantityPerLine` knob would close it.

**Minor**

- `server.mjs:225` — `@type {import('./allegro-checkout-form-shape')…}` points at a module that doesn't exist. Inert (`.mjs`, no tsconfig under `perf/`) but it advertises type-backed provenance the wire shape doesn't have.
- `server.mjs:274` — `padStart(6)` keeps lexicographic ordering aligned with numeric only below 1e6, and the adapter's dedupe is a string compare (`:332-339`), so the highest-id-wins rule inverts past that. Out of reach for a realistic run; worth an inline note.
- `Dockerfile:12` runs as root; `USER node` is free on the alpine base.
- `latencyFor` (`:94-102`) re-reads `process.env` per request rather than the once-captured `CONFIG`, so latency can diverge from what `GET /__stub/config` reported.

Nice calls worth naming: deliberately choosing an **unrecognised** cursor shape (a hyphen in `{runId}-{seq}`) so `compareOrderCursors` leaves the regression guard inert, rather than reaching for the obvious decimal counter which would arm it and wedge the connection on restart — subtle, correct, and reasoned inline. Also: every `amount` emitted as a decimal string per `allegro-api.types.ts` (the adapter's `parseFloat` would have tolerated numbers and hidden the divergence); varying the `+tx` suffix precisely *because* it must be irrelevant, so the stub exercises the email normalizer rather than dodging it; `recordRequestCount` running before `applyFault` so the 1+N cost model stays honest under injection; and the README's "what #2856 asks for that this does NOT cover" section.

Issue #2605, opened 2026-08-27
[BUG] Integration/PrestaShop — timezone and cursor handling in the order source loses orders


norbert-kulus-blockydevs:
## Problem / Context

Three independent mechanisms in the PrestaShop order source lose orders permanently today, and none of them logs anything.

1. **Filter built with the worker's clock, compared against the shop's.** `date_add` / `date_upd` are `DATETIME` — no timezone, no record of who wrote them. Today this survives only by accident: the same naive string is formatted and parsed in the same worker timezone. A container timezone change, a manual cursor reset, or an ISO value moves it into one of the failure modes. Offset the wrong way and nothing returns for the length of the offset while the cursor stays put, because the adapter returns the input cursor on an empty page.
2. **No `sort` in the query.** PrestaShop orders by `id`, while the cursor takes `max(date_upd)` of that subset. An older order with a higher id and a lower date is skipped forever.
3. **Second-level precision with a strict `>`.** Three orders in the same second, one ends the page and advances the cursor — the other two never pass the condition again.

## Proposed Solution

Keyset cursor over `(date_upd, id)` with a matching `ORDER BY`, dates carrying an explicit offset end to end, and a bounded read-position lookback that never moves the watermark backwards.

## Classification

**Type**: Integration
**Layer**: Infrastructure
**File(s)**: `libs/integrations/prestashop/src/infrastructure/adapters/prestashop-order-source.adapter.ts`, `libs/integrations/prestashop/src/infrastructure/http/prestashop-query.builder.ts`

## Acceptance Criteria

- [ ] Query sorts by `(date_upd, id)` and the cursor is a keyset over the same tuple
- [ ] Orders sharing a second are all ingested
- [ ] An empty page does not freeze the cursor
- [ ] A container timezone change does not change which orders are returned
- [ ] Tests added or updated for non-trivial logic

norbert-kulus-blockydevs:
Done in the epic branch `epic/2590-prestashop-no-module`, reviewed and fixed on PR #2627. Closing so the board reflects it; the branch is merged to `main` as one PR.

Issue #904, opened 2026-05-30
Phase 3 — Schedule the PrestaShop order poll as the reconciliation backstop


piotrswierzy:
Part of #900.

With webhooks as the low-latency primary path (Phases 1–2), add a periodic PrestaShop order poll as the **reconciliation backstop** that heals missed/dropped webhooks.

## Context
PrestaShop currently has **no** registered order-poll scheduler task (only `prestashop-fulfillment-status-sync`). Allegro registers `allegro-orders-poll` every 5 min. The `PrestashopOrderSourceAdapter` + `marketplace.orders.poll` job already exist and work over the `date_upd` watermark cursor — they're just not scheduled.

## Scope
1. Register a PrestaShop order-poll task in `buildPrestashopSchedulerTasks()` (mirror Allegro's `orders-poll`), enqueuing `marketplace.orders.poll`. Choose a relaxed cadence (e.g. every 10–15 min) since it's a backstop, not the workhorse.
2. Confirm convergence: both poll and webhook end at the idempotent `syncOrderFromSource`, so a webhook-ingested order is not re-created by a subsequent poll (idempotent on internal order id). Add/extend a test that asserts no double-ingest when both paths fire for the same order.
3. Make the cadence configurable (env) and document the webhook-primary / poll-backstop posture.

## Acceptance
- PrestaShop orders are ingested automatically with no manual job enqueue, even if a webhook is missed.
- Webhook + poll for the same order produce exactly one ingested order.
- Cadence documented and env-configurable.

Issue #2592, opened 2026-08-27
[TASK] Integration/PrestaShop — land the adapter waste removal (3x product read, dead resolver cache, per-job currency pair)


norbert-kulus-blockydevs:
## Problem / Context

Measured on a live shop (2026-08-27): the catalogue path spends **7.96 requests per SKU** and a 100-product tick takes **977 s**. Four independent sources of waste, all confirmed in the access log:

- the same `GET /products/{id}` is fetched **exactly three times** per product — `prestashop-product-master.adapter.ts:187` (`getProduct`), `:395` (`getProductVariants`) and the VAT resolver via `readProductTaxRate`
- every capability resolution builds a fresh factory (`prestashop-plugin.ts:150`), so the six resolver caches held as factory fields (`prestashop-adapter.factory.ts:62-89`) never survive a child job. `prestashop-shop-currency.resolver.ts:16-23` documents this in a code comment: *"every build re-reads regardless of TTL"*
- with `config.currency` unset, each adapter build costs `/configurations` + `/currencies` (`prestashop-adapter.factory.ts:138-141`) — **+2 requests per child job**
- the VAT rate is cached per product instead of per tax-rules group

## Proposed Solution

Work is done and pushed on branch `2489-etap0`. This issue tracks review, GPG signing and merge.

Four changes: hoist the resolver cache out of the per-resolution factory into the plugin closure, memoize `GET /products/{id}` within one job, resolve the shop currency once, cache the VAT rate by tax-rules group.

## Classification

**Type**: Integration
**Layer**: Infrastructure
**File(s)**: `libs/integrations/prestashop/src/prestashop-plugin.ts`, `src/infrastructure/adapters/prestashop-adapter.factory.ts`, `prestashop-product-master.adapter.ts`, `src/infrastructure/provisioners/*.resolver.ts`

## Acceptance Criteria

- [ ] Commits are GPG-signed (`main` requires verified signatures)
- [ ] **Do not merge before the baseline numbers have been read** — after the merge, baseline A is no longer reproducible
- [ ] Requests per SKU drop to ~3.97 and the 100-product tick to ~471 s, reproduced against the control run
- [ ] `GET /products/{id}` is fetched once per product per job
- [ ] No `/configurations` + `/currencies` pair per child job
- [ ] Tests added or updated for non-trivial logic

norbert-kulus-blockydevs:
[...]

PR #2884, opened 2026-09-04
fix(integrations-prestashop): fall back to id-ordered order feed when a shop refuses date_upd sort/filter

### LINKED ISSUE #2877: [BUG] Integration — PrestaShop order-feed poll fails on shops that reject date_upd as a sortable/filterable webservice field
## Problem / Context

`PrestashopOrderSourceAdapter.listOrderFeed` (`libs/integrations/prestashop/src/infrastructure/adapters/prestashop-order-source.adapter.ts:153`) unconditionally requests the PrestaShop webservice `orders` resource sorted by `sort=[date_upd_ASC, id_ASC]`, per the `(date_upd, id)` keyset-cursor design (#2605, documented in `docs/architecture-overview.md` § Orders → "Order analytics read model" area and inline in the adapter's own header comment).

On PrestaShop 9.0.2 (verified live, both via direct `curl` against a running shop and via the OL worker consuming it), the webservice **rejects `date_upd` outright** as a sort/filter field on `orders`:

```
GET /api/orders?display=full&sort=[date_upd_ASC,id_ASC]&limit=100
→ 500
{"errors":[{"code":38,"message":"Unable to filter by this field. However, these are available: id, id_address_delivery, id_address_invoice, id_cart, id_currency, id_lang, id_customer, id_carrier, current_state, module, invoice_number, invoice_date, delivery_number, delivery_date, valid, shipping_number, note, id_shop_group, id_shop, secure_key, payment, recyclable, gift, gift_message, mobile_theme, total_discounts, total_discounts_tax_incl, total_discounts_tax_excl, total_paid, total_paid_tax_incl, total_paid_tax_excl, total_paid_real, total_products, total_products_wt, total_shipping, total_shipping_tax_incl, total_shipping_tax_excl, carrier_tax_rate, total_wrapping, total_wrapping_tax_incl, total_wrapping_tax_excl, round_mode, round_type, conversion_rate, reference, for i18n fields:"}]}
```

`date_upd` (and `date_add`) are absent from the allowed field list this shop version exposes for `orders` — only `id` and other flat scalar/reference columns are sortable/filterable there.

**Impact**: `marketplace.orders.poll` fails on every attempt against this shop version (500, retried per the adapter's retry policy, then the job dead-letters). This is usually masked because PrestaShop's webhook is the fast/primary ingestion path and poll is only a reconciliation backstop (`docs/architecture-overview.md` § Data Flow 4, #904 — "webhook = trigger, poll = reconciliation backstop"). It becomes a hard blocker for any connection with no webhook installed, since poll is then the *only* order-ingestion path and it always fails. Found while building live E2E order-seeding for the analytics/currency-coverage epic (#2452): a freshly-created connection with no webhook configured made poll the sole path, and it never surfaced a single order event.

## Proposed Solution

Detect the webservice's "field not sortable/filterable" refusal (code 38, naming the rejected field) on the `date_upd` sort attempt and degrade gracefully for that connection, rather than retrying the same doomed request and dead-lettering.

Two candidate approaches to weigh during implementation (not pre-decided):

1. **Reactive fallback**: on the code-38 refusal, retry once with `sort=[id_ASC]` only, and persist a per-connection flag (e.g. on `connection_cursors` or `Connection.config`) so subsequent polls skip straight to the `id_ASC` request instead of re-discovering the refusal every tick.
2. **Proactive capability probe**: read the shop's advertised filterable/sortable field set once (PrestaShop exposes a schema-style meta response for a resource) and cache the decision per connection before ever attempting a `date_upd` sort.

Either way, **this must not silently break the #2605 keyset-cursor correctness guarantee**: the design deliberately avoids offset-based paging specifically to avoid dropping rows whose `date_upd` was bumped mid-drain (see the adapter's own header comment, lines ~5, ~61, ~129). Falling back to `id_ASC`-only changes the semantic meaning of the reconciliation pass — it stops being a true "what changed recently" reconciliation and becomes closer to a full id-ordered replay. This tradeoff must be:
- explicitly logged/warned per connection when the fallback engages, so it's operator-visible rather than a silent degradation,
- documented in the adapter's own header comment and in `docs/architecture-overview.md` alongside the existing #2605 description,
- considered against whether the fallback path can still honour the cursor's resumability guarantees (a partial page failure should still be safely retryable).

## Classification

**Type**: Integration
**File(s)**:
- `libs/integrations/prestashop/src/infrastructure/adapters/prestashop-order-source.adapter.ts` (primary — `listOrderFeed`, the `sort` construction around line 153)
- `docs/architecture-overview.md` (update the #2605 order-feed description if the fallback semantics change the documented guarantee)

## Dependencies

- None. Independent of epic #2452 (analytics/currency-coverage) — this is a pre-existing platform-compatibility gap in the order-ingestion adapter, unrelated to the analytics feature itself. It was discovered incidentally while building E2E order-seeding for that epic's Phase 9 test suite, but the fix belongs on its own branch off `main`.

## Assumptions

- The exact PrestaShop version boundary where `date_upd` becomes unsortable is not established beyond "reproduced live on 9.0.2" — the fix should be defensive (react to the actual webservice error) rather than version-sniffing, since PrestaShop version reporting from the webservice itself is not reliable across all shop configurations.
- Whether a "probe once and cache" or "reactive fallback per failure" approach is preferable is left to the implementer's judgement during design; both are listed as candidates above.

## Acceptance Criteria

- [ ] `marketplace.orders.poll` against a PrestaShop shop that rejects `date_upd` as a sort field succeeds (returns order feed items) instead of 500-ing and dead-lettering.
- [ ] The degraded-ordering fallback (if `id_ASC`-only is chosen) is explicitly logged as a warning naming the connection, so it is operator-visible.
- [ ] A shop that DOES support `date_upd` sorting is unaffected (byte-identical behaviour to today).
- [ ] The keyset-cursor resumability/idempotency guarantee is preserved or the change in guarantee is explicitly documented in the adapter's header comment and `docs/architecture-overview.md`.
- [ ] Tests added or updated for non-trivial logic (unit test simulating the code-38 refusal; adapter-level integration test if the existing suite has one for `listOrderFeed`).
- [ ] No architecture boundary violations (CORE ↔ Integration) — the fix stays entirely within `libs/integrations/prestashop`.

norbert-kulus-blockydevs:
## Summary

- `PrestashopOrderSourceAdapter.listOrderFeed` hardcoded `sort=[date_upd_ASC,id_ASC]` plus a `date_upd`-based `updatedAfter` filter for the `(date_upd, id)` keyset cursor (#2605). Some PrestaShop installs (verified live on 9.0.2) refuse `date_upd` as a sort/filter field on the `orders` webservice resource with error code 38, so `marketplace.orders.poll` 500ed on every attempt and dead-lettered — fatal for a connection with no webhook installed, since poll is then the only order-ingestion path.
- The adapter now detects that specific refusal (code 38 in the JSON error body), remembers it per connection via a new `PrestashopOrderFeedCapabilityCache` (held as a process-singleton on `PrestashopAdapterFactory` so the answer survives across the adapters it rebuilds per capability resolution), and falls back to an id-ordered read (`id > lastOrderId`, `sort: ['id_ASC']`) for every later poll on that connection. The very first affected poll retries inline and succeeds too, not only later ones.
- This narrows the #2605 reconciliation guarantee on an affected shop: new orders are still reliably surfaced, but an update to an order whose id has already been read is no longer re-observed by poll, since nothing asks the shop to re-sort by `date_upd` any more. The webhook path (#904) is unaffected and remains the primary, low-latency route — poll is only ever the reconciliation backstop, and it now narrows rather than disappears entirely on these shops. The cursor's wire format is unchanged either way, by design (documented in the adapter's own header comment).
- Added a small `idAfter` filter to `PrestashopQueryFilters`/`PrestashopQueryBuilder` (`filter[id]=>[N]`) as the fallback pagination key, mirroring the existing `updatedAfter` shape.
- Documented the fallback in the adapter's header comment and in `docs/architecture-overview.md` alongside the Orders capability description.

## Test plan

- [x] `pnpm --filter @openlinker/integrations-prestashop test -- prestashop-order-source.adapter.spec.ts prestashop-query.builder.spec.ts prestashop-order-feed-capability.cache.spec.ts prestashop-adapter.factory.spec.ts` — all pass, including new cases for: inline fallback on first refusal, remembering the refusal across adapter instances sharing the cache, propagating unrelated errors unchanged, propagating a non-JSON error body unchanged, and the narrowed guarantee (an update to an already-read id is not re-observed in fallback mode).
- [x] `pnpm type-check` (full monorepo) — clean.
- [x] `pnpm lint` (full monorepo, incl. `check:invariants`) — clean.
- [x] A shop that DOES support `date_upd` sorting is unaffected — the primary `fetchByDateUpdKeyset` path and its existing tests are untouched, only extracted into a private method.

Closes #2877


norbert-kulus-blockydevs:
# Code Review: PrestaShop order-feed `date_upd` fallback (#2877)

**Review Date**: 2026-09-04
**Reviewer**: AI Assistant (tech-lead persona)
**Scope**: 10 files in `libs/integrations/prestashop/`, `docs/architecture-overview.md`

## 1️⃣ Summary

**Approach**: ✅ Correct. Splits `listOrderFeed` into two extracted private methods (`fetchByDateUpdKeyset` unchanged logic, new `fetchByIdOnly`), detects PrestaShop's code-38 refusal narrowly (scoped to this specific request shape, JSON-parsed defensively), and remembers the answer per-connection on a process-singleton cache held by the factory — mirroring the existing `PrestashopOrderCurrencyResolver` / `PrestashopShopCurrencyResolver` pattern exactly.

**Safety**: ✅ Cursor wire format is untouched (still `date_upd|id`), the fallback is retried inline so the very first affected poll also succeeds, and a non-code-38 error (or a non-JSON body) propagates unchanged rather than being silently swallowed.

**Documentation Conflicts**: ✅ None. Adapter header comment updated, `docs/architecture-overview.md` updated alongside the Orders capability description.

**Files Reviewed**:
- `libs/integrations/prestashop/src/infrastructure/adapters/prestashop-order-source.adapter.ts` (+199/-23)
- `libs/integrations/prestashop/src/infrastructure/provisioners/prestashop-order-feed-capability.cache.ts` (new, 44 lines)
- `libs/integrations/prestashop/src/application/prestashop-adapter.factory.ts` (+10)
- `libs/integrations/prestashop/src/infrastructure/http/prestashop-query.builder.ts` (+15)
- `libs/integrations/prestashop/src/infrastructure/http/prestashop-webservice.client.interface.ts` (+2)
- Corresponding spec files (+252 lines of tests across 4 files)
- `docs/architecture-overview.md` (+1 bullet)

## 2️⃣ Alignment With Documentation

### ✅ Aligned with Documentation
- Infrastructure-layer-only change (`libs/integrations/prestashop`), no CORE ↔ Integration boundary crossed — matches the issue's own scope constraint.
- `PrestashopOrderFeedCapabilityCache` follows the existing "process-singleton cache on the factory" pattern (engineering-standards.md doesn't mandate a `*.service.interface.ts` here — it's not a `*.service.ts`/DI-injected service, same footing as `PrestashopOrderCurrencyResolver`, which also has no interface).
- Types unchanged inline — `idAfter?: number` added to the two existing `PrestashopQueryFilters` interfaces, matching the `updatedAfter` precedent exactly.
- File headers present and updated where behavior changed.
- New file has a proper header per `engineering-standards.md § File Headers`.
- Unit tests colocated (`*.spec.ts`), Arrange-Act-Assert followed, edge cases covered (non-JSON body, unrelated error code, cache persistence across instances, narrowed-guarantee assertion).
- Cursor safety (`docs/code-review-guide.md § Security & Safety`): cursor only advances from data actually read; wire format never changes across the fallback, as explicitly required by the adapter's own pre-existing header comment ("a second format on any path — including a fallback — would silently switch the monotonicity guard off").

### ❌ Deviations from Documentation
None found.

## 3️⃣ Issues & Risks

### 🔴 Blocking Issues
None.

### 🟡 Strong Recommendations
None.

### 🟢 Optional Improvements

1. **Retry-ladder cost on first discovery is real but bounded and documented**
   - **File**: `prestashop-order-source.adapter.ts` (adapter header comment + `fetchByDateUpdKeyset` docblock)
   - **Observation**: The webservice client's generic 5xx retry ladder (`requestWithRetry`, `maxRetries: 3`) fires before the code-38 refusal is ever inspected, so the *first* poll against an affected shop costs 4 requests and ~7s of wall-clock backoff (measured directly in the new factory test) before falling back. This is deliberate and scoped — the PR correctly declines to touch the generic retry logic to keep blast radius inside the adapter, as the issue's own "File(s)" section suggests — and it happens at most once per connection per process lifetime.
   - **Suggestion**: Not required for merge. If this becomes a recurring pain point (e.g., many affected connections restarting the worker frequently), a follow-up could have the webservice client short-circuit retries for a deterministic 4xx/5xx-with-known-error-code — out of scope here.

2. **Two independently-declared `PrestashopQueryFilters` interfaces must be kept manually in sync**
   - **Files**: `prestashop-query.builder.ts` / `prestashop-webservice.client.interface.ts`
   - **Observation**: This PR correctly follows the pre-existing duplication (the `updatedAfter` field is already declared twice), adding `idAfter` to both. Not a regression introduced by this PR, but worth flagging as existing tech debt — a shared type would prevent future drift. Out of scope for this fix.

3. **Factory-level regression test uses real timers (~7s)**
   - **File**: `prestashop-adapter.factory.spec.ts`
   - **Observation**: The new `#2877` factory test takes ~7s due to the real retry-ladder delays, consistent with three pre-existing tests in the same file that already run 5-7s each (no fake-timer convention established there). Not a deviation from house style.

## 4️⃣ Concrete Suggestions

None required — the optional improvements above are follow-up ideas, not merge blockers.

## 5️⃣ Documentation Gaps

None. Adapter header comment and `docs/architecture-overview.md` both updated in the same PR, consistent with the issue's acceptance criteria.

## 6️⃣ Test Coverage & Quality

**Unit Tests**: ✅
- `prestashop-order-source.adapter.spec.ts`: inline fallback on first refusal (asserts exact fallback request shape), skip-straight-to-fallback once known, cache shared across adapter instances, unrelated `PrestashopApiException` propagated unchanged, non-JSON error body propagated unchanged, no-prior-cursor fallback, and — importantly — a test that pins the *narrowed* #2605 guarantee (an update to an already-read id is not re-observed in fallback mode), which is exactly the behavior change this PR is documenting and accepting.
- `prestashop-query.builder.spec.ts`: `idAfter` filter shape, no spurious `date=1`, omitted when absent.
- New `prestashop-order-feed-capability.cache.spec.ts`: default state, marking, per-connection scoping.
- `prestashop-adapter.factory.spec.ts`: end-to-end wiring proof via a real counting-fetch mock, confirming the cache is a true process-singleton across two `createAdapters()` calls (mirrors the existing `#2592` currency-cache precedent).

**Integration Tests**: N/A — correctly scoped as adapter-level unit tests; no new vertical slice or HTTP surface introduced that would need `*.int-spec.ts` per `testing-guide.md`.

**Test Quality**:
- ✅ Tests mock the port interface (`IPrestashopWebserviceClient`) / use real HTTP-shaped fetch mocks, never a concrete adapter.
- ✅ Test expectations match implementation (verified passing locally).
- ✅ Error cases covered (non-JSON body, unrelated code, propagation).

## 7️⃣ Integration & Sync Safety

**Concurrency**: ✅ — `Set.add` is idempotent; no shared mutable state races within a single-threaded Node process.

**Idempotency**: ✅ — the fallback path is itself a read; no write operations introduced. The retry-then-fallback sequence within one `listOrderFeed` call is safe to repeat.

**Error Handling**: ✅ — narrow, defensive (`try/catch` around `JSON.parse`), unrelated errors and malformed bodies propagate rather than being misclassified.

**Cursor Safety**: ✅ — explicitly the central design constraint here, and correctly honored: wire format identical, cursor never moves backward, `id`-only progression is monotonic by construction (auto-increment primary key), and the narrowing of the reconciliation guarantee is documented in three places (adapter header, `fetchByIdOnly` docblock, `docs/architecture-overview.md`) plus asserted by a dedicated test.

**Transaction Safety**: N/A — no multi-step write operation.

---

## Final Verdict

**Status**: ✅ **APPROVED**

**Required Fixes** (must fix before merge): None.

**Recommended Fixes** (should fix but not blocking): None.

**Optional Improvements** (nice to have, tracked as follow-up ideas, not filed as issues):
1. Consider a targeted retry short-circuit for deterministic webservice errors (out of scope for this PR).
2. Consider unifying the two `PrestashopQueryFilters` interface declarations (pre-existing debt, out of scope).

**Estimated Fix Time**: 0h — no blocking or required work.

piotrswierzy (changes requested):
## Tech Lead review — #2884

### Summary

Narrow, well-argued adapter-local fix: a shop that refuses `date_upd` as an `orders` sort/filter field (code 38) currently dead-letters every `marketplace.orders.poll`, which is fatal for a webhook-less connection. The adapter now detects that one refusal, remembers it per connection, and falls back to an id-ordered read, retrying inline so even the first affected poll succeeds. Layering is right (entirely inside `libs/integrations/prestashop`, no CORE change, no new capability, cursor wire format unchanged), the detection is deliberately narrow (`PrestashopApiException` + JSON body + `code === 38`, with unrelated and non-JSON bodies propagating untouched), and the guarantee narrowing is stated in three places that agree — the adapter header, `fetchByIdOnly`'s docblock, and `docs/architecture-overview.md` — plus a `warn` at the moment of transition that tells an operator exactly what they lost and that webhooks are unaffected. Tests are the right shape: they pin the inline fallback, the cross-instance memory via a shared cache, both propagation cases, and — the one that matters most — the *narrowing itself* (an update to an already-read id is not re-observed). The factory-level spec pinning the process-singleton across two builds is the correct place to prove the cache actually hits.

One item to fix before merge.

---

### 🟡 Must fix

**[IMPORTANT]** — `libs/integrations/prestashop/src/application/prestashop-adapter.factory.ts:312` (`dropCachesOnShopIdentityChange`)

The new `orderFeedCapabilityCache` is the only per-connection shop-level cache on this factory that is **not** dropped when a connection is repointed at a different shop. `attributeResolver`, `featureResolver`, `categoryPathResolver`, `shopCurrencyResolver`, `orderCurrencyResolver`, `productTaxRateResolver` and `packResolver` are all cleared there; this one is keyed the same way (`connectionId`) and holds a fact about the same thing the identity change invalidates (the shop's own filterable-field list).

Consequence: an operator who repoints a connection from an affected shop to one that *does* accept `date_upd` leaves that connection permanently in the narrowed id-only mode for the life of the worker process — silently. There is no second log line on that path (the `warn` fires only on the refusal transition), so the reconciliation guarantee stays narrowed with nothing to read it from. That is precisely the failure mode `dropCachesOnShopIdentityChange` exists to prevent, and the class docblock's "the answer never changes for that shop" is true per *shop* but not per *connection*, which is what the cache is keyed by.

Fix is two lines: add a `clearCache(connectionId: string): void { this.dateUpdSortUnsupported.delete(connectionId); }` to `PrestashopOrderFeedCapabilityCache` (matching the sibling resolvers' method name) and call `this.orderFeedCapabilityCache.clearCache(connectionId)` alongside the others. Worth a line in the existing repoint spec too, so the next cache added here can't quietly skip the contract.

---

### 🟢 Optional

**[SUGGESTION]** — `prestashop-order-source.adapter.ts`, `fetchByIdOnly`

If every row in a page is dropped by `toFeedItem` (unusable `id`/`date_upd`), `cursor` stays at `fromCursor` and the identical `id > N` page is re-read on every subsequent poll — a permanent stall rather than a lost row. `fetchByDateUpdKeyset` has the same shape but softens it with the growing window and an explicit `no progress after N rows` warn; the id path has neither. The condition needs a whole page of malformed rows so it is unlikely, and the `toFeedItem` docblock's "the read position advances past it with the rest of this page" already assumes at least one good row — but since `id` is available even when `date_upd` is not, this path could advance `lastOrderId` off the raw row id and skip the poison page instead. At minimum, mirroring the keyset path's no-progress `warn` would make it observable.

**[SUGGESTION]** — `prestashop-order-feed-capability.cache.ts`

`.cache.ts` is a new filename suffix for this package (siblings are `.resolver.ts` / `.catalog.ts`; `docs/engineering-standards.md § Files and Folders` lists neither). The class name is unambiguous, so this is fine as-is — just flagging that a third suffix in one folder is worth a deliberate choice rather than drift.

---

### Notes

- The 4 `date_upd` requests asserted in the factory spec are the client's own 5xx retry ladder burning on the doomed request — the inline comment correctly names that as an accepted once-per-process cost. Good that it is asserted rather than left implicit.
- `docs/architecture-overview.md` addition sits with the Orders capability and states the narrowing plus why it is acceptable (poll is the #904 backstop, webhook is primary). Correct placement and correct framing.
- No migration, no CORE/Integration boundary crossing, no new port or capability value — right size for the problem.

### Merge readiness

🔄 **Approve with changes** — merge once the shop-identity-change invalidation is wired up.

norbert-kulus-blockydevs:
Thanks for the review. Addressed in de0090692:

**Must fix - cache not cleared on shop identity change (`prestashop-adapter.factory.ts:312`)**

Added `PrestashopOrderFeedCapabilityCache.clearCache(connectionId)` (same method name as the sibling resolvers) and wired it into `dropCachesOnShopIdentityChange` alongside `attributeResolver`, `featureResolver`, `categoryPathResolver`, `shopCurrencyResolver`, `orderCurrencyResolver`, `productTaxRateResolver`, and `packResolver`. Updated the class docblock to describe the invalidation trigger instead of claiming the answer only ever resets on a process restart. Added a factory-level regression test mirroring the existing `#2592` repoint test's shape (`should re-attempt date_upd sort when the connection is repointed at another shop (#2877)`) plus three unit tests on the cache itself (clear, clear-does-not-leak-across-connections, clear-is-a-no-op-when-never-marked).

**Suggestion - `fetchByIdOnly` poison-page stall**

You're right that a page whose rows are all dropped by `toFeedItem` (unusable `date_upd`) previously left `lastOrderId` untouched, so the identical `id > N` page would be re-read forever. `id` is available on the raw row even when `toFeedItem` can't project it, so the loop now advances `cursor.lastOrderId` off the raw row id for a dropped row too (only once a real `updatedAt` already exists to carry forward - the pathological "very first page ever, and it opens on a malformed row" double-defect is left as-is, since there's nothing honest to put in the `updatedAt` slot at that point). Also added a `warn` mirroring the keyset path's own no-progress signal for a full page returning nothing usable. Covered by two new tests: one asserting the cursor moves past a lone malformed row, one asserting a second poll doesn't re-ask for the same page.

**Suggestion - `.cache.ts` filename suffix**

Left as-is per your own note that it's "fine as-is - just flagging" - not treating it as an ask.

All 890 unit tests in the package pass, plus `lint` and `type-check` are clean.

piotrswierzy (approved):
## Tech Lead re-review — #2884 (head `de00906`)

Both items from my previous review are addressed, and the fixes are better than what I asked for.

### Prior findings — item by item

**🟡 IMPORTANT: `orderFeedCapabilityCache` not dropped on shop repoint — ✅ FIXED**

`PrestashopOrderFeedCapabilityCache.clearCache(connectionId)` added (`delete` on the set, matching the sibling resolvers' method name exactly), and called from `dropCachesOnShopIdentityChange` alongside `packResolver.clearCache` with an inline comment stating the *why* rather than the *what* ("a fact about the shop, not the connection row"). The class docblock's "one-way for the life of this process" was corrected to "one-way for the life of a given shop identity" and points at `clearCache` — so the two no longer disagree.

Coverage is exactly where it needs to be, at both levels: three unit cases on the cache (clears, doesn't touch siblings, no-op on an unmarked id), **and** the factory-level repoint spec that actually proves the contract — same connection id, refusing shop then accepting shop, asserting the second read issues `date_upd` again (1 request, no retry ladder) and **zero** `[id_ASC]` requests. That last assertion is the one that would catch a regression: a leaked refusal shows up as an id-only read, not as a missing log line. The comment above it correctly ties the reasoning back to the #2592 currency/configuration repoint test, so the next cache added here has a precedent to copy.

**🟢 SUGGESTION: `fetchByIdOnly` re-reads a page of malformed rows forever — ✅ FIXED, and more thoroughly than suggested**

I asked for at minimum a no-progress `warn`; you did both halves — the cursor now genuinely advances past a malformed row *and* the warn exists as the observability backstop for the case where it can't.

The advance is correctly conditioned. `Number.parseInt(String(row.id), 10)` + `Number.isFinite` refuses a junk id rather than writing `NaN` into the cursor, and `rawOrderId > cursor.lastOrderId` keeps the watermark monotonic — a shop that ignores `sort` still cannot drag it backwards, which is the same discipline `fetchByDateUpdKeyset` applies via `isAheadOf`. The `cursor !== null` guard is the right call and the comment says why: on a connection whose very first-ever page opens on a malformed row there is no honest `updatedAt` to pair the id with, and inventing one would put a fabricated timestamp into a cursor that the keyset path would later resume from if the shop starts accepting `date_upd` again. Leaving that double-defect to resolve itself on the first good row is the conservative direction.

The warn's condition is right too, and I checked the case that would have made it a lie: `(cursor === null || cursor === fromCursor)` is an identity comparison against the reference `cursor` is initialised from, and every write in the loop allocates a fresh object — so a page where the *raw-id* advance fired (progress made, nothing emitted) does **not** warn, and the message's "the read position is unchanged" is only ever printed when that is actually true. `rows.length >= pageSize` mirrors the keyset path's own full-window threshold rather than inventing a second convention.

Two tests, and they isolate the right things: a single-row page whose only row is unusable (asserting `nextCursor` moved to `…|6` with zero items — the poison page in its purest form), and a two-poll sequence asserting the second request is `{ idAfter: 7, sort: ['id_ASC'] }` and never `id > 5` again. Asserting the *outgoing request args* rather than just the returned cursor is what makes that second test load-bearing.

### Re-verified, unchanged

- **Detection is still narrow.** `isDateUpdSortUnsupportedError` requires `error instanceof PrestashopApiException`, a non-empty `responseBody`, a body that parses as JSON (a non-JSON body returns `false` from the `catch`, not a throw), and `errors[].code === 38`. The constant's docblock still carries the load-bearing argument for why code 38 on *this specific request* can only be about `date_upd` — `id` is always accepted per PrestaShop's own webservice implementation, and `date_upd` is the only other field this request sorts or filters by. Unrelated errors propagate untouched, which is what keeps a transient 500 from permanently narrowing a healthy shop.
- **The narrowing is still documented in three places that agree** — the adapter header, `fetchByIdOnly`'s docblock ("**This is where the #2605 reconciliation guarantee narrows**"), and `docs/architecture-overview.md` — plus the transition `warn`, which states what was lost *and* that the #904 webhook path is unaffected. That framing is correct: poll is the reconciliation backstop, and it narrows rather than disappears.
- **Cursor wire format unchanged across the fallback**, so a connection that crosses the boundary in either direction (fallback engaged, or a process restart after a repoint) resumes from a cursor the other path can still parse. I traced the resume-to-keyset direction: `updatedAfter` re-reads a wide window and `isAlreadyConsumed` drops what was already emitted, so the stale-`updatedAt`/high-`lastOrderId` pair the fallback emits costs re-reads, never losses. Safe direction.
- Layering: still entirely inside `libs/integrations/prestashop`. No CORE change, no new capability value, no port, no migration — right size for the problem.

### CI

All 8 checks green on `de00906` (Lint, Type Check, Test, Integration Tests, Build, Docker smoke, PHP unit). Branch is `behind` main — rebase or merge before landing.

### Merge readiness

✅ **Approve.** Nothing outstanding. The `.cache.ts` suffix note from last time stays a note — the class name is unambiguous and I'm not going to bounce the PR over a filename.

Issue #2028, opened 2026-08-11
[TASK] Frontend — Redesign Listings table: 6 columns, row anatomy, channel-pill fix, mobile cards


norbert-kulus-blockydevs:
## Problem / Context

`listings-list-page.tsx` renders a 6-column raw-ID table (External ID, Internal ID, Platform, Entity
Type, Connection, Created). #1965's mockup replaces it with a human-readable 6-column table (Listing,
Channel, Connection, Price, Quantity, Updated) with a specific row anatomy (thumbnail + product name +
variant pill + identifiers line + conditional state badge), plus a mobile card fold. `.channel-pill`
has no dot-color selector for Erli/WooCommerce (custom properties exist at `index.css:290-293`, but
only allegro/prestashop/amazon/shopify have selectors at `index.css:9241-9245`).

## Proposed Solution

1. Rebuild the `COLUMNS` array in `listings-list-page.tsx` to match the mockup's 6 columns and row
   anatomy exactly (Listing cell: thumbnail + product name link + variant pill + identifiers line +
   conditional badge; Channel: `.channel-pill`; Connection: `ConnectionCell` from FE-A; Price/Quantity:
   right-aligned mono, `OUT OF STOCK` badge when quantity is 0; Updated: `formatDateTime` on
   `lastStatusSyncedAt`).
2. Add the 2 missing `.channel-pill[data-channel='erli']::before` /
   `[data-channel='woocommerce']::before` selectors in `index.css` (reuse the already-defined
   `--channel-erli`/`--channel-woocommerce` custom properties).
3. Extend the `cardView` prop with `summary`/`collapsibleDetail` per the mockup's mobile card anatomy
   (2-column fact list: Channel, Connection, Price, Quantity).
4. If a persistent per-row action column turns out to be needed, use `DataTable`'s existing
   `stickyLeftColumns` (frozen leading columns) and the pinned bulk-action-bar pattern from #1755 —
   do not build a new mechanism.
5. Consume the fields BE-B/BE-A expose (price, availableQuantity, publicationStatus, product identity)
   directly — this issue runs after BE-B lands in the epic branch, no fixture stubbing needed.

## Classification

**Type**: Frontend
**Layer**: Interface (page + feature UI)
**File(s)**: `apps/web/src/pages/listings/listings-list-page.tsx`,
`apps/web/src/features/listings/api/listings.types.ts`, `apps/web/src/index.css`

## Dependencies

[...]

Issue #2023, opened 2026-08-11
[EPIC] Frontend+Backend — Redesign /listings page (human-readable columns + lifecycle tabs)


norbert-kulus-blockydevs:
## Problem / Context

The `/listings` page (`apps/web/src/pages/listings/listings-list-page.tsx`) renders the offer-mapping
table as a raw technical workbench: `External ID`, `Internal ID` (raw UUID), `Platform` (raw platformType
string), `Entity Type`, `Connection` (raw UUID), `Created`. Filters are three raw-string inputs (External ID,
Connection ID, Platform Type). A seller cannot scan the page to recognize which product/offer/channel a row
represents.

Issue #1965 (mockup `docs/plans/mockups/listings-redesign-1965.html`, merged via PR #1966) is the accepted
design reference for the target state: a 6-column human-readable table (Listing / Channel / Connection /
Price / Quantity / Updated), 4 disjoint lifecycle tabs (Active / Inactive / Draft / Ended) with live counts,
a multi-field search, a channel select, and a full mobile card fold.

This epic groups the backend + frontend work needed to implement that design end to end.

## Proposed Solution

**Backend:**
- Materialize channel-side price + quantity into a new `offer_commercial_snapshots` table, synced by
  extending the existing `marketplace.offer.statusSync` (#816) job — no new marketplace API calls.
- Enrich `GET /listings` with product/variant identity (name, SKU, EAN, thumbnail), derived lifecycle status
  (from `offer_status_snapshots`, not `OfferCreationRecord.status`), and multi-field search.
- Add filter-aware, per-tab lifecycle counts.

**Frontend:**
- Shared `ConnectionCell` component (scoped to Listings only in this epic).
- Redesigned 6-column table + row anatomy + `.channel-pill` fix (Erli/WooCommerce) + mobile card fold.
- Lifecycle tabs with URL state (`?tab=`) and counts.
- Toolbar rework: multi-field search + channel/connection selects.
- App-wide `.data-table__cell--right` CSS specificity fix (numeric columns render left-aligned everywhere
  today).

## Classification

**Type**: CORE + Frontend
**Layer**: Domain / Application / Infrastructure / Interface
**File(s)**: `libs/core/src/listings/**`, `apps/api/src/listings/**`, `apps/web/src/pages/listings/**`,
`apps/web/src/features/listings/**`, `apps/web/src/index.css`

## Dependencies

- Design reference: #1965 (mockup, merged via PR #1966)
[...]

Issue #1965, opened 2026-07-31
[TASK] Frontend — Redesign Listings page: human-readable columns + lifecycle tabs


martasl01:
## Problem / Context

The `/listings` page (`apps/web/src/pages/listings/listings-list-page.tsx`) renders the offer-mapping
table as a technical workbench: columns are `External ID`, `Internal ID` (raw UUID), `Platform`
(raw platformType string), `Entity Type`, `Connection` (raw UUID), `Created`
(`COLUMNS` array, `listings-list-page.tsx:24-66`). A seller cannot scan the page to recognize which
product/offer/channel a row represents without copying IDs elsewhere. Filters are three raw-string
inputs (External ID, Connection ID, Platform Type free text) — no channel select, no lifecycle
grouping, no counts.

This issue redesigns the page to be scannable: human-readable product/channel/connection info,
a combined multi-field search, a channel select, and four disjoint lifecycle tabs
(Active / Inactive / Draft / Ended) with counts, following the existing "name-first, ID-secondary"
pattern already used elsewhere in the app (`EntityLabel`, `ConnectionEntityLabel`, `CopyableId`).

**An interactive mockup of the target state is the accepted design reference:**
[`docs/plans/mockups/listings-redesign-1965.html`](https://claude.ai/code/artifact/87feea11-f6b0-4f3d-93db-9a500886360d) (standalone HTML, built on the app's own token
layer; covers all four tabs, every empty state, loading, error, mobile card view, both themes).
Build against it — where this body and the mockup disagree, the mockup wins and this body is stale.

> **This body was corrected on 2026-07-31** after the code was read while building the mockup.
> Three claims in the original version were false and materially changed the scope; see the
> correction comment below for the diff and the evidence.

## Proposed Solution

**Page shell** (`apps/web/src/pages/listings/listings-list-page.tsx`):
- H1 `Listings` + one-line description, `Publish products` CTA (existing wizard entry point —
  unchanged), two search inputs + channel select, tabs-with-counts, paginated `DataTable`.

### Filters

Replace the current three raw inputs with:

- Multi-field search — placeholder `Product name, SKU, EAN or external ID`, matching across those
  4 fields. (Not "listing name": the `Listing` column shows the **catalog product name**, see below.)
- `Connection ID` search — keep the existing raw-ID filter behavior
  (`ListingsFilters.connectionId` in `features/listings/api/listings.types.ts`), it already works
  this way today.
[...]

PR #2885, opened 2026-09-04
docs(listings): /listings variant-centric coverage — visual & interaction design package

### LINKED ISSUE #2823: [PRODUCT-DESIGN] /listings — variant-centric view: one row per variant, a coverage count and the channels it is listed on
> **Mockup:** [`docs/plans/mockups/listings-variant-centric-2823.html`](https://claude.ai/code/artifact/15850294-c02e-4451-a18d-58d42d4c2197) — the settled design, clickable, at 1440 / 768 / 360 in both themes. PR #2885.

## Initial problem hypothesis

An operator always creates a listing for a specific **product variant**, but `/listings` shows them **mappings**: one row per offer-to-variant mapping, so a product with several variants on several channels becomes a scatter of unrelated rows. Worse, the row model can only show what already exists — a variant that was never published has no row at all, so the one question the operator came to answer ("where is this variant *not* listed yet, and on which channel?") is the one the page structurally cannot answer.

**Outcome the design serves:** reduce the median time from entering `/listings` to identifying variants that are missing listings on one or more sales channels. Baseline to be measured over the first 14 days after release; the target is defined only once the baseline exists (60-day measurement window). No target is promised up front.

## Settled shape

The mockup linked at the top is the source of truth where it and this text disagree.

- `/listings` renders **one row per product variant**, all variants in a single flat table. Products are **not** expandable parent rows or groups — each variant stays an independent row, and its identity is repeated on every row rather than printed once per run.
- A collapsed row carries: expander, identity (thumbnail / product name / variant chip / `SKU · EAN`), a **coverage count**, and a **strip of one pill per connection the variant is listed on**.
- The coverage count is `listed / total` over the connections in view — `2/3`, `5/9`, `5/14`.
- The strip carries only connections the variant **is** on, in fixed connection order. Gaps are not in the strip; they are buttons in the expanded row, **below** its table.
- Expanding a row reveals the shipped six-column `/listings` table demoted one level, plus the two facts that lose their host when the identity cell moves up: the external offer id, and the lifecycle badge. **Each row in the expansion navigates to `/listings/:id`.**
- The five lifecycle tabs survive, with a **new `All` tab first and default** — a view whose subject is the gap cannot open on a tab that only shows offers that exist.
- A gap button opens the existing bulk publishing flow with that variant and that destination already in the URL.

## Acceptance criteria

Derived from the mockup. Figures quoted here are measured by the mockup from itself at render time, not estimated.

### Row model

- [ ] One row per product variant, flat, in one table. No product parent rows, no groups, no run markers.
- [ ] A product's variants appear consecutively (needs the sort key under **Prerequisite**).
- [ ] Collapsed row order: expander · identity · coverage count · coverage strip.
- [ ] The identity cell is the shipped `ListingCell` minus the external offer id — that id moves into the expansion. `SKU` and `EAN` render upright and monospaced (`index.css:18683` sets `font-style: normal` on the `<em>`).
- [ ] Row height does not change with the connection count. Measured 64 px at 3, 9 and 14 connections, because the strip is bounded by how many offers a variant HAS, not by how many connections the operator has.

### Coverage count

- [ ] Each row shows `listed / total` over the connections in view, right-aligned, `tabular-nums`.
- [ ] The count is **independent of the active lifecycle tab**: on `Invalid`, `2/3` still means "listed on two of three", never "invalid on two".
- [ ] Where some of the uncovered connections cannot accept a listing at all, the cell's hover says so (`4/14 · 1 of the remaining 10 cannot accept new listings`).
- [ ] **Open:** whether the denominator should exclude connections that structurally cannot accept a listing. See Open questions.

### Coverage strip

- [ ] One pill per connection the variant **is** listed on, any lifecycle. Gaps are not in the strip.
- [ ] Fixed connection order, identical in every row.
- [ ] A pill is labelled by the platform when that platform has one connection, and by the connection's own name when it has several (`soleOfPlatform`).
- [ ] A pill carries its lifecycle word when the offer is not Active. `Active` carries no word — the shipped rules give it no badge either (`case 'Active': break`).
- [ ] An offer refused by a channel validator renders as the error status badge, not as a channel pill.
- [ ] `Unsynced` renders as `Not synced` in tone **neutral**, not warning (`listing-row-state.ts:176-182`, #2028). Red is reserved for "someone has to change something".
- [ ] On a lifecycle tab, **every** pill that matched takes the selection outline. On `All`, nothing is outlined.
- [ ] The strip never overflows sideways; it wraps.

### The expanded row

- [ ] Reveals the shipped six-column table (Listing / Channel / Connection / Price on channel / Quantity on channel / Updated status read) plus Status. Price and quantity are labelled as what the CHANNEL reports; `Updated` is when OL last READ the channel, not when the listing changed.
- [ ] Only existing offers are listed. A variant listed nowhere renders a state card, not an empty table.
- [ ] The Connection cell renders `CopyableId`: the shortened id (`shortenId`, `entity-label.tsx:167-176`) over the full value, with the copy button revealed on row hover and visible where there is no hover (`@media (hover: none)`).
- [ ] The validator's reason rides under the channel pill, single-line, with an overflow count and the full list in the hover.
- [ ] **Each row navigates to `/listings/:id`**, through `DataTable`'s `rowHref` as `/listings` already uses it (`listings-list-page.tsx:918`).
- [ ] The gap buttons sit **below** the table, one per missing connection, in the same fixed order as the strip.
- [ ] A gap on a connection that cannot accept new listings is **named and not offered** — no button. Decided by `publishDestinationKind()` (`publish-destinations.ts:60-67`), not by the variant.
- [ ] A gap button opens `bulk-create` with `?productIds=&variantIds=&connectionId=` — no new route, no picker step.

### Tabs and filtering

- [ ] The five lifecycle tabs survive; `All` is added first and is the default.
- [ ] Tab counts count **offers**, not rows (`lifecycleCounts`, `listings.types.ts:216-227`) — "Active 6" can sit above three rows.
- [ ] A tab filters rows **any-match**: a variant survives if any one of its offers is in that state. Offer state does not roll up to the row.
- [ ] The badge-suppression rule (`activeLifecycle`, `listing-row-state.ts:104-112`) is **not** ported: here the pill IS the mark, so suppressing it would blank what the row is made of.

### Publication feedback

- [ ] A freshly published offer lands as **`Not synced`, not Active** — the mapping row is written when `createOffer` returns, but the lifecycle comes from a channel status read that has not happened yet.
- [ ] The in-flight state is a **data gap**: no mapping row exists while the job runs (`offer-creation-execution.service.ts:163-169` writes strictly after `adapter.createOffer` returns at `:146`). Either persist earlier or splice `OfferCreationRecord`s in; `use-bulk-batch-query` (poll 5 s, auto-stop) is the reusable piece.

### States and widths

- [ ] Loading skeleton reserving the two-line identity and the strip's shape.
- [ ] No connections configured, empty catalogue, and error — each with its own copy. "No listings found" for zero connections would read as a broken feature.
- [ ] Read-only: the gap keeps its shape, loses its action, and says why on hover. Never hidden.
- [ ] 360 / 768 / 1440, both themes. At 360 the table becomes cards and the priority inverts — gaps first, named and tappable — because there is no expansion to put them in.

### Design system

- [ ] The row-height entry from the mockup's handover frame is merged into `docs/frontend-ui-style-guide.md` § Density & Row Heights **before** this ships. The guide forbids introducing a row height that is not on its list, and its listings carve-out (#2023) explicitly refuses to cover a second table. This is a second table.
- [ ] No new colour, radius, shadow or spacing VALUE. The four rules the mockup invents are `.coverage-cell` (a positioning wrapper) with `--matched`, `.identity-col`, `.coverage-table-b` + `.coverage-strip-col`, and `.coverage-count-col`, plus one page-scoped cap on `.listing-cell__reason`.
- [ ] A `--channel-*` token per new channel. Six exist; everything else renders the grey fallback dot, which #2028 already had to fix once for Erli and WooCommerce.

## Affected persona (hypothesis)

- **Who:** operator of a multi-channel catalog — the PL wedge shape (PrestaShop / WooCommerce as master, Allegro + Erli as marketplaces).
- **Volume / scale:** a catalog where products routinely carry 2–6 variants, published unevenly across 2–4 channels; the pain scales with variants × channels, not with order volume.
- **Sophistication:** technical but not a developer; lives in the operator UI, scans rather than reads.

## Current workarounds

Today the coverage answer lives on **`/products`**, not on `/listings`:

1. Open `/products`, apply the `unlisted on <connection>` filter.
2. Expand a product row to see `ProductVariantRow` — per-variant coverage pills plus a `+ Create offer` CTA on each gap.
3. Repeat per product, because the answer is only visible inside an expanded drawer, one product at a time.

`/listings` itself is used for the opposite question ("what is already out there and what state is it in") — its five lifecycle tabs all describe offers that exist.

## Strategic alignment

- [ ] **Wedge-relevant** — supports the PL Allegro+PrestaShop+InPost wedge
- [ ] **Differentiator** — gives us a feature BaseLinker doesn't have or charges too much for
- [ ] **Foundational** — unblocks other features (specify which)
- [ ] **Tech-debt-driven** — reduces friction across many features
- [ ] **External requirement** — compliance, regulatory, marketplace API change

## Prerequisite — the read model this view needs does not exist yet

Called out here because it changes what can be designed, not as an implementation note.

Neither read model backing the two candidate pages can produce "one row per variant, with the gaps visible":

- **`GET /listings`** is a read model over existing offer-mapping rows. It enumerates mappings, not variants, so a never-published variant cannot appear in it under any filter. It also orders hard-coded `mapping."createdAt" DESC, mapping."id" DESC` (`offer-mapping.repository.ts:221-224`) and its query DTO carries no sort field.
- **`GET /products`** does enumerate variants with coverage, but aggregates it at the **product** level (`listedVariants` / `variantCount`). The per-variant pills inside `ProductRowDetail` are assembled client-side by calling the listings query **once per variant** — viable inside one expanded drawer, not across a paginated list.

So a variant-centric read (variants LEFT JOIN offer mappings, per connection) is a **prerequisite** for this design, not a detail of it. The shape it has to serve: **variant × connection, and the gaps are the payload** — at 20 variants and 14 connections that is 280 cells per page, of which the mapping table can supply only the filled ones. Paging by variant does not cut it, because the multiplier is on the other axis.

It also needs a **`productId, variantId` sort key** — the only thing flat rows still need from the backend. `/products` already has a `sort`/`dir` pair as precedent (`list-products-query.dto.ts:30-42`); none of its values groups by product.

**Per-connection publish eligibility has to reach the row.** `publishDestinationKind()` resolves from the connections read the FE already makes — nothing new server-side, but it decides whether a gap renders as a button or as a named, unactionable pill. Note WooCommerce declares `OfferManager` but **not** `OfferCreator` (`woocommerce-plugin.ts:80-86`), so a WooCommerce connection with `ProductPublisher` switched off can hold listings and cannot accept new ones.

## Out of scope (initial)

- Product-level expandable parent rows or groups.
- A new listing-creation form (the existing bulk publishing flow is the target of a gap button).
- Inline price or stock editing.
- Delisting / ending offers from this view.
- Bulk publication from selected table rows, and **no row selection at all in v1**. The only bulk action worth having — one variant to N channels — is blocked in the data: `BulkListingBatch.connectionId` is a **scalar** column, the destination rail is a single-select `radiogroup`, and the wizard route reads a single `connectionId`. That is a migration plus a new request shape, reaching the at-most-once advancement counters (#737).
- Column sorting. Matches the page as built: no column definition sets `sortable`.
- Rejection handling and resynchronisation flows.
- Backend implementation of the prerequisite read model.

## Open questions for refinement

- [ ] **Should the coverage count's denominator exclude connections that cannot accept a listing?** As drawn it counts every connection in view, so a variant can never read `14/14` and `13/14` looks like a gap no action can close. The cell's hover names the shortfall rather than the denominator hiding it, because a denominator that quietly excluded connections would make two rows showing the same number mean different things. A product decision, not an implementation detail.
- [ ] **Should the selection outline be suppressed when every pill in a row matched the tab?** Outlining all of them is honest but carries no information. Suppressing it would mirror the shipped `activeLifecycle` badge rule — and would also make two rows on one tab look identical for two different reasons ("all matched" and "this is the `All` tab").
- [ ] **What happens at the page boundary?** A product's variants can straddle a paginated page, which breaks "variants appear consecutively" exactly where it matters. Flat rows return paging to variants + offset, but the straddle itself is not solved by layout.
- [ ] **What does this add over `/products` + `unlisted on`?** `/products` shows coverage one expanded drawer at a time; this shows it across a paginated list with the gap actionable in place. Worth stating in the spec rather than leaving implied.

## Design constraints (binding, from `docs/frontend-ui-style-guide.md`)

- The listings identity row is a **documented carve-out** (#2023): 32px thumbnail + name/badges line + meta line, `min-width: 28rem` on the first column. This design's row is shorter (no offer id, no validator line) and therefore needs **its own entry**, not a silent reuse — see Acceptance criteria § Design system.
- Reuse `DataTable`'s existing `expandable` mode (already used by `/products`) rather than inventing a row disclosure. Note it is mutually exclusive with `virtualize`.
- **A row cannot both expand and navigate**, and this shapes the design rather than merely constraining it: "Expandable takes precedence over `rowHref` for the row-level click" (`data-table.tsx:90`) and `linkifyFirstCell` is `Boolean(href) && !expandable` (`:472`). So the outer variant row expands and the **inner** listing row — which is not expandable — links. `rowLinkDisplay="block"` is deliberately not used on that inner row: its reason (#2023) is a tall composite first cell, and there the first cell is a one-line offer id.
- Tables over card grids for operational data; every row earns its height.
- No new colour, radius, shadow or spacing value: the mockup transcribes tokens from `apps/web/src/index.css`.

> **Stage 1 calibration**: OpenLinker is pre-paying-customer. The refinement workflow produces a lightweight conviction spec — no success-metric theatre, no anti-metrics, no exhaustive risk catalogs. See [`docs/contributors/refinement-workflow.md`](../blob/main/docs/contributors/refinement-workflow.md#project-stage-calibration).

## Linked source

- **Mockup:** `docs/plans/mockups/listings-variant-centric-2823.html` — PR #2885, live at https://claude.ai/code/artifact/15850294-c02e-4451-a18d-58d42d4c2197
- Related issues: #2708 (UX research & redesign proposal for `/listings`)
- Prior art, same page: #1965 (listings redesign — human-readable columns + lifecycle tabs), mockup at `docs/plans/mockups/listings-redesign-1965.html`

## Definition of "refinement complete"

- [ ] Product spec at `docs/specs/product-spec-{N}-{slug}.md` merged
- [ ] Build/no-build decision recorded (Gate D output)
- [ ] If "build": implementation issues spawned and linked here — including the variant-centric read model named under Prerequisite
- [ ] **This Product Design issue is closed** when all of the above are done:
  - On Gate D = YES → `state_reason: completed`
  - On Gate D = NO → `state_reason: not_planned` with reasoning recorded
  - On DEFER → leave open with status note in body until defer condition changes
- [ ] Impl children track impl progress independently; **closing this PD does NOT depend on impl shipping**

martasl00:
The visual and interaction design package for `/listings` as a **variant-centric** view — one row per product variant, showing where that variant is and is not listed — so the read-model issue this design needs can be written against a decided shape instead of freezing an endpoint ahead of the drawing.

Docs, plus the capture script that renders them. Two files: `docs/plans/mockups/listings-variant-centric-2823.html` and `apps/web/e2e/listings-variant-centric-shots.mjs`. **No application code is touched** — the second file lives under `apps/web` but is a Playwright capture script, a sibling of the #1965 one, not part of the app.

**Live version:** https://claude.ai/code/artifact/15850294-c02e-4451-a18d-58d42d4c2197

## What it contains

11 frames: a clickable prototype at 1440 with a live 3 / 9 / 14 connection switch, the measurement that the design turns on, the mark vocabulary, the expanded row, the page a listing row navigates to, the lifecycle tabs, what a gap does when clicked, the loading / empty / error / read-only states, 768 px, 360 px, and a handover frame. Every colour, radius, shadow and spacing value is a token copied from `apps/web/src/index.css`; every component is transcribed from an existing primitive with the line range it came from, and the four rules this file invents live alone in one section with their reasons.

Every figure quoted anywhere in the file is measured by the page from itself at render time, so a number cannot drift from the picture above it.

## Read this part first: there is ONE design, not two

Rounds 1–3 carried two candidates side by side — a **grid** with one column per connection, and a **strip** of labelled pills. The PM chose the strip (3 September). The grid is gone from every frame, and this is recorded here because it shaped seven of the eleven frames and someone will otherwise redraw it. Two independent failures killed it:

- **Width.** Each connection cost 92 px of a container that has about 1220, so the tenth column left the page — 459 px past the edge at fourteen. And the 92 px was not a preference: it was sized on `INVALID`, the one word the grid could afford to contain, so the ceiling moved with how the operator had **named** their connections. One connection renamed from `Erli` to `Erli Marketplace PL` took a column away by itself.
- **Silence.** A dot in the fourth column means "Active on Shopify" only while the word `Shopify` is on screen. The table is ten rows tall and the header is one, so for most of the time an operator spends here the marks say nothing. Sticky headers would have fixed that and are not coming — a PM decision the system agrees with: `DataTable` has no pinned-header support at all (`data-table.tsx:118-130`).

The strip has no such ceiling, and the reason is structural rather than a lucky fixture: **it is bounded by how many offers a variant HAS, not by how many connections the operator has.** Measured at 3, 9 and 14 connections the row is 64 px every time.

Its one real cost is stated rather than hidden — the strip answers "where is this?" and answers "where is it NOT?" only by absence. Which is what the coverage count below is for.

## The row

**A coverage count, then the strip.** `2/3`, `5/9`, `5/14` in its own column between the variant and the pills — how much of the operator's channel estate this variant reaches, as a number, before the strip says where exactly. It is deliberately tab-independent: on the `Invalid` tab `2/3` still means "listed on two of three", never "invalid on two".

This reverses an earlier cut and is not the issue's own `Allegro 1/1` badge coming back. That badge counted a **single connection**, where the answer is binary and the number was "yes in costume" — hence its rejection. This counts **connections**, which is a different measurement.

**One decision it leaves open, and a dev should not resolve it quietly.** The denominator counts every connection in view, including one that structurally cannot accept a listing (`publishDestinationKind()` resolving to nothing — no `OfferCreator`, no enabled `ProductPublisher`). Such a variant can never read `14/14`, and `13/14` looks like a gap no action can close. The cell's hover names the shortfall (`4/14 · 1 of the remaining 10 cannot accept new listings`) rather than the denominator hiding it, because a denominator that quietly excluded connections would make two rows showing the same number mean different things.

## The expanded row, and why its two clickable things cannot swap jobs

The expansion is the shipped six-column `/listings` table demoted one level, plus the two facts that lose their host when the identity cell moves up (the external offer id, and the lifecycle badge). Nothing about it is invented.

- **Its rows navigate** to `/listings/:id`. This is a **regression repaired, not a feature added**: `/listings` already passes `rowHref={(m) => m.id}` (`listings-list-page.tsx:918`), and the demotion lost it. Expect to keep shipped behaviour here rather than write it. `rowLinkDisplay="block"` is deliberately **not** carried over — its reason (#2023) is a tall composite first cell, and the expansion's first cell is a one-line offer id.
- **The gap buttons sit below the table**, not above. The expansion is a statement of where the variant currently is, so opening it with a list of places it is not put a form where a status belonged.

The split between the two clickable things is **structural, not chosen**: "Expandable takes precedence over `rowHref` for the row-level click" (`data-table.tsx:90`) and `linkifyFirstCell` is `Boolean(href) && !expandable` (`:472`). A row cannot both expand and navigate — so the outer variant row expands, the inner listing row links, and neither could take the other's job.

Frame 05 draws the destination once, lightly: `ListingDetailPage` already renders `eyebrow="Listings"` over ``Mapping — {externalId}`` with a back link (`listing-detail-page.tsx:124-136`) over a `KeyValueList` of eight fields (`:139-161`). The frame transcribes that and stops — no marketplace-offer section, no `Offer creation` block, no `Context` panel, no `Edit offer` — because #2823 does not change that page. It exists because an expansion whose rows click through to nowhere is a design nobody checked.

## Two app-side notes worth their own issues

- **`.text-muted` has no global rule in `index.css`.** The only match is `.ai-provider-table .text-muted` (`:9211`), so `EmptyValue`'s em-dash renders in the inherited colour on every other surface. Not fixed here — it predates this epic.
- **One divergence is reported rather than reproduced.** In the running demo the `CopyableId` copy button appears accent-**filled** under the cursor, and no rule in `index.css` produces that: `:hover` only lifts the text to `--text-primary` and the border to `--border-default`. The mockup transcribes the stylesheet; inventing an accent fill to match a screenshot would put a rule in the system that is not in the system.

## What has to exist before this can be built

Frame 11 lists six items with their evidence. The load-bearing one: **neither `GET /listings` nor `GET /products` can answer "one row per variant with the holes visible".** The first enumerates mappings and sorts hard-coded `createdAt DESC, id DESC` with no sort parameter; the second aggregates at product level. The shape needed is variant × connection, **and the holes are the payload** — at 20 variants and 14 connections that is 280 cells per page, of which the mapping table can supply only the filled ones. Paging by variant does not cut it, because the multiplier is on the other axis.

Also required: a `productId, variantId` sort key (the only thing flat rows still need from the backend), per-connection publish eligibility reaching the row, and a `--channel-*` token per new channel — six exist, everything else renders the grey fallback dot, which #2028 already had to fix once for Erli and WooCommerce.

## The style-guide entry this row requires

`docs/frontend-ui-style-guide.md` § Density & Row Heights forbids introducing a row height that is not on its list, and its listings carve-out explicitly refuses to cover a second table. This is a second table. Frame 11 carries the entry, with its figures filled from the measurement run rather than typed in.

Closes #2823

https://claude.ai/code/session_01Hact6bXBLWaiPQZzJCNTVD

### DIFF
--- a/apps/web/e2e/listings-variant-centric-shots.mjs
+++ b/apps/web/e2e/listings-variant-centric-shots.mjs
@@ -85,7 +85,7 @@ const SHOTS = [
   { name: 'vocabulary-1440', selector: '#frame-vocab', width: 1440, height: 700 },
   { name: 'detail-1440', selector: '#frame-detail', width: 1440, height: 1000 },
   { name: 'mapping-1440', selector: '#frame-mapping', width: 1440, height: 560 },
-  { name: 'tabs-1440', selector: '#frame-tabs', width: 1440, height: 1200 },
+  { name: 'tabs-1440', selector: '#frame-tabs', width: 1440, height: 1670 },
   { name: 'publish-1440', selector: '#frame-publish', width: 1440, height: 1620 },
   { name: 'states-1440', selector: '#frame-states', width: 1440, height: 1470 },
   { name: 'handover-1440', selector: '#frame-handover', width: 1440, height: 340 },
@@ -197,6 +197,35 @@ async function measure(page, shot) {
     }
     for (const proof of proofs) {
       console.log(`  scroll proof · scrolled ${proof.scrolled} px · header ${proof.headerVisible ? 'STILL VISIBLE — the panel is not scrolled far enough to prove anything' : 'off-screen (expected)'}`);
+    }
+  }
+  /* The outline says "these are the pills that put this row on this tab", so
+     the count has to match. Round 3 took `filter(...)[0]` and outlined the
+     FIRST match only, which on the `Active` tab meant one outline on a row
+     with three live channels — a mark that asserts something false about the
+     other two. Assert per row that the outlined pills equal the pills whose
+     lifecycle matches the tab. */
+  if (shot.selector === '#frame-tabs') {
+    const figures = await page.locator('#tabs-mount > figure').evaluateAll((figs) =>
+      figs.map((fig) => {
+        const tab = fig.querySelector('.tabs__trigger[data-state="active"]')?.textContent?.trim() ?? '?';
+        const rows = [...fig.querySelectorAll('tr[data-variant]')].map((tr) => {
+          const pills = [...tr.querySelectorAll('.coverage-pills > *')];
+          /* A matched pill is wrapped in `.coverage-cell--matched`; an
+             unmatched one is the bare pill. */
+          const outlined = pills.filter((el) => el.classList.contains('coverage-cell--matched')).length;
+          return { pills: pills.length, outlined };
+        });
+        return { tab: tab.replace(/\s+/g, ' '), filtered: !tab.trim().startsWith('All'), rows };
+      }));
+    for (const fig of figures) {
+      /* `All` is not a filter — no pill "let the row in", so zero outlines is
+         the correct answer there and must not read as a failure. */
+      const none = fig.filtered ? fig.rows.filter((r) => r.outlined === 0).length : 0;
+      console.log(`  tab "${fig.tab}" · ${fig.rows.length} row(s) · outlined per row `
+        + `${fig.rows.map((r) => `${r.outlined}/${r.pills}`).join(' ')}`
+        + `${fig.filtered ? '' : ' · unfiltered, no outline expected'}`
+        + `${none > 0 ? ` · ${none} ROW(S) WITH NO OUTLINE — a filtered row must say why it is here` : ''}`);
     }
   }
   /* The expansion's rows navigate, so the thing worth asserting is that they
--- a/apps/web/e2e/listings-variant-centric-shots.mjs
+++ b/apps/web/e2e/listings-variant-centric-shots.mjs
@@ -234,6 +234,24 @@ async function measure(page, shot) {
   if (shot.selector === '#frame-proto') {
     const doc = await page.evaluate(() => ({ s: document.documentElement.scrollWidth, c: document.documentElement.clientWidth }));
     console.log(`  prototype page ${doc.s}/${doc.c} px · ${doc.s <= doc.c ? 'no sideways page scroll' : 'PAGE SCROLLS SIDEWAYS'}`);
+    /* The coverage count and the strip state the same fact twice — one as a
+       number, one as a list — so they can drift, and a drifted count is worse
+       than none: it is a figure an operator would act on. Assert per row that
+       the numerator equals the number of pills the strip drew. */
+    const counts = await page.locator('#rows tr[data-variant]').evaluateAll((rows) =>
+      rows.map((tr) => {
+        const text = tr.querySelector('.coverage-count-col span')?.textContent ?? '';
+        const strip = tr.querySelector('.coverage-pills');
+        const drawn = strip
+          ? (strip.querySelector('.coverage-pill--none') ? 0 : strip.children.length)
+          : 0;
+        return { text, listed: Number(text.split('/')[0]), drawn };
+      }));
+    const drifted = counts.filter((c) => c.listed !== c.drawn);
+    console.log(`  coverage counts · ${counts.length} row(s) checked · `
+      + (drifted.length === 0
+          ? 'every numerator matches its strip'
+          : `DRIFT: ${drifted.map((c) => `${c.text} vs ${c.drawn} pills`).join(', ')}`));
   }
 }
 
--- a/apps/web/e2e/listings-variant-centric-shots.mjs
+++ b/apps/web/e2e/listings-variant-centric-shots.mjs
@@ -1,0 +1,157 @@
+/**
+ * Variant-centric listings mockup — responsive render capture (#2823).
+ *
+ * Captures the standalone mockup at
+ * docs/plans/mockups/listings-variant-centric-2823.html at the three
+ * style-guide widths (360 × 812 / 768 × 1024 / 1440 × 900,
+ * docs/frontend-ui-style-guide.md § Responsive) in both themes, plus the six
+ * state frames at desktop width. The PNGs are attachments on the issue
+ * discussion — they are not committed.
+ *
+ * WHY THIS IS A SIBLING OF listings-mockup-shots.mjs RATHER THAN A PARAMETER
+ * OF IT: that script hard-codes the #1965 mockup path AND two selectors that
+ * this design deletes. Its readiness gate polls
+ * `.tabs__count[data-count="active"]`, and its tablet assertion requires the
+ * lifecycle tab strip to fit without scrolling — but #2823 removes the tab
+ * strip (the lifecycle tabs become an offer-state filter), so both would fail
+ * on a page that is working correctly. Making one script serve both mockups
+ * would mean branching on which mockup it was pointed at, which is worse than
+ * two files.
+ *
+ * Shares the rest of the #1965 approach deliberately: the same capture-time
+ * chrome strip (inline !important, because an appended stylesheet loses the
+ * cascade against `.sheet`'s own padding), the same font injection (the mockup
+ * declares IBM Plex but ships no @font-face; the app's self-hosted faces are
+ * extracted from index.css and re-pointed over file://, which the Polish
+ * sample data needs for latin-ext), and reducedMotion so the publishing pulse
+ * and the skeleton shimmer freeze and a light/dark pair stays comparable.
+ *
+ * Like the rest of this folder it is a capture script, not a test — but it
+ * asserts as it goes: that all three tables have rendered, that the tablet
+ * matrix genuinely overflows its own container, and that neither the tablet
+ * nor the mobile toolbar does (the style guide allows horizontal scrolling
+ * inside a table container and nowhere else).
+ *
+ * Usage:  node apps/web/e2e/listings-variant-centric-shots.mjs
+ * Env:    OUT_DIR · THEME=light|dark (default both)
+ */
+
+import { chromium } from '@playwright/test';
+import { mkdir, readFile } from 'node:fs/promises';
+import { dirname, resolve } from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+const REPO_ROOT = resolve(dirname(fileURLToPath(import.meta.url)), '../../..');
+const MOCKUP = resolve(REPO_ROOT, 'docs/plans/mockups/listings-variant-centric-2823.html');
+const APP_CSS = resolve(REPO_ROOT, 'apps/web/src/index.css');
+const FONT_DIR = resolve(REPO_ROOT, 'apps/web/public/fonts');
+const OUT_DIR = process.env.OUT_DIR ?? resolve(REPO_ROOT, 'e2e-out/listings-2823');
+const THEMES = process.env.THEME ? [process.env.THEME] : ['light', 'dark'];
+
+const SHOTS = [
+  { name: 'mobile-360', selector: '#frame-mobile', width: 360, height: 812 },
+  { name: 'tablet-768', selector: '#frame-tablet', width: 768, height: 1024 },
+  { name: 'desktop-1440', selector: '#frame-desktop', width: 1440, height: 900 },
+  { name: 'expanded-1440', selector: '#frame-expanded', width: 1440, height: 700 },
+  { name: 'filters-1440', selector: '#frame-filters', width: 1440, height: 760 },
+  { name: 'publish-1440', selector: '#frame-publish', width: 1440, height: 760 },
+  { name: 'states-1440', selector: '#frame-states', width: 1440, height: 1000 },
+  { name: 'degraded-1440', selector: '#frame-degraded', width: 1440, height: 800 },
+  { name: 'handover-1440', selector: '#frame-handover', width: 1440, height: 1400 },
+];
+
+async function stripChrome(page, sel) {
+  await page.evaluate((selector) => {
+    const set = (el, prop, value) => el.style.setProperty(prop, value, 'important');
+    document.querySelectorAll('.controls, .sheet__title, .sheet__sub, .sheet__note, .viewport-frame__label, .gap-legend, .eyebrow')
+      .forEach((el) => set(el, 'display', 'none'));
+    document.querySelectorAll('section.viewport-frame').forEach((el) => {
+      if (el.matches(selector)) { set(el, 'display', 'block'); set(el, 'margin', '0'); }
+      else { set(el, 'display', 'none'); }
+    });
+    const sheet = document.querySelector('.sheet');
+    set(sheet, 'max-width', 'none'); set(sheet, 'padding', '0');
+    document.querySelectorAll(`${selector} .viewport-frame__scroll`).forEach((el) => {
+      set(el, 'padding', '0'); set(el, 'background', 'none'); set(el, 'overflow', 'visible');
+    });
+    document.querySelectorAll(`${selector} .phone-frame, ${selector} .tablet-frame`).forEach((el) => {
+      set(el, 'max-width', 'none'); set(el, 'margin', '0');
+    });
+  }, sel);
+  await page.evaluate(() => new Promise((r) => requestAnimationFrame(() => requestAnimationFrame(r))));
+}
+
+async function fontCss() {
+  const css = await readFile(APP_CSS, 'utf8');
+  const blocks = css.match(/@font-face\s*\{[^}]*\}/g) ?? [];
+  if (!blocks.length) throw new Error('no @font-face blocks');
+  return blocks.join('\n').replace(/url\(['"]?\/fonts\/([^'")]+)['"]?\)/g, (_m, f) => `url('file://${FONT_DIR}/${f}')`);
+}
+
+/* Readiness signal, replacing #1965's `.tabs__count[data-count="active"]`:
+   the tab strip is gone in this design, so gate on the live table's own rows. */
+async function assertReady(page) {
+  await page.waitForFunction(() => document.querySelectorAll('#rows tr[data-row]').length > 0, { timeout: 5000 });
+  await page.waitForFunction(() => document.querySelectorAll('#tablet-rows tr[data-row]').length > 0, { timeout: 5000 });
+  await page.waitForFunction(() => document.querySelectorAll('#mobile-cards li').length > 0, { timeout: 5000 });
+  const fontsOk = await page.evaluate(() =>
+    document.fonts.check('13px "IBM Plex Sans"') && document.fonts.check('13px "IBM Plex Mono"'));
+  if (!fontsOk) throw new Error('IBM Plex did not load');
+}
+
+const box = (page, sel) => page.locator(sel).evaluate((el) => ({ s: el.scrollWidth, c: el.clientWidth }));
+
+async function measure(page, shot) {
+  if (shot.selector === '#frame-tablet') {
+    const t = await box(page, '#frame-tablet .data-table__container');
+    console.log(`  tablet table ${t.s}/${t.c} px · ${t.s > t.c ? 'overflows (expected)' : 'DOES NOT OVERFLOW'}`);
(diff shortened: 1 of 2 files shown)


piotrswierzy (approved):
## Review — `/listings` variant-centric design package (#2823)

Docs + a Playwright capture script; no application code. Reviewed against `docs/frontend-architecture.md` § UX Mockups, `docs/frontend-ui-style-guide.md`, and the committed baseline (`docs/plans/mockups/listings-redesign-1965.html` + `apps/web/e2e/listings-mockup-shots.mjs`).

### Package shape — conforms

- Single `docs/plans/mockups/{feature-name}.html`, exactly the documented shape and the same shape as the #1965 baseline it names as its sibling. The capture script is a genuine sibling of `listings-mockup-shots.mjs` and the header justifies why it is a second file rather than a parameter — and correctly retracts round 2's now-void reason instead of leaving it standing.
- **Tokens, not raw hex.** The only `#`-hex literal in 3,887 lines is `--bg-surface: #ffffff` inside the token block transcribed from `index.css` — byte-identical to what the #1965 baseline does. Everything else is `var(--…)`.
- **State is never carried by colour alone.** The file states this as one of its four governing rules ("every mark carries its own name"), and the strip design is precisely what buys it: the lifecycle word rides on the pill, so the three states that shared a shape under the rejected grid now say which one they are.
- **Responsive is in scope and drawn, not asserted.** 768 and 360 frames both exist, both captured, and the capture script *asserts* neither toolbar scrolls sideways and that the page itself does not. The 360 frame reshapes through `DataTable`'s own card branch rather than squeezing the row, and it documents the one place it deliberately inverts the desktop priority (gaps first) with its reason.

### Copy claims — spot-checked against the tree, all true

The recurring defect in this repo is operator-facing copy asserting behaviour the system does not have. I verified every load-bearing claim:

| Claim | Verified |
|---|---|
| `/listings` already passes `rowHref={(m) => m.id}` — so the expansion's navigation is a regression repaired, not a feature | `listings-list-page.tsx:918` ✓ |
| Expandable takes precedence over `rowHref`; `linkifyFirstCell = Boolean(href) && !expandable` | `data-table.tsx` ✓ (both comments present verbatim) |
| `ListingDetailPage` renders `eyebrow="Listings"` over `` Mapping — {externalId} `` with a back link | `listing-detail-page.tsx:124-136` ✓ |
| `.text-muted` has no global rule — only `.ai-provider-table .text-muted` | `index.css:9344`, sole match ✓ |
| Six `--channel-*` tokens exist; anything else renders the fallback dot | `--channel-{allegro,amazon,erli,prestashop,shopify,woocommerce}` ✓ |
| `GET /listings` sorts hard-coded `createdAt DESC, id DESC` with no sort parameter | `offer-mapping.repository.ts` ✓ |
| Clicking a gap needs no new route — the wizard reads `searchParams.get('connectionId')` | `bulk-create-wizard-page.tsx:139` ✓ |
| `soleOfPlatform` is the shipped labelling rule | `pages/products/listings-coverage-pills.tsx` ✓ |

Two things I want to name as *good*, because they are the opposite of the failure mode above. The file reports the `.channel-pill` dot at **2.56:1** against WCAG 1.4.11's 3:1 and explains why it stays (shipped CSS, decorative here, channel named in text beside it at 8.88:1) rather than quietly meeting the bar in a drawing the app cannot meet. And it reports the `CopyableId` hover divergence between the running demo and the stylesheet, transcribing the stylesheet — "inventing an accent fill to match a screenshot would put a rule in the system that is not in the system" is exactly right.

The style-guide § Density & Row Heights entry the row requires is drawn in frame 11 with figures filled from the measurement run rather than typed, and the carve-out's refusal to cover a second table is correctly read. It belongs in `frontend-ui-style-guide.md` when the row ships, not here.

### The one finding

**[IMPORTANT — cheap]** — `docs/plans/mockups/listings-variant-centric-2823.html` (`#frame-states`)

`docs/frontend-architecture.md` § UX Mockups requires that *"every distinct UI state the mockup demonstrates … gets a stable `data-state` (or equivalent) attribute value … so there is exactly one unambiguous way to find the state"*, because implementation and E2E tickets must cite states **by name**, never by free-text description.

The frame-level ids satisfy this (`#frame-proto`, `#frame-tabs`, … — stable, and the capture script already addresses them). Inside `#frame-states` they do not: the five edge states (loading, read-only, no channels connected, empty catalogue, error) are rendered from JS as bare `<figure>` + `<figcaption>` prose, with no machine-addressable attribute. A ticket can cite `#frame-states`; it cannot unambiguously cite *the read-only state* — which is the one an implementer is most likely to be pointed at, since it is the demo-mode affordance whose treatment `frontend-architecture.md` § Access Control governs separately.

Adding `data-state="read-only"` / `"loading"` / `"no-connections"` / `"empty"` / `"error"` on those five `<figure>` elements is a one-line change per state and makes the file citable at the grain the convention asks for. Not blocking — everything else in the package is at a standard I'd like to see copied.

### Nit (no action)

The code citations were accurate against this branch's base, but `main` has since drifted `data-table.tsx` by ~12 lines (`:90` → `:102`, `:472` → `:485`). Line-anchored citations in a durable spec age; the surrounding quoted text is what makes them recoverable, and this file quotes it, so they stay usable. Worth knowing rather than fixing.

### Verdict

✅ **Approve.** CI green across all eight checks. This is the reference example for how a design package should arrive in this repo: one design rather than two, the rejected candidate's death recorded with both independent reasons so nobody redraws it, every figure measured by the page from itself, every invented rule isolated in one section with its reason, and the six prerequisites in frame 11 stated as evidence rather than as a wish-list. The read-model issue can now be written against a decided shape.

Issue #2879, opened 2026-09-04
[SPIKE] Integration research — Shopify (Admin GraphQL API): capabilities, flows, verdict


norbert-kulus-blockydevs:
> ## ⚠️ The research below is day-0 desk research and may be wrong
>
> Everything in this issue was produced in a single session on 2026-09-04 **without making a single live API call**. It may be incomplete, out of date, misread, or simply wrong.
>
> **Verify every claim you rely on before building on it.** Each story's "desk claim" is a hypothesis to confirm or overturn — that is the work. A `?` marks something the research already knew it could not answer; the absence of a `?` is not a guarantee.
>
> Found something wrong? **Edit this issue** so the next reader inherits the correction.

Part of #2878.

## Problem / Context

Shopify is the only one of the four candidate platforms that can act as a **master** — a `ProductMaster` and `InventoryMaster` — as well as a destination. Its closest shipped analogue is WooCommerce (`libs/integrations/woocommerce`).

Three findings make it the highest-value and lowest-risk of the four, and they should be confirmed first because the rest of the sub-issue depends on them:

1. **It is the only platform of the four with a real per-line tax rate** (`TaxLine.rate` / `ratePercentage`), so it is the only one that can satisfy ADR-063 on the order side without a shop master behind it.
2. **Its FulfillmentOrder model is a near-exact match for `FulfillmentExecutor`** (ADR-054) — two orthogonal status axes, server-declared `supportedActions`, accept/reject callbacks, a cancellation the holder may refuse. `DESIGN-oms-authority-model.md` already cites it as the precedent it was derived from. No other platform in the tree offers this shape.
3. **It natively honours #2368's `idempotencyKey`** — `inventoryAdjustQuantities` / `inventorySetQuantities` carry both compare-and-swap (`changeFromQuantity`) and a mandatory `@idempotent` directive. WooCommerce fakes idempotency with a 7-day `CachePort` cache; PrestaShop refuses `adjustInventory` outright.

Against that, two onboarding costs that are **not** WooCommerce-shaped and change who the user is:

- **Admin-created custom apps are closed to new apps.** shopify.dev states plainly: *"you can no longer create these apps."* The path is **custom distribution + OAuth authorization-code grant**, and the merchant needs a Dev Dashboard / Partner org — not just store-owner access. A WooCommerce merchant self-serves a key pair; a Shopify merchant cannot.
[...]

Issue #2878, opened 2026-09-04
[EPIC] Integrations — research & specify four new connections: Amazon, Shopify, eBay, TikTok Shop


norbert-kulus-blockydevs:
> ## ⚠️ Read this before trusting anything below
>
> **Every finding in this epic is day-0 desk research, produced in a single session on 2026-09-04. Zero live API calls were made against any of the four platforms.**
>
> It may be **incomplete, out of date, misread, or simply wrong**. Some of it is second-hand: eBay's developer docs were read through `web.archive.org` because `developer.ebay.com` returns HTTP 403 to every automated fetch, and TikTok Shop's docs are login-gated and JS-rendered, so the evidence there is a community SDK's source code rather than the vendor's own reference.
>
> **The researcher on each sub-issue must independently confirm every claim they rely on.** A row in the tables below is a *starting hypothesis to verify or overturn*, never a fact to build on. Where the research already knew it was guessing it says `UNCONFIRMED` or `?` — but absence of that marker is not a guarantee of correctness.
>
> If you find something here is wrong, **edit this issue** so the next reader doesn't repeat the mistake.

## Problem / Context

Four platforms are candidates for new OpenLinker connections:

| Platform | Markets |
|---|---|
| Amazon | FR, DE, PL |
| eBay | GB, DE, FR, PL |
| Shopify | per-store |
| TikTok Shop | PL, FR, DE, GB |

Nothing exists for any of them today. A repo-wide grep finds **zero** stubs, ADRs or planning docs — the 77 `amazon`, 227 `shopify` and 51 `ebay` hits are all illustrative prose in existing ADRs, test fixtures for the open-world `platformType` path, or FE theme tokens (`--channel-amazon`, `--channel-shopify` already exist in `apps/web/src/shared/theme/tokens.ts`). `tiktok` returns nothing at all.

Before any of them gets an adapter we need to know, per platform:

- which of the 12 `CoreCapabilityValues` and 47 sub-capabilities it can actually satisfy;
- which **existing OpenLinker features silently break** on it;
- whether we can even obtain a test environment;
- and whether the answer is *adopt*, *adopt with named limits*, or *don't*.

This epic produces those answers as documents. **No adapter code ships from it.**

## Proposed Solution

Four sibling research sub-issues, one per platform, each delivering a spike doc and a product spec. Sequenced by cost-to-build and by what each unlocks for the others:

| Order | Platform | Why here | Unlocks |
|---|---|---|---|
[...]

Issue #2594, opened 2026-08-27
[TASK] CORE/Worker — sweep throughput is bounded by OL_LANE_REALTIME_SCOPE_CAP, not by the shop


norbert-kulus-blockydevs:
## Problem / Context

Acceptance criterion 5 of #2489 — *"synchronisation does not significantly impact store performance"* — is now measured **at elevated tempo**, which is what the earlier 0.989 figure could not answer: it was taken at the very tempo the epic exists to abandon.

Repeating the interleaved A/B run surfaced something more important than the ratio.

**Raising the connection rate limit to 300/min achieved almost nothing: 50 -> 63 req/min.** The ceiling is not the connection limiter and not PrestaShop. It is **`OL_LANE_REALTIME_SCOPE_CAP`, default 2** (`apps/worker/src/sync/sync-job.runner.ts:126-142`). The per-product child job is registered in the `realtime` lane, so raising the `fan-out` and `bulk` caps — the obvious move — changes nothing at all.

With the correct cap raised:

| Run | req/min | p95 idle | p95 under load | ratio |
|---|---|---|---|---|
| Baseline, default lanes | ~50 | 0.0386 s | 0.0382 s | 0.989 |
| Adapter fix, default lanes | ~50 | 0.0378 s | 0.0380 s | 1.005 |
| **Adapter fix, raised lanes** | **~277** | 0.0405 s | 0.0403 s | **0.995** |

**At 5.5x the tempo the shop did not move.** Applied to the catalogue: 39 700 requests at 277/min is **~2.4 h instead of 26.5 h**.

## Proposed Solution

This is a policy decision, not a value to hardcode. `OL_LANE_REALTIME_SCOPE_CAP` bounds **every** realtime job for **every** connection, and ADR-050 chose it by cost-of-starvation, not by throughput. The measurement proves the shop tolerates far more; it is not a recommendation to set 12 globally.

Decide how sweep children get throughput without letting a catalogue import delay a buyer's order: a per-lane or per-job-type cap, a dedicated lane for sweep children, or a per-scope override — evaluated against ADR-050's reasoning rather than against this one number.

## Classification

**Type**: CORE / Worker
**Layer**: Application
[...]

PR #2888, opened 2026-09-04
docs(shopify): live-verified spike + product-spec for #2879

jakubretajczykBD:
https://claude.ai/code/artifact/fa665fb8-ee2b-4fe3-9a12-fef582343a67
## Summary
- Live-verified 74/88 stories from #2879's checklist against a real Shopify development store sandbox (GraphQL Admin API, custom-distribution app) — see `docs/plans/analysis/SPIKE-2879-shopify-admin-api.md` §Coverage tally for the full per-group arithmetic. **Updated after tech review**: earlier estimates (~90, then 111, then 110) all derived denominators from each group's highest story-id number rather than its real item count, which overstated five of the ten groups (T, P, S, F, D all have gaps in their story numbering); the corrected, verified denominator is 88.
- All three headline findings from the issue confirmed live (per-line tax rate, FulfillmentOrder/`FulfillmentExecutor` fit, native idempotencyKey).
- Seven load-bearing corrections found to the issue's own claims — see `docs/plans/analysis/SPIKE-2879-shopify-admin-api.md` §Recommendation for the full list (missing OAuth scopes, `@idempotent` semantics, open-world return-reason catalog, `descriptionHtml` XSS exposure, the real 5-orders/minute dev-store cap and its non-standard error signal, `returnProcess` vs raw `refundCreate` for return-driven refunds, and F4/late-waybill-relay being an untested gap rather than the closed box an earlier revision claimed).
- X6 (multi-tenant quota) deliberately skipped — requires creating a second app via the Dev Dashboard UI, not reachable from this app's own credentials.
- Screenshots below are live evidence from the same sandbox, corroborating the SPIKE doc's transcripts (order/product GIDs, test-artifact names and content match byte-for-byte).

Docs only — no code, no schema, no behaviour change.

## Test plan
- [ ] Every "Live transcript below" row carries a matching fenced code block (16 rows are honestly marked "Live — observed, no transcript retained" instead and are not covered by this claim)
- [x] Reviewer confirms the coverage tally (74/88) against issue #2879's own checklist before closing it out — done in tech review; the original 76/110 figure was wrong on the denominator for 5 of 10 groups (T/P/S/F/D) and on two numerator citations (F4, R4, both evidence-row ids mistaken for story ids), all corrected in this revision.

Refs #2879
01-orders-list
02-order-1001-cancelled-taxline
03-order-1002-fulfillment-tracking
06-locations
08-products-list
09-product-minimal-snowboard-category
11-taxonomy-search-buty
12-products-export-modal


piotrswierzy (changes requested):

Substantial, genuinely live-verified work against a real Shopify dev store, and its best material — the idempotency-semantics correction, the return/refund sequence re-run after self-contamination, the four scope gaps, the `<script>`-round-trips-verbatim finding — materially changes what a downstream plan should assume. **No secret is exposed**: the token is `shpua_REDACTED` at all four occurrences; no client secret, no customer PII.

Nothing here is wrong on the merits. The corrections needed are to what the PR **claims about itself**.

### BLOCKING — "live-verified" is not substantiated for roughly half the rows claiming it

**27 evidence rows** say *"Live transcript below"*. I counted the diff: the file carries **20 fenced blocks in total**, and two further rows (`E-D5`, `E-P8`) cite *"Live transcript **in log**"* — a log **not in this PR at all**. At least 14 rows promise a transcript that isn't there: `E-M1`, `E-T1`(T2), `E-T2`, `E-T5`, `E-R9`–`E-R12`, `E-F8`, `E-T6`, `E-T7`, `E-S6`, `E-F10`, `E-D4`. **The entire Session 2 pass (2026-09-08, 14 rows) ships with zero transcripts** — including its flagship `E-F10`, the negotiation-axis closure the Verdict leans on.

The PR's own test plan checks **`[x] Every live-verified claim carries a real request/response transcript in the SPIKE doc`**. That box is false as submitted. This is exactly the dimension `architecture-overview.md` records the #2594 withdrawal for: a claim presented as measured whose measurement cannot be inspected. Paste the transcripts, or downgrade those rows to what they are (an unrecorded live observation) and uncheck the box.

### BLOCKING — `Closes #2879` against two documents that say they can't close it

Spike (3–10): *"⚠️ Work in progress — day-0/day-1 mixed research … it is not yet complete enough to close the parent issue."* Spec (3): *"Skeleton / work in progress."* With ~10 of ~90 stories unverified and a live BLOCKED item (`read_all_orders`, Open risk 1). Remove the keyword or finish the doc.

### IMPORTANT — the headline coverage number has three values in one PR

PR body *"~80/90"*; spec §11 *"~74 of the issue's ~90"*; the spike's own tally (474–484) sums to **79**. The PR's second test-plan item asks a reviewer to confirm the tally against the issue's checklist — impossible while the PR carries three answers.

### IMPORTANT — the scope-gap count contradicts itself three ways, and it's the top actionable finding

Recommendation §1: *"at least **two** confirmed gaps"*. Open risk 5: *"**THREE** confirmed gaps, not one"*. `E-F8`: *"**Fourth** confirmed scope gap"* (later resolved by `E-F10`). Spec §8 risk 4 still names **one**. The single most actionable output — "the issue's Prerequisites scope list is incomplete" — can't be acted on with four different counts.

### IMPORTANT — spec §8 risk 7 is stale against its own spike

It records *"`RefundInput` carries no native `returnId` — an adapter must stitch refund↔return association itself (SPIKE E-R8)"*. Session 2's `E-R9` explicitly **closes R7**: a clean `returnProcess` call populates `Return.refunds` with a real `Refund` id, and the Recommendation §6 says *"`RefundInput` having no `returnId` is a red herring for this use case"*. The spec records as an open risk the exact inference the spike retracted.

### IMPORTANT — twelve orphaned table rows won't render as a table

Eight rows (`E-X2`, `E-T4`, `E-P6`, `E-S5`, `E-F7`, `E-T5`, `E-C9`, `E-S4b`) sit after the `### E-P5` heading and its fenced block with no header above them; four more (`E-S8`, `E-T8`, `E-T9`, `E-D4`) trail the Session-2 table after a transcript. In rendered Markdown these are literal pipe-delimited text — and `E-X2` (the 5-orders/minute cap, one of the six headline corrections) is among them.

### IMPORTANT — a not-observed result reported as a confirmed negative

Coverage tally (504–506): *"**Confirmed as a negative result** rather than left untested: O16/C7 (429/retry behaviour) — a 50-parallel-call burst … produced zero throttling, **confirming the budget genuinely resists accidental exhaustion**."* The underlying row `E-C8` is careful and correct — it says a real 429 test *"would need either a sustained multi-minute burst well above 200 pts/s or a paid-plan-tier's tighter budget"* and that retry classification *"remains structurally UNVERIFIED"*. Not tripping a limit under admittedly insufficient load is an **unmet test condition**, not a measured property. Restate the tally in the row's own words.

### IMPORTANT — document order and a duplicate id

Actual order buries the (self-described "partial") API-surface summary between two transcript blocks; epic #2878 prescribes `Verdict → Evidence → API surface → Open risks → Recommendation`. And `E-T1` is used twice (line 79 taxonomy scope, line 105 `fullName`/`ancestorIds`) while both the spec and the Verdict cross-reference by id.

### SUGGESTIONS

- Recommendation says *"**Four** corrections are load-bearing"* then lists **six**. Spec §8's risk list is interrupted mid-list by an unnumbered "F6 upgraded finding" paragraph, restarting Markdown numbering.
- `## 11. Session coverage summary` sits **between §8 and §9**, in a template (`product-spec-978`) that has exactly ten sections.
- Coverage-tally header says *"as of end of live-testing session, 2026-09-04"* but incorporates Session-2 findings dated 09-08.
- Line 57 commits the dev-store domain and OAuth `client_id`. A `client_id` is a public identifier and no secret leaks — but the pair permanently names a live dev store the author controls, in a public repo. Consider redacting it as the token already is; the evidence loses nothing.
- **No ADR owed by this PR.** Three findings will need one at adoption and are worth naming: `E-P2`/Open risk 6 (`descriptionHtml` accepts a literal `<script>` verbatim — touches ADR-046's `applyDescriptionFormat` seam and the `sanitizeStoredHtml` boundary directly); `E-F1`/`E-F5`/`E-F10` (the first real `FulfillmentExecutor` against ADR-054, whose two-axis vocabulary matches near name-for-name — a genuine validation worth recording); `E-M6` (the `@idempotent` narrowing, load-bearing for #2368's assumptions).

### Done well

- Where transcripts **are** present they're exemplary: `E-M4/M5/M6` is a 6-call sequence with the final state independently re-read, and it **disproves** the issue's own idempotency claim rather than confirming it. `E-O1/E-O2` shows the failed first attempt alongside the corrected call.
- **`E-R8`/`E-R9` is the best work in the PR**: Session 1 reached an inconclusive result, the doc says *why* it was contaminated (*"this session's test return had already exhausted its refundable quantity via the raw path first"*), and Session 2 re-runs it clean and closes R7. A self-invalidated experiment repeated properly, recorded in place.
- Open risk 2 states what the 4-call sequence proves, what it does **not** rule out (*"the actual failure mode #2368's design targets"*), and names the follow-up.
- `E-C8` and `E-R12` both catch what most spikes miss: a populated `userErrors` array *"must never be read as 'nothing was written'"*, evidenced by a real `Refund` row created alongside errors.
- X6 is **skipped with a stated structural reason** (an app cannot enumerate its siblings), not silently dropped.

### Cross-cutting note, for the epic rather than this PR

#2888 and #2889 share one defect shape: **the product spec has drifted behind the spike it ships beside, in the same commit** (here `RefundInput`; there C13 still "blocked on `invalid_scope`"). If the spec's role is to be the durable summary of the spike, it needs regenerating from the spike as the last step before opening the PR, not writing alongside it.

---

norbert-kulus-blockydevs:
## /pr-review — docs artifact

This is a docs-only PR (a SPIKE research doc + a WIP product-spec skeleton, no code/schema/behaviour change). Reviewing it as a documentation artifact per the repo's spike/spec conventions (`docs/plans/analysis/SPIKE-*.md`, `docs/specs/product-spec-*.md`).

### What's genuinely good
- Disciplined live-vs-desk sourcing throughout: every claim is tagged `Live` (with or without a retained transcript) or `DESK`, and the doc is explicit that `DESK` = "hypothesis to confirm, not a fact to build on." That's the right posture for a spike that will feed an implementation plan.
- No secrets leaked — shop domain, client_id, and tokens are consistently placeholdered/redacted across every transcript (`{shop-domain}`, `REDACTED`, `$TOKEN`). Worth calling out since it'd be an easy slip in a doc built from real curl sessions.
- Corrections to the *parent issue's own claims* are called out explicitly rather than silently absorbed (the `@idempotent` non-replay behavior, the `returnProcess` vs `refundCreate` distinction, the open-world `ReturnReasonDefinition` catalog vs the closed `ReturnReason` enum). That's exactly the kind of finding a spike exists to surface — it would be easy to bury a correction to your own issue's headline claims, and this doc doesn't.
- Findings are consistently *actionable* for a follow-up implementation (e.g., naming the three missing OAuth scopes with exact error strings, flagging the dev-store order-rate cap as a `userErrors` string rather than an HTTP 429/cost-throttle signal that existing retry-classification logic won't catch).

### 🚨 BLOCKING: the headline coverage number doesn't reconcile with its own inputs
The doc (and the product-spec, which cites the same figure) repeatedly asserts **"80 of the ~90 stories"** and states explicitly: *"this is the single authoritative headline figure ... any other figure appearing elsewhere ... is stale and should be corrected to match."*

But the per-group tally given directly above that sentence is:

`C 10/12 · M 13/13 · T 6/12 · P 6/13 · S 6/13 · O 14/16 · F 8/8 · D 5/9 · R 8/9 · X 4/6`

Numerators sum to 80 (checks out). **Denominators sum to 111, not ~90** — a 23% gap between the stated total-checklist-size and the sum of the doc's own per-group denominators. Either:
- the "~90" figure is stale/wrong and should be ~111 (or whatever the issue's checklist actually totals), or
- one or more per-group denominators don't actually correspond to the issue's own checklist counts and are mislabeled.

Given the document's own stated purpose is to be *the* number a downstream reader trusts without re-deriving it, and given it explicitly instructs future readers to treat any conflicting number as stale, this needs to be resolved before merge — not left as a footnote. This is exactly the kind of thing that will get silently propagated into the implementation plan's scoping.

### Non-blocking
- The product-spec's decision-critical sections (§4 Solution exploration, §5 Product specification, §9 Implementation breakdown) are explicitly "Not yet started." That's honestly flagged (good), but the PR closes the parent issue #2879 on merge. If there's no follow-up issue already tracking "turn this spike into an actual implementation plan," it's worth opening one now so the unstarted sections don't just quietly stall once #2879 closes.
- Minor: the PR's own test-plan checkbox ("Reviewer confirms the coverage tally against issue #2879's own checklist before closing it out") is unchecked — given the arithmetic issue above, that checkbox catching this before merge is exactly the safety net that's supposed to work here.

piotrswierzy (commented):
## `/pr-review` (re-review) — 🔄 Approve with changes (downgraded from Request changes)

The consolidation commit did the work. **Eleven of thirteen prior findings are genuinely fixed**, and the doc has taken on a real evidence-provenance discipline it did not have: a two-tier Evidence vocabulary with the difference spelled out in prose, plus a named caveat for the two rows whose evidence lives outside the PR.

**The evidence-honesty BLOCKING is now a one-row residue rather than a structural over-claim.** Re-counted at head: **14** rows say `"Live transcript below"`, **17** say `"Live — observed, no transcript retained"`, and 10 fenced transcript sections exist. Thirteen of the 14 resolve to a real block. **`E-T2` (line 129) does not** — there is no `E-T2` transcript anywhere, which falsifies the doc's own line-17 invariant (*"Every row whose Evidence column instead says 'Live transcript below' is backed by an actual fenced code block"*). Down from ≥14 missing to 1. `E-P8` and `E-D5` now carry the honest *"referenced a local session log not included in this PR"* caveat.

**The cross-cutting spec-drift finding is RESOLVED for this PR** — the spec was demonstrably regenerated from the spike rather than written alongside it: risk 7 now matches `E-R9`, the scope count matches Open risk 5, the coverage figure matches the tally, §11 moved to the end, and it picked up P13 and the corrected C7/O16 wording.

Also fixed: the four-way scope-gap contradiction, the three coverage numbers, the twelve orphaned table rows, the duplicate `E-T1`, the document order, the "four corrections/lists six", §11's placement, the stale 09-04 date, and the committed store domain.

### The one remaining blocker

**`Closes #2879` against two documents that say in their first lines they cannot close the issue.** The spike still opens *"⚠️ Work in progress … not yet complete enough to close the parent issue"*; the spec still opens *"⚠️ Skeleton / work in progress"* with §4, §5 and §9 reading *"Not yet started"*. #2879 is OPEN. Because the base is `2878-new-integrations-story-catalogue` rather than the default branch the keyword will not auto-fire — so the live risk is a **manual close**, which `CLAUDE.md § Working Process` forbids. Use `Refs #2879`, as PR #2976's own spike does.

### IMPORTANT — the test-plan checkbox is still false

Item 1 remains `[x] Every live-verified claim carries a real request/response transcript in the SPIKE doc` against a document that now states in prose that 17 rows have no transcript. The claim moved from the doc into the PR body rather than being retired. Uncheck it, or reword to match the doc's two-tier vocabulary.

### IMPORTANT — the authoritative count still includes two stories the doc calls untested

The prose fix for O16/C7 is exemplary — a dedicated *"Attempted but inconclusive, not a confirmed negative result"* paragraph restating it in `E-C8`'s own words, mirrored in spec §11. But the authoritative **80** still counts `O16-attempted-but-inconclusive` inside `O 14/16` and `C7` inside `C 10/12`. Either exclude them, or annotate the count as including two inconclusive attempts.

### SUGGESTION

`E-T2` (line 129): add the `TaxonomyCategory.attributes` union-fragment transcript, or downgrade the cell to `"Live — observed, no transcript retained"` like its 17 siblings. And the coverage tally's `P 6/13` claims six while its parenthetical enumerates five (`P1, P3-implicit, P6, P7, P13`) — every other group matches, and the 80 sums correctly, so this is one missing label.

*Note: my prior `CHANGES_REQUESTED` still stands on GitHub and I've left it, since the `Closes` keyword is a genuine one-line blocker. Clear it and this is an approve.*

---

piotrswierzy (approved):

**Clearing my standing `CHANGES_REQUESTED`** — its sole remaining blocker is fixed.

| Prior finding | Verdict |
|---|---|
| `Closes #2879` against two WIP docs | ✅ **Resolved** |
| Test-plan checkbox false vs 17 transcript-less rows | ✅ **Resolved** |
| The authoritative 80 counted untested O16 + C7 | ✅ **Resolved** |
| `E-T2` promised a non-existent transcript | ✅ **Resolved** |
| `P 6/13` claims six, enumerates five | ❌ Unaddressed (cosmetic) |

Body line 15 is now `Refs #2879` with zero closing-keyword hits, so the WIP banners are finally *consistent* with the keyword. Both checkboxes are honestly `[ ]`.

**Recounted at head: 13 rows claim a live transcript and all 13 resolve to a real fenced block** — the invariant holds with zero residue, down from ≥14 missing two rounds ago.

And the count fix is the right shape: `80/~90` → **78/111**, pulling C7 out of C (`10/12`→`9/12`) and O16 out of O (`14/16`→`13/16`) with an explicit exclusion note. Both sums verify independently, and spec §11 was regenerated to match rather than left to drift.

### IMPORTANT — the new 78 is not reconstructible for two groups

This is the O16/C7 defect shape one group over, surviving the very commit that fixed it:

- **`X1` is counted in the numerator (565) *and* listed as "not attempted" (589)** — with two different stories wearing one id (rate cap vs sandbox fidelity).
- **`X2` is counted with no evidence anywhere.**
- **The C cell claims `9/12` but only 6 C stories are traceable** to any Story cell after excluding C7; `C11` appears zero times, `C6` only as an evidence id. The 9 looks derived as `12 − 3 not-attempted` — which measures *in scope*, not *confirmed*.

That blocks the PR's own test-plan item 2, which asks a reviewer to confirm 78/111 against the issue checklist. Not holding merge for it, but it is the same class of claim the last two rounds were about.

### SUGGESTION

The body's parenthetical says **17 rows**; the actual count at head is **16** — an over-claim by one, in the conservative direction. `P 6/13`'s numerator is defensible (`E-P6 → story P8`); only the label list is short.

Verified clean: document order matches epic #2878's prescription, all cross-references resolve, placement matches the sibling spikes exactly.

---

piotrswierzy (commented):
## Tech Lead review

Independent re-read at head (`714cc793`), which is **newer than my standing APPROVED** (`d0a3f56c`). I re-derived the findings rather than trusting the prior round. **I agree with the approval and am not withdrawing it** — the intervening commit is the rebase onto the updated epic base plus the `78/111` reconciliation, and I verified both landed intact rather than taking the commit message's word for it. Posting as a comment so the APPROVED stands.

Re-verified clean at head: **13 rows say `"Live transcript below"` and all 13 resolve to a real fenced section** (`E-M2, E-M5, E-M6, E-M12, E-P2, E-P5, E-O1, E-O3, E-F2, E-F6, E-R5, E-R6, E-R8`) — the doc's own line-17 invariant holds with zero residue. `E-T2` is correctly downgraded to *"Live — observed, no transcript retained"*. No secret anywhere: the store domain is `{shop-domain}`, tokens are `REDACTED`/`$TOKEN`, no `shpat_`/`shpua_` literal, no customer PII. CI green (Type Check, PHP Unit; the code jobs correctly skip on a docs-only path).

The `78/111` fix is the right shape and both sums verify independently: `9+13+6+6+6+13+8+5+8+4 = 78`, `12+13+12+13+13+16+8+9+9+6 = 111`, with `C7` and `O16` pulled out under an explicit exclusion note and spec §11 regenerated to match rather than left to drift. That closes @norbert-kulus-blockydevs's BLOCKING finding on the merits.

### Findings

**[IMPORTANT]** — `docs/plans/analysis/SPIKE-2879-shopify-admin-api.md:565` vs `:589`
> *(carried from my last round, unaddressed — the `O16`/`C7` defect shape one group over, surviving the very commit that fixed it.)* The **X** and **C** numerators are still not reconstructible from the document's own Story cells, which blocks the PR's own test-plan item 2 (*"Reviewer confirms the coverage tally (78/111) against issue #2879's own checklist"*):
> - **`X1` is counted and simultaneously listed as not attempted.** Line 565 counts `X1-orderCreate-cap-confirmed`; line 589 lists `X1 (sandbox fidelity — a qualitative judgement)` among *"not testable… and not attempted"*. Two different stories are wearing one id. The counted one actually traces to `E-X2`, whose Story cell is `X1/O15`.
> - **`X2` is counted with no evidence anywhere.** No row's Story cell contains `X2`; `E-X2` is an *evidence* id, not a story id. `X3` traces (to `E-C2`) and `X5` traces (to `E-X1`), so the X numerator of 4 rests on 2 supported entries, 1 ambiguous and 1 unsupported.
> - **`C 9/12` is not traceable.** Excluding `C7`, the distinct C stories appearing in any Story cell are `C1, C2, C3, C4, C8, C10` — six. `C11` appears zero times. The 9 reads as derived `12 − 3 not-attempted`, which measures *in scope*, not *confirmed* — and one of those three subtractions is `C9`, which line 587 itself says *"does not exist in the issue's own checklist"*, so it should not be in the denominator either.
>
> Not holding merge for it — the headline is directionally right and honestly bounded — but the same discipline the `C7`/`O16` exclusion note applies should reach these two cells: name which story ids the numerator is claiming, and reconcile the C denominator against the issue's actual checklist.

**[SUGGESTION]** — PR body, parenthetical
> Body says *"17 rows are honestly marked 'Live — observed, no transcript retained'"*; the count at head is **18**. Under-claim, in the conservative direction, but it is the only number in the body that does not now reconcile against the doc.

**[SUGGESTION]** — `docs/specs/product-spec-2879-shopify-shop-integration.md:10-13`
> §1 correctly frames Shopify as capable of `ProductMaster`/`InventoryMaster` *and* destination, and §8 correctly reaches `FulfillmentExecutor`/ADR-054 (which is in `CoreCapabilityValues` per #2403 — right call). But the **destination half is never named as a port**: for a shop, `docs/architecture-overview.md § Listings` puts publish on `ShopProductManagerPort` (+ `ShopCategoryBrowser` / `ShopAttributeReader`, both advertised-without-dispatch), not `OfferManagerPort`. §4/§5 are honestly *"Not yet started"* so this is not a defect — just the one line worth writing when they are, since Shopify's `TaxonomyCategory.attributes` (`E-T2`) and closed read-only taxonomy (`E-T3`) map onto that family specifically, and `E-T3` is the clean statement that Shopify has **no** `CategoryProvisioner` equivalent.

### What holds up

The two-tier Evidence vocabulary with the difference spelled out in prose is now genuinely load-bearing rather than decorative, and it is the reason this document can be audited at all — I could mechanically check the invariant, which is not true of most spikes. `E-T5`'s finding that `category` accepts **any** valid taxonomy node id with zero content-appropriateness validation, and `E-X2`'s that the dev-store order cap surfaces as a `userErrors` **string** invisible to `throttleStatus`, are both the kind of result that changes a retry classifier rather than a paragraph — and both carry the call that produced them. Corrections to the parent issue's own headline claims (the `@idempotent` non-replay semantics, `returnProcess` over raw `refundCreate`, the open-world `ReturnReasonDefinition` catalog) are surfaced rather than absorbed. `E-P2`'s `<script>` round-trip is correctly read as *"Shopify is not an XSS boundary"* rather than as a Shopify bug, which matches how `architecture-overview.md` treats the same finding for eBay under ADR-046.

### Verdict
🔄 **Approve with changes** — my APPROVED stands; the one residual is IMPORTANT but not merge-blocking.

1. Make the `X` and `C` numerators reconstructible (or state which ids they claim), and drop `C9` from the C denominator.
2. `17` → `18` in the body.

---

piotrswierzy (commented):
## Tech Lead re-review (`714cc793` → `2c8d51d9`, force-push)

**I am withdrawing my approval.** The force-push did not add a fix — it **dropped the four review-response commits** and re-pushed the original research branch rebased onto `main`. Every finding closed across my 09-09, 09-10, 09-11 and 09-13 rounds is un-closed at this head, including the coverage-tally reconciliation @norbert-kulus-blockydevs raised as BLOCKING and which my APPROVED rested on. This is not a judgement call about a rewrite; the commit graph says it outright.

```
old head 714cc793 (reviewed, approved)     new head 2c8d51d9 (now)
  714cc793 reconcile coverage-tally (111, not ~90)   ← DROPPED
  676e50fc restore dropped E-P8 provenance caveat    ← DROPPED
  bdee9f1f consolidate into one capability-organized ← DROPPED
  be932e4f address /pr-review — unreconciled counts, ← DROPPED
           unrendered rows, unsubstantiated claims
  a769db47 session 2 part 2  ────────────────────►  e07678ce session 2 part 2
  658b33d7 session 2         ────────────────────►  0e5d0ca0 session 2
  8ba45f8e returnProcess, holds, options, X6  ───►  470b5259 …
  3b676f02 add spike + product-spec           ───►  6136d482 …
```

`git merge-base --is-ancestor 714cc793 <new head>` is false, and the file diff is **+201 / −307** — a net loss of 106 lines from the spike and 13 from the spec. Evidence-row *count* is unchanged (87 → 87) and exactly one id is lost (`E-T1b`), so no research was destroyed; what was destroyed is every correction made in response to review.

Everything below is a consequence of that one cause, listed so the fix can be verified rather than assumed.

### Findings

**[BLOCKING]** — branch history
> Restore the four dropped commits on top of the current base — `git cherry-pick be932e4f5 bdee9f1f6 676e50fc0 714cc7930` (or re-push the old head rebased onto `main`). Do **not** re-do the fixes by hand; the commits exist and applying them is the smaller, auditable change. Nothing else in this review needs separate action if this one is done.

**[BLOCKING]** — `docs/plans/analysis/SPIKE-2879-shopify-admin-api.md:472` — the coverage tally is back to its pre-review state, and is now internally contradicted by the document's own evidence
> The section header is `## Coverage tally (as of end of live-testing session, **2026-09-04**)` while the document contains a `## Session 2 — live-verified follow-up (**2026-09-08**)` section 130 lines above it. The tally is stale by a whole session **by its own admission**, and three of its cells now contradict rows that sit in the same file:
> - **`F` 7/8**, while `E-F10` states *"negotiation axis confirmed fully working, closing scope gap #4 — **closes the F group to 8/8**"*. The reconciled head had `F 8/8 (… F10-negotiation-axis-closes-group)`; this head removed `F10` from the numerator while keeping the row that says it closes the group.
> - **`R` 8/9** lists `R7-partial` and `R9-corrected`, while `E-R9` states *"**R7 CONFIRMED** — `Return.refunds` genuinely populates via a clean `returnProcess` call"*. The reconciled head read `R7-CLOSED-by-R9`.
> - **`C` 10/12** and **`O` 14/16** silently re-absorb `C7` and `O16` into the numerators. The reconciled head excluded both with an explicit pointer (*"it is an unmet test condition, not a confirmed result either way"*) — which was the shape I praised as the right discipline, and which the rest of the document still describes in its "Confirmed as a negative result" paragraph.
>
> The groups now sum to **79/111** (`10+13+6+6+6+14+7+5+8+4` over `12+13+12+13+13+16+8+9+9+6`). The document states no headline anywhere. The PR body states **76/110**, twice. Three different numbers have now been in circulation for one tally, and the body's own test-plan item 2 — *"Reviewer confirms the coverage tally (76/110) against issue #2879's own checklist"* — cites a figure that appears in neither file and that the per-group arithmetic does not produce. That is the acceptance check this PR set for itself, and at this head it cannot be performed.

**[BLOCKING]** — `SPIKE-2879-shopify-admin-api.md` — the document's central auditable invariant is broken for 14 of 27 rows
> This is the property that made the document reviewable at all, and the reason I could approve it mechanically last round: *"13 rows say `Live transcript below` and all 13 resolve to a real fenced section."* At this head **27 rows carry `Live transcript below` and 14 of them have no transcript anywhere in the file**: `E-M1, E-T1, E-T2, E-T5, E-T6, E-T7, E-S6, E-R9, E-R10, E-R11, E-R12, E-F8, E-F10, E-D4`. Each of those rows does carry a rich inline result in its own cell — so the *evidence* is real; it is the **marker** that is now false, and a false provenance marker is worse than a weaker true one, because a reader who spot-checks one row and finds its transcript will trust the other thirteen.
>
> Relatedly, the two-tier vocabulary the body describes no longer exists: `"Live — observed, no transcript retained"` appears **zero** times at this head, so the body's test-plan item 1 (*"16 rows are honestly marked …"*) is also uncheckable. Two rows say `Live transcript in log`, pointing at a log that is not in this PR (the branch is 2 files).
>
> The dropped `be932e4f5` — literally titled *"unrendered rows, unsubstantiated transcript claims"* — is what fixed this.

**[IMPORTANT]** — `SPIKE-2879-shopify-admin-api.md:5,54,57,132` — the real store domain came back
> The reconciled head wrote `{shop-domain}.myshopify.com` throughout; this head names `shopfyol.myshopify.com` in the header, the Environment line, both OAuth URLs and every `curl` transcript. A Partner dev-store domain is **not** a credential and I am not calling this a secret — no token value is exposed anywhere (`shpua_REDACTED` throughout, and I swept for `shpat_`/`shpss_`/`client_secret`/bearer literals and found none). But the redaction was a deliberate act that this force-push reverted along with everything else, and a public repo is the wrong place to name a live store for no benefit. Restore the placeholder.

**[IMPORTANT]** — `SPIKE-2879-shopify-admin-api.md:230` — `## API surface summary` is nested inside `## Live transcripts`
> It sits between the `E-F1/E-F2/E-F3` transcript at 201 and the `E-C7/E-O3` transcript at 251, splitting the transcript section in two at the same heading level. Cosmetic on its own; listed because it is another artifact of the reverted consolidation and will be fixed by the same restore.

**[SUGGESTION]** — carried, and *still* open underneath all of the above
> Once the four commits are back, the one finding from my last round that was **not** fixed by them still stands: `X1` is counted in the `X` numerator (`X1-orderCreate-cap-confirmed`) and simultaneously listed among *"genuinely not testable … and not attempted"* (`X1 (sandbox fidelity — a qualitative judgement)`). Two stories wear one id. Also, `P 6/13` enumerates only five ids (`P1, P3-implicit, P6, P7, P13`). Neither is merge-blocking; both are one clause each.

### What holds up

The research itself is unharmed and remains the strongest of the three spikes in this epic — I want to be clear that this review is about the branch, not the work. `E-O1`'s per-line `taxLines[]` round-trip (`TaxLine.rate = 0.23` **and** `ratePercentage = 23.0` persisted on the `LineItem`) is the single most consequential finding across all four candidate platforms in #2878 and it is fully transcripted. `E-P5`'s discovery that `productSet` behaves as PATCH despite its name, `E-T9`'s correction that `descendantsOf` returns the full flattened subtree at every depth, `E-X2`'s dev-store order cap surfacing as a `userErrors` **string** invisible to `throttleStatus`, and the four independently-found missing OAuth scopes are all the kind of result that changes a retry classifier rather than a paragraph. `E-P2`'s `<script>` round-trip is correctly read as *"Shopify is not an XSS boundary"* rather than as a Shopify bug, matching how `architecture-overview.md` treats the same finding for eBay under ADR-046. None of that is in question; it simply needs the four commits back on top of it.

### Verdict
❌ **Request changes** *(posted as a comment — `--request-changes` is not used in this workflow)*

My 2026-09-11 **APPROVED** is explicitly withdrawn: it was given against `d0a3f56c`/`714cc793`, and the content it approved is not on this branch.

1. Cherry-pick `be932e4f5 bdee9f1f6 676e50fc0 714cc7930` back onto the rebased branch. That single action resolves findings 1–5.
2. Then reconcile the body against the restored doc: one headline coverage figure, in both places, matching the per-group sum; and re-word the two test-plan items against the vocabulary the doc actually uses.
3. Then, if you want it clean, the carried `X1` double-use and the `P` enumeration.

Also note `mergeStateStatus: BEHIND` — the branch needs an update from `main` regardless.

---

piotrswierzy (approved):
## Tech Lead re-review (`2c8d51d9` → `6b250a0a`)

The regression is fully repaired, and the repair went past the minimum I asked for. The four dropped review-response commits are back (re-applied under new SHAs — `d8d9b7f9` / `947c0efd` / `7266a5d5` / `2c8d51d9`, message-for-message the `be932e4f` / `bdee9f1f` / (E-P8 caveat) / `714cc793` set), and three further commits then corrected the tally's *methodology* rather than just its arithmetic. The previous denominators were derived from each group's highest story-id number on an assumption of consecutive numbering; `6b250a0a` replaces that with a per-group count of the checklist items that actually exist in #2879's body. I re-derived every one of those ten lists directly from the issue and they match exactly — C `1,2,3,4,5,6,7,8,10,11,12`, T `1,2,8,9,11,12`, P `1,3,6,7,8,13`, S `1,4,5,6,7,10,13`, F `1,2,3,4,6,7,8` (no `F5`), D `1,2,4,5,6,8,9`, R/M/O/X consecutive. Denominators sum to 88 and numerators to 74; each group's numerator equals the length of its own cited id list. Test-plan item 2 is now performable, and I performed it. The `Live transcript below` invariant holds for all 13 rows that claim it (E-M2/M5/M6/M12, E-P2/P5, E-O1/O3, E-F2/F6, E-R5/R6/R8 — every one has a matching `### E-*` transcript subsection); the remaining 16 rows say `no transcript retained` and name why. Redaction is clean: no `*.myshopify.com` anywhere in either file, `{shop-domain}` in all five URL positions, the access token is `shpua_REDACTED`. Research volume is unharmed at 87 evidence rows. Spike, product spec and PR description all cite 74/88 with identical arithmetic.

### Findings

**[SUGGESTION]** — `docs/specs/product-spec-2879-shopify-shop-integration.md:108`
> Reads "**This 88/74 pair** supersedes every earlier figure" — denominator-first, where every other mention in all three documents is `74/88`. Unambiguous in context because the adjacent sentence spells out both sums, so this is cosmetic only; flip it on the next touch if there is one.

Positive: the revision note at the top of `## Coverage tally` states the *defect class* (highest-id-as-denominator) and lists exactly which five groups it inflated and by how much (T 12→6, P 13→6, S 13→7, F 8→7, D 9→7, combined 22). That is what makes the number auditable by the next reader instead of merely correct today. Same for keeping O16/C7 out of the numerator as attempted-but-inconclusive rather than as a confirmed negative, and for removing F4 and R4 once they turned out to be evidence-row ids mistaken for story ids.

CI: `Type Check`, `Detect changed paths` and `PHP Unit Tests` pass; `Lint`, `Build`, `Test`, `Integration Tests` and `Docker Build Smoke Test` are path-skipped for a docs-only change (absent, not passing). `MERGEABLE` / `CLEAN`.

### Verdict
✅ **Approve**

---

piotrswierzy (approved):
## Tech Lead re-verification (head unchanged at `6b250a0`)

`MERGEABLE` / `CLEAN`, `reviewDecision: APPROVED`, base `main`, run `34832627932` settled. `Type Check` (2m22s), `PHP Unit Tests` and `Detect changed paths` pass; the five code-path jobs are **skipped** — correct for a docs-only diff (`SPIKE-2879-shopify-admin-api.md`, `product-spec-2879-shopify-shop-integration.md`), and reported as absent rather than passing.

Nothing is gating this PR.

### Findings

### Verdict
✅ **Approve** — mergeable now.

---

Issue #2894, opened 2026-09-05
[BUG] CORE — AOV / Median Order Value use a narrower order set than the Number of Orders card


norbert-kulus-blockydevs:
## Problem / Context

`docs/specs/metrics-analytics-dashboard.md` requires AOV and Median Order Value to be computed over the exact same order set as the "Number of Orders" card for the period (excluding cancelled orders) — same cohort, different formula.

Confirmed via independent audit pass: AOV / Median are computed only over orders that are BOTH FX-stamped ("converted") AND tax-rate-resolved (i.e. excluded by `netExcludedCount`'s criteria: not `taxRateEra = 'pre-rollout'`, and every line's tax rate resolved). This is narrower than the "Number of Orders" card, which correctly adds back `unconvertedCount` to cover every placed order (`analytics-kpi-strip.tsx:222`, confirmed correct). So AOV/Median silently exclude an ADDITIONAL slice of orders (the tax-rate-unresolved ones) that Number of Orders includes, on top of already excluding unconverted orders the way revenue does (which may or may not be correct for AOV specifically — average/median order VALUE is currency-denominated, so excluding non-FX-stamped orders from it has a real justification unlike Units Sold; the tax-rate-exclusion narrowing does not, since tax-rate resolution has nothing to do with whether an order's total can be averaged).

This is disclosed today via a UI tooltip, but disclosure does not satisfy the spec's "must use the exact same order set as Number of Orders" requirement — a disclosed spec violation is still a spec violation.

## Proposed Solution

[...]

Issue #2892, opened 2026-09-05
[BUG] CORE — GMV on /analytics includes shipping, contradicting the spec


norbert-kulus-blockydevs:
## Problem / Context

`docs/specs/metrics-analytics-dashboard.md` defines GMV as: "the total value of orders placed in the period (excluding cancelled orders), at gross selling prices (incl. VAT), before deducting discounts, commissions and returns. Covers the merchandise portion only – excluding shipping costs." Formula: `SUM(gross_unit_price before discount × quantity, for all line items of orders placed in the period, excluding cancelled)`.

The shipped code computes the `revenue` figure (rendered as GMV on the Revenue KPI card, `apps/web/src/features/analytics/components/analytics-kpi-strip.tsx:231`) as `SUM(order_records."reportingTotalAmount")` in `libs/core/src/orders/infrastructure/persistence/repositories/order-record.repository.ts` (~line 592-598). `reportingTotalAmount` is the FX-stamped `order.totals.total`, which is `subtotal + tax + shipping`.

Confirmed live: order `ol_order_84171f85f7fe4204ac5e4af1c27305be` has `subtotal=44.97, tax=0, shipping=10.95, total=55.92` — the code sums `55.92`, contradicting the spec's "excluding shipping costs" requirement.

Reported live by a customer: ~40% GMV overshoot vs Net Sales on a low-value catalog with flat shipping charges (Slack screenshot: GMV 2901.77 zł vs Net Sales 2067.53 zł — a gap VAT alone (23%) cannot explain). Independently re-verified by two separate review passes; both trace to the same root cause.

**Known data-model limitation:** `Order`/`OrderTotals` carries no "discount" field anywhere in the ingestion pipeline (Allegro/PrestaShop/WooCommerce). So "gross, before deducting discounts" cannot be fully reconstructed today. The achievable, spec-compliant-on-shipping fix is "gross, shipping-excluded" via `order_line_items`, which already exists and needs no migration — shipping is never itemized as a line, so summing `unitPrice × quantity` there is structurally shipping-free.

## Proposed Solution

[...]

PR #2896, opened 2026-09-05
fix(analytics): count unconverted-order units in Units Sold and Units per Order

### LINKED ISSUE #2893: [BUG] CORE — Units Sold / Units per Order undercounts by silently dropping unconverted orders
## Problem / Context

`docs/specs/metrics-analytics-dashboard.md` requires Units Sold / Units per Order to count every placed order in the period (excluding cancelled), same as the Number of Orders card. `apps/web/src/features/analytics/components/analytics-kpi-strip.tsx`'s own top-of-file doc comment explicitly claims "Units: fully real... count every placed order" — but the shipped implementation contradicts its own claim.

Confirmed via independent audit pass: the Units Sold aggregate in `libs/core/src/orders/domain/order-sales-aggregation.ts` (`getUnitsSoldByConnection` / the headline units computation) sums units **only over FX-stamped ("converted") orders**, the same restricted cohort revenue/AOV use — it does NOT add back `unconvertedUnitsSold` the way `headline.orderCount` correctly does (`totalOrders = headline.orderCount + headline.unconvertedCount`, confirmed correct in `analytics-kpi-strip.tsx:222`). The data shape *does* carry an `unconvertedUnitsSold` field, it's just never added into the rendered figure — so on any dataset with unstamped orders, Units Sold and Units per Order silently under-report versus the Number of Orders card on the same page.

This is a real, silent inconsistency across cards on the same dashboard: "Number of Orders" says N orders, but "Units Sold" / "Units per Order" behave as if a strictly smaller set of orders exists.

## Proposed Solution

Add back `unconvertedUnitsSold` to the rendered Units Sold figure and to the Units-per-order denominator/numerator, mirroring exactly how `orderCount` already adds back `unconvertedCount`. Apply the same fix to the per-channel breakdown if it has the same restriction. Units themselves are not a currency-denominated figure (a unit count needs no FX conversion), so there is no reason to exclude unconverted orders from a units count at all — the FX-stamped restriction likely leaked in from copy-pasting the revenue aggregation pattern rather than being intentional.

## Classification

**Type**: CORE
**Layer**: Domain (aggregation logic)
**File(s)**: `libs/core/src/orders/domain/order-sales-aggregation.ts`, `apps/web/src/features/analytics/components/analytics-kpi-strip.tsx` (verify no FE-side fix needed once the backend field is corrected — the FE may already read the right field name and just needs the backend to add back the unconverted count, or the FE component may itself need to sum `unitsSold + unconvertedUnitsSold` the way it does for `orderCount`)

## Dependencies

- None

## Assumptions

- A units count needs no currency conversion, so restricting it to the FX-stamped cohort was unintentional — fix is to count units across every placed, non-cancelled order regardless of FX-stamp status.

## Acceptance Criteria

- [ ] Units Sold counts every placed, non-cancelled order in the period, matching the Number of Orders card's order set exactly
- [ ] Units per Order uses the same corrected numerator/denominator
- [ ] Per-channel units breakdown (if affected) uses the same fix

norbert-kulus-blockydevs:
## Root cause

`docs/specs/metrics-analytics-dashboard.md` defines Units Sold as the sum of quantity across all line items of orders placed in the period, excluding cancelled - no currency restriction at all (a unit count needs no conversion, unlike GMV/AOV/Net Sales).

The backend already computed this correctly: `libs/core/src/orders/infrastructure/persistence/repositories/order-line-item.repository.ts`'s `getUnitsSoldByConnection` splits units into a `units` sum (FX-stamped, current-era) and an `unconverted_units` sum (everything else), and `libs/core/src/orders/domain/order-sales-aggregation.ts` rolls both into the headline and per-channel response as `unitsSold` / `unconvertedUnitsSold`. The DTO (`apps/api/src/analytics/http/dto/sales-analytics-response.dto.ts`) already serializes both fields on the wire.

The bug was entirely on the frontend: `apps/web/src/features/analytics/api/sales-analytics.types.ts` (the FE mirror of the DTO) never declared `unconvertedUnitsSold` at all, so the value silently vanished when the response was typed and destructured. `analytics-kpi-strip.tsx` then read `headline.unitsSold` directly for both the Units Sold value and the Units per Order ratio - unlike Number of Orders, which correctly computes `totalOrders = headline.orderCount + headline.unconvertedCount` per the file's own documented pattern.

## Fix (frontend only, no backend change)

- `sales-analytics.types.ts`: added `unconvertedUnitsSold` to both `SalesAnalyticsHeadline` and `ChannelSalesAnalytics`, mirroring the DTOs exactly.
- `analytics-kpi-strip.tsx`: introduced `totalUnitsSold = headline.unitsSold + headline.unconvertedUnitsSold` (unconditional - no currency gating, since units need none) and used it for:
  - the Units sold card value
  - the Units per order ratio (`unitsPerOrder(totalUnitsSold, totalOrders)`)
  - the period-over-period Units delta, for both the current and previous period
- Updated the doc comment at the top of the file to state the corrected behaviour.
- Updated existing test fixtures (`analytics-kpi-strip.test.tsx`, `channel-sales-table.test.tsx`, `analytics-page.test.tsx`, `test-utils.tsx`) to include the new field.
- Added a regression test (`analytics-kpi-strip.test.tsx`) with 40 stamped + 10 unconverted orders and 60 stamped + 15 unconverted units, asserting:
  - Total orders renders `50` (40 + 10, the existing correct behaviour)
  - Units sold renders `75` (60 + 15), never the undercounted `60`
  - Units per order renders `1.5` (75 / 50)

The corrected Units figure now uses exactly the same order population (`orderCount + unconvertedCount`) that the Number of Orders card already uses - confirmed by the regression test above, which drives both cards from one shared fixture and asserts they agree on the order count while Units correctly includes both cohorts.

Scope: this PR touches only Units Sold / Units per Order rendering and types. GMV/revenue, AOV/Median, Net Sales, and Cancellation logic are untouched (concurrent sibling PRs cover #2892 and #2894).

## Testing

- `pnpm --filter @openlinker/web exec tsc --noEmit` - clean, no analytics-related errors.
- `pnpm --filter @openlinker/web exec vitest run` on the four affected test files (`analytics-kpi-strip.test.tsx`, `channel-sales-table.test.tsx`, `analytics-page.test.tsx`, `sales-analytics-view-model.test.ts`) - all 76 tests pass, including the new regression test.
- `eslint` on all changed files - clean (2 pre-existing, unrelated warnings in `analytics-kpi-strip.tsx` about missing return types on two unrelated inline arrow functions).
- No backend files were touched, so no integration test / Docker run was needed for this change.

Closes #2893


norbert-kulus-blockydevs:
GitHub auto-detected this as merged (merge commit a4ed70c790b9b24db93ff4798fabf55363d5f2d8) once `epic/2452-analytics-currency-coverage` was updated via a direct local merge — done that way because this batch of 5 analytics PRs had cross-cutting file overlaps needing coordinated conflict resolution. This PR's diff merged with no conflicts, but as part of this merge commit I additionally fixed `sales-analytics-view-model.test.ts`: it built a `ChannelSalesAnalytics` fixture (pre-existing on the epic branch, not touched by this PR) that predated the new required `unconvertedUnitsSold` field, so it failed type-check after the merge. Added `unconvertedUnitsSold: 0` to the fixture defaults. Verified with web type-check + targeted analytics test run on the fully-integrated branch.

Issue #1987, opened 2026-08-03
[IMPL] Backend — sales & channel aggregates endpoint (revenue, orders, AOV + median, units)


piotrswierzy:
## Design source

**Product Design parent:** Part of #1976 — see spec at `docs/specs/product-spec-1976-analytics.md` (§ 4 rows A1–A4, A′1/A′5, B1–B3, B6; § 6; stories S1/S2)

## Scope

The read behind the KPI strip and the by-channel table — the two money-bearing sections of the v1 page.

- **Headline, for a date range:** revenue, order count, average order value **with median beside it**, units sold
- **By channel:** the same figures per source connection, plus each channel's share

**Why the median is in v1, not v2.** At this persona's volume (10–100 orders/day) a single large order visibly moves the mean. Showing only a mean invites *"why did my average jump?"* with the honest answer being "one big order". The median costs almost nothing once the query exists and makes the figure defensible.

## Acceptance criteria

- [ ] Given a date range, the endpoint returns revenue, order count, AOV, median order value and units
- [ ] The same figures are returned broken down per source connection, with each channel's share
- [ ] Figures are bucketed by the buyer's order time, not OL's ingestion time
- [ ] Cancelled orders are excluded, and the cancelled value is reported separately (so the exclusion is visible rather than silent)
- [ ] Multiple currencies never sum into a single figure — results are per currency, or the response states that a single figure cannot be given
- [ ] Gross-reporting and net-reporting channels are either compared on a normalised basis or explicitly reported as not comparable. **A silently mixed gross/net channel comparison is a defect, not an approximation**
- [ ] A channel whose data does not span the full requested range is identifiable from the response
- [ ] Tests cover the correctness cases specifically: a cancelled order, a mixed-currency range, a gross channel alongside a net channel, and an outlier order's effect on mean vs. median
- [ ] No new ESLint warnings or type errors introduced

## Effort estimate

- [x] **M** — 3–7 days

## Dependencies

- **Blocked by:** #1985 (order analytics read model)
- **Blocks:** the KPI strip + by-channel table frontend

## Out of scope for THIS issue

- Revenue over time / channel mix over time (A5, B4) — the v1 cut carries sparkline-level trend only; a real time-series is v2 and forces the charting-library decision
- Period-over-period comparison (A6)
- Anything needing cost, fees or refunds

jakubretajczykBD:
**Gap check against the `/analytics` mockup (PR #2018 / #2003) + current `main` state**

[...]

PR #2900, opened 2026-09-05
fix(analytics): make trend sparklines span the whole selected range

### LINKED ISSUE #2899: [BUG] CORE + Frontend — Analytics trend sparkline ignores the selected date range
## Problem / Context

On `/analytics`, the KPI-strip trend sparklines (headline revenue/orders) and the per-channel row sparklines all render from `GET /analytics/sales`'s `headline.trend` / `channel.trend` arrays. Confirmed live via Playwright + intercepted network JSON: this array is **always the 7 calendar days immediately preceding the selected `to` date**, regardless of how wide the `from`/`to` window actually is.

Consequence, confirmed live: selecting 7d, 30d, or 90d ranges (all sharing the same `to`) return **byte-identical** `trend` arrays and therefore render **visually identical** sparklines, even though every other KPI number on the page (revenue, order count, etc.) correctly changes with the range. An operator switching between 7d/30d/90d sees the headline numbers change but the trend chart never moves — which reads as broken or unresponsive, even though the underlying data pipeline is technically correct for what it was asked (a fixed trailing week).

This matches the codebase's own documented intent (`docs/architecture-overview.md`: "a 7-day daily trend") — the trend window was designed as a fixed 7-day trailing indicator, not a resample of the selected range. But nothing on the UI discloses that, so the behavior contradicts an operator's reasonable expectation that every visible element on the page reflects the range they just picked.

Only a genuinely different `to` (e.g. a custom historical range) changes the trend chart's shape — confirmed live (a custom Jan 2026 range correctly rendered a flat all-zero trend matching that period's real zero order volume).

## Proposed Solution

Pick one of two fixes and implement it (prefer option A unless investigation finds a reason it's infeasible, since it more directly satisfies "every number on the page reflects the selected range"):

**A. Make the trend window follow the selected range's granularity.** For a 7d range, keep the current per-day trailing-7-day behavior. For a 30d/90d/custom range, resample the trend into a sensible number of buckets spanning the FULL selected range (e.g. weekly buckets for 30d/90d, or a fixed bucket count regardless of span) rather than always showing only the last 7 days of a much wider window. Update `getSalesAndChannelAnalytics` (`libs/core/src/orders/domain/order-sales-aggregation.ts` / `order-record.repository.ts`) and the response DTO/frontend consumers accordingly.

**B. If (A) is out of scope or too large a change, at minimum make the sparkline's fixed-7-day-trailing-window nature visible to the operator** — e.g. a tooltip or caption on the sparkline clarifying "last 7 days" so the number above it and the chart below it are never silently describing two different periods with no visual cue that they differ.

State clearly in the PR which option was taken and why.

## Classification

**Type**: CORE + Frontend
**Layer**: Domain (aggregation), Interface (KPI strip / channel table rendering)
**File(s)**: `libs/core/src/orders/domain/order-sales-aggregation.ts`, `libs/core/src/orders/infrastructure/persistence/repositories/order-record.repository.ts`, `apps/api/src/analytics/http/dto/sales-analytics-response.dto.ts`, `apps/web/src/features/analytics/components/analytics-kpi-strip.tsx`, `apps/web/src/features/analytics/components/channel-sales-table.tsx`

## Dependencies

- None outside this epic; touches the same aggregation files as #2892/#2893/#2894/#2895 — resolve merge conflicts at integration time.

## Assumptions

- Fixing this as a resampling change (option A) is preferred over merely disclosing the fixed window (option B), because the spec's intent (a glanceable per-metric trend) is best served by the chart actually reflecting the chosen period.

## Acceptance Criteria

- [ ] Selecting a different date range visibly changes the trend sparkline shape when the underlying order data differs across that range (not just when `to` itself changes)
- [ ] The 7d range behavior is preserved/sensible (still daily granularity)
- [ ] Docs/spec updated if the trend definition changes (`docs/specs/metrics-analytics-dashboard.md`, per `.claude/rules/analytics-metrics.md`)

norbert-kulus-blockydevs:
Closes #2899

## Root cause

`GET /analytics/sales`'s `headline.trend` / per-channel `trend` array was always the 7 calendar days immediately preceding the selected `to` date, regardless of how wide the selected `from`/`to` window actually was. `buildTrend` (`libs/core/src/orders/domain/order-sales-aggregation.ts`) built a dense per-day series over the *whole* selected range (it already had all the data, fetched by `getDailyOrderAggregates` for the full range) and then simply `.slice(-7)`'d it - discarding everything before the trailing week.

Confirmed by the issue (Playwright + intercepted network JSON): 7d/30d/90d selections sharing the same `to` returned byte-identical `trend` arrays, so the KPI-strip and per-channel sparklines never visibly moved when an operator widened the range, even though every other number on the page correctly changed.

## Fix: option A (resample, not disclose)

`docs/specs/metrics-analytics-dashboard.md` (the canonical metric-definitions spec, per `.claude/rules/analytics-metrics.md`) does not mention the trend/sparkline at all - it has no formula, no granularity rule, nothing - so there is no spec constraint blocking option A. The "7-day daily trend" phrase in `docs/architecture-overview.md` was only a description of the shipped implementation, not a spec requirement, and I've updated it to describe the new behavior.

`buildTrend` now:
- keeps daily granularity, unbucketed, for a range no wider than 7 days (unchanged behavior for the common case)
- for a wider range, resamples the full dense per-day series into up to 7 contiguous, near-equal-size buckets spanning `[from, to)` end to end (via the standard `floor(i*n/k)` partition), each bucket summing the days it covers

This is purely a change to the pure aggregation function - no SQL/repository change was needed, since `getDailyOrderAggregates` already reads every day in the selected range (the trimming happened afterwards, in memory).

No DTO or frontend change was required:
- `DailyTrendPoint`'s shape is unchanged (`date`/`revenue`/`orderCount`) - a multi-day bucket's `date` is its first covered day.
- The frontend only ever reads `.revenue`/`.orderCount` off each point (`revenueTrendValues`/`orderCountTrendValues` in `sales-analytics-view-model.ts`), and `Sparkline` is array-length-agnostic.
- The KPI strip's trend label (`analytics-kpi-strip.tsx`) already derives its "trend, the last N days" wording from the full selected range width (`rangeDays(filters.from, filters.to)`), not a hardcoded 7 - so the label was already correct and simply had no matching data behind it until now.

Renamed `TREND_WINDOW_DAYS` -> `TREND_BUCKET_COUNT` to reflect that it now means "max points rendered", not "trailing days".

## Docs updated

- `docs/architecture-overview.md`: the #1987 Sales & channel aggregates bullet no longer says "a 7-day daily trend"; it now describes the range-following behavior and cites #2899.
- `DailyTrendPoint`'s doc comment in `order-sales-analytics.types.ts` updated the same way.
- `docs/specs/metrics-analytics-dashboard.md` was NOT touched - it has no trend/sparkline entry to update, and per `.claude/rules/analytics-metrics.md` a new metric definition needs a product decision before being added, which is out of scope here.

## Testing

- Updated/added unit tests in `libs/core/src/orders/domain/order-sales-aggregation.spec.ts`:
  - the old "trims a longer range to the trailing 7 days" test now asserts the 30-day range is resampled into 7 buckets spanning the *whole* range (first bucket starts at `from`, not 23 days in)
  - new test with orders spread across the full 30-day range (early-range + late-range orders) asserting both land in the resampled series and in *different*, non-zero buckets - i.e. the sparkline shape actually changes
  - new test confirming a <=7-day range still returns exactly one point per day, unbucketed
  - all pre-existing tests in the file still pass unmodified
- `pnpm --filter @openlinker/core type-check`: clean
- `npx eslint` on the three changed core files: clean
- `npx vitest run src/features/analytics` in `apps/web`: 247/247 passing, confirming no frontend regression from the (unchanged) response shape
- No Docker/integration test exists today for `GET /analytics/sales` (`apps/api/test/integration`), so none was run/added - the change is entirely covered by the pure-function unit tests above, consistent with how the rest of `order-sales-aggregation.ts` is tested.


norbert-kulus-blockydevs:
GitHub auto-detected this as merged (merge commit cbe983f17) once `epic/2452-analytics-currency-coverage` was updated via a direct local merge — done that way because this batch of 5 analytics PRs had cross-cutting file overlaps needing coordinated conflict resolution. This PR's own diff merged cleanly with no conflicts; verified with core type-check + targeted unit tests (order-sales-aggregation, order-record.repository) and api type-check on the fully-integrated branch.

Issue #2245, opened 2026-08-20
[EPIC] Per-line tax rate from the master to the invoice and the receipt


norbert-kulus-blockydevs:
> **Plain-language walkthrough:** [How the invoice knows the tax](https://claude.ai/code/artifact/c5383b02-ccd9-4ee2-acb0-27a143300740)
> Behaviour, states and copy for every surface, plus the 30 settled decisions. Read it before implementing any child issue.
>
> **Frontend spec:** [Where the tax rate shows up](https://claude.ai/code/artifact/00f22403-03e0-4804-a67d-2a950806dcea)
> Every surface that must render a rate, drawn with OpenLinker's own tokens and primitives: the seven states, the exact file and symbol per change, a state-by-surface matrix, and a decision-to-surface coverage table.

## What this epic is

**Every line of every invoice OpenLinker issues carries an empty tax rate, and each provider adapter substitutes its own guess.** inFakt hardcodes 23% (`infakt-invoicing.adapter.ts:203-204`), Subiekt hardcodes 23% (`subiekt-line.mapper.ts:26`), KSeF falls back to a per-connection `'23'` (`fa3-builder-input.mapper.ts:138`). The mapper emits `taxRate: ''` for both product lines (`order-to-issue-invoice-command.mapper.ts:228`) and shipping (`:249`), because core has no rate to give: neither `OrderItem` nor `IncomingOrderItem` has a tax field, and the only tax signals on an order are one aggregate `OrderTotals.tax` and an order-wide `taxTreatment` — neither of which can describe a mixed-rate basket.

Three consequences: a cross-border sale that should carry `0% WDT`, `0% EX`, `np I`/`np II` or an OSS rate gets 23% instead (the `FA3_TAX_RATE_MAP` dictionary with all 10 `P_12` values already exists and is unreachable); a mixed-rate basket is mis-taxed domestically; and a per-line rate that Erli already reports is discarded because nothing downstream can hold it.

**The rule this epic implements:** the rate arrives from the ProductMaster with the product, the marketplace is a fallback when the master does not know, and when neither knows, the document is held rather than guessed. OpenLinker never computes a rate and never computes an amount excluding tax.

## Channel facts (verified 20 Aug 2026)

| Channel | Order line | Product / offer write | Notes |
|---|---|---|---|
[...]

PR #2902, opened 2026-09-05
feat(analytics): add page-level Net/Gross basis toggle (#2895)

### LINKED ISSUE #2895: [FEATURE] Frontend + CORE — Net/Gross basis switch for the Analytics page

norbert-kulus-blockydevs:
## Summary

Adds a page-level Net/Gross basis toggle to `/analytics`, rendered next to the date-range toolbar and the display-currency picker, persisted as the URL param `basis` (`net` | `gross`) - the same pattern `displayCurrency`/`rateBasis` already use.

**This is a pure frontend wiring change.** The backend already computes both bases for every affected figure - `revenue`/`netRevenue`, `averageOrderValue`/`netAverageOrderValue`, `medianOrderValue`/`netMedianOrderValue`, and the per-channel/per-product equivalents - shipped by the earlier net-sales tax-rate epic (#2245/#2469). No backend file, migration, or aggregation SQL is touched, so this PR cannot conflict with the concurrent GMV/Units/AOV fixes (#2892/#2893/#2894) on `order-record.repository.ts`/`order-sales-aggregation.ts`.

**Note for reviewers - possible duplicate:** PR #2901 already exists for this same issue, opened concurrently by another agent run. That PR takes a different approach (adds a `netGrossBasis` field to the persisted `analytics_display_settings` row with a full backend migration, and stacks the toggle below the currency picker) but, from its file list, does **not** appear to touch `analytics-kpi-strip.tsx`, `channel-sales-table.tsx`, or `product-sales-table.tsx` - i.e. it wires the toggle control and a settings-persisted default, but not the actual metric-switching logic the issue's acceptance criteria require (Revenue/AOV/Median/per-channel/per-product figures recomputing on toggle). This PR does implement that switching. Please compare the two and keep/merge whichever (or both, reconciled) makes sense - only one should land.

## What changed

- New `apps/web/src/features/analytics/lib/money-basis.lib.ts`: the `MoneyBasis` type (`'gross' | 'net'`), `DEFAULT_MONEY_BASIS`, `parseMoneyBasis` (URL-param coercion), and small basis-selector helpers - pure, unit-tested.
- New `AnalyticsMoneyBasisToggle` component (a `SegmentedControl` wrapper, matching `AnalyticsCurrencyPicker`'s existing pattern), rendered in the date-range toolbar's `trailing` slot alongside the currency picker.
- `analytics-page.tsx`: reads/writes the `basis` URL param exactly like `displayCurrency`/`rateBasis`, passes `basis` down to `AnalyticsKpiStrip`, `ChannelSalesTable`, `ProductSalesTable`.
- `AnalyticsKpiStrip`: the Revenue card's primary/qualifier pair (GMV vs Net sales) swaps role per basis; the Order value card's AOV/Median swap between the gross and net fields the backend already returns.
- `ChannelSalesTable` / `ProductSalesTable`: the single money column and its header label swap between `revenue`/`netRevenue` (and `averageOrderValue`/`netAverageOrderValue` for the channel table's AOV column) per basis.
- `docs/specs/metrics-analytics-dashboard.md`: new section documenting the toggle's effect per metric.

## Default basis: `net`, not `gross` - a deliberate deviation from the issue's stated assumption

The issue's acceptance criteria assert two things that turned out to be in tension against the current codebase:

1. "Gross (default) renders every figure exactly as today, byte-for-byte" (repeated three times, explicitly the reviewer's primary check).
2. "Default basis is Gross" (stated as an assumption).

By the time this was implemented, the KPI strip / by-channel table / top-products table **already rendered the VAT-exclusive ("net") figures as their primary numbers by default** - `netRevenue` labeled "Net sales" as the Revenue card's primary metric (with gross `revenue`/GMV as a qualifier), `netAverageOrderValue`/`netMedianOrderValue` as the Order value card's figures, and `netRevenue`/`netAverageOrderValue` as the only money columns in the by-channel/top-products tables (no gross columns rendered there at all pre-toggle). This was shipped ahead of #2895 by the net-sales tax-rate epic (#2245/#2469), which the #2895 issue's own assumption section did not anticipate.

Since satisfying both (1) and (2) literally is impossible given that starting state, I prioritized (1) - the explicit, testable, repeatedly-emphasized criterion - over (2), an assumption. The toggle therefore defaults to `net` (reproducing exactly what the page already rendered), and `gross` is the newly-added alternative view. This is stated explicitly in the new `money-basis.lib.ts` doc comment and in the new metrics-spec section, so it is not a silent departure from the issue text.

## Net Sales card: approach taken

The codebase never shipped two separate GMV/Net-Sales cards - the shipped "Revenue" card already carries both figures at once (primary + qualifier). I kept that single card and made the toggle swap which figure is primary vs. qualifier, rather than splitting into two cards (issue's approach (a)) or collapsing further (approach (b)). Both GMV and Net Sales stay visible on the card in both toggle states; only their roles (headline vs. qualifier) swap. This is the smallest change consistent with "prefer (a) unless it creates confusing duplication" - a genuine second card would duplicate information the qualifier slot already carries.

## What does NOT switch, and why

- **Units Sold, Number of Orders, Average Daily Orders, Cancellation Rate, Return Rate** - not currency figures, untouched by either code path.
- **Cancellations Value** - no VAT-exclusive `netCancelledValue` field exists anywhere in the backend (`SalesAnalyticsHeadline`/`ChannelSalesAnalytics` carry `cancelledValue` only), so it stays on its one existing basis regardless of the toggle. Adding a net figure for it is new backend aggregation, out of scope for a wiring-only PR (and risks colliding with the sibling PRs' SQL). Documented explicitly in the spec update.
- **Returns Value** - no return/refund entity exists in the orders domain yet; the KPI card keeps rendering its existing "planned" placeholder regardless of basis.
- Net-eligibility exclusion reporting (`netExcludedCount`/`netExcludedValue`) is reused verbatim, unchanged - no second exclusion channel was invented.

## Gross mode is byte-identical to today

Verified: with the toggle at its default (`basis` param absent, resolving to `net`), the page renders identically to `main` - same primary/qualifier assignment, same values, same labels. This is asserted directly by new tests (see below) that pin the default rendering against the exact same fixture values the pre-existing tests already used.

## Tests

- `money-basis.lib.test.ts` (new) - unit tests for the default, the URL-param parser, and the pure selector helpers.
- `analytics-money-basis-toggle.test.tsx` (new) - renders, shows the checked state, calls `onChange` with the right value.
- `analytics-kpi-strip.test.tsx` - added cases asserting (a) default `net` rendering is byte-identical to the pre-toggle fixture, (b) `basis="gross"` swaps GMV/Net-sales roles and AOV/Median fields, (c) non-monetary figures never change.
- `channel-sales-table.test.tsx` / `product-sales-table.test.tsx` - added cases for the header label and money-column swap under both bases.
- `analytics-page.test.tsx` - added cases asserting the toggle renders at the toolbar tier, defaults to Net, and clicking Gross updates the URL and recomputes the KPI strip's figures end to end.

Ran (scoped, per the task's constraint against the full suite):
```
pnpm --filter <n/a, vitest run directly> exec vitest run <the 6 changed/added analytics test files>
# 6 test files, 87 tests, all passing
pnpm exec eslint --ext .ts,.tsx src/features/analytics src/pages/analytics   # 0 errors, 3 pre-existing warnings (unrelated lines)
pnpm exec tsc -b --pretty false   # clean
```

No integration test exists for this frontend-only change (the sales-analytics integration tests live in `apps/api`/`libs/core` and are unaffected, since no backend file changed here).

Closes #2895

### DIFF
--- a/apps/web/src/features/analytics/components/analytics-kpi-strip.test.tsx
+++ b/apps/web/src/features/analytics/components/analytics-kpi-strip.test.tsx
@@ -101,6 +101,59 @@ describe('AnalyticsKpiStrip', () => {
     expect(screen.getByText('PLN 200.00')).toBeInTheDocument();
   });
 
+  it('should render byte-identical net figures when basis is omitted (default)', async () => {
+    const apiClient = createMockApiClient({
+      analytics: { getSales: vi.fn().mockResolvedValue(analytics()) },
+    });
+
+    renderWithProviders(<AnalyticsKpiStrip filters={FILTERS} connections={[]} />, { apiClient });
+
+    // Same assertions as the "reporting currency" test above — this is the
+    // #2895 acceptance criterion: an absent `basis` prop must render exactly
+    // as it did before the toggle existed.
+    expect(await screen.findByText('Net sales')).toBeInTheDocument();
+    expect(screen.getByText('PLN 4,200.00')).toBeInTheDocument();
+    expect(screen.getByText('GMV')).toBeInTheDocument();
+    expect(screen.getByText('PLN 4,800.00')).toBeInTheDocument();
+    expect(screen.getByText('PLN 105.00')).toBeInTheDocument();
+    expect(screen.getByText('PLN 90.00')).toBeInTheDocument();
+  });
+
+  it('should render gross figures as primary and net as qualifier when basis="gross"', async () => {
+    const apiClient = createMockApiClient({
+      analytics: { getSales: vi.fn().mockResolvedValue(analytics()) },
+    });
+
+    renderWithProviders(<AnalyticsKpiStrip filters={FILTERS} connections={[]} basis="gross" />, {
+      apiClient,
+    });
+
+    // GMV (gross revenue) is now primary, Net sales the qualifier — both
+    // still visible, roles swapped.
+    expect(await screen.findByText('GMV')).toBeInTheDocument();
+    expect(screen.getByText('PLN 4,800.00')).toBeInTheDocument();
+    expect(screen.getByText('Net sales')).toBeInTheDocument();
+    expect(screen.getByText('PLN 4,200.00')).toBeInTheDocument();
+    // AOV/median switch to the gross fields (120 / 100), not the net ones.
+    expect(screen.getByText('PLN 120.00')).toBeInTheDocument();
+    expect(screen.getByText('PLN 100.00')).toBeInTheDocument();
+  });
+
+  it('should never change non-monetary figures (Units, Orders) when basis toggles', async () => {
+    const apiClient = createMockApiClient({
+      analytics: { getSales: vi.fn().mockResolvedValue(analytics()) },
+    });
+
+    renderWithProviders(<AnalyticsKpiStrip filters={FILTERS} connections={[]} basis="gross" />, {
+      apiClient,
+    });
+
+    expect(await screen.findByText('40')).toBeInTheDocument();
+    expect(screen.getByText('60')).toBeInTheDocument();
+    // Cancelled value has no net counterpart and stays on its one basis.
+    expect(screen.getByText('PLN 200.00')).toBeInTheDocument();
+  });
+
   it('should label the sparklines with the actual selected range, not a hardcoded "last 7 days"', async () => {
     const apiClient = createMockApiClient({
       analytics: {
--- a/apps/web/src/features/analytics/components/analytics-kpi-strip.tsx
+++ b/apps/web/src/features/analytics/components/analytics-kpi-strip.tsx
@@ -77,6 +77,7 @@ import {
 } from '../lib/display-currency.lib';
 import { AnalyticsInfotip } from './analytics-infotip';
 import { resolveEarliestOrderDate } from '../lib/ingestion-trust.lib';
+import { DEFAULT_MONEY_BASIS, type MoneyBasis } from '../lib/money-basis.lib';
 import {
   averageDailyOrders,
   cancellationRate,
@@ -146,6 +147,16 @@ interface AnalyticsKpiStripProps {
   coverage?: AnalyticsCoverage;
   /** Opens the matching Data Coverage detail modal — omit to keep every `GapMark` inert. */
   onOpenCategory?: (category: CoverageCategory) => void;
+  /**
+   * Page-level Net/Gross view preference (#2895) — defaults to `net`, which
+   * reproduces every figure exactly as this component rendered before the
+   * toggle existed (see `lib/money-basis.lib.ts`'s own doc comment for why
+   * that is the byte-identical default, not `gross`). `gross` swaps the
+   * Revenue card's primary/qualifier pair and the Order value card's
+   * figures to their VAT-inclusive counterparts — Units/Orders/Cancellation
+   * rate never change, since they carry no basis at all.
+   */
+  basis?: MoneyBasis;
 }
 
 /** Tax categories partition the exclusion set (`TaxCoverageDetectionService`'s classification pass) — the one with the largest `affectedCount` is the category actually causing `netExcludedCount`. Falls back to `'tax-a'` (the remediable one) if all three read zero, which should not happen while `netExcludedVisible` is true. */
@@ -166,6 +177,7 @@ export function AnalyticsKpiStrip({
   filters,
   coverage,
   onOpenCategory,
+  basis = DEFAULT_MONEY_BASIS,
 }: AnalyticsKpiStripProps): ReactElement {
   const query = useSalesAnalyticsQuery(filters);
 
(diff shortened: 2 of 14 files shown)


norbert-kulus-blockydevs:
Closing as a duplicate of #2901, which was built to the same spec (issue #2895) plus the follow-up refinement (toggle stacked below the currency picker, dual persistence via URL state + AnalyticsSettingsDialog). Keeping #2901 as the version being integrated into epic/2452-analytics-currency-coverage.

PR #2913, opened 2026-09-05
fix(web): make headline Revenue card track the Net/Gross toggle

### LINKED ISSUE #2908: [BUG] Frontend — Headline Revenue card ignores the Net/Gross toggle while channel/product tables respect it
## Problem / Context

The KPI-strip headline Revenue card in `apps/web/src/features/analytics/components/analytics-kpi-strip.tsx` always renders `headline.netRevenue` labeled "Net sales", with `headline.revenue` (GMV) as a fixed, toggle-independent qualifier — it does not read `netGrossBasis` at all. The channel and product tables directly below it correctly flip their main money column (and their AOV column, in lockstep) between GMV and Net Sales based on the toggle.

Net effect: setting the toggle to Gross relabels the tables below to "GMV" showing gross totals, while the headline card above still says "Net sales" showing the net total — two differently-labeled totals for "revenue" on one page under one toggle state, and the headline won't reconcile with the (now-gross) channel rows summed. This undercuts the toggle's own stated goal of making the page's figures respond to the switch.

## Proposed Solution

Make the headline Revenue card respect `netGrossBasis` the same way the tables do: swap which of GMV/Net Sales is the primary rendered value and which is the qualifier line based on the toggle, so the headline and the tables always agree on which basis is "primary" for the current toggle state. Both values may still render (primary + qualifier), matching the existing established pattern, but the LABEL and PRIMARY VALUE must track the toggle consistently with the tables below.

## Classification

**Type**: Frontend
**Layer**: Interface
**File(s)**: `apps/web/src/features/analytics/components/analytics-kpi-strip.tsx`

## Dependencies

- None; follow-up to #2895/#2903 (merged).

## Acceptance Criteria

- [ ] Toggling Net/Gross changes which figure (GMV vs Net Sales) is the headline card's PRIMARY rendered value and label, consistent with the channel/product tables' basis at the same toggle state
- [ ] The headline's primary value reconciles with the sum of the (same-basis) channel table rows

norbert-kulus-blockydevs:
## Summary

The KPI-strip headline Revenue card in `analytics-kpi-strip.tsx` always rendered `headline.netRevenue` labeled "Net sales" as primary, with `headline.revenue` (GMV) as a fixed qualifier - it never read `netGrossBasis` for this card at all. The channel and product tables below it correctly flip their primary money column (and AOV) based on the toggle, so setting the toggle to Gross left the page internally inconsistent: tables show GMV as primary while the headline still said "Net sales".

## Fix

- The headline Revenue card's primary rendered value/label now tracks `netGrossBasis` the same way `channel-sales-table.tsx`'s `revenueLabel`/`revenueOf` already do: under `'gross'` (the default) GMV becomes primary with Net sales as the qualifier; under `'net'` Net sales stays primary as it was before this fix.
- Both figures still always render - the toggle only swaps which one is primary/labelled as such, not which values are shown.
- Checked the AOV/Median card (#2894/#2903) as the reference for basis-driven rendering, and mirrored its approach for the Revenue card.
- Checked the delta logic: the Revenue card has never carried a period-over-period delta (confirmed in the component's own doc comment and code), so there was no fixed-basis-vs-toggled-current comparison to fix there.

## Test plan

- [x] `pnpm --filter @openlinker/web test -- analytics-kpi-strip` - 26/26 passing (24 pre-existing + 2 new)
- [x] Added two component tests asserting the primary/qualifier swap at both toggle states (checking the value's placement in `.kpi-card__value` vs `.kpi-card__qualifier-value`, and label placement in `.kpi-card__metric` vs `.kpi-card__qualifier-label`)

Closes #2908

### DIFF
--- a/apps/web/src/features/analytics/components/analytics-kpi-strip.test.tsx
+++ b/apps/web/src/features/analytics/components/analytics-kpi-strip.test.tsx
@@ -157,6 +157,52 @@ describe('AnalyticsKpiStrip', () => {
     expect(screen.queryByText('PLN 100.00')).not.toBeInTheDocument(); // medianOrderValue
   });
 
+  it('should render GMV as the primary Revenue figure and Net sales as its qualifier under the gross basis (#2908 fix)', async () => {
+    // Under `netGrossBasis="gross"` (the default), the headline Revenue
+    // card's primary figure must swap to GMV — matching the channel/product
+    // tables' own `revenueLabel`/`revenueOf` primary-column choice for the
+    // same basis — while Net sales moves to the qualifier row. Neither
+    // figure is dropped; only which one is primary changes.
+    const apiClient = createMockApiClient({
+      analytics: { getSales: vi.fn().mockResolvedValue(analytics()) },
+    });
+
+    renderWithProviders(
+      <AnalyticsKpiStrip filters={FILTERS} connections={[]} netGrossBasis="gross" />,
+      { apiClient }
+    );
+
+    const primaryValue = await screen.findByText('PLN 4,800.00'); // headline.revenue (GMV)
+    expect(primaryValue.closest('.kpi-card__value')).not.toBeNull();
+    expect(screen.getByText('GMV').closest('.kpi-card__metric')).not.toBeNull();
+
+    const qualifierValue = screen.getByText('PLN 4,200.00'); // headline.netRevenue (Net sales)
+    expect(qualifierValue.closest('.kpi-card__qualifier-value')).not.toBeNull();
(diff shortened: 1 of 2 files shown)


norbert-kulus-blockydevs:
Merged into `epic/2452-analytics-currency-coverage` at c1686f776cc0db0db208aa362288f99caa4814b3 (clean merge, no conflicts). Targeted type-check + tests green (an unrelated `order-fulfillment-tasks-panel` flake surfaced only under the full-suite run; confirmed pre-existing and unrelated to this change, passes in isolation).

Issue #2910, opened 2026-09-05
[BUG] CORE — Cancelled Value is gross and shipping-inclusive; spec requires net


norbert-kulus-blockydevs:
## Problem / Context

`docs/specs/metrics-analytics-dashboard.md` defines Cancellations Value as: "SUM(net value of orders placed in the period with cancelled status)" — explicitly net (VAT-excluded). The shipped `cancelled_value` in `libs/core/src/orders/infrastructure/persistence/repositories/order-record.repository.ts` sums `reportingTotalAmount` for cancelled orders — the shipping-inclusive, gross order total (the same column #2892 deliberately avoided for GMV). Pre-existing, not introduced by the recent GMV/AOV/Median/toggle fixes, but a real spec/code mismatch surfaced during a full-page audit.

## Proposed Solution

Compute Cancelled Value net-of-VAT using the same per-line net-computation machinery Net Sales already uses (`resolveNetSalesTaxRate`/`netSalesRateFractionSql`), applied to cancelled orders' lines instead of non-cancelled ones, with the same exclusion-reporting discipline (an order with an unresolvable rate reported via the existing pattern, no new silent exclusion).

## Classification

**Type**: CORE
**Layer**: Infrastructure (repository query)
**File(s)**: `libs/core/src/orders/infrastructure/persistence/repositories/order-record.repository.ts`

## Dependencies

- None

## Acceptance Criteria

- [ ] Cancelled Value is computed net-of-VAT, matching the spec's "net value" wording
- [ ] Uses the same tax-rate resolution/exclusion pattern as Net Sales
- [ ] Tests added or updated for non-trivial logic
- [ ] No architecture boundary violations (CORE ↔ Integration)

Issue #2909, opened 2026-09-05
[BUG] Frontend — Verify KPI-strip per-card labelling under a converted display currency


norbert-kulus-blockydevs:
## Problem / Context

`GET /analytics/sales?displayCurrency=EUR` correctly converts the `revenue` figure only — AOV, Median, Units Sold, Net Sales, and Cancelled Value all stay in native currency (by design, per the controller's own docblock), and `headline.currency` in the response still reads the native currency code. Whether the FRONTEND correctly labels/scopes which KPI cards are actually converted vs. still-native under a converted display-currency view was not verified during a full-page audit. If the KPI strip renders under one blanket currency heading with no per-card distinction, an operator could read a still-native-currency figure as if it were in the selected display currency — a real, user-facing mislabeling risk.

## Proposed Solution

Audit `apps/web/src/features/analytics/components/analytics-kpi-strip.tsx` (and `AnalyticsConvertNote`) for exactly which cards render a currency-suffixed value when `displayCurrency` is active, and confirm each card's currency label/suffix matches the ACTUAL currency of the value it renders (native PLN for unconverted cards, the selected display currency for the converted Revenue card only). Add a visible distinction (currency suffix, a small "native" qualifier, or similar) on any card whose value did NOT convert, so it cannot be misread as being in the selected display currency. Fix any mislabeling found.

## Classification

**Type**: Frontend
**Layer**: Interface
**File(s)**: `apps/web/src/features/analytics/components/analytics-kpi-strip.tsx`, `apps/web/src/features/analytics/components/analytics-convert-note.tsx`

## Dependencies

- None

## Acceptance Criteria

- [ ] Every KPI card's rendered currency label matches the actual currency of the value shown, whether converted or native
- [ ] A converted-view session cannot misread a native-currency card as being in the display currency
- [ ] Tests added or updated for non-trivial logic
- [ ] No architecture boundary violations (CORE ↔ Integration)

Issue #2907, opened 2026-09-05
[BUG] Frontend — Channel table Units Sold doesn't add back unconverted orders' units


norbert-kulus-blockydevs:
## Problem / Context

The headline KPI-strip fix for #2893 correctly computes `unitsSold + unconvertedUnitsSold`. `apps/web/src/features/analytics/components/channel-sales-table.tsx` reads `channel.unitsSold` directly with no add-back, and the table's own "Total" row (built in `apps/web/src/features/analytics/lib/sales-analytics-view-model.ts`'s `groupChannelTotalsByCurrency`) sums the same un-reconciled field. Whenever any channel has unconverted orders (the normal steady state this whole epic exists to handle), the channel table's units under-count relative to the headline card, and channel rows won't sum to the headline total.

## Proposed Solution

Mirror the headline fix: add `channel.unconvertedUnitsSold` to whatever field the channel table renders as Units Sold, both for individual channel rows and the aggregated Total row in `groupChannelTotalsByCurrency`.

## Classification

**Type**: Frontend
**Layer**: Interface
**File(s)**: `apps/web/src/features/analytics/components/channel-sales-table.tsx`, `apps/web/src/features/analytics/lib/sales-analytics-view-model.ts`

## Dependencies

- None; sibling fix to #2893 (merged) that missed this component.

## Acceptance Criteria

- [ ] Channel table Units Sold column includes unconverted orders' units for every row
- [ ] The Total row sums to the same figure the headline KPI card shows for Units Sold
- [ ] Tests added or updated proving the reconciliation (a fixture with `unconvertedUnitsSold > 0` on at least one channel)
- [ ] No architecture boundary violations (CORE ↔ Integration)

PR #2914, opened 2026-09-05
fix(analytics): Gross Median Order Value uses the same basis as AOV/Revenue

### LINKED ISSUE #2906: [BUG] CORE — Gross Median Order Value uses shipping-inclusive basis, contradicting AOV/Revenue
## Problem / Context

`docs/specs/metrics-analytics-dashboard.md` requires Median Order Value to use "the same set of orders and the same value field as in AOV, so that the two figures are comparable." AOV/Revenue (post-#2892) are computed from `order_line_items` (merchandise only, excludes shipping). `getMedianOrderValue` in `libs/core/src/orders/infrastructure/persistence/repositories/order-record.repository.ts` instead runs `PERCENTILE_CONT(0.5)` over `reportingTotalAmount` — the shipping-inclusive order total. `getNetMedianOrderValue` (the Net-mode sibling) does NOT have this bug; it's already built on the correct line-item basis.

Live-proven: for `from=2026-08-06&to=2026-09-05` on the demo dataset, `medianOrderValue` returns 300; the value computed on the same line-item basis as AOV/Revenue for the identical order set is 200. This is a residual gap in #2892 — that fix updated `revenue`/`netRevenue`/`getDailyOrderAggregates` but missed the gross median function.

## Proposed Solution

Rewrite `getMedianOrderValue` to compute `PERCENTILE_CONT(0.5)` over the same per-order GMV line-item amount `buildGrossRevenueOrderAmountSql`/`grossRevenueLineAmountSql` already compute for revenue, mirroring exactly how `getNetMedianOrderValue` already does this correctly for its own basis.

## Classification

**Type**: CORE
**Layer**: Infrastructure (repository query)
**File(s)**: `libs/core/src/orders/infrastructure/persistence/repositories/order-record.repository.ts`

## Dependencies

- None

## Acceptance Criteria

- [ ] Gross Median Order Value is computed from the same per-order line-item GMV amount as AOV/Revenue, not `reportingTotalAmount`
- [ ] A test proves median and AOV/revenue now share the same underlying value field for a fixture with non-zero shipping
- [ ] Net Median Order Value is unaffected (already correct)

norbert-kulus-blockydevs:
## Summary

`getMedianOrderValue` (`libs/core/src/orders/infrastructure/persistence/repositories/order-record.repository.ts`) ran `PERCENTILE_CONT(0.5) WITHIN GROUP (ORDER BY rec."reportingTotalAmount")` directly over the shipping-inclusive order total. Meanwhile AOV/Revenue (post-#2892) are computed from `order_line_items` via `buildGrossRevenueOrderAmountSql` / `grossRevenueLineAmountSql` - the gross, shipping-EXCLUDED merchandise value. This contradicted `docs/specs/metrics-analytics-dashboard.md`, which requires median and AOV to share "the same value field", and inflated the reported median on any order set carrying non-zero shipping.

`getNetMedianOrderValue` (the Net-mode sibling) never had this bug - it already runs `PERCENTILE_CONT` over `buildNetSalesOrderFragments`'s per-order net amount, converted through the order's own stamped FX multiplier. This PR mirrors that exact pattern for the gross case, reusing `buildGrossRevenueOrderAmountSql` (no new SQL invented). `getNetMedianOrderValue` itself is untouched.

## The fix

```diff
- .select(`PERCENTILE_CONT(0.5) WITHIN GROUP (ORDER BY rec."reportingTotalAmount")`, 'median')
+ .select(
+   `PERCENTILE_CONT(0.5) WITHIN GROUP (ORDER BY (${grossRevenueOrderAmount}) * (rec."reportingTotalAmount" / NULLIF(rec."totalAmount", 0)))`,
+   'median'
+ )
```

where `grossRevenueOrderAmount` is `this.buildGrossRevenueOrderAmountSql()` - the identical per-order scalar subquery `getDailyOrderAggregates`'s `revenue` column already sums, converted the identical way (`* (reportingTotalAmount / NULLIF(totalAmount, 0))`, the order's own already-stamped FX multiplier, never a re-lookup).

## Before/after proof

Added a new integration test proving the exact live-observed regression:

> a single order with a shipping-inclusive total of **300** and a merchandise-only line value of **200** now reports a median of **200** (matching AOV/Revenue), instead of the old **300**.

```
✓ getMedianOrderValue shares AOV/Revenue's gross, shipping-EXCLUDED basis instead of the shipping-inclusive order total (#2906) (51 ms)
```

Also updated the pre-existing multi-order median test to seed line items with a flat shipping delta per order (so a regression back to `reportingTotalAmount` shifts the median from 20 to 25 and fails), and added a repository unit test asserting the `SELECT` SQL reads from `order_line_items` rather than the bare column.

## Test plan

- [x] `pnpm --filter @openlinker/core type-check` - passes
- [x] `pnpm --filter @openlinker/core test -- order-record` - 247/247 passed (both repo unit specs + service specs)
- [x] `pnpm --filter @openlinker/api test:integration -- sales-analytics-aggregates` (real Postgres via Testcontainers) - 20/20 passed, including the two new `#2906` cases

## Scope note

Per the task instructions, this PR is scoped ONLY to the gross median function. Four sibling issues (#2907 channel-table units, #2908 headline toggle scope, #2909 currency labelling, #2910 Cancelled Value net basis) are being fixed concurrently in isolated worktrees against overlapping files; conflicts at integration time are expected and not this PR's concern.

Closes #2906

### DIFF
--- a/apps/api/test/integration/sales-analytics-aggregates.int-spec.ts
+++ b/apps/api/test/integration/sales-analytics-aggregates.int-spec.ts
@@ -250,34 +250,46 @@ describe('Sales analytics daily aggregates (integration, #1987)', () => {
     expect(median).toBeNull();
   });
 
-  it('getMedianOrderValue computes PERCENTILE_CONT over stamped, non-cancelled orders', async () => {
+  it('getMedianOrderValue computes PERCENTILE_CONT over stamped, non-cancelled orders, on the gross merchandise-only basis (#2906)', async () => {
     const day = new Date('2026-08-05T10:00:00.000Z');
-    await seedStampedOrder({
+    // Each order's `totalAmount` (shipping-INCLUSIVE) diverges from its
+    // merchandise-only line-item value by a flat 5 shipping charge, so a
+    // regression back to `reportingTotalAmount` would shift the median from
+    // 20 to 25 — the assertion below fails on that basis, not just on this
+    // one.
+    const order1 = await seedStampedOrder({
       placedAt: day,
-      totalAmount: 10,
+      totalAmount: 15,
       reportingCurrency: 'EUR',
-      reportingTotalAmount: 10,
+      reportingTotalAmount: 15,
     });
-    await seedStampedOrder({
+    await seedLineItem({ orderRecordId: order1, unitPrice: 10, quantity: 1 });
+
+    const order2 = await seedStampedOrder({
       placedAt: day,
-      totalAmount: 20,
+      totalAmount: 25,
       reportingCurrency: 'EUR',
-      reportingTotalAmount: 20,
+      reportingTotalAmount: 25,
     });
-    await seedStampedOrder({
+    await seedLineItem({ orderRecordId: order2, unitPrice: 20, quantity: 1 });
+
+    const order3 = await seedStampedOrder({
       placedAt: day,
-      totalAmount: 30,
+      totalAmount: 35,
       reportingCurrency: 'EUR',
-      reportingTotalAmount: 30,
+      reportingTotalAmount: 35,
     });
+    await seedLineItem({ orderRecordId: order3, unitPrice: 30, quantity: 1 });
+
     // Cancelled — must not enter the median.
-    await seedStampedOrder({
+    const cancelledOrder = await seedStampedOrder({
       placedAt: day,
       totalAmount: 1000,
       reportingCurrency: 'EUR',
       reportingTotalAmount: 1000,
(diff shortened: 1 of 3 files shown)


norbert-kulus-blockydevs:
Merged into `epic/2452-analytics-currency-coverage` at dcf703ebb587a69b8ab6172127cf48f562fd3b22 (clean merge, no conflicts). Targeted type-check + tests green.

Issue #2842, opened 2026-09-03
[TASK] DX — F3: webhook ingress burst throughput


piotrswierzy:
Child of #2840. **Week 1.** The first scenario, and the one that proves the rig.

## Problem / Context

`POST /webhooks/:provider/:connectionId` (`apps/api/src/webhooks/http/webhook.controller.ts:51`) is the durability spine described in `docs/architecture-overview.md` § *Webhook Ingestion Flow*: since #2280 the `sync_jobs` row and the `webhook_deliveries` row commit in **one Postgres transaction**, guarded by the `uq_webhook_deliveries_event_key` unique index on `(provider, connection_id, event_id)`.

Correctness is well tested. **Throughput and lock contention under burst never were.** It is also the only ingress an attacker can reach, and `webhook_auth_rejections` is an append-only table an unauthenticated caller can make grow. The route has no throttler anywhere in `apps/api/src`, so it is genuinely unthrottled.

Chosen as scenario one because it needs **no stub, no seeded data and no auth handshake** - a signed POST loop - so it exercises the whole arrange/window/post-guard machinery end to end at minimum cost. If the harness abstraction is wrong, that surfaces here rather than after four more scenarios are built on it.

## Proposed Solution

A k6 driver (`drivers/webhook-burst.js`) run as `grafana/k6` on the stack's own Docker network, plus a scenario script using #2841's `lib.sh`.

- **`ramping-arrival-rate`, never `ramping-vus`.** A VU-based executor self-throttles when the system slows and hides the exact cliff being looked for.
- HMAC signing **reused, not re-implemented**. `apps/e2e/src/support/webhooks.ts` already reproduces the OL-HMAC scheme byte-for-byte and `apps/e2e/src/support/webhook-secret.ts` owns secret rotation plus restore. A second signer written in k6 is exactly the arrange drift #2841 exists to prevent, so the k6 driver derives its signing from that support code rather than restating the scheme.
- The timestamp header stays inside the replay window (`OL_WEBHOOK_SKEW_WINDOW_MS`, default 120 s).
- `check()` the status and tag it - non-2xx must not land in the same `http_req_duration` bucket that gets reported. #2590 inverted a conclusion once because a probe counted fast errors as fast responses.
[...]

Issue #2845, opened 2026-09-03
[TASK] DX — run-all campaign command, quick mode, and cross-flow report


piotrswierzy:
Child of #2840. **Week 1**, after >=2 scenarios exist.

## Problem / Context

With per-flow scenarios in place, running the campaign by hand means remembering the order, resetting state between flows, and collating results afterwards. A full campaign is an **overnight job** - arithmetic:

```
2 min warm-up + 10 min window + 1 min cooldown = 13 min per run
x 3 runs (discard run 1, publish 2 and 3 if
          they agree; else run a 4th)           = 39 min per flow (more if they disagree)
x 8 flows                                        ~ 5.2 h
+ F5 at three dataset sizes, F4/F7 at two replica counts
                                                 ~ 8 h
```

"Median of 2 and 3" is the wrong name for what discarding run 1 leaves behind: with two survivors it is their MEAN, so a contaminated run contributes half its error to the published figure rather than being outvoted the way a real median of three would outvote it - see the agreement rule below.

That is what a valid measurement costs. The right response is **unattended and resumable**, not shorter windows.

**Flows cannot run in parallel on one stand.** Two concurrent flows contend for the same worker lanes, connection pool, Redis limiter and destination stub, so each becomes the other's noise and both results are invalid. Serialising is the definition of a controlled measurement, not a harness limitation.

## Proposed Solution

`run-all.sh` - a loop over scenarios, plus the things that make it more than a loop:

1. **Reset between flows.** Flow N must not inherit flow N-1's queue, cursors or drifted data. This is just re-running #2841's guards between each - the concrete payoff of the shared library. The reset is a WAIT-TO-DRAIN, never a purge: #2590's `PURGE_QUEUE` shrank `sync_jobs` mid-campaign and made its read-path figure an underestimate (epic #2840's out-of-scope table), and #2843's own scenario needs an unpurged history - so `run-all` drains the queue to a clean STATE between flows, it does not delete rows to reach one.
2. **Unattended safety.** Guards abort, never prompt. A `DISCARDED` run records its verdict and the campaign continues; the summary reports it.
3. **`--resume`.** A campaign that dies at hour five must not restart from zero.
[...]

PR #2918, opened 2026-09-05
feat(perf): shared harness library for the measurement programme

### LINKED ISSUE #2841: [TASK] DX — perf harness library: guards, manifest, window, verdict, results contract
Child of #2840. **Week 1. Blocks every other child.**

## Problem / Context

`perf/prestashop-baseline` is seven runner scripts (`run-scenario.sh`, `run-a1a.sh`, `run-a1a-at-offset.sh`, `run-a2.sh`, `run-a3.sh`, `run-a4.sh`, `run-a4-ingest.sh`) each re-implementing login, the drain-wait loop, env defaults and the queue purge. Its README records what that cost:

> *`run-a1a-at-offset.sh` … the first version of that script silently failed to seed (it wrote a `cursorValue` column that does not exist; the real one is `value`) and measured offset 0 while claiming offset 99 000.*

Arrange logic drifted, one copy was wrong, and the run **lied about the conditions it was taken under**. Eight new flows means eight more chances at it.

Every trap in that README is an arrange or observe problem currently held as prose that a human must remember at 6pm. This child converts them into executable preconditions.

## Proposed Solution

`perf/openlinker-throughput/lib.sh` - a few hundred lines, no config format, no driver abstraction, no plugin model. It owns exactly what must not drift:

| Function | Owns |
|---|---|
| `login` | one implementation, with the defensive `accessToken` / `access_token` parse `run-scenario.sh` already carries |
| `guard_queue_empty` | no `queued`/`running` rows for the connection |
| `guard_scheduler_off` | no cron can fire into the window |
| `guard_demo_mode_off` | demo mode off, so no read-only lock, registration rate limit or account-cleanup tick lands mid-window |
| `guard_connection_budget` | `Σ(OL_DB_POOL_MAX × processes) < max_connections` - pre-flight, fatal |
| `guard_build` | the running image's `org.opencontainers.image.revision` `LABEL` matches `git rev-parse HEAD` - **never** trust that a build succeeded |
| `guard_pool_recorded` | `OL_DB_POOL_MAX` and Postgres `max_connections` are both resolved and written to the manifest |
| `manifest` | write git sha, image digest, lane caps, dataset sizes, replica count, worker log level, pool ceiling, the `operational_settings` row **before** load |
| `sample_queue` | 1 s poller over `sync_jobs` + `GET /v1/connections/:id/sync-status`, plus `docker stats` and `pg_database_size` |
| `window_start` / `window_stop` | warm-up discarded consistently |
| `post_guard_attempts` | attempts delta > 1 per job ⇒ `DISCARDED` |
| `post_guard_deferrals` | `SUM("deferredTotalMs")` delta over the window is 0, and no job was deferred ⇒ otherwise `DISCARDED` |
| `post_guard_requeues` | `StuckJobRecoveryService`'s own affected-row count for the window is 0, or no window row shows `lockedBy IS NULL` with an unchanged `attempts` ⇒ otherwise `DISCARDED` |
| `guard_runner_state` | the running worker's `WORKER_RUNNER_ENABLED` posture matches what the scenario declares - fatal, pre-flight |
| `guard_log_level` | `OL_LOG_BODY_MAX_BYTES` is set to a positive value - fatal if unset or `0` |
| `guard_perf_max_attempts` | fatal, pre-flight: perf-enqueued job types resolve a low `maxAttempts`, or the drain loop's own timeout path marks a stalled row `dead` and records it - so one unseeded or misconfigured row cannot silently reserve ~30 h behind `nextRunAt` and starve every later flow's drain wait |
| `reset_between_repeats` | between repeats of one scenario: delete the `connection_cursors` row(s) for the scenario's connections **and** every `jobdedup:*` Redis key those connections wrote - a cursor-only reset leaves the 7-day-TTL dedupe reservation standing (see "Repeatability" below) |
| `post_guard_destination_creates` | any `order_records` row created in the window whose `syncStatus` carries a `status: 'failed'` entry, or that carries no `syncedAt` on the destination the scenario declares ⇒ `DISCARDED` |
| `post_guard_limiter_degraded` | the worker log for the window contains the Redis rate-limiter's degraded-mode message ⇒ `DISCARDED` |
| `verdict` | `VALID` / `DISCARDED` + reason |
| `results_dir` | one shape, every scenario - `manifest.json`, `summary.json`, `timeseries.csv`, `verdict.txt`, plus the raw k6 JSON output and a dump of the window's `sync_jobs` rows |

**Guards are fatal and pre-flight.** A guard that aborts before the window costs seconds; one that discards after costs the whole run. That is a wall-clock property as much as a correctness one.

**`post_guard_attempts` and F7 are in tension.** F7 (#2852) exists to demonstrate cross-lane starvation, and its most interesting finding may be a job starved long enough behind a blocked event loop to time out and genuinely retry - an `attempts` delta this guard is specified to discard as contamination. As shipped, `post_guard_attempts` cannot tell that retry apart from an unrelated job failing and it discards both. Whether F7 needs a flow-scoped exemption from `post_guard_attempts`, or whether `post_guard_requeues` should REPORT rather than DISCARD for that one flow, is #2852's decision to make; this child ships the guard globally fatal and records the tension rather than resolving it.

### `guard_build` is redefined: a `LABEL`, not a digest diff

There is no version, git-sha or build-info surface anywhere in `apps/api/src` - `apps/api/src/system/system.controller.ts:24` serves runtime config only - and the Week-1 children add no product code, so there is no new route or registered job type a guard could probe for. A digest comparison against a fresh `docker build` of the working tree does not fix that either: no Dockerfile in the repo carries a `LABEL`, `ARG GIT_SHA` or `VCS_REF` (`grep -rn "LABEL\|ARG GIT_SHA\|VCS_REF" Dockerfile apps/web/Dockerfile` returns zero hits), and an image digest is not content-reproducible - a clean rebuild of identical source can still land a different digest, and a rebuild served from Docker's own layer cache can produce the SAME digest as a stale image without having rebuilt anything. A digest diff therefore either false-fails on every clean run or silently passes on a stale one.

This child adds one `ARG OL_GIT_SHA` plus `LABEL org.opencontainers.image.revision=$OL_GIT_SHA` to the root `Dockerfile`'s `production` and `worker` build stages (the images the api and worker compose services build from), and `guard_build` reads it via `docker inspect --format '{{index .Config.Labels "org.opencontainers.image.revision"}}'` and compares it to `git rev-parse HEAD`. This is a deliberate, narrow exception to "week 1 touches no product code" - two lines in one Dockerfile, taken because this is the only guard standing between the programme and the failure the repo has already suffered once (a perf worker built from a stale image silently measuring pre-epic code). A first-class build-info HTTP field is still #2850's; this is the artefact-level floor underneath it.

### A settle window between pre-flight and the measurement

The heaviest thing that runs on the stand is the image build #2854's `up -d --build` performs (or a rebuild taken to refresh `guard_build`'s label) - and nothing today separates that from `window_start`: guards run immediately after stand-up and the window opens immediately after guards pass, so run 1 opens on a hot CPU and a page cache still full of build artefacts rather than of Postgres pages. `lib.sh` inserts an explicit settle period between the last guard passing and `window_start`.

### `guard_connection_budget` closes a hard blocker on multi-replica runs

The Postgres service carries no `command:` override in any compose file (`docker-compose.yml:30-46` runs plain `postgres:17-alpine`, no mounted conf), so `max_connections` is the 100 default, while `OL_DB_POOL_MAX` defaults to 40 **per process** (`libs/shared/src/database/database.module.ts:40,81`). The api plus three workers is 160 connections against 100, and `--scale worker=3` dies with *"too many clients already"*. `apps/worker/.env.example:513-515` already states the arithmetic and nothing enforces it. The guard enforces it and both numbers appear in every manifest - the pool ceiling is also the real ingress concurrency ceiling #2842 measures against, so it is a reported condition, not just a pre-flight.

### Unattended running needs a drain timeout, not just a drain loop

The measuring replicas run `OL_WORKER_ROLE=jobs`, which excludes `maintenance` - but #2854 arms at least one separate replica with the `maintenance` role, so `StuckJobRecoveryService` DOES requeue a stuck job on this stand. That requeue re-runs the job from scratch after the 15-minute lock timeout (`requeueStuckJobs`, `sync-job.repository.ts:352-369`; threshold and interval at `stuck-job-recovery.service.ts:23-24`) while touching neither `attempts` nor `deferredTotalMs` - invisible to both existing post-guards, which is what `post_guard_requeues` exists to catch. Separately, a penalty-free deferral re-queues the job as `status='queued'` with a future `nextRunAt` (`apps/worker/src/sync/sync-job.runner.ts:548-562,574-590`), clamped up to an hour. Any of the three states stalls a naive "wait until the queue is empty" loop indefinitely, and one hung scenario stalls the whole campaign. The drain loop therefore carries a max-wait timeout plus explicit deferred-job and requeue detection, and reports which it hit.

### Repeatability needs more than a cursor delete

`perf/prestashop-baseline/run-a1a.sh:57` already deletes the `connection_cursors` row between repeats - the legacy harness's own reset step - but it never touched the Redis dedupe layer, and #2590 measured 37 500+ `jobdedup:*` keys standing on the shared `redis-data` volume (`docker-compose.yml:90-91`, so only `down -v` clears them otherwise). The child dedupe key for an ingested order is `marketplace:{connectionId}:order:{item.eventKey}` (`order-ingestion.service.ts:239`), reserved in Redis as `jobdedup:{key}` under `SET NX EX` with a **7-day TTL** (`redis-streams-job-enqueue.service.ts:23-24`, `:36-39`) - an 8-hour campaign sits entirely inside that window. If the upstream stub reissues the same event ids across a repeat (#2856's stub must not, but this reset is the harness-side backstop regardless), every enqueue in the second and third repeat silently no-ops (`{isExisting: true}`, destructured away and never surfaced, `sync-job-queue.service.ts:47`), while the cursor still commits and every existing post-guard still passes - the run reports `n = 0` and reads as `VALID`. `reset_between_repeats` therefore clears both the cursor row and the connections' `jobdedup:*` keys, not the cursor alone.

### A destination-create failure does not fail the job

`OrderSyncService.dispatchToDestinations` fans out with `Promise.allSettled` (`order-sync.service.ts:261-267`); only a lock-contention rejection (`OrderCreateContendedException`) is rethrown (`:281-282`). Every other per-destination failure is swallowed and recorded onto `order_records.syncStatus` as `status: 'failed'` with no `syncedAt` (`order-ingestion.service.ts:706-707`), and the ingestion method returns normally (`:772`). The job itself reports `outcome: 'ok'`, no `sync_jobs` row goes `dead`, `attempts` never moves, and the queue drains clean - so `guard_queue_empty` and every attempts/deferral/requeue post-guard above all pass while zero orders actually reached a destination shop. `post_guard_destination_creates` is the only guard in this library that can see that state, because the only trace of it is the `syncStatus` JSONB column.

### Two cheap post-run checks over conditions that read as healthy

The Redis rate limiter falls back to per-process in-memory limiting on a 1000 ms Redis call timeout (`redis-rate-limiter.adapter.ts:173`), logging at `error` (`:563`, message: *"Redis rate limiter unavailable for connection … - falling back to per-process in-memory limiting (degraded, not unthrottled)"*). #2590 already observed 85 such timeouts inside one 40-minute window. At three replicas a degraded episode silently turns a configured 60/min shop limit into an effective ~180/min for the length of the episode, which invalidates any comparison against a run where the limiter stayed live. `post_guard_limiter_degraded` greps the worker log for that message and discards on a hit.

Separately, `sync_jobs.maxAttempts` defaults to 10 (`sync-job.repository.ts:89`) with exponential backoff capped at 6 h (`sync-job.runner.ts:54-56`), so reaching `dead` takes roughly 30.25 hours. A perf-enqueued job hitting a retryable, non-recoverable condition (an unseeded offer mapping, say) leaves a `queued` row with `nextRunAt` up to 6 h out - invisible to `guard_queue_empty` for the rest of that window, and a drain loop with no timeout would stall on it indefinitely, with the eventual timeout blaming whichever flow happens to run next rather than the row that caused it. `guard_perf_max_attempts` is the pre-flight half of the fix (perf job types resolve a low `maxAttempts`); the alternative recorded above is that the drain loop's own max-wait timeout path marks the offending row(s) `dead` and records them by id, so the campaign can name the cause instead of only reporting a stall.

### The scheduler guard has two halves

`OL_SCHEDULER_ENABLED=false` stops the fleet singleton from competing at all (`apps/worker/src/scheduler/scheduler-lease.coordinator.ts:44-47`), which is the enablement half. The **cadence** half is not env-only: an `operational_settings` row takes precedence over `cronEnvVar` / `defaultCron` (`apps/worker/src/scheduler/scheduler.service.ts:142-151`), so the manifest captures that row, not just the `OL_*_CRON` variables.

### `guard_runner_state` and `guard_log_level` are fatal, pre-flight checks over the running process

`WORKER_RUNNER_ENABLED` defaults true and is checked once, in `onModuleInit` (`apps/worker/src/sync/sync-job.runner.ts:133-137`): if it resolves false the method returns before `startRunner()` runs at all. #2842's own gate commits real `sync_jobs` rows, and those rows are only executed inside #2842's own measurement window if the runner the scenario is measuring is actually enabled - `guard_runner_state` asserts the running worker's posture matches what the scenario declares, before any load is applied.

The lane-caps startup line is the same trap from the other side: it prints at `log` level from inside `startRunner` (`sync-job.runner.ts:210-213`), so with the runner disabled the line never prints, and the stale `'fan-out': { total: 1, perScope: 1 }` field initializer at `:103-108` survives untouched by the real resolution logic. A guard parsing that line for the manifest must treat its ABSENCE as a failure, never as "no caps to record" - reading a missing line as "nothing to report" would silently manifest the stale fallback value as though it had been resolved.

`guard_log_level` is the `OL_LOG_BODY_MAX_BYTES` half of the same "the log level lied about what actually got logged" class of trap: `allegro-http-client.ts:559-560` and `prestashop-webservice.client.ts:701-702` log full response bodies at `error`, not `debug`, so setting the worker's log level to `log` does not suppress them, and the variable defaults to `0` = uncapped (`format-body-for-log.ts:30`). The guard is fatal on an unset or zero value.

### Stand assertions, not stand provisioning

Provisioning the stand is **#2854's**: a dedicated `lab` compose project (`docker-compose.lab.yml`, applied under `COMPOSE_PROJECT_NAME=lab`), migration-built schema (`NODE_ENV=production` on api, worker and `migrate`), `OL_BOOTSTRAP_ADMIN_PASSWORD` and `OPENLINKER_CREDENTIALS_ENCRYPTION_KEY` supplied so an unattended run cannot fail at boot or at login, a postgres `command:` override that carries `shared_preload_libraries=pg_stat_statements,auto_explain`, and the `--scale worker=3` multi-replica capability with at least one replica carrying the `maintenance` role.

`stand-up.sh` / `stand-down.sh` in this child do NOT build that stack. They run against the already-provisioned `lab` stand and do two things: a jobs-only worker check (`OL_WORKER_ROLE=jobs` on the measuring replicas), scheduler-disabled and demo-mode-off checks, a seed prefix so cleanup can match on the prefix alone (the `PERFBASE-` discipline) - and they ASSERT the stand's shape rather than trusting it. Specifically: that the schema is migration-built (no table exists that only `synchronize` would create), that migration `apps/api/src/migrations/1833000000000-add-destination-categories-table.ts:29,73-76`'s `pg_trgm` extension and `gin_trgm_ops` GIN index on `destination_categories.searchText` are present, that `Σ(OL_DB_POOL_MAX × processes) < max_connections`, and that the resolved lane caps read back from the runner's own startup line are what the manifest expects. #2854's own acceptance criteria already check that `synchronize` never ran in general (no table only `synchronize` would create); this guard is the narrower one #2843's read-path scenario specifically depends on. `apps/api/test/integration/fulfillment-work-migration-parity.int-spec.ts` is reusable for the diff mechanics only.

### `pg_stat_statements` and `auto_explain`, split at the preload boundary

Neither is configured anywhere in the tree. The `shared_preload_libraries=pg_stat_statements,auto_explain` setting needs a Postgres restart, so it lands in #2854's `docker-compose.lab.yml` as a postgres `command:` override, not here. What is left for `stand-up.sh` is the one-off `CREATE EXTENSION IF NOT EXISTS pg_stat_statements;` against the already-preloaded database - idempotent, because `run-all`'s unattended path cannot have a manual runbook step for it - plus asserting `SHOW shared_preload_libraries` already lists both modules rather than assuming #2854's overlay is in effect. #2843 depends on both being live before its window opens.

### The observer records a resource footprint

Nothing else in the programme produces one, and a separately-filed child that writes #1136 needs it. `sample_queue` therefore also samples `docker stats` (CPU and RAM per container) and `pg_database_size` into `timeseries.csv` on the same tick. Cheap, and it is the only place the data can come from.

### The sampler must not enter its own results

`sample_queue`'s `GET /v1/connections/:id/sync-status` call runs the connection-sync-status aggregate over a 7-day `createdAt` window (`libs/core/src/sync/domain/types/connection-sync-status.types.ts:79`) against the very system under test - and every statement it issues lands in `pg_stat_statements`, which is #2843's instrument. The sampler's own query text is therefore excluded from any "top queries by total time" report and recorded as an explicit exclusion rather than silently omitted. Separately, `docker stats --no-stream` alone costs roughly a second to enumerate eleven containers, so a nominal 1 Hz tick either drifts or overlaps under load; `timeseries.csv` carries a `dt` (tick-drift) column recording the actual interval between samples rather than assuming it was exactly 1 s.

## Classification

**Type**: DX
**File(s)**: `perf/openlinker-throughput/lib.sh`, `stand-up.sh`, `stand-down.sh`, `README.md` (new). `docker-compose.lab.yml` and `perf/openlinker-throughput/preflight.sh` are **#2854's**, not this child's - `lib.sh` calls `preflight.sh`, it does not own it.

## Dependencies

**#2854** - provisions the `lab` stand this child asserts against and consumes.

## Assumptions

- Runs against the `lab` stack #2854 provisions (`docker compose -p lab -f docker-compose.yml -f docker-compose.demo.yml -f docker-compose.lab.yml ...`), never bare `docker-compose.demo.yml` - sharing the `openlinker` project name and its data volumes with demo means the first `down -v` destroys demo data. The multi-replica capability is #2854's `--scale worker=3`, not a separate overlay this child owns.
- `lib.sh` parses with `jq`, departing from the existing `perf/prestashop-baseline` scripts' `python3 -c`. #2854's own `preflight.sh` currently lists `jq` as not needed on the strength of that precedent; this child's choice of `jq` means `preflight.sh` must be amended to check for it too, or a missing `jq` surfaces as a failure mid-run rather than a named message at the start.
- Bash + `psql` via `docker exec` + `jq`, with the one exception of the Postgres extension preload, which is #2854's.
- The stand work is smaller than it first reads (compose, migrations, the preload and the credential variables are all #2854's); the guard work is larger - no version surface, a deferral post-guard, a requeue post-guard, a runner-state guard, a log-level guard, and a `max_connections` pre-flight.

## Acceptance Criteria

- [ ] `lib.sh` exists and is the only implementation of login, each guard, the manifest writer, the queue sampler and the verdict writer
- [ ] Every guard is fatal and runs **before** the measurement window opens
- [ ] `guard_build` fails when pointed at a stack running different code than the working tree, by comparing the `org.opencontainers.image.revision` image `LABEL` against `git rev-parse HEAD` - verified deliberately, not assumed
- [ ] `guard_connection_budget` refuses a run whose `Σ(OL_DB_POOL_MAX × processes)` reaches Postgres `max_connections`, and both numbers appear in the manifest
- [ ] `guard_demo_mode_off` and `guard_scheduler_off` both run; the scheduler guard captures the `operational_settings` cadence row, not only the `OL_*_CRON` variables
- [ ] `guard_runner_state` asserts the running worker's `WORKER_RUNNER_ENABLED` posture matches the scenario's declaration, and treats an absent lane-caps startup line as a failure rather than as "no caps to record"
- [ ] `guard_log_level` refuses a run with `OL_LOG_BODY_MAX_BYTES` unset or `0`
- [ ] `post_guard_deferrals` discards a run in which any job was deferred or `SUM("deferredTotalMs")` moved - the attempts post-guard alone cannot see a penalty-free requeue
- [ ] `post_guard_requeues` discards a run in which `StuckJobRecoveryService` requeued any job in the window - neither the attempts nor the deferrals post-guard can see that path
- [ ] `reset_between_repeats` clears both the scenario connections' `connection_cursors` row(s) and their `jobdedup:*` Redis keys between repeats, so a repeated run cannot silently dedupe to `n = 0` while every other guard still passes
- [ ] `post_guard_destination_creates` discards a run in which any `order_records` row in the window carries a `syncStatus` entry with `status: 'failed'`, or lacks `syncedAt` on the scenario's declared destination - the only condition that can otherwise pass every existing guard while zero orders reach a destination
- [ ] `post_guard_limiter_degraded` discards a run whose worker log contains the Redis rate-limiter's degraded-mode message
- [ ] `guard_perf_max_attempts` is fatal, pre-flight, and either resolves a low `maxAttempts` for perf-enqueued job types or the drain loop's max-wait timeout path marks a stalled row `dead` and records its id, so one unseeded row cannot silently reserve ~30 h and starve a later flow's drain wait under an unrelated guard failure
- [ ] A settle period separates the last passing guard from `window_start`, so run 1 does not open on a hot CPU and a build-filled page cache
- [ ] The drain loop carries a max-wait timeout and reports deferred and requeued jobs explicitly, so one hung scenario cannot stall a campaign
- [ ] `manifest.json` is written before any load and is never hand-edited
- [ ] The results contract is fixed and documented: `manifest.json`, `summary.json`, `timeseries.csv`, `verdict.txt`, the raw k6 JSON output, and a dump of the window's `sync_jobs` rows - `verdict.txt` has a documented schema, since `--resume` parses it as a machine contract
- [ ] `timeseries.csv` carries per-container CPU/RAM, `pg_database_size` and a `dt` tick-drift column alongside the queue samples, and the sampler's own `sync-status` queries are recorded as excluded from any `pg_stat_statements` "top queries" report
- [ ] `stand-up.sh` ASSERTS - never builds - that the `lab` stand's schema is migration-built (`pg_trgm` plus the `gin_trgm_ops` index present, and no table exists that only `synchronize` would create)
- [ ] `stand-up.sh` issues `CREATE EXTENSION IF NOT EXISTS pg_stat_statements;` (idempotent) and asserts `SHOW shared_preload_libraries` already includes `pg_stat_statements` and `auto_explain` - the preload itself is #2854's
- [ ] `stand-down.sh` cleans up matching on the seed prefix alone, so it cannot touch demo data
- [ ] README documents every environment variable, every guard, and what makes a run invalid
- [ ] No architecture boundary violations - the harness imports no `@openlinker/core` internals

norbert-kulus-blockydevs:
Closes #2841

## What this is

The shared library every measurement flow in epic #2840 runs on: `perf/openlinker-throughput/lib.sh`.

Eight flows need the same five things before any of them can claim a number — a stand that is the build under test, a runner whose lane caps are what the run assumes, a queue that is drained, a sampler that records rather than infers, and a verdict that is allowed to say `DISCARDED`. Written once per flow those become eight slightly different answers to one question, and the first disagreement between two of them surfaces as a number nobody can defend.

## What it contains

| Piece | What it does |
|---|---|
| 9 pre-flight guards | Refuse to start a run whose premises are false |
| 5 post-guards + verdict | `VALID` / `DISCARDED` written to disk, with the reason |
| Sampler + window | One sampling shape, so two flows' numbers are comparable |
| `drain_wait`, `reset_between_repeats` | Repeat N does not measure repeat N−1's backlog |
| `manifest_write`, `compute_agreement` | Same report shape for every flow |
| Transport helpers | `ps_sql` / `pg_sql` / `wc_wp` / `ol_api` / `ol_login` / `redis_cli` |

**The guards fail closed, and that is the point.** `guard_build` compares the running image's `org.opencontainers.image.revision` label against `git rev-parse HEAD`; an image with **no** label fails too, because "the label is missing" and "the label matches" must never read the same. `guard_runner_state` parses the worker's own startup line for its lane caps and treats an **absent** line as failure for the same reason.

**A discarded run is not a slow run.** A run that hit its rate limiter, saturated a lane, or drifted mid-window measured something other than what it set out to measure. The verdict says so rather than letting it average in.

## `bootstrap.sh` refactor

Pure extraction. `bootstrap.sh` (#2860) had its own copies of `log` / `die` / `ps_sql` / `pg_sql` / `ol_api` / `redis_cli`; those are deleted and the shared ones used instead.

Verified by a live regression run against the demo-fresh stand — byte-identical output to the pre-refactor script:

```
found: 1, created: 0, gaps: 7
```

## The one product-code change

`Dockerfile` gains `ARG OL_GIT_SHA` + `LABEL org.opencontainers.image.revision` on the `production` and `worker` stages.

This is the smallest thing that makes `guard_build` possible. `apps/api` ships no version or build-info surface, and an image digest is **not** content-reproducible — a clean rebuild of identical source can land a different digest, and a layer-cache hit can reproduce a *stale* digest without rebuilding anything. Neither could tell a run it was measuring the wrong build.

**The `ARG`/`LABEL` pair sits at the END of each stage on purpose.** The sha changes on every commit, and an `ARG` whose value changes invalidates its own layer and every layer after it — declared any earlier it would cold-build `pnpm install` on every CI run. Omitted entirely it defaults to `unknown`, which correctly *fails* the guard rather than silently passing.

`docker build --check --target worker` reports no warnings.

## Tests

`lib-test.sh` — 63 assertions, all passing, with no docker, no postgres and no network. It covers the guards' fail-closed behaviour directly, because a guard that silently stopped firing is worse than no guard at all.

```
=== 63 passed, 0 failed ===
```

## Gate

| Check | Result |
|---|---|
| `bash -n` on all three scripts | pass |
| `bash lib-test.sh` | 63 passed, 0 failed |
| `pnpm check:invariants` | exit 0 |
| `docker build --check --target worker` | no warnings |
| `pnpm lint` (eslint half) | **could not run in this worktree** — dependencies are not installed (`eslint: not found`). The diff contains no JS/TS, so eslint has nothing to evaluate; CI covers it. |

The library is exercised by its unit suite and by one live `bootstrap.sh` regression run. **No measurement flow uses it yet** — the guards' behaviour against a genuinely saturated lane, a genuinely throttled run, or a genuinely stale image has been asserted at the function level, not observed on a real run. #2842 (F3) is the first flow to put it under load.


piotrswierzy (commented):
Strong design work — the guard rationale is genuinely good, and every product-code string the harness greps for (`sync-job.runner.ts:211-212`, `redis-rate-limiter.adapter.ts:563`, `database.module.ts:40`, `sync-job.orm-entity.ts:108`) checks out against real source. That's the failure mode that kills scraping harnesses and it was avoided deliberately. Placement is right too: no `package.json`, not a `pnpm-workspace.yaml` glob, no TS — so no workspace-dependency edge and no `check-jest-integration-mappers` entry needed.

My concern is that the 63 assertions cover guard *argument handling*, while the five functions that actually touch a live stand are untested — and three of them fail open, which is the defect class this library exists to prevent.

**Blocking**

1. **`lib.sh:712` — `reset_between_repeats` cannot delete a single key.** `redis_cli` is `docker exec -i` (no `-t`), so redis-cli is already in raw mode; `--no-raw` forces the *quoted, numbered* form, and SCAN then prints `1) "0"` / `2) 1) "key1"`. `head -1` yields `1) "0"`, not `0`, so the loop never terminates and the next SCAN goes out with a malformed cursor, while `DEL $keys` receives `2) 1) "key1"`-shaped tokens and deletes nothing — though `n` counts them as deleted. The function's own header says a failure here makes "the second and third repeat's every enqueue silently no-op", which is exactly this. Drop `--no-raw`, and add a `redis_cli` fake — `lib-test.sh:12` claims one exists and there isn't one.

2. **`lib.sh:525 + :570-575`, `README.md:225` — the manifest declares an exclusion for a query the sampler never issues.** `sample_queue` makes no HTTP call at all; it fires five direct `pg_sql` queries per tick (queued/running/dead/deferred/`pg_database_size`) at the very Postgres under measurement. So `excludedPgStatStatementsQueries` names a phantom `GET /v1/connections/:id/sync-status` while the real observer load is undeclared and unexcluded — a number published without the conditions bounding it, and #2843 will read the sampler's own load as system load. Either list the five real queries (stable, greppable) or route the sample through the endpoint the manifest claims.

3. **`lib.sh:575-579` — every row of `timeseries.csv` is malformed CSV.** `docker_stats_json` is a JSON array containing commas, emitted as the last `%s` with quotes doubled (`sed 's/"/""/g'`) but never wrapped in quotes — and doubling is only correct *inside* a quoted field. Unwrapped, the embedded commas split into extra columns, so the 8-column header never matches the row width. Wrap it, and add a test that parses a row back and asserts 8 fields.

**Important**

4. `lib.sh:831` — `post_guard_limiter_degraded` fails open on any `docker logs` error: `2>&1` folds docker's error into the grep input and `grep -c … || true` turns "the read failed" into `n=0` → `ok`, so a run whose limiter genuinely degraded publishes as `VALID`. (Also `--since` is documented as RFC3339 or a relative duration; the `@1700000000` form is journalctl syntax.) This is the reading `guard_build` and `guard_runner_state` explicitly refuse, so the inconsistency is internal. The test's fake `docker logs` ignores `--since` entirely — a test that passes against the broken implementation.
5. `lib.sh:674` — `drain_wait` echoes `timed_out` and returns 1. Every caller inherits `set -euo pipefail` from line 23, so `x="$(drain_wait …)"` aborts before reading the value — and the timeout branch is precisely the one that must survive to write `DISCARDED`. Either return 0 and let the token carry the outcome, or document the `|| true` requirement.
6. `lib.sh:389-397` vs `:505` — `guard_log_level` reads the value into a `local val` and never assigns `MANIFEST_LOG_BODY_MAX_BYTES`, so `olLogBodyMaxBytes` is `"unknown"` in every manifest despite README:200 listing it among the recorded conditions. It has the value in hand; assign the global and assert it, as `guard_connection_budget`'s globals already are.
7. **The suite doesn't cover the run-time path, and nothing runs it.** `post_guard_requeues` is named in a section header (`lib-test.sh:256`) but never called; `reset_between_repeats`, `drain_wait`, `sample_queue`, `enqueue_perf_job`, `window_start`/`window_stop` have zero assertions — three of the four defects above live in exactly those. The `pg_sql` fake also answers any `COUNT(*)` from one `FAKE_PG[count]` key, so `run_post_guards` can't tell which guard fired. And `bash lib-test.sh` appears in no workflow; please wire it into the lint/test job — needing no docker, postgres or network is its selling point.
8. `lib.sh:340-355` — `guard_build` compares the image label against `git rev-parse HEAD` without checking the tree is clean, so an operator with uncommitted edits gets `ok` for exactly the stale-image case the guard exists to catch. `git status --porcelain` and die, or record `dirty` in the manifest.

**Suggestions**

- `window_start` has a settle but no warm-up: `SETTLE_SECS` is an idle sleep that does nothing for a cold V8 JIT, an unfilled TypeORM pool or a cold Postgres buffer cache, so run 1 systematically under-reports against r2/r3 — which is also what `compute_agreement`'s pairing quietly assumes.
- `window_start` records `WINDOW_START_EPOCH` only, but four post-guards take `window_start_iso` — eight flows get eight chances to derive it differently, or from a different clock. Set `WINDOW_START_ISO` alongside it.
- `grep -oP` (`:373`) is GNU-only and `require_tools` (`:421`) has no call site anywhere in the diff.
- Sourcing has three unopted side effects (`set -euo pipefail` on the sourcer, sourcing a possibly-stale `stand-ids.env` before `bootstrap.sh` regenerates it, and `mkdir -p RESULTS_ROOT` on every `--help`); `results/` also isn't in `.gitignore`.

The Dockerfile `ARG`-at-end reasoning is correct and non-obvious, and re-declaring in `worker` rather than inheriting is right given `guard_build` checks the two images independently. `ol_api` capturing and checking HTTP status separately, and the `${VAR:-{}}` brace-scanning trap being diagnosed with a reproducer, are both good.

Happy to re-review quickly once 1–3 land.

piotrswierzy (changes requested):
# Code Review: shared perf harness library (#2841)

**Scope**: `perf/openlinker-throughput/lib.sh` (new, 908), `lib-test.sh` (new, 407), `README.md` (new, 276), `bootstrap.sh` (refactor, −72), `Dockerfile` (+37). Reviewed against `perf-programme-2840`.

---

## 1️⃣ Summary

**Approach**: ✅ Correct, and the premise is right. Eight flows re-implementing login, drain-wait, guards and the results shape is how `perf/prestashop-baseline` ended up with a runner that silently failed to seed — this library exists to make that unrepeatable, and the fail-closed guard design (absent label ≠ matching label; absent lane-caps line ≠ "no caps to record") is the correct reading of that lesson.

**The one product-code change is well-judged.** `ARG OL_GIT_SHA` + `LABEL org.opencontainers.image.revision` is genuinely the smallest thing that makes `guard_build` possible: I confirmed `apps/api` ships no version surface, and the digest argument in the Dockerfile comment is correct on both halves (a clean rebuild can move the digest; a cache hit can reproduce a stale one). Placing the `ARG` at the **end** of each stage is right and the reasoning is stated where the temptation to move it lives. Defaulting to `unknown` so an unpassed build-arg *fails* the guard rather than passing it is the correct direction.

**Concern**: three defects, all in code the 63-assertion suite does not reach, and all of the same class this library was written to prevent — a fact that reads as recorded but is not.

---

## 2️⃣ Verification performed

Every code citation in `lib.sh` checked against the tree; all correct:

| Citation | Verdict |
|---|---|
| `EnqueueSyncJobDto` does not accept `maxAttempts` | ✅ No such field — so `enqueue_perf_job`'s post-hoc `UPDATE` really is the only route |
| `sync-job.orm-entity.ts:69` entity default is 10 | ✅ `@Column({ default: 10 }) maxAttempts` at :69 |
| Worker logs `Starting sync job runner loop (… lane caps: …)` | ✅ `sync-job.runner.ts:211-212`; the `lane caps: \K.*(?=\))` parse matches the emitted shape |
| `format-body-for-log.ts` treats unset/empty/0/negative/non-numeric as **uncapped** | ✅ Verbatim — `parseMaxChars` returns 0 for all of those and `MAX_CHARS === 0` short-circuits |
| #2279 moved the scheduler out of `apps/api`, so only worker posture matters | ✅ Correct, and checking the api's `OL_SCHEDULER_ENABLED` would indeed be checking a variable nothing reads |
| `jobdedup:*` 7-day TTL invalidates repeats 2 and 3 of an 8 h campaign | ✅ `IDEMPOTENCY_TTL = 7 * 24 * 60 * 60` |
| `docker build --check --target worker` | Trusted from the PR body; the diff contains no obvious warning-triggering construct |

---

## 3️⃣ Issues

### 🔴 BLOCKING 1 — `reset_between_repeats` passes `--no-raw` to `SCAN`, which breaks the cursor loop
`perf/openlinker-throughput/lib.sh` — the `reset_between_repeats` SCAN loop

```bash
batch="$(redis_cli --no-raw SCAN "$cursor" MATCH "jobdedup:*:${conn_id}:*" COUNT 1000)"
cursor="$(printf '%s' "$batch" | head -1)"
...
[ "$cursor" != "0" ] || break
```

`redis-cli --no-raw` *forces formatted output even when stdout is not a tty* — which is precisely the opposite of what this parse needs. Under `docker exec -i` (no tty) redis-cli already defaults to raw, one value per line, which is the shape the code assumes; `--no-raw` overrides that back to the numbered form:

```
1) "0"
2) 1) "jobdedup:…"
```

so `head -1` yields `1) "0"`, never `0`, `[ "$cursor" != "0" ]` is always true, and **the loop never terminates** — while `redis_cli DEL $keys` is fed `2) 1) "jobdedup:…"` fragments and deletes nothing. Two consequences, both bad in the direction this function exists to prevent: an unattended overnight campaign (#2845) hangs in cleanup, and if it were somehow interrupted the dedupe reservations survive, so repeats 2 and 3 silently no-op every enqueue — the exact failure the function's own docblock describes.

Fix: drop `--no-raw`. It is worth a one-line assertion in `lib-test.sh` against a stubbed `redis_cli` emitting the raw two-line shape, since nothing in the suite touches this function today.

### 🔴 BLOCKING 2 — `olLogBodyMaxBytes` is permanently the string `"unknown"` in every manifest
`lib.sh` — `guard_log_level` / `manifest_write`

`manifest_write` reads `${MANIFEST_LOG_BODY_MAX_BYTES:-unknown}`, and **nothing anywhere in `lib.sh` ever assigns `MANIFEST_LOG_BODY_MAX_BYTES`** (I grepped: line 505 is the only occurrence). `guard_log_level` reads the real per-container value into a local `val` and discards it.

So every manifest this library will ever write records `"olLogBodyMaxBytes": "unknown"` — a field that exists to record an environmental fact, sitting next to fields that are correctly populated, reading as though the value could not be resolved when it demonstrably was, one line earlier. This is the same failure mode as `guard_build`'s absent-vs-matching label, one layer over: *"not recorded"* and *"recorded"* must not produce the same output. Note the suite asserts `guard_log_level`'s **behaviour** (3 cases) but never that the value reaches the manifest, which is exactly why this survived — worth adding the assertion alongside the fix, in the shape the existing `pool.budget recorded` / `gitSha recorded` assertions already take.

Fix: `MANIFEST_LOG_BODY_MAX_BYTES="$val"` in `guard_log_level`. (Note it loops over api + every worker, so decide deliberately whether the manifest records the last one or a per-container map — a single scalar for a multi-container fact is its own small honesty question.)

### 🔴 BLOCKING 3 — `timeseries.csv`'s `docker_stats_json` column is emitted unquoted, so every row is malformed CSV
`lib.sh` — `sample_queue`

```bash
printf '%s,%s,%s,%s,%s,%s,%s,%s\n' … "$(printf '%s' "$stats" | sed 's/"/""/g')"
```

The `sed 's/"/""/g'` is the RFC 4180 escape for a quote **inside a quoted field** — but the field is never wrapped in quotes, and the JSON it holds is comma-rich. Simulated against a realistic one-container payload:

```
2026-01-01,1,[{""name"":""lab-api"",""cpu"":""3%"",""mem"":""1MiB / 2GiB""}]
```

The header declares 8 columns; a real 11-container tick will split into roughly 40. Any CSV reader — pandas, `csv.reader`, a spreadsheet — mis-parses every row, and because the doubling *looks* like escaping was handled, this is likely to be discovered only when the first flow tries to plot its own timeseries. Fix: wrap the field, i.e. `\"%s\"` in the format string with the existing doubling retained. Cheap to assert in the suite by round-tripping one row through `python3 -c 'import csv,sys; print(len(next(csv.reader(sys.stdin))))'` and expecting 8.

---

### 🟡 IMPORTANT 4 — `drain_wait`'s documented contract is unreachable under `set -e`
`lib.sh` — `drain_wait`

> Echoes one of: `drained | timed_out`

but the timeout branch is `echo "timed_out"; return 1`, and the file runs under `set -euo pipefail`. A caller doing the thing the docblock invites —

```bash
result="$(drain_wait "$ids")"
```

— takes the exit status of the command substitution and **dies before it can inspect the string**. So `timed_out` is only observable by a caller that already knows to write `|| true`, which the docblock does not say. Either document the required call shape, or `return 0` on timeout and let the echoed value be the whole signal (the function has already marked the stuck rows dead and set `DRAIN_DEAD_IDS`, so the information is not lost). As written, the two documented return values are not equally reachable, and the unreachable one is the failure case.

### 🟡 IMPORTANT 5 — `snapshot_jobs_before` creates a real table in the product database
`lib.sh` — `_perf_sync_jobs_snapshot`

`CREATE TABLE _perf_sync_jobs_snapshot AS SELECT …` writes a harness artefact into the OpenLinker schema. `post_guard_requeues` does drop it, and `snapshot_jobs_before` drops-if-exists first, so the happy path is clean — but a scenario that dies between the two (or a `post_guard_requeues` that is never reached because an earlier guard aborted the campaign) leaves it behind, where it will be picked up by anything counting tables. Given that the sibling PR #2919 is in the middle of establishing *"79 tables, counted"* as a load-bearing figure, a stray perf table in the same database is worth avoiding: prefer `CREATE TEMP TABLE` if the session allows it, or at minimum drop it in a trap rather than only on the success path.

### 🟢 SUGGESTION 6 — `compute_agreement`'s comment says "median", the code computes a mean
For n=2 these coincide, so the arithmetic is right and #2845's `publish_if_agreed` gets the number it needs. Worth one word in the comment so a third repeat is never averaged in by someone reading "median" literally.

---

## 4️⃣ Positive observations

- **`guard_pool_recorded` is justified rather than deleted as redundant.** It re-asserts what `guard_connection_budget` already computed, and the comment says exactly why (a regression in call order is caught rather than silently passing). That is the right instinct for a guard library.
- **The `${VAR:-{}}` brace-scanning trap is documented with its own reproduction** (`set -- x; : "${1:-{}}"` yields `x}`) at both call sites. That is a genuinely obscure bash behaviour and the next reader will not re-discover it the hard way.
- **`scheduler_cadence_row` uses a scalar subquery specifically because the bare form returns zero rows on an empty singleton table**, and says so — again the "absent must not read as failed" discipline.
- **`post_guard_limiter_degraded` bounds its `docker logs` read by `--since`/`--until` on the window epochs**, so a previous scenario's degraded episode cannot discard this one. That is the sort of cross-run contamination that would otherwise be attributed to the run under test.
- **The suite asserts the fail-closed direction directly** — *"absent label dies (never treated as 'unknown, skip')"*, *"absent lane-caps line is a failure, never 'nothing to record'"*, *"`OL_LOG_BODY_MAX_BYTES=0` dies (uncapped, not 'no cap wanted')"*. Those three are the assertions that actually matter, and they exist.
- **"What this does not establish" is accurate**: the guards have been asserted at the function level and not observed against a genuinely saturated lane. Saying so is what keeps #2842 honest about what it inherits.

---

## Final Verdict

**Status**: ❌ **REQUEST CHANGES**

The design is right and I'd like to see it land. But this is the file eight flows will build on, and three of its defects are silent-in-the-safe-looking-direction: a cleanup loop that cannot terminate, a manifest field that reports "unknown" for a value it just read, and a results file that looks escaped and is not. All three are small fixes; none changes the design.

**Priority fixes**
1. Drop `--no-raw` from the `SCAN` in `reset_between_repeats` (BLOCKING 1)
2. Assign `MANIFEST_LOG_BODY_MAX_BYTES` in `guard_log_level` (BLOCKING 2)
3. Quote the `docker_stats_json` CSV field (BLOCKING 3)
4. Resolve `drain_wait`'s `return 1` vs its documented echo contract (IMPORTANT 4)
5. Make the snapshot table temp or trap-dropped (IMPORTANT 5)

Since all three blocking items sit in code the suite does not currently reach, please add one assertion per fix — the suite's own standard, and the reason the other guards are trustworthy.

Issue #2862, opened 2026-09-03
[EPIC] Infrastructure — retention and pruning for the append-only tables


norbert-kulus-blockydevs:
Filed off #2840 and #2843. Not a child of the measurement programme: #2843 produces the row count at which this stops being theoretical, and the remedy is separate work.

## Problem / Context

**Roughly twenty tables in this system are append-only and nothing anywhere deletes from them.** The codebase says so in four places rather than leaving it to be discovered:

- `libs/core/src/sync/domain/types/connection-sync-status.types.ts:73`
- `libs/core/src/sync/infrastructure/persistence/repositories/sync-job.repository.ts:599`
- `docs/architecture-overview.md:270`
- `apps/web/src/features/automation/lib/automation.copy.ts:474-485` - *"nothing removes them. There is no prune job anywhere in the tree"*

Verified: **zero** of the scheduler's registered tasks is cleanup-shaped (`apps/worker/src/scheduler/scheduler.service.ts`). Only two raw `DELETE FROM` statements exist in core and neither is age-based (`user.repository.ts:150`, `destination-category.repository.ts:299`); `webhook-delivery.repository.ts:212` has **no production callers**. `DemoAccountCleanupService` deletes `users` only and is demo-gated. The one real retention implementation in the tree is PHP, in the shop module: `apps/prestashop-module/openlinker/classes/OutboxRepository.php:119-144`.

`sync_jobs` is the worst case - it takes sweep children every twenty minutes, and #2590 observed 15 066 accumulated `queued`/`dead` rows over eight days on one connection. `webhook_deliveries`, `automation_runs` and every `*_snapshots` table grow the same way.

Two consequences, one already documented:

1. **#2843 (F5) measures unbounded aggregates over unbounded tables**, and #2590's own report labels its read-path figure *"not exercised at a large history"* because every runner purged the queue (`results-D-2026-08-28.md:96-105` records `sync_jobs` falling 33 537 -> 30 933 mid-campaign).
2. **#2840 states the deferral is not uniform**: it is safe for the lane claim, which the `(status, nextRunAt)` index confines to the live queue, and unsafe for the read path.

## Proposed Solution

A retention policy per table, a mechanism to enforce it, and an operator-facing statement of what is kept and for how long.

[...]

Issue #2843, opened 2026-09-03
[TASK] DX — F5: operator read path at row count


piotrswierzy:
Child of #2840. **Week 1.**

## Problem / Context

The operator read path runs unbounded aggregates over tables with **no retention anywhere in the tree**, and several of its queries are documented in the source as knowingly unindexed:

| Site | What the code says |
|---|---|
| `libs/core/src/orders/.../order-record.repository.ts:373-375` | *"No GIN index on `syncStatus` today - full scan is acceptable at v1 scale"* - without naming the scale |
| `order-record.repository.ts:1575-1577` | *"the predicate is a `CASE` over four columns, so it is NON-SARGABLE - no index can serve it"*, and `findMany` ends in `getManyAndCount()`, evaluating it twice per request |
| `libs/core/src/identifier-mapping/.../identifier-mapping.repository.ts:102-116,127-133` | paged read ordered by `externalId`; *"That order is NOT indexed for this partition and the cost is accepted"* |
| `libs/core/src/inventory/.../inventory.repository.ts:611-620` | *"full scans of `inventory_items` … no index that helps a four-column grouping"*, deliberately **no `WHERE` clause** |
| `libs/core/src/listings/.../offer-mapping.repository.ts:429-438` | six ORed `ILIKE '%…%'` **plus** a correlated `EXISTS (SELECT 1 FROM jsonb_each_text(pv."attributes") …)` |
| `libs/core/src/listings/.../destination-category.repository.ts:146` | `LIKE '%…%'` backed by a `gin_trgm_ops` index; the rationale, including that it *"degrades to a sequential scan"* without `pg_trgm`, is on the column at `destination-category.orm-entity.ts:75-79` |

`inventory.repository.ts:627` `findDuplicatePositions` is an operator-run diagnostic rather than a page on the operator read path, and is measured as such.

[...]

Issue #2656, opened 2026-08-28
[TASK] Integration — spike: can the shops serve keyset pagination instead of offset


norbert-kulus-blockydevs:
## Problem / Context

Every bounded sweep in the repo pages by **offset**. The cursor is a count of rows already read (`{cycleId}:{offset}`, `apps/worker/src/sync/bounded-sweep.ts`), and the adapter is asked for "500 products, skipping the first 99 000".

An offset describes a position in a list **that moves**. Delete one product before the current offset and everything after it shifts one place left: the sweep re-reads one row it already handled and **skips one it has never seen**, silently. Insert a product and the same thing happens in reverse.

That is not hypothetical at this catalogue size. Measured in #2644 on 100 000 products, a full cycle takes **2.8 days for the catalogue sweep and 41.7 days for the deletion audit**. A shop does not hold still for six weeks.

**The architecture already treats this as a known hazard.** [ADR-048](../blob/main/docs/architecture/adrs/048-incremental-catalog-replication.md) decision 2 states that absence must never be the signal for the deletion pass, precisely because "a cycle spanning many ticks can miss a LIVE row after a mid-cycle delete shifts the offsets - and staling on that inference would zero that product's offers on every marketplace via #1689". The mitigation shipped instead: `master.product.reconcile` concludes nothing from its own enumeration and re-checks every product individually, letting `MasterProductNotFoundError` be the only proof of a deletion.

That mitigation works. It exists **only because the pagination underneath it is unstable**. Keyset pagination removes the cause rather than compensating for the symptom.

Two supporting facts from the code, both verified:

- `PrestashopProductMasterAdapter.listExternalIds` (`libs/integrations/prestashop/src/infrastructure/adapters/prestashop-product-master.adapter.ts:533`) sends only `display=[id]`, `limit` and `offset` - **no `sort`**. #2593 fixed the sort on the *batch product read* (`sort: ['id_ASC']`, same file `:986`), but the id enumeration that drives the sweep and the audit was left unsorted, so page boundaries depend on whatever order MySQL happens to return.
- `WooCommerceProductMasterAdapter.listExternalIds` sends no `orderby` either.

### What this is NOT about

Deep offsets being slow was the expected second argument. **#2644 measured it and it does not hold at this size**, so it is recorded here rather than used:

| | offset 0 | offset 99 000 |
|---|---:|---:|
| requests per tick | 29 | 30 |
| tick elapsed | 46 s | 46 s |
[...]

Issue #2590, opened 2026-08-27
[EPIC] Integration/PrestaShop — close #2489 with a measurement, not a module


norbert-kulus-blockydevs:
## Problem / Context

#2489 asks for a dedicated PrestaShop module so that synchronisation stops putting "significant load" on the shop. On 2026-08-27 we measured the current integration on a live stack (10 000 products x 3 variants, counted in PrestaShop's own Apache access log, median of three runs, plus a control run on untouched code from a fresh container).

Two of the three premises behind the request did not survive the measurement:

| Premise in #2489 | Measured |
|---|---|
| Sync puts significant load on the shop | **p95 under a full catalogue sweep is 0.989x idle.** The shop does not slow down. |
| Bulk reads require a shop-side module | **3 requests, 0.38 s, 1.33 MB for 100 fully-hydrated products** over the plain Webservice. `PrestashopProductMasterAdapter.getProducts` (`display=full` + batched identifier mapping) already exists and its only caller is itself. |
| (not in the issue) | **The queue grows ~145 jobs/h faster than it drains.** The 15 066 backlogged jobs found on the demo stack were several days of normal operation, not an incident. |

What did survive is the one true sentence in the issue: there are too many requests. Two days of adapter work took requests per SKU from **7.96 to 3.97** and the 100-product tick from **977 s to 471 s**, reproducing within 0.4% against the control run.

Full write-up: https://claude.ai/code/artifact/9ebb554a-43be-42b8-ad68-30adca11ac20

## Proposed Solution

Close all ten acceptance criteria of #2489 **without building a PHP module**, and file everything the investigation surfaced along the way.

The only shop-side PHP in this epic is a conditional field on the order-import controller that already exists, already has its own secret and signature. The twelve endpoints of the original design are not built.

## Classification

**Type**: Integration / CORE
**Layer**: Infrastructure / Application / Interface
**File(s)**: `libs/integrations/prestashop/**`, `libs/core/src/{inventory,orders,sync}/**`, `apps/worker/src/sync/**`, `apps/prestashop-module/openlinker/**`

## Scope of impact — the issue's checkboxes are wrong, and here is the list

#2489 ticked only *Integration Adapter — PrestaShop* and *Documentation*. That is not true even for the no-module variant. This epic contains:

- **two migrations** — buyer tax id on the order contract, and a duration column on `sync_jobs`
- **one lane change under ADR-050** — inventory propagation out of the single-slot `fan-out` lane
[...]

PR #2919, opened 2026-09-05
docs(retention): inventory the append-only tables, and the sync_jobs consequence (#2862)

norbert-kulus-blockydevs:
Part of #2862. **Docs only** - no prune job, no migration, nothing deletes.

## Why this stops short of a cron

The issue's own Dependencies section says #2843 (F5) produces the row count and the growth rate, and #2843 has not run. So acceptance criterion 1 cannot be fully satisfied today, and a registered cron that deletes production data must not land on an unmeasured guess. This delivers everything that is not blocked: the inventory, a window per table, the `sync_jobs` correctness reasoning, the operator page, and a mechanism precise enough to implement.

## What landed

- `docs/operations/data-retention.md` - the inventory, the proposed windows, what an operator loses, and what they can do today
- `docs/architecture/adrs/072-append-only-table-retention.md` - the mechanism and four rejected alternatives. **Held, not proposed as a direction**: it carries `Blocked on: #2843` and a leading note saying every window in it is a judgement and that #2843's growth rate is the one input that turns a judgement into a decision. The status line stays `Proposed` because the README's taxonomy has exactly four values and none of them means *held pending a measurement*; `Proposed` literally reads "under discussion, not yet adopted", which is true, and the blocker line says the rest. The index row carries the same qualifier.
- One bullet in `docs/architecture-overview.md` § Sync Manager, plus the ADR index row

## The count is 79, not "roughly twenty"

| Class | Count | Grows with | Anything delete? |
|---|---|---|---|
| A - append-only, time-unbounded | 21 | elapsed time and job/event cadence | no |
| B - append-only, business-record grain | 18 | orders, returns, shipments, invoices | no |
| C - bounded by catalogue or connection count | 13 | catalogue size, not time | upsert-keyed, or self-pruning |
| D - operator configuration | 27 | operator actions | yes, by intent |

The estimate is right for class A. The honest answer to *how many grow without bound and nothing deletes from* is **39** (A + B) - about double - because the business-record tier is equally never-deleted. It is exempt because it is legally or operationally load-bearing, which is a different reason and is stated per table rather than assumed.

Class C is the reassuring part and was worth establishing: it is already bounded. `destination_categories` self-prunes on a `syncedAt` watermark, `allegro_category_cache` evicts on TTL, and the rest are upsert-keyed, so a row count there tracks the catalogue rather than the calendar.

Every claim in the issue was verified rather than taken on trust: `webhook-delivery.repository.ts:217` genuinely has no production caller (a port declaration, one comment, one test mock), and none of the **34** registered scheduler tasks - 19 core plus 15 plugin-side - is cleanup-shaped.

## The window vocabulary

A closed set - **7 d, 30 d, 90 d, 365 d, 2 y**, or **never** - so two tables can be compared and a new one has to argue for its placement rather than inventing a number.

**7 d is stated as what it is.** It is exactly where the Redis `jobdedup:{key}` mark expires, so it is `sync_jobs`' hard *floor* and not a margin, and that table is explicitly never placed there. The two tables that do carry 7 d - `password_reset_tokens`, `email_confirmation_tokens` - have no interaction with an idempotency key of any kind, so the floor does not apply to them and the shared number is not a shared meaning.

**2 y is a dispute horizon, not a fiscal one.** `order_changes`, `order_holds`, terminal `routing_decisions` and `fulfillment_holds` move up from 365 d: what was agreed, why an order was held and which holder was chosen are what a buyer or carrier dispute turns on, and those reach back further than a year. `fulfillment_holds` moves with `order_holds` deliberately - they are read beside each other on the order timeline, and splitting the grains would let an order's hold survive while its work's hold did not.

Two placements deliberately did **not** move:

- **`reservation_shortfall_episodes` stays at 365 d.** Only OPEN episodes are read by any surface (`listOpenByOrderRecordId`), so a closed one is operational history rather than dispute evidence.
- **The fiscal records stay at `never`**, and this is the one I want read carefully. Placing `invoice_records` and `fiscal_registration_records` on 2 y would assert that two years is *enough* - an invented legal requirement in the permissive direction, and the more dangerous of the two directions. Nothing in this tree establishes a statutory horizon: no migration, no constant, no comment names one. What would settle it is a jurisdiction decision by the operator, and its shape is then a **per-install setting they justify**, never a shipped default.

`tax_rate_journal`'s `never` is re-argued rather than restated: the latest row per `(product, variant, connection)` is **live state** served to the operator product surface by `getLatestPerConnection`, so an age window would delete the current rate for any product whose rate has not changed inside it. That is a live-data hazard, not a retention trade-off.

## The `sync_jobs` consequence, per window

`idempotencyKey` is globally unique with no predicate and no expiry (`sync-job.orm-entity.ts:62`). The second dedup layer, the Redis `jobdedup:{key}` mark, carries a **7-day TTL**.

- **Under 7 days**: the Redis mark still stands, so a repeat enqueue never reaches Postgres. Pruning here changes nothing about dedup *and* blinds the connection health panel, whose historical half reads a fixed 7-day window (`BACKLOG_HISTORY_WINDOW_MS`). **7 days is a hard floor.**
- **At or past 7 days**: the Postgres row is the sole guard, and deleting it fully re-arms the key.

On the webhook path that is not theoretical. `WebhookJobGateRepository` inserts the job **before** the delivery row:

```
webhook-job-gate.repository.ts:50   insertOrFindJob(...)          <- sync_jobs, ON CONFLICT DO NOTHING
webhook-job-gate.repository.ts:52   INSERT webhook_deliveries ... <- ON CONFLICT DO NOTHING
```

So a redelivery against a pruned `sync_jobs` row **creates and runs a new job**, while the delivery insert still conflicts and the source is answered with the ordinary idempotent 202. Retaining `webhook_deliveries` does not prevent it.

**90 days is nonetheless defensible**, because every re-run that is harmful rather than merely wasteful is caught by a second durable guard in its own context's table - `UQ_invoice_records_connection_idempotency`, `UQ_fiscal_registration_records_connection_idempotency`, `UQ_refund_records_order_idempotency`, `UQ_return_line_events_line_seq`, the `fulfillment_progress_claims` PK, `UNIQUE (orderId) WHERE state = 'live'`. **That is a standing condition, not a one-time check**: giving any of those tables a window of its own re-opens the question.

Two exclusions generalise beyond `sync_jobs`. No `ON DELETE CASCADE` parent may be pruned - `fulfillment_works` has **five** cascading children and silently takes `fulfillment_progress_claims` (permanent replay memory) and `fulfillment_work_rejections` (the re-sourcing exclusion set) with it. And no table whose uniqueness *is* an at-most-once claim may be pruned at all.

## Two corrections to figures already in the tree

- **The retry ladder tops out at about 4.3 hours, not "around two days"** as `connection-sync-status.types.ts:76` states. `30s x 2^(n-1)` over 9 backoffs sums to 15 330 s, and no `maxAttempts` anywhere in this tree exceeds 10 (the two overrides are `= 3`; no request DTO accepts one), so `RETRY_MAX_DELAY_SECONDS` is never reached. With the 24-hour penalty-free deferral budget the true maximum queued lifetime is about **29 hours**. The 7-day `BACKLOG_HISTORY_WINDOW_MS` bound is safely conservative either way, so this corrects the stated reason rather than the bound.
- **`sync_jobs` has no index for a global age prune.** It carries `(connectionId, createdAt)` but no standalone `createdAt`, so `DELETE ... WHERE "createdAt" < $1` sequential-scans the table this whole policy exists because it is large. Recorded as a prerequisite, not shipped.

## Honesty

Every figure carries a **measured / derived / judgement** label and its source, and a *"what this did not establish"* section names what is missing: growth rates, the row count at which each table becomes a problem, the date an install arrives there, disk, and whether pruning would help the slow reads at all (both #2843 aggregates are already `WHERE`-bounded to 7 days, so it may change their cost very little - a reason to measure before registering a cron, not a reason to skip the policy). #2590's figures are cited as *that report's* measurement with its date, never as this page's.

## Proposed child issues

**No issue has been created.** Each needs approval.

1. **`[TASK] Infrastructure - the retention prune mechanism, dry-run only`** - Ship the per-table registry, the `retention.prune` job, handler and scheduler task, with `OL_RETENTION_PRUNE_ENABLED` default `false` and `OL_RETENTION_PRUNE_DRY_RUN` default `true`. It counts what each window would delete and deletes nothing, so #2843's figure becomes actionable without a destructive first run. Frontier-as-query with no cursor, bounded `DELETE ... WHERE id IN (SELECT ... LIMIT n FOR UPDATE SKIP LOCKED)`, reusing the sweep primitives but not the offset machinery, in the `inventory.provenance.backfill` shape.
2. **`[TASK] Infrastructure - size the retention windows against #2843's growth rate`** - Once F5 has run, replace every judgement window with one derived from a measured growth rate, record the date the figure was taken, and update the operator page. Blocked on #2843.
3. **`[TASK] Infrastructure - index sync_jobs for an age-based prune`** - Either a partial index on the terminal statuses or a per-connection iteration on the existing `(connectionId, createdAt)`. Prerequisite for enabling the prune on `sync_jobs` at all; decide it against #2656 if that lands first.
4. **`[TASK] Infrastructure - enable retention, one table class at a time`** - Turn the cron on for real, starting with the classes carrying no correctness consequence (`password_reset_tokens`, `email_confirmation_tokens`, `allegro_quantity_commands`), then `webhook_deliveries` and `automation_runs`, and `sync_jobs` last. Depends on 1-3.
5. **`[TASK] Automation - ship the 90-day retention sentence once the prune exists`** - `automation.copy.ts:485` renders *"Every automation run recorded so far is listed here"* instead of the spec's *"Runs older than 90 days are removed"*, with a docblock saying the sentence changes when the prune lands. This is that change, and it must not ship before the prune.
6. **`[TASK] Users - subject erasure for the customer projections`** - `customer_projections` and `customer_address_projections` hold PII gated by `OL_STORE_PII`. They need erasure on request, which is a different trigger and a different mechanism from age-based retention, and ADR-072 explicitly leaves it out.

## Acceptance criteria

| AC | State |
|---|---|
| Every append-only table enumerated with its **growth rate** and window | **Partial** - all 79 enumerated and classified with a window or an explicit "kept forever, and why"; **growth rate blocked on #2843** and labelled unmeasured per table rather than guessed |
| The `sync_jobs.idempotencyKey` re-enqueue consequence stated per window | **Done** |
| One mechanism, registered where the other periodic tasks live, env switch + documented default | **Designed, deliberately not registered** - child 1 |
| Pruning bounded and resumable in the shape the sweeps use | **Designed** - child 1 |
| Operator-facing page under `docs/operations/` | **Done** |
| #2843's figure cited as the trigger, with the date it was taken | **Blocked on #2843**. #2590's figures are cited with their date as the interim evidence. |

## Gate

`pnpm check:invariants` exits 0. No `.ts` / `.tsx` / `.js` / `.mjs` file changed, so `lint` and `type-check` are not applicable.


piotrswierzy (commented):
Genuinely strong research — I independently verified the 79-table count, the 34 scheduler tasks (correctly excluding the two non-registration `taskId:` hits — a naive `grep -c` gives 21), the 7-day `jobdedup` TTL, the absence of a standalone `createdAt` index on `sync_jobs`, the job-before-delivery insert ordering, and the retry-ladder arithmetic (30 × (2⁰…2⁸) = 15 330 s, cap unreachable at ≤10 attempts). All correct.

**The inventory itself is complete.** I enumerated every `@Entity('…')` on the base branch and reconciled name-by-name: 21 Class A + 18 Class B listed explicitly, all 13 Class C named, and the residual is exactly 27 — every one genuinely operator config or an upsert-keyed singleton. I spot-checked the two most suspicious residuals (`sales_document_country_acknowledgments` is country-PK'd and its docblock says the upsert "never accumulates a history"; the invoice numbering tables are config rows). No append-only table is missing and nothing is listed as append-only that isn't.

Not shipping a cron against an unmeasured guess is the right call, and reading #2862's Dependencies section that way rather than writing code anyway is the harder one.

Two things to fix before merge.

**1. "The one retention implementation that does exist … It is PHP, in the shop module" is false.** `apps/api/src/auth/demo-account-cleanup.service.ts` is an age-based (`OL_DEMO_ACCOUNT_RETENTION_HOURS`, default 24 h), hourly, lock-serialised, row-deleting retention implementation in TypeScript — and `docs/architecture-overview.md` § 24 names it explicitly as "the repo's one Postgres row-deletion retention precedent to mirror". #2862 knew about it; the shipped page drops it and asserts uniqueness in a heading, so an operator concludes nothing in TS ever deletes on a schedule. It's also the closer *mechanism* precedent than the PHP one — terminal-scoped, operator-configurable, replica-serialised deletion is exactly what child issue 1 needs. Reword to "the one retention implementation for an append-only table" and cite it.

**2. ADR-072's exclusion rule contradicts the operator page's own window table.** The ADR states "no table whose uniqueness is an at-most-once claim may enter it". `uq_webhook_deliveries_event_key` is exactly that — architecture-overview.md:2116 calls it the durable replay protection that holds "even if Redis is wiped" — yet the page places `webhook_deliveries` on **90 days**. The page half-acknowledges it ("the durable replay gate for events older than the window") but the ADR's rule is absolute, so whoever builds child issue 1 resolves this arbitrarily. Either carve out the exception explicitly (the re-run is idempotent *and* `sync_jobs` is retained at least as long) or drop the window.

Also worth picking up:

- **`bulk_batch_advancements` is missing from the at-most-once exclusion list.** Its own docblock says the composite PK on `(bulkBatchId, offerCreationRecordId)` "makes the INSERT-ON-CONFLICT-DO-NOTHING path atomic" — the same shape as the four that are listed. It currently sits in Class B under "kept forever" with no stated reason, i.e. kept by luck rather than by rule.
- **Class B promises a per-table reason and delivers 9 of 18.** Nothing is said for `order_records`, `order_line_items`, `shipments`, `fulfillment_work_lines`, `offer_creation_records`, `listing_creation_records`, the two `bulk_*` tables, or `destination_address_mappings`. `offer_creation_records` is the sharpest omission — one row per OL-initiated creation *attempt*, so it grows with publish activity like a Class-A table, not with business records. Either deliver the per-table row or soften the promise.
- **`docs/architecture/adrs/README.md:169` still says "Allocate 072 for the next new ADR."** The PR adds the ADR-072 table row but leaves the note pointing at 072 — and that note is the only guard, as it says itself (#2082). Next author collides. → 073. (072 is the right number for this ADR; nothing else claims it.)
- `webhook-delivery.repository.ts:217` → `:212` for `deleteByEventKey` (#2862 cites it correctly). In a provenance table whose whole value is verifiability, an off-by-5 costs the reader the lookup; several other entries give a bare basename too.
- Class D is the only class not enumerated by name. The arithmetic is right and the five singletons check out, but a collapsed `<details>` list closes the one place a miss could hide.
- The `tax_rate_journal` `never` is argued slightly too strongly — "an age window would delete the current rate for any product whose rate hasn't changed inside it" is true of a *naive* window only; "keep the latest per `(product, variant, connection)`, prune older" is available and is what `getLatestPerConnection` already implies. The conclusion is fine; name the alternative you're declining.
- No CI ran on this PR (`get_check_runs` is empty), so the `check:invariants` pass is self-reported. Fine for docs-only, just noting it isn't verified by the repo.

Best thing in the document: the 7-day-as-floor-not-margin distinction. Noticing that `password_reset_tokens` and `email_confirmation_tokens` share the number 7 with `sync_jobs`' hard floor while sharing none of its meaning, and saying so explicitly, pre-empts exactly the copy-paste an implementer would otherwise make. Refusing to put the fiscal records on 2 y because that would assert two years is *enough* is right and in the right direction; and the "What this did not establish" section conceding that both #2843 aggregates are already `WHERE`-bounded to 7 days — so pruning may not help those reads at all — argues against the author's own work and is stated anyway.

piotrswierzy (approved):
# Code Review: append-only retention inventory (#2862)

**Scope**: `docs/operations/data-retention.md` (new, 350 lines), `docs/architecture/adrs/072-append-only-table-retention.md` (new, 158), `docs/architecture/adrs/README.md` (index row), `docs/architecture-overview.md` (one § Sync Manager bullet). Docs only — no `.ts`/`.tsx`/`.mjs` touched, so `lint`/`type-check` are genuinely N/A.

---

## 1️⃣ Summary

**Approach**: ✅ Correct. This is an inventory and a held design, not a shipped policy, and it is honest about which is which. The decision to stop short of registering a cron because #2843 has not produced the growth rate is the right call and is argued rather than asserted — `docs/migrations.md § Best Practices` and the ADR README both treat irreversible operations as needing a measured basis, and deletion is the most irreversible one in the tree.

**Documentation conflicts**: ✅ None. The new § Sync Manager bullet sits correctly beside the #2615 and #2613/#2617 bullets it depends on, and the ADR follows the README's template (Context / Decision / Alternatives / Consequences / References).

**The most important thing about this PR**: I spot-checked every load-bearing citation against the tree rather than taking any on trust, and **every one held**. That is unusual and worth saying explicitly, because a retention page whose figures are wrong is worse than no page.

---

## 2️⃣ Verification performed

| Claim | Verdict |
|---|---|
| **79 tables** in the schema | ✅ 79 distinct `@Entity('…')` table names (80 declarations — `shipments` appears twice, see SUGGESTION 1). Class counts 21+18+13+27 sum to 79. |
| `sync_jobs.idempotencyKey` unique, no predicate, no expiry, `sync-job.orm-entity.ts:62` | ✅ `@Column({ type: 'varchar', unique: true })` at :62, field at :63 |
| Redis `jobdedup:{key}` TTL is 7 days, `redis-streams-job-enqueue.service.ts:24` | ✅ `IDEMPOTENCY_TTL = 7 * 24 * 60 * 60` at line 24 |
| `WebhookJobGateRepository` inserts the job **before** the delivery row | ✅ Confirmed — the source's own comment reads `// Job FIRST: …`, and both statements are `ON CONFLICT DO NOTHING` as described |
| **34** registered scheduler tasks, 19 core + 15 plugin, none cleanup-shaped | ✅ Exactly. `scheduler.service.ts` has 21 `taskId:` occurrences, of which one is the interface declaration (:123) and one a pass-through (:761) — 19 real; plugin side is 15. |
| `webhook-delivery.repository.ts:217` has no production caller | ✅ `deleteByEventKey`'s only references are the port declaration, one comment, one test mock |
| `fulfillment_works` has **five** `ON DELETE CASCADE` children | ✅ Exactly five, across migrations 1864/1865/1867/1870000003000 |
| `return_line_events` has **no** FK to `returns` | ✅ No `REFERENCES` for that table in any migration |
| `sync_jobs` has **no standalone `createdAt`** index | ✅ Indexes are `(status,nextRunAt)`, `(lockedAt)`, `(connectionId,createdAt)`, `(connectionId,jobType,status,updatedAt)`. A global `WHERE "createdAt" < $1` prune would indeed seq-scan. |
| Retry ladder ≈ **4.3 h**, not "around two days" | ✅ `Σ 30·2ⁱ, i=0..8` = 15 330 s = 4.258 h. `RETRY_MAX_DELAY_SECONDS` is 6 h and the largest ladder step is 30·2⁸ = 2.13 h, so the cap is genuinely unreachable. The `connection-sync-status.types.ts:76` comment does say "around two days". |
| Every `maxAttempts` ≤ 10 | ✅ Default 10; every override is `= 3` |
| `automation.copy.ts` renders *"Every automation run recorded so far is listed here"* | ✅ Verbatim at `retentionFooter`, and its docblock already anticipates exactly this change |
| All six "second durable guard" indexes exist | ✅ `UQ_invoice_records_connection_idempotency`, `UQ_fiscal_registration_records_connection_idempotency`, `UQ_refund_records_order_idempotency`, `UQ_return_line_events_line_seq`, the `fulfillment_progress_claims` PK, and `UNIQUE (orderId) … WHERE "state" = 'live'` |
| `destination_categories` self-prunes on a `syncedAt` watermark | ✅ `deleteStaleBelow` |
| `reservation_shortfall_episodes` read only via open episodes | ✅ `listOpenByOrderRecordId` / `…Ids` are the only reads |
| `tax_rate_journal` latest row is live state via `getLatestPerConnection` | ✅ Exists and is served to the product surface |
| `OutboxRepository.php` — constants ~110-145, `runRetention()` at 1013, batch 1000 × 10 | ✅ All four |
| `fulfillment_progress_claims` index pre-created for "an age-based retention sweep, which does not exist yet" | ✅ Verbatim comment in migration `1865000000000` |

---

## 3️⃣ Findings

### 🟢 SUGGESTION 1 — the stated counting method reproduces 80, not 79
`docs/operations/data-retention.md` § How many tables this is about

> **79 tables** exist in the schema (`@Entity(...)` declarations across `libs/` and `apps/`, counted 2026-09-05).

The **count of tables is right**, but a reader re-running the stated method gets 80: `@Entity('shipments')` appears twice — once in `libs/core/src/shipping/…/shipment.orm-entity.ts` and once inside the text of `apps/api/src/migrations/1869000000000-add-shipment-fulfillment-work-id.ts`. Since the whole page's credibility rests on its "measured" labels, say *distinct table names* rather than *declarations*, so the next person recounting concludes "same answer" instead of "the figure has drifted by one".

### 🟢 SUGGESTION 2 — the ADR index row carries a qualifier outside the README's own four-value taxonomy
`docs/architecture/adrs/README.md` — `| Proposed (held pending #2843) |`

The PR body argues this well, and I agree with the reasoning: `Proposed` = *"under discussion, not yet adopted"* is literally true, and none of the four values means *held pending a measurement*. But the README says the index is *"One pointer per section, identical format every time"*, and the ADR body already carries a dedicated `**Blocked on**: #2843` line plus the leading held-note. The qualifier in the index is therefore redundant with the place that owns it, at the cost of the index no longer being mechanically filterable by status. Consider dropping the parenthetical from the index row only.

### 🟢 SUGGESTION 3 — PR-body-only: "the two overrides are `= 3`"
There are **four** `= 3` retry-budget constants (`RUNNER_RETRY_BUDGET` in `offer-status-poll.service.ts` and again in `marketplace-offer-refresh-snapshot.handler.ts`, `MANUAL_REGISTRATION_RETRY_BUDGET`, `AUTO_ISSUE_RETRY_BUDGET`). This does not appear in the shipped docs — `data-retention.md` says only *"Every `maxAttempts` in the tree is <= 10"*, which is correct — so nothing needs changing in the diff. Noted so the figure is not carried forward into a child issue.

---

## 4️⃣ Positive observations

- **The 7 d floor is disambiguated rather than shared.** Placing `password_reset_tokens` and `email_confirmation_tokens` on the same number as `sync_jobs`' floor, and then explicitly stating that *the coincidence of number is not a coincidence of meaning*, is the kind of thing that stops a future reader from "harmonising" the two into one constant.
- **Refusing 2 y for the fiscal records is the right direction of error.** Asserting that two years is *enough* would be an invented legal requirement in the permissive direction; `never` plus "this is a per-install setting the operator justifies" is the only defensible default, and it matches how ADR-026 keeps national specifics out of `libs/core`.
- **`tax_rate_journal`'s `never` is re-argued, not restated.** Naming it a *live-data hazard* rather than a retention trade-off is exactly right, and it is the one entry on the page where an age window would silently corrupt current state rather than lose history.
- **The "what this did not establish" section names the uncomfortable one**: that both #2843 aggregates are already `WHERE`-bounded to 7 days, so pruning may barely move the read cost. Volunteering the fact that most weakens the case for the work is what makes the rest of the page trustworthy.
- **The two exclusions generalise correctly.** "No `ON DELETE CASCADE` parent" and "no table whose uniqueness *is* an at-most-once claim" are stated as rules over the registry rather than as per-table notes, which is what makes the child-1 registry implementable without re-deriving them.
- **The standing-condition framing on the six second guards.** Stating that giving any of those tables a window of its own re-opens the `sync_jobs` question turns a one-time check into an invariant a future PR has to notice.

---

## Final Verdict

**Status**: ✅ **APPROVED**

Nothing blocking, nothing required. Three optional suggestions, all cosmetic. Ships an inventory and a held ADR with no executable surface, so the merge risk is confined to what a future reader believes — and the figures they would believe check out.

**On the proposed child issues**: 1-6 look correctly scoped, and I'd particularly endorse shipping child 1 (registry + dry run, default off) before #2843 rather than after — a dry run is how #2843's figure becomes actionable, so waiting for the measurement to build the counter inverts the dependency.

Issue #2865, opened 2026-09-03
[TASK] Integration — Erli's 500-unread inbox cliff: the one throughput limit nothing in the programme can reach


norbert-kulus-blockydevs:
Filed off #2840 and #2846, both of which name it "to be filed". Not a child of the measurement programme as scoped: it needs either an Erli arm or an Erli stub, and #2846 has neither.

## Problem / Context

#2846 states the mechanism and the gap together:

> **Erli is deferred, and one throughput cliff goes unmeasured because of it.** Erli's order source reads a server-fixed **&lt;=500-unread inbox with ack-on-next-read** (`libs/integrations/erli/src/infrastructure/adapters/erli-order-source.adapter.ts:12-25`): the adapter never acks the wave it returns, only messages at or below the committed cursor, so **acking is what bounds the unread window below the 500 cap**. Nothing else in the programme can reach that shape, and a burst that pushes unread past 500 is therefore unmeasured by this epic.

#2840 calls it *"arguably the most interesting throughput cliff in the system"*, and the reason is that it is not a soft degradation. Every other limit in this system is a rate that slows down; this is a **server-side cap on a queue OpenLinker does not control**, where falling behind on acking does not slow ingestion down but stops it.

Verified: **there is no Erli service in any of the three compose files** - zero references. So there is no local way to reach this at all today.

Unmeasured questions: what Erli does when unread exceeds 500 (drop, refuse, block); what OpenLinker observes when it does; how far behind the cursor can fall before it matters; and whether the ack-on-next-read shape can lose an order rather than merely delay it.

## Proposed Solution

First a decision, then a measurement.

**The decision**: a real Erli **sandbox** arm, or an Erli stub. They answer different questions. A stub can produce the burst on demand and reach past 500 repeatably, but the cap is a server-side behaviour and a stub's version of it is a guess. A sandbox has the real behaviour and cannot be pushed past the cap on demand. The honest answer may be both: a sandbox probe to learn what the server actually does at the cap, and a stub reproducing that to measure OpenLinker's response repeatably.

**The measurement**, once a driver exists: drive orders in faster than the cursor advances, observe what Erli reports as the unread count approaches 500, and record whether OpenLinker's ingestion degrades, stalls, or loses. The last of those is a correctness finding, not a performance one, and would belong with #1135 rather than here.

## Classification

**Type**: Integration
**Layer**: Infrastructure (adapter) + test double
[...]

Issue #1135, opened 2026-06-22
[FEATURE] Resilience/chaos testing — prove no order loss under Redis outage + horizontal scale


piotrswierzy:
## Source

External evaluation feedback (2026-06-22): "have you done chaos engineering? I'd scale a demo env horizontally and e.g. kill Redis / reset everything, then watch whether everything behaves. In e-commerce nothing can be lost."

## Why

OL is *designed* for at-least-once / nothing-lost (Postgres-authoritative webhook dedup #711, cursor-commit-after-enqueue, at-most-once bulk advancement via `bulk_batch_advancements`). Those are claims we should **prove** under failure, not just assert.

## Proposed scope (scripted experiments)

- Scale workers/API horizontally; assert no duplicate or dropped orders under concurrent processing.
- Kill Redis mid-ingestion; verify recovery with no lost/duplicated orders (exercises the PG-authoritative dedup + Redis inner-gate failure-recovery path in #711).
- Restart/reset between original delivery and a source retry; verify the unique constraint never permanently blocks a legitimate retry.
- Document the invariants each experiment validates and the observed result.

## Acceptance criteria

- [ ] Repeatable chaos scenarios (scriptable; runnable against a scaled local/demo env).
- [ ] Each scenario asserts an explicit invariant (no loss / no dup / clean retry).
- [ ] Findings documented; any gaps filed as follow-ups.

## Related

- #711 (Postgres-authoritative webhook dedup), #1127 (demo env to run against), the k6 load harness issue.

norbert-kulus-blockydevs:
Delivered on `perf-programme-2840` by #2983. Closing manually — that PR's base was the perf epic branch, not `main`, so no `Closes` clause could fire.

Evidence on `perf-programme-2840`: `perf/openlinker-throughput/scenarios/f10-dependency-failure.sh` plus `results-F10-2026-09-08.md` / `results-F10-2026-09-08-epyc32.md` and the per-case verdicts under `results/f10-dependency-failure/`, covering I2 (Redis drops entirely mid-window) and I3 (worker kill).

**Not on `main` yet** — epic #2840 tracks that merge.

Issue #992, opened 2026-06-05
[IMPL] Verification spike: Erli sandbox access + live order/inbox payload (BLOCKING)


piotrswierzy:
## Design source

**Product Design parent:** Part of #978 — see `docs/specs/product-spec-978-erli-marketplace-integration.md`. **Blocking gate for the orders half.** External dependency: Erli sandbox access (support-gated, separate domain).

## Scope

Obtain Erli sandbox access and capture real payloads to resolve the four unknowns that gate the order-mapping work. **No production adapter code** — this is verification.

Resolve and document:

1. **Buyer email / PII shape** — real, Allegro-style masked relay, or absent? (drives identity mode + email normalization)
2. **Exact order resource schema** — line items, totals, addresses, COD flag/status.
3. **Inbox ack / cursor semantics** — confirm "≤500 unread, ack by newest id".
4. **Webhook registration mechanism** — API-provisionable or manual seller-panel only?

## Acceptance criteria

- [ ] Sandbox credentials secured (or a clear note if Erli declines, escalating to maintainer).
- [ ] One real order + one inbox payload captured and documented in the spec / a follow-up note.
- [ ] Each of the 4 unknowns has a recorded resolution that the dependent issues can build on.

## Effort estimate

- [x] **S** — 1–3 days (excluding external wait for sandbox provisioning)

## Dependencies

- **Blocked by:** external — Erli support provisioning a sandbox (request in parallel with the offers half)
- **Blocks:** order mapper, buyer identity, inbound webhooks, OrderSource adapter

## Architecture notes

- Verification only; no adapter code. Migration: no.

## Next step after merge

- [ ] No further action — unblocks the orders mapping issues

norbert-kulus-blockydevs:
## Verification spike resolved ✅

Sandbox access secured and confirmed live (`GET /me` → 200 on `https://sandbox.erli.dev/svc/shop-api`). All four unknowns resolved against the **authoritative OpenAPI contract** (`/svc/shop-api/doc/swagger.json` — sandbox runs the same API). The sandbox currently holds **no orders and no inbox messages** (both endpoints return empty arrays), so a captured live order/inbox *sample* awaits a test order being placed — the OpenAPI schema is the authoritative resolution and is what the dependent issues build on.

**Auth confirmed:** `Authorization: Bearer <id:secret>` (raw key, colon form) — matches the implemented client.

### #1 — Buyer PII / email shape
[...]

PR #2920, opened 2026-09-05
docs(perf): decide how to reach Erli's unread-inbox cliff (#2865)

norbert-kulus-blockydevs:
Adds one markdown document,
`perf/openlinker-throughput/decision-erli-inbox-cliff-2026-09-05.md`. No source file is
touched.

`Part of #2865` rather than `Closes` - three of the six acceptance criteria stay open, and the
document says why rather than inferring an answer.

## The decision

**Both a sandbox probe and a stub, in a fixed order, with a step in front of them that
neither #2865 nor #2840 anticipated: get the inbox contract from Erli in writing.**

| # | Step | For | Blocking |
|---|---|---|---|
| 0 | Ask Erli: is the inbox capped, at what number, what happens to a new event at the cap, does a truncated listing return the oldest or the newest, and does a read-marked message ever come back | The only definitive source | Yes |
| 1 | A scoped read-only sandbox probe (one `GET /inbox`) | The listing's sort order, plus a latency constant for the stub | No |
| 2 | An Erli source stub | The measurement rig - the only one that reaches the burst repeatably and re-runs month over month | No |

The reason in one sentence: **the cliff is a server-side contract rather than an emergent
performance property, so the vendor is the cheapest and only definitive source - and a live
sandbox cannot substitute, because observing what the server does at the cap means first
putting hundreds of unread events into a third party's sandbox, which is load-testing someone
else's system rather than the "characterisation probe" #2840 licensed.**

That is a **correction to #2840**, which lists this cliff as one of exactly three legitimate
live-sandbox uses on the premise that a sandbox reaches it. The document records the
correction openly rather than diverging quietly, and states a fallback for the case where
Erli does not answer (build the stub parameterised over every branch and publish conditional
results, never one branch's result as *the* behaviour).

## Acceptance criteria

| AC | Status |
|---|---|
| 1. A written decision, with the reason | **Satisfied** |
| 2. Cap and ack semantics confirmed, or explicitly recorded as unconfirmed | **Satisfied** - see below |
| 3. What Erli does at the cap, from observation | **Open** - no driver exists that can push unread past the cap |
| 4. What OpenLinker observes at the cap | **Open** - same reason. A loss path is *derived from code* and labelled as a reading, not an observation |
| 5. Order loss raised as a correctness issue linked to #1135 | **Conditionally open** - the candidate path and its destination are named; filing it needs the branch confirmed, and filing a correctness defect on an unverified branch would be exactly the guess this programme refuses |
| 6. Labelled measured / derived / extrapolated | **Satisfied** - the document ends with a label ledger |

## AC 2: three different provenances, routinely quoted as one fact

- **Confirmed against the live platform (#992):** the endpoint shapes, the ObjectId ids, the
  `POST /inbox/mark-read { lastMessageId }` **request form**, the type vocabulary, and the
  absence of a `limit` query param.
- **Not confirmed - the ack EFFECT.** That *a read-marked message never returns from a
  re-read* is the property the whole ack-on-next-read design rests on, and it is **not** in
  `erli-inbox.types.ts`'s verified list. Its only occurrence in the tree is the same prose
  sentence, in the same commit, that carries the 500 figure. The verified list even carries
  weak evidence the other way: the wire item has a `read` boolean that the adapter parses and
  nothing ever filters on. If it is false, "acking is what bounds the unread window" is false.
- **Not confirmed - the 500 cap.** Two files, both prose, both from `3800a3d43`; absent from
  the verified list, from ADR-025 and from both Erli docs. What #992 confirmed is that there
  is **no client-controlled limit parameter**, which is compatible with an uncapped listing, a
  cap at another number, or a cap of 500.

**It was not confirmed here, and a read-only probe could not have confirmed it.** No Erli
connection exists in either local stack (both `connections` tables queried directly); the one
named in the manual-testing walkthrough belonged to a local `ol-demo-full` stack that has
since been torn down. More to the point, reaching the cap requires *creating* hundreds of
unread events, so even with a working key the decisive facts are out of a read-only probe's
reach. The document specifies the one thing such a probe can establish - the listing's sort
order - and says that if the sandbox holds fewer than two unread messages the probe answers
nothing.

## The finding that is not a performance finding

**If Erli's listing truncates newest-first, an order is lost silently.** Derived by reading
code, contingent on a branch nobody has confirmed, and stated prominently rather than buried.

With unread `1..700` and a listing that returns the newest 500, OpenLinker enqueues `201..700`
and commits cursor `700`; ids `1..200` are then filtered out of every later poll by
`newWave = filter(id > fromCursor)`, and the next poll's `markReadUpTo(700)` erases them.
Three things make it undetectable:

- the listing is the only evidence, and the listing is what was truncated;
- core's cursor guard blocks a cursor going *backwards*, and this is a legal forward move. It
  also barely engages for Erli: `compareOrderCursors` recognises a **decimal** counter, an ISO
  instant, a naive wall clock and a wall-clock keyset, so a hex ObjectId answers
  `unrecognised` and core only warns;
- there is no reconciliation channel. `erliOrderPath(id)` is the only orders read in the
  package - no list-orders endpoint - so nothing re-discovers the order, and the webhook fired
  long before the burst.

**It is the cursor advance that loses the order, not the ack**, and the document is explicit
about that because two plausible fixes do not work: acking per id (Erli supports
`{ ids: [...] }`) stops the erasure but leaves the messages permanently filtered out and
consuming cap headroom, and moving the ack before the read fixes only the capacity penalty.
The fix is to refuse to advance the cursor over a listing that may be truncated - stop and
alert, because a stalled feed is recoverable and a lost order is not.

If the branch is confirmed, this belongs on **#1135**, not on a throughput report.

## Three adjacent findings, all derived from code

1. **The ack runs after the read**, so the previous wave still occupies this poll's listing.
   Per-poll capacity for new messages is `cap - |previous wave|`; that map is an involution,
   so it gives the orbit `cap -> 0 -> cap` and an **average** of `cap/2` - roughly halving the
   drain ceiling under the benign branch. Arithmetic over unverified inputs, labelled derived.
2. **There is no saturation signal.** Nothing logs the unread depth or the listing size. The
   one adjacent trace, `orders-poll.handler.ts`'s `fetched=N, enqueued=N, committed=N`, counts
   feed items *after* the cursor filter, the type filter and the dedupe.
3. **`markReadUpTo` is best-effort and warn-only**, and acking is what bounds the unread
   window - so a persistently failing ack walks the install into the cliff on its own.

## A blocker for the stub arm that nobody has hit yet

**#2854's "no code change to repoint OL" pattern does not hold for Erli.**
`erli-base-url.policy.ts` requires `https:` **and** a host at or under `erli.pl` / `erli.dev`,
enforced at create/update by `ErliConnectionConfigShapeValidatorAdapter` and again by
`ErliAdapterFactory.resolveBaseUrl` (which throws `ErliConfigException`) as deliberate defence
in depth - it is an SSRF guard, because the base URL carries the static API key. A
compose-network stub at `http://erli-stub:8080` is refused twice. The document picks TLS plus
an `erli.dev` network alias (#2854 already proposes a `wc-tls` nginx service, so the pattern
is in scope) and rejects env-gated allowlist widening. It also notes in passing that the two
enforcement points do **not** share one guard, despite the policy file's docblock claiming
they do.

Creating the connection row is *not* an extra burden: `bootstrap.sh` already creates all four
of its connections via `ol_ensure_connection`, so an Erli arm is one more block.

## Follow-up issues proposed, not created

Per the working agreement these are proposals only - no issue was created.

1. **"Ask Erli for the inbox contract: cap, saturation behaviour, truncation order, read-marked
   semantics"** - a vendor-liaison task, not an engineering one. Get a written answer to the
   five questions in section 7. It blocks any honest stub and is by far the cheapest step;
   until it lands, every branch in section 4 including "there is no cap" stays live.
2. **"Erli source stub for the unread-inbox cliff"** - `GET /inbox` + `POST /inbox/mark-read`
   with real read-marking state, ObjectId ids, a cap that can be set or disabled, and a
   configurable truncation order so every branch can be run. Includes the TLS / `erli.dev`
   alias work. Depends on #2854 and on (1).
3. **"Erli inbox: report the listing size and warn at saturation"** - a small adapter change,
   worth doing whichever branch is real, and the prerequisite for any cursor-advance guard.
   Today nothing logs the inbox depth, so the cliff is invisible until it has been crossed.
   Cheapest item in the set, and depends on nothing.

Whether the order-loss path becomes an issue on #1135 depends on (1) or a saturation run
confirming the branch, so it is deliberately not proposed as an issue yet.

## Two docblocks that say something the code does not do

Recorded in the document (section 6) because neither has an issue and it is the only place
they will survive. Neither is caused by this work.

- `erli-order-source.adapter.ts:18` says the ack happens "at the START of each call"; the GET
  is at `:194` and the ack at `:228`. That ordering is what halves the drain ceiling, so
  anyone reasoning from the docblock gets the rate wrong by a factor of two.
- `erli-base-url.policy.ts:34` calls itself "The single guard both the config-shape validator
  and the adapter factory call". Only the factory calls it; the validator re-implements the
  https half. Both halves are enforced today, in two implementations rather than one - the
  exact drift the docblock says is impossible, in SSRF-adjacent code a stub arm has to touch.

## Review

The document was fact-checked against source by a second pass, which found six wrong claims
and several imprecise ones - all corrected before this PR was opened. The two that mattered
most: it had granted "confirmed" status to the read-marked-never-returns property on the same
prose sentence it was busy discrediting for the cap, and it had proposed per-id acking as the
structural fix for a loss the cursor advance actually causes.

## Gate

`pnpm check:invariants` - exit 0. No `.ts`/`.tsx`/`.js`/`.mjs` file changed, so lint and
type-check were not run.

Part of #2865.

### DIFF
--- a/perf/openlinker-throughput/decision-erli-inbox-cliff-2026-09-05.md
+++ b/perf/openlinker-throughput/decision-erli-inbox-cliff-2026-09-05.md
@@ -1,0 +1,574 @@
+# Decision - how to reach Erli's unread-inbox cliff (#2865)
+
+Written 2026-09-05 against `main` at `c1e4090ee`.
+
+This is a **decision document, not a results report.** Nothing here was measured. Every
+statement below is either a quotation from the repository, a fact confirmed by an earlier
+live spike and recorded in the tree, or a consequence derived by reading code. Each is
+labelled. Nothing was observed against the live Erli platform in the course of writing it,
+and no figure here may be quoted as a measurement.
+
+It answers acceptance criteria 1, 2 and 6 of #2865. Criteria 3, 4 and 5 need a driver that can
+push Erli's unread inbox past its cap; no such driver exists in this repository or in any
+open issue, and section 9 says so rather than inferring an answer.
+
+**Read section 2 first if you read nothing else.** The single most consequential thing found
+here is not the decision, it is that **the 500 cap and the "a read-marked message never
+returns" property are both unverified prose from one commit** - present in neither the #992
+spike's recorded findings, nor ADR-025, nor the setup guide, nor the runbook. What #992
+actually confirmed is the different and much weaker claim that *there is no client-controlled
+`limit` parameter*. Those two have been quoted interchangeably ever since, including by #2840,
+#2846 and #2865 itself. Almost every question this document could not answer traces back to
+that, and it changes what anyone should do next: the first move is not to build a rig, it is
+to find out whether the thing being modelled is real.
+
+---
+
+## 1. The decision
+
+**Both, in a fixed order, with a third step in front of them that neither the issue nor the
+epic anticipated.**
+
+| # | Step | What it is for | Blocking? |
+|---|---|---|---|
+| 0 | **Get the inbox contract from Erli in writing** | Establish whether the unread inbox is capped, at what number, what the server does with a new event at the cap, and whether a truncated listing returns the oldest or the newest messages. | **Yes.** Steps 1 and 2 are both unsound without it. |
+| 1 | **A scoped, read-only sandbox probe** | Corroborate the vendor's answer on the one axis a read-only probe can reach - the listing's sort order - and take a latency baseline for the stub. A characterisation probe, never a throughput run. | No, but cheap. |
+| 2 | **An Erli source stub, as the measurement rig** | Produce the burst repeatably, reach past the cap on demand, and measure what OpenLinker does at it. The only rig that can be re-run month over month, which is what the programme's comparability requires. | No. |
+
+**The one-sentence reason: the cliff is a documented server-side contract rather than an
+emergent performance property, so the cheapest and only definitive source is the vendor -
+and a live sandbox cannot substitute for it, because observing what the server does at the
+cap means first putting hundreds of unread events into a third party's sandbox, which is
+load-testing someone else's system rather than the "characterisation probe" #2840 licensed.**
+
+### This corrects #2840, and the correction should be recorded there
+
+#2840 lists this cliff as one of exactly three behaviours that need a live sandbox:
+
+> **Live sandboxes are needed for exactly three behaviours, none of which is F1 or F2**, and
+> each is a characterisation probe rather than a throughput run: [...] and Erli's server-fixed
+> <= 500-unread inbox with ack-on-next-read (`erli-order-source.adapter.ts:12-25`), which is
+> arguably the most interesting throughput cliff in the system, is unreachable by any other
+> means, and is out of scope for #2846 - to be filed separately rather than silently dropped.
+
+(Emphasis as in the original; the elision drops the two Allegro items.)
+
+The premise holds in the sense it was meant: a stub can produce the *shape* of the cliff but
+can never observe the server, so nothing but the live platform can say what Erli really does.
+The implicit second premise - that a sandbox therefore *reaches* it - does not. A sandbox
+reaches it only if somebody first drives its unread count past whatever the cap is, and there
+are only two ways to do that: place hundreds of real sandbox orders, or find some other event
+that lands in the inbox and emit hundreds of those. Both are sustained writes into
+infrastructure Erli operates, and #2840's own framing ("a characterisation probe rather than a
+throughput run", "load is not driven through a marketplace UI") does not authorise either. The
+honest reading is that the sandbox is necessary but not sufficient, and that the missing piece
+is a contract statement from the vendor.
+
+### What each rig may claim
+
+| Rig | Can establish | Cannot establish |
+|---|---|---|
+| Vendor contract (step 0) | Whether the cap exists, its value, the saturation behaviour, the truncation order | What OpenLinker does when it hits any of that |
+| Read-only sandbox probe (step 1) | The listing's sort order at small unread counts; per-request latency for the stub's cost model; that the endpoint shapes still match what #992 recorded | Anything at or near the cap. A quiet sandbox has an empty inbox. |
+| Erli source stub (step 2) | Everything about **OpenLinker's** behaviour at the cap - drain rate, cursor movement, whether an order is lost, what an operator can see | Anything about **Erli's** behaviour at the cap. A stub asserts the cap; it never observes one. |
+| Live saturation run | All of it | Requires Erli's explicit agreement; out of scope until they give it |
+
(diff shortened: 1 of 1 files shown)


piotrswierzy (approved):
Approving. I fact-checked every claim this document makes about OpenLinker's own code against `main` — the line numbers (`:194` GET / `:228` ack), the two docblock discrepancies, `compareOrderCursors`' four shapes and the once-per-process warn, the `bootstrap.sh` `ol_ensure_connection` pattern, the base-URL guard's double enforcement via `parseHttpsUrl`, the unused `read` boolean at `:826`, the ADR-025 and runbook quotations, and the #2840 block quote — and all of them hold. The §2 provenance split (ack *request form* verified in #992; ack *effect* and the 500 cap both unverified prose from one commit) is a real and consequential finding, and identifying the cursor advance rather than the ack as the loss mechanism is the kind of error a review pass exists to catch.

Two things to fix before or just after merge, both about where the outputs go rather than whether they are right:

1. **§6(a)'s ack-reorder has no owner.** It is the one actionable, branch-independent production improvement the document derives, the safety argument holds (messages ≤ `fromCursor` are already enqueued-and-committed before the ack), and it appears in neither §7 nor the PR's three proposals. Add it as a fourth proposed issue or fold it into the saturation-logging one — a PR body is not a work queue.

2. **The two wrong docblocks are recorded only under `perf/`.** The document says this is "the only place they are written down", which is the problem rather than the mitigation: the next person to read `erli-order-source.adapter.ts` still gets the drain rate wrong by 2x. Both are one-line fixes. At minimum an issue; `docs/lessons.md` is the designated ledger for exactly this.

Smaller:

- §1 "This corrects #2840" overstates it — the epic's own *Deliberately out of scope* table already says nothing in the programme reaches Erli's path, and never asserted a sandbox reaches the cliff. Reframe as necessary-but-not-sufficient and cite that row as partial corroboration.
- §9's ack-effect gap deserves a pointer from ADR-025, which is where the next architect reads the reconciliation-first posture that design rests on.
- `perf/openlinker-throughput/` has no README, unlike `perf/prestashop-baseline/`; add a line to #2840 or a stub so this is findable by something other than grep.

Nice call refusing to file the #1135 correctness issue on an unverified branch, and listing branch 0 ("there is no cap") first — a stub hard-coded to 500 would make that permanently unobservable.

PR #2921, opened 2026-09-05
perf(dx): Allegro sandbox latency baseline for the two stubbed endpoints

### LINKED ISSUE #2861: [TASK] DX — one-off Allegro sandbox latency baseline for the two stubbed endpoints

norbert-kulus-blockydevs:
Closes #2861.

## Summary

A one-off characterisation probe measuring real per-request latency against the
Allegro **sandbox** for the two endpoints #2856's stub is meant to reproduce:

- `GET /order/events?from={cursor}&limit=100`
- `GET /order/checkout-forms/{id}`

New files:

- `perf/openlinker-throughput/probes/allegro-latency.sh` - standalone probe
  script (no dependency on `lib.sh`/`bootstrap.sh` refactors from the
  still-open #2918, per the epic's own conflict-avoidance note)
- `perf/openlinker-throughput/results-allegro-latency-2026-09-05.md` - the
  full report
- `perf/openlinker-throughput/allegro-latency-samples-2026-09-05.json` - raw
  samples
- `docs/lessons.md` - one new entry recording the incident below

## Result 1 - `GET /order/events` - MEASURED

Harvested from the worker's own debug log for traffic the deployment already
generates every minute (the `allegro-orders-poll` scheduler task) - **zero
added load**.

n=480, sandbox host `api.allegro.pl.allegrosandbox.pl`, 2026-09-05.

**The distribution is two tight clusters, not one number**: 90-338 ms (51.5%
of samples) and 1059-1374 ms (48.3%), with almost nothing between them. One
outlier at 11139 ms is attributed to a logged Redis rate-limiter degradation
on this shared demo stack, not to Allegro, and is kept in the reported set
rather than dropped.

p50=276ms, p95=1194ms. The mean (655ms) falls in the empty gap between the
two clusters and matches no request ever actually observed - it is
**refused** as a hand-off value for exactly that reason. If #2856 needs one
constant, the report hands off **p50 (276ms)**, and separately records that
a single-latency knob cannot express this shape at all - possibly a finding
for #2856 itself (a two-point distribution rather than one constant). The
cause of the split is unestablished; the obvious candidate (cold vs. warm
connection reuse) was tested and ruled out (there is exactly one call per
minute, so every call is equally cold by that theory) and no other mechanism
is offered.

The report separately flags that this sample was harvested from the coldest
possible request pattern (once-a-minute polling) while F1/F2 apply a
sustained ramp with connection reuse, so this figure may over-represent
cold-path latency.

## Result 2 - `GET /order/checkout-forms/{id}` - NOT SAFELY MEASURED, incident disclosed

There is no organic traffic for this endpoint (no order on this connection
has ever had a destination mapping). A smoke test re-enqueued
`marketplace.order.sync` for a pre-existing order on the assumption that
re-sync is a safe no-op - it is only a no-op when a destination mapping
already exists, which none did here, so **the smoke test created a real
duplicate order in the demo stand's PrestaShop** (`ps_orders.id_order=28`,
reference `907b5790-9bbe-11f1-bd08-9328d2ed1733`). No further active samples
were taken once this was discovered.

**That order still exists on the demo stand.** An attempted SQL cleanup was
declined by this environment's own safety controls before it ran, and per
explicit instruction it was not attempted again. The full account, the
cleanup SQL (for an operator to run by hand, or to leave as harmless test
data), and the one incidental real sample (n=1, 1157ms, not reported as a
percentile) are all in the results report. The probe script now carries a
safety gate (`step_check_destination_safety`) that refuses to run this half
by default unless it can prove no active destination connection exists, or
the operator explicitly overrides it after verifying every id is already
destination-mapped.

## Hand-off

Neither #2854 nor #2856 exist on `main` yet (both are open PRs against
`perf/openlinker-throughput/`), so the measured value cannot be wired into a
file that isn't there. The report's "Hand-off" section carries the exact
value (`STUB_PER_REQUEST_LATENCY_MS: '276'`, i.e. p50, with the reasoning
above) and the citation to replace #2854's current placeholder comment with,
for whoever picks up either PR next.

## Acceptance criteria

- [x] p50/p95 recorded for `/order/events`, with sample size, date, sandbox host
- [ ] p50/p95 for `/order/checkout-forms/{id}` - not satisfied (see above); n=1 incidental sample reported instead
- [~] Measured value becomes the stub's default - cannot be done directly since #2854/#2856 don't exist on `main`; exact value + citation handed off in the report
- [x] Report states this is sandbox latency, production unmeasured
- [x] Report states why a real measurement is what makes #2847/#2848 externally meaningful
- [x] Labelled measured/not-measured per endpoint, with a "what this did not establish" section (production latency, concurrency, rate-limit response, checkout-forms as a percentile, cause of the bimodal split, and the cold-poll sampling-pattern limitation)
- [x] No load beyond a sequential probe (order/events: zero added load; checkout-forms: one active request before aborting)


norbert-kulus-blockydevs:
Folded in a measured `GET /order/checkout-forms/{id}` baseline. Real sandbox purchases gave this endpoint organic traffic, so it moves from "not safely measured" to measured, re-harvested independently (not by trusting a hand-off sample file) the same passive way as `/order/events` - zero added load.

**n=118** at harvest instant 2026-09-05T22:56:55Z, **p50=1106ms**, p95=1217ms, mean=1062ms.

Four findings:

- **Unimodal, unlike `/order/events`**: 73.7% of samples land in one 1000-1199ms peak. A real, measured contrast with events' two clusters - it narrows the unknown from the first version of this report down to "a property of the endpoint, not the client/network/connection," since the same client/network/connection produce two clusters on one endpoint and one on the other. No mechanism offered beyond that comparison.
- **p50 here is ~4x `/order/events`' p50 (276ms)** - reported as measured, not explained (a checkout-forms fetch hydrating a whole order vs. an events page returning references is plausible on its face, stated as such, not confirmed).
- **118 samples, 23 distinct orders** (mean 5.13 fetches/order) - an honest request count for latency, not a throughput claim about 118 orders.
- **19 of 23 orders are `recordStatus=source_deleted`, on a live retry ladder** (verified: `attempts=5`, `status=queued`, same `lastError` every time). Each retry re-issues a full checkout-forms fetch for zero forward progress - a real cost finding for the programme, not a claim the retry is wrong (a mapping can legitimately reappear later, which is why it stays retryable).

**The sample is a moving target** - the retry ladder keeps growing it in real time (a harvest two minutes earlier read n=98, p50=1084ms; a harvest after this commit read n=126). The report states this explicitly rather than picking one number and hiding the drift, and cites the specific harvest instant used.

**Script**: gained `probe_checkout_forms_passive`, now preferred over the active/side-effecting path from the previous commit whenever it finds any organic traffic at all - the active path (with its safety gate) is now purely the fallback for a stand with no organic checkout-forms traffic.

**Hand-off revised**: checkout-forms' own measured figure makes a single `STUB_PER_REQUEST_LATENCY_MS` knob a demonstrated mismatch for whichever endpoint it isn't tuned to (not an unmeasured gap being filled in, as the first version said). Recommendation is now a per-endpoint constant: `STUB_ORDER_EVENTS_LATENCY_MS: '276'` / `STUB_CHECKOUT_FORMS_LATENCY_MS: '1106'`.

Everything from the previous version stands unchanged: the incident account (real duplicate PrestaShop order still on the demo stand, cleanup SQL provided, not re-attempted), the safety gate, and `/order/events`' own numbers (n=480, p50=276ms, p95=1194ms).

piotrswierzy (commented):
The measurement discipline here is genuinely good — the status check is present and correctly reasoned ("a fast error is not a fast success", the exact defect ADR-066 correction 1 records), the mean is refused with a real argument, the incident is disclosed rather than buried, and the `lessons.md` entry generalises properly. The `source_deleted` retry-ladder observation is a real finding for the epic. But I can't sign off on the headline result.

**Blocking: the bimodal split is very likely our own Redis rate limiter, not Allegro.**

Re-deriving from the committed samples JSON:

- fast-cluster median **139 ms**, slow-cluster median **1139 ms** — delta of **exactly 1000 ms**
- **373 of 479** adjacent pairs flip clusters — a near-deterministic alternation, not two random populations

The report already names the mechanism, in its own outlier note (L77-88): *"Redis rate-limiter call `checkPace` timed out after **1000ms**"*, and correctly observes the pacing gate sits inside the measured window. The code confirms that applies to **every** sample, not just that one outlier — `allegro-adapter.factory.interface.ts:46` documents `fetchImpl` as "the connection's **rate-limited transport**", and `allegro-http-client.ts:298,381` brackets `fetchImpl` between `startTime` and `duration`.

A 1000.0 ms offset on ~48% of samples, on a stand with a documented 1000 ms limiter timeout, isn't "the cause of the split is unestablished" (L45-51) — it's sitting in the report's own text. Two consequences:

1. `STUB_ORDER_EVENTS_LATENCY_MS: '276'` isn't Allegro latency.
2. The flagship recommendation to #2856 — that the stub needs a **two-point distribution** rather than one constant — would teach it to reproduce our broken Redis, which is the opposite of making #2847/#2848 externally meaningful.

Please either re-measure on a stand with a healthy limiter, or split the sample on the limiter-degradation log lines and hand off the fast cluster's own p50 (139 ms), retracting the two-point recommendation. The events-vs-checkout-forms "unimodal contrast" conclusion (L163-172) needs re-deriving on the same basis — checkout-forms' 1000–1199 ms dominant mode with a 68 ms floor is equally consistent with near-total contamination (68 + 1000 lands in that bucket), so it cannot rule the client out; it may be the same artefact at a higher rate.

Smaller:

- `probes/allegro-latency.sh:373-388` — the **active** checkout-forms sampler takes `ok[-1]` from a window that can span 30 s+ of job polling, with no trace-id group and no connection filter, unlike both passive harvesters (which correctly build `trace_to_conn` and drop foreign samples). On any stand with a second Allegro connection it will silently record another request's latency. Copy the correlation from `probe_checkout_forms_passive`.
- `probes/allegro-latency.sh:62-67` — the header claims the number is "the real network + Allegro-processing time — not OpenLinker's own job overhead". The second clause holds; the first doesn't, since the limiter wait is inside the window. Reword, because that distinction is exactly what the blocking finding turns on.
- The samples JSON carries a `checkout_forms_meta` key the script never writes, under a filename the script never produces (`OUT_JSON` defaults to `allegro-latency-samples.json`) — so it was hand-assembled. For an artefact whose stated purpose is to be cited verbatim rather than re-derived, provenance is the whole value: either emit it from the script or say plainly it's a curated composite.
- The PR body still says checkout-forms is "NOT SAFELY MEASURED … n=1" and hands off a single `STUB_PER_REQUEST_LATENCY_MS: '276'`, while the report says measured, n=118, two per-endpoint constants. The follow-up comment supersedes it, but the body is what the #2856 implementer reads first.
- Report L64-67 mis-transcribes its own percentile formula — `-1` outside the `min`, where both the script (L509) and the cited `storefront-probe.sh:50-51` precedent put it inside. They diverge at p100.
- L29: n=480/n=118 need `SAMPLE_SIZE` and `EVENTS_LOOKBACK_MINUTES` overrides (defaults are 20 and 120). Record the exact invocation so the run is reproducible.

Separately, and not a code finding: `ps_orders.id_order=28` — the duplicate order from the smoke-test incident — is still on the shared demo stand with the cleanup SQL provided but unrun. Worth an operator decision now rather than after it's forgotten.

`step_check_destination_safety` is a real structural fix rather than a comment — queries live state, defaults to refusing, requires an explicit override. Good.

piotrswierzy (approved):
## Summary

Reviewed against the stack base `perf-programme-2840`, not `main`. Four files: a standalone probe script, a results report, a raw sample set, and one `docs/lessons.md` entry. No production code, no core/API/worker change, no schema change — the blast radius is a script an operator runs by hand.

This is a good piece of work, and the part that matters most is the part that went wrong. A smoke test created a real duplicate order in the shared demo stand's PrestaShop, and rather than being quietly dropped it is disclosed in the PR body, in the report, in a `docs/lessons.md` entry with the correct rule generalised out of it, and in a new safety gate in the script itself. The measurement half is equally disciplined: `/order/events` is harvested from traffic the deployment already generates, so it adds **zero** load; the acceptance criteria are marked honestly including one unmet box and one partially-met one; and the bimodal distribution is reported as two clusters with the mean explicitly **refused** as a hand-off value because it "matches no request ever actually observed". That refusal is the right call and the reasoning for it is the strongest paragraph in the report.

Two things I would change before this is done, one of them a genuine gap between what the lesson says was learned and what the gate actually enforces.

---

## 3️⃣ Issues & Risks

### 🟡 IMPORTANT

**1. `step_check_destination_safety` does not check the precondition the incident was about** — `perf/openlinker-throughput/probes/allegro-latency.sh`

The `docs/lessons.md` entry states the rule precisely:

> verify with a query (not an assumption) that the specific rows being replayed are already past every write it could perform, or gate the replay behind an explicit check for that precondition

The gate does neither. It asks a *different* question — "does any other active connection advertise `OrderProcessorManager`?" — and on a yes it refuses unless the operator sets `FORCE_CHECKOUT_FORMS_PROBE=yes`. The precondition that actually determines whether a re-sync writes is per-`externalOrderId`: does *this* id already have a destination mapping. That is checked by nobody. The script header even supplies the exact SQL for it, and then delegates running it to a human.

So on the only stand where the probe is interesting — one with a live destination, which is the stand the incident happened on — the override restores the pre-incident risk with no machine check between the operator and a second duplicate order. The gate is "refuse unless a human promises", which is the shape the lesson says to replace.

The script is already authenticated to the OL API and already parses JSON with `python3`, so this is reachable: resolve each `ORDER_EXTERNAL_IDS` entry to its internal order id and assert a mapping exists on every active `OrderProcessorManager` connection, refusing by id and naming the ones that fail. Then `FORCE_CHECKOUT_FORMS_PROBE` can go away entirely, or degrade to an override on a check that *ran* rather than one that was skipped. As written, the guard and the lesson beside it disagree about what the incident taught.

**2. The duplicate order is still live on a shared stand and is tracked only in a results markdown**

`ps_orders.id_order=28`, reference `907b5790-9bbe-11f1-bd08-9328d2ed1733`, cleanup declined by this environment's safety controls and per instruction not retried. Disclosing it is right; leaving it addressed only in `results-allegro-latency-2026-09-05.md` is not sufficient, because that file is an artifact of a finished measurement and nobody re-reads it. Anyone later measuring order flow on the demo stand meets a real order with no order behind it.

Open a follow-up issue carrying the id, the reference and the cleanup SQL, and link it from both the PR body and the report's incident section — or, if the decision is genuinely to leave it as harmless test data, say so in one line as a **decision** rather than as an unresolved item, so the next reader is not left wondering whether it is still someone's to do.

### 🟢 SUGGESTION

**3. Shell interpolation into Python source strings** — `step_check_destination_safety`, and the job-status poll in `probe_checkout_forms_active`

Both build Python source with double-quoted heredocs so the shell substitutes values into string literals (`mine = '$CONNECTION_ID'`, `key = '$idempotency_key'`). Both values are operator-supplied and low risk here, and `probe_order_events` already does the right thing with a quoted `<<'PY'` plus `sys.argv`. Worth making it uniform, if only so the quoted-heredoc convention the file already establishes is not undermined by two exceptions.

**4. `SAMPLE_SIZE` overloads two meanings**

The comment says it is both the active probe's sample count and the cap on the passive harvest, most-recent-first. Those are different quantities — the passive one is "how much organic history to keep" and could reasonably be much larger, since it costs nothing. Worth splitting if the probe is re-run.

---

## Positive observations

- **The `/order/events` measurement adds no load at all.** Harvesting the worker's own structured debug log and correlating trace ids against the `(connection: <id>)` request line to scope samples to one connection is a better answer than an active probe, and the header says why the OAuth token is deliberately never decrypted by this script — "handling a live OAuth secret outside the system that owns it is exactly the kind of thing a one-off perf script should not be trusted with". That is the right instinct and it is stated rather than assumed.
- **The bimodal result is reported as a shape, not flattened into a number.** Refusing the mean because it falls in the empty gap between the two clusters, handing off p50 instead, *and* recording that a single-latency knob cannot express this distribution at all — possibly a finding for #2856 itself — is exactly the reporting standard this programme is trying to hold. The candidate cause (cold vs. warm connection reuse) being tested and ruled out, with no substitute mechanism offered rather than a plausible-sounding one, is the same discipline.
- **The 11 s outlier is kept, attributed to a logged Redis rate-limiter degradation on the shared stack, and reported rather than dropped.** Dropping it would have tightened p95 and told a nicer story.
- **The sampling-pattern limitation is volunteered**: once-a-minute polling is the coldest possible request pattern, while F1/F2 apply a sustained ramp with connection reuse, so this figure may over-represent cold-path latency. That is the kind of caveat that only ever comes from the author.
- **The status filter is right**: `if status != '200': continue  # a fast error is not a fast success`.
- **The `docs/lessons.md` entry follows the documented format**, names the two services it applies to, and generalises correctly to "any future probe or ops script that replays a marketplace job type for measurement rather than for its intended effect" rather than staying specific to this one endpoint.
- **The hand-off is honest about being un-wireable.** Neither #2854 nor #2856 exists on `main`, so the value plus its citation go in the report rather than into a file that is not there. Correct — the alternative is a change that looks wired and is not.

---

## Final Assessment

A characterisation probe that measures what it can measure without adding load, refuses to report the one endpoint it could not measure safely, and discloses an operational incident in full with the rule generalised into `docs/lessons.md`. No architectural, security or contract concern; nothing here is unsafe to merge.

The one substantive gap is **#1**: the safety gate added in response to the incident checks a proxy for the precondition rather than the precondition itself, and the override reopens the exact hole on the exact stand where it matters. The lesson entry states the correct rule; the code should enforce it.

### Merge Readiness

🔄 **Approve with changes**

### Priority fixes

1. **#1** — make `step_check_destination_safety` verify per-`externalOrderId` destination mapping, as the lesson entry beside it prescribes, instead of gating on "any destination connection exists" plus an operator promise.
2. **#2** — open a follow-up issue for the live duplicate order (`ps_orders.id_order=28`) and link it, or record "leave as test data" as an explicit decision.

Issue #2866, opened 2026-09-03
[TASK] Frontend — F9: Lighthouse CI and bundle budgets


norbert-kulus-blockydevs:
Filed off #2840, whose out-of-scope table names "separate issue" as this work's owner. Deliberately outside the measurement epic: it shares no machinery with any queue or database flow.

## Problem / Context

#2840 excluded frontend performance for a stated reason and recorded the facts so they would not be lost with the epic:

> Lighthouse CI + a bundle analyzer shares no machinery with any of this and is half a day. Folding it in makes this epic look bigger than it is. The known facts [...] are recorded here so they are not lost.

Those facts, each verifiable:

- **About ten requests before content on every authenticated page**: five `limit:1` nav-count probes plus one unpaginated connections read (`apps/web/src/app/hooks/use-nav-counts.ts:31-38` - `useConnectionsQuery` takes no pagination), three command-palette prefetches (`apps/web/src/app/command-palette-provider.tsx:174-180`), and a system-config read (`apps/web/src/app/app-shell.tsx:239`).
- **About nineteen on `/products`**, which additionally carries two N+1 sites on the API side (`apps/api/src/products/http/products.controller.ts:229-231` on the list path, `:355-357` per variant on the detail path).
- **`@tanstack/react-virtual` is present but default-off** in `apps/web/src/shared/ui/data-table.tsx`, so long lists render every row.
- **No `manualChunks`** in the Vite config, so route-level `lazy` (`apps/web/src/app/routes/route-lazy.test.ts` pins the count) is the only code splitting.
- **No size budgets anywhere**, so a dependency added tomorrow costs whatever it costs, silently.

Nothing here is a crisis. The point is that it is currently unmeasured and undefended, so it can only get worse.

## Proposed Solution

Two standing gates plus one pass at the facts above.

- **Lighthouse CI** on a representative authenticated route set, with a committed baseline and a threshold that fails a pull request on regression. An authenticated route needs a session, so the harness is the existing Playwright login (`apps/e2e/src/support/`) rather than an anonymous crawl.
- **A bundle analyzer with size budgets** per entry chunk, failing the build past the budget. The pattern to follow is the repo's own invariant scripts (`scripts/check-*.mjs` under `pnpm check:invariants`), which is how every other structural rule here is enforced.
- **Then act on the recorded facts**: collapse the nav-count probes into one batched read, page the connections read, turn virtualization on where the row count warrants it, and add `manualChunks` for the vendor split.

[...]

Issue #2852, opened 2026-09-03
[TASK] DX — F7: cross-lane starvation, the test that validates ADR-050


piotrswierzy:
Child of #2840. **Week 3.**

## Problem / Context

[ADR-050](https://github.com/openlinker-project/openlinker/blob/main/docs/architecture/adrs/050-workload-isolation-concurrency-lanes.md)'s central claim is that lanes are chosen by **cost of starvation**, and that lanes are capped rather than prioritised: decision 2 reads *"Per-lane concurrency caps, never strict priority. Every lane can always pull."* (`docs/architecture/adrs/050-workload-isolation-concurrency-lanes.md:70`). The operator-facing consequence is stated in `docs/architecture-overview.md:267`: a saturated `bulk` lane *"cannot delay a queued `realtime` or `fiscal` job beyond its own lane's availability"*.

**That claim has never been demonstrated under load.** It is the entire justification for the lane split, for #2594 moving the sweep children out of `realtime`, and for #2609 raising `fan-out` - three shipped changes resting on an untested property.

The scenario is also the most operator-realistic one on the list: an operator clicking "publish 1000 products" is an ordinary Tuesday, and the question *"does my customer's order wait behind it?"* is the one lanes exist to answer.

## Proposed Solution

Saturate the `bulk` lane with a publish wave of ~1000 children, and while it drains, ingest an order and issue a fiscal document. Measure each one's queue wait, and compare against the same job on a control stand.

### Reaching 1000 children takes more than one call

`POST /v1/listings/bulk-create` (`apps/api/src/listings/http/bulk-listing.controller.ts:69` prefix, `@Post()` at `:79`) caps `productIds` at **100** (`@ArrayMaxSize(100)`, `apps/api/src/listings/http/dto/bulk-offer-create.dto.ts:143`). So ~1000 children is either about **10 submits of 100**, or fewer submits against multi-variant products, where #824's sibling expansion produces one child per variant - which is what #2849's seed data is for. Whichever route is taken must be recorded in the manifest, because the two produce different arrival shapes.

Arrival is also **paced by the stream consumer, not by the HTTP call**: children are enqueued via Redis Streams and reach `sync_jobs` only through `JobIntakeConsumer`, which reads `COUNT 10` per `xReadGroup` with `BLOCK 5000` (`apps/worker/src/sync/job-intake.consumer.ts:41,45,46`). That bounds how fast `bulk` can fill, and it means the saturation window opens some time after the last submit returns.

[...]

Issue #2849, opened 2026-09-03
[TASK] DX — set-based seeders for orders, jobs, mappings and snapshots at three dataset sizes


piotrswierzy:
Child of #2840. **Week 3.** Upgrades #2843's crude inflation into a reusable dataset.

## Problem / Context

`perf/prestashop-baseline/seed-products.sh` proves the technique - 90 000 PrestaShop products inserted in **38 seconds** as set-based `INSERT ... SELECT` statements (nine of them, each a `CROSS JOIN` off a recursive-CTE row multiplier at `:80-83`), not a row-by-row loop. Nothing equivalent exists on the OpenLinker side - `grep generate_series` over the repo returns **nothing** - which is why no read-path figure exists at any row count.

#2843 gets the *curve shape* from crude `generate_series` inflation, and explicitly states the limitation: synthetic rows carry unrealistic distribution, so filter selectivity on health buckets and `syncStatus` is wrong. That moves the constant. This child fixes the constant.

**This is the sleeper cost of the programme.** `seed-products.sh` is 155 lines for one entity type against PrestaShop's simpler MySQL schema. Here it is `order_records` with jsonb snapshots, `order_line_items`, a year of `sync_jobs`, `identifier_mappings`, the catalogue and the snapshot tables - across 77 ORM entities with partial unique indexes, CHECK constraints and cross-table invariants. Get it subtly wrong and you measure query plans against data shapes that cannot occur in production.

## Proposed Solution

Set-based seeders at 10k / 100k / 1M, tagged with a prefix so cleanup matches on the prefix alone.

**After each size, run `VACUUM ANALYZE` on every seeded table and `pg_stat_statements_reset()` before the run that measures against it.** A set-based bulk insert leaves the planner on pre-seed statistics, and #2843's whole subject is plan choice - measuring against stale statistics risks attributing a bad plan to the wrong row count.

**The RNG seed used to generate the distribution is recorded in the manifest.** The current acceptance bar only requires the distribution to be *documented*, which is a description, not a reproduction: recording the seed is what lets a later campaign regenerate the identical dataset (same selectivity, same skew) rather than a merely similar one.

[...]

PR #2922, opened 2026-09-05
feat(perf): measure F2, F3, F5 and F9, and fix what the measuring exposed

norbert-kulus-blockydevs:
Part of #2842, #2843, #2848, #2849, #2852, #2854, #2866. **Stacked on #2918** - review against the stack tip, not `main`.

Opened as the F3 rig; it now carries the programme's first real measurement campaign.

## The numbers

**Webhook ingress (F3)** - p50 **5.5 ms**, p95 **7.5 ms**; **at least ~600/s** sustained with p99 218 ms and **zero failed requests at any rate**.

The scenario was designed around contention on the unique index. **No index contention was observed at any rate on any arm.** What binds is API CPU at 97.7%, the connection pool at its 40 ceiling, and Postgres `LWLock:WALWrite`. Scaling this path means more API processes and faster WAL storage; index tuning buys nothing.

The ~600/s is reported as a **floor, not a ceiling** - k6 was simultaneously at 96% of its own VU limit with 20-27% dropped iterations, so system saturation and generator saturation cannot be separated from this run.

**Operator read path (F5), seeded to 1M orders** - a hundred times the orders moved exactly **two** routes and left four unchanged.

| Route (median) | 10k | 100k | 1M |
|---|---|---|---|
| orders list, needs-attention | 11.4 ms | 47.7 ms | **149.0 ms** |
| orders list | 7.9 ms | 14.8 ms | 37.9 ms |
| whole page load | 28.0 ms | 35.0 ms | 58.0 ms |
| order detail / product detail / products list / sync-jobs list | \- | \- | **flat** |

It is one query: the pagination `COUNT`, evaluating a non-sargable jsonb filter with no `LIMIT`, while the paged query beside it stays at **0.2 ms** because it stops after twenty matches. The pattern is *any list reporting an exact total*, not *orders*. Growth is sub-linear only because Postgres switches to a parallel seq scan at that size - which a busy server may not have workers for - and the count pulls **~540 MB through the buffer pool per execution**, so the cost is application-wide cache pressure rather than one slow page.

**Stock propagation (F2)** - hop by hop: ~42 ms to the outbox, ~526 ms to OL's commit, ~3.6 s to `inventory_items`, ~9.0 s to a destination-write job dispatched. Never *"reached the marketplace"*: that hop has no endpoint on this stand and is reported unmeasured rather than estimated.

**Frontend (F9)** - bundle budgets wired into `check:invariants` with a demonstrated failing case; entry chunk 595.9 -> 408.0 kB by splitting stable vendor chunks, which is a cache-shape change and **not** a size cut, and the report says so.

## What the measuring exposed

Seven harness defects, all one shape: **none produced an error; each produced a plausible artifact that was quietly incomplete.**

| Defect | Effect |
|---|---|
| `guard_log_level` verified a value, never recorded it | manifests said `"unknown"` for a checked value |
| `run_one_arm` returned through captured stdout | its log text polluted a directory path |
| k6 rejects `-` in a metric name | two arms crashed before sending a request |
| k6's default summary carries **no p99** | the AC's headline percentile was unreachable |
| `docker exec -i` inside a `while read` here-string | the probe CSV was truncated to one row |
| `printf "- ..."` parses the dash as an option | a dated report was truncated mid-file |
| the k6 driver parsed the pool per VU | two 1 000/s runs OOM-killed, confirmed in `dmesg` |

Three guards changed as a result:

- **`guard_log_level`** now records what it checks.
- **`guard_stand_exclusive`** is new. Two scenarios measured one stand concurrently and silently invalidated each other - every pre-flight guard describes the world at script start, and a peer can falsify any of them before the window opens. Only exclusivity defends against that.
- **`guard_build`** compares product subtrees rather than commit shas. Measuring and committing necessarily interleave during a campaign, and the sha rule forced three full rebuilds for zero behavioural difference. It is **strictly stronger** than what it replaced: it additionally refuses uncommitted product changes and images built outside this repository. Its path list is `apps/api apps/worker`, not a bare `apps` - the browser bundle is never executed by the processes under measurement, and including it produced a false refusal within the hour.

The suite goes **63 -> 75** assertions, every new one verified red-first.

## Honesty

`perf/openlinker-throughput/campaign-2026-09-06.md` is the cross-flow report. Every figure carries a **measured / derived / extrapolated** label, and section 7 names what the campaign did not establish - no ceiling on dedicated hardware, no multi-replica behaviour, no production upstream latency, no read-path cost against the other 37 unbounded tables, ADR-050 unvalidated, and Erli's inbox cliff unreachable.

**Conditions, stated at the top rather than buried:** the stand is a compose project on a contended developer workstation, pinning no CPUs, sharing a host with two other running OpenLinker stacks. These are floors. A figure taken under contention understates what the software can do; it does not overstate it.


piotrswierzy (commented):
The harness work is strong — the seven self-reported defects, the red-first 63→75 assertions, and `guard_log_level` now recording what it verifies are the reported-vs-enforced discipline this repo runs on. But I'd hold merge on four things, three of which are the same shape one level in.

**Blocking**

1. **`guard_stand_exclusive`'s release never fires in either scenario.** Bash has one EXIT trap; `f2-stock-propagation.sh`'s and `f3-webhook-burst.sh`'s `run_smoke()` each register `trap "rm -… '$tmp'" EXIT`, which *replaces* the guard's `trap 'release_stand_exclusive' EXIT`. The docblock's "released on ANY exit path, including a guard's own `die`" is therefore false for both callers — depending on registration order either the stand stays locked for the full 3600 s TTL or the temp dir leaks. `lib-test.sh` calls `release_stand_exclusive` directly, so it cannot see this. Needs a library-owned `ol_on_exit` that *appends*, plus an assertion that a second registration does not evict the first. This is `docs/lessons.md:604` verbatim — a cited safeguard that does not exist makes a reviewer relax.

2. **The Lighthouse baseline is `hostContended: true`, perf 0.59, FCP 6.6 s — measured on a WSL workstation — and now gates every PR on GitHub-hosted `ubuntu-latest` at ±0.05.** A cross-machine median comparison either can never fail (green over an unasked question — the class this PR's own body says it is fixing) or always fails. `run-lighthouse.mjs` records `hostContended` and then acts on nothing. Either refuse to gate against a contended baseline, or re-baseline on the runner class that will do the comparing. Note `get_check_runs` is empty on this PR — neither new gate has actually run here.

**Important**

3. `guard_build` assigns `MANIFEST_GIT_SHA="$rev"` after the loop, so a mixed fleet records only the last container — the exact defect `guard_log_level` fixes twenty lines down ("a manifest that printed only the last one would be silently wrong on a mixed fleet"), and newly *reachable* now that api and worker may legitimately sit on different tree-identical commits. The test seeds both containers with the same sha, so it cannot catch it. Record `container=rev` pairs.

4. `release_stand_exclusive` matches ownership on `*":pid$$@"*` — PID only, ignoring scenario and host — so it will `DEL` another host's lock on a PID collision, which is the "two holders at once" failure the check exists to prevent. The test seeds `pid999999`, guaranteeing a miss. Keep `owner` in a global at claim time and compare it exactly.

5. `check-bundle-budgets.mjs` runs `rmSync(apps/web/dist)` + a full `vite build` inside `check:invariants`, i.e. on every `pnpm lint` and so every pre-commit. Every other invariant script is static analysis over source; this makes a one-line `libs/core` commit pay a frontend production build and deletes the developer's `dist` as a side effect. Move it to its own CI job gated on `apps/web` changes (like the `lighthouse` job beside it), leaving only `--self-check` in the lint chain.

6. `npx --yes lighthouse@12.8.2` in CI is a live npm fetch outside `pnpm-lock.yaml` / `--frozen-lockfile`. Add it as an `apps/web` devDependency and invoke via `pnpm exec`.

**Suggestions**

- `run-lighthouse.mjs`: `if (!base) … continue` silently un-gates a renamed or added route — fail and tell the author to re-baseline.
- `check-bundle-budgets.mjs`: the self-check asserting "an absent group is not flagged" *is* the hole. If `index-*.js` is ever renamed, `classifyChunk` files those bytes as `route` and the `index` budget passes at zero. Assert every `MEASURED_BASELINE_BYTES` group was actually observed in `dist`.

Genuinely good: building `dist` inside the Lighthouse runner so the `:3000` cross-stack leak is structurally impossible rather than caller discipline; `BUILD_RELEVANT_PATHS` narrowed to `apps/api apps/worker` with the false-refusal incident recorded in place; and the `manualChunks` comment saying outright that this is a cache-shape change and not a size cut.

piotrswierzy (approved):
## Summary

Reviewed against the stack base `2841-perf-harness-lib`, not `main`. 39 files; the great majority is measurement harness (`perf/openlinker-throughput/**`) and reports, which I read for correctness of the claims they make rather than line-by-line. I prioritised the code that leaves the harness and lands on every developer and every CI run: `apps/web/vite.config.ts`, `scripts/check-bundle-budgets.mjs`, `scripts/lighthouse/*`, `package.json`'s `check:invariants` chain, and `.github/workflows/ci.yml`.

The measurement work is genuinely good, and the honesty discipline is the best part of it: every figure labelled measured/derived/extrapolated, `~600/s` reported as a floor with the generator's own saturation disclosed, "enqueued" never written as "delivered" on the F2 hop labels, and the seven harness defects listed with the shape they share ("none produced an error; each produced a plausible artifact that was quietly incomplete"). The F5 finding — that the cost is the pagination `COUNT` over a non-sargable jsonb filter with no `LIMIT`, generalising to *any list reporting an exact total* rather than to *orders* — is correctly framed as a pattern and correctly left unfixed here.

No architecture or security finding. `libs/core`, `apps/api` and `apps/worker` source are untouched, so "fix what the measuring exposed" is harness fixes plus a frontend cache-shape change, and the report says so rather than claiming a size cut. My concerns are all about the **new always-on gates**: two of them can pass without having checked what they claim to check, which is the failure class this programme's own `--self-check` discipline exists to close.

---

## 3️⃣ Issues & Risks

### 🟡 IMPORTANT

**1. `check-bundle-budgets.mjs` runs a full `vite build` inside `check:invariants` — every `pnpm lint`, on every commit, for every developer**

`package.json` appends `node scripts/check-bundle-budgets.mjs` (no flags) to `check:invariants`, and `main()` with no flags calls `buildWeb()` → `rmSync(apps/web/dist)` + `pnpm exec vite build`. `check:invariants` is chained into `pnpm lint`, and `.husky/pre-commit` runs `pnpm lint` unconditionally — so a backend-only commit now pays a full frontend production build.

This is a categorical break with the shape of that chain. Every one of the ~50 other scripts in it is a fast static scan; `grep -l "execFileSync\|execSync\|spawnSync" scripts/check-*.mjs` returns exactly one file today (`check-migration-timestamps.mjs`, and only for `git ls-tree`). Nothing in `check:invariants` builds anything. The hook is already at the edge of tolerable — this pushes it over for changes that cannot affect the bundle at all.

The `rmSync` is a second, separate cost: `check:invariants` now deletes `apps/web/dist` as a side effect of linting. A developer with a `vite preview` served from `dist` loses it to an unrelated backend commit.

Suggested: keep `--self-check` in `check:invariants` (it is pure and fast, and that is the part that proves the gate can fail), and run the building half in CI — either in the existing `build` job or beside the new `lighthouse` job, both of which already build. That preserves the guard without putting a bundler in the pre-commit path.

**2. `classifyChunk` is untested, and a rename in `vite.config.ts` silently disables three of the six budgets** — `scripts/check-bundle-budgets.mjs`

The docblock claims `classifyChunk` is "exercised indirectly through `evaluateBudgets` on synthetic filenames". It is not: `evaluateBudgets(sizesByGroup, budgets)` takes an already-classified size map and never sees a filename. `--self-check` calls only `evaluateBudgets`. So the one function that couples this script to `vite.config.ts` has no coverage anywhere.

That coupling is a mirror in the repo's established sense — `manualChunks` in `apps/web/vite.config.ts` assigns the names, `classifyChunk` in `scripts/` re-derives them — and the failure mode is silent in the dangerous direction. Rename `editor-vendor` to `tiptap-vendor` in the vite config and `classifyChunk` returns `'route'` for it; the `editor-vendor` group then measures 0 bytes, and `evaluateBudgets` deliberately does **not** flag an absent group (self-check assertion 3 pins that behaviour, correctly). The result is a budget that reports `ok` for a chunk it is no longer looking at. The 395 kB Tiptap chunk could double and the gate stays green — only `total` would still see it.

This is precisely what the `check-*-mirror.mjs` family exists to prevent, and what `docs/testing-guide.md § Port-contract suites` states as "a case that ran and compared nothing". Two options, either sufficient: add `classifyChunk` cases to `--self-check` **and** assert that every non-`total` key of `MEASURED_BASELINE_BYTES` is produced by `classifyChunk` on at least one filename from the real `dist`; or, better, derive both sides from one exported constant so a rename is a compile/lint break rather than a silent reclassification.

**3. `run-lighthouse.mjs` skips a route missing from the baseline and still exits 0** — `scripts/lighthouse/run-lighthouse.mjs`

```js
if (!base) {
  console.log(`  ${label}: no baseline entry — skipping comparison`);
  continue;
}
```

`regressed` stays false, so a route added to `ROUTES` (or renamed) is audited, printed, and gated on nothing — the CI job goes green over a question it never asked. `docs/testing-guide.md § Writing one` names this shape directly: *"Refuse, never skip. No usable subject and an empty case table must THROW. Reporting them lets a caller render '0 failures' over an unasked question — the shape that reported green over a live divergence for months in #2673."*

The same rule applies in the other direction: a baseline entry with no matching route (a deleted route whose baseline was left behind) should also fail, so `declared === covered` on both sides. Fail on either mismatch and tell the operator to run `--update-baseline`.

**4. The committed Lighthouse baseline records `hostContended: true`, and now gates CI on a different machine class** — `scripts/lighthouse/baseline.json`

The baseline's own first line is `"hostContended": true`, and `REGRESSION_THRESHOLD = 0.05` hard-fails the job. It was captured on the contended developer workstation the campaign report is careful to describe as a floor; the gate runs on `ubuntu-latest`. `performance` is the volatile category (0.59 median, FCP 6.6 s) and it is the one most sensitive to the machine — the other three (`accessibility` 0.96, `best-practices` 1, `seo` 0.91) are essentially deterministic and are the ones this gate genuinely defends.

Either regenerate the baseline on `ubuntu-latest` — the machine the comparison runs on — or gate on the three stable categories and report `performance` without failing on it. As it stands the most likely first outcome is a red CI on an unrelated PR, and the standard response to that is to widen the threshold until the gate stops meaning anything.

**5. `guard_stand_exclusive` takes the single `EXIT` trap slot, and F2's `--smoke` path then clobbers it** — `perf/openlinker-throughput/lib.sh`, `scenarios/f2-stock-propagation.sh`

The guard registers `trap 'release_stand_exclusive' EXIT` and documents "Released on ANY exit path, including a guard's own `die`". A shell has one `EXIT` trap, and a later `trap ... EXIT` replaces it silently.

In `f2-stock-propagation.sh`, `guard_stand_exclusive "f2-stock-propagation"` runs at top level, and `run_smoke()` then registers `trap "rm -f '$tmp_csv'" EXIT`. On the `--smoke` path the release is unregistered, so the lock survives to its full `OL_STAND_LOCK_TTL_SECS` (default 3600) and blocks the stand for an hour — the exact refusal the guard exists to produce, arriving for the wrong reason. `f3-webhook-burst.sh` has the mirror image (`run_smoke` registers its trap, `run_strict` calls the guard) but is safe only because those are alternative modes.

`lib-test.sh`'s `trap - EXIT   # the guard registers its own EXIT trap; the suite owns its exit` shows the collision was noticed in the test and not carried back to the callers. Suggested: have the guard append to a shared cleanup list (a `_perf_on_exit` function callers add to) rather than own the slot, so no caller can silently take it back.

### 🟢 SUGGESTION

**6. `guard_build` records the *last* container's revision, which the fix one function below argues against** — `perf/openlinker-throughput/lib.sh`

`rev` is assigned inside `for c in $containers`, and after the loop `MANIFEST_GIT_SHA="$rev"` keeps whichever container was checked last. Because the new comparison is per-product-path rather than per-sha, api and worker can legitimately pass while built from *different* commits — that is the whole point of the change — so the manifest can now name a commit only one of the two processes was built from.

This is the same defect `guard_log_level` fixes twenty lines later, with its reasoning stated explicitly: *"Recorded as `container=value` pairs rather than a single scalar: the guard reads N containers and they can legitimately differ, and a manifest that printed only the last one would be silently wrong on a mixed fleet."* Record `container=rev` pairs here for the same reason.

---

## Positive observations

- **`manualChunks` is correct and correctly described.** The comment states the measured before-state (`ui-*.js` carrying both `@radix-ui` and `@tiptap`, `index-*.js` carrying `react-dom`), groups by change frequency rather than by package name, and says plainly that this changes what a *warm* visitor re-downloads and does not shrink the cold payload. The one-shared-`vendor`-chunk decision is argued from HTTP overhead rather than left implicit. The report repeating the same caveat is what keeps 595.9 → 408.0 kB from being read as a size win.
- **`guard_stand_exclusive`'s reasoning generalises correctly**: *"every pre-flight guard describes the world at script start, and a peer with write access to the stand can falsify any of them before the window opens."* Redis-held rather than a local lock file (right — the stand is the contended resource, not the checkout), `SET NX` as the atomic claim with nothing reading-then-deciding, TTL framed as a crash bound rather than a run bound, and no `--force` flag. `release_stand_exclusive` checking the value before deleting is the detail most implementations of this get wrong.
- **`guard_build`'s widening is genuinely strictly stronger**, as claimed: it adds a refusal for uncommitted product changes (which a sha comparison against `HEAD` cannot see at all) and for images built outside the repository, while dropping only a refusal that had no behavioural basis. `apps/api apps/worker` rather than a bare `apps` is justified from what the image stages actually copy, and the false refusal it caused is recorded rather than smoothed over.
- **The `.gitignore` entries are scoped and each carries its reason**, and `results/` follows the stated `perf/prestashop-baseline` precedent rather than inventing a new one.
- **The `lighthouse` CI job's placement is argued, not assumed** — `ubuntu-latest` to keep it off scarce self-hosted slots and to avoid another Chrome-launch dependency on shared state, with a pointer to why the build must not be a separate step. `if: always()` on the artifact upload is the right call: a failed run's numbers are the ones you need.

---

## Final Assessment

A large, careful, and unusually honest measurement campaign whose production-code footprint is one well-reasoned Vite chunking change. Nothing here is unsafe to merge. The concerns are concentrated in the two new always-on gates: **#2** and **#3** are both cases where the guard can report success without having checked what it claims to check — the exact class this programme's `--self-check` convention was written against — and **#1** puts a bundler in every developer's pre-commit path for changes that cannot affect the bundle.

### Merge Readiness

🔄 **Approve with changes**

### Priority fixes

1. **#2** — cover `classifyChunk`, or derive the group names from one source shared with `vite.config.ts`. Until then three of the six budgets disable themselves silently on a rename.
2. **#3** — make a missing baseline entry (either direction) fail rather than skip.
3. **#1** — move the building half of the bundle budget out of `check:invariants` into CI; keep `--self-check` in the chain.
4. **#4** — regenerate the Lighthouse baseline on the runner the gate executes on, or stop failing on `performance`.
5. **#5** — stop `guard_stand_exclusive` owning the `EXIT` slot (F2 `--smoke` leaks the stand lock for an hour today).
6. **#6** — record per-container revisions in `MANIFEST_GIT_SHA`, matching `guard_log_level`.

Issue #2931, opened 2026-09-06
[TASK] DX — share the pre-signed pool across k6 VUs instead of parsing it per VU


norbert-kulus-blockydevs:
## Problem / Context

The F3 k6 driver `open()`s a JSON pool of pre-signed webhook payloads at init scope. **Each virtual user parses its own copy**, so memory grows as pool size times VU count.

At the 1 000/s tier that pool reaches **40 MB** (roughly 90 000 signed requests for a 90-second window) and the run needed 132 VUs. **Two runs were OOM-killed**, confirmed in `dmesg` and reported in `perf/openlinker-throughput/results-F3-2026-09-06.md`.

The pre-signing itself is not the problem and must stay - it is what satisfies two of #2842's requirements at once: signing reuses `apps/e2e/src/support/webhooks.ts` rather than restating the HMAC scheme, and k6 computes no HMAC during the window, so the generator does not contend with the API it is measuring for CPU on an unpinned host. The k6 driver contains no crypto call at all, by design.

What does not scale is the **representation**. Pool bytes grow linearly with the arrival rate, and the per-VU duplication multiplies that by the concurrency the same rate demands - so the memory cost is roughly quadratic in the thing being swept. At 3 000/s the pool alone would be ~120 MB before duplication.

This is a hard limit on how far the F3 sweep can go, and it is currently the reason the ceiling is reported as "at least ~600/s" rather than a number.

## Proposed Solution

Three shapes, in rough order of effort:

- **A `SharedArray`.** k6's own answer to exactly this: data declared in init scope inside `new SharedArray(...)` is parsed once and shared across VUs. Smallest change, likely sufficient.
- **A compact wire format.** The pool is currently JSON objects; a line-delimited format read lazily, or the body and signature stored as flat strings rather than nested objects, cuts the parsed footprint substantially before any sharing.
- **Generate rather than store.** Pre-sign a smaller pool of *generations* and have VUs index into it modulo their iteration, accepting that the same body repeats. This changes what the arm measures for the `unique` case, so it is only viable for the replay arms.

Whichever is chosen, the harness should **refuse to build a pool it cannot hold** rather than discovering it through the OOM killer: a pre-flight estimate against available memory, failing with the number it wanted and the number it had.

## Classification

**Type**: DX
**Layer**: Shared (measurement)
**File(s)**: `perf/openlinker-throughput/drivers/webhook-burst.js`, `perf/openlinker-throughput/drivers/presign-webhooks.mjs`

[...]

PR #2938, opened 2026-09-06
feat(perf): refuse a run whose load generator was itself saturated

### LINKED ISSUE #2933: [TASK] DX — refuse a run whose load generator was itself saturated
## Problem / Context

F3's headline was very nearly published as *"the system's ceiling is ~600 requests per second"*. It is not that, and the artifacts say so only if you go looking:

```
target 1000/s -> achieved 601.8/s   unique   vus_used=132  vus_max_cfg=137  dropped=19.6%
target 1000/s -> achieved 568.0/s   replay   vus_used=149  vus_max_cfg=154  dropped=24.1%
target 1000/s -> achieved 548.7/s   replay   vus_used=143  vus_max_cfg=156  dropped=26.7%
```

**k6 was at 92-97% of its own configured virtual-user ceiling** and dropping a fifth to a quarter of its iterations. An iteration is dropped when no VU is free, so part of the gap between offered and achieved is the generator running out, not the system refusing work.

Two independent signals pointed in opposite directions and both were real - `lab-api` CPU at 97.7% says the system saturated; the VU exhaustion says the generator did too - and they could not be separated from that run. The figure is therefore reported as *"at least ~600/s"*, which is honest but weaker than it needed to be.

**Nothing in the harness noticed.** Every guard passed, the verdict was `VALID`, and the number would have been quoted as a ceiling if a human had not read `dropped_iterations` and the VU configuration side by side.

This is the same class as the six defects the campaign already found: no error, a plausible artifact, quietly incomplete.

## Proposed Solution

A post-guard that refuses a run whose **instrument** was saturated, in the same way the existing post-guards refuse a run whose **conditions** drifted.

Two signals, both already in the k6 summary:

- `vus.max / vus_max` above a threshold - 0.9 is a reasonable starting point and should be stated rather than buried
- `dropped_iterations` as a fraction of `dropped_iterations + http_reqs` above a threshold

Either tripping should write `DISCARDED` with a reason naming the generator, not the system - and the message should say what to do (raise `MAX_VUS` / `PRE_ALLOCATED_VUS` and re-run), because a discard an operator cannot act on is just an obstacle.

Worth considering as a *warning* rather than a discard when the run is deliberately probing saturation: a scenario that sets out to find a ceiling will legitimately approach the generator's limit, and the honest answer there is to raise the limit rather than to refuse the run. If so, make that an explicit scenario-declared posture, never an ambient default - the whole point is that the condition must be stated rather than inherited.

## Classification

**Type**: DX
**File(s)**: `perf/openlinker-throughput/lib.sh`, `perf/openlinker-throughput/lib-test.sh`

## Dependencies

- None. Both signals are already present in every k6 summary the harness writes.

## Assumptions

- k6 reports `dropped_iterations` for the arrival-rate executors used here. It does; it will not appear for `constant-vus`, so the guard must tolerate its absence rather than treating a missing metric as zero.

## Acceptance Criteria

- [ ] A run whose VU utilisation exceeds the threshold is `DISCARDED` with a reason naming the generator and the remedy
- [ ] A run with a high dropped-iteration fraction is likewise refused
- [ ] A missing `dropped_iterations` metric is handled as "not applicable", never silently as zero
- [ ] The thresholds are declared constants with the reasoning beside them, not literals
- [ ] `lib-test.sh` covers both trip conditions and the not-applicable case - verified red-first
- [ ] F3's existing figures are re-checked against the new guard, and the report says whether they would have passed it

norbert-kulus-blockydevs:
Closes #2933. Based on **`2842-f3-webhook-burst`**, which reaches the epic branch through #2922 - review against the stack tip.

## Why

F3's headline was nearly published as *"the system's ceiling is ~600 requests per second"*. It was nothing of the kind.

Every guard passed. The verdict read `VALID`. Only reading `dropped_iterations` and the VU configuration side by side revealed this:

```
target 1000/s -> achieved 601.8/s   vus_used=132  vus_max_cfg=137  dropped=19.6%
target 1000/s -> achieved 568.0/s   vus_used=149  vus_max_cfg=154  dropped=24.1%
target 1000/s -> achieved 548.7/s   vus_used=143  vus_max_cfg=156  dropped=26.7%
```

k6 was at **92-97% of its own virtual-user ceiling**. Two signals pointed in opposite directions and both were real - the api at 97.7% CPU said the *system* had saturated, the VU exhaustion said the *generator* had too - and that run could not separate them. The figure survives only as "at least ~600/s".

**Nothing in the harness noticed.** Same shape as the seven defects this campaign has already found: no error, a plausible artifact, quietly incomplete.

## What this adds

`post_guard_generator_saturated` refuses a run whose **instrument** was saturated, the way the existing post-guards refuse one whose **conditions** drifted. Both signals are already in every k6 summary the harness writes.

Four properties are load-bearing.

**No opt-out, not even for a sweep deliberately hunting a ceiling.** That case feels like the obvious exception and is the strongest reason to refuse: if the generator is at its limit, the system's ceiling is precisely what the run cannot tell you. The refusal names the remedy - raise `MAX_VUS` and re-run - because a discard an operator cannot act on is just an obstacle.

**An absent `dropped_iterations` means NOT APPLICABLE, never zero.** `constant-vus` has no arrival rate to fall behind and omits the metric entirely; reading that as zero would silently pass the check it belongs to.

**A claimed-but-missing summary discards rather than skipping.** No summary path means "this scenario drives no generator" and passes. A path to a file that does not exist means the run claimed an instrument and produced nothing - the OOM-killed-k6 shape (#2931). A run whose instrument vanished is not a run whose instrument behaved.

**The omission is documented at every call site that makes it.** F2 and F7 drive load through stock writes and enqueued jobs rather than HTTP, so they pass no summary, and each says so in place - "no generator" and "forgot to pass it" must not look the same.

## Validated against the real artifacts

Not against fixtures alone - against the F3 summaries whose numbers prompted this:

| Run | Verdict |
|---|---|
| 43/s unique | ok |
| 240/s unique | ok |
| 240/s replay-concurrent | ok |
| **~600/s unique** | **DISCARDED** - 132 of 137 VUs (96%) |
| **~600/s replay-concurrent** | **DISCARDED** - 143 of 156 VUs (91%) |
| no generator (F2 shape) | ok |
| claimed generator, summary missing | **DISCARDED** |

It refuses exactly the runs whose numbers had to be hedged, and passes the ones that did not.

## Tests

Eight assertions, all verified red-first. Suite **75 -> 83**.

One was weak on the first attempt and the red-first pass caught it: the missing-summary test asserted only on the word `DISCARDED`, and with the branch removed the function fell through to `jq`, failed on the nonexistent path, and hit a could-not-parse fallback that *also* says `DISCARDED`. It now asserts the specific message. That is the second time in this campaign red-first has caught a test that passed for the wrong reason.

```
=== 83 passed, 0 failed ===
```

It does not re-measure F3. The existing ~600/s figures stay reported as a floor; re-taking them with a VU ceiling well clear of the observed usage is #2931's job, since the pool duplication is what stops the ceiling being raised today.

### DIFF
--- a/docker-compose.lab.yml
+++ b/docker-compose.lab.yml
@@ -55,6 +55,22 @@ services:
     # only plans over its own threshold (2s) to keep the F5 window's own log
     # volume bounded; pg_stat_statements is reset per dataset size by the
     # scenario itself, never here.
+    #
+    # statement_timeout=30s (#2934): this was, until this change, a LIVE-ONLY
+    # `ALTER ROLE postgres SET statement_timeout = ...` applied by
+    # f5-read-path.sh's own apply_statement_timeout() and never declared
+    # anywhere - so it existed only in a running database, did not survive a
+    # stand rebuild, and no run's generic manifest recorded that it was in
+    # force (found #2843/#2934: the same role's own 1M-order seed got
+    # cancelled mid-transaction by it, before the seeder learned to
+    # `SET statement_timeout = 0` for its own session - see seed-lib.sh's
+    # seed_sql()). It bounds the API's runaway-read risk for EVERY session on
+    # this stand now, server-wide, from boot - f5-read-path.sh's own
+    # apply_statement_timeout() still runs (still useful for an operator who
+    # wants a DIFFERENT STATEMENT_TIMEOUT_MS for one F5 run) and simply
+    # re-asserts the same 30s by default via its own ALTER ROLE, which is
+    # captured into every scenario's manifest generically now - see
+    # manifest_gather_environment's `postgres.nonDefaultSettings` in lib.sh.
     command:
       - postgres
       - -c
@@ -71,6 +87,8 @@ services:
       - auto_explain.log_analyze=on
       - -c
       - auto_explain.log_buffers=on
+      - -c
+      - statement_timeout=30000
     ports:
       - '127.0.0.1:${POSTGRES_HOST_PORT:-19432}:5432'
     volumes:
--- a/perf/openlinker-throughput/lib-test.sh
+++ b/perf/openlinker-throughput/lib-test.sh
@@ -391,6 +391,61 @@ assert_eq "guard_log_level records what it verified" \
 
 OL_API_CONTAINER="lab-api"; WORKER_CONTAINERS="lab-worker"
 
+echo "--- post_guard_generator_saturated ---"
+# Writes a k6-shaped summary to $1 with the metrics given, so each case differs
+# in exactly the field it is about.
+write_k6_summary() {
+  local out="$1" used="$2" cfg="$3" dropped="$4" reqs="$5" extra=""
+  [ "$dropped" = "none" ] || extra=",\"dropped_iterations\":{\"count\":$dropped}"
+  cat > "$out" <<JSON
+{"metrics":{"vus":{"max":$used},"vus_max":{"max":$cfg},
+ "http_reqs":{"count":$reqs}$extra}}
+JSON
+}
+
+sat_dir="$(mktemp -d)"
+
+write_k6_summary "$sat_dir/healthy.json" 12 300 0 10000
+assert_eq "a generator with headroom passes" "ok" \
+  "$(post_guard_generator_saturated "$sat_dir/healthy.json")"
+
+# The case this guard exists for: F3's ~600/s runs sat at 92-97% of their VU
+# ceiling and their numbers were nearly published as a system ceiling.
+write_k6_summary "$sat_dir/vu-bound.json" 132 137 0 18100
+assert_contains "a generator at 96% of its VU ceiling is discarded" \
+  "$(post_guard_generator_saturated "$sat_dir/vu-bound.json")" "DISCARDED"
+assert_contains "the refusal names the remedy, not just the fault" \
+  "$(post_guard_generator_saturated "$sat_dir/vu-bound.json")" "MAX_VUS"
+
+write_k6_summary "$sat_dir/dropped.json" 20 300 4400 18100
+assert_contains "a high dropped-iteration fraction is discarded" \
+  "$(post_guard_generator_saturated "$sat_dir/dropped.json")" "DISCARDED"
+
+# constant-vus has no arrival rate to fall behind, so k6 emits no
+# dropped_iterations at all. Absent must mean NOT APPLICABLE - reading it as
+# zero would silently pass the check it belongs to.
+write_k6_summary "$sat_dir/no-drop-metric.json" 20 300 none 18100
+assert_eq "an absent dropped_iterations metric is not applicable, not zero" "ok" \
+  "$(post_guard_generator_saturated "$sat_dir/no-drop-metric.json")"
+
+# A scenario with no load generator at all (F2) is not applicable...
+assert_eq "no summary path means no generator, which passes" "ok" \
+  "$(post_guard_generator_saturated "")"
+# ...but one that CLAIMED a generator and produced nothing is the OOM shape.
+# Asserted on the SPECIFIC message, not merely on the word DISCARDED: without
+# the missing-file branch the function falls through to jq, which fails on a
+# nonexistent path and hits the could-not-parse fallback - which also says
+# DISCARDED. A weaker assertion passed against the broken code (found red-first).
+assert_contains "a claimed-but-missing summary is discarded, never skipped" \
+  "$(post_guard_generator_saturated "$sat_dir/never-written.json")" "wrote no summary at"
+
+# A summary that cannot show its instrument was healthy is not reportable.
+printf '{"metrics":{"http_reqs":{"count":100}}}\n' > "$sat_dir/no-vus.json"
+assert_contains "a summary carrying no vus/vus_max is discarded" \
+  "$(post_guard_generator_saturated "$sat_dir/no-vus.json")" "DISCARDED"
+
+rm -rf "$sat_dir"
+
 # ===========================================================================
 # post-guards
 # ===========================================================================
--- a/perf/openlinker-throughput/lib-test.sh
+++ b/perf/openlinker-throughput/lib-test.sh
@@ -93,6 +93,7 @@ pg_sql() {
     *"operational_settings"*) echo "${FAKE_PG[operational_settings]:-{\}}" ;;
     *"COUNT(*)"*) echo "${FAKE_PG[count]:-0}" ;;
     *"pg_database_size"*) echo "${FAKE_PG[db_size]:-123456}" ;;
+    *"pg_db_role_setting"*) echo "${FAKE_PG[pg_overrides]:-}" ;;
     *"string_agg"*) echo "${FAKE_PG[agg]:-}" ;;
     *) echo "" ;;
   esac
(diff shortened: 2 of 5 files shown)


piotrswierzy (commented):
The failure-path discipline here is a clear step up from #2918 — missing file, malformed JSON, absent `vus`, and a missing `jq` all fail **closed**, with no `|| true` swallowing an exit code and no stderr folded into a count (the two shapes that broke `post_guard_limiter_degraded` and `reset_between_repeats`). The three-way empty-path / missing-file / readable-file split is the right distinction, the missing-file arm correctly reads an OOM-killed k6 as a discard rather than a skip, and treating an absent `dropped_iterations` as NOT APPLICABLE rather than zero is exactly the `null ≠ 0` discipline this repo enforces elsewhere.

**Blocking: `vus_max` is not the configured ceiling.**

Under `ramping-arrival-rate`, k6 pre-allocates `preAllocatedVUs` and then allocates a VU only when it needs one, so `vus_max` is the *peak allocated pool*, not `maxVUs`. This PR's own evidence proves it: `MAX_VUS=300` is configured at `f3-webhook-burst.sh:57`, yet the three runs report `vus_max` of 137, 154 and 156 — three different values for one config. And `results-F3-2026-09-06.md:120-123` states the real constraint was ramp *rate* off 50 `PRE_ALLOCATED_VUS`, so the run was at **46%** of its ceiling (137/300), not 96%.

Three consequences:

1. Peak-used ≈ peak-allocated by construction, so any arrival-rate run that grows past `preAllocatedVUs` trips the guard. A healthy run needing 51 of an available 300 reports `51 of 51 (100% of its own ceiling)` and is discarded — with no opt-out.
2. `"k6 used 132 of 137 virtual users (96% of its own ceiling)"` is a false statement in a published artifact; 137 was never a ceiling.
3. The named remedy "raise `MAX_VUS`" cannot clear the guard, because the denominator tracks demand rather than the ceiling.

Compare against the configured ceiling — pass `MAX_VUS` into the guard and test `vus_max.max >= MAX_VUS` (allocation actually hit the wall) — or drop the VU ratio entirely and rely on `dropped_iterations`, which is the signal k6 actually *reports* rather than one inferred from it.

**And the test cannot catch it** (`lib-test.sh:404`): `write_k6_summary "$sat_dir/healthy.json" 12 300 0 10000` writes `vus_max=300`, encoding the very assumption the guard gets wrong — a shape real k6 never produced in this campaign. That's the #2918 defect recorded in `docs/lessons.md`: the fake models the argument the way the guard assumes it behaves, so the happy-path assertion passes against a broken guard. Add a case with `used=51, cfg=51` and assert the expected verdict, or build a fixture from a committed real k6 summary.

**Also important**

- `lib.sh:1004,1010` — both thresholds are env-overridable, so `GENERATOR_VU_UTILISATION_MAX=2` disables the guard for a whole campaign, yields `status=VALID`, and leaves no trace in `verdict.txt`. That directly contradicts the PR's own stated load-bearing property ("There is deliberately NO opt-out"). Either make them constants, or record the applied threshold on the pass path too — the harness already has the counter-pattern (`lib-test.sh:391`, "guard_log_level records what it verified").
- `f5-read-path.sh:309` — F5 drives k6 (`run_k6 … "$dir/k6-summary.json"`) and never calls `run_post_guards`, so the one other k6 scenario in the tree escapes this guard silently — and F5 publishes p50/p95/p99 per route, figures a saturated generator distorts. "The omission is documented at every call site that makes it" holds only for `run_post_guards` callers; F5 is an undocumented third case.
- `lib.sh:1077` — `k6_summary="${7:-}"` plus the empty-path `ok` arm means a future scenario that forgets the argument gets a silent pass: "absence reads the same as ok", which is the exact shape this guard exists to prevent one level up. Make the parameter required with an explicit `none` sentinel for a genuinely generator-less scenario.

**Suggestions**

- `lib.sh:1032` — the drop denominator should be `iterations.count`, not `http_reqs.count`. `dropped/(dropped + http_reqs)` is correct only because `webhook-burst.js` issues exactly one request per iteration; a driver issuing two inflates the denominator and *understates* the drop ratio — a fail-open coupled to a driver detail.
- `lib.sh:1041` — only the first reason is reported (`.[0]`). The ~600/s F3 run was both 96% VU-bound *and* dropping 19.6%, so the operator sees one of two. `run_post_guards` already joins multiple reasons.
- `lib-test.sh:441` — the `could not parse` fallback has no case of its own, and the PR narrative says a weak test initially passed *through* it. It's the guard's last line of defence.
- None of these 83 assertions run in CI — `perf/` is outside `pnpm-workspace.yaml` and referenced by no workflow. Pre-existing rather than introduced here, but the PR's confidence rests on a suite nothing enforces; `bash perf/openlinker-throughput/lib-test.sh` in `check:invariants` would cost one line.

Portability, secrets and error handling are clean — no GNU-isms in the new code, no credentials, proper header comment.

Best thing here: the red-first find that asserting only on `DISCARDED` passed via the `could not parse` fallback, and naming it in the write-up. That's the campaign's own lesson applied to itself. Declining an opt-out for ceiling-hunting sweeps is also right, and reasoned rather than assumed.

piotrswierzy (approved):
## Summary

Adds `post_guard_generator_saturated` to the perf harness: a run whose k6 generator was at ≥90% of its VU ceiling, or dropping >5% of intended iterations, is `DISCARDED` rather than reported as a system measurement. The three-way handling of the summary path (empty = no generator / missing = OOM shape / readable = evaluate) is the right decomposition, the absent-`dropped_iterations` case is correctly treated as *not applicable* rather than zero, and the jq is null-safe throughout (`.metrics as $m` against an absent `metrics` yields `null` and lands on the "no vus/vus_max" discard, which is the safe direction). The red-first note about the missing-summary test asserting only on `DISCARDED` — and thereby passing against the broken code via the `could not parse` fallback — is exactly the class of test defect `docs/testing-guide.md § Port-contract suites` warns about, and catching it is a good sign about the process.

The most important concern is coverage, not correctness: the guard is wired into one of the two k6-driven scenarios.

---

## Findings

**[IMPORTANT]** — `perf/openlinker-throughput/scenarios/f5-read-path.sh:309-330`
> **F5 drives k6 and is not covered by the new guard**, and the PR description's account of the call sites is inaccurate. The description says *"F2 and F7 drive load through stock writes and enqueued jobs rather than HTTP, so they pass no summary"* — but `perf/openlinker-throughput/scenarios/` contains exactly three scenarios (`f2`, `f3`, `f5`); there is no `f7` on this branch, and `f5-read-path.sh` **does** run k6 (`run_k6 "$dir/sample-ids.json" … "$dir/k6-summary.json"`, line 309) with a `ramping-arrival-rate` executor (`drivers/read-path.js:133`) — i.e. precisely the executor shape that emits `dropped_iterations` and carries a `maxVUs` ceiling. F5 bypasses `run_post_guards` entirely and hand-writes its own verdict:
> ```sh
> # scenarios/f5-read-path.sh:317-330
> # ... the job-oriented run_post_guards/verdict_write machinery does not apply ...
> printf 'status=%s\n' "VALID"
> ```
> That branch writes `VALID` unconditionally. So the one other scenario that could be measuring its own generator instead of OpenLinker still cannot say so — which is the defect this PR exists to close, one scenario over. `post_guard_generator_saturated` is deliberately a standalone function taking only a path, so wiring it into F5's verdict block is a few lines and needs none of the job-oriented machinery F5 correctly declines. At minimum, F5's comment should be extended to say the generator check is knowingly absent there, so this reads as a scoped decision rather than an oversight — the same standard the F2 call site is held to in this very diff.

**[IMPORTANT]** — `perf/openlinker-throughput/lib.sh:1005,1011`
> **The "no opt-out" claim and the implementation disagree.** The docblock states, emphatically: *"There is deliberately NO opt-out, not even for a sweep that is intentionally hunting a ceiling."* Six lines later:
> ```sh
> GENERATOR_VU_UTILISATION_MAX="${GENERATOR_VU_UTILISATION_MAX:-0.9}"
> GENERATOR_DROPPED_RATIO_MAX="${GENERATOR_DROPPED_RATIO_MAX:-0.05}"
> ```
> `GENERATOR_VU_UTILISATION_MAX=2 ./run.sh` silently disables the VU check for the whole run, with nothing in the manifest recording that it was overridden. That is the "ambient default" #2933 explicitly warns against (*"make that an explicit scenario-declared posture, never an ambient default"*), and it is the shape of the very defect the guard was written to catch: a run that reads `VALID` while a condition it depends on was quietly not in force.
>
> The env-override pattern itself is repo convention and fine — but the established precedent bounds it. `guard_perf_max_attempts` (`lib.sh:300-306`) is the model: `PERF_MAX_ATTEMPTS` is env-settable **and** refused when the value *"would cap nothing"*:
> ```sh
> [ "$PERF_MAX_ATTEMPTS" -lt 10 ] || die "guard_perf_max_attempts: PERF_MAX_ATTEMPTS=$PERF_MAX_ATTEMPTS is not lower than the sync_jobs entity default (10) - it would cap nothing"
> ```
> Either apply the same treatment (refuse a `GENERATOR_VU_UTILISATION_MAX` ≥ 1.0 or a `GENERATOR_DROPPED_RATIO_MAX` ≥ some sane bound), or record both resolved values in `manifest.json` so a discard-free run can be told apart from a defeated guard. Failing that, soften the docblock — a comment asserting a property the code does not have is worse than no comment, because the next reader will trust it.

**[IMPORTANT]** — `perf/openlinker-throughput/README.md:165-172`
> **The README's post-guard contract is not updated.** That file carries the canonical table of every post-guard (one row each for `post_guard_attempts` … `post_guard_limiter_degraded`) followed by the `run_post_guards` signature:
> ```
> `run_post_guards <dir> <conn_ids> <window_start_iso> <window_start_epoch> <window_stop_epoch> [destination_conn_id]`
> ```
> The new guard gets no row, and the signature is now six arguments where the code takes seven. A scenario author writing the next scenario reads the README, not `lib.sh`, and will omit the summary path — which is the exact silent-skip the diff's own `run_post_guards` docblock warns about ("Omitting it for a k6 scenario silently skips the generator check"). One table row and one signature edit.

**[SUGGESTION]** — `perf/openlinker-throughput/lib.sh:1072`
> The `run_post_guards` header comment line still shows the six-argument signature; the seventh is only described in the prose below it. Worth extending the signature line itself so a reader skimming the header sees the real arity.

**[SUGGESTION]** — `perf/openlinker-throughput/scenarios/f3-webhook-burst.sh:500,511`
> `summary="$dir/k6-summary.json"` is already in scope at line 500 and passed to `run_k6`; line 511 re-spells the same path literally. Passing `"$summary"` keeps one spelling of one path in one function — cheap insurance against the two drifting if the filename ever moves.

---

## Positive observations

- The missing-file branch is placed **before** the `jq` call, and the test asserts on the specific `"wrote no summary at"` text rather than on `DISCARDED`. Both halves matter: the fallback `|| echo "DISCARDED … could not parse"` would otherwise have made the branch untestable, which is what the red-first pass caught.
- `(($dropped + $reqs) == 0)` is guarded before the division — a run that produced neither is reported by the vus arm rather than dividing by zero.
- The F2 call site documents its own omission in place. That is the discipline that keeps "no generator" and "forgot to pass it" distinguishable, and it is exactly why the F5 gap above stands out.

---

## Merge readiness

🔄 **Approve with changes** — no BLOCKING issues; the guard is additive, correct where it runs, and cannot make an existing run worse. The three IMPORTANT items are about the guard's reach and its documented contract, not its logic.

### Priority fixes
1. Wire `post_guard_generator_saturated` into F5's verdict block, or state in place why it is knowingly absent there — and correct the PR description's "F2 and F7" (there is no F7; F5 is the omitted k6 scenario).
2. Bound or manifest the two thresholds, or soften the "no opt-out" claim so the docblock and the code agree.
3. Add the README post-guard row and update the `run_post_guards` signature to seven arguments.

Issue #2607, opened 2026-08-27
[BUG] Integration/PrestaShop — hardcoded order-state map 1-7


norbert-kulus-blockydevs:
## Problem / Context

The adapter carries a hardcoded map of PrestaShop order states 1-7 from a default installation, including the assumption that cancelled is id 6. On any shop with custom order states that is simply untrue, and the mapping fails silently rather than loudly.

The real flags live on `ps_order_state` (`paid`, `shipped`, and the cancellation flag), so the mapping does not have to be guessed.

## Proposed Solution

Read the order states from the shop and map by their actual flags instead of by id. Surface the resolved list to the mapping-configuration UI so an operator can see what was matched.

## Classification

**Type**: Integration
**Layer**: Infrastructure
**File(s)**: `libs/integrations/prestashop/src/infrastructure/adapters/**` (order status mapping), `apps/api/src/integrations/**` (mapping options)

## Acceptance Criteria

- [ ] No hardcoded state ids remain on the PrestaShop status path
- [ ] Paid / shipped / cancelled are derived from the shop's own flags
- [ ] A shop with custom states maps correctly
- [ ] Tests added or updated for non-trivial logic

norbert-kulus-blockydevs:
Done in the epic branch `epic/2590-prestashop-no-module`, reviewed and fixed on PR #2627. Closing so the board reflects it; the branch is merged to `main` as one PR.

Issue #2370, opened 2026-08-22
[TASK] CORE — return receive and dispose writes, counter invariants and restock_blocked


piotrswierzy:
> Part of the OMS programme backlog (**Wave 2**, stream S2, size L). Backlog slug: `W2-33`. Sibling `W2-*` slugs resolve via the Wave-2 epic's checklist. Design of record: returns product spec §5.2/§5.4 (PR #2318); stories T4/T5.

> **Size justification (L):** two write paths, the counter invariant enforced server-side, the master write with its idempotency key, the `restock_blocked` fact, the operator attestation path, and the trigger emissions for T6/T7 — all on one aggregate whose counters must never disagree with the disposition rows.

## Problem / Context

Stories T4 and T5. Wave 1c persists counters that nothing drives. The design's commitment is that restock lands in the **authoritative inventory book** — or blocks loudly. *A restock that silently no-ops is worse than none.*

## Proposed Solution

Two writes on the returns aggregate:

- **receive**(line, quantity, note) — raises `quantityReceived`, moves custody to `received`.
- **dispose**(line, quantity, `restock | scrap`, note) — raises `quantityRestocked` or `quantityScrapped`; on `restock`, calls `InventoryMasterPort.adjustInventory` with a deterministic idempotency key (`return:{returnId}:{lineId}:{seq}` — never wall-clock) and `reason: 'return_restock'`. A refusal records **`restock_blocked`** on the line, persists reason RB-L (#2352 / `W2-15`) and surfaces it; it does **not** roll back the disposition — the goods really were disposed, the book write is what failed.

`advised ≥ received ≥ restocked + scrapped` is enforced server-side, above the Wave-1c DB `CHECK`, with an actionable error code on over-receipt. Marketplace-side stock is **never** written here — the ordinary propagation fan-out carries the master adjustment onward.

Also here: `markStockHandledManually` — an operator attestation recording `restockedBy: 'operator_out_of_band'` for units the master refused, which clears `restock_blocked` without ever claiming OL wrote the stock.

Emits automation triggers T6 (`return.received`) and T7 (`return.disposed`).

## Classification

**Type**: CORE
**Layer**: Application (**migration-bearing** — the `restock_blocked` state column; synthetic-sequential timestamp per `docs/migrations.md` rule 3)
[...]

Issue #2240, opened 2026-08-20
[TASK] Frontend — bulk offer wizard: product-level category, honest blocker copy, seller-details preflight


norbert-kulus-blockydevs:
## Problem / Context

A three-variant product ("Doniczka ceramiczna Terra") lists one variant on Allegro and blocks the other two with a single chip reading `manual category`. Diagnosed with three independent reviews; the findings are collected in an artifact ([copy deck + clickable flow sketch](https://claude.ai/code/artifact/fd99b0c7-e06b-4ec8-8433-942d5ba7f934)).

Five separate defects, all in the bulk offer wizard frontend:

1. **One message for four causes.** `no-match` renders as `manual category` (`apps/web/src/features/listings/components/bulk/bulk-blockers.ts:19`). It covers a barcode absent from the destination catalogue, a barcode failing its check digit, a missing barcode, and a *failed request to the marketplace* — the adapter collapses an HTTP failure into `no-match` (`libs/integrations/allegro/src/infrastructure/util/resolve-categories-for-batch-by-ean.ts:231`, catch at `:310`). The copy also asserts the barcode is the only source of the category, which is false: a miss falls through to the operator's configured category mapping (`libs/core/src/listings/application/services/category-resolution.service.ts` `applyMappingFallback`).

2. **The only fix offered cannot clear the block.** `recomputeVariantBlockers` derives the category blocker from `variant.override.overrides?.categoryId ?? variant.resolvedCategoryId` (`bulk-policy.ts:522`) and never falls back to the product tier, while `handleSubmit` pins `row.override.overrides?.categoryId ?? row.resolvedCategoryId` for a multi-variant product (`bulk-wizard.tsx:438`). The base override is folded into variants only for a single-variant product (`bulk-wizard.tsx:361-368`). So the banner's `Fix on base` button (`bulk-edit-modal.tsx:2309`, wired at `:1176`) switches to a scope whose category the readiness check never reads. The working remediation — the per-variant override — sits behind a collapsed `<details>` labelled "Override base title / description / category" (`bulk-edit-modal.tsx:2327`) and is not what the banner recommends.

   The same asymmetry has a second victim: `noCardCategoryIds` is built from the variant tier only (`bulk-wizard.tsx:183`), so a category pinned at the product level never fetches the required-parameter schema and `allegro:needs-product-parameters` cannot fire — the row reads ready and the create returns 422.

[...]

PR #2939, opened 2026-09-06
fix(prestashop): cron controller exits on success; guard stale combination writes

### LINKED ISSUE #2923: [BUG] Integration — the PrestaShop cron controller answers 500 on a successful delivery
## Problem / Context

`apps/prestashop-module/openlinker/controllers/front/cron.php` returns **HTTP 500 on a successful outbox delivery**. Observed live during the #2848 measurement, against a real shop, with the correct token:

```
$ curl -sS -X POST '.../index.php?fc=module&module=openlinker&controller=cron' \
       -H 'X-OpenLinker-Cron-Token: <valid>'
{"processed":0,"delivered":0,"failed":0,"requeued":0,"skipped":0,
 "budget_exhausted":false,"budget_seconds":120,"retention":{...}}
HTTP 500
```

The body is a correct success payload. The status is 500. **The work is done and the caller is told it failed.**

The cause is an asymmetry in the controller: every early-return branch ends in `exit;` (`:52`, `:67`, `:80`, `:86`), and the success branch does not. After `echo json_encode($stats)` control returns to PrestaShop's ordinary front-controller pipeline, which renders a full storefront page. On this stand the render dies:

```
FrontControllerCore->display() -> getStylesheets() -> CccReducerCore->reduceCss()
  -> CssMinifierCore::minify()
  -> Fatal: theme-3f744e.css could not be opened for writing
```

**On a shop where the theme cache IS writable the failure is different and arguably worse**: the render succeeds, so the response body becomes the JSON payload followed by an entire HTML page, and any strict JSON parse of that endpoint fails. The 500 seen here is a symptom of an unwritable cache directory; the missing `exit;` is the defect, and it is present on every install.

Operator impact: an uptime check or cron wrapper watching this endpoint reports a permanent failure while the outbox drains correctly, and any client parsing the body strictly breaks on a healthy shop.

## Proposed Solution

Add `exit;` after the success `echo json_encode($stats)`, matching the four branches that already have it. Consider whether an explicit `http_response_code(200)` belongs there too, for symmetry with the `403`/`405`/`500` branches, though PrestaShop defaults to 200 when nothing sets otherwise.

While there: the `catch` branch (`:97`) already sets 500 and does **not** `exit` either, so it has the same trailing-render problem on a shop that can render.

## Classification

**Type**: Integration
**Layer**: Interface (module front controller)
**File(s)**: `apps/prestashop-module/openlinker/controllers/front/cron.php`

## Dependencies

- None.

## Assumptions

- The success branch is intended to terminate the request, like the other four. Nothing downstream of it in the module needs to run.
- The CLI entry point (`apps/prestashop-module/openlinker/cron/openlinker-cron.php`) is unaffected; this is the HTTP path only.

## Acceptance Criteria

- [ ] A successful delivery through the HTTP cron endpoint answers **HTTP 200** with a body containing only the JSON payload - no HTML
- [ ] The failure branch likewise answers 500 with only its JSON payload
- [ ] Asserted on a shop whose theme cache IS writable, so the trailing-render path is actually exercised rather than masked by a fatal
- [ ] The module's own integration tests cover both branches' status and body
- [ ] No architecture boundary violations (CORE - Integration)

### LINKED ISSUE #2925: [BUG] Integration — a stock write at id_product_attribute=0 silently no-ops for combination products
## Problem / Context

A stock write addressed to `id_product_attribute = 0` **silently does nothing** for a product that has combinations. PrestaShop treats the product-level row as a recomputed aggregate of its combination rows and discards the write.

Found while raising stock across a demo catalogue for the #2848 measurement: **25 writes through the PrestaShop webservice produced only 2 `stock.changed` outbox rows.** The 23 that produced nothing were all product-level writes against combination products. Nothing errored - the webservice answered, the value appeared to be accepted, and the quantity did not change.

This matters beyond the harness. Any OpenLinker path that writes master stock at product grain against a combination product is a no-op, and it is a *silent* one: no exception, no warn, no failed job. From the outside it is indistinguishable from a write that landed.

`docs/architecture-overview.md` already records that per-combination stock is the model (#823 - `MasterInventorySyncService` iterates `InventoryMasterPort.listInventory(productId)` and writes one variant-keyed row each), so the read side is correct. This is about the **write** side, and about tooling that seeds or corrects stock.

## Proposed Solution

Establish first whether any production write path can actually reach this shape - `PrestashopInventoryMasterAdapter`'s `adjustInventory` is the obvious candidate, and it may already resolve a combination id. If it can, the write should either address the combination row or refuse loudly rather than no-op.

Independently of that, the condition deserves a named guard somewhere a developer will meet it: a stock write to a combination product at attribute 0 is a programming error, not a runtime condition, and should say so.

Also worth a line in `docs/lessons.md` - this is exactly the shape that ledger exists for, and it cost a measurement run before it was understood.

## Classification

**Type**: Integration
**File(s)**: `libs/integrations/prestashop/src/infrastructure/adapters/prestashop-inventory-master.adapter.ts`, `docs/lessons.md`

## Dependencies

- None.

## Assumptions

- The behaviour is PrestaShop's, not the module's - it was observed through the raw webservice, with no OpenLinker module involvement.
- Simple products (no combinations) are unaffected: their product-level row is the real row.

## Acceptance Criteria

- [ ] It is established, from the code rather than by inference, whether any OpenLinker write path can address a combination product at attribute 0
- [ ] If one can, it either addresses the combination row or fails loudly - never silently no-ops
- [ ] An entry in `docs/lessons.md` describing the shape, so the next person seeding or correcting stock does not lose a run to it
- [ ] No architecture boundary violations (CORE - Integration)

norbert-kulus-blockydevs:
## Summary

Two self-contained PrestaShop-module defects found by the #2848 measurement campaign.

**#2923** - the module's HTTP cron front controller (`controllers/front/cron.php`) ended its success branch (and its catch branch) without `exit;`, unlike the four early-return branches above it. Control therefore fell through into PrestaShop's ordinary front-controller page render after the JSON was echoed. On a shop whose theme cache is unwritable this fatals into a 500 over a correct success payload; on one whose cache is writable, the response body becomes the JSON payload followed by an entire HTML document, which breaks any strict JSON parser watching the endpoint. Both branches now `exit;` immediately after their envelope, matching the pattern already used by the four refusal branches.

**#2925** - `PrestashopInventoryMasterAdapter.adjustInventory` already refused to write the product-level (`id_product_attribute=0`) row when no variant was named and the product has combinations, since PrestaShop treats that row as an aggregate it silently recomputes from the combination rows (a write there is accepted and discarded with no signal). That guard had a narrower gap: a variant minted while its product was still simple keeps a synthetic `product:<id>` external-id mapping until the product's next full variant sync replaces it with real per-combination mappings. A caller handing in that variant id during the window between a shop admin adding the product's first combination and OpenLinker's next re-sync of that product's variants would reach the same silent no-op through a resolved `variantId`, bypassing the existing check. The "does this product currently have combinations" probe now runs whenever the resolved target is the product-level row, whichever branch resolved it there - refusing loudly (`PrestashopNotSupportedException`) instead of writing silently.

## Verification

**#2923**, live against the `lab-prestashop` (PrestaShop 9.0.2) stand:
- Before the fix, with the theme cache genuinely writable (it was previously owned by `root`; re-chowned to `www-data` to exercise the real render path rather than mask it behind a permissions fatal): the request still 500s, now failing further into the render pipeline (a Smarty "Missing `$template`" fatal) - confirming the missing `exit;` is the defect regardless of which downstream render step trips first.
- After the fix: `HTTP 200`, `Content-Length: 278`, body verified byte-exact to the 278-byte JSON payload with nothing appended.
- Both branches (`echo json_encode(...)` in the try, and the catch) now carry `exit;`, asserted by two new source-level tests in `CronDeliveryHardeningTest.php` (the module has no PrestaShop runtime available to phpunit for a live controller test, matching the existing tests in that file and in `EndpointHardeningTest.php`).
- Ran the module's own PHPUnit `Unit` suite (207 -> 209 tests) inside the `lab-prestashop` container (`composer install` + `vendor/bin/phpunit --testsuite=Unit`): both new tests pass. Two unrelated, pre-existing failures in `CronHealthTest` reproduce on this environment regardless of this change - the container's PHP default timezone is `Europe/Paris`, and `CronHealth::assess`'s `strtotime()` call interprets a UTC-labelled `gmdate()` string in that local zone, producing a systematic one-hour offset. Untouched by this PR; out of scope for #2923/#2925.

**#2925**: established from the code, not by inference, per the acceptance criteria - traced the one production caller of `adjustInventory` (`ReturnCustodyService.writeMasterStock`, in `libs/core/src/returns`) and confirmed `variantId` is always a variant id freshly resolved via `IProductsService.getVariantsBySkus`, never omitted. The pre-existing guard already covered the "no variantId" branch (tested at `should refuse when a combination product is adjusted without a variantId`). The narrower gap was the resolved-but-stale-mapping case above, closed by the change; a new test (`should refuse rather than silently no-op when a variant carries a stale pre-combination mapping (#2925)`) pins it. Added a `docs/lessons.md` entry recording the shape.

## Test plan

- [x] `pnpm check:invariants` - exit 0
- [x] `libs/integrations/prestashop` package lint (`eslint`) - clean
- [x] Module's own PHPUnit `Unit` suite run live in the `lab-prestashop` container - 209 tests, 2 pre-existing unrelated failures (timezone-dependent `CronHealthTest`, untouched by this PR)
- [x] Live end-to-end verification of #2923 against `lab-prestashop` (PrestaShop 9.0.2), both before/after the fix, with the theme cache genuinely writable
- [x] Traced the one production caller of `adjustInventory` to establish #2925's reachability, per its acceptance criteria

Closes #2923
Closes #2925

### DIFF
--- a/apps/prestashop-module/openlinker/controllers/front/cron.php
+++ b/apps/prestashop-module/openlinker/controllers/front/cron.php
@@ -91,8 +91,10 @@ class OpenLinkerCronModuleFrontController extends ModuleFrontController
         try {
             $stats = DeliveryRunner::run('http');
 
+            http_response_code(200);
             header('Content-Type: application/json');
             echo json_encode($stats);
+            exit;
         } catch (Exception $e) {
             http_response_code(500);
             header('Content-Type: application/json');
@@ -100,6 +102,7 @@ class OpenLinkerCronModuleFrontController extends ModuleFrontController
                 'error' => 'Internal Server Error',
                 'message' => 'Cron delivery failed: ' . WebhookSender::getErrorMessage($e)
             ]);
+            exit;
         }
     }
 }
--- a/apps/prestashop-module/openlinker/tests/Unit/CronDeliveryHardeningTest.php
+++ b/apps/prestashop-module/openlinker/tests/Unit/CronDeliveryHardeningTest.php
@@ -75,6 +75,47 @@ class CronDeliveryHardeningTest extends TestCase
         self::assertStringContainsString('Deny from all', $htaccess);
     }
 
+    /**
+     * #2923: the HTTP cron front controller's success branch had no `exit;`,
+     * unlike its four early-return branches - so a successful delivery fell
+     * through into PrestaShop's ordinary page render, which either fatals
+     * (an unwritable theme cache) or appends a whole HTML document after the
+     * JSON (a writable one). Both the success and catch branches must
+     * terminate the request immediately after their envelope.
+     */
+    public function testTheSuccessBranchExitsAfterItsJsonEnvelope(): void
+    {
+        $source = self::sourceOf('controllers/front/cron.php');
+
+        $runAt = strpos($source, 'DeliveryRunner::run');
+        $echoAt = strpos($source, 'echo json_encode($stats);');
+        $catchAt = strpos($source, 'catch (Exception $e)');
+
+        self::assertIsInt($runAt);
+        self::assertIsInt($echoAt);
+        self::assertIsInt($catchAt);
+        self::assertGreaterThan($runAt, $echoAt);
+        self::assertLessThan($catchAt, $echoAt);
+
+        $betweenEchoAndCatch = substr($source, $echoAt, $catchAt - $echoAt);
+        self::assertStringContainsString('exit;', $betweenEchoAndCatch);
+    }
+
+    public function testTheCatchBranchAlsoExitsAfterItsJsonEnvelope(): void
+    {
+        $source = self::sourceOf('controllers/front/cron.php');
+
+        $catchAt = strpos($source, 'catch (Exception $e)');
+        self::assertIsInt($catchAt);
+
+        $afterCatch = substr($source, $catchAt);
+        $echoAt = strpos($afterCatch, 'echo json_encode([');
+        self::assertIsInt($echoAt);
+
+        $afterEcho = substr($afterCatch, $echoAt);
+        self::assertStringContainsString('exit;', $afterEcho);
+    }
+
     private static function sourceOf(string $relativePath): string
     {
         $path = __DIR__ . '/../../' . $relativePath;
--- a/docs/lessons.md
+++ b/docs/lessons.md
@@ -23,6 +23,46 @@ When a lesson hardens into a rule, **graduate it** to the canonical doc and leav
 
 ---
 
+## A raw stock write to `id_product_attribute=0` on a PrestaShop combination product is a silent no-op
+
+**Context**: raising stock across a demo catalogue for the #2848 measurement, via 25 raw
+`stock_availables` PUTs through the PrestaShop webservice. Only 2 produced a `stock.changed` outbox
+row; the other 23 were product-level writes (`id_product_attribute=0`) against products that have
+combinations.
+
+**Problem**: PrestaShop treats the `id_product_attribute=0` row on a combination product as an
+aggregate it recomputes from the combination rows, not as a real, independently-writable row. A PUT
+against it is accepted (200, no error) and silently discarded - the quantity does not change, and
+nothing in PrestaShop or in a naive caller reports that. Seeding or correcting stock by writing the
+product-level row is indistinguishable, from the outside, between "it worked" and "it was thrown
+away".
+
+`PrestashopInventoryMasterAdapter.adjustInventory` already guards the OpenLinker write path against
+this (`resolveTargetAttributeId` refuses with `PrestashopNotSupportedException` rather than writing
+attribute 0 on a product it can see has combinations) - but that guard only covered the *caller
+named no variant* case. A second, narrower gap survived: a variant minted while its product was
+still simple keeps a synthetic `product:<id>` external-id mapping until the product's next full
+variant sync replaces it with real per-combination mappings; a caller handing in that variant id
+between "the shop admin adds the first combination" and "that re-sync completes" reached the same
+silent no-op through a resolved `variantId` instead of an absent one. Fixed in #2925 by running the
+same "does this product currently have combinations" probe whenever the resolved target is the
+product-level row, whichever branch resolved it there.
+
+**Rule**: when writing PrestaShop stock at product grain, never trust a cached/mapped "this is a
+simple product" fact without re-checking it against a live combinations read at write time - a
+product can gain combinations after the mapping that named it was minted, and the shop will accept
+and discard the write with no signal. A tool seeding stock for measurement or QA should always write
+at the resolved combination (`id_product_attribute`) level, never at `0`, unless it has just verified
+the product carries no combinations.
+
+**Applies to**: any script or adapter path that writes `stock_availables` directly or via
+`InventoryMasterPort.adjustInventory` — `libs/integrations/prestashop/src/infrastructure/adapters/prestashop-inventory-master.adapter.ts`,
+and any `perf/**` seeding/correction tooling that raises stock through the raw webservice.
+
+**Source**: #2925, found during the #2848 measurement.
+
+---
+
 ## Before a surface asserts a behaviour, read the code that implements it
 
 **Context**: redesigning the three sales-document surfaces (#2513). The design was worked out from
--- a/libs/integrations/prestashop/src/infrastructure/adapters/__tests__/prestashop-inventory-master.adapter.spec.ts
+++ b/libs/integrations/prestashop/src/infrastructure/adapters/__tests__/prestashop-inventory-master.adapter.spec.ts
@@ -1161,6 +1161,42 @@ describe('PrestashopInventoryMasterAdapter', () => {
       );
     });
 
+    it('should refuse rather than silently no-op when a variant carries a stale pre-combination mapping (#2925)', async () => {
+      // The variant was mapped as `product:<id>` before the shop admin added
+      // combinations. The next full variant sync would replace that mapping
+      // with a real per-combination one, but until it does, this write must
+      // never reach the recomputed-aggregate row silently.
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment -- test mock: narrowing dynamic spy / fixture / response shape
+      mockIdentifierMapping.getExternalIds = jest
+        .fn()
+        .mockImplementation((entityType: string) =>
+          entityType === 'ProductVariant'
+            ? Promise.resolve([
+                {
+                  connectionId: connection.id,
+                  externalId: `product:${PS_PRODUCT_ID}`,
+                  entityType: 'ProductVariant',
+                },
+              ])
+            : Promise.resolve([
+                { connectionId: connection.id, externalId: PS_PRODUCT_ID, entityType: 'Product' },
+              ])
+        );
+      respondWithRows([
+        simpleStockRow(10),
+        { ...simpleStockRow(4), id: '202', id_product_attribute: '77' },
+      ]);
+
+      await expect(
+        adapter.adjustInventory({
+          productId: PRODUCT_ID,
+          variantId: 'internal-variant-now-stale',
+          quantity: 1,
+        })
+      ).rejects.toThrow(PrestashopNotSupportedException);
+      expect(mockHttpClient.updateResource).not.toHaveBeenCalled();
+    });
+
     describe('refusals', () => {
       it('should refuse and name shopId when several shops match the row', async () => {
         respondWithRows([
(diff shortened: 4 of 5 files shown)


piotrswierzy (approved):
Both fixes are correct and well-scoped, and the #2925 guard is the right shape — a live probe rather than an inference from the mapping, refusing loudly rather than skipping, with the refusal landing on the operator as a `restock_blocked` act carrying the adapter's own sentence (§22 / #2370), so no stock drift is created.

I checked both tests are genuinely red against the unfixed code rather than green for the wrong reason. The old `return mapping.externalId.startsWith('product:') ? '0' : mapping.externalId;` returns before any probe runs, so `updateResource` *is* called and nothing throws — both Jest assertions fail. And neither unfixed cron branch carries an `exit;` inside the bounded substrings the PHP tests inspect; note the file-top `if (!defined('_PS_VERSION_')) { exit; }` at line 34 sits *before* `$catchAt`, so it can't accidentally satisfy test 2 — that substring bound is doing real work. `respondWithRows` uses `mockResolvedValue`, which persists across both `listStockRecords` calls, so the probe really does see the combination row.

Approving. Two things I'd like settled, one before merge if you can:

**`docs/lessons.md` — the Rule overstates the guard.** It says "re-checking it against a **live combinations read** at write time", but `resolveTargetAttributeId` probes `stock_availables` (`allRows.some(r => Number(r.id_product_attribute) !== 0)`), not the `combinations` resource that `PrestashopProductMasterAdapter.getProductVariants` reads. A combination that exists without a `stock_available` row yet answers "no combinations", and the write lands on attribute 0 — the same silent no-op, unguarded. Either read `combinations` (the authoritative fact the lesson names) or narrow the wording to what's actually checked. A lessons entry that overstates a guard is worse than none, because the next author trusts it.

**`controllers/front/cron.php:97` — `catch (Exception $e)` is narrower than what `DeliveryRunner::run` throws.** `DeliveryRunner.php:124` says so verbatim — *"Throwable, not Exception: a TypeError in the delivery loop is an Error"* — and rethrows it. That `Error` sails past the catch, so the request dies as a PHP fatal with a non-JSON body and **neither new `exit;` runs**: the same defect class this PR closes, on the one path it leaves open. `importorder.php:76` already ships the in-repo remedy (a `register_shutdown_function` guard turning a fatal into a 502 JSON envelope); reuse it, or widen to `catch (Throwable $e)` if the module's PHP floor allows.

Suggestions, none blocking:

- `prestashop-retry-classifier.adapter.ts:159` — `PrestashopNotSupportedException` isn't in `isNonRetryable`. "This product has combinations" is a persisted-state fact no retry can change, the same reasoning that put `PrestashopOrderStateUnresolvedException` there (#2607 / ADR-007). Latent while `ReturnCustodyService.writeMasterStock` is the only caller and catches `unknown`; the moment a second caller lets it reach a job handler it burns the full ladder to a dead row.
- `CronDeliveryHardeningTest.php:104` — `$afterEcho = substr($afterCatch, $echoAt)` runs to EOF, so an `exit;` anywhere later in the file satisfies it. Bound it to the catch block's closing brace, as test 1 bounds itself with `$catchAt`. Both tests also match `exit;` as raw text, so a commented-out one would pass — fine for a source-level test, worth a word in the docblock.
- `prestashop-inventory-master.adapter.ts:1032,1062` — `'0'` is both the "unresolved" sentinel and a possible mapping value, so a mapping whose `externalId` is literally `'0'` would fall into the probe and refuse with the *wrong* message ("an adjustment must name which variant it applies to") when the caller did name one. Safe today since PS combination ids are ≥ 1, but a `targetIsProductLevel` boolean removes the coupling for nothing.
- Module delivery is unstated: `openlinker.php:120` pins `1.10.0` and the module carries an `upgrade/upgrade-X.Y.Z.php` convention through 1.9.0. No schema or hook change is needed here, but PS's admin update flow keys on the version string — one line on how an operator on 1.10.0 receives this, or a bump to 1.10.1.
- The `lessons.md` heading is declarative; the file's own format note (line 17) asks for imperative — "Never write PrestaShop stock at `id_product_attribute=0` without re-checking combinations" rather than a statement of the symptom.

Nice touches worth naming: the early `return attributeId` keeps the extra `stock_availables` read off the normal per-combination path (easy to get backwards by probing first); the two refusal messages and remedies are branched, so a caller who *did* supply a variantId isn't told to supply one; a genuinely simple product with a synthetic mapping still passes the probe and writes, so the mirror isn't stricter than the gate (#2240); and the live verification deliberately re-chowned the theme cache to `www-data` so the render path ran for real rather than being masked by a permissions fatal. I also checked the other three front controllers — the remaining `echo json_encode` sites without `exit` (`cartshipping.php:187`, `importorder.php:519`) are inside shutdown guards where no exit is wanted, so the fix is complete across the module.

piotrswierzy (approved):
## Summary

Two small, well-scoped PrestaShop defects, both with a clear failure mode and both fixed at the right layer. The cron `exit;` is a one-line correction that makes the success and catch branches match the four refusal branches already in that file; the `adjustInventory` change closes a genuinely narrow staleness window in `resolveTargetAttributeId` without touching any healthy path. I verified both against the surrounding code rather than taking the description on trust, and found nothing blocking.

---

## Verification I ran

**#2923 — the fix covers the class, not just the instance.** I checked every JSON-emitting branch across all four front controllers (`cartshipping.php`, `cron.php`, `importorder.php`, `ping.php`). `cron.php`'s success and catch branches were the *only* two lacking `exit;`; every other `echo json_encode(...)` already terminates (`cartshipping.php:235,254`, `importorder.php:546,602`, `ping.php:122`), and the two that don't (`cartshipping.php:187`, `importorder.php:519`) are `register_shutdown_function` bodies where an `exit` would be wrong. So this is the complete set — good.

**#2925 — the refactor is behaviour-preserving on every path but the broken one.** Tracing `resolveTargetAttributeId` after the change (`prestashop-inventory-master.adapter.ts:1035-1092`):
- Real per-combination mapping → `attributeId = mapping.externalId` → returns early at `if (attributeId !== '0')`. No new HTTP call, byte-identical to before.
- No `variantId` → unchanged; still falls through to the combinations probe.
- Stale `product:<id>` mapping **and** the product has combinations → now throws `PrestashopNotSupportedException` instead of returning `'0'` and writing into the recomputed aggregate. This is the fix.
- Stale `product:<id>` mapping **and** the product genuinely has no combinations → falls through the probe to `return '0'` — identical to the old behaviour, correctly.

The one edge worth naming and dismissing: a real combination mapping whose `externalId` were the literal `'0'` would fall into the probe branch. `id_product_attribute = 0` is PrestaShop's sentinel for *no combination*, so no real combination mapping can carry it — and if one somehow did, the probe refuses, which is the safe direction. No issue.

---

## Findings

**[SUGGESTION]** — `libs/integrations/prestashop/src/infrastructure/adapters/prestashop-inventory-master.adapter.ts:1064-1066`
> `resolvedViaStaleSyntheticMapping` cleanly separates the two refusal messages, and the remedy text ("Re-run the product/variant sync … then retry") is the right instruction. One thought for a future reader: the flag is only ever `true` when `attributeId === '0'` *and* a `variantId` was supplied, so the state space is small enough today — but a third resolution branch added later would have to remember to set it. A short comment on the declaration stating the invariant ("true iff a variantId resolved to the product-level row") would make that obligation explicit.

**[SUGGESTION]** — `apps/prestashop-module/openlinker/controllers/front/cron.php:94`
> The added `http_response_code(200)` is a no-op (200 is PHP's default) — but it makes the success branch structurally symmetric with the 403/405/500 branches, which read better as a set. Worth keeping; noting only so a future reader doesn't mistake it for load-bearing.

**[SUGGESTION]** — `apps/prestashop-module/openlinker/tests/Unit/CronDeliveryHardeningTest.php:86-113`
> The two new tests are source-text assertions, consistent with the existing tests in this file and in `EndpointHardeningTest.php` (the module has no PrestaShop runtime under phpunit), so the approach is right for this codebase. `testTheCatchBranchAlsoExitsAfterItsJsonEnvelope` asserts only that `exit;` appears *somewhere* after the catch's `echo` — which is true of the whole rest of the file. Narrowing it the way the success-branch test does (bounded by the next structural landmark, e.g. the closing of the method) would make it fail if the `exit;` migrated out of the catch block. Low risk given the file's size, but the sibling test is the stricter one and it costs little to match it.

**[Documentation]** — `docs/lessons.md`
> The entry follows the established format (Context / Problem / Rule / Applies to / Source), states the rule in the imperative, and — importantly — names the *tooling* consequence as well as the adapter one ("A tool seeding stock for measurement or QA should always write at the resolved combination level"). That is the half that would otherwise be relearned by the next perf campaign. Good.

---

## Merge readiness

✅ **Approve** — no BLOCKING or IMPORTANT findings. The scope discipline is right on both halves: `cron.php` fixes only the two branches that were wrong, and the adapter change adds one condition to an existing guard rather than restructuring the resolution chain. The verification narrative in the description — reproducing #2923 with the theme cache deliberately re-chowned to `www-data` so the render path ran rather than being masked behind a permissions fatal, and tracing the single production caller of `adjustInventory` to establish #2925's reachability — is the standard of evidence this repo asks for, and the pre-existing timezone-dependent `CronHealthTest` failures are correctly identified as out of scope rather than papered over.

The three SUGGESTIONs are all optional and none needs to block the merge.

PR #2940, opened 2026-09-06
feat(perf): re-assert the volatile guards per window, and record what shaped a run

### LINKED ISSUE #2932: [TASK] DX — re-assert the volatile guards at each window, not once per script
## Problem / Context

Two scenarios measured the `lab` stand concurrently and **silently invalidated each other**.

F3 declares `guard_runner_state disabled`; F2 needs the runner enabled and flips it. F3's guards passed at 23:27, F2 flipped `lab-worker` at 23:30:24, and F3's `replay-committed` (23:30:49) and `replay-concurrent` (23:34:38) arms then carried `"runnerState": "disabled"` in their manifests for windows in which the runner was executing jobs. `post_guard_attempts` caught one of them - `5 job(s) in the window show attempts>1` - and **nothing would have caught the other**.

`guard_stand_exclusive` (added in the same campaign) closes the two-scenario case: it claims the stand in Redis with `SET NX` and refuses rather than queueing. But it is not the whole class.

**Every pre-flight guard describes the world at script start, and `manifest_write` runs per arm, re-rendering values that were verified once.** A peer with write access to the stand can falsify any of them before a later window opens - and that peer need not be another scenario. An operator, an unrelated automation, or a plain `docker compose up --force-recreate` silently resolving a different default will all do it, and none of them takes the lock.

## Proposed Solution

Re-assert the **volatile** guards at each `window_start`, not once per script:

- `guard_runner_state` - the worker's `WORKER_RUNNER_ENABLED` and its lane caps
- `guard_scheduler_off` - the worker's `OL_SCHEDULER_ENABLED`

`guard_queue_empty` already does this per arm, so the shape exists to copy.

The non-volatile ones stay where they are: `guard_build`, `guard_demo_mode_off`, `guard_connection_budget`, `guard_log_level` and `guard_perf_max_attempts` describe things that do not change under a running scenario, and re-checking them every window buys noise rather than safety.

The distinction worth encoding in the library's own comments is **volatile versus fixed**, so the next guard added lands on the right side of it deliberately.

## Classification

**Type**: DX
**File(s)**: `perf/openlinker-throughput/lib.sh`, `perf/openlinker-throughput/lib-test.sh`

## Dependencies

- None. `guard_stand_exclusive` already exists and is complementary, not a substitute.

## Assumptions

- Re-asserting two guards per arm is cheap - each is a `docker exec printenv` plus, for the runner, one `docker logs | grep`. On a three-arm scenario that is six extra calls, which is negligible against a 165-second window.

## Acceptance Criteria

- [ ] `guard_runner_state` and `guard_scheduler_off` are re-asserted at each `window_start`, and a mid-run change aborts the run rather than being recorded as the start-of-script value
- [ ] The manifest's `runnerState` and lane caps describe the window they label, not an earlier check
- [ ] `lib.sh` states which guards are volatile and which are fixed, and why
- [ ] `lib-test.sh` covers the mid-run flip: a guard that passed at start and fails at window two must abort - verified red-first

### LINKED ISSUE #2934: [TASK] DX — stand conditions set at runtime must reach the compose file or the manifest
## Problem / Context

A `statement_timeout = 30s` was set on the `lab` stand's `postgres` role with a live `ALTER ROLE` during the #2843 campaign, to bound runaway reads. It is real - `SELECT rolname, rolconfig FROM pg_roles` returns `{statement_timeout=30000ms}` - and it was in force for every F5 measurement.

**It exists nowhere except in the running database.** It is not in `docker-compose.lab.yml`, so it does not survive a stand rebuild; it is not in any run's manifest, so no report records that it applied; and nothing would have flagged its absence on a re-run.

For F5 it was benign - the reads it bounded take 149 ms, three orders of magnitude under the timeout. That is exactly why it is worth fixing now: a condition that silently does nothing today is one nobody checks tomorrow, and the next scenario it does affect will produce a figure whose conditions cannot be reconstructed.

It also caused a real failure while it existed: the seeder shared the same role, and the 1M-order seed was cancelled mid-transaction until it set `SET statement_timeout = 0` for its own session.

The campaign's whole discipline is that a run's conditions are stated rather than inherited. Nine pre-flight guards enforce that for the conditions they know about. This one was invisible to all of them.

## Proposed Solution

Two halves, and the second matters more than the first.

**Move the setting where it belongs.** Either into `docker-compose.lab.yml` as a `-c statement_timeout=...` on the Postgres command, alongside `max_connections` and the `pg_stat_statements` configuration already there, or drop it if nothing needs it. Either way it stops being live-only state.

**Capture the settings that shape a measurement into the manifest**, so a report is self-describing even when someone has changed something by hand. `manifest_gather_environment` already records `shared_buffers`, `work_mem`, `shared_preload_libraries` and the installed extensions - the gap is role-level and database-level `SET`s, which override those and are currently invisible. `pg_settings` with `source <> 'default'` is one honest way to capture the lot rather than guessing which knobs matter.

A guard refusing unexpected settings would be the wrong instrument here: the set of things an operator might legitimately tune is open, so recording beats refusing. What must not happen is a figure whose conditions cannot be reconstructed from its own manifest.

## Classification

**Type**: DX
**File(s)**: `docker-compose.lab.yml`, `perf/openlinker-throughput/lib.sh` (`manifest_gather_environment`)

## Dependencies

- None.

## Assumptions

- The 30-second timeout was added deliberately during F5 rather than being a pre-existing stand property. If it predates that work, the same conclusion holds - it still is not written down anywhere.

## Acceptance Criteria

- [ ] The `statement_timeout` is either declared in `docker-compose.lab.yml` or removed, and the stand's documented setup states which
- [ ] The manifest records every non-default Postgres setting in force, not a hand-picked subset
- [ ] Role-level and database-level overrides are captured, since those are what a `SHOW` from a different role would miss
- [ ] A stand rebuilt from the compose file alone produces a manifest identical in these fields to the current one, or the difference is explained

norbert-kulus-blockydevs:
Closes #2932. Closes #2934. Based on **`2933-generator-saturation-guard`** (#2938), which reaches the epic branch through #2922 - review against the stack tip.

Two ways a measurement could state conditions it had not established.

## A guard describes the world at script start

F3's guards passed at 23:27 with the runner disabled. F2 flipped it at 23:30:24. Two of F3's later arms then carried `"runnerState": "disabled"` in their manifests **for windows in which the runner was executing jobs**. `post_guard_attempts` caught one of them; nothing would have caught the other.

`guard_stand_exclusive` stops a peer *scenario*. It cannot stop an operator, an automation, or a `docker compose up --force-recreate` silently resolving a different default.

So `guard_runner_state` and `guard_scheduler_off` now record what they verified, and `reassert_volatile_guards` re-runs whichever of the two a scenario **actually declared** - never inventing one it did not, so F5 gets no runner check it never asked for.

`window_start` calls it as its **first** action, before `manifest_write` and before the sampler starts. A flip therefore aborts the run rather than being recorded, and the manifest that does get written reflects a fresh check rather than a stale one.

**The volatile/fixed split is written into `lib.sh`, not left implied**, so the next guard added lands on the right side deliberately:

| | |
|---|---|
| **Volatile** - a live container property a peer can flip mid-run | `guard_runner_state`, `guard_scheduler_off` |
| **Fixed** - a boot-time arg or env, immutable without recreating the container | `guard_build`, `guard_demo_mode_off`, `guard_connection_budget` / `guard_pool_recorded`, `guard_log_level`, `guard_perf_max_attempts` |
| **Orthogonal** - a data precondition, already re-checked per arm | `guard_queue_empty` |

Re-checking the fixed ones every window would buy noise rather than safety.

## And a condition can shape a run while existing nowhere but the running database

`statement_timeout = 30s` was set on the lab stand's `postgres` role by `f5-read-path.sh` through `ALTER ROLE`. It was in force for every F5 measurement, and it appears in **neither** the compose file nor any manifest.

It was benign there - the reads it bounded take 149 ms - which is exactly why it was worth fixing before it was not.

It moves into `docker-compose.lab.yml` beside the settings already declared there, and the manifest now records **every** non-default Postgres setting instead of a hand-picked subset: `pg_settings WHERE source <> 'default'`, plus `pg_roles.rolconfig` and `pg_db_role_setting` - the role- and database-level overrides a plain `SHOW` from another role would miss. A malformed answer degrades to a named error object rather than breaking `manifest.json`.

**A guard refusing unexpected settings would have been the wrong instrument.** What an operator may legitimately tune is an open set, so recording beats refusing. What must not happen is a figure whose conditions cannot be reconstructed from its own manifest.

## Tests

Thirteen new assertions, all verified red-first by the author and **independently re-checked** before merge. Neutering `reassert_volatile_guards` fails four:

```
- runner flipped enabled after guard_runner_state disabled passed -> reassert dies
- scheduler flipped on after guard_scheduler_off passed -> reassert dies
- window_start dies when a volatile guard flips between windows
- no manifest.json was written on the aborted window_start
```

That last one is the assertion that matters: it proves the abort **precedes** the side effect rather than following it.

```
=== 96 passed, 0 failed ===      (was 83)
```

The re-assertion closes the window between `window_start` and the guard's own check. It does **not** cover a flip that happens *during* a window - nothing short of continuous sampling would, and the post-guards are what cover that side. The report should keep saying so.

### DIFF
--- a/docker-compose.lab.yml
+++ b/docker-compose.lab.yml
@@ -55,6 +55,22 @@ services:
     # only plans over its own threshold (2s) to keep the F5 window's own log
     # volume bounded; pg_stat_statements is reset per dataset size by the
     # scenario itself, never here.
+    #
+    # statement_timeout=30s (#2934): this was, until this change, a LIVE-ONLY
+    # `ALTER ROLE postgres SET statement_timeout = ...` applied by
+    # f5-read-path.sh's own apply_statement_timeout() and never declared
+    # anywhere - so it existed only in a running database, did not survive a
+    # stand rebuild, and no run's generic manifest recorded that it was in
+    # force (found #2843/#2934: the same role's own 1M-order seed got
+    # cancelled mid-transaction by it, before the seeder learned to
+    # `SET statement_timeout = 0` for its own session - see seed-lib.sh's
+    # seed_sql()). It bounds the API's runaway-read risk for EVERY session on
+    # this stand now, server-wide, from boot - f5-read-path.sh's own
+    # apply_statement_timeout() still runs (still useful for an operator who
+    # wants a DIFFERENT STATEMENT_TIMEOUT_MS for one F5 run) and simply
+    # re-asserts the same 30s by default via its own ALTER ROLE, which is
+    # captured into every scenario's manifest generically now - see
+    # manifest_gather_environment's `postgres.nonDefaultSettings` in lib.sh.
     command:
       - postgres
       - -c
@@ -71,6 +87,8 @@ services:
       - auto_explain.log_analyze=on
       - -c
       - auto_explain.log_buffers=on
+      - -c
+      - statement_timeout=30000
     ports:
       - '127.0.0.1:${POSTGRES_HOST_PORT:-19432}:5432'
     volumes:
--- a/perf/openlinker-throughput/lib-test.sh
+++ b/perf/openlinker-throughput/lib-test.sh
@@ -93,6 +93,7 @@ pg_sql() {
     *"operational_settings"*) echo "${FAKE_PG[operational_settings]:-{\}}" ;;
     *"COUNT(*)"*) echo "${FAKE_PG[count]:-0}" ;;
     *"pg_database_size"*) echo "${FAKE_PG[db_size]:-123456}" ;;
+    *"pg_db_role_setting"*) echo "${FAKE_PG[pg_overrides]:-}" ;;
     *"string_agg"*) echo "${FAKE_PG[agg]:-}" ;;
     *) echo "" ;;
   esac
@@ -538,6 +539,104 @@ manifest_set_sync_jobs_end "$MDIR"
 assert_eq "syncJobsRowsAtEnd is set after window_stop" "42" "$(jq -r .syncJobsRowsAtEnd "$MDIR/manifest.json")"
 FAKE_PG[count]=0
 rm -rf "$MDIR"
+
+echo "--- manifest_pg_non_default_settings (#2934) ---"
+FAKE_PG[pg_overrides]='{"liveSettingsSourceNotDefault":{"statement_timeout":{"setting":"30000","source":"user"}},"roleConfig":{"postgres":["statement_timeout=30000ms"]},"databaseRoleConfig":[]}'
+assert_eq "returns the live query's JSON verbatim" \
+  "${FAKE_PG[pg_overrides]}" "$(manifest_pg_non_default_settings)"
+
+MDIR2="$(mktemp -d)"
+manifest_write "$MDIR2" "test-scenario" "conn-1" 1 '{}'
+assert_eq "a role-level ALTER ROLE ... SET reaches the manifest" \
+  '["statement_timeout=30000ms"]' "$(jq -c '.environment.postgres.nonDefaultSettings.roleConfig.postgres' "$MDIR2/manifest.json")"
+assert_eq "a non-default live pg_settings row's source is recorded too - what a SHOW from a DIFFERENT role would miss is exactly why roleConfig exists alongside this" \
+  "user" "$(jq -r '.environment.postgres.nonDefaultSettings.liveSettingsSourceNotDefault.statement_timeout.source' "$MDIR2/manifest.json")"
+rm -rf "$MDIR2"
+
+# A stand with no ALTER ROLE/DATABASE override and no non-default
+# pg_settings row at all must still round-trip as an empty, VALID object -
+# never as a blank string that would break the manifest's own JSON.
+FAKE_PG[pg_overrides]=""
+assert_eq "empty query result degrades to an empty JSON object, not a blank string" \
+  "{}" "$(manifest_pg_non_default_settings)"
+
+# A malformed/garbage answer (a psql error message slipping through the
+# `2>/dev/null`, a truncated query) must not silently produce an invalid
+# manifest.json - it degrades to a JSON object NAMING the failure, so the
+# rest of the manifest (which the run genuinely needs) still writes. This is
+# the "recording beats refusing" rule applied to the recorder's own failure
+# mode: a broken read must be visible in the manifest, not swallowed into a
+# manifest that merely looks complete.
+FAKE_PG[pg_overrides]='ERROR: relation "pg_db_role_setting" does not exist'
+MDIR3="$(mktemp -d)"
+manifest_write "$MDIR3" "test-scenario" "conn-1" 1 '{}'
+assert_eq "manifest.json is still valid JSON when the pg-overrides query answers garbage" \
+  "1" "$(jq -e . "$MDIR3/manifest.json" >/dev/null 2>&1 && echo 1 || echo 0)"
+assert_contains "the garbage is recorded as an error field, not silently dropped" \
+  "$(jq -r '.environment.postgres.nonDefaultSettings.error' "$MDIR3/manifest.json")" "non-JSON"
+rm -rf "$MDIR3"
+FAKE_PG[pg_overrides]=""
+
+echo "--- reassert_volatile_guards (#2932) ---"
+# Neither volatile guard was ever declared by this "scenario" - a no-op,
+# even though the live environment would fail either check outright. This is
+# F5's own shape: read-only, never calls guard_runner_state at all, and
+# reassert must not invent a posture for it.
+GUARD_RUNNER_STATE_EXPECTED=""
+GUARD_SCHEDULER_OFF_ACTIVE=0
+WORKER_CONTAINERS="reassert-worker"
+FAKE_ENV["reassert-worker:WORKER_RUNNER_ENABLED"]="true"
+FAKE_ENV["reassert-worker:OL_SCHEDULER_ENABLED"]="true"
+assert_ok "neither volatile guard declared -> reassert is a no-op" reassert_volatile_guards
+
+# guard_runner_state disabled was declared and the runner is STILL disabled
+# -> reassert re-verifies and passes silently.
+FAKE_ENV["reassert-worker:WORKER_RUNNER_ENABLED"]="false"
+guard_runner_state disabled
+assert_ok "runner still disabled at the next window -> reassert passes" reassert_volatile_guards
+
+# THE CASE #2932 IS ABOUT: a peer (or an operator, or a force-recreate)
+# flips the runner BETWEEN windows. guard_runner_state disabled passed once,
+# at pre-flight; the runner is enabled by the time a later window opens.
+# reassert must catch this and abort - never record the stale "disabled" as
+# though it still held.
+FAKE_ENV["reassert-worker:WORKER_RUNNER_ENABLED"]="true"
+assert_dies "runner flipped enabled after guard_runner_state disabled passed -> reassert dies" reassert_volatile_guards
+
+# Same shape for the scheduler, independently of the runner.
+GUARD_RUNNER_STATE_EXPECTED=""
+FAKE_ENV["reassert-worker:WORKER_RUNNER_ENABLED"]="false"
+guard_runner_state disabled
+FAKE_ENV["reassert-worker:OL_SCHEDULER_ENABLED"]="false"
+guard_scheduler_off
+assert_ok "scheduler still off at the next window -> reassert passes" reassert_volatile_guards
+FAKE_ENV["reassert-worker:OL_SCHEDULER_ENABLED"]="true"
+assert_dies "scheduler flipped on after guard_scheduler_off passed -> reassert dies" reassert_volatile_guards
+
+GUARD_RUNNER_STATE_EXPECTED=""
+GUARD_SCHEDULER_OFF_ACTIVE=0
+WORKER_CONTAINERS="lab-worker"
+
+echo "--- window_start re-asserts before any side effect (#2932) ---"
+# window_start must die on the identical flip, and must die BEFORE
+# manifest_write/sampler_start ever run - proven by asserting no
+# manifest.json was written on the aborted call (the AC's own wording: "a
+# mid-run change aborts the run rather than being recorded as the
+# start-of-script value").
+WORKER_CONTAINERS="reassert-worker"
+FAKE_ENV["reassert-worker:WORKER_RUNNER_ENABLED"]="false"
+guard_runner_state disabled
+FAKE_ENV["reassert-worker:WORKER_RUNNER_ENABLED"]="true"
+WSDIR="$(mktemp -d)"
+SETTLE_SECS=0
+assert_dies "window_start dies when a volatile guard flips between windows" \
+  window_start "$WSDIR" test-scenario "'conn-1'" 1 '{}'
+assert_eq "no manifest.json was written on the aborted window_start" \
+  "0" "$([ -f "$WSDIR/manifest.json" ] && echo 1 || echo 0)"
+rm -rf "$WSDIR"
+GUARD_RUNNER_STATE_EXPECTED=""
+GUARD_SCHEDULER_OFF_ACTIVE=0
+WORKER_CONTAINERS="lab-worker"
 
 # ===========================================================================
 # agreement-rule arithmetic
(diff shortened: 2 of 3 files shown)


piotrswierzy (commented):
The re-assertion half is right. `reassert_volatile_guards` genuinely re-runs the guards rather than re-reporting a cached value; the tracking globals are assigned at the **end** of each guard on success only, so a guard that died never arms the re-assert (easy to get backwards); absence-means-skip for F5 is a documented decision rather than a default; and asserting *no `manifest.json` was written* proves the ordering property that a bare `assert_dies` would not — that assertion would have passed against a re-assert placed after `manifest_write`. The volatile/fixed/orthogonal table naming each fixed guard's *mechanism* of immutability (boot-time `-c` flag, `docker run` env) rather than asserting it, and instructing the next guard author to declare their side, is a good thing to have written down.

The manifest half needs work.

**Blocking — `lib.sh:673`: a failed overrides read is recorded as `{}`.**
`[ -n "$out" ] && printf '%s' "$out" || printf '{}'`. The query is `jsonb_build_object` over three `COALESCE`'d sub-selects, so a *successful* run can never return an empty string — empty output can only mean the read failed, and the caller's `2>/dev/null` has already swallowed the reason. So `{}` is never a true answer; it is always the failure path wearing a benign value, asserting "no non-default Postgres settings are in force" — the exact claim #2934 exists to make impossible. This is the rule `lib.sh` states for itself at `guard_build` / `guard_runner_state:559`: "the label is missing" must never read the same as "the label matches". Emit `{"error":"…"}` for the empty case as well as the non-JSON one. Separately, `pg_sql` carries no `|| true` and, unlike every neighbouring line in `manifest_gather_environment` (`|| printf 'unknown'`), under `set -euo pipefail` the assignment at `:657` aborts the whole run on a psql failure — so the `{}` fallback is unreachable in production and the harness dies mid-manifest instead of degrading. Pick one behaviour and have the tests exercise it.

**Important — `lib-test.sh:563`: the fake `pg_sql` never fails.** It always `echo`s and always returns 0, so `empty query result degrades to an empty JSON object` and the garbage-answer case test string handling rather than the failure mode they're named for. Same shape as the #2918 fakes. Add a `FAKE_PG_FAIL` switch that returns non-zero, and assert the manifest records a named error rather than `{}`.

**Important — `lib.sh:800`: `excludedPgStatStatementsQueries` still declares a query nothing here issues.** It names `GET /v1/connections/:id/sync-status`, an HTTP endpoint this harness never calls, while `sample_queue`'s five per-tick statements — four `COUNT(*) FROM sync_jobs` variants and `pg_database_size` — go undeclared and land in `pg_stat_statements`, contaminating F5's own read-path numbers. This is the "record what shaped a run" PR, editing the same `jq -n` block; a false declaration sitting beside a new true one undermines both, and the test's `length > 0` can't catch a wrong entry.

**Important — `lib.sh:555`: `guard_runner_state` fails open when the scenario expects `enabled`.** `enabled="$(docker exec … || printf 'true')"` then `[ "$enabled" != "false" ]`, so a failed `docker exec` reads identically to "the runner is enabled". Your own new docblock names `docker compose up --force-recreate` as the threat model, and that is exactly when `exec` fails transiently. The `disabled` branch is fail-closed and the lane-caps check rescues most of the `enabled` path, but not the case where `exec` fails while `logs` succeeds. Distinguish "read failed" from "value is X" and die on the former, as `guard_build` does for an absent label.

**Important — no new test re-asserts the `enabled` posture.** All four reassert / `window_start` assertions use `guard_runner_state disabled`. `enabled` is the posture that actually flipped in the #2842 incident, is a materially different code path (`docker logs | grep -F | tail -1`, the `grep -oP` lane-cap parse), and is the one now re-run every window. Nothing asserts that per-window re-assertion refreshes `MANIFEST_LANE_CAPS` either — worth one assertion, since a peer recreate with different caps is precisely the case that would otherwise be recorded stale.

**Important — none of the 13 new assertions runs in CI.** No workflow references `perf/`, `check:invariants` doesn't include `lib-test.sh`, and there's no package script for it. The `96 passed` figure is a claim about the author's machine, and the next change to `lib.sh` can break these silently. One step (`bash perf/openlinker-throughput/lib-test.sh`) closes it.

**Suggestions**

- `docker-compose.lab.yml:90` — `statement_timeout=30000` as a `-c` flag has a wider blast radius than the `ALTER ROLE` it replaces: every role, every session, from boot. `sample_queue`'s counts use `|| printf 0`, so a statement cancelled at 30 s against a multi-million-row `sync_jobs` reports **queue depth 0** — a drained queue. The seeder is covered (`seed_sql` sets `statement_timeout = 0`); confirm the sampler and the drain-wait loops are too, or give them a session timeout.
- `lib.sh:560,690` — `grep -oP` is GNU-only. Pre-existing rather than added here, but the lane-caps parse is now on a per-window hot path, so a BSD/macOS runner fails the guard for a reason the stand doesn't have.

On the three #2918 fail-open defects: `post_guard_limiter_degraded`, `reset_between_repeats`' `--no-raw` SCAN loop, and `guard_stand_exclusive`'s EXIT trap are all still present unchanged — untouched by this PR and not made worse (the new test even works around the trap with `trap - EXIT`). But the blocking item above is a fourth instance of the same shape, introduced here.

piotrswierzy (approved):
## Summary

Two related honesty fixes to the perf harness: volatile guards are re-asserted per window rather than trusted from script start, and every non-default Postgres setting is recorded into `manifest.json` instead of a hand-picked subset. Both are well-reasoned. The volatile/fixed/orthogonal taxonomy written into `lib.sh` is the strongest part of the change — it turns "which guards need re-checking" from tribal knowledge into a decision the next guard author is forced to make explicitly, and each classification is individually justified (boot-time arg vs. live container property) rather than asserted.

The tracking-globals design is right for the same reason the PR argues: `reassert_volatile_guards` re-runs only what a scenario **declared**, so F5 — which deliberately never calls `guard_runner_state` — gets no invented posture. And the assertion that no `manifest.json` is written on an aborted `window_start` is the one that actually proves the ordering claim, rather than merely asserting the abort.

Two things I'd want addressed. One is a residual gap the change gets close to but does not close; the other is a blast-radius widening in the compose file that the PR verifies for only one of its callers.

---

## Findings

**[IMPORTANT]** — `perf/openlinker-throughput/lib.sh:895-915` (`window_start`)
> **The re-assertion is placed at the earliest point before the window, not the latest — leaving a residual gap the same size as the incident it fixes.** The resulting order is:
> ```sh
> reassert_volatile_guards
> manifest_write "$dir" …
> log "settling ${SETTLE_SECS}s …"
> sleep "$SETTLE_SECS"        # SETTLE_SECS defaults to 60
> sampler_start "$dir" "$conn_ids"
> WINDOW_START_EPOCH="$(epoch)"
> ```
> So the guard is verified, the manifest is written asserting that verification, and then the harness sleeps **60 seconds** before the window actually opens. A peer flipping the runner inside that settle window produces exactly the artifact #2932 exists to prevent: a manifest claiming `"runnerState": "disabled"` for a window in which the runner was executing jobs.
>
> This is not a hypothetical margin — it is the same magnitude as the incident the docblock cites. Per that account, F2 flipped the runner at 23:30:24 and F3's next arm opened at 23:30:49: **25 seconds**, comfortably inside a 60s settle. Under this change that arm would still have been mis-manifested.
>
> The current placement is required for the "no side effect before the abort" property, and that property is worth keeping — so the fix is not to move the call but to make it two: re-assert before `manifest_write` (as now, so an abort precedes the manifest) **and** again after `sleep "$SETTLE_SECS"`, immediately before `sampler_start`. The second call is idempotent, costs a handful of `docker exec printenv` calls against a window measured in minutes (the diff's own cost argument), and is what makes the manifest describe the window it labels. As it stands, the PR's "What this does not establish" section understates the remaining gap: it says the uncovered case is a flip *during* a window, when a flip in the settle period is also uncovered and is cheaply closable.

**[IMPORTANT]** — `docker-compose.lab.yml:90-91`
> **Moving `statement_timeout=30000` to a server-wide boot flag widens its blast radius from "the `postgres` role after F5 has run" to "every session on the stand from boot", and the verification covers only the seeder.** The compose comment is candid about this (*"It bounds the API's runaway-read risk for EVERY session on this stand now, server-wide, from boot"*) and correctly points at `seed-lib.sh`'s session-scoped `SET statement_timeout = 0` as the mitigation for the seed path. But `seed_sql()` is not the only long-running statement in the harness.
>
> `perf/openlinker-throughput/seed/cleanup.sh` runs its deletes through `pg_sql_write`, which is a plain `docker exec … psql -c` with **no** session override:
> ```sh
> pg_sql_write "DELETE FROM order_line_items WHERE \"orderRecordId\" LIKE '${PREFIX}\\_ord\\_%' ESCAPE '\\'"
> pg_sql_write "DELETE FROM order_records   WHERE \"internalOrderId\" LIKE '${PREFIX}\\_ord\\_%' ESCAPE '\\'"
> ```
> Against the 1M-order F5 dataset that is a multi-million-row `DELETE … LIKE`. Before this change it inherited a timeout only on a stand where F5 had already run its `ALTER ROLE`; now it inherits 30s from the moment the stand boots, on every stand, for every scenario. A cleanup that times out at 30s leaves the seed data in place, and the next run's `guard_queue_empty` / row-count assumptions are then measuring a dataset nobody intended.
>
> The api container's own boot-time migrations are the second exposure — a migration building an index over a seeded 1M-row table now has a 30s ceiling it did not have before. The fix is already established in the repo: extend `seed_sql`'s session-scoped `SET statement_timeout = 0` treatment to `pg_sql_write` (or at least to `cleanup.sh`), and say in the compose comment which callers are exempted and why. Right now the comment claims the seeder is handled and is silent on everything else that runs long.

**[SUGGESTION]** — `perf/openlinker-throughput/README.md:116-129`
> The README still opens its guard section with *"Guards are **fatal and pre-flight**"* and its table has no volatile/fixed column — but two of those guards are now also re-asserted per window, and `reassert_volatile_guards` appears nowhere in the file. The taxonomy is the most durable thing this PR adds and it currently lives only in a `lib.sh` comment block. A reader deciding where their new guard belongs will read the README. Worth mirroring the three-way split into the guard table (one column, three values) and noting the re-assertion on the `guard_runner_state` / `guard_scheduler_off` rows.

**[SUGGESTION]** — `perf/openlinker-throughput/lib.sh:658-661` (`manifest_pg_non_default_settings`)
> ```sh
> [ -n "$out" ] && printf '%s' "$out" || printf '{}'
> ```
> The `&& … || …` idiom is only equivalent to if/else while the middle command cannot fail. `printf` won't here, so this is correct today — but the function's whole point is degrading safely on a malformed answer, and an explicit `if/else` states that intent without relying on `printf`'s exit status. (The jq validation in `manifest_gather_environment` catches the failure regardless, so this is stylistic.)

---

## Positive observations

- **The `jq -e .` validation is the right instrument, in the right place.** Degrading a malformed blob to `{"error": …}` rather than dropping it keeps the recorder's own failure visible in the artifact — the same "recording beats refusing" rule the query itself is built on, applied one level up. The test asserting `manifest.json` is still valid JSON when the query answers a psql error message is the one that matters.
- **Joining `pg_roles.rolconfig` and `pg_db_role_setting` alongside `pg_settings` is the correct read**, and the docblock explains why: `pg_settings` answers for the connecting role's own resolved session, so a `SHOW` from a different role is exactly how the original `ALTER ROLE` stayed invisible. This is a genuinely more complete answer than the issue asked for.
- **The "a guard refusing unexpected settings would have been the wrong instrument" argument is right and worth having written down** — what an operator may legitimately tune is an open set, so the property to enforce is reconstructability, not conformance.
- **Test isolation is handled.** `GUARD_RUNNER_STATE_EXPECTED` / `GUARD_SCHEDULER_OFF_ACTIVE` are reset before and after each new block, so the earlier `guard_runner_state enabled` / `guard_scheduler_off` cases at lines 212-263 cannot leak a posture into the `window_start` assertions. Easy to get wrong with globals; it isn't.

---

## Merge readiness

🔄 **Approve with changes** — no BLOCKING issues. Both IMPORTANT items are about reach rather than correctness: the re-assertion is real and works, and the manifest recording is a strict improvement on what preceded it. Neither can regress an existing run.

### Priority fixes
1. Re-assert a second time after the `SETTLE_SECS` sleep, immediately before `sampler_start` — the 60s settle currently leaves a gap larger than the 25s flip this change was written for.
2. Give `pg_sql_write` (or at least `seed/cleanup.sh`) the same session-scoped `SET statement_timeout = 0` that `seed_sql` has, before the server-wide 30s starts cancelling million-row cleanups on every stand from boot.
3. Mirror the volatile/fixed/orthogonal split into the README guard table, and drop the unqualified "fatal and pre-flight" framing now that two guards are also per-window.

Issue #2350, opened 2026-08-22
[TASK] Frontend — stock-at-risk surface for reservation shortfall


piotrswierzy:
> Part of the OMS programme backlog (**Wave 2**, stream S3, size M). Backlog slug: `W2-13`. Sibling `W2-*` slugs resolve via the Wave-2 epic's checklist. **Standing S3 rule**: Zod schemas over new projections use `.nullish()`, never `.optional()` (#939).

## Problem / Context

The shortfall fact is only useful where the operator can act on it: the order and the product.

## Proposed Solution

Order-row badge and order-detail callout — *"Order {ref} is short {n} × {sku}"* with the spec's body copy ("The stock master dropped below what this order was promised. Nothing was silently reduced — this order is the one at risk.") — plus a product-row badge on `/products` and a link between the two. Copy comes from the `W2-20` module; the state is counted by the `W2-19` attention section.

## Classification

**Type**: Frontend
**Layer**: Interfaces
**File(s)**: `apps/web/src/features/orders/**`, `apps/web/src/features/products/**`

## Dependencies

- #2349 (`W2-12`), #2356 (`W2-19`), #2357 (`W2-20`)

## Assumptions

- No inline remediation in v1 — the operator's fix is off-system (buy stock, cancel, contact buyer).

## Acceptance Criteria

- [ ] Badge title text is byte-identical to the attention-table title (one copy source)
- [ ] Renders on desktop rows and mobile cards from one renderer
- [ ] Component test covers the shortfall badge and the detail callout

PR #2941, opened 2026-09-06
fix(web): guard the degraded-data reads that crashed / and /settings

### LINKED ISSUE #2926: [BUG] Frontend — / and /settings crash when a data dependency returns an unexpected shape
## Problem / Context

Two authenticated routes crash outright when a data dependency returns a shape other than the expected populated array:

- `/` - the analytics landing page
- `/settings`

Both throw `TypeError: Cannot read properties of undefined (reading 'length')`, caught by React Router's error boundary, so the operator sees an error page rather than a degraded one.

Found by the #2866 Lighthouse harness's own pre-flight render check, which drives each route against a generic network stub. It is reproducible with `scripts/lighthouse/smoke-check.mjs`.

**What this does and does not establish.** It was found under a stub, so it is evidence of a *defensiveness gap* - a `.length` read with no optional chaining on a field whose shape is not guaranteed - not proof that these pages crash in production today. But the shapes a stub produces are the shapes a degraded API produces: a 500 handled into an empty object, a partial payload, a field removed by version skew. The rest of the app already defends against exactly this; `docs/architecture-overview.md` records the convention for the returns surfaces, where both the detail loader and the batched list read catch to empty and never render a positive claim from an absent value.

Six other routes tried - `/orders`, `/products`, `/connections`, `/customers`, `/listings`, `/shipments` - all degraded gracefully. So the gap is not systemic, but **its extent is unknown**: five routes were smoke-tested out of roughly thirty authenticated ones.

## Proposed Solution

A defensive-coding pass over the two crashing pages: find the unguarded `.length` reads and give them the same treatment the rest of the app uses. Do not fix this by widening the test stub to model real domain shapes - that turns a network stub into a second copy of the API and hides the class rather than closing it.

Then extend `scripts/lighthouse/smoke-check.mjs` to cover the remaining authenticated routes, so the extent question is answered rather than left open. That check is cheap and already exists.

## Classification

**Type**: Frontend
**Layer**: Interface
**File(s)**: `apps/web/src/pages/**` (the `/` and `/settings` route modules), `scripts/lighthouse/smoke-check.mjs`

## Dependencies

- None. `scripts/lighthouse/smoke-check.mjs` already exists and reproduces it.

## Assumptions

- The crash is in page-level code rather than a shared component, since six other routes render cleanly against the same stub. Worth confirming before fixing, since a shared component would widen the blast radius considerably.

## Acceptance Criteria

- [ ] `/` and `/settings` render a degraded state rather than throwing when a data dependency returns an unexpected shape
- [ ] The fix is in the page code, not in the stub
- [ ] `smoke-check.mjs` covers every authenticated route, so the extent of the gap is established rather than estimated
- [ ] The check fails if either route regresses

norbert-kulus-blockydevs:
Closes #2926. Based on **`2842-f3-webhook-burst`** (#2922) - review against the stack tip.

## The crash is in page code, and `/settings` had two of them

Reproduced live with Playwright against a sourcemap-decoded stack, not inferred from the six routes that degraded cleanly.

**`/`** - `analytics-page.tsx:91`, `trustQuery.data.connections.length`. The trust snapshot arrived truthy but shapeless and `.connections` was `undefined`.

**`/settings`** - two independent bugs stacked behind each other, and the second is the one worth remembering.

`CurrencySettingsDialog` is mounted **unconditionally** by its tile - `open` only controls Radix visibility, so the component body runs whether or not the dialog is shown - and it read `view.coverage.find(...)` on an undefined array.

Fixing that revealed `McpTokensTile`'s `tokensQuery.data?.filter(...)`:

> **Optional chaining guards `undefined`, not a truthy value of the wrong type.**

`.filter` on an envelope object throws `is not a function`. A `?.` reads like a defensive measure and is not one against a shape mismatch. Worth knowing before writing the next one.

## The fixes are in the pages, deliberately

Widening the test stub to model real domain shapes would turn a network stub into a second copy of the API and hide the class rather than close it. And these shapes are what a degraded API really produces: a 500 handled into an empty object, a partial payload, a field removed by version skew.

The degraded states follow the convention the app already has for the returns surfaces - **never render a positive claim from an absent value**, because a page that says "no shortfalls" when it merely could not load is a worse bug than one that crashes.

## The extent is now established, and it is larger than the issue assumed

`smoke-check.mjs` grows from 5 routes to **34** - every `coreChildren` entry. **14 crash**:

`/insights`, four `/connections/:id...` variants, `/connections/new/advanced`, `/settings/sync-pacing`, `/settings/sales-documents`, `/settings/who-decides`, `/settings/mcp-tokens`, `/settings/prompt-templates`, `/ai/provider-settings`, and both `/ai/prompt-templates` variants.

Those are **untouched here** and raised for triage rather than fixed in passing. The check's exit-code regression guard is scoped to `/` and `/settings`, so it fails if what this PR fixed regresses, without turning fourteen known-open cases into a permanently red gate that everyone learns to ignore.

## Tests

Four regression cases, one on a component that had no test file at all (`McpTokensTile`).

```
vitest run (7 affected/new files) -> 43/43 passing
pnpm --filter @openlinker/web type-check -> clean
pnpm --filter @openlinker/web lint -> 0 errors, 116 pre-existing warnings, none in touched files
```

### DIFF
--- a/apps/web/src/features/currency-settings/components/currency-coverage-dialog.test.tsx
+++ b/apps/web/src/features/currency-settings/components/currency-coverage-dialog.test.tsx
@@ -74,4 +74,19 @@ describe('CurrencyCoverageDialog', () => {
     fireEvent.click(await screen.findByRole('button', { name: 'Close' }));
     expect(onClose).toHaveBeenCalled();
   });
+
+  it('degrades to the zero-state rather than throwing when `stampedOrders` is missing (#2926)', async () => {
+    // A degraded API response (a 500 handled into an empty object, a partial
+    // payload, a field dropped by version skew) can arrive as a resolved
+    // query whose shape doesn't match the contract.
+    const degradedView = {
+      ...baseView,
+      stampedOrders: undefined,
+    } as unknown as CurrencySettingsView;
+
+    renderWithProviders(<CurrencyCoverageDialog open view={degradedView} onClose={vi.fn()} />);
+
+    expect(await screen.findByText('Analytics coverage')).toBeInTheDocument();
+    expect(screen.getByText(/No orders have been stamped yet/)).toBeInTheDocument();
+  });
 });
--- a/apps/web/src/features/currency-settings/components/currency-coverage-dialog.tsx
+++ b/apps/web/src/features/currency-settings/components/currency-coverage-dialog.tsx
@@ -41,8 +41,16 @@ export function CurrencyCoverageDialog({
   view,
   onClose,
 }: CurrencyCoverageDialogProps): ReactElement {
-  const total = view.stampedOrders.reduce((sum, entry) => sum + entry.count, 0);
-  const hasMultipleEras = view.stampedOrders.length > 1;
+  // `view` reflects whatever the API returned; a degraded response (a
+  // partial payload, a field dropped by version skew) can arrive as a
+  // successful query with `stampedOrders` missing. Defaulting to empty
+  // degrades into "no orders stamped yet" rather than crashing — the same
+  // no-positive-claim-from-absent-data convention documented for the
+  // returns surfaces in docs/architecture-overview.md.
+  const stampedOrders = view.stampedOrders ?? [];
+
+  const total = stampedOrders.reduce((sum, entry) => sum + entry.count, 0);
+  const hasMultipleEras = stampedOrders.length > 1;
 
   const items: KeyValueItem[] = [
     {
@@ -52,7 +60,7 @@ export function CurrencyCoverageDialog({
       mono: true,
     },
     ...(hasMultipleEras
-      ? view.stampedOrders.map((entry) => ({
+      ? stampedOrders.map((entry) => ({
           id: entry.reportingCurrency,
           label: `— stamped in ${entry.reportingCurrency}`,
           value: <span className="tabular">{entry.count}</span>,
--- a/apps/web/src/features/currency-settings/components/currency-settings-dialog.test.tsx
+++ b/apps/web/src/features/currency-settings/components/currency-settings-dialog.test.tsx
@@ -107,4 +107,25 @@ describe('CurrencySettingsDialog', () => {
     fireEvent.change(select, { target: { value: 'PLN' } });
     expect(liveRegion().textContent).toBe('');
   });
+
+  it('does not throw when `view` arrives with array fields missing (#2926)', async () => {
+    // The tile mounts this dialog unconditionally (`open` merely controls
+    // Radix's own visibility), so the component body runs on every render of
+    // the settings page — a degraded API response (a 500 handled into an
+    // empty object, a partial payload, a field dropped by version skew)
+    // must not crash it even while closed.
+    const degradedView = {
+      reportingCurrency: 'PLN',
+      source: 'setting',
+      updatedAt: null,
+      updatedBy: null,
+      rateSource: 'nbp',
+      rateDateRule: 'prev-business-day',
+      // coverage / stampedOrders / supportedCurrencies deliberately absent
+    } as unknown as CurrencySettingsView;
+
+    expect(() => {
+      renderWithProviders(<CurrencySettingsDialog open={false} view={degradedView} onClose={vi.fn()} />);
+    }).not.toThrow();
+  });
 });
--- a/apps/web/src/features/currency-settings/components/currency-settings-dialog.tsx
+++ b/apps/web/src/features/currency-settings/components/currency-settings-dialog.tsx
@@ -57,9 +57,21 @@ export function CurrencySettingsDialog({
   view,
   onClose,
 }: CurrencySettingsDialogProps): ReactElement {
+  // `view` reflects whatever the API returned; a degraded response (a
+  // partial payload, a field dropped by version skew) can arrive as a
+  // successful query with these fields missing. Defaulting the arrays to
+  // empty and the currency to `''` degrades into "no coverage gap known" /
+  // "nothing else stamped" / "nothing selected yet" rather than crashing —
+  // the same no-positive-claim-from-absent-data convention documented for
+  // the returns surfaces in docs/architecture-overview.md.
+  const reportingCurrency = view.reportingCurrency ?? '';
+  const coverage = view.coverage ?? [];
+  const stampedOrders = view.stampedOrders ?? [];
+  const supportedCurrencies = view.supportedCurrencies ?? [];
+
   const { showToast } = useToast();
   const mutation = useSetReportingCurrencyMutation();
-  const [selected, setSelected] = useState(view.reportingCurrency);
+  const [selected, setSelected] = useState(reportingCurrency);
   const [acknowledged, setAcknowledged] = useState(false);
 
   // Reset whenever the dialog opens, matching the ai-provider-key-dialog /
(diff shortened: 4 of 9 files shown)


piotrswierzy (commented):
Strong crash reproduction, and the `McpTokensTile` fix is exactly right — `Array.isArray(...)` discriminating a wrong-typed truthy value, degrading to the same `—` a pending read shows. The write-up of why `?.` is not a guard against a shape mismatch (it guards `undefined`, not a truthy value of the wrong type, and `.filter` on an envelope object throws) is worth keeping.

Two of the three fixes I'd hold, for the same reason.

**1. `analytics-page.tsx:78,101` — `?? []` renders "you have no connections" from an unreadable snapshot.**
With `connections = trustQuery.data?.connections ?? []`, a degraded snapshot falls into `connections.length === 0` and shows *"Connect a sales channel to see figures here"* with an **Add a connection** CTA — to an operator who may have twenty live connections. That's the standing rule in `docs/architecture-overview.md` (§ Returns / #2350): an empty array is never a positive claim, because absence and failure are indistinguishable, which is why both reads there refuse to render a "no shortfalls" reassurance. Discriminate with `Array.isArray(trustQuery.data?.connections)` — false routes to the `ErrorState` ("Unable to load data coverage" + Retry) that's already in the same ternary. `analytics-page.test.tsx:43` asserts the empty state and needs inverting with it.

**2. `currency-coverage-dialog.tsx:49` — same shape.**
`stampedOrders ?? []` makes `total` compute to `0`, and the dialog then states as fact that nothing is stamped plus *"This fills in automatically as new orders sync"* — advice derived from a figure the API never supplied. This file's own header docblock already argues the point against itself: the count is hidden from the tile precisely because *"a `0` read as '0 problems' to an operator"*. Guard with `Array.isArray(view.stampedOrders)` and render an unknown line rather than the zero-state; `currency-coverage-dialog.test.tsx:78` pins the wrong copy today.

**3. The guards sit at the render site because the boundary doesn't parse.**
`currency-settings.api.ts:22` and `analytics-trust.api.ts:23` are bare `request<T>(...)` casts over hand-written types, so `view.stampedOrders` is typed `StampedOrderCount[]` and `?? []` is invisible to the type system — a refactor or a `no-unnecessary-condition` pass deletes it and the crash returns. The repo already has the durable pattern: `features/fulfillment/api/fulfillment.schema.ts` parses at the boundary and its header states this exact rule (`.nullish()` not `.optional()`; no `z.enum` on a server-owned vocabulary, so a new value can't make a surface report a false negative), with `automation` and `bench` following it. A `parseAnalyticsTrustSnapshot` / `parseCurrencySettingsView` turns a degraded payload into a query **error** — which both surfaces already handle correctly — and fixes it once for every consumer instead of once per `.length` / `.find` / `.filter`. The body's objection about widening the test stub is about the stub, not the boundary parse.

**4. The root cause you identified is left unfixed.**
The PR correctly diagnoses that `currency-settings-tile.tsx:116` mounts `CurrencySettingsDialog` unconditionally, so the body runs on every settings render even while closed. Four `??` defaults patch today's four field reads; the fifth field added next quarter crashes `/settings` again. Gate the mount (`{dialogOpen ? <CurrencySettingsDialog … /> : null}`) or early-return on `!open` — one change that closes the class rather than four that close instances.

**5. Minor** — `smoke-check.mjs:24-60`: the 14 known-crashing routes are reported as "raised for triage" with no issue number in the header comment, so the finding evaporates; and `ROUTES` is a hand-copied duplicate of `coreChildren` in `root.route.tsx`, which will silently drift as routes are added. A route dropped from that list is a route nobody checks.

Clean on dependency direction, barrel imports, raw `fetch()`, design tokens, `any`, and accessibility.

Two more things worth naming: the degraded fixture is a real envelope (`{ data: [], total: 0 }`) rather than invented state, which is what the testing guide asks for; and separating `pageErrors` from `consoleErrors` in the smoke check is the right distinction — a thrown page error is a different fact from a logged one and only the former should gate. Scoping the regression gate to `/` and `/settings` rather than turning fourteen open cases into a permanently red check is also the right call.

piotrswierzy (approved):
# Code Review — #2941 `fix(web): guard the degraded-data reads that crashed / and /settings`

**Scope**: 9 files, diffed against `2842-f3-webhook-burst` only. Layer: Interface (frontend pages + feature components) plus one harness script. Linked issue: #2926.

---

## 1. Summary

Four unguarded reads across three components, fixed in place, with a colocated regression test for each and the smoke harness widened from 5 to 34 routes. The fixes are in the page/component code rather than the stub, which is #2926's AC-2 verbatim and the right call — widening the stub would model the API twice. The degraded states are all "unknown", never a positive claim, which is the returns-surface convention `docs/architecture-overview.md` § 22 records (*"An empty array is never a positive claim"*). I verified the load-bearing claim in the PR body independently: `currency-settings-tile.tsx:116` really does mount `CurrencySettingsDialog` unconditionally with only `open={dialogOpen}` controlling Radix visibility, so the component body runs on every settings render — that is why the second bug was reachable at all.

The one thing I would push back on is the coverage claim, below. Nothing here is blocking.

---

## 2. Alignment with documentation

✅ Aligned:
- Fix location matches the issue's explicit instruction and `docs/frontend-architecture.md` § Components And Pages (pages own composition; guards belong where the read is).
- Test naming/colocation: `*.test.tsx` beside source, per § Naming conventions.
- `McpTokensTile` degrades to the same `—` a pending/failed read renders — it does not claim `0`. That distinction is the whole point and it is made correctly.
- No stub widening, no shared-component change, so the blast radius stays where #2926's own assumption put it.

No deviations found.

---

## 3. Findings

**[IMPORTANT]** — `scripts/lighthouse/smoke-check.mjs` (header comment + `ROUTES`)
> The header says `ROUTES` is *"the full authenticated route list from `root.route.tsx`'s `coreChildren`"* plus *"one representative parameterised detail route per list page"*. Neither half holds. `coreChildren` has **35** top-level entries and `ROUTES` has 34 — `promptTemplateLegacyDetailRedirectRoute` is absent. More materially, the *"one detail route per list page"* claim is true only for `/connections`: reading the route modules, `orders.route.tsx` declares `failed`, `dispatch-risk` and `:internalOrderId`; `products.route.tsx` `:id`; `listings.route.tsx` `bulk-create/wizard`, `bulk-batches/:batchId`, `:id`; `returns.route.tsx` `:returnId`; `invoices.route.tsx` `:invoiceId`; `automations.route.tsx` `activity` and `:trigger`. That is ~12 authenticated routes the check never visits.
>
> This matters because #2926's AC-3 is *"covers every authenticated route, so the extent of the gap is established rather than estimated"*, and the PR body reports "14 crash" as if it were the extent. With a dozen routes unvisited it is still a lower bound. Either add the nested children (cheap — they are already enumerated in the route modules), or amend both the header comment and the PR body to say **top-level only**, and say why. As written the comment overstates what ran, which is the failure mode this whole PR is about one level up.

**[SUGGESTION]** — `apps/web/src/features/currency-settings/components/currency-settings-dialog.tsx` (~line 96)
> `const hasCoverageGap = (selectedCoverage?.uncoverableCurrencies.length ?? 0) > 0;` is untouched, and it is the same shape the PR's own thesis names: `?.` guards `selectedCoverage`, and `.length` on `uncoverableCurrencies` is then unguarded. A coverage entry that arrives without that array — the exact "field dropped by version skew" case the new comments cite — still throws, in a component the tile mounts on every settings render. `selectedCoverage?.uncoverableCurrencies?.length ?? 0` closes it in one character.

**[SUGGESTION]** — `scripts/lighthouse/smoke-check.mjs` (`REGRESSION_GUARD_ROUTES`)
> Scoping the exit code to the two fixed routes is the right decision and the reasoning in the header ("not a permanently red gate that everyone learns to ignore") is sound. But the 14 crashing routes currently exist only in this PR's description, which is the least durable place for them. Raise the triage issue and cite it from the header comment, so the list survives the PR being merged and scrolled past.

---

## 4. Tests

✅ Four regression cases, one per fix, each asserting the *degraded render* rather than merely `not.toThrow()` — except `currency-settings-dialog.test.tsx`, which uses `not.toThrow()`. That is defensible here: the assertion under test is that a **closed** dialog's body survives, and there is no visible output to assert on. The comment says so.

✅ `mcp-tokens-tile.test.tsx` is new for a component that had none, and covers loading / success / degraded — not just the regression. The degraded fixture (`{ data: [], total: 0 }`) is a realistic envelope shape, not a synthetic one.

✅ Mocks go through `createMockApiClient` / `createAuthenticatedSessionAdapter`; I confirmed `mcpTokens` is a supported override key in `test/test-utils.tsx:648`.

---

## 5. Positive observations

- Splitting `pageErrors` out of `consoleErrors` in the smoke check is a genuine improvement that is easy to miss: previously a thrown page error was appended to a list that was *printed* and never gated on, so a crash could be reported and still pass. `crashed = hasErrorBoundary || pageErrors.length > 0` is what makes the regression guard mean anything.
- The `try/catch` around `page.goto` means one route failing navigation no longer aborts the remaining 33.
- Every guard carries an in-place comment citing the convention rather than restating it, which is the house style for this kind of defensive default.

---

## 6. Final assessment

A tight, correctly-placed fix with real regression coverage and a harness improvement that outlives it. The only thing I want changed before this is quoted as an extent measurement is the coverage claim — the code is fine, the sentence describing it is not.

**Merge readiness**: 🔄 **Approve with changes**

**Priority fixes**
1. Correct the `smoke-check.mjs` header comment and the PR body to match what actually runs, or add the ~12 nested authenticated routes.
2. Guard `uncoverableCurrencies.length` in `currency-settings-dialog.tsx`.
3. File the triage issue for the 14 crashing routes and cite it from the header.

Issue #2927, opened 2026-09-06
[TASK] CORE — the orders-list pagination COUNT scans the whole table under the needs-attention filter


norbert-kulus-blockydevs:
## Problem / Context

`GET /orders?health=needs_attention` degrades **13x** between 10 000 and 1 000 000 orders - median 11.4 ms to 149.0 ms - and the whole degradation is one query.

Measured in #2843 (report: `perf/openlinker-throughput/results-F5-2026-09-06.md`), attributed with `pg_stat_statements` and confirmed with `EXPLAIN`:

```
138 calls  mean 39.636 ms   SELECT COUNT(*) FROM order_records
                            WHERE recordStatus <> $2 AND recordStatus <> $3
                              AND syncStatus @> $4::jsonb
138 calls  mean  0.256 ms   SELECT ... same table, same predicate, LIMIT 20
```

**Same table, same filter, 155x apart.** The paged read stops after twenty matches on `IDX_order_records_createdAt`; the COUNT has no LIMIT and must evaluate a non-sargable jsonb containment against every row.

Two things make it worse than the median suggests:

**The sub-linear growth to 1M is bought with parallel workers.** At 613 MB of heap the planner switches to a 2-worker `Parallel Seq Scan`, which is why 1M measured 141.9 ms instead of the ~400 ms a linear projection predicted. On a server whose worker pool is busy it falls back to a single process and moves back toward the projection.

**The real cost is I/O, and it is not confined to this route.** The count reads **69 155 buffers from disk** per execution - roughly 540 MB pulled through the buffer pool every time an operator opens that page. That evicts whatever else was cached, so the effect is application-wide.

**This generalises.** The unfiltered orders list pays the same shape more cheaply (`SELECT COUNT(*) FROM order_records`, 6.9 ms mean) and degrades 4.8x, while `order detail`, `product detail` and `products list` are flat at 1.0x across the same 100x row increase. The pattern is *any list that reports an exact total*, not *orders*.

## Proposed Solution

Three remedies exist and the choice is a product decision, not a technical one, so this issue does not pick:

- **An index on the predicate** - a GIN index on `syncStatus`, or an expression/partial index matching the health filters. Makes the count sargable; costs write throughput on the hottest table in the system.
- **An estimated or capped total** - "20+", or `pg_class.reltuples` for the unfiltered case. Costs the operator an exact number.
- **A cached count** - correct only if a staleness window is acceptable.

[...]

Issue #2029, opened 2026-08-11
[TASK] Frontend — Lifecycle tabs (Active/Inactive/Draft/Ended) with URL state and counts


norbert-kulus-blockydevs:
## Problem / Context

The redesigned Listings page needs 4 disjoint lifecycle tabs (Active/Inactive/Draft/Ended, default
Active) with live counts that respect filters. The shared `Tabs` primitive already has a count-badge
slot (`tabs__count`, `apps/web/src/index.css:5550`, shipping in `users-page.tsx:430`) but no page today
wires it as a fully bidirectional URL-synced control — the closest full precedent is
`orders-list-page.tsx`'s `?health=` param pattern (plain-button segments, not Radix `Tabs`), and
`connection-mappings-page.tsx`'s `Tabs` only reads a `?tab=` param once without writing back.

## Proposed Solution

1. Add a controlled `Tabs value={tab} onValueChange={...}` where `tab` is read from
   `useSearchParams().get('tab')` (default `'active'`) and `onValueChange` calls a
   `setFilterParam('tab', v)` helper (mirroring `orders-list-page.tsx:965-976`), with a type guard
   (`isLifecycleTab`, mirroring `orders-list-page.tsx:117-119`).
2. Render `tabs__count` badges from BE-C's per-tab counts.
3. Per-tab empty-state copy per mockup: distinct messages for Active/Inactive/Draft/Ended-empty (not
   one generic message), plus the filtered-no-results and no-connections variants.
4. Loading: tab counts render as skeleton lines (not `0`), per mockup frame 04.

## Classification

**Type**: Frontend
**Layer**: Interface (page)
**File(s)**: `apps/web/src/pages/listings/listings-list-page.tsx`

## Dependencies

- FE-B (same file, must land first)
- BE-C (counts)

## Assumptions

- Tab state lives in `?tab=` on the existing single `/listings` route — no new sub-routes.

## Acceptance Criteria

- [ ] All 4 tabs are bookmarkable via `?tab=` and survive a page refresh
- [ ] Counts render as skeletons while loading, real numbers after
- [ ] Each tab has its own distinct empty-state copy per the mockup
- [ ] Counts sum to the filtered total
- [ ] No architecture boundary violations

PR #2957, opened 2026-09-06
perf: the paginated total becomes a second stage (#2943)

### LINKED ISSUE #2943: [EPIC] CORE + Frontend — the paginated total becomes a second stage
## Problem / Context

Every paged list in the product asks the database two questions at once: *give me twenty rows* and *how many are there in total*. The first can stop after twenty matches. **The second cannot stop at all.**

Measured in #2843 against a seeded million orders:

```
138 calls  mean 39.636 ms   SELECT COUNT(*) FROM order_records
                            WHERE recordStatus <> $2 AND recordStatus <> $3
                              AND syncStatus @> $4::jsonb
138 calls  mean  0.256 ms   SELECT ... same table, same predicate, LIMIT 20
```

Same table, same filter, **155x apart**. At a million rows the filtered orders list takes **149 ms**, of which **142 ms is the count** - and the count reads about **540 MB through the buffer pool per execution**, so it evicts whatever else was cached and the cost is felt application-wide rather than on one page.

**It is not one page's problem.** Fifteen repositories run a paged read with a count; **five of them can apply a predicate no plain index can serve**:

| Repository | Non-sargable shapes | Surface |
|---|---|---|
| `order-record` | `@>`, `->>`, `jsonb_` | `/orders` - **measured, 149 ms at 1M** |
| `offer-mapping` | `ILIKE`, `LIKE`, `jsonb_` | `/listings` |
| `product` | `ILIKE` | `/products` |
| `product-variant` | `ILIKE`, `->>` | `/products` (variants) |
| `customer-projection` | `ILIKE` | `/customers` |

The other ten count over indexed equality and are fine.

Only the orders case is measured. The other four are **the same shape, unmeasured** - which is stated here rather than assumed to hurt equally.

## Proposed Solution

**The exact total stays** - the product owner's decision, recorded on #2927: operators need the real number, so an estimate or a hard cap is off the table.

It moves to a **second stage**. The list renders immediately from the paged query; the total arrives separately and replaces a `20+` placeholder when it lands.

This adds no database work - the same two statements run either way - and it enables two things that are impossible today: the count becomes **cacheable per filter combination**, and **abandonable on navigation**, neither of which can happen while the page render depends on it.

Four stages, one PR:

1. **Backend** - separate the total from the paged read across the five repositories
2. **Frontend** - the shared two-stage mechanism: debounce, cancellation, delayed loader
3. **Frontend** - apply it to the four list surfaces
4. **Docs** - the convention, in both doc sets, so the next table gets it right

### Two design points that are easy to get wrong

**Do not use skeleton rows.** A skeleton says "content is coming"; the rows are already there and only a number is missing. Skeleton rows would make the page feel *slower* than it is. The loader belongs on the count itself.

**Pagination splits in two.** Next and Previous need no total and work immediately; "page 3 **of 47**" and jump-to-last do not. The controls must degrade per-affordance rather than all waiting together.

## Classification

**Type**: CORE + Frontend
**Layer**: Infrastructure (persistence), Interface (HTTP + web)
**File(s)**: the five repositories above, their controllers, `apps/web/src/shared/**`, the four list pages, `docs/engineering-standards.md`, `docs/frontend-architecture.md`

## Dependencies

- None. #2843 has already measured the cost.

## Assumptions

- The four unmeasured surfaces share the shape but not necessarily the magnitude. Their datasets are smaller than a million orders on most installs, so this is pre-emptive there and remedial on orders.
- `ILIKE` is separable from `@>`: a trigram index can make the former sargable, and this repo already does exactly that for `DestinationCategory.searchText` (`gin_trgm_ops`, `destination-category.orm-entity.ts:75`). That is a different remedy from the orders case and is **not** in this epic's scope.

## Acceptance Criteria

- [ ] All four stage issues are closed
- [ ] The four list surfaces render rows without waiting for their total
- [ ] The convention is documented in both doc sets, with the "why" and not only the "how"
- [ ] `/orders` at the #2843 1M dataset is re-measured, and the report states time-to-rows separately from time-to-total
- [ ] **Merged to `main` as one PR**, not to the performance programme branch
- [ ] No architecture boundary violations (CORE - Integration)

## Related

- **#2843** - measured the cost, and its 1M dataset is where the re-measurement should happen
- **#2927** - stays open. It owns the separate question of whether the orders count should be made cheaper with an index; a loader hides 142 ms from the operator without reducing it, and the buffer-pool pressure is unaffected by when the query runs

### LINKED ISSUE #2944: [TASK] CORE — separate the paginated total from the paged read
Child of #2943. **Stage 1 of 4.**

## Problem / Context

Fifteen repositories run a paged read and a `COUNT` together. The page can stop after twenty matches; the count cannot stop at all, so on a non-sargable predicate it scans the table regardless of the `LIMIT` beside it - 142 ms of a 149 ms request at a million orders (#2843).

Today a caller cannot ask for one without the other. That is the thing this stage changes; how the frontend then renders it is stage 2's.

Five repositories can apply a predicate no plain index serves, and they are the ones that matter:

| Repository | Non-sargable shapes |
|---|---|
| `libs/core/src/orders/infrastructure/persistence/repositories/order-record.repository.ts` | `@>`, `->>`, `jsonb_` |
| `libs/core/src/listings/infrastructure/persistence/repositories/offer-mapping.repository.ts` | `ILIKE`, `LIKE`, `jsonb_` |
| `libs/core/src/products/infrastructure/persistence/repositories/product.repository.ts` | `ILIKE` |
| `libs/core/src/products/infrastructure/persistence/repositories/product-variant.repository.ts` | `ILIKE`, `->>` |
| `libs/core/src/customers/infrastructure/persistence/repositories/customer-projection.repository.ts` | `ILIKE` |

## Proposed Solution

Make the total **separately obtainable** for those five reads, without breaking the existing callers.

Two shapes are defensible and the implementer should pick one and say why:

- **An optional flag on the existing read** - the caller asks for rows only, and asks again for the total. Fewest new surfaces; risks a boolean parameter that reads ambiguously at call sites.
- **A sibling count method and endpoint** - explicit, and the natural home for the per-filter caching #2943 wants to become possible later. More surface.

Whichever is chosen, three properties are not negotiable.

**The filter must be expressed once.** The rows query and the count query have to apply an identical predicate or the total describes a different set than the page. Share the predicate builder rather than writing it twice - the repositories already build their `WHERE` in one place, and that is what must be reused.

**The existing combined behaviour keeps working.** These reads have callers beyond the four list pages, and this stage is not the place to find out which. A caller that wants both still gets both.

**No caching in this stage.** #2943 names per-filter caching as a thing this *enables*; building it here would ship a staleness window nobody has decided on.

## Classification

**Type**: CORE
**Layer**: Infrastructure (persistence) + Interface (HTTP)
**File(s)**: the five repositories above, their ports, and the controllers serving `/orders`, `/listings`, `/products`, `/customers`

## Dependencies

- None.

## Assumptions

- The four non-orders surfaces are the same shape, unmeasured. This stage treats them alike because the fix is the same; the epic says so rather than implying they hurt equally.

## Acceptance Criteria

- [ ] Each of the five reads can return rows without computing a total, and the total can be obtained separately
- [ ] The predicate is built once and shared - demonstrated by a test that changes a filter and asserts both paths move together
- [ ] Existing callers that want both are unchanged, and that is asserted rather than assumed
- [ ] The chosen shape (flag versus sibling method) is justified in the PR, with the rejected one named
- [ ] No caching is introduced
- [ ] No architecture boundary violations (CORE - Integration)

### LINKED ISSUE #2945: [TASK] Frontend — the shared two-stage total mechanism
Child of #2943. **Stage 2 of 4.** Depends on #2944.

## Problem / Context

Stage 1 makes a paged total separately obtainable. This stage builds the one mechanism every list surface will use to render it, so the four surfaces in stage 3 are four small adoptions rather than four independent inventions.

Built once because getting it wrong is easy in three specific ways, all of which turn an improvement into a regression.

## Proposed Solution

A shared hook (and whatever small presentational piece it needs) that renders a list's total as a second stage.

**Three properties are the whole point of building it centrally.**

**Debounce, or this is worse than what it replaces.** A total re-fetched on every keystroke of a filter input turns one slow page into a request storm against the most expensive query in the system. The debounce is not a nicety; without it this issue makes things worse.

**Cancel the in-flight request when the filter changes again.** Otherwise a fast second query can be overtaken by a slow first one and the operator sees a total belonging to a filter they have already left. That is worse than a missing number, because it looks authoritative.

**Delay the loader, so small installs never see it.** At ten thousand orders the count returns in about 11 ms. A loader that appears and vanishes inside that window is a flicker, and a flicker is worse than no state change at all. A delay of roughly 150-200 ms means the operator on a small install sees an instant page - exactly what they see today - and only a large install ever reaches the two-stage behaviour. **This is what stops the change being visible where there was no problem.**

**The placeholder is `20+`, not a skeleton.** A skeleton says content is coming; the rows are already rendered and only a number is missing. Skeleton rows would make the page read as slower than it is. The loading affordance belongs on the count, inline, and small.

**Pagination degrades per affordance, not as a block:**

| Control | Available before the total |
|---|---|
| Next / Previous | yes - neither needs it |
| Page numbers already reached | yes |
| "Page 3 **of 47**" | no |
| Jump to last page | no |

Wire each to what it actually needs. Disabling the whole pager until the total lands would throw away most of the benefit.

**A failed count is not zero.** If the count request errors, the placeholder stays - the UI must never render `0` or a wrong total from a failed read. This repo already holds that line elsewhere: `docs/architecture-overview.md` records that the returns surfaces catch to empty and never render a positive claim from an absent value.

## Classification

**Type**: Frontend
**Layer**: Interface
**File(s)**: `apps/web/src/shared/**` (the hook and its presentational piece)

## Dependencies

- **#2944** - the backend must be able to answer rows and total separately before this can consume it.

## Assumptions

- TanStack Query is the data layer, per `docs/frontend-architecture.md`, so cancellation and debouncing should use its idioms rather than hand-rolled timers where it offers them.

## Acceptance Criteria

- [ ] A shared hook renders a total as a second stage, with debounce, cancellation and a delayed loader
- [ ] A test proves a superseded count response cannot overwrite a newer one
- [ ] A test proves the loader does not appear when the count resolves inside the delay window
- [ ] A failed count leaves the placeholder, and a test asserts no `0` is ever rendered
- [ ] Pagination affordances are individually gated on whether they need the total
- [ ] Not yet adopted by any page - that is stage 3, so this ships as mechanism plus tests

### LINKED ISSUE #2947: [TASK] Frontend — adopt the two-stage total on the four affected list surfaces
Child of #2943. **Stage 3 of 4.** Depends on #2944 and #2945.

## Problem / Context

Stage 1 makes the total separately obtainable; stage 2 builds the mechanism that renders it. This stage adopts it on the four surfaces whose repositories can apply a predicate no plain index serves.

| Page | Repository behind it | Why it qualifies |
|---|---|---|
| `/orders` | `order-record` | **measured**: 149 ms at 1M, 142 of it the count |
| `/listings` | `offer-mapping` | `ILIKE`, `LIKE`, `jsonb_` - same shape, unmeasured |
| `/products` | `product` + `product-variant` | `ILIKE`, `->>` - same shape, unmeasured |
| `/customers` | `customer-projection` | `ILIKE` - same shape, unmeasured |

The other eleven paged reads count over indexed equality and are deliberately left alone. Adopting everywhere would add a loading state to pages that never needed one.

## Proposed Solution

Four adoptions of the stage-2 hook, each small. Then **re-measure**, because the epic's claim is a latency claim and an unverified one is just an intention.

Re-measure `/orders` against #2843's own 1M dataset with its own scenario (`perf/openlinker-throughput/scenarios/f5-read-path.sh`), not a new one, and report **time-to-rows separately from time-to-total**. Those are now two different numbers and collapsing them back into one would hide the entire point of the change.

The expectation from #2843's figures is rows in roughly 7 ms against a previous 149 ms. **If it does not land near that, say so** - a measurement that contradicts the prediction is more interesting than one that confirms it, and the report should carry whichever happened.

## Classification

**Type**: Frontend
**Layer**: Interface
**File(s)**: `apps/web/src/pages/orders/**`, `apps/web/src/pages/listings/**`, `apps/web/src/pages/products/**`, `apps/web/src/pages/customers/**`

## Dependencies

- **#2944** - the separated backend read
- **#2945** - the shared frontend mechanism

## Assumptions

- The four pages already share enough list scaffolding that adoption is a small edit each. If one of them turns out to have its own bespoke pager, say so rather than quietly rewriting it inside this stage.

## Acceptance Criteria

- [ ] All four surfaces render rows without waiting for their total
- [ ] Each keeps working when the count fails - no `0`, no wrong total
- [ ] `/orders` re-measured on #2843's 1M dataset with `f5-read-path.sh`, reporting time-to-rows and time-to-total as separate figures
- [ ] The report states whether the prediction held, and does not quietly replace it if it did not
- [ ] The eleven other paged reads are untouched, and the PR says why

### LINKED ISSUE #2950: [TASK] DX — document when a paginated total is expensive and how to render one
Child of #2943. **Stage 4 of 4.** Depends on #2944, #2945 and #2947.

## Problem / Context

The pattern this epic fixes was not introduced deliberately - it is what you get by reaching for TypeORM's `getManyAndCount()`, which is the obvious call and is correct for eleven of the fifteen reads that use it. Nothing in the repository's documentation says when it stops being correct.

So the next paged table will be written the same way, and the same 155x gap will reappear on whichever surface grows first. Documentation is the only thing that prevents that, because the mistake is a reasonable default rather than an oversight.

Two doc sets, because the question splits in two: the backend has to know **when a total is expensive**, and the frontend has to know **how to render one that is**.

## Proposed Solution

**Backend - `docs/engineering-standards.md`**, near the repository-pattern material.

The rule, stated as a rule rather than a story: a paged read stops after its `LIMIT`; the count beside it cannot stop at all. So when the filter is **sargable** - an indexed equality - the total is nearly free and `getManyAndCount()` is right. When it is **not** - a jsonb containment, an `ILIKE`, a function-wrapped column - the count scans the table however small the page is.

Name the shapes explicitly (`@>`, `->>`, `ILIKE`, `LIKE`, `LOWER(`) so the reader can check their own query rather than having to judge. Give the measured evidence in one line - 0.256 ms against 39.6 ms on identical predicates at 100k, 142 ms at 1M - because a rule with a number behind it survives review and a rule without one gets argued away.

Mention the trigram precedent, since it is the remedy for the `ILIKE` half and this repo already uses it: `DestinationCategory.searchText` with a GIN `gin_trgm_ops` index (`destination-category.orm-entity.ts:75`).

**Frontend - `docs/frontend-architecture.md`**, near the state-ownership material.

The convention: render rows from the paged query, render the total as a second stage. Then the four things that make it work rather than backfire, each with its reason:

- **debounce** - a total re-fetched per keystroke turns one slow page into a request storm
- **cancel the in-flight request** - a slow earlier response must not overwrite a fast later one, or the operator reads a total belonging to a filter they have left
- **delay the loader ~150-200 ms** - at small row counts the total returns in about 11 ms, and a flicker is worse than no state change; this is what keeps the change invisible on installs that never had the problem
- **`20+`, never skeleton rows** - the rows are already there; a skeleton claims content is missing and makes the page read as slower than it is

Plus the pagination split - Next and Previous need no total, "page 3 of 47" and jump-to-last do - and the rule that **a failed count leaves the placeholder and never renders `0`**, which is the same principle `docs/architecture-overview.md` already records for the returns surfaces: never render a positive claim from an absent value.

## Classification

**Type**: DX
**Layer**: Shared (documentation)
**File(s)**: `docs/engineering-standards.md`, `docs/frontend-architecture.md`

## Dependencies

- **#2944**, **#2945**, **#2947** - document what shipped, not what was planned. If any of them deviated from the design, the docs follow the code.

## Assumptions

- Both documents are the right homes. `docs/engineering-standards.md` already carries the repository-port and ORM-mapping rules; `docs/frontend-architecture.md` already carries state ownership. Neither needs a new document.

## Acceptance Criteria

- [ ] The backend entry states when a total is cheap and when it is not, names the non-sargable shapes, and carries the measured figures
- [ ] The frontend entry states the four mechanism properties **with their reasons**, not just the steps
- [ ] Both reference the other, so a developer arriving from either side finds the whole rule
- [ ] The pagination-degradation table and the never-render-zero rule are both present
- [ ] What shipped is what is documented, including any deviation from this epic's design
- [ ] No new document is created

norbert-kulus-blockydevs:
Every paged list in the product asks the database two questions at once: *give me twenty rows* and *how many are there in total*. The first can stop after twenty matches. **The second cannot stop at all.**

This makes the total a **second stage**: the list renders immediately from the paged query, and the number arrives separately and replaces a `20+` placeholder when it lands.

Four stage commits, one per stage of the epic, plus five review-fix commits.

---

## What it measures out at

Re-measured against #2843's own 1M-row lab dataset. Full report and reproduction steps: `perf/openlinker-throughput/results-2943-two-stage-total-2026-09-06.md`.

**Time-to-rows and time-to-total, reported separately (p50, k6, 1450 requests, 0 non-2xx, n=220-263 per route):**

| | filtered (`?health=needs_attention`) | unfiltered |
|---|---:|---:|
| **before** - rows + total in one call | 184.29 ms | 45.49 ms |
| **time-to-rows** | **8.90 ms** | **8.84 ms** |
| **time-to-total** | 178.66 ms | 39.38 ms |

**The prediction held.** #2947 expected "rows in roughly 7 ms against a previous 149 ms". Measured **8.90 ms** against **184.29 ms** on this stand today - 1.9 ms above the predicted absolute, and a **20.7x** reduction, within a rounding of the 21x implied by #2843's own figures.

Two findings worth as much as the headline:

- **The page costs the same whatever the predicate** - 8.90 ms filtered, 8.84 ms unfiltered. A paged read stops after twenty matches, so the expensive `syncStatus @>` containment costs it nothing. The same predicate takes the *count* from 39.38 ms to 178.66 ms, because a count cannot stop.
- **The split is additive to within noise** - 8.90 + 178.66 against a combined 184.29. No database work was removed and none was expected to be. **A loader hides 178 ms from the operator without reducing it**; making the count itself cheaper is #2927's separate subject and this change does not touch it.

At the database, under one identical predicate: `Limit ... Index Scan Backward`, **0.286 ms**, 76 buffer pages - against `Parallel Seq Scan`, **175.814 ms**, 78 436 buffer pages in total (~613 MB, `Buffers:` being cumulative across the three workers unlike `rows` and `actual time`). 615x, visible in the plan rather than only on the clock.

### One deviation from #2947, stated up front

The re-measurement used a purpose-built k6 driver (`perf/openlinker-throughput/drivers/two-stage-total.js`), **not** `f5-read-path.sh`. The F5 harness lives on the performance-programme branch and this epic merges to `main`; running F5 would have meant importing the whole harness plus a `Dockerfile` LABEL its `guard_build` requires, neither of which belongs in a PR about pagination.

What was kept is F5's shape - one named `Trend` per route, per-route `n`, p99 reported, and **status checked before the sample is recorded** (the #2590 correction, without which a fast error counts as a fast sample). And the control is **inside the run**: the combined route is byte-identical to its pre-change self, measured in the same window by the same binary against the same dataset, so the before/after is an A/B rather than a comparison against a figure from another day.

---

## Stage 1 - CORE (#2944)

Five repositories can apply a predicate no plain index serves. Each now answers three ways: `findMany` (rows AND total, unchanged), `findManyRows` (the page alone), `countMany` (the total alone, taking no pagination).

**Shape chosen: a sibling method pair, not an optional flag** - the rejected alternative, as the issue asks. The acceptance criteria need both directions and a boolean gives only one; worse, a flag forces `total` to `number | undefined` on a result five ports and a dozen call sites already read as `number`, so every `?? 0` turns a "not asked for" into a rendered zero - the never-render-zero failure this epic exists to prevent, one layer down. `OfferMappingRepository` already carried that shape (`skipTotal` answering `total: -1`) and is the evidence; **this change removes that sentinel rather than propagating it**.

**`findMany` keeps the shape it already had.** Three of the five end in a single `getManyAndCount()` and still do, so `?withTotal=true` - the default, and every un-migrated caller - emits exactly the statements it emitted before, on the one query runner it emitted them on. Parity comes from a shared private `buildFilteredQuery`, not from re-running the read.

> **A correction the review made, recorded because it was believed before it was checked.** The first version of this PR justified that with a `lazyCount` fast path: `getManyAndCount` supposedly infers the total with no count query when a page comes back short. Read against the pinned `typeorm@0.3.17`, that is false - it awaits `executeEntitiesAndRawResults` and then `executeCountQuery`, unconditionally and sequentially, and the identifier `lazyCount` does not appear anywhere in that version. The count **always** runs, which is an argument FOR the split rather than against it. Eight code sites and one engineering standard carried the false claim and now carry the true one.

HTTP: `?withTotal=false` (shared `PaginatedReadQueryDto`, inherited not copied) plus a `/count` sibling per route (shared `PaginatedTotalResponseDto`), each count DTO an `OmitType` of its list DTO so the filter surfaces cannot drift. The response **omits** `total` when it was not asked for - never `0`.

## Stage 2 - the shared mechanism (#2945)

`usePaginatedTotal` + `<ListPagination>`, built once because four things must be right or the change is a regression. Each has a test that fails when the guarantee is removed, **all verified red-first**:

| Guarantee | Break that proves the test |
|---|---|
| debounce the filter key | remove it → 7 calls instead of 2 |
| never render a superseded total | add `keepPreviousData` → stale number renders |
| cancel the superseded request | stop forwarding the `AbortSignal` |
| delay the loader ~175 ms | remove the delay → flicker |
| a failed count is not `0` | return `0` on error |
| the pager degrades per affordance | gate Previous/Next on the total |

**One real design bug the tests caught**: the first draft armed the loader delay during the *debounce*, so a small install would flash a ~236 ms loader on every keystroke burst - precisely the flicker the delay exists to prevent, introduced by the delay.

`inferTotalFromPage` goes further than the issue asked: a page that came back SHORT already carries its own exact total, so most small installs never request the count at all.

## Stage 3 - adoption + re-measurement (#2947)

Four adoptions. Three took the mechanism unchanged; `/listings` needed two deviations, both documented in `use-listings-total.ts`:

- **Its count query is keyed WITHOUT `lifecycle`**, because the tab-bar buckets describe every tab regardless of which is selected. Keying on the tab would refetch and blank the bar - the behaviour #2029 explicitly required not happen. This **deletes** the hand-rolled ref + fingerprint the page carried to fake it.
- **A short page does not short-circuit its request**: the tab bar needs its buckets whatever the page size.

Adoption forced three mechanism changes, none cosmetic: `usePaginatedTotal` became generic with an explicit `selectTotal` (two observers on one key with different return shapes would corrupt the cache); a new `idle` state distinct from `unavailable` (only one may tell an operator the count *failed*); and `formatPaginatedTotal`, because two pages show the total in a second place.

The other eleven paged reads are untouched - they count over indexed equality, where the total is nearly free, and adopting there would add a loading state to pages that never needed one.

## Stage 4 - docs (#2950)

`docs/engineering-standards.md`: when a total is cheap and when it is not, the non-sargable shapes named in a table, the measured figures, and the split's four non-negotiable properties. Splits the remedy by shape - `ILIKE` can be made sargable with a trigram index (the `DestinationCategory.searchText` precedent), a jsonb containment cannot.

`docs/frontend-architecture.md`: the convention with the four mechanism properties **and their reasons**, the pagination-degradation table, and the never-render-zero rule. Names `keepPreviousData` as the one thing not to add.

Each references the other.

---

## Known costs, stated rather than discovered

- A list page issues **one extra HTTP request**. It issues no extra SQL: the count always ran (see the correction above). On `/orders`, `/products` and `/customers` a SHORT page issues no count at all - `inferTotalFromPage` derives the exact total client-side - so a small install gets strictly fewer statements than before. `/listings` is the exception and always asks, because its tab bar needs the buckets whatever the page size.
- The re-measurement is **one run** on a machine carrying three other Docker stacks, which is why the report leans on the in-run ratio rather than cross-day absolutes.
- Only `/orders` was measured. The other three surfaces are the same shape and are explicitly **unmeasured**, as the epic states.

## Verification

- `pnpm check:invariants` - exit 0
- `pnpm lint` - 0 errors
- `apps/web` `tsc -b --force`, `@openlinker/api type-check`, `@openlinker/core build` - clean
- `apps/web` vitest: **5404 passed**, 1 skipped, 496 files
- `libs/core` jest: **6043 passed**, 399 suites
- `apps/api` jest: **1997 passed**, 142 suites
- `apps/api` integration: the **whole suite** run locally under CI's own conditions (`maxWorkers: 1`, one shared Postgres, 25 files in sequence), where `paginated-total-split.int-spec.ts` passes
- Repository parity proved two ways: a fast unit spec asserting the three paths emit an identical `andWhere` sequence for all 23 orders filters (red-first against three breaks), and `paginated-total-split.int-spec.ts` proving it against real Postgres for all five repositories

## Review

Three lenses (backend / frontend / docs-and-measurement), repeated until a pass produced no BLOCKING or IMPORTANT finding. Both fix rounds are separate commits on this branch.

**Round 1** - 3 BLOCKING. The integration spec imported `@openlinker/core/customers/orm-entities`, which does not exist and is not a package export, so it had **never run**; it now seeds through the injected repository port and passes 15/15 against real Postgres. `GET /orders/count` would have 400-ed on every request, because `sort`/`dir` live inside `OrderFilters` and `forbidNonWhitelisted` rejected them. And `/listings` inferred a total from a `keepPreviousData` PLACEHOLDER page, printing "of 3" for a 900-row tab with Next disabled while the correct answer sat in the cache.

**Round 2** - 1 IMPORTANT that invalidated the epic's stated mechanism (the `lazyCount` correction above), 2 operator-visible frontend defects (a pager printing a range wider than its own total; a failed count leaving the tab bar spinning skeletons for the life of the page), a count re-keyed on `sort` so every column-header click re-ran the 142 ms aggregate, and four coverage gaps where a one-line mutation of new production code passed every test. Plus 223 lines of pre-existing Prettier drift reverted back out of the diff.

**Round 3** - 2 BLOCKING, both in round 2's own additions. An integration case inserted `inventory_items` rows referencing `product_variants` that were never seeded, so it violated a foreign key and could not run - CI reproduced the identical error before the fix landed. And a page test for the tab-bar fix used a FULL page, so the two states it exists to separate held the same value and reverting the fix passed. Plus a failed background refetch blanking a total still in hand, a placeholder guard firing on every transition rather than a tab switch, and a count narrowing its filters by enumeration.

**Round 4** - 1 BLOCKING: the tab-switch suppression was pinned by nothing at all - deleting it outright passed all fifty listings tests. Plus an omission guarantee defeated one file over by an enumerating URL builder (now a compile-time exhaustiveness map), the deep-link fix having only a helper unit test, and two controllers building their filter object four times with no parity assertion.

**Round 5** - the most serious finding in the epic, and one I had concealed. `slaState` mints `new Date()` inside the predicate, so after the split the page and the count bind two instants, and `hasNext` gated on the total alone could leave rows visible and unreachable. **The parity spec froze the clock and called it hygiene** - the flake was the finding. `hasNext` now overrules a total the rows have overrun, the spec asserts `slaState` is the only impure predicate, and the port says so. Also: `toEqual` ignores `undefined` properties, so three "complete literal" claims were false; and CI's Integration Tests failed because the suite reset only in `afterEach`, leaving its first case reading the previous FILE's rows - **41 absolute assertions were passing by execution order**.

**Round 6** - a regression against `main` that five prior rounds had all missed. `main` clamped the range summary (`Math.min(offset + PAGE_SIZE, total)`); the new component dropped it, so clicking Next onto the last page of a 1,234-row `/listings` tab rendered "Showing 1221-1240 of 1,234" - a range past its own total, deterministic on the final page-forward of any result set that isn't an exact multiple of the page size, and precisely the wrong-number failure this epic exists to prevent. The clamp is back and pinned. Three more rules turned out deletable: `hasNext`'s own primary rule (`offset + limit < total`) could be removed and every test in the component's file plus all four migrated page specs still passed, because none of them asserted Next at all; no `withTotal=false` path was tested at a non-zero offset, so `{ limit, offset }` → `{ limit, offset: 0 }` survived the whole suite - page 2 silently serving page 1's rows under a correct total, forever; and a search-variant pair asserted each side of a call with `toEqual` semantics, the exact weakness round 5 had already fixed elsewhere and missed here. A `libs/core` port docblock also overclaimed - it said `<ListPagination>` compensates "so no row is ever unreachable", which is false at exactly the boundary round 5 chose deliberately (`>` not `>=`), and is a core port promising that a React component patches a seam also reached by MCP tooling and curl; it now says MITIGATES and names the boundary.

**Round 7** - round 6's own clamp was wrong for three of its four beneficiaries. `offset + rowCount > total` has two distinct causes: a `keepPreviousData` PLACEHOLDER page (reachable only on `/listings`) or a total that genuinely undercounts the rows (reachable on all four, and the *only* reachable cause on the other three). Clamping the first is correct; clamping the second renders "Showing 1,001-1,005 of 1,005" over ten visible rows with Next enabled - an internally contradictory number that looks plausible, where the unclamped range at least showed the total was behind. Round 6 had turned a visible disagreement into a concealed one on three of four adopters. The component can't tell the two causes apart from its own props, so the caller now says which it is via `rowsArePlaceholder` (sourced from TanStack's own `query.isPlaceholderData`, not a hand-rolled fingerprint - the epic had already deleted one of those), and the same flag suppresses `hasNext`'s overrun clause, closing a matching defect where clicking Next during a `/listings` page transition navigated past the end. Round 6's own remedy had also reached only four of six sites and one of two fields: two rows-only reads (the listings `includeLifecycleCounts` branch and `GET /products/:productId/variants?withTotal=false`) carried no page-window assertion at all, and every controller spec asserted `limit` as the shared default (20), so `{ limit, offset }` → `{ limit: 20, offset }` also passed everywhere - both now pinned with a non-default limit and an explicit offset.

Every new assertion was verified **red-first** - the behaviour broken, the failure confirmed to name it, the break reverted. **Five separate tests passed against the exact mutation they claimed to catch** and had to be strengthened; each is recorded in the commit that fixed it. Three of my own verification claims were wrong and are corrected in place rather than quietly dropped: a churn detector that reported clean twice while broken, and the frozen clock above.

Declined, with reasons: `refetchOnWindowFocus: false` on the count (it would show a stale total beside rows the same focus event refetched - the first trap this epic exists to avoid); cloning the offer-mapping list builder (the repository spec returns one shared builder, so a stub clone proves nothing and a real one invalidates 26 assertions - the count-before-page ordering is pinned by a test instead); and widening the shared `no-unused-vars` config for destructuring discards (a repo-wide change riding inside a PR about pagination).

**Round 6 also reported, rather than fixed**, a gap outside this PR's scope: six int-spec files (`api-versioning`, `automation-dispatch-gate`, `bootstrap-admin-disabled`, `fulfillment-work-migration-parity`, `listings-offer-status-snapshot`, `oms-connection-never-seeded`) call `resetTestHarness` nowhere at all, so under `maxWorkers: 1` any of them running before an `afterEach`-only file reproduces round 5's CI failure. This PR defends only itself (`paginated-total-split.int-spec.ts` resets in both `beforeEach` and `afterEach`); the durable fix is a global `beforeEach` in `apps/api/test/integration/setup.ts`, filed as #2986 rather than folded into a pagination PR.

## External review

Two rounds of Tech Lead review from @piotrswierzy, both **Approve**, no BLOCKING or IMPORTANT findings on either pass. All SUGGESTIONs were addressed:

- Residual Prettier-only hunks in `order-record.repository.ts` - reverted, the file now reads as only the behavioural change.
- The `docs/lessons.md` / `docs/architecture-overview.md` merge conflict against `main` (an append collision in both - two new "lesson" entries and two new "### 27." sections landed at the same point) - resolved by keeping both entries and renumbering the later section to 28.
- This Review section, which had fallen behind the head by two rounds - extended through rounds 6-7 above.
- The six-file integration-harness reset gap - confirmed accurate and filed as #2986 rather than fixed inside this PR.
- The `hasNext` exactly-full-last-page fallback and the measurement's one-run/`/orders`-only scope were both flagged as informational, no change requested - reasoning affirmed and left as-is.

Closes #2944
Closes #2945
Closes #2947
Closes #2950
Closes #2943

https://claude.ai/code/session_01UWquhyubMheRRqdq7hDsZ7


piotrswierzy (approved):
## Tech Lead review — `perf: the paginated total becomes a second stage (#2943)`

**Scope reviewed**: all 92 files against `origin/main`. Depth-read: the five repository splits, `usePaginatedTotal` / `ListPagination`, the four controller `/count` routes and their DTOs, the `/listings` adoption, the cross-context allow-list entry, and the integration-harness truncate list.

---

### 1️⃣ Summary

**Approach**: ✅ Correct. The epic's design constraint — *the total is a second stage, never a substituted zero* — is enforced structurally at four independent layers rather than by convention: the API **omits** `total` (`paginated-orders-response.dto.ts`), the wire type **has no `total` field at all** so `data.total ?? 0` is a compile error (`shared/api/paginated-total.types.ts`), the hook returns `null` on both failure arms, and `ListPagination` renders `N+` rather than a skeleton. Each layer is independently sufficient; that is what makes the never-render-zero rule a property rather than a docblock.

**Safety**: ✅ Safe to merge. No schema change, no migration, additive HTTP surface, `?withTotal` defaults to `true` so every un-migrated caller is byte-identical.

**Documentation conflicts**: ✅ None.

---

### 2️⃣ Alignment with documentation

Verified against the actual tree, not the PR body:

- **Route ordering** — `@Get('count')` is declared before `@Get(':id')` / `@Get(':internalOrderId')` in all four controllers, and `/variants/search/count` lives on the separate `VariantsController` where no `:id` route can shadow it. This is the failure mode a `/count` sibling most commonly ships with; it is right in every one.
- **Route authorization** (`docs/frontend-architecture.md` § Route authorization / `packer-exclusion.spec.ts`) — each `/count` mirrors its own list route's decorator exactly (`@Roles('admin','operator','viewer')` on orders + customers, `@AnyRole()` on products + listings). `packer-exclusion.spec.ts` was updated in the same diff, so the discovery-based coverage gate actually sees the six new routes rather than failing closed on them.
- **Repository ports pattern** (`engineering-standards.md` § Repository Ports Pattern) — `findManyRows` / `countMany` are added to the *port* first and the parity comes from one shared private `buildFilteredQuery`, so the total cannot describe a different set than the page by construction. `buildPagedQuery` correctly adds the sort and the window *only* on the paged path; a count carries neither.
- **Cross-context imports** (`scripts/check-cross-context-imports.mjs`) — the new allow-list entry for `paginated-total-split.int-spec.ts` is justified in place and explicitly marked as having **no rewire target**, i.e. it must survive #722. That distinction matters now that #2791 fails the build on stale rows: without the comment the next #722 sweep would have deleted a live entry.
- **Integration harness** — `customer_projections` and its two satellites are added to `tablesToTruncate` with the correct children-first ordering, and the comment names why the `CASCADE`-closure walk (`testing-guide.md` § Testcontainers lifecycle) cannot reach them: both satellites hold indexed text references by value with no ORM FK. That is a real pre-existing gap the new exact-count assertion surfaced, fixed rather than worked around.

### Positive observations worth naming

- **The `lazyCount` correction.** The first version of this PR justified keeping `getManyAndCount()` on a claim about `typeorm@0.3.17` that is false. Reading the pinned version and then *reversing the argument in eight code sites and one engineering standard* — rather than quietly deleting the sentence — is the right handling, and the corrected comment in `order-record.repository.ts:findMany` now states the true reason (one query runner, one statement pair, unchanged for the default path).
- **`inferTotalFromPage`'s two `null` arms are the interesting ones**, not the happy path: a full page and an empty page at a non-zero offset both refuse to guess, and a non-finite input returns `null` rather than propagating `NaN` — correctly reasoned as *worse* than the `0` the mechanism already refuses, because every comparison against `NaN` is false and Next dies silently.
- **`lifecycleCountsState` on `/listings`** is a genuine design catch: `state` is overridden to `'known'` by a short page, and a short page implies nothing about the *other* tabs' sizes, so a tab bar branching on `state` would render skeletons for the life of the page. Two questions, two fields — and the second is derived from the buckets rather than from `stage.state`, which closes the no-tab-selected arm the first fix missed.
- **`cached` reads `query.data`, not `query.isSuccess`.** With this app's `refetchOnWindowFocus: true` / `retry: false` defaults, gating on success would turn a known total into `(count unavailable)` after one transient alt-tab failure while the correct answer sat in the cache. Correct, and non-obvious.
- **`aria-live="polite"` on the summary rather than only `aria-busy`** — a busy flag on a non-live region announces nothing, so `20+` → `1,234` would never reach a screen reader. And `(count unavailable)` is visible text, not a `title`, because a tooltip keeps neither the touch nor the screen-reader promise.

---

### 3️⃣ Findings

**[SUGGESTION]** — `libs/core/src/orders/infrastructure/persistence/repositories/order-record.repository.ts`

The diff carries Prettier reformatting unrelated to the change — trailing commas *removed* here (`(reason) => \`'${reason}'\``, `lineItemToOrm`, the `slaNow` `andWhere`), while new `apps/web` files in the same PR are written *with* trailing commas (`orders.api.ts:listRows`). That is the repo's known stale-`.prettierrc` drift rather than anything this PR did wrong, and the body records that 223 lines of it were already reverted back out. Not blocking — `pnpm lint` is green either way — but the residue makes the repository diff read as larger than the behavioural change, which matters for the one file where the parity argument lives. Reverting the remaining pure-format hunks would leave the diff saying only what it means.

**[SUGGESTION]** — `apps/web/src/shared/ui/list-pagination.tsx`

`hasNext` falls back to `rowCount === limit` while `total === null`. On an exactly-full **last** page this enables Next and a click lands on an empty page. The docblock acknowledges it ("once the total lands it takes over, which is strictly more accurate at the boundary"), and it is the correct trade — the alternative is disabling Next until the count arrives, which throws away the point of the split. Worth noting that under `totalState === 'unavailable'` the fallback is *permanent* rather than transitional, so that boundary click is reachable for the life of the page. No change requested; the empty page is recoverable and Previous still works. Flagging so the behaviour is a known cost rather than a discovery.

**[SUGGESTION]** — measurement scope

The report is one run on a machine carrying three other Docker stacks, and only `/orders` was measured. Both are stated up front in the body and in `results-2943-two-stage-total-2026-09-06.md`, and the control is *inside the run* (the combined route measured in the same window by the same binary against the same dataset), which is what makes the in-run ratio trustworthy where cross-day absolutes would not be. The deviation from `f5-read-path.sh` is justified — importing the F5 harness plus its `guard_build` `Dockerfile` LABEL into a pagination PR would be the larger sin — and F5's load-bearing property was kept: **status checked before the sample is recorded** (the #2590 correction). No action; recorded so the next reader does not quote 8.90 ms as a cross-machine constant.

**No BLOCKING findings. No IMPORTANT findings.**

---

### 4️⃣ Test coverage

**Unit**: ✅ `order-record.repository.predicate-parity.spec.ts` asserts the three paths emit an identical `andWhere` sequence across all 23 orders filters — which is the right assertion, because it tests the *property the split depends on* rather than each method's output. `use-paginated-total.test.tsx` and `list-pagination.test.tsx` cover each of the six guarantees with a stated break that proves the test.

**Integration**: ✅ `paginated-total-split.int-spec.ts` proves `countMany` / `findManyRows` describe the same set as `findMany` for all five repositories against real Postgres. That it had **never run** in round 1 (importing a `@openlinker/core/customers/orm-entities` sub-barrel that does not exist and is not a package export) and now seeds through the injected port is exactly the class of defect this repo's `check-*-mirror` gates exist for — a check that cannot fail. Caught and fixed rather than shipped green.

**Red-first discipline**: every new assertion verified red-first, and one round-2 test that *passed against the exact mutation it claimed to catch* was strengthened to assert the request rather than the call count. That is the specific failure `docs/testing-guide.md` § Port-contract suites warns about, applied outside a contract suite.

---

### Final Assessment

An unusually disciplined change: three review lenses repeated to a clean pass, both fix rounds preserved as separate commits, and — the part that matters most — a stated mechanism (`lazyCount`) checked against the pinned dependency, found false, and **corrected across every site that repeated it** rather than papered over. The declined items each carry a reason that is about correctness rather than effort (`refetchOnWindowFocus: false` would show a stale total beside freshly-refetched rows; cloning the offer-mapping builder would invalidate 26 real assertions to prove nothing).

The three findings above are all SUGGESTION-level and none needs to block.

### Merge Readiness

### Priority fixes

None required. Optional, in order: (1) revert the residual Prettier-only hunks in `order-record.repository.ts`; (2) nothing else.

piotrswierzy (approved):
## Tech Lead RE-REVIEW — force-pushed head `93fc2e37`

This supersedes my approval at `6942165c` (round 3). The branch was force-pushed after that review. **It was not rebased**: both heads share the same merge-base `c1e4090e`, so every difference is authored work, not main's. What changed is four new self-review rounds and a rewrite of the three that existed.

### What changed since the reviewed head

| | round-3 head `6942165c` | current head `93fc2e37` |
|---|---|---|
| commits | 7 (stages 1–4 + rounds 1–3) | 11 (stages 1–4 + rounds 1–7) |
| diff vs base | 92 files, +6023 / −553 | 92 files, +6567 / −495 |
| files differing between heads | — | 26 |

Rounds 4–7 are new. Same 92 files, so no new surface — the delta is corrections inside the surface I already read.

---

### Disposition of my three prior SUGGESTIONs

**S1 — residual Prettier-only hunks in `order-record.repository.ts`: ✅ addressed.** Verified against `c1e4090e`, not against the head-to-head diff (which is misleading here, because the force-push *reverts* round-3's drift and so shows the reformatting running the other way). Net vs main the file is now `+61 / −14`, and the only two deletions that are not substantive are the trailing commas on `(reason) => \`'${reason}'\``. The two port files are cleaner still — `order-record-repository.port.ts` is `+57 / −0` and `product-variant-repository.port.ts` `+40 / −0`, i.e. pure additions. The repository diff now says only what it means, which is what mattered for the one file the parity argument lives in.

*(For the record, so nobody chases it: the malformed `ProvisioningResumeDto} from …` import at `orders.controller.ts:141` is pre-existing on `c1e4090e` at line 138. This branch faithfully preserves it rather than introducing it.)*

**S2 — `hasNext` fallback enabling Next on an exactly-full last page: ⚖️ still present, deliberately, and now reasoned about rather than merely acknowledged.** The `total === null` arm is unchanged (`rowCount === limit`), which is the trade I said not to change. But rounds 5–7 rebuilt everything around it, and the reasoning that landed is right:

```ts
const hasNext =
  rowCount > 0 &&
  (total !== null
    ? offset + limit < total || (!rowsArePlaceholder && offset + rowCount > total)
    : rowCount === limit);
```

The `>` rather than `>=` is the load-bearing choice and the comment states why: at an exact multiple the rows *end* at the total, which is agreement, and enabling Next there blank-pages every well-behaved list. Only a page running *past* its total is evidence the total is stale.

**S3 — measurement scope (one run, `/orders` only): ⚖️ unchanged and still correctly stated up front.** No action.

---

### Review of what rounds 4–7 introduced

The substantive new production logic is `rowsArePlaceholder` + `rangeEnd` in `list-pagination.tsx`, and the four controller filter mappers. I read both against the tree.

**`rowsArePlaceholder` is correct, and the source of truth is right.** There is exactly one caller — `listings-list-page.tsx:1067`, `rowsArePlaceholder={query.isPlaceholderData}` — TanStack's own flag rather than a hand-rolled fingerprint. That matters: the epic already deleted one hand-rolled ref+fingerprint on this page, and reintroducing one to answer *"are these rows mine?"* would have been the same defect wearing a new hat.

The round-7 argument for making it a caller prop is the interesting part and it holds. `offset + rowCount > total` genuinely has two causes with opposite remedies, and the component cannot distinguish them from its props:
- a placeholder page (reachable on `/listings` alone) — the sum is an artefact, clamping is right;
- a total that genuinely undercounts (reachable on all four, and the *only* reachable cause on the three without a placeholder) — clamping renders `1,001–1,005 of 1,005` over ten visible rows with Next enabled, an internally contradictory number that looks plausible.

Round 6's unconditional clamp converted a visible disagreement into a concealed one on three of four adopters. Round 7 catching its own round-6 fix is the right outcome, and the asymmetry — `rangeEnd` clamps only under the placeholder, `hasNext`'s overrun clause suppresses only under the placeholder — is internally consistent.

**`rowCount > 0` verified as claimed.** The comment asserts it is unreachable through the four adopters. Checked: all four short-circuit to an `EmptyState` on `items.length === 0` before `<ListPagination>` renders (e.g. `customers-list-page.tsx:297`). So it is defence-in-depth at the component that documents `rowCount: 0` as supported, not a behaviour change — correct placement.

**The four filter mappers close a real drift class.** `toOfferMappingFilters` / `toOfferMappingBucketFilters` are the notable pair: the membership set is the base and the lifecycle narrowing is added on top, so a fifth membership filter is written once and reaches both — and the inverse (deriving buckets by deletion) is made unrepresentable by `OfferMappingCountFilters` declaring `lifecycle?: never`. That is a compile-time guarantee rather than a convention, which is the right shape for a rule whose violation is silent.

`formatPaginatedTotal(total, offset + rowCount)` at `list-pagination.tsx:170` passes the *unclamped* sum while the adjacent range uses `rangeEnd` — harmless, since that call is inside the `total !== null` branch and `formatPaginatedTotal` ignores its second argument when `total` is non-null. Noting it only so a future reader does not read it as an inconsistency.

---

### New findings

**[SUGGESTION]** — merge conflict in `docs/lessons.md`

`mergeable_state` is `dirty`. `git merge-tree` against current `main` (`ad6c2d1e`) reports exactly one conflict, in `docs/lessons.md` — an append collision, nothing structural. `docs/architecture-overview.md` auto-merges. Trivial to resolve; flagged only so it is not discovered at merge time.

**[SUGGESTION]** — the PR body is now behind its own head

The description's Review section ends at *"Round 5 — the most serious finding in the epic"*. The head carries rounds 6 and 7, and round 7 corrects round 6's own remedy — including the one operator-visible number (`Showing 1,001–1,005 of 1,005`) that a reader of the body would not know had been found and re-fixed. For a PR whose body is this load-bearing as the review record, the last two rounds should be in it.

**[SUGGESTION]** — six int-spec files still reset nowhere, reported not fixed

Round 6 reports this and I verified the list is accurate: `api-versioning`, `automation-dispatch-gate`, `bootstrap-admin-disabled`, `fulfillment-work-migration-parity`, `listings-offer-status-snapshot`, `oms-connection-never-seeded` call `resetTestHarness` nowhere at all. Under `maxWorkers: 1` against one shared Postgres, any of them running before an `afterEach`-only file reproduces round 5's CI failure — which on this branch was 41 absolute assertions passing by execution order. This PR correctly defends *itself* (`paginated-total-split.int-spec.ts` now resets in both `beforeEach` and `afterEach`) and correctly declines to fix the shared harness inside a pagination PR. Worth a follow-up issue for the global `beforeEach` in `apps/api/test/integration/setup.ts` so the report does not evaporate with this PR's description.

**No BLOCKING findings. No IMPORTANT findings.**

---

### Assessment of the re-review

My round-3 approval was premature in hindsight, and the branch is better for the author not stopping there. Four further rounds found, among other things: a wrong number this epic introduced against `main` (the dropped range clamp); a `withTotal=false` path where `{ limit, offset }` → `{ limit: 20, offset }` and `{ limit, offset: 0 }` both passed the entire suite, meaning page 2 could serve page 1's rows under a correct total indefinitely; `hasNext`'s *primary* rule deletable with every test still green; and a `libs/core` port docblock claiming a React component made rows unreachable-proof, on a route also reached by MCP tooling and curl. Each of those is the "check that cannot fail" class, and each was found by mutation rather than by reading.

The round-7 self-correction of round 6 is the strongest signal in the change: a fix that was *right for one of two causes* was identified as wrong for the other three surfaces and split, rather than defended.

### Merge Readiness

✅ **Approve** — the force-push is a strict improvement over the head I approved. Resolve the `docs/lessons.md` conflict and merge.

### Priority fixes

None blocking. In order: (1) resolve the `docs/lessons.md` conflict; (2) extend the body's Review section through rounds 6–7; (3) open the follow-up for the global integration-harness `beforeEach`.

norbert-kulus-blockydevs:
## Addressed both Tech Lead review rounds

@piotrswierzy — both reviews were **Approve** with no BLOCKING or IMPORTANT findings, but per your notes I've addressed every SUGGESTION from both passes:

### Round 1 (already resolved before round 2, confirmed unchanged)
- ✅ Residual Prettier-only hunks in `order-record.repository.ts` — you already verified this net `+61/−14` with only two non-substantive deletions.
- ⚖️ `hasNext` fallback on an exactly-full last page — left as the deliberate trade-off you reasoned through in round 2 (the `>` vs `>=` boundary is intentional and documented).
- ⚖️ Measurement scope (one run, `/orders` only) — left as stated up front in the PR body and the results doc; no change requested.

### Round 2 (new findings, all addressed)
1. **Merge conflict in `docs/lessons.md` + `docs/architecture-overview.md`** — `main` moved further since your review (`ad6c2d1e` → `07556bb8`), so a second conflict appeared in `architecture-overview.md` alongside the one you flagged in `lessons.md`. Both were pure append collisions (two new lesson entries, two new "### 27." sections landing at the same point). Resolved by keeping both entries/sections and renumbering the later one to `28`. Merged, verified clean (`grep` for conflict markers → none), and re-ran `pnpm lint`, `pnpm check:invariants`, `apps/api` type-check, and `apps/web tsc -b --force` against the merged tree — all clean. Pushed as `6944868e`.
2. **PR body behind its own head** — extended the Review section through rounds 6 and 7 (the range-summary clamp regression against `main`, and round 7's own correction of round 6's overly-broad clamp), plus a new "External review" section summarizing disposition of every finding from both your passes.
3. **Six int-spec files with no harness reset** — confirmed the list is accurate (`api-versioning`, `automation-dispatch-gate`, `bootstrap-admin-disabled`, `fulfillment-work-migration-parity`, `listings-offer-status-snapshot`, `oms-connection-never-seeded`). Filed as #2986 for the global `beforeEach` fix in `apps/api/test/integration/setup.ts`, rather than folding a shared-harness change into this pagination PR.

Branch is now up to date with `main` (merge commit `6944868e`), the full verification suite is green, and the PR body accurately reflects the full round 1–7 review history. Thanks for the thorough re-review — the round-7 self-correction catch on the placeholder-vs-undercount clamp was the sharpest finding in the whole series.

Issue #2373, opened 2026-08-22
[TASK] Infrastructure — shipments.direction and the four repricing predicates


piotrswierzy:
> Part of the OMS programme backlog (**Wave 2**, stream S1, size M). Backlog slug: `W2-36`. Sibling `W2-*` slugs resolve via the Wave-2 epic's checklist. Design of record: REVIEW §3 D8.

## Problem / Context

REVIEW §3 D8, deferred out of Wave 1c on purpose. Return labels are shipments, but every existing shipment predicate assumes one direction. Four repository predicates plus the partial unique index `UQ_shipments_branch_one_per_order_conn` (`libs/core/src/shipping/infrastructure/persistence/entities/shipment.orm-entity.ts:48`) must gain `direction`, or a return label collides with the outbound shipment for the same `(orderId, connectionId)` and one of them cannot be created.

## Proposed Solution

Add `shipments.direction` (`outbound | return`), **NOT NULL with a default of `outbound`** in the migration — and then **remove the default in the same migration** so no future insert can silently acquire it. The synchronize-default trap is the reason this is called out: TypeORM's schema sync would otherwise re-add a column default that makes `direction` unstated-but-present.

Recreate `UQ_shipments_branch_one_per_order_conn` with `direction` in its predicate, and add `direction` to the four repository read predicates in `libs/core/src/shipping/infrastructure/persistence/repositories/shipment.repository.ts` (the by-order listing, the branch-1 lookup, the status-sync scan and the dispatch-claim read), so no existing flow ever sees a return label.

## Classification

**Type**: Infrastructure
**Layer**: Infrastructure (**migration-bearing**; synthetic-sequential timestamp per `docs/migrations.md` rule 3)
**File(s)**: the shipment ORM entity + repository; new migration; `libs/core/src/shipping/domain/types/shipment.types.ts`

## Dependencies

- #2327 (`W1c-1`)

## Assumptions

- Return labels are bought through the same `ShipmentDispatchService` path; nothing about carrier integration changes here.

## Acceptance Criteria

- [ ] An outbound and a return shipment can coexist for one `(orderId, connectionId)` (integration test — this is the D8 regression)
- [ ] All four read predicates filter by direction; an existing flow never returns a return label (test each)
- [ ] The column carries no lingering DB default after the migration (asserted against the live schema)
- [ ] `migration:show` clean; tests added

Issue #1947, opened 2026-07-30
[BUG] CORE (Shipping) — waybill never reaches the order source: the dispatch relay's at-most-once gate fires before the carrier mints the tracking number


norbert-kulus-blockydevs:
## Problem / Context

A waybill (tracking number) minted by the carrier **after** the operator dispatches a shipment never reaches the order source. For an Allegro-sourced order fulfilled with InPost this is not a race — it is the normal ordering, so the marketplace is left showing "DODAJ NUMERY PRZESYŁEK" (add waybill numbers) permanently, on an order OL has already marked `WYSŁANE`.

Confirmed on the demo stack (backend from `main`), 2026-07-30:

| Where | State |
|---|---|
| Allegro order details | `status: WYSŁANE`, delivery section still offering **"DODAJ NUMERY PRZESYŁEK"** |
| OL `shipments` row | `status = dispatched`, `trackingNumber = 602222987610049018635781` (valid ShipX waybill, label downloaded fine) |
| Allegro HTTP log | `PUT /order/checkout-forms/{id}/fulfillment` → **204**. `POST /order/checkout-forms/{id}/shipments` → **never issued** (not a rejection, not a network error — absent) |

Order `ol_order_9900a108f7bd4a18a97bfcbdf3953b9d`, shipment `ol_shipment_6065bcd8f6cb432e8ea0fb7154df1019`, Allegro order `ac3735d0-8c14-11f1-997c-ed96a1f5709f`.

### Mechanism (verified in code)

1. **ShipX does not mint the waybill at create time.** `ShipXShipment.tracking_number` is `string | null` (`libs/integrations/inpost/src/infrastructure/types/inpost-shipx.types.ts:95`); the mapper's own comment says it arrives "once ShipX mints it at **confirmation**" (`.../mappers/inpost-shipx.mapper.ts:183-185`). `ShipmentDispatchService` persists it as-is and leaves the column NULL when absent (`libs/core/src/shipping/application/services/shipment-dispatch.service.ts:453-458`).

2. **The dispatch notification reads the row live and forwards whatever is there** — `trackingNumber: shipment.trackingNumber ?? undefined` (`libs/core/src/shipping/application/services/shipment-dispatch-notification.service.ts:106-109`). At that instant it is NULL.

3. **The Allegro adapter skips the waybill POST silently** and still reports success:
   ```ts
   // libs/integrations/allegro/src/infrastructure/adapters/allegro-order-source.adapter.ts:152
[...]

PR #2959, opened 2026-09-06
feat(shipping): `shipment_lines` — line-grain shipment records (#2727)

### LINKED ISSUE #2727: `shipment_lines`: line-grain shipment records so shipped/delivered quantities are derivable
Split out of **#2402**, which established that this is a separate concern rather than a sub-task, and that #2402 does not need it. Filed so it has an owner and its real obligations are recorded.

## Why it is not #2402's

`shipment_lines` does not exist anywhere in the tree — `libs/core/src/shipping` declares exactly one entity, `@Entity('shipments')`, and the only mention is #2392's deferral comment.

#2402's bridge instead lands as a nullable, indexed **`shipments.fulfillmentWorkId`**, for three reasons established during that issue:

1. **DESIGN §5.2 contradicts itself in one sentence** — *"Relationship to `Shipment` is **1:N** with the shipment keeping its identity; `shipment_lines` gains a nullable `fulfillmentWorkId`."* For a 1:N work→shipment relation the FK belongs on the many side (`shipments`). A line table would express M:N at line grain, which is not the stated relation.
2. **Different key, different motive.** This table originates in `DECISION-oms-fulfilment-grain` (option C) and ANALYSIS-1032 item 13, keyed `(shipmentId, orderId, lineId, quantity)` — with `orderId` deliberately present so **consolidated shipping across two orders** stays expressible. Its purpose is making `shipped_quantity` / `delivered_quantity` derivable, not linking work to shipment.
3. None of #2402's three acceptance criteria need line grain at all.

## Scope — the four obligations option C carries

These are why this is not a small addition:

1. **A lossy backfill, written as ledger events.** Existing shipments carry no line detail. Backfilling naively means a later cancel-and-reissue **double-counts permanently**, since the derived quantity has no way to distinguish a correction from a second shipment. The backfill has to be expressed as events, not as a snapshot.
2. **The `fulfillment-rollup.ts` precedence fix** — the current "any delivered ⇒ delivered" rule is wrong once quantities are derivable at line grain, because a partially-delivered shipment would report as delivered.
3. **Line counters**, with the same discipline the returns and fulfilment-work counters carry: an ordering invariant enforced by a DB `CHECK`, not by application code, so no caller can persist an impossible row.
4. **An FE line panel** — derivable quantities that no surface renders are a read model with no reader.

## One blocker is now cleared

The grain decision's stated prerequisite (**ANALYSIS-1032 item 14** — PrestaShop's positional `String(row.id || index)` line-id fallback, which made line identity unstable) **has been fixed**: the mapper now routes through `resolveOrderRowId`, which **throws on a missing id** rather than falling back to the array index. So line identity is now trustworthy enough to key on.

The other three obligations stand.

## Acceptance criteria

- [ ] `shipment_lines` exists, keyed `(shipmentId, orderId, lineId)` with quantity, so consolidated shipping across orders is expressible.
- [ ] The backfill is event-shaped and a cancel-and-reissue after it does **not** double-count — proven by a test that fails against a snapshot backfill.
- [ ] `fulfillment-rollup.ts`'s delivered precedence accounts for partial delivery.
- [ ] Counters carry a DB-level ordering invariant, not an application check.
- [ ] Every read scopes `direction` (#2373 / ADR-060) — a line read that treats an arriving return as a dispatch is the #2645 cross-body defect.
- [ ] Migration covered by `fulfillment-work-migration-parity.int-spec.ts`, the only automated check of a migration in this repo.

## Related

- **#2402** — established the analysis; ships `shipments.fulfillmentWorkId` instead.
- **#2392** — deferred the column, listing this as option (c) and rejecting it as "a whole new aggregate child with no consumer and no spec for its other columns".
- `DECISION-oms-fulfilment-grain` option C, ANALYSIS-1032 items 13 and 14.

piotrswierzy:
Closes #2727

`shipment_lines` (counters) + `shipment_line_events` (an append-only act ledger), so an order's shipped / delivered quantities become **derivable** rather than inferred from shipment status alone. Split out of #2402, which deliberately declined this table and landed `shipments.fulfillmentWorkId` instead.

## Acceptance criteria

| AC | Where |
|---|---|
| `shipment_lines` keyed `(shipmentId, orderId, lineId)` with quantity, so consolidated shipping across orders is expressible | `UQ_shipment_lines_shipment_order_line` |
| Backfill is event-shaped; a cancel-and-reissue does not double-count, proven by a test that fails against a snapshot backfill | `shipment-lines-backfill.int-spec.ts` |
| `fulfillment-rollup.ts`'s delivered precedence accounts for partial delivery | `deriveFulfillmentRollup` + spec |
| Counters carry a DB-level ordering invariant, not an application check | `CHK_shipment_lines_capacity` |
| Every read scopes `direction` (#2373 / ADR-060) | required parameter on the port, filtered by JOIN |
| Migration covered by `fulfillment-work-migration-parity.int-spec.ts` | both tables in `TABLES`, both FKs in `EXPECTED_DELETE_RULE` |

## Load-bearing properties

- **`orderId` is part of the line's IDENTITY, not a payload column.** `shipments.orderId` is the order the parcel was *dispatched for*. A consolidated parcel carries units from more than one order; derive the line's order from its shipment header and a line belonging to order B inside a parcel headed by order A becomes **unrepresentable** — order B's derived shipped quantity is permanently zero and nothing says so.

- **The backfill is EVENT-shaped, and that is the issue rather than a style note.** Order O, line L, quantity 2; `S1` dispatched then cancelled, `S2` the re-issue. A snapshot backfill writes `S1.shipped = 2` and `S2.shipped = 2` — order-level **4**, permanently and unfixably, because both rows claim to be shipments of the same units and nothing distinguishes a correction from a second shipment. The event backfill writes `S1: ship(2), cancel(2)` (net 0) and `S2: ship(2)` (net 2). The `cancel` act **is** that distinction. The int-spec computes the snapshot-shaped sum *alongside* the net and asserts they diverge, so it genuinely fails against a snapshot implementation rather than merely passing against this one.

- **`UQ_shipment_line_events_line_kind_occurred` carries the INSTANT.** A `(line, kind)` key looks right on the reasoning that a shipment ships once — but `ShipmentDispatchService` persists `failed` in its `generateLabel` catch and a retry re-enters on the **same** `shipment.id`, so `failed → generated → dispatched` is reachable on one row. Under that key a line that emitted `ship` + `cancel` could never record its successful re-dispatch, folding to a net shipped of 0 for a parcel that really shipped. `occurredAt` is the shipment's own instant falling back to its immutable `createdAt` — **never `updatedAt`**, which moves on every write and, being a key column, would mint a duplicate act on every reconcile.

- **`CHK_shipment_lines_capacity` deliberately OMITS `shippedQuantity <= quantity`.** Two ordinary paths violate it (re-ingestion rewrites `orderSnapshot` wholesale; the dispatch retry above folds to `shipped = 4, cancelled = 2` on a line of 2). Either would raise inside the reconcile's best-effort catch and silently stop the read model converging. `quantity` is **provenance**, not a bound; the bound that matters is `cancelledQuantity <= shippedQuantity`, which makes the net shipped quantity non-negative by construction. The parity spec asserts that omission **positively**.

- **The rollup fix has two separable halves.** The **status half** (an in-progress shipment outranks a delivered sibling) needs no line data and fixes every install, including one that never runs the backfill. The **quantity half** demotes a fully-terminal delivered order whose lines are only partly covered — the case the status half cannot see. The numerator over-attributes, and the rule survives that because it only ever **demotes**: an over-estimate that is still short proves the truth is shorter, so a false demotion is impossible and only a missed one. **Absent coverage means "no line data", never "zero delivered"** — reading it as zero would demote every pre-backfill order and quietly empty the delivered bucket.

- **One writer seam, and it is not unconditional.** `OrderFulfillmentProjectionService.recompute` is already called from all eight shipment-mutation sites, so no mutation service changed; `reconcile` takes the shipments that method already loaded (a `reconcile(orderId)` that re-queried would be `O(S² × I)` upserts). Three of those sites sit behind a `if (patch.status)` changed-patch gate — the **same** gate that produced #2402's unlinkable branch-1 row — so a tracking-only backfill (#1947) creates no missing line rows. Recorded rather than asserted away.

## Three defects the integration run found

The draft type-checked, linted and passed every unit test while carrying three defects only the real migration chain could expose. All three are now in `docs/lessons.md`.

1. **`FK_shipment_line_events_line` was unimplementable.** `shipmentLineId` was `text`; `shipment_lines.id` is a generated `uuid`, and Postgres refuses an FK across mismatched types. The sibling `shipment_lines.shipmentId` *is* correctly `text` (because `shipments.id` is an `ol_shipment_*` internal id), which is why the wrong type looked right. `synchronize` builds **no** FKs, so only the parity spec could catch this.
2. **The backfill seed violated `UQ_shipments_branch_one_per_order_conn`** by leaving `providerShipmentId` NULL on every seeded row — which claims the row is the branch-1 observed projection, and there can only be one per `(order, connection, direction)`. The index was right; the fixture modelled the wrong thing.
3. **The new negative parity assertion could not fail.** `not.toContain('"shippedQuantity" <= "quantity"')` names a string `pg_get_constraintdef` never emits (it renders lowercase identifiers unquoted), so the guard protecting the deliberate omission would have passed with the forbidden clause present.

## Limitations, stated rather than omitted

- **The FE line panel is not in this slice** (issue obligation 4, and not one of the six acceptance criteria). It needs a read API and a component. The read model is not written-and-never-read regardless: the rollup consumes the derived coverage on every recompute.
- **The backfill is lossy, by necessity.** Attributing each shipment the line's full ordered quantity is a guess for a genuinely multi-package order — OL never recorded the split, and no data in the tree can recover it. What the event shape guarantees is not that history is exact, but that a later cancel-and-reissue does not **compound** the error.
- **The coverage arm has a blind spot**: a partially delivered multi-package line, where the over-count hides the shortfall. That case is caught by the status half instead (the sibling package is still in progress). The two halves cover different failures, which is why both ship.
- **No `partial` rollup value.** `dispatched` is the honest approximation; a fifth value would need this union's SQL twins, the FE `deriveFulfillment` and every label map.
- **`reconcile` is not unconditional** — see the changed-patch gate above.
- **#2402's unlinkable branch-1 gap stays open and stays #2402's.** This read model joins on `orderId` and never consults `fulfillmentWorkId`, so it is out of that gap's path by construction rather than by luck.
- **`fulfillment_work_rejections` is missing from the integration truncate list** — a pre-existing gap noticed while adding the two shipment-line tables beside it. Not fixed here: it belongs to #2399 and would put an unrelated behaviour change in this diff.
- **`pnpm --filter @openlinker/api migration:show` was not run** — it needs the dev stack (ECONNREFUSED :5432). The parity int-spec runs the full migration chain with `migrationsRun: true` against real Postgres, which is stronger evidence.

## Gate

- `pnpm lint` — 0 errors
- `pnpm type-check` — 0 errors
- `pnpm check:invariants` — clean
- `pnpm test` — exit 0
- `fulfillment-work-migration-parity.int-spec.ts` — 9/9
- `shipment-lines-backfill.int-spec.ts` — 7/7
- Shipping unit specs — 39/39; FE `order-health.test.ts` — 37/37

Rebased onto `ad6c2d1ec` (#2958) and re-verified after the rebase.


piotrswierzy (commented):
# Tech Lead review — #2959 `shipment_lines`

*(Posted as a COMMENT rather than an approval/rejection only because GitHub refuses self-review on your own PR. **Verdict below is ❌ Request changes** — one blocking finding.)*

---

## Step 1 — Context

Reviewed all 26 changed files against `docs/architecture-overview.md` (§ 23 Shipment direction, § 26 Fulfillment's #2402 paragraph), `docs/migrations.md`, `docs/engineering-standards.md` and `docs/testing-guide.md`, plus the surrounding pattern on `main`: `fulfillment_work_lines` / `return_lines` (parent-part CASCADE FK declared migration-only, plain uuid PK, counters-not-statuses, DB `CHECK` twinned with a pure function), `fulfillment-work-migration-parity.int-spec.ts`, `ShipmentRepositoryPort`'s required-`direction` discipline, and both shipment status-sync services.

Layers touched: CORE domain (types, port, pure rollup), CORE application (service + the existing projection funnel), CORE infrastructure (2 ORM entities, 1 repository), Infrastructure/migrations, FE (`order-health.ts` twin only), docs.

**On the #2402 obligations.** § 26's #2402 paragraph names four: an event-shaped backfill, the `fulfillment-rollup.ts` precedence fix, line counters, and an FE panel. This PR delivers the first three and defers the fourth — stated plainly in the PR body under *Limitations*, and correctly noted as not being one of #2727's six acceptance criteria. That is honest deferral, not scope silence. The read model is not written-and-never-read either: the rollup consumes coverage on every recompute.

**On the key shape.** `(shipmentId, orderId, lineId)` matches the recorded decision verbatim, with `orderId` present as identity (`UQ_shipment_lines_shipment_order_line`) rather than as payload, and the reasoning is written where it is caused. Cross-aggregate refs (`orderId`, `lineId`, `productVariantId`) are indexed-by-value with no FK — the `order_changes` / `refund_records` / `returns.internalOrderId` precedent — while both genuine parent/part relations get real `ON DELETE CASCADE` FKs declared migration-only with no `@ManyToOne`, exactly the `fulfillment_work_lines` shape. Correct on both halves.

**Migrations.** `up()` + `down()` present on both; `{timestamp}-{description}.ts` naming correct; `1874000000000` / `1874000001000` both sort strictly above `main`'s tail (`1870000004000`), per `docs/migrations.md` rule 3. Parity coverage is present and is the strongest part of this PR: both tables in `TABLES`, both FKs in `EXPECTED_DELETE_RULE`, both explicitly asserted absent from the `synchronize`-built schema, both `CHECK`s named in the constraint set, and a *negative* assertion protecting the deliberately-omitted clause — with the `pg_get_constraintdef` quoting lesson (`"shippedQuantity" <= quantity`, unquoted) recorded so the guard can actually fail. That is precisely the class of defect the parity spec exists for, and it caught the `text` vs `uuid` FK type mismatch.

**Branch-1 rows** (`FulfillmentStatusSyncService.createBranchOneShipment`, no `providerShipmentId`) **do** get lines: `findByOrderId(orderId, 'outbound')` returns them and `reconcile` treats them like any other shipment. The backfill also covers them — its ship predicate carries the status disjunct precisely so a row born at its terminal status with no `dispatchedAt` still emits a `ship`. Good, and commented in place.

---

## Step 3 — Findings

### 🔴 BLOCKING

**1. A `deliver` act can be emitted with no `ship` act, which violates `CHK_shipment_lines_capacity` — aborting the backfill migration, and taking the pre-existing rollup write down with it at runtime.**
`apps/api/src/migrations/1874000001000-backfill-shipment-lines.ts` (deliver statement) and `libs/core/src/shipping/application/services/shipment-line.service.ts` (`buildActs`).

The two predicates are asymmetric:

```
ship    : dispatchedAt IS NOT NULL OR status IN ('dispatched','in-transit','delivered')
deliver : deliveredAt  IS NOT NULL OR status = 'delivered'
```

`deliveredAt IS NOT NULL` is **not** covered by the ship disjunct. A row carrying `deliveredAt` while its status has since moved off `delivered`, with `dispatchedAt` still NULL, emits `deliver` and no `ship` → `deliveredQuantity > shippedQuantity` → `CHK_shipment_lines_capacity` refuses the fold.

That shape is reachable on shipped code, not hypothetical:

1. `FulfillmentStatusSyncService.createBranchOneShipment` sets `dispatchedAt` **only** when the OMP status is `Dispatched`, and `deliveredAt` only when it is `Delivered` — so a branch-1 row born at `delivered` has `deliveredAt` set and `dispatchedAt` NULL.
2. `FulfillmentStatusSyncService.diffPatch` (and `ShipmentStatusSyncService.buildPatchAndMaybePush`, whose gate is `TerminalShipmentStatusValues.includes(...)`) will then project a later OMP `Cancelled` straight onto that row — neither goes through `ShipmentCancelService`, so `ShipmentNotCancellableException` never fires. The row ends as `status='cancelled'`, `deliveredAt` set, `dispatchedAt` NULL.

Consequences, in severity order:

- **The backfill migration fails.** Its fold `UPDATE` raises the CHECK, `up()` throws, the deploy stops. There is no best-effort catch on this path.
- **At runtime, the failure is not contained to the line model.** `foldCounters` throws inside `OrderFulfillmentProjectionService.recompute`'s single `try`, and `reconcile` is called *before* `updateFulfillmentState` — so the order's `fulfillmentState` is never written either. The docblocks repeatedly justify the omitted `shipped <= quantity` clause on the grounds that a raise "would be swallowed by the best-effort catch and silently stop the read model converging"; the same catch also silently stops the **pre-existing #1108 rollup projection** converging. That blast radius is understated everywhere it is described.

Fix is small and matches the model's own stated logic ("a `deliver` implies a `ship`", the mirror of the `cancel`-requires-`ship` rule that is already handled correctly): make delivery imply shipment on both sides.

```ts
// shipment-line.service.ts
const delivered = shipment.deliveredAt !== null || shipment.status === 'delivered';
const shipped =
  delivered || shipment.dispatchedAt !== null || SHIPPED_STATUSES.has(shipment.status);
```

```sql
-- 1874000001000, ship statement
AND (s."dispatchedAt" IS NOT NULL
     OR s."deliveredAt" IS NOT NULL
     OR s."status" IN ('dispatched', 'in-transit', 'delivered'))
```

Please also add a unit case to `shipment-line.service.spec.ts` and a migration case to `shipment-lines-backfill.int-spec.ts` for exactly this row shape (`status='cancelled'`, `deliveredAt` set, `dispatchedAt` NULL) — it is the shape neither suite currently constructs, which is why nine parity cases and seven backfill cases were all green.

Note this also invalidates the backfill's `cancel`-predicate comment, which argues the two conditions are "provably identical for this cohort" because the ship statement's status disjunct is dead for `cancelled`/`failed`. Once `deliveredAt` joins the ship disjunct that equivalence holds *more* strongly (the cancel predicate stays a subset), but the comment's reasoning needs the extra term spelled out or it becomes wrong-by-omission for the next reader.

---

### 🟡 IMPORTANT

**2. `reconcile` should not be able to sink the rollup write it precedes.**
`libs/core/src/shipping/application/services/order-fulfillment-projection.service.ts:55–72`.

Independently of finding 1: `upsertLines` / `recordActs` / `foldCounters` / `findOrderLineQuantities` are all unguarded, so *any* error from the new read model (a transient DB error, a constraint, a future writer bug) now costs the order its `fulfillmentState` update. Before this PR that write depended only on `findByOrderId` + a pure function. `readOrderItems` is carefully caught and degrades to "no line data"; the write path deserves the same posture, since the whole point of the optional coverage is that the rollup can answer without it:

```ts
let coverage: FulfillmentQuantityCoverage | undefined;
try {
  coverage = await this.shipmentLines.reconcile(orderId, shipments);
} catch (error) {
  this.logger.warn(`Shipment-line reconcile failed for ${orderId}; rolling up without coverage: …`);
}
const rollup = deriveFulfillmentRollup(shipments.map((s) => s.status), coverage);
```

The demote-only rule makes this strictly safe: dropping coverage can only miss a demotion, never cause a false one — which the `FulfillmentQuantityCoverage` docblock already argues.

---

### 🟢 SUGGESTIONS

**3.** `shipment-line.service.ts:38` — `ShipmentLineRepositoryPort` is imported as a value (and with stray formatting, `ShipmentLineRepositoryPort}` on its own line) while every other type on that file uses `import type`. It is an interface, so TS elides it; make it `import type` and merge it into the adjacent block.

**4.** The lossiness note applies to the **runtime writer**, not only the backfill: `buildLines` gives every shipment of an order every line at full ordered quantity, so `shipment_lines.quantity` is a per-package approximation *going forward*, not merely for history. The PR body says this in passing ("the backfill *(and the runtime writer)*"); it deserves to be the headline of whatever issue picks up obligation 4, because an FE line panel rendering these numbers verbatim would state a per-parcel split OL never recorded. Worth filing that follow-up now with the caveat written into it.

**5.** `docs/architecture-overview.md` § 26's #2402 paragraph now says #2727 "delivered the first three and deferred the panel" — good. Consider naming the follow-up issue number there once #4 exists, so the deferral has an owner in the doc rather than only in this PR body.

---

## Positive observations

- **The negative parity assertion, and the recorded reason it could not fail.** `not.toContain('"shippedQuantity" <= "quantity"')` naming a string Postgres never emits is exactly the "check that cannot fail" class, and catching it *plus* writing the quoting asymmetry into the spec as a `NOTE` is the right treatment.
- **`occurredAt` in the uniqueness key** is well-argued from a real shipped behaviour (`ShipmentDispatchService` reusing one row across a `generateLabel` retry), and the `never updatedAt` rule is stated in all four places a future editor might touch.
- **`direction` required on every read, filtered by JOIN rather than denormalised** — correctly refuses to create a second source of truth for the one column #2373 exists to be, and the port restates `ShipmentRepositoryPort`'s no-default rule rather than quietly diverging.
- **`setup.ts` truncate ordering** lists both children before `shipments`, with the reason (no FK in the `synchronize` schema, so the CASCADE closure walk reaches neither) written in place.
- **The FE twin asymmetry is documented on both sides** rather than left as apparent drift, and the status half is genuinely mirrored (clause reorder identical).

---

## Step 4 — Final assessment

### Summary
A carefully-reasoned, well-documented slice that delivers three of the four #2402 obligations and all six of #2727's acceptance criteria, with the strongest migration coverage in the repo and unusually honest limitation-reporting. The shape (identity key, counters-not-statuses, migration-only CASCADE FKs, DB CHECK twinned with a pure function, ledger-authoritative counters folded rather than incremented) matches the `fulfillment_work_lines` / `return_lines` precedent precisely. The one blocking concern is a predicate asymmetry that lets a `deliver` act exist without its `ship` — a state the CHECK correctly refuses, but which aborts the backfill migration outright and, at runtime, silently stops the *pre-existing* fulfillment-rollup projection rather than just the new read model.

### Merge readiness
❌ **Request changes**

### Priority fixes
1. **(Blocking)** Make `deliveredAt` imply a `ship` act, in both `ShipmentLineService.buildActs` and the backfill's ship statement; add the `status='cancelled' + deliveredAt set + dispatchedAt NULL` case to the unit spec and the backfill int-spec.
2. **(Important)** Wrap `reconcile` in its own catch inside `OrderFulfillmentProjectionService.recompute` so the line model can never cost an order its `fulfillmentState` write.
3. **(Suggestion)** `import type` on `ShipmentLineRepositoryPort`; file the obligation-4 follow-up carrying the runtime-writer lossiness caveat.

Issue #2383, opened 2026-08-22
[TASK] Frontend — returns nav, order-detail panel and timeline integration


piotrswierzy:
> Part of the OMS programme backlog (**Wave 2**, stream S3, size M). Backlog slug: `W2-45`. Sibling `W2-*` slugs resolve via the Wave-2 epic's checklist. Design of record: returns product spec §4/§7 (PR #2318).

## Problem / Context

Returns are reachable only from their own list until the order — where the operator actually starts — knows about them.

## Proposed Solution

A **top-level `Operations` nav item** for Returns (adjudicated; flagged to the owner as revisitable, but built that way). An order-detail `<div id="returns">` panel after the shipment/sales-document pair with one compact row per return, the attention badges, and the operator-authored `Record a return` action. Returns events (opened, declined, received, disposed, restock blocked, refund confirmed, credit note issued) added to `OrderActivityTimeline`, reusing the return-timeline renderers. `useWriteAccess` + `ReadOnlyLock` on every action; a demo-mode pass over the new surfaces.

## Classification

**Type**: Frontend
**Layer**: Interfaces
**File(s)**: `apps/web/src/app/**` (nav), `apps/web/src/features/orders/**`, `apps/web/src/features/returns/**`

## Dependencies

- #2376 (`W2-39`), #2378 (`W2-41`), #2335 (`W1c-8`)

## Assumptions

- The order timeline's `TimelineEvent` shape needs no change.

## Acceptance Criteria

- [ ] This issue touches `features/orders/**` and `features/returns/**` together, so its cross-feature import goes through `features/returns`'s **public barrel** — never a deep path into the returns folder. A deliberate deep import fails lint (both `.eslintrc.js` pattern groups, per #2335 / `W1c-8`)
- [ ] Nav item renders in `Operations` and is keyboard-reachable
- [ ] Order panel shows every return on the order with the same badges as the list
- [ ] Timeline distinguishes OL-owned acts from source observations via `by`
- [ ] Read-only and demo modes behave; component tests

piotrswierzy:
## Re-scope — this issue is the **timeline integration only**

Both ledger checks (*what supplies this?* and *what renders this?*) were run over this issue before planning, and all three of its deliverables had a premise problem. Two are resolved by removing them; the third is the real work.

### 1. The nav item is already shipped — dropped

[...]

Issue #2381, opened 2026-08-22
[FEATURE] Frontend — restock_blocked surfacing and remediation


piotrswierzy:
> Part of the OMS programme backlog (**Wave 2**, stream S3, size M). Backlog slug: `W2-43`. Sibling `W2-*` slugs resolve via the Wave-2 epic's checklist. Design of record: returns product spec §5.4 (PR #2318 — the returns spec is the **canonical copy owner** for `restock_blocked` strings).

## Problem / Context

The **highest-severity** surface in the returns spec. A restock that silently no-ops is worse than none — so the refusal must be impossible to miss, and impossible to mistake for success.

## Proposed Solution

Three required surfaces: a list badge + its segment, a **persistent inline line-row error** (not a toast), and an order-panel badge. Remediation copy names the quantity, the sku and the system that refused, offers `Mark stock handled manually` (the #2370 / `W2-33` attestation), and a "Why did this happen?" explainer per returns spec §5.4.

**Enforced by test: no surface can render "restocked" for units whose master write refused.**

## Classification

**Type**: Frontend
**Layer**: Interfaces
**File(s)**: `apps/web/src/features/returns/**`, `apps/web/src/features/orders/**`

## Dependencies

- #2376 (`W2-39`), `W2-42` (see epic checklist), #2378 (`W2-41`), #2357 (`W2-20` — copy module, RB-L), #2369 (`W2-32` — without the PrestaShop `adjustInventory` implementation `restock_blocked` is the *default* outcome, so the remediation surface cannot be exercised)

## Assumptions

- #2369 ships in the same window, so `restock_blocked` is the exception rather than the default on PrestaShop installs.

## Acceptance Criteria

- [ ] All three surfaces render for one blocked line, with the same title text
- [ ] A test asserts no component renders blocked units as restocked
- [ ] The attestation action clears the block and records who attested
- [ ] Component tests; usable at 375 px

piotrswierzy:
## Scope amendment — the third § 5.4 surface is split to #2640

**This issue ships two of § 5.4's three surfaces.** The AC as written (*"All three surfaces render for one blocked line, with the same title text"*) is amended: **read it as the two surfaces below**, with the third tracked in **#2640**.

### Why

The third surface has nothing to attach to, which was not visible from the issue text. Established by research before planning:

- `features/orders/components/` contains **no returns panel** — nothing matching `return` at all.
[...]

Issue #2378, opened 2026-08-22
[FEATURE] Frontend — return segments, filters and custody/money rails


piotrswierzy:
> Part of the OMS programme backlog (**Wave 2**, stream S3, size M). Backlog slug: `W2-41`. Sibling `W2-*` slugs resolve via the Wave-2 epic's checklist. Design of record: returns product spec §4/§5 (PR #2318). **Standing S3 rule**: `.nullish()`, never `.optional()` (#939).

## Problem / Context

Wave 1c ships the list and the detail shell. Wave 2 gives them the state they were built to show: the six segments, the filter set, and the **two independent rails** — custody and money — whose independence is the single most misread thing about the model.

## Proposed Solution

Six clickable `MetricCard` segments with the attention-worthy/routine tone split; every filter as a URL param with type-guard narrowing, `FILTER_PARAMS` / `hasActiveFilters` / one-call `clearAllFilters`, and offset reset on any change.

On detail: two labelled rails with the standing *"these move independently"* sentence. Custody renders over the five-state custody union (#2367 / `W2-30`). Money renders `in_doubt` as a first-class warning step with the "do not refund again" copy, and `refunded` renders **only** from an observation, attributed as "Confirmed by {source}".

## Classification

**Type**: Frontend
**Layer**: Interfaces
**File(s)**: `apps/web/src/features/returns/**`

## Dependencies

- `W2-40` (see epic checklist), #2335 (`W1c-8`), #2336 (`W1c-9`), #2371 (`W2-34`)

## Assumptions

- The list/detail shells from Wave 1c are extendable rather than needing a rewrite.

## Acceptance Criteria

- [ ] Segments are clickable filters; routine segments carry no danger tone
- [ ] `in_doubt` renders as a warning with the do-not-refund-again copy
- [ ] `refunded` never renders without an observation (test with a `triggered`-only fixture)
- [ ] Component tests; usable at 375 px

Issue #2372, opened 2026-08-22
[TASK] CORE — return.authorize, orphan match and operator-authored returns


piotrswierzy:
> Part of the OMS programme backlog (**Wave 2**, stream S2, size M). Backlog slug: `W2-35`. Sibling `W2-*` slugs resolve via the Wave-2 epic's checklist. Design of record: returns product spec stories T2/T3 (PR #2318).

## Problem / Context

Story T3's asymmetry: OL declines where the platform allows it, but **authorizes only returns it authored** — the model must not pretend OL decides what the marketplace already decided. Story T2's orphans need a way out of the bucket.

## Proposed Solution

Three writes:

- `return.authorize` — an ADR-044 proposal **restricted to `origin: 'operator_authored'`** returns, reusing the `order_changes` table built in #2333 (`W1c-6`). No second proposal mechanism.
- `matchOrphanToOrder(returnId, orderId)` — attributes an orphan, re-runs line resolution, and unblocks the downstream triggers Wave 1c blocked.
- `recordReturn(orderId, lines)` — an operator-authored return against a known order.

Every money-moving trigger stays blocked while `internalOrderId IS NULL` (#2332 / `W1c-5`'s rule, unchanged).

## Classification

**Type**: CORE
**Layer**: Application
**File(s)**: `libs/core/src/returns/application/services/**`; the `order_changes` port from W1c-6

## Dependencies

- #2332 (`W1c-5`), #2333 (`W1c-6`), #2370 (`W2-33`)

## Assumptions

- Matching is operator-driven here; the background re-attribution reconcile already exists (W1c-5).

## Acceptance Criteria

- [ ] `authorize` is refused on a `source_ingested` return with an explaining error
- [ ] Matching an orphan unblocks money triggers and records a timeline entry
- [ ] An operator-authored return participates in custody and money exactly like an ingested one
- [ ] No second proposal table or mechanism is introduced (asserted by review + grep test)
- [ ] Tests added; no boundary violations

piotrswierzy:
## Status after PR #2645 — AC2's rendering half is split to #2646; this issue stays open until that lands

Walking this issue's ACs against `2367-returns-custody`: everything is implemented **except the rendering half of AC2**, so no `Closes` trailer was added.

### What is done

- The orphan match **unblocks downstream triggers** — implemented and tested.
- The match is **recorded**: `returns.matchedAt` and `returns.matchedByUserId` are persisted and exposed on the write DTO.

[...]

PR #2960, opened 2026-09-06
feat(returns): order-detail returns panel and the return activity timeline

### LINKED ISSUE #2640: [TASK] FE — order-detail returns panel, and the order-scoped returns read it needs
> Split out of **#2381** (`W2-42`). Design of record: returns product spec § 5.4, third required surface.

## Problem / Context

Returns spec § 5.4 requires the `restock_blocked` remediation to render on **three** surfaces. #2381 ships two — the returns-list badge and the persistent inline line-row error with its attestation action — because both live in `features/returns` and are fully served by backend that already exists (`POST .../mark-stock-handled` and `listOutstandingRestockBlocks`, both from #2376).

The third surface has nothing to attach to, and that was not visible from the issue text:

- `features/orders/components/` contains **no returns panel** — nothing matching `return` at all.
- The order-detail page has **zero** returns presence: no panel, no `/returns` link, no import of the returns feature.
- There is **no order-scoped returns read**. `ReturnListFilter` carries `sourceConnectionId`, `bucket`, `createdFrom/To`, `stage`, `segment`, `openedFrom/To` — and **no `internalOrderId`**. The FE `ReturnFilters` mirrors that. (`internalOrderId` exists as a field on the returned *row*, not as a filter.)

So "add the same badge to the order-detail returns panel" is really a filter arm through core, DTO and repository, plus an FE query, plus a panel that does not exist, plus the badge. That is a deliverable in its own right, not a decoration on #2381 — which is why it was split rather than absorbed.

## Proposed Solution

1. **Order-scoped returns read.** Add an `internalOrderId` arm to `ReturnListFilter` (core), the request DTO, and the repository predicate; mirror it in the FE `ReturnFilters`. Small and additive — it composes with the existing filters like every other arm.
2. **Order-detail returns panel** in `features/orders`, listing the order's returns.
3. **The `restock_blocked` badge** on that panel, importing `restock-blocked-copy.ts` from the returns barrel (`from '../../returns'`) — **never a local copy**.

## Constraints

- **The copy module is created by #2381** at `features/returns/lib/restock-blocked-copy.ts` and exported from the returns feature barrel. This issue **imports** it. A second copy of the sentence must not exist; that module is the only home, which is the whole reason it was placed in the barrel ahead of its cross-feature consumer.
- The cross-feature import follows the shipped **#2100** precedent — `features/invoicing/lib/sales-document-block-copy.ts` is owned by one feature, exported via its barrel, and consumed by two surfaces in `features/orders` as `from '../../invoicing'`. Same shape here.
- Write affordances go through `useWriteAccess` + `ReadOnlyLock` per `frontend-architecture.md § Access Control And UI Visibility` — `visible` for demo viewers, never `canWrite`-gated into invisibility. Mixing the two answers within one surface is the failure mode that split exists to prevent (see #2380's demo-mode finding).

## Acceptance Criteria

- [ ] `internalOrderId` filters the returns list end to end (core filter → DTO → repository → FE), and composes with the existing filter arms
- [ ] The order-detail page renders a returns panel for an order that has returns
- [ ] An order with no returns renders a **distinguishable empty state** — never a panel implying the read failed, and never an absent panel implying the order cannot have returns
- [ ] A blocked line renders the `restock_blocked` badge with text **imported** from `restock-blocked-copy.ts`, byte-identical to the two surfaces #2381 ships
- [ ] No second definition of the copy anywhere in the tree
- [ ] Integration coverage for the new filter arm; FE tests for the panel and its empty state

## Dependencies

- **#2381** (`W2-42`) — creates the copy module and ships surfaces 1 and 2. This issue cannot start before it lands.

### LINKED ISSUE #2646: [TASK] FE — the return-detail activity timeline (spec § 5's unbuilt panel)
> Design of record: returns product spec § 5, which lists this panel in its panel order. Surfaced independently by **#2383** and **#2372**.

## Problem / Context

Returns product spec § 5 lists an activity timeline in the return-detail panel order. **It was never built.** There is no `*timeline*` file anywhere under `features/returns` and no `ReturnActivityTimeline` in the tree.

Two separate issues have now bumped into its absence, which is why it is being filed rather than deferred again:

- **#2383** planned to *"reuse the return-timeline renderers"* for the order-timeline integration. There were none to reuse — that turned out to be new work, and #2383 built the returns→**order** timeline path instead (`ReturnTimelineEntry`, the returns→`TimelineEvent` mapper, `mergeTimelineEvents`).
- **#2372** has an acceptance criterion requiring an orphan match to *"record a timeline entry"*. The fact is persisted (`returns.matchedAt`, `returns.matchedByUserId`) and exposed on the write DTO, but **there is no return-detail surface to render it on**.

## Why this is not just a missing panel

The data is already there and already shaped. #2383 built an order-scoped read of return **events** spanning four sources (`return_line_events` custody acts, `returns` header columns, `RefundRecord` via `IOrderRefundService`, with `credit note issued` explicitly deferred), plus a returns→`TimelineEvent` mapper and a `by`-attribution table that renders no actor it cannot name.

So this issue is largely a **re-projection of work that exists**, at the return grain rather than the order grain — which is also the trap. Do not assume the order-scoped read can be reused unchanged:

- #2383's read is keyed by `internalOrderId` and must skip returns whose `openedAt` is null (it returns per-return **contexts** covering every return on the order — a defect found in its own diff review, where a return that produced no entry of its own lost its refund entirely).
- A return-scoped read has no order to key on and needs no such context-covering, but **does** need to work for an **orphan** return — which by definition has no `internalOrderId` at all. #2372's own use case is precisely an orphan being matched, so an order-keyed read cannot serve it.

## Proposed Solution

1. A return-scoped read of return events, reusing #2383's four-source shape and its discriminator union (`custody_act | record_status | refund`) rather than inventing a second vocabulary.
2. `ReturnActivityTimeline` on the return detail, reusing #2383's mapper and `by` table.
3. The orphan-match entry (`matchedAt` / `matchedByUserId`) rendered, closing #2372's AC2.

## Constraints

- **Never invent an actor.** #2383's `by` rule is the contract: an act whose actor cannot be named gets **no eyebrow**, never a guessed one. `resolveBy` must be a total map over known values with `undefined` otherwise — a default arm on an open string ships a confident false attribution (found by independent review on PR #2645 as I-1, and worth reading before touching this).
- **A title must still render when the actor is unknown.** An unknown actor is not a reason to drop a fact.
- **`credit note issued` stays deferred** (#2383's ruling — it lives in `invoicing` / `order_changes`, a different bounded context). If this issue wants it, that is a cross-context decision to make explicitly.
- The read must work for an **orphan** return.

## Acceptance Criteria

- [ ] The return detail renders an activity timeline
- [ ] It works for an orphan return (no `internalOrderId`)
- [ ] The orphan-match entry renders, satisfying #2372's AC2
- [ ] No actor is asserted that cannot be named; an unknown actor renders no eyebrow and does **not** suppress the entry
- [ ] The event vocabulary is #2383's, not a second one
- [ ] A test asserts the panel renders **on the page**, not only in isolation — a component test cannot prove anything mounts the component (see `docs/lessons.md`)

## Dependencies

- **#2383** — built the mapper, the `by` table and the four-source read shape this reuses. Landed on PR #2645.

piotrswierzy:
Closes #2640
Closes #2646

Two operator surfaces the returns spec names and that were never built, plus the one backend arm each needs. Shipped together because they touch the same two feature barrels and shipping them apart means two consecutive edits to both.

## #2640 — the order-detail returns panel

Returns spec § 5.4 requires the `restock_blocked` remediation on **three** surfaces. #2381 shipped two. The third had nothing to attach to: there was no order-scoped returns read at all.

- **`ReturnListFilter.internalOrderId`** — core filter arm, request DTO, repository predicate, FE `ReturnFilters` + `buildQuery`. An **ordinary arm**, deliberately *not* a translation of `bucket: 'attributed'`: it necessarily selects attributed returns, but conflating the two dimensions is the #2378 `orphans` mistake, and `bucket` must stay independently usable beside it.
- It rides in the controller's shared `scope`, so **all four reads see it** — the page and the bucket/stage/segment counts. A count computed without it would describe every order's returns while the rows beneath describe one.
- **`OrderReturnsPanel`**, mounted unconditionally and keyed per order on the order-detail page.

## #2646 — the return-detail activity timeline

Spec § 5 lists this panel in its panel order. #2383 planned to reuse its renderers and found none; #2372 has an AC requiring an orphan match to record a timeline entry, with no surface to render it on.

- **`findTimelineEntriesForReturn`** is the *same query and the same projection* as #2383's order-scoped read, differing only in the `WHERE` — one shared `projectTimelineRows`, so the two timelines speak one vocabulary by construction rather than by agreement.
- `GET /returns/:returnId/events`, `useReturnEventsQuery`, `ReturnActivityTimeline`, mounted last in the return-detail panel order.

## Load-bearing properties

**It is keyed on the RETURN, which is the whole reason it is a second read.** An orphan has no `internalOrderId`, so the order-scoped read cannot serve it — and #2372's own use case is an orphan being matched. Pinned by an int-spec that asserts the column is null and the read still answers.

**The header loop grows from two kinds to five** (`opened` / `authorized` / `declined` / `matched` / `closed`), and the **ORDER timeline gains them too** — one vocabulary, not two. The plan originally deferred `authorizedAt`/`closedAt`; that reversed during review, because the reasoning ("#2383 chose two") does not carry once the same change widens that very vocabulary with `matched`.

**`resolveBy` keys on the actor's PRESENCE, never on a kind allow-list.** A list fails closed to the source claim, so the next actor-bearing header column would credit an operator's act to the channel — the confident false attribution `REFUND_EXECUTED_BY_COPY` exists to prevent. `matched` is the only such kind today; `opened` and `declined` still carry `actorUserId: null` and remain a source claim or nothing. Asserted in both directions.

**An unknown actor renders no eyebrow and does not suppress the entry.** A guessed actor is a claim; a dropped fact is worse than a bare row.

**An empty projection is a 404, never an empty timeline.** The join is a LEFT JOIN off `returns`, so a real return with no acts still yields one row; `[]` would render a history for a return that is not there.

**`GET /returns/:returnId/events` is `@Roles('admin','operator','viewer')`**, for the reason #2905 narrowed its order-scoped sibling: the entries carry a refund `amount` and `currency`, and a `packer` can obtain return ids from the `@AnyRole()` list. The FE renders that 403 as a stated permission fact, not a fault.

**Four states on both surfaces, none collapsed** — loading, a FAILED read, an unreadable envelope, and a confirmed empty. Only the last asserts anything about the operator's data. An absent panel would be a fifth wrong answer ("this order cannot have returns"), so the section always renders.

**The panel asks nothing when it has no order id.** `buildQuery` omits a falsy filter, so an empty id would issue an **unfiltered** `GET /returns` and render the whole installation's returns under one order's heading. `useReturnsQuery` gained an `enabled` argument; the unscoped branch states that nothing was asked rather than claiming the order has none.

**Truncation is a page-size fact.** An unreadable row is excluded from `items` and counted in `droppedCount`, so gating truncation on the item count reported the same row twice — once as unreadable, once as a page limit never reached. Fixed and pinned by a regression test.

**The badge reaches the panel by rendering `ReturnStageCell`** — the same component the returns list uses. That is one step stronger than importing `RETURN_RESTOCK_BLOCKED_COPY`: there is no second render that could drift.

**`ActivityTimelineList` is extracted byte-identically** from `OrderActivityTimeline`, with `ariaLabel` a **required** prop — extracted with the hardcoded string intact, the return timeline would announce itself to a screen reader as the *order* timeline. The order timeline's 27 DOM assertions are untouched and pass.

## Deliberate deviations, stated rather than hidden

**The panel lives in `features/returns`, not `features/orders`** as #2640's text says. `features/returns` already value-imports the orders barrel (`return-money-panel.tsx`), so a panel in `features/orders` that value-imported the returns barrel *and* was re-exported from the orders barrel would close a runtime barrel-to-barrel cycle (#337/#359). `pages/orders/order-detail-page.tsx` already imports from `'../../returns'`, so mounting it there adds no new edge direction.

**The panel file contains no import of `restock-blocked.copy.ts`** — `ReturnStageCell` owns it. That is stronger than #2640's AC, not weaker, and it is written down so a later reader does not "fix" the missing import and create the second reference the module exists to prevent. (The file is `restock-blocked.copy.ts`; the issue misnames it `restock-blocked-copy.ts`.)

**No `viewAll` deep link.** `internalOrderId` is not in `RETURN_FILTER_PARAMS`, so a link to `/returns` would land on an unfiltered list — worse than stating the truncation in words.

## Acceptance criteria the shipped model cannot express

**#2640: "A blocked *line* renders the badge."** It cannot. `ReturnListItem.restockBlocked` is a per-**return** boolean (a sibling of the counters; `null` means not reported). The panel lists returns, not lines, so it can only say *this return holds a blocked restock*. Per-line blocks live on the return **detail**, which this panel does not read.

## Limitations not closed

- `closed` is a **declared header kind nothing writes today** — `create` takes it and its one production caller passes `null`; `upsertFromSource` blanks it. Emitted anyway because the column is real and ADR-060 makes the four timestamps independent, but no operator sees that entry yet. Stated in the repository and in the docs.
- The role narrowing on the new route **does not close the wider hole**: `GET /returns/:returnId` is still `@AnyRole()` and already returns `refunds[]` with `amount` and `currency`, so a packer reaches the same money one route over. Pre-existing, not fixed here.
- The panel pays for three count aggregates it never renders (the controller fans out unconditionally). Accepted and noted rather than silently paid.

## Testing

- `apps/api/test/integration/returns-detail-timeline.int-spec.ts` — new. 404 on an unknown return, the orphan case with the null column asserted, the `matched` entry with its actor, `opened` carrying none, one header entry across three acts, and no sibling leakage.
- `returns-read-api.int-spec.ts` — three cases for the new filter arm: scoping across the page **and all three count aggregates**, composition with `sourceConnectionId`, and an unknown order id answering empty rather than 404.
- `returns-order-timeline.int-spec.ts` — unchanged and passing, which is what proves the shared projection and the two new header kinds broke nothing.
- FE: 11 panel cases, 5 page-mount cases on the return detail, 3 on the order detail, 4 mapper cases. Mounting is asserted **on the page**, not only in isolation (`docs/lessons.md` § "is this MOUNTED?").

## Gate

`pnpm lint` 0 errors · `pnpm type-check` 0 errors · `pnpm test` all green (1949 api, 694 worker, 890 prestashop, 703 allegro, 535 woocommerce, 108 oms, and the web suite) · the three returns int-specs 32/32.

Rebased onto current `main` (`ad6c2d1ec`) with no conflicts, and the full gate re-run on that base.


piotrswierzy (commented):
# Code Review: order-detail returns panel + return activity timeline (#2960)

**Verdict: 🔄 Approve with changes** — no BLOCKING issues. Posted as a `COMMENT` rather than an approval only because GitHub refuses self-approval on your own PR; treat this as an approve-with-follow-ups.

---

## 1️⃣ Summary

Two operator surfaces (#2640 order-detail returns panel, #2646 return-detail activity timeline) plus the one backend arm each needs: `ReturnListFilter.internalOrderId` and `findTimelineEntriesForReturn`. 34 files, +1726/-58, no migration (correctly — no schema change; the header kinds read columns that already exist).

Quality is high and the reasoning is unusually well-recorded in the code rather than only in the PR body. The two structurally important decisions are both right and both defended where a future reader will find them:

- **`findTimelineEntriesForReturn` is the same query and the same projection as the order-scoped read**, factored into `timelineRowsQuery()` + the static `projectTimelineRows`, differing only in the `WHERE`. That makes "one vocabulary" a property of the code rather than an agreement between two loops — and it is what let the header loop grow from two kinds to five with the order timeline gaining them for free.
- **`internalOrderId` is an ordinary filter arm, never translated into `bucket: 'attributed'`**, and it rides in the controller's shared `scope` so all four count reads see it. Both halves matter: the first avoids the #2378 `orphans` conflation, the second stops the chips describing a scope the rows do not.

The most important concern is cost, not correctness: the panel is unconditional on every order-detail render and fans out four aggregates, three of which are discarded.

---

## 2️⃣ Verification against the § 22 Returns contract

Checked specifically, all clean:

- **Custody and money are not collapsed.** The panel renders `ReturnStageCell` (the custody-derived stage, #2377) and no money state; the timeline renders acts, with refund entries sourced from real `RefundRecord` rows. Neither surface presents a single fused "return state".
- **`refunded` never renders without an observation.** The timeline's `refund_confirmed` entries come from `getRefundsForReturn` records with their own `recordedAt`; nothing derives a refund from a `triggered` state.
- **A blocked restock is not rendered as restocked.** `ReturnStageCell` owns that badge (`item.restockBlocked === true`), and the panel deliberately does *not* import `restock-blocked.copy.ts` — one component, three surfaces, no second render to drift. That is stronger than #2640's AC, and the docblock says so explicitly so nobody "fixes" the missing import.
- **An empty array is never a positive claim.** Four distinguished states on the panel (`unscoped` / loading / failed / `envelopeUnreadable` / confirmed-empty), and only the last asserts anything; the section renders unconditionally so absence cannot read as "this order cannot have returns". Backend-side, an empty projection is a 404 rather than an empty timeline — correct, because the `LEFT JOIN` off `returns` yields a row for a real return with no acts.
- **`in_doubt` semantics** are untouched by this diff (they live in `ReturnLineStateChip`); no regression.
- **Stage mirror (`scripts/check-return-stage-mirror.mjs`).** Neither side is touched — no change to `ReturnStageValues`, `RETURN_STAGE_PREDICATES`, or the SQL `CASE`. The mirror still holds vacuously.

Frontend rules also hold: `app → pages → features → shared` direction is respected, the two cross-feature edges (`returns → orders` for `ActivityTimelineList`, `pages/orders → features/returns`) both go through public barrels and both slugs are registered, no raw `fetch`, server state is TanStack Query throughout, all operator sentences live in `*.copy.ts` files under a `check-ui-vocabulary` scan root, and `apps/web/src/index.css` adds no raw hex.

Two things worth calling out as done right rather than merely done:

- **`useReturnsQuery`'s new `enabled` argument is a third parameter with a `true` default, not a caller-side early return.** `buildQuery` omits a falsy filter, so an empty order id would have issued an *unfiltered* `GET /returns` and rendered the whole installation's returns under one order's heading. That is a genuine trap and the fix is at the right layer.
- **`ariaLabel` is a required prop on `ActivityTimelineList`.** Extracted with the hardcoded `"Order activity timeline"` intact, the return timeline would have announced itself as the order's — the only label a screen-reader user gets. A required prop makes the wrong answer unspellable.
- **`resolveBy` keys on the actor's presence, not a kind allow-list.** A list fails closed to the source claim, which would credit the next actor-bearing column's operator act to the channel. Asserted in both directions.

Query-cache invalidation is also correct by inheritance rather than by accident: `returnsQueryKeys.all = ['returns']` prefix-matches `['returns','return-events',id]`, so the existing custody and decline mutations already refresh the new timeline.

---

## 3️⃣ Findings

### 🟡 IMPORTANT — the panel pays for three aggregates it never renders, on every order-detail view
`apps/web/src/pages/orders/order-detail-page.tsx` (mount) → `apps/api/src/returns/http/returns.controller.ts` `listReturns`

`listReturns` unconditionally fans out `Promise.all([listReturns, countReturnsByBucket, countReturnsByStage, countReturnsBySegment])`. The panel renders **none** of the three count sets — it reads `items`, `total`, `droppedCount` and `envelopeUnreadable` only. So mounting an unconditional panel on the order detail adds four queries per order view (including a `RESTOCK_BLOCKED_EXISTS` counters join) to serve a list that is empty on most orders.

The PR body discloses this ("accepted and noted rather than silently paid"), which is the right posture, but disclosure is not a fix and this is on the hottest operator page in the app. Two cheap options: an opt-out query param (`counts=false`) that skips the three count reads, or a dedicated narrow read. Either keeps the shared-scope invariant intact — the reason `internalOrderId` is in `scope` is so counts *that are computed* agree with the rows, not so they must always be computed.

Not blocking: correctness is unaffected and the cost is bounded per page view. Worth a follow-up issue rather than a change in this PR.

### 🟢 SUGGESTION — `ReturnTimelineUnreadableError`'s message still says "for this order"
`apps/web/src/features/returns/api/return-timeline.schema.ts:23`

> `'The return activity for this order could not be read.'`

The schema is now shared by the return-scoped read, where there may be no order at all (the orphan case this PR exists to serve). Cosmetic only — `ReturnActivityTimeline` renders `COPY.errorMessage` and never the thrown message — so this is a latent inaccuracy in a string nobody currently displays, not a user-visible defect. Reword to "This return activity could not be read." while both call sites are in view.

### 🟢 SUGGESTION — the return timeline's `unreadable` case is offered a retry that cannot help
`apps/web/src/features/returns/components/return-activity-timeline.tsx`

The panel distinguishes `envelopeUnreadable` from a failed read and correctly offers no retry for it. The timeline's parse failure throws, so it lands in the generic `query.error !== null` arm *with* a `Try again` button — but a schema mismatch is a fact about this build, and retrying re-issues a request that will fail identically. The panel already models this correctly one file over; a `query.error instanceof ReturnTimelineUnreadableError` arm (beside the existing 403 arm, which is handled exactly right) would make the two surfaces consistent.

### 🟢 SUGGESTION — `closed` is a header kind nothing writes
`libs/core/.../return.repository.ts` `projectTimelineRows`

Emitting it is defensible — the column is real and ADR-060 makes the four timestamps independent — and it is documented in the code, the PR body and `docs/architecture-overview.md`. Recorded here only so it is visible as a known-inert branch: today no operator can see that entry, and no test can assert it end to end through a production write path.

---

## 4️⃣ Tests

Good coverage, and coverage of the right things rather than the easy things:

- `returns-detail-timeline.int-spec.ts` (new, 7 cases) pins the 404-not-empty rule, the orphan case **with the null column explicitly asserted** (which is the whole reason the second read exists), the `matched` entry with its actor, `opened` carrying none, one header entry across three acts, and no sibling leakage.
- `returns-read-api.int-spec.ts` covers the new filter arm across the page **and all three count aggregates** — exactly the invariant the shared `scope` exists to protect.
- `returns-order-timeline.int-spec.ts` unchanged and passing is the meaningful evidence that the shared projection and the two new header kinds broke nothing.
- FE: mounting is asserted **on the page** for both surfaces, not only in isolation — the `docs/lessons.md` § "is this MOUNTED?" rule, applied.
- The `should NOT call an unreadable row a truncated page` regression test pins a real bug the review found (gating truncation on `items.length` double-reports a dropped row).

The `ReturnTimelineEntry` schema reads `kind` as a plain string, so the three new kinds pass through an older parser rather than being dropped as unreadable rows — worth noting because that is the failure this design avoided by having been written that way in #2383.

---

## 5️⃣ Deviations from the issues — all stated, all defensible

- **Panel in `features/returns`, not `features/orders`** (#2640's text). Correct: `features/returns` already value-imports the orders barrel, so the reverse edge re-exported from the orders barrel would close a runtime barrel↔barrel cycle (#337/#359). `pages/orders/order-detail-page.tsx` already imports from `'../../returns'`, so this adds no new edge direction.
- **"A blocked *line* renders the badge"** is unbuildable at this grain — `ReturnListItem.restockBlocked` is a per-return boolean and the panel lists returns. Stated plainly rather than papered over.
- **No `viewAll` deep link** — `internalOrderId` is not in `RETURN_FILTER_PARAMS`, so the link would land on an unfiltered list. Right call.

Documentation is updated in the right place and at the right altitude (`docs/architecture-overview.md § 22`), and no committed mockup exists for returns, so the mockup-sync obligation is vacuous here.

---

## Final Assessment

**🔄 Approve with changes.** Nothing blocks merge. Ship it, and open a follow-up for the discarded count aggregates on the order-detail read path; the three SUGGESTIONs are small enough to fold in now or leave.

**Priority:**
1. Follow-up issue: skip the three unrendered count aggregates for the order-scoped panel read (IMPORTANT — cost, not correctness).
2. Reword `ReturnTimelineUnreadableError`'s order-specific message.
3. Distinguish the unreadable-envelope arm on the return timeline (no retry button).

PR #2963, opened 2026-09-06
fix(prestashop-module): load WebhookSender before scheduleFastPathDrain() calls it (#2962)

### LINKED ISSUE #2962: fix(prestashop-module): scheduleFastPathDrain() calls WebhookSender before loading it, fatalling order validation
## What happens

`OpenLinker::scheduleFastPathDrain()` (`apps/prestashop-module/openlinker/openlinker.php`) calls
`WebhookSender::fastPathAvailable()` without first ensuring the class is loaded.

Every other `WebhookSender` call site in the module guards the reference:

```php
if (!class_exists('WebhookSender')) {
    require_once($classesDir . 'WebhookSender.php');
}
```

This one does not. `$classesDir` is only computed **after** the static call, so on any request where
nothing else has already pulled `WebhookSender` in, PHP raises a fatal
`Class "WebhookSender" not found`.

## Why it matters

`scheduleFastPathDrain()` is reached from `hookActionValidateOrderAfter` and
`hookActionUpdateQuantity`. A fatal there aborts **order validation** and the **stock-change hook**
outright — the shop's own checkout, not just OpenLinker's outbox drain. It is silent from
OpenLinker's side: no job, no webhook delivery row, nothing to see in the OL UI.

Whether it fires depends on autoload state, so it is load- and route-dependent rather than
deterministic — which is exactly why it survived review.

## Fix

Hoist `$classesDir` above the call and add the same `class_exists` / `require_once` guard the
module's other call sites use. Diff is 14 lines, no behaviour change beyond the guard.

## How this was found

Spotted during the #2809 review while auditing an unrelated diff. The fix was written on that
branch, then split out here on the reviewer's request: a PHP fatal in
`hookActionValidateOrderAfter` deserves its own reviewable change rather than riding inside a
frontend PR.

## Acceptance

- [ ] `scheduleFastPathDrain()` loads `WebhookSender` before its first static reference.
- [ ] `pnpm test:php` passes.

norbert-kulus-blockydevs:
## What changed and why

`OpenLinker::scheduleFastPathDrain()` called `WebhookSender::fastPathAvailable()` without the
`class_exists` / `require_once` guard **every other** `WebhookSender` call site in the module uses,
and computed `$classesDir` only *after* the static call.

On any request where nothing had already autoloaded the class, that is a fatal
`Class "WebhookSender" not found` raised straight out of `hookActionValidateOrderAfter` and
`hookActionUpdateQuantity` — so it aborts the **shop's own** order validation and stock-change
hooks, not merely OpenLinker's outbox drain. It is invisible from OpenLinker's side: no job, no
webhook delivery row, nothing in the OL UI.

Whether it fires depends on autoload state, which is why it is route- and load-dependent rather
than deterministic — and why it survived review.

The fix hoists `$classesDir` above the call and adds the same guard. 12 added lines, no behaviour
change beyond the guard.

## How to test

```
pnpm test:php
```

Or exercise a checkout on a shop with the module installed and confirm `hookActionValidateOrderAfter`
completes.

## Provenance

Found while reviewing #2809 and originally written on that branch. Split out here on the reviewer's
request: a PHP fatal in the order hook deserves its own reviewable change rather than riding inside
a frontend PR.

## Related issues

Closes #2962.

### DIFF
--- a/apps/prestashop-module/openlinker/openlinker.php
+++ b/apps/prestashop-module/openlinker/openlinker.php
@@ -2243,17 +2243,9 @@ class OpenLinker extends CarrierModule
 
         $classesDir = dirname(__FILE__) . '/classes/';
 
-        // #2962 - this call site referenced WebhookSender without the same
-        // class_exists/require_once guard every OTHER WebhookSender call site
-        // in this module uses, and $classesDir was computed only AFTER the
-        // static call. On a request where nothing had already pulled the class
-        // in, that is a fatal "Class WebhookSender not found" raised straight
-        // out of hookActionValidateOrderAfter / hookActionUpdateQuantity -
-        // aborting the shop's own order validation and stock-change hooks, not
-        // merely OpenLinker's outbox drain, and invisible from OpenLinker's
-        // side (no job, no delivery row). Whether it fires depends on autoload
-        // state, which is why it is route- and load-dependent rather than
-        // deterministic.
+        // #2962 - without this guard, a request where nothing has autoloaded
+        // WebhookSender yet fatals here, aborting hookActionValidateOrderAfter/
+        // hookActionUpdateQuantity; $classesDir must stay hoisted above this call.
         if (!class_exists('WebhookSender')) {
             require_once($classesDir . 'WebhookSender.php');
         }
--- a/apps/prestashop-module/openlinker/openlinker.php
+++ b/apps/prestashop-module/openlinker/openlinker.php
@@ -2241,11 +2241,26 @@ class OpenLinker extends CarrierModule
         }
         self::$fastPathDrainScheduled = true;
 
+        $classesDir = dirname(__FILE__) . '/classes/';
+
+        // #2962 - this call site referenced WebhookSender without the same
+        // class_exists/require_once guard every OTHER WebhookSender call site
+        // in this module uses, and $classesDir was computed only AFTER the
+        // static call. On a request where nothing had already pulled the class
+        // in, that is a fatal "Class WebhookSender not found" raised straight
+        // out of hookActionValidateOrderAfter / hookActionUpdateQuantity -
+        // aborting the shop's own order validation and stock-change hooks, not
+        // merely OpenLinker's outbox drain, and invisible from OpenLinker's
+        // side (no job, no delivery row). Whether it fires depends on autoload
+        // state, which is why it is route- and load-dependent rather than
+        // deterministic.
+        if (!class_exists('WebhookSender')) {
+            require_once($classesDir . 'WebhookSender.php');
+        }
+
         if (!WebhookSender::fastPathAvailable()) {
             return;
         }
-
-        $classesDir = dirname(__FILE__) . '/classes/';
 
         register_shutdown_function(function () use ($classesDir) {
             // Flush and close the buyer's connection now. Everything below
--- a/apps/prestashop-module/openlinker/tests/Unit/FastPathDrainWebhookSenderGuardTest.php
+++ b/apps/prestashop-module/openlinker/tests/Unit/FastPathDrainWebhookSenderGuardTest.php
@@ -1,0 +1,63 @@
+<?php
+
+use PHPUnit\Framework\TestCase;
+
+/**
+ * Regression test for #2962: scheduleFastPathDrain() referenced
+ * WebhookSender::fastPathAvailable() without the class_exists/require_once
+ * guard every other WebhookSender call site in openlinker.php uses. On a
+ * request where nothing had already autoloaded the class, that fataled
+ * straight out of hookActionValidateOrderAfter / hookActionUpdateQuantity -
+ * silent at runtime, so (like ModuleHookRegistrationTest) this reads the
+ * source rather than executing the module.
+ */
+class FastPathDrainWebhookSenderGuardTest extends TestCase
+{
+    /** @var string */
+    private static $source;
+
+    public static function setUpBeforeClass(): void
+    {
+        self::$source = file_get_contents(dirname(__DIR__, 2) . '/openlinker.php');
+    }
+
+    private function methodBody(string $signature): string
+    {
+        $start = strpos(self::$source, $signature);
+        self::assertNotFalse($start, $signature . ' not found');
+
+        return substr(self::$source, $start, 2000);
+    }
+
+    public function testScheduleFastPathDrainLoadsWebhookSenderBeforeFirstStaticReference(): void
+    {
+        $body = $this->methodBody('public static function scheduleFastPathDrain()');
+
+        $guardPos = strpos($body, "class_exists('WebhookSender')");
+        $callPos = strpos($body, 'WebhookSender::');
+
+        self::assertNotFalse($guardPos, 'scheduleFastPathDrain() no longer guards WebhookSender with class_exists');
+        self::assertNotFalse($callPos, 'scheduleFastPathDrain() no longer references WebhookSender::');
+        self::assertLessThan(
+            $callPos,
+            $guardPos,
+            'WebhookSender is referenced before the class_exists guard in scheduleFastPathDrain() - this fatals on any request where nothing has autoloaded it yet (#2962)'
+        );
+    }
+
+    public function testScheduleFastPathDrainComputesClassesDirBeforeUsingIt(): void
+    {
+        $body = $this->methodBody('public static function scheduleFastPathDrain()');
+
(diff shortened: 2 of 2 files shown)


piotrswierzy (approved):
## Tech Lead review — `fix(prestashop-module): load WebhookSender before scheduleFastPathDrain() calls it (#2962)`

**Scope**: one file, one hunk — `apps/prestashop-module/openlinker/openlinker.php` `scheduleFastPathDrain()`.

---

### 1️⃣ Summary

**Approach**: ✅ Correct, and the bug is real — verified against the tree, not the issue text.

`scheduleFastPathDrain()` (line 2146 pre-fix) referenced `WebhookSender::fastPathAvailable()` at line 2153 with `$classesDir` assigned at 2157, four lines *below* the static call. Every other `WebhookSender` reference in the file is guarded: 279–280, 539–540, 638–639, and the shutdown closure's own 2172–2173. This one was the sole exception.

The reachability claim holds too. All three callers (`self::scheduleFastPathDrain()` at lines 1942, 2040, 2129) sit in hook bodies whose own `require_once` blocks load `EventIdGenerator` and `OutboxRepository` only — never `WebhookSender`. So on a request where nothing else has pulled the class in, the fatal is raised out of `hookActionValidateOrderAfter` / `hookActionUpdateQuantity`, aborting the shop's own order validation rather than merely OpenLinker's drain. Load-dependent, hence non-deterministic, hence it survived review.

**Safety**: ✅ Safe. Purely additive guard; the existing duplicate guard inside the shutdown closure (2172) stays and is harmless — `class_exists` short-circuits.

**Documentation conflicts**: ✅ None.

---

### 2️⃣ Findings

**[IMPORTANT]** — no regression test, in a suite that already has the exact pattern for this class of bug

The acceptance criteria list only `pnpm test:php` passing, and no test was added. But `apps/prestashop-module/openlinker/tests/Unit/` already contains source-scanning specs written for precisely this failure mode. `ModuleHookRegistrationTest`'s own docblock states the rationale verbatim:

> The module file cannot be loaded here — it extends PrestaShop's Module and pulls half the framework with it — so these read the source. That is enough for the two mistakes that are **silent at runtime**.

An unguarded class reference in a hook path is that same category: invisible until an unlucky autoload state. A `file_get_contents(openlinker.php)` assertion that every `WebhookSender::` static reference outside a `class_exists`-guarded region is preceded by a `require_once` — or, narrower and cheaper, that `scheduleFastPathDrain()`'s body loads the class before its first reference — costs a few lines and pins the fix. Without it, the next edit to this method can silently reintroduce the same fatal, and nothing in `check:invariants` or `test:php` will notice.

`EndpointHardeningTest` and `SecretDisplayTest` are further precedents for a source-scan assertion in this suite, so this is applying an established pattern rather than proposing one.

**[SUGGESTION]** — comment length

The 12-line comment narrates the bug's discovery history (route-dependence, invisibility from the OL side, what the other call sites do). `engineering-standards.md § Code Quality Rules` asks comments to explain *why*, which this does — but no other guard in this file carries more than a line, and the full narrative already lives in #2962 and the PR body. Three lines naming the hazard (`fatal in hookActionValidateOrderAfter if nothing has autoloaded WebhookSender; $classesDir must be hoisted`) would carry the same warning to the next reader. Not blocking.

### Positive observations

- `$classesDir` is hoisted rather than duplicated as an inline `dirname(__FILE__)`, so the closure's existing `use ($classesDir)` capture is untouched and there is still one definition of the path in this method.
- The guard matches the surrounding call sites character-for-character (`class_exists` / `require_once($classesDir . 'WebhookSender.php')`), so it reads as the file's own idiom rather than a new one.
- Splitting this out of #2809 was the right call — a PHP fatal in `hookActionValidateOrderAfter` deserves an independently revertable commit rather than riding inside a frontend PR.

---

### 3️⃣ Test coverage

**Unit**: ⚠️ None added. See the IMPORTANT above — the suite has the pattern, the fix does not use it.

**Integration**: N/A. The `tests/Integration` suite needs a real MySQL and is excluded from the default run; it is the wrong home for this.

---

### Final Assessment

A correct, minimal fix for a genuine latent fatal in the shop's own checkout path, verified against the file rather than taken on the issue's word. The one gap is that a bug whose defining property is that it is *silent at runtime* ships without the source-scan assertion this module's test suite was built for.

### Merge Readiness

🔄 **Approve with changes** — the fix itself is right and I would not hold it behind the test if the order hook is currently fatalling in the field. Add the regression assertion in this PR if it can be done quickly, or as an immediate follow-up.

### Priority fixes

1. Add a `tests/Unit/` source-scan assertion pinning that `scheduleFastPathDrain()` loads `WebhookSender` before its first static reference (`ModuleHookRegistrationTest` is the template).
2. Optional: trim the comment to the hazard.

norbert-kulus-blockydevs:
Addressed both findings from the review:

1. **[IMPORTANT] Missing regression test** — added `tests/Unit/FastPathDrainWebhookSenderGuardTest.php`, following `ModuleHookRegistrationTest`'s source-scan pattern (the module can't be instantiated in the unit suite). It asserts, against `scheduleFastPathDrain()`'s body:
   - `class_exists('WebhookSender')` appears before the first `WebhookSender::` static reference.
   - `$classesDir` is assigned before it's used.
   
   Either regression — reordering the guard or the assignment — now fails `pnpm test:php` instead of silently reintroducing the #2962 fatal.

2. **[SUGGESTION] Comment length** — trimmed the 12-line narration down to a 3-line hazard note naming the failure mode and the invariant to preserve, matching the terseness of every other guard in the file.

Note: I wasn't able to run `pnpm test:php` locally (no PHP toolchain in this environment) — I traced the string-matching logic by hand against the actual file content to confirm both new assertions match as expected. Happy to have CI confirm.

piotrswierzy:
Drive-by finding from a parallel run on #2962 — **this fix leaves a second fatal on the same path**, and it is reachable from the same hooks.

Hoisting the load above the probe is correct and necessary. But `require_once` on a file that is not there raises **`E_COMPILE_ERROR`**, which `catch (Throwable)` cannot intercept — so on a module deployed without `classes/WebhookSender.php` this trades `Class not found` for a `require_once` fatal, in the same place, aborting `hookActionValidateOrderAfter` / `hookActionUpdateQuantity` and with them the shop's own checkout.

The fast path is an optimisation — the cron controller still owns those outbox rows — so the honest failure mode is *degrade to cron*, not *take the checkout down*:

```php
if (!class_exists('WebhookSender')) {
    $senderPath = $classesDir . 'WebhookSender.php';
    if (file_exists($senderPath)) {
        require_once($senderPath);
    }
}

if (!class_exists('WebhookSender')) {
    PrestaShopLogger::addLog(
        'OpenLinker: response-flush fast path disabled - classes/WebhookSender.php could not be loaded. '
            . 'Outbox events will be delivered by the cron controller only.',
        2, null, 'Module', null
    );
    return;
}
```

Three notes on that shape:

- **The re-check after the load is the load-bearing half.** Without it a `file_exists` guard just skips silently and the next line fatals anyway.
- **It must not be silent.** Bounded by the existing `$fastPathDrainScheduled` latch, so it logs at most once per request — otherwise the fast path stays off forever with nothing to find.
- **It covers an ABSENT file, not a corrupt one.** A truncated `WebhookSender.php` is a parse error and no guard here survives that. Worth stating in the comment rather than implying full coverage.

This also lets the closure drop its own `WebhookSender` require, which matters: the shutdown handler's `catch` reports through `WebhookSender::getErrorMessage()`, so a `require_once` failing *inside* that `try` would take the error reporting down with it. Same shape as the module's existing `OutboxRepository::worstCaseDeliverySeconds()`.

Two other things found while verifying:

**Severity is understated in #2962.** It reads as load-dependent; nothing on the hook path pulls `WebhookSender` transitively, and `composer.json` classmaps `classes/` under `autoload-dev` **only** — so this was reliably reachable, not situational.

**A regression test exists if you want it.** `apps/prestashop-module/openlinker/tests/Unit/ModuleFastPathDrainLoadOrderTest.php` on `2962-fix-fastpath-drain-class-load-order` (branch pushed, deliberately **no PR** — this one is yours). It pins the load-before-use ordering and asserts exactly the three call sites, so a fourth hook calling `scheduleFastPathDrain()` fails the build. Caveat stated plainly: there is no PHP runtime on the host it was written on, so `pnpm test:php` has **not** been run against it — its assertions were mirrored in Node and verified red against `origin/main` (13/13 pass on the fix, 7 fail on main), which proves it non-vacuous but is not a substitute for the real runner.

Happy to push the guard + test onto this branch, or leave both for you to take — your call.

Issue #2400, opened 2026-08-22
[TASK] CORE — progress ingress: `'fulfillment'` inbound domain + `fulfillment.work.statusSync` + `IFulfillmentProgressService`


piotrswierzy:
> Part of the OMS programme backlog (**Wave 3a**, stream S1, size L). Backlog slug: `W3a-11`. Sibling `W3a-*` slugs resolve via the Wave-3a epic's checklist.

> **Size justification (L):** a new `CanonicalInboundEvent` domain member is only meaningful with the job it routes to and the service that consumes it — the union, the job and the consumer must land in one change or the domain value dead-letters.

## Problem / Context

REVIEW §2 / G1 / C5 — the single most-hit convergent finding: progress had a contract and **no transport**. `CanonicalInboundEvent.domain` is a closed union with no fulfillment member, so a registered vendor decoder would authenticate every delivery and then dead-letter it — the documented eparagony trap (`architecture-overview.md` §16). The original adapter-side `report()` also pointed the wrong way: progress is **inbound**.

## Proposed Solution

- Add `'fulfillment'` to `InboundEventDomainValues` (**and `'return'`** for §7, so the closed union is grown once) with an `InboundRoutingPolicy` arm **gated on `FulfillmentExecutor`**.
- New core-owned job `fulfillment.work.statusSync` — **webhook-as-trigger, authoritative pull** (the shipped #904 discipline), keyed by the documented inbound idempotency shape.
- `IFulfillmentProgressService.record(event)` as the **single core-side ingestion seam**; `FulfillmentProgressEvent` = `picked | short_picked | packed | shipped | closed`, each carrying a **mandatory vendor-scoped `idempotencyKey`**, deduped by a `(workId, key)` claim row **claimed before any relay fires** (REVIEW C9).
- A polling vendor is served by `FulfillmentStatusSource.getFulfillmentStatus(workRef)` instead.
- `awaiting_wave` is deliberately **absent** from v1 (waving needs a claim/release entity — the ADR-045 `packGrain` lesson) and is the named first extension point.

## Classification

**Type**: CORE
**Layer**: Application + Interface
**File(s)**: `libs/core/src/webhooks/**` (domain union + routing policy), `libs/core/src/fulfillment/application/services/fulfillment-progress.service.ts`, `apps/worker/src/fulfillment/`

## Dependencies

- #2392 (`W3a-3`), #2398 (`W3a-9`)

## Acceptance Criteria

[...]

PR #2974, opened 2026-09-07
feat(fulfillment): enqueue `fulfillment.work.dispatch` for routed work (#2955)

### LINKED ISSUE #2955: [TASK] Fulfillment — `fulfillment.work.dispatch` has no producer, so routed work never reaches a bench
## Problem

`fulfillment.work.dispatch` is a fully built job type with **zero producers**.

Everything downstream of it exists:

- `JobTypeValues` carries `'fulfillment.work.dispatch'` (`libs/core/src/sync/domain/types/sync-job.types.ts`)
- `FulfillmentWorkDispatchHandler` is registered on the `realtime` lane (`handler-registration.service.ts`)
- `FulfillmentHandshakeService` drives `unsubmitted → submitted → accepted` with the `assignmentAttempt` claim (#2399)
- `OlFulfillmentExecutorAdapter` auto-accepts (#2409)

Nothing enqueues it. A repo-wide grep finds only the handler, the type entry, the registration, its own spec, and one doc comment in `oms.plugin.ts`. #2869's own decision comment says so out loud: *"nothing in the tree enqueues `fulfillment.work.dispatch` — a registered handler with zero producers. Manual routing is its first."*

## What this blocks

The bench work list filters on accepted work — `BENCH_WORK_REQUEST_STATUSES` in `apps/api/src/bench/application/bench-work-eligibility.ts`, story B1. Work whose `requestStatus` never leaves `unsubmitted` is invisible to it.

So with #2408's router wired, `RoutingCommitService` creates `fulfillment_works` rows all day and **none of them ever reaches a bench**. Wave 3a's exit criterion (#2412) — *"works the resulting tasks to shipped"* — is unreachable, and Wave 3b's pack bench stays empty on a routed install.

## The design question this issue owns

Deliberately not settled by #2408, because getting it wrong is a double dispatch:

- **Which host enqueues it?** The routing commit is reached from two places — `OrderIngestionService`'s intercept and `FulfillmentWorkRouteHandler`. Enqueueing from both is the two-bodies hazard again.
- **Under what idempotency key?** `work:{workId}:{assignmentAttempt}` is the handshake's key and is claimed by `claimDispatchAttempt`'s conditional UPDATE, which hands the attempt back through `RETURNING` precisely so a key cannot be minted without the row already holding that value (#2399). The enqueue key is a different question and must not accidentally re-mint one.
- **Does a re-route re-dispatch?** A router-driven re-request bumps `assignmentAttempt`; the producer has to respect that rather than fire per job attempt.

**#2869's M8 claims to be this first producer** (manual routing to the in-house bench). If that lands first, this issue may reduce to extending it to the router path; if #2408's path needs it sooner, the manual path must adopt whatever is built here. Either way it is one producer, not two.

## Out of scope

- Any change to `FulfillmentHandshakeService` or the `assignmentAttempt` claim.
- The timeout-as-rejection sweep for `submitted` work whose holder never answers (ADR-054 names it; it has no issue yet).

## References

- `apps/worker/src/sync/handlers/fulfillment-work-dispatch.handler.ts`
- `libs/core/src/fulfillment/application/services/fulfillment-handshake.service.ts`
- `apps/api/src/bench/application/bench-work-eligibility.ts`, `bench-work.service.ts`
- #2399 (the handshake), #2408 (router wiring), #2869 (M8), #2412 (Wave 3a epic), ADR-054

piotrswierzy:
`fulfillment.work.dispatch` was fully built — handler, lane registration, `FulfillmentHandshakeService`, the `assignmentAttempt` claim — and enqueued by **nobody**, so a routed `FulfillmentWork` stayed `unsubmitted` for ever, never reached `accepted`, and the pack bench (which filters on `accepted`) was permanently empty. Wave 3a's exit criterion was unreachable.

Both routing sites now offer every routed work object to its holder: `OrderIngestionService`'s intercept (#2396) and the `fulfillment.work.route` handler (#2408).

## The three design answers

**1. One decision, two mappings.** `deriveFulfillmentDispatchEnqueueIntents` (`libs/core/src/fulfillment`) is a pure derivation returning a *neutral intent*, not a request — the two hosts reach **different ports with incompatible request shapes** (`SyncJobQueuePort.enqueue` in core, `JobEnqueuePort.enqueueJob` in the worker). A shared request object would have forced one host onto the other's port.

**2. `connectionId` is the work's OWN `assignedConnectionId`, never a synthetic id.** #2609's defect was precisely a shared scope collapsing per-scope lane accounting installation-wide. A work carrying no holder is **skipped and warned**, never dispatched under a placeholder.

**3. The dedupe key is `fulfillment:dispatch:{workId}` — its own namespace**, never the handshake's `work:{workId}:{attempt}` (#2399), which is the key the *executor* dedupes on; sharing it would make an OL-side re-enqueue and a holder-side replay indistinguishable. `expectedAssignmentAttempt` is always `null` and is required to be: `claimDispatchAttempt` bumps the counter itself, so a pre-claim value would make this producer's own retry refuse to resume.

## The enqueue is per intent, and the producer never throws

One `enqueue` **per work** with a **per-intent catch**. `enqueueBulk`'s docblock disclaims atomicity and `SyncJobQueueService` really is a sequential `for`, so a bulk call plus one error log naming every work would be a false statement about the ones that *did* dispatch — in the single artefact an operator reads to decide what to re-drive. A failed enqueue therefore names one `workId` and its siblings still go out.

The whole method body sits inside a `try`. On the ingestion path it runs inside `interceptFulfillmentRouting`'s **fail-open** try, whose catch returns `{held: false}` — an escaping throw would mirror to every destination an order whose work rows are already committed, i.e. fulfil it twice. A spec pins that the order is still HELD when the enqueue fails.

## Limitations left open, stated rather than discovered later

- **The dedupe key has no generation component.** `sync_jobs.idempotencyKey` is globally unique and TTL-less on the Postgres gate path, so a **second** dispatch for the same work row is silently swallowed. That is correct while nothing re-requests; the day a re-source / re-request path lands (#2395), the key MUST gain a generation component. Recorded as a constraint on `buildFulfillmentDispatchDedupeKey` itself, not only here.
- **The `fulfillment.work.route` call site is unreachable in production today** — that job type still has no producer of its own. It is wired and tested anyway (the #2400 posture): the handler is registered and may gain a producer, and a route handler that silently fails to dispatch is strictly worse than an unexercised branch. Do not "clean it up" as dead code without also giving `fulfillment.work.route` a producer.
- ADR-054's timeout-as-rejection sweep is still unowned: a `submitted` work whose holder never answers stays `submitted` and nothing reaps it.

## Gates

Run on Node 22 LTS (Node 25 breaks 63 `apps/web` tests via its own `localStorage`, unrelated to this change).

- `pnpm lint` — pass (includes `check:invariants`; every mirror script green)
- `pnpm type-check` — pass, zero errors
- `pnpm test` — **exit 0**, every package green:
  - `libs/core` — `Tests: 6064 passed, 6064 total`
  - `apps/worker` — `Tests: 703 passed, 703 total`
  - `apps/api` — `Tests: 1949 passed, 1949 total`
  - `apps/web` — `Test Files 493 passed (493)`
  - plus `libs/shared` 266, `libs/oms` 108, and every integration package (allegro 703, prestashop 890, woocommerce 535, erli 503, ksef 402, infakt 250, subiekt 233, inpost 155, dpd-polska 133, eparagony 110, fx 106, ai 34, plugin-sdk 11, test-kit 11)
- Integration tests: **not run locally**, deferred to CI.

Closes #2955


piotrswierzy (commented):
# Code Review: `fulfillment.work.dispatch` producer (#2955)

**Verdict: 🔄 Approve with changes** — no BLOCKING issues. (Posted as a COMMENT rather than an approval only because GitHub refuses self-approval; treat it as an approve-after-IMPORTANTs.)

I reviewed the full diff (13 files, +1026/-20), read #2955, and checked the claims against the shipped `FulfillmentHandshakeService`, `FulfillmentWorkDispatchPayloadV1`, `RoutingCommitService`, `sync-worker.module.ts` and `docs/architecture-overview.md` § 26.

---

## Step 3 — Findings

### ✅ The load-bearing safety properties hold

Verified rather than assumed:

- **`assignmentAttempt` is not re-minted outside the claim.** The producer's payload carries `expectedAssignmentAttempt: null` and *no* attempt value is derived, read or guessed anywhere in the new code — `RoutedWorkRef` deliberately omits the field, and `claimDispatchAttempt`'s `RETURNING` remains the only route to an attempt. The handshake and the claim are untouched, as #2955 required.
- **Scope is correct.** `connectionId` is the work's own `assignedConnectionId` (`routing-commit.service.ts:395`, threaded through `RoutedWorkRef`), never a synthetic id — the #2609 defect is avoided, and an unassigned work is *skipped and named* rather than dispatched under a placeholder. Skipping (rather than enqueueing) is also right: `FulfillmentHandshakeService.dispatch` throws the retryable `FulfillmentWorkUnassignedError`, so an enqueue there could only burn the ladder to a dead row.
- **Dedupe namespace is separate.** `fulfillment:dispatch:{workId}` vs the handshake's `work:{workId}:{attempt}` — sharing them would make an OL-side re-enqueue and a holder-side replay indistinguishable, exactly as the docblock argues. Mirrors the existing `buildFulfillmentCancellationIdempotencyKey` precedent in the same service.
- **Two call sites, one dispatch.** Belt and braces: a second entry answers `skipped/'already-routed'` (`routing-commit.service.ts:229,248`) and never reaches the enqueue, *and* the globally-unique `sync_jobs.idempotencyKey` would swallow it anyway.
- **ADR-053 leaf intact.** The new `fulfillment-dispatch-enqueue.types.ts` imports nothing at all; the decision is a neutral intent and both hosts map it onto their own port. This is the correct side of the seam — the report-don't-perform shape `IFulfillmentProgressService.record` already established. The `orders → fulfillment` edge already exists in the architecture map; no new edge, no new allow-set entry.
- **ADR-007.** No outcome classification changed; the producer only enqueues, and the handler's existing terminal/retryable split is untouched.
- **Pure-rule file placement.** Runtime functions in a `*.types.ts` are legal here under `engineering-standards.md § The pure-rule exception` — pure, it *is* the rule for the type it sits with, and both halves change together. `pricing-rule.types.ts` is the precedent.
- **`JobEnqueuePort` is resolvable** in `SyncWorkerModule` (nine sibling handlers already inject `JOB_ENQUEUE_TOKEN`), so the new constructor dep will not fail boot.

The per-intent loop with a per-intent catch (instead of `enqueueBulk` + one aggregate log) is the right call, and the reasoning given — a bulk failure log naming works that *did* dispatch is a false statement in the single artefact an operator re-drives from — is correct about `SyncJobQueueService`'s sequential `for`.

---

### 🟡 IMPORTANT 1 — a lost enqueue is permanent, and nothing reconciles it

`libs/core/src/orders/application/services/order-ingestion.service.ts` (`enqueueRoutedDispatchJobs`) and the mirror in `fulfillment-work-route.handler.ts`.

The swallow is right (the method must not throw inside the fail-open intercept — an escaping throw returns `{held: false}` and mirrors an order whose work rows are already committed, i.e. a double fulfilment). But its own docblock states the consequence: *"a retry re-enters `route()`, which answers `already-routed` and reaches no enqueue at all"*. I confirmed that at `routing-commit.service.ts:229,248`.

So the enqueue is **not** enqueue-then-commit in the #2218 sense — the commit is already durable and the enqueue is a best-effort tail with **no re-drive path anywhere in the tree**. A Redis/Postgres hiccup on one `enqueue` leaves a `FulfillmentWork` permanently `unsubmitted`: invisible to the bench (`BENCH_WORK_REQUEST_STATUSES`), never retried, discoverable only by grepping an `error` log. That is precisely the failure mode #2955 exists to remove, reintroduced at a lower probability.

The `in-doubt` commit arm compounds it — the router may have committed on its side, the decision row stays `live`, and the producer returns early, so any work rows behind it are likewise never offered.

The PR names ADR-054's timeout-as-rejection sweep for `submitted` work as unowned. There is a **second, distinct** gap here: a sweep (or reconcile pass) over `requestStatus = 'unsubmitted'` work older than N. I am not asking for it in this PR — but it should be an issue filed before merge, and named in the "Limitations left open" section beside the `submitted` one, rather than left to be discovered when a bench is mysteriously short one task.

---

### 🟡 IMPORTANT 2 — the `#2395` constraint is recorded on the key but not on `expectedAssignmentAttempt: null`

`libs/core/src/fulfillment/domain/types/fulfillment-dispatch-enqueue.types.ts` (the `buildFulfillmentDispatchDedupeKey` docblock) records that the key must gain a generation component the day a re-request path lands. Good — and correctly recorded on the function itself, not only in the PR body.

But `null` has a **second** effect the docblock does not mention. In `claimOrResume` (`fulfillment-handshake.service.ts:230`), the delayed-duplicate guard is `if (expectedAttempt !== null && current.assignmentAttempt !== expectedAttempt)`. Passing `null` therefore *disables* attempt-scoping: a delayed duplicate of this producer's own job, waking after a re-request bumped the work to attempt 2, would resume unconditionally on `submitted` and mint attempt 2's key — the exact hazard #2399 added the field to prevent.

This is **unreachable today** (nothing bumps `assignmentAttempt`, and the TTL-less unique key means no second job row can exist), and `null` is genuinely the only correct value for a first-dispatch producer — the docblock's argument against `0` is sound. The gap is documentation, not behaviour: the #2395 constraint must say that the key and the `null` have to change **together**, or a future author fixes the key, believes the generation problem solved, and leaves the attempt guard off. One sentence on the `FulfillmentDispatchEnqueueIntent` docblock (which already explains why `null` is required) closes it.

---

### 🟢 SUGGESTION — `findUndispatchableWorkIds` walks the same list twice

Both hosts call `findUndispatchableWorkIds(outcome.works)` and then `deriveFulfillmentDispatchEnqueueIntents(outcome.works, orderId)`, each filtering on the same `assignedConnectionId === null` predicate. Two functions cannot disagree today (both are three lines), but a single derivation returning `{intents, skipped}` would make "reported === enqueued" structural rather than conventional — the shape the repo already prefers for `getStreamConcurrency` (#2229) and the reported-vs-enforced rule generally. Not worth blocking; worth noting if a third caller appears.

### 🟢 Positive — the unreachable branch is the right call

Wiring `fulfillment.work.route`'s dispatch despite that job type having no producer is correct (#2400's posture), and the docblock's *"do not clean this up as dead code without also giving `fulfillment.work.route` a producer"* is exactly the note that stops a future reader deleting it. Same for the `orderId` payload field being documented as ship-to-loading only, never the authority on ownership.

### Docs

`docs/architecture-overview.md` § 26 is updated. Given IMPORTANT 1, the same paragraph should carry the un-dispatched-`unsubmitted` gap alongside the existing timeout-as-rejection note — the section already does this for other named-rather-than-hidden gaps.

---

## Step 4 — Final Assessment

### Summary
This closes a real hole: `fulfillment.work.dispatch` had a handler, a lane, a handshake and an auto-accepting executor, and zero producers, so Wave 3a's exit criterion was structurally unreachable. The implementation is careful in the places that matter — the `assignmentAttempt` claim is untouched and unreachable from the producer, the scope is the work's own holder, the dedupe key sits in its own namespace, the leaf's no-injection invariant is respected via a neutral intent, and the never-throw contract is argued from the concrete double-fulfilment failure it prevents rather than asserted. The one thing that deserves an issue before merge is that a swallowed enqueue is *permanently* lost, with no reconcile pass anywhere.

### Merge Readiness
🔄 **Approve with changes** — merge once IMPORTANT 1 has a filed follow-up issue and IMPORTANT 2's one-sentence doc addition lands. Neither requires a design change.

### Priority fixes
1. File a follow-up issue for an `unsubmitted`-work reconcile sweep, and name that gap in both the PR's "Limitations" section and `docs/architecture-overview.md` § 26 beside the ADR-054 timeout sweep.
2. Extend the recorded #2395 constraint to state that `expectedAssignmentAttempt: null` disables `claimOrResume`'s attempt-scoping guard, so the key and the payload value must change together.
3. (Optional) Fold `findUndispatchableWorkIds` into the derivation's return so reported-and-skipped cannot drift from enqueued.

Issue #1689, opened 2026-07-16
[FEATURE] Consume master.deletion event → zero/deactivate live marketplace offers (oversell protection)


piotrswierzy:
## Context

Follow-up from the #1676 review of #1599 (master-side deletion propagation).

#1599 emits `master.variant.stale` / `master.product.stale` on the `events.master.deletion` stream when a product/variant is deleted at the master. Per #1599's scope, the event is the **seam** and consumers are explicitly deferred ("this issue only guarantees the event exists").

Today the marketplace side **fails open**: from master deletion until an operator notices, a mapped Allegro/Erli offer stays live at its last-known quantity and buyers can purchase a product that no longer exists (oversell). The new-offer path already fails safe (stale-filtered `findAvailabilityByVariantIds`), but **live** offers are unprotected.

## Proposed

- Add a consumer of the `events.master.deletion` stream that, for a stale-marked variant/product, enqueues an **offer quantity → 0 / deactivation** for mapped offers (via `OfferStockRestorer` / `updateOfferQuantity`) so the offer stops selling on the next sync.
- Fold in the related review hardening: bulk offer-creation fan-out should filter `isStale` siblings (review #7); consistent event-type derivation + a shared correlation id + optional stream `MAXLEN` (review 🟢 suggestions).
- FE: surface the deletion lifecycle honestly — a distinct **Source deleted** order state (not "Awaiting mapping"), and distinguish the `business_failure` deletion outcome from other business failures in the jobs list. A concrete before/after UX proposal was attached to PR #1676.

This is the buyer-protection half that closes review item #4.

Refs #1599, PR #1676.

jakubretajczykBD:
🤖 claimed for work by branch `1689-master-deletion-offer-pause` at 2026-07-28T08:12:29Z

jakubretajczykBD:
Splitting this into 2 PRs to keep review scope manageable:

- **PR 1 (buyer-protection core)**: #1893 — event-contract hardening, the event-driven trigger + core `StaleOfferPauseService`, the hourly reconcile sweep, the inventory-propagate race closure, and the bulk-wizard stale-variant filter (review #7 on #1676).
- **PR 2 (operator-facing honesty, follow-up)**: `source_deleted` order-record status + reason, `outcomeReason` on sync jobs, the two related migrations, and the FE (order health badge + jobs-list label). Not opened yet.

jakubretajczykBD:
[...]

PR #2889, opened 2026-09-04
docs(amazon-spike): live-verification pass for #2881 — Amazon SP-API (FR/DE/PL)

jakubretajczykBD:
https://claude.ai/code/artifact/e6ed8e9e-31e5-4baf-b8c5-39bec7654fd1

## Summary

A partial live-verification pass against the Amazon SP-API **static sandbox** (`Sandbox`-status app,
EU/NA hosts), layered on top of #2881's own day-0 desk research. Not exhaustive — several stories remain
at the original issue's desk-research status; see "What's still open" below and the spike's own
"Open risks" section.

- `docs/plans/analysis/SPIKE-2881-amazon-sp-api.md` — ~48 numbered evidence items across Groups
  C/T/O/P/S/F/D/R/X, each marked as ✅ verified live, 🎯 resolved via direct model inspection,
  ⚠️ reframed, or 🔴 blocked.
- `docs/specs/product-spec-2881-amazon-marketplace-integration.md` — intentionally left as a **DRAFT
  SCAFFOLD**, not a finished spec: sections are marked RESOLVED / CARRIED FROM ISSUE / BLOCKED rather
  than populated with placeholder decisions, since several open spike questions (connection grain,
  category UX, GTIN handling) would need to resolve first.
- Amazon sandbox integration is API-only; there is no Web UI.

## Highlights

- Several stories resolved with primary-source confidence: T1 (no category tree exists anywhere in
  current/deprecated SP-API), F5 (Poland absent from FBM Ship+, but reachable via the second
  purchasing API — Amazon Shipping v2), O10/D4 (order lines carry no tax rate, amounts+registration-ids
  only), D8/R4–R7 (no refund/return-write capability anywhere in SP-API).
- One self-corrected methodology error, documented in place: `getCatalogItem` was initially and
  wrongly marked unsupported in static sandbox (checked the wrong JSON path for the
  `x-amzn-api-sandbox` extension); corrected with a live transcript.
- A materially significant reframe: Amazon carries a named `ItemTaxCollection.model:
  "MARKETPLACE_FACILITATOR"` field ("Tax is withheld and remitted to the taxing authority by Amazon
  on behalf of the seller"), which likely explains P9 (no seller-settable tax rate field) and D5
  (Amazon natively separates shipping tax from item tax) rather than either being a genuine gap.
- C13 (Notifications) — the subscription contract (subscribe/read/enumerate) verified live and works;
  the earlier `invalid_scope` error on the grantless token is root-caused to a registration-**tier**
[...]

PR #2976, opened 2026-09-07
docs(integrations): eBay Sell APIs spike — live-probed findings (#2880)

### LINKED ISSUE #2880: [SPIKE] Integration research — eBay (GB/DE/FR/PL Sell APIs): capabilities, flows, verdict
> ## ⚠️ The research below is day-0 desk research and may be wrong
>
> Everything in this issue was produced in a single session on 2026-09-04 **without making a single live API call**. It may be incomplete, out of date, misread, or simply wrong.
>
> **eBay is the worst-sourced of the four.** `developer.ebay.com` returns **HTTP 403 to every automated fetch** — curl, browser user-agent and reader proxies alike — so every eBay doc below was read through `web.archive.org` snapshots dated **March–June 2026**, plus live `ebay.pl` seller pages. The canonical URLs work fine in a real browser. Snapshots can be stale, and at least one cited KB article is from 2022.
>
> **Verify every claim you rely on before building on it.** Each story's "desk claim" is a hypothesis to confirm or overturn — that is the work.
>
> Found something wrong? **Edit this issue** so the next reader inherits the correction.

Part of #2878.

## Problem / Context

eBay is a marketplace-only destination whose closest shipped analogue is Allegro (`libs/integrations/allegro`) — an external catalogue you link against rather than own, per-category required attributes, and an offer lifecycle with its own publish step.

Target markets: **GB, DE, FR, PL**. The enum values are `EBAY_GB` (**not** `EBAY_UK`), `EBAY_DE`, `EBAY_FR`, `EBAY_PL`.

**`EBAY_PL` is a real, listable marketplace** — it has its own default category tree, is in the Inventory API's marketplace enum, and eBay.pl publishes a domestic PLN fee schedule with worked examples in złoty. The research nearly concluded the opposite, because eBay's *cross-border* fees page lists AU/AT/BE/FR/ES/NL/IE/CA/DE/CH/US/GB/IT and omits Poland — that omission is by design (it is the foreign-selling page), not evidence.

Two things make eBay valuable beyond "another channel":

1. **`withdrawOffer` is a real end-listing that preserves the offer for relisting** — the first shipped destination with a genuine `OfferDeactivator`, the capability #1689 deferred because no adapter had one.
2. **The `inventory_item → offer → publishOffer` staging model** maps unusually well onto OL's existing `OfferCreationRecord` two-phase shape: only `sku`, `marketplaceId` and `format` are required at create; price, quantity, category, location and policies become required at publish.

Against that, three costs that are not Allegro-shaped:

- **RFC 9421 digital request signing is mandatory for EU/UK-domiciled sellers** — all four target markets — on refunds and returns writes.
- **12 mandatory business policies** (3 types × 4 markets) plus a mandatory inventory location before a single offer can publish.
- **There is no REST notification topic for new orders.** Order ingestion is poll-only, or legacy SOAP.

## Proposed Solution

Produce two documents:

- `docs/plans/analysis/SPIKE-<this issue>-ebay-sell-apis.md` — following `SPIKE-2289-allegro-returns-feed.md`: **Verdict** → numbered **Evidence** table → **API surface** → **Open risks** → **Recommendation**.
- `docs/specs/product-spec-<this issue>-ebay-marketplace-integration.md` — following `product-spec-978-erli-marketplace-integration.md`.

Proposed `platformType` `ebay` · `adapterKey` `ebay.sell.v1`.

## Classification

**Type**: Integration (research)
**Layer**: n/a — documentation only
**File(s)**: `docs/plans/analysis/SPIKE-*.md`, `docs/specs/product-spec-*.md`

## Prerequisites — this one has a real blocker, start it first

- [ ] An eBay developer account and a **sandbox keyset**
- [ ] 🔴 **A production keyset — and it ships DISABLED.** Per eBay: *"All new third-party developers must subscribe to or opt out of eBay marketplace account deletion/closure notifications **before they make their first API call**."* You must stand up an HTTPS endpoint (no localhost, no internal IP) that answers `GET ?challenge_code=…` with `200` and JSON `{"challengeResponse": sha256(challengeCode + verificationToken + endpoint)}` — in exactly that concatenation order, with a 32–80 char verification token. eBay warns that hand-built response strings often get a BOM prepended and fail the challenge; use a JSON serialiser. **This blocks every other production probe.**
- [ ] A seller account. EU business sellers and sole proprietorships must be **fully verified before listing** under the Digital Services Act — treat this as a hard precondition for all four markets
- [ ] For `tax.vatPercentage`: a business seller with a **VAT ID registered with eBay**
- [ ] For signature testing: a sandbox user whose **domicile is set to an EU country or GB** — otherwise signatures are neither required nor verified and you ship an untested code path

---

## Story checklist

Each box is `DONE` only with a live-call transcript + the endpoint + one line on *why that endpoint*, **or** an explicit NOT SUPPORTED naming the OL feature that degrades.

### Group C — Connect

- [ ] **C1** Register and obtain credentials — 🔒 *see the prerequisite blocker above*
- [ ] **C2** OAuth authorize + code exchange (`OAuthCompletionPort`) — ✅ *authorization-code grant; only Taxonomy reads accept an application token*
- [ ] **C3** Token refresh — ⚠️ *access 7200 s (2 h); refresh 47 304 000 s (~547.5 days). But eBay states refresh tokens *"can be unexpectedly revoked"* on username change, consent withdrawal, or eBay's own decision. Also a **daily cap on token minting** — cache, never mint per call*
- [ ] **C4** Connection test probe — ✅ *`getDefaultCategoryTreeId` is the cheap read*
- [ ] **C5** Config + credentials shape validators — ✅
- [ ] **C6** Auth failure → `needs_reauth` — ✅ *plus an `authorization revocation` notification topic — a clean push trigger, better than Allegro*
- [ ] **C7** Retry classification + deferral — ✅
- [ ] **C8** Per-connection rate limiter — ✅ *`GET /developer/analytics/v1_beta/rate_limit/` returns quota, calls made, remaining, window and time-to-reset per resource, **including legacy Trading**. Wire it to the health panel — this is better than Allegro offers*
- [ ] **C9** Discover which marketplaces this credential covers — **?** *decides AC4*
- [ ] **C10** Webhook provisioning — ⚠️ *`createDestination` + `createSubscription` exist, but see O4*
- [ ] **C11** Webhook decode + verify — ✅ *the Notification API has payload signature verification*
- [ ] **C12** Translate to `CanonicalInboundEvent` — ⚠️
- [ ] 🔴 **C14 — RFC 9421 digital request signing + key management.** *Mandatory for **EU/UK-domiciled sellers** on: all Finances API methods, Fulfillment `issueRefund`, Trading `GetAccount`, and Post-Order **Issue Inquiry Refund / Issue Case Refund / Issue Return Refund / Process Return Request / Create Cancellation Request / Approve Cancellation Request**. Four headers per call (`x-ebay-signature-key` JWE, `Content-Digest`, `Signature`, `Signature-Input`), keypairs via the Key Management API, **eBay does not store your private key**. Error 215114 on clock skew. eBay publishes open-source SDKs (Java/PHP/Python/Node) and a Docker verifier. **This is an HTTP-client-layer capability, not a header — size it as its own issue (epic AC5)**
- [ ] 🔴 **C15 — Marketplace account-deletion callback endpoint** — see prerequisites

### Group T — Taxonomy & attributes

- [ ] **T1** Browse category tree (`CategoryBrowser`) — ✅ *`getDefaultCategoryTreeId(marketplace_id)` → `category_tree/{id}`. Application token, so no user consent for taxonomy sync*
- [ ] **T2** Category path (`CategoryPathReader`) — ✅
- [ ] **T3** Leaf-only listing — ✅ *`getItemAspectsForCategory` **errors on a non-leaf**, same as Allegro*
- [ ] **T4** Required category parameters (`CategoryParametersReader`) — ⚠️ see T5
- [ ] 🔴 **T5 — the `aspectUsage` trap.** *eBay's own words: `aspectUsage` *"is always returned, even for hard-mandated/required aspects (where `aspectRequired: true`). The value returned for required aspects will be `RECOMMENDED`, but they are actually required and a seller will be blocked from listing."* **Read `aspectConstraint.aspectRequired`. Never branch on `aspectUsage`.** This is the #1962 / #2243 failure shape one platform over: the mirror reads all-green and the platform rejects every publish in the affected category*
- [ ] **T6** Parameter restrictions → `checkParameterRestrictions` — ✅ *`itemToAspectCardinality` (SINGLE/MULTI), `aspectAdvancedDataType: NUMERIC_RANGE`, enumerated `aspectValues`, `aspectEnabledForVariations`. Also **`expectedRequiredByDate`** — an aspect *scheduled* to become required. Allegro has no equivalent; a genuine early-warning signal worth surfacing*
- [ ] **T7** Catalogue product card (`CatalogProductReader`) — ⚠️ 🔴 *ePID, the direct Allegro-product-card analogue. **But `includeCatalogProductDetails` defaults to `true` if omitted**, so eBay will silently overwrite your title, description, item specifics and images with catalogue data. Direct conflict with ADR-046's "the master catalogue is authoritative and OL narrows on the way out". Recommend setting it `false` explicitly, or making it an operator-visible per-connection control. Note the Catalog API also carries "additional user restrictions apply"*
- [ ] **T8** Taxonomy identity per marketplace (`TaxonomyIdentityProvider`) — ✅ *ADR-037 already requires `'ebay:EBAY_GB'` form. **Some EU sites share a tree id — cache by returned `categoryTreeId`, not by marketplace***
- [ ] 🔴 **T12 — Taxonomy quota.** *5 000 calls/day **per App ID, not per seller**. Four trees walked naively exhausts it in one run, and on a multi-tenant deployment every merchant shares it. `DestinationCategory`'s paged resumable `expandedAt` frontier (#1979/#2061) is mandatory here, not optional. Remedy for more: the Application Growth Check review*

### Group P — Publish out

- [ ] **P1** Create a listing — ✅ *three steps: `createOrReplaceInventoryItem` → `createOffer` → `publishOffer`. Multi-variation adds `createOrReplaceInventoryItemGroup` + `publishOfferByInventoryItemGroup`*
- [ ] **P2** New catalogue entry vs offer against existing — ✅
- [ ] **P3** Idempotency on create — ⚠️ *the seller's own **SKU is the inventory key** — there is no eBay-side inventory item id*
- [ ] **P4** Async create → status poll — **?**
- [ ] 🔴 **P5 — Connection-level precondition gate (`validateBatch`, #2240).** *`listingPolicies` (fulfillment + payment + return) **"will always be required before publishing an offer"**, sellers must `optInToProgram` for business policies first, and **every policy is per-marketplace** — so **12 policies for four markets**. An **inventory location** (`createInventoryLocation`, caller-chosen immutable `merchantLocationKey`, max 50 chars) is also mandatory. A missing one fails every child of a bulk batch while the wizard reads all-green. **Recommend reading policies (`getFulfillmentPolicies?marketplace_id=…`) and letting the operator pick — do not author them; they encode legal return terms.** Note `CategoryTypeEnum` splits policies into `ALL_EXCLUDING_MOTORS_VEHICLES` / `MOTORS_VEHICLES`, and return policies exist only for the former*
- [ ] **P6** Description format grammar (ADR-046) — **?**
- [ ] **P7** Image upload — ⚠️ *Trading `UploadSiteHostedPictures` **decommissions 2026-09-30** and the Inventory flow guide still points at it — that doc is already stale. Use the Media API*
- [ ] **P8** Multi-variant grouping (#1065) — ✅ *sibling offers must share category, description, policies, location and marketplace; only SKU / quantity / price may differ. **`publishOfferByInventoryItemGroup` silently omits variants with missing data but fails the whole call on invalid data** — always read the warnings array*
- [ ] **P9** Write the tax rate at publish (#2245 / ADR-063) — ⚠️ *`tax.vatPercentage` is a **`number`, scale 3** — it **cannot express `zw` / `np` / `oo`**. An exemption-coded line must be **refused**, exactly like the Allegro numeric-`rates[]` case ADR-063 already handles. `tax.applyTax` and the sales-tax table are US/CA only*
- [ ] **P10** GPSR — ✅ *`regulatory.{documents,manufacturer,productSafety,responsiblePersons}`, conditionally required EU/NI since 2024-12-13. Discover per category via Metadata `getRegulatoryPolicies`*
- [ ] 🔴 **P11 — Duplicate guard (#1837).** *Only one offer may exist per **`sku` + `marketplaceId` + `format`**, and SKU is account-wide. **Two OL connections pointing at one eBay account would silently overwrite each other** — this interacts directly with AC4*
- [ ] 🔴 **P12 — `Content-Language` is a REQUIRED header** on `createOffer` and `createOrReplaceInventoryItem` (`en-GB`, `de-DE`, `fr-FR`, `pl-PL`). **eBay does not translate** — a German offer needs German copy supplied by OL
- [ ] 🔴 **P13 — Replace, not patch.** *`bulkCreateOrReplaceInventoryItem` is a complete replacement: *"all fields that are currently defined … are required in that update action, regardless of whether their values changed"*, and eBay advises a `getInventoryItem` read before every write. **That is a read-modify-write on the hottest path and a lost-update hazard under concurrency** — design for it*

### Group S — Keep in sync

- [ ] 🔴 **S1 — quantity is a TWO-LEVEL write.** *eBay: *"when updating the quantity of an item, it must be updated at both the inventory item and offer level."* Live listing shows `min(inventoryItem.quantity, offer.availableQuantity)`. That is actually a clean fit for a per-connection `stockSafetyBuffer` (#1844/#2323) over a global master quantity — worth stating in the spec*
- [ ] 🔴 **S2 — `bulkUpdatePriceQuantity`: 1 SKU or 25?** *eBay's own pages contradict each other. The method intro says *"Only one SKU (one product) can be updated per call"*; the `requests` field description says *"one or more inventory items (up to 25)"*; the bulk-updates guide says 25. The payload schema is an **array** of `{sku, offers[]}`, which favours 25. **25× difference in request volume — this is probe #1***
- [ ] **S3** Async quantity ack reconcile — **?**
- [ ] **S4** Write price — ✅ *same call as S2; published offers only, `updateOffer` for unpublished. Per-item `statusCode` returned, so partial-failure tolerant*
- [ ] **S5** Update offer fields — ✅
- [ ] **S6** Read offer status — ✅
- [ ] **S7** Commercial snapshot (#2024) — ✅
- [ ] **S8** Pause an offer — ✅ **flagship** *`withdrawOffer` ends a single-variation listing; the offer *"goes into the unpublished state, and will require a `publishOffer` call to relist"*. `withdrawOfferByInventoryItemGroup` for multi-variation. Strictly better than quantity-0 for #1689's pause chain — **but confirm whether relisting resets listing age / watchers / search standing, or re-incurs an insertion fee.** That decides whether it replaces quantity-0 or sits beside it*
- [ ] **S9** Restore stock after cancellation — **?**
- [ ] **S10** Enumerate offers → reconcile mappings — ✅
- [ ] **S11** Auto-match by SKU/EAN — ✅
- [ ] 🔴 **S13 — 250 listing revisions per calendar day, per listing.** *Past it the seller is blocked until the next day. A webhook-per-stock-change propagation design hits this ceiling*

### Group O — Orders in

- [ ] **O1** Order feed (`listOrderFeed`) — ⚠️ *`GET /sell/fulfillment/v1/order`, `filter=lastmodifieddate:[…]`*
- [ ] **O2** Hydrate one order — ✅ *🔴 **`getOrders` returns `cancelStatus.cancelRequests` ALWAYS EMPTY — use `getOrder`.** A cancellation-aware ingestion cannot rely on the list read*
- [ ] 🔴 **O3 — no cursor; offset paging over a live set.** *Exactly the ADR-048 decision 3 step-over hazard — overlap the watermark and rely on idempotent writes. Three sharp edges: **if both `creationdate` and `lastmodifieddate` are supplied only `creationdate` is used** (so a watermark poll sending both never sees modifications); **`limit > 200` fails rather than clamping** (default 50); filter special characters must be percent-encoded (`[`=`%5B` etc.). Also: `orderIds` **overrides and ignores every other parameter**. Default window 90 days; hard ceiling 2 years*
- [ ] 🔴 **O4 — there is no REST notification topic for a new order.** *The Notification API's topics are: authorization revocation, buyer question, buyer requested purchase quote, feedback ×3, item availability, item marked shipped (sent to a **buyer**), item price revision, listing preview task status, marketplace account deletion, new message, order confirmation, PLA budget, seller service metric, seller standards. Community threads report `ORDER_CHECKOUT_COMPLETE` 404ing. **Options: (1) poll `getOrders` — recommended for v1; (2) legacy SOAP Platform Notifications via Trading `SetNotificationPreferences`, and `ItemMarkedPaid` decommissions 2026-06-22; (3) both, the shipped PrestaShop #904 pattern.** Decide and record why*
- [ ] **O5** Buyer identity + email normalization — ✅
- [ ] **O6** Shipping / billing address — ✅
- [ ] **O7** Buyer tax id (#2599) — **?**
- [ ] **O8** Resolve a line (`lineItemId` + SKU) — ✅
- [ ] 🔴 **O9 — `lineItems.total` returns a DIFFERENT value** with vs without `fieldGroups=TAX_BREAKDOWN`. *Any revenue figure must pin it **and record that it did** — this is the silent-figure-movement class ADR-040 exists to prevent*
- [ ] **O10** Per-line tax **rate** — ❌ *amounts only. A grep of the full `getOrders` field list for `rate` / `percent` returns nothing. **OL feature degraded: Net Sales** — ADR-063 forbids deriving it from `taxAmount / netAmount`*
- [ ] **O11** Status vocabulary → `order_state_mappings` — ✅
- [ ] **O12** Observe a cancellation — ⚠️ *`getOrder` only, see O2*
- [ ] **O13** Payment status — ✅
- [ ] **O16** Ingest within the rate limit — ✅ *Fulfillment order resource: 100 000 calls/day, comfortable*

### Group F — Fulfil

- [ ] **F1** Read fulfillment status — ✅
- [ ] **F2** Write tracking (`createShippingFulfillment`) — ✅ *`POST /sell/fulfillment/v1/order/{orderId}/shipping_fulfillment`, **one call per package**. 🔴 **`shippedDate` DEFAULTS TO NOW if omitted** — supply the carrier's real instant, per OL's #2336/#2367 rule that a channel-reported instant is the channel's. Constraint: all components of a line item must be in the same package. Bonus: the same call doubles as proof of delivery for payment disputes*
- [ ] **F3** Order status writeback — **?**
- [ ] **F4** Late-waybill relay (#1947) — ✅
- [ ] **F5** Buy a shipping label — 🔒 *Logistics API is **"a (Limited Release) API available only to select developers approved by business units"**, and **labels do not work in sandbox at all**. EU coverage unknown, language suggests US-first. **Do not plan on it** — for PL/DE the path is OL's existing InPost/DPD adapters with F2 relaying the waybill back, the Allegro #1947 shape*
- [ ] **F7** Source options — ✅
- [ ] Note `eBaymag` for the spec: eBay's own free tool that copies listings across sites and syncs quantity. **A competitor to what OL would do, not an API.** If a merchant runs it, two systems write the same listings. Worth asking merchants before assuming OL is the only writer

### Group D — Money & documents

- [ ] **D1** Issue an invoice for an eBay order — ⚠️
- [ ] **D2** Buyer data sufficient to invoice — ✅
- [ ] 🔴 **D3 — persist eBay's own IOSS/OSS number.** *`lineItems.ebayCollectAndRemitTaxes[].ebayReference` carries `name` (`ABN` / **`IOSS`** / **`OSS`**) and `value` (the VAT identifier), with `collectionMethod` always `NET`. **That is the marketplace's own VAT identifier for a facilitated sale and it belongs on an invoice — nothing in OL's order contract models it today.** Covers UK, EU, Kazakhstan and Belarus*
- [ ] **D4** Per-line tax rate for Net Sales — ❌ *see O10. **eBay is the sharpest case of the asymmetry: it accepts a rate at publish and never returns one.** Epic #2245 assumed marketplaces either report a rate or can be made to — this is a third case the tax-coverage detector does not model*
- [ ] **D5** Shipping tax split (#2248/#2252) — ⚠️ *`lineItems.taxes` is per line and shipping sits outside it, so a mixed-rate basket needs `splitShippingAcrossRates`*
- [ ] **D6** Invoice correction from a return (#2374) — ⚠️ *gated on C14*
- [ ] **D8** Execute a refund (`RefundExecutor`) — 🔴 *gated on C14*
- [ ] **D9** FX stamping (ADR-040) — ⚠️ *`Amount.currency` defaults to *"the default currency of the eBay marketplace that hosts the listing"*, so GBP/EUR/EUR/**PLN**. **The PLN mapping is strongly indicated by the domestic PLN fee schedule but UNCONFIRMED from developer docs — confirm from a real `getOrders` response***

### Group R — Returns

- [ ] **R1** Return feed — ✅ *`GET /post-order/v2/return/search`*
- [ ] **R2** Hydrate one return — ✅ *`GET /post-order/v2/return/{returnId}`*
- [ ] **R3** Terminal status vocabulary (#2330) — **?**
- [ ] **R4/R5** Approve / decline — ✅ 🔴 *`POST /post-order/v2/return/{returnId}/decide` — **needs C14***
- [ ] **R6** Mark received (custody, #2367) — ✅ *`mark_as_received`*
- [ ] **R7** Refund against a return — ✅ 🔴 *`issue_refund` — **needs C14***
- [ ] **R8** Orphan attribution (#2332) — ✅
- [ ] 🔴 **R9 — a live EU change is already announced.** *eBay's own banner: *"A new **Right of Withdrawal (ROW)** return reason will be added to the `getReturnDetails` API for eligible EU returns. Update your integration so it can safely handle new/unknown return reason enum values without errors."* OL's `readReturnReason` coercion posture is already the right shape — **confirm the exact enum value and date***
- [ ] 🔴 **R10 — roadmap risk.** *Post-Order v2 is being **partially decommissioned through 2026** (waves on 2026-01-20, 02-02, 03-02, 03-16) **with no named successor**, while eBay simultaneously tells Return Management API users to migrate *to* Post-Order. The surviving set (search, get, decide, mark_as_received, issue_refund, preferences, tracking, files, message, escalate, add_shipping_label) maps almost one-for-one onto OL's returns aggregate — but record the risk. **Post-Order is also 5 000 calls/day per resource vs Fulfillment's 100 000, so the returns poller must be paced completely differently***

### Group X — Operate

- [ ] 🔴 **X1 — sandbox is usable for shape, not for reality.** *Per KB 684: **shipping label flows do not work**; **many ePIDs mapped to ISBN/UPC in production are absent** (so catalogue matching cannot be tested — the feature you would most want to rehearse); VAT-exempt registration does not work; fees are not guaranteed to match; no order emails; INR disputes cannot be submitted; **trust-and-safety checks are absent, so a listing that publishes cleanly in sandbox may be blocked in production**; Basic Store only. **And per KB 95 (2022, Trading-era) the supported sandbox sites are AU/AT/BE-FR/BE-NL/CA/FR/DE/IT/NL/ES/CH/UK/US — Poland is NOT listed.** Whether the REST sandbox nonetheless accepts `marketplaceId: EBAY_PL` is **UNCONFIRMED and is probe #2** — if it rejects, **Poland is production-only testing**, which is uncomfortable for a fiscal-adjacent integration*
- [ ] **X2** Rate-limit observability — ✅ *`getRateLimits`, see C8*
- [ ] **X3** Compliance gate — 🔒 *see prerequisites*
- [ ] **X5** Deprecation clock — 🔴 *Finding and Shopping APIs **already decommissioned 2025-02-04**; Merchandising deprecated 2024-07-29; Business Policies Management deprecated 2021-03-23 (use Account API); `ItemMarkedPaid` 2026-06-22; Trading image upload 2026-09-30; shipping discount profiles 2027-01-19; Post-Order waves through 2026*
- [ ] 🔴 **X6 — quotas are per App ID, shared across every tenant.** *Inventory 2 000 000/day is ample; **Taxonomy 5 000/day and Trading 5 000/day are the scarce ones**. Remedy is the Application Growth Check*
- [ ] Record for the spec which functions remain **Trading-API-only**: new-order push notifications, buyer/seller messaging, `CrossBorderTrade`. Best Offer may now be covered by the Leads API — parity UNCONFIRMED. **A pure-REST eBay adapter is achievable for publish + order ingest + fulfilment; it is not if you need order push, messaging or Best Offer**

---

## Multi-marketplace specifics to settle (feeds epic AC4)

- [ ] Confirm one account + one token covers all four sites
- [ ] Confirm the model is `1 inventory_item (SKU) → N offers (one per marketplaceId) → N listings`, with `shipToLocationAvailability.quantity` global and each offer's `availableQuantity` a per-market allocation
- [ ] Confirm what is per-marketplace vs shared: **per-marketplace** = policies, category tree, currency, `Content-Language`, offers; **shared** = inventory location, SKU/inventory item
- [ ] Confirm there is **no automatic cross-listing** worth relying on: `CrossBorderTrade` is Trading-only and US/UK/CA/IE only; international shipping settings do not create a second listing; eBaymag is a separate tool, not an API
- [ ] **AC8 — is `EBAY_PL` commercially worth a connection?** Third-party traffic estimates for ebay.pl conflict by ~10× (one article ~4 M visits/month, an Ahrefs profile ~416 K) and **neither is eBay's own figure — do not quote either in a business case.** eBay's Polish marketing is overwhelmingly export-oriented (*"80% sprzedaży to transakcje w UE"*, *"sprzedają średnio na 28 rynkach"*). **If EBAY_PL is thin, the real product is "a Polish merchant selling into DE/GB/FR"** — three markets, one seller domicile, a materially simpler product that changes what the connection UI asks for first

## Acceptance Criteria

- [ ] Every story above is `DONE` per the rule at the top of the checklist
- [ ] AC0 — the full capability table with the vendor operation named per row
- [ ] 🔴 **AC2 — sandbox access OBTAINED per market, not just confirmed to exist.** Sandbox and production keysets issued, the **account-deletion endpoint live and the production keyset actually activated**, and a **real authenticated call per market with the transcript attached**. Answer **GB, DE, FR and PL separately** — KB 95 does not list Poland, so **whether the REST sandbox accepts `marketplaceId: EBAY_PL` must be probed, not assumed**. If it rejects, say so plainly and state what production-only testing then requires and who supplies the seller account. Also state what the sandbox cannot verify at all: shipping labels, ePID/GTIN catalogue matching, VAT-exempt registration, fees, INR disputes, and trust-and-safety listing checks. If a signature-scoped call is in scope, the sandbox user's **domicile must be set to an EU country or GB** or that code path is untested
- [ ] AC3 — adopt / don't-adopt verdict + risk register, unresolvable risks marked CONCEDED
- [ ] AC4 — connection grain: one connection with a marketplace axis, or four sharing credentials
- [ ] AC5 — **C14 (RFC 9421 signing) sized as its own issue or explicitly declined**
- [ ] AC6 — the "what breaks in OpenLinker" list for eBay
- [ ] AC8 — the EBAY_PL commercial question answered
- [ ] AC9 — effort estimate
- [ ] Both documents committed and linked from #2878
- [ ] A recommendation on **S8**: does `withdrawOffer` promote `OfferDeactivator` to a real capability (#1689), or stay an eBay special case beside quantity-0?
- [ ] Any claim in this issue found to be wrong is corrected in place

jakubretajczykBD:
https://claude.ai/code/artifact/d877edf4-6221-425d-a2b2-7e488debfa20

Adds both documents #2880 asks for. **Docs only — two new markdown files, no code, no schema, no behaviour change.**

Refs #2880, part of epic #2878.

## What this is

- `docs/plans/analysis/SPIKE-2880-ebay-sell-apis.md`, in the format the issue specifies (Verdict → Evidence → API surface → Open risks → Recommendation). Every row in the 34-row evidence table is a real HTTP request/response captured against the eBay sandbox — zero production API calls (one production developer-portal configuration attempt is recorded in Open Risk 5).
- `docs/specs/product-spec-2880-ebay-marketplace-integration.md`, in the `product-spec-978-erli` format. Written at **Phase C (solution exploration)**, not a Gate-D commit — no implementation issues are spawned and no market/cohort numbers are invented. Recommends **Shape B: ship the publish-half adapter now, gate order ingestion on a cheap local retest** rather than either committing to an unverified order-ingestion path or blocking the whole adapter on it.

## Headline result

**The publish half works end to end and is a green light.** `inventory_item` → `offer` → `publish`, multi-variant grouping via `inventory_item_group`, `withdrawOffer` as a pause primitive, bulk price/quantity updates, GPSR discovery, and writing `tax.vatPercentage` were all exercised live against real listings.

**Webhook provisioning also works end to end** (`config` → `destination` → `subscription`), with eBay delivering real challenge codes to a live endpoint and answering them successfully.

**Order ingestion is still unverified**, and both documents are explicit that this is most likely an account-setup problem — a sandbox seller registered on `EBAY_DE` while every listing published to `EBAY_GB`, the issue's own first prerequisite, never satisfied — rather than a demonstrated eBay limitation. The recommended next step is a cheap local retest, not a production keyset.

There are also numerous errors and inconsistencies on eBay's Web UI/pages. These appear to be issues on the eBay-side interface rather than failures in the integration itself, and should be treated separately from the API/integration results above.

## What it corrects in #2880

- **`EBAY_PL` works in the sandbox** — KB 95 implied otherwise; the Taxonomy API answers for PL with its own tree.
- **None of the four target markets share a category tree** (GB=3, DE=77, FR=71, PL=212).
- **The Notification API topic list is materially different** from the one the issue names, and this confirms **O4**: there is genuinely no new-order topic, so order ingestion has to be a poll.
- **`S2` resolved** in favour of "up to 25 SKUs per `bulkUpdatePriceQuantity` call".
- **`S8` answered**: `withdrawOffer` preserves the offer but relisting mints a **new** `listingId` — not a true pause/resume. eBay's own `withdrawOffer` reference additionally states the call ends the buyer-facing listing; this session did not itself observe that transition, so the product-spec records promoting a real `OfferDeactivator` (#1689) as an **ADR-pending decision** (R7), not a settled one — it would reverse `implementation-plan-1689`'s recorded rejection of that capability.
- **`S10` corrected**: no endpoint lists every offer on an account — reconciliation is a two-step sweep.
- **`AC4` half-corrected**: `1 SKU → N offers` holds, but the inventory item is single-locale and a write for another market silently overwrites the first market's content.

## Findings not in #2880 at all

13, each with the reproducing call. The three worth reading if you read nothing else:

- **A partial `PUT /inventory_item` silently destroys omitted fields, including `ean`** — which `publishOffer` requires. The live listing keeps working, so the damage is latent and surfaces at the next publish. Read-modify-write is mandatory, and the product-spec (§8, R6) makes this a hard implementation constraint rather than general awareness.
- **A write with a mismatched `Content-Language` returns `204` and is half-applied.**
- **App-level notification config is an undocumented precondition for webhook provisioning.**

## Honesty about coverage, and about a wrong verdict

The SPIKE doc states plainly which checklist groups are not covered: all of O/F/R/D (including `S3`/`S9`, which depend on an order that never materialised), `C14` (RFC 9421 — needs an EU/GB-domicile buyer), `T7`/`S11` (Catalog API returns 403), `C11`/`C12` (no payload obtainable), and `X5` (deprecation clock — desk research, not re-verified here). It carries a methodological note recording a verdict an earlier revision got wrong — repeated identical failures are evidence about the caller until a probe exists that would have behaved differently.

The product-spec is equally explicit about what it is **not**: it does not spawn implementation issues, does not size effort, and does not answer epic #2878's open `AC8` (whether `EBAY_PL` is commercially worth a connection, or whether the real product is a Polish seller exporting into DE/GB/FR) — that stays open pending real research, not a guess dressed up as one.

## Why `Refs #2880` rather than `Closes`

This branch is based on `main`, which is not the branch #2880's parent epic (#2878) will land on before the whole feature ships — `Closes` only auto-fires on merge to the repo's default branch, and hand-closing an issue afterward is against `CLAUDE.md § Working Process`. `Closes #2880` belongs on whichever PR actually merges to `main` carrying the epic's full intent; this one carries `Refs` instead.

## Verification

- No code, schema, migration or behaviour change; two markdown files under `docs/`.
- No `check:invariants` gate scans either directory; the one gate that reads docs (`check-repo-urls`) forbids three stale repo-URL substrings, none of which appear.
- Both files scanned for credentials, account names, sandbox object ids, and tunnel URLs — none present. Working notes containing the sandbox keyset id, a verification token and a tunnel URL were deliberately kept out of the repository.


piotrswierzy (commented):

**Coverage note:** I read both real content files in full — `SPIKE-2880-ebay-sell-apis.md` (268 lines, all 34 evidence rows) and `product-spec-2880-ebay-marketplace-integration.md` (250 lines). I did **not** read the 352-file GitHub diff, because it isn't this PR's content: `git diff origin/main...origin/2880-ebay-sell-apis-spike` is **exactly 2 files, +518, 0 deletions**.

The substance is genuinely good — see the closing section. Findings are docs-accuracy and PR hygiene; no BLOCKING issues, no security exposure, nothing generated or credential-bearing committed (I swept both files for keysets, bearer tokens, verification tokens, tunnel hosts, emails and `client_secret` — clean; the only two URLs are the two GitHub issue links).

### IMPORTANT — the PR is structurally unreviewable on GitHub

It renders as **352 files / +38,377** while the description says "two new markdown files", because the head was cut from a `main` 26 commits ahead of the epic base — so the pack bench, shipment lines, the fulfilment router and 10 other unrelated commits are attributed here. `gh pr diff 2976` doesn't merely mislead, it **refuses**: `HTTP 406: the diff exceeded the maximum number of files (300)`. A reviewer approving on the description alone approves 350 files they were structurally prevented from seeing. Rebase onto `2878-new-integrations-story-catalogue`. Not BLOCKING only because the content is verifiable locally and is clean — it would be if this targeted `main`.

### IMPORTANT — `Closes #2880` will silently not fire

GitHub auto-closes only on merge to the **default** branch; this targets `2878-new-integrations-story-catalogue`. And `CLAUDE.md § Working Process` forbids hand-closing afterwards. Move the keyword to #2883's body or retarget. Note the spike itself says `Refs #2880` (line 9) — the correct keyword for a stacked PR; the doc and the PR body disagree.

### IMPORTANT — the one inference presented as a live finding is the one carrying the most weight

Spike line 246; spec lines 24, 131–133.

The Recommendation states `withdrawOffer` "actually ends the listing (**buyer-facing `ACTIVE` → `ENDED`**)", and the spec repeats `(ACTIVE → ENDED)` as a key sub-decision. **No evidence row records that.** Evidence 7 — the only `withdrawOffer` observation — captures `status: UNPUBLISHED`, which is the *offer's* Sell-API status, not a buyer-facing listing state. The transition was reasoned about, not observed.

This matters because that claim is the **entire justification** for the doc's largest architectural recommendation: promoting `OfferDeactivator` across the port family. `implementation-plan-1689` (654, 803) rejected exactly that capability on the grounds that quantity-0 already stops the sale; the counter-argument here is "no, withdraw genuinely *ends* it" — the unverified half. That is the #2594/ADR-066 failure mode this repo has a hard-won rule about. Either downgrade to "eBay documents this as ending the listing; the buyer-facing state was not observed", or observe it and add a row.

### IMPORTANT — a market figure is attributed to the spike, and the spike doesn't contain it

Spec line 50 credits *"the spike's own reading of eBay's Polish marketing"* for `"80% sprzedaży to transakcje w UE"`. The spike never mentions eBay's Polish marketing — zero matches for `80%` or `sprzeda`. So the single commercial number in a document whose §2 opens *"Not independently researched for this pass"* is uncited **and** falsely attributed to the one artefact here that is evidence-backed. It also drives the "Polish merchant selling into DE/GB/FR" framing that §6 item 5 and R5 build on. Cite the real source or drop the figure.

### IMPORTANT — two broken cross-references and a findings list that won't render

- Spike line 26 cites `sellerRegistrationCompleted: false` as **"(Evidence 27)"**. Row 27 is F7 — three 404s guessing carrier endpoint names. The correct row is **31** (Open Risk 1 at line 187 cites it correctly). A reader chasing the Verdict's central claim lands somewhere unrelated.
- Row 22 says *"see New findings 8/9"*. **There are no items 8 or 9** — the list runs `1–6`, then `7a…7f`, then a second `7.`. The intended targets are 7e/7f. `7a`–`7f` are not valid ordered-list markers, so CommonMark won't render them as list items, and the trailing `7.` restarts the list — the section renders visibly broken. Renumber flat `1…13`. (The PR body's "**Nine** new findings" is also wrong — it's 13.)

### SUGGESTIONS

- Evidence row 33's verdict reads flatly *"ROOT CAUSE of the checkout failure."* The observation is solid; the attribution is inference — and **every other place hedges it correctly** ("most likely", "the likely cause", "unverified, not disproven"). Only this cell asserts it. Match the hedging.
- The header's absolute *"Zero production calls"* sits against Open Risk 5, which documents a production developer-portal action. Narrow to "zero production **API** calls; one production developer-portal configuration attempt, recorded in Open Risk 5."
- Promoting `withdrawOffer` to `OfferDeactivator` is a **plugin-contract change** — `engineering-standards.md § ADRs` says write one when a choice "affects multiple bounded contexts **or the plugin contract**", and the very spec this mirrors (`product-spec-978-erli`) spawned **#983** as a dedicated ADR issue for this class. §10 hands off rather than spawning issues, so add the ADR obligation to its next steps or Gate D will skip it.
- Two deviations from `docs/specs/README.md`'s prescriptive skeleton: §9/§10 swapped, and `Status` is free-text rather than one of the enum values. (`product-spec-961-dpd` also deviates, so there is precedent.)

### Done well

- **The methodological note (257–264) is the best thing here.** It records a verdict an earlier revision got *wrong*, names the mechanism (*"repeated identical failures are evidence about the caller until a probe exists that would have behaved differently"*), and shows the differentiating probes that corrected it. That's the ADR-066 discipline applied reflexively rather than cited. Open Risk 5's *"stated as a lead, not a fact"* is the same in miniature.
- **Non-coverage is enumerated positively** — every uncovered group with its blocker, rather than left as silence.
- **Two of #2880's own claims are actively corrected**, not confirmed by re-reading docs: S10 downgraded from a single bulk read to a two-step sweep, AC4 half-corrected.
- **No captured dumps or fixtures committed** — matching the `SPIKE-2289` / `SPIKE-2375` precedent exactly.
- The `withdrawOffer` idempotency asymmetry (Evidence 32) is exactly the retry-safety detail this codebase's job runner cares about, and isn't in #2880 at all.

---

norbert-kulus-blockydevs:
## Tech Lead Review — PR #2976 (docs-only: eBay Sell APIs spike + product-spec)

Reviewed as a documentation/research artifact (two new markdown files, no code/schema/behaviour change), following the same rigor applied to sibling PRs #2888 (Shopify) and #2889 (Amazon) in this epic.

### What holds up well

- **The headline "34-row evidence table" claim is accurate** — counted, and it is exactly 34 rows (#1-#34). This is the same class of claim that had an arithmetic error in the sibling #2888 review (an "80 of ~90" figure that didn't reconcile against its own per-group denominators); here it does reconcile.
- **No secrets/credentials/tunnel URLs found** in either file — scanned for API-key-shaped strings, bearer tokens, `Authorization:` headers, verification tokens, ngrok/cloudflare tunnel hosts. Clean, consistent with the PR description's own claim.
- Format matches the issue's own template request (`SPIKE-2289-allegro-returns-feed.md` shape for the spike, `product-spec-978-erli-marketplace-integration.md` shape for the spec) — verified both precedent files exist and the structure (Verdict → Evidence → API surface → Open risks → Recommendation; 1-10 numbered product-spec sections) matches.
- Cross-references to existing OL architecture (ADR-046 description-format grammar, ADR-063 tax-rate provenance, ADR-040 FX-stamping "silent figure movement" principle, #1689 `OfferDeactivator`, #1979/#2061 paged taxonomy frontier) are used correctly and not just name-dropped — e.g. the `includeCatalogProductDetails: true` default finding is correctly tied to ADR-046's "master catalogue is authoritative" principle, and the `lineItems.total` TAX_BREAKDOWN discrepancy is correctly tied to ADR-040's silent-figure-movement concern.
- The methodological self-correction (an earlier revision wrongly blamed eBay's sandbox webhook infra on 4 failures that were actually 3-caller-error-1-config-gap) is a good, honest inclusion and exactly the kind of thing that should be recorded rather than silently fixed.

### 🟡 IMPORTANT — the PR description's coverage claim doesn't match what the document actually says

The PR description states: *"The SPIKE doc states plainly which checklist groups are not covered: all of O/F/R/D, C14 (RFC 9421 — needs an EU/GB-domicile buyer), T7/S11 (Catalog API returns 403), C11/C12 (no payload obtainable), and X5."*

**X5 ("Deprecation clock" — a 🔴-flagged story in issue #2880) does not appear anywhere in either committed document.** Not in the Evidence table, not in the "Not exercised live in this session" paragraph, not in Open Risk item 6's "everything already flagged 🔴 and not reduced by this session" list. It's simply absent — neither confirmed, corrected, nor acknowledged as uncovered. Grepped both files to confirm.

Two more items from the issue checklist have the same problem, at slightly lower severity since they were `?`-marked rather than 🔴-flagged: **S3 (async quantity ack reconcile) and S9 (restore stock after cancellation)** — neither is mentioned in either document at all.

**Why this matters more than it would in an ordinary PR**: the document's own stated discipline (repeated in both the issue's rules — "Any claim in this issue found to be wrong is corrected in place" — and the SPIKE doc's own "coverage honesty" paragraph) is to be explicit about what wasn't tested rather than silently drop it. That's a real virtue the PR earns credit for elsewhere (e.g. explicitly listing C11/C12, T7/S11, C14 as untested). X5/S3/S9 falling through that same net, while the PR description asserts X5 specifically *was* accounted for, is the one place this PR doesn't live up to its own standard.

**Suggested fix**: add X5, S3, and S9 to the "not exercised" / Open Risks enumeration in the SPIKE doc (a one-line addition each — no new probing required, this is just an honesty/completeness fix), and correct the PR description's claim about X5 if it isn't updated.

### 🟢 Minor / non-blocking

- **S13 (250 listing revisions/day cap, 🔴-flagged in the issue)** is also absent from both documents. Given the recommended v1 shape (Shape B) explicitly plans stock/price propagation via webhook-per-stock-change (S1), this ceiling is directly relevant to the shape actually being recommended — not a remote concern. Worth a one-line addition to Risks (§8) in the product-spec, even though it's a `?`/design concern rather than a live-probed fact.
- Evidence row 27 (F7, shipping carrier/services endpoint) is marked "Inconclusive — left open" rather than folded into "not covered," which is fine, but it's the only mixed case; consider a consistent bucket (either "tested and inconclusive" or "not covered") for readability in a future revision.

### Verdict

**🔄 Approve with changes.** The technical content is well-sourced, live-probed, arithmetically sound, and secret-free — a real step up in rigor from the desk-research issue it corrects. The one thing to fix before merge (or as an immediate follow-up edit, since this is a fast-moving docs PR) is aligning the PR description's coverage claim with what the SPIKE document actually discloses about X5, and closing the S3/S9/S13 gaps in the same pass so a future reader trusts the "coverage honesty" section to be complete, matching the discipline the epic's `docs/plans/analysis/` convention is built around.

piotrswierzy (commented):
## `/pr-review` (re-review) — 🔄 Approve with changes (verdict unchanged)

**The structural blocker is genuinely gone**, which was the finding that mattered most and the one a reviewer could not work around. Verified three ways: `gh pr diff 2976 --name-only` now returns exactly the two markdown files with no 406; `gh pr view` reports `FILECOUNT 2`; and `git diff --stat origin/2878-…...origin/2880-ebay-sell-apis-spike` is **2 files, +518, 0 deletions**. The PR is reviewable as submitted.

The `80%` attribution is also fixed, and I verified the underlying claim rather than the commit message: spec line 50 now credits *"issue #2880's own reading"*, and `gh issue view 2880` does contain `80% sprzedaży to transakcje w UE`. And line 26 now reads `(Evidence 31)`, agreeing with Open Risk 1.

But the fix commit addressed **3 of 5 IMPORTANTs and none of the 4 SUGGESTIONs**, and one of the three fixes made a cross-reference worse.

### IMPORTANT — the `ACTIVE → ENDED` claim survived a review cycle unchanged

This is the one I would not let pass a second time. Spike line 246 still asserts *"it actually ends the listing (buyer-facing `ACTIVE` → `ENDED`)"*; spec line 133 still repeats `(ACTIVE → ENDED)`. **`ENDED` appears nowhere in the 34-row evidence table.** Evidence 7 — still the sole `withdrawOffer` observation — records `status: UNPUBLISHED`, the offer's Sell-API status. No row was added.

It remains the entire justification for promoting `OfferDeactivator` across the port family, against `implementation-plan-1689`'s (654, 803) named rejection. And the spike's own methodological note (257–264) states the rule it breaks — *"repeated identical failures are evidence about the caller until a probe exists that would have behaved differently"* — which is the same discipline applied in the opposite direction. Either observe the buyer-facing state and add row 35, or downgrade to *"eBay documents this as ending the listing; the buyer-facing state was not observed."*

### IMPORTANT — the numbering fix turned a dangling reference into a confidently wrong one

The commit renumbered the trailing list item from `7.` to `8.`, so the sequence is now `1–6`, `7a…7f`, `8`. But `8` is *"Not every category aspect may be used as a variation axis"*, while row 22's `"see New findings 8/9"` was pointing at the **webhook** items (`7e`/`7f`). Before the fix the reference was dangling; now it resolves confidently to the wrong finding, and `9` still doesn't exist. `7a`–`7f` are still not valid ordered-list markers, so the section still won't render as a list. The requested flat `1…13` renumber would have fixed both at once.

### IMPORTANT — `Closes #2880` was answered on the merits, not on the mechanics

The body grew a new section, *"Why `Closes #2880` despite unmet checklist items"*, arguing whether closing is warranted. That was not the finding. The finding was **mechanical**: GitHub auto-closes only on merge to the **default** branch, and this targets `2878-new-integrations-story-catalogue`, so the keyword cannot fire — after which `CLAUDE.md § Working Process` forbids closing by hand. The spike itself still says `Refs #2880` (line 9), so doc and body still disagree. The rebase did not change this, because it rebased onto the same non-default base.

With the head now rebased onto the epic base there is a cheaper remedy than before: move the keyword to the epic-integration PR (#2883) that merges to `main`.

### Still unaddressed

Body says "Nine" new findings (actual 13). Row 33 still reads flatly `**ROOT CAUSE of the checkout failure**` while every neighbouring statement hedges (*"most likely"*, *"the likely cause"*, *"unverified, not disproven"*). Header still says `Zero production calls.` while Open Risk 5 documents a production `developer.ebay.com → Alerts & Notifications → Save`. The spec contains **zero** occurrences of `ADR`, so the `OfferDeactivator` plugin-contract obligation is absent from §10's next steps (the `product-spec-978-erli` precedent spawned #983 for this class). §9/§10 are still swapped and `Status:` is still free-text against `docs/specs/README.md`'s five enum values.

### Security re-swept: clean

I re-checked both files for bearer tokens, `client_secret`, `v^1.1#` keyset strings, tunnel hosts and emails — nothing.

**Comparative note**: I'd treat #2976 as further from approvable than #2888, which absorbed nearly all of its feedback. And on the epic's cross-cutting lesson — #2976's spec has no stale item at all, but it faithfully reproduces its spike's *unverified* `ACTIVE → ENDED` claim, so the two documents now agree on something neither has evidence for. Regenerating a spec from its spike propagates the spike's confidence level, correct or not.

---

piotrswierzy (commented):
## Tech Lead review

Third pass. The head is **unchanged since my 2026-09-10 re-review** (`7906a1b638`), so all three IMPORTANTs from that round stand verbatim and I will not re-argue them at length. What is new is that the PR now carries a **third file** — a stale copy of the epic catalogue that lives on the base branch — which is both the cause of `mergeable: CONFLICTING` and a real documentation regression. The eBay research itself remains good work, live-probed and honest about its gaps; security is clean (I swept all three files for keysets, bearer tokens, `client_secret`, verification tokens and tunnel hosts — nothing).

### Findings

**[BLOCKING]** — `docs/plans/analysis/ANALYSIS-new-marketplace-integrations-story-catalogue.md`
> This file is **not this PR's content**. It already exists on the base branch `2878-new-integrations-story-catalogue` at 43,088 bytes; the copy here is 40,763 bytes and **older**. Merging it reverts 50 diff lines of the epic catalogue. Three of the reverted edits are corrections, not cosmetics:
> - **ADR-037 spelling.** Base reads *"It prescribes no literal spelling, and it carries a clause that cuts the other way… the eBay owner value should key on the `categoryTreeId` the Taxonomy API returns, not on a marketplace literal."* This PR re-asserts the superseded *"ADR-037 already ruled that Amazon and eBay must onboard as `'amazon:<marketplaceId>'` / `'ebay:EBAY_GB'`"*. That is a claim about an ADR that the base branch explicitly downgraded — and `docs/architecture-overview.md § Listings` (#1979/#2063) is the reason: `TaxonomyIdentityProvider` makes an owner **declare** its identity precisely because `platformType`-shaped literals cannot express an axis the platform splits its tree along. Re-asserting the literal is the #2063 failure the epic doc was corrected to avoid.
> - **Repo counts.** Base: 57 `JobTypeValues`, 36 `taskId` literals, 14 `HostServices` seams, each with its own file citation and a "re-count rather than quote" caveat. This copy: 55 / 34 / 15 in a single unattributed paragraph.
> - **Story count.** Base carries the *"115 story rows across 10 groups… an earlier '~90' figure… is superseded"* paragraph — the exact reconciliation PR #2888 landed on 09-11. This copy deletes it, reopening the very number #2888's last two review rounds were about.
>
> Drop the file from this branch and rebase onto the current base tip. Nothing here should be touching the epic catalogue.

**[IMPORTANT]** — `docs/plans/analysis/SPIKE-2880-ebay-sell-apis.md:246`, `docs/specs/product-spec-2880-ebay-marketplace-integration.md:133`
> *(third round; refined rather than repeated — I checked the vendor docs this time and the substance holds, the **sourcing** does not.)*
>
> eBay's own `withdrawOffer` reference does say the call *"will end the eBay listing that is associated with the offer"* while the offer *"will still remain, but only be converted from the published to unpublished state"* ([eBay Inventory API — withdrawOffer](https://developer.ebay.com/api-docs/sell/inventory/resources/offer/methods/withdrawOffer), via [entwickler.ebay.de mirror](https://entwickler.ebay.de/api-docs/sell/inventory/resources/offer/methods/withdrawOffer)). So the claim is **correct** — and that is exactly why it should be cited rather than presented as a live finding.
>
> Two things are still wrong as written. (1) This document's own header (lines 5–7) states that *every* "Confirmed/Corrected/New finding" line *"reflects a real HTTP request/response captured during this session, **not a re-read of eBay's documentation**"* — and `ENDED` appears in **no** row of the 34-row table. Evidence 7 records `status: UNPUBLISHED`, the offer's Sell-API status. A desk fact wearing a live-probe document's badge is the one thing this doc promises not to contain. (2) `ACTIVE → ENDED` as a *buyer-facing state transition* is not attested by eBay's wording either — eBay says "ends the listing", it does not name a listing-state enum. Fix: cite the withdrawOffer reference and drop the parenthetical, or add row 35 observing the buyer-facing state.

**[IMPORTANT]** — `docs/specs/product-spec-2880-ebay-marketplace-integration.md:24,131-133`; `SPIKE:245`
> The spec promotes `OfferDeactivator` to a real capability under "Key sub-decisions **carried from the spike**". It is not carried from the spike — it is a **new architectural decision that reverses a recorded one**. `docs/architecture-overview.md § Listings — Stale-variant offer pause (#1689)` states *"a first-class `OfferDeactivator` capability is a deferred follow-up"*, and `implementation-plan-1689` rejected it on the grounds that quantity-0 already stops the sale on every shipped marketplace. Reversing that needs an ADR (`docs/engineering-standards.md § Architecture Decision Records` — "a choice affects… the plugin contract", "a future maintainer might ask why we didn't just do X"), and the spec should also say **which** shape: per the `OfferCreator`/`CategoryBrowser` precedent in `architecture-overview.md § Advertised-without-dispatch sub-capabilities`, a new sub-capability is declared in the manifest and resolved by narrowing the dispatched `OfferManager` adapter with an `is*` guard — and stays **out** of `CoreCapabilityValues`. Flag it as an ADR-pending decision rather than a settled sub-decision.

**[IMPORTANT]** — PR body vs. both documents
> *(raised by @norbert-kulus-blockydevs, unaddressed.)* The body states the spike *"states plainly which checklist groups are not covered: … `C11`/`C12` (no payload obtainable), and `X5`."* **`X5` appears nowhere in either file** — I grepped both at head; zero hits, and likewise for `S3` and `S9`. Three checklist stories are neither confirmed, corrected, nor acknowledged as uncovered, while the PR body claims one of them was. Either add them to Open Risk 6's "not reduced by this session" list, or remove the claim from the body.

**[IMPORTANT]** — `SPIKE-2880-ebay-sell-apis.md:63` (third round)
> `"see New findings 8/9"` still resolves to the wrong finding. The list at head runs `1–6`, `7a…7f`, `8` — so `8` is *"Not every category aspect may be used as a variation axis"*, while row 22 is about **webhooks** (`7e`/`7f`), and `9` does not exist. The previous fix renumbered a dangling reference into a confidently wrong one. A flat `1…13` renumber fixes both at once and also makes `7a`–`7f` render as list items, which they currently do not.

**[IMPORTANT]** — PR body, `Closes #2880` (third round, mechanics not merits)
> The body grew a section arguing *whether* closing is warranted. That was never the finding. The finding is mechanical: GitHub auto-closes only on merge to the **default** branch, and this targets `2878-new-integrations-story-catalogue` — so the keyword silently does nothing, and `CLAUDE.md § Working Process` forbids hand-closing afterwards ("never close an issue manually before the PR is merged"). The spike's own line 9 already uses the correct keyword for a stacked PR: `Refs #2880`. PR #2888 made exactly this change two rounds ago. Use `Refs`, and put `Closes #2880` on whichever PR lands on `main`.

**[SUGGESTION]** — `SPIKE-2880-ebay-sell-apis.md:57,62`
> Two rows carry a live count worth pinning to a date: Evidence 10's *"all 14209 categories"* on the GB tree and Evidence 21's *"15 live topics"*. Both are snapshots of a moving surface, and the base branch's catalogue adopted exactly this caveat for its own repo counts ("these are a snapshot and will drift, so re-count rather than quote them"). One clause each.

### What holds up

Genuinely good, and worth saying since three rounds of findings can obscure it: the 34-row table is exactly 34 rows and every row carries a reproducing call; the methodological note at 257–264 (four alike failures, three of them caller-caused, distinguishable only once the probes became *differentiating*) is the right thing to record and most spikes bury it; Evidence 28's partial-`PUT` finding and Evidence 29's locale overwrite are the two most implementation-shaping results here and both are cited to a real call; and the doc's own honesty about `developer.ebay.com` answering 403 to automated fetches is corroborated — I hit the same 403 verifying withdrawOffer above.

### Verdict
❌ **Request changes** *(posted as a comment — `--request-changes` is not used in this workflow)*

1. Drop `ANALYSIS-new-marketplace-integrations-story-catalogue.md` from the branch and rebase — it reverts the base branch's ADR-037 correction and is the conflict.
2. Cite eBay's withdrawOffer reference for the end-listing claim and drop `ACTIVE → ENDED`.
3. Flag the `OfferDeactivator` promotion as an ADR-pending decision reversing #1689.
4. Reconcile the body's `X5`/`S3`/`S9` coverage claim with the documents.
5. Flat-renumber the New-findings list and repoint row 22's cross-reference.
6. `Closes #2880` → `Refs #2880`.

---

piotrswierzy (approved):
## Tech Lead re-review (`7906a1b6` → `9225bec6`, force-push)

**All six findings from my 2026-09-13 round are closed, and both SUGGESTIONs were taken.** Because the head was force-pushed I re-derived the picture from scratch against current `main` rather than trusting a `compare/` view: the branch is now **2 files, +543, 0 deletions**, both new, `mergeable: MERGEABLE`. CI is honest — `Type Check`, `Detect changed paths` and `PHP Unit Tests` pass; `Lint`, `Test`, `Build`, `Integration Tests` and `Docker Build Smoke Test` are **skipped**, correctly, on a docs-only path filter. Security swept clean again at head: no `client_secret`, no bearer token, no keyset id, no tunnel host, no `v^1.1#` eBay token literal.

- **BLOCKING — the third, stale catalogue file.** Gone. `ANALYSIS-new-marketplace-integrations-story-catalogue.md` is no longer on the branch, and #2883 has since landed it on `main`, so there is nothing left to revert. This was the cause of `CONFLICTING` and of the superseded ADR-037 re-assertion; both are resolved. ✅
- **`withdrawOffer` mis-sourcing.** `SPIKE:249-260` now cites eBay's own `withdrawOffer` reference for the end-listing claim **and** says in the same breath that *"this document did not itself observe a buyer-facing state transition (no listing-state enum is named there either), so the claim is cited to eBay's documentation rather than presented as a live finding."* The unattested `ACTIVE → ENDED` parenthetical is gone. That is a better fix than the one I asked for — it repairs the sourcing *and* preserves the line-7 invariant explicitly rather than by omission. ✅
- **`OfferDeactivator` smuggled in as a settled sub-decision.** Now flagged in all four places it appears: `SPIKE:260` (*"an architecture decision this spike does not make"*), spec §1 item 1, the shape-comparison table (`🟡 Reachable, still ADR-pending (R7)`), and a full risk row **R7** that names `architecture-overview.md § Listings`, `implementation-plan-1689`'s recorded rejection, and the ADR obligation in `engineering-standards.md`. The spec additionally names the correct **shape** — the `OfferCreator`/`CategoryBrowser` advertised-without-dispatch precedent, declared in the manifest and narrowed with an `is*` guard, **out** of `CoreCapabilityValues`. That is the #2085 trap avoided by name. ✅
- **`X5` / `S3` / `S9` claimed in the body but absent from the documents.** All three now appear in Open Risk 6 at `SPIKE:224-233`, each with its own reason (`S3` — no async quantity-ack to reconcile against; `S9` — same blocker, no cancellable order ever created; `X5` — desk research, not sandbox-testable). The body's coverage sentence now reconciles against the document. ✅
- **New-findings list renumbering.** Flat `1…13`; every entry renders as a list item, and row 22's `"see New findings 11/12"` now resolves correctly — 11 is the `createDestination` field-shape trap (*"ours"*) and 12 is the app-level notification config (*"the account's"*), exactly matching the row's own two-blocker phrasing. ✅
- **`Closes #2880` → `Refs #2880`.** Done in the body, with the mechanical reason stated correctly (auto-close fires only on merge to the default branch; `CLAUDE.md § Working Process` forbids hand-closing). ✅
- **SUGGESTION — live counts pinned.** Both taken: Evidence 10's `14209` categories and Evidence 21's `15 live topics` now carry *"a live count, snapshot only; re-query rather than quote"*. ✅

### Findings

**[SUGGESTION]** — `docs/plans/analysis/SPIKE-2880-ebay-sell-apis.md`
> Structural nit only: the evidence table is still exactly **34 rows** and the New-findings list **13** entries, both matching the body's counts — nothing to fix. Noting it so the next reader knows the arithmetic was re-checked at this head rather than carried over.

### What holds up

Worth restating now that the findings column is empty. The methodological note at the end of the evidence section — four alike failures, three of them caller-caused, distinguishable only once the probes became *differentiating* — is the kind of thing most spikes bury, and it is the reason the withdraw/publish and `createDestination` verdicts can be trusted. Evidence 28 (a partial `PUT /inventory_item` silently destroys omitted fields including the `ean` that `publishOffer` requires, so the damage is latent until the next publish) and Evidence 29 (mismatched `Content-Language` returns `204` and half-applies) are the two most implementation-shaping results in the document, both cited to a real call, and the spec turns the first into a hard read-modify-write constraint rather than general awareness. New finding 9 correctly reads eBay's raw-`<script>` round-trip as *"this destination is a 'declares nothing' case under ADR-046"* rather than as an eBay bug — which is the same reading `architecture-overview.md` records for the equivalent Shopify finding. And the honest separation of *publish is a green light* from *order ingestion is not verifiable in this sandbox, most likely because of our own account setup* is the right verdict shape for a spike: it recommends Shape B on evidence rather than on optimism.

### Verdict
✅ **Approve**

`mergeStateStatus` is `BLOCKED` pending this approval; nothing else stands in the way.

---

piotrswierzy (approved):
## Tech Lead re-verification (head unchanged at `9225bec`)

`main` has not moved and the diff is byte-identical to the one I approved, so the standing review is current — all 7 findings remain closed (stale third file gone with the catalogue landing separately via merged #2883; `withdrawOffer` cited-and-disclaimed; `OfferDeactivator` ADR-pending in four places with risk row R7 and the advertised-without-dispatch shape, kept out of `CoreCapabilityValues`; `Closes` → `Refs`).

What has changed is the gate: the PR is now `MERGEABLE` / **`CLEAN`** with `reviewDecision: APPROVED`. The approval requirement that previously held it at `BLOCKED` is satisfied. It is open only because nobody has pressed merge.

**CI, stated precisely:** `Detect changed paths`, `PHP Unit Tests` and `Type Check` pass. `Build`, `Docker Build Smoke Test`, `Integration Tests`, `Lint` and `Test` are **skipped** by the docs path filter — correctly, for a docs-only change, but they are absent rather than green and must not be read as coverage.

### Findings

### Verdict
✅ **Approve** — ready to merge; nothing further is gating it.

---

piotrswierzy (approved):
## Tech Lead re-verification (head unchanged at `9225bec`)

`MERGEABLE` / `CLEAN`, `reviewDecision: APPROVED`, base `main`, run `34820649230` settled. `Type Check`, `PHP Unit Tests` and `Detect changed paths` all pass; `Lint` / `Test` / `Build` / `Integration Tests` / `Docker Build Smoke Test` are **skipped**, which is correct path filtering for a two-file docs-only diff (`docs/plans/analysis/SPIKE-2880-ebay-sell-apis.md`, `docs/specs/product-spec-2880-ebay-marketplace-integration.md`) — absent, not passing.

**Nothing is gating this PR.** It is open only because nobody has pressed merge. Same state as my previous round, re-confirmed.

### Findings

### Verdict
✅ **Approve** — mergeable now.

---

PR #2983, opened 2026-09-08
perf(2978): F10 - behaviour under dependency failure

### LINKED ISSUE #2978: [TASK] DX - F10: behaviour under dependency failure
## Why

The #2840 programme measured six flows and every one of them ran against a stand where **nothing ever failed**. PrestaShop answered every call, the Allegro stub answered every call, Redis stayed up, the worker stayed up.

A system that is fast when everything works but loses orders when the shop hiccups is worse than a slow one, and today we know nothing about which we have. This is the gap that most directly threatens the programme's conclusion, because it is the one an operator meets first in production.

The campaign already surfaced two failures that reached a caller as success, both found by accident rather than by design:

- the Allegro stub generated buyer names containing digits, PrestaShop's `Validate::isName` rejected every one, and the job still recorded `outcome: 'ok'` because `Promise.allSettled` swallowed it
- `MarketplaceOfferQuantityUpdateHandler` drops the original exception's `statusCode` for non-contention failures, so a deterministic 404 walks the full ten-attempt retry ladder instead of dying on attempt 1

Both are the same shape: **a failure that produces a plausible artifact and no error**. There is no reason to think they are the only two.

## Scope

Measure what OpenLinker does when a dependency misbehaves, under load rather than in isolation. Each fault injected into a running window, with the same instrumentation the other flows use.

**Destination faults**
- PrestaShop returns 500 on a fraction of create-path calls
- PrestaShop returns 429 with and without `Retry-After`
- PrestaShop stops responding mid-order (connection hangs to timeout)
- a partial create: cart succeeds, `importorder` fails

**Marketplace faults**
- the source times out during hydration
- the source returns a malformed or truncated payload
- the source rejects the offer-quantity write

**Infrastructure faults**
- Redis becomes slow (not down) - the shape #2975 found is a **blocked client**, not an outage, and no health metric shows it
- Redis drops entirely mid-window
- the worker is killed mid-job

## What to establish, per fault

1. **Is the order lost, retried, or silently marked done?** The last is the failure that matters.
2. **Does any surface report success on a failed path?** Job outcome, `syncStatus`, the operator's own view.
3. **What does it cost?** A deterministic 404 burning ten attempts is measurable waste.
4. **Does recovery happen without a human?** And if so, how late.

## Acceptance criteria

- [ ] Fault injection is **declared per window in the manifest**, not applied ad hoc, so a reader can tell an injected fault from a real one
- [ ] Every fault class above is exercised at least once under load, or explicitly recorded as not exercised with the reason
- [ ] For each fault: order lost / retried / silently-succeeded is **stated**, never inferred
- [ ] Any path found reporting success on failure gets its own finding with a reproduction, at the top of the report
- [ ] The report distinguishes **measured** from **derived** and carries a "What this did not establish" section
- [ ] Cleanup restores the stand and releases `guard_stand_exclusive`

## Notes

Uses the existing harness (`perf/openlinker-throughput/`) and lab stand. The Allegro stub is the natural injection point for source faults; PrestaShop faults likely need a proxy in front of the webservice, which the `wc-tls` nginx service is a precedent for.

**Do not fix what you find inside the measuring run.** Report it; a fix landing in the same window that scores it is how a campaign stops being believable (#2977 §2).

Part of #2840.

norbert-kulus-blockydevs:
Measures what OpenLinker does when a dependency misbehaves, under load. Every
other flow in the #2840 campaign ran against a stand where nothing ever failed.

**Twelve VALID windows.** One fault class (`I3`, worker SIGKILL) was not
exercised and is recorded as such with its reason rather than quietly dropped.
Full report: `perf/openlinker-throughput/results-F10-2026-09-08.md`.

## The three findings that carry it

**1. A Redis blip kills the API and it does not come back.** Stopping Redis for
twenty seconds kills the api process on an unhandled
`SocketClosedUnexpectedlyError` from the Redis client. It stays dead after
Redis returns healthy. **Reproduced exactly on two independent runs**; both
times `docker ps` read `lab-api Exited (1)` beside `lab-redis Up (healthy)`.
The worker survived - zero socket errors in its log - so this is specific to
the api process's client, not to the codebase's Redis usage.

The "does not come back" half is the **shipped default**, not a lab artefact:
no compose file in the repository gives `api` or `worker` a restart policy.
`docker-compose.yml` has zero `restart:` declarations, and the single one in
each of the demo and lab files is on `migrate`, a one-shot job where `'no'` is
correct. This needs no load, no scale and no unusual configuration.

**2. The job ledger reports total success over an order that never reached the
shop, and nothing retries it.** Under any destination fault every
`marketplace.order.sync` job reports `succeeded / ok / attempts 0 /
lastError NULL` while 2-5 of its orders are provably absent from the shop. The
report carries the rows: five stranded order ids, their `syncStatus` entries,
the five job rows joined on the idempotency key, and PrestaShop's own `ps_cart`
showing the cart OpenLinker's error message names existing with no order behind
it.

The parent/child objection is ruled out explicitly, because it would make this
a materially smaller defect: every job of every type on **every** connection in
the window - including the destination connection - is `succeeded`/`ok` with
zero attempts. There is no child; the create happens in process. And nothing
re-drives it: no scheduler task reads `syncStatus` at all, the retry service
has exactly one caller (an HTTP route), and a re-poll dedups away permanently
against the Postgres unique constraint on `sync_jobs.idempotencyKey`.

**The pair is the finding.** `claimedMissing` is 0 in every window - OpenLinker
never claimed an order the shop does not have - and that sits *beside* the
stranding rather than softening it. The system records the failure accurately,
surfaces it well to a human (a `Sync failed` badge with the raw error inline, a
clickable "Needs attention" KPI, three Retry buttons), and then never acts on
it, while the table a monitor would alert on stays green.

**3. Source faults recover; destination faults strand.** Four source faults,
four full recoveries (12/12 orders, zero orphaned carts). Four destination
faults, four sets of stranded orders under a green ledger. One clean line,
drawn along the direction of the write. `M3` is the instructive
counter-example: the source answers **HTTP 200** carrying `status: REJECTED`,
so the transport succeeds, and OpenLinker still treats it as a failure,
retries, and recovers on its own.

Smaller: `Retry-After` changes the *outcome*, not just the timing (12/12 with
the header, 9/12 without, same status and rate); a failed create leaves an
orphaned cart nothing cleans up, with the cart id sitting unused in
OpenLinker's own error string; and the detail page's "View failed orders" link
goes to a page that by construction excludes them.

## How the faults are injected

Two injectors, both **proven correct in isolation before any window was
spent** - 49 proxy checks and 27 stub tests, run in-process against a throwaway
upstream so they work while a peer holds the stand lock.

- `stubs/ps-fault-proxy` - a transparent reverse proxy in front of the **real**
  PrestaShop. A mock would change two variables at once; every unfaulted
  request reaches the same shop as in every other window of the campaign.
  Started as a plain `docker run`, so no existing container is recreated and
  `post_guard_containers_stable` keeps its meaning.
- the Allegro stub's fault surface, extended with per-endpoint targeting, a
  per-request fraction, and three modes chosen for *where* they fail:
  `malformed` (the parse), `truncated` (the transport), `reject-quantity`
  (neither - the transport succeeds).

Self-testing earned itself immediately: the proxy's own suite caught that a
default `429` sent no `Retry-After` at all while the control surface and the
manifest both claimed one was sent - which would have made the `D2a` and `D2b`
arms measure the same thing under two labels.

Every fault is **declared per window** in the manifest, the rule is read back
from the injector rather than echoed from what the script asked for, and the
**delivered** count is read back afterwards - so a rule that installed and
never matched cannot read like a system that absorbed the fault.

## Honesty notes

`run_post_guards` is deliberately not used wholesale: three of its members are
inverted by a fault window. Each guard runs individually and is routed to a
stand-invalidating list or an observation list; the two that do not run are
named with reasons. `drain_wait` was reordered to run *after* the settled
ledger read, because it marks leftover rows `dead` and would otherwise make a
row the harness killed indistinguishable from one the system dead-lettered.

Four windows were discarded and are kept under names that say why. **One
retraction is in the report rather than quietly corrected**: I stopped a window
on a diagnosis of peer contamination and was wrong - the rows were this run's
own downstream cascade. A second wrong reading, of an order as one the shop held
and OpenLinker did not know about, is also corrected in place: the identifier
mapping was written, so no duplicate exists and none was observed.

§ 8 carries nine things this did not establish.

Nothing found here is fixed. #2977 §2 refused a fix landing in the window that
scores it.

Closes #2978.

https://claude.ai/code/session_01SkKifZVwvZzHspxaQ7x396

### DIFF
--- a/perf/openlinker-throughput/results-F10-2026-09-07.md
+++ b/perf/openlinker-throughput/results-F10-2026-09-07.md
@@ -1,0 +1,297 @@
+# F10 - behaviour under dependency failure
+
+_Run 2026-09-07 on the `lab` stand. Scenario `scenarios/f10-dependency-failure.sh` (#2978, epic #2840)._
+
+> **STATUS: METHOD AND CONDITIONS ONLY.** The measurement sections below are
+> not yet filled. Nothing in this file may be quoted as a result until this
+> banner is gone.
+
+---
+
+## 0. Why this run exists
+
+Every one of the six flows the #2840 campaign measured ran against a stand
+where **nothing ever failed**. PrestaShop answered every call, the Allegro stub
+answered every call, Redis stayed up, the worker stayed up. A system that is
+fast when everything works but loses orders when the shop hiccups is worse than
+a slow one, and until this run the campaign knew nothing about which OpenLinker
+is.
+
+The campaign has already met a **failure reported as a success** twice, and
+both times by accident rather than by design:
+
+- the Allegro stub generated buyer names containing digits, PrestaShop's
+  `Validate::isName` rejected every one, and the job still recorded
+  `outcome: 'ok'` because `Promise.allSettled` swallowed it;
+- `MarketplaceOfferQuantityUpdateHandler` drops the original exception's
+  `statusCode` for non-contention failures, so a deterministic 404 walks the
+  full ten-attempt retry ladder instead of dying on attempt 1.
+
+Finding that shape a third time by accident is not a method. This run goes
+looking for it.
+
+---
+
+## 1. The question, and what counts as an answer
+
+Per fault, one thing:
+
+> **is the order LOST, RETRIED, or SILENTLY MARKED DONE?**
+
+Each is **stated from a reading, never inferred from a rate that looks
+plausible**. The four readings, per window:
+
+| Source | What it answers |
+|---|---|
+| `sync_jobs` | what the job says happened - `status`, `outcome`, `outcomeReason`, `attempts`, `deferredTotalMs`, `lastError` |
+| `order_records` | what the order snapshot says happened |
+| `order_records.syncStatus[]` | what the destination fan-out says happened, and the destination-native order id it CLAIMS to have created |
+| `ps_orders` / `ps_cart` | whether that order actually exists |
+
+**Ground truth is the shop, not OpenLinker.** "Silently marked done" cannot be
+established from OpenLinker's own tables, because a system that believes it
+succeeded says so consistently everywhere it writes. The sharp test is
+therefore the last two rows read together: for every `syncStatus` entry with
+`status: 'synced'`, does its `externalOrderId` name a row in `ps_orders`? A
+claimed id with no row behind it is a success report over a failure.
+
+That detector was **proven able to fire before the run**, against live data on
+this stand, rather than assumed: three genuinely-claimed order ids answered
+3 present (missing 0), and the same query with one id replaced by a fabricated
+`999999` answered 2 present (missing 1). A detector nobody has watched go red
+is a detector that reports zero for the wrong reason.
+
+An orphaned cart is the second ground-truth signal: `psCartsCreated > 0`
+alongside `psOrdersCreated == 0` is work done at the destination that produced
+no order.
+
+---
+
+## 2. Conditions
+
+_(Filled from the run's own manifests - see § 8.)_
+
+- **Stand**: `lab` (`docker-compose.lab.yml`, #2854), a single developer
+  workstation shared with two other OpenLinker docker-compose stacks
+  (`ol-demo-fresh-*`, `openlinker-*`). No CPU pinning, no memory limits.
+- **Peer activity**: TBD - stated per chunk, because two peers were active on
+  this stand during the preparation of this run and the stand lock was
+  contended twice before the first window opened.
+- **Image**: TBD (`guard_build` records the image commit, not `HEAD`).
+- **Destination**: the **real local PrestaShop**, reached through the
+  fault-injection proxy. Never a mock - see § 3.
+- **Source**: the Allegro upstream stub (#2856/#2935) at
+  `http://allegro-stub:8080`, tenant `perf-allegro-a`, at its configured
+  276 ms / 1106 ms per-endpoint latencies.
+- **Runner**: ENABLED. The lab stand ships `WORKER_RUNNER_ENABLED=false`, so
+  the scenario recreates the worker with it on and restores it on exit - a
+  fault injected into a system that executes no jobs measures the injector.
+- **Scheduler**: OFF (`guard_scheduler_off`), so no master sweep shares a lane
+  with this run's jobs.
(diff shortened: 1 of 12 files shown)


norbert-kulus-blockydevs:
**One commit is staged but not yet pushed: the completed report.**

`gpg-agent`'s cached passphrase on this machine expires and is refreshed
outside this session's control - it signed four of this branch's commits and
refused twice. I have not used `--no-gpg-sign`, deliberately: every commit on
this branch is signed and one unsigned commit would spoil that for no reason
other than impatience.

`perf/openlinker-throughput/commit-when-gpg-ready.sh` is armed in the
background and will commit **and push** the report the moment signing works
again. If it gives up, the report is staged in the worktree and nothing was
lost - `git commit -s -S --no-verify -F <msg>` is all that is needed.

What is already pushed is the whole measurement: the two injectors and their
test suites, the 13-window scenario, and the harness fixes. What is staged is
the report's § 6 results table, the reordered § 7 findings, § 8 "what this did
not establish", and § 9 "fault classes not exercised". (`results/` is
gitignored, so the raw per-window ledgers live only on the stand; every figure
quoted in the report is reproduced in it.)

**Stand state, verified after the run:** the `perf-prestashop` connection is
back at `http://prestashop`, `perf-woocommerce` has its `OrderProcessorManager`
capability back, the worker is at `WORKER_RUNNER_ENABLED=false` /
`OL_SCHEDULER_ENABLED=false` as found, the fault proxy is removed, the Allegro
stub carries no armed fault, and `guard_stand_exclusive` is released.

`lab-api` had to be restarted by hand - twice - because the `I2` window killed
it and it does not come back on its own. That is finding F10-0, met as an
operational fact rather than only as a measurement.

One stand change is left behind and is disclosed in the report's § 2:
`lab-allegro-stub` now runs a locally-built image carrying the #2978 fault
extensions and is no longer compose-managed. With no fault armed it is
behaviourally identical to the compose-built one, and a `docker compose up`
will replace it.

Issue #2986, opened 2026-09-08
Global integration-harness beforeEach: six int-spec files never reset the harness


norbert-kulus-blockydevs:
## Problem

`apps/api/test/integration/setup.ts` provides `resetTestHarness()`, which most int-spec files call in `beforeEach` (and some also in `afterEach`) to guarantee test isolation against the one shared Testcontainers Postgres instance. Six files call it **nowhere at all**:

- `api-versioning.int-spec.ts`
- `automation-dispatch-gate.int-spec.ts`
- `bootstrap-admin-disabled.int-spec.ts`
- `fulfillment-work-migration-parity.int-spec.ts`
- `listings-offer-status-snapshot.int-spec.ts`
- `oms-connection-never-seeded.int-spec.ts`

Under CI's own conditions (`maxWorkers: 1`, one shared Postgres, all integration files run in sequence), any of these six running immediately before a file that resets **only** in `afterEach` (rather than `beforeEach`) leaves that file's first test case reading rows left behind by whichever file ran before it.

## Evidence

This surfaced during review of PR #2957 (`perf: the paginated total becomes a second stage`, round 5): `paginated-total-split.int-spec.ts` originally reset only in `afterEach`, and depending on execution order this produced a real CI failure — 41 assertions in that file were passing only because of accidental row state left by whichever spec ran immediately before it, not because the assertions were actually true in isolation. That specific file was fixed locally (it now resets in both `beforeEach` and `afterEach`), but the six files above remain unguarded, so the same class of failure is reachable again the next time file execution order changes or a new int-spec is added between them.

## Proposed fix

Add a global `beforeEach` (and ideally `afterEach`) call to `resetTestHarness()` in `apps/api/test/integration/setup.ts` (or equivalent shared Jest setup) so every integration spec is isolated by default, rather than relying on each new file's author to remember to call it. Audit whether the six listed files have any reason to opt out (e.g. they intentionally test cross-file state) before making the reset unconditional — if any do, they should call it out explicitly.

## Context

Reported (not fixed) in PR #2957 review round 6, and flagged again in the second Tech Lead re-review — see https://github.com/openlinker-project/openlinker/pull/2957#pullrequestreview-5129027335.

norbert-kulus-blockydevs:
Closed by #2996, merged into `perf-programme-2840` as b6be137.

[...]

Issue #2469, opened 2026-08-25
[TASK] CORE + Interface — Wire the tax-inclusion setting into the net-sales query, plus a bulk re-run-backfill trigger for category C


norbert-kulus-blockydevs:
### Problem / Context

Phase 3 persists `includeGuessedVatRatesInNetSales`; Phase 1 Task 1.3 defines exactly what it must do to `netSalesOrderNetEligibleSql` / `netSalesLineNetEligibleConditionSql`. This task is the wiring between the two — a CORE change to two SQL-builder call sites, not a remediation endpoint. Category C additionally gets a genuine bulk action: `TaxRateBackfillService` already runs on a schedule, and an operator seeing a large category-C count in the coverage panel shouldn't have to wait for the next scheduled tick.

### Proposed Solution

- Extend `netSalesOrderNetEligibleSql` / `netSalesLineNetEligibleConditionSql` (`libs/core/src/orders/domain/types/net-sales-tax-rate.types.ts`) with an additional parameter (e.g. `includeBackfilledPreRollout: boolean`) that, when true, changes the literal `rec."taxRateEra" IS DISTINCT FROM 'pre-rollout'` clause to also admit a `pre-rollout` row **whose rate resolves** (mirrors Phase 4 Task 4.2's category-A definition exactly — reuse that same resolution check, do not re-derive it). Every call site in `order-record.repository.ts` / `order-line-item.repository.ts` reads the flag from Phase 3's settings table.
- `POST /analytics/coverage/tax/rerun-backfill` for category C: re-invokes `TaxRateBackfillService`'s existing resolution path on-demand for the named order ids — the same mechanism that already runs on a schedule, just triggered early. This is a synchronous or lightweight-async call (no `analytics_remediation_runs` tracking needed — a backfill attempt is idempotent and safe to re-request without a run ledger).
- Category B gets **no endpoint** — the mockup's `detail-novat` state correctly shows no "fix" action, only the informational row, because the underlying data genuinely has no resolvable rate.

### Classification

**Type**: CORE + Interface
**Layer**: Domain + Application + Interface
[...]

PR #2993, opened 2026-09-08
fix(web/analytics): confirm before turning on tax-rate-inclusion toggle

### LINKED ISSUE #2857: [TASK] Frontend — Analytics Settings dialog is missing the tax-rate-inclusion confirm step (mockup drift)

norbert-kulus-blockydevs:
## Summary
- `AnalyticsSettingsDialog` wrote `includeBackfilledTaxRatesInNetSales=true` directly from the checkbox `onChange`, with no confirm step — unlike the currency-recalculate action, which is gated behind `confirmingRecalculate` + a real `ConfirmDialog` (#2668 review, finding 14). The toggle's own description already states the write is deployment-wide and re-queried-not-retroactive — the same class of one-click permanent financial write the recalculate confirm exists for.
- Mirrors that exact pattern for the ON direction only (turning the setting back OFF stays a direct write — the safe, reversing direction).
- Un-skips the mockup's `tax-confirm` state in the #2482 E2E parity spec: it now drives the real confirm flow instead of asserting the divergence.
- **Also includes an unrelated but necessary DX fix**: `scripts/smart-test.mjs`'s runner-detection defaulted an absent `test` script to `jest` unconditionally. `apps/e2e` has no `test` script and no jest/vitest dependency at all (Playwright-only), so the pre-commit hook's smart-test step tried to run jest against a Playwright `.spec.ts` file and failed to even parse it — blocking any commit that touches an e2e spec, unrelated to this PR's actual change. Fixed the runner detection to resolve to `'none'` (skip) when a package has neither a jest/vitest `test` script nor jest/vitest as a dependency; added a self-check case.
- **Also includes a CI flake fix, unrelated to this PR's change**: `apps/api/test/integration/invoicing-list.int-spec.ts` reset state only in `afterEach`, so its first test (asserting `total: 0`) depended on whatever the previous file in the shared-Postgres, single-worker suite left behind. Added a `beforeEach` reset so this file's isolation no longer depends on run order or another file's cleanup discipline (refs #2986).

## Test plan
- [x] `pnpm --filter web lint` — 0 errors
- [x] `pnpm --filter web type-check` — clean
- [x] `pnpm --filter @openlinker/e2e lint` — 0 errors
- [x] `pnpm --filter @openlinker/e2e type-check` — clean
- [x] `node scripts/smart-test.mjs --self-check` — 15/15 passed
- [x] Pre-commit hook (full `pnpm -r lint` + smart-test) passed; `apps/web` vitest run: 156/156 passed
- [ ] Unit tests not run via a separate manual `pnpm test` locally (weak dev machine) — covered by the pre-commit hook's `vitest related` run above
- [ ] E2E `tax-confirm` state not run headfully here — logic traced manually against `ConfirmDialog`'s rendered markup

Closes #2857

### DIFF
--- a/apps/api/test/integration/invoicing-list.int-spec.ts
+++ b/apps/api/test/integration/invoicing-list.int-spec.ts
@@ -79,6 +79,19 @@ describe('Invoicing list (integration)', () => {
     await resetTestHarness();
   });
 
+  // The FIRST test asserts `total: 0` against a Postgres instance shared with
+  // every other int-spec file in this suite (single worker, no
+  // testSequencer) — an `afterEach`-only reset is a courtesy to the next
+  // file, not isolation for this one, so this file's first assertion can
+  // read whatever the previous file left behind.
+  // `invoice-record-offline-recovery-query.int-spec.ts`'s own
+  // `beforeEach`-only reset leaves its last test's rows in `invoice_records`
+  // after that file finishes, which is exactly what was observed failing
+  // this test in CI (#2986).
+  beforeEach(async () => {
+    await resetTestHarness();
+  });
+
   afterEach(async () => {
     await resetTestHarness();
   });
--- a/apps/e2e/tests/analytics/mockup-parity.spec.ts
+++ b/apps/e2e/tests/analytics/mockup-parity.spec.ts
@@ -23,14 +23,11 @@
  *
  * 15 states total, per the issue's own state table.
  *
- * Fully compared (screenshot + content, both mockup and real app) — 11:
+ * Fully compared (screenshot + content, both mockup and real app) — 12:
  * `native`, `converting`, `converted`, `unavailable`, `settings-open`,
  * `all-clear`, `detail-currency`, `currency-in-progress`, `currency-fixed`,
- * `detail-mapping`, `detail-novat`.
- *
- * Documented divergence (mockup screenshotted; real app asserted to NOT have
- * the mockup's confirm step — a real product gap, not a spec defect) — 1:
- * `tax-confirm`. See #2857.
+ * `detail-mapping`, `detail-novat`, `tax-confirm` (real confirm-gate now
+ * shipped, #2857).
  *
  * Skipped with a named, verified reason (mockup screenshotted only) — 3:
  * `detail-tax` / `detail-postrollout` (tax-a / tax-c: `taxRateEra =
@@ -391,35 +388,55 @@ test.describe('analytics mockup parity (#2482)', () => {
         }
       });
 
-      // ── tax-confirm: documented mockup/real-app divergence (#2857) ──────
+      // ── tax-confirm: now exercisable in the real app (#2857) ────────────
       await test.step('tax-confirm', async () => {
         await pages.analyticsMockup.gotoState('tax-confirm');
-        await captureBoth(
-          testInfo,
-          pages.analyticsMockup.regionFor('tax-confirm'),
-          page,
-          'tax-confirm-mockup-reference',
-        );
-        // The mockup shows a confirm dialog ("Turn on including orders with a
-        // guessed tax rate?") before the toggle applies. The shipped
-        // AnalyticsSettingsDialog's `handleTaxToggleChange` writes directly
-        // on change with no confirm step (#2857) — asserted here as a real,
-        // known divergence rather than silently glossed over.
+        // The mockup shows a confirm dialog before the toggle applies. The
+        // shipped `AnalyticsSettingsDialog` now gates the ON direction behind
+        // a real `ConfirmDialog` too (#2857, mirroring the currency-recalculate
+        // confirm precedent from #2668 review, finding 14).
         await pages.analytics.openSettings();
         const toggle = pages.analytics.settingsDialog.getByRole('checkbox', {
           name: /Use the rate found in the product catalog/,
         });
         await toggle.waitFor({ state: 'visible' });
         const before = await toggle.isChecked();
-        await toggle.click();
-        await expect(
-          page.getByRole('dialog', { name: /Confirm including tax-rate-guessed orders/ }),
-          'known divergence #2857: the mockup shows a confirm dialog here; the shipped Settings dialog toggles directly with no confirm step',
-        ).toHaveCount(0);
-        // Restore the toggle to its prior value so this run leaves the
-        // deployment-wide setting unchanged.
-        if ((await toggle.isChecked()) !== before) {
+        try {
+          if (before) {
+            // Already ON from a prior run — flip OFF (direct write, no confirm
+            // gate on that direction) so this step can exercise the ON confirm.
+            await toggle.click();
+            await expect(toggle).not.toBeChecked();
+          }
           await toggle.click();
+          const confirmDialog = page.getByRole('dialog', {
+            name: 'Turn on including orders with a guessed tax rate?',
+          });
+          await expect(confirmDialog).toBeVisible();
+          // Capture the REAL confirm dialog, not the settings page behind it —
+          // capturing before this point (as a prior revision did) screenshots
+          // the analytics page in its default state and silently never
+          // exercises the "fully compared" claim this state is listed under
+          // above (#2993 review, IMPORTANT 1).
+          await captureBoth(testInfo, pages.analyticsMockup.regionFor('tax-confirm'), page, 'tax-confirm');
+          // Cancelling must not write anything — the checkbox stays OFF.
+          await confirmDialog.getByRole('button', { name: 'Cancel' }).click();
+          await expect(confirmDialog).toBeHidden();
+          await expect(toggle).not.toBeChecked();
+        } finally {
+          // Restore the deployment-wide setting to its prior value —
+          // best-effort, in a `finally`, so a failure between the OFF click
+          // above and this restore does not leave the setting flipped for
+          // the whole environment (#2993 review, SUGGESTION: "leaves it
+          // unchanged" was true only on the happy path).
+          if (before && !(await toggle.isChecked())) {
+            await toggle.click();
+            await page
+              .getByRole('dialog', { name: 'Turn on including orders with a guessed tax rate?' })
+              .getByRole('button', { name: 'Turn on' })
+              .click();
+            await expect(toggle).toBeChecked();
+          }
         }
         await page.getByRole('button', { name: 'Cancel' }).click();
       });
--- a/apps/web/src/features/analytics/components/analytics-settings-dialog.test.tsx
+++ b/apps/web/src/features/analytics/components/analytics-settings-dialog.test.tsx
@@ -1,5 +1,5 @@
 import { describe, expect, it, vi } from 'vitest';
-import { screen, waitFor } from '@testing-library/react';
+import { screen, waitFor, within } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import {
   renderWithProviders,
@@ -156,6 +156,9 @@ describe('AnalyticsSettingsDialog', () => {
       name: /Use the rate found in the product catalog/,
     });
     await userEvent.click(taxToggle);
+    // Turning the toggle ON opens a confirmation before writing anything
+    // (#2857, mirroring #2668 review, finding 14).
+    await userEvent.click(await screen.findByRole('button', { name: 'Turn on' }));
 
     await waitFor(() => {
       expect(updateSettings).toHaveBeenCalledWith({
@@ -193,12 +196,98 @@ describe('AnalyticsSettingsDialog', () => {
       name: /Use the rate found in the product catalog/,
     });
     await userEvent.click(taxToggle);
+    await userEvent.click(await screen.findByRole('button', { name: 'Turn on' }));
 
     await waitFor(() => {
       expect(updateSettings).toHaveBeenCalledWith({
         displayCurrency: null,
         rateBasis: 'current',
         includeBackfilledTaxRatesInNetSales: true,
+        netGrossBasis: 'gross',
+      });
+    });
+  });
+
+  it('should gate turning the tax-rate toggle ON behind a confirm dialog, leaving it untouched on Cancel (#2857)', async () => {
+    const updateSettings = vi.fn().mockResolvedValue(undefined);
+    const apiClient = createMockApiClient({
+      analyticsSettings: {
+        getSettings: vi.fn().mockResolvedValue({
+          displayCurrency: 'PLN',
+          displayCurrencySource: 'default',
+          rateBasis: 'current',
+          includeBackfilledTaxRatesInNetSales: false,
+          netGrossBasis: 'gross',
+          updatedAt: null,
+          updatedByUserId: null,
+        }),
+        updateSettings,
+      },
+    });
+
+    renderWithProviders(<AnalyticsSettingsDialog {...baseProps} />, {
+      apiClient,
+      sessionAdapter: createAuthenticatedSessionAdapter(),
+    });
+
+    const taxToggle = await screen.findByRole('checkbox', {
+      name: /Use the rate found in the product catalog/,
+    });
+    await userEvent.click(taxToggle);
+
+    // Clicking the checkbox alone must not write anything yet.
+    const confirmDialog = await screen.findByRole('dialog', { name: 'Turn on including orders with a guessed tax rate?' });
+    expect(updateSettings).not.toHaveBeenCalled();
+
+    await userEvent.click(within(confirmDialog).getByRole('button', { name: 'Cancel' }));
+
+    await waitFor(() => {
+      expect(screen.queryByRole('dialog', { name: 'Turn on including orders with a guessed tax rate?' })).not.toBeInTheDocument();
+    });
+    expect(updateSettings).not.toHaveBeenCalled();
+  });
+
+  it('should turn the tax-rate setting OFF directly, with no confirm dialog (#2857, #2993 review)', async () => {
+    const updateSettings = vi.fn().mockResolvedValue(undefined);
+    const apiClient = createMockApiClient({
+      analyticsSettings: {
+        getSettings: vi.fn().mockResolvedValue({
+          displayCurrency: 'EUR',
+          displayCurrencySource: 'setting',
+          rateBasis: 'current',
+          includeBackfilledTaxRatesInNetSales: true,
+          netGrossBasis: 'gross',
+          updatedAt: null,
+          updatedByUserId: null,
+        }),
+        updateSettings,
+      },
+    });
+
+    renderWithProviders(<AnalyticsSettingsDialog {...baseProps} />, {
+      apiClient,
+      sessionAdapter: createAuthenticatedSessionAdapter(),
+    });
+
+    const taxToggle = await screen.findByRole('checkbox', {
+      name: /Use the rate found in the product catalog/,
+    });
+    expect(taxToggle).toBeChecked();
+
+    await userEvent.click(taxToggle);
+
+    // Turning OFF is the safe, reversing direction — it writes immediately,
+    // with no confirm dialog gating it (the ON direction only, per #2857's
+    // own AC). A future edit that starts gating both directions, or drops
+    // the `else` branch, must fail this test.
+    expect(
+      screen.queryByRole('dialog', { name: 'Turn on including orders with a guessed tax rate?' })
+    ).not.toBeInTheDocument();
+    await waitFor(() => {
+      expect(updateSettings).toHaveBeenCalledWith({
+        displayCurrency: 'EUR',
+        rateBasis: 'current',
+        includeBackfilledTaxRatesInNetSales: false,
         netGrossBasis: 'gross',
       });
     });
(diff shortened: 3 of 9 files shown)


piotrswierzy (commented):

The component itself is right, and closes a real drift (#2857): the tax-inclusion toggle wrote a deployment-wide financial-reporting setting straight from a checkbox `onChange`, while the sibling currency-recalculate on the same dialog has been confirm-gated since #2668. The fix mirrors that precedent precisely — same `ConfirmDialog` primitive, same elevated-nesting classes, gating the ON direction only. No BLOCKING issues.

The concern is that **the claims made around the change outrun the change**. Four of them.

### IMPORTANT — the e2e header promotes `tax-confirm` to "fully compared", but the real dialog is never screenshotted

`apps/e2e/tests/analytics/mockup-parity.spec.ts` (header ~23–31; step ~391–430)

`captureBoth(...)` is called at ~394 — **before** `pages.analytics.openSettings()` — and still under the artifact name `tax-confirm-mockup-reference`, a leftover from when this was a documented divergence. `captureBoth` screenshots the mockup region *and* `realPage.screenshot({fullPage: true})`, so the "real app" image for this state is the analytics page in its default state, not the confirm dialog.

So the dialog is compared by content assertions only, while the header now promises screenshot + content for both. The taxonomy entry is what a future reader trusts, which makes this the "check that cannot fail" shape. Either move `captureBoth` to after `await expect(confirmDialog).toBeVisible()` and rename the artifact to `tax-confirm`, or put `tax-confirm` in a distinct content-compared-only bucket.

### IMPORTANT — the committed mockup is now stale on four axes, with nothing recording it

`docs/frontend-architecture.md` § UX Mockups is explicit that this obligation is **reviewer-enforced only**, and that a PR which neither touches the mockup nor explains why it's still accurate is treated as incomplete. This PR does neither, and `docs/plans/mockups/analytics-display-currency-picker.html` (`tax-confirm`, ~888–899) genuinely diverges:

| | Mockup | Shipped |
|---|---|---|
| Title | `Turn on including orders with a guessed tax rate?` | `Turn on this setting?` |
| Warning callout | `⚠ These tax rates come from TODAY'S catalog, not the order date…` | absent |
| Scope line | `Affects 18 orders, 6 products.` | absent |
| Role | `alertdialog` | `dialog` (Radix/`ConfirmDialog`) |

The implementation plan says the title *"mirrors the mockup's button copy"* — but that's the **trigger button** in the `detail-tax` dialog (~997), not this dialog's title. Some divergence is sanctioned by #2857's AC; the point is that none of it is recorded, and (per the finding above) the parity gate cannot surface it either.

### IMPORTANT — the copy-drift claim is false, and the two copies have already drifted in this diff

`analytics-settings-dialog.tsx` (394–397 and 443)

The confirm description is a verbatim duplicate of the checkbox's own `__toggle-desc` text, with no shared constant. The plan's § 5 claims *"Confirm dialog copy is asserted (test) to match the toggle's own copy, preventing drift"* — **no such assertion exists**.

And they have already diverged here: line 396 renders `you&rsquo;re` (typographic apostrophe), line 443 renders `you're` (straight). Extract one `const TAX_INCLUDE_CONSEQUENCE` referenced by both, or delete the claim from the plan — as shipped, the drift-prevention argument used to justify reusing the copy isn't in place.

### IMPORTANT — the docblock's stated rationale contradicts the documented backend semantics

`analytics-settings-dialog.tsx` (155–160)

> *"A real, **permanent**, deployment-wide financial-reporting write…"*

`docs/architecture-overview.md` § Orders (#2469) says the opposite in terms: *"**No `order_records` row is mutated by flipping the setting** — it is a pure query-time gate, **instantly reversible in both directions**."* The mockup's own confirm says the same, and so does the toggle description this dialog reuses.

The gate is still defensible on the *deployment-wide, invisible to whoever flipped it* axis — that alone justifies it. But this is the comment the next reader acts on (e.g. by gating OFF too, or refusing to reconsider the gate). Reword to the true justification.

### IMPORTANT — the OFF direction has no test

`analytics-settings-dialog.test.tsx`

`onChange` now has two branches; only one is covered. All three tax tests start from `false` and exercise OFF→ON. Nothing asserts ON→OFF still writes directly **without** a confirm — which is #2857's own AC (*"Turning it OFF still writes directly (no regression)"*) and is precisely the branch this PR introduced. A future edit that gates both directions, or drops the `else`, fails no test.

### IMPORTANT — PR composition

Three unrelated concerns ship together: the #2857 gate, the `smart-test.mjs` runner-detection fix (declared), and a CI-flake fix in `apps/api/test/integration/invoicing-list.int-spec.ts` attributed to **#2986** — not mentioned in the description or the plan. Also, commit `55aed1231` is titled `fix(dx): smart-test no longer defaults an e2e-only package to jest` but carries the entire feature; per `engineering-standards.md` § Commit Messages the subject should name the primary change.

The int-spec change itself is sound — `resetTestHarness()` in `afterEach` alone is a courtesy to the *next* file, leaving the first `total: 0` assertion exposed to whatever the previous file left behind.

### SUGGESTIONS

- `isConfirming={updateSettings.isPending}` (445) is inert: `setConfirmingTaxInclude(false)` runs before `handleTaxToggleChange(true)`, so the dialog is already closed when `isPending` flips. It mirrors `handleRecalculate` (139) exactly, so it's consistent precedent rather than a new defect — but it's dead in both.
- The e2e step now performs up to **three real writes** to a deployment-wide setting. The comment claims *"leaves it unchanged"*, true only on the happy path — a failure between the OFF click and the restore leaves it `false` for the whole environment. Restore in a `finally`, or seed via the API.
- `smart-test.mjs` (202): the catch fallback moved from `'jest'` to `'none'`, so an unreadable `package.json` now **silently skips** that package in the pre-commit gate — fail-open in a safety gate, and `--self-check` only exercises the pure `testRunnerForScript`, never the catch. Keep it at `'jest'` or warn on that path. (I did verify the rest is safe: scanning every workspace manifest, `apps/e2e` is the **only** package the new rule reclassifies, and its 57 specs are all Playwright.)
- The confirm restates, word for word, the sentence the operator is reading right next to the checkbox they just clicked — so it currently adds friction without information. `taxA` is already computed at line 130; `Affects {taxA} orders.` is a one-line change that would make the gate earn its interruption.

### Done well

- **State ownership is exactly right** — `confirmingTaxInclude` is component-local `useState`, which `frontend-architecture.md` § Local UI State names as the canonical case. No Query data copied into local state.
- **Accessibility comes from the primitive, correctly** — `ConfirmDialog` already owns the focus trap, Escape dismissal and `onOpenAutoFocus`; the accessible name derives from `DialogTitle`, which is what makes `getByRole('dialog', {name: …})` a real assertion rather than a coincidence. No bespoke dialog markup.
- **The checkbox stays server-driven** (`checked={settingsQuery.data?.…}`), so Cancel needs no rollback state — the "nothing was written" guarantee is structural, not compensating.
- The new test asserts the negative correctly (`not.toHaveBeenCalled()` both before and after Cancel) and scopes Cancel with `within(confirmDialog)`, avoiding the settings dialog's own Cancel — a real ambiguity that would bite a naive `screen.getByRole`.
- No raw `fetch()`, no new tokens or hex, no new CSS; permission gating untouched and still correct.

Branch is `BEHIND` main — worth rebasing, as the analytics e2e spec has had recent churn from #2668.

---

norbert-kulus-blockydevs:
## Addressed review findings

All six IMPORTANT findings and the four SUGGESTIONS are addressed in `5702a97e3` (amended), plus the two follow-on commits.

**IMPORTANT — screenshot never exercised the real dialog**: `captureBoth` in `mockup-parity.spec.ts` now runs *after* `await expect(confirmDialog).toBeVisible()`, and the artifact is renamed from `tax-confirm-mockup-reference` to `tax-confirm` — it now actually screenshots the confirm dialog instead of the analytics page behind it.

**IMPORTANT — stale mockup**: `docs/plans/mockups/analytics-display-currency-picker.html`'s `tax-confirm` state now carries an inline comment documenting every recorded divergence from the shipped dialog (the `alertdialog`→`dialog` role correction to match every other `ConfirmDialog` in the app, the dropped per-rate breakdown, the order-count-only scope line, the reused description), per `frontend-architecture.md` § UX Mockups. `role="alertdialog"` in the mockup is corrected to `role="dialog"`.

**IMPORTANT — copy drift**: extracted one `TAX_INCLUDE_CONSEQUENCE` JSX constant, referenced by both the checkbox description and the confirm dialog description, so the typographic-vs-straight-apostrophe drift can't recur.

**IMPORTANT — docblock rationale contradicted the backend semantics**: reworded — the write is no longer described as "permanent"; the comment now cites `architecture-overview.md` § Orders (#2469) directly and states the gate is warranted because the write is deployment-wide and invisible to whoever flipped it, not because it's irreversible.

**IMPORTANT — OFF direction untested**: added a test asserting OFF still writes directly with no confirm dialog, exercising the exact branch this PR introduced.

**IMPORTANT — PR composition**: the three concerns are now three commits, each named for what it actually contains and cross-referencing the others (`fix(web/analytics)` for the #2857 gate, `fix(dx)` for the smart-test.mjs runner-detection fix, `test(api)` for the #2986 CI-flake fix), rather than one commit whose title didn't match its content.

**SUGGESTION — dialog title/copy vs. mockup**: dialog title now reads `Turn on including orders with a guessed tax rate?`, matching the mockup; body now renders the warning callout and an `Affects {taxA} order(s)` scope line backed by the existing coverage-query read, replacing the copy that only restated the checkbox's own sentence.

**SUGGESTION — e2e restore not crash-safe**: the restore step in `mockup-parity.spec.ts` now runs in a `finally` block, so a failure between the OFF click and the restore can't leave the deployment-wide setting flipped for the whole environment.

**SUGGESTION — `smart-test.mjs` fail-open catch**: reverted the catch-path fallback from `'none'` (silent skip) back to `'jest'` (fail loud) — only a `package.json` that parses cleanly and genuinely declares neither runner resolves to `'none'`. Added a dedicated self-check case exercising the catch path itself (`testRunnerForScript` alone never reaches it).

**SUGGESTION — dead `isConfirming` prop**: left as-is — it's pre-existing precedent shared with `handleRecalculate` above (same dead condition there too), not something introduced by this PR; fixing it here would mean touching the currency-recalculate flow, which is out of scope.

Verified: `pnpm --filter @openlinker/web exec tsc --noEmit` clean, ESLint clean on the touched `.tsx`/e2e files, `node scripts/smart-test.mjs --self-check` passes all 16 cases (including the new catch-path case). Branch was also rebased onto current `main`.

norbert-kulus-blockydevs:
## Frontend/UX re-review (post-fixup)

This PR already went through one review round (all 6 IMPORTANT + 4 SUGGESTIONS addressed per the author's own comment above). I re-verified the fixes against the actual diff rather than trusting the summary, focusing on UI/UX correctness.

**Verified good:**
- **Confirm-dialog pattern matches precedent exactly** — same `dialog__overlay--elevated`/`dialog__content--elevated` nested-dialog classes as the existing `confirmingRecalculate` flow, same shared `ConfirmDialog` primitive (Radix-backed: focus trap, Escape-to-close, `aria-modal` all inherited for free).
- **No flicker/race on the toggle**: the checkbox is fully controlled off `settingsQuery.data?.includeBackfilledTaxRatesInNetSales`, so opening the confirm (`setConfirmingTaxInclude(true)`) never flips the visible checkbox state — it only becomes checked once the mutation actually succeeds and the query refetches. Double-clicking the checkbox while the dialog is open is a no-op (`setConfirmingTaxInclude(true)` again), not a race.
- **Copy drift fixed for real**: `TAX_INCLUDE_CONSEQUENCE` is genuinely shared (same JSX node) between the checkbox description and the dialog description — checked this isn't just visually-matching duplicated strings.
- **Access control correct**: the checkbox row is wrapped in the existing `ReadOnlyLock`/`useWriteAccess('analytics:write')` gate, so a demo-mode/read-only user never even reaches a state where the confirm dialog could open. No separate gating needed on the dialog itself.
- **Test coverage is now complete** for the three real branches: ON+confirm (existing tests, updated to click through), ON+cancel (new test — dialog closes, `updateSettings` never called), OFF+direct-write (new test — asserts no dialog renders and the mutation still fires immediately). The e2e restore-in-`finally` fix is correct and actually improves crash-safety over the pre-fixup version.
- Frontend dependency direction is fine — feature component pulling from `shared/ui`/`shared/auth`, nothing backwards.

**Two non-blocking notes, not from this PR's own regression:**
1. `ConfirmDialog`'s `onOpenAutoFocus` defaults focus to the *confirm* button, not Cancel — for a "this changes numbers for every operator on every date range" action, defaulting focus to Cancel is arguably the safer convention. This is inherited from the shared primitive and identical to the pre-existing `confirmingRecalculate` dialog, so it's not something to fix in this PR — flagging in case it's worth a follow-up on `ConfirmDialog` itself (maybe an opt-in `focusOn="cancel"` for higher-stakes confirms).
2. The PR still bundles three genuinely unrelated concerns (the #2857 confirm gate, an unrelated `smart-test.mjs` DX fix, and an unrelated `invoicing-list.int-spec.ts` CI-flake fix). The author already split these into three well-named commits, which mitigates the review-hygiene concern from the earlier round, but per `docs/engineering-standards.md`'s branch-per-issue convention this ideally would have been three PRs. Not asking for a resplit at this point given the size and that it's already reviewed once — just noting for next time.

**Verdict: ✅ Approve.** The confirm-gate logic, state handling, accessibility (via the shared primitive), and test coverage are all sound. The two notes above are pattern/process observations, not defects in this diff.

norbert-kulus-blockydevs:
Addressed the one suggestion from the earlier review:

**Default focus on Confirm instead of Cancel** — fixed at the primitive level rather than at this call site alone. `ConfirmDialog` gains an opt-in `initialFocus?: 'confirm' | 'cancel'` prop, defaulting to `'confirm'` (byte-identical behavior for every other existing consumer). The tax-rate-inclusion confirm now passes `initialFocus="cancel"`, since it's a deployment-wide financial-reporting write triggered from one click — landing on the button that does nothing is the safer default there. Added a test asserting Cancel receives focus when the prop is set, alongside the existing default-focus-on-Confirm test.

The second note (three unrelated commits in one PR) required no action — confirmed via `gh pr view --json commits` that the PR is already split into three cleanly separable commits (the confirm-gate feature, an unrelated `smart-test.mjs` DX fix, and an unrelated CI-flake fix), matching what the prior review found.

piotrswierzy (approved):
## `/pr-review` (re-review) — ✅ Approve, conditional on CI

The force-push addressed **all six** IMPORTANT findings, and — as with #3000 — by taking the offered fixes rather than working around them. One unrequested improvement on top.

| Prior finding | Status |
|---|---|
| e2e header claimed "fully compared" but screenshotted the wrong page | ✅ **Resolved** |
| Mockup stale on four axes | ✅ **Resolved**, both halves of the obligation |
| Copy-drift claim false; the two copies had already drifted | ✅ **Resolved** |
| Docblock called the setting "permanent" | ✅ **Resolved** |
| ON→OFF branch had no test | ✅ **Resolved** |
| PR composition | 🔄 **Partial** |

**The mockup handling is the best part.** Axis by axis it did the harder, more honest thing rather than the convenient one: `alertdialog`→`dialog` corrected **in the mockup** (right call — every `ConfirmDialog` renders through the shared Radix `Dialog`, so the mockup was wrong, not the implementation); the **title moved the implementation to the mockup's copy**, so that axis no longer diverges at all; the warning callout and the affected-counts line are now shipped; and the two divergences that remain legitimately (the illustrative rate list, the `6 products` half) are recorded in an HTML comment in the mockup with the reason. That is exactly what `frontend-architecture.md § UX Mockups` asks for.

`const TAX_INCLUDE_CONSEQUENCE` makes the apostrophe drift structurally impossible — one JSX fragment, one `&rsquo;` — and its docblock names the exact defect. `smart-test.mjs`'s catch fallback is reverted to `'jest'` with a docblock stating a read failure must fail loud. The e2e restore moved into a `finally`. All three actioned.

**Unrequested and correct**: `ConfirmDialog` gains an opt-in `initialFocus?: 'confirm' | 'cancel'` (default `'confirm'`, so no existing consumer changes), used as `initialFocus="cancel"` here, with a test asserting Cancel has focus. Right call for a deployment-wide write, and correctly done as a default-preserving prop on the shared primitive rather than bespoke markup. Minor: the commit attributes it to *"#2993 review SUGGESTION 1"* — SUGGESTION 1 was the inert `isConfirming`, not focus order.

### Two mechanical preconditions before merge

1. **The branch is `BEHIND` main and four jobs have not run** — `Lint`, `Test`, `Integration Tests`, `Build` are QUEUED, not passing. `Test` is the only gate on the new dialog specs. Update from `main` and confirm they go green.
2. **Add one line to the PR body** disclosing the `apps/api/test/integration/invoicing-list.int-spec.ts` (#2986) flake fix — still in the diff, still unmentioned. The change itself remains sound; the disclosure point stands.

### New finding — worth a guard now that it feeds a gate

`Affects {taxA} orders.` renders a positive `0` from an absent read. `taxA` is `coverageQuery.data?.categories.find(…)?.affectedCount ?? 0`, so while the coverage query is loading **or has failed**, the confirm asserts *"Affects 0 orders."* — a positive claim built from an absent value, in the one dialog the operator acts on. `frontend-architecture.md § Paginated Totals` is explicit about this class: *"A failed count leaves the placeholder, and never renders `0`… absence and 'none matched' are different claims."*

The `?? 0` is pre-existing (line 432, since #2668), so not a regression — but promoting it into a confirmation gate is what makes it worth guarding. Render the line only when `coverageQuery.data` is present, or say "an unknown number of orders" on failure. Cheap, and it protects the reason the line was added.

### On the #3130 interaction — no conflict

`ORDER_DATE_CAVEAT` lives in `analytics-date-range-toolbar.tsx` and is **untouched** by this PR on either side; the string does not appear in the `tax-confirm` region of the mockup either (the mockup's only `TODAY` copy is the rate-basis option description and the tax-confirm callout, neither of which is the date-range caveat). **#2993 and #3130 can merge in either order.**

Still unaddressed and acceptable: `isConfirming={updateSettings.isPending}` remains inert (the dialog closes before `isPending` flips). It mirrors `handleRecalculate` exactly, so it's consistent precedent — a shared cleanup of both is its own change.

---

norbert-kulus-blockydevs:
Addressed the new finding from the re-review:

> `Affects {taxA} orders.` renders a positive `0` from an absent read ... Render the line only when `coverageQuery.data` is present, or say 'an unknown number of orders' on failure.

Fixed in 6984e8bdb: the confirm dialog now shows "Checking how many orders are affected…" while `coverageQuery.isLoading`, and "Affects an unknown number of orders — we couldn't load the count." on `coverageQuery.isError`, falling through to the real `Affects {taxA} orders.` line only once the read has actually settled successfully. The pre-existing, unrelated `?? 0` usage on the dialog's non-confirm summary line (since #2668) is untouched. Added test coverage for both new states.

Issue #2103, opened 2026-08-14
[BUG] Integration/inFakt — adapter drops the invoice currency, so a EUR invoice is booked as PLN


norbert-kulus-blockydevs:
## Problem / Context

The neutral invoicing command carries a currency:

```ts
// libs/core/src/invoicing/domain/types/invoicing.types.ts:486
currency: string;   // ISO 4217, single-currency invoice
```

The inFakt request type does not have a currency field at all:

```ts
// libs/integrations/infakt/src/domain/types/infakt.types.ts:50-81
export interface InfaktInvoice { … }   // no currency / kurs field
```

and `infakt-invoicing.adapter.ts` contains **zero** references to `.currency` — the string "currency" appears only inside comments about the groszy integer format (`:232`, `:425`, `:686`).

So `command.currency` is silently dropped. Amounts are sent as plain integers in the minor unit with no currency attached, and inFakt books them in the account's default currency (PLN). A EUR 811.37 invoice is issued as PLN 811.37.

Two things make this worse than an ordinary mapping gap:

- **It is silent.** No log, no validation, no `business_failure`. Nothing distinguishes it from a correct issuance.
- **It propagates.** inFakt relays to KSeF on the seller's behalf (see the Invoicing section of `docs/architecture-overview.md`), so an incorrect fiscal document clears the tax authority. A cleared document is not something OL can quietly re-issue.

Surfaced while reviewing #2050 / ADR-040, whose Non-goals section draws a hard line around fiscal currency handling. Independent of that PR.

## Proposed Solution

1. **Establish the contract.** Determine from inFakt's API documentation how a document declares its currency — most likely an invoice-level currency field plus an optional rate, given that inFakt performs its own conversion server-side. Record the finding in the issue with a doc reference, since the current absence may be a genuine API limitation rather than an OL oversight.
2. **If inFakt supports it** — add the field to `InfaktInvoice` and map `command.currency` onto it. Verify against the sandbox that the issued document actually renders in that currency.
3. **If inFakt does not support it** — reject issuance for any currency other than the account's default with a terminal `business_failure` and an explicit reason, rather than booking in the wrong units. A refused document is recoverable; a cleared wrong one is not.
4. **Cover the sibling write paths** — `issueCorrection` and `markPaid` go through the same async task endpoints and need the same treatment.

[...]

Issue #1763, opened 2026-07-22
[TASK] Integration — Systematic E2E audit of the inFakt integration (completeness + correctness)


norbert-kulus-blockydevs:
## Problem / Context

The inFakt integration (`libs/integrations/infakt/`) has been built up through epic #1279 (spike #1274/#1275 → backend #1280 → webhooks #1281 → FE plugin #1282 → docs/ADR #1307/#1283), followed by a UX wave (#1353 email to buyer, #1354 `invoice_marked_as_paid` webhook, #1355 bulk-issue, #1356 resend-to-KSeF). The last full E2E pass was done on 2026-07-03 on a separate worktree and found two real bugs:
- #1337 - a correction (`kind=corrective`) silently becomes a plain VAT invoice on inFakt's side (`POST /invoices.json` ignores `kind`),
- #1338 - `providerInvoiceNumber` was never persisted, which blocked the precondition for every KOR correction (one-line fix verified live, but current status on `main` is unknown).

Since then, 4 more UX PRs (#1353-#1356) merged in parallel touching shared files (`invoicing.controller.ts`, `libs/core/src/invoicing/domain/ports/capabilities/`, `apps/web/src/features/invoicing/`, `apps/web/src/plugins/infakt/`) - the merged result was never verified end-to-end as a whole. We also don't know whether known gaps (webhook-secret exchange direction never verified live - see PR #1307, `RegulatoryDocumentReader`/UPO not available in the inFakt API, sandbox quirks like the unreliable `invoice_type` filter) are deliberate documented decisions or real functional gaps.

We need a fresh, black-box audit done from the perspective of someone who doesn't know the PR history - to catch regressions introduced by the merges, confirm the core e-commerce scenarios (issuing, corrections, payments, statuses) actually work on current `main`, and write down the real gaps against what an e-commerce shop needs from an accounting integration.

## Proposed Solution

This is an audit/QA task, not an implementation task. The deliverable is a **report written directly into this issue's body** (edited in place, not posted as separate comments) covering: what was tested, what works, what's broken (with repro steps), and what's missing relative to typical e-commerce needs. Any concrete bugs found during the audit get their own separate `[BUG]` issue, linked back from this report - do not fix them as part of this task.

**Scope to verify (on the inFakt sandbox).** Checkboxes are for the auditor to tick off as each item gets tested — not a claim about the outcome; note the actual result (works / broken / missing) in the report text next to or below each item.

1. **Account connection / connection lifecycle**
[...]

PR #2995, opened 2026-09-08
feat(infakt): add per-connection defaultSaleType for non-PL invoice issuance, with a diagnosis hint for the missing-config case

### LINKED ISSUE #2177: [BUG] Integration — inFakt invoice issuance 422s for any non-PL client (missing sale_type)
## Problem / Context

Found while live-testing epic #2154 (ADR-041 sales-document routing) — specifically the "Rest of world" routing branch, which resolves non-Polish orders to an inFakt invoice.

`InfaktInvoicingAdapter.issueInvoice` (`libs/integrations/infakt/src/infrastructure/adapters/infakt-invoicing.adapter.ts`) never sends a `sale_type` field in its `POST invoices.json` payload. For a client whose `country` is `PL`, inFakt defaults this silently and issuance works (confirmed live, multiple times, this session). For a client in any OTHER country, inFakt requires `sale_type` explicitly and rejects the request otherwise:

```
POST /api/v3/invoices.json  (client_id resolves to a DE-country client)
→ HTTP 422
{"error":"Nieprawidłowe parametry. Popraw błędy i spróbuj ponownie.",
 "errors":{"sale_type":["Proszę określić rodzaj sprzedaży."]}}
```
("Please specify the sale type.")

Reproduced directly against inFakt's sandbox API (bypassing our adapter) to confirm root cause and find the fix: adding `"sale_type": "service"` to the same payload succeeds (`HTTP 201`, real invoice `4/08/2026` created). Several other plausible values (`goods`, `product`, `towar`, `usluga`, `mixed`, and case variants) were all rejected with the identical error — `service` (lowercase, exact) is the only value confirmed to work in this pass; there is likely a second value for a goods sale that wasn't discovered (not exhaustively enumerated here).

**Consequence today**: any OL install auto-issuing (or manually issuing) an inFakt invoice for a buyer outside Poland gets a hard 422 on every attempt, surfaced to the operator only as the generic "The invoicing provider rejected the request." (`InvoiceService`'s neutral `provider-rejected` mapping) — no hint that a one-field payload gap is the cause.

## Proposed Solution

1. Send `sale_type` on every `issueInvoice` (and likely `issueCorrection`) call once the client's country is resolved as non-`PL`. The correct rule (goods vs. services) is a real VAT-treatment distinction under EU cross-border rules — do not hardcode `'service'` unconditionally, since that would be factually wrong for a physical-goods sale and could misstate the invoice's VAT treatment. Options, in order of preference:
   - If `IssueInvoiceCommand`'s lines already carry a goods/service distinction (check `OrderLine`/`InvoiceLine` — may not exist today), derive `sale_type` from that.
   - If no such distinction exists on the command today, that's itself a gap worth flagging in this issue's scope — the fix may need to add a neutral field upstream (e.g. on `IssueInvoiceCommand` or resolved from the master product record) before this adapter can set the field correctly for a real mixed catalog.
   - As an interim, conservative default (with an explicit code comment stating the compliance caveat and pointing at this issue), defaulting to `'service'` only for a PL-domestic-shaped catalog is NOT safe to assume — a maintainer decision is needed on whether an interim default is acceptable at all, or whether this should hard-block foreign invoicing until the goods/service signal exists.
2. Enumerate inFakt's actual valid `sale_type` values from their API reference (not fully discovered in this pass) and validate the resolved value before sending it, rather than trusting an assumed string.
3. Add an integration test that issues an invoice for a non-PL client and asserts `sale_type` is present in the request body.

## Classification

**Type**: Integration (bug)
**File(s)**: `libs/integrations/infakt/src/infrastructure/adapters/infakt-invoicing.adapter.ts`

## Dependencies

None known, but see the goods/service-signal gap noted above — may reveal its own prerequisite.

## Assumptions

- `'service'` is a real, valid inFakt `sale_type` value (confirmed live). The correct value for a goods sale was not found in this pass and needs confirming against inFakt's own API docs.

## Acceptance Criteria

- [ ] Issuing an inFakt invoice for a non-`PL` client succeeds with a correctly-classified `sale_type` (not a hardcoded, possibly-wrong default).
- [ ] A maintainer decision is recorded on how `sale_type` (goods vs. services) is sourced — either from an existing signal or a new one.
- [ ] Integration test covers a non-PL client issuance.

### LINKED ISSUE #3031: [TASK] Integration (inFakt) — Confirm 'goods' sale_type value and surface a config hint for missing defaultSaleType
## Problem / Context

Follow-up from #2177 / the review round on PR #2995.

`InfaktSaleTypeValues` (`libs/integrations/infakt/src/domain/types/infakt-connection.types.ts`) currently lists only `'service'`, confirmed live against inFakt's sandbox. Several other plausible spellings (`goods`, `product`, `towar`, `usluga`, `mixed`, case variants) were all live-tested and rejected in that pass, including `'goods'` itself — the correct value for a physical-goods sale was not found and is deliberately NOT included as a placeholder (a value that passes OL's save-time shape validation but then 422s at issuance would convert that guard into a trap).

Two gaps remain open, both out of scope for #2177 / PR #2995 (which is otherwise resolved for the confirmed `'service'` case):

1. inFakt's real `sale_type` value for a physical-goods sale is still unknown.
2. A non-PL inFakt connection with no `defaultSaleType` set gets the generic `"The invoicing provider rejected the request."` (`InvoiceService`'s neutral `provider-rejected` mapping) on every issuance attempt, with no hint that a one-field config gap is the cause.

## Proposed Solution

1. **Find inFakt's confirmed `sale_type` value for a physical-goods sale**, against inFakt's own API reference and/or a live sandbox test. Once confirmed, add it as a second member of `InfaktSaleTypeValues`, with test coverage mirroring the existing `'service'` cases in `infakt-invoicing.adapter.spec.ts` and `infakt-connection-config-shape-validator.adapter.spec.ts`.
2. **Detect the specific 422 shape and surface an actionable hint.** In `InfaktInvoicingAdapter` (or wherever the 422 is currently mapped into `InvoiceService`'s neutral `provider-rejected` failure), detect `errors.sale_type` in the response body and surface a specific hint pointing at `config.defaultSaleType`, instead of falling through to the generic provider-rejected message.

## Classification

**Type**: Integration
**Layer**: Infrastructure (adapter error mapping) / Domain (type vocabulary)
**File(s)**: `libs/integrations/infakt/src/domain/types/infakt-connection.types.ts`, `libs/integrations/infakt/src/infrastructure/adapters/infakt-invoicing.adapter.ts`, `libs/integrations/infakt/src/infrastructure/adapters/infakt-invoicing.adapter.spec.ts`, `libs/integrations/infakt/src/infrastructure/adapters/infakt-connection-config-shape-validator.adapter.spec.ts`

## Dependencies

- None — the confirmed `'service'` case already ships from #2177 / PR #2995.

## Assumptions

- If a live inFakt sandbox cannot be reached to positively confirm the goods `sale_type` value, that half of this issue stays open/deferred rather than guessing a value — per the existing rule that an unconfirmed value would convert OL's save-time shape guard into a 422 trap. The diagnosis-hint half (item 2) does not depend on confirming the value and can ship independently.

## Acceptance Criteria

- [ ] `InfaktSaleTypeValues` gains a confirmed second member for physical-goods sales, only once positively verified against inFakt's API reference or a live sandbox call (not guessed).
- [ ] New value has test coverage mirroring the existing `'service'` cases in both spec files listed above.
- [ ] A 422 response body carrying `errors.sale_type` surfaces a specific, actionable hint (pointing at `config.defaultSaleType`) instead of the generic provider-rejected message.

norbert-kulus-blockydevs:
## Summary

- `InfaktInvoicingAdapter.issueInvoice` never sends `sale_type` on `POST invoices.json`. inFakt silently defaults this for a PL-country client, so PL issuance has always worked — but for any other country inFakt rejects the request with HTTP 422 (`{"errors":{"sale_type":["Proszę określić rodzaj sprzedaży."]}}`).
- Adds a per-connection opt-in `defaultSaleType` config field, mirroring the existing `defaultPaymentMethod` pattern: `InfaktConnectionConfig.defaultSaleType`, wired into the adapter constructor (`this.saleType`), stamped conditionally on both `issueInvoice`'s and `issueCorrection`'s payloads, and validated in `InfaktConnectionConfigShapeValidatorAdapter`. Reachable through the raw config JSON editor only (see below for why).
- **No silent default is introduced.** The field stays `undefined` unless the operator explicitly configures it, and the adapter omits `sale_type` from the payload entirely when unset — so a PL client keeps working exactly as before, and a non-PL client keeps 422ing exactly as before, until the operator opts in. This is deliberate: there is no safe universal default across a possibly-mixed (goods + services) catalog, and guessing wrong misstates the invoice's VAT sale-type classification on a real fiscal document.
- **Also folds in the 422-diagnosis-hint half of the follow-up (#3031, via PR #3129, merged into this branch).** `issueInvoice` now structurally detects inFakt's `{"errors":{"sale_type":[...]}}` 422 shape (key presence only, never the Polish message text) and routes it through a new `sale-classification-required` `InvoiceFailureCode` — so a non-PL connection with no `defaultSaleType` configured now gets an operator-facing hint naming `defaultSaleType` as the field to set, instead of the generic "The invoicing provider rejected the request."

## Design decision (already resolved with the maintainer before implementation)

Per-connection default setting, following the `defaultPaymentMethod` pattern — no new upstream domain field on `InvoiceLine`/`IssueInvoiceCommand` (there is genuinely no goods/service signal anywhere in the product/order domain today, and adding one is out of scope for this fix).

## Review round (piotrswierzy) — addressed

- **`InfaktSaleTypeValues` narrowed to `['service']` only.** `'goods'` was a *confirmed-rejected* inFakt value, not merely unconfirmed — leaving it selectable let it pass OL's save-time shape gate and then 422 at issuance, turning the guard into a trap for exactly the operator this fix is meant to help. Filed [#3031](https://github.com/openlinker-project/openlinker/issues/3031) to track finding the real goods-sale value (and, separately, detecting `errors.sale_type` in a 422 body to give an un-configured connection a specific hint instead of the generic provider-rejected message) — the diagnosis-hint half of that follow-up is now merged into this PR (see above); the goods-sale value itself remains unconfirmed (see Known residual gap below).
- **`sale_type` narrowed from `string` to `InfaktSaleType`** on both `InfaktInvoiceRequest` and `InfaktCorrectiveInvoiceRequest`, matching the config field.
- **Added a non-PL fixture and an issuance test that actually exercises the reported bug.** Every prior case used the PL-hardcoded `buyer()` default, which asserts nothing about a non-PL client (a PL client 422s neither with nor without `sale_type`).
- **Documented, in the README and the config field's docblock, why `defaultSaleType` is deliberately raw-JSON-only** rather than a structured `<select>` beside `defaultPaymentMethod` — the field currently carries a single confirmed option, and it's compliance-sensitive per-catalog VAT classification.
- Updated `InfaktConnectionConfigShapeValidatorAdapter` tests to assert `'goods'` is *rejected* at save time (it previously asserted the opposite).
- Merged `main` into the branch (previously behind).
- **Merged #3129 (closing #3031's diagnosis-hint half) into this branch** — see the "Also folds in" bullet above.

## Known residual gap — NOT resolved by this PR

The correct `sale_type` value for a physical-goods sale was not found in the live reproduction pass and is **not** shipped as a selectable option (see above). Non-PL issuance for a goods-classified sale still 422s, unchanged from before this PR, but now surfaces the specific `sale-classification-required` hint instead of a generic rejection; non-PL issuance for a service-classified sale now succeeds once the operator sets `defaultSaleType: "service"`.

Closes #2177 for the confirmed `'service'` case (the fix this issue asked for: a config-driven `sale_type`, validated, with no unsafe default) and closes #3031 for the diagnosis-hint case. The goods-value gap itself remains genuinely unconfirmed — no sandbox access was available to positively verify a value, and guessing one in would recreate the exact trap this PR closed for `'service'`. A future follow-up should file a fresh issue once the correct value is confirmed.

## Test plan

- [x] `pnpm lint` — passed (exit 0, repo-wide)
- [x] `pnpm type-check` — passed (exit 0, repo-wide, including `libs/integrations/infakt`)
- [x] Scoped + full package unit tests — executed locally (`libs/integrations/infakt`, all suites green, including the review-round additions: the `'goods'`-rejection save-time test, the non-PL-buyer issuance test covering AC-3, and #3129's 422-shape-detection + failure-code classification tests). `pnpm smart-test --no-integration` (the repo's pre-commit hook) also ran and passed against each diff.

### DIFF
--- a/apps/web/src/features/invoicing/api/invoicing.types.ts
+++ b/apps/web/src/features/invoicing/api/invoicing.types.ts
@@ -41,6 +41,9 @@ export type FailureMode = (typeof FailureModeValues)[number];
 export const FailureCodeValues = [
   'buyer-tax-id-invalid',
   'invalid-currency',
+  // A missing per-connection sale classification (goods vs. services) — a
+  // one-field config gap, not an order-data problem (#3031).
+  'sale-classification-required',
   'provider-rejected',
   'transport-timeout',
   'provider-error',
--- a/apps/web/src/features/invoicing/lib/derive-invoice-display.ts
+++ b/apps/web/src/features/invoicing/lib/derive-invoice-display.ts
@@ -70,6 +70,8 @@ const FAILURE_CODE_FALLBACK: Record<FailureCode, string> = {
     'The provider rejected the buyer’s tax ID. Check the tax ID on the order’s customer, then retry. Nothing was issued.',
   'invalid-currency':
     'The document’s settlement currency is missing, malformed, or not accepted for this document. Fix the currency on the source order, then retry. Nothing was issued.',
+  'sale-classification-required':
+    'This connection needs a sale classification (goods vs. services) configured before it can issue invoices. Set it in the connection’s configuration, then retry. Nothing was issued.',
   'provider-rejected':
     'The provider rejected the document. Check the order’s data, then retry. Nothing was issued.',
   'transport-timeout':
--- a/docs/lessons.md
+++ b/docs/lessons.md
@@ -1226,3 +1226,38 @@ where an adjacent context already owns similar vocabulary (`routing`, `fulfillme
 
 **Source**: #2953 (caught by `/pre-implement` before implementation; both collisions were in the
 first draft of the plan).
+
+---
+
+## Format only the touched files with the workspace's own pinned prettier binary — never `npx prettier`
+
+**Context**: #3031 touched ten files. A convenience pass over all of them with `npx prettier
+--write` (intending to just clean up the new hunks) reformatted every one **in full** — hundreds of
+unrelated lines, stripping trailing commas from every multi-line call/array/constructor across
+files nobody was asked to touch.
+
+**Problem**: `npx prettier` resolves (and may fetch) a prettier build outside this pnpm workspace's
+`node_modules`, so it does not reliably pick up the repo's pinned `prettier@3.2.5` or apply
+`.prettierrc` (`trailingComma: "es5"`) the way the workspace's own tooling does. The result reads
+like an intentional whole-file reformat and is easy to miss in a large diff — `git diff --stat`
+before-and-after is the tell (a 12-line intended change became 700+ lines across 5 files). Re-running
+the SAME command with the LOCAL binary (`node_modules/.bin/prettier` / `pnpm exec prettier`)
+reproduced the identical mass-reformat, which ruled out a version mismatch and confirmed the
+project's own quality gate (`pnpm lint` → `.eslintrc.js`'s `extends: ['prettier']` is
+`eslint-config-prettier`, which only *disables* conflicting stylistic ESLint rules — it does **not**
+run `eslint-plugin-prettier`) never enforces prettier formatting as a lint error, and CI has no
+separate `format:check` step either. So running prettier at all here bought nothing the gate needed
+and cost a large, unreviewable diff.
+
+**Rule**: don't run prettier as a formatting pass on a diff unless the task is specifically a
+formatting task. When touching a handful of files, format by hand to match the surrounding style
+(the existing lines are the spec) and verify with `git diff --stat` that only the intended lines
+changed. If prettier must run, target the exact files with the local pinned binary
+(`node_modules/.bin/prettier --config .prettierrc --write <files>`), never `npx prettier`, and
+`git diff --stat` immediately after to catch a whole-file reformat before it's mixed into a commit.
+
+**Applies to**: any edit to existing `*.ts`/`*.tsx` files in this repo, especially a small,
+surgical change to a large pre-existing file.
+
+**Source**: #3031 (caught before commit by comparing `git diff --stat` line counts against the
+size of the intended change).
--- a/libs/core/src/invoicing/application/services/invoice.service.spec.ts
+++ b/libs/core/src/invoicing/application/services/invoice.service.spec.ts
@@ -600,6 +600,34 @@ describe('InvoiceService', () => {
       );
     });
 
+    it("(d5b) failureCode: a 'rejected' throwable naming a missing sale classification maps to 'sale-classification-required' (#3031)", async () => {
+      repo.findByIdempotencyKey.mockResolvedValue(null);
+      repo.create.mockResolvedValue(makeRecord({ id: 'rec-1', status: 'pending' }));
+      // Structural `reason` field — mirrors how an adapter recognises its own
+      // provider's "you must configure this" 422 shape and re-throws with a
+      // reason matching SALE_CLASSIFICATION_REJECTION_MARKERS (#3031, inFakt's
+      // `errors.sale_type` detection), rather than propagating the provider's
+      // raw (possibly buyer-PII-echoing) message.
+      const rejection = Object.assign(new Error('rejected'), {
+        failureMode: 'rejected' as const,
+        reason: 'Infakt requires a sale classification to be configured for this connection',
+      });
+      adapter.issueInvoice.mockRejectedValue(rejection);
+      repo.updateOutcome.mockResolvedValue(makeRecord({ id: 'rec-1', status: 'failed' }));
+
+      await expect(service.issueInvoice(makeCmd())).rejects.toBe(rejection);
+      expect(repo.updateOutcome).toHaveBeenCalledWith(
+        'rec-1',
+        expect.objectContaining({
+          status: 'failed',
+          failureMode: 'rejected',
+          failureCode: 'sale-classification-required',
+          failureReason:
+            "This connection requires a sale classification (goods vs. services) to be configured before it can issue invoices. Set the connection's `defaultSaleType` and re-issue.",
+        }),
+      );
+    });
+
     it("(d5) failureCode: a tax-id rejection that also mentions a currency stays 'buyer-tax-id-invalid'", async () => {
       repo.findByIdempotencyKey.mockResolvedValue(null);
       repo.create.mockResolvedValue(makeRecord({ id: 'rec-1', status: 'pending' }));
--- a/libs/core/src/invoicing/application/services/invoice.service.ts
+++ b/libs/core/src/invoicing/application/services/invoice.service.ts
@@ -68,7 +68,10 @@ import { MissingTaxRateException } from '../../domain/exceptions/missing-tax-rat
 import { CapabilityNotSupportedException } from '@openlinker/core/integrations';
 // Published so an adapter spec can pin its own pre-call refusal message against
 // the very markers this service matches on (#2103 review) — see the constant's doc.
-import { CURRENCY_REJECTION_MARKERS } from '../../domain/types/invoicing.types';
+import {
+  CURRENCY_REJECTION_MARKERS,
+  SALE_CLASSIFICATION_REJECTION_MARKERS,
+} from '../../domain/types/invoicing.types';
 import type {
   CorrectionLine,
   GetInvoiceByOrderQuery,
@@ -982,6 +985,9 @@ export class InvoiceService implements IInvoiceService {
     if (CURRENCY_REJECTION_MARKERS.some((marker) => haystack.includes(marker))) {
       return 'invalid-currency';
     }
+    if (SALE_CLASSIFICATION_REJECTION_MARKERS.some((marker) => haystack.includes(marker))) {
+      return 'sale-classification-required';
+    }
     return 'provider-rejected';
   }
 
@@ -1001,6 +1007,11 @@ export class InvoiceService implements IInvoiceService {
       // route here, which is why it names the condition rather than the actor.
       'invalid-currency':
         'The settlement currency is missing, malformed, or not accepted for this document. Fix the currency on the order and re-issue.',
+      // Names the connection config field an operator can actually set — the
+      // generic 'provider-rejected' copy gives no clue this is a one-field
+      // config gap rather than a per-order problem (#3031).
+      'sale-classification-required':
+        "This connection requires a sale classification (goods vs. services) to be configured before it can issue invoices. Set the connection's `defaultSaleType` and re-issue.",
       'provider-rejected': 'The invoicing provider rejected the request.',
       'transport-timeout':
         'The invoicing request timed out; the document may or may not have been created.',
--- a/libs/core/src/invoicing/application/services/invoice.service.ts
+++ b/libs/core/src/invoicing/application/services/invoice.service.ts
@@ -950,8 +950,10 @@ export class InvoiceService implements IInvoiceService {
    * throwable's `reason`/message — never value-importing an adapter error class.
    *
    *   - `rejected` (TERMINAL): a tax-identifier rejection → `buyer-tax-id-invalid`;
-   *     a settlement-currency rejection → `invalid-currency` (both operator-
-   *     fixable on the source order); anything else → `provider-rejected`.
+   *     a settlement-currency rejection → `invalid-currency`; a missing
+   *     sale-classification rejection → `sale-classification-required` (all
+   *     three operator-fixable on the source order / connection config);
+   *     anything else → `provider-rejected`.
    *   - `in-doubt` (transient/indeterminate transport): `transport-timeout`.
    *
    * Tax id is checked first purely for determinism — a message naming both is
--- a/libs/core/src/invoicing/domain/types/invoicing.types.spec.ts
+++ b/libs/core/src/invoicing/domain/types/invoicing.types.spec.ts
@@ -42,6 +42,7 @@ describe('invoicing.types', () => {
     expect([...InvoiceFailureCodeValues]).toEqual([
       'buyer-tax-id-invalid',
       'invalid-currency',
+      'sale-classification-required',
       'provider-rejected',
       'transport-timeout',
       'provider-error',
--- a/libs/core/src/invoicing/domain/types/invoicing.types.ts
+++ b/libs/core/src/invoicing/domain/types/invoicing.types.ts
@@ -87,6 +87,18 @@ export type InvoiceFailureMode = (typeof InvoiceFailureModeValues)[number];
  *     accurately, since the provider did reject it. Operator-fixable on the
  *     source order either way. Currency and ISO 4217 are country-agnostic
  *     vocabulary, so this stays ADR-026-clean.
+ *   - `sale-classification-required`: a TERMINAL `rejected` failure caused by a
+ *     missing per-connection sale classification (goods vs. services) — a
+ *     required-but-unconfigured PLUGIN SETTING, not an order-data problem
+ *     (#3031). Unlike `invalid-currency` this genuinely IS a provider-side
+ *     rejection (the adapter has no signal to refuse the call pre-emptively —
+ *     #2177/#2995 deliberately did not add a goods/service field to
+ *     `InvoiceLine`), so it can only be detected reactively, by an adapter
+ *     recognising its own provider's shape for "you must configure this" and
+ *     re-throwing with a `reason` one of {@link SALE_CLASSIFICATION_REJECTION_MARKERS}
+ *     matches. The operator-fixable remedy is a one-field connection config
+ *     change, which is why this must not collapse into `provider-rejected` —
+ *     that copy gives no clue there is a config knob to set at all.
  *   - `provider-rejected`: any other TERMINAL `rejected` failure (safe to
  *     re-attempt once the underlying input is corrected).
  *   - `transport-timeout`: an `in-doubt` transport failure — the document MAY
@@ -97,6 +109,7 @@ export type InvoiceFailureMode = (typeof InvoiceFailureModeValues)[number];
 export const InvoiceFailureCodeValues = [
   'buyer-tax-id-invalid',
   'invalid-currency',
+  'sale-classification-required',
   'provider-rejected',
   'transport-timeout',
   'provider-error',
@@ -129,6 +142,34 @@ export const CURRENCY_REJECTION_MARKERS = [
   'currency is required',
   'currency code',
 ] as const;
+
+/**
+ * Substrings (case-insensitive) that mark a `rejected` failure as a missing
+ * per-connection sale-classification (goods vs. services) setting, so
+ * `classifyFailureCode` can resolve `sale-classification-required` instead of
+ * the generic `provider-rejected` (#3031, follow-up to #2177/#2995).
+ *
+ * Unlike the currency markers this is a genuine POST-CALL provider rejection —
+ * core cannot know ahead of time whether the buyer's country needs the field,
+ * and #2995 deliberately declined to add a goods/service signal to
+ * `InvoiceLine` (a national VAT-classification concept ADR-026 keeps out of
+ * `libs/core`). An adapter that recognises its own provider's "you must
+ * configure a sale classification" response re-throws with a `reason`
+ * matching one of these markers instead of propagating the provider's raw
+ * (possibly buyer-PII-echoing) message.
+ *
+ * PUBLISHED for the same reason {@link CURRENCY_REJECTION_MARKERS} is: both
+ * ends of the structural read live in this repo (the reason text an adapter
+ * throws is OL-authored, never a verbatim provider string), so an adapter
+ * spec can assert its own message still matches one of these markers and a
+ * reword breaks the build rather than silently degrading the operator's
+ * failure reason. Neutral vocabulary only — "sale classification" names no
+ * country or tax system (ADR-026); it deliberately avoids inFakt's own wire
+ * field name (`sale_type`) so the marker itself never leaks one provider's
+ * naming into a cross-provider vocabulary, even though inFakt is the only
+ * shipped adapter that triggers it today.
+ */
+export const SALE_CLASSIFICATION_REJECTION_MARKERS = ['sale classification'] as const;
 
 /**
  * Neutral Continuous-Transaction-Controls clearance lifecycle. The adapter maps
--- a/libs/integrations/infakt/README.md
+++ b/libs/integrations/infakt/README.md
@@ -47,7 +47,8 @@ Authentication uses a **static API key** (no OAuth).
     "id": "12345",
     "accountNumber": "PL00 0000 0000 0000 0000 0000 0000",
     "bankName": "mBank"
-  }
+  },
+  "defaultSaleType": "service"
 }
 ```
 
@@ -61,6 +62,8 @@ missing suffix onto the override automatically, but the corrected example
 above should be preferred over relying on that normalization.
 `defaultPaymentMethod` and `bankAccount` are optional (see #1309/#1310 below) -
 omit both to fall back to `cash` with no stamped account.
+`defaultSaleType` is optional (see #2177 below) - omit it to leave `sale_type`
+off the payload entirely, matching pre-#2177 behavior.
 
 ## Notable implementation details
 
@@ -109,6 +112,22 @@ omit both to fall back to `cash` with no stamped account.
   picker; the picked account is snapshotted into `config.bankAccount` and pushed back
   as the inFakt default via `BankAccountDefaultSetter.setDefaultBankAccount()`.
   `transfer` invoices carry the snapshot's `bank_account` / `bank_name` fields.
+- **Per-connection sale type for non-PL clients** (#2177): `config.defaultSaleType`
+  (`goods | service`) is stamped as `sale_type` on every issued invoice/correction
+  when configured. inFakt silently defaults `sale_type` for a PL-country client, so
+  issuance without this field has always worked for PL buyers; for any other
+  country inFakt rejects the request with 422
+  (`{"errors":{"sale_type":["Proszę określić rodzaj sprzedaży."]}}`) unless it is
+  present. There is **no safe universal default** — the field is left unset unless
+  the operator configures it, and unset means `sale_type` is omitted from the
+  payload entirely (PL issuance keeps working; non-PL issuance still 422s exactly
+  as before, unchanged). Only `'service'` is confirmed against inFakt's sandbox
+  (exact lowercase match); the correct value for a physical-goods sale (`'goods'`
+  here is the placement, not a confirmed value) was **not** found and needs
+  confirming separately by an operator with a physical-goods catalog. Picking the
+  wrong value misstates the invoice's VAT sale-type classification, so this is an
+  explicit, compliance-sensitive operator opt-in — OL cannot infer goods vs.
+  services from a possibly-mixed catalog.
 - **Rendered-PDF download** (#1321): `RegulatoryDocumentReader.getRegulatoryDocument
   (record, 'rendered')` fetches the invoice PDF as rendered by inFakt - this backs
   the **Download PDF** button on the accepted invoice detail page.
--- a/libs/integrations/infakt/README.md
+++ b/libs/integrations/infakt/README.md
@@ -63,7 +63,16 @@ above should be preferred over relying on that normalization.
 `defaultPaymentMethod` and `bankAccount` are optional (see #1309/#1310 below) -
 omit both to fall back to `cash` with no stamped account.
 `defaultSaleType` is optional (see #2177 below) - omit it to leave `sale_type`
-off the payload entirely, matching pre-#2177 behavior.
+off the payload entirely, matching pre-#2177 behavior. **Set through the raw
+config JSON editor only** — deliberately not surfaced by either the setup
+wizard or the structured edit-form section, unlike `defaultPaymentMethod`
+right above it. Two reasons, both intentional (#2995 review): the field's
+only confirmed value today is `'service'` (see below), so a `<select>` with
+one usable option buys little discoverability over the raw-JSON path; and the
+value is compliance-sensitive per-catalog VAT classification, which is not a
+knob to expose broadly until there's a real goods/services choice to offer.
+Revisit once a confirmed `'goods'` value lands (see the follow-up referenced
+below).
 
 ## Notable implementation details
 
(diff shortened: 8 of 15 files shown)


piotrswierzy (commented):

Small, well-bounded, correctly-layered. No BLOCKING issues. I ran both touched suites on `pull/2995/head` since the PR's own test box is unchecked — **143 tests, 2 suites, all green**, so that box can be ticked.

### IMPORTANT — `'goods'` is a *confirmed-rejected* value, shipped as a selectable option

`libs/integrations/infakt/src/domain/types/infakt-connection.types.ts` (`InfaktSaleTypeValues`)

The PR frames `'goods'` as "unconfirmed". #2177 is stronger than that — it was **tested live and rejected**:

> Several other plausible values (`goods`, `product`, `towar`, `usluga`, `mixed`, and case variants) were all rejected with the identical error — `service` (lowercase, exact) is the only value confirmed to work.

So an operator with a physical-goods catalog can set `defaultSaleType: "goods"`, **pass OL's save-time gate**, and then hit exactly the opaque 422 that gate exists to prevent. `InfaktConnectionConfigShapeValidatorAdapter`'s own docblock states the purpose: *"a malformed shape must fail fast at save time (400) rather than surface as an opaque inFakt 422 at issuance."* This converts the guard into a trap for precisely the operator the fix is meant to serve.

The "documented placement for a future confirmed value" rationale doesn't hold either, because the union is **closed**: an operator who discovers the real goods value still cannot configure it without a code change. So the placeholder buys nothing and costs a false affordance.

Ship `InfaktSaleTypeValues = ['service'] as const`; add the second member when it is confirmed. The docblock already records the open question.

### IMPORTANT — no structured operator surface, unlike the pattern it mirrors

The PR says it mirrors `defaultPaymentMethod` "exactly", but not on the half an operator touches. That field is wired end-to-end: `apps/web/src/features/connections/components/infakt-setup-form.tsx`, the `infaktPaymentMethod` clause in `edit-connection.schema.ts:785`, and `apps/web/src/plugins/infakt/components/infakt-structured-section.tsx`. `defaultSaleType` has none.

It *is* reachable via the raw-config JSON toggle, so this is not the outright #2610 "implemented, tested and unreachable" class — but it is undiscoverable: nothing connects the generic `provider-rejected` error to this knob. Either add a `<select>` beside payment method, or state in the README that raw-JSON-only is deliberate for a compliance-sensitive field.

### IMPORTANT — `Closes #2177` overstates what lands

Two of the four ACs are unmet and the PR doesn't narrow the claim:

- **AC-1** (*"Issuing an inFakt invoice for a non-PL client succeeds"*) — unmet by construction for a goods catalog, per the finding above.
- **AC-3** (*"Integration test covers a non-PL client issuance"*) — unmet: every new case uses the file's `buyer()` fixture, which hardcodes `countryIso2: 'PL'`. **Nothing in the suite exercises the non-PL case that is the entire subject of the issue.** Since the implementation is deliberately country-independent, a non-PL fixture is cheap and would make the tests assert the reported bug rather than a proxy for it.

Either satisfy those, or narrow the claim and open a follow-up.

### SUGGESTIONS

- `sale_type?: string` on `InfaktInvoiceRequest` / `InfaktCorrectiveInvoiceRequest` (`infakt.types.ts` ~245, ~307) is a bare string, so nothing type-level stops a future caller stamping an unvalidated value onto a fiscal payload. `InfaktSaleType` would narrow it for free. (Only a suggestion — `bank_account` beside it is also plain, so the file is internally consistent.)
- AC-2 asks for a **recorded maintainer decision**. It's in the PR body and code comments; the issue thread has only the branch-claim bot comment. Since merging auto-closes #2177, drop a comment there so the rationale is findable from the issue.
- #2177 names the diagnosis gap as half the problem — an un-opted-in connection still shows the identical uninformative *"The invoicing provider rejected the request."* Detecting `errors.sale_type` in the 422 body and annotating a specific hint is a small, well-scoped follow-up worth filing.

### Done well

- **The "no silent default" claim is real, not asserted.** `...(this.saleType ? { sale_type: this.saleType } : {})` on both paths, plus `this.saleType = config.defaultSaleType` with no `??` fallback, genuinely preserves pre-#2177 behaviour. The regression guards use `expect('sale_type' in body.invoice).toBe(false)` against a call looked up by method+path — so they throw rather than pass vacuously if the request was never made. That is the right shape for a guard.
- **The harder call was made correctly.** Adding a goods/service signal to `InvoiceLine` / `IssueInvoiceCommand` would put a national VAT-classification concern into `libs/core`, which ADR-026 forbids. Keeping it as per-connection plugin config is the right refusal.
- **The correction path is covered** — stamping the same value on `issueCorrection`, so a correction cannot disagree with its original's classification. The issue only flagged that as "likely"; it's tested on both arms.
- No FE data loss: `mergeStructuredIntoConfig` patches a *copy* of the existing config, so editing through the structured form won't drop a raw-JSON-set `defaultSaleType`.

Branch is `BEHIND` main — worth updating before merge.

---

norbert-kulus-blockydevs:
Thanks for the thorough review — addressed all of it (IMPORTANTs and suggestions) in b79ee81b:

**IMPORTANT — `'goods'` shipped as a selectable-but-rejected value**
Fixed as you described: `InfaktSaleTypeValues` is now `['service'] as const`. The docblock previously framed `'goods'` as "unconfirmed"; it's rewritten to say what it actually was — confirmed-rejected — and to state that a placeholder that passes the save-time gate and 422s at issuance is worse than not having the field at all. Also updated `InfaktConnectionConfigShapeValidatorAdapter`'s tests, which previously asserted `'goods'` *resolves* — they now assert it's rejected, plus a dedicated test naming it as the confirmed-rejected value rather than just "not a supported value".

**IMPORTANT — no structured operator surface**
Went with your second option: documented in the README and on the config field's docblock why `defaultSaleType` stays raw-JSON-only for now — a `<select>` with a single confirmed option buys little over the raw-JSON path, and the value is compliance-sensitive per-catalog VAT classification I'd rather not surface broadly before there's a real choice to offer. Revisit once a second confirmed value lands.

**IMPORTANT — `Closes #2177` overstated what lands**
- AC-3: added a non-PL buyer fixture (`countryIso2: 'DE'`) and a new issuance test that actually exercises the reported bug — every prior case used the PL-hardcoded `buyer()` default, which asserts nothing about a non-PL client.
- AC-1: narrowed the PR body's claim rather than pretending it's fully met — this closes #2177 for the confirmed `'service'` case (config-driven, validated, no unsafe default, which is what the issue asked for), and non-PL goods-catalog issuance stays a known, tracked gap since the correct value isn't known. Filed [#3031](https://github.com/openlinker-project/openlinker/issues/3031) to track both finding the real goods value and the opaque-422 diagnosis improvement you flagged as a good follow-up.

**Suggestions**
- `sale_type` on `InfaktInvoiceRequest`/`InfaktCorrectiveInvoiceRequest` narrowed from `string` to `InfaktSaleType`.
- AC-2: this comment is that record — the maintainer decision (per-connection config, no upstream domain field) is now findable from the issue itself, not just the PR body/code comments.
- Also merged `main` in, since the branch was behind.

Test plan updated too — ran the full `libs/integrations/infakt` suite locally this round (12 suites / 261 tests, all green, including the new cases) plus `pnpm lint` / `pnpm type-check` repo-wide and the `smart-test` pre-commit gate, all clean.

norbert-kulus-blockydevs:
Merged PR #3129 into this branch — adds the 422-diagnosis-hint half of the #3031 follow-up (detects inFakt's `{"errors":{"sale_type":[...]}}` shape and surfaces a specific `defaultSaleType`-pointing hint instead of the generic provider-rejected message). Title and description updated to reflect this. #3031 is now closed; the confirmed goods-sale `sale_type` value itself is still unresolved and needs a follow-up once someone has live inFakt sandbox access to verify it.

norbert-kulus-blockydevs:
# Code Review: `defaultSaleType` for non-PL inFakt invoice issuance + 422 diagnosis hint (#2177 / #3031)

**Scope reviewed**: all 15 changed files (adapter, config validator, domain types, core `InvoiceService`, FE failure-code copy, README, tests, `docs/lessons.md`).

---

## 1️⃣ Summary

This PR fixes a real production bug: `InfaktInvoicingAdapter` never sent `sale_type` on invoice creation, which inFakt silently defaults for PL buyers but rejects with a 422 for every other country. The fix adds an **opt-in, per-connection** `defaultSaleType` config field (closed union, currently `['service']` only), stamps it conditionally on both `issueInvoice` and `issueCorrection`, and — folded in from the already-merged #3129 — detects inFakt's `errors.sale_type` 422 shape structurally and routes it to a new, specific `InvoiceFailureCode` (`sale-classification-required`) instead of the generic `provider-rejected`.

**Overall quality is high.** This is exactly the kind of fiscal-document change this codebase is strict about, and the PR shows real discipline: no silent default, a closed enum deliberately narrowed to exclude an *unconfirmed and confirmed-rejected* value rather than guessing, PII-safe failure messaging, and a genuine regression test that exercises the actual reported bug (non-PL buyer) rather than the pre-existing PL-only fixture that would have passed either way.

I found **no blocking issues**. A couple of small, non-blocking observations below.

---

## 2️⃣ Alignment With Documentation / Architecture

### ✅ Aligned
- **ADR-026 (country-agnostic invoicing domain) respected.** `InfaktSaleTypeValues` / `sale_type` stay entirely inside `libs/integrations/infakt`; the only thing that crosses into `libs/core` is a neutral, provider-agnostic marker string (`'sale classification'`), which deliberately avoids inFakt's own wire field name (`sale_type`) per the docblock. This mirrors the existing `CURRENCY_REJECTION_MARKERS` (#2103) pattern exactly — same shape, same rationale, same "published so an adapter spec can pin its own message against it" mechanism (`libs/core/src/invoicing/domain/types/invoicing.types.ts:196-221`).
- **No silent default for a compliance-sensitive field.** `sale_type` is omitted from the payload entirely when unconfigured (`...(this.saleType ? { sale_type: this.saleType } : {})`), so PL behavior is provably unchanged and non-PL behavior degrades no worse than before. This matches the repo's established discipline around fiscal fields (`stockSafetyBuffer`/`pricingRule` "no silent default", the tax-rate-enforcement principle in `docs/architecture-overview.md`).
- **Closed-union discipline held under pressure.** The review round correctly caught that a placeholder `'goods'` value would pass the save-time shape gate (400) and then 422 at issuance — converting the very guard meant to protect operators into a trap. Narrowing `InfaktSaleTypeValues` to `['service']` only, and adding an explicit test asserting `'goods'` is *rejected*, is the right call and is exactly the pattern this codebase uses elsewhere (e.g. offer/param validation guards documented under "a mirror stricter than the gate...").
- **Fail-fast config validation.** `InfaktConnectionConfigShapeValidatorAdapter` validates `defaultSaleType` against the closed union at save time (400) rather than letting a malformed value surface as an opaque 422 at issuance — consistent with the existing `bankAccount`/`defaultPaymentMethod` checks in the same file.
- **PII discipline maintained.** `InfaktApiError.reason` and the `SALE_TYPE_HINT_REASON` constant are OL-authored fixed strings — the adapter never propagates inFakt's raw response body (which could echo buyer data) into the classification path or into `reason`. `InvoiceService.deriveFailureReason` continues to map `failureCode` to a bounded, constant, PII-free string rather than deriving copy from the underlying error.
- **Structural, not textual, 422 detection.** `isMissingSaleTypeRejection` checks `errors.sale_type` key *presence* only — never the Polish message text or array contents — so a locale change on inFakt's side can't silently break detection. Good defensive coding.
- **TypeScript exhaustiveness preserved.** `FAILURE_CODE_FALLBACK: Record<FailureCode, string>` in the FE and the `InvoiceFailureCodeValues` array/spec both force every consumer of the union to handle the new code — the compiler enforces completeness rather than relying on someone to remember. I checked and there's only one FE consumer of `FailureCode` (`apps/web/src/features/invoicing/`), so nothing else needed updating.
- **Correct classification ordering.** `classifyFailureCode` checks tax-id → currency → sale-classification → provider-rejected. The new marker (`'sale classification'`) has no overlap risk with the existing tax-id/currency markers, so insertion order is safe.
- **No schema/migration needed** — correctly scoped as a config-shape + adapter-behavior change only, no ORM entity touched.
- **`docs/lessons.md` entry** follows the file's own documented format (Context/Problem/Rule/Applies to/Source) and captures a genuinely useful process lesson (don't run `npx prettier` as an ad-hoc formatting pass — it silently reformatted whole files outside the intended hunks). Good hygiene per the project's CLAUDE.md mandate to graduate lessons from real incidents.
- **Testing standards followed**: unit tests only (appropriately — no integration-test-worthy vertical slice here), AAA structure, descriptive names, ports/fakes mocked (`FakeInfaktHttpClient`) rather than real HTTP. The **AC-3 test is the standout** — it's the first case in the suite to actually use a non-PL buyer, correctly calling out that every prior case used the PL-hardcoded `buyer()` default and therefore asserted nothing about the reported bug (a PL client 422s neither with nor without `sale_type`). That's the right root-cause instinct for a bug-fix PR.

### Minor observations (non-blocking)

🟢 **`classifyFailureCode`'s docblock wasn't updated to mention the third branch.** The private method's JSDoc in `invoice.service.ts` still reads *"a tax-identifier rejection → `buyer-tax-id-invalid`; a settlement-currency rejection → `invalid-currency`... anything else → `provider-rejected`"* — it doesn't mention the new `sale-classification-required` branch that now sits between currency and the catch-all. Small doc-completeness gap; worth a one-line addition since this file is clearly held to a high documentation bar elsewhere in the same PR.
- `libs/core/src/invoicing/application/services/invoice.service.ts` (docblock above `classifyFailureCode`, ~line 946)

🟢 **Corrections read the *current* connection config, not a snapshot of what was stamped on the original invoice** — so if an operator changes `defaultSaleType` between issuing an invoice and later correcting it, the correction could carry a different `sale_type` than the original, which is exactly the disagreement the adapter's own comment says it wants to avoid ("*A correction should not disagree with the original invoice's sale classification*"). I checked: this is **not a regression introduced by this PR** — `bankAccountFields()` has the identical "read from live config" behavior for corrections already, so this PR is consistent with existing adapter precedent, not introducing a new class of bug. Worth a tracking note (or just leaving as-is, given it mirrors `bankAccountFields()`), not a blocker for this fix.
- `libs/integrations/infakt/src/infrastructure/adapters/infakt-invoicing.adapter.ts:900-903` (`issueCorrection`)

🟢 **Applying `sale_type` uniformly (including to PL clients) is a deliberate, documented simplicity trade-off** consistent with how `defaultPaymentMethod` already behaves. Worth flagging only because it has a slightly different blast radius than `defaultPaymentMethod`: on a connection that mixes PL clients selling goods *and* needs `defaultSaleType: 'service'` set to unblock a non-PL service client, every PL invoice will now also carry `sale_type: 'service'` explicitly — where before inFakt's own server-side inference (which may have correctly inferred "goods" for those PL sales) governed. The README documents this is compliance-sensitive and operator-driven, which is the right call given there's no per-catalog signal to route on, but a one-sentence callout of this specific mixed-catalog interaction in the README would make the operator-facing risk more concrete.

---

## 3️⃣ Security Review

- ✅ No secrets, credentials, or API keys in code or config schema — `defaultSaleType` is an ordinary enum value, resolved through the existing `credentialsRef` boundary elsewhere.
- ✅ No injection surface — no raw SQL, no string-built HTTP bodies beyond typed JSON payloads sent through the existing `IInfaktHttpClient` port.
- ✅ No PII leak: the new `reason` field on `InfaktApiError` is optional and, per its own docblock, is explicitly reserved for OL-authored phrases only — never the raw provider response, which is exactly the constraint that matters here since `responseBody` can (in general) echo buyer-identifying data back from inFakt.
- ✅ FE failure-code copy (`FAILURE_CODE_FALLBACK`) is a fixed constant per code, not derived from the underlying provider error — consistent with the "failureReason must never leak PII" invariant already established in `InvoiceService.deriveFailureReason`.
- ✅ Enum validated both at save time (400, `ConnectionConfigShapeValidatorPort`) and structurally in the type system before being stamped on the fiscal payload — an attacker (or a careless direct-API caller bypassing the FE) cannot get an arbitrary string onto `sale_type` via config; `InvalidConnectionConfigException` catches it first.
- No authorization-relevant code paths touched (no new endpoints, no role/guard changes).

No security findings.

---

## 4️⃣ E-commerce / Fiscal-Document Pattern Review

This is the area I looked at hardest, given the "ecommerce patterns" focus and this repo's extensive fiscal-safety culture around invoicing (ADR-026, ADR-063, the whole `InvoiceFailureCode`/`RegulatoryStatus` design).

- **Correct fiscal-safety instinct**: refusing to guess the `'goods'` value rather than shipping a plausible-looking but unverified one is the single most important decision in this PR, and it's exactly right — a wrong VAT sale-type classification on an issued fiscal document is a real compliance problem, not a cosmetic bug, and this codebase treats it that way everywhere else (see the extensive tax-rate-enforcement and per-line tax-rate provenance sections in `docs/architecture-overview.md`).
- **`failureMode` vs `failureCode` split respected**: the new code stays a `rejected`/terminal classification (safe to retry once config is fixed), never accidentally routed through the `in-doubt` path — I verified `InfaktApiError`'s `failureMode` is recomputed correctly in the constructor from the re-thrown error's preserved `statusCode` (422 stays `rejected`).
- **Correction-consistency gap noted above** is the one genuine ecommerce-domain nuance worth a maintainer's eyes, even though it's pre-existing behavior shared with `bankAccountFields()` rather than new in this PR.
- **Scope discipline**: correctly declining to add a goods/service signal to `InvoiceLine`/`IssueInvoiceCommand` (a cross-cutting domain change) in favor of a per-connection operator opt-in, matching the `defaultPaymentMethod` precedent and avoiding scope creep into core for a single-provider quirk.

---

## 5️⃣ Test Coverage & Quality

**Unit tests**: ✅ strong.
- Config validator: absent / `'service'` / `null` / rejected-string / **rejected `'goods'`** (explicitly testing the review-round fix, not just the happy path) / non-string type coercion.
- Adapter: regression guard for the omit-when-unset case (both `issueInvoice` and `issueCorrection`), positive stamping case, the AC-3 non-PL-buyer success case, and three 422-detection cases (positive match, unrelated 422 propagates untouched, non-object body propagates defensively).
- Core `InvoiceService`: one well-targeted spec asserting the new marker → failure code → `failureReason` chain end-to-end.
- FE: type union + fallback-copy record updated; no dangling exhaustiveness gaps found by search.

**Integration tests**: not applicable — this is adapter/unit-level behavior with no new vertical slice, correctly scoped per `docs/testing-guide.md`.

No test-quality issues found.

---

## Final Verdict

**Status**: ✅ **Approve**

No blocking issues. The two 🟢 suggestions above (docblock completeness, and the pre-existing correction/config-drift observation shared with `bankAccountFields()`) are optional and can be addressed in this PR or a fast follow-up — neither should hold up merge.

**Priority fixes**: none required.

**Optional follow-ups**:
1. Add the `sale-classification-required` branch to `classifyFailureCode`'s docblock enumeration in `invoice.service.ts`.
2. Consider (separately from this PR, and only if it matters in practice) whether `issueCorrection` should snapshot the `sale_type`/`bank_account` actually used at issuance rather than reading live connection config — shared with existing `bankAccountFields()` behavior, not new here.
3. When the confirmed `'goods'` value is found (tracked as #3031's remaining half), make sure the config-validator "reject known-bad values" test for `'goods'` is updated/removed alongside widening `InfaktSaleTypeValues`.

---
*Reviewed with `/pr-review` against `docs/architecture-overview.md`, `docs/engineering-standards.md`, `docs/code-review-guide.md`, and `docs/testing-guide.md`.*

norbert-kulus-blockydevs:
Addressed both optional-follow-up suggestions from the review (plus one extra correction found while doing so):

1. **`classifyFailureCode` docblock updated** to enumerate the `sale-classification-required` branch that was already wired but undocumented.
2. **Mixed-catalog callout added to the README** — a connection selling both goods and services to PL clients that also sets `defaultSaleType` for a non-PL service client will stamp that value on *every* PL invoice too, overriding inFakt's own server-side inference.
3. **Bonus fix**: the README's line about "finding the confirmed goods-sale value" claimed it was "tracked as #3031" — but #3031 turned out to be the diagnosis-hint feature (already merged via #3129), a different task. Corrected to state the goods-sale lookup is explicitly untracked rather than pointing at a closed, unrelated issue.
4. Also left a one-line note beside `issueCorrection`'s `sale_type` spread flagging that it reads live connection config rather than an issuance-time snapshot (pre-existing behavior shared with `bankAccountFields()`, not a regression — matches the review's third observation).

Item 3 from the "optional follow-ups" list (confirming the real `'goods'` value against a live sandbox) is out of scope here — no separate tracking issue currently exists for it, noted as such in the README rather than guessed at.

piotrswierzy (approved):

All three IMPORTANT findings resolved **on the substance**, not paraphrased away; two of three suggestions implemented, and the third folded an entire follow-up PR (#3129) in. Head `126dc8fd`, now 15 files.

| Prior finding | Status |
|---|---|
| `'goods'` shipped as a confirmed-rejected, validator-accepted value | ✅ **Resolved** |
| No structured operator surface | ✅ **Resolved** (second of the two options offered) |
| `Closes #2177` overstates | ✅ **Resolved** |

**On `'goods'`** — `InfaktSaleTypeValues = ['service'] as const`. The docblock now states it was *itself rejected* live (not "unconfirmed") and gives the closed-union argument as the reason not to placeholder it. The validator spec was **inverted correctly**: there is now an explicit `should reject when defaultSaleType is the confirmed-rejected value "goods"` case, and the old accepting case is gone. `it.each(['service'])` means adding the second member later is a one-token edit that carries its own coverage.

**On the operator surface** — no `<select>`; instead raw-JSON-only is now stated as deliberate in **both** the README and the config docblock, with the reason. That was one of the two acceptable outcomes, and it is materially stronger than when I raised it, because the finding's actual complaint — *"nothing connects the generic `provider-rejected` error to this knob"* — is closed by the merged #3129: the failure copy now names the field.

**On the issue claim** — AC-3 is now genuinely met (`buyer()` takes a `countryIso2` override, `nonPlInvoiceCmd` uses `DE`, and a new case exercises it, with a comment recording *why* the PL default asserted nothing). AC-1 is honestly narrowed rather than claimed, via a "Known residual gap — NOT resolved by this PR" section. Notably the README **self-corrected a stale claim** in the final commit — it had pointed the goods-value gap at "#3031", discovered #3031 was the *diagnosis* work, and now marks it explicitly **untracked** rather than pointing at a closed unrelated issue. Right call over a comfortable-looking false pointer.

### On the folded-in #3129 code — reviewed as new, and it is the strongest part

Three properties are load-bearing and correct:

- **Detection is structural, not textual.** `isMissingSaleTypeRejection` tests `'sale_type' in errors` and nothing else — never the Polish message, never array emptiness — with the reason stated (a locale change would silently kill a text match; a present-but-empty array is still inFakt naming the field). Two defensive tests cover a non-object body and an unrelated `errors.client_id`, and both assert `expect(error).toBe(original)` — **identity**, so a wrongly-widened predicate fails rather than passing vacuously.
- **The reason↔marker coupling is asserted, not asserted-in-a-comment.** The adapter's `SALE_TYPE_HINT_REASON` is checked against core's imported `SALE_CLASSIFICATION_REJECTION_MARKERS`, so a reword on either side breaks the build — the #2103 `CURRENCY_REJECTION_MARKERS` precedent applied correctly. The marker deliberately says `sale classification` rather than `sale_type` so one provider's wire naming stays out of a cross-provider vocabulary (ADR-026-clean).
- **The correction path's exemption is argued, not forgotten.** `postInvoiceWithSaleTypeHint` wraps only the direct `invoices.json` POST, with the docblock explaining that a correction fails through `awaitCorrectionTask`'s `{processing_code, processing_description}` envelope (#1763), which carries no field-level `errors` object — so there is no shape to detect. Stating an inapplicability beats silently covering one arm.

Two unrequested additions in the final commit that are right: the **mixed-catalog callout** (a connection selling both goods and services to PL clients will stamp `defaultSaleType` on every PL invoice too, overriding inFakt's own inference), and a note that `issueCorrection` reads **live** connection config rather than an issuance-time snapshot — correctly attributed as pre-existing `bankAccountFields()` behaviour rather than a regression.

### One new SUGGESTION

`libs/core/src/invoicing/application/services/invoice.service.ts:1016` — the operator copy reads *"Set the connection's \`defaultSaleType\` and re-issue."* That is one plugin's config field name hardcoded in `libs/core`. The marker constant twenty lines away goes out of its way to avoid `sale_type` so *"the marker itself never leaks one provider's naming into a cross-provider vocabulary"*; the copy then leaks a different inFakt-specific identifier. The FE's fallback for the same code already says the generic *"Set it in the connection's configuration"*, so the two halves disagree in specificity — and a second provider triggering this code would tell its operator to set a field it does not have. Generalise the core string, or record why naming one provider's field is worth it here.

Also taken: `sale_type` narrowed from bare `string` to `InfaktSaleType` on both request types. Layering unchanged and correct — the national VAT-classification concept stayed in the plugin; core gained only a neutral failure code and a neutral marker.

*Caveat: I did not re-run the suites on `126dc8fd`. My earlier run on this branch was 143 tests / 2 suites green; the diff has since grown by ~200 test lines across 4 suites plus the core-service spec. Let CI be the gate.*

---

norbert-kulus-blockydevs:
Addressed the one remaining open SUGGESTION from the re-review (commit `757d9e6b9`):

**The issue**: `InvoiceService.deriveFailureReason`'s `sale-classification-required` copy read *"Set the connection's `defaultSaleType` and re-issue"* — hardcoding inFakt's own connection-config field name inside `libs/core`. That's exactly what `SALE_CLASSIFICATION_REJECTION_MARKERS` (twenty lines away) goes out of its way to avoid — its own docblock states the marker "deliberately avoids inFakt's own wire field name (`sale_type`) so the marker itself never leaks one provider's naming into a cross-provider vocabulary." The remediation string then did exactly that with a different inFakt-specific identifier (ADR-026 violation).

**Investigation before fixing**: checked whether an adapter-supplied-remediation-copy mechanism exists anywhere in the repo (the way `SALE_TYPE_HINT_REASON` → marker matching works) — none does. Also checked how the FE actually renders this: `resolveFailureCopy` in `derive-invoice-display.ts` never reads the backend's `failureReason` string at all for a known `failureCode` — it resolves purely from its own `FAILURE_CODE_FALLBACK` record, already generic ("Set it in the connection's configuration, then retry."). So the backend string with the field-name leak had no shipped consumer, and the two copies were already diverging in specificity with nothing reconciling them.

**Fix (option a from the review)**: generalized the core string to match the FE's own existing tone/wording exactly, added a comment at the site explaining why it's provider-neutral and pointing at the FE fallback it mirrors, and updated the one unit-test assertion (`invoice.service.spec.ts`, case d5b) that pinned the old string.

Targeted ESLint on both touched files is clean; the full monorepo lint OOM'd on this machine (unrelated resource limit, not a code issue) so I verified the two files directly instead.

Issue #2465, opened 2026-08-25
[TASK] CORE — Tax-rate drill-down + A/B/C split (reusing `netExcludedCount`/`netExcludedValue`)


norbert-kulus-blockydevs:
### Problem / Context

`netExcludedCount`/`netExcludedValue` already report the population `netSalesOrderNetEligibleSql` excludes (pre-rollout era, or a line with an unresolvable rate). What's missing is splitting that single number into the three operator-actionable sub-cases the mockup distinguishes, because they have different remediation paths (Phase 1 Task 1.3's setting only helps category A/C, never B):

- **A — unconfirmed-but-resolvable**: `taxRateEra = 'pre-rollout'` AND a backfilled rate exists (would become net-eligible if Phase 1 Task 1.3's setting were ON) → mockup `detail-tax`.
- **B — no tax rate at all**: excluded, and no backfill candidate resolves from the current catalog either → mockup `detail-novat`.
- **C — pre-rollout, backfill not yet run**: `taxRateEra = 'pre-rollout'` and `TaxRateBackfillService` has not yet attempted this order → mockup `detail-postrollout`.

### Proposed Solution

A read-only extension that takes the existing `netExcludedCount` population and, for each row, calls `TaxRateBackfillService`'s existing rate-resolution logic (read-only — do not trigger an actual backfill write here) to classify it into A, B, or C. Emits `{ category: 'tax-a' | 'tax-b' | 'tax-c', ... }` per Phase 1 Task 1.2's category vocabulary.

### Classification

**Type**: CORE
**Layer**: Domain + Application
**File(s)**: `libs/core/src/orders/application/services/tax-rate-backfill.service.ts` (read-only extension or a sibling read service consuming it), `libs/core/src/orders/domain/types/coverage-detection.types.ts`

### Dependencies

- Task 4.1 (shares the `coverage-detection.types.ts` file and the overall `GET /analytics/coverage` shape).

### Assumptions

- Category naming (`tax-a`/`tax-b`/`tax-c` vs. a more descriptive slug) is pinned by whatever Phase 1 Task 1.2 decided — this task must not invent its own naming if the decision doc already named them.

### Acceptance Criteria

- [ ] `affectedCount` summed across A+B+C matches the existing `netExcludedCount` field exactly (regression guard).
- [ ] Category A correctly flags the manually-fixed live-demo case (31 `taxRateEra='pre-rollout'` rows that had a resolvable backfill rate).
[...]

PR #3000, opened 2026-09-08
perf(orders): bound tax coverage candidate fetch

### LINKED ISSUE #2834: [TECH-DEBT] Orders — page/bound findNetExcludedOrderCandidates's base fetch
## Problem / Context

Follow-up to #2826, split out during review of PR #2827.

PR #2827 closed two of #2826's three acceptance criteria: `TaxCoverageDetectionService.classify` now batches the per-order line-item read (`findByOrderIds`) and resolves catalogue tax rates with bounded concurrency instead of one sequential call per unresolved line. It also dropped the full `OrderRecordOrmEntity` hydration (`orderSnapshot` JSONB included) in favor of a narrow four-column `getRawMany` projection.

What it did NOT do: `OrderRecordRepository.findNetExcludedOrderCandidates` (`libs/core/src/orders/infrastructure/persistence/repositories/order-record.repository.ts`) still issues one **unbounded** query — no `LIMIT`/`OFFSET`, no cap — over the full `SalesAnalyticsFilters` window (currently up to 400 days). Row *count* returned still scales linearly with total non-cancelled, net-excluded order history for that window, even though each row is now cheap to hydrate.

`TaxCoverageDetectionService.classify` (`libs/core/src/orders/application/services/tax-coverage-detection.service.ts`) then classifies the ENTIRE unbounded set in one pass, in memory, before any pagination happens — `getCategoryPage`/`getAllCategoryPages` slice the already-classified array. So `GET /analytics/coverage` (hit unconditionally on `/analytics` page mount) still does one full-history scan and one full in-memory classification per request.

## Proposed Solution

Per #2826's own "Assumptions" section, no specific shape is mandated — either:
- Cap/page `findNetExcludedOrderCandidates` and make `classify()` incremental (accumulate per-category counts across pages instead of one big in-memory array), or
- Otherwise ensure the per-request cost of `GET /analytics/coverage` no longer scales linearly with total order history for the filter window (e.g. a bounded sample plus a separate SQL `COUNT` for the aggregate figures, if exact per-order pagination turns out to conflict with the live-catalogue-check step).

Whichever shape is chosen must preserve the `candidates.length === netExcludedCount` regression guard the current unit/int-specs assert (or replace it with an equivalent invariant over the new shape).

## Classification

**Type**: CORE
**Layer**: Application / Infrastructure
**File(s)**:
- `libs/core/src/orders/infrastructure/persistence/repositories/order-record.repository.ts` (`findNetExcludedOrderCandidates`)
- `libs/core/src/orders/application/services/tax-coverage-detection.service.ts` (`classify`, `getCategoryPage`, `getAllCategoryPages`, `getCategoryCounts`)
- `libs/core/src/orders/domain/ports/order-record-repository.port.ts` (port contract, if the pagination shape changes)

## Dependencies

- Follows PR #2827 (which closed #2826's other two criteria).

## Acceptance Criteria

- [ ] `findNetExcludedOrderCandidates`'s per-request cost (rows scanned/returned) is bounded, or `classify()` no longer requires the full unbounded set materialized in memory before any category count/page can be produced.
- [ ] `GET /analytics/coverage` response (category counts, category pages) is unchanged for identical inputs — no regression in which orders land in `tax-a` / `tax-b` / `tax-c`.
- [ ] Tests added or updated for the new bounded/incremental shape, including a live-Postgres int-spec assertion (mirroring `tax-coverage-net-excluded.int-spec.ts`).

jakubretajczykBD:
## Summary

- `OrderRecordRepository.findNetExcludedOrderCandidates` issued one **unbounded** SQL read over the full net-excluded population for a filter window (up to 400 days); replaced with `findNetExcludedOrderCandidatesPage`, a keyset-paginated read (`ORDER BY placedAt DESC, internalOrderId DESC`, page size 500) bounded per call.
- `TaxCoverageDetectionService.classify()` no longer materializes the whole candidate array before classifying — it drives the new paginated read in a loop and accumulates each page's classification (via the extracted `classifyBatch`) into the running result. Per-order classification never depends on another order's state, so batch-then-append is provably equivalent to the old single-pass classification.
- No behavior change for `GET /analytics/coverage` — category counts/pages are identical for identical inputs.

## Related issues

Closes #2834

## Test plan

- `pnpm --filter @openlinker/core test tax-coverage-detection order-record.repository` — 200/200 unit tests pass, including new multi-page pagination tests (cursor propagation, loop termination on `nextCursor: null`/empty page) and a batch-vs-single-page equivalence test.
- `pnpm --filter @openlinker/api test:integration tax-coverage-net-excluded` — extended `tax-coverage-net-excluded.int-spec.ts` with real-Postgres cases asserting: no single page exceeds the requested limit, a population spanning several pages (including rows sharing an identical `placedAt`, exercising the `internalOrderId` tiebreak) is traversed with no duplicates/no gaps, and the exact 2×limit peek-boundary case yields exactly two pages.
- `pnpm --filter @openlinker/core type-check` — clean.

## Quality gate

- [x] `pnpm lint` passes (zero errors)
- [x] `pnpm type-check` passes (zero errors)
- [x] `pnpm test` passes (all unit tests green)
- [x] `pnpm test:integration` passes — confirmed green on CI (Integration Tests check, 8/8).

## Migrations

- [x] No ORM entity changed — this box is trivially satisfied.

## ADR

- [x] No new architectural decision beyond what's already documented inline (repository/service doc comments) — this box is trivially satisfied.


piotrswierzy (commented):

The mechanism is right. Keyset pagination on `(placedAt DESC, internalOrderId DESC)`, peek-one-extra, and a loop that runs **to exhaustion with no page cap** — the opposite of the `MAX_PAGES` shape ADR-048/068 exist to kill. No operator-facing count becomes a sample: `getCategoryCounts` / `getCategoryPage` / `getAllCategoryPages` all derive from the exhaustive `classify()`, and the `sum(page.items.length) === netExcludedCount` guard is preserved and now asserted across a multi-page population.

⚠️ **CI is currently red** — `Type Check` and `PHP Unit Tests` are failing. That needs resolving regardless of the below.

### On the four dimensions I checked specifically

- **ADR-068** — satisfied on the truncation axis. No ceiling in the loop; int-spec asserts full traversal with no duplicates and no gaps.
- **TypeORM safety** — clean. Every predicate parameter-bound. I also verified `take()` isn't a trap here: `applySalesAnalyticsScope` adds no join and `netEligible` is an `EXISTS`/`FILTER` subquery rather than a `joinAttribute`, so TypeORM takes the no-join branch and `take` becomes a plain `LIMIT`. It would have silently done nothing with a join present.
- **Boundary case** — covered. The exact-2×limit case is asserted, which is the peek idiom's one real trap.

### IMPORTANT — #2826's dedup guarantee is silently narrowed, and its spec can no longer fail

`tax-coverage-detection.service.ts` (190, 235)

`resolveRates` is now called inside `classifyBatch`, so `uniqueRateKeys` is built **per page**. A `(productId, variantId)` pair referenced on page 1 and page 3 is looked up twice, and `IProductsService.getEffectiveTaxRate` has no cache — it's 1–2 real DB queries per call. That is extra database work on the exact path #2826 optimised, in a PR whose purpose is performance.

The evidence problem is worse than the cost: the spec named

> `'resolves the catalogue rate ONCE per distinct (productId, variantId) pair, even when several orders/lines share it (#2826)'`

**still passes** — because its fixture is a single page. It now asserts something strictly weaker than its own name claims and cannot fail on the regression it was written to catch. That is the "check that cannot fail" shape.

The plan doc calls this "deliberately accepted", but it's accepted without measurement and without re-scoping the test. Cheap fix that keeps the memory win: hoist `rateByKey` (and the seen-key set) into `classify()` and pass it into `classifyBatch` — dedup stays population-wide, only *candidates* are held one page at a time. If page-scoping is genuinely intended, re-fixture the #2826 spec across two pages and rename it, so a reader isn't told a promise the code no longer keeps.

### IMPORTANT — the early empty-page break depends on an unstated port invariant

`tax-coverage-detection.service.ts` (165)

`if (page.items.length === 0) break;` fires *before* consulting `nextCursor`. That's safe only if "an empty page implies `nextCursor === null`" — true of this repository, but the port JSDoc documents neither that invariant nor its converse. A conforming-but-different implementer (one that filters rows after the peek and can return an empty page mid-population) makes `classify()` stop early and under-report every category count, silently. That is the defect ADR-068 was written about, one layer up.

Preferred fix: **delete the early break.** `classifyBatch([])` is free (`OrderLineItemRepository.findByOrderIds` short-circuits on an empty array) and the `nextCursor === null` check below already terminates — so the loop depends on one contract instead of two. Alternatively state the invariant in the port JSDoc.

### IMPORTANT — unbounded `for (;;)` on a synchronous request path

`tax-coverage-detection.service.ts` (157)

Against the shipped repository this cannot spin — the strict `<` tuple predicate guarantees forward progress — so this is defence-in-depth, not a live defect. But `classify()` is coded against the *port*, `GET /analytics/coverage` is hit unconditionally on `/analytics` mount, and a non-advancing cursor from a future implementer hangs the request forever holding a pooled connection. A generous ceiling (say 10 000 pages) that **throws** a named exception costs three lines. Please not a silent `break` — that's the flag-shaped alternative ADR-068 rejects.

### IMPORTANT — the integration tests carry all the evidence and were not run

The quality-gate box for `pnpm test:integration` is unticked ("no Docker in this environment"). Every unit spec mocks the query builder, so the keyset SQL — the `<` tuple predicate, the `addOrderBy` tiebreak, and whether `take` actually emits `LIMIT` against a real server — never executes in `pnpm test`. The three new int-spec cases are the entire proof of correctness for the central mechanism.

### SUGGESTIONS

- The renamed test `'never calls findNetExcludedOrderCandidatesPage more than once, regardless of category count'` is now literally false — with a real multi-page population it *is* called once per page. Its intent still holds; rename to state the intent (`'classifies once for every category, never once per category'`).
- `docs/plans/implementation-plan-tax-coverage-bounded-fetch.md` is committed with session-scope prose that is now false in the repo — *"no `git commit`, no `git push`, no PR, no GitHub comments"*, twice. A future reader of a durable `docs/plans/` file will read a permanent instruction not to do the thing that plainly happened. Strip both paragraphs.
- Port JSDoc: `limit`'s stated justification is test convenience, which invites a future caller to pass a small value in earnest — say plainly that no production caller passes it. And mid-loop consistency is undocumented: a row whose FX stamp lands mid-loop can enter or leave the predicate and be skipped. One sentence saying this is not a consistent snapshot, and why that's fine for a diagnostic panel, would stop the next reader assuming otherwise.

### Done well

- **The `internalOrderId DESC` tiebreak is a latent bug fix, not just a keyset requirement.** `getCategoryPage` slices by offset; under `placedAt DESC` alone, rows sharing a `placedAt` had no deterministic order, so a drill-down's page 2 could repeat or skip a row between requests. The int-spec deliberately seeds an identical `placedAt` pair across a page boundary — exactly the right test.
- **`findByOrderIds`' `IN`-list is now bounded too**, as a side effect. That read was unbounded in the same way; nobody claimed it, but it's real.
- **The peek idiom is implemented correctly**, including the detail most often got wrong: `nextCursor` comes from the last row of the *trimmed* page, never the peeked row — and that's pinned by a dedicated unit test.
- **The `classifyBatch` extraction is faithful** — diffed line by line against the pre-change body; filter, batched line read, dedup map, `resolveRates` and the `classifyOne` loop are unchanged in logic.
- **The port rename is fully swept** — service, both specs, int-spec, the controller spec's mock, the `apps/e2e` comment, the barrel. One implementer, and no plugin depends on a repository port, so no published-contract break.

---

norbert-kulus-blockydevs:
## Tech Lead Review — PR #3000

Reviewed as a senior backend architect: **orders** context, CORE layer, replacing an unbounded SQL read with a keyset-paginated one behind `TaxCoverageDetectionService.classify()`.

### Summary

This is exactly the class of fix that's usually where a "bound the fetch" perf PR quietly introduces silent data loss — and this one gets the hard parts right. `findNetExcludedOrderCandidates` (one unbounded `getRawMany()`) is replaced by `findNetExcludedOrderCandidatesPage`, a proper **keyset** pagination (`ORDER BY placedAt DESC, internalOrderId DESC`, cursor predicate `(placedAt < :cursor OR (placedAt = :cursor AND internalOrderId < :cursorId))`) — not `LIMIT`/`OFFSET`, which would silently skip/duplicate rows under concurrent writes. The classification loop in `classify()` drives it to exhaustion inside one synchronous request and accumulates per-page results, which is provably safe here because per-order classification never depends on another order's state (stated explicitly in the class doc comment, and it's true — worth confirming it stays true if this classifier ever grows a cross-order rule).

### What's done right

1. **Correct keyset direction and tie-break.** DESC-DESC ordering paired with `<`/`<` cursor comparison is the correct direction for a descending walk (a common off-by-direction bug in hand-rolled keyset pagination). `internalOrderId` as tiebreak is justified (unique, non-null, already used as a map key elsewhere in this service) and is int-spec-tested against **rows sharing an identical `placedAt`** — the one case that actually exercises the tiebreak.
2. **Peek-one-extra-row idiom**, correctly implemented: `take(limit + 1)`, `hasMore = rows.length > limit`, `nextCursor` derived from the **trimmed** page's last row, never the peeked row. Also int-spec'd at the exact 2×limit boundary.
3. **Population-wide dedup preserved across pages.** This is the standout finding: per the PR description and the `fix(orders): preserve population-wide tax-rate dedup across pages` commit, the *first* draft apparently reset the `(productId, variantId)` → rate dedup map per page, silently regressing #2826's population-wide catalogue-lookup dedup — a real correctness/perf regression that a single-page-fixture test couldn't have caught. It was caught in review and fixed by threading one `rateByKey` map owned by `classify()` through every page's `classifyBatch()` call. This is the right instinct and the right fix.
4. **Defence-in-depth against a non-conforming port implementation** (`TaxCoveragePageCeilingExceededError`, `MAX_PAGES = 10_000`) — the service is coded against the *port*, and a hypothetical future/third-party implementer whose cursor never returns `null` would otherwise hang the request forever. Throwing instead of silently `break`-ing is the correct choice — a silent break would under-report every category with no signal, which is a worse failure mode than a loud error.
5. **Exact-parity test discipline.** Multi-page-vs-single-page equivalence is asserted, not assumed; the #2465 `candidates.length === netExcludedCount` regression guard is preserved across the pagination boundary via a live-Postgres int-spec, not just a mocked unit test.
6. **No ORM entity changed → no migration needed**, correctly checked as trivially satisfied.

### Items worth a look (non-blocking)

- **[SUGGESTION]** `NET_EXCLUDED_CANDIDATE_PAGE_SIZE = 500` is a plain file-local `const`, explicitly *not* wired into `OL_OPERATIONAL_SETTINGS` (per docs/architecture-overview.md's Operational Settings / ADR-069 pattern for sweep/page-size tuning). The PR's own doc comment justifies this as "request-scoped, not operator-tunable" — reasonable, since this runs inside one synchronous HTTP request rather than a background sweep. I'd only push back if `GET /analytics/coverage` starts timing out for large windows in production; at that point the fix is more pages, not a bigger page, so the constant is probably fine as shipped.
- **[SUGGESTION]** No explicit index change accompanies the new `ORDER BY placedAt DESC, internalOrderId DESC` + cursor predicate. If `order_records` doesn't already have a composite index covering `(reportingCurrency, cancelledAt, placedAt, internalOrderId)` (or close to it) for the sales-analytics scope, each page still does a sort-heavy scan over the filtered window — bounded now, but potentially still slow per page on a large window. Given this doesn't *regress* anything (the old unbounded query had the identical sort with no LIMIT), this is a "worth profiling" note rather than a blocker.
- **[SUGGESTION]** The `MAX_PAGES = 10_000` ceiling throws a plain `Error` subclass with no HTTP-layer mapping visible in this diff — worth a quick check that `GET /analytics/coverage`'s controller/exception filter maps `TaxCoveragePageCeilingExceededError` to something other than a raw 500, so an operator sees an actionable message rather than a stack trace, on the vanishingly unlikely path where it fires.

### Merge Readiness

✅ **Approve.** Correct keyset-pagination mechanics, a real cross-page correctness bug caught and fixed during the PR's own review cycle (visible in its commit history), strong parity test coverage against real Postgres, and no architecture-boundary or migration concerns. Nothing here blocks merge.

piotrswierzy (approved):

All four IMPORTANT findings from my previous review are resolved, and **each took the offered fix rather than routing around it**. CI went from two red checks to fully green (8/8, including `Integration Tests`).

| Prior finding | Status |
|---|---|
| #2826 dedup narrowed to page-wide | ✅ **Resolved** |
| Empty-page break on an unstated invariant | ✅ **Resolved** (preferred fix) |
| Unbounded `for (;;)` | ✅ **Resolved** (throws, not a silent break) |
| Integration tests not run | ✅ **Resolved** by CI |

**The dedup fix is the one I care most about, and it's done properly.** `rateByKey` is declared in `classify()` and threaded into `classifyBatch(page.items, rateByKey)`; `classifyBatch` builds `uncachedRateKeys` skipping keys already in the map and merges `resolveRates`' output back. Dedup is population-wide again; only candidate rows are page-bounded.

Crucially the "check that cannot fail" is closed **by a new test, not by renaming the old one** — `spec.ts:422` drives a genuine two-page fixture sharing `productId: 'shared-product'` and asserts `getEffectiveTaxRate` was called **once**. That test fails against the per-page implementation. The single-page #2826 spec is retained beside it, and the plan doc's "deliberately accepted" paragraph was replaced with an accurate account of the reversal.

The ceiling is exactly what ADR-068 asks for: `MAX_PAGES = 10_000` throwing the new `TaxCoveragePageCeilingExceededError` (correct layer, correct naming, exported from the barrel), whose docblock argues against a silent `break` and cites the ADR. Off-by-one is right — it fires having read exactly 10 000 pages.

### Two new suggestions, neither blocking

**A cached `null` now poisons every later page for that key.** `resolveRates` records a *failed* lookup as `null` (deliberately, folded into `'not-checked'`), and the skip test is `rateByKey.has(key)` — so a `null` counts as resolved and a transient `getEffectiveTaxRate` failure on page 1 is permanent for the rest of the run rather than retried per page. This is an inherent consequence of the population-wide dedup I asked for, and arguably the better spelling (classification becomes deterministic across pages instead of luck-of-the-page). But it's a real behaviour delta from both the pre-#2834 code and the intermediate version, and the comment explaining the map doesn't mention it. One sentence on `rateByKey`'s docblock would stop the next reader reading the skip as purely an optimisation.

**The ceiling test performs 10 000 real loop iterations** — each an `await` plus a `classifyBatch`. It passes and it's the honest way to test the guard, but it's by far the slowest unit test in the file. Asserting the constant, or injecting the ceiling, would be cheaper. Genuinely optional.

### Carried over, all doc/naming polish

- `spec.ts:764` — `'never calls findNetExcludedOrderCandidatesPage more than once…'` is still literally false under pagination. Rename to the intent.
- Port JSDoc — `limit` is still justified as test convenience; still no "not a consistent snapshot" note.
- Tick the `test:integration` box / drop the stale "no Docker" note, now that CI proves it.

---

norbert-kulus-blockydevs:
Addressed the remaining doc/naming suggestions from the re-review:

- Renamed the test asserting `classify()` runs once for the whole multi-category read — `'never calls findNetExcludedOrderCandidatesPage more than once, regardless of category count'` was literally false under real pagination (it *is* called once per page); it's now `'classifies once for every category, never once per category'`, which states the actual guarantee.
- `findNetExcludedOrderCandidatesPage`'s port JSDoc now says plainly that `limit` has no production caller (it exists for test fixtures to exercise multi-page behaviour without a large seed) and that the read is **not** a point-in-time-consistent snapshot across the page loop — a row whose predicate inputs change mid-loop can enter/leave the population between pages, same caveat `getDailyOrderAggregates`'s `net_excluded_count` already carries and for the same reason.
- Added a note on `rateByKey` in `classify()`: because a failed catalogue lookup is cached as `null` and `uncachedRateKeys` skips any key already present (null included), a transient failure on an early page is never retried on a later page within the same `classify()` call — a deliberate consequence of making dedup population-wide, not a bug.
- Updated the PR checklist: ticked `pnpm test:integration` and dropped the stale "no Docker in this environment" note, since CI confirmed it green (8/8, Integration Tests included).

Left the ceiling-test performance suggestion (10 000 real loop iterations) as-is — you flagged it as genuinely optional, and reducing it would mean either exporting `MAX_PAGES` for injection or weakening the exact-boundary assertion, which felt like more churn than the "neither blocking" suggestion warranted. Happy to do it if you'd still like it.

Committed with `--no-verify` per request and pushed.

piotrswierzy (approved):
## Tech Lead review

Independent second review (there is already an approval from @piotrswierzy on `5db3e41`; the only commits since are the doc/naming polish that review asked for, and CI is now green 9/9). **I agree with the approval.** The mechanism is right: keyset pagination on `(placedAt DESC, internalOrderId DESC)` with a strict tuple predicate, peek-one-extra with `nextCursor` taken from the *trimmed* page, a loop that runs **to exhaustion** rather than truncating (ADR-068), and population-wide `(productId, variantId)` dedup preserved across pages by threading one `rateByKey` map — the regression that would otherwise have silently undone #2826. I verified `applySalesAnalyticsScope` enforces `placedAt IS NOT NULL`, so retyping `placed_at` from `Date | null` to `Date` and omitting a null-ordering case is sound rather than optimistic; and `buildNetSalesOrderFragments(includeBackfilledPreRollout)` still takes the flag as a required argument, so ADR-063/#2469's "required parameter, never a repository-held default" rule on the net-eligibility SQL is unchanged by this diff.

On the two specific conventions I was asked to check: `docs/engineering-standards.md § When A Paginated Total Is Expensive` (#2950/#2944) does not bind here — that rule governs a *list + total* read where a `COUNT` scans the table beside a bounded page, and prescribes the `findMany` / `findManyRows` / `countMany` split. This read emits no `COUNT` at all and has exactly one predicate builder, so the properties that rule protects (one shared predicate, no sort/pagination on the count, never rendering `0` for an absent total) are satisfied vacuously. Nothing here introduces a total that could be reported as `0`.

### Findings

**[SUGGESTION]** — `libs/core/src/orders/application/services/tax-coverage-detection.service.ts:36-55` (class doc) and `:143-152` (`classify` doc)
> The memory claim is overstated in a way a future reader will act on. `classify()` accumulates a `TaxCoverageOrderRow` per candidate into `result` for the whole population (`TaxCoverageClassification` is three arrays over every candidate), and `getCategoryCounts` / `getCategoryPage` require that — so what is bounded is the size of any **single SQL statement and its result set**, not the request's peak memory or its total work, which both still scale linearly with the net-excluded population for the window. The class doc says batching "only changes how many round trips the base fetch costs and **how much is held in memory at once**"; the second half is materially false, since the accumulating result is the dominant term. `classify`'s own docblock is more careful ("beyond the accumulating `result`"), so the two disagree. Say it once, plainly, in both places: this removes the *unbounded single read* and leaves per-request cost linear in the window — which is the correct trade-off, because #2834's alternative shape (a bounded sample plus a separate SQL `COUNT`) would turn an operator-facing count into a sample, exactly what ADR-068 exists to refuse. Worth stating explicitly so nobody later reads #2834's "no longer scales linearly with total order history" as closed.

**[SUGGESTION]** — `libs/core/src/orders/domain/exceptions/tax-coverage-page-ceiling-exceeded.error.ts`
> `TaxCoveragePageCeilingExceededError` has no HTTP mapping anywhere in the diff, so on the (unreachable-against-the-shipped-repository) path where it fires, `GET /analytics/coverage` answers a raw 500. That is arguably the honest status — a non-conforming port implementation *is* a server fault, not a client one — so I would not add a filter just to change the number; the note is only that the choice is currently implicit. One sentence on the exception's docblock saying 500 is the intended answer would close it.

Nothing else. In particular the two things most likely to be wrong in a hand-rolled keyset walk are both right and both pinned by the live-Postgres int-spec rather than by a mocked builder: rows sharing an identical `placedAt` across a page boundary (the tiebreak's only real test), and the exact 2×limit case (the peek idiom's only real trap). I also confirmed `take()` is not a trap here — `applySalesAnalyticsScope` adds no join and `netEligible` is an `EXISTS`/`FILTER` subquery rather than a `joinAttribute`, so TypeORM emits a plain `LIMIT`.

### Verdict
✅ **Approve**

Both suggestions are doc-only and neither blocks merge.

---

Issue #2791, opened 2026-09-01
check-cross-context-imports' ALLOW_LIST carries 8 stale entries, and nothing verifies the map


piotrswierzy:
Found during the #2673 sweep for checks-that-cannot-fail.

## The defect

`scripts/check-cross-context-imports.mjs`'s `ALLOW_LIST` silences a specific `(file, symbol)` pair. All 107 declared paths resolve, but **4 symbols no longer appear in the files they are listed against** — verified 0 occurrences each, and each appears twice (service + spec), so 8 dead rows:

| symbol | file |
|---|---|
| `ProductRepositoryPort` | `inventory-query.service.ts` |
| `ProductVariantRepositoryPort` | `order-item-ref-resolver.service.ts` |
| `ProductVariantRepositoryPort` | `offer-mapping-sync.service.ts` |
| `ProductVariantRepositoryPort` | `offer-builder.service.ts` |

These are exactly the #718 slice-1 rewires the script's own comment at `:103` claims were "dropped from this list". **The code shipped; the entries did not.**

## Why it matters

The per-symbol gate is the mechanism that makes this allowlist safe — a new deny-pattern import in an already-listed file still fails, because only the *named* symbol is silenced. A stale entry inverts that: those four core services can silently re-acquire a `products` repository port, in direct violation of the cross-context rule (`architecture-overview.md § Cross-context dependencies in core`), with the build green and no reviewer prompt.

This is not hypothetical scope-widening — it is a live hole in the guard that #713/#718/#721 exist to provide.

## The fix

`scripts/check-architecture-gates.mjs` already re-verifies its own allowlist entries. This script only ever *reads* the map (`:757`). Add the symmetric check: **an allowlist entry whose symbol no longer appears in its file is a FAILURE**, telling the author to delete the row.

That direction is safe to enforce strictly — removing a stale row can only make the gate stricter, never looser — and it makes the allowlist self-retiring as #718 and #722 land, instead of accumulating permissions nobody remembers granting.

## Acceptance criteria

- [ ] A stale `(file, symbol)` allowlist entry fails `check:invariants` naming the row to delete.
- [ ] The 8 dead rows above are removed.
- [ ] Demonstrated by making it fail first (add a bogus entry, see it named).

Related: #2673, #2790 (same class), #718, #722.

piotrswierzy:
Closing — both halves shipped.

[...]

Issue #2301, opened 2026-08-21
[TASK] CORE — terminal state for a poison stream entry on the two remaining PEL consumers


piotrswierzy:
## Problem / Context

ADR-049 records a **Known gap — no terminal state for a poison entry**. #2164 made repeated delivery reachable for the first time (before it, a failing entry was simply never redelivered), so an entry whose handler always throws is now retried **indefinitely**.

#2280 closed this for the webhook path only — that path no longer has a PEL in its durable spine, and its ingress routing failures become durable `deadlettered` delivery rows. The ADR amendment records that reversal gate as fired. **Two consumers still carry the gap:**

- `apps/worker/src/sync/job-intake.consumer.ts`
- `apps/worker/src/events/master-deletion-to-job.handler.ts`

Today's mitigation is containment, not resolution: per-entry error isolation stops a poison entry starving its siblings, the drain pages forward with an exclusive cursor so one stuck entry cannot re-present its own page, and `RecoveryAttemptTracker` (`libs/shared/src/redis/stream-consumer.ts`) alarms **once** on crossing `MAX_RECOVERY_ATTEMPTS`. After that alarm the entry is retried forever with no further signal and no terminal state.

Auto-dead-lettering was deliberately rejected at the time: neither consumer can construct its dead-letter payload from a raw pending entry (job-intake needs a parsed job request; the deletion handler needs a decoded event), and discarding the entry instead would be unrecoverable loss.

## Proposed Solution

**This issue is deliberately gated, not ready to build.** ADR-049's reversal gate is *prose-only* and reads: "the first poison entry observed in production, or the Wave 5 spine (decision 1) removing the PEL from the durable path entirely." Wave 5 removed the PEL from the *webhook* path only, so the gate is half-fired.

Two candidate directions, to be decided when the gate actually fires:

1. **Extend decision 1 to these paths** — the preferred shape, and the one that dissolves the problem rather than managing it. `job-intake` is the strong candidate: its whole job is to write a `sync_jobs` row, which is exactly the durable work row decision 1 describes, so the stream hop may be removable outright the way the webhook path's was.
2. **A terminal state for an un-decodable entry** — for whatever PEL remains, a bounded attempt budget after which the entry is ACKed into a durable record carrying the raw fields (not a typed payload, which is precisely what cannot be constructed). This needs a decision on where that record lives, since neither consumer has a `webhook_deliveries` equivalent.

[...]

Issue #2280, opened 2026-08-21
[IMPL] Wave 5 — durable work-row spine on the webhook path (ADR-049 decision 1)


piotrswierzy:
> **Sub-issue of #2162** (Wave 5 — implements [ADR-049](https://github.com/openlinker-project/openlinker/blob/main/docs/architecture/adrs/049-durability-spine-and-domain-event-contract.md) decision 1, webhook path first)

## Problem / Context

ADR-049 decision 1 is stated but not yet implemented anywhere: *"the spine is the work row, written in the same transaction as the business change."* On the webhook path today, durability sits **seven hops** from ingress: HTTP → `webhook_deliveries` row → `XADD events.inbound.webhooks` → `WebhookToJobHandler` → routing → `XADD jobs.sync` → `JobIntakeConsumer` → `sync_jobs` row. #2164 made the intermediate PEL hops *recoverable*, but recovery is mitigation, not removal: the stream entries remain part of the durable path, the `jobdedup:*` Redis gate can strand a trimmed-but-unconsumed `jobs.sync` entry behind its 7-day TTL (documented in [ADR-049](https://github.com/openlinker-project/openlinker/blob/main/docs/architecture/adrs/049-durability-spine-and-domain-event-contract.md) / `docs/operations/redis-stream-retention.md` as operator-recovered only), and the **poison-entry known gap** (an entry whose handler always throws is redelivered indefinitely; no terminal state) stays open — ADR-049 names "the Wave 5 spine removing the PEL from the durable path" as one of its two closing conditions.

ADR-049 chose the webhook path first because loss there is least detectable: a dropped order still reads `webhook_deliveries.status = 'published'`.

## Proposed Solution

Implement decision 1 on the webhook path: **the `sync_jobs` row (the work row) is written in the same Postgres transaction as the `webhook_deliveries` gate row**, moving the durable write to hop one. Redis becomes a wake-up hint, not a system of record, for this path.

- **Routing moves to ingress**: the translate→route step (`WebhookEventTranslator` → `InboundRoutingPolicy`, ADR-015) runs synchronously in `apps/api/src/webhooks/` so the jobType + payload are known inside the gate transaction. The capability-translated routing contract is unchanged — only *when* it runs moves.
[...]

PR #3011, opened 2026-09-08
chore(webhooks): remove LegacyInboundWebhookDrain and the events.inbound.webhooks streams

### LINKED ISSUE #2300: [TECH-DEBT] API — remove LegacyInboundWebhookDrain and the events.inbound.webhooks stream it drains
## Problem / Context

`apps/api/src/webhooks/application/handlers/legacy-inbound-webhook-drain.ts` is a **one-shot upgrade-migration artifact** shipped with #2280 (ADR-049 decision 1). Its own file header states the removal condition explicitly:

> Removal of this drain (and the stream names it reads) is a follow-up release — see `docs/operations/redis-stream-retention.md`.

Nothing tracked that follow-up, which is what this issue exists to fix.

It exists because the durable spine retired the always-on `WebhookToJobHandler` consumer (confirmed absent from `main`: `webhooks.module.ts` no longer references it). Entries already sitting in `events.inbound.webhooks` at upgrade time would otherwise be stranded — the source's redelivery bounces off the Postgres gate (a `published` row exists) *without* creating a job, i.e. silent order loss. The drain consumes that backlog through the new resolve→gate path, then exits.

Once every deployment has booted at least once on a post-#2280 release, the backlog is drained and the drain is dead code that still runs (detached) on every api boot, still holds a Redis client, and still keeps `events.inbound.webhooks` alive as a concept in the retention policy and stream-name union.

## Proposed Solution

1. Delete `legacy-inbound-webhook-drain.ts` + its spec, and unregister it from `apps/api/src/webhooks/webhooks.module.ts`.
2. Remove `events.inbound.webhooks` from the `RedisStreamName` union and its retention entry (`libs/shared/src/redis/`) — `scripts/check-stream-writes.mjs` types stream names off that union, so a lingering writer becomes a build failure rather than a silent revival.
3. Update `docs/operations/redis-stream-retention.md` and the ADR-049 amendment to record the drain as removed, plus a one-line operator note that the stream key can be `DEL`'d.

**Do not ship this in the release immediately following #2280.** The drain must survive at least one full upgrade cycle, since an operator skipping a release still needs it. State the assumed minimum-version floor in the PR description.

## Classification

**Type**: CORE / Infrastructure
**Layer**: Interface (api webhooks) + Shared (redis stream registry)
**File(s)**: `apps/api/src/webhooks/application/handlers/legacy-inbound-webhook-drain.ts` (+ `.spec.ts`), `apps/api/src/webhooks/webhooks.module.ts`, `libs/shared/src/redis/`, `docs/operations/redis-stream-retention.md`, `docs/architecture/adrs/049-durability-spine-and-domain-event-contract.md`

## Dependencies

- #2280 (durable webhook spine) — merged, this is its named follow-up
- Requires a release boundary: at least one shipped release carrying the drain

## Assumptions

- No deployment upgrades across more than one release boundary without booting. If that is not safe to assume, the drain stays and this issue is closed `not_planned` with the reason recorded.

## Acceptance Criteria

- [ ] Drain class, spec and module registration removed
- [ ] `events.inbound.webhooks` removed from the stream-name union and retention policy; `pnpm check:invariants` passes
- [ ] ADR-049 amendment and `docs/operations/redis-stream-retention.md` record the removal
- [ ] PR description names the minimum version an operator must have booted before upgrading
- [ ] Webhook integration suites pass unchanged (the drain is not on the live ingress path)

piotrswierzy:
Closes #2300

#2280 moved webhook routing to ingress, so the `sync_jobs` work row commits in the same Postgres transaction as the `webhook_deliveries` gate row and nothing publishes to `events.inbound.webhooks` any more. `LegacyInboundWebhookDrain` was the one-shot boot pass that recovered the pre-#2280 backlog through the new resolve-then-gate path. It shipped in v0.8.0 and has run at every api boot through v0.10.0 — three releases against a condition one clean boot resolves.

**18 files, +407 / −887.**

## What goes

The drain, its spec, and `webhook-handler.types.ts` — whose `WebhookPayload` / `WebhookMetadata` had the drain as their only consumer, so the `handlers/` directory goes with them (verified: zero surviving references anywhere in the tree).

`events.inbound.webhooks` and `events.inbound.webhooks.dead` leave `REDIS_STREAM_NAMES` and `STREAM_BOUNDS`. After #2280 neither had a writer, and removing the drain removed their last reader — a name in that map with neither is a retention decision about nothing.

## Four things worth reviewing rather than skimming

**Both `ALLOW_LIST` rows in `scripts/check-cross-context-imports.mjs` are deleted in this same commit.** Since #2791 a row naming a file that no longer imports the symbol *fails the build*, so this is not optional tidying — the guard reports `117 entries across 100 files`, down from `119/102`.

**The shared PEL primitives are deliberately NOT touched.** `ackTrimmed` and `resolveConsumerName` are still called by both live consumers (`job-intake`, `master-deletion-offer-pause`), and `toPendingRows` — which loses its last *external* caller here — remains the engine of `readOwnPending` and `reclaimOrphans`. Live code with no external caller, not dead code. #2301 builds directly on these.

**The int-spec's negative assertion is RETAINED**, with its constant renamed to `RETIRED_INBOUND_WEBHOOK_STREAM`: `EventPublisherPort.publish` takes a plain `string`, so reviving the stream would not be a compile error, and this is its only runtime proof.

**One test honestly lost discriminating power, and says so.** After the retirement the surviving `~` MAXLEN threshold case coincides with `DEFAULT_STREAM_BOUND`, so it can no longer tell "a registered bound is honoured" from "fell through to the default". The comment records that and names the `healthcheck` exact-trim case as where the property is still genuinely pinned — rather than leaving the assertion reading stronger than it is.

## Operator-facing: the upgrade floor is v0.8.0

`git tag --contains 25111eb93` returns exactly `v0.8.0, v0.9.0, v0.10.0`. An operator jumping from **v0.7.0 or earlier** straight to a post-removal release never ran the drain, and can still hold a backlog recorded `published` with no job. That, and the one-time `DEL` reclaiming both keys, are documented in `docs/operations/redis-stream-retention.md`. Upgrade via ≥ v0.8.0 first.

## Testing

Pre-commit gate green in full (lint, type-check, `smart-test --no-integration`), and the full `pnpm lint` / `type-check` / `test` re-run is reported in review. **`pnpm test:integration` has not been run yet** — `webhook-ingestion.int-spec.ts` is modified by this PR and will be exercised serially before merge; flagged rather than implied.


piotrswierzy:
## Review: ✅ Approve — CI green after one test-isolation fix

**The integration failure was not this PR's product code.** `invoicing-list.int-spec.ts` asserted a **global** `total: 0` and got `total: 2` — two `ol_order_*` / `ksef` invoice rows created ~17 minutes earlier by a different suite.

That spec resets only in `afterEach`: it cleans up after itself and inherits whatever the preceding *suite* left in `invoice_records`. The table **is** in the truncate set (`setup.ts:135`, listed explicitly *because* it has no FK and so sits outside the CASCADE closure — the comment there says `resetTestHarness()` "would never clear it" otherwise), and `maxWorkers: 1` rules out parallel-worker interference. The sole exposure was that nothing reset *before* the first case.

That makes the assertion order-dependent, and jest's suite order is not stable — it derives from cached per-file timings. This PR perturbed both inputs: api boot no longer runs the one-shot drain, and it deleted a describe block from `webhook-ingestion.int-spec.ts`. Either alone can reshuffle the run so an invoice-creating suite lands ahead of that one.

Fixed in `6a2c5e96a` where the fragility is — a `resetTestHarness()` in `beforeAll` — rather than by hunting whichever suite happened to sort first. The existing `afterEach` already covers isolation between that file's own cases (several later ones create records), so only the entry point needed it. Deliberately **not** weakened to a connection-scoped assertion: *"the list is empty on a clean database"* is the property that case exists to prove, and scoping it would keep the test green while testing less. **Now 9/9 green.**

It is a genuine latent bug that would have bitten any PR shifting suite timing — worth knowing it existed independently of this work.

## The change itself

Small and correct: the drain and its spec deleted, both `events.inbound.webhooks*` names out of `REDIS_STREAM_NAMES` and `STREAM_BOUNDS`, the module provider gone, both stale `ALLOW_LIST` rows removed (117/100, down from 119/102).

Three things it got right that were easy to get wrong:

- **The int-spec assertion that ingress writes nothing to the retired stream is retained**, not deleted with the drain — correctly justified, since `EventPublisherPort.publish` takes a plain `string`, so a revival would not be a compile error. It is not vacuous either: it fails iff something publishes.
- **The repointed `~`-MAXLEN test is honest about what it lost.** `masterDeletion`'s 10 000 equals `DEFAULT_STREAM_BOUND`, so that case no longer distinguishes "registered bound honoured" from "fell through to default" — stated in the comment, with `healthcheck` named as where the property is still pinned. Most repoints would have silently weakened the test.
- **The upgrade floor is stated, not hand-waved** — v0.8.0, in the runbook, ADR-049 and the architecture doc, with the `DEL` that reclaims both keys.

One IMPORTANT finding applied in `fa1e4ead4`: `§ 5 Master-Deletion Event Flow` still opened by describing `events.master.deletion` as *"a second, narrower core-domain-event stream alongside `events.inbound.webhooks`"* — a present-tense live-sibling claim about the stream this PR deletes. The same dangling contrast was repointed in two other places; that one was missed.

PR #3012, opened 2026-09-08
feat(fulfillment): reap dispatches whose holder never answered (#2712)

### LINKED ISSUE #2712: ADR-054's timeout-as-rejection sweep: a `submitted` fulfilment work whose holder never answers is never reaped
Deferred from **#2399** (merged) with the constraint recorded there. Filed so it has an owner.

## The gap

#2399 shipped the executor handshake: OpenLinker dispatches work, the holder accepts or rejects, and the attempt claim makes the offer at-most-once per attempt.

**Nothing handles a holder that never answers.** A `FulfillmentWork` in `submitted` whose holder is silent stays `submitted` indefinitely — no sweep reaps it, no timeout converts it to a rejection, and nothing re-routes it.

ADR-054 anticipates this as a timeout-as-rejection pass; #2399's scope stopped at the handshake itself.

## What #2399 got right, and must be preserved

The current behaviour is **safe, not merely absent**: the work is never silently retried. The resume path is attempt-scoped and terminal-safe, so a stalled work cannot quietly become a double-dispatch. That is the correct failure direction — a shipment that is late is recoverable; one that ships twice is not.

Any sweep must keep that property. Specifically, converting a timeout into a rejection **frees the work to be re-routed**, so the sweep is the moment where the double-ship becomes possible if the claim is not respected.

## Scope

1. Define the timeout — per holder, per work, or a declared value the executor supplies. Note #2229's rule: a ceiling an adapter declares should be **reported === enforced** through one resolution path, so a number shown to an operator cannot drift from the one applied.
2. Convert an expired `submitted` into a rejection **through the same claim discipline** the handshake uses, so a holder answering late cannot collide with a sweep that already moved on. #2399's guard yields distinct attempts precisely so a stale answer is identifiable.
3. Decide what happens next — re-route (which is #2395's re-source loop, already noted there as a deferred constraint) or hold for an operator.
4. Make the reaping **observable**. A work that timed out and was re-routed must say so; a silent substitution of holder is worse than the stall it replaces (the same reasoning as #2704).

## Acceptance criteria

- [ ] An expired `submitted` work reaches a terminal or re-routable state rather than sitting indefinitely.
- [ ] A holder answering *after* the sweep cannot cause a second dispatch — proven by an overlapping test, not a sequential one (a sequential test passes against no guard at all).
- [ ] The timeout value is resolved through one path that both reports and enforces it.
- [ ] The reaping is visible to an operator.

## Related

- **#2399** — the handshake and the attempt claim this builds on.
- **#2395** — owns the re-source loop; the `orderId` lineage is already on the row.
- **ADR-054** — names the timeout-as-rejection pass.

piotrswierzy:
Closes #2712

ADR-054 names a timeout-as-rejection sweep and #2399 shipped everything except it, so a `FulfillmentWork` offered to a holder that never answered sat in `submitted` for ever — the operator saw work assigned to a holder that will never ship it, and re-sourcing could not begin because the negotiation axis had not terminalised. `fulfillment.work.timeoutSweep` converts such a work into `rejected`, which **is** the re-routable state (`CLAIMABLE_FROM` already contains it).

**35 files, +2412 / −17.**

## The decision the slice turns on: the reap is NON-blocking

A holder's own `blocking: true` is that holder **declaring** it cannot do the work. A silence is OpenLinker **inferring** it — and the standing rule (#2243) is that a destination's declaration is a fact while our inference only warns.

The practical half is decisive: exclusions accumulate as rows with nothing that clears them, and on the only shipped topology there is exactly **one** executor (`openlinker.oms.v1`, #2409). A blocking timeout would therefore permanently exclude the only possible holder after a transient outage, with no in-product remedy.

The consequence is recorded as a forward constraint on **#2395** in the constant's own docblock: because a timeout deliberately does not exclude, the re-source loop must bound its own attempts rather than rely on exclusion to terminate.

## It adds no new writer of `requestStatus`

The reap goes **through** the unchanged `recordRejection`, so the guarded `submitted -> rejected` transition and the rejection row stay one statement pair with one owner — the writer discipline `fulfillment-work.repository.ts`'s header table exists to protect.

It also **pins the attempt it read**. A sweep is a delayed actor by definition, so a row that left and re-entered `submitted` inside the read-then-write window would otherwise take a rejection naming an attempt that is no longer live. `expectedAssignmentAttempt` is optional, so the existing handshake path is byte-identical by omitting it.

## Frontier-as-query, no cursor and no offset

The candidate set is `requestStatus = 'submitted' AND updatedAt < cutoff`, and **every page consumes its own selection** — a reaped row leaves the set. An advancing offset would step over rows, which here means a work that is never reaped: the exact defect this sweep exists to fix, reintroduced by its own pagination. `bounded-sweep.ts` draws that distinction in its own header; #2317 and #2346 recorded the identical reasoning for the same self-consuming shape. The sweep *primitives* are reused and the offset machinery is not, and no `MasterSweepKind` member is added — `sweepLockKey` renders `master:{kind}:…`, a false name for a pass with no master.

## Lane: `bulk`, not the `realtime` its siblings take

ADR-050 picks by cost of starvation. This is a cron-paced reconciler over work **already stalled by definition** — nobody is waiting on it, and one more tick of delay on a dispatch silent for hours is immaterial. Its three `fulfillment.work.*` siblings are `realtime` because a route, a dispatch or a status sync is waited on; letting a reaper take a `realtime` slot from a live dispatch is the wrong trade in the one direction that costs a shipment. Direct precedent: `inventory.reservations.expire`.

## No migration

`IDX_fulfillment_works_request_status` on `['requestStatus','updatedAt']` was created by #2392 **for this sweep by name**, and is verified identical in `1864000000000-create-fulfillment-works.ts:151` and `fulfillment-work.orm-entity.ts:90` — already diffed by `fulfillment-work-migration-parity.int-spec.ts`. No ORM entity changed, so `migration:generate` has nothing to emit. The frontier read is ordered `updatedAt ASC` to match that index's column order.

## Also in scope

Lists `fulfillment_work_rejections` in the integration truncate set — a **pre-existing #2399 gap** (its FK to `fulfillment_works` is migration-only, so the `CASCADE` closure walk cannot reach it). This sweep is the first thing to write rejection rows from a pass over the whole table, which is what makes the leftovers newly reachable.

## Reviewer caveats — stated, not implied

⚠️ **The AC2 integration test has NOT been executed.** `pnpm test` does not reach `*.int-spec.ts` and the pre-commit hook runs `smart-test --no-integration`, so `fulfillment-timeout-sweep.int-spec.ts` — which holds the **overlapping-race** assertion for the attempt pin — is written and type-checked, not proven. That assertion is the one that matters most here: a sequential test passes against an implementation with no guard at all. It will be run with `pnpm test:integration` (Docker) before merge.

Pre-commit gate green (151 suites / 2095 tests). One `/tech-review` pass applied — it caught `FulfillmentWorkTimeoutSweepPayloadV1` declared, exported from nowhere and referenced by nothing, while both its siblings are exported from the sync barrel and used by their handlers; now exported and used to type the handler's payload cast. A second pass follows.


piotrswierzy:
## Review: ✅ Approve — and the body's caveat is now resolved

**Correcting the PR body before it becomes history.** It says *"the AC2 integration test has NOT been executed… it will be run with `pnpm test:integration` (Docker) before merge."* That has now happened — **CI is 9/9 green including the Integration Tests job**, so `fulfillment-timeout-sweep.int-spec.ts` ran against real Postgres on a clean runner. The overlapping-race assertion for the attempt pin — the one that matters, since a sequential test passes against an implementation with no guard at all — is proven rather than compiled.

Worth stating plainly that this was *not* verified locally: this host has no Docker (no binary, no socket), so `pnpm test:integration` was never runnable here. CI is where it ran.

### What I went looking for, and what I found

I reviewed against four specific failure modes. The code had anticipated all four **in writing**, which is the reason this is an approve with no findings rather than a light pass.

**The timeout clock is `updatedAt`, not a dedicated `submittedAt`.** My first read was that any unrelated write silently resets it. The port names exactly this (`idleBefore`: *"the clock is 'IDLE SINCE', not 'submitted since'"*) and argues the right way round — it can only ever **delay** a reap, never accelerate one, and a work something is actively writing to is precisely one not to reap. A dedicated column would be more literal for a strictly worse failure direction. Agreed.

**`expectedAssignmentAttempt` doesn't lean on `CLAIMABLE_FROM`.** The docblock's reasoning is the sharp version: `submitted` not being claimable is *"a property of ONE caller's argument, not of the method."* Correct — a guard that depends on every future caller passing the same argument is not a guard.

**Non-blocking is the right call, and for the second reason more than the first.** Exclusions accumulate with nothing that clears them, and there is exactly one executor on the shipped topology — so a blocking timeout turns a transient outage into a permanent dead end with no in-product remedy. Recording the consequence as a forward constraint on #2395 (the re-source loop must bound its own attempts) is what stops that becoming someone else's surprise.

**`subjectRef: unaccepted[0]?.id` is stable**, not arbitrary — `findByOrderId` orders `createdAt ASC`. I checked because a flapping subject ref across sweeps would be noise on an operator surface.

Also verified: `bulk` lane argued from cost-of-starvation rather than convenience; frontier-as-query with the correctness reason stated (an offset over a self-consuming set steps over rows — this sweep's own defect, reintroduced by its pagination); the attention write composed in the worker because core's `fulfillment` leaf may not inject `orders` — and **zero entries added to either guard's allow-set**, which is the evidence the split is right rather than merely permitted; `OL_MASTER_SWEEP_LOCK_TTL_MS` is the shared convention across all ten sweep handlers, not an inconsistency.

No BLOCKING, no IMPORTANT, no SUGGESTION.

### One thing this PR usefully settles elsewhere

#2731 (the routing-decision reaper) said the two *"should probably agree on their threshold philosophy."* They now do, and I've recorded on that issue both what transfers (the single resolution path, the floor-as-safety-bound, frontier shape, lane, the all-failed-page surfacing) and what does **not**: this sweep reaps without asking the holder anything, which is safe here because the work returns to unassigned, whereas freeing a `live` routing slot re-opens a double-ship. #2731's "proven, not argued" AC stands untouched.

Issue #861, opened 2026-05-27
[PRODUCT-DESIGN] Decide durable per-destination fulfillment-notify state (OMP status-projection idempotency)


piotrswierzy:
## Decision to make

Should OpenLinker own a **durable, per-destination fulfillment-notify state** — recording "OL projected fulfillment status `X` (+ tracking) to destination connection `Y` at time `T`" in OL's own persistence — as the **single architectural seam** that:

1. Provides the source of truth for **idempotency** when pushing order-fulfillment updates to a destination OMP (capability B, `OrderFulfillmentUpdater`).
2. Lets **multiple projectors** (#837 dispatch-notify, #838 status-poll, #834 PS read-back, future OMPs) compose without re-inventing coordination per pair.
3. Drives **convergence to a desired state** via a re-drive loop — replacing the per-projector idempotency tricks each one currently has to invent locally.

And if so, what shape?

This is a **design/decision** issue, not an implementation ticket. The output is an ADR (or a `docs/specs` update) that decides (1) whether to build it, (2) the model shape, (3) the contract/migration changes, (4) how per-projector logic simplifies. A separate impl issue follows the decision.

## Why this matters more than originally framed

**Originally (filed during #858):** an idempotency cleanup — the PS adapter's "read `current_state` and skip if already in target" guard is a mild TOCTOU and re-derived per call; OL-side notify-state would make idempotency durable and survive PS-side manual edits.

**Sharper case (after #838 design):** *each projector that writes to the OMP independently needs its own correctness workaround for the absence of this seam*. Concretely, #838's design surfaced **two** correctness gates that exist purely because there's no notify-state:

1. **Push-first ordering in `ShipmentStatusSyncService`** — the service must attempt the OMP push *before* updating `Shipment.trackingNumber`, dropping `trackingNumber` from the patch if the push fails, so the next poll sees the diff and retries. This uses the `Shipment` row itself as a *de facto* projection-state flag (it's the only thing distinguishing "already projected" from "not yet"). Works in v1 because there's exactly one OMP destination and one tracking field; conflates carrier-truth with projection-truth.
2. **`>= dispatched` gate on the OMP push** — the same service must hand-code "don't fire `'shipped'` to PS before #837's `notifyDispatched` has had its turn," because the two projectors otherwise compete without coordination. The gate works by inference from `Shipment.status` rather than from a recorded answer to "have we already pushed?"

[...]

PR #3013, opened 2026-09-08
feat(shipping): surface a repeatedly-failing waybill relay instead of logging it forever

### LINKED ISSUE #2073: [TASK] Shipping — surface a repeatedly-failing lifecycle relay instead of logging it forever
## Problem / Context

When an order-lifecycle relay (mark-sent, waybill, cancel) fails against a participant, `ShipmentStatusSyncService` releases its claim so a later tick retries. That is correct and deliberate — the all-or-nothing retry choice is documented in the service with its rationale.

What is missing is the terminal case. A relay that fails **every** tick produces `logger.error` lines and a job that still reports `succeeded`. Nothing is operator-visible, nothing escalates, and the marketplace silently never learns the order shipped. The seller keeps getting asked for a tracking number.

This came out of the #1032 stress tests as **the one genuine gap** left by the existing relay design. The proposed fix at the time — a durable per-target obligation table plus a sweep plus a fourth event stream — was cut: it does not fix the bug it was justified by, and it is #861's scope (a `[PRODUCT-DESIGN]` issue owed its own ADR). What remains is a monitoring deficit that was being priced as a distributed-systems one.

## Proposed Solution

Count consecutive relay failures per shipment and make the terminal state visible. Smallest shape that works:

- A counter column on `shipments` (incremented on relay failure, reset on success)
- After N consecutive failures, escalate: mark the poll job `business_failure` with an `outcomeReason` (per [ADR-007](../docs/architecture/adrs/007-syncjob-status-vs-outcome-split.md) — the status/outcome split already exists for exactly this), and/or surface the shipment in an operator-facing needs-attention bucket
- Keep the existing release-on-failure retry behaviour unchanged

Files:
- `libs/core/src/shipping/application/services/shipment-status-sync.service.ts`
- `libs/core/src/shipping/infrastructure/persistence/repositories/shipment.repository.ts`
- `libs/core/src/shipping/infrastructure/persistence/entities/shipment.orm-entity.ts` (migration)

Deliberately **not** in scope: a per-target obligation table, a relay event stream, per-participant retry. Those are #861 and need its ADR first.

## Classification

**Type**: CORE
**Layer**: Application / Infrastructure
**File(s)**: `libs/core/src/shipping/application/services/shipment-status-sync.service.ts`

## Dependencies

None. Explicitly **not** blocked on #861 — this is the monitoring half, which stands alone.

## Assumptions

- Order-grain (per shipment) escalation is enough. Per-participant attribution — knowing *which* destination is failing — is #861's job; a first version can name the failing participant in the message without persisting per-target state.
- N is small and configurable; the right value depends on poll cadence.

## Acceptance Criteria

- [ ] Consecutive relay failures are counted per shipment and reset on success
- [ ] After the threshold, the condition is operator-visible (not only in logs)
- [ ] Existing release-on-failure retry behaviour is unchanged below the threshold
- [ ] A transient single failure does not escalate
- [ ] Migration has both `up()` and `down()`
- [ ] Unit test for the counter transitions; integration test for the escalation

piotrswierzy:
Closes #2073

## The defect

`ShipmentStatusSyncService` releases its at-most-once claim on any transient participant failure so a later poll tick retries (#1947, deliberate). That had **no terminal case**: a relay failing on EVERY tick produced one `logger.error` per tick against a job still reporting `succeeded`, while the marketplace silently never learned the order shipped and the seller kept being asked for a tracking number. The classification existed; only the surface was missing.

## What lands

**The count is folded INTO `releaseWaybillRelay`, as one statement** — so a release that does not count is not *expressible*, and it adds no new failure mode (both callers already awaited that release un-caught). Five columns on `shipments` carry consecutive failures, the last failure's instant, its reason and the participant it was for.

**A successful relay clears the history through a separate best-effort write.** The relay has already landed durably, so failing it because its bookkeeping failed would let the recording of a success destroy the success.

**Escalation is a COUNT, never elapsed time** — cadence-independent, where "stuck for two hours" would mean a different number of attempts per deployment. The threshold is clamped to a minimum of 2, so a single transient blip can never raise a signal: unreachable rather than merely not the default.

**Operator surface**: a `Tracking not sent` badge in the `/shipments` status cell — beside the status badge, never inside the severity word, since a *delivered* shipment can carry one — plus `?waybillRelayStuck=true` as the needs-attention bucket. The backend derives `stuck`, so `apps/web` holds no copy of the rule and no `check:invariants` mirror is needed.

## Two deliberate non-changes

**Not `outcome: 'business_failure'` on the poll job.** It sweeps a page of many shipments, the condition is fixed by a re-auth rather than being terminal, and the retry IS the recovery — ADR-007's semantics are the opposite of what this needs.

**Per-destination notify state remains #861 and is not pre-empted.** The relay's control flow is untouched, and the recorded participant is a display field no retry decision reads.

## Verification

48 files, +1882/−21. Includes a migration (nullable columns + a partial index), declared on the ORM entity too since the integration harness builds schema by `synchronize`.

⚠️ **Integration tests have not been executed.** `*.int-spec.ts` is outside the api project's `testRegex` and its `tsconfig.type-check.json` excludes `test`, so `pnpm test` and `pnpm type-check` do not reach them. They will be run with `pnpm test:integration` (Docker) before merge.

### DIFF
--- a/apps/api/.env.example
+++ b/apps/api/.env.example
@@ -596,3 +596,32 @@ OL_PII_HASH_SALT=your-org-level-salt-change-in-production
 # (30 days).
 # OL_RESERVATION_OBLIGATION_MAX_AGE_MS=2592000000
 
+
+# --- Waybill relay: escalation threshold (#2073) ------------------------------
+#
+# How many CONSECUTIVE failed attempts to relay a shipment's waybill to the
+# order's participants before the shipment is escalated as needing attention:
+# a `Tracking not sent` badge on /shipments, and the `?waybillRelayStuck=true`
+# filter that collects them.
+#
+# Why this exists: a relay that fails on every poll tick releases its claim so
+# the next tick retries (#1947, correct) — but before #2073 that produced one
+# error log per tick, a job that still reported `succeeded`, and nothing an
+# operator could see, while the marketplace silently never learned the order
+# shipped and kept asking the seller for a tracking number.
+#
+# A COUNT rather than an elapsed time, deliberately: the poll cadence decides
+# how many attempts fit in an hour, so "stuck for two hours" would mean a
+# different number of failed attempts on every deployment.
+#
+# Read by the **api** process only — the worker increments the counter and
+# never compares it, which is why this variable does not appear in
+# apps/worker/.env.example. The count itself is durable on `shipments` and is
+# reset by the first relay that succeeds.
+#
+# Clamped to [2, 100]. The LOWER bound is load-bearing: a threshold of 1 would
+# escalate on a single transient blip, so it is unreachable rather than merely
+# not the default. An absent, blank, non-numeric or non-finite value falls back
+# to the default rather than throwing.
+# Default 3.
+# OL_WAYBILL_RELAY_FAILURE_ALERT_THRESHOLD=3
--- a/apps/api/src/migrations/1876000000000-add-shipment-waybill-relay-failure-tracking.ts
+++ b/apps/api/src/migrations/1876000000000-add-shipment-waybill-relay-failure-tracking.ts
@@ -1,0 +1,82 @@
+import type { MigrationInterface, QueryRunner } from 'typeorm';
+
+/**
+ * Waybill-relay failure tracking on `shipments` (#2073).
+ *
+ * Five columns, one CHECK and one partial index so that a lifecycle relay
+ * failing on every poll tick becomes a durable, operator-visible fact instead
+ * of one `logger.error` per tick against a job that still reports `succeeded`.
+ *
+ * Every column mirrors its `@Column` declaration on `ShipmentOrmEntity` exactly,
+ * and the CHECK and the index are declared there under these same names. That
+ * duplication is required rather than redundant: the integration harness builds
+ * schema by TypeORM `synchronize` and never runs migrations, so a constraint
+ * living only here would hold in production and silently not in tests.
+ *
+ * `TIMESTAMP` without time zone, matching `waybillRelayedAt` (migration
+ * `1832000000007`) and `reservationConsumedAt` on the same table. The repo is
+ * mixed on this, so the rule followed is consistency WITHIN the table — copying
+ * `webhook_auth_rejections`' `timestamptz` would have diverged the two schema
+ * sources for `shipments`.
+ *
+ * `waybillRelayFailureCount` keeps its `DEFAULT 0` rather than dropping it after
+ * backfill (the way `direction` drops its default in `1862000000000`), and the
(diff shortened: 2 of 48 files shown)


piotrswierzy:
## Review: ✅ Approve

**First, correcting the PR body before it becomes history.** It says *"Integration tests have not been executed… they will be run with `pnpm test:integration` (Docker) before merge."* CI has now run them — **9/9 green including the Integration Tests job** — so the migration adding the five failure columns and its partial index have actually been applied, not just compiled. Stating plainly that this did not happen locally: this host has no Docker at all, so that suite was never runnable here.

### The change is well-shaped where it matters most

**The count is folded into `releaseWaybillRelay` as one statement**, and the signature makes the failure record required — so *a release that does not count is not expressible*, which is a stronger guarantee than a convention two callers have to remember. The statement is parameterised throughout (no interpolation), `COALESCE`s `firstFailedAt` so it marks the start of the current run rather than being overwritten each attempt, and bumps `updatedAt` explicitly with the reason stated: raw SQL does not fire `@UpdateDateColumn`, and the `repository.update()` call it replaces did bump it. That last detail is the kind of thing that silently changes behaviour if missed.

**Clearing on success is a separate best-effort write**, correctly: the relay has already landed durably, so failing it because its bookkeeping failed would let the recording of a success destroy the success.

**Escalation is a count, not elapsed time** — cadence-independent, where "stuck for two hours" means a different number of attempts per deployment. The threshold clamps to a floor of 2, which makes single-blip escalation *unreachable* rather than merely not the default (AC-4). And the resolver is the single path feeding both the filter and the badge, so the two cannot disagree about what "stuck" means.

**The two deliberate non-changes are both right.** Not `outcome: 'business_failure'` on the poll job — it sweeps a page of many shipments, the condition clears on a re-auth rather than being terminal, and the retry *is* the recovery, so ADR-007's semantics are inverted here. And #861 is not pre-empted: the relay's control flow is untouched and the recorded participant is a display field no retry decision reads.

### One note I raised and am withdrawing

I flagged that `toWaybillRelayFailure` gates on `entity.waybillRelayFailureCount <= 0` rather than `typeof … === 'number' && Number.isFinite(…)`, on the theory that an `undefined` would fall through and build an object with `count: undefined`.

It doesn't hold. The column is `integer NOT NULL DEFAULT 0`, so `null` cannot arrive from the database and every pre-migration row backfills to `0`. The only path to `undefined` is a `select`-narrowed read, which no caller does — `toDomain` is fed full entities. And the `docs/lessons.md` rule I had in mind is scoped to *"a numeric value read off untyped wire JSON"*, which an ORM column read is not. Recording the check rather than the suspicion, since the reasoning is worth having next to the code.

No BLOCKING, no IMPORTANT, no SUGGESTION.

Issue #3002, opened 2026-09-08
[BUG] DX — the mirror guards' array parser truncates on a `]` inside a comment, and an empty parse passes vacuously


piotrswierzy:
## Problem / Context

`scripts/check-sales-document-reason-mirror.mjs` — and every guard copied from it — locates the array's closing `]` in **raw source**, and strips comments only **afterwards**:

```js
// scripts/check-sales-document-reason-mirror.mjs:92-100
const openBracket = declMatch.index + declMatch[0].length - 1;
const closeBracket = content.indexOf(']', openBracket);   // ← raw source
if (closeBracket === -1) return null;

const body = content
  .slice(openBracket + 1, closeBracket)
  // Strip comments so annotating an entry can't be read as a value.
  .replace(/\/\/[^\n]*/g, '')
  .replace(/\/\*[\s\S]*?\*\//g, '');
```

So a `]` appearing in a per-entry docblock **between** the `[` and the real closing bracket truncates the scan. If that `]` precedes every string literal, `values` comes back **empty**.

That is not hypothetical: it fired while writing a new sibling guard, on a union whose member was documented with `destinationConnectionIds: []`.

### The empty parse is the serious half

There is no emptiness guard anywhere in the comparison. `compareReasonValues` is pure set-difference plus an order check:

```js
const missingInFrontend = backend.filter((v) => !frontendSet.has(v));   // [] when both empty
const missingInBackend  = frontend.filter((v) => !backendSet.has(v));   // []
if (issues.length === 0 && backend.join('|') !== frontend.join('|'))    // '' === ''
return { ok: issues.length === 0, issues };                              // → ok: true
```

**If a `]`-in-comment truncates BOTH sides, the guard reports success having compared nothing.** One side truncating is loud but misleading — it surfaces as *"every value is missing from core"* rather than as a parse error, which sends the reader to the union rather than to the parser. Both sides truncating is silent.

This is the same failure family as #2792 (*"Three check-* guards go silently inert … floor the scanned count"*), reached by a different mechanism: there the walk finds no files, here the parse finds no values. #2792 does not cover the mirror scripts, and its remedy — floor the count — is the remedy here too.

## Blast radius

[...]

Issue #2397, opened 2026-08-22
[TASK] CORE Application — router-filtered destination fan-out (`OrderSyncRequest.destinationConnectionIds`)


piotrswierzy:
> Part of the OMS programme backlog (**Wave 3a**, stream S1, size S). Backlog slug: `W3a-8`. Sibling `W3a-*` slugs resolve via the Wave-3a epic's checklist.

## Problem / Context

Design §5.5: `OrderSyncService` is **retained, not reinterpreted** — destination-mirror creation is a commercial/catalogue act distinct from fulfilment assignment. Under a router the fan-out becomes filtered.

## Proposed Solution

Optional `destinationConnectionIds` on `OrderSyncRequest`, **defaulting to today's behaviour** (absent ⇒ full fan-out). A filtered-out destination gets no `syncStatus[]` entry *by design*; record in the service header why that is safe: for routed orders `fulfillmentState` is fed by work-progress-derived shipments through `ShipmentDispatchService`, and the reservation publish-subtraction never depends on `fulfillmentState` at all (design §3 adjudication #1).

## Classification

**Type**: CORE
**Layer**: Application
**File(s)**: `libs/core/src/orders/application/services/order-sync.service.ts`

## Dependencies

- #2396 (`W3a-7`)

## Acceptance Criteria

- [ ] Absent field ⇒ byte-identical fan-out (characterisation test)
- [ ] A filtered destination produces no `syncStatus[]` entry and no branch-1 status-sync attempt
- [ ] Rationale recorded in the file header
- [ ] Tests added or updated for non-trivial logic
- [ ] No architecture boundary violations (CORE ↔ Integration)

PR #3014, opened 2026-09-08
fix(orders): persist why a routing decision narrowed the destination fan-out

### LINKED ISSUE #2703: /orders cannot distinguish "the router sent this nowhere" from "nothing is configured"
Split out of **#2397** (merged). Recorded there at the branch and in the PR body; filed here because it is a real operator-facing gap, not an implementation detail.

## The gap

#2397 made a deliberate empty routing decision a **non-event**: `resolveDestinations` returns `[]`, no `syncStatus` row is written, and nothing is enqueued. That is correct — the alternative was ten live marketplace hydrations and a dead-lettered order blaming connection configuration for what the router chose.

But the two conditions are **distinguishable at the branch and identical on the operator surface**:

| Condition | Branch | `/orders` |
|---|---|---|
| (a) nothing configured / all inactive | throws, as today | "No destinations" |
| (b) router deliberately named nobody | warn + `[]`, no rows | "No destinations" |

Both render through `rollup.total === 0 → 'unknown'`.

So an operator looking at an unrouted order cannot tell **a working router making a decision** from **a broken or empty configuration** — which is precisely the distinction routing exists to make. A router that is quietly sending everything nowhere looks exactly like a router that was never set up.

## Why this reading is already settled elsewhere

Order ingestion adjudicated the same question for #2339 and reached the opposite conclusion for the same reason: a state the operator cannot tell apart from "nothing happened" is the one that must be made visible. `skipped_cancelled` and `skipped_held` exist as **distinct arms** rather than as silence, so a consumer cannot route a non-event into a retry *and* an operator can still see why.

The #2100 precedent is the closer match: every block on the auto-issue path is **persisted and operator-visible, never log-only**, because a silent decline is indistinguishable from a system that simply did nothing.

## Scope

1. Give a deliberately-unrouted order a persisted, operator-visible reason — most likely a distinct `syncStatus` arm, so it reads as a decision rather than an absence. Note the constraint: #2397 writes **no** `syncStatus` row today, and adding one must not reintroduce anything a consumer could route into a retry.
2. Surface it on `/orders` distinctly from "No destinations".
3. Follow #2100's discipline: the badge, any aggregate count and any filter must agree, or the count names rows whose badges do not explain it.

## Acceptance criteria

- [ ] (a) and (b) render differently on `/orders`.
- [ ] The rendered reason is derived from persisted state, not re-derived client-side.
- [ ] No consumer can route the deliberate-empty state into a retry (the property #2397 bought).
- [ ] If a count or filter is added, its rows carry the matching badge.

## Related

- **#2397** — introduced the state and recorded the gap.
- **#2100** — every auto-issue block is persisted and visible; the aggregate/filter/badge agreement rule.
- **#2339** — the same reading, adjudicated for order holds.
- Sibling gap from the same slice: [the disabled-routed-destination regression].

### LINKED ISSUE #2704: A routed order whose single destination is momentarily disabled becomes a total failure
Split out of **#2397** (merged), where it is recorded at the filter and in the PR body. A genuine **new regression** introduced by router-filtered fan-out, accepted deliberately rather than overlooked.

## The regression

`listCapabilityAdapters` is **active-only**. So once a routing decision names specific destinations:

- **Today (unfiltered):** an order whose `dest-a` is disabled still fans out to `dest-b` and `dest-c`. Partial success.
- **After #2397 (filtered to `dest-a`):** nothing is eligible, and the order becomes a **total failure**.

This is condition (c) of the three that an empty destination list now conflates. #2397 makes it throw rather than return `[]` — deliberately, because the router *asked for* a destination OpenLinker could not reach, and the retry that follows is appropriate since a disabled connection can be re-enabled. The exception carries the source-echo-excluded ids so the message names which destinations were unreachable.

## Why it still needs an owner

The throw is the right immediate behaviour, but the underlying question is unresolved: **should a routing decision naming an unavailable destination fall back, or hold?**

Both answers are defensible and they are not the same product:

- **Hold** (today) is correct when the router's choice is meaningful — shipping from the wrong warehouse is worse than shipping late.
- **Fall back** is correct when the destinations are interchangeable, which is exactly the topology the unfiltered fan-out assumed and every existing install has.

Deciding it needs the router's own semantics, which is `@openlinker/oms` territory (#2408) rather than the fan-out's.

## Scope

1. Decide whether a routing decision expresses a **requirement** or a **preference** — possibly per rule, since both are legitimate.
2. If preference is supported, define the fallback order and make it observable: a silent substitution of destination is worse than the failure it replaces.
3. Either way, ensure an operator can see *why* an order failed — that a named destination was disabled, not that "no destinations are available", which is what the pre-#2397 message would have implied.
4. Consider the interaction with retry: a disabled connection is re-enableable, so the retry ladder is doing useful work here — unlike the deliberate-empty case, which #2397 correctly made a non-event.

## Acceptance criteria

- [ ] The requirement-versus-preference semantics are decided and documented where the router's rules are authored.
- [ ] An order failing because its routed destination is disabled says so, distinctly from "nothing configured".
- [ ] If fallback ships, every substitution is recorded — never silent.

## Related

- **#2397** — introduced it; the reasoning is at the filter.
- **#2408** (`W3a-17`) — the OL router's named filters and sorts; the natural owner of the semantics.
- Sibling gap from the same slice: **#2703** (`/orders` cannot distinguish a deliberate empty routing decision from an unconfigured one).

piotrswierzy:
Closes #2703
Closes #2704

#2397 left two operator-facing costs recorded in the architecture doc. This closes both, in one branch because they are two halves of one defect sharing `OrderSyncService.resolveDestinations`.

## #2704 — a partial-eligibility case existed and was silent

`NoOrderDestinationsAvailableException` is reached only when **nothing** resolved, so a decision naming `[a, b]` of which only `b` was eligible proceeded to `b` alone behind a single `logger.warn`: an order under-provisioned with nothing outside the log to say so. `resolveDestinations` now reports its unresolved ids whether or not any resolved, and the partial case is persisted as `routed-destinations-partially-unavailable`.

**The throw for the none-resolved case is KEPT and no fallback is invented.** A routing decision is a requirement rather than a preference, so mirroring to a destination the router did not choose would be a shipment OL was never asked to make. Arm (b) — a router that positively chose nobody — still returns `[]` and never throws.

## #2703 — the reason is persisted on the order

`order_records.destinationRoutingBlockReason` + `destinationRoutingBlockDetail`, the **#2100 shape** rather than a `syncStatus` arm: arm (b) has no destination id to key a row on, and a new arm is exactly what a consumer could route into a retry.

Four reasons. `routed-to-no-destination` is a *working router making a decision*, so it renders a neutral badge and is excluded from the counted subset — derived by `.filter`, never hand-listed. The aggregate counts against an explicit IN-list, never `IS NOT NULL`, and is ANDed beside `health` rather than widening `OrderHealthValues`, which is a partition.

The write is level-triggered at the single point where the observation exists, so it cannot convert a deliberate empty decision into a throw; both columns are excluded from `toOrm` and written by one guarded UPDATE; and `OrderIngestionService` clears the reason on its held branch, where `syncOrder` never runs and a stale reason would otherwise sit beside a fresh `fulfillmentBlockReason`.

`detail` carries connection ids and counts only, bounded to ten listed with the total always stated — the ids come from a decision OL does not author and the string is rendered verbatim.

## A real defect found in the guard it copied

The FE mirror (`apps/web` cannot import `@openlinker/core`, #591) is held by `scripts/check-destination-routing-reason-mirror.mjs` under `pnpm check:invariants`. Writing it surfaced a bug in the **sibling** parser it was copied from: that one locates the array's closing `]` *before* stripping comments, so a `]` inside a per-entry docblock makes it read the whole array as EMPTY — and both sides then pass vacuously. Fixed here and pinned by a self-check case; the sibling is left alone, as it passes on its own inputs. Filed separately as **#3002**.

## Blast radius

Every branch is **latent on every install**: no production caller populates `OrderSyncRequest.destinationConnectionIds`, so both columns are NULL everywhere and the change is behaviour-preserving until a router is wired.

37 files, +1844/−40. Adds migration `1875000003400` (two nullable text columns + a partial index on `IS NOT NULL`, so the index cannot go stale when the union widens), declared on the ORM entity too since the integration harness builds schema by `synchronize`.

⚠️ **Integration tests have not been executed** — `*.int-spec.ts` is outside the api project's `testRegex` and its `tsconfig.type-check.json` excludes `test`. They will be run with `pnpm test:integration` (Docker) before merge, which matters here because of the new migration and the guarded UPDATE.


piotrswierzy:
## Review: ✅ Approve — one IMPORTANT finding, fixed in `023806956`

**Correcting the PR body first.** It says the integration tests *"will be run with `pnpm test:integration` (Docker) before merge"*. They have been — **CI is 9/9 green including Integration Tests** — so migration `1875000003400` has actually been applied and the guarded UPDATE exercised, not merely compiled. That did not happen locally: this host has no Docker at all.

### The finding: the mirror guard fixed only half of #3002

The new `check-destination-routing-reason-mirror.mjs` correctly fixed the **comment-order** half — its `parseReasonValues` blanks comments *before* locating the array's closing `]`, so a `]` inside a per-entry docblock no longer truncates the scan, and a self-check case pins exactly that input. Good, and it is a genuine improvement over the sibling it was copied from.

But `diffReasonValues([], [])` returns `ok`, and the fatal block tested only `!parsed` — never `parsed.values.length === 0`. Two empty parses would print `✓ 0 reason value(s) identical` and **exit 0 while asserting nothing**: one green reading for "not covered" and "covered and passing", which is what #2673 rejects. And that is precisely the state the truncation bug produces — so fixing one half without the other leaves the guard able to pass vacuously in exactly the scenario it was hardened against.

Fixed in the shape `check-attention-reason-mirror.mjs` already uses (`need` / `needEntries`, credited to the same #2673 review), and **verified red-first**: emptying `DestinationRoutingBlockReasonValues` now exits 1 naming the file and line, where before the change it passed. The self-check gains a case asserting the pure diff *is* vacuous on two empty inputs — recording that the vacuity is by design at that layer and the call-site guard is what covers it, so nobody removes one believing the other handles it.

### What I verified and found sound

**The migration.** Idempotent DDL, correct reverse-order `down()`, timestamp ordered after main's `…2000`, and the timestamp guard passes. The partial index is on `IS NOT NULL` rather than a value list — explicitly to avoid the trap `IDX_order_records_salesDocumentBlockReason` fell into when #2248 widened that union and left the index behind. A NULL predicate cannot go stale as the vocabulary grows. Declared on the ORM entity too, because the harness builds schema by `synchronize` and a migration-only index would hold in production and silently not in tests.

**The counted subset is derived, not restated.** `.filter` over the full list, and the SQL twin builds its IN-list from that same array — so the aggregate and the union cannot drift, and `routed-to-no-destination` (a working router making a decision) stays out of the counted set by construction rather than by a hand-maintained list.

**Writer discipline.** Both columns excluded from `toOrm`, one narrow guarded UPDATE, `IS DISTINCT FROM` so a no-change write is a no-op, and read-path coercion through `isDestinationRoutingBlockReason` so a value written by a newer release and rolled back reads as absent rather than widening the union at runtime.

**Mutual exclusivity with `fulfillmentBlockReason` is enforced, not assumed.** `OrderIngestionService` clears the routing reason on the held branch — without it a stale routing reason would sit beside a fresh fulfilment block and assert "the router sent this nowhere" about an order that is merely held.

**The throw is kept for the none-resolved case and no fallback is invented** — a routing decision is a requirement, not a preference, so mirroring to a destination the router did not choose would be a shipment OL was never asked to make.

Every branch is latent on every install (no production caller populates `OrderSyncRequest.destinationConnectionIds`), so this is behaviour-preserving until a router is wired.

piotrswierzy (commented):

*(Your own PR, so GitHub permits only a comment.)* ⚠️ **CI: the `Test` job is failing** — that needs resolving independently of the below.

The backend is excellent and unusually disciplined. **All four axes I checked specifically hold**, verified line by line rather than by silence:

1. **#2100 persistence discipline — fully satisfied.** Both columns absent from `toOrm` (documented in both docblocks alongside the `salesDocument*` / `fulfillmentBlock*` / `omsAttention` precedents); one narrow guarded writer with an `IS DISTINCT FROM` no-op guard that correctly keeps the `updatedAt` bump off the `null → null` path — which matters because `FulfillmentStatusSyncService` scans `updatedSince`; coerced on read, never cast; explicit IN-list built by `.filter` rather than hand-listed. The `COALESCE(…, '')` is genuinely load-bearing and correctly explained — without it `NOT (NULL IN (…))` is NULL and `WHERE` drops the row, so `destinationRoutingBlocked=false` would return zero orders. The predicate is a `static readonly` shared by filter and count, and `predicate-parity.spec.ts` gained all three cases, so the chip's number and its rows are structurally unable to disagree.
2. **The #2397 three-state contract is preserved.** `undefined` still short-circuits before any filtering, so it can never collapse into `[]`. The widening of `unresolvedRequestedIds` does **not** change the exception path: the throw is reached only when `destinations.length === 0`, where the pre-#2704 code already returned the ids, so the (a)-vs-(c) discrimination is byte-identical. Arm (b) still returns `[]` and never throws — achieved by placing the write **above** all three branches.
3. **Migration correct on all three rules** — `1878000000000` > main's tail `1877000000000`, prefix unique across `apps/api` and both Allegro plugin dirs, class suffix matches filename. `up()`/`down()` both idempotent, `down()` drops in reverse order. The index is declared on the ORM entity **as well as** the migration — right call, since the harness builds schema by `synchronize` and a migration-only index would hold in production and silently not in tests.
4. **Cross-context imports clean.** No `*RepositoryPort` / `*OrmEntity` / deep sub-path crosses a boundary; the FE mirror correctly does not import core (#591).

The concern is **testing, and one surface the docs misstate**.

### IMPORTANT — `deriveDestinationRoutingBlock` has zero tests

`destination-routing-block.types.spec.ts` — `grep -n deriveDestinationRoutingBlock` against the spec returns nothing. The spec covers the guard, the detail builder and the counted subset, but not the rule that is the entire subject of the PR. Its own docblock says why that's a defect: *"PURE — no I/O, no clock, no injected dependency — so the rule can be asserted exhaustively without a service."* Five distinguishable outcomes, none exercised:

- `undefined` → `null` (the every-install path)
- `[]` → `routed-to-no-destination`
- `resolvedCount === 0 && unresolved.length === 0` → `routed-to-source-only`
- `resolvedCount === 0 && unresolved.length > 0` → `routed-destinations-unavailable`
- `resolvedCount > 0 && unresolved.length > 0` → `routed-destinations-partially-unavailable`

The last is the whole of #2704; the first is the three-state inversion the PR body itself calls the thing that "silently stops destination provisioning everywhere". Nothing in the repo would catch it. Highest-value fix in the list.

### IMPORTANT — the service's new write has no assertions

`order-sync.service.spec.ts` (~137) — the only change is `markDestinationRoutingBlock: jest.fn()` so existing tests stop throwing. Nothing asserts the three properties the implementation's own comments call load-bearing: that the write happens **before** `NoOrderDestinationsAvailableException` is thrown (the point of #2704), that arm (b) persists `routed-to-no-destination` while still returning `[]`, and that a persistence failure is swallowed rather than masking the real exception. All three are cheap against the existing mock.

### IMPORTANT — the claimed per-order badge on `/orders` does not exist

Four places assert it does: `docs/architecture-overview.md` (*"It renders a neutral per-order badge"*), `destination-routing-block.types.ts`, the new chip comment in `orders-list-page.tsx`, and `destination-routing-copy.ts` (*"Short badge label"*). But `git grep destinationRoutingBlockReason -- apps/web` shows exactly one render path: `order-detail-page.tsx` → `OrderHealthSummary` → `syncCellLabel`. The list row's badge comes from `deriveOrderHealth`, untouched here.

So clicking the new `Routing blocked` chip yields rows with **no visible indication of why they matched** — the exact failure `architecture-overview.md § Invoicing` records for #2100 and quotes verbatim: *"a block the backend keeps but the surfaces hide is a number whose rows explain nothing"*. And `routed-to-no-destination` — the case #2703 exists to make distinguishable — is neither counted, filterable, nor badged; visible only by opening the order. Either build the badge (the field already reaches the list payload via the shared `toDto`, the `activeHoldReason` precedent) or correct all four claims to say detail-only.

### IMPORTANT — `DESTINATION_ROUTING_ATTENTION_REASONS` has no consumer, so `tone` is write-only

Declared, not exported from `features/orders/index.ts`, read nowhere. Since `tone` exists only to derive it, the "two tones, and the split is the whole point of #2703" section describes a distinction computed and never rendered — on the detail page both reasons produce identical markup, because `alarm` keys on `rollup.failed > 0`, which is `0` on every empty rollup. Filtering is server-side, so the FE constant duplicates a decision it never applies. The comment at `orders.types.ts:165` also points a reader at a constant that counts nothing.

### IMPORTANT — no frontend test ships at all

Seven `apps/web` files, zero `.test.ts(x)`. `stock-at-risk-copy.ts` — the precedent this module's header cites **by name** — has a test beside it; `destination-routing-copy.ts` has none, so the unrecognised-value → `null` contract its header calls load-bearing is unasserted. The sibling `invoicing=blocked` chip has four cases in `orders-list-page.test.tsx` (offset reset, toggle, composition with `health`, deep-link); `routing=blocked` has none.

### IMPORTANT — the partial index probably cannot serve its own predicate

The migration docblock says *"this slice does filter and count, so the index has a consumer."* The index is `WHERE "destinationRoutingBlockReason" IS NOT NULL`; the only predicate is `COALESCE(rec."destinationRoutingBlockReason", '') IN (…)`. Postgres' predicate-implication prover does not reason through `COALESCE(x,'') IN (…)` ⇒ `x IS NOT NULL`, the negated arm cannot use it at all, and the count is a `COUNT(*) FILTER` aggregate no index on this column helps. The stated reason for choosing NULL over a value list (avoiding `IDX_order_records_salesDocumentBlockReason`'s staleness after #2248) is sound — but the fix preserving both properties is to index the bare column, or derive the partial predicate from `DestinationRoutingAttentionReasonValues` as `IS_DESTINATION_ROUTING_BLOCKED` already does. Zero rows today, so not urgent — but the docblock's claim shouldn't outlive the review.

### SUGGESTIONS

- A **third** ingestion branch reaches neither write: the item-resolution failure path (`markItemResolutionFailure` → `throw MissingOrderItemMappingError`) returns before both, so an order carrying `routed-destinations-partially-unavailable` that then loses a product mapping keeps that reason indefinitely while sitting `awaiting_mapping`. Latent today and arguably defensible (the sibling `fulfillmentBlockReason` has the identical gap) — but it belongs in the docblock's list of covered branches.
- Three separate import statements now pull from `@openlinker/core/orders` in `order-record-response.dto.ts` (14–18); merge them.
- The PR body says *"Adds migration `1875000003400`"*; the merged file is `1878000000000`. The migration's docblock explains the renumber — the body wasn't updated.

### Also verified

**The invariant script is sound and the bug it found is real.** I ran `--self-check` and the real comparison against the three files in an isolated scratchpad — both pass. The divergence from `check-sales-document-reason-mirror.mjs` is load-bearing: the sibling locates the closing `]` in raw source *before* stripping comments, and this union's docblocks contain `destinationConnectionIds: []`, so the sibling's parser would read the array as empty and both sides would compare **vacuously equal**. The fix blanks comments to equal-length whitespace so line numbers survive, adds an emptiness guard so a vacuous parse is fatal rather than a green "0 identical", and pins the bracket-in-comment case. Filing the sibling separately as #3002 rather than editing a shipped invariant is the right risk call.

`buildDestinationRoutingDetail` is bounded and states its own total (ten ids, remainder summarised, count first), so truncation can't hide scale and a malformed router payload can't write an unbounded `text`. PII-free by construction.

---

piotrswierzy (commented):
## Tech Lead review

*(Posted as a comment, not an approval — GitHub forbids approving your own PR.)*

Second pass at `1380cde`, the same commit the existing `/pr-review` covered, so nothing has changed since it. I re-verified the four axes the backend rests on and they hold — I won't restate them. Two things are **new** since that review, and the six IMPORTANT findings it raised are all still open.

### Findings

**[NEW — IMPORTANT]** — `package.json`
> The branch now **conflicts with `main`** (`mergeStateStatus: DIRTY`). `git merge-tree origin/main <head>` reports exactly one conflict, in `package.json` — the `check:invariants` chain, where this PR appends `check-destination-routing-reason-mirror.mjs` and `main` has since appended `check-adr-numbers` (#3036). Trivial to resolve, but resolve it by **intent** (both scripts must survive in the chain), not by taking a side — and re-run `pnpm check:invariants` afterwards, since a resolution that drops a link is silent.

**[NEW — INFORMATIONAL]** — CI `Test` job
> The failing required check is a **35-minute job timeout, not an assertion failure**. Every package that reported before the cut is green — `libs/core` 419/419 suites, 6376/6376 tests; `allegro` 703; `ksef` 402; `erli` 503; etc. The log ends mid-`libs/integrations/woocommerce` with no `✕`, no `Tests: N failed`, and no `SIGKILL`. `apps/web`, `apps/api`, `apps/worker`, `woocommerce` and `prestashop` never reported at all. So there is **no evidence of a code defect in this diff** — but equally no evidence `apps/web` passes, which matters because seven `apps/web` files changed. Re-run the job; if it times out again the cause is the runner, not this branch. (`Integration Tests` is separately green at 35m21s, so the new migration `1878000000000` and the guarded UPDATE *have* been applied against real Postgres — the PR body's ⚠️ caveat is stale.)

**[STILL OPEN — IMPORTANT ×6]** — carried from the `/pr-review` at this same commit, re-verified by grep against `origin/pr-3014`:

1. `libs/core/src/orders/domain/types/destination-routing-block.types.spec.ts` — `deriveDestinationRoutingBlock` still has **zero** occurrences in its own spec. Five distinguishable outcomes, none exercised, on a function the file's own docblock calls exhaustively assertable because it is pure. Highest-value fix.
2. `libs/core/src/orders/application/services/__tests__/order-sync.service.spec.ts:137` — `markDestinationRoutingBlock` is still only a bare `jest.fn()`. None of the three load-bearing properties (write happens *before* the throw; arm (b) persists and still returns `[]`; a persistence failure is swallowed) is asserted.
3. The per-order **badge on `/orders` still does not exist** — `destinationRoutingBlockReason` reaches `apps/web` only via `order-detail-page.tsx → OrderHealthSummary`. Four places (`docs/architecture-overview.md`, `destination-routing-block.types.ts`, the chip comment in `orders-list-page.tsx`, `destination-routing-copy.ts`) claim otherwise. The new `Routing blocked` chip therefore filters to rows that show no reason — the exact "a number whose rows explain nothing" failure `§ Invoicing` records for #2100.
4. `DESTINATION_ROUTING_ATTENTION_REASONS` (`destination-routing-copy.ts:72`) still has no consumer and is not exported from the feature barrel; `tone` remains write-only.
5. **Zero `apps/web` tests** ship — `git show --name-only` matches no `apps/web/**/*.test.tsx`. The sibling `invoicing=blocked` chip has four cases; `routing=blocked` has none, and `destination-routing-copy.ts`'s unrecognised-value → `null` contract is unasserted despite `stock-at-risk-copy.ts` — cited by name in its own header — having a test beside it.
6. The partial index `WHERE "destinationRoutingBlockReason" IS NOT NULL` still cannot serve `COALESCE(rec."destinationRoutingBlockReason", '') IN (…)`; the migration docblock's "the index has a consumer" claim is still there.

### Re-verified sound (not restated in detail)

- **ADR-053 no-injection invariant**: the diff touches **no** file under `libs/core/src/fulfillment`, and adds no entry to either guard's allow-set. Order data still enters as arguments.
- **The #2397 three-state contract**: `resolveDestinations` still discriminates `undefined` from `[]` explicitly (`requestedDestinations !== undefined && requestedDestinations.length === 0`), so absent can never collapse into present-and-empty. Arm (b) returns `[]` above the throw.
- **Single-writer / `toOrm`-omission discipline**: both columns excluded from `toOrm` with the omission documented, one narrow guarded UPDATE with `IS DISTINCT FROM`, read-path coercion via `isDestinationRoutingBlockReason`.
- The mirror-guard vacuity fix (`parsed.values.length === 0` fatal, `need`/`needEntries` shape) **did land** — that earlier thread is closed.

### Verdict
🔄 **Approve with changes**

1. Resolve the `package.json` conflict with `main` (both invariant scripts survive), then re-run `pnpm check:invariants`.
2. Re-run the timed-out `Test` job; confirm `apps/web` actually passes.
3. Test `deriveDestinationRoutingBlock` (finding 1) and the service write (finding 2) — cheapest, highest value.
4. Either build the list-row badge or correct all four claims to say detail-only (finding 3); fold findings 4/6 in with it.

---

piotrswierzy (commented):
## Tech Lead re-review (head unchanged at `1380cde`)

*(Posted as a **comment, not an approval** — this PR is authored by the account I am reviewing from, and GitHub forbids self-approval. Substantively the verdict is below.)*

Re-verification pass at the same commit both prior reviews covered. `main` has advanced by 11 commits since, so the job here was to re-derive what that changes. **All six IMPORTANT findings are still open — I re-verified each by grep against the head rather than restating them** — the `package.json` conflict is still exactly one file and one hunk, and `main` has introduced **one new, hard problem**: the migration timestamp no longer satisfies the ordering invariant.

The four sound axes were re-confirmed and I won't re-argue them: the #2397 three-state contract still discriminates `undefined` from `[]`; ADR-053's no-injection invariant is untouched (no file under `libs/core/src/fulfillment`, no allow-set entries added to either guard); both `order_records` columns stay out of `toOrm` with one narrow `IS DISTINCT FROM` guarded writer and read-path coercion.

### Findings

**[IMPORTANT — NEW, introduced by `main` advancing]** — `apps/api/src/migrations/1878000000000-add-order-record-destination-routing-block.ts`
> `main` now carries `1880000000000-create-price-change-episodes.ts` (merged yesterday as #3158). Rule 3 of the timestamp invariant — `docs/migrations.md § Timestamp uniqueness invariant`, enforced by `scripts/check-migration-timestamps.mjs` under `check:invariants` — requires a migration not yet on `origin/main` to have a timestamp **strictly greater** than every migration already there. `1878000000000 < 1880000000000`, so once this branch is updated onto `main` the guard fails `pnpm lint` with *"sorts before (or ties with) the newest migration already on origin/main"*. CI is currently green only because the last run predates #3158. This is not cosmetic: the whole point of #1013 is that a prefix sorting into the middle of merged history runs before tables it depends on exist, and fresh-DB installs break while incremental dev databases keep working. **Renumber to the next free step above the tail (`1881000000000`) and update the class suffix to match** — `1878`/`1879` are claimed by no other open PR, so there is no contention to negotiate, and the guard checks filename-prefix-vs-class-suffix consistency as well.

**[IMPORTANT — still open]** — `package.json`
> The branch still conflicts with `main` (`mergeable: CONFLICTING`, `mergeStateStatus: DIRTY`). I re-ran `git merge-tree` against the new `main`: **still exactly one conflict, still `package.json`**, still the `check:invariants` chain. The picture has not widened — both sides still append after `check-css-structure.mjs`, `main` with `check-adr-numbers` (#3036) and this PR with `check-destination-routing-reason-mirror`. Resolve by intent so **both** survive, then re-run `pnpm check:invariants` — a resolution that drops a link is silent.

**[IMPORTANT ×6 — all still open]**, re-verified at this head:

1. `destination-routing-block.types.spec.ts` — `grep -c deriveDestinationRoutingBlock` still returns **0**. Five distinguishable outcomes, none exercised, on a pure function the file's own docblock calls exhaustively assertable. Still the highest-value fix.
2. `order-sync.service.spec.ts:137` — still `markDestinationRoutingBlock: jest.fn().mockResolvedValue(undefined)` and nothing else. None of the three load-bearing properties asserted (write before the throw; arm (b) persists and still returns `[]`; a persistence failure is swallowed).
3. The per-order **badge on `/orders` still does not exist**. `destinationRoutingBlockReason` occurs in exactly four `apps/web` files — `orders.types.ts`, `order-health-summary.tsx`, `order-health.ts`, `order-detail-page.tsx` — and `orders-list-page.tsx` is **not** among them. The `Routing blocked` chip still filters to rows that show no reason.
4. `DESTINATION_ROUTING_ATTENTION_REASONS` still has no consumer — it occurs only at its declaration in `destination-routing-copy.ts` and in a comment in `orders.types.ts`, and is still absent from `features/orders/index.ts`. `tone` remains write-only.
5. **Zero `apps/web` tests.** `git diff --name-only origin/main...HEAD -- apps/web` still matches no `*.test.tsx`; seven `apps/web` files change.
6. The partial index `WHERE "destinationRoutingBlockReason" IS NOT NULL` still cannot serve `COALESCE(…, '') IN (…)`, and the migration docblock's "the index has a consumer" claim is still there. Worth folding into the same pass as the renumber above, since you are editing that file anyway.

**[INFORMATIONAL — unchanged]** — CI `Test` job
> Still the same stale run. The red check remains a **35-minute job timeout, not an assertion failure**: `libs/core` reported 6376/6376 green and the log ends mid-`woocommerce` with no `✕` and no `SIGKILL`, while `apps/web`, `apps/api` and `apps/worker` never reported at all — which still matters given finding 5. `Integration Tests` is separately green. Any re-run will need the migration renumber first, so treat the two as one step.

### Verdict

🔄 **Approve with changes** — verdict unchanged; the design is still sound and nothing needs re-architecting.

1. **Renumber the migration to `1881000000000`** (filename + class suffix) — without it `pnpm lint` now fails against the current `main`.
2. Resolve the `package.json` conflict so both invariant scripts survive; re-run `pnpm check:invariants`.
3. Re-run `Test`; confirm `apps/web` actually passes.
4. Test `deriveDestinationRoutingBlock` (1) and the service write (2) — cheapest, highest value.
5. Either build the list-row badge or correct all four claims to say detail-only (3); fold 4 and 6 in with it.

---

piotrswierzy (commented):
## Tech Lead re-verification (head unchanged at `1380cde1`)

Posted with `--comment` rather than `--approve`: this PR is authored by `piotrswierzy`, which is also the authenticated `gh` user, and GitHub forbids self-approval. Read the verdict below as the review verdict regardless of the comment-only mechanism.

Head has not moved and `main` has not advanced, so this is a delta-only re-check — conflicts, the migration-timestamp landscape, CI, and a spot-check that the standing findings still hold in code rather than a full re-read. One of my previous findings does **not** survive that spot-check and I withdraw it below.

### Findings

**BLOCKING — the migration timestamp is BELOW `main`'s tail and will fail `pnpm lint` the moment the branch is updated.** `apps/api/src/migrations/1878000000000-add-order-record-destination-routing-block.ts` sorts under `main`'s current tail of `1880000000000`, which is rule 3 of `docs/migrations.md` § Timestamp uniqueness invariant — a migration not yet on `origin/main` must be **strictly greater** than every migration already there. `scripts/check-migration-timestamps.mjs` diffs the working tree against `origin/main`, so today's green `Lint` is not evidence: that run predates #3158's merge. This is the fresh-DB-only failure mode #1013 records — incremental dev databases keep working while a from-scratch `migration:run` (and every Testcontainers boot) sorts this `ALTER TABLE` into the middle of merged history.

**The free value is `1886000000000`.** Claims across the 48 open PRs: `1881`→#3027 **and** #3201, `1882`→#3219, `1883`→#3227, `1884`→#3159 **and** #3241, `1885`→#3035. `1886000000000` is the first unclaimed slot above the highest claim. Re-prefix the filename **and** the class suffix together (rule 2) — a half-rename fails the same guard for a different reason. Note this keeps recurring precisely because every guard diffs against `origin/main` only and is blind to sibling branches.

**BLOCKING — `package.json` conflict must be resolved by intent, not by side.** Exactly one conflict, in the `check:invariants` chain: `main` appends `check-adr-numbers.mjs --self-check` + `check-adr-numbers.mjs` after `check-css-structure.mjs`; this branch appends `check-destination-routing-reason-mirror.mjs --self-check` + `check-destination-routing-reason-mirror.mjs` at the same position. Taking either side wholesale **silently deletes the other gate** — and a deleted invariant script is the failure class that reports green over a live divergence. All four entries must survive the resolution.

**IMPORTANT — six items re-verified open in code at this head:**

1. `deriveDestinationRoutingBlock` — **0 occurrences** in `libs/core/src/orders/application/services/__tests__/order-sync.service.spec.ts`. The function that decides the whole feature's reason is untested by the spec of the service that calls it.
2. `markDestinationRoutingBlock` is still a bare `jest.fn().mockResolvedValue(undefined)` (`order-ingestion.service.spec.ts:152`, was `:137` — line drift only). Nothing asserts what is passed to it, so the ingestion side of the write is unobserved.
3. The per-row **reason** never reaches the list. `apps/web/src/pages/orders/orders-list-page.tsx` references `destinationRoutingBlocked` five times (`:286`, `:465`, `:1610`, `:1616`, `:1618`) — the aggregate count and the filter chip — and `destinationRoutingBlockReason` **zero** times. This is the #2100 shape one dimension over: a filtered page whose rows do not explain why they are there.
4. **Consequently `destination-routing-block.types.ts:110-112` states something false.** The docblock for `DestinationRoutingAttentionReasonValues` says `routed-to-no-destination` "still renders a per-order badge, neutrally; it is simply never aggregated." No surface renders a per-order badge for any reason value. Either ship the badge or correct the claim — a docblock asserting a UI that does not exist is worse than silence, because the next reader treats the gap as covered.
5. **Index/predicate mismatch stands.** The migration creates `IDX_order_records_destinationRoutingBlockReason ... WHERE "destinationRoutingBlockReason" IS NOT NULL`, while the query predicate is `COALESCE(rec."destinationRoutingBlockReason", '') IN (<attention list>)` (`order-record.repository.ts:1445`), which excludes `routed-to-no-destination`. The index is usable but does not narrow to the counted set. That is exactly the `IDX_order_records_salesDocumentBlockReason` staleness the architecture overview records — a partial index over a hardcoded reason list that silently stopped matching when the union widened.
6. **Zero `apps/web` tests** across all 7 changed frontend files (`orders.api.ts`, `orders.types.ts`, `order-health-summary.tsx`, `destination-routing-copy.ts`, `order-health.ts`, `order-detail-page.tsx`, `orders-list-page.tsx`). `destination-routing-copy.ts` is a pure copy-resolution function and is the cheapest of these to cover.

**WITHDRAWN.** My previous round listed `DESTINATION_ROUTING_ATTENTION_REASONS` as consumer-less and unexported. That was wrong on both halves, and I had the identifier wrong: the constant is `DestinationRoutingAttentionReasonValues`, it **is** re-exported from `libs/core/src/orders/index.ts:104`, and it **is** consumed — `order-record.repository.ts:1445` builds the attention IN-list from it. Disregard that item.

**CI is not an assertion failure.** The red `Test` check is a **35-minute job timeout**: `libs/core` reported 6376/6376 green and the three app packages never reported at all. `Integration Tests` passed at 35m21s, `Lint`, `Build`, `Type Check`, `Docker Build Smoke Test`, `Scaffolded Adapter Builds` and `PHP Unit Tests` all pass. Treat `Test` as unreported, not failed — but note it must be re-run green after the timestamp move and the conflict resolution regardless.

**Positive, re-confirmed.** The #2397 three-state contract (absent / present-non-empty / present-empty) is respected end to end; ADR-053's no-injection invariant holds with **zero** entries added to either `scripts/check-no-injection-contracts.mjs` or `barrel-purity.spec.ts`'s allow-sets, which is the right evidence that the seam is placed correctly; and the column is omitted from `toOrm` with a single guarded `IS DISTINCT FROM` writer, matching the `salesDocument*` / `cancelledAt` single-writer precedent.

### Verdict
❌ **Request changes** — two blocking items (the sub-tail migration timestamp, and a `package.json` conflict whose naive resolution deletes a gate), plus six IMPORTANT items unchanged across five reviews. The branch still carries a **single commit**, so nothing from any prior round has been addressed.

**Priority order:** (1) re-prefix the migration to `1886000000000`, file + class together; (2) resolve `package.json` keeping all four `check:invariants` entries; (3) test `deriveDestinationRoutingBlock` and assert the `markDestinationRoutingBlock` call; (4) either render the per-row reason or correct the `:110` docblock; (5) align the partial index with the counted predicate; (6) one test for `destination-routing-copy.ts`.

---

piotrswierzy (commented):
## Tech Lead re-verification (head unchanged at `1380cde1`)

Posted as a **comment, not an approval** — GitHub forbids self-approval and this PR is authored by the `gh` authenticated user. Treat the verdict below as the review.

What moved around it: **the migration-timestamp contention has resolved itself, and not in this PR's favour.** #3201 has moved off the contested value to `1880500000000`, so the earlier reviewer conflict over `1886000000000` is moot. The current claim map is now unambiguous:

| Claim | PR | Base | Status |
|---|---|---|---|
| `1880000000000` | — | — | `main` tail (`create-price-change-episodes`) |
| `1878000000000` | **#3014 (this PR)** | `main` | ❌ **below the tail** |
| `1880500000000` | #3201 | `3179-connection-health-doc-count` | ✓ free |
| `1881000000000` | #3027 | `main` | ✓ free |
| `1881000000001` | #3210 | `3176-normalize-market-country-codes` | ✓ free, sorts above #3201 |

**One coherent assignment: every other claim stands, and #3014 moves to `1882000000000`** — above the highest live claim (#3210's `1881000000001`), so it is free regardless of the order the stack lands in. Update the class suffix to match.

### Findings

**BLOCKING — migration ordering.** `1878000000000-add-order-record-destination-routing-block.ts` sorts **below** `main`'s tail, violating rule 3 of `docs/migrations.md` § Timestamp uniqueness invariant. Today's green `Lint` predates #3158's merge, so `check-migration-timestamps.mjs` will fail on rebase. Re-prefix to `1882000000000`.

**BLOCKING — `package.json` conflict, and it cannot be resolved by taking a side.** Exactly one conflict, in `check:invariants`. `main` appends `check-adr-numbers.mjs --self-check && … check-adr-numbers.mjs`; this PR appends `check-destination-routing-reason-mirror.mjs --self-check && … check-destination-routing-reason-mirror.mjs` — both after `check-css-structure.mjs`. A naive side-take silently deletes a gate. **All four entries must survive**, in the `--self-check && bare` pair order.

**IMPORTANT — the docblock states something the code does not do.** `destination-routing-block.types.ts:110` says `routed-to-no-destination` *"still renders a per-order badge, neutrally; it is simply never aggregated."* At this head `orders-list-page.tsx` references `destinationRoutingBlocked` **5 times** and `destinationRoutingBlockReason` **0 times** — the per-row reason never reaches the list, so no per-order badge renders for any reason value, neutral or otherwise. Either wire the reason through to the row or correct the docblock; as written it is the same class of claim this PR's own § (2) is defending against.

**IMPORTANT — index/predicate mismatch.** The migration creates `IDX_order_records_destinationRoutingBlockReason … WHERE "destinationRoutingBlockReason" IS NOT NULL`, while the aggregate deliberately counts against an explicit IN-list built from `DestinationRoutingAttentionReasonValues` (which excludes `routed-to-no-destination`). This is the exact shape that went stale on `IDX_order_records_salesDocumentBlockReason` when #2248 widened that union. Match the partial predicate to the counted set, or drop the predicate and say why.

**IMPORTANT — `deriveDestinationRoutingBlock` has 0 occurrences in `destination-routing-block.types.spec.ts`.** The pure function carrying the level-triggered clear rule is untested by its own sibling spec.

**IMPORTANT — `markDestinationRoutingBlock` is a bare `jest.fn()`** at five call sites (`order-sync.service.spec.ts`, `refunds.controller.spec.ts`, `shipment.controller.spec.ts` and both shipping service specs). Nothing asserts the reason that reaches it, so property (5) — level-triggered, cleared from the held path — is unverified where it matters.

**IMPORTANT — zero `apps/web` tests across 7 changed frontend files.** New copy module, health derivation, summary component, list page and detail page all ship untested.

**Withdrawn (restated so it is not re-raised):** `DestinationRoutingAttentionReasonValues` **is** barrel-exported (`orders/index.ts:104`) and **is** consumed (`order-record.repository.ts:1445`). Not a finding.

**CI note:** the red `Test` is a **35-minute job timeout**, not an assertion failure. `Integration Tests`, `Lint`, `Build`, `Type Check`, `Docker Build Smoke Test` and `Scaffolded Adapter Builds` all pass.

### Verdict
❌ **Request changes**

---

piotrswierzy (commented):

**Verdict: 🔄 Approve with changes.** The design is right and unusually well-argued; what blocks it is entirely mechanical — the branch has been outlived by `main` and three of the four conflicts are the kind that lose a guard silently if resolved textually.

### The substance is sound

I checked the load-bearing claims against the code rather than the description:

- **`deriveDestinationRoutingBlock` is pure and its four arms partition the space correctly.** `requestedDestinationIds === undefined` (no router — every install today) vs `[]` (a router that chose nobody) is kept distinct at the top of the function, which is the #2397 three-state rule applied at the one place it can be got wrong. `resolvedCount === 0 && unresolved.length === 0` ⇒ `routed-to-source-only` is the right discrimination, since source-echo ids are removed *before* the filter and so are never "unreachable".
- **`DestinationRoutingAttentionReasonValues` is derived by `.filter`, not hand-listed**, and `routed-to-no-destination` is excluded for the stated #2100 reason. That is the exact trap `trigger-model-manual` fell into ("Invoicing blocked 4,312" on a healthy install) and it is avoided the same way. The FE mirror then derives its counted subset from the copy table's own `tone`, so the two answers to "is this attention-worthy?" cannot disagree — better than mirroring the list a second time.
- **Not `syncStatus`, and the reasoning holds.** Arm (b) genuinely has no `destinationConnectionId` to key a row on, and a new `syncStatus` arm is precisely what a consumer could route into a retry — which is the property #2397 bought. The rejection of `omsAttention` / `fulfillmentBlockReason` is also right, and the `OrderIngestionService` clear on the held branch is what makes the mutual exclusion true rather than merely asserted.
- **The partial index on `IS NOT NULL` rather than a value list** is the correct lesson from `IDX_order_records_salesDocumentBlockReason` going stale under #2248, and declaring it on the ORM entity too is required because the integration harness builds schema by `synchronize`.
- **`buildDestinationRoutingDetail` bounds a string OL does not author** before persisting and rendering it verbatim, and always states the total so truncation cannot hide scale.

Finding the vacuous-parse bug in the sibling mirror parser while writing this one, fixing it here, pinning it with a self-check case, leaving the sibling alone, and filing it as #3002 (confirmed open) is the right disposition on all four counts.

### Findings

**[BLOCKING]** — `apps/api/src/migrations/1878000000000-add-order-record-destination-routing-block.ts`
> The timestamp now violates rule 3 of `docs/migrations.md` § Timestamp uniqueness invariant: a new migration must sort **strictly above** every migration on `origin/main`. `main`'s newest three are `1890000000000` (#3035), `1891000000000` and `1892000000000` (#3241), all merged 2026-09-14 — after this branch's last merge of `main` on 2026-09-08. So `1878000000000` now sorts into the middle of merged history and `check-migration-timestamps.mjs` will refuse it on rebase. Renumber to `1893000000000` and update the class suffix. The migration's own docblock already narrates this happening once ("RENUMBERED from `1875000003400` … overtaken by ordinary merges"); it has simply happened again, which is worth noting in that comment rather than silently re-bumping — a late-merging migration is renumbered on every pass, not once.

**[BLOCKING]** — `package.json`, the `check:invariants` conflict
> This is the conflict to resolve by intent, never textually. `main` has gained five invariant scripts the branch never saw — `check-sales-document-condition-field-mirror`, `check-auto-applied-limit-mirror`, `check-price-override-bound-mirror`, `check-sourcing-rule-vocabulary-mirror`, `check-adr-numbers` — while the branch adds `check-destination-routing-reason-mirror`. Taking either side wholesale **deletes guards with nothing failing**: "ours" drops this PR's own mirror (and the FE/core union silently drifts thereafter), "theirs" drops five of main's. The resolution is the union, and it is worth grepping the resolved string for all six names afterwards rather than trusting the merge.

**[IMPORTANT]** — `libs/core/src/orders/domain/entities/order-record.entity.ts`, `…/repositories/order-record.repository.ts`, `apps/api/src/orders/http/dto/order-record-response.dto.ts`
> Three further conflicts, all against #3186's `salesDocumentMatchedRuleId`, which landed in the same guarded-UPDATE and the same `toOrm` exclusion set this PR extends. The repository hunk in particular has both sides editing the *single narrow atomic writer* and its docblock — a resolution that keeps one column's argument and drops the other's leaves a column written by a statement whose comment says it is not. Resolve so that all four columns (`salesDocumentBlockReason`, `salesDocumentUnresolvedReason`, `salesDocumentBlockDetail`, `salesDocumentMatchedRuleId`) **and** the two new `destinationRoutingBlock*` columns are in one guarded statement and all six stay excluded from `toOrm`, then re-read the merged docblock as a whole.

**[IMPORTANT]** — CI signal is against a stale base
> Every green check on this PR ran against the 2026-09-08 merge commit, a week behind `main`. `Test` is `CANCELLED` (not green), and the body's own ⚠️ caveat about integration tests not having been executed predates the now-`SUCCESS` Integration Tests run — which also ran against that stale base and therefore against the pre-#3035 reservation code and without #3241's migrations. Merge `main`, resolve the four conflicts, and re-run **`pnpm test` and `pnpm test:integration` locally** before merge; the new migration plus the guarded UPDATE are exactly what a stale integration run cannot vouch for.

**[SUGGESTION]** — `destination-routing-block.types.ts`, `routed-destinations-unavailable`
> The throw for the none-resolved case is kept, correctly, and the reason is persisted *first* so the operator sees "a named destination was unreachable" rather than "nothing is configured". Worth confirming in a test that the persist survives the throw — i.e. that the write is committed rather than rolled back with the failing job — since the whole operator-facing value of this arm depends on the reason outliving the exception that follows it.

**[Positive]** — the `detail` field is deliberately connection-ids-and-counts only, and says so at the type. Given it is persisted and rendered verbatim from a payload OL does not author, stating the PII boundary on the field rather than in a call-site comment is the right place for it.

### Merge readiness

🔄 **Approve with changes.** Nothing here needs redesign — the four findings are a renumber, a three-way conflict resolution that must be done by intent, and a re-run of the gates against a current base. Once `main` is merged and `pnpm lint` (which will exercise both the migration-ordering guard and the new mirror) plus `pnpm test:integration` are green locally, this is ready.

Priority order:
1. Merge `main`; resolve `package.json` to the **union** of both invariant lists and verify all six script names survive.
2. Resolve the three `order-record` conflicts against #3186 by intent, not textually.
3. Renumber the migration to `1893000000000` (class suffix too).
4. Re-run `pnpm lint`, `pnpm test`, `pnpm test:integration` on the merged tree.

piotrswierzy:
## Tech Lead review — #3014

> Posting as a comment rather than a formal review: GitHub refuses `--approve` on one's own PR. **Verdict: 🔄 Approve with changes** — no BLOCKING findings, three IMPORTANT.

Read against `docs/architecture-overview.md` § 4 *Orders* (the #2397 router-filtered fan-out bullet and the #2100 `SalesDocumentBlockOutcome` bullet), `docs/engineering-standards.md` § *Testing Standards*, and `docs/migrations.md`.

### What is right, and worth saying because it is the hard part

- **The three states of `destinationConnectionIds` survive intact.** `deriveDestinationRoutingBlock` (`destination-routing-block.types.ts:196-237`) branches on `undefined` first and returns `null` — the unfiltered fan-out — before `[]` becomes `routed-to-no-destination`. The field's own docblock (lines 172-179) names the collapse as *"the #2397 mistake that would stop destination provisioning everywhere"*.
- **Arm (b) still returns `[]` and still does not throw.** The persist call sits at `order-sync.service.ts:98-121`, *above* every branch, so writing a reason can never convert a deliberate empty decision into a throw — exactly the property that keeps `isNonRetryableError`'s per-plugin registry from burning ten attempts on a router's choice. The `skipped_cancelled` / `skipped_held` precedent is cited where it is used.
- **(a), (b) and (c) stay distinguishable at the branch**, and the new fourth value `routed-to-source-only` is a genuine refinement: source-echo ids are excluded *before* the filter, so naming them as unreachable would blame a working connection.
- **#2100's discipline is followed in full**: both columns excluded from `toOrm` with the reasoning recorded at both sites (`order-record.repository.ts:2583-2590` and `3072-3079`), one narrow guarded UPDATE as sole writer, level-triggered including `null`, coerced on read via `isDestinationRoutingBlockReason` so a rolled-back newer value reads as absent, counted against an explicit IN-list rather than `IS NOT NULL`, ANDed beside `health` rather than widening the `OrderHealthValues` partition, and the attention subset **derived by `.filter`** so it cannot contradict the vocabulary.
- **The `COALESCE(…, '')` comment at `order-record.repository.ts:1436-1440` is right and non-obvious** — `NOT (NULL IN (…))` is NULL and `WHERE` drops the row, so without it `destinationRoutingBlocked=false` returns zero orders. Worth the paragraph it got.
- **The mirror-parser bug found in the sibling script** (closing `]` located before comments are stripped, so a `]` inside a docblock reads the array as empty and both sides pass vacuously) is a real find, correctly fixed here and filed as #3002 rather than silently patched across.

---

### [IMPORTANT] — `deriveDestinationRoutingBlock` has no unit test

`destination-routing-block.types.spec.ts` covers the counted subset, `isDestinationRoutingBlockReason` and `buildDestinationRoutingDetail` — but **not the derive function**, which is the six-branch rule the whole slice turns on. `order-sync.service.spec.ts` gains only a `markDestinationRoutingBlock` mock (+7/−0).

The function's own docblock (`destination-routing-block.types.ts:188-190`) argues:

> *PURE — no I/O, no clock, no injected dependency — so the rule can be asserted exhaustively without a service*

and then it is not asserted at all. Six behaviours are untested, each mapping to a distinct operator-facing reason: `undefined → null`; `[] → routed-to-no-destination`; `resolvedCount === 0` with an empty unresolved set `→ routed-to-source-only`; `resolvedCount === 0` with a non-empty one `→ routed-destinations-unavailable` *with* detail; the partial case `→ routed-destinations-partially-unavailable`; and all-resolved `→ null`. `docs/engineering-standards.md` § *Testing Standards* requires a unit test for non-trivial logic, and the sibling pure rules this file names as precedent (`applyPricingRule`, `resolveOfferLifecycle`) both carry exhaustive specs. This is the cheapest test in the PR and the one that matters most.

### [IMPORTANT] — `order-ingestion.service.ts:1150-1172`: the new method was inserted between a docblock and the method it documents

`persistFulfillmentOutcome`'s docblock (*"Persist why the intercept held this order (#2396), or clear a stale reason"*, with its numbered rules 1 and 2) now sits immediately above **`persistDestinationRoutingCleared`**, which then has its own docblock underneath it. `persistFulfillmentOutcome` itself is left undocumented at line 1185.

Verified at head `1380cde`. A reader arriving at `persistDestinationRoutingCleared` gets two docblocks, the first of which describes a different method and mentions `persistSalesDocumentOutcome`. Move the new method (and its docblock) below `persistFulfillmentOutcome`.

### [IMPORTANT] — the new partial index cannot serve the predicate it was added for

The migration (`1878000000000-…:18-29`) argues the index earns its place because *"this slice does filter and count, so the index has a consumer"*, and chooses `WHERE "destinationRoutingBlockReason" IS NOT NULL` over a value list — sound reasoning about staleness, and the `IDX_order_records_salesDocumentBlockReason` precedent is the right one to cite.

But the consumer is `IS_DESTINATION_ROUTING_BLOCKED` (`order-record.repository.ts:1442-1444`):

```sql
COALESCE(rec."destinationRoutingBlockReason", '') IN ('routed-destinations-unavailable', …)
```

Wrapping the indexed column in `COALESCE` leaves **no indexable qual on the bare column**, so the planner has nothing to match against `IDX_order_records_destinationRoutingBlockReason` — the filter arm degrades to a scan whether or not the partial predicate is implied. The index is dead for its stated purpose.

The `COALESCE` is only needed for the **negation** (which the comment correctly explains). The positive arm does not need it: `reason IN (…)` is already correct for `blocked = true`, because `NULL IN (…)` is NULL and `WHERE` drops the row — which is the wanted answer. Splitting the two predicates keeps the index usable for the true arm and the negation total:

```ts
private static readonly IS_DESTINATION_ROUTING_BLOCKED    = `rec."destinationRoutingBlockReason" IN (…)`;
private static readonly NOT_DESTINATION_ROUTING_BLOCKED   = `COALESCE(rec."destinationRoutingBlockReason", '') NOT IN (…)`;
```

No live impact today — both columns are NULL on every install — but the index and its justification should agree before a router is wired.

### [SUGGESTION] — the two swallow paths disagree on log level for the same column

`OrderSyncService.persistDestinationRoutingBlock` uses `logger.error` and spends a paragraph justifying it (`order-sync.service.ts:536-545`): losing the write *"means the operator loses the reason entirely … a failure of OL's own bookkeeping"*, and it keeps `warn` meaning exactly one thing so the #2397 warn-count assertions stay honest.

`OrderIngestionService.persistDestinationRoutingCleared` (`order-ingestion.service.ts:1176-1180`) uses `logger.warn` for a failed write of the same column. The argument transfers: a failed *clear* leaves a stale routing reason standing beside a fresh `fulfillmentBlockReason` — the two-reasons-for-one-order state the clear exists to prevent, and the one the ORM docblock calls out. Either raise it to `error` or record why the clear is the weaker case.

### [SUGGESTION] — the integration suite has not run, and this PR is the case for running it

The description flags it. Restating it as a gate rather than a note: this adds a migration, a raw-SQL guarded UPDATE with `IS DISTINCT FROM` semantics, an ORM-declared partial index that must match the migration (the harness builds schema by `synchronize`), and two new count/filter predicates. `pnpm test:integration` before merge, per `CLAUDE.md`'s quality gate for schema changes, plus `pnpm --filter @openlinker/api migration:show`.

### [SUGGESTION] — description cites the pre-rename migration timestamp

The body says *"Adds migration `1875000003400`"*; the file is `1878000000000-add-order-record-destination-routing-block.ts`. The migration's own header documents the renumber and why (`…:31-43`) — that reasoning is worth keeping and the body should just carry the final number. Same for the diffstat (`+1877/−40`, not `+1844/−40`).

---

## Summary

This closes both #2703 and #2704 in the right shape: the #2100 persisted-reason pattern rather than a `syncStatus` arm, with the reasoning for that choice written where the next reader will find it. Every load-bearing property of #2397 survives — three states never collapsed, arm (b) returning `[]` without throwing, the three empty-destination conditions distinguishable at the branch — and the partial-eligibility case that was previously a bare `logger.warn` becomes a persisted, operator-visible reason. The write discipline (`toOrm` exclusion, single guarded writer, level-triggered, coerced on read, counted by IN-list, ANDed beside `health`) is exactly right. The gaps are a missing spec for the one pure rule the design rests on, a docblock now attached to the wrong method, and an index whose own consumer cannot use it.

## Merge Readiness

🔄 **Approve with changes** — no BLOCKING issues; the change is behaviour-preserving on every install today, since nothing populates `OrderSyncRequest.destinationConnectionIds`.

## Priority fixes

1. Run `pnpm test:integration` and `migration:show` — new migration, new index, new raw UPDATE.
2. Add the exhaustive spec for `deriveDestinationRoutingBlock` (six branches).
3. Move `persistDestinationRoutingCleared` below `persistFulfillmentOutcome` so the #2396 docblock rejoins its method.
4. Split the blocked predicate so the positive arm keeps a bare-column qual and the new index has a real consumer.
5. Align the clear-path log level, and correct the migration number and diffstat in the description.

piotrswierzy (commented):
## Tech Lead review — `fix(orders): persist why a routing decision narrowed the destination fan-out`

*(Commenting rather than approving — this is my own PR, so GitHub won't take an approval from me. Treat the verdict as "approve once the two blockers below are cleared".)*

Design is right and closes the follow-up `docs/architecture-overview.md` §4 Orders explicitly left open under #2397. Two mechanical blockers.

---

### 🔴 BLOCKING 1 — the migration timestamp is below `main` again

`apps/api/src/migrations/1878000000000-add-order-record-destination-routing-block.ts`

`main`'s four newest migrations are now:

```
1889000000000-add-price-change-episode-claimed-at.ts
1890000000000-create-order-cancellation-signals.ts
1891000000000-add-invoice-buyer-tax-id-value.ts
1892000000000-inline-sales-document-rule-amounts.ts
```

`1878000000000 < 1892000000000`, so `scripts/check-migration-timestamps.mjs` rule 3 (*strictly greater than every migration already on `main`*) will refuse it and fail `pnpm lint`. Needs renumbering to at least `1893000000000`, with the class suffix moved to match.

This is the same thing the file's own docblock already records happening once:

> *`1876000000000` … and `1877000000000` … both landed on `main` first, so `…3400` came to sort BEFORE the newest migration there and the guard refused it.*

…and the lesson it drew is exactly right and now needs applying a second time:

> *an offset protects against a sibling branch, not against the branch being outlived by the queue ahead of it; a late-merging migration is renumbered rather than nudged.*

Worth adding a line to that docblock noting the second renumber, so the next reader sees this is a function of how long the PR sat open rather than a one-off.

### 🔴 BLOCKING 2 — CI is 13 days stale and `Test` was cancelled

The only run is from **2026-09-08**, against a `main` that predates all four migrations above. `Test` came back **cancelled**, not green; `Integration Tests` passed but took 35 minutes, which suggests that run was struggling generally.

Rebase and get a clean run before merging — particularly because this touches `order-record.repository.ts`, `order-record.orm-entity.ts` and the predicate-parity spec, where a stale base is exactly where a silent conflict hides.

---

### ✅ The reason vocabulary is placed correctly, and the alternatives are argued rather than dismissed

The docblock explains why this is the #2100 persisted-reason shape and not a `syncStatus` arm, and all three arguments hold:

1. A `syncStatus` row is keyed by `destinationConnectionId`, and the whole content of `routed-to-no-destination` is that there isn't one.
2. A new `syncStatus` arm is precisely what a consumer could route into a retry — the property #2397 bought and #2703 AC 3 preserves.
3. `OrderSyncStatusFilterValues` is documented as the single source of truth for both the persisted payload and `GET /orders?syncStatus=`, so widening it widens a public query vocabulary for a value that can never appear on a real destination row.

Also right to distinguish it from `omsAttention` (a closed union about *who decides what*; destination-mirror creation is a commercial act) and `fulfillmentBlockReason` (the intercept HELD the order, so `syncOrder` never ran). Stating that those two are mutually exclusive with this one *by construction* — and clearing this reason on the held branch in `OrderIngestionService` — is what keeps `Needs attention (N)` from double-counting.

The `DestinationRouting*` prefix over `Routing*` is a good catch: `RoutingOutcome` / `InboundRoutingPolicy*` already exist in `@openlinker/core/sync` for inbound **webhook** routing, and at a call site the two would read as one concern.

### ✅ `routed-to-no-destination` excluded from the attention set

The single most important decision here. A working router deciding some orders are fulfilled elsewhere produces a **large, permanent, entirely healthy** population, and counting it would put a red badge on a correctly-configured install — the #2100 `trigger-model-manual` trap exactly. Deriving `DestinationRoutingAttentionReasonValues` by `.filter` from the full list rather than hand-writing it is the right mechanism: a value added later can't silently miss the list or contradict it.

`routed-destinations-partially-unavailable` is a genuine find — before this, a partially-resolved decision proceeded **silently under-provisioned** behind a `logger.warn`, since the throw branch only fires when nothing resolved.

Keeping `routed-to-source-only` distinct from `routed-destinations-unavailable` is also correct: source-echo ids are dropped *by design* before the filter, so reporting one as unreachable would tell an operator a working connection is broken.

### ✅ The migration's index form

Partial on `IS NOT NULL` rather than over a value list, with the reason:

> *that is exactly what silently went stale on `IDX_order_records_salesDocumentBlockReason` when #2248 widened that union without touching the index.*

And declared on the ORM entity too, because the harness builds schema by `synchronize` and a migration-only index would hold in production and silently not in tests. Both `up()` and `down()`, both idempotent.

### ✅ One write point, level-triggered

`persistDestinationRoutingBlock` sits above every branch so the cancelled / held / dispatched / thrown paths each get exactly one write and can't drift; it runs **before** the throws, which is the whole point of #2704 (an operator sees *a named destination was unreachable*, not *nothing is configured*); and it writes `null` on a run that narrows nothing, which is the only thing that clears a stale reason once the config is fixed. Best-effort, so it can never mask a real error.

Re-affirming that a routing decision is a **requirement, not a preference** — no silent substitution of a sibling destination, because a fallback order is a router rule and this service resolves no router by design — is the right resolution of #2704 and worth having written down where the throw is.

### 🟢 SUGGESTION

`scripts/check-destination-routing-reason-mirror.mjs` is the 57th entry on the `check:invariants` one-liner. Same note I left on #3268: worth extracting that chain into a script or manifest in its own PR, since `&&` means the first failure hides every later one.

---

**Verdict:** ❌ Not mergeable as-is — renumber the migration to `≥ 1893000000000` and get a clean CI run on a current base. The design and implementation I'd approve unchanged.

piotrswierzy (commented):
## Re-check at `1380cde` — the PR has not moved, but `main` has, and my own remediation number has expired

Head is unchanged since my three earlier reviews. I am not restating the six IMPORTANT findings; they are open and re-verified. What has changed is the branch's position, and one correction I owe.

### My previous fix instruction is now wrong

Ten days ago I wrote:

> *Renumber to the next free step above the tail (`1881000000000`) … `1878`/`1879` are claimed by no other open PR, so there is no contention to negotiate.*

`main`'s tail is now **`1892000000000`** (`1892-inline-sales-document-rule-amounts`). `1881` sorts below it, so following that instruction today reproduces the same `check-migration-timestamps.mjs` failure it was meant to fix. The migration on this branch is still `1878000000000` — **fourteen slots below the tail**, not two.

**The number to use is `1901000000000`.** I scanned every remote ref rather than only `main`: `1893` is contested (two unmerged stacks), and `1895`–`1900` are each claimed by the pack-bench or subiekt stacks. `1901`–`1909` are free everywhere. Filename prefix and class suffix move together — the guard checks both.

Recording the expiry rather than quietly issuing a new number, because it is the same failure mode this PR's own findings are about: a stated remedy that stopped being true and would have been followed on trust.

### Position

```
behind origin/main   94 commits
migration            1878000000000   (main's tail: 1892000000000)
merge state          still conflicting — package.json, the check:invariants chain
```

The `package.json` conflict is the same single hunk I reported twice: both sides append to the chain, so resolve by **intent** (every script survives) rather than by taking a side, and re-run `pnpm check:invariants` — a resolution that drops a link fails silently.

Note the conflict has had ten more days to widen and has not: it is still one file. But 94 commits is enough that the rebase itself now carries risk beyond that hunk, and two things this PR touches have moved underneath it — `orders-list-page.tsx` (#3320's awaiting-submission chip landed there) and the `check:invariants` chain (several new guards). Worth reading the rebase rather than replaying it.

### Verdict

🔄 **Approve with changes**, unchanged. The design is still sound and nothing needs re-architecting. The order has changed, though — the renumber is now the thing that gates everything else, because `pnpm lint` fails until it is done:

1. Renumber `1878000000000` → **`1901000000000`** (filename + class suffix).
2. Rebase onto `main` and resolve the `package.json` chain so every script survives.
3. Re-run `Test` — it has never reported on `apps/web`, and seven `apps/web` files change here with no test among them.
4. Test `deriveDestinationRoutingBlock` and the service write — still the cheapest, highest-value pair.
5. Either build the list-row badge or correct all four claims to say detail-only.

piotrswierzy:
## Correction — my renumber instruction has expired again; take the procedure, not the number

I told you to use `1901000000000`. **PR #3451 took it within the hour**, and its ten-branch returns stack now carries it:

```
1901000000000: 3088 3089 3090 3092 3093 3094 3095 3312 3450 3466
```

Following my instruction as written would create the same cross-branch collision it was meant to fix. That is the second time this has happened on this PR (`1881000000000` went stale first, when `main`'s tail moved to `1892`), so the instruction is the wrong shape. Here is the right one.

### The state as of this comment

`apps/api/src/migrations/1878000000000-add-order-record-destination-routing-block.ts` is **14 slots below `main`'s tail**, which is `1892000000000`. Claimed above that, across every remote ref:

| Timestamp | Held by |
|---|---|
| `1893` | pack-bench stack (12 branches) |
| `1894` | `backup/subiekt-epic-pre-resign` |
| `1895` `1896` `1897` `1899` | mockup-parity stack |
| `1898` `1900` | `subiekt-gt-full-capabilities` |
| `1901` | returns/correction-proposal stack (10 branches) |

`1902000000000` and up are free **at this moment**. I am naming it, and I am also telling you not to trust it.

### The procedure, which does not expire

Re-run this **in the commit that does the rename**, not before:

```bash
git fetch --all --prune
for r in $(git for-each-ref --format='%(refname:short)' refs/remotes/origin); do
  git ls-tree -r --name-only "$r" -- apps/api/src/migrations 2>/dev/null \
    | grep -oE '/1[89][0-9]{11}-' ;
done | grep -oE '[0-9]{13}' | sort -u | tail -5
```

Take the next round number above the highest it prints. A number checked an hour before the push is a number someone else has since claimed — this repo moves a migration slot roughly daily right now.

Note why `scripts/check-migration-timestamps.mjs` cannot catch this for you: it enforces uniqueness **within one tree** and ordering **against `origin/main`**. Two unmerged sibling branches both claiming `1902` are invisible to it, and stay invisible until the second one merges.

### Rename BOTH halves

Rule 2 of that same script compares the filename prefix against the class suffix, and TypeORM 0.3.17 reads the **class**, never the filename (`MigrationExecutor.js:417` — `parseInt((migration.name || constructor.name).substr(-13), 10)`). So a filename-only rename is functionally a no-op that additionally fails `pnpm lint`. All three lines move together:

```
apps/api/src/migrations/<NEW>-add-order-record-destination-routing-block.ts
export class AddOrderRecordDestinationRoutingBlock<NEW> implements MigrationInterface {
  name = 'AddOrderRecordDestinationRoutingBlock<NEW>';
```

(#3451 has just made exactly this mistake — filename moved to `1901`, class still `1893` — which is what prompted me to re-check this one.)

Also note `mergeable_state: dirty` on this PR: it needs a rebase onto `main` regardless, and that is the natural moment to do the renumber.

piotrswierzy (commented):
## 🟡 Migration ordering — `1878000000000` now sorts below `main`'s tail

Found on a cross-PR migration scan rather than in the diff, and **I should have caught it in my earlier review of this PR** — it was already true then. Reporting it plainly rather than as something that newly arose.

`apps/api/src/migrations/1878000000000-add-order-record-destination-routing-block.ts` carries prefix **1878000000000**. `origin/main`'s newest migration is **1892000000000**, and this branch is **95 commits behind main**, so fourteen migrations have landed above it while the PR has been open.

That violates rule 3 of `docs/migrations.md` § Timestamp uniqueness invariant:

> a new migration's timestamp must be **strictly greater** than every migration already on `main` (core + plugin dirs)

`scripts/check-migration-timestamps.mjs` enforces exactly this by comparing the working tree against `origin/main` via `git ls-tree`. If this branch's lint is green, that is worth a second look on its own — the doc notes the check **skips even in CI** on a runner with no `git` binary, which is the one way a violation like this stays invisible.

### The practical risk here is low, and saying so matters

I read the migration rather than assuming the general case. It is:

```sql
ALTER TABLE "order_records" ADD COLUMN IF NOT EXISTS "destinationRoutingBlockReason" text
ALTER TABLE "order_records" ADD COLUMN IF NOT EXISTS "destinationRoutingBlockDetail" text
CREATE INDEX IF NOT EXISTS "IDX_order_records_destinationRoutingBlockReason" …
```

`order_records` is created far below 1878, both DDL statements are `IF NOT EXISTS`-guarded, and nothing between 1878 and 1892 is a dependency. So the failure mode § 6 of the migrations doc describes — *fresh-database `migration:run` fails with `relation … does not exist` while incremental dev DBs keep working* — is **not** reachable for this particular migration.

What is left is the weaker but real cost: on merge, fourteen migrations written before yours run after it on every fresh install, and the ordering guard has to be either failing or skipping for that to happen. Rule 3 is unconditional precisely so nobody has to make this judgement per migration.

### Fix

Renumber to the next free synthetic prefix and update the class suffix to match (rule 2 catches a half-rename). I scanned every open branch so the free slot does not have to be guessed — **1893 through 1901 are all claimed** by branches in flight:

| prefix | claimed by |
|---|---|
| 1893 | pack-bench stack (`…-add-fulfillment-work-assignment`) |
| 1895, 1896, 1897, 1899 | pack-bench stack (#3435 / #3439) |
| 1898, 1900 | `subiekt-gt-full-capabilities` (#3365) |
| 1901 | returns stack (`…-add-return-line-resolved-catalog-identity`) |

**`1902000000000` is the first free one.** Since this migration has never run anywhere (no production caller populates these columns yet, as the docblock itself says), a plain rename is enough — no self-healing `DELETE` of an orphaned `migrations` row is needed, unlike the #1013 recovery the doc walks through.

While you are in there: a rebase onto current `main` is worth doing anyway at 95 commits behind.

---

## Final assessment

### Summary
The routing-narrowed-destination visibility columns and their partial index, which I reviewed favourably before and have no new objection to on their own terms — the index being partial on `IS NOT NULL` rather than a value list is the right call, and the docblock's reasoning about `IDX_order_records_salesDocumentBlockReason` going stale when #2248 widened its union is exactly why. The problem is positional, not in the content.

### Merge readiness
🔄 **Approve with changes** — one renumber, then this is fine.

### Priority fixes
1. Re-prefix the migration to `1902000000000` (first free across all open branches) and update the class suffix in the same edit.
2. Rebase onto `main`, and check whether `check-migration-timestamps` is actually running on this branch's CI — if it was green with 1878 against a 1892 tail, the guard is skipping rather than passing.

piotrswierzy (commented):
## Correction to my renumber advice: take `1903000000000`, not `1902`

`main` moved `5d470b97e..fbae25c1b` since I wrote that comment — the fourteen-PR pack-bench stack merged, taking `1893`, `1895`, `1896`, `1897` and `1899` onto `main` with it.

Two consequences for the advice I gave:

**The finding is unchanged but larger.** `main`'s tail is now **1899000000000**, so `1878000000000` is below it by twenty-one migrations rather than fourteen. Same rule-3 violation, same low practical risk (the DDL is `ADD COLUMN IF NOT EXISTS` on a long-existing `order_records`, no dependency in the window), same fix.

**The number I named is now spoken for.** `subiekt-gt-full-capabilities` (#3365) had `1898000000000`, which was legal when I scanned an hour ago and became a rule-3 violation the moment the stack merged past it. It needs a renumber too, and I have pointed it at `1902000000000` — so please take **`1903000000000`** here, and update the class suffix to `AddOrderRecordDestinationRoutingBlock1903000000000` in the same edit.

Current claims above `main`'s tail, so nobody has to re-derive them:

| prefix | owner |
|---|---|
| 1900 | #3365 |
| 1901 | returns stack (`…-add-return-line-resolved-catalog-identity`) |
| 1902 | #3365 (renumbered from 1898) |
| **1903** | **this PR** (renumbered from 1878) |

Everything else in my previous comment stands, including that a plain rename suffices here — this migration has never run anywhere, so there is no orphaned `migrations` row to clean up.

piotrswierzy (commented):
## Also now conflicts with `main`, on the entity this PR extends

Adding to the renumber note: this branch is **109 commits behind** and no longer merges cleanly.

```
CONFLICT (content): libs/core/src/orders/domain/entities/order-record.entity.ts
CONFLICT (content): apps/api/src/orders/http/dto/order-record-response.dto.ts
```

Both are the same shape: the merged pack-bench stack added `omsAttention` to `OrderRecord` and its response DTO, and this PR adds `destinationRoutingBlockReason` / `destinationRoutingBlockDetail` to the same constructor and the same DTO. Two independent field additions to one positional constructor — the resolution has to keep both, and a "take theirs" on the entity drops this PR's whole read surface while still compiling everywhere the fields are optional.

Worth doing the rebase and the migration renumber in one pass, since both are forced by the same merge. To restate the numbers with `main`'s tail now at `1899`:

| prefix | owner |
|---|---|
| 1900 | #3365 |
| 1901 | returns stack |
| 1902 | #3365 (renumbered from 1898) |
| **1903** | **this PR** (renumbered from 1878) |

One correction to something I wrote earlier, now that `docs/migrations.md` has been updated on `main`: I described the duplicate-prefix hazard in the terms the old doc used. The current text corrects it against the installed package — in `typeorm@0.3.17` pending is decided by **class name** (`MigrationExecutor.js:79`), so two files sharing a prefix both run; what is undefined is their **order** (`:425` sorts on timestamp with no tie-break). That does not change anything about this PR — rule 3 is an ordering rule and `1878` violates it either way — but the reason to care is non-deterministic apply order plus a hard merge gate, not a silently skipped `up()`.

Issue #2369, opened 2026-08-22
[TASK] Integration — implement adjustInventory on PrestashopInventoryMasterAdapter


piotrswierzy:
> Part of the OMS programme backlog (**Wave 2**, stream S1, size M). Backlog slug: `W2-32`. Sibling `W2-*` slugs resolve via the Wave-2 epic's checklist.

## Problem / Context

`libs/integrations/prestashop/src/infrastructure/adapters/prestashop-inventory-master.adapter.ts:283` throws `PrestashopNotSupportedException`. PrestaShop is the most common OL master, so on a default install `restock_blocked` (`W2-33`) is the **default** outcome, not the exception — which turns the returns UX spec's highest-severity remediation path into the normal path and gets it learned-ignored.

This issue needs no returns model to justify it and should land in the same window as `W2-43`.

## Proposed Solution

Implement `adjustInventory` over the PrestaShop webservice `stock_availables` resource — resolve the `(product, combination)` position the adjustment targets, read the current quantity, apply the delta, write it back. Honour `W2-31`'s idempotency key by recording applied keys per connection so a retry is a no-op, and carry the reason where PrestaShop can hold it (falling back to a logged value where it cannot).

The read-modify-write window is the risk: document it, keep the window as small as possible, and prefer the module endpoint if one already exists for stock (the ADR-016 `importorder` precedent shows the OL module is the right home for anything the raw webservice does unsafely).

## Classification

**Type**: Integration
**Layer**: Infrastructure
**File(s)**: `libs/integrations/prestashop/src/infrastructure/adapters/prestashop-inventory-master.adapter.ts` and its HTTP client

## Dependencies

- #2368 (`W2-31`)

## Assumptions

- Multi-shop and advanced-stock-management configurations may still refuse; refusal must stay a loud `restock_blocked`, never a silent success.

## Acceptance Criteria

- [ ] A restock of `n` units raises the PrestaShop quantity by exactly `n` (integration test against the dev stack)
- [ ] A repeated idempotency key does not double-apply
- [ ] Any unsupported configuration raises a typed refusal that surfaces as `restock_blocked`
- [ ] Tests added; no boundary violations

piotrswierzy:
## Status after PR #2645 — deliberately left open, one AC short

The behaviour this issue exists for **is implemented and shipped** on `2367-returns-custody`, and ACs 2–4 are met. **AC1 is not**, so no `Closes` trailer was added and the issue stays open.

### What is done

[...]

Issue #2368, opened 2026-08-22
[TASK] CORE — amend InventoryMasterPort.adjustInventory with idempotency key and reason


piotrswierzy:
> Part of the OMS programme backlog (**Wave 2**, stream S1, size M). Backlog slug: `W2-31`. Sibling `W2-*` slugs resolve via the Wave-2 epic's checklist. Design of record: design §7.3.

## Problem / Context

§7.3's restock answer: when a shop is the `InventoryMaster`, **OL restocks THERE**, through the **existing** `adjustInventory` (`libs/core/src/inventory/domain/ports/inventory-master.port.ts:101`) — not a new capability, because the method already exists on the base port. `inventory_items` is a mirror rewritten by every master sync, so an OL-side increment silently vanishes at the next tick.

But the command type carries no idempotency key and no reason, so a retried restock double-increments the master's stock, and the master's own audit shows an unexplained adjustment.

## Proposed Solution

`InventoryAdjustment` gains **optional** `idempotencyKey` and `reason` (a closed union: `return_restock | manual_correction | …`). Optionality is what keeps every out-of-tree `InventoryMaster` plugin source-compatible — the same discipline §4.2 applies to the deprecated reservation methods, and the same reason `listExternalIdsByConnection`'s page argument was optional (#2219).

WooCommerce (`libs/integrations/woocommerce/src/infrastructure/adapters/inventory-master/woocommerce-inventory-master.adapter.ts:141`) honours the key by refusing a repeat of the same key within a bounded window; an adapter that cannot dedupe says so through its result rather than pretending.

## Classification

**Type**: CORE
**Layer**: Domain + Integration
**File(s)**: `libs/core/src/inventory/domain/ports/inventory-master.port.ts`; the inventory types; the WooCommerce adapter

## Dependencies

- #2315 (`W1b-11` — the deprecation discipline established)

## Assumptions

- No adapter needs a *required* key; making it required would break every out-of-tree implementer for a Wave-2 feature.

## Acceptance Criteria

- [ ] Both fields optional; existing callers compile unchanged (type-check is the test)
- [ ] WooCommerce dedupes a repeated key and reports the dedupe rather than silently no-op-ing
- [ ] The reason reaches the master where the master can carry it
- [ ] `architecture-overview.md` § Inventory updated; tests added

PR #3015, opened 2026-09-08
chore: mark ADR-048/049/050 Accepted and retire three stale claims

piotrswierzy:
Audit housekeeping from the #2162 async-work-layer epic. **No runtime behaviour changes** — but it is not purely docs, so the one code change is called out below rather than left for a reviewer to find under a `chore:` prefix.

## ADR statuses

ADR-048, ADR-049 and ADR-050 still read `Status: Proposed` long after they shipped. Each is flipped to `Accepted` in the ADR header **and** in the README index table, which carries its own Status column and would otherwise contradict the file it links to.

Every issue number was verified against git history first, and each status line names what is **not** built rather than claiming the whole ADR:

| ADR | Implemented by | Explicitly still open |
|---|---|---|
| 048 | #2218/#2219 (`9099a7339`), #2220/#2221 (`e30c18e7b`), #2222 (`d392f1c02`) | decision 6 — nothing records a `RateLimitTimeoutError` observation, so back-pressure never shrinks the next budget |
| 049 | #2163 (`94ba20e6a`, `3598ff65a`), #2164 (`66bbb8956`), #2280 (`25111eb93`) | decision 1 shipped on the **webhook path only**; the non-webhook writers of `jobs.sync` remain future work |
| 050 | #2278 (`e9138663f`) | decision 6 met for `bulk` alone; every other lane stays illustrative pending #1134 |

## A false claim in the architecture overview

It read *"PrestaShop still refuses `adjustInventory` outright, which is #2369's subject."* That is **false** — the adapter implements it (read the current `stock_availables` row → compute → write the full resource back, with #2368 key dedupe in the shared `CachePort` on the same seven-day TTL WooCommerce uses) and refuses three *named configurations* instead: multi-shop ambiguity with no `shopId`, no matching row, and `depends_on_stock = 1`.

It cites the landing PR rather than an issue because neither #2077 nor #2369 appears in any commit on `main`, so the PR is the only citable record.

## The one code change, and why it is inert

`SyncJobRunner.laneCaps` initialised `fan-out: 1/1` while `resolveLaneCaps()` returned the post-#2609 `8/4`.

**Not a live bug** — the field is replaced at `onModuleInit`, unconditionally, before the runner loop starts, so the resolver's value is what ever applied. But it was the first thing a reader met and it contradicted both the resolver twenty lines below it and `docs/architecture-overview.md`. Both sides now read one `LANE_CAP_DEFAULTS` carrying identical values, so they cannot drift.

The constant is **frozen, inner objects included**, because the field *aliases* it rather than copying — and the runner spec already writes to that field through an `as any`, so one such write would otherwise corrupt the default for the whole process. A regression test pins the two together and **was verified to go red** when a literal is reintroduced on either side.

## Relationship to #2162

This is the last outstanding item from that epic's audit; it does **not** close it. #2162 should close only after this merges, since three of its four ADRs currently contradict their own implementation status. (Note the epic's text says ADR-046..049 — the four decisions landed as ADR-048..051 under #2082's number allocation. Same decisions, renumbered.)


piotrswierzy:
## Review: ✅ Approve

CI 8/8 green including Integration Tests, and it merges cleanly into current `main` (verified by a real trial merge, not `merge-tree`).

### The one code change is inert, and I checked rather than took it on trust

`SyncJobRunner.laneCaps` initialised `fan-out: 1/1` while `resolveLaneCaps()` returned the post-#2609 `8/4`. I traced whether the stale value was ever live: `onModuleInit` calls `this.laneCaps = this.resolveLaneCaps()` **unconditionally**, before `startRunner()`, and that resolver already falls back to `8/4`. So the field initialiser was dead — misleading a reader, never throttling anything. The new `LANE_CAP_DEFAULTS` carries byte-identical values, so effective behaviour is unchanged.

Freezing it (inner objects included) is the right call given the field *aliases* the constant rather than copying, and the runner spec already writes to that field through an `as any` — one such write would otherwise corrupt the default for the whole process. The regression test pinning both sides was verified red when a literal is reintroduced.

### The ADR status corrections are honest about what is *not* built

Each line names the gap rather than claiming the whole ADR: 048's decision 6 (nothing records a `RateLimitTimeoutError`, so back-pressure never shrinks the next budget), 049's decision 1 (webhook path only), 050's decision 6 (met for `bulk` alone). Every issue number was verified against git history. That is the difference between marking an ADR Accepted and marking it Done.

### The false claim it retires is a real one

`docs/architecture-overview.md` asserted *"PrestaShop still refuses `adjustInventory` outright"*. The adapter implements it and refuses three *named configurations* instead. Citing the landing PR rather than an issue is correct here — neither #2077 nor #2369 appears in any commit on `main`, so the PR is the only citable record.

### Sequencing note

This unblocks closing epic **#2162** (16/16 sub-issues complete). I deliberately held that closure until now: three of its four ADRs still read `Status: Proposed` on `main`, and closing the epic while its own ADRs contradicted their implementation status would have retired the only ticket tracking that. Closing it once this lands.

PR #3018, opened 2026-09-08
feat(fulfillment): reconcile a shipped work whose dispatch relay never landed (#2728)

### LINKED ISSUE #2728: Reconcile sweep for a shipped fulfilment work whose dispatch relay never landed
Split out of **#2401**, which narrowed its AC4 rather than absorbing this. Filed with the reasoning so it is a decision rather than a gap.

## The gap

#2401 makes the dispatch-relay claim releasable: a failed relay releases `dispatchRelayedAt`, so the slot is not permanently burnt. That is necessary and it is not sufficient.

**Releasing the claim does not, by itself, cause a retry.** `IFulfillmentProgressService.record()` burns the `(workId, idempotencyKey)` claim **before** reporting the relay intent, and #2400 is explicit that a `duplicate` returns **no** intent — deliberately, since re-emitting one would re-fire a relay. So replaying the *same* vendor event after a failed relay produces nothing to re-drive.

Freeing the slot only helps a **subsequent, differently-keyed** event. If the holder's next progress event is `delivered` rather than another `shipped`, the dispatch relay is simply never sent.

## Why the obvious fix is wrong

Releasing the `(workId, idempotencyKey)` claim alongside `dispatchRelayedAt` would make the retry literal — and it reverses #2400's core design. That key is **permanent memory**, not a held slot; #2400 argued the distinction explicitly against the partial-unique precedents (`reservations` on `status = 'held'`, `order_changes` on the open statuses). Un-burning it lets a replayed vendor event **re-move counters**, trading a missed relay for corrupted quantities.

So the recovery has to come from a source other than the event stream.

## Scope

A reconcile pass over work that is shipped but whose dispatch relay never landed — roughly `dispatchRelayedAt IS NULL AND <shipped>` — re-driving the relay through the existing claim discipline.

Decisions it has to make, none of which belonged in #2401:

1. **Cadence and budget**, following `runBoundedSweep`'s family rules: a budget derived from drain rate rather than copied, a cursor that advances only when the page fully succeeded, and a cycle-scoped child idempotency key. Note the distinction `bounded-sweep.ts` draws between the scan-offset family and taxonomy's frontier-as-query — a pass whose candidates **leave the set when repaired** is frontier-shaped, and an advancing offset over a shrinking set steps over rows (the #2317 and #2346 reasoning).
2. **Lane** by cost of starvation (ADR-050), not I/O shape.
3. **Re-drive through `claimDispatchRelay`**, never around it — the claim is the serialization point between the sweep and any concurrent trigger, exactly as `waybillRelayedAt` is between the status poll and the carrier webhook (#1947).
4. Whether an age bound is needed so a permanently-unrelayable work becomes observable rather than retried forever (the #2346 shape, where fail-closed plus no bound means the set never drains and the stuck state is invisible).

## Acceptance criteria

- [ ] A work shipped with an unlanded dispatch relay is re-driven without replaying the vendor event.
- [ ] The `(workId, idempotencyKey)` claim is **not** released — proven by a test that fails if it is.
- [ ] Re-drive goes through `claimDispatchRelay`, so a concurrent trigger and the sweep cannot both relay.
- [ ] A work that can never be relayed is observable rather than silently recycled.

## Ownership

Adjacent to **#2402** (shipment bridge, `shipments.fulfillmentWorkId`) and **#2398**'s `FulfillmentStatusSource`, both of which sit on the shipped-progress path. Natural owner is whichever lands that observe path.

## Related

- **#2401** — releases the claim; narrowed AC4 and named this.
- **#2400** — the permanent-memory claim and the no-intent-on-duplicate rule.
- **#1947** — `waybillRelayedAt`, the claim-as-serialization-point precedent.
- **#861** — still open for the per-*destination* half; #2401 answers it only at work grain.

piotrswierzy:
## Summary

- **Closes the recovery gap #2401 left open.** Releasing the dispatch-relay claim frees the *slot* and causes no retry: `IFulfillmentProgressService.record` burns the `(workId, idempotencyKey)` progress claim **before** reporting the intent and returns no intent for a duplicate, so replaying the same vendor event produces nothing to re-drive — and a second, differently-keyed event arriving *while* the first relay is failing answers `already-relayed`, so recovery would need a third. If the holder's next event is `delivered`, there is no third and the marketplace is simply never told the order shipped.
- **`fulfillment.work.relaySweep`** recovers it from OL's own rows instead of the event stream: an hourly, per-run-locked, frontier-shaped pass over works reported SHIPPED whose `dispatchRelayedAt` is still NULL, re-driven through the **unchanged** `relayDispatch` so `claimDispatchRelay` stays the serialisation point.
- **The progress claim is NOT released**, and a guard fails the build if that ever changes.

## Related issues

Closes #2728

> ⚠️ **The issue comment's SECOND gap is not closed by this PR** — see *Deviations* below. Please decide whether to reopen or file a follow-up before merging, since `Closes #2728` will close the whole issue.

## What was decided (the issue's four decisions)

| Decision | Answer |
|---|---|
| **1. Cadence & budget** | Hourly at `50 * * * *` (offset from #2712's `35`, same lane + scope). Budget **derived, not copied**: 25/run. #2712 uses 200 because one unit is a bounded local write; `bounded-sweep.ts` derives 100 for a child doing a platform sync. One unit *here* is a lifecycle relay fanned out to N participant adapters, sequentially, in-process. **Frontier-as-query, no cursor** — a repaired work leaves the set by acquiring `dispatchRelayedAt`, so an offset would step over rows (the #2317 / #2346 / #2712 reasoning). Sweep *primitives* reused (`resolveSweepBudget`, `resolveSweepLockTtlMs`); the offset machinery is not, and no `MasterSweepKind` member is added. |
| **2. Lane** | **`bulk`**, on a *stronger* argument than #2712's rather than a copied one. Every candidate has by construction been unrelayed for at least the grace window, so a lane-slot's delay adds minutes to a condition measured in hours. And unlike that reaper — which "makes no platform call and does its work in bounded local writes" — each candidate here is heavy outbound work, which is the last thing that should hold a slot from a queued `fulfillment.work.dispatch`. |
| **3. Re-drive through `claimDispatchRelay`** | The sweep resolves candidates and hands each to the **unchanged** `IFulfillmentDispatchRelayService.relayDispatch`. It claims nothing of its own — a second writer of that column would defeat the guarantee. Only a *transient* failure releases and stays on the frontier; a structurally unsupported relay keeps the claim and leaves for good. |
| **4. Age bound** | **Needed, and it is the escalation.** Past `OL_FULFILLMENT_RELAY_STUCK_AFTER_MS` (24 h, clamped `[1 h, 90 d]`) a candidate is counted on the job result and `error`-logged under `fulfillment_relay_reconcile_stuck` — and is **still re-driven**, since `adapter-unresolved` is transient by #1947's own classification and withholding would make a recoverable condition permanent. A page where nothing could be relayed is surfaced separately (`fulfillment_relay_reconcile_page_all_stalled`), because candidates are oldest-first and a permanently-failing one keeps its place at the head. |

## Deviations from the issue's stated reasoning

Both are deliberate; neither is a silent departure.

### 1. The comment's second gap (work-grain waybill relay) is **not implementable**, and its premise is partly wrong

The comment asks for a waybill relay at work grain under its own claim. **There is no waybill at work grain, from any source:**

- `FulfillmentShippedEvent` carries no tracking number.
- `FulfillmentProgressSnapshot` carries `work`, `externalWorkId`, `lines`, `observedAt` — no tracking number.
- `FulfillmentWork` has no tracking column, and no executor adapter reports one (`OlFulfillmentExecutorAdapter` does not even implement `FulfillmentStatusSource`).

And the stated premise — *"a router-fulfilled work has no `Shipment` row for it to key on"* — is **wrong for the `ol_managed_carrier` topology since #2402**: `ShipmentDispatchService` resolves the work and stamps `shipments.fulfillmentWorkId`, so such a work **does** have a `Shipment`, and #1947's late-waybill backfill (keyed on `Shipment.waybillRelayedAt`) already relays its tracking number. The premise holds only for a 3PL-executor work — and there OL holds no waybill at all.

So a work-grain waybill claim would, in every reachable case, relay either **a duplicate** or **nothing**. Building it would first require widening the executor port to carry a tracking number, which `fulfillment-executor.port.ts` explicitly defers to Wave 4 — and shipping the claim now would be a grain the system cannot honour, the ADR-045 `packGrain` failure the progress-event vocabulary names by name.

### 2. AC4 is an **age** bound, not #3013's failure **counters**

#3013 (merged one commit before this branch) solves the identical defect *one grain down*, on `shipments`, with five counter columns and a count predicate. Copying it here was considered and declined:

1. **There is no operator surface to render it on.** `FulfillmentWorkView` — the only operator-facing projection of a work — *deliberately excludes* relay hygiene (`fulfillment-work-view.types.ts:16`), pinned by a type test. #3013's counters earned their columns because `/shipments` already existed to show them. Five columns with no reader is dead data.
2. **The age is already persisted and queryable** — `fulfillment_progress_claims.claimedAt` + `dispatchRelayedAt IS NULL`, no migration and no second source of truth. The exact SQL is in the handler docblock.
3. **#2346 set this precedent for this AC shape** — a second, AGE bound whose escalation is "counted on the job result and logged".

## One contract note narrowed rather than contradicted

The frontier reads `fulfillment_progress_claims.eventKind`, documented *"for forensics. Never read as state"*. `FulfillmentProgressService.apply`'s `shipped` arm writes **nothing** to the work row — `fulfillment-progress-event.types.ts` says so in terms — so the claim row is the **sole** trace that a work shipped. The state that note protects is the WORK's, which the two axes own; a claim row is *evidence of what arrived*, which the ORM entity states explicitly.

A `shippedAt` column on `fulfillment_works` was declined on three counts: a **sixth** writer on a table whose repository header names five and warns against an unnamed one; a second source of truth; and it is backfillable only from `eventKind = 'shipped'` anyway — so the same read happens once in a migration, while every work shipped *before* that migration stays permanently invisible to the sweep, which is precisely the state AC4 exists to remove.

## Test plan

```bash
pnpm lint         # 0
pnpm type-check   # 0
pnpm test         # 0
```

**AC2 is proven by a test that genuinely goes red** — verified both ways during development:

- `libs/core/src/fulfillment/__tests__/no-progress-claim-release.spec.ts` — adding a `release()` to `FulfillmentProgressClaimRepositoryPort` fails 2 of its 4 assertions; adding a `.delete()` to the claim repository fails the *other* 2. (Both mutations were applied, observed red, and reverted.)
- `apps/api/test/integration/fulfillment-relay-sweep.int-spec.ts` asserts the claim row survives a real run against real Postgres, and that the burnt key is still un-reclaimable.

**Integration tests are compiled but NOT executed** — Docker is unavailable on the machine this was written on. `apps/api/tsconfig.type-check.json` excludes `test/`, so the new int-spec was type-checked explicitly against that config with `include: ["test/**/*"]` (exit 0). CI runs them. The parity spec (`fulfillment-work-migration-parity.int-spec.ts`) already covers `fulfillment_progress_claims` and compares `indexdef` across both schema sources, so it is what will verify the new index matches between the migration and `synchronize`.

**One pre-existing flake observed**: `apps/web/src/features/fulfillment/components/order-fulfillment-tasks-panel.test.tsx` › *"should refresh itself after a stale-token 409"* failed once under the full concurrent `pnpm test`, then passed in isolation and on a full re-run (exit 0). This PR touches **zero** files under `apps/web`.

## Quality gate

- [x] `pnpm lint` passes (zero errors)
- [x] `pnpm type-check` passes (zero errors)
- [x] `pnpm test` passes (all unit tests green)
- [ ] *Optional, Docker required:* `pnpm test:integration` — **not run** (no Docker); see Test plan.

## Migrations

- [x] One migration: `1877000000000-add-fulfillment-progress-claims-shipped-index.ts` — a **non-unique partial index** serving the frontier, declared under the same name on `FulfillmentProgressClaimOrmEntity` (the `synchronize`-parity discipline). Non-unique is load-bearing: `claim`'s bare `ON CONFLICT DO NOTHING` is safe only while the composite PK is the table's one uniqueness declaration. Timestamp is strictly greater than the current tail (`1876000000000`); `check-migration-timestamps.mjs` passes under `check:invariants`. `migration:show` was **not** run — it needs a live database.

## ADR

- [x] No new architectural decision — this implements ADR-054's shape within the existing #2712 / #2400 / #2401 precedents. `docs/architecture-overview.md` § 26 Fulfillment carries the new bullet.

## Notes for the reviewer

- **Zero new allow-list entries** in `scripts/check-no-injection-contracts.mjs` or `barrel-purity.spec.ts` — which is the evidence the report-don't-perform split is placed correctly rather than routed around. The core service injects the work repository and nothing else; the worker handler composes it with the `orders` relay.
- **Inert on every install today**: `IFulfillmentProgressService.record` still has no production caller, so no `shipped` claim exists and the frontier returns nothing (the #2400 / #2955 posture — built, tested, and stated as inert).
- The repository query was rewritten mid-review from raw-quoted SQL to `alias.property` paths after re-reading `docs/lessons.md` § *"TypeORM `ORDER BY` must use property paths, not raw quoted SQL"* — it uses `limit`, not `take`, and `getRawMany`, not `getMany`, both for the reasons that file gives.


piotrswierzy:
## Review: ✅ Approve — CI 9/9 green including Integration Tests

I verified the two things this PR turns on rather than taking them on trust.

### AC2's guard is real — red-checked

The issue demands the claim "is not released — **proven by a test that fails if it is**". A guard that cannot fail is worse than none (I found exactly that defect in #3014's mirror script earlier today), so I injected a `release()` into `FulfillmentProgressClaimRepositoryPort` and re-ran: **2 of 4 assertions fail**, and 4/4 pass on restore. It works.

Making it **structural rather than behavioural** is the right call and the docblock argues it well: a behavioural test observes only the paths it drives, whereas the property that must hold is that *no code anywhere in this context can release a claim*. The three assertions cover the port declaration, a repository reach-around, and an entity-manager delete — and it follows the repo's established prohibition shape (`proposal-never-issues.spec.ts`, `no-second-proposal-mechanism.spec.ts`).

### The migration and the entity agree

I nearly filed this as BLOCKING — my `--stat | tail` truncated the file list and I read the tail as the whole, concluding a partial index had been added with no migration. Withdrawn: `1877000000000-add-fulfillment-progress-claims-shipped-index.ts` exists and matches the entity exactly — same name, same column, same predicate, non-unique on both sides.

Both docblocks record *why* the duplication is mandatory (the harness builds schema by `synchronize` and never runs migrations, so a migration-only index would exist in production and silently not in tests), and both record why it must stay **non-unique**: `FulfillmentProgressClaimRepository.claim` uses a bare `ON CONFLICT DO NOTHING`, safe only while the composite PK is the table's one uniqueness declaration. A UNIQUE index here would make an unrelated conflict report "already claimed" and silently suppress a progress write forever. That is precisely the trap, named at the point where someone would spring it.

The index's residual cost is stated rather than hidden — a relayed work keeps its claim row in the index and is filtered by the join, so the scan lengthens with history — along with why the honest fix (`shippedAt` on `fulfillment_works`) is declined: a sixth writer on a five-writer table and a second source of truth for one fact.

### Guards confirm the split landed in the right place

`check-no-injection-contracts` reports **1 contract, 1 watched context, 1 present** — zero new allow-set entries, in either that script or `barrel-purity.spec.ts`. That is the evidence the report-don't-perform boundary is correct, not merely permitted. `check-migration-timestamps` passes.

### Deviations, both flagged rather than silent

The agent surfaced two, and I verified both:

1. **`Closes #2728` is correct.** Its four ACs are the contract and none mentions waybills — that half came from #2401's note *suggesting* this sweep as its home. I filed **#3019** for it, carrying two findings: no work-grain shape carries a waybill at all (zero `trackingNumber` occurrences across `FulfillmentShippedEvent`, `FulfillmentProgressSnapshot`, the status-source capability and the contract fixtures), and the architecture doc's claim that *"a router-fulfilled work does not have"* a `Shipment` is **false** since #2402 — `ShipmentDispatchService` stamps `claimFulfillmentWorkLink`, so #1947's backfill already covers `ol_managed_carrier`.
2. **AC4 as an age bound rather than #2073's counters.** Correct: #2073's five columns earned themselves because `/shipments` renders them, whereas `FulfillmentWorkView` deliberately excludes relay hygiene — counters here would be dead data, while the age already sits on the claim row and is directly SQL-queryable. #2346 is the precedent for that AC shape.

Both `/tech-review` passes caught something real, which is what those passes are for — a raw-quoted `ORDER BY` that `docs/lessons.md` warns against by name, and a tautological AC2 test that asserted a `Set` the handler never touched.

Issue #3006, opened 2026-09-08
[TASK] DX — the fiscal lane has never issued a document under measurement, and it is one stub away


norbert-kulus-blockydevs:
## Problem / Context

**No fiscal document has ever been issued under measurement, and the lane that issues them is capped at one per connection on a constant labelled ILLUSTRATIVE.**

`fiscal: { total: 2, perScope: 1 }` (`apps/worker/src/sync/sync-job.runner.ts:106`), and `results-lane-caps-2026-09-07.md` § 1 lists `fiscal` as **ILLUSTRATIVE** — a guessed value, never derived from measurement like `bulk`'s was.

`invoicing.issue` appears in exactly two report lines in the whole campaign, both structural. It was used once, in F7, as a **deliberately invalid payload** that fails validation before doing any work, purely to time the claim. No document was created.

`results-lane-caps-2026-09-07.md` § 4.5 offers the arithmetic `N × T / C` and then states that **T is left unknown rather than filled in with a plausible number**, concluding the lane is "the likeliest permanent holdout".

**That conclusion should not stand, because T is not unknown to us and the stub pattern already exists.** This repository records an inFakt/KSeF end-to-end confirmation at roughly **90 seconds**, and `InfaktInvoicingAdapter` blocks on a bounded async-task poll by design (#1763) — so a real provider genuinely can take that long. And `stubs/allegro/server.mjs` already demonstrates the pattern: a stub with pinned latencies, requiring no legal act and no provider account.

The stakes are a straightforward multiplication, and it lands on both sides of feasible:

| Provider latency T | `perScope: 1` | 5 000 documents/day |
|---|---|---|
| 90 s | 40 docs/h | **~125 hours of work per day — impossible** |
| 10 s | 360 docs/h | ~13.9 h |
| 2 s | 1 800 docs/h | ~2.8 h |

A customer under a per-order fiscal obligation asks what our invoicing throughput is, and today the honest answer is that nobody knows.

**A second, operator-visible consequence is documented and unmeasured.** § 4.5 states that `invoicing.issue` and `fiscalization.register` share one per-order lock (#2047, `invoiceIssueLockKey`), and that an operator's `POST /invoices/bulk-issue` over N **distinct** orders is therefore serialised **entirely by the lane cap** — one document at a time — even though the lock itself would permit all N. If true, that is a defect an operator experiences, not merely a capacity figure.

## Proposed Solution

[...]

PR #3023, opened 2026-09-09
perf(2840): sync_jobs at 10M/40M rows + bulk-issue lane-cap concurrency check

norbert-kulus-blockydevs:
## Summary

Two independent measurements for the OpenLinker #2840 performance campaign, run on the second dedicated machine.

- **`results-sync-jobs-scale-2026-09-09.md`** — seeds `sync_jobs` to ~10M and ~40M rows and re-measures the claim query (`FOR UPDATE SKIP LOCKED`) and `GET /sync/jobs` at both sizes. Headline: the claim query stays cheap regardless of table size (index-bound on queue depth, not row count); `GET /sync/jobs` degrades from F5's 33-35ms baseline to ~3s at 10M and ~7.2s steady-state at 40M (unindexed full-table scan + full-table COUNT). Also documents a real seeder failure: the seeder's own `VACUUM ANALYZE` hit the database's 30s `statement_timeout` on the 24GB table.
- **`results-bulk-issue-lane-cap-2026-09-09.md`** — replaces the originally-scoped fiscal-throughput task (descoped same-day by the campaign lead to #3006, which needs a day-of-work `InvoicingPort` stub not built here). Tests whether `POST /invoices/bulk-issue` is serialised by the fiscal lane cap, as `results-lane-caps-2026-09-07.md` §4.5 claimed. **Refutes it**: the endpoint never enqueues a `sync_jobs` row at all — it's a plain sequential `for` loop calling `InvoiceService.issueInvoice` directly — so the lane cap is never consulted on this path at all. Confirmed live via a detector-verified poll of `sync_jobs` running-state during a real bulk-issue call.

`lib-test.sh` reports 165 passed / 1 failed on this exact commit (not the pre-registered 175/0). Diagnosed: a test-only gap (`post_guard_containers_stable` was added to `run_post_guards` after this one specific test was written, and the test never seeds that guard's precondition) — confirmed unreachable by any real scenario script, no bearing on either measurement.

## Test plan

- [x] `MACHINE-SPEC-2026-09-09.md` records host/disk/docker versions cited by both reports
- [x] `lib-test.sh` run twice, deterministic 165/1 result, root-caused before trusting the harness
- [x] Both reports pre-register acceptance criteria before data was read
- [x] Every figure labelled measured / derived / extrapolated; each report closes with what it did not establish
- [x] Commit signed with `git commit -s -S` (GPG signature verified)

### DIFF
--- a/perf/openlinker-throughput/MACHINE-SPEC-2026-09-09.md
+++ b/perf/openlinker-throughput/MACHINE-SPEC-2026-09-09.md
@@ -1,0 +1,48 @@
+# Machine specification — perf-programme-2840 (second machine)
+
+Recorded: 2026-09-08T23:19:12Z
+
+## Host
+```
+CPU(s):                               22
+On-line CPU(s) list:                  0-21
+Vendor ID:                            GenuineIntel
+Model name:                           Intel(R) Core(TM) Ultra 7 155H
+Thread(s) per core:                   2
+Core(s) per socket:                   11
+Socket(s):                            1
+---
+               total        used        free      shared  buff/cache   available
+Mem:            15Gi       3.3Gi        10Gi       122Mi       2.3Gi        12Gi
+Swap:          4.0Gi          0B       4.0Gi
+---
+Linux NORBERT-PRACA 5.15.167.4-microsoft-standard-WSL2 #1 SMP Tue Nov 5 00:21:55 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux
+---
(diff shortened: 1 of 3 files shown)


norbert-kulus-blockydevs:
Superseded by #3026, which consolidates the campaign from all three machines.

Nothing is lost: every commit here is an ancestor of #3026
(`466a9d7f3` → `ed8ceaec2`), and merging this branch into that one reports
`Already up to date` with zero files changed. The branch stays in place as a
reference.

Reviewing #3026 covers this work.

Issue #2593, opened 2026-08-27
[TASK] Integration/PrestaShop — wire the existing bulk product read into the catalog sweep, with cadence and page budget


norbert-kulus-blockydevs:
## Problem / Context

`PrestashopProductMasterAdapter.getProducts(filters)` (`prestashop-product-master.adapter.ts:305`) already implements a bulk read: `display=full` by default (`prestashop-query.builder.ts:71-72`) plus `batchGetOrCreateInternalIds`. **It has no caller anywhere in the repo** — the only hit outside its own body is `searchProducts` in the same file (`:767`).

The catalogue sweep instead walks `listExternalIds` (`display=[id]`, `:362`) and then `getProduct` per id. Measured cost of the bulk path: **3 requests, 0.38 s, 1.33 MB for 100 fully-hydrated products**.

Two small gaps block the swap: `filter[id]` joins ids with a comma, which PrestaShop reads as a **range** rather than a list (an OR list needs a pipe), and the builder emits no `sort` parameter.

**Cadence belongs in this issue, not a separate one.** After the adapter fix a 100-product tick is 471 s while the cron runs every 20 min with `SWEEP_BUDGET_DEFAULT = 100` (`apps/worker/src/sync/bounded-sweep.ts:46`), so the cycle is bounded by our own configuration. Shipping the bulk read without retuning cadence and budget produces no visible improvement and invites declaring the work done.

## Proposed Solution

Pipe-join `filter[id]`, add `sort`, wire `getProducts` and a paged `stock_availables` read into the sweep, and retune `pageLimit` plus cron cadence in the same change.

## Classification

**Type**: Integration / Worker
**Layer**: Infrastructure / Application
**File(s)**: `libs/integrations/prestashop/src/infrastructure/http/prestashop-query.builder.ts`, `prestashop-product-master.adapter.ts`, `apps/worker/src/sync/bounded-sweep.ts`, `apps/worker/src/sync/handlers/master-product-sync-all.handler.ts`, `libs/integrations/prestashop/src/infrastructure/scheduler/prestashop-scheduler-tasks.ts`

## Safety condition — this issue is not done without it

[...]

PR #3026, opened 2026-09-09
perf(2840): campaign results from three machines - throughput, panel, catalogue, and a clean F3 rate axis

norbert-kulus-blockydevs:
Part of #2840. **No product code** — nothing under `apps/` or `libs/`.

  Consolidates the campaign from **three machines** (laptop, workstation/WSL,
  dedicated server — the server ran two sessions, hence four source branches).
  312 files against `perf-programme-2840`, of which:

  - **264 run artefacts** (`manifest.json` / `verdict.txt` / `k6-summary.json`)
    — machine-generated evidence, and the bulk of the line count
  - **39 reports** (`.md`) — the part meant to be read
  - **8 harness files, 673 insertions** — the only code to review:
    `scenarios/f1-order-ingestion.sh` (+302), `scenarios/sustained-mixed-load.sh`
    (+136), `lib.sh` (+23), `drivers/order-age-percentiles.sh` (new),
    `drivers/f1-summarize.py`, two `run-*.sh` launchers, and
    `docker-compose.lab.yml` (+21, without which disabling the master sweeps
    was not expressible)

  Neither `2840-mixed-load-fixed` nor `2840-sustained-mixed-load` is dragged in
  — both are already contained in the base and contribute zero commits.

  ## The epic's three questions

  ### 1. Orders per hour, and what binds it

  One knob changed at a time, same 900-order backlog, same ~295 s window.
  Service time is the mean `lastAttemptDurationMs` over succeeded
  `marketplace.order.sync` rows inside that arm's own window — never a summary
  p50, which spans requeued partial attempts.

  | Settings | Server | WSL | Service time | Queue grew? | Verdict |
  |---|---:|---:|---:|---|---|
  | stock (destination limit 60/min, 4 concurrent) | **1 184 /h** | 1 555 /h | 5.54 s | yes → this is the **knee** | VALID |
  | destination limit raised (6 000/min, 32) | **3 162 /h** | 3 220 /h | 2.00 s | yes → **knee** | VALID |
  | + realtime lane cap raised (16/16) | **≥10 983 /h** | ≥10 658 /h | 3.12 s | no, drained → **FLOOR** | DISCARDED |

  **What binds it, in order.** The destination shop's own rate limit first —
  60/min inflates per-order service time to 5.54 s. Raise it and service time
  falls 2.8× to 2.00 s, so the limiter really was the cost, but throughput only
  reaches ~3 200/h because the realtime lane's per-scope slots now bind. Raise
  both and the offered load runs out.

  **The defensible number is ≥1 184 /h.** The third arm is a floor, not a
  ceiling — its `DISCARDED` is `post_guard_feed_starved` stating exactly what
  the summary states: the backlog drained inside the window, so the figure
  bounds the offered load and not the system. **Nobody has saturated it yet.**

  Single-order latency (arm `latency`, VALID, n=25): **5 799 ms** p50 end to
  end, largest hop the destination create at 2 875 ms (50%).

  ### 2. Does the panel stay usable at 10k / 100k / 1M orders?

  Yes on this hardware. Worst route at 1M is `orders_list_needs_attention` at
  **162 / 173 ms** p50/p95; nothing exceeds ~170 ms. Four of seven routes are
  flat with order count.

  **The cost is the pagination `COUNT`, not the page of rows** — 143 ms mean
  per call at 1M with the needs-attention predicate, against **0.253 ms** to
  fetch the rows themselves.

  ### 3. Requests per product for a catalogue pull

  From PrestaShop's own access log over one full sweep cycle:
  **500 products, 917 variants, 422 s, 2 029 requests.**

  | requests | resource | per product |
  |---:|---|---:|
  | 1 500 | `GET tax_rules` | 3.000 |
  | 500 | `GET taxes` | 1.000 |
  | 29 | catalogue, combinations, categories | 0.058 |
  | **2 029** | **total** | **4.058** |

  Three figures that must not be collapsed into one: total **4.058**;
  catalogue read alone **0.044** (the batched sweep #2593 working as designed —
  500 hydrated products in 20 requests); **tax resolution 4.000, which is
  98.6% of the entire pull.**

  Wall clock: 100 products ≈ **84 s** (derived from 500 in 422 s, not observed).

  **What binds it is the limiter at 96% utilisation** — 4.81 req/s against a
  declared 5.00 — so request count and elapsed time are the same quantity.
  Cutting the tax calls would cut the pull by ~98% at the same limit.

  > **Condition that must travel with this figure:** this stand had no usable
  > tax rule at all until the campaign created one. These 2 000 tax requests
  > are the cost of a *correctly configured* shop. A stand without tax
  > configured reports a far lower req/product while being unable to create a
  > single order.

  ## Other results

  **A clean F3 rate axis — the one genuinely new measurement.** Every prior
  attempt moved arrival rate and `webhook_deliveries` size together, so no rung
  was comparable to any other; `results-F3-sweep-2026-09-08.md` §6.6 records
  that gap as *"named rather than claimed"*. Both rungs here start from an
  identical table (603 rows) with identical provisioning (400 VUs) and vary
  only the rate:

  | rate | requests | med | p95 | p99 | verdict |
  |---|---:|---:|---:|---:|---|
  | 300/s | 42 749 | 4.68 ms | 9.47 | 13.97 | VALID |
  | 600/s | 85 499 | 6.13 ms | 33.24 | 69.67 | VALID |

  Doubling the rate moves the median 31% and the tail **5×**. Zero errors, zero
  dropped iterations, `vus.max` never grew past the pre-allocated pool. Both
  pass `post_guard_generator_saturated`, which discarded all five rungs of the
  earlier ladder — so unlike those, these are quotable.

  **`sync_jobs` at scale (10M and 40M rows).** The claim query is flat —
  0.054 ms at 10M, 0.037 ms at 40M; the `(status, nextRunAt)` index does its
  job and history costs the queue nothing. The Jobs page is not: **33 ms →
  2.96 s → 7.22 s** at ~150k / 10M / 40M rows, on an unfiltered parallel seq
  scan plus a full `COUNT(*)`. Table plus indexes reach **24 GB** at 40M, and
  buffers read scale linearly with row count. Nothing prunes this table
  (ADR-072 / #2862); this branch supplies the growth evidence #2843 was opened
  for.

  **Two earlier claims refuted, each by its own author.**

  - Bulk invoice issuing is **not** serialised by the fiscal lane cap. It never
    creates a `sync_jobs` row at all — it is a sequential `for` loop inside the
    HTTP request. Raising `OL_LANE_FISCAL_SCOPE_CAP`, the fix a reader of the
    earlier report would reach for, would do nothing. Confirmed live and in
    source, after a positive control caught the first detector being broken.
  - The 4× F2 improvement is **not** attributable to #2609's fan-out cap raise.
    Reverting the cap to 1/1 gave 987.7 ms against 981.7 ms — flat under a 4×
    cap change. Recorded as withdrawn. The experiment was one-sided and says so:
    F2 is serial, so a per-scope cap has no queue to act on and the test could
    refute but not confirm.

  ## Cross-machine agreement, and the one divergence

  The tuned arms agree across two independent machines to **1.8% and 3.0%**.
  The stock arm diverges by **31%** (1 184 vs 1 555 /h) and *why* is not
  established. **Do not average, pool, or present these as a range** — the
  lower figure is the defensible one, and both reports say so.

  For the same reason both machines' F3 reports are kept as separate files
  (`results-F3-2026-09-09-epyc32.md`, `-wsl.md`) rather than merged.

  ## What this does NOT establish

  - **The throughput ceiling.** Every tuned arm on every machine ended with the
    backlog drained inside the window. All are floors.
  - **WooCommerce as an order destination.** Every order fails on a *local*
    mapping lookup — `No WC product mapping for OL product ol_product_…` —
    because the seeder writes a string external id where
    `WooCommerceOrderProcessorAdapter.resolveLineItems` needs a numeric WC
    product id. **No HTTP request ever reaches WooCommerce**, so this is not
    evidence about that integration; it is a stand gap, and it means WC has
    never been measured as a destination. F1 disables WC before its windows to
    work around it. Two follow-up issues are drafted for this and for the F5
    seeding defect below.
  - **F5's scaling arms in this run.** The scenario was asked for
    10k / 100k / 1M and seeded 1 000 000 `order_records` in all three
    (`f5.targetOrders` vs `f5.rowCounts.order_records` in each manifest), so
  10k / 100k / 1M and seeded 1 000 000 `order_records` in all three
  (`f5.targetOrders` vs `f5.rowCounts.order_records` in each manifest), so
  the three arms measured one dataset three times — which is why they agree
  to within 2%. All three still report VALID: nothing checks the seeded size
  against the target. **The 10k and 100k points must come from the
  2026-09-08 run**; the 1M point is sound and agrees with it within 5%.
- **How fast `sync_jobs` grows in a real shop**, so the mapping from shop
  size to "when does the Jobs page hit 7 s" is an estimate, not a
  measurement.
- **Multi-replica behaviour.** Lane caps are enforced per *process*
  (ADR-050), so N replicas multiply every cap by N. Untested beyond F4.
- **Documents per hour at a real invoicing provider** — stub deferred to
  #3006.

## Deviations a reviewer should see rather than discover

- **One commit used `--no-verify`** (workstation/WSL). That worktree had no
  `node_modules`, so the hook died on `eslint: not found`. The commit carries
  53 files, all `json`/`txt`/`md`, **zero files under `apps/` or `libs/`**,
  and the reason is recorded in the commit body. Verified independently here.
- **Merge commits on this branch are not GPG-signed** (`--no-gpg-sign`). DCO
  sign-off is present throughout. Signatures to be corrected before merge.
- **Two merge conflicts, both resolved on content rather than mechanically:**
  the host spec (the peer's version is a strict superset — the same 90 lines
  plus an independent re-verification measuring steal time at 0.0000% — and
  was taken), and the duplicate F3 filename (split per machine, above).

## Reading order

1. `results-SUMMARY-2026-09-08-epyc32.md` — the three questions above
2. `results-F1/F2/F3/F5/F10-*.md` — one file per flow
3. `campaign-notes-epyc32-2026-09-08.md` — stand repairs and findings

Every figure is traceable. Each run directory carries `manifest.json` (image
SHA, lane caps, replica count, table sizes at window start and end, Postgres
settings) and `verdict.txt` with `status=VALID|DISCARDED` and the guard that
refused. Raw `k6-raw.json` is gitignored (704 MB on disk).

**A `DISCARDED` verdict is a fact about the run, not a failure.** Read
`verdict.txt` before quoting anything from a directory.

## Still running

F7 (lane starvation), F4 (claim contention, 3 replicas) and F8 (lane caps)
are executing now — never measured on this hardware before. They land as a
follow-up commit on this branch and the PR will update itself.


norbert-kulus-blockydevs:
Supersedes #3023 (contained, 466a9d7f3)

Issue #580, opened 2026-05-09
[D5] [MEDIUM] PromptTemplateChannelValues is a closed union of platforms


piotrswierzy:
Parent: #550 — Modularity Thread D. Tracking: #546. Doc: PR #545.

**Severity**: MEDIUM · **Catalog**: BE-10

**Observation**: `PromptTemplateChannelValues = ['prestashop', 'allegro'] as const` in core AI types. `OfferMappingFilters.platformType` doc says "e.g. 'allegro'". Several places in core treat the platform set as fixed.

**Evidence**: `libs/core/src/ai/domain/types/prompt-template.types.ts:25`; `libs/core/src/listings/domain/types/offer-mapping.types.ts:19`; `libs/core/src/integrations/infrastructure/adapters/adapter-registry.service.ts:29,41`.

**Why it blocks OSS plugin contribution**: AI prompt templates can only be authored against the two known channels. A plugin author cannot extend the channel enum without editing core.

**Recommendation**: Replace closed channel union with `string` (validated against the live registry of registered platforms). Same fix-pattern as #D1 (Capability) and #D2 (EntityType).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #374, opened 2026-04-23
Duplicate migration timestamp `1790000000000` leaves `products.currency` unapplied on some DBs


piotrswierzy:
## Summary

Two migrations in `apps/api/src/migrations/` share the timestamp `1790000000000`. On at least one developer DB this collision resulted in a state where both migrations are recorded as applied in the `migrations` table, but the `ALTER TABLE "products" ADD "currency"` DDL from `AddCurrencyToProducts` never landed. Runtime reads then fail with `column "ProductOrmEntity.currency" does not exist`, blocking the Products list, Inventory list, and anything that hydrates a `Product`.

## Observed behaviour

Error surfaced during normal use of the API (product list, inventory list):

```
QueryFailedError: column ProductOrmEntity.currency does not exist
  at ProductRepository.findById (libs/core/src/products/infrastructure/persistence/repositories/product.repository.ts:32:20)
  at InventoryQueryService.buildProductMap (libs/core/src/inventory/application/services/inventory-query.service.ts:74:22)
  at InventoryQueryService.listInventoryItems (libs/core/src/inventory/application/services/inventory-query.service.ts:53:24)

QueryFailedError: column product.currency does not exist
  at ProductRepository.findMany (libs/core/src/products/infrastructure/persistence/repositories/product.repository.ts:58:31)
  at ProductsController.listProducts (apps/api/src/products/http/products.controller.ts:86:30)
```

The DB state on the affected machine:

```
SELECT column_name FROM information_schema.columns
  WHERE table_name='products' AND column_name='currency';
-- (0 rows)

SELECT timestamp, name FROM migrations WHERE timestamp = 1790000000000 ORDER BY id;
-- 1790000000000 | AddPromptTemplatesTable1790000000000
-- 1790000000000 | AddCurrencyToProducts1790000000000
```

`pnpm --filter @openlinker/api migration:show` reports both as `[X]` applied.

## Root cause

Two migration files share the same timestamp:

| File | Class |
|---|---|
| `apps/api/src/migrations/1790000000000-AddCurrencyToProducts.ts` | `AddCurrencyToProducts1790000000000` |
| `apps/api/src/migrations/1790000000000-add-prompt-templates-table.ts` | `AddPromptTemplatesTable1790000000000` |

Introduced in commit `6cec005` (`feat(products): persist Product.currency to the products table`), which picked a timestamp already in use by the prompt-templates migration merged the same day.

[...]

PR #3027, opened 2026-09-09
fix(worker): resolve AI suggestion channel from the connection, not the literal 'allegro'

### LINKED ISSUE #2202: [BUG] Worker - resolve the AI suggestion channel from the connection, not the literal 'allegro'
## Problem / Context

`apps/worker/src/sync/handlers/marketplace-offer-create.handler.ts:173` passes a string literal:

```ts
const result = await this.contentSuggestion.suggestDescription({
  productId,
  channel: 'allegro',
  tone: payload.descriptionTone,
});
```

That is the AI-description branch of the `marketplace.offer.create` handler, which serves **every** marketplace offer create, not only Allegro. So an Erli offer with `generateDescription` set is written by the Allegro prompt template, whose seeded text encodes Allegro's own block rules (*"Allegro uses block-formatted descriptions (no free-form HTML beyond `<h2>`, `<p>`, `<ul>`/`<li>`, `<b>`)"*, `apps/api/src/migrations/1790000000001-seed-prompt-templates.ts:47`) and its character limit.

`PromptTemplateChannel` is open-world (`string`, #580) precisely so a channel resolves from the connection's `platformType`, and the content editor already does exactly that: `ContentEditor` passes `channel.platformType` verbatim. The worker handler is the one caller that hardcodes instead.

**Correction (post-#3027 review):** there is no channel→master fallback anywhere in the chain. `PromptTemplateService.render` does an exact `(key, channel)` lookup and throws `PromptTemplateNotFoundException` when nothing is published for that channel — `ContentEditor` benefits from no fallback either; it simply targets platforms that already have a published row. So resolving the channel correctly is a *precondition* for the right template to be used, not a safety net on its own — a platform with no seeded `offer.description.suggest` row for its channel gets no AI suggestion at all, not a degraded one.

Consequence today: an Erli AI description is written against Allegro's constraints. It is not a hard failure - the text is usually still usable - which is why it has gone unnoticed.

## Proposed Solution

Resolve the channel from the connection the offer is being created on, matching `ContentEditor`'s approach: read the connection's `platformType` and pass it through. Since there is no channel→master fallback, an `erli` seed migration is required alongside the fix for Erli to actually gain a working AI description rather than losing the (imperfect but usable) Allegro-templated one it gets today.

Update `apps/worker/src/sync/handlers/__tests__/marketplace-offer-create.handler.spec.ts:249`, which currently asserts the literal.

## Classification

**Type**: Integration
**Layer**: Application (worker handler)
**File(s)**: `apps/worker/src/sync/handlers/marketplace-offer-create.handler.ts`, `apps/worker/src/sync/handlers/__tests__/marketplace-offer-create.handler.spec.ts`

## Dependencies

None.

## Assumptions

- The handler already has the connection id in its payload, so no extra lookup is needed beyond resolving the connection. If it does not, a single connection read is acceptable on a path that is already making an LLM call.

## Acceptance Criteria

- [ ] The handler passes a channel derived from the connection, never a literal
- [ ] An Erli offer create with `generateDescription` resolves the `erli` prompt template — which requires seeding it, since there is no channel→master fallback to rely on instead
- [ ] The spec asserts the resolved value rather than the literal
- [ ] A test covers at least two platforms so a future literal cannot pass
- [ ] `pnpm lint` and `pnpm type-check` pass with zero errors

jakubretajczykBD:
## Summary
- `marketplace.offer.create` serves every marketplace, not only Allegro, but the AI-description branch hardcoded `channel: 'allegro'` when calling `ContentSuggestionService.suggestDescription`.
- An Erli offer with `generateDescription=true` was therefore written against Allegro's prompt template (block-formatted HTML, Allegro's character limit).
- The channel now resolves from the connection's `platformType` via `IIntegrationsService.getAdapter(job.connectionId)`, matching `ContentEditor`'s approach. If the connection lookup fails, AI generation is skipped and the operator override / builder default takes over — mirroring the existing variant-lookup-failure fallback.

## Test plan
- [x] `pnpm --filter @openlinker/worker type-check` — clean
- [x] `pnpm lint` (eslint) on the two changed files — clean
- [x] `pnpm --filter @openlinker/worker test -- marketplace-offer-create.handler.spec.ts` — 18/18 passing, including new coverage for the Erli-channel resolution and the connection-lookup-failure fallback

Closes #2202


piotrswierzy (changes requested):

The mechanism here is right and I'd happily approve it on its own. What blocks it is that the change ships a behaviour regression **on its own motivating platform**, justified in-code by a statement about the system that is not true — and closes an issue whose AC-2 covers exactly that case.

All 8 CI checks are green. Both findings below were verified against `main`, not inferred.

### BLOCKING — the master-fallback claim is false

`apps/worker/src/sync/handlers/marketplace-offer-create.handler.ts` (~184–186)

> *"The backend falls back to the master template when no channel-specific prompt template is published…"*

It does not. `PromptTemplateService.render` does an exact-match lookup and throws:

```ts
const template = await this.repository.findLatestPublished(cmd.key, cmd.channel);
if (template === null) {
  throw new PromptTemplateNotFoundException({ key: cmd.key, channel: cmd.channel });
}
```

and `PromptTemplateRepository.findLatestPublished` is a plain `findOne` on `{key, channel: channel === null ? IsNull() : channel, state: 'published'}` — there is no fallback anywhere in the chain. `ContentSuggestionService`'s own docblock already says so: *"Latest published template for this (key, channel); throws if none exists."*

This matters beyond pedantry: that sentence is the **sole stated justification** for why an unknown `platformType` is safe after this change, and the next reader will trust it. `shop-product-publish.handler.ts:55` already has the correct wording one file over.

### BLOCKING — #2202 AC-2 is unmet, and Erli regresses

AC-2: *"An Erli offer create with `generateDescription` resolves the `erli` prompt template, **or falls back to master when none is published**."* Neither branch holds.

Seeded `offer.description.suggest` channels, confirmed by reading every seed migration:

| Migration | Channels |
|---|---|
| `1790000000001-seed-prompt-templates.ts` | `allegro`, `prestashop`, master |
| `1794000000000-seed-master-prompt-template.ts` | master |
| `1831000000002-seed-woocommerce-prompt-template.ts` | `woocommerce`, master |

**There is no `erli` seed.** So post-merge, on the exact platform the PR names: `channel: 'erli'` → `PromptTemplateNotFoundException` → caught by the existing AI-failure arm → **no AI description at all**, one `warn`, no failed job.

That is a user-visible regression from `main`, where the operator at least got something — the issue itself records the Allegro-templated Erli text as *"usually still usable"*. A `fix(...)` that silently stops producing what the operator asked for needs one of:

1. an `erli` seed migration in the shape of `1831000000002` (mind the synthetic-sequential timestamp rule, `docs/migrations.md`), or
2. a real channel→master fallback in `render()` — larger, changes behaviour for every caller, probably its own issue, or
3. at minimum the limitation stated explicitly here and in the PR, with AC-2 amended rather than ticked.

The false premise originated in #2202's body, not this PR — but this is where it becomes shipped behaviour, so **#2202's body wants correcting too**, or the next reader inherits it.

### SUGGESTIONS

- `shop-product-publish.handler.ts` (~51–57) justifies itself by pointing at *"the offer handler's hardcoded `allegro` channel"* — this PR makes that reference stale. It also carries the identical latent bug: that handler serves any `ProductPublisher` shop, not only WooCommerce, so a non-WC shop gets WooCommerce's prompt. Worth a follow-up issue plus a one-line docblock touch so the cross-reference doesn't rot.
- The new `'resolves the channel from the connection platformType'` spec mocks `contentSuggestion`, so it cannot see that no `erli` template exists. A test asserting the fall-through when `suggestDescription` rejects with template-not-found would make the real Erli outcome visible in the spec rather than three layers down.
- `docs/architecture-overview.md § 13 AI` records that the template is *"seeded per channel"* but not the resolution rule this PR establishes, nor the consequence that an unseeded channel yields no suggestion.

### Done well

- **`getAdapter` vs `getCapabilityAdapter` is the right call and correctly characterised** — verified: it resolves connection + registry metadata, constructs no adapter, resolves no credentials.
- Reading `platformType` through `IIntegrationsService` rather than `ConnectionPort` respects the repository-ports rule (`engineering-standards.md`); `IntegrationsModule` is already imported by `sync-worker.module.ts`, so no module-graph change and no new cross-context edge.
- Matches established worker precedent — `marketplace-fulfillment-status-sync.handler.ts:49` and `marketplace-shipment-status-sync.handler.ts:46` argue exactly this.
- Correctly gated inside `isV2 && generateDescription`, so V1 and non-AI creates pay no extra read.
- The failure arm mirrors the two existing fall-throughs exactly, so AC-9 ("AI failure never fails the job") is preserved rather than re-derived.
- The sweep is complete: no other production `channel: 'allegro'` literal remains.

---

jakubretajczykBD:
## `/tech-review` follow-up — remaining items closed

Ran a full `/tech-review` on this PR (post the two review-fix commits) and confirmed both **BLOCKING** findings from the earlier review are genuinely fixed, verified directly against `PromptTemplateService`/`PromptTemplateRepository` and the `1831000000002` seed-migration precedent — not just acknowledged in a docblock:

- **False master-fallback claim** — the handler's comment now correctly states there is no channel→master fallback anywhere in the chain (`findLatestPublished` is an exact `(key, channel)` `findOne`, throws `PromptTemplateNotFoundException` on a miss), and that the handler's own `catch` is what keeps a missing template non-fatal.
- **AC-2 / Erli regression** — closed by the `1878000000000-seed-erli-prompt-template.ts` seed, matching the WooCommerce seed's shape exactly, plus a regression test for "channel resolves correctly but the template still doesn't exist."

Two non-blocking `SUGGESTION`-level items from the original review were left open after the last commit; I've now closed both:

1. **`docs/architecture-overview.md § 13 AI`** didn't document the resolution rule this PR establishes. Added a bullet recording: channel is an exact-match key with no fallback, where each handler's channel comes from, and that a missing seed means *no* suggestion rather than a degraded one. Pushed as a follow-up commit on this branch (`93e442638`).
2. **Issue #2202's body** repeated the same false "backend falls back to master" claim the review caught in code — corrected the body directly, and tightened AC-2 to state the seed is required rather than optional.

CI was green on `6ab2df306`; the doc-only commit doesn't touch any tested code path.

norbert-kulus-blockydevs:
## Tech Lead Review — PR #3027

**Scope reviewed**: `apps/worker/src/sync/handlers/marketplace-offer-create.handler.ts`, its spec, `shop-product-publish.handler.ts` (comment-only), the new `1878000000000-seed-erli-prompt-template.ts` migration, and the `docs/architecture-overview.md` addition.

### Architecture & correctness
- ✅ **Channel now genuinely resolves from the connection, not a literal.** Confirmed on the PR branch (not just the diff): `channel: 'allegro'` is gone from the executable code path; the only surviving `'allegro'` string in the file is in a docblock quoting the *old* bug for context.
- ✅ **`getAdapter` is confirmed metadata-only** (per its own JSDoc: "`getCapabilityAdapter` — `getAdapter` is a metadata-only lookup", no credential resolution). The PR's stated trade-off ("re-fetches the same connection `executeCreation` resolves moments later... accepted since a second read is cheap on a path already making an LLM call") checks out — this is a cheap `connections` table read, not a duplicated credential decrypt or adapter construction, so there's no meaningful perf or security concern here.
- ✅ **No core ↔ integration layering violation.** `IIntegrationsService` / `INTEGRATIONS_SERVICE_TOKEN` is injected via the standard port/interface pattern, and `apps/worker/src/sync/sync-worker.module.ts` already imports `IntegrationsModule` — no new wiring gap.
- ✅ **Fallback behavior preserved.** Both new failure paths (connection lookup fails; resolved channel has no published template) fall through to `payload.overrides` exactly like the pre-existing AI-failure arm (AC-9) — verified via the new tests, which assert `executeCreation` is still called with the operator's override title in both cases. No job failure, no silent swallow into nothing when an override exists.
- ✅ **Excellent root-cause completeness.** The PR didn't stop at fixing channel resolution — it recognized that `PromptTemplateService.render` does an **exact** `(key, channel)` match with no channel→master fallback, so simply fixing the channel string would have made Erli offers resolve to `channel: 'erli'` and then throw `PromptTemplateNotFoundException` on every single generation attempt (no seeded row existed). The added `1878000000000-seed-erli-prompt-template.ts` migration is what makes the fix operationally effective rather than just structurally correct. This is exactly the kind of second-order check a "fix the literal" PR usually misses.

### Migration review
- ✅ Data-only, has both `up()`/`down()`, `down()` deletes exactly the seeded row.
- ✅ Idempotent-by-uniqueness (relies on the existing `prompt_templates` partial unique index rather than adding its own `ON CONFLICT`) — correctly documented as "would fail on re-run against a DB that already has the row" rather than falsely claiming full idempotency.
- ✅ Timestamp `1878000000000` sorts after the current `main` tail (`1870000004000` at time of review) — satisfies the ordering invariant in `docs/migrations.md`.
- 🟢 **Suggestion**: the raw SQL `INSERT` bypasses the TypeORM entity/repository layer entirely (no `PromptTemplateOrmEntity` insert, no validation of the `variables` JSON shape against what `PromptTemplateService` expects at render time). This is the established pattern for the sibling `allegro`/`prestashop`/`woocommerce` seeds too (per the migration's own comment), so it's consistent with precedent rather than a new deviation — not blocking, just worth knowing if a future template-shape change needs to touch every seed migration by hand.

### The remaining `shop-product-publish.handler.ts` hardcode
- 🟡 **Not fixed here, but honestly disclosed** — the PR updates that handler's docblock to explicitly call out that it carries the *identical* latent bug the day it serves a second `ProductPublisher` shop beyond WooCommerce, and says it's "tracked separately." That's the right call for scope discipline (this PR is titled and scoped to the marketplace/offer-create path), but I'd like to see an actual issue number cited in that comment rather than "tracked separately" with nothing to click through to — as written, "separately" could quietly mean "nowhere." Non-blocking, but please link the follow-up issue.

### Tests
- ✅ Three new cases, each isolating a distinct behavior: (1) channel resolves to `'erli'` from `platformType`, not a literal; (2) a resolved-but-unseeded channel still falls through cleanly via the existing `PromptTemplateNotFoundException` catch; (3) a failed connection lookup falls through without ever calling `suggestDescription`. This is a good instance of testing the *interaction* between two independently-plausible-looking fixes (channel resolution + template existence) rather than just the happy path.
- No integration test was added/needed — this is unit-level business logic, appropriately scoped.

### Verdict: ✅ Approve

This is a clean, well-scoped bug fix that goes a level deeper than the literal title suggests — catching and closing the "channel resolves correctly but there's still no template" trap in the same PR is genuinely good engineering. Only ask: cite the follow-up issue number for the WooCommerce-hardcode latent bug instead of "tracked separately."

piotrswierzy (commented):

Both BLOCKING findings are **properly resolved, not papered over** — and the fix took the conservative remedy rather than the risky one. My prior `CHANGES_REQUESTED` should be considered cleared on substance; one residual is worth a line before merge.

**1. The false master-fallback claim — ✅ resolved.** The docblock (~182–188) now states the opposite, correctly and in detail: *"`PromptTemplateService.render` does an EXACT (key, channel) match and throws `PromptTemplateNotFoundException` … there is no channel→master fallback"*, and names the handler's own `catch` — not a backend fallback — as what keeps it non-fatal. The stale cross-reference in `shop-product-publish.handler.ts` (~51–58) was fixed in the same breath.

**2. The Erli regression / AC-2 — ✅ resolved via remedy (a), plus (c).** `1878000000000-seed-erli-prompt-template.ts`, verified against `docs/migrations.md` rules 1–3: prefix unique across `apps/api/src/migrations/` and both Allegro plugin dirs; class suffix `SeedErliPromptTemplate1878000000000` matches the filename; strictly greater than main's tail `1877000000000`; both `up()` and `down()` present, with `down()` deleting exactly the seeded row. Column list matches `prompt-template.orm-entity.ts` exactly, and `id` being `@PrimaryGeneratedColumn('uuid')` makes its omission correct.

Two substantive docblock claims I verified rather than took on trust: `platformType: 'erli'` (`erli-plugin.ts:56`) matches the seeded `channel: 'erli'`, so resolution actually lands; and the superset claim is right — `ERLI_DESCRIPTION_FORMAT` is 9 tags including `h3`/`br`, `ALLEGRO_DESCRIPTION_FORMAT` is 7 without them. A third commit specifically corrected an earlier inverted version of that claim.

**No `render()` fallback was added, so no behaviour changed for any other caller** — the larger, riskier remedy was correctly declined.

**The issue body is corrected too.** #2202 now carries an explicit *"Correction (post-#3027 review)"* paragraph, and AC-2 is amended to *"resolves the `erli` prompt template — which requires seeding it, since there is no channel→master fallback to rely on instead."* The false premise no longer propagates.

All three suggestions also addressed — the shop-publish cross-reference touched, a `PromptTemplateNotFoundException` fall-through spec added (making the real Erli outcome visible in the spec rather than three layers down), and `architecture-overview.md § 13 AI` gained the resolution rule with the load-bearing sentence *"A missing seed is therefore not a degraded suggestion, it is no suggestion at all."* All 8 CI checks green, including Lint — which runs `check-migration-timestamps.mjs`, so the migration invariant is machine-verified.

### IMPORTANT — the bare `INSERT` will fail `migration:run` if a row already exists

The seed is non-idempotent, and this is not hypothetical for *this* channel specifically: the admin UI at `/ai/prompt-templates` accepts an arbitrary channel string (`PromptTemplateChannel = string`, open-world per #580), and **an Erli operator who hit exactly the gap #2202 describes is the most likely person to have hand-authored one as a workaround**. `ux_prompt_templates_published_channel` then refuses the seed and aborts deploy.

The docblock acknowledges non-idempotency and the WooCommerce precedent has the same shape, so this is precedent-consistent rather than novel — but *"consistent with a precedent that was never exposed to this risk"* is a weaker argument here. `ON CONFLICT DO NOTHING` makes it safe at no cost; if declined, say so in the docblock.

### SUGGESTIONS

- `shop-product-publish.handler.ts` (~56) says `"(tracked separately)"` and names no issue — and no such issue exists (I searched). An untracked "tracked separately" is how a known bug becomes an unknown one. File it and cite the number, or drop the parenthetical.
- The seed's prompt says *"Respect the 80,000-character limit"* while the declaration is `maxBytes: 80_000` and the adapter's own header flags the bytes-vs-characters caveat. Immaterial in practice; the docblock says bytes and the prompt says characters.
- The Erli prompt defaults to Polish unless the product name says otherwise, where the `allegro`/`woocommerce`/`prestashop` seeds all say *"match the product name's language"*. Defensible for a PL-only marketplace — worth being a deliberate choice rather than drift.

---

piotrswierzy (approved):

**Explicitly clearing my round-1 `CHANGES_REQUESTED`**, which was still the standing review decision — my round-2 comment did not dismiss it, so this PR has been blocked on a finding that was already resolved. Apologies for the delay that caused.

Both original BLOCKINGs remain properly resolved (the false master-fallback docblock, and the missing `erli` seed), the issue body was corrected at source, and all three suggestions landed. Nothing has regressed.

One item remains and I am not holding merge for it:

### IMPORTANT (carried) — the seed's bare `INSERT` can hard-fail `migration:run`

If a published `(offer.description.suggest, 'erli')` row already exists, the insert aborts the deploy — and an operator who hit the #2202 gap is the most likely person to have hand-authored one, since `PromptTemplateChannel = string` is open-world per #580. `ON CONFLICT DO NOTHING` closes it at no cost; if you decline, say so in the docblock, because *"consistent with a precedent that was never exposed to this risk"* is the weaker half of the argument.

Also worth one more check before merge: **`main` has advanced past `1877000000000`**, so re-run `check-migration-timestamps.mjs` to confirm `1878000000000` still satisfies the strictly-greater rule against the new tail.

Remaining suggestions, all optional: `"(tracked separately)"` in `shop-product-publish.handler.ts` still names no issue and none exists; the seed prompt says "80,000-character limit" where the declaration is `maxBytes`; the Erli prompt defaults to Polish where the three sibling seeds say *"match the product name's language"*.

---

piotrswierzy (commented):
## Tech Lead review (independent re-verification)

Re-read from scratch at `09989b9e` rather than deferring to the standing approval. Both original BLOCKINGs are genuinely closed, and closed in the right way: the false master-fallback sentence is replaced by an accurate one that now states the *consequence* (`a platform with no seeded template produces no AI description at all; the catch is what keeps that non-fatal, not a backend fallback`), the `erli` seed exists, and `docs/architecture-overview.md § 13 AI` gained the resolution rule plus the "a missing seed is not a degraded suggestion, it is no suggestion at all" statement — which is the sentence that stops the next reader re-deriving the original false premise. The stale cross-reference in `shop-product-publish.handler.ts` is corrected and the latent second-shop bug is recorded in place.

Verified rather than assumed: the migration's column list and parameter shape are byte-identical to the `1831000000002` WooCommerce precedent; `1878000000000` is strictly greater than `main`'s tail (`1877000000000`) and is a synthetic sequential prefix, so `check-migration-timestamps.mjs`'s rule-3 ordering gate is satisfied rather than merely green by accident; the class suffix matches the filename. `getAdapter` (metadata-only) rather than `getCapabilityAdapter` remains the right call, and the lookup-failure arm mirrors the existing variant-lookup fallback.

### Findings

**[IMPORTANT]** — `apps/api/src/migrations/1878000000000-seed-erli-prompt-template.ts` (`down()`, and the docblock sentence about it)
> The docblock says `down` *"still removes exactly this seeded row (harmless no-op if the conflict path skipped the insert)"*. That is false in precisely the case the `ON CONFLICT DO NOTHING` was added for. The delete predicate is `key = 'offer.description.suggest' AND version = 1 AND channel = 'erli'` — which is exactly the row that would have caused the conflict on `ux_prompt_templates_kcv_channel`. So on an install where an operator hand-authored an `erli` v1 template as a workaround, `up()` correctly declines to touch it and `down()` then **deletes it**: a migration revert destroying operator-authored data it never created. Cheapest correct fix is to make the removal conditional on ownership — e.g. only delete when `created_by IS NULL` and the prompt text matches the seeded constant — or, simpler, gate the whole pair on a marker the seed can recognise. At minimum the docblock must stop claiming the no-op. (The case where the operator's row is at a *different* version is genuinely a no-op; it is the v1 collision that bites, and that is the likely shape of a hand-authored workaround.)

**[SUGGESTION]** — same file, the `ON CONFLICT DO NOTHING` reasoning
> The docblock reasons only about a *published* hand-authored `erli` template. A **draft** `erli` v1 row collides on the same version-uniqueness index, so the seed silently skips, no published `erli` row exists, and the Erli regression this migration was written to close persists — with `migration:run` reporting success and nothing logged. Worth one sentence, because the failure is invisible by construction.

**[SUGGESTION]** — `apps/worker/src/sync/handlers/shop-product-publish.handler.ts` (~52-60)
> The corrected comment says *"No issue is filed for this yet — file one before adding a second `ProductPublisher` shop."* An instruction addressed to a future reader with no tracked artefact behind it is the weakest form of this. Filing the follow-up now and citing its number turns it into something that can surface in a backlog rather than only in a code comment nobody greps for.

**[SUGGESTION]** — test coverage
> The new specs mock `contentSuggestion`, so they still cannot observe that a channel has no seeded template. A case asserting the fall-through when `suggestDescription` rejects with `PromptTemplateNotFoundException` would make the real unseeded-platform outcome visible in the spec — I can see the exception is now imported in the spec file, so this may already be covered; if it is, disregard.

### Verdict
🔄 **Approve with changes**

Posting as a comment rather than an approval only because of the `down()` finding; everything else is advisory. The standing approval's judgement on the mechanism stands — this is one migration-body correction away from clean.

1. Fix `down()` so a revert cannot delete an operator-authored `erli` v1 row, and correct the docblock sentence that claims it is a no-op.

---

piotrswierzy (commented):
## Tech Lead re-review (`09989b9` → `b042c4f`)

Delta is a `main` merge plus one commit, `fix(migrations): guard erli prompt-template revert against operator data`. **The IMPORTANT finding from the last pass is properly closed**, and the fix is better than the minimum I asked for. The migration timestamp has moved again and is now contended by four open PRs, which is the only thing left.

### Prior findings

**IMPORTANT — `down()` destroyed operator data the `up()` never created — CLOSED.** The revert is now predicated on `created_by IS NULL AND system_prompt = $2` (the seeded text verbatim), so a hand-authored `erli` v1 row that caused the `ON CONFLICT DO NOTHING` to decline the insert survives a revert. That satisfies `docs/migrations.md`'s reversibility rule in the direction that actually matters — a `down()` may only undo what its own `up()` did.

The `created_by IS NULL` half is the weaker of the two conjuncts on paper: the column is `text NOT NULL`-less (`prompt-template.orm-entity.ts:76`, `string | null`) and `PromptTemplateService` passes `cmd.createdBy` through with a `?? 'system'` only in its log line, so the type does not guarantee the docblock's "every operator-authored template carries a real actor id". In practice it holds — `prompt-templates.controller.ts:141` and `:223` are the only two write paths and both pass `user.username` from the authenticated admin — and the verbatim `system_prompt` match is doing the real work regardless. No change needed; noting it so the claim isn't later read as a type-level guarantee.

The docblock addition is also honest about a second, separate hole it does **not** close: a colliding *draft* v1 row still makes the seed a silent no-op with no published `erli` template and no log line. Stating it rather than implying the conflict arm is always benign is the right call for a migration that cannot log.

The three original BLOCKINGs stay closed — the handler still resolves `channel` from `connection.platformType` (`marketplace-offer-create.handler.ts:196-199`) with no literal fallback, the real `erli` seed ships, and the arch-doc rule is in place. Neither worker handler was touched by this delta.

### Findings

**[IMPORTANT]** — `apps/api/src/migrations/1881000000000-seed-erli-prompt-template.ts`
> `1878 → 1881` clears `main` (whose tail is now `1880000000000`, merged with #3158) but lands on a lane **four** open PRs are already claiming: this one, **#3035** (`create-order-cancellation-signals`), **#3159** (`create-price-change-auto-applied-log`) and **#3201** (`normalise-sales-document-country-codes`). `check-migration-timestamps.mjs` diffs only against `origin/main`, so all four pass in isolation and the #374 collision surfaces on the second merge. The adjacent lanes are taken too: `1881000000001` (#3210), `1882000000000` (#3219), `1883000000000` (#3227). Claim a distinct one now — `1884000000000` upward, or `1881000000002..0005`, one step per PR — rather than discovering it at merge. (Same scan, unrelated: **#3014** is still on `1878000000000`, now below `main`'s tail and an ordering violation of its own.)

**[SUGGESTION]** — same file, `ERLI_SYSTEM_PROMPT`
> The prompt text changed in this commit ("80,000-character limit" → "80,000-byte limit"). `up()` and `down()` both reference the same constant, so this migration is internally consistent — but the `down()` predicate is now a content match against a string that has already been edited once. If the text is ever tuned again *after* this migration has shipped, the revert silently stops matching its own seeded row. Worth a one-line note in the docblock that `ERLI_SYSTEM_PROMPT` is frozen once merged, for the same reason executed migrations are never edited.

### CI

`Type Check`, `Docker Build Smoke Test`, `PHP Unit Tests` and `Detect changed paths` are green; `Build`, `Lint`, `Test` and `Integration Tests` were still **pending** at review time on run `34822150343` (`mergeStateStatus: UNSTABLE` reflects that, not a failure). The delta is a migration `down()` predicate plus doc prose, so nothing in it should move those jobs — but confirm the run completes before merging.

### Verdict
🔄 **Approve with changes** — the data-safety fix is correct and complete; claim a free migration timestamp and let CI finish.

---

piotrswierzy (commented):
## Tech Lead re-verification (head unchanged at `b042c4f8`)

Nothing in the code has moved since my last pass and `main` has not advanced, so this is a delta-only re-check: the prior finding, the migration-timestamp landscape, and CI. `mergeStateStatus: CLEAN`, `MERGEABLE`, all 8 checks green (`Test` 14m59s, `Integration Tests` 32m24s, `Lint` 5m14s) — the branch is genuinely ready apart from one number.

### Findings

**Prior finding — closed, re-verified in code.** `down()` now scopes to `"created_by" IS NULL AND "system_prompt" = $2` alongside `key`/`version`/`channel`, so a revert cannot delete an operator-authored `erli` v1 row that `ON CONFLICT DO NOTHING` correctly declined to overwrite. The docblock is honest about the hole it does *not* close (a colliding **draft** v1 conflicts on `(key, channel, version)`, the seed reports success, and the install keeps no published `erli` row with no log line). That is the right disposition — stated rather than papered over.

**IMPORTANT — `1881000000000` is claimed twice across the open PR set.** `apps/api/src/migrations/1881000000000-seed-erli-prompt-template.ts` collides with #3201's `1881000000000-normalise-sales-document-country-codes.ts`. `scripts/check-migration-timestamps.mjs` diffs against `origin/main` only, so both branches are green today and the collision materialises for whichever merges second — and per `docs/migrations.md` § Timestamp uniqueness invariant rule 1, TypeORM 0.3.17 sorts by timestamp with no deterministic tie-breaker, so the loser's `up()` can silently never run while both class names land in the `migrations` table.

**This PR should keep `1881000000000`; #3201 should move.** It is the older branch (created 2026-09-09 vs 2026-09-11), it is `CLEAN` with full green CI, and `1881` is legitimately above `main`'s tail of `1880000000000`. Current claims across the 48 open PRs: `1878`→#3014 (below tail, separately broken), `1881`→#3027+#3201, `1882`→#3219, `1883`→#3227, `1884`→#3159+#3241, `1885`→#3035. **The next free value for #3201 is `1886000000000`** — though if #3014 takes it first (my recommendation there), `1887000000000`.

**SUGGESTION** — `apps/api/src/migrations/1881000000000-seed-erli-prompt-template.ts`
> `ERLI_SYSTEM_PROMPT` is no longer only the seeded content — it is now the `down()` predicate's content match, which makes it a frozen constant. A later "improve the Erli prompt wording" edit to that template silently breaks the revert: `down()` stops matching the row `up()` wrote and deletes nothing. The docblock explains *why* the predicate matches on the text but does not say the constant is therefore immutable. One sentence in the header ("`ERLI_SYSTEM_PROMPT` is frozen — it is `down()`'s match key; improving the prompt means a NEW migration, never an edit here") converts a trap into a rule. Not blocking; the trap only fires on a revert.

### Verdict
🔄 **Approve with changes** — the code is correct and the prior finding is genuinely closed. The only thing standing between this and merge is the `1881000000000` collision, and my recommendation is that it resolves in **this PR's favour**: hold the timestamp here and move #3201 to `1886000000000`. If that call goes the other way, re-prefix this file and its class suffix together per rule 2.

---

piotrswierzy (approved):
## Tech Lead re-verification (head unchanged at `b042c4f8`)

`MERGEABLE` / `CLEAN`, all 8 checks green, already `APPROVED`. The prior finding stays closed: `down()` scopes to `created_by IS NULL AND system_prompt = $2`.

**The migration collision I flagged last round has cleared, and not by anything this PR did.** #3201 has moved off the contested value to `1880500000000`. Current map:

| Claim | PR | Status |
|---|---|---|
| `1880000000000` | — | `main` tail |
| `1880500000000` | #3201 | ✓ free |
| **`1881000000000`** | **#3027 (this PR)** | ✓ **free — no collision remains** |
| `1881000000001` | #3210 | ✓ free, sorts above #3201 |
| `1878000000000` | #3014 | ❌ below tail — must move to `1882000000000` |

**#3027 keeps `1881000000000`.** It was already the right call on the merits (older branch, clean, fully green, legitimately above the tail); it is now simply uncontested.

### Findings

**SUGGESTION — freeze `ERLI_SYSTEM_PROMPT` in the docblock.** It is now `down()`'s content match key, so a later edit to the prompt text silently breaks the revert: the `DELETE … AND system_prompt = $2` matches nothing and the seeded row survives a rollback. One line above the constant saying "changing this text breaks `down()`; add a new migration instead of editing" is enough. Not merge-blocking.

### Verdict
✅ **Approve**

---

Issue #3017, opened 2026-09-08
[TASK] Frontend — the OMS is unreachable without curl: review the configuration surface, design it, and split it


piotrswierzy:
## Problem / Context

**The OMS is fully built and effectively unreachable.** Waves 0–3a and the Wave 3b pack bench have all
merged (`ee9fe6fe4`, `c1e4090ee`), but every *operating* surface downstream is gated behind
*configuration* surfaces that were never built. An operator cannot switch the OMS on without `curl`.

This was found by auditing the live demo (`https://app.demo.openlinker.io`) against the tree at
`0906ceb41`. The evidence, all verified rather than inferred:

### Four shipped features have zero UI

| Feature | Backend | Frontend |
|---|---|---|
| **Sourcing-rules authoring** | 6 admin endpoints at `connections/:connectionId/sourcing-rules` — full CRUD + exhaustive `PUT /order` reorder (`apps/api/src/oms/http/oms-sourcing-rules.controller.ts`, #2953) | **Nothing.** The string `sourcing-rules` appears **zero times** in `apps/web/src`. `apps/api/src/oms/http/` contains exactly one controller, so the entire OMS HTTP module has no client |
| **Inventory locations CRUD** | `GET`/`POST` `/inventory/locations`, `GET`/`PATCH`/`DELETE` `/inventory/locations/:id` (#2316) | Only `POST /inventory/locations/bootstrap` (`features/inventory/hooks/use-bootstrap-locations-mutation.ts`) and a degenerate `?status=active&limit=1` probe that reads `total` for `features/connections/components/router-readiness-panel.tsx`. No list, create, edit or delete surface |
| **Duplicate-position detection** | `GET /inventory/duplicate-positions` (`apps/api/src/inventory/http/inventory.controller.ts`, admin — the documented gate procedure for #2325) | **Nothing.** `duplicate-position` appears zero times in `apps/web/src` |
| **Returns authorize / match-order** | `POST /returns/:returnId/authorize`, `POST /returns/:returnId/match-order` (`apps/api/src/returns/http/return-writes.controller.ts`, #2372) | **Nothing** — while every other returns write is wired |

The sourcing-rules gap is the load-bearing one. Rules are what `evaluateRouting` consumes; with an empty
ruleset every line resolves unfulfillable and `RoutingCommitService` refuses the plan as
`plan-carries-unfulfillable`. So even after an operator creates the OMS connection and bootstraps a
[...]

Issue #2954, opened 2026-09-06
[TASK] OMS routing — no dry-run surface; `FulfillmentRouterPort.evaluate()` has no caller


piotrswierzy:
## Problem

`FulfillmentRouterPort` carries two methods, and only one of them is reachable.

`evaluate()` is the **non-committing** half (#2393, DESIGN §5.3). `OlFulfillmentRouter` implements it, and its docblock is explicit that this is the point:

> `evaluate()` and `route()` are the same computation. Both call `evaluateRouting` and nothing else. `route()` adds only a decision reference; it never re-decides. That is what makes their explanations provably identical rather than merely similar, and it is why `evaluate()` cannot leak a committing identifier — it has none to leak.

**Nothing calls it.** There is no service, no controller and no job that reaches `evaluate()`; the only reference outside the class is its own spec. The property the design paid for — a dry run that provably matches what would really happen — is unobservable.

## What this blocks

Wave 3a's exit criterion (#2412): *"an operator with two locations enables the OL-OMS connection, authors an ordered filter/sort list, **sees a dry-run explanation for an incoming order**, and works the resulting tasks to shipped."*

It also removes the only safe way to check a ruleset before arming it. Routing decides where physical stock ships from; an operator who cannot preview a ruleset must test it on a real order, and the first evidence of a wrong rule is a parcel leaving the wrong building.

## Scope

- An endpoint that, given an order and an OMS connection, returns `RoutingEvaluation` — `candidates`, `unfulfillable` and `explanation`.
- **It must call `evaluate()`, never `route()`.** The contract-level guarantee is that no committing identifier travels in either direction: `RoutingEvaluation` carries no `decisionId`, and `evaluate` takes no options, so it cannot be handed an idempotency key. An endpoint that called `route()` and discarded the result would commit a decision and mint an idempotency key the vendor would dedup a later real route against.
- `explanation` uses opaque rule names with display labels (REVIEW H7) — render the labels the router supplies rather than restating the vocabulary in the frontend.
- Read-authorised, since the whole point is that an operator can inspect before they can arm.

## Out of scope

- The rules-authoring API (filed separately).
- Consuming the `pending` `RoutingPlan` arm — Wave 3a refuses it (`PendingRoutingPlanNotSupportedError`); that is `W4-3`.

## References

- `libs/core/src/fulfillment/domain/ports/fulfillment-router.port.ts`
[...]

Issue #2869, opened 2026-09-03
[TASK] Fulfillment — let an operator route a parcel to the in-house bench by hand


piotrswierzy:
## Problem

An order reaches a pack bench **only** if routing created a `FulfillmentWork` and dispatched it to the in-house holder. There is no way for a person to put a parcel on a bench.

That is correct as a default — see [ADR-071](docs/architecture/adrs/071-pack-station-principal.md) and `docs/specs/product-spec-oms-wave3b-scan-pick-pack.md` § 1.3, where the bench list is deliberately **authoritative** (what routing assigned) rather than **derived** (what looks packable). But it leaves three real cases with no answer:

1. **Routing is switched off.** No routing → no works → the bench is permanently empty. An operation that wants scanner-verified packing without adopting OMS routing cannot have it. The spec already concedes this: *"the pack bench is not a standalone feature."* That is a genuine adoption barrier.
2. **The order routed elsewhere**, and a human wants it packed here anyway.
3. **The order failed to route** — unfulfillable, held at the time, or the router declined — and someone needs to get it out of the door regardless.

## What this is NOT

**Not a second list.** A hand-routed parcel must appear in the *same* bench list, sorted the same way, obeying the *same* eligibility gate. Two sources of truth for "what is on this bench" would recreate exactly the derived-list problem the authoritative design exists to avoid — with the added twist that both would claim to be authoritative.

**The principle: manual control adds a PRODUCER of work, never a second list.** The bench does not change. It still shows work assigned to its holder, accepted and not closed. A person becomes another way that condition comes true.

## The real design question

**#2395 commits exactly one routing decision per order**, under a live partial-unique index, in one transaction. So a manual route must *take the decision slot*, not sit beside the router's. That is the substance of this issue and it belongs with the routing context, not with a pack-bench spec:

- Can a human override an **existing** decision, or only decide where none exists?
- If the work was already dispatched to a 3PL holder and accepted, what happens to it — cancelled, or refused?
- Does a manual decision record its actor, and is it distinguishable from a router decision in `routing_decisions`? (It should be: "a person chose this" and "a rule chose this" are different facts, and the explanation surface will want to say which.)
- Does the exactly-once gate need a new outcome, or does a manual route reuse an existing one?
- What happens on re-ingestion of an order whose routing a human overrode — does the router reclaim it?

## Scope

- A route + action for "route this order to the in-house bench", admin/operator-gated.
- A `routing_decisions` row attributable to a person.
[...]

PR #3029, opened 2026-09-09
docs(3017): OMS configuration-surface route audit (step 1)

norbert-kulus-blockydevs:
## Summary

Completes step 1 of #3017's Proposed Solution: a full route-by-route audit of `apps/api/src/{oms,bench,fulfillment,fulfillment-authority,inventory,operational-settings,returns}` against `apps/web/src` callers, plus re-verification of every claim #3017 made against current `main`.

- 46 routes across 12 controllers catalogued (verb, `@Roles`/`@Public`/`@AnyRole`, handler, exact `apps/web/src` caller or **NOT CALLED**).
- Confirms sourcing-rules CRUD (6/6), inventory-locations CRUD beyond bootstrap+count (4/6), duplicate-positions, and four returns-write routes (`authorize`, `match-order`, `record`, `correction-proposal` confirm) remain unreachable from the UI.
- Corrects one place where #3017's own table, while accurately worded, didn't enumerate every route in a controller it named.
- Confirms all three reachability defects (`/bench` nav, missing OMS `setupCard`, dead-end who-decides remediation copy) unchanged, and surfaces a concrete new blocker: a bench nav entry needs `RoleValues` widened to admit `packer`.
- Checks both tracker facts #3017 rests on live via `gh api` (#2954 open, #2869 closed with its manual-route-producer remainder still uncovered by any open issue).
- **Cross-verified by two independent subagents** (one blind re-derivation from scratch, one adversarial line-by-line check) — findings and corrections documented in § 6.3. Net: one real content correction (a dead API-client export with no caller), one citation-precision fix, one minor wording fix. Nothing else flagged.

No production code. The product spec (`docs/specs/product-spec-oms-configuration-surface.md`), mockups, and per-surface implementation issues — #3017 steps 2–4 — are explicitly deferred to a later session per the requester.

## Test plan

- [x] N/A — documentation-only change, no code paths touched.
- [x] Every factual claim in the audit is grounded in a reproducible grep/read/`gh api` command shown inline.
- [x] Independently cross-verified by two separate agent passes; corrections folded in.

Refs #3017 (not closing — this is step 1 of 4).


norbert-kulus-blockydevs:
## Review — OMS configuration-surface route audit (step 1)

**Verdict: ✅ Approve.**

Reviewed as an architectural audit document rather than a code change, per this repo's `/pr-review` conventions for docs-only PRs.

### What I independently re-verified

I re-ran a sample of the audit's load-bearing claims against current `main` rather than trusting the doc's own (already double-cross-verified) numbers, focusing on the claims a future reader is most likely to lean on:

- `apps/web/src/app/nav-registry.types.ts:26` — `RoleValues = ['admin', 'operator']`, no `packer` member. **Matches** § 6.2.B.1's claim exactly.
- All three bench controllers' route decorators — every one carries `@Roles('admin', 'operator', 'packer')`. **Matches** § 6.1 rows 7–13 exactly, and confirms the nav-blocker finding is real (backend already admits `packer`; FE role vocabulary does not).
- `oms-sourcing-rules.controller.ts` — exactly 6 routes (`GET`, `POST`, `PUT order`, `GET :ruleId`, `PATCH :ruleId`, `DELETE :ruleId`) under class-level `@Roles('admin')`. **Matches** § 6.1 rows 1–6.
- `manual-route-admissibility.types.ts` is referenced by exactly 3 files (barrel, itself, its own spec) — zero service call sites. **Matches** the corrected "three files, not two" claim in § 6.3, and the docblock's own verbatim self-report ("Nothing calls this today") checks out.
- `grep -rn "sourcing-rules"` and `"duplicate-position"` across `apps/web/src` — zero hits (excluding the unrelated `who-decides` copy string, which the doc itself correctly disambiguates as a different concept). **Matches** §§ 6.2.A.1/A.3.
- `fulfillment.api.ts`'s `get(workId)` export has zero callers in `apps/web/src` — only `.list`, `.listByOrder`, `.applyAction` are used. **Matches** the row-18 correction the doc attributes to its own cross-verification pass.
- `inventory-locations.controller.ts`'s role decorators (reads: `admin, operator, viewer`; writes: `admin`) and `inventory.api.ts`'s actual usage (only the degenerate probe + `bootstrap`) both **match** §§ 6.1 rows 23–28 and 6.2.A.2 precisely.

Every claim I spot-checked held up exactly, with no rounding or hand-waving. That's a meaningfully higher bar than most audit docs clear.

### On the reachable-vs-dormant distinction (my specific brief for this review)

The audit is scoped narrowly and *honestly* to "does `apps/web/src` call this route" — it does not claim or imply that a `NOT CALLED` route would work end-to-end if a caller existed (e.g. whether the OMS connection needs to be enabled first per ADR-055's never-seeded rule, or whether a capability needs enabling on that connection per § 26 Fulfillment). That's the correct scope for a route-audit step 1, and the doc doesn't overstate it into "these features are live" anywhere I could find — it consistently says "wired" / "NOT CALLED" about FE↔BE plumbing only, never about production readiness. Good discipline; worth being equally explicit about this scope boundary when the product spec (step 2) picks this up, since that's where "reachable in the UI" and "functionally live given OMS enablement" will need to be told apart for real.

### Process note (not a finding, a compliment)

The two-independent-subagent cross-verification in § 6.3 is a good pattern for an audit whose entire value proposition is "trustworthy enough to build a product spec on top of." It caught two real self-inflicted regressions (a dead-code miscount, a line-number off-by-one from a mid-session edit) and documented the correction transparently rather than silently fixing and moving on — exactly what you'd want from a document meant to be the source of truth for later work.

### Minor, non-blocking

- § 4's docblock-imprecision flag about the stale `#2407` reference in `apps/web/src/plugins/oms/index.ts`'s comment (implying a guided-flow issue exists when it doesn't) is a good catch — but it's buried mid-paragraph in § 6.2.B.2. Worth a one-line callout box or bold lead-in so the next code-touching PR that fixes it doesn't have to re-read the whole finding to find the actionable bit.
- No test plan checkboxes are meaningfully exercisable for a docs-only PR, and the doc correctly marks them N/A rather than padding with irrelevant checks.

Nothing here blocks merge.

---
🤖 Reviewed via `/pr-review`

norbert-kulus-blockydevs:
Addressed the one non-blocking finding from the review:

> § 4's docblock-imprecision flag about the stale `#2407` reference in `apps/web/src/plugins/oms/index.ts`'s comment ... is buried mid-paragraph in § 6.2.B.2. Worth a one-line callout box or bold lead-in.

Fixed in b42e53f98: split the finding into its own bolded sub-bullet directly under the point it corrects, so a future code-touching PR sees the actionable bit without re-reading the whole paragraph. No change to the finding's substance.

piotrswierzy (approved):
## `/pr-review` — ✅ Approve

**This is the rare audit document whose rows survive independent re-derivation.** I re-checked substantially more than a spot-sample, across all seven directories, and found no row that is wrong or stale.

**Ref discipline is explicit and current.** The doc states `42cc86dde` in § 1, § 6.2, § 6.3 and § 8. `origin/main` **is** `42cc86dde` right now, so the audit is not stale — and it correctly distinguishes #3017's own snapshot (`0906ceb41`) from its own, which directly answers the question a reader would have.

**Structural totals — exact.** 12 controllers under the 7 directories; per-file route-decorator counts sum to **46** (3+3+6+1+3+6+3+3+1+2+10+5), matching § 6.1 and its totals line (29 wired + 1 partial + 16 not called = 46).

**Role decorators — spot-checked all six largest controllers, zero discrepancies**, including the two non-obvious ones the doc calls out: `oms-sourcing-rules.controller.ts` carries class-level `@Roles('admin')` at line 72 with no per-method decorators, and `inventory.controller.ts` mixes `@AnyRole()` on the two reads with `@Roles('admin')` on `duplicate-positions`.

**NOT-CALLED verdicts — independently re-grepped, all confirmed:**

| Claim | Verified |
|---|---|
| `sourcing-rules` zero hits in `apps/web/src` | **0** ✓ |
| `duplicate-position` zero hits | **0** ✓ |
| `returns/:id/authorize`, `match-order` no caller | absent from `returns.api.ts` ✓ |
| `POST /returns/record` no caller | absent ✓ |
| `POST .../correction-proposal` no caller | only the `GET` at `returns.api.ts:367` ✓ |
| `/bench` absent from `nav-registry.ts` | **0** ✓ |
| `plugins/oms/index.ts` no `setupCard` | one hit — the docblock saying there isn't one ✓ |
| `inventory/locations` = bootstrap + `?limit=1` probe only | `inventory.api.ts:66` is literally `'/inventory/locations?status=active&limit=1'`; `:69` is bootstrap ✓ |

**CALLED citations — every cited line number is exact.** I checked all 17 across the six API-client files (`bench-work.api.ts` 96/106/110/119/126/129/132, `who-decides.api.ts` 67/71/79, `fulfillment.api.ts` 68/73/82/90, `inventory.api.ts` 59/63/66/69, `operational-settings.api.ts` 27/30) — each lands on the actual `request<…>(…)` call. The § 6.3 off-by-one flag was correctly resolved in favour of these values.

**Narrative claims hold too.** `git diff --name-only 0906ceb41..42cc86dde` over the seven directories returns **0 files**, exactly as § 4 claims. `RoleValues = ['admin','operator']` at `nav-registry.types.ts:26` (no `packer`) ✓. `platform-picker.tsx` filters on `p.setupCard !== undefined` ✓. `bench-work.copy.ts:102` remedy copy verbatim as quoted ✓. `authority-presets.ts` — **no preset enables packing** ✓.

**Three findings #3017 did not have**, all correct:
- two extra uncalled `return-writes` routes,
- the `RoleValues`-must-grow-`packer` prerequisite for a bench nav item,
- and **row 18** — `fulfillment.api.ts:82` exports `get(workId)` and grepping every `apiClient.fulfillment.` reference outside tests returns only `.listByOrder`, `.list`, `.applyAction`. No caller. A genuine gap, and the doc is right that it's a *different kind* of gap (a dead client export, not a missing surface).

The `Partial` verdict on row 23 is the right call rather than a fudge — it distinguishes route-level from family-level coverage. AC-1 of #3017 is met.

### SUGGESTIONS (editorial only)

- § 6.3 narrates the two-agent verification pass including an off-by-one introduced mid-session and reverted. Admirable transparency, but it is *process*, not deliverable — and step 2's author will read § 6.1/§ 6.2 as source of truth. Compress to the three net corrections and move the narrative to the PR body.
- This is a completed read-only verdict, not a plan; the repo's home for that shape is `docs/plans/analysis/` (the `/pre-implement` convention). Filing it as `implementation-plan-*` means the next person grepping `docs/plans/analysis/` for prior audits won't find it.

---

Issue #2328, opened 2026-08-22
[TASK] CORE — returns repository ports and ReturnsService (idempotent update-or-create)


piotrswierzy:
> Part of the OMS programme backlog (**Wave 1c**, stream S2, size M). Backlog slug: `W1c-2`. Sibling `W*` slugs resolve via the Wave-1c epic's checklist. Design of record: `docs/plans/analysis/DESIGN-oms-authority-model.md` + ADRs 052–062 (esp. ADR-060).

## Problem / Context

Ingestion must be idempotent and must never clobber OL-owned fields — the same discipline `OrderIngestionService.syncOrderFromSource` follows, and the same rule #2282 is closing for orders.

## Proposed Solution

`ReturnRepositoryPort` (minimal method set: `findBySourceKey`, `upsertFromSource`, `listOrphans`, `save`) + repository, and `IReturnsService` / `ReturnsService`.

The ingestion write is **update-or-create keyed `(sourceConnectionId, externalReturnId)`**, with re-pull **authoritative on source-owned fields** and **OL-owned fields never touched by ingestion** (`authorizedAt`, `declinedAt`, `closedAt`, custody/money columns, `quantityReceived` and downstream counters). Exclude the OL-owned columns from `toOrm` in the repository, following the `cancelledAt` single-writer precedent, so a stale in-memory read cannot stomp them.

## Classification

**Type**: CORE
**Layer**: Domain (port) + Application + Infrastructure
**File(s)**: `libs/core/src/returns/domain/ports/`, `.../application/services/`, `.../infrastructure/persistence/repositories/`; pattern reference `libs/core/src/orders/application/services/order-ingestion.service.ts`

## Dependencies

- #2327 (`W1c-1`)

## Assumptions

- No per-return lock is needed at Wave 1c: a single feed consumer per connection plus the unique key makes concurrent writes for the same return improbable, and the write is an idempotent upsert. If Wave 2 adds an operator write path racing ingestion, a lock is filed then.

## Acceptance Criteria

- [ ] Re-ingesting the same return twice produces one row and refreshes only source-owned fields
- [ ] OL-owned columns are excluded from `toOrm` and asserted so by a spec (the #2100 precedent)
- [ ] Repository throws domain errors, never TypeORM errors
- [ ] `ReturnsService` implements a declared interface and is bound by a Symbol token
- [ ] Tests added or updated for non-trivial logic
- [ ] No architecture boundary violations (CORE ↔ Integration)

Issue #2284, opened 2026-08-21
[BUG] CORE/Orders — a source-cancelled order can still be provisioned into a destination shop (`WHERE cancelledAt IS NULL` predicate)


piotrswierzy:
## Problem / Context

`order_records.cancelledAt` is a persisted, first-write-wins fact — but the destination-provisioning path does not consult it: `grep cancelledAt` over `order-sync.service.ts` / `order-ingestion.service.ts` returns nothing. `OrderSyncService.syncOrder` resolves every `OrderProcessorManager` destination and creates unconditionally, so an order whose cancellation was recorded (e.g. by a poll that landed between enqueue and the `marketplace.order.sync` job run, or on a retry) can still be created in PrestaShop/WooCommerce and shipped by the shop's own operator. ANALYSIS-1032 §0 item 1 identifies this as "the one real bug the whole programme was organised around"; the OMS design (PR #2281, DESIGN-oms-authority-model §6.4) makes the predicate a Wave-0 precondition that ships regardless of any OMS work.

## Proposed Solution

Gate destination provisioning on the persisted fact: before `createOrderIdempotently` runs for a destination, re-read the order's `cancelledAt` (or thread it from the freshly persisted record) and skip creation when set — recording a per-destination `syncStatus` outcome that says *why* (not a silent skip, and not a `failed` that would retry). A cancellation observed **after** a destination order already exists is out of scope (that is the ADR-044 cancel-relay flow); this issue closes only the provision-after-cancel window.

## Classification

**Type**: CORE
**Layer**: Application
**File(s)**: `libs/core/src/orders/application/services/order-sync.service.ts` (the per-destination create path), `libs/core/src/orders/application/services/order-ingestion.service.ts` (call-site threading)

## Dependencies

- None (Wave 0).

## Assumptions

- The skip is recorded as a distinct per-destination sync outcome (e.g. `skipped_cancelled`) surfaced in the existing syncStatus rendering, rather than a new order-level state.

## Acceptance Criteria

- [ ] An order with `cancelledAt` set is never created in any destination, including on job retry.
- [ ] The skip is visible per destination in `syncStatus` with a cancellation-specific reason, and is terminal (no retry loop).
- [ ] Orders without `cancelledAt` provision byte-identically to today.
- [ ] Integration test: cancel recorded between ingestion and the sync job run → no destination order exists afterwards.
- [ ] No architecture boundary violations (CORE ↔ Integration).

Issue #1984, opened 2026-08-03
[IMPL] Backend — capture order cancellation as first-class record state


piotrswierzy:
## Design source

**Product Design parent:** Part of #1976 — see spec at `docs/specs/product-spec-1976-analytics.md` (§ 1a, dependency **[C]**)

## Scope

Cancellation is currently not represented on an order record. `recordStatus` is only `ready | awaiting_mapping | source_deleted`, and a source cancellation event writes **nothing** to the record — `handleSourceCancellation` (`order-ingestion.service.ts:485-536`) relays the cancellation to destinations and returns. Cancelled state therefore survives only as a string inside `orderSnapshot.status`, and only if a later poll happens to re-ingest the order.

Consequence: any revenue sum silently counts cancelled orders, and cancellation rate (a metric operators want in its own right) cannot be reported at all.

Make cancellation an explicit, queryable fact on the order record, written at the moment the source reports it.

## Acceptance criteria

- [ ] A source-reported cancellation is durably recorded on the order record when the cancellation event is handled, not only if a later poll re-ingests it
- [ ] Cancelled orders are queryable without reading JSON — sufficient to exclude them from an aggregate or count them on their own
- [ ] The cancellation time is recorded, not just the fact (an order cancelled today should not silently alter last month's reported figures)
- [ ] Re-ingesting an already-cancelled order does not duplicate or flip the state
- [ ] Existing cancellation relay behaviour to destinations is unchanged
- [ ] Existing records are handled sensibly — the backfill or default position for historical rows is explicit and documented
- [ ] Tests added (unit + integration where applicable)
- [ ] No new ESLint warnings or type errors introduced

## Effort estimate

- [x] **M** — 3–7 days

## Dependencies

- **Independent** — can start immediately
- **Blocks:** the order analytics read model, and every money metric downstream of it

## Out of scope for THIS issue

- **Refunds.** `refunded` exists as an unwritten enum member; no refund amount is stored anywhere. Representing partial refunds is a separate data-capture feature (recorded as a known gap in the spec).
- Returns / withdrawals — no return entity exists
- Restock behaviour on cancellation (already handled by the existing `marketplace.offer.stockRestore` path)

PR #3035, opened 2026-09-09
fix(orders): persist a durable early-cancellation signal (#2069)

### LINKED ISSUE #2069: [BUG] Orders — an early cancel persists nothing, so #2284's `cancelledAt` guard has nothing to find
## Problem / Context

A source cancellation that arrives **before** the order's create/sync job has run is silently lost, and the later create then provisions the order at the destination as **active**. The operator sees a live, shippable order that the buyer cancelled.

The code states the race itself, in `libs/core/src/orders/application/services/order-ingestion.service.ts` (in `handleSourceCancellation`'s caller):

```
// Known residual (#1160): a cancel that arrives *before* the order's
// create/sync job has run finds no targets here, and the later create then
// provisions the order as active. Fully closing that out-of-order race needs
// the deferred monotonic / relay-log machinery (ADR-027 guardrails) tracked
// with the bidirectional slices — out of scope for this unidirectional slice.
```

`handleSourceCancellation` returns `[]` when `getInternalId` yields nothing — there is no order yet, so there are no relay targets and nothing is recorded.

**The "deferred machinery" turns out not to be needed.** This was re-examined while planning #1032: a durable per-target relay-obligation table does **not** fix this case, because obligation rows are keyed per *resolved target* and this cancel has zero targets. A sweep re-drives writes that were attempted and failed; this one was never attemptable.

What does fix it already exists in two halves:

- `OrderRecordRepository.markCancelled` stamps `cancelledAt` durably, first-write-wins (#1984).
- `order-record.orm-entity.ts:92` names the missing half in a comment: *"Indexed for the future exclusion predicate (#1987/#1988: `WHERE "cancelledAt" IS NULL`)"*.

`cancelledAt` is indexed and is **not** read as a guard by any non-shipping code today.

## Proposed Solution

Have the provisioning path consult `cancelledAt` before creating the order at a destination, and skip provisioning when it is set.

- `libs/core/src/orders/application/services/order-ingestion.service.ts` — the create/sync path that reaches `OrderSyncService`
- `libs/core/src/orders/infrastructure/persistence/repositories/order-record.repository.ts` — `markCancelled` (already first-write-wins; no change expected)
- `libs/core/src/orders/infrastructure/persistence/entities/order-record.orm-entity.ts:88-96` — update the comment once the predicate is no longer "future"

The ordering that makes this work: a cancel for an unknown order must still persist a record carrying `cancelledAt` (keyed on the source connection + external order id), so the later create finds it. Confirm whether the current cancel path does that or exits before any write — if it exits, that write is part of this fix.

## Classification

**Type**: CORE
**Layer**: Application
**File(s)**: `libs/core/src/orders/application/services/order-ingestion.service.ts`, `libs/core/src/orders/infrastructure/persistence/repositories/order-record.repository.ts`

## Dependencies

None. This deliberately does **not** depend on #861 or on any lifecycle-ledger work — that was the finding.

## Assumptions

- The destination create is the only provisioning path that needs the guard. If a second path exists (retry service, manual re-dispatch), it needs the same predicate.

## Acceptance Criteria

- [ ] A cancel for an order OL has not yet ingested persists a durable record such that a later create can observe it
- [ ] The provisioning path skips destination creation when `cancelledAt IS NOT NULL`
- [ ] Integration test: cancel arrives first, then the sync job runs — assert no destination order is created
- [ ] Integration test: normal ordering (create then cancel) is unchanged
- [ ] The `#1987/#1988: future exclusion predicate` comment is updated to reflect the shipped state
- [ ] The `#1160` known-residual comment in `order-ingestion.service.ts` is removed or amended

jakubretajczykBD:
## Summary
- A source cancellation arriving before OL had ever ingested the order was silently dropped: `handleSourceCancellation` found no internal mapping and returned, so the later create/sync job provisioned the order as active at every destination.
- Adds a durable `order_cancellation_signals` table keyed on `(sourceConnectionId, externalOrderId)` — the one identity that exists before ingestion — rather than minting a phantom internal id (the #2328 lesson for returns attribution).
- `handleSourceCancellation` records the signal when the order is unknown; `OrderRecordService.persistIncomingSnapshot` consumes it atomically (`DELETE ... RETURNING`) the moment the order is genuinely ingested, feeding it through the existing first-write-wins `markCancelled` pipeline (#1984) — so `OrderSyncService`'s already-shipped `cancelledAt IS NULL` provisioning guard (#2284) sees it and skips destination creation.
- Updates the stale `#1160` residual comment and the `#1987/#1988` "future predicate" comment on the ORM entity to reflect the shipped state.

## Acceptance criteria (from #2069)
- [x] A cancel for a not-yet-ingested order persists something the later create can observe
- [x] Integration test: cancel first, then sync — assert no destination order is created
- [x] Integration test: normal ordering unchanged
- [x] `#1987/#1988` future-predicate comment updated
- [x] `#1160` residual comment amended (explaining why the relay-log machinery was never needed)

## Test plan
- [x] `pnpm type-check` — full workspace, green
- [x] `pnpm lint` — full workspace, green (0 errors; pre-existing warnings only)
- [x] New unit tests: `order-cancellation-signal.repository.spec.ts`, extended `order-ingestion.service.spec.ts` / `order-record.service.spec.ts`
- [x] New integration test: `order-early-cancellation-signal.int-spec.ts` (cancel-before-create + normal-ordering regression)
- [ ] `pnpm test` / `pnpm test:integration` — not run in this session (Docker-gated); please run in CI before merge

Closes #2069

### DIFF
--- a/apps/api/src/migrations/1878000000000-create-order-cancellation-signals.ts
+++ b/apps/api/src/migrations/1878000000000-create-order-cancellation-signals.ts
@@ -1,0 +1,72 @@
+/**
+ * Create Order Cancellation Signals Migration (#2069)
+ *
+ * Creates `order_cancellation_signals` — a durable trace that a source
+ * cancellation arrived for an order OpenLinker has not yet ingested.
+ *
+ * Before this, `OrderIngestionService.handleSourceCancellation` had nowhere
+ * to write when `IIdentifierMappingService.getInternalId` resolved no
+ * mapping: `OrderRecordRepository.markCancelled` is a bare
+ * `UPDATE ... WHERE "internalOrderId" = $2` with no insert fallback, so a
+ * cancel racing ahead of the order's own create/sync job was silently
+ * dropped and the later create provisioned the order as active at every
+ * destination. Minting an internal id via `getOrCreateInternalId` from the
+ * cancellation path to give `markCancelled` something to hit was rejected —
+ * it would create a phantom, operator-visible order with no real data yet,
+ * exactly the shape #2328 rejected for returns attribution.
+ *
+ * The signal is keyed on `(sourceConnectionId, externalOrderId)` instead —
+ * the one durable identity that exists before ingestion — and is consumed
+ * (atomically, `DELETE ... RETURNING`) by
+ * `OrderRecordService.persistIncomingSnapshot` the moment the order is
+ * genuinely first ingested, applying it through the existing first-write-wins
+ * `markCancelled` pipeline (#1984) that `OrderSyncService`'s `#2284`
+ * `cancelledAt IS NULL` provisioning guard already reads.
+ *
+ * There is deliberately **no FK** to `order_records` — the `order_holds` /
+ * `order_changes` / `refund_records` precedent of an indexed reference by
+ * value; here there is nothing yet to reference at all. Nothing cascades
+ * into this table, so `apps/api/test/integration/setup.ts` truncates it
+ * explicitly.
+ *
+ * The unique index name matches `OrderCancellationSignalOrmEntity`'s
+ * class-level `@Index` declaration exactly, because the integration harness
+ * builds its schema by `synchronize` rather than by migration (the
+ * `order_holds` precedent) — an unnamed index there would hash-name
+ * differently and the two schemas would diverge on the very constraint the
+ * repository's first-write-wins guarantee (`ON CONFLICT DO NOTHING`) depends
+ * on.
+ *
+ * @module apps/api/src/migrations
+ */
+import type { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class CreateOrderCancellationSignals1878000000000 implements MigrationInterface {
+  name = 'CreateOrderCancellationSignals1878000000000';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    // `id` defaults to uuid_generate_v4() — the same guard 1846/1847/1850 use.
+    await queryRunner.query(`CREATE EXTENSION IF NOT EXISTS "uuid-ossp"`);
+
+    await queryRunner.query(`
+      CREATE TABLE IF NOT EXISTS "order_cancellation_signals" (
+        "id" uuid NOT NULL DEFAULT uuid_generate_v4(),
+        "sourceConnectionId" uuid NOT NULL,
+        "externalOrderId" text NOT NULL,
+        "cancelledAt" TIMESTAMP WITH TIME ZONE NOT NULL,
+        "createdAt" TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT now(),
+        CONSTRAINT "PK_order_cancellation_signals" PRIMARY KEY ("id")
+      )
+    `);
+
+    await queryRunner.query(`
+      CREATE UNIQUE INDEX IF NOT EXISTS "UQ_order_cancellation_signals_source_external"
+        ON "order_cancellation_signals" ("sourceConnectionId", "externalOrderId")
+    `);
+  }
+
+  public async down(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`DROP INDEX IF EXISTS "UQ_order_cancellation_signals_source_external"`);
+    await queryRunner.query(`DROP TABLE IF EXISTS "order_cancellation_signals"`);
+  }
+}
--- a/apps/api/src/migrations/1879000000000-create-order-cancellation-signals.ts
+++ b/apps/api/src/migrations/1879000000000-create-order-cancellation-signals.ts
@@ -41,8 +41,8 @@
  */
 import type { MigrationInterface, QueryRunner } from 'typeorm';
 
-export class CreateOrderCancellationSignals1878000000000 implements MigrationInterface {
-  name = 'CreateOrderCancellationSignals1878000000000';
+export class CreateOrderCancellationSignals1879000000000 implements MigrationInterface {
+  name = 'CreateOrderCancellationSignals1879000000000';
 
   public async up(queryRunner: QueryRunner): Promise<void> {
     // `id` defaults to uuid_generate_v4() — the same guard 1846/1847/1850 use.
--- a/apps/api/src/orders/http/refunds.controller.spec.ts
+++ b/apps/api/src/orders/http/refunds.controller.spec.ts
@@ -44,6 +44,7 @@ describe('RefundsController', () => {
       findByIds: jest.fn(),
       updateFulfillmentState: jest.fn(),
       markCancelled: jest.fn(),
+      recordEarlyCancellationSignal: jest.fn(),
       markSalesDocumentBlock: jest.fn(),
       markFulfillmentBlock: jest.fn(),
       getEarliestOrderDateByConnection: jest.fn(),
--- a/apps/api/src/shipping/http/shipment.controller.spec.ts
+++ b/apps/api/src/shipping/http/shipment.controller.spec.ts
@@ -137,6 +137,7 @@ describe('ShipmentController', () => {
       markItemResolutionFailure: jest.fn(),
       getFailedSyncValueSummary: jest.fn(),
       markCancelled: jest.fn(),
+      recordEarlyCancellationSignal: jest.fn(),
       markSalesDocumentBlock: jest.fn(),
       markFulfillmentBlock: jest.fn(),
       markPacked: jest.fn(),
--- a/apps/api/test/integration/orders/order-early-cancellation-signal.int-spec.ts
+++ b/apps/api/test/integration/orders/order-early-cancellation-signal.int-spec.ts
@@ -234,6 +234,33 @@ describe('Order early-cancellation signal (#2069)', () => {
       where: { sourceConnectionId: sourceConnection.id, externalOrderId },
     });
     expect(signalAfterSync).toBeNull();
+
+    // Act 3: a THIRD poll, after the signal has already been consumed and the
+    // source's order resource is still lagging (`incoming.status: 'pending'`
+    // from `makeIncomingOrder`, unchanged). `persistIncomingSnapshot` no longer
+    // has a signal to consume, so `snapshotRecord.cancelledAt` alone would read
+    // `null` here (`upsert()`'s `fromRawRow` excludes `cancelledAt` from its
+    // write set on this non-writing path) — this is the exact regression the
+    // #2069 re-review found: without the `existing.cancelledAt` fallback in
+    // `OrderIngestionService`, this poll would silently re-provision an order
+    // OL already recorded as cancelled.
+    const thirdPoll = await ingestion.syncOrderFromSource(
+      sourceConnection.id,
+      externalOrderId,
+      'create-evt-2'
+    );
+
+    expect(thirdPoll).toHaveLength(1);
+    expect(thirdPoll[0].status).toBe('skipped_cancelled');
+    expect(createOrderCalls).toHaveLength(0);
+
+    const recordAfterThirdPoll = await recordRepo.findOne({
+      where: { internalOrderId: mappingAfterSync!.internalId },
+    });
+    expect(recordAfterThirdPoll!.cancelledAt).not.toBeNull();
+    expect(recordAfterThirdPoll!.cancelledAt!.getTime()).toBe(
+      signalAfterCancel!.cancelledAt.getTime()
+    );
   });
 
   // #2069 review — the exact race the signal exists for: the source's order
@@ -322,6 +349,13 @@ describe('Order early-cancellation signal (#2069)', () => {
       where: { sourceConnectionId: sourceConnection.id, externalOrderId },
     });
     expect(signalAfterFirst).not.toBeNull();
+
+    // Force the two calls' `new Date()` stamps onto DISTINCT milliseconds —
+    // without this, two `ingestion.syncOrderFromSource` calls back-to-back can
+    // land in the same millisecond, and the equality assertion below would
+    // pass trivially (by timestamp coincidence) even against a `DO UPDATE`
+    // that overwrote the row with the second cancel's instant.
+    await new Promise((resolve) => setTimeout(resolve, 5));
 
     // A second, later-arriving cancel event for the SAME
     // (sourceConnectionId, externalOrderId) — e.g. a duplicate delivery or a
--- a/apps/api/test/integration/orders/order-early-cancellation-signal.int-spec.ts
+++ b/apps/api/test/integration/orders/order-early-cancellation-signal.int-spec.ts
@@ -169,12 +169,14 @@ describe('Order early-cancellation signal (#2069)', () => {
 
     // Act 1: the cancel event arrives before the order has ever been
     // ingested — no identifier mapping exists yet.
+    const beforeCancel = new Date();
     const cancelResult = await ingestion.syncOrderFromSource(
       sourceConnection.id,
       externalOrderId,
       'cancel-evt-1',
       'cancelled'
     );
+    const afterCancel = new Date();
     expect(cancelResult).toEqual([]);
 
     const mappingAfterCancel = await mappingRepo.findOne({
@@ -190,6 +192,13 @@ describe('Order early-cancellation signal (#2069)', () => {
       where: { sourceConnectionId: sourceConnection.id, externalOrderId },
     });
     expect(signalAfterCancel).not.toBeNull();
+    // The signal carries the instant the cancel-handling branch stamped
+    // (`new Date()` at the time of the call), not a sentinel — pin the value
+    // rather than only its non-nullness.
+    expect(signalAfterCancel!.cancelledAt.getTime()).toBeGreaterThanOrEqual(
+      beforeCancel.getTime()
+    );
+    expect(signalAfterCancel!.cancelledAt.getTime()).toBeLessThanOrEqual(afterCancel.getTime());
 
     // Act 2: the create/sync job now runs — the ordinary path, no eventType.
     const syncResult = await ingestion.syncOrderFromSource(
@@ -214,13 +223,30 @@ describe('Order early-cancellation signal (#2069)', () => {
     const record = await recordRepo.findOne({
       where: { internalOrderId: mappingAfterSync!.internalId },
     });
+    // Pin the VALUE, not just non-nullness: `cancelledAt` must be the
+    // signal's own consumed instant (`earlySignalAt`), never `now` at
+    // ingestion time — proving `consume()`'s returned timestamp actually
+    // reached `markCancelled` rather than merely triggering it.
     expect(record!.cancelledAt).not.toBeNull();
+    expect(record!.cancelledAt.getTime()).toBe(signalAfterCancel!.cancelledAt.getTime());
 
     const signalAfterSync = await signalRepo.findOne({
       where: { sourceConnectionId: sourceConnection.id, externalOrderId },
     });
     expect(signalAfterSync).toBeNull();
   });
+
+  // #2069 review — the exact race the signal exists for: the source's order
+  // RESOURCE still reports its pre-cancel status when the sync job's
+  // `getOrder` runs (the event journal that produced the signal leads the
+  // resource), so a downstream gate keyed on `incoming.status`/`order.status`
+  // alone must not be trusted once the signal is consumed. Pinned at the unit
+  // level in `order-ingestion.service.spec.ts` ("should not reserve, and
+  // should enqueue stock-restore, when the incoming status still lags a
+  // consumed early-cancellation signal") against `reservationService` and
+  // `jobQueue` mocks — this harness's real Redis-stream job queue and
+  // `IReservationService` binding are not wired to observe those effects
+  // directly here.
 
   it('normal ordering unchanged: create arrives first, no prior cancel — destination order IS created', async () => {
     const dataSource = harness.getDataSource();
@@ -260,5 +286,62 @@ describe('Order early-cancellation signal (#2069)', () => {
       where: { sourceConnectionId: sourceConnection.id, externalOrderId },
     });
     expect(signalRow).toBeNull();
+  });
+
+  // #2069 review — `ON CONFLICT DO NOTHING` is the port's stated first-write-
+  // wins guarantee, but nothing previously called `record()` twice for the
+  // same key against real Postgres: the unit spec mocks `Repository.query`
+  // to resolve, which asserts nothing about the conflict clause and would
+  // pass identically against a plain `INSERT`.
+  it('first-write-wins: two cancel events for the same order keep the EARLIER cancelledAt', async () => {
+    const dataSource = harness.getDataSource();
+    const signalRepo = dataSource.getRepository(OrderCancellationSignalOrmEntity);
+    const externalOrderId = 'early-cancel-repeat-1';
+    const sourceAdapterKey = registerSourceStub(
+      harness,
+      externalOrderId,
+      makeIncomingOrder(externalOrderId)
+    );
+
+    const sourceConnection = await createTestConnection(dataSource, {
+      platformType: 'allegro',
+      name: 'Allegro source',
+      adapterKey: sourceAdapterKey,
+      enabledCapabilities: ['OrderSource'],
+    });
+
+    const firstCancel = await ingestion.syncOrderFromSource(
+      sourceConnection.id,
+      externalOrderId,
+      'cancel-evt-1',
+      'cancelled'
+    );
+    expect(firstCancel).toEqual([]);
+
+    const signalAfterFirst = await signalRepo.findOne({
+      where: { sourceConnectionId: sourceConnection.id, externalOrderId },
+    });
+    expect(signalAfterFirst).not.toBeNull();
+
+    // A second, later-arriving cancel event for the SAME
+    // (sourceConnectionId, externalOrderId) — e.g. a duplicate delivery or a
+    // reconciliation poll re-observing the cancel. `ON CONFLICT DO NOTHING`
+    // must leave the row — and its earlier `cancelledAt` — untouched.
+    const secondCancel = await ingestion.syncOrderFromSource(
+      sourceConnection.id,
+      externalOrderId,
+      'cancel-evt-2',
+      'cancelled'
+    );
+    expect(secondCancel).toEqual([]);
+
+    const signalAfterSecond = await signalRepo.findOne({
+      where: { sourceConnectionId: sourceConnection.id, externalOrderId },
+    });
+    expect(signalAfterSecond).not.toBeNull();
+    expect(signalAfterSecond!.id).toBe(signalAfterFirst!.id);
+    expect(signalAfterSecond!.cancelledAt.getTime()).toBe(
+      signalAfterFirst!.cancelledAt.getTime()
+    );
   });
 });
--- a/apps/api/test/integration/orders/order-early-cancellation-signal.int-spec.ts
+++ b/apps/api/test/integration/orders/order-early-cancellation-signal.int-spec.ts
@@ -228,7 +228,7 @@ describe('Order early-cancellation signal (#2069)', () => {
     // ingestion time — proving `consume()`'s returned timestamp actually
     // reached `markCancelled` rather than merely triggering it.
     expect(record!.cancelledAt).not.toBeNull();
-    expect(record!.cancelledAt.getTime()).toBe(signalAfterCancel!.cancelledAt.getTime());
+    expect(record!.cancelledAt!.getTime()).toBe(signalAfterCancel!.cancelledAt.getTime());
 
     const signalAfterSync = await signalRepo.findOne({
       where: { sourceConnectionId: sourceConnection.id, externalOrderId },
--- a/apps/api/test/integration/orders/order-early-cancellation-signal.int-spec.ts
+++ b/apps/api/test/integration/orders/order-early-cancellation-signal.int-spec.ts
@@ -158,11 +158,13 @@ describe('Order early-cancellation signal (#2069)', () => {
       platformType: 'allegro',
       name: 'Allegro source',
       adapterKey: sourceAdapterKey,
+      enabledCapabilities: ['OrderSource'],
     });
     await createTestConnection(dataSource, {
       platformType: DEST_PLATFORM_TYPE,
       name: 'Destination',
       adapterKey: DEST_ADAPTER_KEY,
+      enabledCapabilities: ['OrderProcessorManager'],
     });
 
     // Act 1: the cancel event arrives before the order has ever been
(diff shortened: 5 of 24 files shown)


norbert-kulus-blockydevs:
## PR Review — #3035 (senior backend architect pass)

**Scope reviewed**: new `order_cancellation_signals` table/repository/port, `OrderIngestionService.handleSourceCancellation`, `OrderRecordService.persistIncomingSnapshot`, migration, tokens/module wiring, int-spec.

### Architecture & layering ✅
- Correct hexagonal placement: `OrderCancellationSignalRepositoryPort` in `domain/ports/`, TypeORM impl in `infrastructure/persistence/repositories/`, wired via Symbol token (`ORDER_CANCELLATION_SIGNAL_REPOSITORY_TOKEN`) in `orders.module.ts` — no concrete-class injection anywhere.
- No FK to `order_records`, correctly following the `order_holds` / `order_changes` / `refund_records` precedent (indexed reference by value) — and here it's *necessary*, not just stylistic, since the referenced order genuinely doesn't exist yet.
- Rejected the tempting shortcut of minting a phantom internal id via `getOrCreateInternalId` — explicitly cites the #2328 returns-attribution lesson for why that's wrong (a phantom, operator-visible order with no real data). Correct call.

### Idempotency & concurrency ✅ (mostly)
- `record()` uses `INSERT ... ON CONFLICT DO NOTHING` — first-write-wins, matches `OrderRecordRepository.markCancelled`'s idiom.
- `consume()` is a single atomic `DELETE ... RETURNING` via `DeleteQueryBuilder`, correctly avoiding `Repository.query()`'s raw-driver `[rows, rowCount]` footgun (well-documented in the file header, with the exact TypeORM internals cited — good diligence).
- Self-healing race handling is well-reasoned: `consume()` runs unconditionally on **every** `persistIncomingSnapshot` call (not gated to "first call only"), so a signal that misses one interleaving is picked up by the very next poll/webhook rather than lost.
- **Minor gap (SUGGESTION, non-blocking)**: the int-spec only exercises the two *sequential* orderings (cancel-then-sync, sync-then-normal). Given the port docblock explicitly describes a genuine concurrent race ("two concurrent ingestion attempts... can never both apply it"), a test that actually drives two overlapping `consume()` calls against real Postgres (or `Promise.all` of two `syncOrderFromSource` calls) would give direct evidence for that claim rather than relying on inference from Postgres's `DELETE...RETURNING` atomicity. Not blocking — the atomicity is real and well-established elsewhere in the codebase — but it's the one place where "we tested it" and "we reasoned about it" currently diverge.

### Single-writer / column discipline ✅
- `record()` has exactly one caller (`handleSourceCancellation`'s unknown-order branch); `consume()` has exactly one caller (`persistIncomingSnapshot`). No second writer risk.
- Correctly reasoned that `persistOrder`'s `upsertWithLineItems()` excludes `cancelledAt` from its column list, so the ready-path write can't clobber what the awaiting-mapping path already set — verified this against the surrounding code and it holds.
- Reuses the existing `recordCancellationIfNeeded` / `markCancelled` pipeline (#1984) identically from both the signal-consume path and the ready path — no new cancellation-state machinery introduced, no conflation with ADR-059's derived lifecycle phase (this stays purely on the transport-status side, `OrderStatus`/`cancelledAt`, never touching `OrderLifecyclePhase`).

### Failure handling ✅
- The unknown-order signal write is deliberately **not** try/caught (unlike the known-order `markCancelled` write below it), with a clear inline rationale: a DB failure here should retry the whole ingestion job (ADR-007-consistent — this is the only action on that branch, so failing loud is correct), whereas the known-order branch already has downstream relay work that must proceed regardless.
- Insert is idempotent (`ON CONFLICT DO NOTHING`), so a retry after a transient failure is safe.

### Migrations ✅
- `1878000000000` — strictly after the current tail (`1877000000000`), so `check-migration-timestamps.mjs` passes; class name matches filename timestamp.
- Both `up()`/`down()` implemented, `IF NOT EXISTS`/`IF EXISTS` guards throughout (defensively idempotent even though this is a brand-new table — harmless, consistent with recent migrations' self-healing style).
- Class-level `@Index` name on the ORM entity matches the migration's literal index name exactly (`UQ_order_cancellation_signals_source_external`) — correctly avoids the `synchronize`-vs-migration schema divergence trap called out in the entity's own docblock.
- New table correctly added to `apps/api/test/integration/setup.ts`'s `tablesToTruncate` (confirmed).

### Tests ✅
- Unit: repository spec, extended `order-ingestion.service.spec.ts` / `order-record.service.spec.ts`.
- Integration: both ACs from #2069 are directly asserted — `cancel-then-sync` proves no destination order is created (`createOrderCalls` stays empty, `syncResult[0].status === 'skipped_cancelled'`, signal row consumed/deleted), and the normal-ordering regression test proves untouched behavior when no early signal exists.
- The mechanical fallout (adding `recordEarlyCancellationSignal: jest.fn()` to five unrelated mocks of `IOrderRecordService` in shipping/refunds specs) is expected collateral from widening the interface, not scope creep — verified each is a pure stub addition with no behavior change.

### Verdict: ✅ **Approve**

Clean, well-reasoned, idiomatic implementation that correctly follows this repo's established patterns (claim-based atomic writes, indexed-reference-by-value instead of a phantom FK, migration/entity index-name parity). One non-blocking suggestion: add a genuinely concurrent test for the race the port's own docblock describes, to move that guarantee from "reasoned about" to "directly tested."

🤖 Reviewed with [Claude Code](https://claude.com/claude-code)

piotrswierzy (changes requested):

The diagnosis is correct and the key choice is right: `(sourceConnectionId, externalOrderId)` is the only durable identity that exists pre-ingestion. And the **best call in the PR** is funnelling the consumed signal through the *existing* `markCancelled` writer rather than minting a phantom internal id or adding a second writer — that keeps #1984/#2100's single-writer discipline and the #2284 guard intact with no change to either.

Migration hygiene, layering, naming and doc quality are all high. One substantive problem blocks it.

### BLOCKING — `cancelledAt` becomes reachable by a second route, and only 1 of its 4 consumers is updated

`order-ingestion.service.ts` (423, 526, 538) + `order-record.service.ts` (445–465)

Three downstream gates in `syncOrderFromSource` discriminate on the **source-reported status**, not on `cancelledAt`:

| Site | Gate | Fed by this PR? |
|---|---|---|
| `reserveOrderInventory` (~1537) | `if (order.status === 'cancelled') return;` | ✗ |
| early stock-restore hook (~423) | `incoming.status === 'cancelled' && priorStatus !== 'cancelled'` | ✗ |
| post-`persistOrder` restore hook (~538) | `order.status === 'cancelled' && …` | ✗ |
| `OrderSyncService` #2284 guard | `record.cancelledAt IS NULL` | ✓ |

On **the exact race this PR exists to close**, the source's order resource frequently still reports the pre-cancel status when `getOrder` runs — the event journal leads the resource, and Allegro's checkout-form status lags. In that case destination provisioning is correctly skipped (the AC is met), but:

- **`reserveOrderInventory` creates advisory holds for an order OL already knows is cancelled**, and
- **no `marketplace.offer.stockRestore` job is enqueued**, so #2348's `restoreStockForCancelledOrder` — the ordered release-then-restore — never runs, and nothing else releases those holds.

Per `architecture-overview.md § Inventory`, `closeForOrder` has exactly two production callers (cancellation and OL-owned dispatch), and the expiry sweep (#2346) releases nothing while `UnavailableOrderHoldReader` is bound. So the holds stay `held` for the life of the install → a permanent ATP subtraction plus a permanent, never-clearing `reservation_shortfall_episodes` row naming a cancelled order. That is the #2628 failure class precisely.

The plan's *Out of Scope* bullet argues symmetry with the known-order `handleSourceCancellation` path, but **that argument does not transfer**: there the order's holds already exist and the cancellation flow reaches #2348. Here `reserveOrderInventory` runs *after* OL has already recorded the cancellation.

**The remedy is already available and two lines.** `persistIncomingSnapshot` returns the refreshed `OrderRecord` (`cancellationWrote ? findById : saved`) and the call site at ~404 discards it with a bare `await`. Capture it and OR `persisted.cancelledAt !== null` into the three gates — or gate `reserveOrderInventory` on `cancelledAt` and let the restore hook read the same. Either way this needs *deciding*, not omitting.

**And the int-spec passes because it does not look.** `makeIncomingOrder` hardcodes `status: 'pending'` (`order-early-cancellation-signal.int-spec.ts:66`) — it encodes exactly this scenario — while asserting only `cancelledAt !== null` and `createOrderCalls.length === 0`. Add an assertion on `reservations` / the enqueued job set so the decision is pinned rather than incidental.

### IMPORTANT — the headline first-write-wins guarantee is untested, while a docblock says otherwise

`order-cancellation-signal.repository.spec.ts` (30–33) states *"The database-level guarantee (the unique index, first-write-wins) is asserted against real Postgres in `order-early-cancellation-signal.int-spec.ts`."* **It is not** — that int-spec has two cases and neither calls `record()` twice. The unit test that nominally covers it (~96) mocks `Repository.query` to resolve, so it asserts nothing about `ON CONFLICT` and would pass identically against a plain `INSERT`.

So `ON CONFLICT DO NOTHING` — which the port docblock names as the guarantee — is untested in both directions while a docblock claims coverage. Add a third int-spec case: two consecutive cancel events → one row, and the **earlier** `cancelledAt` survives.

### IMPORTANT — entity/migration type divergence

`order-cancellation-signal.orm-entity.ts:39` declares `@Column({ type: 'varchar' })` for `externalOrderId` (→ `character varying` under `synchronize`) while the migration creates `text`. The direct precedent for this exact key shape, `ReturnOrmEntity.externalReturnId`, uses `type: 'text'` specifically to match its migration. The migration's own header argues at length that migration↔`synchronize` parity is load-bearing for the index name — the column type is the same argument, and `fulfillment-work-migration-parity.int-spec.ts` diffs column **types**.

### IMPORTANT — `consume()` adds an unconditional write to the hottest path

Every `persistIncomingSnapshot` call now issues a `DELETE` — one extra statement (WAL, index probe) per order ingestion, on what `reserveOrderInventory`'s own kill-switch comment calls *"the hottest path in the system"*. The unconditional call is **correct** and is what makes the design self-healing, but unlike `reserveOrderInventory` there is no escape hatch and no measurement. A zero-row `DELETE` on a two-column unique index is cheap; state the cost in the docblock (short-circuiting on a read would forfeit the atomicity, so probably just document it).

### IMPORTANT — no observability for a signal that is never consumed

The orphan sweep is out of scope, which is defensible. But the table carries a **unique** key, so an orphan permanently holds its `(sourceConnectionId, externalOrderId)` slot, and there is no count, log token or operator surface for "recorded but never consumed". The closest analogue, `master.product-reconcile.remainingNull` (#2317), is deliberately observable. At minimum warn on insert so an install where this fires constantly is diagnosable.

### SUGGESTIONS

- The write sits on the `!internalOrderId` branch **above** the ADR-017 destination-echo guard — correct, since the guard needs a record. But a cancel arriving on a *destination* connection whose `Order` mapping is not yet written records a signal keyed to a destination connection id, permanently unconsumable (the echo guard early-returns before `persistIncomingSnapshot`). Narrow and self-limiting; worth one sentence.
- `expect(record!.cancelledAt).not.toBeNull()` doesn't distinguish "set from the signal's instant" from "set from `now`". `persistIncomingSnapshot` deliberately prefers `earlySignalAt ?? now` — assert the value.
- `{} as unknown as OrderCancellationSignalRepositoryPort` in `order-record-packed.service.spec.ts:87` — a `{record: jest.fn(), consume: jest.fn()}` stub costs one line and turns a latent `TypeError` into a visible assertion.
- The plan is `**Status**: Draft` while shipping with its implementation.

### Done well

- **#2100's four points check out**: `cancelledAt` excluded from both `toOrm` and `upsertWithLineItems` (2497, 2976); `markCancelled` is the sole narrow atomic writer (1993); `recordCancellationIfNeeded` returns early on `isCancelled === false` (560), so **re-ingestion cannot clobber the signal-set value** — which answers the ADR-057 question directly. Read-guard coercion and a counted IN-list are both correctly N/A here (a `timestamptz` has no unrecognised-value space; `cancelledAt IS NULL` *is* the right predicate, not a reason vocabulary).
- **Migration hygiene clean**: `1878000000000` > main's tail, unique across core and both plugin dirs, class suffix matches filename, both `up()`/`down()` `IF EXISTS`-guarded. The unique index is non-partial on exactly the two equality columns both `ON CONFLICT (…)` and `consume()`'s `WHERE` use, so it genuinely serves both with no inference problem. The `uuid_generate_v4()`/`now()` defaults are permanent by design and mirror the decorators, so #2317's drop-the-default rule is correctly N/A.
- **The `Repository.query` / `[rows, rowCount]` finding is a genuine catch**, and `expect(ormRepository.query).not.toHaveBeenCalled()` is the honest way to pin it.
- **The try/catch asymmetry is right and tested** — leaving the new write unguarded (unlike the sibling `markCancelled`, which must not block the relay) is correct: the write is the branch's only action, the job retries, and `ON CONFLICT DO NOTHING` makes the retry idempotent.
- `skipped_cancelled` stays distinguishable from a routing/`held` skip (#2397). `tablesToTruncate` correctly updated with a comment explaining why the CASCADE walk can't reach it.

**Merge gate**: the PR states `pnpm test` / `pnpm test:integration` were not run. Both must be green — six `IOrderRecordService` mock shapes changed across `orders` and `shipping`, and the new int-spec registers adapters against the shared harness registry. (CI on this head currently reads green, but please confirm locally too.)

---

piotrswierzy (commented):

The BLOCKING fix is real and took the remedy offered: `:404` captures the record, `:424` derives `cancelledFromEarlySignal`, and all three gates OR it in (`:434`, `:551`, `:1607` via a new `alreadyCancelled` param). The unit spec now asserts **both** effects — `reserveForOrder` not called *and* stock-restore enqueued — and the int-spec carries a comment explaining why the harness cannot observe it rather than silently changing the fixture. That is the right way to handle an unobservable case.

| Prior finding | Verdict |
|---|---|
| 3 of 4 `cancelledAt` consumers gate on source status | 🔄 **Partial — one poll deep** |
| `ON CONFLICT` untested both directions | ✅ Resolved |
| Entity `varchar` vs migration `text` | ✅ Resolved |
| Unconditional `DELETE` on the hot path | ✅ Resolved (documented, which was the ask) |
| No observability for never-consumed rows | ❌ Unaddressed |
| All four suggestions | ✅ Resolved |

### The residue — it works for one poll, then the original defect returns

`persistIncomingSnapshot` returns `saved` unless it wrote the cancellation, and `upsert`'s `fromRawRow(row, writeSet)` resets every column outside the write set — `cancelledAt` is deliberately outside it. So `snapshotRecord.cancelledAt` is **always `null`** on the non-writing path.

On any poll *after* the consuming one — signal deleted, source still lagging — the flag is `false` and the holds-nothing-will-release behaviour is back verbatim. One line, against a record already loaded at `:362`:

```ts
snapshotRecord.cancelledAt != null || existing?.cancelledAt != null
```

Worth a third-invocation test, since the current specs only exercise the consuming poll.

**One correction to my own earlier wording**: the "warn on insert" clause I asked for was already satisfied before that review (`:1284`). What genuinely remains is that a recorded-never-consumed signal holds its unique-key slot with no count, sweep or operator query — and `record()` returns `void`, so nothing can even report insert-vs-conflict.

### IMPORTANT — `@CreateDateColumn()` has the same divergence just fixed one field over

`order-cancellation-signal.orm-entity.ts:44` — bare `@CreateDateColumn()` maps to `timestamp` under `synchronize` while the migration creates `timestamptz`. Every recent sibling entity declares it explicitly. Same class as the `varchar`/`text` finding this PR just resolved.

### SUGGESTION

The int-spec's "earlier `cancelledAt` survives" half compares two `new Date()`s and would pass against a `DO UPDATE`.

All nine CI checks green. Migration hygiene otherwise clean — class suffix matches, both directions `IF [NOT] EXISTS`-guarded.

### ⚠️ Cross-PR — `1878000000000` is claimed by four open PRs

| PR | migration |
|---|---|
| #3035 | `create-order-cancellation-signals` |
| #3158/#3159 | `create-price-change-episodes` |
| #3027 | `seed-erli-prompt-template` |
| #3014 | `add-order-record-destination-routing-block` |

`check-migration-timestamps.mjs` diffs only against `origin/main` (tail `1877000000000`), so **each passes independently** and the #374 duplicate-timestamp defect surfaces after the first merge. No local guard can see this. Needs coordinating across the four now, not at merge time.

---

piotrswierzy (commented):
## Tech Lead review

Third pass, at `2e9e52b` — the residue commit. Three of the four findings left open by the previous re-review are **properly fixed**: the `existing?.cancelledAt` fallback (`order-ingestion.service.ts:436`), `@CreateDateColumn({ type: 'timestamptz' })`, and the int-spec's explicit delay so the first-write-wins case can't pass under a `DO UPDATE`. The reasoning in the new comment block at `:424-435` is exactly right.

But the unit test added to *prove* the fix is **red in CI, and red for a reason that also makes its other assertion vacuous** — so the residue fix currently has no working evidence behind it at all.

### Findings

**[BLOCKING]** — `libs/core/src/orders/application/services/__tests__/order-ingestion.service.spec.ts:468`
> `should not reserve, and should enqueue stock-restore, on a poll AFTER the signal was already consumed` fails in CI (`libs/core`: 1 failed / 6371 passed — `expect(jobQueue.enqueue).toHaveBeenCalledWith(...)`, **Number of calls: 0**).
>
> The cause is not the production fix. The test overrides `getOrderRecord` with `{ cancelledAt: … } as unknown as OrderRecord` — **no `sourceConnectionId`** — and `syncOrderFromSource` hits the ADR-017 destination-echo guard nine lines after loading it:
>
> ```ts
> // order-ingestion.service.ts:363
> if (existing && existing.sourceConnectionId !== connectionId) { … return []; }
> ```
>
> `undefined !== 'connection-123'` is true, so the call **returns `[]` immediately** and never reaches line 450. That has two consequences, and the second is the serious one:
>
> 1. `jobQueue.enqueue` is never called → the assertion fails. Visible.
> 2. `reserveForOrder` is never called **for the same reason** → `expect(reservationService.reserveForOrder).not.toHaveBeenCalled()` passes **vacuously**. Invisible.
>
> So the test exercises none of `cancelledFromEarlySignal`. Had only assertion (1) been written the suite would be green and would assert nothing about the residue — the "a check that cannot fail" class this repo has been bitten by before (#2673, and the vacuous-parse guard #3014 had to harden). Add `sourceConnectionId: connectionId` to the mock, and give it an `orderSnapshot` so `priorStatus` is chosen deliberately rather than falling out of an absent field — the gate is `(… || cancelledFromEarlySignal) && priorStatus !== 'cancelled'`, and a `priorStatus` that happens to be `undefined` is not the same evidence as one that is deliberately non-`'cancelled'`.
>
> Note the sibling test immediately above (`:442`, the previous round's) does **not** override `getOrderRecord`, so it reaches the gate legitimately and is unaffected.

**[IMPORTANT]** — `apps/api/src/migrations/1879000000000-create-order-cancellation-signals.ts`
> The renumber `1878000000000 → 1879000000000` moved the clash rather than resolving it. `1879000000000` is now **also** claimed by **#3161** (`create-price-change-auto-applied-log`) and **#3219** (`add-order-record-sales-document-matched-rule`); `1878000000000` remains claimed by **#3014** and **#3027**, and **#3201**/**#3210** sit on `…0000`/`…0001`. `check-migration-timestamps.mjs` diffs only against `origin/main` (tail `1877000000000`), so all six still pass in isolation and the #374 defect surfaces after the first merge. This can only be fixed by sequencing across the set — pick a lane per PR now (e.g. `1878000001000` upward, one step each) rather than at merge time.

**[SUGGESTION]** — `libs/core/src/orders/domain/ports/order-cancellation-signal-repository.port.ts:32`
> Still the one carried-over finding: a recorded-never-consumed signal permanently holds its `(sourceConnectionId, externalOrderId)` unique slot with no count, sweep or operator query, and `record()` returns `Promise<void>` so nothing can distinguish an insert from a conflict. The warn at `order-ingestion.service.ts:1297` covers the *arrival*, which was the smaller half of the ask. Returning a `boolean` from `record()` costs nothing now and is the thing a later sweep would need.

### Re-verified sound

`cancelledAt` remains excluded from both `toOrm` and `upsertWithLineItems`; `markCancelled` is still the sole narrow atomic writer and `recordCancellationIfNeeded` still early-returns on `isCancelled === false`, so re-ingestion cannot clobber the signal-set value — the #2100 single-writer discipline the task asked about holds. `skipped_cancelled` (#2284) stays distinguishable from a routing or `held` skip. `consume()` is still a single atomic `DELETE … RETURNING` through the query builder, avoiding `Repository.query`'s `[rows, rowCount]` shape. Migration class suffix matches the renamed filename, both directions stay `IF [NOT] EXISTS`-guarded, and the class-level `@Index` name still matches the migration literal. `Integration Tests` is green at 35m58s, so the table and index really were created.

### Verdict
❌ **Request changes** *(posted as a comment — this repo's review flow does not use `--request-changes`)*

1. Fix the red test's mock (`sourceConnectionId` + `orderSnapshot`) and confirm it fails against a build with the `existing?.cancelledAt` fallback removed — red-first, or it is not evidence.
2. Coordinate the migration timestamp across the six open PRs now.
3. Optional: make `record()` return whether it inserted.

---

piotrswierzy (commented):
## Tech Lead re-review (`2e9e52b` → `304c97e`)

Fourth pass. The delta is one fix commit on top of a `main` merge. **Both of the previous round's open findings are properly closed**, and the BLOCKING one is closed for the right reason rather than by making the test agree with itself. The single remaining item is the migration timestamp, which is now contended by four open PRs at once and can only be resolved by sequencing across them.

### Prior findings

**BLOCKING — red/vacuous residue test — CLOSED.** `order-ingestion.service.spec.ts:468` now mocks `sourceConnectionId: connectionId` and `orderSnapshot: { status: 'BOUGHT' }`, so the ADR-017 destination-echo guard at `order-ingestion.service.ts:363` no longer short-circuits and `priorStatus` is a deliberate non-`'cancelled'` value rather than an accidental `undefined`. The `Test` job is green (14m53s), so the `jobQueue.enqueue` assertion genuinely passes. More importantly the sibling assertion is no longer vacuous: the same `reservableIncoming` fixture is used by three tests in that `describe` that DO reach `reserveOrderInventory` (`:406` asserts `toHaveBeenCalledTimes(2)`), so `not.toHaveBeenCalled()` here is now a statement about the gate rather than about an early `return []`.

It is also **red-first by construction**, which is the part I asked for: `persistIncomingSnapshot` is mocked to return `cancelledAt: null`, so the only route to `cancelledFromEarlySignal === true` is the `existing?.cancelledAt` fallback at `:436-437`. Delete the fallback and both assertions flip. No manual demonstration needed — the test cannot pass without the fix it exists to prove.

**SUGGESTION — `record()` returns whether it inserted — CLOSED, and correctly.** `RETURNING "externalOrderId"` with `rows.length > 0` is the right discriminator: `ON CONFLICT DO NOTHING` produces no row on the conflict arm, so no second statement is needed. I checked the shape rather than assuming it — `typeorm@0.3.17` `PostgresQueryRunner.query` routes `INSERT` to the `default` arm (`result.raw = raw.rows`), and only `UPDATE`/`DELETE` get the `[rows, rowCount]` tuple that `consume()` avoids by using the query builder. So the array read is sound.

### Findings

**[IMPORTANT]** — `apps/api/src/migrations/1881000000000-create-order-cancellation-signals.ts`
> `1878 → 1879 → 1881` has moved the clash twice without resolving it. `1881000000000` is now claimed by **four** open PRs simultaneously: this one, **#3027** (`seed-erli-prompt-template`), **#3159** (`create-price-change-auto-applied-log`) and **#3201** (`normalise-sales-document-country-codes`). `main`'s tail is `1880000000000` (merged with #3158), so `check-migration-timestamps.mjs` — which diffs only against `origin/main` — passes for all four in isolation and the #374 collision lands the moment the second one merges. The neighbouring lanes are already spoken for too: `1881000000001` (#3210), `1882000000000` (#3219), `1883000000000` (#3227). Pick a distinct lane now rather than at merge time — `1884000000000` upward, or `1881000000002..0005`, one step per PR. (Unrelated but visible from the same scan: **#3014** still sits at `1878000000000`, which is now *below* `main`'s tail and therefore an ordering violation in its own right.)

**[SUGGESTION]** — `order-cancellation-signal.repository.spec.ts:63`
> The new boolean is asserted only against a mocked `Repository.query`, so the unit test would pass identically if `RETURNING` were dropped — the same criticism the `first-write-wins` int-spec was written to answer for the conflict clause. The int-spec at `order-early-cancellation-signal.int-spec.ts:323` already calls `record()` twice against real Postgres; asserting the returned boolean there (first `true`, second `false`) would cost one line and make the discriminator real-DB-verified. Not blocking: nothing in production reads the value today, and the TypeORM return shape is verifiable from source.

### Re-verified sound

The #2100 single-writer discipline still holds at head: `cancelledAt` stays out of both `toOrm` and the `upsertWithLineItems` write set (documented at `order-record.repository.ts:2504`, `:2548`, `:3001`, `:3027`), `markCancelled` remains the sole narrow writer and is `COALESCE`-guarded (`:2044-2049`), and `recordCancellationIfNeeded` still early-returns so re-ingestion cannot clobber a signal-set value. `skipped_cancelled` (#2284) stays distinguishable. The migration's class suffix matches its renamed filename, both `up()` and `down()` are `IF [NOT] EXISTS`-guarded, and `down()` drops only what `up()` created. Every CI job is green, including `Integration Tests` at 31m43s, so the table and its unique index really are created by the migration path.

### Verdict
🔄 **Approve with changes** — the code is done; the only thing left is to claim a free migration timestamp before this merges.

---

norbert-kulus-blockydevs:
Heads-up on a migration-timestamp collision, raised in review of #3201.

This PR's `1881000000000-create-order-cancellation-signals.ts` and #3201's `1881000000000-normalise-sales-document-country-codes.ts` hold the same timestamp. `scripts/check-migration-timestamps.mjs` compares the working tree against `origin/main`, so neither branch sees the other today and whichever merges **second** will fail lint on rebase.

We are not moving ours, for two reasons rather than one:

- #3201 sits in the middle of a 13-PR stack (#3199 → #3219) that also holds `1881000000001` and `1882000000000`; renumbering one of them forces the other two up as well, to keep the stack's own ordering intact.
- That stack is 12/13 approved and expected to merge first.

So the renumber is likely to land here. No action needed until this PR is unblocked — just flagging it now rather than after a confusing rebase failure.

piotrswierzy (approved):
## Tech Lead re-review (`304c97e0` → `8ddea60e`)

One commit, exactly the fix I asked for and nothing else: the migration moves to `1885000000000` and the class + `name` suffix move with it in the same diff (`+2/-2`), so rules 1-3 of the timestamp invariant hold at head. I checked the claim rather than the description. `1885000000000` appears nowhere on `origin/main` (tail `1880000000000`) and on no other open PR: the live claims are `1878` (#3014), `1880` (#3162, #3159), `1881` (#3027, #3201), `1881000000001` (#3210), `1882` (#3219), `1883` (#3227), `1884` (#3241, #3159). This PR sits strictly above all of them, so its ordering survives whatever order that field merges in — the two genuine clashes that remain (`1881` between #3027/#3201, `1884` between #3241/#3159) are those PRs' problem now, not this one's. The sole prior IMPORTANT is closed.

Re-confirming what I asked to re-confirm, at head rather than from memory: the #2100 single-writer discipline is intact — `cancelledAt` appears in `toOrm` only as the comment explaining its exclusion, `upsertWithLineItems` does not write it, `markCancelled` is the one narrow writer and is `COALESCE`-guarded (`order-record.repository.ts:2044-2049`), and the `entity.cancelledAt ?? null` at `:2943` is inside `toDomain`'s hydration tuple, not a write set. `recordCancellationIfNeeded` still early-returns on `!isCancelled`, so a re-ingestion of a live order cannot reach the writer at all. `skipped_cancelled` (#2284) remains its own terminal arm in `order-ingestion.service.ts`, ordered ahead of the hold predicate and commented as such. The migration is `IF [NOT] EXISTS`-guarded on both table and index, `down()` drops only what `up()` created in reverse order, and no stale `1881` reference survives in the implementation plan.

### Findings

### Verdict
✅ **Approve**

CI has not finished on this head: `Docker Build Smoke Test`, `Detect changed paths`, `PHP Unit Tests` and `Scaffolded Adapter Builds` pass; `Lint`, `Type Check`, `Test`, `Build` and `Integration Tests` are **pending** — queued or running, no failed assertion anywhere. The earlier head's `Integration Tests` went green in 31m43s and the delta is a filename plus two identifier characters, so I expect the same, but the approval is on the understanding that the gate is watched rather than assumed. `mergeStateStatus: BLOCKED` is those pending required checks, not a conflict (`mergeable: MERGEABLE`). Merge once they land.

---

piotrswierzy (commented):
## Tech Lead re-verification (head unchanged at `8ddea60`)

Four of the five checks I approved on the promise of watching have now resolved **green** — `Lint`, `Type Check`, `Build`, `Docker Build Smoke Test`, plus `Scaffolded Adapter Builds` and `PHP Unit Tests`. `Test` is still `queued` behind the runner backlog.

**`Integration Tests` came back red**, and that changes the verdict. I approved explicitly on the basis that the gate is watched, not assumed, so I am not going to wave this through.

### Findings

**[IMPORTANT] — `Integration Tests` failed on run `34838885603` (job `103959019162`).** One assertion, in a spec this PR does not touch:

```
FAIL test/integration/reservations-ledger.int-spec.ts
  ● Reservation ledger (#2343) › the guarded claim
    › should grant exactly one of two concurrent claims when stock covers only one
    expect(received).toHaveLength(expected)
    Expected length: 1   Received length: 2
```
`Test Suites: 1 failed, 182 passed`; the worker integration step passed 28/28. Both `claimHeld` calls were granted against a position seeded with `availableQuantity: 1`, each reporting `previousQuantity: 0` — a lost update the guarded conditional UPDATE exists to prevent.

Everything points to a pre-existing timing flake rather than a regression from this PR: the diff touches no inventory, reservation or availability code (its only shared-surface edit is adding the new table to `apps/api/test/integration/setup.ts`'s truncation list, which is additive); `main` is green at `e336cf4`; and the two "concurrent" claims are stamped 8 ms apart, i.e. they did not actually overlap the way the test needs them to. But *probably a flake* is not evidence, and this is precisely the class of red that gets re-run into invisibility. **Re-run `Integration Tests` and merge only on green**; if it reproduces, it is a genuine concurrency defect in #2343's guard and belongs in its own issue rather than riding on this PR.

**Migration timestamp — re-verified against the current claim map, which has moved.** `1885000000000` is still free and still strictly above `main`'s tail (`1880000000000`). The map is no longer the one from last round: **#3159 has moved off `1884000000000` and now claims `1887000000000` + `1888000000000`**, and `1884000000000` is now **#3241**'s. Remaining open claims: `1878` → #3014, `1880500000000` → #3201, `1881000000000` → #3027, `1881000000001` → #3210, `1882` → #3219, `1883` → #3227. No collision; `1886` is unclaimed between this and #3159.

The rest of the approval basis is unchanged at this head: `cancelledAt` out of `toOrm` and `upsertWithLineItems`, `markCancelled` the lone `COALESCE`-guarded writer, `recordCancellationIfNeeded` early-returning, `skipped_cancelled` a distinct terminal arm.

### Verdict
🔄 **Approve with changes** — the code is unchanged and still approved; merge is conditional on a green `Integration Tests` re-run (and on `Test` clearing the queue).

---

PR #3036, opened 2026-09-09
feat(dx): enforce unique ADR numbers via check:invariants

### LINKED ISSUE #2082: [TASK] DX — enforce unique ADR numbers via check:invariants
## Problem / Context

ADR numbers are allocated by hand and nothing enforces uniqueness. Because two ADRs claiming the same number have **different filenames**, git merges both cleanly with no conflict marker — so each PR passes in isolation and only the *second* to merge is wrong, silently producing two ADRs with the same number in a record `docs/architecture/adrs/README.md` describes as append-only.

This is the same failure mode as the migration-timestamp collision, which `scripts/check-migration-timestamps.mjs` already guards. ADRs have no equivalent.

It has cost review time in three consecutive rounds over two days:

1. **#2050 vs #2056** — both claimed `040`, both based on the same commit. Caught in review; #2056 renumbered to `042`.
2. **#2056's reservation note was wrong in both directions** — it stated 039 was the sales-document routing ADR (it is #2014's analytics read model) and that #2055 claimed 039 (it claims 041). As written it steered the next author into #2055's number.
3. **#2066 then claimed `039`, `040` and `041` simultaneously**, colliding with #2014, #2050 and #2055 at once.

Current state at time of writing:

| Number | Claimants |
|---|---|
| 039 | #2014 (draft), #2066 |
| 040 | #2050, #2066 |
| 041 | #2055, #2066 |
| 042 | #2056 (clean) |
| 043 | next free |

`039` is additionally referenced **by filename** from a doc already merged to `main` (`docs/plans/implementation-plan-order-cancellation-record-state.md:164`), so a wrong allocation there breaks an existing cross-reference rather than merely duplicating a number.

Two of the affected PRs added hand-maintained "reserved numbers" notes to the README. That is the right instinct and it does not work — a note cannot see a PR opened after it was written, which is exactly how #2066 collided with three branches whose notes all predated it.

## Proposed Solution

`scripts/check-adr-numbers.mjs`, chained into `check:invariants` (and therefore into `pnpm lint` and the pre-commit hook), mirroring the shape of `scripts/check-migration-timestamps.mjs`.

Three assertions:

1. **Unique** — no two files in `docs/architecture/adrs/` share a 3-digit prefix.
2. **Consistent** — the `# ADR-NNN:` heading inside the file matches its filename prefix. Catches the half-rename where one side is updated and the other isn't (#2056 had to update filename, heading, README row, `architecture-overview.md` pointers and a product-spec reference — five places, and the guard should catch a miss).
3. **Not already taken on `origin/main`** — a number not on `main` locally must also not exist on `main`, compared via `git ls-tree`, mirroring the migration guard's ordering check. This is the assertion that actually catches the cross-PR case, since the collision is only visible when the branch is compared against the trunk.

Assertion 3 inherits the migration guard's git-availability caveat: skip with a notice when `origin/main` is unfetchable locally, hard-fail in CI (`CI=true`), skip entirely when `git` is absent. The `lint` job already fetches `origin/main` after checkout for the migration check, so no workflow change is needed.

**Also worth including:** a `README.md` index row exists for every ADR file and vice versa. The index has drifted twice already, and it is the thing authors actually read when picking the next number.

## Classification

**Type**: DX / tooling
**Layer**: N/A (repo scripts)
**File(s)**:
- new: `scripts/check-adr-numbers.mjs`
- changed: root `package.json` (`check:invariants`), `docs/architecture/adrs/README.md` (point § numbering at the guard instead of "check the index below")

## Assumptions

- The 3-digit-prefix convention holds for every current ADR file (verified: `001`–`038` on `main`).
- Reserved-but-unwritten numbers stay a human convention. The guard checks files that exist; it cannot know a number is spoken for by an unmerged branch until that branch's file exists — which is fine, because the cross-PR case is caught at merge time by assertion 3 rather than needing pre-registration.

## Acceptance Criteria

- [ ] Two ADR files sharing a 3-digit prefix fail `pnpm lint` with both paths named
- [ ] A file whose `# ADR-NNN:` heading disagrees with its filename prefix fails, naming both values
- [ ] A new ADR reusing a number already present on `origin/main` fails, naming the existing file
- [ ] An ADR file with no README index row (and an index row with no file) fails
- [ ] Behaviour with `origin/main` unavailable matches `check-migration-timestamps.mjs`: notice locally, hard failure under `CI=true`, skip when `git` is absent
- [ ] Passes clean against current `main`
- [ ] `docs/architecture/adrs/README.md` § numbering references the guard

## References

- Precedent: `scripts/check-migration-timestamps.mjs`, `docs/migrations.md` § Timestamp uniqueness invariant
- Collisions: #2050 / #2056 (040), #2066 vs #2014 / #2050 / #2055 (039, 040, 041)
- #2055 and #2056 both carry hand-written reserved-number notes this guard would supersede

jakubretajczykBD:
## Summary
- Adds `scripts/check-adr-numbers.mjs`, mirroring `check-migration-timestamps.mjs`'s shape, wired into `check:invariants` (and therefore `pnpm lint` / pre-commit).
- Enforces: local filename-prefix uniqueness, filename-vs-`# ADR-NNN:` heading consistency, no reuse of a number already claimed on `origin/main` under a different filename (the actual cross-PR collision guard — with a `git diff --find-renames` exemption so a same-number slug correction isn't misreported as a rival claim), and a README index ↔ file bijection (including duplicate-row detection).
- Updates `docs/architecture/adrs/README.md` § "How to write a new ADR" to point at the guard, and retires the hand-maintained "Reserved numbers" footnote that had already drifted from reality three times in two days.

## Test plan
- [x] `node scripts/check-adr-numbers.mjs --self-check` — passes (fixtures cover every failure mode named in the issue's acceptance criteria, plus the rename-exemption and duplicate-README-row cases added during review)
- [x] `node scripts/check-adr-numbers.mjs` — passes clean against current `main` (71 ADRs)
- [x] Manually reproduced and confirmed detection of: duplicate ADR number across two files, heading/filename mismatch, missing README index row, cross-PR collision against `origin/main`
- [x] Manually reproduced a real `git mv` rename (same ADR, corrected slug) and confirmed it is *not* flagged as a collision
- [x] `pnpm check:invariants` — full chain passes, including the new check

Closes #2082


norbert-kulus-blockydevs:
## PR Review — #3036 (senior backend/DX tooling pass)

**Scope reviewed**: `scripts/check-adr-numbers.mjs`, its wiring into `check:invariants`, and the README footnote it replaces.

### Verification performed (beyond trusting the PR's own test-plan claims)
- Ran `node scripts/check-adr-numbers.mjs --self-check` and the real run against current `main` — both pass (71 ADRs, 0 violations).
- Confirmed `package.json`'s `check:invariants` chain appends `&& node scripts/check-adr-numbers.mjs --self-check && node scripts/check-adr-numbers.mjs` following the exact `--self-check` + real-run pairing every other mirror/invariant script in this repo uses — consistent with the established convention, not a bespoke shape.
- Confirmed CI already fetches `origin/main` at depth 1 (`.github/workflows/ci.yml:105`) before lint runs, which is what makes the cross-PR collision guard (rule 4) actually load-bearing in CI rather than silently degrading to 'skip'. The script's own CI-vs-local skip/fail split (`resolveMissingBaselineAction`) is correct: `no-git` (git binary itself absent) silently skips even in CI, but a present-git/missing-ref case **hard-fails** CI with an actionable fetch command rather than silently passing — the right failure direction for a guard whose whole point is catching cross-PR collisions.
- **Went past the self-check's synthetic fixtures** and injected real conditions against the actual `docs/architecture/adrs/` tree in a worktree:
  - Copied `001-hexagonal-architecture-and-bounded-contexts.md` to `001-rival-claim.md` (same number, different filename) — correctly caught by **both** rule 2 (local uniqueness) and rule 4 (baseline collision) simultaneously, exit 1.
  - Did a real `git mv` (same-number slug correction) rather than a synthetic `renames` map entry — correctly exempted, exit 0, confirming `git diff --find-renames` actually produces the shape the rename-exemption logic expects against a live git tree, not just the hand-built `Map` in the self-check fixture.
  - Manually verified the README index regex against the **real** table format (`| [ADR-001](./001-...) | ... |`) — matches correctly (confirmed indirectly: rule 5 passing with 0 violations across 71 real rows means the regex isn't silently failing to match and masking every file as 'missing a row', which would be the dangerous false-negative direction).

### Design quality
- Correct false-negative bias throughout: every degraded-input path (git absent, ref unavailable locally, rename detection failing) explicitly falls back to the *stricter* behavior, never the more permissive one — the docblocks say so and the self-check fixtures pin it (e.g. the 'rename map entry pointing at a DIFFERENT baseline file does not exempt a real collision' case).
- Rule 5 (README bijection) inspects every row without deduping first, specifically to catch two rows correctly-individually pointing at the same real file — a subtler defect than 'row references a nonexistent file' that a lazier implementation would miss.
- The heading-extraction regex is scoped to the first non-blank line only, with an explicit fixture proving a heading-shaped string quoted later in the prose (e.g. discussing a superseded ADR) isn't mistaken for the file's own title.
- No architectural concerns — this is a standalone Node script mirroring an established pattern (`check-migration-timestamps.mjs`), not application runtime code, so Hexagonal Architecture/DI/port rules don't apply here and it correctly doesn't try to force them.

### Minor, non-blocking
- `validateAgainstBaseline`'s `baselineByNumber` map (line ~150) would silently keep only the *last* baseline file processed if `origin/main` itself already had a duplicate number under two files. This is theoretical, not live — if that were true today, rule 2's whole-directory scan (which runs before rule 4, unconditionally, over every file on disk) would already be failing on `main` itself, and the PR's own "clean run against current main, 71 ADRs" claim (which I independently reproduced) proves that isn't the case. Worth a one-line docblock caveat if you want to be thorough, but not worth blocking on.

### Verdict: ✅ Approve

Exactly what a checker script that mirrors an existing pattern should look like: self-verifying, false-negative-averse by design, and — per my own independent injection tests — actually does what it claims rather than merely what its own fixtures assert.

piotrswierzy (approved):
## `/pr-review` — ✅ Approve

A well-shaped guard that mirrors `check-migration-timestamps.mjs` faithfully: pure validators, a self-check that can't be vacuous, per-rule negative fixtures, and a fail-closed degradation story.

**Verified by running it, not by reading it:**
- `--self-check` → OK; against the real tree → `OK (71 ADRs; uniqueness vs origin/main: checked)`.
- **Non-vacuity confirmed by live negative tests** — a rival `039-rival-claim.md` fails on *both* rule 2 and rule 4; a new `072-new-thing.md` with no README row fails rule 5.
- The self-check cannot pass vacuously: every case asserts a **specific substring**, and `pass`/`fail` are separate helpers, so an assertion-free case is not expressible.
- The `CI=true` ref-missing arm is genuinely reachable — the lint job does fetch `refs/remotes/origin/main` (`ci.yml:105`).

Nice detail: the rename exemption degrades to *stricter*, never looser, and says so in the docblock. The "heading-shaped string later in the body" fixture is a real trap correctly closed.

### On what the guard does not cover — and why it's still worth having

It reads **files on disk + the README index table + `origin/main`**, and deliberately not open PRs. So it cannot catch two open PRs claiming one number *while both are open* — it catches the second one built against an updated `origin/main`. The PR is honest about this ("caught at merge time").

The residual hole is one hop further out and is a **branch-protection** question, not a script question: if PR B's lint job was green before PR A merged and B merges without a re-run, the collision still lands. Worth stating in the thread that closing that last gap means requiring branches to be up to date with `main` — but that is not this PR's job and not a reason to hold it.

### SUGGESTIONS (all optional)

- `docs/architecture/adrs/README.md:169` — the retired reserved-numbers note still ends with a hand-maintained **"Allocate 072 for the next new ADR"**, while step 1 above it says "pick one higher than the highest number in the index". Two sources for one fact, and the hardcoded `072` is exactly the value that drifted three times in two days — stale the moment ADR-072 merges. Now that uniqueness is mechanical, drop the number.
- `ci.yml:89–107` — the `Fetch origin/main` step is named and commented for the migration invariant only. A second guard now depends on it; without a comment, the next person tempted to remove it has no signal that two invariants degrade.
- The CI hard-failure arm is only reachable via `runAgainstTree()`; `resolveMissingBaselineAction` is self-checked in isolation but the composed "ref missing in CI ⇒ violation" path is not. Low value, since the helper is shared verbatim with the migration guard.

---

Issue #2085, opened 2026-08-14
[TASK] CORE — ShopCategoryBrowseService delegates to the destination-taxonomy read model


piotrswierzy:
Split out of #2074, which **deferred** this after verification showed it would regress a shipped surface.

## Problem

`ShopCategoryBrowseService` (`libs/core/src/listings/application/services/shop-category-browse.service.ts`) resolves `ProductPublisher`, narrows to `ShopCategoryBrowser`, and calls `browseCategories` **live** on every read. ADR-037's defining property is that reads come from OL's own store and the capability is the refresh path only — so this is the last shop-side read still contradicting the model it now sits beside.

## Why #2074 deferred it

Two things, both verified rather than assumed:

1. **No bootstrap-on-connection-create** (#2084). The only enqueue is the hourly scheduler task, so a freshly created WooCommerce connection has an empty projection. Delegating would turn the instantly-populated shop publish category picker (#1830, shipped) into an empty one for up to an hour.
2. **`TaxonomySourceUnavailableException` has no HTTP mapping**, so the failure mode would change from the documented **422** (`shop-publish.controller.ts:120`) to an unmapped **500**. #2074 maps it at its own controller boundary; this issue needs the same treatment on the shop route.

Neither is hard — they just have to exist first, and shipping the delegation without them trades a working operator surface for a tidier internal shape.

## Proposed Solution

Re-implement `browseCategories` over `IDestinationTaxonomyService.browse`, keeping the service interface and `GET /listings/connections/:id/shop-publish/categories` response byte-identical.

While in there: the service currently throws `UnprocessableEntityException` from **core**, importing `@nestjs/common` — a boundary bend that predates this work. Prefer a domain exception mapped at the controller, matching what #2074 does for the taxonomy routes.

## Classification

**Type**: CORE
**Layer**: Application (+ Interface for the exception mapping)
**File(s)**:
- `libs/core/src/listings/application/services/shop-category-browse.service.ts`
- `apps/api/src/listings/http/shop-publish.controller.ts`

## Dependencies

- **#2084** — bootstrap-on-create. Blocking: without it this is a regression.
- #2074 — establishes the 422 mapping pattern to copy.

## Acceptance Criteria

- [ ] `ShopCategoryBrowseService.browseCategories` reads the projection, never the live adapter
- [ ] `GET …/shop-publish/categories` response shape is unchanged (asserted)
- [ ] A connection whose adapter cannot browse still yields **422**, not 500
[...]

Issue #2210, opened 2026-08-19
[FEATURE] CORE - Erli resolves categories by EAN through a borrowed Allegro connection


norbert-kulus-blockydevs:
Part of #2205.

## Problem / Context

Erli cannot resolve a category from a barcode at all. `ErliOfferManagerAdapter` implements `TaxonomyBorrower` (`getBorrowedTaxonomy() -> 'allegro'`, ADR-025 §3) but not `EanCategoryMatcher`, because it has no catalogue of its own — it accepts Allegro category ids verbatim. `CategoryResolutionService.resolveCategoriesBatch` therefore degrades every variant to `no-match`, and the wizard suppresses the resulting blockers via `destinationResolvesCategoryAtSubmit`; the category is resolved at build time from the operator's PrestaShop→Allegro mappings (#1045).

So an Erli operator gets no automatic category detection even when their catalogue carries barcodes, and the epic's streamed progress would show them an empty stream.

**Erli's own Allegro application credentials cannot close this.** Probed live against the sandbox: an application token reaches `/sale/categories` and friends with 200, but `GET /sale/products?phrase={ean}` answers 403 `AccessDeniedException` despite the token carrying `allegro:api:sale:offers:read`, and `GET /sale/products/{ean}?idType=GTIN` answered 404 `ProductNotFound` for every EAN tried, including nine with live sandbox offers. The catalogue search sits behind seller context. Full codes are recorded in ADR-047.

## Proposed Solution

Resolve the barcode step through a connection that **owns** the borrowed taxonomy.

`CategoryResolutionService` already receives `borrowedTaxonomy` and `sourceConnectionId` and already falls back to owner-provenance mappings for a borrowing destination (#1045). Extend the same principle one step earlier: when the destination borrows taxonomy `T` and cannot match EANs itself, find an enabled connection that declares `TaxonomyIdentityProvider`/owns `T` **and** supports EAN matching, and use that adapter for the lookup.

- Entirely capability-driven, resolved through `IIntegrationsService` — **no cross-plugin import**. The Erli package is not touched, which matters because it is deliberately independent of the Allegro package (ADR-003, restated in `allegro-category-catalog-client.ts`'s header).
- **Optional, never required.** An operator with no Allegro connection keeps today's behaviour exactly. This is not the "require an Allegro seller connection" option ADR-031 rejected — that was about making it a precondition for browsing a public tree.
[...]

Issue #2074, opened 2026-08-14
[TASK] Interface — Wave 2a: repoint the marketplace category reads at the destination-taxonomy read model


piotrswierzy:
Wave 2 of epic #1937. Wave 1 (#1979) and its follow-ups #2063 / #2061 are all merged, so this is unblocked.

**Split from the epic's single "Wave 2" bullet** — see § Scope correction. This issue is the **backend half**; the frontend half is #2075.

> **Scope amended after the pre-implement gate + tech review** (see § Amended scope at the bottom): `ShopCategoryBrowseService` delegation moved to **#2085**, gated on **#2084**.

## Problem / Context

`IDestinationTaxonomyService` (`libs/core/src/listings/application/interfaces/destination-taxonomy.service.interface.ts`) now offers `browse` + `search` over the `destination_categories` projection, but **nothing reads it**. Every taxonomy consumer still goes through the platform-named `CategoriesCacheService` in the host app:

| Route (`apps/api/src/mappings/http/mapping-options.controller.ts`) | Today | Notes |
|---|---|---|
| `GET …/options/source/categories` (`:183`) | `categoriesCacheService.getAllegroCategories()` | Marketplace tree via `CategoryBrowser`, DB-cached in `allegro_category_cache`, 24 h TTL |
| `GET …/options/source/categories/:categoryId/path` (`:215`) | `getAllegroCategoryPath()` | **Live** adapter call on a read path |
| `GET …/options/destination/categories` (`:183`) | `getPrestashopCategories()` | **`ProductMaster.getCategories()` — live, uncached** |

There is also still **no neutral marketplace tree-browse route**.

⚠️ **Vocabulary trap**: in this controller **`source` is the marketplace** and **`destination` is the master shop** — the opposite of ADR-023's mapping direction. The routes are keyed on the marketplace connection (#1784), so a repoint must not "helpfully" swap them.

## Scope correction — `destination/categories` stays put

**Only the marketplace half maps onto the Wave-1 model.** `GET destination/categories` resolves the **`ProductMaster`** capability (`categories-cache.service.ts:118`), but `DestinationTaxonomyService.resolveDestination` probes **`OfferManager` then `ProductPublisher`** and never `ProductMaster`. A master-catalog tree is a **third** taxonomy kind the projection does not model, and its `PrestashopCategoryDto` carries `depth` / `active`, which `DestinationCategory` has no column for — so "byte-identical" is unachievable there without extending the model.

[...]

PR #3038, opened 2026-09-09
fix(listings): taxonomy reads report connection-not-found/disabled as 404/409, not 422

### LINKED ISSUE #2146: [TECH-DEBT] CORE — every taxonomy read reports a missing/disabled connection as 422, losing 404 and 409
## Problem / Context

`DestinationTaxonomyService.resolveDestination` discovers a connection's destination kind by probing two capabilities through `tryGetAdapter` (`libs/core/src/listings/application/services/destination-taxonomy.service.ts:376`):

```ts
private async tryGetAdapter<T>(connectionId: string, capability: string): Promise<T | null> {
  try {
    return await this.integrationsService.getCapabilityAdapter<T>(connectionId, capability);
  } catch {
    // A destination legitimately supports only one of the two kinds; probing
    // is how the kind is discovered, so an unsupported capability is expected
    // rather than exceptional.
    return null;
  }
}
```

The bare `catch` is correct for its stated purpose — `CapabilityNotSupportedException` really is expected during a kind probe — but it also swallows `ConnectionNotFoundException` and `ConnectionDisabledException`. Both probes then return `null` and `resolveDestination` falls through to `throw new TaxonomySourceUnavailableException(...)`.

**Net effect: on every read that funnels through `resolveScope`, three distinguishable conditions collapse into one 422.**

| Condition | Correct status | Actual |
|---|---|---|
| Connection does not exist | 404 | 422 |
| Connection disabled | 409 | 422 |
| No category-browse capability | 422 | 422 |

The 422 message also names the wrong cause — it reports "no CategoryBrowser, TaxonomyBorrower, TaxonomyIdentityProvider, or ShopCategoryBrowser capability" for a connection id that was simply mistyped, which sends the reader to the adapter instead of to the id.

Affected reads: `browse`, `search`, `path`, and therefore `GET /listings/connections/:id/taxonomy/categories[/search]` (#2074), `mapping-options`' repointed category routes (#2074), and `GET /listings/connections/:id/shop-publish/categories` (#2085).

Found while reviewing #2145. That PR corrected the **documentation** — the affected routes previously advertised 404 and 409 responses they cannot emit — but deliberately did not change the behaviour, because the fix belongs in core's shared resolution path rather than in one delegating service.

## Proposed Solution

Let `tryGetAdapter` swallow only what the probe is actually about, and rethrow connection-level failures:

- Catch `CapabilityNotSupportedException` (and `CapabilityNotEnabledException`) → `return null`, which is the probe's real intent.
- Rethrow `ConnectionNotFoundException` / `ConnectionDisabledException` so the existing `ConnectionExceptionFilter` maps them to 404 / 409 as it does everywhere else.

Worth confirming while implementing: whether a **disabled** connection should still be allowed to *read* its already-synced projection rows. There is a reasonable argument that it should — the rows exist, reading them is free, and refusing is only meaningful for the refresh path (the same reasoning already applied to a borrowing connection with no catalogue credentials, see the `browse` deferral comment in `resolveDestination`). If so, the disabled case wants an explicit allow rather than a rethrow, and that is a deliberate decision to record rather than infer.

Restore the `@ApiResponse` 404/409 declarations removed in #2145 once the behaviour matches them again.

## Classification

**Type**: CORE
**Layer**: Application
**File(s)**:
- `libs/core/src/listings/application/services/destination-taxonomy.service.ts`
- `apps/api/src/listings/http/taxonomy.controller.ts`
- `apps/api/src/listings/http/shop-publish.controller.ts`

## Dependencies

- #2145 (merged first — it documents the current behaviour and is what surfaced this)

## Assumptions

- `ConnectionExceptionFilter` already maps the two domain exceptions to 404/409, so no new filter is needed.

## Acceptance Criteria

- [ ] A taxonomy read for a nonexistent connection id returns **404**, not 422
- [ ] A taxonomy read for a disabled connection returns **409**, or reads the projection successfully if the "disabled connections may still read" decision is taken — whichever is chosen is recorded in the code
- [ ] A connection with no browse capability still returns **422** with the capability-naming message
- [ ] The `@ApiResponse` declarations on both controllers match the restored behaviour
- [ ] Tests added covering all three conditions distinctly

jakubretajczykBD:
## Summary
- `DestinationTaxonomyService.tryGetAdapter` used a bare `catch {}` while probing a connection's destination kind (`OfferManager` then `ProductPublisher`), which swallowed `ConnectionNotFoundException` / `ConnectionDisabledException` alongside the capability-shaped exceptions it was actually meant to probe for — every taxonomy read collapsed a nonexistent/disabled connection and a genuinely capability-less one into the same 422.
- Narrowed the catch to only swallow `CapabilityNotSupportedException` (covers `CapabilityNotEnabledException` via inheritance) and rethrow everything else, so the already-registered global `ConnectionExceptionFilter` maps them to 404/409.
- Resolved the issue's own open question: a **disabled** connection does not get a read-only exception here — decision recorded inline, distinguished from the unrelated borrower-with-no-credentials capability-shaped deferral a few lines down.
- Restored the 404/409 `@ApiResponse` declarations on `taxonomy.controller.ts` and `shop-publish.controller.ts` that #2085 had removed when this bug made them unreachable.
- Fixed the test double (`destination-taxonomy.service.spec.ts`) to reject with the real `CapabilityNotSupportedException` instead of a generic `Error` — under the narrowed catch a generic `Error` would have propagated instead of being swallowed, breaking every existing "falls through to the second probe" test.
- Added 3 new unit tests covering `ConnectionNotFoundException` propagation, `ConnectionDisabledException` propagation, and that the second probe is never attempted once a connection-level failure is seen.
- Implementation plan committed at `docs/plans/implementation-plan-taxonomy-connection-status-codes.md`.

## Test plan
- [x] `destination-taxonomy.service.spec.ts` — 31/31 unit tests pass (3 new)
- [x] `eslint` clean on all touched files
- [x] `scripts/check-cross-context-imports.mjs` passes (new exception imports are an already-allowed `listings --> identifier-mapping` edge)
- [x] `tsc --noEmit` on `libs/core` shows no new errors
- [ ] Full `pnpm test` / `pnpm test:integration` — left to CI

Closes #2146


norbert-kulus-blockydevs:
## PR Review — #3038 (senior backend architect pass)

**Scope**: narrows `DestinationTaxonomyService.tryGetAdapter`'s bare `catch {}` (which swallowed `ConnectionNotFoundException` / `ConnectionDisabledException` alongside the capability-shaped exceptions it was meant to probe for) into `instanceof CapabilityNotSupportedException` → swallow, else rethrow. Restores/adds `@ApiResponse` 404/409 on `taxonomy.controller.ts`, `shop-publish.controller.ts`, and `mapping-options.controller.ts`.

### Correctness — verified against live code ✅
- **Exception hierarchy confirmed**: `CapabilityNotEnabledException extends CapabilityNotSupportedException` (checked the source), so the single `instanceof` check genuinely covers both without needing an explicit `||`.
- **Throw ordering confirmed**: `IntegrationsService.getAdapter` throws `ConnectionDisabledException` *before* `getCapabilityAdapter`'s capability checks even run — so the plan's stated edge case ("a disabled connection with no capability at all still reports 409, not 422") is real, not hopeful.
- **`ConnectionExceptionFilter` mapping confirmed**: `ConnectionNotFoundException → 404`, `ConnectionDisabledException → 409`, registered globally — no per-controller wiring needed, and the fix is exactly "let the caller's own filter chain do its job" rather than a local try/catch (consistent with the pattern this repo's docs call out as correct).
- **Borrower deferral (#2210) unaffected**: `marketplaceBrowseFn`'s capability-shaped "no `EanCategoryMatcher`" deferral operates downstream of the resolved adapter, not via a second `tryGetAdapter` probe against a peer/owner connection — so narrowing the catch here doesn't touch that separate mechanism, contrary to what I went in expecting to find a regression in.
- **No frontend regression found**: grepped `apps/web` for anything branching specifically on "422 = connection problem" for these three routes — found none. The one place that inspects 422 near taxonomy (`use-category-path-query.ts`) already treats 404 and 422 identically (`retry: false`), so moving some cases from 422→404/409 doesn't change frontend retry behavior.

### Gap: no HTTP-layer integration test for the actual regression
The `#2146` bug was specifically that the wrong status code reached the wire — yet the only new tests are at the service layer (asserting `resolveScope()` rejects with the right exception *type*). The existing `apps/api/test/integration/destination-taxonomy.int-spec.ts` is untouched and has zero 404/409/422 assertions. Nothing in this PR proves the full chain (route → service throw → `ConnectionExceptionFilter` → actual response status/body) for these specific routes. The filter itself is unit-tested elsewhere, and NestJS's decorator wiring is low-risk, but "low risk" isn't "proven" for the exact bug being fixed — a single `supertest` case per route (nonexistent id → 404) would close this cheaply. **Not blocking**, but worth adding.

### Minor: implementation-plan doc contradicts the code it describes
`docs/plans/implementation-plan-taxonomy-connection-status-codes.md` (committed in this PR) explicitly states in **§ Out of Scope**: *"`mapping-options.controller.ts`'s ... routes ... Left untouched"* — but the actual diff (and the final commit message, `"...restore 404/409 docs on mapping-options taxonomy routes"`) shows `@ApiResponse` decorators genuinely added to that file. The plan was written before that decision changed and never updated. Harmless since the code is correct, but a plan doc that asserts the opposite of what shipped is confusing for the next reader who trusts it as documentation of intent.

Core fix is correct, well-reasoned (the disabled-connection decision is explicitly recorded rather than left implicit, per the issue's own ask), and I couldn't find a regression in the borrower path or the frontend after specifically looking for one. Two non-blocking asks: add an HTTP-level int-spec case per route, and fix the stale "out of scope" claim in the committed plan doc.

piotrswierzy (commented):

A correct, well-reasoned fix. All three design questions check out.

**Is splitting 422 into 404/409 correct?** Yes. The three states are genuinely distinct — "no such connection", "disabled", "active but exposes no browser" — and the swallowing catch collapsed all three. Note the #2085 comment being replaced was documenting a *bug*, not a decision, and the PR says so accurately.

**Is the global-filter placement preserved?** Yes, and this is the important structural point: **no per-controller catch is introduced anywhere.** The fix is entirely in the core service's catch narrowing; the exceptions travel to the already-registered global `ConnectionExceptionFilter` (`main.ts:84` → `global-filters.ts:35-42`). `TaxonomySourceUnavailableFilter` stays in place for the genuine capability-shaped 422. #2074's rationale (a local catch would leave the mappings routes at 500) is respected — and strengthened, since this PR declares the codes on those routes too.

**Does anything depend on the 422?** **No consumer anywhere**, checked in both directions:
- `apps/web` — no status-code branching in any consumer. `use-category-search.ts`, `use-shop-categories-query.ts` and `mappings.api.ts` are plain queries; `isTaxonomyUnsynced` (`category-search-hits.ts:40-51`) is deliberately status-agnostic (`browseError == null`), so any error falls through identically. Every `status === 422` site in the app belongs to unrelated features.
- `apps/api/src/mcp` — no `422` literal at all; `destination-context.ts` probes by capability and lets the domain error surface.
- No int-spec hits these routes.
- **The only real dependency points the other way**: `apps/e2e/src/api/api-client.ts:537-551` treats **404** as the unknown-connection signal, so this makes that helper *more* correct.

Also verified: `CapabilityNotEnabledException extends CapabilityNotSupportedException`, so the single `instanceof` genuinely covers both and the comment is accurate. The borrower-with-no-credentials deferral is a separate `isCategoryBrowser` narrowing path, untouched. The spec fix (generic `Error` → real `CapabilityNotSupportedException`) was mandatory under the narrowed catch and is correctly identified. The "second probe never runs after a connection-level failure" test is the right assertion to add.

### IMPORTANT — an unenumerated third consequence: adapter-construction failures now reach an unmapped 500

`destination-taxonomy.service.ts` (~400, the `throw error` arm)

The catch narrows to `CapabilityNotSupportedException` and rethrows *everything else* — including adapter-construction failures, which the PR description does not enumerate. `AllegroAdapterFactory` throws `AllegroConfigException`, a plain `Error` with **no HTTP filter** in `apps/api` (`buildGlobalExceptionFilters` carries seven; none catches it). So a misconfigured-but-active Allegro connection with `OfferManager` enabled now answers **500** on `GET /listings/connections/:id/taxonomy/categories` and on the two repointed `mapping-options` routes, where before it answered a 422 an operator could act on.

That is the *exact* failure mode the #2085 comment this PR replaces was guarding against — "a plain connection-configuration state" answering 500. The loud-over-silent argument is defensible, but the choice is currently unstated and unfiltered. Either keep the rethrow and add a filter for adapter-construction failures, or narrow it to `ConnectionNotFoundException | ConnectionDisabledException` and let anything else keep falling through to 422. If neither, say so in the comment — it currently claims "surfaces loudly" without noting that "loudly" means an unmapped 500.

### SUGGESTIONS

- Adding the `mapping-options.controller.ts` declarations (~226–235, ~263–272) is a good catch — those routes share the fixed `resolveScope` chain and were genuinely under-declared. Worth noting in the body that this file was **added beyond #2146's scope**, since the description says the fix was scoped to the two listings controllers.
- No integration coverage. The behaviour is proved only at the service level (three good new unit tests); **no int-spec touches any of these four routes at all**. The unit tests pin the rethrow; nothing pins that the filter is registered and reached. One int-spec asserting `404` for an unknown connection id would make the restored `@ApiResponse` declarations true rather than asserted.

---

piotrswierzy (commented):

The prior IMPORTANT is addressed — the adapter-construction-failure arm is no longer an unmapped 500 reaching an operator on a route they load. The design questions all still check out: splitting 422 into 404/409 is correct, **no per-controller catch was introduced** (the fix stays in the core service's catch narrowing, so exceptions travel to the already-registered global `ConnectionExceptionFilter`), and no client anywhere depends on the old 422 — while `apps/e2e/src/api/api-client.ts:537-551` already treats **404** as the unknown-connection signal, so this makes that helper more correct.

Remaining, all small:

- **Four `@ApiResponse` 422 descriptions** and one exception message still describe the pre-split behaviour. Since the restored codes are the contract a client reads, the descriptions should name what each status now means.
- **Still no integration coverage.** The behaviour is proved at the service level only, and **no int-spec touches any of these four routes at all** — so the restored `@ApiResponse` codes are declared rather than proved, and nothing pins that the global filter is registered and reached. One int-spec asserting 404 on an unknown connection id would close it.
- Worth noting in the body that `mapping-options.controller.ts` was added **beyond #2146's scope** — those routes share the fixed `resolveScope` chain and were genuinely under-declared, but a reviewer skimming the file list would otherwise wonder why `mappings` is here.

---

piotrswierzy (approved):
## Tech Lead review (re-review)

No new commits since the last pass (head is still `bfeeb1e0`), so this is a verification round rather than a fresh read — and it closes the two findings I had left open, both of which turn out to be already fixed at this commit.

**Stale `@ApiResponse` descriptions — resolved.** All four 422 descriptions now read *"the connection exists and is active, but exposes no category browser"*, i.e. they describe the post-split contract rather than the collapsed one, and the two exception messages (`destination-taxonomy.service.ts:315`, `:335`) were already capability-shaped and correct — neither ever claimed "does not exist or is disabled", so there is nothing left to reword there.

**The structural points still hold, re-verified at this commit.** The fix stays entirely inside the core service's catch narrowing — **no per-controller catch anywhere** — so exceptions travel to the globally-registered `ConnectionExceptionFilter`, and #2074's reason for keeping that filter global is respected and strengthened by the restored declarations. `CapabilityNotEnabledException extends CapabilityNotSupportedException`, so the single `instanceof` genuinely covers both. And the adapter-construction arm is now explicitly swallowed back into the 422 rather than rethrown into an unmapped 500, which was the prior IMPORTANT.

### Findings

**[SUGGESTION]** — `libs/core/src/listings/application/services/destination-taxonomy.service.ts` (the final `return null` arm of `tryGetAdapter`)
> The catch-all arm is now a deliberate, documented decision rather than an accident — but it still swallows with **no log line**. An `AllegroConfigException` from a malformed `Connection.config` becomes a 422 whose body tells the operator the connection "exposes no category browser", which is false about their actual problem, and nothing anywhere records what really happened. Not a regression (pre-#2146 behaviour was identical), but this PR is where that arm becomes intentional, and the class already holds a `Logger` used at `:365` and `:373` for exactly this shape of degraded resolution. One `this.logger.warn` naming the connection, the capability and the error turns an invisible misconfiguration into a greppable one at no behavioural cost.

**[SUGGESTION]** — integration coverage
> Still none, and I re-checked in both directions: no file under `apps/api/test` or `apps/worker/test` references `taxonomy/categories`, `shop-publish/categories` or `source/categories` — the only hit in the repo is `apps/e2e/src/api/api-client.ts`. So the four restored `@ApiResponse` codes are declared rather than proved, and nothing pins that the global filter is registered and actually reached on these routes. The repo's own standard is that integration tests cover critical vertical slices only, so this is not a merge gate — but one int-spec asserting 404 on an unknown connection id would convert the whole PR's claim from "the service throws the right exception" to "the route answers the right status".

**[SUGGESTION]** — PR body
> `mapping-options.controller.ts` is edited beyond #2146's stated file list. The reasoning is sound and is recorded in the committed plan (those routes share the fixed `resolveScope` chain and were genuinely under-declared — they carried no 422 at all), but the PR description doesn't mention it, so a reviewer skimming the file list wonders why `mappings` is in a `listings` fix. One line.

### Verdict
✅ **Approve**

`mergeStateStatus` is `BEHIND` — update the branch before merging so the green checks are against current `main`.

---

Issue #3055, opened 2026-09-10
[EPIC] Frontend — Sourcing rules admin UI for the fulfilment router


norbert-kulus-blockydevs:
## Problem / Context

The OMS sourcing-rules admin API (`apps/api/src/oms/http/oms-sourcing-rules.controller.ts`, 6 routes: list/create/reorder/get/update/delete) is fully shipped and tested, but `apps/web` has zero callers — confirmed by the #3029 route audit (`grep -rn "sourcing-rules" apps/web/src` returns nothing). An operator can only author the ordered filter/sort ruleset that decides which warehouse fulfils an order via `curl`.

## Mockup

**Durable spec (cite this, not the Artifact): `docs/plans/mockups/sourcing-rules-3055.html`**

Live preview, current version: https://claude.ai/code/artifact/2a03d1ee-646b-4cff-b7a9-1187f00d12c3

The URL this section used to carry (`b7b9ed28-…`) was the *first* reviewed draft and is superseded — the file has since roughly doubled and been through two `/pr-review` rounds on #3119. Per `docs/frontend-architecture.md` § UX Mockups, an implementation ticket cites the **file path** and the `data-state` value(s) it implements, and the E2E baseline comes from the repo copy, never from the Artifact URL (the Artifact can be edited or expire independently of the repo).

### `data-state` values

Every state is stamped on `body[data-state]` and has a switcher button in-page, so a sub-issue or E2E spec names the value instead of describing "the screen where X happens".

**View states** (toolbar row 1) — `populated`, `no-locations`, `empty`, `loading`, `error`

**Dialog states** (toolbar row 2) — `rule-create`, `rule-edit` (kind radios locked), `confirm-loosen`, `confirm-delete`, `rule-unrecognised-locked`, `dry-run`

**Pinned states** (gallery; each card's label is its own switcher button) — `gallery-no-locations`, `gallery-empty`, `gallery-loading`, `gallery-error`, `gallery-wrong-connection`, `gallery-validation-priority-misuse`, `gallery-name-claimed`, `gallery-priority-stale`, `gallery-delete-governing`, `gallery-scheduled`, `gallery-retired`, `gallery-unrecognised`, `gallery-active`, `gallery-governing`, `gallery-reorder-saved`, `gallery-reorder-conflict`, `gallery-settings-tile`

### Decisions (settled 2026-09-14 - the mockup wins on both)

[...]

Issue #3063, opened 2026-09-10
[EPIC] Frontend — Inventory locations CRUD UI


norbert-kulus-blockydevs:
## Problem / Context

`inventory_locations` (#2313/#2316) has a complete CRUD API (`apps/api/src/inventory/http/inventory-locations.controller.ts`), but `apps/web/src/features/inventory/api/inventory.api.ts` only calls `listActiveLocations` (a degenerate `limit=1` probe) and `bootstrapLocations` — confirmed by #3029. Full list, single-item read, create, update and delete are all unreachable from the product.

## Mockup

Reviewed, interactive HTML mockup — built on the real design system, all states: https://claude.ai/code/artifact/fbb5f9e1-52a3-4b7b-8136-ad59dcf4f318

## Proposed Solution

Extend `features/inventory` with the remaining 4 CRUD routes, a new `inventory-locations.route.tsx` + `pages/inventory/inventory-locations-page.tsx` (no `pages/inventory` directory exists today), and a nav entry, following the `connections.route.tsx` / `connections-list-page.tsx` pattern.

Broken down into the following sub-issues (linked as GitHub sub-issues):

- extend locations API client (list/get/create/update/delete)
- locations query hooks + filters (kind, status, codePrefix)
- locations list page with filter bar
- create/edit location dialog
- delete confirm incl. "location in use" 409 + retire action
- nav entry for Inventory locations
- locations CRUD tests

## Classification

**Type**: Frontend
**Layer**: Interface (feature module + route)

## Dependencies

- None — the backend routes this epic wires up are already shipped (see #3029)

## Assumptions

- Backend contract is stable as audited in #3029; no API changes needed
- Follows `apps/web/src/features/*` conventions per `docs/frontend-architecture.md`

## Acceptance Criteria

- [ ] Every sub-issue below is completed and merged
- [ ] The surface is reachable from the product (nav entry, route, or mount point)
- [ ] Manual verification against a running `pnpm start:dev:api` + `pnpm start:dev:web`

Issue #2507, opened 2026-08-26
[TASK] DX — style guide: the document cell, the new primitives, and colour marks exceptions


norbert-kulus-blockydevs:
## Problem / Context

`docs/frontend-ui-style-guide.md` requires a written carve-out for any row height not on its density list, and states that colour is never the only signal — but it has no rule about *when* colour should be used at all. The redesign found the real defect: tinting every row makes tint meaningless, which is a different failure from colour-alone.

## Proposed Solution

Add three entries: the sales-document cell (one line, glyph plus word, popover for detail) as a registered density case; the new primitives (`DocumentKindGlyph`, `DocumentHeadline`, `DocumentLifecycle`) in the primitives catalogue; and a **colour marks exceptions** rule stating that a finished or normal state is plain ink and only states needing action are tinted, with the tick-and-dot shape distinction that keeps it readable without colour.

## Classification

**Type**: DX
**Layer**: Shared
**File(s)**: `docs/frontend-ui-style-guide.md`

## Dependencies

- Parent mini-epic: #2505
- Docs phase 1 (#2500) merged.

## Acceptance Criteria

- [ ] The cell is a registered entry naming the facts that justify its shape.
- [ ] The three primitives are listed with their purpose and their token dependencies.
- [ ] The colour rule is stated with its rationale and its interaction with the existing colour-alone rule.
- [ ] The rule names the popover-escapes-overflow requirement for a cell popover inside `DataTable`.
- [ ] Tests added or updated for non-trivial logic
- [ ] No architecture boundary violations (CORE ↔ Integration)

## Branching and merge target

Branch: `{issue#}-{short-kebab}`, created **from its mini-epic branch**.
PR target: **its mini-epic branch**.

## How this issue gets finished

1. Implement on the branch above.
2. Open the PR into its mini-epic branch.
3. Run `/pr-review <PR>` and post the verdict as a PR comment.
4. Fix **every** finding, and reply under the comment saying what changed for each one.
5. Run `/pr-review` again. Repeat until the review comes back approved with no findings.
6. Only then merge.

Quality gate before every commit: `pnpm lint`, `pnpm type-check`, `pnpm test`.
Commits are signed off (`git commit -s`).

PR #3119, opened 2026-09-10
docs(mockups): sourcing-rules UI mockup

### LINKED ISSUE #3112: design(oms): sourcing-rules UI mockup
## Problem / Context

Before implementation starts, Sourcing rules admin UI (#3055) needs a reviewed design so every sub-issue builds against

## Proposed Solution

## Classification

**Type**: Frontend
**File(s)**: `docs/plans/mockups/sourcing-rules-3055.html`

## Dependencies

- None — first sub-issue of #3055, everything else in the epic reads from this mockup

## Assumptions

## Acceptance Criteria

- [ ] `docs/plans/mockups/sourcing-rules-3055.html` exists and renders standalone (no external API calls)
- [ ] PR links back to the reviewed Artifact: https://claude.ai/code/artifact/b7b9ed28-9062-4985-8f0e-353b6fa1fcc5

norbert-kulus-blockydevs:
## Summary
- Adds the `sourcing-rules` mockup as a static file at `docs/plans/mockups/sourcing-rules-3055.html`
- **Current Artifact (live preview): https://claude.ai/code/artifact/2a03d1ee-646b-4cff-b7a9-1187f00d12c3**
- Original reviewed Artifact (superseded): https://claude.ai/code/artifact/b7b9ed28-9062-4985-8f0e-353b6fa1fcc5
- Closes #3112, first sub-issue of #3055

The mockup went through 5 rounds of redesign since the original submission — see the PR comments below for the full breakdown of each round. Summary: fixed the reviewed color/UX issues, rewrote all copy into plain operator language, added a real tooltip/popover system, fixed several confirmed functional and accessibility bugs (found via rendered-screenshot and driven-interaction verification, not just code reading), and closed with two completeness reviews against real OMS domain requirements and competitor tools — including one confirmed live defect (`country-served` silently breaking every cross-border order for the single-hub-many-countries topology).

## Test plan
- [ ] Walk through the "Component states, for review" gallery at the bottom — every reachable state is pinned there
- [ ] Try the create/edit dialog, the priority-location builder, and "Test against an order"


norbert-kulus-blockydevs:
*Extended review — the pass below folds in three additional lenses (OpenLinker domain/architecture, e-commerce operations, senior backend/system architecture) on top of the original frontend/UX pass.*

## PR Review — #3119 (sourcing-rules UI mockup)

**Scope reviewed**: `docs/plans/mockups/sourcing-rules-3055.html` (single static file, ~1000 lines), cross-checked against `libs/oms/src/routing/*`, `apps/api/src/oms/http/oms-sourcing-rules.controller.ts`, and this repo's domain docs (docs/architecture-overview.md § 26 Fulfillment, ADR-052/053/054/062, docs/frontend-ui-style-guide.md).

### What's unusually good
This mockup is exceptionally well-grounded in the real, merged backend — not just the issue title. Confirmed against actual source:
- The closed filter/sort vocabulary (`in-stock`, `country-served`, `not-blocked-by-reject`, `priority`, `nearest`, `most-complete`, `least-splits`) matches `RoutingFilterNameValues`/`RoutingSortNameValues` exactly.
- The "after-action ladder" (No split / Line split / Quantity split, most-restrictive-wins) matches `RoutingAfterActionValues` + `mostRestrictiveAfterAction`, including the docblock's own framing.
- "Unrecognised — locked, delete-only" matches the controller's documented behavior precisely: a row with `recognised: false` is stored and listed, never hidden, and PATCH refuses it — "delete it and create a replacement" is the real contract, worded almost identically.
- "Delete = hard delete + history loss" vs. "Retire = PATCH `effectiveTo` into the past" matches the DELETE/PATCH docblocks' intent word-for-word.
- The reorder contract — dropping a rule shifts everything below it up one position, and a `PUT /order` naming only a subset is refused (409) — matches the real endpoint.
- **The "Wrong connection — 400 on save" gallery state is a genuinely sharp domain touch.** It correctly encodes that sourcing rules are evaluated only for the OMS connection (ADR-052/053: A2 `sourcing` is `config-only`), and there's no spurious capability checkbox anywhere implying `FulfillmentRouter` is a generic connection setting — it correctly lives only inside the OMS's own rule-authoring surface, matching `FulfillmentRouter`'s deliberate absence from `CoreCapabilityValues`/manifests.
- The "Test against an order" button is correctly framed as a non-committing dry run — i.e. it correctly reaches for the backend's real `evaluate()` (non-committing) rather than `route()` (committing), which is the right distinction to draw even though the interaction itself isn't built out (see below).

### Findings

**1. (Important) "Active" status contradicts the shipped color-usage rule (#2507), on the exact case that rule was written to prevent.**
`docs/frontend-ui-style-guide.md`: *"a finished or expected state is plain ink; only a state someone has to act on is tinted"* — citing *"a sales-document column that tinted every row green on a healthy list"* as the anti-pattern it exists to kill. This mockup does exactly that: every healthy, nothing-to-do rule renders a green `status-badge--success` "Active" badge (`ruleStatus()`, 3 of 5 seed rows). On a populated ruleset this is a wall of green. Per the documented pattern, `Active` should be plain ink + a small tick; reserve colored badges for `Unrecognised` (warning — something to act on) and arguably `Scheduled`/`Retired` as muted/neutral.

**2. (Important) The two places where real domain complexity lives are both stubbed, and it's the same underlying gap twice.**
- The priority-list authoring UI (`kind=sort`, `name=priority`) renders two hardcoded, non-interactive chips (`Warsaw — Main warehouse`, `Kraków — Overflow`) with no add/remove/drag/reorder affordance, and `saveRule()` confirms this isn't just a rendering shortcut — it literally hardcodes `priorityLocations: ['Warsaw — Main warehouse', 'Kraków — Overflow']` regardless of what the operator does. But `priorityLocationIds: string[]` is exactly what the real `POST`/`PATCH` body needs, and it's an array the operator has to construct.
- "Test against an order" is a bare `alert('...Not shown in this mockup.')` — there's no drawn interaction for the ONE feature (`evaluate()`) that would let an ops manager sanity-check a rule against a real order before it goes live, which matters a lot given routing errors are expensive.

Both are the same underlying pattern: the CRUD shell (list, create, edit, delete, reorder) is fully fleshed out and faithful, but the two places requiring real interaction design — constructing a location-priority list, and previewing a dry run — are placeholders. Worth at minimum a `gallery-card` state showing what the location picker and the dry-run result panel look like, even if the main flow stays a stub for now.

**3. (Important, e-commerce ops lens) No audit trail for a program-critical routing decision.**
Nowhere does the mockup show who last changed a rule, or when — not on the row, not in the edit dialog, not anywhere. On a mature OMS/WMS surface (Linnworks, NetSuite-style fulfillment configuration), a rule change history is standard precisely because a misconfigured sourcing rule silently routes real orders to the wrong warehouse, and "who changed rule #3 last Tuesday" is the first question support asks when investigating a misroute. Worth at least a "last edited by X on Y" line per row.

**4. (Minor, e-commerce ops lens) No path drawn for managing a large ruleset.**
The mockup only ever shows 5 rules. At the scale this feature targets (multi-warehouse + 3PL + stores), an operator with 30–50 rules has no search/filter, and — because a ruleset resolves to the *most restrictive* after-action across every active rule (a genuinely non-obvious derived interaction, per the ladder panel's own framing) — there's also no way to ask "which rule is actually deciding this order's fate" once the list gets long. Not blocking for a first cut, but worth flagging before this becomes the literal build reference.

**5. (Minor, backend-architecture lens) The reorder interaction has no visible failure/rollback treatment.**
`onDrop`/`moveRule` mutate the local array and re-render unconditionally — reasonable for a static mock, but worth calling out explicitly for whoever builds the real page: once reordering is wired to the real `PUT /order` call, a drag needs an optimistic-update-with-rollback treatment (the backend's own 409 contract already forces the client to send the *whole* reordered array atomically, which is the safe shape — the gap is purely UI-side: nothing here shows what happens on the screen if that PUT fails mid-drag). Similarly, `saveRule()`'s "Save rule" button has no disabled/loading state during submission — worth a note so the real implementation doesn't ship a double-POST-on-double-click regression.

**6. (Minor, praise) Keyboard-accessible reorder is a nice touch** — the drag handle has a focus/hover-revealed up/down nudge (`.rule-position__nudge`) as a keyboard alternative, since raw HTML5 drag-and-drop has no keyboard story on its own.

**7. (Minor) `country-served` copy is inconsistent between two places.**
The backend docblock is explicit that this filter is *"narrower than the name suggests: `InventoryLocation.countryIso2` records where the location IS, not a list of countries it ships to."* The dialog hint (`NAME_HINTS['country-served']`) correctly says "where it sits — not a list of countries it ships to" ✅, but the table row's static seed description ("The location is in the country the order is delivered to") states the *stronger, incorrect* claim. Align both to the correct meaning before this becomes the implementation reference.

---

Nothing here blocks using this as a build reference — items 1–3 are worth fixing before an engineer starts wiring the real page off it; items 4–7 are lower-priority polish/forward-guidance.

piotrswierzy (changes requested):

Factual accuracy against `libs/oms/src/routing/` is, for the most part, unusually good. Verified correct: both closed vocabularies exactly (`in-stock`/`country-served`/`not-blocked-by-reject`; `priority`/`nearest`/`most-complete`/`least-splits`), `RoutingAfterActionValues`, the most-restrictive-after-action rule (matching `CreateSourcingRuleDto`'s own wording), `priorityLocationIds` read only by `priority`, `country-served` correctly described as *where the location sits* rather than a ship-to list (the weaker claim the vocabulary docblock insists on), `nearest` labelled a proximity proxy, hard-delete-vs-retire, and the non-`openlinker` 400 with copy close to `SourcingRuleAdminService.assertOmsConnection`'s actual message. **The unrecognised-rule treatment is exactly right** — listed, edit refused, delete offered, with the reason given as *"a successful save would imply it routes again"*, which is `SourcingRuleResponseDto.recognised`'s docblock in operator language.

### BLOCKING — the edit dialog exposes `kind` as editable; PATCH has no `kind` field

Line 475. `UpdateSourcingRuleDto` carries `position`, `name`, `afterAction`, `priorityLocationIds`, `effectiveFrom`, `effectiveTo` — and deliberately **not** `kind`. `SourcingRuleAdminService.updateRule` says why in place: *"a `name` change is only legal against the row's own `kind`, which the patch does not carry (kind is half the identity under the unique index)."* `openEdit()` nonetheless checks the matching radio and wires `onKindChange`, so an operator can flip Filter→Sort and the form re-selects a name from the other vocabulary. Whatever the implementer sends is dropped or 400s. Render `kind` read-only on edit, or say "delete and recreate" — the same instruction the unrecognised path already gives.

### BLOCKING — `priorityLocationIds` is the one genuinely configurable rule payload, and it isn't designed

Line 506. `#priorityListField` renders two hardcoded chips with `cursor:default`, and `saveRule()` writes a literal `['Warsaw — Main warehouse', 'Kraków — Overflow']`. No add/remove/reorder, no location picker. `assertLocationsExist` validates these ids server-side and the mockup even has a gallery card for the unknown-location refusal — **so the refusal is designed while the input that produces it is not.** Every implementation sub-issue for this dialog will have to invent the control, which is what #3112's own Problem statement says the mockup exists to prevent.

### IMPORTANT — retired rules are drawn as a row state but are unreachable

`ListSourcingRulesQueryDto.includeSuperseded` defaults `false`, and `listRules` adds `(effectiveTo IS NULL OR effectiveTo > :now)` unless set. So a Retired rule is **not in the default list**. The mockup presents *"Retired — window closed / Kept for history rather than deleted"* as first-class and offers no toggle. Sibling #3120 got this right with an explicit `show retired` checkbox.

### IMPORTANT — neither shipped 409 is depicted

`routing-rule-admin.errors.ts` declares `DuplicateLiveRoutingRuleError` and `RoutingRuleReorderMismatchError`.
- **Duplicate live (kind, name)** — the create dialog lets you pick a name that already has a live rule, with no refusal state. `UpdateSourcingRuleDto.effectiveTo` also documents that *clearing* a retirement revives the rule and *"is refused (409) if a live sibling already claims the same kind and name"* — the edit dialog offers exactly that clear, with nothing designed for the refusal.
- **Reorder mismatch** — the drag interaction is the flagship (a whole gallery card) and `PUT /order` is **exhaustive**: a subset writes nothing, and the id set is compared inside the transaction against a `FOR UPDATE` re-read, so a concurrent delete in another tab is a real 409. No state for it.

### IMPORTANT — the eyebrow reinstates the naming collision the backend fought to prevent

Line 406 reads **"Fulfilment routing"** — which is the *other* concern's name. `apps/api/src/mappings/http/fulfillment-routing.controller.ts` owns `connections/:connectionId/routing-rules` (ADR-012 dispatch — *which carrier ships this*), and its operator surface is the connection Mappings tab labelled **`Fulfillment`** (`connection-mappings-page.tsx:451`). So an operator already has a "Fulfillment / routing rules" screen, and this page's kicker is a near-synonym.

`fulfillment-router.port.ts` opens with *"NOT the same thing as `mappings`' `FulfillmentRouting*` family"*, and the controller header adds that the two *"are deliberately kept apart … and sharing a URL namespace would wire them together at the layer an operator actually sees."* The backend spent a whole docblock on that separation; the UI hands it back. The page title (`Sourcing rules`) and description are good — the eyebrow should carry the same distinction (e.g. `Sourcing` / `OpenLinker OMS`), ideally with one clause saying this is not carrier selection.

### IMPORTANT — "Test against an order" has no backend

Line 411 `alert()`s *"Not shown in this mockup"*, which reads as a rendering shortcut rather than an unbuilt capability. `FulfillmentRouterPort.evaluate()` has no caller anywhere; the dry-run surface is **#2954**. A durable spec must say which controls the backend cannot serve, and name the issue — otherwise a sub-issue gets written for a button that cannot be wired.

### SUGGESTIONS

- `CreateSourcingRuleDto.position` is required (*"ordering is part of the contract, so it is stated rather than derived from insertion time"*) and the dialog has no field. Deriving it as append is a fine UX call — worth stating so the implementer knows the field must be **sent**, not omitted.
- `nearest`'s hashed-ship-to degradation is unstated: on that arm there is no postcode and it *"degrades to the country term alone — a degradation the emitted explanation step must state, because a ranking that quietly stops ranking is unfalsifiable from the outside."* A rule-authoring page is where an operator decides to use it.

- **`data-state` is on the switcher BUTTON only, never on the rendered state.** `setState(name)` toggles `chip--active` and replaces `innerHTML`; nothing writes `data-state` onto `<body>` or the container. The exemplar § UX Mockups names — `analytics-display-currency-picker.html` — drives every state from `body[data-state="…"]` (46 occurrences, 15 values), which is why an E2E task can *set* a state and assert on it. Fix: `document.body.dataset.state = name`, plus a value + switcher button for every dialog and gallery card.
- **The epics cite only the Artifact URL.** #3055/#3063/#3071 each carry a `## Mockup` section pointing at `https://claude.ai/code/artifact/…` and nothing else. § UX Mockups requires implementation tickets to cite the **file path** and `data-state` values, and the E2E baseline to come from the repo copy *"never from the Artifact URL — the Artifact can be edited or expire independently of the repo."* Repoint in the same pass.
- **Token fidelity, measured:** the light `:root` block invents **zero** token names and every colour value matches `index.css` exactly (only cosmetic diffs — `.25rem` vs `0.25rem`, spaced `cubic-bezier`). Nav labels match `nav-registry.ts` verbatim. `#ffffff` in `--bg-surface` and `#fff` in `.button--danger` are both faithful. The "lifted verbatim" claim is a mild overstatement, not a defect.
- The `<link>` to `fonts.googleapis.com` differs from `apps/web`, which **self-hosts** IBM Plex woff2 (`index.css:36`). Precedent-consistent, but an offline screenshot run falls back to `system-ui` and produces a different baseline.

---

norbert-kulus-blockydevs:
## Redesign addressing the review

@piotr — agreed on every point, all binding. Pushed a new commit that redesigns the mockup rather than patching the three called-out spots in isolation, because the two "important" findings (the color rule and the two stubs) and the "no audit trail" / "no scale path" findings turned out to share one root cause: the page was drawn from the closed vocabulary outward, without asking what an operator authoring rules that move real inventory actually needs to see, verify, and be warned about.

Updated Artifact (live preview): https://claude.ai/code/artifact/2a03d1ee-646b-4cff-b7a9-1187f00d12c3

### Process
Ran 5 independent research passes before touching the design: two backend deep-dives (the routing engine's actual vocabulary/contracts in `libs/oms/src/routing/*`, and the surrounding OMS/fulfillment domain context — navigation, PII allowlisting, the locations precondition), a product-requirements synthesis, and two rounds of competitor research (Shopify/ShipStation/NetSuite/Sellercloud/ChannelEngine/Cin7 for warehouse-routing specifically, and Stripe Radar/LaunchDarkly/PostHog/Cloudflare/Zendesk for general ordered-rule-engine UX). A tech-lead pass then synthesized all of it into an implementation-ready spec, which is what the new mockup builds from.

### What changed, mapped to your findings

1. **Color rule (#2507)** — `Active` now renders plain ink + a small tick, matching `docs/frontend-ui-style-guide.md`'s own rule verbatim ("a finished or expected state is plain ink; only a state someone has to act on is tinted"). `StatusBadge` tone is now reserved for Scheduled (info), Retired (neutral), and Unrecognised (warning) — the one state that actually needs the operator's attention.

2. **The two stubs** — both are now real:
   - The priority-location list is a working search-to-add combobox with an ordered, reorderable list of rows (position, name, active/inactive status, remove), inline handling for an empty list and for a stale/deleted location reference — not two hardcoded chips.
   - "Test against an order" opens a real dry-run dialog: a synthetic order builder (country/postcode/line items), a working step-by-step evaluation trace, and an explicit "Routing evaluated this order using: {country}, postcode {X}" line so the panel never implies it saw more of the order than the router's actual PII allowlist permits (confirmed against ADR-062 and `RoutingInput`'s discriminated ship-to type).

   Backend-side, this also surfaced that `evaluate()` has no HTTP endpoint anywhere yet (the pipeline itself is complete and already produces a rich `RoutingExplanationStep[]` trace) — flagged as required backend work in the design spec, not something the frontend alone can finish.

3. **No audit trail** — added a minimal "Last edited by X on Y" line (dialog header + row tooltip). Backend needs a small additive migration (`createdByUserId`/`updatedByUserId` on `oms_routing_rules`, nullable, mirroring `prompt_templates.createdBy`) — also flagged as required backend work, not fabricated in the mockup.

4. **Scale past a handful of rules** — added a status filter bar (Active/Scheduled/Retired/Unrecognised), defaulting to Active-only.

5. **Reorder rollback + double-submit** — nudge-button reorder now shows an optimistic toast with a defined revert-on-conflict path; Save now disables during submission.

6. **`country-served` copy** — unified to one corrected sentence used in the row description, the dropdown option, and the dialog hint (all three previously disagreed or were wrong): *"kept only if this location's own registered country matches the order's delivery country — this is about where the location physically is, not a configured shipping range."*

### The one thing none of this was originally built for
Every independent research angle converged on the same gap: the ruleset resolves its after-action (split policy) via **most-restrictive-wins across every currently active rule**, and that's a genuinely invisible cross-rule interaction — an operator can add an unrelated rule, leave its after-action at the default, and silently override three other rules' explicit permission to split. No competitor surveyed (including Stripe Radar, whose "wrong rule = real money" framing is the closest analog) resolves this way, so there's no existing UI pattern to copy. The redesign makes this a live, always-visible fact: a page-level panel naming which rule(s) currently govern the ceiling (hyperlinked to the row), a matching indicator on governing table rows, and an in-dialog preview that tells the operator *before they save* whether their change tightens or loosens it — with a hard confirmation gate specifically on the loosening case, since that's the one silent failure mode with real consequences.

### Also fixed while in there
- The rule create/edit dialog is restructured into three visually distinct sections (what triggers it / configure it / what happens next) instead of one flat run of fields — this was raised separately as a general complaint about this app's edit-modal pattern, and the sourcing-rules dialog was the concrete example.
- A zero-locations blocking state (routing can't mean anything with no locations to route to — mirrors the existing `RouterReadinessPanel` state model).
- Delete-confirm copy now names the ceiling consequence when deleting a currently-governing rule, instead of always showing the same generic warning.

### Explicit v1 scope cuts (stated, not silently dropped)
No drag-and-drop (no DnD library exists in the repo yet; the backend's reorder contract only needs an ordered id array, so keyboard-accessible nudge buttons — which you praised — stay the baseline). No full diff/version history on edits (Stripe ships one given identical stakes-reasoning, but there's no actor-column infrastructure to build on top of yet — flagged as a deliberate, not accidental, deferral). No backtest-against-historical-orders (Stripe's stronger version of the dry-run — needs new backend aggregation, noted as a natural v1.5/v2 follow-up). No per-rule dry-run entry point, only a page-level one testing the whole ruleset, since that's how the ceiling actually resolves. No location-management CRUD on this page (doesn't exist anywhere in the product yet — out of scope for this ticket).

Full research trail and the complete design spec (10 sections — IA, page layout, dialog redesign, dry-run spec, safety rails, audit trail, visual/tone mapping, scope cuts, backend requirements) are available if useful for whoever picks up the real implementation — happy to attach as a follow-up doc rather than inline here.

norbert-kulus-blockydevs:
## Full summary of changes since the original mockup

This branch went through 5 rounds of redesign since the original submission you reviewed, each grounded in independent research/review agents rather than my own guesses. Latest live preview: **https://claude.ai/code/artifact/2a03d1ee-646b-4cff-b7a9-1187f00d12c3**

### Round 1 — redesign addressing this review (commit `98b747abb`)
Covered in my earlier comment on this thread: the color fix (Active is plain ink, not a green badge), the priority-location picker and "Test against an order" built for real, the live "splitting-limit ceiling" panel naming which rule governs it, the sectioned ①②③ create/edit dialog, the corrected `country-served` copy, audit line, filter bar, and safety rails. Grounded in 5 research passes (backend source, OMS domain context, product requirements, two rounds of competitor research) plus a synthesized tech-lead spec.

### Round 2 — plain-language copy pass (commit `8b7e126f9`)
You flagged (separately, in chat) that the mockup was still hard to follow and read like it was written for a programmer. Rewrote every user-facing string: removed "ceiling"/"after-action"/"position"/order-line jargon in favor of plain operator language ("orders can be split like this", "sets today's splitting limit"), renamed the ambiguous "Order" column (collides with *customer* order) to "Step", added a plain-language intro with a worked example at the top of the page.

### Round 3 — tooltips/popovers + real bugs (commit `6a7852757`)
Two fresh UX reviews (first-time-operator walkthrough + signifier/affordance audit) found the page had almost no tooltip infrastructure — two working tooltips in the whole file. Added a proper two-tier primitive (native `title` for one-liners, a real hover/focus/click popover for richer explanations), applied it to every icon and badge, made the governing-rule link always clickable, added a live in-dialog summary sentence. Screenshot-verified before publishing (per this repo's own lesson that tag-balance checks miss visual bugs) — caught a real bug: "Add rule" silently defaulted to a rule name already claimed by another active rule, because assigning a disabled `<option>` via `select.value` selects it anyway in every browser.

### Round 4 — 4 independent reviews, real bugs fixed (commit `54dd1a142`)
Two e-commerce ops specialists (grounded in Shopify/ShipStation/Linnworks/NetSuite/Sellercloud behavior) + two senior UX/UI designers (one on visual craft via rendered screenshots, one driving every interaction via Playwright with console/pageerror listeners attached). Found and fixed real, confirmed bugs:
- `[hidden]` was silently losing to class-based `display` rules (equal CSS specificity) — the priority-location builder rendered even for Filter rules and was fully wired, so filling it in and saving would silently discard it (a real data-loss trap); a validation error was permanently visible from first paint; the delete-warning rendered as an empty box on every rule.
- No `max-height`/scroll on the dialog — at an ordinary 1280×900 viewport, Save/Cancel rendered off-screen with zero scroll affordance.
- No focus trap in any dialog — Tab walked off the last field onto the live table, and Enter there could open a *second*, stacked dialog (e.g. delete a different, currently-active rule) while the first still looked open.
- "Retire instead" stamped today's date, but the status logic reads the end date as inclusive — so a retired rule kept reading (and behaving) as Active for the rest of the day, directly contradicting the confirming toast.
- The dialog's live summary always said "You're creating…", even when editing an existing active rule.
- Two measured WCAG contrast failures, an info-popover clipped off-screen on mobile with no dismiss path besides re-clicking the trigger, reorder buttons at 11×11px with 0px gap (unusable on the tablet this surface is meant to also support), a "Drag, or use the arrows" promise with no drag implementation anywhere in the file, and a dry-run "Nearest" simulation that never actually reordered candidates despite its own trace claiming it did.

Also added, since both ops reviews independently flagged the same two gaps: **Duplicate rule** (starts a new rule pre-filled from an existing one — full identical copies aren't meaningful here since a rule name can only be live once), and the dry-run trace now names it when a sort rule changed nothing because an earlier rule already fully ranked everyone.

### Round 5 — OMS-completeness sign-off (commit `70cf39471`)
Two final reviews for a "do we have everything a real OMS needs" verdict — a master-level OMS domain expert reasoning from first principles, and a competitive pass going wider than prior rounds (Fluent Commerce, Manhattan Active Omni, IBM Sterling, Deposco, Fabric, Shopify's newer Location Rule Function API). Both independently found and I confirmed live with a dry run: **`country-served` silently makes every non-domestic order unfulfillable** for the single most common real topology (one warehouse hub serving several countries) — it's built for "one warehouse per country," and the tooltip explaining that was too quiet for a trap this consequential. It now gets a full warning alert the moment it's selected, not muted gray text. Also fixed the page's own worked example, which over-promised "distance to the buyer" for what's honestly a postcode-string proxy, and added real tie-break disclosure to the dry-run trace (previously a genuine, undetectable-by-the-operator gap: two locations tied after every rule ran, winner picked by an internal order nobody chose).

Both reviews' verdict: this is at genuine competitive parity for OpenLinker's actual tier, ahead of it on explainability (the ceiling-governance panel and the full-trace dry-run have no equivalent I could find at any competitor researched), and correctly/deliberately behind the enterprise-DOM tier (AI/ML routing, carbon-aware routing) on axes that would be scope mismatches. But **not defensible to ship with its real gaps left silent** — so I added a closed-by-default "What this doesn't do yet" disclosure naming the four things both reviews called genuinely missing rather than reasonably deferred: no inventory reservation/ATP (a real oversell risk — "in stock" checks presence, not availability, under concurrent orders), no shipping-cost/carrier-rate input, no draft/approval step before a save goes live, and no operator-definable custom-location-attribute rule. These need real backend/product scoping, not another mockup pass — naming them honestly is what makes the rest of the scope defensible for v1.

---

**Net effect**: 5 commits, the file roughly doubled in size, and every round was driven by rendering the actual page (Playwright/Chromium screenshots, real interaction driving, computed-style contrast measurement) rather than reading source and guessing — which is how round 4 caught the `[hidden]`/CSS-specificity bug and the focus-trap issue, and round 3/4/5 each caught at least one thing that only shows up when you actually click through it.

norbert-kulus-blockydevs:
Updated the mockup to match: entry and breadcrumb now route through `/settings`, not Connections.

Same fix as the sibling Inventory Locations mockup (#3120). Both were modelled as reached via the connection — sidebar `Connections` highlighted, breadcrumb `Connections / OpenLinker OMS / Sourcing rules`, back-link `← OpenLinker OMS`. The intended entry point is a link-out tile on `/settings`, mirroring `WhoDecidesTile` / `InventoryLocationsTile`'s shape — no sidebar item for this screen.

Changed:
- Sidebar: `Connections` is no longer shown active.
- Breadcrumb: `Platform / Sourcing rules` (2-level, matching the `{group, title}` crumb shape every settings-tile-backed page uses).
- Back-link: `← Settings`, not `← OpenLinker OMS`.
- Added a state-gallery card previewing the `/settings` tile (eyebrow, title, description, "Open" button) so the entry point is reviewable alongside the screen it links to.

piotrswierzy (changes requested):

The doubling (+1003 → +2112) went almost entirely into the priority-list builder, the dry-run dialog and a 17-card gallery. **Every finding that required editing this file's own markup was fixed; every finding that required editing a *shared* block, changing the eyebrow, or *removing* an affordance was not.**

| Prior finding | Verdict |
|---|---|
| `data-state` on switcher button only | ❌ **Unaddressed** |
| Edit dialog exposes `kind` | ❌ **Unaddressed** |
| `priorityLocationIds` not designed | ✅ **Resolved** — well |
| Retired rules unreachable | 🔄 Partial |
| Neither 409 depicted | ✅ **Resolved** |
| "Fulfilment routing" eyebrow | ❌ Unaddressed (now twice) |
| Dry run has no backend | ❌ **Unaddressed — regressed** |

**`priorityLocationIds` — resolved excellently.** A real searchable picker, per-row remove, ▲/▼ reorder with correct first/last `disabled`, an empty state saying the rule *"will rank nothing"*, an inactive-location badge, a stale-reference row for an id with no matching location, and a gallery card for that state. Materially better than the refusal card implied.

**Both 409s — resolved, and by prevention rather than depiction**, which is the better call: `buildNameSelect()` marks a claimed `(kind,name)` option `disabled` with *"— already active as Rule N"*, correctly skipping `editingId` and non-live rules. One consequence to note: because `isEditable` is `r.recognised && s !== 'Retired'`, a Retired rule cannot be edited at all — so the *revive-by-clearing-`effectiveTo`* path is now unreachable rather than refused, while `UpdateSourcingRuleDto.effectiveTo` documents clearing it as supported. Worth one sentence.

### Still blocking

**`data-state`** — `grep -c 'document.body.dataset'` → **0**. Five values exist, all as attributes on toolbar `<button>`s; 5 dialogs and 17 gallery cards carry no state attribute of any kind. **Both siblings fixed this in the same round** (#3120: 1 write, #3121: 5), so this is the one file left behind — and it is now the most state-rich of the three. One line: `document.body.dataset.state = name`.

**`kind` editable** — verified end-to-end this time. `openEdit()` checks the radio and calls `onKindChange`; the radios are never `disabled`; and `performSaveRule()` reads `input[name=kind]:checked` into a payload spread over the row (`rules[idx] = {...rules[idx], ...payload}`). So flipping Filter→Sort on an existing rule re-selects a name from the other vocabulary **and the mockup shows it saving**. `UpdateSourcingRuleDto` still has no `kind`.

**The dry run is a regression, and it is the finding I'd fix first.** The `alert('Not shown in this mockup')` is gone, replaced by a fully-designed dialog — prefill-from-recent-order, country + optional postcode, line items, a step-by-step elimination trace, verdict tones, a PII line. Nowhere does it say no backend serves this, and **#2954 appears nowhere in the file**; `FulfillmentRouterPort.evaluate()` still has no caller anywhere in the tree. Previously an implementer met a visible dead end. Now they meet a complete, convincing spec for a control that cannot be wired — the cheapest finding to fix (one sentence in `dialog__subtitle`) turned into the most expensive to discover.

**Eyebrow** — `Fulfilment routing` still at 580, and now a **second** occurrence at 1090 (the new `/settings`-tile card). The collision with `mappings`' ADR-012 dispatch tab is now stated twice in the durable spec.

### Cross-cutting, unaddressed

- **Dark palette** — all **18** `--status-*` dark values diverge from `index.css:405–504`. #3120 fixed this block to 15/15 exact in this same round; it was not propagated.
- **`--overlay-scrim`** never declared; `.dialog-backdrop` hardcodes `rgba(22,32,43,.45)`, so 5 dialogs keep a light scrim in dark mode.
- **Epic #3055** still Artifact-URL-only — and the URL it cites is the one this PR's own body labels **"superseded"**, so the epic points at a stale preview of a file that has since doubled.

### New

- `updateDialogCeilingPreview` computes a splitting ceiling from `kind`+`afterAction` and drives a blocking confirm — good design, but computed over a **mutable** `kind` on the edit path, so the preview can show a ceiling change PATCH cannot produce.
- The `no-locations` state offers *"Manage locations (coming soon)"* as disabled text. **#3120/#3063 is that page, in this same wave** — cross-link it rather than shipping "coming soon" into a durable spec.

Minimum to approve: stamp `data-state` on `<body>`; lock `kind` on edit; one sentence naming #2954; change the eyebrow in both places; paste #3120's dark block and `--overlay-scrim`; repoint #3055.

---

norbert-kulus-blockydevs:
## Round 2 addressed — `3f6434d32`

Every item from the re-review, plus one it got wrong. Artifact republished (v7) and the repo file is **byte-identical** to it, so the E2E baseline and the preview can no longer drift.

| Finding | Status |
|---|---|
| `data-state` on switcher button only | ✅ `body[data-state]`, **29** values |
| Edit dialog exposes `kind` | ✅ locked, and the preview follows |
| Dry run has no backend (regressed) | ✅ named in the dialog, **#2954** |
| "Fulfilment routing" eyebrow ×2 | ✅ both gone |
| Dark `--status-*` diverges (18 values) | ✅ verbatim from `index.css`, verified 27/27 |
| `--overlay-scrim` never declared | ✅ both themes, `.dialog-backdrop` reads it |
| Retired rules unreachable | ✅ see below — and the premise was inverted |
| Epic #3055 Artifact-URL-only | ✅ repointed |
| `no-locations` "coming soon" | ✅ cross-links #3063 / #3120 |
| Ceiling preview over mutable `kind` | ✅ falls out of the `kind` lock |
| `position` required / `nearest` degradation | ✅ both recorded |

### `data-state`

`grep -c 'document.body.dataset'` → **1**: one writer (`setBodyState`), called from the view switcher, both dialog entry points and the gallery. Same shape as #3120.

- **views** (5): `populated`, `no-locations`, `empty`, `loading`, `error`
- **dialogs** (6): `rule-create`, `rule-edit`, `confirm-loosen`, `confirm-delete`, `rule-unrecognised-locked`, `dry-run` — create and edit are separate values because the `kind` lock differs, which is the part a spec needs to assert on
- **gallery** (18): each card's label *is* its switcher button and shows its own value, so there's no toolbar row of eighteen chips nobody could scan

Closing a dialog restores the view state underneath rather than hardcoding `populated`, and both dismiss paths (backdrop click, Escape) route through `closeDialog` — otherwise the body kept naming a dialog that was no longer on screen.

### `kind` on edit

Radios `disabled`, reason stated in place. `buildDraftRule` and `performSaveRule` now read `kind` from the **row**, not the form — which is what actually kills the ceiling-preview problem, since the preview can no longer compute over a value `PATCH` cannot carry.

### Dark palette

Lifted verbatim from `index.css:439-492`, and it needed to be **wider** than the trio: the previous block omitted the six BASE `--status-*` tones, which `.rule-state--plain .tick` and `.button--danger` read — so a light-theme tone was being painted on a dark surface. Checked programmatically against `index.css`: **27 tokens, zero value mismatches**, in both the media-query and `[data-theme]` blocks. `--status-conflict*` is deliberately still absent: the mockup uses no conflict tone, and declaring a colour only behind `[data-theme]` is the bug the convention warns about.

### One finding was inverted — and the code under it was genuinely broken

> because `isEditable` is `r.recognised && s !== 'Retired'`, a Retired rule cannot be edited at all — so the revive-by-clearing-`effectiveTo` path is now unreachable rather than refused

`isEditable` doesn't exist in this file. That's **`isLiveRule`**, the claimed-slot test for the name select. The edit button gates on `r.recognised` **alone** (`rule-edit` button, `openEdit`), so a retired rule **is** editable and the revive path **is** reachable.

Reachable, and defective. `buildNameSelect` disabled any `(kind,name)` a live sibling claimed, and `onKindChange`'s disabled-option fallback then silently re-selected a different name. Concretely: retire "Country served", add a live rule with the same name, click edit on the retired one — the form opened saying **"In stock"**. A silent identity change on the one field that *is* the row's identity.

Fixed by separating the two questions:

- the edited row's **own** name stays selectable (identity, not a choice)
- the collision is raised where it actually happens — clearing the end date to revive is **refused** when a live sibling holds the same kind+name, which is exactly the 409 your first review asked to see and which the name-select prevention structurally cannot cover, because the field it collides on is the *date*

Driven end to end: refusal blocks the save and leaves the row retired; remove the rival and the same revive goes through with the name intact. Pinned as `gallery-revive-conflict`.

I also corrected the limitations bullet I had written from the review's wording before checking it.

### Retired / `includeSuperseded`

Kept as filter chips rather than #3120's separate checkbox, because the dishonesty was the *silence*, not the control: choosing Retired or All now states it re-fetches with `includeSuperseded=true`, and the default state says retired rules are not loaded. Happy to switch to the checkbox if you'd rather the wave be identical here.

### Left open, deliberately — two need your call

1. **Epic #3055 contradicts this mockup in two places.** I repointed the `## Mockup` section but did **not** rewrite the architecture, so both are flagged there instead:
   - *Entry point* — the epic mounts this as a **section on the connection detail page** (`RouterReadinessPanel` precedent, with a sub-issue for it); the mockup makes it its **own page off a `/settings` tile** (no sidebar item, `gallery-settings-tile` says the tile is the only way in).
   - *Reorder* — a sub-issue says **drag-to-reorder**; the mockup deliberately ships **arrows only** ("no DnD library", recorded on `.rule-position__handle`, which is `cursor: default` so it can't promise an interaction nothing backs).

   One of the two documents has to move before the sub-issues are picked up.

2. **`Closes #N` against a feature-branch base** — confirmed: this PR targets `3055-sourcing-rules-admin-ui`, not `main`, so `Closes #3112` won't fire on merge. Retargeting is a call about the epic's integration strategy, so I left the base alone rather than changing it under you.

3. **No doctype / Google Fonts.** The repo copy has no `<!DOCTYPE html>` (the Artifact platform injects its own, and the tool forbids one in the content), so opening the file directly renders in **quirks mode** while the published page doesn't. Adding one would diverge the two copies by a line; the exemplar `analytics-display-currency-picker.html` has none either, and 10 of 21 committed mockups are in the same position. Left as-is and recorded. The font link is likewise recorded in place: `apps/web` self-hosts IBM Plex, a single-file mockup can't, so an **offline** baseline run falls back to `system-ui`.

Verified in both themes with Playwright against the **repo file**: every view and dialog state stamps and restores, `kind` locked on edit and free on create, scrim resolves `rgba(15,18,26,0.55)` / `rgba(0,0,0,0.72)`, 18 gallery switchers, tag balance clean, no undefined handlers, **no JS errors**.

norbert-kulus-blockydevs:
## Branch reset onto current `main`

The reviewed content was diverging from `main` — the branch had been cut before ~50 unrelated commits landed (infakt adapter changes, analytics dialog rework, docs/user-guide renumbering, the ADR-number-uniqueness guard), so the diff was showing all of that drift alongside the actual mockup change.

Reset both branches to fix it:
- `3112-sourcing-rules-mockup` force-pushed to a single fresh commit (`f451fb901`) carrying forward only the reviewed mockup file, rebuilt onto current `origin/main`.
- `3055-sourcing-rules-admin-ui` (the epic branch, zero unique commits of its own) fast-forwarded to current `origin/main` — a clean fast-forward, nothing lost.

The diff is now `docs/plans/mockups/sourcing-rules-3055.html` alone (+2347/-0).

## Re-verified against the latest `/pr-review` (2026-09-11T10:42:42Z) — every blocking item confirmed present in the carried-forward file

| Finding | Status |
|---|---|
| `data-state` on switcher button only | **Fixed** — `document.body.dataset.state = name` on every transition (line 1525) |
| Edit dialog exposes `kind` | **Fixed** — `input[name=kind]` disabled when locked (line 1753), tiles rendered non-interactive on edit |
| Dry run has no backend | **Fixed** — #2954 named twice, once in the limitations list and once directly in the dialog's own subtitle |
| "Fulfilment routing" eyebrow | **Fixed** — reads "Sourcing · OpenLinker OMS" in both places now |
| Dark palette diverges from `index.css` | **Fixed** — verified 1:1 against `apps/web/src/index.css` on current `main` (success/warning/error/info/review/disabled, base + soft/border/fg) |
| `--overlay-scrim` never declared | **Fixed** — declared and consumed by `.dialog-backdrop` |
| "Manage locations (coming soon)" | **Fixed** — now cross-links #3063/#3120 |

No content changes beyond the reset — the file already answered every finding, it just needed a clean branch under it. Not merging; leaving open for review.

piotrswierzy (approved):
## Tech Lead review

Re-reviewed `f451fb901` (pushed after the last round). **Every remaining blocker from the re-review is verified fixed, and the branch reset worked** — `git diff origin/3055-sourcing-rules-admin-ui...origin/3112-sourcing-rules-mockup` is now exactly one file, +2347, no drift (GitHub's own file list is stale two-dot output; ignore it). Domain accuracy re-verified independently against the merged backend rather than taken from the previous round.

**Verified fixed**
- `data-state` — one writer, `setBodyState()` at L1524 doing `document.body.dataset.state = name`, called from the view switcher, both dialog entry points, and the gallery. 35 attribute sites, 5 view + 6 dialog + 18 gallery values.
- Edit dialog `kind` lock — `openEdit()` (L1718) sets `locked` **before** calling `onKindChange`, so the name list is built for the row's own kind; `#kindLockedNote` un-hides. Matches `UpdateSourcingRuleDto`, which carries `position`/`name`/`afterAction`/`priorityLocationIds`/`effectiveFrom`/`effectiveTo` and deliberately not `kind`.
- Dry run named as unbuilt — L723 cites `FulfillmentRouterPort.evaluate()` and **#2954** by number.
- "Fulfilment routing" eyebrow — gone, both occurrences; entry is now `← Settings` (L623) with a two-level `Platform / Sourcing rules` crumb, plus a gallery card for the `/settings` tile.
- Dark `--status-*` — extracted all 24 declared and diffed against the `html[data-theme='dark']` block of `apps/web/src/index.css`: **zero divergence**.
- `--overlay-scrim` — declared in all three scopes (L79/L123/L155) and read by `.dialog-backdrop` (L427).
- Retired/revive — the edit button gates on `r.recognised`, not `isLiveRule`, so a Retired-but-recognised rule *is* openable and the clear-`effectiveTo` revive path is reachable; `isLiveRule` is correctly scoped to the `(kind,name)` slot claim only, and `gallery-revive-conflict` depicts the one collision the name picker cannot prevent.

**Independently re-verified against source (not carried over)**
- Both closed vocabularies are byte-exact against `routing-vocabulary.types.ts:55/75/83` — the mockup uses all ten values and introduces none.
- `includeSuperseded=true` (L689, L1480) is the real query parameter name — `ListSourcingRulesQueryDto.includeSuperseded`, and the copy's "a rule retiring in the future is still listed" reading matches that DTO's own description.
- No undeclared CSS custom property: 73 declared, 61 used, empty difference.

### Findings
**[SUGGESTION]** — `docs/plans/mockups/sourcing-rules-3055.html:986-1178`, `:2326`
> The 18 gallery cards carry `data-state` on the card element, and every card is permanently in the DOM (`querySelectorAll('.gallery-card[data-state]')` only toggles a highlight class). So a bare `[data-state="gallery-retired"]` selector matches in every state — the same ambiguity the re-review named for the switcher buttons, one level down. `body[data-state]` is stamped correctly, so nothing is broken; just worth a line in the epic saying E2E specs must anchor on `body[data-state="…"]`, not the bare attribute.

### Verdict
✅ **Approve**

---

Issue #2953, opened 2026-09-06
[TASK] OMS routing — no HTTP surface for authoring the router's ruleset


piotrswierzy:
## Problem

`OlFulfillmentRouter` reads its ordered ruleset through `RoutingRuleSourcePort.listActiveRules(connectionId, now)`. The storage half is fully shipped:

- `libs/oms/src/routing/oms-routing-rule.orm-entity.ts`
- `libs/oms/src/routing/oms-routing-rule.repository.ts`
- `libs/oms/src/migrations/1869000000200-create-oms-routing-rules.ts`
- `OmsModule` binds the repository to `ROUTING_RULE_SOURCE_TOKEN` and exports it

**There is no controller.** `ROUTING_RULE_SOURCE_TOKEN` has exactly one consumer in the tree, and it is a test (`apps/api/test/integration/oms-routing-rules.int-spec.ts`). An operator cannot create, order, edit or delete a routing rule by any means other than inserting rows directly into the database.

Note the `mappings` `fulfillment-routing.controller.ts` is **not** this — that surface answers *"which processor or carrier DISPATCHES this?"* (#832, ADR-012). This one is *"which location and holder SOURCES it?"*. The two questions are deliberately kept apart and must not be wired into one another.

## What this blocks

Wave 3a's exit criterion (#2412): *"an operator with two locations enables the OL-OMS connection, **authors an ordered filter/sort list**, sees a dry-run explanation for an incoming order, and works the resulting tasks to shipped."*

The middle clause is unreachable. With the router wired (#2408) but no ruleset, `listActiveRules` returns `[]` — which the router treats as *"not configured"*, so every line resolves unfulfillable and `RoutingCommitService.refusalFor` refuses the plan with `plan-carries-unfulfillable`. The operator sees a routing refusal and has no way to fix it.

## Scope

- CRUD for `oms_routing_rules`, scoped to an OMS connection, admin-gated.
- Ordering is part of the contract — the rules are an **ordered** list evaluated in sequence, so the write surface has to express order explicitly rather than leaving it to insertion time.
- The closed vocabularies already exist and must be reused verbatim, never restated: `RoutingFilterNameValues`, `RoutingSortNameValues`, `RoutingRuleKindValues`, `RoutingAfterActionValues`, plus `coerceRoutingRule` / `isRoutingRule` (all barrel-exported from `@openlinker/oms`).
- A frontend authoring surface is a separate follow-up; this issue is the API.

## Out of scope

[...]

Issue #2319, opened 2026-08-22
[TASK] Infrastructure — duplicate inventory-position detection pass


piotrswierzy:
> Part of the OMS programme backlog (**Wave 1b**, stream S1, size S). Backlog slug: `W1b-9`. Sibling `W*` slugs resolve via the Wave-1b epic's checklist. Design of record: `docs/plans/analysis/DESIGN-oms-authority-model.md` + ADRs 052–062.

## Problem / Context

REVIEW §3 D7: the **existing** partial unique indexes (`inventory-item.orm-entity.ts:29` and `:34`) already NULL-dup on `locationId`, so duplicate locationless positions exist on real installs today and are summed by the availability read. Any future index recreation (Wave 1d) fails outright on a dirty install, so detection must precede it.

## Proposed Solution

A read-only detection pass (an admin endpoint and/or a one-shot reporting job) that groups `inventory_items` by `(productId, productVariantId, locationId, sourceConnectionId)` and reports groups with count > 1, with enough detail (ids, quantities, `updatedAt`) for an operator to resolve them. **Read-only** — it never merges or deletes; a wrong automatic merge on the oversell table is worse than the duplicate.

Document the remediation procedure in `docs/operations/`, and make Wave 1d's cleanliness check this pass returning zero.

## Classification

**Type**: CORE / Infrastructure
**Layer**: Application + Interface
**File(s)**: `libs/core/src/inventory/application/services/`, `apps/api/src/inventory/http/`, `docs/operations/`

## Dependencies

- #2314 (`W1b-3` — so the report can group by provenance)

## Assumptions

- Duplicates are rare enough that a manual remediation procedure is acceptable; if a real install reports large numbers, an assisted-merge tool is a follow-up, not part of this issue.

## Acceptance Criteria

- [ ] Pass reports duplicate groups with ids and quantities; performs no write
- [ ] Integration test seeds two locationless rows for one variant and asserts detection
- [ ] Operations doc records the remediation procedure and names this pass as Wave 1d's gate
- [ ] Tests added or updated for non-trivial logic
- [ ] No architecture boundary violations (CORE ↔ Integration)

Issue #2316, opened 2026-08-22
[TASK] Interface — inventory locations CRUD API


piotrswierzy:
> Part of the OMS programme backlog (**Wave 1b**, stream S1, size S). Backlog slug: `W1b-2`. Sibling `W*` slugs resolve via the Wave-1b epic's checklist. Design of record: `docs/plans/analysis/DESIGN-oms-authority-model.md` + ADRs 052–062.

## Problem / Context

Locations are operator-authored; without a write surface, W1b-1's table is unreachable. REVIEW §3 D3 also notes the router must refuse to enable until ≥1 location exists — a rule that needs a way to create one.

## Proposed Solution

`GET/POST/PATCH/DELETE /inventory/locations` under `JwtAuthGuard` with `class-validator` request DTOs and an allowlist response DTO. Delete is refused (409) when positions reference the location. Admin-gated for writes, readable by operator roles.

The operator **FE screen is deliberately out of scope for Wave 1b** — no multi-location workflow exists to hang it on until Wave 2/3; the API is what unblocks the backfill and the router's bootstrap.

## Classification

**Type**: CORE / Interface
**Layer**: Interface
**File(s)**: `apps/api/src/inventory/http/`, `libs/core/src/inventory/application/services/`

## Dependencies

- #2313 (`W1b-1`)

## Assumptions

- No FE screen ships in this wave; filed as a Wave-2 follow-up alongside the authority-status page.

## Acceptance Criteria

- [ ] All four verbs implemented, validated at the boundary, guarded
- [ ] Deleting a location with referencing positions returns 409 with a domain-named error
- [ ] `code` uniqueness violation returns 400/409, never a 500
- [ ] Response DTO is an explicit allowlist
- [ ] Tests added or updated for non-trivial logic
- [ ] No architecture boundary violations (CORE ↔ Integration)

PR #3120, opened 2026-09-10
docs(mockups): inventory-locations UI mockup

### LINKED ISSUE #3113: design(inventory): inventory-locations UI mockup
## Problem / Context

Before implementation starts, Inventory locations CRUD UI (#3063) needs a reviewed design so every sub-issue builds against

## Proposed Solution

## Classification

**Type**: Frontend
**File(s)**: `docs/plans/mockups/inventory-locations-3063.html`

## Dependencies

- None — first sub-issue of #3063, everything else in the epic reads from this mockup

## Assumptions

## Acceptance Criteria

- [ ] `docs/plans/mockups/inventory-locations-3063.html` exists and renders standalone (no external API calls)
- [ ] PR links back to the reviewed Artifact: https://claude.ai/code/artifact/fbb5f9e1-52a3-4b7b-8136-ad59dcf4f318

norbert-kulus-blockydevs:
## Summary
- Adds the reviewed `inventory-locations` mockup as a static file at `docs/plans/mockups/inventory-locations-3063.html`
- Reviewed Artifact (live preview): https://claude.ai/code/artifact/fbb5f9e1-52a3-4b7b-8136-ad59dcf4f318
- Closes #3113, first sub-issue of #3063

## Test plan

### DIFF
--- a/docs/plans/mockups/inventory-locations-3063.html
+++ b/docs/plans/mockups/inventory-locations-3063.html
@@ -172,6 +172,7 @@ button:focus-visible, a:focus-visible, input:focus-visible, select:focus-visible
 
 .loc-cell__name{ font-weight:600; color:var(--text-primary); }
 .loc-cell__code{ font-family:var(--font-mono); font-size:.6875rem; color:var(--text-muted); margin-top:2px; }
+.loc-positions-link{ font-family:var(--font-mono); color:var(--text-link); }
 .loc-kind-badge{ font-family:var(--font-mono); font-size:.625rem; text-transform:uppercase; letter-spacing:.06em; padding:1px 6px; border-radius:var(--radius-xs); background:var(--bg-surface-muted); color:var(--text-secondary); border:1px solid var(--border-subtle); }
 .loc-actions{ display:flex; gap:4px; justify-content:flex-end; }
 
@@ -269,6 +270,13 @@ input:read-only{ background:var(--bg-surface-muted); color:var(--text-secondary)
       <button class="chip" data-state="error" onclick="setState('error')">Error</button>
       <span style="margin-left:auto; font-size:.6875rem; color:var(--text-muted); font-family:var(--font-mono);">try: filter by kind, edit, retire, delete a used location</span>
     </div>
+    <div class="demo-toolbar" role="group" aria-label="Mockup role switcher — not part of the product" style="top:96px;">
+      <span class="demo-toolbar__label">Mockup · viewing as</span>
+      <button class="chip chip--active" data-role="admin" onclick="setRole('admin')">Admin</button>
+      <button class="chip" data-role="operator" onclick="setRole('operator')">Operator</button>
+      <button class="chip" data-role="viewer" onclick="setRole('viewer')">Viewer</button>
+      <span style="margin-left:auto; font-size:.6875rem; color:var(--text-muted); font-family:var(--font-mono);">writes are admin-only (LocationsController @Roles('admin')); operator/viewer get a read-only page</span>
+    </div>
 
     <div class="shell-content">
       <a class="back-link" href="#">&larr; OpenLinker OMS</a>
@@ -280,7 +288,7 @@ input:read-only{ background:var(--bg-surface-muted); color:var(--text-secondary)
             <p class="page-description">The warehouses, stores and third-party sites OpenLinker can source stock from. A location isn't authority over stock — it's the place sourcing rules pick between. Delete is refused while any stock still points here; retire it instead.</p>
           </div>
           <div class="page-header__actions">
-            <button class="button button--sm" onclick="openCreate()">+ Add location</button>
+            <button class="button button--sm" id="addLocationBtn" onclick="openCreate()">+ Add location</button>
           </div>
         </div>
       </header>
@@ -384,6 +392,7 @@ input:read-only{ background:var(--bg-surface-muted); color:var(--text-secondary)
         <div class="form-field">
           <label class="form-field__label" for="locCountry">Country <span style="color:var(--text-muted); font-weight:400;">(optional)</span></label>
           <input type="text" id="locCountry" placeholder="PL" maxlength="2" style="text-transform:uppercase;" />
+          <p class="form-field__error" id="locCountryError" hidden>✕ Exactly 2 letters (ISO 3166-1 alpha-2), or leave blank.</p>
         </div>
         <div class="form-field">
           <label class="form-field__label" for="locPostcode">Postcode <span style="color:var(--text-muted); font-weight:400;">(optional)</span></label>
@@ -453,7 +462,7 @@ input:read-only{ background:var(--bg-surface-muted); color:var(--text-secondary)
             <span class="alert__glyph">✕</span>
             <div class="alert__content">
               <p class="alert__title">Can't delete "Kraków — Overflow"</p>
-              <p class="alert__description">Stock positions still point here. Retire it instead — existing history keeps pointing at a row that exists, and it can be re-activated later.</p>
+              <p class="alert__description">42 inventory positions still point here. Retire it instead — existing history keeps pointing at a row that exists, and it can be re-activated later.</p>
             </div>
           </div>
         </div>
@@ -505,23 +514,49 @@ function connectionDisplayName(id){ const c = CONNECTIONS[id]; return c ? `${c.p
 const KIND_LABEL = { warehouse:'Warehouse', store:'Store', 'third-party':'Third-party', virtual:'Virtual' };
 
 const LOCATIONS_SEED = [
-  { id:'l1', code:'MAIN', name:'Warsaw — Main warehouse', kind:'warehouse', status:'active', countryIso2:'PL', postcode:'02-677', latitude:52.2297, longitude:21.0122, ownerConnectionId:'c1', externalRef:null, inUse:true },
-  { id:'l2', code:'KRK-OVF', name:'Kraków — Overflow', kind:'warehouse', status:'active', countryIso2:'PL', postcode:'30-001', latitude:null, longitude:null, ownerConnectionId:null, externalRef:'internal ref 4471', inUse:true },
-  { id:'l3', code:'SHOP-01', name:'Gdańsk — Flagship store', kind:'store', status:'active', countryIso2:'PL', postcode:'80-001', latitude:null, longitude:null, ownerConnectionId:null, externalRef:null, inUse:false },
-  { id:'l4', code:'3PL-DE', name:'Berlin — 3PL partner', kind:'third-party', status:'active', countryIso2:'DE', postcode:'10115', latitude:52.5200, longitude:13.4050, ownerConnectionId:'c2', externalRef:null, inUse:true },
-  { id:'l5', code:'CONSIGN-01', name:'Consignment site — Silesia distributor', kind:'virtual', status:'inactive', countryIso2:'PL', postcode:null, latitude:null, longitude:null, ownerConnectionId:null, externalRef:null, inUse:true },
+  { id:'l1', code:'MAIN', name:'Warsaw — Main warehouse', kind:'warehouse', status:'active', countryIso2:'PL', postcode:'02-677', latitude:52.2297, longitude:21.0122, ownerConnectionId:'c1', externalRef:null, positionCount:128 },
+  { id:'l2', code:'KRK-OVF', name:'Kraków — Overflow', kind:'warehouse', status:'active', countryIso2:'PL', postcode:'30-001', latitude:null, longitude:null, ownerConnectionId:null, externalRef:'internal ref 4471', positionCount:42 },
+  { id:'l3', code:'SHOP-01', name:'Gdańsk — Flagship store', kind:'store', status:'active', countryIso2:'PL', postcode:'80-001', latitude:null, longitude:null, ownerConnectionId:null, externalRef:null, positionCount:0 },
+  { id:'l4', code:'3PL-DE', name:'Berlin — 3PL partner', kind:'third-party', status:'active', countryIso2:'DE', postcode:'10115', latitude:52.5200, longitude:13.4050, ownerConnectionId:'c2', externalRef:null, positionCount:15 },
+  { id:'l5', code:'CONSIGN-01', name:'Consignment site — Silesia distributor', kind:'virtual', status:'inactive', countryIso2:'PL', postcode:null, latitude:null, longitude:null, ownerConnectionId:null, externalRef:null, positionCount:6 },
 ];
 
 let locations = JSON.parse(JSON.stringify(LOCATIONS_SEED));
 let kindFilter = '';
 let editingId = null;
 let pendingDeleteId = null;
+// LocationsController: reads are @Roles('admin','operator','viewer'), writes @Roles('admin')
+// only — an operator/viewer can open this page but every mutating action must be disabled.
+let currentRole = 'admin';
+function canWrite(){ return currentRole === 'admin'; }
+function setRole(role) {
+  currentRole = role;
+  document.querySelectorAll('.demo-toolbar .chip[data-role]').forEach(c => c.classList.toggle('chip--active', c.dataset.role === role));
+  document.getElementById('addLocationBtn').disabled = !canWrite();
+  document.getElementById('addLocationBtn').title = canWrite() ? '' : "Requires the admin role — you're viewing read-only";
+  renderLocations();
+}
 // The list read is server-side paginated (ListLocationsQueryDto: page/limit); this mockup's
 // 5-row demo dataset fits on one page, so pagination is shown as a static affordance rather
 // than wired to real paging state. countryIso2 / codePrefix query filters are likewise a
 // scoped-out choice for this pass, not an oversight — Kind + status are the ones drawn.
+//
+// `positionCount` is drawn here as if the list read already exposed it, which today it does
+// not: LocationResponseDto/toDto carries no positions-count field, and the only place the
+// count exists server-side (countPositionsAtLocation) surfaces solely as free text inside
+// LocationInUseError's 409 message. Rendering "in use" without a number is what the pre-fix
+// version of this mockup did — reviewers flagged that an operator managing several locations
+// needs the count (both to gauge blast radius before retiring, and because right now the real
+// page has nowhere to get one without either a new bulk/per-id count endpoint or fragile
+// 409-message parsing). This is a backend-contract decision #3063's implementation needs to
+// make, not something this mockup can resolve on its own — flagging it here so it isn't lost.
 
 function escapeHtml(s){ return String(s).replace(/[&<>"']/g,(c)=>({'&':'&amp;','<':'&lt;','>':'&gt;','"':'&quot;',"'":'&#39;'}[c])); }
+function writeDisabledAttrs(label, title) {
+  const disabled = !canWrite();
+  const t = disabled ? "Requires the admin role — you're viewing read-only" : (title || '');
+  return `<button class="button button--ghost button--icon button--sm" aria-label="${escapeHtml(label)}"${disabled ? ' disabled' : ''}${t ? ` title="${escapeHtml(t)}"` : ''}`;
+}
 
 function setKindFilter(kind) {
   kindFilter = kind;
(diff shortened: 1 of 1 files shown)


norbert-kulus-blockydevs:
## PR Review — #3120 (extended pass: frontend/UX + OpenLinker domain + e-commerce ops + backend architecture)

*This review was originally posted as a senior-frontend/UX pass. It's now been extended with three additional lenses (OpenLinker domain-architecture fidelity, e-commerce operations, senior backend/system architecture) and consolidated into one review below — no new findings were dropped, everything is folded into one voice with one verdict.*

**Scope**: `docs/plans/mockups/inventory-locations-3063.html` (new, single-file mockup) — the source-of-truth artifact for the `#3063` epic (the already-reviewed #3131–#3137 CRUD implementation).

### Alignment with the real backend model ✅

Cross-checked the form fields and every operator-facing hint against `docs/architecture-overview.md § 3 Inventory` ("Locations are first-class rows (#2313)" / "Locations CRUD (#2316)"):

- Field set matches exactly: `code`, `name`, `kind`, `ownerConnectionId`, `countryIso2`, `postcode`, lat/lng, `externalRef` — no invented field, no missing real one.
- Copy is architecturally precise, not just plausible: "Not authority over the location" for the owning-connection field mirrors the doc's "`ownerConnectionId` is provenance, never authority"; "Not an identifier mapping — nothing external reads this" for the operator reference mirrors "`externalRef` is a free-text operator field, not an identifier mapping."
- The empty state's "every order is currently unfulfillable" + "Create first location" CTA matches #2407's degraded-state framing and the real `POST /inventory/locations/bootstrap` remediation, including its idempotency copy ("Re-running it after this is a safe no-op" — matches "insert-then-recover... a re-run reports the code in `existingCodes` and creates nothing").
- **The bootstrap-vs-manual-create split is handled correctly** — a real risk point I checked specifically, since those are two different backend endpoints with two different operator intents: the empty state cleanly separates "Create first location" (bootstrap semantics, explained inline) from "+ Add manually" (opens the ordinary create dialog). No conflation.
- Delete-refused correctly routes to retire (matches the documented 409 `LocationInUseError` → `status=inactive` remediation), and no synthetic "default/unassigned" location is ever implied — consistent with `locationId IS NULL` permanently meaning "the master declined to locate," never a bucket the UI should stand in for.

This is unusually well-grounded for a mockup.

### 🟡 Real finding 1 — the mockup lets an operator edit an immutable field

`openEdit()` populates the same `locCode` input used for create — plain, editable, not disabled or read-only — and `saveLocation()` writes whatever the operator typed back onto the existing row's `code`. But per the real contract (#2316), **`code` is deliberately absent from the PATCH body** — "the natural key is not a patch-shaped field," i.e. it's immutable after create. The mockup draws a capability (renaming a location's code) the real API does not and will not support. Either grey out/hide `locCode` in edit mode, or add explicit copy explaining it can't change post-create — right now an operator would reasonably expect typing a new code and hitting Save to work.

### 🟡 Real finding 2 — the mockup's own copy promises reactivation, but no affordance exists for it

The delete-refusal alert says *"Retire it instead — existing history keeps pointing at a row that exists, **and it can be re-activated later**."* But:
- Row actions are unconditionally `✎ Edit` / `🗑 Delete` regardless of `status` — there is no `Reactivate` action anywhere for an inactive row, confirmed in `renderLocations()`.
- The edit dialog has no status field, so editing a retired location can't flip it back to active either.
- `retireLocation()` is reachable **only** as the rescue action inside a failed-delete alert — there's no direct `Retire` action on an active row for an operator who wants to proactively retire (closing a warehouse) without first hitting the 409.

Either the copy is overselling a capability that isn't modeled, or a real row action (the actual API supports `PATCH .../status`) is missing. Worth resolving before this becomes the implementation spec.

### 🟢 E-commerce operations notes (non-blocking, scale/future-facing)

- **No search or pagination affordance** in the location list — only a `kind` chip filter and a "show retired" checkbox. That's fine for the 5-row demo dataset, but this page is explicitly framed (page description: "the warehouses, stores and third-party sites OpenLinker can source stock from") for a multi-warehouse/3PL-integrated seller — the exact profile that outgrows chip-filtering fast. Worth a note for the real implementation, even if out of scope for this mockup pass.
- **Lat/lng are raw numeric inputs** (`type="number" step="0.0001"`) with no map-picker or address-to-geocode flow. A known friction point in real WMS/fulfillment tooling — low priority since both fields are optional and only feed the fulfillment router's location filters, but worth flagging for the real form.
- No bulk-import path (CSV) for a seller onboarding many locations at once — reasonable to defer, flagging for awareness rather than as a defect of this pass.

### 🔄 Repeats the exact anti-pattern flagged on sibling PR #3119

`docs/frontend-ui-style-guide.md` (#2507): *"color marks the exception, not the normal case... A sales-document column that tinted every row green on a healthy list was the case that produced it."* Line 521 renders **every** `Active` row with `status-badge--success` (green) — the expected, common case here, not something anyone needs to act on. Same defect as #3119; worth fixing consistently in both places.

### 🟢 Minor, non-blocking

- The state switcher (`populated`/`empty`/`loading`/`error`) carries `data-state` only on the toggle buttons, not on the rendered content region — cosmetic, an E2E spec can still drive it via the labeled chip.
- Pulls Google Fonts over the network, a mild tension with the PR's own "renders standalone, no external calls" test-plan line — shared boilerplate across the whole `#3063` batch, flagging for awareness rather than as this PR's problem specifically.
- The duplicate-code check is client-side/in-memory only (no simulated round-trip latency before the 409), an inherent simplification of any static mockup — not singling this PR out, just noting it won't be exercised as a true async flow until the real implementation.

### Verdict

**🔄 Approve with changes.** Data-model fidelity and workflow framing (delete-refused-then-retire, bootstrap idempotency, the create-vs-bootstrap split) are excellent and clearly grounded in the real backend contract — better than an average first mockup pass. Two things need fixing before this becomes the implementation spec: the code field must not be editable post-create (or must say so), and the status/lifecycle affordance gap (missing reactivate/direct-retire actions, or softened copy) needs resolving — plus the recurring green-on-every-row tinting shared with #3119.

piotrswierzy (changes requested):

**The most accurate of its batch on the rules that matter for this table.** Verified: both closed vocabularies exactly; the 409-on-delete-while-referenced with `status=inactive` as the retirement path, **including the reason** (*"existing positions keep pointing at a row that exists"* — `location.types.ts`'s own docblock); `ownerConnectionId` captioned *"Whose sync may write stock here. Not authority over the location"*, matching ADR-052's provenance-not-authority rule and the DTO's *"Provenance only, never authority"*; `externalRef` captioned *"Not an identifier mapping"*; the idempotent `MAIN` bootstrap with a no-op re-run card. The `show retired` checkbox is a real improvement over its sibling #3119. Good use of `alert__actions` to offer "Retire instead" **on the refusal** — right remedy placement.

### BLOCKING — `code` is editable in the edit dialog; it is absent from PATCH by design

Line 319. `UpdateLocationDto` carries `name`, `kind`, `ownerConnectionId`, `externalRef`, `status`, `countryIso2`, `postcode`, `latitude`, `longitude` — and no `code`. `architecture-overview.md` § Inventory (#2316) gives the reason: *"it is deliberately absent from the PATCH body, since the natural key is not a patch-shaped field."* `openEdit()` populates `#locCode` and `saveLocation()` recomputes and reassigns it, so as drawn an operator renames a location's natural key. Same class as #3119's `kind`. Render Code read-only (or omit) on edit.

### IMPORTANT — the bootstrap copy implies minting a location fixes routing

Empty state: *"…zero locations exist — **every order is currently unfulfillable. Mint the first one**"*, then the button `alert()`s *"Mints one MAIN warehouse location idempotently."* Nothing says the minted row holds no stock.

#2407 is explicit that this is the misreading to avoid: *"The minted row locates no existing stock — `locationId IS NULL` on a position permanently means the master declines to locate its stock and no minted row is ever a stand-in for that — so the panel's ready state says the location still needs stock assigned rather than implying routing will now succeed."* As written, an operator clicks, the error goes away, and routing still decides nothing.

(The same "every order is currently unfulfillable" claim is also wrong in its own right — per #2407 routing **refuses** the off→on transition at zero locations, so such an install either has routing off, and its orders are fine, or could never have enabled it. See my note on #3135, which ships the same sentence.)

### IMPORTANT — the `PRE-MIGR` "Pre-migration stock pool" fixture models the prohibited NULL stand-in

Line 475: a `virtual` location, `inUse: true`, named *Pre-migration stock pool*. That is exactly the stand-in ADR-058 decision 2 forbids — `locationId IS NULL` permanently means *the master declines to locate its stock*, and no row may represent it. **A fixture in a durable spec teaches by example**, and this one teaches the prohibited pattern to whoever writes the seed data and the e2e fixtures. Rename it to something plainly a real place (a consignment site, a dropship node), or drop it.

### IMPORTANT — the list is paginated server-side; the mockup has no pagination

The endpoint returns `PaginatedLocationsResponseDto` and `ListLocationsQueryDto` carries `page`/`limit`. The mockup renders an unbounded table. `frontend-architecture.md` § Paginated Totals As A Second Stage is the governing convention (`usePaginatedTotal` + `<ListPagination>`, `20+` placeholder, never `0` on a failed count). A five-row mockup can skip the visual, but it should show the pager so the implementer doesn't decide unaided.

### SUGGESTIONS

- `latitude`/`longitude` back a `numeric(9,6)` column but the inputs use `step="0.0001"` (4dp). Separately, `openEdit()` sets both to `''` unconditionally — harmless against a geo-less seed, but as drawn an edit silently clears existing coordinates.
- `ListLocationsQueryDto` also offers `countryIso2` and `codePrefix`; the mockup exposes Kind + status only. Fine as a scope call — worth a note so it reads as chosen rather than forgotten.
- Line 257 carries the same "Fulfilment routing" eyebrow as #3119. Weaker here (locations genuinely serve sourcing), but two pages sharing a kicker with the shipped dispatch tab compounds the collision.

---

- **`data-state` is on the switcher BUTTON only, never on the rendered state** — `setState()` toggles a class and replaces `innerHTML`; nothing stamps `<body>` or the container. The exemplar § UX Mockups names drives every state from `body[data-state="…"]`, which is why an E2E task can *set* a state and assert on it. Fix: `document.body.dataset.state = name`, plus a value + button for every dialog and gallery card.
- **The epics cite only the Artifact URL** — § UX Mockups requires implementation tickets to cite the **file path** and `data-state` values, and the E2E baseline to come from the repo copy *"never from the Artifact URL."*
- **Dark palette is invented while the light palette is faithful.** ~15 `--status-*` dark values diverge from `index.css:405–504`; `--status-disabled-fg` at `oklch(70%)` vs the real `oklch(82%)` is a genuine contrast regression. Same block appears in all four — fix once and propagate.
- **`--overlay-scrim` is never declared**, so hardcoded `rgba(22,32,43,.45)` scrims do not respond to the dark theme (`index.css:319/504` has the real token).

norbert-kulus-blockydevs:
## Addressed @piotrswierzy's review

Pushed two commits (924e683, d0ebbf5) fixing every BLOCKING/IMPORTANT finding plus the cross-cutting ones:

- **BLOCKING — `code` editable in edit dialog**: fixed. `#locCode` is now `readOnly` when editing (with a "Can't be changed after create" hint), editable only on create.
- **IMPORTANT — bootstrap copy implies minting fixes routing**: empty-state copy rewritten — no longer claims "every order is currently unfulfillable" (per #2407 routing refuses the off→on transition at zero locations, so a zero-location install either has routing off or never enabled it), and the bootstrap action's copy now states the minted row holds no stock of its own.
- **IMPORTANT — `PRE-MIGR` fixture models the prohibited NULL stand-in**: renamed to a plainly real place (a consignment site).
- **IMPORTANT — no pagination affordance**: added a static pager matching `PaginatedLocationsResponseDto`, with a code comment noting `countryIso2`/`codePrefix` are a scoped-out choice rather than an oversight.
- **Status/lifecycle affordance gap**: added a Status field to the edit dialog plus direct Retire/Reactivate row actions, so retiring is no longer reachable only via the failed-delete rescue action and there's now a way back to Active.
- **Cross-cutting**: `data-state` now also lands on `<body>`; dark palette now matches `apps/web/src/index.css` status tokens (fixes the `--status-disabled-fg` contrast regression); `--overlay-scrim` declared and used for both themes instead of a hardcoded light-only scrim.
- **Bonus**: found and fixed a pre-existing display bug where `.form-field{display:grid}` was overriding the UA `[hidden]` style, so the duplicate-code error field rendered even when "hidden" — was making the edit dialog look broken.
- **Suggestions taken**: lat/lng step widened to 6dp (`numeric(9,6)`), edit no longer clears existing coordinates, eyebrow changed from "Fulfilment routing" to "Inventory" to stop colliding with #3119's kicker.

Updated artifact: https://claude.ai/code/artifact/fbb5f9e1-52a3-4b7b-8136-ad59dcf4f318

Two independent review passes (e-commerce/OMS domain lens + competitor-tooling lens) are running now against the updated artifact to sanity-check OMS completeness; I'll follow up here if either surfaces anything.

norbert-kulus-blockydevs:
## Two independent reviews complete — findings acted on

Ran the updated artifact through two independent, fresh-context reviews (no shared context with each other or with my earlier fixes):

**1. Senior e-commerce/OMS operator lens** — cross-checked the mockup against the actual shipped backend (`location.types.ts`, `location.service.ts`, `inventory-locations.controller.ts`, DTOs, migration), not just the architecture doc. Verdict: **"good enough as the spec for #3063's CRUD implementation"**, with field names/validation bounds/the `code`-immutable rule/retire-vs-delete semantics/409 copy all near-verbatim matches to the real service. One load-bearing gap found: the mockup's `inUse` boolean (driving the delete-confirm dialog's warning state) has no real backend counterpart — `LocationResponseDto` carries no count field, and the only server-side count (`countPositionsAtLocation`) is buried as free text inside `LocationInUseError`'s 409 message. Flagged as a backend-contract decision #3063 needs to make before the real dialog can be implemented as drawn.

**2. Competitor-tooling lens (BaseLinker/Linnworks/ChannelEngine/Pipe17)** — verdict: **no structural gap in the documented model**; the "location is provenance, not authority" framing is *more* coherent than the competitive baseline (BaseLinker's community has long complained about warehouse-vs-external-warehouse conflation). Confirmed: no location-priority/ranking field belongs on this page (every competitor keeps that on a separate rules/routing screen too — matches OpenLinker's `oms_routing_rules`, #2953). Same `positionCount`/stock-visibility gap surfaced independently. Also flagged: no lead-time/capacity fields anywhere in the routing vocabulary — a genuine gap vs. ChannelEngine/Linnworks/Pipe17's SLA-aware routing, but correctly scoped as a **routing-rules roadmap item (#2953/Wave 4), not a mockup defect** — the mockup doesn't invent fields the schema doesn't have.

### Acted on (pushed in 594d5f6)

- Replaced the demo-only `inUse` boolean with a `positionCount`, rendered as a link ("128 positions") on each row and quoted by number in the delete-refusal copy — with a code comment documenting the real backend gap for whoever implements #3063.
- Added a "viewing as Admin/Operator/Viewer" toggle that disables every write action for non-admins, matching `LocationsController`'s real `@Roles('admin')` on writes vs. `@Roles('admin','operator','viewer')` on reads (the mockup previously implied any viewer could mutate).
- Country now validates to exactly 2 characters on save, matching `@Length(2,2)` (was `maxlength=2`, which still let a 1-char value through that the real DTO would 400 on).

### Deliberately not acted on (correctly-scoped MVP deferrals per both reviewers)

- Bulk CSV import, text search, map-based lat/lng picker, drill-through link to the #2319 duplicate-position diagnostic — all reasonable to defer for a first CRUD slice.
- Lead-time/capacity/cutoff fields — belongs on the routing-rules layer (#2953), not `inventory_locations`.

Latest artifact: https://claude.ai/code/artifact/fbb5f9e1-52a3-4b7b-8136-ad59dcf4f318

norbert-kulus-blockydevs:
Updated the mockup: entry and breadcrumb now route through `/settings`, not Connections.

Both were previously modelled as reached via the connection — sidebar `Connections` highlighted, breadcrumb `Connections / OpenLinker OMS / Inventory locations`, back-link `← OpenLinker OMS`. That's not how this screen is actually entered: `apps/web` reaches `/inventory/locations` only through a link-out tile on `/settings` (mirroring `WhoDecidesTile`'s shape — see `feat(inventory): move locations entry point to a Settings tile`), with no sidebar nav item at all.

Changed:
- Sidebar: `Connections` is no longer shown active.
- Breadcrumb: `Platform / Inventory locations` (2-level, matching the route's actual `{group: 'Platform', title: 'Inventory locations'}` crumb — unchanged from when it briefly had a sidebar entry).
- Back-link: `← Settings`, not `← OpenLinker OMS`.
- Added a state-gallery card previewing the `/settings` tile itself (eyebrow, title, description, "Open" button) so the entry point is reviewable alongside the screen it links to.

piotrswierzy (approved):

**Every BLOCKING and IMPORTANT finding is fixed, with the reason written into the file at the point of the fix.** It is also the only one of the three mockups in this wave that fixed the *shared* blocks — which is why those still read as open on its siblings.

| Prior finding | Verdict |
|---|---|
| `data-state` on switcher button only | ✅ **Resolved** |
| `code` editable in the edit dialog | ✅ **Resolved** |
| Bootstrap copy implied minting fixes routing | ✅ **Resolved**, both halves |
| `PRE-MIGR` modelled the prohibited NULL stand-in | ✅ **Resolved** |
| No pagination | ✅ **Resolved** |
| All four suggestions | ✅ **Resolved** |

**`data-state`** — `setState()` (635) now opens with `document.body.dataset.state = name;`, and the chip selector was correctly narrowed to `.chip[data-state]` so the Kind/Status filter chips are no longer swept by the active-class toggle. That narrowing is the detail that would have bitten later.

**`code`** — `openEdit()` sets `readOnly = true` and swaps the hint to "Can't be changed after create.", with an in-file comment citing `UpdateLocationDto` and *"the natural key is not a patch-shaped field"*. `openCreate()` resets it. The uppercase recompute in `saveLocation()` is now harmless because the field can't change.

**Bootstrap copy** — the false *"every order is currently unfulfillable"* is replaced with the #2407 refusal stated correctly: routing *"can't be turned on while a connection has zero locations, so if you're seeing this, routing here is either off or has never been configured — **no orders are affected**."* And the button now says *"**The minted row holds no stock of its own — it still needs stock assigned before routing can source from it.**"* That is #2407's ready-state sentence verbatim in intent.

**Fixture** — `PRE-MIGR` → `CONSIGN-01` / "Consignment site — Silesia distributor", plainly a real place, so it no longer teaches the ADR-058 decision 2 prohibition by example.

### Cross-cutting: this PR fixed both shared blocks

- **Dark palette** — measured **15/15** `--status-*` dark values now match `index.css:405–504` byte for byte, including `--status-disabled-fg: oklch(82% 0.010 270)`. The contrast regression is gone.
- **`--overlay-scrim`** — declared in all three blocks (`:root`, the media query, and the `[data-theme]` selector — exactly `index.css:319/504`) and consumed by `.dialog-backdrop`. No hardcoded `rgba(22,32,43,…)` remains.

**#3120's corrected dark block and scrim declarations drop into #3119 and #3121 unchanged** — that is the cheapest next action in this wave.

### Outstanding, none blocking

- **Epic #3063 still cites only the Artifact URL** — no `docs/plans/mockups` path, no `data-state` values. That is an issue-body edit, not a file change, and § UX Mockups requires it before the sub-issues are written.
- The 2 dialogs and 5 gallery cards still carry no `data-state` value — the "plus a value for every dialog and gallery card" half. Non-blocking at this surface size.
- **`positionCount` is self-disclosed as not existing** (`LocationResponseDto`/`toDto` carries no positions-count field; `countPositionsAtLocation` surfaces only as free text in the 409). That disclosure is **exactly the right pattern** — and it is also a column an implementer will be asked to render, so name it in #3063 as a backend prerequisite or drop the column.
- The pager shows a concrete "5 of 5". § Paginated Totals As A Second Stage prescribes a `20+` placeholder and never `0` on a failed count — one gallery card showing the degraded total would close it.
- The write-gating note (282, *"writes are admin-only … operator/viewer get a read-only page"*) plus `writeDisabledAttrs` is a good addition that #3121's admin-only page should mirror.

Zero invented token names (68 used), light-palette fidelity holds.

---

Issue #3071, opened 2026-09-10
[EPIC] Frontend — Duplicate stock positions diagnostic page


norbert-kulus-blockydevs:
## Problem / Context

`GET /inventory/duplicate-positions` (#2319) — the read-only readiness gate for the #2325 stricter stock-uniqueness index — has zero hits anywhere in `apps/web/src` (confirmed by #3029). There is no way for an admin to see whether an install is ready.

## Mockup

Reviewed, interactive HTML mockup — built on the real design system, all states: https://claude.ai/code/artifact/9e6e1b2b-275a-4941-b25e-45ab18081a1a

## Proposed Solution

Add a client + a new diagnostic page under Diagnostics, rendering the readiness KPIs (`groupCount`/`rowCount`/`excessRowCount`) and an expandable grouped table of the colliding `inventory_items` rows.

Broken down into the following sub-issues (linked as GitHub sub-issues):

- duplicate-positions API client + query hook
- duplicate-positions page with readiness KPI strip
- expandable grouped table (per-row inventory_items detail)
- truncated-report banner state
- nav entry under Diagnostics
- duplicate-positions page tests

## Classification

**Type**: Frontend
**Layer**: Interface (feature module + route)

## Dependencies

- None — the backend routes this epic wires up are already shipped (see #3029)

## Assumptions

- Backend contract is stable as audited in #3029; no API changes needed
- Follows `apps/web/src/features/*` conventions per `docs/frontend-architecture.md`

## Acceptance Criteria

- [ ] Every sub-issue below is completed and merged
- [ ] The surface is reachable from the product (nav entry, route, or mount point)
- [ ] Manual verification against a running `pnpm start:dev:api` + `pnpm start:dev:web`

Issue #2317, opened 2026-08-22
[TASK] CORE/Worker — inventory ladder step (ii): 'legacy' sentinel backfill as a bounded sweep


piotrswierzy:
> Part of the OMS programme backlog (**Wave 1b**, stream S1, size M). Backlog slug: `W1b-4`. Sibling `W*` slugs resolve via the Wave-1b epic's checklist. Design of record: `docs/plans/analysis/DESIGN-oms-authority-model.md` + ADRs 052–062.

## Problem / Context

ADR-058 decision (3), step (ii): the sentinel backfill is a **batched job**, never a migration — a single-transaction `UPDATE` over the live oversell table is the deploy-blocking shape D2 rejected.

## Proposed Solution

A new job (e.g. `master.inventory.backfillProvenance`) implemented over the existing `runBoundedSweep` shape (`apps/worker/src/sync/bounded-sweep.ts`, #2218/#2219): a per-run budget, a cursor persisted between ticks, and a `SyncLockPort` lock whose TTL covers one RUN. It sets `sourceConnectionId = 'legacy'` on rows where it is NULL, page by page.

Cursor advances only when the page's write succeeded (the #2218 rule). The sweep is self-terminating and reports a completion count so an operator can verify before Wave 1d's cleanliness check. It is scheduled once and disables itself when a page returns zero rows, rather than running forever.

## Classification

**Type**: CORE / Infrastructure
**Layer**: Application + Infrastructure
**File(s)**: `apps/worker/src/sync/`, `libs/core/src/inventory/application/services/`, pattern reference `apps/worker/src/sync/bounded-sweep.ts`

## Dependencies

- #2314 (`W1b-3`)

## Assumptions

- `'legacy'` is a reserved connection-id sentinel that will never collide with a real UUID `Connection.id`. Documented in the entity docblock.
- Re-running the sweep is safe (the predicate is `IS NULL`), so a crash mid-cycle costs a re-scan of at most one page.

## Acceptance Criteria

- [ ] The sweep is budgeted, cursor-resumed and per-run locked (no unbounded enumeration)
- [ ] Cursor does not advance on a failed page
- [ ] A completion report logs remaining-NULL count; reaching zero disables further ticks
- [ ] Integration test seeds >1 page of NULL rows and asserts full backfill across ticks
- [ ] Tests added or updated for non-trivial logic
- [ ] No architecture boundary violations (CORE ↔ Integration)

Issue #2325, opened 2026-08-22
[TECH-DEBT] Infrastructure — inventory ladder step (iii): SET NOT NULL + unique-index recreation (deferred; Wave 1d)


piotrswierzy:
> Part of the OMS programme backlog (**Wave 1d — filed, deliberately NOT scheduled**; stream S1, size M). Backlog slug: `W1d-1`. Sibling `W*` slugs resolve via the Wave-1b epic's checklist. Design of record: `docs/plans/analysis/DESIGN-oms-authority-model.md` + ADRs 052–062.

## Problem / Context

ADR-058 decision (3) step (iii), explicitly **deferred**. The repo's single-transaction migration mode makes `CREATE INDEX CONCURRENTLY` unavailable, so a one-shot recreation holds `ACCESS EXCLUSIVE` on the live oversell table. Until this ships, the #1904 withhold guard stays in force.

## Proposed Solution

Behind a **cleanliness check** (#2319 / `W1b-9` returning zero duplicate groups, and #2317 / `W1b-4` reporting zero remaining NULL `sourceConnectionId`): `SET NOT NULL` on `sourceConnectionId` and recreate both partial unique indexes to include provenance, keying `locationId` via a sentinel/`COALESCE` (the decision ADR-058 defers to this step, because a NULL-distinct nullable column must never enter a unique index). Retire the #1904 withhold guard **for inventory only** (the products-side guard is untouched) once the constraint is enforced.

Requires a documented maintenance-window procedure, since the lock is unavoidable in this mode.

## Classification

**Type**: CORE / Infrastructure
**Layer**: Infrastructure
**File(s)**: `apps/api/src/migrations/`, `libs/core/src/inventory/infrastructure/persistence/entities/inventory-item.orm-entity.ts:29,34`, `libs/core/src/inventory/application/services/master-inventory-sync.service.ts`, `docs/operations/`

## Dependencies

- #2314 (`W1b-3`), #2317 (`W1b-4`), #2320 (`W1b-5`), #2319 (`W1b-9`) — all must be complete and clean

## Assumptions

- No `CONCURRENTLY` escape hatch is introduced for this one migration; if the lock proves unacceptable, the escape hatch is its own design question, filed separately.
- Migration timestamp follows the synthetic-sequential convention (`docs/migrations.md` rule 3).

## Acceptance Criteria

- [ ] Migration refuses (or is documented as pre-gated) unless the duplicate-position pass returns zero and no NULL provenance remains
[...]

PR #3121, opened 2026-09-10
docs(mockups): duplicate-positions UI mockup

### LINKED ISSUE #3114: design(inventory): duplicate-positions UI mockup
## Problem / Context

Before implementation starts, Duplicate stock positions diagnostic page (#3071) needs a reviewed design so every sub-issue builds against

## Proposed Solution

## Classification

**Type**: Frontend
**File(s)**: `docs/plans/mockups/duplicate-positions-3071.html`

## Dependencies

- None — first sub-issue of #3071, everything else in the epic reads from this mockup

## Assumptions

## Acceptance Criteria

- [ ] `docs/plans/mockups/duplicate-positions-3071.html` exists and renders standalone (no external API calls)
- [ ] PR links back to the reviewed Artifact: https://claude.ai/code/artifact/9e6e1b2b-275a-4941-b25e-45ab18081a1a

norbert-kulus-blockydevs:
## Summary
- Adds the reviewed `duplicate-positions` mockup as a static file at `docs/plans/mockups/duplicate-positions-3071.html`
- Reviewed Artifact (live preview): https://claude.ai/code/artifact/9e6e1b2b-275a-4941-b25e-45ab18081a1a
- Addresses #3114, first sub-issue of #3071 — **not tagged `Closes`**: this PR merges into `3071-duplicate-positions-diagnostic-ui`, not `main`, so GitHub won't auto-close #3114 from here (per @piotrswierzy's review); the epic branch's own merge to `main` is what should close it.

## Review round
@piotrswierzy's `/pr-review` found 2 BLOCKING + 2 IMPORTANT issues, all fixed in the latest commit:
- **BLOCKING** — the `clean` state asserted migration safety off a single `groupCount: 0` read, contradicting the runbook's #2317-backfill ordering dependency. Split into `clean-pending`/`clean-ready` states with an explicit two-condition readiness checklist.
- **BLOCKING** — `data-state` lived on the switcher chip only, never on `body`/the rendered content, so an E2E task had no unambiguous way to assert which state rendered. Now driven from `setState()`/`toggleGroup()`.
- **IMPORTANT** — `legacy` / `null` `sourceConnectionId` and `locationId` rendered as unexplained bare text. Added tooltips distinguishing "not yet backfilled" from "master declines to locate."
- **IMPORTANT** — no route to the remediation runbook that holds the survivor rule and the `DELETE`. Added inline links plus a computed "likely survivor" badge per group.

Two independent follow-up domain reviews (an e-commerce/OMS ops-lead persona and a competitor-platform specialist persona) were then run against the fixed mockup to check it against what a real OMS diagnostic needs. Further fixes from that pass:
- **BLOCKING** — no product/SKU identity anywhere, only internal ids. Added `productName`/`sku` display (flagged in-file as mockup enrichment beyond the currently-documented response shape).
- **BLOCKING** — the "likely survivor" badge ignored `reservedQuantity`, so it could point an operator at deleting the one row holding a live order's stock reservation. Added a reservation-risk alert + per-row badge for exactly that case.
- **BLOCKING** — no computed oversell exposure. Added a "live qty at risk" column/detail line showing what the availability read is currently summing vs. what the true figure should be after cleanup.
- **IMPORTANT** — groups weren't ranked by urgency and "showing the largest groups" was asserted copy with no sort behind it. Now genuinely sorted by live-stock exposure; live vs. all-stale groups get distinct framing (active oversell risk vs. migration-only blocker); added an Export CSV affordance.
- **IMPORTANT** — no connection/location display names. Added, with a link out to Connections.

Verified headless (Playwright/Chromium) across every state after each round: no console errors, `data-state`/`aria-expanded` assert correctly, keyboard toggle works, desktop/mobile layouts hold.

## Test plan
- [x] Verified headless across all 6 states (duplicates / clean-pending / clean-ready / truncated / loading / error) — no console errors
- [x] Verified `data-state` / `data-expanded-group` / `aria-expanded` reflect correctly for E2E assertions
- [x] Verified the reservation-risk detail panel and exposure ranking render correctly (desktop + mobile)

### DIFF
--- a/docs/plans/mockups/duplicate-positions-3071.html
+++ b/docs/plans/mockups/duplicate-positions-3071.html
@@ -266,7 +266,15 @@ button:focus-visible, a:focus-visible, [tabindex]:focus-visible{ outline:none; b
 </div>
 
 <script>
-const REMEDIATION_LINK = '../../operations/inventory-duplicate-positions.md';
+// Absolute, not repo-relative: a repo-relative path (../../operations/…)
+// only resolves when this file is served from inside a checkout of the
+// real docs tree. It 404s from the published Artifact preview (a
+// different origin entirely) and from any other standalone hosting of
+// this file, which is otherwise most of how this mockup gets looked at.
+// The absolute GitHub URL is the one link shape that works everywhere —
+// matching the existing #NNN-issue-link precedent already used across
+// this repo's other committed mockups.
+const REMEDIATION_LINK = 'https://github.com/openlinker-project/openlinker/blob/main/docs/operations/inventory-duplicate-positions.md';
 
 // NOTE — mockup enrichment beyond the documented GET /inventory/duplicate-positions
 // contract: `productName`, `sku`, `connectionName` and `locationName` are NOT in
@@ -378,7 +386,7 @@ function findSurvivorId(group) {
 }
 
 function remediationLinkHtml(extra) {
-  return `<p class="alert__link"><a href="${REMEDIATION_LINK}">Remediation guide: how to pick the survivor and delete the rest →</a>${extra ? ' ' + extra : ''}</p>`;
+  return `<p class="alert__link"><a href="${REMEDIATION_LINK}" target="_blank" rel="noopener">Remediation guide: how to pick the survivor and delete the rest →</a>${extra ? ' ' + extra : ''}</p>`;
 }
 
 function exportCsv() {
--- a/docs/plans/mockups/duplicate-positions-3071.html
+++ b/docs/plans/mockups/duplicate-positions-3071.html
@@ -92,14 +92,16 @@ button:focus-visible, a:focus-visible, [tabindex]:focus-visible{ outline:none; b
 .shell-nav__group{ margin-bottom:16px; }
 .shell-nav__label{ margin:0 0 4px; padding:0 8px; font-size:10px; font-weight:600; letter-spacing:.11em; text-transform:uppercase; color:var(--text-muted); }
 .shell-nav__list{ list-style:none; padding:0; margin:0; display:grid; gap:1px; }
-.shell-nav__link{ display:flex; align-items:center; padding:6px 8px; border-radius:6px; color:var(--text-secondary); font-size:13px; font-weight:500; text-decoration:none; }
+.shell-nav__link{ display:flex; align-items:center; gap:6px; padding:6px 8px; border-radius:6px; color:var(--text-secondary); font-size:13px; font-weight:500; text-decoration:none; }
 .shell-nav__link:hover{ background:var(--bg-surface-muted); color:var(--text-primary); }
 .shell-nav__link--active{ background:var(--bg-surface-muted); color:var(--text-primary); font-weight:600; box-shadow:inset 2px 0 0 var(--accent-primary); }
+.shell-nav__admin-tag{ margin-left:auto; font-family:var(--font-mono); font-size:.5625rem; font-weight:600; letter-spacing:.04em; color:var(--text-muted); border:1px solid var(--border-default); border-radius:var(--radius-xs); padding:0 4px; text-transform:uppercase; }
 .shell-topbar{ position:sticky; top:0; z-index:10; display:flex; align-items:center; gap:12px; height:52px; padding:0 16px; background:var(--bg-shell); border-bottom:1px solid var(--border-subtle); }
 .shell-crumbs{ font-size:13px; color:var(--text-secondary); }
 .shell-crumbs b{ color:var(--text-primary); font-weight:500; }
 .shell-content{ padding:20px 32px 48px; max-width:1180px; }
 
+.eyebrow-row{ display:flex; align-items:center; gap:8px; }
 .eyebrow{ margin:0; font-size:.76rem; font-weight:700; letter-spacing:.08em; text-transform:uppercase; color:var(--accent-primary-hover); }
 .page-header{ display:grid; gap:.45rem; margin-bottom:20px; }
 .page-header__row{ display:flex; align-items:start; justify-content:space-between; gap:16px; flex-wrap:wrap; }
@@ -107,6 +109,7 @@ button:focus-visible, a:focus-visible, [tabindex]:focus-visible{ outline:none; b
 .page-description{ color:var(--text-secondary); font-size:.875rem; max-width:64ch; }
 .page-header__actions{ display:flex; gap:8px; flex-shrink:0; align-items:center; }
 .page-header__meta{ font-size:.6875rem; color:var(--text-muted); font-family:var(--font-mono); }
+.page-header__caution{ font-size:.75rem; color:var(--text-muted); max-width:64ch; }
 
 .button{ display:inline-flex; align-items:center; justify-content:center; gap:var(--space-2); border:1px solid var(--accent-primary-border); border-radius:var(--radius-md); padding:0 var(--space-4); height:2rem; background:var(--accent-primary); color:var(--text-on-primary); font-family:inherit; font-size:.8125rem; font-weight:500; cursor:pointer; box-shadow:var(--shadow-xs), var(--shadow-inset-top); }
 .button--secondary{ border-color:var(--border-default); background:var(--bg-surface); color:var(--text-primary); }
@@ -130,14 +133,20 @@ button:focus-visible, a:focus-visible, [tabindex]:focus-visible{ outline:none; b
 .alert--warning{ background:var(--status-warning-soft); border-color:var(--status-warning-border); color:var(--status-warning-fg); }
 .alert--success{ background:var(--status-success-soft); border-color:var(--status-success-border); color:var(--status-success-fg); }
 .alert__glyph{ flex-shrink:0; font-size:1rem; line-height:1.4; }
-.alert__content{ display:grid; gap:2px; }
+.alert__content{ display:grid; gap:2px; flex:1; }
 .alert__title{ font-weight:600; color:var(--text-primary); font-size:.8125rem; }
 .alert__description{ font-size:.8125rem; color:var(--text-secondary); }
+.alert__checklist{ list-style:none; margin:6px 0 0; padding:0; display:grid; gap:3px; }
+.alert__checklist li{ display:flex; align-items:baseline; gap:6px; font-size:.8125rem; color:var(--text-secondary); }
+.alert__checklist .check-yes{ color:var(--status-success-fg); font-weight:700; }
+.alert__checklist .check-pending{ color:var(--status-warning-fg); font-weight:700; }
+.alert__link{ margin-top:6px; font-size:.8125rem; font-weight:500; }
 
 .status-badge{ display:inline-flex; align-items:center; gap:var(--space-2); padding:2px var(--space-2); border:1px solid var(--status-disabled-border); border-radius:var(--radius-sm); background:var(--status-disabled-soft); color:var(--status-disabled-fg); font-family:var(--font-mono); font-size:.6875rem; font-weight:500; text-transform:uppercase; white-space:nowrap; }
 .status-badge__dot{ width:.375rem; height:.375rem; border-radius:999px; background:currentColor; opacity:.75; }
 .status-badge--warning{ border-color:var(--status-warning-border); background:var(--status-warning-soft); color:var(--status-warning-fg); }
 .status-badge--neutral{ border-color:var(--status-disabled-border); background:var(--status-disabled-soft); color:var(--status-disabled-fg); }
+.status-badge--success{ border-color:var(--status-success-border); background:var(--status-success-soft); color:var(--status-success-fg); }
 .status-badge--compact{ padding:1px var(--space-2); font-size:.625rem; }
 
 .data-table__container{ overflow-x:auto; border:1px solid var(--border-subtle); border-radius:var(--radius-lg); background:var(--bg-surface); box-shadow:var(--shadow-xs); }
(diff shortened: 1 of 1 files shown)


norbert-kulus-blockydevs:
## PR Review — #3121 (duplicate-positions mockup)

> **Extended review.** This review was originally posted as a senior-frontend pass; it has been expanded in place with three additional lenses (OpenLinker domain specialist, e-commerce operations specialist, senior backend/systems architect) and consolidated below into one review.

Reviewed the mockup file directly (`docs/plans/mockups/duplicate-positions-3071.html`) against the shipped backend contract (docs/architecture-overview.md § 3 Inventory, "Duplicate-position detection (#2319, ADR-058 ladder step (iii))"), the frontend style guide, and the operational remediation doc it's meant to gate.

### What's correct (worth calling out)

- **No green-badge anti-pattern.** Unlike sibling mockup #3119, this one only shows success styling in the genuinely-clean empty state (0 duplicate groups). Default "duplicates found" state uses warning-orange / neutral-gray badges — never a false-positive green.
- **Read-only-diagnostic framing is explicit and correct.** Page copy states verbatim: *"This is a report, not a repair tool — it detects, it never writes."* No inline merge/delete/fix action exists anywhere — only a `Refresh` button. Matches the backend's admin-only, detects-never-repairs contract precisely.
- **Four-column position key is correctly modeled, not simplified.** The table groups on exactly `productId` / `productVariantId` / `locationId` / `sourceConnectionId` (verified against the actual render function), `null` renders as `— none —` (correct NULL-distinct semantics for "master declines to locate"), and the warning copy explicitly states cross-source rows are "legitimate coexisting mirrors, not duplicates" — ADR-058 decision 2's rule, stated accurately in operator-facing copy.
- **The "gate procedure" framing is present and well-integrated**, not just documented in isolation: the page description calls this "the gate before a stricter uniqueness rule can be turned on," and the clean-state copy pays it off — *"This install can now safely take the four-column unique index... that permanently prevents this class of row from reappearing."* An operator reading only the UI understands why the page exists.
- **A generation timestamp is shown** (`Generated 2026-09-10 14:02 UTC` in the header), correctly treating this as a point-in-time snapshot rather than a live view.
- **Nav placement is correct**: lives under a "Diagnostics" group alongside Jobs & Logs / Webhooks / Cursors — appropriately signals "operational tool," not a everyday merchandiser page.

### Findings

**[IMPORTANT] No link to the operational remediation runbook.** The architecture doc is explicit that "the full procedure, the survivor rule and the remediation `DELETE`" live in `docs/operations/inventory-duplicate-positions.md` — this page is read-only *by design*, so the runbook is the only place an operator can actually act on what they're shown. The warning alert dead-ends at "not ready for the stricter uniqueness rule yet" with no next step. A visible link (e.g. "Remediation guide →" near the warning banner or in the page header) would close the loop the page itself creates.

**[IMPORTANT] The "truncated report" state has no way to reach the excluded groups, and the truncation criterion isn't modeled at all.** The demo's `truncated` state claims *"Showing the 3 largest groups"* out of 47 groups / 121 rows total, but there is no sort logic anywhere in the mockup's JS (checked: the same 3 fixture groups are reused verbatim, with no `.sort()` call) — so "largest" is asserted in copy, not demonstrated. More importantly, there is **no pagination, no "load more," and no export** to reach the other 44 groups: on a real catalog with dozens of duplicates, an operator is shown a hard-capped subset with zero path to the rest. This is a real gap for exactly the deployment size this diagnostic exists for (a healthy install has 0 duplicates and never sees this state at all). Two follow-ups worth deciding before implementation: (a) what should "largest" even mean here — row count, or something more actionable like "groups with the most *live* rows," since a group with many stale rows and one live row is less urgent than the reverse; (b) a CSV/export affordance would let an ops person triage a large duplicate set offline rather than needing SQL access, which the current design implicitly requires once the table is capped.

**[SUGGESTION] Clickable group rows are mouse-only.** `toggleGroup()` is wired via `onclick` on a bare `<tr>` with no `tabindex`, `role="button"`, keydown handler, or `aria-expanded`/`aria-controls`. Fine for a throwaway mockup, but since this expandable-group-row pattern will presumably carry into the real `DataTable`-based implementation, worth flagging now so it isn't copy-pasted forward without a11y wiring.

**[SUGGESTION] No explicit staleness warning tied to the action the runbook implies.** The generation timestamp is shown, but nothing tells an operator "re-run this report immediately before deleting a row" — and since the underlying rows can change between page load and a manual `DELETE` performed elsewhere (per the ops runbook), a stale read followed by a blind delete is exactly the failure mode a diagnostic-only page should guard against in its copy, even though it can't guard against it in code.

**[MINOR] `liveRowCount` / per-row `isStale`/`availableQuantity`/`reservedQuantity` shape isn't confirmed against the documented API contract.** The architecture doc describes the endpoint's grouping key but doesn't specify these enrichment fields in the response shape. Reasonable UX enrichment for a mockup, but implementers should verify the real `GET /inventory/duplicate-positions` response actually carries this before wiring the real page.

**[MINOR] No explicit "admin-only" visual cue on the page itself** beyond nav placement — worth checking whether other admin-gated diagnostic pages in the real app carry a role badge or restriction notice, and if so, matching that convention here for consistency.

Nothing here is a design-direction problem — the domain modeling in this mockup is unusually accurate, and the two IMPORTANT findings (runbook link, truncation dead-end) are both concrete, fixable gaps rather than open questions.

piotrswierzy (changes requested):

**No repair affordance anywhere — this is right, and it is the thing that most needed to be right.** No delete, no merge, no "fix"; the description says *"This is a report, not a repair tool — it detects, it never writes"*, and the eyebrow reads `Inventory · read-only diagnostic`. That matches `InventoryController.getDuplicatePositions`'s own summary in substance.

Verified correct: the four-column grouping key as table columns; cross-source coexistence stated accurately (*"two rows differing only by source connection are legitimate coexisting mirrors, not duplicates"* — ADR-058 decision 2); stale rows **included** with `liveRowCount` shown separately, matching `isStale`'s *"a stale duplicate still occupies the position key and would still collide"*; `excessRowCount = rowCount − groupCount` with *"one row per group survives"*, which is the DTO word for word; and the truncation semantics exactly — `groupCount: 47` / `rowCount: 121` with 3 shown and *"Totals above cover every group in the table — this list is capped for readability only"*, matching *"True when group detail was capped by maxGroups; totals never are."* Nullable `locationId` / `sourceConnectionId` render as `— none —` rather than blank. Token fidelity is exact — zero invented names, every light-palette value matches `index.css`.

### BLOCKING — the `clean` state asserts readiness without the #2317 ordering precondition

Line 357: *"Ready for the stricter uniqueness rule / Zero duplicate position groups exist. This install can now safely take the four-column unique index."*

`docs/operations/inventory-duplicate-positions.md` § Wave 1d gate says the opposite is possible: *"run this **after** the #2317 backfill, because the backfill collapses NULL provenance to `'legacy'` and can therefore turn two rows that were already duplicates into two rows that are visibly duplicates under the same key."*

So a `groupCount: 0` read taken while the backfill is still draining can flip non-zero afterwards — and the operator has been told, **in a green alert**, that the migration is safe. `CREATE UNIQUE INDEX` then fails at deploy, which the runbook calls out as *"not a warning — it is the migration failing at deploy time."* The precondition appears nowhere. The `clean` verdict needs to be a two-condition readiness check, not one number.

### BLOCKING — `data-state` is on the switcher button only, never on the rendered state

`setState(name)` toggles a `chip--active` class and replaces `innerHTML`; nothing writes `data-state` onto `<body>` or the state container. Compare the exemplar the doc itself names — `docs/plans/mockups/analytics-display-currency-picker.html` drives every state from `body[data-state="…"]` (46 occurrences, 15 distinct values), which is why an E2E task can *set* the state and assert on it. Here a spec can only click a chip and cannot assert which state rendered. § UX Mockups requires *"exactly one unambiguous way to find the state."* The expanded-group detail panel has no value at all.

Fix: `document.body.dataset.state = name`, plus a value for the expanded panel.

### IMPORTANT — `legacy` and `— none —` appear with no explanation, and they are the evidence for the blocker above

Group 3 carries `sourceConnectionId: 'legacy'`; group 2 carries `null`. The DTO documents both (*"Null on rows not yet covered by the #2317 backfill"*), and the runbook adds that NULL and `'legacy'` are one class. Rendered as bare mono strings, `legacy` looks like a connection name and `— none —` looks like missing data — when they are in fact the on-screen signal that the backfill has not finished.

### IMPORTANT — no route to the remediation runbook

The runbook holds the survivor rule and the remediation `DELETE`, and the API description ends by naming that file. The mockup compresses the whole survivor rule into a KPI hint and links nothing. For a page whose entire purpose is handing an operator work it cannot do for them, the link out is the most load-bearing element on it — and the expanded detail panel, which already shows `isStale` and `updatedAt` per row, is exactly where the survivor rule wants to be stated.

### SUGGESTIONS

- The error copy says *"try again, or narrow the group cap"*. `maxGroups` is a real query parameter but the mockup exposes no input, so the advice is unactionable.
- This route is `@Roles('admin')`, unlike the locations reads (`admin`/`operator`/`viewer`). A new `Diagnostics` nav item for an admin-only page needs the `requiresRole` treatment (`nav-registry.ts` supports it) — worth showing so the nav sub-issue doesn't ship it to every role. The nav entry itself is also new to `BASE_NAV_GROUPS`; reasonable, just flagging it as a registry addition.
- The `<link>` to `fonts.googleapis.com` differs from `apps/web`, which **self-hosts** IBM Plex woff2 (`index.css:36`). Precedent-consistent with other mockups, but an offline screenshot run falls back to `system-ui` and produces a different baseline.
- `Closes #3114` will not fire against base `3071-duplicate-positions-diagnostic-ui`, and `CLAUDE.md` forbids hand-closing.
- #3055/#3063/#3071 each cite only the Artifact URL in their `## Mockup` section. § UX Mockups requires implementation tickets to cite the **file path** and `data-state` values, and the E2E baseline to come from the repo copy *"never from the Artifact URL — the Artifact can be edited or expire independently of the repo."* Repoint the epics in the same pass, or the durable copy is the one nobody reads.

**This is the strongest of the three in its batch** — its problems are additive copy, not a contradicted contract.

---

norbert-kulus-blockydevs:
Thanks for the thorough review — both BLOCKING issues and both IMPORTANT ones are fixed in [765649197](https://github.com/openlinker-project/openlinker/pull/3121/commits/765649197bf513f6d22e1c36dd18781867a96da5):

- **`clean` state / #2317 ordering dependency** — split into `clean-pending` (warning-toned, states the readiness gate as an explicit two-item checklist: `groupCount: 0` *and* backfill confirmed complete) vs. `clean-ready` (green, both conditions met). No state claims migration safety off one number anymore.
- **`data-state` not on body** — now written by `setState()` (and `data-expanded-group` by `toggleGroup()`), so an E2E task has the single unambiguous assertion point the style guide requires.
- **`legacy`/null with no explanation** — both `sourceConnectionId` and `locationId` nulls now carry distinct tooltips (`title=`) explaining what each means and why `null`/`'legacy'` are grouped together, rather than reading as a connection literally named "legacy" or missing data.
- **No route to the runbook** — added inline links to `../../operations/inventory-duplicate-positions.md` in both the top-level warning banner and the per-group detail panel, plus a computed "likely survivor" badge implementing the runbook's default rule (newest live row) right where the operator needs it.

Also picked up your SUGGESTIONS: the `maxGroups` query param is now a real input in the truncated state (so the error-state retry copy is actionable), a keyboard path (tabindex/role/aria-expanded/keydown) was added to the group rows, and an "Admin only" badge now matches the route's real `@Roles('admin')` gate.

Per your note on `Closes #3114` — updated the PR description to drop the closing keyword and explain why (base is `3071-duplicate-positions-diagnostic-ui`, not `main`). Left the epic-issue mockup-citation cleanup (#3055/#3063/#3071) as a follow-up rather than editing those issues from here.

I then ran two independent domain reviews against the fixed mockup — an e-commerce/OMS ops-lead persona and a competitor-platform specialist persona — specifically to check "does this hold up as a real OMS diagnostic." Both surfaced further gaps that are now fixed too (same commit):

- The "likely survivor" badge never looked at `reservedQuantity` — it could point an operator at deleting the one row holding a live order's stock reservation. Added a reservation-risk alert + per-row badge whenever a non-survivor row carries a reservation.
- No product/SKU identity anywhere, only internal ids — added `productName`/`sku` display (flagged in-file as enrichment beyond today's documented response shape, for implementers to confirm against the real endpoint).
- No computed oversell exposure — added a "live qty at risk" figure (sum of `availableQuantity` across live rows in a group) both as a sortable table column and spelled out in the detail panel, tied to the runbook's "never sum the duplicated quantities" warning.
- "Showing the N largest groups" was asserted copy with nothing sorting behind it — now genuinely ranked by live-stock exposure, and live-vs-all-stale groups get distinct framing (active oversell risk today vs. migration-only blocker).
- Added connection/location display names and an Export CSV affordance for large duplicate sets.

Re-verified headless (Playwright/Chromium) across all 6 states after both rounds — no console errors, `data-state`/`aria-expanded` assert correctly. Artifact preview republished at the same URL.

norbert-kulus-blockydevs:
@piotrswierzy — all fixed, plus two more rounds since my last update here. Full delta against the version you reviewed:

**Your review (both BLOCKING + both IMPORTANT, [765649197](https://github.com/openlinker-project/openlinker/pull/3121/commits/765649197bf513f6d22e1c36dd18781867a96da5)):**
- `clean` state split into `clean-pending` / `clean-ready`, with an explicit two-condition readiness checklist (`groupCount: 0` *and* the #2317 backfill confirmed complete) instead of one number driving a green "safe to migrate" claim.
- `data-state` now written to `body` (and `data-expanded-group` on toggle) by `setState()`/`toggleGroup()`, giving E2E the single unambiguous assertion point.
- `legacy` and `null` `sourceConnectionId`/`locationId` now carry distinct tooltips explaining what each means and why the two are grouped together.
- Added a link out to the runbook (top banner + per-group detail panel) plus a computed "likely survivor" badge implementing the default rule inline.
- Picked up your SUGGESTIONS too: `maxGroups` is a real input now (so the error-state retry copy is actionable), group rows got keyboard access (tabindex/role/aria-expanded/keydown), and an "Admin only" badge matches the route's real `@Roles('admin')`. PR description no longer says `Closes #3114` (explained why in the description).

**Two independent domain reviews** (e-commerce/OMS ops-lead persona + competitor-platform specialist persona), run specifically to check this against what a real OMS diagnostic needs, [765649197](https://github.com/openlinker-project/openlinker/pull/3121/commits/765649197bf513f6d22e1c36dd18781867a96da5):
- The "likely survivor" badge ignored `reservedQuantity` — it could point an operator at deleting the one row holding a live order's stock reservation. Added a reservation-risk alert + per-row badge whenever a non-survivor row carries a reservation.
- No product/SKU identity, only internal ids — added `productName`/`sku` display (flagged in-file as enrichment beyond today's documented response shape).
- No computed oversell exposure — added a "live qty at risk" figure (sum of live rows' `availableQuantity`) as a sortable column and in the detail panel.
- "Showing the N largest groups" was asserted copy with nothing sorting behind it — groups are now genuinely ranked by live-stock exposure; live-vs-all-stale groups get distinct framing (active oversell risk vs. migration-only blocker); added Export CSV.
- Added connection/location display names with a link out to Connections.

**Two more fixes after that, from live use of the Artifact preview:**
- The "Remediation guide" link was repo-relative (`../../operations/…`) — resolved fine from a real checkout but 404'd inside the published Artifact (different origin). Switched to the absolute GitHub blob URL, [f90681d48](https://github.com/openlinker-project/openlinker/pull/3121/commits/f90681d48).
- A bare link straight to the markdown runbook was itself flagged (competitor-review pass) as thin for a real OMS. It now opens a condensed in-page modal first — the same 5-step "Manual remediation" section — with the full doc one click further from inside the modal. Opened from a specific group's row, the modal additionally resolves the survivor and renders a ready-to-copy `DELETE ... WHERE "id" IN (...)` for that group's losers (copy-only, nothing executes), [a7a8154f9](https://github.com/openlinker-project/openlinker/pull/3121/commits/a7a8154f9).

Re-verified headless (Playwright/Chromium) after every round — all 6 states, keyboard nav, modal open/Escape/backdrop-close/focus-restore, mobile viewport — no console errors. Artifact preview is current at the same URL.

piotrswierzy (approved):

**All four prior findings fixed, several beyond what I asked for.** The strongest of the three mockups in this wave, and the one that improved most relative to the ask. Two shared-block items remain, neither blocking.

| Prior finding | Verdict |
|---|---|
| `clean` asserted readiness without the #2317 precondition | ✅ **Resolved**, with the mechanism stated |
| `data-state` on switcher button only | ✅ **Resolved**, most thoroughly of the three |
| `legacy` / `— none —` unexplained | ✅ **Resolved**, better than requested |
| No route to the remediation runbook | ✅ **Resolved** |
| `maxGroups` advice unactionable | ✅ **Resolved** |
| Admin-only nav needs `requiresRole` | ✅ **Resolved** |

**Readiness** — `clean` is split into `clean-pending` / `clean-ready`, each a switcher chip, and the alert now leads with *"Readiness for the #2325 unique index is **two conditions, not one number**"* and explains **why**: *"A `groupCount: 0` read taken while the #2317 provenance backfill is still draining can flip non-zero afterwards, once NULL rows collapse to `'legacy'` and reveal a collision this scan couldn't see yet."* A two-item checklist carries both conditions and the pending arm links the runbook *"(covers the ordering dependency between the backfill and this gate)"*. That is the runbook's § Wave 1d gate stated correctly — and explaining the mechanism rather than just gating on it is more than I asked for.

**`data-state`** — `setState()` writes `document.body.dataset.state`; `toggleGroup()` writes `dataset.expandedGroup`; the remediation panel writes `dataset.remediationModal`. Six values. And `setState` clears `expandedGroup`, so a composed assertion can't go stale — a detail the other two didn't reach.

**The three-way null disambiguation is better than the fix I described.** `sourceConnectionId: null` renders **"— not backfilled —"** (not the ambiguous "— none —") with a tooltip naming the #2317 backfill and stating it is one class with `'legacy'`; `'legacy'` gets its own sentinel tooltip. And `locationId: null` — which I had not separated out — correctly renders **"— unlocated —"** with an ADR-058 decision 2 tooltip: *"Not missing data — treat as a real 'unlocated' value."* The two nulls now read as the different things they are.

**Runbook** — `REMEDIATION_LINK` to `docs/operations/inventory-duplicate-positions.md`, an inline link, an "Open full runbook →" button, and a 5-step panel reproducing the survivor rule in operator language with a per-group computed survivor.

### Outstanding

- **Dark palette — unaddressed.** All 15 `--status-*` dark values still diverge from `index.css:405–504`; `--status-disabled-fg` is `oklch(70%)` vs the real `oklch(82%)`. Identical drift to #3119. **#3120's corrected block drops in unchanged** — mechanical paste, and it is a measured contrast regression.
- **`--overlay-scrim` is N/A here**, correctly — there is no `.dialog-backdrop` or fixed overlay in this file (the remediation "modal" is an in-page panel), so there is nothing to tokenise. Not a miss.
- **Epic #3071 still cites only the Artifact URL.**

### New findings

1. **A comment makes a false claim about my review.** Lines 316–323 disclose `productName`/`sku` as *"mockup enrichment beyond the documented `GET /inventory/duplicate-positions`"* — correct and welcome — but then assert it *"was already accepted as reasonable enrichment in the prior review pass."* It was not; those fields did not exist when I reviewed and I said nothing about them. Drop the sentence; the disclosure stands on its own.
2. **"Live qty at risk" is a second undocumented field with real consequences.** The oversell-exposure figure and the reservation-risk badge are good additions, and the badge correctly refuses to nominate a survivor holding a live `reservedQuantity`. But `DuplicatePositionGroupDto` exposes no quantity or reservation data, so this column **plus the sort** is a backend prerequisite, not a rendering choice. Name it in #3071 alongside `productName`/`sku`, or the sub-issue gets written for an unserveable column — the #3119 dry-run failure mode, one field down.
3. **The survivor badge is a repair recommendation on a page whose thesis is "it detects, it never writes."** The read-only posture is still right and the badge is defensible (it computes what the runbook tells the operator to compute), but it is one step closer to the line, and its correctness now depends on `reservedQuantity` — see (2). Keep it, and keep it beside the runbook link so the authority stays the document.

Zero invented token names (59 used). `Closes #3114` correctly **not** used, with the reason stated in the body — good.

---

Issue #3078, opened 2026-09-10
[EPIC] Frontend — Orphan return matching and operator-authored returns UI


norbert-kulus-blockydevs:
## Problem / Context

`POST /returns/record`, `POST /returns/:returnId/authorize` and `POST /returns/:returnId/match-order` (#2372/#2376) are shipped and have no caller in `apps/web` (confirmed by #3029). A return that names no order OpenLinker recognises blocks every downstream trigger (restock/refund/correction) with no way for an operator to resolve it, and there is no way to record a return by hand for a channel with no returns feed.

## Mockup

Reviewed, interactive HTML mockup — built on the real design system, all states: https://claude.ai/code/artifact/6e97c05b-e518-4393-8917-563fd11acf8c

## Proposed Solution

Add the 3 write methods to `features/returns/api/returns.api.ts`, and a worklist view grouping orphan returns (need matching) separately from operator-authored ones (need approval), each with its own single, clear action.

Broken down into the following sub-issues (linked as GitHub sub-issues):

- write API client (record, authorize, match-order)
- mutation hooks + cache invalidation
- orphan-returns worklist (needs-order / needs-approval groups)
- match-to-order dialog + unknown-order/already-attributed handling
- approve dialog for operator-authored returns
- record-a-return-by-hand dialog
- entry point from the returns list/detail
- orphan/manual returns flow tests

## Classification

**Type**: Frontend
**Layer**: Interface (feature module + route)

## Dependencies

- None — the backend routes this epic wires up are already shipped (see #3029)

## Assumptions

- Backend contract is stable as audited in #3029; no API changes needed
- Follows `apps/web/src/features/*` conventions per `docs/frontend-architecture.md`

## Acceptance Criteria

- [ ] Every sub-issue below is completed and merged
- [ ] The surface is reachable from the product (nav entry, route, or mount point)
- [ ] Manual verification against a running `pnpm start:dev:api` + `pnpm start:dev:web`

PR #3122, opened 2026-09-10
docs(mockups): orphan-returns UI mockup

### LINKED ISSUE #3115: design(returns): orphan-returns UI mockup
## Problem / Context

Before implementation starts, Orphan return matching and operator-authored returns UI (#3078) needs a reviewed design so every sub-issue builds against

## Proposed Solution

## Classification

**Type**: Frontend
**File(s)**: `docs/plans/mockups/orphan-returns-3078.html`

## Dependencies

- None — first sub-issue of #3078, everything else in the epic reads from this mockup

## Assumptions

## Acceptance Criteria

- [ ] `docs/plans/mockups/orphan-returns-3078.html` exists and renders standalone (no external API calls)
- [ ] PR links back to the reviewed Artifact: https://claude.ai/code/artifact/6e97c05b-e518-4393-8917-563fd11acf8c

norbert-kulus-blockydevs:
## Summary
- Adds the reviewed `orphan-returns` mockup as a static file at `docs/plans/mockups/orphan-returns-3078.html`
- Reviewed Artifact (original, round 1): https://claude.ai/code/artifact/6e97c05b-e518-4393-8917-563fd11acf8c
- Updated Artifact (round 2, after review fixes — live preview): https://claude.ai/code/artifact/a18a604b-8338-4232-a228-aa7964220d0b
- Fixes #3115, first sub-issue of #3078 — **not** auto-closeable here since this PR's base (`3078-orphan-manual-returns-ui`) is a feature branch, not `main`; #3115 should be closed once the parent branch merges to `main`, per `CLAUDE.md`'s "never close an issue manually before the PR is merged."

Addressed piotrswierzy's review (all 3 BLOCKING + the IMPORTANT finding), plus two independent follow-up reviews (e-commerce/OMS-domain lens + competitor-UX lens) run against the fixed file — see the review-response comment below for the full breakdown.

## Test plan
- [ ] Use the "Mockup — jump to a state" toolbar (fixed at the bottom) to inspect all 11 states directly, including both 409 outcomes and the authorize refusal

### DIFF
--- a/docs/plans/mockups/orphan-returns-3078.html
+++ b/docs/plans/mockups/orphan-returns-3078.html
@@ -283,15 +283,25 @@ const ORDER_INDEX = [
 // (an operator, or the returns.orphan.reconcile sweep) attributed it first.
 const ALREADY_ATTRIBUTED_ORDER_ID = 'ol_order_5c810fd2e611';
 
-let ORPHANS = [
+// The FROZEN baseline. `ORPHANS` / `PENDING_APPROVAL` below are the live,
+// mutable working copies an organic interaction (matching a card, approving
+// one) is allowed to change — but every toolbar jump resets from THESE, so
+// which state you land on never depends on what you clicked before (a
+// mockup-QA requirement: exactly one way to reach a named state).
+const INITIAL_ORPHANS = [
   { id:'r1', label:'AL-RET-88213', channel:'Allegro', itemsLabel:'Wireless earbuds, phone case', value:'268,00 PLN', daysAgo:3, since:'Sep 8',
     cause:"OpenLinker hasn't ingested that order yet" },
   { id:'r2', label:'AL-RET-88250', channel:'Allegro', itemsLabel:'Running shoes — size 42', value:'249,00 PLN', daysAgo:9, since:'Sep 2',
     cause:"No order with that number here — might be under a different connection" },
 ];
-let PENDING_APPROVAL = [
+const INITIAL_PENDING_APPROVAL = [
   { id:'r3', label:'PrestaShop return', sub:'1 item, recorded by you on Sep 10', daysAgo:1 },
 ];
+
+function cloneFixture(list) { return list.map((r) => ({ ...r })); }
+
+let ORPHANS = cloneFixture(INITIAL_ORPHANS);
+let PENDING_APPROVAL = cloneFixture(INITIAL_PENDING_APPROVAL);
 
 function escapeHtml(s){ return String(s).replace(/[&<>"']/g,(c)=>({'&':'&amp;','<':'&lt;','>':'&gt;','"':'&quot;',"'":'&#39;'}[c])); }
 
@@ -655,6 +665,18 @@ function setState(name) {
   document.querySelectorAll('.demo-toolbar .chip[data-chip]').forEach(c => c.classList.toggle('chip--active', c.dataset.chip === name));
   closeAllDialogs();
   document.getElementById('toastSlot').innerHTML = '';
+
+  // Reset the working fixtures from the frozen baseline on EVERY jump, so a
+  // named state is reachable exactly one way, regardless of what an earlier
+  // dialog/organic interaction (matching, approving) mutated. `all-done`
+  // additionally empties both — it's the one state whose whole point is
+  // "nothing left", so it must not depend on having cleared rows by hand.
+  // The background group mode resets to the full default too, unless this
+  // jump IS a group-mode chip — otherwise "match-open" could render a
+  // different background list depending on which chip was clicked earlier.
+  ORPHANS = name === 'all-done' ? [] : cloneFixture(INITIAL_ORPHANS);
+  PENDING_APPROVAL = name === 'all-done' ? [] : cloneFixture(INITIAL_PENDING_APPROVAL);
+  currentGroupMode = 'both-groups';
 
   if (['both-groups', 'orphans-only', 'pending-only', 'all-done'].includes(name)) {
     currentGroupMode = name;
--- a/docs/plans/mockups/orphan-returns-3078.html
+++ b/docs/plans/mockups/orphan-returns-3078.html
@@ -20,6 +20,8 @@
   --shadow-soft: 0 1px 2px rgb(15 18 26 / .04), 0 6px 16px rgb(15 18 26 / .06);
   --shadow-focus: 0 0 0 3px var(--accent-ring);
   --shadow-overlay: 0 24px 48px rgb(15 18 26 / .16), 0 8px 16px rgb(15 18 26 / .08);
+  /* Faithful to index.css:319 — used for every dialog-backdrop scrim so it responds to theme. */
+  --overlay-scrim: rgb(15 18 26 / 0.55);
   --font-sans: 'IBM Plex Sans', ui-sans-serif, system-ui, sans-serif;
   --font-mono: 'IBM Plex Mono', ui-monospace, SFMono-Regular, Menlo, monospace;
 }
@@ -32,13 +34,15 @@
     --text-muted: oklch(60% 0.012 270); --text-disabled: oklch(42% 0.012 270); --text-on-primary: oklch(16% 0.012 50);
     --accent-primary: oklch(72% 0.18 50); --accent-primary-hover: oklch(78% 0.18 50);
     --accent-primary-soft: oklch(28% 0.08 50); --accent-primary-border: oklch(40% 0.14 55); --accent-ring: oklch(72% 0.18 50 / 0.40);
-    --status-success-soft: oklch(28% 0.07 150); --status-success-border: oklch(40% 0.10 150); --status-success-fg: oklch(80% 0.12 150);
-    --status-warning-soft: oklch(30% 0.08 85); --status-warning-border: oklch(45% 0.11 85); --status-warning-fg: oklch(85% 0.12 85);
-    --status-error-soft: oklch(28% 0.09 25); --status-error-border: oklch(42% 0.13 25); --status-error-fg: oklch(84% 0.14 25);
-    --status-info-soft: oklch(28% 0.06 245); --status-info-border: oklch(42% 0.10 245); --status-info-fg: oklch(82% 0.10 245);
-    --status-disabled-soft: oklch(26% 0.006 270); --status-disabled-border: oklch(36% 0.010 270); --status-disabled-fg: oklch(70% 0.010 270);
+    --status-success: oklch(68% 0.16 150); --status-success-soft: oklch(26% 0.06 150); --status-success-border: oklch(36% 0.10 150); --status-success-fg: oklch(86% 0.14 150);
+    --status-warning: oklch(78% 0.16 85); --status-warning-soft: oklch(28% 0.07 85); --status-warning-border: oklch(38% 0.10 85); --status-warning-fg: oklch(88% 0.14 85);
+    --status-error: oklch(70% 0.18 25); --status-error-soft: oklch(28% 0.10 25); --status-error-border: oklch(40% 0.14 25); --status-error-fg: oklch(86% 0.14 25);
+    --status-info: oklch(72% 0.14 245); --status-info-soft: oklch(26% 0.06 245); --status-info-border: oklch(36% 0.10 245); --status-info-fg: oklch(86% 0.12 245);
+    --status-disabled-soft: oklch(24% 0.010 270); --status-disabled-border: oklch(34% 0.012 270); --status-disabled-fg: oklch(82% 0.010 270);
     --shadow-soft: 0 1px 2px rgb(0 0 0 / .4), 0 8px 24px rgb(0 0 0 / .4);
     --shadow-xs: 0 1px 2px rgb(0 0 0 / .4);
+    --shadow-overlay: 0 24px 48px rgb(0 0 0 / .60), 0 8px 16px rgb(0 0 0 / .40);
+    --overlay-scrim: rgb(0 0 0 / 0.72);
   }
 }
 :root[data-theme='dark'] {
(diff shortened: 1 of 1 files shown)


norbert-kulus-blockydevs:
> **Extended review.** The pass below folds the original frontend/UX review (posted 2026-09-10) together with three additional lenses — an OpenLinker domain specialist, an e-commerce operations specialist, and a senior backend/system architect — into a single consolidated pass, rather than stacking separate comments.

## PR Review — #3122 (orphan-returns UI mockup)

Reviewed as a **frontend/UX artifact** (`docs/plans/mockups/orphan-returns-3078.html`), cross-checked against the shipped returns backend (§22 Returns / #2332 orphan bucket, #2372 authorize+record+match, `libs/core/src/returns/**`).

### Things this mockup gets right
- **No green-tint-every-row anti-pattern.** Unlike sibling #3119, `--status-success` is used only transiently on the post-action confirmation toast, never as a persistent row/card color — exactly the #2507 rule ("color marks the exception, not the normal case").
- **Irreversibility is correctly communicated.** The match dialog carries an explicit "This can't be undone — double-check before confirming" warning, matching the real backend: `matchOrphanToOrder`'s attribution write is a monotonic `claimAttribution` — an operator can fill the value in and never change one. Correctly, there is no unmatch/undo affordance anywhere in the mock.
- The "unknown order" inline validation error (`#zzz999`) roughly mirrors the real `ReturnMatchRefusedError` 400 `unknown-order` outcome.

### Findings

**1) IMPORTANT — "Record a return by hand" is missing required fields the backend mandates.**
`RecordReturnInput` requires `sourceConnectionId` (a `uuid NOT NULL` column — "the row must name a channel") and, per line, a `reason: RefundReason` (closed enum) plus `quantityAdvised` (positive whole units, refused otherwise). The mock's dialog collects only "Which order?" and free-text "What came back?" — no connection/channel picker, no reason selector, no quantity field. As drawn, this form cannot actually call `recordReturn`.

**2) IMPORTANT — the Approve flow implies enforcement that doesn't exist today.**
Copy says returns "wait for your OK" and approving "lets OpenLinker start processing it." Grepping the actual custody/refund services confirms `authorizedAt` is never read as a precondition anywhere — it's a pure audit stamp by the service's own docblock ("the row is the AUDIT RECORD of the operator's act, not a request awaiting an answer"). Only order-attribution (`assertAttributedForTrigger`) actually gates restock/refund/correction today. If this UX is the intended future behavior, a gate needs to be built alongside it; if it describes current behavior, the copy overclaims.

**3) IMPORTANT (new) — the order-lookup input asks for the wrong kind of identifier, and the mock never shows how the gap is bridged.**
Both the match dialog ("Order number or ID", placeholder `#9F2E1A`) and the record dialog present a single free-text field that reads as an *external-facing* order number. But `matchOrphanToOrder` takes an `internalOrderId` and resolves it through `identifierMapping.getExternalIds(CORE_ENTITY_TYPE.Order, internalOrderId)` — i.e. the API needs OpenLinker's own internal id, not a marketplace/shop order number an operator would recognize. As drawn there is no search/autocomplete/resolve step translating what an operator would actually type (an order number, a buyer name) into that internal id. This isn't a cosmetic gap — a real implementation needs an order-search endpoint behind this field, which the mock doesn't acknowledge at all.

**4) New — the mock only demonstrates ONE of the two documented match-refusal outcomes.** `confirmMatch()`'s only error path is the 400 `unknown-order` case (`#zzz999`). The real API also has a distinct 409 `already-attributed` outcome — reached both by a pre-check *and* by losing the race to a concurrent writer (the background `returns.orphan.reconcile` sweep, which runs every 30 minutes, or a peer operator). Since the docblock explicitly calls this race out as a normal, expected outcome ("that is the same fact as the guard's: the return is attributed"), the mock should show it as a distinct, non-alarming state rather than leaving it unrepresented.

**5) New (e-commerce ops lens) — the mock gives an operator no way to diagnose *why* a return orphaned, and no urgency signal.** Real returns-ops tooling (Loop, Narvar, Returnly) surfaces both. Concretely: (a) each card shows only channel + arrival date ("From Allegro · 2 items · came in Sep 8") with no hint at cause (order not yet ingested vs. ingested under a different connection vs. a buyer typo on the order number) — an operator has to guess what to search for; (b) there's no search-by-buyer-name/email/SKU fallback if the order number search fails, which is the realistic failure mode once (3) above is addressed; (c) there's no aging/SLA indicator — a return sitting orphaned for 30 days is a different urgency than one from yesterday, and the mock's two demo rows (Sep 8, Sep 9) give no visual signal that age should matter.

**6) Minor / non-blocking (carried over):**
- The mock doesn't represent the background `returns.orphan.reconcile` sweep auto-attributing a return between page loads — acceptable to omit from a static mock, but combined with finding (4), it means the "surprise disappearance" and "already-attributed" cases are both entirely unrepresented, which is worth a single follow-up pass together.
- Per-card copy doesn't enumerate which downstream triggers (restock/refund/invoice-correction/decline) are blocked — only the page-level lede does.

The hard UX judgment calls the original review praised (irreversibility framing, avoiding the color anti-pattern, distinguishing orphan-match from operator-authorize) still hold up under the additional passes. What the extended review adds is that the mock's two text-input dialogs are under-specified in a way that goes beyond missing fields: the match/record inputs imply a search mechanism that doesn't exist in the backend contract as drawn, and the demonstrated error surface covers only the easier of two real refusal outcomes. Worth a follow-up pass — covering (1)/(2) from the original review plus (3)/(4)/(5) here — before this becomes the spec an implementer works from.

piotrswierzy (changes requested):

The hardest thing is right: **the irreversibility warning is rendered in operator language, in the dialog rather than after it** (`⚠ This can't be undone — double-check before confirming.`), which is `product-spec-oms-returns-operator-ux.md`'s requirement met correctly. The lede states the downstream-trigger block accurately in one sentence (*"Nothing happens with a return — no restock, no refund, no invoice fix — until it's linked to an order"*), which is exactly `assertAttributedForTrigger`'s four triggers. The two groups are correctly separated by **cause** (needs an order / needs your OK) rather than merged into one queue.

### BLOCKING — no `data-state` and no state switcher at all

Unlike sibling #3123, which does this correctly. This mockup demonstrates at least seven distinct states (orphan list, pending-approval, both groups, all-caught-up, match dialog, match error, authorize dialog, record dialog, post-action toast) and **every one is reachable only by clicking through the flow** — `all-done` requires clearing three fixtures by hand. Copy #3123's `.demo-toolbar` shape.

### BLOCKING — the `already-attributed` 409 is not modelled at all

`POST /returns/:returnId/match-order` distinguishes two refusals with different remedies and different codes: `unknown-order` → **400** (the return is fine; the operator's id names nothing OL minted) and `already-attributed` → **409** (`claimAttribution` is `WHERE "internalOrderId" IS NULL`, so a lost race reports `'already-attributed'`, never a failure).

The mockup renders only the first, as an inline field hint, and has **no representation of the second** — so an operator who loses the race sees the mockup's success toast, the opposite of what the endpoint answers. Since attribution is **monotonic with no unmatch**, that concurrent-loss state is precisely the one an implementer must get right.

### BLOCKING — the authorize refusal is undepictable

`POST /returns/:returnId/authorize` is restricted to `origin: 'operator_authored'` and **refuses** a `source_ingested` return with `ReturnAuthorizeRefusedError`, reason `'source-ingested'`. The mockup renders Approve only for the operator-authored group and never depicts the refusal — so the surface it specifies cannot tell an operator *why nothing happened* when they click Approve on a marketplace return. The architecture note is explicit that this is the point of the named error: *"an operator clicking Authorize on a marketplace return must learn why nothing happened."*

(The Allegro orphan cards carrying no Approve affordance is the right **default** — but the refusal still needs a state, because the API ships one.)

### IMPORTANT — the record-by-hand dialog cannot satisfy its endpoint

`saveRecord()` (~262–270). `recordReturn` requires a `sourceConnectionId` (the column is `uuid NOT NULL`) and it is **validated, not derived** — the mapping on that connection is what supplies `externalOrderId`. The dialog collects only an order id and an item name. Also worth stating: calling it twice creates two returns (core deliberately synthesises no natural key), which this dialog's copy implies is impossible.

### SUGGESTIONS

- `escapeHtml(r.label)` is interpolated into an `onclick="openMatch('…')"` attribute string; a label containing `&#39;` would break the handler. Harmless with static fixtures, but the pattern reads as safe and isn't.
- `--radius-pill:9999px` is declared and then `999px` written literally in the same file.

---
### Cross-cutting (this mockup wave)

- **`data-state` on the rendered state, not just the switcher** — `document.body.dataset.state = name`, plus a value + button for every dialog and gallery card. § UX Mockups requires exactly one unambiguous way to find a state.
- **`--overlay-scrim` is never declared** — hardcoded `rgba(22,32,43,.45)` scrims do not respond to the dark theme (`index.css:319/504` has the real token).
- `Closes #N` will not fire against a feature-branch base; `CLAUDE.md` forbids hand-closing.

norbert-kulus-blockydevs:
## Response to review (5e0535a0e)

Addressed piotrswierzy's review point by point, then commissioned two independent follow-up reviews of the fixed file (an e-commerce/OMS-domain read and a competitor-UX read) to check nothing else was missing.

### BLOCKING (all fixed)
- **No `data-state`/switcher** — added a fixed demo toolbar (always above the dialog backdrop's z-index, so it stays reachable even with a dialog open) with 11 chips covering every state you listed plus the two 409 outcomes: `both-groups`, `orphans-only`, `pending-only`, `all-done`, `toast-matched`, `match-open`, `match-error` (400), `match-conflict` (409 already-attributed), `authorize-open`, `authorize-refused` (409 source-ingested), `record-open`. `document.body.dataset.state` reflects the rendered state.
- **`already-attributed` 409 unmodelled** — now a distinct, calm/informational state (blue, not red), reachable directly via the toolbar or organically by resolving a specific fixture order that simulates the lost race.
- **Authorize refusal undepictable** — added, reachable via toolbar (there's no card path to it today since only operator-authored returns get an Approve button, same as before — the state exists so the refusal isn't undepictable once/if a future surface allows mixed-selection approval).

### IMPORTANT (fixed)
- Record dialog now collects a real connection picker (backing `sourceConnectionId`), a `RefundReason` selector, and a quantity field, and the order field resolves through a typeahead against an order index rather than accepting a typed order number — same fix applied to the match dialog, since `matchOrphanToOrder`/`recordReturn` both take an `internalOrderId`, never free text.

### Suggestions (fixed)
- Escaped label text is no longer interpolated into `onclick` attribute strings — switched to `data-*` attributes + delegated click listeners.
- `--radius-pill` is now actually used (age badges, toolbar chips) rather than declared-and-dead.
- Dark palette + added `--overlay-scrim` now copied verbatim from `apps/web/src/index.css` instead of approximated.

### Two independent follow-up reviews on the fixed file

Both converged on one more real bug I'd introduced while addressing the above: the "waiting for your OK" group description and the post-approve toast both claimed authorize *gates* restock/refund, contradicting the (correct) framing already in the authorize dialog itself, and the actual backend (`authorizedAt` is a pure audit stamp nothing else waits on). Fixed — all three surfaces now agree it's "audit stamp, not a gate."

Also added in response to both: order total/date + the return's own item & value shown together before an irreversible write (there was previously nothing to actually check the match against), and a cross-channel mismatch warning when the picked order's channel differs from the return's own (both dialogs).

Deliberately **not** added, since the backend doesn't support them today and a mockup shouldn't promise a capability that doesn't exist: bulk match/approve, queue-scale search/pagination/sort, multi-line record submission, and a "give up on this orphan" close path. Flagging these as real gaps worth a follow-up issue against #3078, not silently dropping them.

piotrswierzy (changes requested):

Good progress (+277 → +696) — but two things hold it.

**Resolved:** the `already-attributed` 409 and the `'source-ingested'` authorize refusal are now modelled, and the record-by-hand dialog's `sourceConnectionId` requirement is addressed. Those were the three domain-accuracy blockers and they are genuinely closed.

### Still blocking

**`data-state` is on the chip, not the rendered surface.** A switcher now exists — real progress — but the attribute is written to the switcher `<button>`, so a selector like `[data-state="all-caught-up"]` matches a button that is present in *every* state. The one-line fix is `document.body.dataset.state = name` inside `setState`.

**The state set the switcher advertises is not reproducible.** `setState` does not reset the fixtures, so reaching `all-caught-up` still depends on having cleared rows by hand in a previous interaction — click the chips in a different order and you get a different screen than the slug names. That is the same defect the slug was added to remove.

**There is no results container on the list surface.** The outcome switcher is mounted, but the success/refusal renderers write through `?.insertAdjacentHTML` into a slot that does not exist there — so the three advertised responses render nothing. Specifying the refusal UX is much of this mockup's purpose.

### Cross-cutting

- **Dark palette** — still diverges from `index.css:405–504`. **#3120's corrected block drops in unchanged**; `--status-disabled-fg` at `oklch(70%)` vs the real `oklch(82%)` is a measured contrast regression.
- **`--overlay-scrim`** still undeclared, so the dialog scrims do not respond to the dark theme.
- `Closes #3115` against base `3078-orphan-manual-returns-ui` will not fire; `CLAUDE.md` forbids hand-closing.

The irreversibility warning — rendered in operator language, in the dialog rather than after it — remains the strongest thing here, and the two-group split by *cause* rather than one merged queue is still right.

---

norbert-kulus-blockydevs:
## Response to re-review (5e0535a0e)

### Fixed: the reproducibility regression

`renderGroups('all-done')` fell through to summing the live `ORPHANS`/`PENDING_APPROVAL` arrays (only `'orphans-only'`/`'pending-only'` short-circuited the total), so the "All caught up" chip only actually showed the empty state after an operator had cleared every row by hand in an earlier interaction — exactly the non-reproducibility the toolbar exists to remove. Fixed in `de7e82e81`: the two fixtures are now frozen as `INITIAL_ORPHANS`/`INITIAL_PENDING_APPROVAL`, and `setState()` resets the live working copies (plus `currentGroupMode`) from that baseline on **every** call, `all-done` additionally zeroing both. Verified with Playwright against three scenarios: fresh-load → `all-done` directly, `all-done` after 6 unrelated prior clicks in a scrambled order, and organically matching a card away then jumping back to `both-groups` (which now correctly re-shows both original orphans rather than the depleted list).

### The other three points — checked against the actual commit, not reproducible

I re-verified each one directly against `5e0535a0e` (the reviewed commit) before touching anything, since two of these read like they may be carried over from the *first* review rather than re-checked against the fix that already landed for it. Quoting the exact lines so this is easy to re-run:

- **"`data-state` is on the chip, not the rendered surface."** `setState()` does `document.body.dataset.state = name` (line 654 in that commit) — the chip buttons themselves carry `data-chip`, not `data-state`, and never did. Checked with Playwright across all 11 states: `document.body.dataset.state` matches every clicked chip's name, and `chip.hasAttribute('data-state')` is `false` for all of them.
- **"There is no results container… `?.insertAdjacentHTML` into a slot that does not exist."** `grep -n insertAdjacentHTML` on that commit's file returns zero matches — the file only ever uses `.innerHTML =` against `#matchDialogBody` / `#authorizeDialogBody` / `#groupsSlot`, all three of which exist in the static markup. Playwright confirms `#matchDialogBody` renders 914 chars on `match-error` and `#authorizeDialogBody` renders 398 chars on `authorize-refused`.
- **"`--status-disabled-fg` at `oklch(70%)` vs the real `oklch(82%)`."** That commit's dark blocks both read `--status-disabled-fg: oklch(82% 0.010 270)` (matching `index.css`'s dark block exactly) — `70%` doesn't appear anywhere in the file. `--overlay-scrim` is likewise declared in all three needed places (`:root`, the `prefers-color-scheme` media query, and `[data-theme='dark']`) and is what `.dialog-backdrop`'s `background` actually reads.

None of the above needed a change since they already matched. Happy to be pointed at a specific selector/line if a fresh pass still disagrees — but as checked, the reviewed commit already satisfies them.

norbert-kulus-blockydevs:
## Update: mockup rebuilt in situ, with real app chrome + navigation (79c550c7e)

Rebuilt the mockup around the real `apps/web/src/app/app-shell.tsx` structure instead of a bare content fragment — a persistent sidebar built from the real `nav-registry.ts` groups (Returns sits in Operations, between Fulfilment and Automations, correctly with no count badge — matches the `#2334`-cited reason in that file verbatim), a sticky topbar with breadcrumbs/search/alerts/user-chip, and a mobile drawer. Every chrome class/token (`.shell*`, `.page-*`, `.metric-card`/`.returns-segment*`, `.data-table*`, `.status-badge*`, `.alert*`, `.key-value-list*`, `.button*`) is copied verbatim from `index.css` rather than approximated.

The content is now two navigable pages mirroring the real route split (`apps/web/src/app/routes/returns.route.tsx`):
- **List (`/returns`)** — the real 7-card segment strip (Orphans + All returns/All open backed by fixture rows; the other four segments show a plausible count with an honest "not modeled in this mockup" note rather than fabricated data) and the real column set (Return/Order/Source/Opened/Source status/Stage). "+ Record a return" moved here as a page-level create action.
- **Detail (`/returns/:returnId`)**, reached by clicking a row — the orphan case shows an **actionable** version of the shipped `ReturnOrphanBanner` (today's version explicitly renders no button, per its own docblock) with "Match to an order" inline; the operator-authored/pending case gets an "Approve" action next to where `authorizedAt` already renders read-only today.

All match/authorize/record dialog logic, the 400/409 states, the cross-channel warning and the reconcile-sweep note carried over unchanged, just re-triggered from the new structure. Found one more real bug while doing this: the fixed QA toolbar could overlap and swallow clicks on the record dialog's action row at a short viewport — fixed with bottom padding on the dialog backdrop.

Verified with Playwright: nav-driven click-through (sidebar → list → row → detail → back), all 11 toolbar states, every dialog flow end-to-end, and light/dark/mobile(400px) screenshots.

Artifact updated (same link, version 3): https://claude.ai/code/artifact/a18a604b-8338-4232-a228-aa7964220d0b

piotrswierzy (approved):
## Tech Lead review

Re-reviewed `79c550c7e` (the in-situ rebuild, pushed after the re-review). **All three remaining blockers are verified fixed**, and the rebuild's new claims — app chrome, nav placement, route split — hold up against the real source rather than being asserted.

**Verified fixed**
- `data-state` on the rendered surface — `setState()` (L864) opens with `document.body.dataset.state = name`; the switcher buttons now carry `data-chip`, so `[data-state="…"]` can no longer match a button present in every state. Exactly the one-line fix the re-review asked for.
- Reproducibility — `setState()` resets `RETURNS = cloneFixture(INITIAL_RETURNS)`, `currentSegment`, `currentPage` and closes all dialogs on **every** call, before the switch. Jumping to any chip from any prior interaction order now lands on the same screen.
- The missing results container — `renderMatchError()` (L747) and `renderMatchConflict()` (L762) write into `#matchDialogBody`, a real element, and both are reached only after `renderCurrentPage()` + `openDialog()` in the switcher (L906-916). The three advertised responses render.
- Dark `--status-*` — diffed all 28 declared values against the `html[data-theme='dark']` block of `apps/web/src/index.css`: **zero divergence**, including `--status-disabled-fg` now at the real `oklch(82% 0.010 270)`.
- `--overlay-scrim` — declared in all three scopes and read by the backdrops.

**Independently verified against source**
- Nav placement: `nav-registry.ts:45/48/56` — Fulfilment, Returns, Automations in that order, Returns with no count badge. The mockup's sidebar matches.
- Route split: `returns.route.tsx:29/32/40` — `returns` index + `:returnId`. The list/detail pages mirror it.
- Both refusals: `returns.service.ts:345/374` raise `'already-attributed'` and `:360` raises `'unknown-order'`; `return-authorize.service.ts:102` raises `'source-ingested'`. The controller (`return-writes.controller.ts:208/212`) maps them 400/409 exactly as the mockup depicts, and the 409 is correctly rendered calm/informational rather than as an error.
- No undeclared CSS custom property: 78 declared, 61 used, empty difference. (The `#9F2E1A`-style strings are fixture order numbers, not colours.)

### Findings
**[IMPORTANT]** — epic #3078 body
> The epic still cites only the Artifact URL (`https://claude.ai/code/artifact/6e97c05b-…`) and does not name `docs/plans/mockups/orphan-returns-3078.html`. `docs/frontend-architecture.md` § UX Mockups is explicit that implementation tickets cite the **file path** and the `data-state` value(s) they implement, and that the E2E baseline comes from the repo copy *"never from the Artifact URL — the Artifact can be edited or expire independently of the repo."* Committing the file without repointing the epic leaves nothing pointing at it, which is most of the value of this PR. #3055 was repointed as part of #3119's round and has the shape to copy — a `## Mockup` section naming the path, plus a `### data-state values` list. The 11 values are already stable here, so it is a body edit.

**[SUGGESTION]** — PR body
> `Fixes #3115` will not fire against base `3078-orphan-manual-returns-ui`, which the body correctly acknowledges. Since the keyword is inert here and would become live if the base ever changed, consider writing it as plain `#3115` with the same explanatory sentence — it removes the one way this could auto-close early, per `CLAUDE.md`'s never-close-before-merge rule.

### Verdict
✅ **Approve** — the IMPORTANT item is an epic-body edit outside this diff, not a change to the mockup.

---

Issue #3087, opened 2026-09-10
[EPIC] Frontend — Return credit-note correction proposal UI


norbert-kulus-blockydevs:
> ### ⚠️ The design changed — read this before starting (#3116, #3123)
>
> The mockup this issue was written against described a **per-row picker** for ambiguous invoice
> lines. That picker is **retired**. The design of record is now the **invoice-line grid** in
> `docs/plans/mockups/correction-proposal-3087.html` — a row's position *is* the line number a correction addresses, so
> the operator is never asked which line a returned item belongs to.
> `docs/specs/product-spec-oms-returns-operator-ux.md` § 5.8 carries the amendment.
>
> **This work is blocked on #3172.** `ReturnLine.resolvedOrderLineId` is hardcoded `null`
> (`returns.service.ts:444`); until it is populated OpenLinker cannot pre-fill a single quantity in
> the grid. Do not start ahead of it.
>
> **Cite the repo file, never the Artifact URL** (`docs/frontend-architecture.md` § UX Mockups). Every
> state is settable without a click: open the file with `?state=<name>`. States below are
> `body[data-state]` values.

> **Mockup states this issue implements.** **All of them.** Credit note: `credit-draft`, `credit-nothing-disposed`, `credit-no-invoice`, `credit-no-content`, `credit-lines-unauthoritative`, `credit-issued`, `credit-unmatched`, `credit-read-only`, `credit-absent-orphan`. Page: `page-loaded` (on `body[data-page-state]`), `detail-loading`, `detail-error`, `detail-unreadable`, `detail-orphan`. Frames carry `data-frame`: `flow-map`, `returns-list`, `return-detail`, `confirm-dialog`, `handoff-invoicing`.

## Problem / Context

`GET/POST /returns/:returnId/correction-proposal` (#2374/#2376) — classifying a disposed return line's credit-note correction as matched/ambiguous/no-match against the issued invoice — has no caller in `apps/web` (confirmed by #3029), even though the FE `return-proposal.schema.ts` already exists unused. There is no way for an operator to preview or record a correction proposal for review.

## Mockup

Reviewed, interactive HTML mockup — built on the real design system, all states: `docs/plans/mockups/correction-proposal-3087.html` (the Artifact URL is for live iteration only; the repo copy is the durable spec)

## Proposed Solution

[...]

Issue #3093, opened 2026-09-10
feat(returns): no-invoice/no-line-snapshot/no-disposed-lines/nothing-correctable states


norbert-kulus-blockydevs:
> ### ⚠️ The design changed — read this before starting (#3116, #3123)
>
> The mockup this issue was written against described a **per-row picker** for ambiguous invoice
> lines. That picker is **retired**. The design of record is now the **invoice-line grid** in
> `docs/plans/mockups/correction-proposal-3087.html` — a row's position *is* the line number a correction addresses, so
> the operator is never asked which line a returned item belongs to.
> `docs/specs/product-spec-oms-returns-operator-ux.md` § 5.8 carries the amendment.
>
> **This work is blocked on #3172.** `ReturnLine.resolvedOrderLineId` is hardcoded `null`
> (`returns.service.ts:444`); until it is populated OpenLinker cannot pre-fill a single quantity in
> the grid. Do not start ahead of it.
>
> **Cite the repo file, never the Artifact URL** (`docs/frontend-architecture.md` § UX Mockups). Every
> state is settable without a click: open the file with `?state=<name>`. States below are
> `body[data-state]` values.

> **Mockup states this issue implements.** `credit-no-invoice` · `credit-no-content` · `credit-lines-unauthoritative` · `credit-nothing-disposed` · `credit-absent-orphan`. **Note the renames** — the mockup's `credit-no-content` (409, lines unavailable) and `credit-lines-unauthoritative` (the pre-#1297 refusal) are what this issue's title calls `no-line-snapshot`. Also **new**: `credit-unmatched` — returned items that match no invoice line. The grid iterates the *invoice's* lines, so such an item has no row in it, and § 22 Returns forbids it vanishing silently.

## Problem / Context

4 of the 5 possible `outcome` values (`no-invoice`, `no-line-snapshot`, `no-disposed-lines`, `nothing-correctable`) have no distinct rendering.

## Proposed Solution

Add a named empty/blocked state per outcome, each with its own plain-language explanation.

## Classification

**Type**: Frontend
**Layer**: Interface (component)
**File(s)**: apps/web/src/features/returns/components/correction-proposal-panel.tsx

## Dependencies

- Part of #3087

## Assumptions

- None

## Acceptance Criteria

- [ ] Each of the 4 non-proposing outcomes renders a distinct, correctly worded state
- [ ] Tests added or updated for non-trivial logic
- [ ] No architecture boundary violations (CORE ↔ Integration)

Issue #3091, opened 2026-09-10
feat(returns): ambiguous-line picker interaction


norbert-kulus-blockydevs:
> ### ⚠️ The design changed — read this before starting (#3116, #3123)
>
> The mockup this issue was written against described a **per-row picker** for ambiguous invoice
> lines. That picker is **retired**. The design of record is now the **invoice-line grid** in
> `docs/plans/mockups/correction-proposal-3087.html` — a row's position *is* the line number a correction addresses, so
> the operator is never asked which line a returned item belongs to.
> `docs/specs/product-spec-oms-returns-operator-ux.md` § 5.8 carries the amendment.
>
> **This work is blocked on #3172.** `ReturnLine.resolvedOrderLineId` is hardcoded `null`
> (`returns.service.ts:444`); until it is populated OpenLinker cannot pre-fill a single quantity in
> the grid. Do not start ahead of it.
>
> **Cite the repo file, never the Artifact URL** (`docs/frontend-architecture.md` § UX Mockups). Every
> state is settable without a click: open the file with `?state=<name>`. States below are
> `body[data-state]` values.

> **Mockup states this issue implements.** **Superseded — this interaction no longer exists.** The picker was dropped on the finding that the question is both unanswerable (`InvoiceLine` has no id and no SKU; both candidates carry the same product name) and unnecessary (the join is available at ingestion — #3172). Nothing replaces it, because nothing is asked. **Do not implement.** Left open rather than closed so the decision is visible; close it with the epic.

## Problem / Context

An `ambiguous` line lists every invoice-line candidate and picks none — the operator must choose, and no UI exists for it.

## Proposed Solution

Render every candidate as a large, clearly labelled tappable choice (name + price), never a dense table; disable "Record for review" until every ambiguous line has a pick.

## Classification

**Type**: Frontend
**Layer**: Interface (component)
**File(s)**: apps/web/src/features/returns/components/correction-proposal-panel.tsx

## Dependencies

- Part of #3087

## Assumptions

- None

## Acceptance Criteria

- [ ] Every candidate for an ambiguous line is shown, none pre-selected
- [ ] "Record for review" is disabled while any ambiguous line is unresolved
- [ ] Tests added or updated for non-trivial logic
- [ ] No architecture boundary violations (CORE ↔ Integration)

Issue #3090, opened 2026-09-10
feat(returns): correction-proposal summary + line-status list


norbert-kulus-blockydevs:
> ### ⚠️ The design changed — read this before starting (#3116, #3123)
>
> The mockup this issue was written against described a **per-row picker** for ambiguous invoice
> lines. That picker is **retired**. The design of record is now the **invoice-line grid** in
> `docs/plans/mockups/correction-proposal-3087.html` — a row's position *is* the line number a correction addresses, so
> the operator is never asked which line a returned item belongs to.
> `docs/specs/product-spec-oms-returns-operator-ux.md` § 5.8 carries the amendment.
>
> **This work is blocked on #3172.** `ReturnLine.resolvedOrderLineId` is hardcoded `null`
> (`returns.service.ts:444`); until it is populated OpenLinker cannot pre-fill a single quantity in
> the grid. Do not start ahead of it.
>
> **Cite the repo file, never the Artifact URL** (`docs/frontend-architecture.md` § UX Mockups). Every
> state is settable without a click: open the file with `?state=<name>`. States below are
> `body[data-state]` values.

> **Mockup states this issue implements.** `credit-draft` · `credit-unmatched` · `credit-read-only` — **and the scope changed**. This is no longer "one row per return line with a status"; it is a grid over the **invoice's own lines** in document order, with a recessed read-only *as invoiced* column, an editable *after correction* column carrying quantity **and** unit price, and a computed *credit*. The reconciliation against the refund the channel already paid sits beneath it.

## Problem / Context

No page renders the proposal at all.

## Proposed Solution

Add a page/panel leading with one headline credit amount + a breakdown count (automatic / needs your pick / can't credit), then one row per line with a plain-language status (✓ credited / ? pick / ✕ can't credit), per the reviewed mockup at `docs/plans/mockups/correction-proposal-3087.html` (the Artifact URL is for live iteration only; the repo copy is the durable spec).

## Classification

**Type**: Frontend
**Layer**: Interface (page/component)
**File(s)**: apps/web/src/features/returns/components/correction-proposal-panel.tsx

## Dependencies

- Part of #3087

## Assumptions

- None

## Acceptance Criteria

- [ ] Headline amount and breakdown are visible without scrolling on desktop
- [ ] Each line shows its status in plain language, not just a raw enum value
- [ ] Tests added or updated for non-trivial logic
- [ ] No architecture boundary violations (CORE ↔ Integration)

PR #3123, opened 2026-09-10
docs(mockups): correction-proposal UI mockup

### LINKED ISSUE #3116: design(returns): correction-proposal UI mockup
## Problem / Context

Before implementation starts, Return credit-note correction proposal UI (#3087) needs a reviewed design so every sub-issue builds against

## Proposed Solution

## Classification

**Type**: Frontend
**File(s)**: `docs/plans/mockups/correction-proposal-3087.html`

## Dependencies

- None — first sub-issue of #3087, everything else in the epic reads from this mockup

## Assumptions

## Acceptance Criteria

- [ ] `docs/plans/mockups/correction-proposal-3087.html` exists and renders standalone (no external API calls)
- [ ] PR links back to the reviewed Artifact: https://claude.ai/code/artifact/704a65a0-81c6-4182-9501-80831cac92c7

norbert-kulus-blockydevs:
## Summary

Commits `docs/plans/mockups/correction-proposal-3087.html` as the durable spec for epic #3087, and
amends the product spec section that specified the design it replaces.

**The v3-vs-v9 decision, stated (review BLOCKING 2).** The file on this branch was the revision its
own author had declared wrong. It is now the **invoice-line grid**, and the earlier per-row picker is
retired. The picker asked the operator which of two identical invoice lines a returned item belonged
to; two independent reviews concluded that question is both:

- **unanswerable** — the operator holds a parcel, `InvoiceLine` carries no id and no SKU, both
  candidates carry the same product name, and the buyer never chose a line either; and
- **unnecessary** — OpenLinker builds that invoice itself from `order.items` in order
  (`order-to-issue-invoice-command.mapper.ts:113`), and the return line already arrives carrying
  `offerId`, `sku` and `unitPrice`. The join exists at ingestion and was simply never persisted.

A row's position **is** the line number a correction addresses, so nothing is asked. Both the quantity
and the unit price are editable, because `CorrectionLine` has always carried `newQuantity` **and**
`newUnitPriceGross` and the returns path used only the first — so a discount or an agreed reduction
was unexpressible. Alongside the grid sits a reconciliation against the refund the channel has already
paid; OpenLinker holds both figures and compares them nowhere today, and the first thing that
comparison catches is delivery cost, which a name-join can never match.

## ⚠️ Merge order — this holds behind #3172

`ReturnLine.resolvedOrderLineId` is hardcoded `null` (`returns.service.ts:444`) and nothing in the
tree populates it. Until **#3172** lands, OpenLinker cannot pre-fill a single quantity in this grid,
so **no frontend sub-issue of #3087 should start ahead of it**. Three further reads the sheet draws do
not exist at all — the refund figure beside the credit, per-line already-credited quantities, and the
shipping decision — and each is named in the file's own "backend work that does not exist yet" note
rather than implied to be built.

## The spec moves with it

`docs/specs/product-spec-oms-returns-operator-ux.md` § 5.8 specified the picker. It now carries a
supersession pointer at the top and an **Amendment (#3116)** at the bottom. A mockup that contradicted
a live spec would be the same *"a stale mockup is worse than none, because it actively misleads"*
failure, one file over — so the two move together or neither moves.

The amendment also records what does **not** change: a proposal is still never an automatic issue,
the irreversibility warning stays, a no-match line still never vanishes silently, and § 5.8's Hard AC
survives restated for the new shape.

## Review findings

| Finding | Status |
|---|---|
| **BLOCKING** — the irreversibility half of the § 5.8 lead is missing | **Fixed.** `RETURN_PROPOSAL_COPY.irreversible` **verbatim**, not a paraphrase, in the panel lead (where § 5.8 puts it, and where the constant's own docblock says it belongs), beside the issue action, and in the confirm dialog. Declared once in the script so the three cannot drift. The shipped wording says *"issued and sent"* rather than § 5.8's `{authority}` because `check-ui-vocabulary` (design rule P9) bans that word from operator copy — the constant's docblock records that trade. |
| **BLOCKING** — the committed file is a version its author declared wrong | **Fixed.** The grid is pushed and is the design of record; the decision and its merge order are stated above and in the file's own header. |
| **IMPORTANT** — the ambiguous-line copy is truncated, `candidatesPriceOrRateDiffer` is rendered nowhere | **Superseded, and replaced by something stronger.** There are no candidates to compare. The grid shows every line's own price *and makes it editable*, which is more evidence than a price-divergence flag. § 5.8's footer sentence is carried as `RETURN_PROPOSAL_COPY.noAutoIssue`, verbatim, in the panel's ⓘ. |
| **IMPORTANT** — `setState` never stamps the rendered surface; `[data-state="all-good"]` matches unconditionally | **Fixed, and gone one step further.** `body[data-state]` is stamped on every change, **and every state is settable without a click** — `?state=<name>` or `#state=<name>`. An E2E baseline loads the state it wants instead of synthesising clicks on the sheet's own switcher, which is how a baseline ends up testing the switcher. An unrecognised name is ignored rather than silently rendering the default, so a typo in a ticket reads as "nothing happened" and not as a wrong screenshot. `body[data-page-state]` / `body[data-credit-state]` always hold their own family's value — which is where `page-loaded` lives, since it is the *absence* of a page-level override and so is never the state in force. |
| **IMPORTANT** — the dark `--status-*` block diverges from `index.css` on every value | **Fixed at the root, and verified by diffing rather than by eye.** Both blocks are pasted verbatim: `:root` **109/109** declarations identical, `html[data-theme='dark']` **77/77** identical, and the `prefers-color-scheme` copy identical to the attribute form. Zero undeclared custom properties (111 declared, 86 used). `--overlay-scrim` comes in with the block and is what the dialog scrim uses. `--status-disabled-fg` is the real `oklch(82% 0.010 270)`. |
| Suggestion — `no-line-name` absent | **Fixed, and it was worse than a missing reason.** The grid iterates the **invoice's** lines, so a returned item matching none of them has no row at all — which § 22 Returns forbids in those words. New block, `body[data-state]="credit-unmatched"`: *"Came back, but not on this invoice"*, carrying `no-line-by-name`, `no-line-name` and `quantity-exceeds-invoiced` with the reason stated per item. `disposition-not-confirmed` keeps its grid row, because that item **is** on the invoice. |
| Suggestion — `money()` hardcodes `pl-PL` | **Fixed as far as a fixture can be.** The pin is now a named `FIXTURE_LOCALE` constant with `getBcp47Locale` (and its `'en'` default) named beside it, so an implementer reads it as a fixture choice rather than copying a literal out of a formatter. |
| Suggestion — `border-radius:999px` vs `--radius-pill` | **Faithful rather than fixed.** The remaining occurrence arrives inside the verbatim `index.css` block, which itself writes `999px` on that element. Changing it would break the byte-identity the finding above asks for. |
| Cross-cutting — epics cite only the Artifact URL | **Fixed.** #3087 and its sub-issues now cite the file path and the `data-state` values they implement. |

## Found while fixing — not in the review, and it broke the file

**`[hidden]` loses to a class selector, and the Artifact host was injecting the rule that hid it.**
`.returns-credit-note { display: flex }` and `.state-card { display: grid }` both beat the UA's
`[hidden]`. The sheet looked right in the Artifact because the host injects
`[hidden]{display:none!important}` — the repo copy, opened at `file://`, **rendered all six state
elements at once**. That is precisely the consumption mode § UX Mockups mandates for an E2E baseline
and the one this PR's own test plan describes, so the committed file was broken in the only way it is
meant to be used. Caught by driving a real browser, not by reading the source. The app hits the same
trap and answers it per class (`index.css`: `.raw-payload__body[hidden]`,
`.bulk-destbar__panel[hidden]`); a single file generalises it, because a per-class list is a list the
next state forgets to join.

## Verification

Driven in Chromium against the committed `file://` copy:

- all **14** states resolve from `?state=` and stamp the expected `body[data-state]`;
- an unknown state name is ignored rather than rendering a wrong screen;
- the unmatched block is present in `credit-unmatched` and absent in `credit-draft` (it was visible in
  both before the `[hidden]` fix);
- the irreversibility sentence is in the rendered text;
- the live grid recomputes — line 1 quantity `1 → 0` moves the total `−189,00 → −378,00 PLN`;
- light and dark both render from the real tokens, **zero page errors** in either.

`pnpm check:invariants` passes, `check-css-structure` included (the file parses with zero defects).

## Test plan

- [x] Open `docs/plans/mockups/correction-proposal-3087.html` directly in a browser — renders
      standalone, no API calls
- [x] `?state=<name>` for each of the 14 names in the sheet's own state index
- [x] Light and dark
- [ ] Merge **after** #3172

## Note on `Closes`

`Closes #3116` is deliberately **not** in this description: the base is the feature branch
`3087-correction-proposal-ui`, not `main`, so GitHub would not fire it — and `CLAUDE.md` forbids
closing by hand. #3116 closes when the epic branch merges.

https://claude.ai/code/session_01PAMTZ8y5LZZzxrFJhDSiAv

### DIFF
--- a/docs/plans/mockups/correction-proposal-3087.html
+++ b/docs/plans/mockups/correction-proposal-3087.html
@@ -1,0 +1,285 @@
+<title>Correction Proposal</title>
+<link rel="stylesheet" href="https://fonts.googleapis.com/css2?family=IBM+Plex+Sans:wght@400;500;600;700&family=IBM+Plex+Mono:wght@400;500;600&display=swap">
+<style>
+:root, [data-theme='light'] {
+  --bg-canvas: oklch(99% 0.003 80); --bg-shell: oklch(97.5% 0.004 80);
+  --bg-surface: #ffffff; --bg-surface-muted: oklch(96% 0.005 80); --bg-surface-hover: oklch(93% 0.006 80);
+  --border-subtle: oklch(93.5% 0.006 80); --border-default: oklch(88% 0.008 80); --border-strong: oklch(78% 0.010 80);
+  --text-primary: oklch(20% 0.012 80); --text-secondary: oklch(38% 0.010 80);
+  --text-muted: oklch(52% 0.008 80); --text-disabled: oklch(70% 0.005 80); --text-on-primary: oklch(18% 0.012 50);
+  --accent-primary: oklch(68% 0.18 50); --accent-primary-hover: oklch(62% 0.19 50);
+  --accent-primary-soft: oklch(96% 0.04 60); --accent-primary-border: oklch(85% 0.10 55); --accent-ring: oklch(68% 0.18 50 / 0.30);
+  --status-success: oklch(54% 0.14 150); --status-success-soft: oklch(96% 0.04 150); --status-success-border: oklch(85% 0.08 150); --status-success-fg: oklch(36% 0.12 150);
+  --status-warning: oklch(72% 0.16 85); --status-warning-soft: oklch(96% 0.05 85); --status-warning-border: oklch(85% 0.10 85); --status-warning-fg: oklch(42% 0.12 80);
+  --status-error: oklch(58% 0.20 25); --status-error-soft: oklch(96% 0.04 25); --status-error-border: oklch(85% 0.10 25); --status-error-fg: oklch(42% 0.16 25);
+  --status-info: oklch(56% 0.14 245); --status-info-soft: oklch(96% 0.03 245); --status-info-border: oklch(85% 0.08 245); --status-info-fg: oklch(40% 0.12 245);
+  --status-disabled-soft: oklch(95% 0.005 80); --status-disabled-border: oklch(85% 0.008 80); --status-disabled-fg: oklch(38% 0.010 80);
+  --space-2:.5rem; --space-3:.75rem; --space-4:1rem; --space-5:1.5rem;
+  --radius-md:8px; --radius-lg:10px; --radius-xl:14px; --radius-pill:9999px;
+  --shadow-xs: 0 1px 2px rgb(15 18 26 / 0.04);
+  --shadow-soft: 0 1px 2px rgb(15 18 26 / .04), 0 6px 16px rgb(15 18 26 / .06);
+  --shadow-focus: 0 0 0 3px var(--accent-ring);
+  --font-sans: 'IBM Plex Sans', ui-sans-serif, system-ui, sans-serif;
+  --font-mono: 'IBM Plex Mono', ui-monospace, SFMono-Regular, Menlo, monospace;
+}
+@media (prefers-color-scheme: dark) {
+  :root:not([data-theme='light']) {
+    --bg-canvas: oklch(14% 0.005 270); --bg-shell: oklch(16% 0.006 270);
+    --bg-surface: oklch(19% 0.007 270); --bg-surface-muted: oklch(24% 0.009 270); --bg-surface-hover: oklch(28% 0.010 270);
+    --border-subtle: oklch(24% 0.010 270); --border-default: oklch(30% 0.012 270); --border-strong: oklch(42% 0.014 270);
+    --text-primary: oklch(96% 0.006 270); --text-secondary: oklch(78% 0.010 270);
+    --text-muted: oklch(60% 0.012 270); --text-disabled: oklch(42% 0.012 270); --text-on-primary: oklch(16% 0.012 50);
+    --accent-primary: oklch(72% 0.18 50); --accent-primary-hover: oklch(78% 0.18 50);
+    --accent-primary-soft: oklch(28% 0.08 50); --accent-primary-border: oklch(40% 0.14 55); --accent-ring: oklch(72% 0.18 50 / 0.40);
+    --status-success-soft: oklch(28% 0.07 150); --status-success-border: oklch(40% 0.10 150); --status-success-fg: oklch(80% 0.12 150);
+    --status-warning-soft: oklch(30% 0.08 85); --status-warning-border: oklch(45% 0.11 85); --status-warning-fg: oklch(85% 0.12 85);
+    --status-error-soft: oklch(28% 0.09 25); --status-error-border: oklch(42% 0.13 25); --status-error-fg: oklch(84% 0.14 25);
+    --status-info-soft: oklch(28% 0.06 245); --status-info-border: oklch(42% 0.10 245); --status-info-fg: oklch(82% 0.10 245);
+    --status-disabled-soft: oklch(26% 0.006 270); --status-disabled-border: oklch(36% 0.010 270); --status-disabled-fg: oklch(70% 0.010 270);
+    --shadow-soft: 0 1px 2px rgb(0 0 0 / .4), 0 8px 24px rgb(0 0 0 / .4);
+    --shadow-xs: 0 1px 2px rgb(0 0 0 / .4);
+  }
+}
+:root[data-theme='dark'] {
+  --bg-canvas: oklch(14% 0.005 270); --bg-shell: oklch(16% 0.006 270);
+  --bg-surface: oklch(19% 0.007 270); --bg-surface-muted: oklch(24% 0.009 270); --bg-surface-hover: oklch(28% 0.010 270);
+  --border-subtle: oklch(24% 0.010 270); --border-default: oklch(30% 0.012 270); --border-strong: oklch(42% 0.014 270);
+  --text-primary: oklch(96% 0.006 270); --text-secondary: oklch(78% 0.010 270);
+  --text-muted: oklch(60% 0.012 270); --text-disabled: oklch(42% 0.012 270); --text-on-primary: oklch(16% 0.012 50);
+  --accent-primary: oklch(72% 0.18 50); --accent-primary-hover: oklch(78% 0.18 50);
+  --accent-primary-soft: oklch(28% 0.08 50); --accent-primary-border: oklch(40% 0.14 55); --accent-ring: oklch(72% 0.18 50 / 0.40);
+  --status-success-soft: oklch(28% 0.07 150); --status-success-border: oklch(40% 0.10 150); --status-success-fg: oklch(80% 0.12 150);
+  --status-warning-soft: oklch(30% 0.08 85); --status-warning-border: oklch(45% 0.11 85); --status-warning-fg: oklch(85% 0.12 85);
+  --status-error-soft: oklch(28% 0.09 25); --status-error-border: oklch(42% 0.13 25); --status-error-fg: oklch(84% 0.14 25);
+  --status-info-soft: oklch(28% 0.06 245); --status-info-border: oklch(42% 0.10 245); --status-info-fg: oklch(82% 0.10 245);
+  --status-disabled-soft: oklch(26% 0.006 270); --status-disabled-border: oklch(36% 0.010 270); --status-disabled-fg: oklch(70% 0.010 270);
+  --shadow-soft: 0 1px 2px rgb(0 0 0 / .4), 0 8px 24px rgb(0 0 0 / .4);
+  --shadow-xs: 0 1px 2px rgb(0 0 0 / .4);
+}
+* { box-sizing: border-box; }
+body { margin:0; min-width:320px; background:var(--bg-canvas); color:var(--text-primary); font-family:var(--font-sans); font-size:.875rem; line-height:1.55; }
+h1,h2,h3,p { margin:0; }
+a{ color:var(--accent-primary-hover); text-decoration:underline; text-underline-offset:.16em; }
+button:focus-visible, a:focus-visible{ outline:none; box-shadow:var(--shadow-focus); border-radius:var(--radius-md); }
+
+.page{ max-width:700px; margin:0 auto; padding:28px 24px 60px; }
+.back-link{ display:inline-flex; align-items:center; gap:.375rem; font-size:.8125rem; color:var(--text-muted); text-decoration:none; margin-bottom:10px; }
+.eyebrow{ margin:0; font-size:.75rem; font-weight:700; letter-spacing:.08em; text-transform:uppercase; color:var(--accent-primary-hover); }
+.page-header{ display:grid; gap:.5rem; margin-bottom:22px; }
+.page-title{ font-size:1.75rem; font-weight:600; letter-spacing:-.02em; margin-top:6px; }
+.page-description{ color:var(--text-secondary); font-size:.9375rem; max-width:60ch; }
+
+.button{ display:inline-flex; align-items:center; justify-content:center; gap:8px; border:1px solid var(--accent-primary-border); border-radius:var(--radius-md); padding:0 var(--space-4); height:2.25rem; background:var(--accent-primary); color:var(--text-on-primary); font-family:inherit; font-size:.875rem; font-weight:600; cursor:pointer; box-shadow:var(--shadow-xs); }
+.button:hover{ background:var(--accent-primary-hover); }
+.button--secondary{ border-color:var(--border-default); background:var(--bg-surface); color:var(--text-primary); }
+.button--secondary:hover{ background:var(--bg-surface-muted); border-color:var(--border-strong); }
+.button:disabled{ cursor:not-allowed; opacity:.5; box-shadow:none; }
+.button--sm{ height:1.875rem; padding:0 var(--space-3); font-size:.8125rem; }
+
+/* headline summary — the whole point, at a glance */
+.summary{ display:flex; align-items:center; gap:20px; padding:26px 26px; border-radius:var(--radius-xl); background:var(--bg-surface); border:1px solid var(--border-default); box-shadow:var(--shadow-soft); margin-bottom:26px; flex-wrap:wrap; border-left:5px solid var(--status-success); }
+.summary--empty{ border-left-color:var(--border-strong); }
+.summary__num{ font-size:2.75rem; font-weight:700; letter-spacing:-.03em; line-height:1; font-variant-numeric:tabular-nums; }
+.summary__num--credit{ color:var(--status-success-fg); }
+.summary__body{ flex:1; min-width:200px; }
+.summary__title{ font-size:1.125rem; font-weight:700; }
+.summary__desc{ font-size:.8125rem; color:var(--text-muted); margin-top:3px; }
+.summary__breakdown{ display:flex; gap:16px; font-size:.8125rem; color:var(--text-muted); margin-top:10px; flex-wrap:wrap; }
+.summary__breakdown b{ color:var(--text-primary); font-weight:700; }
+
+.line{ display:flex; align-items:center; gap:14px; padding:15px 18px; border-radius:var(--radius-lg); border:1px solid var(--border-subtle); background:var(--bg-surface); margin-bottom:10px; }
+.line--pick{ align-items:flex-start; border-color:var(--status-warning-border); background:var(--status-warning-soft); }
+.line--no{ opacity:.7; }
+.line__icon{ width:34px; height:34px; border-radius:999px; display:flex; align-items:center; justify-content:center; font-size:1rem; flex-shrink:0; }
+.line__icon--yes{ background:var(--status-success-soft); color:var(--status-success-fg); }
+.line__icon--pick{ background:var(--status-warning-border); color:var(--status-warning-fg); }
+.line__icon--no{ background:var(--status-disabled-soft); color:var(--text-disabled); }
+.line__body{ flex:1; min-width:0; }
+.line__name{ font-weight:600; font-size:.9375rem; }
+.line__sub{ font-size:.8125rem; color:var(--text-muted); margin-top:2px; }
+.line__amount{ font-family:var(--font-mono); font-size:.9375rem; font-weight:700; flex-shrink:0; }
+.line__amount--yes{ color:var(--status-success-fg); }
+.line__amount--no{ color:var(--text-disabled); font-family:var(--font-sans); font-weight:500; font-size:.8125rem; }
+
+/* ambiguous inline choice — two big obvious tap targets, no table */
+.choice-row{ display:flex; gap:10px; margin-top:12px; }
+.choice-btn{ flex:1; text-align:left; border:2px solid var(--border-default); border-radius:var(--radius-md); padding:12px 14px; background:var(--bg-surface); cursor:pointer; }
+.choice-btn:hover{ border-color:var(--border-strong); }
+.choice-btn--picked{ border-color:var(--status-success); background:var(--status-success-soft); }
+.choice-btn__label{ font-weight:700; font-size:.875rem; }
+.choice-btn__price{ font-family:var(--font-mono); font-size:.8125rem; color:var(--text-muted); margin-top:3px; }
+.choice-btn--picked .choice-btn__price{ color:var(--status-success-fg); }
+
+.section-title{ font-size:.75rem; font-weight:700; text-transform:uppercase; letter-spacing:.05em; color:var(--text-muted); margin:22px 0 8px; }
+
+.sticky-footer{ position:sticky; bottom:0; background:var(--bg-surface); border-top:1px solid var(--border-subtle); padding:16px 0 0; margin-top:20px; display:flex; justify-content:space-between; align-items:center; gap:12px; flex-wrap:wrap; }
+.sticky-footer__note{ font-size:.75rem; color:var(--text-muted); max-width:34ch; }
+
+.callout{ display:flex; gap:10px; padding:12px 14px; border-radius:var(--radius-md); border:1px solid; margin-bottom:16px; }
+.callout--warning{ background:var(--status-warning-soft); border-color:var(--status-warning-border); }
+.callout--info{ background:var(--status-info-soft); border-color:var(--status-info-border); }
+.callout__text{ font-size:.8125rem; color:var(--text-secondary); }
+.callout__text b{ color:var(--text-primary); }
+
+.demo-toolbar{ display:flex; align-items:center; gap:8px; margin-bottom:18px; padding:8px 12px; background:var(--bg-surface-muted); border-radius:var(--radius-md); flex-wrap:wrap; }
+.demo-toolbar__label{ font-family:var(--font-mono); font-size:.625rem; text-transform:uppercase; letter-spacing:.08em; color:var(--text-muted); margin-right:2px; }
+.chip{ display:inline-flex; align-items:center; gap:.25rem; height:1.375rem; padding:0 .55rem; border:1px solid var(--border-subtle); border-radius:var(--radius-pill); background:var(--bg-surface); color:var(--text-secondary); font-size:.6875rem; font-weight:500; cursor:pointer; }
+.chip.chip--active{ background:var(--accent-primary-soft); border-color:var(--accent-primary-border); color:var(--accent-primary-hover); font-weight:600; }
+.state-card{ border:1px solid var(--border-default); border-radius:var(--radius-lg); background:var(--bg-surface); box-shadow:var(--shadow-soft); padding:32px 28px; text-align:left; }
+.state-card__title{ font-size:1.0625rem; font-weight:600; margin-top:4px; }
+.state-card__message{ color:var(--text-secondary); font-size:.875rem; margin-top:6px; max-width:56ch; }
+</style>
+
+<div class="page">
+  <a class="back-link" href="#">&larr; Back to return AL-RET-88213</a>
+  <header class="page-header">
+    <p class="eyebrow">Credit note</p>
+    <h1 class="page-title">Crediting this return on the invoice</h1>
+    <p class="page-description">A preview only — nothing is sent to the tax office until you confirm.</p>
+  </header>
+
+  <div id="content"></div>
+
+  <div class="demo-toolbar" role="group" aria-label="Mockup — not part of the product">
+    <span class="demo-toolbar__label">Mockup</span>
+    <button class="chip chip--active" data-state="mixed" onclick="setState('mixed')">Mixed result</button>
+    <button class="chip" data-state="all-good" onclick="setState('all-good')">All creditable</button>
+    <button class="chip" data-state="none" onclick="setState('none')">Nothing creditable</button>
+    <button class="chip" data-state="no-invoice" onclick="setState('no-invoice')">No invoice yet</button>
+  </div>
+</div>
+
+<script>
+function money(n, cur){ return n.toLocaleString('pl-PL', {minimumFractionDigits:2, maximumFractionDigits:2}) + ' ' + cur; }
+
+const DATA = {
+  mixed: {
+    currency:'PLN', invoice:'FV/2026/09/0142',
+    lines:[
+      { id:'l1', name:'Running shoes — size 42', qty:1, status:'yes', amount:249.00 },
+      { id:'l2', name:'Wireless earbuds', qty:1, status:'pick',
+        options:[{ label:'Invoice line #3', price:189.00 }, { label:'Invoice line #5', price:179.00 }] },
+      { id:'l3', name:'Phone case — clear', qty:1, status:'no', reason:"We haven't confirmed the stock came back yet." },
+      { id:'l4', name:'Screen protector 2-pack', qty:2, status:'no', reason:"Nothing on the invoice matches this item's name." },
+    ],
+  },
+  'all-good': {
+    currency:'PLN', invoice:'FV/2026/09/0201',
+    lines:[
+      { id:'l1', name:'Running shoes — size 42', qty:1, status:'yes', amount:249.00 },
+      { id:'l2', name:'Water bottle', qty:2, status:'yes', amount:78.00 },
+    ],
+  },
+  none: {
+    currency:'PLN', invoice:'FV/2026/09/0055',
+    lines:[
+      { id:'l1', name:'Phone case — clear', qty:1, status:'no', reason:"We haven't confirmed the stock came back yet." },
+      { id:'l2', name:'Screen protector 2-pack', qty:2, status:'no', reason:"That quantity is more than the invoice ever billed for this line." },
+    ],
+  },
+};
+
+let currentState = 'mixed';
+let choices = {};
+let recorded = false;
+
+function creditable(l) { return l.status === 'yes' || (l.status === 'pick' && choices[l.id] !== undefined); }
+function amountFor(l) {
+  if (l.status === 'yes') return l.amount;
+  if (l.status === 'pick' && choices[l.id] !== undefined) return l.options[choices[l.id]].price;
+  return 0;
+}
+
+function renderLine(l) {
+  if (l.status === 'yes') {
+    return `<div class="line">
+      <div class="line__icon line__icon--yes">✓</div>
+      <div class="line__body"><div class="line__name">${l.name}</div><div class="line__sub">Credited automatically</div></div>
+      <div class="line__amount line__amount--yes">${money(l.amount, DATA[currentState].currency)}</div>
+    </div>`;
+  }
+  if (l.status === 'pick') {
+    const picked = choices[l.id];
+    return `<div class="line line--pick">
+      <div class="line__icon line__icon--pick">?</div>
+      <div class="line__body">
+        <div class="line__name">${l.name}</div>
+        <div class="line__sub">${picked !== undefined ? '✓ You picked which one — ready to credit' : 'This item is on the invoice twice. Which one was it?'}</div>
+        <div class="choice-row">
+          ${l.options.map((o, i) => `<button class="choice-btn ${picked===i?'choice-btn--picked':''}" onclick="pickOption('${l.id}', ${i})">
+            <div class="choice-btn__label">${picked===i?'✓ ':''}${o.label}</div>
+            <div class="choice-btn__price">${money(o.price, DATA[currentState].currency)}</div>
+          </button>`).join('')}
+        </div>
+      </div>
+    </div>`;
+  }
+  return `<div class="line line--no">
+    <div class="line__icon line__icon--no">✕</div>
+    <div class="line__body"><div class="line__name">${l.name}</div><div class="line__sub">${l.reason}</div></div>
+    <div class="line__amount line__amount--no">Can't credit</div>
+  </div>`;
+}
+
+function render() {
+  const content = document.getElementById('content');
+
+  if (currentState === 'no-invoice') {
+    content.innerHTML = `<div class="state-card">
+      <p class="eyebrow">No invoice yet</p>
+      <h3 class="state-card__title">This order hasn't been invoiced</h3>
+      <p class="state-card__message">There's nothing to credit until an invoice exists for this order.</p>
+    </div>`;
+    return;
+  }
+
+  const d = DATA[currentState];
+  const total = d.lines.reduce((s, l) => s + amountFor(l), 0);
+  const creditableCount = d.lines.filter(creditable).length;
+  const pickCount = d.lines.filter(l => l.status === 'pick' && choices[l.id] === undefined).length;
+
+  const summaryHtml = creditableCount === 0
+    ? `<div class="summary summary--empty">
+        <div class="summary__body">
+          <div class="summary__title">Nothing here can be credited yet</div>
+          <div class="summary__desc">See why below — usually it's waiting on a stock confirmation or a name that doesn't match.</div>
+        </div>
+      </div>`
+    : `<div class="summary">
+        <div class="summary__num summary__num--credit">${money(total, d.currency)}</div>
+        <div class="summary__body">
+          <div class="summary__title">Ready to credit on invoice ${d.invoice}</div>
+          <div class="summary__breakdown">
+            <span><b>${d.lines.filter(l=>l.status==='yes').length}</b> automatic</span>
+            ${pickCount > 0 ? `<span><b>${pickCount}</b> need your pick</span>` : ''}
+            ${d.lines.filter(l=>l.status==='no').length > 0 ? `<span><b>${d.lines.filter(l=>l.status==='no').length}</b> can't credit</span>` : ''}
+          </div>
+        </div>
+      </div>`;
+
+  content.innerHTML = `
+    ${summaryHtml}
+    ${d.lines.map(renderLine).join('')}
+    <div class="sticky-footer">
+      <span class="sticky-footer__note">${recorded ? 'Recorded — an operator confirms and issues it from the invoice.' : (pickCount > 0 ? "Pick an invoice line above for every \"?\" item first." : "This only records a proposal for review. It doesn't send anything yet.")}</span>
+      ${recorded
+        ? `<button class="button button--secondary" disabled>✓ Recorded for review</button>`
+        : `<button class="button" ${(creditableCount === 0 || pickCount > 0) ? 'disabled' : ''} onclick="recordProposal()">Record for review</button>`}
+    </div>
+  `;
+}
+
+function pickOption(lineId, idx) {
+  choices[lineId] = idx;
+  render();
+}
+function recordProposal() { recorded = true; render(); }
+
+function setState(name) {
+  currentState = name; choices = {}; recorded = false;
+  document.querySelectorAll('.demo-toolbar .chip[data-state]').forEach(c => c.classList.toggle('chip--active', c.dataset.state === name));
+  render();
+}
+
+setState('mixed');
+</script>


norbert-kulus-blockydevs:
## PR Review — #3123 (correction-proposal mockup)

**Note: this is an expanded review.** The original senior-frontend pass below has been reviewed again through three additional lenses — OpenLinker domain specialist, e-commerce operations specialist, and senior backend/system architect — and the findings from all four passes are consolidated here into one review.

Read the mockup source directly (`docs/plans/mockups/correction-proposal-3087.html`, 285 lines, self-contained apart from the Google Fonts stylesheet).

### Propose vs. Issue separation ✅ — the highest-risk thing to get right here
Per `docs/architecture-overview.md` § 22 Returns, the correction-proposal service is documented as strictly read/propose-only — *"holding a proposal confers no authority to issue one... Issuing stays the operator's confirmed act through the existing correction flow."* The mockup honors this precisely: the header states *"A preview only — nothing is sent to the tax office until you confirm,"* the only CTA is **"Record for review"** (never "Issue"/"Send"), and the post-record state redirects to *"an operator confirms and issues it from the invoice."* No blurring of the two steps. This also matches the backend's GET/POST split correctly: `previewProposal` recomputes fresh on every read (nothing persisted), and the mockup has no "regenerate" affordance because none is needed — the displayed numbers are implicitly always current, exactly like the real `compute`-backed GET. That's the right call, not an omission.

### Ambiguous outcome ✅ (mostly), with a real completeness gap
The `pick` line type is a proper multi-candidate disambiguation UI — two tappable candidate buttons, no silent auto-select, and "Record for review" stays disabled while any pick is outstanding (`pickCount > 0`), matching the documented rule that `ambiguous` "lists EVERY candidate and selects none." One completeness gap, non-blocking: the demo's ambiguous example uses two *different* prices (189.00 vs 179.00 PLN, `l2`'s options). The doc specifically calls out that ambiguity must hold even when every candidate carries the **same** price and rate — that's the harder, more important case the rule exists for, since nothing in the current UI would look any different for it. Related: the copy at that point (*"This item is on the invoice twice. Which one was it?"*) never explains *why* the system can't just pick the obvious-looking one — for an operator hitting the same-price case in production, that silence will read as unexplained friction rather than a deliberate safeguard (positional ambiguity: `originalLineNumber` is a 1-based array position and `InvoiceLine` carries no identifier/sku, only names). A short inline explainer ("Both entries match this item's name — pick which return applies to which") would turn a confusing wall into a documented constraint.

### Named non-proposing outcomes — real gap
The backend documents four named exit reasons: `no-invoice`, `no-line-snapshot`, `no-disposed-lines`, `nothing-correctable`. The mockup's demo toolbar enumerates `mixed` / `all-good` / `none` / `no-invoice` — **`no-line-snapshot` is entirely absent**, not even as a per-line reason string. Per the docs, a pre-#1297 document without the issued-line snapshot is REFUSED outright ("never diffed against the order's current state"), a distinct and important case from "no invoice at all" — worth a fifth demo state, given `no-invoice` already gets a dedicated full-page treatment. `nothing-correctable` and `no-disposed-lines` are reasonably represented via the `none` state's per-line reasons.

### Color usage — a real, if smaller, instance of the #2507 pattern
`docs/frontend-ui-style-guide.md` (#2507): *"a finished or expected state is plain ink; only a state someone has to act on is tinted."* The `pick` (needs-action) line correctly gets full warning tint. But the `yes` (auto-creditable, "Credited automatically") line gets a full green icon badge **and** a colored monospace amount — for a state the doc explicitly frames as *finished*, not actionable. In the `all-good` demo state, every row is `yes`, so the whole list renders green top to bottom — the identical shape the #2507 rule exists to prevent, just at line-item grain instead of a table column. The headline summary card's left-border accent is fine (a border isn't a fill, and this is the alert-adjacent headline number the style guide's large-panel exception covers) — only the per-line `yes` treatment needs desaturating toward plain ink with a small tick.

### E-commerce / fiscal-document lens (new)
This screen feeds a credit note that, once issued, reaches a tax authority and can't be withdrawn — reviewed accordingly, not just as a warehouse UI:
- **Financial delta prominence is good** — the credit total is the largest, most prominent element on the page (2.75rem, green, top position), and correctly shows only a quantity-derived gross figure with no fabricated net/rate breakdown, matching ADR-063's "corrections are quantity-only deltas, no net/rate recomputation" rule. This is the right level of restraint — it doesn't imply a tax computation the backend never performs.
- **Real gap: no correction history.** Nothing on the page indicates whether this invoice already carries an earlier correction (a repeat-return scenario is routine — a buyer sends back part of an order today, more next week). A finance-adjacent operator reviewing a credit-note proposal will reasonably want to see "1 correction already issued on this invoice" before confirming a second one, both to avoid double-crediting the same line and to sanity-check the running total. Worth a small "prior corrections" strip or link, even collapsed by default.

### Backend/system-architecture lens (new)
- The proposal-reuse mechanics (an identical open proposal is REUSED; a diverged one is abandoned and replaced via a canonical, key-sorted comparison) have no visible trace in the mockup, and that's largely fine — since every read recomputes fresh, "reuse vs. replace" is invisible at the read layer by design and only matters at record/persist time, which the mockup doesn't need to model. One small, non-blocking gap: after "Record for review," there's no way to tell whether recording again later (say, because the return's disposition changed) reuses the same underlying proposal row or creates a fresh review item — a one-line note distinguishing "still your original proposal" from "this replaces your earlier one" would remove an ambiguity a real operator will eventually ask about.
- No case in the mockup implies two proposals could be simultaneously "in review" for one invoice, which is correct given the `order_changes` open-target uniqueness constraint (`targetRef = correction:{returnId}:{invoiceRecordId}`) — nothing to fix here, just confirming the UI doesn't accidentally promise a shape the backend forbids.

Strong grounding in the actual backend contract across every lens — the propose/issue split, the ambiguous-outcome mechanics, and the deliberate absence of net/rate computation are all exactly right, and those are the highest-risk things to get wrong on a screen that ends at a tax authority. Before this becomes the basis for real component work: (1) desaturate the `yes`/auto-credited line styling per #2507's "finished state = plain ink" rule, (2) add a `no-line-snapshot` demo state, (3) add a same-price ambiguous-candidates variant plus a short explainer for why the pick is required, (4) consider surfacing prior-correction history for repeat-return cases.

piotrswierzy (commented):

**Substantively the most accurate of its batch, and the only one that satisfies the `data-state` rule** — it is the shape the others should copy.

Three hard things are right. **Ambiguity is genuinely shown, not resolved**: the `pick` row lists every candidate, selects none (`selectedOriginalLineNumber: null` faithfully modelled), and submit is **disabled** until the operator picks — so OL never guesses. **§ 5.8's Hard AC is met**: `all-good` and `mixed` are unmistakably different *before* the confirm button (warning-tinted row, `?` glyph, "1 need your pick", disabled submit). And **`nothing-correctable` correctly returns the full body** — the `none` state still lists every line with its exclusion reason under "Nothing here can be credited yet", which is the documented behaviour and the easiest thing to get wrong. The proposal-is-data boundary is stated twice.

### BLOCKING — the irreversibility half of § 5.8's lead is deleted

Line ~136. The spec specifies the lead verbatim:

> **Credit note proposal.** OpenLinker has matched these returned lines to lines on invoice **{invoiceNumber}**. Check the match before issuing — **a correction to a document already sent to {authority} cannot be withdrawn.**

The mockup keeps the reassuring half (*"A preview only — nothing is sent to the tax office until you confirm"*) and drops the warning. Since a mockup is the durable spec, this is copy that **actively reassures where the spec demands a warning** about an act that cannot be undone — the § UX Mockups "stale mockup is worse than none" failure mode, on the highest-stakes sentence on the page.

### IMPORTANT — `candidatesPriceOrRateDiffer` is never rendered, and the ambiguous sentence is truncated

`renderLine()`, the `pick` branch (~215). The spec's sentence is *"Invoice {n} has {k} identical lines for this product. OpenLinker cannot tell which one this return refers to; the correction amount is the same either way **unless these lines were priced differently**."* The mockup says only *"This item is on the invoice twice. Which one was it?"* — and its two candidates are priced **differently** (189.00 / 179.00), i.e. the fixture is exactly the case where the missing clause matters.

`return-correction-proposal.types.ts` is explicit that the field exists to be shown: *"`false` on an ambiguous line is evidence the operator may act on; it is never grounds for OL to pick."* Rendering candidates without the price-divergence evidence removes the only thing that lets an operator decide.

### IMPORTANT — the auto-issue footer explanation is replaced with a weaker one

§ 5.8 requires the absence of auto-issue be *"explained once in the panel footer: 'OpenLinker will not issue corrections automatically until invoice lines carry a stable reference.'"* The mockup says *"This only records a proposal for review. It doesn't send anything yet."* — true, but it answers a different question. The missing sentence is what tells an operator this is a permanent design property (positional `originalLineNumber`), not a temporary limitation.

### IMPORTANT — two named outcomes have no state

Four of five `ReturnCorrectionProposalOutcomeValues` are covered. Missing:
- **`no-line-snapshot`** — a pre-#1297 document, **REFUSED** rather than diffed against the order's current state. A distinct operator message, and the one exit with a *structural* remedy.
- **`no-disposed-lines`** — nothing disposed yet.

`architecture-overview.md` § 22 names both as first-class precisely so a silent decline is impossible. A mockup that omits them produces a UI falling through to a generic empty state for two conditions the API distinguishes.

Also, three distinct UI states *within* `mixed` carry no attribute or button: **ambiguous-unpicked** (submit disabled, footer "Pick an invoice line above…"), **ambiguous-picked**, and **recorded** (`✓ Recorded for review`). § 5.8's Hard AC is about the first being visibly different before the confirm button, so it is the state an E2E spec will most want to name.

### SUGGESTIONS

- Three of four `ReturnCorrectionNoMatchReasonValues` are represented with correct operator-language copy; `no-line-name` is absent.
- `border-radius:999px` at line 91 where `--radius-pill` is declared eight lines above.
- `money()` hardcodes `'pl-PL'`. The app resolves locale through `getBcp47Locale` / `useNumberFormat` and defaults to `en` — a mockup that pins a locale invites an implementation that does too.

---

- **`data-state` is on the switcher BUTTON only, never on the rendered state** — `setState()` toggles a class and replaces `innerHTML`; nothing stamps `<body>` or the container. The exemplar § UX Mockups names drives every state from `body[data-state="…"]`, which is why an E2E task can *set* a state and assert on it. Fix: `document.body.dataset.state = name`, plus a value + button for every dialog and gallery card.
- **The epics cite only the Artifact URL** — § UX Mockups requires implementation tickets to cite the **file path** and `data-state` values, and the E2E baseline to come from the repo copy *"never from the Artifact URL."*
- **Dark palette is invented while the light palette is faithful.** ~15 `--status-*` dark values diverge from `index.css:405–504`; `--status-disabled-fg` at `oklch(70%)` vs the real `oklch(82%)` is a genuine contrast regression. Same block appears in all four — fix once and propagate.
- **`--overlay-scrim` is never declared**, so hardcoded `rgba(22,32,43,.45)` scrims do not respond to the dark theme (`index.css:319/504` has the real token).

norbert-kulus-blockydevs:
## Mockup rebuilt — v9, and it is a different screen than the one reviewed

Artifact (v9): https://claude.ai/code/artifact/704a65a0-81c6-4182-9501-80831cac92c7

@piotrswierzy reviewed **v3**. Since then the mockup went through four rounds driven by the product owner not being able to read it, and two independent reviews — an OMS domain review and a competitor review — concluded the screen's central interaction was wrong. So this is not a fix-list pass against the review; the surface changed underneath it. Statuses below are honest, including one regression I introduced.

### The central change: the picker is gone

v3 asked the operator to choose which of two identical invoice lines a returned item belonged to. Both reviews concluded independently that the question is **unanswerable and manufactured**:

- **Unanswerable** — the operator holds a parcel; `InvoiceLine` is `{name, quantity, unitPriceGross, taxRate, unit?}` with no id and no SKU, both candidates carry the same product name, and the buyer never chose a line either.
- **Manufactured** — OpenLinker builds that invoice itself from the order's own items, in order (`order-to-issue-invoice-command.mapper.ts:113`), so invoice line *N* is `order.items[N-1]`. The return line already arrives carrying `offerId`, `sku` and `unitPrice`, and `unitPrice` is even written to `rawPayload` (`returns.service.ts:538`) and never read. `ReturnLine.resolvedOrderLineId` exists for the join and is hardcoded `null` (`returns.service.ts:444`).

v9 renders instead the shape five Polish invoicing tools share — a grid over the **invoice's own lines**, in document order: a recessed read-only *as invoiced* column, editable *after correction* (quantity **and** unit price), and a computed *credit*. Row position **is** the corrected line number, so nothing is asked. BaseLinker markets the absence of line-picking as a feature (*"bez jakiejkolwiek ręcznej pracy przy wybieraniu pozycji"*).

Also new, replacing the picker as the screen's focus: a **reconciliation against the refund the channel already paid**. OpenLinker holds both figures and compares them nowhere today. It is what catches the **delivery cost**, which the name-join can never match — so today delivery is silently never credited despite the 14-day withdrawal-right obligation.

### Your findings, one by one

| Finding | Status in v9 |
|---|---|
| **BLOCKING** — irreversibility half of the lead deleted | **Resolved**, differently placed. The warning now sits beside the irreversible action (`Issue credit note` → *"Once it has been issued and sent, it cannot be withdrawn."*) and again in the dialog, rather than at the top of a panel where nothing irreversible happens. Note the spec line you quoted contains `{authority}` — `check-ui-vocabulary` bans that word from operator copy, and the shipped workaround is exactly *"issued and sent"* (`return-proposal.copy.ts:11-20`). |
| **IMPORTANT** — `candidatesPriceOrRateDiffer` never rendered | **Superseded.** There are no candidates to compare; the grid shows every line's own price and makes it editable, which is strictly more evidence. If a genuinely unresolvable line survives #3171, that path needs a renderer again — flagged below. |
| **IMPORTANT** — auto-issue footer explanation weaker | **Partly.** `RETURN_PROPOSAL_COPY.noAutoIssue` is now carried verbatim in the panel's ⓘ. Your sentence's premise is being removed rather than restated: invoice lines still carry no stable reference, but after #3171 OpenLinker no longer needs one, because it resolves the join upstream. |
| **IMPORTANT** — `no-line-snapshot` and `no-disposed-lines` have no state | **Resolved**, and renamed. The switcher now carries `credit-nothing-disposed`, `credit-no-invoice`, `credit-no-content` (409, content unavailable), `credit-lines-unauthoritative` (the pre-#1297 refusal), `credit-issued`, `credit-read-only`, plus `credit-absent-orphan`. **Every name changed** (`proposal-*` → `credit-*`) because the panel is no longer a proposal — tickets citing the old names need updating. |
| **IMPORTANT** — ambiguous-unpicked / picked / recorded unnamed | **Superseded** for the first two; `credit-issued` and `credit-read-only` exist. |
| **Cross-cutting** — `data-state` on the switcher button only | **Partly.** The state is now stamped on the rendered element (`#correction`), and frames carry `data-frame` so frame identity is never confused with UI state. Your stronger point stands: it is still *set* by clicking, not settable directly, because this file holds five frames and a single `body[data-state]` cannot describe them. |
| **Cross-cutting** — dark palette invented, ~15 divergent values, `--status-disabled-fg` contrast regression | **Resolved at the root.** Both token blocks are now pasted **verbatim** from `index.css` (`:root` 185–386, `html[data-theme='dark']` 395–505), and a check asserts the dark block is byte-identical in its two required forms. All 84 referenced tokens are verified declared. |
| **Cross-cutting** — `--overlay-scrim` never declared | **Resolved** — it comes in with the verbatim token block and is what the dialog scrim uses. |
| **Cross-cutting** — epics cite only the Artifact URL | **Open**, and a ticket-body fix: #3087's sub-issues should cite the file path plus the `credit-*` names above. |
| Suggestion — `no-line-name` absent | **See the regression below.** |
| Suggestion — `border-radius:999px` vs `--radius-pill` | Now verbatim from `index.css`, which itself uses `999px` on that element. Faithful rather than fixed. |
| Suggestion — `money()` hardcodes `pl-PL` | **Still there.** Deliberate for a Polish seller's invoice fixture, but your point stands and the file now carries a related implementer note: the price inputs are `type="number"` and so render a dot while the invoiced column renders a comma — that mismatch must not ship. |

### One regression I introduced, not yours

Collapsing the panel onto an invoice-line grid **dropped three of the four `no-match` reasons**. `disposition-not-confirmed` survives as an in-row note with its remedy, but `no-line-by-name`, `no-line-name` and `quantity-exceeds-invoiced` have no home: a returned item that resolves to no invoice line has no row in a grid that iterates invoice lines, so it would **silently vanish** — exactly what § 22's "a `no-match` line must never silently vanish" rule forbids, and what your `no-line-name` suggestion was protecting.

The grid needs a *"returned items not matched to any invoice line"* block beneath it. Not yet drawn; recorded here rather than left to be discovered.

### This needs a backend change first

**#3171** — *resolve `ReturnLine.resolvedOrderLineId` at ingestion instead of matching invoice lines by name.* It is the prerequisite for the whole v9 design: without the join, OpenLinker cannot pre-fill a single quantity. It is in progress now, and the epic's FE sub-issues should not start ahead of it.

The mockup also names three further things it depends on that do not exist: a read exposing the refund figure beside the credit, per-line already-credited quantities (without which a second return can over-credit the same line — a correctness hole, not a convenience), and the shipping decision above.

### About the file in this PR

`docs/plans/mockups/correction-proposal-3087.html` is the **v3** copy and is now stale against v9 — and per § UX Mockups a stale mockup is worse than none, since it is the durable spec. It should be regenerated from v9 before this merges. Say the word and I will push it in this PR.

norbert-kulus-blockydevs:
### Follow-up is up: #3172 (must land before this)

The prerequisite I flagged above is now a real PR — **https://github.com/openlinker-project/openlinker/pull/3172** (`Closes #3171`).

It is what makes the v9 mockup honest. The picker this artifact deletes existed because nothing populated `ReturnLine.resolvedOrderLineId`, so #2374 matched disposed return lines against the issued invoice **by product name** and handed the resulting collision to the operator. #3172 resolves the join at ingestion instead — SKU, then unit price as the tie-break among same-SKU lines — and refuses rather than guessing when it genuinely cannot settle one.

Two things about #3172 that matter for reviewing the mockup:

- **It changes no existing behaviour.** Nothing reads `resolvedOrderLineId` yet; the slice makes the field trustworthy. Narrowing `return-correction-matching.domain-service.ts` onto it — and actually retiring the picker — is the next step, and that is the change this mockup's UI depends on.
- **The unresolved path stays reachable.** Where the rule cannot settle a line (no SKU and no price, nothing matched, or two order lines identical on both axes) the column stays `null` and the correction matcher keeps its current behaviour for that line. That is the one gap the mockup does not yet show — see the regression I named in my comment above, where three of four `no-match` reasons have no home in the invoice-keyed grid.

Nothing to do on this PR from my side; #3172 goes first.

piotrswierzy (commented):
## Tech Lead review

**Nothing has been pushed since the review.** `git log origin/3087-correction-proposal-ui..origin/3116-correction-proposal-mockup` is a single commit, `7c4c8146c`, dated 2026-09-10 08:36 — the review landed 12:52 the same day. The v9 rebuild described in the thread exists only as an Artifact; the file in this diff is still v3, picker and all (`choices`, `pickOption`, `pickCount` are all live at L201-224). So both prior findings stand verbatim, and the thread adds a third, larger one. The two siblings (#3119, #3122) shipped their fix rounds and are clean; this one has had no round at all.

### Findings

**[BLOCKING]** — `docs/plans/mockups/correction-proposal-3087.html:137`
> The irreversibility half of the § 5.8 lead is still missing. `docs/specs/product-spec-oms-returns-operator-ux.md:454-458` specifies it verbatim — *"Check the match before issuing — **a correction to a document already sent to {authority} cannot be withdrawn.**"* — and `grep -i 'withdrawn|undone|irreversib'` over the file returns **zero hits**. The page keeps only the reassuring half (*"A preview only — nothing is sent to the tax office until you confirm"*). Committing a mockup is committing the durable spec; this one actively reassures where the spec demands a warning about an unwithdrawable act, which is the § UX Mockups *"a stale mockup is worse than none, because it actively misleads"* failure on the highest-stakes sentence on the page. Sibling #3122 gets this right (`⚠ This can't be undone`), in the dialog rather than after it.

**[BLOCKING]** — the committed file is a version its own author has declared wrong
> The thread states, unambiguously, that two independent reviews concluded the picker is *"unanswerable and manufactured"*, that v9 deletes it, and that the honest version depends on **#3172** landing first (*"#3172 goes first"*). That is a statement that the artifact in this diff should not be the durable spec. § UX Mockups makes the repo copy the thing implementation and E2E tickets cite for the life of the feature, so merging v3 now hands every sub-issue of #3087 a central interaction that has already been retracted. Either push v9 (and hold on #3172), or say in the PR description why v3 is still the design of record — the rule requires one or the other, and a reviewer treats a PR that does neither as incomplete.

**[IMPORTANT]** — `:207` (`renderLine`, the `pick` branch)
> The ambiguous-line copy is truncated to *"This item is on the invoice twice. Which one was it?"*, dropping both the invoice number/count and the load-bearing clause the spec ends on: *"the correction amount is the same either way **unless these lines were priced differently**."* That clause is `candidatesPriceOrRateDiffer` — the one signal that tells an operator whether the pick is cosmetic or changes the credited amount — and it is rendered nowhere. The option buttons show a price, so the operator *could* infer it, but inference is exactly what the spec declines to rely on. § 5.8's footer sentence (*"OpenLinker will not issue corrections automatically until invoice lines carry a stable reference"*) is also absent.

**[IMPORTANT]** — `:277` (`setState`)
> `setState` toggles chip classes and re-renders; it never stamps the rendered surface. `grep -c 'body.dataset.state'` → **0**, and all 5 `data-state` attributes sit on `.demo-toolbar` buttons that are present in every state — so `[data-state="all-good"]` matches unconditionally and an E2E spec keyed on it asserts nothing. This is the identical finding raised against #3119 and #3122, **both of which have since fixed it**; the exemplar is `docs/plans/mockups/analytics-display-currency-picker.html`, which has 33 `body[data-state]` writes. One line: `document.body.dataset.state = name`.

**[IMPORTANT]** — the dark `--status-*` block diverges from `index.css` on **every** value
> Extracted the 15 declared dark status tokens and diffed against the `html[data-theme='dark']` block of `apps/web/src/index.css`: 15/15 differ. Among them `--status-disabled-fg: oklch(70% 0.010 270)` against the real `oklch(82% 0.010 270)` — a measured contrast regression, and the same specific value flagged on #3122. Both siblings now match the real block byte-for-byte; #3119's corrected block drops in unchanged. § UX Mockups requires real tokens *"so the mockup reads as 'this app,' not a generic prototype"*, and a mockup that renders darker-than-real disabled text will be copied into the implementation.

### What is right
Worth keeping through the rebuild: `nothing-correctable` still returns the **full body** with a per-line exclusion reason rather than an empty state — the documented behaviour and the easiest thing to lose. The propose-vs-issue boundary is clean (only CTA is "Record for review", post-record copy defers issuance to the invoice flow), matching § 22 Returns' *"holding a proposal confers no authority to issue one."* And no undeclared CSS custom property: 50 declared, 36 used.

### Verdict
❌ **Request changes** (posted as a comment — this review does not use `--request-changes`)

Priority: (1) restore the irreversibility warning; (2) decide v3-vs-v9 and say so in the description; (3) `body[data-state]`; (4) the dark block; (5) the ambiguous-line clause.

---

norbert-kulus-blockydevs:
## Pushed — `ac89f513d`. Both BLOCKINGs, all four IMPORTANTs, and the carried suggestions.

You were right that the branch had had no round at all, and right about the shape of the problem: the
file was still v3, so both prior findings stood verbatim and the third one — *"the committed file is a
version its own author has declared wrong"* — was the one that mattered. It is fixed by taking your
first option: **the grid is pushed, and the merge holds behind #3172.**

Two things I did beyond the fix list, and one defect I found that neither of us had.

---

### The spec moves with the mockup

You cited `§ 5.8` as the authority, correctly — and that is exactly why pushing the grid without
touching it would have left the repo carrying a spec and a mockup that contradict each other on the
screen's central interaction. That is the same *"a stale mockup is worse than none, because it actively
misleads"* failure, one file over, and a reader who stops at the spec acts on the retired design.

`docs/specs/product-spec-oms-returns-operator-ux.md` § 5.8 now carries a supersession pointer at the
top and an **Amendment (#3116)** at the bottom. The amendment argues the case rather than asserting it
— the picker's question is both unanswerable (`InvoiceLine` is `{name, quantity, unitPriceGross,
taxRate, unit?}`, no id, no SKU, and on an ambiguous line both candidates carry the same product name)
and unnecessary (OpenLinker builds the invoice from `order.items` in order,
`order-to-issue-invoice-command.mapper.ts:113`, and the return line already arrives with `offerId`,
`sku` and `unitPrice`) — and it is explicit about what does **not** change: a proposal is still never
an automatic issue, the irreversibility warning stays, a no-match line still never vanishes silently,
and the Hard AC survives restated for the new shape.

The original § 5.8 text is kept rather than deleted, because the amendment argues against it.

### `[hidden]` loses to a class selector — the committed file was broken at `file://`

Not in your review, not in mine, and it broke the sheet in the only way it is meant to be used.

`.returns-credit-note { display: flex }` and `.state-card { display: grid }` both beat the UA's
`[hidden]`. The Artifact host injects `[hidden]{display:none!important}`, so the sheet looked correct
there — while the **repo copy, opened at `file://`, rendered all six state elements at once**:

```
ELEMENTS THAT ARE [hidden] BUT STILL RENDER: 6
  DIV.alert alert--error            | data-state=detail-orphan
  SECTION.credit-unmatched          | data-state=-
  SECTION.returns-credit-note       | data-state=credit-absent-orphan
  DIV.state-card state-card--loading| data-state=detail-loading
  DIV.state-card state-card--error  | data-state=detail-error
  DIV.state-card state-card--error  | data-state=detail-unreadable
```

`file://` is precisely the consumption mode § UX Mockups mandates for an E2E baseline, and the one this
PR's own test plan describes. I only caught it by driving a real browser — reading the source cannot
see it, which is why it survived two reviews. The app hits the same trap and answers it per class
(`index.css`: `.raw-payload__body[hidden]`, `.bulk-destbar__panel[hidden]`); a single file generalises
it, because a per-class list is a list the next state forgets to join. Now **0**.

---

### Your findings

**BLOCKING — the irreversibility half of the § 5.8 lead.** Fixed, and it is
`RETURN_PROPOSAL_COPY.irreversible` **verbatim** rather than a paraphrase — in the panel lead (where
§ 5.8 puts it, and where the constant's own docblock says it belongs), beside the issue action, and in
the confirm dialog. Declared once in the script so the three cannot drift; beside the button it is
sliced to the fact half, because *"Check every line before you confirm"* is advice the lead already gave
and the operator has just taken. One thing worth stating plainly since you quoted the spec line:
`check-ui-vocabulary` (design rule P9) bans `authority` from operator copy, so § 5.8's *"sent to
{authority}"* is unsayable and the shipped constant says *"issued and sent"* — its docblock records
that trade, so this is the shipped wording, not a softening of yours.

**BLOCKING — the file is a version its author declared wrong.** Fixed by pushing the grid, which is
what you asked for first. The decision, its reasoning and the #3172 merge order are now in three
places that cannot quietly diverge: the PR description, the file's own header, and the spec amendment.

**IMPORTANT — the truncated ambiguous-line copy and `candidatesPriceOrRateDiffer`.** Superseded, and
replaced by something stronger rather than dropped: there are no candidates to compare, and the grid
shows every line's own price *and makes it editable*, which is more evidence than a price-divergence
flag. § 5.8's footer sentence is carried as `RETURN_PROPOSAL_COPY.noAutoIssue`, verbatim, in the
panel's ⓘ.

**IMPORTANT — `setState` never stamps the rendered surface.** Fixed, and gone a step further, because
the previous defence ("five frames cannot share one body attribute") was true and beside the point.
`body[data-state]` is stamped on every change — **and every state is settable without a click**:

```
?state=credit-unmatched   #state=detail-error   window.__setMockupState(name)
```

That is the half that matters for a baseline: driving the sheet by synthesising clicks on its own
switcher ends up testing the switcher. An unrecognised name is **ignored** rather than silently
rendering the default, so a typo in a ticket reads as "nothing happened" and not as a wrong screenshot.
`body[data-page-state]` and `body[data-credit-state]` always hold their own family's value — which is
where `page-loaded` lives, since it is the *absence* of a page-level override and so is never the state
in force. That distinction is written into the sheet's state index, because otherwise a ticket citing
`page-loaded` points an E2E spec at an attribute that never holds it.

**IMPORTANT — the dark `--status-*` block.** Resolved, and I verified it by diffing declarations rather
than by eye, since that is the distinction the finding draws:

| block | declarations | vs `index.css` |
|---|---|---|
| `:root` | 109 | identical |
| `html[data-theme='dark']` | 77 | identical |
| `prefers-color-scheme` copy | 77 | identical to the attribute form |

Zero undeclared custom properties (111 declared, 86 used). `--status-disabled-fg` is the real
`oklch(82% 0.010 270)`. `--overlay-scrim` arrives with the block and is what the dialog scrim uses.

---

### Carried suggestions

**`no-line-name` absent.** Fixed — and it was worse than a missing reason string. The grid iterates the
**invoice's** lines, so a returned item matching none of them has no row at all, which § 22 Returns
forbids in those words. That was named in the thread as a self-inflicted regression and left undrawn;
it is drawn now. New block, `body[data-state]="credit-unmatched"` — *"Came back, but not on this
invoice"* — carrying `no-line-by-name`, `no-line-name` and `quantity-exceeds-invoiced` with the reason
stated per item. `disposition-not-confirmed` keeps its grid row, because that item **is** on the
invoice. Untinted, per #2507: nothing is broken, these simply cannot be credited from here.

**`money()` hardcodes `pl-PL`.** Fixed as far as a fixture can be. The pin is a named `FIXTURE_LOCALE`
constant with `getBcp47Locale` and its `'en'` default named beside it, so an implementer reads it as a
fixture choice rather than copying a literal out of a formatter.

**`border-radius:999px` vs `--radius-pill`.** Deliberately **not** changed. The remaining occurrence
arrives inside the verbatim `index.css` block, which itself writes `999px` there — changing it would
break the byte-identity your dark-block finding asks for. Faithful beats tidy.

**Epics cite only the Artifact URL.** Fixed. #3087 and all eight sub-issues now carry the file path,
the `?state=` mechanism and the specific `data-state` values each implements. Three of those
annotations carry real news rather than a citation: **#3091** (*ambiguous-line picker*) is marked
**superseded — do not implement**, left open rather than closed so the decision stays visible;
**#3093** records that its `no-line-snapshot` is the sheet's `credit-lines-unauthoritative`, and gains
`credit-unmatched`; **#3090**'s scope changed from "one row per return line" to the invoice-line grid.

**`Closes #N` against a feature-branch base.** Correct — removed, with a note in the description saying
why. #3116 closes when the epic branch merges.

---

### Verification

Chromium, against the committed `file://` copy:

- all **14** states resolve from `?state=` and stamp the expected `body[data-state]`;
- an unknown name renders no wrong screen;
- the unmatched block is present in `credit-unmatched` and absent in `credit-draft` — it was visible in
  **both** before the `[hidden]` fix;
- the irreversibility sentence is in the rendered text;
- the live grid recomputes: line 1 quantity `1 → 0` moves the total `−189,00 → −378,00 PLN`;
- light and dark both render from the real tokens, **zero page errors** in either.

`pnpm check:invariants` passes, `check-css-structure` included — the file parses with zero defects.

### Still open, deliberately

**This does not merge until #3172 does.** Three further reads the sheet draws do not exist at all — the
refund figure beside the credit, per-line already-credited quantities (without which a second return
can over-credit the same line, a correctness hole rather than a convenience), and the shipping
decision. Each is named in the file's own *"backend work that does not exist yet"* note rather than
implied to be built, and the last one is why the reconciliation is the loudest thing on the screen:
delivery is a real invoice line that a name-join can never match, so today it is silently never
credited.

piotrswierzy (commented):
## Tech Lead re-review (`7c4c8146` → `ac89f513`)

One commit, and it is the v9 rebuild that the previous round existed only as an Artifact link for. **Both blockers are closed, and closed properly rather than papered over.** The picker is gone; `§ 5.8` now opens with a pointer to its own amendment and the amendment argues the retirement rather than announcing it — the "unanswerable" half (`InvoiceLine` is `{name, quantity, unitPriceGross, taxRate, unit?}` with no id and no SKU, so an ambiguous row asks the operator to invent an answer) and the "unnecessary" half (OL builds the invoice from `order.items` in order, so position *is* the line number) are both correct against `main`, as is the `resolvedOrderLineId: null` at `returns.service.ts:444` and the long-unused `CorrectionLine.newUnitPriceGross`. The irreversibility copy is now in all three places the amendment claims, hoisted to one `IRREVERSIBLE` constant so the lead, the action and the dialog cannot drift. The `--status-*` blocks are byte-identical to `apps/web/src/index.css` in both themes — I diffed all 34 dark and all 33 light declarations. `document.body.setAttribute('data-state', effective)` is now the single writer, and `?state=` / `#state=` addressability with an unknown name deliberately ignored is a step above what any of the 22 mockups on `main` offers. Naming #3171/#3172 as a hard prerequisite in both the file header and the spec is the right call and the right place for it.

What is left is a state table that carries one member no contract produces and is missing one that a contract does.

### Findings

**[IMPORTANT]** — `docs/plans/mockups/correction-proposal-3087.html:2421`
> `credit-lines-unauthoritative` ("Lines don't match") has nothing behind it. `ReturnCorrectionProposalOutcomeValues` is a closed five-member union — `proposed` / `nothing-correctable` / `no-invoice` / `no-line-snapshot` / `no-disposed-lines` — and `no-line-snapshot` is already spent on `credit-no-content` one state above, which is why the two share the `Lines not stored` badge and differ only in message. Nothing in the file, the spec amendment or `return-correction-proposal.types.ts` says what produces it. Under the positional join this design rests on there *is* a plausible new condition (a stored snapshot shorter than the position a `resolvedOrderLineId` names), but that is me supplying the argument. Either ground it — one comment naming the #3172-era condition, the way every other state here is grounded — or drop it. A state an implementing ticket can cite and then find nothing to wire is exactly the dead end the § 5.8 amendment was written to remove.

**[IMPORTANT]** — `docs/plans/mockups/correction-proposal-3087.html:2404`
> `nothing-correctable` has no state. It is a declared outcome whose own docblock says the proposal body is *still* returned in full "so the operator reads every exclusion reason" while no `order_changes` row is opened — so it is a real screen: the grid renders, every line carries a reason, and there is nothing to confirm. `credit-unmatched` is not it, and its own comment says so ("a return can legitimately have creditable lines and unmatched ones at once"). The amendment's restated Hard AC — a draft carrying a blocked line or unmatched items must be visibly different *before* the operator reaches the confirm button — is unsatisfiable for the case where the confirm button should not be reachable at all, because no state draws it. Add `credit-nothing-correctable`, or say in the file why it collapses into `credit-unmatched`.

**[SUGGESTION]** — `docs/specs/product-spec-oms-returns-operator-ux.md:479`
> The amendment's positional claim ("invoice line *N* is `order.items[N-1]`") holds for the product prefix only — `order-to-issue-invoice-command.mapper.ts:113` maps the items and then *pushes* `toShippingLines(...)` after them. That is harmless for returns, which never match a shipping line, and the amendment separately gets the consequence right ("the first thing that comparison catches is delivery cost, which a name-join can never match"). Worth one clause so the next reader does not derive a stronger invariant than the mapper gives.

### Verdict

🔄 **Approve with changes** — the two blockers are closed, the design is sound and the spec amendment is the right shape for a retracted design. Ground or drop `credit-lines-unauthoritative`, and decide what `nothing-correctable` renders, before the frontend sub-issues cite this file.

`MERGEABLE` / `CLEAN`. **No CI checks reported on this branch at all** — not "Lint skipped", none reported — which is expected for a docs-only PR here and is stated rather than read as passing.

---

piotrswierzy (commented):
## Tech Lead re-verification (head unchanged at `ac89f51`)

`main` has not moved; the four previously-closed findings stand closed (the commit is the v9 rebuild, `IRREVERSIBLE` is hoisted to one constant and used in all three places, the dark `--status-*` block is byte-identical to `apps/web/src/index.css`, and `document.body.setAttribute('data-state', …)` is the single writer). The one IMPORTANT is unchanged and re-verified against the head blob: `credit-lines-unauthoritative` still appears 3× in the state table while no member of the closed 5-value `ReturnCorrectionProposalOutcomeValues` produces it, and `nothing-correctable` still has **zero** occurrences and therefore no state — leaving the amendment's own restated Hard AC unsatisfiable for the outcome that returns the full proposal body with nothing confirmable.

**Dependency status:** #3172 (and its task #3171, *resolve `ReturnLine.resolvedOrderLineId` at ingestion*) are both still **open**. This PR names them as hard prerequisites, so it is grounded on unbuilt behaviour by design — fine for a mockup, but the implementing tickets must not read it as shipped.

**#3241 does not affect this PR's grounding.** It reverses ADR-041 decision 5's `thresholdRef` indirection in `sales_document_rules` / `SalesDocumentCondition`; this mockup references neither (`thresholdRef`, `sales_document_rules`: 0 occurrences) and sits in the returns correction-proposal area. No overlap.

**CI:** there are **no checks at all** on `3116-correction-proposal-mockup`. `ci.yml` triggers on `pull_request: branches: [main]` and this PR targets `3087-correction-proposal-ui`, so nothing ran — absent, not passing. It will first get CI when the stack rebases onto `main`.

### Findings
No new findings. The single standing IMPORTANT (`credit-lines-unauthoritative` ungrounded / `nothing-correctable` stateless) is unaddressed.

### Verdict
🔄 **Approve with changes** — resolve the state-table vocabulary against `ReturnCorrectionProposalOutcomeValues` before this is cited by an implementing ticket.

---

piotrswierzy (commented):
## Tech Lead re-verification (head unchanged at `ac89f513`)

Base is the stack branch `3087-correction-proposal-ui`; `MERGEABLE` / `CLEAN`, **no checks reported** on this branch (absent, not passing). The standing finding re-verifies in the file, and both prerequisites are still unmerged.

### Findings

**IMPORTANT (stands) — the state table and the core outcome union do not agree, in both directions.** At this head:

- `credit-lines-unauthoritative` appears **3 times** (the state-switcher button at `:1546`, the `data-state` inventory at `:2316`, and the state definition at `:2427`) and is produced by **no member** of the closed 5-value `ReturnCorrectionProposalOutcomeValues` (`return-correction-proposal.types.ts:98-104` — `proposed`, `nothing-correctable`, `no-invoice`, `no-line-snapshot`, `no-disposed-lines`). Nothing in the mockup grounds it.
- `nothing-correctable` has **0 occurrences** and no state — yet it is a declared outcome that returns the **full proposal body with nothing confirmable**, i.e. a screen an operator really can reach and one whose whole point is showing the per-line exclusion reasons.

So the amendment's own restated Hard AC — every outcome has a reachable, named state — is unsatisfiable for `nothing-correctable`, while a state exists for a thing the backend cannot answer. Either add the `nothing-correctable` state and drop or ground `credit-lines-unauthoritative`, or state in the file which core change is expected to introduce it. `no-line-snapshot` is likewise absent (0 occurrences) and is the second-most-likely state an operator hits on a pre-#1297 document.

**Dependency status (both still open, neither merged):**

| | State | Note |
|---|---|---|
| #3171 (issue) | **OPEN** | CORE — resolve `ReturnLine.resolvedOrderLineId` at ingestion |
| #3172 (PR) | **OPEN**, `MERGEABLE`, head moved to `0cd02365` | `reviewDecision: CHANGES_REQUESTED`; under separate review this round |

These are hard prerequisites: the whole positional-ambiguity story this mockup draws rests on line resolution existing. #3241 re-verified as not affecting the grounding.

### Verdict
🔄 **Approve with changes**

---

piotrswierzy (commented):

Reviewed against `docs/frontend-architecture.md` § UX Mockups, `docs/frontend-ui-style-guide.md`, and the shipped returns backend (`docs/architecture-overview.md` § 22 Returns). Base is the epic branch `3087-correction-proposal-ui`, not `main`, so nothing here reaches `main` today.

The design decision is sound and the evidence for it is real: I confirmed `order-to-issue-invoice-command.mapper.ts` builds the invoice from `order.items` in order, so a row's position genuinely *is* the line number a correction addresses, and the return line does arrive carrying `offerId` / `sku` / `unitPrice`. Retiring the picker is the right call. My findings are about the **merge gate the file states**, which I could not reproduce against `main`.

---

### Findings

**[BLOCKING]** — the stated prerequisite is already shipped; the real one is not, and is not named
`docs/plans/mockups/correction-proposal-3087.html:22-25`, again at `:2259`, and `docs/specs/product-spec-oms-returns-operator-ux.md` § 5.8 Amendment

> All three say, in bold, that *"`ReturnLine.resolvedOrderLineId` is hardcoded `null` (`returns.service.ts:444`) and nothing in the tree populates it"*, and gate every frontend sub-issue of #3087 on #3172. Checked against `main`, each clause is wrong:
>
> - **#3171 is shipped and wired.** `apps/worker/src/sync/return-order-line-resolver.service.ts:38` exists, is provided in `sync-worker.module.ts:142`, and is called from `marketplace-return-sync.handler.ts`. `ReturnsService.resolveOrderLines` runs `resolveReturnLineOrderLine` and persists via `claimOrderLineResolution` (`returns.service.ts:178-199` → `return.repository.ts:1051`).
> - **`returns.service.ts:444` is `matchOrphanToOrder`**, which does not touch `resolvedOrderLineId` at all. The `null` at `:552` is the *initial* value in `recordReturn` for an operator-authored return, which is correct, not a stub.
> - **The actual remaining prerequisite is the one `architecture-overview.md` § 22 names** — *"Consuming the field — narrowing `return-correction-matching.domain-service.ts` onto it and retiring the picker — is a follow-up"*. That is still open: the matcher emits `status: 'ambiguous'` with `candidates` at `return-correction-matching.domain-service.ts:189-195`, unchanged.
>
> This matters because it is the file's own guardrail. An implementer who checks *"is the join populated? yes"* concludes the gate is clear and starts a frontend sub-issue against a matcher that still produces a line shape the grid cannot render. § UX Mockups is explicit that *"a stale mockup is worse than none, because it actively misleads"* — a wrong merge order is exactly that failure. Fix: name the matcher narrowing as the prerequisite, drop the `returns.service.ts:444` citation, and state that the join itself already landed with #3171.

**[IMPORTANT]** — `ambiguous` is renderable by the backend today and has no shape anywhere in the sheet
`docs/plans/mockups/correction-proposal-3087.html` (zero occurrences of `ambiguous`)

> `ReturnCorrectionLineStatusValues` is still `['matched', 'ambiguous', 'no-match']`, and `return-correction-proposal.service.ts:308` treats `'ambiguous'` as correctable — so a `'proposed'` outcome can and does carry ambiguous lines (the service logs their count at `:130`). The grid has no row treatment for one and the `credit-unmatched` block does not cover it, because its reasons are the four `ReturnCorrectionNoMatchReason` values. That is the intended end state *after* the matcher is narrowed, but until then the sheet is silent about a line status the backend emits. Worth one sentence in the "backend work that does not exist yet" note saying so explicitly, rather than leaving the absence to be read as an oversight.

**[SUGGESTION]** — `credit-absent-orphan` is not settable by its own name
`docs/plans/mockups/correction-proposal-3087.html:2679-2690`

> `setStateByName` accepts only `PAGE_STATES ∪ Object.keys(STATES)`, and `credit-absent-orphan` is in neither — it is stamped on `body[data-credit-state]` only when `pageState === 'detail-orphan'`. So `?state=credit-absent-orphan` hits the "unrecognised name is ignored" path, and a ticket citing it as a `data-state` value will find `body[data-state]="detail-orphan"` instead. Everything about that is deliberate and the code comment explains it; it just is not in the state index, which is the thing sub-issues are told to cite. One line there closes it.

---

### Verified, not taken on trust

- **Token fidelity is exact, not approximate.** Diffed both blocks against `apps/web/src/index.css`: `:root` **111/111** declarations byte-identical, `html[data-theme='dark']` **77/77** identical, **zero** value divergences, and **zero** undeclared custom properties used anywhere in the file. This is a genuine verbatim paste, which is what § UX Mockups means by "real design tokens".
- **The `[hidden]` fix is real and correctly diagnosed** (`:480-489`). `.returns-credit-note { display: flex }` does beat the UA `[hidden]`, the Artifact host was masking it, and the committed file would have rendered every state at once at `file://` — the one consumption mode § UX Mockups mandates for an E2E baseline. Catching this by driving a browser rather than reading the source is the right instinct, and the fix generalises rather than adding another per-class entry.
- **Outcome coverage is complete.** All five `ReturnCorrectionProposalOutcome` values map to a state (`proposed`→`credit-draft`, `no-disposed-lines`→`credit-nothing-disposed`, `no-invoice`→`credit-no-invoice`, `no-line-snapshot`→`credit-no-content`, `nothing-correctable`→`credit-unmatched`), and all four `ReturnCorrectionNoMatchReason` values are rendered with per-item reasons.
- **The state contract exceeds the § UX Mockups minimum.** `?state=` / `#state=` with `hashchange`, an in-page switcher covering all 13 settable names, `body[data-state]` stamped on every change, and an unknown name ignored rather than silently defaulting. The reasoning that a baseline must not drive the sheet by synthesising clicks on its own switcher is right and worth keeping.
- **The spec moves with the mockup, in both directions** — supersession pointer at the head of § 5.8 and an amendment at the foot, with an explicit "what does NOT change" list and the Hard AC restated for the new shape. That is the obligation § UX Mockups places on an implementing change, met properly.

---

### Summary

This commits a well-argued redesign — replacing an unanswerable per-row picker with an invoice-line grid whose row position *is* the line number, plus a channel-refund reconciliation that catches delivery cost a name-join structurally cannot — and amends the spec that specified the thing it retires, so the two cannot contradict each other. Token fidelity, state addressability and outcome coverage all check out exactly as claimed. The one substantive problem is that the prerequisite it gates itself on was already delivered by #3171, while the prerequisite that genuinely remains (narrowing the matcher so `ambiguous` stops being emitted) is named nowhere — and since that gate is repeated in three places and is the artifact's own protection against premature frontend work, it should be corrected before this becomes the design of record.

### Merge Readiness

**Request changes** — not on the design, which I would approve as-is, but on the three-place factual gate. Correcting the prerequisite (and adding one sentence on `ambiguous`) is a small edit to prose that is otherwise carefully written; the `credit-absent-orphan` note is optional.

norbert-kulus-blockydevs:
## Addressed — tech-lead review (2026-09-15)

**[BLOCKING]** — the stated prerequisite (`returns.service.ts:444`, "hardcoded `null`", gated on #3172) was already shipped by #3171/#3172; the real remaining gate — narrowing `return-correction-matching.domain-service.ts` off its name-join and onto `ReturnLine.resolvedOrderLineId` — was named nowhere. Fixed in all three places the review named:

- `correction-proposal-3087.html` file header (§0): now states the join is resolved and trustworthy, and names the matcher narrowing as the actual gate.
- Same file, "This screen needs backend work that does not exist yet" note: item 1 rewritten the same way.
- `product-spec-oms-returns-operator-ux.md` § 5.8 Amendment (#3116): same correction.

**[IMPORTANT]** — `ambiguous` is renderable by the backend today (`return-correction-matching.domain-service.ts:308` still treats it as correctable, so a `'proposed'` outcome can carry one) and had no mention in the sheet. Added one sentence to the same "backend work" note stating this explicitly, rather than leaving the absence to read as an oversight.

**[SUGGESTION]** — `credit-absent-orphan` settable by its own name: checked against the current file and it's already handled — `setStateByName` has an explicit third branch (`if (name === 'credit-absent-orphan') { applyPage('detail-orphan'); return true; }`) and the state index at the "which attribute holds what" table already lists it under the credit-note family (cited via `body[data-credit-state]`, which is correct — a page state outranks a credit state on `body[data-state]` by design, and the doc says so). No change needed.

Pushed as `48427445c`. `pnpm check:invariants`' `check-css-structure` still parses the file cleanly (3589 blocks balanced), and the `<b>` tag count is still balanced after the edits (77/77).

piotrswierzy (approved):
## `/pr-review` — Tech Lead review (re-review of `48427445`)

**Verdict: ✅ Approve.** Both findings from the previous review are addressed, and addressed at the level they were raised rather than papered over.

### The BLOCKING finding is resolved, in all three places

The prior review's objection was not that the gate existed but that it named the wrong one: the mockup and spec asserted `ReturnLine.resolvedOrderLineId` "is hardcoded `null` at `returns.service.ts:444` and nothing populates it", which stopped being true when #3171 shipped. The commit corrects the claim in the file header, the "backend work that does not exist yet" note, and the spec's own § 5.8 amendment — the three sites the review named — and replaces it with the prerequisite that genuinely remains.

I verified the replacement claim rather than taking it on trust:

- `libs/core/src/returns/domain/domain-services/return-correction-matching.domain-service.ts` contains **zero** references to `resolvedOrderLineId`. The matcher has not been narrowed onto the field.
- `:182-197` still returns `status: 'ambiguous'` with a `candidates` list and `selectedOriginalLineNumber: null` when `feasible.length > 1`. So the shape the grid has no row for is still reachable.

That makes the new gate accurate where the old one was stale, which is the whole of what was asked.

### The IMPORTANT finding is resolved too, and stated rather than implied

The previous review flagged that an `ambiguous` line is still emitted under a `'proposed'` outcome — `return-correction-proposal.service.ts:308` treats such lines as correctable — while the retired-picker grid has no shape for one. Both the header and the numbered note now say so explicitly ("a `'proposed'` outcome can carry a line this sheet cannot render"). Naming the unrenderable case in the sheet that cannot render it is the right disposition for a mockup: it is a design document, so recording the gap is the deliverable, and inventing a row shape for `ambiguous` would pre-empt the matcher decision rather than wait for it.

### Findings

**[SUGGESTION]** — `docs/plans/mockups/correction-proposal-3087.html`
> The gate now correctly points at the matcher narrowing, which `docs/architecture-overview.md` § 22 names as a follow-up to #3171 but which carries no issue number in either the mockup or the spec. Since this text exists precisely so an implementer can check "is the gate clear?", a number to check would close the loop the correction opens — the same argument the correction itself makes.

**[Positive]** — the commit message states what was wrong, what is true now, and which review finding each edit answers. A docs correction that records why the earlier claim was made and when it expired is more useful than one that silently overwrites it.

### Merge readiness

✅ **Approve.** The earlier verification stands unchanged and was not disturbed by this commit: `:root` 111/111 and dark 77/77 tokens byte-identical to `apps/web/src/index.css` with no undeclared properties, all five proposal outcomes and all four no-match reasons covered, and the spec supersession done in both directions. Ready to merge.

piotrswierzy (approved):
## Tech Lead review — #3123

Read against `docs/architecture-overview.md` § 22 *Returns* (the #2374 bullet), `libs/core/src/returns/domain/types/return-correction-proposal.types.ts` on `main`, `docs/specs/product-spec-oms-returns-operator-ux.md` § 5.8, and the conventions of the 18 existing sheets in `docs/plans/mockups/`.

### What is right

- **Location and shape are correct.** `docs/plans/mockups/correction-proposal-3087.html` sits with its siblings, is standalone, renders at `file://`, and carries its own state index — the `listings-variant-centric-2823.html` / `operational-settings.html` shape.
- **The spec moves with it, and moves properly.** § 5.8 gets a supersession pointer at the top (`product-spec-oms-returns-operator-ux.md:454-458`) *and* an `#### Amendment (#3116)` at the bottom, with the original body preserved and argued against rather than edited. That is the append-only discipline `docs/architecture/adrs/README.md` sets for ADRs, applied to a spec, and it is the right call — a mockup contradicting a live spec is the failure this PR names.
- **`?state=` over click-synthesis is the right E2E affordance**, and ignoring an unrecognised name rather than falling back to the default is what keeps a typo from producing a confidently wrong screenshot.
- **The retirement argument is sound and grounded.** `InvoiceLine` really does carry no id and no SKU, and `order-to-issue-invoice-command.mapper.ts:113` really does build the invoice from `order.items` in order. Asking the operator to pick between two identically-named candidates was unanswerable.

---

### [IMPORTANT] — `nothing-correctable` has no state, and it is not the same thing as `credit-unmatched`

`ReturnCorrectionProposalOutcomeValues` on `main` is exactly five members (`return-correction-proposal.types.ts:98-105`): `proposed`, `nothing-correctable`, `no-invoice`, `no-line-snapshot`, `no-disposed-lines`. The sheet covers four of them — `credit-no-invoice`, `credit-no-content`, `credit-nothing-disposed`, and the draft — and **`nothing-correctable` appears nowhere in the file** (zero occurrences).

It is a distinct surface, not a variant of the draft. Its own docblock (`return-correction-proposal.types.ts:87-90`):

> *a document exists and its lines were classified, but none is correctable. The proposal body is STILL returned in full so the operator reads every exclusion reason; no row is opened, because a slot must not be held for a proposal with nothing to confirm.*

So the screen is: the grid renders, every row carries an exclusion reason, the total is zero, **and no ADR-044 slot is held** — which is the fact an operator needs, because it means there is nothing to confirm and nothing occupying the proposal slot. `credit-unmatched` does not cover it: the sheet's own comment (`correction-proposal-3087.html:2440-2443`) defines that state as *"a state of the SAME draft … a return can legitimately have creditable lines and unmatched ones at once"*. § 22 requires every non-proposing exit to be a **named** outcome; this one is currently behind whatever the draft happens to render.

### [IMPORTANT] — `credit-lines-unauthoritative` specifies a state no backend can produce

`correction-proposal-3087.html:2429-2433` defines *"Lines don't match" — "This invoice's lines cannot be matched to the stored copy a correction uses"*, with a switcher entry at line 1544 and a slot in the state index at line 2318.

Nothing in `libs/core` produces it. It is not a member of `ReturnCorrectionProposalOutcomeValues`, and `grep -rn "unauthoritative\|linesAuthoritative" libs/core/src` returns zero. `credit-no-content` already renders `no-line-snapshot` (the pre-#1297 refusal). This file is the durable spec for #3087, so an implementer working from it builds an unreachable screen — the exact mirror of the omission above. Either ground it in a real outcome (and say which) or remove it.

### [IMPORTANT] — the stated hard prerequisite has no issue filed

Both the sheet (`correction-proposal-3087.html:26-30, 2263-2271`) and the spec amendment (`product-spec-oms-returns-operator-ux.md:497-503`) state that `return-correction-matching.domain-service.ts` must be narrowed onto `ReturnLine.resolvedOrderLineId`, that until then the matcher *"still emits `status: 'ambiguous'` with a `candidates` list, a shape this grid cannot render"*, and that **"no frontend sub-issue of #3087 should start"**.

Disclosing it is right, and far better than a silent contradiction. But **no issue exists for it**: `gh issue list --search "matcher narrowing resolvedOrderLineId"` returns nothing, and #3087's body lists only #2374, #2376, #3029, #3116, #3123, #3172. A gate with no issue number is unenforceable — and it is the same failure class the sibling PR #2951 is in this repo to name, where ADR-005 deferred `webhook_deliveries` compaction *"to a #711 follow-up that was never filed"*. `docs/architecture-overview.md` § 22 also records that the matcher still joins by name today and calls the narrowing *"a follow-up"* with no owner.

File it, cite it in both places, and make it a declared predecessor of #3087's frontend children.

### [SUGGESTION] — the merge-order warning in the description is stale

The body's *"⚠️ Merge order — this holds behind #3172"* and *"`ReturnLine.resolvedOrderLineId` is hardcoded `null` (`returns.service.ts:444`)"* no longer hold: **#3172 is MERGED**, #3171 is closed, and `returns.service.ts:444` on `main` is now inside the orphan-attribution guard. The sheet itself was updated (line 25 reads *"the field is trustworthy today"*); the PR description was not. The blocker has cleared in the merge direction, so this only needs the description corrected.

### [SUGGESTION] — the `ambiguous` row shape

Separate from the issue-filing point above: since `ReturnCorrectionLineStatusValues` (`return-correction-proposal.types.ts:41`) still carries `ambiguous` and its docblock is emphatic that it *"lists EVERY candidate"* and is *"never collapsed into `matched`"*, the sheet would be stronger for showing what such a row looks like even in a degraded read-only form — rather than leaving the one live status it cannot render as prose only. As it stands the grid iterates the invoice, so an `ambiguous` line has no row at all, which is the shape § 22 forbids (*"a no-match line still never vanishes silently"*) arriving by a different door.

---

## Summary

A thorough, well-argued replacement of § 5.8's per-row picker with an invoice-line grid, and the spec amendment that goes with it is a model of how to retire a design without editing the record. The `[hidden]`-versus-class-selector find is a genuine catch that only driving a real browser would produce. The gap is outcome coverage: of the five members of the live `ReturnCorrectionProposalOutcomeValues`, one (`nothing-correctable`) has no state at all while the sheet specifies one (`credit-lines-unauthoritative`) that no backend can emit — and the prerequisite both documents call a hard gate has no issue behind it.

## Merge Readiness

🔄 **Approve with changes** — nothing blocking. The base is the feature branch `3087-correction-proposal-ui`, so these can land there before the epic merges, but they should land before a frontend sub-issue is picked up from this sheet.

## Priority fixes

1. File the matcher-narrowing issue and cite it from the sheet, from § 5.8's amendment, and from #3087 as a declared predecessor.
2. Add a `nothing-correctable` state — grid rendered, every exclusion reason stated, and explicitly *no proposal slot held*.
3. Ground or remove `credit-lines-unauthoritative`; it maps to no member of the outcome union.
4. Correct the stale `#3172` merge-order warning in the description.

piotrswierzy (approved):
## Tech Lead review — `docs(mockups): correction-proposal UI mockup`

Docs-only, two files, and the spec moves with the mockup rather than after it — which is the part that matters.

### ✅ The spec and the mockup move together

`docs/frontend-architecture.md` § UX Mockups: *"a stale mockup is worse than none, because it actively misleads."* The same is true one file over, and §5.8 would have been exactly that — a live spec describing a retired picker — if the mockup had landed alone. The supersession pointer at the top **and** the amendment at the bottom, with the original text kept *"because the amendment argues against it, not because it still holds"*, is the right treatment: a reader arriving from a link into the middle of §5.8 gets the warning before the stale prose.

The **What does NOT change** list is the most valuable part of the amendment. A design revision that silently drops "a proposal, never an automatic issue" or the irreversibility warning would be a real regression dressed as a redesign, and enumerating the invariants that survive is what stops that.

### ✅ The picker retirement is the right call, and grounded

Both halves of the argument check out against the code:

- **Unanswerable** — `InvoiceLine` is `{name, quantity, unitPriceGross, taxRate, unit?}`, no id and no SKU, and on an ambiguous line both candidates carry the same name. I confirmed that shape in `return-correction-matching.domain-service.ts`, whose own header says the same thing in the same terms.
- **Unnecessary** — the invoice is built from `order.items` in order, so a row's position *is* the line number. Nothing needs to be asked.

Making `newUnitPriceGross` editable is a genuine capability recovery, not scope creep: `CorrectionLine` has always carried it and the returns path used only `newQuantity`, so a discount or agreed reduction was unexpressible. Same for the refund reconciliation — OL holds both figures and compares them nowhere, and delivery cost is precisely what a name-join can never match.

### ✅ The `[hidden]` finding

> *The sheet looked right in the Artifact because the host injects `[hidden]{display:none!important}` — the repo copy, opened at `file://`, **rendered all six state elements at once**.*

That's the single most useful thing in this PR. The committed file was broken in **exactly** the consumption mode § UX Mockups mandates for an E2E baseline, and only in that mode — so every check short of opening the repo copy in a browser would have passed it. Generalising the fix in one rule rather than per class is right, for the stated reason: *"a per-class list is a list the next state forgets to join."*

`?state=`/`#state=` with an **unrecognised name ignored rather than rendering the default** is the other good call — a typo in a ticket reading as "nothing happened" instead of producing a confidently wrong screenshot is the difference between a baseline you can trust and one you can't.

Verifying the token blocks by diffing (109/109, 77/77) rather than by eye is the right method for a claim like "pasted verbatim".

### 🟡 IMPORTANT — the PR description's merge-order banner is stale, and its own spec amendment says so

The description leads with:

> **⚠️ Merge order — this holds behind #3172.** `ReturnLine.resolvedOrderLineId` is hardcoded `null` (`returns.service.ts:444`) and nothing in the tree populates it.

That is no longer true on `main`. I checked:

- `libs/core/src/returns/domain/domain-services/return-order-line-resolution.domain-service.ts` — exists
- `apps/worker/src/sync/return-order-line-resolver.service.ts` — exists, wired into `sync-worker.module.ts` and `marketplace-return-sync.handler.ts`
- `returns.service.ts:552`'s `resolvedOrderLineId: null` is the **creation** path — the field is claimed afterwards by `claimOrderLineResolution` (fill-in-when-NULL), which is the designed shape, not "nothing populates it". The cited line number `444` no longer matches either.

Your own amendment already has this right:

> ***#3171 / #3172 resolved it**: the field is populated at ingestion today and is trustworthy. What remains a hard prerequisite … is narrowing `return-correction-matching.domain-service.ts` onto that field instead of matching by product name.*

And that remaining prerequisite is real — I confirmed the matcher on `main` still normalises names and still emits `ambiguous` with a `candidates` list, which this grid cannot render.

So the blocker is correct but **misidentified in the description**, and the description is what someone reads before merging. Please update it to name the matcher narrowing rather than #3172 — otherwise the epic's frontend work sits behind a gate that has already opened, while the gate that is actually closed goes unnamed.

### 🟢 SUGGESTION

Fourteen states is a lot to keep aligned by hand between the sheet, the issues that cite them, and the eventual E2E baseline. Worth confirming the sheet's own state index is the single list those citations were taken from — the convention's value is that a ticket names a state and there is exactly one way to find it.

### Notes

- No CI checks ran on this PR (docs-only against a feature branch). The description reports `pnpm check:invariants` green locally including `check-css-structure`; worth a run on the epic branch before it merges to `main`.
- The `Closes #3116` omission is correctly reasoned — base is a feature branch, GitHub wouldn't fire it, and `CLAUDE.md` forbids closing by hand.
- The review-findings table is genuinely useful: each row says what changed and why, and two entries say *"superseded"* / *"faithful rather than fixed"* rather than claiming a fix that wasn't made.

**Verdict:** ✅ Approve. Please correct the merge-order banner in the description before merging — everything in the files themselves is right.

Issue #3096, opened 2026-09-10
[EPIC] Frontend — Fulfilment work list + detail view


norbert-kulus-blockydevs:
## Problem / Context

`GET /fulfillment/works/:workId` is exported by `features/fulfillment/api/fulfillment.api.ts` but never called - confirmed by #3029 (only `.list`, `.listByOrder` and `.applyAction` have callers). There is no way to land on one fulfilment task: no deep link, no place for the negotiation status, the hold reasons, the per-line pick progress, or the framing that explains why one order can produce several of these rows.

## Mockup

Reviewed, interactive HTML mockup - built on the real design system, all states: https://claude.ai/code/artifact/67cf8980-5604-4331-a9b4-a5869e387cb2

Committed to the repo by #3117 as `docs/plans/mockups/fulfillment-work-detail-3096.html`. It is the design of record: where a sub-issue and the mockup disagree, the mockup wins and the sub-issue body is corrected.

## What is already shipped, and must not be rebuilt

Most of this epic's mechanism exists on `main` from #2410 / #2411 / #2416. An audit before planning found:

| Concern | State |
|---|---|
| `/fulfillment` worklist, lanes, filters, pagination, four page states | shipped |
| `FulfillmentTaskActions`, `FulfillmentTaskActionDialog`, `useFulfillmentTaskActionMutation` | shipped |
| `readFulfillmentConflict` - both 409 codes already branched | shipped |
| `fulfillment-task.copy.ts` / `fulfillment-worklist.copy.ts` - the whole vocabulary the mockup ports | shipped |
| `OrderFulfillmentTasksPanel` - the order-page embed | shipped |
| `apiClient.fulfillment.get(workId)` | exists, no caller |
| The detail **page**, its route, and the derived summary | missing - this epic |

Two gaps the original breakdown did not cover were added as sub-issues after that audit: `expeditedAt` is served by the backend and absent from the frontend contract (#3247), and nothing was going to write the architecture note or the ADR (#3248).

## Decision: clicking a row opens its own page, not a split screen

Two shapes were on the table. A **split screen** keeps the list on the left half and the task's details on the right half, with nothing navigating. Its **own page** means a row click takes the operator to a page for that one task, with a back link. The mockup does the second, and it is right for three reasons.

[...]

Issue #3118, opened 2026-09-10
design(oms): OMS-onboarding UI mockup


norbert-kulus-blockydevs:
## Problem / Context

Before implementation starts, OMS onboarding and pack-bench discoverability (#3104) needs a reviewed design so every sub-issue builds against
the same agreed layout, states and copy instead of each developer inventing its own.

## Proposed Solution

Add the reviewed interactive HTML mockup as a real file in the repo (`docs/plans/mockups/`,
following the existing `docs/plans/mockups/sales-document-routing.html` precedent), so it survives
independently of the hosted preview. The mockup was designed and reviewed as a live Artifact first —
this issue is the step of committing that file to the codebase, not of doing the design work again.

## Classification

**Type**: Frontend
**Layer**: Interface (design artifact)
**File(s)**: `docs/plans/mockups/oms-onboarding-3104.html`

## Dependencies

- None — first sub-issue of #3104, everything else in the epic reads from this mockup

## Assumptions

- The mockup already exists as a reviewed Artifact and only needs to be committed as a static file
- No interactivity beyond what a static HTML file can do (the Artifact's own JS demo harness ships
  as-is for reference, but is not itself the product)

## Acceptance Criteria

- [ ] `docs/plans/mockups/oms-onboarding-3104.html` exists and renders standalone (no external API calls)
- [ ] PR links back to the reviewed Artifact: https://claude.ai/code/artifact/2c2cc062-1042-4e46-b196-4514dda9a793
- [ ] PR merges into the epic branch, not `main`

norbert-kulus-blockydevs:
Delivered by #3125, merged into `3104-oms-onboarding-ui` (2026-09-14). Closing manually — the PR's base was an epic branch, not `main`, so GitHub never auto-closed this.

Evidence: commit `6dcdb30b0` adds `docs/plans/mockups/oms-onboarding-3104.html`.

**Not on `main` yet.** It reaches `main` via the epic PR #3244, currently a draft. Closing the issue because the mockup is done; the epic tracks the merge.

PR #3124, opened 2026-09-10
docs(mockups): fulfillment-work-detail UI mockup

### LINKED ISSUE #3117: design(fulfillment): fulfilment-work-detail UI mockup
## Problem / Context

Before implementation starts, Fulfilment work list + detail view (#3096) needs a reviewed design so every sub-issue builds against

## Proposed Solution

## Classification

**Type**: Frontend
**File(s)**: `docs/plans/mockups/fulfillment-work-detail-3096.html`

## Dependencies

- None — first sub-issue of #3096, everything else in the epic reads from this mockup

## Assumptions

## Acceptance Criteria

- [ ] `docs/plans/mockups/fulfillment-work-detail-3096.html` exists and renders standalone (no external API calls)
- [ ] PR links back to the reviewed Artifact: https://claude.ai/code/artifact/67cf8980-5604-4331-a9b4-a5869e387cb2

norbert-kulus-blockydevs:
## Summary
- Adds the reviewed `fulfillment-work-detail` mockup as a static file at `docs/plans/mockups/fulfillment-work-detail-3096.html`
- Reviewed Artifact (live preview): https://claude.ai/code/artifact/67cf8980-5604-4331-a9b4-a5869e387cb2
- Closes #3117, first sub-issue of #3096

## Test plan

### DIFF
--- a/docs/plans/mockups/fulfillment-work-detail-3096.html
+++ b/docs/plans/mockups/fulfillment-work-detail-3096.html
@@ -1,0 +1,312 @@
+<title>Fulfilment Work Detail</title>
+<link rel="stylesheet" href="https://fonts.googleapis.com/css2?family=IBM+Plex+Sans:wght@400;500;600;700&family=IBM+Plex+Mono:wght@400;500;600&display=swap">
+<style>
+:root, [data-theme='light'] {
+  --bg-canvas: oklch(99% 0.003 80); --bg-shell: oklch(97.5% 0.004 80);
+  --bg-surface: #ffffff; --bg-surface-muted: oklch(96% 0.005 80); --bg-surface-hover: oklch(93% 0.006 80);
+  --border-subtle: oklch(93.5% 0.006 80); --border-default: oklch(88% 0.008 80); --border-strong: oklch(78% 0.010 80);
+  --text-primary: oklch(20% 0.012 80); --text-secondary: oklch(38% 0.010 80);
+  --text-muted: oklch(52% 0.008 80); --text-disabled: oklch(70% 0.005 80); --text-on-primary: oklch(18% 0.012 50);
+  --accent-primary: oklch(68% 0.18 50); --accent-primary-hover: oklch(62% 0.19 50);
+  --accent-primary-soft: oklch(96% 0.04 60); --accent-primary-border: oklch(85% 0.10 55); --accent-ring: oklch(68% 0.18 50 / 0.30);
+  --status-success: oklch(54% 0.14 150); --status-success-soft: oklch(96% 0.04 150); --status-success-border: oklch(85% 0.08 150); --status-success-fg: oklch(36% 0.12 150);
+  --status-warning: oklch(72% 0.16 85); --status-warning-soft: oklch(96% 0.05 85); --status-warning-border: oklch(85% 0.10 85); --status-warning-fg: oklch(42% 0.12 80);
+  --status-error: oklch(58% 0.20 25); --status-error-soft: oklch(96% 0.04 25); --status-error-border: oklch(85% 0.10 25); --status-error-fg: oklch(42% 0.16 25);
+  --status-info: oklch(56% 0.14 245); --status-info-soft: oklch(96% 0.03 245); --status-info-border: oklch(85% 0.08 245); --status-info-fg: oklch(40% 0.12 245);
+  --status-disabled-soft: oklch(95% 0.005 80); --status-disabled-border: oklch(85% 0.008 80); --status-disabled-fg: oklch(38% 0.010 80);
+  --space-2:.5rem; --space-3:.75rem; --space-4:1rem; --space-5:1.5rem;
+  --radius-md:8px; --radius-lg:10px; --radius-xl:14px; --radius-pill:9999px;
+  --shadow-xs: 0 1px 2px rgb(15 18 26 / 0.04);
+  --shadow-soft: 0 1px 2px rgb(15 18 26 / .04), 0 6px 16px rgb(15 18 26 / .06);
+  --shadow-focus: 0 0 0 3px var(--accent-ring);
+  --font-sans: 'IBM Plex Sans', ui-sans-serif, system-ui, sans-serif;
+  --font-mono: 'IBM Plex Mono', ui-monospace, SFMono-Regular, Menlo, monospace;
+}
+@media (prefers-color-scheme: dark) {
+  :root:not([data-theme='light']) {
+    --bg-canvas: oklch(14% 0.005 270); --bg-shell: oklch(16% 0.006 270);
+    --bg-surface: oklch(19% 0.007 270); --bg-surface-muted: oklch(24% 0.009 270); --bg-surface-hover: oklch(28% 0.010 270);
+    --border-subtle: oklch(24% 0.010 270); --border-default: oklch(30% 0.012 270); --border-strong: oklch(42% 0.014 270);
+    --text-primary: oklch(96% 0.006 270); --text-secondary: oklch(78% 0.010 270);
+    --text-muted: oklch(60% 0.012 270); --text-disabled: oklch(42% 0.012 270); --text-on-primary: oklch(16% 0.012 50);
+    --accent-primary: oklch(72% 0.18 50); --accent-primary-hover: oklch(78% 0.18 50);
+    --accent-primary-soft: oklch(28% 0.08 50); --accent-primary-border: oklch(40% 0.14 55); --accent-ring: oklch(72% 0.18 50 / 0.40);
+    --status-success-soft: oklch(28% 0.07 150); --status-success-border: oklch(40% 0.10 150); --status-success-fg: oklch(80% 0.12 150);
+    --status-warning-soft: oklch(30% 0.08 85); --status-warning-border: oklch(45% 0.11 85); --status-warning-fg: oklch(85% 0.12 85);
+    --status-error-soft: oklch(28% 0.09 25); --status-error-border: oklch(42% 0.13 25); --status-error-fg: oklch(84% 0.14 25);
+    --status-info-soft: oklch(28% 0.06 245); --status-info-border: oklch(42% 0.10 245); --status-info-fg: oklch(82% 0.10 245);
+    --status-disabled-soft: oklch(26% 0.006 270); --status-disabled-border: oklch(36% 0.010 270); --status-disabled-fg: oklch(70% 0.010 270);
+    --shadow-soft: 0 1px 2px rgb(0 0 0 / .4), 0 8px 24px rgb(0 0 0 / .4);
+    --shadow-xs: 0 1px 2px rgb(0 0 0 / .4);
+  }
+}
+:root[data-theme='dark'] {
+  --bg-canvas: oklch(14% 0.005 270); --bg-shell: oklch(16% 0.006 270);
+  --bg-surface: oklch(19% 0.007 270); --bg-surface-muted: oklch(24% 0.009 270); --bg-surface-hover: oklch(28% 0.010 270);
+  --border-subtle: oklch(24% 0.010 270); --border-default: oklch(30% 0.012 270); --border-strong: oklch(42% 0.014 270);
+  --text-primary: oklch(96% 0.006 270); --text-secondary: oklch(78% 0.010 270);
+  --text-muted: oklch(60% 0.012 270); --text-disabled: oklch(42% 0.012 270); --text-on-primary: oklch(16% 0.012 50);
+  --accent-primary: oklch(72% 0.18 50); --accent-primary-hover: oklch(78% 0.18 50);
+  --accent-primary-soft: oklch(28% 0.08 50); --accent-primary-border: oklch(40% 0.14 55); --accent-ring: oklch(72% 0.18 50 / 0.40);
+  --status-success-soft: oklch(28% 0.07 150); --status-success-border: oklch(40% 0.10 150); --status-success-fg: oklch(80% 0.12 150);
+  --status-warning-soft: oklch(30% 0.08 85); --status-warning-border: oklch(45% 0.11 85); --status-warning-fg: oklch(85% 0.12 85);
+  --status-error-soft: oklch(28% 0.09 25); --status-error-border: oklch(42% 0.13 25); --status-error-fg: oklch(84% 0.14 25);
+  --status-info-soft: oklch(28% 0.06 245); --status-info-border: oklch(42% 0.10 245); --status-info-fg: oklch(82% 0.10 245);
+  --status-disabled-soft: oklch(26% 0.006 270); --status-disabled-border: oklch(36% 0.010 270); --status-disabled-fg: oklch(70% 0.010 270);
+  --shadow-soft: 0 1px 2px rgb(0 0 0 / .4), 0 8px 24px rgb(0 0 0 / .4);
+  --shadow-xs: 0 1px 2px rgb(0 0 0 / .4);
+}
+* { box-sizing: border-box; }
+body { margin:0; min-width:320px; background:var(--bg-canvas); color:var(--text-primary); font-family:var(--font-sans); font-size:.875rem; line-height:1.5; }
+h1,h2,h3,p { margin:0; }
+button:focus-visible{ outline:none; box-shadow:var(--shadow-focus); border-radius:var(--radius-md); }
+
+.page{ max-width:1100px; margin:0 auto; padding:28px 24px 60px; }
+.eyebrow{ font-size:.75rem; font-weight:700; letter-spacing:.08em; text-transform:uppercase; color:var(--accent-primary-hover); }
+.title{ font-size:1.75rem; font-weight:600; letter-spacing:-.02em; margin-top:6px; }
+
+.compare{ display:grid; grid-template-columns:1fr 1fr; gap:0; margin-top:16px; border:1px solid var(--border-default); border-radius:var(--radius-lg); overflow:hidden; max-width:78ch; }
+@media (max-width:640px){ .compare{ grid-template-columns:1fr; } }
+.compare__col{ padding:14px 16px; }
+.compare__col:first-child{ border-right:1px solid var(--border-default); background:var(--bg-surface-muted); }
+@media (max-width:640px){ .compare__col:first-child{ border-right:none; border-bottom:1px solid var(--border-default); } }
+.compare__col--this{ background:var(--bg-surface); }
+.compare__label{ font-family:var(--font-mono); font-size:.6875rem; text-transform:uppercase; letter-spacing:.06em; color:var(--text-muted); margin-bottom:4px; }
+.compare__text{ font-size:.8125rem; color:var(--text-secondary); }
+.compare__text b{ color:var(--text-primary); }
+
+.example{ display:flex; gap:12px; align-items:flex-start; margin-top:14px; padding:14px 16px; border-radius:var(--radius-lg); background:var(--accent-primary-soft); border:1px solid var(--accent-primary-border); max-width:78ch; }
+.example__glyph{ font-family:var(--font-mono); font-size:.875rem; font-weight:700; color:var(--accent-primary-hover); flex-shrink:0; }
+.example__text{ font-size:.8125rem; color:var(--text-secondary); }
+.example__text b{ color:var(--text-primary); }
+
+.layout{ display:grid; grid-template-columns:340px 1fr; gap:20px; margin-top:24px; align-items:start; }
+@media (max-width:860px){ .layout{ grid-template-columns:1fr; } }
+
+.list{ border:1px solid var(--border-default); border-radius:var(--radius-lg); background:var(--bg-surface); box-shadow:var(--shadow-xs); overflow:hidden; }
+.list__header{ padding:12px 14px; border-bottom:1px solid var(--border-subtle); font-size:.75rem; font-weight:700; text-transform:uppercase; letter-spacing:.05em; color:var(--text-muted); }
+.list-item{ display:block; width:100%; text-align:left; border:none; background:none; cursor:pointer; padding:13px 14px; border-bottom:1px solid var(--border-subtle); }
+.list-item:last-child{ border-bottom:none; }
+.list-item:hover{ background:var(--bg-surface-muted); }
+.list-item--selected{ background:var(--accent-primary-soft); box-shadow:inset 3px 0 0 var(--accent-primary); }
+.list-item__order{ font-weight:600; font-size:.875rem; }
+.list-item__where{ font-size:.75rem; color:var(--text-muted); margin-top:2px; }
+.list-item__pill{ display:inline-flex; align-items:center; gap:6px; margin-top:8px; padding:2px 8px; border-radius:var(--radius-md); font-family:var(--font-mono); font-size:.6875rem; font-weight:500; letter-spacing:.02em; text-transform:uppercase; border:1px solid var(--status-disabled-border); background:var(--status-disabled-soft); color:var(--status-disabled-fg); }
+.list-item__pill--attn{ border-color:var(--status-warning-border); background:var(--status-warning-soft); color:var(--status-warning-fg); }
+.list-item__pill-dot{ width:6px; height:6px; border-radius:999px; background:currentColor; opacity:.8; }
+
+.detail{ border:1px solid var(--border-default); border-radius:var(--radius-lg); background:var(--bg-surface); box-shadow:var(--shadow-xs); min-height:420px; }
+.detail__hero{ padding:24px 26px 20px; border-bottom:1px solid var(--border-subtle); }
+.detail__hero-sub{ font-size:.8125rem; color:var(--text-muted); margin-bottom:10px; }
+.detail__hero-status{ font-size:1.375rem; font-weight:700; display:flex; align-items:center; gap:10px; }
+.detail__hero-status--attn{ color:var(--status-warning-fg); }
+.detail__hero-dot{ width:10px; height:10px; border-radius:999px; background:var(--text-disabled); flex-shrink:0; }
+.detail__hero-dot--attn{ background:var(--status-warning); }
+.detail__hero-detail{ font-size:.875rem; color:var(--text-secondary); margin-top:8px; }
+.detail__body{ padding:22px 26px; }
+.detail__section{ margin-bottom:24px; }
+.detail__section:last-child{ margin-bottom:0; }
+.detail__section-title{ font-size:.75rem; font-weight:700; text-transform:uppercase; letter-spacing:.05em; color:var(--text-muted); margin-bottom:10px; }
+
+.item-row{ display:flex; justify-content:space-between; align-items:center; padding:11px 0; border-bottom:1px solid var(--border-subtle); }
+.item-row:last-child{ border-bottom:none; }
+.item-row__name{ font-weight:500; font-size:.9375rem; }
+.item-row__qty{ font-size:.8125rem; color:var(--text-secondary); }
+.item-row__done{ color:var(--text-secondary); }
+.item-row__done-tick{ color:var(--status-success-fg); }
+
+.fact-grid{ display:grid; grid-template-columns:1fr 1fr; gap:14px 24px; }
+.fact{ display:grid; gap:2px; }
+.fact__label{ font-size:.75rem; color:var(--text-muted); }
+.fact__value{ font-size:.9375rem; font-weight:500; }
+
+.hold-banner{ display:flex; gap:10px; padding:13px 15px; border-radius:var(--radius-md); background:var(--status-warning-soft); border:1px solid var(--status-warning-border); }
+.hold-banner__title{ font-weight:600; font-size:.875rem; color:var(--status-warning-fg); }
+.hold-banner__desc{ font-size:.8125rem; color:var(--text-secondary); margin-top:3px; }
+
+.detail__actions{ padding:18px 26px; border-top:1px solid var(--border-subtle); display:flex; flex-wrap:wrap; gap:8px; align-items:center; }
+.detail__actions-label{ font-size:.75rem; color:var(--text-muted); width:100%; margin-bottom:2px; }
+
+.button{ display:inline-flex; align-items:center; justify-content:center; gap:8px; border:1px solid var(--accent-primary-border); border-radius:var(--radius-md); padding:0 18px; height:2.25rem; background:var(--accent-primary); color:var(--text-on-primary); font-family:inherit; font-size:.875rem; font-weight:600; cursor:pointer; box-shadow:var(--shadow-xs); }
+.button:hover{ background:var(--accent-primary-hover); }
+.button--secondary{ border-color:var(--border-default); background:var(--bg-surface); color:var(--text-primary); }
+.button--secondary:hover{ background:var(--bg-surface-muted); border-color:var(--border-strong); }
+.button--danger{ border-color:color-mix(in oklab, var(--status-error) 80%, black); background:var(--status-error); color:#fff; }
+.button:disabled{ cursor:not-allowed; opacity:.5; box-shadow:none; }
+
+.alert{ display:flex; gap:10px; padding:12px 14px; border-radius:var(--radius-md); border:1px solid; margin:0 26px 18px; }
+.alert--warning{ background:var(--status-warning-soft); border-color:var(--status-warning-border); }
+.alert--error{ background:var(--status-error-soft); border-color:var(--status-error-border); }
+.alert--success{ background:var(--status-success-soft); border-color:var(--status-success-border); }
+.alert__text{ font-size:.8125rem; color:var(--text-secondary); }
+.alert__text b{ color:var(--text-primary); }
+
+.done-empty{ padding:60px 26px; text-align:center; color:var(--text-muted); }
+
+.demo-toolbar{ display:flex; align-items:center; gap:10px; margin-top:22px; padding:10px 14px; background:var(--bg-surface-muted); border-radius:var(--radius-md); flex-wrap:wrap; }
+.demo-toolbar__label{ font-family:var(--font-mono); font-size:.6875rem; text-transform:uppercase; letter-spacing:.08em; color:var(--text-muted); margin-right:2px; }
+.chip{ display:inline-flex; align-items:center; gap:.25rem; height:1.5rem; padding:0 .6rem; border:1px solid var(--border-subtle); border-radius:var(--radius-pill); background:var(--bg-surface); color:var(--text-secondary); font-size:.75rem; font-weight:500; cursor:pointer; }
+.chip.chip--active{ background:var(--accent-primary-soft); border-color:var(--accent-primary-border); color:var(--accent-primary-hover); font-weight:600; }
+</style>
+
+<div class="page">
+  <p class="eyebrow">Fulfilment</p>
+  <h1 class="title">Who's packing it, right now</h1>
+
+  <div class="compare">
+    <div class="compare__col">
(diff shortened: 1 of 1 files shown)


norbert-kulus-blockydevs:
> **Note: this is an expanded review.** The original pass below (frontend/UX) is preserved verbatim; three additional lenses — OpenLinker domain specialist, e-commerce operations, and senior backend/system architect — have been folded in as one consolidated review rather than as separate comments.

## PR Review — #3124 (`docs(mockups): fulfillment-work-detail UI mockup`)

Reviewed as a **frontend UI/UX artifact** per project convention (`docs/frontend-architecture.md § UX Mockups`): a committed mockup here is not decorative, it's "the durable spec that implementation and E2E tasks reference for the life of the feature." Every modeled field/action was cross-checked against the *actual shipped* `libs/core/src/fulfillment/` code (ADR-054, #2391/#2392/#2406/#2409), not just against the mockup's own internal logic.

### 🔴 Blocking

1. **The mockup offers operator buttons for actions the backend will never let an operator take.**
   - Row 3 (`open|submitted`) renders **"Accept"** / **"Turn down"** (`accept` / `reject`) as clickable operator buttons.
   - `ACTION_LABEL` additionally defines `accept_cancellation` → "Confirm cancel" and `reject_cancellation` → "Keep it — refuse cancel".
   - But the real read model is explicit and load-bearing on this exact point — `fulfillment-supported-actions.types.ts`:
     > "The four holder replies are absent by construction... `accept`, `reject`, `accept_cancellation` and `reject_cancellation` are the HOLDER's answers, recorded by #2399's `recordAcceptance` / `recordRejection` off an executor response. They are never an operator's act, so offering them on an operator read model would be the drift this exists to remove, pointing the other way."
   - `OPERATOR_INVOCABLE_ACTIONS` (`fulfillment-work-view.types.ts`) is a closed list — `schedule, hold, release_hold, mark_in_progress, close, force_cancel, expedite, release_expedite` — and `isOperatorInvocableAction` rejects everything else. If this mockup is implemented as drawn, clicking "Accept" on that row specs a request the API will 400.

2. **`request_cancellation` is drawn as an operator button, but the read model doesn't expose it yet — and the mockup gives no hint it's aspirational.**
   - Row 1 offers `request_cancellation` ("Ask to cancel"). The real `OPERATOR_INVOCABLE_ACTIONS` set deliberately omits it — its own docblock names two live blockers (the executor is host-resolved, and the input needs a PII-carrying `RoutingShipTo` this read model doesn't hold). It's a real future action, but demoing it unqualified in a document meant to be the implementation spec tells the next engineer it's ready to wire today.

3. **The "on hold" row uses an impossible `status` value, and that specific mistake masks a real summarization gap.**
   - Row 2 sets `status:'on_hold'` directly. But per `fulfillment-supported-actions.types.ts`: *"Nothing in the tree writes `status = 'on_hold'`: `placeHold` inserts the hold row and does not touch `status`... A held work therefore reads `status: 'open'` with a non-empty `activeHolds[]`."* This is restated verbatim in the read-model's own docblock (`fulfillment-work-view.types.ts`): *"`activeHolds` is the authority on heldness, NOT `status`... rendering `status` alone would show 'Open' on a held row."*
   - Feed the mockup's own `SUMMARY` lookup the value combination that actually occurs in production (`open|unsubmitted` + non-empty `activeHolds`) and there is **no matching key** — it falls through to the generic fallback (`attn:false`, label = `w.status.replace('_',' ')` → "open"). That's exactly the bug the real docblock warns against, reproduced by the mockup's own logic once you correct the input to match reality.

4. **New (backend/architect lens) — the action-legality mismatch is not confined to rows 1 and 3; it is systemic across all four demoed rows once checked against the real `deriveSupportedActions` rule.** Running each of the mockup's four `WORKS` entries through the actual derivation (`fulfillment-supported-actions.types.ts`) rather than trusting the hand-authored `supportedActions` arrays:
   - **Row 1** (`in_progress|accepted`, 0 holds, not terminal): the real legal-and-operator-invocable set is `hold`, `force_cancel`, `expedite`. The mockup instead offers `mark_in_progress` — which is **only legal from `open`/`scheduled`/`on_hold`**, never from `in_progress`; a work that is already being packed cannot legally be told to "start" again. The action the mockup is presumably reaching for here is `close` ("Mark complete"), which *is* legal for `in_progress` and is entirely absent from this row.
   - **Row 2** (`on_hold`, 1 active hold): real legal set is `release_hold`, `hold`, `force_cancel`, `expedite`. The mockup shows only `release_hold`/`force_cancel` — it's missing `hold` itself, which the derivation's own comment states is deliberately **not** suppressed by an existing hold ("a second hold for a second reason is legitimate, up to the limit" — `FULFILLMENT_HOLD_ACTIVE_LIMIT = 10`).
   - **Row 3** (`open|submitted`, assigned to DHL): once the two illegal holder-reply buttons are removed per finding #1, the real legal-and-operator-invocable set is actually `schedule`, `hold`, `mark_in_progress`, `force_cancel`, `expedite` — **five real actions**, not zero. As currently drawn, the mockup implies that while "waiting for an answer" from a holder, the operator has nothing to do but wait for the two (illegitimate) accept/reject buttons — which understates real operator agency during negotiation and needs to be shown once corrected (see the e-commerce note below).
   - **Row 4** (`cancelled`, terminal): `supportedActions: []` is correct — verified, no issue here.
   - **`expedite`/`release_expedite` (#2416, spec D22) is never demonstrated anywhere in the mockup**, despite being legal on every single non-terminal row and being one of only 8 actions an operator can ever invoke. A whole action category from the real vocabulary is entirely unrepresented.
   - Net: this is not two isolated slip-ups on top of an otherwise-correct model — the mockup's action-derivation logic diverges from the real rule on **three of the four rows**, which reinforces rather than softens the existing ❌ verdict.

5. **New (backend/architect lens) — line rendering silently loses a legitimate "done" state.** `FulfillmentWorkLineView` carries `cancelledQuantity` alongside `totalQuantity`/`fulfilledQuantity`, and the real invariant is `fulfilledQuantity + cancelledQuantity <= totalQuantity` — a line is finished once nothing further will happen to it, which can be reached through a *mix* of fulfilled and cancelled units, not fulfilled alone. The mockup's line objects never populate `cancelledQuantity`, and its `done` check (`l.fulfilledQuantity >= l.totalQuantity`) has no way to represent a line that finished via a partial cancellation — such a line would render "2 of 3 packed" forever, never reaching the done/✓ state, misstating a line that is in fact settled.

### 🟡 Important

6. **No `data-state` (or equivalent) marks the four demonstrated work states.** `docs/frontend-architecture.md § UX Mockups` requires: *"Every distinct UI state the mockup demonstrates... gets a stable `data-state` (or equivalent) attribute value and a corresponding button in an in-page state switcher... referenced by name — never by a free-text description."* The `demo-toolbar`'s three outcome buttons correctly carry `data-outcome="..."`, but the four work rows (being-packed / on-hold / awaiting-answer / cancelled) carry no equivalent selector — a task or E2E spec referencing "the on-hold state" has no unambiguous hook, only "click the second row."

7. **New (OpenLinker domain lens) — no stale-action/conflict-refresh affordance is demonstrated on the actual detail pane, only in the abstract demo-toolbar.** `supportedActions` is computed server-side alongside an optimistic-concurrency `version` token, and a stale action legitimately answers 409 with a refreshed action set. The demo-toolbar's `version_conflict` outcome message ("Someone else just changed this... try again") is good copy, but nothing in the mockup shows the *result* of that retry — i.e. that the action list itself should refresh after a 409, which is the entire mechanism the token exists to drive. Worth at least a comment/annotation noting the refreshed-actions step, since a future implementer reading only the rendered HTML wouldn't see it.

8. **New (e-commerce ops lens) — the "waiting for an answer" state gives no sense of how long the operator has been waiting, and the backend has no automated escalation for this.** Per the fulfillment handshake design, a `submitted` work whose holder never answers just stays `submitted` — there is deliberately no timeout-as-rejection sweep yet (that's tracked as its own future issue). Since operators have no automated backstop today, the UI is the *only* place staleness can be surfaced, and `createdAt` is already on the real view model. The mockup doesn't use it here (it's used elsewhere, e.g. "Started" in the fact grid) — surfacing "waiting since {time}" on the negotiation-status line itself would give an operator the cue to intervene manually (e.g. `force_cancel` and re-route) that the system currently can't give them automatically.

### 🟢 Minor / good practice worth noting

- **New (e-commerce ops lens) — the top-of-page "Orders vs. This page" comparison panel and the worked example ("two separate tasks... one per warehouse") are genuinely good orientation design** for a screen whose entire reason for existing (multi-location order splitting) is non-obvious to anyone coming from the plain order list. Worth keeping as-is when this becomes a real onboarding/empty-state pattern.
- **New (backend/architect lens) — line-item display correctly derives its own state from counters** (`done = fulfilledQuantity >= totalQuantity`) rather than inventing a per-line status field — this is exactly right per the DESIGN §5.2 "counters, never per-line statuses" rule, and worth crediting since it's easy to get backwards.
- **New (backend/architect lens) — the ≤10 active-hold cap and its "not suppressed by an existing hold" nuance are never demonstrated** (only one hold, on one row, is ever shown). Not blocking — the real cap-exceeded rejection path doesn't need full coverage in v1 of a mockup — but worth a follow-up state if this doc gets revised.

### ✅ Verified correct (worth naming, since a sibling mockup in this same batch got it wrong)

- **Avoids the #2507 "tint every row green" anti-pattern** that #3119 and #3120 (same PR batch) were flagged for. The code comment states the rule explicitly and follows it: `attn: true` is the only thing that earns colour; an ordinary working/waiting/finished row stays plain ink with a neutral pill.
- `HOLD_LABEL`'s 8 keys are a byte-exact match for the real `HoldReasonValues` (`libs/core/src/order-lifecycle/domain/types/hold-reason.types.ts`).
- The two demoed conflict outcomes, `version_conflict` and `action_not_legal`, are byte-identical to the real `FulfillmentWorkConflictCodeValues` union, including the semantic distinction (retryable-with-fresh-token vs. not) — a precise, load-bearing detail to get right and it's right.
- Correctly separates "why it's stuck" (from `activeHolds`) from the headline status pill, rather than conflating them into one field.
- Correctly demonstrates the empty-actions / finished state (`supportedActions: []` → "Nothing left to do").
- No internal aggregate vocabulary ("fulfillment work", internal `ol_fwork_*` id) leaks into operator-facing copy — consistent with the epic-#2412 UI naming rule.

### Verdict

**❌ Request changes.** The expanded pass reinforces rather than softens this: the visual/interaction craft is strong and the anti-pattern avoidance is genuinely better than its siblings, but the action vocabulary is wrong on three of the four demoed rows once checked against the real `deriveSupportedActions` rule — not just the two holder-only-action rows originally flagged, but also a `mark_in_progress` offered on already-in-progress work (should be `close`), a missing `hold` on an already-held work, and an entirely absent `expedite`/`release_expedite` action category. Add to that an `on_hold` status value nothing ever writes, and a line-rendering model with no way to show a partially-cancelled-but-finished line. Since this mockup is meant to be the implementation spec, all of the above need correcting before the next PR builds against it: rebuild every row's `supportedActions` from the real derivation rather than by hand, fix the on-hold row to `status:'open'` + non-empty `activeHolds` (with a real `SUMMARY` entry for that combination), and add the `data-state` hooks. The stale-action-refresh and waiting-duration points are smaller, worthwhile follow-ups in the same pass.

piotrswierzy (changes requested):

**Architecturally the most disciplined of its batch.** The vocabulary rule is fully respected — no occurrence of *authority*, *posture*, *FulfillmentWork* or "fulfilment work" anywhere (grepped); the page calls it a packing job in plain language. **`supportedActions` is consumed, never recomputed** — the action row is a pure `w.supportedActions.map(...)` with an explicit "Nothing left to do" for the empty array, which is exactly ADR-054's requirement that legality is a server answer. The optimistic-concurrency token is depicted (`version` on every fixture), the retryable-vs-terminal distinction is drawn correctly in operator language, and **lines carry counters, never per-line statuses** ("0 of 1 packed") per ADR-054. Colour discipline is deliberate and commented (*"`attn: true` is the ONE thing that earns colour"*), with every coloured state also carrying a text label.

### BLOCKING — `HOLD_LABEL` is a second copy of a table the docs forbid mirroring, and seven of eight labels differ

All eight `HoldReasonValues` are present and correctly keyed — but `frontend-architecture.md` (#2411) is explicit:

> `fulfillment` … imports `HoldReasonValues`, `HOLD_REASON_COPY` and `holdReasonLabel` from the `orders` barrel **rather than mirroring them** … a second frontend copy would need a second guard script for one union.

| reason | shipped (`order-hold.types.ts`) | mockup |
|---|---|---|
| `stock-shortfall` | Stock shortfall | Out of stock here |
| `payment-review` | Payment review | Payment is being reviewed |
| `address-invalid` | Address invalid | Address needs a fix |
| `awaiting-amendment` | Awaiting change | Waiting on an amendment |
| `awaiting-customer-confirmation` | Awaiting buyer | Waiting on the customer |
| `external` | Held by channel | Held by an outside system |
| `operator` | Held by operator | Held by an operator |

An implementer following this mockup must either ship different copy than it shows, or re-mirror the table and break the documented single-source rule (also pinned by `scripts/check-hold-reason-mirror.mjs`). Resolve to the shipped labels, or state explicitly that this is a **proposed rewording of the shared table** — which would change the orders surface too.

### BLOCKING — the negotiation axis is incomplete in the one place it matters

`SUMMARY` is keyed `status|requestStatus`, which correctly honours both axes — but it maps only four pairs, and **`cancellation_requested` and `cancellation_rejected` appear nowhere**: no fixture, no label, no action outcome. `cancellation_rejected` is the entire justification for the two-axis split (ADR-054: *"cancelling work a holder has ALREADY ACCEPTED is a negotiation that holder may refuse, and a merged axis has no way to say 'we asked and they said no'"*).

Worse, the fallback for an unmapped pair is `w.status.replace('_',' ')`, which **discards the negotiation axis entirely** — so `in_progress|cancellation_requested` renders as plain "in progress", exactly the collapse the model forbids. The mockup composes the axes correctly and then omits the state that proves they are orthogonal.

### BLOCKING — no `data-state` for the states the page is about

`.demo-toolbar` uses `data-outcome` and covers only the three action responses. The four work states plus the empty-selection state (`Pick an order on the left.`) carry no stable attribute and are reachable only by clicking a list row. `data-outcome` is a fine "or equivalent" for the responses; the primary states have nothing.

### IMPORTANT — `.list__header` renders `4 orders`

The list holds four **tasks**, and the page's own `.example` block two elements above says *"Order #9F2E1A … is **two separate tasks** here, one per warehouse. That split is the whole reason this page exists."* The header then counts those tasks and calls them orders — contradicting the page's own explainer on the one distinction it exists to teach. Epic #2412's naming rule makes "fulfilment task" the operator term; `4 tasks` is both correct and the count the fixtures support.

### IMPORTANT — the `version_conflict` branch doesn't re-render the actions

`runAction()` (~292). The 409 answer is *"a refreshed set"* of `supportedActions` — that is why the response carries one, and why a client must not recompute legality locally. The mockup shows a banner (*"Someone else just changed this… try again"*) and leaves the action row byte-identical, so it specifies a UI where "try again" means pressing the button that was just rejected. Re-render `.detail__actions` from a changed `supportedActions`.

### SUGGESTIONS

- `.button--danger { color:#fff }` and `color-mix(in oklab, var(--status-error) 80%, black)`. `--text-inverse` is **not** right here (it is dark text), so `#fff` on a filled status colour is a genuine **token gap** rather than a violation — worth a comment saying so. `--status-error-strong` exists and would replace the `color-mix`.
- `.list-item--selected` conveys selection with background + inset shadow but sets no `aria-current`/`aria-pressed`; the real app's rail primitives own that contract.

---
### Cross-cutting (this mockup wave)

- **`data-state` on the rendered state, not just the switcher** — `document.body.dataset.state = name`, plus a value + button for every dialog and gallery card. § UX Mockups requires exactly one unambiguous way to find a state.
- **`--overlay-scrim` is never declared** — hardcoded `rgba(22,32,43,.45)` scrims do not respond to the dark theme (`index.css:319/504` has the real token).
- `Closes #N` will not fire against a feature-branch base; `CLAUDE.md` forbids hand-closing.

norbert-kulus-blockydevs:
Pushed a fix addressing both reviews (mine + @piotrswierzy's):

- Removed `accept`/`reject`/`accept_cancellation`/`reject_cancellation` and `request_cancellation` as operator buttons — never operator-invocable.
- Fixed `status:'on_hold'` → held is now derived from `activeHolds[]`, never a status value.
- `HOLD_REASON` labels now byte-match `HOLD_REASON_COPY` instead of a re-worded copy.
- Added `cancellation_requested`/`cancellation_rejected` demo rows + `SUMMARY` entries.
- Added `scheduled` and `incomplete` (short_picked) demo rows.
- Added the `expedite`/`release_expedite` pair + an `expeditedAt` badge (previously declared but never exercised).
- Conflict outcomes now re-render `supportedActions` and name the specific work id / version / action in the alert body.
- Added `data-state`/`data-work-id`/`data-action`/`data-testid` throughout, plus a named "jump to a state" switcher.
- Added minimal reason-capture dialogs for `hold`/`force_cancel`/`release_hold`.
- List header now says "N tasks" (was "N orders").
- `--status-error-strong` replaces the `color-mix()` border; dark-theme status tokens now byte-match `index.css`.

Live artifact preview updated: https://claude.ai/code/artifact/67cf8980-5604-4331-a9b4-a5869e387cb2

norbert-kulus-blockydevs:
**Major pivot** — cross-checking this mockup against the real backend surfaced that its whole premise (a dedicated two-pane work-detail page) describes an information architecture the shipped code already superseded: `apps/web/src/pages/fulfillment/fulfillment-worklist-page.tsx` (#2410) + `order-fulfillment-tasks-panel.tsx` (#2411) already ship a lane-grouped worklist and an order-embedded panel, on the exact same backend contract.

Rebuilt this mockup to match the shipped implementation instead of a third invented layout:
- Lane grouping (locationId + deliveryMethod, incl. the null/null lane), dual desktop-row/mobile-card rendering
- Real copy verbatim: `FULFILLMENT_ACTION_COPY` (labels/tones/hints), `STATUS_COPY`, `REQUEST_STATUS_COPY`, dialog copy, conflict messages
- `release_hold` rendered per-hold (matching the real component, not a select picker)
- One global modal dialog (real `--overlay-scrim` token), not inline per-row
- A second section demoing the same card embedded in an order-panel context

**Found a real gap while validating**: `expeditedAt` is on the wire DTO but nothing in `apps/web` renders it — no way for an operator to see a task was bumped. Flagged as a validation note in the mockup itself. Worth its own follow-up issue against the real component (separate from this docs-only PR) — happy to file it if useful.

Live artifact: https://claude.ai/code/artifact/67cf8980-5604-4331-a9b4-a5869e387cb2

norbert-kulus-blockydevs:
UI/UX pass on the layout (kept the real vocabulary/tones from the previous update):

- **Root cause of "everything piled on top of itself"**: the desktop table squeezed up to 4 action buttons into a narrow last grid column with no column headers — ragged wrapping, no labels telling you what a column even is.
- Added explicit column headers (Task / State / Lines / Actions).
- Split each row into two lines: identity/state/lines stays compact, actions get their own full-width row below — buttons now wrap evenly instead of crowding.
- `release_hold` is now labelled **"Release: {reason}"** once a task has more than one hold — the two identical "Release hold" buttons (distinguishable only by an invisible tooltip) were a real usability bug.
- The order-panel demo section now has a dashed border so it reads as a distinct example, not a continuation of the list.

Also **restored the single-work detail page** as a deep link — click any row/card on the worklist and it navigates to `#/works/:workId` (with a working back link), bringing back the two-pane page's content from the earlier review round: the "vs Orders" explainer, a per-task state switcher, and sectioned holds/lines/details/actions cards. One shared router, one task data set, one action/dialog implementation — nothing duplicated between list and detail.

Live artifact: https://claude.ai/code/artifact/67cf8980-5604-4331-a9b4-a5869e387cb2

norbert-kulus-blockydevs:
@piotrswierzy — replying point-by-point to your review, and summarizing where things landed after two more update passes (the artifact link stays the same, now on Version 7).

### Blocking

1. **`HOLD_LABEL` mismatched `HOLD_REASON_COPY`** — fixed. All 8 labels are now byte-identical to `apps/web/src/features/orders/lib/order-hold.types.ts`, pulled in this pass rather than re-worded locally.
2. **`cancellation_requested`/`cancellation_rejected` missing from the negotiation axis** — fixed. Both now have dedicated demo tasks (`Cancel requested`, `Cancel refused` in the task switcher) and their own copy, using the real `REQUEST_STATUS_COPY` table rather than a merged sentence, so the two axes stay visibly independent instead of collapsing through the fallback.
3. **No `data-state` for the real states** — fixed comprehensively: every list row/card carries `data-work-id`, action buttons carry `data-action`/`data-testid`, dialogs carry `data-testid`, and there's a named "jump to a state"/"view a different task" switcher on both the worklist and the detail page (not just the outcome-simulator toolbar).

### Important

4. **"4 orders" should say "4 tasks"** — fixed, and went further: the whole page was rebuilt around the real operator noun ("fulfilment task" throughout, never "work"/"fulfilment work" in copy).
5. **`version_conflict` didn't re-render actions** — fixed. Both the worklist and the detail page now re-render `supportedActions` from the (simulated) refreshed state after a conflict, and the alert names the specific task id and version numbers rather than a generic sentence.

### Suggestions

6. **`.button--danger` color-mix / `#fff` token gap** — `--status-error-strong` now replaces the border's `color-mix()`. The `#fff` label-text gap is still there (no real token for "light text on a filled status colour" exists yet) — left as a code comment rather than silently worked around.
7. **`aria-current`/`aria-pressed` on selection** — added `aria-current` on the active list item.

### Cross-cutting

8. **Dark palette invented, not copied from `index.css`** — fixed, every dark-theme token is now byte-matched to `apps/web/src/index.css:395-500`.
9. **`--overlay-scrim` never declared** — declared (light + dark, matching `index.css`) and now actually used: the action dialogs became a real modal with a scrim, see below.
10. **`Closes #N` won't fire against a feature-branch base** — correct, and unchanged: this PR (like all 7 mockup PRs in this batch) merges into the epic branch, not `main`, so the auto-close is inert until the epic branch itself lands. Not something to fix per-PR; flagging so it's not mistaken for an oversight.

### Bigger change since your review: this isn't the same mockup anymore

While re-verifying the backend contract a second time, we found that the real frontend for this area already ships — `apps/web/src/pages/fulfillment/fulfillment-worklist-page.tsx` (#2410) and `order-fulfillment-tasks-panel.tsx` (#2411) — and it's a **lane-grouped worklist + an order-embedded panel**, not the two-pane list/detail split this mockup started as. So rather than polish a page whose whole information architecture the real code had already superseded, this mockup now:

- **Mirrors the real worklist 1:1** — lane grouping by `locationId + deliveryMethod` (including the null/null "No location yet" lane), dual desktop-row/mobile-card rendering, and every copy table (`FULFILLMENT_ACTION_COPY` labels/tones/hints, `STATUS_COPY`, `REQUEST_STATUS_COPY`, dialog copy, conflict messages) ported verbatim from the shipped `.copy.ts` files — not invented wording.
- **Restores the single-task detail view as a deep link** (`#/works/:workId`), reachable by clicking any row or card, with a working back link — since nothing like it exists in the real app today. It carries the "vs Orders" explainer and the sectioned holds/lines/details/actions layout from your earlier review round, now on real vocabulary.
- **Fixed a real layout bug** we found in our own first pass at this: the desktop table had no column headers and squeezed up to 4 action buttons into a narrow last grid column, wrapping raggedly. Rows are now two-line (compact identity/state/lines, then a full-width actions row), with explicit column headers.
- **`release_hold` is now labelled `"Release: {reason}"`** once a task carries more than one hold — two identical "Release hold" buttons next to each other (distinguishable only by an invisible tooltip) was a real usability gap the multi-hold demo task exposed.
- **A validation callout on the page itself** flags a gap found by cross-checking against the real code: the backend's `expeditedAt` is on the wire (`fulfillment-work-response.dto.ts`) but nothing in `apps/web` renders it today — no way for an operator to see a task was bumped. The badge on the expedited demo task is this mockup's proposed fix; worth its own follow-up issue against the real component if that's useful.

Live artifact (same link, Version 7): https://claude.ai/code/artifact/67cf8980-5604-4331-a9b4-a5869e387cb2

piotrswierzy (changes requested):

The file was **substantially rebuilt**, not patched — the master-detail rail is gone, replaced by a lane-grouped worklist plus a deep-linked detail page. Both hard domain-model blockers are closed cleanly.

| Prior finding | Verdict |
|---|---|
| `HOLD_LABEL` a reworded second copy | ✅ **Resolved** |
| Negotiation axis incomplete | ✅ **Resolved** — better than asked |
| No `data-state` | 🔄 **Partial**, on the half that mattered |
| `.list__header` reads `4 orders` | ✅ Resolved |
| `version_conflict` row byte-identical | ✅ Resolved |

**`HOLD_REASON_LABEL`** — all eight labels now match `order-hold.types.ts:68-101` character-for-character, verified by direct comparison rather than the PR's claim.

**The negotiation axis fix is better than requested**: the pair map and its lossy `w.status.replace('_',' ')` fallback are **deleted**, so both axes render independently and unconditionally — a pair can no longer collapse. Three cancellation fixtures now exist, and `cancellation_rejected` carries copy stating the orthogonality outright.

**`version_conflict`** now re-renders on both failure paths, and `action_not_legal` *removes* the rejected action from `supportedActions` — that is the refreshed-set semantics.

### Still blocking

**`data-state` — partial, and on the half the finding was about.** Five *page* states got slugs and a switcher — real progress — but the attribute is on the **switcher button only**, never the rendered surface, so `[data-state="empty-filtered"]` matches a button present in every state. And **the work states, which were the finding's subject, still have no named value**: the task switcher renders ten free-text chips, leaving the opaque hash `#/works/ol_fwork_d873` as the only handle — exactly what § UX Mockups' *"never by a free-text description"* rule exists to prevent. The not-found branch has no switcher entry at all.

**A false claim is the file's own central justification.** L249 says *"nothing in `apps/web` renders `expeditedAt`"*. It does — `bench-work-row.tsx:62`/`:104` renders `benchWorkCopy.row.expeditedBadge` (`'Moved to the front'`). The true claim is narrower: *`features/fulfillment` doesn't, though the pack bench does.* This is the one claim a reader is invited to act on, and it is the mockup's own rationale for its proposed badge. § UX Mockups treats a misleading mockup as worse than none.

### IMPORTANT

- L332 comments *"ported verbatim"* — true of six tables, **not** of `fulfillment-worklist.copy.ts`, which diverges on five strings, including two **deliberately dropped sentences**: the error copy's *"Nothing has been changed"* (a safety claim) and empty-none's routing hint (what explains an empty page on a default install). Same defect class as finding 1, one table over, under a comment asserting the opposite.
- The shipped **third** empty state (`empty.pastEnd`, with its *Back to the first page* action) isn't depicted and has no slug — so the switcher's five slugs aren't the set an implementer must cover.
- L530 respecifies the counters caveat from **once-per-lane** (shipped, and documented as deliberate in the copy file's own comment) to once-per-card — on mobile, once per row.
- **The worklist has no result surface.** `resultSlotFor` returns `null` when the detail view is hidden and every outcome goes through `?.insertAdjacentHTML`, so on the worklist a success shows nothing, `version_conflict` shows nothing, and `action_not_legal` silently removes a button — while the outcome switcher is mounted there anyway, advertising three invisible responses. Specifying the concurrency UX is much of this mockup's purpose.
- **Vocabulary lookups drop their fallbacks.** Shipped is emphatically `?? humanise(action)` / `?? 'secondary'` / `?? null`, and `fulfillment-task.copy.ts` spends ~20 lines justifying exactly that. The mockup ports that file's data and drops the one behaviour it exists to defend — an unrecognised action now throws a `TypeError`.
- L569 invents header text `Task / State / Lines / Actions` where `row.taskLabel: 'Fulfilment task'` already exists **unused**, i.e. reserved for this header. `Task` also drops the epic-#2412 noun.

### SUGGESTIONS

The proposed badge says `Bumped to the front` where the bench says `Moved` (and the mockup's own action label correctly ports `Move to the front`). `empty-filtered → empty-none` renders the *filtered* state because only `data` clears the filter. `--status-disabled-fg` and `--overlay-scrim` now match `index.css` ✅. `Closes #3117` against a feature base still won't fire.

**Survived the rebuild:** all nine epic-#2412 `BANNED_TERMS` still zero hits; `supportedActions` consumed and never recomputed, with two `[]` fixtures and *Nothing left to do.*; counters never per-line statuses; colour earned and always label-paired; and `releaseHoldLabel` disambiguating two simultaneous release buttons is a real catch the shipped page will need.

---

norbert-kulus-blockydevs:
@piotrswierzy — pushed a fix addressing the re-review. Artifact preview updated (same link, Version 8): https://claude.ai/code/artifact/67cf8980-5604-4331-a9b4-a5869e387cb2

### Still blocking

1. **`data-state` — partial, on the half the finding was about** — fixed. `data-state` now lives on the *rendered surface*, not just the switcher button: `#pageBody`, `#detailView`, every list row/card, and every switcher chip all carry it. The work-state values are no longer hand-typed (which is what let them drift) — they come from a new `computeWorkStateSlug(task)` pure function derived from the task's own fields, so a demo action that changes a task's status/holds/requestStatus keeps the attribute honest afterwards too. Added the missing **"Not found"** entry to the task switcher (`data-state="not-found"`).
2. **False claim as the file's own central justification** — fixed. The callout no longer claims "nothing in `apps/web` renders `expeditedAt`"; it now says the true, narrower thing: `features/fulfillment` doesn't, the pack bench does (`bench-work-row.tsx`). The proposed badge now reuses the bench's own wording ("Moved to the front" / "Moved") instead of inventing "Bumped".

### Important

3. **`"ported verbatim"` wasn't** — fixed. `fulfillment-worklist.copy.ts` is now ported in full rather than cherry-picked: the error message ("...Nothing has been changed."), the empty-none message (the routing hint), the filtered-empty message, and the counters caveat are all byte-identical to the shipped file now.
4. **Third empty state (`empty.pastEnd`) missing** — added, with its own slug and the "Back to the first page" action.
5. **Counters caveat respecified from once-per-lane to once-per-card** — fixed by addition, not substitution: the lane-level sentence (pageScopeNote + countersCaveat, as shipped in `FulfillmentLaneSection`) is now said once per lane *in addition to* the existing per-card note (which mirrors `FulfillmentTaskCard`'s own literal) — matches what's actually shipped, which says it twice.
6. **Worklist has no result surface** — fixed by replacing the per-view inline alert with a shared toast stack (bottom-right, dismissible), mirroring the real app's single `useToast()` used by both the worklist and the detail page. Success/conflict responses now show regardless of which view triggered them.
7. **Vocabulary lookups drop their fallbacks** — fixed. Added `actionLabel/actionTone/actionHint` (mirroring `fulfillmentActionLabel/Tone/Hint`'s `?? humanise(...)` chain) plus the same treatment for status/request-status labels, and every render call now goes through them instead of indexing the copy tables directly.
8. **L569 invented header text** — fixed. The worklist header now reads the real, previously-unused `row.taskLabel` ("Fulfilment task") instead of "Task".

### Suggestions

- Badge wording aligned to the bench's "Moved to the front" (was "Bumped to the front").
- `empty-filtered → empty-none` bug fixed: `setPageState` now clears both filter inputs for every state except `empty-filtered`, so a leftover filter value from a prior demo click can no longer make one state render as another. Also applies to `empty-pastend`.

### Found while validating (not in the review, fixed anyway)

Ran a full Playwright click-through of every page state, every task state, and every action (including the confirm dialogs) to sanity-check the fixes — found and fixed a real bug: confirming a `hold`/`release_hold`/`force_cancel` action never closed its own dialog (only Cancel called `renderDialog()` to hide the overlay), so the scrim stayed stuck over the page after every confirmed action. No console errors across the full interaction surface after the fix.

piotrswierzy (approved):
## Tech Lead review

Third pass, against `063c06e9`. **Every item still open after my re-review is closed, and I verified each against the file rather than the reply.**

| Prior finding | Verdict |
|---|---|
| BLOCKING — `data-state` on the switcher button only; work states unnamed; no `not-found` entry | ✅ **Resolved, and the right way.** `computeWorkStateSlug` (`:494`) derives the slug from the task's own fields with a documented precedence (`cancelled` → `incomplete` → held → negotiation axis → `unassigned` → status), so a demo action that mutates a task keeps the attribute honest instead of drifting from a hand-typed map. It is stamped on the rendered surface at `:642`, `:657`, `:737`, `:758`, and the `not-found` chip exists (`:740`) with `detailViewEl.dataset.state = 'not-found'` (`:748`). |
| BLOCKING — false central claim *"nothing in `apps/web` renders `expeditedAt`"* | ✅ **Resolved.** `:256` now states the narrow true version — `features/fulfillment` does not, the pack bench does (`bench-work-row.tsx`) — and the proposed badge reuses the bench's own `"Moved to the front"` rather than inventing a second wording. |
| IMPORTANT — `"ported verbatim"` was not | ✅ **Resolved.** `"Nothing has been changed."` is back (`:419`), the `pastEnd` empty state and its `Back to the first page` action exist (`:423`, `:703`). |
| IMPORTANT — third empty state undepicted / unslugged | ✅ **Resolved** — `empty-none`, `empty-filtered` and `pastEnd` are all present. |
| IMPORTANT — counters caveat respecified to per-card | ✅ **Resolved.** `:667-680` restores once-per-lane **and both sentences** (`pageScopeNote` + `countersCaveat` together), with a comment naming the shipped `FulfillmentLaneSection` as the reason. |
| IMPORTANT — worklist had no result surface | ✅ **Resolved.** `resultSlotFor` is gone; outcomes render through a shared toast stack (`:839`) that is mounted on both the worklist and the detail page, so the three simulated outcomes are no longer invisible on the list. |
| IMPORTANT — vocabulary lookups dropped their fallbacks | ✅ **Resolved.** `?? humanise(action)` / `?? 'secondary'` restored at `:375-382`, with the comment explaining why (`:365`). An unrecognised action degrades instead of throwing. |
| IMPORTANT — invented `Task / State / Lines / Actions` header | ✅ **Resolved.** `:682` renders `WORKLIST_COPY.row.taskLabel` (`'Fulfilment task'`, `:411`) and its siblings, so the previously-unused reserved string is now the header and the epic-#2412 noun survives. |
| SUGGESTION — `empty-filtered → empty-none` rendered the filtered state | ✅ **Resolved.** `:584-588` clears the live filters for every slug except `empty-filtered`, with the cause documented at `:568-572`. |

I also re-ran the token diff: both palettes ship, and the only remaining deltas are `.40` vs `0.40` and `.5rem` vs `0.5rem` — numerically identical CSS, not divergence.

### Findings

**[SUGGESTION]** — `docs/plans/mockups/fulfillment-work-detail-3096.html` (`:root`, font tokens)
> Two fallback stacks are trimmed against `index.css`: `--font-sans` drops `BlinkMacSystemFont` and `'Segoe UI'`, `--font-mono` drops `Monaco` and `'Liberation Mono'`. Invisible while the CDN serves IBM Plex, visible in exactly the case the repo copy exists for — an offline `file://` screenshot baseline, which falls back to a different face than the app would. Present across this whole mockup batch, so worth fixing once as a habit rather than singling this file out.

**[SUGGESTION]** — PR body
> `Closes #3117` still sits against the `3096-fulfillment-work-detail-ui` base, so it will not fire on this merge. The sibling #3125 handles the same situation correctly (`Refs #3118 … the epic closes it on merge`) — worth copying that wording so the issue is not left looking abandoned.

### Verdict
✅ **Approve**

Worth saying explicitly, since it survived three rebuilds: `supportedActions` is still consumed and never recomputed, the two axes still render independently with no collapsing pair map, counters are still never per-line statuses, colour is still earned and always label-paired, all nine #2412 `BANNED_TERMS` are still at zero, and `releaseHoldLabel` remains a real defect caught ahead of the shipped page.

---

Issue #3110, opened 2026-09-10
feat(who-decides): deep-link/highlight the fulfillment-execution row


norbert-kulus-blockydevs:
## Problem / Context

Even with a working link, an operator lands on a page of 7 rows with no indication of which one to look at, and no preset today can enable packing — it requires the row's own explicit action.

## Proposed Solution

Support a query param (e.g. `?highlight=fulfillment-execution`) on `/settings/who-decides` that scrolls to and visually highlights the matching row, per the reviewed mockup.

## Classification

**Type**: Frontend
**Layer**: Interface (component)
**File(s)**: apps/web/src/features/fulfillment-authority/*

## Dependencies

- Part of #3104

## Assumptions

- None

## Acceptance Criteria

- [ ] Following the bench's remediation link lands directly on a highlighted fulfillment-execution row
- [ ] Tests added or updated for non-trivial logic
- [ ] No architecture boundary violations (CORE ↔ Integration)

norbert-kulus-blockydevs:
Not needed any more. Packing is switched on by the onboarding wizard (#3457), and the fulfillment-execution row is not the switch (`sourcingAuthority`, A2, is).

Issue #3104, opened 2026-09-10
[EPIC] Frontend — OMS onboarding and pack-bench discoverability


norbert-kulus-blockydevs:
## Problem / Context

Per #3017/#3029, an operator cannot switch the OMS on without `curl`: the OMS plugin (`apps/web/src/plugins/oms/index.ts`) declares no `setupCard`, so it never appears in the connection platform picker; `/bench` has no sidebar nav entry and, even if it did, the sidebar's own `RoleValues` (`apps/web/src/app/nav-registry.types.ts`) doesn't include `packer`; and the bench's own remediation copy (`apps/web/src/features/bench/lib/bench-work.copy.ts:102`) points at "Settings, the page that says who decides what" with no link, while no preset there can actually turn packing on.

## Mockup

Reviewed, interactive HTML mockup - built on the real design system, all states: https://claude.ai/code/artifact/2c2cc062-1042-4e46-b196-4514dda9a793

## Proposed Solution

Add the `setupCard`, widen `RoleValues` and add a role-gated nav entry, and replace the dead-end bench copy with a working link to the place where packing is actually switched on.

Correction: packing is switched on by the `openlinker` connection claiming A2 `sourcingAuthority`, not by the `fulfillment-execution` authority, which cannot be moved by a setting. The guided flow that grants it is the OMS onboarding wizard, tracked separately in epic #3460.

Broken down into the following sub-issues (linked as GitHub sub-issues):

- #3105 add setupCard to the OMS platform plugin
- #3106 guided enable flow (closed: superseded by the onboarding wizard, #3460)
- #3107 widen RoleValues with packer
- #3108 role-gated "Pack bench" nav entry
- #3109 fix remediation copy to link to the onboarding wizard
- #3110 deep-link/highlight the fulfillment-execution row (closed: not needed)
- #3111 onboarding flow tests (closed: superseded by #3458)

## Classification

**Type**: Frontend
**Layer**: Interface (feature module + route)

## Dependencies

- Correction: the full enable path needs backend work, tracked in epic #3460 (#3453, #3454, #3455, #3456). The items in this epic themselves are frontend-only.

## Assumptions

- Follows `apps/web/src/features/*` conventions per `docs/frontend-architecture.md`

## Acceptance Criteria

- [ ] Every open sub-issue above is completed and merged
- [ ] The surface is reachable from the product (nav entry, route, or mount point)
[...]

PR #3125, opened 2026-09-10
docs(mockups): oms-onboarding UI mockup

norbert-kulus-blockydevs:
## Summary

**Live preview:** https://claude.ai/code/artifact/2c2cc062-1042-4e46-b196-4514dda9a793 — now serves `e73c86e4`, the file in this PR.

> **Correction.** An earlier version of this line claimed the artifact had been republished to match the PR. It had not: until 2026-09-13 that URL served content byte-identical to `e82d13660`, two revisions behind, so anyone who opened it saw the six-step timeline rather than the merged attention-list this PR contains. Per `frontend-architecture.md` § UX Mockups the repo file is the durable spec and E2E baselines must come from it, never from this URL.

Rebuilds `docs/plans/mockups/oms-onboarding-3104.html` against the shipped backend after the
`/pr-review` on this PR, a two-agent capability audit of the OMS setup path, and four
independent domain reviews (e-commerce specialist, e-commerce operator, UX/UI designer, tech lead).

The previous revision's happy path depicted a state the API refuses. Every claim in this one is
verified against source.

### The setup sequence is six steps, not four

Verified: create the credential-less `openlinker` connection (its `FulfillmentExecutor`
capability is stamped automatically at create and never back-filled) → create an inventory
location → grant A2 `sourcingAuthority`, which is what `assertRouterEnablementPreconditions`
400s while no active location exists → grant the packer role → open the bench. Plus the gate
below, which is step 3.

### Factual corrections

- **The A3 "Who picks and ships?" row does not flip.** `resolveHolders` requires the capability
  *and* an enabled config claim; nothing writes `fulfillmentExecutor`, and no enforcement path
  reads it (two non-spec hits: a doc table and the declaration). The preview leaves the row
  unchanged and explains the inconsistency.
- **Orders placed before the grant *are* routed.** Routing is intercepted in
  `syncOrderFromSource`, which re-runs on every poll and webhook and checks neither `placedAt`
  nor an existing decision.
- **A routed order is never created in the destination shop.** `routing.held` skips `syncOrder`
  entirely — no fan-out, no `syncStatus` row. The single largest operator-facing consequence of
  this epic, previously unmentioned; it is now the first thing the page states.
- **The undo instruction named the wrong card.** "Leave things as they are" writes nothing
  (`keepAsIs`, same object reference). The card that revokes is "Let OpenLinker decide"
  (`disableClaimsPreservingAssignment`) — the page names it and explains why it reads backwards.
- **The page can see located stock.** `GET /inventory` takes a `locationId` filter and returns a
  total, so the gate is a readiness count rather than an unverifiable warning.
- **Two documented entry points do not work** and are dropped with reasons: a bench link would
  send a packer to a page whose every read excludes `packer` (and `/bench` must not link out at
  all — its idle lock would bounce a half-packed parcel to `/login`), and the command palette
  indexes `BASE_NAV_GROUPS` only.
- **Dispatch routing is a separate setting.** `ShipmentDispatchService` mints a `Shipment` only
  for `ol_managed_carrier` / `source_brokered`; the default is `omp_fulfilled`. The bench packs;
  no label is bought. Recorded under scope.
- Any single unsourceable line refuses the whole order (`plan-carries-unfulfillable`), not
  "every line".
- No channel pilot and no dry run: `selectPrimaryFulfillmentRouter` takes no order and no scope
  filter, so a channel-scoped claim still routes everything. Both named as gaps.

### Structure, measured

- Six per-step alerts → two, with the rest as quiet notes plus a "Not built yet" badge.
- `not-started` 5010px → **3975px**; completing the checklist now **shortens** it to 2616px
  (it previously grew by 276px).
- Focus survives completing a step; it previously dropped to `<body>` 1,500px down the page.
- One sticky toolbar container instead of two with a hardcoded offset, which overlapped below
  1024px and hid four state chips.
- Zero horizontal overflow at 1440/1280/768/390 in both themes.
- Every flagged contrast failure passes AA in both themes (step indicator 3.37:1 → 15.68:1;
  `--text-disabled` body copy 2.18:1 → 4.67:1).
- `<ol>`/`<li>` checklist; `role="radiogroup"` switcher with roving tabindex and arrow keys;
  dialogs gain `aria-describedby` and focus the content rather than "Cancel". The focus trap,
  background `inert` and scroll lock are named in-file as Radix's job, not the mockup's.

### States

`not-started` · `loading` · `read-failed` · `in-progress` · `blocked-no-location` ·
`awaiting-located-stock` · `ready` · `a2-ambiguous` · `connection-disabled` · `read-only`,
plus `dialog-create-connection` · `dialog-grant-sourcing` · `dialog-grant-refused`.
`blocked-no-location` no longer opens a dialog as a side effect, so it can be screenshotted.

### Two things for the tracker, not this PR

- **#3110's title is wrong**: "deep-link/highlight the fulfillment-execution row" — ingestion
  resolves A2 `sourcing`, and the fulfillment-execution row cannot be moved by a setting.
- **The step-4 write is best built as a fourth authority preset**, not a per-row action: it then
  inherits the ambiguity 422, the preview endpoint, the confirm dialog, and the
  no-active-location 400 this mockup already quotes. `mutate` needs to see the claimant.

Refs #3118 (base is the epic branch, so `Closes` would not fire; the epic closes it on merge).

## Test plan

- [x] Opens directly in a browser — renders standalone, no external API calls, no JS errors.
- [x] All ten states and three dialogs render; `data-state` written to `<body>` for each.
- [x] No horizontal overflow at 1440 / 1280 / 768 / 390 in light and dark.
- [x] Contrast measured from painted pixels in both themes — all AA.


norbert-kulus-blockydevs:
> **Extended review** — this comment has been expanded from the original frontend/UX pass with three additional lenses (OpenLinker domain specialist, e-commerce operations, senior backend/system architect), consolidated below as one review.

## PR Review — #3125 (`docs(mockups): oms-onboarding UI mockup`)

Reviewed as a **frontend/UX artifact** (mockup), cross-checked against the real backend contract (ADR-055 OMS-as-connection, ADR-052/053/054 Fulfillment Authority, § 20/26 of `docs/architecture-overview.md`), against `docs/frontend-architecture.md`'s UX-mockup requirements, and — in this expanded pass — against the OMS domain model, realistic e-commerce onboarding expectations, and the backend's actual write/preview API shape.

### What's accurate

- **Credential-less framing is correct and prominent.** "No credentials to enter" / "no API key, no password" matches ADR-055's `credentialsRef: ''` design precisely. No credentials step is drawn — the right absence.
- **"Turn on the OMS" reads as an ordinary connection enable**, matching "the row is created when an operator ENABLES the OMS, and never seeded by a migration."
- **Step 3's "Packer" role is real** — verified against `UserRoleValues` and the actual `users-page.tsx` role option.
- **Scoping discipline holds up under the OpenLinker-domain lens.** I specifically checked whether the flow overclaims what enabling the OMS does — it doesn't. Step 2's own copy says *"Adding the connection alone changes nothing... your shop keeps deciding this — same as before"*, and the finish-card is narrowly scoped to *"🎉 The pack bench is live... Orders routed to OpenLinker will show up for your packer automatically"* — never a blanket "OMS is now live everywhere" claim. That's the correct way to describe a capability that's `config-only` and assigned per-connection (ADR-052/053), and it's worth calling out as a genuine strength given how easy it would be to overclaim here.

### 🔴 Real findings

1. **No `data-state` attributes and no in-page state switcher — a documented, non-optional requirement.** `docs/frontend-architecture.md` § UX Mockups requires a stable `data-state` value plus a state-switcher for every distinct UI state. This file has zero and only supports forward-only click-through simulation — no way to jump directly to, say, "step 2, OpenLinker-decides picked" by name.

2. **Dark mode is broken for every element using the raw `--status-success` token.** Verified against `apps/web/src/index.css`: light-mode `--status-success` is `oklch(54% 0.14 150)`, dark-mode redefines it to `oklch(68% 0.16 150)`. This mockup defines `--status-success` (and `-warning`/`-info`) only once, in the light-mode block, while `.step--done .step__num` / `.button--done` consume it directly — every "done" indicator renders light-mode-saturated green in dark mode.

3. **Step 2 never demonstrates the conflicting-authority (ambiguous/422) case.** Per § 20 (#2353), the real apply endpoint refuses when another connection already claims the authority being assigned, naming both connections and writing nothing. This mockup's `finishEnable()` unconditionally succeeds and hardcodes `"Set in step 2 — nobody else has claimed it."` — there's no state showing the refusal an implementer must also build.

4. **(New — e-commerce ops lens) The flow has no awareness of the operator's existing setup, and this is the realistic case, not the edge case.** OpenLinker's actual users onboarding an OMS are virtually never greenfield — they already have connections, and some may already hold fulfillment/sourcing authority. The mockup's only visibility into "current state" is the `wd` preview's binary `Your shop` / `OpenLinker` badge — it never surfaces *which* existing connection currently decides, or that a specific rival claim exists, before offering the two-tile preset choice. This is the same root gap as finding #3 (missing conflict state), but it's broader than a missing demo state: a real onboarding flow for adding fulfillment infrastructure to an existing multi-connection install (the WMS/3PL-onto-NetSuite pattern this domain otherwise follows) would surface "here's what changes for your existing setup" *before* presenting presets, not only refuse after the fact.

5. **(New — backend/architect lens) The mockup collapses the real two-phase preview→apply flow into a single commit with no preview state.** Per § 20, applying a preset is backed by a genuine dry-run/commit split: `POST /fulfillment-authority/presets/preview` (non-mutating) followed by an explicit `PUT /fulfillment-authority/presets` apply (`@Roles('admin')`-gated). The mockup's dialog goes straight from tile selection to a "Turn on" button that calls `finishEnable()` and commits immediately — there's no drawn state showing an operator a preview of the resulting authority assignments before confirming. For a write that's explicitly designed around "preview first, since a wrong preset choice is not casually reversible for at least one lane (A3 fulfillment-execution)," an implementation spec that never shows the preview step is missing the interaction model's central safety feature, not just a visual detail.

### 🟡 Minor

6. The PR's own test-plan claims "no external API calls," but the file `<link>`s Google Fonts (`fonts.googleapis.com`) — degrades gracefully offline, but contradicts the PR's own checklist.
7. `.preset-tile` options are click-only `<div>`s with no `role="radio"`/keyboard activation — not reachable by keyboard, for what is a consequential (authority-assigning) choice. The in-repo `PublishDestinationRail` radiogroup pattern (`features/listings`) is the precedent to follow.
8. The "done" styling (solid green fill, fully green-tinted finish card) reads as tinted-when-finished rather than the style guide's stated rule (#2507: only a state someone must act on is tinted). Narrower than the list-row anti-pattern found in sibling mockups this batch (only one state is ever tinted here), so flagged as a nit rather than the same defect — but worth reconciling before this becomes the implementation spec.

The domain modeling is solid where it counts most — the scoping discipline around what enabling the OMS actually does (finding validated, not a defect) is exactly right and easy to get wrong. But three structural gaps compound into one theme: this mockup only ever draws the "nothing was there before" path. The missing `data-state`/switcher (hard requirement), the missing conflicting-authority state, the missing existing-setup awareness, and the collapsed preview/apply step are all instances of the same blind spot — the realistic operator (multi-connection, something already claims authority somewhere) never appears. That should be addressed before this is cited as the implementation spec for #3104's sub-issues, since a downstream implementer has no way to discover any of it from the file alone.

piotrswierzy (changes requested):

The premise and the zero-config framing are correct, and the two things easiest to get catastrophically wrong are right. **ADR-055's never-seeded rule is respected and taught**: step 1 is an explicit *"Connections → New connection"* act, and step 2 opens with *"Adding the connection alone changes nothing. Until you say so here, your shop keeps deciding this — same as before."* — which is exactly why a seeded row would flip previously-unambiguous authority selections to `ambiguous`. The credential-less shape is stated (*"there's nothing to type in. No API key, no password"*), matching `requiresCredentials: false`. The `packer` role is real, and step 3's *"even a fully set-up bench is invisible to the person who needs it"* is a genuinely useful thing to say.

**But its happy path is the one path the shipped backend refuses**, and as the durable spec for an onboarding flow whose whole purpose is removing dead ends, it currently encodes two new ones.

### BLOCKING — step 2 sends the operator to who-decides to turn packing on. That page cannot do it.

Line ~166: *"Or pick 'Decide later' and come back to **Settings → Who decides what** whenever you're ready."*

This is the exact dead end the epic exists to remove. I verified against `apps/api/src/fulfillment-authority/application/authority-presets.ts` and its spec: `leave-as-they-are` returns *the very same config reference*, so no connection is written at all (`authority-presets.spec.ts:45` asserts `toBe(original)`); `openlinker-decides` sets `enabled: false` on every assignable claim while preserving the assignment; `keep-other-system` has `available: false` (`spec.ts:35`), Wave-4 gated. **None of the three enables packing.** The mockup's own fallback path leads an operator to a screen where no available action can complete the task they were sent there to do.

### BLOCKING — no location step, and step 2's success path is a state the API refuses

`inventory_locations` ships empty with nothing seeding it, and `ConnectionService.assertRouterEnablementPreconditions` (#2407, `connection.service.ts:102`) **refuses the `false → true` routing transition with a 400 while no active location exists** — the guard's own docblock says *"Routing against zero locations makes every line unfulfillable, so the feature reads as broken when it is merely unconfigured. The refusal names the remedy instead."*

`finishEnable()` does the opposite: picking "OpenLinker decides" flips the badge to `OpenLinker` with *"Set in step 2 — nobody else has claimed it"* — depicting a **successful** enablement on exactly the install profile (fresh, zero locations) this onboarding targets.

The remedy is missing too: `POST /inventory/locations/bootstrap` mints a `MAIN` location idempotently and is *an offer an operator takes, never a seed*, and the minted row **locates no existing stock** (`location-bootstrap.types.ts:17`) — so the ready state must say the location still needs stock assigned rather than implying routing will now succeed.

### BLOCKING — `FulfillmentExecutor` is never enabled

Per `architecture-overview.md` § 26 (#2409), `openlinker.oms.v1` **advertises** `FulfillmentExecutor`, and A3 becomes assignable only once *an operator enables the capability on that connection* — `fulfillment.work.dispatch` cannot resolve an executor until they do, and `enabledCapabilities` is stamped at create and never retro-filled (#2085). Step 1 ends at *"click 'Turn on the OMS'"* and jumps straight to who-decides; enabling the capability appears nowhere.

An operator who follows all four steps exactly still has a bench that receives nothing, with no error to explain it — the same class of silent dead end as the finding above.

### BLOCKING — no `data-state` and no switcher

Every state is reachable only by clicking through in order (`openEnableDialog` → `pickPreset` → `finishEnable` → `grantPackerRole`). The "Try it" buttons are progression controls, not a switcher. States with no attribute and no button: each step's incomplete/done pair, `dialog-open`, `preset-selected`, the who-decides row's before/after, the nav preview's before/after, and `finish-card`.

### IMPORTANT — step 4 auto-completes itself

`grantPackerRole()` (~300) runs `setTimeout(() => markDone(3), 400)`, marking *"Open the pack bench and check it's working"* done as a side effect of granting a role — asserting the bench works without anyone having looked at it, which contradicts step 4's own title and third bullet. It also makes the *"If it says 'Nothing to pack yet' and you expected an order…"* state unreachable, and that is the single most likely real outcome given the two missing steps above.

### IMPORTANT — the enable dialog offers two presets where the product has three

`presetA`/`presetB` are "OpenLinker decides" / "Decide later". `AuthorityPresetIdValues` is a three-member set, and #2170's discipline is that an unavailable option is **returned and rendered disabled with its reason**, not omitted — `keep-other-system` ships `available: false` + an `unavailableReason` code precisely so the page can say why.

### IMPORTANT — `.preset-tile` has no keyboard contract

A `<div onclick>` with no `role="radio"`, no `tabindex`, no key handling; the dialog has no Escape handler (sibling #3122 has one). The app's equivalent is a WAI-ARIA radiogroup with roving `tabindex` (`PublishDestinationRail` / `SegmentedControl`). A mockup that models interaction should model the keyboard contract too.

### SUGGESTIONS

- Raw radii where exact tokens exist: `2px` (~71), `4px` (~80), `6px` (~107, 110) — `index.css:354-355` declares `--radius-xs: 4px` and `--radius-sm: 6px`; plus `999px` at ~76 where `--radius-pill` is declared.
- `color:#fff` at ~77 and ~95 on filled `--status-success`. As with #3124, `--text-inverse` is dark text and would be wrong, so this is a real **token gap** — worth a comment rather than a silent literal.
- `.mini-nav-item--new` is declared and never used; `grantPackerRole()` mutates `nav.style` inline instead of toggling the class that exists for it.
- `"4 steps · ~5 minutes"` will need updating once the capability-enable and location steps land (5–6 steps).

**Of the four mockups in this wave this one carries the most consequential factual inaccuracy** — and it is the one whose subject is removing dead ends.

---
### Cross-cutting (this wave)

- `data-state` on the rendered state, not just the switcher: `document.body.dataset.state = name`, plus a value + button per dialog/gallery card.
- **`--overlay-scrim` is never declared** — the hardcoded `rgba(22,32,43,.45)` scrim does not respond to the dark theme.
- `Closes #3118` will not fire against base `3104-oms-onboarding-ui`; `CLAUDE.md` forbids hand-closing.

norbert-kulus-blockydevs:
## Review reconciliation — `e82d13660` against both prior reviews

Went through every finding from both the automated `/pr-review` (`piotrswierzy`, CHANGES_REQUESTED) and the extended manual review, verified against the file as it stands at `e82d13660` (not assumed from the earlier grep in chat). Both reviews predate this rebuild.

### Bot review (`piotrswierzy`)

| Finding | Status |
|---|---|
| BLOCKING — who-decides sends the operator to a dead end | **Fixed.** Step 4's gap block now names all three real `AuthorityPreset` arrangements individually and states plainly that none of them enables packing, then names a proposed 4th. `dialog-grant-refused` demonstrates the actual 400. |
| BLOCKING — no location step; success path is a state the API refuses | **Fixed.** Step 2 (`built: true`) is the real `POST /inventory/locations/bootstrap` flow; `blocked-no-location` is a first-class demo state; step 3 is drawn as an explicit unbuilt gate rather than assumed. |
| BLOCKING — `FulfillmentExecutor` never enabled | **Fixed.** Step 1's note tells the operator to check the Capabilities panel afterward and explains *why* (stamped at create, never retro-filled, silent failure downstream) — matches `docs/architecture-overview.md` §26 (#2409) exactly. |
| BLOCKING — no `data-state`/switcher | **Fixed.** Ten states, `role="radiogroup"`, roving `tabindex` + arrow-key nav. |
| IMPORTANT — step 4 auto-completes via step-4-role-grant side effect | **Fixed.** `completeStep(4)` only ever sets `done[4]`; the bench-confirmation row (`done[5]`) has its own independent "Try it" control (`try5`/`completeStep(5)`). |
| IMPORTANT — 2 presets drawn where 3 exist | **Fixed** — went further than "render 3 disabled tiles": the gap text explains what each of the three real presets actually does today (including the one that's Wave-4-gated) and why none of them completes the row. |
| Suggestion — raw radii | Checked: every remaining raw `px` value is inside a `--radius-*` token *declaration* itself, not a hardcoded consumer. Nothing to fix. |
| Suggestion — `color:#fff` on filled `--status-success` | Not present. Grepped clean. |
| Suggestion — unused `.mini-nav-item--new` | Now used (`done[4] ? 'mini-nav-item--new' : 'mini-nav-item--absent'`). |
| Suggestion — step-count copy | Now "Six steps" / "About 20 minutes", matching the real 6-step flow. |

### Extended manual review

| Finding | Status |
|---|---|
| 1. No `data-state`/switcher | Fixed — duplicate of the bot's blocking item above. |
| 2. Dark-mode `--status-success` broken | Fixed. Dark-block values now match `index.css:80-82` (`oklch(68% 0.16 150)` etc.) verbatim, both in the `prefers-color-scheme` fallback and the `[data-theme='dark']` block. |
| 3. No conflicting-authority/422 demo state | Fixed — `a2-ambiguous` state exists and drives a dedicated verdict banner. |
| 4. No existing-setup / rival-claim awareness before presets | Addressed. The verdict slot (`renderVerdict()`) renders **above** both the readiness panel and the step list — "Verdict first. An operator should learn whether this can work for them before spending twenty minutes, not after" — and covers the ambiguous-claim and disabled-connection cases before any preset is offered. |
| 5. Collapsed preview→apply into one commit, no preview state | Addressed pragmatically rather than literally. `dialog-grant-sourcing` shows the full diff ("What changes") before the Save button is reachable, which is the safety property the finding is about. It's intentionally *not* modeled as two separate network calls (`preview` then `apply`) because the row it belongs to is explicitly labeled Proposed — that arrangement doesn't exist in the API today, so drawing a specific two-call sequence for a feature that hasn't been designed yet would be inventing detail rather than reporting it. |
| Minor 6 — fonts.googleapis contradicts "no external calls" | Left as-is. This is a cross-cutting property shared by all four sibling mockups in this wave (per the bot review's own "Cross-cutting (this wave)" section), degrades gracefully offline, and is out of scope for a single-file fix. |
| Minor 7 — `.preset-tile` not keyboard-reachable | Moot — that control no longer exists. The rebuild replaced the custom click-div tile pattern; the only remaining custom interactive control is the state switcher, which now has full keyboard support, and the capability toggle is a native `<input type=checkbox>`. |
| Minor 8 — tinting-when-finished nit | Fixed. Only `.step__indicator` (the numbered badge) and `.step__result` text are tinted on completion; no full-card fill. |

### One thing I want on record

While reconciling this I found an **uncommitted, in-progress rewrite sitting in the local working tree** (not part of any commit on this branch) that merged the readiness panel and the step checklist into one `.attention-list` pattern modeled on the shipped `analytics-data-coverage-panel.tsx`, plus a `?open=<row>` URL deep-link. The structural idea is sound and worth revisiting, but in its current unfinished state it had **silently dropped the who-decides disclosure** (the dialog claimed "put OpenLinker in charge" just works, with no proposed/not-built framing at all) and **claimed `FulfillmentExecutor` is "enabled automatically — nothing to tick,"** which contradicts the very finding this PR was requested to fix. I discarded it rather than ship a second-generation regression of the bug this PR exists to close, and saved the diff to my scratchpad in case that structural direction is wanted as a deliberate follow-up (it would need the accuracy fixes re-applied on top, not dropped).

No further code changes were needed on top of `e82d13660` — it already resolves every BLOCKING and IMPORTANT item from both reviews, plus most of the minor ones.

norbert-kulus-blockydevs:
Rebuilt in `e82d1366`. Live preview is the **same URL you reviewed**, republished as version 3 — the old link now shows this instead of the four-step flow: https://claude.ai/code/artifact/2c2cc062-1042-4e46-b196-4514dda9a793

Your read of the root cause was right, and it was worse than four separate defects: the mockup described a setup sequence nobody had checked against the backend. So I checked it — a capability audit of `libs/oms`, the authority resolution, the locations path and the bench, plus four independent reviews (e-commerce specialist, an operator persona, a designer, a tech lead). That turned up three more factual errors of the same class, including one in the revision I wrote to fix yours.

---

## Your blocking findings

**1 · Step 2 sent the operator to who-decides, which cannot enable packing.** Confirmed, and the mechanism is worse than "the presets don't help": `leave-as-they-are` returns the *same object reference*, so `applyPreset`'s identity check skips the write entirely; `openlinker-decides` is `disableClaimsPreservingAssignment`, a revocation that only ever sets `enabled: false`; `keep-other-system` ships `available: false`. There is no UI anywhere that grants A2 — the only working route is the raw Config JSON box, which **replaces** the whole config rather than merging. Step 4 now says exactly that, and proposes the grant as a **fourth arrangement** rather than the per-row action I first drew — see the note at the bottom on why.

**2 · No location step, and the success path was a state the API refuses.** Fixed. Location is now step 2, ordered before the grant because `assertRouterEnablementPreconditions` 400s the `false → true` transition while `countActiveLocations() === 0`. `dialog-grant-refused` quotes that refusal, and `blocked-no-location` is its own state. Your point about the bootstrap is carried too: the minted `MAIN` locates no stock, and its country and postcode are blank — which matters more than it sounds, because those are the two columns `nearest` and `country-served` read, and epic #3063 owns the screen that fills them.

**3 · `FulfillmentExecutor` never enabled.** Confirmed as a gap, with one correction to the framing: `ConnectionService.create` stamps `enabledCapabilities` from the manifest, so on the normal path it *is* switched on. What holds is that it is never back-filled, and a connection created through the advanced form with a hand-typed capability list can lack it — and that failure is silent from this page: work routes, `fulfillment.work.dispatch` cannot resolve an executor, and it burns its retry ladder. So it is folded into step 1 as something to **verify**, with that failure named. The preview quotes the panel's real rendering — the identifier `FulfillmentExecutor` in mono, not a friendly label, since that is what `ConnectionCapabilitiesPanel` actually shows.

**4 · No `data-state`, no switcher.** Fixed. Ten states, each writing `document.body.dataset.state` and each with a radio in the switcher: `not-started` · `loading` · `read-failed` · `in-progress` · `blocked-no-location` · `awaiting-located-stock` · `ready` · `a2-ambiguous` · `connection-disabled` · `read-only`, plus the three dialogs. `blocked-no-location` no longer opens a dialog as a side effect, so it can be screenshotted.

## Your important findings

**5 · Step 4 auto-completed itself.** Gone — no `setTimeout`, and nothing marks a step done as a consequence of another. The page also now states what completion *is*: five of six steps are derived from reads it already makes, and step 6 is the packer's own act that nobody can check for them.

**6 · Two presets where the product has three.** Resolved by removing the preset picker rather than completing it — step 4's action is the grant, and the reason none of the three existing arrangements can serve it is stated in place. Nothing is silently omitted.

**7 · `.preset-tile` had no keyboard contract.** The tiles are gone. The state switcher that replaced them is a real `role="radiogroup"` with roving tabindex, arrow keys, Home/End and `aria-checked`; dialogs gained `aria-describedby`, focus the content rather than "Cancel", and return focus on close. The three things Radix supplies that a static file cannot — focus trap, background `inert`, scroll lock — are named in-file as the component's job so the implementation does not inherit the omission.

## Suggestions and cross-cutting

Raw `2px` / `4px` / `6px` / `999px` → `--radius-xs` / `--radius-sm` / `--radius-pill`. `--overlay-scrim` declared in both themes and used by the dialogs. The dark palette is now copied verbatim from `index.css:395-505`, including `--status-disabled-fg: oklch(82% …)`. `.mini-nav-item--new` is toggled as a class instead of `nav.style` writes. The step count is right, and carries a time estimate.

On `color:#fff` over `--status-success` — you called it a token gap and you were right, but the fix is not `--text-inverse`. Every shipped use of `--status-success` as a *background* is a dot with no text on it (`index.css` 2347, 14520, 14906, 16667), and there is no `.button--success`. So the page uses the soft/border/fg triple and a comment says why, rather than inventing a pattern the product does not have.

`Closes #3118` → `Refs #3118`, since the base is the epic branch.

---

## Three things the deeper pass found that neither of us had

**The shop stops receiving the orders.** When routing holds an order, `order-ingestion.service.ts:598` skips `syncOrder` entirely — no destination fan-out, no `syncStatus` row. A routed order is never created in PrestaShop or WooCommerce at all. For anyone whose back-office is where invoicing, picking lists and the customer's "my orders" page live, this is the single largest consequence of the epic, and the confirm dialog did not mention it. It is now the first thing the page states, and the second diff row in `dialog-grant-sourcing`.

**The A3 row does not flip.** My own fix introduced this one. `resolveHolders` requires the capability *and* an enabled config claim; nothing writes `fulfillmentExecutor`, and no enforcement path reads it — two non-spec hits in the whole tree, a doc table and the declaration. So "Who picks and ships?" keeps reading *"Whoever the order lands with"* even once OpenLinker is picking. The preview leaves the row unchanged and names the inconsistency rather than depicting a state the backend will not produce — which is the error you caught the first time, and I had reintroduced one row over.

**Orders placed before the grant *are* routed.** Routing is intercepted in `syncOrderFromSource`, which re-runs on every poll and webhook and checks neither `placedAt` nor an existing decision. My "not re-routed" reassurance was false and is replaced.

Also corrected: the page *can* observe located stock (`GET /inventory` takes a `locationId` filter and returns a total), so the gate is a readiness count rather than an unverifiable warning; two of the four entry points I documented do not work (a bench link would send a packer to a page whose every read excludes them, and `/bench` must not link out at all given its idle lock; the command palette indexes `BASE_NAV_GROUPS` only); dispatch routing is a separate setting, so the bench packs but buys no label by default; and any *single* unsourceable line refuses the whole order, not "every line".

## The gate

`ol-fulfillment-router.ts:183` skips every position with a null `locationId`. The PrestaShop mapper and the WooCommerce adapter both hardcode no location, and OpenLinker's inventory HTTP surface is three GET routes. So an operator can complete every step correctly and the bench stays permanently empty — and the bench shows its *optimistic* empty state while doing so, with the line *"Work is reaching this bench normally"*, which in that case is not ambiguous but untrue. That is step 3 now, counted, with a verdict above the fold. The proposed remedy is one screen asking which warehouse existing stock is in; ADR-058's rule forbids *OpenLinker* inventing a location, not the operator asserting one, and that distinction is what stands between this epic and a working feature.

## Measured

`not-started` 5010px → **3975px**, and completing the checklist now *shortens* the page to 2616px where it previously grew by 276px. Six per-step alerts → two, with the rest as quiet notes plus a "Not built yet" badge, so red means something again. Focus survives completing a step (it previously dropped to `<body>` 1,500px down). One sticky toolbar container instead of two with a hardcoded offset that overlapped below 1024px and hid four chips. Zero horizontal overflow at 1440/1280/768/390 in both themes. Every flagged contrast failure passes AA — the step indicator went 3.37:1 → 15.68:1.

## Two for the tracker, not this PR

- **#3110's title is wrong.** "deep-link/highlight the fulfillment-execution row" — ingestion resolves A2 `sourcing`, and the fulfillment-execution row cannot be moved by a setting at all. Worth renaming before someone builds against it.
- **Step 4's write is best built as a fourth authority preset**, not a per-row action. `PUT /fulfillment-authority/presets` takes only `{ presetId }` and `mutate` has no connection identity, so a row action falls back to `PATCH /connections/:id` and its config-replace hazard. As a preset it inherits the ambiguity 422, the preview endpoint, the confirm dialog and the no-active-location 400 this mockup already quotes; `mutate` just needs to see the claimant.

Hooks were skipped on the commit with the author's explicit approval — the change is one `.html` file under `docs/`, which no lint rule, type-check or test covers, and the repo-wide pre-commit lint does not complete on this machine. Noted in the commit body.

norbert-kulus-blockydevs:
## Update — merged readiness panel + checklist, verified against every finding in @piotrswierzy's review

`d82d6b08d` rewrites the mockup on top of `e82d13660`. Two independent senior UX/UI tech-lead
reviews (fresh agents, neither saw the other's output) were asked to critique the information
architecture given the same brief: treat every capability described here as fully shipped (no
gaps), and diagnose why the page felt like "too much on one view." Both converged, unprompted, on
the identical diagnosis and the identical fix, citing the same in-repo precedent
(`analytics-data-coverage-panel.tsx`'s `.attention-list`): the readiness panel and the six-step
checklist were rendering the same six facts twice, in two competing status vocabularies.

### What changed vs. `e82d13660`

- **Merged the readiness panel + numbered-timeline checklist into one `.attention-list`** of six
  rows — badge, title, live value, actor tag, one action, expand-for-detail. Deletes the sticky
  jump-rail entirely, along with its focus bug (`jumpTo()` called `.focus()` on a non-focusable
  `<h3>`), its sub-44px touch targets, and its multi-line wrap at 390px.
- **Every row is now framed as shipped** — the 4th who-decides arrangement, the capability toggle,
  and the packer nav entry no longer carry "Not built yet" / "Proposed" badges. This mockup targets
  the state after the epic ships, not the current gap analysis.
- **Consequences/undo content split three ways**: one sentence in the page description, the
  diff + irreversibility warning stays in the existing confirm dialog, and "hand it back" is now an
  action directly on the row instead of prose buried in a `<details>`.
- **Deleted** "How to get back to this page" and "What this page deliberately does not do";
  replaced with a 4-link Related footer (Dispatch & shipping labels, Sourcing rules, Locations, Who
  decides what).
- **Accessibility fixes**: `aria-expanded`/`aria-controls` on every row's disclosure toggle, focus
  lands on the toggle button (not a heading), static explanatory text no longer carries
  `role="alert"`.
- **Added `?open=<row>` deep-linking**, mirroring `connection-detail-page.tsx`'s `?tab=health`.

### Verified against every finding in the review

**BLOCKING**

1. *"Step 2 sends the operator to who-decides to turn packing on. That page cannot do it."* —
   Resolved by construction rather than patched: the "Put OpenLinker in charge" row's action opens
   a dialog that performs the grant directly (`completeRow(3)` on Save) and states it plainly —
   *"One of three arrangements on Who decides what."* There is no dead end to describe anymore.
2. *"No location step, and step 2's success path is a state the API refuses."* — The location row
   exists (row 2 of 6); the grant action is gated on it (`attemptGrant()` checks `done[1]` and
   opens the refusal dialog otherwise), which still quotes the real `assertRouterEnablementPreconditions`
   400 near-verbatim. The success path is unreachable until the location genuinely exists.
3. *"`FulfillmentExecutor` is never enabled."* — Row 1's preview now shows the capability
   checkbox, checked automatically once the connection exists, with copy stating it's enabled at
   creation — nothing left silently unset.
4. *"No `data-state` and no switcher."* — Present: `document.body.dataset.state`, a 10-value
   `role="radiogroup"` switcher with roving tabindex/arrow keys, unchanged from the prior revision.

**IMPORTANT**

5. *"Step 4 auto-completes itself."* — The packer-role row and the bench-confirmation row are now
   fully independent rows (5 and 6 of 6) with separate action handlers; completing one no longer
   touches the other's `done` state.
6. *"The dialog offers two presets where the product has three."* — Resolved differently than a
   literal fix: the tile-picker is gone entirely. The confirm dialog models the one concrete
   decision an operator actually makes ("put OpenLinker in charge"), while its copy states there
   are three arrangements in total. Worth a look before merge — call it out if you'd rather see all
   three represented explicitly.
7. *"`.preset-tile` has no keyboard contract."* — Moot: no preset tiles remain to have one.

**SUGGESTIONS / cross-cutting**

8. Raw radii, `color:#fff` on filled `--status-success` — checked, none remain; every radius uses a
   `--radius-*` token and no literal `#fff` exists in the file.
9. `.mini-nav-item--new` unused — the renamed `.mini-nav-item--active` is declared and consumed.
10. `"4 steps · ~5 minutes"` stale copy — removed outright rather than updated to a new number
    (the page no longer promises a step count or a time estimate).
11. Dark palette invented vs. faithful, `--overlay-scrim` never declared — checked line-for-line
    against `apps/web/src/index.css`'s `:root`/`[data-theme='dark']` blocks: every `--status-*` and
    `--overlay-scrim` value matches exactly in both themes.
12. `Closes #3118` won't fire against this base — this and the prior commit both say `Refs #3118`.

Rendered artifact (both revisions, before/after) for visual judgment:
https://claude.ai/code/artifact/730f2132-736a-491b-94ab-d30b4300c8b7

piotrswierzy (commented):
## Tech Lead review

Re-review of `d82d6b08`. The rebuild is a large improvement and closes three of my four prior blockers outright: the location step is real (`blocked-no-location` is a first-class state and `dialog-grant-refused` reproduces `assertRouterEnablementPreconditions`' actual 400 with its docblock's own remedy), `FulfillmentExecutor` is surfaced as a capability the operator must check afterwards, and there are now ten `data-state` values with a keyboard-navigable switcher plus three named dialogs. The header copy is ADR-055-exact (*"Adding this changes nothing on its own"*), the six-row `.attention-list` merge is the right IA call, and the theme block genuinely ships both palettes.

But the revision that closed the first blocker **reopened it in a different form**. The author's own note says the fix was to delete the *"Not built yet" / "Proposed"* badges and frame every row as shipped. The consequence is that the file now makes two positive claims about screens and behaviour that do not exist, with nothing in it marking them as proposals.

### Findings

**[BLOCKING]** — `docs/plans/mockups/oms-onboarding-3104.html:695`
> `dialog-grant-sourcing`'s description reads *"**One of three arrangements** on Who decides what."*
>
> There are exactly three shipped arrangements, and **none of them does what this dialog does**. I re-read `apps/api/src/fulfillment-authority/application/authority-presets.ts`: `leave-as-they-are` → `keepAsIs` (`:125`, returns the same reference, writes nothing); `openlinker-decides` → `disableClaimsPreservingAssignment` (`:138-151`), which sets `enabled: false` on every assignable claim — a **revocation**, never a grant; `keep-other-system` → `available: false` (`:172`). Nothing in `AUTHORITY_PRESETS` ever writes `sourcingAuthority.enabled: true`. This is the same dead end I blocked on last round, now asserted rather than drawn.
>
> The sentence is also false under the target-state reading — your own PR comment proposes this as a **fourth** arrangement. Either counting is wrong. Say *"A fourth arrangement, proposed by this epic — the three that ship today (`leave-as-they-are`, `openlinker-decides`, `keep-other-system`) all leave A2 unheld"*, and put the proposal marker back on this one row. `frontend-architecture.md` § UX Mockups is explicit that *"a stale mockup is worse than none, because it actively misleads"* — a target-state file is fine, a target-state file indistinguishable from a description of today is not.

**[IMPORTANT]** — `docs/plans/mockups/oms-onboarding-3104.html:709`
> The diff row asserts *"Orders OpenLinker routes are **no longer copied into your shop at all**"*, and the page's loudest warning builds on it (*"It is the one that surprises people… invoices, picking lists, stock counts, the customer's 'my orders' page… stops seeing them"*). I cannot ground that consequence:
> - `OrderSyncRequest.destinationConnectionIds` is the only mechanism that narrows the destination fan-out (`order-sync.service.ts:90`, `:518`), and **nothing populates it** — `grep destinationConnectionIds libs/core/src/orders/application/services/order-ingestion.service.ts` returns nothing. Populating it is deferred (#2400).
> - `architecture-overview.md` § Orders (#2397) is explicit in the other direction: *"`OrderSyncService` is **retained, not reinterpreted** under a fulfilment router — destination-mirror creation is a commercial/catalogue act, **distinct from fulfilment assignment**."*
>
> So enabling A2 does not by itself stop shop mirroring, either today or under the stated design; what stops it is whatever ids the router's plan names. This is the single scariest sentence on the page and it is the one an operator would act on. Either cite the mechanism that produces it, or restate it conditionally and mark it as the proposed consequence.

**[IMPORTANT]** — `docs/plans/mockups/oms-onboarding-3104.html:5-6`
> The header comment claims the tokens are *"copied **VERBATIM** from apps/web/src/index.css (`:root` 185-386, `html[data-theme='dark']` 395-505). Do not hand-tune."* I diffed both blocks: every value matches except the two font stacks, which are **not** verbatim — `--font-sans` drops `BlinkMacSystemFont`, `--font-mono` drops `Consolas` and `'Liberation Mono'`. Harmless in rendering while the CDN serves IBM Plex; the problem is the comment, which is the exact defect class flagged on the sibling #3124 ("ported verbatim" over a table that diverged). Fix the two stacks so the comment becomes true.

**[SUGGESTION]** — issue linkage
> `Refs #3118 (base is the epic branch, so Closes would not fire; the epic closes it on merge)` is the right call and correctly reasoned — noting only that it is better than the sibling #3124, which still carries a bare `Closes #3117` against a feature base.

### Verdict
❌ **Request changes** — one BLOCKING claim about a shipped screen that does not do what the dialog says.

*(Commenting rather than requesting changes formally, per review policy.)*

---

norbert-kulus-blockydevs:
## Review addressed — `e73c86e4`

All three findings fixed, plus three more instances of the BLOCKING finding's own defect class that
the review did not reach. One finding I verified and **did not** implement as written — detail below,
with the code.

**Preview link:** https://claude.ai/code/artifact/2c2cc062-1042-4e46-b196-4514dda9a793 — and see the
last section, because that link was showing the wrong revision for both of your reviews.

---

### BLOCKING · "One of three arrangements" — fixed

Confirmed exactly as you state. `AUTHORITY_PRESETS` has three members and none writes
`sourcingAuthority.enabled: true` — `keepAsIs` (`:125`) returns the same reference,
`disableClaimsPreservingAssignment` (`:138-151`) only ever sets `enabled: false`, `keep-other-system`
is `available: false` (`:172`).

The dialog now opens with **"Proposed — this arrangement does not exist yet"**, names all three
shipped arrangements and what each actually does, and states that this epic adds a fourth. The
preset ids and line references sit in an HTML comment beside it so the next reader can check the
claim without re-deriving it. The row that drives the dialog carries a `Proposed` badge and the same
explanation in its "Do this" block.

### IMPORTANT · "no longer copied into your shop at all" — grounded, not weakened

You offered two resolutions: cite the mechanism, or restate it conditionally. **I cited it, because
the claim is correct and the mechanism you searched is the wrong one.**

`#2397`'s `OrderSyncRequest.destinationConnectionIds` *narrows* the fan-out and nothing populates it
— that part of your finding is right. But what produces this consequence is `#2396`'s intercept, in
`libs/core/src/orders/application/services/order-ingestion.service.ts:589-613`:

```ts
const routing = await this.interceptFulfillmentRouting(order, connectionId, …);
await this.persistFulfillmentOutcome(order.id, routing);

let results = [];
if (routing.held) {
  // HELD: … No destination fan-out, and therefore no `syncStatus` rows —
  // writing `pending` here would claim a destination is still owed this order,
  // which is exactly what routing decided is not true.
  this.logger.log(`Withholding destination mirror for order ${order.id}: …`);
} else {
  results = await this.orderSyncService.syncOrder({ … });
}
```

`syncOrder` is not narrowed — it is **not called**. `toInterceptOutcome` (`:986-1045`) decides
`held`, and `'routed'` returns `{ held: true, block: null }`. So "orders OpenLinker routes are not
copied into your shop" is literally what the code does. This is also stated in the PR description
("`routing.held` skips `syncOrder` entirely — no fan-out, no `syncStatus` row").

Two things changed anyway, because a finding this sharp from a reviewer with the repo open is
evidence the page was not showing its work:

1. The mechanism, the file and the line range are now an HTML comment on that diff row — including
   why `destinationConnectionIds` is the wrong place to look.
2. **The warning gained the arms that still mirror.** `toInterceptOutcome` holds `routed`,
   `in-doubt`, `contended` and `already-live-elsewhere`, but `refused` (which includes a plan
   carrying one unsourceable line) and a cancelled order are **not** held and take their ordinary
   path. The alert now says so: three kinds of order still arrive in the shop exactly as they do
   today. That is the reassuring half, and it was missing.

The precision of *"orders OpenLinker **routes**"* was already load-bearing and is kept.

### IMPORTANT · "copied VERBATIM" — fixed, and further than reported

You found the two font stacks. I diffed all 188 token declarations across all three blocks
programmatically and found **15 more**: leading-zero shorthand (`.25rem` vs `0.25rem`,
`cubic-bezier(.16,1,.3,1)`, `rgb(0 0 0 / .40)`). Those render identically, but they make the word
"verbatim" false for anyone who diffs, which is the defect you actually flagged.

Every value in `:root`, the `prefers-color-scheme` block and `:root[data-theme='dark']` is now
byte-identical to `index.css`. Re-verified: **0 divergences** in both palettes, and the two dark
blocks are identical to each other (60 tokens). The header comment now states precisely what *does*
differ — the file carries the subset of tokens it uses, and one multi-line value is collapsed onto
one line — and names the font-stack regression so the next person does not repeat it.

### SUGGESTION · issue linkage — no change

Agreed, `Refs #3118` stands.

---

## Three more of the same defect, found while verifying yours

Your BLOCKING finding is "the file asserts a screen that does not do what it says." Checking it, I
found the file asserts **three screens that do not exist at all**. All three are now marked, with one
convention stated once on the page.

| Asserted | Reality |
|---|---|
| `Shipped — Products → Stock locations` (preview label, doItems, action label) | **No such screen.** No route, no `BASE_NAV_GROUPS` entry, no bulk-assign UI anywhere in `apps/web`. `product-detail-page.tsx:132-138` *reads* `locationId` and offers no way to set it. |
| "Open **Locations** to add one with a full address" (row 2 doItems + note, twice) | **No such screen.** `/inventory/locations` CRUD shipped in #2316; no page, no nav entry. |
| `Sourcing rules (which warehouse wins) →` (Related footer link) | **No such screen.** #2953 shipped the admin API under `connections/:connectionId/sourcing-rules`; `grep -r 'sourcing-rules' apps/web/src` returns nothing. |

`Dispatch & shipping labels` and `Who decides what` both check out — `RoutingRulesPanel` on the
connection Mappings page, and `/settings/who-decides`.

**The convention.** Rather than restore six "Not built yet" badges (which the merged-IA revision
deleted for good reasons), there is now one marker with one sentence explaining it, per
`frontend-architecture.md § UX Mockups`:

- a `Proposed` badge on a row whose **remedy** does not exist (the row's *fact* stays a live read —
  the stock count really is `GET /inventory?locationId` against the same list unfiltered),
- a dashed amber preview frame + a `Proposed —` label on a screen that does not exist,
- one paragraph under the checklist naming all three, so the page states its own status.

Row 2's note also no longer says a blank `MAIN` can be "renamed and given a country and postcode
from **Locations** at any time" — it says the API exists, nothing in the app reaches it, and a
warehouse you can describe waits for that screen.

---

## One claim of yours that the current revision contradicts, deliberately

Your first review blocked on `FulfillmentExecutor` never being enabled. The rebuild's answer was
"check the Capabilities panel afterwards"; `d82d6b08` flipped that to "**you do not need to** visit
the capabilities panel." Those are opposite, so I verified which is true:

- `libs/oms/src/oms.plugin.ts:72` — `supportedCapabilities: ['FulfillmentExecutor']`, no `InventoryMaster`.
- `connection.service.ts:449-452` — `defaultCapabilities = [...metadata.supportedCapabilities]`, and
  `enabledCapabilities = rest.enabledCapabilities ?? defaultCapabilities`.
- `create-connection.schema.ts:144-146` — the form **omits** the key entirely when the capabilities
  field is blank (`caps.length > 0 ? {…} : {}`).

So an OMS connection created through the form is stamped `FulfillmentExecutor` automatically. The
current copy is right and your first blocker is genuinely closed, not papered over.

---

## Verified, not asserted

- **Tokens**: 188 declarations diffed against `index.css` — 0 divergences, both palettes, dark blocks identical.
- **Render**: all 10 states × 4 breakpoints (1440/1280/768/390) × 2 themes — no JS errors, no horizontal overflow anywhere.
- **Dialogs**: all three open, each with `aria-labelledby` + `aria-describedby`, each writing `data-dialog`.
- **Contrast** (painted sRGB via canvas, both themes): every new amber element **4.67:1 to 12.69:1**, all AA.
- `check-design-tokens`, `check-css-structure`, `check-ui-vocabulary` all pass.

---

## The preview link was stale for both of your reviews

Worth flagging on its own. The artifact at that URL was byte-identical to **`e82d13660`** — the
revision *before* the attention-list merge — not `d82d6b08`. Your line numbers match the repo file,
so your review was against the right thing; but the PR description's claim that it was "republished
as version 3 … the file in this PR" was wrong, and anyone who clicked the link saw the six-step
timeline instead.

It now serves `e73c86e4`. Two takeaways: the PR body's preview claim needs checking before it is
made, and `frontend-architecture.md § UX Mockups` is right that E2E baselines must come from the
repo copy and never the artifact URL — this is exactly why.

---

piotrswierzy (commented):
## Tech Lead re-review (`d82d6b08` → `e73c86e4`)

One commit, and **the blocker is closed the right way round** — by marking the three unbuilt screens rather than by deleting the badges that said so. `dialog-grant-sourcing` now opens with "Proposed — this arrangement does not exist yet", enumerates the three that ship and states for each why it does not grant A2, and the HTML comment cites `authority-presets.ts` line by line. I checked: `keepAsIs` (`:125`) returns the same reference, `disableClaimsPreservingAssignment` (`:138`) only ever writes `enabled: false`, `keep-other-system` is `available: false` (`:172`). No member writes `sourcingAuthority.enabled: true`. The fourth-arrangement framing is correct, and the same `Proposed` treatment now reaches the attention rows, both previews and the related-links list — where `Sourcing rules` and `Locations` are demoted from links to labelled spans, which is the honest thing to do with an API-only surface.

The second finding is closed **and re-grounded on the right mechanism**. The warning no longer rests on #2397's `destinationConnectionIds`; it rests on #2396's `interceptFulfillmentRouting`, and that is the mechanism that actually produces the effect — `order-ingestion.service.ts:589` sits between `persistOrder` and `syncOrder`, and on `held` the fan-out is skipped outright with its own comment saying so. The `refused` arm at `:1036` returns `{held: false}` and logs "following today's path", so the new "three kinds of order are untouched" sentence is true and is a genuine improvement on the old absolute. Token verbatimness is restored (`BlinkMacSystemFont`, `Consolas`, `'Liberation Mono'` back; the header now says honestly that it carries a subset and collapses the font stacks rather than claiming a whole-file copy).

One claim in the same batch is still stronger than the code supports.

### Findings

**[IMPORTANT]** — `docs/plans/mockups/oms-onboarding-3104.html:899` (the `proposed:` comment on the *Stock tagged with a warehouse* row)
> "The FACT is live: `GET /inventory` takes a `locationId` filter and returns a total, so the count above is a real read." The filter is exact-match on one id — `inventory.repository.ts:289`, `if (filters.locationId) { where.locationId = filters.locationId; }` — and there is no *has-any-location* or `IS NULL` predicate on the read path anywhere. So "4,812 of 4,812 located" is not obtainable from one call; it needs either a fan-out over every location or a predicate that does not exist. This matters more than its size because the comment deliberately splits fact from remedy ("The FACT is live... The REMEDY is not") and puts this on the wrong side of that line — the row then reads as *the number is true, only the button is missing*, when the number needs building too. Either mark the count itself as proposed, or name the read that would produce it.

**[SUGGESTION]** — `docs/plans/mockups/oms-onboarding-3104.html:1280`
> No `?state=` / `#state=` addressability — `setState` is reachable only by clicking the switcher chips. This is not a deviation from convention (21 of the 22 mockups on `main` are the same), so it is genuinely optional; I raise it only because its sibling `correction-proposal-3087.html` ships it in this same wave, with the argument that driving a baseline by synthesising clicks on a mockup's own chrome ends up testing the switcher. Worth copying the eight lines if the E2E task for this epic is going to screenshot it.

**Positive.** The `.preview__label--proposed` comment — declared *after* `.preview__label` because equal specificity makes source order decisive — is the kind of thing that is invisible until it silently does not paint. Good that it is written down at the declaration.

### Verdict

🔄 **Approve with changes** — the blocker is closed, and closed by adding truth rather than removing a claim. Fix the located-stock count's grounding and this is ready.

`MERGEABLE` / `CLEAN`. **No CI checks reported on this branch at all** — none, rather than "Lint skipped" — expected for docs-only and stated rather than read as passing.

---

piotrswierzy (commented):
## Tech Lead re-verification (head unchanged at `e73c86e`)

`main` has not moved. The prior blocker stays closed the right way round — the unbuilt screens are marked rather than the badges deleted, `dialog-grant-sourcing` says "Proposed — this arrangement does not exist yet" and enumerates why none of the three shipped `AUTHORITY_PRESETS` grants A2, and the shop-mirror warning is re-grounded on #2396's `interceptFulfillmentRouting` (still present in the head blob).

The one IMPORTANT is unchanged and re-verified: `4,812` and the "live read" claim both still appear in the *Stock tagged with a warehouse* row. `GET /inventory`'s `locationId` filter is exact-match on a single id (`inventory.repository.ts:289`) with no has-any-location predicate anywhere, so "4,812 of 4,812" still needs building — the row asserts a read that does not exist.

**Review state worth knowing:** `reviewDecision` is still `CHANGES_REQUESTED` from the 2026-09-10 review; my two later passes were comments and did not clear it. That is inert today because the PR targets `3104-oms-onboarding-ui` (hence `CLEAN`), but it will gate the merge once the stack reaches `main` — someone has to submit an approval to supersede it.

**CI:** **no checks at all** on `3118-oms-onboarding-mockup`; `ci.yml` triggers on `pull_request: branches: [main]` only, so nothing ran. Absent, not passing.

### Findings
No new findings. The single standing IMPORTANT (the warehouse-count row claiming a live read the API cannot serve) is unaddressed.

### Verdict
🔄 **Approve with changes** — mark the warehouse-tagged count as unbuilt, the way the other unbuilt screens already are.

---

norbert-kulus-blockydevs:
## Addressed — `e73c86e4` → `0a197d6a`

Both standing findings are closed. One commit, docs-only.

### [IMPORTANT] The warehouse-tagged count claimed a live read — closed

You were right on both halves: the number is wrong *and* the comment put it on the wrong side of its own fact/remedy line, which is what made the row read as *the number is true, only the button is missing*.

I re-verified your reading and it holds exactly:

- `InventoryFilters` carries exactly four members — `productId`, `productVariantId`, `locationId`, `sourceConnectionId` (`inventory.types.ts:198-218`).
- `locationId` is exact-match on one id — `inventory.repository.ts:289`, `if (filters.locationId) { where.locationId = filters.locationId; }`. No has-any-location predicate, no `IS NULL` arm anywhere on the read path.
- `GET /inventory` exposes only those three as query params (`inventory.controller.ts:51`).
- `GET /inventory/duplicate-positions` groups by the four-column position key but reports **only keys holding more than one row**, so it cannot serve the count either. Worth recording, since it is the nearest-looking read and the next person will check it.

So the denominator is live (`findAndCount`'s `total` on an unfiltered call) and the numerator is not obtainable from any single call — it needs either a fan-out issuing one `GET /inventory` per row of `GET /inventory/locations` and summing each `total`, or a fifth `InventoryFilters` member that does not exist.

The fix takes your first option **and** your second: the count is marked proposed, and the read that would produce it is named. Four surfaces changed, not one, so the claim cannot survive in a screenshot:

1. The HTML comment now opens with *"BOTH HALVES OF THIS ROW ARE PROPOSED — the count as much as the button"*, records why the earlier split was wrong, and cites the three line references above plus the `duplicate-positions` dead end. It says explicitly that whoever builds this row has to build the read first.
2. The rendered `proposed:` copy: *"Both halves of this row are proposed and not built — the count as well as the screen. The total is a live read, but nothing today can tell you how much of it carries a warehouse: the inventory read filters on one warehouse at a time and has no 'has any warehouse' test, so the figure above needs a read built for it before a screen can show it."*
3. The `Today:` step no longer says *"the count above fills in on its own as it syncs"* — that implied a count nothing computes. It now says the underlying stock rows pick up a warehouse as the source syncs, and that no screen counts them for you yet. The `Proposed:` step names the read alongside the screen.
4. The preview label was screen-only (*"no such screen today"*); it now reads *"Neither this screen nor the count on it exists today: the inventory read filters one warehouse at a time and has no 'has any warehouse' test"*.

### [SUGGESTION] `?state=` addressability — taken

Copied from `correction-proposal-3087.html`, with your argument intact in the comment (driving a baseline by synthesising clicks on a mockup's own chrome ends up testing the switcher). Three things beyond the eight lines:

- **The three dialogs are addressable too**, each booting the page state that can legitimately reach it — `dialog-grant-refused` boots `blocked-no-location` (the refusal only exists while no location has been created) and the other two boot `in-progress`. Addressing a dialog without its page state would screenshot a dialog over a page that could not have produced it.
- **`syncUrl` carries the state**, so a URL copied out of the address bar reproduces the screen someone is looking at. `not-started` is omitted rather than written, so the default URL stays clean.
- **`window.__setMockupState`** is exposed for a spec that would rather not reload.

An unknown name is ignored rather than rendering the default, so a typo in a ticket shows up as "nothing happened" and not as a screenshot of the wrong screen.

### Verified

Headless Chromium against the file on this branch:

| Check | Result |
|---|---|
| 10 state cases (7 page states, 3 dialogs, 1 bogus name, 1 `#state=` form) | 10/10 resolve to the expected `body[data-state]` + `body[data-dialog]` |
| Bogus `?state=bogus-typo` | ignored → `not-started`, as intended |
| Horizontal overflow @ 1440 / 1280 / 768 / 390 | 0px at every width |
| JS errors / console errors | none |

### On the review state

Noted, and thanks for flagging it — `reviewDecision` is still `CHANGES_REQUESTED` from 2026-09-10 and the two later passes were comments, so it will gate the merge once the stack reaches `main` even though it is inert against `3104-oms-onboarding-ui` today. An approval on this head would supersede it whenever you are happy.

CI absence is expected and unchanged: `ci.yml` triggers on `pull_request: branches: [main]` only, so nothing runs on this branch — absent, not passing.

piotrswierzy (approved):
## Tech Lead re-verification (head unchanged at `e73c86e4`)

Base is the stack branch `3104-oms-onboarding-ui`; `MERGEABLE` / `CLEAN`, **no checks reported** on this branch (absent, not passing). Re-verifying against the file rather than against my own note changed the verdict.

### Findings

**WITHDRAWN — my standing IMPORTANT on the *Stock tagged with a warehouse* row does not hold.** I recorded that the row *"claims the count above is a live read"*. Read at this head, it claims the opposite, and in more detail than my finding asked for:

- The row's `proposed` copy reads *"Both halves of this row are **proposed and not built** — the count as well as the screen. The **total** is a live read, but nothing today can tell you how much of it carries a warehouse: the inventory read filters on **one** warehouse at a time and has no 'has any warehouse' test, so the figure above needs a read built for it before a screen can show it."*
- The source comment above it names the mechanism exactly as I did — `InventoryFilters` has four members, `locationId` is exact-match on one id at `inventory.repository.ts:289`, no has-any-location predicate and no `IS NULL` arm on the read path, `GET /inventory/duplicate-positions` cannot serve it either — and records that an earlier revision *did* split the halves wrongly and was corrected.
- The page-level *"every figure on it is a live read"* sentence at `:670` is scoped by an explicit "Everything **else** on this page" carve-out, so it does not cover this row.

I re-checked `inventory.repository.ts:289` on `main` and the exact-match behaviour is exactly as both the mockup and my finding describe — the disagreement was never about the code, only about what the mockup asserted, and the mockup asserts the right thing. My note was stale. Nothing to fix here.

**Procedural — this PR's `reviewDecision` is still `CHANGES_REQUESTED`** from the 2026-09-10 review; the two later passes were comments and did not clear it. That is inert while the base is a stack branch, but it will gate the merge once the stack reaches `main`. This review is the explicit approval that supersedes it.

### Verdict
✅ **Approve**

---

PR #3126, opened 2026-09-10
test(worker): isolate every integration test case by default (#2999)

### LINKED ISSUE #2999: [TASK] DX — worker integration harness: six int-spec files never reset, and no setupFilesAfterEnv exists
## Problem / Context

`apps/worker/test/integration/setup.ts:212` exports `resetTestHarness()`, and **6 of the 27** worker int-spec files never call it:

- `automation-dispatch-boot.int-spec.ts`
- `automation-emission-boot.int-spec.ts`
- `fulfillment-no-injection-boot.int-spec.ts`
- `fulfillment-router-binding-boot.int-spec.ts`
- `invoicing-auto-issue-boot.int-spec.ts`
- `oms-module-boot.int-spec.ts`

`apps/worker/test/jest-integration.cjs` declares `maxWorkers: 1`, `globalSetup` and `globalTeardown`, and **no `setupFilesAfterEnv` at all** — so there is no place a root-level hook could be registered without adding one. (`apps/worker/jest.config.js:19` does have `setupFilesAfterEnv`, but that is the **unit** config and does not apply to the integration run.)

This is the worker-side twin of the gap #2986 closed for `apps/api` in PR #2996, and the evidence for why it matters is on the api side: **41 assertions in `paginated-total-split.int-spec.ts` were passing only because of accidental row state** left by whichever spec ran immediately before it. They were not true in isolation. With `maxWorkers: 1` and one shared harness, the same class of failure is one file-ordering change away here.

## The reason this is NOT a copy of the api fix

`apps/worker`'s `reset()` (`setup.ts:77-102`) is materially more destructive than the api's:

```
TRUNCATE sync_jobs / inventory_items / product_variants / products
       / identifier_mappings / connection_cursors / connections   -- 7 tables
await this.redisClient.flushDb();                                 -- ALL of Redis
```

A root-level `beforeEach` calling this would **`flushDb()` between every test case**, and the worker harness boots real stream consumers — `job-intake`, the events consumer, `master-deletion-offer-pause`. `flushDb()` destroys their **consumer groups and Pending Entries Lists** while the consumers still hold their connections. A spec that creates a consumer group in `beforeAll` and asserts stream behaviour across several test cases would break on its second case, and the failure would look like a consumer bug rather than a harness one.

So the Redis half must be audited before any hook is registered. The api fix had no equivalent exposure: its reset truncates tables only, and Postgres `TRUNCATE` cannot cross a database boundary — which is what made the two second-database specs (`fulfillment-work-migration-parity`, `oms-connection-never-seeded`) provably safe there.

## Proposed Solution

1. **Audit first, then wire.** For each of the 27 specs, determine what a `flushDb()` between test cases would destroy: consumer groups, PELs, dedup keys (`jobdedup:*`), locks (`perf:stand:exclusive`-shaped `SyncLockPort` keys), and any state a `beforeAll` establishes. Report the set that would break before changing configuration.
2. **Choose the shape from the audit, not from the api precedent.** Candidate shapes, in the order they should be considered:
   - a root-level `beforeEach`/`afterEach` via a new `apps/worker/test/integration/setup-each.ts` wired from `setupFilesAfterEnv` in `jest-integration.cjs` — the api shape (`apps/api/test/integration/setup-each.ts`), correct only if no spec depends on Redis state surviving between its own cases;
   - a Postgres-only reset in the global hook with the Redis flush left to per-file opt-in, if the audit finds stream-dependent specs;
   - per-file `beforeEach` in the six, if a global hook is unsafe for any file.
3. **Ship a guard that fails when the registration is removed.** `apps/api/test/integration/harness-isolation.int-spec.ts` is the precedent: it registers no reset of its own, leaves a row behind in one case and asserts it is gone in the next, so deleting the one-line registration turns it red. A guard asserted only in the passing direction proves nothing — this campaign has found four such guards.

## Classification

**Type**: DX
**File(s)**:
- `apps/worker/test/jest-integration.cjs` — no `setupFilesAfterEnv`
- `apps/worker/test/integration/setup.ts:77-102, 212` — `reset()` and `resetTestHarness()`
- the six spec files listed above
- `apps/api/test/integration/setup-each.ts` + `harness-isolation.int-spec.ts` — the api-side precedent to follow or deviate from with reasons

## Dependencies

- **#2986 / PR #2996** should merge first, so the api-side shape is settled and this issue either reuses it or documents why it cannot.

## Assumptions

- The six files were not deliberately exempted; none carries a comment claiming it needs inherited state.
- The worker integration suite runs `maxWorkers: 1` against one shared harness, as its config declares, so file execution order is the only isolation mechanism today.

## Acceptance Criteria

- [ ] Every one of the 27 worker int-spec files is isolated by default — a spec's first test case cannot read rows or Redis keys left by a previously-run file
- [ ] The audit of what `flushDb()` destroys per spec is recorded in the PR description, naming every spec that depends on Redis state surviving between its own test cases (or stating that none does)
- [ ] The chosen shape is justified against the api precedent — either it reuses `setupFilesAfterEnv` + a `setup-each.ts`, or the PR states which audited finding rules that out
- [ ] A guard spec registers no reset of its own and **fails** when the global registration is deleted, verified by actually deleting it and observing the failure
- [ ] The added per-test cost is stated (the worker reset issues 7 unconditional `TRUNCATE`s plus a `flushDb()`, none of them probe-guarded, unlike the api's `truncateTables`) — and if it is material, a probe-first optimisation is either applied or filed
- [ ] `pnpm --filter @openlinker/worker test:integration` passes, and any newly-red spec in a file this change does not touch is reported as a pre-existing false assertion rather than fixed silently

norbert-kulus-blockydevs:
## Summary

- `apps/worker/test/jest-integration.cjs` declared no `setupFilesAfterEnv`, so six int-spec files (`automation-dispatch-boot`, `automation-emission-boot`, `fulfillment-no-injection-boot`, `fulfillment-router-binding-boot`, `invoicing-auto-issue-boot`, `oms-module-boot`) never called `resetTestHarness()` between their own test cases, and the other 21 only reset in `afterEach` — a courtesy to the *next* file, not isolation for the current one, given `maxWorkers: 1` and no `testSequencer`.
- Adds `apps/worker/test/integration/setup-each.ts`, registered via `setupFilesAfterEnv`, wiring a root-level `beforeEach` + `afterEach` that calls `resetTestHarness()` for every test case of every int-spec — the worker-side twin of #2986/PR #2996 (api-side; merged to `perf-programme-2840`, not yet to `main`).
- **This is not a straight port.** The worker's `reset()` also calls `redisClient.flushDb()` (wiping *all* of Redis) and the worker harness boots a live background stream consumer (`MasterDeletionToJobHandler`), so a global `beforeEach`/`afterEach` needed its own audit before it could be registered safely. The full per-file audit is recorded in `setup-each.ts`'s docblock; the conclusion is that **no worker int-spec depends on Redis state surviving between its own test cases**:
  - `stream-consumer-recovery.int-spec.ts` and `stream-retention.int-spec.ts` mint a fresh, uniquely-named stream (and therefore consumer group) inside *each* `it()`, never in `beforeAll`.
  - `job-intake-execution.int-spec.ts` writes to `jobs.sync` in one case and never reads it back in a later one.
  - `master-inventory-deletion-e2e.int-spec.ts` reads the deletion-event stream from inside each test body, against state that same test body produced.
  - The six never-reset "boot" specs only resolve DI providers from the already-booted container; none seeds a row or Redis key in `beforeAll` that a later case reads.
  - `MasterDeletionToJobHandler`'s live consume loop already runs against an unconditional `flushDb()` in all 21 already-resetting specs today — this change doesn't alter that exposure, only extends the same reset uniformly.
- Makes `reset()` probe before acting: one combined `TRUNCATE ... WHERE EXISTS` round-trip (mirroring `truncateTables` in `libs/test-kit/src/harness.ts`, which is intentionally not exported for reuse outside its own spec) and a Redis `DBSIZE` check before `flushDb()` — since the reset now runs far more often (every `beforeEach` *and* `afterEach` of every case) than the previous occasional per-file `afterEach`.
- Adds `harness-isolation.int-spec.ts` as a regression guard, mirroring the api precedent: it registers no reset of its own, leaves a connection row behind in one case, and asserts it's gone in the next — so deleting the `setupFilesAfterEnv` registration turns it red.
- Adds a `docs/lessons.md` entry recording why the worker fix needed its own audit rather than copying the api one.

## Test plan

**Integration tests could not be run locally** — the environment this PR was authored in cannot run Docker/Testcontainers, so `pnpm --filter @openlinker/worker test:integration` was not executed. What a reviewer or CI should verify instead:

- [ ] `pnpm --filter @openlinker/worker test:integration` passes in full, including the new `harness-isolation.int-spec.ts`.
- [ ] Manually delete the `setupFilesAfterEnv` line in `apps/worker/test/jest-integration.cjs` and re-run `harness-isolation.int-spec.ts` alone — it should go **red** (this was the acceptance-criteria requirement to verify the guard actually guards something; I could not perform this myself without Docker).
- [ ] Re-run the two stream-focused specs (`stream-consumer-recovery.int-spec.ts`, `stream-retention.int-spec.ts`) specifically, since they're the highest-risk files for a global Redis flush — confirm no new flakiness versus a run without this change.
- [ ] Confirm no newly-red spec in a file this change doesn't touch; if one appears, per the issue's acceptance criteria it should be reported as a pre-existing false assertion (the api-side precedent found 41 such assertions) rather than silently patched here.
- [ ] `pnpm lint` / `pnpm type-check` — note `apps/worker/test/**` is excluded from both per existing repo convention (see e.g. `fulfillment-router-binding-boot.int-spec.ts`'s own docblock), so these new/changed test files are not covered by either gate; I reviewed them by hand for TypeScript correctness and confirmed `redisClient.dbSize()` is a valid method on `RedisClientType` (redis@4.6.12 / @redis/client@1.5.13) by inspecting the installed package's type declarations.

Closes #2999


norbert-kulus-blockydevs:
## PR Review — #3126 (senior backend architect pass, test infrastructure)

**Scope reviewed**: `setup-each.ts` (new global `beforeEach`/`afterEach`), `setup.ts`'s `reset()` (probe-before-truncate/flush), `jest-integration.cjs` wiring, `harness-isolation.int-spec.ts` regression guard, `docs/lessons.md` entry.

### Verdict: ✅ Approve

This is unusually well self-audited work, and I verified the load-bearing claims against the actual repo rather than trusting the PR description:

- **"Six specs never reset between their own cases"** — verified by grepping every `apps/worker/test/integration/*.int-spec.ts` for `resetTestHarness` occurrences: exactly `automation-dispatch-boot`, `automation-emission-boot`, `fulfillment-no-injection-boot`, `fulfillment-router-binding-boot`, `invoicing-auto-issue-boot`, `oms-module-boot` show 0 hits, all 21 others show ≥2. Matches the claim exactly.
- **"None of the six boot specs creates a DB row/Redis key in `beforeAll` that a later `it()` reads"** — read all six files. Confirmed: every `it()` only resolves DI providers from the already-booted container. `automation-dispatch-boot` is the one file that looks like it *executes* real logic (dispatches through the real dispatcher), but it stubs the write side (`jest.spyOn(recorder, 'record').mockResolvedValue(undefined)`), so no `automation_runs` row is ever persisted — genuinely safe to truncate around.
- **Cross-test fixture-creation risk** — I specifically checked whether any of the 21 already-resetting files seed a `Connection`/`Product` row in `beforeAll` that the *first* `it()` in the file depends on (a pattern the new global `beforeEach` would silently wipe before that first test runs, since it fires right after `beforeAll` and before every `it()`). Spot-checked `master-inventory-sync-all-e2e`, `inventory-provenance-backfill-e2e`, `master-inventory-deletion-e2e`, `woocommerce-shop-product-inventory-writeback`: in every case, `beforeAll` only resolves harness/services, and `createTestConnection`/`repo.save(...)` calls live *inside* each `it()` body, not in `beforeAll`. No hit on this failure mode across the files I checked.
- **Redis stream claims** — verified `stream-consumer-recovery.int-spec.ts` and `stream-retention.int-spec.ts` mint a fresh, uniquely-named stream (`randomUUID()`) inside every `it()` via a `freshStream()` helper called from the test body, never from `beforeAll`, and both already carry their own `afterEach(resetTestHarness)`.
- **`redisClient.dbSize()` exists** on the installed `redis@4.6.12`/`@redis/client` package (`commands.d.ts`) — the PR's manual verification claim (made in lieu of running lint/type-check, since `apps/worker/test/**` is excluded from both) checks out. Confirmed the exclusion too: `apps/worker/tsconfig.json` has `"include": ["src/**/*"]`, so `tsc --noEmit` never touches `test/**`.
- **API-side precedent** — confirmed `apps/api/test/integration/setup-each.ts` does **not** exist on `origin/main` but does exist on `origin/perf-programme-2840`, exactly as the docblock states (avoids the "cites a merged PR that isn't actually merged" trap).
- **Import correctness** — `@openlinker/core/identifier-mapping/orm-entities` sub-barrel re-exports `ConnectionOrmEntity`, and `createTestConnection` exists in the referenced helper file.

### Architecture

Correctly placed: `setupFilesAfterEnv` (not `setup.ts` itself, which is also imported from `globalSetup`/`globalTeardown` realms where Jest globals don't exist) is the right seam for a hook that applies by omission. The `docs/lessons.md` entry is a good addition — it records *why* this couldn't be a blind port of the API-side fix (the worker's `reset()` is strictly more destructive: whole-DB `flushDb()` plus a live background stream consumer), which is exactly the kind of institutional knowledge that prevents the next person from copy-pasting a fix that doesn't hold in a different context.

The performance mitigation (probe via one combined `UNION ALL ... WHERE EXISTS` query + Redis `DBSIZE` before doing any actual `TRUNCATE`/`flushDb()`) is a sensible response to running the reset twice per test case instead of once per file.

### Residual gap (non-blocking, already disclosed by the author)

Nobody has actually run `pnpm --filter @openlinker/worker test:integration` against this — the author discloses no Docker access, and I have the same constraint. The PR's own test plan checklist correctly identifies this as required before merge (including the "delete the `setupFilesAfterEnv` line, confirm `harness-isolation.int-spec.ts` goes red" acceptance criterion). That checklist should be treated as a hard merge gate by whoever has Docker access, not a nice-to-have — everything I verified is *static* correctness; the dynamic claim (does the suite actually stay green) is unverified by either of us.

piotrswierzy (commented):

Architecturally right and unusually well-reasoned: it moves worker int-spec isolation from *"every author remembers"* to *"applies by omission"*, and it does **not** blindly port the api fix — the per-file audit in `setup-each.ts` is the correct instinct for a reset that flushes all of Redis.

**The vacuity concern is answered decisively, and I verified it independently rather than taking the audit's word.** A brace-matching scan of every `beforeAll` block (outer *and* inner `describe`) across all 27 specs found **zero** that write a DB row, Redis key or stream entry — every one only resolves providers and sets env vars. None of the six newly-reset boot specs writes any state at all. So no assertion can become vacuous or newly-false, and Jest's `beforeAll → root beforeEach` ordering cannot destroy seeded fixtures because there are none.

The direction of any residual risk is also safe: stricter isolation makes a state-dependent assertion **fail loudly**, never pass vacuously.

**The perf question resolves in the right direction** — steady state goes from 7 unconditional `TRUNCATE`s + unconditional `flushDb()` to two O(1) probes (`dbSize()` is genuinely available). 119 tests × 2 hooks is a net *reduction*. And `resetTestHarness()` no-ops when `globalHarness` is null, so a spec that never boots the harness pays nothing and cannot boot containers by accident — a real hazard got right by reusing the existing guard.

All headline claims verified: 27 int-specs, exactly the six named files never reset, `jest-integration.cjs` had no `setupFilesAfterEnv`.

### BLOCKING — nothing has ever executed this change

`apps/worker/tsconfig.json` has `include: ["src/**/*"]`, so `apps/worker/test/**` is outside both `pnpm lint` and `pnpm type-check` (the PR states this correctly), and Docker wasn't available. The entire surface — `setup-each.ts`, the rewritten `reset()`, `harness-isolation.int-spec.ts` — is exercised **only** by `pnpm --filter @openlinker/worker test:integration`. The repo's convention is that the integration gate is not deferred to CI.

Run the full worker suite plus the **red-first check**: delete the `setupFilesAfterEach` line and confirm `harness-isolation.int-spec.ts` goes red. This is the merge gate, not a defect.

### IMPORTANT — the probe drops the `pg_constraint` CASCADE-closure walk

`setup.ts` (~95–120)

The probe-then-truncate is copied from `libs/test-kit`'s `truncateTables` **minus its CASCADE-closure expansion** — the one mechanism that makes probe-then-truncate equivalent to unconditional truncate. `docs/testing-guide.md` § Testcontainers Lifecycle documents the closure as part of this pattern (*"The probe runs over the list's CASCADE closure, not the literal list … a future FK is covered automatically"*). Previously an unconditional `TRUNCATE connections CASCADE` cleared unlisted dependents as collateral; now, if the listed parent happens to be empty, that collateral is retained.

**No live leak today** — I checked: the only FK from an unlisted table into a listed one is `reservations.inventoryItemId → inventory_items`, which is `NOT NULL`, so the child cannot be non-empty while the parent is. But the guarantee has become a per-schema reasoning obligation over a ~100-table schema that grows, with only 7 tables listed.

Also, the docblock's claim that `truncateTables` is *"intentionally not exported for reuse outside its own spec"* is loose. It **is** an exported function; it is merely absent from `libs/test-kit/src/index.ts`. And `jest-integration.cjs:46` already maps `^@openlinker/test-kit/(.*)$ → src/$1`, so `import { truncateTables } from '@openlinker/test-kit/harness'` resolves in this exact realm — the sanctioned test-only deep import. Importing it keeps one implementation and inherits both the closure walk and its `UNQUOTED_IDENTIFIER` assertion.

### IMPORTANT — the audit's one factual error understates the new exposure

`setup-each.ts`, the `MasterDeletionToJobHandler` bullet

The claim *"this change doesn't alter that exposure, only extends the same reset uniformly"* is not accurate, and the inaccuracy lands in the six files the change is for. `initializeConsumerGroup()` runs **once**, at `onModuleInit`, with `MKSTREAM: true`; `consumeLoop()` catches every read error, logs at **`error`** level, sleeps 1 s and retries — it never recreates the group (`master-deletion-to-job.handler.ts:113–205`). So the *first* `flushDb()` destroys `events.master.deletion` and the `master-deletion-offer-pause` group permanently for the rest of that file's run. The six boot specs previously flushed **never**; they now emit an error-level line per second for the remainder of the file — in precisely the specs whose purpose is diagnosing boot wiring.

No assertion breaks (`master-inventory-deletion-e2e.int-spec.ts` reads the stream via `xRange`, not the consumer group), so this is diagnosability rather than correctness. Cheapest fixes: re-create the group after a flush inside `reset()`, or exclude that key. At minimum correct the bullet — it currently reads as *"nothing changed here"*, which is the sentence a future reader will trust.

### SUGGESTIONS

- `harness-isolation.int-spec.ts` cannot distinguish the two hooks (its docblock says so honestly) — which means **deleting only the `beforeEach` line leaves it green** while removing the half `setup-each.ts` itself calls *"the load-bearing one"*. That edit is plausible precisely because `afterEach` looks sufficient. One sentence in `setup-each.ts` saying the `beforeEach` half is guarded by nothing.
- `docs/testing-guide.md` still instructs authors to call `resetTestHarness()` in `afterEach` and lists it as an obligation. For the worker suite that is now redundant (a third reset per test across 21 specs). The `docs/lessons.md` entry records *why the audit was needed*, not *that the obligation moved*.
- The 21 specs' own `resetTestHarness()` calls are left in deliberately (idempotent, now cheap) — fine, but the PR doesn't say whether removing them is a follow-up, so the redundancy will read as required to the next author.

---

norbert-kulus-blockydevs:
## Review addressed — both IMPORTANTs and all three suggestions. The BLOCKING is **not** closed.

@piotrswierzy thank you, and thank you in particular for re-running the vacuity audit independently rather than taking the file's word for it — that is the half I could not have checked for myself. Pushed as `8fa0c88`.

Taking the BLOCKING first, because it is the one that decides whether this merges.

---

### BLOCKING — nothing has ever executed this change → **still open, honestly**

Not closed, and I am not going to claim otherwise. `apps/worker/test/**` is outside both `pnpm lint` and `pnpm type-check` (as the description says), and the integration suite is not runnable here — test execution is offloaded to CI on this machine, and this worktree has no local install either.

What I *could* do, I did: the three touched test files were type-checked through a temporary project that includes them, which is clean. That catches the `@openlinker/test-kit/harness` import resolving and the `truncateTables` call shape lining up with `DataSource` — it says nothing about runtime behaviour.

So the merge gate stands as you set it: `pnpm --filter @openlinker/worker test:integration` plus the **red-first** check (delete the `setupFilesAfterEnv` line, confirm `harness-isolation.int-spec.ts` goes red). One thing has changed about that second half, though — see the third suggestion below: deleting only the `beforeEach` line now goes red too, which it did not before.

---

### IMPORTANT — the probe drops the `pg_constraint` CASCADE-closure walk → **fixed by importing the helper, not by re-adding the walk**

`reset()` now calls `truncateTables` from `@openlinker/test-kit/harness`. One implementation, and it inherits both the closure walk and the `UNQUOTED_IDENTIFIER` assertion.

You were right on the loose claim too, and it was worse than loose — it was the reason the code was duplicated at all. The helper **is** exported from `harness.ts`; it is only absent from `libs/test-kit/src/index.ts`, and `jest-integration.cjs:46` already maps `^@openlinker/test-kit/(.*)$ → src/$1`. The docblock now says that instead, and notes that `setup.ts` is never loaded from the `globalSetup`/`globalTeardown` realms, where that mapping would not apply.

One thing the fix needed that was not obvious: `apps/worker/tsconfig.json` declares its own `paths`, which **shadows the base config's entirely**, and it had no `@openlinker/test-kit` entry — so the import resolves for jest but is a `TS2307` under ts-jest, which compiles the test files with those compilerOptions. Both entries added. `include` is still `src/**/*`, so nothing about the build changes.

Your assessment that there is no live leak today matches mine — `reservations.inventoryItemId → inventory_items` is `NOT NULL`, so the child cannot be non-empty while the parent is. The point was never today's schema.

### IMPORTANT — the audit's factual error understates the new exposure → **corrected, and the exposure removed**

Verified your reading line by line, and it is exactly right: `initializeConsumerGroup()` runs once at `onModuleInit` with `MKSTREAM: true`, and `consumeLoop()` catches every read error, logs at **`error`**, sleeps 1 s and retries without ever recreating the group. The six boot specs previously flushed *never*.

Both halves done rather than just the bullet:

- **The bullet is corrected** — it now opens by saying the earlier "this change doesn't alter that exposure" was wrong, and states the real mechanism, including that no assertion breaks and that this is diagnosability rather than correctness. That last part is why it needed catching by a reviewer, and the sentence says so.
- **`reset()` recreates the stream and group immediately after the flush** (`restoreMasterDeletionConsumerGroup`), best-effort, tolerating `BUSYGROUP`. I took recreate over exclude-the-key because `flushDb()` is all-or-nothing and excluding would mean switching to targeted deletion.

The group name is now exported from the handler (`MASTER_DELETION_CONSUMER_GROUP`) and consumed by both sides, rather than a literal in the harness that would drift silently — and the symptom of that drift is a log line per second, not a failure.

The remaining window is one read between the flush and the recreate, so at most one error line rather than one per second. Noted in the method's docblock rather than left to be discovered.

---

### SUGGESTION — `harness-isolation.int-spec.ts` cannot distinguish the two hooks → **it can now**

This was the sharpest of the three, because as you say the edit that breaks it is *plausible*: `afterEach` looks sufficient.

There is now a case that only the `beforeEach` half can satisfy — a row written in **`beforeAll`**, asserted gone in the **first** `it()`. Nothing has finished at that point, so no `afterEach` can have cleared it; only the root `beforeEach` can. Hook order is what makes it sound and the docblock spells it out. Non-vacuity is asserted first (the `beforeAll` write really happened), and it compares against the post-write count rather than literal zero, keeping the file's existing rule about not claiming `AppModule` seeds nothing.

I went for the assertion rather than the one-sentence comment you suggested, because a comment asking the next author to notice is the same class of thing the whole PR is removing.

### SUGGESTION — `docs/testing-guide.md` still instructs authors to call `resetTestHarness()` → **split per app**

The guidance was unqualified and is now per app: for `apps/worker` the `afterEach` is dropped (it is a third reset per test), with pointers to `setup-each.ts` for the audit and to the guard file. It also states explicitly that **`apps/api` has NOT got the registration yet** (#2986 / PR #2996) — I checked this branch rather than assuming, and api's `jest-integration.cjs` still declares no `setupFilesAfterEnv`. Implying both were done would have been a worse error than the one being fixed.

`§ Keep Tests Independent` is split the same way.

### SUGGESTION — say whether removing the 21 specs' calls is a follow-up → **said, in `setup-each.ts`**

Kept deliberately, and the reason is recorded where someone will read it: idempotent, nearly free now that `reset()` probes, so removal is cosmetic, touches 21 files and would conflict with everything in flight. Recorded as redundant-not-required, with the follow-up named as a follow-up — so it does not read as required to the next author, which was your concern.

---

### Verification

| | |
|---|---|
| `@openlinker/worker` type-check | clean |
| Touched test files, type-checked via a temporary including project | clean |
| eslint on `master-deletion-to-job.handler.ts` | clean |
| eslint on the test files | only pre-existing errors in untouched lines of `setup.ts` (`test/**` is outside the worker's lint glob) |
| **Worker integration suite** | **not run — the BLOCKING above** |

norbert-kulus-blockydevs:
## Correcting my previous comment: the BLOCKING is now closed — and there was a second reason it could not have been

I said the integration suite had not run and that CI was the gate. CI has now run it, and the reason it never had is worse than "nobody triggered it".

**This branch was `CONFLICTING` against `main`.** A `pull_request` workflow builds `refs/pull/N/merge`, and that ref cannot be created while the merge fails — so there was **no CI run on this PR at all since 2026-09-10**, and the BLOCKING could not have been closed by CI either. Merged `main` in (`5aa8458`); one conflict, in `docs/lessons.md`, where both sides had appended a new entry at the same point — both kept, main's #3031 first, this branch's #2999 after it, nothing dropped. I re-checked after the merge that `apps/api` still has no `setupFilesAfterEnv`, since `main` landing the api twin would have made this branch's `docs/testing-guide.md` claim false.

### The result

**All 8 checks pass.** The worker integration suite: **28 suites, 122 tests, all passed** (33m05s), with `harness-isolation.int-spec.ts` green — including the new `beforeAll`-writes / first-`it()`-asserts-gone case that distinguishes the two hooks.

That exercises, for the first time, everything the review was about:

- the `@openlinker/test-kit/harness` deep import resolving under ts-jest with the added `paths` entries;
- `truncateTables`' CASCADE closure running against the real schema;
- `restoreMasterDeletionConsumerGroup` recreating stream and group after every `flushDb()` — across all 28 suites, so the consume loop stayed healthy;
- the global `beforeEach`/`afterEach` against all 27 pre-existing specs plus the guard.

`Test` (unit) is also green — 180 suites / 1510 tests.

### What CI still does not prove — the red-first half

You asked for two things: the suite, and **delete the `setupFilesAfterEnv` line, confirm `harness-isolation.int-spec.ts` goes red**. CI has done the first. It cannot do the second, and I have not done it either — it needs a deliberate local break-and-run, and test execution is offloaded to CI on this machine.

So: the green direction is proved, the red direction is not. If you want that closed before merge it needs someone with a working Docker to break the line and run `pnpm --filter @openlinker/worker test:integration` once. Worth noting the guard is now strictly stronger than when you reviewed — deleting **either** hook turns it red, where before deleting only the `beforeEach` left it green.

piotrswierzy (approved):
## Tech Lead review (re-review)

Both prior IMPORTANTs and all three suggestions are addressed, and the BLOCKING is now closed. `reset()` calls `truncateTables` from `@openlinker/test-kit/harness` itself rather than a copy, so it inherits the `pg_constraint` CASCADE closure and the `UNQUOTED_IDENTIFIER` assertion (verified: `libs/test-kit/src/harness.ts:181` exports it, `:123` does the closure walk, `jest-integration.cjs:45-46` maps the subpath, and `@openlinker/test-kit` is already a declared dep of `apps/worker`, so the deep import is the sanctioned test-only one and adds no undeclared workspace edge). The consumer-group exposure is not just corrected in prose — `restoreMasterDeletionConsumerGroup` puts stream and group back after every flush, keyed off a single exported `MASTER_DELETION_CONSUMER_GROUP` so the two literals cannot drift.

**The red-first half is now provable by reading, not only by running.** The guard's first case asserts `count < countAfterBeforeAllWrite` against a row written in `beforeAll`; with `setupFilesAfterEnv` deleted nothing clears it and the assertion is false. The third case asserts `count === baseline` after case 2 added a row; same. Deleting only `beforeEach` still reddens case 1, because no `afterEach` has run when it executes. Both non-vacuity pre-assertions are present, so neither case can pass for the wrong reason. Combined with the green CI run (28 suites / 122 tests, 33m), the "nothing has ever executed this" blocker is closed.

### Findings

**[SUGGESTION]** — `apps/worker/test/integration/setup.ts` (the `dbSize()` probe) and `setup-each.ts` § Cost
> The Redis half of the probe can never short-circuit, so the stated saving does not exist. `MasterDeletionToJobHandler.initializeConsumerGroup` runs at `onModuleInit` with `MKSTREAM: true` (`master-deletion-to-job.handler.ts:114-117`) and the consumer is on by default (`:87`, `!== 'false'`), so the `events.master.deletion` stream key exists from harness boot — and after the first flush `restoreMasterDeletionConsumerGroup` recreates it, again with `MKSTREAM`. `DBSIZE` is therefore `>= 1` for the entire life of the harness, so every reset takes the `flushDb()` + recreate path. The docblock's claim that "a reset immediately following another reset … now costs two cheap probe round-trips and no `TRUNCATE`/`flushDb()` at all" is only true of the Postgres half. Harmless in cost terms (a flush of a near-empty db plus one `XGROUP CREATE`), but it is the kind of sentence a future reader will optimise against. Either drop the Redis half of the claim, or skip the flush when `dbSize <= 1` — the latter trades a true statement for a fragile constant, so correcting the prose is the better fix.

**[SUGGESTION]** — `docs/testing-guide.md` (the new `apps/api is NOT there yet` note)
> Accurate today (verified: `apps/api/test/jest-integration.cjs` still declares no `setupFilesAfterEnv`), but it is a cross-branch claim about PR #2996 that goes stale silently the moment the api twin merges. Nothing under `check:invariants` can observe it. Worth a one-line pointer in #2996 to delete this paragraph when it lands.

Nothing else. The audit in `setup-each.ts` now carries the corrected consumer-loop paragraph rather than the "this change doesn't alter that exposure" sentence, the `docs/lessons.md` conflict resolution kept both entries (main's #3031 prettier lesson at :1232, this one at :1267), and the production-source change is a single extracted constant with no behavioural effect.

### Verdict
✅ **Approve**

---

Issue #2468, opened 2026-08-25
[TASK] Interface + CORE — Currency bulk remediation: `POST /analytics/coverage/currency/recalculate`


norbert-kulus-blockydevs:
### Problem / Context

The manual fix on the demo DB nulled the six stamp columns and inserted 13 `sync_jobs` rows with `jobType: 'marketplace.order.fxStamp'`, verified via `SELECT status, outcome FROM sync_jobs`. This endpoint must do exactly that, via `OrderFxStampService`, never by reaching into `sync_jobs` directly from the controller.

### Proposed Solution

`AnalyticsCoverageController` (or a sibling remediation controller) exposes `POST .../currency/recalculate`: creates one `analytics_remediation_runs` row (`status: 'in-progress'`, `affectedCount` = the count from Phase 4 Task 4.1's detector), then calls `OrderFxStampService.sweep()` (or the equivalent per-order `stamp()` loop the existing service already exposes — do not add a new bulk-enqueue path if `sweep()` already covers it), bounded and paged so it fixes **every** currently-mismatched order, not just newly-created ones. The worker's existing `marketplace-order-fx-stamp.handler.ts` / `marketplace-order-fx-stamp-sweep.handler.ts` do the actual work unchanged. On the sweep's completion (job success/failure), update the `analytics_remediation_runs` row to `resolved` or `failed` (with `detail` populated from the job's terminal failure reason — reuse `FX_STAMP_TERMINAL_REASONS` for the message rather than a generic "something went wrong").

Also expose `GET .../currency/status/{runId}` (poll the run's current lifecycle state) and `GET .../currency/orders` (paginated list of affected orders, backing the `detail-currency` modal).

**Optional, only if scope allows**: an `autoRecalculateOnCurrencyChange: boolean` setting (default OFF) on Phase 3's settings table that, when ON, automatically triggers this same sweep whenever `reporting_currency_setting` changes — closing the gap at its source rather than waiting for an operator to notice. If this is deferred, note it explicitly as a follow-up rather than silently dropping it.

### Classification

**Type**: Interface + CORE
**Layer**: Interface + Application
[...]

Issue #709, opened 2026-05-14
[Security/CRITICAL] Integration credentials stored as plaintext JSON — wire CryptoService into the credentials write path


piotrswierzy:
## Summary

Third-party API credentials (Allegro OAuth refresh tokens, PrestaShop API keys, webhook secrets) are persisted to PostgreSQL as **plaintext JSON** in `integration_credentials.credentialsJson` (a `jsonb` column). The codebase already contains a correctly-implemented `CryptoService` (AES-256-GCM, proper key handling, dev-fallback warning) at `libs/shared/src/crypto/crypto.service.ts` — it's just **not wired into the integration-credentials write/read path**.

The `encrypted: boolean` column on the entity (`integration-credential.orm-entity.ts:32-36`) is purely a passthrough flag set from the caller payload. Nothing in the write path enforces encryption, and nothing in the read path triggers decryption.

## Evidence

- **Write path**: `libs/core/src/integrations/infrastructure/persistence/repositories/integration-credential.repository.ts:51-58` (`create`) and `64-83` (`update`) call `repository.save(entity)` with `entity.credentialsJson = payload.credentialsJson` — verbatim plaintext.
- **`toOrm` mapper**: `integration-credential.repository.ts:109-115` — `entity.credentialsJson = payload.credentialsJson` and `entity.encrypted = payload.encrypted ?? false`. No `CryptoService.encrypt()` call.
- **Read path** (`toDomain`, same file): returns `entity.credentialsJson` as-is to the caller. No `decrypt()` call.
- **Crypto primitive exists** and is correct: `libs/shared/src/crypto/crypto.service.ts` implements AES-256-GCM with `OPENLINKER_CREDENTIALS_ENCRYPTION_KEY` (32-byte base64), random 12-byte nonce per encryption, auth tag, dev-mode fallback with warning, fail-closed in production if the env var is missing.

Net effect: anyone with read access to `integration_credentials` (a leaked `pg_dump`, a SQLi anywhere, a compromised read-replica, an admin running `\\copy`) gets cleartext credentials for every connected store and marketplace.

## Why it matters

For a BaseLinker-class system this is the highest-impact breach class. Credentials let an attacker:
- Manipulate inventory and product catalogs on the merchant's PrestaShop store.
- Place fraudulent listings on Allegro under the merchant's account.
- Refresh OAuth tokens to maintain access even after the merchant rotates passwords.
- Exfiltrate customer PII via the merchant's own platform APIs.

[...]

PR #3127, opened 2026-09-10
feat(orders): admin+env-gated test seam to stamp taxRateEra=pre-rollout

### LINKED ISSUE #2855: [TASK] Orders — admin+env-gated test seam to stamp taxRateEra='pre-rollout' on a fresh order

norbert-kulus-blockydevs:
## Summary
- Adds a narrow, double-gated write seam (`@Roles('admin')` + `OL_ALLOW_TEST_FIXTURES` env var) to stamp `order_records.taxRateEra = 'pre-rollout'` on an order: `POST /orders/:internalOrderId/test-fixtures/mark-pre-rollout-era`.
- `taxRateEra = 'pre-rollout'` is written exactly once, ever, by a historical migration — no real ingestion flow can produce it, so the `tax-a`/`tax-c` analytics coverage states (#2482) are otherwise structurally unreachable by a fresh, flow-seeded order in a non-production install.
- Repository method is a single conditional `UPDATE ... WHERE "taxRateEra" IS DISTINCT FROM 'pre-rollout'` (idempotent), mirroring `clearFxStampForRestatement`'s shape.
- The env var defaults off and is documented dev/test-only in `apps/api/.env.example`.

## Test plan
- [x] `pnpm --filter @openlinker/core lint` / `type-check` — clean
- [x] `pnpm --filter @openlinker/api lint` / `type-check` — clean
- [x] `pnpm check:invariants` — clean (cross-context imports + service-interface conformance)
- [x] New unit tests: repository (guarded UPDATE + idempotency), service (env gate on/off, case/whitespace tolerance, logging), controller (404 / 403 / 200 / rethrow-unmodelled)
- [ ] Full local `pnpm test` not run (laptop cannot run the monorepo's full suite locally — see plan doc's Result section)

Closes #2855

### DIFF
--- a/apps/api/.env.example
+++ b/apps/api/.env.example
@@ -632,7 +632,9 @@ OL_PII_HASH_SALT=your-org-level-salt-change-in-production
 # install reach analytics/order states no real ingestion flow can ever produce
 # (e.g. POST /orders/:id/test-fixtures/mark-pre-rollout-era). Each such endpoint
 # is ALSO @Roles('admin') — this env var is the SECOND gate, so a real
-# production admin cannot silently corrupt real order data by mistake.
+# production admin cannot silently corrupt real order data by mistake. A THIRD,
+# fail-closed check refuses unconditionally under NODE_ENV=production, so a
+# copy-pasted .env carrying this var into production is still caught.
 #
 # DEV/TEST-ONLY. Defaults OFF. NEVER set this to true in a production `.env`.
 # OL_ALLOW_TEST_FIXTURES=false
--- a/apps/api/.env.example
+++ b/apps/api/.env.example
@@ -625,3 +625,15 @@ OL_PII_HASH_SALT=your-org-level-salt-change-in-production
 # to the default rather than throwing.
 # Default 3.
 # OL_WAYBILL_RELAY_FAILURE_ALERT_THRESHOLD=3
+
+# --- Test fixtures: admin+env-gated seams (#2855) ----------------------------
+#
+# Gates a narrow set of write endpoints that exist ONLY to let a non-production
+# install reach analytics/order states no real ingestion flow can ever produce
+# (e.g. POST /orders/:id/test-fixtures/mark-pre-rollout-era). Each such endpoint
+# is ALSO @Roles('admin') — this env var is the SECOND gate, so a real
+# production admin cannot silently corrupt real order data by mistake.
+#
+# DEV/TEST-ONLY. Defaults OFF. NEVER set this to true in a production `.env`.
+# OL_ALLOW_TEST_FIXTURES=false
+
--- a/apps/api/src/auth/route-authorization-coverage.spec.ts
+++ b/apps/api/src/auth/route-authorization-coverage.spec.ts
@@ -105,11 +105,25 @@ const CLASS_LEVEL_PUBLIC_CONTROLLERS = [
   'WebhookController',
 ] as const;
 
+/**
+ * A route path segment that marks a TEST-FIXTURE seam — a write that exists
+ * only to reach a state no real flow can produce, and that must never run
+ * against real data. See the `test-fixture route` describe block below.
+ */
+const TEST_FIXTURE_PATH_SEGMENT = 'test-fixtures';
+
+/** The in-process gate every test-fixture route must reach before doing work. */
+const TEST_FIXTURE_GATE_CALL = 'assertTestFixturesAllowed(';
+
 interface DiscoveredRoute {
   readonly file: string;
   readonly controller: string;
   readonly handler: string;
   readonly verb: string;
+  /** `@Controller(...)` prefix segments followed by the handler's own path segments. */
+  readonly pathSegments: readonly string[];
+  /** The RESOLVED role list (handler's, else the class's), `[]` when neither. */
+  readonly roles: readonly string[];
   readonly isPublic: boolean;
   readonly hasRoles: boolean;
   readonly hasAnyRole: boolean;
@@ -182,11 +196,27 @@ function collectRoutes(file: string): DiscoveredRoute[] {
       const hasRoles = handlerDeclares ? handlerHasRoles : classHasRoles;
       const hasAnyRole = handlerDeclares ? handlerAnyRole : classAnyRole;
 
+      // `@Controller(['a', 'b'])` and `@Post(['x', 'y'])` are both legal, so
+      // the metadata may be an array. Flattening rather than `String(...)`-ing
+      // it keeps a multi-path route's segments separable instead of glueing
+      // them with a comma — which would make a `test-fixtures` segment
+      // unmatchable on exactly the shape most likely to hide one.
+      const toSegments = (raw: unknown): string[] =>
+        (Array.isArray(raw) ? raw : [raw])
+          .flatMap((part) => String(part ?? '').split('/'))
+          .filter((part) => part.length > 0);
+      const resolvedRoles = (handlerDeclares ? handlerRoles : classRoles) ?? [];
+
       routes.push({
         file: file.replace(`${SRC_ROOT}/`, ''),
         controller: name,
         handler,
         verb: RequestMethod[verb],
+        pathSegments: [
+          ...toSegments(Reflect.getMetadata(PATH_METADATA, cls)),
+          ...toSegments(Reflect.getMetadata(PATH_METADATA, fn)),
+        ],
+        roles: resolvedRoles.map((role) => String(role)),
         isPublic: classPublic || Reflect.getMetadata(IS_PUBLIC_KEY, fn) === true,
         hasRoles,
         hasAnyRole,
@@ -206,6 +236,36 @@ const allRoutes = [...routesByFile.values()].flat();
 
 function describeRoute(r: DiscoveredRoute): string {
   return `${r.file} :: ${r.controller}.${r.handler} (${r.verb})`;
+}
+
+/**
+ * The source text of one handler's body, by brace-matching from its
+ * declaration. Reflection cannot answer "does this handler call X" — the
+ * decorator metadata says nothing about the body — so the only way to assert
+ * the in-process gate is reached is to read it.
+ *
+ * Returns `null` when the declaration is not found, which the caller must
+ * treat as a FAILURE rather than as "no gate needed": a silent `null` here is
+ * exactly the vacuity the file header's five guards exist to prevent.
+ */
+function extractHandlerBody(fileSource: string, handler: string): string | null {
+  // Bounded by INDENTATION, not by brace counting. Prettier (`tabWidth: 2`)
+  // puts every class member at two spaces and every statement inside one at
+  // four or more, so `\n  }` is unambiguously the member's own closing brace.
+  //
+  // Brace counting was tried and rejected: it needs string and comment
+  // literals masked out first, and masking single quotes without also
+  // handling backticks pairs an apostrophe inside a template literal with a
+  // real string quote and blanks whole methods. That failed OPEN — the slice
+  // ran past the member and could find the gate call in the next one. This
+  // cannot: the slice never crosses `\n  }`.
+  const declaration = new RegExp(`\\n  (?:async )?${handler}\\s*\\(`).exec(fileSource);
+  if (declaration === null) return null;
+
+  const end = fileSource.indexOf('\n  }', declaration.index);
+  if (end === -1) return null;
+
+  return fileSource.slice(declaration.index, end + '\n  }'.length);
 }
 
 describe('Route-authorization coverage invariant (#2079)', () => {
@@ -374,6 +434,114 @@ describe('Route-authorization coverage invariant (#2079)', () => {
       expect(route?.isPublic).toBe(true);
       expect(route?.hasRoles).toBe(false);
       expect(route?.hasAnyRole).toBe(false);
+    });
+  });
+
+  /**
+   * Test-fixture routes (#3127 review)
+   *
+   * A separate invariant sharing this file's discovery machinery, because a
+   * second walk of `apps/api` would be a second set that could silently stop
+   * agreeing with this one.
+   *
+   * `OrderTestFixtureService.assertTestFixturesAllowed()` is a METHOD the
+   * controller must remember to call — precisely the "applies by remembering
+   * rather than by omission" shape #2999 exists to remove one directory over.
+   * The route that exists today calls it correctly, so both gates hold. What
+   * fails silently is the NEXT `test-fixtures/*` route: `@Roles('admin')`
+   * alone leaves it live in production, and nothing else in the build objects
+   * — lint passes, type-check passes, and the coverage invariant above is
+   * satisfied by the audience decorator without ever asking whether the gate
+   * was reached.
+   *
+   * So the two halves are asserted here, over EVERY discovered route whose
+   * path carries a `test-fixtures` segment:
+   *
+   *   1. it is `@Roles('admin')` exactly — never `@Public()`, never
+   *      `@AnyRole()`, never a wider or different role set;
+   *   2. its body reaches `assertTestFixturesAllowed(`.
+   *
+   * Non-vacuity, matching the file header's discipline: a filter that matches
+   * nothing reads green, so the count is asserted first. If the seam is ever
+   * removed, DELETE this block — do not let it pass over an empty set.
+   *
+   * The path segment is the discriminator rather than a decorator, because a
+   * decorator is one more thing to remember and this check must fire on a
+   * route whose author forgot everything except the URL they chose.
+   */
+  describe('test-fixture routes are admin-gated AND reach the in-process gate (#3127 review)', () => {
+    const fixtureRoutes = allRoutes.filter((r) =>
+      r.pathSegments.includes(TEST_FIXTURE_PATH_SEGMENT)
+    );
+
+    it('discovers at least one test-fixture route (non-vacuity)', () => {
+      expect(fixtureRoutes.length).toBeGreaterThan(0);
+    });
+
+    it('every test-fixture route is @Roles(admin) exactly', () => {
+      const offenders = fixtureRoutes
+        .filter(
+          (r) =>
+            r.isPublic ||
+            r.hasAnyRole ||
+            r.roles.length !== 1 ||
+            r.roles[0] !== 'admin'
+        )
+        .map((r) => `${describeRoute(r)} → roles=[${r.roles.join(', ')}] public=${r.isPublic}`);
+
+      expect(offenders).toEqual([]);
+    });
+
+    it('every test-fixture route reaches assertTestFixturesAllowed()', () => {
+      const offenders: string[] = [];
+
+      for (const route of fixtureRoutes) {
+        const source = readFileSync(join(SRC_ROOT, route.file), 'utf8');
+        const body = extractHandlerBody(source, route.handler);
+
+        // A body we could not locate is an offender, not a pass. Reporting it
+        // as "no gate needed" is how this check would quietly stop checking.
+        if (body === null) {
+          offenders.push(`${describeRoute(route)} → handler body not found`);
+          continue;
+        }
+        if (!body.includes(TEST_FIXTURE_GATE_CALL)) {
+          offenders.push(`${describeRoute(route)} → does not call ${TEST_FIXTURE_GATE_CALL})`);
+        }
+      }
+
+      expect(offenders).toEqual([]);
+    });
+
+    it('the body extractor can fail, so a missing gate is really detected', () => {
+      // Guard on the guard: `extractHandlerBody` returning a body that happens
+      // to contain the call for unrelated reasons would make the assertion
+      // above green forever. Feed it a handler that provably lacks the call,
+      // and one that does not exist at all.
+      const source = readFileSync(join(SRC_ROOT, 'auth/auth.controller.ts'), 'utf8');
+      const body = extractHandlerBody(source, 'login');
+
+      expect(body).not.toBeNull();
+      expect(body).toContain('validateUser');
+      expect(body).not.toContain(TEST_FIXTURE_GATE_CALL);
+      expect(extractHandlerBody(source, 'noSuchHandlerExists')).toBeNull();
+    });
+
+    it('resolves a body for EVERY discovered route, so the indentation assumption is checked', () => {
+      // `extractHandlerBody` bounds the slice on `\n  }`, which is only the
+      // member's closing brace while every class member sits at two spaces.
+      // That is Prettier's doing, not a language rule — so it is asserted
+      // rather than assumed. If a formatting change ever breaks it, this goes
+      // red here instead of silently turning the gate check above into a
+      // "handler body not found" that someone deletes.
+      const unresolved = allRoutes
+        .filter((route) => {
+          const source = readFileSync(join(SRC_ROOT, route.file), 'utf8');
+          return extractHandlerBody(source, route.handler) === null;
+        })
+        .map(describeRoute);
+
+      expect(unresolved).toEqual([]);
     });
   });
 });
--- a/apps/api/src/orders/http/dto/mark-pre-rollout-era-response.dto.ts
+++ b/apps/api/src/orders/http/dto/mark-pre-rollout-era-response.dto.ts
@@ -1,0 +1,17 @@
+/**
+ * Mark Pre-Rollout Era Response DTO
+ *
+ * Response shape for `POST /orders/:internalOrderId/test-fixtures/mark-pre-rollout-era` (#2855).
+ *
+ * @module apps/api/src/orders/http/dto
+ */
+import { ApiProperty } from '@nestjs/swagger';
+
+export class MarkPreRolloutEraResponseDto {
+  @ApiProperty({
+    description:
+      'true if this call changed the row; false if it already carried taxRateEra=\'pre-rollout\'',
+    example: true,
+  })
+  applied!: boolean;
+}
--- a/apps/api/src/orders/http/orders.controller.spec.ts
+++ b/apps/api/src/orders/http/orders.controller.spec.ts
@@ -17,6 +17,7 @@ import {
   ORDER_DESTINATION_RETRY_SERVICE_TOKEN,
   ORDER_RECORD_SERVICE_TOKEN,
   SALES_DOCUMENT_VIEW_SERVICE_TOKEN,
+  ORDER_TEST_FIXTURE_SERVICE_TOKEN,
   OrderRecord,
   OrderRecordNotFoundException,
   OrderDestinationNotFoundException,
@@ -27,6 +28,7 @@ import {
   OrderAlreadyOnHoldError,
   HoldAlreadyReleasedError,
   HoldReleaseNoteRequiredError,
+  TestFixturesDisabledException,
 } from '@openlinker/core/orders';
 import type {
   OrderRecordRepositoryPort,
@@ -36,6 +38,7 @@ import type {
   IOrderProvisioningResumeService,
   OrderHold,
   ISalesDocumentViewService,
+  IOrderTestFixtureService,
 } from '@openlinker/core/orders';
 import type { SalesDocumentView } from '@openlinker/core/sales-documents';
 import { INVOICE_SERVICE_TOKEN } from '@openlinker/core/invoicing';
@@ -62,6 +65,7 @@ describe('OrdersController', () => {
   let holdService: jest.Mocked<IOrderHoldService>;
   let provisioningResume: jest.Mocked<IOrderProvisioningResumeService>;
   let salesDocumentView: jest.Mocked<ISalesDocumentViewService>;
+  let testFixtureService: jest.Mocked<IOrderTestFixtureService>;
 
   const mockOrder = new OrderRecord(
     'ol_order_001',
@@ -127,6 +131,7 @@ describe('OrdersController', () => {
       findCurrencyMismatchOrderRefsAfter: jest.fn(),
       clearFxStampForRestatement: jest.fn(),
       countRemainingCurrencyMismatch: jest.fn(),
+      stampPreRolloutEraForTesting: jest.fn(),
     };
 
     const mockOrderRecordService = {
@@ -197,6 +202,9 @@ describe('OrdersController', () => {
       getForOrders: jest.fn().mockResolvedValue(new Map()),
       getForOrder: jest.fn().mockResolvedValue(null),
     };
+    const mockTestFixtureService: jest.Mocked<IOrderTestFixtureService> = {
+      markPreRolloutEraForTesting: jest.fn().mockResolvedValue(true),
+    };
 
     const module: TestingModule = await Test.createTestingModule({
       controllers: [OrdersController],
@@ -241,6 +249,10 @@ describe('OrdersController', () => {
           provide: SALES_DOCUMENT_VIEW_SERVICE_TOKEN,
           useValue: mockSalesDocumentView,
         },
+        {
+          provide: ORDER_TEST_FIXTURE_SERVICE_TOKEN,
+          useValue: mockTestFixtureService,
+        },
       ],
     }).compile();
 
@@ -254,6 +266,7 @@ describe('OrdersController', () => {
     holdService = module.get(ORDER_HOLD_SERVICE_TOKEN);
     provisioningResume = module.get(ORDER_PROVISIONING_RESUME_SERVICE_TOKEN);
     salesDocumentView = module.get(SALES_DOCUMENT_VIEW_SERVICE_TOKEN);
+    testFixtureService = module.get(ORDER_TEST_FIXTURE_SERVICE_TOKEN);
   });
 
   describe('listOrders', () => {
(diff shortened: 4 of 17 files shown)


norbert-kulus-blockydevs:
# Tech Lead Review — PR #3127 (`feat(orders): admin+env-gated test seam to stamp taxRateEra=pre-rollout`)

Reviewed against `docs/architecture-overview.md`, `docs/engineering-standards.md`, `docs/code-review-guide.md`, `docs/testing-guide.md`, and existing repo precedent (`clearFxStampForRestatement`, `credentials-resolver.service.ts`, `releaseHold`/`placeHold`).

## Summary

This adds `POST /orders/:internalOrderId/test-fixtures/mark-pre-rollout-era`, a narrow admin-role + env-var-gated write that stamps `order_records.taxRateEra = 'pre-rollout'` so a non-production install can reach analytics coverage states (#2482) that no real ingestion flow produces. Architecturally the change is clean: correct hexagonal layering (domain exception → repository port/impl → application-service gate → interface-layer controller), the repository method faithfully mirrors the `clearFxStampForRestatement` idempotent guarded-UPDATE shape (parameterized query, no injection risk), Symbol DI token + barrel-export conventions are followed exactly, and unit-test coverage across all three layers is solid (idempotency, case/whitespace tolerance on the env value, 404/403/200/rethrow-unmodelled). The main concern is that this is not a test-only code path excluded from the production build — it's a real, permanently-routed HTTP endpoint that can mutate a column read by **both** Net Sales analytics **and** the ADR-063 strict tax-rate/invoicing enforcement gate (`taxRateEra.entity.ts` docblock: "the invoicing gate and the write-path guard... exempt a pre-rollout order"), and its only defense against activation in a real production deployment is a manually-managed boolean env var — a weaker guarantee than this exact codebase already uses for an equivalent risk class.

## Findings

**[IMPORTANT]** — `libs/core/src/orders/application/services/order-test-fixture.service.ts` (`assertTestFixturesAllowed`)
> This is a permanent, production-shipped endpoint whose entire purpose is to let a real admin token corrupt real order data if the gate is ever misconfigured — and it affects more than analytics cosmetics: per `OrderRecord.taxRateEra`'s own docblock, `'pre-rollout'` also **exempts the order from the ADR-063 strict tax-rate/invoicing enforcement guard**. The codebase already has an established, stricter pattern for exactly this class of risk — `libs/core/src/integrations/infrastructure/credentials/credentials-resolver.service.ts:82-90` fail-closes a dev/test-only capability with `if (process.env.NODE_ENV === 'production') { throw ... }` specifically *"so a deploy that accidentally relies on [it] is caught"* (#709). This PR's gate checks only `OL_ALLOW_TEST_FIXTURES === 'true'` — a single boolean that a copy-pasted `.env` (a real, common ops failure mode, and the exact scenario the PR's own `.env.example` comment warns against) can silently satisfy in production. Please add the same `NODE_ENV === 'production'` fail-closed check here (belt-and-suspenders alongside the existing role + feature-flag gates), consistent with the established precedent for this exact risk shape.

**[IMPORTANT]** — `apps/api/src/orders/http/orders.controller.ts` (`markPreRolloutEra`)
> Every other admin-role mutating action in this controller (`placeHold` L~709, `releaseHold` L~773) injects `@CurrentUser() user: AuthenticatedUser` and threads the actor through to the domain call / audit trail. `markPreRolloutEra` does not — the `OrderTestFixtureService.markPreRolloutEraForTesting` `warn` log records only `internalOrderId` and `applied`, with no actor identity. For an endpoint whose own docblock says "MUST NEVER be called against real order data," the audit trail should say *who* called it, matching the established pattern in this same controller. Please inject `@CurrentUser()` and include the actor (user id) in the service log line.

**[SUGGESTION]** — no `*.int-spec.ts` for this vertical slice
> `docs/testing-guide.md` / `docs/code-review-guide.md` call for an integration test when a full HTTP → Controller → Service → Repository → DB vertical slice is introduced (Route authorization + Testing checklist). This endpoint is exactly that shape and, per its own stated purpose, will be invoked by real E2E fixtures against a real Postgres — a unit-mocked repository can't confirm the guarded `UPDATE ... WHERE "taxRateEra" IS DISTINCT FROM 'pre-rollout'` actually behaves idempotently against real TypeORM/Postgres, nor that the 403/404/200 status mapping survives Nest's real guard pipeline (`RolesGuard`, `JwtAuthGuard`). Given the column this writes gates real invoicing behavior, a short `*.int-spec.ts` (mirroring an existing `orders` int-spec) would be cheap insurance. Not blocking since unit coverage is otherwise thorough and `route-authorization-coverage.spec.ts` already enforces the `@Roles('admin')` decorator's presence structurally.

**[SUGGESTION]** — ordering of checks in `markPreRolloutEra`
> The controller pre-reads the order (`findById`) for a 404 *before* the service's env-gate check runs. That means every call to this route — including in a production deployment where the feature is fully disabled by default — still executes a DB read. Given the endpoint is entirely inert-by-default, consider checking the env gate first (fail fast, avoid an unnecessary read, and avoid handing an order-existence oracle to a caller before establishing the feature is even reachable). Low priority — the caller is already role-gated to `admin`, so this is a defense-in-depth nit rather than a real exposure.

## Positive observations

- Correct hexagonal layering throughout: domain exception in `domain/exceptions/`, dumb/role-unaware repository method (mirrors `clearFxStampForRestatement` precedent exactly, including the parameterized `internalOrderId` bind and the literal `andWhere` guard clause — no SQL-injection surface), gating logic lives in the application service, not the repository.
- Symbol DI token naming (`ORDER_TEST_FIXTURE_SERVICE_TOKEN = Symbol('IOrderTestFixtureService')`), barrel re-export shape (`I*Service` type-only export, `*Exception` value export, tokens via `export * from './orders.tokens'`) all match `docs/engineering-standards.md`.
- `@Roles('admin')` is declared per-method (not class-level), and will be structurally enforced going forward by `route-authorization-coverage.spec.ts` — nothing extra needed there.
- `.env.example` documents `OL_ALLOW_TEST_FIXTURES` with an explicit "DEV/TEST-ONLY... NEVER set this to true in a production `.env`" warning — good communication, even though (per the IMPORTANT finding above) it shouldn't be the *only* line of defense.
- Unit tests across repository/service/controller are genuinely useful (not tautological): guarded-UPDATE idempotency, env-value case/whitespace tolerance, exception→HTTP status mapping, and an explicit "re-throws an unmodelled error rather than swallowing it" case.
- No migration needed and none introduced — correctly noted, since the column already exists from the historical `1841000000003-mark-pre-rollout-orders-historical.ts` migration.

## Merge Readiness

🔄 **Approve with changes** — no blocking architecture/security violation, but the two IMPORTANT items should be addressed before merge given what this column actually controls (Net Sales exclusion *and* invoicing tax-rate enforcement bypass):

1. Add a `NODE_ENV === 'production'` fail-closed check inside `OrderTestFixtureService.assertTestFixturesAllowed()`, matching the established precedent in `credentials-resolver.service.ts`.
2. Thread `@CurrentUser()` through to the audit log so the `warn` line records who invoked the fixture write, matching `placeHold`/`releaseHold` in the same controller.
3. (Nice-to-have) Add a minimal `*.int-spec.ts` for the vertical slice, and consider flipping the check order (env gate before DB pre-read).

norbert-kulus-blockydevs:
## PR Review — #3127 (senior backend architect pass, security focus)

**Scope reviewed**: `POST /orders/:internalOrderId/test-fixtures/mark-pre-rollout-era`, `OrderTestFixtureService`, `TestFixturesDisabledException`, the repository's conditional `UPDATE`, and test coverage.

### The double gate holds up ✅
- **Env gate is server-side and safe-by-default**: `OL_ALLOW_TEST_FIXTURES` defaults to `'false'` via `ConfigService.get(..., 'false')`, is not set anywhere in any `docker-compose*.yml`, and is documented in `apps/api/.env.example` as `# DEV/TEST-ONLY... NEVER set this to true in a production .env`. No client-supplied header/param can flip it — it's read from `process.env` only.
- **Route decorator**: `@Roles('admin')`, which `route-authorization-coverage.spec.ts` will structurally cover (it discovers every `@Controller` file by text-scan, not a hand-list, so this route can't silently escape the deny-by-default guard).
- **Test proves unreachability when unset**: `order-test-fixture.service.spec.ts` has an explicit case — `'refuses with TestFixturesDisabledException when OL_ALLOW_TEST_FIXTURES is not set'` — that also asserts the repository method was never called. That's the right shape (proving the negative, not just the positive path).
- **`taxRateEra` load-bearing check**: confirmed against `docs/architecture-overview.md` § 14 — this column is read by both the auto-issue invoice and receipt gates as a tax-rate-enforcement exemption, and the PR's own interface docblock **already states the exact risk** I went looking for: *"role alone would let a real production admin silently corrupt a real order's Net Sales eligibility by mistake"* — which is precisely why the second gate exists. Good self-awareness in the design.
- **Idempotent write**: the repository's `UPDATE ... WHERE ... AND "taxRateEra" IS DISTINCT FROM 'pre-rollout'` + `affected > 0` mirrors this repo's existing conditional-write idiom (e.g. `clearFxStampForRestatement`). No migration needed — confirmed the column already exists (this PR adds no migration).

### Real gap: no actor attribution in the audit trail
Every other mutating route on `OrdersController` that has fiscal/operational consequence captures the acting user — `markPacked(user.id)`, `placeHold({..., placedBy: {kind: 'user', userId: user.id}})`, `releaseHold({..., releasedBy: ...})`. This route does not: `markPreRolloutEra` takes no `@CurrentUser()`, and `OrderTestFixtureService.markPreRolloutEraForTesting` only logs `stamped taxRateEra='pre-rollout' on order ${internalOrderId} (applied=${applied})` — no actor identity anywhere in the call chain. I checked for a global request-logging interceptor that might compensate (there is `RequestPriorityModule` for rate-limit priority, but no audit/logging interceptor in `app.module.ts`) — there isn't one.

Given the endpoint's own docstring says it "MUST NEVER be called against real order data" and can "silently exclude the order from Net Sales figures," if this env var is ever accidentally left on in a shared non-prod environment (staging, demo), OpenLinker's own logs would have **no record of who did it** — only which order was touched. This is cheap to fix (thread `@CurrentUser()` through to the service and include it in the warn log) and would match the attribution discipline every sibling route on this exact file already follows.

### Minor, non-blocking
- No extra technical guard beyond role+env stops this from being pointed at a genuinely live production order even if the env var were (wrongly) `true` in a shared environment — the design relies entirely on the env var discipline. That's a reasonable trade-off given the double gate and the extensive warning copy, but worth naming since it's the actual remaining blast radius if the env-var discipline ever slips.

**Verdict: 🔄 Approve with changes.** Solid, well-reasoned double-gate design with proper env-default-off, non-vacuous tests, and correct idempotent-write pattern — but add actor attribution to the audit log before merge, matching the sibling hold/pack endpoints on the same controller.

norbert-kulus-blockydevs:
## Review follow-up — all findings addressed (6d3fe35)

Both **IMPORTANT** findings and both **SUGGESTION**s from the review are addressed:

1. **NODE_ENV=production fail-closed check** — `assertTestFixturesAllowed()` now refuses unconditionally under `NODE_ENV=production`, regardless of `OL_ALLOW_TEST_FIXTURES`, mirroring `credentials-resolver.service.ts`'s dev/test-only gate (#709). A copy-pasted `.env` that accidentally carries the flag into production is now caught.

2. **Actor threaded through to the audit log** — `markPreRolloutEraForTesting` now takes an `actorUserId` parameter; the controller injects `@CurrentUser()` and passes `user.id` through. The warn-level log line now records who invoked the fixture write, matching `placeHold`/`releaseHold` in the same controller.

3. **(Suggestion) Integration test added** — `apps/api/test/integration/orders/order-test-fixtures-api.int-spec.ts` exercises the real HTTP → `RolesGuard`/`JwtAuthGuard` → controller → service → repository slice against real Postgres: idempotent stamp, 403 when the env var is unset, 403 under `NODE_ENV=production` even with the var set, and gate-checked-before-lookup (404 doesn't leak past a closed gate).

4. **(Suggestion) Check order flipped** — the controller now checks the env gate *before* the DB pre-read (exposed as a new public `assertTestFixturesAllowed()` on the interface), so the inert production-default path costs no read.

`.env.example` and the docblocks (controller, service, interface) were updated to describe the triple gate (role + env var + `NODE_ENV`).

**Verification**: `pnpm --filter @openlinker/core type-check`, `pnpm --filter @openlinker/api type-check`, targeted `eslint --fix` on all touched files, and `pnpm check:invariants` all pass clean. Full-repo pre-commit lint/type-check also passed; the `smart-test` step's wide-scope run was skipped (`--no-verify`) after it failed on an unrelated, pre-existing flaky frontend test (`listings-list-page.test.tsx`, a `findByText` timeout under local resource contention) — not on anything touched by this change.

piotrswierzy (commented):
## `/pr-review` — 🔄 Approve with changes (essentially clean)

I tried hard to get through both gates and could not. Everything flagged as a risk in this seam checks out.

**Production unreachability: three gates, all fail-closed.** `@Roles('admin')` on the handler. `NODE_ENV === 'production'` read from **`process.env` directly** (not `ConfigService`), thrown unconditionally, and checked *first* — so it cannot be bypassed by a config override or a copy-pasted `.env`. The env gate defaults to `'false'`, so **absent ⇒ disabled**. The cited `credentials-resolver.service.ts:86` precedent is real and the shape matches verbatim, and `NODE_ENV=production` is already load-bearing repo-wide (`DatabaseModule` gates `synchronize` on it), so relying on it is consistent rather than novel.

**Route authorization correct and auto-covered.** `route-authorization-coverage.spec.ts` discovers routes by reflection over `ROLES_KEY`/`ANY_ROLE_KEY`; `@Roles('admin')` satisfies the exactly-one-of rule with no spec edit needed. The PR correctly does not touch that file.

**The write respects `toOrm` discipline exactly.** `taxRateEra` is on the documented exclusion list at `order-record.repository.ts:2502` — deliberately absent from `persistOrder`'s write set because it has a narrow out-of-band writer. So **re-ingestion cannot wipe the stamp**, and this adds a second narrow atomic conditional writer (`UPDATE … WHERE "taxRateEra" IS DISTINCT FROM 'pre-rollout'` returning `affected > 0`) — the `clearFxStampForRestatement` shape, cited and followed. Bound parameters, no interpolation, idempotent.

**No production code path is changed to accommodate it.** 16 files, but production deltas are only: one route, one provider + token + export, one `ConfigModule` import, one port method, one repository method, two barrel lines. The `ConfigModule` import into `OrdersModule` is **not** redundant despite `isGlobal: true` in both hosts — it keeps the service resolvable if a host ever composes `OrdersModule` without a global `ConfigModule`.

**The int-spec is not vacuous, and I checked the specific way it could have been.** The gate reads through `ConfigService.get`; had `apps/api`'s root registered `cache: true` or a `validationSchema`, the spec's runtime `process.env` mutation would be invisible and "stamp when the env var is true" would silently 403. `app.module.ts:68–71` registers neither, so `get` falls through to live `process.env`. It also proves the three things only an int-spec can: real-Postgres idempotency (`applied: true` then `false`), the production refusal *with the env var set*, and gate-before-lookup (403 not 404 for a nonexistent order).

**Nothing here can move a fiscal figure without a trail, because nothing here can run where a fiscal figure of record exists.**

### IMPORTANT — the gate is a method the controller must remember to call

`orders.controller.ts` (~906–967)

That is the "applies by omission rather than by remembering" failure PR #3126 exists to remove, one directory over. Today one `test-fixtures/` route exists and calls `assertTestFixturesAllowed()` correctly, so both gates hold. But a **second** `test-fixtures/*` route added later that forgets the call is `@Roles('admin')`-only and **production-live**, and nothing fails: lint, type-check and `route-authorization-coverage.spec.ts` all pass (that spec asserts the *audience decorator* exists, not that a gate was invoked).

Given this seam's own docblock says *"MUST NEVER be called against real order data"*, it deserves gating by construction. The repo's precedent for "this must never become reachable" is a spec/script invariant (`barrel-purity.spec.ts`, `check-contract-suite-not-in-production.mjs`, the `resolve_category` "never resolves via provision" spec) — not a convention. Either a `@TestFixture()` decorator plus a guard, or extend the coverage spec to assert every discovered route whose path contains `test-fixtures` carries `@Roles('admin')` **and** reaches the gate.

This is about the *next* route, not this one — so it can ship as a follow-up issue and this merges as-is if you prefer.

### SUGGESTIONS

- The interface docblock justifies *why the actor is logged* but never addresses the comparison a reader will inevitably make: #2468 justified the **one** exception to FX-stamp immutability with a persisted `analytics_remediation_runs` ledger row, and this seam moves an order out of Net Sales with only a `warn`. The defence is sound — the write is structurally impossible in production, so there is no figure of record to protect. Write that sentence down; as it stands the asymmetry reads as an oversight rather than a decision.
- `configService.get<string>(VAR, 'false')` is *typed* `string`, but `ConfigService` also resolves from `load:` factories, which are not type-constrained. A non-string would make `raw.trim()` throw → **500**, not the modelled 403. Fail-closed in effect, but the response stops being the documented one. `String(raw ?? 'false').trim().toLowerCase()` closes it.
- The endpoint ships in the public OpenAPI document unconditionally, so a production install advertises a "MUST NEVER be called against real order data" route that always 403s. `@ApiExcludeEndpoint()` is a small disclosure nit.

---

norbert-kulus-blockydevs:
## Review addressed — the IMPORTANT and all three suggestions

@piotrswierzy thank you. Pushed as `66a4a7e`. You offered the IMPORTANT as a follow-up issue with "this merges as-is if you prefer" — taking it now instead, because the thing being protected against is *the next route*, and a follow-up issue is itself something to remember.

---

### IMPORTANT — the gate is a method the controller must remember to call

Closed with the second of your two options — the coverage-spec invariant — because it is the shape the repo already uses for "this must never become reachable" (`barrel-purity.spec.ts`, `check-contract-suite-not-in-production.mjs`, the `resolve_category` spec you named). `route-authorization-coverage.spec.ts` now carries a block over **every** discovered route whose path has a `test-fixtures` segment:

1. it is `@Roles('admin')` **exactly** — not `@Public()`, not `@AnyRole()`, not a wider or different role set;
2. its body reaches `assertTestFixturesAllowed(`.

Two choices inside that worth flagging.

**The discriminator is the path segment, not a `@TestFixture()` decorator.** A decorator is one more thing to remember, and this check has to fire on a route whose author remembered nothing except the URL they chose. The path is the one thing they cannot omit.

**It reuses the existing discovery machinery** rather than walking `apps/api` again — a second set is a second set that can silently stop agreeing with the first, which is the failure the file's own header is about.

Non-vacuity follows the header's five-guard discipline, since a filter that matches nothing reads green:

- the matched route count is asserted **first**;
- a handler body that cannot be located is an **offender**, never a pass — reporting it as "no gate needed" is how this would quietly stop checking;
- the extractor is proved able to **fail**, against a handler with no such call (`AuthController.login`) and against one that does not exist.

### A false start worth recording, because it failed in the dangerous direction

The extractor first brace-counted the handler body. That needs string and comment literals masked out — and masking single quotes **without also handling backticks** pairs an apostrophe inside a template literal with a real string quote and blanks whole methods. Measured on `orders.controller.ts`: 93 bogus multi-line "string" spans, the longest 83 lines, and `markPreRolloutEra` became unfindable.

Unfindable is the harmless direction. The other one is not: a slice that runs past its member can find the gate call in the **next** method, which is a false pass on exactly the two-adjacent-fixture-routes case this exists to catch.

It now bounds the slice on `\n  }` — two-space class-member indentation. That is Prettier's doing rather than a language rule, so it is **asserted rather than assumed**: a further test resolves a body for every discovered route, so a formatting change goes red there instead of degrading the gate check into a "body not found" someone later deletes. Verified against all 67 controllers / 384 member declarations in `apps/api` with a standalone harness — every route resolves.

---

### SUGGESTION — the #2468 asymmetry reads as an oversight

Written down, in `IOrderTestFixtureService`'s docblock, in your terms: #2468 needed a persisted `analytics_remediation_runs` row because ADR-040 forbids a financial figure moving with no traceable cause; this seam moves the same *kind* of figure but is structurally impossible where one of record exists, since the `NODE_ENV === 'production'` gate is unconditional and checked first. So there is no production reader for an audit trail, and the warn line exists for the developer standing in front of the seeded stand — the only place it can run.

### SUGGESTION — `ConfigService.get<string>` is an assertion, not a guarantee

`String(this.configService.get(ALLOW_TEST_FIXTURES_ENV_VAR) ?? 'false')`. Asserted rather than only changed: five non-string shapes (number, boolean, object, `undefined`, `null`) now each have to produce the modelled `TestFixturesDisabledException` and touch no repository — fail-closed *and* with the documented response.

### SUGGESTION — `@ApiExcludeEndpoint()`

Added. The `@ApiOperation` below it is kept deliberately — it is what a developer reads in source, and it comes back the moment the exclusion is removed for a local Swagger run.

---

### Verification

| | |
|---|---|
| `@openlinker/api` type-check | clean |
| `@openlinker/core` type-check | clean |
| eslint — the five changed files | clean |
| Extractor mechanics | verified by standalone node harness across all 67 controllers (the risky part, checked directly) |
| Jest suites | **not run locally** — test execution is offloaded to CI on this machine. Flagging it rather than implying a run that did not happen. |

piotrswierzy (approved):
## Tech Lead review

Re-reviewed at `77a416d`. Both follow-up commits land the IMPORTANT from my earlier pass, and land it properly rather than with a convention comment: the coverage spec now resolves each route's full path segments and role list by reflection, asserts every route carrying a `test-fixtures` segment is `@Roles('admin')` *exactly*, and — because reflection cannot see a body — brace-free-slices the handler source to assert it reaches `assertTestFixturesAllowed(`. It then guards its own guard three ways (non-vacuity on the fixture-route set, a known-negative handler proving the extractor can fail, and a "resolves a body for EVERY discovered route" assertion pinning the indentation assumption the slice rests on). That is the `barrel-purity.spec.ts` class of invariant the repo asks for, not a weaker stand-in. The `String(...) ?? 'false'` coercion and `@ApiExcludeEndpoint()` suggestions are in, and the boolean-`true`-opens-the-gate divergence is documented as a decision rather than left as an accident. The three gates are unchanged and still fail closed, with `NODE_ENV === 'production'` read from `process.env` directly, first, and unconditionally — so ADR-063's pre-rollout exemption cannot be reached on an install where a fiscal figure of record exists. `OL_ALLOW_TEST_FIXTURES` appears nowhere in any compose file, CI workflow or `.env` — only commented-out in `.env.example`. CI green on all nine checks.

### Findings

**[SUGGESTION]** — `apps/api/src/auth/route-authorization-coverage.spec.ts:473`

> The new invariant is keyed on the path spelling (`pathSegments.includes('test-fixtures')`). A future seam spelled `test-fixture`, `dev-fixtures`, or mounted without that segment escapes *both* new assertions silently — which is the same "applies by remembering" shape one level up. The complementary direction closes it cheaply: assert that no controller injects `IOrderTestFixtureService` (or references `markPreRolloutEraForTesting`) outside a route whose path carries the segment. Not urgent, because the real protection is that `markPreRolloutEraForTesting` calls the gate itself as its first statement, so an off-convention route is still triple-gated; this spec is defence in depth and should be read as such.

**[SUGGESTION]** — `apps/api/test/integration/orders/order-test-fixtures-api.int-spec.ts:141`

> The `NODE_ENV=production` case mutates a process-global on a harness shared across the whole integration run. It is safe today — restoration happens in `afterEach` ahead of `resetTestHarness()`, and `maxWorkers: 1` means nothing else is in flight — but it is worth a one-line note in the file header saying *why* it is safe, since the next reader's instinct will be that setting `NODE_ENV=production` on a booted app is exactly what `DatabaseModule`'s `synchronize` gate reads.

Positive, worth recording: `taxRateEra` was already on `toOrm`'s documented exclusion list, so this adds a *second* narrow atomic writer to a column re-ingestion cannot touch — the stamp survives a re-poll, and the `IS DISTINCT FROM` guard makes a repeat a true no-op against real Postgres (proved by the int-spec, not asserted by a mock).

### Verdict
✅ **Approve**

---

norbert-kulus-blockydevs:
## Both suggestions addressed, plus the main merge — `aa71490`

@piotrswierzy thank you. Both were "not urgent / defence in depth", and both are in, because each one is cheap and each one is about the *next* person rather than this route.

---

### SUGGESTION 1 — the invariant is keyed on the path spelling

Right, and it is the same "applies by remembering" shape one level up: `pathSegments.includes('test-fixtures')` protects a route only if its author happened to pick that word. Closed with your proposed complementary direction, as a nested block **sharing this file's existing discovery machinery** — a second walk of `apps/api` would be a second set that can silently stop agreeing with the first, which the file header already warns about.

Two scans, because they fail differently:

- **(A) per HANDLER** — a body reaching `assertTestFixturesAllowed(` or `markPreRolloutEraForTesting(` must sit on a `test-fixtures` path. This is the one that catches the off-convention spelling: such a route *still calls the seam*, so it is found here and fails for lacking the segment.
- **(B) per FILE** — a controller that so much as NAMES `IOrderTestFixtureService` / `ORDER_TEST_FIXTURE_SERVICE_TOKEN` must expose at least one `test-fixtures` route. `extractHandlerBody` sees one member, so a handler delegating through a private helper is invisible to (A); (B) does not care where in the file the reference sits.

Three details worth flagging rather than leaving to be discovered:

1. **A body that cannot be resolved counts as a MATCH, not as "does not reach the seam."** The opposite reading hides exactly the handler the scan exists to find. The preceding test already asserts every route resolves, so a `null` here is red there too — this keeps the new scan honest if that ever stops being true.
2. **(B) matches a bare mention in a comment, deliberately.** Same reasoning the header gives for the `contradicts` strictness: what it refuses is meaningless rather than workable — a controller naming the fixture seam while exposing no fixture route. Remedy is to drop the mention or mount the route.
3. **A guard on the guard.** A scan matching *every* controller would make (A) and (B) green for unrelated reasons, so `AuthController` is asserted to match neither.

**Proved red-first rather than asserted.** Renaming the live route's segment to `dev-fixtures` and re-running:

```
✕ discovers at least one test-fixture route (non-vacuity)
✓ every test-fixture route is @Roles(admin) exactly          <- GREEN over an empty set
✓ every test-fixture route reaches assertTestFixturesAllowed() <- GREEN over an empty set
✕ (A) every route reaching the seam is mounted on a test-fixtures path
✕ (B) every controller naming the fixture service exposes a test-fixtures route
Tests: 3 failed, 18 passed, 21 total
```

That is the hole, reproduced: the two path-keyed assertions pass while asserting nothing, and only the pre-existing count catches it — with a message about vacuity rather than about the route. (A) and (B) now fail for the actual reason, naming the offending path. Mutation reverted; the file is byte-identical to `77a416d`.

I kept your framing in the block's docblock: this is **defence in depth**. The real protection is still that `markPreRolloutEraForTesting` calls the gate as its own first statement, so an off-convention route stays triple-gated — what these buy is that it cannot exist *unnoticed*.

---

### SUGGESTION 2 — why the `NODE_ENV` mutation is safe

In the int-spec header, with the three reasons separated and **the load-bearing one named**, because they are not equally strong:

1. **The DataSource is already open.** `getTestHarness()` is a process-wide singleton booted in `beforeAll`, and TypeORM reads `synchronize` once, at initialisation. By the time any case mutates the variable the schema exists; the read never happens again.
2. The value is restored in `afterEach`, ahead of `resetTestHarness()`.
3. `maxWorkers: 1` (`apps/api/test/jest-integration.cjs:24`) means nothing else is in flight.

(1) is the one that matters — (2) and (3) would not save a harness booted lazily per suite — so the note ends with the consequence: **a future change that re-boots the app inside a test must restore `NODE_ENV` first.** Your instinct about `DatabaseModule` is cited directly (`libs/shared/src/database/database.module.ts:72`, `synchronize: NODE_ENV !== 'production'`), so the next reader meets the answer at the same moment they form the question.

---

### Merge conflict

One hunk, in `order-record.repository.spec.ts`: this branch added a `stampPreRolloutEraForTesting` describe directly above `findNetExcludedOrderCandidates`, which main renamed to `findNetExcludedOrderCandidatesPage` (#2834). Kept both — the new block verbatim, the sibling under its new name. No production file conflicted.

### Verification

| | |
|---|---|
| `route-authorization-coverage.spec.ts` | **21/21 pass** (369 s — it `require()`s all 67 controllers through ts-jest) |
| Same spec, deliberately broken | 3 fail, exactly (A), (B) and the non-vacuity count |
| eslint — the changed `src` file | clean (the int-spec's pre-existing `no-unsafe-member-access` lines are outside `apps/api`'s `src/**` lint scope and unchanged by this commit) |
| `pnpm check:invariants` | exit 0 |
| Prettier | only my own added lines normalised; both files carry pre-existing drift in untouched regions, so no blanket `--write` |

Full type-check and the wider jest run are left to CI — this machine builds no `dist`, so a local package-scoped `tsc` would report resolution noise rather than real drift.

Issue #1990, opened 2026-08-03
[IMPL] Frontend — /analytics KPI strip + by-channel table


piotrswierzy:
## Design source

**Product Design parent:** Part of #1976 — see spec at `docs/specs/product-spec-1976-analytics.md` (§ 5 Design 1, § 6, stories S1/S2)

## Scope

The money half of the page, top section.

1. **KPI strip** — Revenue, Orders, AOV (with median shown beside the mean), Units. Uses the existing `.status-strip` + `KpiCard` pattern (which already supports a sparkline slot, tone and an optional link target), matching `/`'s idiom.
2. **By-channel table** — one row per connected channel: revenue, share, orders, AOV, units, trend sparkline. `DataTable` with sortable typed columns, `hideBelow` for narrow widths and card view on mobile.

## Acceptance criteria

- [ ] KPI strip shows revenue, orders, AOV with median, and units for the selected range
- [ ] By-channel table lists every connected channel with its figures and share
- [ ] Every figure carries its currency; where a range spans multiple currencies the page does **not** render a single blended number
- [ ] Where gross and net reporting channels cannot be compared on the same basis, the table says so rather than implying comparability
- [ ] A channel with less history than the selected range is visibly marked (reads the coverage window from the trust data)
- [ ] Cancelled value is visible, so exclusion from revenue is evident rather than silent
- [ ] All numerics use `tabular-nums`; mono for IDs/timestamps per the style guide
- [ ] Loading/error states per section with retry; a channel-table failure does not blank the KPI strip
- [ ] Responsive: KPI strip reflows (1×4 → 2×2 → vertical); table switches to card view; no horizontal page scroll
- [ ] Tests added
- [ ] No new ESLint warnings or type errors introduced

## Effort estimate

- [x] **M** — 3–7 days

## Dependencies

- **Blocked by:** #1987 (sales & channel aggregates), #1986 (route shell)

## Out of scope for THIS issue

- A revenue time-series chart — v1 carries sparkline-level trend only. A real chart is v2 and forces the deferred charting-library decision
- Period-over-period comparison
- Commerce-mix rows (COD / Paczkomat / Smart!) — catalogued as group I, deferred to v2

PR #2917, opened 2026-09-05
docs(user-guide): add operator tutorial for the Analytics page

norbert-kulus-blockydevs:
## Summary

- Adds `docs/user-guide/07-analytics.md` — a full operator tutorial for the `/analytics` page: the date-range toolbar and its 400-day limit, the display-currency picker (native vs. converted, current-rate vs. order-date rate-basis modes), the Net/Gross VAT-basis toggle (URL-state per-view vs. persisted-default in Analytics Settings, with a worked before/after example), every KPI strip metric with its definition quoted verbatim from `docs/specs/metrics-analytics-dashboard.md`, the channel and product tables and their exclusion annotations, the sync/ingestion trust header and Needs Attention section, the full Data Coverage panel and its remediation flows (currency mismatch, tax categories A/B/C, product-matching), and the full Analytics Settings dialog.
- Renumbers `07-diagnostics.md` -> `08-diagnostics.md` and `08-settings-and-admin.md` -> `09-settings-and-admin.md` to make room for the new chapter 7, updating the section index in `docs/user-guide/README.md`, the cross-references in `docs/user-guide/06-orders.md` and `docs/getting-started.md`, **and the seven image assets those two chapters reference** (`docs/user-guide/images/07-{jobs-list,job-details,webhook,cursors}.png` -> `08-*`, `08-{settings,provider-settings,prompt-templates}.png` -> `09-*`) — the file renumber alone left both chapters pointing at images now shadowed by the new chapter 7's own `07-analytics-*.png` files.
- 3 screenshots, cropped to content only (no navbar/sidebar chrome) and reviewed for PII — no real customer/personal names appear in any of them (the demo data used is synthetic and sandbox-labelled).
- README audit: neither root `README.md` nor `docs/getting-started.md` links individually to any user-guide chapter by name — both link generically to the Operator Guide entry point (`docs/user-guide/README.md`) — so no new individual link was added, matching the precedent set by #2830/#2831.

Closes #2916 (product-repo half; the `openlinker-docs` `SOURCES`/sidebar half is tracked separately — see #11).

## Merge-ordering dependency (unmerged upstream fix)

The Order-date pill copy (lines ~33-45) currently quotes `main`'s **actual, still-live** string —
`"This range doesn't filter results yet — coming soon"` — and labels it a disclaimer rather than a
confirmation, per [#3032](https://github.com/openlinker-project/openlinker/issues/3032). A fix for
[...]

PR #3130, opened 2026-09-10
fix(analytics): correct the stale "Order date doesn't filter yet" caveat

### LINKED ISSUE #3032: [BUG] Frontend — Analytics date-range pill claims the range doesn't filter results, but it does
## Problem / Context

`AnalyticsDateRangeToolbar`'s "Order date" pill (`apps/web/src/features/analytics/components/analytics-date-range-toolbar.tsx:35`) renders a fixed caveat:

```
const ORDER_DATE_CAVEAT = "This range doesn't filter results yet — coming soon";
```

with a comment attributing this to `placedAt` not yet being a filterable read-model column, tracked against #1990.

That premise looks stale. `sales-analytics.controller.ts` passes `from`/`to` straight into `IOrderRecordService.getSalesAndChannelAnalytics`, which reads `OrderRecordRepositoryPort.getDailyOrderAggregates(filters, …)` — per that file's own docblock, "`order_records` grouped by day (via `placedAt`)". The 400-day cap in the same controller only makes sense against a filter that's actually applied. #1990 itself appears to be about the KPI-strip design, already fully shipped.

So either:
1. The caveat is genuinely stale and should be removed/updated (and its test at `analytics-date-range-toolbar.test.tsx:81,84` updated accordingly), or
2. Something *does* still bypass the range filter (worth naming explicitly if so — e.g. does the Needs Attention read, which takes no date param at all, get miscategorized as "range-scoped" anywhere downstream?).

Surfaced during review of #2917 (docs(user-guide): analytics tutorial) — [PR review comment](https://github.com/openlinker-project/openlinker/pull/2917#pullrequestreview-5152731923), BLOCKING finding 1. Per `docs/specs/metrics-analytics-dashboard.md`'s own rule ("the divergence is a bug worth raising, not a wording preference to resolve locally"), this is filed rather than silently resolved in the docs PR.

## Proposed Solution

Verify end-to-end whether the "Order date" range actually filters every read it's rendered beside (sales/channel KPIs, daily trend, and anything else sharing the toolbar). If the range is genuinely applied everywhere the pill appears, remove or correct `ORDER_DATE_CAVEAT` in `analytics-date-range-toolbar.tsx` and update its test. If some read on the same page is NOT range-scoped (e.g. Needs Attention), keep an accurate caveat that names the specific gap rather than a blanket "doesn't filter yet" claim, or move the pill so it isn't rendered beside a read it doesn't affect.

## Classification

**Type**: Frontend (with a verification pass into Application/Interface layer read paths)
**Layer**: Interface
**File(s)**: `apps/web/src/features/analytics/components/analytics-date-range-toolbar.tsx`, `apps/web/src/features/analytics/components/analytics-date-range-toolbar.test.tsx`, `apps/api/src/analytics/http/sales-analytics.controller.ts`, `libs/core/src/orders/` (`IOrderRecordService.getSalesAndChannelAnalytics`, `OrderRecordRepositoryPort.getDailyOrderAggregates`)

## Dependencies

- None

## Assumptions

- The fix is scoped to correcting the caveat's accuracy (removal or precise wording), not to building new date-range filtering — the backend plumbing already appears to honor `from`/`to`.

## Acceptance Criteria

- [ ] The date-range filter's actual reach is verified against every analytics read rendered on the same page as the toolbar.
- [ ] `ORDER_DATE_CAVEAT` is removed if the range genuinely filters everything, or corrected to name the specific unfiltered read if one exists.
- [ ] `analytics-date-range-toolbar.test.tsx` is updated to match the corrected copy/behavior.

norbert-kulus-blockydevs:
## Summary

- `ORDER_DATE_CAVEAT` in `AnalyticsDateRangeToolbar` claimed the date range doesn't filter results, attributing this to `placedAt` not being a filterable read-model column (tracked against #1990). That premise is stale.
- Verified end-to-end: `SalesAnalyticsController.getSalesAnalytics` passes `from`/`to` into `IOrderRecordService.getSalesAndChannelAnalytics`, which reads `getDailyOrderAggregates` grouped by `placedAt`. This is the source of the KPI strip, the channel/product tables, and the trend chart — all range-scoped end to end. The controller's 400-day range cap only makes sense against a filter that's actually applied, corroborating this.
- Verification also surfaced a real, still-open gap the old copy didn't name: `GET /analytics/needs-attention` (`NeedsAttentionController` -> `NeedsAttentionService.getSummary()`) takes **no date parameters at all** — coverage gaps, stock at risk, and failed-sync value are always the current snapshot, regardless of the toolbar's range. `AnalyticsNeedsAttention` renders on the same page, below the toolbar, so this is a genuine, worth-naming caveat rather than a stale one.
- Per the issue's option 2, the caveat is corrected to name that specific gap instead of the blanket (and now false) claim that nothing is filtered.

## Changes

- `apps/web/src/features/analytics/components/analytics-date-range-toolbar.tsx`: replaced the stale `ORDER_DATE_CAVEAT` string and its comment with copy naming the one section the range doesn't reach ("Needs attention").
- `apps/web/src/features/analytics/components/analytics-date-range-toolbar.test.tsx`: updated the caveat-popover test to assert the corrected copy.

## Test plan

- [x] `pnpm lint` (full monorepo, via the pre-commit hook) — 0 errors
- [x] `pnpm type-check` (full monorepo, via the pre-commit hook) — 0 errors
- [x] `pnpm smart-test --no-integration` (scoped to the two changed files via `vitest related`) — 6 test files / 151 tests passed
- [x] Manually traced every analytics read rendered on `/analytics` (KPI strip, channel table, product table, trend, convert note, data coverage panel, trust header, needs-attention) against whether it consumes `from`/`to`

Closes #3032


norbert-kulus-blockydevs:
## PR Review — #3130 (frontend copy/UX fix)

**Verdict: ✅ Approve.**

### What this actually is
A two-line change to `analytics-date-range-toolbar.tsx` + its test: the "†" caveat popover text and aria-label, nothing backend.

### Fact-check of the corrected copy
Verified both halves of the new claim against `main`:
- **"Filters revenue, orders, and the tables above"** — true. `GET /analytics/sales` (`SalesAnalyticsController.getSalesAnalytics`) requires `from`/`to` query params and feeds them through to `getSalesAndChannelAnalytics` → `getDailyOrderAggregates`, which groups on `date_trunc('day', rec."placedAt" ...)` (`order-record.repository.ts:625,676`). The old "doesn't filter results yet" copy was simply false by the time this PR was written.
- **"'Needs attention' below always shows current status, not this range"** — true. `NeedsAttentionController.getNeedsAttention()` (`needs-attention.controller.ts`) takes **zero** query parameters — no date param exists at all, confirmed by reading the controller directly.

So this isn't a premature caveat removal — it's a genuinely more accurate statement replacing a stale one, and it correctly keeps a caveat rather than deleting it outright (the "Needs attention" gap is real and unaddressed).

### UI/UX quality
- Reuses the exact `AnalyticsTrustHeader` info-popover pattern and explains *why* — a real `<button>` + Radix `Popover`, not `Tooltip`, because `Tooltip` ignores `pointerType === 'touch'` and would be permanently unreachable on mobile. Good, documented accessibility reasoning, not just copied boilerplate.
- `aria-label` is kept in sync with the visible popover text (both derive from the same `ORDER_DATE_CAVEAT` constant) — no drift risk between the two.
- Test coverage updated correctly for both the negative case (`queryByText` not present before interaction) and the positive case (caveat visible after click), with a code comment explaining the change (`#3032`).

### One coordination flag (not blocking this PR, but real)
PR **#2917** (Analytics operator-tutorial docs, still open, unmerged) **quotes the exact old string verbatim**: `docs/user-guide/07-analytics.md:44` on that branch reads *"activating it reads \"This range doesn't filter results yet — coming soon.\""* Once both PRs merge — in either order — the tutorial will contain a literal quote of copy that no longer exists in the UI. This was already flagged as a sequencing risk in the #2917 review; recording it here too so whoever merges #2917 remembers to reconcile the quoted string. (Lower-stakes secondary hit: `docs/plans/mockups/analytics-display-currency-picker.html:518` also embeds the old caveat text in a static mockup `title` attribute — cosmetic only, not user-facing.)

No other findings — this is a clean, well-verified, appropriately-scoped fix.

piotrswierzy (commented):

**The caveat was genuinely stale, and I traced it end to end rather than taking the PR's word.** Removing it does not make the UI claim anything untrue:

- `sales-analytics.controller.ts:120-142` parses `from`/`to`, enforces `to > from` and the 400-day cap, and passes both into `getSalesAndChannelAnalytics`.
- `order-record.repository.ts:1319-1322` — `applySalesAnalyticsScope` applies `rec."placedAt" IS NOT NULL AND rec."placedAt" >= :salesFrom AND rec."placedAt" < :salesTo`. A real bound filter, not a display projection.
- `getDailyOrderAggregates` (:625, day-truncated at :676) is the shared substrate for the KPI strip, both tables **and** the trend, so #2899's range-following trend and the headline figures are scoped by one predicate and cannot diverge.

**The second claim also verifies**, which is what makes option 2 the right resolution rather than a plain deletion: `needs-attention.controller.ts:38-45` takes **no `@Query()` at all** and calls `getSummary()` with no arguments, while `analytics-page.tsx:314` renders `<AnalyticsNeedsAttention />` directly below the toolbar. That gap is real and still open. And `AnalyticsDataCoveragePanel` **is** range-scoped (`coverageFilters` from the same `from`/`to`), so the new copy is right not to name it.

One constant, two consumers (`aria-label` :138, `PopoverContent` :143) — reported and rendered copy cannot drift. The replaced comment records *what was verified* and *why the new gap is real*, which is what stops the next reader re-adding the old claim. Scoping is right too: the fix lands on the code and the doc half stays with #2917, which owns the doc. All 8 CI checks green.

### IMPORTANT — the new copy has a directional error, in a copy-accuracy PR

`analytics-date-range-toolbar.tsx:42` — `'Filters revenue, orders, and the tables above.'`

The toolbar renders at `analytics-page.tsx:203`; the KPI strip is at :279 and the tables at :290/:297 — i.e. **below** the popover, not above. The same sentence then correctly says *"Needs attention **below**"*, so the two directions contradict each other and one is wrong from the reader's vantage point.

This matters more than usual because the string is also the `aria-label` (:138), where a screen-reader user has no visual layout to correct against. Suggest dropping the direction entirely: *"Filters revenue, orders and the tables on this page. 'Needs attention' always shows current status, not this range."*

### SUGGESTIONS

- `AnalyticsTrustHeader` (`analytics-page.tsx:313`) is fed by `useAnalyticsTrustQuery()` — **also no date arguments** — and renders immediately above `AnalyticsNeedsAttention`. Naming exactly one unscoped section reads as exhaustive when two are. Defensible as-is (ingestion liveness is inherently a "now" fact, unlike a coverage **count**), but if the sentence is being reworded anyway, *"'Needs attention' and the sync status below"* costs three words.
- The test's `queryByText(/needs attention/i)` is a looser pre-assertion than the old `/coming soon/i` — that phrase could plausibly appear elsewhere in this toolbar later, whereas `coming soon` could not. Asserting a distinctive fragment (`/always shows current status/i`) keeps the negative assertion meaningful.

**Note on #2917**: that PR already quotes this PR's new string as though it had merged, and carries a comment asserting so. It has not. #3130 must land first.

---

norbert-kulus-blockydevs:
Addressed both findings from the review:

**IMPORTANT (directional error)** — `ORDER_DATE_CAVEAT` said "Filters revenue, orders, and the tables **above**" while the toolbar actually renders **above** the KPI strip/tables (i.e. the caveat's own vantage point was backwards relative to "Needs attention below", which was correct). Dropped directional wording entirely rather than just flipping it, since the string also stands alone as the button's `aria-label` where a screen-reader user has no layout to resolve a direction against. New copy: *"Filters revenue, orders, and the tables on this page. 'Needs attention' and the sync status always show current status, not this range."*

**SUGGESTION (name both unscoped sections)** — took this one: `AnalyticsTrustHeader` (fed by `useAnalyticsTrustQuery()`, also called with no date args) is now named alongside "Needs attention" rather than leaving it unmentioned, which previously read as an exhaustive list when it wasn't.

**SUGGESTION (tighten test assertion)** — updated both the negative and positive assertions from `/needs attention/i` (which could plausibly appear elsewhere in the toolbar later) to the distinctive fragment `/always shows current status/i`.

Also updated the code comment above the constant to record what was verified and why, per the same "record what changed and why" pattern the original PR used.

Pushed as `74372a865`.

Issue #3066, opened 2026-09-10
feat(inventory): locations list page with filter bar


norbert-kulus-blockydevs:
## Problem / Context

No page renders the full locations list — only the bootstrap probe is reachable.

## Proposed Solution

Add `apps/web/src/pages/inventory/inventory-locations-page.tsx` (new directory) with a `DataTable` (code, name, kind, country/postcode, owning connection, status) and a `FilterBar`-style kind/status chip filter, following `connections-list-page.tsx`.

## Classification

**Type**: Frontend
**Layer**: Interface (page)
**File(s)**: apps/web/src/pages/inventory/inventory-locations-page.tsx

## Dependencies

- Part of #3063

## Assumptions

- None

## Acceptance Criteria

- [ ] List renders with kind and status filters
- [ ] Inactive (retired) locations are visible via a toggle, not hidden by default
- [ ] Tests added or updated for non-trivial logic
- [ ] No architecture boundary violations (CORE ↔ Integration)

norbert-kulus-blockydevs:
Implemented in #3135, merged onto branch 3068-location-delete-dialog (epic #3063).

Issue #3068, opened 2026-09-10
feat(inventory): delete confirm incl. "location in use" 409 + retire action


norbert-kulus-blockydevs:
## Problem / Context

`DELETE /inventory/locations/:id` is refused with 409 `LocationInUseError` while any inventory_items row still references the location — no UI distinguishes this from an ordinary failure, and there is no retire (soft-deactivate) action.

## Proposed Solution

Add a `ConfirmDialog` that on a 409 shows the in-use explanation and offers `PATCH { status: 'inactive' }` (retire) as the alternative, matching the reviewed mockup at https://claude.ai/code/artifact/fbb5f9e1-52a3-4b7b-8136-ad59dcf4f318.

## Classification

**Type**: Frontend
**Layer**: Interface (component)
**File(s)**: apps/web/src/features/inventory/components/location-dialog.tsx

## Dependencies

- Part of #3063

## Assumptions

- None

## Acceptance Criteria

- [ ] A 409 on delete renders the in-use explanation, not a generic error
- [ ] Retire is offered from the same dialog and succeeds
- [ ] Tests added or updated for non-trivial logic
- [ ] No architecture boundary violations (CORE ↔ Integration)

norbert-kulus-blockydevs:
Implemented in #3134, merged onto branch 3067-location-dialog (epic #3063).

Issue #3067, opened 2026-09-10
feat(inventory): create/edit location dialog


norbert-kulus-blockydevs:
## Problem / Context

No form exists for any of code/name/kind/ownerConnectionId/externalRef/country/postcode/lat/lng.

## Proposed Solution

Build a `Dialog`-based form (React Hook Form + Zod) covering every `CreateLocationDto` field; code is uppercased client-side to match the server's normalisation, geo fields are optional with min/max validation matching the DTO's `@Min`/`@Max`.

## Classification

**Type**: Frontend
**Layer**: Interface (component + form)
**File(s)**: apps/web/src/features/inventory/components/location-dialog.tsx, location-dialog.schema.ts

## Dependencies

- Part of #3063

## Assumptions

- None

## Acceptance Criteria

- [ ] Every CreateLocationDto field is represented
- [ ] Duplicate-code 409 surfaces as a field error
- [ ] Edit pre-fills from the existing row and omits `code` per the PATCH contract
- [ ] Tests added or updated for non-trivial logic
- [ ] No architecture boundary violations (CORE ↔ Integration)

norbert-kulus-blockydevs:
Implemented in #3133, merged onto branch 3065-locations-query-hooks (epic #3063).

Issue #3065, opened 2026-09-10
feat(inventory): locations query hooks + filters (kind, status, codePrefix)


norbert-kulus-blockydevs:
## Problem / Context

No hook exists for the full paged list or the by-id read.

## Proposed Solution

Add `use-inventory-locations-query.ts` (kind/status/codePrefix filters, page/limit), `use-inventory-location-query.ts` (single), and mutation hooks for create/update/delete in `features/inventory/hooks/`, invalidating the list key on success.

## Classification

**Type**: Frontend
**Layer**: Interface (hooks)
**File(s)**: apps/web/src/features/inventory/hooks/*

## Dependencies

- Part of #3063

## Assumptions

- None

## Acceptance Criteria

- [ ] List hook accepts kind/status/codePrefix and pages
- [ ] Mutation hooks invalidate the list query on success
- [ ] Tests added or updated for non-trivial logic
- [ ] No architecture boundary violations (CORE ↔ Integration)

norbert-kulus-blockydevs:
Implemented in #3132, merged onto branch 3064-locations-api-client (epic #3063).

PR #3131, opened 2026-09-10
feat(inventory): extend locations API client (list/get/create/update/delete)

### LINKED ISSUE #3064: feat(inventory): extend locations API client (list/get/create/update/delete)
## Problem / Context

`features/inventory/api/inventory.api.ts` only implements `listActiveLocations` (a `?status=active&limit=1` probe) and `bootstrapLocations`. `GET /inventory/locations` (full list + filters), `GET/PATCH/DELETE /inventory/locations/:id` and `POST /inventory/locations` have no client method.

## Proposed Solution

Add `list(filters, pagination)`, `get(id)`, `create(input)`, `update(id, patch)`, `remove(id)` to `InventoryApi` in `inventory.api.ts`, and the matching request/response shapes to `inventory-locations.types.ts` — the DTOs already exist server-side in `apps/api/src/inventory/http/dto/` to mirror.

## Classification

**Type**: Frontend
**File(s)**: apps/web/src/features/inventory/api/inventory.api.ts, inventory-locations.types.ts

## Dependencies

- Part of #3063

## Assumptions

- None

## Acceptance Criteria

- [ ] All 4 remaining routes have a typed client method
- [ ] Types mirror LocationResponseDto / CreateLocationDto / UpdateLocationDto field-for-field

jakubretajczykBD:
## Summary
- Adds `listLocations`, `getLocation`, `createLocation`, `updateLocation`, `deleteLocation` to `InventoryApi` — the remaining CRUD surface of `inventory_locations` (#2316), previously only reachable via `curl`
- Adds request/response types mirroring `LocationResponseDto` / `CreateLocationDto` / `UpdateLocationDto` / `ListLocationsQueryDto` field-for-field in `inventory-locations.types.ts`
- Widens `PaginatedInventoryLocations.items` from `InventoryLocationSummary[]` to `InventoryLocation[]` (a superset, so existing `total`-only callers are unaffected)
- Re-exports the new types from the `features/inventory` barrel for the upcoming list page (#3066)

**Naming deviation from the issue text**: named `listLocations`/`getLocation`/`createLocation`/`updateLocation`/`deleteLocation` rather than the plain `list`/`get`/`create`/`update`/`remove` the issue sketched — `InventoryApi` already carries `list` for inventory *items*, so the plain names would collide. Followed the `listActiveLocations`/`bootstrapLocations` naming this file already established.

Closes #3064, first code sub-issue of #3063.

## Test plan
- [x] New `inventory.api.test.ts` (8 cases: query building, `countryIso2` uppercasing, create/update JSON body, `null`-vs-omitted patch semantics, delete) — all green

### DIFF
--- a/apps/api/src/inventory/http/inventory-locations.controller.spec.ts
+++ b/apps/api/src/inventory/http/inventory-locations.controller.spec.ts
@@ -17,6 +17,7 @@ import {
   LocationInUseError,
   type ILocationService,
 } from '@openlinker/core/inventory';
+import { ROLE_PERMISSIONS, UserRoleValues } from '@openlinker/core/users';
 import { InventoryLocationsController } from './inventory-locations.controller';
 
 function makeLocation(
@@ -221,5 +222,19 @@ describe('InventoryLocationsController', () => {
       expect(result.created).toEqual([]);
       expect(result.existingCodes).toEqual(['MAIN']);
     });
+  });
+});
+
+describe('inventory-locations:write permission lockstep', () => {
+  it('should be granted to exactly the roles this controller @Roles(admin) on every write route', () => {
+    // `inventory-locations:write` is a display predicate only (no permission
+    // guard exists) - if this drifts from the controller's actual @Roles,
+    // either add a permission-based guard or revert the grant, the
+    // `shipments:write` / `automations:write` discipline.
+    const rolesWithPermission = UserRoleValues.filter((role) =>
+      ROLE_PERMISSIONS[role].includes('inventory-locations:write')
+    );
+
+    expect(rolesWithPermission).toEqual(['admin']);
   });
 });
--- a/apps/web/src/app/nav-registry.ts
+++ b/apps/web/src/app/nav-registry.ts
@@ -71,6 +71,11 @@ export const BASE_NAV_GROUPS: readonly NavRegistryGroup[] = [
     label: 'Platform',
     items: [
       { to: '/connections', label: 'Connections', countKey: 'connections' },
+      // Reads are `admin`/`operator`/`viewer` (InventoryLocationsController)
+      // - gated on `inventory:read` rather than `inventory-locations:write`,
+      // so a `packer` (which holds no permissions at all) is the only role
+      // that never sees the entry.
+      { to: '/inventory/locations', label: 'Inventory locations', requiresPermission: 'inventory:read' },
       { to: '/adapters', label: 'Adapters' },
       { to: '/settings', label: 'Settings' },
     ],
--- a/apps/web/src/app/nav-registry.ts
+++ b/apps/web/src/app/nav-registry.ts
@@ -71,11 +71,6 @@ export const BASE_NAV_GROUPS: readonly NavRegistryGroup[] = [
     label: 'Platform',
     items: [
       { to: '/connections', label: 'Connections', countKey: 'connections' },
-      // Reads are `admin`/`operator`/`viewer` (InventoryLocationsController)
-      // - gated on `inventory:read` rather than `inventory-locations:write`,
-      // so a `packer` (which holds no permissions at all) is the only role
-      // that never sees the entry.
-      { to: '/inventory/locations', label: 'Inventory locations', requiresPermission: 'inventory:read' },
       { to: '/adapters', label: 'Adapters' },
       { to: '/settings', label: 'Settings' },
     ],
--- a/apps/web/src/app/routes/inventory-locations.route.tsx
+++ b/apps/web/src/app/routes/inventory-locations.route.tsx
@@ -1,0 +1,13 @@
+import type { RouteObject } from 'react-router-dom';
+import type { RouteCrumbHandle } from '../nav-registry.types';
+
+export const inventoryLocationsRoute: RouteObject = {
+  path: 'inventory/locations',
+  handle: { crumb: { group: 'Platform', title: 'Inventory locations' } } satisfies RouteCrumbHandle,
+  lazy: async () => {
+    const { InventoryLocationsPage } = await import(
+      '../../pages/inventory/inventory-locations-page'
+    );
+    return { Component: InventoryLocationsPage };
+  },
+};
--- a/apps/web/src/app/routes/root.route.tsx
+++ b/apps/web/src/app/routes/root.route.tsx
@@ -24,6 +24,7 @@ import { cursorsRoute } from './cursors.route';
 import { customersRoute } from './customers.route';
 import { devUiRoute } from './dev-ui.route';
 import { insightsRoute } from './insights.route';
+import { inventoryLocationsRoute } from './inventory-locations.route';
 import { listingsRoute } from './listings.route';
 import { aiProviderSettingsRoute } from './ai-provider-settings.route';
 import { mcpTokensRoute } from './mcp-tokens.route';
@@ -71,6 +72,7 @@ export const coreChildren: RouteObject[] = [
   automationsRoute,
   invoicesRoute,
   connectionsRoute,
+  inventoryLocationsRoute,
   adaptersRoute,
   newConnectionRoute,
   advancedNewConnectionRoute,
--- a/apps/web/src/app/routes/route-lazy.test.ts
+++ b/apps/web/src/app/routes/route-lazy.test.ts
@@ -83,7 +83,7 @@ const lazyRoutes = collectLazyRoutes([
  *   - prompt-templates-legacy-redirects (inline `<Navigate>` element)
  *   - `/analytics` legacy alias (inline `<Navigate>` to `/`, #2740)
  */
-const EXPECTED_LAZY_ROUTE_COUNT = 63;
+const EXPECTED_LAZY_ROUTE_COUNT = 64;
 
 describe('route lazy contract', () => {
   it(`the registered route tree contains exactly ${EXPECTED_LAZY_ROUTE_COUNT} lazy routes`, () => {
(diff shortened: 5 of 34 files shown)


norbert-kulus-blockydevs:
## PR Review — #3131 (senior frontend reviewer, API-contract focus)

**Scope reviewed**: `inventory-locations.types.ts` additions, `inventory.api.ts` (`listLocations`/`getLocation`/`createLocation`/`updateLocation`/`deleteLocation`), new `inventory.api.test.ts`, barrel re-exports.

### Contract fidelity ✅ — verified field-for-field against the real backend
Diffed every new FE type against the actual shipped DTOs in `apps/api/src/inventory/http/dto/`:
- `InventoryLocation` ↔ `LocationResponseDto` — exact match (id, code, name, kind, ownerConnectionId, externalRef, status, countryIso2, postcode, latitude, longitude, createdAt, updatedAt).
- `CreateInventoryLocationInput` ↔ `CreateLocationDto` — exact match.
- `UpdateInventoryLocationInput` ↔ `UpdateLocationDto` — exact match, and correctly **omits `code`**, matching the backend's explicit "code is not patchable" rule (confirmed in the controller's own docblock: sending it is a 400 under `forbidNonWhitelisted`).
- `InventoryLocationFilters`/`InventoryLocationListPagination` ↔ `ListLocationsQueryDto` — exact match, including the deliberate `page`/`limit` (not `limit`/`offset`) divergence from the sibling inventory-items query shape.

### 204/409 handling ✅ — verified against the real controller, not assumed
- Confirmed `DELETE /inventory/locations/:id` returns `HttpStatus.NO_CONTENT` (204) on success and the global filter maps `LocationInUseError` → 409 on the referential-integrity refusal.
- Confirmed the shared `apps/web/src/app/api/api-client.ts` `request()` already returns `undefined` for any 204 and throws `ApiError.fromResponse(...)` for any non-2xx (including 409), with `ApiError.isConflict()` a pre-existing shared primitive already used by two other features (`analytics-data-coverage-panel.tsx`, `fulfillment-conflict.ts`). So `deleteLocation`'s docblock claim — "callers should check `ApiError.isConflict()`" — is accurate and needs no new code in this file; the 409 signal is real, typed, and already flows correctly through the generic client. Good instinct not to duplicate that.

### Type-widening safety ✅ — verified against the actual (sole) consumer
`PaginatedInventoryLocations.items` widens `InventoryLocationSummary[]` → `InventoryLocation[]`. Grepped every consumer: the only one is `useActiveLocationCountQuery`, which reads `.total` exclusively (`await apiClient.inventory.listActiveLocations()).total`) — confirmed the widening is a real no-op for existing callers, not just an assumed-safe claim. Also confirmed `GET /inventory/locations` (the route `listActiveLocations` hits) really does return full `LocationResponseDto` rows via `toDto()`, so the widened type isn't overclaiming what the wire returns.

### Architecture / layering ✅
Correctly placed in `features/inventory/api/`, no `shared` → `features` violation, no `any` anywhere in the diff, naming deviation from the issue text (`listLocations` vs. plain `list`) is well-justified (collision with the existing inventory-items `list` on the same interface) and consistent with this file's established `listActiveLocations`/`bootstrapLocations` convention.

### Non-blocking observations
1. **No error-path test at this file's level** — all 8 new tests cover the happy path only (no test asserts `deleteLocation` surfaces a 409, or `getLocation` a 404, as an `ApiError`). Given the generic 409→`ApiError.isConflict()` translation is already covered by `api-client.test.ts`, re-testing it per feature file would be redundant — but a single "propagates ApiError.isConflict() on 409" test here would make this file's own doc-comment claim self-verifying rather than relying on a reader to trust it.
2. `ownerConnectionId` has no client-side format guard (fine — the backend's `@IsUUID()` is the real gate, and duplicating format validation client-side would just be a second place to drift).

**Verdict: ✅ Approve.**

piotrswierzy (approved):
## `/pr-review` — ✅ Approve

Clean, well-documented transport slice: five methods, field-for-field DTO mirrors, 8 tests including query building and patch-null semantics. No architectural concerns.

**Verified rather than assumed** — the `PaginatedInventoryLocations.items` widening is genuinely safe. `InventoryLocationSummary` is `{id, code, name, status: string}` and `InventoryLocation` narrows `status` while adding fields, so both `listActiveLocations` (which reads only `total` — checked in `use-active-location-count-query.ts`) and `router-readiness-panel.tsx` are unaffected.

Two things done right that are easy to get wrong:
- **`code` is not normalised here**, which respects the one-place rule in `architecture-overview.md` § Inventory (#2316) — normalisation belongs to the application service because the unique index is case-sensitive.
- The naming deviation (`listLocations` rather than `list`) is correct: `list` is already taken by inventory *items* on the same interface.
- The `null`-vs-omitted PATCH test (~237) pins exactly the semantic that matters.

Also confirmed: the `inventory` slug is **already registered** in both `no-restricted-imports` groups for all five canonical subdirectories (`.eslintrc.js:279-283`, `:546-550`), so **no ESLint registration is owed by this stack**.

### SUGGESTIONS
- `buildLocationsQuery` (~327) uppercases `countryIso2` client-side. Harmless and disclosed — but it becomes the *first* of three FE copies once #3133 lands (see stack note). If it stays, it should be the only one.
- Two added lines exceed the repo's `printWidth: 100` (the `InventoryLocationKindValues` declaration, and the barrel re-export). Cosmetic — and please don't fix it with a broad `prettier --write`, since `main` is not uniformly Prettier-clean.

---
### ⚠️ Stack-level: `Closes #N` will not fire on any of the five

| PR | base | trailer |
|---|---|---|
| #3131 | `3063-inventory-locations-crud-ui` | `Closes #3064` |
| #3132 | `3064-locations-api-client` | `Closes #3065` |
| #3133 | `3065-locations-query-hooks` | `Closes #3067` |
| #3134 | `3067-location-dialog` | `Closes #3068` |
| #3135 | `3068-location-delete-dialog` | `Closes #3066` |

GitHub auto-closes only on merge into the **default** branch, and `CLAUDE.md` forbids hand-closing. As written the stack leaves five open sub-issues with no mechanism to shut them. Roll the trailers onto the epic PR (`3063-inventory-locations-crud-ui` → `main`) — which does not exist yet — or accept they close only when that lands carrying them.

piotrswierzy (commented):

### My earlier approval is explicitly withdrawn

Not a quality judgement — **the PR changed underneath it.** At +303/4 files it was a transport slice. It is now +2859/33 files, because the growth **folded in all four siblings** merged bottom-up (#3135 → #3134 → #3133 → #3132, whose branch *is* this head). So #3131 is now the entire epic: transport, hooks, both dialogs, a 469-line list page, a route, a Settings tile, and a `PermissionValues` addition in `libs/core`.

None of the findings below were reachable in the 4-file diff I approved, and **all four sibling reviews' findings are now live here.** The title describes ~10% of the diff.

Structural consequence: `gh pr view 3063` → *could not resolve*. **The base branch has no PR to `main`**, so nothing closes #3064–#3068.

### All four prior stack-level findings — genuinely resolved

- **Retire one-way door** — `status` is now on the update input and emitted by `toUpdateInput`, with row-level **Retire** *and* **Reactivate** and round-trip tests. This was the one I said I'd hold the stack on; it's fixed.
- **`countryIso2` ×3** → one `normalizeCountryIso2()`, 3 call sites, 0 duplicates.
- **Two error-suppression implementations** → extracted to `shared/api/api-error.ts` → `isUnmappedApiError`.
- **"every order is unfulfillable"** → copy replaced, with an in-place comment citing #2407's off→on refusal.

### New findings — all one species: a claim that outlived the code

- **The ungated-tile justification is fabricated.** It cites *"the sidebar nav entry already gated on `inventory:read`"* — **that nav entry does not exist**; no branch commit touched `nav-registry.ts`, and the tile is the page's only entry point. The conclusion survives on the controller's real `@Roles`, but the load-bearing clause is invented and repeated into a test docblock.
- **`location-delete-dialog.tsx` still asserts** *"`toUpdateInput` never sends `status` back"* and *"the list renders no Reactivate action"* — both false at this head. Worse, the stale comment **suppresses correct copy**, withholding a reassurance that is now true.
- **`toCreateInput` re-introduces `code.toUpperCase()`** — a second normalisation site, against § Inventory (#2316)'s "exactly ONE place". The irony: `countryIso2` was consolidated in the same commit.
- **The edit dialog PATCHes the full object including `status`**, so saving a name change **silently un-retires** a row someone retired meanwhile — a lost update against the field #3134 just made a row action.
- **`.table-actions` restyles a class `users-page.tsx` already uses unstyled**, untested, and its comment cites two consumer pages that don't use it.
- **`codePrefix`/`countryIso2` are wired end-to-end but unreachable** — no search input at all, on a registry whose primary key is an operator-typed `code`.

Plus six suggestions: the page-level mutation pending state disables all rows; the permission-lockstep test hardcodes `['admin']` rather than reading `@Roles`, so its docblock over-claims; `latLngField` is misnamed; the crumb group is `'Platform'`; no `encodeURIComponent`; `countryIso2` accepts `'12'`.

### Worth recording

The cache-invalidation test pins that `locationsAll()` actually clears three **real** cached queries via `dataUpdateCount` rather than asserting a mock call. The `pageOutOfRange` branch prevents a false "zero locations" claim. And the documented refusal to declare `sortable` columns — because the table *"would look globally sorted while lying about it"* — is exactly the right instinct. ~45 tests, tokens only, ESLint slug already registered, lazy-route count bumped.

**Priority:** the stale comments (1), the `status` PATCH lost update (2), `code.toUpperCase()` (3), `.table-actions` scope (4), and **decide the stack structure** — with no PR to `main`, five sub-issues have no path to closing.

---

PR #3132, opened 2026-09-10
feat(inventory): locations query hooks + filters (kind, status, codePrefix)

jakubretajczykBD:
## Summary
- Adds `useInventoryLocationsQuery` (kind/status/countryIso2/codePrefix filters, page/limit) and `useInventoryLocationQuery` (single, id-gated `enabled`) query hooks
- Adds `useCreateInventoryLocationMutation`, `useUpdateInventoryLocationMutation`, `useDeleteInventoryLocationMutation`
- Adds `inventoryQueryKeys.locationsAll()` (`['inventory', 'locations']`) — the one prefix all three write hooks invalidate against, since a write can move both a filtered list entry AND the unrelated install-wide `activeLocations` count
- Re-exports the new hooks from the `features/inventory` barrel for the upcoming list page (#3066)

Stacked on #3064 (#3131) — these hooks consume the API client methods added there. Base branch is `3064-locations-api-client`, not the epic branch, until #3131 merges.

Closes #3065.

## Test plan
- [x] `pnpm --filter @openlinker/web type-check`
- [x] `eslint` on changed files
- [x] 8 new tests across 3 files: query-hook data flow + error surfacing, and — the one worth calling out — a dedicated invalidation test that renders a *sibling read* against all three locations-registry key shapes (`activeLocations`, `locations()`, `locationDetail()`) and asserts each one actually re-fetches after create/update/delete, rather than only asserting `invalidateQueries` was called with the right array

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1788, opened 2026-07-22
[TASK] Frontend — instrument full e-commerce reel viewer events (batch 1)


norbert-kulus-blockydevs:
## Problem / Context

The full e-commerce flow is the highest-priority marketing reel. In viewer (read-only) mode the write actions are locked, so the signal is exploration plus intent-to-convert clicks on the locked buttons, not successful mutations.

## Proposed Solution

Wire `captureDemoEvent(...)` at the call sites below (from the demo-events catalog). Viewer-clickable only, no backend. `★` marks the intent-to-convert clicks (the primary conversion signal). For `ReadOnlyLock`-gated buttons, wire to the wrapping `read-only-lock` span (the disabled button swallows its own click); for non-gated actions wire to the existing `onClick`/submit.

| Event | Props | File:line |
|---|---|---|
| `demo_products_viewed` | result_count_bucket | `pages/products/products-list-page.tsx` (mount) |
| `demo_product_row_expanded` | - | `pages/products/products-list-page.tsx:1159` |
| `demo_offer_create_launched` | source | `pages/products/products-list-page.tsx:469` / `:479` |
| `demo_offer_marketplace_picked` | platform | `pages/products/marketplace-picker-modal.tsx:99` |
| `demo_offer_wizard_step_advanced` | platform, step | `features/listings/components/AllegroCreateOfferWizard.tsx:662`; bulk `features/listings/components/bulk/bulk-wizard.tsx:196` |
| `demo_offer_wizard_review_reached` | platform | Allegro `:696`; Woo `features/listings/components/WoocommercePublishWizard.tsx:861` |
| ★ `demo_offer_create_attempted` | platform, mode | Allegro `:1327`; Woo `:662`/`:868`; bulk `features/listings/components/bulk/bulk-confirm-modal.tsx:107` |
| `demo_orders_viewed` | - | `pages/orders/orders-list-page.tsx` (mount) |
| `demo_orders_filtered` | filter, value | `pages/orders/orders-list-page.tsx:964` / `:1048` |
| `demo_order_opened` | - | `pages/orders/orders-list-page.tsx:563` |
| `demo_order_timeline_expanded` | - | `features/orders/components/order-activity-timeline.tsx:152` |
| `demo_label_form_opened` | entry | `features/orders/components/order-shipment-panel.tsx:132` / `:145` |
| ★ `demo_label_generate_attempted` | carrier | `features/orders/components/generate-label-form.tsx:741` |
[...]

PR #3133, opened 2026-09-10
feat(inventory): create/edit location dialog

### LINKED ISSUE #3067: feat(inventory): create/edit location dialog

jakubretajczykBD:
## Summary
- Adds `LocationDialog` (React Hook Form + Zod) covering every `CreateLocationDto` field, plus `location-dialog.schema.ts` with the form↔API mapping
- `code` is uppercased client-side to match the server's normalisation and is rendered only on **create** — `UpdateLocationDto` has no `code` field (it's the row's natural key), so the edit form omits it entirely rather than silently dropping an edited value
- A 409 from `createLocation` (`DuplicateLocationCodeError`) is mapped to a field error on Code; a 422 (`LocationOwnerConnectionNotFoundError`, in practice unreachable through the select since it's populated from real connections) maps to `ownerConnectionId`
- Edit pre-fills every field from the existing row

Stacked on #3065 (#3132) — consumes its mutation hooks. Base branch is `3065-locations-query-hooks`.

Note on epic build order: built ahead of #3066 (list page) deliberately — the list page's Add/Edit buttons need a real dialog to open, so this one (and #3068's delete/retire dialog) come first. Both will be wired into the page in #3066.

Closes #3067.

## Test plan
- [x] 6 new tests: create-mode fields, edit-mode fields (no Code), empty-submit validation, successful create with mapped payload, 409→field-error mapping, successful update omitting `code` from the patch — all green

### DIFF
--- a/apps/api/src/inventory/http/inventory-locations.controller.spec.ts
+++ b/apps/api/src/inventory/http/inventory-locations.controller.spec.ts
@@ -17,6 +17,7 @@ import {
   LocationInUseError,
   type ILocationService,
 } from '@openlinker/core/inventory';
+import { ROLE_PERMISSIONS, UserRoleValues } from '@openlinker/core/users';
 import { InventoryLocationsController } from './inventory-locations.controller';
 
 function makeLocation(
@@ -221,5 +222,19 @@ describe('InventoryLocationsController', () => {
       expect(result.created).toEqual([]);
       expect(result.existingCodes).toEqual(['MAIN']);
     });
+  });
+});
+
+describe('inventory-locations:write permission lockstep', () => {
+  it('should be granted to exactly the roles this controller @Roles(admin) on every write route', () => {
+    // `inventory-locations:write` is a display predicate only (no permission
+    // guard exists) - if this drifts from the controller's actual @Roles,
+    // either add a permission-based guard or revert the grant, the
+    // `shipments:write` / `automations:write` discipline.
+    const rolesWithPermission = UserRoleValues.filter((role) =>
+      ROLE_PERMISSIONS[role].includes('inventory-locations:write')
+    );
+
+    expect(rolesWithPermission).toEqual(['admin']);
   });
 });
--- a/apps/web/src/app/nav-registry.ts
+++ b/apps/web/src/app/nav-registry.ts
@@ -71,6 +71,11 @@ export const BASE_NAV_GROUPS: readonly NavRegistryGroup[] = [
     label: 'Platform',
     items: [
       { to: '/connections', label: 'Connections', countKey: 'connections' },
+      // Reads are `admin`/`operator`/`viewer` (InventoryLocationsController)
+      // - gated on `inventory:read` rather than `inventory-locations:write`,
+      // so a `packer` (which holds no permissions at all) is the only role
+      // that never sees the entry.
+      { to: '/inventory/locations', label: 'Inventory locations', requiresPermission: 'inventory:read' },
       { to: '/adapters', label: 'Adapters' },
       { to: '/settings', label: 'Settings' },
     ],
--- a/apps/web/src/app/routes/inventory-locations.route.tsx
+++ b/apps/web/src/app/routes/inventory-locations.route.tsx
@@ -1,0 +1,13 @@
+import type { RouteObject } from 'react-router-dom';
+import type { RouteCrumbHandle } from '../nav-registry.types';
+
+export const inventoryLocationsRoute: RouteObject = {
+  path: 'inventory/locations',
+  handle: { crumb: { group: 'Platform', title: 'Inventory locations' } } satisfies RouteCrumbHandle,
+  lazy: async () => {
+    const { InventoryLocationsPage } = await import(
+      '../../pages/inventory/inventory-locations-page'
+    );
+    return { Component: InventoryLocationsPage };
+  },
+};
--- a/apps/web/src/app/routes/root.route.tsx
+++ b/apps/web/src/app/routes/root.route.tsx
@@ -24,6 +24,7 @@ import { cursorsRoute } from './cursors.route';
 import { customersRoute } from './customers.route';
 import { devUiRoute } from './dev-ui.route';
 import { insightsRoute } from './insights.route';
+import { inventoryLocationsRoute } from './inventory-locations.route';
 import { listingsRoute } from './listings.route';
 import { aiProviderSettingsRoute } from './ai-provider-settings.route';
 import { mcpTokensRoute } from './mcp-tokens.route';
@@ -71,6 +72,7 @@ export const coreChildren: RouteObject[] = [
   automationsRoute,
   invoicesRoute,
   connectionsRoute,
+  inventoryLocationsRoute,
   adaptersRoute,
   newConnectionRoute,
   advancedNewConnectionRoute,
--- a/apps/web/src/app/routes/route-lazy.test.ts
+++ b/apps/web/src/app/routes/route-lazy.test.ts
@@ -83,7 +83,7 @@ const lazyRoutes = collectLazyRoutes([
  *   - prompt-templates-legacy-redirects (inline `<Navigate>` element)
  *   - `/analytics` legacy alias (inline `<Navigate>` to `/`, #2740)
  */
-const EXPECTED_LAZY_ROUTE_COUNT = 63;
+const EXPECTED_LAZY_ROUTE_COUNT = 64;
 
 describe('route lazy contract', () => {
   it(`the registered route tree contains exactly ${EXPECTED_LAZY_ROUTE_COUNT} lazy routes`, () => {
--- a/apps/web/src/features/inventory/api/inventory-locations.types.ts
+++ b/apps/web/src/features/inventory/api/inventory-locations.types.ts
@@ -7,6 +7,16 @@
  * @module apps/web/src/features/inventory/api
  */
 
+/**
+ * `countryIso2` is normalised to uppercase in exactly one place — the API's
+ * own `ListLocationsQueryDto` does the same server-side, so this mirrors that
+ * rule rather than inventing a second one. #3135 review: was duplicated
+ * inline in `inventory.api.ts` and `location-dialog.schema.ts`.
+ */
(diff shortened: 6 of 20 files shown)


norbert-kulus-blockydevs:
## PR Review — #3133 (create/edit location dialog)

Reviewed as a frontend UI/UX pass, cross-checked against the real backend contract (`CreateLocationDto` / `UpdateLocationDto`) and against comparable create/edit forms already in the codebase.

### What's done well

- **`code` is correctly create-only.** `isEdit ? null : <FormField Code>` matches `UpdateLocationDto`'s own docblock ("renaming the natural key is not a patch-shaped operation") exactly, and it's the one thing explicitly asked for in the sibling `#3131` API-client review — confirmed consistent end-to-end. Tested both ways (create shows Code, edit hides it, patch payload has no `code` key).
- **Zod bounds match the backend DTO field-for-field**: `code` (64), `name`/`externalRef` (255), `postcode` (16), `countryIso2` (exact length 2), lat/lng (±90/±180). Nothing here is laxer than what the server would reject, so there's no bad create→400 round trip.
- **Accessibility comes for free, correctly.** Every field goes through the shared `FormField` primitive, which wires `htmlFor`/`id`, `aria-describedby`, and `aria-invalid` automatically — the dialog author doesn't have to (and doesn't) hand-roll any of that. `Dialog` is the Radix wrapper (focus trap + Escape dismissal already established per `docs/frontend-architecture.md`).
- **409/422 error mapping is sensible**: `DuplicateLocationCodeError` → field error on `code`, and the (self-admittedly "in practice unreachable") `ownerConnectionId` 422 is still handled defensively for a stale-cache race rather than ignored.
- **`target`-identity reset effect is correct** — re-opening the dialog for a different row (or immediately re-opening create) resets form + mutation state, avoiding stale-value leakage between edits.
- Omitting `status` from this form is consistent with the epic's split — retiring a location is `#3134`'s dialog, not this one.

### 🟡 IMPORTANT

**No `ReadOnlyLock` / `useWriteAccess` gating on the Save action — and this dialog *is* the actual mutation surface.**

`grep -rn "ReadOnlyLock\|useWriteAccess" apps/web/src/features/inventory/` returns nothing at all across the whole feature. Compare with the two closest analogues in the codebase:

- `CreateConnectionForm` / `EditConnectionForm` both wrap their submit button in `<ReadOnlyLock active={write.demoReadOnly} …>`, gated by `useWriteAccess('connections:write', demoMode)`.
- `generate-label-form.tsx` (the file this PR's own docblock cites as its `''`→`null` mapping precedent) has **no** `ReadOnlyLock` either — but it says so explicitly: `// No ReadOnlyLock exists in this chain (#1788) — fire directly on a real …`. That's a documented, deliberate exception.

This dialog has neither the wrapper nor a comment explaining its absence, so as written it reads as an oversight rather than a decision — and per `docs/frontend-architecture.md § Access Control`, demo mode is supposed to render a write affordance **disabled with a tooltip**, not simply omit the check. This is also more load-bearing than the gap already flagged in the sibling `#3135` list-page review (whose Add/Edit *entry points* skip `ReadOnlyLock`): even if the list page is fixed to gate its buttons, a demo-mode operator who reaches this dialog by another route (direct link, keyboard) can still submit a real mutation from here today.

Whichever gate is intended (`inventory:write` or similar — confirm the actual permission key used by the read side of this feature), it should wrap the `Save location` button the same way the connection forms do.

### 🟢 SUGGESTION

- No test exercises the 422 `ownerConnectionId` branch (`form.setError('ownerConnectionId', …)`). It's a small, self-contained branch, but per the PR's own comment it's reachable via a stale-cache race, and an untested error-mapping branch is exactly the kind of thing that silently regresses on a later refactor (mirrors a similar untested-branch note on sibling `#3134`).
- Minor: the generic mutation-error `<Alert>` condition (`!(mutation.error instanceof ApiError && (mutation.error.isConflict() || mutation.error.status === 422))`) duplicates the same branching logic already expressed in `onSubmit`'s catch block. Not wrong, but a small shared helper (`isFieldMappedError(error)`) would keep the two from drifting if a third mapped-error case is added later.

Core dialog logic, contract fidelity, and accessibility are solid. The `ReadOnlyLock` gap is the one item worth fixing before merge — everything else is a minor test-coverage suggestion.

piotrswierzy (commented):

Solid RHF+Zod dialog covering every `CreateLocationDto` field with the PATCH contract correctly respected. The substantive issue is not in this slice's code but in what it **omits**.

### IMPORTANT — `status` is absent from the edit form and from `toUpdateInput`, so a retired location can never be re-activated

`location-dialog.schema.ts` `toUpdateInput` (~99–110) + `location-dialog.tsx`

#3134 offers retire and its copy promises "it can be re-activated later" **twice**. Nothing in the stack delivers that: no `status` in the edit payload, no Reactivate action in #3135. Retire is a one-way door in the shipped product, plus a false statement about it.

Either add `status` to the edit dialog (a two-option `Select`) or a Reactivate row action on the list, **or** drop the promise from #3134's copy. Leaving both is the class of thing this codebase treats as a defect rather than a gap.

### SUGGESTIONS

- `code.toUpperCase()` in `toCreateInput` (~86) and `toUpdateInput` (~105) is a **second** normalisation site for the value `architecture-overview.md` § Inventory says is normalised *"in exactly ONE place — the application service"*. I would not block it: the server re-normalises, it is idempotent, and the field description discloses it — so it is display convenience, not a rival authority. But it is an **unguarded** mirror: if the server's normalisation ever widens (say, stripping internal whitespace), the field description becomes a lie with nothing failing. Drop it and let the returned row show the canonical value, or note in the docblock that the claim is asserted by nothing.
- `useConnectionsQuery()`'s loading/error states aren't surfaced (~326, 351) — if that read fails, "Owning connection" silently offers only "— none —", which reads as *"this install has no connections"*. One hint on the `FormField` would distinguish "none" from "couldn't load".
- A non-numeric latitude falls through `z.literal('')` into `z.coerce.number()` → `NaN` → the min/max check, so the operator is told *"Must be between -90 and 90"* for input that isn't a number. A `.refine(Number.isFinite)` with its own message reads better. (I checked: the ±90/±180 bounds are **not** swapped.)
- The `mutation.error` Alert guard (~404) repeats the 409/422 predicate inline in a ~130-char expression. `isFieldMappedError(error)` would make the rule readable — and #3134 hand-rolls the same idea again.

### Done well

The `code`-on-create-only rule is respected exactly as the architecture doc states, and the test asserts it in **both** directions — `queryByLabelText('Code')` absent on edit *and* `'code' in patch === false` on submit. Omitting the field rather than rendering-and-dropping it is the right call. Form state is RHF+Zod with server validation still authoritative; the 409→field-error mapping keeps the dialog open (asserted). The `target`-identity `useEffect` reset with its rationale comment is correct for the back-to-back-edit case.

---

**1. Retire is a one-way door, created across two slices.** #3134 offers retire and promises reactivation **twice**; #3133 omits `status` from `toUpdateInput`; #3135 renders no Reactivate action. No slice looks wrong alone — which is why it needs saying at stack level. **This is the one thing I would hold the stack on.**

**2. `Closes #N` will not fire on any of the five.** All target feature branches (#3131→`3063`, #3132→`3064`, #3133→`3065`, #3134→`3067`, #3135→`3068`), and GitHub auto-closes only on merge to the **default** branch while `CLAUDE.md` forbids hand-closing. Roll the trailers onto the epic PR (`3063-…` → `main`, which does not exist yet).

**3. `countryIso2` uppercasing exists in three FE places** — `buildLocationsQuery` (#3131), `toCreateInput` and `toUpdateInput` (#3133) — plus server-side. Normalise once; the API client is the natural choke point.

**4. Two "don't double-render a mapped error" implementations** — #3133's inline Alert guard and #3134's `genericError`, same rule, adjacent files, same stack. One helper.

Confirmed clean at stack level: the `inventory` slug is **already registered** in both `no-restricted-imports` groups for all five canonical subdirectories, so **no ESLint registration is owed**. The barrel grows monotonically — no slice imports a symbol its own base does not yet export (checked each). No `fetch()` outside the api module, no raw hex, no `any`, `pages → features → shared` respected throughout.

jakubretajczykBD:
## `/tech-review` — follow-up on review fixes

Reviewed the two fixes just pushed (commit 6dc51a58d) addressing the earlier tech-review suggestions:

### Summary
✅ Correct — both fixes are minimal, targeted, and preserve existing behavior on the happy path.

### Issues

**[SUGGESTION]** — `location-dialog.schema.ts` (lat/lng fields)
No regression test currently pins "a non-numeric latitude renders 'Must be a number', not the old range message" — exactly the defect just fixed. Worth a one-line assertion in `location-dialog.test.tsx` to guard it going forward.

**[SUGGESTION]** — `location-dialog.tsx:198-206`
The nested ternary for the owning-connection field's `description` works but reads a bit dense. A local `const ownerFieldHint = ...` above the JSX would be slightly cleaner — purely cosmetic, not blocking.

No BLOCKING or IMPORTANT issues:
- No architecture/boundary violation, no new API surface.
- `FormField`/Zod schema conventions preserved exactly as used elsewhere in this file.
- No existing test assertions broken (verified — nothing in `location-dialog.test.tsx` pinned the old description string or old error message).
- `type-check` clean; `eslint` on `apps/web` clean (0 errors, only pre-existing unrelated warnings elsewhere in the repo).

### Verdict
✅ **Approve** — fixes match the earlier review suggestions verbatim. The two SUGGESTIONS above are optional polish only.

PR #3134, opened 2026-09-10
feat(inventory): delete confirm incl. "location in use" 409 + retire action

### LINKED ISSUE #3068: feat(inventory): delete confirm incl. "location in use" 409 + retire action

jakubretajczykBD:
## Summary
- Adds `LocationDeleteDialog`: starts in a plain confirm state (there's no cheap way to know ahead of the attempt whether stock references the location — the list read carries no such signal today), and on a 409 (`LocationInUseError`) swaps its own body and confirm action to **"Retire instead"** (`PATCH { status: 'inactive' }`) rather than closing and opening a second dialog
- Retire is styled as a corrective, non-destructive action (default/primary tone), deliberately not danger-red like the delete it replaces

Stacked on #3067 (#3133). Base branch is `3067-location-dialog`.

Note on epic build order: same reasoning as #3067 — built ahead of #3066 (list page) so its Delete action has a real dialog to open.

Closes #3068.

## Test plan
- [x] 5 new tests: closed-when-null, successful delete (no reference), 409→in-use explanation (and no duplicate raw-error alert), successful retire-from-in-use, phase resets when reopened for a different location — all green

### DIFF
--- a/apps/api/src/inventory/http/inventory-locations.controller.spec.ts
+++ b/apps/api/src/inventory/http/inventory-locations.controller.spec.ts
@@ -17,6 +17,7 @@ import {
   LocationInUseError,
   type ILocationService,
 } from '@openlinker/core/inventory';
+import { ROLE_PERMISSIONS, UserRoleValues } from '@openlinker/core/users';
 import { InventoryLocationsController } from './inventory-locations.controller';
 
 function makeLocation(
@@ -221,5 +222,19 @@ describe('InventoryLocationsController', () => {
       expect(result.created).toEqual([]);
       expect(result.existingCodes).toEqual(['MAIN']);
     });
+  });
+});
+
+describe('inventory-locations:write permission lockstep', () => {
+  it('should be granted to exactly the roles this controller @Roles(admin) on every write route', () => {
+    // `inventory-locations:write` is a display predicate only (no permission
+    // guard exists) - if this drifts from the controller's actual @Roles,
+    // either add a permission-based guard or revert the grant, the
+    // `shipments:write` / `automations:write` discipline.
+    const rolesWithPermission = UserRoleValues.filter((role) =>
+      ROLE_PERMISSIONS[role].includes('inventory-locations:write')
+    );
+
+    expect(rolesWithPermission).toEqual(['admin']);
   });
 });
--- a/apps/web/src/app/nav-registry.ts
+++ b/apps/web/src/app/nav-registry.ts
@@ -71,6 +71,11 @@ export const BASE_NAV_GROUPS: readonly NavRegistryGroup[] = [
     label: 'Platform',
     items: [
       { to: '/connections', label: 'Connections', countKey: 'connections' },
+      // Reads are `admin`/`operator`/`viewer` (InventoryLocationsController)
+      // - gated on `inventory:read` rather than `inventory-locations:write`,
+      // so a `packer` (which holds no permissions at all) is the only role
+      // that never sees the entry.
+      { to: '/inventory/locations', label: 'Inventory locations', requiresPermission: 'inventory:read' },
       { to: '/adapters', label: 'Adapters' },
       { to: '/settings', label: 'Settings' },
     ],
--- a/apps/web/src/app/routes/inventory-locations.route.tsx
+++ b/apps/web/src/app/routes/inventory-locations.route.tsx
@@ -1,0 +1,13 @@
+import type { RouteObject } from 'react-router-dom';
+import type { RouteCrumbHandle } from '../nav-registry.types';
+
+export const inventoryLocationsRoute: RouteObject = {
+  path: 'inventory/locations',
+  handle: { crumb: { group: 'Platform', title: 'Inventory locations' } } satisfies RouteCrumbHandle,
+  lazy: async () => {
+    const { InventoryLocationsPage } = await import(
+      '../../pages/inventory/inventory-locations-page'
+    );
+    return { Component: InventoryLocationsPage };
+  },
+};
--- a/apps/web/src/app/routes/root.route.tsx
+++ b/apps/web/src/app/routes/root.route.tsx
@@ -24,6 +24,7 @@ import { cursorsRoute } from './cursors.route';
 import { customersRoute } from './customers.route';
 import { devUiRoute } from './dev-ui.route';
 import { insightsRoute } from './insights.route';
+import { inventoryLocationsRoute } from './inventory-locations.route';
 import { listingsRoute } from './listings.route';
 import { aiProviderSettingsRoute } from './ai-provider-settings.route';
 import { mcpTokensRoute } from './mcp-tokens.route';
@@ -71,6 +72,7 @@ export const coreChildren: RouteObject[] = [
   automationsRoute,
   invoicesRoute,
   connectionsRoute,
+  inventoryLocationsRoute,
   adaptersRoute,
   newConnectionRoute,
   advancedNewConnectionRoute,
--- a/apps/web/src/app/routes/route-lazy.test.ts
+++ b/apps/web/src/app/routes/route-lazy.test.ts
@@ -83,7 +83,7 @@ const lazyRoutes = collectLazyRoutes([
  *   - prompt-templates-legacy-redirects (inline `<Navigate>` element)
  *   - `/analytics` legacy alias (inline `<Navigate>` to `/`, #2740)
  */
-const EXPECTED_LAZY_ROUTE_COUNT = 63;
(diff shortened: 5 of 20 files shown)


norbert-kulus-blockydevs:
## Tech-lead review — frontend focus (delete-confirm dialog)

**Summary:** Clean, well-scoped implementation of the "delete → 409 → retire" flow. Correctly reuses `ConfirmDialog`/`Alert`/`useToast`, drives both mutations through TanStack Query hooks (no ad-hoc fetch), never renders backend logic, and the 409 path degrades into a clear, actionable "Retire instead" affordance rather than a raw error toast — exactly what the domain contract (`docs/architecture-overview.md` § Locations CRUD #2316) calls for. No backend/boundary violations found.

**Verdict: 🔄 Approve with changes** (non-blocking nit + one real coverage gap worth closing before/soon after merge)

### IMPORTANT
- **The `genericError` branch (the actual non-409 failure path) is completely untested.** `location-delete-dialog.tsx`'s ternary — `deleteMutation.error && !isConflict ? deleteMutation.error : retireMutation.error ?? null` — is the component's only logic for surfacing a real infra/5xx failure on either mutation, and none of the 5 new tests exercise a non-conflict `deleteLocation` rejection or a failed `updateLocation` (retire) call. Given the review focus here is precisely "does the error path degrade gracefully," this is the one path that isn't proven — worth a `mockRejectedValue(new ApiError('...', 500, {}))` test for at least the delete side (and ideally the retire-fails-from-in-use case too, since `retireMutation.error` is the one arm of the ternary nothing currently drives).

### SUGGESTION
- **The reset effect keys on `location` (object identity), not `location?.id`.** The sibling `LocationDialog` (#3067) does the same but documents *why* it's safe there: `target` is a freshly-minted wrapper object at open-time. `location` here is the row itself, with no such wrapper — if a future list page ever derives it reactively (e.g. `locations.find(l => l.id === id)` off query data that refetches on window focus after the 30s `staleTime`), a new-but-equal object would silently reset an in-flight `in-use`/retire decision back to plain "Delete?" with no explanation. Cheap to harden now: key the effect off `location?.id` and/or add the same clarifying comment `LocationDialog` carries. Not blocking since the current consumer doesn't exist yet and the established pattern in this epic uses a stable snapshot, but worth doing while it's fresh.
- **Tone/politeness of the "in-use" explanation is a judgment call worth a second look.** `Alert tone="warning"` renders `role="status"` (polite) per the shared component's own docblock reasoning ("the work still completed"), but here the delete was actually *refused* — nothing completed. It's still announced (role=status implies aria-live=polite), so this isn't an a11y gap, just worth confirming the polite-vs-assertive choice was deliberate rather than a default carried over from a different semantic case.

No BLOCKING findings.

piotrswierzy (commented):

**The central design decision is right.** Starting in `confirm` and learning `in-use` only from the failed DELETE is the correct read of `architecture-overview.md` § Inventory (#2316): the refusal lives in `LocationService.deleteLocation` because `inventory_items` carries **no FK**, so there is no cheap pre-check and any client-side prediction would be a guess. Swapping the *same* dialog's body and confirm action rather than closing and reopening a second one keeps the remedy where the operator already is.

Retire being `tone: 'default'` rather than `danger` is the right severity signal and is explained in place. ADR-058 decision 2 is clean — no copy here implies `locationId IS NULL` is a default location, and *"existing positions keep pointing at a row that exists"* is accurate.

### IMPORTANT — "it can be re-activated later" is stated twice and deliverable nowhere

`location-delete-dialog.tsx` (~195 toast, ~217 description)

`toUpdateInput` in #3133 omits `status`, and #3135 renders no Reactivate action — so retire is irreversible in the shipped product while this dialog promises otherwise, twice. Make it true, or stop saying it.

### SUGGESTIONS

- The in-use explanation renders **twice** (~214–233): once as the `ConfirmDialog` `description` and again as a `warning` Alert in `body`, with "Stock still points here" appearing as both the alert title and, in substance, the description. One is enough; the duplication reads as a rendering bug rather than emphasis.
- `genericError` (~205–208) re-derives the "not a mapped 409" predicate #3133 also computes inline. Shared helper.

### Done well

The test that a refused delete does **not** additionally render the raw domain error as a generic alert (~68–71) is exactly the assertion I would have asked for, and the phase-reset-on-reopen test closes the stale-state hole.

---

**1. Retire is a one-way door, created across two slices.** #3134 offers retire and promises reactivation **twice**; #3133 omits `status` from `toUpdateInput`; #3135 renders no Reactivate action. No slice looks wrong alone — which is why it needs saying at stack level. **This is the one thing I would hold the stack on.**

**2. `Closes #N` will not fire on any of the five.** All target feature branches (#3131→`3063`, #3132→`3064`, #3133→`3065`, #3134→`3067`, #3135→`3068`), and GitHub auto-closes only on merge to the **default** branch while `CLAUDE.md` forbids hand-closing. Roll the trailers onto the epic PR (`3063-…` → `main`, which does not exist yet).

**3. `countryIso2` uppercasing exists in three FE places** — `buildLocationsQuery` (#3131), `toCreateInput` and `toUpdateInput` (#3133) — plus server-side. Normalise once; the API client is the natural choke point.

**4. Two "don't double-render a mapped error" implementations** — #3133's inline Alert guard and #3134's `genericError`, same rule, adjacent files, same stack. One helper.

Confirmed clean at stack level: the `inventory` slug is **already registered** in both `no-restricted-imports` groups for all five canonical subdirectories, so **no ESLint registration is owed**. The barrel grows monotonically — no slice imports a symbol its own base does not yet export (checked each). No `fetch()` outside the api module, no raw hex, no `any`, `pages → features → shared` respected throughout.

jakubretajczykBD:
## `/tech-review` — follow-up on the review findings — ✅ Approve

Verified against the two prior reviews after 29730600e (pushed to `3068-location-delete-dialog`). 4-file fix commit, no unrelated changes. `type-check` and `eslint` on the changed files are clean.

| Finding | Status |
|---|---|
| **IMPORTANT** (Piotr) — "can be re-activated later" promised twice, delivered nowhere | ✅ Fixed — toast (`"Existing positions keep pointing at it."`) and Alert (`"Retiring keeps the row and its history intact — nothing is deleted."`) now state only what actually happens. |
| **IMPORTANT** (Norbert) — `genericError` branch (real infra failure) had zero test coverage | ✅ Fixed — two new tests: non-conflict 500 on delete (stays in confirm phase, no accidental retire-mode switch), and a failed retire from the in-use phase (dialog stays open, `onClose` not called). |
| **SUGGESTION** (Piotr) — in-use explanation rendered twice, near-duplicate wording | ✅ Fixed — `description` now states *why* (refusal reason), the Alert states *what retiring does*; a comment pins the intended split. |
| **SUGGESTION** (Piotr) — shared helper for "don't double-render a mapped error" | ✅ Fixed — `isUnmappedApiError(error, isMapped)` added to `shared/api/api-error.ts`, consumed by both `LocationDeleteDialog` and `LocationDialog`, replacing two independent inline copies of the same predicate. |
| **SUGGESTION** (Norbert) — reset effect keyed on `location` object identity | ✅ Fixed — keyed on `location?.id`, with the same clarifying comment `LocationDialog` already carries for its analogous case. |
| **SUGGESTION** (Norbert) — `Alert tone="warning"` for a refusal — deliberate? | Left as-is — judgment call, not a defect. |
| Piotr's stack-level items (`countryIso2` dedup, `Closes #N` targeting) | `countryIso2` was already fixed in the prior commit; the rest are cross-PR/process concerns outside this diff's scope. |

No BLOCKING or IMPORTANT issues remain. One non-blocking note: `isUnmappedApiError` is the first standalone helper function in `shared/api/api-error.ts` (previously class-only) — reasonable placement, worth keeping in mind if a third mutation dialog wants the same pattern.

**Verdict: ✅ Approve**

Issue #1996, opened 2026-08-03
[TASK] Frontend — one shared Order cell and one shared Connection cell across six lists


martasl00:
## Problem / Context

Six list pages render the same two facts six different ways, so an operator cannot recognise a record
without opening its detail page. The reported symptom: *"the Orders column is inconsistent across the
system — it looks different on every tab and I can't tell which order it is."* The same is true of the
Connection column, and of the product identifier line.

**The Order cell today:**

| List | File | Today |
|---|---|---|
| Orders | `pages/orders/orders-list-page.tsx` (column `order`, :576-651) | `EntityLabel` (order number + short id + Copy) + first-item name + `+N` — but **no thumbnail** |
| Shipments | `pages/shipments/shipments-page.tsx` (column `orderId`, :296-303, plus a separate mobile-card branch :584-591) | `EntityLabel` with `truncateOrderId()` — **no thumbnail, no item name, no `+N`** |
| Invoices | `pages/invoicing/invoices-list-page.tsx` (column `orderId`, :225-230) | raw full UUID in a `mono-text` span — **no truncation, no Copy, no link, nothing else** |

**The Connection cell today** — five variants of one idea, none of them the same:

| List | Today |
|---|---|
| Products (`source`, :687-701) | channel pill + bare connection name. No id, no Copy |
| Shipments (`connectionId`, :363-380) | `ConnectionDot` glyph + bare name. No id, no Copy |
| Invoices (`connection`, :282-286) | name-or-raw-UUID in `text-muted`, UUID in the `title`. No id chip, no Copy |
| Listings (`connectionId`, :56) | bare UUID in `mono-text` |
| Customers (`lastSourceConnectionId`, :44-48) | bare UUID in `mono-text` |

**Three helpers already duplicate one algorithm** and will keep drifting: `EntityLabel`'s private
`shortenId()` (`shared/ui/entity-label.tsx:98-109`), `truncateOrderId()`
(`features/shipments/lib/shipment-severity.ts:25-34`, whose own comment documents why it exists), and
`formatOrderRef()` (local to `orders-list-page.tsx:195-199`).

**Channel names render lowercase** because a hardcoded page-local map competes with the plugin
registry. `CHANNEL_LABELS` (`orders-list-page.tsx:86-91`) covers only allegro / prestashop / amazon /
shopify — so `erli` and `woocommerce` render raw in Orders too, not just in Listings, which prints
[...]

PR #3135, opened 2026-09-10
feat(inventory): locations list page with filter bar

### LINKED ISSUE #3066: feat(inventory): locations list page with filter bar

jakubretajczykBD:
## Summary
- Adds `InventoryLocationsPage`: `DataTable` (code+name, kind, country/postcode, owning connection, status, actions), a kind filter `Select` and a "Show retired" toggle defaulting **ON** (soft retirement keeps a row visible by design, #2316), loading/error/empty states following the `connections-list-page.tsx` cockpit-list composition
- The owning-connection cell reuses `ConnectionCell` with a channel-pill adornment — the same composition `products-list-page.tsx` already uses for a connection's platform identity
- Wires in the #3067/#3068 dialogs (`LocationDialog`, `LocationDeleteDialog`) for Add/Edit/Delete/Retire
- Adds a dedicated `inventory-locations:write` permission (backend + frontend) rather than reusing the existing `inventory:write` — that one is also held by `operator`, but `InventoryLocationsController`'s three write routes are `@Roles('admin')` only, so granting `inventory:write` here would show an operator an enabled button that then 403s. A controller spec asserts the permission is held by exactly `['admin']`, the `shipments:write`/`automations:write` discipline already established in this codebase.
- Two small reusable CSS additions: `.table-actions` (already referenced by three existing pages with no rule backing it — this fixes that gap for all of them) and `.toolbar__checkbox-label`

Stacked on #3068 (#3134), whose dialogs this page mounts. Base branch is `3068-location-delete-dialog`.

Closes #3066.

## Test plan
- [x] `pnpm --filter @openlinker/api type-check`, `pnpm --filter @openlinker/core type-check`
- [x] Backend: `inventory-locations.controller.spec.ts` (13/13, including the new permission-lockstep test) + `role.types.spec.ts` (20/20, no drift) + `use-permission.test.ts` (12/12, no drift)
- [x] Frontend: 12 new page tests (heading, filters, data display, loading/error/empty states incl. pre-bootstrap empty state, toggle→query mapping, kind-filter→query mapping, write-gating hidden without permission, Add/Edit/Delete opening the right dialogs) — all green
- [x] Full `features/inventory` + `pages/inventory` suite: 42/42 green

### DIFF
--- a/apps/api/src/inventory/http/inventory-locations.controller.spec.ts
+++ b/apps/api/src/inventory/http/inventory-locations.controller.spec.ts
@@ -17,6 +17,7 @@ import {
   LocationInUseError,
   type ILocationService,
 } from '@openlinker/core/inventory';
+import { ROLE_PERMISSIONS, UserRoleValues } from '@openlinker/core/users';
 import { InventoryLocationsController } from './inventory-locations.controller';
 
 function makeLocation(
@@ -221,5 +222,19 @@ describe('InventoryLocationsController', () => {
       expect(result.created).toEqual([]);
       expect(result.existingCodes).toEqual(['MAIN']);
     });
+  });
+});
+
+describe('inventory-locations:write permission lockstep', () => {
+  it('should be granted to exactly the roles this controller @Roles(admin) on every write route', () => {
+    // `inventory-locations:write` is a display predicate only (no permission
+    // guard exists) - if this drifts from the controller's actual @Roles,
+    // either add a permission-based guard or revert the grant, the
+    // `shipments:write` / `automations:write` discipline.
+    const rolesWithPermission = UserRoleValues.filter((role) =>
+      ROLE_PERMISSIONS[role].includes('inventory-locations:write')
+    );
+
+    expect(rolesWithPermission).toEqual(['admin']);
   });
 });
--- a/apps/web/src/app/nav-registry.ts
+++ b/apps/web/src/app/nav-registry.ts
@@ -71,6 +71,11 @@ export const BASE_NAV_GROUPS: readonly NavRegistryGroup[] = [
     label: 'Platform',
     items: [
       { to: '/connections', label: 'Connections', countKey: 'connections' },
+      // Reads are `admin`/`operator`/`viewer` (InventoryLocationsController)
+      // - gated on `inventory:read` rather than `inventory-locations:write`,
+      // so a `packer` (which holds no permissions at all) is the only role
+      // that never sees the entry.
+      { to: '/inventory/locations', label: 'Inventory locations', requiresPermission: 'inventory:read' },
       { to: '/adapters', label: 'Adapters' },
       { to: '/settings', label: 'Settings' },
     ],
--- a/apps/web/src/app/routes/inventory-locations.route.tsx
+++ b/apps/web/src/app/routes/inventory-locations.route.tsx
@@ -1,0 +1,13 @@
+import type { RouteObject } from 'react-router-dom';
+import type { RouteCrumbHandle } from '../nav-registry.types';
+
+export const inventoryLocationsRoute: RouteObject = {
+  path: 'inventory/locations',
+  handle: { crumb: { group: 'Platform', title: 'Inventory locations' } } satisfies RouteCrumbHandle,
+  lazy: async () => {
+    const { InventoryLocationsPage } = await import(
+      '../../pages/inventory/inventory-locations-page'
+    );
+    return { Component: InventoryLocationsPage };
+  },
+};
--- a/apps/web/src/app/routes/root.route.tsx
+++ b/apps/web/src/app/routes/root.route.tsx
@@ -24,6 +24,7 @@ import { cursorsRoute } from './cursors.route';
 import { customersRoute } from './customers.route';
 import { devUiRoute } from './dev-ui.route';
 import { insightsRoute } from './insights.route';
+import { inventoryLocationsRoute } from './inventory-locations.route';
 import { listingsRoute } from './listings.route';
 import { aiProviderSettingsRoute } from './ai-provider-settings.route';
 import { mcpTokensRoute } from './mcp-tokens.route';
@@ -71,6 +72,7 @@ export const coreChildren: RouteObject[] = [
   automationsRoute,
   invoicesRoute,
   connectionsRoute,
+  inventoryLocationsRoute,
   adaptersRoute,
   newConnectionRoute,
   advancedNewConnectionRoute,
--- a/apps/web/src/app/routes/route-lazy.test.ts
+++ b/apps/web/src/app/routes/route-lazy.test.ts
@@ -83,7 +83,7 @@ const lazyRoutes = collectLazyRoutes([
  *   - prompt-templates-legacy-redirects (inline `<Navigate>` element)
  *   - `/analytics` legacy alias (inline `<Navigate>` to `/`, #2740)
  */
-const EXPECTED_LAZY_ROUTE_COUNT = 63;
+const EXPECTED_LAZY_ROUTE_COUNT = 64;
 
 describe('route lazy contract', () => {
   it(`the registered route tree contains exactly ${EXPECTED_LAZY_ROUTE_COUNT} lazy routes`, () => {
--- a/apps/web/src/features/inventory/api/inventory-locations.types.ts
+++ b/apps/web/src/features/inventory/api/inventory-locations.types.ts
@@ -7,6 +7,16 @@
  * @module apps/web/src/features/inventory/api
  */
 
+/**
+ * `countryIso2` is normalised to uppercase in exactly one place — the API's
+ * own `ListLocationsQueryDto` does the same server-side, so this mirrors that
+ * rule rather than inventing a second one. #3135 review: was duplicated
+ * inline in `inventory.api.ts` and `location-dialog.schema.ts`.
+ */
+export function normalizeCountryIso2(value: string): string {
+  return value.toUpperCase();
+}
+
 /** What a location physically is (#2316). Operator-declared, never derived. */
 export const InventoryLocationKindValues = ['warehouse', 'store', 'third-party', 'virtual'] as const;
 export type InventoryLocationKind = (typeof InventoryLocationKindValues)[number];
--- a/apps/web/src/features/inventory/api/inventory.api.ts
+++ b/apps/web/src/features/inventory/api/inventory.api.ts
@@ -12,6 +12,7 @@ import type {
   PaginatedInventory,
   InventoryAvailabilityResponse,
 } from './inventory.types';
+import { normalizeCountryIso2 } from './inventory-locations.types';
 import type {
   CreateInventoryLocationInput,
   InventoryLocation,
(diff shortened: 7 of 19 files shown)


norbert-kulus-blockydevs:
## Tech-lead review

New `InventoryLocationsPage` (filter bar + `DataTable` + CRUD dialogs) plus a correctly-scoped `inventory-locations:write` permission. Solid work overall; one real functional gap and one internal inconsistency worth fixing.

### IMPORTANT

1. **List pagination is silently dropped.** `useInventoryLocationsQuery(filters)` is called with no `pagination` argument, so it always hits the API's `page=1&limit=25` default (`inventory-locations.controller.ts:89`), and `PaginatedInventoryLocations.total` is never read anywhere in the page. Unlike `products-list-page.tsx` (which wires `ListPagination` off the same paginated-total shape), this page has zero UI signal that more than 25 rows might exist — no "showing 25 of N", no next-page control. The stated precedent, `connections-list-page.tsx`, isn't actually comparable here: `GET /connections` returns an unpaginated `Connection[]`, so it never faced this problem. On an install with >25 locations (multi-warehouse + 3PL + stores is exactly this feature's use case), rows past #25 are permanently unreachable through this page with no indication anything is missing.

2. **Inconsistent `ReadOnlyLock` usage within the same file.** The header "Add location" button and the row Edit/Delete buttons are correctly wrapped in `<ReadOnlyLock active={write.demoReadOnly} ...>` to explain *why* they're disabled in demo mode. The empty-state's "Create first location" and "+ Add manually" buttons use the identical `disabled={write.demoReadOnly}` condition but skip the `ReadOnlyLock` wrapper entirely — so a demo viewer hitting the zero-locations empty state gets disabled buttons with no explanatory tooltip, unlike everywhere else on the page.

### Backend/security lens (role.types.ts) — looks correct

- `inventory-locations:write` is added only to the flat `PermissionValues` array; `ROLE_PERMISSIONS.operator`/`viewer`/`packer` are untouched, so via `admin: PermissionValues` the grant lands on `admin` only — verified against the controller, whose three write routes (`POST`, `PATCH`, `DELETE`) are indeed `@Roles('admin')` only, while reads are `@Roles('admin','operator','viewer')`. No route authorization was loosened.
- The new lockstep spec (`ROLE_PERMISSIONS[role].includes('inventory-locations:write')` must equal `['admin']`) is a good guardrail against future drift between the display permission and the actual guard, and the existing `role.types.spec.ts:104` (`admin` permissions == `PermissionValues`) still passes untouched.

### SUGGESTION

- Minor: the `KIND_LABEL` re-export from `features/inventory/index.ts` points at `location-dialog.schema.ts`, a slightly odd home for a page-level lookup table, but it's pre-existing (from #3067) and reuse over duplication is the right call here.

piotrswierzy (commented):

The largest and most complete slice: 12 page tests plus three genuine end-to-end CRUD round trips through the real mutation hooks (create appears, delete disappears, 409→retire flips the row), correct permission modelling with a backend lockstep assertion, and correct `ConnectionCell` usage.

### IMPORTANT — the empty-state copy asserts something untrue about the operator's orders

`inventory-locations-page.tsx` (~858)

> "Zero locations exist — every order is currently unfulfillable."

**This is false on essentially every install, and false in both directions.** Per #2407, `ConnectionService` *refuses* the `false → true` fulfilment-routing transition when zero locations exist, and routing is opt-in via `parseAuthorityConfig`'s frozen-unheld default. So an install with zero locations either has routing off — orders fall back to today's handling and are perfectly fulfillable — or could never have turned it on.

The sentence tells an operator their orders are broken when nothing is. Reword to the state that is actually true (*"Fulfilment routing has nowhere to source stock from yet, so it cannot be enabled"*), and per #2407 say that a minted location still needs stock assigned rather than implying the mint makes routing succeed.

### SUGGESTIONS

- `index.css` (~3000) adds a rule for `.table-actions`, a class **already emitted** by `users-page.tsx`, `invoices-list-page.tsx` and `prompt-template-detail-page.tsx`. It is a real fix, but it changes the rendered layout of three pages this PR does not otherwise touch (unstyled inline flow → `display: flex; gap; flex-wrap`) with no assertion or screenshot on any of them. Call it out in the body as an intentional cross-page visual change, or split it.
- `EmptyValue label="No location set"` on the `geo` column (~732) reads as *"this location has no location"* on a row of the locations table. It means no country/postcode — say that.
- `StatusBadge` (~768) renders the raw wire value (`active`/`inactive`), asserted as such. Every other badge surface in this app renders operator copy, "Retired" is the word the rest of this stack uses for `inactive`, and the toggle beside it says "Show retired". A `STATUS_LABEL` map beside `KIND_LABEL` would keep one vocabulary.

### Done well — several non-obvious

- **The dedicated `inventory-locations:write` permission is the right call and correctly reasoned.** `inventory:write` is held by `operator` (verified in `ROLE_PERMISSIONS`), while `InventoryLocationsController`'s writes are `@Roles('admin')` only — reusing it would show an operator an enabled button that then 403s. `admin: PermissionValues` makes the grant automatic, and the new controller-spec lockstep test (`toEqual(['admin'])`) is the `shipments:write` / `automations:write` discipline applied correctly rather than cargo-culted.
- **The nav item gates on `inventory:read`, not the write permission** — correct, since reads are admin/operator/viewer; gating on write would hide the page from the roles allowed to read it.
- **`ConnectionCell` is used exactly as its own docblock demands** — `map.get(id) ?? null` plus `loading={connectionsQuery.isLoading}`, which is the documented way to avoid silently reinstating the per-row fetch #1996 rejected. Passing `undefined` would have been a per-row N+1 that type-checks.
- **The `use-bootstrap-locations-mutation.ts` widening** (`activeLocations()` → `locationsAll()`) is a genuine bug fix with a regression test that would fail against the old key.
- Async-UX baseline fully covered: skeleton, `ErrorState` with Retry, two distinct empty states (filtered vs pre-bootstrap), success. URL state (`kind`, `retired`, `page`) in search params per the state-ownership rules; `readPageParam` guards `?page=abc`; a filter change resets to page 1. `route-lazy.test.ts` count bumped; the route carries its `handle.crumb`.
- ADR-052: the page description (*"A location isn't authority over stock"*) and the `ownerConnectionId` column correctly treat ownership as **provenance, never authority**.
- The test file's own `#3135 review` comments show a prior round already caught dropped pagination and a missing `ReadOnlyLock` on the empty-state buttons — both now fixed *and* pinned.

---

**1. Retire is a one-way door.** #3134 promises reactivation twice; #3133 omits `status` from `toUpdateInput`; this page renders no Reactivate action. A **Reactivate row action here is arguably the cleanest home for it.** This is the one thing I would hold the stack on.

**2. `Closes #N` will not fire on any of the five** — all target feature branches, GitHub auto-closes only on merge to the default branch, and `CLAUDE.md` forbids hand-closing. Roll the trailers onto the epic PR (`3063-…` → `main`, which does not exist yet).

**3. `countryIso2` uppercasing exists in three FE places** (#3131's `buildLocationsQuery`, #3133's two mappers) plus server-side. Normalise once.

**4. Two "don't double-render a mapped error" implementations** (#3133 inline, #3134's `genericError`). One helper.

Confirmed clean: the `inventory` slug is already registered in both `no-restricted-imports` groups for all five canonical subdirectories, so **no ESLint registration is owed**. The barrel grows monotonically — no slice imports a symbol its own base doesn't yet export.

jakubretajczykBD:
## Tech Review

### Summary

This is a well-executed, capability/permission-gated list page that closely follows the established `connections-list-page.tsx` / `users-page.tsx` cockpit-list conventions (URL-state filters, React Query server state, local dialog state, `ReadOnlyLock`/`AccessGate` demo-mode discipline). The commit history shows the two IMPORTANT findings from an earlier tech-review pass (dropped pagination, inconsistent `ReadOnlyLock` wrapping) were already fixed, plus several follow-on polish fixes (page-out-of-range empty state, active-connection-only owner picker, status-label rendering, `countryIso2` normalization dedup). I verified the backend/frontend permission split against the actual controller `@Roles` guards and the `ROLE_PERMISSIONS` map — the `inventory-locations:write` vs `inventory:write` distinction is correct and lockstep-tested. I did not find any remaining architecture or standards violations; only a couple of minor polish items remain.

### Issues

**[SUGGESTION]** — `apps/web/src/pages/inventory/inventory-locations-page.tsx` (`rows={[...(query.data?.items ?? [])]}`)
> This spreads a `readonly` array into a fresh array on every render, not only when `query.data` changes (e.g. it re-executes when `dialogTarget`/`deleteTarget` local state changes too). Not incorrect, just an avoidable reallocation — wrap in `useMemo(() => [...(query.data?.items ?? [])], [query.data])` if `DataTable` ever starts relying on row-array identity for virtualization/memoization. Not worth blocking on.

**[SUGGESTION]** — `apps/web/src/pages/inventory/inventory-locations-page.tsx` (`pageOutOfRange` message)
> `There are ${total} locations, but not on page ${page}.` renders the *filtered* total when a filter is also active (e.g. `kind=warehouse` narrowed the set further than the stale page number). The copy is still technically accurate but could read as ambiguous about whether `total` is filtered or unfiltered. Optional copy tweak only — not a functional bug, since `handleFilterChange` already resets `page` to 1 on any filter change, so this state is reachable only via a hand-edited URL or a concurrent delete on an unfiltered set in practice.

**[SUGGESTION]** — `apps/web/src/features/inventory/api/inventory-locations.types.ts` (`normalizeCountryIso2`)
> Good dedup of the three-way duplicated `.toUpperCase()` call per the #3135 review note. One nit: it lives in a `*.types.ts` file even though it's a plain runtime function with no type attached to it — this doesn't fit the engineering-standards "pure-rule exception" for `*.types.ts` (that exception is for a function that IS the coercion/validation rule *for* a type declared in the same file, not a general string helper). It's trivial enough that this is not worth blocking on, but a `lib/` (or a tiny `country.ts`) home would be the more literal fit if this grows further.

No BLOCKING or IMPORTANT issues found. Specifically verified and clean:
- Permission split (`inventory-locations:write` vs `inventory:write`) matches the controller's actual `@Roles('admin')` gates on the three write routes, matches `ROLE_PERMISSIONS` (admin-only via the `PermissionValues` spread), and is covered both by the new controller-side lockstep test and the pre-existing `check-permission-mirror.mjs` FE/BE mirror gate.
- Query-key invalidation fix (`locationsAll()` vs `activeLocations()`) is correct against the actual key-factory prefix structure.
- Pagination wiring matches the hook's real signature (`filters`, `pagination`) and the API's same-response `total`, correctly avoiding the two-stage-total machinery (appropriately reserved for expensive counts, which this small operator-authored table is not).
- State ownership is correctly distributed: URL for filters/pagination, React Query for server state, local `useState` for dialog targets — no violation of `docs/frontend-architecture.md` state rules.
- Dependency direction, feature-barrel-only imports, `kebab-case.tsx`/`*.route.tsx`/`*.test.tsx` naming, and `ReadOnlyLock`/`useWriteAccess` demo-mode discipline all check out against the documented conventions.
- Breadcrumb `handle.crumb`, nav registry gating (`inventory:read`, correctly not the write permission), and the `EXPECTED_LAZY_ROUTE_COUNT` bump are all consistent and enforced by existing contract tests.
- Reasonable test coverage across layers (controller permission lockstep, dialog owner-filtering, page filters/pagination/empty-states/CRUD round-trips).

### Verdict

✅ **Approve** — ready to merge. The two SUGGESTION-level items are optional polish and don't need to block.

PR #3136, opened 2026-09-10
feat(inventory): nav entry for Inventory locations

jakubretajczykBD:
## Summary
- Adds `inventory-locations.route.tsx` (path `inventory/locations`, lazy-loaded, `Platform` breadcrumb group) and registers it in `root.route.tsx`
- Adds the sidebar entry under **Platform**, right after Connections, gated on `inventory:read` rather than `inventory-locations:write` — the page's reads are `admin`/`operator`/`viewer` (`InventoryLocationsController`), so gating on the write permission would hide the entry from `operator` and `viewer`, who can read the page fine. `packer` (which holds no permissions at all) is the only role that never sees it.
- Bumps `route-lazy.test.ts`'s exact lazy-route-count contract (63 → 64)

Stacked on #3066 (#3135). Base branch is `3066-locations-list-page`.

Closes #3069.

## Test plan
- [x] `pnpm --filter @openlinker/web type-check`
- [x] `eslint` on changed files
- [x] `route-lazy.test.ts` + `route-handle.test.ts` + `nav-registry.test.ts` — 130/130 green

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #3069, opened 2026-09-10
feat(inventory): nav entry for Inventory locations


norbert-kulus-blockydevs:
## Problem / Context

No sidebar entry or route exists — the page from a prior sub-issue is unreachable.

## Proposed Solution

Add `apps/web/src/app/routes/inventory-locations.route.tsx` and register it in `apps/web/src/app/nav-registry.ts` under an appropriate group (Platform or Operations).

## Classification

**Type**: Frontend
**Layer**: Interface (routing)
**File(s)**: apps/web/src/app/routes/inventory-locations.route.tsx, apps/web/src/app/nav-registry.ts

## Dependencies

- Part of #3063

## Assumptions

- None

## Acceptance Criteria

- [ ] Page is reachable from the sidebar with no direct URL entry required
- [ ] Tests added or updated for non-trivial logic
- [ ] No architecture boundary violations (CORE ↔ Integration)

norbert-kulus-blockydevs:
Implemented in #3136, merged onto branch 3066-locations-list-page (epic #3063).

PR #3137, opened 2026-09-10
test(inventory): locations CRUD tests

### LINKED ISSUE #3070: test(inventory): locations CRUD tests
## Problem / Context

## Proposed Solution

Add colocated `*.test.tsx` covering list filtering, create/edit validation, the in-use-409-vs-retire branch, and the delete-succeeds path, per `docs/testing-guide.md`.

## Classification

**Type**: Frontend
**Layer**: Interface (tests)
**File(s)**: apps/web/src/features/inventory/**/*.test.{ts,tsx}, apps/web/src/pages/inventory/**/*.test.tsx

## Dependencies

- Part of #3063

## Assumptions

- None

## Acceptance Criteria

- [ ] Create/edit/delete/retire flows are covered
- [ ] pnpm test passes

jakubretajczykBD:
## Summary
Every prior sub-issue (#3064–#3069) already carries colocated unit/component tests for its own layer — the API client, the query/mutation hooks (incl. a dedicated cache-invalidation test), the create/edit dialog (incl. validation and the 409 duplicate-code mapping), the delete/retire dialog (incl. the 409-in-use branch), and the list page's filters/states/write-gating. None of them proves the **page** actually reflects a successful write, since each stops at "the request was made" or "the dialog opened".

- Adds three end-to-end round-trip tests to `inventory-locations-page.test.tsx`:
  - **create** — the new row appears in the table after submit
  - **delete** — the row disappears after confirm
  - **retire** — a 409-refused delete followed by "Retire instead" flips the row's status badge from Active to Retired

Each drives the real mutation hooks' cache invalidation via a second `listLocations` response (`mockResolvedValueOnce` × 2), rather than asserting a mocked table update — so a regression in `inventoryQueryKeys.locationsAll()` invalidation (see #3065's PR) would be caught here even though it isn't touched by this diff.

Stacked on #3069 (#3136). Base branch is `3069-locations-nav-route`.

Closes #3070, the **last** sub-issue of #3063.

## Test plan
- [x] `inventory-locations-page.test.tsx`: 15/15 green (12 existing + 3 new)
- [x] Full `features/inventory` + `pages/inventory` suite: 45/45 green

### DIFF
--- a/apps/web/src/pages/inventory/inventory-locations-page.test.tsx
+++ b/apps/web/src/pages/inventory/inventory-locations-page.test.tsx
@@ -6,6 +6,7 @@ import {
   createMockApiClient,
   renderWithProviders,
 } from '../../test/test-utils';
+import { ApiError } from '../../shared/api/api-error';
 import { InventoryLocationsPage } from './inventory-locations-page';
 import type { InventoryLocation, PaginatedInventoryLocations } from '../../features/inventory';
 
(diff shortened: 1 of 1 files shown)


norbert-kulus-blockydevs:
**Summary**: Adds three end-to-end round-trip tests (create/delete/retire) to `inventory-locations-page.test.tsx`, closing out #3070/#3063 by proving the page actually reflects a write via the real mutation hooks' cache invalidation, not just "the request was made" or "the dialog opened."

**Verdict: ✅ Approve**

This is a clean, well-targeted test-only PR. A few things worth calling out as genuinely good, plus two low-priority nits — nothing blocking.

**What's done well:**
- **Non-brittle round-trip proof**: each test drives a second `listLocations` `mockResolvedValueOnce` and asserts `toHaveBeenCalledTimes(2)`, exercising the real query-invalidation path rather than asserting on a mocked table update. That's exactly the right lens per `docs/testing-guide.md` — it would actually catch a regression in `inventoryQueryKeys.locationsAll()` invalidation even though this diff doesn't touch that code.
- **Reuses established patterns rather than inventing new ones**: the `/^delete$/i` confirm-button selector, the `createAuthenticatedSessionAdapter()` write-access setup, and the 409→"Retire instead" flow all match the sibling `location-delete-dialog.test.tsx` and the file's own pre-existing tests. I checked the underlying `Radix Dialog` (portal + implicit background hiding) and `StatusBadge` (renders raw `location.status`) to confirm the selectors and text assertions (`'active'`/`'inactive'`) are correct against real component output, not assumptions.
- Correct colocation (`inventory-locations-page.test.tsx` next to the page) and naming convention (`*.test.tsx`) per `frontend-architecture.md`.
- RBAC is not re-litigated here — permission-gated visibility of Edit/Delete/Add was already covered by earlier tests in this same file; reusing the authenticated adapter for these three round trips is consistent with that.

**SUGGESTION** — Test title vs. assertion drift: the retire test's description says the badge "flips ... to Retired," but the actual assertion (correctly) checks for the raw lowercase `'inactive'` string, since `StatusBadge` renders `{location.status}` verbatim with no humanized label anywhere. The test is accurate to the real behavior, but the title overstates it — worth a rename so a future reader doesn't assume a "Retired" label exists in the UI (it doesn't, as far as this stack goes).

**SUGGESTION** — None of the three new tests assert the call payload (e.g. `deleteLocation` called with `'ol_location_1'`, `createLocation` called with the typed code/name). The round trip is proven via `listLocations` call count alone. Given payload correctness is already covered at the hook layer (per the PR description), this is low priority, but a one-line `toHaveBeenCalledWith` per mutation would tighten the "real" round-trip claim further.

Nothing else stood out — the assertions match real component/hook wiring, edge-case coverage (409→retire) is exercised, and scope is appropriately narrow for a test-only PR.

Issue #3141, opened 2026-09-10
design(listings): price-changes review queue mockup


norbert-kulus-blockydevs:
## Problem / Context

Part of #3139 (epic). **Stack position: second — branches from the ADR sub-issue's branch (not `main`).**

Before any implementation sub-issue starts, the reviewed, interactive HTML mockup needs to be a durable, repo-committed artifact per `docs/frontend-architecture.md` § UX Mockups — "A high-fidelity, interactive HTML mockup ... must be committed to the repo ... not left as an ephemeral Claude Artifact link." It was designed and iterated live as a Claude Artifact across many review rounds (6 specialist sub-agent reviews, 3 rounds of user-driven copy/UX/architecture corrections) and is now converged.

Live Artifact (design-history reference only, not the source of truth once this issue merges): https://claude.ai/code/artifact/3895933b-4f60-4935-b43b-84eb99e437b1

## Proposed Solution

Commit the artifact's current HTML verbatim to `docs/plans/mockups/price-changes-review-queue.html`, following the exact precedent of `docs/plans/mockups/correction-proposal-3087.html` (issue #3116/#3123 in this repo) and `docs/plans/mockups/analytics-display-currency-picker.html`.

The file is a **real, interactive, self-contained** HTML prototype (no build step, no external API calls — inline `<style>`/`<script>`) with:
- A `data-preview` state switcher (design-review-only affordance, clearly labeled, NOT to be shipped in the real feature) covering `live` / `readonly` / `loading` / `empty` / `error`.
- Explicit `data-state` attributes on every major container and dialog (`page-listings`, `page-connection`, `page-source`, `queue-live`, `queue-loading`, `queue-empty`, `queue-error`, `dialog-single-accept`, `dialog-edit-price`, `dialog-bulk-accept`, `dialog-stale-info`, `dialog-connection-picker`, `dialog-auto-applied`, `bell-panel`, `bulk-action-bar`, `bulk-publish-progress`) plus per-row `data-state` (`row-pending` / `row-accepted` / `row-accepted-custom` / `row-ignored` / `row-needs-refresh` / `row-publishing`) — every sub-issue below (#7–#11) MUST implement the state its title says using these exact names, and #12 (E2E) references them directly.
- Full `id` + `data-testid` coverage on every interactive element (buttons, checkboxes, form fields, filter chips) — reused as-is by real components' `id`/`data-testid` where practical, and by #12's Playwright selectors regardless.

[...]

Issue #3139, opened 2026-09-10
[EPIC] Opt-in recurring price propagation with review queue (#3010)


norbert-kulus-blockydevs:
## Problem / Context

`libs/core/src/identifier-mapping/domain/types/pricing-rule.types.ts` (`readPricingRule` / `applyPricingRule`) computes a destination price from the master catalog price only at **offer-create / product-publish time** (`OfferBuilderService.buildCreateOfferCommand`, `ProductPublishBuilderService.buildPublishProductCommand`) — confirmed by code-audit sub-agent, zero other call sites. There is **no ongoing re-price** when the master price changes later: a price update at the source (e.g. PrestaShop) never reaches an already-published Allegro/Erli/WooCommerce listing. This is the gap #3010 named and deferred as "a product decision, not a settled feature gap."

Decision made across a design session (mockup iterated live, reviewed by 6 specialist sub-agents — 3 project-fit, 2 UX, 1 e-commerce — plus a real-codebase backend-gap audit): build it as **opt-in per (source connection, destination connection) pair**, default off, surfaced as a review queue on the Listings page, with a live interactive HTML mockup reviewed and iterated to convergence.

**Reviewed mockup** (committed to the repo by #3141, this epic's first implementation sub-issue — read it before touching any other sub-issue):
`docs/plans/mockups/price-changes-review-queue.html`

Live Artifact used during design (for design-history reference only — the **repo-committed file is the durable spec**, per `docs/frontend-architecture.md` § UX Mockups): https://claude.ai/code/artifact/3895933b-4f60-4935-b43b-84eb99e437b1

## Proposed Solution

Ship the feature as: (a) a `price_change_episodes` aggregate inside the existing `listings` context (episode pattern, mirroring `ReservationShortfallEpisode`'s partial-unique-index-on-open-row shape); (b) a default + per-(source,destination)-pair override pricing-rule model, replacing the flat single-rule `Connection.config.pricingRule`; (c) a review queue UI (Listings → "Price changes" tab) with per-row Accept/Ignore/Edit, bulk apply with live per-connection progress, and a "Pricing & sync" settings surface reachable from **two** entry points (a picker modal from the queue, and a read-only rollup card on the feeding source connection's own page) landing on the owning **destination** connection's page.

[...]

PR #3154, opened 2026-09-10
docs(adr): recurring price propagation & review queue

### LINKED ISSUE #3140: docs(adr): recurring price propagation & review queue
## Problem / Context

Part of #3139 (epic). **Stack position: first — branches from `main`.**

The recurring price-propagation feature (originating gap: #3010) involves several non-trivial architectural decisions that affect multiple bounded contexts (`listings`, `sync`, `worker`) and reverse/extend an existing pattern (`Connection.config.pricingRule` moves from a flat single rule to a default + per-source-override model). Per `docs/engineering-standards.md` § Architecture Decision Records, this qualifies: it affects multiple contexts, has genuine rejected alternatives (three sync modes vs. two; new bounded context vs. extending `listings`; episode pattern vs. simple per-event rows), and a future maintainer will plausibly ask "why didn't we just add a `digest` mode" or "why is this on the destination, not the source."

## Proposed Solution

Write `docs/architecture/adrs/0NN-recurring-price-propagation.md` (copy `template.md`; next free number — check `docs/architecture/adrs/README.md`'s index at write time, do not hardcode `072`, it may already be claimed by unrelated concurrent work per `docs/architecture-overview.md`'s reference to an append-only-retention ADR). Content — one section per decision, each citing the rejected alternative and why:

1. **Episode pattern for the pending record** — `price_change_episodes`, modeled on `ReservationShortfallEpisode` (`libs/core/src/inventory/domain/entities/reservation-shortfall-episode.entity.ts`): a row opened once, a partial unique index enforcing at-most-one-open-row per `(productVariantId, destinationConnectionId, sourceConnectionId)`, re-detection updates the same open row. Rejected alternative: a simple insert-per-detection-event row, which duplicates on re-detection before review.
2. **Pricing rule ownership: destination, not source** — default rule + per-source overrides live on the DESTINATION connection (`Connection.config.pricingRule` extended, not replaced structurally — same JSONB column, new shape: `{ default: PricingRule, overrides: Record<sourceConnectionId, PricingRule> }`). Cite the real precedent (`applyPricingRule`/`readPricingRule` already destination-scoped) and the cardinality argument (one destination commonly has many sources; the reverse is rare). The feeding source connection gets a READ-ONLY rollup, never its own editable rule.
3. **Two sync modes, not three** — `manual` / `automatic`. State plainly that a `digest` mode was considered and rejected for v1 because no notification/summary surface exists anywhere in the codebase to build on (verified by full-repo grep during design) — record this as the reason future readers should NOT casually add a third mode without first scoping the summary UI it needs.
4. **Currency: same-currency-only in v1** — a source/destination currency mismatch is a persisted `blockReason` (following the `SalesDocumentBlockOutcome` pattern in `libs/core/src/sales-documents`: pure resolver, owning-service persistence, re-decided every transition, level-triggered). No FX conversion attempted. Cite `ADR-040` (order-time FX stamping) as the reason this is deliberately out of scope, not an oversight.
5. **Pinned/custom price is a NEW mechanism** — not a reuse of the existing destination-reported `frozen` field (#2262, `offer-fields-update.types.ts`), which is structurally the wrong direction (Erli-reports-to-OL, not OL-reports-to-itself). New nullable columns on `price_change_episodes`.
6. **No cost/margin data** — record explicitly that `Product`/`ProductVariant` carry no COGS field (verified by grep during design) and that the "steep" magnitude signal in the UI is NOT a margin/profitability check and must never be presented as one. Name real cost tracking as a future prerequisite for a genuine margin-below-cost warning, out of scope here.
7. **Bulk apply reuses `BulkOfferCreationBatch` / `bulk_batch_advancements`** rather than inventing a second progress-counter mechanism.
8. **Publish failure classification follows ADR-007** (status vs. outcome split) — a deterministic block is a terminal `business_failure` on the same attempt, never a throw that burns the retry ladder against a condition retrying cannot change.

Status: `Proposed` (flips to `Accepted` once #3139's backend sub-issues land and the shape proves out — do not pre-emptively mark `Accepted`).

## Classification

**Type**: DX / Architecture documentation
**Layer**: N/A (documentation)
**File(s)**: `docs/architecture/adrs/0NN-recurring-price-propagation.md`, `docs/architecture/adrs/README.md` (index entry)

## Dependencies

- None (first sub-issue in the stack, branches from `main`)

## Assumptions

- The ADR number is resolved at write time from the actual highest existing file in `docs/architecture/adrs/`, not assumed from this issue text.
- This issue is implemented directly in this session, not handed to a future implementer — see epic body.

## Acceptance Criteria

- [ ] ADR file exists, follows `template.md` structure exactly (Context / Decision / Alternatives considered / Consequences / References)
- [ ] All 8 decisions above are covered, each with its rejected alternative stated
- [ ] `docs/architecture/adrs/README.md` index updated with the new entry
- [ ] `docs/architecture-overview.md` is NOT edited by this issue (that happens once implementation lands and the pattern is real — an ADR for unimplemented work should not retroactively claim space in the living architecture doc)
- [ ] PR references `docs/plans/mockups/price-changes-review-queue.html` (landing in the next stacked PR, #1) as the UX companion to this architectural record

norbert-kulus-blockydevs:
## Summary
- Adds ADR-072 recording the architectural decisions made during the price-changes review-queue design session (episode data model, destination-owned pricing rules with per-source overrides, two sync modes, same-currency-only v1, new pinned-price mechanism, reuse of bulk-batch progress).
- Updates the ADR index (`docs/architecture/adrs/README.md`).

**Stack position**: first in this epic's stacked-PR chain (#3139). This branches from `main`; the next PR (#3141, mockup commit) branches from this one.

Closes #3140

## Test plan
- [ ] Docs-only change — no runtime code touched. Verify the ADR renders correctly and matches `docs/architecture/adrs/template.md`'s structure.
- Note: pre-commit hook was skipped (`--no-verify`) for this commit with explicit user authorization — `apps/e2e` has a pre-existing, unrelated lint failure on `main` (missing `node_modules`, not caused by this change). CI should be checked to confirm this doesn't block the PR; if CI also fails on the same pre-existing issue, that's a separate problem to raise, not something this PR introduced.


piotrswierzy (commented):

An unusually honest ADR — seven rejected alternatives each with a real reason, two acknowledged compromises (the "steep change" signal will both miss and over-flag; the currency restriction leaves cross-currency connections with zero automation), and three genuine precedents reused rather than three new mechanisms invented. It clears the "when to write one" bar comfortably and follows the template section-for-section.

**Guard result, since you asked:** `scripts/check-adr-numbers.mjs` does not exist on the PR head (the branch predates #3036), so I drove main's exported validators against the post-merge file set: **rules 1–3 OK, rule 4 (unique vs `origin/main`) OK, rule 5 (README bijection) OK** — 72 rows, the ADR-072 row present and correctly formed. `072` is free (main's highest is `071`), the heading matches the filename prefix. `check-architecture-gates.mjs` is also clean. **The ADR itself is guard-clean on all five rules.**

### BLOCKING — the README edit collides with #3036, which rewrote that exact paragraph

`docs/architecture/adrs/README.md` (166–167). `main` now reads:

> *Reserved numbers: this was previously a hand-maintained log of in-flight number claims … it drifted from reality at least three times in two days … `check-adr-numbers.mjs` (#2082) now enforces uniqueness mechanically at `pnpm lint` time … so this note is no longer needed*

This PR extends the **old** long note, adding *"072 is taken by #3139/#3140 … Allocate 073 for the next new ADR"*, and preserves *"Nothing currently guards ADR numbering mechanically — tracked as #2082"* — which is now **false**, since #3036 shipped exactly that guard.

Git will conflict on the line, so it isn't silent; the risk is a textual resolution taking this PR's side and **reinstating the hand-maintained reservation log that main deleted because it drifted three times in two days.** Correct resolution: keep the new ADR-072 index row (rule 5 requires it), take main's paragraph, drop the reserved-numbers edit. Rebasing is the clean path.

### IMPORTANT — "confirmed the only two call sites by grep" understates the surface by five readers and two invariant scripts

Line 9. True of `applyPricingRule` on the *publish* path; **false** as a statement about who reads `Connection.config.pricingRule`, which is what decision 2 and the Cons bullet propose reshaping (flat → `{default, sourceOverrides}`). Additional readers I verified:

- `apps/web/src/features/connections/lib/stock-and-pricing-preview.ts` — **a second `applyPricingRule` implementation**, the #2610 browser mirror
- `stock-and-pricing-section.tsx`, `EditConnectionForm.tsx` (`readPricingRuleForm`), `edit-connection.schema.ts`
- `apps/api/src/integrations/application/services/connection.service.ts` — **server-side** validation

And two scripts lock the seam: `check-stock-and-pricing-preview-mirror.mjs` guards `applyPricingRule` by token comparison, and `check-architecture-gates.mjs` carries `{helper: 'readPricingRule', key: 'config.pricingRule'}` in `KNOWN_CONFIG_KNOBS`.

`README.md § How to write a new ADR` rule 4: *"An ADR that confidently states incorrect specifics is worse than no ADR — it corrupts the audit trail."* The Cons bullet does say *"every existing reader … with a backward-compatible fallback"*, so the direction is right — but for the governing ADR of a 12-PR epic, **the browser mirror is precisely where #2610's byte-identity discipline either survives the shape change or silently breaks**, and those two scripts are where the epic will trip `check:invariants`. Name them.

*(For what it's worth: #3166 has since shipped a second pricing editor beside `StockAndPricingSection` with no precedence — exactly the reader the Cons bullet meant and the PR's grep claim obscured.)*

### IMPORTANT — the guard could not run on this PR, and that's worth a follow-up against #3036

CI's `Detect changed paths` filter excludes `docs/**` from `code`, and the **Lint** job — the only thing running `check:invariants` → `check-adr-numbers.mjs` — is gated on it. Confirmed empirically here: `Lint`, `Test`, `Build`, `Integration Tests` all **SKIPPED**. Combined with the documented `--no-verify`, **#3036's guard ran nowhere for this PR** — not locally, not in CI. It runs post-merge, so a collision would redden `main` *after* the fact, which is the "silent duplicate discovered by the second merger" failure #3036 exists to prevent. Add `docs/architecture/adrs/**` to the `code` filter, or run the guard ungated.

### SUGGESTIONS

- The eight sub-decisions never state **where the episode sits relative to the shipped pricing path** — a reader can't tell whether an approved episode re-enters the existing publish path ending in `applyPricingRule`, or introduces a second pricing path. One sentence settles it. On substance it *does* square with the shipped model: `passthrough|markup|margin` preserved, flat rule read as the new `default`, no backfill migration, every connection defaults to Manual review.
- Several decisions are gate-shaped ("two sync modes, not three", "same-currency only in v1", "no margin warning") but none is written as a reversal gate. `check-architecture-gates.mjs` enforces *marking*, not *having*, so not a blocker — but ADR-071 next door names its re-take conditions, and for a 12-PR epic that's worth three lines.
- 960 words against the 500 target, permitted by the stated exception (7 alternatives). `docs/plans/mockups/price-changes-review-queue.html` doesn't exist yet (#3141) — fine for a stacked PR, a dangling link if #3141 lands later.

**Verified accurate** per rule 4: all six cited paths and three ADR cross-links resolve; the "no cost/COGS field anywhere in `Product`/`ProductVariant`" claim holds; `ReservationShortfallEpisode` exists and does use the partial-unique-on-open-row pattern decision 1 describes; all cited issue numbers match their described roles.

---

norbert-kulus-blockydevs:
Addressed the review. Summary:

### BLOCKING — resolved
Merged `origin/main` into this branch (merge commit, not a rebase — 11 downstream branches in the stack are already based on this one). This picked up #3036's rewritten "reserved numbers" paragraph verbatim; this PR's stale long-form note (which added "072 is taken" while leaving the now-false "nothing currently guards ADR numbering mechanically" sentence) is gone entirely — the new paragraph already describes the mechanical guard, so no replacement text was needed. The ADR-072 index row is preserved.

### IMPORTANT — fixed
The `§ Context` claim "confirmed the only two call sites by grep" was true only of `applyPricingRule` on the *publish* path, false as a statement about `Connection.config.pricingRule` itself (the value decision 2 reshapes). Widened the claim to name every real reader: the `#2610` browser mirror (`stock-and-pricing-preview.ts`), the connection form/schema, server-side validation in `connection.service.ts`, and the two invariant scripts that lock the seam (`check-stock-and-pricing-preview-mirror.mjs`, `check-architecture-gates.mjs`'s `KNOWN_CONFIG_KNOBS`). Also named these explicitly in the Cons bullet, and noted #3166's newly-shipped second pricing editor as exactly the second reader that bullet warns about.

### IMPORTANT — acknowledged, no code change needed
`node scripts/check-adr-numbers.mjs` now runs cleanly against this branch post-merge: `OK (72 ADRs; uniqueness vs origin/main: checked)`. `check-architecture-gates.mjs` also passes clean. Nothing further to do here — the guard existing on `main` and this branch now including it was the fix.

### SUGGESTIONS
- Addressed: added one paragraph stating an approved episode re-enters the *existing* publish path ending in `applyPricingRule` rather than introducing a second one.
- Addressed: added a "Revisit when" section (matching ADR-071's precedent) naming reversal-gate conditions for decisions 3 (two sync modes), 4 (same-currency-only), and 6 (no margin warning).
- Declined: word count (960 vs 500 target) and the dangling mockup link — the review itself already judged both acceptable given the stated 7-alternatives exception and the stacked-PR ordering (#3141 hasn't landed yet).

**Verification**: `node scripts/check-adr-numbers.mjs` → OK. `node scripts/check-architecture-gates.mjs` → OK. Diff is docs-only (one file changed on top of the merge commit); no code, tests, or migrations touched.

piotrswierzy (approved):
## Tech Lead review

Re-reviewed at `bcc654c`. All three findings from my earlier pass are addressed, and the BLOCKING one is resolved exactly right: the README diff against `main` is now **one line** — the ADR-072 index row — with #3036's rewritten reserved-numbers paragraph taken verbatim and this PR's stale long-form note (and its now-false *"nothing currently guards ADR numbering mechanically"* sentence) gone entirely. Guard rules re-checked by hand against the post-merge file set: `072` is free (`main`'s tail is `071`), the `# ADR-072:` heading matches the filename prefix, and the README↔file bijection holds. The §Context claim is now accurate about the reader surface it describes — it names the `#2610` browser mirror, the connection form/schema, server-side `connection.service.ts` validation, and both invariant scripts, and the Cons bullet calls out that all three must move in one change or `check:invariants` trips. The "Revisit when" section now names re-take conditions for decisions 3, 4 and 6, and the new opening paragraph settles the question the eight sub-decisions left open — an approved episode re-enters the **existing** publish path ending in `applyPricingRule`, it does not introduce a second one.

Checked against shipped invariants and found no contradiction: `#1843 pricingRule`'s `passthrough|markup|margin` shapes are preserved verbatim with the flat rule read as the new `default` and no backfill; `#1844 stockSafetyBuffer` and ADR-061's `applyPublishControls` single-reader seam are untouched (pricing was never folded into Controls, so "both builders call `applyPricingRule`" is still true); `#2610`'s mirror discipline is named as the thing at risk rather than assumed safe; and decision 5's reading of `frozen` as destination-*reported* matches how #2245 and #988/#1737 describe it.

### Findings

**[SUGGESTION]** — CI path filter (restated from my previous pass; not this PR's to fix)

> `Detect changed paths` still excludes `docs/**` from `code`, so **Lint is SKIPPED at this head too** — #3036's `check-adr-numbers.mjs` has run nowhere for this PR, in CI or locally (`--no-verify`). I verified the guard's rules by hand instead, so this PR is safe; but a guard that cannot run on an ADR-only PR is a guard that catches an ADR collision only post-merge, on `main`, which is the failure mode #3036 exists to prevent. Worth a one-line follow-up issue adding `docs/architecture/adrs/**` to the `code` filter (or running that one script ungated).

**[SUGGESTION]** — `docs/architecture/adrs/072-recurring-price-propagation.md:71`

> The mockup reference points at `docs/plans/mockups/price-changes-review-queue.html`, which does not exist yet (#3141, next in the stack). Correct for a stacked PR; just make sure #3141 lands before or with this one so `main` never carries a dangling reference — `docs/frontend-architecture.md § UX Mockups` makes the repo-committed mockup the durable spec that implementation tickets cite.

### Verdict
✅ **Approve**

---

Issue #3156, opened 2026-09-10
[BUG] Interface — POST /auth/register leaks username/email existence via echoed identifier


norbert-kulus-blockydevs:
## Problem / Context

`POST /auth/register` (`apps/api/src/auth/auth.controller.ts`, `register()`) allows user enumeration. `RegistrationService.register()` (`apps/api/src/auth/registration.service.ts`, lines ~78-88) runs two independent uniqueness checks and throws a distinct-per-case exception whose message embeds the raw identifier that collided:

```ts
const [existingByUsername, existingByEmail] = await Promise.all([
  this.userRepository.findByUsername(username),
  this.userRepository.findByEmail(email),
]);

if (existingByUsername) {
  throw new UserAlreadyExistsException(username);
}
if (existingByEmail) {
  throw new UserAlreadyExistsException(email);
}
```

`UserAlreadyExistsException` (`libs/core/src/users/domain/exceptions/user-already-exists.exception.ts`) builds its message as `` `User already exists: ${identifier}` ``. The controller then forwards that message verbatim to the client:

```ts
if (error instanceof UserAlreadyExistsException) {
  throw new ConflictException(error.message);
}
```

Result: submitting a registration with a guessed email (and any unused username) returns `409 {"message": "User already exists: alice@example.com", ...}` — a direct oracle confirming that email is registered, no login/password needed. The same applies for a guessed username. (Corrected framing vs. the original report: the echoed identifier always matches the field that actually collided — it does not cross-leak username→email or vice versa. The real defect is that *either* branch echoes the caller-supplied identifier back at all, which is itself sufficient for enumeration.)

This directly contradicts the enumeration-hardening pattern already established elsewhere in the same controller:
- `resendConfirmation` — "Always returns 200 to prevent user enumeration."
- `forgotPassword` — same pattern.
- `confirmEmail` — deliberately does not forward `error.message` to the client; logs the real reason server-side and returns one generic message instead.

`register` has no equivalent protection.

Affects both normal-mode and demo-mode registration — demo mode is a config flag on the same endpoint (`RegistrationService.register`), not a separate route.

## Proposed Solution

Stop returning distinguishable, identifier-echoing responses for username-taken vs. email-taken.

[...]

Issue #2348, opened 2026-08-22
[TASK] CORE — repoint the ADR-028 cancellation restore onto ATP, ordered after release


piotrswierzy:
> Part of the OMS programme backlog (**Wave 2**, stream S1, size M). Backlog slug: `W2-11`. Sibling `W2-*` slugs resolve via the Wave-2 epic's checklist.

## Problem / Context

ADR-028's cancellation stock restore predates the ledger. Run in the wrong order it double-counts: the restore republishes the quantity while the reservation is still `held`, so ATP momentarily reads too high, and a fast poller can publish that number to a marketplace. Story I5 ("a cancelled or expired order returns its claim and my offers go back up") requires exactly one increase, in one order.

## Proposed Solution

On cancellation: **release the reservations first** (`releaseHeld`, which lowers `olReservedQuantity`), **then** run the ADR-028 restore, and let the ordinary propagation fan-out carry one recomputed ATP outward. The restore is repointed to publish from `IAvailabilityService` rather than computing its own number, so there is one formula.

Ordering is enforced in code (one service method, documented), not by convention across two handlers.

## Classification

**Type**: CORE
**Layer**: Application
**File(s)**: the ADR-028 cancellation restore service under `libs/core/src/inventory/` / `libs/core/src/orders/`; `libs/core/src/inventory/application/services/availability.service.ts`

## Dependencies

- #2344 (`W2-7`), #2345 (`W2-8`)

## Assumptions

- Cancellation after dispatch does not restore (the shipment already consumed) — the contradiction is displayed, per story L6, not reconciled.

## Acceptance Criteria

- [ ] Cancelling a reserved order produces exactly one ATP increase, of the right size (integration test)
- [ ] Release strictly precedes restore; an ordering-inverting refactor fails a test
- [ ] A cancelled-after-dispatch order does not double-restore
- [ ] Tests added; no boundary violations

Issue #2229, opened 2026-08-20
[TECH-DEBT] Integration — make the effective category-resolve in-flight ceiling operator-visible


norbert-kulus-blockydevs:
## Problem / Context

The streamed EAN category-resolve path (#2215, epic #2205) caps its outbound Allegro calls at a hard-coded in-flight ceiling that the operator can neither see nor change:

```ts
// libs/integrations/allegro/src/infrastructure/util/resolve-categories-for-batch-by-ean.ts:98
export const STREAM_CONCURRENCY = 9;
```

`streamCategoriesForBatchByEan` uses it as `options?.concurrency ?? STREAM_CONCURRENCY`, so on the wizard's Resolve step it is the effective ceiling on `GET /sale/products` for that connection.

Two things make that invisible rather than merely undocumented:

1. **`allegroAdapterManifest` declares no `defaultRateLimit`, and deliberately never will** (#1810 §1: a manifest default is meant for merchant-hosted platforms; a fabricated RPM for a multi-tenant marketplace would be surfaced to the operator as "adapter default"). So the usual "the operator's configured cap wins" only holds once `connection.config.rateLimit` is set. With none set, **9 is the cap**, chosen in code.
2. **The existing rate-limit surface actively says otherwise.** `GET /connections/:id/rate-limit-status` returns `enabled: false` when neither `config.rateLimit` nor a manifest default applies, and `RateLimitSection` renders that as no cap being in effect. That is true of the shared outbound limiter and false of what the resolve path actually does — the ceiling lives inside the adapter's own resolver, below the limiter. An operator reading the connection page is told nothing is throttling them while 9 is.

A second fact belongs on the record here, because the constant's justification is easy to over-read. `STREAM_CONCURRENCY = 9` was chosen as "what a 3-chunk batch already sustained in production": before the streamed step, the wizard split a batch into 50-variant requests fired in parallel, each request building its own adapter capped at `DEFAULT_CONCURRENCY = 3`, so the effective count was `3 * ceil(variants / 50)`. That premise only holds from roughly 150 variants up. The client now splits at the route's 200-item cap, so **a 45-variant batch went from 3 in flight to 9** — a straight 3x on outbound calls for small batches, accepted deliberately in #2215 but never surfaced.

Reactive protection is unaffected either way: a 429 still parks `AllegroHttpClient` on `Retry-After`. This issue is about the proactive ceiling being an invisible code constant.

[...]

PR #3157, opened 2026-09-10
fix(auth): stop leaking username/email existence via registration errors

norbert-kulus-blockydevs:
## Summary

- `POST /auth/register` returned a 409 whose message echoed the exact colliding identifier (`User already exists: alice@example.com`), letting an unauthenticated caller confirm whether a specific username or email is already registered.
- `UserAlreadyExistsException` now carries a fixed, non-identifying message (`"Username or email is already in use"`), identical for both the username-collision and email-collision branches. The colliding identifier stays on the exception's `identifier` field as structured context for a debugger or an error sink; nothing logs it and nothing reads it into a response.
- `RegistrationService` logs **which field** collided, at `debug`, never the submitted value — `register` is `@Public()` and unthrottled outside demo mode, so the value is attacker-supplied.
- The registration form renders the server's own field-neutral message instead of hardcoding `"This email is already registered."`, which asserted a specific email exists and was outright false for a username-only collision.

## Scope — this narrows the leak, it does not close it

`register` still answers **409** for a taken identifier and **201** for a free one. The caller controls both fields, so `{username: <random-unique>, email: victim@example.com}` reads "registered" off the status alone; the message never had to name the field. Closing that means always answering 201 and notifying by email (the `forgotPassword` shape) **plus** hashing unconditionally, since the collision path returns before `bcrypt.hash` and is measurably faster.

So this PR is `Refs #3156`, **not** `Closes` — merging it must not retire a live issue. The surviving channel and its remedy are recorded where a reader meets them: the `UserAlreadyExistsException` docblock, the 409 Swagger description, and a characterisation test that a future always-201 fix flips.

Also relevant to prioritising that follow-up: the registration rate limiter is gated on `demoMode`, so a self-hosted install with `OL_REGISTRATION_ENABLED=true` has **no throttle on `register` at all** and the 409 oracle can be walked at full speed. The `429` Swagger response is now documented as demo-mode-only.

## Withdrawn claim

An earlier revision of this description claimed parity with `resendConfirmation` / `forgotPassword`. Those return **200 unconditionally**; `register` still returns 409 vs 201, so the claim is withdrawn. The only pattern genuinely shared is "do not echo the submitted identifier".

## Test plan

- [x] `pnpm --filter @openlinker/core lint` / `type-check` — 0 errors
- [x] `pnpm --filter @openlinker/api lint` / `type-check` — 0 errors
- [x] `pnpm --filter @openlinker/web type-check` + eslint on `features/users` — 0 errors
- [x] Service-level test: a username collision and an email collision throw byte-identical, non-identifying messages
- [x] Controller-level test: both collisions produce an identical **status and body** (the service-level message assertion is not the same claim — what a caller observes is the response)
- [x] Characterisation test pinning the surviving 409-vs-201 channel, so it cannot be closed silently or reopened unnoticed
- [x] FE test asserts the server message is rendered and the old email-specific sentence is gone
- [ ] Unit suites not re-run locally this round (test execution is offloaded to CI on this machine) — CI is the gate

**Known stale text, left deliberately:** `docs/plans/implementation-plan-1125-user-management.md:139` still quotes the original `User already exists: ${username}` message. It is a historical implementation plan describing what was built at the time, not live guidance, so it is not rewritten — but an earlier "grepped repo, none found" line in this description was inaccurate and has been removed.

Refs #3156


piotrswierzy (commented):

A faithful, complete implementation of exactly what #3156 asked for — **and #3156 under-scoped the vulnerability.** The identifier echo is gone from all four throw/catch sites (including `user.repository.ts:113-120`, which throws the same exception from the `23505` race path and inherits the generic message for free) with no leak path for the newly-public `identifier` field: the controller constructs a *fresh* `ConflictException(error.message)`, and I checked all seven global filters — none spreads or stringifies an error object. Tests 39/39 pass locally; all 9 CI checks green.

### The channel the fix does not close

| Channel | Closed? |
|---|---|
| Error **message** | ✅ Closed, completely |
| HTTP **status** (409 vs 201) | ❌ **Open — this is the actual oracle** |
| Response **shape** | ❌ Open, follows the status |
| **Timing** | ❌ Open |

To test whether `victim@example.com` is registered: POST `{username: <random-unique>, email: victim@example.com}`. **409 means taken, 201 means free.** The attacker controls both fields, so the message never had to name the field.

#3156's framing — *"the real defect is that either branch echoes the caller-supplied identifier back at all, which is itself sufficient for enumeration"* — mistakes a sufficient condition for a necessary one. The 409 alone is sufficient; the echo was never required.

### IMPORTANT — `Closes #3156` will retire a live vulnerability

Change it to `Refs #3156` and leave the issue open naming the surviving status-code channel, or widen the fix to always-201 + notify-by-email (the `forgotPassword` shape). This matters more than usual because **rate limiting is demo-mode only** — `if (demoMode && clientIp)` — so on a production install with `OL_REGISTRATION_ENABLED=true` there is no throttle on `register` at all, and the 409 oracle can be walked at full speed against the whole user base. (Related: the controller documents `@ApiResponse({ status: 429 })` for a response unreachable outside demo mode.)

### IMPORTANT — the two new `logger.warn` lines write attacker-supplied PII at warn level

On a `@Public()`, unthrottled endpoint. Raw email and username verbatim — both a PII-into-logs problem and a log-flooding vector an unauthenticated caller controls. It also diverges from the repo's own precedent one file over: `password-reset.service.ts:52` deliberately logs **without** the address and at `debug`:

```ts
this.logger.debug('Password reset requested for unknown email');
```

Match it — `debug` level, and log *which field* collided without the value.

### IMPORTANT — the FE still makes the enumeration claim, and now makes a false one

`register-form.tsx:98-100`:
```tsx
register.error.isConflict() ? 'This email is already registered.' : register.error.message
```

Three consequences: the product surface still tells an anonymous visitor the email exists; the message this PR genericised is **discarded** for exactly the 409 case, inverting the PR's premise; and because a **username**-only collision also returns 409, the user is told their *email* is taken, changes it, resubmits, and gets the same sentence — an unbreakable loop. Previously the server at least named the real field.

That is also the direct answer to "does this degrade a genuine UX need": **yes**, and the PR body does not acknowledge the trade-off.

### IMPORTANT — the test asserts identical *messages*, not identical *responses*

#3156 step 5 asked for an identical response body/message. A controller-level test asserting that a taken and a free registration produce the same **status + body** would have surfaced the 201-vs-409 gap. As written the suite can never fail on the real leak.

### SUGGESTION — timing oracle, for the follow-up

The collision path returns **before** `bcrypt.hash(password, 10)` (≈60–100 ms) and before `save` + the confirmation email. Existing ⇒ fast, free ⇒ slow, trivially measurable. Moot while the 409 stands, but a later always-201 fix would be defeated by it unless the hash runs unconditionally.

Also: the body claims parity with `resendConfirmation`/`forgotPassword`, which return **200 unconditionally** — `register` still returns 409 vs 201, so that claim should be withdrawn. And `docs/plans/implementation-plan-1125-user-management.md:139` still documents the old raw message, so the "grepped repo, none found" claim is inaccurate (leaving the historical plan text is fine).

`@Public()` on `register` is unchanged and correct, and `route-authorization-coverage.spec.ts` is discovery-based so it stays covered with no edit.

---

norbert-kulus-blockydevs:
## Review addressed — all four IMPORTANTs and every suggestion

@piotrswierzy thank you, and the framing correction is accepted in full: **#3156 under-scoped the vulnerability, and the 409 was always the oracle**. Everything below follows from taking that as the premise rather than arguing with it.

Pushed as `926ec0f` (plus a message-only amend of the original commit — no inline comments existed, so nothing was orphaned).

---

### IMPORTANT — `Closes #3156` would retire a live vulnerability → **fixed, and written down in three places**

`Closes` → `Refs`, in both the commit message and the PR description. The issue stays open and now carries [a comment](https://github.com/openlinker-project/openlinker/issues/3156#issuecomment-5638501911) naming the surviving status-code channel, the timing channel, the unthrottled-outside-demo-mode aggravator, and the three steps to close it — including the one that is a product decision rather than a mechanical fix (an email collision can be reported out-of-band; a **username** collision cannot, and that is why this was not widened here).

A PR description is read once. So the limit is also stated where someone will actually meet it:

- `UserAlreadyExistsException`'s docblock — a "Scope limit, stated so it is not mistaken for a closed hole" paragraph.
- The **409 Swagger description**, which now says the status is itself an oracle regardless of the message.
- The **429** description, which now says the limiter is demo-mode-gated so a self-hosted install never returns it (your parenthetical — the response was documented as reachable when it is not).

### IMPORTANT — the two new `logger.warn` lines write attacker-supplied PII at warn level → **fixed**

Matched to the precedent you named rather than invented:

```ts
this.logger.debug('Registration rejected: username already taken');
```

`debug`, and **which field** collided without the value — the `password-reset.service.ts:52` shape. The comment at the call site records why (`@Public()`, unthrottled outside demo mode, so both a PII-into-logs problem and a log-flooding vector an anonymous caller controls).

`identifier` stays on the exception, but its docblock no longer claims the service logs it — that was true when you reviewed and is not any more. It is now described as structured context for a debugger or error sink, with the "never read it into a response" rule kept.

### IMPORTANT — the FE still makes the enumeration claim, and now a false one → **fixed**

`register-form.tsx` renders `register.error.message` — the server's own field-neutral sentence. The `isConflict()` branch and the `ApiError` import are gone.

Your third consequence was the one that decided the shape of the fix: a **username**-only collision returns the same 409, so the old copy sent the user to edit an email that was never the problem, forever. And yes — the answer to *"does this degrade a genuine UX need"* is **yes**, and the PR description now says so instead of quietly not mentioning it.

The FE test was asserting the bad copy, so it was inverted rather than deleted: it now asserts the server message renders **and** that the old email-specific sentence is absent.

### IMPORTANT — the test asserts identical *messages*, not identical *responses* → **fixed**

New `describe('POST /auth/register')` in `auth.controller.spec.ts`, at the layer where the response is shaped:

- both collision kinds produce an identical **status and body**, and the serialised body matches no submitted value;
- a second, explicitly-labelled **characterisation** test pins the surviving 409-vs-201 difference.

That second test is the point of the pair. You are right that the suite as written could never fail on the real leak; it still cannot *close* it, but the gap can no longer be closed silently or reopened unnoticed, because a future always-201 fix has to flip a test that says in its own comment what it is for.

### SUGGESTION — timing oracle → **recorded, not fixed**

Carried into the #3156 comment as a hard precondition on the follow-up: hashing must become unconditional, or step 1 (always-201) is defeated on its own. Moot while the 409 stands, exactly as you say, so no change here.

### SUGGESTION — the `resendConfirmation`/`forgotPassword` parity claim → **withdrawn**

The description now has a "Withdrawn claim" section saying those return 200 unconditionally while `register` still returns 409 vs 201, so the only pattern genuinely shared is "do not echo the submitted identifier".

### SUGGESTION — `implementation-plan-1125-user-management.md:139` → **claim corrected, text left**

You said leaving the historical plan text is fine and I agree — it describes what was built at the time, it is not live guidance. What was wrong was the description's "grepped repo, none found" line, which is removed and replaced with an explicit "known stale text, left deliberately" note naming the file and line.

---

### Verification

| | |
|---|---|
| `@openlinker/api` type-check | clean |
| `@openlinker/core` type-check | clean |
| `@openlinker/web` type-check | clean |
| eslint — `api/src/auth`, `core/src/users`, `web/src/features/users` | clean (only pre-existing warnings in untouched files) |
| Unit suites | **not re-run locally this round** — test execution is offloaded to CI on this machine. CI is the gate; flagging it rather than implying a run that did not happen. |

piotrswierzy (approved):
## Tech Lead review

Re-review at `e7b3c22`. All four IMPORTANTs from the previous round are fixed, and each was verified against the code rather than the description: `closingIssuesReferences` is empty so merging retires nothing; both new log lines are `logger.debug` naming the field and never the value (the `password-reset.service.ts:52` shape); `register-form.tsx` renders `register.error.message` with the `isConflict()` branch and the `ApiError` import gone (`isConflict()` is still used by `fulfillment-conflict.ts` and `analytics-data-coverage-panel.tsx`, so nothing is dead); and the new `describe('POST /auth/register')` asserts an identical **status and `getResponse()` body** plus a serialised-body regex against every submitted value, with a labelled characterisation test pinning the surviving 409-vs-201 channel. I re-checked the blast radius myself: the other three throw/catch sites are clean — `user.repository.ts:119` inherits the generic message, `bootstrap-admin.service.ts:88` logs its own sentence and never `error.message`, and `auth.controller.ts:171` constructs a fresh `ConflictException(error.message)`. No consumer of the old `User already exists: …` string exists outside the historical plan doc. I also independently confirmed the two new Swagger claims are true: `registration.service.ts:75` is `if (demoMode && clientIp)`, so the 429 really is demo-mode-only, and the collision branches at lines 83-89 really do return before `bcrypt.hash` at line 90. 9/9 CI checks green.

### Findings

**[SUGGESTION]** — `libs/core/src/users/domain/exceptions/user-already-exists.exception.ts:10`
> `identifier` is now a `public readonly` own **enumerable** property, so `JSON.stringify(err)` yields `{"identifier":"alice@example.com",…}` where it previously yielded `{}`. The docblock's "nothing may read it into a response" is the human half of that rule, and I confirmed it holds today — none of the seven `ExceptionFilter`s serialises an error object and none of them handles this exception anyway, since the controller catches it first. But the repo's taste elsewhere is to make a rule structural rather than conventional (`#2348`'s ordering-in-the-signature, `#2229`'s single-function reported-equals-enforced). `Object.defineProperty(this, 'identifier', { value: identifier, enumerable: false })` would make a future structured-error sink physically unable to pick it up, at the cost of one line.

**[SUGGESTION]** — `apps/api/src/auth/registration.service.ts:103,106`
> Pre-existing, untouched by this PR, and flagged only because the PR's own new comment now argues the opposite two lines above it. The success path logs `New user registered and pending approval: ${username}` / `Demo account registered; confirmation email sent: ${username}` at `log` level — an attacker-supplied identifier written verbatim on the same `@Public()` endpoint that, per this PR's own 429 note, has no throttle at all outside demo mode. The new rejection-path comment justifies `debug` + no-value as "both PII-into-logs and a log-flooding vector an unauthenticated caller controls"; both of those apply to the success path too, and arguably more so since each also writes a row. A registered username is legitimately more operator-relevant than a rejected attempt, so this may well be the right call — but a reader meeting the new comment will immediately see its neighbours contradict it. Either bring them into line or say in the comment why success is treated differently.

**[SUGGESTION]** — `apps/api/src/auth/auth.controller.spec.ts:337`
> The controller spec asserts the `ConflictException` that is *thrown*, not the bytes that reach the wire. That is the correct closest layer for a unit test and it is what the previous round asked for, so this is not a gap in the PR. Worth noting only that `apps/api/test/integration/user-email-uniqueness.int-spec.ts` already exists and already boots the users slice, so a supertest assertion that the two collisions produce an identical HTTP body would be cheap and would close the last inch — the assumption that Nest's default filter renders `getResponse()` verbatim. Low priority; the assumption is true today.

**[SUGGESTION]** — `apps/web/src/features/users/components/register-form.tsx:105`
> Rendering `register.error.message` makes a string authored in `libs/core` into operator-facing copy, which `scripts/check-ui-vocabulary.mjs` can never see (it reads JSX text literals and `*.copy.ts`, and this is an expression). `docs/architecture-overview.md § 20` states the preferred shape — every field on the wire is a code, and operator copy is the frontend's. The honest reading is that this change is *strictly better* than the hardcoded sentence it replaces (which was both an enumeration claim and factually wrong for a username collision), and there is no code on this exception's wire shape to key off. Noted as the direction a future pass would take, not as a change requested here.

### Positive observations

- The characterisation test is the right instrument and is labelled as such in its own comment — it cannot close the 409 oracle, but it makes closing it a deliberate act and reopening it a failing test. That is exactly what the previous round was asking for.
- Stating the scope limit in the exception docblock, the 409 Swagger description and the PR body means a reader meets it wherever they enter, rather than only in a description that is read once.
- The "Withdrawn claim" section retracting the `resendConfirmation`/`forgotPassword` parity assertion, and the correction of the "grepped repo, none found" line to an explicit known-stale-text note, are both the right call.

### Verdict

✅ **Approve**

Everything raised is a SUGGESTION; none blocks merge. The one thing to keep true after merge: **#3156 must stay open**, and the header of the follow-up is the always-201 + notify-by-email shape **plus** unconditional hashing, since the collision branch demonstrably returns before `bcrypt.hash` and would defeat step one on its own.

---

PR #3155, opened 2026-09-10
design(listings): price-changes review queue mockup

norbert-kulus-blockydevs:
## Summary
- Adds `docs/plans/mockups/price-changes-review-queue.html` — the reviewed, interactive HTML mockup for the price-changes review queue, per `docs/frontend-architecture.md` § UX Mockups.
- Self-contained, no external calls beyond the Google Fonts `<link>` already in the file — every `data-preview` state (live/readonly/loading/empty/error) and every `data-state`-tagged container/dialog/row is clickable and verified working (verified via Playwright against a headless Chromium during design).
- Reviewed Artifact (design-history reference, not the source of truth once this merges): https://claude.ai/code/artifact/3895933b-4f60-4935-b43b-84eb99e437b1

**Stack position**: second in this epic's stacked-PR chain (#3139). This branches from #3154 (the ADR PR, #3140). Once #3154 merges to `main`, retarget this PR's base to `main`.

Closes #3141

## Test plan
- [ ] Open `docs/plans/mockups/price-changes-review-queue.html` directly in a browser (`file://`) — renders standalone, no build step
- [ ] Click through every `data-preview` state (Live / Read-only / Loading / All caught up / Error) and confirm each renders correctly
- [ ] Exercise the core interactive flows by hand: accept a row, edit a row's price, bulk-select + accept, open the connection settings page and toggle a mode, open the source connection rollup page
- Note: pre-commit hook was skipped (`--no-verify`) for this commit with explicit user authorization — same pre-existing, unrelated `apps/e2e` lint failure noted on the ADR PR (#3154).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #3149, opened 2026-09-10
feat(web): connection Pricing & sync settings page


norbert-kulus-blockydevs:
## Problem / Context

Part of #3139 (epic). **Stack position: tenth — branches from #9's branch.**

Build the destination connection's "Pricing & sync" settings page from `docs/plans/mockups/price-changes-review-queue.html`'s `#page-connection` (`data-state="page-connection"`). Backend contract: #7.

## Proposed Solution — cite mockup `id`/`data-state`

Extend the existing connection detail page (`apps/web/src/features/connections/`) — this is a SECTION on that page, following the exact `StockAndPricingSection` pattern (#2610, `apps/web/src/features/connections/components/stock-and-pricing-section.tsx`) — read that file before starting, it's the real precedent this issue must match stylistically and structurally (explicit Save/Discard, not instant-apply).

**Default rule section** (mockup lines ~526-547):
- `#conn-mode-switch` — two-button segmented control: `#default-mode-manual` / `#default-mode-automatic` (`data-testid="default-mode-manual"/"default-mode-automatic"`). **Only two options** — no "Daily digest" button (cut per ADR decision 3).
- `#conn-mode-hint` — dynamic text under the switch: Manual = "New prices wait for your OK on the Price changes page before they go live.", Automatic = "New prices go live the moment they change in your shop — no review, no waiting."
- `#conn-rule-note` — plain-language rule summary (NOT a formula — "Prices it so you keep a 22% margin, then rounds up so it ends in .99." — build this sentence server-side or via a shared pure formatter, do not hand-write per-rule-type template strings in the component; check if #6/#7's API already returns a pre-composed summary string vs. structured `{type, pct, rounding}` for the FE to compose — confirm with #7 which approach it took and match it).
- `#conn-edit-rule` button toggles `#conn-rule-form` (hidden by default) — three fields: `#rule-type` (select: "Use shop price as-is" / "Add a percentage" / "Keep a percentage margin" — these are the real copy for `passthrough`/`markup`/`margin`, use verbatim), `#rule-pct` (percent input, hidden/disabled when type is `passthrough`), `#rule-rounding` (select: "No rounding" / "Round to a whole number" / "Round to end in .99").
[...]

Issue #3143, opened 2026-09-10
feat(listings): price-change detection + currency-mismatch blocking


norbert-kulus-blockydevs:
## Problem / Context

Part of #3139 (epic). **Stack position: fourth — branches from #3's branch.**

With the schema from #3 in place, nothing yet DETECTS that a price changed, or computes/persists the currency-mismatch `blockReason` (ADR decision 4). This must hook into wherever the master catalog price is currently synced.

## Proposed Solution

**1. Detection hook.** Find the real write path where a product's master price is written from a `ProductMasterPort`/`MasterProductSyncService` sync (`libs/core/src/products/`) — confirm the exact call site during implementation, it wasn't in this epic's original audit scope. After a master price write, for every active destination connection this product/variant is mapped to (via `OfferMapping`/`ShopProductMapping`), for every source this destination has an open-episode-eligible relationship with:
   - Resolve the effective rule via `readPricingRuleForSource` (#3)
   - Compute the new destination price via the existing `applyPricingRule`
   - Compare against the episode's `computedOldAmount` (or, if no open episode exists, against the LAST APPLIED price — read from the most recently `resolved`/`accepted` episode, or from the offer/shop-product's own last-known price if no episode has ever existed for this key)
   - If different: `upsertOpen` the episode (creates or updates the SAME open row — never a duplicate, per #3's partial unique index)
   - If the connection/source pair's mode (default or override) is `automatic`: do NOT create a reviewable episode at all — instead enqueue directly for the apply job (#4 sub-issue, i.e. epic's #4) and record it in the lightweight "recently applied automatically" log the frontend reads (#11)

**2. Currency-mismatch blocking.** Before creating/updating an episode, check `Connection.config.currency` (source) against the destination connection's currency. If they differ: write `blockReason: 'currency-mismatch'` on the episode instead of proceeding to a normal pending state. Follow the `SalesDocumentBlockOutcome` discipline exactly (`libs/core/src/sales-documents`, referenced extensively in `docs/architecture-overview.md` § Sales Documents): a **pure function** decides the block reason, the **owning service** persists it, it is **re-decided on every relevant transition** (so a later currency fix clears it), and it is **never inferred client-side** — the API (#6) must expose `blockReason` as a field, not force the frontend to recompute it.

[...]

PR #3158, opened 2026-09-10
feat(listings): price-change episode schema + per-source pricing rule model

### LINKED ISSUE #3142: feat(listings): price-change episode schema + per-source pricing rule model
## Problem / Context

Part of #3139 (epic). **Stack position: third — branches from the mockup sub-issue's branch.**

No entity in the repo stores "the last price OpenLinker computed/published for `(productVariant, destinationConnection, sourceConnection)`" — required to detect that a price "changed" at all. `Connection.config.pricingRule` (`readPricingRule`/`applyPricingRule`, `libs/core/src/identifier-mapping/domain/types/pricing-rule.types.ts`) is a flat single-rule JSONB shape with no per-source dimension. Per the ADR (#0) decisions 1 and 2.

## Proposed Solution

**1. `price_change_episodes` entity + migration**, in `libs/core/src/listings/` following the existing hexagonal layout (`domain/entities/`, `infrastructure/persistence/entities/*.orm-entity.ts`, `infrastructure/persistence/repositories/`). Columns (draft — refine against `ReservationShortfallEpisode`'s actual shape at implementation time, `libs/core/src/inventory/domain/entities/reservation-shortfall-episode.entity.ts`):
   - `id`, `productVariantId`, `destinationConnectionId`, `sourceConnectionId`
   - `sourceOldAmount` / `sourceNewAmount` / `sourceCurrency` (what changed at the master)
   - `computedOldAmount` / `computedNewAmount` (what `applyPricingRule` resolves to for this connection/source pair)
   - `manualPriceOverride` (nullable) / `manualPriceOverrideSetAt` (nullable) — the pinned/custom price mechanism, ADR decision 4
   - `blockReason` (nullable string) — ADR decision 4/5, written by #4, not this issue
   - `detectedAt`, `resolvedAt` (nullable — NULL means open), `resolution` (`'accepted' | 'accepted-custom' | 'ignored'`, nullable, set alongside `resolvedAt`)
   - `resolvedByUserId` (nullable)
   - **Partial unique index**: `(productVariantId, destinationConnectionId, sourceConnectionId) WHERE resolvedAt IS NULL` — at most one OPEN episode per key, mirroring `ReservationShortfallEpisode`'s own partial-unique-on-open-row shape exactly.

**2. Extend `Connection.config.pricingRule` to a default + per-source-override shape.** Current: a flat `PricingRule` object. New: `{ default: PricingRule, sourceOverrides?: Record<string /* sourceConnectionId */, PricingRule> }`. This is a config-shape change, not a schema migration (JSONB column, no DDL) — but every existing reader of `Connection.config.pricingRule` (`OfferBuilderService.buildCreateOfferCommand`, `ProductPublishBuilderService.buildPublishProductCommand`) must be updated to resolve through the new shape, defaulting an absent/legacy flat-shape value to `{ default: <old value>, sourceOverrides: {} }` for backward compatibility with existing connections (do not require a backfill migration — resolve at read time, matching the `readStockSafetyBuffer` config-coercion precedent, #1844).

**3. `readPricingRuleForSource(connection, sourceConnectionId): PricingRule`** — the resolution helper: `sourceOverrides[sourceConnectionId] ?? default`. Pure function, colocated with `readPricingRule`/`applyPricingRule` in `pricing-rule.types.ts`, per the project's "pure-rule exception to types-only files" convention (`docs/engineering-standards.md` § The pure-rule exception).

**4. Repository port** (`PriceChangeEpisodeRepositoryPort` in `domain/ports/`) with at minimum: `findOpenByKey(variantId, destConnId, srcConnId)`, `upsertOpen(...)` (the episode-pattern "update the same open row on re-detection" operation — a conditional UPDATE-or-INSERT against the partial unique index, `ReservationShortfallEpisode`'s conflict-arm shape), `resolve(id, resolution, userId)`, `findOpenForConnection(destConnId, filters)`.

Do NOT build detection logic, the apply job, or API endpoints here — those are #4/#5/#6/#7. This issue is schema + the pure rule-resolution helper + the repository port + its ORM/TypeORM implementation only.

## Classification

**Type**: CORE
**Layer**: Domain / Infrastructure
**File(s)**: `libs/core/src/listings/domain/entities/price-change-episode.entity.ts`, `libs/core/src/listings/infrastructure/persistence/entities/price-change-episode.orm-entity.ts`, `libs/core/src/listings/infrastructure/persistence/repositories/price-change-episode.repository.ts`, `libs/core/src/listings/domain/ports/price-change-episode-repository.port.ts`, `libs/core/src/identifier-mapping/domain/types/pricing-rule.types.ts` (extend), a new migration under `apps/api/src/migrations/`

## Dependencies

- #2 (mockup committed) — branch stacks on it
- Follow `docs/migrations.md` for the migration workflow; run `pnpm --filter @openlinker/api migration:generate`

## Assumptions

- Column list above is a draft to refine against the real `ReservationShortfallEpisode` shape during implementation — the AC below is about the SHAPE properties (partial unique index, resolution model), not the exact column names.
- `sourceConnectionId` is always populated (no NULL-source episodes) — every price change has a known master/source connection in this feature's scope.

## Acceptance Criteria

- [ ] `price_change_episodes` table exists with a partial unique index enforcing at-most-one-open-row per `(productVariantId, destinationConnectionId, sourceConnectionId)`
- [ ] `readPricingRuleForSource` resolves override-then-default correctly, with unit tests covering: override present, override absent (falls back to default), and legacy flat-shape `Connection.config.pricingRule` (backward compatibility)
- [ ] Repository `upsertOpen` correctly updates an existing open episode's `sourceNewAmount`/`computedNewAmount` rather than creating a duplicate row, with an integration test proving it against the real partial unique index (Postgres via Testcontainers, per `docs/testing-guide.md`)
- [ ] `pnpm --filter @openlinker/api migration:show` reports no pending migrations after this PR
- [ ] Tests added for non-trivial logic
- [ ] No CORE ↔ Integration boundary violations

norbert-kulus-blockydevs:
## Summary

Part of #3139 (epic). **Stack position: third of ten** — branches from `3141-price-changes-mockup` (#3155, still open at the time of this PR; per the epic's stacked-PR workflow, base moves to `main` once that PR merges).

- **`price_change_episodes`** (`libs/core/src/listings`): a new episode aggregate mirroring `ReservationShortfallEpisode` exactly — `UQ_price_change_episodes_open` is a partial unique index on `(productVariantId, destinationConnectionId, sourceConnectionId) WHERE "resolvedAt" IS NULL`, so at most one open episode exists per key and a re-detection updates that same row (`upsertOpen`, raw `INSERT ... ON CONFLICT DO UPDATE`) rather than duplicating it. Columns per the issue's draft, refined against the real `ReservationShortfallEpisode` shape (numeric amounts, `blockReason`, `manualPriceOverride`/`manualPriceOverrideSetAt` for the pinned-price mechanism, `refreshedAt` as the staleness marker consumed by #3143/#3145's accept-guard).
- **`Connection.config.pricingRule`** extended to a `{ default, sourceOverrides }` shape (ADR-072 decision 2). `readPricingRuleConfig` coerces the legacy flat shape into `{ default: <that rule>, sourceOverrides: {} }` at read time — no backfill migration. `readPricingRule` (both existing call sites, `OfferBuilderService`/`ProductPublishBuilderService`) is **unchanged in behavior** — it still resolves only the connection's default rule, since neither call site has a source-connection context to resolve against. `readPricingRuleForSource(config, sourceConnectionId)` is the new pure helper #3143's detection service will use.
- **`Connection.config.priceSyncMode`** (new, sibling key): the manual/automatic axis (ADR-072 decision 3), same default+override shape, `readPriceSyncModeForSource`. Kept as an independent key from `pricingRule` since "how to price" and "whether to ask before publishing" are orthogonal settings an operator sets independently on #3149's page.
- Repository port + TypeORM implementation, wired into `ListingsModule`/`listings.tokens.ts`.
- Migration `1878000000000-create-price-change-episodes.ts`.

## Test plan

- [x] `pnpm --filter @openlinker/core type-check` — clean on every file this PR touches (pre-existing, unrelated `@openlinker/shared` dist-staleness errors on `main` are untouched by this diff — verified by stashing and re-running).
- [x] Unit tests added: `readPricingRuleForSource` (override present / absent / legacy flat-shape backward compat), `price-sync-mode.types.spec.ts`.
- [x] Integration test (`apps/api/test/integration/listings-price-change-episode.int-spec.ts`) proving `upsertOpen`'s conflict-arm write against the real partial unique index (Testcontainers Postgres): first detection opens, re-detection refreshes the SAME row, resolution frees the key for a fresh episode, `resolve`/`reopenIgnored` guards.
- [ ] `pnpm --filter @openlinker/api migration:show` — **could not run**: no local Postgres reachable in this environment (`ECONNREFUSED 127.0.0.1:5433`, dev stack not up). The migration mirrors `1861000000000-create-reservation-shortfall-episodes.ts`'s exact shape/style; please re-run this check in CI or against a running dev stack before merge.

Per this session's user-approved exception, this commit used `git commit -s --no-verify` — the local pre-commit hook fails on a pre-existing, unrelated `apps/e2e` `node_modules` gap (same condition already hit and approved for #3154/#3155), not on anything in this diff.

Closes #3142


apps/api/test/integration/listings-price-change-episode.int-spec.ts
```diff
@@ -0,0 +1,183 @@
...
+ * creating a duplicate — asserted against the real partial unique index
+ * (`UQ_price_change_episodes_open`), not a mock.
+ *
+ * @module apps/api/test/integration
+ */
+import type { IntegrationTestHarness } from './setup';
+import { getTestHarness, teardownTestHarness } from './setup';
+import {
+  PRICE_CHANGE_EPISODE_REPOSITORY_TOKEN,
+  type PriceChangeEpisodeRepositoryPort,
```
norbert-kulus-blockydevs, line 16: **BLOCKING — `pnpm lint` fails.**

```
node scripts/check-cross-context-imports.mjs
  apps/api/test/integration/listings-price-change-episode.int-spec.ts:16
    import: { PriceChangeEpisodeRepositoryPort } from '@openlinker/core/listings'
    rule:   matches deny pattern RepositoryPort$
```

`package.json` has `"lint": "pnpm -r lint && pnpm check:invariants"`, so CI lint is red. The repo convention for an int-spec that legitimately needs a repository port is an explicit `ALLOW_LIST` row with a reason comment (see the `returns-ingestion.int-spec.ts` and `paginated-total-split.int-spec.ts` rows). Add one here — *"the vertical slice asserts the partial-index conflict arm, which only the repository can drive"* — rather than leaving the gate failing.

The other three violations in the stack belong to #3161 and #3163 and need a service seam, not an allow-list row.

libs/core/src/listings/infrastructure/persistence/repositories/price-change-episode.repository.ts
```diff
@@ -0,0 +1,224 @@
...
+        resolvedByUserId,
+        resolvedAt,
+        manualPriceOverride: manualPriceOverride === null ? null : String(manualPriceOverride),
+        manualPriceOverrideSetAt: manualPriceOverride === null ? null : resolvedAt,
+      }
+    );
+    return (result.affected ?? 0) > 0;
+  }
+
+  async reopenIgnored(id: string): Promise<boolean> {
```
norbert-kulus-blockydevs, line 237: **BLOCKING — this can raise a raw 23505 and surface as HTTP 500 on an ordinary operator sequence.**

Setting `resolvedAt = NULL` puts the row back into `UQ_price_change_episodes_open`. Reachable sequence: operator ignores episode **E** → E leaves the index → detection opens a fresh episode **F** for the same `(variant, destination, source)` key (this PR's own int-spec proves a resolved row frees the key) → operator clicks Undo on E → unique violation.

`reopenIgnored` has no `try/catch`, `PriceChangesService.unresolve` has none, and #3162's `wrapDomainErrors` catches only the four `PriceChange*Exception` types — so the operator gets a 500 naming a Postgres constraint.

This is also the one place the "mirrors `ReservationShortfallEpisode` **exactly**" claim is false in a way that matters: the precedent has **no reopen at all** ("a recurrence after a close mints a NEW id"), so it never has to reconcile a resurrection with the open-row invariant.

Fix: add a `NOT EXISTS (… "resolvedAt" IS NULL)` conjunct so a rival open episode yields `false` → 409, **or** catch 23505 matched on the constraint name (per #2392's rule that catching every 23505 would mis-report a PK collision) and raise a typed `PriceChangeEpisodeSupersededError`. Cover it with an ignore → detect → unresolve int-spec.

libs/core/src/listings/infrastructure/persistence/repositories/price-change-episode.repository.ts
```diff
@@ -0,0 +1,224 @@
...
+        destinationConnectionId,
+        sourceConnectionId,
+        resolvedAt: Not(IsNull()),
+      },
+      order: { resolvedAt: 'DESC' },
+    });
+    return row ? this.toDomain(row) : null;
+  }
+
+  async upsertOpen(
```
norbert-kulus-blockydevs, line 93: **IMPORTANT — no domain-error conversion, and an unguarded `rows[0]`.**

`docs/engineering-standards.md § Error Handling`: *"Repositories must throw domain errors, not infrastructure errors … Never expose infrastructure-specific error types (TypeORM, database) through ports."* The claimed precedent does exactly that — `ReservationShortfallRepository.openEpisode` wraps its raw INSERT in `catch (QueryFailedError)` and rethrows `ReservationLedgerConstraintError`. `upsertOpen`, `resolve` and `reopenIgnored` have no catch at all, so `CHK_price_change_episodes_amounts_non_negative` and the `UQ_…_open` collision above both reach the HTTP boundary raw.

Second, smaller: `const row = rows[0]` with no length check. The precedent guards (`if (rows.length === 0) return null`) and its `raw()` docblock warns at length that `INSERT … RETURNING *` and a `SELECT` "do NOT come back in the same shape". On an empty reply this throws `TypeError: Cannot read properties of undefined`.

libs/core/src/listings/infrastructure/persistence/repositories/price-change-episode.repository.ts
```diff
@@ -0,0 +1,224 @@
...
+    // Guarded to only reverse an 'ignored' resolution — an already-published
+    // price (accepted / accepted-custom) is never un-published by Undo.
+    const result = await this.episodes.update(
+      { id, resolution: 'ignored' as PriceChangeResolution, resolvedAt: Not(IsNull()) },
+      { resolution: null, resolvedByUserId: null, resolvedAt: null }
+    );
+    return (result.affected ?? 0) > 0;
+  }
+
+  async countOpen(filters?: PriceChangeEpisodeFilters): Promise<number> {
```
norbert-kulus-blockydevs, line 280: **IMPORTANT — materialises every open episode to produce an integer, and the only caller does it in a loop.**

`countOpen` is `(await this.findOpen(filters)).length` — `getMany()` plus full `toDomain` hydration per row. #3163's `connection-pricing-sync.service.ts` calls it **once per source inside a `for` loop**, immediately after `findOpenForConnection` has already loaded the very rows those counts are derivable from. For a destination with S sources and N open episodes that is 1 + S full reads to compute S integers.

This is the shape `docs/engineering-standards.md § When A Paginated Total Is Expensive` exists for, and #2083's rule states it directly: the batched read is called *once*, before the loop, never inside it.

Fix: SQL `COUNT` when no derived filter is set, plus a batched `countOpenBySource(destinationConnectionId): Promise<Map<string, number>>` (one `GROUP BY "sourceConnectionId"`) so #3163's loop becomes one read.

libs/core/src/listings/domain/ports/price-change-episode-repository.port.ts
```diff
@@ -0,0 +1,82 @@
...
+   * already-open episode (as opposed to opening a fresh one) — the caller
+   * uses that to decide whether to stamp `refreshedAt` (#3143's staleness
+   * marker), which only makes sense on the refresh path.
+   */
+  upsertOpen(
+    input: UpsertOpenPriceChangeEpisodeInput
+  ): Promise<{ episode: PriceChangeEpisode; wasRefresh: boolean }>;
+
+  /** Every open episode for a destination connection, paged + filtered for the review queue. */
+  findOpenForConnection(
```
norbert-kulus-blockydevs, line 67: **IMPORTANT — the docblock advertises paging the signature does not have, and the operator-facing list is unbounded.**

"Every open episode for a destination connection, **paged** + filtered" — but there is no `limit`/`offset` anywhere in the chain, and `findOpenAll` reads every open episode install-wide. #3162's `GET /listings/price-changes` then fans that unbounded set into three further full-set reads and returns all of it.

One supplier price-file import across a 20k-SKU master with two destinations opens ~40k episodes, each of which stays open until a human decides. `docs/architecture-overview.md § 28 Paginated reads` is the standard, and the port is where the parameters belong.

Either add pagination here and thread it through #3162, or delete the word "paged" and file the follow-up — but don't leave the claim standing.

libs/core/src/listings/domain/ports/price-change-episode-repository.port.ts
```diff
@@ -0,0 +1,82 @@
...
+   * shape against the partial unique index).
+   *
+   * Returns the resulting row plus whether this call REFRESHED an
+   * already-open episode (as opposed to opening a fresh one) — the caller
+   * uses that to decide whether to stamp `refreshedAt` (#3143's staleness
+   * marker), which only makes sense on the refresh path.
+   */
+  upsertOpen(
+    input: UpsertOpenPriceChangeEpisodeInput
+  ): Promise<{ episode: PriceChangeEpisode; wasRefresh: boolean }>;
```
norbert-kulus-blockydevs, line 64: **IMPORTANT — `refreshedAt` has four different behaviours across three docblocks and the SQL.**

- This port: *"the caller uses `wasRefresh` **to decide whether to stamp `refreshedAt`**"*.
- The entity and ORM entity: stamped on a re-detection *"with a changed `sourceNewAmount`"*.
- The SQL: `"refreshedAt" = now()` — **unconditional**, on every conflict, and already written before the caller sees `wasRefresh`, so no caller can decide anything.

The consequence is operator-visible downstream: #3162 derives `needsRefresh: refreshedAt !== null`, nothing ever clears the column while the episode is open, and #3164 uses that flag to strip the row of **all** its actions. So one re-detection makes a row permanently unactionable while the copy tells the operator to refresh.

Pick one contract and make all four agree — and state the conflict arm's write set on the port explicitly (that `sourceOldAmount`, `computedOldAmount` and `detectedAt` are deliberately **not** updated), which #3159 already depends on by name. The `ReturnRepositoryPort.upsertFromSource` "the write set IS the contract" treatment is the precedent.

libs/core/src/listings/domain/types/price-change-episode.types.ts
```diff
@@ -0,0 +1,77 @@
...
+/** Filters for the review-queue list read. */
+export interface PriceChangeEpisodeFilters {
+  destinationConnectionId?: string;
+  sourceConnectionId?: string;
+  /** `undefined` = both directions. */
+  direction?: 'up' | 'down';
+  /** When `true`, only episodes with `|deltaPct| >= 10` (mockup's "Big changes"). */
+  magnitudeLargeOnly?: boolean;
+  /** Exclude episodes whose variant/offer mapping is stale (#1689 precedent). */
+  excludeStaleVariants?: boolean;
```
norbert-kulus-blockydevs: **IMPORTANT — `excludeStaleVariants` is a declared safety filter that filters nothing.**

Grepped the whole tree at the top of the stack: the only occurrences are this declaration and its docblock. `findOpen` never reads it. Worse, the real stale filtering was implemented **elsewhere** — #3162's `PriceChangesService.listOpen` computes `staleVariantIds` via `IProductsService` and reports `hiddenStaleCount` — so this field is not awaiting a consumer, it is superseded and misleading. A future caller setting it would believe the #1689 stale-variant guard is applied and get episodes for master-deleted products whose offers #1689 paused to quantity 0.

(#2380 records the repo's position on this exactly: *"a control the backend cannot serve is dead code that type-checks."*)

Delete it, and note on `PriceChangeEpisodeFilters` that stale exclusion is applied above the repository. Note it is also the field #3162 would need in SQL to paginate the queue — so if you keep it, wire it.

libs/core/src/identifier-mapping/domain/types/pricing-rule.types.ts
```diff
@@ -59,23 +59,100 @@ export interface PricingRule {
...
+ * overrides) from a connection config.
+ *
+ * Backward-compatible with the pre-#3142 flat shape: `config.pricingRule`
+ * holding a bare `{type, percent, rounding}` object (no `default` key) is
+ * read as `{ default: <that rule>, sourceOverrides: {} }` — resolved at read
+ * time, matching the `readStockSafetyBuffer` config-coercion precedent
+ * (#1844). No backfill migration is required; an existing connection's
+ * behaviour is unchanged until an operator explicitly edits it.
+ */
+export function readPricingRuleConfig(
```
norbert-kulus-blockydevs, line 88: **BLOCKING (root cause; the two broken readers are outside this diff).**

Widening `config.pricingRule` to `{default, sourceOverrides}` leaves two shipped readers behind:

1. `apps/api/src/integrations/application/services/connection.service.ts:273` — `validateStockAndPricingConfig` destructures the flat shape. Under the nested one `type`, `percent` and `rounding` are all `undefined` and every check short-circuits, so **#2610's margin-≥100% refusal never fires** and `sourceOverrides` are never validated. Its own docblock says why that matters: *"core degrades a margin of 100% or more back to the catalogue price, so without this the operator saves happily and then quietly publishes an unchanged price."* A negative `markup` percent also survives and publishes `0`.
2. `apps/web/src/features/connections/components/EditConnectionForm.tsx:165` — `readPricingRuleForm` reads `raw.type` → `''`, so the section renders as unconfigured; `edit-connection.schema.ts:865` then writes `pricingRule = null` on any save of that form, destroying the default **and** every override.

Fix here: extract a `validateOneRule` and apply it to `default` plus every `sourceOverrides` entry (keeping the legacy flat arm), and teach the FE reader/serializer the nested shape. Both belong in this PR — it is the one that changes the contract.

norbert-kulus-blockydevs (commented):

**3 blocking.** The schema work itself is good — the partial-index conflict target repeating its `WHERE "resolvedAt" IS NULL` predicate, the explicit `"updatedAt" = now()` in the raw upsert (which `@UpdateDateColumn` would otherwise miss), the narrow conflict-arm write set, and `readPriceSyncModeConfig` failing closed to `manual` on every malformed input are all correct and non-obvious.

What must not ship:

1. **The #2610 server-side pricing guard silently stops enforcing under the shape this PR introduces.** `ConnectionService.validateStockAndPricingConfig` (`apps/api/src/integrations/application/services/connection.service.ts:273`, not in this diff) destructures the *flat* `{type, percent, rounding}`; under `{default, sourceOverrides}` all three are `undefined` and every check is guarded on `!== undefined`, so nothing runs. `{default:{type:'margin',percent:120}}` persists (→ `applyPricingRule` degrades to the catalogue price) and so does `{type:'markup',percent:-5000}` (→ **a published price of `0`**). `sourceOverrides` are never validated at all. ADR-072's own Cons section names this obligation verbatim: *"requires **every existing reader** to be updated with a backward-compatible fallback."*
2. **`EditConnectionForm` destroys the new shape.** `readPricingRuleForm` reads `raw.type` → `''`, so `StockAndPricingSection` renders "no pricing rule" for a connection that has one; then `mergeStructuredIntoConfig` writes `pricingRule = null` on **any** save of that form, wiping the default and every per-source override.
3. **`reopenIgnored` can violate `UQ_price_change_episodes_open` and answer 500** — see the inline comment.

Also: `price_change_episodes` is missing from `truncateTables` in `apps/api/test/integration/setup.ts` (the table carries no FK, so the cascade closure cannot reach it and it will leak between specs); and no migration-parity spec covers the new table, though `fulfillment-work-migration-parity.int-spec.ts`'s header says a context adding tables should copy it.

**Merge note:** safe to land on `main` alone once the above are fixed — nothing here depends on #3154/#3155.
---
*Part of a whole-stack review of #3158–#3168 (epic #3139). Cross-cutting items that affect this PR but are tracked once: `pnpm lint` is red on the stack (`check-cross-context-imports` ×4, `check-architecture-gates` ×1); `docs/architecture-overview.md § Listings` is untouched by all ten PRs though ADR-072 names it as the primary doc section; `docs/plans/mockups/price-changes-review-queue.html` has not changed since #3155 despite ~20 behavioural divergences, which `docs/frontend-architecture.md § UX Mockups` treats as incomplete; and four sites cite "ADR-072 decision 6/7/9" using **epic #3139's** numbering — ADR-072 has 8 decisions and they diverge from #6 onward.*

libs/core/src/identifier-mapping/domain/types/price-sync-mode.types.ts
```diff
@@ -0,0 +1,82 @@
...
+  return typeof value === 'string' && (PriceSyncModeValues as readonly string[]).includes(value);
+}
+
+/**
+ * Read the per-connection price-sync-mode config (default + per-source
+ * overrides). An absent/malformed config coerces to the safe default
+ * (`manual` everywhere) — a connection that predates this feature, or that
+ * was never configured, never auto-publishes.
+ */
+export function readPriceSyncModeConfig(
```
norbert-kulus-blockydevs, line 50: **IMPORTANT — a 7th config knob that evaded `check-architecture-gates` on a phrasing accident, and has no write-time validation.**

`scripts/check-architecture-gates.mjs` detects a config-coercion knob by the conjunction of the literal string `Connection.config` **and** an exported `read*`/`parse*` helper. This file exports two such helpers over untrusted JSONB but never writes that literal (it says "its own config key, `config.priceSyncMode`"), so the gate is silent and `KNOWN_CONFIG_KNOBS` was not updated. Registering it matters: `KNOB_THRESHOLD = 7` and the registry holds 6, so registering **fails the build** with the #1032 "reconsider a shared per-connection rules model instead of accreting knob-by-knob in JSONB" message — which is the conversation the gate exists to force. (#3159's `price-change-block.types.ts` is already flagged by the same gate at the top of the stack, so this is unavoidable for the epic either way.)

Separately: there is **no** server-side validation for `config.priceSyncMode` at all, and line 53's `typeof raw !== 'object'` means an operator typing the obvious `"priceSyncMode": "automatic"` in the raw-JSON editor gets a 200 and no effect, silently, forever. It fails *closed*, which is the right direction — but it is the reported-≠-enforced gap #2610's docblock is about. Bound it in the same traversal as the `pricingRule` fix.

norbert-kulus-blockydevs:
Addressed all 9 review findings. Summary per comment:

1. **`pnpm lint` failure (cross-context import in the int-spec)** — added an `ALLOW_LIST` row in `scripts/check-cross-context-imports.mjs` for `apps/api/test/integration/listings-price-change-episode.int-spec.ts` importing `PriceChangeEpisodeRepositoryPort` directly, with the same rationale style as the `returns-ingestion.int-spec.ts` precedent (the vertical slice needs to drive the partial-index conflict arm directly). `pnpm lint` is green.

2. **`reopenIgnored` 23505 race** — added a `NOT EXISTS` conjunct against `UQ_price_change_episodes_open`'s key columns so the common ignore → detect → undo ordering is refused with `false` rather than a raw constraint violation, and wrapped the residual race (a concurrent `upsertOpen` committing between the guard and the write) in a catch matched by constraint name (the #2392 rule), raising a new `PriceChangeEpisodeSupersededError`. Added an int-spec case reproducing the exact sequence.

3. **No domain-error conversion / unguarded `rows[0]`** — `upsertOpen`, `resolve` and `reopenIgnored` now convert every `QueryFailedError` to a domain error (`PriceChangeEpisodePersistenceError`, mirroring `FulfillmentPersistenceError`/`RoutingDecisionRepository`). `upsertOpen`'s `rows[0]` read is now guarded and throws the domain error on an empty reply instead of a raw `TypeError`.

4. **`countOpen` materialising every row** — `countOpen` now issues a real SQL `COUNT` (via a shared `buildOpenQuery` builder) for every sargable filter combination, falling back to the full read only for the two derived (non-sargable) predicates, `direction`/`magnitudeLargeOnly`. Added `countOpenBySource(destinationConnectionId)` — one `GROUP BY` read — for the per-source-loop caller the review flagged. Int-spec coverage added for both.

5. **Overclaimed "paged" docblock** — removed the "paged" claim from `findOpenForConnection`/`findOpenAll`; the docblock now states plainly that both reads are unbounded, that `countOpen`/`countOpenBySource` exist so the counters don't pay that cost, and that pagination is deferred to the #3162 HTTP surface (no downstream consumer of the port exists yet on this branch to thread it through). Filed as a stated follow-up rather than guessed at here.

6. **`refreshedAt`/`wasRefresh` inconsistency** — reconciled to one contract: the conflict-arm SQL now stamps `refreshedAt` only when the incoming `sourceNewAmount` genuinely differs from what's already stored (`CASE WHEN ... IS DISTINCT FROM ...`), matching what the entity/ORM-entity docblocks already said. `wasRefresh` now documents precisely that it reports "the conflict arm fired," which is a different fact from "refreshedAt was (re)stamped." The port docblock also states the conflict arm's write set explicitly (which columns are never touched: `sourceOldAmount`, `computedOldAmount`, `detectedAt`). Added an int-spec case proving a repeated identical price does not stamp `refreshedAt` while a genuinely different one does.

7. **Dead `excludeStaleVariants` filter** — removed from `PriceChangeEpisodeFilters`; the docblock now states that stale-variant exclusion is applied above the repository (the real #1689 guard lives in the future `PriceChangesService.listOpen`, #3162).

8. **Widened `pricingRule` broke two readers** — fixed both:
   - `apps/api/src/integrations/application/services/connection.service.ts`: split the pricing-rule validation into `validatePricingRuleConfig` + `validateOnePricingRule`, applied to both the legacy flat shape and the `#3142` nested `{default, sourceOverrides}` shape (mirroring `readPricingRuleConfig`'s own discriminator), so the margin-≥100% refusal and every other bound now fires for the widened shape too. Also added `validatePriceSyncModeConfig` for the new `config.priceSyncMode` key (no legacy shape to support — a malformed value, including a bare string, is refused rather than silently falling back to `manual`).
   - `apps/web/src/features/connections/components/EditConnectionForm.tsx` (`readPricingRuleForm`) and `edit-connection.schema.ts` (`mergeStructuredIntoConfig`'s pricing-rule clause): the FE now reads the `default` sub-object out of the nested shape (mirroring `readPricingRuleConfig`'s discriminator) instead of reading the wrapper object's own `type`/`percent`/`rounding`, and the serializer now *preserves* any existing `sourceOverrides` by re-nesting the edited default underneath them, instead of unconditionally writing the flat legacy shape (which previously destroyed every per-source override on any unrelated save).

9. **Undetected 7th config knob / no write-time validation** — registered `price-sync-mode.types.ts` (`readPriceSyncModeConfig`, `config.priceSyncMode`) in `scripts/check-architecture-gates.mjs`'s `KNOWN_CONFIG_KNOBS` and added a docblock breadcrumb so the gate's own discovery pass (which the phrasing had accidentally evaded) also finds it going forward. This trips `KNOB_THRESHOLD`, so it's raised from 7 to 8 with an explicit rationale following the #2304/#2305 precedent (the consolidation conversation is deferred to #2169 with the same three-point justification: no persistence/binding to host a shared model yet, the knob is required by ADR-072 decision 3's independent-axes design, and the raise buys exactly one more knob before the gate fires again). Write-time validation was added in the same change as part of #8's `validatePriceSyncModeConfig`.

Verification: `pnpm --filter @openlinker/core lint/type-check`, `pnpm --filter @openlinker/api lint/type-check`, `pnpm --filter @openlinker/web lint/type-check`, and `pnpm check:invariants` (including `check-cross-context-imports` and `check-architecture-gates`) all pass with 0 errors. Per instructions, `pnpm test`/integration tests were not run locally; CI will exercise the new/updated int-spec cases.

piotrswierzy (changes requested):
## `/pr-review` — 🔴 Request changes

The strongest PR in the epic, and the only one with a **green CI board** — `Lint`, `Type Check`, `Test`, `Integration Tests`, `Build`, `Docker Build Smoke Test` all SUCCESS, which is worth saying out loud because its siblings in this stack have no checks at all and their test claims are therefore unverified.

Four choices are right and were not obvious. The episode mirrors `ReservationShortfallEpisode` rather than inventing a second episode mechanism — same partial-unique-on-open-row key, same conflict-arm refresh, so a re-detection updates the row and the occurrence id survives for an edge-triggered automation to key on. The legacy flat `pricingRule` is coerced at **read** time, so there is no backfill and an existing connection is byte-identical until an operator edits it (the `readStockSafetyBuffer` precedent, cited correctly). `readPricingRule` is preserved as a thin `readPricingRuleConfig(...).default` wrapper, so both shipped call sites keep their exact behaviour rather than being rewritten to carry a source they do not have. And `priceSyncMode` is kept as an **independent key**, which is the correct modelling: "how to price" and "whether to ask before publishing" are orthogonal, and folding the second into the first would make an operator unable to set one without restating the other. The `{ default, sourceOverrides }` write-preservation in `mergeStructuredIntoConfig` — and its comment explaining that an unconditional flat write would destroy every override on the first unrelated save — is exactly the kind of trap I expected to find unguarded here and did not.

### BLOCKING — `{ sourceOverrides: … }` with no `default` key is accepted by the validator and then silently discarded by the reader

`readPricingRuleConfig` discriminates the two shapes on `'default' in candidate` (`pricing-rule.types.ts`), and `validatePricingRuleConfig` mirrors that discriminator deliberately — the docblock says so, *"so validation can never disagree with what gets read back."* It does disagree, on one input:

```json
{ "pricingRule": { "sourceOverrides": { "<src-uuid>": { "type": "margin", "percent": 150 } } } }
```

- **Validation passes.** No `default` key ⇒ the legacy branch runs `validateOnePricingRule(candidate, 'config.pricingRule')` against the *container*, whose `type`, `rounding` and `percent` are all `undefined`, so every check short-circuits. The `percent >= 100` margin refusal never runs on the override.
- **Read discards it.** Same branch taken ⇒ `coercePricingRule(candidate)` sees no valid `type` and returns `null`, `sourceOverrides` is `{}`. Every override is gone.

So a 150% margin — the value `validateStockAndPricingConfig` exists to refuse, since `applyPricingRule` falls back to the base price there — is saved with a 200 and then silently ignored. That is both halves of the failure the method's own docblock is about: *"the operator saves happily and then quietly publishes an unchanged price"*, and *"failing closed with no operator feedback is the reported-≠-enforced gap."* The raw JSON editor sits on the same page, and curl and MCP bypass the browser entirely — the docblock's own reasoning for why the server check exists at all.

**`priceSyncMode` gets this right** and is the model: `readPriceSyncModeConfig` and `validatePriceSyncModeConfig` both read `sourceOverrides` **unconditionally**, with no `default`-presence gate. The asymmetry between the two sibling keys added in one PR is the tell.

Two-line fix, and I'd take both:

1. In `readPricingRuleConfig`, discriminate on `'default' in candidate || 'sourceOverrides' in candidate` — or drop the discriminator and read `sourceOverrides` unconditionally, exactly as the price-sync reader does. A legacy flat rule can never carry either key (`PricingRule` is `{type, percent, rounding}`), so widening the test cannot misclassify one.
2. In `validatePricingRuleConfig`, mirror whichever you pick — and please add the negative unit test, since the current `pricing-rule.types.spec.ts` cases cover override-present, override-absent and legacy-flat, i.e. the three shapes that *do* work.

There is a third consequence worth fixing in the same commit: `mergeStructuredIntoConfig`'s `existingSourceOverrides` also gates on `'default' in existingContainer`, so a connection carrying the headless shape has its overrides destroyed by the very clause written to preserve them.

### IMPORTANT — `1878000000000` is claimed by three different migrations across open PRs, and the guard cannot see it

`main`'s tail is `1877000000000-add-fulfillment-progress-claims-shipped-index.ts`, so `1878000000000` is correctly the next slot per `docs/migrations.md` rule 1. But three **distinct** migrations currently claim it:

| PR | file |
|---|---|
| #3158 / #3159 | `1878000000000-create-price-change-episodes.ts` *(same file, stacked — fine)* |
| #3035 | `1878000000000-create-order-cancellation-signals.ts` |
| #3014 | `1878000000000-add-order-record-destination-routing-block.ts` |

`scripts/check-migration-timestamps.mjs` diffs only against `origin/main`, so each of the three passes in isolation and the collision surfaces as a reddened `main` for whoever merges second. (`1879000000000` is likewise shared by #3161/#3164, though those two are one stack.) Nothing to change in this diff if it merges first — but the epic needs a claimed-slot ledger, or the second merger pays for it.

### SUGGESTIONS

- `validateOnePricingRule` accepts a rule with `percent` but **no `type`** (both checks are `!== undefined`-guarded independently). Harmless today — `coercePricingRule` drops it — but it means `{percent: 150}` is a 200 that persists a rule nothing will ever read. A `type`-required check on a non-empty rule object would make accepted ≡ effective.
- The `refreshedAt` staleness marker is described as *"consumed by #3143/#3145's accept-guard"*; nothing in this diff reads it, which is correct for a schema-first slice — worth one line in the PR body saying so explicitly, since a column with no reader is otherwise indistinguishable from a forgotten one.
- The test-plan's unrunnable `migration:show` box is handled honestly. Given the collision above, re-run it against a database that has #3035 or #3014 applied rather than a clean one.

**Verified**: both reader discriminators, the four validator arms and the exact short-circuit path above; `main`'s migration tail and all six migration filenames across the open PRs; that `readPricingRule`'s two call sites are behaviourally unchanged; and the CI rollup.

---

norbert-kulus-blockydevs:
Addressed @piotrswierzy's `/pr-review` findings (commit 0f4af9832).

## BLOCKING — fixed, all 4 sub-parts

The root cause: `readPricingRuleConfig`, `validatePricingRuleConfig`, and `mergeStructuredIntoConfig`'s `existingSourceOverrides` all discriminated the nested `{default, sourceOverrides}` shape from the legacy flat shape by testing `'default' in candidate` alone. A headless `{ sourceOverrides: {...} }` container (real overrides, no `default` key) fell through to the legacy branch everywhere:

1. **`readPricingRuleConfig`** (`libs/core/src/identifier-mapping/domain/types/pricing-rule.types.ts`) — widened the discriminator to `'default' in candidate || 'sourceOverrides' in candidate`. A legacy flat `PricingRule` can never carry either key, so this can't misclassify one.
2. **`validatePricingRuleConfig`** (`connection.service.ts`) — mirrored the same widened discriminator, so a 150% margin inside a headless `sourceOverrides` container is now refused with a 400 instead of accepted-then-silently-dropped.
3. **Negative unit tests added** at all three layers: `pricing-rule.types.spec.ts` (core reader), `connection.service.spec.ts` (validator — both the "accept within bounds" and "reject >=100% margin" cases for the nested shape, plus the specific headless-container regression case), and `edit-connection.stock-pricing.test.ts` (FE merge).
4. **`mergeStructuredIntoConfig`'s `existingSourceOverrides`** — mirrored the same discriminator, so the clause that exists specifically to preserve per-source overrides no longer destroys them on the first unrelated save of a connection carrying a headless container.

Also widened `ConnectionConfig.pricingRule`'s TS type (`default` is now optional in the nested shape) so this legitimate persisted state actually type-checks — the new tests wouldn't compile against the old, `default`-required type.

## IMPORTANT — migration timestamp collision (not fixed here)

Confirmed: `1878000000000` is currently claimed by three distinct migrations across open PRs (this one, #3035, #3014). Per your note, I'm leaving this untouched in this diff — renumbering here would just move the collision, and it needs a maintainer decision at merge time about which of the three lands last and gets bumped. Flagging it here for visibility rather than guessing.

## SUGGESTIONS

- **#1 (type-required-on-non-empty-rule)** — done. `validateOnePricingRule` now refuses a rule carrying `percent`/`rounding` but no `type`, so "accepted" implies "will actually take effect" for that shape too (a bare `{}` is still left alone — it configures nothing either way).
- **#2 (`refreshedAt` docblock accuracy)** — checked: #3143/#3145 haven't landed yet (no commits for them exist), so the docblock's "consumed by #3143/#3145's accept-guard" is accurately describing a not-yet-built future consumer, phrased in the future tense ("can refuse a stale accept") — consistent with this repo's schema-first documentation convention elsewhere. Left as-is; happy to revisit once those land if the wording drifts from what actually ships.
- **#3 (migration re-check against a non-clean base)** — not practically testable in my current environment (no local DB / can't check out the sibling PRs' branches to build a merged state without more coordination). Noting it as unverified rather than guessing; the maintainer resolving the timestamp collision above is the natural point to also confirm this.

Verified independently (pre-commit hook skipped, per my setup): `pnpm --filter @openlinker/core lint` (0 errors), type-check across `@openlinker/core`/`@openlinker/api`/`@openlinker/web` (clean), and `pnpm check:invariants` (all green).

piotrswierzy (approved):
## `/pr-review` (round 2) — ✅ Approve

Every half of the previous BLOCKING finding is fixed, and fixed the way that makes the two sides unable to disagree rather than merely agreeing today.

**The reader's discriminator is now `'default' in candidate || 'sourceOverrides' in candidate`** (`pricing-rule.types.ts:110`), and the comment states the property that makes widening it safe rather than just asserting it: a legacy flat `PricingRule` is `{type, percent, rounding}` and can carry neither key, so the wider test cannot misclassify one. **The validator mirrors it verbatim** (`connection.service.ts:314`) — which was the actual requirement, since the previous defect was not that either side was wrong alone but that they were wrong *identically* and therefore silently. A headless `{sourceOverrides: {"src": {type: "margin", percent: 150}}}` now validates as a margin override and is refused at 150, instead of passing every check by virtue of every field being `undefined` and then being discarded on read.

**The FE preservation clause is fixed too** (`edit-connection.schema.ts:891`), which is the half that is easiest to forget: `mergeStructuredIntoConfig` gated `existingSourceOverrides` on the same narrow test, so the clause written to *preserve* overrides destroyed them for exactly the shape the reader could not see. It now tests both keys.

**And the negative test would genuinely fail against the pre-fix code** (`pricing-rule.types.spec.ts`, the last case): it asserts `readPricingRuleConfig` returns the override for a container with no `default`, *and* that a source with no override reads `null` rather than the container itself being mis-read as a legacy flat rule. That second assertion is the one that pins the classification rather than the happy path — the difference between a test that documents the fix and one that prevents the regression.

The SUGGESTION landed as well: `connection.service.ts:357` now refuses `{percent: 150}` with no `type`, so accepted ≡ effective.

### IMPORTANT — `1878000000000` is now claimed by THREE distinct migrations, one more than last round

`main`'s tail is still `1877000000000-add-fulfillment-progress-claims-shipped-index.ts`, so the slot is correct per `docs/migrations.md` rule 1. The contention is not:

| slot | migration | PRs |
|---|---|---|
| `1878000000000` | `create-price-change-episodes` | this stack (#3158/#3159/#3162/#3164) |
| `1878000000000` | `add-order-record-destination-routing-block` | #3014 |
| `1878000000000` | `normalise-sales-document-country-codes` | **#3201 (new since last round)** |
| `1879000000000` | `create-price-change-auto-applied-log` | #3161/#3164 |
| `1879000000000` | `create-order-cancellation-signals` | **#3035 (moved here from 1878)** |

#3035 moving off `1878` did not resolve anything — it landed on the epic's *second* slot instead, which is the shape to expect while every branch picks `main`'s tail + 1 independently. `check-migration-timestamps.mjs` diffs only against `origin/main`, so all five pass in isolation and the collision surfaces as a reddened `main` for whoever merges second. Nothing to change in this diff if it merges first; the epic needs a claimed-slot ledger, or the coordination cost lands on the last merger.

### SUGGESTION

`refreshedAt` still has no reader in this slice, which remains correct for a schema-first PR — the body now says so, which is what I asked for.

---

norbert-kulus-blockydevs:
Addressed the round-2 review.

**IMPORTANT — `1878000000000` three-way collision**: renamed the migration to `1880000000000-create-price-change-episodes.ts` (file + exported class + `name` property), clear of both currently-contended slots the review identified — `1878000000000` (this file, plus #3014's `add-order-record-destination-routing-block` and #3201's `normalise-sales-document-country-codes`) and `1879000000000` (this epic's own `create-price-change-auto-applied-log`, plus #3035's `create-order-cancellation-signals`). Grepped the repo for any other reference to the old class/file name — none found. `check-migration-timestamps.mjs` still passes (`main`'s tail is unchanged at `1877000000000`).

Per the review's own read, nothing strictly *had* to change in this diff for it to merge cleanly first — the underlying issue is epic-wide (five sibling branches independently picking `main`'s tail + 1, invisible to the per-branch check). This bump is a defensive step that removes this PR's migration from the currently-known contention entirely; it doesn't resolve the coordination gap itself, which still needs a claimed-slot ledger or falls on whoever merges last.

**SUGGESTION — `refreshedAt` reader**: already addressed before this round (the PR body states the schema-first rationale), per the review's own note ("which is what I asked for"). No further action.

Also merged the updated `3141-price-changes-mockup` base in (its tip moved with a mockup-review fix commit) and re-ran the full quality gate: `check:invariants` (all green, incl. `check-migration-timestamps`), `@openlinker/api` and `@openlinker/core` `type-check` (clean).

piotrswierzy (approved):
## Tech Lead review (round 3) — ✅ Approve

Re-reviewed independently rather than deferring to the round-2 approval. Every BLOCKING half of that round is still in place and still correct: `readPricingRuleConfig`'s discriminator is `'default' in candidate || 'sourceOverrides' in candidate` with the comment stating *why* widening is safe (a legacy flat `PricingRule` is `{type, percent, rounding}` and can carry neither key), `validatePricingRuleConfig` mirrors it verbatim so the two cannot disagree silently, `mergeStructuredIntoConfig`'s `existingSourceOverrides` tests both keys, and the negative test pins the classification rather than the happy path. The only change since that approval is the migration renumber plus a base merge.

**The renumber worked and I verified it against every open PR, not just `main`.** `1880000000000-create-price-change-episodes.ts` is now claimed by this stack alone; `1878000000000` is contended by #3201, #3014 and #3027, and `1879000000000` by #3035, #3219 and this stack's own #3161. `main`'s tail is `1877000000000-add-fulfillment-progress-claims-shipped-index.ts`, so the slot is both free and strictly greater per `docs/migrations.md` rule 3. CI is fully green on all nine checks.

Migration ↔ entity parity checks out on the axes `synchronize` cannot emit: the `CHECK` is declared class-level under the migration's exact name, the partial unique index repeats `WHERE "resolvedAt" IS NULL` on both sides, `blockReason`/`resolution` are `varchar(32)` rather than PG enums, and there are no FKs by design. The raw `upsertOpen` conflict target repeats the partial predicate (Postgres will not infer it otherwise), writes `"updatedAt" = now()` explicitly because `@UpdateDateColumn` does not fire on a raw statement, and keeps `sourceOldAmount`/`computedOldAmount`/`detectedAt` out of the `DO UPDATE SET` list so the operator keeps seeing the original baseline across refreshes. `reopenIgnored`'s `NOT EXISTS` conjunct plus the constraint-name-matched 23505 catch is the right two-layer answer, and the subquery correctly cannot match its own row (the target is guarded `resolvedAt IS NOT NULL`).

### Findings

**[SUGGESTION]** — `apps/api/test/integration/setup.ts`

> `price_change_episodes` carries no FK by design, so `truncateTables`' CASCADE closure can never reach it — the exact condition the file's own `reservation_shortfall_episodes` and `oms_routing_rules` comments exist to warn about. The shared-list entry lands in #3161, not here, and `listings-price-change-episode.int-spec.ts` truncates inline in `afterEach`, so nothing actually leaks at this commit. Flagging only so the stack ordering is deliberate: if this PR ever lands ahead of #3161 and a second suite starts writing episodes, the inline truncate covers only the spec that remembers.

**[SUGGESTION]** — `libs/core/src/listings/infrastructure/persistence/entities/price-change-episode.orm-entity.ts:19`

> The docblock says *"Every constraint is declared class-level under the SAME NAME the migration uses"*. The `CHECK` and both indexes are; the primary key is not — the migration names `PK_price_change_episodes` while `@PrimaryGeneratedColumn` makes `synchronize` mint a hash name. That is precisely the divergence `fulfillment-work-migration-parity.int-spec.ts` found on its first green run ("the migration named the primary keys while `synchronize` minted hash names"), and that file's header says a context adding tables should copy it. Either soften the claim or add the parity spec — this table is a good candidate, since its partial-unique predicate is the one thing an `indexdef` diff catches and a functional test cannot.

**[SUGGESTION]** — `libs/core/src/listings/infrastructure/persistence/repositories/price-change-episode.repository.ts:132`

> `RETURNING *, (xmax = 0) AS "wasInserted"` introduces the MVCC-internal insert/update discriminator that #2328 explicitly declined ("distinguishing insert from update would need the MVCC internal `xmax = 0` (no precedent in this tree)"). It is the standard trick for `ON CONFLICT` and I am not asking for a change — but it is now a precedent, and `wasRefresh` means *"the conflict arm fired"*, which is deliberately a different fact from *"`refreshedAt` was stamped"*. #3159 already depends on that distinction by name; worth one line on the port docblock saying the two are not interchangeable, since a future reader will reach for `wasRefresh` when they want the second.

### Verdict

✅ **Approve**

`KNOB_THRESHOLD` 7 → 8 is registered rather than exempted, which is the right direction, and the rationale block follows the #2304/#2305 form including the "a reviewer who disagrees should push back on this raise specifically" invitation. I accept it; as I said last round, the line holds at the next one.

Note the base is `3141-price-changes-mockup`, not `main` — this approval is of the diff, and the merge target still needs to move to `main` once #3155 lands.

---

PR #3159, opened 2026-09-10
feat(listings): price-change detection + currency-mismatch blocking

norbert-kulus-blockydevs:
## Summary

Part of #3139 (epic). **Stack position: fourth of ten** — branches from `3142-price-change-episode-schema` (#3158, open — base moves to `main` once that merges).

### The reverse-dependency problem, and how it's solved

The proposed solution in #3143 says to hook detection into "wherever the master catalog price is currently synced" (`MasterProductSyncService`, in the `products` context). But `products -> listings` would be a **new module-load cycle**: `listings` already imports `ProductsModule` for per-variant master stock (#824), so `products` cannot import `listings` back.

This PR resolves it the way `FulfillmentRouterBindingModule` already resolves an analogous `libs/core -> libs/oms` edge: the **port is owned by the caller** (`PriceChangeObserverPort`, defined in `products`), the **implementation lives where the logic belongs** (`PriceChangeDetectionService`, in `listings`), and a **host-side `@Global()` binding module** (one each for `apps/api` and `apps/worker`) rebinds the token at the composition root. The injection into `MasterProductSyncService` is `@Optional()`, so a host that omits the binding module degrades to "no price-change detection" — never a boot failure — and every existing `MasterProductSyncService` unit test is untouched (no observer wired -> the old-price read is skipped entirely, zero extra queries).

### What's implemented

- `MasterProductSyncService` reads each variant's prior price (only when an observer is wired) before its `upsertProduct`/`upsertVariants` call, and reports every variant whose price actually changed, best-effort (caught + logged, never fails the sync).
- `PriceChangeDetectionService` (listings), for every destination connection the changed variant is mapped to via an `Offer` or `ShopProduct` identifier mapping (`IIdentifierMappingService.getExternalIds`, both entity types are keyed by variant id — no new repository queries needed):
  - `resolvePriceChangeBlockReason` (pure, `SalesDocumentBlockOutcome` discipline) sets `blockReason: 'currency-mismatch'` when the destination's `config.currency` differs from the source's — re-decided every pass, never inferred client-side.
  - `readPricingRuleForSource` + `applyPricingRule` compute the destination price; `readPriceSyncModeForSource` picks `manual` (open/refresh a reviewable episode) vs `automatic` (bypass the queue, enqueue the new `pricing.propagateToMarketplaces` job type directly).
[...]

Issue #3145, opened 2026-09-10
feat(api): price-changes queue endpoints (list, accept, ignore, edit)


norbert-kulus-blockydevs:
## Problem / Context

Part of #3139 (epic). **Stack position: sixth — branches from #5's branch.**

Frontend sub-issues #7/#8 need HTTP endpoints to list, filter, accept, ignore, edit, and bulk-accept price-change episodes. This issue is the API surface over #3's schema and #5's apply job.

## Proposed Solution

New controller in `apps/api/src/listings/` (or wherever the existing offer/listing controllers live — follow the existing layout), guarded with `@UseGuards(JwtAuthGuard)` and `@Roles('admin', 'operator')` reusing `connections:write`-equivalent gating (ADR decision 6 — confirm the exact `@Roles`/permission mechanism against `docs/frontend-architecture.md` § Access Control And UI Visibility, since the FE reads `session.permissions[]`, not a raw role string).

Endpoints (map 1:1 to the mockup's interactions in `docs/plans/mockups/price-changes-review-queue.html`):

- `GET /listings/price-changes` — list open episodes. Query params: `connectionId` (filter, mirrors mockup `#filter-conn-*` chips), `direction` (`up`/`down`, mirrors `#filter-direction-*`), `magnitude` (`large`, mirrors `#filter-magnitude-large`, ≥10% threshold). Response includes per-row: variant/product identity, source label + old/new source amount, destination connection, computed old/new destination price + currency, `deltaPct`, `blockReason` (nullable — the API returns it, the FE never infers it), `needsRefresh` (boolean, if the episode was updated after last read — see #4's re-detection behavior), rule summary (type/pct/rounding) used to compute the price, so the FE can render the mockup's `.rule-note` sentences without re-implementing `applyPricingRule` client-side.
- `GET /listings/price-changes/stale-count` (or fold into the list response as a `hiddenStaleCount`) — mirrors the mockup's `#stale-note` ("N more change isn't shown here because that listing is paused") — episodes whose variant offer is `isStale` (#1689 precedent) are excluded from the main list but counted separately.
[...]

Issue #824, opened 2026-05-24
feat(listings,allegro): list multi-variant products as auto-grouped Allegro variant offers


piotrswierzy:
## Context

Follow-up to **#822** (variant-keyed master inventory) and **#823** (per-combination master stock). This is the **destination (Allegro) half** of multi-variant support: actually listing every variant of a multi-variant product as Allegro offers, with per-variant stock.

## Key finding — the explicit variant-set API is gone (verified against developer.allegro.pl)

Allegro's old multi-variant resources — `POST/GET/PUT/DELETE /sale/offer-variants` (the `setId` model), plus the parameter-metadata flags `options.variantsAllowed`, `variantsByColorPatternAllowed`, `variantsEqual` — were **deprecated and removed on 14 April 2026** and now return `404`. **Do NOT build toward `/sale/offer-variants`.**

Replacement model: **variants are created automatically from the Product Catalog.** A seller lists separate offers that reference the **same catalog product** and differ by a variant-distinguishing parameter; Allegro auto-groups them into one buyer-facing listing. Crucially, **stock stays per-offer** — each variant is its own offer carrying its own `stock: { available: N }`. (Confirmed in the current "list a product offer in one request" tutorial.)

## Why this fits OL's existing architecture

OpenLinker already models **1 OL variant → 1 standalone Allegro offer**, with stock written to `body.stock.available` and quantity updated per-offer via `PUT /sale/offer-quantity-change-commands/{commandId}`. Because Allegro stock is inherently per-offer, **no new "multi-variant offer" payload is needed** — the destination half is a composition/expansion problem, not a new stock model.

## Current gap

The bulk flow creates **one offer per product** (the row's single *primary variant*; the FE `productIds` are actually variant ids). A multi-variant product only lists its primary variant. There is no variant grouping and no per-variant parameter emission.

## Proposed direction (settle in `/plan`)

- **Expand** a multi-variant product in the bulk flow into **one offer per variant** (today: one per product).
- Each variant-offer **references the same catalog product** (reuse the `productCardId` thread from #808) and **emits the variant-distinguishing parameter value(s)** mapped from the OL variant's `attributes`, so Allegro auto-groups them.
- Each variant-offer's `stock.available` comes from that variant's master stock (**requires #823**).
- Per-variant stock updates need no new shape — `updateOfferQuantity` is already single-offer.

## Open questions (resolve before implementation — verify against developer.allegro.pl / sandbox, don't guess)

[...]

PR #3161, opened 2026-09-10
feat(worker): price-change apply job with bulk progress

### LINKED ISSUE #3144: feat(worker): price-change apply job with bulk progress
## Problem / Context

Part of #3139 (epic). **Stack position: fifth — branches from #4's branch.**

Once an episode is accepted (or a connection is in `automatic` mode), the new price must actually be published to the destination via `OfferFieldUpdater.updateOfferFields` (marketplace) or the shop-product update path (`ProductPublishBuilderService`-adjacent). This must be a worker job (async, not a synchronous API call), following ADR-007's status/outcome split and reusing `BulkOfferCreationBatch`/`bulk_batch_advancements` for progress (ADR decision 7/9).

## Proposed Solution

**1. `price.propagateToMarketplaces` job type** (or similar — follow `docs/architecture-overview.md` § Sync Manager job-naming convention), registered against ADR-050's lane model. Mirror `InventoryPropagateToMarketplacesHandler`'s shape (`apps/worker/src/sync/handlers/inventory-propagate-to-marketplaces.handler.ts`) as the closest real precedent, but note it does NOT persist a pending record — this job DOES, via the episode row from #3/#4.

**2. Outcome classification per ADR-007** (this epic's ADR decision 8): a deterministic refusal (e.g. the episode carries a `blockReason` that somehow wasn't caught earlier — defensive check) returns a terminal `business_failure` on the SAME attempt. A transient marketplace API error (rate limit, 5xx) throws and lets the existing lane/retry machinery (ADR-050) handle it. **Do not let a deterministic block retry 10× before dead-lettering** — this was called out explicitly during design as the failure mode to avoid.

**3. Bulk progress via `BulkOfferCreationBatch`.** When multiple episodes are accepted together (bulk apply from the frontend, #8), create a `BulkOfferCreationBatch`-shaped aggregate (or a sibling reusing the same `bulk_batch_advancements` counter table — `libs/core/src/listings/infrastructure/persistence/entities/bulk-batch-advancement.orm-entity.ts`) so the frontend's live-progress widget (mockup `#publish-progress`, `data-state="bulk-publish-progress"`, with `#publish-progress-count` / `#publish-progress-fill` / `#publish-progress-lanes` grouped per destination connection) can poll the SAME mechanism the existing `apps/web/src/pages/listings/bulk-batch-progress-page.tsx` consumes — do not invent a new polling contract.

**4. Successful apply writes back to the episode**: `resolvedAt`, `resolution: 'accepted'` or `'accepted-custom'` (if `manualPriceOverride` was set — see #3), `resolvedByUserId`.

**5. Automatic-mode path**: when #4's detection service enqueues directly (no episode), this job still runs but writes its result to the lightweight "applied automatically" log the frontend reads via #11's endpoint contract — coordinate the exact log shape with #6 (API) during implementation, since #6 exposes it.

## Classification

**Type**: Infrastructure (worker)
**Layer**: Infrastructure / Application
**File(s)**: `apps/worker/src/sync/handlers/price-change-apply.handler.ts` (or per convention), lane registration per ADR-050 (`assertFullLaneCoverage` must pass), reuses `libs/core/src/listings/infrastructure/persistence/entities/bulk-batch-advancement.orm-entity.ts` and its repository

## Dependencies

- #4 (detection + currency-mismatch blocking)

## Assumptions

- The exact lane (per ADR-050's cost-of-starvation rule) is decided at implementation time against real traffic characteristics — likely `bulk` (mirroring the catalogue-sweep precedent), but confirm against how urgently a SINGLE accepted price change should publish vs. a large batch. Do not default to `realtime` without justifying it against ADR-050's lane table.
- The marketplace write capability used is `OfferFieldUpdater.updateOfferFields` for marketplace destinations and the shop-product update path for shop destinations — confirmed accurate by the design-phase code audit (`OfferBuilderService`/`ProductPublishBuilderService` already resolve price via this exact path at create/publish time).

## Acceptance Criteria

- [ ] Accepting a single episode enqueues a job that, on success, marks the episode resolved and publishes the new price via the existing `OfferFieldUpdater`/shop-product write path
- [ ] A deterministic block (e.g. currency mismatch slipping through) reports `outcome: 'business_failure'` on the FIRST attempt — assert this with a test that the job does NOT retry
- [ ] A transient marketplace error causes a throw and is retried per the existing lane/retry ladder — assert this with a test using a mocked transient failure
- [ ] Bulk-accepting N episodes creates one `bulk_batch_advancements`-tracked batch whose progress is readable via the SAME query shape `bulk-batch-progress-page.tsx` already polls
- [ ] `assertFullLaneCoverage()` passes at worker boot (new job type is registered with a lane)
- [ ] Tests added for non-trivial logic, including both outcome-classification branches

norbert-kulus-blockydevs:
## Summary

Part of #3139 (epic). **Stack position: fifth of ten** — branches from `3143-price-change-detection` (#3159, open — base moves to `main` once that merges).

- New job type `pricing.propagateToMarketplaces` (added to `JobTypeValues` in #3143) gets its worker handler, `PriceChangeApplyHandler`, registered on the `bulk` lane (mirrors `marketplace.offer.create`'s reasoning: single-unit work that can arrive up to N wide from a bulk-accept wave — not the single-item "someone is waiting" shape `realtime` is for).
- **ADR-007 status/outcome split (ADR-072 decision 8)**: a defensive re-check of the episode's `blockReason` (belt-and-braces — neither the detection service nor #3145's future accept endpoint should ever enqueue a blocked episode) returns a terminal `business_failure` on the first attempt rather than retrying 10x before dead-lettering. A transient marketplace/shop error throws `SyncJobExecutionError` and lets ADR-050's lane/retry ladder handle it.
- **Publish mechanism**: tries `OfferManagerPort.updateOfferFields` first (catching `CapabilityNotSupportedException` to detect "this connection is a shop, not a marketplace"), falling back to a `ShopProductManagerPort.publishProduct` re-publish for shops — reading current stock via `IInventoryQueryService.getAvailabilityByVariantIds` (master-authoritative-including-0, #824's rule) and republishing at `status: 'published'`, which is correct because a price-change episode/automatic-apply only ever exists for an already-mapped, already-live listing.
- **Episode-based success** resolves the episode (`'accepted'` or `'accepted-custom'` when `manualPriceOverride` was set).
- **Automatic-mode success** (no episode — enqueued directly by #3143's detection service) writes a new `price_change_auto_applied_log` row instead — the lightweight, deliberately non-paginated stand-in for the cut "Daily digest" mode (ADR-072 decision 3), which #3145's `GET /listings/price-changes/auto-applied` will read.
- **Bulk progress reuses the existing mechanism verbatim** (ADR-072 decision 9): when the payload carries a `batchId` (to be minted by #3145's future bulk-accept endpoint via the existing `BulkListingBatchRepositoryPort.create`), the handler calls `IBulkListingProgressService.advanceBatchStatus(batchId, childId, outcome)` — the SAME at-most-once counter-advancement gate `marketplace.offer.create`'s children use. No new progress table.

## Test plan

- [x] `pnpm --filter @openlinker/core|api|worker type-check` — clean on every file this PR touches (remaining errors identical to the previously-confirmed pre-existing baseline: stale `@openlinker/shared` dist, missing `@openlinker/integrations-eparagony`/`@openlinker/integrations-fx`/`sanitize-html` deps in this environment).
- [x] Unit tests (`PriceChangeApplyHandler`): marketplace publish success + auto-applied log write, episode resolve on accept (plain and custom-override), terminal `business_failure` on a blocked episode with no publish attempt, throw-and-retry on a transient marketplace error, shop-path fallback when `OfferManager` is unsupported, bulk-progress advancement when a `batchId` is present, payload validation failure.

Per this session's user-approved exception, this commit used `git commit -s --no-verify` (pre-existing, unrelated `apps/e2e` `node_modules` gap in the pre-commit hook).

Closes #3144


apps/worker/src/sync/handlers/price-change-apply.handler.ts
```diff
@@ -0,0 +1,326 @@
...
+   */
+  private async publishPrice(payload: PriceChangeApplyPayload): Promise<void> {
+    let marketplaceAdapter: OfferManagerPort | null = null;
+    try {
+      marketplaceAdapter = await this.integrationsService.getCapabilityAdapter<OfferManagerPort>(
+        payload.destinationConnectionId,
+        'OfferManager'
+      );
+    } catch (error) {
+      if (!(error instanceof CapabilityNotSupportedException)) {
```
norbert-kulus-blockydevs: **BLOCKING — capability discovery by thrown exception, and the exception caught is a *superclass* of "the operator disabled this capability".**

`libs/core/src/integrations/domain/exceptions/capability-not-enabled.exception.ts` says it outright:

> `CapabilityNotEnabledException` … **Extends `CapabilityNotSupportedException`** so existing `instanceof` callers continue to catch both cases; **new callers that want to distinguish "not supported by adapter" from "disabled on this connection" can catch this subtype explicitly.**

This is a new caller and does not. So `getCapabilityAdapter` refusing because **the operator has not enabled `OfferManager`** is silently read as *"this connection is a shop"* — and that is the **default WooCommerce configuration**: `woocommerce-plugin.ts:79` states `OfferManager` "Defaults OFF on new connections when `InventoryMaster` is also in the manifest set", and it is.

The other configuration fails too: with `OfferManager` enabled, `WooCommerceOfferManagerAdapter` implements **only** `updateOfferQuantity`, so `isOfferFieldUpdater` is false at line 199 and a generic `Error` is thrown → retried 10× → dead. The price is never applied and the shop branch is never reached.

The repo's documented pattern is manifest-based: capability presence is answered by `supportedCapabilities` / `listCapabilityAdapters({lazy: true})`, which constructs no adapter.

**Fix:** decide the branch from `resolveAdapterMetadata` (metadata-only, works on a disabled connection — the #2353 `AuthorityStatusService` precedent) intersected with `connection.enabledCapabilities`, and additionally require `isOfferFieldUpdater` before electing the marketplace branch. Never branch on a caught exception. If a catch is kept, catch `CapabilityNotEnabledException` **first** and rethrow it.

apps/worker/src/sync/handlers/price-change-apply.handler.ts
```diff
@@ -0,0 +1,326 @@
...
+    // correct re-publish target, not a guess.
+    const command = await this.productPublishBuilder.buildPublishProductCommand({
+      internalVariantId: payload.productVariantId,
+      connectionId: payload.destinationConnectionId,
+      stock,
+      status: 'published',
+      price: { amount: payload.amount, currency: payload.currency },
+    });
+
+    await adapter.publishProduct(command);
```
norbert-kulus-blockydevs: **BLOCKING — this POSTs a brand-new product to the live storefront on every price change.**

`command.externalProductId` is what selects upsert-vs-create, and the **builder never sets it** — `ProductPublishExecutionService` does, from the `ShopProduct` identifier mapping:

```ts
// product-publish-execution.service.ts:104
const existingExternalProductId = await this.resolveExistingExternalProductId(input);
command = await this.builder.buildPublishProductCommand({...});
if (existingExternalProductId) { command = { ...command, externalProductId: existingExternalProductId }; }
```
```ts
// woocommerce-product-publisher.adapter.ts:149
const isUpsert = cmd.externalProductId != null && cmd.externalProductId !== '';
raw = isUpsert ? await this.httpClient.put(...) : await this.httpClient.post(...);
```

Here it is `undefined`, so `isUpsert` is false and the adapter creates. Nothing persists a mapping for the new product, so the *next* price change creates another one — and each retry attempt inside a single job creates another. Unbounded duplicate-product pollution, silently, reported as `outcome: 'ok'`.

The PR body's premise ("only ever exists for an already-mapped, already-live listing") is **true** — #3143 does walk `ShopProduct` mappings — but the handler never *reads* that mapping. The premise being true is what makes the bug invisible in review.

Also lost by the bypass: `variantGroup.externalParentProductId` resolution (#1836), `ShopProduct` mapping persistence, #1845's `DuplicateIdentifierMappingError` / `ShopProductMappingConflictException` handling, and the stale-mapping recovery.

**Fix:** route the shop branch through `IProductPublishExecutionService.executePublish`.

**Separately, same call:** passing only `{stock, status, price}` means `destinationCategoryIds`, `parameters` and `content` are all *derived* by the builder — so a **price-only** change re-provisions the shop's category tree, re-projects attribute parameters, and overwrites channel-specific content with the master description. #1831 made all three per-item overridable precisely because the server-derived default is not always right. `ShopProductManagerPort` has no partial-field update; that constraint belongs in ADR-072 rather than being discovered as a data-loss incident.

apps/worker/src/sync/handlers/price-change-apply.handler.ts
```diff
@@ -0,0 +1,326 @@
...
+  private async publishToShop(
+    adapter: ShopProductManagerPort,
+    payload: PriceChangeApplyPayload
+  ): Promise<void> {
+    // Master is authoritative including 0 (#824) — the zero-filled read is
+    // the right one here, since this call is about to PUBLISH a quantity.
+    const [availability] = await this.inventoryQuery.getAvailabilityByVariantIds([
+      payload.productVariantId,
+    ]);
+    const stock = availability?.totalAvailable ?? 0;
```
norbert-kulus-blockydevs: **BLOCKING — publishes `totalAvailable`, and defaults an unknown ATP to `0`.**

The contract on this read forbids exactly this (`inventory-query.service.interface.ts:42`):

> Each row also carries `availableToPromise` (#2323) … It is `null` exactly when OpenLinker does not know, which **a publishing caller must treat as "suppress the write", never as zero and never as a reason to fall back to `totalAvailable`.**

Two separate harms. Reading `totalAvailable` publishes a quantity **not net of OL's own published reservations** (#2321/#2345), so the shop over-publishes by exactly what OL has already promised. And `?? 0` written as an **absolute** quantity is the #1689 offer-pause primitive — a variant with no row, or an unknown ATP, silently **deactivates a live listing**.

The in-code comment cites the wrong rule: #824's "master is authoritative including 0" is about the master's *own* reported zero, not about OL not knowing. Both shipped publishing consumers read `availableToPromise` and **exclude** the variant on `null`.

To be precise about what is *not* violated: the operator's `stockSafetyBuffer` / `stockZeroThreshold` Controls **are** applied, because `ProductPublishBuilderService:172` calls `applyPublishControls` and throws `AvailabilityUnknownError` on a null result — so `no-direct-buffer-read.spec.ts` is intact. It is the ATP half that is broken.

**Also:** this path writes the whole product body *including quantity* with no `observedAt` and never advances the `inventory.offerQuantity.observedAt:offer:{id}` mark — so it bypasses ADR-067/#2617's write-ordering guard **and** poisons it: the guard will subsequently refuse a genuinely newer quantity write believing something newer is live.

apps/worker/src/sync/handlers/price-change-apply.handler.ts
```diff
@@ -0,0 +1,326 @@
...
+    const payload = this.getPayload(job);
+
+    this.logger.log(
+      `Executing pricing.propagateToMarketplaces job ${job.id} for connection ${job.connectionId} ` +
+        `(variant=${payload.productVariantId}, amount=${payload.amount} ${payload.currency}, automatic=${payload.automatic})`
+    );
+
+    // Defensive re-check (ADR-072 decision 8): a currency-mismatch block must
+    // never retry 10x before dead-lettering — it can never succeed.
+    let previousOldAmount: number | undefined;
```
norbert-kulus-blockydevs: **BLOCKING — every auto-applied log row records `oldAmount === newAmount`.**

`previousOldAmount` is assigned only inside `if (payload.episodeId)` (line ~103) and read only inside the `else` branch (line ~153). The two are mutually exclusive on the same condition, so it is **always `undefined`** where it is read and the row persists `oldAmount = payload.amount = newAmount`.

It is operator-visible: #3168's `auto-applied-dialog.tsx` renders `<s>{formatAmount(item.oldAmount)}</s> {formatAmount(item.newAmount)}`, so every entry reads `~~399.00~~ 399.00` — the "was → now" that is the entire content of ADR-072 decision 3's digest stand-in.

Two further problems in the same three lines: `sourceOldAmount` is the **source/master** price while `newAmount` is the **computed destination** price, so even on the episode path the pair is not comparable (`computedOldAmount` is the right field); and #3143's producer carries no old amount at all, so on the automatic path the value is structurally unknowable.

**Fix:** add `previousAmount` to the payload, populate it in `enqueueAutomaticApply` from the observation's computed old amount, and make it required for the log write rather than defaulting.

norbert-kulus-blockydevs (commented):

The scaffolding is solid and in places better than average: payload validation narrows every required field with a type check and returns a fresh literal rather than the cast input; `business_failure` is returned rather than thrown for the `blockReason` case with `advanceBulkProgress(…, 'failed')` before it; lane registration is a single call site with a comment justifying the choice against ADR-050 decision 1; `isOfferFieldUpdater` **is** used rather than trusting the type; and the bulk-progress reuse is genuine — I verified `advanceBatchStatus`'s counter-based terminal derivation has no `OfferCreationRecord` dependency and `getBatch` tolerates empty `records`, so a price-change batch *can* terminalise.

But the terminal write does not work. **No price change can currently reach a WooCommerce destination correctly in either capability configuration**, and one of the two paths creates unbounded duplicate products on the operator's live storefront while reporting success. Five blocking items, all inline.

Note the ordering dependency between the first two: today WooCommerce reaches the shop branch only *via* the capability misclassification, so fixing that alone converts "the price never applies" into "duplicate products are created." They must be fixed together.
---
*Part of a whole-stack review of #3158–#3168 (epic #3139). Cross-cutting items tracked once: `pnpm lint` is red on the stack (`check-cross-context-imports` ×4, `check-architecture-gates` ×1); `docs/architecture-overview.md § Listings` is untouched by all ten PRs though ADR-072 names it the primary doc section; the mockup has not changed since #3155 despite ~20 behavioural divergences; and four sites cite "ADR-072 decision 6/7/9" using **epic #3139's** numbering — ADR-072 has 8 decisions and they diverge from #6 onward.*

apps/worker/src/sync/handlers/price-change-apply.handler.ts
```diff
@@ -0,0 +1,326 @@
...
+} from '@openlinker/core/inventory';
+import type { OfferManagerPort, ShopProductManagerPort } from '@openlinker/core/listings';
+import {
+  isOfferFieldUpdater,
+  OFFER_MAPPINGS_SERVICE_TOKEN,
+  PRODUCT_PUBLISH_BUILDER_SERVICE_TOKEN,
+  PRICE_CHANGE_EPISODE_REPOSITORY_TOKEN,
+  PRICE_CHANGE_AUTO_APPLIED_LOG_REPOSITORY_TOKEN,
+  BULK_LISTING_PROGRESS_SERVICE_TOKEN,
+  PriceChangeEpisodeRepositoryPort,
```
norbert-kulus-blockydevs: **BLOCKING — two cross-context `*RepositoryPort` imports; `check-cross-context-imports` fails, so `pnpm lint` is red.**

`docs/architecture-overview.md § Cross-context dependencies` lists `*RepositoryPort` as an explicit deny: *"Intra-context contract — exposes persistence concerns the source context controls. Cross-context callers go through `I*Service`."*

The underlying cause is placement, not the import line. `docs/architecture-overview.md § 7 Sync Manager`: *"Sync orchestration policies live in core application services … not in worker handlers."* `BulkListingProgressService`'s own header restates it: *"the terminal-status derivation rule lives here, with the handler reduced to a thin shell."* This handler is 326 lines of capability selection, mapping lookup, availability read, command building, adapter dispatch, episode resolution, log write and progress advancement.

**Fix:** move the orchestration into a `listings` application service (an `IPriceChangeApplyService` beside `PriceChangeDetectionService`), inject the repositories there, and reduce the handler to payload-narrow → call → map outcome. That resolves this finding, makes the defects above testable in the layer that owns them, and is strictly better than an ALLOW_LIST row, which would silence the symptom and keep the misplacement.

apps/worker/src/sync/handlers/price-change-apply.handler.ts
```diff
@@ -0,0 +1,326 @@
...
+    }
+
+    await this.advanceBulkProgress(payload, 'succeeded');
+
+    return { outcome: 'ok' };
+  }
+
+  /**
+   * Publishes the amount via whichever capability the destination supports —
+   * `OfferFieldUpdater.updateOfferFields` for a marketplace, or a
```
norbert-kulus-blockydevs: **IMPORTANT — `UpdateOfferFieldsReport.frozenFields` is discarded, so a dropped price is recorded as applied.**

`offer-field-updater.capability.ts` states why the return type exists:

> A destination that can refuse a field — Erli drops anything the seller froze, #988 / ADR-025 §4b — reports it in an `UpdateOfferFieldsReport` so a caller can tell "written" from "silently dropped"; the tax-rate journal (#2250) depends on that distinction **to avoid recording a write that never happened**.

The return is ignored, and the handler then resolves the episode `'accepted'` or writes an auto-applied log row — recording exactly the write that never happened. ADR-072 decision 5 discusses `frozen` explicitly, which makes the omission more surprising here than elsewhere.

**Also, five permanent conditions are classified transient and burn the full retry ladder** (episode not found, `ProductPublisher` unsupported, adapter has no `updateOfferFields`, no offer mapping, a marketplace 4xx). Each costs `maxAttempts` with backoff plus a dead row, and `lastError` blames a transient failure. Nothing registers this path in the per-plugin retry-classifier registry, so `isNonRetryableError` cannot rescue it. Return `business_failure` for the structural cases — and **advance the batch with `'failed'`** on each, or a dead-lettered child leaves the batch `running` forever and #3165's widget spins indefinitely.

**And there is no claim marker:** publish → resolve/log → advance, with a `SIGKILL` between the first two leaving the episode open with the price already live. The comparable shipped paths (`waybillRelayedAt` #1947, `reservationConsumedAt` #2347, `dispatchRelayedAt` #2401) all take a conditional claim before crossing the boundary. `episodes.resolve` also returns a guarded `boolean` that is discarded, so a lost race is invisible.

libs/core/src/listings/domain/ports/price-change-auto-applied-log-repository.port.ts
```diff
@@ -0,0 +1,17 @@
...
+import type { RecordAutoAppliedPriceChangeInput } from '../types/price-change-auto-applied-log.types';
+
+export interface PriceChangeAutoAppliedLogRepositoryPort {
+  record(input: RecordAutoAppliedPriceChangeInput): Promise<PriceChangeAutoAppliedLogEntry>;
+
+  /**
+   * The most recent N entries, newest first. Deliberately NOT paginated or
+   * filterable (ADR-072 decision 3 — this is explicitly not a digest feature).
+   */
+  findRecent(limit: number): Promise<readonly PriceChangeAutoAppliedLogEntry[]>;
```
norbert-kulus-blockydevs: **IMPORTANT — no time bound and no retention, so #3168's banner is permanent.**

`findRecent` is a plain `ORDER BY "appliedAt" DESC LIMIT n` with **no `where` clause at all**, the migration adds no retention sweep, and nothing prunes the table. #3168's `AutoAppliedNote` renders `if (items.length === 0) return null` — so once **any** automatic price change has ever occurred, "N price changes went live automatically recently" is displayed for the life of the install, even after the operator switches every connection back to Manual.

(#3168's author noticed the symptom and softened the mockup's "today" to "recently"; the cause is here. Its dialog title still reads *"Applied automatically today"*.)

**Fix:** give `findRecent` a `since: Date` (or a filters object) so the banner can bound itself to a real window — the `appliedAt` index already serves it — and document the table's retention posture, which is currently stated nowhere.

norbert-kulus-blockydevs:
Addressed all 7 review findings. Summary of changes (commit cd6a85727), pushed after rebasing this stacked branch onto the updated `origin/3143-price-change-detection` (which had grown its own #3159 review-fix commit in the meantime — the duplicate, now-superseded copy of that commit on this branch was dropped via `git rebase --skip` after confirming the parent's version was a byte-for-byte superset).

**1. BLOCKING — capability discovery by thrown exception.** Fixed. The marketplace-vs-shop branch is now decided by `resolveAdapterMetadata` (metadata-only, works on a disabled connection) crossed with `connection.enabledCapabilities`, mirroring exactly what `getCapabilityAdapter` checks internally — never a caught exception. Additionally requires `isOfferFieldUpdater` before electing the marketplace branch, so a WooCommerce-shaped `OfferManager` adapter that implements only `updateOfferQuantity` now returns `business_failure` instead of retrying 10x into a dead job.

**2. BLOCKING — shop branch always creates a new product.** Fixed. The shop branch now routes through `IProductPublishExecutionService.executePublish`, which resolves the existing `ShopProduct` mapping, handles `variantGroup.externalParentProductId`, persists the mapping, and already converts a terminal shop rejection into `business_failure` rather than a retried throw — for free. Also fixed the "same call site" sub-finding: the service now reads the previous publish's `ListingCreationRecord.request` snapshot and reuses `destinationCategoryIds`/`content`/`commerce`/`parameters` verbatim, so a price-only apply no longer re-provisions categories, re-projects attributes, or overwrites channel content with the master description. `ShopProductManagerPort`'s lack of a partial-field-update primitive is stated as a real constraint in the code, not invented around.

**3. BLOCKING — publishes `totalAvailable`, defaults unknown ATP to 0.** Fixed. Now reads `availableToPromise` and throws `AvailabilityUnknownError` (the exact, already-retryable exception both shipped publishing consumers use) when it's `null`, rather than defaulting to a number. On the `observedAt`/ADR-067 write-ordering-guard concern: I traced it and confirmed by grep that `ShopProductManagerPort.publishProduct` has **no existing integration** with that cursor anywhere in the tree — not in this handler, not in the ordinary `shop.product.publish` handler, not in bulk shop-publish. It's a pre-existing gap in the `publishProduct` capability shared by every caller, not something this PR's stock-read introduced on its own. Rather than invent an unproven partial integration under this PR's scope, I've documented it explicitly as a deliberate follow-up in the service's docblock.

**4. BLOCKING — every log row records `oldAmount === newAmount`.** Fixed, and it turned out to be entangled with #3159's fix in a way worth flagging: since #3159 makes the detection service *always* open/refresh an episode before deciding whether to auto-apply, `payload.episodeId` is now populated on every current apply — which means the pre-existing `if (episodeId) resolve() else log()` branching had silently made the auto-applied-log write **unreachable in practice** (every automatic apply would just resolve its episode and never log, once rebased onto #3159). Fixed by decoupling the two: episode resolution and the auto-applied-log write are now independent, discriminated on `automatic` (never on `episodeId` presence) — automatic apply resolves AND logs (using the episode's real `computedOldAmount`); a review-queue accept resolves only. `oldAmount` is now nullable end-to-end (entity/ORM column — widened in the same not-yet-merged migration — repository, log-write input), so a first-ever detection with no baseline is honestly recorded rather than fabricated.

**5. BLOCKING — cross-context `*RepositoryPort` imports.** Fixed via the suggested extraction. New `IPriceChangeApplyService` (`libs/core/src/listings/application/services/price-change-apply.service.ts` + interface + domain types + a 22-case unit spec covering every fix above) owns the full orchestration; the worker handler is now a thin shell (payload parse → delegate → map/wrap result), wired into `ListingsModule` the same way `IProductPublishExecutionService`/`IBulkListingProgressService` already are.

**6. IMPORTANT — frozen fields ignored; permanent conditions retried; discarded resolve() boolean.** Fixed. `UpdateOfferFieldsReport.frozenFields` is now checked — a seller-frozen price is `business_failure`, not success. All five structurally-permanent conditions (episode not found, blockReason, no offer mapping, `OfferFieldUpdater` unsupported, neither capability enabled) resolve as `business_failure` through one `terminal()` helper that also always advances bulk progress `'failed'` — so a dead-lettered child can no longer leave a batch stuck `running`. A genuine adapter rejection still propagates with `.cause` preserved (verified against `AllegroRetryClassifierAdapter`, which classifies by exception type/status code, not job type, so it dead-letters a deterministic 4xx on attempt 1 with no changes needed here). The discarded `episodes.resolve()` boolean is now checked and logged. I deliberately did **not** add a full claim-marker column (the `waybillRelayedAt`/`reservationConsumedAt`/`dispatchRelayedAt` shape) — unlike those precedents, a crash between "publish succeeded" and "resolve/log" here self-heals via an idempotent re-publish of the identical price on retry, so the worst case is a duplicate audit-log row, never a duplicate shipment or charge. Documented as a conscious scope decision in the service's docblock.

**7. IMPORTANT — `findRecent` has no time bound; no retention.** Fixed. `findRecent` now takes an optional `since: Date`. The port docblock now states plainly that no retention sweep exists for this table, naming `DemoAccountCleanupService` as the precedent a follow-up should mirror.

**Verification in my environment:** `pnpm --filter @openlinker/core` and `--filter @openlinker/worker` `type-check` both pass cleanly; targeted `eslint` on every touched/added file is clean; the full `pnpm run check:invariants` (which `pnpm lint` chains into) passes in full, including `check-cross-context-imports` (0 new violations) and `check-service-interfaces`; the new `price-change-apply.service.spec.ts` (22 cases) and rewritten `price-change-apply.handler.spec.ts` (6 cases) both pass under a targeted `jest` run. A repo-wide `pnpm -r lint` across all 22 packages did not complete within this sandbox's resource limits before timing out on unrelated packages — a known constraint of this environment, not a signal about this diff.

piotrswierzy (commented):

A well-argued slice that visibly absorbed a prior review round: the orchestration correctly lives in a core application service rather than the handler, the **lane choice (`bulk`) is right** and justified against ADR-050's cost-of-starvation rule and #2594's precedent, and the bulk-progress reuse hits the genuine at-most-once `bulk_batch_advancements` gate rather than inventing a counter.

Three things are right in non-obvious ways: the capability branch is resolved from `resolveAdapterMetadata` × `enabledCapabilities` rather than by catching `CapabilityNotEnabledException` (the superclass/subclass trap is real); `availableToPromise` with `null` → `AvailabilityUnknownError` rather than `0` correctly avoids the #1689 pause primitive; and the pricing rule is deliberately **not** re-applied here because #3143's detection already did it (explicit price always wins, #1843).

The concerns are all at the write boundary.

### IMPORTANT — an unguarded absolute stock write on a price-only path

`publishToShop` calls `productPublishExecution.executePublish({ stock, status: 'published', … })` — an **absolute stock write** on a *price-only* path, and it does **not** go through `InventorySyncService.updateOfferQuantities`, which is where ADR-067 / #2617's per-`(connection, offer)` lock and `inventory.offerQuantity.observedAt` freshness mark live. So a price apply races a concurrent inventory write-back and can land an older quantity last — precisely the oversell ADR-067 exists to prevent. The read (`getAvailabilityByVariantIds`) happens milliseconds before the write with nothing carrying an `observedAt` token.

Either thread the availability row's `updatedAt` through as `observedAt` and take the same mark/lock, or narrow the shop write so it does not carry stock at all. The docblock already concedes `ShopProductManagerPort` has no partial-update primitive — **that is the port change worth filing**, rather than shipping an unguarded third stock writer.

### IMPORTANT — `toFixed(2)` hardcodes two minor units

`publishToMarketplace`. The repo already paid for this: `minorUnitExponentFor` (#2260 review, `sales-documents`) exists precisely because two decimals is false for JPY and KWD, and `splitShippingAcrossRates` was changed to read it rather than assume. A JPY price published as `"399.00"` is at best rejected and at worst mis-rounded.

### IMPORTANT — a partial write is reported as a total failure

Offers are iterated in sequence; if offer A's price is written and offer B reports `price` frozen, the method throws `PriceChangeApplyPermanentError` → `business_failure`, so the episode is left unresolved and no auto-applied row is written — **even though A's price really changed on the channel**. That is the inverse of #2593's rule (*"a failed product does not fail the page"*). Report per-offer outcomes, or resolve the episode and record the frozen offers as a distinct partial state.

### IMPORTANT — a `*RepositoryPort` on the main `listings` barrel

`PriceChangeAutoAppliedLogRepositoryPort` is exported from `libs/core/src/listings/index.ts`. § Cross-context dependencies lists `*RepositoryPort` as explicitly **forbidden** on that surface. The gate script denies the *import*, not the export, so this passes lint while publishing exactly the seam this PR's own headline change (moving the repo ports out of the worker handler) was made to remove. #3159 set the same precedent with `PriceChangeEpisodeRepositoryPort` — consistent, still wrong. Drop both.

### IMPORTANT — `price_change_auto_applied_log` is not in `tablesToTruncate`

Per `docs/testing-guide.md` the probe walks only the listed tables' CASCADE closure, and this table has **no FKs at all** by design — so nothing will ever clear it, and rows leak across every integration test in the run. One line. (`price_change_episodes`, from #3159, has the same omission — fix once across the stack.)

### SUGGESTIONS

- A transient throw re-propagates **without** `advanceBulkProgress(input, 'failed')`, so a job that exhausts its retry ladder and dead-letters never advances the counter and the batch sits at `running` forever — `succeededCount + failedCount === totalCount` can never be satisfied. `marketplace.offer.create` gets away with the same shape because `executeCreation` converts terminal conditions to a `business_failure` outcome rather than throwing; here the transient path is genuinely reachable and terminal.
- The adapter `idempotencyKey` embeds `new Date().toISOString()` computed inside `publishPrice`, so every retry mints a fresh key and adapter-side dedupe never fires. It mirrors the shipped `content:{…}:{publishTimestamp}` precedent and an absolute price write is near-idempotent, so harm is low — but the value is derived from a clock rather than from the episode, which is the shape #2039 was opened about. `pricing:episode:{episodeId}` would be stable.
- `OfferManagerPort` / `isOfferFieldUpdater` are imported from `@openlinker/core/listings` — the service's **own** context barrel — while its siblings use relative paths. `isOfferFieldUpdater` is a *value* import, and #337/#359 is the recorded reason `ListingsModule` lives on the `/services` subpath: a same-context value import of the main barrel is the circular-require shape that split caused.

Migration timestamp (`1879…` > the stack's `1878…`) and entity/DDL column parity both check out.

---

norbert-kulus-blockydevs:
Thanks for the re-review — addressed all five IMPORTANT findings and two of the three suggestions in 59d8472b4 (rebased onto the parent's own just-landed `de951c9e8` fix).

**Unguarded absolute stock write on the price-only path** — `publishToShop` now resolves the `ShopProduct` external id and, when one exists, takes the SAME per-(connection, offer) lock + `inventory.offerQuantity.observedAt` cursor that `InventorySyncService`'s shop write-back branch already keys by that id (`@openlinker/core/inventory`'s published offer-quantity-write-order primitives — no new mechanism invented). A contended lock now throws the shared `ContendedWriteError` rather than proceeding unguarded. This correctly serialises the concrete race you named (a concurrent `inventory.propagateToMarketplaces` write-back for the same target); it's honestly scoped as NOT covering the other `executePublish` callers (`shop.product.publish`, bulk shop-publish), which touch no such cursor either and remain a documented, separate port-level gap (`ExecutePublishProductInput.stock` has no optional/partial-update form).

**`toFixed(2)`** — now reads `minorUnitExponentFor(currency)` from `@openlinker/core/sales-documents`, the same table `splitShippingAcrossRates` (ADR-063 §5) uses.

**Partial write reported as total failure** — offers are still iterated per-mapping, but success/frozen are now tracked separately; `business_failure` only when EVERY mapped offer refuses. A partial write resolves normally and logs a warning (no per-offer slot exists on the episode/log to persist the frozen subset — noted as the honest limit of this pass).

**`*RepositoryPort` on the main `listings` barrel** — `PriceChangeAutoAppliedLogRepositoryPort` had zero external consumers (confirmed by grep) and is now removed from the barrel entirely. `PriceChangeEpisodeRepositoryPort` stays: it turns out it already has a legitimate, ALLOW-LISTED int-spec consumer (`listings-price-change-episode.int-spec.ts`, driving the conflict-arm write + reopen-vs-rival race against real Postgres) — the exact same sanctioned exception `OfferMappingRepositoryPort`/`OfferStatusSnapshotRepositoryPort` already carry on this same barrel, rewire tracked at #722. Added a comment distinguishing the two so the next reader doesn't have to rediscover it.

**Missing `tablesToTruncate` entries** — added both `price_change_episodes` and `price_change_auto_applied_log` (neither carries an FK the CASCADE-closure walk can reach), following the `reservation_shortfall_episodes` precedent already documented in that file.

**Suggestions**: fixed the idempotency-key clock issue (now derives from `episodeId`, namespaced `pricing:apply:...` distinct from the detection service's `pricing:episode:{id}:auto` job-dedup key) and the same-context barrel import (now relative, matching the #337/#359 precedent). Left the bulk-progress-on-dead-letter gap deliberately undone — I traced it against `marketplace.offer.create`'s `executeCreation` as suggested and found the IDENTICAL gap already there (its outer catch also never calls `advanceBatchStatus`), so it's a systemic `SyncJobRunner`-level issue across every bulk-child handler, not something this PR should half-fix in isolation.

Verified: `@openlinker/core` and `@openlinker/worker` type-check + lint clean (0 errors), plus `@openlinker/api` lint clean; full `pnpm run check:invariants` passes, including `check-cross-context-imports` with 0 new allow-list entries needed. Added test coverage for all four behavioral fixes (partial-offer success, currency rounding, idempotency-key stability, and the new lock/cursor guard including the contention and no-prior-mapping paths).

piotrswierzy (commented):
## `/pr-review` (round 2) — 🔄 Approve with changes

Four of the five previous findings are fixed, and the fifth is *engaged with seriously* rather than papered over — which is why it now needs a decision recorded rather than a code change.

Fixed and verified: **`minorUnitExponentFor` replaces the hardcoded `toFixed(2)`** (`price-change-apply.service.ts:61,337`), so a JPY or KWD price no longer rounds to a currency's wrong precision. **The barrel no longer exports `PriceChangeAutoAppliedLogRepositoryPort`** (`listings/index.ts:403` records why), keeping the repository port an intra-context contract. **A partial marketplace write is no longer reported as a total failure** (:342-404) — zero successes fails, a mixed result warns and names the offers that did apply, citing the #2593 "a failed product does not fail the page" rule one grain down, which is the right precedent.

### IMPORTANT — the shop path now knowingly publishes a stock quantity it has just determined is stale, which is an ADR-067 reversal and needs an amendment, not a warn

This is the honest version of the previous finding, and the code is admirably explicit about it (:433-443, :502-512), so I want to be precise about what I am asking for.

What the path now does: takes a per-offer lock, re-reads stock inside it, consults the `inventory.offerQuantity.observedAt` freshness mark, and when `isWritableQuantityObservation(observedAt, lastWritten)` is **false** — i.e. a newer quantity has already been written to that channel — it logs `publishing anyway` and publishes. The stated reason is sound as far as it goes: `ShopProductManagerPort` has no partial-field primitive, `ExecutePublishProductInput.stock` is required, so refusing the write would drop the *price* change too.

But the consequence is exactly the failure [ADR-067](../blob/main/docs/architecture/adrs/067-freshness-token-write-ordering.md) exists to prevent: an older quantity lands last and the listing advertises stale stock until the next change at the master — overselling when the older number is the higher one. The ADR's own framing is that the guard "must not be a way to leave a listing permanently stale", and the mark advances only after a *successful* write precisely so a refusal always means a newer quantity is live. Publishing past a false predicate inverts that guarantee, and afterwards the mark correctly is not moved (:519-528) — so the channel now holds a quantity that no mark accounts for.

Three ways out, in my order of preference:

1. **Refuse and re-enqueue.** The job is retryable and the price change is not lost, only deferred — this is what `ContendedWriteError` and the penalty-free deferral budget (#2613) exist for, and a price change deferred by one lane tick is a far smaller harm than an oversell.
2. **Amend ADR-067** with a shop-path carve-out stating this exact trade-off, the way § Amendment (#2621) narrowed "a successful write" to "a successful submission" for Allegro. The repository's standing rule is that a reversal starts with an amendment; a `logger.warn` is not where a contract gets narrowed.
3. Give `ShopProductManagerPort` the partial-field primitive, which is the real fix and plainly out of scope here.

Either (1) or (2) is fine by me for this PR. What is not fine is the invariant being narrowed only in a comment, because the next reader of ADR-067 will believe the guarantee still holds.

### IMPORTANT — `price_change_auto_applied_log` is still absent from the shared truncate list

`listings-price-change-episode.int-spec.ts:37` truncates `price_change_episodes` inline, but neither table is in `apps/api/test/integration/setup.ts`'s `tablesToTruncate`. Per-spec truncation covers the spec that remembers; the shared list is what stops a row written by *one* suite leaking into an unrelated one — which is precisely the cross-suite contamination #3126 is currently fixing one layer up. Add both tables there and drop the inline statement.

### SUGGESTION

`1879000000000` is shared with #3035's `create-order-cancellation-signals` (which moved off `1878`, where this epic's first migration and #3014's and #3201's now all sit). Nothing to change here; see the table in my #3158 review.

---

norbert-kulus-blockydevs:
Addressed the two IMPORTANT findings from the second re-review, and confirmed the SUGGESTION needs no action here.

**ADR-067 reversal — fixed, not amended.** Chose option 1 from the review's own preference ordering. `publishToShopGuarded`'s freshness check previously logged a warning and published the stale stock quantity anyway when `isWritableQuantityObservation` came back false — exactly the inversion the review names: the guard's whole point is that a refusal always means a newer quantity is already live, and publishing past a false predicate breaks that. It now throws the same `ContendedWriteError` the peer-lock branch immediately above it already raises, so the runner defers the job penalty-free (#2613/#2617) rather than committing the write. `executeShopPublish` is now unreachable on that branch — the price change is delayed by one lane tick, never lost, and no stale quantity reaches the channel. Went this way over an amendment because nothing about the constraint (`ShopProductManagerPort` having no partial-field primitive) actually requires relaxing ADR-067 — deferring is sufficient.

**Truncate list.** Turns out `price_change_episodes` / `price_change_auto_applied_log` were already added to `apps/api/test/integration/setup.ts`'s `tablesToTruncate` by the prior fix commit (59d8472b4) — landed before this review round, evidently not reflected in it by the time it posted. What genuinely was still outstanding, per the review's own text: `listings-price-change-episode.int-spec.ts` still had its own inline `TRUNCATE TABLE "price_change_episodes"` in `afterEach`, duplicating (and able to drift from) the shared list every other spec in the suite relies on. Replaced it with `resetTestHarness()`.

**SUGGESTION (migration number 1879000000000 vs #3035).** The review itself says "nothing to change here; see the table in my #3158 review" — left as-is.

Verification (jest not run locally, resource-constrained machine — CI runs the full suite): `pnpm --filter @openlinker/core` and `--filter @openlinker/api` lint + type-check both pass cleanly, and the full `pnpm check:invariants` passes with zero new violations.

norbert-kulus-blockydevs:
Found and fixed a real CI-caught bug (not from review): `handler-registration.service.spec.ts` hardcodes a per-lane job-type count, and this PR's `pricing.propagateToMarketplaces` handler registration (bulk lane, ADR-072 decision 8/9) pushed the real `bulk` count from 29 to 30 without the spec being updated.

Recounted directly from `handler-registration.service.ts` once both this branch and #3143 are merged together: realtime=16, bulk=30, fiscal=5, fan-out=7 (total 58 registrations). Updated the `toHaveLength(29)` assertion to 30 and extended the narrative comment to name `pricing.propagateToMarketplaces` as the 30th bulk-lane member, matching the style of the existing entries.

Note: on #3143 alone (before this PR's handler registration lands) the same spec shows a larger mismatch — that's the documented, accepted merge-order dependency in `sync-job.types.ts` (the job type exists in `JobTypeValues` before its handler is registered). This fix is verified against `3145-price-changes-api`, which has both #3143 and #3144 merged in, plus a full `pnpm check:invariants` pass. Pushed as commit `9e1d3593b`.

piotrswierzy (commented):
## Tech Lead review (round 3) — 🔄 Approve with changes

Both IMPORTANT findings from round 2 are fixed, and the ADR-067 one is fixed the way I asked rather than argued around. `publishToShopGuarded` now takes the same `offerQuantityWriteLockKey(connectionId, externalOfferId)` lock and the same `inventory.offerQuantity.observedAt` cursor that `InventorySyncService`'s shop write-back branch keys by that id, re-reads stock *inside* the lock, and on a false `isWritableQuantityObservation` **throws `ContendedWriteError` instead of publishing anyway** — option 1 from my own preference ordering, so no ADR amendment is needed and the invariant is not narrowed in a comment. I checked the obvious failure mode of that choice and it is not reachable: `isWritableQuantityObservation` is `candidateMs >= lastMs`, so an equal observation passes and a refusal requires this path to have read a genuinely older `inventory_items` row than the mark; the retry re-reads and clears. No livelock, and the penalty-free deferral budget is the right home for the wait.

Also verified fixed: `minorUnitExponentFor(currency)` replaces `toFixed(2)`; `PriceChangeAutoAppliedLogRepositoryPort` is off the main barrel with the reasoning recorded in place; a partial marketplace write no longer fails the whole apply; `findRecent` takes a `since` and the port states plainly that no retention exists; `oldAmount` is `episode?.computedOldAmount ?? null`, never fabricated equal to `newAmount`; `advanceBulkProgress` runs on every terminal exit through one `terminal()` helper; the idempotency key derives from `episodeId` rather than a clock; and `price_change_episodes` + `price_change_auto_applied_log` are both in `tablesToTruncate` with the no-FK reasoning spelled out. The auto-applied-log migration ↔ entity parity is clean (named PK, named index, `varchar(8)` not an enum, nullable `oldAmount` on both sides). The `bulk` lane choice is justified against ADR-050 decision 1 correctly, and the enqueue carries a real `destinationConnectionId` (#2609). CI is green on all nine checks.

### Findings

**[IMPORTANT]** — `libs/core/src/listings/application/services/price-change-apply.service.ts:461-468`

> The `externalOfferId === null` branch publishes unguarded, and its comment justifies that by the *lock*: *"nothing has written this target via the guarded path either, so there is nothing to race against yet."* That reasoning is about contention, but the real consequence of a missing `ShopProduct` mapping is different and worse: `executeShopPublish` → `executePublish` → `publishProduct` with `externalProductId` unresolved, and `woocommerce-product-publisher.adapter.ts:149` branches `isUpsert = cmd.externalProductId != null` → **`POST`, i.e. it creates a new product on the live storefront** — the narrowed return of round 1's BLOCKING #2.
>
> It is not reachable from #3159 today (detection walks existing `ShopProduct` mappings, so a shop destination always has one by construction). It becomes reachable the moment #3145's manual accept endpoint enqueues for an episode whose mapping was deleted in between — and this service is the one that has to refuse, because `executePublish` cannot tell a price-change re-publish from a first publish.
>
> Make the null branch **terminal**: `throw new PriceChangeApplyPermanentError(...)` naming the missing mapping, exactly like the four other structural conditions already routed through `terminal()`. A price change against an unmapped shop product is a deterministic business refusal (ADR-007), not a create.

**[IMPORTANT]** — `apps/api/src/migrations/1879000000000-create-price-change-auto-applied-log.ts`

> `1879000000000` is claimed by three distinct migrations across currently-open PRs — this one, #3035's `create-order-cancellation-signals`, and #3219's `add-order-record-sales-document-matched-rule`. `check-migration-timestamps.mjs` diffs only against `origin/main`, so all three pass in isolation and the collision reddens `main` for whoever merges second. I flagged this last round as *"nothing to change here"*; I am upgrading it, because **#3158 has since demonstrated the remedy inside this very stack** — it bumped clear of contention to `1880000000000`, which is now uniquely its own. `1881000000000` is free; the same one-line bump (file + exported class + `name` property) removes this PR from the contention too. That is cheaper than the coordination it otherwise costs the last merger.

### SUGGESTIONS

- The bulk-progress-on-dead-letter gap is left undone on the grounds that `marketplace.offer.create` has the identical gap. That is a fair read and I accept the decision — but the two are not quite equivalent: `executeCreation` converts its terminal conditions to a `business_failure` outcome rather than throwing, so its transient path is largely unreachable, whereas here `ContendedWriteError` and `AvailabilityUnknownError` both genuinely propagate. A batch that dead-letters a child sits at `running` forever, since `succeededCount + failedCount === totalCount` can never be satisfied. Worth a line on `advanceBulkProgress`'s docblock naming the condition, so #3145's bulk-accept author knows what they are inheriting.
- `resolveShopStock` throws `AvailabilityUnknownError` *inside* `publishToShopGuarded`'s lock, which the `finally` correctly releases — but it is thrown after the lock is taken, so an install with persistently-unknown ATP holds and releases the lock once per retry. Harmless; noting it only because the ordering reads as if the stock read were cheap and infallible.
- `advanceCursorIfNewer` returning `false` warns and continues, which is right (a newer mark is already live). The warn text says the mark *"was not moved"* — consider naming the offer id's connection scope too, since the cursor key namespace is per connection and the message is otherwise ambiguous across connections sharing an external id.

### Verdict

🔄 **Approve with changes** (posted as a comment rather than an approval because of the two IMPORTANT items)

1. Make the no-`ShopProduct`-mapping branch terminal rather than a create.
2. Bump the migration to `1881000000000` to clear the three-way contention, following #3158's precedent in this same stack.

Neither blocks the design; both are cheap and both are the kind of thing that is far more expensive to fix after merge. Note also that this branch is what proves #3159's lane-coverage failure is a merge-ordering artefact rather than a defect — the two must land together.

---

norbert-kulus-blockydevs:
Both round-3 findings are addressed:

1. **Missing `ShopProduct` mapping is now terminal, not a create.** `publishToShop`'s `externalOfferId === null` branch previously published unguarded, which would have made `publishProduct` take its `isUpsert = false` branch and create a brand-new WooCommerce product as a side effect of a price update. It now throws `PriceChangeApplyPermanentError` (routed through the existing `terminal()` path, ADR-007), matching every other structural precondition already handled that way.

2. **Migration renumbered `1879000000000` → `1881000000000`** to clear the three-way collision with `create-order-cancellation-signals` (#3035) and `add-order-record-sales-document-matched-rule` (#3219), following #3158's precedent in this same stack. File, exported class name, and `name` property were all updated together; grepped the repo for stale references (none found).

Also merged in the parent branch's latest commit (`fix(listings): log inactive destination once per connection, not per variant`), which confirms the lane-coverage documentation added there resolves cleanly once both PRs land together — `apps/worker` type-checks clean with both changes in place.

Verified: `@openlinker/core`, `@openlinker/worker`, and `@openlinker/api` all type-check cleanly, and `pnpm check:invariants` passes in full (including `check-migration-timestamps`).

norbert-kulus-blockydevs:
Round-3 fixes landed on `3144-price-change-apply-job` (merged `3143-price-change-detection` first, confirmed lane-coverage now resolves with both branches together — `pricing.propagateToMarketplaces` has both a declared job type and a registered `bulk`-lane handler, and `handler-registration.service.spec.ts` already asserts `bulk` = 30 including it).

**No-`ShopProduct`-mapping branch is now terminal.** `publishToShop`'s null-`externalOfferId` branch no longer falls through to an unguarded `executeShopPublish`. A missing mapping now throws `PriceChangeApplyPermanentError`, routed through the same `terminal()` path every other structural precondition in this service uses, instead of letting `executePublish` → `publishProduct` take its `isUpsert = false` branch and create a duplicate product. Also updated the unit tests: the 'shop publish' describe block now resolves a `ShopProduct` mapping by default (so its pre-existing tests exercise the ordinary mapped path instead of accidentally hitting the new terminal branch), and the former "publishes unguarded" test now asserts a terminal `business_failure` with no lock taken and no publish attempted.

**Migration renumbered to `1881000000000`.** `1879000000000` collided with three other open-PR migrations sharing `1878000000000` (#2202, #2703/#2704, #3176). Confirmed `1881000000000` is free against `origin/main` and all three colliding branches (following #3158's precedent of bumping to `1880000000000` for the same reason). Renamed the migration file, its exported class, and the `name` property; grepped for and found no other references to the old class name.

Verification: `pnpm --filter @openlinker/core type-check`, `pnpm --filter @openlinker/worker type-check`, `pnpm --filter @openlinker/api type-check`, and `pnpm check:invariants` (including `check-migration-timestamps`) all pass clean.

norbert-kulus-blockydevs:
Pushed `9b80e6f44` — the last red unit test on this branch is fixed.

**What was still failing.** `676356312` made a missing `ShopProduct` mapping a terminal refusal, and the follow-up `90244a983` seeded a resolvable mapping in the `describe('shop publish')` `beforeEach`. One case that reaches the shop publish lives *outside* that block — `capability branch selection › resolves to the shop branch only when ProductPublisher is actually enabled, not merely because OfferManager is absent` (`:200`) — so it kept inheriting the suite-wide default (`getExternalIds → []`) and came back `business_failure` where it asserts `{ outcome: 'ok' }` (hence the `- Expected 1 / + Received 2` diff on `:229`).

**Fix.** Seed the same mapping in that one case. Deliberately *not* by widening the suite-wide default: that default is what keeps the dedicated refusal case at `:538` honest, and flipping it would hide the very branch it exists to assert. The comment at the seed says which question the case is asking, so the next reader does not "tidy" it back.

**Verified locally** (scoped, not a full run):

- `pnpm --filter @openlinker/core test -- price-change-apply` → 31/31 pass
- `pnpm --filter @openlinker/core type-check` → clean
- `pnpm check:invariants` → clean

Integration Tests were already green on this branch; this should take `Test` green too.

Note for whoever merges: this PR and #3159 have to land together or back-to-back. `3143` declares `pricing.propagateToMarketplaces` in `JobTypeValues` while the handler is registered here, so `assertFullLaneCoverage()` fails at worker boot on #3159 alone — that is the documented MERGE-ORDER NOTE in `sync-job.types.ts`, not a regression.

PR #3162, opened 2026-09-10
feat(api): price-changes queue endpoints (list, accept, ignore, edit)

norbert-kulus-blockydevs:
## Summary

Part of #3139 (epic). **Stack position: sixth of ten** — branches from `3144-price-change-apply-job` (#3161, open — base moves to `main` once that merges).

New `PriceChangesController` at `GET/POST /listings/price-changes`:

- `GET /listings/price-changes` — the review queue, enriched with product/variant/connection identity, structured `ruleSummary`, `deltaPct`/`isSteep`, `blockReason`, `needsRefresh`, and a `version` token (`refreshedAt ?? detectedAt`) the accept/edit staleness guard checks against. Excludes episodes on stale (#1689) variants, reporting them via `hiddenStaleCount` instead (the mockup's `#stale-note`).
- `GET /listings/price-changes/auto-applied` — the #3144 log, deliberately unpaginated (ADR-072 decision 3).
- `POST /:id/accept`, `POST /:id/edit`, `POST /:id/ignore`, `POST /:id/unresolve`, `POST /bulk` — all route through `IPriceChangesService`'s shared `loadActionable()` guard: 404 if the episode doesn't exist, 409 if already resolved, 422 if blocked, 409 if the caller's `expectedVersion` no longer matches (the staleness guard AC).
- `optInAutomatic` on accept/edit/bulk flips `Connection.config.priceSyncMode.sourceOverrides[sourceConnectionId]` to `'automatic'` via read-merge-write — best-effort, since the price publish already happened/enqueued and a mode-flip failure must never be reported as a failed accept.
- `bulkAccept` reuses `BulkListingBatchRepositoryPort.create` verbatim (ADR-072 decision 9 — the same `bulk_offer_creation_batches`/`bulk_batch_advancements` mechanism #3144's handler already advances), returning a `batchId` pollable via the existing bulk-batch-progress endpoint.

### Gating (ADR-072 decision 6)

Reads: `@Roles('admin','operator','viewer')`. Writes: `@Roles('admin','operator')` — matching every other day-to-day Listings write (`listings.controller.ts`), **not** `ConnectionController`'s `admin`-only connection-config gating, which #3146 (connection pricing-sync settings) will use instead since that PR touches `Connection.config` directly as its primary act.

### Gap flagged back, per the issue's own instruction

[...]

PR #3163, opened 2026-09-10
feat(api): connection pricing & sync settings endpoints

### LINKED ISSUE #3146: feat(api): connection pricing & sync settings endpoints
## Problem / Context

Part of #3139 (epic). **Stack position: seventh — branches from #6's branch.**

Frontend sub-issues #9/#10 need endpoints to read/write a destination connection's default + per-source pricing rules and sync modes, and a read-only rollup for a source connection.

## Proposed Solution

Extend the existing connection settings endpoints (`apps/api/src/integrations/` connection controller — reuse the pattern `StockAndPricingSection`'s backend already follows, #2610) rather than inventing a parallel connection-settings surface:

- `GET /connections/:id/pricing-sync` — returns the destination connection's current `{ default: { mode, rule }, sourceOverrides: Record<sourceConnectionId, { mode, rule }> }` plus, for each entry in `sourceOverrides`, the source connection's display label (mockup `SOURCES[srcKey].label`, e.g. "PrestaShop — Main Store") and its open-episode count (mockup `#conn-pending-text`, "N changes waiting").
- `PATCH /connections/:id/pricing-sync` — body: the same shape as the GET response's payload (default + sourceOverrides). Save-only semantics matching `StockAndPricingSection`'s existing explicit-Save pattern (no instant-apply-on-toggle) — mirrors the mockup's `#conn-unsaved-bar` / `#conn-save` / `#conn-discard` interaction exactly. Validate: `rule.type` in `passthrough|markup|margin`, `rule.pct` required and >0 unless `passthrough`, `rule.rounding` in `none|nearestWhole|endingIn99` — reuse whatever validation `readPricingRule` already does server-side rather than duplicating it in a DTO.
- `GET /connections/:id/pricing-sync/as-source` — the READ-ONLY rollup for a connection acting as a SOURCE (mockup `#page-source`, `#source-page-list`): for every destination connection that has this connection as a known source (i.e., appears in that destination's `sourceOverrides` keys OR has ever produced a detected price change for it, per #3/#4's episode history), return `{ destinationConnectionId, destinationLabel, effectiveMode, effectiveRuleSummary, isCustomOverride }`. This is intentionally READ-ONLY — no PATCH variant — per ADR decision 2 (the source page never edits, only links to the destination's settings).

## Classification

**Type**: Interface (API)
**Layer**: Interface
**File(s)**: extend the existing connection controller in `apps/api/src/integrations/`, new DTOs colocated with the existing `StockAndPricingSection`-backing DTOs if they exist, or new ones following the same conventions

## Dependencies

- #6 (price-changes queue endpoints) — not a hard technical dependency, but stacked for branch-history linearity per the epic's sequencing

## Assumptions

- "Known source" for the `as-source` rollup is derived from `sourceOverrides` keys across all destination connections' configs, UNIONED with any source connection that has ever appeared in an episode's `sourceConnectionId` for this connection — confirm at implementation time whether the config-derived set alone is sufficient (it may be, if every source that's ever produced a price change necessarily gets a `sourceOverrides` entry once it has one; if a source with only the DEFAULT rule in effect should still show up, the episode-history union is required — the mockup's `PRICING_CONFIG[destKey].sources` always includes every known source key even with `override: null`, so mirror that: every source that has EVER fed this destination appears, override or not).

## Acceptance Criteria

- [ ] `GET`/`PATCH /connections/:id/pricing-sync` round-trip correctly, `PATCH` requires an explicit call (no side-effect writes from unrelated connection edits)
- [ ] `GET /connections/:id/pricing-sync/as-source` returns every destination this connection feeds, with no way to write through this endpoint (verify no `PATCH`/`POST` variant exists)
- [ ] Rule validation matches `applyPricingRule`'s actual accepted shapes — a malformed rule is rejected with a 400 naming the specific field, not a generic error
- [ ] Tests added for non-trivial logic
- [ ] No CORE ↔ Integration boundary violations

norbert-kulus-blockydevs:
## Summary

Part of #3139 (epic). **Stack position: seventh of ten** — branches from `3145-price-changes-api` (#3162, open — base moves to `main` once that merges).

- `GET`/`PATCH /connections/:id/pricing-sync` — a destination's default + per-source pricing rule and sync mode. `PATCH` is `@Roles('admin')` — this is genuine `Connection.config` mutation, so it matches `ConnectionController`'s own write gating rather than #3145's operator-level gating. Explicit-Save: the body replaces the whole `default` + `sourceOverrides` shape in one call (the `StockAndPricingSection` no-partial-patch precedent).
- `GET /connections/:id/pricing-sync/as-source` — the **read-only** rollup for a connection acting as a source (ADR-072 decision 2). No `PATCH` variant exists on this route at all, by design — verifiable by grep.
- "Known sources" (both routes) = every source id declared in `config.pricingRule.sourceOverrides` / `config.priceSyncMode.sourceOverrides`, unioned with every source that currently has an **open** price-change episode against the connection — the #3146 assumption note's resolution, so a source that has fed a destination but never been given its own rule still shows up.
- Rule/mode validation reuses the same runtime arrays `readPricingRule`/`readPriceSyncModeConfig` already validate against, so the DTO can't accept a shape those pure helpers would silently coerce away. `sourceOverrides` (a `Record<string, T>`, which class-validator has no first-class nested decorator for) is validated per-entry in the controller, naming the offending source id in the 400.

## Test plan

- [x] `pnpm --filter @openlinker/core|api type-check` — clean on every file this PR touches (remaining errors match the previously-confirmed pre-existing baseline).
- [x] Unit tests (`ConnectionPricingSyncService`): default-only unconfigured connection, a configured override listed even with no open episode, an unconfigured-but-episode-active source listed with `isCustomOverride: false`, `PATCH` merges `pricingRule`/`priceSyncMode` while preserving unrelated config keys, `getAsSource` unions config-derived and episode-derived destinations and excludes self.

Per this session's user-approved exception, this commit used `git commit -s --no-verify` (pre-existing, unrelated `apps/e2e` `node_modules` gap in the pre-commit hook).

Closes #3146


apps/api/src/integrations/application/services/connection-pricing-sync.service.ts
```diff
@@ -0,0 +1,163 @@
...
+  type Connection,
+  type ConnectionConfig,
+  readPricingRuleConfig,
+  readPriceSyncModeConfig,
+  readPricingRuleForSource,
+  readPriceSyncModeForSource,
+} from '@openlinker/core/identifier-mapping';
+import {
+  PRICE_CHANGE_EPISODE_REPOSITORY_TOKEN,
+  type PriceChangeEpisodeRepositoryPort,
```
norbert-kulus-blockydevs, line 33: **BLOCKING — cross-context `*RepositoryPort` import; `pnpm lint` fails.**

```
node scripts/check-cross-context-imports.mjs
  apps/api/src/integrations/application/services/connection-pricing-sync.service.ts:33
    import: { PriceChangeEpisodeRepositoryPort } from '@openlinker/core/listings'
    rule:   matches deny pattern RepositoryPort$
```

`check:invariants` is chained into `pnpm lint`, so the gate CLAUDE.md requires before every commit is red. **Allow-listing is not the fix here** — the core-to-core list is empty as of #2791, which additionally fails the build on a *stale* row (*"a standing licence for the coupling to come back"*).

**A compliant seam already exists and is exported from the same barrel.** `IPriceChangesService` declares `countOpen(filters)` with the exact signature line 135 calls; only the two list reads need adding. Extending an `I*Service` for a cross-context read is the established pattern (`ISyncJobsService.findLastSucceededJob` for `analytics-trust`; `IOrderRecordService.getEarliestOrderDateByConnection` for #2083).

Adding `listOpenSourceConnectionIds(destinationConnectionId)` / `listOpenDestinationConnectionIds(sourceConnectionId)` returning **distinct ids** fixes this *and* the read-cost finding below in one move, since both callers want ids rather than hydrated rows.

apps/api/src/integrations/application/services/connection-pricing-sync.service.ts
```diff
@@ -0,0 +1,163 @@
...
+      sourceOverridesMode[sourceId] = setting.mode;
+    }
+
+    const updatedConfig: ConnectionConfig = {
+      ...connection.config,
+      pricingRule: { default: input.default.rule, sourceOverrides: sourceOverridesRule },
+      priceSyncMode: { default: input.default.mode, sourceOverrides: sourceOverridesMode },
+    };
+
+    const updated = await this.connectionService.update(connectionId, { config: updatedConfig });
```
norbert-kulus-blockydevs, line 78: **BLOCKING — the write correctly goes through `IConnectionService`, and the validator it reaches no longer validates anything.**

`ConnectionService.validateStockAndPricingConfig` (`connection.service.ts:273`) reads the **flat** shape — `const { type, percent, rounding } = pricingRule` — and every check is guarded on `!== undefined`. Under #3158's `{default, sourceOverrides}` all three are `undefined`, so the function returns having fired nothing. I transcribed it and ran both shapes:

```
--- LEGACY flat (what #2610 was written against) ---
flat margin 120 (the #2610 threat)  -> REFUSED
flat type garbage                   -> REFUSED
--- NEW nested, written by this endpoint ---
nested default margin 120           -> PASS
nested default margin 99999         -> PASS
nested sourceOverride margin 500    -> PASS
nested default type garbage         -> PASS
nested rounding garbage             -> PASS
nested negative percent             -> PASS
```

Its own docblock says why this matters: *"core degrades a margin of 100% or more back to the catalogue price, so without this the operator saves happily and then quietly publishes an unchanged price."* `applyPricingRule` still does `percent >= 100 ? basePrice : …`.

`PricingRuleDto` does not close it — `percent` has `@IsNumber()` + `@Min(0.0001)` and **no `@Max`**, no margin ceiling. Of those six cases the DTO catches four; the one it misses is the only one #2610 calls out as mattering most. And `PATCH /connections/:id` with a raw nested `pricingRule` — the JSON config editor path the docblock names — is unvalidated entirely.

The shape change is #3158's, but this is the endpoint whose whole purpose is writing that shape. **Fix:** extract `validateOneRule` and apply it to `default` + every `sourceOverrides` entry (keeping the legacy flat arm), and add a margin ceiling to `PricingRuleDto` so the boundary answers a field-naming 400 — which is also this issue's own AC.

apps/api/src/integrations/application/services/connection-pricing-sync.service.ts
```diff
@@ -0,0 +1,163 @@
...
+      if (!knownAsSource) {
+        continue;
+      }
+
+      entries.push({
+        destinationConnectionId: destination.id,
+        destinationLabel: destination.name,
+        effectiveMode: readPriceSyncModeForSource(destination.config, connectionId),
+        effectiveRuleSummary: readPricingRuleForSource(destination.config, connectionId) ?? {
+          type: 'passthrough',
```
norbert-kulus-blockydevs, line 108: **IMPORTANT — a GET→Save round-trip turns "no pricing rule" into a real rule and silently switches on rounding.**

`buildView` erases the distinction: `pricingConfig.default ?? {type:'passthrough', percent:0, rounding:'none'}`. But `PricingRuleConfig.default` is `PricingRule | null`, and `applyPricingRule`'s docblock is explicit that the two are **not** equivalent:

> `rule === null` (no configured rule) returns `basePrice` completely unchanged — no rounding, no float cleanup … Once a rule is configured (even `type: 'passthrough'`), rounding applies.

`UpdatePricingSyncDto.default` is required and `PricingRuleDto.type` is required, so `default: null` is **unexpressible** through the API. On a connection that has never had a `pricingRule`, one visit to #3166's settings page and one Save persists `{type:'passthrough', percent:0, rounding:'none'}` — and every price that connection publishes from then on goes through `round2dp`. A real change to published prices, produced by saving a page without touching the default, contradicting ADR-072's promise that behaviour is unchanged until an operator opts in.

**Fix:** make `default.rule` nullable end to end and have `buildView` report `rule: null` rather than synthesising a passthrough. Same for `effectiveRuleSummary` and `getAsSource`.

apps/api/src/integrations/http/dto/pricing-sync-setting.dto.ts
```diff
@@ -0,0 +1,50 @@
...
+
+export class PricingRuleDto {
+  @ApiProperty({ enum: PricingRuleTypeValues })
+  @IsIn(PricingRuleTypeValues)
+  type!: PricingRuleType;
+
+  @ApiProperty({ description: 'Required (and must be > 0) unless type is passthrough.' })
+  @ValidateIf((o: PricingRuleDto) => o.type !== 'passthrough')
+  @IsNumber()
+  @Min(0.0001)
```
norbert-kulus-blockydevs, line 32: **IMPORTANT — the API refuses a payload its own GET produced.**

`coercePricingRule` coerces a missing/non-finite `percent` to `0`, and `validateStockAndPricingConfig` accepts `percent: 0` (it only rejects `< 0`), so `{type:'markup', percent:0}` is a legal, currently-saveable config via `PATCH /connections/:id`. Then:

1. `GET /pricing-sync` returns `default.rule = {type:'markup', percent:0, rounding:'none'}`.
2. #3166's FE clones it into its draft and submits it verbatim on Save.
3. `@ValidateIf(type !== 'passthrough')` + `@Min(0.0001)` → **400, "percent must not be less than 0.0001"**.

The settings page is unsavable on such a connection until the operator notices and edits a number they never set, and the message names an epsilon rather than a rule. `@IsPositive()` expresses the intended ">0" without the magic constant — but the round-trip still needs either `@Min(0)` or a GET that does not emit a value the PATCH rejects.

Also missing here: no `@Max` and no margin ceiling (see the blocking comment on the service).

apps/api/src/integrations/application/types/connection-pricing-sync.types.ts
```diff
@@ -0,0 +1,39 @@
...
+export interface ConnectionPricingSyncView {
+  default: ConnectionPricingSyncSetting;
+  sources: ConnectionPricingSyncSourceEntry[];
+}
+
+export interface UpdateConnectionPricingSyncInput {
+  default: ConnectionPricingSyncSetting;
+  sourceOverrides: Record<string, ConnectionPricingSyncSetting>;
+}
+
```
norbert-kulus-blockydevs, line 32: **IMPORTANT — welding `{mode, rule}` into one override entry materialises pricing overrides the operator never authored.**

Storage keeps the two axes **independent** (`pricingRule.sourceOverrides` and `priceSyncMode.sourceOverrides` are separate maps), and #3162's `optInAutomatic` writes **only** the mode half. This contract collapses them into one required pair, and `buildView` reports `isCustomOverride` as an **OR** across both maps. Concrete reachable sequence:

1. Operator accepts a change with "also set to Automatic" → `priceSyncMode.sourceOverrides['src-1'] = 'automatic'`, nothing in `pricingRule.sourceOverrides`.
2. `GET` returns `src-1` with `isCustomOverride: true` and `effective.rule` = **the connection default** (the fallback).
3. #3166's FE seeds `customSources` from `isCustomOverride` and submits every member with its `effective` snapshot.
4. This service writes `pricingRule.sourceOverrides['src-1'] = <the default rule as it was at page load>`.

`src-1` now carries an explicit rule override nobody wrote, permanently decoupled from the default — so a later change to the default silently stops reaching that source. This is the class #2016/#2610 guard against elsewhere (a cleared knob writes an explicit `null` rather than being conflated with absence).

**Fix:** `sourceOverrides: Record<string, {mode?: PriceSyncMode; rule?: PricingRule | null}>`, writing each half only when present, and split the response signal (`modeOverridden` / `ruleOverridden`) so a client can render "mode overridden, rule inherited".

apps/api/src/integrations/application/services/connection-pricing-sync.service.ts
```diff
@@ -0,0 +1,163 @@
...
+      ...Object.keys(syncConfig.sourceOverrides),
+    ]);
+
+    const openEpisodes = await this.episodes.findOpenForConnection(connection.id);
+    for (const episode of openEpisodes) {
+      sourceIds.add(episode.sourceConnectionId);
+    }
+
+    const sources = [];
+    for (const sourceId of sourceIds) {
```
norbert-kulus-blockydevs, line 133: **IMPORTANT — `2 + 2N` queries per read, one of them unbounded, on a viewer-callable route.**

`buildView` (reached by **both** `getPricingSync` and the tail of `updatePricingSync`) issues `connections.get`, then `episodes.findOpenForConnection` — which has no pagination and ends in a bare `getMany()`, hydrating **every open episode for the destination** into domain entities purely to derive a set of distinct source ids — and then, sequentially in this loop, per source: `connections.get(sourceId)` **and** `episodes.countOpen({destination, source})`.

The N `countOpen` calls are pure waste: the array fetched above already contains every open episode for this destination, so the per-source counts are an in-memory group-by. The N `connections.get` calls should be one `list()` — #2083's rule again. The route is `@Roles('admin','operator','viewer')`, so a viewer triggers it.

Folding both into the `I*Service` method from the blocking comment above solves this and the lint gate together.

norbert-kulus-blockydevs (commented):

**The thing I checked first is right:** `updatePricingSync` writes through `IConnectionService.update`, not `ConnectionPort` — so `adapterKey` immutability, the per-plugin `ConnectionConfigShapeValidatorRegistryService` dispatch, the rate-limit bounds check and the taxonomy bootstrap all still run. (#3162 is the one that bypasses it.) Layer and folder placement match the `rate-limit-status` / `webhook-status` triple in this same directory; the nested route prefix follows the `catalog-trust` / `sync-status` convention; `as-source` really is read-only with no PATCH; and reusing core's runtime vocabulary arrays in the DTO rather than re-declaring string literals is the right instinct.

I also hunted three hypotheses and found nothing: no route-path collision (`connections/:connectionId/pricing-sync` is unique and `@Get(':id')` cannot shadow a two-segment path), no Swagger DTO class-name collisions across `apps/api` and `libs`, and `route-authorization-coverage.spec.ts` genuinely covers the new controller.

Two blocking, both inline: the lint gate, and the fact that **#2610's margin refusal is a complete no-op for the shape this endpoint writes** — I transcribed `validateStockAndPricingConfig` and ran both shapes through it; `margin: 120`, `margin: 99999`, a `sourceOverride` of `margin: 500`, a garbage `type`, a garbage `rounding` and a negative `percent` all **PASS** under `{default, sourceOverrides}`.

Three more that change behaviour operators will notice: a GET→Save round-trip silently switches on rounding; the API rejects a payload its own GET produced; and the welded `{mode, rule}` contract materialises pricing overrides nobody wrote.
---
*Part of a whole-stack review of #3158–#3168 (epic #3139). Cross-cutting items tracked once: `pnpm lint` is red on the stack (`check-cross-context-imports` ×4, `check-architecture-gates` ×1); `docs/architecture-overview.md § Listings` is untouched by all ten PRs though ADR-072 names it the primary doc section; the mockup has not changed since #3155 despite ~20 behavioural divergences; and four sites cite "ADR-072 decision 6/7/9" using **epic #3139's** numbering — ADR-072 has 8 decisions and they diverge from #6 onward.*

apps/api/src/integrations/application/services/connection-pricing-sync.service.ts
```diff
@@ -0,0 +1,163 @@
...
+    @Inject(PRICE_CHANGE_EPISODE_REPOSITORY_TOKEN)
+    private readonly episodes: PriceChangeEpisodeRepositoryPort
+  ) {}
+
+  async getPricingSync(connectionId: string): Promise<ConnectionPricingSyncView> {
+    const connection = await this.connections.get(connectionId);
+    return this.buildView(connection);
+  }
+
+  async updatePricingSync(
```
norbert-kulus-blockydevs, line 59: **IMPORTANT — no server-side destination gate, so ADR-072 decision 2 is enforced only by the browser.**

Decision 2 is binding: *"A default rule + optional per-source overrides live on the DESTINATION connection. The feeding source connection's own page gets a read-only rollup, **never its own editable copy**."* The only thing implementing that is #3167's `pricing-sync-page.tsx:30`, in the browser. `updatePricingSync` has no equivalent check, so `PATCH` will happily stamp `config.pricingRule` onto a pure `ProductMaster` source connection — giving it precisely the editable copy the decision forbids.

This controller's own docblock cites #2610's reasoning that the form is not the only way in; the same reasoning applies to its own gate.

**Fix:** refuse the write when the connection is not a viable destination, in the shape #2407 uses (400 naming the remedy). Resolve destination-ness from `resolveAdapterMetadata().supportedCapabilities` rather than `enabledCapabilities` — per #2085 the latter is stamped at create and never retro-filled, and `resolveAdapterMetadata` constructs no adapter and works on a disabled connection.

**Also, no concurrency token.** `ConnectionRepository.update` is a full-row read-modify-write `save()`, and #3162's `optInAutomatic` plus #3165's Undo write the **same two keys** from two other places, unserialised. Accept an `expectedUpdatedAt` so a clobber 409s instead of succeeding.

apps/api/src/integrations/http/connection-pricing-sync.controller.ts
```diff
@@ -0,0 +1,101 @@
...
+  @Patch()
+  @Roles('admin')
+  @ApiOperation({ summary: 'Save the default + per-source pricing rule and sync mode (explicit-Save, no partial patch).' })
+  @ApiResponse({ status: 200, type: ConnectionPricingSyncResponseDto })
+  @ApiResponse({ status: 400, description: 'A rule/mode entry does not match the accepted shapes.' })
+  async update(
+    @Param('connectionId') connectionId: string,
+    @Body() dto: UpdatePricingSyncDto
+  ): Promise<ConnectionPricingSyncResponseDto> {
+    const sourceOverrides = await this.validateSourceOverrides(dto.sourceOverrides ?? {});
```
norbert-kulus-blockydevs, line 55: **IMPORTANT — the hand-rolled loop drops the global pipe's whitelist, and validates no key at all.**

`main.ts:73` configures the global pipe with `whitelist: true, forbidNonWhitelisted: true`. `sourceOverrides` is decorated only `@IsObject()`, so the pipe passes its nested content through untouched; this then re-validates with `plainToInstance` + `validate`, **neither of which strips**. So keys the rest of the API refuses with a 400 survive inside each `rule` and are written straight into `Connection.config` JSONB. `coercePricingRule` normalises on read so it is not behaviourally exploitable, but unbounded operator-supplied JSON accretes in a config column on the one route that opted out of the global contract.

Separately, **the map keys are entirely unvalidated** — no `@IsUUID`, no existence check. I tested the two hostile keys:

```
own keys of sourceOverrides: [ '__proto__', 'constructor' ]
result own keys after loop:  [ 'constructor' ]
({}).mode leaked globally?   undefined
service saw source ids:      [ 'constructor' ]
```

No prototype pollution — but a `__proto__`-keyed override is accepted with a **200 and silently discarded**, which is the wrong answer either way. And an arbitrary junk id persists forever, then costs one `connections.get` on every GET and renders as `'Unknown connection'`, indistinguishable from a genuinely deleted connection.

**Fix:** `plainToInstance(..., { excludeExtraneousValues: true })` with `@Expose()`, and `@IsUUID` on the keys. Note the same validation must also cover #3162's `optInAutomatic` writer — validation that lives in one controller is validation the other path bypasses.

piotrswierzy (changes requested):

**On scope, the good news first:** the `gh pr diff` figure of 54 files / +3502 is entirely an artefact of a stale base, **not** scope creep. The genuine slice is nine new files, ~590 lines — `GET`/`PATCH /connections/:id/pricing-sync` plus a read-only `GET …/pricing-sync/as-source`. Right-sized, and nothing in it belongs in a sibling slice.

And it **does not create a second pricing authority**: `applyPricingRule` / `applyRounding` / `round2dp` are untouched, so the slice only adds a *selection* dimension over the one arithmetic, and `check-stock-and-pricing-preview-mirror.mjs` is still complete (it guards those four by token comparison and none changed).

### BLOCKING — merging this reverts reviewed fixes in 44 files, including the validation its own writes depend on

I verified against the branches rather than the diff view:

```
merge-base:  9110a51ba
base tip:    2eab490df
behind by:   8 commits
both-modified files: 44
```

The branch also carries **re-authored duplicates** of the three earlier slices (`f62aba1ca`, `5efd2d930`, `ac190c0b1` mirror base's commits by message, not SHA), so a three-way merge takes this branch's **pre-review-fix** side on every non-conflicting hunk. Two concrete regressions, both verified:

**1. `connection.service.ts` loses the server-side margin refusal.** Count of `validateOnePricingRule`:

| ref | occurrences |
|---|---|
| `origin/3145-price-changes-api` (base) | **5** |
| `origin/3146-…` (this head) | **0** |

That commit (`28c4e002e`) is the *only* thing enforcing #2610's "a margin ≥ 100% is refused server-side as well as in the form" under the nested `{default, sourceOverrides}` shape **this PR writes**. Merging as-is ships the write path without its guard. The same commit carries the `mergeStructuredIntoConfig` fix that stops `EditConnectionForm` destroying `sourceOverrides` on any unrelated save — without it, the two writers onto `config.pricingRule` actively fight.

**2. `app.module.ts` re-adds `PriceChangeObserverBindingModule`** and restores the exact comment base commit `cd6a85727` deleted as *"dead wiring with a false justification"* — the reinstated text claims "a manually-triggered master-product-sync from the API surface", which base's own review recorded does not exist anywhere in `apps/api/src`.

**Rebase onto `2eab490df` (or onto `main` once #3162 lands) and re-push**, then re-verify both regressions are gone. Everything below is contingent on that.

### BLOCKING — a `*RepositoryPort` crosses a context boundary in production code

`connection-pricing-sync.service.ts` (32–34, 50–51) imports `PRICE_CHANGE_EPISODE_REPOSITORY_TOKEN` / `PriceChangeEpisodeRepositoryPort` from `@openlinker/core/listings`. `engineering-standards.md § Cross-context dependencies` lists `*RepositoryPort` as an unconditional **deny**, and `check-cross-context-imports.mjs` has no allow-list entry for this file — the only price-change entry base added is for an int-spec, whose comment says *"Rewire via a read/application service once #3162's `PriceChangesService` lands (#722)."* **It has landed, on this PR's own base.**

So `pnpm check:invariants` (→ `pnpm lint`) fails. The test plan says lint was verified "via `npx eslint`", which does not run `check:invariants`, and the commit used `--no-verify` — which is how it escaped. `CLAUDE.md § Never do these` lists that flag without exception.

Fix needs **no allow-list entry**: inject `IPriceChangesService`, already exported from the same barrel and exposing exactly the two reads used here — `countOpen({destinationConnectionId, sourceConnectionId})` and `listOpen({…})`.

### IMPORTANT — validation is asymmetric between `default` and `sourceOverrides`

`connection-pricing-sync.controller.ts` (88–89) calls `validate(instance)` with **no options**, while the global pipe is `{whitelist: true, forbidNonWhitelisted: true}` (`main.ts:73-77`). So an unknown key on a `sourceOverrides` entry is accepted and written verbatim into JSONB, while the same key on `default` is a 400. Pass the same options so the hand-rolled pass matches the pipe it stands in for.

### IMPORTANT — `sourceOverrides` keys are never validated

`PATCH` with `{"sourceOverrides": {"not-a-connection": {…}}}` persists that key permanently, and the subsequent `GET` renders `sourceLabel: 'Unknown connection'` with no way to tell a typo from a deleted connection. Admin-only bounds the blast radius, but an endpoint with an unconstrained keyspace is a write-anything-into-config surface. Validate each key as a UUID at minimum; better, resolve against `ConnectionPort` and 400 naming the unknown id — the shape `validateSourceOverrides` already uses for rule errors.

### IMPORTANT — a bare catch-to-null conflates deletion with failure

`connection-pricing-sync.service.ts:134` — `connections.get(sourceId).catch(() => null)` renders both a deleted connection and a failed database read as `'Unknown connection'`. This repo states the rule repeatedly (§ Returns: *"absence and failure are indistinguishable and no reassurance is ever rendered"*; § Currency: `'unknown'` is a distinct value, never a healthy-looking default). Narrow to `ConnectionNotFoundException`.

### IMPORTANT — request DTOs reused as response schema types

`connection-pricing-sync-response.dto.ts` (12, 18, 29, 44) reuses `PricingSyncSettingDto` / `PricingRuleDto` for responses. `@nestjs/swagger` keys definitions by **class name**, so one schema serves both directions and any future request-side tightening silently narrows the documented response — the same hazard base's own review flagged on `PriceChangeResponseDto`. Also line 43: `effectiveMode!: string` drops the union; should be `@ApiProperty({ enum: PriceSyncModeValues })` as the request DTO correctly does.

### IMPORTANT — unguarded read-modify-write on a whole-map replacement

`get()` → build → `connectionService.update({config})`, and `ConnectionService.update` is a full-row `save()`. Two admins saving concurrently: the later write silently discards the earlier one's **entire** `sourceOverrides` map. #3162 added `expectedVersion` for exactly this class. Carry a version token and reject a stale `PATCH`, serialize per connection with `SyncLockPort` (the `invoiceIssueLockKey` shape), or at minimum state last-write-wins in the controller docblock.

### IMPORTANT — no controller spec and no int-spec

`validateSourceOverrides` is the only non-trivial hand-rolled logic in the slice — the per-entry loop, the error assembly, the offending-source-id naming, the `?? {}` default — and none is covered. And nothing asserts the three routes' status codes; the 404-via-`ConnectionExceptionFilter` and the 400-from-`ConnectionService` paths are inferred, not tested. `testing-guide.md § Integration Tests` calls a new HTTP→service→DB path a vertical slice.

### SUGGESTIONS

- The unconfigured-default literal `{type:'passthrough', percent:0, rounding:'none'}` is minted in **three** places in one file (plus twice in the spec) — the `stockSafetyBuffer` four-copies problem #2323 exists to prevent. One exported constant.
- `buildView` already holds `openEpisodes`, then issues a **second** `countOpen` per source, plus an N+1 `connections.get`. The counts are derivable from the array in hand.
- `percent` has `@Min(0.0001)` and no upper bound while the server refuses `margin >= 100` — the schema advertises what the server rejects. And `@Min(0.0001)` refuses `{type:'markup', percent:0}`, a legitimate "no markup"; per #2610 `0` and unset must round-trip apart, and the `GET` returns `percent: 0`, so an FE rendering it into a form and flipping the type gets an opaque 400.
- The service injects `ConnectionPort` for reads and `IConnectionService` for the write. #2353's `AuthorityStatusService` — same folder — deliberately does both through `IConnectionService` because *"a repository port is an intra-context contract"*. Pick one.
- Editing artefact in `update-pricing-sync.dto.ts` (5–7): *"see `PriceChangesController`... actually `ConnectionPricingSyncController`'s `updatePricingSync`"* — and that method is `update` on the controller.
- `@Param('connectionId')` takes no `ParseUUIDPipe`, unlike `ConnectionController` and base's own `PriceChangesController` — and base's #3162 review found a live route-shadowing bug caused by exactly that asymmetry. No shadowing risk here, but a non-UUID reaches the repository and surfaces as a driver error rather than a 400.

### Done well

Route authorization is correct and **structurally** covered — all three handlers carry exactly one `@Roles(...)`, write is `admin`, reads are `admin|operator|viewer`, `packer` is named nowhere, and both guard specs discover controllers by reflection so the new one is covered with no registry edit. **ADR-072 decision 2 is honoured literally** — `as-source` has `@Get` only, verifiable by grep. `connections.list()` is called with **no filters**, so a `disabled`/`needs_reauth` destination still appears in the rollup — the right call, avoiding the `analytics-trust` active-only trap. Explicit-Save whole-shape replacement matches the `StockAndPricingSection` no-partial-patch precedent, and `@Roles('admin')` is the right gating for a genuine `Connection.config` mutation.

One follow-up worth filing separately: `StockAndPricingSection`'s worked example shows the **default** rule only and cannot show a per-source override, so an operator with an override now sees an understated preview.

---

norbert-kulus-blockydevs:
Addressed all 8 review findings in 6d2eba66f (rebased onto `3145-price-changes-api`'s tip after its own #3162 fix — see note on finding 2).

**BLOCKING**

1. **Cross-context `*RepositoryPort` import.** Removed. Extended `IPriceChangesService` with `countOpenBySource` (delegates to the existing repo aggregate) and a new `listOpenDestinationConnectionIds` (backed by a new repository method, `DISTINCT destinationConnectionId WHERE resolvedAt IS NULL AND sourceConnectionId = :id`). `ConnectionPricingSyncService` now injects `IPriceChangesService` instead of the repository port — `check-cross-context-imports.mjs` passes clean, and this is the same move that fixes finding 6 (below).
2. **Validator not handling the nested `{default, sourceOverrides}` shape.** This was fixed independently on the parent branch (`28c4e002e`, part of #3158's own review pass) *before* this branch rebased onto it — `ConnectionService.validateStockAndPricingConfig` now validates `default` + every `sourceOverrides` entry, margin ceiling included. Re-verified live: nested `margin: 120`, nested `margin: 99999`, and a nested `sourceOverride` margin all now refuse with 400. Added a matching DTO-level `PricingRuleMarginCeilingConstraint` (mirrors the core wording exactly) as a faster boundary check — core stays authoritative.

**IMPORTANT**

3. **GET→Save round-trip silently switching rounding on.** `default.rule` (and `effectiveRuleSummary`, and the per-source `effective.rule`) are now `PricingRule | null` end-to-end, and `buildView`/`getAsSource` report `null` verbatim instead of synthesizing `{type:'passthrough', percent:0, rounding:'none'}`. Widened `ConnectionConfig.pricingRule`'s stored-shape type (`default: PricingRule | null`) to match — it was stricter than the read-time `PricingRuleConfig` it's coerced through.
4. **`percent: 0` round-trip rejected.** `@Min(0.0001)` → `@Min(0)` on `PricingRuleDto.percent`, matching what `coercePricingRule`/core's validator actually accept.
5. **Welded `{mode, rule}` override materializing an unauthored rule.** New `SourceOverrideSettingDto`/`ConnectionPricingSyncSourceOverrideInput`: `mode`/`rule` are independent and both optional, and `updatePricingSync` only writes the half actually supplied. Response splits `isCustomOverride` into `modeOverridden`/`ruleOverridden`.
6. **`2 + 2N` queries.** Folded into finding 1's fix — `buildView` is now one `countOpenBySource` call + one batched `connections.list()`, no per-source loop.
7. **No server-side destination gate + no concurrency guard.** `updatePricingSync` now refuses (400) when `resolveAdapterMetadata().supportedCapabilities` includes neither `OfferManager` nor `ProductPublisher` (checked against `supportedCapabilities`, never `enabledCapabilities`, per #2085). Added a per-connection `SyncLockPort` lock (mirrors `PriceSyncModeOverrideService`'s) plus an optional `expectedUpdatedAt` optimistic-concurrency field on the DTO — a lock miss or a stale value both 409.
8. **Whitelist bypass + unvalidated map keys.** Replaced the hand-rolled controller loop with a new `@ValidateSourceOverrides` decorator (the existing `@ValidateRecordValues` pattern, reused rather than reinvented), reached through the ordinary global `ValidationPipe`. It validates every entry with `whitelist: true, forbidNonWhitelisted: true` *and* validates every map key as a UUID — closing the `__proto__` case (which was actually being silently dropped via JS's own `result['__proto__'] = x` prototype-set footgun, not stored) plus the "junk key persists forever" cost, since a non-UUID key now 400s outright.

**Cross-cutting note (finding 8):** the same whitelist gap likely exists on `#3162`'s `optInAutomatic` writer if it accepts any nested caller-supplied shape — flagging for visibility, not fixed here (out of scope for this PR's own files).

**Verification:** `pnpm --filter @openlinker/core type-check/build`, `pnpm --filter @openlinker/core lint`, `pnpm check:invariants`, and a targeted `eslint` pass over every touched file are all clean. Targeted jest on `connection-pricing-sync.service.spec.ts` passes 13/13 (expanded from 4 to cover every finding, including the lock/concurrency/destination-gate paths). Added integration coverage for the new `listOpenDestinationConnectionIds` method alongside the existing `countOpenBySource` int-spec. The repo-wide `pnpm -r lint` pre-commit hook could not complete in this sandbox (same documented constraint as the parent branch's own fix commits) — committed with `--no-verify` after the independent verification above.

norbert-kulus-blockydevs:
## Response to `/pr-review` (Piotr, 2026-09-11T10:40:11Z)

Note on staleness: this review landed 7 minutes *before* the parent branch's own review-fix commit (`9d72e90d6` on `3145-price-changes-api`, at `10:47:40Z`). I rebased onto the latest `origin/3145-price-changes-api` first (skipping this branch's own re-authored duplicate-message ancestor commits via `git rebase --onto`), then verified every finding against the resulting code before touching anything.

### BLOCKING — verified already resolved by the rebase

**1. 44-file revert of #2610's margin-refusal fix + reinstating a deleted `PriceChangeObserverBindingModule`.** Confirmed: after rebasing onto `origin/3145-price-changes-api`, `validateOnePricingRule` appears **5 times** in `connection.service.ts` (matching the parent base exactly), and `PriceChangeObserverBindingModule` is absent from `apps/api/src/app.module.ts` (worker-only, as intended). This was a branch-staleness artifact, closed by the rebase — no code change needed once history was fixed.

**2. Cross-context `*RepositoryPort` import.** Confirmed: `connection-pricing-sync.service.ts` injects `IPriceChangesService` (`@openlinker/core/listings`), never `PriceChangeEpisodeRepositoryPort` directly. Re-ran `node scripts/check-cross-context-imports.mjs` against the file specifically — passes clean (3031 cross-context imports checked, 0 new allow-list entries needed).

### IMPORTANT — genuinely fixed in this pass

- **Asymmetric/missing `sourceOverrides` validation** — already closed by the prior fix round via `@ValidateSourceOverrides` (UUID-keyed, `whitelist`/`forbidNonWhitelisted` per entry). I added the test coverage that was missing (see below), including a case exercising the `__proto__`-as-plain-key discard the original hand-rolled loop silently accepted.
- **Bare catch-to-null conflating deletion with failure** — already closed: `buildView` now resolves source labels from one batched `connectionService.list()` call into a `Map`, not a per-source `.get(id).catch(() => null)`.
- **Request DTOs reused as response schema** — **fixed now.** Split `PricingRuleResponseDto` / `PricingSyncSettingResponseDto` (plain `@ApiProperty`, no `class-validator` decorators) out of the request-side `PricingRuleDto` / `PricingSyncSettingDto`, and typed `effectiveMode` against `PriceSyncModeValues` instead of a bare `string`.
- **Unguarded read-modify-write** — already closed: `expectedUpdatedAt` optimistic-concurrency check + a per-connection `SyncLockPort` lock around `get` → `update`.
- **No controller spec / no int-spec** — **fixed now.** Added `connection-pricing-sync.controller.spec.ts` (thin pass-through + DTO projection), `validate-source-overrides.decorator.spec.ts` (exercises the real `class-validator` pipeline: UUID key rejection, whitelist enforcement, nested validation), and `apps/api/test/integration/connection-pricing-sync.int-spec.ts` (all three routes over real HTTP, including the viable-destination 400, the stale-`expectedUpdatedAt` 409, and role-gating).

### Suggestions addressed

- Removed the redundant `ConnectionPort` injection — `IConnectionService` already exposes `get()`/`list()`, so the service now depends on one connection seam.
- Added `ParseUUIDPipe` to `:connectionId`, matching `connection-sync-status.controller.ts` and other `connections/:connectionId/*` siblings.
- The unconfigured-default literal (`{type:'passthrough', percent:0, ...}` minted 3×) and the N+1 `connections.get()` were both already resolved as side effects of the prior fix round (verbatim `rule: null` + the batched `list()` read).
- `percent`'s floor/ceiling and the docblock typo were also already resolved by the prior round.

### Verification

Independently ran (no `--no-verify` shortcuts on the checks themselves, just skipping the git hook per this session's standing instruction):
- `pnpm --filter @openlinker/core lint` — 0 errors (92 pre-existing warnings, unchanged)
- `pnpm --filter @openlinker/api lint` — 0 errors (17 pre-existing warnings, unchanged)
- `pnpm check:invariants` — all ~50 checks pass, including `check-cross-context-imports` and `check-service-interfaces`

Pushed as `31e4f9eda` (force-with-lease, since the rebase rewrote history).

piotrswierzy (approved):
## `/pr-review` (round 2) — ✅ Approve

Both previous findings are addressed, and the second one is resolved better than I asked for.

**The duplicate-rule concern is gone, and replaced with a stated precedence rather than a deletion.** `pricing-sync-setting.dto.ts:10-20` now records that `ConnectionService.validateOnePricingRule` is *the authoritative gate* (reached through `ConnectionService.update`) and that the DTO's decorators are "a same-shaped, faster 400 at the boundary". That is exactly the posture `docs/engineering-standards.md` takes — the browser is not a trust boundary, the raw JSON editor and curl and MCP all bypass the form, so a DTO decorator is a convenience for the OpenAPI document and a fast rejection, never the gate.

The reasoning behind `@Min(0)` rather than a positive epsilon is the part worth keeping: `coercePricingRule` coerces a missing or non-finite `percent` to `0` and the service validator accepts `percent: 0`, so `{type: 'markup', percent: 0}` is a legal GET-able config — **a stricter floor here would make the endpoint refuse a payload its own GET had just produced.** A round-trip that fails on unchanged data is the kind of defect that only shows up in an operator's hands, and catching it at design time is the right call.

### SUGGESTION — still behind `main`

`origin/3146-connection-pricing-sync-settings-api` is **6 commits** behind `origin/main` (down from 8 last round). Not blocking and not this PR's fault — it sits seventh in an eleven-deep stack, so it inherits whatever its predecessors carry — but the whole chain will want a rebase before the first merge, and the `1878000000000` / `1879000000000` migration slots (see my #3158 review) are the thing most likely to bite on the way in.

---

norbert-kulus-blockydevs:
Merged the updated \`3145-price-changes-api\` in (merge commit \`7369ded6c\`, no rebase — 5 more PRs are stacked on top of this one). Branch is now caught up: 0 commits behind, was 6.

Picked up along the way:
- The \`1878000000000\`/\`1879000000000\` migration slot you flagged — already resolved upstream by the time this merged: \`1878000000000-create-price-change-episodes.ts\` was renumbered to \`1880000000000\` on the parent branch, and this PR carries no migrations of its own, so there was nothing to reconcile here.
- The new \`libs/core/src/listings/testing\` package-exports subpath + \`price-change-derived-filter.fixtures.ts\` — this PR's own specs don't touch those fixtures, so no follow-up needed there either.

Merge was clean, no conflicts. Re-ran the full verification pass post-merge:
- \`@openlinker/core\` type-check + lint: clean (0 errors)
- \`@openlinker/api\` type-check + lint: clean (0 errors)
- \`pnpm check:invariants\`: all green, including \`check-migration-timestamps\` and \`check-cross-context-imports\`

Nothing else outstanding from the review — thanks for the approve.

norbert-kulus-blockydevs:
Caught this branch up with the two CI-fix commits from `3145-price-changes-api` (merged, not rebased, to avoid disrupting the 5 stacked PRs on top):

- `fix: export PriceChangeDetectionService from libs/core/src/listings/services/index.ts` (missing barrel export)
- `fix: update hardcoded lane-count assertion in handler-registration.service.spec.ts (29 -> 30 for the bulk lane)`

Both caught by real GitHub Actions CI runs, not review feedback. Merge was clean, no conflicts. Verified `pnpm --filter @openlinker/core type-check`, `pnpm --filter @openlinker/api type-check`, and `pnpm check:invariants` all pass.

piotrswierzy (commented):
## Tech Lead review

Independent re-read, and I **agree with the round-2 approval on substance**. This is the best-behaved slice I have read in this stack, and several of the things it gets right are exactly the things its sibling #3162 gets wrong — worth naming, because the contrast is instructive rather than incidental.

**Route authorization is textbook.** Three routes, one decorator each (`@Roles('admin','operator','viewer')` on both reads, `@Roles('admin')` on the write), nothing class-level and widening, nothing undecorated — `engineering-standards.md § Route authorization` satisfied without needing `route-authorization-coverage.spec.ts` to catch anything. The `admin`-on-write split is argued from the act rather than from the folder: this PATCHes `Connection.config`, so it matches `ConnectionController`'s gating rather than #3145's operator-level Listings gating, and both controllers' docblocks now cross-reference each other on that divergence.

**Every `@Param` carries `ParseUUIDPipe`**, and `@ValidateSourceOverrides` validates the `Record` **keys** as UUIDs, not just the values (`validate-source-overrides.decorator.ts:67-70`). That is the boundary-validation rule applied to the one shape `class-validator` has no first-class nested decorator for, and it is precisely the hole #3162 leaves open on its own `:id` params — so this slice is not merely compliant, it closes the class of defect its base introduces.

**Layering is clean.** The service injects `IPriceChangesService`, `IConnectionService`, `IIntegrationsService` and `SyncLockPort` — service interfaces and ports only, no `*RepositoryPort` crossing a context boundary, which is the rule `engineering-standards.md § Repository Ports Pattern` states and `check-cross-context-imports.mjs` enforces. Writing through `IConnectionService.update` rather than `ConnectionPort` keeps `adapterKey` immutability, the per-plugin config-shape validator dispatch, the rate-limit bounds check and the taxonomy bootstrap all in the path. Interface, Symbol token and implementation are in separate files under `application/interfaces/` + `application/services/`.

**And the `pricingRule` type widening is reasoned from behaviour, not from the compiler.** `default` is optional *and* nullable because a stored `null` means "unconfigured", and `applyPricingRule` applies rounding for a configured `passthrough` but not for `null` — so synthesizing `{type:'passthrough'}` to dodge the nullability would have been a silent behaviour change. That is the right way to justify a core type change.

### Findings

**[IMPORTANT]** — `apps/api/test/integration/listings-price-change-episode.int-spec.ts:331`

> **CI is red, and one of the four failures is this PR's own new spec.** Three are inherited from #3162 and will clear when that PR is fixed; the fourth is introduced here:
> ```
> ● listOpenDestinationConnectionIds() reports distinct destination ids for a source,
>   deduplicated across variants (#3163 review)
>     PriceChangeEpisodePersistenceError: … invalid input syntax for type uuid: "no-such-source"
> ```
> `price_change_episodes.sourceConnectionId` is a `uuid` column, so the literal `'no-such-source'` is rejected by Postgres rather than returning an empty array. Unlike #3162 this is **not** a live production defect — `getAsSource` reaches the port only with a `ParseUUIDPipe`-validated `connectionId`, and the PATCH path's source ids are `isUUID`-checked by the decorator, so no caller-supplied string can get there malformed. It is a test asserting a contract (*"a source with no open episodes returns `[]`"*) that the port honours only for a **well-formed** uuid.
>
> Fix is one line: use a real random uuid for the no-match case. If the port genuinely intends to tolerate an arbitrary string, say so on `PriceChangeEpisodeRepositoryPort.listOpenDestinationConnectionIds`'s docblock and guard it — but the uuid is the better answer, since every real caller is already uuid-validated.

**[SUGGESTION]** — branch base

> Still stacked seventh of ten and behind `main`. Not this PR's fault and not blocking, but the chain wants a rebase before the first merge, and the migration-slot collisions flagged on #3158 are the thing most likely to bite on the way in.

### Verdict

🔄 **Approve with changes** — I agree with the existing approval; the design, layering, gating and validation are all correct and I found nothing to add to that assessment. Posted as a comment rather than a second approval because the gate is red.

1. Point the `listOpenDestinationConnectionIds` no-match assertion at a real uuid.
2. Merge after #3162 goes green (its three failures are inherited into this branch's run).

---

norbert-kulus-blockydevs:
Pushed `fe05ab796` (plus `0c381ffe0`, the parent catch-up).

**The fixture.** `listOpenDestinationConnectionIds('no-such-source')` now reads `UNKNOWN_CONNECTION_ID = '88888888-8888-4888-8888-888888888888'`, declared beside the existing `UNKNOWN_EPISODE_ID` with the same explanation one column over. Your read of it was exactly right — the port's "no matching row" contract only applies to a well-formed id, and every real caller is uuid-validated before it gets there.

I also moved the `countOpenBySource()` assertion at `:301` onto the same constant, which your review did **not** flag. It never failed — `bySource.has(...)` is a `Map` lookup with no SQL round trip — but a non-uuid key cannot exist in that map under any circumstances, so `expect(bySource.has('no-such-source')).toBe(false)` was vacuously true and proved nothing. It now asserts the thing the case means: a real-shaped connection id with no open episodes is absent from the map. Say the word if you would rather that stayed as it was.

**Verified against real Postgres**, not just reasoned about:

```
pnpm exec jest --config ./test/jest-integration.cjs listings-price-change-episode
Test Suites: 1 passed, 1 total
Tests:       27 passed, 27 total
```

plus `pnpm --filter @openlinker/api type-check` and `pnpm check:invariants` clean.

**On the other two red checks, neither of which is this PR's code:**

- `Integration Tests` — the three inherited failures were `#3162`'s `'no-such-id'` fixtures, fixed there in `de6f5670a`. This branch now carries that fix through the merge.
- `Docker Build Smoke Test` — a network flake, not a build error. The log ends `failed to fetch oauth token: Post "https://auth.docker.io/token": read tcp …: connection reset by peer` while resolving `node:22.23.1-alpine3.24`. Nothing to fix; this push re-runs it.

One thing worth flagging for the merge, since it affects the order you asked about: `#3162`'s branch was carrying a broken `price-change-apply.service.spec.ts` inherited from `#3161`, so its run would have gone red on a test unrelated to either PR. Fixed at the source on `#3161` (`9b80e6f44`) and cascaded through `3145` (`6e6f53846`), so the whole chain below it now has it.

**On the base:** still stacked seventh and still behind `main` — deliberately not addressed here. Merging `main` into a mid-stack branch would diverge it from its own parent chain; the catch-up belongs at `#3159` and cascades down, which is what I did for the `#3161` fix above.

piotrswierzy (approved):
## Tech Lead re-review (`c6b092b3` → `a404aad1`)

Exactly one substantive delta commit — `fe05ab79` — everything else in the range is the `3145` branch merging forward, which I have just reviewed separately on #3162. **My single prior finding is closed.** `listings-price-change-episode.int-spec.ts` now introduces `UNKNOWN_CONNECTION_ID = '88888888-8888-4888-8888-888888888888'` beside the existing `UNKNOWN_EPISODE_ID`, and both `listOpenDestinationConnectionIds()` and `countOpenBySource()` assert against it instead of `'no-such-source'`. That is the right shape rather than the minimum one: the `listOpenDestinationConnectionIds` case was genuinely red (`sourceConnectionId` is a `uuid` column, so Postgres rejected the literal before the port's empty-array contract could apply), while the `countOpenBySource` case was a `Map.has` lookup that could never fail — a non-uuid key cannot exist in that map under any circumstances, so it asserted nothing at all. Both now assert a real-shaped id with no open episodes, and the commit message states plainly why this is a test-fidelity fix and not a live defect: every production caller is uuid-validated before the port (`ParseUUIDPipe` on the route, `@IsUUID()` on the source-override keys), so the port is only ever asked about well-formed ids.

Re-checked the rest of the surface at head and found nothing regressed by the merge. All three routes on `ConnectionPricingSyncController` declare exactly one authorization decorator (`GET` and `GET as-source` at `admin|operator|viewer`, `PATCH` at `admin`), every `:connectionId` is `ParseUUIDPipe`-piped, `IConnectionPricingSyncService` sits behind a Symbol token per the DI convention, the service lives in `apps/api/src/integrations` — correct, since writes must route through `IConnectionService` for `validateStockAndPricingConfig` / `validateConfigShape` and a core service cannot reach an app-layer construct — and the `@Roles('admin')` write gate agrees with the refusal #3162's controller documents itself as deliberately matching.

### Findings

### Verdict
✅ **Approve**

Same CI caveat as #3162, stated rather than assumed: Type Check, Docker Build Smoke, PHP Unit and Detect-changed-paths are green; Lint / Test / Build / **Integration Tests** are still queued at 0s. Integration Tests is the job the fixture finding was red on — I have verified the cause is fixed in source, not that the job has re-run. `mergeable: MERGEABLE`; `UNSTABLE` reflects those pending checks, not a conflict. Note this PR stacks on #3162, so merge order matters.

---

norbert-kulus-blockydevs:
No findings — thanks. Note for the merge: this branch now carries the `1884000000000` migration renumber from the base, resolving the four-way `1881000000000` collision (#3201, #2069, #2202) for the stack.

piotrswierzy (approved):
## Tech Lead re-verification (head unchanged at `aadc6fe`)

**The CI condition on my previous approval is RESOLVED: `Integration Tests` passed (32m24s)** — and so did `Lint` (5m12s), `Test` (15m15s) and `Type Check`. Only `Build` is outstanding, and it is `queued` behind a self-hosted runner backlog, never failed. `MERGEABLE`, no conflict.

Re-verified in code: `UNKNOWN_CONNECTION_ID = '88888888-…'` replaces `'no-such-source'` in both `listOpenDestinationConnectionIds()` and the `countOpenBySource()` `Map.has` assertion that could never fail; all three routes carry exactly one auth decorator (`@Roles('admin','operator','viewer')` on both reads, `@Roles('admin')` on the write); `ParseUUIDPipe` is on every `:connectionId`; and `IConnectionService` is the sole injected dependency, so writes keep going through its re-validation.

Stack note: this is position 3 of 5 and its base (#3162) has picked up one IMPORTANT finding this round — an already-merged migration edited in place. Nothing in that touches this PR's diff; it only means #3162 lands first.

### Findings

### Verdict
✅ **Approve**

---

piotrswierzy (approved):
## Tech Lead re-verification (head unchanged at `c1f9040`)

Head is unchanged since approval and `main` has not moved, so the standing review is reproduced rather than repeated. CI: a fresh run (34845733508) has re-queued the long jobs — **Type Check, Docker Build Smoke Test, PHP Unit Tests, Scaffolded Adapter Builds, Detect changed paths all pass; Build, Lint, Test and Integration Tests are queued behind the self-hosted runner backlog. Nothing has come back red on any run.** (Integration Tests passed on the previous run of this same head; the current run re-queued it.) Re-spot-checked the four things the approval rested on: `UNKNOWN_CONNECTION_ID` present at both call sites; three routes, one auth decorator each (`@Roles('admin','operator','viewer')` on both reads, `@Roles('admin')` on the PATCH); `ParseUUIDPipe` on all three `:connectionId` params; the service injects `CONNECTION_SERVICE_TOKEN` / `IConnectionService` alone and writes through `IConnectionService.update` so the re-validation is reached.

Stack: base is `3145-price-changes-api`, i.e. this sits on `main → #3159 → #3162 → here`. **Both #3159 and #3162 still carry an open BLOCKING** — each edits the already-merged migration `apps/api/src/migrations/1880000000000-create-price-change-episodes.ts` in place, so an already-migrated database silently diverges from a fresh install. This PR carries no migration and is not directly affected, but **nothing in the chain can merge until those two are fixed.** `mergeable: MERGEABLE`, `mergeStateStatus: UNSTABLE` (pending checks only).

### Findings

### Verdict
✅ **Approve**

---

Issue #3152, opened 2026-09-10
test(e2e): price-changes review queue — full state coverage + screenshots


norbert-kulus-blockydevs:
## Problem / Context

Part of #3139 (epic). **Stack position: thirteenth (post-implementation) — branches from #12's branch. Requires ALL backend (#3-#7) and frontend (#8-#12) sub-issues merged first — this is the full-stack integration checkpoint, not a per-PR unit-test task.**

Per `docs/frontend-architecture.md` § UX Mockups: "The final E2E task for the feature takes its comparison screenshots from the repo-committed mockup file... never from the Artifact URL." This issue is that final E2E task, using `apps/e2e` (Playwright, per `apps/e2e/playwright.config.ts`) against the REAL implemented app — plus, per explicit instruction, capturing screenshots of every real state for reuse in #14's documentation tutorial.

## Proposed Solution

New Playwright spec(s) under `apps/e2e/` (follow existing spec organization conventions in that folder). **Every flow and every state below must have its own test (or its own clearly-named step within a flow test) and its own named screenshot** — do not write one giant "does the feature work" test. Reference mockup `data-state`/`id` values from `docs/plans/mockups/price-changes-review-queue.html` for element selectors (the real components built in #8-#12 should carry equivalent `id`/`data-testid` attributes — if an implementer skipped adding them, that's a bug in #8-#12, not something to route around here with brittle text/CSS selectors).

### Flow 1 — Queue page states
- [ ] `queue-live`: populated table with ≥1 row (seed via API/fixture, not by waiting on real detection timing) — screenshot `price-changes-queue-populated.png`
- [ ] `queue-loading`: initial fetch in flight — screenshot `price-changes-queue-loading.png`
- [ ] `queue-empty`: zero open episodes — screenshot `price-changes-queue-empty.png`
- [ ] `queue-error`: simulated API failure, verify retry button works and re-fetches — screenshot `price-changes-queue-error.png`

### Flow 2 — Row states (one seeded episode per state, verify visual + verify action affordances match the state)
- [ ] `row-pending`: Ignore/Edit/Accept buttons present and enabled
- [ ] `row-accepted`: checkmark status, actor+time shown, no action buttons
- [ ] `row-accepted-custom`: pin icon, "Applied · custom" text, pin tooltip present on hover
- [ ] `row-ignored`: "Ignored" text + Undo button; clicking Undo re-opens the row to `row-pending` — verify against a REAL API "unresolve" call, not just client state
[...]

Issue #3148, opened 2026-09-10
feat(web): accept/edit dialogs + bulk apply with live progress


norbert-kulus-blockydevs:
## Problem / Context

Part of #3139 (epic). **Stack position: ninth — branches from #8's branch.**

Build the three action dialogs (Accept, Edit, Bulk-Accept) and the live bulk-publish progress widget from `docs/plans/mockups/price-changes-review-queue.html`. Backend contract: #6 (accept/edit/bulk endpoints), #5 (bulk batch progress mechanism).

## Proposed Solution — cite mockup `id`/`data-state`

**Single-accept dialog** (mockup `#dialog-single`, `data-state="dialog-single-accept"`):
- `#dialog-single-title` ("Publish new price"), `#dialog-single-desc` (product + connection + "from {source label}")
- Price-hero: `#dialog-single-old` / `#dialog-single-new` — big before/after, "Live now" / "Will publish" labels
- `#dialog-single-rule` — the rule sentence, e.g. "Allegro — PL price for {source} = source price **+22% margin**, rounded ending in .99 ⇒ **419,99 zł**. Computed on net price; VAT is re-added for this connection." — build this sentence from the API's returned rule summary (#6), do not hardcode strings; the "computed on net, VAT re-added" clause is REQUIRED copy per the design's gross/VAT-transparency decision.
- Opt-in checkbox `#dialog-single-optin` / `#dialog-single-optin-title` (dynamic: "Also set {source} → {connection} to Automatic") — checking this, on confirm, must ALSO call the settings PATCH from #7 to write the `sourceOverrides` entry with `mode: 'automatic'`, not just accept the one price. **On success, show a toast with an inline Undo** that reverts the mode back to `manual` (mirrors mockup's `toast(..., { undo: ... })` pattern) — this was an explicit UX finding (accidental auto-sync opt-in needs an immediate, in-context undo, not a buried settings toggle).
- Fixed disclaimer line: "This only updates the live listing — it never changes the price of an order already placed." — static copy, always shown, non-configurable.
- `#dialog-single-cancel` / `#dialog-single-confirm` ("Publish price").

**Edit dialog** (mockup `#dialog-edit`, `data-state="dialog-edit-price"`):
- `#dialog-edit-source` — "Source ({label}): **{old} → {new}**."
[...]

PR #3164, opened 2026-09-10
feat(web): Price changes tab + review queue table

### LINKED ISSUE #3147: feat(web): Price changes tab + review queue table
## Problem / Context

Part of #3139 (epic). **Stack position: eighth — branches from #7's branch.**

Build the "Price changes" tab on the Listings page and its review-queue table, per `docs/plans/mockups/price-changes-review-queue.html` (open it in a browser and click through it — this issue describes structure, the mockup IS the interaction spec). Backend contract: #6.

## Proposed Solution — cite mockup `id`/`data-state` for every piece

**Page/tab structure** (mockup `#page-listings`, `data-state="page-listings"`):
- Real route: extend the existing Listings page (`apps/web/src/pages/listings/`, per `docs/frontend-architecture.md` § Routing Conventions). Add a tab alongside the existing listings view — mockup's `#tab-all` ("All listings", unchanged existing view) / `#tab-queue` ("Price changes", new — mockup `#tab-queue-count` shows the open-episode count badge, wire to `GET /listings/price-changes`'s count).
- Header copy: mockup line 447 area, `<p>` inside `.tab-toolbar` — "One row per destination — a price that changed once at the source can produce a separate, independently-reviewable change on every marketplace or shop it's published to. Prices include VAT." Use this copy verbatim (it survived a dedicated copy-review pass).
- **Do not build "Accept all" / "Ignore all" top buttons** — removed during design as redundant with select-all + bulk bar (see epic decision 11). Only `#btn-pricing-rules` ("Pricing rules" button, opens the connection picker — built in #10) belongs in the toolbar actions area.
- **`#preview-strip` and all `data-preview="*"` buttons are a DESIGN-REVIEW-ONLY affordance in the mockup — do NOT build a shipped equivalent.** It exists so reviewers could jump to loading/empty/error states without driving the real app into them. The REAL states (loading/empty/error/populated) must instead occur naturally: loading while `GET /listings/price-changes` is in flight, empty when the list is genuinely empty, error on a failed fetch (with a "Try again" retry action, mirroring `#error-retry`'s copy and behavior), per `docs/frontend-architecture.md` § Async UX Conventions.

**Table** (mockup `#queue-body`, rows `id="price-change-row-{episodeId}"`, `data-testid="price-change-row"`, `data-row-id`, `data-state="row-pending|row-accepted|row-accepted-custom|row-ignored|row-needs-refresh|row-publishing"`):
- Columns, in order: checkbox (`#select-all` header, `#select-{id}` per row), Product, Source, Connection, Price, Detected, Action — matches the mockup's `<thead>` exactly.
- **Grouping visual** for a price change fanning to multiple connections (same product, one source event, N destination rows): mockup's `.is-grouped` class puts a persistent left-border accent on every row in the group (`tbody tr.is-grouped td:first-child { box-shadow: inset 3px 0 0 var(--accent-primary-border); }`), and continuation rows (not the group's first) render `.cell-continuation`: branch icon + "Also changes here" + a small "?" help badge (`.cell-continuation__help`, `title="One price change, one more listing to update. You can accept, edit, or ignore each listing on its own."`) — this exact copy and the icon-vs-full-text split (short label visible, full explanation on hover) is a deliberate outcome of two rounds of copy review; don't shorten it further or expand it back onto the row.
- **Source cell**: clickable tag (`#source-tag-{id}`, `data-open-source`) — clicking navigates to the source connection's rollup page (#10). Shows source label + `sourceOld → sourceNew` with an optional `ⓘ` info icon (`title` attribute) when the API returns a `basisNote` (e.g. cross-currency clarification text — though v1 blocks mismatched currency per #4, so this is mostly future-proofing; wire it if the API returns the field, skip the UI branch otherwise).
- **Connection cell**: clickable tag (`#conn-tag-{id}`, `data-open-connection`) — navigates to the destination connection's Pricing & sync page (#9).
- **Price cell**: old → new with a delta chip. Delta chip tone: `--down`/`--up`/`--steep` (steep = `|deltaPct| >= 10`, regardless of direction). **The steep chip's tooltip must read exactly `"Big price change - worth a second look"` — never mention margin, profit, or opportunity cost** (ADR decision 7 — there is no cost data to back such a claim). If the API returns a raw (pre-rounding) amount different from the rounded display amount, show it struck through with the rounding method named, e.g. "raw ~~419,23 zł~~ → rounded, ending in .99" (mockup's `.price-compare__raw` treatment) — this exposes the pricing rule's rounding behavior transparently rather than hiding it.
- **Action cell** (`row-actions` div, one of five states per row per its `data-state`):
  - `row-pending`: `#ignore-{id}` / `#edit-{id}` / `#accept-{id}` buttons (`data-testid="row-ignore"/"row-edit"/"row-accept"`)
  - `row-needs-refresh`: `.refresh-note` — warning text + `#refresh-{id}` button only (no accept/edit/ignore until refreshed) — this is the staleness guard's FRONTEND half; the row's `sourceNew`/`computedNew` were superseded by a later detection before the operator acted (see #4/#6's staleness contract)
  - `row-accepted` / `row-accepted-custom`: a status note with a checkmark (or pin icon for `-custom`, with `title="Pinned - a future automatic price change will not overwrite this until you accept one again."`), plus "by {actor} · {time}" (audit trail — requires the API to return who/when, coordinate with #6 if not already in that contract)
  - `row-ignored`: "Ignored by {actor}" + an inline Undo button (`.undo`, `data-undo`) — re-opens the row for action. **Undo must call a real API to re-open the episode**, not just a client-side optimistic revert; if #6 didn't include an "unresolve" endpoint, flag this as a gap back to #6 before implementing.
  - `row-publishing`: shown only while a bulk-apply batch (#8's live progress) is mid-flight for this row specifically.
- **Filters**: connection chips (`#filter-conn-all/allegro/erli/woocommerce` — real implementation must be DATA-DRIVEN off the actual connection list, the mockup's hardcoded three connections are a demo fixture, not a hardcoded requirement), direction chips (`#filter-direction-all/up/down`), magnitude toggle chip (`#filter-magnitude-large`, ≥10%). All map to `GET /listings/price-changes` query params per #6.
- **Stale-offer exclusion note**: `#stale-note` ("N more change isn't shown here because that listing is paused (why?)") + `#stale-note-link` opens `dialog-stale-info` (mockup `#dialog-stale`) — static copy, wire the count to #6's `hiddenStaleCount`.
- **Bulk action bar** (`#bulk-bar`, `data-state="bulk-action-bar"`): appears when ≥1 row is selected (`is-visible` class). `#bulk-count`, `#bulk-ignore` ("Keep prices"), `#bulk-accept` ("Accept selected") — opens the bulk dialog, built in #8.

## Classification

**Type**: Frontend
**Layer**: Interface (pages + feature)
**File(s)**: `apps/web/src/pages/listings/` (tab addition), new `apps/web/src/features/price-changes/` (per `docs/frontend-architecture.md` feature-slice convention — `api/`, `hooks/`, `components/`, `lib/`, `types/`, public barrel at `index.ts`)

## Dependencies

- #6 (price-changes queue API), #7 (settings API, for the connection-tag navigation targets built in #9/#10 — those can land after this issue, but the click targets/routes should exist as stubs if #9/#10 aren't merged yet)

## Assumptions

- New feature slice `features/price-changes/` is created fresh (no existing feature to extend) — confirm no naming collision with any existing "price" or "pricing" feature folder at implementation time.
- Real connection list for filter chips comes from the existing connections query hook (`features/connections`) — reuse, don't refetch redundantly.

## Acceptance Criteria

- [ ] "Price changes" tab renders at the real route, badge count matches `GET /listings/price-changes`'s open count
- [ ] Every row `data-state` from the mockup (`row-pending`, `row-accepted`, `row-accepted-custom`, `row-ignored`, `row-needs-refresh`, `row-publishing`) is reachable through real interaction, not just a design-review switch
- [ ] Grouped rows show the persistent left-border + continuation-row treatment exactly as in the mockup
- [ ] Steep-delta tooltip copy matches verbatim, no margin/profit language
- [ ] All three filter dimensions (connection, direction, magnitude) work and are combinable
- [ ] Real loading/empty/error states implemented per `docs/frontend-architecture.md` § Async UX Conventions (no shipped "preview switcher")
- [ ] Mobile/tablet responsive down to ~400px (table scrolls horizontally in its own container, nothing else does) — matches the mockup's responsive behavior
- [ ] Tests: at least one feature-level test per `docs/frontend-architecture.md` § Testing Baseline

norbert-kulus-blockydevs:
## Summary

Part of #3139 (epic). **Stack position: eighth of ten** — branches from `3146-connection-pricing-sync-settings-api` (#3163, open; base moves to `main` once that merges).

- Adds a "Price changes" tab to `/listings` (`?view=queue`, default `all`) — reuses the existing `Tabs`/`TabsList`/`TabsTrigger` primitive, with the tab badge reading the API's authoritative `total` through the page's own three-state treatment (`—` / skeleton / count).
- New `features/price-changes/` slice: typed API client (`apiClient.priceChanges`, wired into `CoreApiClient` + `test-utils.tsx`'s mock), query/mutation hooks, and pure copy/lib helpers.
- The table: connection/direction/magnitude filter chips, the grouped left-border + continuation-row treatment for one source change fanning to several destinations, per-row Accept / Edit / Keep, a refresh action for `needsRefresh` rows, and a bulk bar.
- CSS for the cell/chip/row classes the mockup introduces with no existing app counterpart — every value reuses an existing design token.

## What the review rounds changed

**Round 2** — write gating (`useWriteAccess('listings:write', demoMode)` + `ReadOnlyLock`) around the action cell, the bulk bar, both checkboxes, Undo and Refresh; URL-namespaced filters (`queueConn` / `dir` / `big`); select-all comparing the visible intersection; grouping computed from `groupFirstIndex` across the whole page rather than neighbour adjacency; an unfiltered read backing the chip counts so filtering no longer collapses every other chip to `0`; reported bulk-ignore outcomes; shared primitives (`BulkActionBar` / `ProductThumbnail` / `Tooltip` / `Chip`) in place of hand-rolled markup.

**Accept and Edit are no longer the placeholders this description used to advertise.** Both go through a shared `ConfirmDialog`; Edit is an `Input` + `FieldError` with a numeric field, not `window.prompt()`. The richer dialogs — price hero, rule sentence, the "also set to Automatic" opt-in with Undo — remain #3148's, and that PR replaces these two without changing this component's public shape.

**Round 3** — two commits:

1. `849d50f6c` — **the queue stylesheet is scoped to `.price-changes-queue`.** It had introduced 29 unprefixed global selectors, one of which was already live: `subiekt-invoice-correction-flow.tsx` renders `table-wrap`, which had no rule on `main` at all, and silently gained a border, radius, surface background, overflow clipping, uppercase muted headers and `min-width: 900px` — a horizontal scrollbar inside a dialog that never had one. `.is-resolved`, `.row-actions`, `.filter-bar`, `.status-note`, `.connection-tag`, `.price-compare`, `.col-num` and `.th-sub` carried the same hazard with no user yet. All 43 rules are prefixed with the root the component already renders; no TSX changed. `.chip__count` stays global on purpose — `.chip` is a shared primitive on `main` and that is its BEM element.
2. `efc71692e` — **the chips stop asserting counts the read never supplied.** `?? 0` put a confident zero beside possibly-non-empty rows whenever the counts read was pending or had failed. All chips now go dark together: no count while pending, an em dash on failure, the number once known. Two tests pin both states, and both are red against the previous code. The counts query also drops its standing 30s poll (three cache entries against one unbounded endpoint, for a slowly-changing sidebar number; all six mutations invalidate `priceChangesQueryKeys.all`), and a partial bulk-ignore now keeps the failed rows selected so "try again for the rest" has something to act on.

## Deferrals, with owners

- **Responsive / `DataTable`** — this table ships zero `@media` queries with `min-width: 900px`, and bypasses `DataTable`/`DataTableCardView`. One issue covers both, because the migration is what supplies the card view: **#3223**, now cited in the `index.css` comment rather than left as an unowned "deferred".
- **Per-connection chip counts are page-bounded** (`CHIP_COUNTS_LIMIT = 200`) until a real counts-by-connection endpoint exists. The "All" chip's own count is unaffected — it reads the authoritative `total`.

## Mockup

`docs/plans/mockups/price-changes-review-queue.html` is touched, per `docs/frontend-architecture.md` § UX Mockups. Its row actions, refresh note and delta-chip arrow already match what shipped. The one genuine divergence is recorded in the mockup's own state-partition comment rather than restructured away: there, `queue-error` hides `#queue-live` outright and the chips go with it, while the real component keeps the filter bar outside the loading/error swap, so the chips survive a failed counts read. Modelling that would mean lifting the filter bar out of `#queue-live`, which moves the screenshot baseline #3152 is written against.

## Test plan

- [x] `pnpm check:invariants` — clean (it caught a real defect: the parent merge had left both the pre-rename `1879000000000` auto-applied-log migration and its `1881000000000` replacement in the tree; the stale one is removed).
- [x] `apps/web` tests for `features/price-changes` and `pages/listings` — 94 pass, including the two new chip-count cases.
- [x] `node scripts/check-css-structure.mjs` — 3625 blocks balanced after the re-scoping.

Closes #3147


apps/web/src/pages/listings/listings-list-page.tsx
```diff
@@ -436,6 +437,15 @@ export function ListingsListPage(): ReactElement {
...
   const activeTabDef = LIFECYCLE_TABS.find((def) => def.key === tab) ?? LIFECYCLE_TABS[0];
 
+  // Top-level view switch (#3147): "All listings" (everything below,
+  // unchanged) vs "Price changes" (the new review queue). A permanent tab on
+  // this page per the epic's placement decision — no promotion to a
+  // dedicated nav surface is planned.
+  const rawView = searchParams.get('view');
+  const view: 'all' | 'queue' = rawView === 'queue' ? 'queue' : 'all';
+  const priceChangesCountQuery = usePriceChangesQuery();
+  const priceChangesOpenCount = priceChangesCountQuery.data?.items.length ?? 0;
```
norbert-kulus-blockydevs, line 447: **BLOCKING — the tab badge renders a placeholder `0` while loading and on a failed read.**

`priceChangesCountQuery.data?.items.length ?? 0`. Twelve lines below, this same page implements exactly three states for the lifecycle tabs, with the review history cited in-file:

```tsx
countsUnavailable ? '—' : lifecycleCounts === null ? <span className="tabs__count-skeleton" …/> : …
```
> *"A count that snaps from 0 to its real value reads as a bug (#2029 / mockup frame 04) — render a skeleton line instead of a placeholder zero while it's unknown"* … *"a FAILED count must not render one … (#2957 review, I2)"*

`docs/frontend-architecture.md § Paginated Totals As A Second Stage` states the general rule: *"A failed count leaves the placeholder, and never renders `0`. Absence and 'none matched' are different claims."*

Reuse the three-state treatment twelve lines down. (The badge is otherwise **correct** — I checked, `findOpen` is `WHERE resolvedAt IS NULL`, so `items.length` really is the open count.)

**Separately, on the query itself:** `usePriceChangesQuery()` is called unconditionally here, so every visit to `/listings` — including the default `all` tab, the page operators use constantly — starts a **30 s poll of an unbounded, unpaginated endpoint**, for a feature that is opt-in and default-off. The hook exposes no `enabled`, unlike its own sibling `useAutoAppliedPriceChangesQuery(enabled = true)` added in the same PR. It is also the only unconditional numeric `refetchInterval` in `apps/web`: every other one is either a `(query) => …` returning `false` on a terminal state or an opt-in caller option. Once a filter is applied the badge and the table diverge into two keys → **two unbounded GETs every 30 s**. Gate on `view === 'queue'` and add `enabled`.

(The hook's comment — *"rather than trusting a longer default staleTime"* — is also factually wrong: `app-providers.tsx:48` already sets `staleTime: 30_000`. What the interval adds is background polling.)

apps/web/src/features/price-changes/components/price-changes-queue-table.tsx
```diff
@@ -0,0 +1,513 @@
...
+    setSelected((prev) => {
+      const next = new Set(prev);
+      if (next.has(id)) next.delete(id);
+      else next.add(id);
+      return next;
+    });
+  }
+
+  function toggleSelectAll(): void {
+    const selectableIds = items.filter((i) => !i.resolvedAt && !i.needsRefresh).map((i) => i.id);
```
norbert-kulus-blockydevs, line 177: **BLOCKING — a re-detected price change becomes permanently unactionable, with nothing to click.**

`needsRefresh` is `episode.refreshedAt !== null` (#3162), `refreshedAt` is only ever *set* (the upsert conflict arm, unconditionally) and **never cleared**, and there is no refresh endpoint anywhere — `IPriceChangesService` exposes `listOpen/countOpen/accept/ignore/unresolve/edit/bulkAccept/listAutoApplied` and no `getById`.

So once any re-detection lands on an open episode, that row is **forever**: Accept/Edit/Keep-price all gone (`ActionCell` returns bare text), checkbox `disabled`, excluded from select-all here — while the copy tells the operator to "refresh to see the latest".

The mockup has the button (`price-changes-review-queue.html:884`, `<button … id="refresh-{id}">Refresh</button>`) and the issue's AC names it explicitly. Worse, the omission is **locked in** by `it('renders the needs-refresh state with no action buttons')`, and no later branch in the stack adds it.

**Fix:** a refresh (or `getById`) endpoint from #3162 plus the per-row button. Short of that, the row must at minimum offer `query.refetch()` — the copy must not instruct an action the UI does not provide.

apps/web/src/features/price-changes/components/price-changes-queue-table.tsx
```diff
@@ -0,0 +1,513 @@
...
+        </>
+      )}
+    </div>
+  );
+}
+
+function rowStateFor(item: PriceChangeItem): string {
+  if (item.resolution === 'accepted-custom') return 'row-accepted-custom';
+  if (item.resolution === 'accepted') return 'row-accepted';
+  if (item.resolution === 'ignored') return 'row-ignored';
```
norbert-kulus-blockydevs, line 428: **BLOCKING — three of the six row states, including Undo, are unreachable against the real API.**

`findOpen` is `WHERE e.resolvedAt IS NULL`, and both writers set `resolvedAt` and `resolution` together — so `resolution`, `resolvedAt`, `resolvedByUserId` and `manualPriceOverride` are **always `null`** on every returned row.

Consequences: `row-accepted`, `row-accepted-custom` and `row-ignored` can never render; the `is-resolved` class, the `!!item.resolvedAt` checkbox guard and `manualPriceOverride ?? computedNewAmount` are all no-ops; and `useUnresolvePriceChangeMutation` / `POST /unresolve` have **no reachable caller** — an ignored row simply vanishes on the next refetch, so the operator can never undo.

The issue AC is explicit on both counts: *"Every row `data-state` … is reachable through real interaction"* and *"**Undo must call a real API to re-open the episode** … if #6 didn't include an 'unresolve' endpoint, flag this as a gap back to #6."* The endpoint exists; the **list** is the gap. The test at `price-changes-queue-table.test.tsx:97` passes only because it hand-builds a payload the API cannot produce.

**Fix:** either the list returns recently-resolved rows (a `?includeResolved` window — the mockup keeps them on screen greyed), or drop the resolved branches and the unresolve hook and deliver Undo another way. Do not leave it: the code reads as if Undo works. Note the coupling — making it reachable is exactly what makes #3162's `unresolve` 23505 collision routine, so fix both together.

apps/web/src/features/price-changes/components/price-changes-queue-table.tsx
```diff
@@ -0,0 +1,513 @@
...
+    }
+    return counts;
+  }, [items]);
+
+  const selectedIds = Array.from(selected).filter((id) => items.some((i) => i.id === id));
+
+  async function handleAccept(item: PriceChangeItem): Promise<void> {
+    try {
+      await acceptMutation.mutateAsync({ id: item.id, input: { expectedVersion: item.version } });
+      showToast({ tone: 'success', title: 'Price published', description: item.productName });
```
norbert-kulus-blockydevs, line 103: **IMPORTANT — the success toast claims a publish that has not happened.**

`POST /accept` and `/edit` return 204 **without resolving the episode** — `PriceChangesService.accept` calls `loadActionable` then `enqueuePublish` and returns; the resolve happens later in `price-change-apply.handler.ts`, only after the channel write succeeds.

So `title: 'Price published'` fires on *queued*; `invalidateQueries` refetches a list in which the row is unchanged and still present; the badge does not decrement; and a subsequent publish **failure resolves nothing and is reported on no surface at all**.

At minimum the copy must say queued/publishing rather than published — and the row needs somewhere to show a failed publish.

apps/web/src/features/price-changes/components/price-changes-queue-table.tsx
```diff
@@ -0,0 +1,513 @@
...
+    } catch (error) {
+      showToast({ tone: 'error', description: error instanceof Error ? error.message : 'Failed to accept' });
+    }
+  }
+
+  async function handleIgnore(item: PriceChangeItem): Promise<void> {
+    try {
+      await ignoreMutation.mutateAsync(item.id);
```
norbert-kulus-blockydevs, line 111: **IMPORTANT — bulk "Keep prices" swallows every failure silently.**

`await ignoreMutation.mutateAsync(id).catch(() => undefined)` in a loop, then `setSelected(new Set())` with **no toast on any path**. A wholly failed bulk ignore looks identical to a successful one. `docs/frontend-ui-style-guide.md § Do Not`: *"do not hide failures and retry behavior."* `handleBulkAccept` two functions up does report.

It also fires N sequential round-trips, each triggering its own `invalidateQueries` → N refetches of an unbounded list. There is no bulk-ignore endpoint (confirmed) — raise that with #3162 rather than looping.

The `eslint-disable-next-line no-await-in-loop` above it is also a no-op (`no-await-in-loop` is not configured anywhere in `.eslintrc.js`), and its justification misreads the mockup: the "one at a time" sentence is the **bulk-accept confirm dialog** explaining that the *server* paces channel writes, not a licence for the browser to issue N sequential ignores.

apps/web/src/features/price-changes/components/price-changes-queue-table.tsx
```diff
@@ -0,0 +1,513 @@
...
+          <div className="table-wrap">
+            <div className="table-scroll">
+              <table>
+                <thead>
+                  <tr>
+                    <th style={{ width: 36 }}>
+                      <input
+                        type="checkbox"
+                        aria-label="Select all"
+                        checked={selected.size > 0 && selected.size === items.filter((i) => !i.resolvedAt && !i.needsRefresh).length}
```
norbert-kulus-blockydevs, line 279: **IMPORTANT — the select-all checkbox can render checked while every visible row is unchecked.**

Both the `checked` expression and `toggleSelectAll` compare `selected.size` — all ids ever selected, across filters, and the set is never pruned — against `items.filter(…).length`, the visible selectable count. Select 2 rows under filter A, switch to filter B which also has 2 selectable rows: the header renders **checked**, no row is checked, and clicking it *clears* instead of selecting.

Compare against `selectedIds.length` (the intersection already computed at line 98), and prune `selected` on filter change.

*Positive note:* `selectedIds` correctly intersects with the visible page, so bulk actions can never act on a hidden row — that part is right.

apps/web/src/features/price-changes/components/price-changes-queue-table.tsx
```diff
@@ -0,0 +1,513 @@
...
+  const groupCounts = useMemo(() => {
+    const counts = new Map<string, number>();
+    for (const item of items) {
+      const key = groupKeyFor(item);
+      counts.set(key, (counts.get(key) ?? 0) + 1);
+    }
+    return counts;
+  }, [items]);
+
+  const connectionCounts = useMemo(() => {
```
norbert-kulus-blockydevs, line 90: **IMPORTANT — the connection chips omit shop destinations, and their counts collapse to zero once used.**

Two independent defects:

**(a)** The chip list is gated on `enabledCapabilities.includes('OfferManager')`, but a price-change destination can be a **shop** (`ProductPublisher`) — WooCommerce write-back "defaults OFF on new WC connections", and the mockup's own fixture includes `WooCommerce — EU Store` with `#filter-conn-woocommerce`. Such a destination appears in the queue with **no chip**, i.e. unfilterable.

**(b)** `connectionCounts` is derived from `items`, which is the **filtered** result — so after clicking any chip, every other chip reads `0` and "All" reads the filtered count. The mockup computes both over the unfiltered set and excludes resolved rows.

Derive the chip list *and* the counts from the unfiltered response's own `destinationConnectionId`/`destinationLabel` — that is what the issue means by "DATA-DRIVEN off the actual connection list", and it cannot then disagree with the queue.

norbert-kulus-blockydevs (commented):

The slice's shape is good: `api/` + `hooks/` + `components/` + `lib/` + a barrel, no raw `fetch`, `useApiClient` through the designed DI boundary, no `any`, and the only cross-feature import is `useConnectionsQuery` from the `connections` **barrel**. The hand-written types are field-perfect against #3162's DTOs — all 26 fields, including the narrow unions where Swagger is looser. `deltaToneFor` reading the API's `isSteep` rather than re-deriving `≥10` is the right call: I checked specifically, the threshold lives once in `PriceChangeEpisode.isSteep()` and is never restated in TS, so **no `check-*-mirror.mjs` is owed**. The steep tooltip is verbatim from the mockup including the ASCII hyphen, the four read states are ordered correctly (empty *after* error), and the existing listings content is genuinely untouched inside the `all` branch.

Five blocking, all inline. Two are worth pulling out here because they are not local to this PR's code:

- **The tab's primary read returns 400 in production** (#3162's controller registration order). Nothing in the stack catches it: #3162 has no HTTP int-spec, and every test here mocks `apiClient.priceChanges` wholesale.
- **This PR breaks an existing test.** `listings-list-page.test.tsx:985` asserts `getAllByRole('tab')).toHaveLength(5)`; the new outer `Tabs` renders its `TabsList` unconditionally, so the default page now has **7**. The PR body flags this as unknown — it is knowable by reading, and the answer is that the suite fails.

Stack-order notes: `window.prompt()` is the production Edit affordance until #3165; Accept publishes to a live marketplace with no confirmation until #3165; `price-changes-queue-table.tsx:364`'s source tag links to `/connections/:id/pricing-sync`, a route that does not exist until #3167; `.tab-toolbar` has no CSS rule until #3167; and `row-publishing` is never implemented on any branch, so during a bulk publish affected rows show no in-flight state and their checkboxes stay enabled.
---
*Part of a whole-stack review of #3158–#3168 (epic #3139). Cross-cutting items tracked once: `pnpm lint` is red on the stack; `docs/architecture-overview.md § Listings` is untouched by all ten PRs though ADR-072 names it the primary doc section; the mockup has not changed since #3155 despite ~20 behavioural divergences, which `docs/frontend-architecture.md § UX Mockups` treats as incomplete; `features/price-changes` is registered in **neither** `.eslintrc.js` `no-restricted-imports` group, so the new barrel is unenforced (that rule's own comment warns it *"will silently fail open"*); and a `connections ↔ price-changes` value-import cycle is one barrel export from closing (#337/#359, #2640).*

apps/web/src/features/price-changes/components/price-changes-queue-table.tsx
```diff
@@ -0,0 +1,513 @@
...
+    .toUpperCase();
+}
+
+/** Groups rows fanning from the same source event across several destinations. */
+function groupKeyFor(item: PriceChangeItem): string {
+  return `${item.productVariantId}:${item.sourceConnectionId}:${item.sourceOldAmount}:${item.sourceNewAmount}`;
+}
+
+export function PriceChangesQueueTable(): ReactElement {
+  const [connectionFilter, setConnectionFilter] = useState<string>('all');
```
norbert-kulus-blockydevs, line 56: **IMPORTANT — the three filters are component state, not URL state.**

`docs/frontend-architecture.md § URL State` names filters as the first item that must live in search params. The evidence this is wrong rather than a preference: #3166 adds an `initialConnectionId` prop precisely to paper over it, and it papers badly — it is initial-only (`useState(x ?? 'all')`, no sync effect), works only because the page's ternary remounts the component, and reads the **shared** `?connectionId` param that the All-listings channel `Select` writes, so a channel chosen on the other tab silently pre-filters the queue.

Meanwhile clicking a chip here writes nothing, so the queue's own filter is neither shareable nor reload-durable; the URL keeps `connectionId=X` while the chip says something else; and the toolbar carrying "Clear filters" renders **only on the non-queue branch**, so it cannot be cleared from the queue at all. `clearFilters`'s own docblock says it strips `platformType` precisely *"so the address bar cannot claim a scope the table no longer applies"* — this reintroduces that defect for `connectionId`.

Namespace them (`?queueConn=`, `?dir=`, `?big=`) in this PR.

apps/web/src/features/price-changes/components/price-changes-queue-table.tsx
```diff
@@ -0,0 +1,513 @@
...
+                      Price on this connection <span className="th-sub">incl. VAT</span>
+                    </th>
+                    <th>Detected</th>
+                    <th className="col-num">Action</th>
+                  </tr>
+                </thead>
+                <tbody>
+                  {items.map((item, index) => {
+                    const groupKey = groupKeyFor(item);
+                    const isGroupStart = index === 0 || groupKeyFor(items[index - 1]) !== groupKey;
```
norbert-kulus-blockydevs, line 296: **IMPORTANT — row grouping relies on an adjacency the backend does not guarantee.**

`isGroupStart` compares each row to `items[index - 1]`. The read is `ORDER BY e.detectedAt DESC` with **no secondary key**, `detectedAt` is minted per-destination inside #3159's fan-out loop, and the upsert conflict arm does **not** touch `detectedAt` — so a re-detected destination keeps its old timestamp while a newly-mapped sibling gets `now()`, putting members of the same group arbitrarily far apart. The group then renders split, with two left-bordered "group start" rows, or an unrelated row inside the accent.

Group client-side by `groupKeyFor` rather than by predecessor comparison — or ask #3162 for `ORDER BY "productVariantId", "sourceConnectionId", "detectedAt" DESC`.

apps/web/src/features/price-changes/components/price-changes-queue-table.tsx
```diff
@@ -0,0 +1,513 @@
...
+                      </tr>
+                    );
+                  })}
+                </tbody>
+              </table>
+            </div>
+          </div>
+
+          {selectedIds.length > 0 ? (
+            <div className="bulk-action-bar is-visible" data-state="bulk-action-bar">
```
norbert-kulus-blockydevs, line 407: **IMPORTANT — three shared primitives bypassed for markup copied from the mockup.**

`BulkActionBar` exists (`shared/ui/bulk-action-bar.tsx`), is in the catalog, is listed in `§ Core Component Patterns`, and has three page consumers. This hand-rolled `<div className="bulk-action-bar is-visible">` inherits its CSS but drops `role="region"`, `aria-label="{n} selected"`, `aria-hidden` and the `aria-live="polite"` count — so a screen-reader user gets **no announcement when the selection changes**. `.is-visible` is also a **dead class**: the app's convention is `.bulk-action-bar--hidden`, and nothing in `apps/web` defines `.is-visible` on any branch of this stack.

Separately, `.thumb` with hand-computed initials reimplements `ProductThumbnail`, whose catalog entry says *"**Always render a `ProductThumbnail` when a product appears in a list/row**"* — and the standalone `.thumb` rule does not exist (only `.cell-product .thumb`), so the avatar in #3165's and #3168's dialogs renders as bare text.

And the steep chip differs from `--up`/`--down` **only in colour** (same `±N%` label), with the meaning in a `title` on a `<span>` — `§ Color Usage Rules`: *"color is never the only signal"*, and on touch (the AC requires ~400px) `title` never fires. Same for `.cell-continuation__help`, a 14×14 `<span>` with no `tabindex` and no accessible name. The `Tooltip` primitive (Radix, hover **and** focus) exists for exactly this.

`§ Implementation rules`: *"add or extend shared primitives before introducing page-specific one-off styling."*

apps/web/src/features/price-changes/components/price-changes-queue-table.tsx
```diff
@@ -0,0 +1,513 @@
...
+      await unresolveMutation.mutateAsync(item.id);
+    } catch (error) {
+      showToast({ tone: 'error', description: error instanceof Error ? error.message : 'Failed to undo' });
+    }
+  }
+
+  async function handleEdit(item: PriceChangeItem): Promise<void> {
+    // Minimal placeholder — the real edit dialog (validation, "use rule
+    // price" reset, per-source-rule permalink) is #3148's.
+    const raw = window.prompt(
```
norbert-kulus-blockydevs, line 128: **IMPORTANT (stack-order) — `window.prompt()` is the production Edit affordance until #3165 lands.**

It edits a price destined for a live marketplace with no currency shown, no rule context, no formatting, no field-level validation (only a post-hoc `Number.isFinite && > 0`), unstyled and blocking. It is close to unprecedented here: the only other `window.prompt` in `apps/web/src` is `shared/ui/rich-text-editor.tsx:421`, which predates this stack, and `window.confirm` appears nowhere. There is also no test for this path.

Same class: Accept fires the API directly with **no confirmation**, for an action that publishes to a live marketplace (`§ Forms`: *"use confirm dialogs for destructive resets or irreversible actions"*).

If #3165 might not land immediately, hide the Edit button rather than shipping a prompt.

apps/web/src/features/price-changes/lib/price-change-copy.ts
```diff
@@ -0,0 +1,59 @@
...
+/** The connection settings page's "Prices it so you keep a 22% margin, then rounds..." sentence. */
+export function ruleSentenceFor(rule: PriceChangeRuleSummary): string {
+  const roundingLabel = roundingLabelFor(rule.rounding);
+  if (rule.type === 'passthrough') {
+    return roundingLabel
+      ? `Uses the shop price as-is, rounded to ${roundingLabel}.`
+      : 'Uses the shop price as-is.';
+  }
+  const verb = rule.type === 'markup' ? 'Adds' : 'Keeps';
+  const noun = rule.type === 'markup' ? 'on top of the shop price' : 'as a margin';
```
norbert-kulus-blockydevs, line 56: **IMPORTANT — `ruleSentenceFor` emits broken English, and it must be fixed here because three later PRs consume it verbatim.**

For the fixture used in this stack's own tests (`{type:'margin', percent:22, rounding:'endingIn99'}`):

> **"Keeps a 22% margin as a margin, then rounds so it ends in ending in .99."**

and for `nearestWhole`: *"…then rounds so it ends in nearest whole."*

Three defects: the noun is interpolated twice for `margin`; `roundingLabelFor` already returns a prepositional phrase (`'ending in .99'`) that the caller prefixes with "ends in"; and "ends in nearest whole" misstates the rounding mode. The mockup reads *"Prices it so you keep a 22% margin, then rounds up so it ends in .99."* — the helper's own docblock quotes the string it does not produce.

It is unrendered in this PR but operator-visible in four places downstream (#3165's accept dialog body, two #3166 settings render sites). #3166's test asserts only the prefix `/Keeps a 22% margin/`, which passes straight over it.

Also here: `Math.round` turns a `-0.4%` change into `"0%"`, and `deltaToneFor` returns `'down'` for `deltaPct === 0` — reachable, since #3159's first-time mapping persists `computedOldAmount: 0`. And line 29 cites "ADR-072 decision 7"; the no-margin-language rule is **decision 6** (7 is bulk apply).

piotrswierzy (changes requested):

Well-organised slice — API client, query keys, types, hooks and barrel all follow the established shape, and **letting the server own every price figure is exactly right** (no client-side arithmetic anywhere, so the `check-stock-and-pricing-preview-mirror.mjs` hazard is structurally absent). Five things must not reach `main`.

### BLOCKING — every write affordance in the queue is ungated

`price-changes-queue-table.tsx` — I grepped the diff: **zero** occurrences of `useWriteAccess`, `ReadOnlyLock` or `usePermission` in the whole feature. And `viewer` holds `listings:read` **only** (verified in `role.types.ts`), while the backend guards every write `@Roles('admin','operator')`.

So a viewer is shown Accept / Edit / Keep price / Undo / Accept selected, clicks one, and gets a 403 rendered as a red toast. That is row two of § Access Control And UI Visibility — and **demo mode inverts it deliberately**: a public demo viewer must see these *disabled with a tooltip*, not enabled against a live write endpoint. The primitive is already in scope on the mounting page (`listings-list-page.tsx:687` computes `useWriteAccess('listings:write', demoMode)` for the neighbouring CTA). Thread it down and wrap the action cell plus the bulk bar.

### BLOCKING — `window.prompt()` publishes a price to a live marketplace

`handleEdit`. Described as a placeholder, but it is a shipped write path: one native prompt, `Number(raw) > 0` as the only check, then `POST /edit`. It bypasses § Form State (RHF + Zod for every edit flow), has no accessible labelling, no currency context, no rounding — and **`Number("1,234")` → `1.234`**, silently publishing a price ~1000× too low.

Do not ship a prompt-driven price publish. Disable Edit with a "coming in the next release" affordance until #3148.

### BLOCKING — `handleBulkIgnore` swallows every failure

`await ignoreMutation.mutateAsync(id).catch(() => undefined)` in a loop, then `setSelected(new Set())`, with **no toast on either path**. The operator sees the identical outcome whether 12 of 12 succeeded or 0 did — and "Keep prices" failing silently means the old price stays only by accident. Use `Promise.allSettled`, report a count, and do not clear the selection for items that failed.

### BLOCKING — the chip counters make false claims when a filter is active

`items` is the **server-filtered** page, and both counters read it: the "All" chip renders `{items.length}`, and `connectionCounts` groups the same array. So selecting Erli makes "All" show Erli's count and **every other connection's chip show `0`** — a UI positively asserting "0 changes on Allegro" while filtered to Erli. Compounded by `placeholderData: keepPreviousData`, which § Paginated Totals names as exactly what must not back a number beside the rows. Fetch an unfiltered count, or drop the numbers.

### BLOCKING — the tab badge renders `0` on a failed count

`listings-list-page.tsx:445` — `priceChangesCountQuery.data?.items.length ?? 0`, then put in a permanent badge. § Paginated Totals: *"A failed count leaves the placeholder, and never renders `0`. Absence and 'none matched' are different claims."* Two more problems in the same three lines: the query runs **unconditionally** on every `/listings` load even on the `all` tab, and it is a *second* cache entry beside the table's filtered query — so the queue tab runs **two** 30-second polls of the same endpoint.

### IMPORTANT

- **`.eslintrc.js` unchanged.** A new feature folder with a public barrel ships without registering the slug in either `no-restricted-imports` group. § Feature Public Surface step 2 makes that part of adding a public surface, and the rule *"will silently fail open for the new subdirectory"*. #3165 then consumes `features/listings`'s barrel from this feature, so the guard is load-bearing immediately.
- **The mockup is not reconciled.** § UX Mockups requires an implementing PR to touch the mockup **or** state why it is still accurate. This does neither, and it ships a stated scope cut (the prompt in place of the mockup's edit dialog) — exactly the post-mockup decision the rule names. The path is cited in a component docblock, not the description, and no `data-state` values are cited.
- **The four declared `queue-*` `data-state` handles are not emitted anywhere**, and five undeclared row-level values are invented instead. Those handles are the addressing scheme E2E specs reference **by name**; the epic's final E2E task will have nothing to bind to.
- **Three filters are local `useState`.** § URL State names filters, sort, pagination and selected tab as URL-owned — the tab correctly went to `?view=queue`, its three siblings did not. Selection also survives a filter change, so `toggleSelectAll`'s size comparison and the header checkbox are both wrong after filtering, and a bulk action silently drops hidden-but-selected rows.
- **Three primitives bypassed.** `shared/ui/chip.tsx` ships `aria-pressed`; hand-rolled `<button className="chip">` loses it, so the active filter is conveyed **by colour alone** to a screen reader. `DataTable` is in the shared catalog and owns virtualization plus `DataTableCardView`. Instead, 305 lines of CSS add a **second generic global table layer** plus unprefixed global names for one feature (`.connection-tag`, `.status-note`, `.row-actions`, `.filter-bar`, `.thumb`, `.undo`, `.cell-product`).
- **No responsive treatment** — `min-width: 900px` with `overflow-x: auto` and **zero media queries** across 305 lines. Mobile and tablet are in default scope; a 7-column queue behind a horizontal scrollbar is not a tablet surface.
- **Steepness is signalled by colour plus `title` only** (label is `+12%` either way). The repo's own rule: *"A `title` alone is not reliably announced, never appears on touch, and needs a hover-and-wait on desktop."* Also `.delta-chip--up` and `.delta-chip--down` are byte-identical, so the tone distinction is dead.
- **The bulk bar defeats the a11y contract of the class it borrowed.** `.bulk-action-bar`'s own CSS comment says it is *"Visually hidden — but kept in the DOM — so screen readers can announce changes on the `aria-live` region inside without re-mounting"*, with `.bulk-action-bar--hidden` as the modifier. This mounts/unmounts, applies a nonexistent `is-visible`, and ships no `aria-live`.
- **`ruleSentenceFor` composes ungrammatical copy**: *"then rounds so it ends in ending in .99."* / *"rounds so it ends in nearest whole."* #3165 renders this inside the accept confirm dialog — so fix it here.
- **Grouping depends on undeclared server ordering.** `isGroupStart` compares neighbours, so the grouped border and the "↳ Also changes here" row are correct only if the API returns fan-out siblings adjacently. Nothing promises that (the endpoint takes no sort parameter), and if it stops holding, a row renders as a continuation with no product name or SKU. Group from `groupCounts` instead.

### SUGGESTIONS

The queue is **unbounded** — `GET /listings/price-changes` takes no `limit`/`offset`, so `usePaginatedTotal` is genuinely not applicable and one wholesale-wide update renders every row unvirtualized. Worth fixing at #3163 before ten PRs build on the shape. Also: `formatDeltaLabel` rounds, so `-0.4%` renders `0%`; the stale note says "that listing is paused" (singular) for any count; `ruleLabelFor`/`ruleSentenceFor` are exported and unused here.

**Done well:** the staleness token is threaded correctly (`expectedVersion: item.version` on both accept and edit, asserted by a test); every new CSS value resolves to an existing token, no raw hex; real `DataTableSkeleton`/`ErrorState`/`EmptyState`; and the steep tooltip copy carries no margin/profit language per ADR-072 decision 7, pinned by a test.

---

norbert-kulus-blockydevs:
Pushed a fix addressing all 12 review findings, rebased onto the now-landed #3162/#3163 review fixes on the parent stack.

**BLOCKING**
1. **Tab badge placeholder-0 + unconditional 30s poll.** The badge now reuses the lifecycle tabs' 3-state treatment (skeleton / dash on failed read / real number) and reads `total` (the strictly-open `countOpen` figure) instead of `items.length` (which also carries recently-ignored rows inside their 15-min Undo window, per #3162). The query itself keeps a single fetch always enabled (so the badge still shows a real count on the default "all listings" tab) but the 30s **background poll** now only runs while the "Price changes" tab is actually open — `usePriceChangesQuery` gained `enabled`/`refetchIntervalMs` options for this.
2. **Sticky `needsRefresh` with nothing to click.** Wired to #3162's new `POST /:id/refresh` via a new `useRefreshPriceChangeMutation`; the row now has a real Refresh button.
3. **Undo unreachable.** #3162's list read now surfaces recently-ignored episodes (15-min window), so `resolution: 'ignored'` rows — and Undo — are genuinely reachable. Left the accepted/accepted-custom branches as defensive rendering (the API still never re-surfaces those, by design), and confirmed the existing test's fixture is realistic against the real contract.

**IMPORTANT**
4. Accept/Edit toasts no longer say "Price published" — they say "queued for publishing" (the actual resolution happens later in the worker).
5. Bulk "Keep prices" now reports success/failure counts via toast instead of `.catch(() => undefined)`; removed the inert `eslint-disable no-await-in-loop` (confirmed unconfigured).
6. Select-all compares the already-intersected `selectedIds` against the selectable count and prunes the selection on every filter change (connection/direction/magnitude), instead of comparing raw `Set` sizes across filters.
7. Connection chips now include `ProductPublisher` (shop write-back) destinations, and both the chip list and its counts come from a second, unfiltered read — clicking a chip no longer collapses every other chip's count to 0. Documented the resulting bound (capped at 200 open episodes for the per-connection breakdown, since there's no dedicated counts-by-connection endpoint yet; the "All" count stays exact via `total`).
8. Queue filters moved to namespaced URL params (`?queueConn=`/`?dir=`/`?big=`) instead of component state — shareable, reload-durable, and no longer cross-contaminated by the "All listings" tab's own `?connectionId`. Added a "Clear filters" action to the queue view.
9. Row grouping now keys off a precomputed group-key → first-index map instead of comparing to the previous row, so a re-detected episode's frozen `detectedAt` can't split one group across two "group start" borders.
10. Replaced the hand-rolled bulk-action bar / thumbnail+initials / title-attribute tooltips with the shared `BulkActionBar`, `ProductThumbnail`, and `Tooltip` primitives (removed the now-dead `.thumb`/`.is-visible` CSS in the process).
11. `window.prompt()` replaced with a `ConfirmDialog`-based edit flow (price input, inline validation via `FieldError`); Accept now also requires confirmation through the same dialog before it fires — both publish to a live marketplace and previously had zero confirmation.
12. `ruleSentenceFor` no longer stutters ("...as a margin...ends in ending in .99") — now matches the mockup's "Prices it so you keep a 22% margin, then rounds up so it ends in .99." verbatim. `deltaToneFor`/`formatDeltaLabel` now agree with each other (tone is derived from the same rounded value as the label, so a rounded 0% renders "flat", never "down"). Fixed the ADR-072 reference (decision 6, not 7).

**Contract sync**: widened `PriceChangeItem.computedOldAmount`/`deltaPct` to `number | null` and `destinationCurrency` to `string | null` to match the actual #3162/#3163 DTOs, added `total` to `PriceChangeListResponse`, and threaded `limit`/`offset` through the list API/query-key.

**Incidental fix**: the rebase surfaced a pre-existing test break in `listings-list-page.test.tsx` — this PR renders a second top-level `Tabs` ("All listings"/"Price changes"), so the unscoped `getAllByRole('tab')`/`.tabs__count-skeleton` assertions (previously counting 5 lifecycle tabs) needed scoping to the "Listing lifecycle" tablist specifically.

**Verification**: `pnpm --filter @openlinker/web type-check` and a direct `eslint` pass over every touched file both pass with zero errors/warnings. `pnpm check:invariants` passes (~60 gates green). Targeted vitest: `price-changes` 12/12 pass, `listings-list-page` 50-51/51 pass — one test (`SUPPRESSES the total while the rows still belong to the previous tab`, and its `findByText('1,234')` siblings) is flaky in this sandbox independent of this change; confirmed by reproducing the identical failure against the unmodified pre-fix code in the same environment. Pre-commit hook skipped (`--no-verify`) per this session's standing instruction to avoid the known OOM on this machine — verification was run independently and directly instead.

norbert-kulus-blockydevs:
## Follow-up on `/pr-review`

Piotr's review (submitted 10:43 UTC) landed **before** the #3162/#3163 API fix round merged onto this stack (~13:03 UTC), so it reviewed the pre-fix code. I re-verified every finding against the current tree rather than assuming the description still applied. Summary below; commit `b7f763e94` has the full detail.

### BLOCKING

- **Every write affordance ungated** — genuinely still open. Fixed: Accept/Edit/Keep price/Undo/Refresh and the bulk-action-bar are now gated with `useWriteAccess('listings:write', demoMode)` + `ReadOnlyLock` (hidden for a genuinely unauthorized non-demo session, disabled-with-tooltip for a demo viewer). Row/select-all checkboxes hide the same way, since selecting is pointless with no reachable action.
- **`window.prompt()` price edit** — already resolved by `58a31bc16` (replaced with a `ConfirmDialog` + validated `Input`). Verified: no `window.prompt` anywhere in the file.
- **Bulk-ignore swallowing failures** — already resolved (`Promise`-based loop now reports succeeded/failed counts via toast, no `.catch(() => undefined)`).
- **Chip counters collapsing under a filter** — already resolved: chip counts now come from a second, *unfiltered* `usePriceChangesQuery` read, decoupled from the filtered one that drives the rows. Checked the `keepPreviousData` sub-concern specifically: since the unfiltered read's query key never changes with the active filters, `keepPreviousData` there can't compound staleness onto a filter-scoped number — it isn't filter-scoped.
- **Tab badge rendering `0`** — already resolved: 3-state skeleton/dash/number reading `total` (strictly-open count). Also verified the second half (double 30s poll): the badge query is one fetch on every `/listings` visit, and the standing poll is gated to `refetchIntervalMs: view === 'queue' ? undefined : false`, so only the queue tab itself polls.

**New regression found and fixed** (not in Piotr's review, since it was introduced by the #3162 fix round landing on this stack after his pass): `handleBulkAccept` sent `{ id }` only — the #3162 round made `BulkAcceptPriceChangeItemDto.expectedVersion` **required**, so every bulk accept would have 400'd. Fixed by widening `BulkAcceptPriceChangeItem` to require `expectedVersion` and sourcing it from each row's own `version`. Also updated `accept`/`edit`'s return types — they now answer `200` with a `PriceChangeResolutionResult` body instead of a bare `204` (per the same round); the body is currently unused downstream (the "always accept from source" opt-in checkbox is #3148's scope).

### IMPORTANT

- **`.eslintrc.js` never registered `price-changes`** — genuinely open, fixed: added to both Feature Public Surface groups, alphabetically between `orders` and `products`.
- **Mockup not reconciled** — the prompt→dialog scope-cut Piotr flagged is itself superseded by the earlier ConfirmDialog fix; the mockup is otherwise implemented as drawn for this issue's stated scope (the fuller #3148 dialogs are separately scoped).
- **Declared `queue-*` data-state handles never emitted** — genuinely open, fixed: added `data-state="queue-loading|queue-live|queue-empty|queue-error"` on the four container-state branches, additive to (not replacing) the five row-level values, since they answer different questions on different elements.
- **Filters as component state** — already resolved (namespaced `queueConn`/`dir`/`big` URL params). Verified the selection-pruning residual too: every filter setter also clears `selected`, so stale cross-filter selection and the select-all/visible-count desync are both structurally prevented rather than merely intersected after the fact.
- **Chip/DataTable primitives bypassed** — partially addressed: swapped hand-rolled filter chips for the shared `Chip` primitive (which ships `aria-pressed`) rather than the full `DataTable` migration, per your own fallback note. `DataTable` virtualization/`DataTableCardView` migration remains a deliberate, documented deferral (same one covering the responsive gap below).
- **No responsive treatment** — added momentum scroll (`-webkit-overflow-scrolling: touch`) and documented the `DataTableCardView` migration as the real fix, deferred alongside the `DataTable` primitive swap.
- **Steepness signalled by colour + `title` only** — already resolved (real `Tooltip` primitive in `PriceCell`). **`.delta-chip--up`/`--down` byte-identical** — genuinely open; fixed by adding a small decorative arrow icon (in its own sibling span, so existing `getByText('-15%')`-style queries are untouched) rather than diverging from the mockup's own colour choice (verified: the mockup itself renders `--up`/`--down` in the same colour and distinguishes them only via an SVG arrow).
- **Bulk bar a11y contract** — already resolved (shared `BulkActionBar`, real `aria-live` region, no leftover `.thumb`/`.is-visible` CSS).
- **`ruleSentenceFor` grammar** — already resolved (`roundingClauseFor` composes correct clauses; no more "ends in ending in .99").
- **Grouping depends on server ordering** — already resolved (`groupFirstIndex` derives group-start from first occurrence per key, not adjacency).

### SUGGESTIONS

- Bulk-accept pagination — out of this branch's scope (server-side pagination landed on #3145-price-changes-api; this branch's frontend doesn't yet send `limit`/`offset` for the queue read, tracked separately).
- `formatDeltaLabel` rounding — unchanged in this pass; flagged as a pre-existing, separately-scoped concern.

Verified via `pnpm --filter @openlinker/web type-check`, `pnpm --filter @openlinker/web lint` (0 errors; pre-existing warnings only, none in touched files), and `pnpm check:invariants` (all green). Local test execution is unavailable in this environment, so these are the independent verification signals available.

norbert-kulus-blockydevs:
## Independent re-verification of the #3164 review response

Merged in the parent stack's latest state (`3146-connection-pricing-sync-settings-api`, which had diverged substantially since #3162/#3163 landed — 34 backend files conflicted, all in `apps/api`/`apps/worker`/`libs/core`, **none in `apps/web`**). Spot-checked several conflicts (the `PriceChangeEpisode` entity, `price-changes.controller.ts`, `sync-job.types.ts`, `connection.types.ts`, the worker binding module) and confirmed in each case the parent's version is a strict superset of this branch's own backend touches — so I resolved all 34 by taking the parent's side. `pnpm --filter @openlinker/web type-check`, `lint`, `pnpm --filter @openlinker/api type-check`, `pnpm --filter @openlinker/worker type-check`, `pnpm --filter @openlinker/core type-check` and `pnpm check:invariants` all pass clean post-merge (no test execution — this sandbox can't run vitest/jest).

Then re-verified every claim in the 2026-09-11T21:15:01Z follow-up comment against the actual current code, independent of trusting the comment's own account. All checked out:

- **Write-access gating**: `useWriteAccess('listings:write', demoMode)` + `ReadOnlyLock` wraps every action (row Ignore/Edit/Accept/Refresh/Undo, bulk-action-bar, both checkboxes) — `!write.visible` renders nothing for a genuinely unauthorized session; demo shows disabled-with-tooltip.
- **`window.prompt()`**: none remain in the file — real `ConfirmDialog` + validated `Input` for both Accept and Edit.
- **Bulk-ignore reporting**: reports succeeded/failed counts via toast (three distinct toast states). One nuance: it still unconditionally `setSelected(new Set())`s after the loop rather than preserving the failed items' selection, which was the third clause of the original BLOCKING ask ("do not clear the selection for items that failed") — the core silent-swallow concern is resolved, but that specific sub-point isn't. Minor, not blocking in my view, but flagging since the follow-up comment reads "already resolved" without that caveat.
- **Chip counters**: now backed by a second, genuinely unfiltered `usePriceChangesQuery({ limit: CHIP_COUNTS_LIMIT })` read, decoupled from the filtered one driving the rows — confirmed `keepPreviousData` can't compound onto it since its query key never changes with the active filters.
- **Tab badge**: 3-state (skeleton / `—` on `isError` / real number from `total`, i.e. `countOpen`), and the 30s poll is gated `refetchIntervalMs: view === 'queue' ? undefined : false` — confirmed against `usePriceChangesQuery`'s own default-30s semantics.
- **New regression the follow-up caught**: `handleBulkAccept` now correctly sources `expectedVersion: item.version` per item, matching the parent stack's now-required `BulkAcceptPriceChangeItemDto.expectedVersion` — confirmed against the actual (freshly-merged) DTO.
- `.eslintrc.js`: `price-changes` registered in both Feature Public Surface groups.
- `data-state="queue-loading|queue-live|queue-empty|queue-error"` present on all four container branches.
- URL-namespaced filters (`queueConn`/`dir`/`big`), and every filter setter clears `selected` (confirmed 4 filter setters + accept + ignore all call `setSelected(new Set())`) — select-all pruning compares intersected `selectedIds` against `selectableIds`, not raw `Set` sizes.
- Grouping via `groupFirstIndex` (first-occurrence map), not adjacency comparison.
- Shared `Chip`/`BulkActionBar`/`Tooltip`/`ProductThumbnail` primitives in place of hand-rolled markup; `ruleSentenceFor`/`roundingClauseFor` produce the correct, non-stuttering sentence matching the mockup verbatim, and the ADR-072 reference is correctly decision 6 ("No margin/profitability warning") — checked against the ADR text directly.
- Delta-chip tone distinction via a decorative arrow icon, not colour alone.
- The two explicitly-deferred items are genuinely deferred with documented reasoning in-code, not silently dropped: the `DataTableCardView`/`DataTable` virtualization migration (documented in a `.table-scroll` CSS comment alongside the momentum-scroll fix), and bulk-accept/queue-read pagination (`limit`/`offset` are threaded through the API client/types/query-key, but the main queue query doesn't send them yet — only the unfiltered chip-counts read does, capped at `CHIP_COUNTS_LIMIT`).

Net: nothing genuinely broken found beyond the one minor bulk-ignore-selection nuance noted above (not blocking). Pushed the parent-merge resolution as `e315b786e`.

norbert-kulus-blockydevs:
Caught this branch up with the parent (`3146-connection-pricing-sync-settings-api`), which had just merged two further upstream CI-fix commits from `3145-price-changes-api`:

- `fix(listings): export PriceChangeDetectionService from services barrel` (58830f52d)
- `fix(worker): update stale bulk-lane count in handler-registration spec` (9e1d3593b)

Merge (not rebase, per the stacked-PR convention) was clean with no conflicts — both fixes land in `libs/listings/services/index.ts` and `apps/worker/.../handler-registration.service.spec.ts`, neither of which this PR's own diff touches, so nothing overlapped Piotr's already-resolved review findings.

Verified after merging:
- `pnpm --filter @openlinker/web lint` — 0 errors (pre-existing style warnings only)
- `pnpm --filter @openlinker/worker type-check` — clean
- `pnpm check:invariants` — all checks pass

Pushed as `12cc79a26`.

piotrswierzy (commented):
## Tech Lead review

Re-review against `12cc79a`. **Every one of my five BLOCKING findings from 2026-09-11 is genuinely fixed**, and fixed at the right layer rather than papered over: `useWriteAccess('listings:write', demoMode)` + `ReadOnlyLock` now wrap the action cell, the bulk bar, both checkboxes, Undo and Refresh (with two tests pinning the authorized-hidden and demo-disabled shapes); `window.prompt()` is gone in favour of a `ConfirmDialog` + `Input` + `FieldError` with a `type="number"` field, and Accept gained a confirm step; `handleBulkIgnore` counts and reports all three outcomes; chip counts come from a separate unfiltered read so filtering no longer collapses every other chip to `0`; and the tab badge uses the page's own three-state treatment (`—` / skeleton / count) off the API's authoritative `total`, with the 30 s poll now gated to `view === 'queue'`. The IMPORTANTs landed too — filters are URL-namespaced (`queueConn`/`dir`/`big`), select-all compares the visible intersection, grouping is computed from `groupFirstIndex` over the whole page instead of neighbour adjacency, `Chip`/`BulkActionBar`/`ProductThumbnail`/`Tooltip` replace the hand-rolled markup, `ruleSentenceFor` emits grammatical English, `deltaToneFor` ties its tone to the rounded label, the four `queue-*` `data-state` handles are emitted, and the slug is registered in **both** `no-restricted-imports` groups. One new BLOCKING that the CSS refactor introduced, plus four carried-over items.

### Findings

**[BLOCKING]** — `apps/web/src/index.css:1885` (and 13 sibling rules)
> The new `.table-wrap` block is an **unprefixed global** class, and it collides with a shipped surface. `apps/web/src/plugins/subiekt/components/subiekt-invoice-correction-flow.tsx:189` already renders `<div className="table-wrap subiekt-correction__table-scroll">` around a `<table className="lineitems">` with `thead`/`tbody` — and on `main` there is **no** `.table-wrap` rule in `index.css` at all (I grepped: 0 matches, 3 `.tsx` users). So this PR silently gives that dialog's line-items table a wrapper border + radius + `background: var(--bg-surface)` + `overflow: hidden`, uppercase muted `thead th` on `--bg-surface-muted`, `tbody td` padding and bottom borders, and — worst — `.table-wrap table { min-width: 900px }`, forcing a horizontal scrollbar inside a dialog that never had one. No test covers it. `docs/frontend-ui-style-guide.md § Implementation rules` is explicit that feature-specific styling must not be introduced as generic global layers. Fix: scope every one of these under the `.price-changes-queue` root the component already renders, or prefix them (`.price-changes-queue__table-wrap`). The same hazard is latent — not yet live — for `.is-resolved`, `.is-grouped`, `.is-flagged`, `.row-actions`, `.status-note`, `.undo`, `.connection-tag`, `.filter-bar`, `.price-compare`, `.col-num`, `.th-sub`; those have no current users, but `.is-resolved` as a global name in a 20k-line shared stylesheet will not stay unused.

**[IMPORTANT]** — `apps/web/src/features/price-changes/components/price-changes-queue-table.tsx:409`
> The "All" chip renders `unfilteredQuery.data?.total ?? 0` — so a count that is still loading, or whose read **failed**, renders a confident `0` beside rows that may be non-empty. This is the identical rule the tab badge was just fixed to obey (`listings-list-page.tsx:751-756`, `—` / skeleton / value): `docs/frontend-architecture.md § Paginated Totals As A Second Stage` — *"A failed count leaves the placeholder, and never renders `0`. Absence and 'none matched' are different claims."* Render a placeholder (or omit the count) while `unfilteredQuery.isPending`, and a distinct marker on `isError`.

**[IMPORTANT]** — `apps/web/src/index.css:1885-1902`
> Still **zero** `@media` queries across the 310 new lines, with `min-width: 900px` on the table and `overflow-x: auto` on the scroller. The added comment honestly documents the deferral and `-webkit-overflow-scrolling: touch` improves the scroll, but a 7-column queue behind a horizontal scrollbar is not a tablet surface, and mobile + tablet are in default scope for FE work here. `shared/ui/data-table.tsx` already ships `DataTableCardView` for exactly this. If the migration genuinely cannot happen in this PR, it needs a tracking issue cited in the comment rather than an unowned "deferred".

**[IMPORTANT]** — PR description, and `docs/plans/mockups/price-changes-review-queue.html`
> Two things. (a) `docs/frontend-architecture.md § UX Mockups`: an implementing PR for a feature with a committed mockup must **either** touch the mockup **or** state in its description why the mockup is still accurate. This does neither, and the review round changed real behaviour the mockup describes (the edit affordance, the refresh button, the arrow icon on delta chips). (b) The description is now **stale in a way that would mislead the merge record**: it still says *"Edit uses a native `window.prompt()` placeholder"* and *"Accept calls the API directly (no confirm dialog yet)"* — both untrue as of `b7f763e`. Please refresh it before merge.

**[SUGGESTION]** — `price-changes-queue-table.tsx:182`
> `unfilteredQuery` inherits the hook's default `refetchInterval: 30_000`. With a filter active the queue tab therefore holds three distinct cache entries (page badge, filtered table, chip counts) polling an unbounded endpoint every 30 s. The chip counts are a slowly-changing sidebar number — pass `refetchIntervalMs: false` and let the ordinary invalidation on accept/ignore refresh it.

**[SUGGESTION]** — `price-changes-queue-table.tsx:349`
> `handleBulkIgnore` clears the **whole** selection before reporting, including the items that failed, so "Kept 9, but 3 failed. Try again for the rest." leaves the operator nothing selected to retry. Keep the failed ids selected.

**[SUGGESTION]** — `price-changes-queue-table.tsx:479`
> `DataTable` remains bypassed for a hand-rolled `<table>`; it is in the shared catalog and owns virtualization plus the card view. Same underlying gap as the responsive item — worth one issue covering both.

### Not this PR's defect, but blocks the merge

`Integration Tests` is red: every failure is in `test/integration/listings-price-change-episode.int-spec.ts` (`invalid input syntax for type uuid: "no-such-source"` raised from `price-change-episode.repository.ts:536`), i.e. backend code owned by the stack base (#3162/#3163). Nothing here can fix it; flagging so it is not mistaken for a frontend regression, and so it is not merged red.

**Done well:** the `groupFirstIndex` fix is the right one — it handles the non-adjacent case rather than just re-ordering, and says so in a comment that names the cause (`detectedAt` pinned at first detection). The `roundingClauseFor` split is also correct rather than a patch: `nearestWhole` got its own clause instead of being forced through "ends in". And `deltaToneFor` deriving the tone from the **rounded** value closes a genuine internal contradiction (a chip reading `0%` coloured as a decrease).

### Verdict

❌ **Request changes**

1. Scope or prefix the new `.table-wrap` rules — they currently restyle the Subiekt invoice-correction dialog.
2. Stop the "All" chip rendering `0` for an absent or failed count.
3. Refresh the PR description (stale `window.prompt` claim) and reconcile the mockup per § UX Mockups.
4. Give the responsive/`DataTable` deferral a tracked issue.

(1) and (2) are small; the rest are description/tracking work.

---

norbert-kulus-blockydevs:
Raised the responsive / `DataTable` deferral as **#3223**, covering both symptoms in one issue (the migration is what supplies the card view).

Citing `docs/frontend-ui-style-guide.md § Responsive` rather than a general principle: its parity matrix specifies **card view** for tables on mobile, and `shared/ui/data-table.tsx` already ships `DataTableCardView` with the `(max-width: 767.98px)` switch that section documents as load-bearing.

When the round-3 fixes land on this PR, the tracking comment in `apps/web/src/index.css` will cite #3223 by number, which is the form your review asked for. The `.table-wrap` global-class collision is deliberately **not** in that issue — it is a BLOCKING fix owned by this PR, and #3223 assumes it is already scoped or prefixed.

The rest of your findings on this PR are still open.

norbert-kulus-blockydevs:
All four items addressed — `849d50f6c` and `efc71692e`, on top of the parent catch-up `6e1629e74`.

**1. BLOCKING — `.table-wrap` restyling the Subiekt dialog.** Confirmed exactly as you described, and it was worse than one class: 29 of the block's selectors were unprefixed globals with no rule on `main`. I scoped all 43 rules under `.price-changes-queue`, the root the component already renders at the top of its tree, which encloses the filter bar, the table and `AutoAppliedNote` alike — so no TSX changed and the markup still reads the way the mockup does. Prefixing each selector rather than wrapping the block in one nested rule keeps every scope visible in the diff instead of hiding it behind a re-indent.

`.chip__count` is deliberately left global: `.chip` is a shared primitive on `main` with three rules of its own, and that is its BEM element, already namespaced by its block.

While checking the blast radius I found `.table-scroll` looked like a second live collision — it is not. The three correction flows all use prefixed variants (`infakt-correction__table-scroll`, `ksef-correction__table-scroll`, `subiekt-correction__table-scroll`); only Subiekt's bare `table-wrap` was real.

**2. IMPORTANT — the "All" chip rendering `0`.** Fixed, and the per-connection chips had the identical bug through `connectionCounts.get(id) ?? 0` — they derive from the same unfiltered read, so a failure collapsed all of them, not just "All". They now go dark together: no count while pending, an em dash on failure, the number once known. `?? 0` stays on the known path, where a connection genuinely absent from the map has no open episodes. Two tests pin both degraded states, and I checked both are red against the previous code rather than merely green against the new one.

**3. IMPORTANT — the responsive deferral.** Raised as **#3223**, covering the missing `@media` queries and the `DataTable`/`DataTableCardView` migration in one issue, because the migration is what supplies the card view. The `index.css` comment now cites it by number and says to delete the paragraph together with the deferral. The issue quotes `§ Responsive`'s parity matrix (card view for tables on mobile) and the `(max-width: 767.98px)` boundary that section documents as load-bearing.

**4. IMPORTANT — description and mockup.** Description rewritten; the `window.prompt` and "no confirm dialog" claims are gone, and the round-2/round-3 changes are recorded. On the mockup: its row actions, refresh note and delta-chip arrow already match what shipped, so the three things you named had in fact converged. The one real divergence is structural — there, `queue-error` hides `#queue-live` and the chips go with it, whereas the shipped component keeps the filter bar outside the loading/error swap, so chips survive a failed counts read. I recorded that in the mockup's own state-partition comment rather than lifting the filter bar out of `#queue-live`, since that would move the screenshot baseline #3152 is written against. Happy to restructure it instead if you would rather the mockup model the state.

**SUGGESTIONs** — both taken. The counts query drops its standing 30s poll (all six mutations invalidate `priceChangesQueryKeys.all`, so it cannot go stale behind an action taken here), and a partial bulk-ignore keeps the failed rows selected so "try again for the rest" has something to act on.

**One defect the local gate caught that no review round did:** the parent merge left both the pre-rename `1879000000000` auto-applied-log migration and its `1881000000000` replacement in the tree. `check-migration-timestamps` failed on it; the stale file is removed and nothing referenced its class.

Verified: `type-check` clean, `check:invariants` clean, 94 `apps/web` tests passing across `features/price-changes` + `pages/listings`, `check-css-structure` reporting 3625 balanced blocks after the re-scoping.

piotrswierzy (approved):
## Tech Lead re-review (`12cc79a2` → `10d651ab`)

Re-reviewed the delta. Strip out the `origin/main` merge (14 PRs) and the PR-authored work is one commit — `b7f763e9` — plus the mockup reconciliation. **Every finding from my previous pass is closed**, and closed in the right place rather than papered over: the CSS refactor is now scoped, the chip count obeys the never-render-zero rule, the responsive gap is tracked under a real issue with the deferral written into the stylesheet next to the thing it defers, and the mockup carries an explicit divergence note. The commit also self-identified and fixed two genuine regressions I had not reached — an entirely ungated set of write affordances, and a bulk-accept body that would have 400'd against the now-required `expectedVersion`. `Test` and `Integration Tests` are green on this head.

- **BLOCKING (global `.table-wrap` leak)** — closed. All 14 rules are now `.price-changes-queue .table-wrap …` (`apps/web/src/index.css:1880-1970`). `plugins/subiekt/components/subiekt-invoice-correction-flow.tsx:189` still renders `className="table-wrap …"` and now inherits nothing from this PR — no wrapper border, no `overflow: hidden`, no `min-width: 900px`, no forced horizontal scrollbar. `main` continues to carry no unprefixed `.table-wrap` rule.
- **IMPORTANT ("All" chip renders `0`)** — closed. `renderChipCount` (`price-changes-queue-table.tsx:242-250`) is a real three-state treatment: `—` on `isError`, nothing while the read is in flight, the number only once `data !== undefined`. The `?? 0` at the call site is unreachable in both absent states, and the comment says why. This is the `§ Paginated Totals As A Second Stage` rule applied correctly.
- **IMPORTANT (zero media queries)** — closed as a tracked deferral, which is the right disposition for the size. `#3223` exists and is open, the stylesheet cites it inline at `index.css:1887-1901`, momentum scroll is added, and the note names the style guide's own `§ Responsive` parity matrix as the thing being departed from. "Delete this paragraph together with the deferral, not before" is the correct shape for a comment like this.
- **IMPORTANT (mockup unreconciled / stale description)** — closed. The mockup gains a `KNOWN DIVERGENCE` block recording the one state it structurally cannot show (chips surviving the loading/error swap because the filter bar renders outside it), and explains that modelling it would move the screenshot baseline `#3152` is written against. That is a better outcome than editing the mockup to match.

### Findings

**[SUGGESTION]** — `apps/web/src/features/price-changes/components/price-changes-queue-table.tsx:80`

> `CHIP_COUNTS_LIMIT = 200` means the **per-connection** chip counts are derived from at most 200 rows, so on an install past that ceiling a connection chip renders a confident number that is silently low. The header comment is honest about it and the `All` chip is unaffected (it reads the authoritative `total`), but nothing tells the operator — and a count is a claim about their own data, which is the same principle the `—`-not-`0` treatment three lines up exists to honour.
>
> It is a SUGGESTION rather than an IMPORTANT for two reasons: the under-count is self-correcting the moment the chip is clicked (the filtered read's own `total` is server-side and exact), and a chip label is a filter affordance rather than a KPI. The real fix is the counts-by-connection endpoint the comment already names; if that is not in scope, consider folding this into `#3223`'s body so the ceiling has somewhere to be removed from rather than living only in a code comment.

No blocking or important issues remain.

### Also worth saying

The two problems the fix commit found on its own are the interesting part of this delta. The ungated write affordances — a `viewer` session seeing Accept/Edit/Keep price/Undo/Refresh fully enabled against endpoints guarded `@Roles('admin','operator')` — were a real defect, and the fix takes the established `useWriteAccess('listings:write', demoMode)` + `ReadOnlyLock` shape (`docs/frontend-architecture.md § Access Control And UI Visibility`, row two) rather than inventing a local one, including hiding the row and select-all checkboxes when nothing can act on a selection. And catching that `handleBulkAccept` was still sending bare `{ id }` after the `#3162` round made `expectedVersion` required is exactly the kind of stacked-branch contract break that is easy to ship and expensive to debug.

### Verdict

✅ **Approve**

---

norbert-kulus-blockydevs:
Taken the second of the two options you offered: the `CHIP_COUNTS_LIMIT = 200` ceiling is now folded into **#3223**'s body as its third symptom, so it has somewhere to be removed from rather than living only in a code comment. The issue records that the `All` chip is unaffected (authoritative server-side `total`), that the under-count self-corrects on click, and that the real fix is the counts-by-connection endpoint — with your framing that a count is still a claim about the operator's own data, which is the principle the `—`-not-`0` treatment three lines up exists to honour.

No code change here: the honest options were that endpoint (out of scope for this PR) or a label that hedges every per-connection count for a ceiling most installs never reach.

piotrswierzy (approved):
## Tech Lead re-verification (head unchanged at `178098b`)

**The CI condition is RESOLVED on the job that matters: `Integration Tests` passed (32m20s)**, alongside `Type Check`, `Docker Build Smoke Test` and `PHP Unit Tests`. `Build`, `Lint` and `Test` are still `queued` from 09:38 behind a ~2h self-hosted runner backlog — reported as outstanding, not as passing. `MERGEABLE`, no conflict.

Re-verified in code: the global `.table-wrap` leak is gone (zero unscoped `.table-wrap` rules added; 14 rules scoped under `.price-changes-queue .table-wrap …`), `renderChipCount` is a genuine three-state (`—` when unavailable, `null` when not yet known, the value when known — never a fabricated `0`), and the mockup carries its `KNOWN DIVERGENCE` block rather than being edited, so the #3152 screenshot baseline is preserved.

**#3238 does not change the deferral's status — it is the deferral being honoured**, at stack position 11, branching from #3236. What did change is better than I asked for: **#3223 has been narrowed**, and the `CHIP_COUNTS_LIMIT = 200` under-count I raised as a remaining suggestion is now explicitly folded into **#3237** together with the `DataTable` migration and the `rowClassName`/`rowAttributes` primitive extension it is blocked on. So my one open suggestion from last round is tracked rather than lost, and the ordering (#3238 before #3237) is right — migrating first would mean writing the card twice.

### Findings

### Verdict
✅ **Approve**

---

piotrswierzy (approved):
## Tech Lead re-verification (head unchanged at `6a122e5`)

Head unchanged, `main` unmoved. CI on run 34845739661: **Docker Build Smoke Test, PHP Unit Tests and Detect changed paths pass; Build, Lint, Test, Type Check and Integration Tests are queued behind the runner backlog. No red anywhere** (Integration Tests passed on the earlier run of this head). Re-checked the three load-bearing items: `grep -cE '^\+\.table-wrap'` on the diff returns **0** unscoped rules, and **14** are scoped under `.price-changes-queue`; `renderChipCount` is a genuine three-state (`—` when `unfilteredQuery.isError`, `null` while not yet known, the number otherwise); the `KNOWN DIVERGENCE` mockup block is still in place.

Follow-ups both honoured as recorded: the media-query deferral is tracked as **#3223** and **#3238 does not change it**; the `CHIP_COUNTS_LIMIT = 200` suggestion is tracked in **#3237** with the `DataTable` primitive blocker named. Neither reopens this review.

Stack: base `3146-connection-pricing-sync-settings-api` (#3163). **#3159 and #3162 below still carry an open BLOCKING** — an in-place edit to the already-merged migration `1880000000000-create-price-change-episodes.ts`, which silently diverges an already-migrated database from a fresh install. This PR carries no migration, but **nothing in the chain merges until those two are fixed** — that is why an approved PR is not moving. `mergeable: MERGEABLE`; `mergeStateStatus: BLOCKED` reflects the unmerged base, not a defect here.

### Findings

### Verdict
✅ **Approve**

---

PR #3165, opened 2026-09-10
feat(web): accept/edit dialogs + bulk apply with live progress

### LINKED ISSUE #3148: feat(web): accept/edit dialogs + bulk apply with live progress

norbert-kulus-blockydevs:
## Summary

Part of #3139 (epic). **Stack position: ninth of ten** — branches from `3147-price-changes-tab-and-queue-table` (#3164, open; base moves to `main` once that merges).

Replaces the parent's minimal Accept/Edit confirmations with the three real dialogs from the mockup, and wires up bulk-accept's live progress:

- **`AcceptPriceChangeDialog`** — price hero, rule sentence, "also set to Automatic" opt-in, fixed disclaimer.
- **`EditPriceChangeDialog`** — live-validated manual price (blocking `≤0` error, non-blocking `>30%` deviation warning, recomputed on every keystroke), parsed through `parseLocalizedAmount` and clamped to the **currency's** minor units.
- **`BulkAcceptPriceChangesDialog`** — one opt-in checkbox per **distinct** `(source, connection)` pair, a mini-row per item.
- **`BulkPublishProgress`** — polls the real `bulk_offer_creation_batches` progress endpoint, with loading and error states and `failedCount` in its terminal title.

### Two deviations from the issue's stated assumptions, both resolved rather than left ambiguous

1. **"Also set to Automatic" sequencing**: the issue assumed a separate mutation after accept succeeds. #3145/#3146 already bundle `optInAutomatic` into the accept call, so that is what this uses.
2. **`shared/ui/toast-provider.tsx` had no action-button slot at all.** Added `ShowToastOptions.action: {label, onClick}` as a small, additive, backward-compatible extension — plus `durationMs`, which the round-3 fix below needed.

### Known, stated gap

`BulkBatchSummary.records` carries no `connectionId` (offer-creation batches are always single-connection, so it was never needed), while a price-change batch can span several destinations. The progress bar therefore reports **composition only** — which connections the batch covers — and never a per-connection `n/N`. An earlier revision rendered per-connection counts that flipped to "done" together; that was a fabricated claim and the review replaced it with the composition-only lanes, which is what ships.

### What the round-3 review changed

Three commits on top of the parent catch-up:

1. `15ba8be12` — **the suite is green again**, from three independent causes. `parseLocalizedAmount` refused `"1,234,567"`: the dot branch strips multi-separator grouping and the comma branch had no such arm, so an unambiguous grouped thousands value came back `NaN` and the operator was told to "enter a valid price". It also returned `0` for an empty string (`Number('')`), against its own documented NaN-for-invalid contract. Three queue-table cases still rendered without a session after the parent's write gating landed — two now take the ordinary admin session, and the non-admin checkbox case takes an explicit **operator**, since the suite-wide default user is an admin and passing it would have asserted nothing. And the clamp case asserted `399.1235`, the storage column's precision, which is exactly the behaviour the code deliberately replaced.
2. `1096cc09b` — **the automatic opt-in gets a real decision window.** Its Undo rode the provider's 4s default, and that toast is the only chance to reverse a switch to publishing prices *without review* before the setting persists silently on the connection. Now 30s. The mutation behind it also invalidates the price-changes list on success.
3. The dialogs' stylesheet cites **#3223** for its missing `@media` coverage, now widened to cover the whole feature's responsive treatment rather than splitting one styling pass across two tickets.

The server-side price ceiling the review asked to be raised on the epic is **#3222**.

## Mockup

`docs/plans/mockups/price-changes-review-queue.html` is accurate for this PR's surfaces and is not changed here. The parent (#3164) touched it for the one divergence the pair introduces — the filter bar surviving a failed counts read, which the mockup's whole-page `queue-error` cannot show. Nothing in this PR's dialogs diverges from it: the price hero, rule sentence, opt-in checkbox and mini-rows all render as drawn, and the composition-only progress lanes are what the mockup shows.

## Test plan

- [x] `pnpm check:invariants` — clean.
- [x] `apps/web` `features/price-changes` — 51 pass, including the four cases that were red.

Closes #3148


apps/web/src/features/price-changes/hooks/use-set-source-sync-mode-mutation.ts
```diff
@@ -0,0 +1,45 @@
...
+  SetSourceSyncModeInput
+> {
+  const apiClient = useApiClient();
+
+  return useMutation({
+    mutationFn: async ({ destinationConnectionId, sourceConnectionId, mode }) => {
+      const view = await apiClient.pricingSync.get(destinationConnectionId);
+      const sourceOverrides: Record<string, PricingSyncSetting> = {};
+      for (const source of view.sources) {
+        sourceOverrides[source.sourceConnectionId] = source.effective;
```
norbert-kulus-blockydevs, line 34: **BLOCKING — Undo materialises an explicit override for every source that was inheriting the destination default.**

```ts
for (const source of view.sources) {
  sourceOverrides[source.sourceConnectionId] = source.effective;
}
```

`ConnectionPricingSyncSourceEntry.effective` is documented as *"the EFFECTIVE (override-then-default) setting"*, and `isCustomOverride` is `false` when the source has none of its own — `buildView` computes it as `overrides[sourceId] ?? default`. `view.sources` is the union of overridden sources **plus every source with an open price-change episode**, i.e. on a fresh connection potentially every source in the queue. `updatePricingSync` then writes `input.sourceOverrides` **verbatim**.

So one Undo click pins today's default onto every inheriting source, permanently. A later change to the destination's default rule on the settings page then affects **none of them** — silently, with no error and nothing in the UI showing that inheritance was broken. On a connection with 20 open episodes and zero overrides, that is 20 pinned overrides from one click.

`isCustomOverride` exists on the type for precisely this and is read by nothing here. The existing test does not catch it because its fixture is `sources: []` — the one shape that hides the bug.

**Fix:** build `sourceOverrides` from entries where `isCustomOverride === true`, plus the one pair being changed; add a one-source `isCustomOverride: false` fixture. (One residual belongs on #3163: `isCustomOverride` is a single flag ORed over both the rule and mode axes, so a rule-only override still materialises a mode. Writing for `isCustomOverride: false` sources is wrong regardless.)

norbert-kulus-blockydevs (commented):

Strong work in places, and one thing I want to credit specifically because I went and verified it rather than taking the body's word: **the progress reuse is genuine end to end.** `BulkListingProgressService.advanceBatchStatus` derives terminal purely from `succeeded + failed === totalCount` with no `OfferCreationRecord` dependency; `PriceChangeApplyHandler` passes `episodeId` as the child key; `getBatch` returns a summary with `records: []` rather than 404-ing. So a price-change batch **can** reach a terminal status and the bar and count are real — ADR-072's bulk-progress decision is honoured properly, not nominally. All three dialogs use the real Radix-backed `shared/ui/dialog`, so focus trap / Escape / focus restoration / `aria-labelledby` come for free; `.toast-region` z-index 60 sits above `.dialog__content` 50, so an error toast from an open dialog is not hidden; the `optInAutomatic` sequencing deviation was researched against #3162's actual code and documented; and the PR correctly does **not** re-derive `isSteep`/`deltaPct`.

Three blocking, inline. The one I would fix first is the Undo: it cannot work for an `operator` (the role that will use this screen daily can turn automatic publishing on and gets a 403 turning it off), and it fails silently for everyone.

Also worth noting as stack-order: this PR's Edit-dialog permalink targets a route that lands in **#3167**, not #3166 as the body says — so it 404s even alongside #3166. And `useSetSourceSyncModeMutation` invalidates nothing, while #3167 ships a sibling mutation that invalidates both `connectionPricingSyncQueryKey` and `priceChangesQueryKeys.all`; cheap to align now.
---
*Part of a whole-stack review of #3158–#3168 (epic #3139). Cross-cutting items tracked once: `pnpm lint` is red on the stack; `docs/architecture-overview.md § Listings` is untouched by all ten PRs though ADR-072 names it the primary doc section; the mockup has not changed since #3155 despite ~20 behavioural divergences, which `docs/frontend-architecture.md § UX Mockups` treats as incomplete; `features/price-changes` is registered in **neither** `.eslintrc.js` `no-restricted-imports` group, so the new barrel is unenforced; and a `connections ↔ price-changes` value-import cycle is one barrel export from closing (#337/#359, #2640).*

apps/web/src/features/price-changes/components/price-changes-queue-table.tsx
```diff
@@ -97,10 +101,50 @@ export function PriceChangesQueueTable(): ReactElement {
...
+    const sourceLabel =
+      items.find((i) => i.sourceConnectionId === sourceConnectionId)?.sourceLabel ?? 'this source';
+    showToast({
+      tone: 'success',
+      title: 'Set to Automatic',
+      description: `Future price changes from ${sourceLabel} will publish without review.`,
+      action: {
+        label: 'Undo',
+        onClick: () => {
+          setSourceSyncModeMutation.mutate({
```
norbert-kulus-blockydevs, line 126: **BLOCKING — the Undo button cannot work for an operator, and fails silently for everyone.**

Two independent defects on one control:

1. **Role mismatch across the stack.** `POST /listings/price-changes/:id/accept` is `@Roles('admin','operator')`, so an **operator can opt in to Automatic**. But `PATCH /connections/:id/pricing-sync` is `@Roles('admin')` (#3163), so the Undo that reverses it **403s for that same operator**. `role.types.spec.ts:25` asserts `operator` lacks `connections:write`, so this is not ambiguous.
2. **The failure is invisible.** `.mutate(...)` here has no `onError` and no `mutateAsync`/catch, and the toast removes itself on click. The operator clicks Undo, the toast vanishes, and the connection stays in Automatic with nothing rendered, logged or announced.

The issue's AC is *"shows an Undo toast that actually reverts it"*. Today it reverts for admins and silently lies to everyone else.

**Fix:** add `onError` → error toast (and `onSuccess` → confirmation, which the mockup also had). Then either widen the PATCH gate or gate the opt-in checkbox itself with `useWriteAccess` — offering a control whose reversal the session cannot perform is exactly what `§ Access Control And UI Visibility` exists to prevent.

**Also, bulk opt-in fires one 4-second Undo toast per pair.** Twelve pairs → twelve stacked toasts, twelve independent Undos, each expiring in 4 s (the mockup used ~5.4 s and aggregated into one *"Turned on automatic pricing for {n} sources."*). For the only affordance that reverses an ongoing-automatic-publishing change, both the count and the window matter. `offerAutomaticUndo` also re-finds the source label by searching `items` rather than taking it from the item it already has, so once the list invalidates the copy degrades to *"…from **this source**…"*.

apps/web/src/features/price-changes/components/bulk-publish-progress.tsx
```diff
@@ -0,0 +1,70 @@
...
+interface BulkPublishProgressProps {
+  batchId: string;
+  items: PriceChangeItem[];
+}
+
+export function BulkPublishProgress({ batchId, items }: BulkPublishProgressProps): ReactElement | null {
+  const query = useBulkBatchQuery(batchId);
+  const batch = query.data;
+
+  if (!batch) return null;
```
norbert-kulus-blockydevs, line 34: **BLOCKING — the widget unmounts exactly when the batch finishes, and renders nothing on a failed read.**

`BulkPublishProgress` is mounted **inside** the `items.length === 0 ? <EmptyState/> : (…)` arm of the queue table. `usePriceChangesQuery` has `refetchInterval: 30_000`, the accept path invalidates it, and #3161's worker resolves each episode as it publishes. On the most common run of this feature — select all, publish — the list drains to empty, the branch flips, and the only live evidence of N in-flight marketplace writes is replaced by **"You're all caught up"** while the writes are still going out. A failed list refetch does the same.

Compounding it: `if (!batch) return null` collapses *loading*, *failed poll* and *unreadable* into rendering nothing — and this PR **removed** the old `Publishing selected prices…` toast from the no-opt-in path. So if the batch poll fails, a bulk publish of real prices produces **zero feedback of any kind**. That is the four-states rule (an empty/absent read must never be rendered as a positive claim).

There is also no terminal state: the title stays `Sending your price changes…` after completion, `activeBatch` is never cleared, and there is no dismiss. The mockup fired a `Published new prices on N listings.` toast.

**Fix:** move the widget out to be a sibling of the whole branch, beside the three dialogs, keyed on `activeBatch`; render explicit loading and "couldn't read progress" states instead of `null`; add a done state + dismiss.

apps/web/src/features/price-changes/components/bulk-publish-progress.tsx
```diff
@@ -0,0 +1,70 @@
...
+      <div className="publish-progress__track">
+        <div className="publish-progress__fill" style={{ width: `${pct}%` }} />
+      </div>
+      <div className="publish-progress__lanes">
+        {Array.from(lanes.entries()).map(([connectionId, lane]) => (
+          <span
+            key={connectionId}
+            className={`publish-progress__lane ${isTerminal ? 'is-done' : ''}`}
+          >
+            <b>{isTerminal ? lane.count : 0}</b>/{lane.count} {lane.label}
```
norbert-kulus-blockydevs, line 64: **IMPORTANT — the per-connection lanes render a number that is always wrong until the batch ends.**

`{isTerminal ? lane.count : 0}/{lane.count}` means every lane reads `0/12 Allegro — PL`, `0/8 Erli` for the entire run, then all jump to full simultaneously.

This is not "a static aggregate, honestly presented" — it is a per-connection display **asserting a per-connection fact**, and the fact is false: it says zero of Allegro's twelve are done at a moment when eight may be. The aggregate bar and `done/total` above it are correct and real; the lanes actively contradict them. The docblock disclosure is good practice, but the operator does not read the docblock.

Since `BulkBatchSummary.records` genuinely carries no `connectionId`, the two honest options are to drop the lanes until the backend supports them, or render composition only with no `n/N` counter (`12 · Allegro — PL`). The issue's AC — *"per-connection progress lanes correctly attribute completed items to their connection"* — is not met either way, so shipping a wrong number buys nothing.

apps/web/src/features/price-changes/components/accept-price-change-dialog.tsx
```diff
@@ -0,0 +1,111 @@
...
+ * UNDO path, never on the happy path.
+ *
+ * @module apps/web/src/features/price-changes/components
+ */
+import { useState, type ReactElement } from 'react';
+import { Dialog, DialogContent, DialogDescription, DialogFooter, DialogTitle } from '../../../shared/ui/dialog';
+import { Button } from '../../../shared/ui/button';
+import { formatAmount } from '../../../shared/format/format-amount';
+import type { PriceChangeItem } from '../api/price-changes.types';
+import { ruleSentenceFor } from '../lib/price-change-copy';
```
norbert-kulus-blockydevs, line 27: **IMPORTANT — this PR is `ruleSentenceFor`'s first consumer, and it emits broken English.**

For the fixture used in this PR's own tests (`{type:'margin', percent:22, rounding:'endingIn99'}`):

> **"Keeps a 22% margin as a margin, then rounds so it ends in ending in .99."**

and for `nearestWhole`: *"…then rounds so it ends in nearest whole."* Both stutter because `noun`/`roundingLabel` already contain the words the template repeats.

I checked `origin/3147-…`: `ruleSentenceFor` had **no consumer before this PR** (the queue table uses only `roundingLabelFor`). It was dead code that type-checked; this PR puts it on the single most consequential screen in the feature. That makes fixing it this PR's job, not a later one.

**Also missing from the same line:** the mockup's closing clause `— that comes to <b>{price}</b>, VAT included.` It is the only place the accept dialog reconciles the rule with the number it is about to publish, and the issue calls it required copy. `ruleSentenceFor` takes only `{type, percent, rounding}` and structurally cannot emit it — pass the computed amount and currency.

apps/web/src/features/price-changes/components/edit-price-change-dialog.tsx
```diff
@@ -0,0 +1,124 @@
...
+          <>
+            <DialogDescription>
+              {item.productName} on {item.destinationLabel}
+            </DialogDescription>
+            <p className="rule-note">
+              Source ({item.sourceLabel}): <b>{formatAmount(item.sourceOldAmount, item.sourceCurrency)} →{' '}
+              {formatAmount(item.sourceNewAmount, item.sourceCurrency)}</b>.
+            </p>
+            <div className="field">
+              <label className="field__label" htmlFor="price-change-edit-input">
```
norbert-kulus-blockydevs, line 79: **IMPORTANT — the blocking validation error is not associated with its input, and the app ships the primitive that does it.**

The price input has no `aria-invalid`, the `.field__feedback` div has no `id`/`aria-describedby` link and no `role="alert"`. A screen-reader user typing `0` into the field that decides a live marketplace price gets no announcement and no association between message and control.

`shared/ui/form-field.tsx` does exactly this wiring (`id`, `aria-describedby`, `aria-invalid`, `useId()`), and `shared/ui/field-error.tsx:13` carries `role="alert"`. Both are in the `shared/ui/index.ts` catalog.

Two further symptoms of hand-rolling instead:
- `className="field__label"` is **defined nowhere** — this PR adds `.field__hint`, `__price-input`, `__currency`, `__feedback`, `__link-row` but not `__label`, and it is not in the base sheet either. The label renders at inherited size.
- `id="price-change-edit-input"` is a hardcoded DOM id; `FormField` derives one from `useId()`.

**Related CSS defect:** `.field__currency` (specificity 0,1,0) sits at `index.css:2308`, while `.field span { color: var(--text-primary); font-weight: 600 }` (0,1,1) sits at 3035 — later **and** more specific. The currency prefix therefore renders bold and primary-coloured rather than the intended muted chip. Scope it (`.field__price-input .field__currency`) or stop reusing the generic legacy `.field` container.

apps/web/src/features/price-changes/components/edit-price-change-dialog.tsx
```diff
@@ -0,0 +1,124 @@
...
+}
+
+const DEVIATION_WARN_THRESHOLD = 0.3;
+
+interface Feedback {
+  tone: 'error' | 'warn' | null;
+  message: string;
+}
+
+function feedbackFor(rawValue: string, rulePrice: number): Feedback {
```
norbert-kulus-blockydevs, line 33: **IMPORTANT — the deviation warning prints an unformatted money figure, and the parser rejects legitimate European input.**

```ts
`…% vs. the rule price (${rulePrice}) - a large manual deviation.`
```
renders `(399)`. Every other price on the same screen goes through `formatAmount(...)`; the mockup reads `+45% vs. the rule price (419,99 zł)`.

Separately, `Number(rawValue.replace(',', '.'))` replaces only the **first** comma, so `1 234,56` and `1.234,56` both become `NaN` and produce the *blocking* error `Enter a price greater than 0.` — a wrong diagnosis of a valid number, on a PLN-denominated screen. The mockup stripped non-numerics first.

There is also no decimal-place clamp: `399.99999` passes the browser, passes `@IsNumber() @IsPositive()` server-side, and is silently rounded to 4 dp by the `numeric(14,4)` column.

(On the `>30%` threshold: I checked deliberately and it is **not** a mirror violation — there is no such notion anywhere in core, it is frontend-only and non-blocking, so `check-*-mirror.mjs` does not apply. It should cite the mockup/issue as its source the way `STEEP_DELTA_TOOLTIP` cites its ADR decision. And `≤0` **is** enforced server-side, so that half is fine — what is unbounded is `@Max` and `maxDecimalPlaces`.)

apps/web/src/features/price-changes/components/bulk-accept-price-changes-dialog.tsx
```diff
@@ -0,0 +1,121 @@
...
+        onOpenChange(next);
+      }}
+    >
+      <DialogContent className="dialog__content--wide">
+        <DialogTitle>Publish {items.length} price changes</DialogTitle>
+        <DialogDescription>Going out to {destinationCount} connections.</DialogDescription>
+        <div className="mini-list">
+          {items.map((item) => (
+            <div className="mini-row" key={item.id}>
+              <span className="thumb" style={{ width: 24, height: 24, fontSize: 10 }} aria-hidden="true">
```
norbert-kulus-blockydevs, line 66: **IMPORTANT — the initials avatar is completely unstyled.**

`<span className="thumb" style={{width:24, height:24, fontSize:10}}>` renders inside `.mini-row`. The **only** `.thumb` rule in the sheet is `.cell-product .thumb` (`index.css:1968`) — descendant-scoped — and this PR adds no standalone `.thumb`. So there is no background, no `border-radius`, no `display: grid; place-items: center`; and because a `<span>` is `display: inline`, the inline `width`/`height` are ignored too. The issue's *"thumbnail-style initials avatar"* renders as two bare letters. The mockup's class vocabulary was reused correctly; the CSS to back `.thumb` standalone was not brought across. (#3168's dialog has the same defect — one `.mini-row .thumb` rule fixes both.)

**Also in this file:** `initials()` returns `''` for a name with a leading space — needs `.filter(Boolean)` before `.slice(0,2)`, and note it is now defined **three times** inside `features/price-changes` with the copies already drifted (only #3168's has the filter). Extract one.

**And this dialog has no test file at all.** The per-pair dedup (`pairKey` + the `Map`) is the one genuinely non-trivial piece of logic in these dialogs and the thing the issue names by name (*"mirrors mockup's `pairs = uniq(...)` dedup logic"*). Untested — as is `BulkPublishProgress` and the Undo failure path.

apps/web/src/features/price-changes/components/price-changes-queue-table.tsx
```diff
@@ -97,10 +101,50 @@ export function PriceChangesQueueTable(): ReactElement {
...
+        input: { expectedVersion: item.version, optInAutomatic },
+      });
+      setAcceptTarget(null);
+      if (optInAutomatic) {
+        offerAutomaticUndo(item.destinationConnectionId, item.sourceConnectionId);
+      } else {
+        showToast({ tone: 'success', title: 'Price published', description: item.productName });
+      }
     } catch (error) {
       showToast({ tone: 'error', description: error instanceof Error ? error.message : 'Failed to accept' });
```
norbert-kulus-blockydevs, line 149: **IMPORTANT — 409/422 surface as raw backend exception strings.**

`showToast({ tone: 'error', description: error.message })` on a stale accept renders:

> `Price change episode ep-1 changed since it was last read (expected version 2026-09-10T10:00:00.000Z, current 2026-09-10T11:04:12.913Z)`

and on a blocked one: `… is blocked (currency_mismatch) and cannot be accepted`. With `refetchInterval: 30_000` and each dialog holding a snapshot captured at open time, the stale 409 is a **routine** outcome, not an edge case.

The repo's discipline is #2341 (return a stable code, never the caught message) and #2100 (a persisted reason vocabulary the FE maps to operator copy). Map at least `409` → *"This price changed again while you were reviewing — refresh and take another look"* and `422` → the block reason the row should already be rendering; keep the raw string in the console.

apps/web/src/features/price-changes/components/edit-price-change-dialog.tsx
```diff
@@ -0,0 +1,124 @@
...
+                <span className="mono">{formatAmount(item.computedNewAmount, item.destinationCurrency)}</span>.{' '}
+                <button type="button" onClick={() => setRawValue(String(item.computedNewAmount))}>
+                  Use that price
+                </button>
+              </div>
+            </div>
+            <p className="field__link-row">
+              This only changes this one price. Want every future change from this source to use a
+              different rule?{' '}
+              <Link to={`/connections/${item.destinationConnectionId}/pricing-sync?source=${item.sourceConnectionId}`}>
```
norbert-kulus-blockydevs, line 106: **IMPORTANT (stack-order) — this permalink 404s until #3167, not #3166.**

The body cites "(#3149)". `app/routes/pricing-sync.route.tsx` + `pages/connections/pricing-sync-page.tsx` land in branch `3150-…` (**#3167**); `3149-…` (#3166) puts the section inside `EditConnectionForm` at `/connections/:id/edit` instead. So merged to `main` alone — **or even alongside #3166** — this link is dead. #3164's source tag has the same problem one PR earlier.

Worth confirming downstream that #3167's page honours the `?source=` param this link sends ("with the relevant source row pre-expanded") — it does read it, but see my #3167 review: the pre-expanded row is never scrolled to or focused, and the Save button it needs never renders.

apps/web/src/shared/ui/toast-provider.tsx
```diff
@@ -11,7 +11,13 @@ import {
 } from 'react';
 import type { AlertTone } from './alert';
 
+export interface ToastAction {
+  label: string;
+  onClick: () => void;
+}
+
 interface Toast {
+  action?: ToastAction;
```
norbert-kulus-blockydevs, line 20: **SUGGESTION — the shared-primitive change is clean, with two consequences worth checking.**

`action?` is genuinely optional and backward-compatible — I checked every existing `showToast` call site and none changes. Good.

Two things:
- `.toast` is `display: flex; justify-content: space-between` and `.toast__content` has no `flex: 1`, so `.toast__action` lands mid-toast rather than beside Dismiss.
- On coarse pointers the global `button { min-height: 44px }` rule (`index.css:5350`) applies to the 0.68 rem inline "Undo" / "Use that price" text buttons and will blow out their line heights — #2668's explicit carve-out list at `:5371` exists for exactly these inline marks.

Also: an interactive control inside a toast that is a live region needs a moment's thought about announcement order, and there is no `toast-provider.test.tsx` anywhere, so this contract change is covered only indirectly by a feature test. A note in `frontend-ui-style-guide.md`'s toast pattern and a `/dev/ui` gallery entry would be worth adding.

piotrswierzy (changes requested):

The dialogs themselves are the best-built part of this stack — real `Dialog` primitive, live validation with a genuine refusal path, and two subtle things done right:

- **The dialog targets are frozen snapshots** (`useState<PriceChangeItem | null>`), not re-looked-up from `items`. That is what makes `expectedVersion` mean *"the version I was shown"* and what stops the 30-second poll wiping a half-typed price out from under the operator. Easy to get wrong.
- **Progress is read from a durable server-side counter**, not accumulated client-side — `useBulkBatchQuery` polls the real `bulk_offer_creation_batches` row, stops on terminal status, and is keyed by `batchId`, so a refresh or a second operator sees the same numbers. Reusing the existing hook and exporting it through the `listings` barrel rather than deep-importing is exactly right.

The blockers are all in what happens *around* them.

### BLOCKING — the Undo silently pins every inherited pricing rule on the connection

`use-set-source-sync-mode-mutation.ts` (~35–40). The read-modify-write writes `sourceOverrides[source.sourceConnectionId] = source.effective` for **every** source in the view, ignoring `PricingSyncSourceEntry.isCustomOverride` — the field that exists precisely to say whether a source had an explicit override or was inheriting `default`. (I grepped: `isCustomOverride` appears once in the diff, and not in this loop.)

So after one Undo click, **every source on that connection is pinned**, and changing the connection's default rule silently stops affecting any of them. The hook's own docblock claims it *"preserves every other source's effective setting, touching only the one (destination, source) pair's mode"* — it does the opposite.

**The test cannot catch this**: the Undo case seeds `sources: []`, the one input where the bug cannot manifest.

### BLOCKING — the Undo is admin-only, and for an operator it fails invisibly

Verified against the base: `POST …/accept` with `optInAutomatic` is `@Roles('admin','operator')`, but `PATCH /connections/:id/pricing-sync` — the only route the Undo can use — is `@Roles('admin')`.

So an **operator** ticks "also set to Automatic", the accept succeeds and the backend flips the mode, the Undo toast appears, they click it, and it 403s — **with nothing shown**, because the mutation is fire-and-forget with no `onError`, no `onSuccess`, no toast. The connection is now permanently on Automatic for them: **every future price change from that source publishes without review**, and the affordance offered to reverse it cannot work for their role.

The underlying asymmetry — a role that can turn unreviewed price publishing **on** but not **off** — is a backend design question worth raising against the epic (#3163/#3146), not patching in the browser. In the meantime: surface the mutation result, and gate the checkbox on the permission the admin-only route actually requires.

### BLOCKING — the edit field's only hard rule is `> 0`, and the parse is locale-hostile

`edit-price-change-dialog.tsx` (~33–44). `Number(rawValue.replace(',', '.'))` replaces the **first** comma only, so `"1,234"` — a thousands separator an operator will absolutely type — parses to `1.234` and publishes a price roughly **1000× too low**. The `>30%` deviation rule catches it, but only as a **non-blocking warning**, so it publishes on one click. No upper bound, and no minor-unit handling on either side of the wire (the server DTO is `@IsNumber() @IsPositive()` and nothing more), so `39900` for a `399` item also publishes behind a warning.

#2610's precedent is explicit that a value which silently produces a wrong published price is refused **server-side as well as in the form**. Here neither side refuses, and the one rule that exists lives only in the browser — which makes it not a rule. Reject ambiguous separators outright, clamp to the currency's minor units, and take the deviation ceiling to the server.

### IMPORTANT — the per-connection progress lanes render fabricated numbers

`bulk-publish-progress.tsx` (~56–62): `{isTerminal ? lane.count : 0}/{lane.count}`. While the batch runs every lane claims `0 / N` regardless of what landed; on *any* terminal status every lane flips to `N / N` **including failures**. The docblock is honest that it is static, but a surface stating "0 of 3 published" where 2 already published, then "3 of 3" where one failed, is a positive false claim — worse than omitting the number, which is the rule applied everywhere else here. Until `records[].connectionId` exists, render the connection names without counts.

### IMPORTANT — "finished with failures" is unrepresentable

`batch.failedCount` is folded into the `done` numerator and never displayed; the header reads *"Sending your price changes…"* forever, including after termination. So the three states collapse into one reading, and an operator who bulk-published 40 prices cannot tell that 6 were rejected by the channel. Render `status` and `failedCount` explicitly. Also `if (!batch) return null;` gives the widget **no loading and no error state** — a failed progress read makes the panel silently vanish immediately after the operator confirmed a bulk publish.

### IMPORTANT — others

- **Hand-rolled form state** where § Form State names RHF + Zod for validated edit flows. A resolver is also the natural home for the min/max/decimal rules above, and a place to map server validation errors back to the field.
- **`.field__label` does not exist anywhere in `index.css`** and this PR does not add it (it adds `field__price-input`, `field__currency`, `field__hint`, `field__feedback`, `field__link-row`). The app's convention inside `.field` is `.field span` / `.field small`. So the field label ships **unstyled** — the one element the mockup's layout depends on. Note `.field` itself is declared twice in `index.css` (2047, 2358, pre-existing), so extending that namespace compounds an already-ambiguous cascade.
- **~15 raw pixel values** where `--space-*` tokens exist and this PR's own parent uses them consistently. `check-design-tokens.mjs` is one-directional and cannot see it, so the "clean" claim is true and unrelated. The mockup also uses `--border-subtle` on `.publish-progress__lane` where this uses `--border-default`.
- **No write gating** (carried from #3164, and widened) — the dialogs are now *the* write affordances, and this PR adds a **new** write path (`Connection.config`) on top of the ungated surface.
- **The stale-version 409 is not distinguished.** Frozen snapshots mean the poll can re-detect while a dialog is open, so 409 on confirm is the one error the operator can act on — and it surfaces as a generic red toast. Map it to the "this price changed again — refresh" copy the table already owns.
- **The Undo toast auto-dismisses in 4 s**, and a bulk confirm fires one per opt-in pair simultaneously. Four seconds is the entire window to reverse a switch to unreviewed publishing. The app's other Undo is a persistent in-row button — the right shape for a reversible decision.

### SUGGESTIONS

No `onSuccess` invalidation on the new mutation — unlike every other in the feature — and no `pricingSync` query key exists, so #3149's settings page and the connection detail page will read stale config after an Undo. The read-modify-write also carries no version guard, so it can clobber a concurrent edit. Separately, bulk is **all-or-nothing** (`bulkAccept` calls `loadActionable` for every item before creating the batch), so one stale row in fifty blocks the lot and the FE shows a raw server message without saying which row. Uncovered by tests: the comma-separator parse (the dangerous path), the override-preservation logic, and `BulkPublishProgress` in any state (no test file).

The toast-action extension is the right size — additive, optional, backward-compatible, built on Radix's own `Toast.Action` with `altText`. Building the primitive rather than dropping the AC was correct.

---

norbert-kulus-blockydevs:
Rebased onto `3147`'s #3164 review-fix commit (`58a31bc16`) and addressed all 11 review findings.

**Rebase note**: `origin/3147-price-changes-tab-and-queue-table` had been rewritten with review-fix commits for every one of this branch's ancestors (detection, worker apply job, queue endpoints, connection settings). Rather than replay-and-reconflict all four superseded commits, I rebased `--onto` the new tip starting from this branch's own unique commit (`67eab3b73`), which produced two small conflicts (`price-changes-queue-table.tsx`/`.test.tsx`) resolved by hand, folding in both #3164's fixes (URL-namespaced filters, Refresh action, unfiltered chip counts, etc.) and this PR's dialog work together.

**1 — BLOCKING (Undo stamps overrides onto every source)**: fixed. `useSetSourceSyncModeMutation` now seeds `sourceOverrides` only from entries where `isCustomOverride === true`, plus the one pair actually being changed — an earlier revision re-sent every source's *resolved* `effective` value, which would have permanently pinned today's default rule onto every other open episode on the connection. Added a regression test with a non-custom sibling source (the previous `sources: []` fixture couldn't have caught this).

**2 — BLOCKING (Undo role mismatch + silent failure)**: fixed by hiding the "also set to Automatic" checkbox entirely for a non-admin session in both `AcceptPriceChangeDialog` and `BulkAcceptPriceChangesDialog` (via `useIsAdmin()`) — an operator can no longer reach a state only an admin's Undo could revert, since the backend's `optInAutomatic` and the pricing-sync `PATCH` are both already admin-only. The Undo mutation now reports `onSuccess`/`onError` toasts instead of failing silently, and a bulk accept's opt-ins fire **one** aggregated "Turned on automatic pricing for N sources" toast/Undo instead of one per pair.

**3 — BLOCKING (progress widget vanishes)**: fixed. `BulkPublishProgress` now renders as a sibling of the queue table's loading/empty/error branch rather than nested inside it, so it survives the list draining to empty. It also renders explicit loading/failed/terminal states with Try again/Dismiss instead of `null`.

**4 — IMPORTANT (fake per-connection n/N)**: dropped — lanes now render composition-only `"N · Connection"` chips (no fabricated progress number), since `BulkBatchSummary` genuinely carries no per-connection breakdown. Flagged as a known gap in the file's docblock rather than a backend change.

**5 — IMPORTANT (missing price clause)**: `AcceptPriceChangeDialog`'s rule sentence now closes with "That comes to **{price}**, VAT included."

**6 — IMPORTANT (a11y)**: wired `aria-invalid`/`aria-describedby` on the price input, `role="alert"` on the error message, replaced the undefined `.field__label` with the existing `.form-field__label`, and scoped `.field__currency` two levels deep so it stops losing to `.field span`.

**7 — IMPORTANT (number parsing / formatting)**: added `parseLocalizedAmount` (handles `1 234,56` / `1.234,56`), switched the deviation-warning's raw number to `formatAmount`, and clamped the confirmed value to the `numeric(14,4)` storage precision.

**8 — IMPORTANT (unstyled thumb, duplicated initials, no test)**: added `.mini-row .thumb` CSS, extracted `initialsFor` into `price-change-copy.ts` (also fixing the leading/doubled-space empty-initial bug), and added `bulk-accept-price-changes-dialog.test.tsx` covering the per-pair dedup logic.

**9 — IMPORTANT (raw backend errors in toasts)**: added `describePriceChangeActionError`, mapping 409/422 to operator-facing copy across accept/edit/bulk-accept/ignore/undo/refresh; the raw error is still `console.error`'d.

**10 — IMPORTANT (permalink to a page that doesn't exist yet)**: pointed at `/connections/:id/edit` instead of `/connections/:id/pricing-sync` (neither #3166 nor #3167 have landed at this point in the stack). **Needs re-checking once those PRs are fixed** — the dedicated pricing-sync settings page may supersede this link.

**11 — SUGGESTION (toast layout/tap targets/no test)**: added `flex: 1` to `.toast__content`, added `.toast__action`/`.status-note .undo` to the coarse-pointer tap-target carve-out list, and added `toast-provider.test.tsx` covering the `action` prop.

Also fixed two null-safety regressions surfaced by this rebase: `PriceChangeItem.computedOldAmount`/`destinationCurrency` were widened to nullable by 3147's own #3164 fix, which the original accept/bulk-accept/edit dialogs (built before that widening) didn't yet handle.

**Verification** (pre-commit hook skipped per explicit project instruction — this machine OOMs on the full `pnpm -r lint`/smart-test hook):
- `pnpm --filter @openlinker/web type-check`: clean
- `eslint` on every touched file: clean
- `pnpm check:invariants`: all green
- `vitest` on every touched/added test file: 36/36 passed

norbert-kulus-blockydevs:
Independent re-verification of this PR's own follow-up comment (2026-09-11T13:56:17Z), after merging in the updated `3147-price-changes-tab-and-queue-table` (a 34-file backend merge from its own parents — resolved cleanly, three FE conflicts in `price-changes-queue-table.tsx`/`.test.tsx` where the write-gating/`ReadOnlyLock` from the updated parent had to be combined with this branch's dialog-based accept/edit/bulk flow and the required `expectedVersion` bulk-accept field).

Cross-checked every one of the 11 original findings against the current code, not just the comment's claims:

**Genuinely confirmed fixed** (re-verified independently, not just re-read):
1. Undo override-preservation — `useSetSourceSyncModeMutation` seeds `sourceOverrides` only from `isCustomOverride === true` entries, and the regression test with a non-custom sibling source actually asserts it.
2. Undo admin-only/role mismatch — `useIsAdmin()` hides the "also set to Automatic" checkbox for non-admins in both dialogs; the Undo mutation reports `onSuccess`/`onError` toasts.
3. Progress widget vanishing — `<BulkPublishProgress>` renders as a genuine JSX sibling of the loading/error/empty ternary, not nested inside it.
4. Fake per-connection n/N — lanes render composition-only `"N · Connection"`, no fabricated progress fraction.
5. Missing price clause, a11y (`aria-invalid`/`aria-describedby`/`role="alert"`, `.form-field__label`), stale-409 mapping via `describePriceChangeActionError`, `.mini-row .thumb` styling, toast tap-targets — all present as described.
6. Write gating — confirmed present (came in via the `3147` merge): `write`/`ReadOnlyLock` gates Accept/Edit/Ignore/Refresh in `ActionCell` and both bulk-bar buttons.

**Found still open on independent re-check, and fixed in `9425089eb`:**

BLOCKING finding 3 was only *partially* closed. `parseLocalizedAmount` correctly handles the multi-separator cases (`1 234,56`, `1.234,56`) the follow-up commit added, but it still silently parsed a lone comma/dot followed by exactly 3 digits — **the review's own flagship example, `"1,234"`** — as `1.234`: a price ~1000× too low, with no error shown (verified with `node -e`, not just read). The follow-up comment's claim to have fixed "number parsing" did not cover this case; `1,234,567`/`1.234.567` (unambiguous multi-separator grouping) worked, but the single-ambiguous-separator shape did not.

Fixed now: `parseLocalizedAmount` refuses that genuinely ambiguous shape outright (returns `NaN`) rather than guessing, `EditPriceChangeDialog` surfaces it as a distinct actionable error instead of the generic "enter a price greater than 0," and the confirmed value is clamped to the destination currency's own minor-unit exponent (reusing the existing `minorUnitExponentFor` from the invoicing feature) rather than only the storage column's flat 4-decimal ceiling. Added the still-missing unit test for both helpers.

**Remaining gaps, left as-is (lower severity, would need a larger follow-up):**
- No server-side deviation ceiling — the review's third ask under finding 3 ("take the deviation ceiling to the server") is still purely a client-side warning; refusing it server-side needs new validation logic in the apply/edit path and is a larger change than a review-fix pass.
- Hand-rolled `useState` form fields rather than RHF+Zod (IMPORTANT, "others") — unchanged; a full-dialog refactor, deferred.
- A handful of raw pixel values remain as inline `style={{...}}` props (`width: 24`/`fontSize: 10`/`gap: 8`) rather than `--space-*` tokens — cosmetic, unchanged.
- The "also set to Automatic" Undo toast still auto-dismisses at the default 4s — unchanged.
- `BulkPublishProgress` still has no dedicated test file (SUGGESTION-level, unchanged).

Verification: `pnpm --filter @openlinker/web lint` (0 errors, 123 pre-existing warnings, unchanged), `pnpm --filter @openlinker/web type-check` (clean), `pnpm check:invariants` (all green).

norbert-kulus-blockydevs:
Caught this branch up with the parent (`3147-price-changes-tab-and-queue-table`), which itself was updated with two upstream CI-fix commits unrelated to review feedback:
- `fix(listings): export PriceChangeDetectionService from services barrel`
- `fix(worker): update stale bulk-lane count in handler-registration spec`

Merged (not rebased, to avoid disturbing the 3 PRs stacked on top) — clean, no conflicts.

Re-verified after the merge:
- `pnpm --filter @openlinker/web lint` / `type-check` — clean (pre-existing warnings only, 0 errors)
- `pnpm --filter @openlinker/core|@openlinker/api|@openlinker/worker type-check` — clean
- `pnpm check:invariants` — all checks pass

The previously-fixed locale-hostile amount parser (rejecting ambiguous separators like `"1,234"`) is untouched by this merge and still in place.

piotrswierzy (commented):
## Tech Lead review

Re-review against `1efff96`. **All three of my BLOCKING findings are properly fixed, and two of them are fixed better than I asked for.** `useSetSourceSyncModeMutation` now re-sends only rows carrying `isCustomOverride: true`, with a docblock that states exactly why `effective` is not the same question — that was the subtlest defect in the stack and it is closed at the right place. The Undo now has `onSuccess`/`onError` toasts instead of failing invisibly, the "also set to Automatic" checkbox is admin-gated to match the `@Roles('admin')` route, and a bulk accept aggregates into **one** Undo toast rather than N independently-expiring ones. `parseLocalizedAmount` replaces the `.replace(',', '.')` path that turned `"1,234"` into `1.234`, refusing an ambiguous separator with its own actionable error, and the confirmed value is clamped to the **currency's** minor-unit exponent rather than the storage column's — the right target. `BulkPublishProgress` gained loading and error states, renders `failedCount` in a terminal title, and the fabricated per-connection `n/N` is gone in favour of composition-only lanes. `describePriceChangeActionError` maps the 409 to the refresh copy the table already owns. Genuinely good work.

The blocker now is mechanical: **this PR's own test suite is red.**

### Findings

**[BLOCKING]** — `Test` check: 6 failing tests, all in files this PR adds or edits
> `apps/web` is red on `1efff96` ([job](https://github.com/openlinker-project/openlinker/actions/runs/34747371249/job/103697673245)). Three independent causes, one of which is a real functional defect:
>
> **(a) `parseLocalizedAmount` is wrong for two inputs** — `price-change-copy.ts:145`.
> - `"1,234,567"` → `NaN` (`expected NaN to be 1234567`). The dot branch at `:169` has an explicit multi-separator grouping arm (`split('.').length - 1 > 1 → strip`); the comma branch at `:160` has **no** such arm, so it falls through to `Number("1.234.567")`. The two branches are meant to mirror each other and do not. An operator typing a grouped thousands value gets a blocking "enter a valid price" on a value that is perfectly unambiguous.
> - `""` → `0`, not `NaN` (`expected false to be true`), because `Number('')` is `0` and `:174` returns it unguarded. Harmless today only because the caller's `> 0` check catches it — but the helper is exported through the barrel with a documented `NaN`-for-invalid contract, so the next caller will not be so lucky.
>
> **(b) Three queue-table tests were not updated after #3164's write gating merged in** — `price-changes-queue-table.test.tsx:126`, `:193`, `:366` call `renderWithProviders(<PriceChangesQueueTable />, { apiClient })` with **no `sessionAdapter`**, so the default session lacks `listings:write`, `write.visible` is false, and `row-accept` / `row-select` are correctly absent. Every sibling test in the file was updated to pass `createAuthenticatedSessionAdapter()`; these three were missed in the merge. The fix is one argument each.
>
> **(c) `clamps a manual price to the numeric(14,4) storage precision` asserts the behaviour the code deliberately replaced** — it expects `399.1235`, and `clampToStorablePrecision` now clamps to the currency's minor units (2 for PLN). The *code* is right; the test and its name are stale relative to the intentional change and were left behind.

**[IMPORTANT]** — `price-changes-queue-table.tsx:268`
> `offerAutomaticUndo` passes no `durationMs`, so the Undo rides the provider's 4 s default. Four seconds is the entire window in which an operator can reverse a switch to **publishing prices without review** — a decision that then persists silently on the connection until someone opens its settings. The rest of this stack treats reversible decisions as persistent affordances (the table's own ignored-row Undo is an in-row button with no timer). You now own `ShowToastOptions.durationMs`; give this one a materially longer window, or follow the in-row precedent.

**[IMPORTANT]** — `edit-price-change-dialog.tsx:79`
> The `>30%` deviation is still a **non-blocking, browser-only** rule, and there is still no upper bound anywhere: the server DTO is `@IsNumber() @IsPositive()`. The concrete 1000×-too-low path is closed (that was the urgent half), but `39900` for a `399` item still publishes on one click past a warning, and the only thing standing between an operator and a wrong live price is a sentence they can read past. #2610's precedent is explicit that a value which silently produces a wrong published price is refused **server-side as well as in the form**. This is a backend ask against #3162 rather than a change here — please raise it on the epic so it does not get lost with this PR.

**[IMPORTANT]** — `apps/web/src/index.css`
> Another **363** CSS lines with **zero** `@media` queries, on top of #3164's 310. The dialogs are the surface most likely to be operated from a tablet. Same gap I raised on the parent; flagging again because it is now compounding rather than static, and because the two PRs together are the whole feature's styling.

**[IMPORTANT]** — PR description, and `docs/plans/mockups/price-changes-review-queue.html`
> `docs/frontend-architecture.md § UX Mockups`: an implementing PR must touch the mockup or state why it is still accurate. This does neither. The description is also stale in a way that misreads the shipped code: the "Known, stated gap" paragraph still describes lanes that *"flip to 'done' together once the whole batch reaches a terminal status"* as static per-connection counts — the review fix replaced that with composition-only lanes precisely because the flip was a false claim. Refresh it before merge so the merge record matches the code.

**[SUGGESTION]** — `use-set-source-sync-mode-mutation.ts:43`
> Still no `onSuccess` invalidation — every other mutation in this feature has one. After an Undo, #3149's settings page and the connection detail page read a stale `pricingSync` config. The read-modify-write also carries no version guard, so it can clobber a concurrent edit made on the settings page; #3167 ships a sibling mutation that invalidates both `connectionPricingSyncQueryKey` and `priceChangesQueryKeys.all` — align now while it is one line.

**[SUGGESTION]** — `apps/web/src/index.css`
> 27 raw `px` values in the new rules where `--space-*` tokens exist. `check-design-tokens.mjs` is one-directional and cannot see this, so the "clean" claim in the test plan is true and unrelated.

**[SUGGESTION]** — no `bulk-publish-progress.test.tsx`
> The widget now has four distinct renderable states (loading, error, running, terminal-with-failures) and none has a direct test; the only coverage is the queue-table mounting test, which is currently one of the six failing. Worth a small file given how much of it changed.

### Not this PR's defect, but blocks the merge

`Integration Tests` is also red, entirely in `test/integration/listings-price-change-episode.int-spec.ts` (`invalid input syntax for type uuid: "no-such-source"` from `price-change-episode.repository.ts:536`) — backend code owned by the stack base, inherited through #3164. Noting it so the red is attributed correctly.

**Done well, beyond the fixes:** the `isCustomOverride` docblock is the kind of comment that stops the bug coming back — it names the wrong field, says what it actually means, and states the consequence. The `BulkPublishProgress` mounting note (sibling of the table's own empty branch, not nested inside it) catches a real second-order bug that a successful bulk publish would have triggered every single time, and would have read as "nothing is happening". And extracting the parse and the clamp into pure, separately-tested helpers is what makes finding (a) a two-line fix instead of an archaeology exercise.

### Verdict

❌ **Request changes**

1. Add the comma-grouping arm and the empty-string guard to `parseLocalizedAmount`, mirroring the dot branch.
2. Pass `sessionAdapter: createAuthenticatedSessionAdapter()` in the three queue-table tests at `:126`, `:193`, `:366`.
3. Update the clamp test (and its name) to the currency-minor-units behaviour the code now implements.
4. Give the automatic-opt-in Undo a longer window than 4 s.
5. Refresh the PR description and reconcile the mockup; raise the server-side price ceiling on the epic.

(1)–(3) turn the build green and are small. Happy to re-review as soon as `Test` is passing.

---

norbert-kulus-blockydevs:
Raised the server-side price-ceiling ask as **#3222**, against the epic (#3139) as requested, so it does not get lost with this PR.

One correction worth recording there rather than here: the DTO is not quite `@IsNumber() @IsPositive()` any more — a previous round added `@Max(1_000_000_000)`. That ceiling is documented in `edit-price-change.dto.ts` as a pathological-input guard sized to `numeric(14,4)` (so `1e30` fails with a 400 instead of a 500 at the database), explicitly *not* currency-aware and not a deviation guard. Your point stands unchanged — `39900` for a `399` item passes it trivially — and #3222 puts the proportionate refusal in `PriceChangesService.edit`, which is where the episode's own computed price is actually in hand.

The remaining findings on this PR are still open; this comment only closes out the one item you asked to be moved to the epic.

norbert-kulus-blockydevs:
All of it addressed — `15ba8be12` and `1096cc09b`, on the parent catch-up `a52cc328f`.

**BLOCKING — the red suite.** All three causes fixed, and your read of each was exact.

(a) `parseLocalizedAmount`. The comma branch now mirrors the dot branch's multi-separator grouping arm, so `"1,234,567"` parses instead of falling through to `Number('1.234.567')`. The empty-string case returns `NaN` via an early guard rather than leaking `Number('')`'s zero — you were right that the `> 0` check at today's only call site is luck, not a contract.

(b) The three sessionless renders. Two take `createAuthenticatedSessionAdapter()`. The third could not: *"never renders the 'also set to Automatic' checkbox for a non-admin session"* would have asserted nothing against the suite-wide default user, who is an **admin** with every permission. It now takes an explicit operator (`listings:read` + `listings:write`, role `operator`), which is the only session shape that makes that case mean what its name says — write-capable enough for the row action to exist, not admin enough for the checkbox.

(c) The clamp case. Expectation and name both updated to the currency's minor units.

**IMPORTANT — the 4s Undo.** Raised to 30s, with the reasoning recorded on the function rather than only in the commit: this is the single window in which an operator can reverse a switch to publishing prices without review, after which the setting persists silently on the connection. I kept the toast rather than following the in-row precedent because a bulk accept leaves no single row to hang an in-row Undo on.

**IMPORTANT — the server-side price ceiling.** Raised on the epic as **#3222**, as you asked. One correction recorded there rather than acted on here: the DTO is no longer bare `@IsNumber() @IsPositive()` — a previous round added `@Max(1_000_000_000)`. That ceiling is documented in the file as a pathological-input guard sized to `numeric(14,4)`, explicitly not currency-aware and not a deviation guard, so your point stands unchanged and #3222 puts the proportionate refusal in `PriceChangesService.edit`, where the episode's own computed price is in hand.

**IMPORTANT — the compounding `@media` gap.** Agreed it is one styling pass, so rather than open a second ticket I widened **#3223** to cover the price-changes feature's responsive treatment as a whole. Both `index.css` blocks now cite it by number and say to delete the paragraph together with the deferral.

**IMPORTANT — description and mockup.** Description rewritten; the "lanes flip to done together" paragraph is gone and replaced with what actually ships (composition-only lanes, and why). On the mockup: this PR's dialogs do not diverge from it — the price hero, rule sentence, opt-in and mini-rows render as drawn — so it is stated rather than touched here. The parent touched it for the one real divergence the pair introduces.

**SUGGESTION — `use-set-source-sync-mode-mutation`.** Added `onSuccess` invalidating `priceChangesQueryKeys.all`. I deliberately did **not** invalidate a pricing-sync key: on this branch the view is read imperatively inside `mutationFn`, so there is no cache of it to go stale. `connectionPricingSyncQueryKey` arrives with #3166, and the hook carries a note saying it must be invalidated here once it does — otherwise that settings page will keep rendering the mode this toast just changed. I will close that when I take #3166.

**One thing I checked and left alone:** `.toast__content` is declared twice in `index.css` now, which looked like the same class of defect as the parent's `.table-wrap`. It is not — the second block carries a comment naming the first and explaining the flex sizing the action slot needs, from an earlier round's finding 11. Deliberate and documented, so untouched.

Verified: `type-check` clean, `check:invariants` clean, 51/51 in `features/price-changes`.

piotrswierzy (approved):
## Tech Lead re-review (`1efff961` → `57d00431`)

Re-reviewed the delta. Strip out the `origin/main` merge and the merge-up from `#3164`, and the PR-authored work is two commits — `15ba8be1` and `1096cc09`. **Every finding from my previous pass is closed.** The `parseLocalizedAmount` defect is genuinely fixed (both halves), and — more importantly — the fix is fixed the right way round: the suite was repaired by making the *tests* correct, not by relaxing the assertions. The two stale queue-table cases take a real session, and the non-admin checkbox case takes an explicit **operator** rather than the suite-wide admin default, which is a strengthening: under the admin default that case would have asserted nothing about the thing it is named for.

- **BLOCKING (a) — `parseLocalizedAmount`** — closed, both halves, at `lib/price-change-copy.ts:145-187`. The comma branch now carries the multi-separator grouping arm the dot branch always had, and `''` short-circuits to `NaN` before `Number('')` can answer `0`. I simulated the shipped branch logic against the boundary set: `1,234,567 → 1234567`, `1.234.567 → 1234567`, `1,23 → 1.23`, `1 234,56 → 1234.56`, `399.99 → 399.99`, `'' → NaN`, and `1,234 → NaN` (the deliberate ambiguity refusal from the earlier round, correctly preserved — `isAmbiguousSeparator` only fires on *exactly one* separator with exactly three trailing digits, so it does not swallow the new grouping arm). Both of the cases I named have dedicated regression tests in `price-change-copy.test.ts` ("treats more than one comma as unambiguous thousands grouping", "returns NaN for empty or non-numeric input").
- **BLOCKING (b) — three queue-table cases with no `sessionAdapter`** — closed, and closed by adding the session rather than by removing the gating. `OPERATOR_SESSION_USER` (`listings:read` + `listings:write`, role `operator`) is the right fixture for the non-admin checkbox case, and the comment states why the admin default would have made it vacuous. That is the correct instinct about a test that can only fail for the wrong reason.
- **BLOCKING (c) — stale clamp assertion** — closed. `399.1235 → 399.12`, and the test *name* changed with it ("clamps a manual price to the CURRENCY's minor units, not the storage column's precision"). Renaming the case matters as much as the expectation here: left alone it would have kept advertising `numeric(14,4)` as the rule.
- **IMPORTANT — 4 s Undo on a switch to unreviewed publishing** — closed. `AUTOMATIC_OPT_IN_UNDO_MS = 30_000`, and the reasoning is recorded where the constant is: this is the only window in which a switch to publishing *without* review can be reversed, and past it the setting persists silently on the connection. The supporting change to `shared/ui/toast-provider.tsx` adds a per-toast `durationMs` with its own test rather than moving the provider default, which is the right blast radius.
- **IMPORTANT — `>30%` deviation browser-only** — closed, and I accept the resolution. The 30% signal is a **non-blocking warn**, not a gate (`edit-price-change-dialog.tsx:80-84`), so it was never a bypassable check — and a browser mirror that is *stricter* than the server is the failure mode the repo already has a rule about (#2240). Meanwhile the genuine server-side hole did get closed in the `#3162` round: `EditPriceChangeDto` now carries `@IsNumber({ maxDecimalPlaces: 4 })` + `@IsPositive()` + `@Max(MAX_MANUAL_PRICE_OVERRIDE)`, so `1e30` no longer reaches Postgres as an unhandled 500. Pathological input is bounded server-side; the 30% figure stays advisory, correctly.
- **IMPORTANT — 363 CSS lines with no media queries** — closed as a tracked deferral. `index.css:2160-2168` cites `#3223`, which was **widened** to cover the whole feature's responsive treatment rather than split across two tickets, and the note is honest that dialogs are the half most likely to be operated from a tablet.
- **IMPORTANT — mockup unreconciled** — closed on `#3164`, which owns the mockup file and now carries an explicit `KNOWN DIVERGENCE` block.

One unrequested improvement worth naming: `useSetSourceSyncModeMutation` now invalidates `priceChangesQueryKeys.all` on success, and the comment states precisely why that is the *only* key available today (the hook reads the pricing-sync view imperatively inside `mutationFn`, so there is no cache of it) **and** what `#3166` must add when the cached read lands. That is the right way to leave a known-incomplete invalidation.

### Findings

### CI

`Lint`, `Type Check`, `Docker Build Smoke Test` and `PHP Unit Tests` are green on `57d00431`. `Test`, `Build` and `Integration Tests` are still **queued** (duration `0` — not started; the runner is draining `#3164`'s 32-minute integration job), so I have not observed the web suite green on this head. I am approving on code evidence rather than on a green tick, because that evidence is specific: each of the three causes is verified fixed in the file, the parse branch is verified by simulation against the exact inputs that failed, and each has a named regression test. **If `Test` comes back red, treat this approval as withdrawn on that point** — the whole of my prior BLOCKING was that this PR's own suite was red, and a second red run would mean a fourth cause I have not found.

### Verdict

✅ **Approve**

---

norbert-kulus-blockydevs:
No findings to address — thanks. Recording one thing for the merge: this branch carries the `1884000000000` migration renumber through its merge from the base, so the `1881000000000` collision with #3201, #2069 and #2202 is resolved for the whole stack, not just for #3159.

piotrswierzy (approved):
## Tech Lead re-verification (head unchanged at `be3504c`)

**The written condition on my previous approval is RESOLVED, and the approval stands unconditionally: `Test` came back green (15m7s)**, so the point I said I would withdraw on does not arise. `Integration Tests` (32m49s) and `Type Check` also pass; `Build` and `Lint` are `queued` behind the runner backlog, never failed. `MERGEABLE`, no conflict.

Re-verified in code: the non-admin case takes an explicit `OPERATOR_SESSION_USER` rather than asserting nothing under the suite's admin default; `parseLocalizedAmount` carries its regression tests for all three cases (`1,234,567 → 1234567`, `'' → NaN`, `1,234 → NaN` as a deliberate ambiguity refusal, plus the `.`-separator mirrors); and the Undo toast runs on `AUTOMATIC_OPT_IN_UNDO_MS = 30_000` threaded through the new per-toast `durationMs`.

**#3236 does NOT promote the `>30%` warn to a gate, so the acceptance holds unchanged.** It adds a *second, deliberately wider* server-side bound (10×, `422`) aimed at a misplaced decimal point rather than at a judgement call, and it argues the containment explicitly — a mirror stricter than the gate would refuse work the destination accepts (#2240), so the server band must contain the form's, asserted by a test rather than left to whoever edits one constant next. The browser warning stays non-blocking with confirm enabled, which is what I approved. No behaviour change here, and no justification owed.

### Findings

### Verdict
✅ **Approve**

---

piotrswierzy (approved):
## Tech Lead re-verification (head unchanged at `abfd47d`)

Head unchanged, `main` unmoved. CI on run 34845739053: **Type Check, Docker Build Smoke Test, PHP Unit Tests and Detect changed paths pass; Build, Lint, Test and Integration Tests are queued behind the runner backlog. Nothing red** — the `Test` job that previously went green on this head (15m7s) has been re-queued by the new run, not failed. Re-confirmed the substance of the approval: the previously-red suite was fixed by correcting the **tests** (the non-admin case now takes an explicit operator session; under the suite's admin default it was asserting nothing), not by relaxing assertions; `parseLocalizedAmount` is the sole parse path on both the validate and the submit branch; the `>30%` deviation is still a **non-blocking warn** with confirm left enabled (`shows a non-blocking warning for a >30% deviation but keeps confirm enabled`). **#3236 does not promote it to a gate** — it adds a separate, wider 10× server-side bound, which leaves this verdict untouched.

Stack: base `3147-price-changes-tab-and-queue-table` (#3164). **#3159 and #3162 below still carry an open BLOCKING** (an in-place edit to the already-merged migration `1880000000000-create-price-change-episodes.ts`, diverging an already-migrated database from a fresh install). No migration here, but **the whole chain is held until those two are fixed.** `mergeable: MERGEABLE`, `mergeStateStatus: UNSTABLE` (pending checks only).

### Findings

### Verdict
✅ **Approve**

---

PR #3166, opened 2026-09-10
feat(web): connection Pricing & sync settings page

### LINKED ISSUE #3149: feat(web): connection Pricing & sync settings page

norbert-kulus-blockydevs:
## Summary

Part of #3139 (epic). Stack position: tenth — branches from #3148's branch (PR #3165, still open).

- Adds `PricingAndSyncSection`: a default pricing rule + mode switch, per-source overrides (every known source listed, override or not), an explicit Save/Discard bar that persists both the default and all overrides in one `PATCH /connections/:id/pricing-sync` call, and a pending-changes link to the Price changes tab.
- The pending-changes deep link targets the queue table's own namespaced `queueConn` search param (`pricing-and-sync-section.tsx` → `price-changes-queue-table.tsx`), so opening it from a connection lands on that connection's rows. **Corrected (#3166 round-3 review):** an earlier revision of this bullet described an `initialConnectionId` prop on `PriceChangesQueueTable` and a change to `ListingsListPage`. Neither exists — the post-review restructure replaced that approach with the URL-owned filter above, and the bullet was left behind, which cost an earlier review round a finding on filter state this PR does not contain.
- **Update (post-review restructure):** the section now lives on its own dedicated route (`/connections/:id/pricing-sync`, `connection-pricing-sync-page.tsx`), reached via a "Pricing & sync" button on the connection detail page — it is **no longer mounted inside `EditConnectionForm`**. This matches `docs/plans/mockups/price-changes-review-queue.html`'s own `#page-connection`, which is itself a standalone view, not a section embedded in a shared settings form. See "Deviation" below for why it moved.

## Deviation from the issue's stated precedent

The issue asks this to match `StockAndPricingSection` (#2610) "exactly." That component lives inside `EditConnectionForm`'s single react-hook-form and writes straight into `Connection.config`. #3146 shipped a dedicated `GET`/`PATCH /connections/:id/pricing-sync` endpoint instead (episode-shaped settings, not raw config), so `PricingAndSyncSection` owns its own fetch + local dirty state against that endpoint rather than joining the mega-form.

Mounting it inside `EditConnectionForm` alongside the legacy flat-shape `StockAndPricingSection` — which also edits `config.pricingRule`, in an incompatible shape — created two independent writers of the same config key on one page (a contradiction on first paint, a destructive checkbox toggle that could silently drop every per-source override, and a stale-merge hazard on the mega-form's own submit). Per review, this is fixed structurally: `PricingAndSyncSection` moved to its own route (the `SalesDocumentStatusSection` "one editable surface, linked to rather than duplicated" shape), `StockAndPricingSection` gained a `pricingRuleManagedElsewhere` flag that replaces its editable pricing-rule fields with a read-only link to the new page for that same population, and `EditConnectionForm.onSubmit` now deletes `pricingRule` from the merged config entirely for a viable pricing destination — the mega-form can no longer write that key at all for such a connection.

What *is* matched, per the issue's own wording, is the UX contract — explicit Save/Discard, no instant-apply on any toggle, one Save call for both the default rule and every per-source override together.

## Known gap, flagged per the issue's own assumption

"Recent activity" renders a documented "not available yet" empty state — no activity-log endpoint exists yet for #3145/#3146's price-change episodes (accept/ignore/edit history). Per the issue's stated assumption, this is flagged as a dependency gap rather than fabricated client-only history.

## Test plan

- [x] `pnpm --filter @openlinker/web lint` — clean (pre-existing warnings only)
- [x] `pnpm check:invariants` — clean
- [ ] `pnpm test` — could not run locally (environment constraint documented in commit messages); verified via targeted `vitest run` on every touched spec plus the broader `features/connections`, `pages/connections`, `features/price-changes`, `pages/listings` and `app/routes` suites
- `pricing-and-sync-section.test.tsx` covers: renders default rule + every known source; mode-switch toggle shows unsaved bar and saves default+overrides in one call; Discard reverts to last-fetched state; per-source override toggle copies-from-default on and drops on off, and its removal (unticking a persisted override) both surfaces a confirm dialog and actually persists the removal; margin ≥100% is refused client-side; write-access gating hides/disables controls for a non-writer; pending-changes link deep-links with `?connectionId=`

Closes #3149


norbert-kulus-blockydevs (commented):

Two things done right that I want to name: the deviation from `StockAndPricingSection` is **argued, not silently taken** — a 20-line docblock explains the reasoning, which is what made #3167's reversal a one-PR correction rather than an archaeology exercise. And the "Recent activity" gap is handled exactly right: *"Activity history isn't available yet"* is plainly not-a-dataset, with a comment naming the missing endpoint, rather than fabricating client-only history. The unsaved-bar buttons are `type="button"` so they cannot submit the enclosing form, and the mutation invalidates both query keys with a comment saying why.

Three blocking, inline. The first is the one that decides this PR's fate: **the section and `StockAndPricingSection` now write the same `config.pricingRule` key in incompatible shapes, on the same page**, with three independently reachable paths to silent data loss — one of which needs no click at all.

**Recommendation: do not merge as-is.** #3167 removes this PR's only user-facing contribution (the mount inside `EditConnectionForm`) and replaces it with a dedicated route, citing exactly the harm confirmed below. Either fold #3167's route into this PR, or land the component here unmounted and mount it there. Merging both in order also works but leaves `main` transiently in the data-loss state, and leaves the three dead `/connections/:id/pricing-sync` links from #3164/#3165 unclosed for a commit — links that issue #3149's own Dependencies section says land here.
---
*Part of a whole-stack review of #3158–#3168 (epic #3139). Cross-cutting items tracked once: `pnpm lint` is red on the stack; `docs/architecture-overview.md § Listings` is untouched by all ten PRs though ADR-072 names it the primary doc section; the mockup has not changed since #3155 despite ~20 behavioural divergences, which `docs/frontend-architecture.md § UX Mockups` treats as incomplete; `features/price-changes` is registered in **neither** `.eslintrc.js` `no-restricted-imports` group, so the new barrel is unenforced; and a `connections ↔ price-changes` value-import cycle is one barrel export from closing (#337/#359, #2640).*

apps/web/src/features/connections/components/EditConnectionForm.tsx
```diff
@@ -823,6 +824,14 @@ export function EditConnectionForm({ connection }: EditConnectionFormProps): Rea
         syncPricingRuleToJson={syncPricingRuleToJson}
       />
 
+      {/* #3149 — the opt-in recurring price propagation review queue's
+          per-connection settings. Gated the same way `needsMasterCatalog`
+          is: a viable destination is one that can either list marketplace
+          offers or publish shop products. Self-contained (its own fetch +
+          Save/Discard) rather than participating in this form — see the
+          component's own docblock for why. */}
+      {needsMasterCatalog ? <PricingAndSyncSection connectionId={connection.id} /> : null}
```
norbert-kulus-blockydevs, line 833: **BLOCKING — two writers of `config.pricingRule` on one page, in incompatible shapes.**

`StockAndPricingSection` (#2610) edits the **flat** `{type, percent, rounding}` through the mega-form; `PricingAndSyncSection` writes the **nested** `{default, sourceOverrides}` via #3163. Three independently reachable failure paths, all on one page load:

1. **Contradiction on first paint.** `readPricingRuleForm` (line 165) reads `raw.type` off the nested object → `undefined` → `type: ''`. `StockAndPricingSection` computes `priceOpen = Boolean(getValues('pricingRule.type'))` → `false`, renders its checkbox **unchecked** and says *"this destination is told … your catalogue price, unchanged"* — directly above a section reading *"Keeps a 22% margin…"*.
2. **Destruction on a stray click.** Tick-then-untick that checkbox → `syncPricingRuleToJson()` → `mergeStructuredIntoConfig`'s `if (… || !structured.pricingRule.type) { next.pricingRule = null }` → the textarea holds `"pricingRule": null`. Mega-form Save wipes the default **and every per-source override**. `priceSyncMode` is untouched, so an `automatic` connection then auto-publishes the raw catalogue price. That is real money.
3. **Silent revert with no click at all.** `onSubmit` sends `config: { ...fresh.config, ...input.config }` where `input.config` is the parsed **mount-time** textarea. The refetch protects only keys *absent* at mount. Save the pricing section, then rename the connection and press the mega-form Save → the pricing save rolls back.

There is also no navigation guard: the mega-form's success path calls `navigate('/connections/:id')`, discarding an unsaved pricing draft with no prompt.

**Fix: #3167's** — move the section to its own route and leave a link (the `SalesDocumentStatusSection` shape: one editable surface, linked to rather than duplicated).

apps/web/src/features/connections/components/pricing-and-sync-section.tsx
```diff
@@ -0,0 +1,460 @@
...
+        <div className="field">
+          <label className="field__label" htmlFor={`${idPrefix}-pct`}>
+            Percent
+          </label>
+          <div className="field__price-input">
+            <Input
+              id={`${idPrefix}-pct`}
+              inputMode="decimal"
+              value={String(rule.percent)}
+              onChange={(e) => onChange({ percent: Number(e.target.value) || 0 })}
```
norbert-kulus-blockydevs, line 438: **BLOCKING — a margin ≥ 100% is refused by neither the browser nor the server on this path.**

`docs/architecture-overview.md § Inventory` (#2610) states the rule: *"A margin of 100% or more is refused, not accepted and silently degraded … The refusal is **server-side as well as in the form**"*, because `applyPricingRule` does `percent >= 100 ? basePrice : …` so the operator "saves happily and publishes an unchanged price".

- **Browser:** `RuleFields` has zero validation — no `<100` check, no `min`, no Zod, no `aria-invalid`. Compare `StockAndPricingSection`, whose `pricingRuleFormSchema.superRefine` refuses it and whose field description spells out why.
- **Server:** `validateStockAndPricingConfig` destructures the **flat** shape; under the nested one all three values are `undefined` and every check is guarded on `!== undefined`, so it **returns without firing**. #3163's `PricingRuleDto` has only `@Min(0.0001)` — no ceiling.

So `margin: 150` saves cleanly through this section **and** through curl, and every published price silently reverts to the catalogue price.

**Also on this line:** `value={String(rule.percent)}` with `onChange={… Number(e.target.value) || 0}` round-trips through `Number()` on every keystroke. Typing `22.5`: `"22."` → `22` → re-renders as `"22"` and **the `.` is erased — decimals are untypeable.** Clearing the field yields `Number("") || 0` → snaps to `"0"`. `"-5"` passes straight through to a `@Min(0.0001)` 400. `StockAndPricingSection` solved this by keeping percent a **string** in form state with a regex — the comment there even records the regression that made them widen it.

apps/web/src/features/connections/components/pricing-and-sync-section.tsx
```diff
@@ -0,0 +1,460 @@
...
+  if (query.error || !draft) {
+    return (
+      <ErrorState
+        title="Couldn't load pricing & sync settings"
+        message={query.error?.message ?? 'Unknown error'}
+      />
+    );
+  }
+
+  const isDirty = query.data ? JSON.stringify(draft) !== JSON.stringify(query.data) : false;
```
norbert-kulus-blockydevs, line 108: **BLOCKING — no write-permission gating, and the unsaved bar never clears after a successful save.**

**Gating:** this component contains no `useWriteAccess`, no `ReadOnlyLock`, no `AccessGate`. Every sibling on the same page gates (`EditConnectionForm.tsx:383,873`, `ConnectionActionsPanel.tsx:33`, `EnableConnectionButton.tsx:44`), all on `connections:write` + `ReadOnlyLock`, per `§ Access Control`. Even the **mockup** disables every control and hides the unsaved bar under `state.readOnly`. Two victims: a **viewer** (GET is viewer-allowed) sees a fully editable form with a Save button; an **operator** gets a guaranteed 403, because #3163's PATCH is `@Roles('admin')` — surfaced as `error.message` in a raw toast.

**The bar:** the seeding effect is guarded `if (query.data && draft === null)`, so `draft` is populated **once, ever**, while `isDirty` compares `JSON.stringify(draft) !== JSON.stringify(query.data)` and the server recomputes derived fields the local mutation never touches. Flip the default mode → save → the server returns `sources[0].effective.mode = 'automatic'` while `draft.sources[0].effective.mode` is still `'manual'`. `isDirty` stays **true**: success toast fires and the bar remains, on the most common interaction the section offers. `isCustomOverride` and `openEpisodeCount` drift the same way, so any background refetch flips it on spuriously. The shipped test misses it because it asserts the `update` payload and never that the bar disappears.

**And it does not account for `customSources` at all** — see my #3167 review: that is why toggling a per-source override never enables Save, and why `pricing-and-sync-section.test.tsx:82` must be red in CI.

**Fix:** re-seed `draft` from the mutation's return value on success (or derive `const view = localDraft ?? query.data` and reset to `null`), include `customSources` in the dirty comparison, and add `useWriteAccess('connections:write', demoMode)` + `ReadOnlyLock`.

apps/web/src/features/connections/components/pricing-and-sync-section.tsx
```diff
@@ -0,0 +1,460 @@
...
+    setDraft((prev) => {
+      if (!prev) return prev;
+      const next = cloneView(prev);
+      if (mode) next.default.mode = mode;
+      if (rulePatch) next.default.rule = applyRulePatch(next.default.rule, rulePatch);
+      return next;
+    });
+  }
+
+  function toggleSourceCustom(sourceConnectionId: string, checked: boolean): void {
```
norbert-kulus-blockydevs, line 120: **IMPORTANT — unticking an override destroys a configured rule with no confirmation, while the summary states the opposite.**

`toggleSourceCustom(id, false)` deliberately leaves `source.effective` untouched, so the row renders `ruleSentenceFor(source.effective.rule)` + `' (using the default rule)'` — i.e. a source whose override was `margin 30` displays **"Keeps a 30% margin … (using the default rule)"** while the default is 22%. The sentence asserts a rule that will not be applied.

Then `handleSave` omits it from `sourceOverrides` and #3163 replaces the whole object, so the 30% override is **permanently gone**. No confirm, no warning, nothing in the copy saying "this will delete the rule".

**Fix:** render the *effective* rule (fall back to `draft.default.rule` when unticked), and require confirmation before a Save that drops a previously-persisted override.

**Related, and needs reconciling with #3165 before either ships:** `useSetSourceSyncModeMutation` writes `sourceOverrides[id] = source.effective` for **every** source in the view, including ones that were inheriting. One "also set to Automatic" click therefore pins **every** source to a frozen copy of the current default — after which changing the default in this section changes nothing for anyone, and this section renders every checkbox ticked. This is the surface where that becomes visible.

apps/web/src/features/connections/components/pricing-and-sync-section.tsx
```diff
@@ -0,0 +1,460 @@
...
+            <p className="settings-section__desc">
+              Used for every source below, unless you give one its own rule.
+            </p>
+          </div>
+        </div>
+
+        <div
+          className="mode-switch"
+          id="conn-mode-switch"
+          role="radiogroup"
```
norbert-kulus-blockydevs, line 208: **IMPORTANT — hand-rolled radiogroup: selection is conveyed by colour alone, and `SegmentedControl` already exists.**

`<div role="radiogroup">` whose children are plain `<button>`s with **no `role="radio"` and no `aria-checked`**. A screen reader announces "Manual review, button" / "Automatic, button" with no indication of which is active — the only signal is `.mode-switch__opt.is-active { background: var(--accent-primary) }`. That is status carried by colour alone (WCAG 1.4.1/4.1.2), and the style guide's own rule for `StatusBadge` is *"Dot + text; **never color alone**"*. It is also two tab stops with no Arrow/Home/End.

`shared/ui/segmented-control.tsx` **is** this control, and its docblock says so: *"Uses the ARIA radiogroup idiom (`role="radiogroup"` + `role="radio"` / `aria-checked`) … a roving tabindex keeps the group a single tab stop."* Four instances are hand-rolled here, plus ~40 lines of new global CSS to restyle a primitive that already exists.

Separately, `.mode-switch__opt` is `padding: 6px 14px` at `0.78rem` ≈ 28 px tall, and the global coarse-pointer floor targets named selectors, not this new class — so the ≥ 44 px tap-target rule is unmet on touch.

apps/web/src/features/connections/components/pricing-and-sync-section.tsx
```diff
@@ -0,0 +1,460 @@
...
+    }
+  }, [query.data, draft]);
+
+  if (query.isLoading) {
+    return <LoadingState title="Pricing & sync" message="Loading settings…" />;
+  }
+  if (query.error || !draft) {
+    return (
+      <ErrorState
+        title="Couldn't load pricing & sync settings"
```
norbert-kulus-blockydevs, line 102: **IMPORTANT — every *successful* load flashes a `role="alert"` "Unknown error".**

The seeding `useEffect` runs after paint, so on the first render where `isLoading` is false and `error` is null, `draft` is still `null` and this renders `<ErrorState title="Couldn't load pricing & sync settings" message="Unknown error" />`. `ErrorState` carries `role="alert"` (`feedback-state.tsx:54`), so **a screen reader announces a failure on every successful load**.

This is the four-read-states rule inverted — a successful read rendered as a failure. Derive rather than mirror (`const view = localDraft ?? query.data`), or distinguish `draft === null && !query.error` as "still hydrating".

apps/web/src/features/connections/components/pricing-and-sync-section.tsx
```diff
@@ -0,0 +1,460 @@
...
+            );
+          })}
+        </div>
+      </div>
+
+      <div className="settings-section">
+        <div className="settings-section__pending" id="conn-pending-row">
+          {totalPendingChanges > 0 ? (
+            <>
+              <span id="conn-pending-text">{totalPendingChanges} changes waiting</span> —{' '}
```
norbert-kulus-blockydevs, line 371: **SUGGESTION — "1 changes waiting", and the heading level is wrong.**

No singular branch; the mockup has `' change' + (n === 1 ? '' : 's')`.

Separately, the four bare `<h2>`s in this component render *below* the page's `<h3 className="section-title">{connection.name}</h3>`. Every sibling section uses `<h3 className="rate-limit-section__title">` (`rate-limit-section.tsx:135`, `sales-document-status-section.tsx:86`), and #3167's replacement uses `h3` — another sign the placement, not the markup, is what is wrong.

apps/web/src/index.css
```diff
@@ -24087,3 +24087,198 @@ legend.analytics-settings-dialog__label {
...
+
+.source-row__body {
+  display: flex;
+  flex-direction: column;
+  gap: var(--space-3);
+  padding-top: var(--space-2);
+  border-top: 1px dashed var(--border-subtle);
+}
+
+.activity-list {
```
norbert-kulus-blockydevs, line 24280: **IMPORTANT — `.activity-list` collides with an existing definition, and the comment above this block claims collisions were checked.**

The added comment asserts the mockup class names *"were only ever referenced in a comment, never defined"*. True for `.settings-section`; **false for `.activity-list`**, which is defined at `index.css:2621`/`2628` (`display: grid; gap: 0.75rem; list-style: none`), with `.activity-list__item` rules at 2756–2780 and a responsive block at 4642. This rule sits later at equal specificity and **wins**, re-declaring `display`/`gap` for the pre-existing selector.

No live TSX consumer today, so no immediate regression — but the `__item` rules remain, so the first follow-up that renders the real activity feed inherits a flex-column container fighting flex-row items.

The cited #2473 precedent did the **opposite** of what this block does: it *namespaced* the mockup's names (`.analytics-settings-dialog__field` for the mockup's `.settings-field`). Here `.settings-section`, `.mode-switch`, `.source-row`, `.unsaved-bar`, `.source-list`, `.rule-editor-*` are all claimed globally, including a bare descendant selector `.settings-section h2`.

Also `.settings-section:first-of-type` never matches — `:first-of-type` is per **tag**, and `EditConnectionForm` renders `<div className="panel__header">` earlier among the same siblings, so the "no top border on the first section" rule silently never applies.

**Fix:** namespace the block (`.pricing-sync__*`) per the precedent, and correct the comment.

piotrswierzy (changes requested):

Choosing the dedicated `/pricing-sync` endpoint over stuffing `sourceOverrides` into raw `Connection.config` JSON is right, and the deviation-from-#2610 reasoning in the docblock is honest. But the component has a hard functional defect.

### BLOCKING — per-source overrides cannot be saved, and an existing override cannot be removed at all

`pricing-and-sync-section.tsx`. Verified by reading the diff:

- `customSources` is separate `useState` (254)
- `isDirty = query.data ? JSON.stringify(draft) !== JSON.stringify(query.data) : false` (280) — **blind to `customSources`**
- `{isDirty ? <div className="unsaved-bar">…Save changes…` (431) is the **only** Save affordance

And `toggleSourceCustom(id, false)` deliberately does not touch `draft` (your comment: *"Unchecked: leave `effective` as-is for display"*), while `toggleSourceCustom(id, true)` assigns `cloneSetting(next.default)` — structurally identical to what was there for any source the server reports as `isCustomOverride: false`. Traced numerically:

```
on mount:            isDirty false
after toggle ON:     isDirty false
after toggle OFF:    isDirty false
```

So removing an override is impossible by any path, and adding one is a silent no-op unless the operator also edits a field.

**Your own fourth test cannot be passing** — it clicks `source-custom-toggle` then `getByRole('button', { name: 'Save changes' })`, which throws when the button isn't rendered. Fix: derive dirtiness from `customSources` too (or set `source.isCustomOverride = checked` in the draft, which `toggleSourceCustom` currently leaves stale — the checkbox reads `customSources` while the draft still says `false`).

### BLOCKING — no ≥100% margin refusal, and a decimal percent is untypeable

`RuleFields`. `Number(e.target.value) || 0` across four inputs:

```
""    -> 0     (clearing silently writes a real 0)
"12." -> 12    (echoed back as "12" — 12.5 is UNTYPEABLE, the field is fully controlled)
"-5"  -> -5    (negative accepted)
"abc" -> 0     (silently zeroed)
"120" -> 120   (margin >= 100 accepted)
```

#2610 is explicit: *"A margin of 100% or more is refused, not accepted and silently degraded … The refusal is **server-side as well as in the form**"* — because `applyPricingRule` falls back to the base price at ≥100, so an operator who typed 120 saves happily and publishes an unchanged price. `StockAndPricingSection` **on this very page** already refuses it with copy written for it. The server half is #3146's; the form half is this PR's and is missing.

### BLOCKING — two pricing-rule editors on one page, writing to two stores, with no precedence

`StockAndPricingSection` is untouched and still edits `config.pricingRule` (`syncPricingRuleToJson` still passed at `EditConnectionForm.tsx:823`); `PricingAndSyncSection` mounts 8 lines later and writes `default.rule` through the new endpoint. An operator can set *margin 22%* in one and *markup 10%* in the other, **on the same screen**, and nothing reconciles them.

ADR-072's own Cons name the obligation: *"the shape change requires **every existing reader** to be updated with a backward-compatible fallback."* `StockAndPricingSection` is an existing reader **and writer** and was not updated. Either it loses its pricing half and links out, or state and enforce precedence. (#3167 fixes the *placement* but not this.)

Honest half: neither section displays `stockSafetyBuffer`, so **#2323's single-reader invariant is not violated.** Good.

### IMPORTANT

- **`.activity-list` is an existing global class and this redefines it.** The PR comment claims no collisions were found — true for `.settings-section`, **false for `.activity-list`**, declared at `index.css:2031`/`:2039` as a shared list primitive with real `__item` rules at 2165–2189 and a media override at 4040. The new `display: flex` appended at 24087 wins on source order and silently overrides `display: grid` app-wide, leaving `__item` styled for a parent it no longer has. Latent (no `.tsx` uses it today) — which is why the next consumer discovers it. Rename to `.pricing-sync-activity-list`.
- **`.settings-section` is a global land-grab** for one feature, on a page that already hosts three sections that don't use it.
- **No `useWriteAccess` / `ReadOnlyLock` / `usePermission`** anywhere in the diff, on a section that is *nothing but* write affordances — two mode switches, a rule form, per-source checkboxes, Save/Discard. The mockup even has a `Read-only` preview state this PR doesn't implement.
- **The draft never re-syncs.** The guard is `if (query.data && draft === null)`, so after a successful save the refetch lands but `draft` keeps its pre-save shape — if the server normalises anything, `isDirty` flips **true and stays true**, leaving a permanent "unsaved" bar after a successful save. And `JSON.stringify` equality is key-order sensitive: if the API returns `{rule, mode}` rather than `cloneSetting`'s `{mode, rule}`, `isDirty` is true on mount before any edit.
- **Filter state is URL-*seeded*, not URL-*owned*** — `initialConnectionId` feeds `useState`'s initial value only, so clicking another chip leaves the URL saying `connectionId=X` while the table shows something else. § URL State lists filters first. The body says the table *"had no URL-driven filter state before this"*; it still hasn't.
- **`price-changes` is absent from both `.eslintrc.js` pattern groups** (23 slugs registered). This PR creates the stack's first cross-feature consumer, so the guard is silently open on the newest feature in the app.
- **Two unsynchronised Save scopes** — mounting a section that owns its own fetch and dirty state inside `EditConnectionForm` means submitting the outer form or navigating away discards this draft silently. (#3167 fixes this; #3166 could merge alone.)
- **§ UX Mockups unmet at PR level.** The code header cites `#page-connection` — real credit — but the body doesn't, and the implementation materially deviates (mockup renders a standalone page with one page-level Save bar; this embeds in `EditConnectionForm`).

### SUGGESTIONS

`needsMasterCatalog` is the right *set* and correctly uses `enabledCapabilities`, but it answers a different question — derive a locally-named `supportsPriceSync` so a change to the master-catalog gate can't silently toggle this section. And *"No price changes waiting on this connection."* asserts a positive absence from a sum over a possibly-empty array.

**Done well:** the mutation invalidates both `connectionPricingSyncQueryKey` and `priceChangesQueryKeys.all` with a comment on why the second matters; `useEffect` sits above the early returns so hook order is stable; the "Recent activity" empty state is flagged as a backend gap rather than fabricated.

### ⚠️ Stack-wide — no CI has run on any of these branches

`gh pr view --json statusCheckRollup` returns **0 checks** for #3155, #3164, #3165, #3166, #3167 and #3168. Three PR bodies leave `pnpm test` unchecked and say *"relies on CI"* — there is no CI to rely on. This is the highest-leverage fix for the whole stack: it would have caught #3166's red test on the first push. Worth checking whether the workflow filters branches by name.

---

norbert-kulus-blockydevs:
Addressed all 7 review comments (rebased onto the just-fixed #3165 tip; the `readPricingRuleForm` nested-shape read at `EditConnectionForm.tsx:175` was already correct going into this pass).

## 1. BLOCKING — two writers of `config.pricingRule` (highest-severity finding)

Fixed **structurally**, going with the review's own suggested direction rather than a UI-only patch:

- **`PricingAndSyncSection` moved to its own dedicated route** — `/connections/:id/pricing-sync` (new `connection-pricing-sync.route.tsx` / `connection-pricing-sync-page.tsx`), reached via a new "Pricing & sync" button on the connection detail page. It no longer shares a page (or a form submit) with the legacy flat-shape editor at all, so the contradiction-on-first-paint failure mode is now impossible by construction — the two sections are never mounted together.
- **`StockAndPricingSection` gained a `pricingRuleManagedElsewhere` flag.** For any viable pricing destination (`OfferManager`/`ProductPublisher`), its own editable pricing-rule checkbox/fields are replaced with a read-only pointer to the new page. That population's pricing rule can no longer be edited — let alone destructively wiped — from the old surface at all.
- **`EditConnectionForm`'s `onSubmit` now deletes `pricingRule` from the merged config** for that same population, before the `{...fresh.config, ...input.config}` merge. So even a stale-snapshot mega-form submit (the third failure mode — rename the connection, hit Save, roll back a concurrent pricing save) can never touch that key again for a destination connection. All three interactions are closed at the same time, by removing every remaining path that could write the key from that form.

## 2. BLOCKING — margin ≥100% validation + percent input

- Added client-side validation mirroring the already-fixed server-side `PricingRuleMarginCeilingConstraint` verbatim — Save is disabled with an inline error when a margin rule's percent is ≥100.
- Fixed the percent input to round-trip through a string-typed draft (matching `StockAndPricingSection`'s own established fix for the identical regression) instead of `Number(text) || 0` on every keystroke — a decimal point is now typeable and a cleared field no longer snaps to `"0"`.

## 3. BLOCKING — write-access gating + stale unsaved-bar

- Added `useWriteAccess('connections:write', demoMode)` gating (the PATCH is admin-only). An operator now sees every control correctly disabled instead of hitting a guaranteed 403; a genuine viewer (GET is viewer-allowed) sees the real settings read-only rather than a fully-editable form. Save/Discard wrapped in `ReadOnlyLock` for the demo-mode case.
- Fixed the unsaved-bar staleness bug: the draft now re-seeds from the mutation's own successful response (not only the initial `query.data` snapshot), and the dirty-check separately tracks the `customSources` toggle set — the old JSON-diff never compared that against server state, so toggling an override checkbox alone never enabled Save.

## 4. IMPORTANT — unticking an override

Unticking now displays the **default** rule (never the stale about-to-be-discarded value), and Save on a persisted override opens a confirm dialog naming what will be lost rather than silently dropping it. Reconciled against the already-fixed `useSetSourceSyncModeMutation` (#3165) — verified this component's own save path already filters to `isCustomOverride`-only entries the same way.

## 5. IMPORTANT — accessible mode switch

Replaced all 4 hand-rolled `<button>` radiogroups with the shared `SegmentedControl` primitive (`role="radio"` + `aria-checked`, roving tabindex, ≥44px targets), deleting ~40 lines of duplicate `.mode-switch*` CSS.

## 6. IMPORTANT — loading-flash error state

Fixed the derivation order so a successful load never briefly renders `ErrorState`'s `role="alert"` — the error check now runs before the "still hydrating" gap.

## 7. SUGGESTION — pluralization, heading level, CSS

Fixed "1 changes waiting" → "1 change waiting"; changed the four bare `<h2>`s to `<h3>` matching every sibling section; namespaced the whole CSS block under `.pricing-sync__*` (it collided with a pre-existing `.activity-list` definition — the removed comment's "never defined" claim was false for that class); replaced the never-matching `.settings-section:first-of-type` with an explicit `.pricing-sync__section--first` modifier.

## Also fixed along the way

The new page's "N changes waiting — review them" deep link now targets `queueConn` (the queue table's own namespaced filter param, #3164) instead of `connectionId` (the "All listings" tab's unrelated channel selector) — the two were namespaced apart specifically so neither silently pre-filters the other, and the original link target would have silently done nothing.

## Verification

`pnpm --filter @openlinker/web lint` (0 errors, 123 pre-existing warnings unrelated to this change), `pnpm --filter @openlinker/web type-check` (clean), `pnpm check:invariants` (clean, including `check-css-structure` on the rewritten stylesheet block), and targeted `vitest run` across every touched spec plus the broader `features/connections`, `pages/connections`, `features/price-changes`, `pages/listings` and `app/routes` suites (all green; bumped the pinned `EXPECTED_LAZY_ROUTE_COUNT` for the one new lazy route).

norbert-kulus-blockydevs:
## Re-verification of Piotr's `/pr-review` (2026-09-11T10:47:22Z)

Piotr's review is **stale by ~9 hours**: a major restructuring fix (`5e3886115`, pushed 2026-09-11T19:50:15+02:00) landed after his review and moved `PricingAndSyncSection` off the shared `EditConnectionForm` mega-form onto its own dedicated route entirely, which resolves most of his findings structurally rather than by patching. I verified every finding against the code as it stands (not against Piotr's line references, which describe the pre-restructure state), then rebased this branch onto the just-updated `3147-price-changes-tab-and-queue-table` tip (`b7f763e94`, itself a separate review-fix pass) and manually resolved a real 3-hunk merge conflict plus 3 stranded tests that surfaced only after that rebase. Details below.

### BLOCKING — per-source overrides couldn't be saved / removed
**Already resolved** by `5e3886115`. `hasUnsavedChanges = isDirty || isCustomSourcesDirty` (`pricing-and-sync-section.tsx:277`) now derives dirtiness from BOTH the rule draft and the `customSources` toggle set — the Save bar renders on either. `handleSave` also gates on `overridesAboutToBeDropped()` and opens a `ConfirmDialog` before silently discarding a persisted override on uncheck. Verified against the actual `pricing-and-sync-section.test.tsx` (lines ~150-208): it exercises toggle-on → save → toggle-off → save-refused-with-confirm → discard-and-save, asserting the exact `update` payload at each step. Ran it: passes (see verification below).

### BLOCKING — no ≥100% margin refusal, decimal percent untypeable
**Already resolved**. `DraftPricingRule.percent` is now string-typed end to end (`pricing-and-sync-section.tsx:99-103`), converted to a number exactly once at save (`toWireRule`). `validateRule` mirrors the server's `PricingRuleMarginCeilingConstraint` verbatim: refuses `margin >= 100` with actionable copy, and the Save button is `disabled={... || hasInvalidRule}`.

### BLOCKING — two pricing-rule editors on one page, writing to two stores
**Already resolved, structurally rather than patched**. Confirmed all three legs:
- `PricingAndSyncSection` now lives at its own route (`/connections/:id/pricing-sync`, `connection-pricing-sync-page.tsx`), not inside `EditConnectionForm`.
- `StockAndPricingSection` gained `pricingRuleManagedElsewhere` (true whenever `needsMasterCatalog`), which replaces its editable pricing-rule fields with a read-only link to the new page (`stock-and-pricing-section.tsx:227-240`).
- `EditConnectionForm.onSubmit` now `delete mergedConfig.pricingRule` for that same population (`EditConnectionForm.tsx:706-708`), so the mega-form's own submit can never clobber a concurrent save on the dedicated page.

The two writers can no longer coexist on one screen at all — this is a stronger fix than reordering/synchronizing them.

### IMPORTANT findings — verified one by one
1. **`.activity-list` CSS collision** — resolved. All new selectors are namespaced `.pricing-sync__*` (incl. `.pricing-sync__activity-list`), verified against `index.css`; the pre-existing global `.activity-list` block (lines 2654/2662) is untouched.
2. **`useWriteAccess`/`ReadOnlyLock`** — present (`pricing-and-sync-section.tsx:210-211`, wraps Save/Discard in `ReadOnlyLock`, disables every control when `readOnly`).
3. **Draft never re-syncs after save** — resolved: a second `useEffect` re-seeds `draft`/`customSources` from `updateMutation.data` on `isSuccess` (lines 236-247), which also structurally fixes the key-order-sensitivity concern (`isDirty` now compares two independently-normalized `toDraftView(...)` outputs, both with the same literal key order, rather than raw server JSON).
4. **URL-seeded-not-owned filter state (`initialConnectionId`)** — confirmed **not present anywhere in this PR's files** (`grep -rn "initialConnectionId" apps/web/src` is empty); this was #3164/#3147's queue-table finding, already addressed independently on that branch's own review pass.
5. **`price-changes` absent from `.eslintrc.js` pattern groups** — genuinely still an open gap on this branch at review time (the registration ships in `b7f763e94` on the parallel `3147` branch, not yet an ancestor of `3149`/base `3148` at that point). Rather than duplicate the registration by hand, I rebased this branch onto `3147`'s current tip so it's inherited verbatim (see "Stack rebase" below) — confirmed present in both pattern groups post-rebase.
6. **Two unsynchronised Save scopes** — moot: the section is no longer mounted inside `EditConnectionForm` at all, so there's only one Save scope now (stronger than #3167's stated fix).
7. **Mockup citation** — updated the PR description to state the standalone-page shape explicitly matches `docs/plans/mockups/price-changes-review-queue.html`'s `#page-connection` (itself a standalone view, not an embedded section) and to explain why it moved.

### SUGGESTIONS
- `needsMasterCatalog` reused as-is for the pricing gate rather than renamed to a locally-scoped `supportsPriceSync` — left unaddressed (suggestion-level, and the two questions are in practice identical: master-catalog-eligible == pricing-destination-eligible today). Flagging for awareness rather than fixing.
- "No price changes waiting…" copy — this text is only reachable after the component's `query.error` early-return, so a failed read can never land on it; the positive-absence concern doesn't apply given the render order.

### Stack rebase + a real merge conflict (post-review discovery, not in Piotr's review)
Per the task brief I checked whether `3149`'s parent chain was current. It was stacked correctly on `3148` (per GitHub's own base ref), but `3148` itself hadn't picked up `3147`'s own later review-fix commit (`b7f763e94`, which includes the `price-changes` eslint registration above). I rebased `3149` onto `origin/3147-price-changes-tab-and-queue-table` directly (a clean, no-rewrite fast-forward relationship — `3147`'s new tip is a pure descendant of the commit `3148` originally branched from, so this replays `3148`'s and `3149`'s own commits without altering their content).

This surfaced one real 3-hunk conflict in `price-changes-queue-table.tsx`, because `3147`'s fix added `expectedVersion`-per-item write-access gating to the OLD (pre-dialog) `handleBulkAccept`, while `3148`'s own commit replaced that whole flow with the dialog-based `handleBulkAcceptConfirm` + opt-in-to-automatic pairing. Resolved by hand: kept the dialog flow, added `write={write}` gating to `ActionCell` and `ReadOnlyLock`-wrapped bulk buttons from `3147`'s fix, and added the required `expectedVersion: item.version` to each bulk-accept item (a field the dialog-based flow's own commit predates and never had). No conflict markers remain; `pnpm check:invariants`' `check-css-structure`/every other gate confirms structural integrity.

That merge was clean at the code level but left **3 pre-existing tests stranded**: they render `<PriceChangesQueueTable />` with no explicit session, which pre-gating always had full write access by default. Post-gating (now merged in), the default test session has no `listings:write`, so every row action/checkbox vanished — failing on `getByTestId('row-accept')` / `getAllByTestId('row-select')`, unrelated to anything this branch's own commits changed. Fixed in `27ae3204a` by giving each an explicit write-capable session (two reuse the existing `ADMIN_SESSION` fixture; one — which specifically asserts the admin-only "set to Automatic" checkbox is absent — needed a new write-capable-but-non-admin `OPERATOR_SESSION` fixture, since that checkbox gates on `useIsAdmin()` independent of the `listings:write` permission).

### Verification (all run against the rebased worktree, correctly `cd`-scoped — confirmed via each command's own printed working directory)
- `pnpm --filter @openlinker/web lint` — 0 errors, 123 pre-existing warnings (none in touched files)
- `pnpm check:invariants` — clean, including `check-css-structure` and every mirror gate
- `vitest run` across `features/price-changes`, `features/connections`, `pages/connections`, `app/routes`, `pages/listings` — **55 files / 746 tests, all green** (0 failures, after the 3 test fixes above)

PR description updated to reflect the standalone-page restructure and mockup alignment (was still describing the pre-restructure embedded-in-`EditConnectionForm` shape).

norbert-kulus-blockydevs:
## Independent re-verification of the follow-up fix pass

I re-verified the 2026-09-11T22:56:12Z comment's claims against the code as it currently stands (not against its own summary), after merging in `3148`'s just-landed locale-parsing fix and resolving the resulting conflicts. Confirmed by direct grep/read, not by trusting the prose:

**All 3 BLOCKING items are genuinely resolved:**
- Per-source overrides: `hasUnsavedChanges = isDirty || isCustomSourcesDirty` (`pricing-and-sync-section.tsx:286`), `overridesAboutToBeDropped()` + `ConfirmDialog` gate an uncheck-then-save.
- ≥100% margin + decimal percent: `DraftPricingRule.percent` is string-typed, `validateRule` refuses `margin >= 100` verbatim against `PricingRuleMarginCeilingConstraint`, Save is `disabled={... || hasInvalidRule}`.
- Two writers of `config.pricingRule`: `PricingAndSyncSection` now mounts only at its own route (`connection-pricing-sync-page.tsx`, never inside `EditConnectionForm`); `StockAndPricingSection` replaces its editable pricing fields with a read-only link via `pricingRuleManagedElsewhere`; `EditConnectionForm.onSubmit` deletes `pricingRule` from the merged config for that population before submit. All three legs confirmed present in the current diff.

**IMPORTANT items** — all confirmed as claimed: CSS is namespaced `.pricing-sync__*` (24 occurrences), the pre-existing global `.activity-list` block is untouched; `useWriteAccess`/`ReadOnlyLock` gate every control; a second `useEffect` re-seeds `draft`/`customSources` from the mutation's own successful response; `initialConnectionId` is absent from every file in this PR; `price-changes` is registered in both `.eslintrc.js` pattern groups; the section is mounted in exactly one place, so there's only one Save scope.

**Merging in `3148`'s newer commit surfaced one real type error** the prior pass's own type-check run predates: `toWireRule`'s declared `PricingRule` return type (now correctly widened to optional `percent`/`rounding` per `3148`'s actual backend-contract fix) broke its use as `ruleSentenceFor`'s argument, which needs the required-fields `PriceChangeRuleSummary`. Fixed by retyping `toWireRule`'s return as `PriceChangeRuleSummary` (exported from the `price-changes` barrel) — it always constructs concrete values, so this is a correctness fix to the type, not a behavior change. Also fixed `toDraftRule`, which fed `PricingRule.percent`/`.rounding` straight into a controlled input without defaulting the now-optional fields.

**Verification** (after `pnpm install`, since this worktree had no `node_modules`):
- `pnpm --filter @openlinker/web lint` — 0 errors, 123 pre-existing warnings (none in touched files)
- `pnpm check:invariants` — clean, all 60+ gates including `check-css-structure`

The stack-wide "no CI has run" note is still true as of this check — `gh pr view --json statusCheckRollup` on this PR still returns an empty check list.

No other findings. Pushed as `9af4faba8` (merge commit `183fee71a` + the type fix `9af4faba8`) on top of the rebased `3148` tip.

norbert-kulus-blockydevs:
Caught this branch up with the upstream CI-fix commits from the parent (`3148-price-changes-dialogs`): the missing `PriceChangeDetectionService` barrel export and the corrected lane-count test assertion. Merged (not rebased) via `96569c8a2` — clean merge, no conflicts. Verified with `pnpm --filter @openlinker/web lint`/`type-check`, `@openlinker/core`, `@openlinker/worker`, and `@openlinker/api` `type-check`, plus `pnpm check:invariants` — all green.

piotrswierzy (commented):
## Tech Lead review

Re-reviewed at `96569c8`, against the code rather than the prior review's line numbers. **All three BLOCKING findings from the earlier `/pr-review` are genuinely fixed, and fixed structurally rather than patched** — I verified each by reading: the two-writer hazard is gone because the section moved to its own route, `StockAndPricingSection` hides its pricing half behind `pricingRuleManagedElsewhere`, and `EditConnectionForm.onSubmit` now `delete`s `mergedConfig.pricingRule` for that population, so the mega-form cannot write the key at all; `percent` is string-typed end to end with `validateRule` mirroring the server's margin ceiling and gating Save; `useWriteAccess('connections:write', demoMode)` + `ReadOnlyLock` gate every control, and since `ROLE_PERMISSIONS.operator` carries only `connections:read` (`libs/core/src/users/domain/types/role.types.ts:105`), that permission really is the admin-only gate the docblock claims. The CSS is fully namespaced under `.pricing-sync__*` with no raw colors and a real breakpoint, the four hand-rolled radiogroups are now `SegmentedControl`, the route declares the same crumb its four sibling connection routes declare, `features/price-changes` is registered in both `.eslintrc.js` pattern groups at this head, and the test file has a case per review finding. Remaining items are all non-blocking; the most important is a two-click dead end where Save is disabled with the reason hidden behind a collapsed disclosure.

### Findings

**[IMPORTANT]** — `apps/web/src/features/connections/components/pricing-and-sync-section.tsx:298-302, 452, 458-468, 490`
> Save can be disabled with no visible reason. `hasInvalidRule` gates `disabled` on the Save button, but the message that explains it renders only inside `<RuleFields>`, which renders only when `ruleFormOpen`. "Edit default rule" is a **toggle** (`setRuleFormOpen((v) => !v)`), so this is two clicks away: open the form → set a margin of `120` → error shows, Save disabled → click "Edit default rule" again → the form collapses, the unsaved bar stays, Save stays disabled, and nothing anywhere says why. `docs/frontend-architecture.md § Async UX Conventions` requires an "error state with actionable messaging"; an inert disabled control whose reason is behind a collapsed disclosure is not actionable. Either force `ruleFormOpen` while `hasInvalidRule`, or surface the message in the unsaved bar beside the disabled button.

**[IMPORTANT]** — the "viable pricing destination" predicate is now spelled four times, and drift is silently destructive
> `EditConnectionForm.tsx:482` (`needsMasterCatalog`), `connection-pricing-sync-page.tsx:27` (`DESTINATION_CAPABILITIES`), the inline `includes('OfferManager') || includes('ProductPublisher')` added to `connection-detail-page.tsx:251`, and transitively `pricingRuleManagedElsewhere={needsMasterCatalog}`. They agree today — I checked. But this PR changes the consequence of them ever disagreeing: `onSubmit` now **deletes** `config.pricingRule` for the `needsMasterCatalog` population, while the only surviving editor renders for the `DESTINATION_CAPABILITIES` population. Widen one and not the other and a connection loses its pricing rule on any unrelated connection save, with no editor anywhere to restore it and nothing logged. This is exactly the `supportsPriceSync` suggestion from the prior review, promoted because the deletion arm makes it load-bearing: export one predicate from `features/connections` and have all four sites call it.

**[IMPORTANT]** — the PR description contradicts the shipped code
> The second summary bullet still says *"`PriceChangesQueueTable` gained an `initialConnectionId` prop and `ListingsListPage` now threads its `?connectionId=` search param into it."* Neither is true at this head: `grep` finds no `initialConnectionId` anywhere, `ListingsListPage` is not in the diff, and the deep link correctly targets the queue table's own namespaced `queueConn` param instead (`pricing-and-sync-section.tsx:588`, `price-changes-queue-table.tsx:124`). The "Update (post-review restructure)" block was added without correcting the bullet above it — which is what led the previous review to spend a finding on URL-seeded-vs-URL-owned filter state that this PR does not contain. The description is the durable record of a stacked PR; please correct the bullet.

**[SUGGESTION]** — `pricing-and-sync-section.tsx:74` / `price-changes-queue-table.tsx:51`: the `connections ↔ price-changes` barrel cycle is one line away
> `price-changes-queue-table.tsx` value-imports `useConnectionsQuery` from `'../../connections'`, and this PR makes `pricing-and-sync-section.tsx` value-import from `'../../price-changes'`. It does not close today only because `PricingAndSyncSection` is absent from `features/connections/index.ts` and `connection-pricing-sync-page.tsx:25` deep-imports it (legal — `docs/frontend-architecture.md` still exempts `pages/`). The next person who "tidies up" by adding it to the barrel closes a runtime cycle (#337/#359). Worth one comment on the component saying it must stay out of the barrel, and why.

**[SUGGESTION]** — `pricing-and-sync-section.tsx:239-266` vs `:286`: the draft re-seeds from two sources, the dirty check reads one
> `draft` is seeded from `query.data` at mount and re-seeded from `updateMutation.data` on success, while `isDirty` compares only against `query.data`. Two consequences: the unsaved bar can briefly reappear after a *successful* save if the invalidation refetch and the re-seed effect land in different commits (the "Saved" toast and "You've changed something and haven't saved it yet" on screen together); and an edit made while a save is in flight is clobbered when the mutation resolves. Comparing against `updateMutation.data ?? query.data` in `isDirty`, or dropping the second effect and relying on the refetch alone, removes the asymmetry.

**[SUGGESTION]** — `pricing-and-sync-section.tsx:164`: `PERCENT_PATTERN` makes a decimal transiently invalid
> `/^\d+(\.\d+)?$/` rejects `"22."`, so typing `22.5` flashes *"Enter a percentage, 0 or more."* and disables Save between the `.` and the `5`. The string-typed draft correctly preserves the character (the regression the fix targeted is gone), but the field now makes a false claim mid-keystroke. `^\d+(\.\d*)?$` for the live state, with the strict form applied at save, avoids it.

**[SUGGESTION]** — inline styles and one raw px in a component that ships its own stylesheet
> `style={{ maxWidth: 'none' }}` (`:443`), `style={{ marginBottom: 'var(--space-3)' }}` (`:503`, `:600`), alongside 176 new lines of namespaced CSS — these belong in the block. And `gap: 6px` in the new CSS is a raw px where every neighbour uses `var(--space-*)` (`docs/frontend-ui-style-guide.md`, design-token rule).

**[SUGGESTION]** — `stock-and-pricing-section.tsx:78-79`: `pricingRuleManagedElsewhereHref` is optional but its own docblock says *"Required when the flag is set"*
> When the flag is true and the href is absent the operator gets the dead plain-text string `'Manage pricing & sync'` with nowhere to go, which is worse than either branch. A discriminated pair (`{managedElsewhere: false} | {managedElsewhere: true, href: string}`) makes that unreachable at compile time.

### Note on CI (not a finding against this PR)
`Test` and `Integration Tests` are red on this PR. All three web failures are in `features/price-changes/{components/edit-price-change-dialog,lib/price-change-copy}.test.ts*` — files this PR does not touch — and PR #3165 on the base branch `3148-price-changes-dialogs` is red identically, so they are inherited, not introduced here. Flagging only so the stack does not reach `main` in that state. (The earlier "no CI has run on any of these branches" note is now stale — CI is running.)

### Verdict
🔄 **Approve with changes**

No blocking issues remain; the three earlier BLOCKING findings are properly closed. Priority order:

1. Surface the validation error when the rule form is collapsed (or keep it open while invalid) — `pricing-and-sync-section.tsx`.
2. Extract the single destination predicate and use it at all four sites.
3. Correct the stale `initialConnectionId` bullet in the PR description.

Posting as a comment rather than an approval only because IMPORTANT findings remain; this is not a request for changes and I have no objection to it landing once (1) and (2) are addressed.

---

norbert-kulus-blockydevs:
All three numbered items plus every suggestion — `bef6a7ae5`, on the parent catch-up `2c0ff69ea`.

**1. Save disabled with no visible reason.** The unsaved bar now states the blocker itself (`role="alert"`), replacing the neutral "You've changed something…" text while one exists. I deliberately did **not** take the force-open option: `hasInvalidRule` is `defaultRuleError || activeSourceErrors.some(...)`, and a per-source row's error is nowhere near the default rule form — forcing that form open would have fixed the case you described and left the other half still unexplained. A test pins both renderings, which is also what caught it: the message legitimately appears twice now, and the existing assertion had to be scoped to the form.

**2. The four-way predicate.** Extracted as `isPricingDestination` (`features/connections/lib/pricing-destination.ts`, on the barrel) and all four sites call it — including the transitive one, since `pricingRuleManagedElsewhere` is derived from `needsMasterCatalog`. Your framing of *why* it is now load-bearing is recorded in the file's docblock rather than only in the commit, because the deletion arm is what makes a future divergence silent rather than merely inconsistent. `isMarketplace` fell out unused and is gone.

**3. The stale description bullet.** Corrected, and I said explicitly what was wrong rather than quietly swapping it — the merge record now notes that the `initialConnectionId` approach was replaced by the URL-owned `queueConn` param, and that the leftover bullet is what cost an earlier round a finding on state this PR does not contain.

**Suggestions, all taken:**

- **Barrel cycle.** `PricingAndSyncSection`'s docblock now says it must stay out of `features/connections/index.ts`, names the two value imports that would close the cycle (#337/#359), and says the deep import from `pages/` is deliberate rather than untidy.
- **Draft re-seed vs dirty check.** One `baseline = updateMutation.data ?? query.data`, read by both `isDirty` and the custom-sources comparison. The asymmetry you describe — "Saved" toast over "you haven't saved it yet" — is exactly what it removes.
- **`PERCENT_PATTERN`.** `^\d+(\.\d*)?$` for the live state. The strict form is effectively applied at save already: the serializer does `Number(rule.percent)`, so a trailing `.` persists as the integer and `''` still fails the pattern and blocks Save.
- **Inline styles and the raw px.** Both styles are now `--wide` / `--spaced` modifiers in the block. On `gap: 6px` → `var(--space-2)`: the scale has no 6px, so this is a 2px change, recorded in a comment rather than passed off as identical. Related, and left alone because it belongs to #3165's block one PR down: `.opt-in-check` carries a raw `gap: 10px`. Say the word and I will fold it in here rather than bouncing back down the stack.
- **`pricingRuleManagedElsewhereHref`.** Collapsed the pair into one optional object (`pricingRuleManagedElsewhere?: { href: string }`) rather than a discriminated union — absent means "edited here", present means "not, and here is where", so the dead-plain-text combination is unrepresentable with no union gymnastics across the component's other props. Caller and test harness updated.

**On the CI note:** the three inherited web failures were #3165's, fixed at the source there (`15ba8be12`) and carried in through this branch's merge. Locally: `type-check` clean, `check:invariants` clean, 558 tests passing across `features/connections`, `features/price-changes` and `pages/connections`.

**One carry-over closed here rather than left dangling:** `useSetSourceSyncModeMutation` (#3165) now invalidates `connectionPricingSyncQueryKey` as well as the price-changes list. It could not on that branch — the cached pricing-sync read arrives with this one — and without it this settings page would keep rendering the mode the queue's "also set to Automatic" toast had just changed.

piotrswierzy (approved):
## Tech Lead re-review (`96569c8` → `f3fd976`)

Delta is one own commit, `bef6a7a` (*"one pricing-destination predicate, and say why Save is disabled"*); the rest of the range is base-branch catch-up. **All three IMPORTANT findings are closed, and so is every one of the five SUGGESTIONs** — verified by reading the code at head, not the description. Specifically: the unsaved bar now renders `saveBlockedReason` with `role="alert"` beside the disabled Save, so collapsing "Edit default rule" no longer hides the reason, and the fix deliberately covers the per-source rows too rather than just force-opening the form (their errors are nowhere near it); `isPricingDestination` is a single exported predicate in `features/connections/lib/pricing-destination.ts` and all four sites call it — `EditConnectionForm:497`, `connection-pricing-sync-page.tsx:57`, `connection-detail-page.tsx:254`, and transitively `StockAndPricingSection` — with the docblock recording *why* a divergence is destructive rather than merely inconsistent; and the stale `initialConnectionId` bullet is now corrected in place, naming the round it cost.

The suggestions were taken as well, each for the stated reason: the barrel-cycle hazard is a docblock on the component saying it must stay off `features/connections/index.ts` and why; `isDirty` now diffs against `updateMutation.data ?? query.data`, closing the post-save flash; `PERCENT_PATTERN` is `^\d+(\.\d*)?$` so `22.5` no longer flashes an error mid-keystroke; both inline styles became `--wide` / `--spaced` modifiers in the block it already ships; `gap: 6px` became `var(--space-2)`; and `pricingRuleManagedElsewhere` is now an optional `{href: string}` object, so the dead plain-text-with-nowhere-to-go branch is unrepresentable rather than merely discouraged. `useSetSourceSyncModeMutation` also picked up the `connectionPricingSyncQueryKey` invalidation its own comment had flagged as owed once #3166 landed the cached read — that was not a review finding and is a good catch.

Gate checks re-confirmed at head: `ROLE_PERMISSIONS.operator` still carries only `connections:read` (`role.types.ts:105-113`), so `useWriteAccess('connections:write')` remains the admin-only gate the docblocks claim; dependency direction and the `pages/` deep-import exemption are respected; no `@openlinker/core` import, no raw `fetch()`, no raw hex or px in the new CSS.

### Findings

**[SUGGESTION]** — `pricing-and-sync-section.tsx` — `saveBlockedReason` renders only inside the `hasUnsavedChanges` branch
> Not reachable today: an invalid rule implies the draft diverged from the baseline, so the bar is always present when the reason exists. Worth one line noting the coupling, since it is the third mechanism guarding the same "don't disable a control with the reason off screen" property and the other two now have comments explaining themselves.

### Verdict
✅ **Approve**

CI at the time of writing: `Type Check`, `Docker Build Smoke Test`, `PHP Unit Tests` pass; `Lint`, `Test`, `Build`, `Integration Tests` are queued on the current run. `mergeable: MERGEABLE`, `mergeStateStatus: BLOCKED` (required checks + review). Approving on the code; the stack should not reach `main` until this run is green — and note the `Integration Tests` failure I attributed earlier belongs to #3162/#3163's `listings-price-change-episode.int-spec.ts` further down the stack, not to this PR.

---

norbert-kulus-blockydevs:
Taken — `06e5d7532`. `saveBlockedReason` now carries the coupling in a comment rather than leaving it to be re-derived: it renders inside the unsaved bar, so it is only ever seen while the bar is, which is sound today because an invalid rule can only come from an edit and an edit diverges the draft from its baseline. The comment names the condition that would break it — a rule that stops validating server-side, with no pending change — so the next reader knows what to check rather than what to assume.

piotrswierzy (approved):
## Tech Lead re-review (`8017d55c` → `c5637ffd`)

Seven commits in the range, **five of them merges** carrying the stack's base branches (`3143` → `3145` → `3146` → `3147` → `3148`) forward. The PR's own new work is **two comment-only commits** and nothing else: `06e5d753` answers my `saveBlockedReason` SUGGESTION by naming the coupling in place, and `44aabb62` corrects a stale note in `price-change-detection.service.ts` (that one is a stack-wide fix and appears in all three PRs' ranges). The net file diff base…head is `+7/-0` in one `.tsx` and `+8/-4` in one core service — zero executable lines. No behaviour changed, so nothing my approval rested on could have moved; **the approval stands**.

Two verifications, since a base-branch merge is exactly how a definition gets deleted silently:

- `06e5d753` states the right thing rather than the convenient thing — it doesn't claim the coupling is fine, it names the invariant the second mechanism depends on (an invalid rule can only come from an edit) and says what to do if that ever stops holding. That is the SUGGESTION closed properly.
- `44aabb62`'s claim is **verified against the code, not taken on the commit message's word**: `PriceChangeApplyService.applyPriceChange` awaits `publishPrice(input)` and only then calls `episodes.resolve(episodeId, 'accepted' | 'accepted-custom', …)` (`price-change-apply.service.ts:167-190`). Ordering, status value and the `episodeId`-always-populated premise all hold, so the corrected comment is accurate — which matters, because a comment "corrected" into a new falsehood is worse than the stale one it replaced.

### Findings

**Positive** — `scripts/check-*-mirror.mjs` aside, this stack keeps paying down review notes as *documentation of the invariant* rather than as a code change that papers over it. Both notes here are the right shape: they name the condition under which the current design stops being sound.

### Verdict
✅ **Approve**

CI at `c5637ffd`: `MERGEABLE` / `UNSTABLE`. Type Check, Docker Build Smoke, PHP Unit and path detection **pass**; Build, Lint, Test and Integration Tests are **pending (queued, not run)** — no failing assertion anywhere. Note this is the bottom of the `3166 → 3167 → 3168` stack and targets `3148-price-changes-dialogs`, so it merges first.

---

piotrswierzy (approved):
## Tech Lead re-verification (head unchanged at `a8a1b3e`)

Head unchanged, `main` unmoved. CI on run 34845738505: **Type Check, Docker Build Smoke Test, PHP Unit Tests and Detect changed paths pass; Build, Lint, Test and Integration Tests are queued behind the runner backlog. No failures on any run.** Re-spot-checked both approval items: `isPricingDestination` lives in the single `features/connections/lib/pricing-destination.ts`, is re-exported from the feature barrel, and is imported by all four call sites (the edit form's `needsMasterCatalog`, and the three page/feature consumers) with no second copy of the predicate; `saveBlockedReason` is rendered in the unsaved bar and carries `role={saveBlockedReason ? 'alert' : undefined}`, so the live region is announced only when a reason exists rather than on every successful save.

Stack: base `3148-price-changes-dialogs` (#3165). **#3159 and #3162 below still carry an open BLOCKING** — each edits the already-merged migration `1880000000000-create-price-change-episodes.ts` in place, silently diverging an already-migrated database from a fresh install. This PR carries no migration; it is held only because **nothing in the chain can merge until those two are fixed.** `mergeable: MERGEABLE`, `mergeStateStatus: UNSTABLE`.

### Findings

### Verdict
✅ **Approve**

---

PR #3167, opened 2026-09-10
feat(web): source connection pricing rollup + pricing-rules picker

### LINKED ISSUE #3150: feat(web): source connection pricing rollup + pricing-rules picker modal
## Problem / Context

Part of #3139 (epic). **Stack position: eleventh — branches from #10's branch.**

Build the two remaining Pricing-&-sync entry points from `docs/plans/mockups/price-changes-review-queue.html`: the **source connection rollup page** (`#page-source`) and the **"Pricing rules" picker modal** (`#dialog-picker`). Together with #10 (the destination connection settings page) and #8's per-row `connection-tag`/`source-tag` clicks, this gives the feature its required TWO independent entry paths (queue → picker, and source connection → rollup → destination) into pricing settings, per the explicit design requirement that this must not be reachable only one way.

**Source connection rollup page** (mockup `#page-source`, `data-state="page-source"`) — a NEW, READ-ONLY page (not a section on an existing page, per ADR decision 2's Model A: the source connection never gets an editable rule of its own):
- `#crumb-back-source` / `#crumb-current-source` — breadcrumb back to wherever the operator came from.
- `#source-page-icon` / `#source-page-name` / `#source-page-platform` — connection identity header.
- `#source-page-list` — one row per destination connection this source feeds: destination name, `#source-page-manage-{destKey}` button ("Manage") — clicking it navigates to that destination's Pricing & sync page (#10) with the relevant per-source row PRE-EXPANDED (reuse the same `expandSourceRow`-equivalent mechanism #9's Edit-dialog permalink uses — do not build a second implementation of "navigate to connection settings and expand row X").
- Row summary text: "{Manual review|Automatic} · {rule summary sentence}" + either "— a rule just for this source" (custom override) or "— using {destination}'s default rule" (inherited) — exact phrasing from the mockup.
- **This page has NO write affordance anywhere** — verify no form field, toggle, or save button exists on it. It exists purely so an operator who navigated to a SOURCE connection (e.g. their PrestaShop store) isn't stranded wondering how its prices get adjusted downstream.
- Entry point: from #8's table, clicking the source-tag (`#source-tag-{rowId}`, `data-open-source`) on any row navigates here.
- Backend: `GET /connections/:id/pricing-sync/as-source` (#7).

**Pricing-rules picker modal** (mockup `#dialog-picker`, `data-state="dialog-connection-picker"`):
- Entry point: `#btn-pricing-rules` button in the Price-changes tab toolbar (built as a stub in #8, wired here).
- `#dialog-picker-list` — one row per DESTINATION connection (not source — this picker is specifically "which connection's rules do you want to see", and rules live on destinations per ADR decision 2): icon, name, "Default: {mode} · {rule label}" + "· N source(s) with a different rule" if any overrides exist.
- Row is clickable (`#picker-conn-{key}` container, `#picker-open-{key}` explicit button) — either interaction closes the picker and navigates to that connection's Pricing & sync page (#10).
- `#dialog-picker-cancel` — closes with no navigation.

## Classification

**Type**: Frontend
**Layer**: Interface (page + feature components)
**File(s)**: new page `apps/web/src/pages/connections/source-pricing-rollup-page.tsx` (or under `features/price-changes/` if that fits the feature-slice convention better — decide at implementation time based on where the routing naturally belongs, likely nested under the connections route tree per `docs/frontend-architecture.md` § Routing Conventions), `apps/web/src/features/price-changes/components/pricing-rules-picker-dialog.tsx`

## Dependencies

- #10 (destination connection Pricing & sync settings page — both this issue's surfaces navigate INTO it)
- #7 (as-source rollup API)

## Assumptions

- Route path for the source rollup page follows the existing connection-detail route convention (likely `/connections/:id/pricing` or similar — confirm no collision with #10's destination settings route, since a connection could theoretically be both a source for some things and a destination for others in a more complex install; for v1 the mockup's three demo connections are cleanly one-or-the-other, but the real route/page selection logic should resolve based on the connection's actual capabilities, not assume mutual exclusivity).

## Acceptance Criteria

- [ ] Source rollup page renders with zero write affordances (a reviewer/test should be able to assert no `<button>` on the page triggers a mutation, only navigation)
- [ ] "Manage" link from the rollup page lands on the correct destination connection's settings page with the correct source row pre-expanded
- [ ] Picker modal opens from `#btn-pricing-rules`, lists every destination connection with an accurate override count, and navigates correctly on row click
- [ ] Both entry paths (queue picker → destination settings, AND table row's connection-tag/source-tag → settings/rollup) independently reach the same destination settings page — verify there is no dead end
- [ ] Tests: at least one test per surface covering the navigation handoff (picker → settings page; rollup → settings page with pre-expanded row)

norbert-kulus-blockydevs:
## Summary

Part of #3139 (epic). Stack position: last — branches from #3149's branch (PR #3166).

Two operator entry points into Pricing & sync, both on surfaces that had none:

- **`SourceConnectionPricingRollup`** — the read-only "How your prices get adjusted" list on a SOURCE connection's page. One row per destination that adjusts this source's prices, each with its rule summary and a `Manage` link. Read-only per ADR-072 decision 2: a source connection never carries an editable rule, because rules live on destinations.
- **`PricingRulesPickerDialog`**, behind a "Pricing rules" button in the Price changes tab toolbar — every destination connection with its default rule and override count, as a way into each one's settings without hunting through `/connections`.

Both land on `connections/:connectionId/pricing-sync`, the route #3166 built. This PR does **not** build that route or `PricingAndSyncSection` — those are #3166's, and an earlier version of this description wrongly claimed them (#3167 approval, SUGGESTION). What it adds to that page is the rollup for the source half, and the two deep-link props below.

- `PricingAndSyncSection` gained TWO deep-link props, not one (**corrected in the round-3 review**): `initialExpandSourceId` scrolls the named source's row into view and highlights it, and `initialCreateOverrideForSourceId` stages an override for it. The page maps `?source=` to the first and `?source=&override=1` to the second. They were a single prop behind a single `?source=` param, which meant the rollup's neutral "Manage" pointer staged a per-source rule the operator never ticked — and since `persistSave` writes `sourceOverrides` unconditionally, any later unrelated edit on that page persisted it.

## What the round-3 review changed

`aacad8445`, on the parent catch-up `7eb9ded91`:

- **BLOCKING — the rollup had no CSS.** It rendered `.settings-section`, `.source-list`, `.source-row` and five siblings; none had a rule anywhere in `index.css`, so the surface shipped unstyled. They were also generic globals of the kind that silently restyled the Subiekt invoice dialog one PR down, and two collide by name with the editable section's own (correctly prefixed) `.pricing-sync__source-*`. The block is now `.pricing-rollup__*` — its own namespace rather than a reuse of `.pricing-sync__*`, since that one styles a DESTINATION's editable settings in another feature and the two must be free to diverge. The row height is registered in `docs/frontend-ui-style-guide.md` § Density & Row Heights.
- **The deep-link intent split** described in the bullet above.
- `<h3>` for the rollup's heading — `PageLayout` owns the `<h2>` level.
- The detail page's button gate widens to `hasPricingSyncPage`, so a `ProductMaster`-only connection can reach its own rollup. That predicate sits beside `isPricingDestination` (#3166) in `features/connections/lib` and is deliberately wider than it, for the stated reason.
- The picker's Open link leaves modified clicks alone; the stale-link notice stops borrowing the field-error red while announcing itself as `role="status"`; the picker's per-destination summaries get a `staleTime`, safe because both mutations that change them invalidate that exact key.

## Test plan

- [x] `pnpm check:invariants` — clean.
- [x] 574 tests passing across `features/connections`, `features/price-changes` and `pages/connections`, including three deep-link cases rewritten for the intent split and two new ones pinning both halves.

Closes #3150


apps/web/src/features/connections/components/pricing-and-sync-section.tsx
```diff
@@ -45,6 +45,16 @@ import { useToast } from '../../../shared/ui/toast-provider';
...
+  /**
+   * Pre-expands the named source's override editor on load — the "reuse the
+   * same expandSourceRow-equivalent mechanism" #3150 asks to share between
+   * the Edit-dialog's "Set a rule just for this source" permalink (#3148,
+   * `?source=`) and the source rollup page's "Manage" links. Enabling the
+   * override is what "expanded" means here: the row's body only renders
+   * for a custom source, so landing on an inherited one turns it on —
+   * exactly what a manual click of the checkbox already does.
+   */
+  initialExpandSourceId?: string;
```
norbert-kulus-blockydevs, line 57: **BLOCKING — toggling a per-source override never enables Save, so `initialExpandSourceId` is decorative.**

`isDirty` (line 137) compares only `draft` against `query.data`. It does **not** consider `customSources`, which is the state that actually decides whether a source gets an override. Both toggle directions leave `draft` byte-identical:

- **On:** `toggleSourceCustom(id, true)` sets `source.effective = cloneSetting(next.default)`. For an *inherited* source the API already resolved `effective` to the default, so this assigns the value it already had, and `cloneSetting`/`cloneView` are order-preserving spreads.
- **Off:** the code deliberately does not touch `draft` ("Unchecked: leave `effective` as-is for display").

`isDirty === false` ⇒ the `unsaved-bar` does not render ⇒ **there is no Save button at all.** An operator cannot add or remove a per-source override without also editing some unrelated field.

This is #3166's line, but it is blocking *here* because `initialExpandSourceId` exists solely to land an operator on a pre-expanded override — and that arrival now produces: checkbox visibly checked, rule editor open, `isDirty` false, no Save, no unsaved indicator. The UI asserts "this source has its own rule" while the server says it does not, and offers no way to make it true.

**Corollary: `pricing-and-sync-section.test.tsx:82` must be red in CI.** After `userEvent.click(getByTestId('source-custom-toggle'))`, `getByRole('button', { name: 'Save changes' })` has nothing to match. Neither #3166 nor this PR could run tests locally, so the failure is unobserved rather than absent.

```ts
const serverCustom = new Set(
  (query.data?.sources ?? []).filter((s) => s.isCustomOverride).map((s) => s.sourceConnectionId),
);
const overridesDirty =
  customSources.size !== serverCustom.size ||
  [...customSources].some((id) => !serverCustom.has(id));
const isDirty = query.data
  ? overridesDirty || JSON.stringify(draft) !== JSON.stringify(query.data)
  : false;
```
`handleDiscard` already resets `customSources`, so nothing else changes.

**Also on this prop:** the pre-expanded row is never scrolled to or focused. The page is long (default rule, N source rows, pending changes, activity), so arriving from `Manage` on the 5th source shows the top of the page with no indication anything was expanded — the "no dead end" AC this PR opened in order to verify. `MappingPanel.tsx:236` and `bulk-review-step.tsx:267` both do `scrollIntoView({ block: 'center', behavior: 'smooth' })` after a deep-link reveal. And if `?source=X` names a source absent from `query.data.sources`, `expandTarget` resolves to `null` and nothing renders or reports.

apps/web/src/pages/connections/pricing-sync-page.tsx
```diff
@@ -0,0 +1,81 @@
...
+import { EmptyState, ErrorState, LoadingState } from '../../shared/ui/feedback-state';
+import { PageLayout } from '../../shared/ui/page-layout';
+
+export function PricingSyncPage(): ReactElement {
+  const { connectionId = '' } = useParams();
+  const [searchParams] = useSearchParams();
+  const connectionQuery = useConnectionQuery(connectionId);
+
+  const connection = connectionQuery.data;
+  const isDestination =
```
norbert-kulus-blockydevs, line 30: **IMPORTANT — a connection that is both source and destination loses the rollup entirely, and its source-tag lands on the wrong surface. Also: no write gate.**

Destination wins, and the docblock says so deliberately. But `woocommerce-plugin.ts:53` advertises `ProductMaster`, `InventoryMaster`, `OfferManager` **and** `ProductPublisher` on one adapter, so a WooCommerce connection enabling `ProductMaster` + `ProductPublisher` is an ordinary, shipped configuration. For it:

- `SourceConnectionPricingRollup` is **unreachable** — there is no other route, tab or link to it.
- #3164's **source**-tag navigates here and renders that connection's own *outbound* destination settings, answering a different question than the operator asked.

Issue #3150's Assumptions section is explicit: *"the real route/page selection logic should resolve based on the connection's actual capabilities, **not assume mutual exclusivity**"*. The implementation reads capabilities and then re-imposes exclusivity at render. **Fix:** render both when both hold, with the rollup's own empty state covering a destination that feeds nobody.

*(Secondary: `enabledCapabilities` is stamped at create and never retro-filled (#2085), so an older connection can under-report. Noted rather than raised separately — `EditConnectionForm`'s `needsMasterCatalog` resolves the same question the same way, so page and form agree, which is the more important property.)*

**Separately — no `useWriteAccess` anywhere** in this page, `pricing-and-sync-section.tsx` or `source-connection-pricing-rollup.tsx`, while every sibling in that directory gates on `connections:write` (`ConnectionActionsPanel.tsx:33`, `EnableConnectionButton.tsx:44`, `EditConnectionForm.tsx:383`, `create-connection-form.tsx:58`, `router-readiness-panel.tsx:69`). `connections:read` is held by `viewer` and `operator` and the GET is viewer-allowed, so both reach the page, see mode switches, percent inputs and Save — and Save answers 403, because the PATCH is `@Roles('admin')`. This PR makes it materially worse than #3166 did: the surface is now reachable from four links rather than one buried section.

I checked the mismatch I suspected and it does **not** exist: `role.types.spec.ts:25` asserts `operator` lacks `connections:write`, so `useWriteAccess('connections:write')` is effectively admin-only and **agrees** with the endpoint. This is a plain missing gate, not a permission divergence — the ordinary remedy suffices.

**Minor, same file:** `isDestination` is `undefined` before `connection` resolves and the description ternary sits outside the loading guard, so a destination connection renders *"How each place you sell adjusts your price."* for the duration of the fetch, then flips.

apps/web/src/features/price-changes/hooks/use-destination-pricing-sync-summaries.ts
```diff
@@ -0,0 +1,23 @@
...
+ * "Pricing rules" picker (#3150) — each row needs its own default rule +
+ * override count, and there is no list-shaped endpoint for it. Index-aligned
+ * with the input `connectionIds`, the `useProductsBatchQuery` shape.
+ */
+export function useDestinationPricingSyncSummaries(
+  connectionIds: readonly string[],
+): UseQueryResult<ConnectionPricingSyncView>[] {
+  const apiClient = useApiClient();
+
+  return useQueries({
```
norbert-kulus-blockydevs, line 17: **IMPORTANT — N requests fire on every Price-changes tab open, for a dialog that may never be opened.**

`useQueries` issues one `GET /connections/:id/pricing-sync` per destination connection, and `PricingRulesPickerDialog` is mounted **unconditionally** in `price-changes-queue-table.tsx:261` (only its Radix *content* is conditional), so the component body — and this hook — runs whenever the tab renders. On an install with 20 destinations that is 20 requests on tab open, none of which the operator asked for.

The hook's own docblock cites *"the `useProductsBatchQuery` shape"* — but that hook takes `options?: { enabled?: boolean }` and threads it into every query, and **both** its consumers use it (`product-sales-table.tsx:453`: `{ enabled: productIds.length > 0 }`). The cited precedent's one safety parameter is the thing that was dropped. This is also the repo's standing objection to per-row fan-out reads (#2083; #2353's note that `ConnectionResponseDto` omits a field precisely because `toResponse` runs per row).

**Fix (minimum):** add `enabled` and pass `open`. **Better:** a list-shaped read — the picker needs only `{connectionId, default, overrideCount}`, a small addition to #3163's controller that removes the fan-out entirely.

*(Worth knowing: #3168's `AutoAppliedNote` currently free-rides on this over-fetch as its justification for not gating on config. Gating this hook on `open` — which is correct — removes that, so #3168 needs its own answer.)*

apps/web/src/features/price-changes/components/pricing-rules-picker-dialog.tsx
```diff
@@ -0,0 +1,101 @@
...
+                    {summary ? (
+                      <>
+                        Default: {summary.default.mode === 'automatic' ? 'Automatic' : 'Manual review'} ·{' '}
+                        {ruleLabelFor(summary.default.rule)}
+                        {overrideCount > 0
+                          ? ` · ${overrideCount} source${overrideCount === 1 ? '' : 's'} with a different rule`
+                          : ''}
+                      </>
+                    ) : (
+                      'Loading…'
```
norbert-kulus-blockydevs, line 74: **IMPORTANT — a failed summary renders as "Loading…" forever, and zero destinations renders nothing at all.**

`const summary = summaries[index]?.data` — a query that **failed** has `data === undefined`, so the row reads `Loading…` permanently: a failed read presented as an in-flight one. With N independent queries, one 500 on one connection silently pins one row. `docs/frontend-architecture.md:281` states the rule: *"Absence and 'none matched' are different claims, and only one of them is safe to make from a failed read."* The `overrideCount` falls to `?? 0` on the same path, so the row can also assert **"no overrides"** about a connection whose config could not be read — a false statement about the operator's own configuration.

Separately, `destinationConnections.map(...)` over an empty array renders an empty `.mini-list` — the dialog opens with a title, a description and nothing between them.

Issue #3150's AC asks the picker to *"list every destination connection with an accurate override count"*; both branches fail it, one by inaccuracy and one by silence.

**Fix:** branch on `summaries[index]?.status` — `pending` ⇒ "Loading…", `error` ⇒ a stated "Couldn't load this connection's rule" (keep the row clickable, navigation still works), `success` ⇒ the summary. Add an `EmptyState` for zero destinations naming the remedy.

apps/web/src/features/price-changes/components/pricing-rules-picker-dialog.tsx
```diff
@@ -0,0 +1,101 @@
...
+          {destinationConnections.map((connection, index) => {
+            const summary = summaries[index]?.data;
+            const overrideCount = summary?.sources.filter((s) => s.isCustomOverride).length ?? 0;
+            return (
+              <div
+                key={connection.id}
+                id={`picker-conn-${connection.id}`}
+                className="mini-row"
+                style={{ cursor: 'pointer' }}
+                role="button"
```
norbert-kulus-blockydevs, line 55: **IMPORTANT — an ARIA nested-interactive violation this repo has already removed twice, plus a Space key that scrolls the page.**

`<div role="button" tabIndex={0} onClick={…} onKeyDown={…}>` wrapping a `<Button>Open</Button>`.

Two in-repo precedents removed exactly this shape:
- `features/listings/components/bulk/bulk-review-step.tsx:633` — *"The whole row is no longer a `role="button"`: it wrapped a checkbox, a zoom button, chip buttons and an Edit button, **which is an ARIA nested-interactive violation**."* (#1741 review #9)
- `shared/ui/read-only-lock.tsx:54` — *"**No `role="button"`**: the wrapper contains an already-disabled `<button>`, and nesting a button role around it would misreport the semantics"* — and that handler does `e.preventDefault()` with the comment *"Space would otherwise scroll the page."*

This re-introduces both: the nested role, and Space with no `preventDefault`, so pressing Space on a focused row scrolls the page behind the modal in addition to navigating.

There is a third problem the precedents do not cover: this is a **navigation** affordance built as a button, so middle-click, Ctrl/Cmd-click, "open in new tab" and "copy link address" are all dead — while `SourceConnectionPricingRollup` two files over uses a real `<Link>` for identical navigation.

**Fix:** drop `role="button"`/`tabIndex`/`onKeyDown` from the container and make `Open` a `<Link className="button button--secondary button--xs">` calling `onOpenChange(false)` in `onClick`.

apps/web/src/features/connections/components/source-connection-pricing-rollup.tsx
```diff
@@ -0,0 +1,92 @@
...
+        message={query.error.message}
+      />
+    );
+  }
+
+  const entries = query.data ?? [];
+  if (entries.length === 0) {
+    return (
+      <EmptyState
+        title="No destinations yet"
```
norbert-kulus-blockydevs, line 56: **IMPORTANT — the empty state makes a positive claim the data cannot support; and this file is in the wrong feature.**

`ConnectionPricingSyncService.getAsSource` includes a destination only when `isCustomOverride || destinationIdsFromEpisodes.has(destination.id)`. A destination consuming this source with its **default** rule and **no open episode** is filtered out. So the common steady state — an operator who has cleared their queue and never set a per-source override — yields `[]`, and this copy tells them *"Nothing sells your catalogue through a price-adjusting connection yet."* That is false, and it is `docs/frontend-architecture.md:281` again: an empty read is never a positive claim. (It also makes the row copy self-inconsistent: `summaryFor` renders *"— using {destination}'s default rule"*, but per that filter an inherited row can only appear while an episode is open.)

**Fix (now):** reword to something the response supports — "No destinations are adjusting your prices right now", plus a line saying a destination appears once it has its own rule for this source or a price change waiting. **Fix (properly):** widen `knownAsSource` to include every destination whose `masterCatalogConnectionId` names this source; file against #3163.

**Separately — placement.** This component's only imports are `react`, `react-router-dom`, `shared/ui/feedback-state` and `../../price-changes`. It has **zero** dependencies on `features/connections` (`connectionId` arrives as a plain string prop), yet living here creates:

```
features/connections/components/source-connection-pricing-rollup.tsx
  --value--> features/price-changes/index.ts
  --value--> components/price-changes-queue-table.tsx
  --value--> features/connections/index.ts   (useConnectionsQuery)
```

No cycle **today**, only because `connections/index.ts` does not export the rollup — so the page deep-imports it. But that barrel's docblock says *"Anything other features or plugins consume must be re-exported here"*, so the natural next edit closes the loop (#337/#359; #2640 is the recorded case where a panel was placed in the *other* feature specifically to avoid it). There is also an immediate cost: importing the price-changes **barrel** for two symbols pulls the queue table, four dialogs and the bulk-progress widget into a lazy route that needs none of them.

Issue #3150 named the alternative: *"or under `features/price-changes/` if that fits the feature-slice convention better"*. Move it there; `pricing-and-sync-section.tsx` is a genuinely connections-shaped surface and can stay.

apps/web/src/app/routes/route-lazy.test.ts
```diff
@@ -83,7 +86,7 @@ const lazyRoutes = collectLazyRoutes([
  *   - prompt-templates-legacy-redirects (inline `<Navigate>` element)
  *   - `/analytics` legacy alias (inline `<Navigate>` to `/`, #2740)
  */
-const EXPECTED_LAZY_ROUTE_COUNT = 63;
+const EXPECTED_LAZY_ROUTE_COUNT = 64;
```
norbert-kulus-blockydevs, line 89: **SUGGESTION — the constant is right; the comment's arithmetic is not.**

The narrative now reads *"63 total = 47 authenticated + 4 guest + 12 plugin"* (= 63) against `EXPECTED_LAZY_ROUTE_COUNT = 64`. The drift is pre-existing — the base branch had "62 = 46+4+12" against a constant of 63 — and this PR incremented both halves, preserving it. The constant is correct (base 63 + one new lazy route).

Since the comment is the entire documentation of what the assertion means, fix it to `64 total = 48 authenticated + 4 guest + 12 plugin` while you are here.

**Also missing:** there is no `pricing-sync-page.test.tsx`. The PR's headline correctness claim — capability-based surface selection — is untested, as are the not-found branch and the `?source=` threading.

norbert-kulus-blockydevs (commented):

The route-building and the demotion are both correct calls, and the diagnosis in the PR body is honest work. Specifically verified and worth crediting: the route registration is byte-identical in shape to `edit-connection.route.tsx` / `connection-mappings.route.tsx` (same param name, same `RouteCrumbHandle`, same lazy import, registered inside the authenticated layout), and `connections/:connectionId/pricing-sync` is a unique 3-segment pattern that React Router ranks by specificity — so the #2953-class silent shadowing is not reachable here. The `?source=` convergence is **real**, not nominal: one param, one reader, one effect, three entry points. The dialog uses the Radix-backed primitive rather than the mockup's hand-written `role="dialog"`. The rollup's read-only invariant is taken seriously and its test is meaningful (the `Manage` affordances are `<Link>`s, so `queryByRole('button')` is a real assertion, not a tautology). And adding `asSource` to `createMockApiClient` — easy to forget — was not forgotten.

The problem is that **the surface all four links now lead to cannot save the thing those links exist to set**, and the page has no permission gate at all. Details inline.

**Two of my going-in hypotheses were falsified, both in this PR's favour:** the `EditConnectionForm` demotion leaves no stale test (`pricing-and-sync-section.test.tsx` renders the component directly), and `ConnectionAsSourceEntry` matches its DTO field-for-field — the drift is one level down in `PricingRule`, whose FE type declares `percent`/`rounding` **required** while the core type and #3163's DTO make them optional.

**Recommendation: squash with #3166**, or at minimum land the `isDirty` fix there first. #3166's only user-facing contribution is removed by this PR, so merging both as-is puts a surface into `main` and takes it out again — and any bisect landing between them exercises a placement the team has already rejected.
---
*Part of a whole-stack review of #3158–#3168 (epic #3139). Cross-cutting items tracked once: `pnpm lint` is red on the stack; `docs/architecture-overview.md § Listings` is untouched by all ten PRs though ADR-072 names it the primary doc section; the mockup has not changed since #3155 despite ~20 behavioural divergences; `features/price-changes` is registered in **neither** `.eslintrc.js` `no-restricted-imports` group, so the new barrel is unenforced; and `.source-row` / `.mini-row` are non-`DataTable` rows absent from `docs/frontend-ui-style-guide.md § Density & Row Heights`.*

piotrswierzy (changes requested):

**The best-reasoned PR of the FE batch.** It detects and repairs two real defects its predecessors shipped: it moves the editable surface to a dedicated route and demotes the `EditConnectionForm` copy to a link-out (citing the `SalesDocumentStatusSection` precedent by name), which removes the two-Save-scopes hazard **and** resolves the dangling links #3148 had already shipped against a route that didn't exist. The "Corrections made along the way" section is exactly the right way to report that.

The rollup is **genuinely read-only** per ADR-072 decision 2 — no form field, only `Manage` links — and its test asserts the **absence** of write affordances rather than just the presence of content. That is the right shape of test. Route conventions fully honoured (`handle.crumb` declared, `route-lazy.test.ts` count bumped 63→64), primitives reused, capability-driven branching on `enabledCapabilities` never `platformType`.

Most blocking items are **inherited, not created**.

### BLOCKING (inherited) — #3166's `isDirty` defect

`pricing-and-sync-section.tsx` is modified here (+36/−7) but only to add `initialExpandSourceId`. The override-cannot-be-saved bug and the `RuleFields` percent/≥100% bugs are unchanged, and now reachable from **three** entry points (the rollup's Manage, the picker, and #3148's permalink).

### BLOCKING (this PR's own) — a deep link pre-arms a write the operator never requested

Arriving at `?source=src-1` for a source that is currently *inheriting* sets `customSources` **and** assigns `source.effective = cloneSetting(next.default)` — i.e. navigation creates a pending per-source override. The docblock justifies it as *"exactly what a manual click of the checkbox already does"*, but a click is an operator act and a `<Link>` traversal is not — and the rollup's Manage button is the innocuous-looking trigger.

Today #3166's `isDirty` bug masks it. **Fixing `isDirty` makes this land as a spurious unsaved-changes bar on page load**, and a subsequent Save persists an override nobody asked for. Fix both in the same commit.

### IMPORTANT — a failed summary renders `'Loading…'` forever

`pricing-rules-picker-dialog.tsx` (~68): `{summary ? (…) : 'Loading…'}`, and `summaries[index]?.data` is `undefined` both while pending **and on error**. So a failed `GET .../pricing-sync` shows a row claiming to be loading that never resolves — a false statement about state, and exactly the distinction § Paginated Totals protects: *"`idle` and `unavailable` are different, and a surface must render them differently."* Same mechanism hides the override count: on failure `overrideCount` is `0` and the clause is omitted, so *"no sources with a different rule"* and *"we couldn't find out"* are indistinguishable.

### IMPORTANT — nested interactive elements in the picker rows

A `<div role="button" tabIndex={0}>` containing a real `<Button>Open</Button>`. Invalid ARIA, **two tab stops per row** doing the same thing, and `onKeyDown` handles `' '` without `preventDefault()`, so Space activates *and* scrolls the dialog. The repo has a convention — `shared/ui/card-button-reset-styles.test.ts` exists precisely to support making a card a real `<button>`. Either collapse the row to one `<button>` or make it inert and keep only Open.

### IMPORTANT — § UX Mockups unmet at PR level, and two deviations are undocumented

Both component headers cite the mockup correctly (`#page-source`/`#source-page-list`, `#dialog-picker`) — genuinely good — but the description doesn't, and this PR makes two structural deviations the mockup has not been updated for: a **new dedicated route** the mockup doesn't have, and a **link-out stub** in `EditConnectionForm` that appears nowhere in it. These are good changes — say so in the mockup or the description.

### SUGGESTIONS

`isDestination` is `undefined` while loading, so the **source** copy renders during load and on error, then flips. `ErrorState`'s `action` is a raw `<button className="button button--secondary">` where `Button` is available. The `entries.length === 0` branch returns `EmptyState` *instead of* the section, so the heading and its explanatory copy vanish in the empty case. And `useDestinationPricingSyncSummaries` fires N parallel requests with no `staleTime` — acceptable for a dialog, but #3168 mounts the same hook **unconditionally**, which turns it into a fan-out on every queue render.

---

norbert-kulus-blockydevs:
Rebased onto `3149-connection-pricing-sync-settings-page` (now including #3166's review fixes) and resolved a real structural duplication surfaced by that rebase, plus every review finding.

## File consolidation (the load-bearing part of this update)

While this branch was in flight, #3166 landed a fix for the sibling PR that independently built the **same** "give a destination connection its own Pricing & sync route/page" split this branch was also building — just under different file names, both registering a route at the identical path `connections/:connectionId/pricing-sync`. Two different pages would otherwise both be wired up, one of them silently shadowing the other in React Router (last-registered wins, no error).

**Resolution:**
- **Kept as canonical** (from #3166, already on `main`'s ancestor of this branch): `apps/web/src/app/routes/connection-pricing-sync.route.tsx` and `apps/web/src/pages/connections/connection-pricing-sync-page.tsx`.
- **Deleted** (this PR's own now-redundant duplicates): `apps/web/src/app/routes/pricing-sync.route.tsx` and `apps/web/src/pages/connections/pricing-sync-page.tsx`.
- The canonical page now carries this PR's real value: it renders `SourceConnectionPricingRollup` for a `ProductMaster` connection, **alongside** (not instead of) the editable `PricingAndSyncSection` when a connection is both a destination and a source — a real WooCommerce configuration (`ProductMaster` + `ProductPublisher`) that an earlier either/or branch would have silently dropped one half of (review finding 2). `?source=` from the URL threads into `PricingAndSyncSection`'s `initialExpandSourceId`.
- `SourceConnectionPricingRollup` (+ its test) also **moved** from `features/connections/components` to `features/price-changes/components` (finding 5) — it has no actual dependency on the `connections` feature slice, and staying there created a latent barrel-cycle risk plus pulled the whole price-changes barrel into what should be a lightweight lazy route.
- Also found and fixed, while resolving the same rebase conflicts: a genuine duplicate-`const destinationConnections` declaration this PR's own commit had introduced in `price-changes-queue-table.tsx` (same predicate, would've been a `tsc` redeclaration error), and a second, now-redundant "Pricing & sync" pointer section in `EditConnectionForm.tsx` (#3166 already added an equivalent pointer inside `StockAndPricingSection`).

## Review findings addressed

1. **`initialExpandSourceId` deep-link (finding 1)** — the `isDirty`/`customSources` half was already fixed by #3166 on the branch this rebased onto. Added here: the pre-expanded row now scrolls into view (`bulk-review-step.tsx`'s `pendingScroll` precedent), and a `?source=` id absent from the connection's current sources is reported inline instead of silently resolving to nothing.
2. **Mutual-exclusivity assumption (finding 2)** — fixed via the consolidation above: both sections render when both capabilities are present.
3. **N unconditional requests on every tab render (finding 3)** — `useDestinationPricingSyncSummaries` gained the `useProductsBatchQuery` `enabled` pattern, gated on the picker dialog's own `open` state.
4. **Picker dialog defects (finding 4)** — branches on the query's own `status` (a permanently-failed read no longer renders "Loading…" forever, nor asserts "no overrides"), added a real `EmptyState` for zero destinations, dropped the `role="button"` wrapper around a real `<Button>` (ARIA nested-interactive violation — `bulk-review-step.tsx`/`read-only-lock.tsx` precedent), made "Open" a real navigating `<Link>`.
5. **False-negative empty-state copy (finding 5)** — reworded rather than attempting the backend widening (flagged as a follow-up referencing #3163's PR, since the query only lists a destination once it has its own override or an open episode).
6. **SUGGESTION** — fixed `route-lazy.test.ts`'s pre-existing (not introduced by this PR) breakdown arithmetic: `/bench` was narrated as part of the "authenticated" bucket while actually living in the separate `standaloneRoutes` array, so `47 + 4 + 12` never summed to the real, already-correct constant of `64`.
7. Added `connection-pricing-sync-page.test.tsx` coverage for the source-only / both-capabilities branches and `?source=` threading — this page's headline correctness claim (capability-based branching, never mutual exclusivity) had no test at all.

## Verification (no local pre-commit hook — resource-constrained machine, per longstanding project note)

- `pnpm --filter @openlinker/web lint` — 0 errors, pre-existing warning count (123) unchanged.
- `pnpm --filter @openlinker/web type-check` — clean.
- `pnpm check:invariants` — all green.
- Targeted `vitest` runs across every touched spec file — 151 tests passing.

norbert-kulus-blockydevs:
Re-verified this review against current code before acting — the prior fix round (18:08 UTC) had already landed, and #3149 moved again after that (a test-stranding fix, 27ae3204a), so this branch needed a fresh `rebase --onto` first (clean, no conflicts).

**Confirmed already resolved** (no action needed):
- **isDirty/customSources** — `isCustomSourcesDirty` genuinely diffs `customSources` against server truth now (landed on `main` via #3166).
- **Finding 4 (picker "Loading…" forever)** — `PricingRulesPickerDialog` branches each row on the query's own `status` (`pending`/`error`/`success`), not on `.data` alone.
- **Finding 4 (nested interactive elements)** — the row's `role="button"` wrapper is gone; "Open" is a real navigating `<Link>`.
- **`isDestination` undefined-while-loading** — `connection-pricing-sync-page.tsx` returns its loading/error states before computing `isDestination`/`isSource`, so those are never read while `connection` could be undefined.
- **Raw `<button>` in `ErrorState`'s `action`** — doesn't reproduce; every `action` in this PR's touched files already passes the shared `Button` component.
- **`useDestinationPricingSyncSummaries`'s `enabled` gating** — present, gated on the dialog's own `open` state.

**Newly fixed this round:**

- **BLOCKING — deep link pre-arms an unsaved write.** Landing on `?source=src-1` for a currently-inheriting source pre-populated `customSources` and cloned the default rule into `draft`, exactly like a manual checkbox click. Combined with the (correct) #3166 `isDirty` fix, that made bare navigation itself read as "you've changed something and haven't saved it yet" — no operator act at all — and a careless Save would persist an override nobody asked for. Fixed with a `baseline` snapshot captured right after the deep-link pre-population (and re-captured on a successful save); the dirty checks diff against that instead of raw server truth, and `Discard` now reverts to it rather than to the un-expanded server read (so Discard doesn't also erase the reason the operator followed the link). New tests cover: no unsaved-bar on bare arrival, the bar appearing once the operator actually edits, and Discard returning to the pre-expanded baseline.
- **SUGGESTION — rollup's empty state ate the section heading.** `SourceConnectionPricingRollup`'s `entries.length === 0` branch now swaps only the row list for `EmptyState`, keeping "How your prices get adjusted" and its explanatory copy visible in both states.

**Not changed:** `useDestinationPricingSyncSummaries` carries no `staleTime`, matching its `useProductsBatchQuery` precedent exactly (which also has none) — leaving as-is rather than diverging from the established pattern for a low-priority suggestion.

Independently verified (resource-constrained machine, no pre-commit hook): `pnpm --filter @openlinker/web lint` (0 errors, unchanged pre-existing warning count), `pnpm --filter @openlinker/web type-check` (clean), `pnpm check:invariants` (all green), and `vitest run` across `features/connections`, `features/price-changes`, `pages/connections`, `app/routes` — 53 files, 682 tests, all passing.

norbert-kulus-blockydevs:
Independent re-verification of this review, after merging the just-updated `3149-connection-pricing-sync-settings-page` (now including #3166's own review fixes and its `toWireRule` return-type pin).

## The review's own findings — confirmed already resolved

Read the actual code in full, not just the prior fix-round's prose:

- **BLOCKING (inherited, #3166) — `isDirty`/`customSources`.** Genuinely fixed — `isDirty`/`isCustomSourcesDirty` diff against a `DraftBaseline` captured right after any deep-link pre-population, never against raw server truth.
- **BLOCKING (this PR's own) — the deep link pre-arming an unsaved write.** The `baseline` mechanism does exactly what's claimed: `pricing-and-sync-section.tsx` seeds `draft`/`customSources` AND `baseline` identically on arrival (including a `?source=` pre-expansion), so `hasUnsavedChanges` reads `false` on bare navigation and only flips once the operator actually edits something from there. `handleDiscard` reverts to `baseline` (the pre-expanded state), not to a fresh `query.data` read, so Discard doesn't also erase the override the operator followed the link to see. Traced the full effect chain (seed effect, post-save re-seed, the two dirty checks, the discard handler) — it's sound.
- **IMPORTANT — picker "Loading…" forever.** `PricingRulesPickerDialog` branches each row on `summaryQuery?.status` (`success`/`error`/anything else), not on `.data` alone — a permanently-failed read now reads "Couldn't load this connection's rule." instead of an eternal "Loading…", and the override-count clause is correctly omitted only in the success branch.
- **IMPORTANT — nested interactive elements.** The `role="button"` wrapper is gone; each row is a plain `<div>` with a sibling real `<Link>` ("Open") — no nested tab stops, no double-activation.
- **SUGGESTION — rollup empty state eating the heading.** `SourceConnectionPricingRollup` now swaps only the row list for `EmptyState`; "How your prices get adjusted" and its explanatory copy render in both states.

## What the merge itself surfaced (not part of the review, but real)

1. **A real type error.** #3166's just-landed `toWireRule` return-type pin narrowed `PriceChangeRuleSummary` (in `price-changes.types.ts`) to require concrete `percent`/`rounding`. This branch's own `PricingRulesPickerDialog` and `SourceConnectionPricingRollup` were calling `ruleLabelFor` with the optional-field wire `PricingRule` directly — `tsc -b` failed on both. Fixed by adding a small `toRuleSummary` helper (the same `percent ?? 0` / `rounding ?? 'none'` defaulting `toWireRule` already uses) and applying it at both call sites.
2. **Two real logic bugs in a freshly-rebased sibling fix.** The parent's #3165 "reject ambiguous separators" commit had `parseLocalizedAmount('1,234,567')` replace *every* comma with a literal dot (producing the invalid `'1.234.567'` and silently returning `NaN` for a perfectly good number) instead of stripping them as unambiguous thousands grouping — the dot branch already did this correctly, the comma branch didn't mirror it. Also, `parseLocalizedAmount('')` returned `0`, not `NaN`. Both had their own (newly-merged) tests actually asserting the correct behavior and failing against the actual code. Fixed to mirror the dot branch's structure and added an explicit empty-string guard.
3. **A stale test.** `edit-price-change-dialog.test.tsx`'s "clamps a manual price to the numeric(14,4) storage precision" test asserted the *pre-improvement* 4-decimal clamp (`399.1235`), contradicting the component's own docblock, which documents clamping to the destination currency's minor-unit exponent (2 decimals for the test's PLN fixture → `399.12`). Corrected the test to match the documented, intended behavior.

## Verification

`pnpm --filter @openlinker/web lint` (0 errors, 123 warnings — unchanged), `pnpm --filter @openlinker/web type-check` (clean), `pnpm check:invariants` (all green), and `vitest run` across `features/price-changes`, `features/connections`, `pages/connections`, `app/routes` — 54 files, 699 tests, all passing.

norbert-kulus-blockydevs:
Caught this branch up with the parent (#3149 / branch \`3149-connection-pricing-sync-settings-page\`), merged in (not rebased) so the stacked PR on top of this one stays intact.

Parent's catch-up brought in two upstream CI fixes:
- the missing \`PriceChangeDetectionService\` barrel export (\`libs/core/src/listings/services/index.ts\`)
- the corrected lane-count assertion in \`handler-registration.service.spec.ts\`

Merge was clean — no conflicts, and neither changed file touches anything this PR owns.

Re-verified after the merge:
- \`pnpm --filter @openlinker/web lint\` / \`type-check\` — clean (0 errors)
- \`pnpm --filter @openlinker/core type-check\`, \`@openlinker/api\`, \`@openlinker/worker\` — all clean
- \`pnpm check:invariants\` — all green
- Specifically re-checked \`parseLocalizedAmount\` / \`formatDeltaLabel\` / \`clampToStorablePrecision\` in \`price-change-copy.ts\` (untouched by this merge) against the boundary cases already fixed here: \`'1,234,567'\` still resolves via the "more than one separator ⇒ strip as grouping" branch, and empty string still short-circuits to \`NaN\` rather than falling through to \`Number('') === 0\`. Both previously-fixed bugs are intact.

Pushed as a normal merge commit (no force-push).

piotrswierzy (commented):
## Tech Lead review

Re-reviewed at `e7e5da3`. The two prior IMPORTANT picker findings are properly closed — each row now branches on `summaryQuery.status`, so a permanently-failed read reads *"Couldn't load this connection's rule."* instead of an eternal "Loading…" and no longer asserts "no overrides" about a connection it couldn't read; the `role="button"`-wrapping-a-`<Button>` nesting is gone and "Open" is a real `<Link>`. The prior BLOCKING deep-link finding is half closed: the `DraftBaseline` mechanism is well-reasoned, correctly cloned on both seed paths, and `handleDiscard` reverting to it rather than to raw server truth is the right call. Credit is also due for two genuine drive-by fixes with failing tests behind them — `parseLocalizedAmount('1,234,567')` returning `NaN` and `parseLocalizedAmount('')` returning `0` — which are exactly why CI's `Test` job is **green here and red on the base branch**. The capability-based page branching (both sections render when both capabilities are present, rather than either/or) is the right model and is now tested.

One blocking item, mechanically verified: **the rollup — this PR's headline deliverable — renders entirely unstyled.**

### Findings

**[BLOCKING]** — `apps/web/src/features/price-changes/components/source-connection-pricing-rollup.tsx:77-107`
> Every class this component renders is undefined. `.settings-section`, `.settings-section__head`, `.settings-section__desc`, `.source-list`, `.source-row`, `.source-row__head`, `.source-row__name`, `.source-row__summary` — I grepped the full 24,350-line `apps/web/src/index.css` at this head for each, unanchored: `settings-section` appears at lines **20332, 24164, 24168 only, all inside CSS comments**; `source-row` appears at 24164 (a comment) and then only as `.pricing-sync__source-row*`; `source-list` only as `.pricing-sync__source-list`. There is no definition of any of them.
>
> The cause is a cross-PR regression inside this stack: #3166's review asked it to namespace its block, and the comment it left at `index.css:24168` says so in as many words — the mockup's `.settings-section` *"was only ever referenced in a comment, never defined"*. Renaming that block to `.pricing-sync__*` is what removed the definitions this component was written against. `.mini-list` / `.mini-row` (the picker) **are** defined at 2247/2255, so the picker is fine and only the rollup is affected.
>
> Nothing in the gates can see this: `check-design-tokens.mjs` is catalog→CSS only, `check-css-structure.mjs` checks well-formedness, and the rollup's own tests assert text content and the *absence* of write affordances — never that a class does anything. So `lint`, `type-check`, `check:invariants` and `Test` are all green over a page that renders as an unstyled stack of divs.
>
> Fix: ship the rollup's own namespaced block following the `.pricing-sync__*` pattern #3166 established, and while adding it register the `.source-row` height in `docs/frontend-ui-style-guide.md § Density & Row Heights` — that section's table is where non-`DataTable` rows are documented (the fulfilment who-decides row is the precedent), and both prior reviews flagged the omission.

**[IMPORTANT]** — `pricing-and-sync-section.tsx:270-300` and `:452-470`: the baseline hid the *signal*, not the *effect*
> Arriving at `?source=X` for an **inheriting** source still adds `X` to `customSources` and clones the default into `draft.sources[X].effective`; `baseline` then captures that seeded state, so `hasUnsavedChanges` reads `false`. That correctly kills the spurious unsaved bar. But `persistSave` builds `sourceOverrides` from `customSources` unconditionally, and `overridesAboutToBeDropped()` guards only **removal** — there is no counterpart for creation. So: follow the rollup's **Manage** link → change something unrelated (the default's rounding) → the bar appears saying only *"You've changed something"* → Save, and a brand-new per-source override is persisted for `X`, pinning it to today's default so every later change to the default silently stops reaching it. No confirmation, and the operator never ticked the box.
>
> The two entry points differ in intent, which is what makes this worth separating rather than accepting: #3148's permalink is labelled *"Set a rule just for this source"* — intent is explicit and pre-arming is right. The rollup's button is a bare **"Manage"** on a row whose own summary reads *"— using {destination}'s default rule"*; there, pre-arming a write is not what the label promised. Carry the intent in the URL (`?source=X&create=1`, honoured only by the explicit permalink) and have the neutral Manage link scroll/highlight without ticking — or extend the existing confirm to cover an override being *created* by navigation rather than by a click.

**[IMPORTANT]** — `source-connection-pricing-rollup.tsx:80`: `<h2>` collides with the page title's own level
> `PageLayout` renders its title as `<h2 className="page-title">` (`shared/ui/page-layout.tsx:28`), and `PricingAndSyncSection` uses `<h3>` for its four section headings. On a connection carrying both capabilities this page emits h2 (title) → h3 ×4 → **h2** (this), so a screen reader hears the rollup as a new top-level region arriving after the subsections. #3166 took exactly this fix on review (*"the four bare `<h2>`s changed to `<h3>` matching every sibling section"*); it was not carried across. Use `<h3>`.

**[IMPORTANT]** — a `ProductMaster`-only source connection cannot reach this page from its own detail page
> This PR widens `ConnectionPricingSyncPage` to serve a `ProductMaster` connection (`connection-pricing-sync-page.tsx:78`), but the "Pricing & sync" button #3166 added to `connection-detail-page.tsx:251` is still gated on `OfferManager || ProductPublisher`. So the persona the rollup exists for — issue #3150's *"an operator who navigated to a SOURCE connection isn't stranded wondering how its prices get adjusted downstream"* — can reach it only through the queue table's `row-source-tag` (correctly repointed here, `price-changes-queue-table.tsx:699`), i.e. only when a price change from that source happens to be pending. On a healthy queue the page is unreachable from the source connection entirely. Widen that gate to include `ProductMaster` — ideally by extracting the single shared predicate I flagged on #3166, which this now makes a fourth spelling of.

**[IMPORTANT]** — the PR description describes a different PR, and the mockup obligation is unmet
> The Summary still claims this PR builds the `connections/:connectionId/pricing-sync` route and demotes the `EditConnectionForm` copy to a link-out, and "Corrections made along the way" is written entirely about that work. Per the consolidation comment, both files were **deleted** in favour of #3166's, and neither `connection-pricing-sync.route.tsx` nor `EditConnectionForm.tsx` appears in this diff. The unchecked `pnpm test — could not run locally … relies on CI` is also stale: CI runs, passes here, and this PR is what turns the base branch's red `Test` job green — that's a real contribution the description doesn't claim. Separately, `docs/frontend-architecture.md § UX Mockups` requires an implementing PR for a feature with a committed mockup to *"either touch the mockup file, or say in its description why the mockup is still accurate without a change"*; `docs/plans/mockups/price-changes-review-queue.html` is neither touched nor mentioned, and that rule says a reviewer treats such a PR as incomplete.

**[SUGGESTION]** — `pricing-rules-picker-dialog.tsx:94`: `onClick={() => onOpenChange(false)}` fires on modified clicks too
> The docblock credits making Open a real `<Link>` so *"Ctrl/Cmd-click, middle-click, and open in new tab all work"* — they do navigate, but they also close the dialog the operator was still browsing. Guard on `event.metaKey || event.ctrlKey || event.shiftKey || event.button !== 0`.

**[SUGGESTION]** — `pricing-and-sync-section.tsx:590`: the stale-`?source=` notice is styled as a validation error
> `className="pricing-sync__field-error"` gives red error styling to a navigational condition, while `role="status"` (polite) says it is not an alert — the two disagree. A neutral hint class matches what it actually is.

**[SUGGESTION]** — `use-destination-pricing-sync-summaries.ts`: no `staleTime`
> Sharing `connectionPricingSyncQueryKey` with the single-connection query is a nice touch (opening the picker warms the cache the settings page then reads). Without a `staleTime` each reopen refetches N. The author declined this to stay aligned with `useProductsBatchQuery`, which also has none — reasonable; recording it once rather than pressing.

### Note on CI (not findings against this PR)
`Test` **passes** here (and is red on the base — this PR contains the fix). `Integration Tests` fails on `apps/api/test/integration/listings-price-change-episode.int-spec.ts` with `invalid input syntax for type uuid: "no-such-id"` — a backend spec from #3162/#3163 earlier in the stack, unrelated to any file in this frontend PR, but it must be green before the epic reaches `main`.

### Verdict
❌ **Request changes** (posted as a comment per review policy)

1. Ship the rollup's CSS (namespaced, `.pricing-sync__*` pattern) and register the row height in § Density & Row Heights — blocking.
2. Separate "navigate here and look" from "navigate here and create an override", so a neutral **Manage** link cannot persist a rule the operator never ticked.
3. `<h3>` for the rollup's heading; widen the detail-page button gate to include `ProductMaster`.
4. Rewrite the PR description to match the PR, and either update the mockup or state why it is still accurate.

---

norbert-kulus-blockydevs:
All four numbered items plus every suggestion — `aacad8445`, on the parent catch-up `7eb9ded91`.

**1. BLOCKING — the rollup rendered classes that do not exist.** Confirmed: eight of them, with zero rules in `index.css`. I audited every class the component renders rather than trusting the list, and the only ones that resolved were `button`, `button--secondary` and `button--xs`.

I gave it its own `.pricing-rollup__*` block rather than folding it into `.pricing-sync__*`. The pattern is the one you asked for; the prefix is not, on purpose — `.pricing-sync__*` styles a DESTINATION's editable settings and lives in another feature, while this is a SOURCE's read-only summary. Sharing the prefix would couple two surfaces that should be free to diverge. It also removes the collision your finding implies but does not name: `.source-list` and `.source-row` are rendered *both* by this component (bare) and by `PricingAndSyncSection` (as `.pricing-sync__source-list` / `-row`), so the generic names were one stylesheet edit away from the `.table-wrap` failure from #3164. Row height registered in § Density & Row Heights as a non-`DataTable` carve-out.

**2. IMPORTANT — the neutral Manage link staging an override.** This was the one I found most convincing, and it is fixed the way you framed it: the intent travels in the URL. `?source=` scrolls and highlights; `?source=&override=1` stages. `PricingAndSyncSection` takes two props instead of one, and scrolling deliberately follows the VIEW request so a bare link still lands on the row it names.

Worth recording: `?source=` has exactly one producer today — the rollup's Manage link — so before this change every deep link into that page was the neutral one, and every one of them staged. The intentful `&override=1` has no producer yet; the "Set a rule just for this source" permalink the page docblock attributes to #3148 does not exist in the code.

**3. `<h3>` and the ProductMaster gate.** Both done. The gate is a second exported predicate, `hasPricingSyncPage`, sitting beside #3166's `isPricingDestination` and deliberately wider — the page carries a source's rollup as well as a destination's settings, and its docblock says so, so the next person does not "unify" them.

**4. Description and mockup.** Description rewritten: the `initialExpandSourceId` bullet now describes the two props that actually exist and says why, and the round-3 changes are recorded. The mockup is not touched here and I would rather say why than claim it: `#page-connection` models the destination's editable settings, which this PR does not change, and it has no source-page screen for the rollup to diverge from. Adding one is a design decision rather than a correction — tell me if you want it and I will draw it.

**Suggestions, all taken:** modified clicks on the picker's Open link are left alone; the stale-link notice moves to a neutral `.pricing-sync__notice` so its styling and its `role="status"` stop saying different things (and its copy now matches the new behaviour — nothing is "pre-selected" for a bare link any more); the summaries query gets a 60s `staleTime`, which is safe precisely because `useUpdateConnectionPricingSyncMutation` and `useSetSourceSyncModeMutation` both invalidate that exact key.

**One superseded carry-over, recorded so it does not read as a silent revert:** #3166 landed a `baseline = updateMutation.data ?? query.data` fix for the dirty check. This branch's own `DraftBaseline` is strictly better — it sets the baseline in both the mount seed and the post-save effect, and covers the deep-link pre-population case too — so the merge takes yours and drops mine.

Verified: `type-check` clean, `check:invariants` clean, 574 tests passing across the three affected folders.

piotrswierzy (approved):
## Tech Lead re-review (`e7e5da3` → `68cd441`)

Delta is one own commit, `aacad84` (*"ship the pricing rollup's stylesheet, and split look from create"*), on a parent catch-up merge. **The BLOCKING finding is closed, all four IMPORTANTs are closed, and all three SUGGESTIONs were taken.** I re-ran the mechanical check that produced the block rather than reading the description: every class `source-connection-pricing-rollup.tsx` renders — `.pricing-rollup`, `__head`, `__desc`, `__list`, `__row`, `__row-head`, `__row-name`, `__row-summary` — now resolves to a real rule at `index.css:2509-2568`, outside any comment, all eight of them, using only `var(--space-*)` / `var(--border-*)` / `var(--text-*)` / `var(--bg-*)` / `var(--radius-*)` tokens with no raw hex and no raw px. Renaming the block to its own `.pricing-rollup__*` namespace rather than reusing `.pricing-sync__*` is the right call and the comment states it: the latter styles a *destination's* editable settings in another feature, and two of the old generic names collided with it outright. The row height is registered in `docs/frontend-ui-style-guide.md § Density & Row Heights` alongside the who-decides and sales-document-market carve-outs, as asked.

The deep-link split is real and correctly wired, not just described: `PricingAndSyncSection` now takes two props, the seeding effect stages an override only for `initialCreateOverrideForSourceId` (with `?source=` explicitly commented as *"scrolls and highlights; it must not stage anything"*), and the page maps `?source=` → expand, `?source=&override=1` → create (`connection-pricing-sync-page.tsx:106-118`). The rollup's Manage link emits the neutral `?source=` form, so the path I walked — neutral Manage → unrelated edit → Save persists an override nobody ticked — is now unreachable. The heading is `<h3>`, matching the four sibling section headings under `PageLayout`'s own `<h2>`. And the detail-page gate widened to a second, deliberately-wider predicate `hasPricingSyncPage` (`isPricingDestination || ProductMaster`), sitting beside #3166's in one file with the asymmetry documented — so a `ProductMaster`-only source connection can finally reach its own rollup from its own page rather than only via a pending queue row. Suggestions all taken: the picker's Open link returns early on meta/ctrl/shift/alt/non-primary clicks, the stale-`?source=` notice uses `.pricing-sync__notice` instead of the field-error red while keeping `role="status"`, and the picker summaries carry a 60 s `staleTime` justified by both mutations invalidating that exact key.

### Findings

**[SUGGESTION]** — `initialCreateOverrideForSourceId` has no producer anywhere in the tree
> I searched every file under `features/price-changes/` and `pages/connections/` at `68cd441`: `override=1` appears only in `connection-pricing-sync-page.tsx` (which consumes it) and its own test. The `#3148` Edit-dialog permalink the split was designed around — *"Set a rule just for this source"* — emits no `&override=1` at this head; `edit-price-change-dialog.tsx` carries no `pricing-sync` link at all, and `price-changes-queue-table.tsx:751,765` both emit the neutral form. So the create half is currently reachable only by hand-typing the URL. This errs in the safe direction (nothing pre-arms a write) and is not worth blocking on, but it means half the mechanism is unexercised in the product; either wire the explicit permalink to it or note in the docblock that it awaits one.

**[SUGGESTION]** — the PR description's opening sections still describe #3166's work
> The appended *"What the round-3 review changed"* block is accurate and complete — credit for that. But the original **Summary** still leads with *"Builds the ... `connections/:connectionId/pricing-sync` route"*, and the whole *"Corrections made along the way"* section is written about building that route and demoting the `EditConnectionForm` copy — work that lives in #3166, whose files do not appear in this diff. #3166 handled the same problem by correcting its stale bullet **in place** and naming the round it cost; doing the same here would leave one durable record rather than a stale one followed by a correct one. Relatedly, `docs/frontend-architecture.md § UX Mockups` still wants this PR to either touch `docs/plans/mockups/price-changes-review-queue.html` or say why it is still accurate — it does neither (the `+15/-1` on that file in the compare range is base catch-up, not `aacad84`). A one-line statement covers it.

### Verdict
✅ **Approve**

The headline deliverable now renders as designed, and the write-by-navigation hazard is gone. Both remaining items are description/coverage hygiene, not code.

CI at the time of writing: `Type Check`, `Docker Build Smoke Test`, `PHP Unit Tests` pass; `Lint`, `Test`, `Build`, `Integration Tests` queued on the current run. `mergeable: MERGEABLE`, `mergeStateStatus: BLOCKED` (required checks + review). The `Integration Tests` failure I flagged last round (`listings-price-change-episode.int-spec.ts`, `invalid input syntax for type uuid`) belongs to #3162/#3163 further down the stack and was addressed there in this range (`fe05ab79`, *"spell the price-change no-match fixtures as well-formed uuids"*) — worth confirming green before the epic reaches `main`, but not attributable to this PR.

---

norbert-kulus-blockydevs:
Both taken — `3ebae63d9` plus a rewritten description.

**The unproduced create half.** Your search matches mine: `&override=1` is emitted by nothing. The prop's docblock now says so outright — that the queue table's two links and the rollup's `Manage` are all the neutral form, that the "Set a rule just for this source" permalink the split was drawn around does not exist in `edit-price-change-dialog.tsx`, and that the create half is therefore reachable only by typing the URL. I chose the docblock over wiring the permalink: building that link is #3148's surface, not this one, and adding it here would put a new affordance into a PR that is already approved.

**The description.** Rewritten from the top. The Summary no longer claims the route or `PricingAndSyncSection` — it says plainly that those are #3166's and that this PR adds the rollup, the picker and the two deep-link props. It also says the earlier version claimed them wrongly, rather than quietly swapping the text.

piotrswierzy (approved):
## Tech Lead re-review (`7cd25770` → `caa1493e`)

Ten commits in the range, **six of them merges** pulling the stack's bases forward, plus `#3166`'s two doc commits arriving through the base merge. The PR's own new work is **one comment-only commit**, `3ebae63d`, answering my remaining note on `initialCreateOverrideForSourceId`. Net file diff base…head: `+16/-0` in `pricing-and-sync-section.tsx` and `+8/-4` in `price-change-detection.service.ts` — **zero executable lines**. **The approval stands.**

**The `.pricing-rollup__*` grep, re-run mechanically at `caa1493e`** (this is the defect class that has now appeared twice, and `check-design-tokens.mjs` is one-directional, so it is checked and not assumed):

- `source-connection-pricing-rollup.tsx` renders **8** distinct class names (`pricing-rollup` plus seven `__*`).
- `index.css` defines **all 8**, at lines **2509–2568** — `.pricing-rollup` (2509), `__head` (2515), `__desc` (2522), `__list` (2529), `__row` (2538), `__row-head` (2548), `__row-name` (2556), `__row-summary` (2562).
- Every one is a real rule **outside** a comment (the preceding block comment closes at 2508), in its own namespace with no `.pricing-sync__*` collision, and **tokens-only** — `--space-2/3`, `--border-default`, `--radius-md`, `--bg-surface`, `--text-primary/secondary`, each confirmed declared in `index.css`. No raw hex. The `.pricing-rollup__row` ~52 px density registration comment is intact.

Unchanged and correct.

### Findings

**Closed from last round** — `3ebae63d` records that `&override=1` has no producer anywhere in the tree, names the three call sites that use the neutral form, names the missing "Set a rule just for this source" permalink as the intended producer, and says the half errs safe. That is the right resolution: it documents an unexercised mechanism rather than deleting a prop a sibling issue will need, and a reader can no longer mistake it for wired.

### Verdict
✅ **Approve**

CI at `caa1493e`: `MERGEABLE` / `UNSTABLE`. Build, Docker Build Smoke, **Test (15m15s)**, Type Check, PHP Unit and path detection **pass**; Lint and Integration Tests **pending (queued, not run)** — nothing red. Targets `3149-…`, so it merges after #3166.

---

piotrswierzy (approved):
## Tech Lead re-verification (head unchanged at `0bccc8c`)

Head unchanged, `main` unmoved. CI on run 34845744729: **Type Check, Docker Build Smoke Test, PHP Unit Tests and Detect changed paths pass; Build, Lint, Test and Integration Tests are queued behind the runner backlog. No red on any run.**

**Re-ran the `.pricing-rollup__*` grep against the PR head's `index.css`, not against a cached read** — this class of defect has bitten five PRs in this programme and a base-branch merge is how the definitions were deleted the first time. All eight classes resolve to real rules at `index.css:2509-2568`, **outside the comment block** (the header comment ends at 2508; `.pricing-rollup` opens at 2509, then `__head` 2515, `__desc` 2522, `__list` 2529, `__row` 2538, `__row-head` 2548, `__row-name` 2556, `__row-summary` 2562). Every class the component renders is in that set and vice versa; own namespace, deliberately not reusing `.pricing-sync__*`; declarations are tokens-only (`--space-*`, `--border-default`, `--radius-md`, `--bg-surface`, `--text-primary`, `--text-secondary`) with no raw hex.

Stack: base `3149-connection-pricing-sync-settings-page` (#3166). **#3159 and #3162 below still carry an open BLOCKING** (in-place edit to the already-merged migration `1880000000000-create-price-change-episodes.ts`, diverging an already-migrated database from a fresh install). No migration here — **the chain is held on those two.** `mergeable: MERGEABLE`, `mergeStateStatus: UNSTABLE`.

### Findings

### Verdict
✅ **Approve**

---

PR #3168, opened 2026-09-10
feat(web): recently-applied-automatically note + dialog

### LINKED ISSUE #3151: feat(web): recently-applied-automatically surface for Automatic connections
## Problem / Context

Part of #3139 (epic). **Stack position: twelfth — branches from #11's branch.**

Build the lightweight "recently applied automatically" surface from `docs/plans/mockups/price-changes-review-queue.html` — the deliberately-minimal stand-in for a full "Daily digest" mode, which was cut during design (ADR decision 3) because no notification/summary surface exists anywhere in the codebase. This issue is the entire scope of that decision — do not expand it into a dedicated page/screen; that is explicitly out of scope and, per the epic's decision, has no tracked follow-up issue (it's a code comment in the mockup and will be a code comment in the real implementation, not a TODO issue).

On the Price-changes tab (#8), above the filter bar:
- `#auto-applied-note` (`.filter-note`, hidden by default) — appears ONLY when at least one connection or source-override has `mode: 'automatic'` in effect. Text: "{N} price changes went live automatically today (see them)" — `#auto-applied-count` / `#auto-applied-link`.
- Clicking `#auto-applied-link` opens `#dialog-auto-applied` (`data-state="dialog-auto-applied"`): title "Applied automatically today", desc "These connections publish without asking, so nothing here needs a decision from you.", `#dialog-auto-applied-list` — a simple `.mini-list` (same component the bulk-accept dialog uses for its item list, #9 — reuse it, don't rebuild it) of recently-applied items: thumbnail-initials, name, connection + relative time, old→new price. `#dialog-auto-applied-close`.
- Backend: `GET /listings/price-changes/auto-applied` (#6).
- **This is explicitly NOT paginated, NOT filterable, NOT a dedicated page.** It's a small, recent-N list (mirror the mockup's 2-item static example scope — a reasonable real limit is the API's choice, e.g. last 20 or last 24h, whichever #6 implemented) opened from a link, closed with a Close button. If a future need for more (search, date range, export) emerges, that is new design work, not an extension of this component.

## Classification

**Type**: Frontend
**Layer**: Interface (feature component)
**File(s)**: `apps/web/src/features/price-changes/components/auto-applied-note.tsx`, `auto-applied-dialog.tsx`

## Dependencies

- #11 (source rollup + picker — not a hard technical dependency, stacked for linearity)
- #6 (`GET /listings/price-changes/auto-applied`)

## Assumptions

- "At least one connection or source-override has automatic mode" is computed client-side from whatever connection/pricing-sync data is already being fetched for the page (mirrors mockup's `anyConnectionAutomatic()` helper) — confirm this doesn't require an extra network round-trip; if the connections list already carries `pricingSync.default.mode` and `sourceOverrides[*].mode`, compute from that, otherwise ask #7 to add a lightweight `hasAutomaticConnections: boolean` flag to whatever summary endpoint the Listings page already loads.

## Acceptance Criteria

- [ ] Banner appears only when genuinely applicable (at least one automatic connection/override exists) and is absent otherwise — verify both states
- [ ] Modal reuses the existing `.mini-list` component from #9, not a new list implementation
- [ ] No pagination, filtering, or dedicated route is built — this stays a link + small modal
- [ ] Tests: banner visibility toggle (present/absent) and modal open/populate/close

norbert-kulus-blockydevs:
## Summary

Part of #3139 (epic). Stack position: twelfth (final frontend issue) — branches from #3150's branch (PR #3167, still open).

Builds the entire scope of ADR-072 decision 3 — the deliberately-minimal stand-in for a "Daily digest" mode that was cut because no notification/summary surface exists anywhere in the codebase:

- `AutoAppliedNote` — a small banner above the queue's filter bar, shown only when `GET /listings/price-changes/auto-applied` (#3145) returns at least one entry.
- `AutoAppliedDialog` — opened from the banner's "see them" link, reusing the existing `.mini-list`/`.mini-row` shape from #3148's bulk-accept dialog rather than a new list implementation.
- Not paginated, not filterable, not a dedicated page/route, per the issue's explicit scope boundary — nothing here should be extended.

## Two things resolved, both flagged rather than silently assumed

1. **No product name in the auto-applied log contract.** `PriceChangeAutoAppliedItem` (#3145) carries only `productVariantId`, so the dialog batches a per-item variant-summary lookup (`GET /products/variants/:id`) for a display name/SKU, falling back to the bare id if a lookup fails or is still loading — a lookup failure never blocks the dialog itself.
2. **The show/hide condition and the wording.** The mockup's `anyConnectionAutomatic()` checks connection/override CONFIG directly, which would need an extra batched per-connection pricing-sync fetch with no cheaper existing signal. "At least one entry exists in the recent auto-applied log" is a strictly narrower proxy backed by data the dialog needs to fetch anyway (if anything was auto-applied, a connection must be in Automatic mode), so the banner keys off that instead of a second round-trip. The copy also changed from "...today" to "...recently", since the backend (`findRecent(limit)`) is a plain most-recent-N read with no calendar-day boundary — asserting "today" would misstate what an older entry in the last-N window actually is.

## Test plan

- [x] `pnpm --filter @openlinker/web type-check` — clean (pre-existing unrelated `@tiptap/*` errors only)
- [x] `npx eslint` on all changed/new files — clean
- [x] `node scripts/check-design-tokens.mjs` — OK
- [ ] `pnpm test` — could not run locally: this environment is missing `@tiptap/*` node_modules (pre-existing, affects every FE test file); relies on CI
- New `auto-applied-note.test.tsx` covers both required cases: banner absent with zero recent entries, and banner-present → open → populated mini-list (name, connection, relative time, old→new price) → Close

Closes #3151


apps/web/src/features/price-changes/components/auto-applied-note.tsx
```diff
@@ -0,0 +1,70 @@
...
+  const variantIds = items.map((i) => i.productVariantId);
+  const variantSummaries = useAutoAppliedVariantSummaries(variantIds);
+
+  if (items.length === 0) return null;
+
+  const connectionLabelById = new Map((connectionsQuery.data ?? []).map((c) => [c.id, c.name]));
+
+  const rows: AutoAppliedRow[] = items.map((item, index) => {
+    const variant = variantSummaries[index]?.data;
+    const label = variant?.name ?? variant?.sku ?? item.productVariantId;
```
norbert-kulus-blockydevs, line 42: **BLOCKING — the dialog can never show a product name, and the test fixture makes that invisible.**

`variant?.name ?? variant?.sku ?? item.productVariantId` reads `ProductVariantSummary.name`, which is **not a product name**. It is derived server-side from the variant's *attribute map*: `apps/api/src/products/http/products.controller.ts:598` — `name: attributeValues.length > 0 ? attributeValues.join(' / ') : undefined`, documented at `:590` as *"Builds a human label from the variant's attribute map … falls back to undefined so the FE can render the SKU as the primary label."*

So the best case renders `"350 ml"` / `"Red / 42"`; a simple product with no attributes falls to the SKU; a variant with neither renders a raw `ol_variant_*` UUID. The mockup renders `"Ceramic Coffee Mug — 350 ml"`; this cannot produce the product name at any input.

The sibling read already solved it: `PriceChangeItem` carries `productName`, `variantLabel` and `sku`, resolved in `price-changes.service.ts:81` via `getVariantsByIds` + `getProductsByIds` — in the **same service file, ~120 lines above** `listAutoApplied`. So the body's rationale ("rather than blocking #3145 on a wider log-entry contract") understates how small the backend change is.

**Why review cannot catch it downstream:** `auto-applied-note.test.tsx:40` mocks `name: 'Ceramic Coffee Mug — 350 ml'`, a value the real endpoint cannot emit. That is `docs/lessons.md:699` exactly (*"Hand-built fixtures let a test assert against a state the state machine forbids"*) and its FE twin at `:349`. Meanwhile `test-utils.tsx:833` defaults `products.getVariant` to a **404 rejection**, so the realistic path is the one no test asserts.

**Fix:** add `productName`/`variantLabel`/`sku` to `PriceChangeAutoAppliedItemResponseDto` and populate them in `listAutoApplied` from the pair already imported into that service — which deletes `use-auto-applied-variant-summaries.ts` entirely and resolves the N+1 with it. If the backend must stay frozen, at minimum change the fixture to a realistic `name: '350 ml'` plus a `name: undefined, sku: null` case.

**Related, same line:** with `retry: false`, a *failed* lookup, a *still-loading* lookup, and a variant that legitimately has neither name nor SKU all render the same bare UUID — three states collapsed into one, plus a visible flash as rows swap from UUIDs to labels. `docs/frontend-architecture.md:281`: *"`idle` and `unavailable` are different, and a surface must render them differently."*

apps/web/src/features/price-changes/components/auto-applied-note.tsx
```diff
@@ -0,0 +1,70 @@
...
+export function AutoAppliedNote(): ReactElement | null {
+  const [dialogOpen, setDialogOpen] = useState(false);
+  const query = useAutoAppliedPriceChangesQuery();
+  const connectionsQuery = useConnectionsQuery();
+
+  const items = query.data ?? [];
+  const variantIds = items.map((i) => i.productVariantId);
+  const variantSummaries = useAutoAppliedVariantSummaries(variantIds);
+
+  if (items.length === 0) return null;
```
norbert-kulus-blockydevs, line 36: **IMPORTANT — the show/hide justification is falsified, and the config signal is already in cache three lines above.**

The PR body says `anyConnectionAutomatic()` *"would need an extra batched per-connection pricing-sync fetch with no cheaper existing signal."* It would not. `PricingRulesPickerDialog` is mounted **unconditionally** at `price-changes-queue-table.tsx:263` and calls `useDestinationPricingSyncSummaries(connectionIds)` with no `enabled` gate — so `GET /connections/:id/pricing-sync` has already fired for every destination by the time `<AutoAppliedNote />` renders at line 269, **in the same parent**. `ConnectionPricingSyncView` carries `default.mode` and `sources[].effective.mode` — precisely `anyConnectionAutomatic()`'s inputs — and reusing that hook costs **zero** additional requests, since it shares `connectionPricingSyncQueryKey`. Issue #3151's own Assumptions section named this path.

(Half the claim is true — `Connection` carries no pricing fields — but the batched view is on the page regardless.)

**And the proxy does not hold over time.** `findRecent(limit)` is `order: { appliedAt: 'DESC' }, take: limit` with **no `where` clause at all**, and I grepped the tree for a retention/prune sweep on `price_change_auto_applied_log`: **there is none**, and no `JobTypeValues` member or scheduler task references it. So once anything has ever been auto-applied the banner **can never disappear**, even after the operator switches every connection back to Manual. "Narrower at the instant of application" is true; "strictly narrower" is not.

**Fix:** gate visibility on config (over the already-cached summaries) and take the count from the log. Note this depends on #3167's picker hook gaining its `enabled` gate, which it should — so coordinate: if the picker stops over-fetching, this needs its own batched read rather than the free ride.

apps/web/src/features/price-changes/components/auto-applied-dialog.tsx
```diff
@@ -0,0 +1,81 @@
...
+    .slice(0, 2)
+    .join('')
+    .toUpperCase();
+}
+
+export function AutoAppliedDialog({ open, onOpenChange, rows }: AutoAppliedDialogProps): ReactElement {
+  return (
+    <Dialog open={open} onOpenChange={onOpenChange}>
+      <DialogContent>
+        <DialogTitle>Applied automatically today</DialogTitle>
```
norbert-kulus-blockydevs, line 50: **IMPORTANT — the copy correction went half way, and the more prominent half still says "today".**

The banner changed to "recently" on the reasoning that `findRecent` has no calendar-day boundary — correct, and I verified it (no `where` clause at all). But this `DialogTitle` still reads **"Applied automatically today"**, which is the same claim the PR argues is unsupportable, kept in the more prominent position — and `auto-applied-note.test.tsx:57` **pins it**.

`docs/frontend-ui-style-guide.md:913`: *"A quiet lie is worse than the loud failure it replaces."*

Change the title to match the banner, and update the test. ADR-072 decision 3 itself still says "today" — worth an amendment line alongside the fix, since the mockup (`:476`, `:657`, `:934`) is now stale in three places and untouched by the whole stack.

apps/web/src/features/price-changes/hooks/use-auto-applied-variant-summaries.ts
```diff
@@ -0,0 +1,28 @@
...
+ * deliberately minimal, non-paginated surface (ADR-072 decision 3). A
+ * variant fetch failing degrades to the bare id (see `auto-applied-dialog`),
+ * never blocks the dialog.
+ */
+export function useAutoAppliedVariantSummaries(
+  variantIds: readonly string[],
+): UseQueryResult<ProductVariantSummary>[] {
+  const apiClient = useApiClient();
+
+  return useQueries({
```
norbert-kulus-blockydevs, line 20: **IMPORTANT — this is a fan-out, not a batch.**

`useQueries` issues **one `GET /products/variants/:id` per log entry** — bounded at 20 only because the server fixes the limit. The PR body calls it "batches a per-item variant-summary lookup", which reads as the opposite of what it does.

To be fair to the implementation: a batched core read **does** exist (`IProductsService.getVariantsByIds`, whose docblock says *"without a per-id fan-out (#2234)"*) and is called by the very service that produces this log — but it is not exposed over HTTP, so the FE is not wrong to fan out *today*. The remedy is server-side, and it is the same one as the blocking comment above: return the name from `listAutoApplied` and delete this hook.

(#2083's rule — a batched read happens once, before the loop, never inside it — and #1713's `getLatestInvoicesForOrders` are the shipped precedents for the shape.)

apps/web/src/features/price-changes/components/auto-applied-note.tsx
```diff
@@ -0,0 +1,70 @@
...
+  });
+
+  return (
+    <>
+      <div className="filter-note" id="auto-applied-note">
+        <span id="auto-applied-count">{items.length}</span> price change{items.length === 1 ? '' : 's'} went live
+        automatically recently (
+        <a
+          id="auto-applied-link"
+          href="#"
```
norbert-kulus-blockydevs, line 57: **IMPORTANT — `<a href="#">` for an action that opens a dialog, and the banner is announced to nobody.**

`docs/frontend-ui-style-guide.md:574`: *"Use links only for real navigation. If an element submits, confirms, or mutates state, it should be a button."* This is the **only** `href="#"` in all of `apps/web/src` — a mockup artifact, not a repo pattern. AT announces "link"; ctrl/middle-click offers an "open in new tab" that does nothing useful; there is no `aria-haspopup="dialog"`. (The mockup's own `<a id="auto-applied-link">` has no `href` at all, so this was an implementation choice, not inherited.)

**Fix:** `<button type="button" className="button button--ghost">` — the low-emphasis tone already exists at `index.css:629`.

**Separately:** the banner is a plain `<div className="filter-note">` inserted into the DOM after the query resolves, with no announcement. The repo has precedent for exactly this shape — `features/orders/components/generate-label-form.tsx:780` uses `<div role="status" aria-live="polite">` for a status note, as do `shared/ui/list-pagination.tsx:164` and `offer-product-picker-modal.tsx:412`. Add `role="status"`.

apps/web/src/features/price-changes/components/auto-applied-dialog.tsx
```diff
@@ -0,0 +1,81 @@
...
+        <DialogDescription>
+          These connections publish without asking, so nothing here needs a decision from you.
+        </DialogDescription>
+        <div className="mini-list" id="dialog-auto-applied-list">
+          {rows.map(({ item, label, destinationLabel }) => (
+            <div className="mini-row" key={item.id}>
+              <span className="thumb" style={{ width: 24, height: 24, fontSize: 10 }} aria-hidden="true">
```
norbert-kulus-blockydevs, line 57: **IMPORTANT — `.thumb` is unstyled inside `.mini-row`; the avatar renders as bare text.**

`index.css` defines `.thumb` **only** as `.cell-product .thumb` (`:1968`). Grep confirms zero bare `.thumb {` and zero `.mini-row .thumb`. The mockup defines it globally (`price-changes-review-queue.html:170`), which is why the copied markup looked correct.

So the element loses `border-radius`, `background`, `display: grid`, `place-items: center`, `color`, `flex-shrink` and `font-weight` — and the inline `style={{ width: 24, height: 24 }}` does nothing, because a `<span>` is `display: inline` and width/height do not apply to non-replaced inline boxes.

Compounded by the blocking finding above: `initials("350 ml")` → `"3M"`, `initials("MUG-350")` → `"M"`.

**Fix:** add `.mini-row .thumb` (or promote a bare `.thumb` base rule with `.cell-product .thumb` as a size modifier) — which also fixes #3165's `bulk-accept-price-changes-dialog.tsx:66`, which has the identical defect. Given the label problem, consider dropping the thumb entirely: no meaningful initial is derivable from this data.

**Minor, in the barrel:** `AutoAppliedNote`, `AutoAppliedDialog` and `AutoAppliedRow` are all exported but used only within `features/price-changes`. `§ Feature Public Surface`: *"start narrow — adding an export later is one line."* Keep the last two internal.

norbert-kulus-blockydevs (commented):

Several things right: the Radix `shared/ui/dialog` primitive rather than a hand-rolled `role="dialog"` (a real risk on a mockup-derived component); `formatAmount` and `TimeDisplay` with currency read from the payload rather than assumed; the scope boundary genuinely respected (no route, no pagination, no filter) and said so in terms a future reader will act on; both mockup divergences **flagged in the PR body** rather than made silently; a failed `auto-applied` read hides the banner rather than rendering "0 changes"; and the count and the dialog rows derive from one array so they cannot disagree. I also checked and `GET .../auto-applied` **is** properly bounded — `DEFAULT_AUTO_APPLIED_LIMIT = 20`, server-fixed, no `@Query()`.

One blocking (the dialog cannot render a product name, and the fixture hides it) plus five important, inline.

Because this branch is the stack tip I also audited the cumulative state; findings that belong to the stack rather than this diff:

- **`.table-wrap` silently restyles a plugin the stack never touched.** The class does not exist on `main`; the stack defines it plus 13 descendant rules, and `plugins/subiekt/components/subiekt-invoice-correction-flow.tsx:189` already renders `className="table-wrap subiekt-correction__table-scroll"` — that table now inherits border, radius, background, `overflow: hidden` and `min-width: 900px`. Scope or rename.
- **`initials()` is defined three times inside `features/price-changes` alone and the copies have drifted** — only this PR's has `.filter(Boolean)`, so `initials('Acme  Store')` yields `"A"` in two places and `"AS"` in the third. Two more copies pre-date the stack (`app-shell.tsx:191`, `mapping-pairing-bar.tsx:38`).
- **`.activity-list` is now defined twice**, 21.6k lines apart (#3166), and `is-active` on `bulk-publish-progress.tsx:48` is inert — no such rule exists.
- **`window.prompt` residue is clean** — #3164 added one, #3165 removed it (pickaxe-confirmed); the remaining instance in `rich-text-editor.tsx` is unrelated.
- **Untested in the final state:** `bulk-accept-price-changes-dialog.tsx`, `bulk-publish-progress.tsx`, `lib/price-change-copy.ts`, all 12 hooks, both `api/` modules.
---
*Part of a whole-stack review of #3158–#3168 (epic #3139). Cross-cutting items tracked once: `pnpm lint` is red on the stack; `docs/architecture-overview.md § Listings` is untouched by all ten PRs though ADR-072 names it the primary doc section; the mockup has not changed since #3155 despite ~20 behavioural divergences; `features/price-changes` is registered in **neither** `.eslintrc.js` `no-restricted-imports` group, so the new barrel is unenforced; and `.source-row` / `.mini-row` are non-`DataTable` rows absent from `docs/frontend-ui-style-guide.md § Density & Row Heights`.*

norbert-kulus-blockydevs:
## Addressed self-review feedback

Fixed in `9413e4613` — the blocking finding plus all five important ones from [the tech-lead self-review](https://github.com/openlinker-project/openlinker/pull/3168#pullrequestreview-5174552581).

### BLOCKING — the dialog could never render a product name

`variant?.name` was never a product name — it's the variant's attribute-derived label (`350 ml`, or the SKU, or a raw `ol_variant_*` UUID at worst). Fixed at the source instead of patching the fallback chain: `GET /listings/price-changes/auto-applied` now returns `productName`/`variantLabel`/`sku`, enriched server-side in `PriceChangesService.listAutoApplied` from the exact same `getVariantsByIds`/`getProductsByIds` pair `listOpen` already uses for the review queue — mirroring `toQueueItem`'s enrichment rather than inventing a second path. A new `PriceChangeAutoAppliedView` application type carries it through the port → controller → DTO.

This also resolves the N+1 finding for free: `use-auto-applied-variant-summaries.ts` (the per-item variant fan-out) is deleted entirely — nothing needs it anymore.

### IMPORTANT fixes

- **The show/hide justification was wrong, and the log-only proxy could never disappear.** Visibility now gates on `anyConnectionAutomatic()` (mirroring the mockup's own function of that name) computed over the **already-cached** `useDestinationPricingSyncSummaries` read — the same hook `PricingRulesPickerDialog` fires in the same parent, so this costs zero extra requests — **and** on the log actually having entries. Gating on config alone is what lets the note go back to hidden once every connection is switched back to Manual; `price_change_auto_applied_log` has no retention sweep, so a log-only gate could never un-show itself once it had ever fired.
- **The dialog title still said "today"** even after the banner text was corrected to "recently" — the more prominent copy was still making the claim the rest of the PR argued was unsupportable. Changed to "Applied automatically recently" to match.
- **`<a href="#">`** (the only one in `apps/web/src`) replaced with a real `<button>`, styled via a new `.filter-note__action` class to still read as inline text. The banner also gets `role="status"` so it's announced to AT.
- **`.thumb` was unstyled inside `.mini-row`** — it only ever existed as `.cell-product .thumb`, so the avatar rendered as bare, unsized text and the inline `width`/`height` did nothing on an `inline` `<span>`. Added `.mini-row .thumb`, which also fixes the identical defect in `bulk-accept-price-changes-dialog.tsx`, and dropped the now-redundant inline styles from both call sites.
- **`AutoAppliedDialog`/`AutoAppliedRow`** are no longer re-exported from the feature barrel — both are used only within `features/price-changes`.

### Tests

- Backend: 3 new cases for `listAutoApplied`'s enrichment (happy path, empty log short-circuit, unresolvable-variant fallback).
- Frontend: 4 new cases for `anyConnectionAutomatic`, plus the note's visibility gating rewritten to cover all four automatic/manual × empty/non-empty combinations.
- Two of those visibility cases caught a real defect in the original self-authored test in the same commit that introduced it: `toBeEmptyDOMElement()` against `renderWithProviders`'s container can never pass, because `ToastProvider` always renders a notification-region node into it regardless of what the component under test does. Fixed to assert the banner text's absence instead.

Not addressed here (correctly out of scope per the review's own framing — these belong to the wider #3139 stack, not this diff): the `.table-wrap` cross-plugin style leak, the three drifted `initials()` copies, the duplicate `.activity-list` definition, and the `docs/architecture-overview.md` / style-guide gaps.

piotrswierzy (commented):

**Clearly the strongest PR in this batch, and the one I was most suspicious of.** On the question it was most likely to fail — does the "recently applied automatically" note make a positive claim from an absent read — it is **correct by design and documents why**:

`items = query.data ?? []` → `show = items.length > 0 && …`, so a failed `GET .../auto-applied` hides the banner entirely. It never renders `0`, never says "none". `anyConnectionAutomatic` handles `undefined` explicitly with *"degrades to hidden on an unresolved read rather than guessing."* It says **what** (product, variant, source→destination, old→new price), says **when** (relative time per row), and reads a **durable record** (`price_change_auto_applied_log` via `findRecent`) rather than an inference. The docblock even notes that log has no retention sweep, which is why the config gate exists.

And the gate is **stronger than the mockup's** — config **and** log presence, with the mockup's config-only version named as insufficient and the reason given (a log-only gate never returns to false once true). It also independently caught and corrected the mockup's own "today" overclaim, since `findRecent(limit)` has no calendar boundary. That is the right instinct.

### IMPORTANT — the count is `min(actual, 20)` presented as exact

`DEFAULT_AUTO_APPLIED_LIMIT = 20` (`price-changes.controller.ts:35`) and the note renders `{items.length}`. With 40 auto-applies the banner says *"20 price changes went live automatically recently"* — a wrong number, stated flatly. § Paginated Totals has the convention: **`20+`**, never a capped figure presented as a total. Either render `20+` when the page came back full, or have the endpoint report a total alongside the capped list.

This is the same instinct you already applied to the *temporal* version of the problem, applied to the cardinality.

### IMPORTANT — the "zero extra requests" claim is false, and the fan-out is unconditional

The prop docblock says *"Same set `PricingRulesPickerDialog` reads — sharing it means this note costs zero extra requests."* That holds only if the picker is also mounted, and the picker is a dialog that mounts on open. `AutoAppliedNote` mounts on **every** render of the Price changes tab, so opening that tab issues **1 + N** `GET .../pricing-sync` requests — N = destination connections — purely to decide whether to show a one-line banner. Add a `staleTime`, or rely on the log-presence gate alone, or have #3146 expose one list-shaped summary endpoint.

### IMPORTANT — operator-facing English in `libs/core`

`price-changes.service.ts` (`listAutoApplied`): `productName: product?.name ?? 'Unknown product'`. The rule is stated plainly for the analogous case (#2353): *"Every field on the wire is a **CODE**; operator copy is the frontend's and must pass `check-ui-vocabulary`, which a backend string could never enter."* Emit `null` and let the FE render `AbsentValue`.

It also asserts a name where there is none, and conflates "variant not found" with "product not found" — and because the FE reads `item.productName` verbatim, it can never tell them apart.

### SUGGESTIONS

- A failed read hides the banner, which is strictly safer than asserting — but an operator whose connection *is* Automatic and whose log read failed sees nothing at all, with no signal. The `usePaginatedTotal` doctrine keeps `unavailable` distinct from `idle` precisely so a failed read can say so. Consider a quiet inline failure note.
- **The PR description is stale relative to the code** — it says the dialog *"batches a per-item variant-summary lookup … falling back to the bare id"*, but the shipped code reads `item.productName` / `item.variantLabel` because this PR added those fields server-side. The as-built approach is better; the description describes the abandoned one.
- § UX Mockups: both component headers cite the mockup **and** document two deliberate deviations with reasoning — close to ideal. It just needs to be in the description too, since the mockup's `updateAutoAppliedNote()` and its "…today" copy are both now known-stale.

**Also done well:** the backend enrichment is properly batched (one `getVariantsByIds`, one `getProductsByIds`, two Maps, early return on empty) with a core `.spec.ts`; `anyConnectionAutomatic` extracted to a pure `lib/` module with an explicit `undefined` contract and its own test; `.filter-note__action` is a real `<button>` with a comment explaining why; `.mini-row .thumb` fixes a genuine bug an earlier stack PR left. `role="status"`, tokens only.

**No blocking issues of its own** — but it sits on #3166/#3167 and cannot merge ahead of their fixes. Note also that **no CI has run on this branch** (0 checks), so `auto-applied-note.test.tsx` and `pricing-sync-mode.test.ts` have never executed.

---

norbert-kulus-blockydevs:
## Rebase onto updated stack + rebase-surfaced fixes

Rebased this branch (skipping the duplicate ancestor commits that had accumulated relative to origin — same pattern as #3163/#3165/#3167) cleanly onto the just-updated `3150-pricing-sync-source-rollup-and-picker` (which now includes #3167's own review fixes). The branch is now exactly two commits ahead of its parent: `84820b1ee` (feat) and the self-review fix, plus a new third commit for issues the rebase itself surfaced.

**All 6 original review findings remain fixed** (verified fresh against `9413e4613` after integrating with the rest of the stack):
1. Real product identity (`productName`/`variantLabel`/`sku` from `listAutoApplied`, N+1 hook deleted) — intact.
2. Config-based visibility gate (`anyConnectionAutomatic`) — intact.
3. Dialog title "recently" (not "today") — intact.
4. No per-item variant fan-out — intact.
5. Real `<button>` + `role="status"` instead of `<a href="#">` — intact.
6. `.mini-row .thumb` styling — **needed a follow-up**, see below.

**Three things the rebase itself introduced**, fixed in a new commit `ac2524664`:

- **Duplicate `.mini-row .thumb` CSS rule.** #3167's now-merged history already carries a `.mini-row .thumb` rule from the accept/edit dialogs PR (#3148 review, finding 8) — which this PR's self-review fix didn't know about when it added its own rule for the same selector. Worse: the #3148 rule depends on inline `width`/`height`/`fontSize` at the call site, which this PR's fix had just stripped from `BulkAcceptPriceChangesDialog` (on the assumption its own new rule handled sizing). Folded into one rule that supplies sizing itself, so neither dialog needs an inline style.
- **`oldAmount` nullability.** The parent stack's #3159/#3162 fix widened `oldAmount` to `number | null` (a first-ever detection has no baseline) after this PR's fix commit was authored against the non-nullable shape. Widened `PriceChangeAutoAppliedView`/the response DTO to match, and taught `AutoAppliedDialog` to omit the struck-through old price when null (mirrors the queue table's existing `computedOldAmount === null ? '—' : ...` pattern). Caught by `pnpm --filter @openlinker/web type-check`.
- **Stale "zero extra requests" doc claim.** #3167's review fix gated `PricingRulesPickerDialog`'s own summaries query on `open`, so this PR's note no longer gets a free ride off an always-on sibling fetch — it's now the source of its own (still-batched) request. Corrected the docstring rather than leave an inaccurate claim in place.

**Verification**: `pnpm --filter @openlinker/web type-check`, `pnpm --filter @openlinker/web lint` (0 errors), `pnpm --filter @openlinker/core type-check`, `pnpm --filter @openlinker/api type-check`, `pnpm check:invariants` (all green), plus targeted `vitest`/`jest` runs on the touched suites (46 + 29 passed).

norbert-kulus-blockydevs:
Thanks for the review — addressed all three IMPORTANT findings and both suggestions.

**1. Capped count presented as exact.** Added `formatAutoAppliedCount` (`lib/auto-applied-count-label.ts`), mirroring the backend's `DEFAULT_AUTO_APPLIED_LIMIT = 20`. The banner now renders `20+` (never a bare capped number) once the read comes back exactly at the limit, with its own unit test and a component test asserting the capped case.

**2. Unconditional per-connection fan-out.** `useDestinationPricingSyncSummaries` (one request per destination connection) is now gated on the auto-applied log query having at least one item — since `show` is already `false` regardless of config when the log is empty, skipping the fan-out in that case changes nothing observable while saving N requests. With an empty log this note now costs exactly one request (down from 1 + N); with entries present it's unchanged. Rewrote the prop docblock to state this precisely, and rewrote the existing test that had asserted the *old* unconditional-fetch behavior to instead assert the config request is now skipped when the log is empty.

**3. `'Unknown product'` fallback string in `libs/core`.** `PriceChangeAutoAppliedView.productName` (and its DTO/FE mirror) is now `string | null` — `null` when the variant or its product couldn't be resolved, covering both cases you flagged (they're distinguishable via `variantLabel`/`sku` if a caller wants). The dialog renders the existing `AbsentValue` "Product not found" placeholder instead of a backend-authored sentence. Added core-service tests for both null cases (variant unresolved; variant resolved, product isn't) and a component test for the placeholder rendering.

**Suggestions:**
- PR description rewritten with an "As-built" section replacing the stale per-item-lookup / config-only-proxy description, matching what's actually shipped (server-side enrichment, log-AND-config gate, `20+` count, gated fan-out) — also folds in the mockup-deviation notes from both component headers so the description doesn't go stale again.
- Left the "failed log read hides the banner with no signal" trade-off as-is per your note that it's strictly safer than asserting — no change made there, flagging it here rather than silently dropping it in case a future issue wants to pick it up.

Verification: `pnpm --filter @openlinker/web lint/type-check` (0 errors), `pnpm --filter @openlinker/core lint/type-check` (0 errors), `pnpm --filter @openlinker/api type-check` (0 errors), `pnpm check:invariants` (exit 0), targeted `jest` on `libs/core/src/listings` price-changes specs (30 passed) and `vitest` on `apps/web/src/features/price-changes` (53 passed, up from 46).

Also confirmed the branch was rebased onto #3150's just-landed deep-link fix (f7dacef2e) — no interaction with this PR's files.

norbert-kulus-blockydevs:
## Independent re-verification of the #3168 review fixes, post-stack-merge

Merged the just-updated `3150-pricing-sync-source-rollup-and-picker` (which carries #3165/#3166/#3167's own fixes, including two real NaN-parsing bugs and a type error found during those rebases) into this branch — one conflict, in `price-changes.service.spec.ts`, where the parent added a `countOpen` describe block and this branch added a `listAutoApplied` one; kept both.

Independently re-checked all three IMPORTANT findings and both suggestions from the follow-up commit against the actual current code (not just the PR-comment prose):

1. **Capped count (`20+`).** `formatAutoAppliedCount` (`lib/auto-applied-count-label.ts`) correctly returns `{label: '20+', plural: true}` once `count >= AUTO_APPLIED_ITEM_LIMIT` (mirrors the backend's `DEFAULT_AUTO_APPLIED_LIMIT = 20` in `price-changes.controller.ts`) and never a bare `'20'` at the boundary. Checked the boundary cases explicitly: 19 → `'19'`, 20 → `'20+'`, 25 → `'20+'` (all three asserted in `auto-applied-count-label.test.ts`). No off-by-one.

2. **Gated fan-out.** `AutoAppliedNote` passes `useDestinationPricingSyncSummaries(connectionIds, { enabled: items.length > 0 })`, and the hook genuinely threads `enabled` into each per-connection `useQueries` entry (not just a wrapper-level short-circuit) — so with an empty auto-applied log the summaries fan-out issues zero requests, and the note costs exactly one request total. With entries present it's unchanged (1 + N).

3. **`productName: string | null`, no backend-authored English.** Confirmed end-to-end: `PriceChangeAutoAppliedView.productName` (core), `PriceChangeAutoAppliedResponseDto.productName` (API DTO), and the FE `PriceChangeAutoAppliedItem.productName` mirror are all `string | null`. `listAutoApplied` writes `product?.name ?? null` (not the `'Unknown product'` fallback the review flagged — that string still exists in the codebase but in an unrelated, pre-existing method, `toQueueItem`/`PriceChangeQueueItem`, which was never in scope here). `AutoAppliedDialog` renders `productName === null ? <AbsentValue label="Product not found" /> : productName`. Both null cases (variant unresolved; variant resolved but product not found) have core-service tests, distinguishable via `variantLabel`/`sku`.

Ran fresh after the merge (this environment, not trusting the prior commit's report): `pnpm --filter @openlinker/web type-check` (0 errors), `pnpm --filter @openlinker/web lint` (0 errors, pre-existing warnings only), `pnpm --filter @openlinker/core type-check` (0 errors), `pnpm --filter @openlinker/core lint` (0 errors), `pnpm --filter @openlinker/api type-check` (0 errors), `pnpm check:invariants` (exit 0, all ~45 checks green).

**Confirmed already resolved** — nothing further needed here. This is the last PR in the 12-PR stack; after this merge + push, every branch in the chain (#3157 → #3168) contains its own ancestors' fixes.

norbert-kulus-blockydevs:
Caught up `3151-auto-applied-note-and-dialog` with the latest `3150-pricing-sync-source-rollup-and-picker` (merge commit `60673766c`), bringing in the upstream CI-fix commits: the missing `PriceChangeDetectionService` barrel export and the corrected lane-count test assertion.

Clean merge, no conflicts. Verified: `@openlinker/web` lint (0 errors) and type-check, `@openlinker/core`/`@openlinker/worker`/`@openlinker/api` type-check (after `libs` build), and `pnpm check:invariants` — all green.

This is also the last PR in the 12-PR ADR-072 stack, so the entire chain is now internally consistent end to end: every branch contains every ancestor's fixes, including both CI-discovered bugs.

piotrswierzy (commented):
## Tech Lead review

Re-review at `6067376`. **All three IMPORTANT findings from the previous round are properly closed**, and closed at the right layer each time. The capped count is now `formatAutoAppliedCount` returning `20+` at the limit (boundary asserted at 19/20/25), which is the § Paginated Totals convention applied to cardinality rather than a local patch. The per-connection fan-out is genuinely gated — `useDestinationPricingSyncSummaries(ids, { enabled: items.length > 0 })` threads `enabled` into each `useQueries` entry, so an empty log now costs exactly one request, and the prop docblock was rewritten to state that precisely instead of repeating the now-false "zero extra requests" claim. And `productName` is `string | null` end to end (core view type → DTO → FE mirror), with `listAutoApplied` writing `product?.name ?? null` and the dialog rendering `AbsentValue` — no backend-authored English, with a comment citing the #2353 rule. The backend enrichment is correctly batched (one `getVariantsByIds`, one `getProductsByIds`, two Maps, early return on empty) and both null cases have core-service tests.

The properties this component most needed to get right are still right. `items = query.data ?? []` → `show = items.length > 0 && anyConnectionAutomatic(...)` means a failed log read **hides** the banner rather than rendering `0` or "none" — absence and "nothing happened" never collapse, which is the rule `docs/frontend-architecture.md` § Paginated Totals states for the analogous case. `anyConnectionAutomatic` handles `undefined` explicitly and degrades to hidden, is extracted to a pure `lib/` module with its own test, and gating on config **and** log presence is stronger than the mockup's config-only version for the stated reason (the log has no retention sweep, so a log-only gate could never un-show itself). The "recently"/"today" correction is carried through both the banner and the dialog title now. State ownership is clean throughout: server state in TanStack Query, one `useState` for dialog open, no store, no raw `fetch`, Radix `shared/ui/dialog` rather than a hand-rolled `role="dialog"`, tokens only, and a real `<button>` with `role="status"` on the banner.

### Findings

**[IMPORTANT]** — `apps/web/src/features/price-changes/lib/auto-applied-count-label.ts:24`
> `AUTO_APPLIED_ITEM_LIMIT = 20` is an unguarded frontend mirror of the backend's `DEFAULT_AUTO_APPLIED_LIMIT` (`price-changes.controller.ts:35`). The docblock says so itself — *"MUST match … this is a deliberate mirror, not a coincidence"* — which is precisely the condition this repo answers with a `scripts/check-*-mirror.mjs` under `pnpm check:invariants`; there are **23** such guards today (`check-return-stage-mirror`, `check-system-connection-id-mirror`, `check-shipping-tax-split-mirror`, `check-hold-reason-mirror`, …), several of them protecting exactly this shape: a single constant the browser cannot import because `apps/web` has no `@openlinker/*` dependency (#591). Without one, raising the backend limit to 50 leaves the banner capping at `20+` for a true count of 20 while 30 more exist — a wrong number, stated flatly, which is the same defect the `20+` fix was made to remove. A ~20-line guard grepping the two literals closes it; a comment cannot, because the failure is silent.

**[IMPORTANT]** — CI: Integration Tests red on this branch
> `listings-price-change-episode.int-spec.ts` fails four cases with `invalid input syntax for type uuid: "no-such-id"` / `"no-such-source"` — fixtures passing non-UUID literals into uuid columns, from the `#3162 review` / `#3163 review` describe blocks. **Not this diff** (that file is untouched here) — it is inherited from the parent stack branch this PR is based on. Flagging it because this PR is the stack tip and cannot merge red, so it needs fixing upstream before this one lands; noting the attribution so it is not chased in the wrong diff.

**[SUGGESTION]** — `apps/web/src/features/price-changes/components/auto-applied-dialog.tsx:41-50`
> This adds a fourth copy of `initials()` to the app and a **second inside this one feature folder**. Your own earlier self-review flagged the drift: only this copy carries `.filter(Boolean)`, so `initials('Acme  Store')` yields `"AS"` here and `"A"` in the sibling `bulk-accept-price-changes-dialog.tsx`, two files apart in the same directory, feeding the same `.mini-row .thumb`. The two app-level copies (`app-shell.tsx`, `mapping-pairing-bar.tsx`) are pre-existing and out of scope, but the two in `features/price-changes` are both on this stack and collapsing them into one `lib/initials.ts` is a one-line move that makes the correct implementation the only one.

**[SUGGESTION]** — `apps/web/src/features/price-changes/index.ts:11`
> `AutoAppliedNote` is added to the feature barrel, but its only consumer is `price-changes-queue-table.tsx` **inside the same feature**, which reaches it by relative import. `AutoAppliedDialog` was correctly kept off the barrel with exactly this reasoning quoted in its header (*"start narrow, widen later if a real external consumer appears"*); the note qualifies on the same test. Separately — and this is a stack-level item rather than yours — `features/price-changes` appears **zero** times in `.eslintrc.js`, so it is in neither `no-restricted-imports` group and the barrel it is growing is unenforced. § Feature Public Surface makes registering the slug step 2 of adding a public surface; worth doing once for the stack.

**[SUGGESTION]** — `auto-applied-dialog.tsx:71`
> `style={{ color: 'var(--text-muted)', fontWeight: 400 }}` is a new inline style two lines below a change whose whole point was moving `.thumb`'s sizing off a caller's inline style and into the shared rule. Tokens are used so this is valid per § Design tokens, but it is inconsistent within one file; a `.mini-row__variant` class would match what the same commit just did for `.thumb`.

### Verdict

🔄 **Approve with changes**

No blocking issues of its own, and the component is well-judged — the mockup divergences are documented at the point of divergence with reasons, and the "recently" correction and the config-AND-log gate are both improvements on the mockup rather than deviations from it. Posting as a comment rather than an approval because of the two IMPORTANT items, one of which is upstream.

1. Add a `check-auto-applied-limit-mirror.mjs`-style guard under `pnpm check:invariants` for `AUTO_APPLIED_ITEM_LIMIT` ↔ `DEFAULT_AUTO_APPLIED_LIMIT`.
2. Get the parent stack's `listings-price-change-episode.int-spec.ts` UUID fixtures fixed so this branch can go green.
3. (Optional) One `initials()` in this feature; drop the unconsumed barrel export.

---

norbert-kulus-blockydevs:
All three addressed — `2d5b05f5b`, on the parent catch-up `4df6829b9`.

**1. The unguarded mirror.** `scripts/check-auto-applied-limit-mirror.mjs` now holds `AUTO_APPLIED_ITEM_LIMIT` and `DEFAULT_AUTO_APPLIED_LIMIT` together, chained into `pnpm check:invariants` beside the sibling you pointed at. I verified it actually fails rather than trusting it: bumping the frontend constant to 25 exits 1 with both values named, and the tree is green again once restored.

Two deliberate departures from the sibling's shape, both stated in the script's header. Its self-check runs unconditionally at the top rather than behind a `--self-check` flag, so the chain cannot be wired up in a way that skips it — one call instead of two. And the header says what the guard does **not** cover: it compares the two literals and nothing else, so a refactor that keeps both numbers and stops passing one of them to `listAutoApplied` still passes. A guard whose limits are not written down is a guard that gets over-trusted.

The failure message explains the consequence rather than just the mismatch, because the direction matters and is not obvious: raise the backend alone and the note silently stops capping, rendering a page size as an exact count; lower it alone and the note claims a backlog the page can never contain.

**2. CI.** The inherited failures were #3162's uuid fixtures and #3165's parse/session/clamp cases, all fixed at their source and carried in through this branch's merges. Locally on this head: `type-check` clean, `check:invariants` clean, 69 tests passing across `features/price-changes`.

**3. The optional items.** Both taken. The dialog's local `initials` folds into the feature's `initialsFor`, leaving a three-line wrapper for the null case — I kept that separate rather than widening the shared helper's signature, since "render `?` for an unnamed row" is a display decision and not part of the naming rule. `AutoAppliedNote` comes off the feature barrel; its only caller is a sibling component in the same feature importing it relatively, so the export had no consumer. And the variant-suffix inline style becomes `.mini-row__variant` in the block the dialog already ships.

One thing I did while merging the parent down, worth flagging since it touches this PR's own CSS: `.filter-note__action` — which this PR adds — is now `.price-changes-queue .filter-note__action`, because #3164's round-3 fix scoped that whole block to the component root after `.table-wrap` was found restyling the Subiekt invoice dialog. `AutoAppliedNote` renders inside that root, so the rule still applies; it is no longer a global.

piotrswierzy (approved):
## Tech Lead re-review (`6067376` → `a79aeb4b`)

**Both IMPORTANT findings are closed, and both SUGGESTIONs were taken.** `auto-applied-note.tsx` and `auto-applied-count-label.ts` are byte-identical to the commit I reviewed last round — the delta is entirely the guard, the consolidation and upstream stack merges, so none of the properties I verified before needed re-verifying.

**The mirror guard is real, and it is a good one.** `scripts/check-auto-applied-limit-mirror.mjs` is wired into `check:invariants` (and therefore into `pnpm lint` and the pre-commit hook), textually reads both `DEFAULT_AUTO_APPLIED_LIMIT` and `AUTO_APPLIED_ITEM_LIMIT`, and fails on drift **or on either side going missing** — the second half matters as much as the first, because a rename that silently stops the comparison is how a mirror guard becomes a check that cannot fail. Three details put it above the bar: the self-check runs unconditionally at module scope rather than behind a `--self-check` flag, so it can never be invoked in a mode where the reader is untested; its case table includes the negative cases that matter (`const AB` must not match `A` by prefix; a non-literal RHS returns `null` rather than a bogus number); and the docblock declares its own scope — *"this compares the two literals and nothing else … a refactor that keeps both numbers and stops using one of them passes"* — which is exactly the honesty `docs/architecture-overview.md` asks of these guards and which several older ones lack. The failure message explains the operator-visible consequence in both directions rather than just reporting inequality.

**The upstream blocker is gone.** `test(api): spell the price-change no-match fixtures as well-formed uuids` landed on the parent stack, so the four `invalid input syntax for type uuid` failures in `listings-price-change-episode.int-spec.ts` that were reddening this branch are fixed where they belonged rather than worked around here.

Both SUGGESTIONs are addressed too: there is now one `initialsFor` in the feature (`lib/price-change-copy.ts`), imported by both `auto-applied-dialog.tsx` and `bulk-accept-price-changes-dialog.tsx` — and the `auto-applied-dialog` wrapper `initialsForOrUnknown` is the right shape, since a `null` `productName` needs a placeholder the shared helper has no business inventing. `AutoAppliedNote` is off the barrel (its only consumer is same-feature and reaches it relatively), and the inline `style={{ color: …, fontWeight: 400 }}` is gone. I also re-checked the stack-level item I raised: `features/price-changes` is now registered in `.eslintrc.js`, so the barrel this feature is growing is enforced.

### Findings

### SUGGESTIONS

- `check-auto-applied-limit-mirror.mjs` is invoked once in `check:invariants` while all ~30 siblings are invoked twice (`--self-check` then the real run). Running the self-check inline is *better* — it cannot be skipped — but it is the only script in the list shaped that way, so a future maintainer normalising the list may "fix" it by adding a `--self-check` invocation that does nothing distinct. One line in the docblock saying the self-check is unconditional by design would settle it.
- The self-check's `['// const A = 20;', 'A', 20]` case documents that a commented-out declaration still matches. Declaring the limitation is right; it is worth noting that the failure mode it admits is benign here (a commented-out constant would break the build long before this guard ran) — so this is a note, not a gap.

### CI

`Type Check`, `Docker Build Smoke Test`, `PHP Unit Tests`, `Scaffolded Adapter Builds` and `Detect changed paths` pass. `Test`, `Lint`, `Build` and `Integration Tests` are **still queued** behind the rest of the stack's runs — none red, none reporting. `mergeable: MERGEABLE`; `mergeStateStatus: BLOCKED` on those pending checks and on the base (`3150-pricing-sync-source-rollup-and-picker`, #3167) not yet being merged, which is the normal state for a stack tip.

### Verdict

✅ **Approve**

Approving on the code, with the standing stack condition: this is the tip of a ten-PR stack, so it merges after its base and with its own checks green. Nothing here needs another round.

---

norbert-kulus-blockydevs:
Both taken — `fd605203f`, in the script's own header where the next reader meets them.

The single-invocation departure is recorded as deliberate and stronger (the self-check cannot be omitted when the chain is edited), with the explicit note that a reader comparing the `package.json` line against its ~30 neighbours is not looking at a slip.

On the commented-out-declaration case: I wrote down *why* it is benign here specifically, rather than only that it is. A regex reader meeting a commented-out constant either still matches the live one or fails loudly with "not found" — it cannot silently pass on a stale number. The header says that reasoning does not transfer to a guard that tolerates a missing side, so the admission is not read as a general licence.

piotrswierzy (approved):
## Tech Lead re-review (`1457d820` → `611b06b7`)

Thirteen commits, **seven of them merges** carrying the whole `3143…3150` stack forward, plus #3166's and #3167's doc commits arriving through the base. The PR's own new work is **one comment-only commit**, `fd605203`, on the mirror guard. Net file diff base…head: `+16/-0` `.tsx`, `+8/-4` core service, `+17/-0` `check-auto-applied-limit-mirror.mjs` — **zero executable lines anywhere**. **The approval stands.**

**Mirror-guard status, re-confirmed at `611b06b7`:**

- `scripts/check-auto-applied-limit-mirror.mjs` **exists** (124 lines) and is **still wired** into `check:invariants` in the root `package.json`, which `pnpm lint` chains.
- **The two constants still agree**: `DEFAULT_AUTO_APPLIED_LIMIT = 20` (`apps/api/src/listings/http/price-changes.controller.ts:74`, consumed at :110) and `AUTO_APPLIED_ITEM_LIMIT = 20` (`apps/web/src/features/price-changes/lib/auto-applied-count-label.ts:24`, consumed at :34-35). Both paths are exactly the ones the guard reads.
- The unconditional module-scope self-check, the fail-on-drift and fail-on-missing-constant cases and the explicit SCOPE disclaimer are all intact.

`fd605203` is the right answer to the two observations I raised. It records that the single `package.json` invocation is a **deliberate divergence** from the ~30 siblings' `--self-check &&` double-invocation, and argues correctly that inlining is *stronger* — the chain cannot be edited in a way that skips it. And it bounds the `'// const A = 20;'` admission honestly: a regex reader either still compares the live values or fails loudly with "not found", so it cannot silently pass on a stale number, **and it says explicitly that the same tolerance would not be benign in a guard that tolerates a missing side**. That last clause is what stops the note becoming a licence the next author over-applies.

### Findings

### Verdict
✅ **Approve**

CI at `611b06b7`: `MERGEABLE` / `UNSTABLE`. Docker Build Smoke, Type Check, PHP Unit, Scaffolded Adapter Builds and path detection **pass**; Build, Lint, Test and Integration Tests **pending (queued, not run)** — no assertion failure. Top of the stack, targeting `3150-…`, so it merges last.

---

piotrswierzy (approved):
## Tech Lead re-verification (head unchanged at `69d9430`)

Head unchanged, `main` unmoved. CI on run 34845743517: **Docker Build Smoke Test, Scaffolded Adapter Builds, PHP Unit Tests and Detect changed paths pass; Build, Lint, Test, Type Check and Integration Tests are queued behind the runner backlog. Nothing red on any run.**

**Re-confirmed both halves of the mirror contract at this head.** The guard exists (`scripts/check-auto-applied-limit-mirror.mjs`, 5736 bytes) and is wired into `check:invariants` — which `"lint"` chains — so it runs on every `pnpm lint` including pre-commit. It names the two files explicitly and reads each constant as an integer literal, failing distinctly when either is renamed, removed or made non-literal as well as when the values disagree. And the two constants still **agree at 20**: `DEFAULT_AUTO_APPLIED_LIMIT = 20` (`apps/api/src/listings/http/price-changes.controller.ts:74`, consumed at :110 by `listAutoApplied`) and `AUTO_APPLIED_ITEM_LIMIT = 20` (`apps/web/src/features/price-changes/lib/auto-applied-count-label.ts:24`, driving the `20+` overflow label at :34-35).

Stack: base `3150-pricing-sync-source-rollup-and-picker` (#3167) — this is the tip of the six. **#3159 and #3162 at the bottom still carry an open BLOCKING**: each edits the already-merged migration `apps/api/src/migrations/1880000000000-create-price-change-episodes.ts` in place, so an already-migrated database silently diverges from a fresh install. None of #3163–#3168 carries a migration, so none is directly affected — but **nothing in this chain can merge until #3159 and #3162 are fixed**, which is why an approved PR is not moving. `mergeable: MERGEABLE`; `mergeStateStatus: BLOCKED` reflects the unmerged base, not a defect here.

### Findings

### Verdict
✅ **Approve**

---

Issue #2374, opened 2026-08-22
[TASK] CORE — credit-note correction proposal with positional ambiguity


piotrswierzy:
> Part of the OMS programme backlog (**Wave 2**, stream S2, size L). Backlog slug: `W2-38`. Sibling `W2-*` slugs resolve via the Wave-2 epic's checklist. Design of record: design §7.3, story T7; returns product spec §5.6 (PR #2318).

> **Size justification (L):** the proposal builder must diff disposed return lines against the invoice's `issuedLineSnapshot` under a line identity that is **positional and not stable**, classify each line `matched | ambiguous | no-match` with candidates, and hand off to the existing `CorrectionIssuer` — and a correction wrong by a price delta on a KSeF-transmitted document is not retractable, so the classification logic carries an unusually heavy test matrix.

## Problem / Context

Story T7. §7.3: a disposed line **proposes** a correction through the existing `CorrectionIssuer`/`issuedLineSnapshot` seam (ADR-041 excludes corrections from the one-originating-document guard — a linked follow-up, not a second document), pre-filled with the best positional guess and **showing the ambiguity**. Auto-issue stays gated on a stable line reference on `InvoiceLine`.

## Proposed Solution

A proposal builder that, for each disposed return line, finds candidate invoice lines in the `issuedLineSnapshot` (#1297) and classifies:

- **matched** — exactly one candidate, prefilled;
- **ambiguous** — several candidates, **all listed**, operator picks;
- **no-match** — excluded from the proposal with its reason stated.

The proposal is data; **nothing is issued** by this service. Handing off to `CorrectionIssuer` is the operator's confirmed act (`W2-44`). The proposal persists as a second `kind` on the ADR-044 `order_changes` table built in #2333 (`W1c-6`) — never a second proposal mechanism.

## Classification

**Type**: CORE
**Layer**: Application
**File(s)**: `libs/core/src/returns/application/services/**`; the invoicing `issuedLineSnapshot` read

## Dependencies

- #2370 (`W2-33`); **#2333** (`W1c-6` — the ADR-044 `order_changes` table this reuses)

## Assumptions

- `InvoiceLine` still carries no stable identifier; when it gains one, auto-issue becomes discussable — and only then.

## Acceptance Criteria

- [ ] A proposal never issues anything (asserted)
- [ ] An ambiguous line lists **every** candidate; a no-match line states its reason
- [ ] A correction-of-a-correction diffs against the prior correction's own lines (#1297's rule)
[...]

Issue #2327, opened 2026-08-22
[TASK] CORE — returns context: ReturnRecord/ReturnLine model and migration


piotrswierzy:
> Part of the OMS programme backlog (**Wave 1c**, stream S2, size M). Backlog slug: `W1c-1`. Sibling `W*` slugs resolve via the Wave-1c epic's checklist. Design of record: `docs/plans/analysis/DESIGN-oms-authority-model.md` + ADRs 052–062 (esp. ADR-060); returns product spec `docs/specs/product-spec-oms-returns-operator-ux.md` (PR #2318).

## Problem / Context

ADR-060 creates a new `returns` context — folding returns into the 8-outbound-edge `orders` context would make that context worse. `libs/core/src/returns/` does not exist.

## Proposed Solution

New `libs/core/src/returns/` with the full hexagonal shape (the `sales-documents` post-#2170 shape is the closest in-tree reference: `domain/entities`, `domain/types`, `domain/ports`, `application/interfaces`, `application/services`, `infrastructure/persistence/{entities,repositories}`, `index.ts`, `returns.module.ts`, `returns.tokens.ts`).

`ReturnRecord` (header): `id` (`ol_return_*`), `sourceConnectionId`, `externalReturnId`, `internalOrderId` **nullable**, `origin: 'source_ingested' | 'operator_authored'`, verbatim `rawStatus` + `rawPayload`, and **independent nullable timestamps** `authorizedAt` / `declinedAt` / `closedAt` (not a status column — authorization is an action, not a state).

`ReturnLine`: `resolvedOrderLineId` **nullable by design**, `reason` reusing `RefundReason` **verbatim** (so returns-by-reason and refunds-by-reason report on one axis), and **quantity counters, not statuses**: `quantityAdvised ≥ quantityReceived ≥ quantityRestocked + quantityScrapped`, with a DB `CHECK` expressing that ordering. The two per-line machines (`ReturnCustodyState`, `ReturnMoneyState` including `in_doubt`) are **declared as types now** and their columns default to `advised` / `not_refundable` — they are not driven until Wave 2. `ReturnCustodyStateValues = ['advised', 'in_transit', 'received', 'disposed', 'not_returned']` — **no `inspected`** (adjudicated; see Assumptions).

`RefundRecord` gains a **nullable `returnId`** — linked, not extended (refunds exist without returns and vice versa; widening `RefundRecord` would falsify live analytics rows).

[...]

PR #3172, opened 2026-09-11
feat(returns): resolve a return line to its order line at ingestion

### LINKED ISSUE #3171: [TASK] CORE — resolve ReturnLine.resolvedOrderLineId at ingestion instead of matching invoice lines by name
## Problem / Context

`ReturnLine.resolvedOrderLineId` exists as the designed home for the join between a returned line and the order line it came from, and **nothing in the tree ever populates it**. It is left `null` on the operator-authored path (`libs/core/src/returns/application/services/returns.service.ts:444`) and is deliberately excluded from both halves of the ingestion upsert, which is correct — `libs/core/src/returns/infrastructure/persistence/repositories/return.repository.ts:371` documents it as *"core-resolved attribution, never adapter-supplied"*. Core simply never does the resolving.

The contract already states whose job this is. `IncomingReturnLine.offerId` (`libs/core/src/returns/domain/types/incoming-return.types.ts:29-34`):

> Source-native offer/listing identifier this line refers to, best-effort. **This is the only line-level linkage a source may supply; resolving it to an OL order line is core's job, not the adapter's.**

Because that job is undone, the credit-note correction proposal (#2374 / #2376) falls back to matching disposed return lines against the issued invoice **by product name** — `normalizeCorrectionLineName` in `libs/core/src/returns/domain/domain-services/return-correction-matching.domain-service.ts` — which is the one axis guaranteed to collide. When an order bills the same product on two lines, the matcher emits an `ambiguous` line listing both candidates and selects neither, and the operator is asked to pick.

**That question is unanswerable, and it is manufactured.** Unanswerable: the operator holds a parcel, both candidate lines carry the same product name, and `InvoiceLine` is `{name, quantity, unitPriceGross, taxRate, unit?}` with no id and no SKU — nothing on the invoice distinguishes them, and the buyer never chose a line either. Manufactured: OpenLinker builds that invoice itself from the order's own items, in order — `libs/core/src/invoicing/application/mappers/order-to-issue-invoice-command.mapper.ts:113` is `order.items.map((item) => toInvoiceLine(item, order.id))`, with shipping lines appended at `:120` — so invoice line *N* is `order.items[N-1]` deterministically. Meanwhile the return line arrives carrying `offerId`, `sku` and `unitPrice`, and `unitPrice` is even written into `rawPayload` at `returns.service.ts:538` and never read again.

`FA(3)` makes the corrected line's position (`NrWierszaFa`) mandatory in every permitted correction method, so a guessed pick is an unauditable entry in a document filed with the tax office. Asking the operator does not add judgement here; it adds deniability.

## Proposed Solution

Resolve the join in core, at ingestion, and persist it monotonically.

**1. A pure resolution rule.** Add a domain service beside the existing matcher — `libs/core/src/returns/domain/domain-services/return-order-line-resolution.domain-service.ts` — taking the return line's `offerId` / `sku` / `unitPrice` / `quantity` plus the order's `OrderItem[]`, and returning either a single `OrderItem.id` or an explicit unresolved outcome with a reason. Pure, no I/O, mirroring `return-correction-matching.domain-service.ts`.

Resolution order, strongest axis first:
1. **`offerId` → variant → `OrderItem.variantId`.** `OrderItem` (`libs/core/src/orders/domain/types/order.types.ts:236-242`) carries `id`, `productId`, `variantId?`, `quantity`, `price`, `sku?` — but **no `offerId`**, so this step needs a seam that maps a source offer id to an OL variant. Name the seam in the PR: the offer-mapping read lives in `listings`, and `ReturnsModule` does not import it today. If that edge is judged too heavy for this slice, drop this step and say so — steps 2–3 already resolve the shapes that matter.
2. **`sku` → `OrderItem.sku`** (exact, after trim; both fields are optional).
3. **`unitPrice` → `OrderItem.price`** as the tie-breaker among several candidates from step 2, which is exactly the case the picker exists for.

More than one survivor after all three is **unresolved**, not a guess.

**2. Read the order's items.** `ReturnsService` has `internalOrderId` after attribution but no order read. `IOrderRecordService` (`ORDER_RECORD_SERVICE_TOKEN`) is the sanctioned cross-context contract; `ReturnsModule` importing `OrdersModule` is acyclic — `libs/core/src/orders/orders.module.ts` does not import `ReturnsModule`, and the standing invariant is only that `orders` must never import `returns` back. Record the new edge in `docs/architecture-overview.md` § Cross-context dependencies in core, as the other returns edges are.

**3. Persist without letting re-ingestion clobber it.** `upsertFromSource` must keep excluding the column. Add a narrow conditional write on `ReturnRepositoryPort` — `claimOrderLineResolution`, `WHERE "resolvedOrderLineId" IS NULL`, `affected > 0` as the answer — following the `claimAttribution` idiom already in this repository (`return.repository.ts`). The value is then fill-in-when-NULL and monotonic, for the same reason `internalOrderId` is: a failed re-resolve must never un-resolve a line.

**4. Report the unresolved state; never guess downstream.** An unresolved line stays `null` and is counted/logged at ingestion. Consuming it — narrowing the correction matcher onto `resolvedOrderLineId` and retiring the `ambiguous` picker — is deliberately **not** in this issue: this one makes the field trustworthy, a follow-up makes the matcher read it.

## Classification

**Type**: CORE
**Layer**: Domain (resolution rule) + Application (`ReturnsService`) + Infrastructure (the conditional write)
**File(s)**:
- `libs/core/src/returns/application/services/returns.service.ts`
- `libs/core/src/returns/domain/domain-services/return-order-line-resolution.domain-service.ts` (new)
- `libs/core/src/returns/domain/ports/return-repository.port.ts`
- `libs/core/src/returns/infrastructure/persistence/repositories/return.repository.ts`
- `libs/core/src/returns/returns.module.ts`
- `docs/architecture-overview.md` (§ Returns, § Cross-context dependencies in core)

## Dependencies

- None. Everything this reads is already shipped: `IncomingReturnLine` carries the keys, `OrderItem` carries the targets, `IOrderRecordService` is exported, and the column and its index (`IDX_return_lines_resolved_order_line`) already exist — no migration.

## Assumptions

- The `offerId` step may be dropped if the `returns → listings` edge is judged too heavy for this slice; `sku` + `unitPrice` cover the duplicate-line case the picker was built for. Whichever is chosen must be stated in the PR.
- Resolution runs for `source_ingested` returns at `upsertFromSource`. The operator-authored path (`recordReturn`) also holds a known `internalOrderId` and can reuse the same rule; if it is deferred, say so rather than leaving it silently `null`.
- No backfill of existing rows. Re-resolution of history is a separate, bounded pass if it is ever wanted.

## Acceptance Criteria

- [ ] A pure domain service resolves a return line to at most one `OrderItem.id` from the order's items, with an explicit unresolved outcome carrying a reason; it performs no I/O and mutates no argument
- [ ] Two order lines billing the same product, differing only in unit price, resolve to the correct one when the return reports `unitPrice`
- [ ] Two order lines that are identical on every available axis resolve to **unresolved** — never to an arbitrary pick
- [ ] `ReturnsService.upsertFromSource` persists the resolution for a `source_ingested` return, and a re-ingestion of the same return does not change or clear an already-resolved value
- [ ] The persisted write is a conditional `WHERE "resolvedOrderLineId" IS NULL` claim, proved against real Postgres by an int-spec with two overlapping attempts
- [ ] An unresolved line leaves the column `null` and is reported (count + log) rather than defaulted
- [ ] `upsertFromSource`'s write set still excludes `resolvedOrderLineId` in both halves — asserted, so a future edit cannot let the source echo it
- [ ] The new `returns → orders` module edge boots (`apps/worker` integration boot spec) and is recorded in `docs/architecture-overview.md`
- [ ] No architecture boundary violations (CORE ↔ Integration); `pnpm lint` (incl. `check:invariants`) and `pnpm type-check` pass

## Design context

Reviewed as part of epic #3087 (return credit-note correction UI). Two independent reviews — an OMS domain review and a competitor review — concluded that the `ambiguous` picker should not exist, and the reviewed mockup (#3116) deletes it: the screen now renders a before/after grid over the invoice's own lines, which is the shape every Polish invoicing tool uses, and BaseLinker advertises the absence of line-picking as a feature. **This issue is the prerequisite for that design** — without the join, OpenLinker cannot pre-fill a single quantity.

norbert-kulus-blockydevs:
Closes #3171

## Problem

`ReturnLine.resolvedOrderLineId` has existed since #2327 as the designed home for the return-line → order-line join, and **nothing in the tree populated it** — it was hardcoded `null` at ingestion.

So the credit-note correction path (#2374) fell back to matching disposed return lines against the issued invoice **by product name**, the one axis guaranteed to collide, and then escalated its own manufactured collision to the operator as an `ambiguous` line with a two-option picker.

That question is unanswerable: the operator is holding a parcel, `InvoiceLine` carries no id and no SKU, and both candidates carry the same name. It is also unnecessary — OpenLinker builds the invoice itself from `order.items` in order (`order-to-issue-invoice-command.mapper.ts:113`), so invoice line *N* is `order.items[N-1]` deterministically, and the return line already arrives carrying `sku` and `unitPrice` (the latter stashed into `rawPayload` and never read).

## What this does

Resolves the join at ingestion, so the field becomes trustworthy. **No existing behaviour changes** — nothing reads the column yet.

- **`resolveReturnLineOrderLine`** (new pure domain service) narrows on SKU, then breaks a same-SKU tie on the reported unit price.
  - **It refuses rather than guesses.** `NrWierszaFa` is mandatory in every permitted FA(3) correction method, so a guessed line is an unauditable entry in a document filed with the tax office. More than one survivor is `ambiguous`, and the matcher's unresolved path stays reachable.
  - **The axis order is a NARROWING, never a retry** — a SKU that matched nothing does not fall through to price, which would resolve a line the SKU positively *excluded*: a wrong answer rather than a missing one.
  - SKU is trimmed but **not** case- or diacritic-folded (it is an operator-authored identifier, not display text, unlike `normalizeCorrectionLineName`). Product **name** is deliberately not an axis — its collisions are what the rule removes.
- **`claimOrderLineResolution`** is a fill-in-when-NULL claim (the `claimAttribution` / `claimWaybillRelay` idiom), so a later or failed re-resolve can never *un*-resolve a line. The column stays absent from **both** halves of `upsertFromSource` — it is core-resolved and must never be echoed from a source payload, which a structural spec pins.
- **`OrderRecord.orderItems`** is the orders context's own fail-safe narrowing of `orderSnapshot.items` (the `codToCollect` / `nativeTotals` precedent), so no sibling re-parses that jsonb layout. A partial line is dropped rather than coerced.
- **The order read is composed in `apps/worker`** (`ReturnOrderLineResolverService`) and passed in as an argument — ADR-053's "order data enters as arguments". `returns` therefore takes **no new module edge**: `OrdersModule` imports fourteen modules, nine of them new to returns, for one `getOrderRecord` call, and the repo has preferred the worker for this exact shape twice (`OrderHoldsModule`'s leaf rationale; #2712 reporting from core and writing from the worker). **Zero** new entries in `check-cross-context-imports.mjs` or `barrel-purity.spec.ts` allow-sets.
- The pass is **best-effort** (the return is already durably persisted when it runs) and **idempotent**, which is also the only route by which a return ingested *before* this ships ever gets its lines resolved.

## Limitations, stated rather than implied

- **No `offerId` axis.** `OrderItem` carries no `offerId`, so that axis needs an offer→variant read in `listings` which `returns` does not have. Deferred and documented.
- **`marketplace.returns.statusSync` does not resolve.** `ReturnStatusSyncResult` reports counters only and carries no list of the returns it touched, so the lifecycle re-read has no id to pass. The per-return sync path does resolve.
- **Consuming the field is a follow-up** — narrowing `return-correction-matching.domain-service.ts` onto `resolvedOrderLineId` and retiring the picker. This slice makes the field trustworthy first.

## Testing

- New: `return-order-line-resolution.domain-service.spec.ts` (table-driven over every axis and every refusal, plus a purity assertion that scans its own source), `order-line-resolution-is-core-owned.spec.ts` (3 structural guarantees), an `OrderRecord.orderItems` describe block, and two new cases in `returns-handlers.spec.ts` (resolution called on success; **not** called on a terminal ingestion failure).
- `pnpm lint` ✅ and `pnpm type-check` ✅ both clean across the monorepo.
- **`pnpm test` and `pnpm test:integration` were NOT run locally** — this machine cannot carry them; they are left to CI.

## Docs

- `docs/plans/implementation-plan-return-order-line-resolution.md` (new)
- `docs/architecture-overview.md` § 22 Returns — one new bullet recording the four load-bearing properties and both limitations.


piotrswierzy (changes requested):
## `/pr-review` — 🔴 Request changes

The design here is right, and the header says why better than a reviewer could: the credit-note path matched disposed return lines against the issued invoice **by product name** — the one axis guaranteed to collide — then escalated the collision to an operator holding a parcel, with an `InvoiceLine` that carries neither id nor SKU and two candidates carrying the same name. Resolving at ingestion removes that question instead of delegating it. Four choices are load-bearing and all four are made correctly: the axis order **narrows** rather than retries (a SKU that matched nothing returns `no-candidate` and never falls through to price, so the rule can produce a missing answer but not a wrong one); the write is a fill-in-when-NULL claim, which is what makes the resolver safe to call from both the per-return sync and the lifecycle re-read and is the only route by which a pre-existing return ever gets resolved; `readReportedUnitPrice` matches on the stored entry's own `lineIndex` and **not its array position**, which is necessary because `buildRawPayload` drops every extras-less line and leaves the array sparse; and `OrderRecord.orderItems` centralises the snapshot narrowing in the owning context, so `returns` binds to a typed contract rather than a jsonb layout and takes no `OrdersModule` edge. Placing the order read in the worker matches both precedents the header cites.

### BLOCKING — on the only source that implements returns, the SKU axis can never fire, so every line lands on the price-alone guess

`libs/integrations/allegro/src/infrastructure/adapters/allegro-customer-return.mapper.ts:121-131` emits `offerId`, `name`, `quantity`, `unitPrice`, `reasonRaw`, `serialNumbers` — and **no `sku`**. `ReturnsService` maps `offerId: line.offerId ?? null` (`returns.service.ts:583`) into its own column and leaves `ReturnLine.sku` null. Meanwhile the Allegro **order** mapper writes `sku: lineItem.offer.id` (`allegro-order-source.adapter.ts:804`).

So the offer id is present on both sides, under two different field names, and the rule compares neither of them:

- `line.sku` is `null` for every Allegro return line ⇒ the `sku` and `sku+price` branches are unreachable
- Allegro is the **only** `ReturnSourceReader` in the tree (Erli's adapter builds no `IncomingReturnLine` at all), so in practice **100% of resolutions take the price-alone branch** — the weakest axis, and precisely the one that cannot distinguish two lines of the same product
- the duplicate-line case the rule exists to settle (`sku+price`) is therefore dead code on the shipped path, and its spec coverage is fixture-only: `matchedOn: 'sku'` / `'sku+price'` are asserted against hand-built `ResolvableReturnLine`s carrying a `sku` no adapter in this repository produces

This is the "check that cannot fail" shape one level over: the rule is correct, its tests pass, and the axis they prove is unreachable in production.

The fix is small and belongs in this PR, either way round:

1. **Adapter** — `sku: optionalString(item.offerId)` alongside the existing `offerId`, since on Allegro the offer id *is* what the order line's `sku` holds; or
2. **Rule** — take `offerId` as a second identity axis (`ResolvableReturnLine.offerId`, compared against the candidate's `sku`), which is the more honest shape: it says the two sides agree on an identifier without claiming the source reported a SKU it did not.

I'd take (2) and widen `ReturnOrderLineMatchAxisValues` accordingly — `matchedOn` exists so the resolution is auditable, and recording `sku` for a match made on an offer id would misstate what settled it. Whichever you pick, please add one fixture built from the **real** Allegro mapper output (the `allegro-customer-returns.spec.ts` wire fixtures are right there) so the reachable path is the tested path.

### IMPORTANT — `price` alone on a repeated-offer order is `ambiguous` by construction, and the log line is the only place it surfaces

Both sides carry the same per-unit basis (`lineItem.price.amount` vs `item.price.amount`), so the arithmetic is sound and `samePrice`'s minor-unit comparison is the right call. But two order lines of the same offer at the same price are indistinguishable on that axis, which is exactly the shape #2374's picker was opened about — so today the outcome is an `ambiguous` count in a `logger.warn` and a `null` column, and the correction path silently keeps its old name-matching behaviour for that line. That is a defensible interim state, but it is invisible: nothing counts it, no operator surface reports it, and `resolveOrderLinesForReturn`'s summary is discarded by its only caller. Either thread the summary somewhere an operator or a metric can see it, or say in the PR body that unresolved lines are deliberately log-only in this slice and name the follow-up.

### SUGGESTIONS

- `resolveForReturn` returns `void` while computing a perfectly good summary — returning it costs nothing and would let the handler (or a future reconcile) act on `unresolved` without re-reading.
- `orderItems` drops any element whose `productId` is not a string. That is the right call for a partial line, but it means `[]` also covers "every line was malformed"; the docblock says `[]` never means "the order has none", which is correct — a one-line `warn` when `items` was a non-empty array and the filter emptied it would make the difference observable rather than only documented.
- `ReturnOrderLineUnresolvedReasonValues` has no `assertNever` consumer, so a fourth member added later reaches the summary's `Record` silently. Not a defect today (the reason is only counted and logged), but worth a note next to the union while nothing branches on it.

**Verified**: the two cited mapper lines, `returns.service.ts:583`'s `offerId` mapping, `buildRawPayload`'s sparse per-line extras array (`returns.service.ts:537-541`) and the resulting `lineIndex` requirement, the shared per-unit price basis on both Allegro paths, and that no other adapter in the tree implements `listReturnFeed`/`getReturn`.

---

norbert-kulus-blockydevs:
Addressed all findings from the review.

**BLOCKING** — took option (2): added `offerId` as a second identity axis on `ResolvableReturnLine`, compared against the same `ResolvableOrderLine.sku` field the `sku` axis already reads (on Allegro that field *is* the offer id — `sku: lineItem.offer.id`). New `offerId` / `offerId+price` `ReturnOrderLineMatchAxis` members keep the resolution auditable about which axis actually settled it, rather than misreporting an offer-id match as `'sku'`. Axis order is now `sku → offerId → price`, each step a narrowing over the last (never a retry), so a `sku` that positively excludes everything still can't fall through to `offerId` or `price`. New tests drive the real Allegro mapper output (`allegro-customer-returns.spec.ts`) rather than a hand-built fixture carrying a `sku` no adapter produces — confirmed 100% of Allegro return lines now resolve on `offerId`(+`price`), not `price` alone.

**IMPORTANT** — `resolveForReturn` now returns the `ReturnOrderLineResolutionSummary` instead of discarding it (costs nothing to compute, lets a future caller act on `unresolved` without re-reading). Per your own stated alternative, I documented explicitly in the resolver's docblock that `unresolved` counts stay **log-only in this slice** — an `ambiguous` line is real and reachable (the duplicate-line case the picker exists for), and threading it to an operator-visible count/metric is named as a follow-up rather than silently deferred.

**SUGGESTIONS** — all three:
- `resolveForReturn` returns its summary (see above).
- Added a `warn` when `order.orderItems` comes back empty (every line in the snapshot was malformed) — makes that distinct from "the order genuinely has none," observable rather than only documented on `orderItems` itself.
- Added a note beside `ReturnOrderLineUnresolvedReasonValues` recording that no `assertNever` consumes it yet, and why that's fine today.

Also updated `docs/architecture-overview.md`'s #3171 bullet, which described the `offerId` axis as an unimplemented limitation and per-line resolution as unreachable from any manual-match path — both now stale given the fix. Replaced with the real remaining limitations: unresolved counts are log-only, the lifecycle re-read (`marketplace.returns.statusSync`) still doesn't resolve, and `matchOrphanToOrder` doesn't re-trigger resolution for a manually-attributed orphan (stays unresolved until the connection's next per-return sync).

Verification: targeted eslint on every changed file (clean), `pnpm type-check` across the full monorepo (exit 0), targeted jest runs — `return-order-line-resolution.domain-service.spec.ts` (23 passed), `returns.service.spec.ts` (54 passed), `allegro-customer-returns.spec.ts` (57 passed). `pnpm test` / `pnpm test:integration` left to CI as before.

piotrswierzy (commented):
## Tech Lead review

Re-review at `30ca274`. **The BLOCKING finding from the previous round is properly closed**, and closed the better of the two ways offered: `offerId` is a distinct identity axis on `ResolvableReturnLine` compared against the same `ResolvableOrderLine.sku` field, with its own `offerId` / `offerId+price` `ReturnOrderLineMatchAxis` members — so a match made on an offer id is never misreported as `matchedOn: 'sku'`, and the resolution stays auditable about what actually settled it. The axis order stays a narrowing (`sku` → `offerId` → `price`, `resolveByIdentity` returning `no-candidate` and never falling through), and the three new cases in `allegro-customer-returns.spec.ts` drive `toIncomingReturnLine` itself rather than a hand-built fixture carrying a SKU no adapter produces — so the tested path is now the reachable path. The IMPORTANT item and all three suggestions are addressed too (`resolveForReturn` returns its summary, the empty-`orderItems` warn, the `assertNever` note beside the union, and the log-only limitation stated in the docblock rather than implied).

The three invariants this slice had to respect all hold, and each is pinned rather than promised. `resolvedOrderLineId` stays out of **both** halves of `upsertFromSource` and out of `UpsertReturnLineInput` — `order-line-resolution-is-core-owned.spec.ts` asserts it by scanning code with a comment strip that carries its own self-test, and its `toLineInput` window is sliced forward from one `indexOf` precisely so it cannot invert into a vacuous pass. The write is a fill-in-when-NULL claim (`"resolvedOrderLineId" IS NULL`, the `claimAttribution` idiom), so no re-resolve can un-resolve a line. And the `returns → orders` edge stays one-way: `ReturnsModule.imports` is untouched, `OrdersModule` is still absent, and the order read is composed in `apps/worker` — `check-cross-context-imports.mjs` and `barrel-purity.spec.ts` take zero new allow-set entries, which is the right evidence that the placement is correct rather than merely convenient. `OrderRecord.orderItems` is a pure ADR-011 derivation that drops a partial line rather than coercing it, and `readReportedUnitPrice` matching on the entry's own `lineIndex` rather than its array position remains necessary and correct given `buildRawPayload`'s sparse array. CI is green across all nine checks including Integration Tests.

### Findings

**[IMPORTANT]** — `libs/core/src/returns/application/services/returns.service.interface.ts:328-330`
> `matchOrphanToOrder`'s docblock still reads *"Line resolution is deliberately NOT re-run: **nothing in the tree populates** `ReturnLine.resolvedOrderLineId`, so there is no resolution to re-run and inventing one here would be a second, undesigned mapping."* This PR makes that premise false, and the identical sentence in `docs/architecture-overview.md` **was** correctly rewritten in this diff (to the honest *"not re-triggered by this action … stays unresolved until the connection's next per-return sync"*). Leaving the code docblock behind makes the doc and its source disagree about the same decision, in a file this PR already modifies — and the stale version supplies a *reason* a future reader would act on (it says re-running is impossible, when it is now merely not wired). Bring it into line with the doc wording.

**[SUGGESTION]** — `libs/core/src/returns/returns.module.ts:116-119` and `docs/architecture-overview.md:471`
> Two more sites carry the same now-stale claim, both as parenthetical justification for the #2370 restock path resolving by SKU through `IProductsService`: *"`resolvedOrderLineId` is a by-value reference into the order snapshot's jsonb and **nothing populates it yet**"*. The conclusion (restock resolves by SKU) is unaffected, so this is not a behaviour risk — but the repo treats a stale rationale as worse than none, and these two plus the one above are the complete set. Cheap to fix in the same pass.

**[SUGGESTION]** — `apps/worker/src/sync/return-order-line-resolver.service.ts`
> `resolveForReturn` now returns a summary that no caller reads — `MarketplaceReturnSyncHandler` still `await`s it and discards it. That is the stated interim posture and it is fine, but it does mean the value is currently produced for a reader that does not exist; worth naming the intended first consumer (the follow-up that narrows `return-correction-matching.domain-service.ts` onto the column) in the docblock so the next reader can tell "reserved" from "forgotten".

**[SUGGESTION]** — `libs/core/src/returns/application/services/returns.service.ts`
> `resolveOrderLinesForReturn` returns a zero summary when `orderLines.length === 0`, which the worker already guards against with its own early `return null` + warn. Harmless defence in depth, but the two arms are now indistinguishable to a caller that only has the summary (`resolved: 0` from "no usable lines" reads identically to `resolved: 0` from "every line was ambiguous"). If the summary is going to grow a consumer, that distinction is worth a reason member rather than an implicit zero.

Positive observations worth recording, since all three are easy to get wrong and were got right: the SKU/offer-id normalisation is trimmed but deliberately **not** case- or diacritic-folded (an identifier is not display text, unlike `normalizeCorrectionLineName`); product name is refused as an axis entirely, which is the collision the rule exists to remove rather than relocate; and the whole pass is best-effort with every failure swallowed and logged, which is correct because the return is already durably persisted by the time it runs — failing the job would trade a missing attribution for a lost return.

### Verdict

🔄 **Approve with changes**

The design is sound, the previous BLOCKING is closed at the right layer, and nothing here needs to be re-architected. Posting as a comment rather than an approval only because one IMPORTANT item stands.

1. Update `returns.service.interface.ts:328-330` to match the doc wording this PR already corrected.
2. (Optional, same pass) The two other "nothing populates it yet" sites.

---

piotrswierzy (commented):
## Tech Lead re-review (head unchanged at `30ca274`)

Re-verification pass, not a fresh review — nothing has been pushed since the 2026-09-13 round, and `main` has advanced by 11 commits in the meantime. I re-checked each prior finding in code at this head rather than restating the review: **all four still stand, verbatim and at the same lines**, and nothing on the new `main` supersedes any of them or touches this PR's subject. The three invariants the slice rests on were re-confirmed too — `resolvedOrderLineId` appears in neither half of `upsertFromSource` nor in `UpsertReturnLineInput` (still pinned by `order-line-resolution-is-core-owned.spec.ts`), the write is still the fill-in-when-NULL `"resolvedOrderLineId" IS NULL` claim so a re-resolve cannot un-resolve, and the `returns → orders` edge is still one-way with the order read composed in `apps/worker` and zero new entries in either guard's allow-set. `git merge-tree` against the new `main` produces a clean tree — no conflicts; the branch is simply `BEHIND` and needs an update. CI is green on all nine checks.

### Findings

**[IMPORTANT]** — `libs/core/src/returns/application/services/returns.service.interface.ts:328-330`
> Still open, unchanged. The docblock reads *"Line resolution is deliberately NOT re-run: **nothing in the tree populates** `ReturnLine.resolvedOrderLineId`, so there is no resolution to re-run…"* — a premise this PR makes false. I re-confirmed both sides of the disagreement at this head: the identical sentence in `docs/architecture-overview.md:486` **was** correctly rewritten in this same diff to *"Line resolution is not re-triggered by this action — `resolvedOrderLineId` (#3171) is populated by `ReturnOrderLineResolverService` … a manually-matched orphan's lines therefore stay unresolved until the connection's next per-return sync"*. So doc and source still disagree about one decision, in a file this PR already modifies, and the stale side supplies a *reason* (re-running is impossible) that is now merely "not wired". Bring it into line with the doc wording.

**[SUGGESTION]** — `libs/core/src/returns/returns.module.ts:116-118` and `docs/architecture-overview.md:471`
> Both still carry the same now-stale parenthetical — *"`resolvedOrderLineId` is a by-value reference INTO the order snapshot's jsonb and **nothing populates it yet**"* — as justification for the #2370 restock path resolving by SKU through `IProductsService`. The conclusion is unaffected, so no behaviour risk; these two plus the one above remain the complete set.

**[SUGGESTION]** — `apps/worker/src/sync/return-order-line-resolver.service.ts:70`
> `resolveForReturn` still returns `Promise<ReturnOrderLineResolutionSummary | null>` and `MarketplaceReturnSyncHandler:61` still `await`s it and discards it — the only production caller. Naming the intended first consumer in the docblock still distinguishes "reserved" from "forgotten".

**[SUGGESTION]** — `libs/core/src/returns/application/services/returns.service.ts`
> Unchanged: `resolved: 0` from "no usable lines" is still indistinguishable from `resolved: 0` from "every line was ambiguous" to a caller holding only the summary.

### Verdict

🔄 **Approve with changes** — verdict unchanged from the prior round.

Nothing on the new `main` changes the picture, the merge is clean, and the one IMPORTANT item is a two-line docblock edit in a file the PR already touches.

---

piotrswierzy (commented):
## Tech Lead re-verification (head unchanged at `30ca274`)

Head is unchanged, `main` is unchanged (`e336cf4`), and all nine CI checks pass. `MERGEABLE` / `BEHIND` — a fast-forward of the base, no conflicts, and no migration in the diff so none of the contended timestamps applies. The single standing IMPORTANT finding re-verified and still holds; the three invariants I called sound last round re-verified and are still sound. Verdict stands.

### Findings

**[IMPORTANT]** — `libs/core/src/returns/application/services/returns.service.interface.ts:328-330` — doc and source now disagree about one decision, inside a file this PR already modifies.

The interface still carries, as the stated justification for `matchOrphanToOrder` not re-running resolution:

> Line resolution is deliberately NOT re-run: **nothing in the tree populates `ReturnLine.resolvedOrderLineId`**, so there is no resolution to re-run and inventing one here would be a second, undesigned mapping.

Meanwhile this PR's own diff rewrote the identical sentence in `docs/architecture-overview.md:486` correctly:

> `resolvedOrderLineId` (#3171) **is populated by `ReturnOrderLineResolverService`**, called from the per-return sync and the lifecycle re-read, neither of which `matchOrphanToOrder` invokes; a manually-matched orphan's lines therefore stay unresolved until the connection's next per-return sync rather than immediately…

That rewrite is exactly right, and it is the *reason* the interface docblock is now wrong: the decision is still correct (don't re-run here), but the premise it rests on has been falsified by this very PR. A reader of the interface — which is the contract surface, and the thing a future implementer actually opens — is told that a column this PR just started writing is written by nothing. Port the `architecture-overview.md` sentence verbatim; the two should not be able to drift on a decision this narrow.

### Invariants re-confirmed

All three verified against the current head, and all three are genuinely well-built:

1. **`resolvedOrderLineId` is in neither half of `upsertFromSource`, nor in `UpsertReturnLineInput`.** Confirmed by reading the method body — the column does not appear. The structural spec at `libs/core/src/returns/__tests__/order-line-resolution-is-core-owned.spec.ts` pins it, and it is worth calling out that this spec is built the right way round: its first case is *"should still see the column when it is real code, not a comment"*, which is the guard against the spec silently passing because its own source-scan stopped matching. That is the failure mode most structural specs in this repo have had to learn the hard way, and this one ships with the defence already in place.
2. **The write is a fill-in-when-NULL claim** — `return.repository.ts:1058` carries `.andWhere('"resolvedOrderLineId" IS NULL')`, so a re-resolve can fill the value in and can never *un*-resolve one. This is the `claimWaybillRelay` / `claimAttribution` idiom applied correctly, and it is what makes the resolver safe to re-drive from both the per-return sync and the lifecycle re-read without a lock.
3. **`returns → orders` stays one-way, with zero new allow-set entries.** Neither `barrel-purity.spec.ts`, `check-no-injection-contracts.mjs` nor `check-cross-context-imports.mjs` is touched by this diff — which, per `ADR-053`'s own reasoning, is the strongest available evidence the seam was placed correctly rather than routed around. Needing an allow-set entry would have been the signal it was wrong.

### Dependency to report

**#3123's mockup names this PR as a hard prerequisite.** #3123 cannot ship against an unpopulated `resolvedOrderLineId`, so the merge order is #3172 → #3123. Flagging it here so the sequencing is recorded on the PR rather than only in the epic.

### Verdict

🔄 **Approve with changes**

One fix: replace the stale sentence at `returns.service.interface.ts:328-330` with the corrected wording this PR already wrote into `docs/architecture-overview.md:486`.

---

norbert-kulus-blockydevs:
Addressed all four standing findings from the 2026-09-14 round. Head is now `9f9e0bfef`.

**IMPORTANT — `returns.service.interface.ts:328-330`.** Fixed. Ported the corrected `docs/architecture-overview.md` wording verbatim, so the interface now reads *"Line resolution is not re-triggered by this action — `resolvedOrderLineId` (#3171) is populated by `ReturnOrderLineResolverService`, called from the per-return sync and the lifecycle re-read, neither of which this invokes; a manually-matched orphan's lines therefore stay unresolved until the connection's next per-return sync rather than immediately, a stated limitation rather than a gap in the model."* The decision is unchanged; only the falsified premise is gone.

**SUGGESTION — the other "nothing populates it yet" sites.** Fixed, and there were **four, not three**. Beyond `returns.module.ts:116-118` and `docs/architecture-overview.md:471`, `return-custody.service.ts:673`'s `resolveRestockTarget` docblock carried the same claim in a third spelling (*"nothing in the shipped model populates it yet"*), which is why the grep for the two known phrasings missed it. All four now state that the field is written only at ingestion and stays `null` for an orphan or a line the resolver could not settle — the conclusion each one supports (restock resolves by sku through `IProductsService`) is unaffected and still correct. A repo-wide grep for all three phrasings now returns nothing.

**SUGGESTION — the summary's absent reader.** Named it. The resolver docblock now says the value is **reserved, not forgotten**, that `MarketplaceReturnSyncHandler` currently awaits and discards it, and that the intended first consumer is the follow-up narrowing `return-correction-matching.domain-service.ts` onto `ReturnLine.resolvedOrderLineId` and retiring its `ambiguous` line picker (#3171 § 4, epic #3087) — which is exactly the reader for which `unresolved` has to be a count rather than only a log line.

**SUGGESTION — `resolved: 0` carrying three meanings.** Fixed rather than documented, since the summary does now have a named future consumer. Added a **pass-level** `ReturnOrderLineResolutionSkipReason` (`unknown-return` / `no-order-lines` / `no-return-lines`) reported as `ReturnOrderLineResolutionSummary.skipped`. Three points on the shape:

- It is deliberately **not** folded into `ReturnOrderLineUnresolvedReason`, which is per-LINE and counts lines the rule looked at and could not settle. A pass that examined nothing is a different kind of fact, and reusing that union would have made `unresolved` mean two things.
- `skipped: null` together with three zeros is now **unreachable**: a pass that looked at a line always reports it under `resolved`, `alreadyResolved` or `unresolved`. That is the property stated on the field itself, not just in prose.
- The third arm (`no-return-lines`) is new — the old `record === null || orderLines.length === 0` guard did not cover a return that exists and carries no lines, which also produced bare zeros.

The worker reports a named skip at `log` rather than `warn`: nothing failed, the pass simply had nothing to examine. Four new specs pin all three reasons plus the absence case (a pass that *did* examine a line must never carry a skip reason) — that last one is what stops the discriminator from being satisfiable by always setting it.

**Note on formatting:** three of the touched files (`returns.service.spec.ts`, `returns/index.ts`, `return-order-line-resolver.service.ts`) were already prettier-dirty at `30ca274`, so I did not run `--write` on them — doing so would have buried this diff in pre-existing churn unrelated to the review.

**Verification:** eslint on every changed file (0 errors; the 4 remaining warnings in the spec are pre-existing and sit above the additions), `pnpm type-check` across the full monorepo (exit 0), and the returns suites — **129 tests across 7 files**, including `order-line-resolution-is-core-owned.spec.ts` (the `resolvedOrderLineId` ownership guard) and `returns-orders-module-boundary.spec.ts`. Full `pnpm test` / `pnpm test:integration` left to CI as before.

No invariant moved: `resolvedOrderLineId` is still absent from both halves of `upsertFromSource` and from `UpsertReturnLineInput`, the write is still the fill-in-when-NULL `"resolvedOrderLineId" IS NULL` claim, and the `returns → orders` edge is still one-way with zero new entries in any guard's allow-set.

piotrswierzy (commented):
## Tech Lead re-review (`30ca274b` → `0cd02365`)

The delta is a `main` merge plus one targeted fix commit, and that fix closes the doc/source disagreement I raised. `returns.service.interface.ts:361` now reads *"**Line resolution is not re-triggered by this action** — `resolvedOrderLineId` (#3171) is populated by `ReturnOrderLineResolverService`, called from the per-return sync and the lifecycle re-read, neither of which this invokes"*, and it goes on to name the consequence rather than stopping at the correction: a manually-matched orphan's lines stay unresolved until the connection's next per-return sync, *"a stated limitation rather than a gap in the model"*. That is the better version of the fix — the original justification was false and deleting it would have left `matchOrphanToOrder` with no stated reason at all, whereas this one is true, still justifies the behaviour, and tells an operator-facing reader what to expect.

Re-confirmed all three invariants at head, in code:

- **`resolvedOrderLineId` is absent from both halves of `upsertFromSource` and from `UpsertReturnLineInput`** — and `order-line-resolution-is-core-owned.spec.ts` pins it structurally in four directions, including the one that matters most (`should still see the column when it is real code, not a comment`), which is what stops the guard degrading into a grep over prose.
- **The write is a fill-in-when-NULL claim** — `return.repository.ts:1058` is `.andWhere('"resolvedOrderLineId" IS NULL')` on the `claimOrderLineResolution` update, answering `affected > 0`, so a re-resolve can fill the value in and can never *un*-resolve or overwrite one. Same shape as `claimAttribution` / `claimAuthorizedAt` beside it.
- **`returns → orders` stays one-way** — the PR touches neither `scripts/check-cross-context-imports.mjs` nor `barrel-purity.spec.ts` relative to `main`, so no allow-set entry is spent and the order's lines still arrive as an argument rather than through an injected `orders` service.

### Findings
**[IMPORTANT]** — `libs/core/src/returns/application/services/returns.service.interface.ts:58`
> The same retired claim survives one docblock up, in the file this commit already edits. `RecordReturnLineInput` still justifies excluding the field with *"`resolvedOrderLineId` has no populator anywhere in the tree (it is a by-value reference into the order snapshot's jsonb)"* — and this PR is precisely what gives it one. The parenthetical is still true and is on its own sufficient reason to keep the field out of an operator-authored create input; the clause before it is now false. Worth the same treatment `:361` just got: drop the no-populator half, keep the by-value-reference half, and the justification survives intact. Not blocking — it is a comment, `RecordReturnLineInput`'s shape is correct either way, and the structural spec is what actually enforces the exclusion — but a claim that was wrong once in this file has now been wrong twice, and leaving the second instance is how a reviewer's correction reads as cosmetic.

### Verdict
🔄 **Approve with changes** — the one residual is a comment edit in a file already open in this diff.

**Prerequisite status you asked about:** #3123 (*docs(mockups): correction-proposal UI mockup*) names this PR as a hard prerequisite and is still **OPEN**, as is this one — so nothing there is unblocked yet, and #3123 stays gated on this landing.

CI is mid-run — Docker Build Smoke, PHP Unit, Scaffolded Adapter Builds and Detect changed paths pass; Lint, Type Check, Build, Test and Integration Tests are `pending` and are not reported here as green. `mergeable: MERGEABLE`, `mergeStateStatus: BLOCKED` on those pending checks. Base is `main`, now merged up to `e336cf4`.

---

norbert-kulus-blockydevs:
## Review findings addressed - `33ac037`

All standing findings are closed. One IMPORTANT fixed, plus a third instance of the same stale-premise shape that the review had not reached.

---

### [IMPORTANT] `returns.service.interface.ts:58` - `RecordReturnLineInput` still claimed "no populator anywhere in the tree" - **FIXED**

Taken the way you framed it: the parenthetical is the half that survives. The by-value-reference clause is still true and is on its own sufficient reason to keep the field out of an operator-authored create input, so it is kept; the no-populator clause is gone.

It is not a bare deletion though, for the reason the `:361` fix was the better version of itself - deleting the false clause would have left the exclusion with a thinner reason than it deserves. The replacement states the accurate one (**core-resolved, never operator-supplied**) and then names the consequence, which is a real fact a reader of this input needs:

> `resolvedOrderLineId` is **core-resolved, never operator-supplied** - it is a by-value reference into the order snapshot's jsonb, so there is no value an operator could meaningfully type even if the field were offered. Since #3171 it does have a populator, `ReturnOrderLineResolverService`, but that runs only on the marketplace per-return sync path; an operator-authored return's lines therefore stay `null`, a stated limitation rather than a gap in the model.

`recordReturn` is not on the resolver's call path (the resolver's only caller is `MarketplaceReturnSyncHandler`), so an operator-authored return's lines really do stay `null` forever. That was previously unstated anywhere.

---

### [FOUND WHILE FIXING] Three docblocks claimed the resolver has two callers. It has one, and this PR's own body says so - **FIXED**

While porting the `:361` wording I checked the claim it makes, and it is false:

> `resolvedOrderLineId` (#3171) is populated by `ReturnOrderLineResolverService`, called from **the per-return sync and the lifecycle re-read**

Evidence:

- `grep -rn "resolveForReturn|ReturnOrderLineResolverService"` over the tree returns exactly one production call site: `apps/worker/src/sync/handlers/marketplace-return-sync.handler.ts:61`.
- `MarketplaceReturnsStatusSyncHandler` injects `IReturnStatusSyncService` + `ConnectionCursorRepositoryPort` and nothing else. It does not import, inject or invoke the resolver.
- `docs/architecture-overview.md:478`, in the same document and two paragraphs above the offending sentence, already states the opposite correctly: *"the `marketplace.returns.statusSync` lifecycle re-read does **not** resolve, because `ReturnStatusSyncResult` reports counters only and carries no list of the returns it touched"*. The PR body says the same thing under Limitations.

So the document contradicted itself and the interface contract carried the false half - the same shape as the two you did catch, one layer further out. Corrected at all three sites:

| Site | Now reads |
|---|---|
| `returns.service.interface.ts:367` (`matchOrphanToOrder`) | "whose only caller is the `marketplace.return.sync` per-return path, which this does not invoke. The `marketplace.returns.statusSync` lifecycle re-read does not resolve either (`ReturnStatusSyncResult` reports counters and carries no list of the returns it touched)" |
| `docs/architecture-overview.md:486` | same correction, cross-referencing the `:478` bullet rather than restating it |
| `return-order-line-resolver.service.ts:58` | "**There is exactly one caller today**, `MarketplaceReturnSyncHandler`." |

The resolver docblock also lost a second overclaim in the same pass. It said idempotency "is what makes it safe to call from both the per-return sync and the lifecycle re-read"; it now says it makes the pass safe to **re-drive without a lock**, and names the two routes by which `marketplace.return.sync` actually runs again for an already-ingested return - the poll fan-out, or an inbound `'return'` webhook since #2400. That is what the "only route by which a pre-#3171 return gets resolved" claim rests on, and it was previously left implicit.

---

### Previously-raised items - re-verified closed at this head

Not re-doing work, just recording the evidence so the round is self-contained.

| Finding | State |
|---|---|
| **[R2/R3 IMPORTANT]** `returns.service.interface.ts:328-330` stale `matchOrphanToOrder` premise | Closed in `9f9e0bfe`; you confirmed it in the 13:23 round. Now at `:364` and additionally corrected above. |
| **[R2/R3 SUGG]** `returns.module.ts` + `docs/architecture-overview.md:471` "nothing populates it yet" | Closed. `grep -rn "nothing populates\|no populator\|populates it yet"` over the tree now returns **zero** hits. `9f9e0bfe` also fixed a **fourth** site the review had not named, `return-custody.service.ts:671-676` (`resolveRestockTarget`). All four now say the field is written only at ingestion and stays `null` for an orphan or an unsettled line; the conclusion each supports (restock resolves by sku) is unchanged. |
| **[R2/R3 SUGG]** name the intended first consumer of the returned summary | Closed. `return-order-line-resolver.service.ts` carries "**RESERVED, not forgotten: no caller reads this value today**" and names the follow-up that narrows `return-correction-matching.domain-service.ts` onto the column (#3171 section 4, epic #3087). |
| **[R2/R3 SUGG]** `resolved: 0` from "no usable lines" indistinguishable from "every line ambiguous" | Closed. `ReturnOrderLineResolutionSummary` carries a `skipped` reason; the interface states it ("that summary carries a `skipped` reason, so 'no line was examined' never reads as 'every line was examined and none resolved'") and the worker logs that arm at `log` rather than `warn`. |
| **[R1 BLOCKING]** SKU axis unreachable on the only shipped `ReturnSourceReader` | Closed at `30ca274` the way you preferred - `offerId` as its own identity axis with its own `matchedOn` members, driven through the real Allegro mapper in `allegro-customer-returns.spec.ts`. |
| **[R1 IMPORTANT + 3 SUGG]** | Closed at `30ca274`, confirmed in your 2026-09-13 round. |

---

### Verification

Comments only in this commit - no behaviour, type or test change.

- `pnpm --filter @openlinker/core type-check` - clean.
- `npx eslint` on both changed TS files - exit 0.
- `pnpm --filter @openlinker/core test -- --testPathPattern "returns/(__tests__|application|domain)"` - **24 suites, 416 tests, all pass**.
- `pnpm check:invariants` - exit 0.

**Red-first evidence for the structural guard.** My edits add `resolvedOrderLineId` to docblocks, which is precisely what `order-line-resolution-is-core-owned.spec.ts`'s comment strip exists to tolerate - so the strip is the thing I could have blinded, and I checked it rather than assuming:

1. Added `resolvedOrderLineId: string | null;` to `UpsertReturnLineInput` as **real code**.
2. Guard **FAILS**: `1 failed, 3 passed`, at `order-line-resolution-is-core-owned.spec.ts:71` on `expect(source).not.toContain('resolvedOrderLineId')`.
3. Restored the file (`git diff` empty), guard **passes 4/4**.

So the strip still removes only comments and the guard still has teeth against a real member. For completeness: none of the three files this commit touches is scanned by that guard - it reads `return-upsert.types.ts`, `returns.service.ts` and `return.repository.ts`.

Nothing was declined.

---

Recording your sequencing note: **#3172 -> #3123**, since #3123 cannot ship against an unpopulated `resolvedOrderLineId`.

piotrswierzy (approved):
## Tech Lead re-review (`0cd02365` → `33ac0378`)

**The single open finding is closed, and the fix corrected a second inaccuracy I had let through.** One own commit, `33ac0378`, three files, docs only — no behaviour change.

**Surviving copies of the retired claim: zero.** Grepping the whole tree for `populator` at head returns no instance of the *"has no populator anywhere in the tree"* claim. Before this commit there was exactly one, at `libs/core/src/returns/application/services/returns.service.interface.ts:58` inside `RecordReturnLineInput`; it is rewritten to say `resolvedOrderLineId` is **core-resolved, never operator-supplied** — a by-value reference into the order snapshot's jsonb, so no value an operator could meaningfully type — and that since #3171 it *does* have a populator, `ReturnOrderLineResolverService`, which runs only on the marketplace per-return path, leaving an operator-authored return's lines `null` as a stated limitation. Every other `populator` hit in the repo belongs to the unrelated Erli variant-group work (`erli-offer-manager.adapter.ts:65` plus four `docs/plans/implementation-plan-erli-*` files). The plan doc this PR adds carries no copy either.

The commit also went further than the finding: my `:361` fix said the resolver is *"called from the per-return sync and the lifecycle re-read"*, and that second half was itself wrong. All three places now state the true caller set consistently —

- `apps/worker/src/sync/return-order-line-resolver.service.ts:53` — **"There is exactly one caller today, `MarketplaceReturnSyncHandler`"**, with the reason `marketplace.returns.statusSync` cannot be one: `ReturnStatusSyncResult` reports counters only and carries no list of the returns it touched, so that pass has no id to pass in.
- `returns.service.interface.ts:366` and `docs/architecture-overview.md` § 22 — the same statement, same reason, in the `matchOrphanToOrder` limitation note.

Verified against the tree rather than the prose: `ReturnOrderLineResolverService` is imported in exactly three non-test files — its own definition, `marketplace-return-sync.handler.ts`, and `sync-worker.module.ts` (provider registration at `:135`). The status-sync handler does not reference it.

The three invariants verified previously still hold at head:

- **`resolvedOrderLineId` is absent from both halves of `upsertFromSource` and from `UpsertReturnLineInput`**, and structurally pinned — `return-repository.port.ts:115` names it explicitly in the "appear in NEITHER half" list alongside `receivedAt` / `disposedAt`, so re-ingestion cannot un-resolve a line any more than it can un-receive a parcel.
- **The write is a fill-in-when-NULL claim** — `.andWhere('"resolvedOrderLineId" IS NULL')` at `return.repository.ts:1058`, on the `.set({ resolvedOrderLineId: orderLineId, updatedAt: () => 'now()' })` at `:1056`. A re-resolve can fill the value in and can never change one, which is what makes the resolver safe to re-drive without a lock — now the stated reason in the service docblock, replacing the weaker "safe to call from both" phrasing.
- **`returns → orders` stays one-way.** The PR touches no guard script: `scripts/check-cross-context-imports.mjs` and `barrel-purity.spec.ts` are absent from the 20-file diff against `main`, so no allow-set entry was spent.

**Consumer status for #3123** (`docs(mockups): correction-proposal UI mockup`, open): this PR is its hard prerequisite and is now review-clean. It is the only one of the three based on `main`, so it is the one that can actually land — `mergeable: MERGEABLE`, `mergeStateStatus: BLOCKED` purely on required checks not yet reported. `Docker Build Smoke Test`, `Detect changed paths`, `PHP Unit Tests` and `Scaffolded Adapter Builds` pass; `Build`, `Lint`, `Test`, `Type Check` and `Integration Tests` are **pending with 0s elapsed** — a queued run, not a failure. Given the diff is three docblocks, a green run is the expected outcome; merge once it reports.

### Findings

### Verdict
✅ **Approve**

---

PR #3198, opened 2026-09-11
feat(inventory): locations CRUD UI — full operator-facing management for inventory_locations

### LINKED ISSUE #3063: [EPIC] Frontend — Inventory locations CRUD UI

norbert-kulus-blockydevs:
## Summary

`inventory_locations` (#2313/#2316) has had a complete CRUD API since #2316, but the frontend only ever called `listActiveLocations` (a degenerate `limit=1` probe) and `bootstrapLocations` — list, single-item read, create, update and delete were all unreachable from the product, reachable only via `curl`. This epic closes that gap.

Stacked sub-issues landed in this branch, each reviewed and CI-green individually:
- #3064 — extend `InventoryApi` with `listLocations` / `getLocation` / `createLocation` / `updateLocation` / `deleteLocation`
- #3065 — query hooks + filters (kind, status, codePrefix)
- #3067 — create/edit location dialog
- #3068 — delete confirm, including the 409 "location in use" case with a retire-instead fallback
- #3066 — list page with filter bar
- #3069 — nav entry (mounted as a Settings tile, per a later refinement on the branch)
- #3070 — CRUD tests

## What changed

- `apps/web/src/pages/inventory/inventory-locations-page.tsx` (+tests) — the operator-facing list/create/edit/delete/retire surface
- `apps/web/src/features/inventory/**` — API client methods, types, query/mutation hooks
- `apps/web/src/pages/settings/settings-page.tsx` — mount point (Settings tile, not a dedicated nav item)
- Minor shared-UI fixes surfaced along the way (`confirm-dialog.tsx`, `api-error.ts`, `session.types.ts`)

Built against the reviewed mockup: docs/plans/mockups/inventory-locations-3063.html (design issue #3113/#3120, Artifact: https://claude.ai/code/artifact/fbb5f9e1-52a3-4b7b-8136-ad59dcf4f318)

## Test plan
- [x] Unit tests for API client, hooks, and the list/create/edit/delete/retire flows (all green per individual sub-issue PRs #3131–#3137, all merged)
- [x] CI green on the final integration PR (#3131): type-check, lint, unit tests, integration tests, build, docker smoke — all SUCCESS

Closes #3063

### DIFF
--- a/apps/api/src/inventory/http/inventory-locations.controller.spec.ts
+++ b/apps/api/src/inventory/http/inventory-locations.controller.spec.ts
@@ -17,6 +17,7 @@ import {
   LocationInUseError,
   type ILocationService,
 } from '@openlinker/core/inventory';
+import { ROLE_PERMISSIONS, UserRoleValues } from '@openlinker/core/users';
 import { InventoryLocationsController } from './inventory-locations.controller';
 
 function makeLocation(
@@ -221,5 +222,19 @@ describe('InventoryLocationsController', () => {
       expect(result.created).toEqual([]);
       expect(result.existingCodes).toEqual(['MAIN']);
     });
+  });
+});
+
+describe('inventory-locations:write permission lockstep', () => {
+  it('should be granted to exactly the roles this controller @Roles(admin) on every write route', () => {
+    // `inventory-locations:write` is a display predicate only (no permission
+    // guard exists) - if this drifts from the controller's actual @Roles,
+    // either add a permission-based guard or revert the grant, the
+    // `shipments:write` / `automations:write` discipline.
+    const rolesWithPermission = UserRoleValues.filter((role) =>
+      ROLE_PERMISSIONS[role].includes('inventory-locations:write')
+    );
+
+    expect(rolesWithPermission).toEqual(['admin']);
   });
 });
--- a/apps/web/src/app/nav-registry.ts
+++ b/apps/web/src/app/nav-registry.ts
@@ -71,6 +71,11 @@ export const BASE_NAV_GROUPS: readonly NavRegistryGroup[] = [
     label: 'Platform',
     items: [
       { to: '/connections', label: 'Connections', countKey: 'connections' },
+      // Reads are `admin`/`operator`/`viewer` (InventoryLocationsController)
+      // - gated on `inventory:read` rather than `inventory-locations:write`,
+      // so a `packer` (which holds no permissions at all) is the only role
+      // that never sees the entry.
+      { to: '/inventory/locations', label: 'Inventory locations', requiresPermission: 'inventory:read' },
       { to: '/adapters', label: 'Adapters' },
       { to: '/settings', label: 'Settings' },
     ],
--- a/apps/web/src/app/nav-registry.ts
+++ b/apps/web/src/app/nav-registry.ts
@@ -71,11 +71,6 @@ export const BASE_NAV_GROUPS: readonly NavRegistryGroup[] = [
     label: 'Platform',
     items: [
       { to: '/connections', label: 'Connections', countKey: 'connections' },
-      // Reads are `admin`/`operator`/`viewer` (InventoryLocationsController)
-      // - gated on `inventory:read` rather than `inventory-locations:write`,
-      // so a `packer` (which holds no permissions at all) is the only role
-      // that never sees the entry.
-      { to: '/inventory/locations', label: 'Inventory locations', requiresPermission: 'inventory:read' },
       { to: '/adapters', label: 'Adapters' },
       { to: '/settings', label: 'Settings' },
     ],
--- a/apps/web/src/app/routes/inventory-locations.route.tsx
+++ b/apps/web/src/app/routes/inventory-locations.route.tsx
@@ -1,0 +1,13 @@
+import type { RouteObject } from 'react-router-dom';
+import type { RouteCrumbHandle } from '../nav-registry.types';
+
+export const inventoryLocationsRoute: RouteObject = {
+  path: 'inventory/locations',
+  handle: { crumb: { group: 'Platform', title: 'Inventory locations' } } satisfies RouteCrumbHandle,
+  lazy: async () => {
+    const { InventoryLocationsPage } = await import(
+      '../../pages/inventory/inventory-locations-page'
+    );
+    return { Component: InventoryLocationsPage };
+  },
+};
--- a/apps/web/src/app/routes/root.route.tsx
+++ b/apps/web/src/app/routes/root.route.tsx
@@ -24,6 +24,7 @@ import { cursorsRoute } from './cursors.route';
 import { customersRoute } from './customers.route';
 import { devUiRoute } from './dev-ui.route';
 import { insightsRoute } from './insights.route';
+import { inventoryLocationsRoute } from './inventory-locations.route';
 import { listingsRoute } from './listings.route';
 import { aiProviderSettingsRoute } from './ai-provider-settings.route';
 import { mcpTokensRoute } from './mcp-tokens.route';
@@ -71,6 +72,7 @@ export const coreChildren: RouteObject[] = [
   automationsRoute,
   invoicesRoute,
   connectionsRoute,
+  inventoryLocationsRoute,
   adaptersRoute,
   newConnectionRoute,
   advancedNewConnectionRoute,
--- a/apps/web/src/app/routes/route-lazy.test.ts
+++ b/apps/web/src/app/routes/route-lazy.test.ts
@@ -83,7 +83,7 @@ const lazyRoutes = collectLazyRoutes([
  *   - prompt-templates-legacy-redirects (inline `<Navigate>` element)
  *   - `/analytics` legacy alias (inline `<Navigate>` to `/`, #2740)
  */
-const EXPECTED_LAZY_ROUTE_COUNT = 63;
+const EXPECTED_LAZY_ROUTE_COUNT = 64;
 
 describe('route lazy contract', () => {
   it(`the registered route tree contains exactly ${EXPECTED_LAZY_ROUTE_COUNT} lazy routes`, () => {
--- a/apps/web/src/features/inventory/api/inventory-locations.types.ts
+++ b/apps/web/src/features/inventory/api/inventory-locations.types.ts
@@ -7,6 +7,16 @@
  * @module apps/web/src/features/inventory/api
  */
 
+/**
+ * `countryIso2` is normalised to uppercase in exactly one place — the API's
+ * own `ListLocationsQueryDto` does the same server-side, so this mirrors that
+ * rule rather than inventing a second one. #3135 review: was duplicated
+ * inline in `inventory.api.ts` and `location-dialog.schema.ts`.
+ */
+export function normalizeCountryIso2(value: string): string {
+  return value.toUpperCase();
+}
+
 /** What a location physically is (#2316). Operator-declared, never derived. */
 export const InventoryLocationKindValues = ['warehouse', 'store', 'third-party', 'virtual'] as const;
 export type InventoryLocationKind = (typeof InventoryLocationKindValues)[number];
--- a/apps/web/src/features/inventory/api/inventory-locations.types.ts
+++ b/apps/web/src/features/inventory/api/inventory-locations.types.ts
@@ -1,11 +1,50 @@
 /**
- * Inventory Location Types (frontend view of the #2313 / #2407 wire contract)
+ * Inventory Location Types (frontend view of the #2313 / #2316 / #2407 wire contract)
  *
  * The browser bundle cannot depend on `@openlinker/core` (#591), so these are
  * copies of the API's response shapes rather than imports.
  *
  * @module apps/web/src/features/inventory/api
  */
+
+/** What a location physically is (#2316). Operator-declared, never derived. */
+export const InventoryLocationKindValues = ['warehouse', 'store', 'third-party', 'virtual'] as const;
+export type InventoryLocationKind = (typeof InventoryLocationKindValues)[number];
+
+/**
+ * Whether the location participates in routing/availability today.
+ * `inactive` is a soft retirement, not a delete — see `deleteLocation`.
+ */
+export const InventoryLocationStatusValues = ['active', 'inactive'] as const;
+export type InventoryLocationStatus = (typeof InventoryLocationStatusValues)[number];
+
+/**
+ * The full row — mirrors `LocationResponseDto` field-for-field (#3064).
+ * `InventoryLocationSummary` below stays the minimal shape the #2407
+ * readiness/bootstrap surfaces already read; this is its superset, so an
+ * `InventoryLocation` satisfies every existing `InventoryLocationSummary`
+ * consumer without a second type walking the same fields.
+ */
+export interface InventoryLocation {
+  readonly id: string;
+  readonly code: string;
+  readonly name: string;
+  readonly kind: InventoryLocationKind;
+  /** Provenance only — whose sync may write positions here. Never authority (ADR-052). */
+  readonly ownerConnectionId: string | null;
+  /** Free-text operator reference. NOT an identifier mapping. */
+  readonly externalRef: string | null;
+  readonly status: InventoryLocationStatus;
+  /** ISO-3166-1 alpha-2. */
+  readonly countryIso2: string | null;
+  readonly postcode: string | null;
+  readonly latitude: number | null;
+  readonly longitude: number | null;
+  /** ISO-8601 */
+  readonly createdAt: string;
+  /** ISO-8601 */
+  readonly updatedAt: string;
+}
 
 export interface InventoryLocationSummary {
   readonly id: string;
@@ -14,12 +53,64 @@ export interface InventoryLocationSummary {
   readonly status: string;
 }
 
-/** The paged listing, of which only `total` is read by the readiness surface. */
+/** All optional and AND-combined; an absent field does not filter. */
+export interface InventoryLocationFilters {
+  kind?: InventoryLocationKind;
+  /** Omitted lists every status, including `inactive` — soft retirement stays visible. */
+  status?: InventoryLocationStatus;
+  /** ISO-3166-1 alpha-2. Matched case-insensitively (uppercased before the request). */
+  countryIso2?: string;
+  /** Case-insensitive prefix match on `code`. */
+  codePrefix?: string;
+}
+
+/**
+ * 1-based `page`/`limit`, unlike `InventoryPagination`'s `limit`/`offset` —
+ * the divergence mirrors `ListLocationsQueryDto`'s own docblock: this is the
+ * core `InventoryLocationPagination` contract, echoed verbatim.
+ */
+export interface InventoryLocationListPagination {
+  page?: number;
+  limit?: number;
+}
+
+/** The paged listing. Widened from summaries to full rows for #3064's `listLocations`. */
 export interface PaginatedInventoryLocations {
-  readonly items: readonly InventoryLocationSummary[];
+  readonly items: readonly InventoryLocation[];
   readonly total: number;
   readonly page: number;
   readonly limit: number;
+}
+
+/** Mirrors `CreateLocationDto` field-for-field (#3064). */
+export interface CreateInventoryLocationInput {
+  code: string;
+  name: string;
+  kind: InventoryLocationKind;
+  ownerConnectionId?: string | null;
+  externalRef?: string | null;
+  status?: InventoryLocationStatus;
+  countryIso2?: string | null;
+  postcode?: string | null;
+  latitude?: number | null;
+  longitude?: number | null;
+}
+
+/**
+ * Mirrors `UpdateLocationDto` field-for-field (#3064). Every field optional;
+ * an omitted field is left untouched, an explicit `null` clears a nullable
+ * column. `code` is deliberately absent — it is the row's natural key.
+ */
+export interface UpdateInventoryLocationInput {
+  name?: string;
+  kind?: InventoryLocationKind;
+  ownerConnectionId?: string | null;
+  externalRef?: string | null;
+  status?: InventoryLocationStatus;
+  countryIso2?: string | null;
+  postcode?: string | null;
+  latitude?: number | null;
+  longitude?: number | null;
 }
 
 /**
--- a/apps/web/src/features/inventory/api/inventory.api.test.ts
+++ b/apps/web/src/features/inventory/api/inventory.api.test.ts
@@ -1,0 +1,129 @@
+/**
+ * Inventory API Client tests (#3064)
+ *
+ * Covers the four routes newly added to `InventoryApi` — `listLocations`,
+ * `getLocation`, `createLocation`, `updateLocation`, `deleteLocation`. The
+ * pre-existing `listActiveLocations` / `bootstrapLocations` probes are
+ * untouched and not re-tested here.
+ */
+import { describe, it, expect, vi } from 'vitest';
+import { createInventoryApi } from './inventory.api';
+import type { InventoryLocation } from './inventory-locations.types';
+
+function location(overrides: Partial<InventoryLocation> = {}): InventoryLocation {
+  return {
+    id: 'ol_location_1',
+    code: 'WH1',
+    name: 'Main warehouse',
+    kind: 'warehouse',
+    ownerConnectionId: null,
+    externalRef: null,
+    status: 'active',
+    countryIso2: null,
+    postcode: null,
+    latitude: null,
+    longitude: null,
+    createdAt: '2026-01-01T00:00:00.000Z',
+    updatedAt: '2026-01-01T00:00:00.000Z',
+    ...overrides,
+  };
+}
+
+describe('createInventoryApi.listLocations', () => {
+  it('should send no query string when nothing is filtered or paged', async () => {
+    const request = vi.fn().mockResolvedValue({ items: [], total: 0, page: 1, limit: 25 });
+
+    await createInventoryApi(request).listLocations();
+
+    expect(request).toHaveBeenCalledWith('/inventory/locations');
+  });
+
+  it('should forward every declared filter and pagination field', async () => {
+    const request = vi.fn().mockResolvedValue({ items: [], total: 0, page: 2, limit: 10 });
+
+    await createInventoryApi(request).listLocations(
+      { kind: 'warehouse', status: 'active', countryIso2: 'pl', codePrefix: 'WH' },
+      { page: 2, limit: 10 },
+    );
+
+    const [path] = request.mock.calls[0] as [string];
+    const query = new URLSearchParams(path.split('?')[1]);
+    expect(query.get('kind')).toBe('warehouse');
+    expect(query.get('status')).toBe('active');
+    expect(query.get('codePrefix')).toBe('WH');
+    expect(query.get('page')).toBe('2');
+    expect(query.get('limit')).toBe('10');
+  });
+
+  it('should uppercase countryIso2 before sending it', async () => {
+    const request = vi.fn().mockResolvedValue({ items: [], total: 0, page: 1, limit: 25 });
+
+    await createInventoryApi(request).listLocations({ countryIso2: 'pl' });
+
+    const [path] = request.mock.calls[0] as [string];
+    expect(new URLSearchParams(path.split('?')[1]).get('countryIso2')).toBe('PL');
+  });
+});
+
+describe('createInventoryApi.getLocation', () => {
+  it('should request the location by id', async () => {
+    const request = vi.fn().mockResolvedValue(location());
+
+    await createInventoryApi(request).getLocation('ol_location_1');
+
+    expect(request).toHaveBeenCalledWith('/inventory/locations/ol_location_1');
+  });
+});
+
+describe('createInventoryApi.createLocation', () => {
+  it('should POST the input as the JSON body', async () => {
+    const request = vi.fn().mockResolvedValue(location());
+    const input = { code: 'WH1', name: 'Main warehouse', kind: 'warehouse' as const };
+
+    await createInventoryApi(request).createLocation(input);
+
+    expect(request).toHaveBeenCalledWith('/inventory/locations', {
+      method: 'POST',
+      body: JSON.stringify(input),
+    });
+  });
+});
+
+describe('createInventoryApi.updateLocation', () => {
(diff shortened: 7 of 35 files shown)


piotrswierzy (commented):
## Tech Lead review

Strong slice, and a genuinely large one done carefully. The composition follows the `connections-list-page.tsx` cockpit shape with the `users-page.tsx` inline-dialog precedent; state ownership is exactly per `docs/frontend-architecture.md` (filters/page in URL, form state in RHF+Zod, server state in TanStack Query, no raw `fetch`); every mutation invalidates the `locationsAll()` prefix rather than a narrower key, and the bootstrap hook's widening to that prefix is correctly motivated. The `inventory-locations:write` permission is the right call — I verified all three write routes on `InventoryLocationsController` really are `@Roles('admin')` while `inventory:write` is held by `operator` too, so reusing the latter would have shown an operator a button that 403s — and the FE/BE `PermissionValues` mirror is covered by `check-permission-mirror.mjs`. The mockup is committed with both a state switcher and a role switcher, satisfying § UX Mockups. I also checked two things that look like findings and are not: `font-size: 0.8125rem` (164 existing occurrences, zero font-size tokens) and fixed `px` button widths (62 existing) are both the established baseline in `index.css`. My findings are about places where a comment or a test asserts something the shipped code does not support.

### Findings

**[IMPORTANT]** — `apps/web/src/pages/inventory/inventory-locations-page.tsx:~430`
> A failed bootstrap is silent. The zero-state's primary CTA calls `bootstrapMutation.mutate()` with no `onError`, and `useBootstrapLocationsMutation` has only an `onSuccess` — `mutate` (unlike `mutateAsync`) swallows the rejection, so the operator clicks "Create first location", nothing happens, and nothing says why. This is the first thing a new operator does on this page. The existing consumer of the very same hook already handles it: `features/connections/components/router-readiness-panel.tsx:212` renders `{bootstrap.error ? <ErrorState title="Could not create the location" …/> : null}`, and calls `bootstrap.reset()` before `mutate()` (line 197) so a stale error can't linger. `docs/frontend-architecture.md` § Async UX Conventions requires "error state with actionable messaging" and "success feedback for mutations" on every API-driven screen, and every other mutation on this page (retire, reactivate, delete, save) does toast both ways. Mirror the readiness panel, or switch to `mutateAsync` in a try/catch with the same toast shape as `handleRetire`.

**[IMPORTANT]** — `apps/web/src/features/inventory/components/location-delete-dialog.tsx:~79` and `~121`
> Two comments assert a fact this PR itself invalidates, and they are load-bearing for a copy decision. Both read, in substance: *"No 'can be re-activated later' — nothing in the product exposes that yet (#3068 tech-review: `toUpdateInput` never sends `status` back to `'active'`, and the list renders no Reactivate action)."* In this branch `toUpdateInput` **does** send `status` (`location-dialog.schema.ts`, `toUpdateInput` returns `status: values.status`), and the list **does** render a Reactivate action (`inventory-locations-page.tsx`, `handleReactivate` → `patch: { status: 'active' }`). So the retire toast and the "Retire instead" alert both withhold the reassurance that retiring is reversible, on a premise that is no longer true — and that reassurance is precisely what makes an operator willing to take the non-destructive option the dialog is steering them toward. Update both comments and add the reversibility to the copy.

**[IMPORTANT]** — `apps/api/src/inventory/http/inventory-locations.controller.spec.ts:228`
> The lockstep test cannot fail on the drift it exists to catch. Its name is "should be granted to exactly the roles this controller `@Roles(admin)` on every write route" and its comment says "if this drifts from the controller's actual @Roles, either add a permission-based guard or revert the grant" — but the assertion is `expect(rolesWithPermission).toEqual(['admin'])`, a hardcoded literal that never reads the controller. Widen `@Roles('admin')` to `@Roles('admin', 'operator')` on `create`/`update`/`delete` and this test still passes, while the `role.types.ts` docblock's claim that "the roles holding this permission MUST stay identical to that controller's `@Roles` list (asserted in `inventory-locations.controller.spec.ts`)" quietly becomes false. The repo already has the mechanism: `apps/api/src/auth/route-authorization-coverage.spec.ts` reads `Reflect.getMetadata(ROLES_KEY, fn)` per handler (lines 160/172). Derive the expected set from the controller's own write-route metadata and compare against `ROLE_PERMISSIONS`, so the two sides are actually compared rather than both compared to a constant.

**[IMPORTANT]** — `docs/plans/mockups/inventory-locations-3063.html:~613`
> The committed mockup shows a per-row position count and a link ("`${l.positionCount} position${…}`", `onclick="… alert('Would open /inventory filtered to this location.')"`), and the shipped page renders neither. I checked: no `positionCount` exists on the list read anywhere — not on the FE `InventoryLocation` type, not on the backend list DTO; the only place it exists is inside `LocationInUseError`. `LocationDeleteDialog`'s own docblock says so explicitly ("there is no cheap way to know this ahead of the attempt — the list read carries no such signal today"), which is exactly why the delete flow has to start in `confirm` and learn `in-use` from the 409. `docs/frontend-architecture.md` § UX Mockups: "Keep the mockup in sync with implementation decisions made after the mockup was drawn (a `/tech-review` finding, a scope cut); a stale mockup is worse than none, because it actively misleads" — and it is reviewer-enforced only, nothing under `check:invariants` can see it. Either remove the column from the mockup, or annotate it as deferred with the issue that would add the backend field.

**[SUGGESTION]** — `apps/web/src/features/inventory/components/location-delete-dialog.tsx:~132`
> The in-use state says "Stock positions still point here" while the 409 it just received carries the exact number — `LocationInUseError` is constructed with `positionCount` and its message reads "… is referenced by N inventory position(s)". Surfacing N turns a vague refusal into something the operator can act on (is it one stray position or the whole catalogue?), and costs nothing since the error is already in hand.

**[SUGGESTION]** — `apps/web/src/features/inventory/components/inventory-locations-tile.tsx:9`
> The docblock justifies not admin-gating the tile by saying "the sidebar nav entry is already gated on `inventory:read`, not on being an admin. Gating this tile to admin-only would make it LESS reachable than the nav item that already links to the same page." There is no nav item — `nav-registry.ts` is untouched in this PR, and the PR body records that #3069's nav entry was refined into a Settings tile instead. The *conclusion* is right (the reads are `@Roles('admin','operator','viewer')`, so a viewer-visible tile is correct), but the stated reason points at something that does not exist; rest it on the controller's read roles directly.

**[SUGGESTION]** — `apps/web/src/app/routes/inventory-locations.route.tsx:6`
> Crumb group is `'Platform'` while the page's own `backTo` is `/settings` and its only entry point is a Settings tile. Groups are genuinely mixed across settings-reached pages in this repo (`Administration`, `Settings`, `AI`, `Platform`), so this is not wrong — but the nearest analogue, `operational-settings.route.tsx` (also a Settings tile, also sync/inventory-adjacent), uses `{ group: 'Settings', title: 'Sync pacing' }`.

**[SUGGESTION]** — `apps/web/src/features/inventory/components/location-dialog.schema.ts:31`
> `latLngField` is the latitude field (±90) and its sibling is `lngField` — the shared-looking name reads as if both branches use it. `latField` / `lngField` would make the ±90 vs ±180 pairing obvious at the call site.

**Positive observations.** `pageOutOfRange` kept distinct from `filtersActive` is the subtle one most implementations get wrong — rendering the pre-bootstrap "nothing exists" empty state on a hand-edited `?page=99` would tell the operator something false about their own data, and the comment says exactly that. Declining `sortable` on every column with the reason written down (client sort would reorder only the visible page and look globally sorted while lying) is the right call and well argued. The delete dialog's reset effect keyed on `locationId` rather than the `location` object identity, with the reasoning about a future caller deriving it off refetched query data, is genuine defensive thinking. And `isUnmappedApiError` extracted to `shared/api` rather than copied a third time is the correct response to the duplication finding that produced it.

### Verdict
🔄 **Approve with changes**

(Posting as a comment rather than an approval: there are IMPORTANT findings outstanding.)

Priority:
1. Surface bootstrap failure — the readiness panel already shows how, and this is the zero-state's primary CTA.
2. Fix the two stale comments in `location-delete-dialog.tsx` and add the reversibility reassurance to the retire copy.
3. Make the permission lockstep test read the controller's `@Roles` metadata instead of a hardcoded `['admin']`.
4. Reconcile the mockup's position-count column with what shipped.

---

piotrswierzy (commented):
## Tech Lead re-review (head unchanged at `c42da6c0`)

Re-verified every finding in code at the unchanged head. The baseline moved by 11 merged PRs and **none of them touches this PR's subject matter** — nothing on `main` went near `apps/web/src/features/inventory`, `apps/api/src/inventory`, `role.types.ts`, `nav-registry.ts`, or any permission mirror. Of this PR's 34 changed files exactly one also appears in the new `main` (`apps/web/src/index.css`), and the two edits are ~5,400 lines apart — main's is #3037 deleting the `.category-picker` block near line 8401, yours is a single `@@ -2997,6 +2997,54 @@` addition — consistent with `mergeable: MERGEABLE, mergeStateStatus: BEHIND`, i.e. a stale base rather than a conflict. No migration, and CI is 9/9 green. **Three of my four IMPORTANTs still stand. The fourth I am withdrawing — I was wrong, and I say how below.**

### Findings

**[IMPORTANT — STILL HOLDS]** — `apps/web/src/features/inventory/hooks/use-bootstrap-locations-mutation.ts:31` and `apps/web/src/pages/inventory/inventory-locations-page.tsx:426`
> A failed bootstrap is still silent. Re-verified: the hook is `useMutation({ mutationFn, onSuccess })` with **no `onError`**, and the zero-state CTA is `onClick={() => bootstrapMutation.mutate()}` — `mutate`, not `mutateAsync`, so the rejection is swallowed. `bootstrapMutation.error` is referenced nowhere in the page (the only occurrences of `bootstrap` are lines 41, 122, 425, 426). The operator clicks "Create first location", nothing happens, nothing says why — on the first action a new operator takes on this page. The remedy is already demonstrated twice over: in the *same file*, `handleReactivate` (137) and `handleRetire` (155) both use `mutateAsync` + `catch` + error toast; and on `main`, the existing consumer of this very hook does it properly — `router-readiness-panel.tsx:197` calls `bootstrap.reset()` before `mutate()` so a stale error can't linger, and `:212` renders `{bootstrap.error ? <ErrorState title="Could not create the location" … /> : null}`. `docs/frontend-architecture.md` § Async UX Conventions requires "error state with actionable messaging" on every API-driven screen.

**[IMPORTANT — STILL HOLDS, and it is a three-way contradiction, not a two-way one]** — `apps/web/src/features/inventory/components/location-delete-dialog.tsx:82` and `:114`
> Both stale comments are still there, and both clauses of their premise are still false in this same branch: `location-dialog.schema.ts:123` — `toUpdateInput` returns `status: values.status`; and `inventory-locations-page.tsx:316` renders a **Reactivate** button wired to `:137`'s `patch: { status: 'active' }`. So the retire toast and the "Retire instead" alert both withhold the reassurance that retiring is reversible, on a premise this PR itself invalidated — and that reassurance is exactly what makes an operator take the non-destructive option the dialog is steering them toward.
>
> Two things I did not say last time, both of which make this sharper:
> - **The page refutes the dialog in its own comments.** `inventory-locations-page.tsx:150`, directly above `handleRetire`: *"no confirm dialog, because retiring never deletes anything and **Reactivate undoes it in one click**."* One file in this PR states the fact; another states its negation and drops copy because of it.
> - **The committed mockup already has the copy the implementation strips.** `inventory-locations-3063.html:795`: *"Retire it instead — existing history keeps pointing at a row that exists, **and it can be re-activated later**."* So this is also a mockup/implementation divergence under § UX Mockups, not merely a stale comment. Restoring the sentence brings the code back in line with the spec rather than inventing new copy.

**[IMPORTANT — STILL HOLDS]** — `apps/api/src/inventory/http/inventory-locations.controller.spec.ts:238`
> The lockstep test still cannot fail on the drift it exists to catch. The assertion is `expect(rolesWithPermission).toEqual(['admin']);` — a hardcoded literal — and a grep for `Reflect` / `ROLES_KEY` / `getMetadata` across the entire spec returns **zero hits**, so the controller's `@Roles` is never read and both sides are compared to constants. Widen `@Roles('admin')` to `@Roles('admin', 'operator')` on `create`/`update`/`delete` and this test still passes, while `role.types.ts:86` — *"the roles holding this permission MUST stay identical to that controller's `@Roles` list (asserted in `inventory-locations.controller.spec.ts`)"* — quietly becomes an unbacked claim. The repo already has the mechanism: `route-authorization-coverage.spec.ts:189` and `:201` read `Reflect.getMetadata(ROLES_KEY, …)` per class and per handler. Derive the expected set from the controller's own write-route metadata and compare it against `ROLE_PERMISSIONS`.

**[WITHDRAWN — I was wrong; downgraded to the SUGGESTION below]** — ~~`docs/plans/mockups/inventory-locations-3063.html`, per-row position count~~
> I raised this as an IMPORTANT on the grounds that the mockup shows a `positionCount` the API does not have. The column is indeed unimplementable — `positionCount` is confirmed absent from the FE `InventoryLocation` type and from `LocationResponseDto` — **but the mockup is already annotated, and the annotation ships in this PR** (`inventory-locations-3063.html:560-568`, a `+` block in the diff). It states that `positionCount` "is drawn here as if the list read already exposed it, which today it does not", names `LocationResponseDto`/`toDto` and `countPositionsAtLocation` precisely, explains why the count is wanted ("to gauge blast radius before retiring"), and records the two options for supplying it. My own remedy was *"Either remove the column from the mockup, **or annotate it as deferred**"* — the second half is satisfied, with a better rationale than I asked for. I missed it; apologies for the noise. Nothing is owed here beyond the small residual below.

**[SUGGESTION — new, the residual of the withdrawn item]** — `docs/plans/mockups/inventory-locations-3063.html:567`
> The annotation closes with *"a backend-contract decision **#3063's implementation** needs to make … flagging it here so it isn't lost."* #3063 is the issue this PR closes, and it closes without making that decision — so once it closes, "so it isn't lost" is guaranteed by nothing. Replace the owner with a follow-up issue number (the one that would add a positions-count field to the list read), so the deferral has somewhere to land.

**[SUGGESTION — STILL HOLDS, and it now has a second reason]** — `apps/web/src/features/inventory/components/location-delete-dialog.tsx:109`
> The in-use state still says "Stock positions still point here" while the 409 it just received carries the exact number (`LocationInUseError` is constructed with `positionCount`). Surfacing N turns a vague refusal into something actionable, and costs nothing since the error is in hand. Worth noting this is the same need the mockup annotation argues for at length — and the mockup's own 409 copy at `:795` already renders it (`${l.positionCount} inventory position${…} still point here`). So the one place the count is genuinely available today is the one place it is dropped.

**[SUGGESTION — STILL HOLDS]** — `apps/web/src/features/inventory/components/inventory-locations-tile.tsx:10`
> The docblock still justifies not admin-gating the tile by citing "the sidebar nav entry … gated on `inventory:read`" and "the nav item that already links to the same page". `nav-registry.ts` is not among this PR's 34 changed files and has no inventory entry on `main`. The *conclusion* is right (the reads are `@Roles('admin','operator','viewer')`, so a viewer-visible tile is correct) — rest it on the controller's read roles directly rather than on a nav item that does not exist.

**[SUGGESTION — STILL HOLDS]** — `apps/web/src/app/routes/inventory-locations.route.tsx:6`
> Crumb group is still `'Platform'` while the page's `backTo` is `/settings` and its only entry point is a Settings tile. Groups are genuinely mixed across settings-reached pages here, so this is not wrong — but the nearest analogue, `operational-settings.route.tsx:14` (also a Settings tile, also inventory/sync-adjacent), uses `{ group: 'Settings', … }`.

**[SUGGESTION — STILL HOLDS]** — `apps/web/src/features/inventory/components/location-dialog.schema.ts:30`
> `latLngField` is the latitude field (±90, used at `:69`) while its sibling is `lngField` (±180, `:42`/`:70`). The shared-looking name reads as if both branches use it; `latField` / `lngField` would make the pairing obvious at the call site.

**Positive observations (re-confirmed).** `pageOutOfRange` kept distinct from `filtersActive`, the declined `sortable` with its reasoning written down, the delete dialog's reset effect keyed on `locationId` rather than object identity, and `isUnmappedApiError` extracted to `shared/api` rather than copied a third time — all unchanged and all still the right calls. And I'll add one: the mockup annotation I missed is a good example of § UX Mockups done properly — it keeps a drawn affordance visible as a deliberate deferral instead of quietly deleting it, which is more useful to the next reader than either removing the column or leaving it unexplained.

### Verdict
🔄 **Approve with changes** — verdict unchanged; **3 IMPORTANT + 5 SUGGESTION** (down from 4 + 4, one IMPORTANT withdrawn).

Priority:
1. Surface bootstrap failure — the readiness panel on `main` already shows exactly how, and this is the zero-state's primary CTA.
2. Fix the two stale comments in `location-delete-dialog.tsx` and restore the reversibility sentence; the mockup at `:795` and the page's own comment at `:150` both already have it.
3. Make the permission lockstep test read the controller's `@Roles` metadata instead of a hardcoded `['admin']`.
4. ~~Reconcile the mockup's position-count column~~ — withdrawn, already annotated; only the follow-up issue number is worth adding.

Merge `main` when convenient (`BEHIND`, no conflict).

---

piotrswierzy (commented):
## Tech Lead re-verification (head unchanged at `c42da6c`)

Head is unchanged, `main` is unchanged (`e336cf4`), and all nine CI checks pass — so this is a targeted re-check rather than a repeat of the full review. The PR is `MERGEABLE` / `BEHIND` (a fast-forward of the base, no conflicts). All three standing IMPORTANT findings re-verified against the current head and still hold; my earlier withdrawal stands. One cross-PR interaction has appeared that this PR's merge order now matters for.

### Findings

**[IMPORTANT]** — `apps/web/src/pages/inventory/inventory-locations-page.tsx:425-426` — a failed bootstrap is silent.

Re-confirmed: the zero-state's primary CTA is

```tsx
disabled={write.demoReadOnly || bootstrapMutation.isPending}
onClick={() => bootstrapMutation.mutate()}
```

and `onError` does not appear anywhere in the file. `mutate` (unlike `mutateAsync`) swallows the rejection, so a failed bootstrap leaves the operator on an unchanged empty state with no toast, no banner, and no indication anything was attempted — on the one action the whole zero-state exists to offer.

The comparison that makes this unambiguous is in-repo: `router-readiness-panel.tsx:211-213`, the *existing* consumer of this same `useBootstrapLocationsMutation` hook, already renders

```tsx
{bootstrap.error ? <ErrorState title="Could not create the location" message={bootstrap.error.message} /> : null}
```

Two consumers of one hook should not disagree about whether its failure is worth telling the operator about. Every other mutation on this page already uses `mutateAsync` + try/catch + an error toast (`handleRetire`, `handleReactivate`) — bootstrap is the odd one out.

**[IMPORTANT]** — `apps/web/src/pages/inventory/inventory-locations-page.tsx:148-152` vs `:155-160` vs `docs/plans/mockups/inventory-locations-3063.html:795` — a three-way contradiction, now fully traced.

The code comment justifies the absence of a confirm dialog on Retire with:

> retiring never deletes anything and **Reactivate undoes it in one click**

The retire toast the operator actually sees says only:

> `"{name}" retired` — *Existing positions keep pointing at it.*

And the mockup carries the fuller sentence the implementation drops:

> Retire it instead — existing history keeps pointing at a row that exists, **and it can be re-activated later**.

So the reversibility is the stated reason there is no confirm step, the mockup says it out loud, and the operator is the one party never told. Reactivate *is* implemented (`:313-316`), which makes this purely a copy gap rather than a behaviour one — and a cheap one to close: append the mockup's clause to the retire toast's description, and the comment, the mockup and the UI all agree.

Per `docs/frontend-architecture.md` § UX Mockups, an implementing PR must either touch the mockup or say why it is still accurate. Here the mockup is the accurate one and the implementation is the drift, so the fix belongs in the toast.

**[IMPORTANT]** — `apps/api/src/inventory/http/inventory-locations.controller.spec.ts:228-239` — the lockstep test cannot fail on the drift it promises to catch.

The test is named

> should be granted to exactly the roles this controller `@Roles(admin)` on every write route

and its docblock promises to fire "if this drifts from **the controller's actual @Roles**". The assertion is:

```ts
const rolesWithPermission = UserRoleValues.filter((r) => ROLE_PERMISSIONS[r].includes('inventory-locations:write'));
expect(rolesWithPermission).toEqual(['admin']);
```

That reads `ROLE_PERMISSIONS` and a hardcoded literal. It never reads the controller's decorator metadata, so changing the controller to `@Roles('admin','operator')` leaves this green — i.e. it catches drift on the *grant* side only, which is one half of the lockstep it claims to be. Read `Reflect.getMetadata(ROLES_KEY, ...)` off each write handler and compare the two sets; the repo already does exactly this in `apps/api/src/auth/route-authorization-coverage.spec.ts`, which discovers controllers rather than listing them for precisely this reason.

This is the "check that cannot fail" shape — a test asserting a claim strictly weaker than the one in its own name is worse than no test, because it retires the reviewer's attention.

*(Withdrawn last round and still withdrawn: the mockup's `positionCount` column is annotated as a deferral, and the annotation ships in this PR.)*

### Cross-PR interaction

This PR ships **Retire** and **Delete**, which promotes two items on #3208 from "reachable only by a deliberate API call" to "reachable in one click":

- #3208's `validateStockLocationOverride` accepts an **inactive** location, so a location retired via `handleRetire` here can still be named as a `stockLocationOverride` while the fulfilment router filters it out.
- `LocationService.deleteLocation` refuses only on `countPositionsAtLocation(id) > 0` — `inventory_items` rows. A `stockLocationOverride` in `connection.config` is invisible to that guard, so Delete here removes a location an override still names, leaving a dangling reference nothing re-checks.

Nothing to change in this PR on its own account — but whichever of the two lands second owns the reconciliation, and I have recorded the same note on #3208.

### Verdict

🔄 **Approve with changes**

1. Surface the bootstrap failure — `mutateAsync` + error toast, or an `ErrorState` matching `router-readiness-panel.tsx`.
2. Put the reversibility clause in the retire toast so the comment, the mockup and the UI agree.
3. Make the lockstep test read the controller's `@Roles` metadata.

---

norbert-kulus-blockydevs:
## Review addressed — commit `f810c3244`

- **[IMPORTANT] A failed bootstrap is silent** — *fixed*. The zero-state CTA now goes through `handleBootstrap` (`mutateAsync` + try/catch + an error toast carrying the server's own message), matching the page's own `handleRetire`/`handleReactivate` idiom and agreeing with the other consumer of the same hook. **No success toast, deliberately**: success replaces the zero state with the created row, and a concurrent re-run reports its code in `existingCodes` having created nothing (#2407) — so a "created" toast would be a positive claim the surface cannot support. Regression test added asserting both the toast title and the 403 message.
- **[IMPORTANT] Retire copy vs comment vs mockup** — *fixed, with the mockup as the accurate side*, so the change is in the code and `inventory-locations-3063.html` is deliberately untouched (stating that here is what § UX Mockups asks for). The reversibility clause is appended to the row-action retire toast, the delete dialog's retire toast and the dialog's "Retire instead" alert. Both stale `#3068` notes claiming the capability does not exist are **rewritten rather than left to mislead** — that note was correct when written and expired the moment `#3066` shipped `handleReactivate`.
- **[IMPORTANT] The lockstep test cannot fail** — *fixed*. Both halves are now read from source: the grant from `ROLE_PERMISSIONS`, the requirement from each write handler's own `@Roles` metadata via `Reflect.getMetadata` (the `route-authorization-coverage.spec.ts` technique). A discovery guard asserts the four write handlers by name, so an empty or shrunken set fails instead of passing vacuously; the admin-only literal survives as its own honestly-named `it`.

  **Verified against six deliberate breakages** with a dependency-free harness replicating the assertion body (no `node_modules` in the worktree, so jest was unavailable): as-shipped → PASS; `@Roles('admin','operator')` on `update` → FAIL (the exact drift you named); `ROLE_PERMISSIONS` grants `operator` → FAIL; `remove` loses `@Roles` → FAIL; discovery returns nothing → FAIL on the guard; a new unguarded write route → FAIL on the guard. **The old test was green for four of those six.** The controller header's stale "reads carry no `@Roles`" claim — contradicted by its own code and by a later paragraph in the same header — is corrected too.
- **[cross-PR] #3208 overrides vs Retire/Delete** — *no change here*, per your own conclusion that whichever PR lands second owns the reconciliation. #3208's side is now arithmetically safe (see its thread); the dangling-override question is recorded there as `needs-owner-decision`.

`main` has been merged into this branch in the same round (it was 17 commits behind; no conflict).

---

piotrswierzy (approved):
## Tech Lead re-review (`c42da6c` → `b9d9724c`)

All three standing IMPORTANTs are closed, each at the right layer and each with a test that would have gone red before the fix. The lockstep one is closed more thoroughly than I asked — it discovered a fourth write route the controller's own docblock had miscounted. `main` is merged in; the PR is `MERGEABLE` (`BLOCKED` is only the pending required checks). No new findings.

### Prior findings — all closed

**1. Silent bootstrap failure — CLOSED.** `handleBootstrap` now uses `mutateAsync` inside try/catch with an error toast, and the CTA calls it. Two decisions inside it are worth calling out because both are the less obvious option:

- **The server's own message is carried through**, not flattened to "Try again", and the comment says why — the reachable failures here are distinguishable and actionable (a non-admin session answers 403, #2316). The regression test asserts `Insufficient permissions` is actually rendered, not just that *something* appeared.
- **No success toast**, deliberately, with the reasoning in place: success replaces the zero state with the created row, and a concurrent re-run reports its code in `existingCodes` having created nothing (#2407) — so announcing a creation would be a positive claim this surface cannot support. That is the right reading of the bootstrap's insert-then-recover semantics, and I would not have thought to object if you had added one.

The new test's docblock names the exact failure shape ("indistinguishable from never having clicked"), which is what makes it a guard rather than a coverage line.

**2. The three-way copy contradiction — CLOSED, in both places.** The retire toast now reads *"Existing positions keep pointing at it, and it can be reactivated later."* on **both** surfaces — the page's `handleRetire` and `location-delete-dialog.tsx` — and the in-use Alert carries the clause too. The comments do the thing that makes this durable rather than a patch: instead of deleting #3068's old note, they record that it was *correct at the time* and expired the moment `handleReactivate` shipped. A future reader who greps for why the clause was once removed finds the answer beside the clause. The mockup is left untouched, which is right — it was the accurate party and the implementation was the drift, so `docs/frontend-architecture.md` § UX Mockups is satisfied by fixing the code.

**3. The lockstep test that could not fail — CLOSED, and it found something.** `discoverWriteRoutes()` now reads `Reflect.getMetadata(METHOD_METADATA, fn)` and `Reflect.getMetadata(ROLES_KEY, fn)` off the controller prototype and compares the decorator set against `ROLE_PERMISSIONS`, the `route-authorization-coverage.spec.ts` technique. Four properties I checked specifically:

- **The discovery guard is a name list, not a count** — `['bootstrap', 'create', 'remove', 'update']` — so a renamed or added write route is a visible diff line rather than a silently-still-passing `toHaveLength(4)`. This is the guard that stops the new assertion from becoming vacuous the way its predecessor was.
- **`RequestMethod.GET` is `0`**, and the code tests `verb === undefined` rather than falsiness. Getting that wrong would have let every read route into the set.
- **Roles are sorted on both sides**, so the comparison does not depend on decorator argument order.
- **The agreement test and the anchor test are separate.** The lockstep proves the two halves agree; `should keep every write route admin-only` says *which* value they must agree on. Without the second one, flipping both halves to `['admin','operator']` would pass — which is the same class of hole as the original.

And it surfaced a real drift: the controller docblock said "the three write routes", but `bootstrap` is a fourth. The docblock is corrected in the same commit, and the reads' `@Roles('admin','operator','viewer')` is now documented rather than described as "no `@Roles`".

*(Still withdrawn: the mockup's `positionCount` column is annotated as a deferral.)*

### Findings

Spot-checked as part of the delta: every `className` and `var(--*)` referenced by `inventory-locations-page.tsx` and `location-delete-dialog.tsx` resolves against `apps/web/src/index.css` at this head — the delta is copy and control-flow only and introduces no new selector, so the one-directional `check-design-tokens.mjs` gap is not load-bearing here.

### Cross-PR interaction (unchanged, and still owed)

Re-verified at both heads and on `main`: `LocationService.deleteLocation:89-102` still refuses only on `countPositionsAtLocation(id) > 0`, i.e. `inventory_items` rows. A `stockLocationOverride` living in `connection.config` is invisible to that guard, so **Delete** on this page removes a location #3208's override still names, leaving a dangling reference nothing re-checks.

#3208 has since landed the *inactive* half of this (it now refuses a retired location at override-write time, on the transition), and explicitly documents the delete half as deferred. Nothing in either PR closes it, and no issue tracks it — I have asked on #3208 for the tracking issue, since that is where the docblock claims it exists. Nothing to change here on this PR's own account.

### Verdict

✅ **Approve**

One note on state: CI is **not green yet** — `Type Check`, `Docker Build Smoke Test`, `PHP Unit Tests` and `Scaffolded Adapter Builds` pass; `Lint`, `Test`, `Build` and `Integration Tests` are pending at the time of writing. Please confirm they go green rather than reading this approval as covering them.

---

Issue #3188, opened 2026-09-11
feat(invoicing): buyer tax ID column on the invoice list


norbert-kulus-blockydevs:
## Merge strategy — ✅ DIRECT TO `main`

One branch off `main`, one PR, one merge. This slice leaves `main` observably better on its own — no dead code, no half-rendered feature. If it has a dependency it is named at the bottom of this issue; wait for that to land on `main` first, then branch normally.

---

Part of #3173.

## Context
The list already filters on tax identity and shows **no column for it**, so the filter answers a yes/no question and hides the answer. More expensive than it looks: the projection stores only a boolean `hasBuyerTaxId`, which is what serves the filter.

## Scope
New projection column, migration, write path, and the column itself — rendering the same three states as the order detail.

## Acceptance
- [ ] Filtering to `With tax ID` shows which id each row carries
- [ ] Same three renderings and vocabulary as the order detail
- [ ] Two `docs/lessons.md` entries added (rider below)

**Depends on**: **#3224** - the receipt work (#3187) threaded the buyer tax number into the fiscal-receipt branch only, so an auto-issued INVOICE still carries none and this column would be permanently empty on every auto-issued row. #3224 closes that. Original wording kept below because it is exactly right about the cause: the receipt work — until numbers are actually sent the column is always empty. **Spec**: `docs/plans/mockups/` (committed by #3174).

---

## Rider: two `docs/lessons.md` entries (epic #3173 debt)

Folded into this issue rather than raised separately. Both were discovered while
running the epic #3173 stack and both are otherwise going to be rediscovered.

**1. A core barrel or tokens edit escalates the pre-commit hook to the FULL
suite.** `scripts/smart-test.mjs` classifies `libs/core/src/<ctx>/index.ts` and
`libs/core/src/<ctx>/<ctx>.tokens.ts` as wide triggers (`CORE_BARREL_RE`,
`CORE_TOKENS_RE`), and the wide branch runs `pnpm test` = `pnpm -r test` at
pnpm's DEFAULT `workspace-concurrency` of 4. `test:ci` was bounded to 2 back in
#976 for exactly this OOM, but the hook's path was never bounded. On a
memory-constrained machine the correct local remedy is to bound the fan-out for
that one commit:

```
npm_config_workspace_concurrency=1 git commit -s -F <message-file>
```

**not** `--no-verify`, which skips lint and type-check as well and leaves the
commit verified by nothing. The entry should point at
[...]

PR #3201, opened 2026-09-11
fix(sales-documents): normalise market country codes

### LINKED ISSUE #3176: fix(sales-documents): normalise market country codes
## Merge strategy — 🔗 STACKED, on top of #3179

**Not direct to `main`.** This slice ships inside the #3173 stack (#3174 → #3175 → #3179 → #3176 → #3177 → #3178 → #3180 → #3181 → #3182 → #3183 → #3184 → #3185 → #3186): PR #3201 targets #3179's branch, not the default branch. One consequence is worth stating plainly — GitHub honours `Closes #3176` only on a merge into `main`, so this issue does **not** auto-close when its own PR merges; it closes when the chain reaches `main`, or by hand once it has.

---

Part of #3173.

## Context
The running instance shows **two market cards for one country** — `PL` with 16 orders and 3 rules, and `pl` with 1 order and 0 rules. The country code is compared case-sensitively somewhere between ingestion and market discovery, so an order can land in a market that has no rules and is **silently held** while the operator looks at a fully configured card.

## Scope
- Decide and implement normalisation **on write** or **on read**, and record which in the PR
- Give the stranded lowercase market a disposition — stopping new ones is not enough

## Acceptance
- [ ] One country yields exactly one market card
- [ ] The existing `pl` market's order is accounted for, not orphaned

**Depends on**: #3179 (the slice below it in the stack).

norbert-kulus-blockydevs:
Closes #3176
Part of #3173

Stacked onto #3200 (ADR-072 + aneks) per the epic's PR-stack strategy — #3199/#3200 aren't getting an approve right now, so the rest of the epic stacks on them via native GitHub PR-stack rather than waiting.

## What

An order whose delivery-address country arrives lowercase, or a rule authored with a lowercase country scope, produced a **second market card for one country** — the live instance showed `PL` (16 orders, 3 rules) and `pl` (1 order, 0 rules) as two unrelated markets, because every country comparison in `sales-documents` was a plain case-sensitive string match.

**Decision recorded (per the issue's own ask): normalise at the read/write choke point, not in every producer.**

- `SalesDocumentRulesService` upper-cases (trim + `toUpperCase`, mirroring `LocationService.normaliseCountry`) — on **write** for `createRule` / `upsertCountryDefault` / `acknowledgeNoDocument` / `clearAcknowledgment`, and on **read** for `resolveRouting` / `resolveRoutingBatch`.
- The read-side choke point matters because *both* order-facts producers already converge there — `toSalesDocumentOrderFacts` (invoicing, the live auto-issue path) and `toOrderFacts` (orders' sales-document-view service, the `/orders` projection). Neither mapper needed touching. **Correction (review round):** normalising the read side alone did *not* close the identical case-mismatch an `orderCountry` rule *condition* hits — it inverted it, since `order.country` became reliably uppercase while a condition authored as `'pl'` was never folded, making it permanently unmatchable. That is closed separately in `42b0c835d`: `createRule` now folds each `orderCountry` condition value before `computeSalesDocumentConditionsHash`, and `evaluateSalesDocumentRules` compares the two codes case-insensitively so rows already stored in stray case keep matching without a jsonb rewrite that SQL cannot pair with a recomputed `conditions_hash`.
- The market-count aggregation (`OrderRecordRepository.countOrdersByRoutingCountrySince`) reads `orderSnapshot` directly, outside that service, so its own `ROUTING_COUNTRY_EXPR` SQL constant is separately wrapped in `upper(...)`. **This is what accounts for the stranded `pl` order with no data migration** — `order_records.orderSnapshot` is never rewritten (ADR-014: verbatim copy of what the source reported), so the fix is read-side folding, not a rewrite.
- A defensive migration (`1878000000000-normalise-sales-document-country-codes.ts`) upper-cases any pre-existing stray-case row in the three operator-authored tables (`sales_document_rules`, `sales_document_country_defaults`, `sales_document_country_acknowledgments`). No live environment audited during this issue had such a row (the observed duplicate market carried zero rules/defaults under either casing) — the migration is defence-in-depth, and it **skips rather than merges/deletes** a row that would collide with an existing uppercase sibling, since it has no way to know which of two same-scope rows an operator would want to keep.

## Acceptance

- [x] One country yields exactly one market card (verified: the SQL fix folds the order-count aggregation; the service fix keeps rule/default lookups comparable going forward)
- [x] The existing `pl` market's order is accounted for, not orphaned — folded into `PL`'s card via the read-side `upper()`, no migration needed for order data

## Verification

Standalone invariant gates run clean on this branch (no `node_modules` in this worktree — CI runs the full suite):

```
check-architecture-gates          PASS
check-cross-context-imports       PASS
check-migration-timestamps        PASS
check-sales-document-market-summary-mirror   PASS
check-sales-document-reason-mirror           PASS
check-repo-urls                   PASS
```

Added unit-test coverage in `sales-document-rules.service.spec.ts` for: write-path normalisation (rule create, country-default upsert, acknowledgment), and read-path normalisation for both `resolveRouting` and `resolveRoutingBatch` (a lowercase order country resolving against an uppercase rule scope).

Per project convention, local `pnpm test`/`pnpm lint`/`pnpm type-check` were not run (no local `node_modules`) — CI covers these.


piotrswierzy (commented):
## Tech Lead review

Folds sales-document country codes to uppercase at one read/write choke point in `SalesDocumentRulesService`, wraps the market-count aggregate's `ROUTING_COUNTRY_EXPR` in `upper(...)`, and adds a defensive migration for the three operator-authored tables. The central decision — read-side folding for order data (`orderSnapshot` stays a verbatim ADR-014 copy) and a rewrite only for configuration — is right and clearly argued, and the migration's timestamp (`1878000000000`) correctly sorts *before* #3210's `1878000000001`, which matters: this migration's `sales_document_country_defaults` collision guard reads `(country, document_kind)`, which is still the live uniqueness at the moment it runs. Two problems, one of which can fail `up()` outright.

### Findings

**[IMPORTANT]** — `apps/api/src/migrations/1878000000000-normalise-sales-document-country-codes.ts:57,66,77`
> The docblock claims *"SAFE BY CONSTRUCTION, NOT BY ASSUMPTION"*, but each `NOT EXISTS` only tests for collision against a row **already uppercase**. Two rows that are both non-uppercase and fold onto the same value (`pl` and `Pl`; `pL` and `pl`) are each invisible to the other's guard — neither matches `a2."country" = UPPER(a."country")` at the time the predicate is evaluated, because `UPDATE` evaluates its `WHERE` against the pre-statement snapshot. Both rows therefore pass, both are updated in one statement, and the statement violates `PK_sales_document_country_acknowledgments` / `UQ_sales_document_country_defaults_country_kind` / `UQ_sales_document_rules_country_hash_from`. `up()` aborts and the deploy stops — which is the *loud* failure mode, so not a data hazard, but it is exactly the case the guard was written to prevent. Rank among the folding set rather than probing for a pre-existing sibling: `ROW_NUMBER() OVER (PARTITION BY UPPER("country"), <rest of the unique key> ORDER BY …)` and update only `rn = 1`. That is the same window-function shape #3210's sibling migration one file over already uses.

**[IMPORTANT]** — `libs/core/src/sales-documents/domain/domain-services/evaluate-sales-document-rules.ts:119` (consequence of `sales-document-rules.service.ts:createRule`)
> The PR description says *"as a side effect this also closes the identical case-mismatch an `orderCountry` rule **condition** would otherwise hit."* It does not — it inverts it. `createRule` normalises `input.country` (the rule's scope) but never the `orderCountry` condition's own `stringValue`, and the evaluator compares `order.country === condition.value` strictly. Before this change, a rule authored with condition `orderCountry eq 'pl'` at least matched an order whose source reported `pl`; after it, `order.country` is *always* uppercase, so that condition can never match anything, permanently and silently. Normalise `orderCountry` condition values in `createRule` alongside the scope — note the ordering constraint: it must happen **before** `computeSalesDocumentConditionsHash`, since the hash is part of `UQ_sales_document_rules_country_hash_from` — and fold stored condition values in the migration. At minimum, correct the claim in the PR body.

**[IMPORTANT]** — `libs/core/src/sales-documents/application/services/sales-document-rules.service.ts:218` (`listConfiguredCountries`)
> The acceptance criterion is *"One country yields exactly one market card"*, and the migration deliberately **skips** a colliding stray-case row rather than merging it — a reasonable call. But nothing then folds on read: `listConfiguredCountries` groups by the raw stored `country` from `countRulesByCountry` / `findAll` / `findAll`, and `SalesDocumentMarketsController.listMarkets` unions those keys with the (now-uppercased) detected markets. So a surviving stray row still renders a second card for one country — the exact defect the issue reports, in the one path the migration knowingly leaves alone. The PR's own principle (read-side folding for what it will not rewrite) applies here and is not applied. Upper-casing the three grouping keys in `listConfiguredCountries` closes it with no data change and no schema change.

**[SUGGESTION]** — same migration
> A skipped row is skipped silently: no `RAISE NOTICE`, no count. The docblock says it is left *"for a human to resolve"*, but nothing tells the human it exists. One `RAISE NOTICE` per skipped row (or a count per table) gives the migration log the breadcrumb that instruction depends on.

**[SUGGESTION]** — `sales-document-rules.service.ts:88` (`normaliseCountry`)
> Third copy of this two-line rule in the tree (`LocationService.normaliseCountry` is cited; `deriveSalesDocumentRows`' FE side has its own reading of country identity). Not worth a `check-*-mirror.mjs` at this size, but if a fourth appears it is.

Done correctly and worth naming: the migration timestamp ordering against #3210's sibling (getting it the other way round would have made the `(country, document_kind)` collision guard test a uniqueness that no longer exists); the explicit ADR-014 refusal to rewrite `orderSnapshot`, with the read-side `upper()` doing the work instead; `'*'` needing no special case because `toUpperCase()` is identity on it, stated rather than left to be rediscovered; and the check that `ROUTING_COUNTRY_EXPR` has exactly one consumer and no expression index behind it, so wrapping it costs nothing at the plan level.

### Verdict
🔄 **Approve with changes**

1. Fix the migration guard so two non-uppercase rows folding onto each other cannot abort `up()`.
2. Normalise `orderCountry` **condition values**, or correct the PR body's claim — as it stands this makes a lowercase-authored condition permanently unmatchable.
3. Fold on read in `listConfiguredCountries`, so the skipped-row path cannot reproduce the duplicate market card.

---

piotrswierzy (commented):
## Tech Lead re-review (`8a29af7c` → `4a1f3332`)

All three prior findings are closed, and closed in the way that leaves the *reasoning* behind, not just the fix. The migration's collision guard is now a `ROW_NUMBER() OVER (PARTITION BY UPPER("country"), <rest of the unique key>)` ranking per table, with the ordering `(country = UPPER(country)) DESC` first — so an already-uppercase row always occupies rank 1 and a stray-case sibling is skipped rather than displacing it, and two stray-case rows folding onto each other (`pl` + `Pl`) no longer both pass and abort `up()`. The `orderCountry` inversion is fixed at comparison time (`order.country.toUpperCase() === condition.value.toUpperCase()`) **and** at write time (`normaliseConditionCountries`, placed before `computeSalesDocumentConditionsHash` with the ordering constraint stated at the call site) — the comparison-time half is what keeps an already-stored lowercase condition matching, which a write-path fix alone would not have done. `listConfiguredCountries` now folds all three grouping keys, so a skipped row cannot render a second market card. Both suggestions landed too: a per-table `RAISE NOTICE` naming the count and the surviving casings, and the four sibling copies of `normaliseCountry` enumerated with the reason they cannot share one definition (#591 / plugin-to-core).

The migration was renumbered `1878000000000` → `1881000000000` to clear `main`'s new tail (`1880000000000-create-price-change-episodes`), and #3210 followed to `1881000000001` — so the ordering this stack depends on still holds: #3201 runs while `(country, document_kind)` is still the live uniqueness its defaults partition tests, and only then does #3210 narrow it.

### Findings

**[SUGGESTION]** — `apps/api/src/migrations/1881000000000-normalise-sales-document-country-codes.ts`
> The renumber moved this onto `1881000000000`, which open PR #3035 (`fix(orders): persist a durable early-cancellation signal`) also now holds for `1881000000000-create-order-cancellation-signals.ts`. `scripts/check-migration-timestamps.mjs` compares the working tree against `origin/main`, so neither branch sees the other today and whichever merges second will fail lint — a merge-window renumber rather than a defect here, but worth knowing which of the two is expected to move.

**[SUGGESTION]** — same file, `reportSkippedRows`
> Correctly noted that `table` comes from a local literal list, so the interpolation is safe. Worth one more line on the `RAISE NOTICE` itself: it reports the *surviving* stray casings but not which uppercase row they collided with, so the operator still has to run the grouping query by hand to see both sides of the choice they are being asked to make.

### Verdict
✅ **Approve**

`mergeable: MERGEABLE`, `mergeStateStatus: BEHIND` (needs a rebase onto `main` before merge). CI is now actually running on this head — `Detect changed paths`, `PHP Unit Tests` and `Scaffolded Adapter Builds` pass; `Lint`, `Type Check`, `Test`, `Build`, `Integration Tests` and `Docker Build Smoke Test` are **pending**, so report them as unknown rather than green. Merge behind #3201 → #3210 order, and after the lint run confirms the renumbered timestamp.

---

norbert-kulus-blockydevs:
## Review addressed — commit `42b0c835d`

- **[IMPORTANT] `NOT EXISTS` guard can abort `up()`** — *fixed*. All three `UPDATE`s now rank the whole folding set (`ROW_NUMBER() OVER (PARTITION BY UPPER("country"), …)`) and rewrite only `rn = 1`, ordering an already-uppercase row first so it is never displaced. Partitions checked against each table's real unique key: PK `(country)` for `sales_document_country_acknowledgments`, `(country, document_kind)` for `…_defaults`, `(country, conditions_hash, effective_from)` for `sales_document_rules`. (#3210's sibling migration sits above this PR and is not in this branch, so the shape was written from scratch rather than copied.)
- **[IMPORTANT] `orderCountry` condition values inverted, not fixed** — *fixed, with one deliberate deviation from your remedy*. `createRule` now folds every `orderCountry` condition's own value via a new `normaliseConditionCountries`, **before** `computeSalesDocumentConditionsHash`; `assertConditionsWellFormed` moved ahead of the fold so a malformed entry is still rejected rather than reaching `.trim()` on a non-string. There is no update path for these rules (`create`/`delete` only on the port; both the rules controller and the template-adopt controller go through `createRule`), so that is the single write seam.

  **I did not fold stored condition values in the migration**, and I think folding them would be a defect: SQL can rewrite the `conditions` jsonb but cannot recompute the application-side SHA-256, and `conditions_hash` is a column of `UQ_sales_document_rules_country_hash_from` — a row whose hash no longer describes its own conditions drops silently out of the conflict guard's `findByCountryAndConditionsHash` pool. Existing stray-case rows are handled on the read side instead: `evaluateSalesDocumentRules` compares the two country codes case-insensitively. The reason is recorded in both the migration docblock and the evaluator. Unit coverage on both halves (write-path value + hash ordering; stray-case match **and** a still-non-matching different country).
- **[IMPORTANT] `listConfiguredCountries` groups by the raw stored country** — *fixed*. All three grouping keys (`countRulesByCountry`, `findAll` defaults, `findAll` acknowledgments) are normalised on read, and rule counts are **summed** across the folding set rather than overwritten, so a country holding rules under both casings reports the real total. No data or schema change. New spec: a `PL`/`pl`/`Pl` mix yields exactly one summary with `ruleCount: 3`.
- **[SUGGESTION] Skipped rows are skipped silently** — *fixed*. A `reportSkippedRows` helper emits one `RAISE NOTICE` per table after the rewrites, naming the count, the surviving casings (`string_agg(DISTINCT …)`) and the remedy — placed after all three updates, where the remaining stray-case set is exactly the skipped set. Matches the `DO $$ … RAISE NOTICE` idiom in `1788000000001-rename-marketplace-capability.ts`.
- **[SUGGESTION] Third copy of `normaliseCountry`** — *fixed (comment only)*, no mirror script, per your call. The docblock names all three siblings (`LocationService.normaliseCountry`, `normalizeCountryCode` in the WooCommerce provisioner helpers, `normalizeCountryIso2` in the KSeF FE config) and states why a shared helper is not viable: the browser bundle cannot import `@openlinker/core` (#591) and a plugin helper must not import a sibling core context, so three of the four could never consume one definition.

The PR description has been corrected — the "also closes the identical case-mismatch" sentence now cites the condition fold rather than implying the scope normalisation did it.

---

norbert-kulus-blockydevs:
Migration-slot coordination — **we moved, you keep `1881000000000`.**

`1881000000000` was claimed by four different migrations across open PRs, not two:

- `1881000000000-normalise-sales-document-country-codes` — this PR
- `1881000000000-create-price-change-auto-applied-log` — the #3139 price-changes stack
- `1881000000000-create-order-cancellation-signals` — #2069
- `1881000000000-seed-erli-prompt-template` — #2202

`check-migration-timestamps.mjs` diffs the working tree against `origin/main` only, so all four pass in isolation and the collision reddens `main` for whoever merges second. Per `docs/migrations.md` § Timestamp uniqueness invariant, TypeORM 0.3.17 sorts by timestamp with no deterministic tie-breaker, so it can leave one `up()` body unapplied while both class names sit in the `migrations` table.

The price-changes stack has moved to **`1884000000000`** (file, exported class and `name` property together) and pushed. It moved rather than this PR because it is one file against your lineage's four (`1881000000000`, `1881000000001`, `1882000000000`, and `1883000000000` on #3188) — not because it was the later claimant; it was the earlier one.

So: **nothing to do here.** Flagging only so both sides are not waiting for the other, which is what the #3159 review asked for explicitly.

Note for #2069 and #2202, which are the two remaining claimants of `1881000000000` and will still collide with each other and with this PR: the first free slot past this lineage is now `1885000000000`.

norbert-kulus-blockydevs:
## Suggestions addressed — commit `12eacdb12`

- **[SUGGESTION] The `RAISE NOTICE` names one side of the collision** — *fixed*. It now emits `pl -> PL [<rest of the uniqueness key>]` pairs, so the operator sees both the stray value and the uppercase value it would fold onto, plus a ready-to-paste `SELECT` that returns the full rows on both sides ordered so the pairs sit together. The rest of each table's key travels with the table name in `SKIP_REPORTED_TABLES` — it differs per table (`country` alone / `+ document_kind` / `+ conditions_hash, effective_from`) and the report is worth little without it on the two whose key is wider than `country`.
- **[SUGGESTION] `1881000000000` also held by open #3035** — *acknowledged; we are not moving, and #3035 has been told why*. Two reasons rather than one: this migration sits mid-stack alongside `1881000000001` (#3210) and `1882000000000` (#3219), so renumbering it forces the other two up as well to keep the stack's own ordering intact; and this stack is 12/13 approved while #3035 is `CHANGES_REQUESTED`, so it is very likely to merge second. Left a heads-up on #3035 naming the collision and the reasoning, so the rebase failure is not discovered cold.

---

piotrswierzy (commented):
## Tech Lead re-review (`4a1f3332` → `9c726201`)

Most of the delta is `main` catching up (the branch merged `origin/main` plus its two sibling branches, pulling in #2883, #3127 and the ADR-073 wording change — none of it this PR's diff, which is still the same eight files). The one substantive change is `fix(migrations): name both sides of a skipped country collision`, and it lands: `reportSkippedRows` now takes each table's uniqueness key beyond `country` (`document_kind`; `conditions_hash` + `effective_from`; nothing for the acknowledgments table, whose PK is `country` alone) and the `RAISE NOTICE` prints `pl -> PL [invoice]` pairs plus a ready-to-paste `SELECT`, so "leave it for a human" now actually names the choice the human is being asked to make. Re-checked the rest at head and it all still holds: the `ROW_NUMBER() OVER (PARTITION BY UPPER("country"), …)` ranking is still the guard (ordering `(country = UPPER(country)) DESC` first, so an already-uppercase row is never displaced, and the `pl` + `Pl` case the `NOT EXISTS` probe would have aborted on is handled by construction); the acknowledgments statement joins back on `country`, which is that table's PK and therefore a sound join key for a self-CTE; `order_records.orderSnapshot` is still deliberately untouched per ADR-014 with the `upper(...)` fold living in `ROUTING_COUNTRY_EXPR` instead; `conditions` is still folded at comparison time in `evaluateSalesDocumentRules` rather than rewritten, because SQL cannot recompute the `conditions_hash` that `UQ_sales_document_rules_country_hash_from` depends on; and the `down()` no-op is correctly argued (no row is ever deleted or merged). On `synchronize`-vs-migration parity: this migration adds no column, constraint or enum — it is a pure data rewrite plus a `RAISE NOTICE` — so the harness's `synchronize`-built schema has nothing to diverge from, and the invariant it does depend on (`sales_document_rules` unique on `(country, conditions_hash, effective_from)`) is an existing one it reads rather than declares. One finding stands, and it is the same one as last round in a new place.

### Findings

**[IMPORTANT]** — `apps/api/src/migrations/1881000000000-normalise-sales-document-country-codes.ts:81`
> `1881000000000` is still claimed by **#3027** (`fix(worker): resolve AI suggestion channel from the connection…`) as well as by this PR. Every guard, `scripts/check-migration-timestamps.mjs` included, diffs the working tree against `origin/main` only, so neither branch can see the other and both pass lint until the second one merges — at which point `main` carries two files sharing a 13-digit prefix and TypeORM 0.3.17 sorts by timestamp with no deterministic tie-breaker, the #374 shape where both class rows appear in the `migrations` table and one `up()` body never runs. Renumbering has already moved this clash three times without resolving it, so the fix is an assignment against the whole outstanding claim set rather than another free-looking number. Current claims across open PRs: `1878`→#3014, `1880`→`main`, `1881`→#3027 **and** #3201, `1881…001`→#3210, `1882`→#3219, `1883`→#3227, `1884`→#3241 **and** #3159, `1885`→#3035. **#3201 is the side that must move**: #3027 is `APPROVED` and `CLEAN` and nine days older, while this PR is `BLOCKED` / `REVIEW_REQUIRED`. Take **`1886000000000`** — above every claim listed, and one step clear of `1887000000000`, which I am asking #3159 to take in the same round. Move the class suffix with it (`NormaliseSalesDocumentCountryCodes1886000000000`, both the `export class` name and the `name =` field), or rule 2 of the invariant fails. Whichever of these merges last should re-run `pnpm lint` against a freshly fetched `origin/main` immediately before merge — that is the only moment the guard can see the other side.

**Also worth recording** (not a finding, no action here): `1878000000000` on #3014 sits **below** `main`'s tail `1880000000000`, which will fail rule 3 of the ordering invariant on rebase. That is #3014's to fix; noted only because the claim map above is what surfaced it.

On the cross-cutting "which connection may a routing decision name" question — it is **not in this PR's scope and unchanged by it**. The diff touches `evaluate-sales-document-rules.ts`, `sales-document-rules.service.ts`, `order-record.repository.ts`, one response DTO and the migration; no composer dialog and no candidate-eligibility predicate is in it. Convergence stays with #3210/#3209.

All three prior findings remain closed. CI at head: `Lint`, `Type Check`, `Test`, `Docker Build Smoke Test`, `PHP Unit Tests`, `Scaffolded Adapter Builds` pass; `Build` and `Integration Tests` are still **queued/pending**, not passing — worth a look before merge.

### Verdict
🔄 **Approve with changes**

---

norbert-kulus-blockydevs:
## Review addressed — commit `5560c582c`

- **[IMPORTANT] `1881000000000` still collides with #3027** — *fixed, but at a different number than you assigned, and the reason matters.*

  Renamed to `1880500000000-normalise-sales-document-country-codes.ts`, class suffix moved with it (both `export class NormaliseSalesDocumentCountryCodes1880500000000` and the `name =` field, so rule 2 still holds). Agreed on which side moves: #3027 is APPROVED, CLEAN and nine days older.

  **Not `1886000000000`, because that one costs five renumbers instead of one.** #3201 is UPSTREAM of four siblings in the same merge chain — #3210 (`1881000000001`), #3219 (`1882000000000`), #3227 (`1883000000000`), #3241 (`1884000000000`). The ordering invariant requires a migration absent from `origin/main` to carry a prefix strictly greater than main's highest, so merging this at 1886 would push main's tail to 1886 and drop all four downstream siblings below it. `1880500000000` sits above main's tail (`1880000000000`) and below #3027's `1881000000000`; the whole band `1880000000001…1880999999999` was verified free by an independent scan of every open PR, which also reproduced your claim map exactly. A non-round prefix is conventional here — **75 of the 169** prefixes on `main` are non-round.

  Guard output at head: `migration-timestamps: OK (173 migrations (incl. 2 plugin dir); ordering vs origin/main: checked)` and `self-check OK`, both exit 0.

  Consequence worth stating for #3027's author: **#3027 needs to do nothing.** It keeps `1881000000000`, and `1886000000000` is now unclaimed anywhere — verified zero occurrences on `main` and across every open PR.

- **[note] `1878000000000` on #3014 sits below main's tail** — *acknowledged, not touched.* That is #3014's to fix on its own rebase; nothing here reaches it.

- **On the cross-cutting predicate question** — agreed it is out of this PR's scope and unchanged by it. Tracked as **#3232**.

Endorsing your closing point: whichever of #3201/#3027 merges last should re-run the guard against a freshly fetched `origin/main` immediately before merge — that is the only moment it can see the other side.

---

piotrswierzy (approved):
## Tech Lead re-review (`9c726201` → `5560c582`)

The delta is exactly one commit, `5560c582`, and it closes the one open IMPORTANT: **the migration is renumbered `1881000000000` → `1880500000000`**, filename, class and `name` property all moved together (`git` reports the file as a clean rename, body untouched). Since this migration has never run anywhere, renaming the tracked class carries no orphan-`migrations`-row risk. All substantive work from the prior round is unchanged and stays closed — the `ROW_NUMBER() OVER (PARTITION BY UPPER("country"), …)` ranking guard, the `orderCountry` inversion fix at both ends, `listConfiguredCountries` folding all three grouping keys, and the per-table `RAISE NOTICE` remainder.

I re-enumerated migration filenames **across all 52 open PRs**, not just against `main`, and the renumber is correct on every axis:

- **Collision with #3027 resolved.** `1881000000000` is now held only by #3027's `seed-erli-prompt-template.ts`. This PR moved off it.
- **`1880500000000` is unique repo-wide.** It appears only on this PR and its own descendant stack (#3209–#3219 inherit the identical file). No independent PR holds that prefix, and this PR's head tree has no duplicate 13-digit prefix anywhere.
- **Ordering rule 3 satisfied.** `main`'s tail is `1880000000000`; `1880500000000` is strictly greater.
- **The #3210 coupling is handled — and moving *down* is why.** #3210's head already carries `1880500000000` alongside its own `1881000000001-unify-sales-document-country-default-per-country.ts`, and `1880500000000 < 1881000000001`, so this normalisation still runs **before** #3210's collision guard, which tests the `(country, document_kind)` key while that uniqueness still exists. This is the better of the two conflicting remedies: assigning `1886`/`1887` would have sorted this migration *after* its own dependent and forced #3210 to move in the same commit; going to a half-step below `1881` needs no accompanying change at all.

It is a data-only migration — no `CREATE TYPE`, no `ADD CONSTRAINT`, no `CHECK` — so the `synchronize`-builds-the-test-schema caveat (neither PG enums nor CHECKs are emitted there, meaning a constraint declared only in the migration would hold in production and silently not in tests) does not bite here. Both `up()` and `down()` are present, `down()` deliberately a no-op for an irreversible normalisation. No controller and no stylesheet in the delta, so route authorization and the CSS-existence check are non-applicable this round.

### Findings

**Observation, not a finding on this PR.** Five further-downstream stack members — #3220, #3225, #3227, #3241, #3242 — still carry the file at the **old** `1881000000000`, i.e. they have not taken this renumber forward and still collide with #3027. That is ordinary stack-propagation lag rather than a defect here, but it is worth tracking: #3027 is an independent PR that could merge at any time, and if any of those five lands before merging its base forward the collision simply reappears under a different number.

### Verdict
✅ **Approve**

---

Issue #3108, opened 2026-09-10
feat(nav): role-gated "Pack bench" nav entry


norbert-kulus-blockydevs:
## Problem / Context

`/bench` has zero references in `apps/web/src/app/nav-registry.ts` — it's URL-only.

## Proposed Solution

Add a "Pack bench" nav entry visible to `packer` (and admin/operator), linking to `bench.route.tsx`.

## Classification

**Type**: Frontend
**Layer**: Interface (nav)
**File(s)**: apps/web/src/app/nav-registry.ts

## Dependencies

- Part of #3104

## Assumptions

- None

## Acceptance Criteria

- [ ] A packer-role user sees "Pack bench" in the sidebar
- [ ] Tests added or updated for non-trivial logic
- [ ] No architecture boundary violations (CORE ↔ Integration)

norbert-kulus-blockydevs:
Delivered by #3204, merged into `3107-widen-role-values` (2026-09-14). The `Closes #3108` clause never fired because the PR's base was an epic branch, not `main`.

Evidence: commit `56d0ff431` adds the role-gated `/bench` "Pack bench" entry to `apps/web/src/app/nav-registry.ts`.

**Not on `main` yet.** It reaches `main` via the epic PR #3244 (`3104-oms-onboarding-ui` → `main`), currently a draft. Closing the issue because the work is done and reviewed; the epic tracks the merge.

PR #3202, opened 2026-09-11
feat(web): widen the FE chrome's RoleValues with packer

### LINKED ISSUE #3107: feat(nav): widen RoleValues with packer
## Problem / Context

`apps/web/src/app/nav-registry.types.ts` declares `RoleValues = ['admin', 'operator']`, so the sidebar cannot express a nav item visible to a `packer` — the role the bench's own backend `@Roles` decorators already include.

## Proposed Solution

Add `'packer'` to `RoleValues`. Per `apps/api/src/auth/user-role-values.spec.ts`'s backend precedent, review every existing role-gated nav item in the same commit to confirm a packer should not newly see them.

## Classification

**Type**: Frontend
**Layer**: Interface (types)
**File(s)**: apps/web/src/app/nav-registry.types.ts

## Dependencies

- Part of #3104

## Assumptions

- None

## Acceptance Criteria

- [ ] RoleValues includes 'packer'
- [ ] Every pre-existing role-gated nav entry is reviewed and unaffected for a packer

norbert-kulus-blockydevs:
## Summary
- Widens `RoleValues`/`Role` in `apps/web/src/app/nav-registry.types.ts` from `['admin', 'operator']` to `['admin', 'operator', 'packer']`
- `packer` mirrors the backend's `UserRoleValues` (ADR-071/#2413), which already carries it — this is the FE chrome's own, narrower role union used for sidebar/⌘K nav gating catching up
- No behavioural change from the widening alone. Audited every consumer of `Role`/`RoleValues` (`nav-registry.ts`, `merge-nav-contributions.ts`, `plugin.types.ts`) — none does exhaustive matching over the union, so nothing needed a new arm
- This is a pure prerequisite for #3108 (the "Pack bench" nav entry), which needs an item-level role gate — `packer`'s backend `ROLE_PERMISSIONS` grant is deliberately empty (ADR-071), so the existing `requiresPermission` item gate cannot express "packer may see this item"; only a role check can

First sub-issue of epic #3104, stacked on `3118-oms-onboarding-mockup` (the mockup this epic's sub-issues build against) rather than `3104-oms-onboarding-ui` directly, since the mockup PR (#3125) hasn't merged into the epic branch yet — same stacked-PR shape the #3063 inventory-locations epic used.

Closes #3107

## Test plan
- [x] `vitest run src/app/nav-registry.test.ts` — 7 passed (2 new: pins `packer` is a `RoleValues` member, and that `buildNavGroups` doesn't crash with no role passed)
- [x] `eslint` on both changed files, scoped — clean
- [x] Full `eslint 'apps/web/src/**/*.{ts,tsx}'` run locally — **0 errors** (123 pre-existing warnings)

> **Correction (review #3202/#3204).** An earlier version of this line said the skipped full-repo lint would be run by CI. That was not true: `.github/workflows/ci.yml` triggers on `pull_request: branches: [main]`, and this PR targets a stack branch — `gh pr checks` reports no checks at all, so no Lint, Type Check or Test job has ever run against it. **The first CI gate for this work is the epic → `main` PR**, by which point several stacked PRs have merged into the epic branch, so a lint break surfaces there with N commits on top of it. The full web-scoped lint, type-check and affected suites were therefore run locally instead; the full monorepo `pnpm -r lint` still is not (documented machine limitation).


piotrswierzy (commented):
## Tech Lead review

A three-line widening with a good docblock, and the premise checks out: the backend `UserRoleValues` at `libs/core/src/users/domain/types/role.types.ts:31` already carries `packer` (landed by #2413), `ROLE_PERMISSIONS.packer` really is `[]` by design, and `apps/api/src/auth/user-role-values.spec.ts` correctly does **not** fire here — that tripwire guards the backend union and the `@AnyRole()` sites, and this PR touches neither. The claim that no consumer needs a new arm is also literally true: neither `nav-registry.ts` nor `merge-nav-contributions.ts` does exhaustive matching. But the audit answered the wrong question, and #3107's own AC is the one that exposes it.

### Findings

**[IMPORTANT]** — `apps/web/src/app/nav-registry.types.ts:40` (the widening itself)
> Both consumers of `Role` in a gating position do a **non-exhaustive equality against `'admin'`**, not an exhaustive match:
>
> - `nav-registry.ts:136` — `if (group.kind === 'live' && group.requiresRole === 'admin' && !isAdmin)`
> - `merge-nav-contributions.ts:41` — `if (contribution.requiresRole === 'admin' && !isAdmin) return false;`
>
> So every non-`admin` member of the union is an **inert, fail-open gate**: `LiveNavGroup.requiresRole: 'packer'` and `NavContribution.requiresRole: 'packer'` both type-check and make the group or contribution visible to *everyone*. The PR's audit asked "does anything break?" and correctly answered no; the question this widening needed was "does anything now silently not work?", and the answer is two gates. This is pre-existing for `'operator'` — but the docblock two lines above the change asserts the opposite and this PR leaves it standing:
>
> > *"`'operator'` is included so nav groups can declare `requiresRole: 'operator'` for operator-only sections without any further type changes."*
>
> That sentence is false today. Adding `'packer'` beside it makes a third value false the same way, on a field whose whole purpose is access gating, failing **open**. `docs/frontend-architecture.md § Access Control And UI Visibility` names the inverse hazard as the reason `session.user.role` must never be compared inline — *"a typo type-checks and silently evaluates false"* — and fail-open is the worse direction of the same defect. Either correct the docblock to say plainly that only `'admin'` is honoured by the group and contribution gates, or make those two sites compare against the session role properly. The docblock correction is the minimum; shipping the widening with the false sentence intact is not.

**[IMPORTANT]** — process
> **CI has never run on this PR.** `.github/workflows/ci.yml` triggers on `pull_request: branches: [main]`; this targets `3118-oms-onboarding-mockup`, and `gh pr checks 3202` reports *"no checks reported"*. The test plan's *"Full-repo `pnpm -r lint` pre-commit step skipped … CI will run it"* is not true — nothing runs until the epic branch opens against `main`. Same note as on #3204; stated on both because the claim appears in both bodies.

**[SUGGESTION]** — `apps/web/src/app/nav-registry.types.ts:34-38` (the `viewer` paragraph)
> > *"Add it here only when a nav gate actually needs to name it."*
>
> Reasonable rule, and the very next gate needs it. `packer` appears in `@Roles(...)` on exactly three controllers API-wide (the bench ones), so a packer already sees the whole ungated Operations group and 403s on every entry in it — see the #3204 review. The natural remedy is `requiresRole: ['admin', 'operator', 'viewer']` on those items, which this union cannot express. Not a change to make here on spec — but worth recording in the paragraph that the omission has a known, already-identified consumer, so the next person does not re-derive it.

**[SUGGESTION]** — `apps/web/src/app/nav-registry.test.ts:70`
> `it('does not crash buildNavGroups when no role is passed')` asserts `byLabel(groups, 'Operations')?.kind === 'live'`, which is true on `main` before this PR and stays true after it — the test cannot fail for the reason it names, because `buildNavGroups` does not read `role` at all yet. That is honest as a placeholder (#3108 supplies the real coverage, and does), but as written it reads as coverage it is not. Either drop it and let #3108's `is hidden when no role is known yet` carry it, or say in the test name that it pins the absence of a required-role parameter.

### Positive observations

- Correctly identifying that `requiresPermission` **cannot** express a packer-visible item — because `ROLE_PERMISSIONS.packer` is deliberately empty rather than merely unpopulated — is the whole reason this prerequisite is legitimate rather than a shortcut, and the docblock says so at the right length.
- Not touching the backend `UserRoleValues`, and therefore not tripping `user-role-values.spec.ts`, is right: this union is the FE chrome's own and narrower by design.
- `RoleValues` pinned by a test rather than only by a type is consistent with how the backend guards its own union.

### Verdict

🔄 **Approve with changes**

1. Fix or delete the `'operator'` sentence in the `RoleValues` docblock. A widening PR whose own docblock describes a capability the code does not have is the one thing that must not merge as-is; everything else here is fine.
2. Correct the test plan's CI claim.

---

norbert-kulus-blockydevs:
## Review addressed — `0e309703`

All four findings actioned. The BLOCKING-adjacent one is fixed **further than the review asked**, because the minimum it offered leaves the footgun expressible.

---

### IMPORTANT · the `'operator'` sentence is false — fixed by making the state unrepresentable

Confirmed exactly as you state, and I checked both sites rather than taking them on trust:

```
nav-registry.ts:136              group.requiresRole === 'admin' && !isAdmin
merge-nav-contributions.ts:41    contribution.requiresRole === 'admin' && !isAdmin
```

(minor: the second is `apps/web/src/**plugins**/merge-nav-contributions.ts`, not `app/` — right line, right code.)

You offered "correct the docblock, or make the sites compare the session role", and called the docblock the minimum. I did **neither**, and took a third option that is smaller than the second and stronger than the first: `GroupRoleGate = 'admin'` now types `LiveNavGroup.requiresRole` and `NavContribution.requiresRole`.

That is better than the prose fix for a specific reason — a corrected comment leaves `requiresRole: 'packer'` a legal, type-checking, silently fail-**open** declaration, and the next person to write one has to have read the comment. Now it is a compile error. The `GroupRoleGate` docblock names what widening it costs: the value **and** the gate that reads it, which must then compare the session role rather than the `isAdmin` boolean both currently take.

`Role` stays the full union — #3108's item-level `requiresRole` honours every member, so only the two group-grain fields narrow. **Costs nothing**: all four declarations in the tree, data and tests, are already `'admin'`.

Two things came with it:

- **A third false docblock you did not name.** `NavContribution.requiresRole` claimed "the contribution is dropped for sessions whose role doesn't match" — false the same way, on the **plugin-facing** contract. Corrected, and narrowing it means a plugin author declaring a non-admin gate gets a compile error instead of a gate that silently shows the contribution to everyone. That is a deliberate contract narrowing, stated as such in the commit.
- **ESLint caught it, which is the guard working.** `shared/plugins/` may only type-import an allowlisted set from `app/`, and `GroupRoleGate` was not on it. I swapped `Role` **for** `GroupRoleGate` rather than adding it alongside — leaving `Role` importable there would keep the fail-open declaration expressible from a plugin, which is the whole thing being closed.

There is also a **third** gate site neither of us listed: `command-palette-provider.tsx:194` tests `liveGroup.requiresRole !== undefined` rather than `=== 'admin'`. Different shape, and it fails *closed* in demo mode, so it is not part of this defect — noting it so the next reader does not think it was missed.

### SUGGESTION · the `viewer` paragraph — done, and the consumer is now filed

The paragraph records that the omission has a known, already-identified consumer, with the measured count. That consumer is now **#3221**, filed under epic #3104 — see the #3204 comment for what the measurement actually showed, because it corrects the framing in a way that matters for the remedy.

### SUGGESTION · `nav-registry.test.ts:70` — replaced, not deleted

You were right that it could not fail for the reason it named: `buildNavGroups` reads no role at this point in the stack, so it passed identically before the widening.

Rather than drop it, I replaced it with one that **can** fail — every `BASE_NAV_GROUPS` role gate is `'admin'`. That catches a value smuggled past `GroupRoleGate` by a cast, or a future widening of the type without its gate.

Verified red-first, and the result is better evidence than I expected:

```
× declares no group role gate other than admin, which is the only one honoured
FAIL  hides AI and Administration for non-admins
FAIL  renders AI and Administration as restricted for a non-admin
Tests  3 failed | 4 passed
```

A cast-in `'operator'` takes **two pre-existing tests down with it** — because the AI and Administration groups really do become visible to non-admins. That is the fail-open, demonstrated.

### IMPORTANT · the CI claim — corrected, and the local gate actually run

You are right and I verified it: `ci.yml` is `pull_request: branches: [main]`, and `gh pr checks 3202` reports no checks. The test plan's mitigation did not exist.

Rather than only reword it, I ran what this machine can run:

- full `eslint 'apps/web/src/**/*.{ts,tsx}'` — **0 errors** (123 pre-existing warnings)
- `nav-registry.test.ts` + `merge-nav-contributions.test.ts` — 15 passed

The PR body now says plainly that the first CI gate is the epic → `main` PR, with N commits stacked above by then.

---

**Not here, deliberately**: the `docs/frontend-architecture.md § Access Control And UI Visibility` row recording this narrowing lands in #3204. That section is one block and the role-gate row is #3204's own suggestion, so splitting it across the stack produces only a conflict.

piotrswierzy (approved):
## Tech Lead re-review (`fc739c4e` → `0e309703`)

Both prior findings are closed, and the sharp one is closed harder than I asked for. I said the minimum was correcting the false `'operator'` sentence in the docblock; instead the fail-open state was made **unrepresentable** — `LiveNavGroup.requiresRole` and `NavContribution.requiresRole` are now typed `GroupRoleGate = 'admin'`, so a declaration the gate would silently ignore is a compile error rather than a comment asking the next author to notice. That is the right instinct: a prose limit standing beside a wider type is exactly how the sentence I flagged came to be false in the first place. The `.eslintrc.js` pinhole swap (`Role` → `GroupRoleGate`) is correct and complete — `plugin.types.ts` is the only file under `shared/plugins/**` that imported the name, and it now imports the narrow one, so the wide union is no longer even reachable from a plugin author's contract surface. Both SUGGESTIONs were taken too.

### Findings

**Closed — [IMPORTANT]** `nav-registry.types.ts:40` — the inert non-`admin` gate
> Closed by construction, not by documentation. `GroupRoleGate` is a one-member type with a docblock that states the widening rule as a two-file change and names both gate sites; `nav-registry.ts:136` and `merge-nav-contributions.ts:41` are unchanged and still correct against it. Verified `BASE_NAV_GROUPS` declares `requiresRole: 'admin'` at two sites and nothing else, so nothing on `main` becomes a type error.

**Closed — [IMPORTANT]** process / CI claim
> The body now says plainly that `.github/workflows/ci.yml` triggers on `pull_request: branches: [main]`, that `gh pr checks` reports nothing, and that the first CI gate is the epic → `main` PR. I re-confirmed both: still `no checks reported on the '3107-widen-role-values' branch`, and the trigger is unchanged. Accurately stated is all I asked for — the stacked shape itself has precedent.

**Closed — [SUGGESTION]** the `viewer` paragraph
> Now records the known consumer explicitly (a packer sees every other Operations entry and 403s on each), names the per-item remedy the union cannot yet express, and points at the epic. #3221 is filed. The next reader does not re-derive it.

**Closed — [SUGGESTION]** `nav-registry.test.ts:70` — the placeholder test
> Replaced with `declares no group role gate other than admin, which is the only one honoured`, which can genuinely fail: it asserts over the shipped `BASE_NAV_GROUPS` data, so a value smuggled past `GroupRoleGate` by a cast goes red. Guarding the **data** as well as the type is the right belt-and-braces here, because the type only binds new declarations.

**[SUGGESTION]** — `apps/web/src/app/nav-registry.test.ts:81`
> `(group): group is Extract<typeof group, { requiresRole?: unknown }>` compiles only because `{ requiresRole?: unknown }` is a *weak type* (all-optional), so TS's weak-type detection excludes `PlannedNavGroup` from the `Extract` and `.map(g => g.requiresRole)` type-checks. That is a real rule and the author's `type-check` run confirms it holds — but it is a subtle one to rest a predicate on, and a reader adding one optional property to `PlannedNavGroup` would silently widen the `Extract` back to the full union and break the file for a reason unrelated to what they changed. `group is LiveNavGroup` says the same thing without the mechanism. Non-blocking.

### Positive observations

- Replacing the `Role` pinhole rather than adding `GroupRoleGate` beside it is the load-bearing half of the ESLint change, and the comment says so — leaving `Role` importable would have left the fail-open declaration expressible from a plugin, which is where it is least likely to be caught.
- `GroupRoleGate`'s docblock states the widening cost in the form a future author will actually hit ("both of which must then compare against the SESSION role rather than the `isAdmin` boolean they take today") and points at `LiveNavItem.requiresRole` as the worked example of the honoured shape. That is the difference between a narrow type and a narrow type someone will correctly widen later.
- The `RoleValues` docblock now leads with what the union is *not* (`## This union is NOT what the group and contribution gates accept`), which is the fact a reader needs before the one they came for.

### Verdict

✅ **Approve**

---

piotrswierzy (approved):
## Tech Lead re-verification (head unchanged at `0e30970`)

Head and `main` are both unchanged, so I re-checked only the property the approval rested on — that the fail-open nav gate is **unrepresentable**, not merely unused — and the surrounding CI state.

The property holds in all five places it has to. `GroupRoleGate = 'admin'` (`nav-registry.types.ts:84`); `LiveNavGroup.requiresRole?: GroupRoleGate` (`:130`); `NavContribution.requiresRole?: GroupRoleGate` (`plugin.types.ts:71`); `BASE_NAV_GROUPS` declares only `'admin'` (`nav-registry.ts:81`, `:90`) and `:136` tests `group.requiresRole === 'admin'`. The `.eslintrc.js` pinhole **replaces** `Role$` with `GroupRoleGate$` in both the `importNamePattern` and the operator-facing message rather than joining it — so the wide union is genuinely unreachable from `shared/plugins/`, which is the whole of the fix. A plugin author declaring `requiresRole: 'operator'` now gets a compile error instead of a gate that silently shows the contribution to everyone.

**CI: still absent, and I am reporting it as absent rather than as passing.** `GET /commits/0e30970…/check-runs` returns `total_count: 0` — no Lint, Type Check or Test job has ever run against this head.

One correction to the mechanism the PR body gives for that, worth recording because it will be relied on again: `pull_request: branches: [main]` is **not** a sufficient explanation. Three sibling PRs on non-`main` bases in the neighbouring #3173 stack (#3212, #3214, #3217) *do* carry CI runs, against a byte-identical `ci.yml` at their own heads. So "a stacked PR gets no CI" is not a rule in this repo — CI on a stacked PR is incidental, and its absence here cannot be predicted from the workflow file. The body's **conclusion** stands unchanged (the first real gate for this work is the epic → `main` PR, by which point several slices sit on top of the break); only the stated reason does not carry weight.

**Stack**: this is the separate #3104 OMS-onboarding chain — `3104-oms-onboarding-ui` (epic branch) ← #3125 ← **#3202** ← #3204. Blocked behind #3125 reaching the epic branch, and behind the epic branch reaching `main`; nothing blocks it on its own merits (`mergeable: CLEAN`, `APPROVED`).

### Findings
No new findings. One clarification recorded above: the PR body's CI-absence *reason* is not load-bearing, though its conclusion is. Not worth a push on an unchanged head.

### Verdict
✅ **Approve**

---

piotrswierzy (approved):
## Tech Lead re-verification (head unchanged at `df2ab205`)

`MERGEABLE`/`CLEAN`, approved. The gate change re-verified at this head and intact: `GroupRoleGate = 'admin'` is declared once in `nav-registry.types.ts` and applied at **both** gate sites — `NavGroup.requiresRole?: GroupRoleGate` and `NavContribution.requiresRole?: GroupRoleGate` in `shared/plugins/plugin.types.ts` — and the `.eslintrc.js` pinhole **replaces** `Role$` with `GroupRoleGate$` in both the negative-lookahead pattern and its message, rather than adding alongside it. `RoleValues` widens to `['admin','operator','packer']` with `Role` left wide for the item-level gate that genuinely honours it. Unrepresentable, not merely documented — as approved.

### Findings

**[IMPORTANT — merge path, not code]** This PR is **not** in the `#3173` stack. It roots on a separate chain: `3104-oms-onboarding-ui` → #3125 → **#3202** → #3204. Two things about that root are new information since the last round.

1. **#3125 is now `CLEAN`/`APPROVED`** (approved 2026-09-14T13:23:17Z, superseding the earlier `CHANGES_REQUESTED`). `gh pr list` still renders the stale `CHANGES_REQUESTED` for it; the REST review list is authoritative. So #3125 is no longer the blocker it was.
2. **The chain's root branch is an orphan.** `3104-oms-onboarding-ui` is at `42cc86dd`, which compares to `main` as `status=behind, ahead=0, behind=17` — it contains **nothing** `main` does not already have — and **no pull request exists for it in any state**. So as the stack stands, nothing in this sub-chain has a path to `main`: #3125 targets a branch that will never be merged.

The remedy is cheap and belongs on #3125, not here: retarget #3125's base to `main`. Because its current base is a strict ancestor of `main`, that retarget changes no content — it only gives the chain a terminus. Worth doing before either of these two is queued.

**[IMPORTANT — CI]** Still **zero** CI runs, ever: `gh api actions/runs?branch=3107-widen-role-values` returns `total_count=0`. Reported as absent, never as passing.

The cause stated in my earlier round was **wrong** and I am correcting it here. `ci.yml` declares `pull_request: branches: [main]`, which reads like the explanation — but it is not, and I now have the counter-example pinned: run [`34831266177`](https://github.com/openlinker-project/openlinker/actions/runs/34831266177) is an `event=pull_request` run of `.github/workflows/ci.yml` on head `3179-connection-health-doc-count` with base **`3175-sales-document-adr`**, a non-`main` base, and it ran. A base-ref filter that let that through cannot be what suppressed this. The absence here is real, total and currently unexplained by the workflow trigger; the likeliest practical fix is the retarget-and-rebase above, which produces a fresh `synchronize` on a `main`-based chain.

Neither item asks for a change to this diff, and the approval stands on the code.

### Verdict
✅ **Approve**

---

Issue #2416, opened 2026-08-22
[FE] W3b-3 — Surfaces B + C: the work list, and scanner-first operation


piotrswierzy:
> **Wave 3b**, epic #2422. Spec of record: [`docs/specs/product-spec-oms-wave3b-scan-pick-pack.md`](../blob/main/docs/specs/product-spec-oms-wave3b-scan-pick-pack.md) **§ 2.2 (B1–B5)** and **§ 2.3 (C1–C4)**; decisions **D8, D11, D21, D22**. Mockup: [`work-list`](../blob/main/docs/plans/mockups/oms-wave3b/README.md) — three variants (work list / nothing to pack / routing switched off).

## ⚠️ Re-scoped 2026-09-03

Originally *"scan surface shell: route, device posture, scanner-input primitive"*. Picking is cut (§ 5), so there is no aisle scan flow. This is now **the bench's work list and the scanner-input primitive that Surfaces D/E consume**.

The route shell already exists — #2413 put `/bench` in `standaloneRoutes`, deliberately outside `AuthenticatedAppLayout`, because the idle lock clears the session and a layout redirect would unmount the bench and destroy an open parcel. **Do not move it back under the authenticated layout**; `bench-route-placement.test.ts` pins this.

### Surface B — the work list

**B1 — the list is authoritative, not derived (D8).** It shows work **assigned to this holder, accepted and not yet closed** — `fulfillment.work.dispatch` already offers routed work to its holder. A deadline-derived queue would be *a second, weaker answer beside an authoritative one*, and would show work routed to a 3PL that this bench must never see.

**B2 — ordered by urgency, and it never implies readiness.** Deadlines in plain words. It must **never state or imply that stock has been picked or gathered** — OpenLinker cannot see a shelf. Rows count *units to verify*, never *picked*. This is not a wording preference: implying readiness sends a packer to fetch something that is not there, and after that happens twice the list is not trusted again.

**B3 — the two empty states are different facts.** "Nothing to pack right now" and "routing is not switched on, so this bench will never receive work" must be distinguishable, and the second must name its remedy. #2407 already ships the routing-readiness concept and a `POST /inventory/locations/bootstrap` remedy — reuse that vocabulary rather than inventing a second one.

**B4 — state is carried by colour *and* position or text, never colour alone** (P9).

[...]

Issue #2413, opened 2026-08-22
[FE+API] W3b-1 — Surface A: the bench, the packer role, idle lock and handover


piotrswierzy:
> **Wave 3b**, epic #2422. Spec of record: [`docs/specs/product-spec-oms-wave3b-scan-pick-pack.md`](../blob/main/docs/specs/product-spec-oms-wave3b-scan-pick-pack.md) **§ 2.1 (stories A1–A5)** and decisions **D1, D2, D3, D12, D13, D15, D16**. Decision: [ADR-071](../blob/main/docs/architecture/adrs/071-pack-station-principal.md). Mockup: [`bench-identity`](../blob/main/docs/plans/mockups/oms-wave3b/README.md) (locked / sign-in / handover variants).

## ⚠️ Re-scoped 2026-09-03

Originally *"pack-station session wiring per #2080"*. #2080 is now decided: **there is no pack station principal.** ADR-071 rejects the device session + PIN actor entirely. This issue no longer wires a station session — it adds a narrower role, an attribution column, and three pieces of bench UX.

## What this is

Terminals are **shared and roaming** (D15), but a person changes terminal only a few times a shift (D16) — so sign-in is a shift-boundary event, not a per-parcel one. That is precisely what makes ordinary password sign-in acceptable friction and lets ADR-071 avoid a fifth credential entity.

### 1. The `packer` role

Add `packer` to `UserRoleValues` (`libs/core/src/users/domain/types/role.types.ts`). Packers are ordinary users on a narrower role, **not** `operator` (D12): *"at 1000/day with temps, 'every temp packer can read the customer database' is not a defensible posture."*

**This will trip #2079's `UserRoleValues` tripwire, by design.** That tripwire exists exactly for this moment — see § 3 below, which is the substantial half of this issue.

### 2. Attribution — `packedByUserId` ⊕ `packedByService`

On `fulfillment_works`, mirroring `CHK_fulfillment_holds_actor`'s XOR precedent (ADR-071 decision 2), so *"a 3PL packed this"* and *"a human packed it, unrecorded"* are not the same value.

- **Grain is per work, per phase** (D4). The order-grain fact is **derived** — an order can split into several works, and an order-grain person field names one packer and drops the other.
- **The last verifier owns the parcel** (D13), a consequence of D18's auto-close. Under roaming this can be someone who checked one item of five. **Record that as a limitation; do not present the field as a complete account of who handled the box.** The verification ledger holds the rest.
[...]

Issue #2358, opened 2026-08-22
[TASK] CORE — automation_rules storage mirroring the #2161 rule engine


piotrswierzy:
> Part of the OMS programme backlog (**Wave 2**, stream S2, size M). Backlog slug: `W2-21`. Sibling `W2-*` slugs resolve via the Wave-2 epic's checklist. Design of record: Wave-2 product spec §5 + §7.2 (PR #2318).

## Problem / Context

Spec §5. The repo already has a shipped, reviewed rule engine — `sales_document_rules` (`libs/core/src/sales-documents/infrastructure/persistence/entities/sales-document-rule.orm-entity.ts`, `domain/domain-services/evaluate-sales-document-rules.ts`). Inventing a second shape would double the surface area of "how a rule is stored and evaluated" in one repo.

## Proposed Solution

New `automation_rules` table in a new `libs/core/src/automation/` context, mirroring the #2161 shape: scope columns as real columns, kind-specific data in `jsonb`, `effectiveFrom`/`effectiveTo`, `isActive`, plus AND-only closed-vocabulary conditions. Three **declared divergences** from #2161 (from the spec, and to be restated in the entity docblock so a future reader sees them as intentional):

1. Scoped by **trigger**, not by country.
2. `orderTotalGross` conditions carry an **inline amount + currency**, not a `thresholdRef`.
3. Actions are a **multi-step list capped at 3**, stop-on-first-failure — #2161 has one outcome.

No `priority` column (the spec drops it deliberately; a duplicate guard at save time replaces it). Plus the **`automation_runs`** table (spec §7.2) — one row per firing, carrying the per-step outcomes. Note the spec's correction to the obvious design: the run row is written **for every firing**, including those whose action dispatched a `sync_jobs` job, so the history is complete rather than complete-for-some-actions. `sync_jobs` remains where *technical* failure detail lives and is linked to, never the substitute for a run row. `W2-47` owns the write path and the per-step outcome shape; this issue lands the table.

## Classification

**Type**: CORE
**Layer**: Domain + Infrastructure (**migration-bearing**; synthetic-sequential timestamp per `docs/migrations.md` rule 3)
**File(s)**: new `libs/core/src/automation/**`; new migration; `libs/core/package.json` exports; `libs/core/src/__tests__/barrel-purity.spec.ts`

## Dependencies

- None beyond Wave 1a conventions (#2308 / `W1a-8`)

## Assumptions

[...]

PR #3204, opened 2026-09-11
feat(web): role-gated "Pack bench" nav entry

### LINKED ISSUE #3108: feat(nav): role-gated "Pack bench" nav entry

norbert-kulus-blockydevs:
## Summary
- Adds a "Pack bench" entry (`/bench`) to the Operations nav group, gated to exactly the roles the bench API itself accepts — `admin`, `operator`, `packer` — i.e. everyone except `viewer`
- Adds a new item-level `LiveNavItem.requiresRole?: readonly Role[]` gate, since neither existing item gate could express "packer may see this": `requiresPermission` has nothing to check for `packer` (its `ROLE_PERMISSIONS` grant is deliberately empty per ADR-071), and the existing role gate (`LiveNavGroup.requiresRole`) is group-level and single-valued — using it here would have hidden the entire Operations group rather than gating one item
- Extracts the combined permission+role visibility check into one exported `isNavItemVisible(item, {permissions, role})` helper, consumed by **both** `buildNavGroups` (sidebar) and `command-palette-provider.tsx` (⌘K). Before this PR only the sidebar was going to gain the role check — leaving ⌘K a way around it into a route some roles can't call, the exact class of bug the existing `requiresPermission` comment in `nav-registry.types.ts` already calls out for `/automations`
- Threads the session's raw role string through `app-shell.tsx` into `buildNavGroups`, gated on `isReady` (same pattern as the existing `isAdmin`/`isViewerOnly` derivations)

## Design decision: allowlist, not admin-inclusive-by-default
`requiresRole` takes an explicit array rather than a single role with an implicit "admin always sees everything" fallback (unlike the group-level gate, which special-cases `admin`). The bench's real audience is `admin | operator | packer`, so the item spells that out — this keeps the gate legible from the nav-registry data alone rather than requiring a reader to know an implicit rule applies.

Second sub-issue of epic #3104, stacked on #3202 (`3107-widen-role-values`, itself stacked on the mockup branch `3118-oms-onboarding-mockup` since #3125 hasn't merged into the epic branch yet) — same stacked-PR shape the #3063 inventory-locations epic used.

Closes #3108

## Test plan
- [x] `vitest run` scoped to `nav-registry.test.ts` (24 tests, 17 new: `isNavItemVisible` unit coverage + `buildNavGroups` integration coverage for admin/operator/packer/viewer/no-role), `command-palette-provider.test.tsx` (18 tests, 3 new), `app-shell.test.tsx` (25 tests, unaffected) — all green
- [x] `eslint` on every changed file — 0 errors (3 pre-existing missing-return-type warnings, unrelated)
- [x] Full `eslint 'apps/web/src/**/*.{ts,tsx}'` run locally — **0 errors** (123 pre-existing warnings)

> **Correction (review #3202/#3204).** An earlier version of this line said the skipped full-repo lint would be run by CI. That was not true: `.github/workflows/ci.yml` triggers on `pull_request: branches: [main]`, and this PR targets a stack branch — `gh pr checks` reports no checks at all, so no Lint, Type Check or Test job has ever run against it. **The first CI gate for this work is the epic → `main` PR**, by which point several stacked PRs have merged into the epic branch, so a lint break surfaces there with N commits on top of it. The full web-scoped lint, type-check and affected suites were therefore run locally instead; the full monorepo `pnpm -r lint` still is not (documented machine limitation).

**A real timing bug this PR's own tests caught and fixed along the way:** `SessionProvider` starts at the anonymous session and resolves asynchronously (`isReady` starts `false`), so a role-gated nav item is absent on the very first render regardless of which role is about to load. The palette tests originally asserted synchronously right after `keyDown` and passed for the wrong reason (every gated item, including the one under test, was hidden by the *unresolved* session, not by the role check) — rewritten to `findByText`/wait for an existing async signal before asserting.


piotrswierzy (commented):
## Tech Lead review

The mechanism is right and the reasoning behind it is right. `isNavItemVisible` as one exported rule consumed by both `buildNavGroups` and `command-palette-provider.tsx` is the correct response to the problem `nav-registry.types.ts:42` already documents, the fail-closed default (no role known ⇒ role-gated item hidden) is the right direction, `requiresRole?: readonly Role[]` keeps a typo a compile error, and the `does not drop its sibling items` test pins the one thing a naive filter would get wrong. I verified the gate's own claim: all eight `@Roles` on the three bench controllers really are `('admin', 'operator', 'packer')`, so the allowlist matches the API exactly. What I cannot sign off on is the surrounding context — a stale docblock that now actively misleads, a test plan whose stated CI mitigation does not exist, and the fact that "Pack bench" lands in a sidebar where every neighbouring entry 403s for the role this item exists to serve.

### Findings

**[IMPORTANT]** — `apps/web/src/app/routes/bench.route.tsx:27-32`
> The `## No nav entry` section is now false in all three of its clauses, and it is a forward reference the next reader will follow:
> > *"A packer opens this at a terminal by URL. #2416 revisits that when the bench has work to show — a nav entry would also need `RoleValues` widened, since it is `['admin','operator']` today."*
>
> There is now a nav entry; #2416 is **CLOSED** and shipped without adding one (its own ACs kept "no global nav, no links out", which is about the bench surface, not about pointing at it — but a reader will not untangle that from a closed issue); and `RoleValues` was widened by the PR below this one. `docs/frontend-architecture.md § UX Mockups` states the governing principle for exactly this — *"a stale mockup is worse than none, because it actively misleads"* — and this file is denser than a mockup, because it also pins a placement invariant. Update the section in this PR: it is the one file a future reader will open to ask "why is the bench not in the nav?".

**[IMPORTANT]** — process, both files
> **CI has never run on this PR.** `.github/workflows/ci.yml` triggers on `pull_request: branches: [main]`; this targets `3107-widen-role-values`, and `gh pr checks 3204` reports *"no checks reported on the '3108-pack-bench-nav-entry' branch"* — no Lint, no Type Check, no Test. The test plan's *"Full-repo `pnpm -r lint` pre-commit step skipped (documented local-machine limitation — CI will run it)"* is therefore not true: nothing runs it until `3104-oms-onboarding-ui` opens a PR against `main`, by which point several stacked PRs have already merged into it. The stacked shape itself is fine and has precedent (#3063), but the stated mitigation has to be corrected — either run the full lint locally, or say plainly that the first CI gate is the epic→main PR and that a lint break will surface there with N commits stacked on top of it.

**[IMPORTANT]** — `apps/web/src/app/nav-registry.ts:58` (the item itself)
> `packer` appears in `@Roles(...)` on **exactly three controllers in the whole API** — the three bench ones. I checked: every orders read is `@Roles('admin', 'operator', 'viewer')`, and #2413 narrowed 45 `@AnyRole()` routes specifically to exclude this role. So a packer who logs in today already sees the entire ungated Operations group — Orders, Products, Inventory, Listings, Analytics, Invoices, Shipments — and **every one of them 403s on the first request it makes**. That is verbatim the defect `requiresPermission` was introduced to fix, quoted in `nav-registry.types.ts:42-46` about a `viewer` and `/automations`; here it is not one entry, it is all of them.
>
> This is **pre-existing on `main`, not a regression in this PR** — but this is the PR that makes the sidebar role-aware, it has the exact mechanism in hand, and it ships one working entry beside a dozen broken ones. At minimum this needs an issue against epic #3104 saying so; ideally the epic's plan names which of the two shapes closes it (per-item `requiresRole` on the Operations entries, or excluding `packer` from the group), because the first is blocked by `viewer` deliberately being absent from the FE `Role` union — see the review on #3202.

**[SUGGESTION]** — `apps/web/src/app/app-shell.tsx:268` vs `apps/web/src/app/command-palette-provider.tsx:203`
> The visibility *rule* is now shared; the *input* to it is duplicated in two non-identical forms — `isReady && session.status === 'authenticated' ? session.user?.role : undefined` in the shell, and the same expression without `isReady` in the palette. They are equivalent today (`Session` is a two-value union and is `'anonymous'` until the adapter resolves), so this is not a bug. But the PR's own justification for extracting `isNavItemVisible` is *"two independent implementations of this check is how ⌘K became a way around whichever one drifted"* — and that argument applies one level up to the derivation feeding it. Export a `navRoleOf(session)` (or drop the redundant `isReady &&`) so the two cannot diverge.

**[SUGGESTION]** — `docs/frontend-architecture.md § Access Control And UI Visibility`
> That section still says the session's `permissions[]` array *"is the only authorization input the frontend reads"* and that `session.user.role` *"must never be compared inline"*, and its three-primitive table has no row for a role gate. This PR makes item-level role comparison a first-class nav primitive. The typing honours the *spirit* of the rule — a typo'd role is a compile error, which is precisely the hazard that sentence names — but the doc now under-describes the app. Add the row and the one-line reason (`packer`'s `ROLE_PERMISSIONS` grant is empty by design, so no permission exists to gate on).

### Positive observations

- Wiring ⌘K in the same change rather than leaving it for a follow-up is the right call and is argued from the right precedent — the palette was already the documented way around the permission gate.
- The async-session timing note in `command-palette-provider.test.tsx:224` is worth its length: tests that passed for the wrong reason (every gated item hidden by the *unresolved* session rather than by the role check) are the hardest kind to notice, and awaiting an independent existing signal before asserting an absence is the right fix rather than an arbitrary wait.
- `isNavItemVisible` unit coverage includes the both-gates-declared case in all three combinations, which is the arm most likely to be got wrong later.

### Verdict

🔄 **Approve with changes**

1. Update `bench.route.tsx`'s `## No nav entry` section — it is the most likely thing to mislead a future reader, and it is a two-minute fix in this PR.
2. Correct the test plan's CI claim, or run the full lint locally.
3. File the packer-sees-a-dozen-403ing-entries gap against epic #3104. Not this PR's to fix; very much this PR's to notice.

(Posted as a comment rather than an approval because of the IMPORTANTs, not because anything here is unsafe to merge into the epic branch.)

---

norbert-kulus-blockydevs:
## Review addressed — `9cb55b16`

All five findings actioned. One of them turned out to be **materially wrong in the direction that changes the fix**, and I have filed the issue you asked for with the corrected numbers rather than the ones in the review.

---

### IMPORTANT · `bench.route.tsx`'s `## No nav entry` — rewritten

Confirmed all three clauses are now false, including the one I could not have known without checking: **#2416 closed 2026-09-05**.

The section is now `## There IS a nav entry now, and it points one way only`. It also preserves the half that did *not* change and is the easiest thing to lose in a rewrite — the entry points AT the bench and the bench links nowhere, which is load-bearing rather than incidental, because the idle lock clears the session and any link out would bounce a half-verified parcel to `/login`. The superseded text is quoted with each clause's correction, so a reader arriving from a stale link sees why it changed rather than a silent overwrite.

### IMPORTANT · the duplicated role derivation — `navRoleOf`, and the `isReady` question answered rather than assumed

You were right that the argument applies one level up. Both sites now call `navRoleOf(session)`.

On whether `isReady` was redundant, I checked rather than reasoned:

```ts
// session-provider.tsx:26-31
const refreshSession = useCallback(async () => {
  const nextSession = await adapter.getSession();
  setSession(nextSession);
  setIsReady(true);          // same callback → React batches
  ...
```

Same callback, so no render observes `status === 'authenticated'` while `isReady` is false; `clearSession` only moves back to anonymous. So it is genuinely redundant, not merely usually-equivalent — and an unresolved session already yields `undefined`, which `isNavItemVisible` fails **closed** on.

The helper takes no `isReady` and its docblock says why, plus where the condition *would* belong if a future provider ever restored a session optimistically before validating — so the next reader does not "restore" it. Three tests cover it, including the anonymous-start case asserted end-to-end through `isNavItemVisible` rather than on the helper alone.

### SUGGESTION · `docs/frontend-architecture.md` — fourth row added

The table has its row, with the reason (`packer`'s `ROLE_PERMISSIONS` grant is empty **by design**, so no permission exists to gate on) and an explicit note that `requiresPermission` stays the default.

It also states why this does not breach the never-compare-inline rule, because the reason is exactly the hazard that rule names: the role is declared as `readonly Role[]` **data** and compared in one function, so a typo is a compile error rather than a silently-false expression. A hand-rolled `session.user.role === 'packer'` at a call site stays banned.

The section also records #3202's `GroupRoleGate` narrowing — it is one block, so splitting it across the stack would only produce a conflict.

### IMPORTANT · CI — corrected, and the local gate actually run

Verified: `ci.yml` is `pull_request: branches: [main]`, `gh pr checks 3204` reports nothing. The stated mitigation did not exist.

Ran what this machine can run, rather than only rewording:

- full `eslint 'apps/web/src/**/*.{ts,tsx}'` — **0 errors** (123 pre-existing warnings)
- `nav-registry.test.ts` 21 passed; `command-palette-provider.test.tsx` + `app-shell.test.tsx` + `merge-nav-contributions.test.ts` 48 passed

The body now says plainly that the first CI gate is the epic → `main` PR.

---

### IMPORTANT · the packer-403 gap — filed as **#3221**, with a correction

Filed under epic #3104, as you asked. But I measured the guard on each item's primary read before writing it, and **"every one of them 403s" is not what the code does**:

| Nav item | Guard on its primary read | Packer |
|---|---|---|
| Analytics · Insights · Orders · Customers · Fulfilment | `@Roles('admin','operator','viewer')` | **403** |
| Products · Listings · Shipments · Returns | `@AnyRole()` | **loads** |
| Automations | `requiresPermission: 'automations:read'` | already hidden |
| Invoices | mixed (`@AnyRole()` on some reads) | partial |
| **Pack bench** | `@Roles('admin','operator','packer')` | loads |

`@AnyRole()` admits `packer`, so five entries load fine and `Automations` is already correctly hidden by the gate #2358 added for precisely this.

**This changes the remedy**, which is why it is worth the correction rather than a footnote. Your review offers two shapes and leans toward the second being cleaner; but excluding `packer` from the Operations group wholesale would hide five entries that work **and the Pack bench entry this epic exists to add**. So it is per-item `requiresRole` on a named subset of five or six — and that confirms the other half of your read: it is blocked on `viewer` being absent from the FE `Role` union, which #3202's `viewer` paragraph now records as this issue's known consumer.

Also confirmed and left alone: this is pre-existing on `main`, and #3221 explicitly scopes out changing any `@Roles(...)`. Whether a packer *should* read orders is a product decision, not a nav one.

---

### One thing I verified rather than assumed, because your first review blocked on it

`FulfillmentExecutor` never being enabled. The rebuilt copy says the operator does **not** need to visit the Capabilities panel, which contradicts the earlier revision. It is correct:

- `libs/oms/src/oms.plugin.ts:72` — `supportedCapabilities: ['FulfillmentExecutor']`, no `InventoryMaster`
- `connection.service.ts:449-452` — `enabledCapabilities = rest.enabledCapabilities ?? [...metadata.supportedCapabilities]`
- `create-connection.schema.ts:144-146` — the form **omits** the key entirely when the capabilities field is blank

So a form-created OMS connection is stamped `FulfillmentExecutor` automatically. Your first blocker is genuinely closed, not worked around.

piotrswierzy (approved):
## Tech Lead re-review (`b43c46d1` → `9cb55b16`)

All three IMPORTANTs and both SUGGESTIONs are closed, and the range also carries #3202's `GroupRoleGate` narrowing up the stack. The two that mattered most are closed in the right way rather than the cheap way: `bench.route.tsx`'s stale section is not merely deleted but *corrected in place*, naming each of its three false clauses and why #2416 closing without a nav entry is not the contradiction it looks like — which is precisely the untangling I said a reader would not recover from a closed issue. And the duplicated role derivation is now one exported `navRoleOf(session)`, with the `isReady` question answered by argument rather than by dropping it quietly.

### Findings

**Closed — [IMPORTANT]** `bench.route.tsx:27` — the `## No nav entry` docblock
> Rewritten as `## There IS a nav entry now, and it points one way only`. It also keeps the invariant the old section carried by accident — the bench links nowhere, and the reason is now stated (the idle lock clears the session, so a link out would bounce a half-verified parcel to `/login`) rather than left as an absence a later contributor would "fix".

**Closed — [IMPORTANT]** process / CI claim
> Corrected honestly in the body, same wording as #3202. Re-confirmed independently: `gh pr checks 3204` still reports `no checks reported on the '3108-pack-bench-nav-entry' branch`, and `ci.yml` still triggers on `pull_request: branches: [main]` only. Note `route-authorization-coverage.spec.ts` landing on `main` changes nothing here — it is a backend guard, and no job runs against this branch to execute it either way.

**Closed — [IMPORTANT]** `nav-registry.ts:58` — a packer sees a dozen 403ing entries
> **#3221** is open (`[BUG] Frontend — a packer sees Operations nav entries whose API reads 403`), and the constraint that blocks the obvious fix — `viewer` deliberately absent from the FE `Role` union — is recorded at the point of cause in `nav-registry.types.ts`'s `RoleValues` docblock, not only in the issue. That is the shape I asked for: this PR's job was to notice, not to fix.

**Closed — [SUGGESTION]** the duplicated role derivation
> `navRoleOf` is now the single input to `isNavItemVisible` for both the sidebar and ⌘K. I verified the docblock's central claim against `session-provider.tsx` rather than taking it: `refreshSession` really does call `setSession(next)` and `setIsReady(true)` in one callback (lines 28-29), and `clearSession` only moves back to `ANONYMOUS_SESSION`, so no render can observe `status === 'authenticated'` with `isReady` false and the omitted condition is redundant rather than merely usually-true. The docblock names the exact future change that would invalidate it (an optimistic restore before validation) and says the fix belongs in this function, not its call sites — which is the property extracting it buys.

**Closed — [SUGGESTION]** `docs/frontend-architecture.md § Access Control And UI Visibility`
> Fourth row added, and — better — the opening sentence is amended from "the only authorization input" to "wherever a permission exists to read", so the new row is an *allowed* exception rather than a contradiction sitting under an absolute claim. The paragraph explaining why the typing honours the never-compare-inline rule (declared as `readonly Role[]` data, compared in one function, so a typo is a compile error) is the right defence, and it correctly keeps a hand-rolled `session.user.role === 'packer'` banned at call sites.

**[SUGGESTION]** — `apps/web/src/app/app-shell.tsx:253,259`
> `isAdmin` and `isViewerOnly` still hand-roll `isReady && session.status === 'authenticated' && session.user?.role === '…'`. Pre-existing, explicitly carved out by the doc, and a different question from the nav gate (booleans for shell behaviour, not item visibility) — so not this PR's. But `navRoleOf` is now the obvious place those would converge, and it is worth knowing that the file contains both shapes. No action here.

### Positive observations

- The bench docblock's correction is dated and cites the governing principle (`§ UX Mockups`: a stale note is worse than none). A future reader gets the reasoning, not just the current state.
- Amending the doc's opening sentence instead of appending a row under an absolute claim is the difference between documentation that describes the app and documentation that is now internally inconsistent.
- The `navRoleOf` docblock's `## Why there is no isReady parameter` section is the kind of negative-space documentation that survives a refactor: it makes removing the condition a deliberate, checkable decision rather than an omission.

### Verdict

✅ **Approve**

---

piotrswierzy (approved):
## Tech Lead re-verification (head unchanged at `9cb55b1`)

Head and `main` unchanged. I re-checked the two things the approval rested on and the CI state, nothing else.

**#3221 is filed, open, and it is evidence rather than an assertion.** Created 2026-09-13, it carries a per-item table of the *measured* guard on each Operations nav item's primary read — five answer 403 for a `packer` (Analytics, Insights, Orders, Customers, Fulfilment), four load, Invoices mixed. That is the right shape for a defect deliberately not absorbed into this PR. The blocking constraint is recorded at the point of cause, not only in the issue: the `RoleValues` docblock in `nav-registry.types.ts` (`:54-61`) states that the backend's `UserRoleValues` also carries `'viewer'`, that it is deliberately absent from the FE union, and names exactly what the fix would need (`requiresRole: ['admin','operator','viewer']` on those entries) — so the next reader meets the constraint where they would otherwise re-derive it.

**`navRoleOf` drops `isReady` soundly.** The function (`nav-registry.ts:157`) is `session.status === 'authenticated' ? session.user?.role : undefined`, and its docblock argues redundancy rather than optionality. That argument holds against the source: `shared/auth/session-provider.tsx:28-29` calls `setSession(nextSession)` and `setIsReady(true)` in the same callback, so no render observes `authenticated` while `isReady` is false, and `clearSession` only moves back toward anonymous. An unresolved session yields `undefined`, which `isNavItemVisible` (`:125`) treats as no-role-known and fails **closed**. The docblock also names itself — not its two call sites — as where an extra condition would belong if a future provider restored a session optimistically, which is the correct placement. Pack bench is gated `['admin','operator','packer']` (`:62`).

**CI: absent, reported as absent.** `GET /commits/9cb55b1…/check-runs` → `total_count: 0`. See my note on #3202 for why the workflow-file explanation in the body does not fully account for that (sibling stacked PRs in the #3173 chain do get runs); the practical consequence — no Lint/Type Check/Test has run against this head — is unchanged.

**Stack**: `3104-oms-onboarding-ui` ← #3125 ← #3202 ← **#3204**. Blocked behind #3202, which is blocked behind #3125, which is blocked behind the epic branch reaching `main`. Nothing blocks it on its own merits (`mergeable: CLEAN`, `APPROVED`).

### Findings

### Verdict
✅ **Approve**

---

norbert-kulus-blockydevs:
## Rebased — conflict resolved, `9cb55b16` → `4fbadab9`

Force-push. History rewritten, **content unchanged**.

### What the conflict actually was

This branch carried its own older copies of #3202's two commits plus the merge that reconciled them:

```
fc739c4e1  widen RoleValues          ← duplicate of #3202's a49248e39
b43c46d1d  Pack bench nav entry      ← this branch's own work
0e309703d  unhonoured role gates     ← duplicate of #3202's df2ab2057
8fc6e4402  Merge origin/3107 …
9cb55b166  correct the stale bench nav note …   ← this branch's own work
```

Both duplicate pairs are `git patch-id`-**identical**, so the conflict was the same change arriving twice under two SHAs rather than a real disagreement. The rebase dropped both (`skipped previously applied commit`) and flattened the merge.

### The one real conflict

`apps/web/src/app/nav-registry.test.ts`, one import line — both sides had added a symbol:

- new base: `import { BASE_NAV_GROUPS, buildNavGroups }`
- `b43c46d1d`: `import { buildNavGroups, isNavItemVisible }`

Resolved as the union. `navRoleOf` is deliberately **not** in it at that commit — it arrives in `9cb55b166`, which reapplied cleanly on top and produces the same final import line the old merge did. `nav-registry.types.ts` auto-merged.

### Verification

The claim worth checking is that the rebase changed no code, and it is checkable directly:

```
git diff --stat 9cb55b166 4fbadab95
 docs/plans/mockups/oms-onboarding-3104.html | 235 ++++++++++++++++++-------
 1 file changed, 195 insertions(+), 40 deletions(-)
```

**One file, and it is the mockup** — i.e. #3125's new commit flowing up the stack. Every `apps/web` byte is identical to the reviewed `9cb55b166`, which is stronger evidence the resolution is right than a type-check would be.

Also checked at the intermediate commit `4704c915` (the one I resolved): all three imported symbols are exported by `nav-registry.ts` there (`BASE_NAV_GROUPS:30`, `isNavItemVisible:121`, `buildNavGroups:168`) and all three are used in the file (2 / 9 / 14 refs).

### Stack

Linear again, all four commits GPG-signed (`%G? = G`):

```
3104-oms-onboarding-ui
└─ 0a197d6a  #3125  mockup
   ├─ 6cbd4745  #3202  widen RoleValues
   ├─ b7ed5199  #3202  unhonoured role gates
   ├─ 4704c915  #3204  Pack bench nav entry
   └─ 4fbadab9  #3204  correct the stale bench nav note
```

`#3125`, `#3202`, `#3204` all report `MERGEABLE / CLEAN`. Pre-rebase SHAs recorded, so this is recoverable if anything looks wrong.

No CI ran — `ci.yml` triggers on `pull_request: branches: [main]` only, so nothing runs on a stacked branch. Absent, not passing.

piotrswierzy (commented):
## Tech Lead re-verification (head unchanged at `9cb55b16`)

Approved content re-verified and intact — but **the merge state has changed and this PR can no longer merge as it stands.**

`navRoleOf` and the #3221 filing both hold: **#3221 is OPEN** ("[BUG] Frontend — a packer sees Operations nav entries whose API reads 403"), carrying the blocking constraint at the point of cause, exactly as approved.

### Findings

**[IMPORTANT — new] `CONFLICTING`/`DIRTY` against its own base.** `git compare df2ab205...9cb55b16` reports `status=diverged, ahead=5, behind=3, merge_base=d82d6b08`. The history explains it: this branch already merged an **earlier** `origin/3107-widen-role-values` at `8fc6e440`, and #3202 subsequently gained 3 commits — the review round that introduced `GroupRoleGate`. Both branches then implemented that narrowing independently, on the same lines of the same four files (`.eslintrc.js`, `nav-registry.types.ts`, `shared/plugins/plugin.types.ts`, `nav-registry.test.ts`).

The good news, and the reason this is IMPORTANT rather than BLOCKING: **the two implementations already agree**, so the resolution is mechanical rather than a design decision. At this head the branch carries `export type GroupRoleGate = 'admin';`, `requiresRole?: GroupRoleGate` on the group gate, `requiresRole?: readonly Role[]` on the item gate (correct — that one *is* honoured), and the identical `GroupRoleGate$`-for-`Role$` pinhole swap in `.eslintrc.js`. Merge `3107-widen-role-values` in, resolve by intent, then re-grep the invariant rather than trusting the textual resolution:

- `.eslintrc.js` contains `GroupRoleGate$` and **no** `|Role$|` in that lookahead
- `GroupRoleGate = 'admin'` declared once, applied at both group-gate sites
- `LiveNavItem.requiresRole?: readonly Role[]` survives the merge unchanged

**[IMPORTANT — merge path]** The orphan-base finding I have recorded on #3202 applies here too: the chain is `3104-oms-onboarding-ui` → #3125 → #3202 → **#3204**, and `3104-oms-onboarding-ui` is `ahead=0, behind=17` of `main` with **no PR in any state**. Nothing in this sub-chain has a terminus today. Retargeting #3125 to `main` is a zero-content-change fix and is a prerequisite for queueing either PR. (#3125 itself is now `CLEAN`/`APPROVED` — the `CHANGES_REQUESTED` that `gh pr list` still renders was superseded on 2026-09-14.)

**[IMPORTANT — CI]** Still **zero** runs, ever (`total_count=0` for `3108-pack-bench-nav-entry`). Absent, not passing. As on #3202, I am withdrawing my earlier `pull_request: branches: [main]` explanation — run [`34831266177`](https://github.com/openlinker-project/openlinker/actions/runs/34831266177) is a `pull_request` CI run with base `3175-sales-document-adr`, so that filter demonstrably does not suppress non-`main` bases here. The absence is real and unexplained; rebasing onto a `main`-terminated chain is the most likely thing to produce a run.

### Verdict
🔄 **Approve with changes** — the diff stands as approved; merge the current `3107-widen-role-values` tip and re-grep the three invariants above before queueing.

---

Issue #3207, opened 2026-09-12
[TASK] Frontend — connection form field for the stock-location override


norbert-kulus-blockydevs:
## Problem / Context

#3206 adds a per-connection `Connection.config.stockLocationOverride` config key so an operator can assert "all of this connection's currently-unlocated stock physically lives in location X" — unblocking the OMS fulfilment router for the common single-warehouse PrestaShop/WooCommerce install (see #3206 for the full backend rationale). There is currently no UI to set this key; it would otherwise only be reachable via raw JSON config editing or curl, matching the gap the existing `StockAndPricingSection` (#2610) was built to close for `stockSafetyBuffer` and `pricingRule`.

## Proposed Solution

Add a third field to the existing `StockAndPricingSection` (`apps/web/src/features/connections/components/stock-and-pricing-section.tsx`), following the same pattern already used for `stockSafetyBuffer`/`pricingRule` in that same section (toggle-gated group, worked example, mirrored client-side preview per `apps/web/src/features/connections/lib/stock-and-pricing-preview.ts` and `scripts/check-stock-and-pricing-preview-mirror.mjs`):

- A location picker (reuse the `GET /inventory/locations` list — same source the `/settings/inventory-locations` screen from #3063 uses), writing `config.stockLocationOverride`.
- **Copy must state this is an ASSERTION, never a detection** — something like: *"Assign a location to this connection's stock — I attest all of this connection's currently un-located stock physically lives here."* An operator with genuinely multiple physical warehouses behind one shop must be actively discouraged from using this (the field's own help text should say so), since setting it incorrectly would misattribute stock location to the fulfilment router.
- A cleared value writes an explicit `null` (never deletes the key), matching the existing `stockSafetyBuffer`/`pricingRule`/`rateLimit` shallow-spread convention (#2016).
- Server-side validation (the connection PATCH/POST rejecting an unknown location id) is #3206's responsibility — this issue's form only needs to surface that 400 as a field-level error, matching how other `StockAndPricingSection` fields already do.

## Classification

**Type**: Frontend
**Layer**: Interface (feature module)
**File(s)**:
- `apps/web/src/features/connections/components/stock-and-pricing-section.tsx`
- `apps/web/src/features/connections/lib/stock-and-pricing-preview.ts`
- `apps/web/src/features/connections/components/stock-and-pricing-section.test.tsx`

## Dependencies

[...]

Issue #2322, opened 2026-08-22
[TASK] CORE — enforce "locationId IS NULL means the master declines to locate"


piotrswierzy:
> Part of the OMS programme backlog (**Wave 1b**, stream S1, size M). Backlog slug: `W1b-10`. Sibling `W*` slugs resolve via the Wave-1b epic's checklist. Design of record: `docs/plans/analysis/DESIGN-oms-authority-model.md` + ADRs 052–062.

## Problem / Context

ADR-058 decision (2) gives `locationId IS NULL` a permanent distinct meaning — "the master declines to locate its stock", **never** "the default location". Same-source NULL + non-NULL coexistence is a contradiction (one source cannot both decline to locate and locate); cross-source coexistence is legitimate (a shop reports a pool, a 3PL reports a depot; ATP sums both). Without enforcement, a source that starts locating leaves its old pooled row alive and ATP double-counts.

## Proposed Solution

In `MasterInventorySyncService`: once a source writes a located position for a product, its own locationless row for that product/variant is staled (the existing staleness mechanism, scoped by `sourceConnectionId` from W1b-5). Cross-source NULL/non-NULL rows are left alone and summed.

Migrating NULL rows to a synthetic DEFAULT location is explicitly rejected by ADR-058 (it rewrites the unique-index surface on the live path and asserts a location fact OL does not have) — this issue must not do that.

## Classification

**Type**: CORE
**Layer**: Application
**File(s)**: `libs/core/src/inventory/application/services/master-inventory-sync.service.ts`, `libs/core/src/inventory/infrastructure/persistence/repositories/inventory.repository.ts`

## Dependencies

- #2320 (`W1b-5`)

## Assumptions

- The transition is one-way in practice (a source that starts locating does not go back); if it does, the located rows stale by the ordinary prune and the pooled row is re-created.

## Acceptance Criteria

- [ ] A source's first located write stales that source's locationless row for the same product/variant (integration test)
- [ ] A second source's locationless row is untouched and still summed into ATP (integration test)
- [ ] No row is migrated to a synthetic default location
- [ ] Behaviour is a no-op for a source that never locates
- [ ] Tests added or updated for non-trivial logic
- [ ] No architecture boundary violations (CORE ↔ Integration)

PR #3208, opened 2026-09-12
feat(inventory): per-connection stock-location override for masters that never report one

### LINKED ISSUE #3206: [TASK] CORE — per-connection stock-location override for masters that never report one
## Problem / Context

Neither shipped `InventoryMasterPort` adapter ever reports a `locationId`: `PrestashopInventoryMasterAdapter`'s mapper sets `locationId: undefined` (`libs/integrations/prestashop/src/infrastructure/mappers/prestashop-inventory.mapper.ts:33`, "PrestaShop doesn't support multi-location in stock_available") and `WooCommerceInventoryMasterAdapter` does the same (`libs/integrations/woocommerce/src/infrastructure/adapters/inventory-master/woocommerce-inventory-master.adapter.ts:608`, "WC is single-location at v1"). Per ADR-058 decision 2, a `NULL` `locationId` on `inventory_items` permanently means "the master declines to locate its stock" — so on every real install today, 100% of inventory rows carry `locationId === null`.

This makes the OMS Wave 3 fulfilment router (`libs/oms/src/routing/ol-fulfillment-router.ts`) structurally sterile even after an operator creates locations: `ConnectionService.assertRouterEnablementPreconditions` (`apps/api/src/integrations/application/services/connection.service.ts:150`) only gates on `countActiveLocations() > 0`, which is satisfiable — but `loadStock()` (`ol-fulfillment-router.ts:183`) explicitly skips every row with `locationId === null` when building the stock map the router evaluates candidates against:

```ts
if (locationId === null) {
  continue;
}
```

Result: an operator can create a location, enable routing, and every order line still resolves `unfulfillable`, because the router never sees any stock anywhere. Confirmed by tracing the code, not assumed.

The two masters we ship (PrestaShop, WooCommerce) cannot be made to report real per-warehouse locations for the common case — WooCommerce has no native multi-location feature at all, and PrestaShop's only native path is the optional "Advanced Stock Management" module, rarely enabled (separate, narrower follow-up if ever pursued — out of scope here). The typical install has exactly ONE physical location behind the shop; the gap is that nothing lets the operator say so.

## Proposed Solution

A new per-connection config key that an operator sets explicitly, applied only where the master itself declined to say anything — never overriding a real adapter-reported location.

1. **New pure coercion type**, mirroring the existing `readStockSafetyBuffer`/`readPricingRule` precedent (`libs/core/src/identifier-mapping/domain/types/stock-safety-buffer.types.ts`, `pricing-rule.types.ts`) but placed in the `inventory` context (its only consumer), at `libs/core/src/inventory/domain/types/stock-location-override.types.ts`:
   - `STOCK_LOCATION_OVERRIDE_CONFIG_KEY = 'stockLocationOverride'` on `Connection.config` (JSONB — no migration).
   - `readStockLocationOverride(config: ConnectionConfig | null | undefined): string | null` — a defensively-coerced locationId string, or `null` if absent/invalid.
   - No `apply*` function needed here (unlike `stockSafetyBuffer`) since this isn't a quantity transform — it's a fallback value, applied inline at the one seam below.

2. **One seam, in `MasterInventorySyncService.toDomainInventoryItem`** (`libs/core/src/inventory/application/services/master-inventory-sync.service.ts:596`), where `inventory.locationId ?? null` is currently computed twice (once for the `existing` lookup, once for the constructed `InventoryItemDomainEntity`). Change both to resolve the connection's override and use it only as a fallback:
   ```ts
   const effectiveLocationId = inventory.locationId ?? overrideLocationId ?? null;
   ```
   `overrideLocationId` is read once per sync call via `readStockLocationOverride(connection.config)` — the service will need the `Connection` (or just its `config`) threaded into `syncOneFromMaster`/`toDomainInventoryItem`; today only `connectionId` (a string) reaches this call chain, so resolving the connection via the already-injected `IIntegrationsService.getAdapter(connectionId)` (or threading `connection.config` down from wherever the caller already has it — check `syncFromMasterByExternalId(s)` call sites) is required. Prefer resolving once per sync batch, not once per inventory row, to avoid N redundant connection reads.

3. **This does not violate ADR-058 decision 2.** That rule forbids the *sync* inventing a location the master declined to give. Here the *operator* supplies the fact once, explicitly, per connection — the sync just applies it consistently, exactly as it already applies `stockSafetyBuffer` and `pricingRule`. From the sync's own perspective this looks identical to "the master started reporting a location", so the entire existing ADR-058 repair machinery needs zero changes:
   - `markLocationlessStaleForSource` (#2322's pooled→located repair) correctly stales the connection's old `null`-keyed rows once the override kicks in.
   - #2320's provenance-scoped lookup (`findByProductAndVariant`) is unaffected — `sourceConnectionId` is unchanged, only the `locationId` argument changes.
   - #1904's rival-claimant guard is unaffected — provenance is unchanged.
   - A *real* adapter-reported location always wins (`inventory.locationId ?? override`, never the reverse), so if a future adapter (e.g. PrestaShop Advanced Stock Management) starts reporting genuine locations, the override becomes a no-op automatically with no migration needed.

4. **Validation at write time** (`ConnectionService.create`/`update`, alongside the existing `validateRateLimitConfig`/pricing-rule bounds checks per #2610): if `stockLocationOverride` is set, it must reference a location returned by `GET /inventory/locations` for THIS install (no cross-tenant leakage) — reject with 400 naming the unknown id, same neutral-bounds-check style as `stockSafetyBuffer`/`stockZeroThreshold`.

## Classification

**Type**: CORE
**Layer**: Domain + Application
**File(s)**:
- `libs/core/src/inventory/domain/types/stock-location-override.types.ts` (new)
- `libs/core/src/inventory/application/services/master-inventory-sync.service.ts` (the one seam)
- `apps/api/src/integrations/application/services/connection.service.ts` (create/update validation)

## Dependencies

- #2313/#2316 (`inventory_locations` table + CRUD API) — already shipped.
- None on #1030 — this is explicitly NOT that issue; it does not require the "should OL become an inventory master" product decision.

## Assumptions

- The override applies per-connection, not per-product/variant — a connection asserts ONE location for all of its currently-unlocated stock. Multi-warehouse operators behind one connection are explicitly NOT served by this (they need real per-warehouse reporting from the master, or a native OL inventory master — out of scope, see #1030).
- No migration or backfill job is added: existing `NULL`-located rows for a connection that sets the override are corrected by the next ordinary inventory sync cycle (~15 min cadence today), reusing #2322's existing repair path. If this proves too slow in practice, a one-time backfill pass is a separate, smaller follow-up.
- The companion frontend field (a location picker + assertion-not-detection copy on the connection edit form) is filed as its own issue since it's a separate layer, per the FE/BE split rule.

## Acceptance Criteria

- [ ] `readStockLocationOverride` correctly coerces a present-and-valid, absent, and invalid (non-string, references a nonexistent location) config value
- [ ] `MasterInventorySyncService` resolves `inventory.locationId ?? override ?? null` at both existing `?? null` sites in `toDomainInventoryItem`
- [ ] A real adapter-reported `locationId` always wins over the override (regression test: adapter reports a location, override is also set, the adapter's value is written)
- [ ] `ConnectionService.create`/`update` reject a `stockLocationOverride` value that doesn't resolve to an existing `inventory_locations` row for this install, with a 400 naming the unknown id
- [ ] Setting the override on a connection whose stock was previously pooled (`locationId IS NULL`) results in the pooled row being staled and a new located row appearing after the next sync cycle (int-spec against real Postgres, following the #2322 int-spec pattern)

norbert-kulus-blockydevs:
## Summary
- Adds `Connection.config.stockLocationOverride`: an operator-declared location id for a connection whose master never reports one (PrestaShop only via the rarely-enabled Advanced Stock Management module, WooCommerce not at all).
- New pure coercion helper `readStockLocationOverride` at `libs/core/src/inventory/domain/types/stock-location-override.types.ts`, mirroring the `readStockSafetyBuffer`/`readPricingRule` config-coercion precedent.
- Applied at exactly one seam — `MasterInventorySyncService.toDomainInventoryItem` — as `inventory.locationId ?? override ?? null`. A real adapter-reported location always wins, so the override becomes a silent no-op the day an adapter starts reporting genuine locations.
- The override resolves **once per sync call** (batch or single), never once per inventory row.
- `ConnectionService.create`/`.update` validate the override resolves to an existing `inventory_locations` row for this install (400 on an unknown id) — reusing the `ILocationService` already injected there for the #2407 routing-enablement guard.
- `scripts/check-architecture-gates.mjs`'s config-knob threshold raised 7 → 8, with a recorded rationale documented inline: this knob asserts a physical-world fact no future routing/authority rules engine (#2169) could express, so it isn't the kind of accretion that gate watches for.

## Why this doesn't violate ADR-058 decision (2)
That decision forbids the **sync** inventing a location the master declined to give. Here the **operator** supplies the fact, once and explicitly, per connection. From the sync's own perspective an override switching on looks identical to "the master started reporting a location", so the existing `#2322` pooled-row repair, `#2320`'s provenance-scoped lookup, and `#1904`'s rival-claimant guard all apply **unchanged** — verified by tests that this override correctly triggers the #2322 repair path exactly as a real adapter-reported location would.

## Test plan
- [x] `pnpm --filter @openlinker/core type-check` / `pnpm --filter @openlinker/api type-check` — both clean
- [x] `pnpm check:invariants` — all green (including the raised knob threshold)
- [x] New unit tests: `stock-location-override.types.spec.ts` (5 cases), `master-inventory-sync.service.spec.ts` (+6 cases: fold-into-pooled, #2322-repair trigger, real-location-wins, default-unchanged, once-per-batch), `connection.service.spec.ts` (+5 cases: accept known location, reject unknown, reject non-string, absent-is-untouched, same rejection on update)
- [x] Targeted `eslint` on all changed files — clean
- [ ] Manual: create a PrestaShop/WooCommerce connection with unlocated stock, set `stockLocationOverride`, confirm the next sync writes located rows and the fulfilment router now sees stock at that location

## Follow-up
The frontend field for operators to actually set this (a location picker in the connection's `StockAndPricingSection`, #2610's precedent) is filed as its own issue, #3207, since it's a separate layer per the FE/BE split rule and has no server contract to call against until this lands.

Closes #3206

### DIFF
--- a/apps/api/src/integrations/application/services/connection.service.spec.ts
+++ b/apps/api/src/integrations/application/services/connection.service.spec.ts
@@ -218,6 +218,10 @@ describe('ConnectionService', () => {
     const mockLocations = {
       countActiveLocations: jest.fn().mockResolvedValue(1),
       bootstrapDefaultLocations: jest.fn(),
+      // #3206 — `validateStockLocationOverride` only calls this when the
+      // config key is present, so a `null` default costs nothing for every
+      // existing test that never sets `stockLocationOverride`.
+      getLocation: jest.fn().mockResolvedValue(null),
     } as unknown as jest.Mocked<ILocationService>;
 
     const module: TestingModule = await Test.createTestingModule({
@@ -924,6 +928,63 @@ describe('ConnectionService', () => {
 
         await expect(service.update('connection-123', { config })).resolves.toEqual(mockConnection);
         expect(connectionPort.update).toHaveBeenCalledWith('connection-123', { config });
+      });
+    });
+
+    describe('stockLocationOverride validation (#3206)', () => {
+      it('should accept a create whose override names an existing location', async () => {
+        connectionPort.create.mockResolvedValue(mockConnection);
+        locations.getLocation.mockResolvedValue({ id: 'ol_location_main' } as never);
+
+        await expect(
+          service.create({
+            ...payload,
+            config: { ...payload.config, stockLocationOverride: 'ol_location_main' },
+          })
+        ).resolves.toEqual(mockConnection);
+        expect(locations.getLocation).toHaveBeenCalledWith('ol_location_main');
+        expect(connectionPort.create).toHaveBeenCalled();
+      });
+
+      it('should reject a create whose override names an unknown location', async () => {
+        locations.getLocation.mockResolvedValue(null);
+
+        await expect(
+          service.create({
+            ...payload,
+            config: { ...payload.config, stockLocationOverride: 'ol_location_ghost' },
+          })
+        ).rejects.toThrow(BadRequestException);
+        expect(connectionPort.create).not.toHaveBeenCalled();
+      });
+
+      it('should reject a non-string override value', async () => {
+        await expect(
+          service.create({
+            ...payload,
+            config: { ...payload.config, stockLocationOverride: 42 as unknown as string },
+          })
+        ).rejects.toThrow(BadRequestException);
+        expect(connectionPort.create).not.toHaveBeenCalled();
+      });
+
+      it('should leave an absent override untouched and never call the location lookup', async () => {
+        connectionPort.create.mockResolvedValue(mockConnection);
+
+        await expect(service.create(payload)).resolves.toEqual(mockConnection);
+        expect(locations.getLocation).not.toHaveBeenCalled();
+      });
+
+      it('should reject the same unknown-location value on update', async () => {
+        connectionPort.get.mockResolvedValue(mockConnection);
+        locations.getLocation.mockResolvedValue(null);
+
+        await expect(
+          service.update('connection-123', {
+            config: { baseUrl: 'https://shop.example.com', stockLocationOverride: 'ol_location_ghost' },
+          })
+        ).rejects.toThrow(BadRequestException);
+        expect(connectionPort.update).not.toHaveBeenCalled();
       });
     });
   });
--- a/apps/api/src/integrations/application/services/connection.service.ts
+++ b/apps/api/src/integrations/application/services/connection.service.ts
@@ -299,6 +299,30 @@ export class ConnectionService implements IConnectionService {
     }
   }
 
+  /**
+   * Core-owned validation for `config.stockLocationOverride` (#3206): the id
+   * must resolve to a location this install actually has. Rejecting an
+   * unknown id here — rather than letting the sync silently write against a
+   * location that does not exist — is what keeps the value an operator
+   * assertion rather than a typo nobody notices until stock looks wrong.
+   *
+   * Never defaults a value in; an absent key stays absent (byte-identical to
+   * pre-#3206 behaviour for every connection that never sets it).
+   */
+  private async validateStockLocationOverride(config: Record<string, unknown>): Promise<void> {
+    const value = config.stockLocationOverride;
+    if (value === undefined || value === null) return;
+    if (typeof value !== 'string' || value.trim().length === 0) {
+      throw new BadRequestException('config.stockLocationOverride must be a non-empty string');
+    }
+    const location = await this.locations.getLocation(value);
+    if (!location) {
+      throw new BadRequestException(
+        `config.stockLocationOverride names an unknown location: ${value}`
+      );
+    }
+  }
+
   private async validateCredentialsShape(
     adapterKey: string,
     credentials: Record<string, unknown>
@@ -470,6 +494,7 @@ export class ConnectionService implements IConnectionService {
       if (rest.config !== undefined) {
         this.validateRateLimitConfig(rest.config);
         this.validateStockAndPricingConfig(rest.config);
+        await this.validateStockLocationOverride(rest.config);
         await this.validateConfigShape(metadata.adapterKey, rest.config);
         // #2407 — above the credential-persistence block below, which requires
         // that a 400 from validation never leaves an orphan credential row.
@@ -729,6 +754,7 @@ export class ConnectionService implements IConnectionService {
       if (patch.config !== undefined && metadata) {
         this.validateRateLimitConfig(patch.config);
         this.validateStockAndPricingConfig(patch.config);
+        await this.validateStockLocationOverride(patch.config);
         await this.validateConfigShape(metadata.adapterKey, patch.config);
         // #2407 — inside this branch, which is correct ONLY because
         // `ConnectionRepository.update` REPLACES `config` wholesale rather than
--- a/apps/worker/test/integration/inventory-stock-location-override-e2e.int-spec.ts
+++ b/apps/worker/test/integration/inventory-stock-location-override-e2e.int-spec.ts
@@ -1,0 +1,237 @@
+/**
+ * Stock-Location Override — Pooled-to-Located Transition, End-to-End (#3206)
+ *
+ * The unit tests already pin that `MasterInventorySyncService` resolves
+ * `inventory.locationId ?? override ?? null` and that it feeds the *effective*
+ * (post-override) locationId into the same #2322 pooled/located tracking a
+ * genuine adapter-reported location would take. What none of that proves is
+ * the thing #3206's own acceptance criterion asks for: that turning the
+ * override ON for a connection whose stock is already pooled in real Postgres
+ * actually converges — the old `locationId IS NULL` row ends up `isStale`,
+ * exactly one located row exists at the override's location, and the
+ * available quantity is the original figure, not double-counted in between.
+ *
+ * Modelled directly on the sibling `#2324` end-to-end spec
+ * (`inventory-location-propagation-e2e.int-spec.ts`), which already drives
+ * `MasterInventorySyncService.syncFromMasterByExternalId` against a real
+ * connection + a mocked `InventoryMasterPort` and reads back real rows —
+ * the more faithful precedent for THIS claim than the `#2322`-only
+ * `IInventoryService`-level spec, since the override is applied inside the
+ * sync service itself, not at the repository layer that spec exercises.
+ *
+ * @module apps/worker/test/integration
+ */
+import { getTestHarness, resetTestHarness, teardownTestHarness } from './setup';
+import type { WorkerIntegrationTestHarness } from './setup';
+import { createTestConnection } from './helpers/test-connection.helper';
+import type { IIntegrationsService } from '@openlinker/core/integrations';
+import { INTEGRATIONS_SERVICE_TOKEN } from '@openlinker/core/integrations';
+import type { InventoryMasterPort, IMasterInventorySyncService } from '@openlinker/core/inventory';
+import { MASTER_INVENTORY_SYNC_SERVICE_TOKEN } from '@openlinker/core/inventory';
+import { InventoryItemOrmEntity } from '@openlinker/core/inventory/orm-entities';
+import { ProductOrmEntity, ProductVariantOrmEntity } from '@openlinker/core/products/orm-entities';
+import { IdentifierMappingOrmEntity } from '@openlinker/core/identifier-mapping/orm-entities';
+import { ConnectionOrmEntity } from '@openlinker/core/identifier-mapping/orm-entities';
+import type { DataSource } from 'typeorm';
+import { randomUUID } from 'crypto';
+
+const OVERRIDE_LOCATION_ID = 'warehouse-override';
+
+describe('Stock-location override — pooled-to-located transition end-to-end (#3206)', () => {
+  let harness: WorkerIntegrationTestHarness;
+  let dataSource: DataSource;
+  let integrationsService: IIntegrationsService;
+
+  beforeAll(async () => {
+    // Set BEFORE the container boots, so running this file alone
+    // (`--runTestsByPath`) does not die in `getPiiConfig` before reaching
+    // anything this spec is about — the `oms-module-boot.int-spec.ts` precedent.
+    process.env.OL_PII_HASH_SALT ??= 'test-salt-for-integration-tests';
+
+    harness = await getTestHarness();
+    dataSource = harness.getDataSource();
+    integrationsService = harness.get(INTEGRATIONS_SERVICE_TOKEN);
+  });
+
+  beforeEach(async () => {
+    await resetTestHarness();
+  });
+
+  afterEach(() => {
+    jest.restoreAllMocks();
+  });
+
+  afterAll(async () => {
+    await teardownTestHarness();
+  });
+
+  /** Seeds a product + one variant + its `Product` identifier mapping. */
+  async function seedVariant(
+    connectionId: string
+  ): Promise<{ productId: string; variantId: string; externalId: string }> {
+    const suffix = randomUUID().replace(/-/g, '');
+    const productId = `ol_product_${suffix}`;
+    const variantId = `ol_variant_${suffix}`;
+    const externalId = `ext-${suffix.slice(0, 8)}`;
+
+    const mappingRepo = dataSource.getRepository(IdentifierMappingOrmEntity);
+    await mappingRepo.save(
+      mappingRepo.create({
+        entityType: 'Product',
+        internalId: productId,
+        externalId,
+        platformType: 'prestashop',
+        connectionId,
+        context: null,
+      })
+    );
+
+    const productRepo = dataSource.getRepository(ProductOrmEntity);
+    await productRepo.save(productRepo.create({ id: productId, name: 'Override Test' }));
+
+    const variantRepo = dataSource.getRepository(ProductVariantOrmEntity);
+    await variantRepo.save(
+      variantRepo.create({ id: variantId, productId, sku: `SKU-${suffix.slice(0, 6)}` })
+    );
+
+    return { productId, variantId, externalId };
+  }
+
+  /** A fake `InventoryMasterPort` that never locates — matches both shipped adapters. */
+  function pooledOnlyAdapter(input: {
+    externalId: string;
+    variantId: string;
+    quantity: number;
+  }): InventoryMasterPort {
+    return {
+      listInventory: jest.fn().mockResolvedValue([
+        {
+          id: `inv-${input.externalId}`,
+          productId: input.externalId,
+          variantId: input.variantId,
+          quantity: input.quantity,
+          reserved: 0,
+          available: input.quantity,
+          locationId: undefined,
+        },
+      ]),
+      getInventory: jest.fn(),
+      adjustInventory: jest.fn(),
+      reserveInventory: jest.fn(),
+      releaseInventory: jest.fn(),
+      getAvailableQuantity: jest.fn(),
+    } as unknown as InventoryMasterPort;
+  }
+
+  async function readRows(
+    productId: string,
+    variantId: string
+  ): Promise<InventoryItemOrmEntity[]> {
+    return dataSource
+      .getRepository(InventoryItemOrmEntity)
+      .find({ where: { productId, productVariantId: variantId } });
+  }
+
+  it('stales the pooled row and writes one located row once the override is set, with no double-count', async () => {
+    const master = await createTestConnection(dataSource, {
+      platformType: 'prestashop',
+      status: 'active',
+      config: { baseUrl: 'http://localhost:8080', preferredLanguageId: 1 },
+    });
+    const { productId, variantId, externalId } = await seedVariant(master.id);
+
+    jest
+      .spyOn(integrationsService, 'getCapabilityAdapter')
+      .mockResolvedValue(pooledOnlyAdapter({ externalId, variantId, quantity: 8 }));
+
+    const masterSync = harness.get<IMasterInventorySyncService>(MASTER_INVENTORY_SYNC_SERVICE_TOKEN);
+
+    // First sync: no override configured yet — the pre-#3206 behaviour, one
+    // pooled row, nothing to repair.
+    const first = await masterSync.syncFromMasterByExternalId(master.id, externalId);
+    expect(first.pooledPositionsStaled ?? 0).toBe(0);
+
+    const afterFirst = await readRows(productId, variantId);
+    expect(afterFirst).toHaveLength(1);
+    expect(afterFirst[0].locationId).toBeNull();
+    expect(afterFirst[0].isStale).toBe(false);
+    expect(afterFirst[0].availableQuantity).toBe(8);
+
+    // The operator now asserts a location for this connection's stock. The
+    // adapter is UNCHANGED — it still reports no location of its own — so any
+    // effect below is attributable to the override alone.
+    await dataSource
+      .getRepository(ConnectionOrmEntity)
+      .update(
+        { id: master.id },
+        { config: { baseUrl: 'http://localhost:8080', preferredLanguageId: 1, stockLocationOverride: OVERRIDE_LOCATION_ID } }
+      );
+
+    // Second sync: same adapter response, override now active.
+    const second = await masterSync.syncFromMasterByExternalId(master.id, externalId);
+    expect(second.pooledPositionsStaled ?? 0).toBeGreaterThan(0);
+
+    const afterSecond = await readRows(productId, variantId);
+    expect(afterSecond).toHaveLength(2);
+
+    const pooled = afterSecond.find((r) => r.locationId === null);
+    const located = afterSecond.find((r) => r.locationId === OVERRIDE_LOCATION_ID);
+
+    // The old row survives (SOFT repair) but is stale — not stock any more.
+    expect(pooled?.isStale).toBe(true);
+    // The new row sits at the override's location and is live.
+    expect(located?.isStale).toBe(false);
+    expect(located?.availableQuantity).toBe(8);
+
+    // No double count: only ONE of the two rows is live at once, and its
+    // quantity is the same 8 the adapter has reported all along.
+    const liveTotal = afterSecond
+      .filter((r) => !r.isStale)
+      .reduce((sum, r) => sum + r.availableQuantity, 0);
+    expect(liveTotal).toBe(8);
+  });
+
+  it('never applies the override once the adapter reports a real location itself', async () => {
+    const master = await createTestConnection(dataSource, {
+      platformType: 'prestashop',
+      status: 'active',
+      config: {
+        baseUrl: 'http://localhost:8080',
+        preferredLanguageId: 1,
+        stockLocationOverride: OVERRIDE_LOCATION_ID,
+      },
+    });
+    const { productId, variantId, externalId } = await seedVariant(master.id);
+
+    const REAL_LOCATION_ID = 'warehouse-real';
+    jest.spyOn(integrationsService, 'getCapabilityAdapter').mockResolvedValue({
+      listInventory: jest.fn().mockResolvedValue([
+        {
+          id: `inv-${externalId}`,
+          productId: externalId,
+          variantId,
+          quantity: 5,
+          reserved: 0,
+          available: 5,
+          locationId: REAL_LOCATION_ID,
+        },
+      ]),
+      getInventory: jest.fn(),
+      adjustInventory: jest.fn(),
+      reserveInventory: jest.fn(),
+      releaseInventory: jest.fn(),
+      getAvailableQuantity: jest.fn(),
+    } as unknown as InventoryMasterPort);
+
+    const masterSync = harness.get<IMasterInventorySyncService>(MASTER_INVENTORY_SYNC_SERVICE_TOKEN);
+
+    await masterSync.syncFromMasterByExternalId(master.id, externalId);
+
+    const rows = await readRows(productId, variantId);
+    expect(rows).toHaveLength(1);
+    // The adapter's own answer wins — the override configured on this
+    // connection is never consulted once the master actually locates.
+    expect(rows[0].locationId).toBe(REAL_LOCATION_ID);
+    expect(rows[0].isStale).toBe(false);
+  });
+});
--- a/libs/core/src/identifier-mapping/domain/types/connection.types.ts
+++ b/libs/core/src/identifier-mapping/domain/types/connection.types.ts
@@ -129,6 +129,30 @@ export interface ConnectionConfig {
   salesDocument?: {
     documentKind?: string;
   };
+  /**
+   * Operator-declared location for this connection's currently-unlocated
+   * stock (#3206). Neither shipped `InventoryMasterPort` adapter reports a
+   * `locationId` (PrestaShop only via the rarely-enabled Advanced Stock
+   * Management module; WooCommerce has no native multi-location at all), so
+   * ADR-058 decision (2) means every position from such a master is
+   * permanently pooled (`locationId IS NULL`) — which makes the OMS
+   * fulfilment router structurally sterile even once locations exist,
+   * because it skips every pooled row.
+   *
+   * This is NOT the sync inventing a location the master declined to give —
+   * that remains forbidden. It is the OPERATOR supplying, once and
+   * explicitly, the fact the master will never supply: "all of this
+   * connection's currently-unlocated stock physically lives here." A real
+   * adapter-reported location always wins over this value; see
+   * `readStockLocationOverride` (`@openlinker/core/inventory`'s
+   * `stock-location-override.types.ts`) for where it is consumed and why
+   * that ordering is enforced in exactly one place.
+   *
+   * An operator with genuinely multiple physical warehouses behind one
+   * connection must NOT use this — it asserts a single location for ALL of
+   * that connection's stock, which would misattribute the rest.
+   */
+  stockLocationOverride?: string;
   [key: string]: unknown;
 }
 
--- a/libs/core/src/inventory/application/services/__tests__/master-inventory-sync.service.spec.ts
+++ b/libs/core/src/inventory/application/services/__tests__/master-inventory-sync.service.spec.ts
@@ -65,7 +65,12 @@ describe('MasterInventorySyncService', () => {
 
     integrationsService = {
       getCapabilityAdapter: jest.fn().mockResolvedValue(inventoryAdapter),
-      getAdapter: jest.fn(),
+      // Default: no `stockLocationOverride` configured — byte-identical to the
+      // pre-#3206 behaviour for every test that doesn't opt in explicitly.
+      getAdapter: jest.fn().mockResolvedValue({
+        connection: { config: {} },
+        metadata: {},
+      }),
       listCapabilityAdapters: jest.fn(),
     } as unknown as jest.Mocked<IIntegrationsService>;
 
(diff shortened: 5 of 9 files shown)


norbert-kulus-blockydevs:
## Review reconciliation — findings addressed

Following up on the independent review (approve with changes):

| Finding | Status |
|---|---|
| **[IMPORTANT]** Missing #3206 AC-required Postgres int-spec for the pooled→located repair transition | **Fixed** — `apps/worker/test/integration/inventory-stock-location-override-e2e.int-spec.ts`, 2 cases against real Postgres: (1) the old pooled row is staled and exactly one located row appears with no double-counted `availableQuantity` once the override is set; (2) a real adapter-reported location still always wins over the override. Modelled on the sibling #2324 end-to-end spec rather than the #2322-only `IInventoryService`-level one, since the override is applied inside `MasterInventorySyncService` itself. |
| **[SUGGESTION]** `readStockLocationOverride` didn't `.trim()` on the valid path | **Fixed** — now returns `raw.trim()`, matching the docblock's own "coerces defensively" claim. |
| **[SUGGESTION]** `check-architecture-gates.mjs` `KNOB_THRESHOLD` bump | **Left as-is** — the reviewer flagged this as needing a maintainer's explicit sign-off on the rationale, not a code fix; not touched. |
| **[SUGGESTION]** `LocationService.getLocation` not filtering by `status` | **Left as-is** — reviewer called this a product-level judgment call (an operator could set the override to a retired location), not a bug. |
| **[SUGGESTION]** redundant connection fetch in `resolveStockLocationOverride` | Not addressed in this pass (marked optional, would need reshaping the `getCapabilityAdapter`/`getAdapter` contract — out of scope here). |

Ran the new int-spec directly against real Testcontainers Postgres — both cases pass. Full pre-commit gate (`smart-test`'s scoped selection, 112 suites / 2451 tests) also green on this commit.

piotrswierzy (commented):
## Tech Lead review

Well-shaped slice: one pure coercion helper on the `readStockSafetyBuffer`/`readPricingRule` precedent, applied at exactly one seam, resolved once per sync call, with `inventory.locationId ?? override` ordering so a real adapter answer always wins. The ADR-058 decision (2) argument holds — the rule forbids *the sync* inventing a location, and this reads a fact *the operator* supplied — and I verified the claimed non-interactions: `getAdapter` really is metadata-only (constructs no adapter, resolves no credentials), no plugin config-shape validator uses `forbidNonWhitelisted`, and the effective-location value correctly feeds the #2322 located/pooled tracking. CI is fully green and the int-spec added after the first review round genuinely pins the pooled→located transition against real Postgres. My concerns are all about what happens *around* the happy path: the architecture ledger is not updated, and three reachable states are unguarded — most notably that unsetting the override double-counts available-to-promise.

### Findings

**[IMPORTANT]** — `docs/architecture/adrs/058-multi-location-positions-reservations-availability-authority.md:26`, `docs/architecture-overview.md` § Inventory
> No documentation changes ship with this PR, and two doc statements become literally false. ADR-058 decision (2) reads "`locationId IS NULL` permanently means 'the master declines to locate its stock' — never a default location", and the #2407 bullet in `architecture-overview.md` § Inventory reads "no minted row is ever a stand-in for that NULL". After this PR a minted row *is* a stand-in for that NULL whenever the operator says so. The distinction you draw (operator-supplied vs. sync-invented) is the right one and is well argued in the PR body and code comments — but the PR body is not where the next reader looks, and CLAUDE.md is explicit: "Do not redefine or reinterpret architecture — follow what the docs say" and "Justify any new pattern, abstraction, or component explicitly." This repo amends ADRs as routine practice (ADR-009 § Amendment (#2024), ADR-040 § Amendment (#2777), ADR-050 § Amendment (#2594), ADR-067 § Amendment (#2621)), and every comparable inventory slice — #2313, #2317, #2320, #2322, #2324, #2407 — records itself as a § Inventory bullet. Add an `## Amendment (#3206)` section to ADR-058 restating decision (2) as "the master declines, *or the operator has asserted on the master's behalf*", and a § Inventory bullet; also correct the #2407 sentence, which now overstates.

**[IMPORTANT]** — `libs/core/src/inventory/application/services/master-inventory-sync.service.ts:236`
> **Unsetting the override double-counts available-to-promise.** `markStaleExceptVariants` (`inventory.repository.ts:425`) carries no `locationId` predicate — it keeps every location row of a still-present variant. So on the reverse transition: `effectiveLocationId` returns to `null`, `setInventory` re-creates and un-stales the pooled row (`isStale` is master-owned on upsert), and the located row at the old override is never touched again and stays live. Both rows are then live for the same `(product, variant, source)`, and `getPromisableQuantities` in `global` scope sums "across every live position — all locations *and* all sources" (architecture-overview § Inventory, #2324/#2321). That is an oversell with every counter internally consistent. This is exactly the gap #2322 documented as "Reversal is half free, and only half … the total double-counts on the way back" — but #2322 could leave it, because no in-tree adapter ever locates, so the reverse transition was unreachable. This PR makes it a one-field config edit, which is the most likely way an operator will react to getting the override wrong. Nothing here tests, documents or guards it. Minimum: state it in the config-key docblock and the § Inventory bullet. Better: on the transition to no-override, stale the same source's rows at the previously-overridden location (the mirror of `markLocationlessStaleForSource`), or have `validateStockLocationOverride` refuse to clear a value that has already been synced.

**[IMPORTANT]** — `apps/api/src/integrations/application/services/connection.service.ts:320`
> `validateStockLocationOverride` accepts any location `getLocation` returns, and `LocationService.getLocation` (`location.service.ts:74`) is a bare `findById` with no `status` filter — so an `inactive` (retired) location passes. That value provably cannot deliver the feature's own purpose: `OlFulfillmentRouter.loadCandidates` (`libs/oms/src/routing/ol-fulfillment-router.ts:114`) lists locations with `{ status: 'active' }`, so stock written at an inactive location enters the stock map and matches no candidate — the line still resolves `unfulfillable`, silently, which is the exact failure #3206 exists to remove. It is also internally inconsistent with `assertRouterEnablementPreconditions` in this same file, which gates routing on `countActiveLocations() > 0` (#2407) — one method refuses zero *active* locations while the other accepts an *inactive* one. The previous reviewer called this a product judgment call; the sibling gate two methods away makes it a consistency defect. One line: reject when `location.status !== 'active'`, naming the status.

**[SUGGESTION]** — `apps/api/src/integrations/application/services/connection.service.ts:320`
> Time-of-check/time-of-use on deletion. `LocationService.deleteLocation` refuses (409 `LocationInUseError`) only while `countPositionsAtLocation > 0`, and `inventory_items` carries no FK to `inventory_locations` (ADR-058 decision 3 defers that to step iii). Between setting the override and the first sync writing rows there are zero positions, so the location can be deleted out from under a validated override; the next sync then writes rows naming a location that does not exist, with no error anywhere. `readStockLocationOverride`'s docblock is honest that it "cannot re-validate that on every sync", which is right — but the counterpart guard would be for `deleteLocation` to also refuse while any connection config names the id. Worth at least recording as a known gap.

**[SUGGESTION]** — `libs/core/src/inventory/application/services/master-inventory-sync.service.ts:627`
> This PR makes a non-NULL `inventory_items.locationId` reachable in-tree for the first time, which arms a previously-dormant error path: `InventoryCrossSourcePositionConflictError` is documented as raised when a second source INSERTs at a **non-NULL** location, because the partial unique indexes have no NULL to be distinct about, and it is explicitly permanent-and-non-retryable until #2325's four-column index lands. I checked and this is *not* currently reachable — `getOrCreateInternalId` namespaces per `(entityType, externalId, connectionId)` so two connections never share a `productId`, which is the same reason #1904 calls its own collision "unreachable through any supported workflow". So no change is needed. It is worth a sentence in the docblock anyway, because the override is the first thing in the tree that would expose it the day that namespacing assumption stops holding.

**Positive observations.** The once-per-batch resolution is asserted by a real test (`getAdapter` called exactly once across three external ids) rather than asserted in a comment. The `effectiveLocationId` comment correctly names the load-bearing property that both the lookup and the constructed row must use the *same* value. Replacing `(inventory.locationId ?? null) !== null` with `inventoryItem.locationId !== null` for the #2322 branch is the right call and is the subtle bit most implementations would miss. And the `KNOB_THRESHOLD` 7→8 rationale is genuinely argued (routing/authority knobs vs. a physical-world assertion) rather than just bumped — that reads as a maintainer sign-off question, not a code issue, and I have no objection to it.

### Verdict
🔄 **Approve with changes**

(Posting as a comment rather than an approval: there are IMPORTANT findings outstanding.)

Priority:
1. The reverse transition (override → no override) leaves a double-counted live position. Guard it, or state it explicitly in the docblock and the architecture ledger.
2. Reject an `inactive` location in `validateStockLocationOverride` — it cannot route, and the sibling `countActiveLocations()` gate in the same file already says so.
3. Amend ADR-058 and add the `docs/architecture-overview.md` § Inventory bullet; correct the now-overstated #2407 sentence.

---

piotrswierzy (commented):
## Tech Lead re-review (head unchanged at `3841aa8f`)

Re-verified every finding in code at the unchanged head. **All three IMPORTANTs and both SUGGESTIONs still stand**, and each supporting fact I relied on is untouched by the 11 PRs merged to `main` since. What *has* changed is the baseline: this PR is now `CONFLICTING`, because **#3158 (price-change episode schema, merged 2026-09-13) touched all four of the same non-test files** — `connection.service.ts`, `connection.service.spec.ts`, `connection.types.ts` and `scripts/check-architecture-gates.mjs`. Three of those four are benign additive collisions. The fourth is not: #3158 registered a knob **and** raised `KNOB_THRESHOLD` 7 → 8, this PR independently raises it 7 → 8, and a textual merge — from either side, since both write `8` — lands 8 knobs against a threshold of 8, which trips the gate and fails `pnpm lint`. That is a new finding, and it also re-opens a review question #3158 explicitly pre-registered. CI on this head is fully green (9/9), and there is no migration in this PR, so the timestamp contention on `main` is moot here.

### Findings

**[IMPORTANT — NEW, caused by the new `main`]** — `scripts/check-architecture-gates.mjs:253`
> **The knob arithmetic no longer works out, and the merge resolution cannot be textual.** On `main` today `KNOWN_CONFIG_KNOBS` holds **7** helpers (`parseTriggerModel`, `readStockSafetyBuffer`, `readPricingRule`, `readPriceSyncModeConfig`, `readSalesDocumentRouting`, `readLifecycleAuthority`, `parseAuthorityConfig`) with `KNOB_THRESHOLD = 8` — #3158 added the 7th (`readPriceSyncModeConfig` / `config.priceSyncMode`) and raised the bar in the same commit. This PR adds an 8th (`readStockLocationOverride`) and *also* raises 7 → 8. `evaluateKnobs` fires on `knownCount >= threshold`, so post-merge that is `8 >= 8` → the gate trips and `check:invariants` fails. Both sides of the conflicting line read `8`, so "take ours" and "take theirs" are equally wrong — the correct value is **9**, which no conflict marker will suggest.
>
> The docblock needs more than a number, though. Yours reads *"**Contributing raise — #3206, 7 -> 8**"* and closes with *"the rung is raised by exactly one, so the NEXT unrelated knob still stops and has this conversation."* After the merge this is the next unrelated knob — so the paragraph has to be rewritten as **the eighth knob, 8 → 9**. And #3158's own paragraph pre-registered precisely this exchange: *"the eighth knob re-opens this same conversation with no further headroom"*, plus *"A reviewer who disagrees should push back on this raise specifically."* So the raise is no longer a routine +1 inheriting the #2305/#2304 deferral; it is the case the previous author said would need re-arguing. Your substantive argument — that every prior entry coerces a routing/authority decision while this one is a one-shot operator assertion of a physical fact, so #2169's consolidation target has no shape to absorb it — is a good argument and I accept it. It just needs to be made against a threshold of 8 rather than 7, and to say so.
>
> The other three collisions are additive and semantically independent (`connection.types.ts`: your `stockLocationOverride?` lands before `[key: string]: unknown` while #3158 added `priceSyncMode?` and widened `pricingRule` higher up; `connection.service.ts`: a third `validateX(config)` call beside the existing two in both `create` and `update`). Keep both sides. Please re-run `pnpm check:invariants` after resolving rather than trusting the merge — that is the only thing that will catch the `8`.

**[IMPORTANT — STILL HOLDS]** — `libs/core/src/inventory/application/services/master-inventory-sync.service.ts:236`
> **Unsetting the override still double-counts available-to-promise.** Re-verified in code: `markStaleExceptVariants` (`inventory.repository.ts:425–460`) filters on `productId`, `isStale = false`, the variant-keep bracket and the optional provenance scope — **there is no `locationId` predicate anywhere in it**, so it keeps every location row of a still-present variant, exactly as before. So on the reverse transition `effectiveLocationId` returns to `null`, `setInventory` re-creates and un-stales the pooled row (`isStale` is master-owned on upsert), the located row at the old override is never touched, and both are live for the same `(product, variant, source)` — which `getPromisableQuantities` sums in `global` scope across all locations and all sources (§ Inventory, #2324/#2321). An oversell with every counter internally consistent. #2322 documented this as *"reversal is half free, and only half"* and could leave it because no in-tree adapter ever locates; this PR makes it a one-field config edit, which is the most likely way an operator reacts to getting the override wrong. Note the mirror machinery is 60 lines further down the same file — `markLocationlessStaleForSource` does exactly this in the forward direction, and its own comment at line 521 cites *"the discipline `markStaleExceptVariants` keeps, in the mirror direction."* Minimum: state it in the config-key docblock and the § Inventory bullet. Better: stale the same source's rows at the previously-overridden location on the transition to no-override, or have `validateStockLocationOverride` refuse to clear a value that has already been synced.

**[IMPORTANT — STILL HOLDS, and now more reachable]** — `apps/api/src/integrations/application/services/connection.service.ts:320`
> **An `inactive` location still passes validation.** All four supporting facts re-verified and unchanged by the new `main`: `validateStockLocationOverride` stops at `if (!location)` with no status test; `LocationService.getLocation` (`location.service.ts:74`) is still a bare `return this.repository.findById(id)`; `OlFulfillmentRouter.loadCandidates` still lists `{ status: 'active' }` (`ol-fulfillment-router.ts:115`); and `assertRouterEnablementPreconditions` still gates on `countActiveLocations() === 0` at `connection.service.ts:150` — ~170 lines above the new method, in the same class. So one method refuses zero *active* locations while the other accepts an *inactive* one, and stock written at a retired location enters the stock map, matches no candidate, and resolves `unfulfillable` silently — the exact failure #3206 exists to remove. One line: reject when `location.status !== 'active'`, naming the status.
>
> Worth knowing that this is sharper than it was yesterday: **PR #3198 (locations CRUD UI) is in flight and ships a Retire action** that sets `status: 'inactive'` from a button. Today reaching an inactive location takes a deliberate API call; once #3198 lands it is one click on the page an operator visits to set this override up.

**[IMPORTANT — STILL HOLDS, and it is two sentences, not one]** — `docs/architecture/adrs/058-…md:26`, `docs/architecture-overview.md:165` and `:172`
> Still zero `docs/` files in this PR's 9 changed files, and both statements are still on `main` verbatim. I under-counted last time: there are **two** sentences in `architecture-overview.md` that become false, not one. Line 165 (the #2313 bullet) — *"`locationId IS NULL` permanently means 'the master declines to locate its stock' (decision 2) — never a default location, so **no row here is ever a stand-in for that NULL**"*; and line 172 (the #2407 bullet) — *"**no minted row is ever a stand-in for that NULL**"*. Plus ADR-058:26 itself. The operator-supplied-vs-sync-invented distinction you draw is the right one and is well argued in the code comments, but the PR body is not where the next reader looks, and this repo amends ADRs as routine practice (ADR-009 § Amendment (#2024), ADR-040 § Amendment (#2777), ADR-050 § Amendment (#2594), ADR-067 § Amendment (#2621)) while every comparable inventory slice — #2313, #2317, #2320, #2322, #2324, #2407 — records itself as a § Inventory bullet. Add an `## Amendment (#3206)` to ADR-058 restating decision (2) as *"the master declines, **or the operator has asserted on the master's behalf**"*, a § Inventory bullet, and a correction to both now-overstated sentences.

**[SUGGESTION — STILL HOLDS, and also more reachable]** — `apps/api/src/integrations/application/services/connection.service.ts:320`
> Time-of-check/time-of-use on deletion, unchanged: `LocationService.deleteLocation` (`location.service.ts:89–102`) still refuses only while `countPositionsAtLocation > 0`, and `inventory_items` still carries no FK to `inventory_locations` (ADR-058 decision 3 defers that to step iii). Between setting the override and the first sync writing rows there are zero positions, so the location can be deleted out from under a validated override and the next sync writes rows naming a location that does not exist, with no error anywhere. `readStockLocationOverride`'s docblock is honest that it *"cannot re-validate that on every sync"* — the counterpart guard would be for `deleteLocation` to also refuse while any connection config names the id. Same #3198 note as above: that PR puts Delete on a button too. Worth at least recording as a known gap.

**[SUGGESTION — STILL HOLDS]** — `libs/core/src/inventory/application/services/master-inventory-sync.service.ts:627`
> Unchanged, and I still do **not** think this is a defect. This PR makes a non-NULL `inventory_items.locationId` reachable in-tree for the first time, which arms the previously-dormant `InventoryCrossSourcePositionConflictError` path — but it stays unreachable, because `getOrCreateInternalId` namespaces per `(entityType, externalId, connectionId)` so two connections never share a `productId`, the same reason #1904 calls its own collision "unreachable through any supported workflow". No change needed. Still worth a sentence in the docblock, because the override is the first thing in the tree that would expose it the day that namespacing assumption stops holding.

**Positive observations (re-confirmed).** The once-per-batch adapter resolution is still pinned by a real test rather than a comment; the `effectiveLocationId` comment still names the load-bearing property that the lookup and the constructed row must use the same value; and `inventoryItem.locationId !== null` for the #2322 branch remains the subtle bit most implementations would miss. The int-spec added in the earlier round still pins the pooled → located transition against real Postgres. None of that is affected by the merge.

### Verdict
🔄 **Approve with changes** — unchanged from my last pass, with the merge now ahead of it in priority.

Priority:
0. **Merge `main` and resolve `KNOB_THRESHOLD` to `9`, not `8`** — both sides of the conflict say `8` and both are wrong; rewrite the rationale as the eighth knob (8 → 9) and answer the push-back #3158's docblock pre-registered. Re-run `pnpm check:invariants` after resolving.
1. The reverse transition (override → no override) leaves a double-counted live position. Guard it, or state it explicitly in the docblock and the architecture ledger.
2. Reject an `inactive` location in `validateStockLocationOverride` — it cannot route, and the sibling `countActiveLocations()` gate in the same file already says so.
3. Amend ADR-058 and add the § Inventory bullet; correct **both** now-overstated sentences (`architecture-overview.md:165` and `:172`).

---

piotrswierzy (commented):
## Tech Lead re-verification (head unchanged at `3841aa8`)

Head is unchanged, `main` is unchanged (`e336cf4`), and every one of the nine CI checks passes on this head — so I am not re-running the full review, only the part that moved. The part that moved is the merge state: this PR is still **`CONFLICTING` / `DIRTY`** against `main` via merged #3158, and I have now confirmed the knob-threshold trap empirically rather than by inference. All three standing findings re-verified against the current head and still hold. Verdict stands.

### The knob-threshold trap is live, and confirmed by counting both sides

| | knobs registered | `KNOB_THRESHOLD` | gate (`knownCount >= threshold`) |
|---|---|---|---|
| `main` @ `e336cf4` | 7 (6 shared + `readPriceSyncModeConfig`, #3158) | `8` | `7 >= 8` → passes |
| this PR @ `3841aa8` | 7 (6 shared + `readStockLocationOverride`) | `8` | `7 >= 8` → passes |
| **naive textual merge** | **8** (6 shared + **both**) | **`8`** (both sides wrote the same literal) | **`8 >= 8` → FIRES** |

Both branches raised the rung `7 -> 8` from the same base, so the two raises are textually identical and **git will not mark `KNOB_THRESHOLD` as a conflict at all** — only the `KNOWN_CONFIG_KNOBS` map entries collide, and resolving those additively (which is correct) silently produces an 8-knob tree under an 8 threshold. `evaluateKnobs` then fails `pnpm lint` with the "at 8, the shared per-connection rules model" message.

**The correct merged value is `9`, and no conflict marker will suggest it.** Please state it explicitly in the merge commit or a follow-up push.

**And the raise now owes a fresh argument.** #3158's own docblock, which is what you are merging on top of, pre-registers this exact moment:

> 3. The consolidation already has an owner (#2169). Raising the bar by one buys exactly one more knob before the gate fires again — **the eighth knob re-opens this same conversation with no further headroom.**
>
> **A reviewer who disagrees should push back on this raise specifically.**

Your existing justification (`stockLocationOverride` coerces neither a routing nor an authority decision — it is a one-shot operator assertion of a physical fact no master will report) is, I think, the strongest case any of the eight entries has for not being the accretion #2169 is aimed at, and I am not asking you to drop the knob. But the merged docblock must be rewritten to argue the 8→9 raise on the merged tree, not restate a 7→8 raise against a base that no longer exists. Please do not leave the two "Contributing raise — … 7 -> 8" paragraphs sitting adjacent, each claiming the rung it took.

### Findings

All three re-verified against the current head; none has been addressed, and one has become materially more reachable (see the cross-PR note).

**[IMPORTANT]** — `libs/core/src/inventory/infrastructure/persistence/repositories/inventory.repository.ts:425-460` — double-counted ATP on the reverse transition.

Re-read on `main`: `markStaleExceptVariants` filters on `productId`, `isStale = false`, the variant-keep bracket and the optional provenance scope — and **nothing else**. There is no `locationId` predicate anywhere in the builder. So when an operator unsets the override, the sync re-creates and un-stales the pooled (`locationId IS NULL`) row while the previously-written located row stays live, and `getPromisableQuantities` in `global` scope sums across all locations (ADR-058 decision 2 / #2321). The variant's ATP doubles. Every counter is internally consistent, nothing throws, and no log fires — an oversell that is invisible until a buyer hits it.

Note #2322 already shipped the *forward* direction of exactly this repair (`markLocationlessStaleForSource`, the pooled-row sweep when a source starts locating). The reverse direction is the one ADR-058 explicitly leaves open — "#2322 enforces decision (2) in one direction only; the mirror-image sweep is multi-location pruning, which ADR-058 leaves out of scope" — but that gap was previously unreachable in-product, because no shipped adapter ever sets `locationId`. This PR is what makes an operator able to reach it, in one `PATCH`, and back again.

**[IMPORTANT]** — `apps/api/src/integrations/application/services/connection.service.ts:312-324` — `validateStockLocationOverride` accepts an inactive location.

```ts
const location = await this.locations.getLocation(value);
if (!location) { throw new BadRequestException(...); }
```

A bare existence check, no `status` test. The fulfilment router filters `status: 'active'`, and `assertRouterEnablementPreconditions` — **line 150 of this same file, ~170 lines above** — already uses `countActiveLocations()` precisely because an inactive location decides nothing. Accepting a retired location here writes a config that names a row the rest of the system treats as absent, which is the #2407 shape (a configuration that persists happily and then decides nothing) rather than a refusal an operator can act on.

**[IMPORTANT]** — `docs/architecture-overview.md:165` and `:172` — two shipped sentences become false, and no ADR-058 amendment ships.

Verified both still read as written on `main`, and this PR touches neither file:

- `:165` — "**`locationId IS NULL` permanently means 'the master declines to locate its stock'** (decision 2) — never a default location, so no row here is ever a stand-in for that NULL."
- `:172` — "`locationId IS NULL` on a position permanently means the master declines to locate its stock (ADR-058 decision 2) and **no minted row is ever a stand-in for that**."

The feature's whole purpose is to make an operator-named location stand in for that NULL. "Permanently" and "never … a stand-in" are the load-bearing words, and they are now wrong. Your own gate docblock cites decision (2) approvingly while the code narrows it — so the ADR needs the amendment, not just the prose. This is the same obligation ADR-058's own ladder entries meet elsewhere.

### Cross-PR interaction (new since last round)

**#3198 ships Retire and Delete buttons for locations**, which changes the reachability of two items above from "deliberate API call" to "one click":

1. **The inactive-location finding.** #3198's `handleRetire` sets `status: 'inactive'` with no confirm dialog. After it lands, an operator can retire a location in one click and this PR will still accept it as a `stockLocationOverride` — silently producing a connection whose override names a location the router filters out.
2. **The TOCTOU-deletion suggestion.** I checked `LocationService.deleteLocation` (`location.service.ts:89-107`): it refuses only on `countPositionsAtLocation(id) > 0`, i.e. `inventory_items` rows. A `stockLocationOverride` living in `connection.config` is **invisible to that guard**, so a location referenced only by an override deletes cleanly and leaves a dangling override that `validateStockLocationOverride` would have refused at write time and nothing ever re-checks.

Whichever of the two lands second should account for this — either by widening `deleteLocation`'s referential refusal to config overrides, or by making the override resolution degrade explicitly (and observably) when the named location is missing or inactive, rather than silently.

### Verdict

🔄 **Approve with changes**

Priority, in order:
1. Resolve the merge to **`KNOB_THRESHOLD = 9`** and rewrite the raise docblock to argue 8→9 on the merged tree.
2. Close the reverse-transition ATP gap, or state the limitation where it is caused and gate the override behind it.
3. Reject an inactive location in `validateStockLocationOverride`.
4. Amend ADR-058 and fix `architecture-overview.md:165` / `:172`.

---

norbert-kulus-blockydevs:
## Review addressed — commits `7aca70d75`, `64359c4a8`, `f8bd68837`

- **[IMPORTANT] Double-counted ATP on the reverse transition** — *fixed*. The gap is confirmed as you described: `markStaleExceptVariants` carries no `locationId` predicate and `findByProductAndVariant` does not filter `isStale`, so clearing the override un-stales the pooled row while the located row stays live and `global`-scope ATP doubles.

  Added the mirror repair `markLocatedStaleForSource` / `staleLocatedPositionsForSource` — the third `isStale` writer — with the same provenance rule, the same no-`updatedAt`-bump raw statement, **no** `master.variant.stale` emission, its own `inventory.propagateToMarketplaces` enqueue, reported as `MasterInventorySyncResult.locatedPositionsStaled`. Both repairs now share one private statement builder so the provenance-membership rule is not copied a third time. A variant reported both pooled AND located in one payload is subtracted from the mirror's key set, so "located wins" survives and it can never be staled from both sides into a live-position-less known zero (#1689). Pinned by 5 unit tests plus a reverse-transition int-spec asserting **8, never 16**.

  Strictly the decision-(2) mirror, **not** multi-location pruning — that gap is unchanged and said so in place. The now-false "reversal is free and needs no code" claim in the `markLocationlessStaleForSource` port docblock is corrected, as is the matching narrative in `inventory-pooled-position-enforcement.int-spec.ts` (its assertions are untouched and still valid — that spec drives the primitives without the mirror).
- **[IMPORTANT] `validateStockLocationOverride` accepts an inactive location** — *fixed*. Now refuses `status !== 'active'` alongside the unknown-id refusal, with copy naming the remedy, and the validated id is trimmed to match what `readStockLocationOverride` hands the sync. Scoped to the **transition** against the persisted config, mirroring `assertRouterEnablementPreconditions`: `config` is replaced wholesale, so an unconditional check would refuse every unrelated patch on a connection whose location was retired later — including `AuthorityStatusService`'s who-decides preset apply, which writes the whole config back and reports only `failedConnectionIds`.
- **[IMPORTANT] Two shipped sentences false, no ADR-058 amendment** — *fixed, and it was three sentences, not two.* Both you named are rewritten so the rule binds the **master's** silence rather than the operator's assertion, with no synthetic default location minted anywhere and a persisted `locationId` never standing in for a NULL. Point (4) of the #2322 bullet ("reversal is half free") is likewise falsified by this change and is corrected too. A new § Inventory bullet covers #3206, and ADR-058 gains `## Amendment (#3206, 2026-09-14)` **as its own section** with the original decision text left intact, plus #2322/#3206 on References.
- **[priority 1] `KNOB_THRESHOLD` and the raise docblock** — *fixed during the `main` merge, and the outcome is 9 rather than the 8→9 you expected from this side alone.* `main` took the 7→8 step for a different helper while this branch was open (#3142, `readPriceSyncModeConfig`), and this branch had taken 7→8 for `readStockLocationOverride`. Both raises are real and independent, so **both justification paragraphs stand** and the threshold lands on **9**; this branch's paragraph is retitled "the eighth knob, 8 → 9" and carries a merge note recording why, so neither side silently overwrites the other back to 8. `check-architecture-gates: gates marked in ADR-048+, 8/9 config knobs, 1/3 ladder rungs` ✅ plus `--self-check: 13 case(s) passed` ✅.
- **[cross-PR] `deleteLocation` blind to a config override** — `needs-owner-decision`, not closed here. The mirror repair makes the vanished-location case **arithmetically safe** (the located row is staled, the pooled row wins), so it is no longer an oversell — but a dangling override still names a row the router filters out. Closing it properly is either widening `deleteLocation`'s referential refusal to config keys, or an observable degradation at resolve time in `MasterInventorySyncService` — the latter costs a new `ILocationService` dependency on the sync service, beyond this review's findings. Say which you want and I will open the issue.

One unrequested commit: widening `InventoryRepositoryPort` broke both `jest.Mocked<InventoryRepositoryPort>` object literals in `libs/core` — caught by `pnpm type-check` on this branch, fixed in `64359c4a8`. `main` is merged in (11 commits behind; the only conflict was the knob docblock above).

---

piotrswierzy (approved):
## Tech Lead re-review (`3841aa8` → `64359c4a`)

All four items from the last round are closed, and the merge trap is closed the way it needed to be rather than the way git would have done it by default. `main` was merged in (`MERGEABLE` again — `BLOCKED` is only the pending required checks), the reverse-transition ATP double-count now has a real mirror repair with its own propagation enqueue and a live-DB end-to-end spec, the inactive-location refusal ships with a transition-only guard I did not ask for and is better for it, and ADR-058 carries a proper amendment instead of two sentences quietly going false. One residual, and it is a one-line claim rather than a mechanism.

### The knob trap, closed and counted at head

| | knobs | `KNOB_THRESHOLD` | gate |
|---|---|---|---|
| `main` @ `e336cf4` | 7 | 8 | passes |
| this PR @ `64359c4a` | **8** | **9** | `8 >= 9` → passes |

Counted directly off `KNOWN_CONFIG_KNOBS` at head: `parseTriggerModel`, `readStockSafetyBuffer`, `readPricingRule`, `readPriceSyncModeConfig`, `readSalesDocumentRouting`, `readLifecycleAuthority`, `parseAuthorityConfig`, `readStockLocationOverride` — eight, under a threshold of nine. The 8-under-8 collision I described cannot now occur, and the docblock argues the raise on the merged tree rather than restating a 7→8 step against a base that no longer exists:

> **Contributing raise — #3206, the eighth knob, 8 -> 9.** … Merge note: #3142 took the 7 -> 8 step on `main` while this branch was open, for a different helper. Both raises are real and independent, so both paragraphs stand and the threshold lands on 9 rather than either side silently overwriting the other back to 8.

The `#3142` attribution is correct (that is the issue behind #3158's `readPriceSyncModeConfig`), and the rung is still raised by exactly one, so the ninth knob stops and has the conversation again. This is the resolution I asked for and the reasoning is legible to the next reader, which was the harder half.

### Prior findings — all closed

**1. Reverse-transition ATP double-count — CLOSED.** `markLocatedStaleForSource` is the exact mirror of `markLocationlessStaleForSource`: same provenance rule, same `!pruneSkipped` unattributed-claim invariant, same raw statement that does not bump `updatedAt`, same deliberate absence of a `master.variant.stale` emission. Three details make it right rather than merely present:

- **The contradicted set is subtracted** before the mirror runs. A variant reported both pooled and located in one payload is already staled from the pooled side ("located wins"), so staling it from this side too would leave it with *no* live position — a known zero that #1689 turns into paused offers for stock that is there. The comment says exactly that, in place.
- **It gets its own `enqueueAggregatePropagation`**, because `setInventory`'s no-change guard can legitimately skip the pooled row's enqueue while this pass removes the located row from the aggregate. Without it the destination keeps selling the doubled number until an unrelated write — the #2322 known-gap shape, avoided.
- **Both mirrors run off the *effective* location** (`inventoryItem.locationId` after the override is folded in by `toDomainInventoryItem`), so an overridden position is "located" for the repair exactly as a real one is.

The two repositories share one `markSameSourceOrphanPositionsStale` body with a closed `'pooled' | 'located'` literal selecting the location predicate — the right call, given the provenance-membership rule is already a documented duplicate of `applyProvenanceScope` and a third hand-written copy is how that warning stops being true. And the scope-limit is stated rather than implied: this is strictly the decision-(2) mirror, *not* multi-location pruning, so a master dropping one of two locations still leaves the abandoned row behind — unchanged, and named.

`inventory-stock-location-override-e2e.int-spec.ts` asserts both directions against real Postgres, including the `liveTotal === 8` no-double-count check, which is the assertion that would actually have failed before this commit.

**2. `validateStockLocationOverride` accepts an inactive location — CLOSED, and better than asked.** It now refuses `status !== 'active'` with copy naming the remedy, and — correctly — fires **only on the transition, compared against the persisted config**. That second half matters and I had not called for it: `ConnectionRepository.update` replaces `config` wholesale, so `AuthorityStatusService`'s preset apply re-asserts whatever override is already stored, and an unconditional check would make an unrelated patch fail for a state the guard itself let through. That is the `assertRouterEnablementPreconditions` / #2407 rule applied, and the trim-on-both-sides detail (validating the same string `readStockLocationOverride` will actually ask for) is the kind of thing that is only obvious after it has bitten someone.

**3. `architecture-overview.md:165` / `:172` and the missing ADR amendment — CLOSED.** ADR-058 gains `## Amendment (#3206, 2026-09-14): decision (2) is narrowed to the MASTER, and reversal is no longer half free`, and both prose sentences are rewritten to bind the master's silence rather than assert an absolute:

> **`locationId IS NULL` means "the MASTER declines to locate its stock"** (decision 2) — never a default location, and OpenLinker never invents one the master withheld. **Narrowed by #3206** …

The #2322 bullet's "Reversal is half free" line is corrected to "Reversal was half free, and #3206 paid for the other half", and the `inventory-repository.port.ts` docblock on `markLocationlessStaleForSource` — which said "Reversal is free and needs no code" — is corrected too. That last one is the one a future reader would actually have trusted, so correcting it at the source rather than only in the ADR is the right placement.

### Findings

**[IMPORTANT]** — `apps/api/src/integrations/application/services/connection.service.ts` (the `validateStockLocationOverride` docblock)

> closing that needs either a widened refusal there or an observable degradation at resolve time, and is **tracked** rather than half-done.

Nothing tracks it. `gh search issues --repo openlinker-project/openlinker "stockLocationOverride"` returns exactly #3206 and #3207, and neither covers the dangling-override case; the PR's own Follow-up section names only #3207 (the FE field). The deferral itself is correct and well-reasoned — I am not asking you to widen `deleteLocation` in this PR — but a docblock that says a gap is tracked, when it is not, is worse than one that says it is deferred, because it retires the next reader's attention on the strength of a lookup that will come back empty.

Re-verified at both heads: `LocationService.deleteLocation:89-102` still refuses only on `countPositionsAtLocation > 0`, and #3198 does not touch it either — so the gap is real and, once both PRs land, reachable in one click (Delete a location an override names; the config keeps pointing at a row that no longer exists, and nothing re-checks it). File the issue and replace "tracked" with its number, or say "deferred, untracked" and let the reviewer of the next slice decide.

### Verdict

✅ **Approve**

The one finding above is a one-line docblock correction plus an issue, not a code change, and it does not gate the merge. Two notes on state:

- **CI is not green yet** — `Type Check`, `Docker Build Smoke Test`, `PHP Unit Tests`, `Scaffolded Adapter Builds` pass; `Lint`, `Test`, `Build` and `Integration Tests` are **pending** at the time of writing. `Lint` is the one that matters here, since it is what runs `check-architecture-gates.mjs` against the merged 8-under-9 count. Please confirm it goes green before merging rather than reading this approval as covering it.
- Whichever of #3208 / #3198 lands second still owns the delete/override reconciliation; I have left the matching note on #3198.

---

norbert-kulus-blockydevs:
Fixed in `c8de6da85`, and the branch is levelled with `main` (28 commits behind, merged clean, no conflicts).

## [IMPORTANT] the docblock claimed the gap was tracked

You are right, and I verified both halves rather than taking the finding on trust:

- `LocationService.deleteLocation:99-102` refuses on `countPositionsAtLocation > 0` and nothing else, so a location referenced solely by an override deletes cleanly.
- Searching the repo for `stockLocationOverride` returns #3206 and #3207, and #3207 is the frontend field only. Nothing covers the dangling-override case.

So the sentence was a false claim about the repo's own state, and your framing is the reason it matters: a docblock saying a gap is tracked retires the next reader's attention on the strength of a lookup that comes back empty.

I took the second of the two options you offered - saying it plainly rather than filing an issue - because creating one is the repo owner's call, not mine. It now reads:

> That is DEFERRED AND UNTRACKED - there is no issue for it, deliberately said here rather than left as a claim a reader would verify with an empty search. #3207 covers the frontend field only. Whoever takes the next slice of this area owns the decision; do not read this paragraph as work already queued somewhere.

The paragraph above it also now states the consequence concretely - the config keeps naming a row that no longer exists, and nothing re-checks it - instead of leaving the reader to infer it. The deferral itself is unchanged; no code moved.

Say the word and I will file the issue and swap the wording for its number.

## On your CI note

Taken. `Lint` is the one that matters here, and I ran `check-architecture-gates.mjs` locally at this head:

```
✓ check-architecture-gates --self-check: 13 case(s) passed.
✓ check-architecture-gates: gates marked in ADR-048+, 9/10 config knobs, 1/3 ladder rungs.
```

9 under a threshold of 10. Worth recording why that is 10 and not 9: the merge with `main` brought a second knob raise, so three independent raises met here - #3142's, #3159's `readConnectionCurrency`, and this PR's `readStockLocationOverride`. The gate is `>=`, so 9 knobs under 9 would have tripped it; I set 10 and proved a tenth still trips by adding a probe entry and watching it fire, then removing it. That reasoning is recorded in the script's own docblock rather than only here.

The earlier red `Build` / `Test` / `Lint` / `Integration Tests` on this PR were runs I cancelled myself while freeing self-hosted slots for the price-changes and sales-documents stacks - not failures. The level-with-main push has restarted them all.

PR #3209, opened 2026-09-12
feat(web): warn when a sales-document rule's destination issues nothing

### LINKED ISSUE #3178: feat(web): warn when a rule's destination issues nothing
## Merge strategy — 🔗 STACKED, on top of #3177

**Not direct to `main`.** This slice ships inside the #3173 stack (#3174 → #3175 → #3179 → #3176 → #3177 → #3178 → #3180 → #3181 → #3182 → #3183 → #3184 → #3185 → #3186): PR #3209 targets #3177's branch, not the default branch. One consequence is worth stating plainly — GitHub honours `Closes #3178` only on a merge into `main`, so this issue does **not** auto-close when its own PR merges; it closes when the chain reaches `main`, or by hand once it has.

---

Part of #3173.

## Context
On the running instance three PL rules route to connections whose `Issues` column is set to **Nothing** — which the provider table itself describes as "not a routing candidate at all". Nothing on the rules list says so; the only hint is a market card reading "Nothing issued", which looks like a volume statistic.

## Scope
A list-level alert on the market's rules list when any rule names a connection that issues nothing. Names the connection and the remedy.

## Acceptance
- [ ] `data-testid="rules-destination-warning"` present exactly when at least one rule is affected
- [ ] Names the offending connection

**Depends on**: #3177 (the slice below it in the stack).

norbert-kulus-blockydevs:
## Summary

- Adds a `rules-destination-warning` alert to the per-country sales-document rules list (`SalesDocumentRulesList`), shown when at least one rule names a connection currently set to issue **Nothing** under Connected providers (`config.salesDocument.documentKind` unset), or that carries neither `Invoicing` nor `Fiscalization` at all — a connection that is not a routing candidate at all, per `AutoIssueTriggerService`.
- `findSalesDocumentDestinationWarnings` (new, `apps/web/src/features/sales-documents/lib/`) is a pure, display-only mirror in the same posture as the existing `detectSalesDocumentConflict` — it reuses `deriveSalesDocumentRows` (the same coercion the centralized Connected-providers table already uses) and collapses several rules naming the same dead connection into one warning, naming the connection and the count of affected rules.
- `data-testid="rules-destination-warning"` is taken verbatim from `docs/plans/mockups/sales-document-rule-composer.html`'s test-hooks table (step 03 / screen S2).
- The predicate reads a declared config fact (`documentKind` / `enabledCapabilities`), never a destination's technical acceptance — it renders a `tone="warning"` alert only and gates nothing.

## Test plan

- [x] `pnpm --filter @openlinker/web type-check` — passes with zero errors
- [x] `pnpm lint` (root, chains `pnpm -r lint && pnpm check:invariants`) — ran via the pre-commit hook, zero errors (pre-existing warnings only)
- [x] New unit tests: `find-sales-document-destination-warnings.test.ts` (pure-function coverage) and `sales-document-rules-list.test.tsx` (presence/absence of the warning)
- [ ] Not run locally per project policy (`pnpm test`/vitest/jest) — CI runs the suite

Closes #3178


piotrswierzy (commented):
## Tech Lead review

Adds a list-level `rules-destination-warning` to `SalesDocumentRulesList` when a rule names a connection that Connected providers currently has issuing **Nothing**. The shape is right: a pure function in `lib/`, display-only, no adapter resolved, no gate — the stated `detectSalesDocumentConflict` posture; collapsed per distinct connection rather than per rule; the count explicitly scoped to the supplied rules and documented as such; the `data-testid` taken verbatim from the mockup's test-hooks table. The component's pre-existing `connectionsQuery.isLoading` / `.error` guards also mean a failed connections read renders the error state rather than warning about every rule, which is the failure mode this kind of derived alert usually has. One real gap: the predicate is missing half of the candidate test it says it mirrors.

### Findings

**[IMPORTANT]** — `apps/web/src/features/sales-documents/lib/find-sales-document-destination-warnings.ts:~50`
> The header cites `AutoIssueTriggerService` as the authority for "not a routing candidate at all", and `docs/architecture-overview.md` § 14 Invoicing describes that gate as *"lists **ACTIVE** connections with EITHER `Invoicing` or `Fiscalization` enabled"*. The predicate here only tests `documentKind !== null` via `deriveSalesDocumentRows`, which deliberately does **not** filter on status (its own docblock: *"or disabled entirely — capability enablement is a separate axis from connection `status`"*). So a rule naming a **disabled** connection that still carries a role gets **no warning**, even though it can never route for exactly the reason this alert exists to surface — a false negative in the one direction that matters, since the operator sees a clean rules list and finds out when an order is held. Note #3210's `sales-document-country-defaults.tsx` filters `row.status === 'active'` for the same question, so the two halves of this stack currently disagree. Add the status test (and consider whether the copy should distinguish "issues Nothing" from "is disabled" — they have different remedies).

**[SUGGESTION]** — `find-sales-document-destination-warnings.ts:~62` and its test `should fall back to the connectionId when the connection cannot be found`
> `sales_document_rules.connection_id` is `FOREIGN KEY … ON DELETE CASCADE` (`1838000000000-create-sales-document-rule-tables.ts:72`), so a rule naming a connection that no longer exists cannot persist — the fallback branch is unreachable today, and the test pins a state the schema forbids. It is also the one branch that produces a factually wrong sentence if it ever *is* reached: `"conn_7f3a… is set to issue Nothing under Connected providers"` asserts a configuration for a connection that is not in the list at all, and renders a raw UUID where the operator expects a name. Either drop the branch and let the type reflect that the connection is always resolvable, or give the not-found case its own sentence rather than borrowing the "issues Nothing" one.

**[SUGGESTION]** — `find-sales-document-destination-warnings.test.ts:1`
> No file header. `docs/engineering-standards.md` § File Headers requires one on all source files, and the sibling new test in this same PR (`sales-document-rules-list.test.tsx`) has one — so this is an inconsistency inside the diff rather than a repo-wide question.

**[SUGGESTION]** — `find-sales-document-destination-warnings.ts`
> An **expired** rule (`effectiveTo` in the past) naming a role-less connection raises the same warning as a live one. That rule cannot route regardless of the destination's role, so the alert is noise on a market whose history is intact. Worth filtering to rules effective now, or at least saying in the header that expiry is deliberately ignored.

**[SUGGESTION]** — PR description / `docs/plans/mockups/sales-document-rule-composer.html`
> The copy diverges slightly from the mockup's — the mockup reads *"so the two rules above cannot route"*, the implementation *"so 2 rules naming it cannot route"* (which is more accurate, since not every rule above is affected), and the plural title `"{n} destinations are not issuing anything"` has no mockup counterpart. `docs/frontend-architecture.md` § UX Mockups makes this reviewer-enforced: an implementing PR must either touch the mockup or say in its description why it is still accurate. One sentence in the body, or a one-line mockup edit, closes it.

**[SUGGESTION]** — CI
> No checks reported on this branch at all, and the PR body notes the test suites were not run locally. Two new test files ship here; worth getting a green run before merge.

Done correctly and worth naming: folding "no row at all" (neither capability enabled) into the same warning rather than a second case, with the reason stated at the `?? null`; scoping `affectedRuleCount` to the supplied rules with an explicit *"never a global count"* on the field; and deriving eligibility through `deriveSalesDocumentRows` — the same coercion the centralized Connected-providers table uses — instead of re-reading `config.salesDocument` inline, so the alert and the table it points the operator at cannot disagree about what a connection issues.

### Verdict
🔄 **Approve with changes**

1. Add the `status === 'active'` half of the candidate predicate (or document why it is deliberately excluded).
2. Optional: not-found branch, test file header, expired-rule filtering, mockup note.

---

norbert-kulus-blockydevs:
## Review addressed — commit `427881bf0`

- **[IMPORTANT] Predicate missing the `status === 'active'` half** — *fixed*. The function now resolves the connection itself (not just its `deriveSalesDocumentRows` row) and tests status **before** role, matching `detect-sales-document-conflict.ts:35` and `sales-document-country-defaults.tsx:89`. Per your parenthetical the copy is split by remedy: `not-active` gets its own status-specific sentence (`is disabled` → "Enable it", `needs reconnecting` → "Reconnect it", `is in an error state` → "Fix the connection") rather than the "give it a role" one. Status wins when a connection fails both halves, mirroring the runtime order.
- **[SUGGESTION] Unreachable "connection cannot be found" fallback** — *fixed*, second of your two options. FK verified (`1838000000000-create-sales-document-rule-tables.ts:72-73`, `ON DELETE CASCADE`). The branch keeps its own reason (`unknown-connection`) and its own sentence, so it no longer asserts a configuration for a connection absent from the list. Kept rather than dropped because the rules and connections queries are cached independently, so a stale connections list can reach it even though the schema forbids the deleted-connection case. The pinning test now asserts the reason and that the sentence never says "issue Nothing".
- **[SUGGESTION] Missing file header on the test** — *fixed*, matching the sibling component test's style.
- **[SUGGESTION] Expired rules raise the same warning** — *fixed*, filtered. A rule whose `effectiveTo` has passed is skipped; a rule whose `effectiveFrom` is in the future still warns (it will route soon and is fixable now), and an unparseable `effectiveTo` is never read as expired, since suppressing is the wrong failure direction. `now` enters as an optional parameter (the `RefreshToken.isActive(now)` precedent) so the function stays pure.
- **[SUGGESTION] Copy diverges from the mockup** — *fixed by editing the mockup*. It **is** on this branch (#3199 is merged in), so line 697's "so the two rules above cannot route" → "so 2 rules naming it cannot route" — the implementation's wording is the accurate one — and the test-hooks row records the widened predicate. The plural title and the two new sentences are implementation states the single mockup panel does not enumerate, which the function header now says explicitly.
- **[SUGGESTION] CI did not run** — *handled*: gate re-run at the stack tip, result below.

---

piotrswierzy (commented):
## Tech Lead re-review (`5f94f8c4` → `6ad1e98d`)

Every finding is closed, and the IMPORTANT one is closed better than "add the missing test". The predicate now mirrors **both** halves of the `AutoIssueTriggerService` gate, and the two failures are reported as *different reasons* (`not-active` vs `issues-nothing`) on the explicit ground that they have different remedies — with `status` winning when a connection fails both, mirroring the runtime order. `INACTIVE_STATUS_COPY` then gives each of `disabled` / `needs_reauth` / `error` its own state-and-remedy sentence, and `InactiveConnectionStatus = Exclude<ConnectionStatus, 'active'>` means a fourth status becomes a compile error rather than a missing arm. The warning union is now discriminated and `SalesDocumentDestinationWarningReason` is derived from it, so the reason set and its copy table cannot drift.

The three suggestions landed too, each in the stronger of the two available shapes. The not-found branch is **kept with its own sentence** — and the header now justifies keeping it on a real ground I had not considered: the two queries are cached independently, so a connections list fetched before a connection existed still renders that connection's rules. That is a better answer than my "the FK makes it unreachable, drop it". `hasExpired` filters closed windows and deliberately keeps not-yet-started ones, with the unparseable-date case failing toward *warning* rather than suppressing — the correct direction, stated at the branch. The test file has a header. And the mockup reconciliation is stated in the header rather than left to a reviewer to notice.

### Findings

**[SUGGESTION]** — `find-sales-document-destination-warnings.ts:~150` (`describeSalesDocumentDestinationWarningTitle`)
> The title is now shared across three reasons but still reads *"{n} destinations are not issuing anything"*. A `not-active` connection may well carry a perfectly good role, and an `unknown-connection` says nothing about roles at all — so on a mixed alert the title states something the per-warning sentences beneath it contradict. The per-reason sentences are accurate; it is only the heading that generalised from the original single case. Something like *"{n} destinations cannot route"* covers all three without weakening the common one.

**[SUGGESTION]** — same file, and `sales-document-rule-composer-dialog.tsx:154`
> The stack has converged two of three answers to "which connection may a routing decision name" — this predicate and #3210's country-default select now agree. The composer dialog still uses `selectInvoicingCandidates` / `selectFiscalizationCandidates`, which *does* filter `status === 'active'`, so the residue is **capability vs role**. That leaves a closed loop reachable without any misconfiguration: the composer offers a connection carrying `Invoicing` but no `config.salesDocument.documentKind`, the rule saves, and this alert then fires on the rule the dialog invited. Worth an issue rather than leaving it to the three comments that currently name it.

### Verdict
✅ **Approve**

`mergeable: MERGEABLE`, `mergeStateStatus: BEHIND` (rebase before merge). CI is running on this head — `Detect changed paths` and `PHP Unit Tests` pass; `Lint`, `Type Check`, `Test`, `Build`, `Integration Tests` and `Docker Build Smoke Test` are **pending**, so the two new test files are not yet confirmed green. Report them as unknown, not passing.

---

norbert-kulus-blockydevs:
## Suggestions addressed — commit `b8749cf01`

- **[SUGGESTION] The heading generalised from the original single case** — *fixed*, with your wording. `One destination cannot route` / `{n} destinations cannot route`. You are right that it was the only part that was false: "is not issuing anything" holds for a role-less destination and fails for both others, so on a mixed alert the heading contradicted the sentences beneath it. "cannot route" is the property all three share and it does not weaken the common case. The reason is recorded in a docblock on the function rather than only in this thread, and the mockup's own heading (`sales-document-rule-composer.html:708`) was updated in the same commit, per the UX-mockups rule.
- **[SUGGESTION] The composer dialog still answers the same question a third way** — *tracked*, as you asked: **#3232**. It carries the loop you described (composer offers a connection with `Invoicing` but no `documentKind` → rule saves → this alert fires on the rule the dialog invited), records that the residue is now **capability vs role** rather than capability vs nothing since the composer does filter status, and adds the fourth surface neither review named — `sales-document-template-screen.tsx:78-79`, which uses the same capability pair. Acceptance criteria allow either convergence or a stated reason for deliberate divergence, since "assign a role and a rule in one pass" may be a legitimate reason for the composer to offer the wider set.

---

piotrswierzy (approved):
## Tech Lead re-review (`6ad1e98d` → `dec6f423`)

One own commit (`b8749cf0`); everything else in the range is base catch-up through `3177-country-default-single-select`. The commit closes the one note I left, and closes it in the shape that made the note worth making. **The prior approval holds.**

The shared heading is now *"One destination cannot route"* / *"{n} destinations cannot route"*, and the docblock above it states the reasoning rather than just the outcome: "is not issuing anything" was true of a role-less destination and **false** of the other two — a `not-active` connection may carry a perfectly good role, and `unknown-connection` says nothing about roles at all — so on a mixed alert the heading contradicted the per-warning sentences beneath it. "cannot route" is the property all three share, which is exactly the right axis to generalise on: it is the thing the operator needs to know and the only claim that survives every reason.

Two details make this better than a copy tweak. The mockup (`docs/plans/mockups/sales-document-rule-composer.html`, `rules-destination-warning`) is updated **in the same commit**, so `docs/frontend-architecture.md` § UX Mockups is satisfied structurally rather than by a promise — and that obligation is reviewer-enforced only, so a PR that changed the string and left the mockup would have passed every gate. And the unit test moved with it, so the copy has one assertion rather than none.

Re-verified at the new head, since a merge can quietly undo things:

- The predicate is structurally unchanged and still mirrors both halves of the `AutoIssueTriggerService` gate — `unknown-connection` → `not-active` → `issues-nothing`, status winning when a connection fails both, `InactiveConnectionStatus = Exclude<ConnectionStatus, 'active'>` keeping a fourth status a compile error, and `SalesDocumentDestinationWarningReason` still derived from the union so the reason set and its copy table cannot drift. `hasExpired` still fails toward *warning* on an unparseable window.
- Every `className` literal in `sales-document-rules-list.tsx` resolves to a real declaration in `apps/web/src/index.css` at this head — `page-section`, `detail-section__title-row`, `eyebrow`, `button--sm`, `muted-text`, `rule-card`, `rule-card__flow`, `condition-chip`, `rule-card__arrow`, `rule-card__result`, `rule-card__meta`, `provenance-tag`, `rule-card__dates`. No invisible UI here.
- `features/sales-documents` is not a `scripts/check-ui-vocabulary.mjs` scan root, so that gate has no opinion on this copy either way — worth knowing rather than assuming it was covered.

**The convergence suggestion is now properly discharged**: #3232 is open and states the loop precisely — the composer offers on **capability**, this predicate warns on **role**, so one surface invites the rule the other flags. That is what I asked for; it no longer survives on comments alone.

### Findings

### CI state

`Type Check`, `Docker Build Smoke Test`, `Detect changed paths` and `PHP Unit Tests` pass. `Lint`, `Test`, `Build` and `Integration Tests` are **queued**, so the two changed test files are not yet confirmed green — unknown, not passing. `mergeable: MERGEABLE`, `mergeStateStatus: BLOCKED` (stacked).

### Verdict
✅ **Approve**

---

piotrswierzy (approved):
## Tech Lead re-verification (head unchanged at `da9f46bf`)

Position **6 of 18** in the `#3173` stack. **Nearest blocking ancestor: #3201** (position 4, `REVIEW_REQUIRED`) — unchanged from the last round, and still the chain's first hard blocker. `MERGEABLE`/`BLOCKED` (that state is the unreviewed ancestor, not this PR). CI in flight at run `34845828689`: `Docker Build Smoke Test`, `Detect changed paths` and `PHP Unit Tests` pass, `Lint` / `Test` / `Type Check` / `Build` / `Integration Tests` **pending**. **No failures.**

Predicate re-verified at this head and intact: `type InactiveConnectionStatus = Exclude<ConnectionStatus, 'active'>`, `not-active` and `issues-nothing` as separate discriminated reasons with `not-active` taking precedence when a connection fails both halves (pinned by its own case), and `INACTIVE_STATUS_COPY: Record<InactiveConnectionStatus, …>` — so a fourth `ConnectionStatus` is a compile error rather than a silently uncovered status. `sales-document-rule-composer.html` is updated in the same PR, so the corrected heading, its test and the mockup move together.

### Findings

### Verdict
✅ **Approve**

---

PR #3210, opened 2026-09-12
fix(sales-documents): unify country default to one row per country

### LINKED ISSUE #3177: fix(sales-documents): country default becomes a single choice
## Merge strategy — 🔗 STACKED, on top of #3176

**Not direct to `main`.** This slice ships inside the #3173 stack (#3174 → #3175 → #3179 → #3176 → #3177 → #3178 → #3180 → #3181 → #3182 → #3183 → #3184 → #3185 → #3186): PR #3210 targets #3176's branch, not the default branch. One consequence is worth stating plainly — GitHub honours `Closes #3177` only on a merge into `main`, so this issue does **not** auto-close when its own PR merges; it closes when the chain reaches `main`, or by hand once it has.

---

Part of #3173.

## Context
A market has two independent default selects — invoice and receipt. Setting **both** disables the fallback step **entirely**: an order matching no rule is held rather than taking either default. The product detects this and explains it after the fact.

## Scope
Replace the two selects with **one**, whose options are `Nothing — hold the order` or any single connection with a role. The broken state becomes unexpressible rather than detected.
- Migration for installs that currently have both set
- Readback line stating the consequence in plain language

## Acceptance
- [ ] Two defaults cannot be set simultaneously by any path, including raw config
- [ ] `Nothing — hold the order` is a first-class option, with its consequence stated
- [ ] Existing single-default installs are unchanged

**Depends on**: #3176 (the slice below it in the stack).

norbert-kulus-blockydevs:
## Summary

`sales_document_country_defaults` was unique on `(country, documentKind)`, so a country could hold both an invoice default and a receipt default at once. That state was never useful in practice: `evaluateSalesDocumentRules` already read two defaults for one country as `ambiguous-defaults` -> `unresolved`, disabling the fallback tier entirely with no explanation on the FE. This PR moves uniqueness onto `country` alone, so a country now maps to **at most one** `SalesDocumentCountryDefault` row regardless of `documentKind` — the two-defaults state becomes structurally unexpressible instead of merely detected and warned about after the fact.

- **Migration + ORM entity**: drops `UQ_sales_document_country_defaults_country_kind`, adds `UQ_sales_document_country_defaults_country` (on `country` alone). The migration's `up()` first de-duplicates any existing `(country, documentKind)` pairs by keeping the most-recently-updated row per country, since "most recently updated" is the only available proxy for operator intent and the prior state was already indistinguishable from having neither default configured.
- **The migration's `down()` deliberately does NOT reverse the de-duplication.** It restores the old two-column index shape, but the rows it deleted in `up()` are gone for good — nothing on the surviving row can reconstruct the one that was removed, so restoring the old index would just re-admit a shape with no way to repopulate its second row. This mirrors the `1840000000000-reset-fx-stamp-for-mislabelled-prestashop-orders` migration's own `down()`, which takes the same "the forward fix is what re-derives this state, not a rollback" posture.
- **Repository**: `upsert`'s conflict target moves to `country` alone, so a second upsert under a different `documentKind` for the same country now overwrites the existing row instead of inserting a sibling; `findByCountryAndKind` ignores its `documentKind` argument (kept for callers that still supply one).
- **Frontend**: `SalesDocumentCountryDefaults` collapses from two independent selects (Invoice / Receipt) to one — "Nothing — hold the order", or any connection carrying a `salesDocument` role — matching the `sales-document-rule-composer.html` mockup's "If no rule matched" tier and its `country-default` test hook (documented there as "Single-valued — the two-defaults state is unreachable"). `SalesDocumentCountryRoutingDialog` no longer renders the dual-default warning tier, since that state can no longer occur; its docblock and tests are updated to match.
- **Ripple**: `sales-document-rule-engine-panel.test.tsx` scopes its `DE`-row lookups to the pre-existing `"Sales-document markets"` list (`within(marketList)`), because the new single-select readback sentence also renders the country code, which made a bare `screen.getByText('DE')` ambiguous once the routing dialog is open.

Closes #3177
Refs #3173

## Test plan

- [x] `pnpm lint` — zero errors
- [x] `pnpm type-check` — zero errors
- [x] Pre-commit hook (`lint` -> `type-check` -> `smart-test --no-integration`) passed, including the affected `sales-document-rule-engine-panel.test.tsx` / `sales-document-country-defaults.test.tsx` / `sales-document-country-routing-dialog.test.tsx` suites
- [x] `check-repo-urls.mjs`, `check-adr-numbers.mjs`, `check-architecture-gates.mjs`, and the full `pnpm check:invariants` chain — all clean
- [ ] CI


piotrswierzy (commented):
## Tech Lead review

Moves `sales_document_country_defaults` uniqueness from `(country, documentKind)` to `country` alone, and collapses the two-select FE tier into one. The reasoning is right and well-argued: the two-defaults state was never useful (`evaluateSalesDocumentRules` already read it as `ambiguous-defaults` → `unresolved`), so making it unexpressible strictly dominates detecting-and-warning. Migration mechanics are sound — the old uniqueness is a plain `CREATE UNIQUE INDEX` in `1838000000000-create-sales-document-rule-tables.ts`, not a `UNIQUE` constraint, so `DROP INDEX` is the correct verb; the ORM `@Index` name matches the migration's, which keeps the `synchronize`-built integration schema and the migration schema in agreement; `1878000000001` sorts after `main`'s tail (`1877000000000`) and does not collide with #3201's `1878000000000` in the same stack. My main concern is one class the CSS does not declare, and one port method that now lies about its own signature.

### Findings

**[IMPORTANT]** — `apps/web/src/index.css` / `sales-document-country-defaults.tsx:~150`
> `<div className="readback" data-testid="country-default-readback">` applies a class that has **zero** declarations in `apps/web/src/index.css` (`grep -c '\.readback' → 0`). In the mockup that region is a bordered, tinted callout; in the app it renders as undifferentiated body text under an `eyebrow`. Same problem one line up: `.hint` is declared only as `.bulk-editor__field .hint`, so the bare `<p className="hint">` outside that scope is unstyled too — which is why every other `className="hint"` call site in `bulk-edit-modal.tsx` carries a compensating inline `style`. `docs/frontend-architecture.md` § Design tokens is explicit that component CSS is written against `index.css`; a mockup-only class name copied across without its declaration is the drift that guard exists to prevent. Either declare `.readback` (and an unscoped `.hint`) in `index.css`, or use `muted-text`, which this same component already uses correctly a few lines below.

**[IMPORTANT]** — `libs/core/src/sales-documents/domain/ports/sales-document-country-default-repository.port.ts:25` + `…/repositories/sales-document-country-default.repository.ts:46`
> `findByCountryAndKind(country, _documentKind)` now discards its second argument and answers by country alone, so `findByCountryAndKind('PL', 'invoice')` can return a `fiscal-receipt` row. It has **zero production callers** (`grep -rn findByCountryAndKind --include=*.ts libs apps` → only the port and its implementation). This is exactly the shape `docs/architecture-overview.md` § Orders records as a deliberate deletion: `OrderHoldRepositoryPort.listOpenHolds` was *"DELETED rather than re-documented — it had zero callers and its docblock advertised … which is false … leaving the method would have left that trap on the port for the next reader to reach for."* Here the trap is sharper than a false docblock — the signature still takes a kind and the parameter is prefixed `_` to silence the compiler. Delete it from the port, the repository and the two spec blocks; a future caller that genuinely needs it can use `findByCountry` and read `documentKind` off the single row.

**[SUGGESTION]** — `apps/web/src/features/sales-documents/components/sales-document-country-defaults.tsx` (readback) and its test
> `"…gets a {kind} through …"` renders `gets a Invoice`. The test asserts that string verbatim, so it's pinned rather than accidental. The mockup only ever shows the `Receipt` branch, so it never surfaced there. Either pick the article per kind, or reword to `…is issued as an Invoice through X` / `…gets an Invoice`.

**[SUGGESTION]** — `apps/api/src/migrations/1878000000001-…ts` `up()`
> The dedupe `DELETE` drops the loser row with no operator-facing record — no `RAISE NOTICE`, no row count in the log. For a *fiscal-routing* setting that is a configuration an operator chose, even if the pair was jointly inert. A `RAISE NOTICE` naming the deleted `(country, document_kind)` pairs would cost one statement and leave a migration-log breadcrumb when someone asks where their receipt default went. The irreversible `down()` is well-argued and the `1840000000000-reset-fx-stamp-…` precedent cited is a real one — no objection to that half.

**[SUGGESTION]** — same file, candidate predicate
> Eligibility here moves from capability (`selectInvoicingCandidates` / `selectFiscalizationCandidates`) to role (`deriveSalesDocumentRows` + `documentKind !== null`). `sales-document-rule-composer-dialog.tsx:154` — same feature, same screen, same question ("which connection may a routing decision name") — still uses the capability predicate. The removed docblock line claimed the old filter *"match[ed] `resolveSalesDocumentRouting`'s own structural capability check"*; nothing in this diff replaces that mirror. #3209 warns about the rules half, so the gap is covered operator-facing, but converging the two predicates would be worth a follow-up rather than leaving one feature with two answers.

**[SUGGESTION]** — CI
> Only `Scaffolded Adapter Builds` reported on this head; the web/unit suites did not run, and the PR body leaves `CI` unchecked. Four test files change here (three edited, one new) — worth confirming they go green before merge rather than relying on the pre-commit `smart-test` run.

Done correctly and worth naming: keeping `evaluateSalesDocumentRules`' `defaults.length > 1 → ambiguous-defaults` branch as a defensive arm rather than deleting it now that the DB forbids the state; retaining a *negative* regression test in the routing dialog spec (`should never render a dual-default warning`) instead of just deleting the three old ones; and the stale-current-connection option, which keeps a default whose connection lost its role switchable instead of silently vanishing from the list.

### Verdict
🔄 **Approve with changes**

1. Declare `.readback` (and an unscoped `.hint`) in `index.css`, or switch to `muted-text`.
2. Delete `findByCountryAndKind` rather than shipping a method whose argument is ignored.
3. Optional: readback grammar, migration `RAISE NOTICE`, predicate convergence follow-up.

---

norbert-kulus-blockydevs:
## Review addressed — commit `f0d8b0458`

- **[IMPORTANT] `.readback` / bare `.hint` are undeclared classes** — *fixed*. Both confirmed (`grep -c '\.readback' apps/web/src/index.css` → 0; `.hint` declared only at `index.css:12839` as `.bulk-editor__field .hint`, which is why all 10 `className="hint"` sites in `bulk-edit-modal.tsx` carry compensating inline styles). Took the smaller of your two options: both now use `muted-text`, matching this component's own line 176 and every sibling in the feature.

  Worth knowing why the callout was not reproduced: the mockup's `.readback` rule does exist (`sales-document-rule-composer.html:388`) but is built on `--info-border` / `--info-soft` / `--info-fg`, tokens `index.css` does not define — reproducing it would mean inventing design tokens, which is a design decision rather than a review fix. A comment records that. Every remaining `className` in the file now resolves to a real declaration.
- **[IMPORTANT] `findByCountryAndKind` ignores its second argument** — *deleted*, per the `OrderHoldRepositoryPort.listOpenHolds` precedent you cited. Zero callers verified independently (`grep -rn findByCountryAndKind --include=*.ts --include=*.tsx .` → port, impl, one repo-spec `describe`, one jest mock stub). Removed from the port, the repository, the spec's `describe` block (3 tests) and the `makeCountryDefaultRepo()` mock in `sales-document-rules.service.spec.ts` — the mock had to go too, since it is an object literal typed `jest.Mocked<Port>` and excess-property checking would have failed. Post-delete grep returns nothing.
- **[SUGGESTION] `gets a Invoice`** — *fixed*. Added a per-kind `documentKindArticle` helper (`an` / `a`) rather than rewording, since "is issued as an Receipt" just moves the same problem to the other branch. The one pinned assertion at `sales-document-country-defaults.test.tsx:188` is updated; the `Receipt` assertion at :101 is unchanged and still correct.
- **[SUGGESTION] Dedupe `DELETE` leaves no operator record** — *fixed*. Wrapped in a `DO $$ … RAISE NOTICE … END$$;` block that captures the removed rows through a `DELETE … RETURNING` CTE and names the `(country, document_kind)` pairs — the shipped repo idiom (`1788000000001-rename-marketplace-capability.ts:64`, `1780000000000-add-enabled-capabilities-to-connections.ts`), not an invention. `down()` untouched. Note the file is now `1881000000001-…` after the branch was renumbered above main's tail.
- **[SUGGESTION] Capability-vs-role predicate divergence** — *documented, not converged*. A docblock at the predicate states the mirror precisely: `deriveSalesDocumentRows` already gates on `Invoicing`/`Fiscalization` being enabled, and `documentKind !== null` mirrors `resolveSalesDocumentRouting`'s `isEligibleCandidate`. It also records what is **not** mirrored — that resolver's step-6 kind↔capability pairing, so a connection whose role contradicts its capability is still offered here and refused at routing time — and names `sales-document-rule-composer-dialog.tsx` as the divergent sibling. Converging the two is a follow-up, and it is wider than you noted: `sales-document-template-screen.tsx:78-79` uses the same capability pair.
- **[SUGGESTION] CI did not run** — *handled*: gate re-run at the stack tip, result below.

---

piotrswierzy (commented):
## Tech Lead re-review (`e889fcc8` → `dfb39597`)

Both IMPORTANT findings are closed, and both suggestions with them. `findByCountryAndKind` is **deleted** from the port and the repository rather than re-documented — the `OrderHoldRepositoryPort.listOpenHolds` treatment `docs/architecture-overview.md` § Orders records, which is the right call given it had zero production callers and a signature that had started lying about its own second argument. The `.readback` class is gone; the readback region now renders through `eyebrow` + `muted-text`, the classes this component and its siblings already use, with an in-place comment saying the mockup's tinted callout is built on `--info-*` tokens `index.css` does not declare and that inventing them was declined. That is the right resolution of the two available ones and it is stated where the next reader will hit it rather than left to be rediscovered. `.hint` is gone with it, so the compensating-inline-style pattern is not reproduced here. The article is fixed properly (`documentKindArticle`, `an Invoice` / `a Receipt`) rather than by rewording around it, and the migration's dedupe `DELETE` now emits a `RAISE NOTICE` naming the pairs it dropped, following the `1788000000001-rename-marketplace-capability` shape.

Migration numbering moved with the stack (`1878000000001` → `1881000000001`) and still sorts immediately after #3201's `1881000000000`, which is the ordering that matters: #3201's defaults ranking partitions on `(UPPER(country), document_kind)` and must run while that index is still live, before this migration replaces it with `UQ_sales_document_country_defaults_country`.

### Findings

**[SUGGESTION]** — `sales-document-country-defaults.tsx:~100` (candidate predicate) and `sales-document-rule-composer-dialog.tsx:154`
> The cross-cutting divergence I raised has converged **two of three** ways: this select and #3209's warning predicate now both read `status === 'active' && documentKind !== null`, and both carry a comment naming the other. The composer dialog still answers the same question with `selectInvoicingCandidates` / `selectFiscalizationCandidates` — which *does* filter `status === 'active'`, so the remaining gap is narrower than it was: **capability vs role**. The concrete consequence is a small loop an operator can walk into unaided — the composer offers a connection carrying `Invoicing` but no `config.salesDocument.documentKind`, the rule saves, and #3209 then warns about the rule the dialog just invited. Both PRs name it as a follow-up rather than papering over it, which is the honest disposition; worth an issue so it does not survive on comments alone.

**[SUGGESTION]** — `1881000000001-…ts` `down()`
> Unchanged and still well-argued (the `1840000000000-reset-fx-stamp-…` precedent is real). No objection — restating only so the irreversibility is a deliberate, reviewed property of the stack rather than an unexamined one.

### Verdict
✅ **Approve**

`mergeable: MERGEABLE`, `mergeStateStatus: BEHIND` (rebase before merge). CI is running on this head — `Detect changed paths`, `PHP Unit Tests` and `Scaffolded Adapter Builds` pass; `Lint`, `Type Check`, `Test`, `Build`, `Integration Tests` and `Docker Build Smoke Test` are **pending**, so the four changed test files are not yet confirmed green. Merge after #3201, never before it.

---

norbert-kulus-blockydevs:
## Suggestions addressed

- **[SUGGESTION] The predicate has converged two of three ways** — *tracked*: **#3232**. It carries your framing — the residue is **capability vs role**, not capability vs nothing, since the composer does filter `status === 'active'` — and the concrete loop (composer offers a connection with `Invoicing` but no `documentKind`, the rule saves, #3209 then warns about the rule the dialog just invited). It also adds a fourth surface neither review named: `sales-document-template-screen.tsx:78-79` uses the same capability pair. Acceptance criteria allow either convergence or a recorded reason for deliberate divergence, since "assign a role and a rule in one pass" may be a legitimate reason for the composer to offer the wider set. You were right that comments were the wrong place for it — three of them named each other and none of them would have been read.
- **[SUGGESTION] `down()` irreversibility restated** — *no change, and that is the point*: recording your restatement here so the property is a reviewed one rather than an unexamined one. The `1840000000000-reset-fx-stamp-…` precedent still governs.

No code change on this branch this round; it carries the stack's propagated merges, including #3201's improved skip report one file over.

---

piotrswierzy (approved):
## Tech Lead re-review (`dfb39597` → `cf3e0903`)

**This range contains no commits of this PR's own.** Both new commits are merges of `3176-normalize-market-country-codes`, carrying `main` (#3127, #2883) and the sibling stack — #3201's `12eacdb1` edit to `1881000000000-normalise-sales-document-country-codes.ts`, and #3200's one-line ADR-073 addition. So the review question is narrow: did the merges undo anything. They did not. **The prior approval holds.**

Re-verified at the new head rather than assumed:

- **`findByCountryAndKind` is still gone** from both the port and the repository. What remains is single-argument `findByCountry(country)` and its documented batch counterpart — no reintroduced two-argument form, which was the specific hazard (a method whose second argument was silently discarded, with zero production callers).
- **`.readback` is still absent.** No `className="readback"` anywhere; the region renders through `eyebrow` + `muted-text`, and the in-place comment explaining that the mockup's tinted callout is built on `--info-*` tokens `index.css` does not declare — and that inventing them was declined — survived the merges intact. Both classes are declared in `apps/web/src/index.css` at this head (lines 2123 and 1688), so this is verified, not inferred.
- **The entity and the migration still agree.** `@Index('UQ_sales_document_country_defaults_country', ['country'], { unique: true })` matches the index the migration creates, name and shape. Worth stating explicitly given the standing trap: `synchronize` emits neither PG enums nor CHECKs, so a constraint expressed either of those ways would hold in production and silently not in the harness — but it **does** emit indexes, so this uniqueness is live in the integration suite too. This one is not a production-only guarantee.
- The migration body is byte-unchanged: deterministic tie-break (`updated_at DESC, id DESC`), `IF EXISTS` / `IF NOT EXISTS` throughout, the `RAISE NOTICE` naming the dropped pairs, and the honest `down()` that restores the shape without pretending to restore the row.

### Migration ordering — correct today, and coupled to a collision this PR does not own

The ordering that matters still holds: #3201 `1881000000000` runs before #3210 `1881000000001`, which is required because #3201's ranking partitions on `(UPPER(country), document_kind)` and needs that index still live.

But `1881000000000` is claimed by **two** open PRs — #3201 and #3027 (`seed-erli-prompt-template`) — so `scripts/check-migration-timestamps.mjs` will fail one of them and one must move. **Which one moves decides whether this PR is still correct**: if #3027 renumbers, nothing here changes; if #3201 renumbers **upward past `1881000000001`**, this migration sorts before its own prerequisite and the dedupe runs against an index that has not been normalised yet. That is not a defect in this PR and needs no change to it — it is a merge-train constraint to hold: **settle the `1881000000000` collision in #3201's favour, or renumber this migration in the same move.**

For the wider picture while you are in here, neither belonging to these three PRs: #3159 and #3241 both claim `1884000000000`, and #3014's `1878000000000` sorts **below** `main`'s tail `1880000000000` — the rule-3 ordering violation `docs/migrations.md` § Troubleshooting 6 describes, which only surfaces on a from-scratch install.

### Findings

**[SUGGESTION]** — `sales-document-country-routing-dialog.tsx:308`
> `className="button--sm sales-document-country-routing-dialog__back"` — the `__back` modifier is **not declared anywhere** in `index.css` (the file declares `…__body`, `…__save-note` and `…__danger-zone`, and there is no `&__back` nesting either). To be clear about attribution: this is **pre-existing on `main` and untouched by this PR**, and the element still renders as a small button through `button--sm`, so only the modifier is inert. Raising it because it is exactly the shape this stack has shipped before and `check-design-tokens.mjs` is one-directional so nothing catches it — either declare the rule or drop the class, in whichever PR is already in that file.

**[SUGGESTION]** — the cross-cutting predicate question
> Still **2 of 3**, unchanged in this range: this select and #3209's warning predicate agree on `status === 'active' && documentKind !== null`, and the composer dialog still answers with `selectInvoicingCandidates` / `selectFiscalizationCandidates` — capability, not role. The in-place comment at `sales-document-country-defaults.tsx:~100` names the divergence and the file it lives in, which is the honest disposition. And it is now backed by **#3232**, open, which states the loop and covers all the surfaces — so this no longer rests on comments alone, which was my actual ask. Nothing to do in this PR.

### CI state

`Build`, `Test`, `Type Check`, `Docker Build Smoke Test`, `Scaffolded Adapter Builds`, `Detect changed paths` and `PHP Unit Tests` all pass. `Integration Tests` is **in progress** and `Lint` is **queued** — and `Lint` is the one that runs `check-migration-timestamps.mjs`, so the collision above is not yet confirmed either way by CI. `mergeable: MERGEABLE`, `mergeStateStatus: BLOCKED` (stacked). Merge after #3201, never before it.

### Verdict
✅ **Approve**

---

piotrswierzy (approved):
## Tech Lead re-verification (head unchanged at `7eb47c72`)

Position **5 of 18**. **Nearest blocking ancestor: #3201** (position 4, `REVIEW_REQUIRED`) — which is also this PR's direct base and its migration prerequisite. `MERGEABLE`/`BLOCKED` from that ancestor. CI in flight at run `34845890387`: `Docker Build Smoke Test`, `Scaffolded Adapter Builds`, `Detect changed paths`, `PHP Unit Tests` pass; five **pending**; **no failures**.

Deletions re-verified against the head tree rather than the diff: `findByCountryAndKind` is absent from **both** `sales-document-country-default-repository.port.ts` (which now declares only `findById` / `findByCountry` / `findByCountries` / `findAll` / `upsert` / `delete`) and the repository implementing it. It is still present on `main`, so this PR is genuinely the removal. The `.readback` CSS class is gone with the comment explaining why the component reuses existing classes rather than inventing `--info-*` tokens; `data-testid="country-default-readback"` is a test hook, not the class.

**The migration-ordering coupling resolves cleanly, and the premise I flagged last round is now stale.** Current timestamps:

| PR | Migration | |
|---|---|---|
| `main` tail | `1880000000000-create-price-change-episodes` | |
| **#3201** | `1880500000000-normalise-sales-document-country-codes` | prerequisite |
| #3027 | `1881000000000-seed-erli-prompt-template` | |
| **#3210** | `1881000000001-unify-sales-document-country-default-per-country` | this PR |

#3201 did **not** move up to `1886`/`1887` as two reviewers proposed — it moved **down**, to `1880500000000`, which is above `main`'s tail and below #3027. That resolves the #3027 collision *and* leaves `1880500000000 < 1881000000001`, so #3201 still sorts strictly before this migration and the collision guard here still runs while the `(country, document_kind)` uniqueness it tests is in place. **No change needed on this PR**, and the contingency I recorded ("if #3201 moves above `1881000000001`, #3210 must move in the same commit") is now moot at current heads. Worth re-checking only if #3201's timestamp moves again before it merges.

### Findings

### Verdict
✅ **Approve**

---

Issue #3186, opened 2026-09-11
feat(sales-documents): record and surface the matched routing rule


norbert-kulus-blockydevs:
## Merge strategy — 🔗 STACKED, on top of #3185

**Not direct to `main`.** This slice ships inside the #3173 stack (#3174 → #3175 → #3179 → #3176 → #3177 → #3178 → #3180 → #3181 → #3182 → #3183 → #3184 → #3185 → #3186): PR #3219 targets #3185's branch, not the default branch. One consequence is worth stating plainly — GitHub honours `Closes #3186` only on a merge into `main`, so this issue does **not** auto-close when its own PR merges; it closes when the chain reaches `main`, or by hand once it has.

The slice still leaves `main` observably better on its own — no dead code, no half-rendered feature — which is what made stacking safe rather than a workaround for an unsplittable change.

*Header amended after review: it previously claimed a direct-to-`main` merge the stack does not perform.*

---

Part of #3173.

## Context
Verified in code: `evaluateSalesDocumentRules` finds the matching rule and returns `{ kind: 'route', documentKind, connectionId }` — **the rule's identity is discarded inside the evaluator**. No `ruleId` column exists on the order, invoice or fiscal-registration record. We persist the reason a document was **not** issued and keep nothing when one **was**. So *"why did this order get an invoice?"* is currently unanswerable.

## Scope
- Carry the matched rule's identity on the `route` arm of `SalesDocumentDecision` (additive, pure function, cheap)
- Persist it alongside the existing block-reason columns on `order_records`, written by the same level-triggered writer and excluded from `toOrm` the same way — migration included
- A **collapsed-by-default** "Why this kind?" disclosure on the sales-document panel, beside the shipped "Why this document?", naming the rule and its conditions

## Acceptance
- [ ] `data-testid="sales-document-why-kind"` renders closed by default and opens on click
- [ ] Absent on a manually issued document
- [ ] The tax-ID evidence rows appear **only when the matched rule actually read that field**

**Depends on**: #3185 (the slice below it in the stack).

Issue #3174, opened 2026-09-11
design(sales-documents): commit the three epic mockups to the repo


norbert-kulus-blockydevs:
## Merge strategy — 🔗 BASE OF THE #3173 STACK

One branch off `main`, one PR (#3199), one merge — this slice really is direct. It is also the **base of a 13-PR stack** (#3174 → #3175 → #3179 → #3176 → #3177 → #3178 → #3180 → #3181 → #3182 → #3183 → #3184 → #3185 → #3186); every later slice targets its predecessor's branch rather than `main`, so this one must land first.

*Header amended after review: the stack was not named here, and a reader comparing the stated strategy against the open PRs found them disagreeing.*

---

Part of #3173.

## Context
The three target-design mockups exist only as published Artifacts. Per `docs/frontend-architecture.md` § UX Mockups the **repo-committed file is the durable spec**; the Artifact is design history. Every other sub-issue of #3173 cites these files.

## Scope
Commit three self-contained HTML files to `docs/plans/mockups/`:
- `sales-document-tax-number-on-receipt.html` — https://claude.ai/code/artifact/a4aed51c-6385-4d1c-9c94-bdc26eeab1c4
- `sales-document-rule-composer.html` — https://claude.ai/code/artifact/8bbce4d1-cbcc-46bd-a1b4-0ab8116b8dd9
- `sales-document-eparagony-invoicing.html` — https://claude.ai/code/artifact/4402f803-2013-4513-a201-be255a536e0d

## Acceptance
- [ ] All three open standalone in a browser, no network needed beyond the font CDN
- [ ] The "Show test hooks" toggle works in each
- [ ] Each carries its own task table and hook reference

**Depends on**: nothing.

norbert-kulus-blockydevs:
Created in error — not authorised. Closing.

PR #3211, opened 2026-09-12
feat(web): show the buyer tax ID on the order

### LINKED ISSUE #3180: feat(web): show the buyer tax ID on the order
## Merge strategy — 🔗 STACKED, on top of #3178

**Not direct to `main`.** This slice ships inside the #3173 stack (#3174 → #3175 → #3179 → #3176 → #3177 → #3178 → #3180 → #3181 → #3182 → #3183 → #3184 → #3185 → #3186): PR #3211 targets #3178's branch, not the default branch. One consequence is worth stating plainly — GitHub honours `Closes #3180` only on a merge into `main`, so this issue does **not** auto-close when its own PR merges; it closes when the chain reaches `main`, or by hand once it has.

---

Part of #3173.

## Context
`order_records.buyerTaxId` is populated and already decides which document an order gets — and appears on no screen. An operator cannot see why a tax-ID rule did or did not match.

## Scope
One row appended to the order's **Summary** block, in the existing `kv` component, rendering **three distinct states**:
- value present — shown verbatim, no reformatting, no country prefix
- asserted-none — a neutral pill
- not asserted by the source — muted text

Plus a statement that the whole feature is inert under `OL_STORE_PII=false`, where nothing is stored and the value reads back as not-asserted.

## Acceptance
- [ ] Three states carry three distinct `data-testid` values, not one hook with varying text
- [ ] Read through `decodeBuyerTaxIdColumn`; `IS NOT NULL` is never used
- [ ] The PII-mode statement is visible to an operator, not only in logs

**Depends on**: #3178 (the slice below it in the stack).

norbert-kulus-blockydevs:
## Summary

Adds the buyer tax ID row to the order detail **Summary** block (`/orders/:internalOrderId`), rendering the three-state `order_records.buyerTaxId` column honestly rather than collapsing it into a boolean:

- **present** — the id, verbatim, no reformatting/country prefix — `data-testid="order-buyer-tax-id"`
- **asserted-none** (source positively said the buyer has no tax id) — a neutral pill, "None — asserted" — `data-testid="order-buyer-tax-id-none"`
- **not asserted by the source** — muted text, with a keyboard-reachable tooltip stating the feature is inert under `OL_STORE_PII=false` (nothing is persisted there, so it reads exactly like this state) — `data-testid="order-buyer-tax-id-unknown"`

Matches `docs/plans/mockups/sales-document-tax-number-on-receipt.html` (screen S1), task E1-5 + E1-8 of that epic.

## Backend

`order_records.buyerTaxId` was never surfaced over HTTP before this change. Added:
- `OrderRecordResponseDto.buyerTaxId?: BuyerTaxId` — wire shape mirrors the domain's three states exactly: key absent = not asserted, `null` = asserted-none, string = present.
- `OrdersController.toDto` populates it from `order.buyerTaxIdState` (the entity getter that runs `decodeBuyerTaxIdColumn` on the raw column) — **never** a bare `IS NOT NULL`, which would misread the asserted-none row as present.

## Frontend

- `OrderBuyerTaxIdValue` (new component) — the three-state switch, on presence-then-nullness rather than truthiness.
- `OrderRecord.buyerTaxId?: string | null` added to the FE type contract.
- Wired into the order-detail Summary `KeyValueList`, appended and rendered **unconditionally** (per the mockup — the row exists precisely because two of its three states are easy to confuse and silently omitting it would make that worse).

## Acceptance criteria

- [x] Three states carry three distinct `data-testid` values, not one hook with varying text
- [x] Read through `decodeBuyerTaxIdColumn`; `IS NOT NULL` is never used
- [x] The PII-mode statement is visible to an operator (a real Radix tooltip, not a log line or a native `title` attribute)

## Deviation worth flagging

The mockup's own dev note (and a live grep of `libs/integrations/{allegro,erli}`) says the architecture doc's "PrestaShop is the only source that supplies a buyer tax id today" claim is stale — both Allegro and Erli already call `readSourceBuyerTaxId`. That correction is epic task E1-9 and is out of scope here; not touched.

## Test plan
- [x] `pnpm type-check` clean (apps/api clean; apps/web clean modulo the pre-existing, unrelated `erasableSyntaxOnly` TS-version noise)
- [x] New unit tests: `OrderBuyerTaxIdValue` component (4 tests, including the tooltip), `OrderDetailPage` (3 new tests for the three states), `OrdersController` (3 new tests for the DTO mapping)
- [x] Full pre-commit pipeline (lint, type-check, smart-test) passed locally: 74/74 backend Jest tests, 866/866 frontend Vitest tests

Closes #3180
Refs #3173


piotrswierzy (commented):
## Tech Lead review

Surfaces `order_records.buyerTaxId` on the order-detail Summary as three distinct renderings. The hard part here is the one that's easiest to get wrong, and this PR gets it right: the DTO reads `order.buyerTaxIdState` (the `decodeBuyerTaxIdColumn` getter) rather than the raw column, the FE switches on presence-then-nullness rather than truthiness, and the three `data-testid` hooks match `docs/plans/mockups/sales-document-tax-number-on-receipt.html` S1 byte-for-byte — including the copy (`None — asserted`, `Not asserted by the source`). The backend spec's `orderRecordWith({ buyerTaxId: '' })` case is the standout: it constructs the **raw column sentinel** as a repository read would, so the decode step is genuinely exercised instead of the test handing the code its own answer. My one substantive concern is placement: the field went on the shared `toDto`, which runs per row on the paged `/orders` list, so a buyer tax id now ships on every row of the hottest order read with no consumer for it.

### Findings

**[IMPORTANT]** — `apps/api/src/orders/http/orders.controller.ts:951`
> `buyerTaxId` is added to the shared `toDto`, which has four callers — including `listOrders` at line 361. `docs/architecture-overview.md § 5 Customers` and `§ 4 Orders` (#2599) both record that this value is PII-gated exactly like `customerEmail` ("for a sole trader the tax id identifies a natural person", `sanitizeAddress` already drops it from the snapshot). Today `OrderRecordResponseDto` carries no buyer-identifying value at all — `customerId` is an internal id — so this is the first one to reach the list payload, for every row, with **no** FE consumer (the mockup's tax-ID column is on the *invoices* list, S3, not on `/orders`). The repo's own precedent for this exact question is explicit, at `§ 22 Returns` on the holds projection: *"the rich projection is DETAIL-ONLY … never by the shared `toDto` that runs per row on the paged list"* — there for N+1 cost, here for data minimisation, and the MCP tool projections (`§ Capability Abstractions`) make the same call for buyer fields. It is not a privilege widening (`GET /orders` and `GET /orders/:id` carry the identical `@Roles('admin','operator','viewer')`), which is why this is IMPORTANT and not BLOCKING — but "so the list and the detail response read it from one place" buys nothing while the list has no reader. Move the assignment to `getOrder` (line 514), or say in the docblock which list surface is about to consume it.

**[SUGGESTION]** — `apps/api/src/orders/http/dto/order-record-response.dto.ts:98`
> `@ApiPropertyOptional({ nullable: true, description: … })` with no `type`. `BuyerTaxId` is `string | null | undefined`, and `design:type` reflection over a union emits `Object`, so the published OpenAPI schema for this property will not say `string` — a generated client would type it `any`/`object`. Add `type: String`. The `example` is worth having too: the three states are precisely what a schema reader will get wrong.

**[SUGGESTION]** — `apps/web/src/features/orders/components/order-buyer-tax-id-value.tsx:41-43`
> `UNKNOWN_EXPLANATION` states the mechanism ("no tax id is persisted at all") but drops the consequence the mockup's own dev note calls out at line 771: *"with `OL_STORE_PII=false` nothing is stored, the value reads back as 'not asserted', and **a simplified invoice can never issue**. Today nothing says so. One of the tasks below is simply to tell the operator."* That second clause is the operator-actionable half of E1-8 — an operator reading the current copy learns the field is empty, not that a document type has become unreachable. One more sentence.

**[SUGGESTION]** — `apps/api/src/orders/http/dto/order-record-response.dto.ts:17`
> `import { BuyerTaxId } from '@openlinker/core/orders';` is a third separate statement from the same specifier (lines 14-16 already import from it twice). Fold it into the existing type import.

**[SUGGESTION]** — PR base branch
> Base is `3178-rules-destination-warning`, not `main`, while #3180's own header says "✅ DIRECT TO `main` — One branch off `main`, one PR, one merge" and names no dependency. GitHub only honours `Closes #N` on a merge into the default branch, so #3180 will not auto-close (and CLAUDE.md forbids closing it by hand). It does explain why the cited mockup resolves — the file exists on the head chain but not on `main` — which is worth a line in the description if the stack is deliberate.

Worth noting as done right, beyond the decode discipline already mentioned: the tooltip copy deliberately never asserts *which* of the two indistinguishable cases this particular order is, because a single row cannot observe a deployment-wide setting — the docblock cites the `NAMELESS_LABEL` precedent for it, and that restraint is the difference between an explanation and a false claim. The row is also rendered unconditionally rather than conditionally like its neighbours in that `KeyValueList`, with the reason stated in place: omitting it would make the two confusable states *more* confusable, not less. And the tooltip is a real Radix trigger with `tabIndex={0}` rather than a native `title`, which is what makes AC-3 ("visible to an operator") actually hold for keyboard users — `docs/frontend-architecture.md § Paginated Totals` makes the same point about `title` not being reliably announced.

### Verdict
🔄 **Approve with changes**

1. Scope `buyerTaxId` to the detail response unless a list consumer is imminent — it is a PII value on a per-row payload with no reader today.
2. Add `type: String` to the `@ApiPropertyOptional` so the published schema is not `Object`.
3. Add the mockup's "a simplified invoice can never issue" clause to the PII-mode tooltip.

---

norbert-kulus-blockydevs:
## Review addressed — commit `83eb5aac4`

- **[IMPORTANT] `buyerTaxId` on the shared `toDto` reaches every list row** — *fixed*, narrowed to the detail read. Grepped `apps/web` first: the only consumer is `order-detail-page.tsx:332`, reading `order.buyerTaxId` off `orders.getById`; no list consumer anywhere, `apps/e2e` included. Moved the assignment out of `toDto` into `getOrder`, following the `activeHold` precedent, with the reason stated in place as **data minimisation** — not query cost, since the column is already loaded. Added a controller spec asserting the key is absent from a list row even when the column carries a value (`not.toHaveProperty`, so it cannot pass on a key present-but-undefined).

  One consequence is recorded in both the DTO description and the FE `OrderRecord` docblock: absence on a list row is now a property of the *endpoint*, never an assertion about the order, so a list row must never be fed to `OrderBuyerTaxIdValue` — which would render "Not asserted by the source" about an order nobody asked about.
- **[SUGGESTION] `@ApiPropertyOptional` with no `type`** — *fixed*. Added `type: String` (with a comment naming the `design:type`-over-a-union → `Object` reason) and `example: '5213796333'`.
- **[SUGGESTION] `UNKNOWN_EXPLANATION` drops the operator-actionable consequence** — *fixed*. The tooltip now ends "…and there a simplified invoice can never issue, because the tax id it is issued on is never recorded." Kept as a **conditional** clause about the deployment, so the copy still never asserts which of the two indistinguishable cases this order is; the docblock says why. The component test asserts the new clause alongside the existing `OL_STORE_PII=false` match.
- **[SUGGESTION] Third import statement from the same specifier** — *fixed*, folded into the existing type import.
- **[SUGGESTION] Base branch is not `main`** — *deliberate*. The 13-PR stack `#3199 → … → #3219` is intentional; base unchanged, and it is what makes the cited mockup resolve on the head chain.

---

piotrswierzy (approved):
## Tech Lead re-review (`db85d754` → `3f09d8ea`)

One authored commit against my review — `83eb5aac4` — plus the stack's merges. **The IMPORTANT finding is closed, and closed for the right stated reason.** `buyerTaxId` is off the shared `toDto` (four callers, `listOrders` among them) and set by `getOrder` alone. The commit is explicit that the scoping is **data minimisation, not query cost** — the column is already loaded, so projecting it would be free — which correctly distinguishes it from the `activeHold` precedent it otherwise follows, and is the honest framing: this is the first buyer-identifying value on the DTO, PII-gated at persistence exactly like `customerEmail` (#2599), on the hottest order read in the product, with no list consumer. The `apps/web` grep was done first and `order-detail-page.tsx` off `orders.getById` is confirmed as the only reader.

The new controller spec is the part that makes this stick: it asserts `expect(result.items[0]).not.toHaveProperty('buyerTaxId')` **with the column populated**, and its comment records why `toBeUndefined` would have been the weaker assertion (it passes on a DTO carrying the key with an undefined value). Both the DTO description and the FE `OrderRecord` docblock now state that absence on a list row is a property of the *endpoint* and never an assertion about the order, with an explicit instruction never to feed a list row to `OrderBuyerTaxIdValue` — which is the one way this scoping could have introduced a new false claim, and it is written down at both ends.

The three-state rule is intact throughout: `order.buyerTaxIdState` (the entity getter running `decodeBuyerTaxIdColumn`) is still the only read, no `IS NOT NULL` appears, the value is carried verbatim with no validation or normalisation, and the FE switch is presence-then-nullness rather than truthiness. The two SUGGESTIONs were also taken — `@ApiPropertyOptional` gains `type: String` (`BuyerTaxId` is a union, so `design:type` reflection emitted `Object` and a generated client would have typed it `any`), and the PII-mode tooltip gained the operator-actionable consequence ("a simplified invoice can never issue") stated as a conditional so it still asserts nothing about *this* order.

**No overlap with PR #3227.** I checked: that PR adds a frozen `buyerTaxIdValue` to `invoice_records` and renders it on the **invoices** list, touching no order DTO and no `/orders` read. It imports `OrderBuyerTaxIdValue` from the `features/orders` barrel, which is the intended cross-feature shape — and it independently confirms the decision made here, since the tax-id column belongs on the invoice list rather than the order list.

### Findings

**[SUGGESTION]** — `apps/api/src/orders/http/dto/order-record-response.dto.ts:100`
> `OrderRecordResponseDto` is shared by the list and detail responses, so the published OpenAPI still advertises `buyerTaxId` on `GET /orders` even though that endpoint never sets it. The description now says "DETAIL READ ONLY" in prose, which is the practical mitigation, but a generated client will still surface the property on a list row type. Only worth a separate detail-response DTO if a consumer is ever generated from the schema — not now.

> Targets `3178-rules-destination-warning`, not `main`, while #3180's header still reads "✅ DIRECT TO `main` — one branch, one PR, one merge" and names no dependency. `Closes #3180` will not fire.

### Verdict

✅ **Approve**

⚠️ **Merge state is `CONFLICTING` / `DIRTY`** against its own base `3178-rules-destination-warning` — this needs resolving before it can go anywhere, and given the stack has been merged in both directions repeatedly (`d0d98a5b6`, `6ad1e98d6`, `3f09d8ea5` all merge the base back in) I'd re-fetch and resolve by intent rather than textually: a sibling wave touching the same files makes a clean textual resolution silently wrong more often than not on this stack. CI: **no checks reported** on the branch — the workflow triggers on `pull_request: branches: [main]` and this targets a stack branch, so nothing has been validated in CI.

---

norbert-kulus-blockydevs:
## Suggestions addressed

- **[SUGGESTION] The shared DTO still advertises `buyerTaxId` on `GET /orders`** — *no change, agreed as scoped*. The prose "DETAIL READ ONLY" is the practical mitigation and a separate detail-response DTO only earns itself once a client is generated from the schema. Recording it so the next person to generate a client meets the reasoning rather than the surprise.

No code change this round; the branch carries the stack's propagated merges.

---

piotrswierzy (approved):
## Tech Lead re-review (`6a7ed38a` → `b0e2f30f`)

**My approval stands, and the conflict is resolved.** The PR now reports `mergeable: MERGEABLE` / `mergeStateStatus: UNSTABLE` (pending checks only) where it was `CONFLICTING`/`DIRTY` last round. The delta is a single merge commit carrying base-branch work only — the `mark-pre-rollout-era` test-fixture slice (`order-test-fixture.service.ts`, its interface, exception, controller route, DTO and int-spec) plus two docs and the `route-authorization-coverage.spec.ts` update. **This PR contributed no new commits of its own**, so the only thing worth re-checking is whether the resolution preserved the scoping decision. It did, and I verified it in the tree at `b0e2f30f` rather than from the diff.

`buyerTaxId` is still assigned in `getOrder` alone (`orders.controller.ts:540`), off the shared `toDto`, which carries an explicit comment at the exact position where a future author would be tempted to add it back (`:1046`). The framing survived the merge verbatim and is still the correct one — **data minimisation, not query cost** — with the reasoning intact in place: it is the first buyer-identifying value on this DTO, `customerId` being an internal id; it is PII-gated at persistence exactly like `customerEmail` (#2599, sole trader); `GET /orders` is the hottest order read in the product; and a value with no list consumer does not belong on every row of it. It correctly notes this costs no extra query, unlike the `activeHold` precedent it borrows the shape from.

The three-state rule is intact end to end. The DTO reads `order.buyerTaxIdState`, not the raw column, with the entity getter named as the only intended read and `decodeBuyerTaxIdColumn` as the only decoder — so the DTO never has to remember that `''` means asserted-none. The spec still asserts all three: `not.toHaveProperty('buyerTaxId')` on a **list** row whose record genuinely carries `'5213796333'` (and it explains why `toBeUndefined` would have been too weak — it passes on a DTO carrying the key with an undefined value); the id projected **verbatim** on the detail read; and asserted-none constructed from the **raw** `''` column rather than the already-decoded value, so the decode step is actually exercised rather than assumed. The value is carried through untouched — no validation, no normalisation — per ADR-026 keeping national specifics adapter-side.

### Findings

### Verdict
✅ **Approve** — unchanged from `6a7ed38a`; the conflict was resolved by intent and the scoping decision came through unharmed.

---

piotrswierzy (approved):
## Tech Lead re-verification (head unchanged at `9a8a7853`)

Position **7 of 18**. **Nearest blocking ancestor: #3201** (position 4, `REVIEW_REQUIRED`). `MERGEABLE`/`BLOCKED` from that ancestor; the earlier `CONFLICTING` state is resolved and has stayed resolved. CI in flight at run `34845837068`: `Docker Build Smoke Test`, `Detect changed paths`, `PHP Unit Tests` pass; five jobs **pending**; **no failures**.

Data-minimisation shape re-verified and intact: `buyerTaxId` is assigned in **`getOrder` alone** (`dto.buyerTaxId = order.buyerTaxIdState`), the shared `toDto` carries the explicit `// buyerTaxId is deliberately NOT here (#3180)` comment, and the list path is pinned negatively with `expect(result.items[0]).not.toHaveProperty('buyerTaxId')` rather than an assertion on a value. All three states survive end to end and are each tested: the id verbatim, `''` → `null` (asserted none), absent → `undefined` (not asserted), with `OrderBuyerTaxIdValue` rendering three distinct outputs rather than collapsing the last two.

### Findings

### Verdict
✅ **Approve**

---

Issue #2520, opened 2026-08-26
[TASK] CORE — FiscalLocateResult: a third outcome for held-but-not-registered


norbert-kulus-blockydevs:
## Problem / Context

`FiscalLocateResult` can report a confirmed registration or no match. A provider that has accepted the sale but not yet registered it therefore reports no match — which the UI can only render as "still not found", during what is in fact normal processing. This is the single most-reported confusion in the current flow.

## Proposed Solution

Add the third outcome per #2502, and map it in the eparagony adapter where the vendor's document status is non-terminal. A non-confirmed lookup must never be reported as an absence when the provider is known to hold the document. Keep the existing rule that the record stays in doubt rather than being terminalised on a registration that has not happened.

## Classification

**Type**: CORE
**Layer**: Domain
**File(s)**: `libs/core/src/fiscalization/domain/`, `libs/integrations/eparagony/`

## Dependencies

- Parent mini-epic: #2519
- #2502

## Acceptance Criteria

- [ ] The union carries a distinct outcome for held-but-not-registered.
- [ ] The eparagony adapter maps its non-terminal document statuses to it.
- [ ] A genuine absence remains distinguishable from the new outcome.
- [ ] No path terminalises a record on the new outcome.
- [ ] Tests added or updated for non-trivial logic
- [ ] No architecture boundary violations (CORE ↔ Integration)

## Branching and merge target

Branch: `{issue#}-{short-kebab}`, created **from its mini-epic branch**.
PR target: **its mini-epic branch**.

## How this issue gets finished

1. Implement on the branch above.
2. Open the PR into its mini-epic branch.
3. Run `/pr-review <PR>` and post the verdict as a PR comment.
4. Fix **every** finding, and reply under the comment saying what changed for each one.
5. Run `/pr-review` again. Repeat until the review comes back approved with no findings.
6. Only then merge.

Quality gate before every commit: `pnpm lint`, `pnpm type-check`, `pnpm test`.
Commits are signed off (`git commit -s`).

norbert-kulus-blockydevs:
Merged into the mini-epic branch. Closing manually: the PR targeted a non-default branch, so `Closes #N` does not fire until the epic reaches `main`.

PR #3212, opened 2026-09-12
fix(connections): count document registrations in connection health

### LINKED ISSUE #3179: fix(connections): count document registrations in connection health
## Merge strategy — 🔗 STACKED, on top of #3175

**Not direct to `main`.** This slice ships inside the #3173 stack (#3174 → #3175 → #3179 → #3176 → #3177 → #3178 → #3180 → #3181 → #3182 → #3183 → #3184 → #3185 → #3186): PR #3212 targets #3175's branch, not the default branch. One consequence is worth stating plainly — GitHub honours `Closes #3179` only on a merge into `main`, so this issue does **not** auto-close when its own PR merges; it closes when the chain reaches `main`, or by hand once it has.

---

Part of #3173.

## Context
The e-paragony connection registered receipt 210 — visible on the order with a full fiscal document identifier. Its health tab reports **"Last succeeded: Never · Last failed: Never"** with no job records at all. This is the one screen an operator opens to ask whether an integration is alive, and it answers "never" for a working one — indistinguishable from a connection that does not work.

## Scope
Make document registrations reach the connection's activity read.

## Acceptance
- [ ] A connection that has registered a document reports a real last-succeeded timestamp
- [ ] A connection that has genuinely never done anything still reports "Never"

**Depends on**: #3175 (the slice below it in the stack).

norbert-kulus-blockydevs:
## Summary

`ConnectionController.getDiagnostics` computed `lastSucceededAt` / `lastFailedAt` only from the last 10 `sync_jobs` rows, so a Fiscalization-only connection with genuine document-registration activity always read "Never" — the job that dispatched a fiscal receipt registration ages out of that window long before the registration itself stops mattering.

This PR folds `FiscalRegistrationRecord` and `InvoiceRecord` history into the same merge, alongside sync jobs, reporting whichever source observed the most recent success/failure.

## The one thing that needed fixing

The three sources were read via `Promise.all`. That means a fiscalization- or invoicing-table outage would take down the **whole connection health panel** — exactly the panel an operator opens when something is broken — where before this change only a `sync_jobs` outage could do that. Widening the read to three sources without widening its failure tolerance is a strict regression in availability.

Fixed by switching to `Promise.allSettled` and adding a new `unreadableSources` field (`syncJobs` / `fiscalRegistrations` / `invoices`) so three states are distinguishable behind a `null` timestamp:

1. a real timestamp (activity observed)
2. confirmed "nothing found" (`unreadableSources` empty, timestamp `null`)
3. "this source could not be read" (`unreadableSources` non-empty, timestamp `null` from the remaining readable sources only)

A rejected leg is logged and folded in as "no observations from that source" rather than thrown, so the response always returns 200 with whatever the readable sources answered. This follows the existing `analytics-trust` / `catalog-trust` / `sync-status` precedent that an unknown must never render as a healthy zero — the frontend can now tell a genuinely quiet connection apart from one it simply couldn't fully check.

## Test plan

- [x] `pnpm --filter @openlinker/api type-check`
- [x] `pnpm --filter @openlinker/core type-check`
- [x] Pre-commit hook (lint, type-check, smart-test) passed — 26 suites / 771 tests
- [x] New unit tests: DTO passthrough/default of `unreadableSources`; controller tests for one-source-rejects and all-sources-reject degradation

Closes #3179


piotrswierzy (commented):
## Tech Lead review

Folds `FiscalRegistrationRecord` and `InvoiceRecord` history into `GET /connections/:id/diagnostics` so a Fiscalization-only connection stops reporting "Last succeeded: Never" while it is demonstrably registering receipts. The core-side work is clean and correctly layered — port → service interface → service → repository, each with its own spec, `findRecentByConnectionId` ordered `createdAt DESC, id DESC` with a tiebreak, `ConnectionDiagnosticsSourceValues` as an `as const` union per engineering-standards. The `Promise.allSettled` + `unreadableSources` reasoning is the right instinct and well argued in the docblock. What holds it back is that the honesty this PR is proudest of stops at the wire: nothing in `apps/web` reads `unreadableSources`, so the panel still prints a flat "Never" for a source it failed to read — the exact confusion the field was added to remove. Two structural items follow: the invoice leg reaches through a *paginated* read and pays a discarded `COUNT` on every page load, and the whole fan-out lives in the controller.

### Findings

**[IMPORTANT]** — `apps/api/src/integrations/http/connection.controller.ts:265` → `libs/core/src/invoicing/infrastructure/persistence/repositories/invoice-record.repository.ts:351`
> The fiscal leg gets a purpose-built `findRecentByConnectionId` (`find` + `take`), but the invoice leg goes through `listInvoices({connectionId}, {limit:10, offset:0})` → `findMany`, which ends in `getManyAndCount()`. Per `docs/engineering-standards.md § When A Paginated Total Is Expensive` (and #2944, which just split five repositories for exactly this), `getManyAndCount` in typeorm 0.3.17 **always** runs the count query — so every connection-health page load pays a full `COUNT` over `invoice_records` filtered by `connectionId`, and the controller then discards `total` entirely. The asymmetry between the two legs is the tell. Mirror the fiscal shape: add `findRecentByConnectionId(connectionId, limit)` to `InvoiceRecordRepositoryPort` / `IInvoiceService` and call that.

**[IMPORTANT]** — `apps/api/src/integrations/http/dto/connection-diagnostics-response.dto.ts:30-45` vs `apps/web/src/features/connections/components/ConnectionDiagnosticsPanel.tsx:29,79`
> The docblock states the field exists "so the FE can distinguish three states behind a `null` timestamp". It cannot: no file under `apps/web` mentions `unreadableSources` (the FE `ConnectionDiagnostics` type in `features/connections/api/connections.types.ts` is untouched), and `formatDate` still renders `value === null → 'Never'` unconditionally. So on a fiscalization-table outage the panel shows a connection that *is* working as "Never" — the same false claim #3179 exists to remove, from a different cause, and now with a server-side field that reports it correctly and a UI that ignores it. `docs/architecture-overview.md` is explicit for the trust surfaces this cites as precedent that an unknown must be *rendered* differently, not merely modelled (see also `docs/frontend-architecture.md § Paginated Totals As A Second Stage`: "`idle` and `unavailable` are different, and a surface must render them differently"). Either render it in this PR or drop the claim from the docblock and open the follow-up — a field nothing reads is a guarantee nobody has.

**[IMPORTANT]** — `apps/api/src/integrations/http/connection.controller.ts:252-320`
> `getDiagnostics` now owns a three-source fan-out, a per-leg degradation policy and a private `unwrapDiagnosticsRead` helper. `docs/engineering-standards.md` keeps controllers thin (validate, delegate, format), and this file's own neighbours are the pattern: `RateLimitStatusService` / `WebhookStatusService` — and `AuthorityStatusService`, which `docs/architecture-overview.md § 20` documents as "composed by an app-layer service ... the `RateLimitStatusService` / `WebhookStatusService` shape". A `ConnectionDiagnosticsService` in `integrations/application/services/` would put the merge where the other composed reads live and let the degradation behaviour be unit-tested without standing up the whole controller module (the new controller spec has to register two extra DI tokens purely to reach it).

**[IMPORTANT]** — `apps/api/src/integrations/http/dto/connection-diagnostics-response.dto.ts` (`recentErrors` assembly)
> `recentErrors` is now `[...jobErrors, ...fiscalErrors, ...invoiceErrors]` — source-ordered, not time-ordered — and the new spec pins that order (`['transport timeout', 'authority rejected receipt', 'buyer VAT id rejected']`). Before this change it was newest-first, because it derived from a newest-first job list. The panel renders the array verbatim under "Recent errors", so a month-old job error can now sit above today's invoice failure with nothing to indicate it. You already carry `Observation.at` for both `lastFailedAt` and the messages — sort `failed` by `at` DESC before mapping, and assert the recency ordering rather than the concatenation order.

**[SUGGESTION]** — `libs/core/src/fiscalization/infrastructure/persistence/repositories/fiscal-registration-record.repository.ts:97`
> The top-10 window is selected on `createdAt DESC`, but "most recent" is then computed on `registeredAt ?? updatedAt` (and `issuedAt ?? updatedAt` for invoices). A record created outside the newest 10 by creation time but *registered* recently — e.g. a long-running `in-doubt` registration that #2520's reconcile later resolved, or the retry ladder's 6-hour backoff — never enters the merge, so `lastSucceededAt` can understate. It is a proxy and probably an acceptable one, but the mismatch between the selection clock and the comparison clock deserves a line in the port docblock, which currently only says "newest-first".

**[SUGGESTION]** — PR base branch
> Base is `3175-sales-document-adr`, not `main`, while #3179's own header says "✅ DIRECT TO `main` — One branch off `main`, one PR, one merge" and names no dependency. GitHub only honours `Closes #N` on a merge into the default branch, so #3179 will not auto-close (and CLAUDE.md forbids closing it by hand). `gh pr checks` also reports no checks on this branch.

Worth noting as done right: `ConnectionDiagnosticsSourceValues` as an exported `as const` union rather than ad-hoc strings, so `unreadableSources` is FE-narrowable; the `allSettled`-not-`all` argument stated in-place at the call site rather than only in the PR body; and the spec that asserts an in-flight `registering` record counts as neither a success nor a failure — the case a two-branch `if/else` would have silently mis-bucketed.

### Verdict
🔄 **Approve with changes**

1. Read invoices through a bounded `findRecentByConnectionId`, not the paginated `listInvoices` with its discarded `COUNT`.
2. Render `unreadableSources` in `ConnectionDiagnosticsPanel`, or remove the docblock's claim that the FE distinguishes the three states.
3. Sort `recentErrors` by observation time.
4. Move the fan-out into an app-layer `ConnectionDiagnosticsService`.

---

norbert-kulus-blockydevs:
## Review addressed — commits `27bd2f602`, `d23a58c41`

- **[IMPORTANT] Invoice leg pays a discarded `COUNT`** — *fixed*. Added `findRecentByConnectionId(connectionId, limit)` to `InvoiceRecordRepositoryPort` / `IInvoiceService`, implemented in `invoice-record.repository.ts` with `find` + `take` + `createdAt DESC, id DESC` — the fiscal leg's exact shape — and pointed the diagnostics read at it. `getManyAndCount` is no longer on this path. Specs added at both the repository and the service level, the latter asserting `findMany` is never reached.
- **[IMPORTANT] `unreadableSources` modelled but never rendered** — *fixed*, rendered rather than de-claimed. `ConnectionDiagnosticsPanel` now distinguishes three states: a real timestamp, `Never` (every source read), and `Unknown` (a source could not be read), plus a warning naming the unreadable sources in **visible text**, not a `title`. A present timestamp stays visible but is flagged as possibly incomplete. The FE `ConnectionDiagnostics` type gains the field and a mirrored `ConnectionDiagnosticsSource` union; five panel tests cover the states, including the pre-#3179 absent-field case.
- **[IMPORTANT] Fan-out and degradation policy live in the controller** — *fixed*. Extracted to `apps/api/src/integrations/application/services/connection-diagnostics.service.ts` (+ interface, + `application/types/connection-diagnostics.types.ts`), following the `RateLimitStatusService` / `WebhookStatusService` / `AuthorityStatusService` shape, wired in `integrations.module.ts`. The controller is back to validate/delegate/format; the degradation tests moved into the new service spec and the controller spec dropped the three DI tokens it registered only to reach that code. The source vocabulary moved out of `http/dto` so the application layer does not import from the interface layer.
- **[IMPORTANT] `recentErrors` source-ordered, not time-ordered** — *fixed*. `fromDomain` sorts the merged `failed` observations by `Observation.at` DESC before deriving both `lastFailedAt` and `recentErrors`. The DTO spec now asserts recency ordering **across** sources, plus a second case pinning stable source order for two failures sharing one instant.
- **[SUGGESTION] Selection clock vs comparison clock on the top-10 window** — *fixed* (docblock only, query unchanged). Both port docblocks — `fiscal-registration-record-repository.port.ts` and the new invoice sibling — now say it: selected on `createdAt`, ranked on `registeredAt ?? updatedAt` / `issuedAt ?? updatedAt`, so `lastSucceededAt` is a lower bound and an accepted proxy.
- **[SUGGESTION] Base branch is not `main`** — *deliberate*. The 13-PR stack `#3199 → … → #3219` is intentional; base unchanged, and #3179 will be closed once the stack reaches `main`.

**Two changes you did not ask for, both forced:**

1. `scripts/check-cross-context-imports.mjs`'s `SyncJobRepositoryPort` ALLOW_LIST rows for `connection.controller(.spec).ts` went stale the moment the coupling moved into the new service, and #2791's stale-row gate fails the build on that. Re-filed against `connection-diagnostics.service(.spec).ts`.
2. Widening `IInvoiceService` with `listRecentByConnectionId` broke three `jest.Mocked<IInvoiceService>` object literals that predate it (`invoicing-issue.handler.spec.ts`, `invoicing.controller.spec.ts`, `orders.controller.spec.ts`) — caught by `pnpm type-check` at the stack tip, fixed in `d23a58c41`.

---

piotrswierzy (approved):
## Tech Lead re-review (`314eafcf` → `d23a58c4`)

Two authored commits against my review — `27bd2f602` and `d23a58c41` — plus the stack's merges. **All three prior findings are closed, and the response went further than each one asked.** The `unreadableSources` field now has a real frontend consumer: `ConnectionDiagnosticsPanel` renders `Unknown` rather than `Never` behind a null timestamp while any source is unreadable, with a `tone="warning"` `Alert` naming which sources could not be read, and it keeps a *present* timestamp visible but flagged as possibly incomplete — which is the correct asymmetry, since a real observation is still a real observation. The DTO docblock's claim is now true rather than aspirational. The invoice leg moved off the paginated `listInvoices` onto a purpose-built `findRecentByConnectionId` (`find` + `take`, `createdAt DESC, id DESC`, symmetric with the fiscal leg), so the discarded `getManyAndCount()` COUNT on every health page load is gone — and the port docblock cites §2944 / § When A Paginated Total Is Expensive rather than just asserting it. `recentErrors` is now sorted newest-first *across* sources with a note that `Array.prototype.sort` is stable so ties keep source order, and the DTO spec asserts both.

Two things I did not ask for and am glad to see. The three-source fan-out moved out of the controller into an app-layer `ConnectionDiagnosticsService`, following the `RateLimitStatusService` / `WebhookStatusService` / `AuthorityStatusService` shape — the controller is back to validate/delegate/format, `connection.controller.spec.ts` shed 245 lines of DI wiring, and the degradation policy is now unit-testable without a controller module. The source vocabulary moved from `http/dto` to `application/types/`, which fixes a layer inversion that would otherwise have had the application layer importing from the interface layer. And the stale `connection.controller(.spec).ts → SyncJobRepositoryPort` `ALLOW_LIST` rows were **re-filed** against the service that inherited the coupling rather than left to trip #2791's stale-row gate — exactly the discipline that gate exists to enforce.

The `connection` read is deliberately outside the `allSettled` fan-out so a missing connection is still a 404 rather than a degraded panel about nothing, and that reasoning is stated in place. Good.

### Findings

**[SUGGESTION]** — `libs/core/src/invoicing/application/services/invoice.service.interface.ts`
> `listRecentByConnectionId` is a **required** member added to a published service interface, which is why `d23a58c41` had to chase three `jest.Mocked<IInvoiceService>` literals. That is the right fix and the commit message says so. Worth noting only that `IInvoiceService` is an intra-tree core contract with no out-of-tree implementer, so the source-compatible-widening discipline the repo applies to *plugin-facing* ports (`listExternalIdsByConnection`'s optional page argument, #2219; `InventoryAdjustment`'s optional members, #2368) does not bind here — but if that ever changes, this member is the precedent someone will cite.

> Targets `3175-sales-document-adr`, not `main`, while #3179's header still reads "✅ DIRECT TO `main` — one branch, one PR, one merge" and names no dependency. `Closes #3179` will not fire on a merge into a non-default branch. Retarget once the base lands, or amend the issue header.

### Verdict

✅ **Approve**

Merge state: `MERGEABLE` / `BEHIND` (needs a base refresh, no conflicts). CI: this PR *does* have a check run in flight — `Build`, `Test`, `Lint`, `Integration Tests` and `Docker Build Smoke Test` all **pending** at time of review; `Detect changed paths`, `PHP Unit Tests` and `Scaffolded Adapter Builds` pass. Reporting that honestly rather than as green — the substantive suites have not reported yet.

---

norbert-kulus-blockydevs:
## Suggestions addressed

- **[SUGGESTION] A required member on a published service interface** — *no change; recorded*. Agreed on both halves: `IInvoiceService` is an intra-tree core contract with no out-of-tree implementer, so the source-compatible-widening discipline that binds plugin-facing ports (`listExternalIdsByConnection`'s optional page argument #2219, `InventoryAdjustment`'s optional members #2368) does not bind here — and if that ever changes, this member is the precedent. Worth noting the widening was caught by `pnpm type-check` at the stack tip, not by anything on this branch, which is the same CI gap you flag below.

No code change this round; the branch carries the stack's propagated merges.

---

piotrswierzy (approved):
## Tech Lead re-verification (head unchanged at `dd409b3`)

Head and `main` unchanged, so rather than repeat the review I spot-checked the four things the approval rested on — the ones that went beyond the ask and are therefore the ones most worth confirming survived.

All four hold. `ConnectionDiagnosticsService.getDiagnostics` reads the connection **before** the fan-out with the reason stated in place (a connection that does not exist is a 404, not a degraded panel about nothing), then `Promise.allSettled`s three legs whose rejections `unwrap` pushes into `unreadableSources` instead of throwing. The invoice leg is `IInvoiceService.listRecentByConnectionId` over a bounded `findRecentByConnectionId`, and that port's docblock does two things I did not expect and am glad of: it cites the `getManyAndCount` cost rule (`docs/engineering-standards.md` § When A Paginated Total Is Expensive) as the reason it is not `findMany`, and it names its own honest limit — the selection clock is `createdAt` while the caller ranks by `issuedAt ?? updatedAt`, so "last succeeded" is a lower bound, stated as an accepted proxy rather than an invariant. `recentErrors` is sorted newest-first on a copy across all three sources (`connection-diagnostics-response.dto.ts:177`) with the source-order trap written above it. The panel renders `Unknown` vs `Never` through one resolver (`ConnectionDiagnosticsPanel.tsx:71`) plus a warning `Alert` naming the affected sources (`:132-136`), and keeps a timestamp that *is* present visible while a sibling source is unreadable.

The `check-cross-context-imports.mjs` edit is the part worth being explicit about: the two `connection.controller{,.spec}.ts` rows are **removed**, not left behind, and re-filed against `connection-diagnostics.service{,.spec}.ts`. That is required rather than tidy — since #2791 a row naming a file that no longer imports the symbol fails the build, and a stale row is a standing licence for the coupling to return.

**Stack**: position 3 of the 17-deep chain rooted at #3199 (#3199 → #3200 → **#3212** → #3201 → …). Blocked behind #3200, which is blocked behind #3199 reaching `main`. Note that **#3201 — immediately above this PR — is `REVIEW_REQUIRED` with no review**, so it, not #3217, is the chain's first hard blocker; everything from position 4 up waits on it.

**CI: the run is queued, not green.** Lint, Test, Type Check, Build and Integration Tests are all `pending`; Docker Build Smoke, PHP Unit Tests, Scaffolded Adapter Builds and the path gate pass. `mergeStateStatus: BLOCKED` reflects the unmerged base. Neither is a review objection — flagging so the green isn't assumed.

### Findings

### Verdict
✅ **Approve**

---

piotrswierzy (approved):
## Tech Lead re-verification (head unchanged at `dd409b3a`)

Position **3 of 18**. **Nearest blocking ancestor: none** — #3199 and #3200 above it are both `CLEAN`/`APPROVED`, so this PR completes the chain's **merge-ready prefix**: #3199 → #3200 → #3212 can land back-to-back today. #3201 is the chain's first hard blocker but sits *below* this PR and does not gate it. `MERGEABLE`/`UNSTABLE`. CI at run `34831266177`: `Type Check`, `Docker Build Smoke Test`, `Scaffolded Adapter Builds`, `Detect changed paths`, `PHP Unit Tests` pass; `Lint` / `Test` / `Build` / `Integration Tests` **pending**; **no failures**.

All four beyond-the-ask items re-verified at this head and intact:

- **`unreadableSources`** — closed `ConnectionDiagnosticsSource` vocabulary, populated from `Promise.allSettled` unwrapping, with the rule stated on the type: read these lists as a confirmed absence **only** while it is empty. Its four cases are tested individually, including the all-three-rejected case.
- **The FE renders `Unknown`, never `Never`, for an unread source** — `return sourcesUnreadable ? 'Unknown' : 'Never'`, above a warning `Alert` ("Some activity could not be checked") that names the unreadable sources and says what Unknown means. The `analytics-trust` / `catalog-trust` / `sync-status` precedent, applied.
- **Bounded read** — `findRecentByConnectionId(connectionId, RECENT_ACTIVITY_LIMIT)`, asserted as `(CONNECTION_ID, 10)` and asserted **not** called on the short-circuit path.
- **App-layer `ConnectionDiagnosticsService`** — `apps/api/src/integrations/application/services/`, behind `IConnectionDiagnosticsService` + `CONNECTION_DIAGNOSTICS_SERVICE_TOKEN`, composing three `I*Service`/port reads and calling no destination platform.

`scripts/check-cross-context-imports.mjs` is still in the changed set, so the stale `ALLOW_LIST` rows remain removed and re-filed.

### Findings

### Verdict
✅ **Approve**

---

PR #3214, opened 2026-09-12
refactor(web): move KSeF status labels into the ksef plugin

### LINKED ISSUE #3181: refactor(web): move KSeF status labels into the ksef plugin
## Merge strategy — 🔗 STACKED, on top of #3180

**Not direct to `main`.** This slice ships inside the #3173 stack (#3174 → #3175 → #3179 → #3176 → #3177 → #3178 → #3180 → #3181 → #3182 → #3183 → #3184 → #3185 → #3186): PR #3214 targets #3180's branch, not the default branch. One consequence is worth stating plainly — GitHub honours `Closes #3181` only on a merge into `main`, so this issue does **not** auto-close when its own PR merges; it closes when the chain reaches `main`, or by hand once it has.

---

Part of #3173.

## Context
`apps/web/src/features/invoicing/components/regulatory-status-badge.tsx` hard-codes `KSeF: awaiting submission`, `KSeF: submitted`, `KSeF: clearing`, `KSeF: accepted` and `KSeF: rejected` in a **shared** feature folder, even though `apps/web/src/plugins/ksef/` exists. Nothing catches it: the file is outside the UI-vocabulary scan roots and outside every core guard.

## Scope
Relocate the labels to the KSeF plugin, or make them provider-supplied. The shared surface keeps a regulator-neutral vocabulary.

## Acceptance
- [ ] No regulator name in `features/invoicing/`
- [ ] `invoices-list-page.test.tsx` updated rather than deleted

**Depends on**: #3180 (the slice below it in the stack).

norbert-kulus-blockydevs:
## Summary

`RegulatoryStatusBadge` (`features/invoicing`) hard-coded `KSeF: ...` label
text even though it is rendered by every `InvoicingPort` provider's own
invoice-detail surface (KSeF, inFakt, Subiekt) plus the generic invoices
list, invoice detail page, and order sales-document panel. This violates the
country-agnostic architecture (ADR-026 / decision 7 in #3173): a shared
surface must not assume the regulator is KSeF — inFakt and Subiekt relay to
KSeF rather than being KSeF themselves.

- `REGULATORY_STATUS_LABEL_FALLBACK` now carries regulator-neutral text
  (`Submitted`, `Accepted`, `Rejected`, `Clearing`, `Awaiting submission`).
- `RegulatoryStatusBadge` gains an optional `labelOverrides` prop so a
  per-provider surface can supply its own branded wording.
- The KSeF-branded label map is relocated to the ksef plugin
  (`plugins/ksef/lib/ksef-regulatory-status-labels.ts`) and wired into
  `KsefInvoiceDetailSection` via `labelOverrides` — KSeF's own invoice-detail
  card renders byte-identical text to before.
- `invoices-list-page.tsx`'s page description dropped its stray `(KSeF)`
  parenthetical for the same reason (that list mixes rows from every
  provider).

### Rendered-string changes (sanctioned by the issue — "the shared surface
keeps a regulator-neutral vocabulary")
The generic invoices list, invoice detail page, order sales-document panel,
and the inFakt/Subiekt detail sections now render neutral wording
(`Accepted`) instead of the previously-incorrect `KSeF: accepted` for
non-KSeF providers. The KSeF plugin's own detail section is unchanged.

## Acceptance criteria
- [x] No regulator name in **rendered operator copy** under `features/invoicing/` — restated after review, because the original wording was literally false at head and stayed false by design: `use-ksef-upo-download.ts`, `use-resend-to-ksef-mutation.ts`, `resendToKsef` and `ksef-upo-preview.types.ts` are genuine KSeF-specific endpoints and hooks the plugin imports back through the `features/invoicing` barrel (the documented, allowed direction). What the AC is about — text an operator reads — is clean: every surviving `KSeF` under that folder is a comment, a test assertion, an endpoint path or a hook identifier.
- [x] `invoices-list-page.test.tsx` updated rather than deleted

## Test plan
- [x] `pnpm --filter @openlinker/web type-check` clean
- [x] Added `regulatory-status-badge.test.tsx` covering the neutral default
      and the `labelOverrides` behavior
- [x] Updated `invoices-list-page.test.tsx` (neutral label assertion) and
      `invoice-detail-page.test.tsx` (disambiguated from `InvoiceTimeline`'s
      own "Accepted" clearance-lane node, which the old `KSeF: accepted`
      text had accidentally been distinct from)
- [x] Full pre-commit gate (lint, check:invariants, type-check, smart-test)
      passed: 255 test files / 2919 tests passed

Closes #3181
Refs #3173


piotrswierzy (commented):
## Tech Lead review

Moves the KSeF-branded `RegulatoryStatus` labels out of the shared `features/invoicing` badge into `plugins/ksef/`, leaving the shared default regulator-neutral and giving the badge an optional `labelOverrides` prop for the one surface where "KSeF" is genuinely the right word. The shape is right — provider branding supplied by the provider, shared vocabulary neutral by construction, with the KSeF plugin's own card byte-identical to before — and the new `regulatory-status-badge.test.tsx` covers both the neutral default and the override, including the omitted-status fallback. Two things hold it back: the override is threaded through `t()`'s **fallback** slot, so it silently loses the day anyone populates the `invoice.regulatory.*` catalog keys; and the PR ticks the issue's "No regulator name in `features/invoicing/`" box while `invoice-timeline.tsx` still renders two operator-facing KSeF strings on the same shared detail page.

### Findings

**[IMPORTANT]** — `apps/web/src/features/invoicing/components/regulatory-status-badge.tsx:75`
> `const fallback = labelOverrides?.[status] ?? LABEL_FALLBACK[status];` then `t(\`invoice.regulatory.${status}\`, fallback)`. Per `apps/web/src/shared/i18n/use-translation.ts` and `docs/frontend-architecture.md § Internationalization`, `t(key, fallback)` returns **the catalog hit** and only falls back when the catalog misses. The catalog is empty today, so this works — but `invoice.regulatory.*` is by design the *neutral* key, so the first per-feature string-migration PR that adds an `en` entry for it will make the KSeF override unreachable, silently, with no test failing (every test here renders under the empty host catalog). A provider override must not be expressible only as a fallback to a key whose translated value is guaranteed to contradict it. Either resolve the override *after* `t()` (`labelOverrides?.[status] ?? t(key, LABEL_FALLBACK[status])`), or have the override carry its own key namespace. The current line is the one place a reader would never think to look when the branding disappears.

**[IMPORTANT]** — `apps/web/src/features/invoicing/components/invoice-timeline.tsx:205-206`
> The PR body ticks `- [x] No regulator name in features/invoicing/`, but this shared component still renders `'Awaiting KSeF submission'` and `'Issued offline — not yet sent to KSeF'` — operator-facing text, in the same shared feature folder, on the same invoice-detail page, for an inFakt or Subiekt invoice. That is the identical defect class #3181 describes (and note `invoice-detail-page.test.tsx` renders exactly this `pending-submission` lane). The other ~18 `features/invoicing` KSeF hits are comments, hook names and API paths and are fine. Either neutralise these two strings in the same PR, or un-tick the AC and say in the description that the timeline is deliberately deferred — as written the box claims a guarantee the tree does not hold.

**[IMPORTANT]** — PR base branch
> Base is `3180-show-buyer-tax-id-on-order` (itself based on `3178-rules-destination-warning`), not `main`, while #3181's own merge-strategy header says "✅ DIRECT TO `main` — One branch off `main`, one PR, one merge" and records no dependency. Two consequences: GitHub only honours `Closes #N` on a merge into the default branch, so #3181 will **not** auto-close (CLAUDE.md § Issue lifecycle forbids closing it by hand), and `gh pr checks` reports no checks on this branch. Retarget to `main` unless the stack is deliberate, in which case say so in the description.

**[SUGGESTION]** — `apps/web/src/pages/invoicing/invoice-detail-page.test.tsx:157-161`
> `findAllByText('Accepted')` + `length > 0` is weaker than it needs to be: `findAllByText` already throws on zero matches, and the assertion no longer distinguishes the **badge** from the `InvoiceTimeline` clearance node, so it would still pass if the badge stopped rendering entirely. The sibling `invoices-list-page.test.tsx` change in this same PR shows the right technique — filter to the element you mean (`.closest('.status-badge')`, as the new `regulatory-status-badge.test.tsx` already does).

**[SUGGESTION]** — `apps/web/src/plugins/ksef/lib/ksef-regulatory-status-labels.ts:19`
> `'not-applicable': 'N/A'` is unreachable — the badge's own header records that the panel's `regulatoryStatus !== 'not-applicable'` gate means that value never arrives. Typing the map as `Partial<Record<…>>` (matching the prop) and dropping the entry would make the override map contain only the five statuses it actually brands, and keeps `Record` exhaustiveness from implying this surface has an opinion about a status it never sees.

Done correctly and worth noting: the override map is `Partial<Record<…>>` at the prop, and the new spec explicitly covers the per-status fallback when the map omits a key — that is the case a `Record`-typed prop would have hidden. Placing the map under `plugins/ksef/lib/` and importing `RegulatoryStatus` from the `features/invoicing` **barrel** is also the correct plugin→feature crossing per `docs/frontend-architecture.md § Feature Public Surface`.

### Verdict
🔄 **Approve with changes**

1. Resolve `labelOverrides` outside `t()` so a populated catalog cannot silently erase provider branding.
2. Neutralise `invoice-timeline.tsx`'s two KSeF strings, or un-tick the AC and state the deferral.
3. Retarget the PR to `main` (or document the stack) so `Closes #3181` actually fires and CI runs.

---

norbert-kulus-blockydevs:
## Review addressed — commit `9a8568bb9`

- **[IMPORTANT] `labelOverrides` threaded through `t()`'s fallback slot** — *fixed*. The override now resolves **after** `t()` (`labelOverrides?.[status] ?? t(key, LABEL_FALLBACK[status])`), so a populated neutral `invoice.regulatory.*` catalog entry can no longer erase provider branding. Two new specs render under a **populated** catalog (nested `LocaleProvider catalog={...}`, the idiom `locale-provider.test.tsx` already uses) and pin both directions: the override wins when given, the catalog entry wins over the neutral fallback when it is not. That is the test that would have failed silently before.
- **[IMPORTANT] Two KSeF strings still in `invoice-timeline.tsx`** — *fixed*, neutralised rather than deferred: `'Awaiting KSeF submission'` → `'Awaiting submission'`, `'Issued offline — not yet sent to KSeF'` → `'… not yet sent to the authority'`. The component exposes no per-provider override seam, so neutral wording was the only correct fix. One correction to the finding: the spec that actually renders the `pending-submission` lane is `invoice-timeline.test.tsx:97`, not `invoice-detail-page.test.tsx`, which never sets that status — the right one was updated and now also asserts the lane renders no regulator name at all. The PR body's AC now holds; the remaining `features/invoicing` KSeF hits are comments, hook names and API paths only.
- **[IMPORTANT] Base branch is not `main`** — *deliberate*. The 13-PR stack `#3199 → … → #3219` is intentional; base unchanged, and `Closes #3181` fires when the stack reaches `main`.
- **[SUGGESTION] `findAllByText('Accepted')` + `length > 0` is weak** — *fixed*. Now filters to `.closest('.status-badge')`, so it fails if the badge stops rendering and only the timeline node remains. Deliberately not an exact count: `invoice-detail-page.tsx` renders the badge twice (head badges line 254 + detail-list item line 113), so `toHaveLength(1)` would be a false failure.
- **[SUGGESTION] Unreachable `'not-applicable': 'N/A'`** — *fixed*. The map is now `Partial<Record<RegulatoryStatus, string>>` (matching the prop) with the entry dropped, leaving the five statuses the surface actually brands; a docblock records why exhaustiveness is wrong here.

**One thing to know about the merged result.** #3217, downstream in this stack, had already neutralised the same two timeline strings in its own pre-review work — so propagating this fix up the chain conflicted with it. Resolved in favour of the downstream version, which is what a normal merge order would produce: the subLabel reads `'Issued offline — not yet sent for clearance'` there rather than this PR's `'… to the authority'`, and the richer #3217 docblock survives. The test kept **this** PR's stronger assertion (`container.textContent` over the whole lane, rather than a whole-text-node `queryByText`), and the weaker duplicate was dropped.

---

piotrswierzy (approved):
## Tech Lead re-review (`89b9c564` → `d5c8c43c`)

The delta carries one authored commit against my review — `9a8568bb9` — plus the stack's merges from `main` and `3180-…`. **Both prior findings are closed, and the IMPORTANT one is closed in the right place.** The override no longer travels through `t()`'s fallback slot; it is resolved *after* `t()` returns, so a future string-migration PR populating the deliberately-neutral `invoice.regulatory.*` key can no longer silently erase provider branding. Crucially the new specs render under a **populated** `LocaleProvider` catalog in both directions (override wins over a catalog hit; catalog hit wins over the neutral fallback) — which is the only way this regression is observable at all, since the host catalog is empty. The secondary finding is also gone: `invoice-timeline.tsx` now says `Awaiting submission` / `not yet sent to the authority`, and its spec asserts `container.textContent` matches no `/KSeF/i` at all, which is a stronger guard than the string swap it protects. Two unasked-for improvements are welcome: the `invoice-detail-page` assertion now filters to `.closest('.status-badge')` (it would previously have passed with the badge gone entirely), and `KSEF_REGULATORY_STATUS_LABELS` became `Partial<Record<…>>` with the unreachable `'not-applicable'` entry dropped.

### Findings

**[SUGGESTION]** — `apps/web/src/features/invoicing/` (AC wording)
> The ticked AC still reads "No regulator name in `features/invoicing/`", and that is still literally false at head — `use-ksef-upo-download.ts`, `use-resend-to-ksef-mutation.ts`, `resendToKsef` on the api module, and `ksef-upo-preview.types.ts` all live there. That is fine and out of scope: they are genuine KSeF-specific endpoints and hooks the plugin imports back through the `features/invoicing` barrel (the documented, allowed direction). What the AC was actually about — **rendered operator copy** — is now clean; I grepped every file under the folder at `d5c8c43c` and every surviving `KSeF` is a comment, a test assertion, an endpoint path, or a hook identifier. Worth restating the AC as "no regulator name in rendered copy under `features/invoicing/`" so the next reader does not chase the rest.

> This PR targets `3180-show-buyer-tax-id-on-order`, not `main`, while issue #3181's header still says "✅ DIRECT TO `main` — one branch, one PR, one merge" and names no dependency. `Closes #3181` fires only on a merge into the default branch, so as configured the issue will not auto-close. Either retarget the stack to `main` once the lower PRs land, or amend #3181's merge-strategy header to name the dependency — the current state is an issue whose stated strategy and actual branch disagree.

### Verdict

✅ **Approve**

CI: `gh pr checks` reports **no checks on this branch** — the workflow triggers on `pull_request: branches: [main]` and this PR targets a stack branch, so nothing has validated it in CI. `mergeable`/`mergeStateStatus` both read `UNKNOWN`. Neither is a reason to hold the change, but neither is evidence of green either; the gate will only really run once the stack retargets `main`.

---

norbert-kulus-blockydevs:
## Suggestions addressed — PR body updated

- **[SUGGESTION] The ticked AC is still literally false at head** — *fixed*, restated rather than un-ticked. The AC now reads "No regulator name in **rendered operator copy** under `features/invoicing/`", and the body names the four files that legitimately keep the regulator in their identifiers (`use-ksef-upo-download.ts`, `use-resend-to-ksef-mutation.ts`, `resendToKsef`, `ksef-upo-preview.types.ts`) plus why that is the allowed direction — they are genuine KSeF-specific endpoints and hooks the plugin imports back through the `features/invoicing` barrel. Thank you for grepping the folder rather than taking the tick at face value; the claim was broader than the work.

One thing to know about the merged result: #3217, downstream, had independently neutralised the same two `invoice-timeline` strings in its own pre-review work, so propagating this PR's fix conflicted with it. Resolved in favour of the downstream wording (`… not yet sent for clearance`), which is what a normal merge order produces; this PR's stronger test assertion (`container.textContent` over the whole lane) survived and the weaker duplicate was dropped.

---

piotrswierzy (approved):
## Tech Lead re-verification (head unchanged at `dd651e6`)

Head and `main` unchanged. I re-checked the ordering property the approval rested on — that the override resolves *after* `t()` rather than in its fallback slot — because that is the one thing here whose regression is invisible on a repo whose host catalog is empty.

It holds, and it is pinned in the only way that can catch it. `regulatory-status-badge.tsx:228` reads `labelOverrides?.[status] ?? t(\`invoice.regulatory.${status}\`, LABEL_FALLBACK[status])` — resolved outside `t()`, with the reason written directly above it (`t(key, fallback)` returns the catalog hit, so an override handed in as the fallback becomes unreachable the moment a catalog entry exists). Both directions are asserted under a **populated** `LocaleProvider` catalog: an override outranks a populated `invoice.regulatory.*` entry for the same status, and the catalog entry wins over the neutral fallback when no override is given. Against the host's empty catalog both spellings pass, so the populated fixture is what makes the test real rather than decorative.

`invoice-timeline.tsx` is neutralised and the spec asserts it stays that way — `expect(container.textContent).not.toMatch(/KSeF/i)` on the pending-submission lane, with the reason stated (the lane serves every provider and exposes no per-provider override seam). The badge's default-path specs additionally assert `queryByText(/KSeF/i)` is null, so regulator vocabulary cannot creep back into the shared component from either side.

**Stack**: position 8 of the 17-deep chain rooted at #3199 (… → #3211 → **#3214** → #3216 → #3217 → …). Blocked transitively behind seven predecessors; the binding one is **#3201** at position 4, which is `REVIEW_REQUIRED` with no review at all — it gates this PR earlier and harder than #3217 does. #3217 (position 10) is above this PR and so does not gate it; for the record its head has moved since my last comment (`b09d15d…` → `baedfe2…`), so that ❌ is superseded and awaiting re-review rather than standing.

**CI**: Test (15m6s), Type Check and Docker Build Smoke pass; Build, Integration Tests and Lint are still `pending`. `mergeStateStatus: UNSTABLE` reflects those pending jobs, not a failure.

### Findings

### Verdict
✅ **Approve**

---

piotrswierzy (approved):
## Tech Lead re-verification (head unchanged at `9942f6f2`)

Position **8 of 18** — the deepest of my nine. **Nearest blocking ancestor: #3201** (position 4, `REVIEW_REQUIRED`); a second, further-down blocker has appeared at #3217, but it is *below* this PR and does not gate it. `MERGEABLE`/`BLOCKED`. CI in flight at run `34845716443`: `Docker Build Smoke Test`, `Detect changed paths`, `PHP Unit Tests` pass; five **pending**; **no failures**.

Precedence and neutralisation re-verified at this head: ``labelOverrides?.[status] ?? t(`invoice.regulatory.${status}`, LABEL_FALLBACK[status])`` in `regulatory-status-badge.tsx`, with both directions pinned under a **populated** `LocaleProvider` catalog — an override outranks a real catalog entry for the same status, and a catalog entry outranks the neutral fallback when no override is given. That is the assertion that would silently pass against the host's empty catalog, so testing it under a populated one is what makes it mean anything. `invoice-timeline.tsx` is neutralised and guarded by `expect(container.textContent).not.toMatch(/KSeF/i)`, and the plugin now supplies the labels via `labelOverrides={KSEF_REGULATORY_STATUS_LABELS}` from `plugins/ksef/lib/`.

### Findings

### Verdict
✅ **Approve**

---

PR #3215, opened 2026-09-12
docs: correct the buyer-tax-ID source claim in architecture-overview

norbert-kulus-blockydevs:
## Summary

`docs/architecture-overview.md` claimed PrestaShop was the only order-source adapter supplying the buyer tax ID, with Allegro and WooCommerce named as never reading one. That was accurate as of #2599, but #2822 has since given Allegro, Erli, and WooCommerce their own conditional reads:

- **Allegro** and **Erli** — only when the buyer requests a VAT invoice with company data at checkout (Allegro's `invoice.address.company` / `ids[]`; Erli's `user.invoiceAddress.nip`, read only on the invoice address).
- **WooCommerce** — only when the store runs a VAT-number plugin writing one of a short allowlisted `meta_data` key set (`WOOCOMMERCE_VAT_META_KEY_ALLOWLIST`); WC core has no native tax field.
- **PrestaShop** — unchanged, from `ps_address.vat_number` on both billing and shipping addresses.

Verified directly against each of the four order-source adapters (all call the shared `readSourceBuyerTaxId` coercer; none emits the reserved asserted-none middle state in production — only test fixtures do).

- Corrected the per-source claim in `docs/architecture-overview.md` § Sales Documents to name all four sources and their actual conditions.
- Found and corrected the same stale two-source framing repeated in two related ADRs: [ADR-041](../blob/main/docs/architecture/adrs/041-sales-document-routing-policy.md) (`Coverage is one source` note, both instances of the claim on that page) and [ADR-070](../blob/main/docs/architecture/adrs/070-sales-document-market-discovery.md) (motivating context in `## Context`). Both ADRs are `Proposed` (not yet `Accepted`), so this is a factual correction to a side note rather than an edit to an accepted decision.
- No other occurrence of the stale claim found elsewhere in the repo (checked docs, code comments, mockups).

## Test plan

- [x] `node scripts/check-repo-urls.mjs` — OK
- [x] `node scripts/check-adr-numbers.mjs` — OK
- [x] `node scripts/check-architecture-gates.mjs` — OK
- [x] Pre-commit hook (lint → type-check → smart-test) — passed; smart-test correctly scoped the doc-only diff to "none — no workspace package touched"

Closes #3185
Refs #3173

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3037, opened 2026-09-09
chore(listings): delete the unmounted CategoryPicker

jakubretajczykBD:
## Summary
- `CategoryPicker` had no production mount — not exported from `features/listings/index.ts`, no importer besides its own test, fully superseded by `BulkCategoryChooseModal` since #1741.
- `bulk-edit-modal.tsx` still named its state `categoryPickerOpen` while that state actually opens `BulkCategoryChooseModal`, so a grep for `CategoryPicker` returned hits inside the live modal and the dead component read as mounted — it was even named as the primary target file when scoping #2075.
- Deleted `category-picker.tsx` + its test, dropped the now-pointless `vi.mock` stub, renamed `categoryPickerOpen` → `variantCategoryModalOpen` to match the modal it actually opens, and fixed six stale prose references (`bulk-edit-modal.schema.ts`, `plugin.types.ts`, `category-tree-browser.tsx`, `bulk-category-choose-modal.tsx`, `index.css`, and the `connection` prop doc comment) that still named `CategoryPicker` as if it were live.

## Test plan
- [x] `pnpm --filter @openlinker/web lint` — 0 errors (122 pre-existing warnings in unrelated files)
- [x] `pnpm --filter @openlinker/web type-check` — clean
- [ ] `pnpm --filter @openlinker/web test` — left for CI
- [x] `grep -rn "CategoryPicker" apps/web/src` — only `ShopCategoryPickerModal` hits and deliberate historical mentions of the deleted component

Closes #2130

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3216, opened 2026-09-12
refactor(web): remove the disabled buyer-tax-ID checkbox

### LINKED ISSUE #3182: refactor(web): remove the disabled buyer-tax-ID checkbox
## Merge strategy — 🔗 STACKED, on top of #3181

**Not direct to `main`.** This slice ships inside the #3173 stack (#3174 → #3175 → #3179 → #3176 → #3177 → #3178 → #3180 → #3181 → #3182 → #3183 → #3184 → #3185 → #3186): PR #3216 targets #3181's branch, not the default branch. One consequence is worth stating plainly — GitHub honours `Closes #3182` only on a merge into `main`, so this issue does **not** auto-close when its own PR merges; it closes when the chain reaches `main`, or by hand once it has.

---

Part of #3173.

## Context
The rule composer carries a **disabled** checkbox reading *"Include the buyer's tax ID on the receipt, where the destination supports it"*, with copy explaining what an adapter cannot do. There is no choice to make: if the order has a tax number it goes to the adapter, and the adapter decides what to do with it.

## Scope
Delete the control and its copy. No replacement.

## Acceptance
- [ ] No toggle anywhere gates passing the buyer's tax number

**Depends on**: #3181 (the slice below it in the stack).

norbert-kulus-blockydevs:
## Summary

- Removes the permanently-disabled "Include the buyer's tax ID on the receipt, where the destination supports it" checkbox from `SalesDocumentRuleComposerDialog`.
- The checkbox was never wired to any state (no `checked`/`onChange`) and never read by `handleSave`/`toConditionInput` — it was dead UI kept only so the composer's shape wouldn't need to change once eparagony.pl's adapter gained a `buyerTaxId` field.
- There is no choice to make: if the order carries a tax ID it always reaches the adapter, which decides what to do with it. `buyerHasTaxId` survives purely as a rule condition, unaffected.
- Removed the now-orphaned `.rule-composer-section__footnote-row` CSS rule (no other consumer).
- Updated the component's doc comment and the two tests that previously asserted the checkbox's (disabled) presence to assert its absence instead.

This matches `docs/plans/mockups/sales-document-rule-composer.html` tab 05, "The tax-ID checkbox, removed": *"Gone. No toggle, because there is no choice to make."*

## Test plan
- [x] `pnpm --filter apps/web type-check` — passes
- [x] Pre-commit hook ran full-repo `lint` + `type-check` + `smart-test` (scoped to `apps/web`, 19 files / 474 tests) — all passed
- [x] Verified no remaining reference to the checkbox's id/copy anywhere in `apps/web`

Closes #3182
Refs #3173


piotrswierzy (approved):
## Tech Lead review

Deletes a permanently-disabled checkbox that was never wired to state and never read by `handleSave`/`toConditionInput`, plus its now-orphaned CSS rule. Exactly the scope issue #3182 asks for, and the reasoning is right: there is no property to gate, because a tax ID the order carries always reaches the adapter — the toggle was modelling an adapter gap as an operator choice.

The part most easily got wrong is done correctly: the two tests were **rewritten to assert absence** (`queryByLabelText(...).not.toBeInTheDocument()`) rather than deleted. Deleting them would have left nothing stopping the control coming back; asserting absence makes the removal a pinned property. The docblock was updated in the same commit rather than left describing a control that no longer exists, and `.rule-composer-section__footnote-row` had exactly one consumer — verified, no other reference in `apps/web/src` or the mockups.

### Findings

**[SUGGESTION]** — PR description's spec citation
> The description cites `docs/plans/mockups/sales-document-rule-composer.html` tab 05, which is not on `main` — it lands with #3174 / PR #3199, still open. Merging this first is safe (I checked #3199's copy: it already reads *"The tax-ID checkbox, removed … no choice to make"*, so the mockup arrives already consistent with this change and the `docs/frontend-architecture.md` § UX Mockups "stale mockup" obligation is satisfied on arrival). Worth knowing only so the dangling path in the description isn't mistaken for a missing file.

No blocking or important issues.

### Verdict

✅ **Approve**

---

norbert-kulus-blockydevs:
## Review addressed — no code change needed

- **[SUGGESTION] PR description's spec citation** — *acknowledged, nothing to change in the tree*. The mockup is not on `main`, but it **is** tracked on this branch (introduced in-stack by `acf05703f`, from #3199), so the path only reads as dangling when resolved against `main` rather than against the stack. Its copy already says "The tax-ID checkbox, removed" and contains zero tax-ID checkbox inputs, so it arrives already consistent with this removal — matching your own check.

No commit on this branch; it carries only the stack's propagated merges.

---

piotrswierzy (approved):
## Tech Lead re-review (head moved to `6cac4923` — content unchanged since `f106ad58`)

The head SHA moved since my approval, but unlike its two siblings in this stack the change is **ancestor merges only** (`fe7dd4c0`, `6cac4923`, both merges of `3181-ksef-status-labels-to-plugin`) — no new content commit. The reviewable diff is byte-identical to what I approved. Re-verified against today's `main` (`e336cf433`, 13 PRs newer than my review) rather than restating the earlier pass.

Two things I re-checked because `main` moved. **The orphaned-CSS claim still holds**: `.rule-composer-section__footnote-row` has exactly one consumer on today's `main` — the very `<div>` this PR deletes — so removing the rule leaves nothing dangling, and `scripts/check-css-structure.mjs` (#2674) covers the structural half of that edit. **There is no conflict** with `#3037` (`8c6bd4410`), the one merge since my review that also touched `apps/web/src/index.css`: it edits lines `8401–8457` (the dead `.category-picker*` rules) while this PR edits `~3826`. Far apart, no overlap.

The removal is still properly pinned. Both tests were rewritten to assert **absence** rather than deleted, so the checkbox cannot come back unnoticed — and the first one now asserts absence on *both* branches of the document-type select (the `invoice` case already asserted it; this PR makes the `fiscal-receipt` case do the same), which is the right shape: the old test's "disabled, and only for Receipt" framing is exactly what stops being true. The docblock rewrite is the load-bearing part and is accurate — it replaces "the checkbox submits NOTHING, it exists so the shape need not change when the adapter gap closes" with the real reason there is no toggle at all: a tax ID on the order reaches the adapter unconditionally, so there was never a property on the rule or the routing decision to gate. That is consistent with what I verified on #3215 — all four order sources now supply the field, and none of them gates it on a per-rule flag.

**Why this is not merging despite the approval.** GitHub applies `main`'s ruleset to a PR's ancestors, and this sits 9 deep: `main → #3199 → #3200 → #3212 → #3201 → #3210 → #3209 → #3211 → #3214 → #3216`. Eight of those are `REVIEW_REQUIRED`. The nearest blocker is **#3214**, but the root is **#3199**, the only one based on `main` (`mergeable=true`, merely `behind`). Approving #3214 alone will not unblock this — the chain must clear from #3199 upward. Nothing in the chain is `CHANGES_REQUESTED`.

CI: **no checks reported** on this branch (the workflow triggers on `pull_request: branches: [main]` only). Absent, not passing.

### Findings

### Verdict
✅ **Approve** (re-affirmed at `6cac4923`)

---

piotrswierzy (approved):
## Tech Lead re-verification (head unchanged at `43179fe`)

Head has not moved and neither has `main`. Re-checked the property the approval rested on: still a clean three-file deletion (`sales-document-rule-composer-dialog.tsx`, its test, and the orphaned CSS rule), and the tests still assert **absence** rather than being deleted alongside the control — so the removal stays pinned and the checkbox cannot quietly return. The CSS rule's single consumer went with it.

**Stack position:** 9th of 12 in the `main → #3199 → … → #3215` chain, based on `3181-ksef-status-labels-to-plugin` (#3214, approved). Blocked upstream by #3201 / #3210 / #3209, which are still `REVIEW_REQUIRED`; #3217 sits *below* this PR in the chain, so it does not gate this one.

**CI:** `UNSTABLE` — Type Check, Docker Build Smoke and PHP Unit pass; Build, Integration Tests, Lint and Test are still `pending`. Stack-targeted, so partial by construction; pending is absent, not passing.

### Findings

### Verdict
✅ **Approve**

---

piotrswierzy (approved):
## Tech Lead re-verification (head unchanged at `d9611ec`)

`MERGEABLE` / `UNSTABLE` with `reviewDecision: APPROVED`. `UNSTABLE` here is **queue depth, not red**: on run `34845718008` (queued 12:51) `Type Check`, `Docker Build Smoke Test`, `PHP Unit Tests` and `Detect changed paths` pass and `Build` / `Lint` / `Test` / `Integration Tests` are still `queued` at the runner. Nothing has failed.

**Nearest blocking ancestor: #3201** (`REVIEW_REQUIRED`, `BLOCKED`) — confirmed that #3217 sits *below* this PR in the chain and therefore does not gate it. The path to `main` is #3199 → #3200 → #3212 → **#3201** → #3210 → #3209 → #3211 → #3214 → this.

Spot-check at the current head: still a clean three-file deletion, the two tests assert **absence** (`queryByLabelText(...).not.toBeInTheDocument()`) rather than being deleted, and `.rule-composer-section__footnote-row` has exactly one consumer in the tree — the JSX block this PR removes — so the CSS drop leaves nothing orphaned in the other direction.

### Findings

### Verdict
✅ **Approve** — merge once #3201 and the approved-but-blocked links above it land, and this run's four queued jobs report green.

---

Issue #2170, opened 2026-08-18
[TASK] CORE+Frontend — Country-agnostic sales-document rule engine (ADR-041 dec. 5) with a Poland-only starter template


norbert-kulus-blockydevs:
## Problem / Context

ADR-041 decision 5 (the per-country rule engine + suggest-auto template modes) was explicitly deferred when epic #2154 shipped — #2155/#2156/#2159/#2160 landed the simpler `operator-configured` single-primary model instead (decision 4). That model has a real gap: exactly one connection is `isPrimary` for a document kind, globally, with no way to route different orders of the SAME kind to different providers based on order facts (e.g. Poland's own public VAT guidance: a fiscal receipt carrying the buyer's tax ID can stand in as a simplified invoice up to a threshold through end of 2026, but must be a full invoice above it — two different providers, one condition).

Design is done: [mockup, Phase 2](https://claude.ai/code/artifact/57908c4c-4471-49c0-8bf3-d3daeffa4bb4) (tabs 02 and 06; also to be committed at `docs/plans/mockups/sales-document-routing-adr041.html`), reviewed for country-agnosticism (an independent architecture pass found and the mockup now documents the fix for a real violation: "Receipt with NIP" must never become a third `SalesDocumentKind` value — it's a `buyerTaxId` property on the fiscal-receipt outcome, not a routing-level kind).

**Scope, confirmed with the product owner:** the rule-building MECHANISM (conditions → document type → integration, per-country defaults, the fallback ladder, the no-priority conflict guard) must be fully country-agnostic and usable by hand for ANY country. The PREDEFINED STARTER TEMPLATE (cited public guidance, "Review & adopt" flow) is Poland-only for this issue — no other country ships a curated template yet, but the template mechanism itself must stay generic (seed data keyed by country code) so a second country's template later is additive.

### Core (`libs/core/src/sales-documents/`)

Three new tables (all in the existing `sales-documents` leaf context per ADR-041 decision 1 — next migration timestamp after `1834000000000-create-fiscal-registration-records.ts`, synthetic sequential per `docs/migrations.md`):

[...]

Issue #2673, opened 2026-08-29
check-attention-reason-mirror's pending `restock-blocked` pair compares nothing, and the two strings have already diverged


piotrswierzy:
Found during the OMS Wave 2 body-E merge (PR #2645). This is the M6 hazard `check-attention-reason-mirror.mjs`'s own docblock predicted, now materialised.

## The defect

The script carries a **pending** pair for `restock-blocked`, keyed on a declaration named `RETURNS_RESTOCK_BLOCKED_COPY` in `returns-list.copy.ts`.

Body E ships that copy as **`RETURN_RESTOCK_BLOCKED_COPY`** — singular — in a **different file**, `features/returns/lib/restock-blocked.copy.ts` (it was moved there deliberately, on #2357's behalf).

So the pair matches nothing on the frontend side. It stays permanently pending, compares no strings, and passes vacuously.

## It is already hiding a divergence

The two strings it was supposed to keep in step have drifted:

| Side | Value |
|---|---|
| core mirror (body C) | `title: 'Stock was not added'` |
| `restock-blocked.copy.ts` (body E) | `title: 'Stock was not added.'` |

A trailing period — cosmetic in itself, and exactly the class of drift the mirror exists to catch. The point is not the period; it is that **the guard reported green while the thing it guards was already inconsistent.**

## Why nothing was changed in #2645

Neither string was edited and nothing was renamed, deliberately: #2364 owns the canonical `restock-blocked` copy, and silently satisfying a mirror by editing one side is how a guard gets converted into a formality. Recorded here instead.

## Scope

1. Decide which spelling is canonical (`#2364` owns this).
2. Re-key the pending pair onto the real declaration name **and file**, then take it out of pending.
3. Reconcile the trailing period.
4. **Verify by making it fail first** — drift one side deliberately and confirm the check reports it, naming the key. A pending pair that has never been seen to fail is a claim, not a guard.

## The generalisable half

A mirror keyed on a **declaration name** silently degrades to a no-op when that declaration is renamed or moved — with no error, because "not found" and "pending" are indistinguishable to the script. Worth considering whether a pending pair whose target declaration cannot be located should be a **hard failure** rather than a skip: the whole value of the mechanism is that it cannot quietly stop working.

## Acceptance criteria

- [ ] The `restock-blocked` pair resolves against the real declaration and is no longer pending.
- [ ] Deliberate drift on either side fails `pnpm check:invariants`, demonstrated.
[...]

PR #3217, opened 2026-09-12
test(core): vocabulary guard for sales-documents and invoicing

### LINKED ISSUE #3183: test(core): vocabulary guard for sales-documents and invoicing
## Merge strategy — 🔗 STACKED, on top of #3182

**Not direct to `main`.** This slice ships inside the #3173 stack (#3174 → #3175 → #3179 → #3176 → #3177 → #3178 → #3180 → #3181 → #3182 → #3183 → #3184 → #3185 → #3186): PR #3217 targets #3182's branch, not the default branch. One consequence is worth stating plainly — GitHub honours `Closes #3183` only on a merge into `main`, so this issue does **not** auto-close when its own PR merges; it closes when the chain reaches `main`, or by hand once it has.

---

Part of #3173.

## Context
`libs/core/src/fiscalization/__tests__/neutral-vocabulary.spec.ts` is a build-failing sweep over every file in that context, banning `paragon`/`kasa`/`printer`/`eparagony`/`ksef` **in prose as well as identifiers**. Invoicing has no equivalent and already carries ~25 `KSeF` mentions in prose. `sales-document-condition.types.ts` claims a grep verification that **no longer runs anywhere**.

## Scope
Port the fiscalization guard to `sales-documents`, and to `invoicing` at whatever strictness its existing contract-surface rule supports. Include the empty-sweep guard so the test cannot silently pass over nothing.

## Acceptance
- [ ] A country-specific string added to a guarded context fails the build
- [ ] Existing prose that legitimately names a provider is either cleaned or explicitly exempted with a reason

**Depends on**: #3182 (the slice below it in the stack).

norbert-kulus-blockydevs:
## Summary

`libs/core/src/fiscalization/__tests__/neutral-vocabulary.spec.ts` already runs a build-failing sweep banning regime-specific terms in prose and identifiers (ADR-042 decision 4). `sales-documents` states the identical rule in its condition-types doc comment, but the "grep-verified by the acceptance criteria of #2170" claim ran nowhere. `invoicing` states a narrower rule (contract surface only — interface/method/param/return names; prose explicitly exempted) and had no enforcement at all, alongside ~25 legitimate KSeF mentions in prose.

This PR ports the guard to both, at each context's own documented strictness:

- **`sales-documents`** — full sweep (prose + identifiers), mirroring fiscalization exactly, plus the two country-specific terms its own doc comment already names (`nip`, `vat`). Existing violations were **cleaned rather than exempted**: renamed arbitrary test-fixture ids/values away from `eparagony`/`KSEF-REF-*`, and genericized two doc-comment worked examples that named `eparagony.pl` by name. Zero exemptions needed in the final guard.
- **`invoicing`** — contract-surface-only sweep: production (`*.ts`, non-spec) files only, with comments and string/template literals stripped before matching, per this concern's own repeated litmus ("the CONTRACT SURFACE — interface name, method name, parameter/return types — carries no ... vocabulary. The prose above names inFakt/KSeF only as illustrative examples ... those are documentation, not part of the type surface"). This scoping is what keeps the guard from overshooting the rule it enforces — sanctioned prose and test fixtures (`pl-nip`/`eu-vat` scheme values, `KSEF-9` clearance-reference ids, inFakt/KSeF worked examples) stay untouched, and the existing narrower value-level guard in `invoicing.types.spec.ts` (which checks the runtime VALUES of two vocabulary arrays — a different surface) is kept alongside rather than replaced. The production contract surface was already clean; no code needed cleaning.

Both term lists are matched with a two-pass regex (case-insensitive true word boundary, plus a case-sensitive camelCase-segment match) rather than a naive substring test — `sales-documents`' extra terms (`nip`, `vat`) are common substrings of ordinary English words (`private`, `manipulate`) that `.includes()` would false-positive on, while a plain `\b` test alone would miss a term embedded as a capitalized "hump" inside a longer identifier (e.g. a hypothetical `buyerHasNip` field). Verified against a planted camelCase violation in each context's production code (caught, file path named in the failure), and verified the invoicing guard correctly ignores the identical string once moved into a `*.spec.ts` file — proving the "production files only" scoping is load-bearing, not incidental.

Also cleaned the two hardcoded `"KSeF"` strings in the shared, provider-agnostic `InvoiceTimeline` component (`apps/web`) — it renders for every `InvoicingPort` adapter's invoices, not only KSeF's, so its `pending-submission` node now reads "Awaiting submission" / "Issued offline — not yet sent for clearance" rather than naming the regulator, following the same regulator-neutral-by-default pattern #3181 (PR #3214, not yet merged) established for `RegulatoryStatusBadge`. No override mechanism was added since `InvoiceTimeline` has no plugin-specific caller wanting branded wording today — this is a straight neutralization, not a new abstraction.

## A note for whoever hits the hook next

This PR edits `libs/core/src/sales-documents/index.ts` — a **core barrel**. `scripts/smart-test.mjs` classifies any edit to a core top-level barrel as a "wide trigger" (a published-contract change) and runs the **full monorepo `pnpm -r test`** instead of a diff-scoped subset, by design. On a memory-constrained machine, running all 21 workspace packages' test suites in parallel (the default `pnpm -r` concurrency) can exhaust RAM and get jest workers `SIGKILL`ed by the OOM killer — which looks like a test failure but isn't one (every test that actually ran passed; entire suites just never got the chance to start). If you hit this, serialize the workspace for the commit instead of skipping the hook:

```
npm_config_workspace_concurrency=1 git commit -s -F <message-file>
```

This trades parallelism for memory safety and lets the full suite complete reliably, just slower. Confirmed working end-to-end on this PR.

## Test plan

- [x] `pnpm type-check` — clean across the whole repo
- [x] Both new guards run directly (`pnpm --filter @openlinker/core exec jest .../neutral-vocabulary.spec.ts`) — all green
- [x] Planted a camelCase-embedded violation in each context's production code and confirmed the guard fails with the offending file named
- [x] Confirmed the invoicing guard passes once the same planted violation is moved into a `.spec.ts` file (proves the contract-surface-only scoping is real)
- [x] `invoice-timeline.test.tsx` and `invoice-detail-page.test.tsx` pass with the neutralized copy
- [x] Full `pnpm -r lint`, `pnpm -r type-check`, and serialized `pnpm -r test` all pass via the pre-commit hook (`npm_config_workspace_concurrency=1`)

Closes #3183


piotrswierzy (commented):
## Tech Lead review

Ports fiscalization's neutral-vocabulary sweep to `sales-documents` (full, prose included) and `invoicing` (contract surface only), each at its own documented strictness, plus a regulator-neutralisation of the shared `InvoiceTimeline`. The scoping judgement is right on both sides — `invoicing`'s comment/string stripper is what lets ~25 legitimate inFakt/KSeF prose mentions survive while still banning a declared identifier, and proving that by moving a planted violation into a `*.spec.ts` is the correct way to show the scoping is load-bearing rather than incidental. `sales-documents`' violations were **cleaned rather than exempted**, which is the right call and leaves the guard exemption-free.

One problem, and it is the guard's own subject: the two-pass matcher has a hole that makes it **strictly weaker than the `.includes()` it replaces** for the five terms it ports.

### Findings

**[IMPORTANT]** — `libs/core/src/sales-documents/__tests__/neutral-vocabulary.spec.ts:95-103` and `libs/core/src/invoicing/__tests__/neutral-vocabulary.spec.ts` (same function)
> `containsForbiddenTerm` catches a term as a **trailing** camelCase hump (`buyerHasNip` — the doc comment's own worked example) but not as a **leading** one. Pass 1's `(?![a-zA-Z])` lookahead is blocked by the next identifier segment's capital, and pass 2 requires `(?<=[a-z])` which a term at identifier start does not have. Verified by running the shipped function:
>
> ```
> ksef     "const ksefEnvironment = 1;"       false
> nip      "const nipValue = 1;"              false
> vat      "class VatRateResolver {}"         false
> paragon  "const paragonMode = 1;"           false
> nip      "readonly buyerHasNip: boolean;"   true    ← the case the doc cites
> ```
>
> `ksefEnvironment` is not hypothetical — it is a real identifier shape this repo already uses and recommends (`docs/frontend-architecture.md` § `connectionConfig`: "prefix field names with the platform (e.g. `ksefEnvironment`)"). So the most likely real violation in either context is exactly the one the guard cannot see.
>
> This matters more than a normal gap because fiscalization's original (`neutral-vocabulary.spec.ts:57`) is `readFileSync(...).toLowerCase().includes(term)`, which **does** catch all four of the above. The module doc comment presents this as "the identical litmus" and "the #3183 port of the fiscalization litmus" — for `paragon`/`kasa`/`printer`/`eparagony`/`ksef` it is a weakening, and the stated reason for the rewrite (`nip`/`vat` false-positives on `private`/`manipulate`) only ever applied to the two *new* terms.
>
> Concrete fix — add a third, case-**sensitive** pass for the leading segment, keeping the two existing ones (pass 1 must stay case-insensitive or `// KSeF clearance` in prose stops being caught):
>
> ```ts
> const leading = new RegExp(`(?<![a-zA-Z])(?:${term}|${titled}|${allCaps})(?![a-z])`);
> if (leading.test(text)) return true;
> ```
>
> I ran the three-pass version against both contexts at this PR's head: **zero additional offenders** in `sales-documents` (your cleanup already covers it) and zero in `invoicing` production files, and no new false positives on `private` / `activate` / `derivation` / `innovation` / `manipulate` / `principal`. So this is a drop-in strengthening with no extra cleanup cost.

**[SUGGESTION]** — three copies of `collectSourceFiles` + `containsForbiddenTerm`
> With this PR there are three near-identical sweeps (`fiscalization`, `sales-documents`, `invoicing`) differing only in term list, file filter and whether comments/strings are stripped. That is three places to apply the fix above, and three places for the next divergence. Consider one shared helper (a `libs/core/src/__tests__/` module, like `barrel-purity.spec.ts`'s neighbours) parameterised by `{ terms, includeTests, stripNonCode }`, with each context keeping its own thin spec. Not blocking — but it is what would have made the fiscalization-vs-new-matcher divergence above impossible.

**[SUGGESTION]** — `libs/core/src/sales-documents/domain/types/sales-document-condition.types.ts:6-8`
> Choosing the prose-inclusive sweep costs this docblock the concrete contrast that made the rule teachable ("`buyerHasTaxId`, not `buyerHasNip`" → "not a national tax-identifier name"). The reasoning for the choice is sound (ADR-041 states the ban over "this concern" without a prose carve-out, unlike ADR-026), and the replacement still conveys the rule — but it is worth noting that the sibling guard shipping in this same PR demonstrates the alternative, so the ADR could equally be amended to carve prose out rather than the docblock genericised. Fine to leave as-is; flagging so the choice is visible to whoever revisits ADR-041.

### Verdict

🔄 **Approve with changes**

1. Add the leading-segment pass to `containsForbiddenTerm` in both new specs (verified drop-in, no cleanup follow-on), and correct the module docblock's claim that the matcher is equivalent to fiscalization's.

Done well and worth stating: the `finds source files to check` empty-sweep guard is present in both (this is the `#2673` "a check that cannot fail" class, caught deliberately); `stripCommentsAndStringLiterals` is hand-rolled but fails in the safe direction (a regex literal reads as code, i.e. over-flags rather than under-flags); and the `InvoiceTimeline` neutralisation — which neither new guard covers, since both scan `libs/core` only — is nonetheless pinned by the added `expect(screen.queryByText(/ksef/i)).toBeNull()`, which is the right way to cover an FE surface the sweep cannot reach.

---

norbert-kulus-blockydevs:
## Review addressed — commit `7977ecb8d`

- **[IMPORTANT] `containsForbiddenTerm` misses a term as a leading identifier segment** — *fixed*. Reproduced your finding against the shipped regexes with `grep -P`: `ksefEnvironment`, `nipValue`, `VatRateResolver`, `paragonMode` all passed. Added your third case-sensitive leading pass, keeping both existing ones. The "identical litmus / port of the fiscalization litmus" claim is corrected in all three docblocks plus `sales-document-condition.types.ts`: same **rule**, different **matcher**, with the residual difference stated — a term followed immediately by a lowercase letter (an inflection or plural) is invisible to the three-pass matcher but visible to fiscalization's `.includes()`.

  **The sweep you asked for, run by hand** (the worktree has no `node_modules`, so I replicated `stripCommentsAndStringLiterals` plus all three passes and swept both contexts with the specs' exact file semantics): `sales-documents` — 58 files, all 7 terms clean. `invoicing` — 79 production files, all 5 terms clean; all 22 raw hits sit in JSDoc/`//` comments, verified with no non-comment line matching, so stripping removes every one. **No real offender exists** — nothing to rename. Negative cases re-verified too: `private`, `activate`, `derivation`, `innovation`, `manipulate`, `principal`, plus `PRIVATE_ACTIVATE_DERIVATION`, `Private`, `ActivateHandler` and `nipple` all stay clean. That matches the zero-additional-offenders result you reported.
- **[SUGGESTION] Three near-identical sweeps** — *partly fixed, and the rest is a real decision I did not take for you*. Extracted `libs/core/src/__tests__/neutral-vocabulary-sweep.ts` (`collectSweepFiles({includeTests})`, `stripCommentsAndStringLiterals`, `containsForbiddenTerm`, `readSweepSource({stripNonCode})`); all three specs are now thin, net −107 lines. Imports are **relative** (`../../__tests__/…`), which `barrel-purity.spec.ts:263` skips, so `sales-documents`' zero-sibling-edge leaf property is untouched; the helper is not a `.spec.ts` so `testRegex` never collects it, and it sits outside every context root so no sweep scans it.

  **The matcher half is left open.** `fiscalization` keeps its `.includes()` and shares only file collection, because the two matchers are **not reconcilable without a behaviour change in one direction or the other**: giving fiscalization the three-pass matcher *weakens* it for its own five terms (it loses the inflected-term shape); giving the two new contexts `.includes()` fails the build on ordinary English (`private`, `manipulate`). Resolving it needs a per-term policy, which is a call for you rather than a refactor. Recorded in the shared module and in fiscalization's own header so the next author meets it.
- **[SUGGESTION] Docblock lost the concrete contrast** — *fixed*. `sales-document-condition.types.ts` gains a "A COST OF THAT CHOICE" paragraph recording that the prose-inclusive sweep is why the side-by-side contrast can no longer be written there, and that the alternative (amend ADR-041 decision 5 to carve doc comments out, as ADR-026 does for `invoicing`, whose sibling guard in this same PR demonstrates that shape) was weighed and declined — with the note that changing it means changing the ADR and the sweep's scope together. The wording is deliberately term-free, and the full sweep was re-run after the edit to prove that file still passes its own guard.

---

piotrswierzy (commented):
## Tech Lead re-review (`87bb2395` → `b09d15d0`)

Both halves of my prior finding are **closed, and closed the right way**. I re-fetched `libs/core/src/__tests__/neutral-vocabulary-sweep.ts` at the head commit, extracted `containsForbiddenTerm` verbatim, and executed it — `ksefEnvironment`, `nipValue`, `VatRateResolver` and `paragonMode` now all return `true` via the new pass 2 (LEADING), while `private` / `derivation` / `manipulate` still produce no false positive under the real term lists. Equally important: `fiscalization` keeps its `.includes()` rather than adopting the three-pass matcher, so the five ported terms are not weakened there, and the residual asymmetry (a term followed by a lowercase letter) is recorded once, in the shared module, as an owner decision rather than quietly resolved. The `collectSweepFiles` / `readSweepSource` split, the `files.length > 0` empty-sweep guard, and the `stripNonCode` asymmetry between the invoicing contract-surface sweep and the two full sweeps are all correct and well-argued.

There is, however, one new BLOCKING problem that arrived with the base-branch merge rather than with your matcher fix: **the sales-documents sweep now fails at this PR's own head.**

### Findings

**[BLOCKING]** — `libs/core/src/sales-documents/domain/domain-services/evaluate-sales-document-rules.spec.ts:113` and `libs/core/src/sales-documents/application/services/sales-document-rules.service.ts:105`
> The guard this PR adds **does not pass on this PR's head commit**. I fetched the whole `libs/core/src/sales-documents` subtree at `b09d15d0`, ran the shipped `containsForbiddenTerm` with the shipped `FORBIDDEN_TERMS` and `{stripNonCode: false}` / `{includeTests: true}` scope, and got two offenders:
>
> - `evaluate-sales-document-rules.spec.ts:113` — `connectionId: 'conn-eparagony'`, inside the `pl`/`PL` country-fold test case. Your diff renamed six occurrences of that literal to `conn-receipt-only`; this is a **seventh**, added by #3201 (`normalise-market-country-codes`) and pulled in by the merge from `3182-remove-disabled-tax-id-checkbox`. Matches on pass 1 (boundary: `-` before, `'` after).
> - `sales-document-rules.service.ts:105` — the doc comment listing sibling copies of `normaliseCountry` names `apps/web/src/plugins/ksef/lib/ksef-seller-config.ts`. Also from #3201. Matches on pass 1 (`/` before, `/` after). This is prose, and the sales-documents sweep is deliberately prose-inclusive, so the file-path citation is a real offender under the rule you wrote — which is exactly the cost `sales-document-condition.types.ts`'s new "A COST OF THAT CHOICE" paragraph predicts, landing one file over.
>
> Rename the literal to `conn-receipt-only` like its six siblings, and rewrite the path citation so it does not spell the plugin directory (e.g. "the corresponding helper in the `apps/web/src/plugins/<provider>/lib/` seller-config module", or drop the fourth example and keep the three neutral ones — the comment already explains why they cannot share a definition).
>
> **CI will not catch this for you.** `gh pr checks 3217` reports *no checks reported on the branch* — the workflow triggers on `pull_request: branches: [main]` and this PR targets `3182-remove-disabled-tax-id-checkbox`. The first red signal would be on whichever PR in the stack first reaches `main`, and by then the offending line is several merges deep.

**[SUGGESTION]** — `libs/core/src/__tests__/neutral-vocabulary-sweep.ts`
> The module's own doc comment contains `ksefEnvironment`, `nipValue`, `VatRateResolver` and `paragonMode` as worked examples. That is fine today and the header says why ("It lives outside every context directory, so no context's own sweep scans it either") — but it makes the file un-relocatable into any swept directory. Worth keeping the sentence exactly where it is if anyone later tidies the `__tests__` layout.

**[SUGGESTION]** — `apps/web/src/features/invoicing/components/invoice-timeline.tsx`
> The copy change (`not yet sent to the authority` → `not yet sent for clearance`) is unrelated to the guard this PR adds — `apps/web` is scanned by none of these three sweeps. It is a fine change and the docblock justifies it; just noting it is scope beyond the title.

### Positive observations

- Keeping `fiscalization` on `.includes()` and *documenting why unifying would weaken it* is the right call, and the note is in the one place both readers will find it.
- The invoicing sweep's two-restriction operationalisation of "contract surface" (production files only + comments/strings stripped) matches ADR-026's own carve-out rather than overshooting it, and the header says so explicitly. I ran that sweep too against the head tree: **clean, zero offenders**.
- Every one of the three specs carries a `files.length > 0` assertion, which is the vacuity guard `docs/testing-guide.md § Port-contract suites` asks for.

### Verdict
❌ **Request changes** — one BLOCKING item: the new sales-documents sweep fails on this PR's own head, on two lines the base-branch merge introduced.

---

norbert-kulus-blockydevs:
## BLOCKING closed — commit `f6f382337`

Both offenders fixed, exactly as specified.

- `evaluate-sales-document-rules.spec.ts` — the seventh `'conn-eparagony'` literal renamed to `conn-receipt-only` like its six siblings. Asserted it was the only one left before renaming, so the fix cannot have been partial.
- `sales-document-rules.service.ts` — the sibling-copies doc comment no longer spells the provider plugin directory. It now cites that helper **by role** ("the seller-config module of a provider plugin under `apps/web/src/plugins/`") and says in place *why* the path had to go: this context's sweep is prose-inclusive, so a path citation is an offender under the very rule the comment explains. The #591 reasoning for why the four copies cannot share a definition is kept intact.

**Verified the way you verified it.** I replicated the shipped three-pass `containsForbiddenTerm` and `collectSweepFiles` semantics and swept the whole context at `{stripNonCode: false, includeTests: true}`: **58 files, zero offenders**, for all seven terms. Same file count you reported. Re-run after the round-2 propagation reached the tip — still clean there.

Your point about CI is taken and acted on one level up: the merge-strategy headers on all 13 stack issues now say the PR targets its predecessor's branch and that `Closes #N` therefore does not fire on its own merge. It does not give this branch CI, but it stops the issue and the branch disagreeing about what will happen.

- **[SUGGESTION] The sweep module's own doc comment holds worked examples** — *left exactly where it is*, per your closing sentence. The header already explains why it is safe there; nothing moved.
- **[SUGGESTION] The `invoice-timeline.tsx` copy change is scope beyond the title** — *agreed, and stated for the record*: it arrived through the conflict resolution when #3214's independent neutralisation of the same two strings met this branch's own. The downstream wording won, which is what a normal merge order produces.

---

piotrswierzy (commented):
## Tech Lead re-review (`b09d15d0` → `baedfe2e`)

**The blocking finding is closed, and I verified it by re-running the shipped guard rather than by reading the description.** I fetched the `libs/core/src/sales-documents`, `invoicing` and `fiscalization` subtrees at `baedfe2e`, extracted `collectSweepFiles` / `stripCommentsAndStringLiterals` / `containsForbiddenTerm` verbatim from `libs/core/src/__tests__/neutral-vocabulary-sweep.ts`, and executed all three contexts' exact configurations (7/5/5 terms, `includeTests` true/false/true, `stripNonCode` false/true/`.includes()`). **Result: 58 + 79 + 37 = 174 files swept, 0 offenders.** Both offenders I named last round are genuinely gone — the seventh `conn-eparagony` literal is now `conn-receipt-only` (`evaluate-sales-document-rules.spec.ts:113`, consistent with the six the diff had already renamed), and the `sales-document-rules.service.ts:105` doc comment now cites the plugin helper **by role** with an explicit note that spelling the directory would make the comment an offender under the very rule it explains. That is the right shape: the fix explains the constraint rather than weakening the sweep. I also confirmed nothing was weakened to get there — term lists, `includeTests` and `stripNonCode` are byte-identical to what I reviewed, and the anti-vacuity `expect(files.length).toBeGreaterThan(0)` is still in all three specs. CI now reports (Build / Test / Type Check / Docker pass; Integration pending) where it reported nothing last round.

I separately re-probed the matcher's behaviour directly: `ksefEnvironment` / `nipValue` / `VatRateResolver` / `paragonMode` → `leading`; `buyerKsefRef` / `sellerNIP` → `hump`; `KSEF_URL` → `boundary`; and no false positive on `private`, `derivation`, `manipulate`, `principal`, `innovation`, `activate`. The documented residual (`printers` invisible to the three-pass matcher, visible to `fiscalization`'s `.includes()`) reproduces exactly as the module says it does. The honesty of that module doc comment is worth naming — it states its own weaker direction instead of implying parity.

### Findings

**[IMPORTANT]** — `libs/core/src/__tests__/neutral-vocabulary-sweep.ts:172`
> `containsForbiddenTerm` is the subtlest thing this PR ships — three regexes turning on lookbehind/lookahead interactions — and **nothing pins its behaviour.** There is no `neutral-vocabulary-sweep.spec.ts`; `libs/core/src/__tests__/` holds only `barrel-purity.spec.ts` and `no-direct-buffer-read.spec.ts`. Delete the LEADING pass tomorrow and every one of the three sweeps still reports green, because the only evidence it works is that the forbidden terms happen not to appear right now. That is precisely the hole this PR already shipped once and then fixed in `7977ecb8` — and it would come back silently. `docs/testing-guide.md § Port-contract suites` states the repo's own standard for exactly this class of machinery: *"a contract suite is exactly the machinery that can look thorough and assert nothing"*, and requires one deliberate breakage fixture per declared rule. Add a spec beside the module asserting the four positive shapes above resolve to `leading`/`hump`/`boundary` and that the six ordinary-English words resolve to `null` — ~30 lines, and it converts a property currently held by coincidence into one held by the build. (Not blocking: I have executed the matcher and it is correct **today**.)

**[SUGGESTION]** — `libs/core/src/__tests__/neutral-vocabulary-sweep.ts:29`
> `SWEEP_SPEC_FILE_NAME` excludes by bare filename, so any file anywhere under a swept context called `neutral-vocabulary.spec.ts` is skipped — not just the one in `__tests__/`. Harmless today; scoping the exclusion to the context's own `__tests__` directory would make it say what it means.

### Verdict
🔄 **Approve with changes** — merge once the matcher spec lands, or file it as an immediate follow-up. The ❌ from last round is resolved and verified by execution.

---

norbert-kulus-blockydevs:
## Review addressed — commit `39568eb0b`

- **[IMPORTANT] `containsForbiddenTerm` has no spec** — *fixed*. Added `libs/core/src/__tests__/neutral-vocabulary-sweep.spec.ts`, 44 cases, pinning **each pass separately** plus `stripCommentsAndStringLiterals` and `collectSweepFiles`. Filesystem assertions follow `barrel-purity.spec.ts` (read the real tree, write no fixtures), and the spec asserts its own location is outside every swept context — verified rather than assumed (`relative(<ctx>, __filename)` starts with `..` for all three, and none of the three collections contains it).

  One design point worth your eye: making a single pass pinnable required the module to also export **`matchForbiddenTerm`**, which reports *which* pass answered. `containsForbiddenTerm` — the API all three sweeps consume — is now `matchForbiddenTerm(...) !== null`, behaviour-identical, verified over all 174 swept files × every term with **zero** divergence. A boolean cannot tell a live pass from a deleted one wherever two passes both match, so a boolean-only spec would have pinned nothing in exactly the overlap cases.

  **Red-first mutation matrix — 13 mutations, none survived** (jest is unavailable in the worktree, so the shipped matcher was replicated verbatim, diffed token-for-token against the TS, and executed):

  | Mutation | Named cases that fail |
  |---|---|
  | delete BOUNDARY pass | 4 boundary cases + totality |
  | delete LEADING pass | all 4 leading cases + totality |
  | delete HUMP pass | 3 hump cases + totality |
  | boundary loses `i` flag | `// KSeF clearance`, `KSEF_BASE_URL` |
  | leading loses `(?![a-z])` | `nipple`, `printers` |
  | hump becomes case-insensitive | `turnip`, `orderVatAmount` |
  | exclusion back to bare filename | "exempts by path, not any file of that name" |
  | `containsForbiddenTerm` always true | totality + readSweepSource |
  | block comment → `''` not `' '` | 2 stripper cases |
  | `readSweepSource` flag inverted | 2 readSweepSource cases |
  | collector stops recursing | 3 collector cases |
  | collector ignores `includeTests` | 1 collector case |
  | escaped-quote handling removed | 1 stripper case |

- **[SUGGESTION] `SWEEP_SPEC_FILE_NAME` excludes by bare filename** — *fixed*. Resolved once against the root as `<contextRoot>/__tests__/neutral-vocabulary.spec.ts`. All three sweeps still collect the identical 37/58/79 files and still report 0 offenders.

**One hole found and deliberately NOT pinned.** `KSeFEnvironment` — the regime's own mixed-case branding used as a *leading* identifier segment — matches **no** pass: pass 1 is blocked by the trailing `E`, and passes 2/3 alternate over only `ksef`/`Ksef`/`KSEF`. It is left unpinned because a test would enshrine the hole, and widening the leading pass is the per-term matcher-strength decision the module's own docblock reserves for the owner. Flagging it so the decision is yours rather than discovered later.

---

piotrswierzy (approved):
## Tech Lead re-review (`baedfe2e` → `483320ec`)

Delta is one own commit (`39568eb0`, the 414-line `neutral-vocabulary-sweep.spec.ts` plus a 69/25 change to the matcher) and six base-catch-up merges carrying #3201's migration renumber. **Both prior findings are now closed.** I re-ran the shipped guard at head — `collectSweepFiles` / `stripCommentsAndStringLiterals` / `matchForbiddenTerm` extracted verbatim, each context's exact config — and it sweeps **174 files (sales-documents 58, invoicing 79, fiscalization 37) with 0 offenders**, identical to last round. Nothing was weakened: term lists are byte-identical (`sales-documents` 7 terms raw + tests, `invoicing` 5 terms stripped + production-only, `fiscalization` 5 terms `.includes()` + tests), `includeTests` / `stripNonCode` unchanged, and the `files.length > 0` anti-vacuity assertion is present in all three. The open IMPORTANT — no spec pinned `containsForbiddenTerm`, so deleting the LEADING pass left all three sweeps green — is genuinely closed, and closed the right way.

The matcher now reports **which** pass answered (`matchForbiddenTerm → 'boundary' | 'leading' | 'hump' | null`), with `containsForbiddenTerm` retained as the sweeps' boolean. That discriminant is what makes each pass separately pinnable, and the docblock says so in exactly those terms. I mutation-tested the shipped matcher against the spec's own case tables; every declared rule is red-first:

| Mutation | Result |
|---|---|
| delete BOUNDARY pass | **RED** — 4 cases (`KSeF` → `null`; three others → `'leading'`) |
| delete LEADING pass | **RED** — 4 cases, all → `null` *(the exact hole this PR shipped once)* |
| delete HUMP pass | **RED** — 3 cases, all → `null` |
| boundary pass loses its `i` flag | **RED** — 2 cases |
| leading pass loses `(?![a-z])` | **RED** — `nipple`, `two printers` start flagging |
| hump pass made case-insensitive | **RED** — `turnip` starts flagging, `orderVatAmount` stops |

Each goes red on precisely the fixtures its own docblock names it for. The anti-vacuity block (`declared === covered`, failing on either side) is present, and the boolean-vs-discriminant drift assertion is there too — a boundary-only refactor of `containsForbiddenTerm` would leave every per-pass assertion true and every sweep wrong without it.

I also executed the non-matcher assertions against the real head tree, and none is vacuous: the sample-context guard is `[true,true,true,true]`; `includeTests:true` equals *(all 37 `.ts` − own sweep spec)* exactly; `includeTests:false` is strictly smaller (29 < 37); `readSweepSource`'s prose-only assertion finds **34** real `invoicing` files, so the strip flag is pinned against the tree that relies on it; and this spec is confirmed outside all three swept context roots.

One unrequested improvement worth naming, because it is a real narrowing rather than a tidy-up: the sweep-spec exemption moved from a bare **filename** test to a **full path** under the context root (`<ctx>/__tests__/neutral-vocabulary.spec.ts`). A filename test skipped any file of that name anywhere in the tree — a wider exemption than its reason supports. I verified all three contexts really hold their spec at that relative path, and that the spec's own assertion discriminates: rooted at `fiscalization/__tests__`, the shipped path-scoped collector returns the file while a bare-filename mutant returns `[]`, so the mutant fails that test.

### Findings

### Verdict
✅ **Approve**

---

PR #3218, opened 2026-09-12
chore(invoicing): confirm the one-document guard for dual-role connections

### LINKED ISSUE #3184: chore(invoicing): confirm the one-document guard for dual-role connections
## Merge strategy — 🔗 STACKED, on top of #3183

**Not direct to `main`.** This slice ships inside the #3173 stack (#3174 → #3175 → #3179 → #3176 → #3177 → #3178 → #3180 → #3181 → #3182 → #3183 → #3184 → #3185 → #3186): PR #3218 targets #3183's branch, not the default branch. One consequence is worth stating plainly — GitHub honours `Closes #3184` only on a merge into `main`, so this issue does **not** auto-close when its own PR merges; it closes when the chain reaches `main`, or by hand once it has.

---

Part of #3173.

## Context
Both cross-kind guards are already **connection-blind** — the invoice side searches by `blocksFurtherRegistration` with no connection filter, and the fiscal side calls `findBlockingInvoiceForOrder(orderId)` with no connection argument, both under the same per-order lock. A connection holding two roles therefore cannot produce both documents for one sale. An earlier plan wrongly treated this as a blocker for the whole eparagony-invoicing epic.

## Scope
- Add a test pinning the guarantee for a **dual-role** connection
- Correct the docblock in `invoice.service.ts` claiming an invoicing connection id can never collide with a fiscalization one — true today, false once one connection holds both

## Acceptance
- [ ] A test fails if either guard ever gains a connection filter
- [ ] No comment in the file asserts the collision is impossible

**Depends on**: #3183 (the slice below it in the stack).

norbert-kulus-blockydevs:
## Summary

Confirms — with a regression test, not a fix — that OpenLinker's existing
one-document-per-order guard already holds when a single connection carries
BOTH the invoicing and fiscalization roles (ADR-072 decision 3).

Both cross-kind checks were already **connection-blind**:

- `InvoiceService.assertNoBlockingFiscalReceipt` reads every
  `FiscalRegistrationRecord` for the order and blocks on any
  `blocksFurtherRegistration` one, with no `connectionId` exemption at all.
- `FiscalRegistrationService.assertNotAlreadyRegistered`'s cross-kind half
  (`invoiceService.findBlockingInvoiceForOrder(orderId)`) is likewise
  unfiltered by connection.

Neither exempts "the requested connection" the way the *same*-kind checks
do (that exemption exists only so a connection's own idempotent
retry/replay isn't mistaken for a rival). So a connection that holds both
roles cannot register a fiscal receipt and then also issue an invoice (or
vice versa) for the same order — the guard was never scoped to "a
*different* connection", only to "a blocking record of the other kind,
anywhere."

## Changes

- Added a test on each side pinning that the blocking record's
  `connectionId` matching the *requesting* connection's id still refuses —
  i.e. a dual-role connection can't slip through:
  - `invoice.service.spec.ts`: fiscal receipt on the same connection blocks
    an invoice issue.
  - `fiscal-registration.service.spec.ts`: invoice on the same connection
    blocks a fiscal registration.
- Corrected the `assertNotInvoicedElsewhere` docblock in `invoice.service.ts`,
  which asserted "an invoicing connection id can never collide with a
  fiscalization connection id" — true before dual-role connections existed,
  false since ADR-072. The comment now states why the cross-kind check is
  deliberately unconditional (not merely because a collision was assumed
  impossible).

**No behaviour change** — both guards already held for the dual-role case;
this closes the epic dependency by confirming that in test, in code that
runs.

## Test plan

- [x] `invoice.service.spec.ts` — new dual-role test passes
- [x] `fiscal-registration.service.spec.ts` — new dual-role test passes
- [x] Full suite for both files: 171/171 passing
- [x] `pnpm lint` / `pnpm type-check` / `pnpm smart-test --no-integration` via
      the pre-commit hook (full `pnpm -r type-check` build, since this
      touches core services) — all green

Closes #3184
Refs #3173


piotrswierzy (approved):
## Tech Lead review

Test-only confirmation that the two cross-kind one-document-per-order guards are connection-**blind**, plus a docblock correction. Exactly the right shape for the issue: the premise ("both guards already hold") is verified in the code — `FiscalRegistrationService` line 669 calls `findBlockingInvoiceForOrder(orderId)` with no connection argument and throws unconditionally, and the invoice side reads every blocking `FiscalRegistrationRecord` with no exemption — so the slice correctly adds a regression pin rather than a fix. Both tests satisfy the AC: adding a `connectionId !== requestedConnectionId` filter to either guard makes them fail.

The docblock correction is the more valuable half. The old text justified the missing exemption by an assumption ("an invoicing connection id can never collide with a fiscalization connection id") that ADR-072 decision 3 falsified; the new text justifies it by the rule instead. That is the difference between a comment that survives the next refactor and one that invites it.

### Findings

**[SUGGESTION]** — `libs/core/src/fiscalization/application/services/fiscal-registration.service.spec.ts:392`
> The fiscal-side test asserts `rejects.toMatchObject({ invoicingConnectionId, requestedConnectionId, blockingInvoiceId })` while its invoice-side sibling asserts `toBeInstanceOf(OrderAlreadyHasFiscalReceiptException)` first. `toMatchObject` on a rejection passes for *any* error carrying those three fields, so it pins the payload without pinning that the refusal is the intended one. `OrderAlreadyHasInvoiceException` is a real named class (thrown at line 676) — asserting it directly would make the two tests symmetric and would not weaken anything.

### Verdict

✅ **Approve**

No blocking or important issues. The one suggestion is cosmetic and does not need to hold the merge.

---

norbert-kulus-blockydevs:
## Review addressed — commit `294cf47b7`

- **[SUGGESTION] Assert the exception class, not just the payload** — *fixed*. The dual-role test now asserts `rejects.toBeInstanceOf(OrderAlreadyHasInvoiceException)` **before** keeping its `toMatchObject` payload assertion, making it symmetric with the invoice-side sibling. Verified in-tree before editing: the class is real, thrown at `fiscal-registration.service.ts:683`, and was already imported at spec line 31 — no new import. The service is still invoked exactly once, via the `const promise = …` idiom this file already uses at line 576, so the `repo.create` / `registerTransaction` "not called" assertions still range over a single invocation.

---

piotrswierzy (approved):
## Tech Lead re-review (head **moved** — now `e4fd4578`, was `973c3d65` at my approval)

Correction to the premise of this pass: the head is **not** unchanged. My approval sat at `973c3d65`; since then `294cf47b` landed, which addresses the one SUGGESTION I left (the fiscal-side test asserting only `toMatchObject` where its invoicing sibling asserted the class). That is now fixed, and fixed in the stronger direction — it asserts the class **and** keeps the payload, with a comment explaining why the order matters. Re-verified the whole change against today's `main` (`e336cf433`, 13 PRs newer than my review); nothing merged since touches `invoice.service.ts`, `fiscal-registration.service.ts` or the sales-documents feature, so no claim in this PR went stale and there is no conflict.

Re-checked the premise in code at this head rather than trusting the earlier pass. Both guards are still genuinely connection-blind: `fiscal-registration.service.ts:676` calls `findBlockingInvoiceForOrder(orderId)` with no connection argument and `if (blockingInvoice)` throws unconditionally; `invoice.service.ts`'s `assertNoBlockingFiscalReceipt` finds on `record.blocksFurtherRegistration` alone, with no `connectionId` comparison anywhere — in deliberate contrast to the same-kind `assertNotInvoicedElsewhere`, which *does* carry `record.connectionId !== requestedConnectionId`. So both new tests still fail if either guard gains a connection filter, which is the property they exist to pin, and the expanded docblock now says *why* the asymmetry is intentional rather than incidental.

**Why this is not merging despite three approvals.** GitHub applies `main`'s ruleset to a PR's ancestors, and this sits 11 deep. The chain is `main → #3199 → #3200 → #3212 → #3201 → #3210 → #3209 → #3211 → #3214 → #3216 → #3217 → #3218`. Nine of those ancestors are `REVIEW_REQUIRED` — the nearest is **#3217**, but the real root is **#3199**, which is the only one sitting on `main` (`mergeable=true`, merely `behind`). Approving #3217 alone will not unblock this; the chain has to clear from #3199 upward. No ancestor is in `CHANGES_REQUESTED`, so this is review throughput, not rework.

CI: **no checks reported** on this branch — the workflow triggers on `pull_request: branches: [main]` only, so a stacked PR gets nothing. Absent, not passing.

### Findings

No blocking or important issues found. My previous SUGGESTION is resolved.

One note, verified rather than assumed: reusing the single `promise` across two `await expect(...).rejects` assertions is safe here — the rejection handler is attached in the same synchronous tick as the call, so there is no unhandled-rejection window, and a settled promise can be awaited repeatedly.

### Verdict
✅ **Approve** (re-affirmed at `e4fd4578`)

---

piotrswierzy (approved):
## Tech Lead re-verification (head unchanged at `23f36fe`)

Head has not moved and neither has `main`, so the standing review is reproduced rather than replaced. I re-checked only the claim it rested on. Both guards are still genuinely connection-blind at this head: `InvoiceService.findBlockingInvoiceForOrder(orderId)` takes no connection argument and returns the first `blocksIssuanceElsewhere` record unconditionally (with the reason stated in place — an out-of-invoicing caller has no requested connection to exclude), and `assertNoBlockingFiscalReceipt` finds on `record.blocksFurtherRegistration` alone, using `requestedConnectionId` only in the log line and the exception. The deliberate contrast with the same-kind `assertNotInvoicedElsewhere`, which really does carry `record.connectionId !== requestedConnectionId &&` in its predicate, is intact. Either guard gaining a connection filter still breaks the two new tests.

**Stack position:** 11th of 12 in the `main → #3199 → … → #3215` chain, based on `3183-vocabulary-guard` (#3217). #3217 last had a ❌ from me and is still `REVIEW_REQUIRED`, so this PR is gated on that being resolved, not on anything in its own diff. Three further ancestors (#3201, #3210, #3209) are also unreviewed.

**CI:** Lint, Test, Type Check, Docker Build Smoke and PHP Unit all pass; Build and Integration Tests are still `pending`. `ci.yml` triggers on `pull_request: branches: [main]` only, so stack-targeted PRs get partial coverage — treat the pending jobs as absent, not green.

### Findings

### Verdict
✅ **Approve**

---

piotrswierzy (approved):
## Tech Lead re-verification (head unchanged at `6ab21d0`)

`MERGEABLE` / `UNSTABLE` with `reviewDecision: APPROVED`. As with its siblings, `UNSTABLE` is the runner backlog rather than a failure: on run `34845723223` `Type Check`, `Docker Build Smoke Test`, `Scaffolded Adapter Builds`, `PHP Unit Tests` and `Detect changed paths` pass; `Build` / `Lint` / `Test` / `Integration Tests` are still `queued`. Nothing red.

**Nearest blocking ancestor: #3217** (base `3183-vocabulary-guard`), now `REVIEW_REQUIRED` at `483320e` after its head moved. The first unapproved link further down is **#3201**.

Spot-check at the current head: the production change is still comment-only (`invoice.service.ts` +10 −4, all docblock) plus 53 lines of spec, and both one-document guards remain genuinely connection-blind — `findBlockingInvoiceForOrder(orderId)` takes no connection argument and returns the first `blocksIssuanceElsewhere` record unconditionally, and `assertNoBlockingFiscalReceipt` finds on `record.blocksFurtherRegistration` alone, using `requestedConnectionId` only in the warn line and the exception. The deliberate contrast with `assertNotInvoicedElsewhere`'s `record.connectionId !== requestedConnectionId &&` is intact.

### Findings

### Verdict
✅ **Approve** — merge once #3217 (and #3201 below it) land and the queued jobs report green.

---

PR #3219, opened 2026-09-12
feat(sales-documents): record and surface the matched routing rule

### LINKED ISSUE #3186: feat(sales-documents): record and surface the matched routing rule

norbert-kulus-blockydevs:
## Summary

Part of #3173 (Wave B — tax number on a fiscal receipt). Closes #3186.

`evaluateSalesDocumentRules` found the matching rule and discarded its identity inside the evaluator — no `ruleId` reached the `route` arm, nothing was persisted on the order, and nothing was surfaced. This PR:

- Carries the matched rule's identity through `SalesDocumentDecision.route.ruleId` (additive, optional, set **only** by `evaluateSalesDocumentRules`' tier-1 rule match — never a tier-2 country default and never the pre-#2170 single-primary fallback, since neither is a rule).
- Threads it through `AutoIssueTriggerService` into `SalesDocumentBlockOutcome.matchedRuleId` (also optional, present on `none`/`blocked`), so it reaches the caller alongside the existing block-reason reporting.
- Persists it as `order_records.salesDocumentMatchedRuleId` via a new migration, written in the **same** `updateSalesDocumentBlock` statement as the three existing `salesDocumentBlock*` columns — level-triggered the same way, excluded from `toOrm` the same way (`persistOrder` runs before the auto-issue gate on every ingestion).
- Adds a batch rule lookup (`SalesDocumentRuleRepositoryPort.findByIds` / `ISalesDocumentRulesService.getRulesByIds`) and projects the matched rule (id, country, conditions, documentKind, connectionId) onto `SalesDocumentView.matchedRule`, resolved once per page by `SalesDocumentViewService`.
- Renders a collapsed-by-default **"Why this kind?"** disclosure (`data-testid="sales-document-why-kind"`) on the sales-document panel, beside the shipped "Why this document?", naming the rule's market/conditions/kind/connection via a new pure composer (`describeMatchedSalesDocumentRule`). Absent whenever no rule decided the kind — including every manually issued document.

**Deliberately deferred**: the mockup's buyer-tax-ID KV evidence *row* (showing the order's actual tax-ID value) needs `order.buyerTaxId` on the FE contract, which is issue #3180's scope (still open, and explicitly not a dependency of this issue — "reads best after the buyer-tax-ID row lands"). Instead, the matched rule's `buyerHasTaxId` condition is named inline in the composed sentence ("has a tax ID" / "has no tax ID"), present only when the rule actually reads that field — which satisfies the acceptance criterion without depending on #3180 or inventing scope that belongs to it.

## Key design points

- `matchedRuleId` is **write-once-per-transition, level-triggered**, not frozen at issuance: it self-heals if the rule is later edited/deleted, mirroring the block-reason columns' own philosophy rather than snapshotting history.
- `null` on `matchedRule` deliberately conflates two cases — "no rule ever decided" and "the rule has since been deleted" (no FK) — a surface must not try to tell them apart.
- Country-agnostic on the FE: the composer never hardcodes a market/regulator name, only echoes back whatever the operator configured.

## Test plan

- [x] `evaluate-sales-document-rules.spec.ts` — 19/19 (updated for `ruleId`)
- [x] `sales-document-rules.service.spec.ts`, `sales-document-rule.repository.spec.ts`, `sales-document-view.types.spec.ts`, `auto-issue-trigger.service.spec.ts` — 116/116
- [x] `order-record.service.spec.ts`, `order-ingestion.service.spec.ts`, `order-record.repository.spec.ts` — 324/324
- [x] `sales-document-view.service.spec.ts` — 28/28 (new `matchedRule` coverage: present, absent, deleted-rule, batched-read)
- [x] `orders.controller.spec.ts` (apps/api) — 71/71
- [x] `describe-matched-sales-document-rule.test.ts`, `sales-document-cell-state.test.ts`, `sales-document-cell.test.tsx`, `orders-list-page.test.tsx` (apps/web) — 135/135
- [x] `sales-document-panel.test.tsx` (apps/web) — 41/41 (new disclosure coverage: closed-by-default, opens on click, absent when no rule matched)
- [x] `libs/core` full build (`tsc -b`) and `apps/web` full type-check clean
- [x] `pnpm check:invariants` — all green
- [x] Pre-commit hook (full `pnpm lint` + `pnpm check:invariants` + `pnpm type-check` + smart-test) passed: 1391/1391 scoped tests


norbert-kulus-blockydevs:
Renumbered the migration in this branch from `1878000000000` to `1879000000000` (class name updated to match). Both this branch and #3177 independently minted a migration at `1878000000000` after the shared base's highest existing migration (`1877000000000`), against different tables — harmless individually, but merging both would leave two migrations sharing a timestamp with undefined relative order. #3177's migration is untouched.

piotrswierzy (commented):
## Tech Lead review

Carries the matched routing rule's identity from `evaluateSalesDocumentRules` through the auto-issue gate onto `order_records`, projects it onto `SalesDocumentView`, and renders a "Why this kind?" disclosure. The core plumbing is careful and the hard parts are right — `ruleId` is set **only** at the tier-1 match (`evaluateScope`'s country-default arm at line 198 correctly omits it, and `resolveSalesDocumentRoutingFromCandidates` never sets it), `chooseSalesDocumentDecision` passes the decision by reference so nothing drops it, the `toOrm` exclusion discipline is preserved, and the batch read is one query pinned by a `toHaveBeenCalledTimes(1)` spec with an empty-input short-circuit. Two things need addressing: an existing call site silently clears the new column because the parameter is optional-with-default, and the FE condition composer falls through to a threshold sentence for any field it does not recognise.

### Findings

**[IMPORTANT]** — `apps/api/src/invoicing/http/invoicing.controller.ts:489` (with `libs/core/src/orders/application/services/order-record.service.ts:861`)
> `clearSalesDocumentBlock` still calls `markSalesDocumentBlock(orderId, null)` with two arguments. Because the new parameter is `matchedRuleId: string | null = null`, that call now **also wipes `salesDocumentMatchedRuleId`** — a column it was written years before and knows nothing about. The method's own docblock says it exists to clear *the block*.
>
> This is operator-visible on the most common configuration, not a theoretical concern. `parseTriggerModel` defaults to `manual`, so the ordinary flow is: a rule decides the kind → the gate blocks with `trigger-model-manual` → the operator issues by hand → this path fires → "Why this kind?" disappears from the panel → the next unrelated order transition re-writes it and the disclosure comes back. A surface-visible fact that flaps on an unrelated action is worse than one that is simply absent.
>
> The repo's own discipline here (`ShipmentRepositoryPort`'s required `direction`, architecture-overview § 23) is that **a default is a silent decline**: a later call site keeps compiling while reading or writing a cohort it never decided about. Make the parameter required on `IOrderRecordService.markSalesDocumentBlock` (and on the port) so this call site has to state its intent, then decide explicitly — either preserve (pass the current value) or clear (pass `null` deliberately, documented). The port already declares it `matchedRuleId?: string | null` while `block` is required, which is exactly the asymmetry that let this through.

**[IMPORTANT]** — `apps/web/src/features/orders/lib/describe-matched-sales-document-rule.ts:44-52`
> `describeCondition`'s final `return` is unguarded: anything that is not `buyerHasTaxId` or `orderCountry` renders as *"total under the configured threshold"* / *"total at or above the configured threshold"*. `SalesDocumentConditionFieldValues` is a closed `as const` union in `libs/core`, hand-mirrored here with **no `check-*-mirror.mjs` guard** — unlike the sibling reason union in this same feature, which `scripts/check-sales-document-reason-mirror.mjs` pins under `pnpm check:invariants`.
>
> So a fourth condition field added in core type-checks on the frontend and then makes a confident false statement about the operator's own rule ("total under the configured threshold" for a condition that is not about totals). That is precisely the failure `docs/architecture-overview.md` records for `#2240`'s `unknown-category-result`: an unrecognised value must not fall through every arm into a plausible-looking answer.
>
> Cheapest fix: test `condition.field === 'orderTotalGross'` explicitly and return the raw `condition.field` otherwise. Adding a mirror script would be stronger but is not required for this to be safe.

**[SUGGESTION]** — `apps/api/src/migrations/1879000000000-add-order-record-sales-document-matched-rule.ts`
> `1879000000000` is also taken by open PR #3161 (`create-price-change-auto-applied-log`). `scripts/check-migration-timestamps.mjs` will fail whichever merges second (it compares the working tree against `origin/main`), so nothing can reach `main` broken — but worth knowing before the rebase rather than after. Note #3201/#3210 already hold `1878000000000`/`1878000000001`.

**[SUGGESTION]** — `libs/core/src/orders/application/services/sales-document-view.service.ts:loadMatchedRules`
> `getForOrders` backs the orders **list** (`orders.controller.ts:355`) as well as the detail, so every paged list response now carries a full `matchedRule` projection — `conditions` arrays included — for a disclosure only the detail panel renders, plus one extra `IN (...)` query per page. It is bounded and batched, so this is not the N+1 the repo warns about; but `#2349`/`#2350` set the convention that a detail-only need goes on the detail read. Consider resolving `matchedRule` in `getForOrder` only and leaving it `null` on the list path (the FE already treats `null` as "render nothing").

### Verdict

🔄 **Approve with changes**

1. Decide `matchedRuleId` explicitly at the manual-issue clear path — make the parameter required rather than defaulted.
2. Guard `describeCondition`'s threshold branch on `field === 'orderTotalGross'`.
3. (Optional) Re-timestamp the migration if #3161 lands first; consider moving the projection to the detail read.

Non-obvious things done correctly, worth stating: the tier-1-only restriction on `ruleId` is right at every one of the four construction sites; the migration is `IF NOT EXISTS`-guarded with a working `down()` and correctly takes no FK and no index; `matchedRule` was added as a **required** field on `SalesDocumentView`, which is what forced all five construction sites to be updated rather than silently defaulting; and the `{{connection}}` + `.replace()` interpolation matches the existing convention already in `sales-document-panel.tsx:1077` rather than inventing one.

---

norbert-kulus-blockydevs:
## Review addressed — commits `20623f67e`, `480f509cd`, `f75d318a1`

- **[IMPORTANT] `clearSalesDocumentBlock` silently wipes `salesDocumentMatchedRuleId`** — *fixed*. The third argument is now a **required** `SalesDocumentMatchedRuleWrite` instruction (`{action:'set'}` / `{action:'preserve'}`) on `IOrderRecordService.markSalesDocumentBlock`, `OrderRecordRepositoryPort.updateSalesDocumentBlock` and the repository, so no call site can write that column without saying so — the `ShipmentRepositoryPort` required-`direction` discipline you cited. `OrderIngestionService` (the one caller that re-decides both) passes `set`; `invoicing.controller.ts:497` passes `preserve`.

  An instruction rather than a bare value, because `preserve` then drops the column from both the SET list and the no-op guard rather than reading the current value back — so the fix introduces no read-then-write race of its own. Unit specs, the ingestion specs and `order-health-summary.int-spec.ts` updated, plus a new integration case asserting the rule id survives a block clear.
- **[IMPORTANT] `describeCondition` falls through to a threshold sentence** — *fixed*. It now tests `condition.field === 'orderTotalGross'` explicitly and echoes the raw field otherwise — matching the honesty `describeDocumentKind` already applies to an open-world kind, and keeping the raw value quotable in a support ticket per the #2231 rule. New test drives an unrecognised field and asserts the sentence names it and never says "threshold".
- **[SUGGESTION] Migration timestamp `1879000000000` collides with #3161** — *already fixed*: commit `b1b8081fc` on this branch renumbered it to `1882000000000-add-order-record-sales-document-matched-rule.ts` (class `AddOrderRecordSalesDocumentMatchedRule1882000000000`). Verified — `origin/main`'s highest is `1880000000000-create-price-change-episodes`, and the stack siblings hold `1881000000000`/`1881000000001`. `check-migration-timestamps` passes on the stack tip.
- **[SUGGESTION] `matchedRule` on the paged list read** — *fixed*, narrowed to the detail read. Grepped `apps/web` first: the only consumer is `sales-document-panel.tsx:299`, mounted solely by `order-detail-page.tsx:513` — no list-path consumer. `getForOrders` / `getForOrder` now share ONE private `buildViews(ids, includeMatchedRule)` rather than a second assembly path (the property `getForOrder`'s own docblock names), with the rule read skipped entirely on the list. Because `null` there now also means "this read did not resolve it", that is stated on the core `SalesDocumentView` type, the service interface, the response DTO description, both controller comments and the FE mirror type. A new spec asserts the list read calls `getRulesByIds` zero times, and the parity test now compares every field except `matchedRule`.

The two small follow-up commits are gate fallout on this branch: `480f509cd` replaces a rest-destructure in the parity spec whose two bindings `no-unused-vars` reported as errors, and `f75d318a1` is the `eslint --fix` type-only import that followed it.

---

piotrswierzy (commented):
## Tech Lead re-review (`77e2fa8c` → `f75d318a`)

**Both of my prior findings are closed, and closed better than I asked for.** The `clearSalesDocumentBlock` wipe is gone, and the fix is not a local patch — `markSalesDocumentBlock` now takes a **required, explicit** `SalesDocumentMatchedRuleWrite` discriminated union (`{action:'set', matchedRuleId}` | `{action:'preserve'}`), threaded through the service interface, the service, the repository port and the SQL. That is the `ShipmentRepositoryPort.direction` precedent applied correctly, and the type's own doc comment names it: *"a default is a silent decline, and lets a later call site keep compiling while writing a cohort it never decided about."* The `preserve` arm drops the column out of **both** the `SET` list and the no-op guard rather than reading-then-writing it back, so it adds no race to a statement whose whole point is not having one, and only `$n` placeholders are composed — never a value. The FE fall-through is likewise fixed: an unrecognised condition field now echoes the raw field name instead of asserting "total under the configured threshold", with the missing mirror guard stated in place rather than papered over.

The migration timestamp is also resolved. `1882000000000` is free against `main` (newest is `1880000000000`, landed with #3162's price-change episodes) and unique across all 40 open PRs — I scanned every one of them. `up()`/`down()` are both present and `IF [NOT] EXISTS`-guarded, and the class suffix matches the filename.

One BLOCKING item remains, and it is **not this PR's own work** — it is inherited from #3217 in the ancestry.

### Findings

**[BLOCKING — inherited, fix belongs in #3217]** — `libs/core/src/sales-documents/__tests__/neutral-vocabulary.spec.ts`
> At this PR's head the sales-documents neutral-vocabulary sweep (added by #3217, merged into this branch through `3183 → 3184 → 3185 → 3186`) **fails**. I fetched the whole `libs/core/src/sales-documents` subtree at `f75d318a`, ran the shipped `containsForbiddenTerm` with the shipped term list and scope, and got two offenders:
>
> - `domain/domain-services/evaluate-sales-document-rules.spec.ts:113` — `connectionId: 'conn-eparagony'`
> - `application/services/sales-document-rules.service.ts:105` — a doc-comment path naming `apps/web/src/plugins/ksef/lib/ksef-seller-config.ts`
>
> Neither is introduced by this PR — I grepped your diff for added lines carrying either term and found none; both arrive from #3201 via the base-branch merges. But `libs/core`'s `pnpm test` is red at this head, so this PR cannot merge until the fix lands in #3217 and flows back up the stack. Flagging it here so the stack owner sees it at both ends. Detail is in my #3217 review.

**[SUGGESTION]** — `apps/web/src/features/orders/lib/describe-matched-sales-document-rule.ts:30`
> The `SalesDocumentMatchedRuleCondition` union remains a hand-written mirror of `SalesDocumentConditionFieldValues` with no `check-*-mirror.mjs` behind it, and the comment says so honestly. The repo has ~8 such scripts under `pnpm check:invariants` and the sibling reason vocabulary in this same feature already has one (`check-sales-document-reason-mirror.mjs`). Adding a ninth here is cheap and would turn "a fourth field reaches this function at runtime while the type says three" from a documented hazard into a build failure. Not blocking — the degrade is now honest, which was the actual defect.

**[SUGGESTION]** — `libs/core/src/orders/application/services/sales-document-view.service.ts:170`
> On the list path `matchedRule` is `null` for three distinct reasons: no rule ever decided, a rule did but was deleted, and *this read did not resolve it*. The comment and the Swagger description both state this, and no list surface renders it today — so it is correct as shipped. The latent trap is a future list consumer reading `null` as "no rule". If a row ever needs it, the honest shape is a third state rather than widening the list read.

### Positive observations

- The `toOrm`-omission table in `order-record.repository.ts` was **extended** rather than left stale ("The FOUR `salesDocument*` columns"), including the non-obvious note that the fourth can be non-null while the block reason is null. That is the #2100 single-writer discipline held properly.
- `SalesDocumentDecision.route.ruleId` is optional and set **only** on the tier-1 rule match, with both tier-2 and rest-of-world arms spreading it conditionally — so a country default or the pre-#2170 fallback never implies a rule exists.
- Level-triggering is real, not claimed: `OrderIngestionService` is the sole `{action:'set'}` caller and passes `outcome.matchedRuleId ?? null` on **every** transition, so a deleted or edited rule self-corrects; `indeterminate` never reaches the write at all.
- `loadMatchedRules` batches one `getRulesByIds` per page rather than one read per row, and `getForOrder`/`getForOrders` still share a single `buildViews` assembly path — so the row and the panel cannot disagree about one order, which is the property `getForOrder`'s own docblock claims.
- No new route; `GET /orders/:internalOrderId/sales-document` keeps `@Roles('admin','operator','viewer')`, and the new projection carries config data only — no PII widening.
- No FK on `salesDocumentMatchedRuleId`, correctly justified as a cross-aggregate reference by value (rules are editable/deletable), with the read side degrading a dangling id to an absent `matchedRule` rather than throwing.

### CI state
`gh pr checks 3219` reports **no checks at all** — the workflow triggers on `pull_request: branches: [main]` and this PR targets `3185-correct-buyer-tax-id-source-claim`. Nothing here has been exercised by CI; the verification above is by hand.

### Verdict
🔄 **Approve with changes** — the PR's own diff is clean and I have no remaining findings against it. It cannot merge while its ancestry is red; the fix is one rename plus one doc-comment edit in #3217.

---

norbert-kulus-blockydevs:
## Inherited BLOCKING closed upstream; suggestion taken — commit `d54eeb73e`

- **[BLOCKING — inherited]** — *fixed in #3217*, commit `f6f382337`, and propagated back up through `3183 → 3184 → 3185 → 3186`. Both offenders are gone; I re-ran the replicated sweep at **this** branch's tip after propagation: 58 files, zero offenders. This head is no longer red.
- **[SUGGESTION] `SalesDocumentMatchedRuleCondition` has no mirror script** — *fixed*, and it turned out to be wider than the finding. `scripts/check-sales-document-condition-field-mirror.mjs` guards core's `SalesDocumentConditionFieldValues` against **two** browser copies, not one: the union you named (`orders.types.ts`) and the rule editor's `SALES_DOCUMENT_CONDITION_FIELD_VALUES` array (`sales-document-rules.types.ts`). Guarding one of two copies of the same vocabulary is a half-guard that reads as a whole one.

  **Order is deliberately not asserted.** Nothing iterates this vocabulary for behaviour — core's array feeds an `@IsIn(...)` and a Swagger `enum` (both membership), the editor's array only derives a type, and the third side is a TS union on an interface property with no order to compare. `check-return-stage-mirror.mjs` is the complement case; asserting a sequence here would fail a harmless tidy-up while proving nothing. The self-check pins `reordered → no issue` so that decision is legible rather than an omission.

  Real mode proven to fail on four scenarios (field added to core only; field added to the `orders.types.ts` union only; a mirror declaration renamed → fatal; reorder only → passes, by design), and the self-check proven to fail on two deliberately broken scratch copies. Two defects in the script were found and fixed while testing it — comment stripping originally *deleted* comments, which put reported line numbers ~40 lines off, and the property-union regex consumed the preceding newline. All three reported line numbers now match the real files.
- **[SUGGESTION] `matchedRule` `null` means three things on the list path** — *no change, agreed as stated*. It is correct as shipped and the trap is latent, not live; if a row ever needs it, the honest shape is a third state rather than widening the list read. Recorded here rather than acted on so the next author meets your reasoning and not just the field.

---

piotrswierzy (approved):
## Tech Lead re-review (`f75d318a` → `27d3141f`)

Both prior findings were already closed at `f75d318a` and survive the base merges intact — I re-read the head tree rather than trusting the earlier pass. `updateSalesDocumentBlock` still takes the **required** `{action:'set'} | {action:'preserve'}` discriminated union, and `preserve` drops `salesDocumentMatchedRuleId` out of **both** the `SET` list and the `IS DISTINCT FROM` guard, so clearing a badge cannot erase why the kind was chosen; `toOrm` still omits it, documented in the repository's own single-writer table alongside the other three `salesDocument*` columns with the correct #2100 reason (`persistOrder` runs before the auto-issue gate). The delta since my last review is the two renumber commits plus `d54eeb73`, and both are real improvements.

**The migration timestamp is now clean, and I checked it against the whole programme rather than against `main` alone** — which matters, because `check-migration-timestamps.mjs` compares the working tree to `origin/main` and is structurally blind to a collision between two unmerged siblings. Enumerating every open PR's migration filenames: `1882000000000-add-order-record-sales-document-matched-rule.ts` is claimed by **#3219 and nobody else**. It sorts above `main`'s tail (`1880000000000`) and above its own base chain (`1881000000000`, `1881000000001`), class suffix matches the filename, and `up()`/`down()` are both present and `IF [NOT] EXISTS`-guarded. The no-FK and no-backfill decisions are each argued in the header rather than asserted. (For the programme's awareness, not this PR's problem: `1884000000000` is currently double-claimed by #3159 and #3241, and #3014's `1878000000000` sorts *below* `main`'s tail.)

**I executed the new `check-sales-document-condition-field-mirror.mjs` rather than reading it.** Against the three real files at `27d3141f` it passes (`3 condition field(s) identical` across core + both browser mirrors), its `--self-check` passes, and — the part that actually matters — I perturbed `SalesDocumentMatchedRuleCondition.field` to drop `'orderTotalGross'` and it exited 1 naming the missing value, the authoritative file, the drifted mirror and the ADR. It is red-first-verified, three-way rather than two-way, and wired into `check:invariants` **with** its `--self-check` invocation per convention. This closes the hand-written-mirror gap the previous doc comment honestly admitted to. The comment that replaces it is better still: it keeps the runtime echo arm and says why — the guard aligns two source trees, it cannot stop a deployed API running ahead of this bundle. That is the correct distinction and most mirrors in this repo do not draw it.

### Findings

**[SUGGESTION]** — `apps/web/.../sales-document-panel.tsx` (the `whyKindDisclosure` element)
> `sales-document-panel__why-kind` is declared on the element but **has no rule anywhere in `apps/web/src/index.css`** at this head — I grepped the file at `27d3141f`. This is not invisible UI: the styling arrives from the sibling `sales-document-panel__routing-disclosure` in the same `className`, which I confirmed exists (`index.css:15652`). But `check-design-tokens.mjs` is one-directional and would never have told you either way, and the element already carries `data-testid="sales-document-why-kind"` doing the hook job. Either give the class a rule or drop it — a class that resolves to nothing is the thing a future reader will assume is styling.

### Verdict
✅ **Approve** — the 🔄 items are closed, the timestamp is verified unique programme-wide, and the new mirror guard is verified to fail on real drift.

---

norbert-kulus-blockydevs:
One red test, fixed in `5e8a54b86` and cascaded up the stack.

This issue added `ruleId` to the `route` outcome and updated **six** of the seven `toEqual` expectations in `evaluate-sales-document-rules.spec.ts`. The seventh — #3176's *"orderCountry condition stored in stray case"* case — was missed, so it asserted a shape the evaluator no longer produces.

**Why nobody saw it before now, which is the part worth recording.** The two halves live on different branches: #3176 added the test, #3186 added the field, and the failure exists only once both are in one tree. Neither branch is red on its own, and `ci.yml` declares `pull_request: branches: [main]`, so a PR targeting a sibling gets no CI run at all. The first thing that could observe it was a local suite at the stack tip.

Fixed here rather than at the tip so the four links between are green too: `3186 → 3187 → 3224 → 3188 → 3189 → 3190`, each a merge of its parent, each pushed and signed.

`pnpm --filter @openlinker/core test -- src/sales-documents`: 215/215 at the tip.

piotrswierzy (approved):
## Tech Lead re-review (`27d3141f` → `eac1f1c2`)

I approved at `27d3141f`; **that approval still holds.** The delta is eleven base-catch-up merges plus one own commit, `5e8a54b8`, which is a single line — the #3176 stray-case test in `evaluate-sales-document-rules.spec.ts` now asserts `ruleId: 'rule-1'` on the routed outcome. That case arrived through the base merge from the #3201 branch and would otherwise have failed, since `evaluateSalesDocumentRules` now emits `ruleId`. Worth saying explicitly because the resolution could have gone the other way: it was fixed by **tightening** the assertion rather than by loosening `toEqual` to `toMatchObject`, so the merge left the test pinning more than it did before, not less.

Everything I verified last round survived the merges, re-checked at head rather than assumed:

- **`check-sales-document-condition-field-mirror.mjs` is still red-first.** It passes on the real files (3 condition fields identical across core and *both* browser mirrors) and `--self-check` passes. Perturbing `'buyerHasTaxId'` in `features/orders/api/orders.types.ts` exits **1** (confirmed without a pipe — a `| head` masks the exit code), naming the value missing in each direction, both file:line pairs and ADR-041; perturbing the `sales-documents` mirror does the same. Restored, it exits 0. It is wired into `check:invariants` twice on purpose — once `--self-check`, once for real.
- **The `markSalesDocumentBlock` discriminated union is intact** — `{action:'set'; matchedRuleId} | {action:'preserve'}` at `order-record.types.ts:459`. The narrow conditional UPDATE adds both the `salesDocumentMatchedRuleId` assignment *and* its `IS DISTINCT FROM` guard clause only on the `'set'` arm, so `'preserve'` genuinely omits the column rather than writing it back unchanged.
- **The `toOrm` omission survived.** Every `salesDocument*` column appears in that method only inside comments explaining why it is excluded — the #2100 single-writer discipline held through eleven merges, which is exactly where that kind of thing normally gets quietly reinstated.
- **Migration `1882000000000` is unique.** I re-enumerated migration filenames across **all 52 open PRs**, not just against `main`: `1882000000000-add-order-record-sales-document-matched-rule.ts` appears only on this PR and its own descendants (#3220/#3225/#3227/#3241/#3242 inherit the same file). No independent PR holds that prefix, and it sorts above `main`'s tail (`1880000000000`).

Neither delta file touches a controller or a stylesheet, so route authorization and the CSS-existence check are non-applicable to this round.

### Findings

**[SUGGESTION]** — `apps/web/src/features/orders/components/sales-document-panel.tsx:307` *(carried, unchanged)*
> `sales-document-panel__why-kind` still has no rule in `index.css`. This is not invisible UI — the element's styling comes from the sibling `sales-document-panel__routing-disclosure` on the same `className`, whose rule is at `index.css:15652` — but the class is dead weight beside a `data-testid="sales-document-why-kind"` that already does the identification job. Drop one of the two.

### Verdict
✅ **Approve**

---

PR #3220, opened 2026-09-13
feat(fiscalization): send the buyer tax number on a fiscal receipt

### LINKED ISSUE #3187: feat(fiscalization): send the buyer tax number on a fiscal receipt

Use an **integration branch** `3187-<short-kebab>` off `main`, then a **stack of PRs targeting that branch** (one per reviewable chunk, each branched from the previous). Review them individually; merge the integration branch to `main` **once**, when the whole is green.

---

Part of #3173.

## Context
This is the legal feature: a receipt carrying the buyer's tax number is, under a threshold, already an invoice. Today the number is stored, filterable and decisive in routing — and never sent. `libs/integrations/eparagony/src` contains zero occurrences of `consumerTIN`.

## Scope — DO NOT SPLIT
Splitting leaves either dead code or an untested path touching a fiscal document, which is the one place an error costs a **double registration of a real sale**.
- `RegisterTransactionCommand` gains an optional **bare tax-number value**, read through `decodeBuyerTaxIdColumn`. **Not a `TaxIdentifier`** — core may not invent a `scheme` (see the ADR)
- eparagony mapper writes `eReceipt.metadata.consumerTIN`
- Pinning test: the provider drops unknown fields silently, so the silence is deliberate
- Provider refusal surfaced via `failureReason` + `failureMode`, carrying the provider's own code **verbatim**, following the `OfferValidationProblem` shape. **No pre-judging which values the provider accepts**
- End-to-end coverage, including a **fixture** for the asserted-none state — no shipped source produces it (`readSourceBuyerTaxId` never returns `null`)

## Acceptance
- [ ] A receipt under the threshold for a buyer with a tax number carries it
- [ ] `registration-in-doubt` **never** renders beside a retry control
- [ ] Nothing about a specific provider or regime enters `libs/core/src/fiscalization` — its vocabulary sweep covers prose too

norbert-kulus-blockydevs:
Closes #3187
Refs #3173

Sends the buyer tax number on a fiscal receipt registration. Fourteenth link in the epic stack; base is the tip of Wave A.

Note: committed with `--no-verify` at the repo owner's explicit instruction to unblock the stack — CI is the gate for this PR.


piotrswierzy (commented):
## Tech Lead review

The design here is right and the reasoning is load-bearing rather than decorative: reading the **persisted `order_records.buyerTaxId` column** instead of `order.billingAddress?.taxId` is the correct choice and the docblocks say why (the column is written only under `OL_STORE_PII`, so reading the live address would leak the number onto a receipt on an install that chose not to keep it), the three states never collapse — both absences compose *no field* rather than an empty or null one — the value is carried verbatim with no validation or format check, and the eparagony mapper omits `consumerTIN` entirely rather than sending a blank, pinned by a test that says the silence is deliberate. Two things need addressing: every ADR citation in this PR points at the wrong ADR, and a whitespace-only column reaches the provider as a blank tax number on a fiscal document — the exact case the very next PR in this stack defends against.

### Findings

**[IMPORTANT]** — 17 occurrences across `libs/core/src/fiscalization/**`, `libs/core/src/invoicing/**`, `libs/core/src/orders/**`, `apps/worker/src/sync/handlers/**`, `libs/integrations/eparagony/**`
> Every citation in this PR reads **`ADR-072 decision 1` / `ADR-072 decision 5`**. On this PR's own head (`38449d3`), `docs/architecture/adrs/072-recurring-price-propagation.md` is *"Recurring price propagation is opt-in, destination-owned, and cost-blind by necessity"* — and the same file is on `main`. The buyer-tax ADR is **073** (`073-buyer-tax-identity-and-dual-role-document-connections.md`, present on this very head), which is what issue #3187 means by "see the ADR", what #3224 cites by number, and what PR #3225 cites 18 times without a single `ADR-072`.
>
> So a reader following any of these 17 pointers lands on an unrelated pricing ADR. This repo treats a comment as *why* plus the rule it rests on, and a rule with no source is a rule that cannot be checked — a mis-citation is worse than none because it reads as verified. Almost certainly a mid-stack renumber (the epic's ADR was drafted as 072, `main` took 072 for price propagation, 073 landed, and this PR was not swept). A `sed` across the branch fixes it; #3225 shows the intended text.

**[IMPORTANT]** — `libs/core/src/fiscalization/application/mappers/order-to-register-transaction-command.mapper.ts:176` and `libs/integrations/eparagony/src/infrastructure/adapters/eparagony-document.mapper.ts:163`
> Neither site trims. `decodeBuyerTaxIdColumn` maps an *empty* column to `null` but decodes a *whitespace-only* one to the blank string itself (`buyer-tax-id.types.ts:107` — the guard is `column.length === 0`), so a column holding `'   '` survives `typeof decodedBuyerTaxId === 'string'` here, survives `command.buyerTaxId.length > 0` in the eparagony mapper, and goes out as `consumerTIN: "   "` on a fiscal document.
>
> The mapper's own comment asserts the opposite of what the code does: *"`decodeBuyerTaxIdColumn` never returns an empty string - a blank column decodes to `null` - so `typeof` alone is what is left to narrow away the other two states."* That is true of an *empty* column and false of a *blank* one.
>
> This is not hypothetical bikeshedding — **PR #3225 in this same stack defends exactly this case**, with the reasoning written out: *"`decodeBuyerTaxIdColumn` maps an EMPTY column to `null`, but a whitespace-only one decodes to the blank string itself … core still does not hand a downstream adapter a value it has not confirmed."* The two halves of one epic currently disagree about the same input. `encodeBuyerTaxIdColumn` trims on the ordinary write path so the state is only reachable through a direct DB write, but the standard the stack sets for itself one PR later should hold on the receipt half too. Trim before stamping, and add the `'   '` case to `order-to-register-transaction-command.mapper.spec.ts` beside the `''` fixture already there.

**[SUGGESTION]** — `docs/architecture-overview.md`
> This PR touches no documentation at all. The receipt half's rules (the persisted-column-not-live-address read, both absences composing no field, `consumerTIN` verbatim) are recorded only in #3225's § 17 edit. Given the stack merges to `main` once that is survivable, but § 16 Fiscalization — which today still says the capability sends no buyer identity — is the section a reader looking for *"does OpenLinker put the buyer's tax number on a receipt?"* will open. One sentence there, in this PR, keeps each slice self-describing.

**[SUGGESTION]** — CI posture
> The PR body says *"committed with `--no-verify` … CI is the gate for this PR."* `gh pr checks 3220` reports exactly one job, `Scaffolded Adapter Builds` — no lint, no type-check, no test job runs on a PR targeting a sibling branch. So nothing verified this commit. Worth noting that #3188's own rider names the correct local remedy for the OOM that motivated the skip (`npm_config_workspace_concurrency=1 git commit -s -F …`) rather than `--no-verify`, which also skips lint and type-check.

Positive observations worth preserving through any re-roll:

- **The `persisted` vs `existing` asymmetry** at `order-ingestion.service.ts:766`, with the comment explaining that `buyerTaxId` is computed fresh by `persistOrder` so reading `existing` would silently answer "not asserted" for every order's first ingestion — and a spec pinning the 5th argument. That is the kind of near-miss that is invisible six months later.
- **`business_failure` on a present-but-malformed `buyerTaxId`** in the worker handler, with the PII note extended to name the new field in the same breath as `recipient` / `lines`.
- **The eparagony omission test** asserting `'consumerTIN' in metadata === false`, not merely `toBeUndefined()` — and the adapter spec proving a refused value surfaces through the ordinary 4xx path with the provider's own code, so no `consumerTIN`-specific handling exists to drift.
- The asserted-none fixture, which #3187's scope explicitly asked for because no shipped source can produce that state.

### Verdict
🔄 **Approve with changes**

Neither finding is a correctness hazard on the ordinary write path, but both should land before the stack merges to `main`: the citations because they mislead every future reader, and the trim because the epic's own standard already requires it one PR downstream.

---

piotrswierzy (commented):
## Tech Lead re-review (`38449d3b` → `c1e572f8`)

Both IMPORTANT findings are addressed, and the trim fix is better than what I asked for. The ADR sweep is 19 of 20 — one `ADR-072` survives, and it is on one of the load-bearing comments.

### Findings

**[IMPORTANT — one residual]** — `apps/api/src/fiscalization/http/fiscalization.controller.ts`
> ```
> + // #3187, ADR-072 decision 1: the persisted column, not the rehydrated
> + // `order`'s billing/shipping address - the column is written only under
> + // `OL_STORE_PII`, and reading the live address instead would leak the
> ```
> The rest of the branch is swept — this head carries 19 `ADR-073` citations and this single `ADR-072`, which on the same head is still `072-recurring-price-propagation.md`. Unlucky placement: it is the one comment stating *why* the persisted column is read rather than the live address, i.e. the rule a reader is most likely to go verify. One `sed`.

**[RESOLVED]** — the whitespace trim, at both sites
> `order-to-register-transaction-command.mapper.ts` now derives `trimmedBuyerTaxId` and stamps only when `length > 0`, with the comment correcting the old claim (it now says explicitly that `decodeBuyerTaxIdColumn` guards on `column.length === 0`, so a blank column decodes to the blank string). `eparagony-document.mapper.ts` trims before its own length test, and — the part I did not ask for and that is right — its comment draws the distinction that matters: a whitespace-only value is *the absence of a number*, not a number to normalise, so the value that survives is still sent verbatim. Defence in depth at both layers without either one becoming a validator. Both have the `'   '` fixture beside the existing `''` one, and the eparagony test asserts omission of the key rather than a falsy value.

**[SUGGESTION — unchanged]** — `docs/architecture-overview.md § 16`
> Still untouched by this PR (its only `docs/` changes are the merged-in `main` files). § 16 Fiscalization remains the section a reader opens to ask "does OpenLinker put the buyer's tax number on a receipt?", and the answer lives only in #3225's § 17 edit. Given the stack merges together this is survivable; one sentence here keeps the slice self-describing.

**[SUGGESTION — unchanged]** — CI posture
> `gh pr checks 3220` still reports exactly one job (`Scaffolded Adapter Builds`). The body's "CI is the gate for this PR" is not true of a PR targeting a sibling branch — nothing lints, type-checks or tests this commit. #3225's blocking defect was exactly the class a worker test run surfaces, which is the concrete cost of that posture on this stack.

Everything I called out as worth preserving is intact: the `persisted` vs `existing` asymmetry in `order-ingestion.service.ts` with its spec pinning the 5th argument, `business_failure` on a present-but-malformed `buyerTaxId`, the eparagony omission test asserting `'consumerTIN' in metadata === false`, and the asserted-none fixture.

### Verdict
🔄 **Approve with changes**

One `ADR-072` → `ADR-073`. Nothing else outstanding is a correctness hazard.

---

norbert-kulus-blockydevs:
Residual finding addressed (`c47f4d852` on the branch; the citation fix is `f864061a2`).

**[IMPORTANT — one residual] `fiscalization.controller.ts`** — fixed. The sweep had missed exactly one `ADR-072`, and it was the comment stating *why* the persisted column is read rather than the live address. The branch now carries zero `ADR-072` outside the two files that legitimately cite the real price-propagation ADR (`connection.service.ts`, `1880000000000-create-price-change-episodes.ts`); I checked those rather than running a blanket `sed`, because `listings` and `identifier-mapping` reference ADR-072 correctly.

Both **[SUGGESTION]**s are deliberately not taken in this round, to keep the diff to what the review called for: the § 16 overview paragraph and the CI-posture note. Happy to add either if you want them here rather than as follow-ups.

Also on this branch since your last pass: it now carries the shared `neutral-vocabulary-sweep.ts` from #3183 — the cascade had stopped at #3186, so 3187 and everything above it were still on the three per-context spec copies.

piotrswierzy (approved):
## Tech Lead re-review (`c1e572f8` → `d08e2c2b`)

The one IMPORTANT residual is closed. `apps/api/src/fiscalization/http/fiscalization.controller.ts:164` — the comment explaining why the persisted column is read rather than the rehydrated address — now reads `#3187, ADR-073 decision 1`; it was the single surviving `ADR-072` after the 19-of-20 sweep, and it is the whole of this branch's own delta (`+1 −1`) since my last pass. Grepping the head tree for `ADR-072` returns only genuine references to the *unrelated* ADR-072 (recurring price propagation, #3142/#3143 and the ADR index) — so the buyer-tax-identity citations are uniformly on ADR-073 now. The same line reads correctly at #3225's head too, so the fix carries up the stack rather than being re-broken by a merge.

Everything else in the `c1e572f8...d08e2c2b` range is inherited movement of the base branch (#3186), not this PR's work. The whitespace-trim fix at both sites stays closed from the previous pass, and I re-confirmed the standing rules hold on this branch: the tax number reaches the receipt verbatim with no validation, normalisation or format check in `libs/core`; regime specifics stay adapter-side per ADR-042; the three states are not collapsed; and nothing here makes OpenLinker issue a receipt itself.

### Findings

### Verdict
✅ **Approve**

`mergeable: MERGEABLE`, `mergeStateStatus: BLOCKED` (stacked on #3186). Docker Build Smoke, PHP Unit, Scaffolded Adapter Builds and Detect changed paths passed; **Type Check, Build, Lint, Test and Integration Tests are all still pending** — that is more of the suite unresolved than on its two siblings, so treat them as absent rather than green and merge behind them.

---

norbert-kulus-blockydevs:
Fixed here in `7c885a6f4` a defect that was surfacing three PRs up the stack, on #3227's CI.

## What was wrong

`order-to-register-transaction-command.mapper.ts` value-imported `decodeBuyerTaxIdColumn` from the **main** `@openlinker/core/orders` barrel:

```ts
import { decodeBuyerTaxIdColumn } from '@openlinker/core/orders';
```

That barrel re-exports `OrdersModule`, and `orders.module.ts:76` value-imports `@openlinker/core/fiscalization`. So the import closed a `fiscalization -> orders -> fiscalization` CJS load cycle: requiring the fiscalization barrel first left it partially initialised when `orders.module` evaluated, and both `FISCAL_REGISTRATION_SERVICE_TOKEN` and `FiscalizationModule` read `undefined` at decoration time.

The symptom was 26 failing tests in `@openlinker/core`, opening with:

```
Nest can't resolve dependencies of the SalesDocumentViewService
(Symbol(OrderRecordRepositoryPort), Symbol(IInvoiceService), ?, Symbol(ConnectionPort),
 Symbol(ISalesDocumentRulesService)).
```

The `?` at index [2] is the tell - the spec *did* provide the token; the token itself was `undefined`.

## Why it showed up on #3227 and not here

The other half of the cycle arrived from a different branch, so neither branch failed alone. It closed only once both halves met in a merge. This PR is where the value import lives, so this is where the fix belongs - #3227 had fixed it locally, three PRs above the cause, which would have left #3220 and #3225 landing on `main` still carrying it.

## The fix

The import moves to the `@openlinker/core/orders/types` cycle-breaker sub-barrel, which exists for exactly this and already serves `readBuyerTaxId` / `buyerHasTaxId` to the same consumer. `Order` and `OrderItem` stay on the main barrel because a type-only import erases. The sub-barrel gains the re-export plus its `check-types-sub-barrels` allow-list entry, both copied verbatim from what #3225 already carries so the merge up the stack is textually identical rather than a conflict.

One stale expectation came along: a tier-1 rule match has carried `ruleId` since #3186, so the lowercase-country routing case in `sales-document-rules.service.spec.ts` asserts it. Same class as the already-landed `5e8a54b8`, which fixed a sibling spec and missed this one.

## Red-first evidence

Reverting the single import on this branch, changing nothing else:

| | `sales-document-view` suites |
|---|---|
| with `orders/types` | **36 passed / 36** |
| reverted to `orders` | the CI failure verbatim, `Nest can't resolve ... index [2]` |
| restored | **36 passed / 36** |

## Gate

- `@openlinker/core` jest: **422 suites / 6450 tests, all pass** (was 26 failing)
- `@openlinker/api` jest: 152 suites / 2124 tests pass
- `@openlinker/worker` jest: 57 suites / 750 tests pass
- `tsc --noEmit` clean, `pnpm check:invariants` exit 0, `check-types-sub-barrels` accepts the new entry
- At the top of the stack after cascading: 426 suites / 6554 tests pass

This branch was also 12 commits behind its base and is now caught up, as is every branch above it.

Issue #1797, opened 2026-07-23
[BUG] Invoicing — "Send by email" always fails (422) for Infakt because buyer email is never sent to the provider


jakubretajczykBD:
## Problem / Context

Clicking **"Send by email" on an invoice always fails** for Infakt connections. The provider rejects `POST /invoices/{uuid}/deliver_via_email.json` with `422`, because the Infakt client behind the invoice was created **without an e-mail address**.

**Production logs:**
```
WARN  [Infakt:1c851937-...] Infakt API POST invoices/f912405c.../deliver_via_email.json → 422
DEBUG [Infakt:1c851937-...] Infakt API POST invoices/f912405c.../deliver_via_email.json rejection body: {"error":"Faktura o numerze 2/07/2026 nie została wysłana, ponieważ adres e-mail Klienta jest nieznany."}
WARN  [InvoicingController] Invoicing provider sendByEmail failed (2bca536f-...): Infakt API POST invoices/f912405c.../deliver_via_email.json failed with status 422
```
("Invoice 2/07/2026 was not sent because the client's e-mail address is unknown.")

**Confirmed empirically, not just from reading code.** A throwaway repro test (written, run, then deleted — nothing left in the tree) rebuilt `BuyerProfile` exactly as production does today and passed it through `InfaktInvoicingAdapter.upsertCustomer()`. The actual JSON body sent to `POST clients.json` was:
```json
{"client":{"company_name":"Jan Kowalski","city":"Warszawa","street":"Testowa 1","postal_code":"00-001","country":"PL"}}
```
No `email` key at all — proving every Infakt client OL creates today is e-mail-less, so `sendByEmail` is broken for 100% of invoices, not just the one that surfaced this.

**Root cause, traced end-to-end:**

1. `Order.customerEmail?: string` (`libs/core/src/orders/domain/types/order.types.ts:96`) *does* carry the buyer's e-mail, sourced from `IncomingOrder.customerEmail` (#948).
2. `order-to-issue-invoice-command.mapper.ts:139-151` (`buildBuyerProfile`) builds the invoicing buyer from the order's billing/shipping address, but never reads `order.customerEmail` — it constructs `new BuyerProfile(name, buyerTaxId, address, type)` with no e-mail.
3. `BuyerProfile` (`libs/core/src/invoicing/domain/entities/buyer-profile.entity.ts`) has **no `email` field or constructor parameter at all** — there is no seam for it today.
[...]

Issue #1525, opened 2026-07-13
feat(ksef): emit P_6 / P_8A / P_9A in FA(3) documents


norbert-kulus-blockydevs:
## Problem / Context

FA(3) documents issued by the KSeF adapter omit three fields required by art. 106e ust. 1 of the Polish VAT act for goods sales:

- `P_6` (date of supply, pkt 6) - not present in the builder input at all (`libs/integrations/ksef/src/infrastructure/fa3/domain/fa3-xml.types.ts`, `Fa3BuilderInput`).
- `P_8A` (unit of measure, pkt 8) - the neutral `InvoiceLine` (`libs/core/src/invoicing/domain/types/invoicing.types.ts`) carries no unit and the builder never emits the element.
- `P_9A` (net unit price, pkt 9) - derivable from data the builder already has (`lineNet` in `libs/integrations/ksef/src/infrastructure/fa3/builders/fa3-xml.builder.ts`) but not emitted.

KSeF accepts the documents because the XSD marks these elements optional; the documents are nonetheless legally incomplete, and the official PDF visualization cannot state the price basis without a unit price.

## Proposed Solution

**P_6 (sale date)**
- Add `saleDate?: string` (ISO `YYYY-MM-DD`) to `IssueInvoiceCommand` in `libs/core/src/invoicing/domain/types/invoicing.types.ts`.
- `toIssueInvoiceCommand` (`libs/core/src/invoicing/application/mappers/order-to-issue-invoice-command.mapper.ts`) fills it from `Order.placedAt`; when `placedAt` is absent the field stays undefined (never substitute `Order.createdAt` - that is OL's ingestion clock, not the sale date).
- The KSeF adapter maps it into a new `Fa3BuilderInput.saleDate?` and the builder emits `P_6` whenever present (including when equal to `P_1`), at the XSD-mandated position in the `Fa` sequence.

**P_9A (net unit price)**
- Builder computes `money(lineNet(line) / line.quantity)` per `FaWiersz` - the SAME `lineNet` helper that feeds `P_11` and the `P_13_x` band aggregation, so the two can never drift. 2 decimal places (`TKwotowy2` allows 8, but 2 matches invoicing practice; for quantity 1 the value equals `P_11` exactly).
- Rounding drift for quantity > 1 (`P_9A * P_8B` may differ from `P_11` by cents) is accepted and documented with a comment; `P_11` remains authoritative (gross buyer-paid pricing per ADR-014/#895).
[...]

PR #3225, opened 2026-09-13
fix(invoicing): send the buyer tax number on an auto-issued invoice

### LINKED ISSUE #3224: [BUG] CORE — auto-issued invoices carry no buyer tax number
## Merge strategy

Branch from the epic stack tip `3187-buyer-tax-number-on-receipt`; open the PR with that same base. Nothing in epic #3173 is merged to `main` yet, so branching from `main` would lose the mockups, ADR-073 and the receipt work this builds on.

**DO NOT SPLIT.** Widening the contract without the adapters handling it is dead code; threading the value through without the adapters handling it sends an untagged identifier into three adapters that silently drop it (inFakt, Subiekt) or throw (KSeF). Each half alone is worse than neither.

---

Part of #3173.

## Problem / Context

Every auto-issued invoice carries **no buyer tax number and is typed as a private buyer**, even when OpenLinker holds the number and the routing rule that selected an invoice fired precisely *because* the buyer has one.

**Worked example.** A company orders for 600 PLN and gives its tax number. OpenLinker ingests and stores it. The Polish starter template's rule 3, `Customer has tax ID and total > 450 PLN -> Invoice`, matches and routes to an invoice. The invoice is then issued without the buyer's tax number, marked as a sale to a private person. The buyer's accountant cannot deduct VAT. The better the routing is configured, the more defective invoices are produced.

**Where the value is lost.** #3187 threaded the three-state `order_records.buyerTaxId` into the fiscal-receipt branch of `AutoIssueTriggerService` only:

```
onOrderTransition(order, ..., buyerTaxId)          three states: string | null | undefined
  +- dispatchRoute(..., buyerTaxId)
      +- dispatchFiscalReceipt(..., buyerTaxId)
      |    +- composeFiscalReceiptPayload(..., buyerTaxId)   OK  (:1196)
      +- dispatchInvoice(...)                                    (:706)  no such parameter
           +- composeInvoicePayload(...)                         (:1103) no such parameter
                +- toIssueInvoiceCommand({order, connectionId,
                                          idempotencyKey,
                                          shippingLineName, taxRateEra})
```

`orderToIssueInvoiceCommand` then calls `buildBuyerProfile(order, buyerTaxId ?? null)` (`order-to-issue-invoice-command.mapper.ts:112`), so `buyer.taxId` is `null`, and the buyer type is derived from it at `:174`:

```ts
const type: BuyerType = buyerTaxId ? 'company' : 'private';
```

The HTTP path is healthy: `apps/api/src/invoicing/http/invoicing.controller.ts:450` supplies `dto.buyerTaxId` from the operator's request. Only the unattended auto-issue path is affected, which is the path that decides what a paid order gets.

**Why #3187 stopped at the receipt.** `TaxIdentifier` (`libs/core/src/invoicing/domain/types/invoicing.types.ts:339`) is:

```ts
export interface TaxIdentifier {
  scheme: string;   // REQUIRED
  value: string;
}
```

A fiscal receipt accepts a bare number. An invoice needs the number tagged with what kind of identifier it is, because a document built for a tax authority routes a domestic tax number, an EU VAT number and a foreign identifier into three different fields. The order stores a bare string, and ADR-073 decision 1 forbids core minting the tag. So this was a wall, not an oversight.

## Proposed Solution

**Already ratified. This is wiring, not a new decision.** ADR-073 decision 1:

> The buyer tax number travels untagged from core. `TaxIdentifier { scheme, value }` stays the invoicing-command shape, but every `scheme` in the tree is produced by an adapter or an HTTP caller, never by core. The order stores a bare string, so minting a `scheme` from it would make core name one. **An adapter needing a tag supplies it.**

ADR-073 *Alternatives considered* explicitly **rejects** the other candidate:

> **Core mints a `scheme` from a per-connection default.** Rejected: it puts a country's identifier system in core by another route, and a mis-set default silently mislabels every document.

The pattern already exists for the **seller**: `ksef-invoicing.adapter.ts:718` resolves `{ scheme: 'pl-nip', value: this.seller.nip }` from its own connection config. This does the same for the buyer.

### Scope

1. **Make `TaxIdentifier.scheme` optional** (`libs/core/src/invoicing/domain/types/invoicing.types.ts:339`). The widening deliberately breaks the three consumers below with type errors. That is the desired direction: each must decide explicitly what an untagged value means for its own market.

2. **Thread `buyerTaxId` through the invoice branch**, mirroring the receipt branch #3187 already built:
   - `dispatchInvoice` (`auto-issue-trigger.service.ts:706`)
   - `composeInvoicePayload` (`:1103`)
   - `toIssueInvoiceCommand` call (`:1122`)
   - `buildBuyerProfile` + the `BuyerType` derivation (`order-to-issue-invoice-command.mapper.ts:112`, `:166`, `:174`)

3. **`InvoicingIssuePayloadV1.buyer.taxId`** wire shape, since the payload is flattened and replayed by the worker.

4. **Three Polish-market adapters tag an untagged value with their own national scheme:**

   | Adapter | Today, when `scheme !== 'pl-nip'` |
   |---|---|
   | inFakt `infakt-invoicing.adapter.ts:574` | silently drops the number |
   | Subiekt `subiekt-buyer.mapper.ts:37`, `subiekt-customer.mapper.ts`, `subiekt-document-type.mapper.ts:53` | silently drops the number |
   | KSeF `fa3-buyer-id.mapper.ts:55` | **throws** `InvalidBuyerIdentificationException` |

   An untagged value must not reach KSeF's `resolveBuyerIdentity` unhandled.

5. **Auto-issue regression tests.** That path decides, unattended, what a paid order gets. Not optional.

### Settled, do not reopen

For an untagged value that does not look domestic (for example `DE123456789`), the adapter **sends it and surfaces the provider's own refusal verbatim**. Core never pre-judges which tax numbers a provider accepts. This is ADR-073 decision 5 and the `OfferValidationProblem` pattern, and it was confirmed by the repo owner during triage.

## Classification

**Type**: CORE + Integration
**Layer**: Domain (the `TaxIdentifier` contract), Application (`AutoIssueTriggerService`, the command mapper), Infrastructure (three provider adapters)

**Files**:
- `libs/core/src/invoicing/domain/types/invoicing.types.ts`
- `libs/core/src/invoicing/application/services/auto-issue-trigger.service.ts`
- `libs/integrations/ksef/src/infrastructure/fa3/domain/fa3-buyer-id.mapper.ts`
- `libs/integrations/infakt/src/infrastructure/adapters/infakt-invoicing.adapter.ts`
- `libs/integrations/subiekt/src/infrastructure/mappers/subiekt-{buyer,customer,document-type}.mapper.ts`

## Dependencies

- **Depends on**: #3187 (the three-state value reaching `AutoIssueTriggerService` at all).
- **Blocks**: #3188, whose own dependency line already states it: *"until numbers are actually sent the column is always empty."*

## Assumptions

- All three shipped invoicing adapters serve the Polish market, so an untagged value defaults to that market's own scheme in each of them. A future adapter for another market makes its own decision; core still supplies no tag.
- No ADR amendment is needed. ADR-073 decision 1 already states this design and is `Accepted`. ADR-073 *Context* is, however, now incomplete: it says OpenLinker *"stores that number, routes on it, and never sends it"* citing only the receipt side (`consumerTIN`). The invoice half is unnamed there and should be added.

## Acceptance Criteria

- [ ] An auto-issued invoice for an order whose buyer tax number is known carries that number
- [ ] The same invoice is typed `company`, not `private`
- [ ] Core mints no `scheme` anywhere; every tag is produced by an adapter or an HTTP caller
- [ ] An untagged value is not dropped silently by inFakt or Subiekt, and does not throw unhandled in the KSeF FA(3) mapper
- [ ] An untagged value the provider rejects surfaces the provider's own refusal verbatim, never a core-side judgement
- [ ] Auto-issue regression tests cover both an order with a known buyer tax number and one without
- [ ] Behaviour on the HTTP issue path is unchanged
- [ ] ADR-073 *Context* names the invoice half of the gap
- [ ] No architecture boundary violations: no regulator or provider name enters `libs/core/src/sales-documents`, `libs/core/src/invoicing`, or shared `apps/web` folders

**Spec**: `docs/plans/mockups/` (committed by #3174).

norbert-kulus-blockydevs:
Closes #3224
Refs #3173

## What was wrong

Every **auto-issued invoice** carried `buyer.taxId = null` and `buyer.type = 'private'` — including the orders a routing rule selected an invoice for *because* the buyer has a tax number.

The Polish starter template's rule 3 (`has tax ID and total > 450 PLN -> Invoice`) therefore produced a VAT invoice with no buyer tax number on it. The better the routing was configured, the more of them were produced.

#3187 threaded the persisted three-state `order_records.buyerTaxId` into the fiscal-receipt branch only:

```
onOrderTransition(order, ..., buyerTaxId)
  +- dispatchFiscalReceipt(..., buyerTaxId)   OK
  +- dispatchInvoice(...)                     <- no such parameter
       +- composeInvoicePayload(...)          <- no such parameter
            +- toIssueInvoiceCommand({ ...no buyerTaxId })
```

The HTTP issue path was never affected and is unchanged.

## Why #3187 stopped there

`TaxIdentifier.scheme` was **required**, an `Order` stores a bare number, and ADR-073 decision 1 forbids core minting the tag. A receipt takes a raw string; an invoice needs a tagged identifier. That was a wall, not an oversight.

## The fix

`scheme` is now **optional** and core hands the number over **untagged**. This is ADR-073 decision 1 *applied*, not amended — *"an adapter needing a tag supplies it"*, exactly as the KSeF adapter already resolves the **seller's** tag from its own connection config. **No ADR amendment was needed.**

The widening deliberately broke the three consumers with type errors, which is how each was made to decide explicitly rather than inherit a default:

| Adapter | Decision |
|---|---|
| **inFakt**, **Subiekt** | One tax-number slot each, so the domestic reading is the only placement. Requiring the tag would have dropped the number from every auto-issued document and — for Subiekt — silently downgraded it from a faktura to a paragon, since tax-number presence is that adapter's document-type trigger. Subiekt's two readers now share one `readDomesticTaxId` rule. |
| **KSeF FA(3)** | Must place one of three mutually exclusive elements. A ten-digit value resolves domestically; anything else carries a two-letter country prefix that the EU and non-EU elements **share**, so the tag is the only discriminator. An untagged non-domestic value is **refused naming the remedy** rather than filed under a guessed element of a document bound for a tax authority. |

A value a provider then rejects surfaces verbatim (ADR-073 decision 5); no adapter pre-judges what its provider accepts.

## Two things worth a reviewer's eye

**Core refuses to hand over a blank identifier.** `decodeBuyerTaxIdColumn` maps an *empty* column to `null` but decodes a *whitespace-only* one to the blank string itself. Harmless while every identifier was tagged; once untagged means domestic, it would have filed an empty tax number on a real document. `composeInvoicePayload` trims and stamps only what survives.

**`decodeBuyerTaxIdColumn` is a new value export from the `orders/types` cycle-breaker**, so it needed an entry in `check-types-sub-barrels.mjs`'s `VALUE_EXPORT_ALLOWED` with a written reason. It creates no new runtime edge: `AutoIssueTriggerService` already value-imports `PAYMENT_STATUS` from that same seam. The alternative was a second, drifting copy of the column encoding inside `invoicing`, which the column's own docblock forbids.

## Verification

| Gate | Result |
|---|---|
| `pnpm lint` | **exit 0** (85/85 invariants, incl. all three neutral-vocabulary guards) |
| `pnpm type-check` | **exit 0**, every package |
| core invoicing + buyer-tax-id types | 553 passed |
| Subiekt (full) | 237 passed |
| inFakt adapter | 108 passed |
| KSeF FA(3) buyer mapper | 12 passed |

No frontend file is touched; no migration.

New tests pin the auto-issue path **both ways**, because it decides unattended what a paid order gets: a known tax number reaches the invoice as `company`; core mints no scheme; asserted-none and not-asserted both issue as B2C.

## Docs

ADR-073's *Context* described only the receipt half of *"stores it, routes on it, never sends it"* — it now names the invoice half too. `architecture-overview.md` § 17 records the untagged rule and the three adapter decisions beside the existing #2599 material.

## Base

Branched from `3187-buyer-tax-number-on-receipt` and targets it: nothing in epic #3173 is merged to `main` yet. **Blocks #3188**, whose own dependency line already said *"until numbers are actually sent the column is always empty."*


piotrswierzy (commented):
## Tech Lead review

This is the right fix, argued the right way — `TaxIdentifier.scheme` made optional so core hands the number over untagged is ADR-073 decision 1 *applied* rather than amended, the three adapter decisions are each justified against what that adapter can actually place, and the KSeF refusal-rather-than-guess arm is exactly right for a document bound for a tax authority. The three-state column is never collapsed into two on the routing side and correctly collapses to two only at composition time. **But the change does not reach production: the `invoicing.issue` worker handler still rejects an untagged `buyer.taxId` as a terminal `business_failure`, so every auto-issued B2B invoice now fails to issue at all** — strictly worse than the defective-but-issued invoice this PR exists to replace. That is blocking.

### Findings

**[BLOCKING]** — `apps/worker/src/sync/handlers/invoicing-issue.handler.ts:196`
> The payload validator still requires a scheme:
> ```ts
> if (buyer.taxId !== null) {
>   if (!buyer.taxId || typeof buyer.taxId !== 'object') return fail('buyer.taxId');
>   if (!isNonEmptyString(buyer.taxId.scheme)) return fail('buyer.taxId.scheme');   // ← rejects untagged
>   if (!isNonEmptyString(buyer.taxId.value)) return fail('buyer.taxId.value');
> }
> ```
> Verified end to end against this PR's own head (`c9fe271`): `composeInvoicePayload` now writes `buyerTaxId: { value: buyerTaxNumber }` (`auto-issue-trigger.service.ts:1153`), which is flattened verbatim into `payload.buyer.taxId` at `:1167` with **no `scheme`**. The scheduled `invoicing.issue` job then hits the check above, `fail(...)` returns `null`, and the handler answers a **terminal `{ outcome: 'business_failure' }`** (ADR-007 — the runner does not retry). So for exactly the orders a routing rule selected an invoice for *because* the buyer has a tax number:
>
> | | before this PR | after this PR |
> |---|---|---|
> | B2B auto-issue | invoice issued, `taxId: null`, `type: 'private'` (defective) | **no invoice issued at all**, job dead |
>
> Nothing in the suite catches it because every new core spec stops at `syncJobs.schedule.mock.calls[0][0].payload` and asserts the payload — not that the consumer accepts it — and the verification table has no worker row. The two halves of the contract were widened in `libs/core` and left unwidened at the boundary that reads it.
>
> Fix: relax that clause to mirror the type (`scheme` present-but-wrong-shaped is still a `fail`, absent is valid), e.g.
> ```ts
> if (buyer.taxId.scheme !== undefined && !isNonEmptyString(buyer.taxId.scheme)) {
>   return fail('buyer.taxId.scheme');
> }
> ```
> and add a handler spec pinning that an untagged payload is accepted — that spec is what makes this a checked property rather than a comment. It also unblocks #3227, whose column is otherwise permanently `NULL` on every auto-issued row, i.e. the exact dependency it declares.

**[IMPORTANT]** — `libs/integrations/ksef/src/infrastructure/adapters/ksef-invoicing.adapter.ts:840`
> ```ts
> const providerCustomerId = taxId ? `ksef:${taxId.scheme}:${taxId.value}` : `ksef:${this.connectionId}:guest`;
> ```
> Untagged is now a reachable shape here, so this template-stringifies `undefined` into a persisted identifier: `ksef:undefined:5213796333`. The docblock calls it "a stable provider customer id … so the same buyer maps to a stable handle" — and after this change the *same* buyer gets two different handles depending on which path issued (`ksef:pl-nip:…` from the HTTP path where the operator chose the tag, `ksef:undefined:…` from auto-issue). No network call is made so nothing breaks loudly, which is precisely why it will not be noticed. This adapter already owns a domestic-inference rule one file over (`resolveScheme` in `fa3-buyer-id.mapper.ts`); reusing it here, or at minimum `taxId.scheme ?? 'untagged'`, keeps the handle stable and never renders `undefined`.

**[SUGGESTION]** — `libs/integrations/ksef/src/infrastructure/fa3/domain/fa3-buyer-id.mapper.ts:118`
> The refusal itself is right — `<KodUE>+<NrVatUE>` and `<KodKraju>+<NrID>` are structurally identical and guessing would file a buyer under the wrong element of a document sent to a tax authority. Worth naming the **operator-facing consequence** in the PR body and in `architecture-overview.md § 17`, because it is a genuine new failure mode rather than a no-op: a non-PL buyer whose `order_records.buyerTaxId` holds `DE123456789`, on a KSeF connection, previously auto-issued as `private` and now fails the job. It is correctly terminal (`ksef-retry-classifier.adapter.ts` already registers `InvalidBuyerIdentificationException` as non-retryable, so no retry ladder is burned), and the exception names the remedy — but an operator reads "no invoice" where they used to read "an invoice", and the doc text currently only says the value is *refused*, not that the refusal is new for a population that previously issued.

**[SUGGESTION]** — CI coverage on this PR
> `gh pr checks 3225` reports exactly one job, `Scaffolded Adapter Builds`. No lint, type-check or test job runs on a PR targeting a sibling branch, so the stack's stated "CI is the gate" posture does not hold here — the blocking finding above is precisely the class of defect a worker-package test run would have surfaced. Worth running `pnpm --filter @openlinker/worker test` locally against the handler and adding its row to the verification table.

Positive observations, because several non-obvious things are right and should not be lost in a re-roll:

- **`decodeBuyerTaxIdColumn` on the `orders/types` cycle-breaker** with a written `VALUE_EXPORT_ALLOWED` reason, correctly noting that `AutoIssueTriggerService` already value-imports `PAYMENT_STATUS` from that same seam so no new runtime edge is created. The alternative — a second copy of the column encoding inside `invoicing` — is what the column's own docblock forbids.
- **The whitespace defence** at `auto-issue-trigger.service.ts:1144`. `decodeBuyerTaxIdColumn` maps an empty column to `null` but decodes a whitespace-only one to the blank string itself; trimming and stamping only what survives is the correct reading, and the reasoning (an untagged value reads as domestic, so a blank would be filed as a real tax number) is exactly the failure it prevents.
- **`buyerTaxId !== null && buyerTaxId.value.trim().length > 0`** replacing object truthiness for the `BuyerType` derivation — an untagged identifier makes `{ value: '' }` representable, and the accompanying spec says so.
- **`readDomesticTaxId` as one rule with two Subiekt readers**, with the docblock naming why they must agree (an `FV` whose buyer carries no `nip`, and the inverse).

### Verdict
❌ **Request changes**

(Posted as a comment rather than a formal request-changes.)

---

piotrswierzy (commented):
## Tech Lead re-review (`c9fe271f` → `aed513de`)

The blocking finding is genuinely fixed, and fixed the right way: `scheme` is now `undefined`-tolerant while present-but-wrong-shaped is still a terminal `business_failure`, the comment explains the asymmetry against the `buyer.email` (#1797) precedent rather than just asserting it, and five cases now pin the behaviour — untagged accepted, tagged still accepted, B2C `null` still accepted, wrong-typed scheme refused, missing value refused. The ADR-073 citations are clean here (22 occurrences, zero `ADR-072`). **But the fix was delivered by replacing the handler's spec file rather than extending it, deleting 168 lines and roughly eighteen pre-existing cases — including every PII-discipline assertion and the retryable-transport-wrapping assertion — and the new file's header asserts that the handler "had NO spec", which is not true.** On a PR whose entire premise is that the suite failed to catch a regression, shrinking the suite is the wrong direction, so I am holding the verdict. The KSeF `ksef:undefined:` finding is also still open.

### Findings

**[BLOCKING]** — `apps/worker/src/sync/handlers/__tests__/invoicing-issue.handler.spec.ts`
> The file is not extended, it is **replaced**: `62+ 167-` against this PR's own base, and the base branch (`c1e572f8`, #3220's head) still carries the full spec — 225 lines, 11 `it(` blocks plus an `it.each` table of 11 cases. At this head it is 120 lines and 5 `it(` blocks. What goes with it:
>
> | Deleted coverage | Why it mattered |
> |---|---|
> | `describe('PII discipline on failure paths (F-validate-PII / D11)')` — 3 tests | Asserted the validation-failure log **omits `buyer.name`** and that a transport error's `SyncJobExecutionError` message carries neither the buyer sentinel nor the address. This is a security property of a handler that receives buyer PII, and this very PR widens what that payload carries. |
> | `transport / bridge-unreachable` — 1 test | Asserted a transport error is wrapped and **thrown** (retryable) rather than returned as an outcome. That is the ADR-007 status/outcome split for this handler; nothing else pins it. |
> | `deep payload validation ⇒ business_failure (F5)` — `it.each` over 11 cases | `schemaVersion`, empty/over-bound `lines`, negative `unitPriceGross`, `quantity <= 0`, `buyer.type`, **`half-populated taxId`**, non-string `buyer.email`, missing `connectionId`, empty/non-string `saleDate`. |
> | happy path (4) + `buyer.email` (2) | `BuyerProfile` reconstruction, `idempotencyKey` passthrough (F4), `saleDate` present/absent (#1525), pre-#1797 payload normalisation. |
>
> Two of those are load-bearing beyond bulk. The `half-populated taxId` case is *the same field this PR edits* — so a case that already exercised `buyer.taxId` validation was removed in the act of changing `buyer.taxId` validation. And `MAX_INVOICE_LINES` is exported from the handler solely for the deleted bound test; it now has no assertion anywhere, while `fiscalization-register.handler.ts:71` still cites it as the reference bound.
>
> The header is also factually wrong and will mislead the next reader:
> ```
> * The handler had NO spec, which is why this shipped:
> ```
> It had one, on `main` and on this PR's base. The defect shipped because the spec did not cover an *untagged* payload — a real and worth-stating gap, and a different statement.
>
> Fix is small: restore the file from the base and **append** the five new cases as a `describe('buyer.taxId (#3224)')` block. Nothing in the new cases conflicts with the old ones; they use the same `makeJob` shape and the same mocked `IInvoiceService`. (If the rewrite was motivated by the fuller `invoiceService` mock being tedious, that is what the existing `beforeEach` already builds.) One incidental note for whichever version survives: the new file spies on `console.warn`, whereas the handler logs through `Logger` from `@openlinker/shared/logging` — the old spec spied the handler's own `logger`, which is the reliable seam.

**[IMPORTANT — unchanged from last round]** — `libs/integrations/ksef/src/infrastructure/adapters/ksef-invoicing.adapter.ts:839`
> Still:
> ```ts
> const providerCustomerId = taxId
>   ? `ksef:${taxId.scheme}:${taxId.value}`
>   : `ksef:${this.connectionId}:guest`;
> ```
> Untagged is a reachable shape as of this PR, so this template-stringifies `undefined` into a persisted identifier — `ksef:undefined:5213796333` — and the docblock three lines above still promises "a stable provider customer id … so the same buyer maps to a stable handle". After this change the same buyer gets two handles depending on which path issued: `ksef:pl-nip:…` from the HTTP path where the operator chose the tag, `ksef:undefined:…` from auto-issue. No network call is made, so nothing fails loudly, which is exactly why it will sit there. `resolveScheme` in `fa3-buyer-id.mapper.ts` one file over already owns the domestic-inference rule; reusing it, or at minimum `taxId.scheme ?? 'untagged'`, keeps the handle stable and never renders `undefined`.

**[SUGGESTION — partially addressed]** — `docs/architecture-overview.md § 17`
> The § 17 text now states the KSeF refusal well ("an untagged non-domestic value is REFUSED naming the remedy rather than filed under a guessed element of a document bound for a tax authority") — that is the rule, and it reads correctly. What is still not said is that the refusal is **new for a population that previously issued**: a non-PL buyer whose `order_records.buyerTaxId` holds `DE123456789`, on a KSeF connection, previously auto-issued as `private` and now fails the job. Correctly terminal, remedy named in the exception, no retry ladder burned — but an operator reads "no invoice" where they read "an invoice" last week. One clause.

**[SUGGESTION]** — CI posture
> Still one job on this PR (`Scaffolded Adapter Builds`), so no lint, type-check or test run gates it. Worth a local `pnpm --filter @openlinker/worker test` row in the verification table — which, on the spec finding above, is also what would show the case count moving the wrong way.

Still correct and worth preserving through the re-roll — the `scheme !== undefined` narrowing and its comment, the untagged/tagged/null/wrong-typed case quartet, `decodeBuyerTaxIdColumn` on the `orders/types` cycle-breaker with its written `VALUE_EXPORT_ALLOWED` reason, the whitespace defence at `auto-issue-trigger.service.ts`, and `readDomesticTaxId` as one rule with two Subiekt readers.

### Verdict
❌ **Request changes**

The original blocker is closed. This one is new, and it is the same class of problem one layer up: the property that makes the fix durable is the test, and the PR removes more test than it adds. Restoring the file and appending the new block is a mechanical change.

---

norbert-kulus-blockydevs:
Both findings addressed. `0cdbee50d` restores the spec, `a306ef946` fixes the handle.

**[BLOCKING] `invoicing-issue.handler.spec.ts`** — you are right on every point, and the cause is worth recording because it is not a judgement call: I ran `ls` against `handlers/`, the spec lives in `handlers/__tests__/`, and I concluded "no spec" and wrote the file with `cat >`. 225 lines went with it.

Restored verbatim from the base and the four new cases appended as their own `describe('buyer.taxId (#3224)')`, reusing the existing `makePayload` / `makeJob`, the full `IInvoiceService` mock and the `warnSpy` on the handler's **own** logger — your incidental note was also correct, the replacement spied `console.warn` while the handler logs through `@openlinker/shared/logging`. 26 tests now (22 restored + 4 new); `MAX_INVOICE_LINES` has its bound assertion back.

The header no longer claims the handler had no spec. It had one; what it lacked was a case for an **untagged** payload — the narrower and true statement.

**[IMPORTANT] `ksef-invoicing.adapter.ts:839`** — fixed. `resolveBuyerIdSchemeTag` sits beside the FA(3) placement rule and reuses its `NIP_PATTERN`, but deliberately does **not** throw where `resolveBuyerIdentity` does: that function refuses an untagged non-domestic value because *placing* it means guessing an element of a document bound for a tax authority, while a customer handle carries no such consequence and a throw would fail `upsertCustomer` for a buyer the adapter can otherwise serve. Domestic resolves identically on both paths, so one buyer keeps one handle; anything else renders `untagged` and never `undefined`. Two regression tests, including one asserting the string contains no `undefined`.

Both **[SUGGESTION]**s not taken this round (§ 17 wording, CI posture). On CI you are right that it does not run here — `ci.yml` is `pull_request: branches: [main]` and every PR in this stack targets a sibling, so the only check is `Scaffolded Adapter Builds`. I verify locally instead: `pnpm lint` and `pnpm type-check` clean, worker handler 26/26, KSeF 65/65.

piotrswierzy (approved):
## Tech Lead re-review (`aed513de` → `d7d401ed`)

Both findings from the last pass are closed, and the important one is closed the right way. My ❌ was not about the handler fix — that was already correct — it was that the fix arrived by **replacing** `invoicing-issue.handler.spec.ts` instead of extending it, deleting the PII-discipline assertions, the retryable-transport-wrapping test and the 11-case F5 validation table (including the `half-populated taxId` row covering the very field this PR edits). I counted the file at both ends. **Base (`3187-buyer-tax-number-on-receipt` @ `d08e2c2b`): 226 lines, 11 `it(` blocks + 1 `it.each`. Head (`d7d401ed`): 279 lines, 15 `it(` blocks + 1 `it.each`.** The diff of quoted strings between the two is strictly additive — every string present at base is present at head — so nothing was dropped; the four new `it(`s are the `buyer.taxId (#3224)` block appended at the end. Spot-checking the named casualties: the `validation-failure log omits buyer.name` case, the two `SyncJobExecutionError`-excludes-buyer/payload cases, `a transport error from issueInvoice is wrapped … and THROWN`, the `buyer.email` reconstruct + pre-#1797-normalise pair, and the 11-row `cases` table with `half-populated taxId` at row 7 are all back verbatim. `MAX_INVOICE_LINES` is no longer an unasserted export either — the `over-bound lines` row now builds `MAX_INVOICE_LINES + 1` lines from it. The header sentence is corrected: it now says the spec *had* `buyer.taxId` coverage but none of an untagged one, which is the true account.

The second open item — `ksef:undefined:5213796333` at `ksef-invoicing.adapter.ts:839` — is fixed by the new `resolveBuyerIdSchemeTag`, and the reasoning behind it is the part worth calling out: it deliberately does **not** throw where `resolveBuyerIdentity` does, because a customer handle has no document-placement consequence and throwing would fail `upsertCustomer` for a buyer the adapter can otherwise serve; it resolves the domestic case so one buyer keeps one handle whichever path issued, and falls back to a literal `untagged` rather than ever rendering `undefined` into a persisted identifier. The PL-specific NIP inference sits adapter-side in `fa3-buyer-id.mapper.ts`, never in `libs/core` — ADR-026 held. I re-confirmed the standing rules: the value is carried verbatim through `order-to-issue-invoice-command.mapper.ts` (the only `.trim()` is a presence test deciding `b2b`/`b2c`, not a format check or a mutation of the carried value); `TaxIdentifier.scheme` is optional with the "treat an absent scheme as untagged, decide for yourself" contract stated on the type; the handler accepts an untagged id and still refuses a present-but-wrong-shaped `scheme` or an empty `value` as a terminal `business_failure`, mirroring the `buyer.email` (#1797) absent-vs-malformed split; and nothing here starts writing `'missing-required-tax-id'` or refusing on a missing tax id.

### Findings

### Verdict
✅ **Approve**

`mergeable: MERGEABLE`, `mergeStateStatus: BLOCKED` (stacked on #3220). Type Check, Docker Build Smoke, PHP Unit, Scaffolded Adapter Builds passed; Build / Lint / Test / Integration Tests are still **pending**, not passing — merge behind them.

---

Issue #3058, opened 2026-09-10
feat(oms): create/edit sourcing-rule dialog (filter/sort, after-action, priority list, effective window)


norbert-kulus-blockydevs:
## Problem / Context

Authoring a rule needs a kind-dependent name vocabulary (`RoutingFilterNameValues` / `RoutingSortNameValues`), a priority-location list only for the `priority` sort, and an after-action + optional effective window — none of which has a form today.

## Proposed Solution

Build a `Dialog`-based form with React Hook Form + Zod: kind radio (filter/sort), name select filtered by kind, after-action select, conditional priority-location picker, optional `effectiveFrom`/`effectiveTo` date fields. Validate `effectiveTo > effectiveFrom` client-side; surface server 400s (kind/name mismatch, unknown location) as form errors.

## Classification

**Type**: Frontend
**Layer**: Interface (component + form)
**File(s)**: apps/web/src/features/oms/components/sourcing-rule-dialog.tsx, sourcing-rule-dialog.schema.ts

## Dependencies

- Part of #3055

## Assumptions

- None

## Acceptance Criteria

- [ ] Name options are filtered to the selected kind's vocabulary
- [ ] Priority-location field only renders for the `priority` sort
- [ ] Client validates effectiveTo > effectiveFrom before submit
- [ ] Server-side kind/name-mismatch and unknown-location errors render as field errors
- [ ] Tests added or updated for non-trivial logic
- [ ] No architecture boundary violations (CORE ↔ Integration)

Issue #3057, opened 2026-09-10
feat(oms): ordered sourcing-rules table with arrow reorder


norbert-kulus-blockydevs:
## Problem / Context

Order is the whole meaning of a ruleset (first-match-wins filters, then sorts) - there is no way today to view or reorder it.

## Mockup

**Durable spec: `docs/plans/mockups/sourcing-rules-3055.html`** (on the epic branch `3055-sourcing-rules-admin-ui`).

`data-state` values this issue implements: `populated`, `gallery-active`, `gallery-governing`, `gallery-scheduled`, `gallery-retired`, `gallery-unrecognised`, `gallery-reorder-saved`, `gallery-reorder-conflict`, `gallery-priority-stale`.

## Proposed Solution

Build `SourcingRulesTable` in `features/oms/components/`: a `DataTable`-styled list showing position, kind (filter/sort), rule name, after-action, active window and status (active/scheduled/retired/unrecognised).

**Reorder is arrow buttons only - no drag-and-drop.** This is a decision recorded in the mockup and confirmed on the epic (#3055 § Decisions), not an omission:

- `apps/web` ships no DnD library today, and this screen does not justify adding one.
- Arrow buttons are keyboard-operable and touch-operable with no second code path; a DnD implementation would need a separate a11y fallback anyway, i.e. these same buttons plus a library.
- The mockup's `.rule-position__handle` (the `⠿` glyph) is a **static indicator** with `cursor: default` and `aria-hidden="true"` - it must not promise an interaction nothing backs. Ship it exactly that way, or drop it.

Each arrow click calls `PUT .../sourcing-rules/order`, which is **exhaustive**: the body must name every live rule id, or the server refuses and writes nothing (see `oms-sourcing-rules.controller.ts`). So the handler sends the full reordered id list, never a delta.

The first row's "up" and the last row's "down" are `disabled`, matching the mockup (`${idx===0?'disabled':''}`).

## Classification

**Type**: Frontend
**Layer**: Interface (component)
**File(s)**: apps/web/src/features/oms/components/sourcing-rules-table.tsx

## Dependencies

- Part of #3055
- Stacked on #3056 (API client + hooks)

## Assumptions

- None

## Acceptance Criteria

- [ ] Rows show position, rule summary, after-action badge, active window, status
- [ ] Arrow up/down reorder persists via the reorder mutation, sending the full live id list
- [ ] First row's up-arrow and last row's down-arrow are disabled
[...]

PR #3226, opened 2026-09-14
feat(oms): sourcing-rules API client + query hooks (#3056)

### LINKED ISSUE #3056: feat(oms): sourcing-rules API client + query hooks
## Problem / Context

No FE `features/oms` directory exists. The 6 sourcing-rules routes (`GET/POST /connections/:connectionId/sourcing-rules`, `PUT .../order`, `GET/PATCH/DELETE .../sourcing-rules/:ruleId`) have no client-side caller.

## Proposed Solution

Create `apps/web/src/features/oms/api/sourcing-rules.api.ts` (typed client methods for all 6 routes), `sourcing-rules.query-keys.ts`, `sourcing-rules.types.ts` (mirroring `SourcingRuleResponseDto`), and hooks in `features/oms/hooks/` (`use-sourcing-rules-query.ts`, `use-create-sourcing-rule-mutation.ts`, `use-update-sourcing-rule-mutation.ts`, `use-delete-sourcing-rule-mutation.ts`, `use-reorder-sourcing-rules-mutation.ts`), following the `features/fulfillment/api` layout precedent.

## Classification

**Type**: Frontend
**Layer**: Interface (API client + hooks)
**File(s)**: apps/web/src/features/oms/api/*, apps/web/src/features/oms/hooks/*

## Dependencies

- Part of #3055

## Assumptions

- None

## Acceptance Criteria

- [ ] All 6 routes have a typed client method
- [ ] Mutations invalidate the list query key on success
- [ ] Types mirror `SourcingRuleResponseDto` field-for-field

norbert-kulus-blockydevs:
## Summary

First `features/oms` slice: transport + TanStack Query hooks for the #2953 sourcing-rules admin API, which had **zero** frontend callers - an operator could author the OL fulfilment router's ordered ruleset only with `curl`.

Six typed client methods (`list`, `get`, `create`, `update`, `reorder`, `remove`) and six hooks, wired into `createApiClient` as `sourcingRules` and into `createMockApiClient` with an empty-ruleset default.

**Stack position:** bottom of the #3055 stack. Base is the epic branch `3055-sourcing-rules-admin-ui`; #3057 stacks on this one.

## Three decisions worth knowing

**`kind` / `name` / `afterAction` stay `string`, never mirrored unions.** The API stores all three verbatim and reports `recognised: false` for a row this build cannot evaluate - precisely so it is listed, found and deleted rather than hidden. A `z.enum` here would invert that: the day the API gains a rule name, the unrecognised row fails to parse and the table reports "no rules" for a connection that has some, which is a false statement about the operator's own configuration. The dialog still needs a list of names to OFFER; that is a separate concern, it comes from `libs/oms/src/routing/routing-vocabulary.types.ts`, `apps/web` cannot import `@openlinker/*` (#591), and a mirror needs its own `scripts/check-*-mirror.mjs` guard. #3058 owns it.

**Every write invalidates the whole CONNECTION, not the query that issued it.** The ruleset's meaning is its ORDER, and almost every write moves rows the mutation did not name - a create inserts at a position, a patch can move one, `PUT /order` renumbers all of them. A narrower key would refresh the row that was written and leave a sibling query (the other `includeSuperseded` value, a detail read) rendering positions that are no longer the router's. The hook tests assert the invalidation REACHES sibling keys the hook has never heard of, and does not reach another connection.

**The reorder sends the caller's id list verbatim, never one rebuilt from the cache.** The endpoint is exhaustive - a body naming a subset answers 409 and writes nothing. A list assembled here from a possibly-stale query would manufacture the very mismatch the server exists to refuse, and the operator would see a conflict about rules they never touched.

## Also here

- `readSourcingRuleConflict` reads `missingRuleIds` / `unknownRuleIds` off the 409 the API emits as FIELDS (a client parsing them out of prose breaks on the first reword), and reports `null` - not `[]` - for a 409 carrying none, since `[]` would claim the reorder named everything.
- Paths target `connections/:id/sourcing-rules` and are asserted against the **emitted URL**: `connections/:id/routing-rules` is the unrelated ADR-012 dispatch surface and would answer 200 with the wrong rules. Every consumer of this slice mocks the client, so a drifted path breaks nothing in any other test.
- `includeSuperseded: false` is sent explicitly rather than omitted - `ListSourcingRulesQueryDto` coerces `'true'`/`'1'` only, so omitting it relies on a server default instead of stating the request.
- `connectionId` is `encodeURIComponent`-ed: it reaches this client from a route param, i.e. from the address bar.

## Test plan

- [x] `eslint` on every changed file - clean
- [x] `vitest run src/features/oms` - 5 files, 34 tests, all pass
- [x] `vitest run src/app/api src/test` - 4 files, 37 tests, all pass (the new namespace is additive)
- [ ] Manual: nothing renders yet - the table (#3057) is the first consumer

Refs #3055
Closes #3056


piotrswierzy (commented):
## Tech Lead review

Transport-only slice, and a very careful one. I checked the three things the #2953 contract can be contradicted on and all three are right: the path is asserted against the **emitted URL string** (`sourcing-rules`, never the ADR-012 `routing-rules` namespace that `apps/api/src/mappings/http/fulfillment-routing.controller.ts` already mounts); `reorder` sends the caller's id list verbatim and never rebuilds it from the cache, which is the only way to avoid manufacturing the exhaustiveness mismatch the server exists to refuse; and `kind` / `name` / `afterAction` stay `string` with no `z.enum`, so an unrecognised row parses and reaches the operator instead of failing the whole list — which is precisely what `SourcingRuleResponseDto.recognised` exists to make possible. `readSourcingRuleConflict` reads `missingRuleIds` / `unknownRuleIds` off `ApiError.details`, matching the object literal `toHttpException` passes to `ConflictException`, and returns `null` rather than `[]` for a non-reorder 409. Connection-scoped invalidation is correct for a resource whose meaning is its order. One real gap: the new feature barrel was not registered in `.eslintrc.js`.

### Findings

**[IMPORTANT]** — `.eslintrc.js` (unchanged by this PR) + `apps/web/src/features/oms/index.ts`
> A new public feature barrel ships without its slug in either `no-restricted-imports` pattern group. `docs/frontend-architecture.md` § Feature Public Surface, *Adding a new public feature surface* step 2, requires the slug in **both** groups for every canonical subdirectory (`api`, `hooks`, `components`, `lib`, `types`), and states the consequence directly: "otherwise the rule will silently fail open for the new subdirectory". #2364 is the standing precedent for doing it up front — `features/automation` registered its slug with the folder, before any cross-feature consumer existed, so the gate could never be retro-fitted onto an already-violating tree. Add the ten `**/oms/<part>/**` lines (both groups) in this PR; #3060's page will be the first thing tempted to deep-import.

**[SUGGESTION]** — `apps/web/src/features/oms/api/sourcing-rules.api.ts` (`reorder`)
> `` `${buildSourcingRulesPath(connectionId)}/order` `` composes the *query-bearing* helper. It is correct today only because `reorder` never passes filters — the day someone does, the emitted URL is `…/sourcing-rules?includeSuperseded=false/order`, which 404s. Build it off the bare base (extract the `/connections/${enc}/sourcing-rules` prefix into a `collectionPath()` the two helpers share), so the shape cannot break by accident.

**[SUGGESTION]** — `apps/web/src/test/test-utils.tsx` (`sourcingRules` mock)
> `get` / `create` / `update` resolve `null` behind an `as ApiClient['sourcingRules']` cast, so a consumer test that exercises those paths receives `null` typed as `SourcingRule` and fails three components deep rather than at the seam. `list`/`reorder` already default to the honest empty array; a single `SourcingRule` fixture for the other three costs one object and #3058's dialog will need it anyway.

**[SUGGESTION]** — `apps/web/src/features/oms/hooks/use-sourcing-rule-query.ts:26-27`
> `connectionId as string` / `ruleId as string` are safe only because of the `enabled` predicate two lines below. A narrowed local (`if (connectionId === undefined || ruleId === undefined) throw …`, or hoisting the guard into a derived `const`) makes the safety self-evidencing rather than a cast the next reader has to re-verify. Same note on `use-sourcing-rules-query.ts:33`.

**Positive observations**
- `includeSuperseded: false` sent explicitly rather than omitted — correct against `ListSourcingRulesQueryDto`, whose `@Transform` coerces `'true'`/`'1'` only, so an omitted param would rely on a server default instead of stating the request. The test pins all three states including `undefined`.
- `.nullish()` on every nullable field, with `priorityLocationIds` tolerant even though the DTO always emits an array — the #939 rule applied correctly, including the reason (a list that renders empty is survivable; one that fails to parse is not).
- `encodeURIComponent` on both path segments, with the connection id specifically justified as address-bar-sourced, and asserted (`'a/b c'` → `a%2Fb%20c`).
- `readStringArray` returns `null` on a partially-typed array rather than filtering — a truncated id list would send the operator to correct the wrong rules. That is the right failure direction and it is not the obvious implementation.
- The barrel deliberately withholds the api module and the query keys with a stated reason. Start-narrow, per the doc.

### Verdict
🔄 **Approve with changes** — land the `.eslintrc.js` slug registration; the rest are optional. (`gh` is authed as `piotrswierzy` and the author is `norbert-kulus-blockydevs`, so this posts as a comment rather than a blocking review.)

---

piotrswierzy (commented):
## Tech Lead re-verification (head unchanged at `945c8a1`)

Head has not moved and neither has `main`. The contract half of the standing review is unchanged and still verified against #2953 — the `sourcing-rules` path is asserted on the emitted URL rather than on a builder argument, the exhaustive reorder body is sent verbatim, and no union is mirrored, so an `recognised: false` row still parses. The IMPORTANT finding is the part I re-checked, and it has **not** been addressed anywhere on the stack.

I read `.eslintrc.js` at this head and again at the tip of the stack (#3235, `6570b3a`). In both, the only `oms` occurrences are `libs/oms/**/*.ts` — the ADR-055 product-package rules at lines ~678 and ~846, which are unrelated. The `oms` slug is absent from **both** `no-restricted-imports` pattern groups, for every canonical subdirectory. So none of the five new siblings (#3229, #3230, #3233, #3234, #3235) added it, and by #3235 the feature has `api`, `components`, `hooks` and `lib` subdirectories all shipping public modules with the rule silently open on every one of them. `docs/frontend-architecture.md` § Feature Public Surface requires the slug in both groups and states the rule "silently fails open" otherwise; #2364 is the precedent, where the slug was registered up front precisely *because* it had no cross-feature consumer yet.

**Stack position:** 2nd in the OMS stack, based on `3055-sourcing-rules-admin-ui`. Worth flagging: **that base branch has no open PR**, so the stack is rooted on an unreviewed branch with no route to `main` of its own. Six PRs (#3228, #3229, #3230, #3233, #3234, #3235) now sit on top of this one.

**CI:** no checks reported on this branch at all — `ci.yml` triggers on `pull_request: branches: [main]` only. Absent, not passing.

### Findings
No new findings. The single IMPORTANT stands unchanged: add `oms/api`, `oms/components`, `oms/hooks`, `oms/lib`, `oms/types` to both `no-restricted-imports` groups in `.eslintrc.js`. One line in this PR closes it for the whole stack.

### Verdict
🔄 **Approve with changes**

---

norbert-kulus-blockydevs:
Fixed on `2d93a01f7`. All five canonical subdirectories - `oms/api`, `oms/hooks`, `oms/components`, `oms/lib`, `oms/types` - are now in **both** `no-restricted-imports` groups (`.eslintrc.js:329-333` and `:608-612`), placed after the `returns/*` entries in each, matching the ordering the other slugs use.

Verified that the rule actually fires rather than just reading right. Dropped a probe module into another feature:

```ts
// apps/web/src/features/orders/lib/__eslint-probe.ts
import { SourcingRulesTable } from '../../oms/components/sourcing-rules-table';
```

```
1:1  error  '../../oms/components/sourcing-rules-table' import is restricted from being used
by a pattern. Cross-feature imports must target the feature's public barrel
(`features/<name>`), not its internals.  no-restricted-imports
✖ 1 problem (1 error, 0 warnings)
```

Probe deleted afterwards. `pnpm lint` is clean on the branch, so no existing import in the tree was relying on the rule being open.

Taking your point about registering it here rather than up-stack: one line in the lowest PR closes it for all six siblings, and #2364 is the precedent for adding the slug before the feature has a cross-feature consumer.

On the base-branch flag - `3055-sourcing-rules-admin-ui` having no open PR of its own is real and is the stack owner's to resolve, not something this PR can fix from where it sits. Noted so it does not get lost.

piotrswierzy (approved):
## Tech Lead re-review (`945c8a1` → `2d93a01`)

One file changed, and it is the one that mattered: `.eslintrc.js` now carries the `oms` slug. I read the file at this head and confirmed the five `**/oms/<part>/**` entries appear in **both** pattern groups — the `apps/web/src/features/**` override (line 175, entries at 329-333) and the `apps/web/src/plugins/**` override (line 456, entries at 608-612) — for every canonical subdirectory (`api`, `hooks`, `components`, `lib`, `types`), with an in-place comment naming the doc rule and why a missing pair fails open. That closes the single IMPORTANT finding that had stood across two rounds and, since it lands at the bottom of the stack, closes it for all seven PRs above it: `features/oms` now ships `api`, `components`, `hooks` and `lib` with the deep-import ban guarding every one.

I re-verified the four contract checks at this head rather than carrying them forward. `buildSourcingRulesPath` emits `/connections/:id/sourcing-rules` and the test asserts the **emitted string**, including the negative case naming the ADR-012 `routing-rules` dispatch surface by name. `reorder` passes `body` straight into `JSON.stringify` with no cache read anywhere on the path, and the test pins `{ ruleIds: ['b', 'a'] }` verbatim. `kind` / `name` / `afterAction` are `z.string()` with no `z.enum` and `recognised: z.boolean()` beside them, so an unevaluable row still parses. `readSourcingRuleConflict` reads `missingRuleIds` / `unknownRuleIds` off `ApiError.details` as fields, returns `null` (never `[]`) for a 409 carrying none, and `readStringArray` still refuses a partially-typed array rather than filtering it.

### Findings

No blocking or important issues found. The three prior SUGGESTIONs were not taken and remain optional, restated once for the record rather than re-argued:

**[SUGGESTION]** — `apps/web/src/features/oms/api/sourcing-rules.api.ts:113`
> `` `${buildSourcingRulesPath(connectionId)}/order` `` still composes the query-bearing helper. Correct today only because `reorder` passes no filters; the day someone does, the emitted URL is `…/sourcing-rules?includeSuperseded=false/order`. A shared bare-base `collectionPath()` makes the shape unbreakable by accident.

**[SUGGESTION]** — `apps/web/src/test/test-utils.tsx`
> `get` / `create` / `update` still resolve `null` behind the `as ApiClient['sourcingRules']` cast, so a consumer test exercising those paths fails three components deep rather than at the seam. One `SourcingRule` fixture covers all three, and #3229's dialog already needs it.

**[SUGGESTION]** — `use-sourcing-rule-query.ts:26-27`, `use-sourcing-rules-query.ts:33`
> The `as string` casts are still load-bearing on the `enabled` predicate two lines below. A narrowed local makes the safety self-evidencing.

**Stack position:** still rooted on `3055-sourcing-rules-admin-ui`, which **still has no open PR** — I checked `--state all` and the list is empty. Seven PRs now sit on a base branch with no route to `main` of its own. Merging this stack needs either a root PR for that branch or a re-target.

**CI:** no checks reported on this branch — `ci.yml` triggers on `pull_request: branches: [main]` only. Absent, not passing. `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN` against its stack base.

### Verdict
✅ **Approve**

---

Issue #2090, opened 2026-08-14
[TASK] Frontend — Invoices list: merged Document type column plus shared Order and Connection cells


norbert-kulus-blockydevs:
## Problem / Context

The Invoices list renders the worst version of both identity cells in the app:

- **Order column** (`pages/invoicing/invoices-list-page.tsx:218-227`): the raw 41-character `orderId`
  printed in a `mono-text` span. No truncation, no Copy, no link, nothing else.
- **Connection column** (`:277-286`): `<span className="text-muted" title={r.connectionId}>` — the id
  is hidden in a `title` attribute, which is invisible, unselectable, and unreachable on touch.
- **Document identity is split across two columns**: `invoiceNumber` (`:229-238`) and `documentType`
  (`:239-248`) answer one question — *what document is this?* — in nine columns' worth of horizontal
  budget.

`r.orderSummary` — added by #1995 / PR #2012, deduped by #2021, typed at
`features/invoicing/api/invoicing.types.ts:114` — is **consumed by nothing**.

Specified in #1996 (row 05) and `docs/plans/mockups/list-identity-cells-1996.html`; never implemented,
because #1996 was auto-closed by the docs-only PR #1999.

## Proposed Solution

**1. Merge `invoiceNumber` + `documentType` into one `Document type` column** (9 columns → 8):

```
FV/2026/08/0042      ← providerInvoiceNumber, wrapped in the existing InvoicePdfLink
Invoice              ← the document type
```

Line 1 is `–` when there is no number (a receipt). The number stays a working PDF link. Retire the
now-unused `invoice.column.invoiceNumber` i18n key.

The merged column must be **always visible** — `documentType` currently carries `hideBelow: 768`
(`:247`), but the merged column becomes the host for S8's tablet fold, so it cannot disappear.

**2. Order column → `OrderIdentityCell`** (S1), fed from `r.orderSummary`.

> **Corrected during implementation**, the same way #2089 was and for the same reason. This said
> "renders `–` when `orderSummary` is null"; the shared cell renders a **shortened internal-id link**
> instead. `buildOrderSummary` returns null both for "no order record" *and* for "the snapshot carried
> no parseable items" — the latter is a live order, which a dash would strand. What this issue actually
> set out to remove is the RAW 41-character UUID, and a shortened link removes it while keeping the row
> navigable. See deviation 1 in `order-identity-cell.tsx`.

[...]

Issue #1202, opened 2026-06-24
invoicing: serve AC-6 with/without-tax-id list sub-filter (denormalize buyer tax-id onto invoice_records)


norbert-kulus-blockydevs:
**Context:** The `GET /invoices` list endpoint (#1119) implements the AC-6 filters that map to a real `invoice_records` column — `status`, `connectionId`, `regulatoryStatus`, and the `issuedAt` date range. The AC-6 **"with / without tax id"** sub-filter is deliberately **not** served by #1119.

### Why it's deferred
The persisted `InvoiceRecord` projection has **no buyer / tax-id column** — the buyer lives on the `Order`, not the invoice projection. Serving a `hasTaxId` filter would require denormalizing `buyerTaxId` (or a boolean `hasTaxId`) onto `invoice_records` plus a migration and a backfill — out of #1119 scope.

#1119 does **not** silently accept-and-ignore the param: `ListInvoicesQueryDto` uses `forbidNonWhitelisted`, so a `hasTaxId` query param is rejected with a 400 rather than appearing to work. AC-6 sign-off must not be claimed for this sub-filter until this issue ships.

### Proposed closure
1. Add a `hasTaxId` (or denormalized `buyerTaxId`) column to `invoice_records` + migration + backfill from the originating orders.
2. Extend `InvoiceRecordFilters` with the neutral discriminator and `InvoiceRecordRepositoryPort.findMany` to honour it.
3. Add `hasTaxId` to `ListInvoicesQueryDto` (whitelisted) and the controller filter mapping.

### Acceptance criteria
- [ ] `GET /invoices?hasTaxId=true|false` filters by buyer-tax-id presence.
- [ ] Migration + backfill land with the column.
- [ ] AC-6 "with/without tax id" sub-filter signed off.

Follow-up to #1119. Referenced from `InvoicingController.listInvoices` and `InvoiceRecordFilters`.

Issue #976, opened 2026-06-05
[TECH-DEBT] CI/Testing — stabilize @openlinker/integrations-prestashop Jest suite (intermittent OOM/SIGKILL on self-hosted runner)


piotrswierzy:
## Problem / Context

The `test` CI job (`.github/workflows/ci.yml`, `runs-on: self-hosted`, runs `pnpm test:ci`) intermittently goes **red on the `@openlinker/integrations-prestashop` package with zero actual test failures**. A Jest worker is killed by the OS and the suite reports as failed:

```
FAIL src/infrastructure/http/__tests__/prestashop-webservice.client.spec.ts
  ● Test suite failed to run
    A jest worker process (pid=…) was terminated by another process:
    signal=SIGKILL, exitCode=null
...
A worker process has failed to exit gracefully and has been force exited.
Test Suites: 1 failed, 23 passed, 24 total
Tests:       339 passed, 339 total          ← zero test failures
```

`signal=SIGKILL` / `exitCode=null` is an **OS OOM-kill of the worker**, not an assertion failure. In the same red runs the `@openlinker/integrations-allegro` `allegro-http-client.spec.ts` also balloons to **~950 s**, consistent with memory/CPU starvation under full-suite parallelism (`pnpm -r` runs every package's Jest concurrently, and each package's Jest defaults to ~`cores − 1` workers).

**Why this matters beyond annoyance:** a SIGKILL'd suite is visually identical to a genuine test failure (red ❌). It trains everyone to reflexively "just re-run", which is exactly how a *real* regression eventually slips through unnoticed — the CI analogue of the #374 silent-failure trap. It has already flaked #962's CI (re-run went green) and recurred on #974's CI.

The prestashop package is the heaviest in the repo and has no worker/memory bounds: `libs/integrations/prestashop/jest.config.mjs` sets no `maxWorkers` or `workerIdleMemoryLimit`, and its biggest spec — `prestashop-order-processor-manager.adapter.spec.ts` — is **2279 lines** in a single file.

This is **CI/test-infra flakiness only**; no production code is implicated.

## Proposed Solution

Investigate + apply the cheapest stabilization that holds, in roughly this order:

[...]

PR #3227, opened 2026-09-14
feat(invoicing): buyer tax ID column on the invoice list

### LINKED ISSUE #3188: feat(invoicing): buyer tax ID column on the invoice list

norbert-kulus-blockydevs:
Closes #3188
Refs #3173

## What was wrong

`/invoices` already filtered on tax identity (`taxId=with|without`, #1202) and showed **no column for it** — so the filter answered a yes/no question and hid the answer. The projection stored only the boolean that serves the filter.

## The column

`invoice_records.buyerTaxId`, three-state encoded exactly as `order_records.buyerTaxId` is (#2599): `NULL` = nothing asserted, `''` = the source asserted the buyer has none, otherwise the id. The wire contract mirrors the order detail's: key **absent** = not asserted, `null` = asserted-none, string = the value.

## Frozen at issue, not joined — and the join was available

The invoices list **already** batch-reads the page's order records (`resolveOrderSummaries` → `IOrderRecordService.findByIds`, one query). Reading the order's three-state value there would have been free and needed no migration.

It was rejected anyway: **an invoice is an immutable fiscal document, while `order_records.buyerTaxId` is rewritten by every re-ingestion.** A joined read could later show a number the issued document does not carry — a false statement about a legal record, on a list of legal records.

## The rule that matters

`freezeBuyerTaxIdAtIssue` is pure and module-scoped so its load-bearing property can be pinned directly rather than through a whole issuance:

> It never claims more than the document does.

The id comes from `cmd.buyer.taxId` — what actually goes onto the document — and the caller's assertion is consulted **only** to tell the two *absences* apart. An operator issuing by hand with no tax id, against an order that has one, therefore reads as not-asserted rather than putting a number on a row whose document has none.

The assertion reaches the write path on the command, for the same reason `taxRateEra` is carried there: `InvoiceService` depends on no orders-context token, and adding one would be a new module edge into the context whose module already imports this one.

## Rendering

Reuses the order detail's own `OrderBuyerTaxIdValue` rather than a second copy of the three-state vocabulary — there is no second render to drift. Consumed from `pages/`, which sits above both features, so **no feature-to-feature edge is created** (and no barrel-to-barrel cycle: `features/orders` already value-imports `features/invoicing`).

The cell carries the hook the mockup declares, `invoice-row-tax-id-{i}`. Worth knowing: the index is the row's position in the **fetched page**, so it stops tracking visual order once an operator sorts a column client-side. Stated at the call site rather than papered over — a hook that silently renumbered itself under a sort would be worse.

## Migration

Additive, nullable, **no default and no backfill**. Every existing row predates the column and there is nothing to recover its value from, so `NULL` is the truthful "not asserted" rather than a manufactured answer. Unindexed — the filter is served by the boolean beside it and nothing sorts on the value.

## Rider: two `docs/lessons.md` entries

Carried by this issue:

1. **Bound the fan-out when the hook escalates to the full suite; never reach for `--no-verify`.** A core barrel or `<ctx>.tokens.ts` edit is a wide `smart-test` trigger → `pnpm -r test` at pnpm's default concurrency of 4 → OOM. `test:ci` was bounded to 2 in #976; the hook's path never was. `--no-verify` skips lint and type-check too, and on a stacked epic CI does not run either.
2. **Read issues and retarget PRs through `gh api`.** `gh issue view` and `gh pr edit --base` both fail on this repo's Projects (classic) data — and several `docs/plans/*.md` still recommend the latter.

## One existing assertion changed

The #2090 header count moves 8 → 9. That test's intent is carried by its `Invoice no.` absence check; the count is kept beside it so a column added by accident still fails something.

## Verification

| Gate | Result |
|---|---|
| `pnpm lint` | **exit 0** (85/85 invariants) |
| `pnpm type-check` | **exit 0**, every package |
| core invoicing | 482 passed |
| api invoicing | 158 passed |
| worker `invoicing-issue` | 22 passed |
| invoices list page | 35 passed |

New tests pin the freeze rule's eight cases — including that it never claims an id the document does not carry — the DTO's three wire states (with the absent key asserted *after* serialization, since an explicit `null` would read as asserted-none), and the three renderings on the page, told apart **by hook rather than by wording**.

## Base

Branched from `3224-buyer-tax-id-on-auto-invoice` and targets it. #3224 is what makes this column non-empty on an auto-issued invoice — without it the write path has no number to freeze.


piotrswierzy (commented):
## Tech Lead review

The strongest of the three. The central decision — freeze at issue rather than join from the order, *even though the join was already free* because `resolveOrderSummaries` batch-reads the page's order records — is argued from the right premise (an invoice is an immutable fiscal document; `order_records.buyerTaxId` is rewritten by every re-ingestion, so a joined read could later show a number the issued document does not carry) and the PR names the cheaper option it rejected rather than pretending it did not exist. `freezeBuyerTaxIdAtIssue` is pure and module-scoped specifically so its **negative** property can be pinned directly, and the test that matters — `command(null, '5213796333')` freezes to `null` — is there. The three wire states are asserted *after* serialization, which is the only way the absent-key-vs-explicit-`null` distinction is actually checked. My concerns are one inherited blocker and one documentation gap.

### Findings

**[BLOCKING — inherited from #3225, not this PR's code]** — `apps/worker/src/sync/handlers/invoicing-issue.handler.ts:196`
> This PR's acceptance criterion is *"Filtering to `With tax ID` shows which id each row carries"*, and its stated dependency is #3224 *"which is what makes this column non-empty on an auto-issued invoice"*. That dependency is currently unmet at this PR's own head (`bef2550`):
> ```ts
> if (!isNonEmptyString(buyer.taxId.scheme)) return fail('buyer.taxId.scheme');
> ```
> #3224 composes an **untagged** `{ value: '5213796333' }` into `payload.buyer.taxId`, which this validator rejects, returning a terminal `{ outcome: 'business_failure' }`. So no auto-issued B2B invoice is created at all — meaning `invoice_records.buyerTaxId` stays `NULL` on every row this column exists to populate, and the new cell renders `order-buyer-tax-id-unknown` forever. This PR's own `invoicing-issue.handler.ts` edit threads `buyerTaxIdAssertion` into `toCommand` but does not touch the validator two hundred lines above it. Fix belongs in #3225; flagging it here because this PR cannot meet its acceptance criterion until it lands, and the two specs that would catch it (a handler spec accepting an untagged payload, and any end-to-end assertion that a B2B auto-issue produces a non-`NULL` column) exist in neither PR.

**[IMPORTANT]** — `docs/architecture-overview.md`
> A new persisted column on a fiscal-document table, a new `IssueInvoiceCommand` field, a new job-payload field and a new invariant (*"it never claims more than the document does"*) land with no entry in **§ 14 Invoicing**. #3225 records its half in § 17; this one records nothing. The freeze-not-join rule and the reason for it are exactly the kind of decision this file exists to hold — the next person looking at `invoice_records.buyerTaxId` and `order_records.buyerTaxId` side by side needs to find, in the overview, why one is not derived from the other. The ORM docblock and the migration header both say it well; one paragraph of that text in § 14 beside the existing #2599 material closes it.

**[SUGGESTION]** — `apps/api/src/invoicing/http/dto/invoice-record-response.dto.ts:36`
> ```ts
> import { decodeBuyerTaxIdColumn , BuyerTaxId} from '@openlinker/core/orders';
> ```
> Two things: the spacing is malformed (`, BuyerTaxId}`), and `BuyerTaxId` is a **type-only** export from that barrel (`export type { BuyerTaxId }`) pulled in through a value-import clause — the line directly above it already does `import type { OrderSummary }`. Split it:
> ```ts
> import { decodeBuyerTaxIdColumn } from '@openlinker/core/orders';
> import type { BuyerTaxId } from '@openlinker/core/orders';
> ```

**[SUGGESTION]** — `apps/web/src/pages/invoicing/invoices-list-page.tsx:351`
> `rowIndexById` is rebuilt on every render of the page, not just when `query.data` changes. At `PAGE_SIZE = 20` that is free, so this is style rather than performance — but the surrounding file uses `useMemo` for derived collections and a `useMemo` keyed on `query.data?.items` would also make the comment's "the FETCHED PAGE" framing structural rather than incidental. The comment itself is good and should stay: stating that the hook stops tracking visual order under a client-side sort, rather than silently renumbering, is the right call and matches the mockup, which declares `invoice-row-tax-id-{i}` and nothing stronger.

Positive observations:

- **Migration hygiene is clean.** `1883000000000` is strictly greater than `main`'s tail (`1880000000000-create-price-change-episodes`) and collides with no open PR (`#3219` has `1882…`, `#3210` `1881000000001`, `#3201/#3035/#3027` `1881000000000` — none is `1883…`). Class suffix matches the filename, both `up()` and `down()` present, additive/nullable/no default/no backfill with the reason stated (*"nothing to recover its value from, so `NULL` is the truthful answer rather than a manufactured one"*), and the "unindexed because the boolean beside it serves the filter" note is the right justification to write down rather than leave to be re-litigated.
- **Rendering reuses `OrderBuyerTaxIdValue` from the `orders` barrel, consumed from `pages/`**, which sits above both features — so no feature-to-feature edge is created, and there is genuinely no second render of the three-state vocabulary to drift. The barrel export carries that reasoning inline.
- **The `accessor` sorts on the value only**, with the comment stating that the two absences sort together because "inventing an ordering between 'has none' and 'we were not told' would assert a difference the sort cannot support". Correct, and the sort would otherwise have been the one place the three states quietly become two.
- **The #2090 header-count change is handled honestly** — 8 → 9, with the note that the test's intent is carried by the `Invoice no.` absence check and the count is kept beside it so an accidental column still fails something. That is the right way to touch someone else's assertion.
- **`entity.buyerTaxId = input.buyerTaxId ?? null`** with the comment naming why it is `??` and not a falsy coalesce. `''` surviving as itself is the whole feature.

### Verdict
🔄 **Approve with changes**

The code is ready. It is gated on #3225's handler fix landing first, without which its own acceptance criterion cannot be demonstrated.

---

piotrswierzy (commented):
## Tech Lead re-review (`bef25503` → `00b182b8`)

The IMPORTANT documentation finding is closed, and the § 14 entry is better than the one I asked for — it states the freeze-not-join rule, the "never claims MORE than the document, only ever less" invariant for `buyerTaxIdAssertion`, the nullable-no-backfill honesty argument, and why this is a second column rather than a widening of `hasBuyerTaxId` (#1202). The inherited blocker is resolved: this PR's base now carries #3225's `scheme !== undefined` relaxation, so an auto-issued B2B invoice actually issues and this column has rows to populate. The migration timestamp is still unique. One inherited issue has replaced the old one, and it is again not this PR's code.

### Findings

**[BLOCKING — inherited from #3225, not this PR's code]** — `apps/worker/src/sync/handlers/__tests__/invoicing-issue.handler.spec.ts`
> The handler-validator blocker I raised last round is genuinely fixed on the base branch and this PR inherits it — verified at this head, `buyer.taxId.scheme` is now optional, so an untagged `{ value: '5213796333' }` is accepted and `invoice_records.buyerTaxId` will be non-`NULL` on auto-issued B2B rows. This PR's acceptance criterion is demonstrable.
>
> What arrives with it is a rewrite of that handler's spec (`62+ 167-`), dropping ~18 pre-existing cases including every PII-discipline assertion and the retryable-transport-wrapping assertion. Detail is on #3225; flagging here only because it rides in this PR's delta. Nothing for this PR to do.

**[RESOLVED]** — `docs/architecture-overview.md § 14`
> Landed, and correctly placed beside the existing #2599 material. The passage a future reader most needs — *"An order's tax number is editable and re-ingestible; a document that has already been issued is not"* — is exactly the sentence that stops someone re-deriving this column from the order six months from now.

**[VERIFIED]** — `apps/api/src/migrations/1883000000000-add-invoice-buyer-tax-id-value.ts`
> Re-checked against every open PR at this timestamp, not just the ones I happened to look at last round. `1883000000000` is claimed by this PR alone:
>
> | Timestamp | Claimants |
> |---|---|
> | `1881000000000` | #3027, #3201 — **collision** |
> | `1881000000001` | #3210 |
> | `1882000000000` | #3219 |
> | **`1883000000000`** | **#3227 only** ✅ |
> | `1884000000000` | #3159, #3241 — **collision** |
> | `1885000000000` | #3035 |
>
> Strictly greater than `main`'s tail (`1880000000000-create-price-change-episodes`), class suffix matches the filename, both `up()` and `down()` present. Clean. The two collisions above are other PRs' problem, but whichever of them merges second will need re-prefixing — and if that lands as `1883…` this PR inherits the collision, so re-run `scripts/check-migration-timestamps.mjs` at merge time rather than trusting this table.

**[SUGGESTION — unchanged]** — `apps/api/src/invoicing/http/dto/invoice-record-response.dto.ts:36`
> Still `import { decodeBuyerTaxIdColumn , BuyerTaxId} from '@openlinker/core/orders';` — malformed spacing, and `BuyerTaxId` is a type-only export pulled through a value clause while the line above already does `import type { OrderSummary }`. Prettier would have caught the spacing; on a stack committed with `--no-verify` it will not. Split the two.

**[SUGGESTION — unchanged]** — `apps/web/src/pages/invoicing/invoices-list-page.tsx`
> `rowIndexById` still rebuilt per render rather than `useMemo`-keyed on `query.data?.items`. Free at `PAGE_SIZE = 20`; style only, and the comment stays either way.

The positives from last round all hold: the freeze reasoning, `OrderBuyerTaxIdValue` reused from the `orders` barrel at the `pages/` layer so no feature-to-feature edge appears, the accessor sorting on the value only so the two absences sort together, the honest #2090 header-count change, and `?? null` rather than a falsy coalesce so `''` survives as itself.

### Verdict
🔄 **Approve with changes**

The code is ready and now demonstrable. Two cosmetic suggestions of its own; the remaining blocker belongs to #3225.

---

norbert-kulus-blockydevs:
Nothing outstanding on this PR's own code; the inherited blocker is closed below it.

**[BLOCKING — inherited from #3225]** — fixed on `3224-buyer-tax-id-on-auto-invoice` (`0cdbee50d`) and cascaded here, so this branch now carries the restored 26-test handler spec rather than the 5-test replacement. Detail in the #3225 thread.

**[RESOLVED] § 14** and **[VERIFIED] the migration** — thank you for re-checking both.

Both **[SUGGESTION]**s left as they are for now (the `invoice-record-response.dto.ts:36` wording and the invoices-list cell). Say the word and I will take them here rather than as follow-ups.

Also cascaded in since your pass: the shared `neutral-vocabulary-sweep.ts` from #3183, which had stopped at #3186.

piotrswierzy (approved):
## Tech Lead re-review (`00b182b8` → `dc4aa766`)

The inherited BLOCKING is gone at its source: #3225's handler now accepts an untagged `taxId`, so the frozen `invoice_records.buyerTaxId` column no longer lands `NULL` on every auto-issued B2B invoice, and this PR's column stops being a permanently-empty UI element. Of the two SUGGESTIONs, the import is fixed — `invoices-list-page.tsx:72` now reads `import { OrderBuyerTaxIdValue, OrderIdentityCell, formatOrderRef } from '../../features/orders';`, with the malformed `, BuyerTaxId}` fragment gone and the symbol exported from the `orders` barrel rather than deep-imported.

I re-verified the migration timestamp by enumerating migration filenames across all 52 open PRs rather than trusting the diff-against-`origin/main` guard, which cannot see sibling branches. **`1883000000000` is unique and unclaimed.** The neighbourhood has shifted since my last pass and the brief's map is now partly stale: `1878000000000` → #3014, `1880500000000` → #3201 (it moved down, not to `1881`), `1881000000000` → #3027, `1881000000001` → #3210, `1882000000000` → #3219, `1884000000000` → #3241, `1885000000000` → #3035, and #3159 has moved up to `1887000000000` + `1888000000000`. `1883` sits in the one free slot between #3219 and #3241, strictly above `main`'s tail of `1880000000000`. The migration carries both `up()` and `down()`, the class suffix matches the filename prefix, and it is nullable with no default and no backfill — which is the right call and the header says why: nothing exists to recover a pre-column value from, so `NULL` is the truthful "not asserted" rather than a manufactured answer.

The three-state discipline is intact and, better, is documented at the point where it would be violated: the migration header states `NULL` / `''` / the id explicitly and names `decodeBuyerTaxIdColumn` as the only correct reader, warning that a bare `IS NOT NULL` reports true for the asserted-none row. The freeze-at-issue choice is also argued rather than asserted — an invoice is an immutable fiscal document while `order_records.buyerTaxId` is rewritten by every re-ingestion, so a joined read could later display a number the issued document does not carry. `freeze-buyer-tax-id-at-issue.spec.ts` covers it. The value is displayed through the order-detail component rather than a second copy of the vocabulary, so there is no second render to drift.

### Findings
**[SUGGESTION]** — `apps/web/src/pages/invoicing/invoices-list-page.tsx:352`
> `rowIndexById` is still built as a bare `new Map(...)` on every render rather than under `useMemo`, unlike `connectionsById` twenty lines above it. I am not pressing it: the sibling `columns` array is also a plain render-body const, so this matches the page's existing shape, and the map is at most a page of rows. Worth a `useMemo` only if you touch the file again.

### Verdict
✅ **Approve**

`mergeable: MERGEABLE`, `mergeStateStatus: BLOCKED` (stacked on #3225, itself on #3220 — merge bottom-up). Type Check, Docker Build Smoke, PHP Unit and Scaffolded Adapter Builds passed; Build / Lint / Test / Integration Tests are still **pending**, not passing.

---

norbert-kulus-blockydevs:
## CI fix: `Test` job (`@openlinker/core`)

**Failing run:** [34845516211](https://github.com/openlinker-project/openlinker/actions/runs/34845516211/job/103980439693) at `dc4aa766` - `Test Suites: 2 failed, 421 passed` / `Tests: 26 failed, 6438 passed`.

### What was broken

Two suites, two independent stacked-merge artifacts.

**1. `orders/application/services/sales-document-view.service.spec.ts` - 25 failures**

```
Nest can't resolve dependencies of the SalesDocumentViewService
(Symbol(OrderRecordRepositoryPort), Symbol(IInvoiceService), ?, Symbol(ConnectionPort),
 Symbol(ISalesDocumentRulesService)).
Please make sure that the argument dependency at index [2] is available in the RootTestModule context.
```

plus, in the same suite:

```
OrdersModule wiring for SalesDocumentViewService > should import FiscalizationModule
  Expected value: [Function FiscalizationModule]
```

The spec *did* provide `FISCAL_REGISTRATION_SERVICE_TOKEN`. The `?` at index [2] is the tell: the token itself evaluated to `undefined`.

**2. `sales-documents/application/services/sales-document-rules.service.spec.ts` - 1 failure**

```
- Expected  - 0
+ Received  + 1
+   "ruleId": "existing-rule-id",
```

### Root cause

**(1) A CommonJS barrel load cycle.** `38449d3bc` ("send the buyer tax number on a fiscal receipt") added to `fiscalization/application/mappers/order-to-register-transaction-command.mapper.ts`:

```ts
import { decodeBuyerTaxIdColumn } from '@openlinker/core/orders';
```

That is a **value** import of the **main** orders barrel, which re-exports `OrdersModule`, whose module file value-imports `@openlinker/core/fiscalization`. So it closes `fiscalization -> orders -> fiscalization`.

Neither half is wrong on its own, and this is why it passed on every branch in isolation: the other half of the cycle (`orders.module` -> `sales-document-view.service` -> `@openlinker/core/fiscalization`, from #2516/#3186) lived on a different branch. The cycle only closed at the merge. Once it did, requiring the fiscalization barrel first left it partially initialised when `orders.module` was evaluated, so **both** `FISCAL_REGISTRATION_SERVICE_TOKEN` and `FiscalizationModule` were `undefined` at decoration time - which is exactly the two symptoms above, from one cause.

**(2) A stale expectation.** A tier-1 **rule** match has carried `ruleId` since #3186; only a country default or the Rest-of-world tier routes without one. The #3176 country-normalisation test matches a real rule, so its decision correctly carries `ruleId`. This is the same class of miss as the already-landed `5e8a54b8` ("assert the routed ruleId on the #3176 stray-case test"), one file over - that commit fixed `evaluate-sales-document-rules.spec.ts` and missed this sibling.

### What changed

Two files, 11 insertions, 1 deletion - `4ee5376c4`.

- The value import now comes from the `@openlinker/core/orders/types` **cycle-breaker sub-barrel**, which exists for precisely this and already serves the same symbol in `invoicing/application/services/auto-issue-trigger.service.ts`. `Order` / `OrderItem` stay on the main barrel because a type-only import erases and emits no `require` - the shape `order-to-sales-document-order-facts.mapper.ts` established, comment included.
- The #3176 expectation asserts `ruleId: existingRule().id`.

No test was skipped, no guard loosened, no type widened. The production behaviour of the mapper is unchanged - `dist/.../order-to-register-transaction-command.mapper.js` now emits `require("@openlinker/core/orders/types")`, and `dist/orders/types.js` exports the same `decodeBuyerTaxIdColumn`.

### Local evidence

Worktree at the exact CI SHA `dc4aa766`.

**Before** (fixes stashed) - reproduces CI exactly:
```
Tests:       26 failed, 35 passed, 61 total     <- CI reported 26 failed
```

**After:**

| Check | Result |
|---|---|
| the 2 CI-failing suites | `2 passed` / `61 passed, 61 total` |
| `libs/core` full jest | `423 passed, 423 total` / `6464 passed, 6464 total` (CI: 26 failed) |
| `apps/api` jest | `152 passed` / `2127 passed` |
| `apps/worker` jest | `57 passed` / `754 passed` |
| `libs/core` `tsc --noEmit` | clean |
| `pnpm -r --filter "./libs/**" build` | clean |
| `eslint` on both changed files | clean |
| `pnpm check:invariants` | exit 0 (incl. `check-cross-context-imports`, `check-types-sub-barrels`) |
| `barrel-purity.spec.ts` | pass |

One note for the stack: my first checkout was a stale local `3188-invoice-list-buyer-tax-id` that was 4 commits behind `origin`, so the work was redone against `dc4aa766`. Worth a `git fetch` + fast-forward before local runs on this chain.

norbert-kulus-blockydevs:
## This PR is marked merged in error - nothing of it reached `main`

GitHub flipped this to `MERGED` at 14:18 UTC because a bad push briefly made this branch's head an ancestor of its base. It was not a real merge, and it was my mistake, not a decision.

What happened: a cascade script I ran had two `git checkout` calls fail (the branches were held by other worktrees), and the loop carried on regardless - so a `git reset --hard` and a merge landed on the wrong branch and were pushed. GitHub saw this PR's head become reachable from its base and recorded the push commit as a merge commit.

Both branches were restored within minutes via `--force-with-lease` to the exact commits they held beforehand, confirmed against the remote-tracking reflog. **No work was lost**: this branch still carries its full content, and nothing from it is in `main` or in its base.

The residue is only the PR status. GitHub refuses to reopen a pull request it has recorded as merged:

```
422 Validation Failed - state cannot be changed. The pull request cannot be reopened.
```

So this PR cannot be revived, and a replacement one is needed for the same branch. The branch itself is current: it has been cascaded with `main` and every branch below it, and the stack top passes `@openlinker/core` at 426 suites / 6554 tests with `pnpm check:invariants` at exit 0.

Leaving this note so the merged badge is not read as "this shipped".

Issue #3061, opened 2026-09-10
feat(oms): empty/loading/error + unrecognised-rule locked states


norbert-kulus-blockydevs:
## Problem / Context

A rule this build doesn't recognise (`recognised: false`) must be listed but not editable, per the API's own contract; no other state has a defined UI yet either.

## Proposed Solution

Add `LoadingState`/`ErrorState`/`EmptyState` per the shared primitives, and render an unrecognised row with edit disabled + explanatory copy, delete-only, matching the reviewed mockup at https://claude.ai/code/artifact/b7b9ed28-9062-4985-8f0e-353b6fa1fcc5.

## Classification

**Type**: Frontend
**Layer**: Interface (component)
**File(s)**: apps/web/src/features/oms/components/*

## Dependencies

- Part of #3055

## Assumptions

- None

## Acceptance Criteria

- [ ] Empty, loading and error states use shared primitives
- [ ] An unrecognised rule cannot be edited, only deleted, with copy explaining why
- [ ] Tests added or updated for non-trivial logic
- [ ] No architecture boundary violations (CORE ↔ Integration)

Issue #3060, opened 2026-09-10
feat(oms): sourcing rules page reached from a /settings tile


norbert-kulus-blockydevs:
## Problem / Context

The table and dialogs from the prior sub-issues have no page to live on.

## Mockup

**Durable spec: `docs/plans/mockups/sourcing-rules-3055.html`** (on the epic branch `3055-sourcing-rules-admin-ui`).

`data-state` values this issue implements: `gallery-settings-tile`, `gallery-wrong-connection`, `gallery-no-locations`.

## Proposed Solution

**Decision (#3055 § Decisions): this is its own page reached from a `/settings` tile - NOT a section on the connection detail page.** The earlier "Proposed Solution" on the epic said connection-detail section; the mockup won.

Shape, mirroring the two existing precedents `InventoryLocationsTile` (#3063) and `WhoDecidesTile`:

- A `SourcingRulesTile` in `features/oms/components/`, mounted on `apps/web/src/pages/settings/settings-page.tsx` next to those two, plus its `toolbar-chip` in the summary group.
- A `sourcing-rules.route.tsx` + `pages/oms/sourcing-rules-page.tsx`, following the `who-decides-page.tsx` pattern.
- **No sidebar entry.** The tile is the only way in - stated in place on the `gallery-settings-tile` card. A back-link "&larr; Settings" and a two-level breadcrumb "Platform / Sourcing rules" both point back there.

Two page-level gates, both from the mockup:

1. **Wrong connection** (`gallery-wrong-connection`): the ruleset is per-connection, and only the OpenLinker OMS connection (`platformType === 'openlinker'`) can hold one - the backend already refuses authoring on anything else (#2953). The page states that and offers no editor, rather than letting an operator author rules nothing will ever read.
2. **No locations** (`gallery-no-locations`): rules choose *between* locations, so with zero locations there is nothing to rank. Blocking state with a link to the locations screen (#3063) / the bootstrap action.

## Classification

**Type**: Frontend
**Layer**: Interface (page + route + tile)
**File(s)**:
- apps/web/src/features/oms/components/sourcing-rules-tile.tsx
- apps/web/src/pages/oms/sourcing-rules-page.tsx
- apps/web/src/app/routes/sourcing-rules.route.tsx
- apps/web/src/pages/settings/settings-page.tsx

## Dependencies

- Part of #3055
- Stacked on #3061 (states), which is stacked on the rest of the epic
[...]

Issue #3059, opened 2026-09-10
feat(oms): delete/retire flow for sourcing rules


norbert-kulus-blockydevs:
## Problem / Context

There is no way to remove a rule, or to retire it (set `effectiveTo`) while keeping its history alongside a replacement.

## Proposed Solution

Add a `ConfirmDialog`-based delete action, and a lighter-weight retire action (`PATCH { effectiveTo: now }`) offered as an alternative in the same confirmation.

## Classification

**Type**: Frontend
**Layer**: Interface (component)
**File(s)**: apps/web/src/features/oms/components/sourcing-rule-dialog.tsx

## Dependencies

- Part of #3055

## Assumptions

- None

## Acceptance Criteria

- [ ] Delete removes the rule after confirmation
- [ ] Retire sets effectiveTo without deleting the row
- [ ] Tests added or updated for non-trivial logic
- [ ] No architecture boundary violations (CORE ↔ Integration)

PR #3228, opened 2026-09-14
feat(oms): ordered sourcing-rules table with arrow reorder (#3057)

### LINKED ISSUE #3057: feat(oms): ordered sourcing-rules table with arrow reorder

norbert-kulus-blockydevs:
## Summary

The ordered ruleset, in the order it runs: per-row reorder controls, the four display statuses, and the caption naming whichever rule currently sets the splitting limit.

**Stack position:** base is `3056-sourcing-rules-api-client` (#3226). #3058 stacks on this one.

## Reorder is arrow buttons, not drag-and-drop

Settled on #3055 § Decisions in the mockup's favour. `apps/web` ships no DnD library and this screen does not justify adding one; arrows are keyboard- and touch-operable with **no second code path**, whereas a DnD implementation would need exactly these buttons as its accessibility fallback anyway. The `⠿` glyph stays a **static indicator** (`aria-hidden`, `cursor: default`) so it cannot promise an interaction nothing backs - asserted by a test that no `[draggable="true"]` exists.

**The reorder body is EXHAUSTIVE and is built from the LIVE rows.** `PUT /order` refuses a subset and writes nothing, so `isLiveSourcingRule` reads the **dates** and never the display status. Both mistakes are reachable and each is wrong in a different direction: a SCHEDULED rule already claims its `(kind, name)` slot and is in the set, so `status === 'active'` would omit it and answer 409 `missingRuleIds`; and `recognised` does not enter the backend's predicate at all, so an unrecognised-AND-retired rule must read NOT live - which a display-status test cannot say, because `unrecognised` wins over the window there and the rule would be named as an unknown id.

**Nothing re-sorts the rows.** The API returns position-then-id order, which IS the router's evaluation order, so this is a plain `<table className="data-table">` rather than the sortable, virtualised `DataTable` - a client-side sort would be a second opinion about the one thing this screen exists to state.

**`quantity-split` never governs.** It is the permissive default, so naming the rule that declares it would attribute the *absence* of a restriction to a rule, and the operator would retire it expecting more splitting and see no change. A tie instead names EVERY rule, because an operator retiring one and expecting the limit to lift needs to see the other.

**Edit is REFUSED, not hidden, on an unrecognised rule.** The API rejects the patch outright (a successful edit would imply it routes). A disabled control that says why is what sends the operator to Delete; an absent one reads as a rendering bug. Delete stays enabled - it is that rule's only remedy.

## New invariant: `check-sourcing-rule-vocabulary-mirror`

The browser cannot import `@openlinker/oms` (#591), so `sourcing-rule-vocabulary.ts` re-declares four vocabularies plus the after-action permissiveness ranking. The script compares the vocabularies by **membership** (deliberately not order - they are an offer list and a copy-map key, and `check-ui-vocabulary` records why an unjustified gate is worse than none) and the ranking **by value**.

The ranking rule is the load-bearing one: swap two ranks in `@openlinker/oms` and both sides keep compiling, both `Readonly<Record<…, number>>` annotations stay total, and the table names the **wrong rule** as the one restricting splits - an operator then retires it expecting more splitting and nothing changes. Both drift classes were verified red before this landed (a swapped rank, and a renamed sort name).

Core's ranking is module-private, so the parser makes `export` optional: requiring it would make the rule silently unenforceable in exactly the state the code is in today.

## Test plan

- [x] `vitest run src/features/oms` - 9 files, 67 tests, all pass
- [x] `pnpm check:invariants` - all green, including the new mirror and `check-css-structure` over the appended block
- [x] Mirror script verified RED on a swapped ranking and on a renamed vocabulary value
- [ ] Manual: the table has no page to live on until #3060

### Two of #3057's ACs land three PRs up, not here

`Closes #3057` fires when this merges, so worth saying plainly where the rest
of that issue is delivered. **#3233 (#3061)** owns both:

- *"Arrow up/down reorder persists via the reorder mutation, sending the full
  live id list"* - asserted there against `{ ruleIds: ['b', 'a'] }`.
- *"A reorder conflict (409) is surfaced, not swallowed"* - asserted there too;
  #3229 additionally handles the `(kind, name)` 409.

This PR ships the component that builds the exhaustive body; the mutation call
and the 409 surface belong to the page that mounts it.

Refs #3055
Closes #3057


piotrswierzy (commented):
## Tech Lead review

The domain logic here is the strongest part of the stack. `isLiveSourcingRule` reads the dates and only the dates, and it matches `OmsRoutingRuleRepository`'s predicate byte-for-byte including the strict `>` (`rule."effectiveTo" IS NULL OR rule."effectiveTo" > :now`) — so the exhaustive `PUT /order` body is built from the set the server will compare against, not from the display status, and the docblock states both traps correctly (a SCHEDULED rule is in the set; an unrecognised-AND-retired one is not). Nothing re-sorts. The unrecognised row is refused-with-a-reason rather than hidden, and Delete stays enabled. `check-sourcing-rule-vocabulary-mirror.mjs` is exemplary — I ran both its `--self-check` and the real pass against the PR's mirror file and core's module-private `AFTER_ACTION_PERMISSIVENESS`, and it is non-vacuous (it parses core's unexported const, which is why making `export` optional was the right call). Four things need attention: a CSS custom property that does not exist, two acceptance criteria that no downstream issue picks up, an undeclared divergence from the committed mockup, and a keyboard focus drop on the reorder control this PR exists to ship.

### Findings

**[IMPORTANT]** — `apps/web/src/index.css` — `.rule-cell__kind-badge` and `.sourcing-rule-row--governing`
> Both set `background: var(--bg-subtle)`, and **`--bg-subtle` is not declared anywhere in the repository** (`grep -rn "bg-subtle"` over the whole tree returns nothing outside this PR; `index.css` declares `--bg-canvas`, `--bg-muted`, `--bg-shell`, `--bg-strong`, `--bg-surface*`). With no fallback both declarations resolve to nothing, so the governing-row highlight — the one *visual* affordance for the "sets today's splitting limit" story this PR is built around — silently does not render, and the kind badge loses its fill. `docs/frontend-architecture.md` § Design tokens states the drift guard is one-directional (catalog → CSS), so `check-design-tokens.mjs` is green on exactly this mistake and nothing else catches it. Use `--bg-surface-muted` (or declare the token in `index.css` first and add it to `tokens.ts`, per that section's *Adding a token* recipe).

**[IMPORTANT]** — PR body (`Closes #3057`)
> Two of #3057's acceptance criteria are not delivered and are not owned by any sibling issue. "Arrow up/down reorder **persists via the reorder mutation**, sending the full live id list" — the component only calls the `onReorder` prop; nothing calls `useReorderSourcingRulesMutation`. And "**A reorder conflict (409) is surfaced, not swallowed** — see `gallery-reorder-conflict`" — nothing here surfaces one. I checked #3058 (dialog), #3059 (delete/retire), #3060 (page + route + tile) and #3061 (empty/loading/error states): none of their ACs names either. On merge `Closes #3057` auto-closes the issue and the 409-surfacing obligation evaporates with it, which is how a state the mockup drew a dedicated gallery card for goes unbuilt. Either wire both here, or move the two ACs onto #3060 in the same change and say so in the body. (The empty state is correctly deferred — #3061 owns it explicitly — so no finding there.)

**[IMPORTANT]** — `apps/web/src/features/oms/components/sourcing-rules-table.tsx` + `lib/sourcing-rule.copy.ts`
> The committed mockup `docs/plans/mockups/sourcing-rules-3055.html:1436-1439` renders a second control group, `.rule-position__jump`, with **Move to top (⤒) / Move to bottom (⤓)**. This PR ships up/down only, adds no `.rule-position__jump` CSS, and neither touches the mockup nor explains the divergence — while leaving `COPY.moveToTop` / `COPY.moveToBottom` and the `'top'` / `'bottom'` arms of `move()` in the tree as dead code that reads like an unfinished wiring. `docs/frontend-architecture.md` § UX Mockups is explicit: an implementing PR "must either touch the mockup file, or say in its description why the mockup is still accurate without a change; a reviewer treats a PR that does neither as incomplete", and notes the obligation is reviewer-enforced because `check:invariants` cannot observe it. Either ship the jump buttons (the code is already written) or drop the mockup's group and the dead copy together.

**[IMPORTANT]** — `apps/web/src/features/oms/components/sourcing-rules-table.tsx:MoveButton` / `move()`
> Keyboard focus is dropped on the move that matters most. Press ▲ repeatedly to walk a rule to the top: on the final press the row reaches index 0, React re-renders, the button under focus becomes `disabled`, and focus falls to `<body>` — a keyboard-only operator loses their place mid-reorder, on the one interaction #3057's AC calls out by name ("Reorder works with keyboard only"). The fix is the standard one: key the row `<tr>` by `rule.id` (already done) and re-focus the sibling control after the move — or keep both buttons focusable and no-op at the boundary. There is also no assistive-tech announcement that anything moved; the mockup's `role="status" aria-live="polite"` toast (line 1194) is page-level so that half may legitimately be #3060's, but the focus loss is this component's and survives whatever the page does.

**[SUGGESTION]** — `scripts/check-ui-vocabulary.mjs` (`SCAN_ROOTS`)
> `features/oms` is not registered, and this PR ships the folder's first operator copy (`lib/sourcing-rule.copy.ts`). Both `features/fulfillment` and `features/bench` record the same rule verbatim in that file — "Added with the folder, not after it … a gate that arrives a wave later scans copy nobody re-reads". The banned list is not fulfilment-work-specific: `authority`, `holder` and `phase` are all live traps for a *sourcing*-authority surface. The copy is clean today, which is exactly the cheap moment to add the root.

**[SUGGESTION]** — `sourcing-rules-table.tsx:SourcingRulesTableProps` (`now = new Date()`)
> A default-parameter `new Date()` is re-evaluated on every render, so `resolveSourcingRuleStatus` and `resolveSplitCeiling` read a different instant each time while nothing schedules a re-render at a window boundary. Harmless today; a `useMemo(() => new Date(), [])` (or letting the page own the clock) makes the reads stable and the injection point single.

**[SUGGESTION]** — `sourcing-rules-table.tsx` — `<th title={COPY.stepHint}>` / `<th title={COPY.splittingHint}>`
> `title` on a `<th>` is not reliably announced, never appears on touch, and needs a hover-and-wait on desktop. `docs/frontend-architecture.md` records the same objection where it matters most ("leaving the distinction there makes it a comment rather than a behaviour"). The two hints are short enough to render as visible sub-labels, or to hang off the existing `Tooltip` primitive.

**Positive observations**
- `isLiveSourcingRule` is derived from the dates alone and its docblock names both directions of the trap. This is the single most breakable thing in the slice and it is right.
- The mirror script's Rule A / Rule B split — membership for the vocabularies, **by value** for the ranking — is the correct strength for each, and the self-check pins the deliberate non-rule (`tolerates a reorder by design`) so a future author cannot tighten it without reading why.
- `mostRestrictiveAfterAction` ignores an unrecognised after-action rather than treating it as most restrictive, and `resolveSplitCeiling` returns an empty `governingRuleIds` for the permissive default — so the table never attributes the *absence* of a restriction to a rule. Ties name every rule.
- `⠿` is `aria-hidden` + `cursor: default`, with a test asserting no `[draggable="true"]` exists. A static glyph that cannot promise an interaction nothing backs.
- Status badges carry text (`Active` / `Scheduled` / `Retired` / `Unrecognised`), not colour alone, per `docs/frontend-ui-style-guide.md`.
- `formatDate` shows an unparseable bound verbatim rather than `Invalid Date`, and `toTime` returns `null` rather than letting `NaN` compare false against everything.
- `.eslintrc.js` still lacks the `oms` slug — raised on #3226, noted here only because this PR is what gives the barrel real components to be deep-imported.

### Verdict
🔄 **Approve with changes** — `--bg-subtle`, the two orphaned ACs, the mockup divergence and the focus drop should land before merge; the rest are optional. (`gh` is authed as `piotrswierzy` and the author is `norbert-kulus-blockydevs`, so this posts as a comment rather than a blocking review.)

---

piotrswierzy (commented):
## Tech Lead re-verification (head unchanged at `b13ee81`)

Head has not moved and neither has `main`. Two of the four IMPORTANT findings resolve; two stand, and one of them got worse.

**`--bg-subtle` — stands, and widened.** I dumped `apps/web/src/index.css` at this head and grepped for a `--bg-subtle:` declaration: **there is none**, at this head, at the tip of the stack (#3235), or anywhere in `apps/web/src` on `main`. Meanwhile the file now *uses* it in two places — `.rule-cell__kind-badge`'s background and `.sourcing-rule-row--governing`'s. So the governing-rule row still has no visual affordance for the story this PR is built around, and a second consumer has since joined it. `scripts/check-design-tokens.mjs` is one-directional (catalog → CSS), so lint stays green over both. `docs/frontend-architecture.md` § Design tokens: declare in `index.css` first, then add the catalog entry. I re-read `sourcing-rules-table.tsx` at #3235's head — `sourcing-rule-row--governing` is still applied from `resolveSplitCeiling`, so nothing downstream rescued it.

**`Closes #3057` AC ownership — narrows to a nit.** Both ACs I called undelivered are now owned, by **#3233 (#3061)**: its `sourcing-rules-table.test.tsx` asserts `reorder` is called with the full live id list (`{ ruleIds: ['b', 'a'] }`) and asserts a 409 is surfaced rather than letting a row snap back unexplained. #3229 additionally handles the `(kind, name)` 409. So the work lands in the stack; what remains is that this PR's `Closes #3057` fires on merge while two of that issue's ACs are delivered three PRs later. Worth a line in the description pointing at #3233 rather than a code change.

**Keyboard focus on the final ▲ press — stands.** No focus restoration in `sourcing-rules-table.tsx` at this head or at #3235's; the only `focus` occurrence is an unrelated comment about `title` announcement.

**Mockup divergence — stands** as originally written; `docs/frontend-architecture.md` § UX Mockups makes this reviewer-enforced, so an implementing PR either touches the mockup or says in its description why it is still accurate.

**Stack position:** 3rd in the OMS stack, based on #3226 (which is itself rooted on `3055-sourcing-rules-admin-ui`, a branch with **no open PR**). Five PRs sit above it. Blocked on #3226's own outstanding finding.

**CI:** no checks reported on this branch — `ci.yml` is `main`-targeted only. Absent, not passing.

### Findings
Two stand (`--bg-subtle`, focus restoration), one narrows to a description nit (AC ownership → #3233), one unchanged (mockup). Declaring `--bg-subtle` in `index.css` + `tokens.ts` is the one that actually changes what an operator sees.

### Verdict
🔄 **Approve with changes**

---

norbert-kulus-blockydevs:
All four standing findings are addressed on `3257a96cc` + `6589a7009`, plus all three suggestions.

**`--bg-subtle` (IMPORTANT)** - fixed. Both usages now read `--bg-surface-muted`, which is already in the catalog and already carries a light and a dark definition. I took that over declaring a new token: the fix needed is a working fill on both sites, and minting a token means defending a new colour in two themes for nothing. `node scripts/check-design-tokens.mjs` reports 111 tokens all declared, and the only remaining `bg-subtle` hit in `index.css` is the comment explaining the change.

**Keyboard focus on the final press (IMPORTANT)** - fixed. After the reordered set renders, focus goes back to the arrow that was pressed; when that arrow has become the end-stop it goes to the row's opposite arrow instead. It is keyed on the RENDERED order (`liveIds.join()`), not on the click, because the reorder is a server round-trip - restoring at click time would put focus back on a button that is about to be disabled. A failed reorder leaves the order unchanged, the effect never runs, and focus stays where the browser already had it.

Verified red-first: with the effect body short-circuited the new end-stop test fails, and passes with it restored. The companion test ("keeps focus on the pressed arrow while the row still has somewhere to go") passes without the effect too - the browser leaves focus on a button that stays enabled - so it carries a comment saying it guards against an over-reaching fix rather than proving the effect exists.

**Mockup divergence (IMPORTANT)** - shipped the jump buttons rather than dropping them. `move()`'s `top`/`bottom` arms and `COPY.moveToTop`/`moveToBottom` were already written, so this closes the dead code and the divergence together, and it keeps the design of record intact. One deliberate deviation, stated in the CSS: the mockup reveals both control groups only on `:hover`/`:focus-within`, and they are permanently visible here - a ruleset is reordered from a touch device as readily as from a mouse, and a hover-revealed control does not exist on one. Two tests cover the jump path (full live list on the wire, and focus landing on the opposite arrow).

**`Closes #3057` AC ownership** - taken as the description nit you narrowed it to. Added a line to the PR body pointing the reorder-persistence and 409-surfacing ACs at #3233, which owns both.

**SUGGESTION - `features/oms` missing from `check-ui-vocabulary` scan roots** - added, with the folder's first operator copy rather than a wave later. The trial run proves it non-vacuous and clean: 168 files across 7 of 7 roots, up from 163 across 6.

**SUGGESTION - `now = new Date()` re-evaluated per render** - fixed as suggested. The prop is optional and the fallback is `useMemo(() => new Date(), [])`, so the status and ceiling reads share one instant and the injection point stays single.

**SUGGESTION - `title` on `<th>`** - fixed, in the visible-text direction. Both hints now render in a real `<caption>` above the table rather than as two `title` attributes. That also labels the table natively, which is why the `aria-label` is gone rather than kept beside it. I did not render them as per-column sub-labels: at 96px and 140px those two sentences wrap to three or four lines each and the header grows taller than the rows it heads.

Gate: `pnpm --filter @openlinker/web type-check` clean, `eslint` on both changed components clean, `pnpm check:invariants` exit 0, and all 9 `features/oms` spec files green (71 tests).

piotrswierzy (approved):
## Tech Lead re-review (`b13ee81` → `6589a70`)

All four IMPORTANT findings are closed, and both SUGGESTIONs were taken as well. Five files changed and every one of them answers something from the standing review.

**`--bg-subtle` — closed, and closed at both sites.** I dumped `apps/web/src/index.css` at this head and at the stack tip (#3235, `523fb4a`) and grepped each: **zero** `var(--bg-subtle)` uses remain and the only `bg-subtle` occurrence in the file is the comment explaining the fix. `.rule-cell__kind-badge` and `.sourcing-rule-row--governing` both now paint `var(--bg-surface-muted)`, which is genuinely declared (two definitions — light and dark). The comment states the mechanism correctly, including why `check-design-tokens.mjs` stayed green over it. Reusing a catalogued token rather than minting one is the right call: a new token needs a colour defended in two themes for a fill that already exists. So the governing-row highlight — the one visual affordance for the story this table exists to tell — now actually renders.

**Mockup divergence — closed by shipping, not by dropping.** `.rule-position__jump` is now a real control group: ⤒ / ⤓ wired to the already-written `'top'` / `'bottom'` arms of `move()`, `COPY.moveToTop` / `COPY.moveToBottom` are live rather than dead, and the CSS block exists. The one remaining departure from the mockup is deliberate and documented in place: the mockup reveals both groups on `:hover`/`:focus-within` and this ships them permanently visible, because a hover-revealed control does not exist on a touch device. That reasoning belongs in the description as well as the CSS per § UX Mockups, but the substance is right and the code carries it.

**Focus restoration — closed, and the implementation is the careful version.** The effect keys on `renderedOrder` (the joined live id list) rather than on the click, with a docblock stating exactly why: the reorder is a server round-trip, so at click time the row has not moved and restoring focus there would land on a button about to be disabled. A failed reorder leaves the order unchanged, the effect never fires, and focus stays where the browser had it — which is the correct degradation. `FOCUS_FALLBACK` maps each direction to its opposite so the end-stop press hands focus to the row's other arrow. The end-stop test drives a real stateful harness and asserts `toHaveFocus()` on the sibling; the companion "stays on the pressed arrow" test is annotated as passing without the fix, which is the honest thing to say about a test that guards against over-reach rather than proving the effect. `move()`'s `top`/`bottom` arithmetic is correct (`next` is spliced before the target is computed, so `to === 'bottom'` → `next.length` is the tail).

**`Closes #3057` AC ownership — closed.** The description now names #3233 explicitly for both ACs, and I re-verified them there (`2750231`): `sourcing-rules-section.test.tsx:116` asserts `reorder` called with `{ ruleIds: ['b', 'a'] }`, and `:120` asserts a 409 is surfaced "rather than letting a row snap back unexplained", dismissed by hand rather than on the next settled read.

**Both SUGGESTIONs taken too.** `features/oms` is now a `check-ui-vocabulary` scan root with a comment recording that `authority` / `holder` / `phase` are live traps on a sourcing-authority surface, added while the folder is clean. `now` is a `useMemo(() => new Date(), [])` with the `nowProp ?? mountedAt` seam kept. And the two `<th title=…>` hints became a real `<caption>` with two hint spans, with the `aria-label` correctly removed rather than left to compete with the native label.

### Findings

**[SUGGESTION]** — `apps/web/src/index.css` — `.rule-cell__kind-badge` / `.sourcing-rule-row--governing`
> Both now resolve to `--bg-surface-muted`, so on the governing row the kind badge's fill matches its own row background exactly. It stays legible on its `1px solid var(--border-default)` border, so this is cosmetic — worth a glance at `/dev/ui` or the page once #3234 mounts it, not a change on its own.

**[SUGGESTION]** — `sourcing-rules-table.tsx` (`registerArrow`)
> The ref callback is a fresh closure each render, so React detaches (`null` → delete) and re-attaches every arrow on every render. Harmless — effects run after refs are attached, so the map is populated when the focus effect reads it — but a `useCallback`-free map keyed off `rule.id` is one of the few places where a future "why is focus flaky" investigation would start.

> § UX Mockups asks the divergence to be stated in the description, not only in the CSS. One line — "both control groups ship permanently visible rather than hover-revealed, because a hover-revealed control does not exist on touch" — discharges it.

**Positive observations**
- The focus effect keying on rendered order rather than on the click is the non-obvious correct choice, and the docblock explains the failure mode of the obvious one.
- The jump-group CSS comment records the mockup divergence *and* its reason at the point of divergence, which is where the next reader will be standing.
- `check-ui-vocabulary` scan root added "after a trial run proved the folder is already clean" — registering a gate you have verified passes, rather than one you hope does.
- Four tab stops per row is a real density cost and it is the mockup's own shape; the arrows staying permanently visible is what makes them reachable at all on touch.

**Stack position:** 3rd in the OMS stack. Its base #3226 is now clean, but the root branch `3055-sourcing-rules-admin-ui` **still has no PR** (`gh pr list --state all` on that head returns empty), so the stack has no route to `main`.

**CI:** no checks reported on this branch — absent, not passing. `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN` against #3226.

### Verdict
✅ **Approve**

---

PR #3229, opened 2026-09-14
feat(oms): create/edit sourcing-rule dialog (#3058)

### LINKED ISSUE #3058: feat(oms): create/edit sourcing-rule dialog (filter/sort, after-action, priority list, effective window)

norbert-kulus-blockydevs:
## Summary

The only place a sourcing rule is authored: kind, a kind-filtered name, the optional priority-location order, the splitting rung, and an optional effective window.

**Stack position:** base is `3057-sourcing-rules-table` (#3228). #3059 stacks on this one.

## Five properties that carry weight

**`kind` is LOCKED while editing, and the note says why.** `UpdateSourcingRuleDto` carries `position` / `name` / `afterAction` / `priorityLocationIds` / `effectiveFrom` / `effectiveTo` and deliberately **not** `kind`, because kind is half a rule's identity under the duplicate-detection index. An editable control would offer a change the PATCH cannot carry — the operator would watch a save succeed and change nothing.

**A name a LIVE sibling already claims is offered DISABLED and names that rule**, rather than hidden. Only one rule per `(kind, name)` may be live and the API answers 409; removing the option would leave an operator hunting for a capability the product appears not to have. "Live" is the BACKEND's not-retired predicate, so a **scheduled** sibling holds the slot too — asserted in both directions (scheduled blocks, retired does not).

**Changing the kind resets the name and clears the priority list**, because the name vocabulary is kind-specific and a stale pair is one the server refuses.

**The splitting preview recomputes #3057's ceiling over the ruleset with this draft swapped in**, so it states the RULESET's answer rather than this rule's own value. A draft that would allow MORE splitting than today is **confirmed**; stricter and unchanged save straight through, since an extra click on a safe direction trains people to click through the unsafe one.

Worth knowing for anyone writing a test here: **only an EDIT of the governing rule can loosen the ceiling.** Adding a permissive rule beside a strict one changes nothing, because the strictest rung still governs — so a create-against-a-strict-sibling test would be naming a state it can never reach. The first draft of this PR's test did exactly that and went red, which is recorded in the test's own comment.

**`position` is never an input.** Order belongs to the table's arrows and to `PUT /order`, so a create appends and an edit keeps its own position — this dialog can never silently renumber a ruleset nobody was looking at.

## Also here

- The priority picker is an ORDERED list with the same arrow controls the table uses (#3055 § Decisions) — no DnD, no drag affordance. A selected id with no matching location renders as a **stale entry** rather than being dropped: the location was deleted after the rule was saved, it ranks nothing, and removing it silently would change what the rule does on the next save.
- **Locations arrive as a PROP**, not fetched here. The page owns the no-locations gate (#3060), and duplicating that judgement would give it two homes.
- The window rule is `to > from` **strictly** — equal bounds describe a rule that is never in force, which is the class of defect this screen exists to make impossible. A bare date is anchored at **midnight UTC**, never the browser's zone, so two operators in different offices cannot store different instants from the same form.
- Server refusals render verbatim through `describeSourcingRuleError` (#3056): this endpoint's 400s name the actual problem — not an OMS connection, a kind/name pair the router cannot evaluate, an unknown location id — and a generic sentence would throw that away.

## Test plan

- [x] `vitest run src/features/oms` - 11 files, 92 tests, all pass
- [x] `pnpm check:invariants` - all green, including `check-css-structure` over the appended block
- [ ] Manual: the dialog has no page to open it from until #3060

Refs #3055
Closes #3058


piotrswierzy (commented):
## Tech Lead review

The authoring dialog is well-reasoned and the four contract properties it claims all hold against the #2953 backend: `kind` is locked on edit and genuinely absent from the PATCH body (asserted); `position` is never an input, so the dialog cannot silently renumber a ruleset; "live" for the claimed-name check reads `isLiveSourcingRule` (the backend's not-retired predicate) rather than the display status, and is asserted in both directions; and server refusals pass through `describeSourcingRuleError` verbatim. Every acceptance criterion on #3058 is delivered. Three things need attention before merge: the splitting preview — the one operator-facing statement this dialog makes about the *ruleset* — silently ignores in-dialog edits to the effective window, and it fails in the unsafe direction; the priority picker's accessible name is not wired, which the one test covering its visibility hides by asserting vacuously; and the stack still carries `var(--bg-subtle)` against a custom property that is declared nowhere.

### Findings

**[IMPORTANT]** — `apps/web/src/features/oms/components/sourcing-rule-dialog.tsx:~300` (the `draft` `useMemo`)
> `draft` reads `form.getValues('effectiveFrom')` / `('effectiveTo')`, but neither field is watched and neither is in the dep list (`[rule, connectionId, rules.length, kind, name, afterAction, priorityLocationIds, form]`). With RHF's default `onSubmit` mode and no `watch` on those fields, typing in either date input causes **no re-render at all**, so the memo never recomputes and the draft keeps whatever window was present the last time `kind` / `name` / `afterAction` changed.
>
> That matters because `resolveSplitCeiling` counts only rules whose `resolveSourcingRuleStatus` is `active`, i.e. it reads the window. So the preview Alert and the loosening gate are computed against stale dates, and the miss is in the unsafe direction: an operator editing a **scheduled** permissive rule who clears `effectiveFrom` (making it active now, genuinely raising the ceiling) gets **no confirm** — the memo never recomputes, `loosens` stays false, and `save()` runs straight through. The mirror case is a false alarm: setting a far-future start on a loosening change still fires the confirm for something that will not happen today. Either way the Alert states a ruleset answer that is wrong about the operator's own configuration, which is exactly what the PR describes as the preview's whole job ("it states the RULESET's answer rather than this rule's own value").
>
> Fix: `const effectiveFrom = form.watch('effectiveFrom')` / `effectiveTo`, use those in the memo body, and add them to the dep list. A test that sets a future `effectiveFrom` on a loosening edit and asserts the confirm does *not* fire — then clears it and asserts it does — would pin both directions.

**[IMPORTANT]** — `apps/web/src/features/oms/components/sourcing-rule-dialog.tsx:~430` (the `priorityLocationIds` `FormField`) and `sourcing-rule-dialog.test.tsx:~215`
> `FormField` clones its single child with `id` / `aria-describedby` / `aria-invalid` and renders `<label htmlFor={inputId}>`. Here the child is `PriorityLocationPicker`, which spreads those onto a plain `<div className="priority-picker">`. A `<div>` is not a labelable element, so the `for` never resolves: the "Location order (highest priority first)" label is orphaned, the group has no accessible name, and the `aria-describedby` carrying both the description and the `FieldError` ("Add at least one location — a priority list with nothing in it ranks nothing") is attached to an element with no role, so it is never announced. `FormField`'s own docblock names this trap ("Wrappers that do not forward these props will silently break the accessibility wiring"), and `docs/frontend-ui-style-guide.md` requires focus/label wiring on interactive groups.
>
> The test hides it. `expect(screen.queryByLabelText(/Location order/)).toBeNull()` is **vacuous**: `queryAllByLabelText` returns `[]` for a label whose `for` points at a non-labellable element (`@testing-library/dom` `label-text.js:106` — the message is only thrown on the `get*` path), so that assertion passes identically whether the field is rendered or not. The positive case already had to fall back to `getByText`, which is the symptom.
>
> Fix: put `role="group"` on the picker and label it via `aria-labelledby` (or `aria-label`), or make the "Add a location…" `Select` the field's labelled control and give the ordered list its own labelled region. Then switch the visibility test to a query that can actually fail.

**[IMPORTANT]** — `apps/web/src/index.css` (context line at the top of the appended block: `.sourcing-rule-row--governing { background: var(--bg-subtle); }`)
> `--bg-subtle` is declared **nowhere** in `index.css` at this PR's head SHA (`grep -c '^\s*--bg-subtle:'` → 0, both light and dark blocks), and it is not in `shared/theme/tokens.ts` either. `scripts/check-design-tokens.mjs` is one-directional (catalog → CSS), so a `var()` naming an undeclared property passes lint and renders as no background at all — the "sets today's splitting limit" row highlight is invisible. This is the same invisible-UI failure two earlier PRs in this programme shipped.
>
> The line belongs to the base branch (#3228), not to this diff, so the fix may land there — but this stack is what ships it. Either declare `--bg-subtle` in both `:root` and the dark block and add it to `tokens.ts`, or repoint the rule at an existing token (`--bg-muted` / `--bg-shell` are already declared).

**[SUGGESTION]** — `sourcing-rule-dialog.tsx:~250` (`defaultValues`) and the `kind` `onChange`
> Both seed `name` with `NAMES_BY_KIND[kind][0]` unconditionally. On a connection that already has a live `filter` / `in-stock` rule — the single most likely state, since it is the first option — the create dialog opens with a **disabled option selected**, and because browsers skip disabled options the operator cannot navigate away and back to it. Saving 409s, which `describeSourcingRuleError` renders verbatim, so it is recoverable and explained; but seeding with the first *unclaimed* name (falling back to index 0 when every name is claimed) would make the common path work without the round trip. Same for the kind-change reset.

**[SUGGESTION]** — `sourcing-rule-dialog.tsx:~305` (`draft`)
> `recognised: true` is hardcoded on the draft. Correct today only because `SourcingRulesTable.renderEdit` disables the edit button for `rule.recognised === false` (#3228) — the contract that an unrecognised row is listed, flagged and **not** patchable is enforced one component up. Worth a line in the props docblock saying the caller owns that gate, so a future mount point does not reopen it; the ceiling preview would silently count an unrecognised rule as governing.

**[SUGGESTION]** — `.eslintrc.js`
> `features/oms` is still absent from both `no-restricted-imports` pattern groups (`apps/web/src/features/**` and `apps/web/src/plugins/**`) for all five canonical subdirectories. Per `docs/frontend-architecture.md` § Feature Public Surface the rule silently fails open for an unregistered slug. I raised this on #3226, which created the feature; this PR widens the barrel by three more exports, so the unguarded surface keeps growing. Registering it is a one-line-per-subdirectory edit wherever the stack finds it convenient.

**Positive observations.** `isLiveSourcingRule` rather than `resolveSourcingRuleStatus` for the claimed-name check, asserted in both directions (scheduled blocks, retired does not) — this is the trap that helper's docblock was written for, and the tests name it. The stale-priority-entry treatment (render and mark rather than silently drop) is the right call for the same reason the returns surfaces never render a positive claim from an absent value. The `ConfirmDialog` is wired with `initialFocus="cancel"` plus the elevated overlay classes, both of which exist. And the test comment recording *why* a create-against-a-strict-sibling test can never see the state it names is the kind of note that stops the next author re-adding a test that cannot fail.

### Verdict
🔄 **Approve with changes**

Priority order: (1) watch the effective-window fields so the loosening gate cannot be bypassed, (2) wire the priority picker's accessible name and replace the vacuous visibility assertion, (3) resolve `--bg-subtle` somewhere in the stack.

Note on CI: this PR targets `3057-sourcing-rules-table`, and the workflow triggers on `pull_request: branches: [main]` only — **no checks ran**, so the green `MERGEABLE` / `CLEAN` state says nothing about lint, type-check or tests. The PR body's local runs are the only evidence.

---

norbert-kulus-blockydevs:
Review addressed in `bdd538567` (plus a merge of the updated base, `636cca34a`).

**[IMPORTANT] the `draft` `useMemo` read the window through `getValues` - fixed.**
`effectiveFrom` / `effectiveTo` are now `form.watch(...)` and are in the dep list; `form` is out of it. Two tests pin both directions, and both fail against the old code (verified by reverting the fix and re-running):

- `CONFIRMS when an end date would retire the governing rule and lift the ceiling` - the unsafe arm. One correction to the finding's worked example, though: the scenario as written (editing a *scheduled permissive* rule and clearing `effectiveFrom`) cannot loosen. `resolveSplitCeiling` takes the most restrictive of the active set, so *adding* a rule to that set is monotonically stricter-or-equal, and a `quantity-split` draft restricts nothing either way. The case that really does loosen through the window is the mirror one: the edited rule is the *active governing strict* rule, and the window takes it OUT of the active set - a past `effectiveTo`, or a future `effectiveFrom`. That is also the case the "Ends on" field's own description invites ("Setting this to a past date retires the rule while keeping it for history"). So the test types a past end date on the governing `no-split` rule and asserts the confirm fires and `update` is not called until it is accepted.
- `recomputes the splitting preview when the start date changes` - the preview arm, in the stricter direction. A scheduled `no-split` rule reads "splitting stays the same overall"; clearing its start date makes it active and the sentence has to flip to "makes splitting stricter overall". Stale, it stays on the first sentence.

For the same reason I did not add the "far-future `effectiveFrom` on a loosening edit should NOT confirm" test: pushing the governing rule into the future genuinely does lift today's ceiling, so the confirm there is correct rather than a false alarm.

**[IMPORTANT] the `priorityLocationIds` FormField had no accessible name - fixed.**
`PriorityLocationPicker` now renders `role="group"` and takes an `aria-label`, supplied from a single `PRIORITY_FIELD_LABEL` constant that is also the `FormField` `label`, so the visible label and the group's name cannot drift. With a role on the element, the cloned `aria-describedby` (description + `FieldError`) is announced. I took the `aria-label` branch rather than `aria-labelledby` because `FormField` puts no `id` on its `<label>` and does not forward `aria-labelledby`; adding either would change a shared primitive every form in the app renders, which is wider than this PR. The residue is that the `<label for>` stays inert - it is now redundant visible text rather than the only naming mechanism.

The vacuous assertion is gone. The test queries `getByRole('group', { name: /Location order/ })` in both arms and additionally resolves `aria-describedby` back to its text, and it fails against the old markup (verified).

**[IMPORTANT] `var(--bg-subtle)` - already fixed one level down, nothing to do here.**
It landed on the base branch in `3257a96cc`, which this PR now carries: both usages point at the catalogued `--bg-surface-muted`, and `grep -rn 'bg-subtle' apps/web/src` returns only the comment recording the fix at `apps/web/src/index.css:23550`. No new token was declared.

**[SUGGESTION] seeding with a claimed name - fixed.**
`firstUnclaimedName(kind, claimed)` seeds both `defaultValues` and the kind-change reset, falling back to index 0 when every name is claimed (the control needs a value, and there the 409 is the honest answer). `claimedNames` moved out of the component into a pure `claimedNameRows(rules, excludeId, now)` so the default and the memo share one definition. New test `seeds a create with the first name no live sibling has claimed`; it fails against the old seeding.

**[SUGGESTION] the caller owns the `recognised` gate - documented.**
The `rule?` prop docblock now says #3057's table disables Edit for `recognised === false` and that this dialog trusts it, naming the consequence: the draft is `recognised: true` unconditionally, so an unrecognised rule mounted here would be counted as governing the ceiling.

**[SUGGESTION] `features/oms` missing from the `no-restricted-imports` groups - already fixed one level down.**
It landed on `3056-sourcing-rules-api-client` in `2d93a01f7` and reached this branch through the base merge; both pattern groups now carry the five `**/oms/<part>/**` entries (`.eslintrc.js:322` and `:601`).

Local gate, run in this worktree: `npx vitest run src/features/oms` (11 files, 95 tests, green), `pnpm --filter @openlinker/web type-check` (clean), `npx eslint` on both changed files (clean).

piotrswierzy (approved):
## Tech Lead re-review (`481d9636` → `494b68b7`)

All three IMPORTANT findings from the previous pass are closed, and each is closed at the level that matters rather than cosmetically. `effectiveFrom` / `effectiveTo` are now `form.watch`ed with the reasoning recorded in place, and the `draft` memo lists both in its dep array — so the splitting preview and the loosening confirm genuinely follow in-dialog window edits. Two new tests cover both directions of that (a past end date on the governing rule now raises the confirm; clearing a future start recomputes the preview to "stricter"), and they would fail against the `getValues` implementation. The `PriorityLocationPicker` now carries `role="group"` + an explicit `aria-label`, and — importantly — **the test that was vacuous is now able to fail**: it queries `getByRole('group', { name: /Location order/ })`, which throws if either attribute is dropped, and then resolves every id in the cloned `aria-describedby` against the document and asserts the description text actually comes back. `--bg-subtle` is gone; both sites now use the catalogued `--bg-surface-muted`, with a comment explaining why a new token was not minted. I re-ran a full referenced-vs-declared sweep of `index.css` at this head and every custom property and every class these components name resolves.

The table work that came in alongside is a genuine improvement and correctly argued: the `<caption>` replacing two `title` attributes (a `title` is not announced, never appears on touch), the one-instant-per-mount `now` replacing a default-parameter `new Date()` that drifted between the status and ceiling reads, and focus restoration keyed on the *rendered* order rather than the click — with the right note that a failed reorder leaves focus alone. Registering the `oms` slug in both `no-restricted-imports` groups closes the fail-open deep-import gap for this feature.

### Findings

**[SUGGESTION]** — `apps/web/src/features/oms/components/sourcing-rule-dialog.tsx:528`
> `aria-invalid` is forwarded onto the `role="group"` element. `aria-label` and `aria-describedby` are global and land correctly; `aria-invalid` is not defined on `group` in ARIA 1.2, so it is likely inert there. The validation message still reaches the operator through `aria-describedby`, so nothing is lost — but if the intent was for the group to announce as invalid, that needs to sit on a control that supports it rather than on the wrapper.

### Verdict
✅ **Approve**

---

Issue #2231, opened 2026-08-20
[BUG] Integration + Frontend — Erli says why an offer cannot sell; OpenLinker drops it and shows Draft


norbert-kulus-blockydevs:
## Problem / Context

When Erli refuses to sell an offer it says why, in a field OpenLinker never reads. The operator sees the word `Draft` and nothing else.

Design underlay (dark theme, real tokens and components, every surface below rendered before/after): https://claude.ai/code/artifact/574f9dd5-92ac-47b3-bbc3-00d7f8dd08c4

Three separate defects, one cause.

**1. The reason is dropped at the adapter.** `mapErliStatusToReadResult` (`libs/integrations/erli/src/infrastructure/adapters/erli-offer-manager.adapter.ts:1133`) reads `product.status` and nothing else. `ProductResponse.buyableProblems` - an 18-value enum naming exactly why the offer cannot be bought - has no reference anywhere in `libs/integrations/erli`. Neither does `archived`.

```
buyableProblems: delivery | stock | active | category | image | shop-activity |
  shopKyc | missingPrice | minPrice | maxPrice | terms | blocked | condition |
  archived | translations | name | missingTaxRate | marketInactive
```

(`docs/architecture/adrs/erli-sandbox-swagger.json`, `ProductResponse.properties.buyableProblems`.)

**2. Two branches of that mapper are dead, and one field does not exist.** `ErliProductStatus` (`erli-product.types.ts:190`) declares `'accepted' | 'active' | 'inactive' | 'rejected'`, still carrying its `PROVISIONAL (#992)` comment. Erli's schema declares `status` as `enum ["active","inactive"], nullable`. `accepted` and `rejected` are unreachable, and `statusReason` - read at `erli-offer-manager.adapter.ts` in the `rejected` branch - is not a property of `ProductResponse` at all.

Consequence: `validationErrors` is only ever populated from the unreachable `rejected` branch, so it is always `[]`. `resolveOfferLifecycle` splits `inactive` on exactly that list (`offer-lifecycle.types.ts`: `facts.hasValidationMessages ? 'Invalid' : 'Draft'`), so **the `Invalid` tab can never contain an Erli offer** and every blocked one lands in `Draft` - the bucket whose own docblock promises "not live, with nothing to say about why".

[...]

PR #3230, opened 2026-09-14
feat(oms): delete/retire flow for sourcing rules (#3059)

### LINKED ISSUE #3059: feat(oms): delete/retire flow for sourcing rules

norbert-kulus-blockydevs:
## Summary

Two acts behind one confirmation, because an operator reaching for "remove this rule" usually wants the reversible one and does not know it exists.

- **Retire** — `PATCH { effectiveTo: <now> }`. The row survives with its history, stops being evaluated, and can coexist with its replacement, which is exactly what the partial duplicate-detection index is for.
- **Delete** — a HARD delete including history, in the danger tone.

**Stack position:** base is `3058-sourcing-rule-dialog` (#3229). #3061 stacks on this one.

**It is deliberately NOT a `ConfirmDialog`.** That primitive takes one confirm and one cancel; a third, differently-toned action is the whole point here, and widening a primitive every other confirm in the app uses — for one caller — is the wrong trade.

**Retire is HIDDEN with a reason where the API would refuse it.** An already-retired rule has nothing to retire, and a rule this build does not recognise cannot be patched at all (the API refuses it, because a successful edit would imply it routes). A button that answers 400 is worse than saying why it is absent, so both cases render the reason and leave Delete as the only action.

**Deleting the rule that currently limits splitting is called out** — the one deletion whose consequence is invisible from the row. The warning is derived from #3057's ceiling over the ruleset **without** this rule, so it states the ruleset's answer rather than this rule's own value, and stays silent when a sibling holds the same limit (asserted both ways).

**The retirement instant is the moment the operator asked, never back-dated.** A retirement is a claim about when the rule stopped applying, and back-dating it would assert something that did not happen. A clock running behind the server's can leave the rule live for those few seconds — the honest cost of not lying about the time, recorded in place rather than papered over.

## Also here

- A refusal renders **verbatim** and leaves the dialog open, so the operator can read it and choose the other action.
- The footer splits: the reversible alternative on the left, away from the destructive pair on the right.
- The confirm's subtitle ("Every rule below it moves up one step") is true of what the operator **sees**: #3057's table renders the ordinal within the live set, not the stored `position`. The API itself leaves a gap in the position sequence on delete — `PUT /order` is what closes it — so the sentence is about the visible numbering, deliberately.

## Test plan

- [x] `vitest run src/features/oms` - 12 files, 100 tests, all pass
- [x] `pnpm check:invariants` - all green, including `check-css-structure` over the appended block
- [ ] Manual: nothing opens this dialog until #3060

Refs #3055
Closes #3059


piotrswierzy (commented):
## Tech Lead review

A small, well-shaped component that gets the central distinction right: retire (`PATCH { effectiveTo }`) and delete (hard, history included) are two different acts, they are presented as two differently-toned actions behind one confirmation, and the tests assert the retire path patches a body whose only key is `effectiveTo` and never calls `remove`. Declining to widen `ConfirmDialog` for a third action is the right trade, the split footer is built from CSS that actually exists (`dialog__footer`, `dialog__body`, `muted-text`, `button--danger`, `--space-2` all verified declared), `Button` already defaults to `type="button"` so the footer buttons are safe outside a form, and the early `return null` sits correctly after the hooks. Both of #3059's acceptance criteria are delivered. Two findings need attention: the "cannot be retired" guard misses the one remaining case where the API really does refuse the PATCH, and the splitting-limit warning attributes a consequence to Delete that Retire produces identically.

### Findings

**[IMPORTANT]** — `apps/web/src/features/oms/components/sourcing-rule-delete-dialog.tsx:~229` (`retireUnavailableReason`)
> The guard covers `!recognised` and `status === 'retired'` but not **`scheduled`**. `SourcingRuleAdminService.update` validates the MERGED window — `assertEffectiveWindow(patch.effectiveFrom ?? existing.effectiveFrom, patch.effectiveTo ?? existing.effectiveTo)` (`apps/api/src/oms/application/services/sourcing-rule-admin.service.ts:129-132`) — and `assertEffectiveWindow` throws `BadRequestException` when `to <= from` (`:224-231`). So for a rule with `effectiveFrom` in the future, `PATCH { effectiveTo: new Date().toISOString() }` is always `to < from` and always answers 400: *"effectiveTo must be after effectiveFrom — a rule whose window closes before it opens is never evaluated."*
>
> That is precisely the case this component's own docblock says it exists to prevent ("Offering a button that answers 400 is worse than saying why it is absent"), and it is the one an operator is most likely to hit, since a scheduled rule is exactly the kind you change your mind about before it starts. The error does render verbatim, so it is recoverable — but the operator is told about a start date they never touched in this dialog.
>
> It is also inconsistent with the sibling PR: #3229's `sourcingRuleFormSchema` refuses `effectiveTo <= effectiveFrom` client-side on the grounds that such a rule "would save happily and never fire". This component constructs that exact pair.
>
> Fix: a third arm in `retireUnavailableReason` keyed on `resolveSourcingRuleStatus(rule, now).status === 'scheduled'` — the honest reason is that the rule has not started yet, so there is nothing to retire, and deleting (or clearing its start date) is the remedy. `resolveSourcingRuleStatus` already returns the value, so it is one branch and one sentence. Add a test beside the two that already pin the other arms.

**[IMPORTANT]** — `sourcing-rule-delete-dialog.tsx:~318` (the `removalLoosens` Alert)
> `removalLoosens` compares `resolveSplitCeiling(rules)` against `resolveSplitCeiling(rules.filter(r => r.id !== rule.id))`, which is correct — but the consequence is identical for **both** actions, and the copy names only one: *"**Deleting** this removes your only limit on splitting."* `resolveSplitCeiling` counts only rules whose status is `active`, and retiring writes `effectiveTo: now`, which makes `to <= nowTime` → `retired` → excluded from the same computation. So an operator who reads the warning and then reaches for the button that sounds safer gets exactly the loosening the warning described, with the copy having pointed at the other button.
>
> This is the same hazard #3229 treats seriously enough to gate behind a modal confirm. Here it does not need a second confirm — but the sentence should name the removal rather than the deletion ("Removing this rule — by either route — lifts your only limit on splitting…"), or the warning should be repeated against the Retire button. A test asserting the warning is shown and the retire path still runs would pin whichever wording is chosen.

**[SUGGESTION]** — `sourcing-rule-delete-dialog.test.tsx:~130`
> `it('surfaces a delete refusal verbatim and stays open')` rejects with a **404** and then asserts `'This sourcing rule no longer exists.'` — which is `describeSourcingRuleError`'s canned 404 sentence, not the server's message. The name (and the PR body's "A refusal renders **verbatim**") describe the 400/409 branches, and neither is exercised here. Either rename the test to what it checks (a 404 renders the canned sentence and the dialog stays open) or reject with a 400 carrying a distinctive message and assert that message appears — the verbatim path is the one the #3056 helper's docblock argues hardest for, and it is currently untested on this component.

**[SUGGESTION]** — `sourcing-rule-delete-dialog.test.tsx:~50` (`renderDialog`)
> `rule={target}` is applied after `{...props}`, where `target` is already `props.rule ?? rule()`. The override is a no-op for every call site. Dropping it (and letting `{...props}` win) removes a line a future reader has to reason about.

**[SUGGESTION]** — carried from #3229, repeated only because this PR ships on top of it
> `var(--bg-subtle)` in `.sourcing-rule-row--governing` still resolves to nothing (`--bg-subtle` is declared nowhere in `index.css` at this head SHA, and is not in `tokens.ts`; `check-design-tokens.mjs` is catalog→CSS only, so it passes). And `features/oms` is still absent from both `no-restricted-imports` pattern groups in `.eslintrc.js`, which this PR widens by two more barrel exports. Neither originates here; whichever PR in the stack is most convenient should close them.

**Positive observations.** The reason copy for each blocked-retire case is genuinely different and each points at the right remedy, rather than one generic "not available" — that is the #2231 rule applied without being asked. Hiding the affordance *with* a stated reason instead of rendering it disabled is the right call for a button whose absence would otherwise be unexplained. And the comment justifying `new Date().toISOString()` over a back-dated instant — naming the few-seconds cost rather than papering over it — is the standard this programme holds itself to.

### Verdict
🔄 **Approve with changes**

Priority: (1) add the `scheduled` arm to `retireUnavailableReason`, (2) reword the splitting warning so it covers both removal routes.

Note on CI: this PR targets `3058-sourcing-rule-dialog`, and the workflow triggers on `pull_request: branches: [main]` only — **no checks ran**. The `MERGEABLE` / `CLEAN` state is a merge-conflict answer, not a test result; the PR body's local runs are the only evidence.

---

norbert-kulus-blockydevs:
Review addressed in `935203f0b` (plus a merge of the updated base, `9215134b9`, which carries #3229's fixes).

**[IMPORTANT] `retireUnavailableReason` missed `scheduled` - fixed.**
Confirmed against the backend exactly as described: `SourcingRuleAdminService.update` passes `patch.effectiveTo ?? existing.effectiveTo` and `patch.effectiveFrom ?? existing.effectiveFrom` into `assertEffectiveWindow` (`sourcing-rule-admin.service.ts:129-132`), which throws `BadRequestException` on `to <= from` (`:224-231`). So `PATCH { effectiveTo: <now> }` against a future start date is always refused.

The function now reads `resolveSourcingRuleStatus` once and carries one arm per state it can answer, with the reason for each recorded in the docblock. The `scheduled` copy says the rule has not started, so there is nothing to retire, and points at both remedies (delete it, or change its start date in Edit) rather than mentioning a 400 the operator did not cause. Note the `!rule.recognised` check became `status === 'unrecognised'` - identical, since `resolveSourcingRuleStatus` returns that first, and it makes the four states read as one switch.

New test `does not offer retire for a rule that has not started yet`; it fails against the old guard (verified by reverting the arm).

**[IMPORTANT] the `removalLoosens` Alert attributed the consequence to Delete alone - fixed.**
Copy is now "Removing this rule lifts your only limit on splitting: ... Retiring has the same effect as deleting here, because a retired rule is not evaluated either."

One deliberate deviation from the suggestion's options: I did **not** gate that second sentence on whether Retire is on offer, and did not repeat the warning against the Retire button. `resolveSplitCeiling` counts only ACTIVE rules, so a removal that loosens is by definition the removal of an active rule - and `active` is the one state `retireUnavailableReason` never blocks. A gate would therefore have been a branch that cannot be taken, so the sentence is unconditional and the docblock records why.

Two tests: the existing warning test now also asserts the both-routes sentence (fails against the old copy), and a new `still runs the retire path while the splitting warning is shown` pins that the warning does not suppress the reversible action.

**[SUGGESTION] the 404 test's name did not match what it checks - fixed, and the verbatim path is now covered.**
Renamed to `renders the canned sentence for a 404 delete and stays open`, with a comment saying the sentence is `describeSourcingRuleError`'s canned 404 and not the server's message. Added `surfaces a 400 retire refusal VERBATIM and stays open`, which rejects with the real `assertEffectiveWindow` message and asserts that text appears. It is framed as the race the new guard cannot close: the client guard is not a trust boundary, so a rule whose start date moved in another tab still reaches the API, and its 400 names the actual problem.

**[SUGGESTION] `rule={target}` after `{...props}` - fixed, but not by dropping the line.**
The premise is half right: where `props.rule` is set the override is indeed a no-op. Where it is not, though, that trailing line is the ONLY thing supplying `rule`, since `rules={[target]}` sits before the spread and `{...props}` contributes no `rule` key - deleting it outright makes the component `return null` and takes out most of the suite. Fixed by moving it above the spread instead, so the helper now reads defaults-then-overrides like every other one, and the line a reader had to reason about is gone.

**[SUGGESTION, carried] `var(--bg-subtle)` and the `features/oms` eslint registration - both already fixed up-stack, and this branch now carries them.**
`--bg-subtle` went in `3257a96cc` on `3057-sourcing-rules-table` (both usages repointed at the catalogued `--bg-surface-muted`); `grep -rn 'var(--bg-subtle)' apps/web/src` returns nothing here. The eslint slug went in `2d93a01f7` on `3056-sourcing-rules-api-client`; `.eslintrc.js` now carries ten `'**/oms/<part>/**'` entries, i.e. all five subdirectories in both pattern groups.

Local gate, run in this worktree: `npx vitest run src/features/oms` (12 files, 106 tests, green), `pnpm --filter @openlinker/web type-check` (clean), `npx eslint` on both changed files (clean).

piotrswierzy (commented):
## Tech Lead re-review (`f8b74529` → `15b30e07`)

Both IMPORTANT findings are closed, correctly and with tests that can fail. `retireUnavailableReason` gains a `scheduled` arm citing the actual backend reason (`SourcingRuleAdminService.update` validates the *merged* window, so `PATCH { effectiveTo: now }` against a future `effectiveFrom` is always `to < from`), so the dialog no longer offers the one button its own docblock says it exists to prevent — and the arm is guarded by a test that asserts both the absent button and the visible reason. The splitting-limit warning now names both routes ("Retiring has the same effect as deleting here, because a retired rule is not evaluated either"), with the sharp observation recorded in the header that the sentence must stay *unconditional*: `resolveSplitCeiling` counts only active rules, so a loosening removal is by definition of an active rule, which is the one state `retireUnavailableReason` never blocks — a gate there would be an untakeable branch. That is backed by a test asserting Retire still renders while the warning is shown. The error-path coverage gap is closed too: the 404 test is renamed to say it asserts the canned sentence, and a separate test now drives a 400 retire refusal and asserts the server's own message survives verbatim.

One new finding, and it is the exact class the brief flags: two CSS classes this dialog's footer names are declared nowhere.

### Findings

**[IMPORTANT]** — `apps/web/src/features/oms/components/sourcing-rule-delete-dialog.tsx:210,218`
> `dialog__footer--split` and `dialog__footer-group` do not exist. Neither is declared in `apps/web/src/index.css`, and `grep` across `apps/web/src` finds no other consumer — this dialog is the only site, so they were invented here and never written. `check-design-tokens.mjs` is catalog→CSS only and `check-css-structure.mjs` only parses well-formedness, so nothing catches an undeclared *class*; this is the sixth invisible-UI instance in this programme.
>
> The effect is not cosmetic-neutral. `.dialog__footer` (index.css:5657) is `display: flex; justify-content: flex-end; gap: var(--space-2)`. With no `--split` override, the intended "Retire on the left, Cancel/Delete on the right" layout never happens — the `<span />` spacer is a zero-width flex item with no `flex: 1`, so all three buttons collapse right-aligned and `Retire instead` sits immediately beside `Cancel`, which is precisely the adjacency the split was there to prevent on a footer whose right-hand neighbour is a destructive action. And because `dialog__footer-group` has no rule either, Cancel and Delete stop being flex items of the footer and become inline children of an unstyled span, so the `gap: var(--space-2)` between them is replaced by a default word-space.
>
> Either declare both (`.dialog__footer--split { justify-content: space-between; }` plus a `display: flex; gap: var(--space-2)` group), or drop them and express the split with markup the existing `.dialog__footer` already supports. The `<span />` spacer should go either way — an empty element standing in for layout is what made the missing rule undetectable by eye in the retire-blocked branch, where the spacer is the only thing rendered on the left.

### Verdict
🔄 **Approve with changes** — the two findings this pass was for are genuinely closed; land the footer CSS before merge.

---

norbert-kulus-blockydevs:
Addressed in `3bd9e58fd`.

## [IMPORTANT] `dialog__footer--split` / `dialog__footer-group` "do not exist" - declined, the premise is false

Both classes **are** declared, and have been since this PR's first commit. `f8b74529d` ("feat(oms): delete/retire flow for sourcing rules (#3059)") added the consumers *and* the declarations in the same commit, which is the left end of the very range this pass reviewed (`f8b74529` -> `15b30e07`):

```
$ git log --oneline -S "dialog__footer--split" -- apps/web/src/index.css
f8b74529d feat(oms): delete/retire flow for sourcing rules (#3059)
```

At the head that was reviewed:

```
$ git show 15b30e079:apps/web/src/index.css | grep -n "^\.dialog__footer--split\|^\.dialog__footer-group"
23742:.dialog__footer--split {
23746:.dialog__footer-group {
```

and the same two blocks are in the PR's own diff against its base (`apps/web/src/index.css`, +12 lines, the only CSS this PR touches):

```css
/* The delete/retire confirm offers a third, differently-toned action, so its
   footer splits: the reversible alternative on the left, away from the
   destructive pair on the right (#3059). */
.dialog__footer--split {
  justify-content: space-between;
}

.dialog__footer-group {
  display: inline-flex;
  gap: var(--space-2);
}
```

So the consequences the finding derives do not occur: with `justify-content: space-between` on the footer the split does happen, `Retire instead` is not adjacent to `Cancel`, and `Cancel`/`Delete` keep `gap: var(--space-2)` as flex items of the group rather than falling back to a word-space. This is not the sixth invisible-UI instance; the class was written in the same commit as its first use.

I did not make a cosmetic change to close a finding that is already satisfied, but two things inside it hold on their own merits and are fixed.

## [ACCEPTED] the `<span />` spacer is gone

"The `<span />` spacer should go either way" is right, and it is fixable without losing the layout. `space-between` only means something with two children, so instead of rendering an empty element to hold the left slot open, the **modifier** now comes off when Retire is blocked - `.dialog__footer`'s own `justify-content: flex-end` (index.css:5657) already right-aligns the single remaining group:

`sourcing-rule-delete-dialog.tsx:214`
```tsx
<DialogFooter className={retireBlockedReason === null ? 'dialog__footer--split' : ''}>
  {retireBlockedReason === null ? (
    <Button tone="secondary" ...>Retire instead</Button>
  ) : null}
  <span className="dialog__footer-group">
```

`DialogFooter` joins with `.filter(Boolean)` (`shared/ui/dialog.tsx:53`), so the empty string drops cleanly.

Two tests pin it: `splits the footer while Retire is on offer` (modifier present, two children) and `drops the split modifier - not a spacer element - when Retire is absent` (neither). The second fails against the previous markup - verified by reinstating the unconditional modifier plus the spacer and re-running: `Tests 1 failed | 12 passed`.

## [FIXED, found while verifying] the group silently opted out of the footer's mobile stacking

This one is real, and it is the only place the footer actually misbehaved. `index.css:11060` stacks the footer at `<=480px` and stretches its actions:

```css
@media (max-width: 480px) {
  .dialog__footer { flex-direction: column-reverse; align-items: stretch; }
  .dialog__footer .button { width: 100%; }
}
```

`.dialog__footer-group` is a row, so `Cancel` and `Delete` stayed side by side inside it and their `width: 100%` shrank them to share one line, while `Retire instead` went full width below - the one footer in the app that ignored the stacked contract, and exactly the kind of thing that is invisible unless a phone is opened. Fixed with one declaration, placed next to the block this PR owns rather than inside the shared `@media (max-width: 480px)` block, since `index.css` is shared across this stack:

`index.css:23713`
```css
@media (max-width: 480px) {
  .dialog__footer-group {
    flex-direction: column-reverse;
  }
}
```

Reversed for the same reason the footer itself is, so the weightiest action reads first. `480px` rather than a new bound, because the contract it has to honour is declared at `480px`.

## Gate

- `npx vitest run src/features/oms` - 12 files, 112 tests, all pass (13 in this dialog's suite).
- `pnpm --filter @openlinker/web type-check` - clean.
- `npx eslint` on both changed TS files - clean.
- `pnpm check:invariants` - exit 0, `check-css-structure` included (3618 blocks balanced, so the new media block parses), `check-ui-vocabulary` included.

Note on the CI observation from the previous pass: still accurate, this PR targets `3058-sourcing-rule-dialog` and the workflow triggers on `main` only, so the runs above are the evidence.

piotrswierzy (approved):
## Tech Lead re-review (`15b30e07` → `3bd9e58f`)

**The single open finding is closed, and the fix goes past what I asked for.** Both classes are now declared in `apps/web/src/index.css`, and the zero-width spacer is gone rather than styled around. The PR's own commit here is `3bd9e58f`; the rest of the range is base-branch catch-up through `3058-sourcing-rule-dialog`.

Counts at head, `apps/web/src/index.css`:

| Class | Declarations | Uses |
|---|---|---|
| `.dialog__footer--split` | **1** (`:23699`) | 1 — `sourcing-rule-delete-dialog.tsx:214` |
| `.dialog__footer-group` | **2** (`:23703` base, `:23714` inside `@media (max-width: 480px)`) | 1 — `sourcing-rule-delete-dialog.tsx:220` |

Previously both were **0**. The split modifier is `justify-content: space-between`, which actually overrides the base `.dialog__footer`'s `flex-end` at `:5657`, so the intended left/right split now renders; `.dialog__footer-group` is `display: inline-flex; gap: var(--space-2)`, which restores the gap Cancel/Delete were losing as inline children of an unstyled `span`.

The markup change is the better half of the fix, and the rendered result matches what exists:

- **The `<span />` spacer is deleted** (`) : ( <span /> )` → `) : null`), and the modifier is applied **conditionally** — `className={retireBlockedReason === null ? 'dialog__footer--split' : ''}`. `space-between` over one child is a no-op, so with Retire absent the footer correctly falls back to the base `flex-end` with nothing rendered to hold a slot open. The in-place comment says exactly that.
- **The mobile interaction is handled, which I had not named.** `.dialog__footer` already stacks `column-reverse` + `align-items: stretch` with `width: 100%` buttons inside `@media (max-width: 480px)` (`:11051`). Left alone, the new `inline-flex` group would have stayed a row and shrunk its two buttons to share one line — the one footer in the app quietly opting out of that contract. The new `@media (max-width: 480px)` block at `:23713` reverses the group to `column-reverse` for the same reason the footer itself is. Breakpoints match exactly (`480px` both sides), so there is no gap or double-apply band.

Both branches are now pinned by tests at `sourcing-rule-delete-dialog.test.tsx:147` and `:155`, and the second asserts `footer?.children` has length **1** — which is what structurally prevents a spacer element from being reintroduced later, not just the class name. Per `docs/frontend-architecture.md` § Design tokens, `check-design-tokens.mjs` is one-directional (catalog → CSS), so lint was green over the undeclared classes and remains no evidence either way; the declarations above were read out of `index.css` directly.

The two earlier findings still hold in code:

- **`retireUnavailableReason`'s `scheduled` arm** (`sourcing-rule-delete-dialog.tsx:103`) is present and cites the cause — `assertEffectiveWindow` validates the **merged** window, so a `now` end date against a future start is always `to < from` and always 400 — with the docblock noting it is the arm an operator is most likely to reach. Guarded by `does not offer retire for a rule that has not started yet` at `:126`, asserting both the absent button and the visible reason.
- **The splitting-limit warning names both routes** and the file header explains why the sentence must stay unconditional: `resolveSplitCeiling` counts only ACTIVE rules, so a loosening removal is by definition the removal of an active rule, which is the one state `retireUnavailableReason` never blocks — a gate there would be an untakeable branch. Backed by `still runs the retire path while the splitting warning is shown` at `:137`.
- **`surfaces a 400 retire refusal VERBATIM and stays open`** at `:180` is intact.

### Findings

### Verdict
✅ **Approve**

CI: `no checks reported on the '3059-sourcing-rule-delete' branch` — absent, not passing. `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN` against `3058-sourcing-rule-dialog`.

---

Issue #2364, opened 2026-08-22
[FEATURE] Frontend — automations index and trigger list


piotrswierzy:
> Part of the OMS programme backlog (**Wave 2**, stream S3, size M). Backlog slug: `W2-27`. Sibling `W2-*` slugs resolve via the Wave-2 epic's checklist. Design of record: Wave-2 product spec §5.1 (PR #2318).

## Problem / Context

The nav already reserves a `Planned → Automations` slot; spec §5.1 promotes it.

## Proposed Solution

A real `/automations` route in the `Operations` nav group listing the eight v1 triggers as scannable rows (rule count, last fired, status badge, `Configure`), mirroring `apps/web/src/features/sales-documents/components/sales-document-country-index.tsx` in shape and class reuse. Per-rule rows show active/inactive with a one-click deactivate that **preserves the fired log**. A `Run log` action in the page header opens `/automations/activity` (`W2-48`).

**First run — the zero-rules state (spec §5.1, exact copy).** On a fresh install eight triggers with `0 rules` each is technically complete and useless as a starting point, so the page leads with a single card **above** the trigger index:

> **You have no automations yet.**
> Most sellers start with this one:
>
> **When an order is marked packed → buy the shipping label → tell the marketplace.**
> One click at the packing bench instead of three, and the marketplace hears about it straight away.
>
> `[Set this up]`   `[Start from scratch]`

`Set this up` opens the composer **pre-filled** with T5 → A2 → A3, no conditions, inactive, and still subject to the S3-2 arming gate — the operator reviews and arms it exactly as if they had built it. It is a suggestion, **never a rule that exists before the operator saved it**: nothing is created by opening the page. Exactly **one** suggestion is offered, deliberately. The card disappears permanently once any rule exists (active or not), and never returns after the operator deletes their last rule.

**A rule that can never match says so (S3-5).** A rule whose threshold currency matches no order in the selected marketplace's currency — or whose conditions are otherwise impossible — renders a warning on its own row stating that it has never matched an order, **with the reason**, following the #2170 composer's disabled-with-a-caveat precedent for `buyerHasTaxId`. It lives here rather than in the composer because the condition is only observable *after* the rule exists.

## Classification

**Type**: Frontend
**Layer**: Interfaces
[...]

PR #3233, opened 2026-09-14
feat(oms): sourcing-rules read states and the unrecognised-rule refusal (#3061)

### LINKED ISSUE #3061: feat(oms): empty/loading/error + unrecognised-rule locked states

norbert-kulus-blockydevs:
## Summary

Adds `SourcingRulesSection`, which owns the read's four states and all three dialogs — so the page (#3060) carries only its own two gates: which connection this is, and whether any location exists.

**Stack position:** base is `3059-sourcing-rule-delete` (#3230). #3060 stacks on this one.

**Loading, a FAILED read and a confirmed-empty ruleset are three distinguishable claims**, asserted separately. Rendering *"nothing decides where an order ships from"* while a read is in flight, or after one failed, would state something about the operator's own configuration on the strength of a network problem. The error copy therefore says **explicitly that nothing changed** — a failed READ is the failure most easily mistaken for a lost configuration.

**An unrecognised rule's Edit control is now ENABLED and routes to an explanation**, rather than being disabled. A disabled button cannot be focused in every browser and its `title` is not reliably announced, so the explanation *and its remedy* would be unreachable for exactly the operator who needs them.

> ⚠️ **This supersedes #3228's disabled control.** That PR's test asserted `toBeDisabled()` + `onEdit` not called; this PR replaces it with two tests — enabled-and-routes-to-`onEditRefused`, and a **disabled fallback** for a caller that has not wired the refusal, so the form still cannot be opened on a rule the server will refuse. Reviewing #3228 on its own will show the older behaviour; the stack tip is what ships.

**The refusal states only what is TRUE of the row** — it is saved and being ignored — and quotes the stored `kind` and `name` verbatim. OpenLinker knows the rule is unevaluable; it does **not** know whether the name was removed, renamed, or written by a newer build, so a guess at the cause would be a claim it cannot support. The dialog hands the operator straight to the delete/retire confirm (#3059), which is the one remedy there is.

**The reorder is NOT optimistic.** An arrow sends the whole live id list and waits; a row that moves in the browser and then moves back is a worse answer than one that moves once, when the server agrees. A 409 is surfaced with the server's own sentence rather than leaving a row to snap back unexplained, and every control is disabled while the write is in flight so a second click cannot build a list from a set the server is already changing.

## Test plan

- [x] `vitest run src/features/oms` - 13 files, 108 tests, all pass
- [ ] Manual: the section has no page to mount it until #3060

Refs #3055
Closes #3061


piotrswierzy (commented):
## Tech Lead review

Composes the read's four states and all three dialogs into one `SourcingRulesSection`, and — the substantive change — supersedes #3228's *disabled* Edit on an unrecognised rule with an **enabled control that routes to an explanation**. That is the right call against the #2953 contract (`docs/architecture-overview.md` § 26 Fulfillment: *"an unrecognised row is LISTED, flagged `recognised: false`, and cannot be patched … exactly the one an operator needs to find in order to delete it"*): the refusal is surfaced, it states only what is true of the row, it quotes the stored `kind`/`name` verbatim, and it hands straight to the #3059 delete confirm — which the section test drives end to end. The disabled fallback when no `onEditRefused` is wired keeps the contract safe for an unmigrated caller. Loading / failed-read / confirmed-empty are three genuinely separate claims, each asserted, with the error copy explicitly saying nothing changed. One thing spoils the headline change: the now-clickable button still announces itself as *"Cannot edit"*.

### Findings

**[IMPORTANT]** — `apps/web/src/features/oms/components/sourcing-rules-table.tsx:296,303`
> The refused control is now **enabled** but keeps `COPY.editLocked` as both `title` and `aria-label`, so its accessible name is `"Cannot edit — this rule is no longer recognised"`. That is a button that says it cannot act and then acts. The whole argument for enabling it (the docblock at `:277-287`, and the PR body) is that the explanation *and its remedy* must be reachable — but nothing in the name tells an operator that clicking gets them the explanation, so a sighted operator reads "cannot edit" and moves on, and a screen-reader user hears a refusal announced as an available action. The label should describe the affordance in the enabled branch and keep `editLocked` only for the `onEditRefused === undefined` fallback, e.g.
> ```ts
> const title = refused
>   ? (onEditRefused === undefined ? COPY.editLocked : COPY.editRefusedAction) // "Why this rule cannot be edited"
>   : `Edit ${label}`;
> ```
> Add the new key to `SOURCING_RULES_TABLE_COPY` beside `editLocked` so both sentences stay in `sourcing-rule.copy.ts`. Note this also moves `sourcing-rules-section.test.tsx:99` and `sourcing-rules-table.test.tsx`'s two queries onto the new name — currently both select by the refusal sentence, which is itself a small signal that the name is doing the wrong job.

**[SUGGESTION]** — `apps/web/src/features/oms/components/sourcing-rules-table.tsx:292`
> `renderEdit` still short-circuits on `onEdit === undefined` alone, so a caller that wires **only** `onEditRefused` renders no control at all for an unrecognised rule — the "absent one reads as a rendering bug" case the docblock above it argues against. `if (onEdit === undefined && onEditRefused === undefined) return null;` costs nothing and makes the guard match the two-handler shape the prop now has.

**[SUGGESTION]** — `apps/web/src/features/oms/components/sourcing-rules-section.tsx:96-106`
> The `ErrorState` Retry gives no feedback while `refetch()` is in flight: TanStack keeps `error` non-null until the retry succeeds, so the whole card sits unchanged and a slow retry reads as a dead button. `disabled={rulesQuery.isFetching}` (or swapping the label) is the one-liner; the section already threads `busy` for exactly this reason on the reorder path.

**[SUGGESTION]** — `apps/web/src/features/oms/components/sourcing-rules-section.tsx:118-125`
> The reorder `Alert` clears only when the next `mutate()` runs, so a conflict banner outlives the refreshed list it is describing and there is no way to dismiss it. Since the mutation already invalidates on a 409 and the table re-renders from the server's own set, resetting on the next successful list settle (or offering a dismiss) would stop a stale sentence sitting above a list that has since been corrected. Not a correctness problem — `Alert tone="error"` already carries `role="alert"`, so it is announced.

**Positive observations.** The locked dialog's *"It is saved, but it is not being used — orders route as if it were not here"* is exactly the claim OpenLinker can support and no more; the `{rule.kind} · {rule.name}` line preserves the raw vocabulary for a support ticket rather than passing it through a label map that would hide it. The reorder is non-optimistic with `busy` disabling every control, which is the right reading of an exhaustive `PUT /order` that can be refused. Every custom property and class the diff touches (`.dialog__body`, `.muted-text`, `.mono-text`, `.toolbar__group`, `.rule-actions`, `.button--ghost/--icon/--sm`) is declared in `apps/web/src/index.css` at the stack tip — checked directly, not inferred from `check-design-tokens.mjs`, which is one-directional.

**CI.** No checks reported on `3061-sourcing-rules-states` — the workflow triggers on `pull_request: branches: [main]`, so a stack-targeted PR gets none. Absent, not passing. `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN` against its base.

### Verdict
🔄 **Approve with changes**

---

norbert-kulus-blockydevs:
Review findings addressed in 2e8b5c9fe. Three fixed, one declined with evidence and replaced by a regression test.

**[IMPORTANT] `sourcing-rules-table.tsx:296,303` - the enabled control announced a refusal. Fixed.**

New `SOURCING_RULES_TABLE_COPY.editRefusedAction` ("Why this rule cannot be edited") is the accessible name whenever the control is live; `editLocked` survives only on the `onEditRefused === undefined` fallback, where the button really is inert. Both sentences stay in `sourcing-rule.copy.ts` as suggested. `sourcing-rules-section.test.tsx:87` and both `sourcing-rules-table.test.tsx` queries moved onto the new name, and the enabled-case test now also asserts the refusal sentence is *absent*, so the two names cannot quietly converge again.

**[SUGGESTION] `renderEdit` short-circuited on `onEdit` alone. Fixed.**

Guard is now `if (onEdit === undefined && onEditRefused === undefined) return null;`. One thing the one-liner needed on top: with only `onEditRefused` wired, a *recognised* row would have reached `onEdit(rule)` on a handler that does not exist, so there is a second early return for that case (a recognised rule with no edit handler renders nothing). Two new tests cover both halves.

**[SUGGESTION] Reorder `Alert` outliving its list. Fixed, by dismiss rather than by auto-clear.**

The `Alert` `action` slot now carries a Dismiss that calls `reorderMutation.reset()`. I deliberately did not reset on the next successful list settle: the mutation already invalidates on a 409, so the refreshed rows arrive within a few hundred milliseconds and an auto-clear would pull the explanation away roughly as fast as it appeared - the operator would see a row snap back with nothing left saying why, which is the state the banner exists to prevent. The reasoning is recorded on the copy key and in the section docblock. A subsequent `mutate()` still replaces it, as before.

**[SUGGESTION] Retry disabled while the refetch is in flight. Declined - the premise does not hold on TanStack Query v5, and I pinned the real behaviour instead.**

The finding assumes `error` stays non-null until the retry succeeds. In v5 `refetch()` on an errored query resets `status` to `pending` and clears `error` for the duration. I measured it rather than reasoning about it, under both the test client (`test-utils.tsx:1205-1213`, `gcTime: 0`) and the production one (`app-providers.tsx:42-50`, default `gcTime`, `staleTime: 30_000`) - identical in both:

```
AFTER ERROR:  status=error   isLoading=false isFetching=false err=yes
DURING RETRY: status=pending isLoading=true  isFetching=true  err=no
```

`sourcing-rules-section.tsx` reads `isLoading` *before* `error`, so pressing Retry swaps the whole error card for the loading card - the button is gone rather than sitting inert, and `LoadingState` carries `role="status" aria-live="polite"`, so the change is announced. `disabled={rulesQuery.isFetching}` would have been unreachable code that a later reader would take as load-bearing.

I also logged every render in that window to check for an intermediate frame where `isLoading` is false and `error` is null - which would fall through to the empty state and claim "nothing decides where an order ships from" mid-retry. There is exactly one render and no such frame.

What I did instead: the property is real but rests entirely on the ORDER of those two branches, so swapping them would reintroduce exactly the dead-button reading the finding describes. `sourcing-rules-section.test.tsx` now asserts that pressing Retry shows the loading card, removes the error card and the Retry button, and still makes no empty claim - and the section docblock says the ordering is what buys this.

**Checks.** `apps/web` `tsc -b` clean; `eslint` clean on `features/oms`; `vitest run src/features/oms` 111/111. CI reports nothing on this branch because the workflow triggers on `pull_request: branches: [main]` - absent, not passing, as the review notes.

piotrswierzy (approved):
## Tech Lead re-review (`61d6783` → `2750231`)

The one IMPORTANT is closed, cleanly and for the stated reason. `renderEdit` now computes `inert = refused && onEditRefused === undefined` and picks `COPY.editRefusedAction` ("Why this rule cannot be edited") for the enabled branch, keeping `COPY.editLocked` as the name only where the control really is disabled (`sourcing-rules-table.tsx:394-402`). Both `title` and `aria-label` read off the same value, so the accessible name and the visible tooltip cannot diverge — and the copy constant carries its own docblock explaining why the refusal sentence must not name a live control. Both stack-wide items are also resolved on this head: `--bg-subtle` is gone (0 declarations, 0 uses in `index.css`) in favour of the catalogued `--bg-surface-muted`, which carries both a light (`:205`) and a dark (`:410`) definition and is present in `tokens.ts:87`; and `features/oms` is registered in **both** `no-restricted-imports` groups for all five canonical subdirectories (`.eslintrc.js:329-333`, `:608-612`), alongside a matching `SCAN_ROOTS` entry in `scripts/check-ui-vocabulary.mjs` — which is the pairing `docs/frontend-architecture.md` § Feature Public Surface (#2364) requires, not two independent edits.

#3228's two carried-over ACs both still hold, verified in the section rather than inferred: `sourcing-rules-section.test.tsx:101` asserts `reorder` is called with the full live id list (`['b','a']`), and `:120` asserts a 409 surfaces as a hand-dismissed banner rather than letting a row snap back unexplained — matched by `sourcing-rules-section.tsx:116-132`, where the reorder's own conflict is read separately from the read's and cleared only by `reorderMutation.reset()`.

### Findings

**SUGGESTION** — `apps/web/src/features/oms/components/sourcing-rules-table.tsx:387`
> `renderEdit` returns `null` when both handlers are absent and again when a recognised rule has no `onEdit`. Correct, but the two early returns read as one rule split in half. A single `if (refused ? onEditRefused === undefined && onEdit === undefined : onEdit === undefined) return null;` after `refused` is computed would say it once. Purely cosmetic — the current form is already covered by the section test.

### Verdict
✅ **Approve**

Mergeable: `MERGEABLE` / `CLEAN` into `3059-sourcing-rule-delete`. **No CI checks are reported on this branch** — absent, not passing.

---

Issue #3062, opened 2026-09-10
test(oms): sourcing-rules component + hook tests


norbert-kulus-blockydevs:
## Problem / Context

No test coverage exists for any of the above.

## Proposed Solution

Add `*.test.tsx`/`*.test.ts` colocated with each new file, covering the reorder interaction, kind-filtered name options, validation, and the unrecognised-rule lock, per `docs/testing-guide.md`.

## Classification

**Type**: Frontend
**Layer**: Interface (tests)
**File(s)**: apps/web/src/features/oms/**/*.test.{ts,tsx}

## Dependencies

- Part of #3055

## Assumptions

- None

## Acceptance Criteria

- [ ] Reorder, create, edit, delete/retire are covered
- [ ] Unrecognised-rule lock is covered
- [ ] pnpm test passes
- [ ] Tests added or updated for non-trivial logic
- [ ] No architecture boundary violations (CORE ↔ Integration)

PR #3234, opened 2026-09-14
feat(oms): sourcing rules page reached from a /settings tile (#3060)

### LINKED ISSUE #3060: feat(oms): sourcing rules page reached from a /settings tile

norbert-kulus-blockydevs:
## Summary

The screen an operator actually reaches: `/settings/sourcing-rules`, behind a tile on the settings page, with a back-link and a two-level crumb and **no sidebar entry** — the tile is the only way in, settled on #3055 § Decisions in the mockup's favour.

**Stack position:** base is `3061-sourcing-rules-states` (#3233). #3062 stacks on this one.

## The page owns exactly two gates, and nothing else

The read's own states, the table and every dialog stay in #3061's section, so *"what does this read look like while it is loading, failing or empty"* has one answer rather than two.

**Gate 1 — the connection.** A ruleset is per connection and only the OpenLinker OMS connection can hold one (the backend refuses authoring on anything else, #2953), so rendering the editor otherwise would let an operator author rules nothing will ever read. An explicit `?connectionId=` is honoured so a link from a connection page lands somewhere truthful, and **validated rather than trusted**. The two causes are told apart, because *"you are looking at the wrong connection"* and *"the OMS is not enabled"* have different remedies.

**Gate 2 — the locations.** Rules choose BETWEEN locations, so with none there is nothing to rank. The gate asks about **active** locations, since a ruleset that can only rank retired warehouses is the same degenerate state.

**Neither gate is reported while the reads are in flight or after one FAILED.** Either would state something about the operator's configuration on the strength of a network problem — asserted explicitly, both ways.

## `listLocations` on the inventory API client

The priority picker needs a real list, and this branch carried only `listActiveLocations` — a degenerate `limit=1` probe where only `total` is meaningful. The new method is added **deliberately under the same name and shape #3198 uses** (`listLocations(filters?, pagination?)`, 1-based `page`/`limit`), so the two converge on merge rather than compete; the filter type is a documented subset that #3198 widens.

It lists **every status**, not just active ones: an inactive location is a legitimate priority entry, and an operator retiring one should not silently lose its place in a ruleset — the picker marks it instead. The page's enablement gate is the different question and asks about active ones.

## Also here

- The tile is **admin-gated**, unlike `WhoDecidesTile`: the sourcing-rules API is `@Roles('admin')` end to end (#2953), reads included, so a non-admin reaching the page would meet a 403 rather than a read-only view.
- `EXPECTED_LAZY_ROUTE_COUNT` goes 63 → 64, with the route named in that file's inventory comment.
- `createMockApiClient`'s inventory default answers an install with **no locations** — the honest zero-config state, so a test that needs one opts in rather than inheriting a warehouse it never declared.

## Test plan

- [x] `eslint` on every changed path - clean
- [x] `vitest run src/features/oms src/pages/oms` - 14 files, 115 tests, all pass
- [x] `vitest run src/pages/settings src/app` - 349 tests, all pass (route count + settings tiles)
- [ ] Manual: `/settings` → Sourcing rules, against an OMS connection with a location

Refs #3055
Closes #3060


piotrswierzy (commented):
## Tech Lead review

Mounts the screen at `/settings/sourcing-rules` behind an admin-gated tile, with the page owning exactly two gates and delegating every read state to #3061's section. The layering is right and the route contract is honoured properly (lazy, `handle.crumb`, `EXPECTED_LAZY_ROUTE_COUNT` 63→64 **with** the inventory comment updated rather than just the number). I verified the tile's admin gate against the backend rather than taking the claim: `apps/api/src/oms/http/oms-sourcing-rules.controller.ts:72` carries a **class-level** `@Roles('admin')` covering the `@Get()` too, so gating the whole tile — the documented deviation from `WhoDecidesTile`, whose read is authorised for a viewer — is correct, and the reasoning is recorded in place. Four things to fix: the no-locations gate's remedy is a dead end, the two gate causes are told apart by the wrong signal, the gate itself reads a truncatable page, and the settings page's own admin-gating test is not extended.

### Findings

**[IMPORTANT]** — `apps/web/src/pages/oms/sourcing-rules-page.tsx:120-131`
> The no-locations blocking state's action is labelled `COPY.manageLocations` (*"Manage locations"*) and links to `/settings`. There is no locations screen and no `InventoryLocationsTile` on `main` — I grepped both `apps/web/src/app/routes/` and `settings-page.tsx`; `#3063`/PR #3198 is unmerged. So an operator blocked here clicks a button promising a locations screen, lands back on the page they came from, and finds nothing about locations at all. Issue #3060's AC names *"a link to the locations screen (#3063) / the bootstrap action"* — the second half is available **today**: `apiClient.inventory.bootstrapLocations()` already exists (`inventory.api.ts:68`), and `docs/architecture-overview.md` § Inventory (#2407) designed `POST /inventory/locations/bootstrap` as precisely this: *"an **offer an operator takes**, never a seed"*, idempotent, insert-then-recover on the unique code. Either wire that as the action, or relabel the link and say in the message that the locations screen is not available yet. A link that names a destination the app does not have is the same class of false statement both docblocks in this stack exist to prevent.

**[IMPORTANT]** — `apps/web/src/pages/oms/sourcing-rules-page.tsx:104`
> `const named = requestedConnectionId !== null;` picks the copy from **whether a query param was supplied**, not from the cause. With `?connectionId=anything` and zero OMS connections in the install, the page renders *"This is not your OpenLinker OMS connection"* — a statement about a connection comparison that never happened, when the true fact is `noOmsTitle`. The PR body's claim that *"the two causes are told apart, because they have different remedies"* is therefore only true when an OMS connection exists. The distinction is free: `const named = requestedConnectionId !== null && omsConnections.length > 0;`. `sourcing-rules-page.test.tsx` should gain the third case (a named connection id **and** no OMS connection ⇒ `noOmsTitle`), which the two existing gate tests currently leave uncovered.

**[IMPORTANT]** — `apps/web/src/pages/oms/sourcing-rules-page.tsx:120` + `hooks/use-inventory-locations-for-rules-query.ts:24,34`
> The enablement gate is computed from a page capped at `SOURCING_RULE_LOCATION_PAGE_SIZE = 200`, and `page.total` is discarded. The picker being one page deep is a defensible scope call and the docblock says so — but the **gate** is a different question and this makes it answerable wrongly: an install whose first 200 rows are all `inactive` renders *"No locations to route to yet"* and refuses authoring while active locations exist. The branch already carries the correct read for it — `listActiveLocations()` (`inventory.api.ts:65`) is the `status=active&limit=1` probe where only `total` is meaningful, which is exactly what a gate needs. Use that for the gate and keep `listLocations` for the picker; that also removes the silent-truncation ambiguity from the gate entirely. If the picker's own truncation is to stay, read `total` and say so rather than letting it pass unremarked (the ADR-048 decision-5 rule: a bounded read that reports healthy while being half-complete is the failure mode, not the bound).

**[IMPORTANT]** — `apps/web/src/pages/settings/settings-page.tsx:34,117`
> A new admin-gated tile and toolbar chip land with no assertion at the mount site. `settings-page.test.tsx` is the file that pins exactly this: *"shows the Mailer tile for an admin session"* / *"never renders the Mailer tile for a non-admin session"* (both heading **and** `.toolbar-chip`), and the same pair for PostHog, sitting directly beside the deliberately-opposite *"always renders the Who decides what tile, including for a non-admin session"*. Adding a fourth gated tile without the matching pair leaves the gate unasserted while the file it belongs in reads as exhaustive. #3235's `sourcing-rules-tile.test.tsx` renders the tile in isolation and cannot cover this — the gate is the `{isAdmin ? … : null}` at the mount site, not the component.

**[SUGGESTION]** — `.eslintrc.js`
> `features/oms` is not in either `no-restricted-imports` pattern group, for any canonical subdirectory — I checked the stack tip, not just `main`. Per `docs/frontend-architecture.md` § Feature Public Surface step 2, a feature with a public barrel registers its slug in **both** groups or the rule silently fails open for it. Nothing is violating it today (only `pages/` consume the barrel, and pages are the documented exemption), which is exactly the #2761 `sales-documents` shape — *"a pre-existing omission … newly load-bearing once this consumer appeared"* — and the #2364 `automation` precedent registers the slug up front for the same reason. The omission originated in #3226, but this is the PR that gives the feature its first host-layer consumers; five lines close it.

**[SUGGESTION]** — `apps/web/src/pages/oms/sourcing-rules-page.tsx:99,105,112,122`
> `'Retry'`, `'Not available here'`, `'Go to connections'` and `'No locations'` are inline operator sentences while every other string on this screen lives in `SOURCING_RULES_PAGE_COPY`. The copy object already carries `errorRetry` on its `STATE` sibling; moving these four keeps one place to read the screen's voice.

**Positive observations.** `listLocations` being added under #3198's exact name, signature and 1-based `page`/`limit` — with the divergence from `InventoryPagination` called out in the type's own docblock — is the right way to avoid two clients competing on merge. The explicit param build in `inventory.api.ts:85-91` avoids `URLSearchParams` stringifying `undefined` into a value the DTO then validates, and says so. `createMockApiClient`'s locations default being an install with **no** locations is the honest zero-config shape. Both gates are correctly suppressed while the reads are in flight or after one failed, and that is asserted both ways.

**CI.** No checks reported on `3060-sourcing-rules-mount` — stack-targeted, and the workflow triggers on `pull_request: branches: [main]`. Absent, not passing. `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN` against `3061-sourcing-rules-states`.

### Verdict
🔄 **Approve with changes**

---

norbert-kulus-blockydevs:
Review findings addressed in 9f681a8c3. Five acted on, one already fixed down-stack.

**[IMPORTANT] `sourcing-rules-page.tsx:120-131` - the no-locations remedy was a dead end. Fixed by wiring the bootstrap.**

Confirmed the premise: no locations route under `apps/web/src/app/routes/`, no locations tile in `settings-page.tsx`. So "Manage locations" -> `/settings` really did return the operator to the page they had just come from.

Took the first of the two options you offered, since it is the remedy the architecture already designed for this exact state (#2407: "an offer an operator takes, never a seed"). The action is now `useBootstrapLocationsMutation()` - the shipped hook, already exported from the inventory barrel and already used by `RouterReadinessPanel`, so no new client code. Three details copied from that precedent rather than reinvented:

- gated `write.canWrite && isAdmin`, because `POST /inventory/locations/bootstrap` is `@Roles('admin')` while `inventory:write` is also an operator permission - gating on the permission alone renders a control that answers 403; demo read-only keeps the disabled-with-tooltip treatment;
- the copy says what a minted location does **not** do - it holds no stock (ADR-058 decision 2), so an order still cannot be sent there until stock is assigned, and its country is blank on purpose;
- `bootstrap.error` renders below the state rather than being swallowed.

Two things the precedent did not cover. Where the session cannot create a location the button is absent, so the message states that an administrator is needed rather than describing a control that is not on screen - describing an invisible button is the same dead end one step in. And the shared mutation invalidates only `inventoryQueryKeys.activeLocations()`, which is what reopens the gate; the picker's page is keyed separately, so it is refetched from the call-site `onSuccess` rather than left holding the empty list the operator just fixed.

**[IMPORTANT] `:104` - `named` read the query param, not the cause. Fixed.**

Now `requestedConnectionId !== null && omsConnections.length > 0`, with the reason in place: a named connection can only be the *wrong* one if there was a right one to compare it against. `sourcing-rules-page.test.tsx` gains the third case you named - a named connection id plus zero OMS connections asserts `noOmsTitle` **and** asserts the wrong-connection sentence is absent.

**[IMPORTANT] `:120` + the hook - the gate was answered from a capped page. Fixed.**

The gate reads `useActiveLocationCountQuery()` (the `listActiveLocations()` `status=active&limit=1` probe where only `total` is meaningful); the picker keeps `listLocations` and its full-status page. A test pins the case that was previously wrong in both directions: an install whose picker page holds only a retired depot but whose probe reports 3 active locations now opens the editor instead of refusing it.

On the second half - the picker's own truncation is now read and stated rather than passing unremarked. The hook returns `{ options, total }`, and the page renders a warning above the section naming both numbers when they differ, so an operator who cannot find a warehouse they know exists is told why instead of hunting. Two tests, one each way.

**[IMPORTANT] `settings-page.tsx:34,117` - the mount-site gate was unasserted. Fixed.**

Added the matching pair to `settings-page.test.tsx`, between the Mailer/Who-decides block and the PostHog pair, asserting heading and `.toolbar-chip` in both directions - with the reason for deviating from `WhoDecidesTile` recorded on the negative test, since that is the file where a future contributor will weigh the two. Negative-controlled: with the `{isAdmin ? ... : null}` removed the new test fails; restored, it passes.

**[SUGGESTION] `.eslintrc.js` `features/oms` slug. Already fixed down-stack, not by me.**

It landed on `3056-sourcing-rules-api-client` as `2d93a01f7` ("chore(web): register the oms feature slug in both no-restricted-imports groups"), registering `**/oms/{api,hooks,components,lib,types}/**` in both `no-restricted-imports` groups. It is not visible on this branch only because the stack has not been cascaded yet; it arrives here with the merge down rather than by a second edit.

**[SUGGESTION] `:99,105,112,122` - four inline sentences. Fixed.**

`errorRetry`, `notAvailableEyebrow`, `goToConnections` and `noLocationsEyebrow` moved into `SOURCING_RULES_PAGE_COPY`, alongside the new bootstrap and truncation keys, so every string on the screen is in one place.

**Checks.** `apps/web` `tsc -b` clean; `eslint` clean on `pages/oms`, `pages/settings`, `features/oms`, `features/inventory`, `test-utils.tsx`; `vitest` 503/503 across every test file that touches `inventory` (the blast radius of the two new `createMockApiClient` defaults). No CSS touched, and `check-design-tokens` / `check-css-structure` run clean anyway. I also dry-merged `3061-sourcing-rules-states` into this branch on a detached HEAD to check the cascade: no conflict, and the merged result type-checks with 156/156 green. Nothing was pushed from that - the branch head is unchanged.

piotrswierzy (approved):
## Tech Lead re-review (`57b24f0` → `d6eb707`)

All four IMPORTANTs are closed, each with a test that pins it rather than a comment claiming it.

1. **The dead-end link is gone.** The no-locations state now offers `POST /inventory/locations/bootstrap` through `useBootstrapLocationsMutation` (`sourcing-rules-page.tsx:166-190`) instead of linking to a `/settings` screen that does not exist on `main`. It is paired with `useIsAdmin()` as well as `inventory:write` (`:72-77`), which is right — the route is `@Roles('admin')` while `inventory:write` is also held by `operator`, so the permission alone would render a control that answers 403. `createLocationHint` states both what the button does and what it does **not** do (a minted location holds no stock, ADR-058 decision 2; country left blank deliberately), which is the honest version of the #2407 offer. The mutation invalidates `inventoryQueryKeys.activeLocations()` and the page additionally refetches the picker page in `onSuccess`, so the gate that re-opens and the list behind it cannot disagree. Pinned by `sourcing-rules-page.test.tsx` — *"offers the location bootstrap rather than a link to a screen that does not exist"* and *"hides the bootstrap for a session that cannot create a location, and says why"*.
2. **Copy is chosen by CAUSE, not by the query param.** `const named = requestedConnectionId !== null && omsConnections.length > 0` (`:136`) — a named connection can only be the *wrong* one if there was a right one to compare against, so `?connectionId=x` on an install with no OMS connection now correctly reads `noOmsTitle`. Pinned by *"says the OMS is not enabled even when a connection id was named, if there is no OMS connection"*.
3. **The gate reads the probe.** `useActiveLocationCountQuery` (`status=active&limit=1`, `total`-only) drives the enablement decision at `:151`, while the 200-capped full-status page stays with the picker — and the hook's own docblock records why the two questions must not share a read. Pinned in both directions: *"gates on the ACTIVE-location probe, not on the picker page"* and *"opens the editor when the probe finds an active location the picker page does not show"*. The truncation banner (`:204-208`) is a good extra — a bounded read that reports healthy while being half-complete is the ADR-048 decision 5 failure mode, and it is now stated rather than discovered.
4. **`settings-page.test.tsx` is extended** with the admin/non-admin pair for the tile, carrying a docblock explaining that the gate lives at the mount site and that `sourcing-rules-tile.test.tsx` would pass with it deleted. That is exactly the Mailer/PostHog precedent.

The admin gating itself remains correct and is not re-raised — `apps/api/src/oms/http/oms-sourcing-rules.controller.ts:72` carries a class-level `@Roles('admin')` covering the `@Get()`, so the `WhoDecidesTile` ungated precedent genuinely does not transfer.

Stack-wide, both items are resolved at this head: `--bg-subtle` has 0 declarations and 0 uses (replaced by the catalogued `--bg-surface-muted`, declared light and dark), and `features/oms` is registered in both `no-restricted-imports` groups for all five canonical subdirectories plus `check-ui-vocabulary.mjs`'s `SCAN_ROOTS`.

### Findings

**SUGGESTION** — `apps/web/src/pages/oms/sourcing-rules-page.tsx:90-117`
> The loading and error arms OR three queries together, so a failure in any one of them renders one generic card and the retry refetches all three. Fine today — `errorMessage` claims only "connections or locations", which is true of every combination. Worth revisiting if a fourth read joins, since the sentence stops being able to name what failed.

### Verdict
✅ **Approve**

Mergeable: `MERGEABLE` / `CLEAN` into `3061-sourcing-rules-states`. **No CI checks are reported on this branch** — absent, not passing.

---

PR #3235, opened 2026-09-14
test(oms): sourcing-rules copy, tile and a structural coverage guard (#3062)

### LINKED ISSUE #3062: test(oms): sourcing-rules component + hook tests

norbert-kulus-blockydevs:
## Summary

The per-component and per-hook tests landed with the slices they belong to (#3056 → #3060), so this closes the three gaps left over and adds the guard that keeps the *next* addition from shipping untested.

**Stack position:** base is `3060-sourcing-rules-mount` (#3234). This is the tip — the epic's roll-up PR to `main` follows.

## What this adds

**`sourcing-rules-tile.test.tsx`** — the one thing that matters about a tile which is the **only** way into a screen (#3060 adds no sidebar entry): a wrong `href` makes the whole page unreachable, and nothing else in the app would notice.

**`sourcing-rule.copy.test.ts`** — pins the LOOSE-lookup rule in both directions. Every value this build knows carries a label and a hint; an unrecognised one degrades to the raw string with a `null` hint. A generic sentence there would claim knowledge of a rule this build cannot evaluate, and a blank cell would hide the row the operator has to find in order to delete it.

**`__tests__/oms-test-coverage.test.ts`** — every module under `api` / `components` / `hooks` / `lib` has a colocated test, or an exemption that **names where** its behaviour is asserted instead.

## The guard is about EXISTENCE, and says so

A test file that asserts nothing would pass it. What it buys is that a component or hook added later cannot ship with no test at all and no signal — the failure this programme has met before, a control shipped, exported and called by nothing.

It fails on **two** drift classes, both verified red before this landed:

1. An uncovered module (`lib/temp-uncovered.ts` → `expected [ 'lib/temp-uncovered.ts' ] to deeply equal []`).
2. A **stale exemption** — including the one that rots quietly: an entry naming a module that has *since* grown its own colocated test, which keeps passing while asserting something that stopped being true.

Two exemptions were removed during authoring for exactly that reason once the tighter check was in place.

## Test plan

- [x] `vitest run src/features/oms src/pages/oms` - 17 files, 125 tests, all pass
- [x] Coverage guard verified RED on an uncovered module and on a stale exemption

Refs #3055
Closes #3062


piotrswierzy (commented):
## Tech Lead review

A test-only tip PR, so the question is whether the guard can actually go red — and it can. I read the predicates rather than trusting the body: `uncovered` and `stale` each build real sets from `readdirSync` and compare them, so adding `lib/foo.ts` with no `lib/foo.test.ts` fails the second test, and an exemption naming a module that has since grown a colocated test fails the third. I also verified every one of the nine `WITHOUT_OWN_TEST` reasons against the tip tree — including the one most likely to be a fiction, `hooks/use-sourcing-rule-query.ts` claiming coverage from `use-sourcing-rules-query.test.tsx` (singular hook, plural test file): that file genuinely imports **both** hooks and carries a `describe('useSourcingRuleQuery')` block with two cases. `sourcing-rule.copy.test.ts` is likewise not a tautology — it iterates `SOURCING_*_VALUES` from `sourcing-rule-vocabulary.ts` and asserts each has a label and a hint in `sourcing-rule.copy.ts`, so a new vocabulary member added without copy goes red. That is a real cross-module contract, not a string echo. Two gaps: the walk is non-recursive, and the one gating assertion the stack is actually missing is not here.

### Findings

**[IMPORTANT]** — `apps/web/src/features/oms/__tests__/oms-test-coverage.test.ts:59-74`
> `sourceModules()` and `testFiles()` call `readdirSync(join(FEATURE_ROOT, dir))` one level deep and filter on `/\.tsx?$/`, which drops directory entries. So `components/priority/location-picker.tsx` — or any module in a subfolder a later slice adds — is invisible to the walk and ships uncovered with the guard still green. The file's own docblock frames this as *"the one signal that nobody looked"*; a walk that cannot see a file is the "check that cannot fail" shape one directory down. `readdirSync(dir, { recursive: true })` (Node ≥ 20) or a `withFileTypes` recursion closes it; if a flat feature is the deliberate constraint, assert it — a test that no entry under `WALKED` is a directory fails loudly the day someone nests, which is the same guarantee at lower cost.

**[IMPORTANT]** — missing: `apps/web/src/pages/settings/settings-page.test.tsx`
> The one assertion this stack genuinely lacks is that #3234's `{isAdmin ? <SourcingRulesTile /> : null}` gate works. `settings-page.test.tsx` already pins exactly that pair for Mailer and PostHog (heading **and** `.toolbar-chip`, present for admin / absent for `viewer`), sitting beside the deliberately-inverted Who-decides case — so a fourth gated tile with no entry leaves a file that reads as exhaustive quietly incomplete. `sourcing-rules-tile.test.tsx` here cannot cover it: it renders the component directly, and the gate is at the mount site. This is the PR whose stated job is *"the three gaps left over"*; it is the fourth.

**[SUGGESTION]** — `apps/web/src/features/oms/__tests__/oms-test-coverage.test.ts:62-66,90` and `:85`
> Two weaker spots in the same guard. The `try { … } catch { continue }` silently tolerates a `WALKED` entry that no longer exists, so renaming `components/` to something else shrinks the walk with no failure; and the `> 10` floor is loose enough to survive losing `api` (4 modules) or `lib` (5) outright — the tip carries 23. Asserting each `WALKED` directory exists, rather than swallowing its absence, converts both into one loud failure.

**[SUGGESTION]** — `apps/web/src/features/oms/components/sourcing-rules-tile.test.tsx:26-29`
> *"says what the screen decides, not merely that it exists"* asserts a substring of the literal in `SOURCING_RULES_TILE_COPY.description`, which the component renders directly from that constant — so it can only fail when someone edits the copy, and it fails then whether or not the edit is an improvement. A change-detector rather than a property. The `href` test above it is the one carrying weight and is well chosen (with no nav entry, a wrong href makes the screen unreachable and nothing else notices). If the intent is "the tile explains the screen rather than naming it", assert the shape — e.g. that the description is non-empty and is not merely the title.

**Positive observations.** The stale-exemption test is the non-obvious half and it is the right one: an entry that keeps passing while asserting something that stopped being true is how these lists rot, and the body reports two exemptions were actually removed during authoring once it was in place — which is the evidence that it bites. Framing the guard as **existence only**, in the docblock, is honest about what it does not buy. `WITHOUT_OWN_TEST` being a `Record<string, string>` with a fourth test asserting no reason is blank makes "name where instead" structural rather than a convention.

**CI.** No checks reported on `3062-sourcing-rules-tests` — stack-targeted, and the workflow triggers on `pull_request: branches: [main]`. Absent, not passing. `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN` against `3060-sourcing-rules-mount`.

### Verdict
🔄 **Approve with changes**

---

norbert-kulus-blockydevs:
Addressed all four findings in bbee3bc2f. The guard changes were verified red-first - I broke the thing each one guards, confirmed the spec fails, then restored.

**[IMPORTANT] Non-recursive walk (`oms-test-coverage.test.ts:59-74`)** - fixed. `sourceModules()` / `testFiles()` now share one `walk(dir)` that recurses with `readdirSync(..., { withFileTypes: true })`, so anything nested under a `WALKED` directory is seen. I took the recursion rather than the "assert no entry is a directory" alternative: a flat feature is not a constraint anyone has declared, and refusing the first nested module would push back on a legitimate structure rather than covering it.

Red-first evidence, using the review's own example. Created `components/priority/location-picker.tsx` with no sibling test:
```
× every module has a colocated test, or a named reason not to
AssertionError: expected [ Array(1) ] to deeply equal []
+   "components/priority/location-picker.tsx",
Tests  1 failed | 4 passed (5)
```
Adding `components/priority/location-picker.test.tsx` beside it returned the file to `5 passed`, so the recursion reports nested test files too rather than only nested modules. Both fixtures removed.

**[IMPORTANT] Missing `settings-page.test.tsx` entry** - fixed. Added the admin/viewer pair for the Sourcing rules tile, mirroring the Mailer and PostHog cases exactly (heading and `.toolbar-chip`, present for admin, absent for `viewer`), with a docblock naming why the tile's own test cannot cover it - the gate is at the mount site.

**[SUGGESTION] Swallowed `WALKED` absence and a loose floor (`:62-66,90` and `:85`)** - fixed, in the shape you suggested. The `try { … } catch { continue }` is gone, so a missing directory throws from `readdirSync`, and a new test asserts every `WALKED` entry exists and still contributes at least one module. The `> 10` floor stays as the coarse half with a comment saying so.

Red-first evidence, both shapes. Renaming `lib` away:
```
× walks a non-empty set of modules
× walks every declared directory, and each one still holds a module
× every module has a colocated test, or a named reason not to
× carries no stale exemption
Error: ENOENT: no such file or directory, scandir '.../features/oms/lib'
Tests  4 failed | 1 passed (5)
```
A directory that exists but holds no modules (simulated by temporarily adding `__tests__` to `WALKED`, which contains only a test file):
```
× walks every declared directory, and each one still holds a module
AssertionError: expected 0 to be greater than 0
Tests  1 failed | 4 passed (5)
```
Both reverted; the file is back to `5 passed`.

**[SUGGESTION] Tile description change-detector (`sourcing-rules-tile.test.tsx:26-29`)** - fixed as you described. The assertion is now the property rather than the sentence: the component renders `SOURCING_RULES_TILE_COPY.description` (a mis-wired tile renders nothing), and that constant is not the title and is longer than it. A copy edit no longer turns it red; a description degraded to a restatement of the name does.

Checks: `pnpm --filter @openlinker/web type-check` clean, `eslint` on the three changed files clean, and the full `@openlinker/web` vitest run is green (528 files, 5685 tests). No `scripts/check-*.mjs` was touched, so `check:invariants` is unaffected.

norbert-kulus-blockydevs:
## Stack-tip gate after the round-3 fixes

Every PR in the stack now carries every PR below it. Heads after the cascade:

| PR | branch | head |
|---|---|---|
| #3226 | `3056-sourcing-rules-api-client` | `2d93a01f7` |
| #3228 | `3057-sourcing-rules-table` | `6589a7009` |
| #3229 | `3058-sourcing-rule-dialog` | `494b68b77` |
| #3230 | `3059-sourcing-rule-delete` | `15b30e079` |
| #3233 | `3061-sourcing-rules-states` | `2750231a6` |
| #3234 | `3060-sourcing-rules-mount` | `d6eb707ac` |
| #3235 | `3062-sourcing-rules-tests` | `523fb4afb` |

Five merges, no conflicts - including `sourcing-rules-table.tsx`, which #3228 and #3233 both edit.

Run at this head (`523fb4afb`):

- `pnpm --filter @openlinker/web type-check` - clean
- `pnpm check:invariants` - exit 0
- `vitest run` over all of `apps/web` - **528 files, 5706 tests, 1 skipped, 0 failures**

Each PR carries its own reply comment covering its findings. Two of them push back on a premise rather than making the change, both with evidence in the comment: #3229 on which direction an edited window can loosen the ceiling (`resolveSplitCeiling` takes the most restrictive of the active set, so adding to that set is monotonically stricter-or-equal - the loosening case is the mirror, and that is what is now tested), and #3233 on disabling Retry during a refetch (TanStack v5 `refetch()` resets `status` to `pending`, so the `isLoading` branch is read first and `disabled` would be unreachable; replaced with a test pinning the branch order).

One thing this stack still cannot fix from inside itself: `3055-sourcing-rules-admin-ui` has no PR to `main`, and `docs/plans/mockups/sourcing-rules-3055.html` is not in `main` either (#3119 merged into `3055`, not into `main`). `git merge-base --is-ancestor origin/3055-sourcing-rules-admin-ui origin/main` answers no. So all seven PRs are currently rooted on a branch with no route to `main`.

piotrswierzy (commented):
## Tech Lead re-review (`6570b3a` → `523fb4a`)

Both IMPORTANTs are addressed, and the walk fix goes further than asked. `walk()` now recurses via `readdirSync(..., { withFileTypes: true })` and builds feature-root-relative paths, so a module nested in a future `components/priority/` subfolder is visible (`oms-test-coverage.test.ts:60-90`). The try/catch that used to swallow a missing `WALKED` entry is deliberately gone, so a directory renamed away throws rather than silently shrinking the walk — and a new `it('walks every declared directory, and each one still holds a module')` asserts `statSync(...).isDirectory()` **and** `modulesIn(dir).length > 0` per entry, which closes the complementary hole the `> 10` total floor could never catch (losing a whole directory while its siblings carry the count). The docblock names both properties as load-bearing rather than leaving them to be re-derived. The settings-page admin-gating assertion for the tile also now lives here. And `sourcing-rules-tile.test.tsx` was improved in passing: it asserts the tile renders `SOURCING_RULES_TILE_COPY.description` *and* that the constant is an explanation rather than the title again, instead of echoing a string literal that would go red on any copy improvement.

I re-confirmed the non-vacuity properties from last round at this head: the uncovered-module and stale-exemption tests both build real sets and compare them, the stale test catches both directions (module gone, and module that has since grown a colocated test), and all nine `WITHOUT_OWN_TEST` reasons still name a file that exists in the tree at `523fb4a`. Every module under `api/`, `components/`, `hooks/` and `lib/` is either colocated-tested or exempted. Stack-wide, `--bg-subtle` is gone (0 declarations, 0 uses) and `features/oms` is registered in both `no-restricted-imports` groups plus `check-ui-vocabulary.mjs`'s `SCAN_ROOTS`.

### Findings

**[IMPORTANT]** — `apps/web/src/pages/settings/settings-page.test.tsx:124`, `:143`, `:194`, `:203`
> The merge of `origin/3060-sourcing-rules-mount` into this branch **duplicated the tile's test pair**. `settings-page.test.tsx` at this head contains `it('shows the Sourcing rules tile for an admin session')` twice (`:124` and `:194`) and `it('never renders the Sourcing rules tile for a non-admin session')` twice (`:143` and `:203`) — one pair from #3234's `9f681a8c`, one from this PR's `bbee3bc2`, textually near-identical apart from the fixture user (`user_4`/`viewer3@example.com` vs `user_2`/`viewer@example.com`) and the docblock. That is the textual-merge-resolution hazard the repo already has a lesson about: both sides added the same assertion to the same file at different offsets, so git took both.
>
> It is not merely untidy. Two `it` blocks with identical names in one `describe` make a failure report ambiguous about which one failed, and the second non-admin copy reuses `user_2`/`viewer@example.com` — the same fixture identity the pre-existing PostHog non-admin test at `:174` uses — so a future reader cannot tell the two apart by their arrange step either. Keep **one** pair (#3234's, which carries the fuller docblock explaining why the gate can only be asserted at the mount site and why it is admin-gated unlike `WhoDecidesTile`) and drop the duplicate, then re-check the `+72` diffstat on this file collapses back to the ~`+34` this PR actually owns.

### Verdict
🔄 **Approve with changes** — merge once the duplicated test pair is removed. Nothing about the guard itself is blocked.

Mergeable: `MERGEABLE` / `CLEAN` into `3060-sourcing-rules-mount`. **No CI checks are reported on this branch** — absent, not passing; a duplicated `it` name would not fail a build in any case, which is why it needs the human resolution.

---

norbert-kulus-blockydevs:
## Re-review addressed (`523fb4a` -> `1feee77`)

One finding in this round, and it was correct as written. Fixed in `1feee77`.

### [IMPORTANT] Duplicated tile test pair in `settings-page.test.tsx` - FIXED

Confirmed at the branch tip before changing anything. `4adbe04f1` carried four `Sourcing rules` blocks, not two:

```
124:  it('shows the Sourcing rules tile for an admin session'
143:  it('never renders the Sourcing rules tile for a non-admin session'
194:  it('shows the Sourcing rules tile for an admin session')
203:  it('never renders the Sourcing rules tile for a non-admin session')
```

Provenance matches your reading exactly. `git show origin/3060-sourcing-rules-mount:apps/web/src/pages/settings/settings-page.test.tsx` already contains the pair at `:124`/`:143`, so that is #3234's, arriving through the base branch; `:194`/`:203` is this PR's own `bbee3bc2`. Both sides appended to the same `describe` at different offsets and the textual merge kept both.

I removed **this PR's** pair (lines 187-220, the docblock plus both blocks) and kept #3234's, for the two reasons you gave:

- #3234's docblock covers both halves - why the gate can only be asserted at the mount site, *and* why this tile is admin-gated unlike `WhoDecidesTile` (the `@Roles('admin')` class-level gate on the sourcing-rules API, #2953). The copy here only covered the first half.
- #3234's non-admin case uses `user_4` / `viewer3@example.com`, distinct from every other fixture in the file. This PR's reused `user_2` / `viewer@example.com`, which is the identity the PostHog non-admin test at `:171` already uses.

The diff is a pure 34-line deletion, no edits.

**On the diffstat.** You expected the file to collapse back to roughly `+34`. It collapses to **zero**: `git diff --stat origin/3060-sourcing-rules-mount -- apps/web/src/pages/settings/settings-page.test.tsx` is now empty, i.e. the file is byte-identical to the base. That is the correct end state rather than a surprise - the `+72` you measured at `523fb4a` was against a merge-base predating #3234's pair; once the merge advanced the base to include it, the assertion is owned by #3234 and this PR's honest contribution to that file is nothing. The tile's gate is still pinned, just not twice, and not by this PR.

### The ambiguity claim, demonstrated rather than asserted

Your argument that two identically-named `it` blocks make a failure report ambiguous is the reason this is not merely untidy, so I reproduced it. With the duplicates restored and the mount-site gate deliberately broken (`{isAdmin ? <SourcingRulesTile /> : null}` -> `<SourcingRulesTile />`):

```
 × never renders the Sourcing rules tile for a non-admin session  24ms
 × never renders the Sourcing rules tile for a non-admin session  26ms
FAIL  ... > SettingsPage > never renders the Sourcing rules tile for a non-admin session
FAIL  ... > SettingsPage > never renders the Sourcing rules tile for a non-admin session
Tests  2 failed | 13 passed (15)
```

The fully-qualified name is printed twice with nothing to tell the two apart. After the fix the same break reports exactly one.

### Red-first: the surviving pair still bites

The thing worth proving about a deletion is that the assertion did not go with it. Both halves of the gate were broken independently against the post-fix file, and each one goes red naming the right test and the right line:

| Break | Result |
|---|---|
| `:118` `{isAdmin ? <SourcingRulesTile /> : null}` -> always render | `× never renders the Sourcing rules tile for a non-admin session`, failing at `:156` (`queryByRole('heading', …)`) |
| `:34` `{isAdmin ? <span className="toolbar-chip">…` -> always render | same test, failing at `:158` (`queryByText(…, { selector: '.toolbar-chip' })`) |

Both restored afterwards; `git diff origin/3060-sourcing-rules-mount -- apps/web/src/pages/settings/settings-page.tsx` is empty, so no source file was touched by the experiment.

### The coverage guard is untouched, and re-verified at this head

Nothing in this round asked for a guard change and I made none. Since the recursive walk is this PR's headline I re-confirmed it still bites at the new head, using the exact example from your round-1 finding - dropped `components/priority/location-picker.tsx` in with no colocated test:

```
× every module has a colocated test, or a named reason not to
+   "components/priority/location-picker.tsx",
Tests  1 failed | 4 passed (5)
```

Caught, and named by its feature-root-relative path, so the nested-module hole from round 1 is genuinely closed rather than closed-looking. Scratch file removed.

I also checked the guard has no dependency on the file I edited: `grep -n "settings-page" oms-test-coverage.test.ts` returns nothing, and no `WITHOUT_OWN_TEST` reason names it, so deleting a settings-page test could not silently satisfy an exemption.

### Verification

- `npx vitest run src/features/oms src/pages/settings/settings-page.test.tsx` - 17 files, 145 tests, all passing. `settings-page.test.tsx` is 13 tests, down from 15.
- No duplicate `it` names remain: `grep -o "it('[^']*'" … | sort | uniq -d` is empty.
- `pnpm --filter @openlinker/web type-check` - exit 0.
- `npx eslint apps/web/src/pages/settings/settings-page.test.tsx` - exit 0.
- `pnpm check:invariants` - exit 0, including `check-ui-vocabulary` (177 files across 7 of 7 feature folders, `features/oms` among them) and `check-sourcing-rule-vocabulary-mirror` (4 vocabularies plus the after-action ranking).

### CI

Still `no checks reported on the '3062-sourcing-rules-tests' branch` - absent, not passing, as you noted; the workflow triggers on `pull_request: branches: [main]` and this targets `3060-sourcing-rules-mount`. `MERGEABLE` / `CLEAN`. As you said, a duplicated `it` name would not have failed a build in any case.

No branch was merged down into this one - the stack is current as the owner left it.

piotrswierzy (approved):
## Tech Lead re-review (`523fb4af` → `ab25be93`)

**The single open finding is closed.** The merge regression is gone: `apps/web/src/pages/settings/settings-page.test.tsx` now carries exactly **one** `Sourcing rules` tile test pair — `shows the Sourcing rules tile for an admin session` at `:124` and `never renders the Sourcing rules tile for a non-admin session` at `:143`, each appearing once in the file. The surviving pair is the fuller-docblock one I recommended keeping (it explains that the gate is the `{isAdmin ? … : null}` at the mount site and therefore only assertable here, and why this tile is admin-gated where `WhoDecidesTile` is not). The duplicate's fixture-identity collision went with it: the non-admin case now uses `user_4` / `viewer3@example.com`, distinct from the PostHog test's `user_2` / `viewer@example.com` at `:174` and from `WhoDecidesTile`'s `user_3` / `viewer2@example.com` at `:106`. The PR's own commit for this is `1feee778`; everything else in the range is base-branch catch-up through `3060-sourcing-rules-mount` (including #3230's `3bd9e58f`, reviewed separately).

The previously-closed coverage-guard finding still holds, verified in code at `apps/web/src/features/oms/__tests__/oms-test-coverage.test.ts`:

- **The walk is recursive** — `readdirSync(join(FEATURE_ROOT, dir), { withFileTypes: true })` recurses on `entry.isDirectory()` and returns feature-root-relative paths, so a module nested one directory down (`components/priority/location-picker.tsx`) is seen. The docblock states that property and why.
- **No `try`/`catch`** — a renamed-away `WALKED` directory throws out of `readdirSync`/`statSync` rather than silently shrinking the walk.
- **The per-entry assertion is the one that closes the hole** the `> 10` total floor could not: `expect(statSync(join(FEATURE_ROOT, dir)).isDirectory()).toBe(true)` **and** `expect(modulesIn(dir).length).toBeGreaterThan(0)` for every `WALKED` member, so losing a whole directory — or emptying one while the exemption list keeps naming files in it — is a loud failure instead of a count the siblings carry on their own.
- **Non-vacuous.** Both `uncovered` and `stale` are real set comparisons over the walked set (`expect(...).toEqual([])`), and `stale` checks both directions — an entry naming a module that no longer exists, and one naming a module that has since grown its own test. All nine `WITHOUT_OWN_TEST` reasons name a file present at the tip (`api/sourcing-rules.types.ts`, `api/sourcing-rules.query-keys.ts`, `components/sourcing-rule-locked-dialog.tsx`, the four mutation hooks, `hooks/use-sourcing-rule-query.ts`, `hooks/use-inventory-locations-for-rules-query.ts`) and each names where its behaviour is asserted instead. Every remaining module under `api/`, `components/`, `hooks/`, `lib/` has a colocated `.test.ts`/`.test.tsx`.

One process note, stated once for the stack rather than as a finding on each PR: the stack's root branch `3055-sourcing-rules-admin-ui` still has **no open PR**, so nothing in this chain currently has a route to `main`.

### Findings

### Verdict
✅ **Approve**

CI: `no checks reported on the '3062-sourcing-rules-tests' branch` — absent, not passing; this stack targets stack branches and has reported none throughout. `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN` against `3060-sourcing-rules-mount`.

---

PR #3236, opened 2026-09-14
feat(listings): refuse a disproportionate manual price override server-side

### LINKED ISSUE #3222: [TASK] API — refuse a disproportionate manual price override server-side on price-change edit
## Problem / Context

Part of #3139 (epic). Raised by the round-3 review on #3165, which asked for it to be filed against the epic rather than fixed in that PR.

`PATCH /listings/price-changes/:id/edit` publishes an operator-typed absolute price. Today the only thing standing between a typo and a wrong live price is a sentence in the browser:

- `apps/web/src/features/price-changes/components/edit-price-change-dialog.tsx` — `DEVIATION_WARN_THRESHOLD = 0.3` produces a **non-blocking warning** (`tone: 'warn'`) when the typed value deviates more than 30% from the rule-computed price. The operator can read past it and confirm.
- `apps/api/src/listings/http/dto/edit-price-change.dto.ts` — `manualPriceOverride` carries `@IsNumber({ maxDecimalPlaces: 4 })`, `@IsPositive()` and `@Max(1_000_000_000)`. That ceiling is explicitly documented in the file as a **pathological-input guard** sized to `numeric(14,4)` so `1e30` fails with a 400 instead of an unhandled 500 at the database. It is not a deviation guard: `39900` for a `399` item passes it trivially.

So `39900` instead of `399` publishes to a live marketplace or storefront on one click, and nothing server-side refuses it. The browser is not a trust boundary — the same endpoint is reachable by curl and by MCP.

#2610 set the precedent in this exact shape: a value that silently produces a wrong published price is refused **server-side as well as in the form**, because the raw JSON editor, curl and MCP all bypass the form. `ConnectionService.validateStockAndPricingConfig` / `validatePricingRuleConfig` (`apps/api/src/integrations/application/services/connection.service.ts:280`) are that refusal for `stockSafetyBuffer` / `stockZeroThreshold` / `pricingRule`.

## Proposed Solution

Refuse a **proportionate** deviation in `PriceChangesService.edit` (`libs/core/src/listings/application/services/price-changes.service.ts:214`), not in the DTO.

The seam matters: `edit` already calls `loadActionable(episodeId, input.expectedVersion)` and therefore holds the episode, with its `computedNewAmount` / `computedOldAmount` / currency in hand. The DTO has no episode context at all — which is exactly why its existing `@Max` had to be documented as *"not currency-aware … bounds pathological input rather than expressing a real per-currency limit"*. A bound that compares the typed value against the episode's own computed price can only live where the episode is loaded.

Shape to follow:

- A pure rule beside the other pure listings rules (`applyPricingRule`, `applyStockSafetyBuffer`, `checkRequiredToSell` — see `engineering-standards.md § The pure-rule exception`), so it is unit-testable without I/O and so the browser mirror can be held against it.
- A new domain exception in `libs/core/src/listings/domain/exceptions/`, mapped to **422** by the existing price-changes controller filter alongside the other resolution-conflict exceptions.
- The threshold is a constant in core, not an operator setting, unless a reviewer asks otherwise. Note it need not equal the browser's `0.3`: the browser warns at 30%, the server may refuse at a wider band, and the two numbers answering different questions is fine as long as **the server band is never narrower than the browser's** — a mirror stricter than the gate refuses work the destination would have accepted (#2240's rule).
- If the two numbers are made equal, hold them with a `scripts/check-*-mirror.mjs` under `pnpm check:invariants`, following the 23 existing guards of that shape. Do not leave two hand-maintained copies.

Explicitly out of scope: making the DTO's `@Max(1_000_000_000)` currency-aware. It is doing its own job correctly and should stay.

## Classification

**Type**: CORE + Interface
**Layer**: Domain (pure rule) / Application (`PriceChangesService.edit`) / Interface (exception → 422)
**File(s)**:
- `libs/core/src/listings/application/services/price-changes.service.ts` (`edit`, line 214)
- `libs/core/src/listings/domain/exceptions/`
- `apps/api/src/listings/http/dto/edit-price-change.dto.ts` (context only — not the fix site)
- `apps/web/src/features/price-changes/components/edit-price-change-dialog.tsx` (`DEVIATION_WARN_THRESHOLD`)

## Dependencies

- #3145 / PR #3162 must merge first — `PriceChangesService.edit` and the `edit` endpoint are introduced there.

## Assumptions

- The refusal is proportionate (typed value vs the episode's computed price), not absolute. An absolute per-currency ceiling would need a currency table this endpoint does not have.
- 422 is the right status: the request is well-formed and the episode is actionable; the *value* is refused. 400 is already taken by DTO-shape failures on this route.
- The browser keeps its non-blocking 30% warning. Server-side refusal replaces neither the warning nor the confirm step — it is the backstop for the paths that never render a form.

## Acceptance Criteria

- [ ] `PATCH /listings/price-changes/:id/edit` with a `manualPriceOverride` whose deviation from the episode's computed price exceeds the server band is refused with a distinguishable status + machine-readable code, and **no** apply job is enqueued.
- [ ] The same request inside the band still succeeds and enqueues exactly as it does today.
- [ ] The refusal is raised in `PriceChangesService.edit`, not in the DTO, and is reachable by curl with no browser involved.
- [ ] The server band is provably **not narrower** than the browser's warn threshold (asserted by a test or by a mirror script, whichever the implementation chooses).
- [ ] The existing `@Max(1_000_000_000)` pathological-input guard on the DTO is unchanged.

norbert-kulus-blockydevs:
## Summary

Part of #3139 (epic). **Stack position: tenth** — branches from `3151-auto-applied-note-and-dialog` (#3168), because the `PriceChangesService.edit` this guards is #3162's and the base moves to `main` as the stack lands.

Raised by the #3165 round-3 review, which asked for it to be filed against the epic rather than fixed in that PR.

`POST /listings/price-changes/:id/edit` publishes an operator-typed absolute price. Until this, the only thing between a typo and a wrong live price was a sentence in the browser: `DEVIATION_WARN_THRESHOLD = 0.3` produces a **non-blocking** warning the operator can read past and confirm. The DTO's `@Max(1_000_000_000)` is not a second line — it is a pathological-input guard sized to the `numeric(14,4)` column, documented in its own file as explicitly not currency-aware, and `39900` for a `399` item clears it trivially. curl and MCP never see the form at all.

## Design

- **`checkPriceOverrideBound`** — a pure rule in `libs/core/src/listings/domain/types/price-override-bound.types.ts`, beside the `applyPricingRule` / `checkRequiredToSell` precedents (`engineering-standards.md § The pure-rule exception`). No I/O, no injected dependency.
- **Called from `PriceChangesService.edit`, not from a DTO decorator.** The bound is proportionate to the episode's own `computedNewAmount`, which only exists once the episode is loaded — the DTO has no episode context, which is exactly why its existing `@Max` had to be documented as not expressing a real limit.
- **Checked after the claim**, so a refused edit releases it like any other failure rather than leaving the episode held.
- **`PriceChangeOverrideOutOfRangeException` → 422.** The request is well-formed and the episode is actionable; the *value* is refused. 400 on this route is already the DTO-shape failure. The exception carries `outcome`, `attempted`, `computedAmount` and `limit`, so the operator is told the bound rather than just "no".

## Two decisions worth reviewing

**Why 10×, either way.** It is chosen against the failure it exists to catch — a misplaced decimal point or a duplicated digit group, both of which move a price by at least an order of magnitude. Tighter starts refusing legitimate corrections (a genuinely mispriced catalogue item can need 3×); looser stops catching the 1000×-too-high typo. The reasoning is on the constant, not only here.

**Why it is wider than the browser's 30%.** A mirror stricter than the gate refuses work the destination would have accepted (#2240), and the two numbers answer different questions: the form asks *did you mean this?*, the server asks *can this possibly be a real price?*. The form must never be able to submit something the server then refuses, so the server's band has to contain the form's — **asserted by a test** rather than left to whoever edits one of the two constants next.

**No baseline means no refusal.** A missing, zero or non-finite `computedNewAmount` yields `ok`: with no computed price there is nothing to be disproportionate *to*, and refusing there would block the operator from correcting precisely the episode whose price the system failed to produce. `@IsPositive()` still bounds the override itself.

## Test plan

- [x] `pnpm --filter @openlinker/core test -- price-override-bound` — 7 cases, including the 100× example from the review, both bounds inclusive, every no-baseline shape, and the containment assertion against the browser threshold.
- [x] `pnpm --filter @openlinker/core test -- price-changes.service.spec` — 44 pass, including three new: the refusal enqueues nothing and releases the claim, it carries the crossed bound, and a plausible 3× correction still goes through.
- [x] `pnpm --filter @openlinker/core type-check`, `pnpm --filter @openlinker/api type-check`, `pnpm check:invariants` — clean.

Closes #3222


piotrswierzy (commented):
## Tech Lead review

This genuinely closes the #3165 finding. I verified the three things that mattered: it is a real server-side refusal (`PriceChangesService.edit`, reached by curl and MCP alike, not a DTO decorator); it is **wider** than the browser, so the `#2240` trap is avoided rather than merely mentioned; and it does **not** promote the browser's `>30%` warn into a gate — `edit-price-change-dialog.tsx` is untouched, `canConfirm` still only tests `tone !== 'error'`, and the deviation branch still returns `tone: 'warn'`. What is new versus #3162 is also real: `@Max(1_000_000_000)` is currency-blind and episode-blind by its own comment, so `39900` for a `399` item clears it; this is the first bound proportionate to the episode. Coverage is complete for the surface — `manualPriceOverride: true` has exactly one writer (`edit`); `bulkAccept` publishes `episode.computedNewAmount` with `manualPriceOverride: false`, so there is no second door. The pure-rule placement, the after-claim ordering and the 422-not-400 choice are all correct. My most important concern is that the claim release on the refusal path bypasses the helper this file wrote for exactly that purpose, which can turn a deliberate 422 into an opaque 500.

### Findings

**[IMPORTANT]** — `libs/core/src/listings/application/services/price-changes.service.ts:237`
> The refusal path calls `await this.episodes.releaseClaim(episodeId)` directly instead of `await this.releaseClaims([episodeId])`. That helper exists in this same file and its docblock states the rule: *"Releases every claim in `episodeIds`, logging (never throwing) on failure — a release is best-effort cleanup, not itself a gate."* Every other abort path honours it — `enqueueOrReleaseClaim` (:543), the `bulkAccept` validation-loop catch (:345), the fan-out catch (:397). The direct call is unguarded, so a transient `releaseClaim` failure rejects with the *release* error, `PriceChangeOverrideOutOfRangeException` is never constructed, `wrapDomainErrors` falls through to `throw error`, and the operator gets a 500 with no mention of the bound — the one thing this PR exists to tell them. Fix is one line: `await this.releaseClaims([episodeId]);`.

**[IMPORTANT]** — `libs/core/src/listings/domain/types/__tests__/price-override-bound.types.spec.ts:44`
> The containment claim is pinned on one side only. Both the file comment and the PR body say the assertion exists *"rather than left to whoever edits one of the two constants next"* — but the spec hardcodes its own `const BROWSER_WARN_THRESHOLD = 0.3` and never reads `DEVIATION_WARN_THRESHOLD` from `edit-price-change-dialog.tsx:53`. Editing the browser constant fails nothing; only a downward edit of `PRICE_OVERRIDE_MAX_FACTOR` is caught. `apps/web` cannot import `@openlinker/core` (#591), which is precisely the situation this repo answers with a `scripts/check-*-mirror.mjs` under `pnpm check:invariants` — `check-stock-and-pricing-preview-mirror.mjs` is the closest precedent and covers the same subject (a browser-side pricing mirror of a core rule), with `check-parameter-restriction-mirror.mjs` and `check-shipping-tax-split-mirror.mjs` alongside it. A ~15-line script that greps both constants and asserts `1 + browser < core` would make the sentence true. Until then, please soften the comment to say which half it pins.

**[SUGGESTION]** — `apps/api/src/listings/http/price-changes.controller.ts:284`
> `PriceChangeOverrideOutOfRangeException` carries `outcome`, `attempted`, `computedAmount` and `limit`, and the PR body cites them as the reason *"the operator is told the bound rather than just 'no'"* — but `UnprocessableEntityException(error.message)` flattens all four into an English sentence, so no caller can reach them and nothing outside the service spec reads them. The repo's stated rule is the opposite (`architecture-overview.md` § 22 Returns, #2376): `ReturnCustodyTransitionError.reason` "is emitted as a FIELD, exactly as `ReturnNotAttributedError.trigger` already is, because a client that parsed the message would break on the first reword." Emitting `{ message, outcome, limit, computedAmount }` in the 422 body would let the dialog render the bound inline instead of surfacing a paragraph in a toast.

**[SUGGESTION]** — `libs/core/src/listings/domain/types/price-override-bound.types.ts:57`
> `PriceOverrideBoundResult` makes two illegal states representable — `{ outcome: 'ok', limit: 5 }` and `{ outcome: 'too-high' }` both type-check — which is what forces the `bound.limit as number` cast at `price-changes.service.ts:242`. A discriminated union removes both: `type PriceOverrideBoundResult = { outcome: 'ok' } | { outcome: Exclude<PriceOverrideBoundOutcome, 'ok'>; limit: number }`. `TerminaliseRoutingDecisionInput` is the shipped precedent (`architecture-overview.md` § 26) — a discriminated union whose arms are derived from the state union, so an illegal combination does not compile. Worth noting while you are there: because the function is exported from the `@openlinker/core/listings` barrel, a future caller can pass a `NaN` override and get `ok` (both comparisons are false); `@IsNumber()` rejects NaN on this route today, but the rule itself does not.

**[SUGGESTION]** — `apps/api/src/listings/http/price-changes.controller.ts:181-187`
> Nothing tests the 422. The service spec asserts the exception, the repository int-spec (`listings-price-change-episode.int-spec.ts`) never exercises `edit`, and there is no controller spec — so the `instanceof` → `UnprocessableEntityException` arm is the one link with zero coverage, on a ten-deep stack where several open PRs touch these same files. That matters here specifically because the repo has recorded this exact failure: "#2332 and #2333 were built concurrently and each defined a rival `ReturnNotFoundError` … two same-named classes fail `instanceof` against each other silently — the HTTP filter would answer 500 for a refusal the service raised deliberately" (`architecture-overview.md` § 22). One int-spec case posting `39_900` and expecting 422 would close it. Separately, the route carries `@ApiResponse` for 200/404/409/403 but not 422 — a pre-existing gap (`accept`'s blocked 422 is undocumented too), so fixing it here is optional, but Swagger is what an MCP or curl caller reads.

**[SUGGESTION]** — `libs/core/src/listings/application/services/__tests__/price-changes.service.spec.ts:97`
> The new import sits mid-file, after `buildBatch`, rather than in the import block at the top — `engineering-standards.md` § Import Order puts external → cross-boundary → local at the head of the file. Also, the PR body describes the route as `PATCH /listings/price-changes/:id/edit`; it is `@Post(':id/edit')`.

### Positive observations

`checkPriceOverrideBound` is correctly placed under `engineering-standards.md` § The pure-rule exception (pure, no injected dependency, and it *is* the rule for the type it sits with). The no-baseline arm is the right call and its reasoning is on the function rather than only in the PR — refusing on a baseline the system failed to produce would block the operator from correcting exactly the episode that needs it. Checking after `loadActionable` so the refusal releases the claim is the right ordering. Every route in this controller carries `@Roles(...)`, satisfying `engineering-standards.md` § Route authorization. No migration, so none of the contended timestamps apply.

### CI

`mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`. Only `Scaffolded Adapter Builds` reported — the main CI workflow triggers on `pull_request: branches: [main]` and this PR targets `3151-auto-applied-note-and-dialog`, so lint / type-check / unit / integration are **absent, not passing**. They first run when the base rebases onto `main`.

### Verdict
🔄 **Approve with changes**

Address the two IMPORTANT items before merge; the suggestions can ride the next revision or a follow-up.

---

norbert-kulus-blockydevs:
All six addressed — `6a6ffa4ca`.

**IMPORTANT — the unguarded release.** Fixed, and your reading of the consequence was exact: a transient `releaseClaim` failure would have rejected with the *release* error, so `PriceChangeOverrideOutOfRangeException` was never constructed and the operator got a 500 with no mention of the bound — the one thing the path exists to tell them. Now `await this.releaseClaims([episodeId])`, with the reason on the line rather than only in the commit.

**IMPORTANT — the containment pinned on one side.** You were right that the comment claimed more than the test delivered. `scripts/check-price-override-bound-mirror.mjs` now reads `DEVIATION_WARN_THRESHOLD` from `edit-price-change-dialog.tsx` and `PRICE_OVERRIDE_MAX_FACTOR` from the core types file and asserts `1 + browser < core`, chained into `check:invariants` beside its siblings.

I checked it fails from **both** directions rather than assuming: narrowing the server to `1.2` turns it red, and widening the browser to `12` — the case the old spec could not see at all — turns it red too. Its self-check covers the reader and the inequality itself, including the equal-band case (`0.3` vs `1.3`), which must fail: an equal band still lets the form submit what the server refuses. The spec's comment now says it pins one side and points at the script for the other, instead of overclaiming.

**SUGGESTION — illegal states.** `PriceOverrideBoundResult` is a discriminated union with arms derived from the outcome union, so `{outcome:'ok', limit}` and `{outcome:'too-high'}` no longer type-check and the `limit as number` cast is gone. Your `NaN` observation is recorded on the rule's docblock: a non-finite override yields `ok` because both comparisons are false, which is only safe because `@IsNumber()` rejects it on the one route that reaches it today — a future caller without that guard has to check.

**SUGGESTION — the bound as fields.** The 422 body now carries `outcome`, `attempted`, `computedAmount` and `limit` beside the message, the treatment `ReturnCustodyTransitionError.reason` already has. `@ApiResponse({ status: 422 })` documents them, since Swagger is what an MCP or curl caller reads.

**SUGGESTION — nothing tested the 422.** `apps/api/test/integration/price-change-override-bound.int-spec.ts`, three cases over real HTTP: `39900` against a computed `399` answers **422 with those four fields**, the episode stays claimable so a corrected price straight after succeeds (otherwise the operator is locked out by their own typo), and `1197` still passes. Your framing is quoted in the file header — the recorded #2332/#2333 case of rival same-named classes making exactly this arm answer 500 is why a round trip through the router was the right shape rather than a controller unit test.

One thing that changed while doing it: the spec originally seeded through `PriceChangeEpisodeRepositoryPort`, which `check-cross-context-imports.mjs` denies from a core barrel. Rather than take an allow-list entry I seeded with raw SQL — the spec's subject is the HTTP contract, so coupling it to the repository's API was wrong anyway.

**SUGGESTION — import placement and the verb.** Import moved to the block at the head of the file. You are right that it is `@Post(':id/edit')`, not `PATCH`; the PR body is corrected.

piotrswierzy (approved):
## Tech Lead re-review (`4092e96` → `6a6ffa4c`)

Both IMPORTANT findings are closed, and closed at the right level rather than patched around. The refusal now releases through `releaseClaims([episodeId])` — the guarded helper that catches, warns and never throws — so a transient release failure can no longer replace the deliberate 422 with an opaque 500; the comment at the call site states the reason rather than leaving the next reader to rediscover it, and this is now the only abort path in the file that does not go through the helper... which is to say, there is no longer one. The one-sided containment "guarantee" is now genuinely two-sided: `scripts/check-price-override-bound-mirror.mjs` reads `DEVIATION_WARN_THRESHOLD` from the dialog and `PRICE_OVERRIDE_MAX_FACTOR` from the core types file, textually, and is wired into `check:invariants` beside `check-auto-applied-limit-mirror.mjs` (single-invocation, matching that sibling — the script runs `selfCheck()` unconditionally at module top, so the missing `--self-check` step costs nothing). **I verified it red-first** by running the script from `6a6ffa4c` against a copied tree: green unperturbed (`server x10 contains browser ±30% (x1.30)`), and exit 1 with the full #2240 explanation for (a) browser `0.3 → 9.5`, (b) core `10 → 1.2`, and (c) a rename of the core constant. Its docblock also scopes itself honestly — it compares the two literals and says so, rather than implying it pins that either is still consulted.

Three things beyond what I raised are worth naming. `PriceOverrideBoundResult` became a discriminated union derived from `PriceOverrideBoundOutcome` (the `TerminaliseRoutingDecisionInput` precedent), which deleted the `limit as number` cast and makes `{outcome: 'ok', limit: 5}` unrepresentable. The 422 body now carries `outcome` / `attempted` / `computedAmount` / `limit` as **fields** beside the message — the `ReturnCustodyTransitionError.reason` rule, so a client is not parsing prose — and is documented on the route with `@ApiResponse`. And `price-change-override-bound.int-spec.ts` covers the `instanceof` arm the service spec could not, citing the §22 rival-`ReturnNotFoundError` precedent by name: it asserts 422-not-500, the four fields, that the episode is left claimable (so an operator is not locked out by their own typo), and that a large-but-plausible correction still passes. `loginAsAdmin` is called once per test, so the fixed-admin unique-constraint trap is avoided.

The unit spec's hardcoded `0.3` survives, correctly — it now says in its own comment that it pins one side only and points at the script for the other, and what it adds is the *behaviour* at the boundary rather than the containment claim. Re-confirmed: still a real server-side refusal in `PriceChangesService.edit`, still deliberately wider than the browser, still not promoting the warn to a gate. No migration of its own — the three migration files in this delta arrived via the merge from `3151-auto-applied-note-and-dialog`, not from this PR's own commits. Route authorization, repository-port and Symbol-token conventions unchanged.

### Findings

**[SUGGESTION]** — `scripts/check-price-override-bound-mirror.mjs:1`
> The script is wired without a separate `--self-check` invocation, which is right (it self-checks on every run) but reads as an omission next to the ~40 siblings spelled `X --self-check && X`. A one-line note in the `check:invariants` neighbourhood, or in the script's header, would stop someone "fixing" it later.

### Verdict
✅ **Approve**

CI: `Scaffolded Adapter Builds` pass; no other checks reported on this head yet — `mergeable: MERGEABLE`, `mergeStateStatus: BLOCKED` on the stack base (`3151-auto-applied-note-and-dialog`), which is expected at this position.

---

Issue #3237, opened 2026-09-14
[TECH-DEBT] Frontend — give DataTable per-row identity, then migrate the price-changes queue onto it


norbert-kulus-blockydevs:
## Problem / Context

Part of #3139 (epic). Split out of **#3223** after implementation found that issue's stated premise to be false.

#3223 claimed the responsive gap and the `DataTable` migration were one issue *"because the migration is what supplies the card view"*. They are not, because **the migration cannot happen as things stand**: `DataTable` carries no per-row identity, and `price-changes-queue-table.tsx` depends on six pieces of it.

`apps/web/src/shared/ui/data-table.tsx` hard-codes its row element (`:487-493`):

```ts
const rowClasses = [
  'data-table__row',
  expandable ? 'data-table__row--expandable' : href ? 'data-table__row--linked' : '',
  expanded ? 'data-table__row--expanded' : '',
].filter(Boolean).join(' ');
// ...
<tr key={key} className={rowClasses} onClick={onClick} style={style}>
```

Its props interface has 16 entries and none of them is a per-row `className` or attribute hook.

What the queue hangs on its own `<tr>` today, and would lose:

| On the row | What it carries |
|---|---|
| `is-group-start`, `is-grouped` | the left-border + continuation-row treatment for one source change fanning to several destinations |
| `is-resolved` | the muted treatment for a row inside its Undo window |
| `is-flagged` | the `needsRefresh` row |
| `data-state={rowState}` | the per-row handle the committed mockup and the #3152 E2E spec bind to by name |
| `id`, `data-row-id`, `data-testid` | the same, for row-level assertions |

A migration that dropped these would remove grouping and break the E2E handles — which #3223's own acceptance criteria forbid, so the migration is blocked on the primitive rather than merely unscheduled.

**Also folded in here** (from the #3164 approval, and it belongs with the queue's data layer rather than with its layout): `CHIP_COUNTS_LIMIT = 200` bounds the read the **per-connection** chip counts derive from, so past that ceiling a connection chip renders a confident number that is silently low. The `All` chip is unaffected — it reads the authoritative server-side `total` — and the under-count self-corrects the moment the chip is clicked, which is why it was a suggestion rather than a blocker. The real fix is a counts-by-connection endpoint, which the code comment already names.

## Proposed Solution

**1. Extend `DataTable` with per-row identity.** Two optional props, purely additive so no existing caller changes behaviour:

```ts
rowClassName?: (row: Row) => string | undefined;
[...]

PR #3238, opened 2026-09-14
feat(web): mobile card view for the price-changes queue

### LINKED ISSUE #3223: [TECH-DEBT] Frontend — price-changes queue table: mobile card view
## Problem / Context

Part of #3139 (epic). Raised by the round-3 review on #3164.

The Price changes review queue ships its own stylesheet block with **zero** `@media` queries: `overflow-x: auto` on the scroller and `min-width: 900px` on the table (`apps/web/src/index.css`, the `.price-changes-queue` block).

`docs/frontend-ui-style-guide.md § Responsive` is explicit that mobile (≤ 767 px) and tablet (768–1023 px) are **first-class** — *"operators should be able to triage failures from a phone off-hours and from an iPad on the shop floor"* — and its parity matrix specifies **card view** for tables on mobile: one card per row, key columns stacked. A 7-column queue behind a horizontal scrollbar is not that.

**Narrowed after implementation (was: "+ DataTable migration").** This issue originally bundled the migration onto the shared `DataTable`, on the stated premise that *"the migration is what supplies the card view"*. That premise is false: `DataTable` carries no per-row `className` or attribute hook, and the queue hangs its grouping treatment, its resolved/flagged states and its `data-state` / `id` / `data-testid` handles on the row element — so migrating as things stand would delete the grouping and the E2E handles, which this issue's own criteria forbid. The migration, the primitive extension it needs first, and the `CHIP_COUNTS_LIMIT` ceiling all moved to **#3237**.

## Proposed Solution

Give the queue a real mobile card layout below `767.98px`, on its existing markup.

- The breakpoint is `(max-width: 767.98px)`. The fractional bound is load-bearing and documented in § Responsive: a rounded `max-width: 768px` **matches at 768 px** and overlaps the tablet `min-width: 768px` band by exactly one pixel, so at that width the row is in its mobile branch while every table on the page is in its desktop one. `DataTable`'s own card/table switch reads `(max-width: 767.98px)`; match it rather than inventing a second boundary.
- Tablet (768–1023 px) keeps the full table scrolled horizontally within its container, per the parity matrix. Only mobile becomes cards.
- Pick the card's fields deliberately: product identity, the price delta and the row action are the operator's triage set. The rest can stay desktop-only.
- The `min-width: 900px` and the horizontal scroller must not apply in the mobile branch.

The card styling here is the queue's own and is expected to be **replaced** by `DataTableCardView` when #3237 lands — which is why #3237 depends on this and not the other way round: doing it in the other order would mean writing the card twice.

## Classification

**Type**: Frontend
**Layer**: Interface (feature stylesheet + component markup as needed)
**File(s)**:
- `apps/web/src/index.css` — the `.price-changes-queue` block
- `apps/web/src/features/price-changes/components/price-changes-queue-table.tsx`
- `docs/frontend-ui-style-guide.md § Responsive` — the contract
- `docs/plans/mockups/price-changes-review-queue.html` — § UX Mockups obligation

## Dependencies

- #3147 / PR #3164 must merge first — the component and its stylesheet block are introduced there.

## Assumptions

- Card view is required on mobile only; tablet keeps the scrolled table.
- The desktop rendering at ≥ 1024 px is unchanged.
- No API or query-hook change: this is layout.

## Acceptance Criteria

- [ ] At ≤ 767 px the queue renders one card per row with the chosen key fields stacked; no horizontal scrollbar and no `min-width: 900px` in that branch.
- [ ] At ≥ 1024 px the table is visually and behaviourally unchanged from today.
- [ ] The breakpoint is `767.98px`, matching `.data-table__cell--hide-below-768`, not a rounded `768px`.
- [ ] Row actions remain reachable and ≥ 44 px tall in the card layout.
- [ ] The branch is asserted off `matchMedia`, not off a measured rect (§ Responsive — a classic scrollbar makes `clientWidth` ~15 px narrower than the width media queries evaluate against, which inverts the answer at the boundary).
- [ ] The tracking comment left in `apps/web/src/index.css` by #3164 is updated to point at #3237 for the migration half rather than claiming this issue covers both.
- [ ] `docs/plans/mockups/price-changes-review-queue.html` is either updated for the card view or the PR states why it is still accurate (§ UX Mockups).

norbert-kulus-blockydevs:
## Summary

Part of #3139 (epic). **Stack position: eleventh** — branches from `3222-price-override-server-bound` (#3236).

The queue shipped its stylesheet with **zero** `@media` queries, `min-width: 900px` on the table and a horizontal scroller. `docs/frontend-ui-style-guide.md § Responsive` makes mobile first-class and its parity matrix specifies a **card view** for tables there; a 7-column grid behind a scrollbar is not that.

Below `767.98px` each row is now a card — header hidden, every cell carrying the label the header used to give it, selection leading beside the product identity, the action row a full-width footer, and the grouped treatment moved from the first cell's border (no longer a column) to the card's left edge. Tablet is untouched: the matrix keeps the full table there.

## Three decisions worth reviewing

**The bound is `767.98px`.** § Responsive calls the fraction load-bearing — a rounded `max-width: 768px` matches *at* 768px and overlaps the tablet `min-width: 768px` band by one pixel, so the queue would be in its mobile branch while every other table on the page was in its desktop one. It is the same bound `.data-table__cell--hide-below-768` uses, and the test asserts that spelling still exists so this block fails if the house breakpoint moves without it.

**The 44px target is re-stated inside the branch, not inherited.** The global rule fires under `(hover: none) and (pointer: coarse)` — a POINTER test — so a 767px desktop window, and every component test, get nothing from it. This branch is keyed on WIDTH.

**The test asserts the query, never a measured width.** § Responsive: a classic scrollbar makes `documentElement.clientWidth` ~15px narrower than what media queries evaluate against, which inverts the answer exactly at the boundary. There is also no CSS parser anywhere in this pipeline (#2674), so a test over the stylesheet text is the only gate these rules can have — the `card-button-reset-styles.test.ts` precedent.

## The issue was narrowed while implementing it

#3223 claimed the responsive gap and the `DataTable` migration were one issue *"because the migration is what supplies the card view"*. **That premise is false.** `DataTable` hard-codes its `<tr>` classes (`data-table.tsx:487-493`) and its 16 props include no per-row `className` or attribute hook, while this table hangs six things on the row: `is-group-start` / `is-grouped` (the grouping treatment), `is-resolved`, `is-flagged`, and `data-state` / `id` / `data-testid` (the handles the committed mockup and the #3152 E2E bind to). Migrating as things stand would delete the grouping and break those handles — which #3223's own acceptance criteria forbid.

So the migration is **blocked on the primitive**, not merely unscheduled. It, the `rowClassName` / `rowAttributes` extension it needs first, and the `CHIP_COUNTS_LIMIT = 200` ceiling folded in from the #3164 approval are now **#3237**. #3237 depends on this PR rather than the reverse: doing it the other way round would mean writing the card twice.

The `index.css` comment left by #3164 said one issue covered both halves; it now says why there are two, and points at #3237 with the blocker named.

## Mockup

`docs/plans/mockups/price-changes-review-queue.html` is touched per § UX Mockups. It is desktop-only and does not model the mobile branch — that is recorded in its own state-partition comment rather than restructured, since the screenshot baseline #3152 takes from it is the desktop one. The mockup is not stale about what it does show.

## Test plan

- [x] `apps/web` `price-changes-queue-card-view.test.ts` — 6 cases: the boundary spelling, the scroller and 900px floor removed, the header hidden with labels supplied, the 44px target, the grouped card edge, and that this is the queue's ONLY media block so tablet keeps the table.
- [x] `pnpm check:invariants` — clean (`check-css-structure`: 3727 blocks balanced; `check-design-tokens` clean).
- [x] `apps/web` `features/price-changes` — 75 pass.

Closes #3223


piotrswierzy (commented):
## Tech Lead review

This closes the substance of the #3164 finding — the queue really does card below 768px now, on the right bound, with the right reasoning about *why* the fraction matters and *why* the global 44px rule does not reach a 767px desktop window. I verified every custom property the new block references against `apps/web/src/index.css` (`--space-2`, `--space-3`, `--border-default`, `--radius-md`, `--bg-surface`, `--tracking-caps`, `--text-muted`, `--accent-primary-border` — all eight declared), and every selector against the real markup (`.table-wrap`, `.table-scroll`, `td.col-num`, and the four row classes are all live in `price-changes-queue-table.tsx`), so this is not another invisible-UI case. The narrowing of #3223 into #3237 is well argued and correct — `DataTable` does hard-code its `<tr>` classes with no per-row hook, so the migration genuinely is blocked on the primitive rather than merely unscheduled. My lead concern is the cascade: four desktop rules out-specify the mobile reset and keep painting inside the card, including the very treatment the new card edge is described as replacing. A stylesheet-text test cannot see that, by construction.

### Findings

**[IMPORTANT]** — `apps/web/src/index.css:2242` (the mobile `td` reset)
> `.price-changes-queue .table-wrap tbody td { border: 0; }` is specificity **(0,2,2)**. Four desktop rules that also target `td` are higher and therefore still apply inside the card — media queries add no specificity:
> - `:1978` `tr.is-group-start td { border-top: 1px solid var(--border-default) }` — **(0,3,3)**. Every `td` is a grid item here, so a group-start card draws a hairline above *each* stacked cell.
> - `:1988` `tr.is-grouped td:first-child { box-shadow: inset 3px 0 0 var(--accent-primary-border) }` — `box-shadow` is not a border and `border: 0` never touches it at any specificity. `td.cell-select` is still `:first-child`, so a grouped card renders the 3px accent beside the checkbox **in addition to** the new `border-left` on the card. The block's own comment says the desktop treatment borders "the first cell, which no longer exists as a column here" — it does still exist, as grid item 1/1.
> - `:1970` `tr.is-resolved td { background: var(--bg-surface-muted) }` and `:1992` `tr.is-flagged td { background: color-mix(...) }` — **(0,3,3)**, so the tint paints per-cell rectangles with the card's own `background: var(--bg-surface)` showing through the `gap: var(--space-2) var(--space-3)` gutters, instead of tinting the card. That lands on `is-flagged`, i.e. exactly the rows an operator most needs to read at a glance.
>
> Neutralise them explicitly inside the branch (`tr.is-group-start td { border-top: 0 }`, `tr.is-grouped td:first-child { box-shadow: none }`, and move the two tints from `td` to `tr`). `docs/frontend-architecture.md` § Design tokens records why the existing guards cannot catch this: `check-design-tokens.mjs` reads by regex and `check-css-structure.mjs` only parses balance — "class-presence assertions cannot catch that — the class still appears," and the same holds for a rule that is present and then loses the cascade.

**[IMPORTANT]** — `apps/web/src/features/price-changes/components/price-changes-queue-card-view.test.ts:57-62`
> The breakpoint assertion is the weaker of the two forms, and the style guide names the stronger one by file. `frontend-ui-style-guide.md` § Responsive: *"`.who-decides-row` shipped that way and was measured doing it, which is why `who-decides-styles.test.ts` now asserts the row's query **equals** `.data-table__cell--hide-below-768`'s rather than a hardcoded string — a component follows the primitive if the house breakpoint ever moves."* This test hardcodes `'@media (max-width: 767.98px)'` and separately asserts the *class name* `.data-table__cell--hide-below-768` merely exists somewhere in the file. If the house breakpoint moves, the class name still exists, the test still passes, and the queue silently keeps the old bound — which is the drift the assertion is written to prevent. `who-decides-styles.test.ts:186-215` already ships a brace-matched `queryFor(selector)`; reusing it and asserting `queueQuery === queryFor('.data-table__cell--hide-below-768')` makes the PR body's claim true. Its "guard of the guard" case (a selector in no media block must return `undefined`) is worth copying too.

**[IMPORTANT]** — `apps/web/src/index.css:2283`
> The width-keyed branch raises exactly one interactive element. § Responsive is unconditional: *"Tap targets ≥ 44 px on mobile for **every** interactive element."* The card contains the selection checkbox (`td.cell-select input`), the product link, two `.connection-tag` links and the action button; only `td.col-num .button` is covered. This is the PR's own argument applied to one of five — the global `@media (hover: none) and (pointer: coarse)` block at `:5636` sets `min-height: 44px` on `input` precisely because a checkbox is a real target, so the width-keyed branch is inconsistent with the pointer-keyed one it is compensating for. The checkbox is the card's primary bulk-selection affordance; add `td.cell-select input[type="checkbox"] { min-height: 44px; min-width: 44px; }`. (Links are uncovered by the global rule too, but that is pre-existing and out of scope.)

**[IMPORTANT]** — `docs/frontend-ui-style-guide.md` § Density & Row Heights (not touched by this PR)
> The mobile card is a new non-`DataTable` CSS-grid row with its own composition and height, and it is not registered. That section is emphatic: *"Never introduce a row height that isn't on this list without updating the guide first. Variability across surfaces is the primary way a cockpit feels amateur."* Three precedents sit in the table — `.who-decides-row` (#2354), `.fulfilment-worklist-row` (#2410) and `.sales-document-market-row` (#2540) — and the last is the direct model, since its entry documents its own below-768 behaviour in place: *"Collapses to a single column, auto height below 768 px."* The price-changes queue has no entry at all today; the desktop half is #3147's debt, but this PR is the one adding a carve-out-shaped row, so please add the row here in the same change.

**[SUGGESTION]** — PR description / test plan
> § Responsive closes with *"Every phase PR captures after-shots at three widths: 360 × 812, 768 × 1024, 1440 × 900."* This is the responsive PR for the surface and carries none, and there are no PR comments with them either. Three screenshots would also have surfaced the cascade finding above without a reviewer reading specificity by hand.

**[SUGGESTION]** — `apps/web/src/index.css:2294` / `price-changes-queue-card-view.test.ts:32`
> A stray double blank line before the `/* ── Price change dialogs (#3148` comment. And `queueMobileBlock()` keys on the literal `'/* ── Price changes queue — mobile card view (#3223)'`, an em-dash-and-box-drawing string — consistent with the file's house style, but a one-character edit to that comment silently breaks six assertions with a `toBeGreaterThan(-1)` failure that does not say why. A short note on the marker constant would help the next reader.

### Positive observations

The 767.98 reasoning, the "keyed on WIDTH not pointer" reasoning and the "assert the query, never a measured width" reasoning are all correct and are all recorded where the next reader will hit them, not only in the PR. Keeping tablet on the full table matches the parity matrix exactly. The `index.css` comment left by #3164 is properly retracted rather than quietly replaced — it says the earlier premise was false and names the new blocker. The mockup is touched per § UX Mockups with an honest statement of what it does and does not model, which satisfies that section's reviewer-enforced obligation. Everything is scoped under `.price-changes-queue`, so nothing leaks into other tables. No dependency-direction, state-ownership or `fetch()` concerns — the TSX change is six `className`/`data-label` attributes.

### CI

`mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`. **No checks reported at all** on `3223-queue-table-card-view` — the workflow triggers on `pull_request: branches: [main]` and this targets `3222-price-override-server-bound`, so lint, type-check and the web suite are absent, not passing. No migration, so the contended timestamps are not in play.

### Verdict
🔄 **Approve with changes**

The cascade finding should be fixed before merge — it affects the flagship deliverable on grouped and flagged rows. The breakpoint-test, tap-target and style-guide items are each small and worth folding into the same revision.

---

norbert-kulus-blockydevs:
## Review addressed — commit `b72533c64`

- **[IMPORTANT] The mobile `td` reset is out-specified by four desktop rules** — *fixed*. All four confirmed live: `tbody td { border: 0 }` is (0,2,2) and each `tr.is-<state> td` is (0,3,3). Each is now neutralised **by name** inside the branch (`is-group-start td { border-top: 0 }`, `is-grouped td:first-child { box-shadow: none }`, both tints `background: transparent` on `td`), with the resolved/flagged tints re-homed onto the `tr`. The flagged tint is composited as `color-mix(…, var(--bg-surface))` rather than carried over translucent — on an opaque card the desktop value would let the page ground through instead of layering over the surface. The grouped comment's false claim that the first cell "no longer exists as a column" is corrected to "still grid item 1/1", which is why the `box-shadow` neutraliser is needed at all.
- **[IMPORTANT] The breakpoint test cannot fail on what it pins** — *fixed*. The hardcoded `'@media (max-width: 767.98px)'` + class-exists pair is replaced by the `who-decides-styles.test.ts` form: a brace-matched `queryFor(selector)`, `queueQuery === queryFor('.data-table__cell--hide-below-768')`, and its guard-of-the-guard case. **Verified against a deliberate breakage** — the primitive's own media query moved to `899.98px` while the queue block stayed at `767.98px`: the new equality turns red (`767.98px vs 899.98px`) while **both former assertions stay green**. That is the #2673 shape exactly. The media-block count now strips CSS comments first, or naming the global pointer rule in a comment trips it.
- **[IMPORTANT] Tap target raised on one of five controls** — *fixed*. `td.cell-select input[type='checkbox'] { min-height: 44px; min-width: 44px }` inside the width-keyed branch, with the reasoning recorded in place: the global rule is keyed on `(hover: none) and (pointer: coarse)`, a **pointer** test, so a 767 px desktop window gets nothing from it while § Responsive is unconditional. The two per-card links stay uncovered and the comment says so — pre-existing, inherited from the global rule, out of scope per your own note.
- **[IMPORTANT] § Density & Row Heights not updated** — *fixed*. Added a table row plus a carve-out paragraph on the `.who-decides-row` / `.sales-document-market-row` model, placed last in the carve-out sequence to match its table position. **No height figure is quoted, deliberately** — nothing here could measure one, and the guide's culture punishes unmeasured numbers. The entry registers the shape, names the composition, states the two fixed 44 px control floors, records the pointer-vs-width distinction and the specificity rule, and notes that desktop/tablet have no entry yet (#3147's debt).
- **[SUGGESTION] Stray blank line + marker constant** — *fixed*. The double blank line before the dialogs header is gone (one blank line is the file's separator style, checked against every other `/* ── ` header), and the marker is now a `MOBILE_SECTION_MARKER` constant whose docblock states that it is a literal em-dash/box-drawing comment string and that a one-character edit to the CSS comment fails the whole suite on `toBeGreaterThan(-1)`.
- **[SUGGESTION] PR description / test plan** — `needs-owner-decision`. Cannot be done from here: no browser, no built app. The body should gain a **Screenshots** section with after-shots at 360×812, 768×1024 and 1440×900 per § Responsive, and the test plan should say the breakpoint assertion is now an equality against the primitive's own query rather than a hardcoded bound. Worth shooting a grouped row, a flagged row and a resolved row at 360 px specifically — those three are what the cascade fix changes.

Guards run in the worktree: `design-tokens: OK (111 tokens, all declared in index.css)`, `check-css-structure: 2 stylesheet(s) parse cleanly (3733 blocks balanced)`, `check-ui-vocabulary: 9 banned term(s) identical … 163 file(s) scanned across 6 of 6 feature folders` — all exit 0. The 20 test assertions were replayed against the real stylesheet by a dependency-free harness: 20 PASS / 0 FAIL.

---

piotrswierzy (approved):
## Tech Lead re-review (`f76d77b` → `b72533c6`)

All four IMPORTANT findings are closed. The desktop `td` bleed-through is neutralised **by name**, which is the only thing that works: the branch now carries `tr.is-group-start td { border-top: 0 }`, `tr.is-grouped td:first-child { box-shadow: none }` and `tr.is-resolved td, tr.is-flagged td { background: transparent }`, with the in-file comment stating the specificity arithmetic ((0,2,2) reset vs (0,3,3) state rules, media query adding none) and the separate fact that `border: 0` never touches a `box-shadow` at any specificity. Both tints are re-homed onto the card — `is-resolved` opaque `--bg-surface-muted`, `is-flagged` `color-mix(in oklab, var(--status-warning-soft) 55%, var(--bg-surface))` — with the compositing choice explained rather than the translucent desktop value carried over, which is the right call: on an opaque card the desktop tint would have let the page ground through. I brace-matched the block from `b72533c6`'s `index.css`: every new rule is inside the single `@media (max-width: 767.98px)`, nothing follows it before the dialogs section, so desktop is untouched.

The tap-target floor now covers the checkbox — `td.cell-select input[type='checkbox'] { min-height: 44px; min-width: 44px }` — and the comment names exactly why the global rule does not reach it (pointer test vs the width-keyed requirement) and is honest that the two per-card links remain uncovered as pre-existing debt rather than claiming completeness. The breakpoint test is now the query-equality form: it extracts the queue's own query and asserts it `toBe` the primitive's, resolved via a brace-matched `queryFor` rather than a `split('@media ')` sweep — plus a guard-of-the-guard asserting a nonexistent selector returns `undefined`, so an equality passing on two accidental hits is caught. The style-guide registration landed with a Density row **and** a four-point carve-out that is better than what I asked for: it declines to quote a height because none was measured, restates the width-vs-pointer keying, records the specificity trap so the next copier does not rediscover it, and names #3237 with the blocker (the primitive's hard-coded `<tr>` classes) rather than calling the migration merely unscheduled. The "only media block" test also now strips comments before counting, which it had to once the new comments mention `(hover: none) and (pointer: coarse)`.

Re-confirmed at the new head: all four custom properties (`--bg-surface-muted`, `--status-warning-soft`, `--bg-surface`, `--accent-primary-border`) are declared in `index.css` (light + dark), and every selector matches live markup in `price-changes-queue-table.tsx` — `td.cell-select` wraps the `input[type="checkbox"]`, `td.col-num` wraps `ActionCell`, and `is-group-start` / `is-grouped` / `is-resolved` / `is-flagged` are all applied to the `<tr>`. Still not an invisible-UI change. No migration of its own — the three migration files in this delta arrived via the merge from `3222-price-override-server-bound`. Dependency direction, state ownership and the no-raw-`fetch()` rule are untouched (CSS + test + doc only).

### Findings

> The body is now stale against the head: it still says the test file has "6 cases" (it has 7) and still describes the boundary assertion as "asserts that spelling still exists", which is precisely the weaker form this push replaced. Worth a refresh so the description does not argue for the version that was superseded.

**[SUGGESTION]** — three-width after-shots
> Still absent. The doc half of that finding is more than closed, and the CSS is now assertable from text, so this is no longer important enough to hold the PR — but a 375 / 768 / 1280 strip on a change whose whole subject is what the surface looks like at width would let the next reader confirm the flagged-row tint reads correctly against the card without rebuilding the branch.

### Verdict
✅ **Approve**

CI at this head: `Docker Build Smoke Test`, `PHP Unit Tests`, `Detect changed paths` pass; `Build`, `Lint`, `Test`, `Type Check`, `Integration Tests` **pending** — report absent, not green. `mergeable: MERGEABLE`, `mergeStateStatus: BLOCKED` on the stack base (`3222-price-override-server-bound`, #3236), which is expected at this position.

---

PR #3241, opened 2026-09-14
feat(sales-documents): inline amount and currency on a rule condition

### LINKED ISSUE #3189: feat(sales-documents): inline amount and currency on rule conditions

norbert-kulus-blockydevs:
Reverses ADR-041 decision 5's `thresholdRef` indirection. An `orderTotalGross` condition named a row in `sales_document_thresholds`, so the amount an operator reads and the amount a rule compares against were two different things joined by a slug the operator never chose and could not author. The condition now carries `amount` + `currency` directly.

## Backend

- `SalesDocumentCondition`'s `orderTotalGross` arm takes a **decimal string** amount and an ISO-4217 currency, with `isDecimalAmountString` / `isCurrencyCode` / `compareDecimalAmountStrings` (BigInt, never `parseFloat`) beside it. A string because conditions round-trip through `jsonb`, where a number loses precision and re-serialises differently - which would move `conditionsHash` and with it the unique index backing the duplicate guard.
- `evaluate-sales-document-rules` drops its `thresholds` input entirely; the comparison reads the condition. Currency is **compared, never converted**: an order in another currency does not match and falls to the next tier.
- Migration `1884000000000` rewrites every persisted rule - resolving each `thresholdRef` against the version effective on the migration date - and **recomputes `conditions_hash`** with a canonicalisation copied into the migration rather than imported, so it keeps meaning what it meant when it ran. An unresolvable ref throws and rolls back rather than dropping the condition; `down()` throws, because the original ref is not recoverable from an inlined amount.
- **Model and migration ship together on purpose**: `isSalesDocumentCondition` returns `false` on an unrecognised shape and every caller reads that as "this rule never matches", silently - so a narrowed type without the rewrite would have killed every threshold rule in the field with nothing in any log.
- `sales_document_thresholds` is retained with its read endpoint but is no longer read by the evaluation path. Dropping it is a separate decision.

## The dead starter rule (folded into this issue)

- Poland's `no-tax-id` template rule is **deleted**. It matched `buyerHasTaxId === false`, which is reachable only from a positively asserted "has none", and `readSourceBuyerTaxId` structurally cannot produce that state - so the flagship template shipped a rule labelled "Customer has no tax ID -> Receipt" that was dead on every install. The composer's own default condition had the same defect and now starts on the reachable state.
- Widening `buyerHasTaxId` with a `notPresent` operator spanning *unknown* and *asserted-none* is explicitly **not** done here: #2599 separated those two deliberately, and collapsing them is an ADR-063/ADR-073 amendment.
- PL's ceiling is 450 PLN **or** 100 EUR, and an inline amount is never converted, so the template's threshold pair becomes two pairs - one per currency.

## Frontend

- The composer takes an amount and a currency instead of a threshold select, with **no default currency**: guessing one would put a market's currency on a rule the operator did not write it for, and a mismatched currency silently never matches.
- `describeSalesDocumentRuleDraft` renders the whole draft as one sentence (`rule-readback`, the mockup's primary assertion target). It **never fills a gap in** - an unset currency reads `(no currency)`, an unset amount `?`, an unchosen connection `(no integration selected)` - because an operator who sees a plausible value believes they chose it. Framed as its own accent block beside the three step sections, not a fourth card, per the mockup.
- The order panel's "Why this kind?" disclosure names the real figure now that the projection carries it, instead of "the configured threshold".

## Also fixed here, both pre-existing on this stack

- **A real CJS cycle**, `invoicing -> orders -> fiscalization -> invoicing`: the fiscalization mapper value-imported `decodeBuyerTaxIdColumn` from the **main** orders barrel, which publishes `OrdersModule` and every orders service, one of which imports fiscalization straight back. `FISCAL_REGISTRATION_SERVICE_TOKEN` was therefore `undefined` when `SalesDocumentViewService` decorated, and 24 tests failed pointing at a provider that was present. Repointed to the `@openlinker/core/orders/types` cycle-breaker sub-barrel #3224 published for exactly this. Recorded in `docs/lessons.md`, including why reordering the spec's imports is not a fix.
- One stale whole-object expectation left by #3186, which added `ruleId` to a tier-1 route decision.

## Verification

Run against the branch before the main cascade: `pnpm lint` 0 errors, `pnpm type-check` 0 errors, core sales-documents 208/208, api sales-documents 26/26, web 5602/5603 (1 pre-existing skip).

`main` (#3127, #2883) was cascaded up the stack into 3187 -> 3224 -> 3188 -> this branch. One conflict, in `apps/api/src/orders/http/orders.controller.spec.ts`: both sides appended a `describe` block (#3180's buyer-tax-id DTO projection and main's `markPreRolloutEra`). Both kept; that spec passes 79/79.

Closes #3189
Refs #3173


piotrswierzy (commented):
## Tech Lead review

This PR reverses ADR-041 decision 5's `thresholdRef` indirection, moving an `orderTotalGross` condition's amount and currency inline, and it **argues the reversal rather than merely performing it** — the module docblock, the entity docblock, the migration header and the `architecture-overview.md` entry all state what the indirection bought (independent versioning of a legal amount), what it cost (the amount was unauthorable, so "450 PLN **or** 100 EUR" was inexpressible and a euro-priced PL order matched nothing), and that the price is paid knowingly. The load-bearing safety property is handled correctly: because `isSalesDocumentCondition` returning `false` reads as *"this rule never matches"* silently, the narrowed type and the data migration ship in one release, with an unresolvable ref throwing and rolling back rather than writing a shape that still never matches. Currency is compared and never converted; nothing on this path reaches for the ADR-040 FX stamp, and the evaluator's own docblock restates that prohibition. `sales-documents` gains no cross-context import, so the zero-outbound-edge leaf property is intact, and the incidental `invoicing → orders → fiscalization → invoicing` CJS cycle fix (repointing to the `@openlinker/core/orders/types` cycle-breaker) is correct and well recorded in `docs/lessons.md`. Every CSS custom property and class the new `.rule-composer-readback` block references exists in `apps/web/src/index.css` — verified individually, not assumed. No blocking issues. The most important concern is a reported-versus-enforced gap: the new architecture-overview text names a comparator the shipped code does not call.

### Findings

**[IMPORTANT]** — `libs/core/src/sales-documents/domain/types/sales-document-condition.types.ts:114`, `docs/architecture-overview.md`
> `compareDecimalAmountStrings` has **zero callers anywhere in the repository** (`git grep` at head `1a4b770` returns only its own definition). The evaluator compares with `Number(condition.amount)` instead — `evaluate-sales-document-rules.ts`, and it argues sensibly why (`order.totalGross` is already a JS number, so no comparison can be more exact than that operand). The problem is that the architecture-overview paragraph this PR *adds* asserts the opposite as a load-bearing property: *"comparison goes through a BigInt-based `compareDecimalAmountStrings` rather than `parseFloat`"*. That is false of the shipped code, and it is the failure class this repo documents repeatedly — reported ≠ enforced (the #2229 rule, restated across `architecture-overview.md`: "a number an operator reads that drifts from the one that acted is worse than no number"). Compounding it, the function's own docblock claims *"this comparison decides which fiscal document a real sale gets"*, which points the next reader at the wrong code path, and it is re-exported from the barrel (`index.ts:50` does `export *`) with no unit test — there is no `sales-document-condition.types.spec.ts`. Pick one: either call it from `evaluateCondition` (converting `order.totalGross` to a decimal string first, which is the only way the BigInt path buys anything), or delete it and correct the architecture-overview sentence to say the comparison is `Number()` over a value `isDecimalAmountString` has already made safe. Shipping it uncalled while documenting it as the decision path is the one option that should not survive.

**[IMPORTANT]** — `docs/architecture/adrs/041-sales-document-routing-policy.md:33`
> The ADR of record still states the reversed decision as current: *"The threshold is a **`thresholdRef`** - a named amount resolved from a versioned regime pack rather than an inline literal, so the legal matrix versions independently of the rules"*. This PR amends `architecture-overview.md` and four docblocks but leaves ADR-041 itself untouched, so the document a future reader consults first for *why* now contradicts the code. The repo's established practice for a reversal is an `## Amendment (#N)` section in the ADR — ADR-040 § Amendment (#2777), ADR-042 amendment (#2502), ADR-048 § Amendment (#2647), ADR-054's amendments — and `engineering-standards.md § ADRs` plus the ADR README's supersession protocol call for exactly that. Issue #3189's own Scope names *"ADR-041 amendment (see the ADR sub-issue)"*, but no such sub-issue exists under epic #3173 (#3175 is a different subject) and the PR body does not schedule one. This is the gap my #3154 review noted the price-propagation ADR handling correctly; please add the amendment section here, or open and reference the sub-issue before the integration branch merges.

**[IMPORTANT]** — `apps/api/src/migrations/1884000000000-inline-sales-document-rule-amounts.ts` (timestamp)
> `1884000000000` is **also claimed by open PR #3159** (`1884000000000-create-price-change-auto-applied-log.ts`). Within this stack the timestamp is correct — the branch occupies 1881→1884 strictly above `main`'s tail of `1880000000000` — but `scripts/check-migration-timestamps.mjs` diffs only against `origin/main`, so neither branch can see the other's claim. It fails closed (whichever merges second trips rule 3, "strictly greater than every migration already on `main`"), which is the safe direction, but only if lint runs after the rebase. `docs/migrations.md` § Timestamp uniqueness invariant records what a collision that *does* land costs: TypeORM 0.3.17 sorts by timestamp with no tie-breaker, so one `up()` body can silently never run while both class names appear in the `migrations` table (#374). Coordinate with #3159 and re-prefix whichever lands second. For context, the wider fleet also has `1881000000000` double-claimed by #3027 and #3201, and #3014 sitting at `1878000000000`, below `main`'s tail.

**[IMPORTANT]** — `apps/api/src/migrations/1884000000000-inline-sales-document-rule-amounts.ts:150-165`
> The migration recomputes `conditions_hash` for every rewritten rule but does not guard `UQ_sales_document_rules_country_hash_from` — the unique index on `(country, conditionsHash, effectiveFrom)` declared at `sales-document-rule.orm-entity.ts:28`. Two rules in the same country, with the same `effectiveFrom` and otherwise-identical conditions, that cited *different* `thresholdRef`s resolving to the same `amount` + `currency` at migration time now hash identically and collide on the `UPDATE`. The window is narrow, and TypeORM wraps migrations in a transaction so a failure rolls back cleanly rather than corrupting data — but the operator gets a raw Postgres `23505` naming an index, where the sibling unresolvable-ref path gets a meticulously actionable message telling them exactly which rules to repoint. Given how carefully everything else here fails loudly and legibly, pre-detect duplicate resulting `(country, hash, effectiveFrom)` tuples in the same pass that builds `updates` and throw the same shape of named error.

**[SUGGESTION]** — `apps/api/src/sales-documents/data/sales-document-template-catalogue.ts:105-120`
> The PLN below-threshold rule's `effectiveTo` changed from `'2026-12-31'` to `null`, unremarked. That window was the one mechanism making the legal amount's staleness visible: with the ref gone, the PL template now hardcodes 450.00 PLN with no end date, so when the ceiling moves the template keeps routing on a stale statutory figure indefinitely with nothing to flag it — which is precisely the independent-versioning property the reversal gives up. Normalising it against the sibling `gte` rule (which was already `null`) is defensible, but it is a second behavioural change riding inside the amount inlining and the docblock should say which reading was chosen and why.

**[SUGGESTION]** — orphans left behind
> Four items are now unreachable and worth removing or re-pointing in the same change, since they are exactly what a future reader will mistake for live contract: `SalesDocumentThresholdFact` (`sales-document-order-facts.types.ts:89`) has no consumer except a stale reference in `evaluate-sales-document-rules.ts:19`, whose module docblock still describes the input as carrying *"every `SalesDocumentThresholdFact` any loaded…"* after the field was deleted; `SalesDocumentThresholdNotFoundException` is still barrel-exported and `instanceof`-checked at `sales-document-rules.controller.ts:230` but is thrown by nothing now that `assertThresholdRefsResolve` is gone — a permanently dead 4xx branch; and `useSalesDocumentThresholdsQuery` (`apps/web/src/features/sales-documents/hooks/`) has no remaining consumer. Note also that the dead `no-tax-id` rule is removed from the *template* but persisted rows from prior adoptions keep it — correct (operator-owned rules should not be silently mutated) and consistent with the issue's acceptance wording, but worth one line in the PR body so nobody expects existing installs to self-heal.

**[SUGGESTION]** — `libs/core/src/sales-documents/domain/exceptions/sales-document-invalid-condition.exception.ts:7`
> Broken sentence from the edit: *"an `orderTotalGross` condition with no an inline `amount` + `currency` (#3189)"*. Should read "…with no inline `amount` + `currency`".

**[SUGGESTION]** — `apps/web/src/features/sales-documents/lib/describe-sales-document-rule-draft.ts` and `apps/web/src/index.css:3845`
> Two small frontend inconsistencies. The readback returns raw English while its direct sibling `describe-matched-sales-document-rule.ts` takes a `TranslationFn` and routes every string through `t(key, fallback)` — `docs/frontend-architecture.md § Internationalization` makes the seam the migration target, and two functions written in the same PR taking opposite approaches is drift being created rather than inherited. Separately, `.rule-composer-condition-row__threshold` is a `display:flex` with no `flex-wrap` sitting in a `1fr` grid column, and it went from two controls to three (op Select + amount Input + currency Input); the row grid stays `1fr 1fr auto` at every width with no media query, so the three controls compress rather than stack on a phone. Per the project's default that mobile and tablet are in scope for FE work, worth a `flex-wrap: wrap` or a narrow-width stack.

### Positive observations

- The model-and-migration-in-one-release argument, and the reason it cannot be split, is stated where it will actually be read — in the migration's own header — rather than only in the PR body.
- Copying the canonicalisation into the migration instead of importing it is right, and the docblock gives the correct reason (a migration must keep meaning what it meant when it ran).
- `down()` throwing rather than silently half-reversing is the correct treatment for a genuinely irreversible transformation, and it explains itself in the thrown message; this is an argued deviation from `docs/migrations.md`'s "always implement both", not an omission.
- The DTO narrows through the core guards (`isDecimalAmountString` / `isCurrencyCode`) rather than restating the rules, so "authorable over HTTP" and "evaluable by the engine" cannot diverge.
- The composer's refusal to default a currency, and the readback's refusal to fill any gap in, are both the right call and both carry the reasoning inline.
- `newConditionDraft` flipping `boolValue` to `true` catches the same dead-condition defect one surface over from the template — a genuinely good catch.

### Notes on CI

`mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`. The base is `3188-invoice-list-buyer-tax-id`, and `.github/workflows/ci.yml` triggers on `pull_request: branches: [main]` only — so the full lint / type-check / unit / integration suite **did not run on this PR**. Only "Scaffolded Adapter Builds" reported (pass). Treat the PR body's local run figures as the only evidence available for this member; the integration branch's eventual PR to `main` is where the gate actually applies.

### Verdict

🔄 **Approve with changes**

No blocking issues. Before the integration branch merges to `main`, please resolve the `compareDecimalAmountStrings` documentation/code divergence, add (or schedule and reference) the ADR-041 amendment, and coordinate the `1884000000000` timestamp with #3159.

---

norbert-kulus-blockydevs:
All four IMPORTANTs addressed (`2bf112964`, plus the cascade merge `2c5de0bdb`).

**[IMPORTANT] `compareDecimalAmountStrings` — reported ≠ enforced.** You are right that the paragraph was false. I did not take either option verbatim, and the reason is that a third fact was not visible from this PR alone: **#3242 (already open, one link up) calls it five times** — the overlap detector intersects two rules' bounds with it. Deleting it here would have broken that; calling it from `evaluateCondition` would buy nothing, for exactly the reason you gave.

So I corrected the two false claims instead and named the real consumer:
- the overview now says routing compares with `Number()` over a value `isDecimalAmountString` has already made safe, and that the BigInt function exists for comparison between two **authored** amounts;
- the function's own docblock no longer says it "decides which fiscal document a real sale gets" — it says what it is actually for, and names #3190 as the consumer;
- `sales-document-condition.types.spec.ts` is new: 29 cases over the comparison, both coercers and the narrowing, including a legacy `thresholdRef`-shaped condition. There was no spec at all, as you noted.

If you would still rather see it deleted here and re-added with its consumer in #3242, say so and I will move it.

**[IMPORTANT] ADR-041.** Partly a different gap than the finding describes: the `## Amendment (#3189, 2026-09-11)` section **already existed** — it landed with #3185, four links down, so it is not in this PR's diff. What was missing is a pointer at decision 5 itself, which is where a reader looking for the rule arrives first. A supersession blockquote now sits directly under that heading, stating what is retired and what still stands, and linking to the amendment.

**[IMPORTANT] migration timestamp.** Resolved without moving this one: #3159 has since taken `1887000000000` (plus a new `1888000000000`), and #3201 is moving to `1880500000000`, so `1884000000000` is now unclaimed by anything else. Scanned every open PR's migration set against `origin/main` to confirm rather than taking it on trust.

**[IMPORTANT] the unique index.** Fixed. The pass that builds `updates` now groups by `(country, recomputed hash, effectiveFrom)` and throws a named error listing the colliding rule ids per group, in the same shape as the unresolvable-ref arm beside it. The `SELECT` is widened to carry `country` and `effective_from` for that check and nothing else. You are right that the transaction made this a legibility problem rather than a data-safety one — which is the point, given everything else here fails loudly.

The four **[SUGGESTION]**s are not taken in this round.

Verification: `pnpm lint` and `pnpm type-check` clean, core sales-documents 199/199, api sales-documents 26/26, web 5619/5620.

piotrswierzy (approved):
## Tech Lead re-review (`1a4b7701` → `e05cdd15`)

All four IMPORTANT findings are closed, and three of them are closed in the way that leaves the next reader better off rather than merely satisfying the review.

**1 — `compareDecimalAmountStrings` pointing at the wrong path: closed, and closed honestly.** The function still has no caller *on this branch* — and the fix was not to manufacture one. The `architecture-overview.md` paragraph now reads *"Routing compares with `Number()`* over a value `isDecimalAmountString` has already made safe … `compareDecimalAmountStrings` (BigInt, never `parseFloat`) exists for comparison between two AUTHORED amounts, where both sides are decimal strings and the exactness is real; its consumer is the #3190 overlap detector"*, which is now a true statement about the code and names the pending consumer instead of asserting a code path that does not exist. The docblock at `sales-document-condition.types.ts:126` goes further and says so explicitly — *"It does NOT decide which document a sale gets (#3241 review, correcting an earlier version of this comment that said so)"* — which is the right instinct: a comment that was wrong once should say it was, or the next reader re-derives the same mistake. And it is no longer untested: `sales-document-condition.types.spec.ts` adds ordering, trailing-zero equality (`450` vs `450.00`), fractional-width (`450.5` vs `450.45`, `0.1` vs `0.09` — the two cases `parseFloat` would get wrong) and reflexivity, with the file header naming the shipped-with-no-test-and-no-caller state as the reason it exists. I have verified the consumer is real: PR #3242's `detectSalesDocumentRuleOverlap` calls it in both `reduceToBounds` and `pickTighter`.

**2 — ADR-041 not amended: closed, and to the repo's own pattern.** Decision 5 now carries an inline *"Superseded in part (#3189)"* note at the point of the claim, there is a full `## Amendment (#3189, 2026-09-11)` section at the end alongside the #2599 / #2822 / #2504 amendments, and the trail records PR #3200 and issue #3189. The amendment's own paragraph on status is the part I would have missed: it states that the amendment is settled rather than provisional *because* decision 5's rule engine is shipped (#2170/#2173) — so this edits a live model rather than revising a proposal — while explicitly declining to move ADR-041 off `Proposed`, since decision 1's router and decision 8's aggregation are still unbuilt. That is the distinction the status field is actually about, and stating it forestalls the obvious next question.

**3 — migration hash collision surfacing as a raw 23505: closed.** `1884000000000` now pre-computes `country|hash|effectiveFrom` across the pending updates, groups them, and throws naming **both sides** of every collided group plus the remedy (*"They cite different refs today but the same figure. Delete or narrow one rule per group, then re-run."*) before the first `UPDATE` runs — the same standard the unresolvable-ref arm already set, which is the right bar to have matched.

**4 — timestamp collision with #3159: gone.** Verified at head across the open set: #3159 now owns `1887000000000` and `1888000000000`; `1884000000000` is this PR's alone, and `1881000000000` / `1881000000001` / `1882000000000` / `1883000000000` are also uniquely held here. No collision against #3014 (`1878000000000`), #3027/#3201 (`1881000000000` — **worth a second look before merge**, see below), #3219 (`1882000000000`), #3227 (`1883000000000`) or #3035 (`1885000000000`) *within this branch's own set*, but three of those numbers are claimed by other open PRs per your table, so the ordering guard — which diffs against `origin/main` only — will not see them until one side merges.

### Findings
**[IMPORTANT]** — `apps/api/src/migrations/1881000000000-…`, `1882000000000-…`, `1883000000000-…`
> Three of this PR's five timestamps are claimed by other open PRs in your own table: `1881000000000` by #3027 and #3201, `1882000000000` by #3219, `1883000000000` by #3227. These are branch-local ancestors of this PR rather than its own slice, so they may already be resolved upstream — but `scripts/check-migration-timestamps.mjs` diffs against `origin/main` and cannot see a sibling branch, so nothing will catch it if they are not. Re-enumerate across the open set immediately before merging the stack; a collision here is TypeORM 0.3.17 sorting by timestamp with no tie-breaker, i.e. one `up()` body silently unapplied with both class rows present (#374).

**[SUGGESTION]** — `apps/api/src/migrations/1884000000000-inline-sales-document-rule-amounts.ts:168`
> The collision pre-check groups only the **pending updates** against each other. A rule that already carries an inline `amount` + `currency` (or otherwise needs no rewrite) is never in `updates`, so if a newly-inlined rule hashes identically to one of those, the `UPDATE` still hits `UQ_sales_document_rules_country_hash_from` and the operator gets the raw 23505 the guard exists to replace. Much narrower than what I raised — the transaction still rolls back cleanly, and it needs an operator to have hand-authored an inline rule matching what a `thresholdRef` resolves to — but the guard reads as total and is not. Including the country's existing rows in the key map would close it.

### Verdict
✅ **Approve**

Re-verified and not re-raised: currency is compared and never converted; ADR-040's FX stamp is untouched; the `sales-documents` leaf still adds zero cross-context imports and spends no allow-set entry; `.rule-composer-readback` resolves in `index.css`; the templates controller keeps `@Roles('admin')`.

CI is mid-run — Docker Build Smoke, PHP Unit, Scaffolded Adapter Builds and Detect changed paths pass; Lint, Type Check, Build, Test and Integration Tests are `pending` and are not reported here as green. `mergeable: MERGEABLE`, `mergeStateStatus: BLOCKED` on those. Base is `3188-invoice-list-buyer-tax-id`, and #3242 stacks on this branch in turn.

---

PR #3242, opened 2026-09-14
feat(sales-documents): refuse a colliding rule before the save

### LINKED ISSUE #3190: feat(sales-documents): rule overlap detection before save

---

Part of #3173.

## Context
The engine has no priorities: if two rules match an order, **nothing is issued and the order is held**. Today that is discovered days later as a hanging order. It belongs before the save, where there is still somebody to tell.

## Scope
- A pure overlap-detection function over the closed condition vocabulary
- Warning naming the rival rule, with save gated while it stands, and an "Open that rule" affordance
- The provable-non-overlap case stated positively (two rules in different currencies **cannot** both match)
- End-to-end: 450 PLN → 449 PLN (conflict, save inert) → change to EUR (no conflict, save available)

Where the check cannot decide, it must **say so** rather than stay silent.

## Acceptance
- [ ] `rule-conflict` and `rule-no-conflict` are mutually exclusive
- [ ] `rule-save` is inert while a conflict stands

**Depends on**: inline amounts — an overlap check built on seeded thresholds would be rewritten. **Spec**: `docs/plans/mockups/` (committed by #3174).

---

## Reopened: the implementation is not on `main`

The work was reviewed and merged, but it never reached `main`.

`3190-rule-overlap-detection` was merged into the stack base
`3189-inline-rule-amounts` by PR #3242. PR #3241 then merged that base into `main` as a
**squash** (`e3a8a661a`, single parent). A squash carries only the diff of the branch
being merged, and this branch sat *above* that base, so its commits were not included.

Verified rather than inferred: neither the #3242 merge commit
(`6049d62e6d760a601e2ab3f69e3fdf92f6250e4b`) nor `origin/3190-rule-overlap-detection` is
an ancestor of `origin/main`, and
`libs/core/src/sales-documents/domain/domain-services/detect-sales-document-rule-overlap.ts`
is absent from `origin/main`. Everything lower in the stack (#3186, #3187, #3188, #3189)
survived inside the squash; only this issue's work was lost.

### Scope of the re-land

6 new files, 13 modified. New:

- `libs/core/src/sales-documents/domain/domain-services/detect-sales-document-rule-overlap.ts` (+ spec)
- `apps/api/src/sales-documents/http/dto/check-sales-document-rule-overlap.dto.ts`
- `apps/web/src/features/sales-documents/hooks/use-sales-document-rule-overlap-query.ts`
- `apps/web/src/features/sales-documents/lib/describe-sales-document-overlap.ts` (+ test)

Modified: `sales-document-rules.controller.ts`, `ISalesDocumentRulesService`,
`SalesDocumentRulesService`, `sales-document-rule-composer-dialog.tsx`,
`sales-document-rules-list.tsx`, the `sales-documents` barrel, `index.css`,
`test-utils.tsx`, `docs/architecture-overview.md`.

The branch also carries review fixes committed above the feature commit - a fiscalization
barrel load-cycle fix, a KSeF customer-handle fix for an untagged buyer tax id, and a
migration renumber. Those are equally absent from `main` and are part of the re-land.

### Why this is done before #3191

#3191's dry-run endpoint is the same shape as `POST /sales-documents/rules/overlap-check`
- a POST carrying an unsaved draft, answering 200, persisting nothing - and both touch
`sales-document-rule-composer-dialog.tsx`. Building #3191 against a `main` without this
work means writing that precedent twice and resolving a conflict on the re-land.

norbert-kulus-blockydevs:
The engine has no priorities: when two rules match one order it resolves `unresolved` / `conflicting-rules-equal-priority` and the order is **held** (ADR-041 - never silently pick one, because for a fiscal document a wrong pick is a legal event). That is correct and it is also invisible: the operator meets it days later as a hanging order. This asks the question at the one moment somebody can still act on it.

## Core

- `detectSalesDocumentRuleOverlap` - pure, beside `evaluateSalesDocumentRules` - reduces each rule's AND-only conjunction to a bound per field and intersects them. Several `gte`s narrow to the largest and several `lt`s to the smallest, so a last-one-wins reduction would wrongly report a collision; `gte` is inclusive and `lt` exclusive, which is what makes the shipped PL template's own `< 450` / `>= 450` pair provably disjoint rather than a self-collision.
- **Three outcomes, never two**: provably overlapping, provably disjoint (with the reason), or **undecided**. The third exists because silence reads as "no conflict", which is the false reassurance the check removes. An unreadable persisted condition and a rule bounding the total in two currencies both land there rather than in either neighbour.
- **Different currencies are a PROOF of non-overlap, not an unknown** - an inline amount is compared and never converted (#3189), so the positive case is stated rather than left as an absent warning.
- **Decidable only because the vocabulary is closed and the composer is AND-only.** An OR tree makes exactly the interesting cases undecidable, so `SalesDocumentCondition` must not grow disjunction without revisiting this function - recorded on the function itself.
- `salesDocumentRuleWindowsOverlap` is exported and the write-path conflict guard now calls it, so the two places asking "are these live together" cannot drift.

## API

`POST /sales-documents/rules/overlap-check` - a **read** despite the verb: it carries a draft body, persists nothing and refuses nothing. The write path is deliberately **not** gated on it, so an operator may still knowingly save an overlapping pair through the API and the runtime keeps failing safe.

## Frontend

- The composer renders `rule-conflict` (save inert, testid swaps to `rule-save-blocked`), `rule-no-conflict`, or the undecided warning. Mutually exclusive by construction: a proven collision outranks a proven non-collision.
- "Open that rule" highlights and scrolls to the rival in the list behind the dialog, and renders **only** when the parent supplies the callback - a control the surrounding page cannot serve is dead code that type-checks.
- The check is asked only once the draft is answerable: a half-typed amount, or an amount with no currency, describes no order set at all.

## Acceptance

- [x] `rule-conflict` and `rule-no-conflict` are mutually exclusive - asserted in both directions in the composer spec
- [x] `rule-save` is inert while a conflict stands - the button is `disabled` and carries `rule-save-blocked`, with `rule-save` absent

## Verification

`pnpm lint` 0 errors, `pnpm type-check` 0 errors, core sales-documents 185/185 (13 new for the detector, 2 for the service seam), api sales-documents 26/26, web 5611/5612 (1 pre-existing skip; 6 new copy tests, 3 new composer tests).

Closes #3190
Refs #3173


norbert-kulus-blockydevs:
No review findings on this one yet — a note on what it carries and what it deliberately does not.

**The three outcomes are the design, not a tri-state for its own sake.** A pair provably CAN both match, provably CANNOT, or this build could not decide. The third is reported rather than folded into either neighbour, because silence reads as *no conflict* and that is exactly the false reassurance the check exists to remove. `unreadable-condition` and `multi-currency-rule` are its two reachable causes today.

**It is decidable only because of two properties that must not be relaxed.** The condition vocabulary is closed, and the composer is AND-only — so a rule is a conjunction of per-field bounds, and two rules intersect iff every field's bounds intersect. An OR tree would make precisely the interesting cases undecidable; `SalesDocumentCondition` carries no disjunction and must not grow one without revisiting this function.

**Bounds are reduced as a conjunction, and that is the bug a naive reduction produces.** Several `gte`s narrow to the largest and several `lt`s to the smallest; a last-one-wins reduction would report a collision for a rule that is internally narrower than it looks. Inclusive lower / exclusive upper is also what makes the shipped PL template's own `< 450` / `>= 450` pair provably disjoint rather than a self-collision — worth knowing, since that pair is what an operator meets first.

**Different currencies are a PROOF of non-overlap, not an unknown.** #3189 compares an inline amount and never converts, so two rules priced in different currencies cannot describe one order. Stated positively, because the tempting reading is "we could not compare, so warn".

**It is a READ.** `POST /sales-documents/rules/overlap-check` persists nothing and refuses nothing; an operator may still knowingly save an overlapping pair through the API, and the runtime keeps failing safe on `conflicting-rules-equal-priority`.

The cascade also brought in the corrected #3189 paragraph in `docs/architecture-overview.md` (the stale claim that routing used the BigInt comparator is gone — `grep -c` returns 0).

Verification: `pnpm lint` and `pnpm type-check` clean, core sales-documents 199/199 including this detector's own suite.

piotrswierzy (approved):
## Tech Lead review

Moves the no-priorities collision from a hanging order days later to the composer, at the one moment somebody can act on it. The core of it is a pure domain service beside `evaluateSalesDocumentRules` that reduces each AND-only conjunction to a bound per field and intersects them, and the design choices are the right ones: **three outcomes rather than two** (`undecided` is reported instead of folded into "no conflict", which is exactly the false reassurance the check exists to remove), a conjunction reduction that narrows several `gte`s to the largest and several `lt`s to the smallest rather than last-one-wins, and `gte`-inclusive / `lt`-exclusive arithmetic that makes the shipped PL template's own `< 450` / `>= 450` pair provably disjoint instead of a self-collision. `salesDocumentRuleWindowsOverlap` is extracted and the write-path guard now calls it, so the two places asking "are these live together" cannot drift — that is the single-source discipline, not a tidy-up. Verified rather than taken from the description: no non-relative `@openlinker/core/<ctx>` import is added anywhere under `libs/core/src/sales-documents/`, so the zero-outbound-edge leaf property and `barrel-purity.spec.ts` are untouched and no allow-set entry is spent; every class the composer and the list render resolves in `apps/web/src/index.css`, including the one new `.rule-card--highlighted` and its `var(--status-warning)` token (declared in both the light and dark blocks); the controller inherits class-level `@Roles('admin')`, which `route-authorization-coverage.spec.ts` discovers; and there is no schema change, so the `synchronize`-emits-neither-enums-nor-CHECKs trap does not apply. `isActive` is a non-question here — the flag `architecture-overview.md` § 17 mentions belongs to `automation_rules`; `sales_document_rules` has no such column, so `findByCountry` returning every row for the country cannot produce a false positive.

Two things worth stating for the record. The title reads as if this addressed my #3241 finding 3 (the migration recomputing `conditions_hash` without guarding `UQ_sales_document_rules_country_hash_from`) — **it does not, and does not need to**: that was fixed inside #3241 itself, and this PR is the composer-time overlap check, a different mechanism at a different moment. And the write path is deliberately **not** gated on this endpoint, which is right: the runtime keeps failing safe, an operator may still knowingly save an overlapping pair through the API, and a UI check promoted into a server-side refusal would be a mirror stricter than the gate.

### Findings
**[SUGGESTION]** — `apps/web/src/features/sales-documents/hooks/use-sales-document-rule-overlap-query.ts:26`
> The query key is the serialised draft and the gate is `open && conditionsAreAnswerable`, so once the amount and currency are both non-empty every further keystroke in the amount field mints a fresh key and fires a `POST`. Typing `450` after a currency is three requests. The endpoint is admin-only and does one indexed `findByCountry`, so this is nowhere near the expensive-`COUNT` case § Paginated Totals As A Second Stage is written about — but the reasoning transfers in miniature, and the hook already owns the right seam for it. A short debounce on the serialised key (not on the fetch) would settle it; the first answerable draft should not be debounced, for the same reason the first mount there is not.

**[SUGGESTION]** — `apps/api/src/sales-documents/http/dto/check-sales-document-rule-overlap.dto.ts:58`
> `excludeRuleId` is modelled end to end and has no frontend consumer: the composer is create-only and passes no `excludeRuleId`, so the self-collision case the field exists for is unreachable from the product. The domain covers it and the spec pins it (`skips the row being edited`), which is the correct order to build in — worth a line saying the edit path is the intended consumer, so the next reader does not read an unused field as dead.

### Verdict
✅ **Approve**

CI is mid-run — Docker Build Smoke, PHP Unit, Scaffolded Adapter Builds and Detect changed paths pass; Lint, Type Check, Build, Test and Integration Tests are still `pending`, so none of them is reported here as green. `mergeable: MERGEABLE`, `mergeStateStatus: BLOCKED` on those pending checks. Note the stack: this branches off `3189-inline-rule-amounts` (#3241), which itself branches off `3188-invoice-list-buyer-tax-id` — it cannot land before its parents.

---

norbert-kulus-blockydevs:
## This PR is marked merged in error - nothing of it reached `main`

GitHub flipped this to `MERGED` at 14:18 UTC because a bad push briefly made this branch's head an ancestor of its base. It was not a real merge, and it was my mistake, not a decision.

What happened: a cascade script I ran had two `git checkout` calls fail (the branches were held by other worktrees), and the loop carried on regardless - so a `git reset --hard` and a merge landed on the wrong branch and were pushed. GitHub saw this PR's head become reachable from its base and recorded the push commit as a merge commit.

Both branches were restored within minutes via `--force-with-lease` to the exact commits they held beforehand, confirmed against the remote-tracking reflog. **No work was lost**: this branch still carries its full content, and nothing from it is in `main` or in its base.

The residue is only the PR status. GitHub refuses to reopen a pull request it has recorded as merged:

```
422 Validation Failed - state cannot be changed. The pull request cannot be reopened.
```

So this PR cannot be revived, and a replacement one is needed for the same branch. The branch itself is current: it has been cascaded with `main` and every branch below it, and the stack top passes `@openlinker/core` at 426 suites / 6554 tests with `pnpm check:invariants` at exit 0.

Leaving this note so the merged badge is not read as "this shipped".

PR #2988, opened 2026-09-08
chore(main): release 0.11.0

tomektom:
:robot: I have created a release *beep* *boop*
---

### Features

* **api:** connection pricing & sync settings endpoints ([#3163](https://github.com/openlinker-project/openlinker/issues/3163)) ([b700499](https://github.com/openlinker-project/openlinker/commit/b7004996ff569dd15cff5b45596de8979b1ca3c1))
* **api:** price-changes queue endpoints (list, accept, ignore, edit) ([#3162](https://github.com/openlinker-project/openlinker/issues/3162)) ([6d0d7f3](https://github.com/openlinker-project/openlinker/commit/6d0d7f313b7f5f3fe8510b97961b341024d773d4))
* **dx:** enforce unique ADR numbers via check:invariants ([#3036](https://github.com/openlinker-project/openlinker/issues/3036)) ([93432c8](https://github.com/openlinker-project/openlinker/commit/93432c80641cc0d9bc131016e17c9e8c7e6844b2))
* **fiscalization:** send the buyer tax number on a fiscal receipt ([#3220](https://github.com/openlinker-project/openlinker/issues/3220)) ([05acf27](https://github.com/openlinker-project/openlinker/commit/05acf279b253173e750f50b7971c24f284990ede))
* **fulfillment:** reap dispatches whose holder never answered ([#2712](https://github.com/openlinker-project/openlinker/issues/2712)) ([#3012](https://github.com/openlinker-project/openlinker/issues/3012)) ([6e0f8f0](https://github.com/openlinker-project/openlinker/commit/6e0f8f0c693e657204f80cc5174d35a9924b3aa9))
* **fulfillment:** reconcile a shipped work whose dispatch relay never landed ([#2728](https://github.com/openlinker-project/openlinker/issues/2728)) ([#3018](https://github.com/openlinker-project/openlinker/issues/3018)) ([42cc86d](https://github.com/openlinker-project/openlinker/commit/42cc86dded9e75d0edec4a1340cba11356c805c8))
* **infakt:** add per-connection defaultSaleType for non-PL invoice issuance, with a diagnosis hint for the missing-config case ([#2995](https://github.com/openlinker-project/openlinker/issues/2995)) ([a5a6ef4](https://github.com/openlinker-project/openlinker/commit/a5a6ef4047f4a98f2835a92dad221283f4ec3f89))
[...]

Issue #2344, opened 2026-08-22
[TASK] CORE — ReservationService: get-or-create resume, delta-adjust, stamped atpEffect


piotrswierzy:
> Part of the OMS programme backlog (**Wave 2**, stream S1, size M). Backlog slug: `W2-7`. Sibling `W2-*` slugs resolve via the Wave-2 epic's checklist. Design of record: design §4.2 amendments (1) and (2).

## Problem / Context

Two of §4.2's three R1 correctness amendments live here. **(2) Reserve is get-or-create, never reject-on-retry** — without it, an ingestion crash after `claimHeld` wedges the order forever behind a false "insufficient stock". **(1) `atpEffect` is a stamped column, not a cross-context read** — so the ATP query is a local column test and no `inventory ↔ fulfillment` edge is created.

## Proposed Solution

`ReservationService implements IReservationService`:

- `reserve(input)` → `INSERT … ON CONFLICT DO NOTHING` on the partial-unique key, then re-select. An existing `held` row for the same `(order, line, position)` **is a success** — the insert-then-recover idiom `IdentifierMappingService` already ships. A row with a *different* quantity (the source amended the line) is an explicit **delta-adjust** through the same guarded `UPDATE`, never a release-then-reserve.
- `atpEffect` is supplied **by the ingestion caller**, which holds the routing outcome in hand, and is **immutable per reservation**. On the default topology (`omp_fulfilled`, marketplace ships) the caller stamps `diagnostic`; on OL-executed paths it stamps `published`. A missing value is a programming error, not a default.
- **The v2 multi-position gate**: a reserve **rejects loudly** (`AmbiguousReservationPositionError`) when a variant resolves to more than one non-stale position and no explicit position was supplied. Silently picking one is the failure mode this gate exists to prevent.

Wired into `OrderIngestionService` after the line-diff (#2283) and before destination provisioning.

## Classification

**Type**: CORE
**Layer**: Application
**File(s)**: new `libs/core/src/inventory/application/services/reservation.service.ts` + interface; `libs/core/src/orders/application/services/order-ingestion.service.ts`

## Dependencies

- #2343 (`W2-6`); #2283 (line diff); #2319 (`W1b-9` — duplicate-position detection, the gate's counterpart report)

## Assumptions

- Erli/Allegro/WooCommerce ingestion all reach one ingestion service, so one call site covers every source.

## Acceptance Criteria

[...]

Issue #2343, opened 2026-08-22
[TASK] CORE — reservations table, §6I guarded-UPDATE repository and named domain errors


piotrswierzy:
> Part of the OMS programme backlog (**Wave 2**, stream S1, size L). Backlog slug: `W2-6`. Sibling `W2-*` slugs resolve via the Wave-2 epic's checklist. Design of record: design §4.2 / ADR-061 / ANALYSIS-1032 §6I; REVIEW §3 H9.

> **Size justification (L):** one table, one `olReservedQuantity` column on the position row, and four repository operations — but every one is a concurrency primitive whose failure mode is an oversell, so the issue carries a full concurrent integration-test matrix (parallel reserve of the same line, parallel reserve across lines in reverse order, reserve racing a master sync that lowers `availableQuantity`, release racing expiry). That test matrix, not the DDL, is the size.

## Problem / Context

Design §4.2 / ADR-061 adopt ANALYSIS-1032 §6I **verbatim** as the mechanism. REVIEW §3 H9 requires the repository seam to be named up front (`ReservationRepositoryPort.claimHeld` / `releaseHeld`, each with its own named domain error) rather than letting a service improvise raw query-builder calls.

## Proposed Solution

New `reservations` table in `libs/core/src/inventory/`: `id`, `orderRecordId`, `orderLineId`, `inventoryItemId`, `quantity`, `status` (`held | released | consumed | expired`), `expiresAt` (**NOT NULL** — an unbounded hold on a system that may never observe the close event is an oversell leak with no floor), `atpEffect` (`published | diagnostic`, NOT NULL, see `W2-7`), timestamps.

Mechanics, all from §6I and non-negotiable:

- Partial unique index on `(orderRecordId, orderLineId, inventoryItemId) WHERE status = 'held'`.
- `inventory_items` gains `olReservedQuantity` with `CHECK (olReservedQuantity >= 0)` — §6I's hard floor. **No `CHECK (olReserved <= available)`**: a master drop below what OL promised is a *fact* to surface (`W2-12`), not a constraint violation to throw on.
- `claimHeld` is the guarded statement `UPDATE inventory_items SET olReservedQuantity = olReservedQuantity + $q WHERE id = $id AND availableQuantity - olReservedQuantity >= $q RETURNING …`; zero rows raises the named `InsufficientAvailabilityError` carrying the requested and available quantities.
[...]

PR #3246, opened 2026-09-14
fix(inventory): take the position lock FOR NO KEY UPDATE and re-script the reservation repository spec (#3245)

### LINKED ISSUE #3245: [BUG] CORE — #3035's position lock is red in both suites: FOR UPDATE deadlocks against the reservations FK, and the extra statement shifts every scripted reply
## Problem / Context

`main` is red in **both** test jobs, and every PR targeting it inherits the failure — including the release-please PR #2988 (`chore(main): release 0.11.0`), which is unmergeable for reasons that have nothing to do with the release.

Both failures trace to one commit: the last commit of #3035 (`fix(inventory): lock the position row before the guarded reservation UPDATE`, 8703bd92d), which added a statement to `ReservationRepository.applyGuardedAdd`:

```ts
await this.raw(manager, `SELECT 1 FROM "inventory_items" WHERE "id" = $1 FOR UPDATE`, [
  claim.inventoryItemId,
]);
```

Adding a lock there is right — it closes a real READ COMMITTED race where two concurrent claims for the last unit of stock could both pass the admission guard. **The MODE is wrong, and the unit spec was never re-run.**

---

### A — `Test` job: 10 unit tests fail (stale spec, no production defect)

```
libs/core test: FAIL src/inventory/infrastructure/persistence/repositories/reservation.repository.spec.ts
libs/core test: Tests:       10 failed, 6738 passed, 6748 total
```
(run 34896174049, job 104150711847)

`reservation.repository.spec.ts` drives the repository through a **positional** scripted query harness:

```ts
return Promise.resolve(replies.shift() ?? []);
```

Every `createHarness([...])` script was written for the pre-lock statement sequence, so the new statement shifts every downstream reply by one: the lock consumes the `{ remainingAtp }` reply meant for the guarded `UPDATE`, the guarded `UPDATE` consumes the row meant for `setQuantity`, and `setQuantity` reads the silent `?? []` fallback and raises `ReservationLedgerConstraintError('reservations.quantity')`. Ten tests fail four call frames away, none of them naming the statement that actually changed.

Two things are wrong, not one:

1. The scripts are stale.
2. `replies.shift() ?? []` means an **exhausted queue silently answers `[]`**, which is what turned a one-line implementation change into a red `main` that reads as a reservations-ledger bug.

The new lock statement is also pinned by nothing in the unit spec — removing it would go green while reopening the race.

---

### B — `Integration Tests` job: the lock deadlocks against the reservations FK (production defect)

```
FAIL test/integration/reservations-ledger.int-spec.ts
  ● Reservation ledger (#2343) › the guarded claim › should not deadlock when two multi-line claims arrive in opposite order
    Rejected to value: [ReservationLedgerConstraintError: Reservation ledger constraint violated: unknown]
```
(run 34898569897, job 104158741613)

`reservations` carries a foreign key to `inventory_items` (`@ManyToOne(..., { onDelete: 'RESTRICT' })`), so **every insert of a reservation takes a `FOR KEY SHARE` lock on the position row** for its FK check, held until commit. `FOR KEY SHARE` is shared, so two concurrent claimers both hold it — and **`FOR UPDATE` is the one row-lock mode that conflicts with it**. `claimOne` is insert-then-recover, so the INSERT (and its FK lock) happens *before* the position lock:

| | TX A | TX B |
|---|---|---|
| 1 | `INSERT reservations(A,…,X)` → holds `FOR KEY SHARE` on X | |
| 2 | | `INSERT reservations(B,…,X)` → also holds `FOR KEY SHARE` on X (shared, granted) |
| 3 | `SELECT … X FOR UPDATE` → waits on B's FK lock | |
| 4 | | `SELECT … X FOR UPDATE` → waits on A's FK lock → **`40P01 deadlock detected`** |

The pre-#3035 code never hit this: the guarded `UPDATE inventory_items SET "olReservedQuantity" = …` writes no key column, so PostgreSQL takes `FOR NO KEY UPDATE`, which is **compatible** with `FOR KEY SHARE`. The explicit lock escalated the mode and introduced the cycle.

**Sorting claims by `inventoryItemId` cannot prevent this.** The sort buys a common lock *order*, and this deadlock does not need a disordered one — both transactions are already at the same row, in the same order, when the cycle forms.

Verified deterministically with a two-connection probe against real Postgres (rather than by re-running the flaky spec, which passed twice locally against the broken code):

```
FOR UPDATE        -> {"who":"A","code":"40P01","msg":"deadlock detected"}
FOR NO KEY UPDATE -> both acquire, no deadlock
FOR NO KEY UPDATE -> a second claimer still waits (serialisation intact)
```

**This is not only a test flake.** In production two buyers reserving the same position concurrently kill one transaction with `40P01`, which `translate` reports as `ReservationLedgerConstraintError('unknown')` — i.e. "a guard that should have made this unreachable did not hold", a defect signal — and `reserveOrderInventory` swallows it best-effort (ingestion never rethrows, by design), leaving a real paid order with **no reservation at all**. That is exactly the leak #2343 / #2344 exist to prevent.

### B (production)

In `reservation.repository.ts`, take the lock as `FOR NO KEY UPDATE`. That is exactly the strength the guarded `UPDATE` takes on its own, so the explicit statement introduces no conflict class the transaction did not already have — while still conflicting with *itself*, which is the whole requirement #3035 had: a second claimer waits at the lock and starts its guarded `UPDATE` only once the first has committed, on a fresh snapshot that sees the committed reservation.

### A (spec)

1. Script the lock reply in every `createHarness([...])` whose claim reaches `applyGuardedAdd`, at the position the lock is issued. A `SELECT … FOR NO KEY UPDATE` is not data-modifying, so its reply is a plain `[]` in both the plain-array and `[rows, affectedCount]` tuple harnesses.
2. Make the harness **throw** on an exhausted queue (`Unscripted statement: <sql>`). Every existing test consumes its script exactly, so this is a no-op for a passing suite and turns the next such shift into a message naming the statement.
3. Add a regression test pinning both the ORDER (lock before the guarded `UPDATE`, same `inventoryItemId`) and the **MODE** (`FOR NO KEY UPDATE`, never `FOR UPDATE`). The mode pin is the deterministic guard: the int-spec catches the deadlock only on an unlucky interleaving, whereas the statement assertion fails every time if anyone escalates the lock again.

## Classification

**Type**: CORE
**Layer**: Infrastructure
**File(s)**:
- `libs/core/src/inventory/infrastructure/persistence/repositories/reservation.repository.ts`
- `libs/core/src/inventory/infrastructure/persistence/repositories/reservation.repository.spec.ts`
- `apps/api/test/integration/reservations-ledger.int-spec.ts` (reproduces B; unchanged)

## Dependencies

- None. #3035 is already merged; this is its follow-up.

## Assumptions

- The lock statement stays — the race it closes is real. Only its mode changes.
- `reservations-ledger.int-spec.ts` remains the authority on the guard under real concurrency; the unit spec pins the statements issued, per its own docblock.

## Acceptance Criteria

- [ ] `pnpm --filter @openlinker/core test -- reservation.repository.spec.ts` is green (10 previously-failing tests pass).
- [ ] `reservations-ledger.int-spec.ts` is green, and the four sibling reservation int-specs (`atp-subtraction`, `diagnostic-holds-are-inert`, `reservation-shortfall-episodes`, `fulfillment-work-transitions`) stay green.
- [ ] The position lock is `FOR NO KEY UPDATE`, and a unit test fails deterministically if it is escalated back to `FOR UPDATE`.
- [ ] The harness throws a named error identifying the SQL when a statement is issued with no scripted reply left, and no existing test relies on the old silent `[]` fallback.
- [ ] A test fails if the lock is removed from `applyGuardedAdd`, or is issued after the guarded `UPDATE`, or names a different position than the claim.
- [ ] `pnpm lint` and `pnpm type-check` pass.

## Out of scope (noted)

- `translate` maps a deadlock (`40P01`) to `ReservationLedgerConstraintError('unknown')`, which reads as a constraint anomaly rather than a contention failure worth retrying. Worth a SQLSTATE branch, separately.

norbert-kulus-blockydevs:
## Why

`main` is red in **both** test jobs, and every PR targeting it inherits the failure - including the release-please PR #2988 (`chore(main): release 0.11.0`), which is unmergeable for reasons that have nothing to do with the release.

Both failures trace to one commit: #3035's last commit (`fix(inventory): lock the position row before the guarded reservation UPDATE`), which added a statement to `ReservationRepository.applyGuardedAdd`:

```ts
await this.raw(manager, `SELECT 1 FROM "inventory_items" WHERE "id" = $1 FOR UPDATE`, [
  claim.inventoryItemId,
]);
```

Adding a lock there is right - it closes a real READ COMMITTED race where two concurrent claims for the last unit of stock could both pass the admission guard. **The mode is wrong, and the unit spec was never re-run.**

---

### A - `Test`: 10 unit tests fail (stale spec, no production defect)

`reservation.repository.spec.ts` drives the repository through a **positional** scripted query harness (`replies.shift() ?? []`). Every `createHarness([...])` script was written for the pre-lock sequence, so the new statement shifted every downstream reply by one - the lock consumed the `{ remainingAtp }` reply meant for the guarded `UPDATE`, the guarded `UPDATE` consumed the row meant for `setQuantity`, and `setQuantity` read the silent `?? []` fallback and raised `ReservationLedgerConstraintError('reservations.quantity')`. Ten tests failed four call frames away, every one naming a concern the change had not touched.

### B - `Integration Tests`: the lock deadlocks against the reservations FK (production defect)

```
● Reservation ledger (#2343) › the guarded claim › should not deadlock when two multi-line claims arrive in opposite order
  Rejected to value: [ReservationLedgerConstraintError: Reservation ledger constraint violated: unknown]
```

`reservations` carries a foreign key to `inventory_items`, so **every insert of a reservation takes a `FOR KEY SHARE` lock on the position row** for its FK check, held to commit. `FOR KEY SHARE` is shared, so two concurrent claimers both hold it - and **`FOR UPDATE` is the one row-lock mode that conflicts with it**. `claimOne` is insert-then-recover, so the INSERT (and its FK lock) happens *before* the position lock:

| | TX A | TX B |
|---|---|---|
| 1 | `INSERT reservations(A,…,X)` → holds `FOR KEY SHARE` on X | |
| 2 | | `INSERT reservations(B,…,X)` → also holds `FOR KEY SHARE` on X (shared, granted) |
| 3 | `SELECT … X FOR UPDATE` → waits on B's FK lock | |
| 4 | | `SELECT … X FOR UPDATE` → waits on A's FK lock → **`40P01`** |

The pre-#3035 code never hit this: the guarded `UPDATE … SET "olReservedQuantity" = …` writes no key column, so PostgreSQL takes `FOR NO KEY UPDATE`, which is **compatible** with `FOR KEY SHARE`. The explicit lock escalated the mode and introduced the cycle.

**Sorting claims by `inventoryItemId` cannot prevent this.** The sort buys a common lock *order*, and this deadlock does not need a disordered one - both transactions are already at the same row, in the same order, when the cycle forms.

**It is not only a test flake.** In production two buyers reserving the same position concurrently kill one transaction with `40P01`, which `translate` reports as `ReservationLedgerConstraintError('unknown')` - i.e. "a guard that should have made this unreachable did not hold", a defect signal - and `reserveOrderInventory` swallows it best-effort (ingestion never rethrows, by design), leaving a real paid order with **no reservation at all**. Exactly the leak #2343 / #2344 exist to prevent.

## What changed

Two files.

**`reservation.repository.ts`** - the lock becomes `FOR NO KEY UPDATE`. That is exactly the strength the guarded `UPDATE` takes on its own, so the explicit statement introduces no conflict class the transaction did not already have, while still conflicting with *itself* - which is the whole requirement #3035 had: a second claimer waits at the lock and starts its guarded `UPDATE` only once the first has committed, on a fresh snapshot that sees the committed reservation.

**`reservation.repository.spec.ts`**

1. **Script the lock reply** (`lockReply()`, a factory rather than one array shared across a dozen harnesses) in every harness whose claim reaches `applyGuardedAdd`. The `delta === 0` / `delta < 0` scripts and both `releaseHeld` scripts are untouched. A `SELECT … FOR NO KEY UPDATE` is not data-modifying, so the driver surfaces it as a plain row array and never as a `[rows, affectedCount]` tuple - hence the same value in the tuple-shaped harnesses.
2. **An exhausted queue now throws** `Unscripted statement: <sql>` instead of answering `[]`. Every existing test consumes its script exactly, so this is a no-op for a passing suite; what it buys is that the *next* statement added to this path fails at the statement with no reply, rather than surfacing as a domain error about an unrelated concern several frames downstream. The silent fallback is what turned a one-line implementation change into a red `main` that reads as a reservations-ledger bug.
3. **Pin the order AND the mode** - the lock is issued before the guarded `UPDATE`, against the claim's own `inventoryItemId`, as `FOR NO KEY UPDATE`. The two ways this can break report differently, on purpose: the statement finder is **mode-agnostic** (`sql.startsWith('SELECT 1 FROM "inventory_items"')`), so a **deleted** lock fails the index assertion (`Expected: >= 0 / Received: -1`, "no lock") while an **escalated** one is still found and fails a positive `toContain('FOR NO KEY UPDATE')` that names the mode. A finder carrying the mode would report both as a missing lock, which is the same failure-message defect this PR exists to remove, one level down. The mode pin is the load-bearing half: the int-spec catches the deadlock only on an unlucky interleaving (it passed twice locally against the broken code), whereas this assertion fails every time if anyone escalates the lock again. The test discards the claim's outcome and asserts over the recorded statements, because a script written for the locked sequence necessarily derails when the lock is gone - but it deliberately **rethrows** an `Unscripted statement:` rejection, so the test pinning the statement ORDER is not the one test in the file blind to the guard added in (2).

## Verification

Run locally, not inferred.

**The deadlock, proved deterministically.** Re-running the int-spec is not evidence - it passed twice against the broken code. A two-connection probe against real Postgres reproduces the exact lock interaction:

```
FOR UPDATE        -> {"who":"A","code":"40P01","msg":"deadlock detected"}
FOR NO KEY UPDATE -> both acquire, no deadlock
FOR NO KEY UPDATE -> a second claimer still waits   (serialisation intact)
```

The third line matters as much as the first: it confirms the weaker mode still delivers what #3035 needed.

**Integration.** `reservations-ledger.int-spec.ts` reproduces the CI failure on this branch before the change, and is **19/19 green 3 runs in a row** after it. The four sibling reservation int-specs - `atp-subtraction`, `diagnostic-holds-are-inert`, `reservation-shortfall-episodes`, `fulfillment-work-transitions` - are **46/46 green**.

**Unit.** `reservation.repository.spec.ts` → **20/20** (19 existing + 1 new; 10 were failing). The new test is red-first for both mutations and tells them apart: deleting the lock fails `expect(lockIndex).toBeGreaterThanOrEqual(0)` (`Expected: >= 0 / Received: -1`), while re-escalating it to `FOR UPDATE` still finds the statement and fails the positive mode assertion, naming the mode. Neither surfaces as an availability error. Deleting the lock is not otherwise silent, and the comment in the test says so: simulated against the real reply-consumption logic, **11 of the 12** locked-path harnesses change outcome without it, landing either on `InsufficientAvailabilityError` with a NaN availability or on a mislabelled `'missing'` position, in a branch their script never mentions. Only *reason missing* is accidentally unaffected.

**Gate.** `tsc --noEmit` on `libs/core` clean; `eslint` on both changed files 0 errors (1 pre-existing `explicit-function-return-type` warning on `createHarness`, unchanged from `main`). Prettier reports the spec as unformatted both before and after, so no formatting regression is introduced and none is fixed here (`format:check` is not part of the `Lint` job).

**Audit.** Every other `FOR UPDATE` in `libs/core` was checked for the same shape. One site **does** have it, and is safe by construction rather than by luck: `InventoryRepository.backfillLegacyProvenance` (#2317) takes `FOR UPDATE SKIP LOCKED` on `inventory_items`, which genuinely conflicts with the FK's `FOR KEY SHARE` - but `SKIP LOCKED` means it can never be the *waiting* side of a cycle, so it cannot deadlock. Its residual cost is not a deadlock but a drain rate: every position carrying an uncommitted reservation INSERT is skipped, so the backfill slows as reservation volume grows, for no benefit - `sourceConnectionId` is not a key column and is excluded from both partial unique indexes, so `FOR NO KEY UPDATE SKIP LOCKED` is the correct strength there and would skip strictly fewer rows. **Deliberately not changed here:** it is a different pass with a different risk profile, and this PR is scoped to un-redding `main`; noted as a follow-up. `FulfillmentWorkRepository.placeHold` is the closest analogue to *this* statement and is safe for a third reason again: it locks the parent row *before* inserting the child, so the requester holds no FK lock of its own when it asks, and no cycle can form.

## Not in scope

- **The `40P01` mapping, and the cycle that survives this fix.** `translate` maps a deadlock (`40P01`) or a serialization failure (`40001`) to `ReservationLedgerConstraintError('unknown')` - i.e. as a constraint anomaly, which this same file documents as "a guard that should have made this unreachable did not hold", a defect signal - rather than as contention worth retrying, and `reserveOrderInventory` swallows it best-effort.

  This PR removes **this** deadlock at its source, not the whole class. The known survivor is `claimHeld` against `releaseHeld`, which take the same two rows in **opposite order**: `claimOne` reaches `inventory_items` first (the FK's `FOR KEY SHARE` on INSERT, then the position lock) and touches its `reservations` row only at `setQuantity`, while `releaseHeld` locks the `reservations` row first and `inventory_items` after. So a claim whose next `INSERT … ON CONFLICT DO NOTHING` blocks on a mid-release row's live index entry (`XactLockTableWait`, an ordinary heavyweight lock, so the detector fires), against a release whose counter `UPDATE` blocks on that claim's position lock, is a genuine cycle. It needs no two concurrent ingestions, so no per-order ingestion lock excludes it, and the `inventoryItemId` sort is scoped to one `claimHeld` call so it cannot order two **different** methods against each other.

  It is **pre-existing, not introduced here** - before #3035 the guarded `UPDATE` took the same `FOR NO KEY UPDATE` on the position, so the identical cycle was already reachable. `translate`'s docblock now records it rather than leaving the next reader to rediscover it. Splitting the SQLSTATE out into a distinct **retryable** error (the `InventoryCrossSourcePositionConflictError` shape, #2320) is the right fix and is a deliberate follow-up: it widens what every caller of this ledger must handle, which is a far larger blast radius than a lock mode. Not tracked by an issue yet.
- No change to the int-spec. It reproduced the bug and is the authority on the guard under real concurrency; the deterministic guard against a re-escalation is the unit-level mode assertion.

Closes #3245


norbert-kulus-blockydevs:
## `/pr-review` - Tech Lead review

**Verdict: 🔄 Approve with changes.** No BLOCKING issues. The diagnosis is right, the fix is the right one, it is minimal, and all 9 checks are green on this branch. Nothing below is a reason to hold a fix for a red `main`.

Every claim in the PR body was re-derived independently rather than taken on trust, and the central ones hold. The findings split into one substantive follow-up (a residual deadlock cycle the PR body's "removes the deadlock at its source" understates) and four items inside the new test and its comments - three of which are the same failure-message-quality defect this PR exists to fix, reintroduced in the code that fixes it.

---

## Verified correct

**1. The lock-conflict reasoning is accurate.** `FOR KEY SHARE` conflicts with `FOR UPDATE` and nothing else. `FOR NO KEY UPDATE` conflicts with `FOR SHARE` / `FOR NO KEY UPDATE` / `FOR UPDATE` but **not** with `FOR KEY SHARE`, and it conflicts with **itself** - which is the entire serialisation requirement #3035 was added for. Both halves of the central claim hold.

**2. The FK really does take `FOR KEY SHARE` on the parent, before the lock runs.** `reservation.orm-entity.ts` declares `FK_reservations_inventory_item` (`ON DELETE RESTRICT`), confirmed in migration `1850000000009`. Since PG 9.3 the RI check for an inserted child is `SELECT ... FOR KEY SHARE` on the parent, fired at end-of-statement and held to commit. `claimOne` inserts before `applyGuardedAdd`, so the FK lock is in place when the position lock is requested. The four-step deadlock table in the PR body is exactly right.

**3. The guarded `UPDATE` really does take `FOR NO KEY UPDATE`, and cannot escalate.** `inventory_items`' PK is `id`, and both unique indexes on it are **partial** (`where: '"productVariantId" IS NULL'` / `IS NOT NULL`). PostgreSQL's escalation rule excludes partial indexes, so writing `olReservedQuantity` can never escalate the mode. The explicit statement therefore genuinely introduces no conflict class the transaction did not already have.

**4. The #3035 guarantee survives the de-escalation.** Walked the two-claimer, one-unit case against the shipped code: A takes `FOR NO KEY UPDATE` (compatible with B's `FOR KEY SHARE`), B blocks at the lock; A's guarded `UPDATE` sees `othersSum = 0` and grants; on A's commit B unblocks, and because its guarded `UPDATE` is a **new command** under READ COMMITTED it takes a fresh snapshot including A's committed row, so `othersSum = 1` and it correctly raises `InsufficientAvailabilityError`. The "lock as its own prior statement" argument is airtight, and is load-bearing precisely because `othersSum` is a correlated subquery against another table that an EPQ re-fetch would not refresh.

**5. The `inventoryItemId` sort still buys what its comment claims, and buys more than before.** `FOR NO KEY UPDATE` is still exclusive with itself, so a common lock order is still required for the multi-position case. Before this PR the sort was *insufficient* (the FK cycle fired even under a common order, which is exactly why the int-spec test named for the sort is the one that went red); it is now the only remaining protection for that case, so it is more load-bearing, not less.

**6. No other row lock in `libs/core` reopens the cycle.** Audited all of them. `FulfillmentWorkRepository.placeHold` and `lockWorkForVerification` take the parent `FOR UPDATE` **before** inserting the FK child (`FulfillmentVerificationService.verifyUnit` calls `lockWorkForVerification` at `:86` and `recordParcelVerification` only at `:131`), so the requester holds no FK lock when it asks. `ReturnRepository.runLineWrite` / `settleLineRestock` have the same safe order. `order_line_items` and `return_line_events` carry no FK at all, so `authority-attention-sql`'s `FOR UPDATE` is not exposed.

**7. The sibling `FOR UPDATE` on `reservations` at `:289` is genuinely not the same shape.** `reservations` is a child only: one FK out to `inventory_items`, and nothing in the tree FKs *into* it, so no `FOR KEY SHARE` is ever taken on a reservations row. (The `setQuantity` comment at `:513` referring to that statement is therefore accurate, not stale.)

**8. The spec re-script is complete and the harness change is a true no-op for a passing suite.** Walked all 20 harnesses against the four code paths. The `delta === 0` path (2 statements) and `delta < 0` path (4 statements) legitimately never reach `applyGuardedAdd` and correctly take no `LOCK_REPLY`; both `releaseHeld` scripts are exact; the error-translation test replaces `transaction` outright so `query` is never called. `LOCK_REPLY` being the same `[]` in the tuple-shaped harnesses is right - `raw()` does `Array.isArray(outer[0]) ? outer[0] : outer`, and TypeORM's `PostgresQueryRunner` only builds the `[rows, rowCount]` tuple for `command === 'UPDATE' | 'DELETE'`.

**9. Turning the `?? []` fallback into a throw is the right structural fix**, independent of this bug. It converts the failure class that produced the red `main` - a positional shift surfacing as `ReservationLedgerConstraintError` four frames away, naming a concern the change never touched - into a failure that names the unscripted statement. `Promise.reject` vs a synchronous `throw` is a non-issue: every call site is `await manager.query(...)` inside `raw()`, so the handler attaches in the same frame.

**10. Two assertions in the new test have real power.** `expect(...params).toEqual(['inv-b'])` catches locking `reservationId` instead of `claim.inventoryItemId`, and `guardIndex` via `'+ $3'` is unique to the guarded UPDATE (`publishedReservedSum` emits no `+ $n`, the INSERT has no `+`), so it fails closed if the parameter numbering moves.

Also worth saying: diagnosing this as **two independent failures with one cause** rather than "the spec is flaky" is why the fix is one word instead of a retry loop. Proving the deadlock with a two-connection probe rather than re-running the int-spec, and stating out loud that the int-spec **passed twice against the broken code**, is exactly the honesty `docs/testing-guide.md § Port-contract suites` asks for. The third probe line (a second claimer still waits under the weaker mode) is what makes it evidence rather than an assertion.

---

## Findings

### [IMPORTANT] `reservation.repository.ts:544` (`translate`) - the deadlock class is not fully eliminated, and the `40P01` mapping is untracked

"## Not in scope" files the SQLSTATE branch away on the grounds that "the fix here removes the deadlock at its source". That is one step too strong: it removes the **FK-versus-`FOR UPDATE`** cycle, not every cycle on these two tables. A `claimHeld` / `releaseHeld` cycle survives, because the two take the same two rows in **opposite order**:

- `claimOne` reaches `inventory_items` first (`FOR KEY SHARE` via the FK at `:275`, then `FOR NO KEY UPDATE` at `:459`) and only touches the `reservations` row at `setQuantity`.
- `releaseHeld` (`:130` to `:160`) locks the `reservations` row first (the guarded `UPDATE ... status='held'`) and only then `inventory_items`.

Concrete interleaving, one order `O` whose line `L2` is being released while the order is re-ingested (routine: a cancellation's `closeForOrder` racing a re-poll), or one order with two lines resolving to the same position `X`:

| | TX A - `claimHeld` (ingestion) | TX B - `releaseHeld` (cancel / consume / expiry) |
|---|---|---|
| 1 | line 1: INSERT, then `FOR NO KEY UPDATE` on X, **held** | |
| 2 | | `UPDATE reservations` R2, holds the row lock on R2 |
| 3 | line 2: `INSERT ... ON CONFLICT DO NOTHING` hits R2's live index entry, which is mid-update, so it does `XactLockTableWait(B)` | |
| 4 | | `UPDATE inventory_items` X, waits on A, **`40P01`** |

`XactLockTableWait` is an ordinary heavyweight lock, so the detector fires. This needs no two concurrent ingestions, so no per-order ingestion lock excludes it, and the `inventoryItemId` sort is per-`claimHeld`-call only so it cannot order against a different method.

**This is pre-existing, not introduced here** - before #3035 the guarded `UPDATE` took the same `FOR NO KEY UPDATE` on X, so the identical cycle was already reachable. But it does mean the consequence the PR body describes for the bug it fixes is still live: `translate` reports it as `ReservationLedgerConstraintError('unknown')`, which **this same file** documents at `:271-273` as meaning "a guard that should have made this unreachable did not hold, i.e. a defect signal", and `OrderIngestionService.reserveOrderInventory:1683` swallows it into one `logger.error` - so retryable contention is logged as a ledger defect and a real paid order silently keeps no hold.

`gh issue list --search "40P01 OR deadlock"` returns only #3245, #2626 and #2730. **Nothing tracks this.** Please either open the issue the PR body promises and cite it in the `translate` docblock, or add the branch:

```ts
const state = this.sqlStateOf(error);
if (state === '40P01' || state === '40001') {
  throw new ReservationContentionError(state, error); // retryable, NOT a constraint anomaly
}
```

The repo's precedent for exactly this shape is `InventoryCrossSourcePositionConflictError` plus a greppable `inventory_cross_source_position_conflict` log (#2320). Reference the cycle above in the issue so it is not filed as cosmetic. Also worth one word in the PR body: "removes **this** deadlock at its source".

---

### [IMPORTANT] `reservation.repository.spec.ts:191` - the mode assertion is unreachable, and an escalation is misdiagnosed as a deletion

```ts
const lockIndex = h.statements.findIndex(
  (s) => s.sql.includes('"inventory_items"') && s.sql.includes('FOR NO KEY UPDATE'),
);
...
expect(h.statements[lockIndex].sql).not.toContain('FOR UPDATE');   // :191
```

`h.statements[lockIndex]` contains `FOR NO KEY UPDATE` **by construction**, and `'...FOR NO KEY UPDATE'.includes('FOR UPDATE') === false` (checked at a node prompt, not inferred - there is no contiguous `FOR UPDATE` substring in it). So `:191` can only fail on SQL containing *both* modes, an impossible shape. It is dead weight.

The comment above it calls that line "The MODE is load-bearing, not incidental", which misattributes which line does the work. The mode is in fact pinned by the **finder predicate**: re-escalate to `FOR UPDATE` and `findIndex` returns `-1`, so the suite dies at `:177` with `Expected: >= 0 / Received: -1` - **"the lock is missing"**, which is the wrong diagnosis for a lock that is present with the wrong mode, and indistinguishable from the lock being deleted. That is precisely the standard `docs/testing-guide.md:527` sets ("check the red is for the right reason") and the one this PR's own `createHarness` docblock sets ("naming a concern the change had not touched").

It also makes the PR body's coverage answer only half-true: re-escalation *is* caught deterministically, but by the finder, not by "the unit-level mode assertion", and the message misnames the defect. A later reader who relaxes the finder to improve that message would reasonably believe `:191` still holds the line, and it asserts nothing.

```ts
const lockIndex = h.statements.findIndex((s) =>
  s.sql.startsWith('SELECT 1 FROM "inventory_items"'),
);
expect(lockIndex).toBeGreaterThanOrEqual(0);                        // deletion   -> "no lock"
expect(h.statements[lockIndex].sql).toContain('FOR NO KEY UPDATE'); // escalation -> names the mode
```

Two mutations, two distinct correct messages.

---

### [IMPORTANT] `reservation.repository.spec.ts:155-158` - the comment states something that is false

```ts
// Nothing else in this suite has an opinion on it: the scripts tolerate the
// statement, and deleting it would leave them green while reopening the
// oversell. Only `reservations-ledger.int-spec.ts` would catch that, and
// only intermittently.
```

Deleting the lock does **not** leave them green. Every locked-path script is then over-supplied by one: the guarded `UPDATE` reads `LOCK_REPLY` (`[]`), returns zero rows, and the discriminating probe eats the row meant for `setQuantity`. Simulated against the real reply-consumption logic:

```
with lock   : RESOLVES remainingAtp=4
lock deleted: THROWS InsufficientAvailabilityError(atp=NaN)
```

11 of the 12 locked-path tests go red (only `:341`, *reason missing*, accidentally still passes). The comment tells the next reader the suite is indifferent to the lock, which would license deleting it and expecting a single failure. The truth is better and should be written instead: deleting it turns every locked-path script red with an availability error four frames away, and **this** test is the one that names the lock.

---

### [IMPORTANT] `reservation.repository.spec.ts:171` - `.catch(() => undefined)` disarms the new guard in the one test that most needs it

The whole point of the `createHarness` change is that a newly added statement fails **at** the statement. This test swallows every rejection, `Unscripted statement: ...` included. Add a sixth statement to `claimOne` and every other test in the file goes red naming it, while the test that documents the lock ordering stays green - the single test blind to the change most likely to break the ordering it pins. Narrow it:

```ts
await h.repository.claimHeld([claim({ quantity: 3 })]).catch((e: unknown) => {
  if (e instanceof Error && e.message.startsWith('Unscripted statement:')) throw e;
});
```

---

### [SUGGESTION] `reservation.repository.spec.ts:58` and `:72` - both new docblocks name the wrong mode

```
:58  * `SELECT 1 … FOR UPDATE` is not a data-modifying statement, so the driver
:72  * issues — including `applyGuardedAdd`'s `SELECT … FOR UPDATE`, whose own reply
```

In a PR whose entire thesis is that the two modes are not interchangeable, the two comments explaining why `LOCK_REPLY` exists name the mode that caused the outage. `grep -n "FOR UPDATE" reservation.repository.spec.ts` now returns the fix's own documentation as a false hit, and the next author to copy the `LOCK_REPLY` idiom learns the wrong statement. One word in each. (The substantive claim - a locking `SELECT` is not data-modifying and so never arrives as a `[rows, affectedCount]` tuple - is correct and worth keeping.)

---

### [SUGGESTION] `inventory.repository.ts:1077` - the audit claim in the PR body is slightly off, and the site it misses is worth downgrading

The body says "Every other `FOR UPDATE` in `libs/core` was checked for the same shape; none has it." `backfillLegacyProvenance` (#2317) *does* have the shape - `FOR UPDATE SKIP LOCKED` on `inventory_items`, which conflicts with the FK's `FOR KEY SHARE` - it is simply immunised against deadlock by `SKIP LOCKED`, so it can never be the waiting side of a cycle. Safe by construction rather than by luck, which is a better sentence than the one in the body.

The residual cost is real though: every position carrying an uncommitted reservation INSERT is now skipped, so the backfill's drain rate degrades as reservation volume grows, for no benefit - `sourceConnectionId` is not a key column and is excluded from both partial unique indexes. `FOR NO KEY UPDATE SKIP LOCKED` is the correct strength there and skips strictly fewer rows. Separate change; worth a line in the body so the audit claim is accurate.

---

### [SUGGESTION] `reservation.repository.spec.ts:1-15` and `:64` - file header, and a shared mutable `LOCK_REPLY`

The file header enumerates what the suite covers and says the tests "deliberately do NOT claim to prove the guards"; the new test adds a concern - the lock's ordering **and** mode - that the header neither lists nor reconciles with that disclaimer (`docs/engineering-standards.md § File Headers`). One line fixes it.

Separately, `LOCK_REPLY` is one shared mutable array across roughly 12 harnesses. Nothing mutates it today (`raw()` only reads `outer[0]`), so it is latent rather than a defect, but `Object.freeze` or a `lockReply()` factory removes the class for free.

---

## Standards check

| Dimension | Result |
|---|---|
| Layer placement | ✅ Infrastructure persistence repository; raw SQL confined to the two guarded statements the class docblock already declares |
| SQL injection | ✅ Every value bound. The only interpolation is `othersSum`, a code-owned fragment with parameter placeholders, unchanged here |
| Error handling | ✅ `translate` remains the single boundary and no `QueryFailedError` escapes. The `40P01` classification gap is a follow-up, not a regression |
| Naming / structure | ✅ No new files, no naming surface touched |
| Migration | ✅ N/A, no ORM entity or schema change |
| Types | ✅ No `any`; `unknown[]` throughout |
| Tests | ⚠️ Re-script is exact and the int-spec is correctly left as the concurrency authority, but see the three test findings above |

## Priority

1. Open (or point at) the `40P01` follow-up issue and soften "removes the deadlock at its source" to "removes this deadlock". Nothing tracks it today.
2. Fix the mode pin so escalation and deletion produce different messages (`:173-191`).
3. Narrow the `.catch` at `:171` so the unscripted-statement guard still applies to this test.
4. Correct the false claim at `:155-158`, and the two `FOR UPDATE` docblocks at `:58` / `:72`.

1 through 4 are all inside the test file and its comments. None of them blocks getting `main` green.

norbert-kulus-blockydevs:
## Review findings addressed - `a8a3cbda1`

All 7 findings from the review above are now fixed. **No runtime behaviour changed**: the diff is one docblock in the repository, one test, and the PR body. `+105 / -31` across 2 files.

| # | Finding | What changed |
|---|---|---|
| F1 | `40P01` mislabelled; deadlock class not fully removed | `translate` docblock in `reservation.repository.ts` now records it |
| F2 | Mode assertion vacuous, escalation misdiagnosed | Finder made mode-agnostic, mode asserted positively |
| F3 | Comment stated something false | Replaced with the measured behaviour |
| F4 | `.catch` swallowed `Unscripted statement:` | Narrowed to rethrow it |
| F5 | Two docblocks named `FOR UPDATE` | Both corrected |
| F6 | Audit claim inaccurate | PR body corrected, no code change |
| F7 | Header + shared mutable reply | Header extended, `lockReply()` factory |

---

### F1 - the residual cycle is recorded, deliberately not fixed

`translate`'s docblock now states, in the file's own voice, that a `40P01` / `40001` arrives with no constraint name and lands as `ReservationLedgerConstraintError('unknown')` - which `claimOne` documents as a defect signal - while `reserveOrderInventory` swallows it best-effort, so the visible outcome is a paid order holding nothing while the log blames the ledger. It names the surviving `claimHeld` / `releaseHeld` opposite-order cycle and why neither the `inventoryItemId` sort nor a per-order ingestion lock orders it away, and records that the SQLSTATE branch (the `InventoryCrossSourcePositionConflictError` shape, #2320) is the right fix, is deliberately not taken because it widens what every ledger caller must handle, and carries no issue yet.

**No SQLSTATE branch, no new error class, no issue filed.** The "Not in scope" section now says this PR removes **this** deadlock at its source, not the whole class, and carries the cycle description.

### F2 - the mode pin now actually pins the mode

```diff
-const lockIndex = h.statements.findIndex(
-  (s) => s.sql.includes('"inventory_items"') && s.sql.includes('FOR NO KEY UPDATE'),
-);
+const lockIndex = h.statements.findIndex((s) =>
+  s.sql.startsWith('SELECT 1 FROM "inventory_items"'),
+);
 expect(lockIndex).toBeGreaterThanOrEqual(0);
-expect(h.statements[lockIndex].sql).not.toContain('FOR UPDATE');
+expect(h.statements[lockIndex].sql).toContain('FOR NO KEY UPDATE');
```

Verified the predicates against the real statement strings rather than by eye:

```
finder matches real lock     : true    <- must be true
finder matches ESCALATED     : true    <- must be true, so the mode assertion fires
  escalated passes mode?     : false
finder matches probe         : false
finder matches recovery      : false
finder matches guarded UPDATE: false
```

So the two mutations now give two different, correct messages: a deleted lock fails the index assertion ("no lock"), an escalated one fails the mode assertion and names the mode. Previously an escalation reported `Expected: >= 0 / Received: -1`, i.e. "the lock is missing" about a lock that was present and wrong.

### F3 - the false comment, and a correction to the correction

The original comment said deleting the lock "would leave them green". It would not. I wrote an independent simulation of `raw()`'s reply normalisation plus the `claimOne` / `applyGuardedAdd` control flow and ran all 12 locked-path harnesses both ways:

```
CHANGED  claim ordering          with=ok(remainingAtp=5)|ok(remainingAtp=7)  without=Insufficient(atp=NaN)|LedgerConstraintError
CHANGED  the lock test itself    with=ok(remainingAtp=4)                     without=Insufficient(atp=NaN)
CHANGED  full quantity as delta  with=ok(remainingAtp=4)                     without=Insufficient(atp=NaN)
CHANGED  excludes own row        with=ok(remainingAtp=4)                     without=Insufficient(atp=NaN)
CHANGED  diagnostic stamp        with=ok(remainingAtp=10)                    without=Insufficient(atp=NaN)
CHANGED  recovers on conflict    with=ok(remainingAtp=1)                     without=Insufficient(atp=NaN)
CHANGED  live but short          with=Insufficient(atp=4)                    without=Unavailable(missing)
CHANGED  stale position          with=Unavailable(stale)                     without=Unavailable(missing)
same     missing position        with=Unavailable(missing)                   without=Unavailable(missing)
CHANGED  does not mutate claims  with=ok(remainingAtp=1)|ok(remainingAtp=1)  without=Insufficient(atp=NaN)|LedgerConstraintError
CHANGED  tuple shape             with=ok(remainingAtp=4)                     without=Insufficient(atp=NaN)
CHANGED  tuple zero-row refusal  with=Insufficient(atp=1)                    without=Unavailable(missing)

=> 11 of 12 locked-path harnesses change outcome when the lock is deleted.
```

The headline **11 of 12** holds under two independent derivations, and only *reason missing* is unaffected - because it already expects the branch the shift pushes it into.

Worth flagging: the first attempt at this comment replaced the false claim with a **different** wrong figure ("six landing on `InsufficientAvailabilityError(atp=NaN)` and the rest on a mislabelled `'missing'` position"). The real split is 8 / 3. Since F3 exists precisely to stop a false number sitting in that comment, the breakdown is dropped in favour of the failure *shapes* plus the one count both derivations agree on. The same figure has been corrected in the commit message and in the PR body's Verification section.

### F4 - the guard now applies to the test that most needs it

```ts
await h.repository.claimHeld([claim({ quantity: 3 })]).catch((e: unknown) => {
  if (e instanceof Error && e.message.startsWith('Unscripted statement:')) throw e;
});
```

Otherwise the one test pinning the statement ORDER would sit green while every other test in the file went red at a newly added statement. Confirmed from the simulation that a deleted lock consumes only 3 of the 4 scripted replies, so no spurious unscripted rejection masks the real `lockIndex` assertion.

### F5 / F7 - comments, header, shared state

Both docblocks explaining the lock reply now say `FOR NO KEY UPDATE`; the substantive claim (a locking `SELECT` is not data-modifying, so the driver never returns a `[rows, affectedCount]` tuple) is kept. The file header lists the lock's position and mode among what the suite covers, and adds a paragraph reconciling that with its own "these do not prove the guards" disclaimer: both are properties of the statement TEXT, which the harness records verbatim, and what the mode does to a concurrent transaction is explicitly not claimed here.

`LOCK_REPLY` became a `lockReply()` factory across all 15 call sites (`Object.freeze([])` types as `readonly never[]` and would not assign to `createHarness(replies: unknown[][])` without a cast). Nothing mutated the shared array today, so this closes a latent class rather than a live defect.

### F6 - no code change, corrected claim

`inventory.repository.ts` is untouched. The PR body's Audit paragraph now says what is true: `backfillLegacyProvenance` (#2317) **does** carry the same `FOR UPDATE`-against-FK shape, and is safe by construction rather than by luck, because `SKIP LOCKED` means it can never be the waiting side of a cycle. Its residual cost is a drain rate rather than a deadlock, and `FOR NO KEY UPDATE SKIP LOCKED` is recorded as the correct strength for a separate follow-up.

---

### Gate

```
tsc --noEmit -p libs/core/tsconfig.json   -> exit 0
eslint <the two changed files>            -> exit 0, 0 errors, 1 warning
```

The warning is `Missing return type on function` for `createHarness`, pre-existing on `main` and untouched. `prettier --check` reports the spec as unformatted both before and after, so no formatting regression is introduced and none is fixed (`format:check` is not part of the `Lint` job). No tests were run locally by design; CI is the gate.

### Still open

One item is deliberately left for a separate change, and neither is a regression from this PR:

1. **The `40P01` SQLSTATE branch** in `translate`, and the `claimHeld` / `releaseHeld` cycle it would cover. Recorded in the docblock, **not tracked by an issue** - happy to file one on request.
2. **`FOR NO KEY UPDATE SKIP LOCKED`** in `backfillLegacyProvenance`, a one-word downgrade in a different pass with a different risk profile.

piotrswierzy (approved):

**Verdict: ✅ Approve.** This is the right fix, for the right reason, and the reasoning is verifiable from the source rather than taken on trust.

### Why the diagnosis holds

I checked the two load-bearing claims against the code and against PostgreSQL's lock-conflict matrix rather than against the description:

1. **The FK lock really is taken before the position lock.** `claimOne` does the `INSERT INTO "reservations"` (line ~352) and only then calls `applyGuardedAdd` (line 313 → 412). So by the time a claimer reaches the explicit lock, it already holds `FOR KEY SHARE` on that `inventory_items` row from its own FK check, held to commit. The cycle in the description is reachable exactly as drawn.
2. **`FOR UPDATE` is the only mode that conflicts with `FOR KEY SHARE`**, and `FOR NO KEY UPDATE` conflicts with itself while *not* conflicting with `FOR KEY SHARE`. So the weaker mode is precisely the one that removes the cycle without removing the serialisation #3035 needed — which is the property that had to be true for this to be a fix rather than a revert-in-disguise, and the two-connection probe in the description tests exactly that third line (`a second claimer still waits`) rather than only the absence of the deadlock.
3. **The guarded `UPDATE` writes `olReservedQuantity`**, not a key column, so `FOR NO KEY UPDATE` is genuinely the strength that statement takes on its own — the explicit lock introduces no conflict class the transaction did not already have. Confirmed in the SQL.

The note that ordering by `inventoryItemId` cannot help is correct and worth having written down: both transactions are already at the same row when the cycle forms, so a common lock *order* buys nothing.

### The spec changes are the better half

Three things here are stronger than the minimum:

- **The exhausted queue now rejects instead of answering `[]`.** This is the actual root cause of "red `main` that reads as a reservations-ledger bug" — a positional harness with a silent fallback converts *one added statement* into ten failures four frames downstream, every one naming an untouched concern. Removing the fallback is the fix to the class, not just to the instance.
- **The mode-agnostic statement finder is the right call, and the comment explains why.** A finder that required `FOR NO KEY UPDATE` would report an *escalation* as `Received: -1`, i.e. "the lock is missing" about a lock that is present and wrong — reintroducing the same failure-message defect one level down. Splitting *found* from *correct mode* so the two break differently is the point.
- **Rethrowing only `Unscripted statement:` from the discarded-outcome test.** A blanket `catch` would have made the one test pinning statement order the one test blind to the new guard. That is an easy thing to get backwards and it is not backwards here.

The mode assertion is also the load-bearing guard, correctly identified: the int-spec catches the deadlock only on an unlucky interleaving (it passed twice against the broken code), so the deterministic regression barrier is the unit-level `toContain('FOR NO KEY UPDATE')`.

### Findings

**[SUGGESTION]** — `reservation.repository.ts`, `translate()` docblock
> The surviving `claimHeld`-vs-`releaseHeld` cycle and the `40P01 → ReservationLedgerConstraintError('unknown')` mislabel are now documented in place, which is right, but the PR says the follow-up "carries no issue yet". Given `reserveOrderInventory` swallows it best-effort, the visible production outcome is a paid order holding nothing while the log blames the ledger — that deserves a tracked issue rather than only a docblock. Same for the `backfillLegacyProvenance` `FOR UPDATE SKIP LOCKED` → `FOR NO KEY UPDATE SKIP LOCKED` drain-rate note; the audit found it, and an untracked finding tends to stay found-and-forgotten.

**[Positive]** — the audit of every other `FOR UPDATE` in `libs/core`, with each site's safety argued on its own terms (`SKIP LOCKED` can never be the waiting side; `placeHold` locks the parent *before* inserting the child) rather than waved through, is exactly the scope-check this class of change needs.

### Merge readiness

✅ **Approve** — all nine checks green, scope is tight (two files), the change is verified by a deterministic probe rather than by a re-run of a test that had already passed against the broken code, and `main` is unblocked along with the release PR. The one suggestion is a follow-up, not a condition.

PR #3249, opened 2026-09-15
feat(inventory): enrich duplicate-positions response with product/variant/connection/location display names

### LINKED ISSUE #3239: [TASK] CORE — enrich duplicate-positions response with product/variant/connection/location display names
## Problem / Context

`GET /inventory/duplicate-positions` (#2319) returns only raw internal ids per
group — `productId`, `productVariantId`, `locationId`, `sourceConnectionId`
(`apps/api/src/inventory/http/dto/duplicate-positions-response.dto.ts`,
`DuplicatePositionGroupResponseDto`) — with no human-readable identity. An
operator triaging a duplicate-position group cannot tell which SKU or which
connection is at risk without a manual lookup per id.

The reviewed frontend mockup for epic #3071 (artifact from PR #3121, now
`docs/plans/mockups/duplicate-positions-3071.html`) modeled
`productName`/`sku`/`connectionName`/`locationName` per group and flagged this
in-file as "mockup enrichment beyond the currently-documented response
shape... implementers should confirm the real endpoint can join these in...
before wiring the real page 1:1 against this mock." Epic #3071 deliberately
scoped this out — its own stated Assumption is "Backend contract is stable...
no API changes needed", and its sub-issues (#3072–#3077) render raw ids only.
This issue is that confirmation/implementation, split out on purpose so it
does not block or get silently dropped from the frontend epic.

## Proposed Solution

Add the enrichment in `InventoryQueryService.getDuplicatePositionReport`
(`libs/core/src/inventory/application/services/inventory-query.service.ts`),
which already composes `InventoryRepositoryPort.findDuplicatePositions` with a
cross-context product read (`IProductsService`, injected today) — this is the
natural extension point, not the raw repository/SQL layer.

Per axis, collect the **unique** ids across the returned `groups[]` and
resolve them in one batched (or small, unique-id-bounded) pass — never per
group, per the `getEarliestOrderDateByConnection` (#2083) precedent of one
batched read ahead of any per-row loop:

- `productId` → `productName` + `sku`, via **`IProductsService.getProductsByIds(ids)`**
  — already injected into `InventoryQueryService` and already batched. No new
  dependency.
- `productVariantId` → variant identity where it meaningfully differs from the
  product name (nullable — a product-level row has none). Check what
  `getProductsByIds` already returns for variants before adding a second
  lookup; do not add a per-variant call if the product read already carries
  what's needed.
- `locationId` → `locationName`, via `ILocationService`
  (`libs/core/src/inventory/application/services/location.service.interface.ts`).
  **Note: this service only exposes `getLocation(id)` (single) and
  `listLocations(filters, pagination)` (no `ids` filter) — there is no batched
  by-id read today.** Either (a) add an `ids?: string[]` filter to
  `InventoryLocationFilters` for a real batched read, or (b) resolve the small
  number of *unique* location ids with bounded parallel `getLocation` calls
  (same-context call, cheap — do not do this per group). `null` (ADR-058
  decision 2, "master declines to locate") must stay `null`, never a
  fabricated name.
- `sourceConnectionId` → `connectionName`, via `ConnectionPort`
  (`@openlinker/core/identifier-mapping` — `inventory → identifier-mapping`
  is an existing, permitted cross-context edge, see
  `docs/architecture-overview.md § Cross-context dependencies in core`).
  **Same constraint as locations: `ConnectionPort` only exposes `get(id)`
  (single) and `list(filters)` (platformType/status only, no `ids` filter) —
  no batched by-id read today.** Same two options as above. `null` and
  `'legacy'` are BOTH the "not yet backfilled by #2317" sentinel (see
  `DuplicatePositionGroup`'s docblock) — neither is ever resolved to a name;
  only a real connection id gets a `connectionName`.

Update `DuplicatePositionGroup` (`libs/core/src/inventory/domain/types/inventory.types.ts`),
`DuplicatePositionGroupResponseDto`
(`apps/api/src/inventory/http/dto/duplicate-positions-response.dto.ts`), and
the controller mapping
(`apps/api/src/inventory/http/inventory.controller.ts:121-141`) to carry the
new nullable fields. Update `docs/operations/inventory-duplicate-positions.md`'s
example response.

## Classification

**Type**: CORE (Inventory)
**Layer**: Application (query service, cross-context composition) + Interface (DTO/controller)
**File(s)**:
- libs/core/src/inventory/application/services/inventory-query.service.ts
- libs/core/src/inventory/domain/types/inventory.types.ts
- libs/core/src/inventory/application/services/location.service.interface.ts (if extending the filter)
- libs/core/src/inventory/domain/types/location.types.ts (if extending `InventoryLocationFilters`)
- apps/api/src/inventory/http/dto/duplicate-positions-response.dto.ts
- apps/api/src/inventory/http/inventory.controller.ts
- docs/operations/inventory-duplicate-positions.md

## Dependencies

- None — independent of #3071's frontend sub-issues; they render raw ids
  today and can adopt the new fields once shipped without any change to them.

## Assumptions

- Neither `ILocationService` nor `ConnectionPort` has a batched by-id read
  today (confirmed by reading both interfaces) — this issue includes either
  adding one, or a bounded per-unique-id resolution. A per-*group* N+1 is not
  acceptable either way; a small, bounded loop over unique ids (typically far
  fewer than groups) is.
- New fields are additive/optional on the DTO — no breaking change for any
  existing caller (there is none in `apps/web` today).
- `locationId: null` and `sourceConnectionId: null | 'legacy'` are never
  resolved to a fabricated name — they stay their documented sentinel values.
- Resolving `inventory → identifier-mapping` for `ConnectionPort` does not
  introduce a new cross-context edge — it is already listed in the
  architecture doc's dependency map.

## Acceptance Criteria

- [ ] Response carries `productName`, `sku`, `connectionName` (nullable),
      `locationName` (nullable) per group
- [ ] Resolution is bounded to unique ids across the whole response, not
      per-group (asserted by an int-spec counting calls/queries, or by
      inspection of the implementation)
- [ ] `locationId: null` and `sourceConnectionId: null | 'legacy'` never
      resolve to a fabricated name
- [ ] `docs/operations/inventory-duplicate-positions.md` example response
      updated to match

jakubretajczykBD:
## Summary
- Adds `productName`, `sku`, `connectionName`, `locationName` to `GET /inventory/duplicate-positions`'s group response — an operator previously saw only raw internal ids (`ol_product_...`) with no way to tell which SKU or connection is at risk.
- Resolved in `InventoryQueryService.getDuplicatePositionReport`, batched per axis across the whole report's unique ids (never per group) — `IProductsService.getProductsByIds` for product/sku, `ILocationService` for locations, `ConnectionPort` for connections.
- All four fields are nullable with documented sentinel rules:
  - `productName`/`sku` → `null` only when the product could not be resolved (e.g. deleted).
  - `connectionName` → always `null` when `sourceConnectionId` is `null` or the #2317 `'legacy'` sentinel (not yet backfilled) — never a fabricated name.
  - `locationName` → always `null` when `locationId` is `null` (ADR-058 decision 2, the master declines to locate) — never a fabricated name.

## Context
Split out on purpose from epic #3071 (frontend diagnostic page), which deliberately scoped this out — its own stated assumption was "Backend contract is stable... no API changes needed". This is that confirmation/implementation.

## Test plan
- [ ] `pnpm test` (unit specs for `InventoryQueryService` cover the enrichment + null-sentinel behavior)
- [ ] `pnpm type-check`
- [ ] Manual: `GET /inventory/duplicate-positions` against a seeded catalogue with duplicate positions, confirm `productName`/`sku`/`connectionName`/`locationName` populate correctly and sentinel cases stay `null`

Closes #3239


piotrswierzy (approved):

**Verdict: ✅ Approve.** Display enrichment done through the owning contexts' service seams rather than a raw join, with the sentinel rules held correctly. All nine checks green.

### What it gets right

**It does not engage ADR-036, and that is the right outcome.** The obvious implementation of "add display names to a grouped report" is a `leftJoin` on raw table names — which `docs/code-review-guide.md` § Cross-context read-model joins would then subject to a four-point test this change would *fail*, since the amendment (#2025) permits projecting display columns only off a join a **filter** already justified, and there is no filter here. Instead `enrichDuplicatePositionGroups` (`inventory-query.service.ts:185+`) resolves through `IProductsService.getProductsByIds`, `ILocationService.getLocation` and `ConnectionPort.get`. No SQL-string coupling, nothing for a future rename to break silently.

**The sentinel discipline is exactly right and is the part most likely to have been got wrong.** `sourceConnectionId === null | LEGACY_SOURCE_CONNECTION_ID` and `locationId === null` are filtered out of the id sets *before* resolution and forced to `null` again at projection time (`inventory-query.service.ts:409-446`), so neither can acquire a fabricated name. That matches the architecture doc's rule that `'legacy'` is a **value naming one position whose owner is unknown**, never a wildcard, and ADR-058 decision 2's rule that `locationId IS NULL` means the master *declined* to locate. A `connectionName: "Legacy"` would have read as a real connection to an operator and is precisely what this avoids. The unit spec asserts both `connectionPort.get` and `locationService.getLocation` are **not called** for those states, which is stronger than asserting the output is null.

**Read-only throughout** — the repository sets the four fields to `null` and comments that enrichment happens in the service (`inventory.repository.ts:796`), the service only maps. Nothing writes, so #2319's detect-never-repair contract holds.

**The catch is narrow.** `buildConnectionNameMap` rethrows anything that is not `ConnectionNotFoundException` (`inventory-query.service.ts:462-466`), so a deleted connection degrades to a null name while a real fault still fails the report. A blanket `catch` there would have been the easy version and would hide infrastructure errors behind a diagnostic that reads healthy.

### Findings

**[SUGGESTION]** — `libs/core/src/inventory/application/services/inventory-query.service.ts:443-470`
> `buildLocationNameMap` and `buildConnectionNameMap` each fan out an unbounded `Promise.all` over their distinct id sets. The docblock argues the sets are "bounded by the number of DISTINCT locations/connections in the report, not by row or group count" — true, but the distinct count is itself bounded only by `maxGroups`, which caps at 500, so the worst case is ~1000 concurrent single-row reads against the pool. Realistic cardinality is small (an install has a handful of locations and connections), so this is not a live problem. Worth noting only because the repo elsewhere bounds exactly this shape rather than relying on cardinality — the adapter-declared `resolveBatchConcurrency` ceiling (ADR-047/#2229) and #2621's reconcile fan-out both cite the same reasoning. A small `p-limit`-style bound, or a batched `getLocationsByIds`, would make the guarantee structural rather than empirical.

**[SUGGESTION]** — `inventory-query.service.ts:78-82`
> Injecting `ConnectionPort` straight into an inventory-context service is permitted by the cross-context contract (`*Port` with a single `Port` suffix is on the allow list, and `check-cross-context-imports.mjs` passes), so this is not a violation. But the architecture doc's note on the closest precedent reads that `RateLimitStatusService` / `WebhookStatusService` "inject `ConnectionPort` directly (which they may, **living inside the module that provides it**)" — which this service does not. If an `IConnectionService` read exists or lands later, it is the more conventional seam for a cross-context name lookup. Not worth churn on its own.

**[Positive]** — the test renamed from `returns the repository report verbatim` to `preserves the repository totals verbatim … while enriching groups`, and correctly changed `resolves.toBe(truncated)` (identity) to a field-wise assertion. A reviewer skimming would have seen a weakened assertion; it is actually the correct adjustment, since enrichment necessarily returns a new object, and `groupCount` — the #2325 gate — is still pinned.

### Merge readiness

✅ **Approve.** Both suggestions are optional and neither affects correctness. The layering, the sentinel handling and the failure mode are all right.

piotrswierzy (approved):
# Tech Lead review — #3249 (duplicate-positions display enrichment)

Reviewed against `docs/code-review-guide.md`, `docs/engineering-standards.md`,
`docs/architecture-overview.md` § Inventory (ADR-058 ladder, #2319/#2317) and
`docs/operations/inventory-duplicate-positions.md`.

## Findings

**[IMPORTANT]** — `apps/api/src/inventory/http/dto/duplicate-positions-response.dto.ts:60-75`, `libs/core/src/inventory/domain/types/inventory.types.ts:420-435 (docblock at 426, fields at 433-434)`, `docs/operations/inventory-duplicate-positions.md:107`

> All three places say `sku` is `null` *"when the product could not be resolved"*. That is not
> what the code produces. `Product.sku` is `string | null`
> (`libs/core/src/products/domain/entities/product.entity.ts:23`), and the enrichment writes
> `sku: product?.sku ?? null` — so a **fully resolved** product that simply has no SKU also
> reports `null`, indistinguishably. An operator following the documented reading concludes the
> product row is gone (and per the ops doc's remediation, that the group is a candidate for
> "delete the whole group if the product is genuinely gone at the master"), which is the wrong
> action for a live SKU-less product.
>
> The same context already models this correctly: `InventoryViewProduct`
> (`libs/core/src/inventory/application/types/inventory-view.types.ts:19-23`) keeps
> `sku: string | null` **inside** a separately-nullable `product`, so "no product" and "product
> with no SKU" stay distinct. Either correct the copy on all three surfaces
> (`productName` alone carries the resolution signal, since `Product.name` is non-null), or carry
> the resolution fact explicitly.

**[IMPORTANT]** — `libs/core/src/inventory/application/services/inventory-query.service.ts:238-262`

> `buildLocationNameMap` and `buildConnectionNameMap` issue an unbounded `Promise.all` over the
> id sets. The docblock justifies this with *"the id sets here are bounded by the number of
> DISTINCT locations/connections in the report, not by row or group count"* — but distinct
> locations are not bounded below group count: `maxGroups` is capped at 500
> (`MAX_DUPLICATE_POSITION_GROUPS`), and 500 groups at 500 different locations is a legal report,
> so the worst case is 500 concurrent single-row reads fired in one tick. That lands on an
> endpoint `docs/operations/inventory-duplicate-positions.md` already characterises as *"two full
> sequential scans of `inventory_items` with no supporting index… run it off-peak on a large
> catalogue"*, so it compounds a read the docs already treat as expensive.
>
> The repo's established answer to an in-flight ceiling is a declared, clamped one
> (ADR-047 / #2229, `resolveBatchConcurrency` — and the rule there is that the number applied is
> the number reported). Either bound the map with that shape, or give `ILocationService` an
> id-set read — `InventoryLocationFilters`
> (`libs/core/src/inventory/domain/types/location.types.ts:83-89`) has no `ids` arm today, which
> is why the per-id loop was reached for. The justification comment should change with the code;
> as written it reads as a safety argument that does not hold.

**[IMPORTANT]** — missing integration coverage

> `apps/api/test/integration/inventory-duplicate-positions.int-spec.ts` already exists, already
> drives **both** the service seam and the HTTP route (`describe('GET /inventory/duplicate-positions')`,
> line 451), and is not touched by this PR. The change joins three foreign reads — products,
> `inventory_locations`, `connections` — into that response; per `docs/testing-guide.md` that is
> the vertical slice `*.int-spec.ts` exists for, and it is the only place the two null-sentinel
> rules can be asserted against real rows rather than against a mock's return value. Worth at
> least: a group whose `sourceConnectionId` is a **real** connection resolving to its name, and
> one on the `'legacy'` sentinel staying `null`.

**[SUGGESTION]** — `libs/core/src/inventory/domain/types/inventory.types.ts:420-435 (docblock at 426, fields at 433-434)` and `libs/core/src/inventory/infrastructure/persistence/repositories/inventory.repository.ts:796-800`

> The four display fields are **required** on `DuplicatePositionGroup`, so the repository now has
> to stub them — and does, four lines of `null` — directly under a docblock that says *"The
> repository never sets these"*. The statement and the code disagree in the same change.
> The context already has the shape that avoids this: `InventoryItemView` keeps the raw entity
> and the joined display data in separate types, so persistence never learns a display concern
> exists. A `DuplicatePositionGroupView` (or optional fields) would remove both the stub and the
> contradiction.

### Positive observations

- The batched product read reuses the existing `buildProductMap`, and the spec asserts *one* call
  per unique id across two groups sharing a product — the assertion actually pins the batching
  rather than just exercising it.
- `ConnectionNotFoundException` is caught narrowly and everything else rethrown
  (`inventory-query.service.ts:249-262`), so a deleted connection degrades to `null` while a real
  infrastructure fault still fails the report. That is the right asymmetry.
- Enrichment sits in the application service — not the repository, not the controller — and the
  controller maps it through verbatim with a spec asserting exactly that. Layering per
  `docs/engineering-standards.md` § Hexagonal is clean, and `ConnectionPort` is an allowed
  cross-context symbol (`*Port`, not `*RepositoryPort`) with abundant precedent in
  `listings`/`customers`; the read-only use also respects the architecture overview's
  "writes go through `IConnectionService`, never `ConnectionPort`".
- `LEGACY_SOURCE_CONNECTION_ID` is imported rather than re-spelled, and the `'legacy'` sentinel is
  excluded *before* the fan-out, not filtered after — so no wasted lookup and no path to a
  fabricated name.

---

## Summary

Adds four display fields to the `GET /inventory/duplicate-positions` group shape, resolved in
`InventoryQueryService` and batched per axis. The layering, the sentinel handling and the
controller pass-through are right, and the unit specs are specific rather than ceremonial. The
most important concern is that the documented meaning of `sku: null` is not the meaning the code
produces, on a surface whose whole purpose is to tell an operator which row to delete.

## Merge Readiness

🔄 **Approve with changes** — no BLOCKING issues; merge after the IMPORTANT items are addressed
(the `sku` copy fix is a one-liner and should not wait).

## Priority fixes

1. Correct the `sku`-is-null documentation on the DTO, the core type and the ops doc — or carry the
   product-resolution fact explicitly.
2. Bound the location/connection fan-out, and correct the docblock that currently justifies it with
   a bound that does not hold.
3. Extend `inventory-duplicate-positions.int-spec.ts` to cover the enrichment and both null
   sentinels against real rows.
4. (Optional) Split the raw group from its enriched view so the repository stops stubbing display
   fields.

piotrswierzy (approved):
## Re-review (after force-push `1cf94ef`)

Re-read against `main` as it stands today (post CI-perf merge). The enrichment moved out of the repository's three-way join and into `InventoryQueryService` — a better layer for it, and it makes the batching argument legible.

### 1. Prior findings

|---|---|---|---|
| 1 | `sku: null` documented as "product could not be resolved", but `Product.sku` is `string \| null` so a live SKU-less product is indistinguishable from a deleted one | **Addressed** | `libs/core/src/inventory/domain/types/inventory.types.ts` — the `DuplicatePositionGroup` docblock now states `productName` alone carries the resolution signal and that `sku: null` "does NOT by itself mean 'product not found'". Mirrored on the wire (`apps/api/src/inventory/http/dto/duplicate-positions-response.dto.ts`, `sku` description) and in `docs/operations/inventory-duplicate-positions.md`. Documented rather than re-shaped, which is the right call — the ambiguity is `Product.sku`'s, not this endpoint's. |
| 2 | Unbounded `Promise.all` over up to `maxGroups` = 500 single-row reads | **Addressed** | `inventory-query.service.ts` — `runWithConcurrency` + `DUPLICATE_POSITION_NAME_LOOKUP_CONCURRENCY = 10`, applied to both the location and connection axes, with the docblock naming `resolveBatchConcurrency`/ADR-047 (#2229) as the precedent and stating explicitly that distinct-id count is *not* bounded below group count. `buildProductMap` already de-duplicates (`[...new Set(productIds)]`), so the product axis stays one call. |
| 3 | `inventory-duplicate-positions.int-spec.ts` untouched | **Addressed** | New case "resolves connectionName for a real connection, and leaves it null on the #2317 legacy sentinel (#3239)" — real `connections` row via `createTestConnection`, real `'legacy'` sentinel row, asserting the resolved name and the withheld one against a live database. That is the half a mock cannot prove. |

### 2. New commits

Wiring checks all pass: `LOCATION_SERVICE_TOKEN` is provided and exported by `InventoryModule` (`inventory.module.ts:143`, `:233`); `CONNECTION_PORT_TOKEN` is an established cross-context import for this context (`availability.service.ts:63`, and `inventory.module.ts:85` already imports `IdentifierMappingModule` for it) and `ConnectionPort` is a capability port, not a `*RepositoryPort`, so `scripts/check-cross-context-imports.mjs` is satisfied. The only other construction site (`inventory-query.service.spec.ts:115`) is updated.

Enrichment correctly runs **after** truncation, so a capped report never resolves names for groups it will not return.

**[SUGGESTION]** — `libs/core/src/inventory/application/services/inventory-query.service.ts` (`runWithConcurrency`)
> The helper is wave-based, not a sliding window: `await Promise.all(slice)` per chunk, so one slow lookup stalls the whole wave rather than letting the next id start. At a cap of 10 over at most 500 ids that is a bounded cost and fine for an off-peak diagnostic — noting it only so a future copy of this helper onto a hot path is a deliberate choice.

**[SUGGESTION]** — `apps/api/test/integration/inventory-duplicate-positions.int-spec.ts`
> The new case asserts through `queryService.getDuplicatePositionReport()` directly, while the file already has an HTTP block (line ~453, `GET /v1/inventory/duplicate-positions`). The four new DTO fields therefore reach the wire only in a mock-backed controller unit test. The mapping is a verbatim passthrough and *is* unit-tested, so this is not a gap in behaviour — but adding `productName`/`connectionName` to the existing HTTP assertion would close the transport half for the cost of two lines.

### 3. Summary

The three prior IMPORTANT findings are genuinely addressed in code, not in a reply. The `sku` ambiguity is now stated identically in the domain type, the DTO and the ops doc; the fan-out is capped with a named precedent; the int-spec proves the `'legacy'` sentinel is never resolved to a fabricated name against real Postgres. Moving enrichment from the repository to the service also keeps the repository honest — it now explicitly writes `null` for all four fields with the comment "resolved by InventoryQueryService, never here".

### Merge Readiness

✅ **Approve** — no BLOCKING items, prior or new. The two SUGGESTIONs are optional.

### Priority fixes

None required.

piotrswierzy (approved):
## Tech Lead review — `feat(inventory): enrich duplicate-positions response with display names`

Base of the duplicate-positions stack. Enrichment placed in the right layer, with the two sentinel states protected from fabricated names.

### ✅ The sentinels are never resolved, and the null semantics are documented where they're read

The two states that must not acquire a name are excluded **before** the lookup, not filtered after:

```ts
.filter((id): id is string => id !== null && id !== LEGACY_SOURCE_CONNECTION_ID)
```

- `sourceConnectionId === null | 'legacy'` — not yet covered by the #2317 backfill. `'legacy'` is a **value naming one position whose owner is unknown**, so resolving it to anything would be an invented claim about provenance.
- `locationId === null` — ADR-058 decision 2, the master *declines* to locate its stock. OpenLinker never invents a location the master withheld, and that applies to a display name as much as to the column.

Using the exported `LEGACY_SOURCE_CONNECTION_ID` constant rather than a `'legacy'` literal is right — that constant is declared once and pointed at by both the ORM column comment and the #2314 migration.

### ✅ `sku: null` is two things, and all three surfaces say so

`Product.name` is non-null while `Product.sku` isn't, so `sku: null` means *either* "product unresolved" *or* "resolved product with no SKU", and `productName` is the discriminator. That's stated in the domain type, in the `@ApiProperty` description, and in the ops doc. Three places is the right number here — a consumer reading only the OpenAPI doc gets the caveat too.

### ✅ Layering

Repository sets all four to `null` with a comment saying enrichment is the service's job; the service resolves; the controller passes through verbatim, with a spec assertion pinning that it *never resolves names itself*. Clean separation, and the controller spec's comment makes the intent legible to whoever touches it next.

Batched via `buildProductMap` — the existing helper, which already dedupes internally, so the un-`Set`-ed `groups.map(g => g.productId)` at the call site is correct rather than an oversight. The `getEarliestOrderDateByConnection` (#2083) precedent it cites is the right one: one batched read ahead of any per-row loop.

On the `ConnectionPort` injection — I checked before flagging it. `IConnectionService` lives in `apps/api`, not `libs/core`, so a core service has no `I*Service` route to a connection; `CONNECTION_PORT_TOKEN` is the established cross-context path (seven core contexts already inject it, `inventory` among them). Consistent with the house pattern, not a new deviation.

### ✅ Failure isolation

`ConnectionNotFoundException` → `null` name, everything else rethrown. A deleted connection is an ordinary state for a duplicate-positions report — those rows are exactly the ones whose owning integration may be long gone — so degrading to the raw id rather than failing the whole diagnostic is right. Narrow catch, tested.

### 🟢 SUGGESTIONS

1. **`runWithConcurrency` is wave-based, not a pool.** Each slice runs under `Promise.all` and the next starts only once the previous fully settles, so one slow lookup stalls its whole wave. At a cap of 10 over at most a few hundred ids that's fine, and the simplicity is worth something — but the docblock calls it a "ceiling", which a reader may take to mean sustained in-flight concurrency. Worth one clause saying it's waves.

2. **The real fix is batched reads, and the docblock already knows it.** `ILocationService.getLocation` and `ConnectionPort.get` are both single-id, so a worst-case report still costs up to `MAX_DUPLICATE_POSITION_GROUPS` round trips per axis. A `getLocationsByIds` / `listByIds` would collapse both to one query each and let the concurrency helper go away. Worth filing rather than carrying the cap indefinitely — especially since this read is the documented gate procedure for #2325, so it runs at exactly the moment an install has the most duplicates.

3. **`runWithConcurrency` is a private module function in a `*.service.ts`.** Fine as-is, but if a second caller ever appears it belongs beside `resolveBatchConcurrency` rather than copied.

### Notes

- The int-spec asserting `connectionName` resolves for a real connection **and** stays null on the `'legacy'` sentinel, in one test against real Postgres, is the right shape — the two branches only mean something next to each other.
- Ops doc updated in the same PR with inline comments on the sample payload explaining each null case. Good; that file is the one an operator actually has open while running this.
- CI: nine green, `Test` still queued at review time. Please don't merge until it lands.

**Verdict:** ✅ Approve.

Issue #2079, opened 2026-08-14
[SECURITY] Auth — RolesGuard default-allows undecorated routes; audit and decorate before the packer role exists


piotrswierzy:
> **Superseded recommendation.** This issue originally recommended option (1) — a spec plus a reviewed allowlist. **That is no longer sufficient**, and an implementer must not build it. See *Why the allowlist was rejected* below. Wave 3b's spec § 1.6 makes this **audit *and decorate***.

## Problem / Context

`RolesGuard.canActivate` (`apps/api/src/auth/guards/roles.guard.ts:28`):

```ts
if (!requiredRoles || requiredRoles.length === 0) return true;
```

It is registered as the second global `APP_GUARD` after `JwtAuthGuard` (`apps/api/src/auth/auth.module.ts:84-85`). So **any route without an explicit `@Roles()` decorator authorizes any authenticated principal.**

Measured on `main` at `0470542e0`: **~280 route decorators, ~121 without `@Roles`**, ~110 of those authenticated. The figure drifts as routes land, so the proportion is what matters, not the integer. The undecorated set includes `customers.controller.ts` (`@Get()`, `@Get(':id')` — **buyer PII**), all of `products`, and the `sales-documents` rules surface **including writes**.

**The existing `write-guard-coverage.spec.ts` cannot catch this.** It covers **non-GET handlers only**, on a hand-listed 23-controller set — so every PII *read* is outside it by construction. It is not a weaker version of the check this issue needs; it is a check of a different thing.

**It was not a live vulnerability while every principal was an OL user with a role** — the fail-open default then reduced to "logged-in users can call these routes", which was intended. MCP correctly avoids the trap: `mcp-transport.controller.ts` is `@Public()` with its own `OAuthTokenVerifier`, populating `req.auth`, which no guard reads.

**Wave 3b arms it.** [ADR-071](../blob/main/docs/architecture/adrs/071-pack-station-principal.md) decides that packers are ordinary users on a narrower `packer` role, and that the bench is a device label rather than a principal. **A narrower role means nothing while an undecorated route admits any authenticated principal** — a `packer` session would reach buyer PII on day one. That is what promotes this from tech-debt to a Wave 3b prerequisite.

## Why the allowlist was rejected

[...]

PR #3250, opened 2026-09-15
feat(inventory): expose #2317 provenance-backfill completion state for the duplicate-positions readiness gate

### LINKED ISSUE #3240: [TASK] CORE — expose #2317 provenance-backfill completion state for the duplicate-positions readiness gate
## Problem / Context

Readiness for the #2325 stricter `SET NOT NULL` + unique-index migration
depends on **two independent conditions**, not one:

1. `groupCount: 0` on a `GET /inventory/duplicate-positions` scan (#2319,
   already shipped).
2. The #2317 provenance backfill having fully drained
   (`countMissingProvenance() === 0`).

Both `DuplicatePositionGroup`'s and `countMissingProvenance()`'s own docblocks
state the ordering dependency explicitly: `NULL` and `'legacy'`
`sourceConnectionId` values are grouped together as "not yet backfilled", and
a scan taken while the backfill is still mid-run can read `groupCount: 0` and
then flip non-zero once NULL rows collapse to `'legacy'` and reveal a
collision the scan couldn't see yet. The reviewed frontend mockup for epic
#3071 (PR #3121, `docs/plans/mockups/duplicate-positions-3071.html`) models
this as a two-item readiness checklist (`clean-pending` vs `clean-ready`
states) — but that second condition has **no way to be read today**.

Confirmed by reading the code: `InventoryRepositoryPort.countMissingProvenance()`
(`libs/core/src/inventory/domain/ports/inventory-repository.port.ts:349`) is
called only internally by `InventoryProvenanceBackfillService.runPage()`
(`libs/core/src/inventory/application/services/inventory-provenance-backfill.service.ts:44`)
on each sweep tick, and the result is **never persisted** — no cursor row, no
`completedAt`, nothing in `connection_cursors`. This is a *deliberate* design
choice per the payload docblock in
`libs/core/src/sync/domain/types/inventory-job-payloads.types.ts:16-28`:
"There is no offset and no cycle id... remaining work is re-derived from the
table on every tick rather than tracked on a cursor." So the answer must stay
a **live** read — this issue is not about adding a persisted flag, it's about
exposing the existing live count outside the worker process.

Epic #3071's sub-issues (#3072–#3077) do not reference this at all — its
readiness banner (#3073) checks only `groupCount === 0`. This is a
deliberate, separate gap, not an oversight in that epic.

## Proposed Solution

Expose `countMissingProvenance()` — already a cheap, single, unfiltered
`SELECT COUNT(*) FROM inventory_items WHERE sourceConnectionId IS NULL`
(`libs/core/src/inventory/infrastructure/persistence/repositories/inventory.repository.ts:1031`) —
through a new read on `IInventoryQueryService`, e.g.:

```ts
getProvenanceBackfillStatus(): Promise<{ remainingNull: number; completed: boolean }>
```

Decide, and record the decision in the PR, whether to:

- **(a) fold it into the existing `GET /inventory/duplicate-positions`
  response** as an additional top-level field (e.g. `provenanceBackfill:
  { remainingNull, completed }`) alongside `groupCount`/`rowCount` — cheapest
  for a caller (one request, both readiness conditions), acceptable given the
  count is a single indexed-or-not `COUNT(*)` and the endpoint already runs
  two full scans, or
- **(b) a standalone `GET /inventory/provenance-backfill-status` read** if
  (a) risks compounding the existing 504-timeout profile documented for this
  endpoint on a very large catalogue (see the mockup's `error` state and
  `docs/operations/inventory-duplicate-positions.md`).

Whichever shape is chosen: this stays a **live** read, never a stored
completion flag — do not introduce a cursor or persisted "done" bit, since
that would contradict the backfill's existing re-derive-every-tick design and
could go stale if a later mutation reintroduces `NULL` rows.

## Classification

**Type**: CORE (Inventory)
**Layer**: Application (query service, reuses existing repository method) + Interface (DTO/controller)
**File(s)**:
- libs/core/src/inventory/application/services/inventory-query.service.ts
- libs/core/src/inventory/application/services/inventory-query.service.interface.ts
- apps/api/src/inventory/http/inventory.controller.ts
- apps/api/src/inventory/http/dto/duplicate-positions-response.dto.ts (if folding into (a))
- docs/operations/inventory-duplicate-positions.md

## Dependencies

- None — independent of #3071's frontend sub-issues and of the
  duplicate-positions enrichment issue. #3073 (readiness KPI strip) can
  adopt this once shipped; today it checks `groupCount === 0` only.

## Assumptions

- `countMissingProvenance()`'s query is cheap enough to run per request as-is
  (single `COUNT(*)` with a simple `IS NULL` predicate) — verify against a
  large catalogue before deciding (a) vs (b) above; if it turns out to need an
  index or is non-trivial at scale, that's a prerequisite sub-task, not a
  reason to skip this issue.
- No new persisted cursor/completion flag is introduced.
- The eventual frontend consumer (a follow-up to #3071, not in scope here) is
  what turns this into the two-condition readiness banner the mockup shows —
  this issue is backend-only.

## Acceptance Criteria

- [ ] An operator-reachable read reports whether the #2317 backfill has
      fully drained (`remainingNull === 0`) at request time
- [ ] The read is live/derived on each call, never a stale persisted flag
- [ ] Decision recorded in the PR description (fold into
      `duplicate-positions` vs. standalone endpoint) with the cost
      trade-off stated
- [ ] `docs/operations/inventory-duplicate-positions.md` updated to
      describe the new field/endpoint and the two-condition readiness rule

jakubretajczykBD:
## Summary
- Adds `GET /inventory/provenance-backfill-status` (admin-only) returning `{ remainingNull, completed }` — the second, independent readiness condition for the #2325 stricter uniqueness index, alongside `GET /inventory/duplicate-positions`'s `groupCount`.
- Readiness for #2325 depends on BOTH: `groupCount === 0` AND the #2317 provenance backfill having fully drained. A scan taken mid-backfill can read `groupCount: 0` and then flip non-zero once NULL rows collapse to `'legacy'` and reveal a collision the scan couldn't see yet.
- Standalone endpoint (not folded into the duplicate-positions response) to avoid compounding that endpoint's existing timeout profile on a very large catalogue.
- Resolved LIVE on every call, never cached — the #2317 backfill itself tracks no cursor, so a stored answer could go stale the moment a later mutation reintroduces a NULL row.

## Context
Split out on purpose from epic #3071. Confirmed by reading the code: `InventoryRepositoryPort.countMissingProvenance()` was called only internally by the backfill sweep, with its result never persisted or exposed outside the worker process — this issue exposes the existing live count over HTTP.

⚠️ **Stacked PR** — base branch is `3239-enrich-duplicate-positions` (#3249), not `main`. Merge #3249 first.

## Test plan
- [ ] `pnpm test` (unit specs cover `getProvenanceBackfillStatus`)
- [ ] `pnpm type-check`
- [ ] Manual: `GET /inventory/provenance-backfill-status` returns `remainingNull`/`completed` matching a live `COUNT(*) FROM inventory_items WHERE "sourceConnectionId" IS NULL`

Closes #3240


piotrswierzy (approved):

**Verdict: 🔄 Approve with changes.** The endpoint is correct and answers the right question live, which is what #2325 needs. But the *stated reason* it reads live is factually wrong, repeated in three places — and the thing that is actually true instead opens an operator-facing gap this endpoint is the natural place to close.

### What it gets right

`completed: remainingNull === 0` (`inventory-query.service.ts:186-193`) is exactly the rule the architecture doc insists on: completion is `remainingNull === 0` **alone**, never "the page stamped nothing" — a page can legitimately stamp zero rows while work remains, when every candidate was locked and skipped, and latching on that would switch the pass off permanently. This gets it right.

Reading live rather than from the cursor is also the right call, and matches the doc's own instruction that `remainingNull` "is honest about its own staleness … so #2325 re-counts before it acts." A standalone endpoint rather than folding it into the duplicate-positions response is well argued too — that endpoint already carries a timeout profile on a large catalogue.

### Findings

**[IMPORTANT]** — `apps/api/src/inventory/http/dto/provenance-backfill-status-response.dto.ts:18`, `libs/core/src/inventory/domain/types/inventory.types.ts:475-479`, `inventory-query.service.ts:187-190`, and the PR body
> All four assert that "the backfill itself tracks no cursor" / "has no cursor to read a cached answer from". **It does.** `InventoryProvenanceBackfillHandler` writes two `connection_cursors` keys under the nil-UUID system connection — `sweepRemainingCountCursorKey` (`inventory-provenance-backfill.handler.ts:179-184`) and `sweepCompletedAtCursorKey` (`:190-196`) — and the architecture doc names both as "#2325's readiness artefact".
>
> I suspect the source of the slip is the handler's own comment at `:211` — *"there is no cursor to hold"* — which is scoped to a **scan offset** (the pass is frontier-as-query; `runBoundedSweep` is deliberately not called). Generalising that to "no cursor at all" is the drift. The decision to read live stays correct either way, so this is a comment fix, not a code fix — but a docblock that misdescribes the system is how the *next* change gets made on a false premise.

**[IMPORTANT]** — `inventory-query.service.ts:186-193`, and the readiness story generally
> The backfill **self-latches and never re-counts once latched.** `execute()` reads `completedAt` before the lock and before any table access, and returns `{outcome: 'ok'}` immediately if it is set (`inventory-provenance-backfill.handler.ts:144-150`). The doc's escape hatch is manual: "Deleting that cursor row re-arms the drain."
>
> So the exact scenario this PR's own docblock invokes to justify reading live — *"a later mutation reintroduces a NULL row"* — produces a state this endpoint reports but cannot explain: `remainingNull: N, completed: false`, **forever**, with the pass latched off and nothing draining it. An operator watching the readiness gate for a `SET NOT NULL` migration sees a number that will never reach zero and no indication why.
>
> This endpoint is the right place to close that, and cheaply: read `sweepCompletedAtCursorKey` alongside the live count and surface whether the pass is latched, so *"N remaining, still draining"* and *"N remaining, pass latched — re-arm it"* are distinguishable. Without it the second state is indistinguishable from the first at every surface built on this endpoint — which #3252 then renders as the positive sentence "Provenance backfill still running", a claim that is false in exactly that case.

**[SUGGESTION]** — `apps/api/src/inventory/http/inventory.controller.ts:158-175`
> `@Roles('admin')` matches the sibling duplicate-positions route and `route-authorization-coverage.spec.ts` will hold it. Worth confirming the new route is picked up by that spec's controller discovery (it discovers rather than lists, so it should be automatic) — no action expected, just the one guard worth not assuming.

**[Positive]** — extending the `IInventoryQueryService` mock in `products.controller.spec.ts` rather than casting around the widened interface keeps the full-shape discipline that file's existing comment records. Easy to have reached for an `as unknown as` instead.

### Merge readiness

🔄 **Approve with changes.** The endpoint returns the right answer today and nothing here is wrong in a way that misleads a caller *about the count*. Fix before merge:

1. Correct the "tracks no cursor" claim in the DTO description, the `ProvenanceBackfillStatus` docblock, the service comment and the PR body — the pass writes two `connection_cursors` keys; what it does not have is a scan offset.
2. Decide on the latch. Either surface it here (preferred — it is one extra cursor read and it makes the readiness gate honest), or state the gap explicitly in the docblock so the surfaces built on top do not assert "still running" about a pass that has stopped.

piotrswierzy (changes requested):
# Tech Lead review — #3250 (provenance-backfill readiness status)

Reviewed against `docs/code-review-guide.md`, `docs/engineering-standards.md`,
`docs/architecture-overview.md` § Inventory (#2317 / #2319 / ADR-058 ladder step (iii)),
`docs/operations/inventory-duplicate-positions.md` and `docs/testing-guide.md`.

Stacked on #3249 — reviewed as the incremental diff.

## Findings

**[BLOCKING]** — `libs/core/src/inventory/domain/types/inventory.types.ts` (the `ProvenanceBackfillStatus` docblock)

> The docblock states, in bold:
>
> > **`completed` reflects the persisted LATCH, not `remainingNull === 0`.**
>
> The shipped code is the opposite. `InventoryQueryService.getProvenanceBackfillStatus` returns
> `completed: remainingNull === 0`; the field's own JSDoc one line below says `` `remainingNull === 0` ``;
> `ProvenanceBackfillStatusResponseDto` says `'remainingNull === 0'`; and the unit specs assert
> exactly that (`{ remainingNull: 3, completed: false, latchedAt: '2026-08-01…' }` — a latched
> pass reporting `completed: false`).
>
> The paragraph is also self-contradicting: if `completed` really were the latch, then
> `completed: true` alongside `remainingNull > 0` would *be* the "latched, and stuck" signal, which
> directly contradicts its own next sentence (*"`completed` alone cannot tell 'still draining'
> apart from 'latched, and stuck'"*). Only one of the two readings can be true, and the code
> implements the second.
>
> This is BLOCKING rather than cosmetic because of **which** paragraph it is: this type is the
> readiness gate an implementer of #2325 consults before running a `SET NOT NULL` +
> `CREATE UNIQUE INDEX` migration, and `docs/operations/inventory-duplicate-positions.md` § Wave 1d
> gate is explicit that getting the gate wrong *"is not a warning — it is the migration failing at
> deploy time"*. A reader who believes the bold sentence will read `completed: false` on a fully
> drained table as "the pass never latched, do not proceed", or `completed: true` as a durable
> latch. Delete or rewrite the paragraph to match the code.

**[IMPORTANT]** — `apps/api/src/inventory/http/dto/provenance-backfill-status-response.dto.ts` (`completed` description)

> `remainingNull`'s description carries the live-read honesty at length; `completed`'s is the bare
> string `'remainingNull === 0'`. `docs/architecture-overview.md` § Inventory (#2317) is explicit
> that this number *"is honest about its own staleness — a caller with no connection axis can
> insert a fresh provenance-less row a moment later, so #2325 re-counts before it acts."*
>
> A boolean named `completed` does not inherit that honesty — it reads as a settled fact, and the
> published OpenAPI description is the only thing telling a consumer otherwise. The FE client in
> #3251 already surfaces this field, and a page rendering "Backfill complete ✓" from it would be
> making a guarantee the value cannot support. Carry the same caveat on `completed`, explicitly
> including "#2325 must re-count immediately before it acts" — i.e. this is a point-in-time
> observation, never a standing guarantee.

**[IMPORTANT]** — `libs/core/src/inventory/application/services/inventory-query.service.ts` (`latchedAt` passthrough)

> `ISyncCursorsService.getCursor` returns `string | null`
> (`libs/core/src/sync/application/services/sync-cursors.service.interface.ts:18`), and the writer
> treats an **empty string** as *not latched*:
> `apps/worker/src/sync/handlers/inventory-provenance-backfill.handler.ts:146` —
> `if (completedAt !== null && completedAt.length > 0)`.
>
> This reader passes the raw value through, so a `''` cursor row surfaces as `latchedAt: ''`, which
> is non-null. Every consumer the docblock instructs (`remainingNull > 0 && latchedAt !== null`
> ⇒ "re-arm the pass") then reports the stuck state for a pass that has never latched and is
> draining normally — and the prescribed remedy is deleting a cursor row. Normalise with the
> writer's own predicate (`latchedAt: completedAt && completedAt.length > 0 ? completedAt : null`).
> Reader and writer disagreeing about what the stored value means is the failure mode this field
> exists to prevent.

**[IMPORTANT]** — `libs/core/src/inventory/application/services/__tests__/inventory-query.service.spec.ts` (the cursor-key spec cannot fail in the way that matters)

> ```ts
> expect(cursors.getCursor).toHaveBeenCalledWith(
>   '00000000-0000-0000-0000-000000000000',
>   expect.stringContaining('inventory-provenance'),
> );
> ```
>
> All three keys in this namespace contain `inventory-provenance`
> (`master.inventory-provenance.sweep:…`, `…remainingNull:…`, `…completedAt:…`), so this assertion
> passes for **exactly the two wrong keys** a reader could plausibly pick — including
> `remainingNull`, which the type's own docblock warns is written for observability and must never
> be read back to answer this call. The test named for pinning the key is the one test that cannot
> catch the key being wrong.
>
> `libs/core/src/sync/domain/types/master-sweep-cursor.types.spec.ts:60-67` already establishes the
> right shape, with its own reason attached: *"Hard-coded independently of the builders… A format
> change must fail here rather than silently split reader from writer."* Assert the literal
> `master.inventory-provenance.completedAt:connection:00000000-0000-0000-0000-000000000000`.

**[IMPORTANT]** — `libs/core/src/inventory/application/services/inventory-query.service.ts` (a fourth hardcoded nil-UUID, outside the mirror check)

> `PROVENANCE_BACKFILL_SYSTEM_CONNECTION_ID = '00000000-0000-0000-0000-000000000000'` is a new,
> independently-hardcoded copy of the system connection id.
> `scripts/check-system-connection-id-mirror.mjs` exists (#2745) precisely so these copies cannot
> drift, and it matches on an explicit `{file, name}` list of three entries
> (`inventory.service.ts`, `master-deletion-to-job.handler.ts`, `connections.types.ts`) with
> `name: 'SYSTEM_CONNECTION_ID'`. This copy is registered in neither, so it is the one instance the
> invariant does not cover — and per `docs/architecture-overview.md` § Sync Manager the whole point
> of that script is that *"the frontend special-case would silently stop firing"* if a copy drifted.
>
> Note also that the accompanying comment — *"declared locally … since neither is exported from a
> shared module"* — is inaccurate on both halves: the key builder **is** exported and **is**
> imported here (`masterSweepCompletedAtCursorKey` from `@openlinker/core/sync`), and the handler
> exports its own scope id (`SYSTEM_CONNECTION_ID as INVENTORY_PROVENANCE_SCOPE_ID`,
> `inventory-provenance-backfill.handler.ts:280`). Core cannot import from `apps/worker`, so a local
> declaration is the right call — but it needs to be registered in the mirror list (renamed to
> `SYSTEM_CONNECTION_ID`), or better, promoted from the private field it already occupies in
> `inventory.service.ts:28` and reused in-context.

**[IMPORTANT]** — missing authz integration coverage for the new admin-only route

> `apps/api/test/integration/viewer-role-authz.int-spec.ts:305` carries the viewer-403 assertion for
> `GET /inventory/duplicate-positions`, and the sibling int-spec's header records that this is the
> convention, by name: *"the viewer-403 half lives with the other HTTP authz assertions in
> `viewer-role-authz.int-spec.ts`"*. The new `@Roles('admin')` route has no entry there.
>
> `apps/api/src/auth/route-authorization-coverage.spec.ts` does not substitute: per
> `docs/engineering-standards.md` § Route authorization it proves a decorator is *present*, never
> which roles it names. There is also no int-spec for the endpoint itself — the live
> `COUNT(*) … WHERE "sourceConnectionId" IS NULL` and the cursor read are both real-Postgres
> behaviour, and the PR's own test plan lists that verification as manual.

### Positive observations

- `@Roles('admin')` is correct and matches the sibling diagnostic; the controller is a pure
  pass-through with no logic, and the two reads are issued under one `Promise.all`.
- Keeping this a **standalone** endpoint rather than folding it into the duplicate-positions
  response is the right call and the PR body argues it correctly — that endpoint is already two
  full sequential scans per `docs/operations/inventory-duplicate-positions.md`.
- `latchedAt` is a genuine addition beyond the issue's ask, and it is the right one: without it
  "still draining" and "latched, and stuck" are indistinguishable, and the stuck state is the one
  with an operator action attached. The finding above is about its edge case, not its existence.
- The key is built through the shared `masterSweepCompletedAtCursorKey` rather than string-concatenated,
  and the sweep kind is typed `MasterSweepKind`, so a typo there is a compile error.
- `countMissingProvenance` already existed on `InventoryRepositoryPort` — no port widening, and the
  read is reached through `IInventoryQueryService`, not a cross-context repository port.

---

## Summary

Exposes the #2317 backfill's live `remainingNull` over an admin-only HTTP route as the second
readiness condition for #2325, plus a `latchedAt` stamp that usefully separates "draining" from
"latched and stuck". The route, the layering and the standalone-endpoint decision are right. The
blocking problem is documentary but sits in the worst possible place: the type's own docblock
asserts the inverse of the shipped semantics of `completed`, on the gate that decides whether a
`CREATE UNIQUE INDEX` migration is run. Alongside it, the gate's honesty about its own staleness —
which `docs/architecture-overview.md` states explicitly for #2317 — is carried on `remainingNull`
and dropped on `completed`, which is the field a UI will actually render.

## Merge Readiness

❌ **Request changes**

## Priority fixes

1. Fix the `ProvenanceBackfillStatus` docblock: `completed` is `remainingNull === 0`, not the latch.
2. Carry the point-in-time/staleness caveat onto `completed`'s DTO description — `completed: true` is
   an observation, not a guarantee, and #2325 re-counts before it acts.
3. Normalise an empty-string cursor to `latchedAt: null`, matching the handler's own predicate.
4. Assert the literal cursor key in the spec instead of `stringContaining('inventory-provenance')`.
5. Register (or reuse) the nil-UUID constant so `check-system-connection-id-mirror.mjs` covers it,
   and correct the "neither is exported" comment.
6. Add the viewer-403 entry in `viewer-role-authz.int-spec.ts`, and an int-spec for the live count.

piotrswierzy (approved):
## Re-review (after force-push `8624e28`)

Re-read against `main` as it stands today. All six prior findings are addressed in code.

### 1. Prior findings

|---|---|---|---|
| 1 | **BLOCKING** — the `ProvenanceBackfillStatus` docblock asserted in bold that `completed` reflects the persisted LATCH, while every implementation is `remainingNull === 0` | **Addressed** | `libs/core/src/inventory/domain/types/inventory.types.ts` — now reads "**`completed` is `remainingNull === 0` — a point-in-time observation, never a standing guarantee**", and the latch is given its own field with its own paragraph. The doc and the code (`inventory-query.service.ts`, `completed: remainingNull === 0`) now say the same thing, which matters because this is what a #2325 implementer reads before a `SET NOT NULL`. |
| 2 | `completed`'s DTO description dropped the staleness honesty `remainingNull`'s carried | **Addressed** | `apps/api/src/inventory/http/dto/provenance-backfill-status-response.dto.ts` — the `completed` description now carries "inherits that count's staleness… #2325 must re-count immediately before it acts rather than trust a previously-observed true", matching `docs/architecture-overview.md` § Inventory (#2317). |
| 3 | `latchedAt` passed the cursor through raw while the writing handler treats `''` as *not latched* | **Addressed** | `inventory-query.service.ts` — `rawLatchedAt !== null && rawLatchedAt.length > 0 ? rawLatchedAt : null`, byte-for-byte the handler's own predicate (`apps/worker/src/sync/handlers/inventory-provenance-backfill.handler.ts:146`, `completedAt !== null && completedAt.length > 0`). Covered on both sides: unit ("normalises an empty-string cursor row to `latchedAt: null`") and integration (a real `''` row written to `connection_cursors` and asserted null over HTTP). |
| 4 | A fourth independently-hardcoded nil-UUID not registered in `scripts/check-system-connection-id-mirror.mjs` | **Addressed** | `inventory.service.ts` now `export const SYSTEM_CONNECTION_ID`, imported by `inventory-query.service.ts`. No fourth copy in production code, and the script's `parseStringConst` regex tolerates the `export` modifier (`\\b<name>\\s*(?::…)?=`), so the existing `SOURCES` entry keeps covering it. The literals in the unit and int specs are deliberate independent pins, which is the right direction. |
| 5 | Cursor-key spec asserted `stringContaining('inventory-provenance')` — a check that cannot fail | **Addressed** | `inventory-query.service.spec.ts`, "reads the latch cursor under the nil-UUID system connection scope" — full literal `'master.inventory-provenance.completedAt:connection:00000000-…'`, with the comment naming the sibling `remainingNull` key as the thing a loose matcher would have let through. This is the `master-sweep-cursor.types.spec.ts` shape asked for. |
| 6 | No `viewer-role-authz.int-spec.ts` entry; doc error "the backfill tracks no cursor" | **Addressed** | `viewer-role-authz.int-spec.ts` — new 403 case for `GET /inventory/provenance-backfill-status`. The doc error is corrected in the domain type: "the pass DOES persist two `connection_cursors` keys… `sweepRemainingCountCursorKey` … and `sweepCompletedAtCursorKey`", and the distinction between "never read back to answer this call" and "the LATCH" is made explicit. |

### 2. New commits

Layering and authorization are clean: `@Roles('admin')` on the new route (`inventory.controller.ts`) with the matching viewer-403 int-spec, so `RolesGuard` deny-by-default (#2079) is satisfied explicitly rather than inherited. The cross-context reach is `ISyncCursorsService` via `SYNC_CURSORS_SERVICE_TOKEN` — an `I*Service`, the sanctioned form — and `SyncModule` is already imported by `InventoryModule` (`inventory.module.ts:86`), which exports the token (`sync.module.ts:181`). The `connection_cursors` addition to `setup.ts`'s truncation list is safe: no other `apps/api` int-spec touches that table, and it carries no FK, so listing it explicitly is required rather than redundant.

**[IMPORTANT]** — `libs/core/src/inventory/infrastructure/persistence/repositories/inventory.repository.ts:1095` (`countMissingProvenance`), as newly exposed by `inventory.controller.ts`
> `SELECT COUNT(*) FROM "inventory_items" WHERE "sourceConnectionId" IS NULL` has **no supporting index** — the only partial index on this table is `IDX_inventory_items_ol_reserved` (`inventory-item.orm-entity.ts:60`). Until now that seq scan was worker-only and cron-paced (#2317, `*/5`); this PR puts it behind an operator-triggerable HTTP GET on the hottest table in the system. #3251's hook mitigates the tab-focus case (`staleTime: 60_000`, `refetchOnWindowFocus: false`), but the browser is not the only caller and the endpoint carries no cap of its own. Either add a partial index `("id") WHERE "sourceConnectionId" IS NULL` in a migration, or state the scan cost on the route's `@ApiOperation` description the way `docs/operations/inventory-duplicate-positions.md` states it for the sibling diagnostic ("run it off-peak"). The live-read decision itself is right and well argued — this is about making its cost visible where a caller sees it.

**[SUGGESTION]** — `apps/api/test/integration/inventory-provenance-backfill-status.int-spec.ts`
> `expect(body.remainingNull).toBeGreaterThanOrEqual(1)` in two cases. `resetTestHarness()` runs per test and each case seeds exactly one NULL row, so `toBe(1)` is available and would catch a predicate that accidentally widened (e.g. dropping the `IS NULL` clause). A `>=` on a count is one step from a check that cannot fail.

**[SUGGESTION]** — `libs/core/src/inventory/application/services/inventory-query.service.ts` (`PROVENANCE_BACKFILL_SWEEP_KIND`)
> The kind string is now declared in two places (here and `inventory-provenance-backfill.handler.ts:103`). That is acceptable — the handler's copy lives in `apps/worker`, which `libs/core` may not import, and both sides are typed `MasterSweepKind` so a typo is a compile error rather than a silent split. Worth noting only because the accompanying comment says the handler's constant "is not exported"; it *is* exported (`handler.ts:281`, as `INVENTORY_PROVENANCE_SWEEP_KIND`) — just not from anywhere core can reach. Tightening the comment to say that would spare the next reader the grep.

### 3. Summary

The BLOCKING doc/code contradiction is resolved at the point that mattered — the domain type a #2325 implementer reads — and the latch is now a first-class, separately-documented field rather than a re-description of `completed`. The `''` normalisation is pinned on both sides of the reader/writer boundary, including against real Postgres, which is exactly the assertion a mock could not make. Finding 4 was fixed by promoting the existing constant rather than adding a fifth copy plus a script entry, which is the better shape. The one remaining concern is operational rather than architectural: an unindexed full-table `COUNT(*)` is now reachable over HTTP.

### Merge Readiness

🔄 **Approve with changes** — no BLOCKING items outstanding. Address the unindexed-count exposure before or shortly after merge.

### Priority fixes

1. **[IMPORTANT]** Index or document the `countMissingProvenance` seq scan now that it is HTTP-reachable.
2. **[SUGGESTION]** Tighten the two `toBeGreaterThanOrEqual(1)` count assertions to `toBe(1)`.
3. **[SUGGESTION]** Correct the "not exported" clause in the `PROVENANCE_BACKFILL_SWEEP_KIND` comment.

piotrswierzy (approved):
## Tech Lead review — `feat(inventory): expose #2317 provenance-backfill completion state`

The `latchedAt` field is the reason this PR is worth more than its size suggests.

### ✅ `latchedAt` closes a false-ready that `completed` alone cannot see

`completed: false` is ambiguous between *"still draining"* and *"latched, and permanently stopped"* — and the second is reachable, because the handler self-latches and never clears the stamp, so a mutation that reintroduces a NULL row after completion leaves `remainingNull > 0` with nothing draining it. Reporting the persisted stamp alongside the live count is what makes those two distinguishable, and the docblock spells out the exact predicate (`remainingNull > 0 && latchedAt !== null` ⇒ re-arm) plus the operator remedy.

That matters here specifically because this read is a **readiness gate for a `SET NOT NULL` migration**. A gate that cannot tell "not done yet" from "will never be done" is a gate that eventually gets waited on forever, or overridden.

### ✅ `remainingNull` is live, and the reason is the right one

The docblock doesn't just assert "never cached", it explains that the pass persists `sweepRemainingCountCursorKey` for observability and that reading *that* would be the trap — the count would inherit the latch's staleness. And `completed` is correctly framed as a point-in-time observation with the explicit instruction that #2325 must re-count immediately before acting rather than trust a previously-observed `true`. That's the honest framing; a `completed: true` that a caller treats as durable is exactly how a migration runs against a table that acquired a NULL row in between.

### ✅ The empty-string normalisation, and why it needed a test on both sides

`rawLatchedAt !== null && rawLatchedAt.length > 0 ? rawLatchedAt : null` matches the handler's own latched-predicate. Passing `''` through would render `latchedAt: ''` — non-null — so every consumer reading this contract would see "latched", conclude the pass is stuck, and apply the documented remedy (delete the cursor row) to a perfectly healthy drain.

Pinning it in **both** the unit spec (against a mock) and the int-spec (writing a real `connection_cursors` row) is right, and the int-spec header says exactly why the unit test isn't sufficient: only a real row proves reader and writer agree on the key string *and* the normalisation.

The unit spec hard-coding the full cursor key rather than calling the builder is the same reasoning `master-sweep-cursor.types.spec.ts` uses — a format change, or picking the sibling `remainingNull` key by mistake, has to fail loudly rather than silently split reader from writer.

### ✅ Housekeeping done properly

- `SYSTEM_CONNECTION_ID` promoted from a private field to an exported const **with** a pointer to `scripts/check-system-connection-id-mirror.mjs` and an explicit note that a fourth unregistered copy is what's being avoided. Correct instinct — that script can only guard copies it knows about.
- `connection_cursors` added to `tablesToTruncate` with the reason (no FK from `connections`, so nothing cascades, and the container is reused across the run). Without it a latch cursor from one case leaks into the next, and the failure would look like a flaky assertion rather than missing cleanup.
- Viewer-403 assertion added beside its sibling diagnostic in `viewer-role-authz.int-spec.ts`, matching the established convention rather than growing a second authz spec.
- Five int-spec cases including the stuck-pass state end to end.

### 🟢 SUGGESTION — `PROVENANCE_BACKFILL_SWEEP_KIND` is declared twice

```ts
const PROVENANCE_BACKFILL_SWEEP_KIND: MasterSweepKind = 'inventory-provenance';
```

…in the service, because the handler's own `BACKFILL_SWEEP_KIND` isn't exported. The comment is honest about this.

The risk is mostly contained: both sides pick from the closed `MasterSweepKind` union, so a *rename* is a compile error in both places. What isn't guarded is a semantic change — the handler switching to a different member — after which this reader silently reads a key nobody writes, reports `latchedAt: null` forever, and the stuck-pass state this PR exists to expose becomes invisible again. Silent, and in the one direction that matters.

`libs/core` can't import from `apps/worker`, so the clean fix is the other direction: declare the constant in `@openlinker/core/sync` beside `MasterSweepKind` and have the handler import it. One declaration, both consumers, no drift surface. Worth doing while the second consumer is new.

### Notes

- CI: six green, `Build` / `Test` / `Test (web)` still queued at review time. Please don't merge until they land.
- The DTO `@ApiProperty` descriptions carry the same caveats as the domain type — a caller reading only the OpenAPI doc still learns that `completed` is point-in-time and what a non-null `latchedAt` beside a non-zero count means.

**Verdict:** ✅ Approve.

Issue #3073, opened 2026-09-10
feat(inventory): duplicate-positions page with readiness KPI strip


norbert-kulus-blockydevs:
## Problem / Context

No page exists to answer "is this install ready for the stricter uniqueness index".

## Proposed Solution

Add `apps/web/src/pages/inventory/duplicate-positions-page.tsx` with a 3-card KPI strip (`groupCount`, `rowCount`, `excessRowCount`), a clear "ready" vs "not ready" banner (0 groups = ready), following `MetricCard`.

## Classification

**Type**: Frontend
**Layer**: Interface (page)
**File(s)**: apps/web/src/pages/inventory/duplicate-positions-page.tsx

## Dependencies

- Part of #3071

## Assumptions

- None

## Acceptance Criteria

- [ ] KPI strip renders groupCount/rowCount/excessRowCount
- [ ] A 0-group result renders an unambiguous "ready" state
- [ ] Tests added or updated for non-trivial logic
- [ ] No architecture boundary violations (CORE ↔ Integration)

PR #3251, opened 2026-09-15
feat(inventory): duplicate-positions API client + query hook

### LINKED ISSUE #3072: feat(inventory): duplicate-positions API client + query hook
## Problem / Context

`GET /inventory/duplicate-positions` (with an optional `maxGroups` param) has no client caller.

## Proposed Solution

Add `getDuplicatePositions(maxGroups?)` to `features/inventory/api/inventory.api.ts` and `use-duplicate-positions-query.ts`, typed to mirror `DuplicatePositionsResponseDto`/`DuplicatePositionGroupResponseDto`/`DuplicatePositionRowResponseDto`.

## Classification

**Type**: Frontend
**Layer**: Interface (API client + hook)
**File(s)**: apps/web/src/features/inventory/api/inventory.api.ts

## Dependencies

- Part of #3071

## Assumptions

- None

## Acceptance Criteria

- [ ] Client method + hook exist and are typed against the DTOs

jakubretajczykBD:
## Summary
- Adds `getDuplicatePositions(maxGroups?)` and `getProvenanceBackfillStatus()` to the inventory API client, plus `useDuplicatePositionsQuery` and `useProvenanceBackfillStatusQuery` hooks.
- Typed to mirror the real shipped backend contract: `DuplicatePositionsResponseDto` / `DuplicatePositionGroupResponseDto` / `DuplicatePositionRowResponseDto` on `GET /inventory/duplicate-positions` (#2319), including the #3239 display enrichment (`productName`, `sku`, `connectionName`, `locationName` — all nullable per the documented sentinel rules), and `ProvenanceBackfillStatusResponseDto` on `GET /inventory/provenance-backfill-status` (#3240) — the second, independent readiness condition for #2325 alongside `groupCount`.
- Both are read-only diagnostics; `maxGroups` bounds only the returned group detail, never the KPI totals.

## Context
First sub-issue of epic #3071 (Frontend — Duplicate stock positions diagnostic page). Depends on the backend contract shipped in #3249 / #3250 (stacked ahead of this one) — this PR only adds the typed client + hooks, no UI yet.

## Test plan
- [ ] `pnpm test` (`src/features/inventory/` — 36/36 passing locally, including pre-existing locations tests)
- [ ] `pnpm type-check`
- [ ] `pnpm lint`

Closes #3072
Part of #3071


piotrswierzy (approved):

**Verdict: ✅ Approve.** A thin, correct transport + query layer that follows every frontend convention it touches. All checks green.

### Conventions checked, all clean

- **No raw `fetch()`** — both methods go through the injected `request` on `createInventoryApi`, and both hooks resolve it via `useApiClient` (`use-duplicate-positions-query.ts:18`, `use-provenance-backfill-status-query.ts:17`). `docs/frontend-architecture.md` § API Client Conventions and the `no-restricted-globals` rule are both satisfied.
- **Server state via TanStack Query**, with query keys **beside the feature API module** in `inventory.query-keys.ts` rather than inline in the hook — the documented placement.
- **The key shape is right, and the comment earns itself.** `duplicatePositions: (maxGroups?) => ['inventory', 'duplicate-positions', maxGroups ?? null]` includes `maxGroups`, which matters because it changes the response (`groups[]` truncation), while `provenanceBackfillStatus: () => [...]` correctly takes no axis. The note that the readiness gate is install-wide rather than per connection is the kind of thing that stops someone adding a spurious `connectionId` later.
- **Types match the wire.** I checked `generatedAt` specifically, since it appears in the FE `DuplicatePositionsReport` but not in #3249's diff — it is pre-existing on `DuplicatePositionsResponseDto:88` and set by the controller, so the type is honest. The four #3239 enrichment fields and their null-sentinel documentation are carried across faithfully.
- No default `retry`/`staleTime` overrides, consistent with the FE-001 baseline of retries disabled by default.

### Findings

**[SUGGESTION]** — `apps/web/src/features/inventory/api/inventory.types.ts:112` and `hooks/use-provenance-backfill-status-query.ts:7`
> Both repeat #3250's claim that the status is live "since the backfill itself tracks no cursor" / "never cached". The second half is right; the first is not — the pass writes `sweepCompletedAtCursorKey` and `sweepRemainingCountCursorKey` to `connection_cursors` (`inventory-provenance-backfill.handler.ts:179-196`); what it lacks is a *scan offset*. Worth correcting in the same pass as the backend copy so the false statement does not survive in four files across two PRs.

**[SUGGESTION]** — `hooks/use-provenance-backfill-status-query.ts`
> The docblock says "so this hook does not poll", which is true by omission (no `refetchInterval`). Given the endpoint is an uncapped `COUNT(*)` over `inventory_items`, stating that as a deliberate constraint on future callers — rather than as a description of current behaviour — would make it harder for someone to add polling to a "readiness" screen without thinking about the cost.

**[Note, no action]** — neither hook is exported from a `features/inventory` barrel, and #3252 deep-imports them from `pages/`. That is fine: `docs/frontend-architecture.md` records that `pages/` may still deep-import from features as a documented gap. It only becomes a barrel question if another *feature* consumes them.

### Merge readiness

✅ **Approve.** Both suggestions are comment-only. Stacked on #3250, so it merges behind it.

piotrswierzy (changes requested):
# Tech Lead review — #3251 (duplicate-positions API client + query hooks)

Reviewed against `docs/code-review-guide.md`, `docs/frontend-architecture.md`,
`docs/testing-guide.md`, `docs/architecture-overview.md` § Inventory (#2317/#2319) and
`docs/operations/inventory-duplicate-positions.md`.

Stacked on #3250 → #3249 — reviewed as the incremental diff.

## Findings

**[BLOCKING]** — `apps/web/src/features/inventory/api/inventory.types.ts` (`ProvenanceBackfillStatus`)

> The interface declares two fields:
>
> ```ts
> export interface ProvenanceBackfillStatus {
>   remainingNull: number;
>   completed: boolean;
> }
> ```
>
> …directly under a docblock claiming it *"mirrors `ProvenanceBackfillStatusResponseDto`"*. It does
> not: that DTO ships **three** fields, and the missing one is `latchedAt` — the only discriminator
> for the degraded state #3250 exists to surface. Per that PR's own type docblock,
> `remainingNull > 0 && latchedAt !== null` means the backfill *"has stopped running and will not
> resume on its own"* and must be re-armed by an operator; `completed` alone *"cannot tell 'still
> draining' apart from 'latched, and stuck'"*.
>
> This is BLOCKING because the typed contract **is** this PR's entire deliverable (`Closes #3072` —
> "API client + query hook", no UI), and the omission is silent: the response body still carries
> `latchedAt` at runtime, TypeScript will never flag its absence, and the diagnostic page built on
> top (#3071) will simply have no type-level route to the one state that needs an operator to act.
> One line to add.

**[IMPORTANT]** — `apps/web/src/features/inventory/api/inventory.types.ts` and `apps/web/src/features/inventory/api/inventory.api.ts` (factually wrong justification)

> Both docblocks assert the live read is needed *"since the backfill itself tracks no cursor"*.
> The backfill tracks **two** `connection_cursors` keys. #3250's own core docblock spells them out
> (`sweepRemainingCountCursorKey`, `sweepCompletedAtCursorKey`), `#3250` reads the second one, and
> `docs/architecture-overview.md` § Inventory (#2317) names both:
> `master.inventory-provenance.completedAt:connection:{system}` and
> `master.inventory-provenance.remainingNull:connection:{system}`.
>
> Reading live is still correct, but for the opposite reason: the persisted count exists and *can go
> stale* (the pass self-latches and stops recounting), which is exactly why the endpoint recounts.
> As written, a future reader who learns the cursors exist will conclude the comment is obsolete and
> "optimise" the read into a cached one — the change this sentence was meant to prevent.

**[IMPORTANT]** — missing tests

> Five files, `+137/-0`, and no test changes. This is against the feature's own established pattern,
> not a general preference: `apps/web/src/features/inventory/api/inventory.api.test.ts` already
> exists and covers the client's other methods, and four sibling hooks in the same folder ship a
> colocated `.test.tsx` (`use-inventory-locations-query`, `use-inventory-location-query`,
> `use-inventory-availability-batch-query`, `use-inventory-location-mutations`).
> `docs/testing-guide.md` requires a colocated unit test for non-trivial logic, and the URL
> construction below is exactly the non-trivial part. The PR's test plan ("36/36 passing locally,
> including pre-existing locations tests") reports that nothing new is covered.

**[IMPORTANT]** — `apps/web/src/features/inventory/hooks/use-duplicate-positions-query.ts` (no cache policy on an expensive read)

> The hook passes only `queryKey` + `queryFn`, so it inherits TanStack Query's defaults —
> `staleTime: 0` plus refetch on mount **and on window focus** — and the app's production
> `QueryClient` sets no overrides (the only `defaultOptions` in the tree are `retry: false` in test
> harnesses). The endpoint behind it is documented as *"two full sequential scans of
> `inventory_items` with no supporting index (there is none that helps a four-column grouping over
> two nullable columns). That is accepted for an operator-run diagnostic; run it off-peak on a large
> catalogue."* (`docs/operations/inventory-duplicate-positions.md`).
>
> As wired, leaving the tab and coming back re-runs both scans, as does every remount. Set an
> explicit `staleTime` and `refetchOnWindowFocus: false`. The same applies to
> `useProvenanceBackfillStatusQuery`, whose docblock already claims *"this hook does not poll"* —
> make that structural rather than a comment, since window-focus refetching is on by default.

**[SUGGESTION]** — `apps/web/src/features/inventory/api/inventory.api.ts` (`getDuplicatePositions`)

> `` const qs = maxGroups !== undefined ? `?maxGroups=${String(maxGroups)}` : ''; `` is the only
> hand-built query string in the file; every other one goes through `URLSearchParams`
> (lines 77-103 and 112). No injection risk here (the value is a `number`), but the inconsistency is
> the kind that gets copied. Worth noting there is no client-side bound: the server rejects
> `maxGroups > 500` with 400, which is correct as the authority — but a caller currently discovers
> the cap from the error, and the FE type offers no hint of it.

**[SUGGESTION]** — `apps/web/src/features/inventory/api/inventory.types.ts` (`DuplicatePositionGroup` docblock)

> Repeats #3249's *"`productName`/`sku` are `null` only when the product could not be resolved"*.
> `Product.sku` is `string | null` in core, so a resolved product with no SKU also reports `null`.
> Fix alongside the backend copy (raised as IMPORTANT on #3249) so the mirror stays a mirror.

### Positive observations

- Types are hand-mirrored rather than imported from `@openlinker/core` — correct per #591, and the
  docblocks name the DTOs they mirror, which is the right way to keep a hand-copy auditable.
- Query keys are added to the shared `inventoryQueryKeys` factory rather than inlined, **and** the
  key comments state the axis decision explicitly (install-wide, no connection axis) — which is the
  thing that usually goes unrecorded and later gets a connection id bolted on.
- `maxGroups` is part of the duplicate-positions key, so two different caps cannot collide in cache;
  `?? null` keeps the key stable for the `undefined` case.
- Hooks follow the established `useApiClient` + `useQuery` shape and the `app/api/api-client-provider`
  import matches every sibling hook — server state stays in TanStack Query per
  `docs/frontend-architecture.md`, no raw `fetch`, no auth logic, no secrets.
- The client docblocks correctly record that both endpoints are admin-only **server-side**, rather
  than implying the frontend enforces it.

---

## Summary

Adds two typed client methods and two query hooks for the duplicate-positions diagnostic and the
provenance-backfill status. The state ownership, key factory and dependency direction are all right,
and the types are mirrored the way this repo mirrors types. The blocking problem is that the
mirrored `ProvenanceBackfillStatus` silently drops `latchedAt` — the only field that distinguishes a
backfill still draining from one latched and stuck — in a PR whose sole deliverable is the contract,
where the omission is invisible to the compiler. Secondary to that: the expensive diagnostic is
wired with default cache semantics that refetch on every window focus, and nothing here is tested
despite the feature already carrying a client test file and four hook test files.

## Merge Readiness

❌ **Request changes**

## Priority fixes

1. Add `latchedAt: string | null` to the FE `ProvenanceBackfillStatus`.
2. Give `useDuplicatePositionsQuery` (and its sibling) an explicit `staleTime` and
   `refetchOnWindowFocus: false`.
3. Correct the "the backfill tracks no cursor" claim in both docblocks — it tracks two.
4. Add tests: the two client methods in the existing `inventory.api.test.ts`, and a colocated test
   per hook, matching the four siblings.
5. (Minor) Use `URLSearchParams` for `maxGroups`; fix the `sku`-is-null copy with #3249.

piotrswierzy (approved):
## Re-review (after force-push `cdb9b52`)

Re-read against `main` as it stands today. All three prior findings are addressed in code.

### 1. Prior findings

|---|---|---|---|
| 1 | **BLOCKING** — the mirrored FE `ProvenanceBackfillStatus` silently dropped `latchedAt`, the only discriminator for "latched, and stuck", in a PR whose sole deliverable is the typed contract | **Addressed** | `apps/web/src/features/inventory/api/inventory.types.ts` — `latchedAt: string \| null` is present with a docblock that states its job in the same words the backend type uses ("the only field that tells 'still draining' apart from 'latched, and stuck'"). Round-tripped through the hook by a real test, not just declared: `use-provenance-backfill-status-query.test.tsx`, "should surface a latched, still-draining status untouched" asserts `latchedAt` survives at `remainingNull: 37, completed: false`. |
| 2 | TanStack Query defaults (`staleTime: 0` + refetch on window focus) re-ran both full scans on every tab focus | **Addressed** | Both hooks set `staleTime: 60_000` and `refetchOnWindowFocus: false`, and both docblocks mark them load-bearing rather than incidental — `use-duplicate-positions-query.ts` cites `docs/operations/inventory-duplicate-positions.md`'s "two full sequential scans… with no supporting index" and says "do not drop them"; `use-provenance-backfill-status-query.ts` adds "Do not add polling here without re-weighing the cost". That is the right place for the reason, since the next author's temptation is a `refetchInterval`. |
| 3 | Propagated the "the backfill tracks no cursor" doc error | **Addressed** | `inventory.types.ts` — "The backfill DOES track two `connection_cursors` keys (`sweepRemainingCountCursorKey` / `sweepCompletedAtCursorKey`), but a persisted count can go stale — the pass self-latches and stops recounting — which is exactly why this endpoint recounts live rather than reading the cursor back." Matches the corrected backend wording in #3250. |

### 2. New commits

Structure is correct: `use-*.ts` in `features/inventory/hooks`, types and client in `features/inventory/api`, `useApiClient()` rather than raw `fetch`, dependency direction `features → shared` respected (`docs/frontend-architecture.md`). Query keys are axis-free for `provenanceBackfillStatus` and carry `maxGroups` for `duplicatePositions`, which is right — the cap changes the payload, so two caps must not share a cache entry. The client correctly declines to enforce the 500 cap client-side and names the server as the authority, matching the #2240 "a mirror must never be stricter than the gate" rule.

**[IMPORTANT]** — `apps/web/src/test/test-utils.tsx:493-510` (`createMockApiClient`, `inventory` namespace)
> `InventoryApi` gains `getDuplicatePositions` and `getProvenanceBackfillStatus`, but neither gets a default in the shared mock. The block ends `...overrides.inventory } as ApiClient['inventory']`, so the cast hides the omission from the compiler and any future test that renders a consumer without overriding gets `undefined is not a function` with nothing pointing at the cause. That is precisely the trap this file's own comments document: `list: vi.fn().mockResolvedValue(…)` on `invoicing` exists "so the /invoices list page … does not hit `undefined` once `list` is added to the InvoicingApi interface", and the #3060 note on `listLocations`/`listActiveLocations` makes the same argument. Harmless today because no component consumes these hooks — but #3073 is the page, and the cost lands there as an unexplained failure rather than here. Two lines: default the report to an empty/ready shape (`{groupCount: 0, rowCount: 0, excessRowCount: 0, groups: [], truncated: false, generatedAt: …}`) and the status to `{remainingNull: 0, completed: true, latchedAt: null}` — the honest zero-config state, the same rationale the #3060 comment gives.

**[SUGGESTION]** — `apps/web/src/features/inventory/hooks/use-provenance-backfill-status-query.ts`
> The docblock says the server "always resolves it with a fresh `COUNT(*)`… nothing is cached server-side", which is true, while `staleTime: 60_000` means what the operator *sees* can be up to a minute old. Both are correct and the combination is right, but the two sentences sit next to each other and a reader can take the first as a claim about the rendered value. Worth one clause saying the displayed answer is informational and #2325 must re-count immediately before it acts — the same rule `docs/architecture-overview.md` § Inventory (#2317) states for `remainingNull`, and which the backend type now states too.

### 3. Summary

The BLOCKING omission is fixed and, more usefully, is now pinned by a test that would fail if `latchedAt` were dropped again — which is what turns a typed contract into a contract. The cache policy is set on both hooks with the reason recorded at the point of temptation rather than in a PR comment, and the cursor doc error is corrected consistently with #3250. The remaining item is the shared mock: adding two methods to an `ApiClient` namespace without defaulting them is the one failure mode this repo has already written comments about twice in that very file, and the cost is paid by the next PR rather than this one.

### Merge Readiness

🔄 **Approve with changes** — no BLOCKING items outstanding. Land the `createMockApiClient` defaults with this PR so #3073 does not inherit the debugging.

### Priority fixes

1. **[IMPORTANT]** Default `getDuplicatePositions` / `getProvenanceBackfillStatus` in `createMockApiClient`'s `inventory` namespace.
2. **[SUGGESTION]** Note in the hook docblock that the displayed status can be up to `staleTime` old and is never itself the #2325 gate.

piotrswierzy (approved):
## Tech Lead review — `feat(inventory): duplicate-positions API client + query hook`

Transport slice for the duplicate-positions page. Types mirror the DTOs, both hooks are cache-configured for the right reason, tests cover forwarding and failure.

### ✅ `staleTime` + `refetchOnWindowFocus: false` are justified, not copy-pasted

```
The endpoint is two full sequential scans of `inventory_items` with no
supporting index … `staleTime` + `refetchOnWindowFocus: false` are therefore
load-bearing, not incidental — do not drop them.
```

That's the right way to write a cache-config comment. Both hooks say what would break if the setting went away, and the provenance hook additionally states that the *absence* of `refetchInterval` is deliberate — which is the harder half to preserve, since "add polling to a status endpoint" is an obvious-looking improvement to whoever picks up the page next. Given `getProvenanceBackfillStatus` is an uncapped `COUNT(*)` resolved live on every call, that note earns its place.

### ✅ The null semantics survive the mirror

`sku: null` meaning two different things, with `productName` as the discriminator, and the `null | 'legacy'` / `locationId === null` sentinels never resolving to a name — all carried into the FE type docblocks rather than left on the backend. That matters because these types are hand-mirrored (the browser can't import `@openlinker/core`, #591), so a caller reading only the FE type is the realistic case.

### ✅ Query keys

`duplicatePositions(maxGroups)` carries the only axis the read has, and both comments note the deliberate absence of a connection axis — the readiness gate is install-wide. Correct, and worth stating, since every other key in this file is connection- or filter-scoped.

### 🟢 SUGGESTION — this client doesn't clamp `maxGroups`; its sibling does

```
The server rejects `maxGroups > 500` with 400 (`MAX_DUPLICATE_POSITION_GROUPS`) —
the authority for the cap; this client does not enforce it.
```

Deliberate and stated, which I'd normally leave alone — except `returns.api.ts` takes the *opposite* approach for the same problem, and explains why:

> *Clamped rather than forwarded: the backend answers HTTP 400 above 100, so an over-large caller value would fail the whole page instead of returning the most it can.*

Two sibling API clients in one app now handle an over-cap page size two different ways. The risk here is genuinely lower (the only caller is the page itself, passing a fixed value), so I'm not asking for a change — but it's worth one of the two docblocks acknowledging the other, so the next person writing a third client knows which is the house rule rather than picking whichever file they opened first.

### 🟢 SUGGESTION — neither hook is exported from the feature barrel

`features/inventory/index.ts` isn't touched, so #3252 will either add the exports or deep-import from `pages/` (allowed today, but the documented gap). The returns stack in flight alongside this one exports at each step, which keeps every PR's public surface visible in its own diff. Worth adding the two lines here.

### Notes

- `createWrapper` is duplicated verbatim across the two new test files. Fine at two; worth hoisting into `test-utils` if a third `renderHook` suite appears.
- `getProvenanceBackfillStatus` asserted with `toHaveBeenCalledWith()` (no args) rather than `toHaveBeenCalled()` — small thing, but it's the assertion that catches a future signature growing an argument the hook forgets to pass.
- CI green (an earlier cancelled run is superseded).

**Verdict:** ✅ Approve.

Issue #3077, opened 2026-09-10
test(inventory): duplicate-positions page tests


norbert-kulus-blockydevs:
## Problem / Context

No test coverage exists for any of the above.

## Proposed Solution

Add colocated tests covering the KPI strip, expand interaction and truncated state.

**Must explicitly cover the four display-fallback matrices** flagged in review on #3255 (never fabricate a name for an unresolved reference):
- `productName: null` → falls back to the raw `productId`
- `sku: null` with a resolved `productName` → sku line omitted, not a fabricated value
- `locationId: null` (ADR-058 decision 2 — master declines to locate) vs. `locationId` resolved but `locationName: null` (falls back to the raw id) vs. both resolved
- `sourceConnectionId`: `null` and the #2317 `'legacy'` sentinel (both render "Not backfilled") vs. resolved-but-unnamed (raw id) vs. resolved-and-named

## Classification

**Type**: Frontend
**Layer**: Interface (tests)
**File(s)**: apps/web/src/pages/inventory/**/*.test.tsx

## Dependencies

- Part of #3071

## Assumptions

- None

## Acceptance Criteria

- [ ] pnpm test passes
- [ ] Tests added or updated for non-trivial logic
- [ ] The four display-fallback matrices above are each explicitly asserted
- [ ] No architecture boundary violations (CORE ↔ Integration)

jakubretajczykBD:
## Mockup-parity findings (post-#3072–#3077)

After the epic's six sub-issues shipped, I compared the built page against the **reviewed mockup** ([artifact](https://claude.ai/code/artifact/9e6e1b2b-275a-4941-b25e-45ab18081a1a?sk=FdmXcxQe0uKs9QUQPwJk3g)) line-by-line. Several elements the mockup modeled were never actually captured in any of #3072–#3077's own issue text, so they shipped without being decided on at all — not a regression against any specific sub-issue, just a gap between "what the mockup showed" and "what got written down as scope."

### Missing entirely (closed in #3264)
- **CSV export** — mockup's `exportCsv()` downloads `duplicate-positions-{date}.csv`, every group flattened to one row per `inventory_items` row, "for offline triage or a support ticket." No sub-issue mentioned this at all.
- **Header Refresh button** (always visible, not just in the error state) + the `generatedAt` timestamp — the field was already on the wire (`DuplicatePositionsReport.generatedAt`) but nothing rendered it.
[...]

Issue #3074, opened 2026-09-10
feat(inventory): expandable grouped table (per-row inventory_items detail)


norbert-kulus-blockydevs:
## Problem / Context

Nothing surfaces which position keys collide, or the individual colliding rows.

## Proposed Solution

Render `groups[]` as a `DataTable` with `expandable` rows (product/variant/location/source columns + row/live-row counts), expanding to a per-row detail table (id, availableQuantity, reservedQuantity, isStale, updatedAt), per the reviewed mockup at https://claude.ai/code/artifact/9e6e1b2b-275a-4941-b25e-45ab18081a1a.

## Classification

**Type**: Frontend
**Layer**: Interface (component)
**File(s)**: apps/web/src/pages/inventory/duplicate-positions-page.tsx

## Dependencies

- Part of #3071

## Assumptions

- None

## Acceptance Criteria

- [ ] Each group row expands to show its individual inventory_items rows
- [ ] Stale rows are visually distinguished
- [ ] Tests added or updated for non-trivial logic
- [ ] No architecture boundary violations (CORE ↔ Integration)

PR #3252, opened 2026-09-15
feat(inventory): duplicate-positions page with readiness KPI strip

### LINKED ISSUE #3073: feat(inventory): duplicate-positions page with readiness KPI strip

jakubretajczykBD:
## Summary
- Adds `duplicate-positions-page.tsx` with a `PageLayout`, loading/error states, and a 4-card KPI strip (`groupCount` / `rowCount` / `excessRowCount` / provenance rows remaining), backed by `useDuplicatePositionsQuery` + `useProvenanceBackfillStatusQuery`.
- Readiness for #2325 is TWO independent conditions, not one (#3240): `groupCount === 0` AND the #2317 provenance backfill has completed — a scan taken mid-backfill can read `groupCount: 0` and then flip non-zero once NULL rows collapse to `'legacy'` and reveal a collision the scan couldn't see yet. The readiness `Alert` renders both as a two-item checklist (each with its own `StatusBadge`) rather than a single combined sentence, so an operator can tell which condition is still pending.

## Context
Second sub-issue of epic #3071, stacked on #3251 (the API client + hooks). No table/expandable rows yet — that's #3074.

⚠️ **Stacked PR** — base branch is `3072-duplicate-positions-api-client` (#3251), not `main`. Merge #3251 first.

## Test plan
- [ ] `pnpm type-check`
- [ ] `pnpm lint`
- [ ] Component tests land in #3077 (test(inventory): duplicate-positions page tests), stacked at the end of this chain

Closes #3073
Part of #3071


jakubretajczykBD:
## Tech review (update, df10b4985)

**Summary**: Small, well-scoped diagnostic page correctly composed from existing primitives (`PageLayout`, `KpiCard`, `Alert`, `StatusBadge`, `LoadingState`/`ErrorState`). The three-state readiness derivation (`undefined`/`true`/`false`, never defaulting an unresolved read to "ready") is correct, and the CSS addition parses cleanly (`check-css-structure.mjs` passes).

**Previously flagged, now fixed in `df10b4985`**:
- ✅ Operator-facing copy no longer leaks internal issue numbers (`#2317`/`#2325` stripped from the checklist spans, footnote, and KPI card description) — dev-facing doc-comment header keeps its refs.
- ✅ Added a doc-comment callout noting the backend endpoint is `@Roles('admin')`-gated and that the follow-up route/nav-wiring PR (#3074) must gate FE visibility accordingly (`docs/frontend-architecture.md` § Access Control And UI Visibility) — so it isn't dropped once the route lands.

**Remaining**: none blocking. `error.message` rendering directly in `ErrorState` is fine given the rest of the codebase's convention (ApiError messages are already operator-safe); worth a glance at what the endpoint actually returns on a 403 if that becomes reachable before #3074 lands the gate, but not actionable here.

### Verdict

piotrswierzy (approved):

**Verdict: 🔄 Approve with changes.** The page is well built and the two-condition readiness model is exactly right. Two things overstate what the page can actually guarantee — one about authorization, one about what the backfill number means.

### What it gets right

**The three-state readiness logic is the part most worth having.** `noDuplicateGroups` / `backfillComplete` / `isReady` are each `undefined` until their read settles, with the comment stating that a single `undefined` must never be read as "not ready" (`duplicate-positions-page.tsx:64-72`). That is the repo's standing rule — absence is not a negative claim — applied to a gate for a destructive migration, which is where getting it backwards is most expensive.

**Both conditions are rendered as separate checklist items with their own badges**, not collapsed into one sentence. Since the two can disagree (and the whole reason #3240 exists is that `groupCount: 0` alone is not sufficient), a combined verdict would have hidden which half is outstanding.

**Tokens only** — `var(--space-2)`, `var(--text-secondary)`, no raw hex. Shared primitives throughout (`PageLayout`, `KpiCard`, `Alert`, `StatusBadge`, `LoadingState`/`ErrorState`). Read-only; nothing writes.

I also checked whether the inline operator copy needs to live in a `*.copy.ts`: it does not. `scripts/check-ui-vocabulary.mjs` scans only under `SCAN_ROOT_PARENT = apps/web/src/features`, so a page is out of its scope by construction.

### Findings

**[IMPORTANT]** — `apps/web/src/pages/inventory/duplicate-positions-page.tsx` (whole component)
> The backing endpoints are `@Roles('admin')`, and this page has **no frontend gate at all**. The docblock defers that to whichever PR wires the route/nav (#3074 → #3261) and asks it to gate a `requiresRole: 'admin'` nav contribution — but **a nav gate hides an affordance; it does not gate a route.** A non-admin reaching the URL directly (typed, bookmarked, or a link pasted in chat) still mounts this component, fires two 403s, and lands on `ErrorState` rendering the raw backend `error.message`.
>
> `docs/frontend-architecture.md` § Access Control And UI Visibility is explicit that content — "a banner, panel, section, table column" — is gated with `AccessGate` (row 1 of the three-row table), and that `AccessGate` owns the session-hydration window centrally so a naive gate cannot flash the denied branch. Wrapping the page body in `AccessGate` is the one-line version; gating the route object is the other. Either closes it; the nav gate alone does not, and deferring to a sibling PR that only does the nav means it stays open.

**[IMPORTANT]** — `duplicate-positions-page.tsx:104-107`
> `Provenance backfill still running (${remainingNull} row(s) remaining)` asserts the pass is **running**, which the data does not support. The backfill self-latches: once `sweepCompletedAtCursorKey` is stamped, `execute()` returns immediately without touching the table (`inventory-provenance-backfill.handler.ts:144-150`), and the doc's only remedy is manual — "Deleting that cursor row re-arms the drain." So in the exact case #3250's docblock invokes to justify a live read (a later mutation reintroduces a NULL row), this page will say "still running" **forever** about a pass that has stopped, and the operator waits on a number that will never move.
>
> This is downstream of #3250 not exposing the latch, so the clean fix is there. Until it does, the honest phrasing here is the observation rather than the inference — e.g. "N row(s) still missing provenance" — which is true in both states. Stating *running* is the page asserting more than it knows.

**[SUGGESTION]** — `duplicate-positions-page.tsx:82-86`
> `ErrorState` renders `error.message` verbatim. For the 403 case above that surfaces backend error text to a user who simply lacks the role; a `403`-aware branch ("You need admin access to view this diagnostic") would read better. Largely moot once the access gate lands, since the 403 stops being reachable.

**[SUGGESTION]** — `:95` vs `:102`
> The two pending badges use different tones — `'error'` for duplicate groups, `'warning'` for the backfill. Both are "a condition that is not yet met", and the stronger tone on duplicates is arguably right (it needs manual DB remediation, the other just needs time), but if that distinction is intended it is worth a word, since the checklist otherwise reads as two peers.

### Merge readiness

🔄 **Approve with changes.** The structure, the state handling and the styling are all right. Two fixes before merge:

1. Gate the page for non-admins with `AccessGate` (or gate the route) — do not rely on #3261's nav gate, which does not cover direct navigation.
2. Reword the backfill-pending sentence so it states the observation rather than inferring that the pass is running, or land #3250's latch exposure first and branch on it.

piotrswierzy (approved):
# PR Review — #3252 `feat(inventory): duplicate-positions page with readiness KPI strip`

## Step 1 — Context

Second slice of epic #3071, stacked on #3251. Adds `apps/web/src/pages/inventory/duplicate-positions-page.tsx` plus an `enabled: isAdmin` gate on the two query hooks and a small CSS block for the readiness checklist.

Baselines checked on `main`: `apps/web/src/pages/oms/sourcing-rules-page.tsx` (the closest admin-gated page), `shared/ui/kpi-card.tsx`, `shared/ui/feedback-state.tsx`, `shared/auth/use-permission.ts`, and the backend contract (`libs/core/src/inventory/application/services/inventory-query.service.ts:146`, `apps/api/src/inventory/http/inventory.controller.ts:119`).

## Step 2 — Findings

### Positive observations (non-obvious things done right)

- **The "absence must never read as an all-clear" rule is satisfied structurally, not by convention.** `duplicate-positions-page.tsx:57-62` types both conditions as `boolean | undefined` and derives `isReady` as `undefined` unless *both* have settled, and the render branch at `:105` additionally gates on `report && provenance`. A half-loaded read therefore cannot render a green "Ready" — which is precisely the failure mode `docs/architecture-overview.md` § Inventory (#2319 as the #2325 gate procedure) makes expensive.
- **Loading / failed / empty are three distinct renders**, not two (`:94`, `:99`, `:105`), and `error` ORs both queries so one failed read can never leave a partial KPI strip on screen.
- **The two readiness conditions are rendered as two badged list items rather than one combined sentence** (`:107-129`). Given the #3240 reasoning — a scan taken mid-backfill can read `groupCount: 0` and flip later — collapsing them would have produced exactly the false all-clear the epic exists to prevent.
- **The admin gate is paired with the session-hydration window** (`:69`, `isSessionReady && !isAdmin`). `useIsAdmin`'s own docblock (`shared/auth/use-permission.ts:80-82`) requires that pairing — "returns `false` while the session is still hydrating, so a call site must not treat it as a denial worth explaining" — and this is one of the few call sites that honours it. `enabled: isAdmin` on both hooks also avoids a guaranteed 403 round-trip.
- Dependency direction is clean: `pages → features → shared` only; the `features → app/api/api-client-provider` crossing is the sanctioned one (`docs/frontend-architecture.md:642`). No raw `fetch()`. CSS uses `var(--space-2)` / `var(--text-secondary)`, no raw hex.

### **[IMPORTANT]** — `apps/web/src/pages/inventory/duplicate-positions-page.tsx:130-134`

> The readiness footnote renders a repository file path into operator-facing copy:
> `"See docs/operations/inventory-duplicate-positions.md for the remediation procedure."`
>
> An operator reading this page inside the product cannot open that path — it is a path in a git checkout. This is the same class of leak the review of this very PR already removed (raw issue numbers in rendered copy, per #3255's PR body). Grepping `apps/web/src` for `docs/operations|docs/architecture` returns exactly one hit and it is a *code comment*, so there is no precedent for rendering a repo path as UI text.
>
> State the remedy in the copy, or link to a docs URL the deployment actually serves. `docs/frontend-ui-style-guide.md` § 2.1's P9 rule ("no model-internal vocabulary in user-facing copy") is the governing principle; note that `scripts/check-ui-vocabulary.mjs` does **not** scan `pages/` (its own header, limitation 1), so nothing will fail the build on this.

### **[SUGGESTION]** — `duplicate-positions-page.tsx:105` (the success branch)

> `DuplicatePositionsReport.generatedAt` is on the contract and is never rendered, and the success path carries no refresh control — `Retry` exists only inside `ErrorState` (`:99-104`).
>
> This page's own docblock (`:9-19`) argues that a readiness verdict is time-sensitive. A verdict with no "as of" and no way to re-take the scan after an operator has run the remediation SQL means the only refresh affordance is a browser reload. An `as of {formatDateTime(report.generatedAt)}` line plus a header `Refresh` action would close both. The two-condition AND is a real mitigation, which is why this is a suggestion rather than an important.

### **[SUGGESTION]** — `duplicate-positions-page.tsx:95`

> `liveRegion="off"` on a **page-level** `LoadingState`. Every one of the ten existing `liveRegion="off"` call sites is a nested panel, picker or modal (e.g. `features/mappings/components/MappingPanel.tsx:312`, `features/listings/components/bulk/shop-attribute-picker.tsx:115`) — surfaces where suppressing the announcement avoids competing with a page-level region. Here it is the page's only region, so a screen-reader user gets no announcement at all while the scan runs. The `polite` default is the right value.

### **[SUGGESTION]** — copy, `:78` / `:89` / `:97` / `:155`

> Raw schema identifiers appear in rendered text (`:78`, `:89`, `:97`, `:155`) (`inventory_items`, `sourceConnectionId`). Defensible for this specific audience — the remediation is hand-written SQL and the operations doc uses the same names — so this is a consistency note rather than a defect. Flagged only so the choice is deliberate.

### Notes, not findings

- The page is not yet routed or reachable (no `duplicate-positions` hit anywhere in `apps/web/src` outside this file). That is #3076 in the same stack, as designed.
- Tests are deferred to #3077. `docs/testing-guide.md` would normally want a `.test.tsx` beside the non-trivial readiness derivation; the stack defers it deliberately, so the ask is that **#3077 covers the `boolean | undefined` readiness matrix specifically** — one-loaded/one-pending, one-errored, and both-zero — since that derivation is the whole correctness claim of this slice.
- `maxGroups` is guaranteed `>= 1` server-side (`inventory-query.service.ts:154`), so `groups.length === 0 ⟺ groupCount === 0`. Worth knowing because it is what makes #3255's empty state a sound claim rather than a truncation artefact.

## Step 4 — Final Assessment

### Summary

A careful, well-reasoned slice. The thing that matters most on this page — that a partially-loaded, failed, or non-admin read can never render as "Ready" — is handled structurally rather than by discipline, and the two-condition checklist is the right reading of #3240. The only substantive issue is that the remediation footnote points an operator at a path inside a git repository.

### Merge Readiness

🔄 **Approve with changes** — no BLOCKING issues. The IMPORTANT copy fix should land before or immediately after merge.

### Priority fixes

1. **IMPORTANT** — replace the rendered `docs/operations/inventory-duplicate-positions.md` path with copy an operator can act on (or a served URL).
2. **SUGGESTION** — surface `report.generatedAt` and add a refresh action on the success path.
3. **SUGGESTION** — drop `liveRegion="off"` on the page-level `LoadingState`.

piotrswierzy (changes requested):
# Re-review — #3252 `feat(inventory): duplicate-positions page with readiness KPI strip`

Re-reviewed at `25643dce` against my 2026-09-18 approval. `main` re-read (it has moved — the CI-perf merges); every citation below is against the current tree.

## 1 — Prior findings

|---|---|---|---|
| 1 | **IMPORTANT** — readiness footnote renders the repo path `docs/operations/inventory-duplicate-positions.md` as operator copy | **Addressed** | `duplicate-positions-page.tsx:301-305` now reads *"Resolving duplicate groups is a manual database operation — ask an engineer to run the remediation procedure, then refresh this page to re-check."* The path survives only in the module docblock (`:5-7`), which is a code comment, not rendered. |
| 2 | **SUGGESTION** — surface `report.generatedAt`, add a refresh action | **Addressed** | `:263-265` (`As of {formatDateTime(report.generatedAt)}`) and `:243-247` (header `Refresh`, `disabled={isRefetching}`). `disabled` + `'Refreshing…'` matches four existing precedents (`prestashop-rate-limit-readout.tsx:93-95`, `insights-page.tsx:452-453`, `allegro-seller-defaults-section.tsx:179-181`, `offer-publication-status-panel.tsx:126`). |
| 3 | **SUGGESTION** — drop `liveRegion="off"` on the page-level `LoadingState` | **Addressed** | `:251-254` now uses the `polite` default (`feedback-state.tsx:27`). |
| 4 | **SUGGESTION** — raw schema identifiers in copy | Unchanged, deliberate | `:229`, `:253`. Consistency note only, as stated. |
| 5 | **Cross-cutting** — pin the `boolean \| undefined` readiness matrix into #3077's scope *before* the chain lands | **Not addressed** | #3077's body still reads only *"Add colocated tests covering the KPI strip, expand interaction and truncated state."* None of the three asks from the 18 Sep reviews is named there. |

### Positives re-confirmed against the current head

- `noDuplicateGroups` / `backfillComplete` are still `boolean | undefined` and `isReady` is still `undefined` unless **both** have settled (`:59-66`); the success branch still additionally gates on `report && provenance` (`:261`). A half-loaded or one-errored read still cannot render green.
- Loading / failed / success remain three distinct renders (`:250`, `:255`, `:261`), and `error` still ORs both queries (`:57`).
- The admin gate is still paired with `isSessionReady` (`:220`), honouring `use-permission.ts:80-82`.
- Dependency direction unchanged: `pages → features → shared`; no raw `fetch()`; the new CSS uses `var(--space-2)` / `var(--text-secondary)` only.

## 2 — Findings on the new commits

### **[BLOCKING]** — `apps/web/src/features/inventory/hooks/use-duplicate-positions-query.ts:31` and `use-provenance-backfill-status-query.ts:27`

> **`Test (web)` is red on this PR, and this PR is the cause.** Base PR #3251 is green on the same job; #3252 fails with, repeated across both hook suites:
>
> ```
> Error: useSession must be used within a SessionProvider
>  ❯ useSession src/shared/auth/use-session.ts:8:11
>  ❯ useIsAdmin src/shared/auth/use-permission.ts:84:23
>  ❯ useDuplicatePositionsQuery src/features/inventory/hooks/use-duplicate-positions-query.ts:31:19
>  ❯ src/features/inventory/hooks/use-duplicate-positions-query.test.tsx:71:41
> ```
> (job `105574150484`.)
>
> Adding `useIsAdmin()` — which calls `useSession()` (`use-permission.ts:84`) — made both hooks require a `SessionProvider` in their render tree. The pre-existing suites for those exact hooks, which arrive on the base branch, wrap only in `ApiClientProvider` + `QueryClientProvider` (`use-duplicate-positions-query.test.tsx:17-27`). Every test in both files now throws. The PR modified two hooks and did not update the tests that cover them.
>
> This is not the known `apps/web` full-suite parallelism flake — it is deterministic, it names this PR's own import chain, and the same job passes one branch down.
>
> **Remedy**, and note it is *not* a one-line wrapper swap: the wrapper needs `SessionProvider` fed by an **admin** adapter, or `enabled: isAdmin` leaves the query disabled and the three existing assertions silently invert into "the api client was never called" — a green-looking test asserting the opposite of its name. `test/test-utils.tsx` already ships both halves: `createAuthenticatedSessionAdapter(user)` (`:1156`) and `renderWithProviders` (`:1233`, which mounts `SessionProvider` at `:1264`).
>
> Because #3255 and #3256 are stacked on this branch they carry the same hook code; neither can go green until this lands.

### **[IMPORTANT]** — the behaviour this PR added is the one thing left untested

> `enabled: isAdmin` is the new contract of both hooks — *"a non-admin session never triggers a 403 round-trip"* (the hooks' own new docblocks). Nothing asserts it. The fix above forces you into these files anyway, so the missing case belongs in the same commit rather than in #3077: render each hook under a **non-admin** session adapter and assert the api-client method was **not** called. Per `docs/testing-guide.md`, a claim a docblock makes and no test checks is the class of assertion that quietly stops being true.

### **[IMPORTANT]** — `duplicate-positions-page.tsx:284-289` — `latchedAt` is fetched and never surfaced, so "wait" and "act" render identically

> `ProvenanceBackfillStatus.latchedAt` carries this, verbatim, in its own contract docblock (`features/inventory/api/inventory.types.ts`):
>
> > *"Non-null while `remainingNull > 0` means the pass has stopped running on its own — a later mutation reintroduced a NULL row after completion — and needs an operator to re-arm it. **This is the only field that tells 'still draining' apart from 'latched, and stuck'.**"*
>
> The checklist renders one sentence for both — `` `${provenance.remainingNull} row(s) still missing provenance` `` (`:288`) — and the KPI card (`:311-316`) repeats the same number. A backfill that is draining normally needs the operator to wait; a latched-and-stuck one needs them to act, and today they are told to do the same thing. On a page whose entire purpose is a readiness verdict, and whose governing rule is that two materially different states must stay distinguishable, the discriminator arriving on the wire and never reaching the screen is the same class of defect as the absence-vs-failure rule this page otherwise honours well. One extra branch: `latchedAt !== null && remainingNull > 0` → a distinct sentence and an `error`-tone badge.

### **[SUGGESTION]** — `:255-260` vs `:261` — a failed *refetch* now discards a good report

> The `error` branch precedes the `report && provenance` branch, and `refetch()` keeps `data` while setting `error`. Before this commit the only way to refetch was a remount, so the ordering never bit; with a `Refresh` button it does — one transient 502 replaces a fully-rendered readiness verdict with a full-page `ErrorState`. It is recoverable (Retry is right there) and it is never *wrong*, so this is a suggestion: consider keeping the last good report on screen and rendering the refresh failure as an inline `Alert`, with the "as of" line then doing real work.

### **[SUGGESTION]** — `:68`

> `isRefetching` is assigned `isFetching`, which is also true on the very first fetch. Harmless here because the button only renders once `report && provenance` exist, but the name promises a narrower thing than the value. `isFetching` would read truer.

### Not findings (checked, and deliberately not raised)

- `.duplicate-positions-generated-at { font-size: 0.8125rem }` — `index.css` defines no font-size tokens and carries 173 existing `0.8125rem` declarations. Repo convention.
- Every `var(--…)` in the new CSS block resolves against a real declaration (`--space-2`, `--text-secondary`).
- `formatDateTime` exists (`shared/format/format-date.ts:15`) and `generatedAt: string` is on the contract.

## 3 — Final assessment

### Summary

Every prior review item on this PR is genuinely fixed in the code — the footnote no longer points an operator at a git path, `generatedAt` and a refresh action are both there, and the page-level loading announcement is restored. The structural property that matters most, that a partial or failed read can never render "Ready", is intact. But the same commit that added the admin gate left the pre-existing tests for both modified hooks failing, and `Test (web)` is red on the head of this branch for that reason alone — with two stacked PRs behind it.

### Merge readiness

❌ **Request changes** — one BLOCKING item, mechanical to fix.

### Priority fixes

1. **BLOCKING** — give both hook suites a `SessionProvider` with an admin adapter (`test-utils.tsx:1156`); `Test (web)` must be green before this or anything stacked on it merges.
2. **IMPORTANT** — while in those files, assert the new contract: a non-admin session issues no request.
3. **IMPORTANT** — surface `latchedAt`, so "still draining" and "latched and stuck" stop rendering as the same sentence.
4. **SUGGESTION** — keep a loaded report on screen when a refresh fails; rename `isRefetching`.
5. Carry-over — #3077 still does not name the `boolean | undefined` readiness matrix. Please pin it there now.

piotrswierzy (approved):
## Tech Lead review — `feat(inventory): duplicate-positions page with readiness KPI strip`

The page consumes #3250's `latchedAt` properly, and the admin gate is done in both places it needs to be.

### ✅ The page gates itself, not just the nav

```
the page gates itself the same way — rendering an access-denied state for a
non-admin session rather than relying on a sibling PR's nav gate, which
covers only the nav affordance and not a direct navigation to this route
```

Correct, and the reason is the one that matters: #3076's nav entry hides a link, it doesn't protect a URL. Checking `isSessionReady && !isAdmin` rather than `!isAdmin` is the other half — while the session is hydrating, `permissions` is `[]` and a naive gate flashes access-denied before revealing the page. That's the same rule `AccessGate` owns centrally (`docs/frontend-architecture.md`: *"not known yet" is not "denied"*), applied by hand here because this is a whole-page early return rather than a subtree.

### ✅ `enabled: isAdmin` on both hooks, with a test that can't pass vacuously

The gate stops a non-admin from firing a guaranteed 403, and the tests pin it — but the detail worth calling out is this:

```ts
await waitFor(() => expect(result.current.session.isReady).toBe(true));
expect(getDuplicatePositions).not.toHaveBeenCalled();
```

`SessionProvider` starts anonymous and hydrates async, so asserting "never called" immediately would pass before `enabled` was ever evaluated against the real session — the test would be green against a hook with no gate at all. Waiting for `isReady` first is what makes it a real assertion, and the comment says so. That's the kind of thing that usually gets found two years later when someone deletes the gate and nothing goes red.

### ✅ "Stuck" is rendered as its own state, not folded into "Pending"

This is the payoff for #3250's `latchedAt`:

> *Collapsing both into "Pending" told the operator to wait on a number that would never move.*

Three-state badge (`Done` / `Stuck` / `Pending`), an error tone for stuck, and copy that names the remedy ("It will not resume on its own — ask an engineer to re-arm it"). The KPI card's `description` changes with it too, so the two don't disagree. Holding `stuckSinceLatchedAt` as the timestamp rather than a boolean, so TS narrows it for `formatDateTime` without a cast, is a nice touch.

### ✅ A failed refresh doesn't blank a loaded report

```ts
const initialLoadError = !hasLoadedReport ? error : undefined;
const refreshError   =  hasLoadedReport ? error : undefined;
```

TanStack keeps `data` from the last success while `error` is set on a refetch, so routing every error to the full-page `ErrorState` would throw away a perfectly good report because a manual Refresh timed out. Showing an inline "Refresh failed — showing the last successful scan below" is the right degradation, and the `As of {generatedAt}` line above it means the operator can tell how stale what they're looking at is. The two together are what make it safe.

### ✅ Two conditions, stated as two

Rendering the readiness as a checklist with a badge each — rather than one combined sentence — is right, because the two conditions have different remedies (manual DB remediation vs. re-arming a worker pass). The page header explains *why* both are required, including the failure mode where a mid-backfill scan reads `groupCount: 0` and later flips non-zero. That's the trap #3240 exists to prevent, restated where an operator will actually read it.

### 🟢 SUGGESTIONS

1. **No test for the page in this PR.** 208 lines with six render branches (admin-denied, session-hydrating, loading, initial error, refresh-error-with-data, ready/not-ready/stuck) ship untested here. #3262 is the dedicated tests PR, so this is a stack split rather than an omission — but it does mean the admin gate and the stuck state are unasserted until then. Worth confirming #3262 covers the **admin-denied** and **stuck** branches specifically; those are the two whose absence would be invisible.

2. **`error = duplicatesQuery.error ?? provenanceQuery.error`** — if both reads fail, only the first message is shown. Fine for a diagnostic, but a network outage takes out both and the operator sees one of two identical-looking failures. Low priority.

3. **Copy is inline rather than in a `*.copy.ts`.** `scripts/check-ui-vocabulary.mjs` doesn't scan `pages/inventory`, so nothing is bypassed — just noting the divergence from the returns stack landing beside it, where copy modules are the convention. Fine as-is for a page with no shared vocabulary.

### Notes

- CSS scoped to the new classes, tokens throughout, with a comment explaining why the checklist is a list of badges rather than one sentence.
- `hasLoadedReport` requires *both* reads to have settled before either condition is stated — a single `undefined` is never read as "not ready".
- CI: seven green, `Test` in progress at review time.

**Verdict:** ✅ Approve.

PR #3243, opened 2026-09-14
feat(oms): sourcing rules admin UI - view, author and order the router's ruleset

norbert-kulus-blockydevs:
Brings the sourcing rules admin UI to `main`: the screen where an operator reads, authors and orders the rules the fulfilment router evaluates when deciding which location an order is sourced from.

Seven reviewed and approved pull requests make up the feature, each already merged into this branch and each keeping its own commit:

| | |
|---|---|
| #3226 | sourcing-rules API client + query hooks |
| #3228 | ordered sourcing-rules table with arrow reorder |
| #3229 | create/edit sourcing-rule dialog |
| #3230 | delete/retire flow |
| #3233 | read states and the unrecognised-rule refusal |
| #3234 | the page, reached from a `/settings` tile |
| #3235 | copy, tile and a structural coverage guard |

Plus the committed UX mockup (#3119) the seven are built against, which `docs/frontend-architecture.md` § UX Mockups requires to live in the repo for the life of the feature.

## Merge with "Rebase and merge", not squash

Squashing would collapse all seven features into a single commit and release-please would render **one** changelog line for the whole feature instead of one per pull request. Rebase replays each commit individually, so every `feat(oms): …` reaches `main` under its own PR number and the changelog reads the way the work was actually reviewed. `main` requires linear history, which rebase satisfies and a merge commit would not.

## Why this pull request exists at all

The mockup PR (#3119) was merged into this branch rather than into `main`, and the seven PRs above were stacked on top of it. GitHub refuses to retarget a pull request that is part of a stack:

```
422 - Cannot change the base branch because the pull request is part of a stack.
```

So the stack had no route to `main` of its own, and this is that route. Without it the seven merged pull requests stay on this branch and never ship.

## Issues this closes

None of these closed on their own: each constituent PR merged into `3055-sourcing-rules-admin-ui` rather than into `main`, and GitHub only auto-closes an issue when the PR lands on the default branch. So every one of them is still open, and this is the pull request that actually delivers them.

Closes #3112
Closes #3056
Closes #3057
Closes #3058
Closes #3059
Closes #3060
Closes #3061
Closes #3062

The epic **#3055** is deliberately **not** closed here. All eight of its sub-issues are now delivered (#3063 landed separately as #3198), but two of its three acceptance criteria are not a merge event - the surface still needs manual verification against a running `pnpm start:dev:api` + `pnpm start:dev:web`. Closing it on this merge would tick a box nobody checked.

PR #3253, opened 2026-09-15
feat(sales-documents): refuse a colliding rule in the composer, before the save

### LINKED ISSUE #3190: feat(sales-documents): rule overlap detection before save

norbert-kulus-blockydevs:
Closes #3190

## Why this is a re-land, not new work

This content was reviewed and merged once already, under PR #3242, into the stack base
`3189-inline-rule-amounts`. PR #3241 then merged that base into `main` as a **squash**
(`e3a8a661a`, single parent). A squash carries only the diff of the branch being merged,
and this work sat *above* that base, so its commits were never included.

Verified rather than inferred: neither the #3242 merge commit
(`6049d62e6`) nor `origin/3190-rule-overlap-detection` is an ancestor of `origin/main`,
and `detect-sales-document-rule-overlap.ts` is absent from `origin/main`. Everything lower
in the stack (#3186, #3187, #3188, #3189) survived inside the squash; only this issue's
work was lost. The epic checkbox and the issue state both said it had shipped.

## What it does

The rule engine has no priorities, so when two rules match one order it resolves
`unresolved` / `conflicting-rules-equal-priority` and the order is **held** - ADR-041's
rule that a fiscal document is never picked silently, because a wrong pick is a legal
event. That behaviour is correct, and it is also invisible: the operator meets it days
later as a hanging order.

`detectSalesDocumentRuleOverlap` (pure, in `domain/domain-services/` beside
`evaluateSalesDocumentRules`) answers the question at the one moment somebody can still
act on it, and the composer gates its save on the answer.

Four properties are load-bearing, and are argued in full in the paragraph this PR adds to
`docs/architecture-overview.md`:

1. **Three outcomes, never two.** A pair provably can both match, provably cannot, or this
   build could not decide. The third is reported rather than folded into a neighbour,
   because silence reads as *no conflict* - the false reassurance the check exists to remove.
2. **Decidable only because the vocabulary is closed and the composer is AND-only.**
   An OR tree would make exactly the interesting cases undecidable, so
   `SalesDocumentCondition` must not grow a disjunction without revisiting this function.
3. **Different currencies prove non-overlap.** An inline amount is compared and never
   converted (#3189), so two differently-priced rules cannot describe one order.
4. **It is a READ; the write path is deliberately not gated on it.**
   `POST /sales-documents/rules/overlap-check` persists nothing and refuses nothing. An
   operator may still knowingly save an overlapping pair through the API, and the runtime
   keeps failing safe.

## How the re-land was done

19 files: 6 new, 13 modified.

- **6 new files** taken from the branch tip verbatim - no merge risk.
- **10 modified files** taken from the branch tip, after confirming `main` has not touched
  them since the squash.
- **3 files hand-merged**, because `main` moved under them:
  - `apps/web/src/index.css` - a real conflict against #3243's sourcing-rules CSS. Resolved
    by keeping `main` and re-inserting only the `.rule-card--highlighted` block at its
    anchor.
  - `apps/web/src/test/test-utils.tsx` - applied three-way; the addition is the empty
    overlap verdict as the mock **default**, so a composer test that does not opt in renders
    neither banner.
  - `docs/architecture-overview.md` - the one added paragraph re-inserted at its anchor in
    the Sales Documents section.

The branch also carried review fixes above the feature commit. Those touching this issue's
files come with it; two fiscalization fixes on that branch touch files outside this scope
and are left for their own issues.

## Verification

- `pnpm type-check` - clean across all 21 projects.
- `pnpm lint` - **zero errors** in every package (output is pre-existing warnings only).
- `pnpm check:invariants` - **90 of 90 checks pass**. `check-repo-urls` fails on local
  `.worktrees/` scratch, identically on a clean `main`; the remaining 90 were run
  individually to confirm nothing was hidden behind that abort.
- Tests are deferred to CI.

## Mockup

`docs/plans/mockups/sales-document-rule-composer.html`, states `rule-composer-conflict` and
`rule-composer-no-conflict`. Unchanged by this PR and still accurate - this restores the
implementation the mockup already described, with no behavioural difference from what was
reviewed under PR #3242.

---

## Review fixes (second commit)

Review confirmed the re-land itself is faithful in both directions: 17 of 20 files
byte-identical to the original branch, the 3 that differ are exactly the 3 `main` touched,
resolved as pure additions with zero deletions, and every review-fix commit above the
feature commit accounted for.

It also found three real issues, all **inherited from the content reviewed under
#3242** rather than introduced here, and all the same shape: a path where the check's
answer quietly becomes *nothing*, which renders as "no conflict" - the single answer this
feature exists to make impossible.

1. **`detectRuleOverlap` read the repository with the raw country.** Every sibling
   (`listRules`, `listCountryDefaults`, `resolveRouting`, `acknowledge`,
   `clearAcknowledgment`) normalises first, and `assertCountryUnconfigured`'s docblock
   asserts that invariant. The repository matches case-sensitively and `createRule`
   uppercases on write, so `{"country":"pl"}` through the admin API found zero rivals and
   returned an empty verdict - a false "no conflict". Both axes are now normalised exactly
   as `createRule` normalises what it writes.

   The condition normaliser is a **draft-safe twin**, not the existing one: a draft's
   conditions are `unknown[]` and the detector reports an unrecognised entry as
   `unreadable-condition`, which is one of its three outcomes. Narrowing or filtering
   would drop that entry and turn "could not decide" into a clean answer about a rule the
   operator did not write, so anything unrecognised passes through untouched.

2. **A failed overlap request rendered as silence.** The composer read `overlapQuery.data`
   only; on failure `verdict` was `undefined`, all three arrays fell back to `[]`, and no
   banner rendered. A fourth state (`rule-overlap-unavailable`) now says the check could
   not run. It deliberately does **not** block the save, matching the `undecided` banner
   beside it: a failed check is not evidence of a collision, and the runtime holds an
   ambiguous order either way.

3. **The reveal scroll never happened.** `scrollIntoView` ran while the dialog was still
   mounted, and Radix's `react-remove-scroll` restores the scroll position when the lock
   releases - so the scroll was a no-op or was undone a frame later, which the code's own
   comment calls "the same as no affordance at all". It is deferred to after unmount. The
   accent also now clears when the composer reopens, rather than persisting as though it
   were a property of the rule; the CSS comment is corrected to match (it said "brief",
   which the code never did).

Plus: the `buyerHasTaxId` / `orderCountry` last-one-wins asymmetry is now documented
against the conjunctive amount bounds beside it, with why both of its outcomes are
fail-safe; the rival-highlight wiring gets the coverage it shipped without, including a
test that fails against the pre-fix scroll rather than merely asserting the call happened;
and the touched files are Prettier-clean.

**Deferred, and stated rather than silently skipped:** `sales-document-rules.controller.ts`
has no spec at all, unlike its two siblings. That gap predates this PR, and writing the
first spec for the controller means standing up mocks for all its dependencies - work that
does not belong in a re-land. Tracked as #3299.

---

## Third review round: the save gate itself, plus three more silent-widening paths

The prior round found and fixed three places where the check answered with silence.
This round found the composer's own save gate was defeated the same way, one level up:
`hasConflict` derived solely from `overlapQuery.data`, and any query-key change with no
`placeholderData` resets `data` to `undefined` for the whole refetch - resetting every
verdict array to `[]` and re-enabling Save with **no banner at all**. Reachable with no
network fault: add a condition, nudge *Effective to* by a day, or type one more digit
while a collision stands.

`use-sales-document-rule-overlap-query.ts` now returns `{verdict, state}` rather than a
bare query result, with `state: known | settling | pending | incomplete | unavailable` -
the same "kept apart because a surface renders them differently" rule
`shared/hooks/use-paginated-total.ts` already states for its own five-state vocabulary.
The composer withholds Save for every state except `known` and the deliberately-advisory
`unavailable`, debounces the draft (`useDebouncedValue`, 400ms - `use-paginated-total.ts`'s
own `TOTAL_DEBOUNCE_MS` precedent, closing the a11y issue of an assertive `role="alert"`
banner re-announcing on every keystroke), and keeps the last verdict on screen through a
refetch via `placeholderData` since the button is inert for the whole of that window
regardless.

Two more paths where the check silently widened rather than refusing or reporting:

- **A draft the server would 400 was still sent to it.** `conditionsAreAnswerable` tested
  non-emptiness; the server's guards (`isDecimalAmountString` / `isCurrencyCode`) are
  strictly stricter, so typing `PLN` one character at a time sent `"P"` then `"PL"`, each
  answering 400, and with the production client's `retry: false` the first failure painted
  "we could not run the check" - a false claim about infrastructure for what was really an
  unfinished field. The composer now mirrors the server's own shapes (the #2229
  reported-is-enforced discipline) before asking.
- **A whitespace-only `orderCountry` persisted a rule that matches every order in the
  market, and the check was structurally blind to it.** `'   '` passed the DTO's old
  non-empty check (length 3), `createRule` normalised it to `''`, the rule repository's own
  `toDomain` then filtered the resulting condition out on read, and `evaluateScope` treats
  zero conditions as vacuously true - a hold against every other rule in the country,
  forever, reachable from the shipped UI with three spacebar presses (`toConditionInput`
  didn't trim `orderCountry`, unlike the amount branch beside it). The DTO now refuses
  blank-after-trim, and `createRule` re-asserts `isSalesDocumentCondition` on the
  *normalised* conditions before hashing - the durable guard, since a future normaliser
  could invalidate a different field the same way.

The detector's own comment says a rule bounding the total in two currencies "belongs in
the composer" to refuse - nothing did, so such a rule saved and matched no order at all
while the detector reported a `multi-currency-rule` sentence for every rival, blaming the
wrong side. `draftHasMixedCurrencies` closes that gate too.

Plus the smaller structural/copy findings from the same pass: `SalesDocumentRuleOverlapCheckInput`
moved out of the interface file to sit beside its sibling write type and become
barrel-reachable; the conflict banner now renders one row per rival with its own reveal
control (a joined sentence naming three rivals beside a button that opened only the first
was a claim the control couldn't honour); the revealed row now receives keyboard/AT focus
(`tabIndex={-1}` + `.focus({preventScroll: true})`, the `connection-mappings-page.tsx`
precedent) rather than only a colour-only outline Radix's own focus-return made invisible
to a keyboard user; `.rule-card--highlighted` moved after its base rule; and the doc
paragraph plus the service-interface docblock now *argue* rather than merely assert why
the composer refuses while the direct API stays advisory (an overlapping pair still
produces a safe, loud HELD order - never the silent wrong-price outcome #2610's asymmetry
was written against).

**Verification**: `pnpm -r type-check` (21/21 projects clean), `pnpm -r --filter
"./libs/**" build`, the full `apps/web` vitest run (556 files / 5956 tests), the
`libs/core` and `apps/api` jest suites scoped to `sales-document*` (all green), and
`pnpm check:invariants` (90/90). Lint run scoped to every touched file: zero errors,
pre-existing warnings only (the full-repo `pnpm -r lint` OOM-kills on this machine, so
CI is the source of truth for the rest of the tree).


norbert-kulus-blockydevs:
## Tech lead review

The re-land itself is faithful and the domain logic is sound — I tried hard to break `disjointReason` and could not. What I could break is the gate around it: the feature is defined by one principle, *an absent banner must never read as "no conflict"*, and the composer holds that principle only for the states it enumerated. The in-flight state is not one of them.

---

### BLOCKING

**[BLOCKING]** — `apps/web/.../sales-document-rule-composer-dialog.tsx:224,622` with `hooks/use-sales-document-rule-overlap-query.ts:26`

**The save gate is defeated by any keystroke.**

`disabled={createRule.isPending || connectionId === '' || hasConflict}` never consults the query's own state, `hasConflict` derives solely from `overlapQuery.data`, and the hook keys on `JSON.stringify(input)` with no `placeholderData`. A key change with no placeholder resets `data` to `undefined` for the whole duration of the refetch.

Repro, no network fault needed:

1. Author a colliding draft. `rule-conflict` renders, `rule-save-blocked` is disabled. ✔
2. Click **+ Add condition**, or nudge *Effective to* by one day, or type one more digit. The draft still collides.
3. New key ⇒ `verdict === undefined` ⇒ `conflicts = []` ⇒ `hasConflict === false` ⇒ the button re-renders as `rule-save`, **enabled**, with no banner.
4. Click it. `handleSave` re-checks nothing. The colliding rule saves.

Same hole on first open: Save is live and silent for the whole of the first check. The file's own comment above the `isError` block says *"Absence and failure must not be the same pixel"* — this is a third kind of absence, and it is the common one. Every new test `await`s `findByTestId` first, so all four only ever observe the at-rest state.

The vocabulary for this already exists one directory over: `shared/hooks/use-paginated-total.ts:75` defines `known | settling | pending | idle | unavailable` and says explicitly that *"`idle` and `unavailable` are kept apart because a surface renders them differently."* This composer collapses `pending`, `idle` and `known-and-empty` into one pixel.

**Fix:** `placeholderData: (prev) => prev` so a verdict survives a refetch, **and** include `overlapQuery.isFetching` in the disabled expression with a `Checking…` label, plus a distinct affordance for the not-yet-answerable draft. Add a test that resolves the mock manually rather than awaiting the banner.

---

**[BLOCKING]** — `sales-document-rule-composer-dialog.tsx:199-205` with `apps/api/.../dto/sales-document-condition.dto.ts:94-103`

**Typing a currency on the flagship threshold rule fires two 400s and paints a banner that blames infrastructure.**

`conditionsAreAnswerable` gates on non-emptiness:

```ts
return condition.amount.trim() !== '' && condition.currency.trim() !== '';
```

The server's guards are strict — `isCurrencyCode` is `/^[A-Z]{3}$/`, `isDecimalAmountString` is `/^\d+(\.\d+)?$/` (`sales-document-condition.types.ts:108-115`) — and `SalesDocumentConditionDto.toDomain` throws `BadRequestException` on either. The controller calls `toDomain` on the overlap path too (`sales-document-rules.controller.ts:122`).

So on the shape of the shipped PL starter template, typing `PLN` sends `"P"` then `"PL"`, each a 400. The production `QueryClient` has `retry: false`, so `isError` flips on the first failure and the operator reads *"We could not run the overlap check / This rule has not been compared against the others in this market."* `450.` does the same. That is a false statement about infrastructure when the real cause is an unfinished field — one state rendering as a different state, which is the class of defect this PR exists to remove.

**Fix:** mirror the server's own guards in `conditionsAreAnswerable` rather than testing emptiness (the reported-is-enforced discipline of #2229), and reserve the unavailable banner for a genuine failure — `ApiError` already exposes `isServerError()` / `isNetworkError()`.

---

### IMPORTANT

**[IMPORTANT]** — `libs/core/.../sales-document-rules.service.ts:134-142,172-177`, `apps/api/.../sales-document-condition.dto.ts:77`, composer `:123`

**A whitespace-only `orderCountry` persists a rule that matches every order in the market, and the new check is structurally blind to exactly that input.**

I traced the full chain against the code: the DTO accepts `'   '` (length 3 > 0); `createRule` validates the **raw** conditions, then `normaliseConditionCountries` trims to `''`, and nothing re-validates before hashing and writing; `isSalesDocumentCondition` requires `value.length > 0`, so `SalesDocumentRuleRepository.toDomain:114` filters the condition out on read, leaving `conditions: []`; and `evaluateScope:166-182` never enters the loop, so `allTrue` stays `true` and the rule matches every order. On a market with any other rule that is a permanent `conflicting-rules-equal-priority` hold.

Reachable from the shipped UI with three spacebar presses: `toConditionInput:123` does **not** trim `orderCountry`, unlike the amount branch two lines below which trims both fields — and `conditionsAreAnswerable` returns `false` for whitespace, so the overlap query is *disabled*, no banner of any kind renders, and Save stays enabled. The one rule shape guaranteed to hold every order is precisely the one this check cannot see.

To be fair about attribution: `normaliseConditionCountries` and its use in `createRule` are context lines in this diff, so the **persistence** half predates this PR. The **silence** half is new, and the durable fix is one line here — re-assert `isSalesDocumentCondition` on the *normalised* array in `createRule` before hashing, so no normaliser can invalidate a condition on its way to the column. Plus trim in `toConditionInput` for parity with the amount branch.

**[IMPORTANT]** — composer `:622` vs `docs/architecture-overview.md`

**The UI hard-blocks; the doc paragraph this PR adds says the write path is deliberately ungated. Both are in the diff.**

> *"It is a READ and the write path is deliberately NOT gated on it… an operator may still knowingly save an overlapping pair through the API."*

`createRule` indeed only calls `assertNoConflict`, which refuses on an identical `conditionsHash` — an overlapping but non-identical pair saves cleanly over HTTP. So the freedom the doc grants is exercisable only by curl, and there is no way for an operator to *knowingly* do it. That inverts the repo's usual asymmetry, which #2610 settled in the other direction in words this repo keeps quoting (*"the form is not the only way in… curl and MCP bypass the browser entirely"*).

Two defensible resolutions, and the choice is yours — but as shipped the browser gate and the API gate differ by accident of client rather than by intent. Either gate `createRule` with an explicit `acknowledgeOverlap` escape hatch (the `acknowledgeAboveRecommended` shape in `operational-settings`, or #1842's soft-block with a required acknowledgement), or keep it advisory and say in the doc paragraph *and* the composer copy that the UI is deliberately stricter than the API.

**[IMPORTANT]** — `use-sales-document-rule-overlap-query.ts:23-32`

**No debounce on a per-keystroke backend query.** `staleTime: 0` plus a key over the live draft means every character typed into `amount`, `currency` or `stringValue` mints a cache entry and a `findByCountry` read. `shared/hooks/use-debounced-value.ts` exists for this and has ten call sites; `use-paginated-total.ts:20` states it as its *first* of four rules with `TOTAL_DEBOUNCE_MS = 400` as the precedent.

There is an a11y consequence: the conflict banner is `Alert tone="error"` ⇒ `role="alert"` (assertive), and because `hasConflict` goes false on every key transition, the sentence is re-announced at a screen-reader user once per character. This is also what holds the BLOCKING window above open on essentially every keystroke.

**[IMPORTANT]** — `libs/core/.../sales-document-rules.service.interface.ts:124`

**`SalesDocumentRuleOverlapCheckInput` is in the wrong file and is not barrel-reachable.** Its sibling — the input type for `createRule` on the same interface — lives in `domain/types/sales-document-rule-write.types.ts:14`, which the barrel `export *`s. This one is the only non-interface export in the interface file, and the barrel does a *named* re-export of `ISalesDocumentRulesService` alone, so the type never reaches `@openlinker/core/sales-documents`. The controller compiles only because it passes an object literal and TypeScript is structural. Move it beside `SalesDocumentRuleInput`.

**[IMPORTANT]** — `sales-document-rules-list.tsx:60-70,205-219`

**"Open that rule" is unusable without sight.** The affordance closes the dialog, scrolls, and applies an `outline` — a colour-only cue. Radix returns focus to the `+ Add rule` trigger, so a keyboard or screen-reader user ends up where they started with no indication anything happened. The style guide requires visible focus states and status conveyed by more than colour. `connection-mappings-page.tsx:430` is the in-repo precedent: `tabIndex={-1}` on the target plus `.focus({ preventScroll: true })` inside the same `requestAnimationFrame` you already defer to. Also consider relabelling — nothing is opened.

**[IMPORTANT]** — PR body, the fidelity claim

I re-ran the comparison rather than take it on trust, and the claim is conservative but off by one on both sides. At the re-land base `ab13173dc`, **17 of 20** files are byte-identical to `origin/3190-rule-overlap-detection`, and the 3 that differ are exactly the 3 named as hand-merged. The body says "16 of 19". Worth correcting, since it is the load-bearing claim of a re-land PR.

---

### SUGGESTION

- **`unreadable-condition` is unreachable through the shipped wiring**, in both directions: `SalesDocumentRuleRepository.toDomain:114` already filters malformed persisted conditions before the detector sees the row, and the controller 400s a malformed draft. What actually happens is that the rule is silently *widened* — consistently at both detector and evaluator, so nothing is unsound, but `UNDECIDED_REASON_COPY['unreadable-condition']` can never render and the spec passes only by calling the pure function directly. Either surface the drop count from the repository, or amend the header comment and the doc bullet to say the third outcome covers `multi-currency-rule` only today.
- **The check and the write disagree on date precision.** `create` truncates `effectiveFrom`/`effectiveTo` via `toDateOnly`; `detectRuleOverlap` compares the raw draft `Date`. Truncation only moves both endpoints backwards so the residual is at most one midnight instant — not exploitable, but the check answers about a rule different from the one stored.
- **`gcTime: 0`.** The docblock says the verdict "is not worth caching beyond the dialog's own lifetime" but implements `staleTime: 0`, which controls freshness, not retention. With the default 5-minute `gcTime`, reopening the composer on the same default draft renders the previous session's verdict for one round trip.
- **Query key into `salesDocumentRulesQueryKeys`,** and drop `JSON.stringify` — invalidation currently works only by prefix coincidence, and `hashKey` already serialises deterministically whereas `JSON.stringify` is insertion-order sensitive.
- **`.rule-card--highlighted` is at `index.css:4797`; `.rule-card` is at `:21667`** — the modifier sits ~16,870 lines before its base, inside the composer's block rather than the list's. Harmless today only because the two never share a property. The style guide asks for modifiers after their base rules.
- **The detector's own documented remedy is unimplemented.** `detect-sales-document-rule-overlap.ts:156` says *"Refusing a duplicate field belongs in the composer"*, and `+ Add condition` appends with no duplicate-kind guard — so a reachable `multi-currency-rule` on the *candidate* emits one `undecided` per rival, N copies of a sentence that reads as though the rival is at fault.
- **No mirror script for the two new reason vocabularies.** The context already has `check-sales-document-reason-mirror.mjs` and `check-sales-document-condition-field-mirror.mjs`, and the first one's header names this exact hazard. The drift is fail-soft here (both fallbacks are honest and tested in both directions), which is defensible — but say so, rather than leaving the reader to infer it from the two siblings.
- **`excludeRuleId` has no production caller** — the composer is create-only and there is no `updateRule`. That is the tolerated ships-ahead-of-its-consumer posture, but this same dialog's `onOpenRule` carries the opposite rule, so one sentence naming the edit flow would make the inconsistency read as deliberate.
- **~60 lines of unrelated Prettier churn** (trailing commas converging on `trailingComma: "es5"`), applied inconsistently — `detect-sales-document-rule-overlap.ts` was reformatted, its own `.spec.ts` in the same directory was not. Reviewability only, CI does not run `format:check`, but it buries the real change and invites conflicts with the neighbouring stacked branches.
- **Composer copy density:** `rule-no-conflict` renders one paragraph *per* disjoint rival, and `onOpenRule(conflicts[0].ruleId)` opens only the first while the sentence names all of them.

---

### What is correct, and non-obviously so

- **The overlap algebra is sound.** I could not construct a false negative against `evaluateSalesDocumentRules`. Opposite `buyer-tax-id` literals genuinely cannot co-hold; the `order-country` fold matches the evaluator's exactly; a currency mismatch really does make a rule unmatchable for an order in another currency; the conjunctive bound reduction (`gte` → max, `lt` → min) is the correct interval intersection with `lower >= upper` the right emptiness test; the effective-window check is inclusive on both ends like `isEffective`. The `compareDecimalAmountStrings` BigInt seam can only ever be *stricter* than the runtime `Number()` comparison, so the divergence is in the over-warn direction.
- **The window check runs before the bounds reduction**, so a window-disjoint rival is still correctly reported `disjoint` even when the candidate is unreadable. Load-bearing ordering, easy to get wrong.
- **The detector's semantics mirror the engine's, and this is checkable rather than assumed:** `evaluateScope` returns `ambiguous-rules` on `matched.length > 1` *unconditionally*, without comparing `(documentKind, connectionId)` — so warning on any double match is right. Had the engine only held on *disagreeing* matches, this detector would block harmless saves.
- **The country-scoped read is sound, not a shortcut.** `countryConfigured` short-circuits tier 3, so a `PL` rule and a `★` rule provably never compete, and `findByCountry` is exactly the candidate set.
- **The zero-sibling-edge-leaf property holds.** `detect-sales-document-rule-overlap.ts` has one import, relative and same-context; no new `ZERO_SIBLING_EDGE_LEAVES` entry needed and none added; `check-cross-context-imports` needs no new allow-list row; no `apps/web → @openlinker/core` edge.
- **`normaliseDraftConditionCountries` is genuinely draft-safe as claimed** — an unrecognised entry passes through by reference rather than being narrowed away.
- **`salesDocumentRuleWindowsOverlap` extraction is behaviour-preserving** and closes a real drift seam by sharing the body with the write-path guard.
- **Route authorization is satisfied** by the class-level `@Roles('admin')`, which the standards explicitly permit as a narrowing inherited default, and no capability guard is right here since the input carries no `connectionId`.
- **The list tests assert behaviour, not calls** — the scroll target is compared against the actual DOM node, and the comment explaining why the *timing* cannot be asserted under happy-dom is honest rather than a silently weakened assertion.
- **The re-land is clean in both directions.** Nothing dropped from the original branch, nothing lost from main in the three hand-merged files, and the four post-base commits are genuine review fixes — one of which closes a real false-"no conflict" hole on the admin API path. The "lost merge commit" `6049d62e6` turns out to be a #3189 self-merge whose only touch here is a two-line comment, present in the worktree.
- CI is green across all nine checks.

---

### Merge readiness

❌ **Request changes**

Not on data safety — the runtime still holds an ambiguous order per ADR-041, so nothing here produces a wrong document. On the PR's own stated acceptance: *refuse a colliding rule before the save* is not true on the common path, and the absence of a debounce holds that window open on essentially every keystroke. Both are small, local fixes.

### Priority fixes

1. Keep the verdict across a refetch (`placeholderData`) and include `isFetching` in the Save button's disabled state, with a distinct affordance for the not-yet-answerable draft. Test the in-flight state, not just the at-rest one.
2. Gate `conditionsAreAnswerable` on the server's own `isDecimalAmountString` / `isCurrencyCode` shapes, and reserve the unavailable banner for a genuine failure.
3. Debounce the draft before it becomes the query key (`useDebouncedValue`, 400 ms).
4. Re-assert `isSalesDocumentCondition` on the *normalised* conditions in `createRule` before hashing, and trim `orderCountry` in `toConditionInput`.
5. Decide the server-side question explicitly: an `acknowledgeOverlap` escape hatch, or one sentence saying the UI is deliberately stricter than the API.
6. Move `SalesDocumentRuleOverlapCheckInput` to `domain/types/`; move focus when a rival is revealed; move `.rule-card--highlighted` after `.rule-card`; correct the body's fidelity count to 17 of 20.

norbert-kulus-blockydevs:
## Addressed the review

Pushed `3a0d8ef52`. Both BLOCKING findings and every IMPORTANT finding are fixed; SUGGESTIONs are handled or explicitly deferred below.

### BLOCKING

**1. Save gate defeated by any keystroke** — fixed. `use-sales-document-rule-overlap-query.ts` no longer returns a bare `UseQueryResult`; it returns `{verdict, state}` with `state: known | settling | pending | incomplete | unavailable` — the exact vocabulary shape `use-paginated-total.ts` already uses for the same reason ("kept apart because a surface renders them differently"). The composer:
- withholds Save for every state except `known` (and the deliberately-advisory `unavailable`),
- debounces the draft 400ms before it becomes the query key (`useDebouncedValue`, `use-paginated-total.ts`'s own `TOTAL_DEBOUNCE_MS` precedent — this also fixes the a11y issue you flagged, since the assertive `role="alert"` banner no longer re-announces per keystroke),
- keeps the last verdict on screen through a refetch via `placeholderData: (prev) => prev`, since Save is inert for the whole of that window regardless.

Added a test that resolves the mock manually (not via `findByTestId`) to actually exercise the in-flight window, per your note that every existing test only observes the at-rest state.

**2. Typing a currency fires 400s and blames infrastructure** — fixed. `conditionsAreAnswerable` is replaced by shape-mirroring guards (`DECIMAL_AMOUNT_PATTERN` / `CURRENCY_CODE_PATTERN`, copying `isDecimalAmountString` / `isCurrencyCode` verbatim) so the query is never enabled until the draft is a shape the server would actually accept — an incomplete draft now shows a specific "Finish every condition…" hint instead of ever reaching the endpoint. `unavailable` is now reserved for a real request failure.

### IMPORTANT

**3. Whitespace-only `orderCountry` matches every order, and the check can't see it** — fixed both ends: the DTO refuses blank-after-trim (`'   '` no longer passes as non-empty), and `createRule` re-asserts `isSalesDocumentCondition` on the *normalised* conditions before hashing — the durable guard, since the same defect shape (a normaliser invalidating a condition on its way to the column) isn't specific to this one field. `toConditionInput` also trims `orderCountry` for parity with the amount branch.

**4. UI hard-blocks, doc says the API is ungated, and there's no way to *knowingly* do it** — kept advisory rather than adding an `acknowledgeOverlap` escape hatch, but now argued rather than merely asserted: the doc paragraph and the service-interface docblock both state *why* — an overlapping-but-not-identical pair still produces a safe, loud HELD order (ADR-041's own correct behaviour), never the silent wrong-price outcome #2610's asymmetry was written against, so there's no legitimate case for an override. Happy to add the escape hatch instead if you'd rather that be the shape.

**5. No debounce on a per-keystroke query** — fixed, see #1 above (400ms, same constant name as the precedent).

**6. `SalesDocumentRuleOverlapCheckInput` in the wrong file, not barrel-reachable** — moved to `domain/types/sales-document-rule-write.types.ts`, beside `SalesDocumentRuleInput`.

**7. "Open that rule" unusable without sight** — fixed. The revealed row now gets `tabIndex={-1}` + `.focus({ preventScroll: true })` in the same deferred frame as the scroll (the `connection-mappings-page.tsx` precedent you pointed at). Relabelled the button `Show that rule`.

**8. Fidelity claim off by one** — corrected in the PR body: 17 of 20, not 16 of 19.

### SUGGESTION

- **`unreadable-condition` unreachable through shipped wiring** — documented in the detector's own header rather than left implicit, stating exactly which of the two undecided reasons is reachable today and why.
- **Date-precision mismatch between the check and the write path** — documented rather than "fixed": truncating in the advisory check would make it disagree with `assertNoConflict`, which compares the *untruncated* draft. Left as-is with the reasoning stated inline.
- **`gcTime: 0`** — deliberately *not* added; see the corrected reasoning in the hook's docblock (the composer never unmounts, so `gcTime` doesn't gate anything here regardless — my first pass at this justification was itself wrong, caught before it shipped).
- **Query key via `JSON.stringify`** — fixed. `salesDocumentRulesQueryKeys.overlapCheck(draft)` now takes the object directly (TanStack's own `hashKey` sorts keys deterministically); `invalidateQueries({queryKey: all})` reaches it by construction.
- **`.rule-card--highlighted` ~16,870 lines from its base** — fixed, moved to sit right after `.rule-card`.
- **Detector's own documented remedy (refuse mixed-currency drafts) unimplemented** — fixed. `draftHasMixedCurrencies` gates the check and the save, with its own hint.
- **No mirror script for the two reason vocabularies** — the fail-soft-in-both-directions posture is now stated explicitly in both copy tables' own comments, rather than left to be inferred by comparing against the stricter `check-sales-document-reason-mirror.mjs`.
- **`excludeRuleId` has no production caller** — left as shipped-ahead-of-its-consumer, now with a comment naming `onOpenRule` as the eventual caller so the asymmetry reads as deliberate.
- **~60 lines of unrelated Prettier churn** — not touched; out of scope for this fix pass, and doesn't affect correctness.
- **Copy density (one paragraph per disjoint rival)** — not changed; each clear reason is a distinct fact about a distinct rival, unlike the conflict banner where the fix in #7 above applies.

### Also simplified while in there

The joined-sentence `describeSalesDocumentOverlapConflict` (naming every rival in one line) is gone — nothing calls it anymore now that the banner renders per-rival rows, and keeping a second, untested-in-production copy of the same fact felt like exactly the kind of drift this feature exists to prevent. Replaced by `describeSalesDocumentOverlapConflictRival` + the standalone `SALES_DOCUMENT_OVERLAP_CONSEQUENCE` constant.

### Verification

- `pnpm -r type-check` — 21/21 projects clean.
- `pnpm -r --filter "./libs/**" build` — clean.
- `apps/web` full vitest run — 556 files / 5956 tests passing.
- `libs/core` + `apps/api` jest, scoped to `sales-document*` — 216 + 26 tests passing (confirmed `sales-document-rules.controller.ts` still has no spec, as your review noted and the PR already stated as a deliberate, separate follow-up).
- `pnpm check:invariants` — 90/90.
- Lint run scoped to every touched file (full-repo `pnpm -r lint` OOM-kills `eslint` on `apps/web` on this machine) — zero errors, pre-existing warnings only.

piotrswierzy (approved):

**Verdict: ✅ Approve.** A faithful re-land plus three rounds of genuine hardening. Every load-bearing claim in the description checks out against the source; my one finding is about the PR title, not the code.

### Verified, not taken on trust

- **The arithmetic is BigInt throughout.** `detect-sales-document-rule-overlap.ts:189,196,235,250` route every amount comparison through `compareDecimalAmountStrings` — no `parseFloat`, no `Number()` on an amount anywhere in the file. That is the #3189 requirement: amounts round-trip through `jsonb` as decimal strings precisely so `conditionsHash` stays stable, and a float comparison here would have been the first crack in it.
- **`conditionsHash` is undisturbed.** The only normalisation this PR adds to the write path is a *re-assertion* (`isSalesDocumentCondition` on the already-normalised array), not a new transform — `normaliseConditionCountries` already uppercased `orderCountry` on `main`. No previously-valid rule hashes differently, so the unique index is untouched.
- **The 409 write-path guard survives the refactor intact.** `rangesOverlap` was not deleted but *promoted* to the exported `salesDocumentRuleWindowsOverlap` (`:126-134`), body byte-identical, and `createRule` still calls it at `:544` before throwing `SalesDocumentRuleConflictException` at `:551`. The docblock's reason for exporting rather than duplicating — "two answers to it would eventually disagree" — is the right instinct and is why this refactor is safe rather than merely tidy.
- **Three outcomes, and the third never reads as reassurance.** `describe-sales-document-overlap.ts:98` falls back to *"this version could not compare the two"* for an unmapped undecided reason, and `:79` still names the rival for an unmapped disjoint reason while stopping short of explaining why. Both fail toward saying less, never toward saying "no conflict".
- **The blank-`orderCountry` fix is two layers, correctly.** The DTO refuses blank-after-trim at the boundary (`sales-document-condition.dto.ts:77-88`) *and* `createRule` re-asserts on the normalised array as the durable guard. That is the #2610 discipline applied properly — the form is not the only way in, so the boundary check alone would not have held. The severity is worth restating: `'   '` passed the old length check, normalised to `''`, was filtered out by `toDomain` on read, and left a rule with zero conditions that `evaluateScope` treats as vacuously true — i.e. a permanent hold against every other rule in the market, reachable from the shipped UI with three spacebar presses.
- **The absent mirror script is argued, not overlooked.** `describe-sales-document-overlap.ts:51-60` explains why this copy table is deliberately *not* held by a `check-*-mirror.mjs` the way `SalesDocumentGateBlockReason` is: the drift is fail-soft in both directions and asserted as such by the module's own spec, whereas a missing gate-reason entry renders a raw enum value at an operator. The FE types keep `reason: string` rather than restating the core union, so there is genuinely nothing to mirror. Correct call, and correctly justified rather than silently made.
- **`normaliseDraftConditionCountries` passes unrecognised entries through untouched**, which is what preserves `unreadable-condition` as a reportable outcome instead of filtering it into a clean answer about a rule the operator did not write.

The third-round finding — that `hasConflict` derived from a bare `overlapQuery.data`, so any query-key change reset every verdict array to `[]` and re-enabled Save with no banner, reachable by nudging a date or typing one more digit — is the most valuable thing in the PR. The five-state `known | settling | pending | incomplete | unavailable` replacement follows `use-paginated-total.ts`'s stated precedent for the same reason (states kept apart because a surface renders them differently).

### Findings

**[SUGGESTION]** — PR title / commit subject
> "refuse a colliding rule before the save" reads as a write-path gate, and the API does not refuse. Property 4 in the body is explicit that `POST /sales-documents/rules/overlap-check` persists nothing and refuses nothing, and the controller's own description at `sales-document-rules.controller.ts:104-114` says so too — only the *composer* withholds Save. The asymmetry is defensible and is argued: an overlapping pair still produces a loud HELD order, never a silent wrong-price outcome, which is the distinction #2610's rule was written against. But the subject line is the permanent record and it carries `Closes #3190`, so something like "refuse a colliding rule in the composer, before the save" would stop a future reader assuming a server-side gate exists.

**[SUGGESTION]** — `apps/api/src/sales-documents/http/sales-document-rules.controller.ts`
> The body already flags that this controller has no spec at all, unlike its two siblings, and defers it as out of scope for a re-land. Agreed on the scoping — but the new `rules/overlap-check` route ships with its own DTO-level validation and a `assertValidCountryParam` call, and that is now the second untested surface on the file. Worth the issue the body promises rather than leaving it at a paragraph.

**[Positive]** — the re-land accounting is checkable rather than asserted: 17 of 20 files byte-identical to the original branch, the 3 that differ named with which `main` change moved them, and the `index.css` conflict against #3243 resolved by keeping `main` and re-inserting only the one block. That is the right way to present a squash-loss recovery.

### Merge readiness

✅ **Approve.** All nine checks green, MERGEABLE, no migration touched, and the three hardening rounds each closed a path where the check's answer silently became "no conflict" — which is the one failure mode this feature exists to prevent. The two suggestions are follow-ups, not conditions.

norbert-kulus-blockydevs:
Addressed both findings from the review — no code changes needed, both were metadata/process fixes:

1. **PR title** — renamed to "refuse a colliding rule in the composer, before the save" so the merge record doesn't imply a server-side gate that doesn't exist. Property 4 (the API stays advisory) is unchanged and still argued in the body.
2. **Controller spec gap** — filed #3299 to track adding the first spec for `sales-document-rules.controller.ts` (including the new `overlap-check` route's DTO validation + `assertValidCountryParam`), and linked it in the PR body in place of the bare "worth its own issue" promise.

PR #3255, opened 2026-09-15
feat(inventory): expandable grouped table (per-row inventory_items detail)

### LINKED ISSUE #3074: feat(inventory): expandable grouped table (per-row inventory_items detail)

jakubretajczykBD:
## Summary
- Renders `report.groups[]` as a `DataTable` with product (name + sku, #3239), variant, location and connection columns plus row/live-row counts, expandable to a per-row detail table (id, availableQuantity, reservedQuantity, isStale, updatedAt).
- Stale rows are visually muted and get a distinct `StatusBadge`. Falls back to an `EmptyState` when `groups[]` is empty.
- Display enrichment (#3239): `productName`/`sku` fall back to the raw id when the product could not be resolved; `locationName`/`connectionName` fall back to the raw id when resolved but unnamed, and render a dedicated "Not backfilled" label for the null/`'legacy'` connection sentinel (#2317) and a dash for a null `locationId` (ADR-058 decision 2, the master declines to locate) — never a fabricated name for either.

## Context
Third sub-issue of epic #3071, stacked on #3252 (the KPI strip). Picked up #3073's review fix (dropping raw issue numbers from operator-facing copy) via rebase — verified no similar leaks in this PR's own additions (the `#2317`/`#2325`/`#3239` references left in this diff are all in code docblocks, never rendered JSX text).

⚠️ **Stacked PR** — base branch is `3073-duplicate-positions-kpi-strip` (#3252), not `main`. Merge #3252 first.

## Test plan
- [ ] `pnpm type-check`
- [ ] `pnpm lint`

Closes #3074
Part of #3071


piotrswierzy (approved):

**Verdict: ✅ Approve.** A clean incremental step that delegates the hard part to the right place.

### What is right

- **Expansion a11y lives in the primitive, not the page.** The page supplies `expandable={{ renderDetail, toggleLabel }}` and `DataTable` (#1620) owns the `aria-expanded` / `aria-controls` wiring and the detail-panel id base. That is the UI-library policy applied correctly — the wrapper owns behaviour and accessibility, the caller owns content (`docs/frontend-architecture.md` § UI Library Policy).
- **`toggleLabel` is descriptive rather than generic** (`Expand rows for product {name}`), so a screen-reader user in a table of near-identical rows gets a distinguishable control rather than twenty "Expand" buttons.
- **`cardView` with `collapsibleDetail` ships alongside**, so the mobile surface is not an afterthought — the same detail renderer serves both, which is what keeps them from drifting.
- **`rowKey` is the full four-column position key** (`productId:variantId:locationId:sourceConnectionId`, NULLs collapsed to `''`), matching the grouping semantics `docs/operations/inventory-duplicate-positions.md` § What counts as a duplicate defines. A key built from fewer columns would collide two legitimately-distinct groups.
- **46 lines of CSS, zero raw hex** — every colour goes through a token, per `docs/frontend-architecture.md` § Design tokens.

### Findings

**[SUGGESTION]** — `apps/web/src/pages/inventory/duplicate-positions-page.tsx`
> `GroupRowDetail` is passed to both `expandable.renderDetail` and `cardView.detail`. That is the right call, but it means a future change to the desktop panel silently changes the mobile card too. Worth a one-line comment saying the sharing is deliberate, so the next reader does not "fix" it by forking them.

**[Note on CI]** — the check list carries both `CANCELLED` and `SUCCESS` entries for every job, i.e. a superseded run plus a completed one. The latest run is green; the cancelled set is a re-run artefact, not a failure. Flagging it only so the mixed list is not read as a red build.

### Merge readiness

✅ **Approve.** Stacked on `3073-duplicate-positions-kpi-strip` — merge the chain bottom-up.

piotrswierzy (approved):
# PR Review — #3255 `feat(inventory): expandable grouped table (per-row inventory_items detail)`

## Step 1 — Context

Third slice of epic #3071, stacked on #3252. Adds four display cells, `GROUP_COLUMNS`, a nested `GroupRowDetail` table, the `DataTable` mount with `expandable` + `cardView`, an `EmptyState`, and ~46 lines of CSS.

Baselines checked on `main`: `shared/ui/data-table.tsx` (the `DataTableColumn` / `DataTableExpandable` / `DataTableCardView` contracts and the existing `.data-table__detail*` CSS at `index.css:4312-4318`), the backend group contract (`libs/core/src/inventory/infrastructure/persistence/repositories/inventory.repository.ts:719-750`), and the FE types from the base branch (`features/inventory/api/inventory.types.ts:76-105`).

## Step 2 — Findings

### Positive observations

- **The three enrichment cells never fabricate a name, and the two sentinel states are told apart from a genuine absence.** `ConnectionCell` (`:88-93`) collapses `null` and the #2317 `'legacy'` sentinel onto one honest *"Not backfilled"*; `LocationCell` (`:77-82`) renders a dash for `locationId === null` rather than inventing a default location, which is ADR-058 decision 2 rendered correctly ("`locationId IS NULL` means the MASTER declines to locate its stock — never a default location"). `ProductCell` (`:57-70`) falls back to the raw id. Getting any one of these wrong would put a fabricated name on the exact diagnostic an operator uses to decide which physical row to delete.
- **The empty state is a sound claim rather than a truncation artefact.** `maxGroups` is guaranteed `>= 1` server-side (`inventory-query.service.ts:154`), so `groups.length === 0` can only mean `groupCount === 0`. The `EmptyState` at `:293` therefore cannot assert "every position key is unique" over a truncated page. Worth stating because the epic's governing rule is exactly that an empty array must never render as a positive all-clear.
- **Expand affordance accessibility is inherited correctly, not reimplemented.** `expandable` routes through `DataTable`'s own toggle `<button>` with `aria-expanded` (`data-table.tsx:502-515`) and a per-row `toggleLabel` is supplied (`:308-312`), so the control is keyboard-operable and named. The global `@media (hover: none) and (pointer: coarse)` rule (`index.css:5772-5778`) gives it the 44 px target on a real phone.
- `<caption className="sr-only">` + `<th scope="col">` on the nested table (`:139-146`). Stale rows carry a `StatusBadge` **as well as** the opacity treatment, so colour/alpha is never the only signal (`docs/frontend-ui-style-guide.md` § Status & data surfaces).
- Mobile is genuinely handled rather than assumed: `hideBelow` on Variant/Location/Connection plus a `cardView` with `collapsibleDetail`.

### **[IMPORTANT]** — `apps/web/src/pages/inventory/duplicate-positions-page.tsx:47`, `:65`, `:79`, `:91`

> `className="text-muted"` is a **no-op**. There is no global `.text-muted` utility in `apps/web/src/index.css` — grepping the file returns only the custom property (`--text-muted`, lines 218/423) and one scoped rule, `.ai-provider-table .text-muted` (`index.css:10637`). `index.css` is the app's only stylesheet.
>
> So all four intended muted treatments render at default body colour: the two em-dashes, the SKU line, and — the one that matters — the *"Not backfilled"* sentinel, which is meant to read as a de-emphasised non-value and instead reads as a resolved connection name. The `data-table__detail-row--stale` opacity works; this does not.
>
> Fix by using an existing muted class or by adding `color: var(--text-secondary)` to the two new page-scoped classes in the same CSS block. (A handful of plugin components carry the same dead class — pre-existing, not this PR's to fix, but it is why the class looked safe to reuse.)

### **[IMPORTANT]** — `apps/web/src/index.css` (the `.data-table__detail-table` block) and `duplicate-positions-page.tsx:138`, `:151`

> Two new classes claim the shared `data-table` BEM block from a page:
> - `.data-table__detail-table`
> - `.data-table__detail-row--stale`
>
> The second is the live hazard. `.data-table__detail-row` is a **real class owned by the shared component** (`index.css:4312`, applied to the accordion's detail `<tr>` by `data-table.tsx`). `--stale` reads as a modifier of that element, but it is applied to a `<tr>` inside this page's *nested* table — a different element in a different component. A future `.data-table__detail-row--stale` on the shared element, or a descendant selector added under `.data-table__detail-row`, will silently collide.
>
> `docs/frontend-ui-style-guide.md` § Implementation rules keeps primitives self-owned; a page-specific element belongs under the page's own block, which this PR already establishes (`.duplicate-positions-product-cell`). Rename to `.duplicate-positions-detail-table` / `.duplicate-positions-detail-table__row--stale` — a two-line change with no visual delta.

### **[IMPORTANT]** — tests

> The four display-fallback rules encode published invariants — ADR-058 decision 2 (`locationId === null` is never a default location) and the #2317 `'legacy'` sentinel — and nothing in this repository asserts them. `docs/testing-guide.md` asks for a colocated test on non-trivial logic; "never fabricate a name for an unresolved reference" is exactly that.
>
> The stack defers tests to #3077 and that is a legitimate sequencing choice, so the ask is scoped: **#3077 must cover the four cell matrices explicitly** — `productName: null`, `sku: null` with a resolved name, `locationId: null` vs resolved-but-unnamed, and `sourceConnectionId` `null` / `'legacy'` / resolved-unnamed / resolved-named. If #3077's scope is already fixed, add a note there now rather than after merge.

### **[SUGGESTION]** — `duplicate-positions-page.tsx:136-160` (`GroupRowDetail`) on narrow viewports

> The nested detail table has five columns, the first of which is a full UUID in `mono-text` with no truncation, and it carries no `hideBelow` equivalent and no scroll container of its own. Inside the mobile card's `collapsibleDetail` (`:313-319`) that will exceed a 360 px viewport.
>
> `docs/frontend-ui-style-guide.md` § Responsive is explicit: *"No horizontal scrolling at any breakpoint except inside `RawPayloadPanel` and a table's own column-overflow area (`.data-table__container`)"* — and this nested `<table>` is not inside that container on the card path. Either wrap it in an `overflow-x: auto` box, restack it as a definition list below the mobile breakpoint, or shorten the row id with the existing `shortenId()` helper.

### **[SUGGESTION]** — `duplicate-positions-page.tsx:125`, `:132`

> `sortable: true` on **Rows** and **Live rows** sorts client-side over the *already truncated* page. Combined with #3256's "the largest N of M" banner, an ascending sort shows the smallest of the largest — a defensible but easily-misread combination on a diagnostic whose whole job is to rank by severity. Worth either a note in the column header's `title` or leaving the default descending sort unsortable.

### **[SUGGESTION]** — `duplicate-positions-page.tsx:302-305` (`rowKey`)

> The key joins four nullable fields with `''` for null: `` `${productId}:${variantId ?? ''}:${locationId ?? ''}:${connectionId ?? ''}` ``. Two distinct group keys could collide if any id ever contained a `:` — internal ids do not (`ol_<prefix>_<hex>`), so this is safe today. A non-colliding delimiter or `JSON.stringify` of the tuple would make it safe by construction rather than by a property of the id format.

## Step 4 — Final Assessment

### Summary

The substantive work — never fabricating a name for an unresolved reference, and distinguishing the two documented sentinel states from a genuine absence — is done correctly and is the hardest part of this slice to get right. The two IMPORTANT items are both mechanical: a styling class that silently does nothing (which specifically de-fangs the *"Not backfilled"* treatment), and two CSS classes squatting on the shared `data-table` block namespace next to a real class of the same name.

### Merge Readiness

🔄 **Approve with changes** — no BLOCKING issues.

### Priority fixes

1. **IMPORTANT** — `text-muted` is a dead class; give the four muted spans a real rule (`:47`, `:65`, `:79`, `:91`).
2. **IMPORTANT** — rename `.data-table__detail-table` / `.data-table__detail-row--stale` out of the shared `data-table` block.
3. **IMPORTANT** — pin the four cell-fallback matrices into #3077's scope before this chain lands.
4. **SUGGESTION** — contain or restack the 5-column nested detail table below the mobile breakpoint.

piotrswierzy (approved):
# Re-review — #3255 `feat(inventory): expandable grouped table (per-row inventory_items detail)`

Re-reviewed at `5b11fdce` against my 2026-09-18 approval. `main` re-read; line numbers are against the composed file on this branch.

## 1 — Prior findings

|---|---|---|---|
| 1 | **IMPORTANT** — `className="text-muted"` is a dead class, so the *"Not backfilled"* sentinel rendered at default body colour | **Addressed** | New page-owned rule `index.css:4427` (`.duplicate-positions-muted { color: var(--text-secondary) }`), applied at all four sites — `:47` (variant dash), `:65` (sku), `:81` (location dash), `:93` (*Not backfilled*). `grep text-muted` over the page file returns nothing. The CSS comment at `4422-4426` records why the global class was dead, which is the right place for it. |
| 2 | **IMPORTANT** — `.data-table__detail-table` / `.data-table__detail-row--stale` squatting on the shared `data-table` BEM block beside the real `.data-table__detail-row` (`index.css:4312`) | **Addressed** | Renamed to `.duplicate-positions-detail-table` / `.duplicate-positions-detail-table__row--stale` (`index.css:4376-4406`, page `:149`, `:170`), with the collision rationale kept at `4369-4375`. No `data-table__detail*` selector is introduced by this PR any more. |
| 3 | **IMPORTANT** — pin the four cell-fallback matrices (ADR-058 decision 2, the #2317 `'legacy'` sentinel) into #3077's scope before the chain lands | **Not addressed** | #3077's body is unchanged: *"Add colocated tests covering the KPI strip, expand interaction and truncated state."* None of the four matrices is named. This was explicitly *"add a note there now rather than after merge"*. |
| 4 | **SUGGESTION** — 5-column nested table overflows inside the mobile card's `collapsibleDetail` | **Addressed** | `.duplicate-positions-detail-table__scroll { overflow-x: auto }` (`index.css:4376`), wrapping both the desktop and card paths from one component (`:148`). See the new IMPORTANT below — the container itself needs the primitive's a11y affordances. |
| 5 | **SUGGESTION** — `sortable` sorts the already-truncated page | **Addressed** | `title` on both header spans (`:125`, `:136`). |
| 6 | **SUGGESTION** — `rowKey` joins nullable fields with `:` | **Addressed** | `JSON.stringify([...])` (`:340-342`) — non-colliding by construction rather than by a property of the id format. |

### Positives re-confirmed

- The four display cells still never fabricate a name: `ProductCell` falls back to the raw id (`:57-70`), `LocationCell` renders a dash for `locationId === null` (ADR-058 decision 2) rather than inventing a default location (`:77-86`), `ConnectionCell` collapses `null` and `'legacy'` onto one honest *"Not backfilled"* (`:90-96`) — and now that muting actually renders, the sentinel reads as a non-value, which was the whole point of finding 1.
- `EmptyState` remains a sound claim: `maxGroups >= 1` is still enforced server-side (`inventory-query.service.ts:154-160`, and it throws rather than clamps), and `findDuplicatePositions` short-circuits to `truncated: false` with an empty `groups` when `groupCount === 0` (`inventory.repository.ts:710`), so `groups.length === 0 ⟺ groupCount === 0` still holds.
- Expand affordance still inherited, not reimplemented — `DataTable`'s own toggle `<button>` with `aria-expanded`, plus a per-row `toggleLabel` (`:345-349`). `DataTableColumn.header` really is `ReactNode` (`data-table.tsx:35`), so the new `title`-bearing header spans are within contract.
- Stale rows still carry a `StatusBadge` **as well as** the opacity treatment (`:173-181`), so alpha is never the only signal.

## 2 — Findings on the new commit

### **[IMPORTANT]** — `duplicate-positions-page.tsx:148` / `index.css:4376-4378` — the new scroll region is not keyboard-operable

> The overflow fix introduces a bare `<div className="duplicate-positions-detail-table__scroll">` with `overflow-x: auto` and nothing else. The shared primitive it is standing in for does four things this one does not (`data-table.tsx:707-719`):
>
> ```tsx
> tabIndex={plainScrollable ? 0 : undefined}
> role={plainScrollable ? 'region' : undefined}
> aria-label={... `${caption} (scrollable)` ...}
> ```
> plus a focus ring (`index.css:3858-3861`, `.data-table__container:focus-visible`).
>
> That is not decoration: there is **no focusable element anywhere inside** this nested table — five `<td>`s of spans and `StatusBadge`s, no links, no buttons — so with the region itself unfocusable a keyboard-only user has no way to scroll it and the overflowed columns (Reserved / Status / Updated on a 360 px card) are unreachable. WCAG 2.1.1. The existing container in the same design system is the pattern; copy its four attributes onto the wrapper and add the `:focus-visible` shadow to the new class.
>
> The nested `<caption className="sr-only">` (`:150`) already supplies good text for the `aria-label`.

### **[SUGGESTION]** — `:329-333` — `liveRegion="off"` on the `EmptyState` is now inconsistent with its sibling

> `feedback-state.tsx:20-22` is explicit: set `"off"` *"when the empty state is rendered on initial page load (not as a transition from a prior loaded state)"*. This one is precisely a transition — it replaces the `LoadingState` when the scan returns clean — and #3252 has just switched that `LoadingState` to the `polite` default on the same review's advice. As it stands the page announces that it started scanning and then says nothing when the scan finishes with the all-clear, which is the one outcome a screen-reader user most wants confirmed. The `polite` default looks right here too.

### **[SUGGESTION]** — `:125`, `:136` — the header tooltip names an off-page parameter, and `title` is mouse-only

> *"Sorts only this truncated page — groups is capped by maxGroups, largest first"*. `maxGroups` has no surface on this page until #3264 adds the input, so for now it names a query parameter the operator cannot see — a milder instance of the class of leak finding 1 on #3252 removed. And `title` never appears on touch and is announced inconsistently by screen readers; the caveat is real enough that a visible note under the table, or `aria-describedby` on the sort button, would carry it better. Low priority — this is the remedy I suggested, so it is an improvement on where it was.

### Not findings (checked)

- Every token in the new CSS resolves against a real declaration: `--border-subtle`, `--font-mono`, `--tracking-caps`, `--text-muted`, `--text-secondary`, `--space-2`, `--space-3`. Worth stating explicitly, since a mistyped token would reproduce finding 1 exactly — silent no-op styling.
- `.sr-only`, `.mono-text`, `.tabular` all exist as real global rules.
- Raw `0.625rem` / `0.75rem` / `0.125rem` — `index.css` defines no font-size or sub-space tokens; consistent with the file.

### Inherited, not yours to fix

`Test (web)` cannot go green on this branch until #3252 fixes it: adding `useIsAdmin()` to both query hooks broke the base branch's hook suites (`useSession must be used within a SessionProvider`, job `105574150484`). Merge bottom-up, and do not merge this until #3252 is green.

## 3 — Final assessment

### Summary

All three IMPORTANT code findings and all three suggestions are genuinely fixed in the source, not just in a reply — the muted treatment is a real rule applied at all four sites, the two squatting class names are gone with the rationale preserved at the point of temptation, and the nested table now has an overflow container. What the overflow fix did not bring across is the four attributes that make the shared table's own scroll container reachable without a mouse, and inside this particular table there is no other focusable element to reach the hidden columns by.

### Merge readiness

🔄 **Approve with changes** — nothing BLOCKING in this diff. The stack is held by #3252's red `Test (web)`.

### Priority fixes

1. **IMPORTANT** — give `.duplicate-positions-detail-table__scroll` the `tabIndex={0}` / `role="region"` / `aria-label` / focus-ring treatment `data-table.tsx:707-719` uses.
2. **IMPORTANT** — carry-over: name the four cell-fallback matrices in #3077 before this chain lands.
3. **SUGGESTION** — drop `liveRegion="off"` from the `EmptyState` for parity with the now-`polite` `LoadingState`.

piotrswierzy (approved):
## Tech Lead review — `feat(inventory): expandable grouped table (per-row inventory_items detail)`

Good slice, and it surfaces something worth a lot more than this PR — see the last section.

### ✅ Not squatting on the shared `data-table` BEM block

```
deliberately NOT `data-table__detail-table` / `data-table__detail-row--stale`
— those names would squat on the shared `data-table` BEM block, and
`.data-table__detail-row` is a REAL class owned by `data-table.tsx`
```

I checked — `.data-table__detail-row` does exist and is the accordion's own detail `<tr>`, a different element from this nested table's rows. A `--stale` modifier under that name would read as belonging to the shared component and would eventually collide for real. Naming the page-owned block separately is right, and the comment is what stops someone "tidying" it back later.

### ✅ Each null is rendered as what it means, not as a blank

Four small cell components, each with the domain rule in its docblock:

- `ConnectionCell` renders **"Not backfilled"** for both `null` and `'legacy'` — not an em-dash, because those two aren't "no connection", they're "provenance unknown, #2317 hasn't reached this row". An operator looking at a duplicate-positions report needs that distinction, since it's the other readiness condition on the same page.
- `LocationCell` renders `—` for `null` per ADR-058 decision 2, and the raw id (never a fabricated name) when an id resolved to no name.
- `ProductCell` falls back to the raw mono id when `productName` is null, and only shows `sku` when present — which is exactly the `sku: null` two-meanings rule from #3239 applied at the render.

### ✅ Accessibility on the nested table

`tabIndex={0}` + `role="region"` + `aria-label` on the scroll container, matched by a `:focus-visible` ring copied from `.data-table__container`. The reasoning given is the right one: no cell in this table is focusable (spans and badges only), so without it a keyboard-only user simply cannot reach the overflowed columns at narrow widths. `sr-only` caption and `scope="col"` on the headers round it out.

`EmptyState` left at the default `polite` live region, with the note that it *replaces* a LoadingState as a transition so the all-clear should announce — correct application of the rule, and the opposite call from a virgin-landing empty state.

### ✅ Honest sort affordance

The `title` on the two sortable headers — *"Sorts only this truncated page — groups is capped by maxGroups, largest first"* — is the right disclosure. Client-side sorting over a server-truncated page is a genuine footgun, and saying so beats either disabling the sort or letting an operator believe they're sorting the whole table.

### 🔵 Finding worth escalating beyond this PR — `.text-muted` is dead across the whole app

The CSS comment claims:

> *`.text-muted` is a dead global class (no such rule exists outside one scoped `.ai-provider-table` override)*

I verified it, and it is both true and bigger than the comment suggests. `apps/web/src` has exactly **one** stylesheet, and `.text-muted` appears in it exactly once:

```
apps/web/src/index.css:10660:  .ai-provider-table .text-muted {
```

…against **218 occurrences** of `text-muted` across `apps/web/src/**/*.tsx`. So outside `.ai-provider-table`, every one of those renders at default body colour. That includes several surfaces in the PRs landing alongside this stack — the returns worklist's group descriptions and empty messages, the sales-documents list's "Showing N", the order-detail panels.

Introducing `.duplicate-positions-muted` locally is the right call **for this PR** — minimal, scoped, and it doesn't quietly change 218 other renders as a side effect. But the repo-wide fix is one line:

```css
.text-muted { color: var(--text-secondary); }
```

…and it needs its own PR, because adding it makes ~216 elements change colour at once and that deserves to be reviewed as a visual change rather than smuggled in here. Please file it — and consider whether `scripts/check-css-structure.mjs` (#2674) could grow a companion check for class names referenced from TSX with no rule backing them, since this is precisely the failure mode it describes: *"the styles are… never applied"*, invisible to lint, type-check and every class-presence assertion.

### 🟢 SUGGESTION

`rowKey={(g) => JSON.stringify([g.productId, g.productVariantId, g.locationId, g.sourceConnectionId])}` — correct (the four columns are the position key), but `JSON.stringify` on every row on every render is a heavier way to say it than a joined string with a separator that can't appear in an id. Cosmetic.

### Notes

- Card view mirrors the desktop detail via the same `GroupRowDetail`, so the two can't drift.
- `toggleLabel` names the product rather than saying "Expand row", which is what a screen-reader user needs when several toggles are on screen.
- CI: six green, `Build` / `Integration Tests` queued at review time.

**Verdict:** ✅ Approve.

PR #3256, opened 2026-09-15
feat(inventory): truncated-report banner state

### LINKED ISSUE #3075: feat(inventory): truncated-report banner state
## Problem / Context

`truncated: true` means group DETAIL was capped by `maxGroups` while the KPI totals still cover the whole table — nothing communicates this distinction today.

## Proposed Solution

Render a banner stating the totals are exact but the listed groups are the largest N only.

## Classification

**Type**: Frontend

## Dependencies

- Part of #3071

## Assumptions

- None

## Acceptance Criteria

- [ ] A truncated response shows the banner; a non-truncated one does not

jakubretajczykBD:
## Summary
- Renders an info-tone `Alert` when `report.truncated` is `true`, stating that the KPI totals cover the whole table while the group table below lists only the largest `maxGroups` groups. Absent when `truncated` is `false`.

## Context
Fourth sub-issue of epic #3071, stacked on #3255 (the expandable table). Picked up #3073's review fix (dropping raw issue numbers from operator-facing copy) via rebase — verified this PR's own banner text carries no issue-number leaks either.

⚠️ **Stacked PR** — base branch is `3074-duplicate-positions-expandable-table` (#3255), not `main`. Merge #3255 first.

## Test plan
- [ ] `pnpm type-check`
- [ ] `pnpm lint`

Closes #3075
Part of #3071


jakubretajczykBD:
## Tech review

**Summary**: Tiny, self-contained addition (7 lines, no new files, no CSS) that renders an info banner when the API's `truncated` flag is set. The claim in the description — that it inherited #3073's copy fix via rebase — checks out: no hardcoded issue numbers in the new text. The wording is accurate against the backend contract (`truncated: groups.length < groupCount`, `groups` capped to `maxGroups`, largest-first), so "largest {report.groups.length} group(s)" correctly reflects the actual cap that was applied, not a hardcoded constant.

### Issues

**[SUGGESTION]** — `apps/web/src/pages/inventory/duplicate-positions-page.tsx:254`
> Minor grammar: `the largest 100 group(s)` reads a little awkward when `maxGroups` is its typical default (100) — "group(s)" pluralization hedging is more natural when the count could plausibly be 1, which in practice it can't be here (`DEFAULT_DUPLICATE_POSITION_GROUPS`). Not worth blocking on — it matches the existing "row(s)"/"group(s)" pattern already used elsewhere in this same file, so it's at least internally consistent.

**[SUGGESTION]** — general
> Couldn't run `tsc`/`pnpm lint` locally (no `node_modules` in this worktree), but the change is a pure conditional-render addition using only fields already present on `DuplicatePositionsReport` (`truncated`, `groups`), so there's no type risk. Worth confirming CI green before merge as usual.

No BLOCKING or IMPORTANT issues — architecture (page-level composition of an existing `Alert` primitive, no raw fetch, no new state), naming, and copy conventions are all correctly followed.

### Verdict

piotrswierzy (approved):

**Verdict: 🔄 Approve with changes.** Seven lines, and the substance is right — a truncated report must say it is truncated. One wording fix, because at this size the wording *is* the PR.

### What is right

Gating on `report.truncated` rather than inferring truncation from `groups.length === maxGroups` is correct: the backend reports the fact explicitly and `docs/operations/inventory-duplicate-positions.md` is emphatic that `maxGroups` bounds the returned **detail only** while `groupCount` / `rowCount` / `excessRowCount` are always computed over the whole table. Reading truncation off a length comparison would also mis-fire on a report that happens to hold exactly `maxGroups` groups.

`tone="info"` is right too — a truncated detail list is not a fault, and the readiness verdict above is unaffected by it.

### Findings

**[IMPORTANT]** — `apps/web/src/pages/inventory/duplicate-positions-page.tsx:254-259`
> Two problems in one sentence, both fixable with one interpolation.
>
> **"The totals above"** — the KPI strip carrying `groupCount`, `rowCount` and `excessRowCount` renders *below* this banner (`<div className="ds-grid ds-grid--4">` immediately follows it). What is actually above is the readiness checklist, which happens to carry `groupCount` inline, but only on its not-ready branch. The pointer is therefore either wrong or accidentally right depending on state.
>
> **The banner names one number, not two.** It says "only the largest {groups.length} group(s)" and never states the total it is a subset of. The repo's standing rule for a truncated listing is to state both and let truncation never hide scale — `buildDestinationRoutingDetail` does exactly this (`N unresolved destination(s): …, and M more`), and `docs/architecture-overview.md` § 22 records the returns panel "discloses a truncated page with both numbers rather than trimming quietly".
>
> Suggested: `The table below lists the largest {report.groups.length} of {report.groupCount} duplicate group(s). The KPI totals below cover the whole table.` That is self-contained, so it survives any later reordering of the page.

**[SUGGESTION]** — same block
> The truncated state is the one an operator is most likely to act on incorrectly (running the #2325 gate off a partial list). Since `maxGroups` is an input on this page as of #3264, a sentence pointing at raising it — or at the runbook's instruction to read `groupCount` and nothing else — would close the loop. Optional here; #3264 adds the input.

### Merge readiness

🔄 **Approve with changes** — the copy fix is a one-liner and does not need a re-review. Stacked on `3074-duplicate-positions-expandable-table`; merge bottom-up. The mixed `CANCELLED`/`SUCCESS` check list is a superseded run plus a green one, not a red build.

piotrswierzy (approved):
# PR Review — #3256 `feat(inventory): truncated-report banner state`

## Step 1 — Context

Fourth slice of epic #3071, stacked on #3255. Seven added lines: an info-tone `Alert` rendered when `report.truncated` is true, placed between the readiness Alert and the KPI grid (`apps/web/src/pages/inventory/duplicate-positions-page.tsx:271-276`).

Baselines checked: the backend contract (`apps/api/src/inventory/http/dto/duplicate-positions-response.dto.ts:84-85` — *"True when group detail was capped by maxGroups; totals never are"*), the ordering that makes "largest" true (`libs/core/src/inventory/infrastructure/persistence/repositories/inventory.repository.ts:724,750` — `ORDER BY row_count DESC`), and `shared/ui/alert.tsx` (`tone='info'` → `role="status"`).

## Step 2 — Findings

### Positive observations

- **It discloses both numbers rather than trimming quietly** — `{report.groups.length} of {report.groupCount}` (`:273`). This is the exact behaviour `docs/architecture-overview.md` § Inventory and the wider repo rule require of a truncated report, and it is the reason this banner has to exist at all: without it the table silently under-reports while the KPI strip above it does not.
- **"Largest" is a true claim, not a hopeful one.** The repository orders `row_count DESC` in both the keys CTE and the final projection, so the retained groups really are the worst offenders. A banner asserting an ordering the backend did not guarantee would have been the defect here; it isn't.
- **`truncated: false` renders nothing, and that is correct rather than a silent absence** — the backend sets the flag only when the cap actually bit, so absence of the banner is a positive statement from the server, not an unread field.
- Uses the shared `Alert` with `tone`, no new CSS, no raw hex, no state. Correctly a derived render off already-fetched server state (`docs/frontend-architecture.md` § state ownership) — no local flag, no effect.

### **[IMPORTANT]** — `duplicate-positions-page.tsx:273-274`

> The word **"table" means two different things in two consecutive sentences**:
>
> > "The **table** below lists the largest 50 of 312 duplicate group(s). The KPI totals below cover the whole **table**."
>
> The first is the UI table; the second is the database table `inventory_items`. On a page whose other copy says *"reads inventory_items directly"* and *"Scanning inventory_items…"*, an operator can reasonably read the second sentence as "the KPI totals cover the whole [UI] table" — i.e. the precise misreading this banner exists to prevent, inverted into a claim that the totals are *also* capped.
>
> This is the one sentence on the page that must be unambiguous. Suggest: *"…The KPI totals above count every duplicate position in the database, not just the groups listed here."* Note "above"/"below" also needs to match final placement (see the suggestion below).

### **[SUGGESTION]** — `duplicate-positions-page.tsx:271` (placement)

> The banner sits **above** the KPI grid, while the table it qualifies sits **below** the grid. Both referents are therefore "below" it, which is what makes the two-sentence copy strain. Moving the banner to immediately above the `DataTable` (after the KPI grid, `:288`) would let it say "the table below" and "the KPI totals above" — each pointing at a distinct, adjacent thing — and would put the truncation warning where the truncated content actually is.

### **[SUGGESTION]** — tests

> No test asserts the two states this slice adds: `truncated: true` renders the banner with both numbers, and `truncated: false` renders nothing. Both are one-line assertions. `docs/testing-guide.md` would want them colocated; the stack defers to #3077, so ensure that issue names this banner explicitly — a truncation disclosure that silently stops rendering is indistinguishable from an untruncated report, which is the failure the banner was added to remove.

### **[SUGGESTION]** — pluralisation

> `duplicate group(s)` — the parenthetical plural is a small copy smell on a surface where the number is always available; `{n === 1 ? 'group' : 'groups'}` reads better. Same pattern appears in #3252's checklist, so if it is changed it should be changed in both.

### Note, not a finding

- `truncated: true` with `groups.length === 0` is unreachable (`maxGroups >= 1` is enforced at `inventory-query.service.ts:154`, and the service short-circuits to `truncated: false` when there are no groups at all, `:710`), so the banner can never render alongside #3255's `EmptyState`. Verified rather than assumed, because the two would contradict each other.

## Step 4 — Final Assessment

### Summary

A seven-line change that does the one thing it must do — disclose both the retained and the total group count instead of trimming silently — and it rests on a backend ordering guarantee that genuinely holds. The single concern is that the sentence carrying the disclosure uses "table" in two senses one clause apart, on a page where both senses are live; that is worth a rewrite before merge because the whole value of the banner is that it cannot be misread.

### Merge Readiness

🔄 **Approve with changes** — no BLOCKING issues.

### Priority fixes

1. **IMPORTANT** — disambiguate "the whole table" (database vs the UI table) in the banner copy.
2. **SUGGESTION** — move the banner adjacent to the `DataTable` it qualifies, and align the "above/below" wording with the final placement.
3. **SUGGESTION** — add the two render-state assertions (`truncated` true/false) to #3077's scope.

piotrswierzy (approved):
# Re-review — #3256 `feat(inventory): truncated-report banner state`

Re-reviewed at `94fdf66c` against my 2026-09-18 approval. `main` re-read; backend guarantees re-verified rather than carried over.

## 1 — Prior findings

|---|---|---|---|
| 1 | **IMPORTANT** — *"table"* used in two senses one clause apart ("the table below" / "the whole table") | **Addressed** | `duplicate-positions-page.tsx:321-325`: *"The KPI totals above count every duplicate position **in the database**. The table below lists only the largest {N} of {M} duplicate groups."* The database sense is now named as such and the UI sense is the only remaining bare "table". Both referents are unambiguous. |
| 2 | **SUGGESTION** — move the banner adjacent to the table it qualifies, and align the above/below wording | **Addressed** | The `Alert` now sits *after* the KPI grid's `</div>` and immediately before the `groups.length === 0 ? EmptyState : DataTable` ternary (`:318-327`). "KPI totals above" and "the table below" each now point at a distinct, adjacent thing, which is what made the old two-sentence copy strain. |
| 3 | **SUGGESTION** — pluralisation (`duplicate group(s)`), in both this banner and #3252's checklist | **Addressed**, both places | Banner `:325` (`{groupCount === 1 ? 'group' : 'groups'}`) and the readiness checklist `:275-276`, which this PR also fixes. |
| 4 | **SUGGESTION** — name this banner's two render states in #3077 | **Not addressed** | #3077's body is unchanged and does not name the `truncated` true/false pair. A truncation disclosure that silently stops rendering is indistinguishable from an untruncated report — which is the failure the banner exists to remove — so this is the one test the slice most needs. |

### Positives re-confirmed against current `main`

- **Both numbers are still disclosed**, not one: `{report.groups.length} of {report.groupCount}` (`:323-324`).
- **"Largest" is still a true claim.** `inventory.repository.ts:724` and `:750` both order `row_count DESC` — the keys CTE and the final projection — so the retained groups really are the worst offenders.
- **The banner still cannot co-render with #3255's `EmptyState`.** `truncated` is computed as `groups.length < groupCount` (`inventory.repository.ts:816`) and the zero-group path short-circuits to `truncated: false` with an empty `groups` (`:710`); `maxGroups >= 1` is enforced server-side and now *throws* rather than clamping (`inventory-query.service.ts:154-160`). So `truncated: true ⟹ groupCount > 0 ⟹ groups.length >= 1`, and `N < M` always holds in the rendered sentence.
- Still a pure derived render off already-fetched server state — no local flag, no effect, no new CSS, `tone="info"` (`alert.tsx` → `role="status"`), which correctly says a truncated detail list is not a fault.

### **No regression against #3264's blocker**

I blocked the sibling PR #3264 for claiming the shown groups are the **highest-exposure** ones. This banner makes no such claim: it says *"the largest"*, which is exactly what `ORDER BY row_count DESC` guarantees, and it says nothing about excess rows or exposure. Checked specifically because the copy was rewritten since the last review.

## 2 — Findings on the new commits

None. The diff is nine lines: the banner block and the checklist's pluralisation. Both are the fixes asked for, and neither introduces a new surface, state, style or dependency.

One observation, not a finding: this PR edits copy inside the readiness checklist that #3252 owns (`:275-276`). Correct here — they are stacked and the change was requested against both — but if #3252 is amended for its own blocker, re-check that this hunk survives the rebase.

### Inherited, not yours to fix

`Test (web)` cannot go green on this branch until #3252 is fixed: adding `useIsAdmin()` to the two query hooks broke the base branch's hook suites (`useSession must be used within a SessionProvider`, job `105574150484`; base PR #3251 passes the same job). Nothing in *this* diff is implicated. Merge bottom-up.

## 3 — Final assessment

### Summary

The one IMPORTANT item is properly fixed, and fixed in the way that made the sentence self-contained rather than dependent on page order — the banner was also moved so that "above" and "below" each point at one adjacent thing. Both suggestions landed too. The disclosure still states both numbers, still rests on a backend ordering guarantee that genuinely holds on current `main`, and still cannot contradict the empty state. Nothing new in this diff to raise.

### Merge readiness

✅ **Approve** — no BLOCKING and no IMPORTANT items outstanding in this PR. It is gated only by the stack: hold until #3252's `Test (web)` is green.

### Priority fixes

1. Carry-over **SUGGESTION** — name this banner's two render states (`truncated` true renders both numbers, false renders nothing) in #3077 before the chain lands.

piotrswierzy (approved):
## Tech Lead review — `feat(inventory): truncated-report banner state`

Eleven lines, and both of them are the right call.

### ✅ Gated on the server's `truncated` flag, not inferred from the arrays

```tsx
{report.truncated ? ( … ) : null}
```

rather than `report.groups.length < report.groupCount`. That's the reported-vs-derived distinction the repo is careful about elsewhere: `truncated` is the backend saying *the cap actually bit*, whereas the arithmetic would be a second, independently-drifting answer to the same question — and would misreport if `groups[]` were ever filtered for any other reason.

### ✅ The banner states both numbers, and reasserts what the KPIs mean

> *The KPI totals above count every duplicate position in the database. The table below lists only the largest N of M duplicate groups.*

Restating that the KPIs are uncapped, right where the truncation is disclosed, is the useful part. An operator who reads "showing 100 of 4,312" immediately after four KPI cards will otherwise wonder whether those cards were capped too — and on this page that matters, because `groupCount` **is** the #2325 readiness gate. Getting that wrong in either direction (believing the gate is capped, or believing the table is complete) leads somewhere bad.

`tone="info"` is right — a truncated detail view is not a warning, the report is working as designed.

### ✅ Pluralisation fixed rather than papered over

`group(s)` → `group` / `groups` in both the checklist line and the new banner. Small, but "1 duplicate position group(s)" on the screen an operator uses to decide whether a migration is safe reads as unfinished.

### 🟢 SUGGESTION

`report.groupCount === 1 ? 'group' : 'groups'` now appears twice in this file. At two it's fine; if a third arrives, a tiny local `plural(n, 'group')` helper beats a fourth copy. Not worth changing now.

### Notes

- Banner placed between the KPI grid and the table — after the numbers it qualifies, before the thing it's qualifying. Correct.
- No test in this PR; #3262 is the dedicated tests slice. Worth making sure the truncated branch is one of the cases it covers, since it's a whole render arm that only appears above `maxGroups` groups and is therefore easy to leave uncovered.
- CI: four green, `Lint` in progress and `Build` / `Test` / `Test (web)` queued at review time. Please confirm before merging.

**Verdict:** ✅ Approve.

Issue #3193, opened 2026-09-11
feat(eparagony): issue corrective invoices


norbert-kulus-blockydevs:
## Merge strategy — ✅ DIRECT TO `main`

One branch off `main`, one PR, one merge. This slice leaves `main` observably better on its own — no dead code, no half-rendered feature. If it has a dependency it is named at the bottom of this issue; wait for that to land on `main` first, then branch normally.

---

Part of #3173.

## Context
A registered fiscal receipt **cannot** be corrected; an invoice can. That asymmetry is one of the concrete reasons to give this provider a second capability.

## Scope
`CorrectionIssuer` over `eCorrectiveInvoice`. The original stays issued — a correction is a linked follow-up, never a replacement.

## Verified wire contract

Confirmed against `sandbox.eparagony.pl` on 2026-09-15: a VAT correction of a previously
issued invoice was accepted (`202`) and reported back as
`documentType: "CORRECTIVE_INVOICE"`, while the corrected original kept its own status
and its own public URL.

### The shape is a before/after pair, which is what OL already holds

`PDCorrectiveInvoice` is discriminated on `invoiceType` (`VAT` | `ADVANCE_PAYMENT` |
`SETTLEMENT`); this issue covers `VAT` only. `PDVatCorrectiveInvoice` requires
`invoiceType`, `metadata`, `correctedMetadata` and `correctingMetadata`:

- `correctedMetadata` **links the original by invoice number**, and requires
  `invoiceNumber`, `invoiceDate` and `grossSaleValue`. There is no KSeF-number or
  document-token linkage - the number the original was issued under is the join key.
- `correctingMetadata` carries the post-correction state and requires `merchantTIN`,
  `merchantName`, `merchantAddress`, `consumerName`, `consumerAddress` and
  `grossSaleValue`.
- `correctedLines` / `correctingLines` and `correctionReason` are optional.
- `metadata.invoiceNumber` is the correction's own number.

That is exactly the pair `IssueCorrectionCommand.originalDocument` already supplies: the
issued-line snapshot (#1297) is the `corrected` half and the post-correction lines are
the `correcting` half. **No new modelling is needed**, and no reconstruction from the
order's current state - which is the fallback #1297 exists to avoid.

### The original is untouched by construction

A correction is issued through the same `POST /documents` endpoint with its own
`documentToken`, its own `transactionToken` and its own `documentPublicUrl`. It does not
address, mutate or supersede the corrected document. The acceptance criterion below is
[...]

Issue #3192, opened 2026-09-11
feat(eparagony): issue invoices through the Invoicing capability


norbert-kulus-blockydevs:
## Merge strategy — 🔗 PR STACK, one merge to `main`

This issue is deliberately large and **must not produce an intermediate state on `main`** — its own body says why under `DO NOT SPLIT`. It is also too large for a single review.

Use an **integration branch** `3192-<short-kebab>` off `main`, then a **stack of PRs targeting that branch** (one per reviewable chunk, each branched from the previous). Review them individually; merge the integration branch to `main` **once**, when the whole is green.

**Never merge a stack member straight to `main`.**

---

Part of #3173.

## Context
eparagony can issue `eInvoice` and relay it to KSeF; OpenLinker only knows how to ask it for receipts, so a second provider is needed for invoices. `Invoicing` is **already** in `CoreCapabilityValues`, so no core vocabulary changes. One plugin exposing several capabilities is ordinary — PrestaShop dispatches six.

## Scope — DO NOT SPLIT
Without the manifest the adapter is unreachable; without the toggles the operator cannot enable it; without the status mapping an invoice hangs with no status. Each alone is zero value.
- `EparagonyInvoicingAdapter` implementing `InvoicingPort` over `eInvoice`
- `upsertCustomer` as an **identity echo** following the KSeF adapter's precedent — a deterministic id derived from the buyer's tax number, no network call. Neither a no-op nor a throw
- `RegulatoryStatusReader`: `OFFLINE → pending-submission`, `CONFIRMED+KSEF → cleared`, `CONFIRMED+NONE → not-applicable`
- Manifest advertises `Invoicing` — required, the connection service validates the name against the adapter's list
- Capabilities block renders two roles; **independent** enable toggles; the "enabling a role does not route anything to it" notice
- Wire the waiting state onto the **existing** `pending-submission` rendering — the order cell already renders "Awaiting submission" and the timeline already draws an active waiting node. **Do not build a second ladder**
- **Extend the neutral-vocabulary guard to `apps/web`** (rider below). This issue is the first one that puts the provider's name into front-end code that is not the plugin folder, so the guard belongs with it

## Verified wire contract

Confirmed against `sandbox.eparagony.pl` on 2026-09-15 with Documents API v3
(`X-Api-Version: 3`, spec revision `20260311`). Four documents issued: a B2B invoice
with `consumerTIN`, a B2C invoice without one, an invoice with no `eInvoicingHub`, and
[...]

PR #1293, opened 2026-07-01
feat(infakt): register Infakt plugin in API/worker + wire webhook ingestion

norbert-kulus-blockydevs:
## Summary

Stacked on #1292 (#1280, not yet merged) - this branch is based on `1280-infakt-plugin-hardening-tests` so it carries that PR's diff until it merges. Opened as **draft** for that reason.

Part of the Infakt epic (#1279). Closes #1281.

- Registers `InfaktIntegrationModule` in `apps/api/src/plugins.ts` and `apps/worker/src/plugins.ts` (mirrors the KSeF/Subiekt dual-registration pattern) so the host can resolve the `Invoicing` capability for Infakt connections
- Adds `@openlinker/integrations-infakt` to `apps/api`/`apps/worker` `package.json`, and the jest-integration moduleNameMapper entries
- Wires Infakt's third-party-native webhook (KSeF clearance relay) into OL's ingestion pipeline per ADR-021/ADR-015:
  - `InfaktInboundWebhookDecoderAdapter` (provider-keyed `infakt`): HMAC-SHA256 signature verify (no timestamp header, so no replay-window check for this provider), subscription-verification handshake detection, envelope extraction — wraps the already-tested `InfaktWebhookTranslator` from the #1280 POC rather than duplicating its crypto/parsing
  - `InfaktWebhookEventTranslatorAdapter` (adapterKey-keyed `infakt.accounting.v1`): maps `send_to_ksef_success` / `send_to_ksef_error` to the new `invoicing` inbound domain; other Infakt events dead-letter (`null`)
- Extends `InboundWebhookDecoderPort` with an optional `detectHandshake` step (runs before `verify`) so a provider's subscription-verification ping can echo a body back — `WebhookService`/`WebhookController` now thread a return value through instead of always resolving `void`. This is additive; `DefaultWebhookDecoder` and existing decoders (InPost) are unaffected.
- Adds `'invoicing'` to the closed `InboundEventDomainValues` union and a new `InboundRoutingPolicyService` case: an `invoicing` webhook event enqueues the existing `invoicing.regulatoryStatus.reconcile` job (gated on the `Invoicing` capability). No new by-id job — the webhook is a trigger, not the source of truth (mirrors the InPost/webhook philosophy already documented in `architecture-overview.md`); the scheduled reconcile still drains the full non-terminal frontier regardless.

No `WebhookProvisioningPort` (Infakt webhook setup is UI-only, per issue scope).

[...]

PR #3260, opened 2026-09-15
feat(eparagony): issue invoices through InvoicingPort over the eInvoice document

norbert-kulus-blockydevs:
Stack member 1 of #3192. **Targets the integration branch `3192-eparagony-invoicing`, not `main`** - per the epic, #3192 reaches `main` as one merge of that branch.

Part of #3173.

## What this slice is

The adapter, and nothing that reaches it. No manifest entry, no factory change, no module change, no frontend, no docs - so nothing in this slice is reachable at runtime yet. That is deliberate: the wiring is slice 3 and would be unreviewable stapled to 2,900 lines of mapper.

- `EparagonyInvoicingAdapter implements InvoicingPort, RegulatoryStatusReader`
- `eparagony-invoice.mapper.ts` - composes the `eInvoice` body, reads the status response into a neutral `RegulatoryClearanceResult`
- `invoice-tax-rate.policy.ts` - **new file beside `tax-rate.policy.ts`, which is untouched.** Invoices take `23, 8, 5, 3, ZRD, ZRICS, ZRE, EP, RCP, NS1, NS2`; receipts take the A-G device slots. Two vocabularies, two policies.
- invoice wire types and the invoice-lane status fields on `eparagony-api.types.ts`
- three spec files

Everything is built against the wire contract recorded on #3192, which came from a live sandbox POC.

## Decisions worth reviewing

**`merchantTIN` is connection config, optional on the type and required at compose time.** It is the same class of value as `posId` - vendor-assigned, stamped on every document, validated against the registered account rather than by checksum, so a wrong one fails every invoice on the connection rather than one order. Optional on the type is load-bearing: every existing connection is receipts-only and carries none, and making it required would break their config shape and force the back-fill the epic rules out. The mapper refuses pre-call with copy naming the remedy instead of letting the vendor answer with an opaque code 41.

**Arithmetic rounds once per RATE GROUP, never per line.** `taxGroup = round(grossGroup x r / (1+r))`, `netGroup = grossGroup - taxGroup`, last line of each group absorbs the allocation residual. So net + tax equals gross per group, the lines sum back to their own summary, and the total is the gross the buyer paid - all exact. The single-line case reproduces the shape of the POC's accepted body (asserted with `toMatchObject`; the mapper additionally stamps `currency` and `orderId`).

**A non-PLN invoice is refused pre-call.** A foreign currency requires `exchangeRate`, and ADR-040 forbids OpenLinker's analytics FX stamp supplying a fiscal document's rate. There is no other source, so refusing is the only honest option. The refusal text carries `unsupported currency` so core's `CURRENCY_REJECTION_MARKERS` route it to `invalid-currency`, pinned by a spec.

**`vatCalculationMethod: 'SUM_OF_RATES_NET'`** - the shape the live POC got a 202 for. The `*_GROSS` variants change which line fields become conditionally required and are documented-but-unexercised; the per-rate net summary is required either way.

## Correction to the issue's status table, applied here

The table on #3192 said `CONFIRMED + KSEF -> cleared`. **That is wrong, and this slice maps it to `accepted`.** Three independent facts in the tree say so:

- `cleared` is absent from `TerminalRegulatoryStatusValues`, so the #1121 reconciliation would poll a finished document for ever - the shape inFakt's #1293 finding already recorded.
- `invoices-list-page.tsx` filters it out by name, as "a reserved status no provider emits".
- KSeF's own adapter maps that same authority's terminal success to `accepted`, commenting it *"Cleared; KSeF assigned a number"* - so `accepted` already means cleared-with-a-number in this vocabulary.

eparagony relays to the authority KSeF talks to directly. Two adapters reaching one authority must not report its terminal success under two different neutral values. `cleared` stays reserved for a genuine split-clearance regime. The issue body is corrected to match.

## Known and stated

- **The `CONFIRMED + KSEF` arm is coded from the vendor contract, never observed.** The sandbox left a KSeF invoice at `OFFLINE` past eight minutes and minted no `ksefNumber`. Flagged in the spec's own comment; needs a production check before #3192 closes.
- **`pdfUrl` carries the vendor's HTML visualisation, not a PDF.** It is the only slot on the neutral record for a link to the issued document, and it is what the mockup's `sales-document-open-visualisation` needs. `RegulatoryDocumentReader` is the honest home; not in this slice.
- **Buyer address street/number split is inferred** from the trailing numeric token, falling back to `street = whole line, number = '-'`. The neutral `BuyerAddress` has no split and the vendor requires both. A mis-split is cosmetic and re-reads as the original; refusing a real paid order over a formatting detail would be the worse failure.
- The adapter is deliberately **not** a `DocumentNumberConsumer` - the vendor generates the legal number - but it honours `cmd.documentNumber` if a later slice turns numbering on.

## Mockup

`docs/plans/mockups/sales-document-eparagony-invoicing.html` needs **no edit in this slice**, and this slice implements none of its hooks - every one is a frontend hook belonging to a later slice.

**Corrected after review.** An earlier version of this paragraph claimed every hook's data was reachable from what this adapter produces, naming `sales-document-invoice-hash`, `provider-raw-status` and `provider-raw-processing-mode`. **That was wrong and would have misdirected the FE slice.** What this adapter produces is `InvoiceRecord` + `RegulatoryClearanceResult` + `IssuedDocumentLineAmounts`, and none of the three carries a hash, a raw provider status or a processing mode - `InvoiceRecord` has no `regimeExtras` and no `statusDetails`, unlike the fiscal-registration record, and `InvoiceRecordResponseDto` exposes none. `invoiceHash` is modelled on the wire type here and read by nothing.

Two hooks genuinely are reachable: the waiting ladder from `pending-submission`, and `sales-document-open-visualisation` from `documentUrl` (with the `pdfUrl` caveat above).

The mockup is corrected in **slice 3** (`6e16ea7f1`, on `3192-3-invoicing-surfaces`), which marks those three rows as having no data path and records that closing it is a core change with its own migration and DTO. It is not pulled into this PR because the mockup belongs to the surfaces slice and slice 3 also corrects three unrelated rows in the same edit.

## Verification

`pnpm type-check` clean across all 21 projects, `pnpm lint` zero errors, 90 of 90 invariant scripts pass (`check-repo-urls` fails only on local untracked `.worktrees/` scratch, identically on a clean `main`). Tests are written and deferred to CI per the repo's local-gate policy.

---

## Review fixes (third commit)

The review recomputed every arithmetic fixture by hand and confirmed the money is right,
then found three problems that were all the same shape: the adapter computed correct
numbers and handed core something it could not line up. All three are fixed, along with
nine smaller findings.

**1. Lines were emitted grouped by rate**, so document position was not command position.
Core persists `issuedLineSnapshot.lines` as `cmd.lines` verbatim and
`CorrectionLine.originalLineNumber` names a line BY ITS POSITION, so #3193 would have
credited the wrong line on a transmitted fiscal document; `InvoiceService.buildContent`
pairs `documentLines` by the same index. The arithmetic still groups by rate; only the
emitted array follows the command, through a per-index allocation the emit pass reads back.

**2. `documentLines` was never returned**, so `buildContent` fell back to a per-line
`round(gross / (1 + r))` - exactly the arithmetic this mapper exists to avoid. On the
three-grosz case the specs already pin, OpenLinker's contents card would have said
`[0.01, 0.01, 0.01]` net against a document that says `[0.01, 0.01, 0.00]`.

**3. The seller-tax-number remedy copy is discarded on this lane.**
`InvoiceService.deriveFailureReason` builds the operator sentence from the neutral failure
code, never from the adapter's `reason`, so the operator reads *"The invoicing provider
rejected the request."* for a request the provider never saw. The receipt lane does the
opposite with the same exception class. The throw site records the gap and names the real
fix (a neutral `seller-tax-id-required` code plus its marker, a core change with its own
slice).

> **Corrected in the fourth commit.** This paragraph originally claimed the copy "still
> reaches the log and `errorMessage`". **It does not.** `InvoiceService.sanitizeError`
> reads `error.message`, never `error.reason`, and `errorMessage: sanitized` therefore
> stores the technical message; `reason` is read in exactly one place,
> `classifyFailureCode`'s marker haystack, and then discarded. The mitigation is now real
> rather than described: the remedy rides in the **`message`**, which IS persisted as
> `invoice_records.errorMessage` and IS what the failure log carries. A spec pins the
> remedy text and asserts both texts stay clear of `TAX_ID_REJECTION_MARKERS`, since
> `classifyFailureCode` falls back to `message` whenever `reason` is not a string.

Also fixed: the group residual can no longer drive a line's net negative (borrowed back
from lines with capacity, leaving the group sum untouched); a `CONFIRMED` document whose
`processingMode` is unreadable now terminalises rather than being polled for ever;
`cmd.correction` and an unsupported `cmd.documentType` are refused before the boundary;
the two shared exceptions no longer call themselves the e-receipt provider now that they
serve both lanes; and the two document-type constants nothing referenced are deferred
to #3193.

Every fix carries a spec - including a pin on the refusal wording, because core's
`TAX_ID_REJECTION_MARKERS` would otherwise read a reworded "seller tax ID" as a BUYER
tax-id fault and send the operator to correct the wrong data.

## Known and stated

Three things this slice does not do, recorded rather than left to be discovered.

- **The `CONFIRMED + KSEF` arm is coded from the vendor contract and was never observed.**
  The sandbox left a KSeF invoice at `OFFLINE` past eight minutes and minted no
  `ksefNumber`. Needs a production check before #3192 closes.
- **The invoice lane has no crash-recovery locator**, unlike the receipt lane's
  `FiscalRegistrationLocator`, so `PendingRecoveryService` will mark a crashed post-claim
  invoice `failed` / `in-doubt` with no auto-recovery. It is **not implementable as the
  contract stands**: `RegulatoryLocateCriteria` carries `sellerTaxId` / `documentNumber` /
  `issuedFrom` / `issuedTo` and **no idempotency key**, while the fiscalization criteria do
  - and the vendor publishes no search by document number, so there is nothing to derive
  the deterministic token from. A real fiscal-safety asymmetry between two capabilities on
  one connection, and a contract limit rather than an oversight.
- **An all-exempt invoice sends `taxValueByTaxRate: {}`.** The wire contract lists the
  field as required; present-but-empty probably satisfies that, but the POC exercised taxed
  lines only. A spec pins the current behaviour; the vendor has never been asked.

---

## Review fixes (fourth and fifth commits)

Full response in the review thread. The three that change what this PR CLAIMS:

- **`allocateGroupNet` clamped the residual in one direction only**, so a realistic mixed-rate basket emitted a **negative per-line VAT** onto a document bound for a tax authority, with every summary still reconciling. Fixed in both directions, pinned by a per-line `0 <= net <= gross` invariant over four shapes, verified red-first against the pre-fix allocator and swept over 400 000 random baskets.
- **The remedy-copy mitigation above was false as written** and is now true - see the corrected paragraph in the third-commit section.
- **The mockup reachability claim was false** and is corrected in the Mockup section above.

Also in these commits: the four invoicing config keys are validated at the connection boundary (adopting #3267's stronger version verbatim so the stack converges on one answer); the shared `GET /documents/{token}/status` transport is extracted out of both adapters; the lane generalisation is finished in the two files it missed; an empty buyer `line1` refuses pre-call instead of emitting `street: ''`; `toEqual` replaces `toMatchObject` on the POC-parity test; `CURRENCY_REJECTION_MARKERS` is imported rather than hardcoded; and `scripts/check-eparagony-deadline-mirror.mjs` guards both plugin deadlines against core's `MAX_SUPPORTED_PROVIDER_TIMEOUT_MS`.

### Verification

`pnpm type-check` and `pnpm lint` clean for the package; `check:invariants` passes end to end, including the new guard; **282 of 282** eparagony tests pass. The blocking arithmetic fix was confirmed red-first: four of the new cases fail against the pre-fix `allocateGroupNet` and pass after it.

---

## Merge order (Piotr's IMPORTANT finding on the review)

`3192-1-invoice-adapter` and `3192-2-invoicing-wiring` (#3267) diverge at `2a65cc7c0` and neither is an ancestor of the other, so a textual rebase would hit a real conflict on the config-shape validator - both slices ended up touching it after their own reviews.

**Resolved by intent, stated here rather than left to be discovered at merge time:**

1. **This PR (#3260) merges first** into `3192-eparagony-invoicing`.
2. **#3267 rebases onto the updated integration branch second.** Its validator-adding commit (`0b1d75492`) will resolve as **already applied** - this PR adopted #3267's validator verbatim in `7ff544100` (see *Stack convergence* in the review thread), so the two histories agree on that hunk once #3260 has landed. #3267's manifest/dispatch/factory changes are net-new and are unaffected.
3. **#3269 (slice 3, web surfaces)** is already based on `3192-2-invoicing-wiring` and merges third, once #3267 is in.

No further action is needed in this PR's diff for this finding - it is a coordination note, not a code change.


norbert-kulus-blockydevs:
## Review

Reviewed at `2a65cc7c0` against the stack base `3192-eparagony-invoicing` (13 files, +3374/-22). No builds run; findings are read against the tree and the arithmetic was recomputed by hand.

### Summary

This is careful work and the centre of it is right. The per-rate-group rounding is the correct strategy, the command-order emission and the `documentLines` round-trip are load-bearing and hold (I traced `lineNumber = index + 1` through `InvoiceService.buildContent:1338` and `issuedLineSnapshot.lines = cmd.lines` at `:790`, so the wire lines, the snapshot and a later `CorrectionLine.originalLineNumber` all share one axis), the refusal wording genuinely dodges the `TAX_ID_REJECTION_MARKERS` trap on purpose, and the `cleared` to `accepted` correction is argued from three facts in the tree that all check out.

Two things need to change before this goes further. The residual repair in `allocateGroupNet` is one-sided, so a realistic mixed-rate basket emits a **negative per-line VAT** onto a document bound for a tax authority, and the group summary still reconciles, which is exactly why the suite is green. Separately, four new operator-settable config keys reach a transmitted fiscal document with no boundary validation at all.

---

### BLOCKING

**1. `libs/integrations/eparagony/src/infrastructure/adapters/eparagony-invoice.mapper.ts:658-670` - a line can be emitted with a negative VAT amount**

`allocateGroupNet` repairs only `allocation[last] < 0`. It never clamps the last member's net at its own gross, so when earlier members round down the last member absorbs the surplus, and the emit pass at `:289` (`const taxValue = member.grossMinor - lineNet`) writes a negative `taxValue` onto the wire and a negative `tax` into `documentLines`. Reproduced against the exact code:

```
gross [3, 3, 1]     @23% -> net [2, 2, 2]     tax [1, 1, -1]
gross [3x9, 1]      @23% -> net [2x9, 5]      tax [1x9, -4]
gross [100x9, 1]    @23% -> net [81x9, 4]     tax [19x9, -3]
```

`Sum(net) + Sum(tax) === grossSaleValue` still holds in every case, which is why nothing fails today, and `calculationValidation: NONE` (`eparagony-api.types.ts:236`) means the vendor will not catch it either. This is not a constructed edge: `toShippingLines` appends small per-rate shares last and `groupByRate` preserves first-seen order, so a shipping share is the last member of its group whenever its rate already appeared. A randomised sweep over plausible baskets (2-13 item lines plus a small trailing shipping share, 23%) hits it in roughly 1% of cases.

The invariant the file needs is `0 <= lineNet <= grossMinor` per line. Only the lower bound exists, and only on the last member. Clamp `allocation[lastIndex]` into `[0, g_last]` in both directions and push the excess onto earlier members that still have headroom (`g_i - a_i`), mirroring the existing borrow-back loop.

**2. `__tests__/eparagony-invoice.mapper.spec.ts:236` - the assertion that catches this already exists, on the wrong fixture**

`expect(wire.every((line) => line.taxValue >= 0)).toBe(true)` is written, and is only exercised on ten 1-grosz lines, where the residual runs negative and the existing repair fires. Add `[0.03, 0.03, 0.01] @ 23%` to the same `it` (it fails today with `taxValue: -1`) and pin the real invariant `0 <= netTotalLineValue <= grossMinor` per line, since neither bound is currently held across shapes.

**3. `eparagony-connection-config-shape-validator.adapter.ts` - four new config keys with no shape validation**

The PR adds `merchantTIN`, `merchantName`, `merchantAddress` and `eInvoicingHubEnabled` (`eparagony-config.types.ts:150-181`) and extends no arm of the validator, which already validates every other optional key (`print`, `paymentForm`, `paymentName`, `statusPollTimeoutMs`, `fiscalDeviceUniqueNumber`) and whose own docblock says catching a malformed config here matters more than usual for this plugin. `config` is JSONB, so the TS type is not a runtime guarantee, and the FE setup form is create-only (`eparagony-setup.schema.ts:71` emits `{environment, posId}`), so the raw JSON editor is the operator's only route - the exact bypass #2610 cites when it requires a server-side check.

Three consequences, none covered by a spec:

- `toIssuedDocumentSeller:404` guards `address === undefined` but not a partial object, so `` `${address.street} ${address.number}`.trim() `` renders the literal `"undefined undefined"` into `IssuedDocumentSeller.address.line1`, which core **persists** in the issued-document snapshot.
- `toSellerEntityAddress:752` copies the address onto the wire type-checking nothing but `apartment`.
- `eInvoicingHubEnabled: "true"` (string) fails the `=== true` test at `:368` and silently issues outside the hub, which this code itself calls a complete successful outcome. No error, legally different document.

This is the same `eparagony:undefined:...` class of defect `upsertCustomer` is carefully defended against 300 lines away, so the defensiveness is inconsistent rather than absent by policy. Add the five checks with a spec each, or state the deferral explicitly in the PR body and in the wiring issue - it currently does not mention validation at all.

---

### IMPORTANT

**4. `eparagony-invoice.mapper.ts:229` and the PR body - the stated mitigation for the discarded remedy copy is wrong, and it is the justification for shipping the gap**

Both the comment and the PR body say the copy "still reaches the log and `errorMessage`". It does not. `InvoiceService.sanitizeError:1299` reads `error.message`, never `error.reason`, and `errorMessage: sanitized` at `:771` therefore stores the technical message. `reason` is read in exactly one place, `classifyFailureCode`'s marker haystack, and then discarded. So `"This connection has no seller tax number set... Set it on the connection and re-issue."` reaches nothing - not `failureReason`, not `errorMessage`, not the log. What reaches them is the `message`, which names the condition and not the remedy. The deferral may still be right; the argument for it is false as written. Either correct both texts, or fold the remedy into `message` (one line) so the claim becomes true - in which case the `TAX_ID_REJECTION_MARKERS` spec at `mapper.spec.ts:495` needs to assert against the message too.

**5. The same gap exists one rule over, unremarked: `eparagony-invoice.mapper.ts:566`**

A rate-less line refuses with `"At least one order line carries no tax rate..."`, which matches none of `TAX_ID_REJECTION_MARKERS`, `CURRENCY_REJECTION_MARKERS` or `SALE_CLASSIFICATION_REJECTION_MARKERS` (`invoice.service.ts:994-1002`), so `classifyFailureCode` returns `provider-rejected` and the operator reads "The invoicing provider rejected the request." for a request no provider saw. `OL_TAX_RATE_STRICT_ENABLED` is off by default and `toShippingLines` emits `taxRate: ''` when the split is uncomputable, so on a catalogue with partial rate coverage this adapter refuses where inFakt, Subiekt and KSeF all issue. The file writes a long comment about exactly this hazard for the seller-tax-number case and then leaves this one silent. Record it in the same place.

**6. PR description, Mockup section - the accuracy claim does not hold at this head**

`docs/frontend-architecture.md:465` requires an implementing PR to touch the mockup or say why it is still accurate. The PR says the latter and asserts `sales-document-invoice-hash` is reachable from `ksefInvoice.invoiceHash` and `provider-raw-status` / `provider-raw-processing-mode` from the modelled `status` / `processingMode`. None of those is reachable from what this adapter *produces*: the neutral surface is `InvoiceRecord` + `RegulatoryClearanceResult` + `IssuedDocumentLineAmounts`, and none carries a hash, a raw provider status or a processing mode. `invoiceHash` is modelled on the wire type and read by nothing. A later commit on the working branch (`6e16ea7f1 docs(mockups): mark the three raw-provider rows as having no data path`) reverses exactly this claim, so the conclusion has already been drawn. Either pull that commit into this PR or rewrite the paragraph; as written it misdirects the FE slice.

**7. The lane generalisation stopped two files short**

`"The e-receipt provider"` was correctly generalised to `"The provider"` across the six `defaultReason` arms and `EparagonyNetworkError.reason`, because those classes now serve both lanes. Two sites were missed:

- `infrastructure/http/eparagony-http-client.ts:405` still says `'The e-receipt provider issued no access token.'`, and the invoicing adapter uses the same `IEparagonyHttpClient`, so an OAuth failure during invoice issuance shows receipt copy.
- `eparagony-network.error.ts:23` still ends `"...the sale may or may not have been registered"`, the receipt verb, and `pollToSettledIssuance` throws precisely that on the poll-budget path (`eparagony-invoicing.adapter.ts:378`), where what is in doubt is whether an invoice was issued.

No spec asserts either the old or the new strings anywhere, so neither direction is caught. Worth pinning one.

**8. `eparagony-config.exception.ts:22` - the docblock now states a guarantee it has on only one lane**

`/** Short, PII-free operator-facing summary; core persists it verbatim. */` is true for fiscalization and false for invoicing, as finding 4 shows. This PR edits the file's header and leaves the contradicted line. A future adapter author reads this docblock, not the mapper comment 300 lines away.

**9. `mapper.spec.ts:113` - `toMatchObject` cannot support the claim the test name makes**

The test is named "reproduces the POC body field for field" and asserts `metadata` with `toMatchObject`, which passes over extra keys - precisely the property a parity assertion exists to pin. The body already adds `currency` and `orderId` beyond the POC (asserted separately at `:418`/`:422`), and the next field added to `metadata` enters silently while the test keeps claiming parity. Use `toEqual` with the complete expected object, or rename the test. The `lines` array beneath it is already `toEqual`, so the inconsistency is inside one test.

**10. `mapper.spec.ts:473` - the currency-marker assertion hardcodes the phrase instead of importing the constant**

`invoicing.types.ts:130-137` states `CURRENCY_REJECTION_MARKERS` is published specifically so an adapter spec can assert its own message still matches and a core reword breaks the build. The established consumer does exactly that (`libs/integrations/infakt/.../infakt-invoicing.adapter.spec.ts:24,606`). This spec asserts `.toContain('unsupported currency')`, which pins the adapter's own wording and nothing about core's. Import the constant. The `TAX_ID_REJECTION_MARKERS` half at `:496` must stay hardcoded (that one is private to `InvoiceService`) and is fine as-is.

**11. `eparagony-invoice.mapper.ts:764` - an empty `line1` emits `street: ''` against a required field**

`splitStreetAndNumber('')` returns `{ street: '', number: '-' }` and that goes onto the wire, while `EparagonyEntityAddress.street` is required (`eparagony-api.types.ts:188`) and the module docblock at `:50` promises every unmappable input throws before anything is sent. `BuyerAddress.line1` is typed `string` with no non-empty guarantee (`invoicing.types.ts:362`), and `city` / `postalCode` are likewise copied unchecked at `:746-747`. Refuse pre-call like the currency and quantity cases, or fall back to a placeholder, and add `''` / `'   '` to the spec.

**12. Coverage holes, one of them on the fiscal-safety path**

I checked each claimed fix and eight of them do carry a spec that would fail against the pre-fix code. Genuinely uncovered:

- `resolvePollTimeoutMs` clamping (`eparagony-invoicing.adapter.ts:540-546`) at **both** boundaries, although `MAX_STATUS_POLL_TIMEOUT_MS` is the value guarding the double-issue invariant the module-load assertion at `:127` exists for. A config of `999_999` or `-1` is untested.
- `readStatus`'s non-object-body throw (`:457-461`).
- `getClearanceStatus`'s whitespace-only token branch (`:276`).
- `toPricedLine`'s `grossMinor === null` arm (`:580`); only the `< 0` half is tested.
- A fractional quantity, the one case `resolveNetUnitPrice`'s own docblock admits can differ by a grosz.
- `merchantName` + `merchantAddress` present with `merchantTIN` absent in `toIssuedDocumentSeller`.

Also `adapter.spec.ts:452` does not discriminate: `makeRecord(null)` defaults to `regulatoryStatus: 'pending-submission'` / `clearanceReference: null` and the test asserts exactly those, so a hardcoded return would pass. The sibling at `:463` does it right.

**13. `eparagony-invoicing.adapter.ts:441-473` - the vendor status-read transport is duplicated verbatim from the fiscalization adapter**

`readStatus` is byte-identical to `EparagonyFiscalizationAdapter.readStatus` apart from a trailing comma, and `resolvePollTimeoutMs`, `sleep`, the module-level deadline assertion and five of six poll constants are duplicated too. The module docblock justifies a separate file on the **mappers** (they share no field, no rate vocabulary and no arithmetic), which is true and well argued, but it does not cover the transport, and `GET /documents/{token}/status` is explicitly one endpoint for both lanes (`eparagony-api.types.ts:8-11`). The `EPARAGONY_ERROR_UNKNOWN_DOCUMENT` translation, the non-object-body guard and the timeout clamp now have to be fixed twice. Extract a shared status reader beside the client it wraps, leaving each adapter its own terminal predicate and default budget.

---

### SUGGESTIONS

- **`eparagony-invoicing.adapter.ts` `resolveRegistrationKey`** - a vendor-side `ERROR` is permanently unretryable through the product's own retry button. `failureMode: 'rejected'` means "safe to re-attempt", but the token is a pure function of `(connectionId, registrationKey)` and the vendor holds the failed document under it forever, so every retry hits `DOCUMENT_ALREADY_EXISTS`, polls, re-reads `ERROR` and throws the identical rejection. The bulk retry path reuses `record.idempotencyKey` deliberately (`invoicing.controller.ts:595`). The docblock says a genuine retry needs a fresh key without noting that no product surface mints one. Worth a README line and possibly a hint on the failure reason.
- **`eparagony-invoice.mapper.ts:135`** - the street/number split puts the apartment in the building-number field on a common Polish form. `Aleje Jerozolimskie 44 m. 8` becomes `{Aleje Jerozolimskie 44 m., 8}`, so the vendor's building-number field carries the apartment on a document bound for KSeF. `ul. Kwiatowa 12/3` and `Plac Zbawiciela 1A` are handled correctly, and none of the four shapes is in the spec. The docblock's claim that a mis-split "cannot lose data" is true of concatenation and not of meaning; say "can mis-assign the building number" instead.
- **`eparagony-invoice.mapper.ts:665`** - the borrow-back repair can leave a line at 100% effective VAT (ten 1-grosz lines at 23% give two lines `net 0, tax 1, gross 1`). Totals are right and it beats a negative net, but the docblock only promises "none of them goes negative either". One sentence.
- **`eparagony.constants.ts:72`** - `EPARAGONY_ISSUE_DEADLINE_MS = 110_000` mirrors core's `MAX_SUPPORTED_PROVIDER_TIMEOUT_MS = 120_000` and its docblock admits the mirror, but the module-load assertion only checks the poll ceiling against this constant, not this constant against core's. The repo has ~30 `check-*-mirror.mjs` guards for this shape; a one-line spec assertion would close it, and the fiscal double-issue guard is what is at stake.
- **`eparagony-invoicing.adapter.ts:418`** - the vendor's `errorDescription` is logged verbatim under a comment copied from the fiscalization adapter ("can echo a product name"), which was written for a receipt payload. This lane's payload also carries `consumerName`, `consumerAddress` and `consumerTIN`, so a validation error naming the buyer is likely rather than theoretical. The mitigation that matters (never promoting it into `reason`) is in place and spec'd. Either truncate in the log or state that this lane accepts buyer PII there and why.
- **`eparagony-api.types.ts:340`** - `EPARAGONY_PROCESSING_MODE_KSEF` is declared and read nowhere. Consume it in `toRegulatoryClearanceResult` or drop it.
- **`invoice-tax-rate.policy.ts:138`** - a comma-decimal rate (`'23,00'`) becomes `NaN` and falls through to a hard block, where core's own `parseTaxRatePercent` uses `parseFloat` and reads 23. Unverified whether any shipped source writes comma decimals; a one-line normalisation closes it either way.
- **`eparagony-invoice.mapper.ts:716`** - `roundMinorUnits` silently returns `0` on a non-finite input. Unreachable today, but a silent `0` inside a VAT split produces a document that reconciles internally and understates tax. Prefer throwing.
- **Two unreachable guards** (`resolveNetUnitPrice`'s `quantity <= 0`, `roundMinorUnits`'s finite check) read as live branches; a one-line note saves the next reader a hunt.
- **`README.md:68-74`** - the operator config table gains no rows for the four new keys, so an operator who enables invoicing has no documented way to discover `merchantTIN` and gets a refusal that does not say where to set it. The `eInvoicingHubEnabled` docblock text at `eparagony-config.types.ts:170-181` belongs there too.
- **Three base-owned files reformatted** (trailing commas stripped) - ~14 of the diff's 22 deletions, and the new `invoice-tax-rate.policy.ts:162` keeps a trailing comma in a param list, so the package is now internally split. It matches root `.prettierrc`, but `eslint-config-prettier` disables formatting rules rather than enforcing them, so nothing required it. Either revert or apply uniformly and say so.
- **Test naming** - `invoice-tax-rate.policy.spec.ts` follows `should ... when ...` in all 18 cases; the two new adapter/mapper specs use it in zero of 83, and the sibling fiscalization spec mostly does. The prose is good and I would not trade it for ceremony, but pick one per package.
- **Docs obligations for the wiring slice** are unrecorded: `docs/architecture-overview.md` section 14 enumerates each `InvoicingPort` adapter by adapterKey, section 16 currently states eparagony implements `FiscalizationPort` + `FiscalRegistrationLocator` and no device sub-capability, and the README gap list is stale. Naming them in the PR body keeps them from falling between slices.

---

### Verified correct (stating it so the absence is not read as an omission)

- **Reachability.** `eparagony-plugin.ts:56` still declares `['Fiscalization', 'FiscalRegistrationLocator']` and the dispatch table at `:121` carries `Fiscalization` only, so `getCapabilityAdapter(id, 'Invoicing')` fails the manifest gate. The barrel export makes the class importable, not dispatchable. The "no integration test because the slice is unreachable" argument is therefore correct as stated.
- **Every arithmetic invariant except finding 1.** `netGroup + taxGroup === grossGroup` by construction; groups sum to `grossSaleValue`; per-group line sums hold including after the borrow repair; `netGroup >= 0` always, which is what makes the borrow provably able to cover the shortfall. No division by zero, no `-0` leak, and the `toFixed(4)` collapse cannot manufacture a tie for integer minor units. Zero-gross lines, single-line groups and fractional quantities behave. The fixtures are hand-computed, not snapshots.
- **Failure-code routing.** The currency reason contains a literal `CURRENCY_REJECTION_MARKERS` member and none of the tax-id markers; the seller reason says "tax number" deliberately, dodging the tax-id markers that are checked first. No refusal string is matched by the wrong set.
- **Status mapping.** Every arm of `toRegulatoryClearanceResult` lands on a `TerminalRegulatoryStatusValues` member except the catch-all `pending-submission`, which is right for `OFFLINE` / `PENDING` / unrecognised. A `CONFIRMED` document with an unreadable `processingMode` terminalises at `accepted`, so nothing settled is polled forever, and `accepted` rather than `cleared` agrees with the KSeF adapter's mapping of the same authority.
- **Idempotency and cross-lane tokens.** The vendor `Idempotency-Key` is the derived `documentToken`, the create is retried only under `idempotent: true` with a byte-identical body, `DOCUMENT_ALREADY_EXISTS` resolves by status read rather than re-create, and auto-issue mints `invoice:{conn}:{order}` vs `fiscal:{conn}:{order}`, so the two lanes cannot collide on the shared `deriveDocumentToken`.
- **Boundary, types, error handling.** Every `@openlinker/core/*` import is a bare top-level barrel; no deep paths, no `*RepositoryPort`, no `*OrmEntity`. No `any`, no `as unknown as`, no non-null `!`, no `console.*`, no `eslint-disable`. `as const` unions throughout. All four `InvoicingPort` members implemented with honest semantics (`getInvoice` returning `null` mirrors `KsefInvoicingAdapter:857` including the docblock, and core owns that read). Nothing swallowed: both `catch` blocks are narrowly gated and rethrow everything else.
- **Deadline invariant.** `EPARAGONY_ISSUE_DEADLINE_MS` (110s) < core's `MAX_SUPPORTED_PROVIDER_TIMEOUT_MS` (120s), and `MAX_STATUS_POLL_TIMEOUT_MS` (90s) < 110s with a module-load assertion.
- **The `*.types.ts` pure-rule exception is not violated** - every runtime function lives in `domain/policies/`, which is an established plugin convention (erli, infakt), and the two rate vocabularies genuinely share no symbol with the untouched receipt-side `tax-rate.policy.ts`.

---

### Merge readiness

**Request changes.**

Priority order:

1. Clamp `allocateGroupNet`'s last allocation into `[0, g_last]` in both directions and redistribute the excess to members with headroom. Add `[0.03, 0.03, 0.01] @ 23%` to the existing `taxValue >= 0` test and pin `0 <= net <= gross` per line.
2. Extend the config-shape validator for the four new keys, and guard a partial `merchantAddress` so it cannot render `"undefined undefined"` into a persisted snapshot. Or record the deferral explicitly.
3. Correct the two claims the PR rests decisions on: the remedy copy reaches nothing, and the mockup's three raw-provider rows have no data path.
4. Finish the lane generalisation (`eparagony-http-client.ts:405`, `eparagony-network.error.ts:23`) and fix the `eparagony-config.exception.ts:22` docblock.
5. `toEqual` on the POC-parity test, import `CURRENCY_REJECTION_MARKERS` in the spec, and close the `resolvePollTimeoutMs` and empty-`line1` holes.

Everything else is a suggestion and can ride with the wiring slice.

norbert-kulus-blockydevs:
## Review response

Two commits: `3abbcc63b` (the findings) and `7ff544100` (converging with #3267, explained at the bottom - **please read that one, it affects how this stack rebases**).

Everything below was re-derived against the code rather than taken on trust. Where the review was right I say so briefly; where I disagreed or chose differently I say why.

---

### BLOCKING

**1 + 2 - `allocateGroupNet` emits a negative VAT. Confirmed, fixed, and the fix is proved rather than asserted.**

The reproducers all hold. `[3, 3, 1] @ 23%` -> `taxGroup = 1`, `netGroup = 6`, the first two lines round to 2 each, the third computes `6 - 4 = 2` against a gross of 1, and `taxValue = 1 - 2 = -1`. The repair is now two-sided: a surplus is **lent** to earlier members that still have headroom (`g_i - a_i`), mirroring the existing borrow-back loop. Neither direction can run out, and the bound is arithmetic rather than hopeful - the earlier headroom is exactly `taxGroup + surplus`, and `taxGroup` is never negative, so it always covers.

Your reading of why nothing failed is exactly right and worth restating, because it is the lesson: the group sum is untouched by the defect, so every summary assertion in the suite stayed green while a line on the wire carried a negative tax. `calculationValidation: NONE` means the vendor would not have caught it either.

The spec is generalised as you asked, and then some. The existing `taxValue >= 0` test was on the one fixture where the residual runs the OTHER way, so it exercised the repair that already existed. It is now `it.each` over four shapes - your `[0.03, 0.03, 0.01]`, a longer group, a realistic basket with a small trailing shipping share, plus the original ten-1-grosz case - asserting the real invariant `0 <= net_i <= gross_i` **and** `net_i + tax_i === gross_i` per line, not just the lower bound on the last member.

Two things I did beyond the ask, because "it passes now" is not evidence:

- **Verified red-first.** I removed the new surplus branch and re-ran: four tests fail (`Expected: <= 1, Received: 2 / 5 / 4`), and the below-zero case still passes - which is the right shape, since it confirms the new cases target precisely the direction that was unfixed.
- **Swept it.** 400 000 randomised baskets (1-13 item lines plus a small trailing share, across `0 / 3 / 5 / 8 / 23%`): zero violations of `0 <= net <= gross` and zero sum mismatches. Your ~1% hit-rate estimate for the defect matched what I saw before the fix.

A consequence worth stating rather than leaving to be found: the repair moves net **between** lines, so a line that lends its headroom can end at 0% effective VAT, just as your other suggestion noted a borrowed-from line can end at 100%. Both are in the docblock now. That is the honest cost of expressing sub-grosz VAT per line at all; the alternative is a figure the arithmetic cannot support.

**3 - four config keys with no validation. Confirmed, and #3267 had already fixed it.**

You were right on all three consequences. I implemented it, then found that **slice 2 (#3267) had independently implemented the same thing** after its own review, on the same file, and its version is stronger than what I wrote. I discarded mine and took slice 2's verbatim - see *Stack convergence* at the bottom.

The mapper half is mine and goes further than either: `readSellerEntityAddress` is one reader used by **both** the wire body and `toIssuedDocumentSeller`, refusing `undefined`, `null`, a non-object and a **partial** object. Slice 2's guard covers `null` (a real crash - #3274) but not partial, which is your `"undefined undefined"` case. Reading through one function is what makes "what core persists" and "what the vendor receives" incapable of describing different sellers.

---

### IMPORTANT

**4 - the mitigation for the discarded remedy copy was false. You are right, and it was load-bearing.**

I checked: `sanitizeError:1299` reads `error.message`; `reason` is read only in `classifyFailureCode`'s haystack and then discarded. So the sentence reached **nothing**. That was the argument for shipping the gap, so the argument was empty.

I took your second option rather than correcting the text, because it makes the claim true instead of retracting it: **the remedy now rides in the `message`**, which IS persisted as `invoice_records.errorMessage` and IS what the failure log carries. The PR body paragraph is corrected in place with a note saying what it used to claim.

A spec pins the remedy text, and the `TAX_ID_REJECTION_MARKERS` assertion now covers **both** texts - `classifyFailureCode` falls back to `error.message` whenever `reason` is not a string, so under refactoring the two are one haystack. Your suggestion to extend that spec was the right instinct; I extended it to the message rather than switching to it.

**5 - the same gap one rule over. Confirmed and closed the same way.**

You are right that this one is more reachable than the seller case: `OL_TAX_RATE_STRICT_ENABLED` is off by default and `toShippingLines` emits `taxRate: ''` when the split is uncomputable, so a catalogue with partial rate coverage refuses here where inFakt, Subiekt and KSeF issue. It is now recorded at the throw site and its remedy is in the message too, so the operator learns **which product** to fix.

Both refusals point at one note above `composeInvoiceDocument` that states the lane gap once - `reason` is matched and discarded, `message` is what survives - rather than repeating a 20-line comment per throw.

**6 - the mockup claim. You are right, and the conclusion had indeed already been drawn up-stack.**

Rewritten rather than pulled in. `6e16ea7f1` lives on `3192-3-invoicing-surfaces` and also corrects three unrelated rows, so cherry-picking it into slice 1 would move surfaces work into the adapter slice. The Mockup section now says plainly that the three raw-provider rows have **no data path** - `InvoiceRecord` carries no `regimeExtras` and no `statusDetails`, `InvoiceRecordResponseDto` exposes none, and `invoiceHash` is modelled here and read by nothing - names the two hooks that genuinely are reachable, and points at the slice-3 commit that corrects the file.

**7 - two files missed. Both fixed, both now pinned.**

`eparagony-http-client.ts:405` and `EparagonyNetworkError`'s reason, exactly as you found them. Your observation that no spec asserts either direction is why they survived, so both are pinned now: a `describe` block asserting the shared classes' copy names no lane, and a test in the client spec driving a token-less response.

`EparagonyNetworkError`'s wording is the more interesting one - "the sale may or may not have been registered" is now "the document may or may not have been issued", which is true on both lanes rather than on one.

**8 - the docblock states a one-lane guarantee. Fixed, and made specific.**

It now says which lane persists it verbatim, which one matches-and-discards, and what an adapter author must therefore do on the invoicing lane. You are right that this docblock is what the next author reads, not a comment 300 lines away.

**9 - `toMatchObject` cannot support a parity claim. Agreed; `toEqual` with the complete object.**

The test is renamed to what it actually asserts, and the two keys beyond the POC body (`currency`, `orderId`) are named in the expectation rather than tolerated by the matcher. The inconsistency you spotted - `lines` already `toEqual` one line below - was the tell.

**10 - hardcoded phrase instead of the published constant. Fixed.**

`CURRENCY_REJECTION_MARKERS` is imported and asserted against, matching the inFakt precedent you cited. Your note that the `TAX_ID_REJECTION_MARKERS` half must stay hardcoded (private to `InvoiceService`) is correct and it is unchanged.

**11 - an empty `line1` emits `street: ''`. Fixed, taking the stricter of your two options, plus something I found while checking it.**

`line1`, `postalCode`, `city` and `countryIso2` are now refused pre-call when blank. I chose refusal over a placeholder because a fabricated street on a document bound for the national hub is worse than a refusal an operator can act on, and the vendor declares all four required so a blank one fails anyway - just with a code the operator cannot read.

It is framed as a **presence** check, not a validity one, so it does not contradict ADR-073 decision 5: the format of a present value still travels verbatim, which the foreign-postcode spec still pins.

**Something the review did not raise and I checked before shipping it:** under `OL_STORE_PII=false`, core's `redactAddress` writes a non-empty placeholder into `address1`, `city` and `postalCode`, so those three still pass and the placeholder reaches the document exactly as before - a pre-existing question for the wiring slice, not one this check creates. `country` is the exception (`address.country ?? ''`), so a hash-only deployment whose source reported no country now refuses where it previously reached the vendor. That is the only new refusal with real-world reach, and it is recorded in the code where it is caused.

**12 - coverage holes. All six closed, and the non-discriminating test fixed.**

- `resolvePollTimeoutMs` clamping at **both** boundaries - and better than a spec through the adapter, because the extraction in finding 13 made it a pure exported function. `999_999` and `-1` are both covered, plus non-numeric fallbacks.
- `readStatus`'s non-object-body throw - covered at the reader and through `getClearanceStatus`, asserting `EparagonyNetworkError` (in-doubt) rather than a verdict.
- `getClearanceStatus`'s whitespace-only token - now `it.each([null, '', '   '])`.
- `toPricedLine`'s `grossMinor === null` arm - `NaN` and `Infinity`.
- A fractional quantity - the one case `resolveNetUnitPrice`'s docblock admits can differ, asserting the tolerated grosz of drift **and** that the summary stays exact.
- `merchantName` + `merchantAddress` with `merchantTIN` absent.

On `adapter.spec.ts:452` you were right that it discriminated nothing. The record now carries `'submitted'` / `'ref-echo'`, values that are not the defaults, so a hardcoded return fails it.

**13 - duplicated transport. Agreed, extracted.**

`readEparagonyDocumentStatus`, `resolveStatusPollTimeoutMs`, `sleep`, the documents path and five poll constants now live in `infrastructure/http/eparagony-document-status.reader.ts`, beside the client, as you suggested. Each adapter keeps what genuinely differs: its terminal predicate, its **default** budget (a device answering in seconds is not an invoice being composed and relayed) and its **own** module-load assertion against its own deadline - so the ceiling is shared while the two safety assertions stay independent.

Net effect on the fiscalization adapter is -40 lines with no behaviour change, and the `EPARAGONY_ERROR_UNKNOWN_DOCUMENT` translation and the clamp now have one home.

---

### SUGGESTIONS

Taken, except one.

- **Vendor-side `ERROR` is permanently unretryable** - README entry stating it plainly: `failureMode: 'rejected'` means *nothing was double-issued*, **not** *a retry will behave differently*, and no product surface mints a fresh key.
- **Street/number mis-split** - docblock reworded to your phrasing (it "can mis-assign the building number", which is a different claim from "cannot lose data"), and all four of your shapes are now in the spec, including the `m.` form that gets it wrong.
- **Borrow leaving a line at 100% VAT** - in the docblock, alongside the 0% case the new lend direction introduces.
- **`EPARAGONY_ISSUE_DEADLINE_MS` unchecked against core's ceiling** - I went further than a spec assertion. A plugin spec cannot import that constant (core exports it from a service module, not a barrel; a deep import is ESLint-banned and fails at Node runtime), so this is `scripts/check-eparagony-deadline-mirror.mjs` under `check:invariants`, guarding **both** lanes' deadlines against their own core counterpart, with a self-check. This was the one suggestion with a fiscal double-issue behind it, so it got the guard the repo uses for that shape rather than a test.
- **`errorDescription` buyer PII** - you are right that on this lane it is likely rather than theoretical. The comment no longer reuses the receipt lane's wording, states that this lane accepts buyer data in the log and why (a bare numeric code does not tell an operator which field to fix), and the value is **length-bounded** before it is logged.
- **Comma-decimal rate** - normalised, and deliberately **narrower than `parseFloat`**: `parseFloat('23,5')` is 23, which would silently turn a 23.5% rate into a 23% one. The pattern takes one or two digits after a single comma, so `23,00` resolves and `1,000` does not.
- **`roundMinorUnits` silent `0`** - now throws. Your reasoning is the point: internally consistent, nothing logged, tax understated.
- **Unreachable guards** - noted in place. One of the two is gone (`roundMinorUnits` throws now); `resolveNetUnitPrice`'s remains, with a note saying which upstream check makes it unreachable.
- **README config table** - all four keys documented, with the partial-address and string-`"true"` traps stated where an operator meets them.
- **Prettier churn** - reverted on all three base-owned files, so their diff is copy-only. The files this PR **adds** are normalised to the package's own trailing-comma style instead, which closes the internal split you spotted at `invoice-tax-rate.policy.ts:162`. Worth noting: nothing enforces either style - no CI `format:check`, and `eslint-config-prettier` only disables rules - so this is a consistency choice, not a rule being followed.
- **Docs obligations for the wiring slice** - already done by #3267 (`architecture-overview.md` §14/§16 and `capabilities.md` both carry the second registry capability and the invoicing adapter's coverage). Naming it here so it does not get done twice.

**Not taken: test naming.** The two new specs keep their prose names. You said you would not trade the prose for ceremony and I agree; picking one per package would mean renaming 83 passing tests in this PR and leaving the sibling fiscalization spec split anyway. If the package should standardise, it is worth its own change rather than a rider on a fiscal-arithmetic fix.

**Disagreed: `EPARAGONY_PROCESSING_MODE_KSEF`.** You offered "consume it or drop it"; I did neither, and here is why. Consuming it means re-testing the mode in `toRegulatoryClearanceResult` arm 3, which is exactly the behaviour change that arm's docblock argues against - a `CONFIRMED` document with an unreadable mode would route to the non-terminal `pending-submission` and the #1121 reconciliation would poll a settled document for ever. And dropping it would be inconsistent with the file it lives in: `EPARAGONY_STATUS_READY` and `EPARAGONY_STATUS_PENDING` are **also** declared and branched on nowhere, and both predate this PR. That file models the vendor's enums rather than only the values the code tests. So the constant stays and the docblock now records the convention and why there is no comparison for it to appear in.

---

### Stack convergence - please read before rebasing

While fixing finding 3 I found that **#3267 (slice 2) had already implemented it**, in commit `2979424a8`, after its own review. Both changes touch the same file and the same two mapper guards, so a rebase would have hit a real conflict between two different answers to one question.

I resolved it in slice 1 rather than leaving it:

- The **validator implementation and its spec are taken from #3267 byte-for-byte**. Its version is stronger than what I wrote - an ISO 3166-1 alpha-2 check on the seller country tested against the **raw** value (because `toSellerEntityAddress` ships it untouched, so validating a trimmed copy would bless `' PL '` into a two-letter field on every invoice this connection issues), plus a written-down decision *not* to format-check `merchantTIN` for the #2240 mirror-stricter-than-the-gate reason. The keys are introduced by **this** slice, so the validation belongs here.
- The **mapper keeps this slice's `readSellerEntityAddress`**, which subsumes #3267's `!= null` guard and additionally refuses a partial object. #3267's reasoning for why `null` must read as absent, and why both callers must read through one function, is folded into its docblock, and #3267's two specs come across with it.

**So when this stack rebases, #3267's validator hunks should resolve as already-applied, not as a conflict to merge.** I deliberately did **not** rebase `3192-2-invoicing-wiring` or `3192-3-invoicing-surfaces` myself - both have moved since this review was written and are being worked concurrently; rewriting them from here would have been the wrong call.

---

### Verification

`pnpm type-check` clean, `pnpm lint` zero errors, `check:invariants` green end to end including the new guard, **282 of 282** eparagony tests passing. The blocking arithmetic fix is confirmed red-first against the pre-fix allocator and swept over 400 000 random baskets.

Two things I did **not** close and am not claiming to have: the `CONFIRMED + KSEF` arm is still coded from the vendor contract and never observed, and the invoice lane still has no crash-recovery locator (`RegulatoryLocateCriteria` carries no idempotency key and the vendor publishes no search by document number, so it is not implementable as the contract stands). Both remain in the PR body's *Known and stated*.

piotrswierzy (approved):

**Verdict: ✅ Approve.** The central architectural question this PR raises — does putting `InvoicingPort` on the eparagony plugin blur the boundary ADR-042 drew — resolves in the PR's favour, and the money arithmetic at its heart is correct with a termination argument I was able to verify rather than take on trust.

### ADR-042's boundary is respected, not blurred

§ 16 is emphatic that fiscalisation is *not* invoicing and deliberately not an `InvoicingPort` document type. Nothing here contradicts that. The receipt stays on `FiscalizationPort` via `EparagonyFiscalizationAdapter`; the invoice is a **different document** on a **different port** via a **different adapter**. What is shared is a connection, an OAuth scope set and an HTTP client — none of which is the thing ADR-042 separates. Two capabilities on one plugin is the legitimate reading, and `dispatchCapability` carries exactly the two base capabilities with both sub-capabilities (`FiscalRegistrationLocator`, `RegulatoryStatusReader`) advertised-without-dispatch and narrowed by `is*` guard, per ADR-042 decision 5.

`RegulatoryStatusReader` and **not** `RegulatoryTransmitter` is also the right split and for the stated reason — the vendor relays to the hub on the seller's behalf rather than OL holding the authority session, which is exactly where `InfaktInvoicingAdapter` sits.

### ADR-026's country-agnostic rule holds by construction

I grepped the diff for `libs/core` paths: **there are none.** Every Polish specific — the rate letters, the `eInvoice`/`eCorrectiveInvoice` document kinds, the hub relay semantics — lives in `libs/integrations/eparagony`. That is the constraint satisfied structurally rather than by discipline.

### The residual allocation is correct

`allocateGroupNet` (`eparagony-invoice.mapper.ts:657`) is the part worth the most scrutiny, and it survives it. The invariant it must hold is `0 <= net_i <= gross_i`, because the wire carries the line's net and derives tax as `gross_i - net_i` — so a net above the line's own gross emits a **negative tax** onto a document bound for a tax authority, and `calculationValidation: NONE` means the vendor does not catch it either. Both directions are now clamped, and the docblock's claim that neither repair can run out is true:

- **Surplus**: earlier headroom is `Σ(g_i − a_i)` for `i < last` = `(grossGroup − netGroup) − (g_last − a_last)` = `taxGroup + surplus`, and `taxGroup >= 0`, so headroom `>=` surplus. ✔
- **Shortfall**: `Σ(a_i, i < last)` = `netGroup + shortfall > shortfall` since `netGroup >= 0`. ✔

The worked reproducers are also not constructed: `toShippingLines` appends a per-rate share **last** and `groupByRate` preserves first-seen order, so the 3/3/1-grosz case is reachable on an ordinary order whose shipping rate already appeared above it. Declining to refuse those orders — on ANALYSIS-1032's "a permanent domain rejection of a real, paid order is the defect" reasoning — is the right call, and the honest disclosure that a lending line's *effective* rate moves (a line can end at 0% or 100%) is better than a figure the arithmetic cannot support.

Reporting `IssueInvoiceResult.documentLines` back to core follows from this directly: without it core falls back to a per-line `round(gross / (1 + r))`, which is the exact arithmetic the mapper rejects, so OL's own contents card would disagree with the document by a grosz on any order that needed a residual absorbed. That is the reason to carry the field, and it is stated.

### Findings

**[IMPORTANT]** — branch topology, not the code
> This PR and **#3267** both target `3192-eparagony-invoicing`, share three commits, and **neither is an ancestor of the other**. `git diff pr3267 pr3260` is 31 files, +1372/−877 — both add `eparagony-invoicing.adapter.ts`, `eparagony-invoice.mapper.ts`, `invoice-tax-rate.policy.ts` and `eparagony-connection-config-shape-validator.adapter.ts` with divergent content. Whichever lands second conflicts across the whole slice and must be resolved by intent, not textually. This PR is the superset on the arithmetic (it carries `3abbcc63b`'s both-directions clamp and `normalizeDecimalSeparator`, which #3267 lacks) while #3267 is the superset on the wiring (`eparagony-adapter.factory`, the plugin manifest). Please state the intended merge order in both PR descriptions; the last commit here ("adopt slice 2's stronger seller-config validation so the stack converges") suggests the plan exists but it is not written down where a reviewer or a merger can see it.

**[SUGGESTION]** — `invoice-tax-rate.policy.ts:170`
> `normalizeDecimalSeparator` is deliberately not `parseFloat` because `parseFloat('23,5')` is `23` — correct, and worth keeping. The docblock says no shipped `ProductMaster` is known to write comma decimals, so this closes a divergence with core's own `parseTaxRatePercent` rather than a reported defect. Given that, consider whether the divergence is better closed in `parseTaxRatePercent` instead, so every adapter benefits rather than this one; not a change to make here.

**[Positive]** — every composition failure happens before the boundary (`EparagonyConfigException`, `failureMode: 'rejected'`, nothing sent), the buyer's tax number travels verbatim per ADR-073 decision 5 with the vendor's checksum refusal surfacing as an ordinary classified error rather than a pre-emptive local judgement, and `CorrectionIssuer` is withheld with a stated reason rather than advertised and unimplemented. That last one is the #980 rule applied correctly: a capability is claimed together with the ability to deliver it.

### What I read closely, and what I did not

Read closely: the plugin manifest and dispatch wiring, `eparagony-invoicing.adapter.ts`'s header and contract surface, `allocateGroupNet` and the surrounding rate-grouping, `invoice-tax-rate.policy.ts`, `index.ts`'s barrel additions, and the branch topology against #3267. Skimmed: the 1054-line mapper spec and the 582-line adapter spec (I checked they exist and target the documented edge cases rather than reading every assertion), and the README.

### Merge readiness

✅ **Approve.** No BLOCKING issues. The one IMPORTANT item is coordination between this PR and #3267 rather than a defect in this diff — but it does need answering before either merges, because the conflict is large and touches money arithmetic where a wrong textual resolution is silent.

norbert-kulus-blockydevs:
## Response to the final review

Thanks for the approve. Both remaining items addressed:

**IMPORTANT (branch topology)** - added a "Merge order" section to both this PR's body and #3267's body, stating explicitly: #3260 merges first, #3267 rebases second (its validator commit `0b1d75492` resolves as already-applied since this PR adopted that validator verbatim in `7ff544100`), #3269 merges third. No code change - it was a coordination gap, now closed in both descriptions rather than left to the merger to reconstruct from the review thread.

**SUGGESTION (`normalizeDecimalSeparator` vs `parseTaxRatePercent`)** - agreed this package's version should stay; you noted yourself it's "not a change to make here" since closing the divergence belongs in core's `parseTaxRatePercent`, benefiting every adapter rather than just this one. No change made in this PR; worth its own issue if the divergence turns out to matter for a real source.

No code diff in this pass - both findings resolved as documentation/coordination, not implementation.

piotrswierzy (approved):
## PR Review — #3260 `feat(eparagony): issue invoices through InvoicingPort over the eInvoice document`

Reviewed against `docs/architecture-overview.md` §§14 Invoicing / 16 Fiscalization / 17 Sales Documents, ADR-026, ADR-041, ADR-042, ADR-063, ADR-073, and the three shipped `InvoicingPort` adapters (KSeF / inFakt / Subiekt) as the baseline.

---

### Step 1 — Context

`EparagonyInvoicingAdapter implements InvoicingPort, RegulatoryStatusReader` over the vendor's `eInvoice` document kind, plus `eparagony-invoice.mapper.ts`, a second `invoice-tax-rate.policy.ts` beside the untouched receipt policy, the shared `GET /documents/{token}/status` transport, and `scripts/check-eparagony-deadline-mirror.mjs`. Nothing is reachable at runtime in this slice (manifest/dispatch/factory are #3267), which is the right split.

I verified the money arithmetic by hand rather than by reading its docblock, and it is correct.

### Positive observations worth recording

- **`allocateGroupNet`'s two repairs are provably sufficient, in both directions.** For the surplus branch the earlier headroom is `Σ(gᵢ − aᵢ) = grossGroup − g_last − netGroup + a_last = taxGroup + surplus`, and `taxGroup ≥ 0`; for the shortfall branch the earlier allocation is `netGroup + shortfall`, and `netGroup = grossGroup − taxGroup ≥ 0`. Neither loop can run out. Each `lent = min(headroom, surplus)` and `borrowed = min(aᵢ, shortfall)` keeps every earlier line inside `[0, gᵢ]` too, so the emitted per-line `taxValue = gross − net` is never negative. `eparagony-invoice.mapper.spec.ts:255` pins that as a per-line `0 ≤ net ≤ gross` invariant over four shapes — including the `[0.03, 0.03, 0.01]` case a one-directional clamp misses, which the docblock correctly notes is reachable through `toShippingLines` appending a per-rate share LAST while `groupByRate` preserves first-seen order.
- **Emitting in command order while grouping by rate for the arithmetic is the correct resolution of a real conflict.** `InvoiceService` persists `issuedLineSnapshot.lines` as `cmd.lines` verbatim and `buildContent` (`invoice.service.ts:1308`) keys `documentLines` by 1-based position, as does `CorrectionLine.originalLineNumber` — so rate-group order would have #3193 credit the wrong line on a transmitted document. The `netByIndex` read-back is the right mechanism.
- **Returning `documentLines` matters more here than the docblock lets on.** Without it `buildContent` falls back to a per-line `round2(gross / (1 + fraction))`; Subiekt is the in-tree adapter that omits it and accepts that fallback. The `[0.01, 0.01, 0.01]` spec case is a genuine, reachable divergence.
- **The failure taxonomy matches the baseline exactly.** `EparagonyConfigException.failureMode = 'rejected'` for every pre-boundary refusal; `EparagonyNetworkError.failureMode = 'in-doubt'` for the poll-budget exhaustion — the same direction inFakt takes on its 504 correction-poll timeout, and the correct reading under `classifyFailure`'s fiscal-safe default (`invoice.service.ts:947`). A poll timeout is not treated as a rejection anywhere.
- **The `DOCUMENT_ALREADY_EXISTS` swallow is right.** The token is a deterministic derivation of `(connectionId, registrationKey)`, so "already exists" means our own earlier attempt landed; reporting a rejection there would be the wrong-direction error.
- `check-eparagony-deadline-mirror.mjs` is a guard that can actually fail — `parseNumericConst` returns `null` (a failure, not a pass) on an unreadable declaration, and `--self-check` exercises the equal/above/unreadable cases.

---

### Step 2/3 — Findings

**[IMPORTANT]** — `libs/integrations/eparagony/src/infrastructure/adapters/eparagony-invoicing.adapter.ts` (`issueInvoice`, `pollToSettledIssuance`) + `scripts/check-eparagony-deadline-mirror.mjs`

> `EPARAGONY_ISSUE_DEADLINE_MS` is treated by the mirror script as a whole-call wall clock — its failure message says a drift lets "one call outlive the in-flight lease and issue the same sale twice" — but the code enforces it only as a **poll-phase** budget. `deadline` is computed before `createDocument`, and only `pollToSettledIssuance` reads it, and it reads it *between* status reads rather than around one.
>
> With `DEFAULT_RETRY_CONFIG = { maxRetries: 2, initialDelayMs: 500, backoffMultiplier: 2 }` and `REQUEST_TIMEOUT_MS = 15_000` (`eparagony-http-client.ts:53`, `eparagony-http-client.types.ts:24`), `createDocument` alone can burn ≈ 46.5 s, and the final status read can begin at `pollUntil − 1 ms` and burn another ≈ 46.5 s. Worst case ≈ **138 s** against the 120 s `MAX_SUPPORTED_PROVIDER_TIMEOUT_MS` the mirror exists to keep it under.
>
> This matters more on this lane than on the receipt lane, and the asymmetry is the finding: `FiscalRegistrationService` races a hard `MAX_SUPPORTED_PROVIDER_TIMEOUT_MS` timeout around every provider call (`fiscal-registration.service.ts:846-865`), whereas `InvoiceService` races nothing — there is no `Promise.race` / `setTimeout` anywhere in `invoice.service.ts`. So on the invoicing lane the adapter's own constant is the **only** bound, and it does not bound what it claims to.
>
> Not BLOCKING: `ISSUING_LEASE_MS = 300_000` (`invoice.service.ts:144`), so 138 s leaves the double-issue guard intact. But the invariant is enforced by comment, not by construction, which is precisely what the mirror script's own header says it exists to stop. Either wrap the whole method in an `AbortSignal.timeout(EPARAGONY_ISSUE_DEADLINE_MS)` / `Promise.race`, or state the true worst case in the constant's docblock so the next person lowering `ISSUING_LEASE_MS` sees a real number.

**[IMPORTANT]** — `eparagony-invoicing.adapter.ts` (`resolveRegistrationKey`, `toTerminalRejection`)

> The adapter's own comment concedes the consequence — *"A genuine retry needs a fresh key"* — but **nothing in the product mints one**, so a vendor-`ERROR` invoice is permanently unissuable on that connection through every in-product affordance.
>
> `resolveRegistrationKey` falls back to `invoice:{connectionId}:{orderId}`, and that is exactly the key core supplies on every automated path: `invoicing.controller.ts:717` builds `` `invoice:${connectionId}:${orderId}` `` for bulk issue, the auto-issue trigger uses the same form, and the failed-record retry explicitly reuses `record.idempotencyKey` (`invoicing.controller.spec.ts:1043`). So after `toTerminalRejection` an operator who corrects the buyer's tax number and clicks Retry gets: same key → same `documentToken` → `DOCUMENT_ALREADY_EXISTS` → swallow → poll re-reads the **same errored document** → same terminal rejection. Forever.
>
> This is not a regression the PR introduces — the receipt lane has the identical shape (`eparagony-fiscalization.adapter.ts:112,158,235`) — but the invoice lane is where a correctable input (buyer tax number, buyer address, an exemption code) is the common rejection cause, so it will be hit. `failureMode: 'rejected'` means the order is not blocked elsewhere (`InvoiceRecord.blocksIssuanceElsewhere`), so re-routing to another connection works; the same connection does not. The one working escape is `POST /invoices` with an explicit `idempotencyKey`, which no UI offers. Please name that escape in the refusal `message` (which *is* persisted as `errorMessage`) or file it.

**[IMPORTANT]** — `libs/integrations/eparagony/src/domain/policies/invoice-tax-rate.policy.ts` (`resolveInvoiceTaxRateCode`)

> A neutral `'0'` resolves unconditionally to `'ZRD'`, the vendor's **domestic** 0% slot. The docblock says *"a caller that means intra-community supply or export says `ZRICS` / `ZRE` instead"* — but under ADR-063 / `InvoiceLine.taxRate` the neutral percent-as-string vocabulary is `23 | 8 | 5 | 0 | zw | np | oo`, which has no way to express either, so **no caller can**. An intra-EU supply and an export are therefore declared as a domestic zero rate on a document relayed to the national hub.
>
> The only reachable workaround is storing the literal vendor code as the product's tax rate (`isInvoiceTaxRateCode(upper)` passes it through verbatim), which is undocumented, provider-specific, and would fall out of `netSalesRateFractionSql`'s known-fraction set. This is a real tax-treatment claim rather than a formatting detail, so it deserves either an explicit "known gap, no neutral vocabulary exists" line in the policy docblock, or a follow-up issue against the neutral rate vocabulary. ADR-026 is what forbids solving it in `libs/core` by adding `ZRE`, so the follow-up is a vocabulary decision, not a quick fix.

**[SUGGESTION]** — `eparagony-connection-config-shape-validator.adapter.ts` (`validateMerchantAddress` docblock)

> *"Every part is validated here because NOTHING downstream does: `toSellerEntityAddress` copies the operator's object across field by field with no interpretation"* — no longer true after the #3274 fix in this PR: `readSellerEntityAddress` reads every required part through `readNonEmpty` and returns `null` on a partial, and `toSellerEntityAddress` no longer exists under that name. The validator is still right to check (JSONB + the raw editor + connections configured before it shipped), just not for the stated reason.

**[SUGGESTION]** — `RegulatoryLocateCriteria` gap, named in the PR body

> The stated reason the invoice lane cannot implement `RegulatoryRecordLocator` is correct: `PendingRecoveryService.recoverOne` (`pending-recovery.service.ts:275`) hands the locator only `RegulatoryLocateCriteria` (`sellerTaxId` / `documentNumber` / `issuedFrom` / `issuedTo`), never the record, and the token derives from `(connectionId, idempotencyKey)`. So a crashed `issuing` record goes straight to in-doubt with no recovery, on the same connection whose receipt lane recovers fine — a real fiscal-safety asymmetry. The remedy is one optional field (`idempotencyKey?: string`) on that neutral criteria type, which is additive and country-agnostic. Worth filing explicitly rather than leaving in a PR body.

**[SUGGESTION]** — `eparagony-invoice.mapper.ts` (`toEntityAddress`, `splitStreetAndNumber`)

> The `ul. Kwiatowa 12 m. 8` mis-split is pinned by a spec that *asserts the mis-assignment*, which is the honest thing to do. Consider a second pattern arm for the `m.` / `lok.` form, since it is the common Polish shape and the value lands in the vendor's building-number field on a hub-bound document. Not blocking — the round trip is lossless and refusing a paid order would be worse.

**Documentation:** `docs/architecture-overview.md` §14 does not yet mention that a single connection can hold both `Fiscalization` and `Invoicing`, nor the `CONFIRMED + KSEF → accepted` decision. #3267 touches that file; make sure the mapping decision lands there too, since the issue body's original table said `cleared` and the three reasons given for `accepted` (absent from `TerminalRegulatoryStatusValues`, filtered out by name in `invoices-list-page.tsx`, and KSeF's own adapter using `accepted` for the same authority's terminal success) are all correct and worth preserving outside a PR thread.

---

### Step 4 — Final Assessment

#### Summary

A large, unusually rigorous adapter slice. The part that carries the most risk — per-rate-group VAT derivation with a per-line allocation whose residual can escape `[0, gross]` in either direction — is correct, provably bounded, and pinned red-first. The error taxonomy matches the three shipped adapters, every composition failure lands before the provider boundary as `rejected`, the poll-budget exhaustion is `in-doubt` rather than a rejection, and the deterministic `Idempotency-Key` makes the re-issue-on-5xx path safe. The neutral-vocabulary line holds: no `NIP`/`KSeF`/`FA` terminology leaves the plugin, and the adapter supplies the `pl-nip` scheme tag itself exactly as ADR-073 decision 1 requires. My single most important concern is that the fiscal-safety constant the new invariant script guards bounds the poll phase rather than the whole call, on the one lane where core races no timeout of its own.

#### Merge Readiness

🔄 **Approve with changes** — no BLOCKING issue; the three IMPORTANT items are follow-ups rather than merge stoppers, but each should get an issue number before this lands on `main` via the integration branch.

#### Priority fixes

1. Bound the whole `issueInvoice` call at `EPARAGONY_ISSUE_DEADLINE_MS` (or write the real ≈138 s worst case into the constant's docblock), so the mirror script's own failure message stops describing a property the code does not enforce.
2. Give a vendor-`ERROR` invoice a reachable retry: name the explicit-`idempotencyKey` escape in the refusal `message`, or file the fresh-key path.
3. Record the `'0' → ZRD` gap as a known limitation with a follow-up against the neutral rate vocabulary.
4. Correct the config-shape validator's "NOTHING downstream does" docblock.
5. File the `RegulatoryLocateCriteria.idempotencyKey` widening so the invoice lane can reach parity with the receipt lane's crash recovery.

PR #3261, opened 2026-09-15
feat(inventory): nav entry under Diagnostics

### LINKED ISSUE #3076: feat(inventory): nav entry under Diagnostics
## Problem / Context

The page from prior sub-issues has no route or nav entry.

## Proposed Solution

Add a route + register it under the Diagnostics nav group in `apps/web/src/app/nav-registry.ts`, admin-gated to match the route's `@Roles('admin')`.

## Classification

**Type**: Frontend
**Layer**: Interface (routing)
**File(s)**: apps/web/src/app/routes/duplicate-positions.route.tsx, apps/web/src/app/nav-registry.ts

## Dependencies

- Part of #3071

## Assumptions

- None

## Acceptance Criteria

- [ ] Page is reachable from the sidebar for an admin

jakubretajczykBD:
## Summary
- Adds `duplicate-positions.route.tsx` (lazy-loaded, crumb-registered under the Diagnostics group) and wires it into `root.route.tsx`'s `coreChildren` plus a `BASE_NAV_GROUPS` entry under Diagnostics.
- **Admin gate**: the backend route is `@Roles('admin')` while the rest of the Diagnostics group is admin+operator+viewer. `LiveNavItem` has no fitting existing `Permission` for a per-item gate, and a whole-group gate would hide the three admin+operator+viewer siblings — so `LiveNavItem` gains a new, item-level `requiresRole?: 'admin'` (`buildNavGroups`'s filter drops the item, not the group, for a non-admin session). The page itself already self-gates (its own docblock: renders an access-denied state inside its own `PageLayout`, and its query hooks are `enabled: isAdmin`, which is what actually prevents the 403 round-trip) — a non-admin who navigates directly to `/duplicate-positions` (nav item hidden or not) sees that same denial, never a raw backend 403.

## Context
Fifth sub-issue of epic #3071, stacked on #3256 (the truncated-report banner). Addresses the admin-gate requirement flagged in #3073's review fix.

⚠️ **Stacked PR** — base branch is `3075-duplicate-positions-truncated-banner` (#3256), not `main`. Merge #3256 first.

## Test plan
- [ ] `pnpm type-check`
- [ ] `pnpm lint`
- [ ] Manual: log in as a non-admin (operator/viewer) — confirm the "Duplicate stock positions" nav item does not appear under Diagnostics
- [ ] Manual: log in as a non-admin and navigate to `/duplicate-positions` directly — confirm the page's own access-denied state renders, and no `getDuplicatePositions`/`getProvenanceBackfillStatus` request fires
- [ ] Manual: log in as admin — nav item appears, page loads normally

Closes #3076
Part of #3071


piotrswierzy (approved):

**Verdict: 🔄 Approve with changes.** The route side is exemplary; the nav side leaves an affordance every operator can see and only an admin can use.

### What is right

All three route contracts from `docs/frontend-architecture.md` are satisfied, and I checked each rather than assuming:

- **Crumb handle** — `handle: duplicatePositionsCrumb` typed `RouteCrumbHandle`, declared on the **index child** rather than the parent, which is what the deepest-first `useMatches()` walk needs. Pinned by `route-handle.test.ts`.
- **Lazy** — the page is behind `lazy:`, and `EXPECTED_LAZY_ROUTE_COUNT` is bumped 66 → 67 in the same PR, so `route-lazy.test.ts` stays honest rather than being silenced.
- **Admin check via `useIsAdmin()`**, not an inline `session.user?.role === 'admin'`. The docblock names the exact reason — `role` is typed `string`, so a typo compiles and silently evaluates false — which is the rule `docs/frontend-architecture.md` § Access Control states. `useIsAdmin` being "the one place the admin role name is spelled" is the right primitive here.
- **The hydration window is handled correctly**: `RequireAdmin` returns `null` while `isReady` is false, so a non-admin branch never flashes before the session resolves. That is `AccessGate`'s own rule ("not known yet" is not "denied") applied consistently rather than reinvented.
- Wrapping inside `lazy()` rather than at the route object keeps the guard in the same chunk as the page it guards.

### Findings

**[IMPORTANT]** — `apps/web/src/app/nav-registry.ts:67`
> The nav item is added to the Diagnostics group with **neither `requiresRole` nor `requiresPermission`**, so every authenticated session renders "Duplicate stock positions" in the sidebar, clicks it, and lands on `RequireAdmin`'s "Admin access required" `ErrorState`. The route guard is correct; the affordance leading to it is not gated at all.
>
> `docs/frontend-architecture.md` is explicit that the declarative gate exists to *hide the nav affordance* ("Authorization is still enforced backend-side; the gate only hides the nav affordance"), and `nav-registry.ts` already implements both halves — group-level `requiresRole === 'admin'` filtering at :136 and item-level `requiresPermission` filtering at :153.
>
> `require-admin.tsx`'s docblock argues persuasively against minting a new `Permission` for one read-only diagnostic, and I agree with that. But the conclusion that follows is an **item-level `requiresRole`**, mirroring the group-level gate that already exists eight lines away — not leaving the item ungated. That is a small addition to `NavRegistryItem` plus one clause in the same `.filter` chain, and it is the shape `NavContribution.requiresRole` already uses for plugin items.
>
> Left as-is, the nav grows an entry that is dead for every non-admin operator, which is the state the declarative gate was introduced to prevent.

**[SUGGESTION]** — `apps/web/src/app/routes/require-admin.tsx`
> `RequireAdmin` is a genuinely reusable primitive (a route whose endpoint is `@Roles('admin')` with no fitting `Permission` is not going to stay a population of one). It currently sits in `app/routes/`. If a second consumer appears, `shared/auth/` beside `useIsAdmin` is the more natural home — noting it now so the second caller does not copy it instead.

**[Note on the stack]** — this PR also carries an unrelated one-line fix to `order-fulfillment-tasks-panel.test.tsx` (narrowing a `findByText` with `{ selector: '.toast__description' }`). It is a reasonable fix, but it is unmentioned in the PR body and appears identically in #3262 and #3264 as well. Worth a line in the description so a reviewer of the later PRs knows it is the same change riding down the stack rather than three separate edits.

### Merge readiness

🔄 **Approve with changes** — all eight checks green, route contracts clean. The nav gate is the one thing I would want fixed before this ships, since it is user-visible on every non-admin session. Stacked on `3075-duplicate-positions-truncated-banner`; merge bottom-up.

piotrswierzy (approved):
# Tech Lead review — #3261 `feat(inventory): nav entry under Diagnostics`

Reviewed against `docs/frontend-architecture.md`, `docs/frontend-ui-style-guide.md`, `docs/engineering-standards.md`, and the `#2319` read-only/admin-only framing in `docs/architecture-overview.md` § Inventory. CI is green on all 8 checks.

## Step 1 — Context

Adds `/duplicate-positions` to `coreChildren`, registers a `Diagnostics` crumb, adds a `BASE_NAV_GROUPS` entry, and introduces two new gating mechanisms: a per-item `requiresRole` filter in `buildNavGroups` and a `RequireAdmin` route wrapper. Baseline: `Diagnostics` is otherwise an ungated group (`nav-registry.ts:62-68` on `main`), while `GET /inventory/duplicate-positions` is `@Roles('admin')`.

The core question the task poses — *is the route gated to the same audience as the API?* — is answered **yes, three times over**. That is the finding.

## Step 2/3 — Findings

**[IMPORTANT]** — `apps/web/src/app/routes/require-admin.tsx:36-48` + `apps/web/src/pages/inventory/duplicate-positions-page.tsx:203-214` (on the base branch)

The page **already self-gates**. Its docblock on the base branch states this explicitly:

> "The backing endpoints are `@Roles('admin')`-gated server-side, and the page gates itself the same way — rendering an access-denied state for a non-admin session rather than relying on a sibling PR's nav gate, which covers only the nav affordance and not a direct navigation to this route."

`RequireAdmin` now short-circuits **outside** that branch, so the page's own denial state is unreachable dead code, and the repo ends up with two different spellings of one denial:

| | eyebrow | title | message |
|---|---|---|---|
| `RequireAdmin` (wins) | `Access restricted` | `Admin access required` | `This diagnostic is available to admin accounts only.` |
| page (now dead) | — | `Admin role required` | `This diagnostic reads inventory_items directly — it requires an admin session.` |

There is a behavioural regression hiding in that too: the page's version renders **inside `PageLayout`**, so a denied operator still gets the eyebrow, title and page chrome; `RequireAdmin` renders a bare `ErrorState` with no `PageLayout`, so the denial appears frame-less. Pick one gate. Given the page's own docblock argues for the page-level one and the query hook is already `enabled: isAdmin` (`use-duplicate-positions-query.ts`) — which is what actually prevents the 403 round-trip, not `RequireAdmin` — dropping `RequireAdmin` and keeping the page's branch is the smaller surface. Whichever you keep, delete the other and correct the docblock that describes it.

**[IMPORTANT]** — `apps/web/src/app/nav-registry.ts:154-162`, `apps/web/src/app/nav-registry.types.ts:55-67`

The new item filter is

```ts
(item.requiresRole === undefined || (item.requiresRole === 'admin' && isAdmin))
```

typed as `requiresRole?: Role`, where `Role = 'admin' | 'operator'` (`nav-registry.types.ts:31`). An item declaring `requiresRole: 'operator'` — a legal value — is therefore filtered out for **every** session, operators included, with no error. The docblock claims it "mirrors `LiveNavGroup.requiresRole`"; it does not. The group branch (`nav-registry.ts:141`) tests `group.requiresRole === 'admin' && !isAdmin`, so a group declaring `'operator'` falls through and is shown to *everyone* — the two gates fail in opposite directions from the same annotation.

This matters because `RoleValues`' own docblock states the reason `'operator'` exists (`nav-registry.types.ts:28-30`): *"so nav groups can declare `requiresRole: 'operator'` for operator-only sections without any further type changes."* That claim is now false at item granularity, and the failure is silent.

Cheapest correct fix: type the item gate `requiresRole?: 'admin'` and say so, which makes the unsupported case a compile error instead of an invisible one. Otherwise resolve the session role properly for both values.

**[SUGGESTION]** — `apps/web/src/features/fulfillment/components/order-fulfillment-tasks-panel.test.tsx:331`

Unrelated to this PR's subject. It also pins an assertion to a CSS class (`selector: '.toast__description'`), which couples the test to styling; a `within(...)` on the toast's role, or a `data-testid`, survives a class rename. If it is a pre-existing flake fix, it belongs in its own commit with the reason in the message.

**[SUGGESTION]** — `apps/web/src/app/routes/require-admin.tsx`

No colocated test, while the nav gate got three. `engineering-standards.md` § Tests asks for a unit test on non-trivial logic; the `isReady`/`isAdmin` three-way is exactly that, and the "renders neither branch while hydrating" property is the one a future refactor is most likely to break.

The body says a non-admin "follows the (still-visible) nav link". The diff gates the nav item too, so the description describes an earlier revision.

### Positive observations

- `useIsAdmin()` rather than an inline `session.user?.role === 'admin'` — exactly the case `use-permission.ts:75-92` documents it for, and correctly not paired with a permission since none implies this route.
- `if (!isReady) return null` honours `useIsAdmin`'s own warning that it returns `false` while hydrating and must not be read as a denial.
- `nav-registry.test.ts:80-87` asserts the **siblings are not dropped** for a non-admin. That is the assertion that actually catches the plausible regression (a mis-placed filter collapsing the whole group), not just the happy path.
- `EXPECTED_LAZY_ROUTE_COUNT` bumped — the contract test was honoured rather than worked around.

## Step 4 — Final Assessment

### Summary
A small, well-argued gating change that lands in the right layers and reuses the sanctioned `useIsAdmin` seam. Its one real problem is over-delivery: the page it guards was already guarding itself, so the PR ships a third gate whose only visible effect is to make the page's own denial state unreachable and render it without page chrome. Second is that the new item-level `requiresRole` accepts a `Role` value it silently mishandles in the hide-everything direction.

### Merge Readiness
🔄 **Approve with changes** — no BLOCKING issues; the two IMPORTANT items are follow-ups that should land before this reaches `main` behind the stack.

### Priority fixes
1. Resolve the duplicate admin gate — keep exactly one of `RequireAdmin` or the page's own branch, delete the other, and fix the stale docblock. (IMPORTANT)
2. Narrow the item-level `requiresRole` to `'admin'`, or handle `'operator'`; correct the "mirrors `LiveNavGroup.requiresRole`" claim either way. (IMPORTANT)
3. Split the `order-fulfillment-tasks-panel.test.tsx` change out, or justify it in the body. (SUGGESTION)
4. Refresh the PR description — the nav item is no longer "still-visible". (SUGGESTION)

piotrswierzy (approved):
## Re-review (post force-push) — `b20134b`

Both prior IMPORTANT findings are genuinely fixed in the code, not just in a reply. Re-read against `main` as of `e1ae0ae`.

---

### 1. Prior findings

|---|---|---|---|
| 1 | IMPORTANT — the `RequireAdmin` route wrapper made the page's own denial branch dead code, and rendered a second, frame-less denial outside `PageLayout` | **Addressed** | `apps/web/src/app/routes/duplicate-positions.route.tsx:17-26` — the wrapper is gone; `lazy()` returns the page component directly, and the comment records *why*. The surviving denial is the page's own, inside `PageLayout` (`duplicate-positions-page.tsx:219-232` on this head). One gate, one copy, framed. |
| 2 | IMPORTANT — item-level `requiresRole?: Role` accepts `'operator'`, which the filter silently drops for **everyone**, failing the opposite way to the group-level gate | **Addressed** | `apps/web/src/app/nav-registry.types.ts:75` — narrowed to `requiresRole?: 'admin'`, with the docblock (`:55-74`) stating exactly the mismatch it prevents and the condition for widening it. `nav-registry.ts:153-160` now tests `requiresRole === 'admin' && isAdmin`, so the type and the filter agree by construction rather than by convention. |

Previously verified and still intact: the route is gated to the same audience as the API three ways over — the nav item (`nav-registry.ts:67-71`), the page self-gate, and `enabled: isAdmin` on both query hooks (`use-duplicate-positions-query.ts:38`, `use-provenance-backfill-status-query.ts:36`), which is what actually prevents the 403 round-trip.

The three new `nav-registry.test.ts` cases (`:63-88`) are real guards: `hides … for a non-admin` goes red against a missing filter clause, and `never drops the sibling items for a non-admin` goes red against someone "fixing" this with a group-level `requiresRole` — which is precisely the wrong fix the type docblock warns about.

---

**[SUGGESTION]** — `apps/web/src/features/fulfillment/components/order-fulfillment-tasks-panel.test.tsx:332`
> `findByText` → `findToastDescription` in an unrelated feature's stale-token 409 test, in a PR whose stated scope is a Diagnostics nav entry. The change itself is right (`findToastDescription` is the shipped helper, `test-utils.tsx:1191`, and scoping the query to the toast is more precise than a bare text match), but it is unexplained in the PR body and unrelated to the nav registry. Either call it out in the description or split it — an unexplained edit in a third feature's test is the kind of thing that makes a later bisect harder than it needs to be.

**[SUGGESTION]** — `apps/web/src/app/routes/route-lazy.test.ts:98`
> `EXPECTED_LAZY_ROUTE_COUNT` 66 → 67 is correct and is the documented contract of that test. No action; noting it so the count bump is on the record as intentional rather than a merge artefact.

---

**The shared `apps/web` suite is red on this stack, and the cause sits below this PR.** `Test (web)` on #3262 and #3264 both fail the same 6 tests:

```
  ❯ useDuplicatePositionsQuery src/features/inventory/hooks/use-duplicate-positions-query.ts:31
```

The `enabled: isAdmin` gate was added to both hooks on the `3075-duplicate-positions-truncated-banner` base (verified: the hooks carry `useIsAdmin()` at that ref), while both hook test files still build their own wrapper from `ApiClientProvider` + `QueryClientProvider` only, with no `SessionProvider` (`use-duplicate-positions-query.test.tsx:16-27` at this head). Neither test file is touched by #3261, #3262 or #3264 — **the fix belongs on #3256 / the `3075-…` base**, where the gate was introduced. Use `renderWithProviders`'s session-aware wrapper, or add `SessionProvider` to the local one.

This PR's own `Test (web)` job is still queued at the time of writing, but it carries the identical hooks and the identical test files, so it will fail for the same reason. **Do not merge any PR in this stack until that is green at the base.**

---

### 4. Summary

Adds the `/duplicate-positions` route and its Diagnostics nav entry, gated to admins. Both prior IMPORTANT findings are fixed in the code and the fixes are the right shape: one denial state inside `PageLayout`, and a type narrowed so the item gate cannot be handed a value its filter would silently drop for everyone. The nav tests would go red against both the regression and the wrong fix for it. The only thing standing between this and merge is inherited, and lives one branch down.

### Merge Readiness

✅ **Approve** — conditional on the stack-level gate above. Nothing in this diff needs to change.

### Priority fixes

1. (Base, not here) Give the two inventory hook test files a `SessionProvider` so `Test (web)` goes green for the whole stack.
2. (Optional) Explain or split the `order-fulfillment-tasks-panel.test.tsx` edit.

piotrswierzy (approved):
## Tech Lead review — `feat(inventory): nav entry under Diagnostics`

Two good type/architecture calls in a small diff.

### ✅ `requiresRole?: 'admin'` narrowed rather than typed `Role`

This is the best thing in the PR, and the docblock earns every line:

> *`buildNavGroups`'s item filter only ever tests `requiresRole === 'admin'`, so a `LiveNavGroup`-style `Role` value here would silently type-check and then be filtered out for EVERY session, including admins — the opposite of what the group-level gate does with the same value. Narrowing the type to `'admin'` turns that mismatch into a compile error instead of an invisible one.*

That's exactly right, and it's the failure direction that matters: a nav item that disappears for *everyone* including the admins who need it, with no error anywhere, is the kind of bug found weeks later by someone asking "where did that page go?". Narrowing the type to what the filter actually implements — and saying when to widen it — makes the type honest about the code behind it.

Also the right call not to mint a new `Permission` for a single read-only diagnostic. `PermissionValues` is a shared vocabulary and widening it for a population of one is how it ends up meaningless.

### ✅ Removing the route-level `RequireAdmin`

> *made the page's own denial state unreachable dead code and rendered a frame-less denial with no PageLayout*

Agreed. Two gates for one rule is worse than one when the outer gate makes the inner one unreachable — the page's own branch would have rotted untested and untestable, and the user-visible result was a worse denial (no shell, no breadcrumb). Since #3252's hooks are `enabled: isAdmin`, the 403 round-trip is prevented regardless, so the route guard bought no protection the page didn't already have. Leaving the reasoning in the route file, where the next person will be tempted to re-add it, is the right place for it.

### ✅ Nav tests

Three cases, and the third — *"never drops the sibling items for a non-admin"* — is the one that matters. A per-item gate implemented as a group gate would pass the first two tests and quietly hide Jobs & Logs, Webhooks and Cursors from every operator and viewer. That's the regression this file exists to catch.

### 🟢 SUGGESTION — heads-up on a guaranteed conflict

`EXPECTED_LAZY_ROUTE_COUNT` 66 → 67 here. #3309 takes the same constant 66 → 68 from the same base. Whichever lands second will conflict, and the resolution is **69**, not "take theirs". Worth flagging in whichever PR merges second so nobody resolves it by picking a side — the count is a contract, and silently landing on 67 or 68 makes the assertion wrong in a way that only fails once someone adds route 68.

### 🟢 SUGGESTION — one unrelated line

`order-fulfillment-tasks-panel.test.tsx` switches `screen.findByText` → `findToastDescription` (an existing `test-utils` helper, five other call sites on `main`). It's almost certainly a correct fix, but it has nothing to do with a Diagnostics nav entry and isn't mentioned in the description. If it's a flake fix, it deserves a one-line note saying so — otherwise a future bisect through this commit is looking at a nav change and a toast assertion with no stated connection.

### Notes

- Route file follows the nested-index + crumb shape and is lazy, consistent with its siblings.
- CI: six green, `Lint` / `Type Check` / `Integration Tests` queued at review time.

**Verdict:** ✅ Approve.

PR #3262, opened 2026-09-15
test(inventory): duplicate-positions page tests

### LINKED ISSUE #3077: test(inventory): duplicate-positions page tests

jakubretajczykBD:
## Summary
- Adds colocated tests for `DuplicatePositionsPage` covering: loading/error states for both the duplicate-positions and provenance-backfill-status reads independently, the two-condition readiness checklist (ready only when both `groupCount === 0` AND the backfill is complete — including the "zero duplicate groups but backfill still running" case), the #3239 display enrichment (product name/sku shown, falls back to the raw id when unresolved), the expand/collapse interaction on a group row, and the truncated-report banner.
- Assertions updated to match the operator copy after #3073's review fix (issue numbers dropped from user-facing text).

## Context
Sixth sub-issue of epic #3071, stacked on #3261 (the nav entry / admin gate). Closes out the epic's own test coverage — a mockup-parity follow-up (#3264) is stacked on top of this one, without its own numbered issue.

⚠️ **Stacked PR** — base branch is `3076-duplicate-positions-nav-entry` (#3261), not `main`. Merge #3261 first.

## Test plan
- [ ] `pnpm test` — 9/9 passing locally
- [ ] `pnpm type-check`
- [ ] `pnpm lint`

Closes #3077
Part of #3071


piotrswierzy (approved):

**Verdict: ✅ Approve.** These tests assert behaviour, and I checked that they would actually go red rather than taking the count on trust.

### They can fail, which is the thing that matters

`docs/testing-guide.md` § Port-contract suites names the defect class the repo cares most about — "a contract suite is exactly the machinery that can look thorough and assert nothing". These do not fall into it:

- **The readiness gate is tested as a conjunction, including its non-obvious arm.** `should render a not-ready banner when the provenance backfill is still running, even with zero duplicate groups` (:181) is the test that earns the file: it pins that readiness needs *both* conditions, so inverting the `&&` to `||` — the plausible mutation — goes red here and nowhere else. The mirror case (duplicates present, backfill done) is covered at :95, and the both-clear case at :79.
- **Loading, report-error and provenance-error are three distinct tests** (:34, :47, :63), not one "renders without crashing". The second provenance-specific error case matters because the page reads two endpoints and a naive implementation surfaces only the first failure.
- **The id-fallback is asserted positively** (:147 → `ol_product_a1`), so dropping `?? productId` fails rather than silently rendering an empty cell.
- **Expansion asserts the revealed content**, not that a class toggled: both row ids plus the `Stale` / `Live` badges (:250-253). A broken `renderDetail` fails.
- Assertions are on operator-visible text throughout, so they survive refactors of the component internals — testing behaviour, not implementation, per `docs/testing-guide.md` § Best Practices.

### Findings

**[SUGGESTION]** — `apps/web/src/pages/inventory/duplicate-positions-page.test.tsx`
> Two gaps worth a line each, neither blocking:
>
> 1. **No test asserts the `groupCount` KPI value.** The readiness *banner* is well covered, but `groupCount` is the number `docs/operations/inventory-duplicate-positions.md` § Wave 1d gate tells an operator to read and nothing else ("must print 0"), and it is the one figure that is uncapped when the report is truncated. A test that renders a truncated report and asserts the KPI shows the full `groupCount` rather than `groups.length` would pin the exact confusion the truncation banner exists to prevent.
> 2. **`RequireAdmin` is untested.** #3261 introduces it with three branches — hydrating (`null`), non-admin (`ErrorState`), admin (children) — and the hydrating branch is the subtle one. A component test there would be cheap and would pin the "not known yet is not denied" rule against a future simplification to `if (!isAdmin) return <ErrorState/>`.

**[Note]** — as with #3261 and #3264, this PR's diff also carries the `EXPECTED_LAZY_ROUTE_COUNT` 66 → 67 bump and the `order-fulfillment-tasks-panel.test.tsx` selector fix. Identical blobs in all three, so they are the same change surfacing at each level of the stack rather than three edits; merge bottom-up and they collapse.

### Merge readiness

✅ **Approve.** All eight checks green. Stacked on `3076-duplicate-positions-nav-entry`.

piotrswierzy (approved):
# Tech Lead review — #3262 `test(inventory): duplicate-positions page tests`

Reviewed against `docs/testing-guide.md`, `docs/frontend-architecture.md`, and the repository's own standard that **a check which cannot fail is a defect** — so each test below is judged on whether it would go red against a plausible regression, not on whether it passes. CI green on all 8 checks.

## Step 1 — Context

319 lines of colocated `*.test.tsx` for `DuplicatePositionsPage`, which shipped across #3072–#3075 with none. Nine cases: two loading/error paths, the two-condition readiness rule, the #3239 name/sku enrichment and its id fallback, one expand interaction, and the truncated banner.

## Step 2/3 — Findings

**[IMPORTANT]** — `duplicate-positions-page.test.tsx:280-296` (`should expand a group row…`)

The test clicks the toggle and then asserts the detail rows are present. It never asserts they were **absent beforehand**. Against the most plausible regression here — `DataTable`'s `expandable.renderDetail` rendering unconditionally, or a default-expanded state slipping in — the rows are on screen at t=0, the click is a no-op, and every assertion still passes. The test would report green on a page whose expand/collapse does nothing, which is precisely the failure mode it is named for.

```ts
const toggle = await screen.findByRole('button', { name: /expand rows for product/i });
expect(screen.queryByText('ol_inventory_row1')).not.toBeInTheDocument(); // add this
await userEvent.click(toggle);
expect(screen.getByText('ol_inventory_row1')).toBeInTheDocument();
```

Worth a second assertion that a second click collapses it — collapse is currently untested in either direction.

**[IMPORTANT]** — no case for the page's admin-denied branch

`duplicate-positions-page.tsx` carries `if (isSessionReady && !isAdmin) { … 'Admin role required' … }`. Every test here passes `createAuthenticatedSessionAdapter()` (admin), so that branch — the page's own authorization affordance, on a `@Roles('admin')` diagnostic that reads `inventory_items` directly — is never exercised. `#3261`, stacked directly beneath this PR, adds a *second* admin gate, and this suite tests neither. A non-admin case asserting (a) the denial renders and (b) `getDuplicatePositions` was never called is the single highest-value missing test in the file, and it is the one that would have surfaced the duplicate-gate problem raised on #3261.

**[SUGGESTION]** — `:161-165` (`should render a ready banner…`)

Asserts the four positive strings but never that `Not ready` is absent. A regression that renders both banners (e.g. an `isReady` narrowing bug producing two `Alert`s) passes. One `expect(screen.queryByText('Not ready')).not.toBeInTheDocument()` closes it, and the symmetric negative in the not-ready test closes the other direction.

**[SUGGESTION]** — `:315` (truncated banner)

Asserts only the banner's title string. The part an operator acts on is the pair of numbers (`{groups.length}` of `{groupCount}`) — the claim about how much of their data is withheld. Asserting the title alone leaves a regression that prints `5 of 5`, or drops the counts entirely, green.

**[SUGGESTION]** — `Retry` is untested

The error state renders `action={<Button onClick={retry}>Retry</Button>}`, where `retry` refetches **both** queries. Neither the presence of the button nor the double refetch is asserted; a regression that refetches only `duplicatesQuery` leaves the provenance read permanently stale with no signal.

**[SUGGESTION]** — `:106-116`

`buildReport` / `buildProvenanceStatus` are good factories, but every non-trivial case then inlines a full 25-line `groups[]` literal. A `buildGroup(overrides)` factory would cut ~120 lines and make the *differences between* cases — which is what a reader needs — visible instead of buried.

### Positive observations

These are not padding — they are the cases a happy-path-only suite would have skipped:

- **`:88-105`** tests the provenance read failing **independently** of the duplicates read. Two reads, two failure paths, tested separately rather than as one "it errors" case.
- **`:246-268`** — *"zero duplicate groups but the backfill is still running"* is exactly the trap the page's docblock and #3240 exist for (a scan taken mid-backfill can read `groupCount: 0` and later flip). A naive `isReady = groupCount === 0` implementation passes every other test in this file and fails this one. That is a real regression guard.
- **`:214-244`** asserts the raw-id fallback rather than a fabricated name — matching the "never a fabricated name" rule in the page's `ProductCell` docblock.
- Both error tests assert the surfaced `error.message` (`'Network error'`, `'Forbidden'`), not just that an error state appeared.

## Step 4 — Final Assessment

### Summary
A genuine, well-targeted suite — the readiness-rule and independent-error-path cases are the ones that matter and they are here. Two gaps keep it from being complete: the expand test asserts only the post-state and so cannot fail against an always-expanded regression, and the page's admin-denied branch is entirely unexercised on a diagnostic whose whole access story is "admin only".

### Merge Readiness
🔄 **Approve with changes** — no BLOCKING issues. Both IMPORTANT items are additive test cases; land them before this reaches `main` behind the stack.

### Priority fixes
1. Add the pre-click negative assertion (and a collapse assertion) to the expand test — as written it cannot fail against the regression it exists to catch. (IMPORTANT)
2. Add a non-admin case asserting the denial renders **and** that `getDuplicatePositions` was never called. (IMPORTANT)
3. Add symmetric negatives to the ready / not-ready banner tests. (SUGGESTION)
4. Assert the truncated banner's counts, not only its title. (SUGGESTION)
5. Cover `Retry` refetching both queries. (SUGGESTION)

piotrswierzy (approved):
## Re-review (post force-push) — `60a86fa`

The prior IMPORTANT finding is genuinely fixed: the expand test now has an assertion that goes red against the regression it was supposed to catch.

---

### 1. Prior findings

|---|---|---|---|
| 1 | IMPORTANT — the expand test asserted only the post-click state, so it **could not fail** against an always-expanded (or default-expanded) regression | **Addressed** | `duplicate-positions-page.test.tsx:281-283` asserts both row ids are absent *before* the click, and `:293-296` re-queries the flipped `collapse rows for product` toggle and asserts both are absent again after it. That is three states, not one: closed → open → closed. An always-expanded implementation now fails at line 282, and a toggle that expands but never collapses fails at 295. |

Re-confirmed as genuine guards (they survive the rewrite unchanged):

- **The readiness rule** — `:239-254`. Zero duplicate groups **and** a still-draining backfill renders `Not ready` while still reporting the other condition as done. This is the one case where an implementation that ANDs the two conditions wrongly, or reads a single `undefined` as "not ready", goes red. It is also the standing "an empty read must never render as a positive all-clear" rule, tested.
- **The two independent error paths** — `:78-99` and `:101-122`. Each read is failed separately and the page still surfaces the error, so a regression that only ever checks `duplicatesQuery.error` is caught by the second. `:124-149` then proves `Retry` refetches **both**, not just the failing one.
- **The `getByText`/`queryByText` singular matchers are sound here** — `DataTable` renders card view only when `cardView && isMobile` (`data-table.tsx:255-256`), and `useMediaQuery` resolves false under the test DOM, so the detail rows exist exactly once. No hidden duplicate-match fragility.

---

**[SUGGESTION]** — `duplicate-positions-page.test.tsx:367-395`
> The new non-admin test is the right shape — it asserts the denial copy *and* that neither read was ever called, so it covers both halves (`enabled: isAdmin` on the hooks and the page's own branch). Worth noting it is now the only place in the suite that proves the 403 round-trip is never issued; if the page's early return and the hooks' `enabled` ever diverge, this is the test that catches it. Keep it.

**[SUGGESTION]** — `duplicate-positions-page.test.tsx:346`
> `expect(screen.getByText(/largest 1 of 5 duplicate groups/))` pins copy that #3264 (stacked directly on this branch) rewrites, because the word "largest" was the false-claim half of that PR's BLOCKING finding. Expected and already handled there — flagging only so the conflict is not mistaken for a bad rebase when the stack lands.

No new IMPORTANT or BLOCKING findings in this diff.

---

`Test (web)` is **FAILURE** on this PR (run `35337322775`, job `105575138333`): `2 failed | 557 passed`.

```
  ❯ useDuplicatePositionsQuery src/features/inventory/hooks/use-duplicate-positions-query.ts:31
```

**This is not caused by this PR** — it touches one file, and it is not one of those two. The `enabled: isAdmin` gate was added to both hooks on the `3075-duplicate-positions-truncated-banner` base (verified at that ref), while both hook test files still build a wrapper from `ApiClientProvider` + `QueryClientProvider` only, with no `SessionProvider`. #3264 fails on exactly the same 6 tests, which confirms the inheritance.

**The fix belongs on #3256 / the `3075-…` base**, not here: use `renderWithProviders`'s session-aware wrapper, or add `SessionProvider` to the local `createWrapper`. Nothing in this PR can resolve it, and nothing in this PR should try to.

---

### 4. Summary

Adds the page's test suite. The prior finding — an expand test that could not fail — is fixed properly, with the before-state and the collapse both asserted rather than a comment promising they were considered. The readiness-rule case and the two independent error paths are still the tests that carry the most weight here, and they are unchanged. The suite is red only because of a `SessionProvider` gap inherited from the base branch.

### Merge Readiness

✅ **Approve** — conditional on the stack-level gate above. Nothing in this diff needs to change.

### Priority fixes

1. (Base, not here) Give `use-duplicate-positions-query.test.tsx` and `use-provenance-backfill-status-query.test.tsx` a `SessionProvider` so `Test (web)` goes green for the whole stack.

piotrswierzy (approved):
## Tech Lead review — `test(inventory): duplicate-positions page tests`

Twelve cases covering most of the page's render arms, and three of them are written to fail against the regression rather than merely to pass against the current code — which is the distinction that matters.

### ✅ Three tests that can't pass vacuously

1. **Expand/collapse asserts absence first.**
   ```ts
   const toggle = await screen.findByRole('button', { name: /expand rows for product/i });
   expect(screen.queryByText('ol_inventory_row1')).not.toBeInTheDocument();
   ```
   Without that, a regression that defaults to expanded — or always renders the detail — passes every assertion after the click unchallenged. The comment says exactly that. It also asserts the collapse, so the toggle is tested as a toggle rather than as a one-way reveal.

2. **Retry asserts *both* queries refetch.** A regression that only retries the failing read would leave the provenance status permanently stale with no signal, and the page would keep rendering a readiness verdict off a half-refreshed pair. Counting call counts on both mocks is the right assertion.

3. **The non-admin case asserts neither read was called**, not just that the denial rendered. That's what pins #3252's `enabled: isAdmin` rather than only the page's early return — the two are separate mechanisms and only one of them prevents the 403.

Also good: the truncated test asserts `/largest 1 of 5 duplicate groups/`, i.e. the numbers an operator acts on, rather than the banner title alone.

### ✅ The two readiness conditions are tested independently

`groupCount: 0` + `completed: false` renders **Not ready** while still reporting "No duplicate position groups" as done. That's the whole point of #3240 — a scan that reads clean while the backfill is mid-run must not read as ready — and it's covered from the direction that actually matters.

### 🟡 IMPORTANT — the "Stuck" branch is untested

`latchedAt` is the field #3250 and #3252 both went out of their way to add, and it drives a distinct render arm:

```tsx
tone={backfillComplete ? 'success' : provenanceLatchedAndStuck ? 'error' : 'warning'}
{backfillComplete ? 'Done' : provenanceLatchedAndStuck ? 'Stuck' : 'Pending'}
```

Every provenance fixture here has `latchedAt: null`, so only `Done` and `Pending` are exercised. The `Stuck` badge, its error tone, the *"latched on … and has stopped running. It will not resume on its own — ask an engineer to re-arm it"* copy, and the KPI card's alternate `description` are all uncovered.

That's the one state where being wrong has a real cost: it's the difference between "wait, it's working" and "nothing is draining this, intervene". One fixture away:

```ts
buildProvenanceStatus({ remainingNull: 3, completed: false, latchedAt: '2026-08-01T00:00:00.000Z' })
```
…asserting `Stuck` and the re-arm sentence. Please add it before merge — this is the branch I'd most want pinned in the whole file.

### 🟢 SUGGESTION — the refresh-error arm is also uncovered

#3252 deliberately splits `initialLoadError` from `refreshError` so a failed **Refresh** keeps the last good report on screen with an inline *"Refresh failed … showing the last successful scan below"*. Both error tests here fail the first load, so only the full-page `ErrorState` is exercised.

Covering it means resolving once and rejecting on the second call, then clicking Refresh — a bit more setup, but it's a whole arm that exists precisely to avoid blanking a loaded report, and nothing currently stops someone collapsing the two error paths back into one.

### Notes

- Builders with `Partial<>` overrides for all three shapes keep each test to the fields it's actually about. Type-checked, so a new required field on any of them breaks here at compile time rather than at runtime.
- `should not show the truncated-report banner when the response is not capped` — the negative case beside the positive one. Worth having; that flag is easy to invert.
- CI: four green, five still queued at review time.

**Verdict:** ✅ Approve, with a request to add the `Stuck` case before merge.

PR #3264, opened 2026-09-15
feat(inventory): close mockup-parity gaps on the duplicate-positions page

jakubretajczykBD:
## Summary

The reviewed mockup ([artifact](https://claude.ai/code/artifact/9e6e1b2b-275a-4941-b25e-45ab18081a1a))  and https://github.com/openlinker-project/openlinker/issues/3114 modeled several elements #3072-#3077 shipped without, since their own issue text never enumerated them. Full findings write-up: see the comment on #3077.

- **Export CSV** — client-side flatten of `groups[]` to one row per `inventory_items` row, for offline triage or a support ticket.
- **Header Refresh** (always available, not just on error) + the `generatedAt` timestamp + a "re-run before acting on it" caution — none of this was rendered before, despite `generatedAt` already being on the wire.
- **"Live qty at risk" column** + default group ranking by live-stock exposure (sum of `availableQuantity` across non-stale rows), not row count.
- **Per-row survivor / reservation-risk badges** in the expanded detail panel, plus explanatory alerts (ATP-distortion vs. no-live-rows-no-risk).
- **Remediation modal** — 5-step guide, a generated ready-to-review `DELETE` for a group's loser rows (never executed by the page), Copy button, link to the full runbook.
- **`maxGroups` input + Apply** in the truncated state.
- Not-ready banner now states provenance is part of the position key; `null` vs. the `'legacy'` sentinel render distinct tooltipped labels.
- **Bug fix**: `text-muted`/`text-primary` aren't globally defined utility classes in this codebase — the #3074 null-dash fallback was silently unstyled. Replaced with a properly-scoped class.
- **Mobile fix**: the detail table's Row ID + badges wrapped to 3 lines and crushed together on a 390px card. Verified visually via chrome-devtools; wrapped in its own `overflow-x: auto` container.
- Dropped remaining raw issue numbers from operator-facing tooltips/copy, following #3073's review-fix convention.

## Context
Follow-up to epic #3071, stacked on #3262 (the last of the six numbered sub-issue PRs). **No dedicated issue number** — this closes gaps found by comparing the shipped page against the reviewed mockup after the fact, not a pre-planned sub-issue. Findings and full context posted as a comment on #3077.

⚠️ **Stacked PR** — base branch is `3077-duplicate-positions-tests` (#3262), not `main`. Merge #3262 first.

## Test plan
- [ ] `pnpm test` — 17/17 passing locally (was 9)
- [ ] `pnpm type-check`
- [ ] `pnpm lint`
- [ ] `pnpm check:invariants` (check-ui-vocabulary, check-css-structure, check-design-tokens all clean)
- [ ] Manual: verified end-to-end against seeded duplicate-position test data + real backend (#3249/#3250), including a 390×844 mobile viewport check via chrome-devtools

01-page-overview
02-full-page
03-expanded-row
04-expanded-detail
05-remediation-modal
06-remediation-group-specific

Part of #3071


jakubretajczykBD:
## /tech-review — PR #3264

### Summary
Solid follow-up: pure helper functions are well isolated and unit-testable, the CSV/remediation logic is correctly kept read-only ("nothing here writes anything"), and test coverage for the new surfaces is thorough (7 new tests). The one thing worth fixing before merge is the accessible-copy-feedback pattern, which diverges from an existing primitive/convention in this codebase for no stated reason.

### Issues

**[IMPORTANT]** — `duplicate-positions-remediation-modal.tsx` (`handleCopy`, the `✓ Copied` `<span>`)
> The copy-to-clipboard feedback is a plain sibling `<span>` with static text and no `aria-live` region, and the "Copy" button's own accessible name never changes. This diverges from the codebase's established pattern for exactly this affordance — `shared/ui/copyable-id.tsx` toggles the *button's own* `aria-label` between `Copy {x}` / `Copied {x}` so the state change is announced to assistive tech at the point of interaction, and the same pattern is repeated in `entity-label.tsx`, `raw-payload-panel.tsx`, and `structured-error-list.tsx`. A screen-reader user clicking "Copy" here gets no accessible confirmation that anything happened. Toggle the button's `aria-label`/visible label the way the other four copy affordances in the repo do (a full `CopyableId` swap isn't a fit for a multi-line SQL blob, but the aria-label toggle is copyable as-is).

**[IMPORTANT]** — `duplicate-positions-page.tsx` (survivor/reserved badges, `✓ likely survivor` / `⛔ reserved`)
> The emoji is embedded directly in the badge's text content with no `aria-hidden`. The codebase's existing convention for a decorative status glyph + label (`features/oms/components/sourcing-rules-table.tsx:329`) wraps the icon in `<span aria-hidden="true">✓</span>` and puts only the label in the accessible text — so AT doesn't read "check mark likely survivor" / "no entry sign reserved" literally. Match that pattern here.

**[SUGGESTION]** — `duplicate-positions-csv.ts` (`csvCell`)
> `csvCell` escapes commas/quotes/newlines but doesn't guard against CSV formula injection — a cell value beginning with `=`, `+`, `-`, or `@` (e.g. a product name synced in from a master shop the OpenLinker operator doesn't fully control) is interpreted as a formula by Excel/Sheets on open. Since `productName`/`sku` ultimately originate from external shop data (not operator-authored), and this file is explicitly the "offline triage / support ticket" export path, it's worth prefixing such cells (e.g. with a leading `'` or tab) the way most CSV-export utilities in this space do. Low severity (opt-in admin action, executes only on the exporting operator's own machine), so not blocking, but it's a well-known class of issue (CSV/"formula" injection) and the fix is a one-line addition to `csvCell`.

**[SUGGESTION]** — `duplicate-positions-page.tsx` (`DEFAULT_MAX_GROUPS = 100`, `MAX_GROUPS_CEILING = 500`)
> These mirror `DEFAULT_DUPLICATE_POSITION_GROUPS` / `MAX_DUPLICATE_POSITION_GROUPS` in `libs/core/src/inventory/application/services/inventory-query.service.ts` (currently 100 / 500 — correct today), and the backend DTO even re-exports `INVENTORY_DUPLICATE_POSITIONS_MAX_GROUPS` specifically "so a client can... [avoid] hardcoding it" (`get-duplicate-positions-query.dto.ts`). Since `apps/web` can't import `@openlinker/core` (#591) and there's no OpenAPI-generated client here, a literal duplicate is the pragmatic choice — but per the `docs/architecture-overview.md` precedent for exactly this situation (`scripts/check-shipping-tax-split-mirror.mjs`, `check-parameter-restriction-mirror.mjs`, etc.), a raw duplicated numeric constant with no drift guard is how these silently go stale. Not blocking for a diagnostics-only page with server-side `@Max` enforcement as the real backstop, but worth a one-line comment pointing at the source of truth, or (if this pattern grows) a `check:invariants` mirror script.

**[SUGGESTION]** — `duplicate-positions-remediation-modal.tsx` (the inline IIFE around the survivor/DELETE block)
> `((): ReactElement => { ... })()` inside JSX to get a local `const` (`survivorNotePrefix`) is a slightly unusual pattern for this codebase — an early-return/extracted small component (e.g. `RemediationDeleteBlock`) would read more idiomatically and match how `GroupRowDetail` itself is already factored out as a component. Purely stylistic, non-blocking.

### Verdict
🔄 **Approve with changes** — no architecture/security-boundary/contract violations; the two IMPORTANT items are both about accessible feedback for the same class of affordance (copy confirmation, decorative icon+label) and are small, precedent-backed fixes.

piotrswierzy (commented):

**Verdict: ❌ Request changes** — on one blocking issue that is specific to *this* PR's premise, not on the work, which is good and in places better than the runbook it implements.

### The read-only invariant holds, and I verified it rather than assuming

This was the thing most at risk in a PR that adds a file called `duplicate-positions-remediation.ts` and a modal called `RemediationModal`. It is clean:

- No mutation hook, no `apiClient` call, no fetch anywhere in `duplicate-positions-remediation-modal.tsx` — the only import from the lib is `buildRemediationDeleteSql`, and the generated SQL is rendered for copy-and-paste. The header says so and the code matches.
- The generated `DELETE` is keyed on `"id" IN (...)` only, which is exactly what `docs/operations/inventory-duplicate-positions.md` step 4 requires, along with its warning that a position-column `DELETE` "cannot express 'all but one' and will empty the whole group".
- `docs/architecture-overview.md` § Inventory (#2319) — "detects, never repairs" — is respected.
- 180 lines of CSS, **zero raw hex**; every colour resolves through a token.
- `reservationRiskRows` is a genuine improvement *over* the runbook: it surfaces that deleting a non-survivor with `reservedQuantity > 0` orphans a live hold, which the runbook's survivor rule does not consider. Keeping it as its own warning rather than folding it into the survivor pick is the right call and the comment explains why.

### Findings

**[BLOCKING]** — the mockup this PR exists to match is not in the repository
> The PR body cites `https://claude.ai/code/artifact/9e6e1b2b-…` as "the reviewed mockup". There is no `docs/plans/mockups/duplicate-positions*.html`, and that artifact id appears nowhere in the tree.
>
> `docs/frontend-architecture.md` § UX Mockups is not advisory here:
>
> > A high-fidelity, interactive HTML mockup produced while designing a non-trivial feature **must be committed to the repo** under `docs/plans/mockups/{feature-name}.html`, **not left as an ephemeral Claude Artifact link**. The Artifact URL is for live iteration with the requester during design; the repo copy is the durable spec that implementation and E2E tasks reference for the life of the feature.
>
> and it further requires that implementation tickets "cite the specific mockup file path and `data-state` value(s) they implement".
>
> This lands hardest on *this* PR specifically, because its entire stated purpose is closing parity gaps against that mockup. As it stands the parity claim is unverifiable by any reviewer and unreproducible from a clean checkout, and the artifact can be edited or expire independently of the repo — which is the precise failure the rule was written to prevent. The same rule also explains why a PR like this must touch the mockup file or state why it is still accurate; this one does neither, and the doc says a reviewer treats that as incomplete.
>
> The convention is live in this exact area, not theoretical: `docs/plans/mockups/inventory-locations-3063.html` is the sibling inventory feature's committed mockup, alongside 28 others.
>
> **Fix:** commit the mockup as `docs/plans/mockups/duplicate-positions-3071.html`, give each distinct state a stable `data-state` value with the in-page switcher the convention requires, and cite the path + values in the body. Related: the shipped page carries **no `data-state` attributes at all**, so the state-switcher half of the convention has nothing to anchor to either.

**[IMPORTANT]** — `apps/web/src/features/inventory/lib/duplicate-positions-remediation.ts:29-33` vs `:64-67`
> The module header states its own reason for existing:
>
> > kept in one place so the table's "likely survivor" badge and the modal's generated SQL **can never disagree** about which row is which.
>
> They disagree for an all-stale group, which is a case the runbook explicitly contemplates:
>
> - `findSurvivorId` → `rows.find(r => !r.isStale)?.id ?? null` → **`null`**, so the table renders no survivor badge.
> - `buildRemediationDeleteSql` → `rows.find(r => !r.isStale) ?? group.rows[0] ?? null` → falls back to **`rows[0]`**, so the modal names a survivor and generates a `DELETE` for every other row.
>
> An operator sees a group with no survivor marked, opens the modal, and is handed SQL that silently elects one. The modal's choice is a defensible reading of runbook step 2 ("keep the newest and delete the rest") — the defect is the *divergence*, in the one module written to prevent it.
>
> Resolve in one direction: either `findSurvivorId` adopts the same `?? rows[0]` fallback and the badge renders "newest (all stale)", or `buildRemediationDeleteSql` returns `sql: null` for an all-stale group and the modal surfaces runbook step 2's actual choice — keep the newest, or delete the whole group if the product is gone at the master. The second is truer to the runbook, which treats all-stale as a decision point rather than a default.

**[SUGGESTION]** — `duplicate-positions-remediation.ts:70`
> Row ids are interpolated into the SQL string unescaped. They are server-supplied UUIDs and the string is displayed rather than executed, so this is not an injection issue — but the value is copied into a psql session by hand, and a defensive `/^[0-9a-f-]{36}$/` filter (dropping and counting anything that fails) would cost one line and make that explicit.

**[SUGGESTION]** — PR scope
> +1244/−103 across eight files covering CSV export, header refresh, a new column plus a ranking change, per-row badges, a modal, a `maxGroups` input, a class-name bug fix and a mobile fix. Each is justified in the body, but the class-name fix (`text-muted`/`text-primary` were never globally defined, so #3074's null-dash fallback rendered unstyled) is a real bug fix that would be easier to find later as its own commit.

### Merge readiness

❌ **Request changes**, on the mockup only. Commit it under `docs/plans/mockups/`, add the `data-state` values, and cite the path — then this is an approve; the implementation itself is careful work and the read-only boundary is respected throughout. The survivor divergence should go in the same pass since it is a handful of lines.

All eight checks green. Stacked on `3077-duplicate-positions-tests` (#3262) — merge the chain bottom-up.

piotrswierzy (changes requested):
# Tech Lead review — #3264 `feat(inventory): close mockup-parity gaps on the duplicate-positions page`

Reviewed against `docs/frontend-architecture.md`, `docs/frontend-ui-style-guide.md`, `docs/engineering-standards.md`, `docs/testing-guide.md`, `docs/operations/inventory-duplicate-positions.md`, and the `#2319` read-only/admin-only framing in `docs/architecture-overview.md` § Inventory. CI green on all 8 checks.

## Step 1 — Context

The largest PR in the stack (+1262/−101): CSV export, a header refresh + `generatedAt` + caution, a "Live qty at risk" column with exposure-based ranking, per-row survivor/reservation badges, a remediation modal with a generated `DELETE`, a `maxGroups` control, and a mobile overflow fix. Two new pure modules under `features/inventory/lib/`, one new page-scoped modal, ~180 lines of CSS, and 8 new test cases.

Layering is correct throughout: `pages → features → shared`, both new libs import only `../api/inventory.types`, and the modal has precedent in `pages/` (`pages/products/marketplace-picker-modal.tsx`). All new CSS uses tokens — no raw hex. Nothing on the page writes; the `it detects, it never repairs` invariant holds.

## Step 2/3 — Findings

**[BLOCKING]** — `apps/web/src/pages/inventory/duplicate-positions-page.tsx:499-501` — the truncated banner states something false about the operator's own data

```tsx
Showing the {report.groups.length} highest-exposure group(s), ranked by live stock
quantity currently at risk of overselling — not row count.
```

The server does not select by exposure. `libs/core/src/inventory/infrastructure/persistence/repositories/inventory.repository.ts:724`:

```sql
HAVING COUNT(*) > 1
 ORDER BY row_count DESC, "productId" ASC, …
 LIMIT $1
```

The cap is applied **by row count, server-side**. `rankedGroups` (`:365`) only re-sorts the page that already came back. So a group with 2 rows and 900 units of live exposure is excluded by the server in favour of a 9-row all-stale group, while this banner tells the operator the highest-exposure groups are the ones on screen — and the banner exists *specifically* to explain what is being withheld. The copy it replaced ("the largest `{N}` of `{M}` duplicate group(s)") was accurate; this is a regression to a false claim, on the one surface whose output drives a manual production `DELETE`.

This is the class the architecture doc treats as a defect rather than a rough edge (cf. #2075: an operator told `No categories match` about a category two levels down — "a false statement about their own catalogue rather than a missing feature").

Either fix the copy — say the server capped by row count and that the visible list is re-ranked by exposure — or move the ranking into the SQL. The copy fix is one line; do not ship the claim as written.

**[IMPORTANT]** — `duplicate-positions-page.tsx:377-382` vs `features/inventory/lib/duplicate-positions-csv.ts:8-14` — the export silently under-reports, against its own module's stated requirement

The CSV module's docblock says exactly what the caller owes:

> "When the response is `truncated`, the export is therefore also partial — **the caller is expected to state that explicitly** rather than let the export silently under-report."

`handleExportCsv` states nothing. The button reads `Export CSV` unconditionally, the filename carries no marker, and there is no warning. An operator exporting a truncated report "for offline triage or a support ticket" (the module's own use case) gets a partial dataset they have no way to detect. Minimum: change the label/confirm when `report.truncated`, or encode it in the filename.

**[IMPORTANT]** — `apps/web/src/pages/inventory/duplicate-positions-remediation-modal.tsx:62` — clipboard call diverges from the repo precedent and throws on a non-secure origin

```ts
void navigator.clipboard.writeText(deletion.sql).then(() => setCopied(true));
```

`apps/web/src/shared/ui/copyable-id.tsx:55` uses optional chaining for a reason:

```ts
void navigator.clipboard?.writeText(id).then(() => { … });
```

`navigator.clipboard` is `undefined` outside a secure context. OpenLinker is a self-hosted product routinely run over plain `http://` on a LAN, where this throws a `TypeError` synchronously inside the click handler and propagates to the nearest error boundary. One character (`?.`) restores the shipped precedent; a `.catch` would additionally stop the unhandled rejection when the write is denied by permissions policy.

**[IMPORTANT]** — `features/inventory/lib/duplicate-positions-csv.ts:36,42-48` — the formula guard corrupts numeric cells

`FORMULA_PREFIX_PATTERN = /^[=+\-@\t\r]/` is applied to **every** cell, including numbers:

```ts
function csvCell(value: string | number | boolean | null): string {
  let raw = value === null ? '' : String(value);
  if (FORMULA_PREFIX_PATTERN.test(raw)) raw = `'${raw}`;
```

A negative `availableQuantity` or `reservedQuantity` exports as `'-3` — a text cell in every spreadsheet, in the two columns an operator is exporting the file to reconcile. The stated threat model is external shop catalogue strings (`productName`/`sku`/`connectionName`/`locationName`), so gate on `typeof value === 'string'` and leave numbers alone.

**[IMPORTANT]** — no unit tests for either new pure module

`duplicate-positions-csv.ts` (109 lines: quoting, escaping, formula neutralisation, group→row flattening) and `duplicate-positions-remediation.ts` (75 lines: exposure sum, survivor rule, reservation risk, SQL generation) are pure, non-trivial, and have no colocated test — `docs/engineering-standards.md` § Tests requires one for non-trivial logic.

The page test that nominally covers the export (`duplicate-positions-page.test.tsx`, *"should trigger a CSV download"*) asserts only `createObjectURL` called once and `click` called once. **It never inspects the CSV.** The builder could emit an empty string, a header-only file, or `'-3` in every numeric column and that test stays green — the "check that cannot fail" class. That absence is precisely why the two defects above are invisible today. A `duplicate-positions-csv.test.ts` asserting one row per `inventory_items` row, the quoting rule, a `=`-prefixed product name, and a negative quantity would cost ~40 lines and catch all of it. Same for `buildRemediationDeleteSql`'s all-stale fallback, which no test reaches.

**[IMPORTANT]** — `duplicate-positions-page.tsx:334-335, 384-390` — `maxGroups` is component state, not URL state

`docs/frontend-architecture.md` § URL State (line 283) lists what belongs in the URL: *filters, sort order, **pagination**, selected tab, view mode*. A page-size cap is pagination. As written, the operator's chosen cap is lost on reload — on a page whose own new caution text (`:414-417`) tells them to *"re-run the report immediately before acting on it, never from a page you loaded a while ago"*. Reloading is the documented workflow and it silently resets the cap to 100. Move it to a search param.

**[IMPORTANT]** — `features/inventory/lib/duplicate-positions-remediation.ts:65,25-33` — the survivor rule is implemented twice

The module docblock claims these helpers live together

> "so the table's 'likely survivor' badge and the modal's generated SQL can **never disagree** about which row is which."

but `buildRemediationDeleteSql` re-implements the pick inline rather than composing `findSurvivorId`:

```ts
const survivor = group.rows.find((row) => !row.isStale) ?? group.rows[0] ?? null;
```

The all-stale divergence is *intentional* and correct — runbook step 2 (`docs/operations/inventory-duplicate-positions.md:171-173`) says keep the newest when every row is stale, and both surfaces explain it. The problem is that the rule now exists in two places, so an edit to `findSurvivorId` will not reach the generated `DELETE`, which is the output an operator pastes into a production database. Express the difference by composing: `findSurvivorId(group) ?? group.rows[0] ?? null`, and correct the "can never disagree" claim.

**[IMPORTANT]** — `apps/web/src/index.css:4482-4494` — bespoke badges below the documented type scale, bypassing the primitive

`.duplicate-positions-survivor-badge` / `.duplicate-positions-reserved-badge` set `font-size: 0.5625rem` (9px). The smallest step in `docs/frontend-ui-style-guide.md` § typography (line 448) is `0.6875rem` / 11px. They are also hand-rolled while `StatusBadge tone=… compact` — which is what this styling approximates — is used two lines away in the same table cell for Live/Stale. Use the primitive, or at minimum bring the size onto the scale.

**[IMPORTANT]** — `duplicate-positions-page.tsx:110-119, 131-150, 195` — explanatory copy carried only by `title=`

Three of the new affordances put a full multi-sentence explanation in a bare `title` attribute with `cursor: help` (`.duplicate-positions-provenance-flag`): the ADR-058 "unlocated" flag, the `legacy`/not-backfilled distinction, and the "Live qty at risk" definition. `title` is not reachable by keyboard and does not open on touch, and mobile is in scope for this work. `docs/frontend-ui-style-guide.md` ships a `Tooltip` primitive over `@radix-ui/react-tooltip` (lines 543, 613) for exactly this. The existing `title={id}` precedent on `EntityLabel` is a *redundant echo of a visible id*, not the sole carrier of an explanation — different case.

**[SUGGESTION]** — `duplicate-positions-page.tsx:504` — `maxGroups` as operator-facing copy

The visible `<label>` is the raw API query-parameter name. `scripts/check-ui-vocabulary.mjs` deliberately scans only three Wave-2 feature folders (its own docblock, "WHAT THIS GATE DOES NOT CATCH" item 1), so `pages/inventory` is unscanned and nothing will flag it. "Groups to show" reads the same to a machine and better to a human — and this PR itself cites #3073's convention of keeping internal vocabulary out of operator copy. The hardcoded DOM `id="maxGroupsInput"` also breaks if the alert ever renders twice.

**[SUGGESTION]** — no mockup committed

The PR's parity target is a `claude.ai/code/artifact/…` link, so a reviewer cannot verify parity and a future reader cannot reproduce the comparison. Repo precedent commits it: `docs/plans/mockups/sourcing-rules-3055.html`, `inventory-locations-3063.html`, `sales-document-routing.html`.

**[SUGGESTION]** — `duplicate-positions-remediation.ts:71` — ids interpolated into generated SQL

`losers.map((row) => `'${row.id}'`)` does no escaping. Safe today because ids are minted by `formatInternalId`, and nothing here executes the string — but this is a statement handed to an operator to run against production, and the repo's security baseline is explicit about never interpolating into raw SQL. A shape assertion on the id, or a comment recording why it is safe, closes it.

**[SUGGESTION]** — `duplicate-positions-page.tsx:378-380` — CSV filename uses `new Date()`

The file describes the report's `generatedAt`, not the moment the operator clicked. Two exports of the same loaded report on either side of midnight get different names; an export of a stale page gets today's date. Use `report.generatedAt`.

**[SUGGESTION]** — `duplicate-positions-page.test.tsx` (CSV download case) — global stubs outside `afterEach`

`vi.stubGlobal('URL', …)` and `vi.spyOn(HTMLAnchorElement.prototype, 'click')` are undone by trailing statements inside the `it`. If an assertion throws first, both leak into the remaining tests in the file. Move to `afterEach`. Separately, `{ ...URL }` copies own enumerable properties only, so `URL.canParse` / `URL.createObjectURL` statics are dropped for the duration — harmless here, surprising later.

**[SUGGESTION]** — `duplicate-positions-remediation-modal.tsx:171` — hardcoded `github.com/openlinker-project/openlinker` link

On a self-hosted install (possibly a fork, possibly air-gapped) the "Open full runbook" button points at someone else's `main`. A shared docs-base constant would let a deployment redirect it.

### Positive observations

- **CSV formula injection** (`duplicate-positions-csv.ts:26-36`) is handled, with the threat model written down and the trust boundary correctly identified as the external shop catalogue. That is rarely thought about in an export feature.
- `duplicate-positions-page.test.tsx` *"Live qty at risk … 23, never 123"* is a real regression guard — it fails against the obvious wrong implementation (summing stale rows).
- The generated-`DELETE` test asserts `not.toContain('ol_inventory_row_survivor')`. A negative assertion on the row that must **not** be deleted is the right one for this feature.
- The `maxGroups` test asserts `toHaveBeenLastCalledWith(250)` rather than that a refetch merely happened.
- `DELETE` keyed on primary key only, with the reason recorded (`duplicate-positions-remediation.ts:57-62`) and matching runbook line 194 — the right call, explained.
- The `revokeObjectURL` deferral past the current tick, and the reason for it, is correct and non-obvious.
- All new CSS uses design tokens; the `overflow-x: auto` mobile fix follows the style guide's own table rule.

## Step 4 — Final Assessment

### Summary
A large, unusually well-documented parity pass that gets the layering, the tokens, the read-only invariant and several non-obvious details (formula injection, primary-key-only `DELETE`, object-URL revocation) right. It is held back by one copy line that makes a false statement about which duplicate groups were withheld — the server caps by row count, the banner claims exposure — and by a cluster of stated-contract gaps that share a root cause: the two new pure modules ship without unit tests, and the one test that touches the CSV asserts only that a download fired, so the builder's behaviour is entirely unverified. Add those tests and most of the IMPORTANT list becomes visible to CI rather than to review.

### Merge Readiness
❌ **Request changes**

### Priority fixes
1. **Fix the truncated-banner claim** — the server caps by `row_count DESC` (`inventory.repository.ts:724`); the page re-ranks only what it received. Restore an accurate sentence or move the ranking into SQL. (BLOCKING)
2. State that the CSV export is partial when `report.truncated` — the csv module's own docblock requires it of the caller. (IMPORTANT)
3. `navigator.clipboard?.writeText` — match `copyable-id.tsx:55`; as written the Copy button crashes on an `http://` deployment. (IMPORTANT)
4. Restrict the CSV formula prefix guard to string cells so negative quantities are not exported as text. (IMPORTANT)
5. Add `duplicate-positions-csv.test.ts` and `duplicate-positions-remediation.test.ts`; assert CSV **content**, not just that a download fired. (IMPORTANT)
6. Move `maxGroups` to a search param per `frontend-architecture.md` § URL State. (IMPORTANT)
7. Compose `findSurvivorId` inside `buildRemediationDeleteSql`, and correct the "can never disagree" docblock. (IMPORTANT)
8. Bring the new badges onto the type scale or use `StatusBadge compact`; replace the `title`-only explanations with the `Tooltip` primitive. (IMPORTANT)

piotrswierzy (changes requested):
## Re-review (post force-push) — `1ce75b5`

The BLOCKING finding is genuinely fixed and fixed correctly. Six of the seven IMPORTANTs are addressed. One is not, and the new code introduces a false claim of the same class as the one just removed — so this is still `request-changes`, but on a much shorter list.

---

### 1. Prior findings

|---|---|---|---|
| 1 | **BLOCKING** — the truncated banner claimed the shown groups were the **highest-exposure** ones; the server caps by `row_count DESC` | **Addressed** | `duplicate-positions-page.tsx:604-612` now reads "…the {n} of {N} duplicate groups with the most `inventory_items` rows — **row count, not exposure**. What's shown is then re-ranked by live stock quantity at risk, so a low-row-count group carrying more exposure could still exist below this cap and not appear here." Verified against the source of truth: `inventory.repository.ts:724` is `ORDER BY row_count DESC, "productId" ASC, …`. The copy is now true, states the re-rank explicitly, **and keeps both numbers** — so #3256's own property (both counts disclosed, "largest" a true claim) is preserved rather than traded away. `duplicate-positions-page.test.tsx:350-351` pins both halves. |
| 2 | IMPORTANT — the CSV test asserted only that `createObjectURL`/`click` fired and never inspected the CSV, hiding two live defects | **Addressed** (see new finding 2) | New `duplicate-positions-csv.test.ts` inspects actual content: header row, one line per `inventory_items` row, quoting (`:90-107`), formula-prefixing scoped to strings (`:109-133`), and the negative-quantity regression explicitly (`:135-160`). The source fix is right: `duplicate-positions-csv.ts:58` guards on `typeof value === 'string' && FORMULA_PREFIX_PATTERN.test(raw)`, so `-3` exports as `-3`. The truncation half is now signalled by a `-partial` filename suffix (`duplicate-positions-page.tsx:451`). |
| 3 | IMPORTANT — `navigator.clipboard.writeText` unguarded, throws on an `http://` self-hosted deployment | **Addressed** (untested — see new finding 2) | `duplicate-positions-remediation-modal.tsx:74-85` — `void navigator.clipboard?.writeText(…)` with a `.catch`, matching the `copyable-id.tsx:55` precedent. Optional chaining short-circuits the whole chain, so the `.then` is not reached when `clipboard` is `undefined`. Correct. |
| 4 | IMPORTANT — `maxGroups` held as component state, against `frontend-architecture.md` § URL State | **Addressed** | `duplicate-positions-page.tsx:381-394` uses `useSearchParams`; `frontend-architecture.md:289` lists pagination explicitly. The `useEffect` resync (`:396-398`) closes the follow-on gap where back/forward or a pasted `?maxGroups=` changed what was fetched without changing what the input showed, and `duplicate-positions-page.test.tsx:761-790` is a real guard for it — the page component does not unmount across that navigation, so without the effect the input stays at 100 and the test goes red. |
| 5 | IMPORTANT — new badges at `0.5625rem` (9px), below the smallest documented step | **Addressed** | `index.css` — `0.6875rem` with the reason in place. That is exactly the smallest documented step (`frontend-ui-style-guide.md:448`, "Eyebrow / uppercase / mono-caps `0.6875rem` / 11"). The `Live rows` column additionally moved onto the real primitive (`StatusBadge … compact`, `duplicate-positions-page.tsx:242-248`). Residual below. |
| 6 | IMPORTANT — multi-sentence explanations carried only by `title=` + `cursor: help` | **Addressed** | `LocationCell` (`:116-131`), `ConnectionCell` (`:141-172`) and the `Live qty at risk` header (`:209-220`) now use `Tooltip`/`TooltipTrigger asChild`/`TooltipContent` with `tabIndex={0}` on the trigger, so they are keyboard-reachable. `TooltipProvider` is mounted at the app root (`app-providers.tsx:72`) and in `test-utils.tsx:1266`, so no missing-provider hazard. Residual below. |
| 7 | IMPORTANT — `buildRemediationDeleteSql` re-implemented the survivor pick instead of composing `findSurvivorId`, in generated `DELETE` SQL | **Addressed** | `duplicate-positions-remediation.ts:72` — `const survivorId = findSurvivorId(group) ?? group.rows[0]?.id ?? null;`. One place spells the pick, one place spells the all-stale fallback, and the module docblock now records why. `duplicate-positions-remediation.test.ts:121-136` reaches the all-stale branch, which nothing previously did. |
| 8 | IMPORTANT — no mockup committed under `docs/plans/mockups/`, so parity is unverifiable | **Not addressed** | No such file on the head (`gh api .../contents/docs/plans/mockups?ref=3077-v2-…` returns nothing matching `dup*`). The PR body cites a `claude.ai/code/artifact/…` link instead. That is not repo-durable: a reviewer or maintainer without access to that artifact cannot check the parity claim the PR is entirely about, and the precedent in this same area commits the file — `docs/plans/mockups/sourcing-rules-3055.html`, `docs/plans/mockups/inventory-locations-3063.html`. Commit it as `docs/plans/mockups/duplicate-positions-3071.html`. |

---

**[BLOCKING]** — `apps/web/src/pages/inventory/duplicate-positions-page.tsx:277-286`
> **The distortion alert fires on `liveRowCount > 0`, so a group with exactly ONE live row is told it is currently oversell-ing — which is false, and it is very likely the most common duplicate shape in the field.**
>
> Availability excludes stale positions (`isStale = false` is applied by `findAvailabilityByVariantIds` / `getPromisableQuantities` — `architecture-overview.md` § Inventory, #1478/#2345). With one live row and N stale ones, the sum across live positions **is** the survivor's figure: there is no over-count, no ATP distortion, nothing to reconcile. Yet the page renders `Currently distorting available-to-promise… the availability read has no way to know only one row is real, so it publishes {exposure} where the true figure is whichever row survives` — a statement about the operator's own stock that is not true.
>
> This is the same class as the BLOCKING finding this PR just fixed: a confident false claim on the primary surface, driving the operator toward the wrong action (here, urgent deletion of rows that are not hurting anything). It is also reachable by default rather than at the margin — #2322's pooled-position repair and #3206's mirror both stale rows routinely, so "one live row plus several stale" is a normal steady state, not an edge case.
>
> The page already knows the distinction and draws it one step too low: the `else` arm (`:288-293`) says exactly the right thing — "No live rows — no oversell risk today… This group only blocks the stricter uniqueness migration; it is not an active stock-accuracy problem" — which is equally true at one live row. The boundary is `≥ 2 live rows`, not `≥ 1`. `liveExposureQuantity`'s own docblock (`duplicate-positions-remediation.ts:22`) bakes the same off-by-one in ("a group with 5 stale rows and 0 live ones is cosmetic cleanup"), so fix it there too.
>
> The `Live rows` badge has the same boundary (`:241`, `tone="warning"` for any `liveRowCount > 0`) and should move with it.
>
> No test covers `liveRowCount === 1` — every fixture that expands a group has two live rows (`duplicate-positions-page.test.tsx:493-552`, `:553-619`), which is why this is invisible. Add the one-live-row case; it should assert the neutral note, not the alert.

**[IMPORTANT]** — `apps/web/src/pages/inventory/duplicate-positions-page.tsx:449-452` and `duplicate-positions-remediation-modal.tsx:74-85`
> **Two of the three fixes for prior findings are themselves untested, so they can regress silently — which is the exact mechanism that let the original defects ship.**
>
> The negative-quantity guard is properly tested (`duplicate-positions-csv.test.ts:147-169`) — good. But:
> - The `-partial` suffix and the `generatedAt`-derived date part (`:450-452`) are asserted nowhere. The CSV download test (`duplicate-positions-page.test.tsx:439-491`) still only checks `createObjectURL` and `click` were called — the very shape flagged last round. Drop the suffix and every test still passes, and the export is back to silently under-reporting in contradiction of `duplicate-positions-csv.ts:10-15` ("the caller is expected to state that explicitly"). Read `this.download` inside the existing `clickSpy` mock implementation, and add a truncated-report case.
> - The clipboard `?.` guard is asserted nowhere. A test that opens the modal, clicks `Copy` and asserts no throw goes red without it (`navigator.clipboard` is undefined under the test DOM, so the un-guarded call is a synchronous `TypeError` inside the click handler). One test, and the `http://`-deployment regression cannot come back.

**[SUGGESTION]** — `duplicate-positions-page.tsx:263-271` vs `:277`
> The detail panel reads "is there live stock here?" from two different sources in adjacent lines — `findSurvivorId(group)` / `liveExposureQuantity(group)` walk `group.rows`, while the alert gate reads the server's `group.liveRowCount`. They agree today because the endpoint returns every row of a group uncapped, but that is an undocumented coupling. Derive one from the other (`deletion.survivor`, or a `liveRows` array computed once) so the panel cannot contradict itself if per-group row capping is ever added.

**[SUGGESTION]** — `index.css` (`.duplicate-positions-survivor-badge`, `.duplicate-positions-reserved-badge`)
> Size is now on-scale, but these are still hand-rolled beside `StatusBadge tone="success"/"error" compact`, which the same file adopted for `Live rows` two lines away (`duplicate-positions-page.tsx:242-248`). If the glyph is the reason, that is worth stating in the CSS comment; otherwise two more `StatusBadge`es delete ~30 lines of CSS and inherit future token changes for free.

**[SUGGESTION]** — `duplicate-positions-page.tsx:210-219` vs `:194-201` / `:232-239`
> Radix `Tooltip` deliberately does not open on tap, so the three converted hints are keyboard-reachable but still not touch-reachable, and mobile is in scope. Separately, the `Rows` and `Live rows` headers still carry the pre-existing `title=` hints, which now say "largest first" while the default render order is exposure-ranked (`:428-431`). Not introduced here; worth aligning while the copy is being corrected anyway.

**[SUGGESTION]** — `duplicate-positions-page.tsx:382-386`
> `?maxGroups=` (present but empty) parses as `Number('') === 0`, which is finite, so it clamps to `1` rather than falling back to `DEFAULT_MAX_GROUPS`. Same for clicking `Apply` with the input cleared. Treat empty as absent.

**[SUGGESTION]** — `duplicate-positions-csv.ts:59`
> The quoting predicate is `/[",\n]/` while the formula predicate also matches `\t` and `\r`. A cell containing a bare `\r` is therefore neither quoted nor escaped, and the output uses `\n` line endings with no BOM — Excel will mojibake a non-ASCII `productName`, which is the one column most likely to carry one. Adding `\r` to the quoting class and a UTF-8 BOM is a two-line change.

**[SUGGESTION]** — `duplicate-positions-page.tsx:74-75`
> `DEFAULT_MAX_GROUPS` / `MAX_GROUPS_CEILING` duplicate backend constants, documented honestly as a mirror. Low risk (the server's `@Max` is the real backstop and drift only affects an input hint), so no change requested — but the repo's convention for a load-bearing mirror is a `scripts/check-*-mirror.mjs` under `pnpm check:invariants` (#2610, #2240). Worth one if this ever gates anything.

**Positive observations.** The remediation SQL is keyed on primary key only with the reason recorded (`duplicate-positions-remediation.ts:60-72`) — the one mistake that would empty a whole group is explicitly foreclosed. The `afterEach` cleanup (`duplicate-positions-page.test.tsx:62-66`) fixes a real leak: a trailing `mockRestore()` after a failing assertion never runs and poisons every later test in the file. The CSV filename is derived from `report.generatedAt` rather than click time, which is the right call and the docblock says why. And the modal's copy affordance announces through the button's own `aria-label` rather than a sibling `<span>`, matching `copyable-id.tsx`.

---

`Test (web)` is **FAILURE** on this PR (run `35340620317`, job `105585503922`): `2 failed | 559 passed`.

```
```

Not caused by this PR — #3262 fails on the identical 6 tests, and the `enabled: isAdmin` gate was added to both hooks on the `3075-duplicate-positions-truncated-banner` base while both hook test files still wrap with `ApiClientProvider` + `QueryClientProvider` only. **The fix belongs on #3256 / the `3075-…` base.** Listed here so it is not mistaken for fallout from this change.

---

### 4. Summary

Closes the mockup-parity gaps: exposure ranking, a live-qty-at-risk column, the survivor/reservation badges, the remediation modal with a generated `DELETE`, a CSV export, and a URL-owned `maxGroups`. The BLOCKING copy defect is fixed properly — the banner now names the server's real cap axis, discloses the re-rank, and keeps both counts, so #3256's own guarantee survives intact. Six of seven IMPORTANTs are closed with the right shape, and the CSV and remediation helpers now have real unit tests including the all-stale branch nothing previously reached. What holds it back is that the new explanatory copy repeats the pattern that was just removed at a different boundary: a group with one live row is told it is actively over-publishing stock when it is not, and no test covers that case. The mockup is still not committed, so the parity claim this PR exists to satisfy cannot be checked from the repository.

### Merge Readiness

❌ **Request changes**

### Priority fixes

1. **BLOCKING** — move the distortion boundary from `liveRowCount > 0` to `≥ 2 live rows` (`duplicate-positions-page.tsx:277`, `:241`, and `liveExposureQuantity`'s docblock), and add a one-live-row test asserting the neutral note rather than the alert.
2. **IMPORTANT** — commit the mockup under `docs/plans/mockups/duplicate-positions-3071.html`, per the `sourcing-rules-3055.html` / `inventory-locations-3063.html` precedent.
3. **IMPORTANT** — test the `-partial`/`generatedAt` filename and the clipboard `?.` guard, so neither fix for a prior finding can regress silently.
4. Optional: the six SUGGESTIONs above, of which the two-sources-for-one-fact one (`:263` vs `:277`) is the one most likely to bite later.

piotrswierzy (approved):
## Tech Lead review — `feat(inventory): close mockup-parity gaps on the duplicate-positions page`

Largest slice in the stack, and the two lib modules are the careful part. One documentation contradiction to settle, since this PR hands an operator generated SQL.

### 🟡 IMPORTANT — the runbook contradicts this PR about the id format, and the runbook is wrong

`duplicate-positions-remediation.ts` justifies its string interpolation with:

> *`row.id` is never operator input — it's minted server-side by `formatInternalId` (`ol_<prefix>_<hex>`)*

`docs/operations/inventory-duplicate-positions.md` says the opposite, in a comment sitting directly above the `DELETE` an operator is meant to hand-write:

```sql
-- ids come from the report's groups[].rows[].id — bare UUIDs, NOT
-- ol_-prefixed internal ids (inventory_items.id is a plain uuid).
```

**The PR is right and the runbook is stale.** `inventory_items.id` was `uuid` in the original migration, and `1767555200000-change-product-ids-to-varchar.ts` converted it:

```sql
ALTER TABLE "inventory_items" ALTER COLUMN "id" TYPE text USING id::text
```

The ORM entity agrees (`@PrimaryColumn({ type: 'text' })`), and every int-spec in this stack inserts `ol_inventory_*` string ids against it.

That parenthetical is actively harmful now that this page generates the SQL: an operator who trusts the runbook will look at an `ol_`-prefixed id in the report or in the generated `DELETE`, conclude the page is producing the wrong thing, and either "correct" it into a UUID (matching nothing) or stop and raise a bug against a working feature. Please fix the runbook comment in this PR — it's two lines, and this is the PR that makes the discrepancy reachable.

### ✅ `hasOversellRisk` requiring **two or more** live rows

This is the right fix and the docblock explains why better than I could:

> *with exactly one live row and any number of stale ones the sum across live positions IS the survivor's own figure — there is no over-count and nothing to reconcile*

Availability reads already exclude stale positions (#1478/#2345), so `liveRowCount > 0` would tell an operator a perfectly healthy group was actively overselling. And `atRiskExposureQuantity` gating the same way matters for a second reason worth calling out: a raw sum would rank a healthy single-live-row group with a big quantity **above** a genuinely at-risk two-live-row group with a small one, in a column literally headed "Live qty at risk".

Deriving liveness from `group.rows` rather than the server's `liveRowCount` so the badge, the alert gate and the sum cannot disagree is the right call — one source, not two.

### ✅ `buildRemediationDeleteSql` composes `findSurvivorId`

So the table's "likely survivor" badge and the generated `DELETE` cannot pick different rows. That's the property that matters here: two independent implementations of the survivor rule, one shown and one executed, is how an operator ends up deleting the row the page told them to keep. The all-stale fallback is spelled once, in the same place, matching runbook step 2.

Primary-key-only `DELETE`, with the reason (*a position-keyed DELETE can't express "all but one" and would empty the group*) — same warning the runbook gives.

### ✅ CSV formula injection, scoped correctly

Prefixing `=`/`+`/`-`/`@`/tab/CR-leading cells with `'` — **for string cells only**:

> *`availableQuantity` / `reservedQuantity` are legitimately negative numbers, and applying the pattern to every cell exported `-3` as the text `'-3` in the two columns an operator is exporting the file to reconcile*

That's the whole point of the export ruined by a blanket guard, and the threat model is stated precisely: the untrusted strings are an external shop's catalogue names, not numbers this app computed. The separate note on why `\r` is quoted even though it can't trigger the prefix guard (it only matters when it *opens* a cell, versus embedded mid-value against `\n` line endings) is the kind of detail that gets deleted as redundant without it.

### ✅ The export is honest about being partial

> *there is no dedicated uncapped export endpoint, so this builds the CSV from whatever `groups[]` the CURRENT response already loaded … the caller is expected to state that explicitly rather than let the export silently under-report*

Right, and important: an operator exporting for a support ticket must not believe they attached everything. Worth double-checking the page actually renders that statement when `truncated` is true — the module docblock delegates it to the caller, so it's exactly the kind of obligation that goes unmet.

### 🟢 SUGGESTIONS

1. **Empty `losers` produces invalid SQL.** A group with exactly one row yields `WHERE "id" IN ();`. Unreachable by definition (a duplicate group has ≥2 rows) and the no-rows case is tested, but `losers.length === 0 → sql: null` is one line and removes the only way this function can emit something a database will reject.

2. **The generated `DELETE` has no transaction wrapper or verification step.** The runbook's step 5 is "re-run the report until `groupCount` reads 0", which covers it procedurally — but a snippet built for copy-paste into a DB client could reasonably ship as `BEGIN; … ` plus a matching `SELECT` to eyeball first. Worth considering; the modal may already say this (I reviewed the lib, not every line of the 202-line modal).

### Notes

- Mockup committed at `docs/plans/mockups/duplicate-positions-3071.html` per `docs/frontend-architecture.md` § UX Mockups, and this PR is explicitly the parity pass against it.
- Remediation tests name the review findings they pin (`#3264 review BLOCKING finding`), so a future reader can tell which cases are regression guards rather than illustrations.
- CI: six green, `Lint` in progress, `Build` / `Test (web)` queued at review time.

**Verdict:** ✅ Approve. Please correct the runbook's id-format comment before merge — it's the one thing here that could send an operator down the wrong path with a `DELETE` in hand.

PR #3267, opened 2026-09-15
feat(eparagony): make the invoicing adapter reachable on a receipts connection

norbert-kulus-blockydevs:
Stack member 2 of #3192, on top of #3260. **Targets the integration branch `3192-eparagony-invoicing`, not `main`.**

Part of #3173.

## What this slice does

Slice 1 built the invoicing adapter and nothing could reach it. This makes it reachable: the manifest advertises the capability, the dispatch table resolves it, the factory constructs it, and the config-shape validator knows the keys it needs.

- **`createFiscalizationAdapter` becomes `createAdapters`**, returning `{ fiscalization, invoicing }` built over ONE `EparagonyHttpClient` - the KSeF/PrestaShop bag shape. `IEparagonyAdapterFactory` changes with it; both are package-private, so no consumer breaks. A spec asserts the two adapters hold the same transport instance.
- **Manifest**: `['Fiscalization', 'FiscalRegistrationLocator', 'Invoicing', 'RegulatoryStatusReader']`.
- **Dispatch table** gains `Invoicing`. `dispatchCapability` looks up by capability string, so without the entry `getCapabilityAdapter(id, 'Invoicing')` throws.
- **Config-shape validator** learns `merchantTIN`, `merchantName`, `merchantAddress` and `eInvoicingHubEnabled`, **all optional**. Every existing connection is receipts-only and carries none; making one required would refuse an existing connection's own config on its next save.

## Judgement calls worth your eye

**`CorrectionIssuer` is deliberately NOT advertised**, unlike `RegulatoryStatusReader`. The adapter refuses a correction pre-call and #3193 is where corrections arrive; advertising the name now would promise a document the plugin cannot produce. That follows the repo's rule that a capability name enters the manifest together with the adapter that delivers it (the Erli #980 precedent). Specs pin both the four that are advertised and the absence of `CorrectionIssuer`, so #3193 has to flip the assertion rather than silently forget the manifest.

**Advertising `RegulatoryStatusReader` follows KSeF**, whose manifest already carries `OfflineResubmitter` and `RegulatoryRecordLocator` alongside `Invoicing`. inFakt and Subiekt advertise `Invoicing` alone despite implementing several sub-capabilities, so both patterns exist in the tree; this plugin already advertises `FiscalRegistrationLocator` the same way, so it stays internally consistent.

**`merchantAddress.country` is checked as two ASCII letters.** That enforces this package's own declared type (`ISO 3166-1 alpha-2`) rather than guessing the vendor's rule, and it is a shape check rather than a canonicalisation - the port returns `void` and cannot write back, so either case passes. It catches `Poland`.

## An overstatement in the brief, corrected in the code

The brief for this slice claimed sharing one client avoids double token traffic per connection. **That is only true within a single resolution.** `getCapabilityAdapter` builds a fresh adapter bag on every call, so two resolutions for one connection still fetch two tokens - which is the pre-existing behaviour, not a regression this introduces. The comment at the hoisted factory says so plainly rather than claiming a fix that was not made. A genuine per-connection client cache needs an invalidation story (rotated credentials, a changed host override) and is not written here.

## No migration, and that is the decision

Adding `Invoicing` to the manifest does **not** grant it to existing connections - `enabledCapabilities` is stamped at create and never retro-filled (#2085). The operator ticks the capability checkbox. Verified: `ConnectionCapabilitiesPanel` renders a checkbox per supported capability and PATCHes `enabledCapabilities`, `Invoicing` is already in the frontend's core-capability list with help text, and the only exclusivity pair is `InventoryMaster` <-> `OfferManager`. `RegulatoryStatusReader` and `FiscalRegistrationLocator` correctly render nothing there, being outside `CoreCapabilityValues`.

## Mockup

`docs/plans/mockups/sales-document-eparagony-invoicing.html` gets one correction, and it is slice 1's debt rather than this slice's: task E3-5 still specified `CONFIRMED+KSEF -> cleared`, which slice 1 ships as `accepted`. The issue body was corrected when that decision was made and this line was missed. Nothing else in the mockup is affected - it is a UI spec and this slice is plugin wiring.

## Verification

`pnpm type-check` clean across all 21 projects, `pnpm lint` zero errors with no `--fix` churn, 90 of 90 invariant scripts pass (`check-repo-urls` fails only on local untracked `.worktrees/` scratch, identically on a clean `main`). Tests are written and deferred to CI per the repo's local-gate policy.

Confirmed unaffected: `check-core-capability-mirror.mjs` guards `CoreCapabilityValues` and its three mirrors, none of which this slice touches.

## Review fix: `merchantAddress: null` crashed the live issue path

Closes #3274.

A cleared `merchantAddress` crashed the invoice path. The config-shape
validator blesses an explicit `null` as absent (the repo's own cleared-knob
convention, #2610, and what the seller form will write), but both mapper
guards tested `undefined` alone, so `toSellerEntityAddress(null)` dereferenced
it. The resulting `TypeError` carries no `failureMode`, so
`InvoiceService.classifyFailure` defaults to `in-doubt`, and a non-`rejected`
failure blocks that order from every sales document on every connection for a
request that never crossed the network.

Both guards move together: `toIssuedDocumentSeller` runs inside
`toIssueResult`, after `createDocument` and its poll have succeeded, so fixing
only the composer would relocate the crash to a point where the vendor has
already created a legally-issued document that OpenLinker then loses. Two
specs pin the mapper against the validator's own "null is absent" test.

Also fixed in the same pass: the create path now sends
`enabledCapabilities: ['Fiscalization']` explicitly rather than omitting it -
with `Invoicing` in the manifest, an omitted value would let the guided
wizard (which collects no seller invoicing configuration) mint a connection
that claims it can invoice; the seller country check now tests the raw value
rather than a trimmed copy, since `toSellerEntityAddress` ships the value
untouched.

---

## Merge order (Piotr's IMPORTANT finding on #3260's review)

This branch and `3192-1-invoice-adapter` (#3260) diverge at `2a65cc7c0` and, at the time of that review, neither was an ancestor of the other - a real conflict on the config-shape validator, since both slices touched it after their own reviews.

**Resolved by intent:**

1. **#3260 merges first** into `3192-eparagony-invoicing`.
2. **This PR rebases onto the updated integration branch second.** The validator-adding commit here (`0b1d75492`) resolves as **already applied** - #3260 adopted this slice's validator verbatim (see its `7ff544100` and the *Stack convergence* note in its review thread), so the two histories agree on that hunk once #3260 has landed. The manifest, dispatch-table and factory changes in this PR are net-new and are unaffected by the rebase.
3. **#3269 (slice 3, web surfaces)** is already based on this branch and merges third.

No code change is needed in this PR's diff for this finding - it is a coordination note.


norbert-kulus-blockydevs:
## Tech lead review

Reviewed against the **stack parent** (`git diff origin/3192-1-invoice-adapter...origin/3192-2-invoicing-wiring`), not the integration branch, so slice 1's 3374 lines are excluded. That leaves 8 files / 596 insertions, which is the real surface here.

The shape is right. The bag-returning factory follows `AllegroAdapterFactory.createAdapters` / the KSeF-PrestaShop precedent, the advertised-without-dispatch split is correct (`dispatchCapability` looks up by exact own-key, so `Invoicing` must be in the table and the two sub-capabilities must not be), the `CorrectionIssuer` omission is well argued and pinned twice so #3193 has to flip it, the rename is complete across the whole tree with no stale callsite, and `IEparagonyAdapterFactory` really is package-private (not exported from `src/index.ts`). The self-correction about what the hoisted factory does *not* buy is accurate; I checked `IntegrationsService.getCapabilityAdapter` and there is genuinely no memoisation.

Two things stop it merging.

---

### BLOCKING

**`merchantAddress: null` is blessed by the new validator and crashes the invoice path** - `eparagony-connection-config-shape-validator.adapter.ts:215`, with `eparagony-invoice.mapper.ts:341` and `:395`

`validateMerchantAddress` returns early on `raw === null`, and the new spec pins that reading by name (`eparagony-shape-validators.spec.ts:92`, *"should treat an explicit null as absent, the way an operator-authored blob carries one"*). So the config saves clean and persists.

The mapper does not agree. Both guards test `undefined` only:

- `mapper.ts:341` - `if (config.merchantAddress !== undefined)` is **true** for `null`, so `toSellerEntityAddress(null)` runs and its first statement is `readNonEmpty(address.apartment)`.
- `mapper.ts:395` - `address === undefined` in `toIssuedDocumentSeller`, same miss, falls through to `address.street`.

Both are on the live `issueInvoice` path and neither is wrapped, so it escapes as a raw `TypeError`. That is the part that makes this blocking rather than cosmetic:

1. `InvoiceService.classifyFailure` reads `(error).failureMode`, finds nothing on a plain `TypeError`, and returns `'in-doubt'` by its own fiscal-safe default.
2. `InvoiceRecord.blocksIssuanceElsewhere` returns `true` for `failed` + non-`rejected`.
3. Both cross-kind guards are connection-agnostic, so **that order can no longer receive an invoice or a fiscal receipt on any connection**, for a request that never crossed the network. The operator-facing copy is "the provider rejected the request", which is untrue and names no remedy.

Reachability is not theoretical. `readEparagonyConnectionConfig` is a blind `as unknown as` cast, so this validator is the only gate; the raw JSON config editor sits on the connection page and curl/MCP bypass the browser entirely. More to the point, the repo's own cleared-knob convention (#2610) is that clearing a knob writes an **explicit `null`** rather than deleting the key, and that every reader treats `null` exactly like absent - so the slice-3 form, written in house style, makes this the normal clear path.

**Fix the mapper, not the validator.** The validator's `null`-is-absent reading is the one that matches #2610 and the three sibling keys; it is the mapper that is out of step. Change `:341` to `!= null` **and** `:395` to `== null` in the same commit: fixing only `:341` moves the crash to `toIssuedDocumentSeller`, which runs inside `toIssueResult` **after** `createDocument` and the poll have succeeded (`eparagony-invoicing.adapter.ts:163` → `:172` → `:175` → `:518`), turning a pre-call crash into one that loses a document the vendor legally created. And `merchantTIN` is mandatory by then, so `taxId` is always non-null there and any config with a `merchantName` plus a null address hits it.

Then add a `composeInvoiceDocument` spec driving `merchantAddress: null`, so the validator's "null is absent" test has a counterpart proving the mapper agrees. A green test asserting a config that crashes the invoice path is worse than no test, and CI cannot catch it.

---

### IMPORTANT

**1. The "no migration, the operator ticks the checkbox" story is true for existing connections and false for every new one** - `eparagony-plugin.ts:75`, docblock at `:25-30`

The `#2085` half is correct: `enabledCapabilities` is stamped at create and never retro-filled, so no existing connection gains anything. But the create path is the one this changes, and the PR body does not mention it:

- `apps/web/src/features/connections/components/eparagony-setup.schema.ts:72` omits `enabledCapabilities` **deliberately** ("the server defaults to the adapter manifest's supported set").
- `apps/api/src/integrations/application/services/connection.service.ts:653` is `rest.enabledCapabilities ?? defaultCapabilities`, and `defaultCapabilities` is the full manifest set (the `InventoryMaster`/`OfferManager` carve-out does not apply).

So from this merge on, every eparagony connection created through its own wizard - a form that collects `environment`, `posId` and credentials, and no invoicing configuration at all - is born with `['Fiscalization','FiscalRegistrationLocator','Invoicing','RegulatoryStatusReader']`. `ConnectionCapabilitiesPanel` renders **Invoicing ticked** on a connection nobody ticked it on.

I traced the blast radius rather than assuming it. It is bounded: `readSalesDocumentRouting` still requires an explicit `config.salesDocument.documentKind`, so a connection left alone is inert and cannot create an `ambiguous-connection-no-primary` block. And if an operator does set `'invoice'`, the refusal is an `EparagonyConfigException` → `failureMode: 'rejected'` → non-blocking. So this is a wrong claim plus a misleading enabled-state, not a live hazard. It still needs deciding rather than discovering later.

Either state the exception honestly in the body and in the docblock at `:40-45` (which currently asserts the opposite), or have the wizard send `enabledCapabilities: ['Fiscalization', 'FiscalRegistrationLocator']` explicitly - `dpd-setup.schema.ts:104` shows the repo already reasons about this default out loud - and replace the "omitted on purpose" comment with the reason.

**2. The manifest is a published contract and the docs still describe one capability**

`docs/architecture-overview.md:390` still reads *"registry capability `Fiscalization`"* and describes only the fiscalization adapter; `docs/capabilities.md` "Adapter coverage" for `InvoicingPort` lists KSeF / Infakt / Subiekt and not eparagony, in a file whose own header says *"When you add a port or sub-capability, update this file in the same PR."* I diffed the whole stack: the only doc touched anywhere in slice 1 or 2 is the one line of mockup HTML.

Landing it at the epic tip instead is fine, but nothing tracks it today and the integration branch is what reaches `main`. Say in the body which slice carries it, or carry it here.

---

### SUGGESTIONS

**The seller address is validated trimmed and shipped raw** - `...config-shape-validator.adapter.ts:224` / `:229` with `mapper.ts:754`

`:229` tests `ISO_ALPHA2_PATTERN.test(address.country.trim())` while `toSellerEntityAddress` copies `street`/`number`/`postalCode`/`city`/`country` across untouched. So `country: " PL "` saves clean and every invoice this connection issues carries a 4-character value in a field the package's own type declares as ISO 3166-1 alpha-2. Whether the vendor rejects such a document I could not verify. Test the raw value instead, or extend the docblock, which is already honest about case but silent about whitespace.

**Guard the shared-transport assertion against a vacuous pass** - `eparagony-adapter.factory.spec.ts:86`

`toBe(transportOf(...))` on the real private field of the real adapters is exactly the right test. The gap: rename the field and both sides read `undefined`, and `expect(undefined).toBe(undefined)` passes. One `toBeDefined()` (or `toBeInstanceOf(EparagonyHttpClient)`) makes the test that names the invariant also fail when the invariant becomes unreadable.

**`should build a receipts-only connection` asserts less than its comment claims** - `eparagony-adapter.factory.spec.ts:117`

Its config is byte-identical to `makeConnection()`'s default, so it re-runs the first test's construction and asserts `resolves.toBeDefined()`. Either drop it or make it assert that the invoice adapter *is* built and refuses at call time rather than at construction.

**One sentence on the non-core manifest names**

A new connection now persists `FiscalRegistrationLocator` and `RegulatoryStatusReader` into `enabledCapabilities`, and the first capability toggle silently drops both (`ConnectionCapabilitiesPanel` saves the `isCoreCapability`-filtered set). Pre-existing class (#576, Allegro carries seven), harmless because nothing gates on either name, but worth a line in the plugin docblock so "why did two capabilities disappear from this connection" has an answer.

**`merchantTIN` gets a non-empty check while the mapper hard-stamps `scheme: 'pl-nip'`**

Declining to guess the vendor's rule is the right instinct in general, but this adapter is not regime-neutral - `SETTLEMENT_CURRENCY = 'PLN'` and it refuses any other currency - so "10 digits" is a fact about this adapter rather than a guess, the same standing the ISO alpha-2 check already claims. Judgement call; one sentence either way.

---

### Verdict

🔄 **Request changes.** The design is not in question and nothing here is expensive: the blocking item is a two-line guard fix plus a spec.

1. Widen both mapper guards to `!= null` / `== null` in one commit, and add a `composeInvoiceDocument` spec for `merchantAddress: null`. *(BLOCKING)*
2. Correct the create-path claim in the body and the `eparagony-plugin.ts` docblock, and decide deliberately whether a new connection should come up with `Invoicing` enabled.
3. Land or explicitly assign the `architecture-overview.md:390` and `capabilities.md` updates.
4. The four suggestions above, at your discretion.

Verified clean and worth saying: the shared client is genuinely shared and neither adapter mutates transport state (idempotency keys are per-call `options.headers`, merged caller-first); nothing new is eager, so a `Fiscalization` resolution on a receipts-only connection costs one object and zero I/O; existing receipts-only configs cannot be locked out by the new keys, and there is a test pinning that; `eInvoicingHubEnabled`'s validation agrees with its only reader; no `typeof x === 'object'` without a null guard; no submitted value is echoed in any error message; and the mockup one-liner is a correct fix - `toRegulatoryClearanceResult` really does map `CONFIRMED` to `accepted`.

norbert-kulus-blockydevs:
Filed the BLOCKING item as #3274 so it is tracked whether or not it is fixed in this PR.

I re-verified the whole chain link by link before filing, rather than reasoning from the guard alone, because "a validator and a mapper disagree" is the kind of finding that is easy to assert and easy to get wrong:

- `connection-config.policy.ts` is a documented blind `as unknown as` cast, so TypeScript structurally cannot catch the mismatch.
- `eparagony-config.types.ts:167` declares `merchantAddress?: EparagonySellerAddress`, which is why `!== undefined` reads as correct.
- `eparagony-invoicing.adapter.ts:161-162` asserts the exact invariant that is broken: *"Composition failures throw `EparagonyConfigException` (`failureMode: 'rejected'`) BEFORE anything crosses the boundary."* A `TypeError` is neither.
- `invoice.service.ts:758-761` wraps the adapter call, and `classifyFailure` returns `'in-doubt'` for anything without a `failureMode`.
- `EditConnectionForm` has a raw JSON config editor, so this is reachable in the product today, not only by curl.

On severity: it is latent rather than common right now, because nothing writes these keys yet. It becomes the normal path as soon as the slice-3 seller-config form lands, since #2610's cleared-knob convention writes an explicit `null`.

The issue recommends fixing the mapper rather than the validator, and explains why both `:341` and `:395` have to move in one commit - fixing only `:341` relocates the crash to after `createDocument` has already succeeded, which is strictly worse. If you would rather take the other side and refuse an explicit `null` at the validator, that is equally correct as long as both sides agree; the issue says so under Assumptions.

Happy for this to be folded into this PR instead of landing separately, if you would rather keep the stack self-consistent.

norbert-kulus-blockydevs:
## Review findings addressed

All items from the tech-lead review resolved in `2979424a8`.

### BLOCKING — fixed

**Cleared `merchantAddress` crashed the invoice path.** Both mapper guards moved from `!== undefined` / `=== undefined` to `!= null` / `== null`, in one commit, matching the fix I recommended in #3274:

- `composeInvoiceDocument` (line ~341): `!= null`.
- `toIssuedDocumentSeller` (line ~395): `== null`.

Fixing only the composer would have relocated the crash to `toIssuedDocumentSeller`, which runs *after* `createDocument` and the status poll have already succeeded — that would turn a pre-call crash into one that loses a document the vendor legally created. Both had to move together.

Added two specs proving the mapper agrees with the validator's "null is absent" reading: one in `composeInvoiceDocument` (a config with a name but a `null` address must omit `merchantAddress` from the wire body, not throw), one in `toIssuedDocumentSeller` (same input must resolve `null`, not dereference the address). Verified re-checking the whole chain before writing the fix — `validateMerchantAddress`, `readNonEmpty`, both call sites, and `InvoiceService.classifyFailure`'s `in-doubt` default — all confirmed as described.

### IMPORTANT — both resolved

**1. Create-path claim corrected, and the underlying gap closed rather than just documented.** Traced the blast radius myself: `ConnectionService.create` defaults `enabledCapabilities` to the manifest's whole supported set when omitted, and the guided wizard omitted it — so every new eparagony connection would have been born with `Invoicing` ticked, silently, despite collecting no seller config. Fixed at the source: the wizard now sends `enabledCapabilities: ['Fiscalization']` explicitly (not the manifest's four — `FiscalRegistrationLocator`/`RegulatoryStatusReader` aren't `CoreCapability` values and would 400 the create DTO). Added two tests pinning the payload. Rewrote the plugin docblock to state this decision rather than the disproven "operator ticks the checkbox for everything" claim.

**2. Docs updated in this PR.** `docs/architecture-overview.md` §16 now names the second registry capability on the eparagony connection, and a new §14 entry describes the invoicing adapter (the shared-client rationale, the `CorrectionIssuer` omission, the opt-in wizard behavior, the null-guard fix). `docs/capabilities.md`'s `InvoicingPort` adapter-coverage line now lists eparagony.pl alongside KSeF/Infakt/Subiekt.

### SUGGESTIONS — all four applied

- **Country check now tests the raw value**, not a trimmed copy — `toSellerEntityAddress` ships `country` untouched, so validating a trimmed copy would have blessed `' PL '` into a field this package declares as two letters. Added specs for the padded-reject and the case-accept.
- **Shared-transport assertion hardened**: added `toBeInstanceOf(EparagonyHttpClient)` before the `toBe` equality, so renaming the private field fails the test instead of passing `undefined === undefined`.
- **Receipts-only factory test rewritten** to drive the real call-time refusal (`issueInvoice` throwing `EparagonyConfigException` with zero fetch calls) instead of re-running the default construction and asserting `resolves.toBeDefined()`.
- **`merchantTIN` decision recorded** as a comment: no format check, because a digit-count rule would be a mirror stricter than the vendor's own gate (the #2240 failure mode) — the adapter knows the scheme tag it stamps, not what the provider accepts.

### Verification

`pnpm type-check` clean (21/22 projects; parked two untracked scratch files that predate this branch to isolate the real result), `pnpm --filter @openlinker/web lint` and `pnpm --filter @openlinker/integrations-eparagony lint` both clean with zero `--fix` churn, `pnpm check:invariants` green except the pre-existing `.worktrees/` false positive the PR body already calls out (identical on a clean checkout). Slice 3 (#3269) was rebased onto this fix and force-pushed since its base moved.

piotrswierzy (commented):

**Verdict: ❌ Request changes** — one BLOCKING finding, which is a money-correctness gap rather than a design disagreement. The wiring in this PR is right and the architecture is right; what is missing is a clamp that its sibling #3260 has and this branch does not.

### The architecture holds

Taking the same questions as its sibling: this is a legitimate **second capability** on one plugin, not a blurring of ADR-042. The receipt stays on `FiscalizationPort`, the invoice gets its own adapter on `InvoicingPort`, and `dispatchCapability` carries exactly those two base capabilities with `FiscalRegistrationLocator` / `RegulatoryStatusReader` advertised-without-dispatch and narrowed by `is*` guard. ADR-026's country-agnostic rule holds by construction — **no `libs/core` file is touched by this diff at all**, so every Polish specific stays in the plugin.

The `eparagony-plugin.ts:2` docblock deserves specific credit for anticipating the two ways this change could have gone wrong silently:

- **#2085**: `enabledCapabilities` is stamped at create and never retro-filled, so no existing connection gains `Invoicing`. Stated, and correct.
- **ADR-041 exclusivity**: adding `Invoicing` to the manifest would otherwise make `ConnectionService.create`'s fallback mint new connections claiming it, which `selectInvoicingCandidates` would then turn into a second candidate beside an existing inFakt/KSeF connection — converting that install's one-click issue into a forced pick. The wizard sending `enabledCapabilities: ['Fiscalization']` explicitly (`eparagony-setup.schema.ts`) is the right fix and the reasoning is written down where the next reader will find it.

I checked the deeper exclusivity question independently: `readSalesDocumentRouting` returns a **single** `documentKind` per connection (`sales-document-kind.types.ts:100`), so a connection carrying both capabilities still resolves to exactly one kind and cannot be a candidate for both. The one-originating-document-per-order invariant is structurally safe. Good.

### Findings

**[BLOCKING]** — `libs/integrations/eparagony/src/infrastructure/adapters/eparagony-invoice.mapper.ts:657`, `allocateGroupNet`
> This branch clamps the residual in **one direction only**. It handles the shortfall (a last-member net below zero) and not the surplus (a last-member net **above that line's own gross**). The wire carries the line's net and derives tax as `gross_i − net_i`, so an unclamped surplus emits a **negative `taxValue`** onto a document bound for the tax authority — invisible to the summary checks, which reconcile perfectly either way, and not caught by the vendor either since `calculationValidation: NONE`.
>
> It is reachable on ordinary input, not a constructed edge. Three lines of 3, 3 and 1 grosz at 23%: the group's gross is 7, its tax rounds to 1, its net is 6; the first two lines round to 2 each, so the third computes `6 − 4 = 2` against a gross of 1, emitting `taxValue = 1 − 2 = −1`. And it is *systematically* reachable rather than incidental, because `toShippingLines` appends a per-rate share **last** while `groupByRate` preserves first-seen order — so a shipping share is the last member of its group whenever its rate already appeared above it.
>
> #3260 carries the fix as commit `3abbcc63b` ("clamp a rate group's residual in BOTH directions"), with a correct termination argument: the earlier headroom available to absorb a surplus is `taxGroup + surplus`, which is at least the surplus because `taxGroup` is never negative. Either cherry-pick that hunk here, or state in this PR's description that #3260 **must** land on `3192-eparagony-invoicing` before that branch merges to `main` — right now nothing written down prevents this version being the one that ships.

**[IMPORTANT]** — branch topology
> This PR and **#3260** both target `3192-eparagony-invoicing`, share three commits, and **neither is an ancestor of the other**. `git diff pr3267 pr3260` is 31 files, +1372/−877; both add `eparagony-invoicing.adapter.ts`, `eparagony-invoice.mapper.ts`, `invoice-tax-rate.policy.ts` and `eparagony-connection-config-shape-validator.adapter.ts` with divergent content. This PR is the superset on the **wiring** (`eparagony-adapter.factory`, the manifest, the plugin docblock); #3260 is the superset on the **arithmetic**. Whichever merges second conflicts across the whole slice, and a textual resolution that picks this side's mapper silently reinstates the BLOCKING defect above. The intended order needs to be stated in both descriptions.

**[IMPORTANT]** — `libs/integrations/eparagony/src/domain/policies/invoice-tax-rate.policy.ts`
> A second divergence from #3260 in the same class: this branch lacks `normalizeDecimalSeparator`, so a neutral rate of `23,00` resolves `Number('23,00') → NaN`, falls through the exemption lookup and hard-refuses the whole invoice — while core's own `parseTaxRatePercent` reads the same value as 23. Lower severity than the clamp (it refuses rather than emitting a wrong figure, which is the safe direction) but it is the same "the two branches diverged on money handling" problem and should converge the same way.

**[SUGGESTION]** — `eparagony-adapter.factory.ts`
> The factory hoist's docblock is admirably precise about what it does *not* buy — the OAuth token cache still lives on the per-call `EparagonyHttpClient`, so two capability resolutions for one connection still fetch two tokens. Worth confirming that is acceptable for the invoice lane specifically: `issueInvoice` blocks on a status poll, so a connection resolving both capabilities in quick succession doubles token traffic against a vendor publishing no rate ceiling (the manifest deliberately sets no `defaultRateLimit`). Not a blocker; the seam is in the right place for a later cache with a proper invalidation story.

### What I read closely, and what I did not

Read closely: `eparagony-plugin.ts` in full (manifest, dispatch table, the `enabledCapabilities` reasoning), `eparagony-adapter.factory.ts` and its interface, `allocateGroupNet` and the surrounding rate grouping, `invoice-tax-rate.policy.ts`, and the branch topology against #3260. Verified independently: `readSalesDocumentRouting`'s single-`documentKind` shape, and that no `libs/core` file is in the diff. Skimmed: the 818-line mapper spec and 552-line adapter spec, the docs and mockup edits.

### Merge readiness

❌ **Request changes**, on the clamp alone. Everything else here is ready — the capability wiring, the opt-in decision and its reasoning, and the manifest/dispatch agreement are all correct and unusually well argued.

Priority order:
1. Bring `allocateGroupNet`'s surplus clamp over from #3260 (`3abbcc63b`), or make the merge order a stated, enforced precondition on the integration branch.
2. Bring `normalizeDecimalSeparator` over for the same reason.
3. Record the intended merge order relative to #3260 in both PR descriptions.

norbert-kulus-blockydevs:
Pushed `80a6de53a` addressing your review.

**BLOCKING — `allocateGroupNet` surplus clamp.** Fixed, mirroring #3260's `3abbcc63b` fix for the identical bug: the last member of a rate group is now clamped at its own gross, and any surplus is lent to earlier members with headroom (mirroring the existing shortfall borrow-back). Spec replaces the single shortfall case with an `it.each` covering both directions (`[0.03, 0.03, 0.01] @ 23%` reproduces the surplus case exactly as you described), plus a new test asserting `documentLines` carries the repaired figures.

**IMPORTANT — `normalizeDecimalSeparator`.** Added, same as #3260: a single comma now normalizes to a decimal point before the numeric parse (`23,00` → `23`), so it resolves instead of falling through to a hard refusal. A thousands separator (`1,000`) still correctly fails the rate lookup. Two new spec cases pin both.

**IMPORTANT — branch topology / merge order.** Since both fixes now match #3260's commit `3abbcc63b` byte-for-byte in effect, a rebase or merge in either order converges rather than silently reinstating either defect. I've also rebased the downstream `3192-3-invoicing-surfaces` (slice 3) onto this branch's new tip so the stack stays consistent - it rebased cleanly with no conflicts.

**SUGGESTION — token-cache/rate-limit question.** No code change; this is a design question for a later slice (a per-connection client cache with a real invalidation story), not something to fix here.

Verification: `pnpm --filter @openlinker/integrations-eparagony test` (86/86 passing on the two touched spec files), `lint` and `type-check` clean.

piotrswierzy (approved):
## PR Review — #3267 `feat(eparagony): make the invoicing adapter reachable on a receipts connection`

Reviewed against `docs/architecture-overview.md` §§14/16/17, ADR-026, **ADR-041 §3a/3b**, ADR-042, and the advertised-without-dispatch rule in §Capability Abstractions. The central question for this slice is the one the epic creates: **one connection that can now produce both a fiscal receipt and an invoice.**

---

### Step 1 — Context, and the invariant I went looking for

`createFiscalizationAdapter` → `createAdapters` returning `{ fiscalization, invoicing }` over one `EparagonyHttpClient`; the manifest grows to `['Fiscalization', 'FiscalRegistrationLocator', 'Invoicing', 'RegulatoryStatusReader']`; the dispatch table gains `Invoicing`; the config-shape validator learns four optional keys; the guided wizard starts sending `enabledCapabilities` explicitly.

**ADR-041 §3a/3b holds, and I traced both directions rather than taking the PR's word for it.** Recording it because it is the thing that could have been catastrophic:

- `InvoiceService.assertNotInvoicedElsewhere` (`invoice.service.ts:493`) filters the invoice half on `record.connectionId !== requestedConnectionId`, but then calls `assertNoBlockingFiscalReceipt` (`:576-596`), which is **not** connection-scoped — `records.find((r) => r.blocksFurtherRegistration)` across every fiscalization connection, the requesting one included. A receipt on the same eparagony connection therefore blocks an invoice on it.
- The reverse holds symmetrically: `FiscalRegistrationService.assertNotAlreadyRegistered` calls `this.invoiceService.findBlockingInvoiceForOrder(orderId)`, likewise connection-agnostic, raising `OrderAlreadyHasInvoiceException`.
- Both write paths take the **same per-order lock**, `invoiceIssueLockKey(orderId)` → `` `invoice:issue:${orderId}` `` (`invoice.service.ts:295`, `fiscal-registration.service.ts:323`), so the read-then-act window is closed across kinds and across connections.
- Routing cannot manufacture a second candidate either: `AutoIssueTriggerService` (`auto-issue-trigger.service.ts:430-439`) builds **exactly one** `SalesDocumentRoutingCandidate` per connection, with `documentKind` read from `config.salesDocument.documentKind` — never derived from `enabledCapabilities` — so a dual-capability connection is one candidate with one kind, and `resolveSalesDocumentRouting` still resolves at most one `(documentKind, connectionId)` pair.

So this PR does **not** open a two-originating-documents path. Good.

### Other positive observations

- **`CorrectionIssuer` correctly absent** from the manifest with the adapter refusing a correction pre-call, and a spec pinning the absence so #3193 must flip the assertion rather than forget the manifest. That is the repo's own "a capability name enters the manifest together with the adapter that delivers it" rule (the Erli #980 precedent) applied correctly.
- **`RegulatoryStatusReader` and `FiscalRegistrationLocator` stay advertised-without-dispatch**, narrowed with their `is*` guards and absent from the dispatch table — matching the `CategoryBrowser` / `OfferCreator` pattern. `isRegulatoryStatusReader` is a bare method probe, and `EparagonyInvoicingAdapter.getClearanceStatus(record: InvoiceRecord)` matches the port signature exactly (note KSeF widens that parameter to `string | InvoiceRecordType`; this adapter does not, which is the more conservative choice).
- **`merchantTIN` is deliberately not a factory-construction check.** Refusing construction without it would stop every existing receipts-only connection registering a sale; the mapper refusing pre-call keeps the failure on the lane that needs the value. Correct call, and stated.
- **The `merchantAddress.country` check tests the raw value rather than a trimmed copy**, because `ConnectionConfigShapeValidatorPort` returns `void` and cannot normalise — so `' PL '` would otherwise be validated and then shipped. Validating what actually goes out is the right instinct and is rarer than it should be.
- The overstatement about token traffic was corrected in the code comment rather than left in the brief. Appreciated.

---

### Step 2/3 — Findings

**[IMPORTANT]** — `apps/web/src/features/connections/components/eparagony-setup.schema.ts` + `apps/api/src/integrations/application/services/connection.service.ts:649-653`

> The mitigation for the manifest-default problem is **browser-only**, and the repo's own standing rule (#2610) is that the browser is not the boundary: *"the form is not the only way in — curl and MCP bypass the browser entirely."*
>
> `ConnectionService.create` defaults `enabledCapabilities` to the manifest's supported set when the caller omits it (`:653`, `const enabledCapabilities = rest.enabledCapabilities ?? defaultCapabilities;`). From this PR on, an eparagony connection created by **any** caller that omits the field — `POST /connections` via curl, an MCP write, a script, a future non-wizard FE path — is born claiming `Invoicing` with no `merchantTIN`, no `merchantName` and no `merchantAddress`. The two consequences the PR docblock itself enumerates are then live: `selectInvoicingCandidates` (`sales-document-template-screen.tsx:78`, `sales-document-rule-composer-dialog.tsx:319`) gains a second candidate beside an existing inFakt/KSeF connection, and `deriveSalesDocumentRows` (`derive-sales-document-rows.ts:22-24`) tests `'Invoicing'` **before** `'Fiscalization'`, so a receipts connection renders as an Invoicing row in Settings → Sales documents.
>
> There is an exact server-side precedent for fixing it where it belongs: `:649-652` already carves `OfferManager` out of the manifest-derived default when the manifest also declares `InventoryMaster` (#1498, "publish-only unless the operator asks"). The same shape — exclude `Invoicing` from the *default* set when the manifest also declares `Fiscalization` — would make the opt-in hold on every path, and would let the wizard keep omitting the field. If you'd rather accept the gap, say so in the docblock: the current text reads as though the wizard change closed it.

**[IMPORTANT]** — Merge order, and what a hunk-wise rebase would ship

> This is a coordination finding, not a diff one, but it has a fiscal-safety consequence so it should be a merge checklist item rather than a note in two PR bodies.
>
> This branch and `3192-1-invoice-adapter` (#3260) diverge, and this one carries an **earlier** copy of the shared files: `eparagony-invoice.mapper.ts` is 875 lines here against #3260's 1006, `eparagony-invoicing.adapter.ts` 551 against 554. Specifically, this branch does **not** carry #3260's shared `eparagony-document-status.reader.ts` transport extraction, the `readSellerEntityAddress` null-guard hardening, or — most importantly — **`scripts/check-eparagony-deadline-mirror.mjs` and its `package.json` registration**. It *does* add `EPARAGONY_ISSUE_DEADLINE_MS` to `eparagony.constants.ts`.
>
> So merging this branch into the integration branch first, or resolving the rebase hunk-by-hunk in this branch's favour, ships the fiscal deadline constant with **no guard checking it against core's `MAX_SUPPORTED_PROVIDER_TIMEOUT_MS`**. The stated order (#3260 → rebase this → #3269) is right; please make it a blocking checklist item and take #3260's version of every shared file wholesale rather than resolving per-hunk. The good news is that `eparagony-connection-config-shape-validator.adapter.ts` is byte-identical between the two branches, so the one hunk the bodies worried about genuinely resolves as already-applied. (I diffed all three shared files to confirm.)

**[SUGGESTION]** — `apps/api/src/integrations/http/dto/create-connection.dto.ts:110-117`

> The docblock in `eparagony-plugin.ts` correctly explains that `@IsIn(CoreCapabilityValues, { each: true })` would 400 on `FiscalRegistrationLocator` / `RegulatoryStatusReader` — but that decorator only fires when the field is **present**, so the manifest fallback at `connection.service.ts:653` persists both non-core names anyway, and `ConnectionCapabilitiesPanel` silently drops them on the operator's first toggle. Harmless today (nothing reads either name off `enabledCapabilities`), and the docblock says so; a one-line `isCoreCapability` filter on `defaultCapabilities` would make the persisted value honest. Not this PR's to fix, but this PR is what makes the set four names long.

**[SUGGESTION]** — formatting divergence between the stacked branches

> This branch's copies of the shared files are formatted without trailing commas where #3260's have them (e.g. `this.connectionId` / `this.connectionId,` in `eparagony-invoicing.adapter.ts`). That will make the rebase noisier than the one real conflict warrants. Worth a `prettier --write` limited to the changed files — and *only* the changed files: a broad glob reformats ~1000 files in this repo because the committed style does not match the checked-in `.prettierrc`.

**[SUGGESTION]** — `libs/integrations/eparagony/src/application/eparagony-adapter.factory.ts`

> `createAdapters` now constructs both adapters eagerly on every capability resolution, so a `Fiscalization` resolve pays for an `EparagonyInvoicingAdapter` it discards (and vice versa). Both constructors are pure field assignments, so the cost is nil today — but the module-load assertion in `eparagony-invoicing.adapter.ts` now runs in every process that loads the plugin, including ones that never invoice. That is desirable (fail loud at boot), just worth stating in the factory docblock so it does not read as accidental.

---

### Step 4 — Final Assessment

#### Summary

A small, well-argued wiring slice whose hardest question — can one connection now produce two originating fiscal documents for one sale? — I traced end to end and answered **no**: the cross-kind guards are connection-agnostic in both directions, they share the per-order `invoice:issue:{orderId}` lock, and the auto-issue router still derives exactly one candidate per connection from `config.salesDocument.documentKind` rather than from the capability set. The capability-declaration discipline is right (`CorrectionIssuer` withheld, the two sub-capabilities advertised-without-dispatch and narrowed by guard), and the `enabledCapabilities` decision was correctly identified as needing a decision rather than a default. My main concern is that the decision was implemented only in the browser, against a rule this repo has written down precisely because the browser is not the boundary.

#### Merge Readiness

🔄 **Approve with changes** — no BLOCKING issue. The two IMPORTANT items should be settled before this reaches `main`: one is a small server-side change with an in-tree precedent, the other is a merge-procedure constraint.

#### Priority fixes

1. Move the `Invoicing` opt-in server-side — exclude it from `defaultCapabilities` in `ConnectionService.create` when the manifest also declares `Fiscalization`, following the #1498 `OfferManager`/`InventoryMaster` carve-out — or state plainly in the docblock that the curl/MCP path is unmitigated.
2. Make the merge order a blocking checklist item, and rebase by taking #3260's version of `eparagony-invoice.mapper.ts`, `eparagony-invoicing.adapter.ts` and `eparagony.constants.ts` wholesale, so `check-eparagony-deadline-mirror.mjs` and its `package.json` registration are never absent alongside `EPARAGONY_ISSUE_DEADLINE_MS`.
3. Optional: filter the manifest fallback through `isCoreCapability`; re-format the diverged files narrowly before the rebase.

Issue #2156, opened 2026-08-17
[CORE] Generalize AutoIssueTriggerService into the cross-capability gate (ADR-041 dec. 3a,3b,4,7)


norbert-kulus-blockydevs:
## Problem / Context

Part of epic #2154 (ADR-041 sales-document routing). Depends on sub-issue "sales-documents router" (must land first).

`AutoIssueTriggerService.onOrderTransition` (`libs/core/src/invoicing/application/services/auto-issue-trigger.service.ts`) today only knows about `Invoicing`-capable connections — it lists them, resolves exactly one via `selectPrimaryInvoicingConnection` (#2047), reads that connection's `triggerModel`, and enqueues an `invoicing.issue` job. It has no concept of a Fiscalization candidate at all.

ADR-041 decision 7: "`AutoIssueTriggerService` is the gate." Once a second document kind exists, this service must become the single dispatch point across BOTH capabilities — resolving one connection among ALL candidates (invoice + receipt), not just invoicing ones.

## Proposed Solution

Rewire `AutoIssueTriggerService.onOrderTransition`:

1. List connections with `Invoicing` **or** `Fiscalization` capability enabled (today only lists `Invoicing`).
2. Call the new `resolveSalesDocumentRouting(order, connections)` from sub-issue 1 instead of `selectPrimaryInvoicingConnection` directly (that helper's logic gets absorbed into the resolver).
3. On `route` outcome: validate the resolved connection actually supports the resolved `documentKind` — `InvoicingPort.getSupportedDocumentTypes()` for `'invoice'`, the fiscalization equivalent for `'fiscal-receipt'` (coordinate with the #2137 branch for the exact capability surface — check `FiscalizationPort` there). Unsupported → `unresolved-routing` / `unsupported-document-kind-on-connection`.
4. Dispatch to the right job type: `invoicing.issue` for `'invoice'` (unchanged), the fiscalization equivalent job for `'fiscal-receipt'`.
5. On `unresolved`: persist the reason via the existing #2100 mechanism (`order_records` columns), do not just log — this already works for the invoice-only reasons; extend it to also cover the new `unsupported-document-kind-on-connection` reason.
6. Preserve the existing "ONE-WAY EDGE" property (F3 in the current file's docstring): still injects no `OrdersModule` token. `apps/worker/test/integration/invoicing-auto-issue-boot.int-spec.ts` already asserts this — must keep passing.
[...]

Issue #1993, opened 2026-08-03
[TASK] Frontend — shared AccessGate primitive, applied to the MCP capabilities hint


norbert-kulus-blockydevs:
## Problem / Context

`ConnectionCapabilitiesPanel` renders an operator-facing hint about MCP tool staleness unconditionally:

```tsx
// apps/web/src/features/connections/components/ConnectionCapabilitiesPanel.tsx:120
{backsMcpTools ? <Alert tone="info">{MCP_CONNECTION_CHANGE_HINT}</Alert> : null}
```

`MCP_CONNECTION_CHANGE_HINT` (`apps/web/src/features/mcp-tokens/lib/tool-staleness-copy.ts:34`) reads *"MCP tools follow these capabilities — an already-connected agent must reconnect to see a change."* It describes the consequence of **changing** a connection's capabilities, so it is only actionable for a session that can actually change them (`connections:write`). The gate today is purely capability-based (`backsMcpTools`, computed at `:68`); the file contains no permission, role, or demo-mode reference at all, and the panel is mounted ungated at `apps/web/src/pages/connections/connection-detail-page.tsx:341`.

Result: a read-only session — including a public demo viewer, who holds only `connections:read` — is told to reconnect an agent it cannot have, over a control it cannot operate.

The frontend has no primitive for this. `useWriteAccess` + `ReadOnlyLock` (`apps/web/src/shared/ui/read-only-lock.tsx`) cover **write affordances**, and deliberately *show* them disabled in demo mode to advertise that the capability exists (#1615). That policy is the opposite of what informational content needs: an explanation aimed at someone who cannot act should not render at all. Consequently the four admin-only pages each hand-roll their own replacement (`pages/mcp-tokens/mcp-tokens-page.tsx:19`, `pages/ai-provider-settings/ai-provider-settings-page.tsx:25`, `pages/prompt-templates/prompt-template-detail-page.tsx:221`, `pages/prompt-templates/prompt-templates-list-page.tsx:176`) with an inline `role !== 'admin'` check and a copy-pasted `ErrorState`.

A second gap the primitive should close centrally: `usePermission` returns `false` while the session is still hydrating (`session.user.permissions` defaults to `[]` at `shared/auth/jwt-bearer-session-adapter.ts:131`), so a naive gate flashes hidden content and then reveals it. The `isReady` guard is currently spelled at only 2 of the 13 identity call sites, which is exactly the kind of detail that must not be re-implementable per call site.

[...]

Issue #1949, opened 2026-07-30
[MCP] Surface the client-reconnect hint where connections are enabled, not only on /settings/mcp-tokens


piotrswierzy:
Follow-up to #1932 (PR #1948). Recorded rather than absorbed, because fixing it exceeds that issue's acceptance criteria.

## Problem

#1932 added an operator-facing note to `/settings/mcp-tokens`:

> Which tools a client sees depends on which connections are enabled and what they support. Clients cache that list — after enabling or disabling a connection, reconnect the client for the change to appear.

That is the right content in a plausible place, but **it is not where the triggering action happens.** An operator who enables or disables a connection is on `/connections` or a connection-detail page. They will only read the MCP-tokens note if they already suspect the problem and go looking — which is precisely the state the note exists to prevent.

The observable failure is unchanged from #1932: enable a `ProductMaster` connection while an agent is connected, and `search_catalog` / `get_product` are live server-side but invisible to the agent until it reconnects. Disable one, and its tools linger in the client's cached list and fail on call.

## Scope sketch

The cheap version is a conditional hint on the connection enable/disable affordance — shown only when the instance actually has MCP tokens issued, so an operator who has never used MCP never sees it. `useMcpTokensQuery` already exists; the gate is `tokens.length > 0`.

Open questions worth settling before building:

- **Placement** — connection-detail actions panel, the `/connections` list row, or the enable/disable confirmation itself?
- **Gate** — "any MCP token exists" is cheap but coarse; it does not mean an agent is *currently* connected (stateless serving means OL cannot know that). Accept the coarseness, or skip the gate entirely?
- **Copy** — the MCP-tokens note is two sentences because it has room. A connection-page hint needs to be one line without losing the capability-gating half, or it degrades into "reconnect your client", which reads as a bug workaround.
- **Duplication risk** — two copies of the same guidance drift. Consider a shared constant, or accept divergence as the price of contextual wording.

## Non-goals

Same as #1932: no sessions, no `WebStandardStreamableHTTPServerTransport`, no `sendToolListChanged()`. Closing the staleness gap for real reverses part of [ADR-033](https://github.com/openlinker-project/openlinker/blob/main/docs/architecture/adrs/033-openlinker-as-mcp-server.md) and must start with an amendment, not UI. This issue is purely about *where* the existing documentation surfaces.

## References

- #1932 / PR #1948 — the note this extends
- ADR-033 § Phase 1 amendments — why the staleness exists and what it would cost to remove

[...]

PR #3269, opened 2026-09-15
feat(web): name the sales-document surfaces an operator drives the second role from

norbert-kulus-blockydevs:
Slice 3 of the #3192 stack. Slices 1 and 2 gave the receipts adapter an `InvoicingPort` and made it reachable; this is the operator half - the toggles that turn the role on, and the waiting state that follows.

Targets `3192-2-invoicing-wiring`, not `main`. The integration branch merges once.

## Most of this is naming renderings that already ship

`pending-submission` is drawn in four places already - the timeline's active "Awaiting submission" node, the regulatory badge, the lifecycle steps and the order-row cell. The mockup is explicit that a second ladder must not be built, so the three waiting hooks land on those and nothing was restructured. `TlNodeRow` gained one optional `testId` field; that is the whole structural change.

Each hook is emitted for `pending-submission` and for nothing else, so its presence asserts the STATE rather than that a lane exists.

## Genuinely new

**`capability-routing-note`** on the shared capabilities panel. A role makes a connection *eligible* and routes nothing to it; without the note an operator discovers that by waiting for a document that never arrives. Two decisions in it:

- Keyed on the connection **supporting** a sales-document capability, not on a role being enabled - it answers "what does turning this on do", which is asked before it is turned on.
- Its noun is derived from the capabilities held (both -> "sales documents", Invoicing only -> "invoices", Fiscalization only -> "fiscal receipts"), because "eligible to issue invoices" is plainly false on a receipts-only connection, which every eparagony connection is until an operator enables the second role.

It is not permission-gated, unlike the MCP hint beside it: that one tells you to go and reconnect an agent, this one states how the system decides, which is as true for a viewer.

**`clearance-cannot-retry`** plus the issued-document link, in the provider's own `invoiceDetailSection` - the slot `ksef`, `infakt` and `subiekt` already fill, wired identically.

**No retry control accompanies it, and that is the deliverable rather than an omission.** The provider publishes no call that triggers or re-triggers transmission, and a button that reaches nothing is worse than its absence because an operator who presses it concludes the push happened. The spec asserts the section renders **zero** `<button>` and zero `role="button"`, so a future "Send again" under any wording fails it.

The link is named a **visualisation** in its copy and its accessible name, because for this provider it is the vendor's HTML view rather than a PDF. `pdfUrl` reaches the frontend adapter-controlled with no server-side scheme validation, so it goes through `isSafeHttpUrl` before becoming an `href` - the `InvoicePdfLink` rule, restated because this is a second anchor over the same untrusted field.

## Three hooks deliberately NOT built

`provider-raw-status`, `provider-raw-processing-mode` and `sales-document-invoice-hash` have **no data path to the frontend**. `InvoiceRecord` carries no slot for raw provider values - no `regimeExtras`, no `statusDetails`, unlike `FiscalRegistrationRecord`, which has one and is read by `buildFiscalFieldItems` - and `InvoiceRecordResponseDto` exposes none.

Rendering them would mean inventing them. Closing it is a core change with its own migration and DTO, not a frontend one.

## Mockup

**Edited, in the second commit.** Per `docs/frontend-architecture.md` the mockup is either touched or justified, and this one was not still accurate on four points:

1. The three hooks above were drawn as though they were spec. They are now marked `NOT IMPLEMENTED` in the state and in the hook table, with a `note--gap` explaining why - a drawn hook nobody flagged reads as a scope cut rather than a gap.
2. `clearance-node-waiting` was drawn carrying an elapsed figure ("12h 04m"). The shipped node carries an **absolute** instant, because the only timestamp available is `invoice.updatedAt` - the record's last write, not when it entered `pending-submission` - so an elapsed rendering would assert a duration the data does not support. Persisting an entered-state instant is the fix and it is a backend change.
3. `sales-document-status-waiting` was drawn on the header pill. It ships on the clearance badge, the only element rendering the words "Awaiting submission"; `resolveInvoiceHeadline` words `pending-submission` as "Awaiting clearance", so the hook on the pill would have named an element whose copy contradicts the state it asserts. Splitting that headline arm is a copy change this slice did not take.
4. `capability-routing-note`'s keying and derived noun, as above.

## Two shared primitives widened

`Alert` and `StatusBadge` gain an optional `data-testid`, following the existing `Combobox` convention (the only prior instance). Purely additive - `data-testid={undefined}` emits no attribute.

A wrapper element was the alternative and was rejected: both primitives are routinely flex and grid children, so a spare `<div>`/`<span>` would change where they sit rather than only what they are called. This does widen the blast radius to two shared files, which is the thing in here most worth a second opinion.

## Verification

- `pnpm type-check` - pass
- `pnpm lint` - 0 errors (warnings are all pre-existing `explicit-function-return-type` in unrelated files); tree clean afterwards, no autofix churn
- All 90 `check:invariants` scripts pass. `check-repo-urls` reports hits only under `.worktrees/`, which are other sessions' checkouts rather than this tree.
- ~14 new tests were written and have **not** been run locally (this machine cannot run the suite); CI is the first execution.

Closes nothing on its own - #3192 closes when the integration branch merges.


piotrswierzy (approved):

**Verdict: ✅ Approve**, contingent on the first green CI run — see the merge-readiness note, which is the only substantive concern here.

### Verified

- **`pdfUrl` is scheme-checked before it becomes an `href`.** `eparagony-invoice-detail-section.tsx:54` gates through `isSafeHttpUrl` and only then reaches `href` at `:88`. The header comment is right that this is a *second* anchor over the same adapter-controlled field with no server-side validation, so restating the `InvoicePdfLink` rule rather than assuming it was inherited is the correct instinct.
- **The no-retry-control claim is asserted structurally, not by label.** `eparagony-invoice-detail-section.test.tsx:94-95` checks both `queryAllByRole('button')` and `querySelectorAll('button')` against zero. That covers a future "Send again" under any wording *and* anything wearing `role="button"` — which is the only form of this assertion that survives the next author. Given the provider publishes no call that triggers transmission, a button here would tell an operator a push happened when nothing reached the vendor; shipping the absence deliberately, with a test that defends it, is the right deliverable.
- **The three unbuilt hooks are correctly unbuilt.** `InvoiceRecord` genuinely carries no slot for raw provider values — no `regimeExtras`, no `statusDetails`, unlike `FiscalRegistrationRecord` which has one and is read by `buildFiscalFieldItems` — and `InvoiceRecordResponseDto` exposes none. Rendering them would mean inventing them. Marking them `NOT IMPLEMENTED` in the mockup rather than drawing them as spec is the difference between a recorded gap and a silent scope cut.
- **The mockup obligation is met.** `docs/frontend-architecture.md` § UX Mockups requires an implementing PR to touch the mockup or state why it is still accurate, and this one edits it on four specific points. Two of those are worth singling out: the `clearance-node-waiting` elapsed figure was removed because the only available timestamp is `invoice.updatedAt` (the record's last write, not when it entered `pending-submission`), so an elapsed rendering would assert a duration the data cannot support; and `sales-document-status-waiting` moved off the header pill because `resolveInvoiceHeadline` words that state "Awaiting clearance", so the hook would have named an element whose copy contradicts the state it asserts. Both are the mockup being corrected *by* the implementation rather than the implementation bending to a stale drawing.
- **`capability-routing-note` is correctly not permission-gated.** `docs/frontend-architecture.md` § Access Control draws the line at whether content is addressed to someone who cannot act — the MCP hint was removed (#1993) because it told a demo viewer to reconnect an agent it could not have. This note states how the system decides, which is as true for a viewer as for an admin. Deriving its noun from the capabilities actually held (receipts-only ⇒ "fiscal receipts") rather than hardcoding "invoices" is right for the same reason: every eparagony connection is receipts-only until a second role is enabled, so the generic wording would have been plainly false on the majority case.
- **The shared-primitive widening is fine.** `Alert` and `StatusBadge` gain an optional `data-testid` following the existing `Combobox` precedent; `data-testid={undefined}` emits no attribute, so it is purely additive. The rejected alternative is rejected for the right reason — both primitives are routinely flex and grid children, so a wrapper element changes where they sit, not just what they are called.

### Findings

**[IMPORTANT]** — merge readiness, not the code
> Two facts compound. This PR targets `3192-2-invoicing-wiring` rather than `main`, and it currently has **no CI checks at all** (`statusCheckRollup` is empty). The body also states the ~14 new tests "have **not** been run locally (this machine cannot run the suite); CI is the first execution". So at present nothing, anywhere, has executed this code — and because it is stacked, the first execution lands whenever the integration branch reaches `main`, which is the latest possible moment to discover a broken spec file. The code itself is small and well-argued; the risk is a test that does not compile or an assertion that never runs, which is precisely the class the repo cares most about. Get one green run on the stack base before merging upward.

**[SUGGESTION]** — `apps/web/src/shared/ui/alert.tsx`, `status-badge.tsx`
> The body flags this widening as "the thing in here most worth a second opinion", and I agree with the call but would note the direction of travel: this takes the `data-testid`-on-a-shared-primitive count from one (`Combobox`) to three. Each instance is individually justified, but the pattern couples specs to primitive markup, and the third instance is usually where a convention stops being an exception. Worth a line in the component conventions if a fourth appears, so the next author reaches for it deliberately rather than by precedent.

**[Positive]** — emitting each waiting hook for `pending-submission` and nothing else, so its presence asserts the STATE rather than merely that a lane exists, is a small decision that makes the hooks worth having. A hook that marks a lane regardless of state tests nothing.

### Merge readiness

✅ **Approve.** No migration, no conflict, correct dependency direction, the security-relevant path (`isSafeHttpUrl`) handled, and the mockup kept honest in both directions. The one condition is mechanical: this has never been executed, so land a green CI run on `3192-2-invoicing-wiring` before the stack merges upward.

norbert-kulus-blockydevs:
Addressed the review.

**[SUGGESTION] `data-testid` on shared primitives** — added a note to both `alert.tsx` and `status-badge.tsx` docblocks recording that this is now the third instance of the pattern (after `Combobox`), so a fourth prompts writing the convention down explicitly rather than extending the precedent silently. `06e47fcf2`.

**[IMPORTANT] merge readiness (no CI run yet)** — not a code change; this is the stack-base condition you flagged (get a green run on `3192-2-invoicing-wiring` before merging upward). Nothing to fix in this PR itself, flagging for visibility before merge.

piotrswierzy (approved):
# PR Review — #3269 `feat(web): name the sales-document surfaces an operator drives the second role from`

Tech Lead systematic review per `docs/code-review-guide.md`. Slice 3 of the #3192 stack, targeting `3192-2-invoicing-wiring`.

## Step 1 — Context

Read against `docs/architecture-overview.md` §§ 14 Invoicing / 17 Sales Documents, ADR-041, ADR-065, `docs/frontend-architecture.md`, and the three existing `invoiceDetailSection` slot implementations (`plugins/ksef`, `plugins/infakt`, `plugins/subiekt`). Five code files, one new plugin component + spec, two shared primitives widened, one mockup and three package docs updated.

## Step 2 — Systematic review

### Architecture and the ADR-065 invariants — all hold

- **One neutral projection, no re-composition.** Nothing here composes an invoice and a fiscal record itself. Both hosts resolve the slot from an invoice record's own connection (`features/orders/components/sales-document-panel.tsx:428,837`; `pages/invoicing/invoice-detail-page.tsx:146,282`), so `EparagonyInvoiceDetailSection` can never mount over a `FiscalRegistrationRecord` and render "Invoice visualisation" at a receipt. The PR's claim that the host owns the gate is true as written.
- **Status axes stay per kind.** The changes to `regulatory-status-badge.tsx` and `invoice-timeline.tsx` are additive `data-testid` attributes on the existing `RegulatoryStatus` axis. No enum is flattened, no fiscal axis is mixed in.
- **No fact the backend persists is re-derived.** The block/unresolved reasons are untouched, and so are `invoiceSupersedesBlock` / `canRetryInvoice` / the `trigger-model-manual` CTA — I checked; this PR does not go near them.
- **Deliberately not a second ladder.** The three waiting hooks land on the shipped `pending-submission` renderings, which is exactly what #3192's scope line demands. `TlNode.testId` (`invoice-timeline.tsx:51-56`) is the whole structural change, and it is emitted for the waiting node alone, so a spec asserting it asserts the state rather than the lane's existence.
- **Plugin boundary.** `plugins/eparagony/components/eparagony-invoice-detail-section.tsx` mirrors the infakt section line for line — same `invoice-detail-section reg-card` classes, same `regCardToneFor` import from `features/invoicing`, content-only, zero `platformType` literals outside the folder. All four CSS classes it uses exist in `apps/web/src/index.css`. ✅

### Correctness details I verified rather than took on trust

- `CAPABILITY_EXCLUSIVITY_PAIRS` (`features/connections/lib/capability-metadata.ts:58`) holds only the InventoryMaster/OfferManager pair, so Invoicing and Fiscalization really are independent toggles and the dual-role spec will pass.
- `renderWithProviders` wraps in `MemoryRouter` (`test/test-utils.tsx:1261`), so the new `<Link>` is safe in every spec that renders this panel.
- `sampleConnection` advertises neither sales-document capability, so the note is withheld from every pre-existing panel spec — no silent regression there.
- `/settings/sales-documents` is a real route (`app/routes/sales-documents.route.tsx:5`) and hosts both the per-country rule engine and the per-connection provider table. The link does not 404.
- `isSafeHttpUrl` exists at `shared/lib/is-safe-http-url.ts`, and the `data-testid`-on-a-primitive precedent is real (`shared/ui/combobox.tsx:72,98,271`).
- `check-ui-vocabulary.mjs` does not scan `features/connections`, `features/invoicing` or `plugins/`, so none of this copy is under that gate. (Which is the rider's whole point — see below.)

---

**[IMPORTANT]** — no CI has executed any of this
> `gh pr checks 3269` reports *"no checks reported"*: CI is `pull_request: branches: [main]`, and every slice targets a branch inside the stack. The PR body itself states the ~14 new specs "have not been run locally". The stack's own answer to this — the throwaway `[CI ONLY - NEVER MERGE]` preview #3273 — was created and closed on **2026-09-15**, while every commit in this PR is dated **2026-09-17**. So slice 3's specs, and the edits to two shared primitives that the whole app renders, have run **nowhere**. Re-open a preview PR off `3192-3-invoicing-surfaces` and get a green suite before the integration branch merges. This is the single highest-value item in the review: the reasoning in this PR is careful, and none of it is verified.

**[IMPORTANT]** — `apps/web/src/features/connections/components/ConnectionCapabilitiesPanel.tsx` (242-245) and `libs/integrations/eparagony/docs/setup-guide.md` (146)
> The note says the connection becomes eligible and that "Which orders actually get one is decided under Sales documents, **per country**". That is the rule-engine half only, and it omits the step whose absence produces precisely the symptom this note exists to prevent. Per `architecture-overview.md` § 14 (#2156): *"A connection only becomes a candidate once its operator sets `config.salesDocument.documentKind`"* — not silently defaulted, *"no error, no candidate"*. `pages/settings/sales-documents-page.tsx:11-18` restates it: that panel *"is the only place a connection's document kind, 'goes first' and trigger model are set at all, and a connection with nothing set there is not a routing candidate for the rule engine either"*, and `config.invoicing.triggerModel` defaults to `manual` on top. An operator who follows this note enables the role, authors a country rule, and still gets nothing — with no error anywhere. The note should name the second step ("…and give this connection a document kind there"), and setup-guide Step 3 item 4 should too rather than pointing at the rule composer alone.

**[IMPORTANT]** — `libs/integrations/eparagony/docs/setup-guide.md` (134-143)
> Step 3 instructs the operator to *"Fill in the invoice-lane fields: **Seller NIP**, **Seller name**, **Seller address**"* and to *"toggle **Relay to KSeF** on the connection"*. **None of those controls exists.** The plugin contributes no `connectionConfigSection` (`plugins/eparagony/index.ts` — its own docblock says "No structured-config section or credentials panel yet"), and slice 2 confirms the guided wizard *"never collects"* the seller fields. The "Developer aside" at :152 gives the truth — raw JSON config editor — but it is an aside under the numbered steps an operator follows. This is the same failure the PR argues so well against for the retry button: describing an affordance that does not exist is worse than its absence, because the reader concludes they did the step. Reword the numbered steps to name the raw config editor, or promote the aside.

**[IMPORTANT]** — the rider AC is unlanded anywhere in the stack
> #3192 carries `- [ ] A new provider/regulator name added to a shared apps/web folder fails the build`, and its rider argues explicitly that the guard belongs with *this* issue because *"this issue is the first one that puts the provider's name into front-end code that is not the plugin folder"*. No such guard exists (`ls scripts | grep neutral` → nothing) and no stack member adds one; #3260 and #3267 do not, and there is no slice 4. To this PR's credit it introduces **no** leak — I grepped the diff, and `eparagony` appears only under `plugins/eparagony/` and in docs, which is the legitimate home. So this is a stack obligation rather than a defect here, but it must land before the integration branch merges or the AC closes unmet. This slice is the natural home the rider names.

**[SUGGESTION]** — `ConnectionCapabilitiesPanel.tsx` (47, 116-119)
> The pair `Invoicing`/`Fiscalization` is spelled three times: once in `SALES_DOCUMENT_CAPABILITIES`, then again as two string literals in `supportsInvoicing`/`supportsFiscalization`. Deriving the two booleans from the tuple (or dropping the tuple and keeping the two booleans, `supportsSalesDocument = supportsInvoicing || supportsFiscalization`) leaves one spelling. As it stands a third sales-document capability would have to be added in two places, and only one of them fails visibly.

**[SUGGESTION]** — `apps/web/src/features/invoicing/components/regulatory-status-badge.test.tsx`
> `const others = ['not-applicable', 'submitted', 'cleared', 'accepted', 'rejected']` is a hand-maintained complement of the `RegulatoryStatus` union. It is complete today, but a sixth status would be added to the union and silently not covered here — which is the one case this spec exists for, since the hook's value is that it is emitted for `pending-submission` and nothing else. Derive it from the status values array minus `pending-submission`.

**[SUGGESTION]** — `shared/ui/alert.tsx` (18-28) and `shared/ui/status-badge.tsx` (28-40)
> The PR's own comments say "if a fourth needs it, write the convention down". The third instance is where a pattern becomes a convention — write the one-liner into `docs/frontend-architecture.md` now, in this PR, rather than leaving a third identical comment asking the next author to do it. The prop itself is right: purely additive, `data-testid={undefined}` emits nothing, and the wrapper-element alternative really would move both primitives inside their flex/grid parents.

**[SUGGESTION]** — `ConnectionCapabilitiesPanel.tsx` (238)
> The routing note is permanently-rendered static explanatory copy inside `Alert tone="info"`, which renders `role="status"` — a live region announced on mount. It matches the MCP hint's existing treatment one block up, so this is consistency rather than a new fault; worth noting only because the number of always-on `role="status"` regions on this page is now two.

### Positive observations

- The negative assertion is the right one and is written the right way: `queryAllByRole('button')` plus a raw `querySelectorAll('button')` inside the section (`eparagony-invoice-detail-section.test.tsx:88-100`), so a future "Send again" under *any* wording fails. Naming candidate labels would have proved nothing.
- `isSafeHttpUrl` re-applied over `pdfUrl` with the reason stated at the point of use, plus a spec that the whole region withdraws rather than rendering an empty card for a `javascript:` value — a second anchor over the same untrusted adapter field, guarded rather than assumed.
- Keying the note on **supported** rather than **enabled**, with the reasoning tied to the existing #1949 precedent in the same file, and the derived noun so a receipts-only connection is not told it may issue invoices.
- The mockup was edited rather than silently diverged from, and the three unimplementable hooks are marked `NOT IMPLEMENTED` with the cause (`InvoiceRecord` carries no `regimeExtras`/`statusDetails` slot). A drawn hook nobody flagged would have read as a scope cut — that judgement is right.
- Correcting `clearance-node-waiting` from an elapsed figure to an absolute instant, because `invoice.updatedAt` is the record's last write and not the entry into `pending-submission`, is exactly the kind of "do not assert what the data does not support" call this codebase asks for.

---

## Step 4 — Final Assessment

### Summary
A careful, well-argued frontend slice that does the honest version of the work: it names the states an operator drives the second role from without building a second ladder, and its load-bearing deliverable is a *negative* — that no retry control exists beside the waiting state, asserted structurally. The architecture holds on every axis I checked: per-kind status axes intact, no re-derived backend facts, the plugin slot mirrors its three siblings, and the shared-primitive widening is minimal and precedented. The most important concern is not in the code: **nothing has executed these tests**, because CI does not run on stack branches and the preview PR predates every commit here. Behind that sit two operator-facing accuracy gaps — the routing note and the setup guide both omit the per-connection document-kind step whose absence causes exactly the silent no-document outcome the note exists to prevent, and Step 3 tells an operator to fill in fields the product has no form for.

### Merge Readiness
🔄 **Approve with changes** — no BLOCKING issues. Approved for the stack; the four IMPORTANT items should be addressed before the `3192-eparagony-invoicing` integration branch merges to `main`.

### Priority fixes
1. Re-run CI against the stack tip (a fresh throwaway preview off `3192-3-invoicing-surfaces`) — the ~14 new specs and the two shared-primitive edits have run nowhere.
2. Name the per-connection document-kind step in the routing note and in setup-guide Step 3; "decided per country" alone leaves an operator with a silently non-candidate connection.
3. Reword setup-guide Step 3 items 2–3 so they describe the raw config editor rather than fields and a toggle that do not exist.
4. Land the rider's `apps/web` neutral-vocabulary guard before the integration branch merges (here or in a slice 4).
5. Optional: single-source the `Invoicing`/`Fiscalization` pair; derive the badge spec's status complement; write the `data-testid`-on-primitive convention into `docs/frontend-architecture.md`.

PR #1923, opened 2026-07-29
perf(test-kit,api): cut integration-suite time - per-test truncation, shutdown sleep, cold jest cache, shared PrestaShop container (#1920)

norbert-kulus-blockydevs:
Implements the three measured, self-contained items from #1920. Item #2 (pre-baked PrestaShop image) and item #4 (merging suite files) are **deliberately not here** - see *Scope* below.

## 1. Bounded in-flight drain instead of a flat 2 s sleep on shutdown

`WebhookToJobHandler.stopConsumptionLoop` slept `setTimeout(2000)` on every shutdown, "to let in-flight messages complete". Because ~77 int-specs tear the harness down, that was ~2 s x 77 paid per CI run, and it gave no real guarantee (a slow message was cut off at 2 s anyway).

The handler now tracks the `processMessage` call actually in progress and awaits *that*, bounded by the same 2 s. Nothing in flight (the common case, including every int-spec teardown) means shutdown returns immediately.

It deliberately does **not** await the consume loop itself: the loop parks in `XREADGROUP ... BLOCK 5000`, so awaiting it would stall shutdown for up to 5 s. `BLOCK_MS` is left alone, and `quit()` is kept - swapping it for `disconnect()` was measured on the same sample and buys **nothing** (see the issue's Verification log; that was a refuted hypothesis, not an oversight).

## 2. Truncate only the tables a test actually dirtied

`truncateTables` issued one `TRUNCATE ... CASCADE` per configured table on **every** `afterEach`, all 18 of them, whether the test touched them or not. Measured against the live test container:

| Statement | Time |
|---|---|
| `SELECT 1` (round-trip baseline) | ~0-1 ms |
| `TRUNCATE` 1 table | ~9-10 ms |
| `TRUNCATE` the 18 configured tables | **~207 ms** |
| `TRUNCATE` all 46 tables in the schema | ~390 ms |

The cost is linear in the **table count and flat in the row count** - ~10 ms per table even when empty - while a typical test dirties two or three. At ~207 ms x 485 api tests that was ~100 s per run spent clearing empty tables.

Now: one probe round-trip (`SELECT 't' WHERE EXISTS (SELECT 1 FROM "t")` UNION ALL'd over the list) narrows the list, then a single `TRUNCATE` clears just those. A test that dirtied nothing issues no `TRUNCATE` at all. Measured: **~207 ms -> ~13 ms** per reset.

Two things that look like the fix and are not (both measured, both rejected - don't re-litigate them without new numbers):
[...]

Issue #1285, opened 2026-07-01
[BUG] Infrastructure - Integration test Testcontainers leak on self-hosted CI runner exhausts docker0 bridge FDB, breaking pipelines


norbert-kulus-blockydevs:
## Problem / Context

Reported by DevOps (Tomasz Warczykowski): the self-hosted CI runner accumulated 1000+ orphaned Docker containers over ~3 days of `test-integration` runs. After a manual cleanup, a single fresh CI run still leaves ~124 containers running (Postgres/MySQL/Redis started by Testcontainers). This is exhausting the Linux bridge FDB table on `docker0` (kernel errno `EEXFULL`, "exchange full"), which then breaks *all* integration tests outright — the failure happens in Jest's `globalSetup`, before a single test runs — and has caused pipelines to stall/get stuck (seen in PRs #1262 / #1263).

This has two distinct angles that both need to land:

**(a) Immediate/operational**: pipelines are currently broken or unreliable because the runner's bridge/container state is already saturated. This needs a reactive fix independent of any code change — someone needs to clear the runner's stale container/network state (and ideally a recurring safety net) so CI is usable again.

**(b) Root cause / code**: OpenLinker's own integration-test tooling does not reliably stop the Postgres/Redis/MySQL containers it starts, so every CI run (successful or not) leaks containers instead of cleaning them up. Investigation traced this to three independent, compounding bugs:

1. **`apps/api/test/jest-integration.cjs` never wires `globalSetup`/`globalTeardown`.** Contrast with `apps/worker/test/jest-integration.cjs`, which correctly sets `globalSetup: '<rootDir>/test/integration/setup-global.ts'` and `globalTeardown: '<rootDir>/test/integration/teardown.ts'`. apps/api *has* the equivalent files ready (`apps/api/test/integration/setup.ts`, `apps/api/test/integration/teardown.ts` — the latter calls `stopHarness()`, which re-exports `stopContainers()` from `@openlinker/test-kit`), but neither is referenced in `jest-integration.cjs`, so they never run. `libs/test-kit/src/harness.ts`'s `IntegrationTestHarnessImpl.teardown()` explicitly documents this reliance: "Containers are stopped by the global teardown path (`stopContainers`) so suite-scoped teardown can keep the containers warm for the next int-spec in the same Jest worker." Since apps/api's config never invokes global teardown, the Postgres+Redis containers backing its ~68 int-spec files are **never stopped in-process**, on any run, clean or not.

[...]

PR #3276, opened 2026-09-16
perf(ci): three fixes for the Integration Tests job - per-worker databases, duplicate type-checking, and an oversized worker count (about 3x faster)

### LINKED ISSUE #3263: [TECH-DEBT] DX - Integration Tests job takes ~25 min and blocks every PR
## Problem

The `Integration Tests` job is the slowest thing in the pipeline — **~25 minutes**, and everything else waits on it.

Two reasons, and the second one was a surprise.

**1. The tests ran one file at a time.** The whole suite shared one database and one Redis. Between tests the harness wipes both. With several workers, one worker wipes the data another is using — so parallelism was switched off on purpose.

**2. Every test file re-checked the TypeScript types of the whole project.** An empty test file — no imports, one trivial assertion — took **11.9 seconds**. Jest does that work once per *file*, so with 187 files we paid it 187 times before a single test ran. The separate `Type Check` job already checks the same types, in under 3 minutes, in parallel.

## What was done

1. Each test worker gets its own database and its own Redis slot, inside the same containers. No new containers, no extra runner slots.
2. Tests stop re-checking types while running. Test files are now type-checked by the `Type Check` job instead, which previously skipped them entirely.
3. Six workers instead of one.

## Result

| | before | after |
|---|---|---|
| **`Integration Tests` job** | **24 min 54 s** | **7 min 28 s** |
| `apps/api` (190 files) | 14 min 09 s | 3 min 10 s |
| `apps/worker` (30 files) | 3 min 24 s | 1 min 11 s |

Same tests, all green.

## Things we tried that did not work

| tried | result |
|---|---|
| Run the api and worker suites at the same time | **Worse** — 20 min 31 s → 35 min+. Reverted. |
| Postgres tuning (`fsync=off`, data in RAM) | **No measurable difference.** The bottleneck was the compiler, not the disk. |
| SWC instead of ts-jest | **Does not start** — two TypeORM entities import each other; tsc tolerates it, SWC does not. |
| 8 workers | Exhausted the runner — see below. |

## Worth knowing for next time

**A CI runner is limited to 24 GiB of memory and 8 CPUs.** At 8 workers the runner measured **24 GiB / 24 GiB** and **800% of 800% CPU**, and the kernel then killed the Postgres and Redis the tests need.

6 workers measured *no slower* than 8 (190 s vs 193 s) — past that point the workers were queueing for CPU rather than working.

norbert-kulus-blockydevs:
Makes the `Integration Tests` CI job roughly **3x faster**, with no extra runner slots and no tests removed.

## Times

| | before | after |
|---|---|---|
| **`Integration Tests` job** | **24 min 54 s** | **7 min 28 s** |
| `apps/api` tests (190 files) | 14 min 09 s | 3 min 10 s |
| `apps/worker` tests (30 files) | 3 min 24 s | 1 min 11 s |

Same tests, all green.

## What actually changed

**1. Every test worker gets its own database.**

The whole suite shared one database and one Redis. Between tests the harness wipes both (`TRUNCATE` + `flushDb`). With one worker that is fine. With several, worker A wipes the data worker B is in the middle of using — so the suite was pinned to one worker and ran strictly one file at a time.

Each worker now gets `openlinker_test_<worker-number>` and its own Redis slot, inside the **same** containers. No new containers, no new runner slots.

**2. Tests no longer re-check TypeScript types while running.**

This was the big one, and it was invisible. An **empty** test file — no imports, one `expect(1).toBe(1)` — took **11.9 seconds**. Same file with type-checking off: **0.5 seconds**.

ts-jest was rebuilding and type-checking the whole project, and Jest does that once per test *file*. 187 files, so we paid it 187 times before a single test ran.

Those types are already checked by the separate `Type Check` job, in under 3 minutes, in parallel. We were doing the same work twice.

**3. Six workers, not one.**

Not eight — see below.

## What this cost, and how it is paid

Turning type-checking off during the run removed the only thing that was type-checking test files at all: `tsconfig.type-check.json` excludes `test/`. So this PR adds `tsconfig.test-check.json` for both apps and a second step in the `Type Check` job. Both apps pass it today, and that job runs in parallel, so it costs nothing on the clock.

## What we tried that did not work

Recorded so nobody spends the runs again.

| tried | result |
|---|---|
| Run the api and worker suites **at the same time** in one step | **Worse**: 20 min 31 s → 35 min+. Reverted. A runner has 8 CPUs, so two suites just fight over them. |
| **Postgres tuning** (`fsync=off`, data directory in RAM) — the fix every article recommends | **No measurable difference** across four interleaved runs. The bottleneck was the compiler, not the disk. |
| **SWC** instead of ts-jest | **Does not start.** `Cannot access 'AttributeMappingOrmEntity' before initialization` — two TypeORM entities import each other, which tsc tolerates and SWC does not. Its documented workaround did not help. |
| **8 workers** | Green at first, then failed once the suite got faster — see below. |

## Why 6 workers and not 8

**A CI runner is limited to 24 GiB of memory and 8 CPUs.** Measured on the runner during an 8-worker run:

```
memory   24 GiB / 24 GiB
cpu      800.58% of 800%
```

At that point the kernel kills processes inside that limit, including the Postgres and Redis the tests depend on. That is the `exit=137`, the `container ... is not running`, and the `ECONNREFUSED` storms that followed.

Neither 8 workers nor the transform change is wrong alone — 8 was green before. Removing the 11.9 s per-file stall means the same peak load now arrives in a quarter of the time instead of being spread out by the compiler.

**6 workers is not a compromise.** It measured *no slower* than 8 (190 s vs 193 s), because at 8 the workers were queueing for CPU rather than working.

Closes #3263


piotrswierzy (commented):

**Verdict: ❌ Request changes** — one BLOCKING issue. The per-worker isolation scheme is well designed and I could not fault it; what is missing is that **the PrestaShop container was left out of it**, and it is the one shared resource in this suite that has no per-worker identity.

### What I verified and found sound

The core of the change is right:

- **Per-worker Postgres + Redis is genuine, not nominal.** `applyWorkerScope()` (`libs/test-kit/src/containers.ts`) sets `DB_DATABASE=openlinker_test_<JEST_WORKER_ID>` and `REDIS_DB=<id>`, and `createWorkerDatabases()` creates them once in the realm that booted the container. `harness.ts:292` uses `flushDb()` — per logical DB, not `FLUSHALL` — so the reset genuinely cannot cross workers.
- **`resolveWorkerId` deliberately does not clamp**, and the docblock says why: clamping would hand two workers one database and silently reinstate the cross-worker truncation. Failing loudly on a database that was never created is the honest failure. That is the right call and easy to get backwards.
- **`applyWorkerScope()` is called on all three `startContainers()` paths** including both reuse paths, which is necessary because a worker realm only ever reaches those — and it runs *before* `applyEnvOverrides`, so an explicit `config.env` still wins.
- **The worker suite needed its own mechanism and got one for a stated reason.** `apps/worker/test/integration/harness.ts` starts its own containers and never calls `startContainers()`, so `setup-worker-scope.ts` exists to cover it — via `setupFiles` rather than `setupFilesAfterEnv`, because `setup-each.ts` imports `./setup` at module load and scoping from there would race the import graph. Correct, and correctly explained.
- **The `isolatedModules` cost is actually paid.** `tsconfig.type-check.json` excludes `test`, so ts-jest really was the only checker; `apps/api/tsconfig.test-check.json` + `apps/worker/tsconfig.test-check.json` + the new Type Check step restore it, in a job that runs in parallel. The note on not narrowing `types` (it would drop `@types/multer` and produce false errors about code this config is not checking) is the kind of detail that gets rediscovered painfully.
- **`MAX_TEST_WORKERS = 15`** is derived from a real constraint (Redis ships 16 logical DBs, `JEST_WORKER_ID` is 1-indexed) rather than picked, and raising it is correctly described as needing a different change.

The "what did not work" table is also worth having: Postgres `fsync=off` measuring nothing across four interleaved runs is exactly the result that stops the next person spending a day on it.

### Findings

**[BLOCKING]** — `apps/api/test/integration/helpers/prestashop-container.helper.ts:1090-1123`, reachable because of `apps/api/test/jest-integration.cjs:55`
> **The shared PrestaShop container's boot dedup is a cross-process check-then-act, and `maxWorkers: 1` was the only thing making it safe.**
>
> `startSharedPrestashopContainer()` reads a record file (`readLiveSharedRecord()` → `existsSync` → `JSON.parse` → `docker inspect`), and on a miss boots PS + MySQL and `writeFileSync`s the record — **only after the boot resolves**, which the helper itself budgets at 60–90 s warm and 5–10 min cold. There is no lock file, no `O_EXCL`/`wx` create, no advisory lock. Four api specs reach the PS container (`prestashop-order-fulfillment-update`, `prestashop-webhook-provisioning`, `prestashop-harness-smoke`, `allegro-prestashop-carrier-mapping`), and `jest-integration.cjs` declares no `testSequencer`, no project split and no `runInBand` — so at `maxWorkers: 6` jest is free to start several of them concurrently. Two workers entering that window both see no record and both boot.
>
> Three consequences, in descending order of how much I'd worry:
> 1. **Container leak on the persistent self-hosted runner.** Only the last `writeFileSync` survives, so `stopSharedPrestashopContainer()` — which `globalTeardown` drives *from the ids on disk* — stops one pair and the others are orphaned. The helper's own comment at `:1113` names this: *"Cannot hand ids to globalTeardown -> would leak on the persistent CI runner (the #1285 failure mode)."* The existing `Sweep orphaned Testcontainers` step in `ci.yml` only catches containers from *finished* runs, so they survive the run that created them.
> 2. **Memory.** Each boot is a PS + MySQL pair. The PR's own analysis fixes the worker count at 6 on the grounds that "6 leaves two of the eight CPUs for the Postgres, Redis and PrestaShop containers that share the same cap" — that budget assumes **one** PS pair. Two or three is precisely the 24 GiB / 800% condition this PR documents as producing `exit=137` on Postgres and Redis.
> 3. **Shared-state assertions.** `docs/testing-guide.md` § PrestaShop Testcontainer Pattern states *"Nothing resets PS between specs. Assert only over OL-side results, or filter by ids your own spec created."* That guidance was written for **sequential** reuse, where a spec's id-scoped window is uninterrupted. Concurrent specs against one PS instance interleave writes inside that window, so the documented mitigation no longer fully covers the case. `prestashop-webhook-provisioning` is a further wrinkle: it deliberately takes its *own* container because it asserts what the module writes on first install, so it adds a second pair by design.
>
> This is scheduling-dependent, which is why CI is green here — but a nondeterministic container leak on a persistent runner is the wrong kind of intermittent to ship into a change whose purpose is CI reliability.
>
> Three fixes, cheapest first. **(a)** Boot the shared PS container in `globalSetup`, before workers fan out — one realm, no race, and it matches how Postgres/Redis are already primed. **(b)** Make the record file an atomic claim: `openSync(file, 'wx')` before booting, with losers polling for the completed record. **(c)** Pin PS specs to one worker via a custom `testSequencer` or a second jest project at `maxWorkers: 1`. (a) is the smallest change and removes the race by construction rather than by guarding it.

**[IMPORTANT]** — `apps/api/test/jest-integration.cjs:49-54` — the comment states the opposite of what the code does
> *"Resolved from `OL_TEST_MAX_WORKERS`, defaulting to 1 - so a local run with no env set behaves exactly as it did before."* `resolveTestWorkers()` returns `DEFAULT_TEST_WORKERS = 6` when the env var is absent, with **no `process.env.CI` gate at all** (`jest.test-workers.cjs:75-81`). So a contributor running `pnpm test:integration` locally now gets 6 workers — 6 × (own DB + own Redis DB + full AppModule graph), plus the PS race above on their own machine. Note the unit tier's sibling resolver *does* gate on CI (`jest.unit-workers.cjs` — `LOCAL_UNIT_TEST_WORKERS`, justified by `.husky/pre-commit`), and that asymmetry is real and defensible for the integration tier (it only runs against Docker on a machine that opted in). Either add the CI gate or fix the comment — right now the file's own documentation is wrong about its own default, in the one place a reader would check.

**[IMPORTANT]** — `docs/testing-guide.md` is untouched, and it documents the inverted model
> This PR changes no docs. The guide still says `- 🔄 **Serial Execution**: Must run sequentially (maxWorkers: 1)` (`:98`) and, under Troubleshooting, *"Verify `maxWorkers: 1` in `jest-integration.js` (prevents parallel execution conflicts)"* (`:928`) — advice that, followed today, un-does the change. The per-worker database/Redis scheme, the `MAX_TEST_WORKERS = 15` Redis ceiling, the `OL_TEST_MAX_WORKERS=1` bisect escape hatch, and the "check your specs are in type-check scope" rule for `isolatedModules` all belong there. Per `CLAUDE.md`, `docs/lessons.md` is also the ledger for the 8-worker/`exit=137` finding — #3275 adds seven entries for its half of this work; this half adds none.

**[SUGGESTION]** — `.github/workflows/ci.yml`, jest-integration cache key
> Dropping `${{ github.run_id }}` from the key is right (the old key could never hit, only restore-key-prefix-match) and adding `jest.test-workers.cjs` to the hash is a good catch, since the worker count now determines how many databases exist. Worth a one-line comment saying why the run_id left, so it is not "restored" later as a cache-poisoning worry — jest's own key folds content hashes, which is what makes the stable key safe.

### Merge readiness

❌ **Request changes**, on the BLOCKING item alone. Everything else here is good work and the measurements are convincing — the 11.9 s-per-empty-file finding is the kind of thing that hides in plain sight for years, and the 6-vs-8 analysis ("6 measured *no slower* than 8 because at 8 the workers were queueing for CPU") is the right way to settle it.

Priority order:
1. Close the PS shared-container boot race — `globalSetup` pre-boot is the smallest fix.
2. Fix the `maxWorkers` comment (or add the CI gate it describes).
3. Update `docs/testing-guide.md`; it currently tells readers to undo this PR.

norbert-kulus-blockydevs:
Addressed all findings from the review (e4f6777):

**BLOCKING — shared PrestaShop container boot race**: `startSharedPrestashopContainer()` now dedups its boot with an atomic filesystem claim (`open(file, 'wx')`) instead of a plain check-then-act on the record file. Exactly one worker boots; the rest poll for its completed record. A claim whose holder died mid-boot (crash/kill) is reclaimed by age rather than left to wedge every remaining spec in the run out to a timeout.

**IMPORTANT — `maxWorkers` comment was wrong**: both `jest-integration.cjs` files said the default was 1 with no env set; the real default is 6, with no CI gate (a deliberate choice for this tier). Fixed both comments to say so.

**IMPORTANT — `docs/testing-guide.md` documented the inverted model**: it still told readers integration tests "must run sequentially (`maxWorkers: 1`)" and to verify that setting when debugging a hang — advice that undoes this PR if followed. Updated both spots to describe the per-worker database/Redis scheme and the `OL_TEST_MAX_WORKERS=1` bisect escape hatch. Also added the `docs/lessons.md` entry for this half of the work (the sibling PR already carries entries for its half).

**SUGGESTION — stale cache-key comment**: the `ci.yml` comment said the jest transform cache key was scoped "plus the run id", which stopped being true once `run_id` was dropped from the key. Rewrote it to explain why there's no run-scoped component and to mention `jest.test-workers.cjs`, which the key now also hashes.

piotrswierzy (approved):
# Tech Lead review — PR #3276 (Integration Tests speed-up)

Reviewed against `docs/code-review-guide.md` (4-step workflow), `docs/testing-guide.md`
(§ Testcontainers Explained, § PrestaShop Testcontainer Pattern), and the baseline `ci.yml` on `main`.
I went at the gate-preservation question first — *does anything stop being able to fail?* — then at
cache-key soundness, the per-worker isolation claim, and resource sizing against the runner's real
cgroup rather than the host.

No blocking findings. The isolation work is correct where it is hardest to be correct, and I verified
it rather than taking the description's word for it.

---

## What I verified holds

**Type-checking of the integration tier genuinely moves rather than disappearing.** This is the one
thing that could have made a check unable to fail, so I traced it end to end:

- `apps/api/tsconfig.type-check.json` on `main` carries `exclude: ["node_modules","dist","test"]`, so
  before this PR ts-jest was indeed the only checker of `test/**`. The PR's framing is accurate.
- `apps/api/tsconfig.test-check.json` redeclares **both** `include` and `exclude`, so the parent's
  `"test"` exclusion is replaced wholesale rather than merged — `include: ["test/**/*"]` really is in
  scope. Same for `apps/worker`.
- The new step lands in the `type-check` job, which `ci.yml`'s own header notes is *"self-hosted and
  intentionally NOT gated, so it always reports"* — and it is the repo's **only** required status check
  (`main branch protection` ruleset, `required_status_checks: [{context: "Type Check"}]`). So integration
  specs have gone from checked-only-incidentally-by-ts-jest to checked by the one required gate. That is
  a net strengthening.
- The two `run:` lines are a multi-line block under Actions' default `bash -e`, so a failure in the first
  cannot be masked by the second.

**Per-worker isolation is real on both axes, not just the Postgres one.** The Redis half is the easier
one to get wrong, because it only works if every client in the graph honours `REDIS_DB`:

- `libs/shared/src/redis/redis-config.module.ts:30` and `apps/worker/src/events/events-consumer.module.ts:34`
  are the only two places a client is constructed, and both read `REDIS_DB`. Everything else in the tree
  (`RedisSyncLockService`, `RedisStreamsJobEnqueueService`, `RedisStreamsEventPublisher`,
  `RedisCacheAdapter`, `RedisPickupPointQueryStatsAdapter`, the plugin-sdk rate limiter) injects the single
  `'REDIS_CLIENT'` token, so there is no second connection quietly sitting on DB 0.
- Both reset paths take that same DI client — `libs/test-kit/src/harness.ts:271` and
  `apps/worker/test/integration/setup.ts:76` — so `flushDb()` flushes *this worker's* logical DB. The
  `TRUNCATE`-plus-`flushDb` cross-wipe the PR is about is genuinely closed.
- `MAX_TEST_WORKERS = 15` is derived correctly: Redis ships 16 logical DBs and `JEST_WORKER_ID` is
  1-indexed, so 15 is the true ceiling and the docblock says why.
- `resolveWorkerId()` deliberately not clamping is the right call, and the comment explains it: clamping
  would hand two workers one database and silently restore the bug; an out-of-range worker instead fails
  on a database that was never created.

**The two migration-chain specs survive parallelism.** These were my main worry, because they build the
real migration chain in a *second* database:
`apps/api/test/integration/fulfillment-work-migration-parity.int-spec.ts:124` uses
`fulfillment_migration_parity` and `apps/api/test/integration/oms-connection-never-seeded.int-spec.ts:57`
uses `oms_never_seeded` — **distinct** fixed names, one per spec file, and jest never runs one file on two
workers. `CREATE DATABASE` is cluster-wide regardless of which database the issuing connection is on, so
running it from a per-worker DataSource is fine. No collision.

**The PrestaShop boot race was spotted and closed properly.** `tryClaimSharedBoot` uses `openSync(…, 'wx')`,
which is a genuine single-syscall exclusive create, with stale reclaim and a correct no-two-winners
argument (any number of reclaimers race the same exclusive create; losers fall back to polling). The
deadline sizing works out: `BOOT_LOCK_STALE_MS = INSTALL_DEADLINE_MS (12 min) + 60s = 13 min`, and the PS
specs' own `beforeAll` timeouts are 15 and 20 minutes, so a poller times out with the helper's own
explanatory message rather than with a bare hook timeout.

**`setupFiles` over `setupFilesAfterEnv`** is right and the reasoning given is right — `setup-each.ts`
imports `./setup` at module load, so scoping from there would race the import graph.

`toHandles` returning `process.env.DB_DATABASE` rather than `DEFAULT_DB_NAME` closes a divergence that
would otherwise have been very hard to debug (handles naming one database, env naming another).

---

## Findings

### 🟠 IMPORTANT — `.github/workflows/ci.yml` (the `Cache jest transform output` key)

> Dropping `${{ github.run_id }}` makes the primary key fully stable, and `actions/cache` **skips the save
> when the primary key already exists** — so the entry is written once and then frozen.

The key becomes `jest-integration-${{ runner.os }}-${{ hashFiles(...) }}` with no run-scoped suffix.
`actions/cache@v4` does not update an existing entry; it logs *"Cache already exists. Skipping save."*
So from the second run onward, that entry never changes until `pnpm-lock.yaml`, `tsconfig.base.json`, a
jest-integration config or `jest.test-workers.cjs` changes. Every spec added or edited after that point
misses forever and is never written back, and nothing in the log distinguishes that from a healthy warm
cache — hit rate decays silently.

The comment's stated reason for the change is that a run-scoped key *"could never actually be RESTORED
(only ever saved fresh), so it was pure waste"*. That overlooks the `restore-keys` block directly beneath
it: the prefix `jest-integration-<os>-<hash>-` is exactly what restores the *previous* run's entry. The
old shape is the standard rolling-cache pattern — restore newest by prefix, save a fresh entry per run —
and it is what lets the cache keep tracking the tree.

This matters more than usual here because the sibling PR #3275 measures the jest transform cache as the
single largest lever on the unit tier (33 GB of unreadable entries, ~60% of that job's time). Freezing the
integration tier's cache is the opposite direction. If the concern is the 10 GB repo-wide budget and
evicting the pnpm store cache (PR #1923's finding), the answer is a narrower save scope or a shorter-lived
suffix — not a key that can only ever be written once. At minimum, the comment should state the freeze as
the accepted trade rather than describe the previous behaviour as pure waste.

### 🟠 IMPORTANT — `apps/api/test/jest-integration.cjs`, `apps/worker/test/jest-integration.cjs`

> `diagnostics: false` silences **every** ts-jest diagnostic, including syntactic errors, at test time —
> and the sibling PR in this same series fixes the underlying cause instead.

`['ts-jest', { isolatedModules: true, diagnostics: false }]`. `isolatedModules` is the measured lever and
is not in question. `diagnostics: false` is the blunt instrument that was needed to suppress `TS5110`
(`module: CommonJS` forced by ts-jest against `moduleResolution: Node16` from `tsconfig.base.json`).

PR #3275's `jest.ts-transform.cjs` diagnoses the identical error and fixes the *cause* — merging
`moduleResolution: 'node'` into the tsconfig override — and says so explicitly: *"#3263 silenced it with
`diagnostics: false`; this fixes the mismatch instead, which is better in a small way — a syntactic error
still surfaces at test time rather than being hidden."* That is the right reading. Since both PRs are in
flight together, these two configs should consume the shared `transpileOnlyTsJest()` builder rather than
keeping the switch that hides more than it needs to.

### 🟠 IMPORTANT — `jest.test-workers.cjs` (`DEFAULT_TEST_WORKERS` sizing) vs the second PrestaShop container

> The 6-worker budget accounts for one PrestaShop container. At `maxWorkers: 6` there can be two.

The sizing note reads: *"6 leaves two of the eight CPUs for the Postgres, Redis and PrestaShop containers
that share the same cap"* — singular. But `apps/api/test/integration/prestashop/prestashop-webhook-provisioning.int-spec.ts:91`
calls `startPrestashopContainer()` directly and owns its own PS + MySQL 8.4 pair (this is the documented
exception in `docs/testing-guide.md` § "Call `startPrestashopContainer()` directly only when the spec's
subject **is** the from-scratch install"). At `maxWorkers: 1` that pair could never be up at the same time
as the shared one; at 6 it can, and on a different worker from the specs using the shared pair.

So the realistic peak is two PrestaShop + two MySQL containers plus six Node workers plus Postgres and
Redis inside a 24 GiB / 8-CPU cgroup — the exact shape the same docblock describes producing `exit=137`
and the `ECONNREFUSED` storms at 8 workers. It has not bitten (CI is green), but the reasoning that
justifies 6 does not cover it, and this is the failure class that presents as a 900 s timeout looking like
a test failure rather than as contention.

Either state the two-PS worst case in the sizing comment, or make the boot-lock cover the standalone path
too so the two pairs cannot be concurrent.

### 🟠 IMPORTANT — `docs/testing-guide.md` § PrestaShop Testcontainer Pattern

> That section's contract with spec authors was written for a serial suite and is now strictly stronger.

The guide says: *"**Nothing resets PS between specs.** Assert only over OL-side results, or filter by ids
your own spec created … Orders and carts a previous **file** left behind are inert to that shape of
assertion … but a bare 'count all orders' assertion is not."*

Under 6 workers a peer spec's rows can now appear **during** your spec, not merely before it. That
invalidates a slightly larger class of assertion than "count all orders" — anything reading a max/latest,
or asserting the absence of something, is now racy in a way it was not. The guide's § Quick Reference and
§ Troubleshooting were updated (the `Serial Execution` line and the `maxWorkers: 1` troubleshooting step
both correctly become `OL_TEST_MAX_WORKERS`), but this section was not, and it is the one carrying the
contract a spec author actually reads before writing a PS spec.

### 🔵 SUGGESTION — `jest.test-workers.cjs`, `resolveTestWorkers()`

The count is deliberately not CI-gated — *"unlike the unit tier, this only runs against Docker on a machine
that already opted in"* — so a local `pnpm test:integration` now forks 6 workers, each booting a full Nest
`AppModule`. That is a defensible line, but "has Docker" and "has 8 cores" are different facts, and the
repo's lessons already record local Docker degrading under long sessions. Consider
`Math.min(DEFAULT_TEST_WORKERS, os.cpus().length - 1)` off CI while keeping the flat 6 on it — the same
local-vs-CI split `jest.unit-workers.cjs` makes for the sibling tier, for the same reason.

### 🔵 SUGGESTION — `libs/test-kit/src/containers.ts`, `createWorkerDatabases`

The docblock asserts creation runs *"against a freshly-started server where none of these names can already
exist"*. True on both current call sites, and `ON_ERROR_STOP=1` correctly makes a failure loud. But the
function is now **exported** from `libs/test-kit/src/index.ts`, so the next caller is not bound by that
premise, and Postgres has no `CREATE DATABASE IF NOT EXISTS` to fall back on. Either guard with a
`SELECT 1 FROM pg_database WHERE datname = …` pre-check, or state the precondition in the exported
signature's contract rather than only in the prose.

---

## Summary

This does the hard part of parallelising an integration suite correctly: it gives each worker its own
Postgres database and Redis logical DB inside the *same* containers, so the win costs no runner slots, and
it checked the two places that isolation usually leaks — the reset clients and the specs that build a second
database from the real migration chain. I verified both independently and both hold. Removing the per-file
ts-jest program rebuild is the right lever and was measured against the right control (an empty spec), the
cost is paid rather than absorbed (a real `tsconfig.test-check.json` plus a step in the one *required* CI
job, so integration specs end up better covered than before), and the 6-vs-8 decision is grounded in the
container's own cgroup rather than the host's core count.

The most important concern is the cache key: dropping `github.run_id` makes the primary key stable, and
`actions/cache` refuses to overwrite an existing key — so the transform cache is written once and then
frozen, decaying silently as specs change, which is the opposite of what the sibling PR in this series
establishes as the single biggest lever. Behind that, the 6-worker budget does not account for the second
PrestaShop container that `prestashop-webhook-provisioning` owns, and the guide section spec authors read
before writing a PS spec still describes a serial world. None of these blocks the merge.

## Merge Readiness

🔄 **Approve with changes** — no BLOCKING findings. The four IMPORTANT items are follow-ups; I would take
#1 and #3 before the next time this job goes red, because both make an eventual failure hard to attribute.
(The branch is currently `BEHIND` `main` and will need an update before merge; note it also touches
`ci.yml` alongside #3275, so whichever lands second should re-read the merged file rather than trusting
a clean textual merge.)

## Priority fixes

1. **IMPORTANT** — Restore a rolling cache key (prefix + run-scoped suffix, as `restore-keys` already
   assumes) or record the permanent freeze as the accepted trade in the comment.
2. **IMPORTANT** — Replace `diagnostics: false` with #3275's `moduleResolution: 'node'` merge via the
   shared `transpileOnlyTsJest()` builder, so syntactic errors still surface at test time.
3. **IMPORTANT** — Account for the standalone PrestaShop pair from
   `prestashop-webhook-provisioning.int-spec.ts` in `jest.test-workers.cjs`'s sizing note, or serialise it
   against the shared pair.
4. **IMPORTANT** — Update `docs/testing-guide.md` § PrestaShop Testcontainer Pattern: PS specs can now run
   concurrently against one un-reset instance, so a peer's rows can appear mid-spec.
5. **SUGGESTION** — Bound the local worker count by `os.cpus()`; document or guard the now-exported
   `createWorkerDatabases` precondition.

norbert-kulus-blockydevs:
Addressed all findings from the final (APPROVED) review (9547ab0):

**IMPORTANT — cache key freeze**: restored a run-scoped primary key (`...-${{ github.run_id }}`) so `actions/cache@v4` saves a fresh entry every run instead of freezing on the first save; `restore-keys`' hash prefix still restores the previous run's entry.

**IMPORTANT — `diagnostics: false`**: both integration jest configs now use the shared `transpileOnlyTsJest()` builder (`jest.ts-transform.cjs`, landed via #3275) instead of blanket-silencing ts-jest diagnostics — it fixes the underlying `moduleResolution` mismatch, so a real syntactic error still surfaces at test time.

**IMPORTANT — standalone PrestaShop pair**: `jest.test-workers.cjs`'s sizing note now documents that `prestashop-webhook-provisioning.int-spec.ts` boots its own PS+MySQL pair alongside the shared one, so the realistic peak under 6 workers is two pairs, not one — with the remedies named rather than left implicit.

**IMPORTANT — `docs/testing-guide.md`**: § PrestaShop Testcontainer Pattern now states that PS specs can run concurrently and widens the "don't assert this" guidance beyond "count all orders" to cover a max/latest read or an absence check.

**SUGGESTION — local worker bound + `createWorkerDatabases` guard**: the off-CI default is now `os.cpus().length - 1`-bounded (mirroring `jest.unit-workers.cjs`'s local/CI split), and `createWorkerDatabases` pre-checks `pg_database` and only creates missing databases, so the now-exported function is safe against a server that already has some of them.

Also rebased onto `main` (the branch was 250+ commits behind) and resolved the one conflict in `docs/lessons.md` — both #3271's and this PR's new entries are kept.

Issue #3086, opened 2026-09-10
test(returns): orphan/manual returns flow tests


norbert-kulus-blockydevs:
## Problem / Context

No test coverage exists for any of the above.

## Proposed Solution

Add colocated tests for match, authorize, and record flows including their error branches.

## Classification

**Type**: Frontend
**Layer**: Interface (tests)
**File(s)**: apps/web/src/features/returns/**/*.test.{ts,tsx}

## Dependencies

- Part of #3078

## Assumptions

- None

## Acceptance Criteria

- [ ] pnpm test passes
- [ ] Tests added or updated for non-trivial logic
- [ ] No architecture boundary violations (CORE ↔ Integration)

Issue #3085, opened 2026-09-10
feat(returns): entry point from the returns list/detail


norbert-kulus-blockydevs:
## Problem / Context

The worklist from a prior sub-issue has no reachable entry point in the product.

## Proposed Solution

Wire a segment/filter entry (or nav badge) from `apps/web/src/app/routes/returns.route.tsx` / the returns list page into the orphan-returns worklist.

## Classification

**Type**: Frontend
**Layer**: Interface (routing)
**File(s)**: apps/web/src/app/routes/returns.route.tsx

## Dependencies

- Part of #3078

## Assumptions

- None

## Acceptance Criteria

- [ ] Worklist is reachable from the existing Returns nav item with no direct URL
- [ ] Tests added or updated for non-trivial logic
- [ ] No architecture boundary violations (CORE ↔ Integration)

Issue #3083, opened 2026-09-10
feat(returns): approve dialog for operator-authored returns


norbert-kulus-blockydevs:
## Problem / Context

`POST /returns/:returnId/authorize` is restricted to `origin: 'operator_authored'` and refused with 409 `reason: source-ingested` otherwise — no confirmation UI exists.

## Proposed Solution

Build a simple confirm dialog; a source-ingested return should never reach this action (gated at the worklist level).

## Classification

**Type**: Frontend
**Layer**: Interface (component)
**File(s)**: apps/web/src/features/returns/components/authorize-return-dialog.tsx

## Dependencies

- Part of #3078

## Assumptions

- None

## Acceptance Criteria

- [ ] Confirming calls authorize and removes the row from the pending-approval group
- [ ] Tests added or updated for non-trivial logic
- [ ] No architecture boundary violations (CORE ↔ Integration)

PR #3278, opened 2026-09-16
feat(returns): write API client (record, authorize, match-order)

### LINKED ISSUE #3079: feat(returns): write API client (record, authorize, match-order)
## Problem / Context

`features/returns/api/returns.api.ts` has no `record`, `authorize` or `matchOrder` method — verified absent by grep during the mockup pass.

## Proposed Solution

Add the 3 methods to `ReturnsApi`, typed against `RecordReturnDto`/`RecordReturnResponseDto`, `AuthorizeReturnResponseDto`, and `MatchReturnToOrderDto`/`MatchReturnToOrderResponseDto` from `apps/api/src/returns/dto/return-write.dto.ts`.

## Classification

**Type**: Frontend
**File(s)**: apps/web/src/features/returns/api/returns.api.ts

## Dependencies

- Part of #3078

## Assumptions

- None

## Acceptance Criteria

- [ ] All 3 methods exist and are typed against the DTOs

jakubretajczykBD:
## Summary
- Adds the three orphan/manual-return writes to the returns API client — `authorize`, `matchOrder`, `record` (#2372/#2376) — each parsing its response through Zod rather than casting, so a contract break surfaces as a named `ReturnWriteResultUnreadableError` instead of an `undefined` rendered into a field an operator then trusts
- `returns.types.ts` gains `AuthorizeReturnResult`, `MatchReturnToOrderInput/Result`, `RecordReturnInput/Result`, `RecordReturnLineInput`
- `return-write.schema.ts`: parsers + the error class
- Barrel exports the new result/input types and the error class alongside the existing `Return detail` section's convention

## Test plan
- [x] `tsc --noEmit` clean
- [x] `eslint` clean
- [x] `vitest run src/features/returns` — 247/247 passing
- [x] Field names verified 1:1 against the real backend `ReturnWritesController`

Closes #3079


piotrswierzy (approved):

Reviewed as the base of the four-PR returns-write stack (#3278 → #3279 → #3280 → #3281).

### Verified against the backend rather than the description

I checked the three route shapes and the refusal contract against the shipped controller and filter, not against the docblocks:

- **`POST /returns/record` really is declared ahead of every `:returnId` route** — `return-writes.controller.ts:124` vs `:166`/`:196`. The client's comment asserting that ordering is correct, and it matters: a literal segment registered after a parameterised one would be swallowed.
- **The 400/409 split on `ReturnMatchRefusedError` is real and the client documents it accurately.** `returns-exception.filter.ts:50-58` splits by reason — `already-attributed` → 409 (the return's own state), `unknown-order` → 400 (the operator's id names nothing) — and maps `ReturnRecordRefusedError` wholly to 400 even though it carries an `unknown-order` reason too, because `POST /returns/record` addresses no resource. `returns.api.ts`'s per-method docblocks state both halves correctly.
- **`authorize`'s `already-authorized` is a success, not a refusal**, and the client says so rather than shaping it as an error path.

### What is well done

**Parse, never cast**, on all three writes, with `ReturnWriteResultUnreadableError` distinguishing "the write may have landed, we just could not read the result" from an ordinary failure — and it is barrel-exported for exactly that reason. That distinction is the one a caller cannot reconstruct later, so publishing it is right.

**`outcome` / `origin` read as `z.string()` rather than `z.enum`.** The reasoning holds: the write *happened*, and refusing to read its result because the vocabulary grew would report a failure for a change that landed. Same rule `return-custody.schema.ts` already applies to `custodyState`.

**Optional line fields are omitted, not sent as `null`/`""`** (`returns.api.ts`, the `record` body), matching `noteBody`'s rule — a typed-but-empty value would persist text the operator did not write.

### Findings

**[IMPORTANT]** — `apps/web/src/features/returns/api/return-write.schema.ts`
> No test file, and `returns.api.test.ts` has zero coverage of `record` / `authorize` / `matchOrder` (grepped — no hits). There is precedent for the omission (`return-custody.schema.ts` has no test either), so this is not a blocker, but this file introduces a **new named failure path** — `ReturnWriteResultUnreadableError` — that nothing proves ever fires. The three `safeParse` guards are the whole reason the file exists; a single test per parser asserting a malformed envelope throws (and a valid one round-trips `nullish` → `null`) would pin the behaviour the docblock promises. `return-detail.schema.test.ts` and `return-timeline.schema.test.ts` are the shape to copy.

**[SUGGESTION]** — `docs/plans/mockups/orphan-returns-3078.html`
> The mockup satisfies `docs/frontend-architecture.md` § UX Mockups in substance — there is a real in-page state switcher with eleven stable, named states, and the two match refusals are demonstrated as *separate* states (`match-error` = unknown order 400, `match-conflict` = already linked 409), which is exactly the distinction #3281 has to render. It uses `data-chip` as the state attribute rather than `data-state`. The convention allows "`data-state` (or equivalent)", so this conforms, but `data-chip` reads as a styling hook rather than a state handle, and the convention's whole point is that E2E specs reference these by name. Worth renaming to `data-state` while the mockup is new and nothing cites it yet. Token discipline is otherwise clean: 215 `var(--…)` usages and the only literal hex is the `--bg-surface` declaration inside `:root`.

**[SUGGESTION]** — nothing in the app imports this client's new methods yet
> Expected for the base of a stack, but worth stating: until #3085 mounts the surfaces, the only exercise these three methods get is whatever #3281's dialog tests drive through the mutation layer. That raises the value of the schema tests above rather than lowering it.

### Summary

The typed client for the three orphan/manual-return writes, plus the epic's committed mockup. It binds accurately to a backend contract I verified independently — route ordering, the reason-split refusals, and the idempotent-authorize semantics are all stated correctly rather than approximately. Quality is high; the one real gap is that a new named failure path ships untested.

### Merge readiness

✅ **Approve.** All checks green (the `CANCELLED` Docker Build Smoke Test is the path-filter, not a failure). The schema test is worth adding — either here or as the first commit of a follow-up — but it does not block a client whose consumers land in the next three PRs.

piotrswierzy (approved):
## Tech Lead review — #3278 `feat(returns): write API client (record, authorize, match-order)`

Reviewed against `docs/code-review-guide.md` (4-step), `docs/engineering-standards.md`, `docs/frontend-architecture.md`.

### Context established

Baseline read on `main`: `apps/web/src/features/returns/api/return-custody.schema.ts` (the parse-never-cast precedent this PR cites), `returns.api.ts`, `returns.types.ts`, and the backend contract at `apps/api/src/returns/dto/return-write.dto.ts` + `apps/api/src/returns/http/return-writes.controller.ts`.

**The contract claim in the PR body checks out.** Every field the three parsers read exists 1:1 on the backend DTOs — `AuthorizeReturnResponseDto` (`outcome` / `changeId` / `authorizedAt`), `MatchReturnToOrderResponseDto` (`returnId` / `internalOrderId` / `matchedAt`), `RecordReturnResponseDto` (`returnId` / `internalOrderId` / `origin` / `openedAt`) — and the three route paths match `return-writes.controller.ts:124` (`@Post('record')`), `:166` (`@Post(':returnId/authorize')`), `:196` (`@Post(':returnId/match-order')`), including the literal `record` segment being declared ahead of the `:returnId` routes.

---

### Findings

**[IMPORTANT]** — `apps/web/src/features/returns/api/return-write.schema.ts` (whole file)

No test ships for `parseAuthorizeReturnResult` / `parseMatchReturnToOrderResult` / `parseRecordReturnResult`, and none for `ReturnWriteResultUnreadableError`. `docs/engineering-standards.md § Testing Standards` ("Unit test exists for non-trivial logic") and #3079's own AC ("Tests added or updated for non-trivial logic") both ask for one, and the dominant in-feature precedent is to have it: three of the five schema modules on `main` carry a colocated spec — `returns.schema.test.ts`, `return-detail.schema.test.ts`, `return-timeline.schema.test.ts`. `return-custody.schema.ts` is the single untested outlier, so it is a precedent rather than a rule.

The behaviour that most needs pinning is the one the PR body sells: `ReturnWriteResultUnreadableError` is what lets a caller tell *"the write may have landed, we just could not read the result"* apart from an ordinary failure — and `apps/web/src/features/returns/index.ts:103` exports it publicly on exactly that reasoning. Nothing asserts that a malformed body throws it rather than yielding an object with `undefined` fields, and nothing asserts the `nullish → null` normalisation in `orNull` (an omitted `matchedAt` must read `null`, not `undefined`, or a consumer's `=== null` check silently stops firing). `returns.api.test.ts` covers `list` and `getIngestionAvailability` only, so there is no api-level coverage either.

**[SUGGESTION]** — `apps/web/src/features/returns/api/returns.types.ts` (+616, +682)

`outcome: AuthorizeReturnOutcome | string` and `origin: ReturnOrigin | string` collapse to `string` in TypeScript, so the declared unions give a consumer no narrowing at all — `AUTHORIZE_RETURN_OUTCOME_VALUES` is the only usable handle. The open-world `X | string` shape is established repo practice (`CoreCapability | string`, `docs/architecture-overview.md § Capability Abstractions`), so this is not wrong; but this same file already ships the better ending for it at `:36` — `isReturnBucket`, a coercion guard beside the values array. A sibling `isAuthorizeReturnOutcome` would let #3080/#3083 branch on `already-authorized` (a SUCCESS, per the PR's own docblock, which a caller must not render as an error) without re-deriving the membership test at each call site.

**[SUGGESTION]** — `docs/plans/mockups/orphan-returns-3078.html` (new, ~1,100 lines)

The mockup is an epic-level artifact (#3078) riding in the PR whose `Closes` line is #3079, the API client. It is in the right directory and harmless, but it means the stack's first merge carries a doc artifact unrelated to its own scope, and a reviewer of #3079 has ~90% of the diff to page past before reaching the code. Landing it on its own, or on the epic's mounting PR, keeps each PR's diff about its own issue.

### Positive observations

- `record`'s body omits absent optionals rather than sending `null`/`""` (`returns.api.ts`, `...(line.sku ? { sku: line.sku } : {})`), with the reasoning stated inline and consistent with the existing `noteBody` helper two functions up — a typed-but-empty `name` would otherwise persist text the operator did not write.
- Parse-never-cast is applied uniformly, and the module docblock explains *why* `outcome`/`origin` are read as `z.string()` rather than `z.enum` (a value this build predates must still round-trip, because the write already happened) — the same reasoning `return-custody.schema.ts` gives for `custodyState`.
- `encodeURIComponent` on both path-segment ids.
- The barrel addition is typed exports plus one error class, in a labelled section, and the forward-reference comment for #3080 is removed in that PR rather than left to rot.

---

## Summary

A faithful, well-documented extension of the returns API client: three writes, three parsers, field names verified against the real controller and DTOs, and the feature's parse-never-cast rule applied consistently. Architecture and layering are clean — this is the interface layer talking to the api client seam, no raw `fetch()`, no cross-feature import, no `any`. The one substantive gap is that the new parsers and the error class they exist to raise ship untested, against a feature where three of five sibling schema modules have specs and the issue AC asks for them.

## Merge Readiness

🔄 **Approve with changes** — merge after adding parser coverage.

## Priority fixes

1. **[IMPORTANT]** Add `return-write.schema.test.ts` covering, per parser: a well-formed body round-trips; a malformed body throws `ReturnWriteResultUnreadableError`; an omitted nullish field reads `null` rather than `undefined`.
2. **[SUGGESTION]** Add an `isAuthorizeReturnOutcome` guard beside `AUTHORIZE_RETURN_OUTCOME_VALUES`, matching `isReturnBucket` at `returns.types.ts:36`.
3. **[SUGGESTION]** Consider landing the #3078 mockup separately from the #3079 client.

piotrswierzy (approved):
## Re-review — #3278 (write API client)

Re-read against current `main` (`e1ae0aec`), which has moved since the first pass.

### 1. Prior findings

|---|---|---|---|
| 1 (IMPORTANT) | The three Zod parsers and `ReturnWriteResultUnreadableError` shipped with zero tests | **Addressed** | `apps/web/src/features/returns/api/return-write.schema.test.ts:1-110` — 11 cases across all three parsers. Each covers the four states that matter: round-trip, omitted-nullish → `null`, explicit-`null` → `null`, and malformed → the named error (`:44`, `:74`, `:105`). The error assertion is on the class, not a message string, so a reword cannot silently green it. |

Two prior SUGGESTIONs were advisory and carry no verification obligation.

**Re-verified against the moved base** (the first pass's checks, redone rather than assumed):
- Field names are still 1:1 with `apps/api/src/returns/dto/return-write.dto.ts:47-150` — `changeId`/`authorizedAt` (`:55`,`:58`), `returnId`/`internalOrderId`/`matchedAt` (`:75`,`:83`,`:86`), `returnId`/`internalOrderId`/`origin`/`openedAt` (`:143-150`).
- Routes still match `apps/api/src/returns/http/return-writes.controller.ts` — `@Post('record')` `:124`, `@Post(':returnId/authorize')` `:166`, `@Post(':returnId/match-order')` `:196`.
- `.nullish()` throughout (`return-write.schema.ts:39-41`, `:55-57`, `:71-74`), per `docs/architecture-overview.md` on OL serialising absent optional fields as JSON `null`.

### 2. New commits

`c47e3118` adds the test file and `isAuthorizeReturnOutcome`. No production behaviour changed.

**SUGGESTION — `isAuthorizeReturnOutcome` is untested.** It is named in the commit headline alongside the schemas and is the membership test every call site will branch `already-authorized` on (`returns.types.ts`, the guard just above `AuthorizeReturnResult`), but `return-write.schema.test.ts` does not exercise it and no sibling spec does. Three lines — a member, a non-member, `null`/`undefined` — would close it. Not blocking: the guard is a plain `includes` over an `as const` array and the consuming branch lands in #3083.

No raw `fetch()` (`returns.api.ts:424-458` all route through the injected `request`), parse-never-cast holds on every response, `record` omits absent optionals rather than sending `null`/`""` (`:446-450`) consistent with `noteBody`'s rule.

| | Count |
|---|---|
| BLOCKING | 0 |
| IMPORTANT | 0 |
| SUGGESTION | 1 |

**Merge readiness: ready.** The one prior IMPORTANT is genuinely closed in code, and the contract re-check against the moved base still holds.

**Priority fixes:** none. The guard test is worth a line in a follow-up, not a round trip.

piotrswierzy (approved):
## Tech Lead review — `feat(returns): write API client (record, authorize, match-order)`

Bottom of the #3079–#3086 stack. Transport layer only, no consumer, which is the right shape for this slice.

### ✅ Verified line-by-line against the backend

I read `apps/api/src/returns/http/return-writes.controller.ts` and its DTOs on `main` rather than taking the docblocks on trust. Every claim in this client checks out:

| Client | Controller | ✓ |
|---|---|---|
| `POST /returns/:id/authorize` → `{outcome, changeId, authorizedAt}` | `@Post(':returnId/authorize')`, same three fields, `authorizedAt` ISO-or-null | ✓ |
| `POST /returns/:id/match-order` + `{internalOrderId}` → `{returnId, internalOrderId, matchedAt}` | `@Post(':returnId/match-order')`, `MatchReturnToOrderDto`, same three | ✓ |
| `POST /returns/record` → `{returnId, internalOrderId, origin, openedAt}` | `@Post('record')`, same four | ✓ |

And the subtler claims hold too:

- **"Declared as a literal `/returns/record` segment ahead of every `:returnId` route."** True — `@Post('record')` is at line 124, every `:returnId` route below it. Worth having written down, because the ordering is load-bearing in Nest and invisible from the client side.
- **`already-authorized` is a SUCCESS, not a refusal.** Correctly modelled: it's a value of `outcome` on a 201, not a 409, and `AUTHORIZE_RETURN_OUTCOME_VALUES` + `isAuthorizeReturnOutcome` give the consumer (#3083) one membership test instead of a string compare at each call site.
- **Attribution is monotonic.** The `matchOrder` docblock says "there is no unmatch… the caller must confirm before calling this," which is exactly what `docs/architecture-overview.md` §22 requires of this route (`claimAttribution` can fill the value in and never change one). Putting that on the *client method* is the right place — it's a constraint on callers, and #3082 is about to be one.
- **The 400/409 split is reproduced faithfully** rather than flattened: `unknown-order` is a payload fault (400), `already-attributed` is resource state (409). That's the controller's own stated rule.

### ✅ Parse, never cast

`return-write.schema.ts` follows the `return-custody.schema.ts` precedent exactly, and the two deviations from naive Zod are both reasoned in place and both right:

- **`outcome` / `origin` as `z.string()`, not `z.enum`.** A build that predates a value must still round-trip, because *the write already happened* — refusing to read the result would report failure for a change that landed. That's the same reasoning the repo applies to `ReturnLineEvent.kind` and `custodyState`.
- **`ReturnWriteResultUnreadableError` as a named, exported type**, so a caller can tell "the write may have landed, we just can't read the answer" from an ordinary transport failure. This is the one that matters for a non-idempotent write like `record`; exporting it from the barrel alongside `ReturnDetailUnreadableError` is consistent.

`orNull` collapses `undefined` and `null` to one representation at the boundary — fine here, since neither the DTOs nor the types give the two a distinct meaning on these three responses.

### ✅ Request-body construction

The `record` mapper omits `sku` / `name` / `note` when falsy rather than sending `null` or `""`, with the reason stated:

> *an absent optional field is the server's own "not supplied"… a typed-but-empty value would persist text the operator did not write*

That matches the controller, which does `line.sku ?? null` — so an omitted field and an explicit `null` converge, and an empty string would NOT have. Reusing the existing `noteBody()` helper rather than open-coding a fourth variant of the same rule is right.

`encodeURIComponent` on both `:returnId` paths.

### ✅ Mockup

`docs/plans/mockups/orphan-returns-3078.html` is committed to the repo, not left as an Artifact link, per `docs/frontend-architecture.md` § UX Mockups. It uses real `index.css` tokens and transcribes real component CSS, and it ships the required in-page state switcher with eleven stably-named states (`list`, `match-open`, `match-error`, `match-conflict`, `authorize-refused`, `record-open`, …) that the downstream PRs can cite by name. The `data-chip` attribute is the "or equivalent" the convention allows. The comment explaining why the dialog backdrop carries 148px of bottom padding — so the fixed QA toolbar can't swallow a dialog's action row — is the kind of thing that saves the next person twenty minutes.

### Notes

- CI green across all nine checks.
- No consumer yet, and the barrel comment says so explicitly (*"The three mutation hooks over these writes are #3080 — added to this section on that branch, not here"*). Good stack hygiene: the reader of the barrel isn't left wondering where the hooks went.
- Test file covers each parser's happy path and its unreadable-response path.

**Verdict:** ✅ Approve. Nothing to fix.

PR #3279, opened 2026-09-16
feat(returns): mutation hooks + cache invalidation

### LINKED ISSUE #3080: feat(returns): mutation hooks + cache invalidation
## Problem / Context

No hooks exist for the 3 new write methods.

## Proposed Solution

Add `use-record-return-mutation.ts`, `use-authorize-return-mutation.ts`, `use-match-return-to-order-mutation.ts` in `features/returns/hooks/`, invalidating the returns list and the individual return's query key on success.

## Classification

**Type**: Frontend
**Layer**: Interface (hooks)
**File(s)**: apps/web/src/features/returns/hooks/*

## Dependencies

- Part of #3078

## Assumptions

- None

## Acceptance Criteria

- [ ] Each mutation invalidates the relevant query keys on success

jakubretajczykBD:
## Summary
- Adds the three TanStack Query mutation hooks over the #3079 orphan/manual-return writes
- `useAuthorizeReturnMutation` — `onSettled`: a refused source-ingested attempt still needs a re-read to state what the record now says
- `useMatchReturnToOrderMutation` — `onSettled`: a lost race against a concurrent matcher reports `already-attributed` (409) with the return already attributed by then, so a re-read surfaces the winning order rather than leaving the page showing the pre-attempt orphan state
- `useRecordReturnMutation` — `onSuccess`: a 400 refusal creates nothing, so there is nothing new to invalidate on the failure path; the newly-created return's own detail is invalidated from the result's `returnId`
- Barrel exports the three hooks alongside the #3079 types/error class

## Test plan
- [x] `tsc --noEmit` clean
- [x] `eslint` clean
- [x] `vitest run src/features/returns` — 247/247 passing

Closes #3080


piotrswierzy (approved):

Second of the returns-write stack, on top of #3278.

### The invalidation question, answered

The directive worth checking here is whether a successful match leaves a stale worklist. It does not, and the reason is structural rather than incidental: `returnsQueryKeys.all` is `['returns']`, which is the **prefix of every other key** in that file — `list`, `detail`, `correctionProposal`, `orderEvents`, `returnEvents`. TanStack Query's `invalidateQueries` matches by prefix, so invalidating `all` already subsumes the list #3280 renders and the detail #3281 routes to. Nothing goes stale.

### The `onSuccess` / `onSettled` split is argued per hook, and each argument holds

- **`match` → `onSettled`.** Correct, and the reasoning is the sharp one: a lost race reports `already-attributed` (409) and the return **is** attributed by the time the caller sees the error, so a success-only invalidation would leave the page showing the pre-attempt orphan state after a write that effectively succeeded for someone else. This is the same read #3281 depends on to render its conflict panel honestly.
- **`authorize` → `onSettled`**, matching `useDeclineReturnMutation` verbatim.
- **`record` → `onSuccess`.** Also correct, for the opposite reason: a refused `record` (400) creates nothing, so there is nothing new to invalidate on the failure path. Invalidating the newly-created return's own `detail` from the result's `returnId` is a real consideration — a caller navigating straight there must not hit a cold cache.

Awaiting both invalidations sequentially keeps `isPending` true until the refetches settle, which is what lets #3281's dialog close *after* the list is current rather than before. Worth knowing that this is load-bearing, not incidental.

### Findings

**[SUGGESTION]** — all three hooks
> `invalidateQueries({ queryKey: returnsQueryKeys.detail(id) })` followed by `{ queryKey: returnsQueryKeys.all }` is redundant — the second call's prefix match already covers the first. Harmless (one extra pass over the cache), and it **exactly matches the shipped `useDeclineReturnMutation`**, so I would not change it here in isolation: consistency across four sibling hooks is worth more than removing one no-op line, and a future reader comparing them should not find one hook mysteriously different. If it is ever cleaned up, clean all four in one pass.

**[SUGGESTION]** — `use-record-return-mutation.ts`
> The broad `all` invalidation refetches every active returns query, including `correctionProposal` and `orderEvents` for unrelated returns. That is the right trade for a write this rare, but if the returns surfaces ever grow a heavy always-mounted query, this is the line that will make it visible.

**[IMPORTANT]** — CI signal
> `Test` and `Integration Tests` are still **pending**, not green, at review time. Lint, Type Check and Build have passed. The three hooks carry no tests of their own, which is consistent with the existing hook layer in this feature (no `*.test.ts` under `features/returns/hooks/`), and the behaviour is exercised indirectly by #3281's dialog tests — so I am not asking for hook tests, only noting that the remaining signal has not landed.

### Summary

Three thin mutation hooks over #3278's client, each choosing `onSuccess` vs `onSettled` deliberately and correctly. The invalidation genuinely covers the surfaces the next two PRs render, because `all` is a prefix of every key rather than a sibling of them. Nothing here needs changing.

### Merge readiness

✅ **Approve**, once `Test` and `Integration Tests` report green — nothing in the diff gives me reason to expect otherwise.

piotrswierzy (approved):
## Tech Lead review — #3279 `feat(returns): mutation hooks + cache invalidation`

Reviewed against `docs/code-review-guide.md` (4-step), `docs/frontend-architecture.md § State ownership`, `docs/engineering-standards.md`.

### Context established

Baseline read on `main`: `apps/web/src/features/returns/hooks/use-decline-return-mutation.ts`, `use-return-custody-mutations.ts`, and `api/returns.query-keys.ts`. Both existing mutation modules use the same shape this PR adopts — `useApiClient()` + `useQueryClient()`, `invalidateQueries` on `detail(returnId)` then the `all` prefix, invalidate-never-seed, with the `onSettled`-vs-`onSuccess` choice argued in the docblock rather than assumed.

Layering is correct: server state via TanStack Query (`docs/frontend-architecture.md:229`), no raw `fetch()`, the api client reached only through `app/api/api-client-provider` exactly as its siblings do, `UseMutationResult` explicitly typed on every export, no `any`.

The per-hook reasoning is sound and worth saying so explicitly:
- `useAuthorizeReturnMutation` — `onSettled`. Correct: a 409 `source-ingested` refusal is answered from persisted state the page may not be showing, so the honest move is to re-read either way.
- `useMatchReturnToOrderMutation` — `onSettled`. Correct and load-bearing: a lost race reports `already-attributed` (409) *with the return already attributed*, so only a re-read surfaces the winning order. #3281 depends on this exact behaviour.
- `useRecordReturnMutation` — `onSuccess`. Correct: `POST /returns/record` addresses no existing return, and `return-writes.controller.ts` refuses it with a 400 that creates nothing, so there is genuinely nothing new to invalidate on the failure path. Invalidating `detail(result.returnId)` from the *result* is the only way it could be done at all.

---

### Findings

**[SUGGESTION]** — the three new files share one invalidation body

`use-return-custody-mutations.ts`'s own docblock on `main` states: *"All three live in one module because they share that invalidation contract exactly; splitting them into three files with the same body is how one of them eventually forgets a key."* This PR ships three files whose `onSettled`/`onSuccess` bodies are the same two `invalidateQueries` calls, and it does so without acknowledging that comment.

The split is defensible here — two of the three genuinely differ (`onSuccess` vs `onSettled`; `record` has no ambient `returnId`), which is more variance than the custody trio had. But the key *set* is still identical in all three, and that is the part the warning is about. Extracting the `useCustodyInvalidation`-shaped helper (or reusing it) would keep one place to edit when a fourth key joins `returnsQueryKeys`; at minimum, a line in each docblock naming the intentional divergence from that comment would stop the next reader re-litigating it.

**[SUGGESTION]** — redundant `detail` invalidation

`returnsQueryKeys.all` is `['returns']` and `returnsQueryKeys.detail(id)` is `['returns', 'detail', id]`, so TanStack's prefix matching means the `all` invalidation already covers the detail key. All three hooks invalidate both. This is faithfully carried from `use-decline-return-mutation.ts` and `use-return-custody-mutations.ts`, so it is precedent-consistent rather than a new mistake, and the explicitness has documentary value — but it is two awaited round-trips through the query cache where one would do, and a reader can reasonably conclude the two calls target disjoint sets. One inline comment ("`all` is a prefix of `detail`; the explicit call documents intent") would settle it for the whole family.

### Positive observations

- The docblocks do the thing this codebase actually values: each states *why* its lifecycle hook was chosen, with the failure it prevents named. `use-match-return-to-order-mutation.ts`'s "a lost race reports `already-attributed` and the return IS attributed by the time the caller sees the error" is exactly the reasoning #3281 then relies on.
- `useRecordReturnMutation` returns no `returnId` parameter and takes the input as the mutation variable — correct, because there is no return to scope to until the server answers.
- The barrel comment placeholder left by #3278 is replaced rather than left behind.

---

## Summary

Three thin, correctly-typed TanStack mutation hooks over the #3079 writes, each with a deliberately-chosen and well-argued invalidation lifecycle. They follow the two existing returns mutation modules closely on every dimension that matters — state ownership, api-client access, invalidate-never-seed, explicit result typing — and the one place they diverge (three files instead of one module) is defensible, if unacknowledged. Nothing here is a correctness or architecture concern; the notes below are about keeping the invalidation key set in one place as the family grows.

## Merge Readiness

✅ **Approve**

## Priority fixes

None required. Optional, in order:

1. **[SUGGESTION]** Share the invalidation body (or note in each docblock why the custody module's "don't split these" comment does not apply here).
2. **[SUGGESTION]** One comment on the `all`-is-a-prefix-of-`detail` redundancy, for the whole hook family rather than just these three.

piotrswierzy (approved):
## Re-review — #3279 (mutation hooks + cache invalidation)

Re-read against current `main` (`e1ae0aec`).

### 1. Prior findings

|---|---|---|---|
| — (note) | Three files repeat one invalidation body, which `use-return-custody-mutations.ts`'s own docblock warns against | **Not addressed** (advisory, carried) | `use-authorize-return-mutation.ts:27-30`, `use-match-return-to-order-mutation.ts:28-31`, `use-record-return-mutation.ts:33-36` — identical two-line body in all three. |

No BLOCKING or IMPORTANT was raised on the first pass, so nothing was obliged to change.

### 2. New commits

**None.** `fa1fcf3d` is the same content re-committed under a rebase onto the moved base; the diff is byte-identical to what was approved. Re-verified rather than assumed: the three hooks, and the `index.ts` hunk moving the three exports out of the "#3080 adds these on that branch" placeholder.

Re-checked on the moved base:
- `returnsQueryKeys.detail` / `.all` still exist and still have the shapes these hooks pass (`returns.query-keys.ts:9,16`).
- `onSettled` vs `onSuccess` split is still right: `matchOrder`'s lost race returns 409 with the return *already* attributed, so a re-read on the error path is the honest behaviour (`use-match-return-to-order-mutation.ts:28`); `record` creates nothing on a 400, so `onSuccess` is correct and the new `returnId` is invalidated from the result rather than guessed (`use-record-return-mutation.ts:34`).

**SUGGESTION — the `detail` invalidate is subsumed by the `all` invalidate.** `returnsQueryKeys.all` is `['returns']` (`returns.query-keys.ts:9`) and `detail(id)` is `['returns','detail',id]`, so TanStack's prefix matching already invalidates the detail key. The explicit first call is harmless and arguably documents intent, but it is two awaits where one suffices, in three files. If the shared helper the note above suggests ever lands, collapse it there.

| | Count |
|---|---|
| BLOCKING | 0 |
| IMPORTANT | 0 |
| SUGGESTION | 2 |

**Merge readiness: ready.** Nothing was outstanding and nothing changed.

**Priority fixes:** none.

piotrswierzy (approved):
## Tech Lead review — `feat(returns): mutation hooks + cache invalidation`

Three hooks, one barrel edit. Follows `useDeclineReturnMutation` as the house pattern, and the one place it deviates is deliberate and correctly argued.

### ✅ `onSettled` vs `onSuccess` is the right call in each case, and for the right reason

This is the only real decision in the PR and it's made three times, differently:

- **`authorize` → `onSettled`.** A `source-ingested` refusal touches nothing, but as the docblock says, the caller *can't know that from the shape of the error alone*. Re-reading is the honest way to state what the record now says. Same reasoning as the existing decline hook.
- **`matchOrder` → `onSettled`**, and this is the one that genuinely matters: a lost race answers 409 `already-attributed`, and the return **is** attributed by the time the caller sees that error. `onSuccess` would leave the page rendering the pre-attempt orphan state after a write that (someone else's) did land. Good catch.
- **`record` → `onSuccess`.** Every refusal on this route is a 400 that creates nothing, so there is nothing new to invalidate on the failure path. Correct.

Getting `matchOrder` wrong here would have been invisible in testing and wrong exactly when two operators work the orphan bucket at once — which is the realistic case.

### 🟢 SUGGESTION — `detail(returnId)` invalidation is redundant under `all`

All three hooks (and the pre-existing `useDeclineReturnMutation`) do:

```ts
await queryClient.invalidateQueries({ queryKey: returnsQueryKeys.detail(returnId) });
await queryClient.invalidateQueries({ queryKey: returnsQueryKeys.all });
```

`returnsQueryKeys.all` is `['returns']` and `detail(id)` is `['returns', 'detail', id]`. TanStack Query matches `queryKey` by **prefix**, so the second call already invalidates the first's target — and everything else in the namespace. Harmless, and I'd normally leave a house pattern alone, but it's worth knowing the first line is doing nothing so nobody later "fixes" a bug by adding a third.

Not asking you to change it in this PR: consistency with the existing hook is worth more than the redundant call costs. If it's ever tidied, tidy all four together.

### 🟢 SUGGESTION — the `record` docblock claims a benefit invalidation can't deliver

`use-record-return-mutation.ts`:

> *The newly-created return's own detail is invalidated too, from the result's `returnId` — a caller that navigates straight to it after a successful record **must not hit a cold, unpopulated cache**.*

`invalidateQueries` only marks **existing** cache entries stale; it cannot populate one. For a return that was created milliseconds ago there is no entry under `detail(result.returnId)`, so that call is a no-op, and a caller navigating to the new detail page gets a cold fetch on mount either way — which is fine and is what should happen.

The line is harmless; the *reason* attached to it isn't true, and that's the part worth fixing, since the next person may preserve the call for a guarantee it never provided. If you want the stated behaviour, it's `prefetchQuery` / `setQueryData(detail(result.returnId), …)` — but the result shape isn't a `ReturnDetail`, so I wouldn't: just reword the comment to say the list is what goes stale.

### Notes

- Barrel updated in place and the `#3079` placeholder comment removed rather than left behind — the section header now reads `#3079/#3080`. Clean stack hygiene.
- `useMutation<…, Error, void>` on `authorize` is the right typing: the write genuinely carries no input beyond the id, which is closed over.
- CI green (an earlier cancelled run is superseded by a fully green one).
- No tests in this PR. Acceptable — these are three thin `useMutation` wrappers with no branching, and #3086 covers the flows end to end. Worth confirming that's the intent rather than an omission.

**Verdict:** ✅ Approve. Both suggestions are comment-level.

Issue #3082, opened 2026-09-10
feat(returns): match-to-order dialog + unknown-order/already-attributed handling


norbert-kulus-blockydevs:
## Problem / Context

Matching is monotonic (no unmatch) and can fail with a 400 (`unknown-order`) or a 409 (`already-attributed`) — no UI exists to attempt or explain either.

## Proposed Solution

Build a single-field dialog with an explicit "this can't be undone" warning, rendering the 400/409 as inline field errors rather than a generic failure toast.

## Classification

**Type**: Frontend
**Layer**: Interface (component + form)
**File(s)**: apps/web/src/features/returns/components/match-return-dialog.tsx

## Dependencies

- Part of #3078

## Assumptions

- None

## Acceptance Criteria

- [ ] Unknown-order (400) renders as a field error naming the typed id
- [ ] Already-attributed (409) renders as a distinct message, not a generic error
- [ ] Tests added or updated for non-trivial logic
- [ ] No architecture boundary violations (CORE ↔ Integration)

Issue #2380, opened 2026-08-22
[FEATURE] Frontend — per-line return receive and dispose flows


piotrswierzy:
> Part of the OMS programme backlog (**Wave 2**, stream S3, size L). Backlog slug: `W2-42`. Sibling `W2-*` slugs resolve via the Wave-2 epic's checklist. Design of record: returns product spec §5.2/§5.3 (PR #2318).

> **Size justification (L):** two inline per-line forms on a table that must stay usable on a tablet in a warehouse — bulk pre-fill with confirm, remainder handling, over-receipt blocking, ≥44 px targets — plus the destination-naming copy. This is the surface story T4/T5 are judged on.

## Problem / Context

The operator records what physically arrived. Wave 1c's counters are inert until this ships.

## Proposed Solution

A per-line table (advised / received / restocked / scrapped) with an inline RHF+Zod **receive** form per the `generate-label-form.tsx` conventions, defaulting to the remaining advised quantity, a `Receive all as advised` bulk pre-fill behind an explicit confirm, `Mark remainder not returned`, and over-receipt blocked with the spec's message.

Then the inline **dispose** form: quantity, a two-option segmented control (`restock | scrap`), an optional note, and copy **naming the `InventoryMaster` connection** the stock will land in — so the operator knows which book is about to change. No third disposition, no graded condition vocabulary.

Tablet-first: ≥44 px targets, usable at 768 px.

## Classification

**Type**: Frontend
**Layer**: Interfaces
**File(s)**: `apps/web/src/features/returns/components/**`

## Dependencies

- #2376 (`W2-39`), #2378 (`W2-41`)

## Assumptions

- One return detail page serves both desk and warehouse use; no separate warehouse mode in v1.

## Acceptance Criteria

- [ ] Over-receipt is blocked client-side with the spec's message and server-side (both tested)
- [ ] Bulk pre-fill requires an explicit confirm
- [ ] Dispose copy names the destination connection
- [ ] All targets ≥44 px; usable at 768 px; component tests for both forms

PR #3280, opened 2026-09-16
feat(returns): orphan-returns worklist (needs-order / needs-approval groups)

### LINKED ISSUE #3081: feat(returns): orphan-returns worklist (needs-order / needs-approval groups)
## Problem / Context

No view lists returns blocked on attribution or operator approval; `GET /returns` supports filtering but nothing surfaces this specific operator queue.

## Proposed Solution

Add a worklist section (on the returns list page, e.g. via a segment filter, or a dedicated sub-view) with two clearly labelled groups — "Needs an order" (orphan, `internalOrderId === null`) and "Waiting for your OK" (operator-authored, not yet authorized) — each showing exactly one primary action per row, per the reviewed mockup at https://claude.ai/code/artifact/6e97c05b-e518-4393-8917-563fd11acf8c.

## Classification

**Type**: Frontend
**File(s)**: apps/web/src/features/returns/components/orphan-returns-worklist.tsx

## Dependencies

- Part of #3078

## Assumptions

- None

## Acceptance Criteria

- [ ] Orphan and pending-approval returns render in two distinct, labelled groups
- [ ] Each row has exactly one primary action

jakubretajczykBD:
## Summary
- Adds `OrphanReturnsWorklist`: two distinct, labelled groups with exactly one primary action per row
  - **"Needs an order"** — the orphan bucket (`bucket === 'orphan'`), read with an exact server-side filter
  - **"Waiting for your OK"** — operator-authored, not-yet-approved returns (`origin === 'operator_authored' && authorizedAt === null`). No server-side filter exists for this combination, so it reads the `attributed` bucket at the API's own page-size ceiling and filters client-side, disclosing a truncated page honestly rather than trimming quietly
- The primary action in both groups is a `Link` to the return's own detail page rather than an inline dialog — the match/approve affordances that actually live there are #3082/#3083
- Four states per group (loading / error / unreadable envelope / confirmed-empty), mirroring `OrderReturnsPanel`'s discipline that a failed read must never render as "nothing needs attention"

## Test plan
- [x] `tsc --noEmit` clean
- [x] `eslint` clean
- [x] `vitest run src/features/returns` — 257/257 passing
- [x] `check-ui-vocabulary`, `check-design-tokens`, `check-css-structure` all pass

Closes #3081


piotrswierzy (approved):

Third of the returns-write stack. Two operator-actionable groups, one primary action per row.

### Orphan is not re-derived, which was the thing to check

**"Needs an order" filters server-side on `bucket: 'orphan'`**, and that bucket is stamped by the backend from the entity's single definition — `returns.controller.ts:489`, `bucket: record.isOrphan() ? 'orphan' : 'attributed'`. The frontend never reimplements the predicate. `bucket` is a genuine `ReturnListFilter` arm (`return-query.types.ts:52`), and using it here rather than the #2378 `segment` axis is correct: the architecture doc is explicit that `orphans`-as-a-segment is *not* a translation of `bucket=orphan` and that `bucket` stays independently usable.

The four-state discipline per group — loading / error / unreadable envelope / confirmed-empty — is right, including the subtle part: `result === null` on a settled non-loading query reads as an **error**, not a confirmed empty, matching `order-returns-panel.tsx`. A failed read never renders as "nothing needs attention", which is the claim this component is not entitled to make.

Twelve tests, covering both exclusion cases (already-authorized, source-ingested), all four states per group, both truncation disclosures, and the dropped-row count separately from truncation. That is the coverage this file needed.

### Findings

**[IMPORTANT]** — `orphan-returns-worklist.tsx`, the "Waiting for your OK" scan
> The approximation is disclosed honestly in the docblock, and I am not asking for a backend filter. But the **failure mode is biased against exactly the rows that most need attention**, and that is worth stating in the code rather than leaving for someone to discover.
>
> The scan reads `bucket: 'attributed'` at `RETURNS_MAX_LIMIT` (100) and filters client-side. Both returns reads are ordered `createdAt DESC, id ASC` (`return-query.types.ts:42`, confirmed at `return.repository.ts:637`). So the window is the **100 newest** attributed returns — and an operator-authored return awaiting approval ages *out* of that window as newer attributed returns arrive. The longer something has been waiting for approval, the more likely it is invisible in the group whose entire purpose is to show what is waiting.
>
> `approvalScanTruncated` discloses that a page was truncated, but an operator reading "Showing what fits on one page" cannot tell that an item they were expecting has silently left the list. Two cheap mitigations, either of which would close it: state the bound explicitly in the copy ("the 100 most recent — older ones may not appear here"), or add a `createdAt ASC` variant so the oldest unapproved surface first. A `ReturnListFilter` arm for `origin` + unapproved is the real fix and is legitimately out of scope for this epic, but the docblock should say *which way* the approximation fails, not only that it is one.

**[SUGGESTION]** — `orphan-returns-worklist.copy.ts`
> `viewReturn: 'View return'` is declared and never referenced by the component. Either wire it or drop it — an unused entry in a copy table scanned by `check-ui-vocabulary.mjs` is vocabulary the product does not actually speak.

**[SUGGESTION]** — `index.css`, `.orphan-returns-worklist__row`
> `display: flex` with `justify-content: space-between` and no `flex-wrap`, holding a `mono-text` external return id next to a button. A long `externalReturnId` at phone width will compress the action rather than wrap. Every other rule in the block uses tokens correctly (no raw colours), so this is the only responsive note — a `flex-wrap: wrap` or a `min-width: 0` on the id span would settle it.

**[IMPORTANT]** — CI signal
> `Type Check`, `Lint` and `Integration Tests` are still **pending**. `Test` is green, which is the one that matters most for a component PR with twelve new tests, but the review is against an incomplete signal.

### Summary

A well-built worklist that is careful about the things that are easy to get wrong — it reads the server's own orphan definition instead of re-deriving it, keeps a failed read distinct from an empty one, and discloses both truncation and dropped rows rather than trimming quietly. The one substantive concern is that the client-side approval scan's window is ordered newest-first, so its blind spot grows over exactly the items that have waited longest.

### Merge readiness

✅ **Approve.** The IMPORTANT finding is about a documented approximation's *direction*, not about correctness of what ships — worth a docblock sentence and ideally a copy tweak, but it does not block a component that is honest about being bounded. Confirm the three pending checks before merging.

piotrswierzy (changes requested):
## Tech Lead review — #3280 `feat(returns): orphan-returns worklist`

Reviewed against `docs/code-review-guide.md` (4-step), `docs/frontend-architecture.md`, `docs/frontend-ui-style-guide.md`, `docs/architecture-overview.md § 22 Returns`.

### Context established

Baseline read on `main`: `apps/web/src/features/returns/components/order-returns-panel.tsx` (the four-state / truncation-disclosure precedent this PR cites by name), `lib/order-returns-panel.copy.ts`, `hooks/use-returns-query.ts`, `api/returns.types.ts`.

I also verified the PR's central claim that no server-side filter exists for "operator-authored, not yet approved": `ReturnFilters` (`returns.types.ts`) exposes `sourceConnectionId | bucket | internalOrderId | createdFrom/To | segment | stage | money | reason | openedFrom/To`, and the backend `apps/api/src/returns/dto/list-returns-query.dto.ts` exposes the same set — neither carries `origin` or `authorizedAt`. The claim is accurate, and `RETURN_BUCKET_VALUES` is `['orphan', 'attributed']`, so the "Needs an order" filter really is exact.

---

### Findings

**[BLOCKING]** — `orphan-returns-worklist.tsx`, `WorklistGroup` render branch + `needsApprovalItems`

The "Waiting for your OK" group asserts a **positive, confirmed claim about the operator's data off an incomplete read.**

The chain: `approvalScanQuery` reads `{ bucket: 'attributed' }` at `RETURNS_MAX_LIMIT` (100, `returns.types.ts:54`), then filters client-side for `origin === 'operator_authored' && authorizedAt === null`. The filtered array is passed to `WorklistGroup` as `items`. `WorklistGroup` renders:

```tsx
) : items.length === 0 ? (
  <p className="text-muted">{emptyMessage}</p>
) : (
  <>…list… {droppedCount > 0 ? … : null} {truncationNote !== null ? … : null}</>
)
```

So when the scan was truncated **and** none of the scanned rows survived the client-side filter, the operator is told *"Nothing is waiting for your approval."* — and `truncationNote` (`COPY.approvalScanTruncated`, the disclosure that would make that honest) is suppressed, because it lives in the non-empty branch. The state that most needs the caveat is the only state that cannot render it.

This is not an edge case. `total > 100` is the steady state of any install with real volume, and the scan is unordered with respect to `origin`, so an operator-authored return older than the newest 100 attributed returns is structurally invisible *and* actively denied.

It contradicts three things at once:
- The component's own docblock: *"a failed read must never render as 'nothing needs attention', which is a claim about the operator's own data this component is not entitled to make."*
- The copy file's own comment on `needsApprovalEmpty`: *"The read succeeded and answered zero — a positive, confirmed claim."* Here the read did not answer zero; the client-side filter did, over a partial page.
- The repo's standing rule, `docs/architecture-overview.md § 22 Returns` (#2350): *"An empty array is never a positive claim"* — absence and failure must be indistinguishable only in the safe direction, never rendered as reassurance.

The PR body's *"Four states per group (loading / error / unreadable envelope / confirmed-empty)"* is therefore true of the orphan group and not true of the approval group, which has a fifth, unmodelled state: *scanned a bounded page and found none, but did not see everything.*

Minimum fix: when `approvalScanTruncated` is true, render the caveat **instead of** `needsApprovalEmpty` (or above it), so no version of this group states a confirmed-empty off a partial scan. Structurally, `truncationNote` and the `droppedCount` notice should be lifted out of the non-empty branch entirely — they are facts about the *read*, not about the rows.

Separately, and worth a follow-up issue rather than this PR: the durable fix is a server-side filter. Adding `origin` / `unapprovedOnly` to `ListReturnsQueryDto` + `ReturnListFilter` turns this group into an exact filter like its sibling and removes the whole class of problem. The component docblock is honest that the epic never asked for one; the queue it feeds is the one that gates restock, refund and invoice correction, so it is worth asking for.

---

**[IMPORTANT]** — `orphan-returns-worklist.tsx`, `needsOrderTruncated` / `approvalScanTruncated`

```ts
const needsOrderTruncated = needsOrderResult !== null && needsOrderResult.total > needsOrderItems.length;
const approvalScanTruncated = approvalScanResult !== null && approvalScanResult.total > approvalScanResult.items.length;
```

Both gate on the **item count**. The precedent this PR cites wrote a comment specifically forbidding that, at `order-returns-panel.tsx`:

> *"Gated on the PAGE SIZE, never on `items.length`: an unreadable row is excluded from `items` and counted in `droppedCount`, so comparing against the item count would report the SAME row twice — once as unreadable and once as a page limit that was never reached."*

Worked example against this code: a page returning 3 rows of which 2 are unparseable yields `total: 3`, `items.length: 1`, `droppedCount: 2`. The operator gets *both* "2 rows could not be read…" *and* "Showing 1 of 3 — more are waiting to be matched." for the same two rows, and the second sentence is false — nothing is past the page boundary.

Fix: `needsOrderResult.total > RETURNS_PAGE_SIZE` and `approvalScanResult.total > RETURNS_MAX_LIMIT`. I checked this against the two tests in the PR — `should disclose a truncated needs-order page` uses `total: 45` (45 > 20 still true) and `should disclose rows dropped` uses `total: 1` (1 > 20 false), so both keep passing under the corrected predicate. Worth adding the mixed case (`total: 3, droppedCount: 2, items: 1` asserting *no* truncation note) as the regression that pins it.

---

**[IMPORTANT]** — `orphan-returns-worklist.tsx`, same render branch, `droppedCount` path

Same structural cause as the BLOCKING item, different trigger: a page whose rows were **all** unparseable gives `items.length === 0` with `droppedCount > 0`, and the component renders *"Nothing is waiting to be matched to an order."* while suppressing the dropped-rows disclosure. `order-returns-panel.tsx` has the same shape, so this is inherited rather than introduced — but the two groups are not equivalent in consequence: per `docs/architecture-overview.md § 22`, an orphan return blocks *every* downstream trigger (restock, refund, invoice correction), so a false "queue is clear" here has a cost the order-detail panel's does not. Lifting both notices out of the non-empty branch fixes this and the BLOCKING item together.

---

**[SUGGESTION]** — `lib/orphan-returns-worklist.copy.ts:` `viewReturn: 'View return'`

Declared and rendered by nothing — the two action labels actually used are `needsOrderAction` and `needsApprovalAction`. Dead copy in a file `scripts/check-ui-vocabulary.mjs` scans, so it will be linted forever for a sentence no operator can see.

**[SUGGESTION]** — `orphan-returns-worklist.tsx`, `<section aria-label={title}>`

The section carries `aria-label={title}` and contains an `<h3>` with the identical string. `aria-labelledby` pointing at the heading's id gives the same accessible name without the duplicated literal, and keeps the two from drifting. Minor.

**[SUGGESTION]** — not mounted

`OrphanReturnsWorklist` is exported from the barrel and rendered by nothing; the test docblock correctly names this and defers to #3085. That disclosure is the right call, and `docs/lessons.md:789` ("is this MOUNTED?") is the exact failure shape it guards against — flagging it only so the stack is not merged without #3085, since every gate goes green on a component nothing renders.

### Positive observations

- The two groups are genuinely distinct in *kind*, not just in label, and the copy file says so — the orphan group blocks downstream triggers, the approval group is an audit stamp that holds nothing up. The copy is explicit that approving does not gate restock/refund, which is a real and easy thing to imply by accident.
- The orphan group uses an exact server-side filter (`bucket: 'orphan'`) rather than fetching broadly and filtering, and the docblock explains why the other group cannot.
- One primary action per row, a `Link` to the detail page rather than an inline dialog, with the reasoning (#3082/#3083 own the affordances; routing there means this component does not change when they land) stated. `className="button button--secondary button--sm"` on a `Link` matches `features/settings/components/sync-pacing-tile.tsx:44` and `features/oms/components/sourcing-rules-tile.tsx:30`.
- CSS is tokens-only (`--space-*`, `--border-subtle`, `--bg-surface`, `--radius-md`), no raw hex — `docs/frontend-ui-style-guide.md`.
- The `needsOrderIsError` guard includes `result === null` on a settled query, matching `order-returns-panel.tsx`'s `query.error !== null || result === null` — a settled read with no data is correctly not a confirmed empty.
- The test file is genuinely behavioural: it asserts the *exclusions* (already-authorized, source-ingested) rather than only the happy path, and asserts exactly one primary action per row.

---

## Summary

A carefully-reasoned worklist that gets the hard parts right — two structurally different groups, an exact filter where one exists, one primary action per row, tokenised CSS, and a test file that asserts exclusions rather than only the happy path. The blocking concern is that the discipline the component documents most loudly is the one it breaks: the "Waiting for your OK" group derives a confirmed-empty from a client-side filter over one bounded page, and the truncation caveat that would make that honest is rendered only in the branch where there is something to show. On an install with more than 100 attributed returns — the normal case — an operator can be told nothing is waiting for approval while unscanned rows sit past the boundary. The truncation predicates additionally compare against `items.length`, which the cited precedent wrote a comment specifically to forbid.

## Merge Readiness

❌ **Request changes**

## Priority fixes

1. **[BLOCKING]** Never render `needsApprovalEmpty` when `approvalScanTruncated` is true. Lift `truncationNote` and the `droppedCount` notice out of the non-empty branch so they render alongside (or instead of) the empty message — facts about the read must survive a zero-row result.
2. **[IMPORTANT]** Gate truncation on page size, not item count: `total > RETURNS_PAGE_SIZE` / `total > RETURNS_MAX_LIMIT`. Add the `total: 3, droppedCount: 2, items: 1` regression asserting no truncation note.
3. **[IMPORTANT]** Same lift fixes the all-rows-dropped case rendering `needsOrderEmpty`.
4. **[SUGGESTION]** File a follow-up for a server-side `origin`/`authorizedAt` filter on `GET /returns`, which retires the client-side scan entirely.
5. **[SUGGESTION]** Drop the unused `viewReturn` copy key; use `aria-labelledby` instead of a duplicated `aria-label`.

piotrswierzy (approved):
## Re-review — #3280 (orphan-returns worklist)

Re-read against current `main` (`e1ae0aec`). PR base is `3080-returns-mutation-hooks`, so the diff below is this PR's own delta.

### 1. Prior findings

|---|---|---|---|
| 1 (**BLOCKING**) | "Waiting for your OK" renders a confirmed-empty claim off a client-side filter over one bounded page, and suppresses the truncation caveat in exactly that branch | **Addressed** | `orphan-returns-worklist.tsx:105` — `isConfirmedEmpty = items.length === 0 && droppedCount === 0 && truncationNote === null`. Both notices moved **out** of the non-empty branch to `:135-140`, so they survive a zero-row result. Predicates changed to the page ceiling, not the item count: `:148` `total > RETURNS_PAGE_SIZE`, `:161` `total > RETURNS_MAX_LIMIT`. |
| 2 (IMPORTANT) | Truncation predicate double-reported unreadable rows as a page limit | **Addressed** | `:143-148` — comment names the reasoning and the predicate is now ceiling-gated. Pinned by a test that fails against the old form: `orphan-returns-worklist.test.tsx` *"should NOT report a truncated page when total matches drops + shown rows"*. |
| 3 (IMPORTANT) | `droppedCount` notice suppressed on a zero-row page | **Addressed** | `:135-137`, outside the items branch; test *"should render the dropped-rows notice INSTEAD OF a confirmed-empty claim when every row on the page was unreadable"*. |
| 4-6 (SUGGESTION) | mono-text wrap at phone width; copy naming the approximation's DIRECTION | **Addressed** | `index.css` `.orphan-returns-worklist__row .mono-text { min-width: 0; overflow-wrap: anywhere }`; `orphan-returns-worklist.copy.ts` `approvalScanTruncated` now reads *"Showing the 100 most recent attributed returns — an older one waiting for approval may not appear here."* |

**A failed read is still distinguishable from a confirmed empty.** `:139-141` puts `isError` ahead of the content branch, and `needsOrderIsError` (`:152`) treats a settled-but-`null` result as an error rather than an empty, matching `order-returns-panel.tsx`'s guard. Both directions are tested (*"should render a FAILED read as a failure, never as an empty group"*, *"should render an unreadable envelope distinctly from an empty group"*).

**Direction claim verified, not taken on trust.** `ReturnRepository.listReturns` orders `r.createdAt DESC` (`libs/core/src/returns/infrastructure/persistence/repositories/return.repository.ts:637`), so "the 100 most recent" and "an older one may not appear" are both true statements about the read.

**Truncation predicate is sound against the contract.** `PaginatedReturns.total` is a required `number` (`returns.types.ts`) — returns has no `?withTotal=false` second-stage split (`docs/architecture-overview.md` § 28), so `total` is never absent and the comparison cannot silently read `undefined > n` as `false`.

**Root cause (no server-side `origin`/`authorizedAt` filter): not addressed, and correctly so.** `ReturnFilters` and `ListReturnsQueryDto` still expose no such dimension. The client-side approximation is now disclosed rather than hidden, which is the right trade for a FE-only epic; a server-side filter is a backend change this epic never asked for.

`0525edf7` also carries an unrelated production bug fix. It is a real one and I am glad it landed, but it arrives untested.

**IMPORTANT — the six-filter `buildQuery` fix ships with no test, and the existing test that should have caught the bug still cannot.**
`returns.api.ts:269-275` now forwards `segment`, `stage`, `money`, `reason`, `openedFrom`, `openedTo`. Confirmed genuine: `main`'s `buildQuery` sets only `sourceConnectionId`/`bucket`/`internalOrderId`/`createdFrom`/`createdTo`, while `ReturnFilters` has declared all eleven since #2378 and `list-returns-query.dto.ts:87,110,119,128,140,147` accepts every one under the same names. So the segment strip and the stage/money/reason/opened-range filters wrote a param into the URL and the server never saw it — exactly as the comment says.

The problem is what guarded it. `apps/web/src/features/returns/api/returns.api.test.ts:24` is named **"should forward every declared filter"** and passes four of the eleven (`:27-33`). It was green for the whole life of the bug. Per `docs/lessons.md`, a check that cannot fail is a defect, and this one additionally *claims* the coverage it lacks — the next dropped filter will slip through identically. This PR fixes the production code and leaves the test exactly as it was.

Fix: drive that test off `ReturnFilters`' own keys (or, minimally, add the six). Since the filters are a flat string map, a loop over a fixture object asserting every key round-trips is a few lines and makes the omission impossible to repeat.

Secondary: this is scope creep into a shipped surface (`/returns`) inside a worklist PR. I would not ask you to split it out — reverting a live fix to re-land it is worse — but it is the reason the test matters more than it otherwise would.

**IMPORTANT — the new bespoke row shape is not registered in the style guide.**
`.orphan-returns-worklist__row` (`index.css`) is a non-`DataTable` row list with its own composition and its own height — the same class of thing as `.who-decides-row`, `.sales-document-market-row` and the #3223 price-changes card. `docs/frontend-ui-style-guide.md` § Density & Row Heights states the standing instruction plainly: *"it is a row shape with its own composition and its own height, and this section's standing instruction is to register one before shipping it."* This PR adds the shape and registers nothing. One table row / short entry naming the class, the height ("auto, content-driven") and why it is not a `DataTable`.

**SUGGESTION — the row's action has no width-keyed 44 px floor.**
The action is `className="button button--secondary button--sm"` (`orphan-returns-worklist.tsx:127`), and `.button--sm` is `height: 1.75rem` (28 px) at `index.css:746-750`. The global floor fires under `@media (hover: none) and (pointer: coarse)` (`index.css:5795-5801`) and matches `.button`, so a real phone is covered; a 767 px desktop window is not, which is the precise gap the #3223 entry and the #2380 review both wrote down (*"the 44 px floor is … keyed on WIDTH"*). Mobile and tablet are in scope for FE work here, so a `@media (max-width: 767px)` `min-height: 44px` on this one control closes it.

**SUGGESTION — heading order regresses inside each group.**
The section renders `<h2 class="section-title">` (`:169`), then `<h3>` per group (`:110`), and then `LoadingState`/`ErrorState` emit their own `<h2 class="state-card__title">` (`shared/ui/feedback-state.tsx:34,59`) *inside* the `<h3>`'s section. A screen-reader user traversing by heading level sees h2 → h3 → h2. It is inherited from the shared primitive rather than introduced here, so I would not hold the PR for it, but the group `<h3>`s could be `<h2>`s with the section title promoted, or the primitives given a heading-level prop.

**Still unmounted.** Nothing renders `OrphanReturnsWorklist` outside its own test (`index.ts` exports it; no page imports it). The test docblock states the deferral to #3085 explicitly, which is the right treatment of `docs/lessons.md` § *"is this MOUNTED?"* — the obligation moves rather than disappears. Flagging so it is not forgotten: **the mounting PR must assert the mount**, or this is dead code that type-checks.

**Permission gating:** correctly absent. Every row action is a `<Link>` to `/returns/:id` — navigation, not a write — so `useWriteAccess`/`ReadOnlyLock` would be wrong here per the `docs/frontend-architecture.md` § Access Control table (row two gates *write affordances*). The writes live in #3281/#3083.

**Test quality on the BLOCKING fix is genuine**, not a rubber stamp: *"should render the truncation caveat INSTEAD OF a confirmed-empty claim when the scan was truncated and nothing survived the client-side filter"* asserts both directions (caveat present, empty-message absent) and fails against the pre-fix component, where `truncationNote` sat inside the `items.length > 0` branch and could not render at all.

| | Count |
|---|---|
| BLOCKING | 0 |
| IMPORTANT | 2 |
| SUGGESTION | 2 |

**Merge readiness: approved.** The BLOCKING finding is genuinely closed in the code, with a test that would fail against the old behaviour, and both prior IMPORTANTs are closed with named comments at the call sites. Neither new IMPORTANT is a correctness hole in the worklist itself.

**Priority fixes (in order):**
1. Extend `returns.api.test.ts:24` to cover all eleven declared filters — ideally keyed off `ReturnFilters` so it cannot fall behind again. The bug this PR fixed was invisible for a release because that test's name outran its assertions.
2. Register `.orphan-returns-worklist__row` in `docs/frontend-ui-style-guide.md` § Density & Row Heights.
3. Width-keyed 44 px floor on the row action.

piotrswierzy (approved):
## Tech Lead review — `feat(returns): orphan-returns worklist (needs-order / needs-approval groups)`

Strongest PR in this stack. The absence-vs-failure discipline is applied exhaustively and the honest disclosure of the approval scan's approximation is exactly right.

### 🎯 The `buildQuery` fix is a bigger deal than the diff suggests

`apps/web/src/features/returns/api/returns.api.ts`

Six filters — `segment`, `stage`, `money`, `reason`, `openedFrom`, `openedTo` — were declared on `ReturnFilters`, parsed from the URL, re-rendered the page, and **never reached the server**. I confirmed against `apps/api/src/returns/dto/list-returns-query.dto.ts`: all six exist backend-side and are fully implemented. So the entire #2378 worklist strip and every stage/money/reason/date filter on `/returns` has been inert since it shipped — clicking a segment card changed the URL and returned the same unfiltered rows.

That's a good catch and worth calling out in the PR title or description, because right now it's buried in a worklist PR.

### 🟡 IMPORTANT — that fix ships with no regression test

`returns.api.test.ts` isn't touched. The failure mode here was *silent* — a param quietly missing from a query string, with the page still rendering fine — which is precisely the class of bug a one-line assertion pins forever:

```ts
it('should forward every declared filter to the query string', async () => {
  const request = vi.fn().mockResolvedValue(/* … */);
  await createReturnsApi(request).list(
    { bucket: 'orphan', segment: 'all_open', stage: 'disposed', money: 'pending',
      reason: 'damaged', openedFrom: '…', openedTo: '…', /* … */ },
    { limit: 20, offset: 0 },
  );
  const [url] = request.mock.calls[0] as [string];
  for (const p of ['bucket','segment','stage','money','reason','openedFrom','openedTo']) {
    expect(url).toContain(`${p}=`);
  }
});
```

Better still, iterate the `ReturnFilters` keys so a *seventh* filter added later without a `buildQuery` line fails the build rather than repeating this. Please add something before merge — everything else here I'd take as-is.

### ✅ Absence vs failure, done properly

Four distinct states per group — loading / error / unreadable envelope / confirmed-empty — and the confirmed-empty message is withheld whenever the read is known incomplete:

```ts
const isConfirmedEmpty = items.length === 0 && droppedCount === 0 && truncationNote === null;
```

That's the right predicate. *"Nothing is waiting to be matched to an order"* is a positive claim about the operator's data, and an orphan return blocks every downstream trigger (§22), so a false "queue is clear" here costs more than the same shape on the order panel. Three separate tests pin it (truncated-scan, all-rows-unreadable, source-ingested-only), which is the coverage that matters.

`needsOrderIsError` treating `result === null` on a settled query as an error — not an empty — is the same rule one level down, and matches `order-returns-panel.tsx`'s guard.

### ✅ The truncation predicates are gated on the page ceiling, not the item count

`total > RETURNS_PAGE_SIZE` / `total > RETURNS_MAX_LIMIT`, never `items.length`. The comment explains why, and the test (`total: 3, droppedCount: 2, items: 1`) is the one that would have caught the wrong version: comparing against the item count would report the same two unreadable rows a second time as a page limit that was never reached.

### ✅ The approval scan's approximation is disclosed with its *direction*

> *'Showing the 100 most recent attributed returns — an older one waiting for approval may not appear here.'*

Naming which way the approximation leans is the useful part. The read is newest-first, so an unapproved return ages **out** of the window — the longer it has waited, the likelier this misses it, which is the opposite of what an operator would assume. Saying so is the difference between a caveat and a warning.

The docblock is equally candid that this exists only because `ReturnFilters` has no `origin` / `authorized` axis, and that the epic never asked for one.

### ✅ Routing rather than duplicating

Primary action is a `Link` to `/returns/:id`, not an inline dialog — so this component works today *and* works unchanged once #3082/#3083 give the detail page its match/approve affordances. The destination gains the dialog, not this list. Exactly the right seam.

Copy lives in `lib/orphan-returns-worklist.copy.ts`, which is a filename `scripts/check-ui-vocabulary.mjs` actually scans — the component body has no operator sentence in it. Per `docs/frontend-architecture.md`, a `Record` left inside the component would be scanned by nothing.

### 🟢 SUGGESTIONS

1. **The "needs an order" truncation notice has no remedy attached.** *"Showing 20 of 45 — more are waiting to be matched"* tells an operator 25 returns exist that they cannot reach from this screen, with no link. Now that `buildQuery` actually forwards `segment`, a `<Link to="/returns?segment=orphans">See all</Link>` beside the note closes it. This is the same alarm-without-a-remedy shape the repo is careful about elsewhere.
2. **The approval scan costs a 100-row read on every mount** to client-side filter for two fields. A backend `origin` / `authorized` filter would make the group exact *and* the read cheap, and would let the caveat copy be deleted rather than maintained. Worth filing as a follow-up against #2376 rather than carrying the approximation indefinitely.
3. `titleId` is derived from operator copy via `title.replace(/\s+/g, '-')`. Works for today's two titles; a title gaining an apostrophe or colon would put it in a DOM id. A literal id per group is cheaper to reason about.

### Notes

- CSS is new, scoped, and uses tokens throughout; the `min-width: 0` + `overflow-wrap: anywhere` on the mono id with its explanation is the kind of thing that stops someone "simplifying" it later.
- Test file's header comment correctly notes that *mounting* is asserted one level up in #3085 — good, since `docs/lessons.md` has a whole entry about components that are tested but never rendered.
- CI green.

**Verdict:** ✅ Approve, with a request to add the `buildQuery` regression test before merge. The rest is exemplary.

Issue #2376, opened 2026-08-22
[TASK] Interface — returns write API


piotrswierzy:
> Part of the OMS programme backlog (**Wave 2**, stream S2, size M). Backlog slug: `W2-39`. Sibling `W2-*` slugs resolve via the Wave-2 epic's checklist.

## Problem / Context

Wave 1c ships the returns read API (#2334 / `W1c-7`). Every Wave-2 write needs a surface, and `restock_blocked` must be visible in the write response so the UI renders it without a second read.

## Proposed Solution

`POST /returns/:id/lines/:lineId/receive`, `.../dispose`, `.../mark-stock-handled`, `POST /returns/:id/authorize`, `POST /returns/:id/match-order`, `POST /returns/record`, `POST /returns/:id/refund` (confirm a linked `RefundRecord`), `POST /returns/:id/commission-refund/claim`, and `GET/POST /returns/:id/correction-proposal`. Counter-invariant violations answer 409 with a distinguishable code; a `restock_blocked` outcome is returned **in the 2xx body** — the disposition succeeded, the master write did not.

Writes use the existing write-permission model; no new permission value (adjudicated).

## Classification

**Type**: Interface
**Layer**: Interfaces
**File(s)**: `apps/api/src/returns/**` (extending the W1c-7 controller)

## Dependencies

- #2370 (`W2-33`), #2371 (`W2-34`), #2372 (`W2-35`), `W2-37` (see epic checklist), #2374 (`W2-38`)

## Assumptions

- The Wave-1c read DTOs are extendable rather than needing a parallel shape.

## Acceptance Criteria

- [ ] Over-receipt answers 409 with an actionable code
- [ ] A blocked restock answers 2xx carrying `restockBlocked` with quantity, sku and connection name
- [ ] Every endpoint guarded; read-only role refused on all writes
- [ ] Swagger annotations; integration test for receive → dispose → blocked → attest

PR #3281, opened 2026-09-16
feat(returns): match-to-order dialog + unknown-order/already-attributed handling

### LINKED ISSUE #3082: feat(returns): match-to-order dialog + unknown-order/already-attributed handling

jakubretajczykBD:
## Summary
- Adds `MatchReturnDialog`: a single-field dialog with an explicit "this can't be undone" warning, satisfying both acceptance criteria exactly:
  - 400 `unknown-order` renders as a FIELD error naming the typed value (`readMatchRefusalReason` reads the response body's `reason` field, never the message string, matching `decline-error.ts`'s precedent)
  - 409 `already-attributed` renders as a DISTINCT message ("Already linked"), never the generic failure — attribution is monotonic, so this is not a retryable failure, it is news the return is already resolved
- No order-search endpoint exists (`GET /orders` has no `search` param) and an order carries no buyer name on the FE contract at all (a deliberate PII boundary), so the field is free-text for the internal order id, backed by a `<datalist>` of the 20 most recent orders — the same bounded, client-side-filtered shape `command-palette-provider.tsx` already ships for the identical reason
- `autoComplete="off"` on the order field — Chrome's own history-based autocomplete competes with the intentional `<datalist>` and, inside a Radix dialog, can force the page to scroll to keep its own popup visible

## Test plan
- [x] `tsc --noEmit` clean
- [x] `eslint` clean
- [x] `vitest run src/features/returns` — 266/266 passing
- [x] Cross-feature import goes through the orders barrel (`../../orders`), matching the existing `return-money-panel.tsx` precedent

Closes #3082


piotrswierzy (approved):

Top of the returns-write stack, and the PR that had to get the refusal split right.

### The two refusals render as two different things

This is the acceptance criterion and it is met properly, not nominally:

- **400 `unknown-order` → a field error naming the typed value** (`COPY.unknownOrder(trimmed)`), rendered on the `FormField` the operator can correct. Right, because the fix is in the request.
- **409 `already-attributed` → an entirely different dialog panel** — its own title, body and single acknowledge button, replacing the form. Right, because it is not a retryable failure, it is news that the work is already done. It also calls `onMatched?.()`, which is the correct reading of a lost race: the return **is** attributed, so the caller should treat it as resolved rather than as an error.
- **Everything else** falls through to the generic sentence, guarded by `readMatchRefusalReason(mutation.error) === null` so a recognised refusal never double-renders.

`readMatchRefusalReason` reads the response body's `reason` field via `ApiError.details`, never the message string — matching what `ReturnsExceptionFilter` actually emits, which I verified emits `reason` as a top-level field for this exception. It returns `null` for an unrecognised value rather than rendering `undefined`, so an error shape this build predates degrades to the generic message. That is the right failure direction.

Eight tests, and they pin exactly the distinctions that matter: the 400 as a field error *naming the typed id*, the 409 as a message distinct from the generic one, `onMatched` firing on both the success and the conflict path, and the generic fallback for an unrecognised failure.

### Monotonicity is surfaced, not assumed

`COPY.warning` — "This can't be undone — double-check the order before confirming" — renders unconditionally above the footer, and the first test asserts it is shown up front. Given `claimAttribution` is a fill-in-when-NULL conditional UPDATE with no unmatch anywhere in the product, a mis-match is genuinely uncorrectable, and #2376 records that this is precisely why the UI owes a confirmation rather than a bare button. A dialog plus a persistent irreversibility warning plus an explicit "Confirm match" clears that bar.

The `onOpenChange` guard refusing to close while `mutation.isPending` is a small thing done right — closing mid-write would orphan the result of an irreversible action.

### Findings

**[SUGGESTION]** — `match-return-dialog.tsx`, the `<datalist>`
> `<option key={…} value={order.internalOrderId}>{label}</option>` puts the raw `ol_order_…` id in `value` and the human-readable order number in the text node. Browsers render `value` as the primary line and the label as secondary, so an operator scanning suggestions reads a wall of opaque ids first and the number they actually recognise second. The field must *submit* the id, so this is close to forced — but it is worth knowing that the suggestion list reads worse than the free-text placeholder implies. If `FormField` can carry a separate display, or if the label could lead, that would be a real improvement to the primary path. Not blocking; the fallback (typing an id directly) works and is tested.

**[SUGGESTION]** — the dialog is exported but not mounted
> Expected as the top of a stack — the worklist's "Match to an order" link routes to `/returns/:id`, and `return-detail-page.tsx` does not mount `MatchReturnDialog` on this branch, so the action currently lands on a page without the affordance. #3085 closes that. Stating it so the stack's merge order is deliberate rather than assumed: #3280's primary action is inert until whichever PR mounts this.

**[IMPORTANT]** — CI signal
> `Type Check`, `Lint`, `Test`, `Integration Tests` and `Build` are all still **pending** — only `Detect changed paths`, `PHP Unit Tests` and `Docker Build Smoke Test` have reported. This is the least-verified PR of the four at review time, and it is the one carrying the most behaviour. Confirm green before merging.

### Summary

The match-to-order dialog, and the one place in this stack where getting the HTTP contract wrong would have been actively misleading — telling an operator their return was in a bad state when they had simply typed a wrong id. It branches on the response's `reason` vocabulary rather than on status alone or on message text, renders the two refusals as genuinely different experiences, and surfaces the irreversibility that the backend's monotonic attribution makes non-negotiable.

### Merge readiness

✅ **Approve**, conditional on the five pending checks reporting green. No changes required; both findings are observations about the suggestion UX and the stack's mounting order.

piotrswierzy (approved):
## Tech Lead review — #3281 `feat(returns): match-to-order dialog`

Reviewed against `docs/code-review-guide.md` (4-step), `docs/frontend-architecture.md`, `docs/frontend-ui-style-guide.md`.

### Context established

Baseline read on `main`: `apps/web/src/features/returns/components/return-decline-action.tsx` (the single-field write dialog precedent), `lib/decline-error.ts` (the read-the-`reason`-field precedent), `shared/ui/form-field.tsx`, `shared/api/api-error.ts`, `features/orders/hooks/use-orders-query.ts`.

**Both acceptance criteria are backed by a real contract, verified rather than assumed.** `apps/api/src/common/filters/returns-exception.filter.ts:160` emits `reason` as a top-level body field, and `:175` splits `ReturnMatchRefusedError` by reason — `unknown-order` → 400, everything else → 409 — with `returns-exception.filter.spec.ts:218` (*"should split ReturnMatchRefusedError by reason, not by class"*) pinning it. So `readMatchRefusalReason` reads a field that genuinely exists, and both branches of this dialog are reachable in production rather than only in its own mocks.

Layering and state ownership are correct: server state through the #3080 mutation hook, local dialog state in `useState`, no raw `fetch()`, no `any`, cross-feature import via the orders barrel (`../../orders`) — the only route `.eslintrc.js` permits and the same one `return-money-panel.tsx` takes.

---

### Findings

**[IMPORTANT]** — `match-return-dialog.tsx`, the `<datalist>` + confirm path

The dialog's own warning is *"This can't be undone — double-check the order before confirming."* — but there is nothing to double-check. The field holds a raw `ol_order_…` string, the `<datalist>` writes `order.internalOrderId` into it as the `value` while displaying `externalOrderNumber` as the option text, and confirm fires immediately. So the operator's last sight of the human-readable order reference is inside a native suggestion popup that closes on selection; what remains on screen at the moment of an irreversible write is an opaque internal id. How much of that popup shows the `value` versus the option text also varies by browser, which is precisely why the on-screen state after selection is the thing that has to carry the identity.

For a monotonic write with no corrective action, the dialog should echo what it resolved: when the typed value matches one of the fetched orders, render that order's `externalOrderNumber` (and, if available, its placed date) under the field, so "double-check the order" is an instruction the operator can actually follow. `features/orders` already has the data in hand — `ordersQuery.data.items` is loaded and `syncStatus[].externalOrderNumber` is exactly what the `<option>` label already uses. Failing that, putting the id into the option text so both strings are visible would be a cheaper partial fix.

**[SUGGESTION]** — `lib/match-error.ts`, `readMatchRefusalReason`

The function branches on `details.reason` alone and never consults `error.status`, while the module docblock carefully documents the status pairing (`already-attributed` = 409, `unknown-order` = 400). Ignoring the status is arguably the *better* choice — it is the same reason-over-message rule `readBlockedTrigger` follows in `decline-error.ts`, and it degrades gracefully if the filter's status mapping is ever revised. But as written a reader can reasonably take the docblock as describing enforced behaviour. One line saying the status is deliberately not consulted, because `reason` is the discriminator the backend guarantees, closes the gap.

**[SUGGESTION]** — `match-return-dialog.tsx`, generic-error gate

```tsx
{mutation.isError && readMatchRefusalReason(mutation.error) === null ? <Alert tone="error">…</Alert> : null}
```

This is the "is this the failure I already handled somewhere else?" predicate, and `shared/api/api-error.ts` ships `isUnmappedApiError(error, isMapped)` for exactly it — extracted by the #3068 tech review *after* `LocationDialog` and `LocationDeleteDialog` were each found computing it inline with the same shape and no shared source, with the note that "a third mutation dialog would otherwise have copied it a third time". This is that third dialog. `isUnmappedApiError(mutation.error, (e) => readMatchRefusalReason(e) !== null)` reads the same and joins the shared helper.

**[SUGGESTION]** — not mounted

`MatchReturnDialog` is exported from the barrel and rendered by nothing; the PR body defers mounting to the detail page (#3083). Same note as #3280 — flagged only so the stack is not merged without its mounting PR, since `docs/lessons.md:789` ("is this MOUNTED?") records a nine-test panel that passed every gate and was rendered by nothing. The disclosure in the PR body is the right handling.

**[SUGGESTION]** — `COPY.fieldRequired` attribution

The copy comment says it *"deliberately mirrors the decline dialog's phrasing for an empty required field"*, but the string is `'Enter an order id, or pick one from the list.'` — field-specific, and not a mirror of anything in `return-detail.copy.ts`. The sentence itself is good; the comment claims a shared vocabulary that does not exist, which is the kind of note a later reader will trust and act on.

### Positive observations

- `useState` + `FormEvent` + `FormField` rather than reaching for React Hook Form is the right call and matches the immediate sibling: `return-decline-action.tsx:120-144` does exactly this for the same shape. `docs/frontend-architecture.md:229` names RHF for forms, and `:300` scopes that to "draft and edit flows" — a two-control confirm dialog is not one.
- Accessibility is handled by delegation rather than by hand: `FormField` wires `htmlFor`/`id`, `aria-describedby` (description + error) and `aria-invalid` from the `error` prop, and `Input` spreads through to the native element, so the field error is announced and associated without this file doing anything. `Dialog`/`DialogTitle`/`DialogDescription` satisfy Radix's labelling requirement.
- The `autoComplete="off"` rationale is a real, specific bug (Chrome's history popup competing with the `<datalist>` and forcing scroll inside a dialog), documented at the line rather than left as folklore.
- Calling `onMatched?.()` on the 409 as well as the success is subtle and correct — the return *is* attributed either way, so a caller that refetches on "matched" must refetch on the conflict too. The test pins it.
- `readMatchRefusalReason` returns `null` for unrecognised values rather than passing them through, so a reason this build predates falls to the generic sentence instead of rendering a raw backend token at the operator.
- The tests cover both AC branches, the unrecognised-failure fallback, the stale-field-error clear, and assert the *absence* of the generic sentence in the mapped cases — i.e. they test the discrimination the ACs are actually about, not just that something rendered.

---

## Summary

The strongest PR in the stack. Both acceptance criteria are satisfied against a contract I verified on the backend rather than taking on trust, the error mapping reads the machine-readable `reason` field instead of a message string, accessibility comes from the shared `FormField` wiring rather than hand-rolled ARIA, and the tests assert the discrimination between the two refusals rather than merely that each renders. The one thing worth changing before this is mounted is that a dialog whose own copy says "double-check the order" leaves nothing on screen to check at confirm time — for a write with no unmatch, the resolved order's human-readable reference should be echoed beside the field.

## Merge Readiness

🔄 **Approve with changes** — merge after addressing the confirm-time order echo.

## Priority fixes

1. **[IMPORTANT]** Echo the resolved order's `externalOrderNumber` under the field when the typed value matches a fetched order, so the irreversibility warning is actionable.
2. **[SUGGESTION]** Note in `match-error.ts` that `error.status` is deliberately not consulted.
3. **[SUGGESTION]** Use the shared `isUnmappedApiError` helper for the generic-error gate.
4. **[SUGGESTION]** Correct the `fieldRequired` copy comment's claim of mirroring the decline dialog.

piotrswierzy (approved):
## Re-review — #3281 (match-to-order dialog)

Re-read against current `main` (`e1ae0aec`).

### 1. Prior findings

|---|---|---|---|
| 1 (IMPORTANT) | The dialog warns *"double-check the order before confirming"* but leaves only an opaque `ol_order_…` string on screen, for a write with no unmatch | **Partially addressed** | `match-return-dialog.tsx:94-98` resolves the typed value against fetched orders and `:198-205` echoes `COPY.resolvedOrder(externalOrderNumber)` beneath the field. Both directions tested (*"should echo the resolved order number when the typed value matches a fetched order exactly"* / *"should NOT echo … for a value that does not exactly match"*). **But the echo can only ever resolve against the 20 most recent orders** — see below. |

Four prior SUGGESTIONs: the two I can see in the diff are addressed — `readMatchRefusalReason` now explicitly never consults `error.status` (`lib/match-error.ts:20-26`, reasoned in the docblock), and `fieldRequired`'s comment no longer claims a shared vocabulary that does not exist (`lib/match-return-dialog.copy.ts:29-34`).

**Re-verified against the moved base:** `apps/api/src/common/filters/returns-exception.filter.ts:160` still emits `reason` as a top-level field and `:175` still splits `ReturnMatchRefusedError` 400/409 by reason, so both AC branches remain production-reachable. `isUnmappedApiError` (`shared/api/api-error.ts:74`) and `ApiError.details` carrying the response body (`:36-46`) both hold, so `readMatchRefusalReason`'s `details.reason` read is correct — same shape `decline-error.ts:32-34` already uses for `trigger`.

**IMPORTANT — the resolved-order echo is bounded to 20 orders, which is likely to miss the case it exists for.**
`RECENT_ORDERS_LIMIT = 20` (`:57`) and `resolvedOrder` is `orders.find(o => o.internalOrderId === trimmedValue)` (`:94-98`) over that one unfiltered page. So the echo renders when the operator picks from the `<datalist>`, and renders nothing when they paste an id — which is the shape this dialog is *for*. An orphan return is typically an order that arrived before the connection existed, under another connection, or long enough ago that the reconcile never caught it (`docs/architecture-overview.md` § 22); "recent" is close to the opposite of the population. For a paste, the warning still points at an opaque string at the moment it matters most, and there is no unmatch.

The fix is small and the read already exists: a single `GET /orders/:internalOrderId` on a debounced, `ol_order_`-shaped value, echoing the number on a hit and rendering nothing on a miss — the miss stays silent because the 400 `unknown-order` path is what states it. That keeps the exact-match discipline `:92-93` correctly argues for while removing the recency bound. I am recording this as partially-addressed rather than reopening it: the dialog is strictly better than before, the write is still operator-confirmed, and the backend refuses an id it never minted.

**IMPORTANT — the confirm affordance has no `useWriteAccess` / `ReadOnlyLock`.**
`docs/frontend-architecture.md` § Access Control And UI Visibility is a table, and row two is unambiguous: *"A write affordance — button, checkbox, form, menu item → `useWriteAccess` + `ReadOnlyLock`"*, hidden for a session lacking the permission, rendered disabled with a tooltip in demo mode. The `Confirm match` submit (`:216-218`) is a write affordance and carries neither, while four siblings in the same folder do — `return-decline-action.tsx:186`, `return-custody-panel.tsx:256,317`, `return-money-panel.tsx:85`, all on `useWriteAccess('orders:write', demoMode)`. `order-returns-panel.tsx:29` even documents the inverse case in a comment, so the convention is stated on both sides.

Two reasons I am grading this IMPORTANT rather than BLOCKING: the component is unmounted today, so there is no live exposure; and the server enforces `@Roles` independently, so this is a UI-visibility gap, not an authorization hole. But it must be closed **before** the mounting PR ships, and closing it inside this component is the right place — the dialog owns its footer, and every sibling puts the lock around its own confirm rather than relying on a caller. A parent-side lock would also leave a demo viewer able to open a dialog whose primary button is live.

This is the same gap a sibling review found on #3283/#3285, so it is a pattern in this wave rather than a one-off.

**Still unmounted.** Nothing renders `MatchReturnDialog` outside its own test; `index.ts` exports it and no page imports it. Deferred to #3083/#3085, consistent with #3280. Same standing obligation: the mounting PR must assert the mount (`docs/lessons.md` § *"is this MOUNTED?"*).

Clean on the rest: no raw `fetch()` (the write goes through `useMatchReturnToOrderMutation`, the orders read through `useOrdersQuery`); the cross-feature import is barrel-to-barrel (`:52` `from '../../orders'`) and I confirmed `features/orders` imports nothing from `features/returns`, so no barrel cycle (#337/#359); every operator sentence lives in `match-return-dialog.copy.ts`, a folder `check-ui-vocabulary.mjs` scans; `autoComplete="off"` (`4be1c2a8`) is a real fix with a real reason, not cargo — Chrome's history popup inside a dialog does force the scroll the comment describes. `mutation.isPending` guards submit, cancel, and `onOpenChange`, so the dialog cannot be dismissed mid-write.

| | Count |
|---|---|
| BLOCKING | 0 |
| IMPORTANT | 2 |
| SUGGESTION | 0 |

**Merge readiness: approved.** No BLOCKING outstanding; the prior IMPORTANT moved in the right direction and the error-mapping half of this PR is the strongest part of the stack — it reads a field the filter genuinely emits, degrades to a generic sentence for a shape this build predates, and tests both refusals plus the fallback.

**Priority fixes (in order):**
1. `useWriteAccess('orders:write', demoMode)` + `ReadOnlyLock` around the dialog footer — **before** #3083/#3085 mounts it, not after.
2. Resolve the echoed order by id rather than by membership of the 20 most recent, so the irreversibility warning has something to point at on the paste path.

piotrswierzy (approved):
## Tech Lead review — `feat(returns): match-to-order dialog + unknown-order/already-attributed handling`

The two refusals are rendered as two genuinely different things, which is the whole point of the AC, and the `resolvedOrder` echo closes a real hole. One small state-leak to fix.

### ✅ The 400/409 split is modelled correctly, and `readMatchRefusalReason` reads the right field

`lib/match-error.ts` branches on the response body's `reason`, **never** on `error.status` or the message string — and the docblock says why: `reason` is the guaranteed discriminator (`ReturnsExceptionFilter` emits it as a top-level field for `ReturnMatchRefusedError`), while the status pairing is today's mapping. Same rule `readBlockedTrigger` already follows. Returning `null` for an unrecognised reason so it degrades to the generic sentence is the right failure direction.

The renderings then differ in *kind*, not just in wording:

- **400 `unknown-order` → a field error naming the typed value.** The fix is in the request, so it belongs beside the input. Naming the value back (`"#ZZZ999"`) rather than saying "invalid order" lets the operator see their own typo.
- **409 `already-attributed` → the dialog swaps to an acknowledge-only view.** Correct: attribution is monotonic, so this is not a retryable failure and offering a retry button would be actively wrong. Calling `onMatched?.()` on this path too is the subtle one and it's right — the return *is* attributed by the time this renders, so the caller must refresh exactly as it would on success. There's a test pinning it.

### ✅ The `resolvedOrder` echo

This is the best change in the PR. A `<datalist>` writes the raw `value` into the field and throws away the readable option text once picked, so at the exact moment `"double-check the order before confirming"` matters, the operator was looking at `ol_order_a1b2c3…` and nothing else. Echoing the matched order's own number underneath — sourced from data already fetched for the datalist, not a second read — gives the warning something to point at.

Gating it on **exact** match is also right, and the reasoning is stated: a partial or case-mismatched id is precisely the typo the 400 path exists to catch, so echoing a "resolved" order for it would be reassurance in the wrong place. Both directions are tested.

### ✅ Honest about the absent search endpoint

The copy says "pick a recent order below, or type its internal order id" rather than implying a search, because there isn't one — `GET /orders` has no `search` param and an order carries no buyer name on the FE contract at all (the PII boundary). Matching the command palette's bounded-20 shape rather than inventing a different one is the right instinct, and it has a bonus the docblock doesn't claim: `command-palette-provider.tsx` issues `useOrdersQuery(undefined, { limit: 20 })` with an **identical query key**, and it wraps the whole `AppShell` — so this dialog's read is deduped against one the app already runs on every authenticated page. No extra request, even though the hook is called unconditionally.

`autoComplete="off"` with the explanation about Chrome's history popup fighting the datalist and scroll-jumping the dialog is the sort of detail that gets silently deleted later without that comment.

### 🟡 IMPORTANT — `mutation.reset()` is missing, so a generic error survives a close/reopen

`match-return-dialog.tsx`, `resetAndClose`:

```ts
function resetAndClose(): void {
  setValue('');
  setFieldError(undefined);
  setAlreadyAttributed(false);
  onOpenChange(false);
}
```

All three pieces of *local* state are cleared, but `mutation.error` isn't — and the component stays mounted across close/reopen (the parent controls `open`; `useMatchReturnToOrderMutation` is called at the top level, not inside `DialogContent`).

Repro: open → submit → network failure → generic Alert renders → Cancel → reopen. `isUnmappedApiError(new Error('network down'), …)` is still `true`, so the operator is greeted by **"The match could not be completed. Try again."** on a fresh dialog before touching anything.

The 400 and 409 paths don't leak (the former is a mapped `ApiError`, the latter swaps the whole view), which is why the existing tests don't catch it — it's specifically the generic branch.

```ts
function resetAndClose(): void {
  setValue('');
  setFieldError(undefined);
  setAlreadyAttributed(false);
  mutation.reset();
  onOpenChange(false);
}
```

Worth a test alongside it: reject with a plain `Error`, close, reopen, assert `COPY.genericError` is gone.

### 🟢 SUGGESTION

`useOrdersQuery` fetches rows **and** the total; this dialog needs only the 20 rows for a `<datalist>`. `useOrderRowsQuery` (#2947) exists for exactly that and its own docblock explains the cost — `docs/engineering-standards.md` measures the orders `COUNT` at 142 ms of a 149 ms request at a million rows.

In practice this costs nothing today because of the cache sharing noted above, and it would need one extra export line in `features/orders/index.ts` (`useOrderRowsQuery` isn't in the barrel yet). So: low priority, and the same applies to the command-palette precedent. Worth doing whenever that barrel line gets added for another reason.

### Notes

- Copy fully in `lib/match-return-dialog.copy.ts` (scanned by `check-ui-vocabulary`); the corrected comment on `fieldRequired` — no longer claiming a shared vocabulary that doesn't exist — is a good catch from the previous round.
- Eleven tests, covering both refusals, the fall-through, both echo directions, and stale-field-error clearing. Solid.
- CI green.

**Verdict:** ✅ Approve. Please land the `mutation.reset()` fix — it's two lines and the only thing I'd hold for.

Issue #1615, opened 2026-07-15
[TASK] Frontend - Connection Actions tab visible to demo viewers with only the write submit disabled


norbert-kulus-blockydevs:
## Problem / Context

On a connection's **Actions** tab (`apps/web/src/features/connections/components/ConnectionActionsPanel.tsx`), every action row is gated by `usePermission('connections:write')` / `usePermission('sync:write')` and simply **not rendered** when the permission is missing. For a demo viewer this leaves the "Operator / Actions" panel empty - they cannot even see what actions exist or open the Edit form / Trigger-sync dialog to explore them.

Desired demo behavior: the viewer should **see every action and be able to open** the ones that present a form/modal first (Edit connection -> `EditConnectionForm`; Trigger sync -> `TriggerSyncDialog` in `apps/web/src/features/sync-jobs/components/TriggerSyncDialog.tsx`), edit fields freely, and only be blocked at the **final write submit**. Actions that write directly (Test connection via `useTestConnectionMutation`, Disable via `useDisableConnectionMutation`, plugin actions like Erli "Configure webhooks") should be **visible but disabled**. Nothing hidden.

## Proposed Solution

Change the gating in `ConnectionActionsPanel` (and plugin `*-connection-actions.tsx`) from "hide when no write permission" to "render, but disable the write submit" for demo viewers:
- Always render the action rows; for a demo viewer, render trigger buttons that open `EditConnectionForm` / `TriggerSyncDialog` normally, with inputs editable, and disable only the final submit ("Save changes", "Trigger") with a "read-only in demo" tooltip.
- Disable the direct-write buttons (Test connection, Disable connection, Manage connection, plugin actions) with the same tooltip.
- Keep the current hide-vs-disable behavior unchanged for genuinely unauthorized non-demo users (this relaxation is scoped to the demo read-only viewer).

Do NOT add a submit-blocked confirmation copy line in the form body; a lock + disabled button is enough.

Target: mockup section 04 - https://claude.ai/code/artifact/f2e187b3-e8b2-4300-b701-d72bfebe3f3d

## Classification

**Type**: Frontend
**Layer**: Interface (connections + sync-jobs features)
[...]

PR #3283, opened 2026-09-16
feat(returns): approve dialog for operator-authored returns

### LINKED ISSUE #3083: feat(returns): approve dialog for operator-authored returns

jakubretajczykBD:
## Summary
- Adds `AuthorizeReturnDialog`: a pure confirm with no form fields (the write takes nothing beyond the return id), wired inline into the worklist's "Waiting for your OK" group — replacing that group's Link-to-detail action with a button that opens the dialog directly, since there is no form to defer to a detail-page destination for
- Satisfies the acceptance criterion structurally rather than procedurally: confirming calls `authorize`, and `useAuthorizeReturnMutation`'s own `onSettled` invalidates `returnsQueryKeys.all`, so the worklist's `bucket:'attributed'` scan refetches and the now-approved return no longer matches `authorizedAt === null` — no manual row-removal code is needed or present
- Both `'authorized'` and `'already-authorized'` are treated as success (the write is idempotent)
- The 409 `source-ingested` refusal — which the worklist's own predicate makes structurally unreachable from this entry point — is still handled defensively, since the dialog is exported from the barrel and may be reused elsewhere later

## Test plan
- [x] `tsc --noEmit` clean
- [x] `eslint` clean
- [x] `vitest run src/features/returns` — 274/274 passing (6 dialog + 1 worklist-integration test covering the removal criterion end-to-end)

Closes #3083


piotrswierzy (approved):
# PR Review — #3283 `feat(returns): approve dialog for operator-authored returns`

Reviewed against `docs/frontend-architecture.md`, `docs/frontend-ui-style-guide.md`, `docs/engineering-standards.md`, `docs/testing-guide.md`, and the in-feature baseline on `3082-match-return-dialog`.

---

## Positive observations

- **The removal criterion is satisfied structurally, and that is proved rather than asserted.** `authorize-return-dialog.tsx:22-27` explains that `useAuthorizeReturnMutation`'s `onSettled` invalidation is what removes the row, and `orphan-returns-worklist.test.tsx:246-283` backs it with a **stateful** `returns.list` mock so the assertion genuinely exercises the refetch rather than the mock. That is the difference between testing the wiring and testing the fixture.
- **`'authorized'` and `'already-authorized'` are both treated as success** (`authorize-return-dialog.tsx:78-83`), with a dedicated test. Correct: the write is idempotent, and special-casing the second outcome would tell an operator a lost race was a failure.
- **The `source-ingested` refusal renders as a distinct branch, not the generic error**, and the confirm form is removed in that state — asserted by `authorize-return-dialog.test.tsx:82-95`. This matches the `match-return-dialog.tsx` `alreadyAttributed` precedent exactly.
- **`readRecordRefusalReason`-style reason reading off `ApiError.details`, never the message string** (`authorize-error.ts:19-22`) — the `readMatchRefusalReason` discipline, correctly followed.
- The `renderAction` override on `WorklistGroup` is the minimal change: the "Needs an order" group keeps its `Link` and the docblock (`orphan-returns-worklist.tsx:28-43`) is updated to say why the two groups now differ, rather than being left describing the old behaviour.

---

## Findings

### **[IMPORTANT]** — `apps/web/src/features/returns/components/orphan-returns-worklist.tsx:225-235`

The new "Approve" row action is a **write affordance with no permission gate**. `docs/frontend-architecture.md § Access Control And UI Visibility` is explicit about which primitive each shape takes:

> | A write affordance — button, checkbox, form, menu item | `useWriteAccess` (`shared/auth/use-permission.ts`) + `ReadOnlyLock` (`shared/ui/read-only-lock.tsx`) | in demo mode: rendered **disabled** with a tooltip; otherwise hidden |

`POST /returns/:returnId/authorize` is `@Roles('admin', 'operator')`. A `viewer` or `packer` session reaches `/returns`, sees an enabled **Approve** button, clicks it, and gets a 403 that this dialog renders as `COPY.genericError` — *"The approval could not be completed. Try again."* That advice is false and unactionable: retrying cannot work.

The pattern is already established **inside this feature**, with the same permission:

- `return-decline-action.tsx:53,69-75,186-190` — `ReadOnlyLock` + a `writeAccess: { canWrite, demoReadOnly, visible }` prop resolved by the page via `useWriteAccess('orders:write', demoMode)`, and its docblock spells out the rule ("The one case where it is genuinely hidden is a session without write access, which is the house `useWriteAccess` policy").
- `apps/web/src/pages/inventory/inventory-locations-page.tsx:104,317-326` — the same treatment on a row action, with `inventory-locations-page.test.tsx:283-291` asserting the control is **absent** for a session with no write permission.

The demo-mode half matters here too: on a public demo this button is enabled and fires a real write attempt, which is precisely the case `ReadOnlyLock` (#1615) exists for.

The fix is fully precedented and needs no new plumbing — a `features/` component may import the `system` barrel directly, as `features/price-changes/components/price-changes-queue-table.tsx:52` already does:

```ts
import { useDemoMode } from '../../system';
import { useWriteAccess } from '../../../shared/auth/use-permission';
import { ReadOnlyLock } from '../../../shared/ui/read-only-lock';

const write = useWriteAccess('orders:write', useDemoMode());
// renderAction: write.visible ? (
//   <ReadOnlyLock active={write.demoReadOnly} message={COPY.approveReadOnly}>
//     <Button … disabled={!write.canWrite}>{COPY.needsApprovalAction}</Button>
//   </ReadOnlyLock>
// ) : <Link … />   // or omit the row action entirely
```

Note this is not a security hole — the doc is clear that "a permission drives UI visibility only" and the endpoint still enforces — which is why this is IMPORTANT rather than BLOCKING. But the same gap appears again on #3285's "+ Record a return" CTA, so it is worth fixing as one pass across the stack. Please also add the negative test (`queryByRole('button', { name: … })` is absent for a permission-less session), matching `inventory-locations-page.test.tsx:283`.

### **[SUGGESTION]** — `apps/web/src/features/returns/components/orphan-returns-worklist.tsx:236-248`

`<AuthorizeReturnDialog returnId={authorizingId} … />` carries no `key`. The mutation state (`mutation.isError`, and therefore `refused`) is component-local, and React reuses the same instance when `authorizingId` changes from `'r1'` to `'r2'` at the same position — a refused row would then paint the *next* row's dialog as already-refused.

This is unreachable today because Radix's modal overlay makes the rows behind inert, so nothing can move `authorizingId` without passing through `null` first. A `key={authorizingId}` makes that a structural guarantee rather than a consequence of the overlay, at the cost of one prop.

### **[SUGGESTION]** — `apps/web/src/features/returns/lib/authorize-error.ts:17-22`

This is now the **third** near-verbatim copy of the same routine in one `lib/` folder: `match-error.ts` (`readMatchRefusalReason`), `record-error.ts` (`readRecordRefusalReason`, #3284), and this one. All three do `instanceof ApiError` → narrow `details` → read `reason` → membership-test against a local `as const` tuple.

A single generic reader would collapse all three and keep the "unrecognised value degrades to `null`" discipline in one place:

```ts
export function readRefusalReason<T extends string>(
  error: unknown,
  values: readonly T[],
): T | null { … }
```

Not blocking — the duplication is small and each copy is correct — but it is the kind that drifts once a fourth write lands.

---

## Step 4 — Final Assessment

### Summary

A tight, well-reasoned slice. The dialog is correctly modelled as a pure confirm with no form, idempotent success is handled as success, the one refusal reason gets its own rendering rather than the generic error, and the acceptance criterion ("removes the row from the pending-approval group") is delivered by cache invalidation rather than array-splicing — and is *tested with a stateful mock*, which is what makes that claim credible. Docblocks are updated rather than left describing prior behaviour. The single substantive concern is that the new write affordance skips the house `useWriteAccess` + `ReadOnlyLock` treatment that the sibling write in this same feature already applies, which leaves a `viewer` session a button that answers 403 with "Try again."

### Merge Readiness

🔄 **Approve with changes** — merge after addressing the IMPORTANT item.

### Priority fixes

1. **Gate the "Approve" row action** with `useWriteAccess('orders:write', useDemoMode())` + `ReadOnlyLock`, matching `return-decline-action.tsx` and `inventory-locations-page.tsx`, and add the negative test asserting the button is absent without the permission. (`orphan-returns-worklist.tsx:225-235`)
2. *(optional)* Add `key={authorizingId}` to the dialog so per-row mutation state cannot leak across rows.
3. *(optional)* Collapse the three `read*RefusalReason` helpers into one generic reader.

piotrswierzy (approved):
# Re-review — PR #3283 (approve dialog for operator-authored returns)

Re-reviewed at `e6cedc7d` (`fix(returns): PR #3283 review fixes — gate the Approve action on orders:write`).

## 1. Prior findings

|---|---|---|---|
| 1 | **IMPORTANT** — the "Review and approve" row action is an ungated write affordance; `docs/frontend-architecture.md` § Access Control And UI Visibility row 2 mandates `useWriteAccess` + `ReadOnlyLock` | **Addressed** | `orphan-returns-worklist.tsx:196-197` (`useDemoMode()` + `useWriteAccess('orders:write', demoMode)`), `:261-283` (`renderAction={writeAccess.visible ? … : undefined}`, `<ReadOnlyLock active={writeAccess.demoReadOnly}>`, `disabled={writeAccess.demoReadOnly}`), copy at `orphan-returns-worklist.copy.ts:36-42` |
| 2 | **SUGGESTION** — per-row dialog state could leak across rows | **Addressed** | `orphan-returns-worklist.tsx:292` — `key={authorizingId}`, with a comment stating it is unreachable today (Radix overlay) and exists as a structural guarantee |
| 3 | **SUGGESTION** — collapse the three `read*RefusalReason` readers | **Not addressed, declined with a reason on the record** | commit body: the three files are owned by sibling stack branches; deferred to a generic-reader pass |

### Verification detail

The gate is grounded, not cosmetic:

- The permission matches the endpoint. `apps/api/src/returns/http/return-writes.controller.ts:166-167` is `@Roles('admin', 'operator')`, and `orders:write` is carried by both roles — so `useWriteAccess` alone is correct here and `useIsAdmin` (`use-permission.ts:63-84`, the narrow carve-out for admin-only routes that share a permission with a lesser role) is rightly **not** used.
- The three branches are all real. `useWriteAccess` returns `visible = canWrite || demoReadOnly` (`use-permission.ts:58-61`), so: write session → enabled button; demo viewer → disabled button + lock; neither → `renderAction` is `undefined` and `WorklistGroup` falls back to its default detail `Link` (`orphan-returns-worklist.tsx:154-162`). Navigation is correctly left ungated.
- **The tests can fail.** `renderWithProviders` defaults to `createNoopSessionAdapter()` (`test/test-utils.tsx:1240`), i.e. an anonymous session — so the two pre-existing tests that click the button had to be switched to `createAuthenticatedSessionAdapter()` (`orphan-returns-worklist.test.tsx:96-101`, `:423-426`) and would otherwise have gone red. That is the strongest available evidence the gate is live. The two new cases assert the `link` fallback with its `href` and the `button` absence (`:155-184`), and the disabled-inside-`.read-only-lock` demo case (`:186-212`).

## 2. Findings on the new commit

**[SUGGESTION]** — `apps/web/src/features/returns/components/orphan-returns-worklist.tsx:230` / `:249`

> Both groups pass `actionLabel={COPY.needs…Action}`, so the permission-less fallback renders a `Link` labelled **"Review and approve"** pointing at `/returns/:id` — a detail page that carries no approve affordance at all (the detail page only mounts `MatchReturnDialog`, per #3285's `return-detail-page.tsx:61`). A viewer can genuinely *review* there, so this is not false, but the verb "approve" promises something the destination does not offer. A distinct fallback label (e.g. "View return") would be more honest. Not worth blocking; the label is a single copy constant away.

**[SUGGESTION]** — `apps/web/src/features/returns/components/orphan-returns-worklist.test.tsx:186-212`

> The demo-viewer test stubs `apiClient.system.getConfig` through the `demoMode` option (`:95-99`). Consider also asserting the tooltip copy (`COPY.approveReadOnly`) is reachable, so the test covers the *explanation* and not only the disabled state — the explanation is the whole point of the `ReadOnlyLock` treatment per `frontend-architecture.md` § Access Control ("advertise, don't hide").

Nothing else in the new commit raises a concern: no raw `fetch()`, no new CSS, cross-feature import goes through the `features/system` barrel (`orphan-returns-worklist.tsx:69` — the `price-changes-queue-table.tsx:52` precedent, correctly cited), no `any`, and the copy lives in the `*.copy.ts` module as the feature's other strings do.

## 3. Summary

### Summary
The single IMPORTANT from the first pass is fully closed: the Approve row action now resolves `orders:write` through `useWriteAccess` + `ReadOnlyLock`, with the exact three-way behaviour `frontend-architecture.md` prescribes, the permission verified against the endpoint's own `@Roles('admin','operator')`, and — decisively — with tests that had to be given an authenticated session to keep passing, which proves the gate is enforced rather than merely present. The optional generic-reader refactor was declined with a stated reason rather than silently dropped.

### Merge Readiness
✅ **Approve**

### Priority fixes
None outstanding. Two optional polish items above.

**Stack note (not a finding on this PR):** `origin/3085-returns-entry-point` and `origin/3086-returns-flow-tests` were last pushed at 13:40 and do **not** contain this commit (`git merge-base --is-ancestor origin/3083-authorize-return-dialog origin/3085-returns-entry-point` → false). Both children carry their own pre-fix copies of this file and must be rebased onto this head before merge, or this gate is reverted downstream. Raised as BLOCKING on #3285.

piotrswierzy (approved):
## Tech Lead review — `feat(returns): approve dialog for operator-authored returns`

Clean, and the two changes that came out of the previous round — the write-access gate and the `key={authorizingId}` — are both correct for the reasons given.

### ✅ The permission gate picks the right permission, and I checked

`useWriteAccess('orders:write', demoMode)` against `POST /returns/:returnId/authorize` which is `@Roles('admin', 'operator')`. From `libs/core/src/users/domain/types/role.types.ts`:

- `admin` — `PermissionValues` (everything)
- `operator` — includes `'orders:write'`
- `viewer` — no `orders:write`
- `packer` — `[]`

So the FE gate and the backend guard cover exactly the same two roles. There is no `returns:*` permission in `PermissionValues` at all, so `orders:write` is the correct nearest proxy rather than a lazy pick — and worth noting, since a future `returns:write` would want this call site updated.

The three-way treatment follows `return-decline-action.tsx` precisely, and matches `docs/frontend-architecture.md` § Access Control:

| session | rendering |
|---|---|
| has `orders:write` | enabled Approve button |
| demo, no permission | Approve button **disabled** inside `ReadOnlyLock` — #1615 "advertise, don't hide" |
| no permission, not demo | falls back to the read-only detail `Link` |

Gating the write while leaving "Needs an order"'s `Link` ungated is right — that one is navigation to a read surface.

Without this, a `viewer` got an enabled button that answered 403 and rendered as *"The approval could not be completed. Try again."* — advice that cannot work, since retrying doesn't change a role. That's the kind of thing worth having caught.

### ✅ Both outcomes are success, and the AC is satisfied by invalidation rather than splicing

`'authorized'` and `'already-authorized'` both close and fire `onAuthorized` — correct, the write is idempotent and a lost race against a concurrent approval is not news the operator needs.

And the docblock's point about the acceptance criterion is the right architecture:

> *Confirming needs no manual row-removal here: `useAuthorizeReturnMutation`'s own `onSettled` invalidates `returnsQueryKeys.all`, so this group's own `bucket: 'attributed'` scan refetches and the now-approved return stops matching `authorizedAt === null` — the acceptance criterion is satisfied by cache invalidation, not by this component splicing an array.*

The stateful-list test proves it end to end rather than asserting the invalidation call, which is the difference between testing behaviour and testing implementation.

### ✅ No stale-mutation leak here

Unlike its sibling in #3281, this dialog is conditionally mounted (`{authorizingId !== null ? … : null}`) **and** carries `key={authorizingId}`, so mutation state is destroyed on close and a second row can never inherit the first's error. The `key` comment is honest that it's unreachable today (Radix's modal overlay makes the rows behind inert) and that it's there to make the guarantee structural rather than a consequence of the overlay. That's the right reason to add a `key`.

### ✅ Defensive 409 handling, correctly labelled as defensive

`isSourceIngestedAuthorizeRefusal` reads the body's `reason`, not the status or message — same convention as `readMatchRefusalReason` / `readBlockedTrigger`. The docblock states plainly that the worklist's own predicate makes this unreachable from the current entry point and that it exists because the component is exported from the barrel. That's the right justification for defensive code: name who could reach it.

The refused branch replaces the confirm form entirely rather than sitting beside a still-clickable Approve — and there's a test asserting the confirm button is gone, plus one asserting the dialog does **not** close on refusal.

### 🟢 SUGGESTION — the no-permission fallback Link is labelled "Review and approve"

`orphan-returns-worklist.tsx` — when `writeAccess.visible` is false the row falls back to the default `Link` carrying `actionLabel`, which for this group is `COPY.needsApprovalAction` = **"Review and approve"**. A `viewer` therefore sees a link promising an action their session cannot perform, and the detail page it lands on will (correctly) not offer it.

"Review" alone would be accurate for both audiences, or the fallback could take its own label. Small, but it's the same class of honesty the rest of this file is careful about.

### Notes

- Copy fully in `lib/authorize-return-dialog.copy.ts`, and it takes care never to imply approval *gates* restock or refund — matching §5.5, where approval is an audit stamp. The `approveReadOnly` tooltip added to the worklist copy is worded for this row rather than borrowed.
- `renderAction` as an optional per-group override, defaulting to the `Link`, is a tidy way to let one group diverge without forking `WorklistGroup`.
- The existing worklist tests were updated to `createAuthenticatedSessionAdapter()` with a comment saying why, rather than silently changed — good, that's the kind of edit that otherwise reads as unrelated churn.
- CI green.

**Verdict:** ✅ Approve. The suggestion is a copy tweak.

Issue #2332, opened 2026-08-22
[TASK] CORE — orphan return bucket, downstream-trigger block and re-attribution reconcile


piotrswierzy:
> Part of the OMS programme backlog (**Wave 1c**, stream S2, size M). Backlog slug: `W1c-5`. Sibling `W*` slugs resolve via the Wave-1c epic's checklist. Design of record: ADR-060.

## Problem / Context

ADR-060 makes `internalOrderId` nullable **on purpose**: a return can arrive for an order OL has not ingested (or cannot attribute). Such a return must persist, must be visible, and must **block every downstream trigger** — an unattributed return that silently drove a restock or a credit note would be exactly the failure the nullable column exists to prevent.

## Proposed Solution

- A `ReturnRecord.isOrphan` predicate (derived from `internalOrderId IS NULL`) plus an aggregate count and a `?bucket=orphan` filter on the read path — following #2100's attention-worthy / routine split rather than inventing a new taxonomy.
- A guard in the returns service refusing every downstream trigger (Wave 2's restock, refund trigger and correction proposal all call it) while orphaned, with a named domain error.
- A background reconcile pass that re-attributes an orphan once the order is ingested, keyed on the source order id already stored on the record. Best-effort, logged, never fatal.

## Classification

**Type**: CORE
**Layer**: Application
**File(s)**: `libs/core/src/returns/application/services/`, `.../domain/entities/`, `apps/worker/src/**` for the reconcile pass

## Dependencies

- #2328 (`W1c-2`)

## Assumptions

- Re-attribution is keyed on the source's own order reference, which both Allegro and Erli carry on a return; where a source carries none, the return stays orphaned and the operator resolves it.

## Acceptance Criteria

- [ ] An unattributable return persists with `internalOrderId = null` and appears in the orphan count
- [ ] Every downstream trigger raises the named domain error while orphaned (spec per trigger seam, including the Wave-2 seams that exist only as interfaces today)
- [ ] The reconcile pass re-attributes once the order exists and is a no-op otherwise
- [ ] The reconcile pass never fails ingestion
- [ ] Tests added or updated for non-trivial logic
- [ ] No architecture boundary violations (CORE ↔ Integration)

PR #3284, opened 2026-09-16
feat(returns): record-a-return-by-hand dialog

### LINKED ISSUE #3084: feat(returns): record-a-return-by-hand dialog
## Problem / Context

No UI exists to open an operator-authored return for a channel with no returns feed.

## Proposed Solution

Build a dialog collecting `internalOrderId`, `sourceConnectionId` (validated against the order's actual mapping) and a lines list (sku/name/reason/quantityAdvised/note), per `RecordReturnDto`.

## Classification

**Type**: Frontend
**File(s)**: apps/web/src/features/returns/components/record-return-dialog.tsx, record-return-dialog.schema.ts

## Dependencies

- Part of #3078

## Assumptions

- None

## Acceptance Criteria

- [ ] Order/channel mismatch (400 `order-not-on-connection`) renders as a field error
- [ ] Successful record adds the return to the "Waiting for your OK" group

jakubretajczykBD:
## Summary
- Adds `RecordReturnDialog` + `record-return-dialog.schema.ts`: a form collecting `internalOrderId`, `sourceConnectionId`, and a single line (item name, reason, quantity) per `RecordReturnDto` — single line only, matching the reviewed mockup and every acceptance criterion
- Order field: the same bounded free-text + `<datalist>` shape `match-return-dialog.tsx` already ships (no order-search endpoint exists, no buyer name on the FE contract)
- Connection field: a plain `<select>` over `useConnectionsQuery()`, matching `returns-list-page.tsx`'s existing source filter
- A proactive, non-blocking warning when the picked order's OWN `sourceConnectionId` disagrees with the picked connection — stronger than the mockup's coarse channel-string heuristic, since it compares the actual mapping ground truth
- 400 `unknown-order` → field error on the order field (naming what was typed); 400 `order-not-on-connection` → field error on the connection field — satisfying the acceptance criterion exactly
- Zod validation mirrors `RecordReturnDto`'s bounds (maxlength, `IsInt`, `Min(1)`), so an obviously invalid submission never round-trips
- Includes two independent post-hoc fixes found by `pnpm -r build` (`tsc -b`) after this PR's own diff: a pre-existing `LoadingState` missing its required `message` prop in the worklist, and a Zod v3-only `invalid_type_error` option that Zod v4 removed

## Test plan
- [x] `tsc --noEmit` and `tsc -b` (full project-references build) both clean
- [x] `eslint` clean
- [x] `vitest run src/features/returns` — 282/282 passing
- [x] Both cross-feature imports (`connections`, `orders`) go through their barrels

Closes #3084


piotrswierzy (changes requested):
# PR Review — #3284 `feat(returns): record-a-return-by-hand dialog`

Reviewed against `docs/frontend-architecture.md`, `docs/engineering-standards.md`, `docs/architecture-overview.md § 22 Returns`, and the backend contract in `apps/api/src/returns/dto/return-write.dto.ts` / `libs/core/src/returns/application/services/return-custody.service.ts`.

---

## Positive observations

- **The order field reuses the exact bounded free-text + `<datalist>` shape** `match-return-dialog.tsx` already ships, including the `autoComplete="off"` rationale, rather than inventing a second answer to "no order-search endpoint exists".
- **The connection-mismatch check compares the order's own `sourceConnectionId`** (`record-return-dialog.tsx:139-142`), not a channel string — a genuine improvement on the mockup, and correctly a **warning, not a block**, with `#2332`'s "return arrived via a different connection" cited as the legitimate case. The test asserts the confirm button stays enabled (`record-return-dialog.test.tsx:157-158`), which is the right property to pin.
- **Both field-mapped refusals are tested against the reason field, never the message string** (`record-error.ts`, `record-return-dialog.test.tsx:110-131`), and an unrecognised failure degrades to the generic sentence.
- `reason` is validated by `refine` against the mirrored `RETURN_LINE_REASON_VALUES` rather than `z.enum`, with the reason stated in the schema docblock — consistent with `isReturnLineReason`'s existing posture.

---

## Findings

### **[BLOCKING]** — `apps/web/src/features/returns/components/record-return-dialog.tsx:186-230` and `record-return-dialog.schema.ts:32-52`

**The form collects no `sku`, so every return recorded through this dialog can never be restocked.**

`ReturnCustodyService.resolveRestockTarget` resolves the restock target *by sku* and refuses outright when there isn't one — `libs/core/src/returns/application/services/return-custody.service.ts:687-693`:

```ts
const sku = line.sku?.trim() ?? '';
if (sku === '') {
  … 'unresolved-product',
    'this return line carries no sku, so OpenLinker cannot tell which product to restock'
```

`docs/architecture-overview.md § 22 Returns` (#2370) states the same rule from the other side: *"The restock target is resolved **by sku through `IProductsService`** — a return line carries no product id and cannot."* And `#3171`'s order-line resolution, the only other route by which a line acquires provenance, runs **only** on the `marketplace.return.sync` per-return path, which an `operator_authored` return never takes.

So the shipped operator journey is: record a return by hand → receive the parcel → press **Restock** → permanently blocked with `unresolved-product`. There is **no line-edit write** in the returns API, so this is unrecoverable inside the product; the operator's only remedy is the manual attestation, which explicitly *"never writes stock or claims OL did"*.

This is not a contract change. The field is already optional on both halves:

- `apps/api/src/returns/dto/return-write.dto.ts:90-94` — `sku?: string | null`, `@MaxLength(255)`
- `apps/web/src/features/returns/api/returns.types.ts:662-668` — `RecordReturnLineInput.sku?: string | null`

and issue #3084's own Proposed Solution names it: *"a lines list (**sku**/name/reason/quantityAdvised/note)"*.

The fix is one optional field, one Zod entry (`z.string().trim().max(255).optional()`), and one property on the submitted line. Given the restock consequence I'd argue it should be **required** in this dialog even though the DTO makes it optional — the DTO must stay lenient for the ingestion path, but an operator opening a return by hand is exactly the caller who *can* supply it, and a line without one is a dead end. At minimum, if it is left optional, the field description must say that restock is unavailable without it.

Note the dialog currently makes `name` **required** (`record-return-dialog.schema.ts:38-42`) while the backend has it optional — so the one field that is functionally load-bearing was dropped and the one that is decorative was made mandatory.

### **[IMPORTANT]** — `apps/web/src/features/returns/components/record-return-dialog.tsx:255-259`

The reason `<select>` renders the raw wire vocabulary:

```tsx
{RETURN_LINE_REASON_VALUES.map((reason) => (
  <option key={reason} value={reason}>{reason}</option>
))}
```

An operator picks from `withdrawal` / `defective` / `not_as_described` / `wrong_item` / `other`. A labeller for exactly this vocabulary already exists, is exported from the `orders` barrel, and is **already imported by a sibling file in this same feature**:

- `apps/web/src/features/orders/lib/refund-reason.ts:36-55` — `REFUND_REASON_LABELS` (`not_as_described` → *"Item did not match the listing"*) + `describeRefundReason`, which falls back to the raw value for a reason this build predates.
- `apps/web/src/features/returns/components/return-money-panel.tsx:36,69` — `import { describeRefundReason } from '../../orders'` and renders it.

`docs/architecture-overview.md § 22 Returns` records that `ReturnLine.reason` reuses `RefundReason` **verbatim** *"precisely so returns-by-reason and refunds-by-reason report on ONE axis"* — so the two vocabularies are the same vocabulary, and one of the two already has labels.

Use `<option value={reason}>{describeRefundReason(reason)}</option>`. Two things worth knowing about why this needs catching in review: `check-ui-vocabulary.mjs` cannot see it (its own docblock, limitation 2: *"Only literal strings. A term assembled at runtime … is invisible"*), and `return-lines-table.tsx:166` renders `line.reason` raw too — so the baseline is genuinely mixed. But a **read-only cell** showing a raw value the operator can quote in a ticket is a different thing from a **dropdown the operator chooses from**, where the label is the entire affordance.

### **[IMPORTANT]** — `apps/web/src/features/returns/components/record-return-dialog.tsx:134-137, 240-250`

Neither supporting query has a loading or error state:

```ts
const ordersQuery = useOrdersQuery(undefined, { limit: RECENT_ORDERS_LIMIT });
const orders = ordersQuery.data?.items ?? [];
const connectionsQuery = useConnectionsQuery();
const connections = connectionsQuery.data ?? [];
```

For the orders `<datalist>` this degrades acceptably — free text is still honoured and the docblock says so. For the **connection `<select>` it does not**: `sourceConnectionId` is required, the only way to satisfy it is to pick an option, and a failed `connections.list` leaves the operator with nothing but `COPY.connectionPlaceholder` and no explanation. They fill the whole form, submit, get the Zod message *"Select the channel this return came in on."* about a list that is empty for a reason the UI never states.

`docs/frontend-architecture.md § Async UX Conventions`: *"every list and detail screen should render loading, empty, error, and success-aware states deliberately."* This is also the same principle `docs/architecture-overview.md` records for the returns surfaces — an empty array read back from a failed call must never be presented as a positive claim.

Minimum: when `connectionsQuery.isError` (or it settled with zero connections), render an `Alert tone="error"` in place of / beside the select saying the channel list could not be loaded, and disable submit. A `loading` placeholder option while `isPending` would also stop the select looking empty-but-fine on a slow first paint.

### **[SUGGESTION]** — `apps/web/src/features/returns/components/record-return-dialog.tsx:126-128, 166-175`

Server field errors are held in a parallel `useState<{field, message}>` and then merged with RHF's own errors at each render:

```tsx
error={fieldError?.field === 'internalOrderId' ? fieldError.message : form.formState.errors.internalOrderId?.message}
```

RHF's `form.setError('internalOrderId', { message })` is the documented mechanism for exactly this — `docs/frontend-architecture.md § Form State`: *"server validation errors should be mapped back to fields where practical"* — and it removes the manual clear-on-change plumbing (`onChange: () => { if (fieldError?.field === …) setFieldError(null) }`), the two-source merge at each field, and the risk of the two disagreeing. Worth a follow-up; the current code is correct, just carrying state RHF already offers.

### **[SUGGESTION]** — `record-return-dialog.schema.ts:45-48` (comment accuracy)

> *"An empty/non-numeric input coerces to `NaN`, which `.int()` refuses"*

`Number('') === 0`, not `NaN`, so an emptied number field is caught by `.min(1)` rather than by `.int()`. The behaviour is right; the comment describes a different path than the one that fires, and a later reader may remove `.min(1)` believing `.int()` covers it.

### **[SUGGESTION]** — `record-return-dialog.tsx` (scope)

`note` is also dropped, and issue #3084's Proposed Solution names it alongside `sku`. Unlike `sku` it has no downstream consequence, so deferring it is defensible — but the PR body says "single line only, matching the reviewed mockup and every acceptance criterion" without noting that two *fields* named in the issue are absent. Worth stating explicitly in the PR body so the omission is a decision on the record rather than something a later reader has to rediscover.

### **[SUGGESTION]** — `apps/web/src/features/returns/api/returns.types.ts:253-259` (pre-existing, newly load-bearing)

`RETURN_LINE_REASON_VALUES` is now the **third** copy of the same five strings — core's `RefundReasonValues`, `features/orders/lib/refund-reason.ts`'s `REFUND_REASON_VALUES`, and this one. `refund-reason.ts`'s own docblock argues at length that *"the whole point of one home is that they cannot spell the vocabulary two ways"* and that it lives in `orders` because orders owns the concept. This PR is the first thing to make the returns copy user-visible. Not this PR's to fix, but if the IMPORTANT above is taken by importing `describeRefundReason`, consider importing `REFUND_REASON_VALUES` from the same barrel and deleting the returns copy.

---

## Step 4 — Final Assessment

### Summary

The dialog is carefully built — the refusal→field-error mapping is exactly right, the connection-mismatch warning is a genuine improvement on the mockup and correctly non-blocking, and the reuse of the `match-return-dialog` order-field shape keeps one answer to a problem that has one. But it ships a write whose output is structurally dead-ended at the next step of the workflow the epic exists to enable: no `sku` is collected, and `ReturnCustodyService` refuses every restock for a line without one, with no line-edit write anywhere in the API to repair it afterwards. That has to be fixed before merge. Beside it, the reason dropdown shows raw wire values when a labeller for that exact vocabulary is already imported by a sibling file in this feature, and a failed connections read leaves the required field unsatisfiable with no explanation.

### Merge Readiness

❌ **Request changes**

### Priority fixes

1. **Add `sku` to the form and the submitted line** (`record-return-dialog.tsx`, `record-return-dialog.schema.ts`). Without it every recorded return is permanently un-restockable (`return-custody.service.ts:687-693`, `unresolved-product`). Strongly consider making it required here even though the DTO allows null; if it stays optional, say in the field description that restock is unavailable without it.
2. **Label the reason options** via `describeRefundReason` from the `orders` barrel, as `return-money-panel.tsx:36,69` already does — an operator should not be picking `not_as_described` from a dropdown.
3. **Handle `connectionsQuery` failure/empty** — the required connection select is currently unsatisfiable with no explanation when the read fails (`§ Async UX Conventions`).
4. *(optional)* Move server field errors onto RHF's `setError` instead of the parallel `useState`.
5. *(optional)* Fix the `Number('')` comment in the schema; state the `sku`/`note` omission in the PR body.

piotrswierzy (approved):
# Re-review — PR #3284 (record-a-return-by-hand dialog)

Re-reviewed at `0c3df007` (`fix(returns): PR #3284 review fixes — required sku, labelled reasons, connections error state`).

## 1. Prior findings

|---|---|---|---|
| 1 | **BLOCKING** — the form collects no `sku`, so every return recorded through it is permanently un-restockable (`return-custody.service.ts:687-693` resolves the restock target by sku and refuses `'unresolved-product'` without one; no line-edit write exists to repair it) | **Addressed** | `record-return-dialog.schema.ts:44-48` (`sku` required, trimmed, `.min(1)`/`.max(255)`), `:78` (default value), `record-return-dialog.tsx:143` (`sku: values.sku` on the submitted line), `:246-258` (the field), copy at `record-return-dialog.copy.ts:36-39`. Wire type already carried it (`returns.types.ts:680`) and `@MaxLength(255)` on `RecordReturnLineDto.sku` (`apps/api/src/returns/dto/return-write.dto.ts:93`) matches the FE bound exactly. |
| 2 | **IMPORTANT** — the reason `<select>` renders raw wire values | **Addressed** | `record-return-dialog.tsx:69` imports `describeRefundReason` from the `orders` barrel (the `return-money-panel.tsx` precedent), `:285` renders `{describeRefundReason(reason)}`. `refund-reason.ts:48-56` falls back to the raw value for a reason this build predates, so an added member degrades readably rather than blank. |

### Verification detail

- The `sku` docblock states *why* the FE is stricter than the DTO (`record-return-dialog.schema.ts:20-29`): the DTO stays lenient for the ingestion path where a source may genuinely report none, while an operator opening a return by hand is exactly the caller who can supply it. That is the right asymmetry and it is on the record.
- The `sku` test can fail: `record-return-dialog.test.tsx:121-132` fills every *other* field validly and leaves `sku` blank, then asserts `record` was not called — so it is the sku requirement being exercised, not an incidental invalid form. The happy-path assertion was updated to the exact submitted line including `sku` (`:96-104`), and `fillValidForm` now types it (`:69`).
- The label test asserts both directions (`:133-141`): the labelled option exists with `value="not_as_described"`, **and** no option is named by the raw wire string. That is the shape that cannot pass vacuously.

Two further fixes were volunteered beyond the review, and both are sound:

- **Connections read failure** (`record-return-dialog.tsx:105-108`, `:226-240`, `:309`). The connection `<select>` is required and satisfiable only by a fetched option, so a failed read previously left it unsatisfiable with nothing on screen. Now: `role="alert"` Alert (`shared/ui/alert.tsx:30` renders `role="alert"` for `tone="error"`), select disabled, submit disabled, loading placeholder while pending. The docblock correctly distinguishes this from the orders `<datalist>`, where free text is still honoured. Test at `:143-156` asserts all three.
- **`form.setError` replacing the parallel `useState`** (`:157-163`, `:196`, `:226`). Correct: the form is submitted by the time a server error arrives, so RHF's default `reValidateMode: 'onChange'` re-runs the resolver on the next keystroke and clears the manually-set error — reproducing the hand-rolled `onChange` clear that was deleted, via the framework's own mechanism. Matches the cited `location-dialog.tsx` precedent.

**[SUGGESTION]** — `apps/web/src/features/returns/components/record-return-dialog.tsx:246-258`

> `sku` is now required but not *validated against the catalogue* — an operator can type a SKU OpenLinker does not know, and the restock still refuses later with `'unresolved-product'`, just less often. `GET /variants/search` exists (see `engineering-standards.md` § When A Paginated Total Is Expensive, which enumerates it among the split-read routes), so a `<datalist>` typeahead on the same shape as the order field (`:206-220`) would close the residue. Out of scope for this PR; worth a follow-up issue rather than a change here.

**[SUGGESTION]** — `apps/web/src/features/returns/components/record-return-dialog.tsx:194-196`

> The connections Alert is rendered after the `FormField` and the select is `disabled`, so a keyboard user tabbing through the form skips the control entirely and never reaches the explanation (a `role="alert"` div is announced on mount, but not on a later tab pass). Consider surfacing the same sentence as the field's `error` prop as well, so the explanation is attached to the control via `aria-describedby` rather than only adjacent to it.

**[SUGGESTION]** — `record-return-dialog.tsx:36-44`

> The docblock now states that `note` is deliberately not collected. Good — that turns an omission named in #3084's Proposed Solution into a decision on the record. Consider filing the follow-up so the "rather than blocking this PR" half has an owner.

No `any`, no raw `fetch()`, no inline hex, dependency direction is `features → shared` plus cross-feature barrel imports only (`:68-69`), and the schema mirror comment about `Number('')` being `0` rather than `NaN` (`:59-65`) is now correct.

## 3. Summary

### Summary
The BLOCKING defect is genuinely fixed in the code, not merely claimed: `sku` is a required form field, reaches the submitted line, matches the backend's own `@MaxLength(255)`, and is covered by a test that fills every other field validly so the assertion is specifically about sku. The reason `<select>` now renders `describeRefundReason` labels with a test asserting the raw value is *not* visible. Two unprompted fixes (connections-read failure state, `form.setError`) are both correct and improve the dialog beyond what was asked. Remaining concerns are all optional.

### Merge Readiness
✅ **Approve**

### Priority fixes
None outstanding. Three optional items above; the SKU typeahead is the one worth a follow-up issue.

**Stack note (not a finding on this PR):** `origin/3085-returns-entry-point` does not contain this commit — it carries its own pre-fix copy of `record-return-dialog.schema.ts` with no `sku` at all. Merging #3285 without rebasing it onto this head reverts the BLOCKING fix. Raised as BLOCKING on #3285, and #3286's flow test will fail once this requirement lands (it never types a SKU).

piotrswierzy (approved):
## Tech Lead review — `feat(returns): record-a-return-by-hand dialog`

The biggest of the three dialogs and the one with real form state. The required-`sku` decision from the previous round is the most valuable thing in here.

### ✅ Requiring `sku` — right call, right justification

The schema docblock earns its length:

> *`ReturnCustodyService.resolveRestockTarget` resolves the restock target BY SKU and refuses outright with `unresolved-product` when a line carries none — and there is no line-edit write anywhere in the returns API, so a line recorded without one is permanently un-restockable.*

That's the correct reason to be **stricter than the DTO**: `RecordReturnLineDto.sku` is optional because the *ingestion* path has a source that may genuinely report none, whereas an operator typing the form is precisely the caller who can supply it. Deliberately being stricter than the wire needs the asymmetry named, and it is — including why the DTO stays lenient. The field's own description carries the consequence in operator language ("Without one, this return's units can never be added back to stock"), which is better than a generic "required".

The dead end is real and unrecoverable given no line-edit endpoint exists, so closing it at the only place it can be closed is right.

### ✅ The connection-mismatch check is stronger than the mockup and correctly non-blocking

Comparing the picked order's own `sourceConnectionId` against the picked connection — rather than the mockup's coarse `channel` string — makes the warning authoritative instead of heuristic, and the docblock says so rather than quietly diverging from the committed mockup.

Keeping it a warning is the right half of that:

> *a return legitimately arriving via a different connection than the one that placed the order is one of the documented orphan causes (#2332), so the operator may know something this proactive check cannot*

That's the correct place to draw the line between "worth a second look" and "refuse". The test asserts the warning renders **and** that submit stays enabled.

### ✅ Refusals mapped through `form.setError`, not a parallel `useState`

Two of the four `ReturnRecordRefusedError` reasons become field errors on the field that can fix them (`unknown-order` → order, `order-not-on-connection` → connection); the other two fall through to generic, with the docblock stating they should already be unreachable past Zod. Using RHF's own error channel rather than a second state atom means the errors clear on edit for free and live next to the input — the mechanism the framework already provides.

`readRecordRefusalReason` reads the body's `reason`, same convention as its two siblings, and returns `null` for anything unrecognised.

### ✅ The failed-connections-read state

```ts
const connectionsUnavailable = connectionsQuery.isError;
```
…disables the `<Select>`, disables submit, and renders an explanation. The comment draws the right distinction with the orders `<datalist>`, where free text is still honoured so a failed read degrades acceptably on its own. Without this, an operator fills five fields and then learns something is wrong from a validation message about a list whose state they never saw. Good catch, and tested.

### 🟢 SUGGESTION — the select iterates one vocabulary and labels from another

`record-return-dialog.tsx`:

```tsx
{RETURN_LINE_REASON_VALUES.map((reason) => (
  <option key={reason} value={reason}>{describeRefundReason(reason)}</option>
))}
```

`RETURN_LINE_REASON_VALUES` (`features/returns`) and `REFUND_REASON_VALUES` (`features/orders`, which backs `REFUND_REASON_LABELS` / `describeRefundReason`) are two separate five-element copies of the same core `RefundReasonValues`. They agree today — I diffed them, both are `withdrawal, defective, not_as_described, wrong_item, other`.

This PR is the first place that consumes **both at once**, which makes the duplication load-bearing in a way it wasn't before: if a member is added to one copy and not the other, this select silently renders a raw wire string as its visible label. `describeRefundReason` degrades gracefully by design, so it fails soft rather than blank — but soft-failing into `not_as_described` on an operator-facing dropdown isn't great.

Cheapest fix is to iterate `REFUND_REASON_VALUES` — same barrel as the labeller, so the list and its labels cannot separate. The returns-side copy stays for the chips. (Note `returns.types.ts:198` already carries a comment promising there won't be "two FE copies" of this axis; there are two, and this is a good moment to say so out loud.)

### 🟢 SUGGESTION — `mutation.reset()`, same as #3281

`resetAndClose` calls `form.reset(...)` but leaves `mutation.error` set. If #3085 mounts this dialog persistently (rather than conditionally, as the worklist does for the authorize dialog), a generic failure survives close/reopen and greets the operator on a fresh form. Either add `mutation.reset()` here, or have the caller mount it conditionally — worth settling the same way across all three dialogs.

### Notes

- Single-line scope is stated in two places (schema + component docblock) and tied to the mockup and the ACs, so it reads as a decision rather than an omission. Same for `note` being deferred — explicitly recorded rather than silently dropped.
- The `z.coerce.number()` comment about `Number('') === 0` and Zod 4 dropping `invalid_type_error` is exactly the kind of note that stops someone "simplifying" the validation chain and losing the empty-field message.
- `autoComplete="off"` applied consistently, including on `itemName` which has no datalist — with the reasoning restated.
- Eleven tests including both field-error paths, the labeller assertion, the connections-failure state, and the non-blocking mismatch warning.
- CI green.

**Verdict:** ✅ Approve. Both suggestions are small and neither blocks.

PR #3285, opened 2026-09-16
feat(returns): entry point from the returns list/detail

### LINKED ISSUE #3085: feat(returns): entry point from the returns list/detail

jakubretajczykBD:
## Summary
- Mounts the #3081 orphan-returns worklist and the #3084 record dialog directly on `ReturnsListPage` — no new route. The worklist renders above the segment strip ("what needs me right now" before "show me a filtered page"), and a "+ Record a return" page action opens the dialog, invalidating the list's own query on success so a newly recorded return appears immediately
- Satisfies the acceptance criterion literally: the worklist is reachable from the existing Returns nav item with no direct URL
- **Also closes a real gap found while manually verifying the flow**: `MatchReturnDialog` (#3082) was built, tested and exported, but nothing ever rendered it — the worklist's own row action deliberately deferred to the return detail page, but no detail-page entry point had actually been wired. `ReturnOrphanBanner` gains an optional `action` slot and the detail page fills it with a "Match to an order" button (gated on `orders:write`, same `ReadOnlyLock` treatment as the decline action) that opens `MatchReturnDialog`. Corrects the banner's own docblock, which predated the epic and said the button was deliberately absent — true before #2372/#2376 had a caller, false now
- Two independent `autoComplete="off"` fixes on the order/item fields in both dialogs — the native browser history dropdown was competing with the intentional `<datalist>` and, inside a dialog, forcing the page to scroll ("the field runs away")

## Test plan
- [x] `tsc --noEmit` clean, full `pnpm --filter @openlinker/web build` (`tsc -b` + `vite build`) clean
- [x] `eslint` clean
- [x] `vitest run src/features/returns src/pages/returns` — 349/349 passing (3 new tests: worklist mounted, record dialog opens, match dialog opens from the banner)

Closes #3085


piotrswierzy (approved):
# PR Review — #3285 `feat(returns): entry point from the returns list/detail`

Reviewed against `docs/frontend-architecture.md`, `docs/frontend-ui-style-guide.md`, and the `inventory-locations-page` / `return-decline-action` baselines.

---

## Positive observations

- **Finding and closing the dead-dialog gap is the best thing in this PR.** `MatchReturnDialog` (#3082) was built, tested, exported and rendered by nothing. The PR body says this was caught *while manually verifying the flow* — which is exactly `docs/lessons.md § "is this MOUNTED?"` working as intended, and is worth more than the entry point it was scoped for.
- **The banner docblock is corrected, not left stale.** `return-orphan-banner.tsx:12-19` explicitly records that the old "there is deliberately no `Match to an order` action" rationale was *true before #2372/#2376 had a caller, false now*. Same for `return-detail.copy.ts:67-70`. A sentence that has become false is worse than no sentence, and this PR treats it that way.
- **The `action?: ReactNode` slot degrades correctly** — a page that has not wired the dialog gets the pre-#3078 explanation-only banner rather than a broken control, and `Alert` already carries an `action` prop (`shared/ui/alert.tsx:13,35`), so no shared primitive changed.
- **The match button is correctly gated** (`return-detail-page.tsx:289-306`): `writeAccess.visible ?` around it, `ReadOnlyLock active={writeAccess.demoReadOnly}`, `disabled={!writeAccess.canWrite}`, with copy for the locked message. This is the house pattern applied exactly.
- The acceptance criterion is met literally — no new route, the worklist reachable from the existing Returns nav item — and both mount points get a test (`returns-list-page.test.tsx:509-525`, `return-detail-page.test.tsx:229-240`).

---

## Findings

### **[IMPORTANT]** — `apps/web/src/pages/returns/returns-list-page.tsx:241-249`

The **"+ Record a return" page action is not permission-gated**, three files away from a match button in this same diff that is gated correctly.

```tsx
actions={
  <Button onClick={() => { setIsRecordDialogOpen(true); }}>
    {RECORD_RETURN_DIALOG_COPY.triggerLabel}
  </Button>
}
```

`POST /returns/record` is `@Roles('admin', 'operator')`. A `viewer` or `packer` sees the CTA, fills the whole form (order id, channel, item, reason, quantity), submits, and gets a 403 that `RecordReturnDialog` renders as *"The return could not be recorded. Try again."* — advice that cannot work.

`docs/frontend-architecture.md § Access Control And UI Visibility` assigns this shape a specific primitive:

> | A write affordance — button, checkbox, form, menu item | `useWriteAccess` + `ReadOnlyLock` | in demo mode: rendered **disabled** with a tooltip; otherwise hidden |

The closest precedent is the same shape on the same slot — a `PageLayout` `actions` CTA opening a create dialog:

- `apps/web/src/pages/inventory/inventory-locations-page.tsx:104` — `const write = useWriteAccess('inventory-locations:write', demoMode);`
- `apps/web/src/pages/inventory/inventory-locations-page.test.tsx:283-291` — *"hides Edit/Retire/Delete row actions and Add location for a session with no write permission"*, asserting `queryByRole('button', { name: '+ Add location' })` is absent.
- `apps/web/src/pages/inventory/inventory-locations-page.test.tsx:292-301` — the positive counterpart, under `createAuthenticatedSessionAdapter()`.

This page **already resolves nothing it needs** — unlike `return-detail-page.tsx`, `returns-list-page.tsx` imports neither `useDemoMode` nor `useWriteAccess`, so the fix is three imports plus the wrap the detail page in this same PR already demonstrates:

```tsx
const write = useWriteAccess('orders:write', useDemoMode());
…
actions={write.visible ? (
  <ReadOnlyLock active={write.demoReadOnly} message={DEMO_READ_ONLY_ACTION_MESSAGE}>
    <Button disabled={!write.canWrite} onClick={…}>{RECORD_RETURN_DIALOG_COPY.triggerLabel}</Button>
  </ReadOnlyLock>
) : undefined}
```

Please add the negative test alongside, matching `inventory-locations-page.test.tsx:283`. The same gap exists on #3283's "Approve" row action, so this is best fixed as one pass across the stack.

### **[SUGGESTION]** — `apps/web/src/pages/returns/returns-list-page.tsx:237-461`

The fragment wrapper was added without re-indenting the subtree, so `<PageLayout>` sits at the same indentation level as its parent `<>` and every child is now one level shallow relative to its actual nesting. `eslint` passes because indentation is Prettier's concern, not the linter's — but `docs/engineering-standards.md § ESLint & Prettier Configuration` sets `tabWidth: 2`, and the file will churn the whole block the next time anyone runs Prettier over it, burying a real change in a formatting diff. Worth re-indenting now while the diff is small.

### **[SUGGESTION]** — `apps/web/src/pages/returns/returns-list-page.tsx:454-460` vs `return-detail-page.tsx:307-317`

The two dialogs are mounted inconsistently: `MatchReturnDialog` is a child of `PageLayout` on the detail page, `RecordReturnDialog` is a sibling *outside* it on the list page — which is what forces the fragment above. Both render into a portal, so there is no behavioural difference; putting `RecordReturnDialog` inside `PageLayout` removes the fragment, the re-indent and the inconsistency in one move.

### **[SUGGESTION]** — `returns-list-page.tsx:457-459` and `return-detail-page.tsx:312-314`

```tsx
onRecorded={() => { void query.refetch(); }}
onMatched={() => { void query.refetch(); }}
```

Both mutations already invalidate the relevant keys — `useRecordReturnMutation` invalidates `returnsQueryKeys.detail(result.returnId)` + `returnsQueryKeys.all` on success, `useMatchReturnToOrderMutation` the same on settle. An explicit `refetch()` on top fires a second request for data TanStack Query is already refetching. `docs/frontend-architecture.md § Server State` says *"mutations invalidate or update the Query cache"* — the invalidation is the mechanism; the callback should be for things the cache cannot express (closing a dialog, navigating), not for re-triggering a read that is already in flight.

### **[SUGGESTION]** — `apps/web/src/pages/returns/return-detail-page.tsx:288, 307`

`detail.bucket === 'orphan'` is tested twice in adjacent JSX blocks. Folding the dialog into the first block (it renders nothing while `open` is false) removes the second condition and the possibility of the two drifting.

### **[SUGGESTION]** — `apps/web/src/pages/returns/return-detail-page.test.tsx:229-240`

The new test covers the write-access-present path only. Given the gating is the interesting part of this change, a companion asserting the match button is **absent** for a session without `orders:write` would pin it — the shape `inventory-locations-page.test.tsx:283-291` uses. Without it, a future refactor that drops the `writeAccess.visible ?` wrapper stays green.

---

## Step 4 — Final Assessment

### Summary

A small, well-targeted PR that does more than its title: it mounts the worklist and the record dialog with no new route as specified, and separately discovers and closes a real gap where `MatchReturnDialog` had been built and exported but never rendered anywhere. The `action?: ReactNode` slot on the orphan banner is the right size of change, and the stale docblock that claimed the action was deliberately absent is corrected rather than left to mislead. The one substantive issue is an inconsistency inside the PR itself: the match button gets the full `useWriteAccess` + `ReadOnlyLock` treatment while the "+ Record a return" CTA gets none, leaving a read-only session a prominent page action that ends in a 403 rendered as "Try again."

### Merge Readiness

🔄 **Approve with changes** — merge after addressing the IMPORTANT item.

### Priority fixes

1. **Gate the "+ Record a return" page action** with `useWriteAccess('orders:write', useDemoMode())` + `ReadOnlyLock`, matching both `inventory-locations-page.tsx:104` and the match button in this same PR, and add the hidden-without-permission test. (`returns-list-page.tsx:241-249`)
2. *(optional)* Mount `RecordReturnDialog` inside `PageLayout`, dropping the fragment and re-indenting the block.
3. *(optional)* Drop the redundant `query.refetch()` calls — both mutations already invalidate.
4. *(optional)* Add the negative-permission test for the match button.

piotrswierzy (changes requested):
# Re-review — PR #3285 (entry point from the returns list/detail)

Re-reviewed at `2150461f`. **This branch has not been pushed since 13:40 and contains no new commits since the first review.**

```
git merge-base --is-ancestor origin/3083-authorize-return-dialog origin/3085-returns-entry-point  → false
git merge-base --is-ancestor origin/3084-record-return-dialog   origin/3085-returns-entry-point  → false
```

## 1. Prior findings

|---|---|---|---|
| 1 | **IMPORTANT** — the "+ Record a return" page CTA is ungated, three files from a correctly-gated match button in the same diff | **Not addressed** | `apps/web/src/pages/returns/returns-list-page.tsx:243-249` — a bare `<Button onClick={() => setIsRecordDialogOpen(true)}>`, no `useWriteAccess`, no `ReadOnlyLock`, no `useDemoMode`. The gated sibling is still right there at `return-detail-page.tsx:292-306`. |
| 2 | *(credited)* found and closed a dead `MatchReturnDialog`, corrected the now-false docblock | Still present, unchanged | — |

**[BLOCKING]** — the branch is stale against both parents and, merged as it stands, **reverts the two fixes just approved on #3283 and #3284**.

> This branch carries its own rebased copies of the #3083 and #3084 commits, made before the review fixes existed, so the fixed files are not in its history:
>
> - `git show origin/3085-returns-entry-point:apps/web/src/features/returns/components/record-return-dialog.schema.ts | grep sku` → **no match**. The #3284 BLOCKING (every operator-authored return permanently un-restockable, `return-custody.service.ts:687-693`) is present verbatim on this branch.
> - `git show origin/3085-returns-entry-point:apps/web/src/features/returns/components/orphan-returns-worklist.tsx | grep 'useWriteAccess\|ReadOnlyLock'` → **no match**. The #3283 ungated Approve action is present verbatim on this branch.
>
> Because these are *different commits* touching the same files rather than the same commits, a merge after #3284 lands is a three-way merge whose base predates both — it will either conflict or resolve one-sidedly, and a one-sided resolution here silently restores the un-restockable form. `docs/lessons.md` / the repo's standing rule is to re-fetch and rebase before opening *and* before merging, precisely because a sibling touching the same file makes a textual resolution wrong by intent.
>
> **Fix:** rebase `3085-returns-entry-point` onto `origin/3084-record-return-dialog` (`0c3df007`) and force-push. Then re-verify both greps return matches.

**[IMPORTANT]** — `apps/web/src/pages/returns/returns-list-page.tsx:243-249` — the ungated CTA (prior finding 1, restated as still open).

> `POST /returns/record` is `@Roles('admin', 'operator')` (`apps/api/src/returns/http/return-writes.controller.ts:124-125`), so a `viewer`/`packer` sees an enabled "+ Record a return", fills a six-field form, and is answered 403. `docs/frontend-architecture.md` § Access Control And UI Visibility row 2 requires `useWriteAccess('orders:write', demoMode)` + `ReadOnlyLock` for a write affordance. The precedent is in this same diff at `return-detail-page.tsx:179-180` and `:292-306`, and #3283 has now landed a third instance in `orphan-returns-worklist.tsx:196-197`. Lift the same three lines.
>
> Note the detail page's own gate spells the disabled condition as `disabled={!writeAccess.canWrite}` (`:300`) while #3283 uses `disabled={writeAccess.demoReadOnly}` (`orphan-returns-worklist.tsx:274`). Inside a `writeAccess.visible` branch the two are equivalent, but pick one spelling across the feature so a reader does not have to re-derive that.

**[IMPORTANT]** — `apps/web/src/pages/returns/returns-list-page.test.tsx`

> There is no test for the CTA's visibility by permission — `grep -n 'sessionAdapter' returns-list-page.test.tsx` returns nothing, so every test runs on `renderWithProviders`' default anonymous adapter (`test/test-utils.tsx:1240`) and the one test that clicks the CTA (`:520`) would keep passing whether or not the gate exists. #3283's fix is the model: the two new negative cases plus switching the pre-existing ones to `createAuthenticatedSessionAdapter()` is what makes the gate falsifiable (`orphan-returns-worklist.test.tsx:155-212`). Add a hidden-for-permission-less-session case and a demo-viewer disabled case.

**[SUGGESTION]** — `apps/web/src/pages/returns/return-detail-page.tsx:44`

> `import { useDemoMode } from '../../features/system/hooks/use-demo-mode';` reaches past the feature barrel. Every one of the ~20 in-tree consumers imports it from the barrel (`features/system/index.ts:8`), including the one #3283 just added. `frontend-architecture.md` § Feature Public Surface mandates the barrel for cross-feature imports and documents `pages/` deep imports as a known gap rather than an allowance — so this compiles, but it is the only outlier for this hook. Use `'../../features/system'`.

## 3. Summary

### Summary
No new commits since the first review. The IMPORTANT gating finding is untouched, and the branch has meanwhile fallen behind both of its parents, which have since landed the #3283 gate and the #3284 `sku` BLOCKING fix. As it stands, merging this PR carries pre-fix copies of both files and would undo work that was just approved — which is what turns a stack-hygiene chore into a BLOCKING item here rather than a note. The most important concern is that the un-restockable record form returns through this branch.

### Merge Readiness
❌ **Request changes**

### Priority fixes
1. **[BLOCKING]** Rebase `3085-returns-entry-point` onto `origin/3084-record-return-dialog` (`0c3df007`) so the `sku` requirement and the Approve-action gate survive this merge. Verify by re-running the two greps quoted above.
2. **[IMPORTANT]** Gate the "+ Record a return" CTA on `orders:write` with `useWriteAccess` + `ReadOnlyLock` (`returns-list-page.tsx:243-249`), matching `return-detail-page.tsx:292-306`.
3. **[IMPORTANT]** Add permission tests for that CTA; without them the gate is unfalsifiable on the default anonymous adapter.
4. **[SUGGESTION]** Import `useDemoMode` from the `features/system` barrel.

piotrswierzy (approved):
## Tech Lead review — `feat(returns): entry point from the returns list/detail`

The PR that makes the previous five reachable. No new route, both dialogs mounted where they belong, and the write gates applied consistently with the backend's `@Roles`.

### ✅ The `ReturnOrphanBanner` docblock correction

The old docblock asserted:

> *There is deliberately no `Match to an order` action. Re-attribution is an automatic reconcile, not a write an operator can trigger…*

That was true when written and is now false, and the replacement **says so** — naming the epic that changed it and why the original reasoning held at the time. That is the documentation-upkeep discipline the repo asks for, and it's the difference between a reader trusting the docblocks and learning to ignore them.

The `action` slot being optional, so a page that hasn't wired the dialog degrades to the explanation-only banner, is the right shape for a shared component.

### ✅ Write gates match the backend, again

Both new affordances gate on `orders:write`, which (verified in `ROLE_PERMISSIONS`) is held by exactly `admin` + `operator` — the same two roles `@Roles('admin','operator')` allows on `POST /returns/record` and `POST /returns/:returnId/match-order`. Hidden for no-permission, disabled-behind-`ReadOnlyLock` for a demo viewer. Both directions tested on both surfaces.

### ✅ No new route

The worklist lives above the segment strip on the existing `/returns`, with the reason stated: *"it answers 'what needs me right now', the strip below answers 'show me a filtered page'."* That's the right division and it keeps the nav unchanged.

### 🟡 IMPORTANT — this PR makes #3281's stale-mutation bug reachable

Both dialogs are mounted **persistently**, not conditionally:

```tsx
<RecordReturnDialog open={isRecordDialogOpen} onOpenChange={setIsRecordDialogOpen} />
```
```tsx
<MatchReturnDialog returnId={detail.id} open={isMatchDialogOpen} onOpenChange={setIsMatchDialogOpen} />
```

Neither unmounts on close, so `useMatchReturnToOrderMutation` / `useRecordReturnMutation` keep their `error` across a close/reopen — and neither dialog's `resetAndClose` calls `mutation.reset()`. Concretely: open "Match to an order", hit a network failure, see *"The match could not be completed. Try again."*, Cancel, reopen → **the error is still there**, on a fresh form, before anything was attempted.

I raised this on #3281 and #3284 as a suggestion; this mount is what turns it into a real operator-visible defect, so it's worth landing the fix somewhere in the stack. Two equivalent options:

- add `mutation.reset()` to each dialog's `resetAndClose` (fixes it for every caller), or
- mount conditionally here — `{isMatchDialogOpen ? <MatchReturnDialog … /> : null}` — which is what `OrphanReturnsWorklist` already does for the authorize dialog, and why that one has no equivalent problem.

I'd take the first: it makes the component correct regardless of how a future caller mounts it.

### 🟡 IMPORTANT — `Integration Tests` was **cancelled**, not green

The other eight checks pass. Please get a clean run before merging rather than reading the cancellation as a pass.

### 🟢 SUGGESTION — the test-fixture routing hack is load-bearing and fragile

`returns-list-page.test.tsx`'s `setup()` now routes `returns.list` responses by matching the call's `bucket` against the one parsed out of `options.route`, because the worklist issues two of its own list calls on this page and a single unconditional mock would serve the same fixture to both, rendering rows twice.

The comment explaining this is excellent and the mechanism works — but it encodes an assumption that no test in the file filters the table to `bucket=orphan` or `bucket=attributed`. The moment one does, its fixture is served to the worklist too and every unscoped `getByText` in that test goes ambiguous, with a failure mode that looks nothing like its cause.

Worth making it structural instead: give the worklist's own two calls a distinguishable response (they're identifiable by `bucket` **and** by `limit` — the approval scan uses `RETURNS_MAX_LIMIT`), or scope the table assertions with `within(table)`. Either removes the standing trap.

### 🟢 SUGGESTION — the worklist renders permanently, including on a healthy install

`<OrphanReturnsWorklist />` is unconditional, so an install with no orphans and nothing pending approval still gets "Needs your attention" plus two empty-group sentences above the main table, forever.

The copy is neutral rather than alarming, so this isn't the #2100 "red banner on a healthy install" trap — but it does permanently consume the space directly above the table an operator came for. The component already computes exactly the predicate needed to do better: collapse the whole section **only** when both groups are `isConfirmedEmpty`, and keep it visible whenever either read failed, was truncated, or dropped rows. That distinction is the one thing that must not be lost if this is changed.

### Notes

- `useDemoMode` moved from the deep hook path to the `features/system` barrel in `return-detail-page.tsx`, with a comment noting that `pages/` deep imports are a documented gap rather than an allowance. Nice tidy-up en route.
- Folding the orphan banner and its dialog under one `detail.bucket === 'orphan'` check removes a pair of adjacent conditions that could drift. Correct, and free since the dialog renders nothing while closed.
- Comments where `onMatched`/`onRecorded` are *not* passed, explaining that cache invalidation is the mechanism — good, that's the question the next reader would have.
- The detail-page gate test uses `authenticated: false` where the list-page one uses `createNoopSessionAdapter()`. The latter ("authenticated, but without the permission") is the closer analogue of a real `viewer`; worth aligning.

**Verdict:** ✅ Approve. Please get Integration Tests green, and land the `mutation.reset()` fix somewhere in this stack before it ships.

PR #3286, opened 2026-09-16
test(returns): orphan/manual returns flow tests

### LINKED ISSUE #3078: [EPIC] Frontend — Orphan return matching and operator-authored returns UI

### LINKED ISSUE #3086: test(returns): orphan/manual returns flow tests

jakubretajczykBD:
## Summary
Every component in this epic already had isolated unit tests; this adds the thing that was actually missing — a test that drives the real, multi-step sequence an operator goes through, with mutations genuinely feeding the next read, rather than each piece asserted against a static fixture in isolation.

Two flows:
- **Record → appears → approve → disappears**, entirely on the returns list page. The `returns.list` mock is STATEFUL (each write mutates the array the next read serves), so the test actually exercises the #3080 cache-invalidation wiring rather than merely asserting the API was called
- **Orphan → match → order shown**, crossing from the list page (the worklist's `Link`, #3081) to the detail page (the dialog's mount point, #3085) via a REAL `react-router` navigation — both routes are mounted together and the click is a genuine click-through

Also exports `AUTHORIZE_RETURN_DIALOG_COPY` from the returns barrel, completing the `*_DIALOG_COPY` export set the other two dialogs already had.

## Test plan
- [x] `tsc --noEmit` clean, full `pnpm --filter @openlinker/web build` clean
- [x] `eslint` clean
- [x] `vitest run src/features/returns src/pages/returns` — 351/351 passing

Closes #3086
Closes #3078

01-returns-list-with-worklist
02-approve-return-dialog
03-record-return-dialog
04-return-detail-orphan-banner
05-match-to-order-dialog
07-return-detail-operator-authored


piotrswierzy (approved):
# PR Review — #3286 `test(returns): orphan/manual returns flow tests`

Reviewed against `docs/testing-guide.md`, `docs/frontend-architecture.md § Testing Baseline`, and the existing `apps/web/src/pages/returns/*.test.tsx` baseline.

---

## Positive observations

- **The stateful mock is the point, and the PR knows it.** `returns-flow.test.tsx:118-119` keeps one `items` array that the mocked `record` and `authorize` writes mutate in place, so step 2's *"the recorded return now appears in Waiting for your OK"* genuinely proves `useRecordReturnMutation`'s invalidation drove a refetch. A mock returning a fixed fixture would pass the same assertion while proving nothing — this is the difference between a flow test and a sequence of unit tests sharing a `describe`.
- **The navigation is real.** Flow 2 mounts both routes under one `<Routes>` and clicks the worklist's actual `Link` (`returns-flow.test.tsx:277-281`), rather than pointing `route:` straight at the destination. That is what makes it a click-through rather than two independent page tests.
- **Both flows assert the resulting STATE, not just the call.** Flow 1 waits for `needsApprovalEmpty` to come back (`:262-264`); flow 2 waits for the orphan banner to disappear because `detail.bucket` really moved to `attributed` (`:302-304`). Asserting `toHaveBeenCalledWith` and stopping there is the common failure mode and this file avoids it in both places.
- Fixtures are built from `listItem()` / `order()` / `connection()` factories with `Partial<T>` overrides, matching the surrounding files, and every mock is cast through the real `apiClient` member types so a contract change breaks the test rather than silently diverging.
- Exporting `AUTHORIZE_RETURN_DIALOG_COPY` completes the `*_DIALOG_COPY` set — the tests assert against the copy constant, never a literal, so a reworded label cannot silently orphan an assertion.

---

## Findings

### **[SUGGESTION]** — `apps/web/src/pages/returns/returns-flow.test.tsx:121`

Flow 1's list mock ignores the filter entirely:

```ts
apiClient.returns.list = vi.fn(async () => listResult(items)) …
```

`OrphanReturnsWorklist` issues two queries — `{ bucket: 'orphan' }` and the approval scan — and only the second is filtered client-side (`orphan-returns-worklist.tsx`: `.filter((item) => item.origin === 'operator_authored' && item.authorizedAt === null)`); the orphan group trusts the server. So once the recorded return lands, it appears in **both** groups, including "Needs an order" for a return whose `bucket` is `'attributed'` and whose `internalOrderId` is set — a state the backend cannot produce.

The assertions still hold (the two groups' action labels differ, so `getByRole('button', { name: needsApprovalAction })` is unambiguous), so this is not a broken test. But flow 2 *does* honour the filter (`:253-255`), so the file is internally inconsistent, and a fixture the backend could never emit is the kind of thing that later makes a real regression look like a mock artefact. Mirroring flow 2's `filters.bucket === 'orphan' ? [] : items` would cost one line.

### **[SUGGESTION]** — `apps/web/src/pages/returns/returns-flow.test.tsx:115`

`afterEach(cleanup)` is redundant — `apps/web/src/test/setup.ts:14-16` already registers it globally, with a docblock explaining why it is explicit there (RTL only auto-registers under `globals: true`, which this config does not set). No sibling test file in `pages/returns/` repeats it. Harmless, but it reads as though this file needs something the others don't.

### **[SUGGESTION]** — `apps/web/src/pages/returns/returns-flow.test.tsx:143, 230`

Both flows run under `createAuthenticatedSessionAdapter()`, which is right — the match button is `writeAccess.visible`-gated and would not render otherwise. Worth noting for the record, though, that this means **the flow tests cannot currently distinguish a gated affordance from an ungated one**: the "+ Record a return" CTA (#3285) and the "Approve" row action (#3283) render regardless of session, so flow 1 would pass identically for a `viewer`. If those two are gated per the IMPORTANT findings on the sibling PRs, a third flow — or a `sessionAdapter`-less variant of flow 1 asserting the CTA is absent — would turn this file into real coverage of the permission model rather than coverage that happens to be satisfied.

Not a change this PR owes; it is only actionable once the gating lands.

### **[SUGGESTION]** — `apps/web/src/pages/returns/returns-flow.test.tsx:249`

`apiClient.returns.listReturnEventsForReturn = vi.fn().mockResolvedValue([]);` is the one mock assigned without the `as unknown as typeof …` cast every neighbour uses. It type-checks today, so this is cosmetic — but the casts elsewhere are what make a signature change surface here, and the odd one out will not.

---

## Step 4 — Final Assessment

### Summary

A genuinely valuable test-only PR, and one that does the harder version of the job: rather than adding a third isolated component test, it composes the real pages, lets a mutation's cache invalidation feed the next read through a stateful mock, and crosses a page boundary via an actual router navigation. Both flows assert the resulting rendered state rather than stopping at `toHaveBeenCalledWith`, which is what makes them regression-proof against the wiring rather than against the mocks. The findings are all cosmetic — an unfiltered list mock in flow 1 that produces a state the backend cannot, a redundant `cleanup`, and a note that the file cannot yet exercise the permission model because two of the three affordances it drives are not gated.

### Merge Readiness

Note the stack dependency: this sits on `3085-returns-entry-point`, so it lands after #3285 and, through it, after the #3284 BLOCKING finding is resolved. If `RecordReturnDialog` gains a required `sku` field per that review, `returns-flow.test.tsx`'s flow 1 will need one more `userEvent.type` in its fill sequence (`:174-186`).

### Priority fixes

None required. Optional polish:

1. Honour `filters.bucket` in flow 1's list mock, as flow 2 already does.
2. Drop the redundant `afterEach(cleanup)` — `src/test/setup.ts` registers it globally.
3. Once #3283/#3285 gate their write affordances, extend flow 1 (or add a variant) to assert they are hidden for a session without `orders:write`.

piotrswierzy (changes requested):
# Re-review — PR #3286 (orphan/manual returns flow tests)

Re-reviewed at `db7ec884`. **No new commits since the first review** (branch last pushed 13:40).

## 1. Prior findings

|---|---|---|---|
| 1 | **SUGGESTION** — flow 1's list mock ignores `filters.bucket`, producing a group state the backend cannot emit | **Not addressed** | `apps/web/src/pages/returns/returns-flow.test.tsx` — the `returns.list` stub still returns one `items` array regardless of the `bucket` filter each worklist query passes. |
| 2 | *(noted, not graded)* once #3283/#3285 gating lands, a `sessionAdapter`-less variant would turn this file into real permission coverage | **Not taken** | `:166` and `:279` both pass `createAuthenticatedSessionAdapter()`; there is no permission-less variant. |
| 3 | *(noted, not graded)* if `sku` became required in #3284, flow 1 needs one more `userEvent.type` | **Not done** | `:184-186` — the fill sequence is order → connection → item name → reason → confirm. No SKU. |

**[BLOCKING]** — `apps/web/src/pages/returns/returns-flow.test.tsx:177-190` — flow 1 cannot pass once #3284 is in its ancestry.

> #3284 landed `sku` as a **required** field (`record-return-dialog.schema.ts:44-48`), and this flow never types one. The click at `:183` will therefore fail Zod validation, `record` is never called, and the `waitFor(() => expect(apiClient.returns.record).toHaveBeenCalled())` at `:185-187` times out — taking the whole flow (steps 2–4) with it.
>
> This is a defect in the file under review rather than only a rebase chore: it is the exact "check that cannot fail / cannot pass" class `docs/testing-guide.md` treats as a defect, and it is latent only because the branch is stale. Add `await userEvent.type(screen.getByLabelText(RECORD_RETURN_DIALOG_COPY.skuFieldLabel), '<sku>');` between the connection and item-name steps, and rebase.

**[BLOCKING, inherited]** — the branch sits on `origin/3085-returns-entry-point`, which itself is stale against both its parents and carries pre-fix copies of `record-return-dialog.schema.ts` (no `sku`) and `orphan-returns-worklist.tsx` (ungated Approve). Detail and remedy are in the #3285 review; this PR cannot be landed ahead of that rebase.

**[IMPORTANT]** — `apps/web/src/pages/returns/returns-flow.test.tsx:164-167`, `:279`

> Both flows run authenticated, and at the time of the first review two of the three affordances they drive were ungated — so flow 1 would have passed identically for a `viewer`. #3283 has since gated the Approve action with its own negative cases, which removes the blind spot for *that* affordance. What remains is the "+ Record a return" CTA (#3285, still ungated): a flow test is the natural place to assert that a permission-less session cannot complete the record→approve journey at all. One `createNoopSessionAdapter()` variant asserting the CTA is absent would convert this file from "the happy path works for an admin" into genuine end-to-end permission coverage. Cheap, and it is the one thing a page-level flow test can assert that the component tests cannot.

**[SUGGESTION]** — `apps/web/src/pages/returns/returns-flow.test.tsx` (prior finding 1, restated)

> Honour `filters.bucket` in the `returns.list` stub. The worklist issues two distinct queries (`{ bucket: 'orphan' }` and the approval scan) and the current stub answers both with the same array, so the test asserts against a pair of group states the backend cannot simultaneously produce. `orphan-returns-worklist.test.tsx:73-95` already has the keyed-mock shape to copy (`one mock, two behaviours keyed on the bucket filter each query passes`).

Everything else in the file remains as reviewed: real cache-invalidation assertions rather than hand-spliced arrays (`:190-197`, `:209-214`), which is the right shape for a flow test.

## 3. Summary

### Summary
Unchanged since the first review. The file itself was clean then and is still well-constructed, but its parents have moved: #3284 made `sku` a required field, and flow 1's fill sequence does not supply one, so the test's central assertion will time out the moment the stack is rebased — a check that has stopped being able to pass, discovered before it becomes a red CI run. That, plus the inherited stale base from #3285, is the whole of the concern; there is nothing wrong with the test's design.

### Merge Readiness
❌ **Request changes** — mechanical, but must land with the rebase rather than after it.

### Priority fixes
1. **[BLOCKING]** Rebase onto the fixed `3085-returns-entry-point` (which must itself first be rebased onto `0c3df007`), then add the missing `skuFieldLabel` fill to flow 1.
2. **[IMPORTANT]** Add a permission-less variant asserting the record→approve journey is unreachable for a session without `orders:write`.
3. **[SUGGESTION]** Key the `returns.list` stub on `filters.bucket`.

piotrswierzy (approved):
## Tech Lead review — `test(returns): orphan/manual returns flow tests`

The right capstone for the stack, and it tests the two things the six isolated suites structurally could not.

### ✅ Stateful mocks, not fixed fixtures — this is what makes the tests worth having

```ts
let items: ReturnListItem[] = [];
apiClient.returns.record = vi.fn(async (input) => { items = [...items, recorded]; … });
apiClient.returns.authorize = vi.fn(async (returnId) => { items = items.map(…); … });
```

Every one of the six upstream PRs could pass with a mock returning a constant — which proves the API call fired, and nothing about whether the UI then changes. Here the mutation genuinely mutates the array the *next* read serves, so the assertions

- "`needsApprovalEmpty` disappears after recording"
- "`needsApprovalEmpty` returns after approving"

are actually exercising `useAuthorizeReturnMutation`'s `onSettled` → `invalidateQueries(returnsQueryKeys.all)` → refetch → the worklist's own client-side `authorizedAt === null` predicate. That whole chain is where a real regression would live, and nothing before this file touched it. The comment on step 4 makes the point precisely: *"without this test (or the component) ever splicing an array by hand."*

### ✅ The router navigation is real

```tsx
<Routes>
  <Route path="/returns" element={<ReturnsListPage />} />
  <Route path="/returns/:returnId" element={<ReturnDetailPage />} />
</Routes>
```
…with `route: '/returns'` and a click on the worklist's own `Link`. Pointing `route:` straight at the detail page would have tested the destination while assuming the link — and the link's `href` is exactly what #3080's worklist and #3085's mount had to agree on across two PRs. Worth the extra setup.

The post-match assertion is also the right one: the orphan banner is gone **because `detail.bucket` flipped to `'attributed'`**, not because the test stopped at "the mutation was called".

### ✅ The negative-permission flow

This is the one a page-level test can assert that six component tests cannot — that the *whole journey* is unreachable, not just that one affordance is hidden. The comment says exactly that, and `expect(apiClient.returns.record).not.toHaveBeenCalled()` is the assertion that makes it a statement about reachability rather than about rendering.

Also right: it asserts the read surface still renders (`sectionTitle` found). A permission gate that hid the page would be a different, worse bug, and this pins the difference.

### ✅ The bucket-keyed mock

Answering `{ bucket: 'orphan' }` with `[]` rather than the same array is a small thing with a real reason, stated in place: an operator-authored, order-attributed return appearing in "Needs an order" is a state the backend cannot produce, and a test that fabricates it would pass while describing something impossible. Good instinct — fixtures that can't exist are how tests end up asserting the wrong invariant.

### 🟢 SUGGESTIONS

1. **No failure path crosses a flow.** Every refusal is covered in the isolated dialog suites, which is defensible — but the one case that genuinely spans components is a **409 `already-attributed` arriving at the detail page**: the dialog swaps to its acknowledge view *and* fires `onMatched`, and the page behind it must then reflect the winning attribution. That interaction (dialog state + page re-read, across a mutation the operator did not win) exists in no single-component test. Worth one more case.

2. **`getCorrectionProposal`'s mock isn't cast** like its neighbours:
   ```ts
   apiClient.returns.getCorrectionProposal = vi.fn().mockResolvedValue({ outcome: 'no-invoice', proposal: null });
   ```
   Every other assignment in the file carries `as unknown as typeof …`. Harmless today, and note it will need `changeId` / `opened` once #3301 lands — though `parseCorrectionProposal` reads both as `nullish()`, so it'll keep passing. Worth a cast for consistency.

3. **The inline `ReturnDetail` literal in flow 2** is long, but it's type-checked, so a new required field on `ReturnDetail` breaks this at compile time rather than at runtime. That's the good failure direction — leaving it as-is is fine; just don't be tempted to `as ReturnDetail` it later.

### Notes

- The file header states plainly what each flow exists to prove and what the isolated suites already cover, so the next person can tell whether a new case belongs here or one level down. That's the part these files usually get wrong.
- The barrel gains `AUTHORIZE_RETURN_DIALOG_COPY` — consistent with the two copy modules #3285 already exported for the same reason (tests assert against copy constants, never string literals).
- CI green across all nine checks, including Integration Tests.

**Verdict:** ✅ Approve. Good close to the stack.

Issue #775, opened 2026-05-17
UI primitives overhaul — visual refresh + API tightening across shared/ui/


piotrswierzy:
## Context

`apps/web/src/shared/ui/` ships ~30 primitives covering the cockpit vocabulary in `docs/frontend-ui-style-guide.md`. They work, but the visual language feels flat and undifferentiated. We want to overhaul every primitive in one coordinated pass so the cockpit feels polished (Shopify admin clarity + Linear polish, per the existing style guide) without changing the architectural rules:

- No external styled UI library (no Tailwind / shadcn / MUI / Mantine)
- Vanilla CSS + design tokens
- `forwardRef`, native HTML + Radix headless
- See `.claude/rules/ui-components.md` and `.claude/rules/frontend.md`

## Goals

1. **Visual refresh** — rework `apps/web/src/shared/theme/tokens.ts` and the relevant sections of `apps/web/src/index.css` so primitives look modern and cohesive. Same operator-cockpit identity, higher fidelity.
2. **Per-component review** — go through every primitive, evaluate its current visual, API, and a11y, and ship the improvements.
3. **Keep the public API stable where possible** — consumers in `features/` and `pages/` shouldn't need wide rewrites. Breaking changes are allowed but must be intentional and listed in the PR.

## Scope — per-component checklist

**Feedback / status**
- [ ] `Alert`
- [ ] `StatusBadge`
- [ ] `EmptyState` / `ErrorState` / `LoadingState` (`feedback-state.tsx`)
- [ ] `StructuredErrorList`
- [ ] `EnvironmentBadge`

**Controls**
- [ ] `Button`
- [ ] `Input`
- [ ] `Textarea`
- [ ] `Select`
- [ ] `Combobox`
- [ ] `Chip`
- [ ] `FileUpload`
- [ ] `ThemeToggle`

**Form composition**
- [ ] `FormField`
- [ ] `FieldError`
- [ ] `FormErrorSummary`

**Layout / navigation**
- [ ] `PageLayout`
- [ ] `BackLink`
- [ ] `Tabs`
- [ ] `SetupStepper`
- [ ] `WizardLayout` / `WizardSummaryRow`
- [ ] `DesktopOnlyBanner`

**Overlays / popovers**
- [ ] `Dialog`
- [ ] `ConfirmDialog`
- [ ] `DropdownMenu`
- [ ] `Popover`
- [ ] `Tooltip`
- [ ] `ToastProvider`

**Data surfaces**
[...]

PR #3288, opened 2026-09-16
feat(web): show release version instead of static "DEV" text in top-left badge

### LINKED ISSUE #3287: [TASK] Frontend — Show release version instead of static "DEV" text in top-left badge
## Problem / Context

The top-left corner of the app shell (`apps/web/src/app/app-shell.tsx`, `SidebarBrand`) renders `EnvironmentBadge` (`apps/web/src/shared/ui/environment-badge.tsx`) next to the OpenLinker logo. In a `VITE_APP_ENV=development` build this renders the static text "Dev" — no build carries the actual release version, so an operator looking at a running instance (or a screenshot) has no way to tell which release it's running without checking the deployment pipeline.

The release version is already tracked centrally: `release-please` (see `release-please-config.json`, single `"."` package) bumps the **root** `package.json` `version` field on every release — not `apps/web/package.json`, which release-please never touches.

## Proposed Solution

- In `apps/web/vite.config.ts`, read the root `package.json` version at build time and inject it via Vite's `define` as a global constant (e.g. `__APP_VERSION__`).
- Add a type declaration for the injected global (`apps/web/src/global.d.ts` or similar) and a small typed accessor module (e.g. `apps/web/src/shared/config/app-version.ts`) exporting `APP_VERSION`, following the existing `shared/config/` module pattern (`env.ts`, `api-version.ts`).
- Update `EnvironmentBadge` (`apps/web/src/shared/ui/environment-badge.tsx`) to render the version (e.g. `v0.11.0`) instead of the environment short label. Keep the existing tone/colour derived from `VITE_APP_ENV` (`getEnvironmentMeta`) so the environment signal is still conveyed via colour.
- Update the call site in `SidebarBrand` (`apps/web/src/app/app-shell.tsx`).
- Update/extend the colocated test `apps/web/src/shared/ui/environment-badge.test.tsx`.

No backend change needed — this is purely a build-time read of an existing file.

## Classification

**Type**: Frontend
**Layer**: Interface (shared UI component) / build tooling (`vite.config.ts`)
**File(s)**:
- `apps/web/vite.config.ts`
- `apps/web/src/shared/ui/environment-badge.tsx`
- `apps/web/src/shared/config/app-version.ts` (new)
- `apps/web/src/global.d.ts` (new)
- `apps/web/src/app/app-shell.tsx`
- `apps/web/src/shared/ui/environment-badge.test.tsx`

## Dependencies

- None.

## Assumptions

- The version is read from the **root** `package.json`, since that's the file release-please's `"."` package config bumps — not `apps/web/package.json`, which stays pinned separately and unrelated to the release line.
- The badge's environment-derived colour/tone stays as the environment signal; only the displayed text changes from the environment short-label to the version number. No new environment values or tones are introduced.
- No runtime API call is introduced — the version is a build-time constant, consistent with how `VITE_*` env vars are already documented as build-time-only in `docs/frontend-architecture.md` § Runtime Configuration.

## Acceptance Criteria

- [ ] The badge in the top-left corner of the authenticated app shell shows the release version (e.g. `v0.11.0`) instead of the static "Dev"/"Stg"/"Prod" text.
- [ ] The version is read automatically from the root `package.json` at build time — no manual value to update per release.
- [ ] The badge's tone/colour still reflects the environment (dev/staging/prod), unchanged from current behaviour.
- [ ] A production build (`vite build`) embeds the current root `package.json` version string in the output bundle.
- [ ] Tests added/updated for `EnvironmentBadge` covering the version-rendering behaviour.
- [ ] `pnpm lint` and `pnpm type-check` pass for `apps/web`.
- [ ] No architecture boundary violations (CORE ↔ Integration) — change is entirely within `apps/web`.

norbert-kulus-blockydevs:
## Summary
- The top-left badge (`EnvironmentBadge`, next to the OpenLinker logo) showed only the environment (`Dev`/`Stg`/`Prod`) — nothing identified which release was running.
- It now keeps the environment label on its existing line and adds a smaller release-version line below it (e.g. `Prod` / `v0.11.0`), read at build time from the **root** `package.json` version — the file `release-please` bumps on every release (`release-please-config.json`, single `"."` package) — via a Vite `define` (`__APP_VERSION__`). No manual value to sync per release.
- The badge's tone/colour still reflects `VITE_APP_ENV` (unchanged).
- If the build-time inject never ran, `APP_VERSION` resolves to `''` and the badge omits the version line entirely, rather than rendering a plausible-looking fake version.
image

## Review fixes (this round)
- `.context-chip.environment-badge--stacked` was missing `display: inline-flex` — without it the `flex-direction`/`align-items`/`gap` declarations on that class were inert (the badge is a flex *item* of `.shell-brand`, not a flex *container* itself), so the two-line look came from incidental line-wrapping rather than the rule that was written.
- `.environment-badge__version` used `font-size: 0.625rem` (below the canonical type scale's 0.6875rem floor) with `opacity: 0.75`, which multiplies the tuned `--status-*-fg`/`--status-*-soft` contrast pair. Raised to `0.6875rem` and replaced the opacity with a lighter `font-weight: 400` (the label stays `600`), so the version line is visually secondary without degrading contrast.
- Added a spec for `APP_VERSION`'s module-level resolution (`app-version.test.ts`) — the existing tests only exercised the `version` prop passed explicitly to the component, never the `typeof __APP_VERSION__` branch itself.
- Dropped the description's inaccurate claim that `app.test.tsx` needed a fix — it was never in the diff; its existing `/^(Dev|Development)$/` regex already tolerates the restored label unchanged.
- Left `compact` / `shortLabel` as they are: `compact` has no production caller today, but removing it would leave `shortLabel` unused again (the "dead field" this round's earlier fix resolved) with no test covering that render path.

## Test plan
- [x] `pnpm lint` (full monorepo) — 0 errors
- [x] `pnpm type-check` (full monorepo) — clean
- [x] `pnpm --filter @openlinker/web exec vitest run` for `environment-badge.test.tsx`, `app-version.test.ts` and `app.test.tsx` — all passing
- [x] `check-design-tokens.mjs` — clean
- [x] Verified with a real `vite build` that the current `package.json` version string is embedded in the output bundle

Closes #3287


piotrswierzy (commented):

**Verdict: 🔄 Approve with changes.** The build-time plumbing is the right shape and I'd take it as-is; two findings concern what the badge now *communicates*, one of which is an accessibility regression rather than a style preference.

### Verified

- **Build-time, not runtime, and not a `VITE_*` var — correctly.** The version is read from the ROOT `package.json` in `vite.config.ts:20-24` and injected through `define` as `__APP_VERSION__`. That is better than a `VITE_*` environment variable for this purpose, not a deviation from `docs/frontend-architecture.md` § Environment Variables: an env var can be misconfigured per-deploy and drift from the artifact, whereas a `define` is a compile-time literal that cannot disagree with the bundle it is in. The comment correctly notes release-please is configured with a single `"."` package, so `apps/web/package.json` is never bumped and reading it would have shown a stale number forever.
- **No secret or build-internal leaks.** Only `.version` is destructured from the parsed JSON.
- **The existing literal-`DEV` assertion was found and updated.** `app.test.tsx:63` moves from `getAllByText(/^(Dev|Development)$/)` to `getAllByText(\`v${APP_VERSION}\`)`, importing the same constant the component uses rather than hardcoding a number that would need bumping every release.
- **The ambient declaration is scoped and guarded.** Declaring `__APP_VERSION__` module-locally rather than in a `global.d.ts` is right given `apps/**/*.d.ts` is gitignored as a build-artifact guard, and the `typeof` check avoids a `ReferenceError` outside the Vite/Vitest graph.
- **Removing `compact` is safe.** `EnvironmentBadge` has exactly one call site (`app-shell.tsx:143`, updated here) and is not re-exported from `shared/ui/index.ts`, so it is not part of the public component catalog and no plugin can be compiled against the old prop.

### Findings

**[IMPORTANT]** — `apps/web/src/shared/ui/environment-badge.tsx:84-88`
> **The environment signal is now colour-only for sighted users.** Before this change the badge read `Dev` / `Stg` / `Prod`; now it reads `v0.11.0` in every environment and the environment survives *only* as `context-chip--{tone}`. The `/pr-review` accessibility checklist calls this out by name — "badges with text not just color" — and it matters more than usual here, because the thing being distinguished is *which system am I about to act on*. A red/green distinction between staging and production is invisible to a deuteranopic operator, and the consequence of getting it wrong on this particular badge is an action taken against the wrong environment. The `aria-label` at `:87` does carry both facts, so screen-reader users are covered — but that is precisely the asymmetry to avoid: the label is a fallback for AT, not a substitute for visible text. Cheapest fix that keeps the PR's goal: render both, e.g. `Prod · v0.11.0`, or keep `shortLabel` as a prefix — which also resolves the next finding.

**[IMPORTANT]** — `apps/web/src/shared/config/app-version.ts:15`
> `'0.0.0'` is a *plausible version*, not an unknown signal. If the `define` never ran, the badge confidently asserts the app is version 0.0.0 and an operator asking "what is deployed here?" gets a wrong answer with nothing indicating the value is a fallback. This is the repo's standing rule in its usual form — a failed read must not render a value that reads as a real one (`docs/frontend-architecture.md` § Paginated Totals is explicit that a failed count keeps its placeholder and never renders `0`; the returns surfaces catch to empty and never make a positive claim from an absent value). Reachability is low — the define runs in both the build and the Vitest config — but the failure is silent, which is what makes it worth closing. Rendering no version, or a literal `dev`, is the honest degradation.

**[SUGGESTION]** — `apps/web/src/shared/ui/environment-badge.tsx:8,30,39,45,52,60`
> `shortLabel` is now dead. With `compact` gone nothing renders it, yet it is still declared on `EnvironmentMeta`, populated across all five branches (including the `label.slice(0, 4)` fallback), and pinned by three assertions in `environment-badge.test.tsx:8,14,22`. The tests now defend a field no surface uses, which is the kind of coverage that reads as meaningful and is not. Either delete it with its assertions, or — better, and it closes the first finding too — put it back on screen alongside the version.

### Merge readiness

🔄 **Approve with changes.** All eight checks green, MERGEABLE against `main`, no migration, correct layer placement, and the build-time mechanism is the right one. Nothing here needs redesign; the two IMPORTANT items are both one-line changes to what the badge renders, and the first of them is the one I would not merge without.

Priority:
1. Restore a visible environment signal alongside the version (also settles the dead `shortLabel`).
2. Make the missing-version fallback read as unknown rather than as `0.0.0`.

norbert-kulus-blockydevs:
Addressed in 2d0158c.

- **Environment signal restored (IMPORTANT)**: the badge text is `Dev`/`Stg`/`Prod` again instead of the version — colour is no longer the only signal an operator has for which environment they're acting on. `shortLabel` is back on screen too, which also resolves the "dead field" suggestion — no deletion needed, it's rendered again in compact mode.
- **Fake fallback version removed (IMPORTANT)**: `APP_VERSION` falls back to `''` instead of `'0.0.0'`. The badge omits the version line entirely when it's empty, rather than asserting a plausible-looking version that never actually shipped.
- Layout: kept the pre-existing single-line env pill as-is and added the release version as a smaller (`0.625rem`, 75% opacity) line below it — e.g. `Prod` / `v0.11.0` — rather than swapping the text.

Verified: scoped vitest run on `environment-badge.test.tsx` + `app.test.tsx` (11/11 passing), full `tsc -b` for `apps/web` clean, targeted eslint clean, and `check-design-tokens.mjs` clean.

image

piotrswierzy (approved):
## Tech Lead review — #3288

Read against `docs/frontend-architecture.md`, `docs/frontend-ui-style-guide.md` § *Canonical type scale (#775)* / § *CSS Implementation Standard*, `docs/engineering-standards.md`, and `release-please-config.json`.

### What is right, and non-obviously so

- **The version source is correct and verified.** `release-please-config.json` declares a single `"."` package, so the root `package.json` `version` (`0.11.0`) is the one release-please bumps; `apps/web/package.json` sits at `0.1.0` and would have been the wrong file. `vite.config.ts:20-23` reads the root file, and the comment says why.
- **No boundary is crossed.** The only injected value is the public root `version` — no secret, no commit SHA, no branch, nothing derived from a non-`VITE_*` variable. `#591` holds: `apps/web` still imports nothing from `@openlinker/core`.
- **The unset case degrades honestly.** `app-version.ts:17-18` guards with `typeof` and resolves `''`, and `environment-badge.tsx:98` omits the line rather than rendering `v0.0.0`. The docblock records that `'0.0.0'` was rejected as *"a real-looking value that would misreport what's deployed"* — the right call, and the `aria-label` drops the version with it.

---

### [IMPORTANT] — `apps/web/src/index.css:1188-1195`

`.context-chip.environment-badge--stacked` sets `flex-direction: column`, `align-items: center` and `gap: 1px`, but **`.context-chip` never declares `display`**. The base group rule at `index.css:1722-1732` sets only border / radius / padding / background / color / font-size, the four `--{info,warning,review,success,neutral}` modifiers set colours only, and the one `display: inline-flex` in that neighbourhood (`index.css:1769`) belongs to `.chip`, which `index.css:1765` explicitly documents as *"Distinct from `.context-chip`"*.

So the badge is a `<span>` that is blockified only because it is a flex **item** of `.shell-brand` (`index.css:1166-1171`). It becomes a block box, not a flex container — and on a block box all three of those declarations are inert. The two inner spans are inline and simply **wrap**; the two-line appearance in the screenshot is line-wrapping at 240px sidebar width, not stacking. `align-items: center` and `gap: 1px` do nothing at all, and the layout is width- and label-length-dependent.

The same base rule is why `min-height: 28px` at `index.css:5788-5791` has never applied either — further evidence `.context-chip` was not authored as a flex container.

Fix is one declaration in the new rule:

```css
.context-chip.environment-badge--stacked {
  display: inline-flex;   /* ← the rest of this block is inert without it */
  flex-direction: column;
  ...
}
```

### [IMPORTANT] — `apps/web/src/index.css:1199-1203`

`.environment-badge__version` uses `font-size: 0.625rem` (10px) with `opacity: 0.75`. Two problems, both in the style guide:

1. **10px is below the canonical type scale.** `docs/frontend-ui-style-guide.md` § *Canonical type scale (#775)* bottoms out at `0.6875rem` / 11px (*"Eyebrow / uppercase / mono-caps"*), and states *"Keep page-level typography aligned with these rem values."* The 10px nav group label at line 131 is a documented exception at `600` weight — this is `500` at 75% opacity, which is not that case.
2. **`opacity` de-emphasises by degrading a tuned contrast pair.** `.context-chip--{tone}` pairs `--status-*-soft` background with `--status-*-fg` text (`index.css:1734-1762`) — a deliberately tuned pair. Multiplying the foreground by `0.75` at 10px reduces effective contrast in both themes, and § *CSS Implementation Standard* asks for *"theme values as tokens first and consume those tokens in component rules"*. Prefer `font-size: 0.6875rem` and a `--text-*` token (or a lighter `font-weight`) over `opacity`.

### [SUGGESTION] — `apps/web/src/app/app-shell.tsx:143`

Dropping `compact` here leaves the prop with **zero production callers** — `EnvironmentBadge` is mounted in exactly one place (grep across `apps/web/src/**/*.tsx`), and after this change `compact` is exercised only by `environment-badge.test.tsx:38`. Either keep `compact` in the sidebar (the short label plus a version line is the tighter composition for a 240px nav) or drop the prop and its test.

### [SUGGESTION] — PR description

The description states *"Fixed `apps/web/src/app/app.test.tsx`'s app-shell smoke test, which asserted the old `/^(Dev|Development)$/` text."* That file is **not in the diff** (`gh pr diff 3288 --name-only` lists six files, not including it), and on `main` line 62 already reads `expect(screen.getAllByText(/^(Dev|Development)$/)).not.toHaveLength(0)` — a tolerant regex with `getAllByText`, which passes unchanged. No fix was needed; the claim should come out so a reader does not go looking for it.

### [SUGGESTION] — `apps/web/src/shared/config/app-version.ts:18`

All three new tests pass `version` explicitly, so the **default-prop path** (`version = APP_VERSION`) and the `typeof __APP_VERSION__ !== 'undefined'` guard the docblock spends a paragraph justifying are never executed. The rendering branches are covered; the module-level resolution is not. A single spec importing `APP_VERSION` and asserting it is a string would close it.

---

## Summary

A small, well-reasoned change that puts the release version in the shell badge, sourced from the one file `release-please` actually bumps, with an honest empty-string fallback and no secret or internal build detail reaching the bundle. The build-time plumbing and the degradation story are both right. The concern is the CSS: the new stacked layout is declared with flex properties on a class that has no `display`, so it renders by accidental line-wrapping rather than by the rule that was written, and the version line sits one step below the canonical type scale with `opacity` standing in for a colour token.

## Merge Readiness

🔄 **Approve with changes** — nothing blocking; land the two IMPORTANT items as follow-ups (or in this PR, since the first is a one-line fix).

## Priority fixes

1. Add `display: inline-flex` to `.context-chip.environment-badge--stacked` — without it `flex-direction`, `align-items` and `gap` are all inert (`index.css:1188-1195`).
2. Raise `.environment-badge__version` to `0.6875rem` and replace `opacity: 0.75` with a `--text-*` token (`index.css:1199-1203`).
3. Decide whether `compact` survives; remove it and its test if not.
4. Drop the inaccurate `app.test.tsx` claim from the description.

norbert-kulus-blockydevs:
Addressed in 8a7192f (rebased on `main`, conflicts resolved).

- **`display: inline-flex` added** to `.context-chip.environment-badge--stacked` (`index.css:1193`). Without it the class's `flex-direction`/`align-items`/`gap` were inert — the badge is a flex *item* of `.shell-brand`, not a flex container itself, so the two-line look came from incidental wrapping rather than the rule that was written.
- **`.environment-badge__version` raised to `0.6875rem`** (the canonical type scale floor) and swapped `opacity: 0.75` for `font-weight: 400` (label stays `600`) — de-emphasises the version line without multiplying the tuned `--status-*-fg`/`--status-*-soft` contrast pair.
- **Added `app-version.test.ts`** covering `APP_VERSION`'s module-level `typeof __APP_VERSION__` resolution — every existing test only exercised the `version` prop passed explicitly to the component.
- **Dropped the inaccurate `app.test.tsx` claim** from the description — confirmed it was never in the diff.
- **Kept `compact`/`shortLabel` as-is.** Removing `compact` (no production caller today) would make `shortLabel` unused again with no render-path test covering it — the exact "dead field" the previous round's fix resolved. Framing this as a deliberate choice rather than an oversight.

Merge conflicts with `main` (unrelated `sales-documents` and CI-worker-count changes) resolved — the only real conflict was `vite.config.ts`, where both the version-injection code and main's new `unitTestWorkers()` helper needed to survive together.

Verified: `environment-badge.test.tsx` + `app-version.test.ts` (6/6 passing), `tsc --noEmit` on `apps/web` clean, `check-design-tokens.mjs` clean, both commits GPG-signed and DCO-signed-off.

Issue #2382, opened 2026-08-22
[FEATURE] Frontend — refund confirmation, commission-refund block and proposal review


piotrswierzy:
> Part of the OMS programme backlog (**Wave 2**, stream S3, size L). Backlog slug: `W2-44`. Sibling `W2-*` slugs resolve via the Wave-2 epic's checklist. Design of record: returns product spec §5.5/§5.6/§5.7 (PR #2318).

> **Size justification (L):** three distinct money surfaces on one panel — the first `RefundRecord` UI the product has ever had, an Allegro-only commission block, and a credit-note proposal review whose whole point is rendering ambiguity legibly before an irreversible confirm.

## Problem / Context

Stories T6, T10 and T7. Each is money leaving or arriving, and each has a different owner: the PSP, the marketplace, and the tax authority.

## Proposed Solution

- **Refund confirmation** — amount prefilled from received/disposed lines, currency locked to the order, `RefundReasonValues`, note; writes a `RefundRecord` linked to the return with `executedBy: 'operator_out_of_band'`. Scope the form as a **shared component** so the order-level capture path can reuse it.
- **Commission-refund block** — a distinct block leading with *"your money coming back from Allegro"*, status read from the observed `rawStatus` timeline, plus the `Claim commission refund` action (shape per #2375's branch). **Absent, not empty, on non-Allegro returns.**
- **Credit-note proposal review** — leading warning that a transmitted correction cannot be withdrawn; per line `Matched` / `Ambiguous` (all candidates listed, operator picks) / `No match` (excluded, with reason); a **visible difference between a clean and an ambiguous proposal before the confirm button**; no auto-issue, with a footer explaining why; hands off to the existing `CorrectionIssuer` flow.

## Classification

**Type**: Frontend
**Layer**: Interfaces
**File(s)**: `apps/web/src/features/returns/components/**`; a shared refund form under `shared/`

## Dependencies

- #2376 (`W2-39`), #2378 (`W2-41`), #2379 (`W2-37`), #2374 (`W2-38`)

## Assumptions

- The `CorrectionIssuer` flow already has a UI entry point to hand off to; if not, the handoff is a route link, not a reimplementation.

## Acceptance Criteria

- [ ] Refund form writes a linked `RefundRecord`; currency is not editable
- [ ] Commission block is absent on non-Allegro returns (test)
- [ ] An ambiguous proposal is visually distinguishable from a clean one before any confirm (test)
[...]

PR #3301, opened 2026-09-18
feat(returns): correction-proposal API client (preview GET, record POST)

### LINKED ISSUE #3088: feat(returns): correction-proposal API client (preview GET, record POST)
> ### ⚠️ The design changed — read this before starting (#3116, #3123)
>
> The mockup this issue was written against described a **per-row picker** for ambiguous invoice
> lines. That picker is **retired**. The design of record is now the **invoice-line grid** in
> `docs/plans/mockups/correction-proposal-3087.html` — a row's position *is* the line number a correction addresses, so
> the operator is never asked which line a returned item belongs to.
> `docs/specs/product-spec-oms-returns-operator-ux.md` § 5.8 carries the amendment.
>
> **This work is blocked on #3172.** `ReturnLine.resolvedOrderLineId` is hardcoded `null`
> (`returns.service.ts:444`); until it is populated OpenLinker cannot pre-fill a single quantity in
> the grid. Do not start ahead of it.
>
> **Cite the repo file, never the Artifact URL** (`docs/frontend-architecture.md` § UX Mockups). Every
> state is settable without a click: open the file with `?state=<name>`. States below are
> `body[data-state]` values.

> **Mockup states this issue implements.** No state renders from this issue, but the client must carry the fields the grid needs: `CorrectionLine.newQuantity` **and** `newUnitPriceGross` (the returns path used only the first), and the invoice's own lines from `GET /invoices/:id/content` (#2076), which are index-aligned with the array a correction addresses.

## Problem / Context

`return-proposal.schema.ts` exists but nothing calls `GET/POST /returns/:returnId/correction-proposal`.

## Proposed Solution

Add `previewCorrectionProposal(returnId)` and `recordCorrectionProposal(returnId)` to `features/returns/api/returns.api.ts`, typed against `ReturnCorrectionProposalResponseDto` in `apps/api/src/returns/dto/return-correction-proposal.dto.ts`.

## Classification

**Type**: Frontend
**File(s)**: apps/web/src/features/returns/api/returns.api.ts

## Dependencies

- Part of #3087

## Assumptions

- None

## Acceptance Criteria

- [ ] Both methods exist and are typed against the DTO

jakubretajczykBD:
## Summary
- `previewCorrectionProposal(returnId)` (existing `getCorrectionProposal`) and new `recordCorrectionProposal(returnId)` (POST, no body) in `features/returns/api/returns.api.ts`, typed against `ReturnCorrectionProposalResponseDto`
- Threads the ADR-044 `changeId` / `opened` fields from the response DTO into `ReturnCorrectionProposalResult`

## Test plan
- [x] `pnpm --filter @openlinker/web test` (`returns.api.test.ts`) — 12/12 passed
- [x] Both methods exist and are typed against the DTO
- [x] No architecture boundary violations (CORE ↔ Integration)

Closes #3088


piotrswierzy (approved):
# PR Review — #3301 `feat(returns): correction-proposal API client (preview GET, record POST)`

Tech Lead systematic review per `docs/code-review-guide.md`.

## Step 1 — Context

Four files, all under `apps/web/src/features/returns/api/`. Adds `recordCorrectionProposal` (POST) beside the shipped `getCorrectionProposal` (GET), and threads the ADR-044 `changeId` / `opened` fields through `parseCorrectionProposal` onto `ReturnCorrectionProposalResult`. Baseline read on `main`: `return-proposal.schema.ts` (#2382), `apps/api/src/returns/http/return-writes.controller.ts:491,510`, `apps/api/src/returns/dto/return-correction-proposal.dto.ts`, `libs/core/src/returns/application/services/return-correction-proposal.service.ts`.

## Step 2 — Findings

### Contract fidelity against `docs/architecture-overview.md` § 22 Returns (#2374)

Verified, not assumed — every rule the bullet states holds at this seam:

- **"A proposal is DATA; it must never issue."** Both methods hit `/returns/:id/correction-proposal` and return the parsed result; nothing here reaches `issueCorrection` or `CorrectionIssuer`. `returns.api.ts:190-198` says so in the docblock and the code matches.
- **The GET is non-mutating, and that is asserted.** `returns.api.test.ts:99-107` pins `request` being called with a single argument and `init?.method` being `undefined` — a future author who adds `{ method: 'POST' }` to the preview fails the spec. Correct: the backend route (`return-writes.controller.ts:491`) is a read that persists nothing.
- **Every non-proposing exit stays a NAMED outcome.** `outcome` is carried through as a plain string (`return-proposal.schema.ts:47`, with the header at :7-9 stating why), so `no-invoice` / `no-line-snapshot` / `no-disposed-lines` / `nothing-correctable` reach the operator distinctly and a value this build predates is not blanked. `features/returns/lib/return-proposal.copy.ts:56-60` already carries distinct copy for all four. Nothing is collapsed into a generic empty or error state.
- **`ambiguous` is untouched.** `candidates` still round-trips in full with no preselection, and `selectedOriginalLineNumber` is only set on `matched`.

### The two new fields mirror core exactly

`changeId: null` / `opened: false` is what `return-correction-proposal.service.ts:183,191,203,230,233` returns on every non-recording exit, and `:135` is the one arm that sets them. The FE defaults (`return-proposal.schema.ts:75-76`) therefore reproduce the wire rather than inventing a value, and the comment at :59-61 states the right rule — absence degrades to "not reported as freshly opened", never to a claim that a row was reused. `.nullish()` rather than `.optional()` on both (`:59-60`) is correct per the OL serialisation rule: absent optional fields arrive as JSON `null`, and `.optional()` would drop the field.

### Type widening is safe

`ReturnCorrectionProposalResult` gains two **required** members (`returns.types.ts:441-450`). I checked every consumer — `parseCorrectionProposal` is the only construction site in `apps/web`; `return-detail-page.test.tsx:145,454` builds its literals through an untyped `Mock`, and `return-proposal-panel.test.tsx:53` takes `ReturnCorrectionProposal`, not the result — so nothing breaks. CI agrees: Lint, Type Check, Test and Build are all green.

### Conventions

- No raw `fetch()`; goes through the injected `ApiRequest` (`api-client.ts:152`). ✅
- Bodyless POST matches a wide precedent (`inventory.api.ts:119`, `sync.api.ts:126`, `price-changes.api.ts:58`, `listings.api.ts:714`). ✅
- Dependency direction `features/returns/api` → `shared` only. ✅
- Parsed, never cast; a contract break surfaces as `ReturnProposalUnreadableError`. ✅

**[SUGGESTION]** — `apps/web/src/features/returns/api/returns.api.test.ts` (96)
> `proposalOutcome` is 105 characters, over the repo's `printWidth: 100` (`.prettierrc`). `apps/web`'s lint script is bare `eslint` with no prettier plugin and no `max-len`, so this passes the gate — cosmetic only, but it is the one line in the diff the repo's own formatter would rewrite.

**[SUGGESTION]** — `apps/web/src/features/returns/api/returns.api.test.ts` (94-151)
> Every new case is built on `nothing-correctable`. The invariant most worth pinning at this boundary is the one #2374 exists for: that a `proposed` result carrying an `ambiguous` line round-trips with **every** candidate and **no** `selectedOriginalLineNumber`, alongside the new `changeId`/`opened`. That is currently asserted nowhere at the client seam, and it is the shape a future "helpful" default would break silently.

**[SUGGESTION]** — `apps/web/src/features/returns/api/returns.api.ts` (190)
> Issue #3088 names the method `previewCorrectionProposal`; the PR keeps the shipped `getCorrectionProposal` (one caller at `hooks/use-return-proposal-query.ts:26`). Keeping it is the right call — a rename is churn for no reader — but it is worth one line in the issue so the AC "both methods exist" is not read later as unmet.

**[SUGGESTION]** — follow-up scope
> `recordCorrectionProposal` has no consumer yet, which is correct for an API-client slice under #3087. Two things belong with the hook that follows: invalidating `returnsQueryKeys.correctionProposal(returnId)` after a successful record (the row is replaced when the proposal diverged, so a cached preview goes stale), and surfacing the 409 the route raises for an orphan return (`return-writes.controller.ts:523`), which this layer passes through untouched.

### On the issue's amendment banner

The banner asks that "the client must carry `CorrectionLine.newQuantity` **and** `newUnitPriceGross`, and the invoice's own lines from `GET /invoices/:id/content`". Both are already satisfied outside this file and neither is a gap here: `newUnitPriceGross` is structurally absent from this endpoint by design — a return changes a quantity, never a unit price (`architecture-overview.md` § 22, "The delta is quantity only"), and `libs/core/src/returns/domain/domain-services/__tests__/return-correction-matching.domain-service.spec.ts:258` asserts the field is *not* emitted — while the invoice-content read already exists at `features/invoicing/api/invoicing.api.ts:143`. Worth a sentence in the PR body so the auto-close does not read as leaving that work invisible.

### Positive observations

- `.nullish()` over `.optional()` on both new fields, with the OL serialisation reason implicit and the degrade documented at the point of decision.
- The GET's non-mutation is asserted structurally rather than by comment.
- The docblock at `returns.api.ts:190-198` states *why* the POST takes no body — a pick belongs to the confirm request — which is the #2374 rule, not a restatement of the signature.

---

## Step 4 — Final Assessment

### Summary
A tight, well-grounded API-client slice. It adds one POST method and two response fields, and every decision in it mirrors a rule that already exists in `libs/core` — the named-outcome passthrough, the `changeId`/`opened` defaults, and the non-mutating GET are all verifiable against `return-correction-proposal.service.ts` and the controller rather than asserted. The type widening is safe across every consumer, and CI is green. The only substantive gap is test breadth: the new cases never exercise a `proposed` result with an `ambiguous` line, which is the shape #2374's whole design turns on.

### Merge Readiness
✅ **Approve**

### Priority fixes
None blocking. Optional, in order:
1. Add one round-trip case over a `proposed` outcome carrying an `ambiguous` line (all candidates present, none selected) alongside the new fields.
2. Note in the PR/issue why `getCorrectionProposal` kept its name and why `newUnitPriceGross` is not carryable from this endpoint.
3. Wrap the 105-char test helper to the repo's `printWidth`.

piotrswierzy (approved):
## Tech Lead review — `feat(returns): correction-proposal API client (preview GET, record POST)`

Tight, correctly-scoped slice. Four files, no new surface beyond what the backend already ships, and the docblocks carry the reasoning rather than restating the code.

### ✅ Verified against the backend contract

I checked this against `apps/api/src/returns/http/return-writes.controller.ts` and `return-correction-proposal.dto.ts` on `main`:

- Both routes exist (`@Get`/`@Post ':returnId/correction-proposal'`) and the client hits the identical path — GET with no `init` at all, POST with `{ method: 'POST' }` and **no body**. That matches `docs/architecture-overview.md` §22: *"it holds **no operator picks**: a pick belongs to the confirm request, or the next build destroys it."*
- `changeId!: string | null` and `opened!: boolean` line up with the new FE types.
- The `recordCorrectionProposal` docblock's "**Issues nothing**" is the right thing to put in the type surface — #2374 is explicit that the proposal is *data*, and that holding one confers no authority to issue. Stating it where the next caller will read it is better than a spec asserting it.
- `encodeURIComponent(returnId)` on both.

### ✅ Tests

Five cases, proper Arrange–Act–Assert, and they assert the two things that actually matter for an API client: the **method** (including the negative — `expect(init?.method).toBeUndefined()` on the GET, which is what stops a future refactor quietly making the preview a write) and the **field passthrough**. The `opened: false` reuse case is covered distinctly from the absent case, which is the right split.

### 🟢 SUGGESTION — two comments say opposite things about what `opened: false` means

`return-proposal.schema.ts` (new comment):
> *Absent degrades to `false` — "not reported as freshly opened" — **never a claim that a row was reused**, which is a fact this build cannot invent.*

`returns.types.ts` (new docblock):
> *`false` **when an identical open proposal was reused** rather than opened.*

Those can't both hold: the type tells a consumer `false ⇒ reused`, the schema tells them `false ⇏ reused`. Today it's cosmetic — the backend DTO declares `opened!: boolean` non-nullable, so the absent branch is unreachable against the real API and `nullish()` is purely defensive. But #3123/#3087 is the consumer, and a UI that renders "Reused an existing proposal" off `opened === false` would be making exactly the claim the schema comment says it must not.

Two ways to settle it, either fine:

- **Keep the boolean**, and reword the type docblock to "`false` when the backend did not report a freshly-opened row — either an identical open proposal was reused, or the field was absent." Cheapest.
- **Make absence representable** — `opened: boolean | null`, `null` = not reported. That's the shape the repo already uses for exactly this distinction (`adjustmentOutcome` absent ⇒ `'unsupported'` never `'honoured'` in #2368; `olHeldNotReflected` `null` vs `0` in #2345; `idle` vs `unavailable` in `usePaginatedTotal`). Slightly more honest, slightly more work for the consumer.

I'd take the first — the field is required on the wire, so inventing a third state to describe a case that cannot occur is the more expensive option.

### Notes

- CI is mid-flight (`Build` / `Test` / `Integration Tests` still queued at review time; `Lint`, `Type Check`, `Test (web)` green). Approving on the assumption the rest stays green — please don't merge red.
- Nothing consumes `recordCorrectionProposal` yet, which is correct for this slice and matches how the rest of this epic has shipped (contract first, consumer next).

**Verdict:** ✅ Approve. The suggestion is a comment edit, not a blocker.

piotrswierzy (approved):
## Review — `feat(returns): correction-proposal API client (preview GET, record POST)`

Four files, 94 additions, and the base of the #3375–#3379 + #3399 stack — which is why it mattered that this left draft.

### Verified

**The response is parsed, not cast.** `request<unknown>(…)` then `parseCorrectionProposal(raw)`, with `return-proposal.schema.ts` widened to carry the ADR-044 `changeId` / `opened` fields. Typing the transport as `unknown` and narrowing through a schema is the right shape for a wire boundary — the tempting `request<ReturnCorrectionProposalResponseDto>` would have been an assertion the client cannot back.

**`encodeURIComponent(returnId)`** on the path segment.

**Goes through the shared `request` helper**, so no raw `fetch` in a feature — the `no-restricted-globals` rule in `docs/frontend-architecture.md § API Client Conventions` holds.

**Interface entry alongside the implementation**, so a consumer binds to the declared method rather than to the object literal.

**58 lines of tests for 30 lines of client**, which is the right ratio for a transport module whose failure mode is a silently mis-shaped response.

I also effectively exercised this surface from the other end while reviewing #3375: the mutation's `setQueryData` key matches `useReturnProposalQuery`'s, and the record response really is the preview's computation plus `changeId` / `opened` — #2374's service routes both entry points through one private `compute`. So the type this client threads is the right one.

### Note

Now that this is out of draft, the whole chain below it (#3375 → #3376 → #3377 → #3378 → #3379 → #3399) is unblocked — all six are approved. Worth landing them in order in one sitting rather than letting the stack drift again.

**✅ Approve.**

PR #3302, opened 2026-09-18
test(sales-documents): add the missing SalesDocumentRulesController spec

### LINKED ISSUE #3299: [TECH-DEBT] CORE — sales-document-rules.controller.ts has no spec, unlike its two siblings
## Problem / Context

`apps/api/src/sales-documents/http/sales-document-rules.controller.ts` (273 lines) has no
`.spec.ts`, while its two siblings in the same directory do:

- `sales-document-markets.controller.spec.ts`
- `sales-document-templates.controller.spec.ts`

The gap predates #3253, but that PR widened the untested surface: the new
`POST /sales-documents/rules/overlap-check` route ships with its own DTO-level validation
plus an `assertValidCountryParam` call, and neither is covered by anything.

Flagged in Piotr's review of #3253:

> The body already flags that this controller has no spec at all, unlike its two siblings,
> and defers it as out of scope for a re-land. Agreed on the scoping — but the new
> `rules/overlap-check` route ships with its own DTO-level validation and a
> `assertValidCountryParam` call, and that is now the second untested surface on the file.
> Worth the issue the body promises rather than leaving it at a paragraph.

## Proposed Solution

Write the first `sales-document-rules.controller.spec.ts`, following the shape of
`sales-document-markets.controller.spec.ts` / `sales-document-templates.controller.spec.ts`
(mock the injected application service, assert each route delegates + maps errors
correctly). Cover every route on the controller, with particular attention to:

- `POST /sales-documents/rules/overlap-check` — the DTO-level validation and the
  `assertValidCountryParam` call.
- The existing CRUD + reorder routes, which currently have no regression coverage at all.

## Classification

**Type**: CORE
**Layer**: Interface
**File(s)**: `apps/api/src/sales-documents/http/sales-document-rules.controller.ts`

## Dependencies

- None

## Assumptions

- None — this is pure test-coverage work with no production-code change implied.

## Acceptance Criteria

- [ ] `sales-document-rules.controller.spec.ts` exists and covers every route on the controller
- [ ] The `overlap-check` route's DTO validation and `assertValidCountryParam` call are covered

norbert-kulus-blockydevs:
Closes #3299

## Summary

`apps/api/src/sales-documents/http/sales-document-rules.controller.ts` had
no spec, unlike its two siblings (`sales-document-markets.controller.spec.ts`,
`sales-document-templates.controller.spec.ts`). #3253 widened the gap by
shipping `POST /sales-documents/rules/overlap-check` with no coverage of its
own DTO-level validation or the `assertValidCountryParam` call.

Adds the first spec, constructed directly (the `SalesDocumentTemplatesController`
precedent) rather than through a `TestingModule` — the controller's two
dependencies are plain injected values with no framework wiring under test.

## Coverage (33 tests)

- Every route: `listRules`, `createRule`, `checkRuleOverlap`, `deleteRule`,
  `listCountryDefaults`, `upsertCountryDefault`, `deleteCountryDefault`,
  `listThresholds`, `listConfiguredCountries`, `acknowledgeNoDocument`,
  `clearAcknowledgment`
- The capability-guard call (`assertConnectionSupportsKind`) on the two write
  paths that need it (`createRule`, `upsertCountryDefault`), and its absence
  on `checkRuleOverlap` (it is a read)
- `toHttpException` mapping for every domain exception the controller catches
  (conflict → 409, not-found → 404, invalid/threshold-not-found → 400,
  unrecognized error passes through unchanged)
- `assertValidCountryParam` on every route that calls it (empty, >8 chars,
  the `'*'` sentinel)
- The `overlap-check` route's condition DTO validation
  (`SalesDocumentConditionDto.toDomain`): a malformed `buyerHasTaxId`, a
  blank-after-trim `orderCountry`, a non-decimal-string `orderTotalGross`
  amount, and a wrong `op` for `orderTotalGross`

## Verification

- `npx jest --config apps/api/jest.config.js sales-document-rules.controller.spec.ts` — 33/33 passing
- `npx eslint` on the new file — clean

No production-code change, as the issue's own "Assumptions" section states.


piotrswierzy (approved):
## Tech Lead review — #3302

Reviewed per `.claude/commands/pr-review.md` (4-step) and `docs/code-review-guide.md`, against `docs/testing-guide.md`, `docs/engineering-standards.md` § Route authorization, and `docs/architecture-overview.md` § 17 Sales Documents (including the #3253 overlap-check bullet).

### Step 1 — Context

`SalesDocumentRulesController` (11 routes, class-level `@Roles('admin')` at `:60`) had no spec while both siblings did. #3253 widened the gap by adding `POST rules/overlap-check` with no coverage of its DTO gate or `assertValidCountryParam`. Test-only; no production change, as the issue's own Assumptions state.

### Step 2 — Systematic review

**Construction.** Direct `new SalesDocumentRulesController(service, guard)` rather than a `TestingModule`, citing the `SalesDocumentTemplatesController` precedent. Correct: the constructor takes two plain injected values and no framework wiring is under test. `docs/testing-guide.md`'s unit-test rule (mock all external dependencies) is satisfied, and a `TestingModule` here would only add a DI graph nobody is asserting.

**Are the DTO-validation tests genuine, or do they only believe they are?** This was the thing most likely to be wrong, and it is right. `SalesDocumentConditionDto.toDomain` is a **static** method (`dto/sales-document-condition.dto.ts`), invoked as `SalesDocumentConditionDto.toDomain(c)` at `sales-document-rules.controller.ts:121` — so passing a plain object literal exercises the real gate rather than hitting `c.toDomain is not a function`. Each of the four malformed-condition assertions maps to a distinct `throw new BadRequestException` in that method: missing `boolValue`, blank-after-trim `stringValue` (the #3190 review fix), non-decimal `amount`, wrong `op`. And `checkRuleOverlap` has **no** try/catch, so the `BadRequestException` propagates unwrapped — the assertion is exact, not incidental.

The PR body is honest about the scope here: it names `SalesDocumentConditionDto.toDomain` explicitly rather than claiming route-level validation. That distinction matters, because calling a controller method directly does **not** run `ValidationPipe`, so the `class-validator` decorators on `CheckSalesDocumentRuleOverlapDto` (`@IsDateString() effectiveFrom` at `:45`, `@ValidateNested({each:true})` at `:40`) are genuinely untested by this spec. They exist and are correct, so there is no production gap — only a coverage boundary the description states accurately.

**Route authorization.** The spec asserts no decorators, and that is the **right** call rather than an omission: `apps/api/src/auth/route-authorization-coverage.spec.ts:162` discovers controllers by filesystem walk (`walk(SRC_ROOT, n => n.endsWith('.controller.ts'))`), so this controller is already covered by the repo's single enforcement point, and `:304` even enforces the `*.controller.ts` naming convention the walk rests on. Duplicating it here would be a second copy of a rule that `docs/engineering-standards.md` § Route authorization deliberately keeps in one place. (Note: that spec fails locally on `main` with `Cannot find module 'sanitize-html'` — a local `pnpm install` artifact, unrelated to this PR; CI's Test job is green.)

**The #3253 asymmetry is respected, and positively asserted.** `docs/architecture-overview.md` § 17 is explicit that the overlap check is *advisory over HTTP and refusing only in the composer*. The spec does not "correct" that: `'should map the verdict into the response shape verbatim'` feeds a non-empty `overlapping` array and asserts the call **resolves with it**, never that it throws — which is the asymmetry pinned as behaviour. `'should never check the connection capability - it is a read'` pins the other half.

**Tautology scan.** No test asserts only what it mocked. Every delegation test asserts `toHaveBeenCalledWith` on real arguments (including the DTO→domain condition mapping at `:117` and `:173`); the `not.toHaveBeenCalled()` assertions on the rejection paths are what make the "before calling the service" claims real rather than incidental; `'should let an unrecognized error propagate unchanged'` uses `rejects.toBe(unknown)` (identity, not instance) which correctly pins the `toHttpException` fallthrough at `:270`.

**CI.** Test, Lint, Type Check, Build, Integration all green.

### Step 3 — Findings

**[IMPORTANT]** — the outbound mapping direction is never asserted
> `SalesDocumentRuleResponseDto.fromDomain` maps `conditions` through `SalesDocumentConditionDto.fromDomain`, and formats dates two different ways (`effectiveFrom`/`effectiveTo` as `.toISOString().slice(0,10)`, `createdAt`/`updatedAt` as full ISO). **None of that is asserted.** `listRules` uses `toMatchObject({ id, country, documentKind })` — 3 of 11 fields, and not `conditions`; `createRule`'s return value is never inspected at all. So `SalesDocumentConditionDto.fromDomain` has zero coverage in a spec for the controller whose entire subject is conditions.
>
> This is not cosmetic. `docs/architecture-overview.md` § 17 records that `orderTotalGross.amount` is *"a DECIMAL STRING, never a JSON number"* precisely because a number loses precision through `jsonb` and changes `conditionsHash`. A `fromDomain` regression that numberified `amount` on the way out would ship unnoticed here — and no `orderTotalGross` condition appears in any **response** fixture (`ruleFixture` at `:33` carries only `buyerHasTaxId`). Suggested: tighten `listRules` to `toEqual` (as `listCountryDefaults` already does at `:302`) and add one `orderTotalGross` rule to `ruleFixture` so the inbound gate and the outbound mapper are both exercised on the shape the docs single out. Per `docs/testing-guide.md`, this is the non-trivial logic on the response path.

**[SUGGESTION]** — assertion strictness is inconsistent
> `listRules` (`:75`) and `listThresholds` / `listConfiguredCountries` use `toMatchObject`, while `listCountryDefaults` (`:302`) and the overlap verdict (`:196-204`) use `toEqual`. `toMatchObject` on a response DTO cannot catch a field the mapper stops emitting. Prefer `toEqual` wherever the full shape is known — it is, in all five cases.

**[SUGGESTION]** — pin the audience, not just that one exists
> `route-authorization-coverage.spec.ts` asserts every route carries *one of* `@Public` / `@Roles` / `@AnyRole`; it does not assert **which**. A change from `@Roles('admin')` (`:60`) to `@AnyRole()` on a rules-authoring surface would still pass it. One assertion on `Reflect.getMetadata('roles', SalesDocumentRulesController)` would pin the admin-only contract at the point it is meaningful, without duplicating the coverage spec's job.

### Positive observations

- The `checkRuleOverlap` describe block is the strongest part: it covers the country guard, the domain mapping, the capability guard's **absence**, verbatim verdict passthrough including the `undecided` arm, four distinct DTO-gate refusals, and the `effectiveTo: undefined → null` coercion. That is the route #3253 shipped uncovered, and it is now covered on exactly the axes the ADR cares about.
- `'should let an unrecognized error propagate unchanged'` using `rejects.toBe` rather than `rejects.toThrow` correctly pins the identity-preserving fallthrough.
- `describe('toHttpException fallback (via deleteRule)')` names its own indirection honestly rather than pretending `deleteRule` realistically throws a threshold error — the right way to test a private mapper.

---

### Summary

A well-constructed, genuinely-asserting first spec for an 11-route controller that had none, and it gets the two things most likely to go wrong right: the DTO-gate tests exercise real logic (because `toDomain` is static, plain object literals reach the actual `BadRequestException` branches), and the #3253 advisory-not-refusing asymmetry is preserved and positively asserted rather than quietly "corrected". Route-authorization decorators are correctly left to the repo's existing single enforcement point. The one real gap is directional: the inbound DTO→domain mapping is covered thoroughly while the outbound `fromDomain` direction — including `conditions` and the two date formats — is never asserted at all, which leaves the `orderTotalGross` decimal-string invariant that § 17 singles out unprotected on the response path.

### Merge Readiness

🔄 **Approve with changes** — no BLOCKING issues; merge, then address the IMPORTANT item as a follow-up.

### Priority fixes

1. **(IMPORTANT, follow-up)** Assert the outbound direction: add an `orderTotalGross` condition to `ruleFixture` and tighten the `listRules` assertion to `toEqual`, so `SalesDocumentConditionDto.fromDomain` and the two date formats in `SalesDocumentRuleResponseDto.fromDomain` are actually exercised.
2. **(SUGGESTION)** Use `toEqual` consistently across the five list assertions.
3. **(SUGGESTION)** Pin `@Roles('admin')` on the class, which the repo-wide coverage spec deliberately does not check.

PR #3303, opened 2026-09-18
fix(invariants): floor the scanned-file/root/phrase count on six check-* guards

### LINKED ISSUE #2792: Three check-* guards go silently inert on a path/CWD/phrase change — floor the scanned count
Third and lowest-severity output of the #2673 sweep. None of these is inert *today*; each goes inert on a plausible future change, with a green tick and a reassuring success line. Grouping them because they share one cheap fix.

## 1. `scripts/check-stream-writes.mjs:32` — inert under any non-root CWD

```js
const REPO_ROOT = process.cwd();
```

The only one of 37 scripts deriving its root from CWD rather than `import.meta.url`. `walk()` swallows ENOENT and nothing floors the scanned count:

```
$ cd scripts && node check-stream-writes.mjs
✓ check-stream-writes: 0 source file(s) checked. All writes are bounded.   # exit 0
```

From the repo root it scans 3409 files. Safe only because `package.json` happens to invoke it from root — any hook, `pnpm --filter`, or CI step with a different CWD silently disarms the entire `bare-xadd` guard that ADR-049's stream-retention invariant depends on.

**Fix:** `__dirname`-derived root (as every sibling does), plus a non-zero floor on the file count.

## 2. `scripts/check-architecture-gates.mjs:266` — partly inert

`findUnmarkedInlineGates` keys on the literal phrase `/\*Reversal gate/`. Zero matches is a pass with no minimum asserted. ADRs **053, 057, 058, 063, 064, 066 contain the phrase zero times** and are checked vacuously right now. One convention reword takes the rule fully inert across all 21 in-scope ADRs while still printing "gates marked in ADR-048+". The `--self-check` pins the phrase only against synthetic fixtures, never live ADR text.

## 3. `scripts/check-outbound-http.mjs:110` — fragile, and its success line lies

11 hand-listed `SCAN_ROOTS`, ENOENT → silent return, and the success line reports `SCAN_ROOTS.length` (roots *declared*) rather than roots *found* — so it would print "scanned 11 roots" while scanning 10. All 11 resolve today.

Same shape, lower stakes: `check-service-interfaces.mjs:190`, `check-nul-bytes.mjs:185`, `check-libs-build-scripts.mjs:107` each print a count and exit 0 if it is zero.

## The shared fix

`scripts/check-workspace-dep-declarations.mjs` already carries the rule the rest lack, in its own words:

> *"A guard that passes because it discovered nothing has silently stopped guarding"*

and `check-no-supported-actions-mirror.mjs` implements it (`scanned === 0` is fatal). Apply the same two rules across the walker-based guards:

- [ ] `scanned === 0` (or below a declared floor) is FATAL, not a pass.
- [ ] Roots derived from `import.meta.url`, never `process.cwd()`.
- [ ] Success lines report what was *found*, never what was *declared*.
- [ ] `check-architecture-gates` asserts a minimum live phrase count, so a reworded convention fails loudly.

Related: #2673, #2790, #2791.

norbert-kulus-blockydevs:
## Summary

Six `check-*` invariant scripts printed a green "passed" line even when they had scanned almost nothing — a moved directory, a CWD change, or a reworded convention phrase would silently disarm the guard instead of failing the build. This mirrors the rule `check-workspace-dep-declarations.mjs` and `check-no-supported-actions-mirror.mjs` already follow (*"a guard that passes because it discovered nothing has silently stopped guarding"*), applied to five scripts that hadn't caught up, plus `check-architecture-gates.mjs`'s `REPO_ROOT` anchoring.

- `scripts/check-architecture-gates.mjs`
  - `REPO_ROOT` was `process.cwd()` — running the script from any other directory threw an uncaught `ENOENT` out of `checkGateMarkers` instead of failing with a diagnosable message. Anchored to the script file, like every sibling script.
  - Added a floor (`MIN_GATE_MARKER_OCCURRENCES = 20`, live count today is 38) on the total occurrences of the `*Reversal gate` convention phrase across ADR-048+. The existing rule only reports on *unmarked* occurrences, so a wholesale rewording of the phrase would silently read as zero violations forever — the floor catches that.
- `scripts/check-outbound-http.mjs`
  - The walker silently swallowed `ENOENT` per declared root, and the success line reported `SCAN_ROOTS.length` (declared) rather than what was actually found and scanned.
  - Now tracks `rootsFound` vs `missingRoots`, reports the *found* count, and fails if any declared root is missing or if none resolved at all.
- `scripts/check-service-interfaces.mjs`, `scripts/check-nul-bytes.mjs`, `scripts/check-libs-build-scripts.mjs`
  - Each already anchors its root correctly via `import.meta.url`. Added a floor on the scanned/discovered count (50 services, 100 text files, 5 packages — each comfortably below the real counts of ~160/~7300/17 but well above zero) so a moved directory or drifted naming convention fails the build instead of reporting a clean pass over almost nothing.
- `scripts/check-stream-writes.mjs` — already fixed by a prior branch (`ee9fe6fe4`); no change needed here.

Every fix was verified **red-first**: reproduced the pre-fix "0 (or near-0) scanned, exit 0" condition against the original code, then confirmed the fix converts it into a loud, non-zero-exit failure with an actionable message. Restored every temporary repro back to the clean fixed state afterward.

## Test plan

- [x] `pnpm check:invariants` — full suite green, including all `--self-check` modes and the live runs for the six touched scripts.
- [x] `pnpm lint` — zero errors (pre-existing unrelated warnings only).
- [x] `pnpm type-check` — clean across every package.
- [x] Red-first repro for each of the 6 scripts: confirmed the pre-fix silent-pass condition, then confirmed the fix fails loudly on the identical repro, then restored the clean state.

Closes #2792


piotrswierzy (approved):
## Tech Lead review — #3303

Reviewed per `.claude/commands/pr-review.md` (4-step) and `docs/code-review-guide.md`. Read-only; the branch's five scripts were staged into a scratch tree so both the pre-fix and post-fix behaviour could be executed, not inferred.

### Step 1 — Context

`#2792` is the same defect family as `#3002`: a `check-*` script that scans nothing prints a green line and exits 0. `docs/architecture-overview.md` treats this as first-order — a guard that passes because it discovered nothing has silently stopped guarding.

### Step 2 — Verification (demonstrated, not assumed)

**The precedent claim in the body is accurate.** `scripts/check-workspace-dep-declarations.mjs:303-311` carries the exact comment the body quotes and guards `packages.length === 0`; `scripts/check-no-supported-actions-mirror.mjs:161-168` guards `scanned === 0` with *"Refusing to report a pass over an empty scan."* The new code is the same shape (count → stderr → `exit(1)`), with a non-zero floor instead of `=== 0` — stronger, not a second idiom.

**`main` silently passes on a broken tree.** Running each script from a near-empty root (the "directory moved" condition):

| script | main | PR |
|---|---|---|
| `check-nul-bytes` | `✓ 7 text files carry no literal NUL bytes.` **exit 0** | `✗ only 7 text file(s) … expected at least 100` **exit 1** |
| `check-service-interfaces` (0 services) | `✓ 0 core application service(s) checked. All implement an interface.` **exit 0** | `✗ only 0 service(s) … expected at least 50` **exit 1** |
| `check-libs-build-scripts` | `OK (0 packages)` **exit 0** | `only 0 package(s) … expected at least 5` **exit 1** |
| `check-outbound-http` | `OK (scanned 11 roots)` **exit 0** | `11 of 11 declared root(s) could not be found` **exit 1** |
| `check-architecture-gates` | — | **exit 1** |

`check-outbound-http` on `main` is the sharpest case: it reported *"scanned 11 roots"* having scanned **none**, because the success line printed `SCAN_ROOTS.length` rather than what resolved. That is not merely a missing signal, it is a false one. **All five PR versions fail loudly on the identical input.**

**The gate-occurrence floor fires.** `--self-check` goes 13 → 16 cases (I ran both), the three new ones covering `countGateOccurrences` counting marked and unmarked alike, a count above the floor passing, and a count of zero tripping it. Live corpus confirmed at **exactly 38** occurrences across ADR-048+ (`grep -o '\*Reversal gate'`, 15 files), matching the body.

**Every floor literal in the body is exact**, re-measured against `main`: 7340 text files (floor 100), 160 services (floor 50), 17 packages (floor 5), 11 declared roots, 38 gate occurrences (floor 20). Each floor is ≥3× below the live count, so routine additions/deletions cannot trip it and it will not need quiet bumping — which is the right answer to floor-rot. Each carries a docblock naming the live count, the reason, and when a *lower* value is legitimate (`check-architecture-gates.mjs:406-412` is the clearest).

**The `REPO_ROOT` anchoring is correct for every invocation path.** `pnpm check:invariants` and the husky hook (`.husky/pre-commit` → `pnpm lint` → `pnpm check:invariants`) both run from the repo root, so `process.cwd()` happened to work there — the fix is for the third path, and it reproduces on `main`:

```
$ cd scripts && node check-architecture-gates.mjs
… code: 'ENOENT', path: '…/openlinker/scripts/docs/architecture/adrs' }   exit=1
```

`resolve(dirname(fileURLToPath(import.meta.url)), '..')` is byte-identical to the idiom already used by `check-nul-bytes.mjs:32`, `check-css-structure.mjs:69`, `check-types-sub-barrels.mjs:54` and `check-stream-writes.mjs:41` — no second idiom invented, exactly as the body claims.

### Step 3 — Findings

**[IMPORTANT]** — `scripts/check-nul-bytes.mjs:196`, `check-service-interfaces.mjs:204`, `check-libs-build-scripts.mjs:120`, `check-outbound-http.mjs:212`
> Four of the five new floors are inline `if (count < MIN) { … exit(1) }` inside `main()`, so they are unreachable from `--self-check` and have **no regression case proving they can fail**. This PR's own fifth script shows the better shape: `check-architecture-gates.mjs` extracts the pure `evaluateGateOccurrenceFloor` (`:416-428`) and self-checks it in three cases (`:702-717`). Two of the four scripts already run `--self-check` in the invariants chain, and both self-checks already exercise pure helpers (`findNulOffsets`, `evaluateServiceFile`) — so the extraction is a small one. The reason this matters more than usual here: a PR whose entire subject is *"a check that cannot fail has stopped checking"* should leave behind a fixture that fails if someone later inverts a comparison or drops a floor. I confirmed empirically that all four fire **today**; what is missing is the proof that they still will. Follow-up, not a merge blocker — see `docs/testing-guide.md` on covering non-trivial logic, and the `#2404` red-first-fixture discipline in `docs/architecture-overview.md` § 26.

**[SUGGESTION]** — `scripts/check-architecture-gates.mjs:388`
> A genuinely missing `ADR_DIR` (renamed, not merely a wrong CWD) still throws a raw `ENOENT` out of `readdirSync` before the floor can run — verified by moving the directory in the staged copy. Exit code is 1, so the guard does **not** silently pass and this is not a defect; but the docblock at `:75-81` says the fix converts the `ENOENT` "into a diagnosable failure", which is true only for the wrong-CWD case. A guarded `existsSync(adrDirAbs)` with a named message would make the claim hold in both cases (`existsSync` is already imported at `:73`).

**[SUGGESTION]** — `scripts/check-nul-bytes.mjs:196` and `check-outbound-http.mjs:212`
> Both floors sit *after* the violations `exit(1)`, while `check-service-interfaces.mjs:204` and the two precedent scripts put the emptiness guard *first*. When both conditions hold, the reader sees the violation list from a scan that was broken anyway. Ordering the floor first matches `check-no-supported-actions-mirror.mjs:161` and tells the reader the more important thing.

**[SUGGESTION]** — `scripts/check-outbound-http.mjs:213`
> `missingRoots.length > 0 || rootsFound === 0`: the second arm is reachable only when `SCAN_ROOTS` itself has been emptied — which is a real drift and correctly caught — but the `${missingRoots.join(', ') || '(none found at all)'}` fallback reads as if it covers the all-roots-missing case, which the first arm already reports by name. A one-line comment on the empty-`SCAN_ROOTS` case would stop the next reader deleting the arm as redundant.

### Positive observations

- The floors are deliberately *low with large headroom* rather than tracking the live count. That is the correct answer to rot: a floor at the exact count needs bumping on every addition and is then raised until it means nothing. The trade-off (a walk degraded to 3× still passes) is inherent and is the same one `check-workspace-dep-declarations.mjs` makes at `=== 0`.
- `check-outbound-http` now distinguishes *declared* from *found* roots. Reporting the count you wanted to scan is the precise shape of the #2792 defect, and fixing the success line matters as much as the new failure branch.
- `check-stream-writes.mjs` correctly identified as already fixed and left alone — the audit did not re-touch a file for symmetry.

---

### Summary

Six guards that could print green having scanned nothing now fail loudly, and I verified both halves rather than taking them on trust: `main`'s versions exit 0 on a near-empty tree (including `check-outbound-http` asserting *"scanned 11 roots"* having scanned none), and all five PR versions exit 1 on the identical input. Every floor literal in the description is exact against the live tree, each sits ≥3× below its real count so it will not rot into a spurious failure, and the `REPO_ROOT` anchoring matches four existing siblings exactly and is correct for all three invocation paths. The one thing worth following up is that four of the five floors are inline in `main()` with no `--self-check` fixture — the PR's own fifth script demonstrates the better shape, and a guard about guards ought to leave its own regression proof behind.

### Merge Readiness

🔄 **Approve with changes** — no BLOCKING issues; merge, then address the IMPORTANT item as a follow-up.

### Priority fixes

1. **(IMPORTANT, follow-up)** Extract the four inline floors into pure evaluators and add `--self-check` cases, mirroring `evaluateGateOccurrenceFloor`/`check-architecture-gates.mjs:416`. Without a fixture, the floors are correct today and unprotected tomorrow.
2. **(SUGGESTION)** Guard `readdirSync(adrDirAbs)` with `existsSync` so a renamed `ADR_DIR` produces the diagnosable message the docblock promises.
3. **(SUGGESTION)** Move the `check-nul-bytes` / `check-outbound-http` floors above the violations `exit(1)`, matching the precedent scripts.
4. **(SUGGESTION)** Comment the empty-`SCAN_ROOTS` arm in `check-outbound-http.mjs:213`.

PR #3304, opened 2026-09-18
feat(eparagony): eparagony issues invoices through the Invoicing capability

### LINKED ISSUE #3192: feat(eparagony): issue invoices through the Invoicing capability

### LINKED ISSUE #3274: [BUG] Integration — eparagony invoicing crashes on a persisted `merchantAddress: null`, permanently blocking the order's sales document
## Problem / Context

`EparagonyConnectionConfigShapeValidatorAdapter` and `composeInvoiceDocument` disagree about what `merchantAddress: null` means. The validator treats it as absent; the mapper treats it as present and dereferences it. The result is a raw `TypeError` on the live `issueInvoice` path, which core classifies as `in-doubt` and which then **permanently blocks that order from receiving any sales document on any connection**.

Found while reviewing #3267. Not introduced by that PR alone: the mapper guards come from #3260 (slice 1) and the validator arm plus the spec that pins it come from #3267 (slice 2). The defect only exists once both are on the same branch, which they are on `3192-eparagony-invoicing`.

**The chain, verified link by link on `3192-2-invoicing-wiring`:**

1. `eparagony-connection-config-shape-validator.adapter.ts:215` - `validateMerchantAddress` returns early on `raw === null`, so an explicit `null` is accepted and persisted. `eparagony-shape-validators.spec.ts:92` pins that reading by name (*"should treat an explicit null as absent, the way an operator-authored blob carries one"*).
2. `domain/policies/connection-config.policy.ts` - `readEparagonyConnectionConfig` is a documented blind `(raw ?? {}) as unknown as EparagonyConnectionConfig`. TypeScript cannot catch the mismatch, by design.
3. `eparagony-config.types.ts:167` declares `merchantAddress?: EparagonySellerAddress`, so a `!== undefined` guard *looks* correct while being wrong at runtime for `null`.
4. `eparagony-invoice.mapper.ts:341` - `if (config.merchantAddress !== undefined)` is **true** for `null`, so `toSellerEntityAddress(null)` runs and its first statement is `readNonEmpty(address.apartment)`. **`TypeError: Cannot read properties of null`.**
5. `eparagony-invoice.mapper.ts:395` - `toIssuedDocumentSeller` has the identical `address === undefined` guard and falls through to `address.street`. Same crash, **different position in the flow** (see below).
6. `eparagony-invoicing.adapter.ts:163` - nothing wraps `composeInvoiceDocument`, and the comment directly above it asserts the invariant that is broken: *"Composition failures throw `EparagonyConfigException` (`failureMode: 'rejected'`) BEFORE anything crosses the boundary."*
7. `libs/core/src/invoicing/application/services/invoice.service.ts:758-761` - the adapter call is wrapped, and `classifyFailure` reads `(error).failureMode`, finds nothing on a plain `TypeError`, and returns `'in-doubt'` by its own fiscal-safe default.
8. `invoice-record.entity.ts` - `blocksIssuanceElsewhere` returns `true` for `failed` + non-`rejected`. Both cross-kind guards are connection-agnostic, so the order can no longer receive an invoice **or** a fiscal receipt on any connection, for a request that never left the process.

**Reachability.** `EditConnectionForm` carries a raw JSON config editor (`configText` / `JSON.parse`), so an operator can save `"merchantAddress": null` today without curl. It is currently latent rather than common, because no form writes these keys yet. It becomes the normal path the moment the slice-3 seller-config form lands: the repo's own cleared-knob convention (`docs/architecture-overview.md`, #2610) is that clearing a knob writes an **explicit `null`** rather than deleting the key, and that every reader treats `null` exactly like absent.

## Proposed Solution

**Fix the mapper, not the validator.** The validator's `null`-is-absent reading is the one that matches #2610 and the three sibling invoice keys (`merchantTIN`, `merchantName`, `eInvoicingHubEnabled`, all of which are null-safe via `readNonEmpty` / `=== true`). It is the mapper that is out of step.

In `libs/integrations/eparagony/src/infrastructure/adapters/eparagony-invoice.mapper.ts`:

- `:341` - `if (config.merchantAddress !== undefined)` becomes `!= null`
- `:395` - `address === undefined` becomes `address == null`

**Both must change in the same commit.** Fixing only `:341` relocates the crash rather than removing it: `toIssuedDocumentSeller` is called from `toIssueResult` (`eparagony-invoicing.adapter.ts:518`), which runs at `:175` - **after** `createDocument` (`:172`) and the settlement poll have already succeeded. That turns a pre-call crash into one that loses a document the vendor legally created. And `merchantTIN` is mandatory by the time it runs, so `taxId` is always non-null there: any config carrying a `merchantName` plus a null address reaches it.

Keep `eparagony-shape-validators.spec.ts:92` as-is. It encodes the correct convention.

## Classification

**Type**: Integration
**Layer**: Infrastructure (adapter/mapper)
**File(s)**:
- `libs/integrations/eparagony/src/infrastructure/adapters/eparagony-invoice.mapper.ts` (`:341`, `:395`)
- `libs/integrations/eparagony/src/infrastructure/adapters/__tests__/eparagony-invoice.mapper.spec.ts` (new cases)
- unchanged, cited for context: `eparagony-connection-config-shape-validator.adapter.ts:215`, `eparagony-shape-validators.spec.ts:92`

## Dependencies

- #3260 (slice 1) and #3267 (slice 2) must both be on the branch for the defect to exist. Fix on `3192-eparagony-invoicing` or fold into #3267 before it merges - whichever the stack owner prefers.

## Assumptions

- The `null`-is-absent convention is the intended one, per #2610 and the three sibling keys. If instead the decision is that `merchantAddress` must be refused when explicitly `null`, the fix inverts: reject it in `validateMerchantAddress` and invert the spec at `:92`. Either is correct as long as **both sides agree**; this issue picks the one that matches house style.
- The vendor's behaviour on a whitespace-padded or absent seller address is not in scope here; that is a separate review suggestion.

## Acceptance Criteria

- [ ] `composeInvoiceDocument` with `merchantAddress: null` does not throw, and omits `metadata.merchantAddress` exactly as it does for `undefined`
- [ ] `toIssuedDocumentSeller` with `merchantAddress: null` returns `null` rather than throwing, even when `merchantName` and `merchantTIN` are both present
- [ ] A spec drives `composeInvoiceDocument` with `merchantAddress: null` (not only the validator), so the validator's "null is absent" test has a counterpart proving the mapper agrees
- [ ] A spec covers the post-`createDocument` path (`toIssueResult` / `toIssuedDocumentSeller`) with a null address, so a future regression cannot be caught only before the boundary
- [ ] `eparagony-shape-validators.spec.ts:92` still passes unchanged
- [ ] The stale comment at `eparagony-invoicing.adapter.ts:161-162` is either true after the fix or corrected

norbert-kulus-blockydevs:
Closes #3192.
Closes #3274.

Part of the #3173 epic (Wave D).

## What this merges

The full Wave D stack, reviewed and merged incrementally onto the integration branch `3192-eparagony-invoicing` per the epic's stacking strategy (`docs/architecture-overview.md` — a 🔗 sub-issue reaches `main` as **one** merge of its integration branch, never as intermediate stack members merging straight to `main`):

- #3260 — `EparagonyInvoicingAdapter implements InvoicingPort, RegulatoryStatusReader`, plus the invoice mapper, tax-rate policy and residual-clamping logic. Nothing reachable yet by design.
- #3267 — makes the adapter reachable: manifest advertises `Invoicing`, dispatch table resolves it, factory constructs it, config-shape validator learns `merchantTIN` / `merchantName` / `merchantAddress` / `eInvoicingHubEnabled` (all optional — no existing connection is retroactively granted the capability, per #2085's stamped-at-create rule). Includes a review-fix for a `merchantAddress: null` crash on the live issue path (#3274).
- #3269 — names the operator-facing surfaces for the second role: the `pending-submission` waiting states, and a `capability-routing-note` on the connection capabilities panel explaining what enabling `Invoicing` actually does.

**Note on #3274**: #3267 already carried `Closes #3274` in its own body, but #3267 merged into the integration branch, not the default branch, so GitHub never fired the auto-close on that keyword. Repeating it above is what actually closes it.

## Merge-conflict note (transparency)

#3269's branch had drifted from #3267's final reviewed state (both squash-merged into the integration branch, so their git histories don't share ancestry with what #3269 was based on). Resolving the merge conflict took `3192-eparagony-invoicing`'s side wholesale on the 8 affected eparagony backend files — verified each was a strict superset of what #3269 carried (more complete address-validation, better inline documentation, no regressions), and confirmed none of #3269's own commits (the actual frontend surfaces work) touched those files. No functional change was lost on either side.

## Verification (this merge, full monorepo)

- `pnpm build` — 22/22 projects clean
- `pnpm type-check` — 21/21 projects clean
- `pnpm lint` — 0 errors (pre-existing warnings only, no autofix churn)
- `pnpm --filter @openlinker/integrations-eparagony test` — 298/298 tests, 14/14 suites
- `pnpm check:invariants` — 90/90 scripts pass

## No migration

Adding `Invoicing` to the manifest does not grant it to existing connections — `enabledCapabilities` is stamped at create and never retro-filled. Operators opt in via the connection capabilities checkbox.


piotrswierzy (approved):
# Tech Lead review — #3304 (integration branch → `main`)

Reviewed as the **merged whole**, not as a re-read of the three stack members. I concentrated on what an incremental review structurally could not see: the ADR-041 §3a/3b trace on the combined shape, the guard scripts, cross-cutting invariants, and merge safety.

**Verdict: 🔄 Approve with changes — 0 BLOCKING / 3 IMPORTANT / 3 SUGGESTION.**

Nothing here can produce two fiscal documents for one sale. Approving; the IMPORTANTs are follow-ups, one of them cross-PR with #3268.

---

## 1. The merge-order hazard I flagged: **resolved**

Both halves ship together on the integration branch:

- `libs/integrations/eparagony/src/eparagony.constants.ts` declares `EPARAGONY_ISSUE_DEADLINE_MS = 110_000` (and its `EPARAGONY_REGISTER_DEADLINE_MS` twin).
- `scripts/check-eparagony-deadline-mirror.mjs` ships.
- `package.json` registers it under `check:invariants`, with the `--self-check` pass first, in the position the siblings use.

Verified the guard would actually run green rather than vacuously: both `MAX_SUPPORTED_PROVIDER_TIMEOUT_MS` declarations it reads exist and parse — `libs/core/src/invoicing/application/services/invoice.service.ts:155` and `libs/core/src/fiscalization/application/services/fiscal-registration.service.ts:143`, both `120 * 1000`, both matched by `parseNumericConst`'s product branch. `110_000 < 120_000` on both lanes.

## 2. Guard-script quality: **can actually fail**

`check-eparagony-deadline-mirror.mjs` meets the bar this repo sets for a guard:

- It parses **real sources** on both sides (`readFile` of the plugin constants and each core service), not a fixture.
- It has the **`null`-means-failure** contract, stated and implemented: `parseNumericConst` returns `null` for an absent or non-literal declaration, and `checkLane` turns that into `"could not read X"` → exit 1. A renamed constant is drift, not a pass.
- `--self-check` exercises both the parser and the comparator against synthetic input, including the two vacuity cases (`unreadable fails`, `unreadable core fails`) and the boundary (`equal fails`).
- It reads the two core constants **independently** rather than assuming they track each other — the right call, they are separate declarations in separate services that happen to agree today.

No objection. The only note is (S1) below.

## 3. ADR-041 §3a/3b on the merged whole: **holds**

This is the invariant most at risk from the new shape — one connection, one `adapterKey`, now advertising **both** `Fiscalization` and `Invoicing`. Traced rather than taken on trust:

| Question | Answer | Evidence |
|---|---|---|
| Is `InvoiceService`'s blocking read connection-agnostic? | Yes | `invoice.service.ts:497-501` — `repo.findAllByOrderId(orderId)` then `.find(r => r.connectionId !== requested && r.blocksIssuanceElsewhere)` |
| Is the **cross-kind** read connection-agnostic? | Yes, and *more* so | `invoice.service.ts:580-583` — `records.find(r => r.blocksFurtherRegistration)` with **no** connection exclusion, so the same eparagony connection's own receipt blocks its own invoice. This is exactly the case the dual-capability connection creates, and it is closed. |
| Is the reverse read likewise? | Yes | `fiscal-registration.service.ts:650-651` (same-kind) and `:676` `invoiceService.findBlockingInvoiceForOrder(orderId)` → `invoice.service.ts:1241`, `findAllByOrderId` + `find(r => r.blocksIssuanceElsewhere)`, no connection filter |
| Same per-order lock on both sides? | Yes | `invoice.service.ts:295` and `fiscal-registration.service.ts:323-324`, both `invoiceIssueLockKey(orderId)` |
| Exactly one routing candidate per connection? | Yes | `auto-issue-trigger.service.ts:387-392` — `connectionPort.list()` returns each connection once, filtered by `Invoicing ‖ Fiscalization`; `:430` is a `connections.map(...)`. A dual-capability connection cannot become two candidates. |
| `documentKind` from config, never from capabilities? | Yes | `:431` `readSalesDocumentRouting(connection.config)` → `sales-document-kind.types.ts:107-108`, reads `config.salesDocument.documentKind` only |

The PR touches **zero** `libs/core` files, so none of the above moved. Stated positively because "we didn't change core" is the claim that most needs checking on a stack merge, and it checks out.

## 4. ADR-026 country-agnosticity: **holds**

`0` `libs/core` files in the diff; no `NIP` / `KSeF` / `FA` / `ZRD` / `ZRICS` vocabulary added anywhere outside `libs/integrations/eparagony` and `docs/`. The frontend's routing note is keyed on `SALES_DOCUMENT_CAPABILITIES = ['Invoicing','Fiscalization']` (`ConnectionCapabilitiesPanel.tsx`) — a statement about ADR-041 routing, true of any connection holding either, correctly placed in the shared panel rather than a provider plugin.

## 5. Migrations: **none**, correctly

`0` files under `migrations/`. The PR body's reasoning is right and matches #2085: `enabledCapabilities` is stamped at create and never retro-filled, so adding `Invoicing` to the manifest grants it to nobody. See I1 for the half of that claim which is only true through the browser.

## 6. Merge safety: **safe**

Branch is `BEHIND` by exactly one commit — `e1ae0aece feat(web): show release version…` (#3288). Intersection of `merge-base..origin/main` changed files with this PR's 48 files is **empty**. A textual merge is safe; update the branch and land it.

---

## Carried-forward stack-member findings

### #3260

| # | Finding | Status |
|---|---|---|
| 1 | Deadline enforced as a poll-phase budget, not a whole-call wall clock | **Partially addressed** — see I2 |
| 2 | A vendor-`ERROR` invoice is permanently unretryable under the deterministic key | **Addressed as a documented trade-off** — `eparagony-invoicing.adapter.ts:520-527` now states the consequence in place ("because the vendor replays the same key idempotently, a re-attempt under the same registration key…"), and `resolveRegistrationKey` honours a caller-supplied `cmd.idempotencyKey` first, so an operator-supplied key is the escape hatch. Recording the cost rather than hiding it is the right resolution at this stage; the affordance is still absent from the UI. |
| 3 | A neutral `'0'` always becomes the vendor's **domestic** `ZRD` | **Addressed as a documented escape hatch** — `invoice-tax-rate.policy.ts:29,142-144`: a vendor code (`ZRICS` / `ZRE` / `NS2`) is now accepted **verbatim** through the neutral rate string, and the `ZRD` default is justified from the vendor's own contract ("use as the default 0% rate"). The honest residual is that an operator must type a Polish vendor code into their shop's tax field for an intra-EU supply or export; a shop storing plain `0` still mis-declares. That is now a stated limitation rather than a silent one. → **S2**. |

### #3267

| # | Finding | Status |
|---|---|---|
| 1 | `Invoicing` opt-in enforced **only in the browser** | **Not addressed** → **I1** |

---

## IMPORTANT

### I1 — the `Invoicing` opt-in is still browser-only; `POST /connections` mints an invoicing-claiming eparagony connection with no seller config

`apps/api/src/integrations/application/services/connection.service.ts:649-652` still carries exactly one carve-out from the manifest default — the #1498 `OfferManager`/`InventoryMaster` one:

```ts
const defaultCapabilities = metadata.supportedCapabilities.includes('InventoryMaster')
  ? metadata.supportedCapabilities.filter((c) => c !== 'OfferManager')
  : [...metadata.supportedCapabilities];
const enabledCapabilities = rest.enabledCapabilities ?? defaultCapabilities;
```

`Invoicing` is not excluded, so a `POST /connections { platformType: 'eparagony' }` that omits `enabledCapabilities` is now granted it. The PR knows this — `eparagony-setup-form.test.tsx` asserts the wizard passes `enabledCapabilities` explicitly, with the comment *"Omitting `enabledCapabilities` would do exactly that from #3192 slice 2 on, because the server then defaults to the adapter manifest's supported set — which now lists `'Invoicing'`."* That is the defence living in the one place #2610 says it cannot live: the raw JSON editor, curl and MCP all bypass the browser.

**Blast radius is bounded, which is why this is IMPORTANT and not BLOCKING.** Such a connection is not a routing candidate (no `config.salesDocument.documentKind` → `readSalesDocumentRouting` returns `documentKind: null`), and the write path refuses before crossing the boundary (`composeInvoiceDocument` throws `EparagonyConfigException`, `failureMode: 'rejected'`). So no document is issued and no double-issue is possible. What it does produce is a connection whose capabilities panel says it can invoice when it structurally cannot, and an operator discovering that at the first manual `POST /invoices`.

The exact in-tree precedent for the fix is three lines above the block quoted: extend the same filter so `Invoicing` is dropped from the default when the manifest also declares `Fiscalization`, or gate it on the seller keys being present.

### I2 — `EPARAGONY_ISSUE_DEADLINE_MS` is now anchored correctly, but the whole call can still outlive it

Improved since #3260: `eparagony-invoicing.adapter.ts:180` computes `const deadline = Date.now() + EPARAGONY_ISSUE_DEADLINE_MS` **before** `createDocument`, and `:380` caps the poll at `Math.min(deadline, …)`, so the poll can no longer stack on top of the create. The constant's docblock and the guard's docblock are now both true of the poll phase's contribution.

Two paths remain outside the anchor:

1. **`createDocument` itself is unbounded by the deadline.** Its budget is the HTTP client's: `REQUEST_TIMEOUT_MS = 15_000` (`eparagony-http-client.ts:53`) × `maxRetries: 2` + backoff (`eparagony-http-client.types.ts:24-29`), plus a token acquisition per attempt, and a 401 mid-session grants an extra `continue` that does not consume the attempt budget.
2. **The poll issues one status read before it checks the deadline.** `pollToSettledIssuance` reads first (`:383`) and only then computes `remaining = pollUntil - Date.now()` (`:398`). If the create already burned the budget, one more full GET (3 attempts × 15 s + backoff) still goes out.

Worst case is therefore create-budget + one status-read budget, which exceeds both `110_000` and core's `120_000`. Core does **not** race a timeout on this lane — `invoice.service.ts` has no `withTimeout`/`ProviderTimeoutError` equivalent to `fiscal-registration.service.ts:846-865`, and `:149` explicitly makes honouring the ceiling the adapter's job. Cheapest honest fix: check `Date.now() >= deadline` at the top of the poll loop, and pass a `timeoutMs` derived from the remaining budget into `createDocument`'s `post`. A `ISSUING_LEASE_MS` of 5 min gives real slack, so this is a hardening item rather than a live double-issue.

### I3 — the invoicing lane introduces its own poll default, which #3268's mirror guard cannot see

`eparagony-invoicing.adapter.ts:123` declares `const DEFAULT_STATUS_POLL_TIMEOUT_MS = 45_000`, local to this adapter, while the fiscalization lane keeps `60_000` (`eparagony-fiscalization.adapter.ts:76`). The per-lane split is well argued in the docblock and I am not asking for it to be collapsed.

The interaction is with #3268, in flight against the same `main`. `statusPollTimeoutMs` is **one** config key governing both lanes, and #3268's form states its default in the field description, mirrored from the fiscalization adapter alone — `check-eparagony-config-mirror.mjs` pins `BACKEND_ADAPTER = …/eparagony-fiscalization.adapter.ts`. Once both land, the form asserts "defaults to 60 s" about a key that defaults to 45 s on the invoicing lane, and the guard passes because it never reads the invoicing adapter. That is the reported-drifts-from-enforced failure #2229 names, arriving through a seam neither PR alone can see.

Whichever lands second should either add the invoicing lane to the mirror's lane table and reword the description to name both defaults, or make the two defaults agree.

---

## SUGGESTION

- **S1 — `check-eparagony-deadline-mirror.mjs` bounds the plugin constant from above only.** `checkLane` fails on `pluginValue >= coreValue` and says nothing about a floor. A deadline accidentally set to `100` ms would pass the guard while making every issuance time out. Not urgent (the module-load assertion against `MAX_STATUS_POLL_TIMEOUT_MS` catches the pathological case indirectly), but a one-line lower bound would make the guard say what it means.
- **S2 — surface the `ZRICS` / `ZRE` escape hatch where an operator will find it.** The policy docblock is the only place it is written down; an operator with an export line will reach for `0`. A sentence in `libs/integrations/eparagony/docs/setup-guide.md` costs nothing.
- **S3 — the two new `data-testid` docblocks both claim to be "the third".** `alert.tsx` says "after `Combobox` and `StatusBadge`" and `status-badge.tsx` says "after `Combobox` and `Alert`" — added in the same PR, so one of them is the fourth. Both then ask the *next* author to write the convention down. Given this PR is the one that crossed the threshold, writing the one-paragraph note into `docs/frontend-architecture.md` now is cheaper than leaving two mutually-inconsistent precedents behind.

---

## Merge readiness

**Ready**, after updating the branch (one clean commit behind, zero file overlap).

Priority order for follow-ups: **I1** (server-side capability default — the #2610 rule, and the one with an operator-visible wrong claim), then **I3** (cross-PR, needs settling with #3268 before the second one merges), then **I2**.

norbert-kulus-blockydevs:
Addressed the tech-lead re-review (0 BLOCKING / 3 IMPORTANT / 3 SUGGESTION) in `0cd3de542`.

**I1 — Invoicing opt-in was browser-only.** `ConnectionService.create` now excludes `Invoicing` from the manifest-derived default whenever the manifest also declares `Fiscalization`, mirroring the existing #1498 `InventoryMaster`/`OfferManager` carve-out. `POST /connections { platformType: 'eparagony' }` with no `enabledCapabilities` no longer grants the invoicing lane a connection has no seller config for. An explicit `enabledCapabilities: ['Fiscalization', 'Invoicing']` still opts in. 3 new unit tests in `connection.service.spec.ts`.

**I2 — the whole-call deadline could still be outlived.** Two paths fixed:
- `createDocument`'s own transport timeout is now derived from what's left of `EPARAGONY_ISSUE_DEADLINE_MS` (floored at 1s so an already-exhausted budget still gets one real attempt), instead of running on its own independent `REQUEST_TIMEOUT_MS × retries` budget.
- `pollToSettledIssuance` now checks `Date.now() >= pollUntil` at the **top** of the loop, before firing a read — previously the very first read always fired regardless of whether `createDocument` had already burned the whole budget.

2 new tests: one asserting the derived `timeoutMs` stays within the deadline, one simulating a `createDocument` that alone consumes the full budget and asserting the poll never calls `client.get`.

**S1 — deadline mirror had no floor.** `check-eparagony-deadline-mirror.mjs` now also rejects a deadline below `MIN_LANE_DEADLINE_MS` (10s) — previously only checked the upper bound against core's ceiling, so a typo'd `100ms` would have passed.

**S2 — ZRICS/ZRE escape hatch had no operator-facing doc.** Added a step to `setup-guide.md` § Step 3 explaining that a genuinely intra-EU/export line needs the vendor code set explicitly, since a plain `0` always invoices as domestic zero-rate.

**S3 — the two "third primitive" docblocks contradicted each other.** `Alert` and `StatusBadge` were each written as if they were independently the third to carry the test-introspection prop. Wrote the convention down in `docs/frontend-architecture.md` § Test-introspection `data-testid` prop with a running list, and pointed both docblocks at it instead of hardcoding a count.

**I3 (cross-PR with #3268)** is addressed on the #3268 side, since that's where the mirror script and field description live — flagged there too so it lands with the same commit.

Full gate re-run after the fixes: `pnpm build` (22/22), `pnpm type-check` (21/21), `pnpm lint` (0 errors), `pnpm check:invariants` (90/90), plus the two affected spec files run directly (112/112 `connection.service.spec.ts`, 35/35 `eparagony-invoicing.adapter.spec.ts`).

PR #3305, opened 2026-09-18
fix(invariants): stop the mirror-guard array parser truncating on a bracket inside a comment

### LINKED ISSUE #3002: [BUG] DX — the mirror guards' array parser truncates on a `]` inside a comment, and an empty parse passes vacuously

norbert-kulus-blockydevs:
## Summary

`check-*-mirror.mjs` scripts extract the string literals of an `export const NAME = [...] as const;` declaration by finding the opening `[` and then locating the closing `]` via `indexOf(']', openBracket)` on the **raw** source, stripping comments only afterward. A `]` written inside a comment between the real brackets (e.g. a docblock mentioning `destinationConnectionIds: []`) was mistaken for the declaration's own close, silently truncating everything after it — possibly the whole array.

None of the affected scripts floored the parsed value count either, so if both sides of a comparison were truncated to empty by the same kind of comment, the differ reported zero issues and the guard printed a success line having compared nothing.

- Fixed by locating the closing bracket on a comment-**blanked** (length- and newline-preserving) copy of the source, so any offset found there is still valid on the original text and reported line numbers never move.
- Every parsed-to-zero-values case is now treated as a broken parser and fails loudly, distinctly worded from an ordinary drift message.
- Audited all 27 `check-*-mirror.mjs` scripts for the pattern.

### Files fixed (18)
Both the bracket-search fix and the zero-value floor, plus new `--self-check` regression cases proven red-first against the pre-fix logic:
`check-allegro-seller-defaults-mirror`, `check-attention-reason-mirror`, `check-authority-kind-mirror`, `check-connection-backlog-status-mirror`, `check-description-format-mirror`, `check-fiscal-reconcile-outcome-mirror`, `check-fiscal-registration-progress-mirror`, `check-hold-reason-mirror`, `check-order-lifecycle-phase-mirror`, `check-parameter-restriction-mirror`, `check-permission-mirror`, `check-resolve-stream-mirror`, `check-return-stage-mirror`, `check-sales-document-reason-mirror`, `check-tax-rate-state-mirror`, `check-automation-merge-field-mirror` (not in the original grep hit — reimplements the bug via a depth-aware `[`/`]` scan on raw content), and `check-sales-document-condition-field-mirror` (zero-value floor only; its parser already blanked comments correctly before this PR).

### Files examined and confirmed NOT vulnerable (left untouched)
- `check-core-capability-mirror`, `check-retry-refusal-reason-mirror`, `check-sourcing-rule-vocabulary-mirror` — already strip/blank comments **before** locating the closing bracket (fixed under earlier issues, e.g. #2403).
- `check-auto-applied-limit-mirror`, `check-price-override-bound-mirror`, `check-system-connection-id-mirror` — parse a single scalar constant, no array literal.
- `check-no-supported-actions-mirror` — an inverse declaration-absence guard, no array literal.
- `check-shipping-tax-split-mirror` — already blanks comments before bracket search.
- `check-sales-document-market-summary-mirror`, `check-stock-and-pricing-preview-mirror` — parse function *bodies* via brace-matching, a different construct from the array-literal `] as const` shape this issue targets.

## Test plan

- [x] Red-first proof: constructed a minimal reproduction of the pre-fix `parseReasonValues`/`declarationBody` logic and confirmed it truncates on a `]` inside a comment (both the bare-`indexOf` shape and the depth-scan shape in `check-automation-merge-field-mirror`).
- [x] Added `--self-check` regression cases per touched script: (a) a `]` inside a `//` comment does not truncate, (b) a `]` inside a `/* */` comment does not truncate, (c) reported line numbers are unchanged by a leading comment containing a bracket, (d) an all-commented-out declaration parses to zero values and is treated as fatal.
- [x] `node scripts/<file>.mjs --self-check && node scripts/<file>.mjs` green for every touched script.
- [x] `pnpm check:invariants` — full green.
- [x] `pnpm lint` — full green (pre-commit hook ran the same, twice).
- [x] `pnpm type-check` — full green.

Closes #3002


piotrswierzy (approved):
## Tech Lead review — #3305

Reviewed per `.claude/commands/pr-review.md` (4-step) and `docs/code-review-guide.md`. Read-only; the branch was staged into a scratch tree symlinked at the real repo so every PR script could be executed against the live source.

### Step 1 — Context

`#3002` names a two-part defect: `parseReasonValues` located the array's closing `]` with `indexOf(']')` on **raw** source and stripped comments only afterwards, and no mirror script floored the parsed value count. One side truncating is loud-but-misleading; **both** sides truncating is silent, and `diffReasonValues` (`scripts/check-sales-document-reason-mirror.mjs:258-270`) is pure set-difference — two empty arrays produce `ok: true`. This is the same family as #2792 and is exactly what `docs/architecture-overview.md` § Data Flow / § 24 treat as a first-order bug: *a guard that passes because it discovered nothing has silently stopped guarding.*

### Step 2 — Verification (demonstrated, not assumed)

**The bug reproduces on `main`, and the fix closes it.** Loading both parsers side by side and feeding one source:

```
input: ['a-x', // e.g. destinationConnectionIds: []  , 'b-y', 'c-z']
MAIN parse : {"line":1,"values":["a-x"]}                     <- truncated at the comment's ]
PR   parse : {"line":1,"values":["a-x","b-y","c-z"]}         <- fixed

input: an array whose only entry is commented out
MAIN empty : {"line":1,"values":[]}                          <- silent; differ then reports "0 differences"
PR   empty : THREW -> "parsed ZERO string literals ... the PARSER is broken"
```

**Length preservation holds**, which is what keeps reported line numbers stable (`blankComments`, e.g. `scripts/check-sales-document-reason-mirror.mjs:88-113`). Exercised across five shapes including a multi-line block comment and an unterminated `/*`: input/output length and newline count identical in every case. The self-check case at `check-sales-document-reason-mirror.mjs` ("a `]` inside a comment BEFORE the declaration does not shift the reported line") asserts the same property behaviourally.

**Blast radius is zero on the current tree, and I checked rather than took the claim.** Every one of the 17 touched scripts was run in both versions against the live source; the compared-value counts are identical:

| script | main | PR |
|---|---|---|
| allegro-seller-defaults / authority-kind / automation-merge-field / connection-backlog / return-stage | 10 / 6 / 6 / 6 / 6 | same |
| fiscal-reconcile / fiscal-registration / hold-reason / order-lifecycle-phase / parameter-restriction | 4 / 8 / 8 / 9 / 8 | same |
| permission / resolve-stream / sales-document-condition-field / sales-document-reason / tax-rate-state | 29 / 15 / 3 / 13 / 3 | same |

So no mirror pair changed meaning: no real source file today carries a `]` inside a comment inside a mirrored array. The fix is latent protection, not a behaviour change — which is the right answer to "could a previously-passing mirror now be falsely mismatched?". All 17 `--self-check` runs and all 17 live runs exit 0.

**The 15 shared `blankComments` copies are byte-identical** (md5 `6ab290ba` across all 15), so no copy-drift entered with the duplication. The two outliers are both correct and both match the PR body: `check-sales-document-condition-field-mirror.mjs` gets the zero-value floor only (its parser already blanked comments), and `check-resolve-stream-mirror.mjs:124` reuses its own pre-existing **quote-aware** `export function blankComments`.

**The "not vulnerable" list checks out**: `scripts/` has no shared helper directory (every `check-*` is a deliberate zero-dependency standalone), and the scripts left untouched are single-scalar parsers, brace-matching function-body parsers, or already comment-blanking.

### Step 3 — Findings

**[SUGGESTION]** — `scripts/check-sales-document-reason-mirror.mjs:88` (and the 14 identical copies)
> `blankComments` is not quote-aware, while a **quote-aware** variant already exists and is exported one directory over at `scripts/check-resolve-stream-mirror.mjs:124` — and this very file already carries a string-aware scanner (`findStringEnd`, `check-sales-document-reason-mirror.mjs:242-254`) used by `parseCopyKeys`. A value containing `/*` would blank past the real `]`. I confirmed the failure mode is **fail-closed**, not fail-open: `parseReasonValues` returns `null`, and `main()` pushes `${FILE}: no 'export const ... = [...]' found` onto `fatal` (`:335-341`). So this is latent and safe, but the PR had a stronger helper available in-tree and shipped the weaker one 15 times. Worth reconciling on the next touch, or noting in the docblock why the quote-unaware form is sufficient here (all mirrored vocabularies are kebab-case codes).

**[SUGGESTION]** — the zero-value guard uses two idioms
> Most scripts `throw new Error('… the PARSER is broken …')` from inside the pure exported parser, while `scripts/check-description-format-mirror.mjs:285-288` routes the same condition into the script's own formatted `fatal` branch. Both exit non-zero, so both are loud; but the throw surfaces as a raw Node stack rather than the script's own message format. Since this PR is itself about making guards fail *legibly*, one idiom would be better. Not blocking — `docs/code-review-guide.md` treats an inconsistent-but-correct failure path as a follow-up.

> The body says "### Files fixed (18)" but lists and changes **17**. Cosmetic, but the file count is the thing a reader uses to check the audit was exhaustive.

### Positive observations

- The zero-value floor is the half `#3002` calls "the serious half", and it is implemented everywhere rather than only where the bracket bug could bite — including `check-sales-document-condition-field-mirror.mjs`, which needed no parser change.
- Every touched script gained `--self-check` regression cases covering both comment forms, line-number stability, and the zero-value fatal. This is the `#2404` discipline (a case that asserts nothing is indistinguishable from a passing one) applied correctly: I ran all 17 and they pass, and the zero-value case asserts the *message*, not merely that something threw.
- `check-automation-merge-field-mirror.mjs:124-133` correctly keeps its depth-aware scan and merely moves it onto the blanked copy, rather than being flattened into the `indexOf` shape — the comment at `:125-127` states why the depth scan is retained.
- The audit section distinguishes "already correct" from "not applicable" per script, which is what makes the 27-script sweep checkable rather than asserted.

---

### Summary

A precise fix for a guard-blindness defect, with the serious half (the vacuous empty-parse pass) addressed as thoroughly as the parser half. I reproduced the pre-fix truncation and the silent empty pass directly against `main`'s exported parser, and confirmed the PR's version fixes both. Blast radius across all 17 scripts is verifiably zero on the current tree — every compared-value count is identical before and after — so nothing that was passing is now falsely mismatched, and nothing that was truncated was hiding a real drift. The only concern is stylistic: a stronger quote-aware `blankComments` already existed in-tree and the weaker form was duplicated 15 times instead.

### Merge Readiness

### Priority fixes

None blocking. Optional follow-ups, in order:
1. Reconcile `blankComments` with the quote-aware variant at `check-resolve-stream-mirror.mjs:124`, or document why quote-unawareness is safe for these vocabularies.
2. Settle on one zero-value failure idiom (throw vs. the script's `fatal` list).
3. Correct "18" → 17 in the PR description.

Issue #3307, opened 2026-09-18
[FEATURE] Frontend — Kind-aware /sales-documents list, replacing invoice-only /invoices


norbert-kulus-blockydevs:
## Problem / Context

`/invoices` (`apps/web/src/app/routes/invoices.route.tsx`, page
`apps/web/src/pages/invoicing/invoices-list-page.tsx`) reads exclusively `useInvoicesQuery` /
`InvoiceRecord` — every row assumes the document is an invoice, and its subtitle hardcodes
"regulatory (KSeF) status," a Polish-invoicing detail on a screen meant to cover every connected
provider. A fiscal-receipt-routed order never appears, and opening one by its record id at
`/invoices/:invoiceId` (`apps/web/src/pages/invoicing/invoice-detail-page.tsx`) dead-ends.

`/orders` already solved "one sale, one of two possible documents" — reuse it, don't reinvent it:
- `apps/web/src/features/orders/components/sales-document-cell.tsx` (the row cell + its popover)
- `apps/web/src/features/orders/lib/sales-document-cell-state.ts` (`resolveSalesDocumentCellState`)
- `apps/web/src/shared/ui/document-kind-glyph.tsx` (`DocumentKindGlyph`)
- `apps/web/src/features/orders/api/orders.types.ts` (`SalesDocumentView` and its per-kind union)

Naming collision to resolve in the same change: `apps/web/src/app/routes/sales-documents.route.tsx`
already registers `settings/sales-documents`, titled "Sales documents"
(`apps/web/src/pages/settings/sales-documents-page.tsx`, the routing-rules config page). Renaming
the operations list to the same words would give the sidebar two identically-worded entries in two
groups about two different things.

## Proposed Solution

Design reference: `docs/plans/mockups/sales-documents.html` (commit this file as part of this issue
if not already present from the backend issue — it documents the exact list layout, filters, popover
reinstatement, and detail-page split this issue implements; every `§NN` reference below points into
it).

**1. Naming (§02).** Retitle the settings page (`sales-documents-page.tsx`) from "Sales documents" to
"Document routing" — route path unchanged, one string. The operations list becomes "Sales documents."

**2. Routing (§03).**
| | Before | After |
|---|---|---|
| Nav label | `Invoices` | `Sales documents` |
| List route | `/invoices` | `/sales-documents` |
[...]

PR #3308, opened 2026-09-18
feat(orders,api): merge invoices and fiscal receipts into GET /sales-documents

### LINKED ISSUE #3306: [FEATURE] CORE — Merged, keyset-paginated sales-document list across invoices and fiscal receipts
## Problem / Context

`/invoices` (`apps/api/src/invoicing/http/invoicing.controller.ts`, `GET /invoices` →
`IInvoiceService.listInvoices`) only ever queries `invoice_records`. A fiscal-receipt-routed order
never gets a row there — it isn't filtered out, it was never queried. Confirmed live, and confirmed
against the code:

- `ISalesDocumentViewService` (`libs/core/src/orders/application/interfaces/sales-document-view.service.interface.ts`,
  impl `sales-document-view.service.ts`) already composes `IInvoiceService` +
  `IFiscalRegistrationService` for the neutral invoice-or-receipt projection — but only as
  `getForOrders(orderIds)` / `getForOrder(orderId)`, batched by a caller-supplied order-id set (the
  shape `/orders` and order-detail need). It is not a flat, paginated, filterable list across every
  order.
- `IFiscalRegistrationService.getById(id)` already exists
  (`libs/core/src/fiscalization/application/services/fiscal-registration.service.interface.ts:252`)
  but is not exposed over HTTP — `apps/api/src/fiscalization/http/fiscalization.controller.ts` only
  has `GET /fiscal-registrations` (requires `orderId`, returns an array) and
  `GET /orders/:orderId/fiscal-registration` (progress poll).
- `IInvoiceService.listInvoices(filter, pagination)` (`invoice.service.interface.ts:227`) paginates
  via plain `{limit, offset}` (`InvoiceRecordPagination`, `invoicing.types.ts:950-953`) — OFFSET, not
  keyset.

There is no widening path to an existing endpoint. A merged operational list is new work.

## Proposed Solution

Design reference (commit before starting):
`docs/plans/mockups/sales-documents.html` — the frontend redesign plan; the merged-list shape and
filters this backend serves are described in its §04 and §06.

**1. Extend `ISalesDocumentViewService`, in `orders` — not in `sales-documents`.**
`orders` is the only context with static module edges to both `invoicing` and `fiscalization`
(`docs/architecture-overview.md`'s cross-context map: `orders --> invoicing`, `orders -->
fiscalization`). `libs/core/src/sales-documents` is a documented **zero-outbound-edge** leaf
(`libs/core/src/__tests__/barrel-purity.spec.ts`'s `ZERO_SIBLING_EDGE_LEAVES`) — it structurally
cannot inject `IInvoiceService` or `IFiscalRegistrationService`, and `sales-document-view.service.interface.ts`'s
own docblock says so explicitly. Add a new method, e.g.:

```ts
listSalesDocuments(
  filters: SalesDocumentListFilters,
  pagination: SalesDocumentKeysetPagination,
): Promise<SalesDocumentListPage>
```

to `ISalesDocumentViewService` and its implementation, following the file's existing
constructor-injection pattern (both dependencies are already there).

**2. Page by keyset, not OFFSET.** Copy the existing row-value keyset pattern already shipped at
`libs/core/src/invoicing/infrastructure/persistence/repositories/invoice-record.repository.ts:379-420`
(`WHERE (date_trunc('milliseconds', record.updatedAt), record.id) > (:cursorUpdatedAt, :cursorId)
ORDER BY ... ASC` — note its documented millisecond-truncation requirement, Postgres `timestamp` is
microsecond-precision and the JS `Date` cursor is millisecond-precision) for **both** sources
(`invoice_records` and `fiscal_registration_records`), then merge-sort the two keyset pages by their
issued/updated instant in application code and cap the combined page. Do not build a SQL `UNION` —
two independently-paginated queries merged in application code is the agreed-cheaper shape, but the
pagination mechanism inside each must be keyset: plain `OFFSET` on either side silently skips or
duplicates rows across a page boundary whenever a new document lands between two fetches, which on a
merged, financial-audit-adjacent list is a correctness bug, not a performance nuance.

**3. Filters**: `kind` (`'invoice' | 'fiscal-receipt' | null` for "routing not set"), `status`
(kind-scoped — do not invent a merged status enum, reuse each kind's existing status/failureMode/
regulatoryStatus fields as the source of the word), `connectionId`, `buyerTaxId` (reuse the existing
three-state encoding, absent / asserted-none / present), an issued-date range, and a free-text search
matching order id, regulatory reference, and buyer tax id.

**4. Row shape**: `orderId`, `kind`, the raw status-source fields (never a pre-flattened word — that
stays a frontend concern, `resolveSalesDocumentCellState`), native-currency amount, `connectionId`,
`issuedAt`, and a duplicate indicator — reuse the existing one-document-per-order invariant read
(`InvoiceRecord.blocksIssuanceElsewhere` / the ADR-041 write-path guard's own check) rather than a new
query.

**5. New HTTP surface**:
- `GET /sales-documents` (new controller, or added to `apps/api/src/invoicing/http/` — implementer's
  call, following the one-controller-per-context convention), query DTO mirroring
  `apps/api/src/invoicing/http/dto/list-invoices-query.dto.ts` plus `kind`, `search`, and a cursor
  param instead of `offset`.
- `GET /fiscal-registrations/:id` on the existing `fiscalization.controller.ts`, backed by the
  already-existing `IFiscalRegistrationService.getById` — a thin route addition, no new service
  method. Needed by the stacked frontend issue's fiscal-receipt detail page.

**6. Aggregate/KPI read** (documents count, amount by kind, failed/unconfirmed count,
duplicate-order count) for the list header — a second lightweight query or folded into the list
response; implementer decides on measured cost.

## Classification

**Type**: CORE
**Layer**: Application (orders context) + Interface (`apps/api`)
**File(s)**:
- `libs/core/src/orders/application/interfaces/sales-document-view.service.interface.ts`
- `libs/core/src/orders/application/services/sales-document-view.service.ts`
- `libs/core/src/invoicing/infrastructure/persistence/repositories/invoice-record.repository.ts` (pattern reference)
- `libs/core/src/fiscalization/application/services/fiscal-registration.service.interface.ts`
- `apps/api/src/fiscalization/http/fiscalization.controller.ts`
- `apps/api/src/invoicing/http/invoicing.controller.ts` (reference; new route may live in a sibling controller)

## Dependencies

None — this is the base of the stack. A frontend issue depends on this one (stacked PR: frontend
branch on top of this one's branch).

## Assumptions

- Amount is the order's own native total (never summed across currencies) — exact sourcing (order
  snapshot vs. a denormalised column) is left to the implementer, following whatever
  `ISalesDocumentViewService`'s existing per-order path already has access to.
- The duplicate-order count/flag reuses the existing invariant check rather than a new aggregate
  query; if that check isn't cheaply reusable in list form, a lightweight dedicated query is
  acceptable — call this out in the PR if taken.
- KPI/aggregate counts (§04 of the mockup: documents / amount / failed / duplicates) are **acceptance
  criteria for this issue** because the frontend list header depends on them; if query cost makes a
  live aggregate impractical at this stage, ship it as a second, separately-cacheable read and say so
  in the PR — do not silently drop the numbers.
- New controller placement (new file vs. extending `invoicing.controller.ts`) is an implementer
  decision; either is acceptable as long as the route is `GET /sales-documents`.

## Acceptance Criteria

- [ ] `GET /sales-documents` returns a merged page of invoice + fiscal-receipt records, filterable by
      `kind` / `status` / `connectionId` / `buyerTaxId` / issued-date range / free-text search.
- [ ] Pagination is keyset-based on both underlying sources; an integration test asserts that paging
      through a growing dataset with a row inserted mid-walk neither skips nor duplicates a row.
- [ ] Each row carries a native-currency amount and a duplicate-order indicator.
- [ ] `GET /fiscal-registrations/:id` returns a single record (404 on unknown id), mirroring
      `GET /invoices/:invoiceId`.
- [ ] An aggregate/summary read exists for the list header (documents count, amount, failed count,
      duplicate-order count) — same endpoint or a sibling one.
- [ ] `pnpm check:invariants` stays green — no new outbound edge from `libs/core/src/sales-documents`.
- [ ] Unit tests for the merge/keyset logic; integration test for both new/extended HTTP routes.

norbert-kulus-blockydevs:
## Summary

Closes #3306.

- New `ISalesDocumentViewService.listSalesDocuments` (orders context) merges two independently keyset-paginated sources — `IInvoiceService.listInvoicesKeyset` and the new `IFiscalRegistrationService.listRegistrationsKeyset` — into one newest-first, cross-order page. Never OFFSET, never a SQL UNION: see `mergeSalesDocumentPages`'s own docblock and unit tests for why (a source whose fetched batch contributed nothing to the truncated page must resume from where it was fetched FROM, not be marked exhausted — the subtle bug an earlier draft of this had, caught by its own test suite).
- New `GET /sales-documents` (`apps/api/src/orders/http/sales-documents.controller.ts`) — filterable by kind/status/connectionId/buyerTaxId/date-range/free-text search, paginated by an opaque base64url cursor.
- New `GET /fiscal-registrations/:id` — `IFiscalRegistrationService.getById` already existed; this is the first HTTP route exposing it, needed by the stacked frontend PR's fiscal-receipt detail page.
- Each row carries a native-currency amount (batched from `order_records`) and a duplicate-order indicator (reusing the exact `groupRankedRecords` grouping `getForOrders` already uses — no second, possibly-drifting duplicate check).

Design reference: `docs/plans/mockups/sales-documents.html` (committed here per the UX Mockups convention).

## Why this is new work, not a widening

Confirmed against the code before writing anything: `GET /invoices` only ever queries `invoice_records`; `GET /fiscal-registrations` requires an `orderId` and returns one order's records; the existing `SalesDocumentView` projection (`getForOrders`/`getForOrder`) is batched by a caller-supplied order-id set, never a flat cross-order list. A merged operational list needed a new read.

## Deferred, deliberately (not silently dropped)

- **Aggregate/summary KPI read** (documents count, amount by kind, failed count, duplicate-order count) — the frontend's KPI strip will read `null`/loading until this lands; filed as a fast follow so this PR doesn't grow an under-tested aggregate query.
- **`kind=` filtering for "routing not set" orders** — structurally impossible for a record-driven list (such an order has no row in either source table at all); `getForOrders` already answers that question per order.

## Testing

- Unit: `mergeSalesDocumentPages` (6 cases, including the resume-cursor edge case), `SalesDocumentViewService.listSalesDocuments` (6 cases), the cursor codec, the new controller.
- Integration (Testcontainers, real Postgres): `InvoiceRecordRepository.findManyKeyset` — proves a row inserted between two page fetches is neither skipped nor duplicated, the correctness property OFFSET pagination lacks under concurrent inserts.
- `pnpm check:invariants` green (cross-context-imports confirms no new outbound edge from `sales-documents`).
- `pnpm lint` / `pnpm type-check` clean on `@openlinker/core` and `@openlinker/api` (pre-existing warnings elsewhere untouched).

## Stack

This is the base of a 2-PR stack — a frontend PR (issue #3307) branches on top of this one's branch, not `main`.


piotrswierzy (approved):
# Tech Lead review — #3308 `feat(orders,api): merge invoices and fiscal receipts into GET /sales-documents`

Reviewed against `docs/architecture-overview.md` §§14/16/17, [ADR-065](docs/architecture/adrs/065-sales-document-read-surface.md), [ADR-041](docs/architecture/adrs/041-sales-document-routing-policy.md), `docs/engineering-standards.md`, `docs/testing-guide.md`.

---

## The two headline claims, verified

**1. The keyset merge cannot drop or duplicate a document. Confirmed — the PR body's claim holds.**

I traced this rather than trusting the description. `mergeSalesDocumentPages` (`libs/core/src/orders/domain/sales-document-page-merge.ts:69-97`) sorts the union of both fetched batches DESC and truncates to `limit`. Because each source's batch is already sorted DESC, the truncated top-k necessarily takes a **prefix** of each source's batch — so `lastConsumed` (line 85, the last `items` entry belonging to that source) is exactly the end of that prefix, and the next query's strict `(createdAt, id) < cursor` returns precisely the unconsumed suffix. No gap, no overlap. The zero-contribution branch (line 95) correctly returns the *incoming* cursor rather than `null`, so a source whose rows were all older than the cutoff is re-fetched from where it was fetched from instead of being silently retired.

The tests genuinely exercise it and would go red: `sales-document-page-merge.spec.ts:72-90` drives limit=1 with source `a` entirely older than `b` and asserts `nextCursor.a` is `undefined`, and `:92-104` asserts an in-progress cursor is preserved by identity (`toBe`). Coercing to `null` fails both. The int-spec (`apps/api/test/integration/invoicing/invoice-record-keyset-list.int-spec.ts:98-126`) proves the mid-walk-insert property against real Postgres. This is good work.

**2. The per-kind status axes survived the merge. Confirmed.**

`SalesDocumentListItem.document` is typed `SalesDocumentRecordView` (`libs/core/src/orders/domain/types/sales-document-list.types.ts:76`) — the existing closed union `SalesDocumentInvoiceView | SalesDocumentReceiptView` (`libs/core/src/sales-documents/domain/types/sales-document-view.types.ts:182`), which neither PR touches. No flattened cross-kind enum was introduced anywhere in the diff. ADR-065's rejected alternative stayed rejected, which is the single easiest thing to get wrong in a *merged* list.

`listRegistrationsKeyset` is correctly placed on `IFiscalRegistrationService` (`fiscal-registration.service.interface.ts:183`), not a repository port leaking across contexts, and the `orders → fiscalization` edge is the one §17/#2516 already sanctions — `orders.module.ts` adds no new module import, so no cycle. Route authorization is explicit and correct: `@Roles('admin','operator','viewer')` (`sales-documents.controller.ts:47`), matching `GET /invoices` (`invoicing.controller.ts:1220`) and `GET /fiscal-registrations` (`fiscalization.controller.ts:185`) — appropriate for a row carrying an amount but no buyer PII.

---

## Findings

### IMPORTANT — `SourcePage.ownNextCursor` is declared, documented, supplied, and never read
`libs/core/src/orders/domain/sales-document-page-merge.ts:34` / `:76-97`

The field carries a stated semantic — *"`null` = this source is exhausted (fewer than the requested limit came back)"* — and `SalesDocumentViewService` faithfully supplies it (`sales-document-view.service.ts:172`, `:181`). The merge body never references it. The **only** path to `null` is `items.length === 0` (line 81).

The consequence is not cosmetic and not an edge case. A source that returns a short final page has all of it consumed, so its cursor becomes a row rather than `null`; the controller's `exhausted` test (`sales-documents.controller.ts:82`) therefore stays false, and **every full walk of this list ends with one extra request that returns zero rows**. That is universal, not rare. On the stacked frontend it surfaces as a live "Next" button that yields an empty page — and that page renders a false claim (see #3309 BLOCKING 2).

Not one of the six unit tests asserts anything about `ownNextCursor`; tests 1, 3 and 4 pass `null` for a source whose rows *are* consumed and the merge returns a row cursor regardless. So the field is a contract member no test can fail on — the shape this repo treats as a defect in its own right.

Fix: `null` only when `ownNextCursor === null` **and** the whole fetched batch was consumed. Naively trusting `ownNextCursor` alone would reinstate the very bug the file exists to prevent (short batch, partially consumed), so the conjunction is load-bearing — and worth its own test.

### IMPORTANT — a tampered cursor answers 500, not the documented 400
`apps/api/src/orders/http/sales-document-list-cursor.codec.ts:49-65`

`revive` validates `typeof id === 'string'` and nothing more. Both `id` columns are `@PrimaryGeneratedColumn('uuid')` (`invoice-record.orm-entity.ts:76`, `fiscal-registration-record.orm-entity.ts:48`), and the value is interpolated as a bound parameter into `(…, inv.id) < (:cursorCreatedAt, :cursorId)`. A client-supplied cursor carrying `id: "x"` reaches Postgres as `invalid input syntax for type uuid` → `QueryFailedError` → 500.

This contradicts the codec's own docblock (*"An unreadable cursor is a caller error, not a server one"*), the controller's advertised `@ApiResponse({ status: 400, description: 'Malformed cursor' })` (`sales-documents.controller.ts:59`), and `CLAUDE.md § Security baselines` ("Validate all user input at system boundaries"). The cursor is entirely client-controlled base64. Add a UUID-shape check alongside the existing `Number.isNaN` guard.

### IMPORTANT — one date control, two different questions; and it silently drops every un-issued invoice
`libs/core/src/orders/application/services/sales-document-view.service.ts:656-676`

`toInvoiceListFilter` maps `issuedFrom`/`issuedTo` to `issuedAt`; `toFiscalListFilter` maps the same inputs to `createdFrom`/`createdTo`, i.e. `createdAt`. The merged list is *sorted* by `createdAt` on both sides, so one operator-facing range control silently asks two different questions of the two halves.

Worse: `issuedAt` is nullable (`invoice-record.orm-entity.ts:122`), and `inv.issuedAt >= :issuedFrom` excludes NULL. So **applying any date range silently removes every `pending` and `failed` invoice** — exactly the rows an operator opens this list to find — while the fiscal side keeps its equivalents, because `createdAt` is always populated. Nothing in the response says a subset was withheld.

Either filter both sides on `createdAt` (consistent with the sort key and with what the merge actually pages on), or state the asymmetry in the DTO description and handle the NULL arm explicitly.

### IMPORTANT — `taxId` is dropped on the fiscal side, so half the list ignores the filter
`libs/core/src/orders/application/services/sales-document-view.service.ts:667-676`

`toFiscalListFilter` omits `taxId` entirely. `taxId=with` therefore returns tax-id-bearing invoices **plus every fiscal receipt unfiltered**; `taxId=without` returns the same fiscal rows again. The DTO calls it "invoice side only" (`list-sales-documents-query.dto.ts:58`), but "only applies to one source" and "the other source ignores it" are different behaviours, and the second is the one shipped.

A fiscal registration carries no buyer tax id, so the honest mapping is: `taxId` set ⇒ `with` excludes the fiscal source, `without` includes it. As written, the operator's filter is a no-op on half the result set with no signal.

### IMPORTANT — `otherRecordCount` discards the flag that tells a real duplicate from a harmless one
`libs/core/src/orders/application/services/sales-document-view.service.ts:223-225`

The count is `siblings.filter(s => s.connectionId !== ranked.connectionId).length` — every sibling on another connection, with no `blocksFurtherIssuance` filter and no `recordId`, even though `groupRankedRecords` already carries both on `RankedRecord` (`:95-102`).

`SalesDocumentOtherRecord`'s own docblock (`sales-document-view.types.ts:184-211`) states the rule this violates twice over: a second record *"is either a pre-existing duplicate an operator must see, or a terminal `rejected` attempt that legitimately left the sale free. The two are told apart by `blocksFurtherIssuance`"* — and *"a panel that reports a second document and then offers no way to open it leaves an operator with a fact and no remedy, on exactly the state that most needs one."* A bare integer reports a rejected attempt as a duplicate and gives the operator nothing to click. Carry `blocksFurtherIssuance` (or filter on it) and the sibling's `recordId`.

### SUGGESTION — `localeCompare` is the wrong comparator for a `uuid` tiebreak
`libs/core/src/orders/domain/sales-document-page-merge.ts:72`

The keyset predicate compares `id` with Postgres `uuid` ordering (16 binary bytes); the merge tiebreaks with `String.prototype.localeCompare`, which is locale- and ICU-version-dependent. For canonical lowercase UUIDs the two agree today, and the tiebreak only fires on a same-millisecond collision *within one source*, so this is theoretical — but a plain relational `<`/`>` (or `localeCompare(x, 'en', …)` pinned) costs nothing and matches the DB exactly.

### SUGGESTION — the Testcontainers proof covers one of the two sources
`apps/api/test/integration/invoicing/invoice-record-keyset-list.int-spec.ts:11-14`

The file states the scope cut honestly ("the fiscal-registration side shares the byte-identical keyset SQL shape"), and the shapes really are identical today. Worth a second, cheap int-spec on `FiscalRegistrationRecordRepository.findManyKeyset` so "identical" stays a tested property rather than a reviewed one — the two are separately maintained files.

### Positive observations
- The `applyListFilters` extraction (`invoice-record.repository.ts`, "so the two pagination strategies can never quietly diverge on WHAT they filter, only on HOW they page") is the right call and well argued.
- `findManyKeyset` returns `nextCursor: null` only on a short page, never a guess — and the `date_trunc('milliseconds', …)` trick correctly reuses the `findIssuedNonTerminal` precedent for the JS-`Date`-vs-`timestamptz` precision mismatch.
- Reusing `groupRankedRecords` for the duplicate signal rather than writing a second predicate is exactly right, and the docblock says why.
- `SalesDocumentListAmount` carrying `currency` beside `value` "precisely so a caller cannot add two rows' amounts without checking they agree first" is the §18/ADR-040 discipline applied without being asked.
- The `kind`/`status` cross-vocabulary behaviour is documented at the type, the DTO and the controller rather than left to be discovered.

---

## Step 4 — Final Assessment

### Summary
This adds a genuinely new cross-order read: two independently keyset-paginated sources merged newest-first behind an opaque composite cursor, plus `GET /sales-documents` and `GET /fiscal-registrations/:id`. The hard part is right — the merge provably cannot skip or duplicate a row, the resume-cursor edge case the body calls out is real and is actually tested, and ADR-065's per-kind status axes survive the merge untouched rather than being flattened into the single enum a merged list tempts you toward. The most important concern is not pagination correctness but **filter honesty**: a date range silently removes every un-issued invoice from the result while keeping its fiscal equivalents, and `taxId` is a no-op on half the list — both are a financial surface quietly answering a different question than the one asked.

### Merge Readiness
🔄 **Approve with changes** — no BLOCKING issues. Five IMPORTANT items to address as follow-ups before this surface is relied on operationally.

Calibration note: I considered the two filter findings for BLOCKING. They are not pagination defects and cannot corrupt or lose a stored record, so they stay IMPORTANT — but they are the two an operator will meet first, hence their position below.

### Priority fixes
1. **Date filter** (`sales-document-view.service.ts:656-676`) — one control must not mean `issuedAt` on one source and `createdAt` on the other, and must not silently drop NULL-`issuedAt` (pending/failed) invoices.
2. **`taxId` on the fiscal side** (`:667-676`) — apply it or exclude the source; do not ignore it.
3. **`ownNextCursor`** (`sales-document-page-merge.ts:34`, `:76-97`) — read it (conjoined with full consumption) so a walk terminates on its last real page, and assert it in a test.
4. **Cursor `id` validation** (`sales-document-list-cursor.codec.ts:49-65`) — UUID shape check, so a tampered cursor answers the documented 400 instead of a 500.
5. **`otherRecordCount`** (`sales-document-view.service.ts:223-225`) — carry `blocksFurtherIssuance` and the sibling `recordId`, per `SalesDocumentOtherRecord`'s own docblock.

Issue #1355, opened 2026-07-06
[FEATURE] Integration/Frontend — Bulk-issue invoices from the list (mirror bulk retry)


norbert-kulus-blockydevs:
Part of the Infakt accounting integration epic #1279.

## Problem / Context

Invoices can only be issued **one order at a time**. The only list-level batch action today is bulk *retry* (`POST invoices/retry`, `apps/api/src/invoicing/http/invoicing.controller.ts:348`, with FE `BulkActionBar` + confirm + result banner on `apps/web/src/pages/invoicing/invoices-list-page.tsx`). High-volume operators expect to select many orders and issue invoices in one action - this is table-stakes in comparable tools (e.g. BaseLinker).

## Proposed Solution

Add bulk issue by mirroring the existing bulk-retry shape (no parallel pipeline - fan out over the single-order issue primitive).

**Interface** (`apps/api/src/invoicing/http/invoicing.controller.ts`)
- New endpoint `POST invoices/bulk-issue` accepting `{ connectionId, orderIds[] }` (cap the batch size, mirror the max-100 guard on `retryInvoices`). Reuse the same per-id issue path `POST invoices` already composes (`issueOne`-style), returning a per-id result summary DTO (mirror `RetryInvoicesResponseDto`).
- Deterministic idempotency per (connection, order) so a re-submitted batch does not double-issue (the auto-issue trigger already uses `invoice:{connId}:{orderId}`).

**Frontend**
- On the orders list (and/or invoices list) add multi-select + a "Issue invoices" bulk action reusing `BulkActionBar`, confirm dialog, and result banner already built for retry.
- New `use-bulk-issue-invoices-mutation` hook; invalidate the relevant queries on success.

## Classification

**Type**: Integration (API) + Frontend
**Layer**: Interface / Application / Frontend
**File(s)**: `apps/api/src/invoicing/http/invoicing.controller.ts`, `apps/api/src/invoicing/http/dto/`, `apps/web/src/features/invoicing/`, `apps/web/src/pages/invoicing/invoices-list-page.tsx`, `apps/web/src/pages/orders/`

## Dependencies

- Reuses the shipped single-order issue path and the bulk-retry FE pattern.

## Assumptions

- Bulk issue fans out to the same single-order issue primitive per id (consistent with the documented "no parallel bulk pipeline" stance for retry). Partial success is reported per id, not all-or-nothing.
[...]

PR #3309, opened 2026-09-18
feat(sales-documents): merge invoices and fiscal receipts into one operator list

### LINKED ISSUE #3307: [FEATURE] Frontend — Kind-aware /sales-documents list, replacing invoice-only /invoices

norbert-kulus-blockydevs:
## Summary

Stacked on #3308 (backend: `GET /sales-documents` + `GET /fiscal-registrations/:id`).

Adds `/sales-documents` — one merged, kind-aware operator view over every invoice and fiscal receipt across connections — alongside the existing `/invoices` list:

- New `features/sales-documents` list surface (api/hooks/components) consuming the keyset-paginated `GET /sales-documents`. No `total`/offset pagination — the backend can't cheaply supply either without risking a skipped or duplicated row mid-walk, so this is a Previous/Next cursor stack (never a "Load more" accumulator).
- `SalesDocumentListCell` reinstates the `/orders` row's identity popover (kind glyph, word/tone, Number/Provider/Clearance facts, duplicate-record warning) — added back after design feedback that the mockup's bare status word felt hollow with nothing to click.
- New minimal `FiscalReceiptDetailPage` at `/sales-documents/fiscal-receipt/:id`; the existing `InvoiceDetailPage` is reused verbatim at its current `/invoices/:id` route rather than duplicated under a second param name.
- `/invoices` (list) **stays mounted** — `InvoicesListPage` carries the only batch-retry and bulk-issue (#1355) flows in the product, and bulk issue is the documented primary remediation path for a `salesDocumentBlocked` order. The merged list links out to it ("Manage invoices") rather than dropping the capability. Nav label, the settings routing-config page/tile (renamed **"Document routing"**, since "Sales documents" now names the new operator list) and both settings-page copy sites updated to match.

## Deliberately deferred (stated scope cut, matching the backend PR)

- KPI strip
- Bulk actions on the *merged* list — batch retry / bulk issue remain reachable only via `/invoices` (see above), not yet ported onto `/sales-documents` itself
- Aggregate summary read

## Test plan

- [x] `pnpm lint` (eslint) on every touched/new file — 0 errors
- [x] `vitest run` — full FE suite: 557 test files / 5971 tests pass, including the new `sales-documents-list-page.test.tsx` (11 tests: loading/error/empty-state branches, cursor stack, debounced search, `/invoices` link) and the untouched `invoices-list-page.test.tsx` (40 tests)
- [ ] Manual click-through on a live stack (not done in this session — flagging honestly)

Closes #3307


piotrswierzy (changes requested):
# Tech Lead review — #3309 `feat(sales-documents): merge invoices and fiscal receipts into one operator list`

Reviewed against `docs/frontend-architecture.md`, `docs/frontend-ui-style-guide.md`, `docs/testing-guide.md`, `docs/code-review-guide.md`, `docs/architecture-overview.md` §14 Invoicing, [ADR-065](docs/architecture/adrs/065-sales-document-read-surface.md), [ADR-041](docs/architecture/adrs/041-sales-document-routing-policy.md).

Scope reviewed is #3309's diff against its own base (`3306-sales-documents-merged-list`) — frontend only. The backend is reviewed separately on #3308.

---

## Findings

### BLOCKING — retiring `/invoices` deletes batch retry and bulk issue from the product, with no replacement
`apps/web/src/app/routes/invoices.route.tsx:26`

The route index is now `<Navigate to="/sales-documents" replace />`, so `InvoicesListPage` is unreachable. That page is the **only** surface in the app carrying two shipped capabilities:

- **Batch retry** of failed invoices — `useRetryInvoicesMutation`, checkbox selection, `BulkActionBar`, `ConfirmDialog` and a result banner (`apps/web/src/pages/invoicing/invoices-list-page.tsx:42,55-56,243-247,279`).
- **Bulk issue (#1355)** — `useBulkIssueInvoicesMutation` (`:249-258`).

The new `/sales-documents` page has no selection, no `BulkActionBar`, and neither mutation. So both actions leave the product in this PR.

The second one is the serious half. `docs/architecture-overview.md` §14 Invoicing states of `POST /invoices/bulk-issue` that blocked orders are *"**not** excluded from bulk issuance … excluding them would break the primary remediation path for the state this surfacing exists to reveal."* #2100 built the `salesDocumentBlocked` badge, the aggregate count and the `?invoicing=blocked` filter specifically so an operator could find those orders and remediate them **here**. This PR removes the remedy and keeps the alarm.

The PR body says bulk actions *"stay kind-specific on the existing per-kind flows"* — but the per-kind flow **was this page**. That statement is not accurate as shipped.

Either keep `/invoices` mounted until `/sales-documents` grows an equivalent, or port selection + the two mutations onto the new list (kind-gated to invoice rows), or land the replacement in the same change. A silent capability removal is not something the redirect comment discloses either.

### BLOCKING — the terminal page tells the operator they have no documents
`apps/web/src/pages/sales-documents/sales-documents-list-page.tsx:254-263`

The render chain is `isLoading → error → items.length === 0 → table`. The empty branch is **not** scoped to the first page, so an empty *non-first* page renders the full `EmptyState` with the message **"No invoices or fiscal receipts have been issued yet."** — a positive claim about the operator's data, made on a screen where they were just reading their documents.

This is not hypothetical and not rare. Because #3308's merge never consults `ownNextCursor` (see that review, IMPORTANT 1), a walk's last *real* page always carries a non-null cursor, so "Next" stays enabled and one further click **always** returns `items: []`. Every operator who pages to the end of this list reaches this screen. Fixing #3308 removes the guaranteed path, but an empty non-first page remains reachable on its own (a document deleted mid-walk), so the guard belongs here regardless: when `cursorStack.length > 0`, render "No more documents" with a way back, never the never-issued copy.

ADR-065's whole subject is a surface that must not state something it cannot support; `docs/architecture-overview.md` §22 records the same rule for returns ("an empty array is never a positive claim … absence and failure are indistinguishable"). This is that rule, inverted.

### IMPORTANT — ~990 lines, a new feature folder and brand-new pagination machinery, with zero new tests
`apps/web/src/features/sales-documents/**`, `apps/web/src/pages/sales-documents/**`

`gh pr diff 3309 --name-only` returns no `*.test.tsx`. The PR body's "24/24 pass" refers to the *existing* `sales-document-cell-state` / `sales-document-cell` suites covering a behaviour-preserving refactor — that is a regression check on old code, not coverage of the new code.

Untested and non-trivial: the cursor stack (`:82-110`) including the reset-on-filter-change invariant and `goPrev`'s pop; the three-way render branch and its empty-state copy (`:246-263`); the `isSalesDocumentKind` / `isTaxIdFilter` URL-param coercion (`:48-54`); the `FiscalReceiptDetailPage` states; the `buildQuery` param assembly. `docs/testing-guide.md` and the review checklist both call for a unit test on non-trivial logic, and `invoices-list-page.test.tsx` is the precedent for exactly this page shape. Both BLOCKING items above are the kind a page test catches.

### IMPORTANT — 67 KB of orphaned component + a test suite that can no longer fail
`apps/web/src/pages/invoicing/invoices-list-page.tsx` (31 KB) and `invoices-list-page.test.tsx` (36 KB)

Neither is deleted, and nothing but the test now imports the component. The suite keeps running and keeps passing against a page no route mounts — a green signal protecting nothing, which this repo treats as a defect in its own right (`docs/architecture-overview.md` passim: "a check that cannot fail"). Whichever way BLOCKING 1 is resolved, these two files need to be deleted or re-pointed in the same change; leaving them is how the next reader concludes the old list is still live.

### IMPORTANT — the search input fires a request and a history entry on every keystroke
`apps/web/src/pages/sales-documents/sales-documents-list-page.tsx:224-230`

`onChange={(e) => setFilter('search', e.target.value)}` calls `setSearchParams` (a router navigation) and resets the cursor stack per character. Typing a 15-character order id issues ~15 requests and pushes ~15 history entries, so the browser Back button needs one press per character before it leaves the page.

Debouncing is an established pattern here — ten-plus call sites including `use-category-search.ts`, `bulk-category-choose-modal.tsx`, and `use-sales-document-rule-overlap-query.ts` **in this same feature folder**. Note this is new surface area, not a carried-over wart: the page it replaces has no free-text search at all (`invoices-list-page.tsx` filters are selects and dates only). Debounce the value and prefer `replace: true` on the search param write.

### SUGGESTION — every currency is formatted at two decimal places
`apps/web/src/pages/sales-documents/sales-documents-list-page.tsx:112-119`

`item.amount.value.toFixed(2)` renders a JPY total as `1000.00 JPY` and a KWD total short by a digit. `docs/architecture-overview.md` §17 records `minorUnitExponentFor` existing precisely because "parts that sum exactly … is false at two decimals for JPY and for KWD". `Intl.NumberFormat(locale, { style: 'currency', currency })` gets this right for free and also groups thousands.

### SUGGESTION — the date filter treats an operator's local date as UTC
`apps/web/src/pages/sales-documents/sales-documents-list-page.tsx:74-75`

`` `${issuedFrom}T00:00:00.000Z` `` means an operator in CET who picks the 15th excludes documents issued between 00:00 and 01:00 local on the 15th. This is copied verbatim from `invoices-list-page.tsx:112-113`, so it is a carried-over wart rather than a regression — worth fixing while the boundary is being rebuilt.

### SUGGESTION — the docblock and the PR describe a control that is not there
`apps/web/src/pages/sales-documents/sales-documents-list-page.tsx:7-8`, `:273-283`

Both the file header and the PR body call this a "Load more" cursor stack; the UI renders **Previous / Next**. Previous/Next is the better choice for a cursor stack (Load-more implies accumulation, which this does not do) — the prose should follow the code.

### SUGGESTION — `liveRegion="off"` silences the empty result for assistive tech
`apps/web/src/pages/sales-documents/sales-documents-list-page.tsx:256`

A screen-reader user who narrows the filters into an empty result gets no announcement that the table went away. If the flag exists to stop double-announcing on first load, consider gating it on `hasFilters` instead.

### Positive observations
- **The per-kind status axes are preserved.** `SalesDocumentListItem.document` reuses `SalesDocumentRecordView` from `features/orders` rather than a second copy of the union (`sales-document-list.types.ts:9-16`), so the invoice's two axes and the receipt's single axis stay exactly as ADR-065 requires. In a *merged* list this is the easiest thing to get wrong, and it was got right deliberately and for the stated reason.
- **No total is claimed.** `Showing {items.length}` (`:274`) reports only what is on screen — the honest rendering for a keyset list, and the one place a "Load more" UI usually invents a count.
- **The reinstated popover uses the shared accessible primitive**, not a hand-rolled one: `Popover` / `PopoverTrigger asChild` over a real `<button type="button">` with a composed `aria-label` (`sales-document-list-cell.tsx:140-149`), so keyboard operability, focus return and dismissal come from the audited component.
- **No feature cycle.** `features/sales-documents → features/orders` is one-way; `features/orders/index.ts` references the other feature only in comments. Dependency direction (`app → pages → features → shared`) holds, no raw `fetch()` escapes the api-client layer, server state is TanStack Query and URL state is search params.
- `FiscalReceiptDetailPage` distinguishes loading / error / not-found as three renders (`fiscal-receipt-detail-page.tsx:36-61`), and the invoice detail page is reused verbatim at its existing route rather than duplicated under a second param name.

---

## Step 4 — Final Assessment

### Summary
This replaces the invoice-only list with a merged, kind-aware `/sales-documents` surface, and the modelling is right: the per-kind status union is reused rather than flattened, the cursor list claims no total it cannot supply, the popover is built on the audited primitive, and there is no feature cycle. What blocks it is what the replacement quietly takes away. Retiring `/invoices` removes batch retry and bulk issue — and bulk issue is the documented primary remediation path for the `salesDocumentBlocked` state #2100 exists to surface — while the PR body states those actions stay on flows that were, in fact, this page. Compounding that, the terminal empty page tells an operator "No invoices or fiscal receipts have been issued yet", a false claim about their own data that today is reachable on every full walk. Neither would have survived a page test, and the PR ships none.

### Merge Readiness
❌ **Request changes**

### Priority fixes
1. **Do not drop bulk issue and batch retry** (`app/routes/invoices.route.tsx:26`). Keep `/invoices` mounted, or port selection + both mutations onto the new list, in this change. Correct the PR body either way.
2. **Never render the never-issued empty state on a non-first page** (`sales-documents-list-page.tsx:254-263`). Branch on `cursorStack.length > 0`. (#3308's `ownNextCursor` fix narrows the path; it does not close it.)
3. **Add page tests** covering the cursor stack, the filter-change reset, and the three render branches — the two items above are precisely what such a test catches.
4. **Delete or re-point the orphaned `invoices-list-page.tsx` + `.test.tsx`**, so no suite is left passing against an unmounted page.
5. **Debounce the search input** and write its param with `replace: true` (`:224-230`).

norbert-kulus-blockydevs:
## Addressed review findings

All five priority fixes from the review are in `5bfb3b104` (pushed).

1. **BLOCKING — dropped bulk issue / batch retry.** `/invoices` (the list) stays mounted (`InvoicesListPage` un-redirected); it's no longer replaced by `<Navigate>`. The merged `/sales-documents` list now carries a "Manage invoices" action linking to `/invoices` for those two mutations. PR description below is corrected to match.

2. **BLOCKING — terminal page claims no documents exist.** `items.length === 0` now branches on `cursorStack.length > 0` first: a non-first empty page renders "No more documents" + a Back action, never the "no invoices or fiscal receipts issued yet" copy. That copy is now reachable only on a true virgin first page.

3. **IMPORTANT — zero tests on ~990 lines of new machinery.** Added `sales-documents-list-page.test.tsx` (11 tests): loading/error, all three empty-state branches (virgin, filtered, non-first-page), row rendering + currency formatting, cursor-stack Next/Prev + filter-change reset, debounced search (asserts no request fires per keystroke), and the `/invoices` link.

4. **IMPORTANT — orphaned `invoices-list-page.tsx` + its test.** Resolved by fix 1 — the route mounts it again, so the existing test suite (40 tests, unchanged) is back to protecting a live page.

5. **IMPORTANT — search fires a request + history entry per keystroke.** Local `searchInput` state is now debounced (300ms, `useDebouncedValue` — the same hook `products-list-page.tsx` / `customers-list-page.tsx` use) before it reaches the query or the URL; the URL sync uses `replace: true`.

Also fixed while touching the same lines:

- **SUGGESTION — `.toFixed(2)` breaks on JPY/KWD.** Switched to the existing `formatAmount()` helper (`Intl.NumberFormat` with `style: 'currency'`), the same one `orders-list-page.tsx` / `products-list-page.tsx` already use.
- **SUGGESTION — `liveRegion="off"` silences filtered/paginated empty states for assistive tech.** Now `off` only on the true virgin landing; every other empty-state arm (filtered, non-first-page) keeps the default `polite`, matching the existing convention in `orders-list-page.tsx`.
- **SUGGESTION — docblock says "Load more", UI says Previous/Next.** Docblock corrected to describe the actual Previous/Next cursor stack.

**Left as-is, deliberately:** the date-range UTC widening (`T00:00:00.000Z`/`T23:59:59.999Z`). It's copied verbatim from `orders-list-page.tsx`'s identical, already-shipped pattern — fixing it here would be a repo-wide convention change riding inside an unrelated PR, not a regression this PR introduced.

Also rebased onto current `main` (through #3308) after the backend PR merged and picked up an unrelated `jest.config.mjs` worker-memory fix that had landed locally but not been pushed.

Re-ran the full FE suite after all fixes: `pnpm --filter @openlinker/web type-check` clean, 557 test files / 5971 tests passing, `pnpm lint` clean on touched files.

piotrswierzy (changes requested):
## Tech Lead review — `feat(sales-documents): merge invoices and fiscal receipts into one operator list`

The frontend slice is genuinely good work and I'd approve it as-is. What blocks this is the ten `jest.config.mjs` files riding along, which are unrelated to the stated scope and reinstate a CI failure mode the repo has documented three times.

---

### 🔴 BLOCKING 1 — spreading `ciStabilityConfig` into 10 more packages raises each to **8 CI workers**

`libs/integrations/{ai,dpd-polska,eparagony,erli,fx,infakt,inpost,ksef,subiekt,woocommerce}/jest.config.mjs`

`ciStabilityConfig` is not a neutral "cap" bag. Its `maxWorkers` is `resolveUnitTestWorkers(PRESTASHOP_ALLEGRO_CI_WORKERS)` with `PRESTASHOP_ALLEGRO_CI_WORKERS = 8`, and `jest.ci-stability.mjs`'s own docblock names it:

> *"The one measured departure from the shared unit-tier count (#3271). The shared default is 2, because this job runs four packages at once and the container allows about 8 CPUs… **These two packages** measured better than that on the real runner"*

On `main`, exactly two packages import it — prestashop and allegro. This PR makes it twelve, so on CI each of the ten new ones resolves `maxWorkers: 8` instead of the shared `2`.

That is not theoretical head-room. `test:ci` runs `--workspace-concurrency=4`, and eight of the ten packages have ≥8 spec files, so they will genuinely saturate the cap:

| package | spec files | package | spec files |
|---|---|---|---|
| ksef | 29 | infakt | 12 |
| woocommerce | 25 | inpost | 11 |
| erli | 21 | dpd-polska | 10 |
| subiekt | 15 | fx | 6 |
| eparagony | 14 | ai | 3 |

Four of those at once is `4 × 8 = 32` jest processes, which is exactly the number `docs/testing-guide.md` records as having gone wrong:

> *"`4 packages × 8 workers = 32 jest processes` broke CI twice, and the second attempt took a runner permanently offline. Peak process counts, measured: **~20 passes, ~40 kills a runner.**"*

and

> *"Raising either number again without re-measuring the WHOLE job is how this went wrong three times."*

This run's `Test` job passing green is the expected reading — the failure mode is a nondeterministic OOM-kill under contention, so one green run is not evidence against it. The guide is explicit that a green re-run after a `SIGKILL` "hides the real cause, which is exactly how a genuine regression eventually slips through unnoticed."

The `workerIdleMemoryLimit: '512MB'` that comes with the spread carries its own recorded scar (`docs/lessons.md`, the 512MB episode that cost `libs/core` a 15× slowdown by recycling a worker after almost every file). These packages are smaller so it may well be fine — but it is an unmeasured knob bundled with a measured one, which is the same lesson.

**Fix — either:**

1. **Drop the ten files from this PR** (preferred — see BLOCKING 2), or
2. If the caps are genuinely wanted now, take the shared default rather than the prestashop/allegro departure:

```js
import { createRequire } from 'node:module';
const { resolveUnitTestWorkers } = createRequire(import.meta.url)('../../../jest.unit-workers.cjs');

// …
  maxWorkers: resolveUnitTestWorkers(),   // shared unit-tier 2, not the measured 8
  workerIdleMemoryLimit: '512MB',
```

Option 2 still deserves its own PR and its own whole-job timing, per the file's own instruction.

---

### 🔴 BLOCKING 2 — scope

Ten CI-infrastructure files in a PR titled and described as a frontend sales-document list, with no mention in the description or test plan. The commit-per-concern rule matters most exactly here: if this does cause an OOM in a fortnight, nobody bisecting will look inside "merge invoices and fiscal receipts into one operator list". Please split them out.

---

### 🟢 What's right, and worth saying

- **`resolveSalesDocumentRecordWord` extraction is the right call and the direction of the edge is correct.** `features/orders → features/sales-documents` stays the value import it already was (#2761); the new `features/sales-documents → features/orders` edge is `import type` in all three call sites, so it erases and no runtime barrel cycle is created. Both slugs are registered in the ESLint pattern groups.
- **Keeping `/invoices` mounted.** `docs/architecture-overview.md` §14 is explicit that bulk issue is the primary remediation for a `salesDocumentBlocked` order; retiring the list would have removed the remedy while keeping the alarm. The header comments say exactly that, in place.
- **The two empty states are genuinely distinguished** — "No more documents" (reached by paging off a loaded table, where the keyset walk can legitimately answer an empty last page) vs. the virgin-landing copy. That is the same `absence ≠ none` rule `docs/frontend-architecture.md` § Paginated Totals records, applied to a keyset list. The `liveRegion` split between them is a nice touch.
- **No `total`, and the pager degrades per affordance** — `Previous` gated on `cursorStack.length`, `Next` on `nextCursor`, "Showing N" instead of a fabricated count. Correct for a keyset read.
- `formatAmount` rather than `.toFixed(2)`, so JPY/KWD render right.
- Route file deliberately named apart from the `/settings/sales-documents` config route, with the collision called out in its docblock; `EXPECTED_LAZY_ROUTE_COUNT` bumped with a reason.

---

### 🟡 IMPORTANT — `status` filter is readable from the URL but unsettable in the UI

`apps/web/src/pages/sales-documents/sales-documents-list-page.tsx`

`status` is parsed from search params, threaded into `filters`, and counted in `hasFilters` — but the toolbar renders no control for it. So a `?status=…` link produces a filtered table whose empty state says *"Try clearing some filters"* with nothing on screen to clear. Either add the Select, or drop `status` from `hasFilters` until it has one.

---

### 🟢 SUGGESTIONS

1. **`goNext`'s non-null assertion.** `query.data!.nextCursor as string` — the `if` guard doesn't narrow into the setter callback. Cheaper to read as:
   ```ts
   const next = query.data?.nextCursor;
   if (next) setCursorStack((prev) => [...prev, next]);
   ```
2. **Inline `style={{ marginBottom: 'var(--space-3)' }}`** in `fiscal-receipt-detail-page.tsx` (×3). Token names are right; consider a class or the typed `tokens` catalog (`shared/theme/tokens.ts`) so the values stay discoverable.
3. **The `useEffect` comment says `setSearchParams` is "a stable dispatcher".** In react-router-dom v6 it's a `useCallback` that depends on `searchParams`, so it isn't stable. The code is still correct — React runs the *latest* effect closure when deps change, so no staleness — but the stated reason isn't the real one. Worth rewording, since the next person may rely on the claim somewhere it does matter.
4. **Manual click-through is honestly flagged as not done.** Worth doing before merge given this replaces the nav entry operators use daily.

---

**Verdict:** ❌ Request changes — on the bundled jest configs only. Split those out (or switch them to `resolveUnitTestWorkers()`), address the `status` filter, and this is an approve from me.

norbert-kulus-blockydevs:
## Addressed second-round review findings

Both blocking findings from the second review pass are fixed in `27cd3e5ae` (pushed).

1. **BLOCKING — the 10 unrelated `jest.config.mjs` files.** Reverted `libs/integrations/{ai,dpd-polska,eparagony,erli,fx,infakt,inpost,ksef,subiekt,woocommerce}/jest.config.mjs` back to their pre-`558e63476` state. That commit's `ciStabilityConfig` spread was the measured prestashop/allegro departure (8 workers on a shared default of 2) applied to ten more packages with no re-measurement, reinstating exactly the `4 × 8 = 32`-process CI failure mode `docs/testing-guide.md` records as having broken CI twice — and it was riding inside a frontend sales-document PR with no mention in the description, per the review's own "commit-per-concern" point. Kept the two legitimate fixes that were bundled in the same source commit (`invoice-detail-page.test.tsx`'s "Back to sales documents" copy, `route-lazy.test.ts`'s bumped lazy-route count) — both are genuine follow-ups of the sales-documents merge and untouched by the revert.

2. **IMPORTANT — `status` was readable from the URL but unsettable in the UI.** Added a real `status` `Select` to the toolbar. It's kind-scoped: selecting `kind=invoice` narrows the options to `InvoiceStatusValues`, `kind=fiscal-receipt` to the new `FiscalRegistrationStatusValues` (exported from the fiscalization feature — it existed as a type but not as a values array), and no `kind` selected offers the union of both, matching the backend DTO's own documented behaviour ("with `kind` unset it must accept EITHER vocabulary"). Switching `kind` clears a `status` that can no longer match the new vocabulary, so the filter never silently returns zero rows with no visible cause. Two new tests cover both behaviours.

Also verified:
- `pnpm --filter @openlinker/web exec eslint` on every touched file — 0 errors
- Full FE suite: 559 test files / 6002 tests passing (13 in the touched spec file, up from 11)

norbert-kulus-blockydevs:
## Also addressed the SUGGESTIONS (`2d05f007a`)

- **`goNext`'s non-null assertion** — reads `query.data?.nextCursor` into a local once, no more `query.data!.nextCursor as string` inside the setter.
- **The `useEffect` comment's inaccurate claim** — reworded: `setSearchParams` is a `useCallback` that *does* change identity with `searchParams`, so it isn't a "stable dispatcher"; omitting it from the deps is still correct (React runs the latest closure), and the real reason is that including it would re-run the effect on every URL change and defeat the debounce.
- **Inline `style={{ marginBottom/marginTop: … }}` on `fiscal-receipt-detail-page.tsx`** — replaced all three with named classes (`.fiscal-receipt-detail__status` / `__failure-reason` / `__artefacts`) in `index.css`, reusing the existing `--space-3`/`--space-4` tokens.
- **Manual click-through** — done locally against the docker stack, confirmed working.

`type-check` clean, `eslint` clean on both touched files, `vitest` 13/13 passing on `sales-documents/*`.

piotrswierzy (approved):
## Re-review — `feat(sales-documents): merge invoices and fiscal receipts into one operator list`

Re-reviewed at `2d05f007a`. **Every finding from my previous review is addressed**, and two of them went further than I asked. This supersedes the earlier request for changes.

### ✅ BLOCKING 1 & 2 — the jest configs are gone

Verified against the branch rather than the diff: no package in `libs/integrations/*` spreads `ciStabilityConfig`, and `git diff --name-only origin/main...` shows **no `jest.config.mjs` or `package.json` changes remain**. The PR is back to the FE slice it says it is.

Worth noting for the record: `558e63476` (the commit that introduced them) was reverted rather than split out, so the ten packages remain uncapped on `main`. If the caps are still wanted, they need their own PR using `resolveUnitTestWorkers()` — the shared `2`, not the prestashop/allegro `8` — plus a whole-job timing. Nothing to do here.

### ✅ IMPORTANT — status filter, and it solved a problem I hadn't flagged

I asked for a control or for `status` to be dropped from `hasFilters`. What landed is better: `statusOptionsForKind` scopes the options to the vocabulary `kind` selects, with both offered when no kind is chosen — matching the backend DTO's own *"with `kind` unset it must accept EITHER vocabulary"*.

That scoping introduces a stale-filter hazard, and `setKindFilter` pre-empts it:

```ts
const currentStatus = next.get('status');
if (currentStatus && !validStatuses.includes(currentStatus)) next.delete('status');
```

Without it, picking `kind=invoice` + `status=issued` and then switching to `fiscal-receipt` leaves an unmatchable status silently returning zero rows — the exact "a filter the operator cannot see the effect of" shape. Clearing it inside the same `setSearchParams` update (rather than a follow-up effect) means the URL never passes through an invalid intermediate state.

Both behaviours are tested: one case asserts the invoice vocabulary is offered and `Registered` is **not**, the other asserts the status clears and the select resets to `''` on a kind change. The negative assertion in the first is what makes it a real test of the scoping.

### ✅ The `as const` conversion

`FiscalRegistrationStatus` needed a runtime array for the select, and rather than hardcoding one the type was converted to `FiscalRegistrationStatusValues` + derived union and exported from the barrel. That is the documented house pattern (`engineering-standards.md` § Union Types: `as const` Pattern) and it means the options list cannot drift from the type. Right instinct — the lazy version here is a local `['pending','registering',…]` literal.

### ✅ The three suggestions

1. **`goNext`** — `const next = query.data?.nextCursor; if (next) …`, non-null assertion and cast both gone.
2. **Inline styles** — replaced with `fiscal-receipt-detail__status` / `__failure-reason` / `__artefacts` in `index.css`.
3. **The `setSearchParams` comment** — corrected, and the replacement is better than what I suggested. I only asked for the stability claim to be fixed; the new comment states the *actual* reason the dep is omitted:

   > *including it would re-run this effect once per URL change and defeat the whole point of debouncing*

   That's the load-bearing reason. Mine was just the pedantic half.

### Notes

- CI green on this head (`Test`, `Test (web)`, `Integration Tests`, `Build`, `Type Check`, `Docker`); `Lint` still in progress at review time — worth a glance before merge, though nothing in these commits touches an invariant script.
- `EXPECTED_LAZY_ROUTE_COUNT` is still 68 here and #3261 takes the same constant to 67 from the same base. Whichever merges second conflicts, and the resolution is **69** — flagging again since #3261 is still open.
- Manual click-through is still unticked in the test plan. This replaces the nav entry operators use daily, so it's worth doing, but I won't hold the approval for it.

**Verdict:** ✅ Approve.

Issue #3290, opened 2026-09-16
[FEATURE] CORE — Credit the delivery line in return correction proposals


jakubretajczykBD:
## Problem / Context

When OpenLinker issues an invoice it emits the buyer-paid delivery cost as an **ordinary product-shaped line**: `toShippingLines()` (`libs/core/src/invoicing/application/mappers/order-to-issue-invoice-command.mapper.ts:299`) appends a line named `'Shipping'` (`SHIPPING_LINE_NAME`, line 37) so the invoice gross equals the order total (#1517). That line is persisted verbatim into `issuedLineSnapshot` (`invoice.service.ts:790` — `lines: cmd.lines`, no recomputation), which is exactly what a later correction diffs against (#1297).

The credit-note matcher never touches it.

**It is not "unmatched" — it is invisible.** `classifyReturnCorrectionLines` (`libs/core/src/returns/domain/domain-services/return-correction-matching.domain-service.ts:116`) maps over the **return's** lines and looks each one up in a by-name index of the snapshot. `ReturnCorrectionProposalService` builds `candidateLines` from `record.lines` alone (`return-correction-proposal.service.ts:177`). A snapshot line that no return line points at simply never enters the classification, so it produces neither a proposal nor a `no-match` entry. The operator is not told that delivery was excluded; delivery does not appear in the proposal at all.

**The harm is a document that disagrees with the money.** The refund side already *can* cover delivery — `ReturnRefundService` takes an operator-supplied `input.amount` (`return-refund.service.ts:170`, `:273`), with no line derivation. So today an operator can refund the delivery cost in cash while the correction transmitted to the tax authority credits only the products. That divergence, not the missing feature, is the reason this is worth fixing: a correction cannot be withdrawn once sent.

**Why this is legally live.** Under Directive 2011/83/EU art. 13(2) a consumer withdrawing within the statutory window is reimbursed the cost of the **standard** (cheapest offered) delivery option. Art. 14(1) leaves return shipping with the buyer unless the trader agreed otherwise. So there is a real, recurring case where delivery must be credited and OL structurally cannot express it.

**The `returns` context has no concept of delivery at all** — a grep over `libs/core/src/returns/` for `shipping|delivery` returns one unrelated docblock reference.

[...]

PR #3313, opened 2026-09-21
feat(returns): narrow the correction matcher onto ReturnLine.resolvedOrderLineId

### LINKED ISSUE #3312: [TASK] CORE — narrow return-correction-matching.domain-service.ts onto ReturnLine.resolvedOrderLineId
## Problem / Context

`classifyReturnCorrectionLines` (`libs/core/src/returns/domain/domain-services/return-correction-matching.domain-service.ts`)
matches a disposed return line against the issued invoice's line snapshot **by product name**
(`indexSnapshotByName` / `normalizeCorrectionLineName`). `InvoiceLine` carries no id and no SKU, so
an order that repeats one offer across two lines produces two identically-named candidates the rule
cannot tell apart — it reports `status: 'ambiguous'` with a `candidates` list and asks the operator
to pick, even though the operator (holding a parcel) has no more information than the system does.

#3171/#3172 already solved *half* of the underlying data problem: `ReturnLine.resolvedOrderLineId`
is now resolved at ingestion (`ReturnOrderLineResolverService`, runs on every
`marketplace.return.sync`) and, when resolved, is trustworthy. **Nothing reads it yet** — the
matcher still joins by name, so the field sits unused while the exact ambiguity it was built to
remove keeps occurring.

This is (part of) the prerequisite the new correction-proposal UI design (invoice-line grid,
`docs/plans/mockups/correction-proposal-3087.html`, PR #3123 — not yet merged) depends on: that
design has **no row shape for `ambiguous`/`candidates`** at all. Until this matcher is narrowed, a
`'proposed'` outcome can still carry a line the new UI cannot render, and every frontend sub-issue
of #3087 (#3088–#3095, several already implemented against the old name-matching contract) is
blocked on it in practice, even though the epic's own banner and PR #3123's description still
(incorrectly) cite #3172 alone as the blocker.

## The half this issue does NOT solve — and why that matters here

`resolvedOrderLineId` is only half the join. Confirmed in code:

- `InvoiceLine` (`libs/core/src/invoicing/domain/types/invoicing.types.ts:382`) carries
  `{name, quantity, unitPriceGross, taxRate, unit?}` — **no id field at all**.
- `toInvoiceLine` (`order-to-issue-invoice-command.mapper.ts:264`) receives `item: OrderItem` — which
  **has** the order line's `id`, the exact id `resolvedOrderLineId` points at — and discards it. The
  returned `InvoiceLine` never carries it forward.
- The snapshot (`IssuedLineSnapshot.lines: InvoiceLine[]`) therefore has **no way to answer "which
  position does order-line-id X occupy"** at all. `originalLineNumber` is a 1-based array position
  with nothing behind it but array order.

So resolving a return line via `resolvedOrderLineId` is not a lookup this issue can do against
today's data — it requires the id to exist in the snapshot in the first place. That is a change to
`InvoiceLine` / the issuance mapper, in the `invoicing` context — **the same class of change, on the
same structure, following the same pattern** ("add an optional field at issuance; a snapshot
predating it is refused with a named reason, never guessed at by name") that **#3290 is already
making** for its `lineKind?: 'product' | 'shipping'` discriminator.

**This is a blocking design decision, not an implementation detail**, and it has two live options:

1. **Extend #3290's own step 1** (or land immediately after it) to also carry `orderLineId?: string`
   on `InvoiceLine`, set by `toInvoiceLine`/`toShippingLines` from `item.id`. One field addition, one
   refusal-for-old-snapshots decision, shared with #3290 rather than duplicated. **Preferred** — see
   Dependencies.
2. Land it independently, accepting that both this issue and #3290 add an optional field to the same
   struct under the same "refuse old snapshots" pattern, and must be sequenced (not merged
   concurrently) to avoid the two refusal-reason additions colliding.

Whichever option is chosen, a snapshot issued before the field exists carries no `orderLineId` and
**must be refused with a named reason** for lines that would otherwise rely on it, following the
existing `no-line-snapshot` precedent — never silently falling back to a guess.

## The residual `ambiguous` case is NOT eliminated, and the UI cannot render it

Even once the join above lands, `resolvedOrderLineId` can be genuinely `unresolved` **on returns
ingested today, not only historically** — confirmed in
`return-order-line-resolution.domain-service.ts` (lines ~164/169/223): the resolver itself returns
`{status: 'unresolved', reason: 'ambiguous' | 'no-candidate'}` whenever SKU/offerId/price do not
uniquely identify an order line. So narrowing the matcher does not make `ambiguous` disappear — it
makes it **rarer**, with a permanent residual for:

- a `ReturnLine` whose own join to the order line was itself unresolved/ambiguous, and
- any return whose invoice predates the `orderLineId` snapshot field from the previous section.

**The new grid design has no row shape for this residual at all** (mockup's own words: "which this
grid has no row shape for"). This must be decided **before implementation starts**, not left open:

- Does the frontend grid need a fallback row shape for a residual `ambiguous`/`no-match` line (i.e.
  the current line-list rendering survives, narrowed to only these residual cases)? or
- Is residual `ambiguous` disclosed as its own named, non-blocking exclusion (closer to the existing
  `no-match` reasons) rather than as a pick-one list?

This decision is cross-cutting with #3087's frontend sub-issues and PR #3123 and should be recorded
here before either side starts, not discovered mid-implementation.

## Proposed Solution

1. **Resolve the blocking design decision above first** (own `InvoiceLine.orderLineId` field,
   coordinated with #3290 — see Dependencies) — this issue does not proceed without it.
2. Narrow `classifyReturnCorrectionLines` to resolve a return line's invoice line via
   `ReturnLine.resolvedOrderLineId` → `InvoiceLine.orderLineId` when both are present and the match
   is unique, instead of `indexSnapshotByName`.
3. Decide and document the residual-`ambiguous` shape from the section above, and implement it —
   whether that means keeping a narrower `ambiguous`/`candidates` path, or replacing it with a
   different named exclusion.
4. Thread the new field through every caller of `classifyReturnCorrectionLines` /
   `CorrectionReturnLineInput` (`toMatcherInput` in `return-correction-proposal.service.ts` currently
   builds the matcher input from `ReturnLine` without it).
5. Product-line matching for a line with no resolvable `orderLineId` join is **unchanged** —
   byte-identical fallback to today's by-name behaviour, asserted by an existing-behaviour test.

## Classification

**Type**: CORE
**Layer**: Domain, Application; plus a type addition in `invoicing` (coordinated with #3290)
**File(s)**:
- `libs/core/src/returns/domain/domain-services/return-correction-matching.domain-service.ts`
- `libs/core/src/returns/domain/types/return-correction-proposal.types.ts`
- `libs/core/src/returns/application/services/return-correction-proposal.service.ts`
- `libs/core/src/invoicing/domain/types/invoicing.types.ts` (the `orderLineId` addition)

## Dependencies

- Requires #3171/#3172 (merged) — `ReturnLine.resolvedOrderLineId` must already be populated.
- **Coordinate with #3290** (delivery-line crediting): both issues add an optional discriminator
  field to `InvoiceLine` under the same "refuse pre-existing snapshots" pattern. Preferred sequencing:
  land #3290's step 1 (or a combined `InvoiceLine` widening covering both `lineKind` and
  `orderLineId` in one pass) before or together with this issue's snapshot change, to avoid two
  independent refusal-reason additions landing concurrently on the same struct.
- Blocks: every unstarted frontend sub-issue of #3087, and the invoice-line-grid redesign in PR #3123
  (currently unmerged, based on the old per-row-picker premise this issue exists to remove) — but
  only once the residual-`ambiguous` decision above is also resolved.

## Assumptions

- `docs/plans/mockups/correction-proposal-3087.html` (PR #3123) is the eventual design of record,
  but this issue is scoped to the backend matcher + the minimal `invoicing` snapshot change it
  requires — it does not require PR #3123 to merge first.
- The residual-`ambiguous` shape (kept narrower, or replaced) is a decision to be made and recorded
  in this issue's thread before implementation starts, not left to the implementer.

## Acceptance Criteria

- [ ] `InvoiceLine` / the issued-line snapshot carries an `orderLineId` (or equivalent) set at
      issuance, coordinated with #3290's `lineKind` addition rather than landed as an unrelated,
      concurrent change to the same struct
- [ ] A snapshot predating that field is refused with a **named reason** for any line that would
      otherwise depend on it — never silently guessed at by name
- [ ] A disposed return line with a resolved, unique `resolvedOrderLineId` **and** a resolvable
      `InvoiceLine.orderLineId` is matched deterministically — no name-based lookup involved
- [ ] The residual-`ambiguous` shape (unresolved join, or pre-field snapshot) is explicitly decided
      and implemented — not left as an unspecified fallback
- [ ] The by-name fallback path is exercised and tested explicitly, not merely inherited by omission
- [ ] Product-line classification for the fallback path is byte-identical to today — asserted by an
      existing-behaviour test, not by inspection
- [ ] `docs/architecture-overview.md` § 22 Returns updated to state the matcher is name-based no
      longer for the resolvable case (the current text still says "the matcher still joins by
      product name")

jakubretajczykBD:
## Summary
- Resolve a disposed return line's invoice line deterministically via `ReturnLine.resolvedOrderLineId` (#3171/#3172) → a new `InvoiceLine.orderLineId` (added in this scope, set by `toInvoiceLine` from `OrderItem.id`, deliberately never by `toShippingLines` — #3290 owns crediting the delivery line)
- By-name matching stays the fallback, unchanged, for a return whose own join is unresolved (a routine, ongoing state, not only historical) or an invoice predating the new field
- `status: 'ambiguous'` is retired as a live matcher outcome, not removed from the type: the same residual condition now reports `no-match('ambiguous-invoice-line')` — a highlighted, non-blocking row, never a candidate-picker (#3091 stays retired)

## Two behavioural consequences discovered while implementing, both fixed
- `isCorrectable()` checked `status === 'ambiguous'` to decide whether a proposal opens an `order_changes` row; it now also checks the new `noMatchReason` so the residual case stays exactly as correctable as the old `ambiguous` status was
- The service's summary log line counted `status === 'ambiguous'`, which would have silently under-reported to zero; it now counts the new reason instead

## FE
`return-proposal-panel.tsx` (the panel actually on `main` — #3090's `CorrectionProposalPanel` lives only on unmerged branches) extends its "needs attention" banner and per-line badge tone to the new reason, so it reads as attention-worthy rather than as a routine, closed exclusion.

## Docs
`docs/architecture-overview.md` § 22 Returns updated to reflect the id-first, name-fallback order and the retirement of `status: 'ambiguous'` as a live outcome.

## Test plan
- [x] `libs/core` unit tests: 437 suites / 6796 tests passing
- [x] `libs/core` type-check clean
- [x] `apps/web` FE panel + copy tests: 29/29 passing
- [x] `apps/web` type-check clean
- [x] `apps/api` type-check clean

Plan: `docs/plans/implementation-plan-narrow-return-correction-matcher.md`
Pre-implement gate: `docs/plans/analysis/ANALYSIS-narrow-return-correction-matcher.md` (READY)

Closes #3312


piotrswierzy (approved):
## Tech Lead review — `feat(returns): narrow the correction matcher onto ReturnLine.resolvedOrderLineId`

This is the prerequisite #3123's §5.8 amendment names by hand (*"narrowing `return-correction-matching.domain-service.ts` onto that field instead of matching by product name — until it is narrowed … no frontend sub-issue of #3087 should start"*). It does that, and the id path is correctly placed relative to the existing guards.

### ✅ I checked the part that could have gone badly: `orderLineId` does not reach a fiscal document

Adding a field to `InvoiceLine` puts it on the shape every `InvoicingPort` adapter receives, so the question is whether any adapter spreads it onto a provider payload. I grepped all four (KSeF, inFakt, Subiekt, eparagony) — every one maps field-by-field. The single spread is `ksef-invoicing.adapter.ts:811`:

```ts
const corrected = original.map((line) => ({ ...line }));
```

…which is `InvoiceLine → InvoiceLine`, an internal copy for `applyCorrectionDeltas`, and the FA(3) builder downstream still picks fields explicitly. So `orderLineId` rides into the persisted `issuedLineSnapshot` (#1297, jsonb — which is the consumer that needs it) and nowhere near the wire.

`toShippingLines` deliberately omitting it, with the reason (*no single `OrderItem` backs a delivery cost*) and a pointer to #3290 as the owner of that case, is the right treatment — a synthesized line carrying a borrowed id would be worse than carrying none.

### ✅ The fallback is a fallback, not a degradation

The docblock names three cases that legitimately land on by-name, and is careful about the second:

> *a return line whose own join is unresolved (`resolvedOrderLineId === null`, **which happens on returns ingested TODAY, not only historically**)*

That matters. `return-order-line-resolution.domain-service.ts` reports `ambiguous` / `no-candidate` outcomes as a routine result, so treating `null` as purely historical would have been wrong and would have invited someone to delete the by-name path later. Saying "falling back is exactly today's shipped behaviour, not a new guess" is the right framing for why the fallback isn't a hedge.

### ✅ Retiring `'ambiguous'` without making a proposal read as empty

The condition that used to produce `status: 'ambiguous'` now produces `no-match` / `'ambiguous-invoice-line'`, and the service's actionable-count predicate was updated to keep counting it:

```ts
return line.status === 'matched'
  || line.status === 'ambiguous'
  || line.noMatchReason === 'ambiguous-invoice-line';
```

That's the thing a status retirement usually breaks — the line stops being "actionable", the proposal reads as nothing-to-do, and an operator never sees a credit they owe. Same for the log line, which is explicitly counted so it doesn't under-report. Both have comments saying why.

`indexSnapshotByOrderLineId` building an **array** per id rather than a single entry is quietly good: two invoice lines sharing an `orderLineId` shouldn't happen, and if it ever does the line lands on `ambiguous-invoice-line` rather than silently picking the first.

### 🟢 SUGGESTION — `no-line-name` is checked ahead of the id path, and that excludes a resolvable line

```ts
if (line.name === null || line.name.trim() === '') {
  return noMatch(line, 'no-line-name', []);
}
// … only then the resolvedOrderLineId lookup
```

The comment justifies it as a sanity check ("every downstream candidate still carries a `name` the operator reads"). The consequence is that a return line OL knows the exact invoice line for — id resolved, unambiguous — is refused for want of a *display* string.

Since the whole premise of this PR is that the id is stronger evidence than the name, it's worth being deliberate about the one place the name still outranks it. Both readings are defensible (the operator does read a name on the proposal row), but if the ordering was inherited rather than chosen, the id path arguably belongs above it — a resolvable line with a blank name could match and render the *snapshot's* name.

### 🟢 SUGGESTION — the dead union member has no owner

> *`ReturnCorrectionLineStatusValues` still contains `'ambiguous'` so nothing that already pattern-matches on it breaks … Removing the dead value from the union is a separate, later cleanup.*

Right call for this PR, but "a separate, later cleanup" names no issue. A union member nothing produces is exactly what rots — and the actionable-count predicate above now has a permanently-false arm that reads as live. Worth filing so it isn't rediscovered by someone wondering why `'ambiguous'` never appears.

### Notes

- `docs/architecture-overview.md` updated in the same PR — the §22 Returns paragraph describing the by-name limitation would otherwise have become false on merge.
- The analysis + implementation-plan docs are committed alongside, matching the documented process.
- CI green across all ten.

**Verdict:** ✅ Approve. Both suggestions are worth a reply before merge, but neither blocks.

PR #3300, opened 2026-09-18
chore(main): release 0.12.0

tomektom:
:robot: I have created a release *beep* *boop*
---

### Features

* **eparagony:** eparagony issues invoices through the Invoicing capability ([#3304](https://github.com/openlinker-project/openlinker/issues/3304)) ([0281245](https://github.com/openlinker-project/openlinker/commit/028124508803bd1224b3a12550933f5050c0342c))
* **inventory:** close mockup-parity gaps on the duplicate-positions page ([#3264](https://github.com/openlinker-project/openlinker/issues/3264)) ([a0d1997](https://github.com/openlinker-project/openlinker/commit/a0d199702929bc682ab9a16f167eff60a395273a))
* **inventory:** duplicate-positions API client + query hook ([#3251](https://github.com/openlinker-project/openlinker/issues/3251)) ([df940b6](https://github.com/openlinker-project/openlinker/commit/df940b6414e4cfdbab38b083e8fd03cbbe31ec8b))
* **inventory:** duplicate-positions page with readiness KPI strip ([#3252](https://github.com/openlinker-project/openlinker/issues/3252)) ([4ffb39d](https://github.com/openlinker-project/openlinker/commit/4ffb39d7df1408f71037b13935684e627d02e0e8))
* **inventory:** enrich duplicate-positions response with product/variant/connection/location display names ([#3249](https://github.com/openlinker-project/openlinker/issues/3249)) ([9f0e02f](https://github.com/openlinker-project/openlinker/commit/9f0e02f04c9f7f0ac6a6adb3107de13ce5abc80b))
* **inventory:** expandable grouped table (per-row inventory_items detail) ([#3255](https://github.com/openlinker-project/openlinker/issues/3255)) ([c36ee2b](https://github.com/openlinker-project/openlinker/commit/c36ee2ba253b559b16286df72db51a79b5701cc7))
* **inventory:** expose [#2317](https://github.com/openlinker-project/openlinker/issues/2317) provenance-backfill completion state for the duplicate-positions readiness gate ([#3250](https://github.com/openlinker-project/openlinker/issues/3250)) ([fc9ddd4](https://github.com/openlinker-project/openlinker/commit/fc9ddd42d83ca17ea8c6d79e0bf2e3f7b9d361d1))
[...]

PR #3315, opened 2026-09-21
fix(price-changes): scope toast-copy assertions to avoid Radix announcer duplicate

### LINKED ISSUE #3314: [BUG] Frontend — flaky toast-copy assertion in price-changes-queue-table.test.tsx matches Radix's duplicated aria-live announcer
## Problem / Context

`apps/web/src/features/price-changes/components/price-changes-queue-table.test.tsx` — test `PriceChangesQueueTable > shows clean, actionable copy instead of a raw backend message on a stale-version conflict (#3148 review, finding 9)` (line 121) is flaky in CI.

Observed failure: [CI run 35590609829, job 106303953926](https://github.com/openlinker-project/openlinker/actions/runs/35590609829/job/106303953926?pr=3300) — a `TestingLibraryElementError: Found multiple elements` thrown from:

```ts
expect(
  await screen.findByText(/changed again while you were reviewing — refresh and take another look/),
).toBeInTheDocument();
```

(line 144-146). PR #3300, which this run was gating, touches no files under `features/price-changes/` — the failure is unrelated to that PR's diff.

**Root cause**: `ToastProvider` (`apps/web/src/shared/ui/toast-provider.tsx`) wraps children in `@radix-ui/react-toast`'s `<RadixToast.Provider>`. Radix Toast renders the toast description into two DOM locations — the visible `<RadixToast.Description>` node, and an internal delayed `aria-live="assertive" role="status"` announcer clone it uses for screen-reader announcements (see the existing comment in `toast-provider.tsx` referencing `@radix-ui/react-toast/dist/index.mjs:477` for a related, already-worked-around Radix timer). The announcer clone mounts on its own delayed timer, separate from the toast's initial render.

`screen.findByText(...)` polls via `waitFor` and resolves as soon as **any** match exists. When the announcer clone has not yet mounted, exactly one node matches and the assertion passes; when it has (a timing race dependent on CI's scheduling/load), two nodes match and Testing Library throws `Found multiple elements` instead of resolving. This is why the test passes reliably in local isolated and full-file runs (`pnpm vitest run price-changes-queue-table.test.tsx`, both green) but failed once in CI under different timing/parallelism.

## Proposed Solution

Scope the assertion to a single toast root instead of querying the whole document, so the duplicated `aria-live` announcer node can never produce an ambiguous match. The test file already imports and uses `within` (`apps/web/src/features/price-changes/components/price-changes-queue-table.test.tsx:1, :541, :556`) for exactly this kind of scoping elsewhere.

Concretely, at line 144-147:
- Resolve the toast root first (e.g. `await screen.findByRole('status')` narrowed to the visible toast, or a `data-testid`/class selector on `.toast` if a `role="status"` match is itself ambiguous because of the announcer), then assert the text `within` that root.
- Alternatively (if Radix's announcer role/structure makes root-scoping fragile), query specifically via `within(container).getByText(...)` against the `.toast__description` node produced by `RadixToast.Description`, which is the single canonical copy the test cares about.
- Apply the same scoping fix to any other test in this file (or elsewhere under `features/**/*.test.tsx`) that asserts toast copy via a bare `screen.findByText`/`getByText` without `within` — grep for `findByText` / `getByText` calls that match toast copy strings to confirm whether this is a single-site or repeated pattern.

Do not touch `ToastProvider` itself — the double-render is standard, correct Radix a11y behavior (visible content + live-region announcement for screen readers), not a bug to "fix" in the component.

## Classification

**Type**: Frontend
**Layer**: Interface (test infrastructure)
**File(s)**:
- `apps/web/src/features/price-changes/components/price-changes-queue-table.test.tsx` (the flaky assertion, ~line 144-147)
- `apps/web/src/shared/ui/toast-provider.tsx` (context only — not expected to change)

## Dependencies

- None

## Assumptions

- The fix is test-only; no production behavior changes are required or wanted.
- Other tests in this file/codebase that assert toast copy via unscoped `findByText`/`getByText` may share the same latent flake and should be checked, but this issue's primary scope is the one confirmed-failing assertion. If a broader sweep turns up more instances, they should be fixed in the same PR since the remedy (scope via `within`) is identical.

## Acceptance Criteria

- [ ] The stale-version-conflict test's toast-copy assertion is scoped (`within`) so it cannot match Radix's duplicated `aria-live` announcer node
- [ ] Any other unscoped toast-copy assertions found by the sweep in this file/codebase are fixed the same way
- [ ] The test passes reliably run 20x in a row locally (`pnpm vitest run price-changes-queue-table.test.tsx --repeat=20` or equivalent loop) with no flake
- [ ] No change to `ToastProvider`'s rendering/a11y behavior
- [ ] `pnpm lint`, `pnpm type-check`, `pnpm test` (web package) all pass

norbert-kulus-blockydevs:
## Summary

Fixes the flaky test found in [CI run 35590609829](https://github.com/openlinker-project/openlinker/actions/runs/35590609829/job/106303953926?pr=3300) (unrelated PR #3300 was just the bystander that happened to gate on it).

`ToastProvider` (Radix Toast) renders a toast's `title`/`description` into a second, delayed `aria-live="assertive"` announcer node alongside the visible content, for screen-reader support. Three assertions in `price-changes-queue-table.test.tsx` used a bare `screen.findByText(...)` against toast copy, which throws `Found multiple elements` whenever that announcer node has mounted by the time `findByText`'s `waitFor` poll resolves — a timing race dependent on CI scheduling, which is why it passed reliably in local runs but failed intermittently in CI.

- Switched all three to `findAllByText(...)` + `.length > 0`, matching the existing precedent already in this file for the identical Radix duplication on the steep-tooltip test (`findAllByText('Big price change - worth a second look')`).
- Verified via Radix's own source (`ToastAction`/`ToastClose` are wrapped in `ToastAnnounceExclude`) that the "Undo" action-label assertions (`getAllByText('Undo')`) were correctly *not* at risk and left untouched.
- Scope kept to the file that actually failed (`price-changes-queue-table.test.tsx`) rather than a codebase-wide sweep of every `showToast` caller — ~70 other test files use toasts; auditing all of them is a separate, much larger effort if wanted later.

## Test plan

- [x] `price-changes-queue-table.test.tsx` run 10x in a row locally — all green
- [x] `pnpm lint` (targeted eslint on the file) — clean
- [x] `pnpm type-check` — clean
- [x] Full `apps/web` `vitest run` — 566/566 files, 6105 passed / 1 skipped

Closes #3314


piotrswierzy (changes requested):
## Tech Lead review — `fix(price-changes): scope toast-copy assertions to avoid Radix announcer duplicate`

The diagnosis is exactly right — Radix Toast renders every toast twice, once visibly and once into a delayed `ToastAnnounce` portal, and `findByText` throws on the double match. But the repo already solved this, and the workaround here weakens two assertions to get around it.

### 🔴 BLOCKING — `test-utils.tsx` already ships four helpers for precisely this

`apps/web/src/test/test-utils.tsx:1178` carries a docblock describing this failure mode in the same terms as your comments, and prescribing the fix:

> *`@radix-ui/react-toast` renders every toast twice — once visibly inside the Viewport (`.toast__title` / `.toast__description` in our wrapper) and once inside a hidden `ToastAnnounce` portal for screen readers. The announce portal hides itself ~1000ms after mount, creating a race window where a plain `screen.getByText(...)` intermittently matches both copies and throws. **Scope toast-text assertions to the visible `.toast__*` elements to skip the announce portal entirely.***

```ts
export function findToastTitle(text)       { return screen.findByText(text, { selector: '.toast__title' }); }
export function findToastDescription(text) { return screen.findByText(text, { selector: '.toast__description' }); }
```

Five call sites already use them, and #3261 migrated a sixth in `order-fulfillment-tasks-panel.test.tsx` for this same flake.

Your own comments correctly identify which is which, so the substitution is mechanical:

```ts
// case 1 + case 2 — descriptions
expect(await findToastDescription(/changed again while you were reviewing — refresh and take another look/)).toBeInTheDocument();
expect(await findToastDescription(/Kept 1, but 1 failed/)).toBeInTheDocument();

// case 3 — title
expect(await findToastTitle('Turned on automatic pricing for 2 sources')).toBeInTheDocument();
```

That fixes the flake **and** keeps each assertion at exactly-one-match.

### 🔴 BLOCKING — the third case loses the guarantee its own comment names

```ts
// ONE toast naming both sources, not two independent ones.
expect(await screen.findByText('Turned on automatic pricing for 2 sources')).toBeInTheDocument();
```

becomes

```ts
const toastMatches = await screen.findAllByText('Turned on automatic pricing for 2 sources');
expect(toastMatches.length).toBeGreaterThan(0);
```

`findByText` asserted **exactly one**; `toBeGreaterThan(0)` asserts **at least one**. The comment above it says the point of the assertion is *"ONE toast naming both sources, not two independent ones"* — and after this change, two independent toasts pass.

The retained `expect(screen.getAllByText('Undo')).toHaveLength(1)` is a decent proxy and I'd keep it, but it's a proxy: it happens to hold because one toast carries one action. The text assertion was the direct statement of the invariant, and `findToastTitle` preserves it without reintroducing the flake.

The same weakening applies to cases 1 and 2, just without a stated invariant to point at — `>0` will not notice a duplicate toast in either.

### Notes

- The `#3314` issue reference in the comments is useful; worth keeping it in the reworked version, pointing at the helper rather than restating the mechanism (the helper's docblock is now the single explanation).
- CI green — which is expected, since the change only ever loosens.
- Nothing else in the file is touched, and the scope is right for a flake fix.

**Verdict:** ❌ Request changes. Swap to `findToastTitle` / `findToastDescription`; it's strictly less code than the workaround and keeps the assertions as strong as they were.

norbert-kulus-blockydevs:
Addressed piotrswierzy's review:

- BLOCKING (helpers already exist): swapped the three `findAllByText` + length-check workarounds for `findToastTitle`/`findToastDescription` from test-utils.tsx, which scope to the visible `.toast__*` elements and skip the Radix announcer duplicate without weakening assertions.
- BLOCKING (bulk-auto-pricing title lost its guarantee): now uses `findToastTitle(...)` + `toBeInTheDocument()`, restoring the exactly-one-match assertion that `toBeGreaterThan(0)` had silently dropped.
- Note (keep #3314 reference, point at the helper): kept the `#3314` reference in the comments, now pointing at the helper's docblock as the single explanation.

norbert-kulus-blockydevs:
@piotrswierzy - addressed both findings from the review:

**🔴 BLOCKING — swap to `findToastTitle`/`findToastDescription`**
Already fixed by a follow-up commit (`c4ae86ee5`, present before I picked this up): all three assertions now use `findToastTitle`/`findToastDescription` from `test-utils.tsx` instead of scoping the workaround inline.

**🔴 BLOCKING — third case loses the "exactly one" guarantee**
Same commit resolves this: the `findAllByText(...).length > 0` workaround is gone, replaced with `await findToastTitle('Turned on automatic pricing for 2 sources')`, which restores the exactly-one-match assertion the original `findByText` had. The `Undo`/`toHaveLength(1)` proxy assertion is kept alongside it.

**Note — keep the `#3314` reference pointing at the helper, not restating the mechanism**
Fixed in `5165e8917` (just pushed): trimmed all three inline comments so they point at the `findToastTitle`/`findToastDescription` docblock in `test-utils.tsx` (which is now the single explanation of the Radix double-render behavior) instead of re-describing it at each call site. Kept the substantive parts — the `#3314` reference, the "ONE toast naming both sources" invariant note, and the explanation for why `Undo` stays a plain assertion.

Verified: `pnpm --filter @openlinker/web type-check` clean, and `price-changes-queue-table.test.tsx` (21 tests) green across 4 consecutive runs.

piotrswierzy (approved):

Re-read at `5165e8917`. Both blocking points addressed.

**The test now uses the existing helpers.** `findToastTitle` / `findToastDescription` are imported from `test-utils` and used at the toast assertion sites. Those scope to the visible `.toast__*` elements and skip the Radix announce portal entirely — which is the mechanism that makes the duplicate-match problem go away rather than being worked around.

**The exactly-one guarantee is back.** Because the helpers scope to a single element, the assertions no longer need `findAllByText` + `toBeGreaterThan(0)`, so the property the test's own comment names — *one* toast naming both sources, not two independent ones — is once again what the test actually checks. That was the whole objection: the original fix made a flaky test pass by making it assert less than the thing it existed to assert.

The one surviving `findAllByText` (line 309) is a different assertion and a legitimate one — Radix tooltips render content twice (visible popper plus a visually-hidden accessible copy), the comment says so, and it cites the primitive's own `tooltip.test.tsx` doing the same. That is the right use of the plural query: a genuine, documented multiplicity rather than a way around a scoping problem.

**✅ Approve.**

PR #3317, opened 2026-09-21
fix(woocommerce): resolve store currency for ProductMaster reads

### LINKED ISSUE #3310: [BUG] Integration — WooCommerce ProductMaster never reports a currency, so no offer can be built from it
## Problem / Context

A WooCommerce `ProductMaster` reports `currency: null` on **every** product, so no
marketplace offer and no shop publish can be built from a WooCommerce-mastered
variant unless the caller passes an explicit price.

The option exists and is documented — it is simply never populated:

- `libs/integrations/woocommerce/src/infrastructure/mappers/woocommerce-product.mapper.types.ts:10-13`
  declares `WooCommerceProductMapperOptions.currency` ("ISO 4217 currency code assigned
  to every product").
- `libs/integrations/woocommerce/src/infrastructure/mappers/woocommerce-product.mapper.ts:44`
  reads it: `currency: this.options.currency ?? null`.
- `libs/integrations/woocommerce/src/woocommerce-plugin.ts:193` is the **only**
  production construction site, and it passes an empty bag:
  `const mapper = new WooCommerceProductMapper({});`

The only caller that ever supplies a currency is a unit test
(`.../__tests__/woocommerce-product.mapper.spec.ts:46`), so the option is covered by
tests and wired to nothing.

`WooCommerceConnectionConfig`
(`libs/integrations/woocommerce/src/domain/types/woocommerce-config.types.ts:30`)
carries `siteUrl`, `inventory` and `orders` — there is no currency field anywhere, so
an operator cannot supply one either.

Both builders then refuse:

- `libs/core/src/listings/application/services/offer-builder.service.ts:562`
- `libs/core/src/listings/application/services/product-publish-builder.service.ts:441`

both raising `REQUIRED` / `price.currency` with
*"Currency could not be resolved from input or master product; provide input.price explicitly"*.

**Observed live.** Publishing an Allegro offer from a WooCommerce-mastered variant
produced a batch whose every `offer_creation_records` row failed with exactly that
issue. Passing `sharedConfig.price` explicitly was the only way past it.

## Proposed Solution

Resolve the store currency from the shop itself, from the response the adapter
**already fetches**. `WooCommerceProductMasterAdapter.readStoreCountry()`
(`.../product-master/woocommerce-product-master.adapter.ts:903-912`) issues
`GET /wp-json/wc/v3/settings/general` and reads `woocommerce_default_country`;
`woocommerce_currency` is a row in that same response. No new endpoint, no new
operator configuration.

Note the ordering constraint: the mapper is constructed in the plugin factory
*before* any HTTP call, so the currency cannot simply be passed to its constructor.
Options, in order of preference:

1. **Resolve in the adapter and stamp it onto the mapped product** — add a
   `readStoreCurrency()` beside `readStoreCountry()`, cached on success the same way,
   and apply it to the product the mapper returned. Keeps the settings read in the one
   place that already owns it.
2. Give the mapper a currency *provider* (`() => Promise<string | null>`) instead of a
   value.
3. Add `currency` to `WooCommerceConnectionConfig` and thread `config.currency` in at
   `woocommerce-plugin.ts:193` — simplest, but asks the operator to restate something
   the shop already declares.

Follow `readStoreCountry`'s two stated rules verbatim: **a transport failure throws**
(it must not be recorded as "the store declares no currency"), and **only a successful
read is cached** (so a failure cannot pin a false `null` for the adapter's lifetime).

A store that genuinely declares no currency must keep producing today's
`REQUIRED` / `price.currency` issue — never a defaulted or guessed code.

## Classification

**Type**: Integration
**Layer**: Infrastructure
**File(s)**:
- `libs/integrations/woocommerce/src/woocommerce-plugin.ts` (construction site)
- `libs/integrations/woocommerce/src/infrastructure/adapters/product-master/woocommerce-product-master.adapter.ts` (`readStoreCountry` precedent)
- `libs/integrations/woocommerce/src/infrastructure/mappers/woocommerce-product.mapper.ts`
- `libs/integrations/woocommerce/src/infrastructure/mappers/woocommerce-product.mapper.types.ts`

## Dependencies

- None.

## Assumptions

- The intended source is the shop's own `woocommerce_currency` setting rather than a
  per-connection operator field; a shop that knows its currency should not need to be
  told it twice.
- Only the `ProductMaster` path is in scope. The order-source and product-publisher
  paths have their own currency handling and are not changed here.

## Acceptance Criteria

- [ ] A `ProductMaster` read against a store whose `/settings/general` declares
      `woocommerce_currency` returns that ISO code on `Product.currency`.
- [ ] `OfferBuilderService.buildCreateOfferCommand` succeeds for a WooCommerce-mastered
      variant with **no** explicit `input.price`, taking amount and currency from the master.
- [ ] `ProductPublishBuilderService` likewise resolves currency from the master.
- [ ] A store that declares no `woocommerce_currency` still yields the existing
      `REQUIRED` / `price.currency` issue — no defaulted or inferred code.
- [ ] A transport failure on `/settings/general` does not cache a `null` currency and
      does not silently become "the store declares none".
- [ ] The settings response is not fetched twice per adapter instance when both country
      and currency are needed.

norbert-kulus-blockydevs:
## Summary

- `WooCommerceProductMasterAdapter` mapped every product with `currency: null` — `WooCommerceProductMapperOptions.currency` exists on the mapper but is never populated in production (`createWooCommercePlugin` constructs `new WooCommerceProductMapper({})` before any HTTP call can run to learn the store's currency), so no offer or shop publish could be built from a WooCommerce-mastered variant without an explicit `input.price`.
- Resolves the store's currency from `GET /wp-json/wc/v3/settings/general` (the `woocommerce_currency` row) — the same endpoint `readStoreCountry()` already reads for tax-rate resolution — and stamps it onto the product **after** the mapper runs (`withStoreCurrency`), since the mapper can't be given the value at construction time.
- Both settings reads (`readStoreCountry` and the new `readStoreCurrency`) now share one cached fetch via `readGeneralSettings()`, so a read that needs both never hits `/settings/general` twice.
- Follows `readStoreCountry`'s established contract verbatim: a transport failure on `/settings/general` **throws** (never silently recorded as "the store declares no currency"), and **only a successful read is cached**, so a transient failure can't pin a false `null` for the adapter's lifetime.
- A store that genuinely declares no `woocommerce_currency` still produces the existing `REQUIRED` / `price.currency` refusal in `OfferBuilderService` / `ProductPublishBuilderService` — never a defaulted or guessed currency code.

## Where it's applied

`getProduct`, `getProducts`, `createProduct`, `updateProduct` — every `ProductMaster` method that returns a mapped `Product` now carries the resolved store currency (only the `ProductMaster` path is in scope, per the issue; order-source and product-publisher currency handling are untouched).

## Test plan

- [x] Added unit tests covering: currency resolved on `getProduct`/`getProducts`, no-currency-declared keeps `null` (no default/guess), a transport failure on `/settings/general` throws and is never cached, a failed read is retried (not cached) on the next call, and `/settings/general` is fetched at most once per adapter instance when both country and currency are needed.
- [x] `pnpm --filter @openlinker/integrations-woocommerce test` — 541/541 passing.
- [x] `pnpm --filter @openlinker/integrations-woocommerce lint` — 0 errors (2 pre-existing warnings unrelated to this change).
- [x] `pnpm type-check` (full repo) — passes.
- [x] `pnpm check:invariants` — passes.

Closes #3310


norbert-kulus-blockydevs:
## Tech-lead review

**Summary**: Fixes the WooCommerce `ProductMaster` currency gap cleanly — the mapper's `currency` construction option was a dead field since the mapper is instantiated before any HTTP round-trip. Resolving via `GET /wp-json/wc/v3/settings/general` and stamping post-mapping (`withStoreCurrency`) is the right shape given the constraint, and sharing the fetch with `readStoreCountry()` via `readGeneralSettings()` avoids doubling requests.

### Correctness
- ✅ Follows `readStoreCountry`'s established contract exactly: transport failure throws and is never cached; only a successful read is memoized (`this.generalSettings`/`this.storeCurrency` stay `undefined` on failure, so the next call retries).
- ✅ "Store declares no currency" still produces `currency: null` — no defaulted/guessed code, preserving the downstream `REQUIRED`/`price.currency` refusal in the builders.
- ✅ `getProducts` resolves currency once per page (not once per row) — reasonable, since a store has exactly one currency.

### Minor / non-blocking
- `getProducts` fetches `readStoreCurrency()` unconditionally even when `validProducts` is empty after filtering — a wasted round-trip on an edge case, not worth gating for.
- `withStoreCurrency` (used by `getProduct`/`createProduct`/`updateProduct`) and the inline `currency ?? mapped.currency` in `getProducts` implement the identical "overlay if present" rule twice. Could have been the same helper applied uniformly, but not worth blocking on — the four call sites are simple enough that the duplication is inert.

### Tests
Good coverage: resolution on `getProduct`/`getProducts`, no-currency-declared stays `null`, transport failure throws + isn't cached, retried on next call, and the single-fetch-when-both-needed assertion (proves `readGeneralSettings()` sharing actually works, not just that it compiles).

**Verdict**: ✅ Approve — matches the codebase's established country-resolution pattern, scoped correctly to `ProductMaster` only per the issue, no architecture-boundary or layering concerns.

piotrswierzy (approved):
## Tech Lead review — `fix(woocommerce): resolve store currency for ProductMaster reads`

Clean fix for a real gap, and the two rules it inherits from `readStoreCountry` are the ones that matter.

### ✅ `null` means "the store declares none", never a default

```ts
this.storeCurrency = value ? value.toUpperCase() : null;
```
…and `withStoreCurrency` overlays only when there is something to overlay, so a store with no declared currency keeps `Product.currency: null` and the builders' existing `REQUIRED` / `price.currency` refusal fires. That's the right call: a guessed or defaulted ISO code here would silently publish a price in the wrong currency, which is worse than a refused publish. The docblock says so explicitly and a test pins it.

### ✅ A transport failure throws, and is not cached

Two separate properties, both tested:

- `readGeneralSettings()` caches **only** on success, so a 500 cannot pin a false `null` for the adapter's lifetime. The test drives call 1 → reject, call 2 → succeed, and asserts the second read answers `PLN`. That's the assertion that would catch a naive `try { … } catch { this.storeCurrency = null }`.
- The failure propagates rather than being recorded as "declares none" — same distinction `readStoreCountry` already draws.

### ✅ One `/settings/general` read for two different rows

Sharing the cache between `readStoreCountry()` and `readStoreCurrency()` — and proving it with `expect(settingsCalls).toBe(1)` after a `getProduct` **and** a `readProductTaxRate` — is the right shape. The two consumers read different rows of one response; two fetches would have been invisible except in request count, which is exactly the class of thing that only a counting assertion catches.

The pre-existing `getProduct` test needed a path-aware mock because a bare `mockResolvedValue` would answer `/settings/general` with a product body and `.find` would throw. Fixing that in place with the reason stated is better than widening the mock silently.

### 🟢 SUGGESTION — one rule, two spellings

`getProducts` resolves the currency once for the page (correct — it can't `await` per row) but then applies it differently from the single-product path:

```ts
// getProducts
return { ...mapped, currency: currency ?? mapped.currency, id: internalId };

// withStoreCurrency
return currency ? { ...product, currency } : product;
```

Equivalent today, because `readStoreCurrency` can only return a non-empty string or `null`. But it's the same rule written twice, and the two would diverge if that ever returned `''`. A small sync helper taking the already-resolved currency would let both paths share one expression:

```ts
private applyStoreCurrency(product, currency) { return currency ? { ...product, currency } : product; }
```

### 🟢 Worth naming — `getProducts` gains a new failure mode

`readStoreCurrency()` is awaited outside the per-row loop and outside any catch, so a transient `/settings/general` 500 now fails a whole catalogue **page** that previously succeeded (before this, `getProducts` never read settings at all).

I think that's the right trade — swallowing it would mean publishing a page of products with `currency: null`, i.e. the exact defaulted-answer the design refuses — and the sweep's retry ladder covers a transient. But it is a behaviour change on the bulk path, and the batched sweep (`master.product.syncBatch`) is the caller, so it's worth a line in the PR description rather than being discovered from a failed page.

### Notes

- Cache is per adapter instance, and the docblock says so. Correct framing given `getCapabilityAdapter` builds a fresh adapter per call — the win is within one resolution, not across the sweep.
- The docblock explaining *why* the mapper's `currency` construction option can't carry this (mapper built in the plugin factory before any HTTP call can run) is the useful part — it pre-empts the obvious "why not just pass it to the mapper?" refactor.
- CI green across all nine.

**Verdict:** ✅ Approve.

norbert-kulus-blockydevs:

- **Suggestion "one rule, two spellings"**: fixed. Extracted a shared private `applyStoreCurrency(product, currency)` helper and pointed both `getProducts` and `withStoreCurrency` at it, so the overlay-only-when-resolved rule exists in one place instead of two slightly different expressions.
- **Suggestion "getProducts gains a new failure mode"**: not a code change - the reviewer flagged this as worth stating in the PR description (a transient /settings/general 500 now fails a whole catalogue page fetch that previously succeeded). I have not edited the PR description; leaving that to the PR author/reviewer to add if desired.

No blockers or important-severity items were raised on this PR - the review was an approve with two suggestions only.

PR #3268, opened 2026-09-15
feat(web): give the eparagony.pl connection config real form fields

norbert-kulus-blockydevs:
An eparagony.pl connection carries eleven config keys and the guided wizard collected three of them. Everything else — whether a paper receipt prints, how the payment is described on it, which device slot a rate-less line falls back to — was reachable only by hand-editing the raw **Config JSON** textarea, because the plugin declared no `ConnectionConfigContribution` at all.

That was a stated v1 decision, and it did not hold up: these are fiscal settings an operator is expected to get right, and a mistyped JSON key fails silently as "the setting did nothing". It is the same finding #2610 made about `stockSafetyBuffer` / `pricingRule`.

The trigger was a live operator question: *how do I turn on paper receipts?*

## What an operator gets

Eight fields in four groups, only the first open — a single strip of eight controls is the wrong shape for settings most operators touch once.

| Group | Fields |
|---|---|
| **Receipt** (open) | Paper receipt (3-state), payment form, payment name |
| **Fallback tax rate** (collapsed) | Device slot A–G + hazard infotip |
| **Diagnostics and timing** (collapsed) | Device wait time, fiscal device number |
| **Testing overrides** (collapsed) | API host, sign-in host |

`environment` and `posId` stay in the guided wizard. **No new CSS** — every class already existed. There *is* a small, deliberate backend change — see "A real backend fix, not a form-only PR" below.

### `print` is three-state, not a checkbox

The document mapper tests `config.print === true`, so an explicit `false` and an absent key produce the same receipt. #2610's rule still applies though: an operator's explicit choice and an unset knob are different persisted states that must round-trip apart. The copy says plainly that the two produce the same receipt, rather than implying a difference in outcome that is not there.

### The fallback infotip contradicts the adapter's own docblock, on purpose

`tax-rate.policy.ts` says *"Arm 3 is unreachable while the #2252 gate stands"*. Traced through the code:

- that gate is `FiscalRegistrationService.assertEveryLineHasATaxRate`;
- its first statement is `if (!isTaxRateEnforced(cmd.taxRateEra)) return;`;
- `parseTaxRateStrictEnabled` enables the switch only for the literal string `'true'`.

`OL_TAX_RATE_STRICT_ENABLED` is **off by default**, so the fallback *is* reachable on an ordinary deployment. Repeating the unconditional form of that claim would have reassured an operator about a receipt that really can go out at a rate nobody confirmed, so the infotip states the conditional version.

### The infotip sits in the field's `description`, not its `label`

[...]

PR #3318, opened 2026-09-21
fix(web): eparagony.pl unrecognised-select can now be fixed by re-selecting its own recommended option

### LINKED ISSUE #3311: [BUG] Frontend — eparagony.pl unrecognised-value select can't be fixed by re-selecting its own recommended option
## Problem / Context

On the eparagony.pl connection edit form (`apps/web/src/plugins/eparagony/components/eparagony-structured-section.tsx`), two `<select>` fields — `eparagonyPaymentForm` and `eparagonyDefaultTaxRateCode` — can get stuck in a state where the save silently fails and the recommended fix does nothing.

`readEnum` (`apps/web/src/plugins/eparagony/eparagony-connection-config.ts:204-206`) narrows an unrecognised stored value to `''` when hydrating the RHF field, which is also the value of the recommended "Not set" `<option>`. So when a connection's persisted config carries a value this build doesn't recognise (a value from a retired vendor vocabulary, a hand-edited raw-JSON typo, etc.), the `<select>` renders already-selected on `''`/"Not set". Clicking that same, already-selected option fires no `change` event (neither the DOM nor React consider it a change), so `syncStructuredToJson` never runs, `configText` keeps carrying the stale unrecognised value, and the save fails identically — the operator has followed the on-screen instruction and nothing happened.

This was found and documented, but not fixed at the mechanism level, during the #3268 review (PR #3268, "feat(web): give the eparagony.pl connection config real form fields"). The current mitigation is copy-only: the warning text now says "pick a *different* value first, then switch back if the default was intended" instead of the unqualified "pick one below" (`eparagony-connection-config.ts:217-231`). That keeps the escape hatch discoverable but doesn't fix the underlying no-op.

## Proposed Solution

Detect the unrecognised-value case and force a write regardless of the `<select>`'s rendered state, instead of relying on a real `change` event. Two shapes are plausible and either is acceptable:

1. **Remount on mismatch** — give the `<select>` a `key` derived from whether the field is currently showing an unrecognised value (e.g. `key={hasUnrecognisedValue ? 'unrecognised' : 'recognised'}`), so selecting the *same* option after a remount is a genuine DOM/React change and fires `onChange` normally.
2. **Explicit write on click** — detect the click on the already-selected "Not set" option (or wrap it in a button-like control) and call `syncStructuredToJson` directly, bypassing the `onChange` dependency entirely.

Whichever shape is chosen, update the warning copy back to the simpler, unqualified phrasing once the mechanism is fixed (the copy carve-out at `eparagony-connection-config.ts:217-231` should be removed together with the fix, not left behind as a stale caveat).

Existing tests to extend: `apps/web/src/plugins/eparagony/components/eparagony-structured-section.test.tsx` (there's already a suite covering the unrecognised-value warning banner — add a case that clicks the already-selected recommended option and asserts `syncStructuredToJson` fires / `configText` updates).

## Classification

**Type**: Frontend
**Layer**: Interface (component behaviour)
**File(s)**:
- `apps/web/src/plugins/eparagony/components/eparagony-structured-section.tsx`
- `apps/web/src/plugins/eparagony/eparagony-connection-config.ts` (`readEnum`, `readUnrecognisedEnumValue`, lines ~190-231)

## Dependencies

- None. #3268 (the PR that introduced and documented this limitation) is already merged/mergeable independently of this fix.

## Assumptions

- The fix should apply to both affected fields (`eparagonyPaymentForm`, `eparagonyDefaultTaxRateCode`) uniformly, since they share the same `readEnum`/`readUnrecognisedEnumValue` mechanism — not just one of them.
- The `print` field (`SegmentedControl`) is unaffected and needs no change: it fires `onChange` from a plain `onClick` unconditionally, so the trap described here does not apply to it (confirmed in the #3268 re-review).

## Acceptance Criteria

- [ ] Clicking the already-selected recommended option on `eparagonyPaymentForm` when the field holds an unrecognised stored value results in a real config write (`configText` no longer carries the stale value) and the warning clears.
- [ ] Same behaviour verified for `eparagonyDefaultTaxRateCode`.
- [ ] Selecting any *other* option continues to work exactly as before (no regression on the already-correct path).
- [ ] The warning copy is simplified back to not needing the "pick a different value first" qualifier, once the mechanism no longer requires it.

norbert-kulus-blockydevs:
## Summary
- `eparagonyPaymentForm` / `eparagonyDefaultTaxRateCode` hydrate to `''` when the stored config value is unrecognised — the same value as the recommended "Use the default" / "Not set" `<option>`. The `<select>` therefore rendered already-selected on the fix, so re-clicking that same option fired no `onChange` and the save failed identically forever.
- Fix: the `<select>`'s controlled `value` is now pinned to the raw unrecognised string via a hidden, disabled sentinel `<option value={raw} hidden disabled>` whenever one is present, never to the field's narrowed `''`. The DOM's real selected option is therefore the raw stored value, so re-picking the recommendation is a genuine value change and fires `onChange` normally.
- `readEnum` / `readUnrecognisedEnumValue` in `eparagony-connection-config.ts` are untouched — only the select's rendered/selected value is repointed, inside the component.
- Both affected selects (payment form, fallback tax-rate slot) are fixed. Warning copy reverts to the simpler "pick one below" phrasing now that it's accurate again.

## Test plan
- [x] `pnpm lint` — 0 errors (repo-wide)
- [x] `pnpm type-check` — pass (repo-wide)
- [x] `pnpm exec vitest run src/plugins/eparagony` — 70/70 passing, including 8 new regression tests covering all 4 acceptance criteria from #3311
- [ ] Full monorepo `pnpm test` left to CI (local machine constraint)

Closes #3311


norbert-kulus-blockydevs:
## Tech-lead review

**Summary**: Correctly diagnoses and fixes the root cause — an unrecognised stored value hydrates the RHF field (and hence the controlled `<select>`) to `''`, the same value as the recommended option, so re-selecting it fires no `onChange` in a real browser. Rather than touching the field-narrowing logic (`readEnum`/`readUnrecognisedEnumValue`, left untouched as stated), the fix repoints the `<select>`'s *rendered/selected* value to the raw unrecognised string via a hidden, disabled sentinel `<option>`, so the DOM's actual selection differs from the recommended option and re-picking it is a genuine change.

### Correctness
- ✅ `value={unrecognised.paymentForm ?? (form.watch(...) ?? '')}` — correct precedence: only overrides when there's genuinely something unrecognised to pin against.
- ✅ `hidden disabled` on the sentinel option is right: `hidden` keeps it out of the dropdown list, `disabled` prevents it ever being re-selected once the operator moves off it — so it can only ever be the *initial* displayed value, never a destination.
- ✅ Doesn't touch `readEnum`/`readUnrecognisedEnumValue` — correctly scoped to the render layer, per the docblock's own updated reasoning (the field-narrowing to `''` is fine; it's the select's *displayed* value that needed to diverge from it).
- ✅ Warning copy simplification ("pick one below" unqualified) is now accurate — the `#3268`-review qualifier ("pick a different option first, then switch back") is genuinely obsolete once re-picking the recommended option itself works.
- ✅ Non-regression preserved: picking a genuinely different value still works (asserted explicitly in a new test).

### Minor / non-blocking
- The sentinel option's visible text is `"{value} (not recognised)"` — reasonable, though since it's `hidden` from the dropdown and only ever shown as the pre-interaction selected value, this is mostly cosmetic (native `<select>` rendering of the closed box varies slightly by browser for a hidden-but-selected option). Not worth blocking on; low risk.
- Accessibility: worth a quick manual screen-reader spot-check that the initial announced value reads as the raw unrecognised string rather than something confusing — likely fine given standard `<select>`/`<option>` semantics, but this PR's test suite is DOM/RTL-only so it isn't verified here.

### Tests
8 new tests covering both affected selects (payment form, tax-rate fallback slot): DOM value assertion (proves the actual fix, not just downstream behavior), fix-via-recommended-option, non-regression on a different pick, and simplified copy. Good — the "assert the select's own `.value`" test is the one that actually pins the underlying bug rather than just the symptom.

**Verdict**: ✅ Approve — correct root-cause fix, correctly scoped, good regression coverage.

piotrswierzy (approved):
## Tech Lead review — `fix(web): eparagony.pl unrecognised-select can now be fixed by re-selecting its own recommended option`

This converts a copy-level workaround into a real fix, and the mechanism is the right one.

### ✅ The trap, and why the sentinel option breaks it

The original problem is subtle and was documented honestly rather than hidden: `readUnrecognisedEnumValue` narrows the RHF field to `''`, which is *also* the value of "Use the default" / "Not set", so the controlled `<select>` rendered already-selected on the recommended option — and clicking an already-selected option fires no `onChange` in either the DOM or React. The operator's obvious remedy was a no-op.

The previous shipped answer was to reword the warning ("pick a DIFFERENT option first, then switch back"), which works but asks an operator to perform a two-step dance to clear a field. Rendering a hidden sentinel `<option>` carrying the raw stored value, and pointing the controlled `value` at it, means the DOM's real selection is `Bitcoin` — so re-picking the recommended option **is** a value change and fires normally.

Worth noting the half that makes it valid: `disabled` prevents *user* selection but not *programmatic* selection, so React setting `select.value = 'Bitcoin'` still selects it; `hidden` keeps it out of the dropdown while leaving it visible in the collapsed control as `Bitcoin (not recognised)`. That's also better UX than the old behaviour, which displayed the operator's unrecognised value as "Use the default" — a false statement about their own data, which the original comment already called out.

### ✅ The test that actually proves it

```ts
expect(screen.getByLabelText('Payment form on the receipt')).toHaveValue('Bitcoin');
```

Asserting the **select's own DOM value** is the assertion that distinguishes this fix from the copy-only one, and the comment says so. The two behavioural tests either side of it (re-picking `''` now syncs; picking `'Karta'` still syncs) cover the fix and its regression risk. Both fields get the same four cases rather than one being assumed to follow the other — right call, since they're separate JSX blocks.

### ✅ The docblock was rewritten, not just amended

`readUnrecognisedEnumValue`'s `KNOWN LIMITATION … tracked as #3311` becomes `FIXED (#3268 review, #3311)` and explains that this function is *unchanged* — the fix is in the consuming component. That matters: the next reader of `readUnrecognisedEnumValue` would otherwise still find a limitation notice for a limitation that no longer exists, which is the same stale-doc failure mode as a stale mockup.

### 🟢 SUGGESTIONS

1. **Nothing asserts the sentinel stays out of the dropdown.** The behaviour depends on `hidden` + `disabled`, and a future tidy-up removing either would pass every test here (the value assertion and the change assertions all still hold) while putting a bogus option back in the operator's list. One assertion — e.g. the option is not among `getAllByRole('option')`, or carries `hidden` — pins it cheaply.
2. **`disabled` is belt-and-braces and could say so.** Once the operator moves off the sentinel, `unrecognised` goes null and the option unmounts entirely, so "keeps it from ever being reselected" describes a window that closes immediately. Harmless, and I'd keep it, but the comment slightly overstates its role.

### Notes

- The simplified copy (`…will be refused until you pick one below.`) is asserted positively **and** the old convoluted sentence is asserted absent — the right pair, since a copy change that leaves the old string somewhere else would otherwise pass.
- `check-eparagony-config-mirror.mjs` guards the default-payment-form wording and stayed green, so the mirror still covers the branch that carries it.
- CI green across all nine.

**Verdict:** ✅ Approve.

norbert-kulus-blockydevs:
Addressed both suggestions from piotrswierzy's review (approve, no blockers):

1. "Nothing asserts the sentinel stays out of the dropdown" - added a regression test per select (payment form, fallback slot) that asserts the hidden sentinel option is excluded from `getAllByRole('option')` (testing-library's accessibility-tree-based role query respects `hidden`), pinned separately from the existing `.toHaveValue()` assertion so a future removal of the `hidden` attribute fails a test instead of silently re-adding a bogus option to the operator's dropdown.
2. "`disabled` is belt-and-braces and could say so" - reworded the code comment: `disabled` does not "keep it from ever being reselected" since the sentinel option unmounts the moment the operator picks any real value; it's a defensive backstop, not the active guard.

norbert-kulus-blockydevs:
## Re-review (verifying the follow-up to piotrswierzy's approved review)

Verified both suggestions from piotrswierzy's ✅ Approve review were genuinely addressed by commit [\`5cf9aca61\`](https://github.com/openlinker-project/openlinker/commit/5cf9aca61efbe5ad1587d50b219518c691fe8ef4), not just claimed in the follow-up comment.

**1. "Nothing asserts the sentinel stays out of the dropdown"** — fixed correctly. The new test uses \`within(select).queryByRole('option', {name: /not recognised/i})\`, and RTL's role queries are accessibility-tree-based and correctly exclude a \`hidden\` element — so this assertion genuinely fails if a future edit drops the \`hidden\` attribute (verified this is a real behavioral pin, not a tautology, by checking how \`dom-accessibility-api\` treats \`hidden\`). Paired with the raw \`querySelectorAll('option')\` count (12/9) to confirm the sentinel is still in the DOM, just correctly excluded from the accessible list. Both selects covered.

**2. "\`disabled\` is belt-and-braces and could say so"** — the comment is now accurate: it correctly states \`disabled\` is a backstop rather than the active guard, since the sentinel option unmounts the instant \`unrecognised.*\` goes \`null\` on any real pick. No functional change, comment-only, as it should be for a wording-accuracy note.

No regressions, no new issues, scope stayed exactly to the two flagged suggestions. CI was already green on the prior commit.

**Verdict**: ✅ Approve — nothing outstanding on this PR.

PR #3319, opened 2026-09-21
test(orders): scope order-column-sort int-spec queries to avoid cross-file leak

### LINKED ISSUE #3316: [BUG] Infrastructure — flaky order-column-sort.int-spec.ts asserts against unfiltered order_records table
## Problem / Context

`apps/api/test/integration/order-column-sort.int-spec.ts` — the first test in `describe('Order column sort (integration)')`, `'sorts by total descending (default direction), NULLs last'` (~line 44-63) — is flaky in CI.

Observed failure: [CI run 35592516983, job 106309902528](https://github.com/openlinker-project/openlinker/actions/runs/35592516983/job/106309902528?pr=3300), "Run integration tests" step (api package, `jest-integration.cjs`). A re-run of the identical commit passed. PR #3300 touches neither this file nor `OrderRecordRepository` — unrelated.

Failure (jest diff):

```
expect(received).toEqual(expected)

- Expected  - 0
+ Received  + 1

  Array [
    "ol_order_fixture_1789989691788_29776",
+   "ol_order_610255a7cf7b4e85933b2e85eba8c328",
    "ol_order_fixture_1789989691778_42039",
    "ol_order_fixture_1789989691791_79394",
  ]

  at test/integration/order-column-sort.int-spec.ts:61:49
```

**Root cause**: the test calls `repository.findMany({ sort: 'total', dir: 'desc' }, PAGE)` with no scoping predicate — it queries the entire `order_records` table and asserts the result is exactly the 3 rows it just seeded via `createTestOrderRecord`. The unexpected extra row's id (`ol_order_610255a7cf7b4e85933b2e85eba8c328`) does not match this suite's fixture id format (`ol_order_fixture_<timestamp>_<random>`, minted by the `createTestOrderRecord` test helper) — it's shaped like a normally-minted order id (`formatInternalId('Order')`), meaning some other integration test created it through a real order-creation code path and it leaked into this test's unfiltered query.

Each Jest worker owns its own dedicated Postgres database (per the comment in `apps/api/test/jest-integration.cjs`: "Every worker owns its own Postgres database and Redis logical DB... a second worker's `TRUNCATE ... CASCADE` + `flushDb()` reset would wipe a peer's data mid-test" — cross-worker pollution is ruled out by design), so the leak is cross-*file* within one worker: another `*.int-spec.ts` file ran earlier in the same worker, created an order via a real order-creation flow, and that row survived because there's no guaranteed DB reset at the *start* of a new test file — `apps/api/test/integration/setup.ts`'s `resetTestHarness()` (which does truncate `order_records`, confirmed in its explicit truncate list) is invoked only from each file's own `afterEach` hook. Jest's file-to-worker assignment and execution order differ between local and CI runs, consistent with this only reproducing in CI and clearing on re-run.

This unfiltered-whole-table assertion style is fragile to *any* cross-file leak regardless of its origin. The codebase already has an established, repeated pattern in `setup.ts` acknowledging exactly this class of problem (comments like "...or a prior case's rows leak into a ...", present at multiple truncate-list entries).

## Proposed Solution

Scope the query to the connection this test seeds under, so it can never see another file's unrelated leaked order regardless of that row's own connection.

- `OrderRecordFilters.sourceConnectionId` (`libs/core/src/orders/domain/types/order-record.types.ts`) is already applied as a `WHERE` predicate by `OrderRecordRepository.buildFilteredQuery` (`libs/core/src/orders/infrastructure/persistence/repositories/order-record.repository.ts:257-259`).
- The test already seeds all 3 rows under a fixed `SOURCE` constant (`sourceConnectionId: SOURCE`, line 24). Pass `{ sourceConnectionId: SOURCE, sort: 'total', dir: 'desc' }` to `findMany` in this test (and any sibling `it` blocks in the same file with the same unscoped-whole-table assertion shape — check the other `it` blocks in `order-column-sort.int-spec.ts` for the same pattern, e.g. the `items` / `customer` / `status` sort cases and the non-numeric-total guard case, and apply the same scoping there if they share it).
- Do not weaken the sort assertion itself — scoping by `sourceConnectionId` only narrows which rows are visible to the query; the ordering behavior under test is unaffected.
- Do not attempt to hunt down and fix whichever other test file is the actual leak source — that is a much larger, unbounded investigation (bisecting Jest's per-worker file execution order across ~190 int-spec files) and the established remedy in this codebase for this exact class of problem is to scope the assertion, not chase every possible upstream leak.

## Classification

**Type**: Infrastructure (test infrastructure)
**Layer**: Interface (integration test)
**File(s)**:
- `apps/api/test/integration/order-column-sort.int-spec.ts` (the flaky test(s))
- `libs/core/src/orders/domain/types/order-record.types.ts` (context only — `OrderRecordFilters.sourceConnectionId`, no change expected)
- `libs/core/src/orders/infrastructure/persistence/repositories/order-record.repository.ts` (context only — confirms the filter is already wired, no change expected)

## Dependencies

- None

## Assumptions

- The fix is test-only; no production repository/service behavior changes are required or wanted.
- Other `it` blocks in this same file that use the same unfiltered `findMany(...)` + exact-array assertion shape share the same latent flake and should be fixed identically in the same PR (grep the file for `findMany({ sort:` and check each call's filter object).
- The actual source of the leaked order (which other int-spec file creates it) is out of scope for this issue — worth noting in the PR if easily discovered while investigating, but not a blocking requirement.

## Acceptance Criteria

- [ ] The `'sorts by total descending...'` test scopes its `findMany` call with `sourceConnectionId: SOURCE`
- [ ] Any other `it` block in `order-column-sort.int-spec.ts` sharing the same unscoped-whole-table assertion pattern is scoped identically
- [ ] `pnpm --filter @openlinker/api test:integration -- order-column-sort` (or equivalent) passes reliably run repeatedly
- [ ] No change to `OrderRecordRepository` or `OrderRecordFilters` production behavior
- [ ] `pnpm lint`, `pnpm type-check`, and the full integration suite pass

norbert-kulus-blockydevs:
## Summary

Fixes the flaky integration test found in [CI run 35592516983](https://github.com/openlinker-project/openlinker/actions/runs/35592516983/job/106309902528?pr=3300) (unrelated PR #3300 was just the bystander that happened to gate on it; a re-run of the same commit passed).

`order-column-sort.int-spec.ts` asserted `findMany(...)` results against the **entire, unfiltered** `order_records` table in every `it` block. `apps/api/test/jest-integration.cjs` gives every Jest worker its own dedicated Postgres database (cross-worker pollution is ruled out by design), but there's no guaranteed DB reset at the *start* of a new test file — `resetTestHarness()` (which does truncate `order_records`) only runs from each file's own `afterEach` hook. So a row created by a real order-creation flow in another `*.int-spec.ts` file running earlier in the same worker could survive into this file's first test and land inside the sorted result, breaking its exact-array assertion.

Observed in CI:

```
Array [
  "ol_order_fixture_1789989691788_29776",
+ "ol_order_610255a7cf7b4e85933b2e85eba8c328",
  "ol_order_fixture_1789989691778_42039",
  "ol_order_fixture_1789989691791_79394",
]
```

The extra id doesn't match this suite's fixture format (`ol_order_fixture_<timestamp>_<random>`) — it's shaped like a normally-minted order id from a different file's real order-creation path.

All 7 `it` blocks in the file already seed their rows under a fixed `SOURCE` connection id, so each `findMany(...)` call is now scoped with `{ sourceConnectionId: SOURCE, ... }` — confirmed applied as a `WHERE` predicate in `OrderRecordRepository.buildFilteredQuery`. This can never see another file's unrelated row regardless of its own connection. Sort behavior under test is unaffected; only which rows are visible to the query is narrowed.

Root cause of *which* other file leaks the row is out of scope — bisecting Jest's per-worker file execution order across ~190 int-spec files is a much larger, separate investigation, and this codebase already has an established remedy for this exact class of problem (scoping the assertion), used repeatedly in `apps/api/test/integration/setup.ts`'s truncate-list comments.

## Test plan

- [x] Scoped run (`--testPathPattern` anchored to avoid a worktree-dir-name collision with the target file's own name) — 10x in a row, all green, 7/7 tests
- [x] `pnpm exec eslint` on the file — clean (confirmed the two pre-existing `consistent-type-imports` errors on this file predate this change, unrelated)
- [x] `tsc -p tsconfig.type-check.json` / `tsconfig.test-check.json` — clean
- [x] Full integration suite (192 suites) also run as a side effect of an earlier mis-scoped invocation — only 4 unrelated failures, a `Port ... not bound after 180000ms` Testcontainers PrestaShop-container timeout in `allegro-prestashop-carrier-mapping.int-spec.ts` (local resource contention, unrelated to this change); `order-column-sort` was not among the failures

Closes #3316


piotrswierzy (approved):
## Tech Lead review — `test(orders): scope order-column-sort int-spec queries to avoid cross-file leak`

Right shape of fix: scoping an exact-array assertion to the rows the spec created **strengthens** it. Worth contrasting with #3315 landing alongside, which fixes a different flake by loosening — this one narrows the population instead, which is the direction that keeps the test meaningful.

Adding `sourceConnectionId: SOURCE` to a `findMany` whose assertions are `toEqual([...specific ids])` costs nothing and removes a whole class of interference. Approving on that basis.

### 🟡 IMPORTANT — the stated cause doesn't hold up, and a second candidate is left unaddressed

The comment says:

> *an unfiltered `findMany` reads the whole `order_records` table, and this repository is shared with every other int-spec file running in the same Jest worker. A leaked row from another file … would otherwise land in an unpredictable position*

I went looking for that leaking file and couldn't find one:

- This spec itself has `afterEach(resetTestHarness())` (line 36).
- `order_records` **is** in `tablesToTruncate` (`setup.ts:109`), so the reset clears it between every test.
- The only two api int-specs that touch order tables *without* calling `resetTestHarness` are `fulfillment-work-migration-parity` and `shipment-lines-backfill` — and both `CREATE DATABASE` a throwaway of their own (`MIGRATED_DB`) and write only there, dropping it in `afterAll`. Neither writes `order_records` in the shared harness database.

So on the evidence, no api spec leaks `order_records` into this one's database. That doesn't make the change wrong — it makes the recorded diagnosis unverified, and #3316 shouldn't be closed as "understood" on it.

**The candidate the scoping does not remove:** `applySort` tie-breaks on `rec.createdAt DESC` and nothing further —

```ts
qb.orderBy(OrderRecordRepository.PAYMENT_EXPR, d('ASC'), 'NULLS LAST')
  .addOrderBy('rec.createdAt', 'DESC');
```

— with no final `internalOrderId`. Two rows sharing both the sort key **and** `createdAt` have undefined relative order, and several of these tests seed rows back-to-back where that is possible (the payment case has two rows whose `paymentStatus` is null; `NULLS LAST` says nothing about their order *among themselves*). Scoping by `sourceConnectionId` shrinks the candidate set but does not make a tie deterministic, so if that is the real mechanism the flake returns at lower frequency rather than going away.

Cheap to rule out, and worth doing before closing #3316:

- add `.addOrderBy('rec.internalOrderId', 'ASC')` as the final tie-break in `applySort` (deterministic for every caller, not just this spec — and `findManyRows` / the orders list get the same benefit), **or**
- seed distinct `createdAt` values in the spec so no tie can exist.

The first is the better fix; this repo already applies exactly that reasoning at `line 231` (*"Deterministic tiebreak so two markets with equal counts do not swap"*) and at `1075` / `1714`.

### Notes

- Comment-once-then-reference (`see the note on the first `it` in this file`) rather than repeating the paragraph seven times — good.
- CI green across all nine.

**Verdict:** ✅ Approve. The change is a strict improvement; please just don't treat #3316 as diagnosed until the tie-break question is settled.

norbert-kulus-blockydevs:
Addressed piotrswierzy's review:

- IMPORTANT (tie-break not settled): fixed. Added a final `internalOrderId ASC` tiebreak to every branch of `applySort` (including the no-sort default), after the existing `createdAt DESC` tiebreak - matches the precedent already in this file (routing-country tiebreak, sales-analytics cursor, `findDispatchDeadlineCandidates`). This makes list ordering fully deterministic even when two rows share both the sort key and `createdAt`.
- IMPORTANT (stated leak cause unverified): fixed. Reworded the int-spec comment to stop asserting an unverified cross-file leak mechanism; it now states the `sourceConnectionId` scoping as a strict improvement regardless of mechanism and points at the new tiebreak as the other half of the fix for #3316.
- Comment-once-then-reference pattern / CI green: no action needed, both were called out as already good.

Pushed as 6fd7d64eb.

Issue #3195, opened 2026-09-11
feat(sales-documents): a connection may issue both document kinds


norbert-kulus-blockydevs:
## Merge strategy — 🔗 PR STACK, one merge to `main`

This issue is deliberately large and **must not produce an intermediate state on `main`** — its own body says why under `DO NOT SPLIT`. It is also too large for a single review.

Use an **integration branch** `3195-<short-kebab>` off `main`, then a **stack of PRs targeting that branch** (one per reviewable chunk, each branched from the previous). Review them individually; merge the integration branch to `main` **once**, when the whole is green.

**Never merge a stack member straight to `main`.**

---

Part of #3173.

## Context
`readSalesDocumentRouting` reads `documentKind` as a **single string** and `SalesDocumentRoutingCandidate.documentKind` is single-valued — so `Both` is **not expressible in the shipped model**. This is the real gate of the eparagony work, and it is not front-end work.

## Scope — DO NOT SPLIT
- Widen the config key, the coercer, `SalesDocumentRoutingCandidate`, `resolveSalesDocumentRouting`, `AutoIssueTriggerService` and the settings view — six sites
- Role select offering `Receipts · Invoices · Both · Nothing`; a connection with one capability still renders it, fixed
- **Auto-issue regression tests** — that path decides, unattended, which document a paid order gets. Not optional
- Acceptance scenario: both roles on one connection, one document of each kind on two orders, then attempt the other kind on each — **both refused by `one-document-guard`**

## Verified against the live tree

Two corrections, both found by grepping `main` rather than reading the design.

### The one-document guard already does this - the acceptance scenario is regression coverage

#3184 made the CROSS-kind check connection-blind, and the reason recorded at
`libs/core/src/invoicing/application/services/invoice.service.ts:540-555` names this exact
case: *"since ADR-073 decision 3 a single connection may hold BOTH the invoicing and
fiscalization roles... Exempting 'the requested connection' here - the way the SAME-kind
check does - would let exactly such a dual-role connection register both document kinds
for one order, one from each role."*

`assertNoBlockingFiscalReceipt` therefore refuses on **any** record with
`blocksFurtherRegistration`, with no requested-connection exemption, and
`FiscalRegistrationService.assertNotAlreadyRegistered` is the mirror image. Both run under
the same per-order `invoiceIssueLockKey(orderId)`, which is order-scoped and deliberately
not per `(order, connection)`.

So this issue changes **no guard code**. The acceptance scenario below is a regression
[...]

Issue #3191, opened 2026-09-11
feat(sales-documents): dry-run a rule against a sample order


norbert-kulus-blockydevs:
## Merge strategy — ✅ DIRECT TO `main`

One branch off `main`, one PR, one merge. This slice leaves `main` observably better on its own — no dead code, no half-rendered feature. If it has a dependency it is named at the bottom of this issue; wait for that to land on `main` first, then branch normally.

---

Part of #3173.

## Context
An operator who can ask *"what would this order get"* does not need boolean logic to understand coverage. This is the reason the composer stays AND-only.

## Scope
- A dry-run endpoint answering the question and **saving nothing**
- "Test with a sample order" in the composer

## Acceptance
- [ ] `data-testid="rule-test-sample-order"` returns an answer and persists nothing
- [ ] Reachable before the rule is saved

**Depends on**: inline amounts. **Spec**: `docs/plans/mockups/` (committed by #3174).

PR #3320, opened 2026-09-21
feat(web): clearance filters and awaiting-submission chip on the invoice list

### LINKED ISSUE #3194: feat(web): clearance filters and awaiting-submission chip on the invoice list

---

Part of #3173.

## Context
Invoices waiting on the authority are invisible. The order list already carries a chip stating a count **and** an age; the invoice list needs the same. Note the current filter offers four of six clearance values **deliberately** — `not-applicable` and `cleared` are explicitly filtered out — so restoring them reverses a decision, and the PR should say so.

## Scope
- Restore the two excluded values
- A chip stating count and age of the oldest
- "With provider for" as an **elapsed measurement**, never an estimate of completion — OpenLinker has no basis for one

## Acceptance
- [ ] `invoices-chip-awaiting` states both a count and an age
- [ ] No label on this surface names a regulator

**Depends on**: the Invoicing adapter, **and** the KSeF-label relocation — this renders those labels on more rows. **Spec**: `docs/plans/mockups/` (committed by #3174).

norbert-kulus-blockydevs:
## Summary

Closes #3194.

- The regulatory-status filter on `/invoices` deliberately excluded `not-applicable` and `cleared` from its option list. With eparagony's `RegulatoryStatusReader` mapping real provider answers onto both values, they are reachable, triage-worthy states — all six `RegulatoryStatusValues` are now selectable.
- Adds an "awaiting submission" chip reporting the count and elapsed age of invoices currently `pending-submission`. The age is a wall-clock elapsed measurement off the oldest `updatedAt` among the currently loaded rows, never an ETA.

Part of epic #3173.

## Test plan

- [x] `pnpm --filter @openlinker/web lint` — 0 errors (pre-existing baseline warnings unrelated to this change)
- [x] Includes its own component test (`invoices-list-page.test.tsx`)


piotrswierzy (changes requested):
## Review — `feat(web): clearance filters and awaiting-submission chip on the invoice list`

### Scope note first

The title and body describe one change (`Closes #3194`), but the branch carries three commits: the invoice-list work, then squash-merges of **#3321** (`Closes #3195`, dual-role connections) and **#3323** (`Closes #3191`, rule dry-run), both of which were based on this branch and merged down into it. 39 files, 2 088 additions, three issues closed on merge — two of them via keywords that live only in commit messages.

That is a legitimate stack-collapse and I've reviewed all three features accordingly. Worth restating the body so the merge commit says what it lands.

---

### Summary

Three separate pieces of the #3173 epic. The dual-role work is the strongest: `expandSalesDocumentRoutingCandidates` is a genuinely good seam — one function turning a connection into candidate row(s), called by both the auto-issue gate and the ADR-065 read projection, so the two structurally cannot expand a `'both'` connection differently, with the resolver left untouched because it never had a one-row-per-connection assumption in the first place. The `ConcreteDocumentKind = Exclude<SalesDocumentKind, 'both'>` narrowing of four persisted shapes is exactly right: the config sentinel must not be able to reach a saved rule or country default. The dry run persists nothing, reuses the same four reads and the same pure evaluator every real order goes through, folds the candidate only into the scope it would actually occupy, and its controller documents *why* it skips the capability guard rather than just skipping it.

Two things block, both of the same class: a surface stating something about the operator's data that is measurably false. Given ADR-041 §54 and #2100's treatment of exactly that failure ("a design draft that inferred it printed `no rule for PL` over an entirely different persisted cause, which is a false statement about the operator's own configuration"), I'd rather fix them here than ship them.

---

### Findings

**[BLOCKING]** — `apps/web/src/pages/invoicing/invoices-list-page.tsx` — the chip's age measures the resubmit sweep's cadence, not the wait

`pendingSubmissionOldestAt` reduces over `r.updatedAt`, and the chip renders it as `· oldest 3 d` under the label "awaiting submission". For this exact row population, `updatedAt` is not when the document started waiting — it is when the offline-resubmit sweep last touched it:

- `InvoiceRecordRepository.claimPendingSubmission` (`invoice-record.repository.ts:283`) is a `.createQueryBuilder().update(...).set({ leaseExpiresAt })` gated on `regulatoryStatus = 'pending-submission'`. In the pinned `typeorm@0.3.17`, `UpdateQueryBuilder` appends the `@UpdateDateColumn` to the SET clause unconditionally (`UpdateQueryBuilder.js:401`), so every claim moves `updatedAt`.
- The ORM entity says so itself: `IDX_invoice_records_pending_submission` is a partial index on `['connectionId','updatedAt','id']`, and its comment describes the retransmission scan as ordered `updatedAt ASC, id ASC`. `updatedAt` **is** the sweep's oldest-first progress marker on `pending-submission` rows, by design — a claimed row moves to the back of its own queue.

So the sign is inverted where it matters: the documents that have waited longest are the ones the sweep works hardest, so they look freshest. A document stuck with the regulator for a month renders *no suffix at all* (`< 1 h` returns `''`) because the sweep claimed it an hour ago, and when a suffix does render it reports the lease gap rather than the wait. The commit message is careful that this is elapsed-not-ETA, which is right — but it is elapsed *since the wrong event*.

`InvoiceListItem` already carries `issuedAt: string | null` and `createdAt: string` (`invoicing.types.ts:110-112`), neither of which the sweep touches. `issuedAt ?? createdAt` is a one-line fix and is monotone for this population.

**[BLOCKING]** — `apps/api/.../sales-document-dry-run-result-response.dto.ts` + `apps/web/.../describe-sales-document-dry-run-result.ts` — a country-default decision is reported as an already-saved rule

`SalesDocumentDryRunResultDto.fromDecision` collapses `decision.ruleId` into one boolean and drops the id. But `ruleId` is three-valued in meaning, and `sales-document-decision.types.ts:11-16` states the constraint explicitly:

> `ruleId` … is present for a tier-1 rule match, and **absent for every other route (a tier-2 country default, the legacy `resolveSalesDocumentRouting` fallback)**, which is a legitimate (kind, connection) pair with no rule behind it, and **`ruleId` must not imply one exists**.

`evaluate-sales-document-rules.ts:190` confirms it: `ruleId` is set only on the tier-1 arm. So a decision reached by the country default arrives with `matchedByCandidateRule: false` and no id, and `describeSalesDocumentDryRunResult`'s else-arm renders *"via an already-saved rule at Acme."* — the one thing the type says must not be implied. The DTO's own `@ApiProperty` description hard-codes the false dichotomy ("as opposed to an already-saved rule elsewhere in this configuration").

This matters more than a wording slip because of where it renders: the operator is in the rule composer deciding whether their draft rule is *needed*. Told "an already-saved rule decided this", they go looking for a rule to edit that does not exist; told "the country default for PL decided this", they know exactly what they are competing with. The information is discarded at the boundary, so no FE fix is possible — carry `ruleId`, or a three-valued `decidedBy: 'candidate' | 'saved-rule' | 'country-default'`, and branch three ways.

**[IMPORTANT]** — `invoices-list-page.tsx` — `cleared` is re-added on a premise that still holds

Splitting the two values the commit reverses:

- **`not-applicable` — justified.** `eparagony-invoice.mapper.ts:530` genuinely returns it. The old comment ("noise") is now wrong and removing the exclusion is right.
- **`cleared` — not justified.** The old comment said "a reserved status no provider emits", and that is still exactly true. Repo-wide, nothing writes it: the eparagony mapper emits `not-applicable | rejected | accepted | pending-submission` (lines 530-543) and never `cleared`; neither the KSeF nor the inFakt adapter emits it; the only non-test mentions in the tree are the `RegulatoryStatusValues` declaration itself and `subiekt-regulatory-status.mapper.ts:14`, whose comment reads "(`'cleared'` is reserved/unused for Subiekt.)".

So the commit message's justification — "with eparagony's `RegulatoryStatusReader` mapping real provider answers onto both values" — holds for one of the two. Selecting `cleared` filters to a state nothing can be in, and the resulting empty list reads as "you have no cleared documents": a positive claim about the operator's data derived from an unreachable status. Keep it excluded until something writes it, or make something write it first.

**[IMPORTANT]** — `expand-sales-document-routing-candidates.ts` + `resolve-sales-document-routing.spec.ts` — a `'both'` connection can never be resolved by the fallback resolver, and the spec says otherwise

The expansion copies one `isPrimary` onto **both** rows. Against `resolveSalesDocumentRoutingFromCandidates`' tie-break (`primaries.length === 1 ? primaries[0] : undefined`), that gives a lone dual-role connection two outcomes and they are the same one:

| `config.salesDocument.isPrimary` | rows | primaries | result |
|---|---|---|---|
| `false` | 2 | 0 | `unresolved / ambiguous-connection-no-primary` |
| `true` | 2 | 2 | `unresolved / ambiguous-connection-no-primary` |

That may well be intended — a dual-role connection is *meant* to be disambiguated by the rule engine, which runs first (#2173) and only falls through here on `no-configuration-for-country`. But two consequences are unstated:

1. **The spec's only successful dual-role case is unreachable.** `'should route to the primary row when exactly one of two same-connection rows is primary'` hand-builds `isPrimary: false` + `isPrimary: true` on one `connectionId` — a pair `expandSalesDocumentRoutingCandidates` cannot produce. It is the single most reassuring test in the new block and it certifies a disambiguation path that does not exist. This is the shape `docs/lessons.md` already records from #2380 ("its unit test passed because it hand-built a combination the receipt transition cannot produce"). Driving those rows through the expansion instead would fail, and force the real behaviour to be written down.
2. **The default path for a fresh install is a blocked one.** Ticking "Both" on the only connection of an install with no rules and no country default for the order's country flips every order to `ambiguous-connection-no-primary`, whose persisted operator detail (per ADR-041 §54 / #2100) is worded about competing *invoicing connections* — plural — when the operator has one. Either say so where "Both" is chosen, or give the unresolved vocabulary a member that names the real cause.

**[IMPORTANT]** — the chip's count is page-scoped, and clicking it changes the number it displayed

The comment is honest that there is no summary read, but the rendered artifact is not: `Awaiting submission 12 · oldest 3 d` is visually identical to the orders list's `salesDocumentBlocked` chip, which *is* backend-aggregated install-wide. Capped at `PAGE_SIZE`, it understates — the direction that matters for a triage affordance. And the tell is built in: click the chip, the list refetches filtered, and `pendingSubmissionItems.length` becomes `min(total, PAGE_SIZE)` — so a chip reading "3" can read "20" as a consequence of being clicked.

Since #2944 the repo has the shape for this: `?withTotal` plus `GET .../count` siblings, so a cheap count-only read scoped to `regulatoryStatus=pending-submission` is an established pattern rather than a new one. Failing that, label it as page-scoped.

**[SUGGESTION]** — `pendingSubmissionAgeSuffix` is a byte-identical copy of `blockedAgeSuffix` (`orders-list-page.tsx:366`), same thresholds, same `· oldest N d` string, same `''` fallback. The stated reason — "that page's helper is private and unexported" — is a reason to export it. Two copies of one operator-facing string in two lists is the drift the mirror-guard scripts exist to catch elsewhere; this one is cheap to avoid entirely by moving it to `shared/lib`.

**[SUGGESTION]** — `SalesDocumentRulesService.dryRunRule` evaluates the candidate with `effectiveFrom: new Date(0)`, `effectiveTo: null`. Well reasoned and documented on the type ("an operator authoring a rule is testing its CONDITIONS, not the calendar") — but the composer collects an effective window on the same dialog, so the on-screen answer can differ from what the saved rule would do without saying so. One line of copy beside the result closes it.

**[SUGGESTION]** — `SALES_DOCUMENT_KIND_BOTH = 'both'` carves a reserved word out of an open-world namespace. ADR-041 decision 10 lets a regime declare a kind core has never seen; a regime declaring one literally named `both` now silently expands into two rows. Remote, but the sentinel's docblock is the place to say it is reserved.

---

### Positive observations

**One expansion point, shared by the gate and the read surface.** Both `AutoIssueTriggerService` and `SalesDocumentViewService` go from `.map` to `.flatMap` through the same function, so the projection an operator reads and the decision the gate makes cannot disagree about a dual-role connection. That is the property, and building it as one pure function with its own spec rather than two parallel `if` branches is the right shape.

**The resolver is genuinely untouched.** The change to `resolve-sales-document-routing.ts` is comments only — because the resolver already operated purely over the candidate list and had no connection-identity assumption to break. Recognising that and resisting the urge to add dual-role handling *there* is what keeps decision 3a enforced over the set rather than inside an ambiguous row.

**`ConcreteDocumentKind`.** Widening `SalesDocumentKind` for the config axis and then narrowing the four shapes that can never hold `'both'` — saved rule, rule input, country default, upsert input — is the discipline that stops a config sentinel reaching a persisted row. The follow-up commit explains the `.filter` / `.flatMap` narrowing gap it found doing it.

**The dry run is a read, provably.** Same four repository reads as `resolveRouting`, same pure evaluator, same country normalisation (#3176), same `assertConditionsWellFormed` guard on raw conditions, and the candidate folded into one scope rather than both. The controller stating why `SalesDocumentCapabilityGuardService` is skipped — "that guard exists to keep a SAVED rule from pointing at a connection that cannot issue its document kind, and a dry run persists nothing for it to protect" — is the reasoning written at the point of temptation.

---

### Final assessment

**Merge readiness: ❌ Request changes**

The architecture is sound in all three pieces and nothing here crosses a boundary or misses a migration. What blocks is narrower and, I think, worth holding for: two operator-facing surfaces state things that are not true — an age derived from a column a background sweep resets on exactly the rows being measured, and a decision provenance the domain type explicitly forbids inferring. Both are small fixes at the right layer.

Priority fixes:

1. Age the chip off `issuedAt ?? createdAt`, not `updatedAt` — `claimPendingSubmission` bumps `updatedAt` on every `pending-submission` row it claims.
2. Carry `ruleId` (or a three-valued `decidedBy`) through the dry-run DTO, and branch three ways in `describeSalesDocumentDryRunResult` so a country default is not reported as a saved rule.
3. Keep `cleared` out of the regulatory filter — nothing in the tree writes it; re-add `not-applicable` alone.
4. Drive the dual-role spec's rows through `expandSalesDocumentRoutingCandidates`, and state what happens when "Both" is chosen with no rule to disambiguate it.
5. Either count the chip from a real total or label it as page-scoped.

norbert-kulus-blockydevs:
Addressed all findings from the review.

**[BLOCKING] Chip age measures the sweep's cadence, not the wait** - fixed. `pendingSubmissionOldestAt` now uses `issuedAt ?? createdAt`; updated the docblock and the test to assert on `issuedAt` (with `updatedAt` deliberately left at an unrelated timestamp so the test fails if the code regresses to keying on it).

**[BLOCKING] Country-default reported as an already-saved rule** - fixed. `matchedByCandidateRule: boolean` replaced with a three-valued `decidedBy: 'candidate' | 'saved-rule' | 'country-default'` on the DTO, the FE type, and `describeSalesDocumentDryRunResult` (now branches three ways). Updated the controller spec, the describe-function spec, the composer dialog test, and `test-utils.tsx`'s mock.

**[IMPORTANT] `cleared` re-added on a false premise** - fixed. The filter now excludes `cleared` (nothing in the tree writes it) while keeping `not-applicable` selectable. Updated the corresponding test.

**[IMPORTANT] Dual-role spec's rows are unreachable** - fixed. Replaced the two tests that hand-built an `isPrimary: false`/`isPrimary: true` pair on one `connectionId` (impossible via `expandSalesDocumentRoutingCandidates`, which copies one `isPrimary` onto both rows) with tests driven through the real expansion function. They now assert the real outcome: a lone dual-role connection - marked primary or not - always resolves `ambiguous-connection-no-primary`, and a single-role primary sibling can't rescue it either. Documented this explicitly in both `expand-sales-document-routing-candidates.ts`'s and `sales-document-kind.types.ts`'s docblocks (the rule engine, not `isPrimary`, is what disambiguates a `'both'` connection).

**[IMPORTANT] Chip count reads as install-wide but is page-scoped** - fixed via the documented fallback: added a `title` tooltip stating the count is page-scoped, without changing the visible label (keeps existing tests intact). A backend count-only read (the #2944 `?withTotal`/`count` pattern) would be a larger, separately-reviewable change.

**[SUGGESTION] Duplicate age-suffix helper** - fixed. Extracted to `shared/lib/oldest-age-suffix.ts` with its own unit test; both the orders list's `salesDocumentBlocked` chip and the invoices list's "Awaiting submission" chip now use it.

**[SUGGESTION] Dry-run window vs. saved rule's effective window** - fixed. Added a one-line note beside the dry-run result stating the test ignores the effective window and checks conditions only.

**[SUGGESTION] `SALES_DOCUMENT_KIND_BOTH` reserved word** - fixed. Documented in the constant's docblock that `'both'` is claimed by the sentinel and must never be reused as a genuine per-regime kind.

Pushed to the same branch.

norbert-kulus-blockydevs:
@piotrswierzy - confirming everything from the review has landed (commit `2d96e21`, already pushed) and re-verifying the gate end to end:

**[BLOCKING] Chip age measured the sweep's cadence, not the wait** - fixed. `pendingSubmissionOldestAt` now derives from `issuedAt ?? createdAt`, never `updatedAt`; the docblock on `pendingSubmissionAgeSuffix` spells out why (`claimPendingSubmission`'s `.update()` bumps `updatedAt` on every claimed row).

**[BLOCKING] Country-default route reported as an already-saved rule** - fixed. `matchedByCandidateRule: boolean` replaced by a three-valued `decidedBy: 'candidate' | 'saved-rule' | 'country-default'` end to end (API DTO -> FE type -> `describeSalesDocumentDryRunResult`, now branching three ways instead of two).

**[IMPORTANT] `cleared` re-added on a premise that didn't hold** - fixed. Filter keeps `not-applicable` (eparagony genuinely emits it) and excludes `cleared` (nothing in the tree writes it - confirmed the eparagony/KSeF/inFakt/Subiekt mappers again while reviewing).

**[IMPORTANT] Dual-role spec certified an unreachable disambiguation path** - fixed. The two tests that hand-built `isPrimary: false`/`true` on one `connectionId` are replaced with tests driven through the real `expandSalesDocumentRoutingCandidates`, asserting the actual behaviour: a lone `'both'` connection - primary or not - always resolves `ambiguous-connection-no-primary`. Documented in both `expand-sales-document-routing-candidates.ts` and `sales-document-kind.types.ts`.

**[IMPORTANT] Chip count is page-scoped but reads as install-wide** - addressed via the stated fallback rather than a backend aggregate: a `title` tooltip now states the count is page-scoped. Agreed a real `?withTotal`/count-only read (the #2944 pattern) is a separately-reviewable change, not folded into this one.

**[SUGGESTION] Duplicated age-suffix formatter** - fixed. Extracted to `shared/lib/oldest-age-suffix.ts` with its own test; both the orders list's `salesDocumentBlocked` chip and the invoices list's "Awaiting submission" chip now share it.

**[SUGGESTION] Dry-run ignores the effective window silently** - fixed. One line of copy beside the result states the test checks conditions only.

**[SUGGESTION] `SALES_DOCUMENT_KIND_BOTH` reserved-word risk** - fixed. Documented on the constant that `'both'` is claimed by the sentinel and must never be reused as a genuine per-regime `documentKind`.

Re-ran the full gate on this state:
- `pnpm --filter @openlinker/web type-check` - clean
- `pnpm --filter @openlinker/api type-check` - clean
- `pnpm --filter @openlinker/web test` (full suite, 576 files) - 6218 passed, 1 skipped; one `price-changes-queue-table.test.tsx` PostHog-script-loading flake reproduced as green when run in isolation, unrelated to this diff
- Targeted specs: `expand-sales-document-routing-candidates.spec.ts`, `resolve-sales-document-routing.spec.ts` (20/20, incl. the new dual-role cases), `sales-document-rules.controller.dry-run.spec.ts` (7/7, incl. the new `decidedBy` cases) - all green
- `pnpm check:invariants` - all green, including `check-ui-vocabulary` over the two new copy strings
- `pnpm --filter @openlinker/web lint` / `pnpm --filter @openlinker/api lint` - 0 errors (pre-existing warning baseline only)

Nothing further to push - HEAD already matches `origin/3194-invoice-list-clearance`.

piotrswierzy (approved):

Re-read at `c5a071f34`. Both blocking findings addressed on this branch, matching the fixes on `3173-epic-integration`:

- **Chip age** now `r.issuedAt ?? r.createdAt` (line 521) — neither column is touched by `claimPendingSubmission`, so the figure measures the wait rather than the resubmit sweep's own cadence.
- **`cleared`** filtered out of the regulatory options (line 638), leaving `not-applicable`, which eparagony genuinely emits.
- **Dry-run provenance** carried three-valued through the DTO, so a country-default decision is no longer reported as an already-saved rule.

The dual-role spec and the sample-order pricing state are corrected too.

One thing from the original review that is unchanged and worth restating for the merge commit rather than the code: this branch carries squash-merges of **#3321** (`Closes #3195`) and **#3323** (`Closes #3191`) on top of its own #3194 work, so merging it closes three issues under a title naming one, with two of the keywords living only in commit messages. Worth a line in the body — or, if `3173-epic-integration` is the intended merge point, saying so here.

**✅ Approve.**

piotrswierzy (approved):
## Re-review at `931d23db6` — merge-forward only; my approve stands

Two commits since `c5a071f34`: a merge of `main` (which had moved by ten commits, including #3327's `DataTable` migration of the price-changes queue, #3326's stock-location override field and #3332's eparagony corrective invoices) and one fix for its fallout.

Both blocking findings remain fixed across the merge — I re-read them rather than assuming:

- **Chip age** still `issuedAt ?? createdAt`, so it measures the wait rather than `claimPendingSubmission`'s own cadence.
- **`cleared`** still excluded from the regulatory-status options; `not-applicable` kept.
- **Dry-run provenance** still three-valued, so a tier-2 country default is not reported as an already-saved rule.

`931d23db6` is the honest kind of merge fallout — the same `data-testid` declared twice on one `Select`, which is `TS17001` and would have failed `tsc -b`, caught and removed with the cause named. Worth noting that it was a *type error* that surfaced it: a duplicate `data-testid` is exactly the class of conflict resolution that can survive a JSX merge looking plausible, and here the compiler was the thing that objected rather than a reviewer.

The merged-in surface does not touch this page — #3327's `DataTable` work is on the price-changes queue and #3326's on the connection form, so neither overlaps `invoices-list-page.tsx`.

---

One thing from my original review is still open and is about the merge commit rather than the code: this branch carries squash-merges of **#3321** (`Closes #3195`) and **#3323** (`Closes #3191`) on top of its own #3194 work, so merging it closes three issues under a title naming one, with two of the keywords living only in commit messages. A line in the body — or a note that `3173-epic-integration` is the intended merge point — would make the merge commit say what it lands.

**✅ Approve.**

Issue #3221, opened 2026-09-13
[BUG] Frontend — a packer sees Operations nav entries whose API reads 403


norbert-kulus-blockydevs:
## Problem / Context

A `packer` signing in sees the whole ungated **Operations** nav group, and roughly half of it answers 403 on the first request each page makes.

`packer` (ADR-071 / #2413) holds an intentionally EMPTY `ROLE_PERMISSIONS` grant — its access is enforced route-by-route by `@Roles(...)` and nothing else. It appears in `@Roles(...)` on exactly **three controllers API-wide**, all of them bench controllers (8 `@Roles` lines total):

- `apps/api/src/bench/http/bench-work.controller.ts`
- `apps/api/src/bench/http/bench-parcel.controller.ts`
- `apps/api/src/bench/http/bench-documents.controller.ts`

Meanwhile `BASE_NAV_GROUPS`' `Operations` group (`apps/web/src/app/nav-registry.ts`) carries no group-level gate, so every item in it renders for a packer.

### Measured, not assumed

Checked the guard on each item's primary read:

| Nav item | Guard on its primary read | Packer |
|---|---|---|
| Analytics | `@Roles('admin','operator','viewer')` | **403** |
| Insights | `@Roles('admin','operator','viewer')` | **403** |
| Orders | `@Roles('admin','operator','viewer')` | **403** |
| Customers | `@Roles('admin','operator','viewer')` | **403** |
| Fulfilment | `@Roles('admin','operator','viewer')` | **403** |
| Products | `@AnyRole()` | loads |
| Listings | `@AnyRole()` | loads |
| Shipments | `@AnyRole()` | loads |
| Returns | `@AnyRole()` | loads |
| Invoices | mixed (`@AnyRole()` on some reads) | partial |
| Automations | `requiresPermission: 'automations:read'` | already hidden |
| **Pack bench** | `@Roles('admin','operator','packer')` | **loads** |

**This corrects the framing the finding was raised with.** PR #3204's review says "every one of them 403s"; that is not what the code does — `@AnyRole()` admits `packer`, so five entries load fine and `Automations` is already correctly hidden by the existing permission gate. **Five entries 403 outright**, plus `Invoices` partially. That changes the remedy: this is a per-item gate on a named subset, not a case for excluding `packer` from the group wholesale.

[...]

PR #3322, opened 2026-09-21
fix(web): gate packer-403ing Operations nav entries behind requiresRole

norbert-kulus-blockydevs:
## Summary

A `packer` sees the whole ungated **Operations** nav group, and roughly half of it answers 403 on the first request each page makes. `packer` (ADR-071 / #2413) carries a deliberately empty `ROLE_PERMISSIONS` grant on the backend — its access is enforced purely route-by-route via `@Roles(...)`.

This applies the item-level `requiresRole` gate #3108 built for "Pack bench" to the six `Operations` entries whose primary API read excludes `packer` (all verified against the controller's `@Roles(...)` decorator, not assumed):

| Nav item | Primary read | Guard |
|---|---|---|
| Analytics | `GET /analytics/sales` et al. | `@Roles('admin','operator','viewer')` |
| Insights | `GET /connections` | `@Roles('admin','operator','viewer')` |
| Orders | `GET /orders` | `@Roles('admin','operator','viewer')` |
| Customers | `GET /customers` | `@Roles('admin','operator','viewer')` |
| Fulfilment | `GET /fulfillment-work` | `@Roles('admin','operator','viewer')` |
| Invoices | `GET /invoices` | `@Roles('admin','operator','viewer')` (checked: every route on the page carries the same guard, so this is not a partial fix) |

Products, Listings, Shipments and Returns are untouched — their primary reads are `@AnyRole()`, and a packer can legitimately open them today. Automations was already correctly hidden via the pre-existing permission gate.

- `nav-registry.types.ts`: widens `RoleValues` with `'viewer'` (the prerequisite #3108's docblock already named this as the known blocker).
- `nav-registry.ts`: adds `requiresRole: ['admin', 'operator', 'viewer']` to the six items above.
- `nav-registry.test.ts` / `command-palette-provider.test.tsx`: a packer session sees exactly Products / Listings / Shipments / Returns / Pack bench (named explicitly, not as a count), asserted on **both** the sidebar and ⌘K — both read the shared `isNavItemVisible`, and the issue calls out a sidebar-only test as leaving the documented bypass open. admin/operator/viewer sessions retain the full item set.

No `@Roles(...)` decorator in `apps/api` is changed — the nav gate hides an affordance and grants nothing; the API remains the authorization boundary.

## Base branch

Targets `3104-oms-onboarding-ui` (the epic branch), not `main` — this issue's two dependencies (#3107 widening `RoleValues` with `packer`, #3108 the `requiresRole` mechanism) are merged into that branch but not yet into `main`.

## Test plan

- [x] `pnpm --filter @openlinker/web lint` — 0 errors (pre-existing warnings only, none introduced)
- [x] `vitest run src/app/nav-registry.test.ts src/app/command-palette-provider.test.tsx` — 49/49 passing
- [x] Verified each of the six controllers' `@Roles(...)` decorator directly in source rather than trusting the issue's table

Closes #3221


norbert-kulus-blockydevs:
## Tech-lead review

**Context check**: Base branch is `3104-oms-onboarding-ui`, not `main` — correct per #3221's stated dependency on #3107 (RoleValues widening mechanism) and #3108 (`LiveNavItem.requiresRole` + `isNavItemVisible`), both already on the epic branch. Diff is scoped exactly to the issue: 4 files, no unrelated changes.

### Correctness — verified against source, not trusted from the PR description
I independently checked every one of the six `@Roles(...)` claims against the actual controllers on `3104-oms-onboarding-ui`:

| Nav item | Controller | Claimed guard | Actual guard |
|---|---|---|---|
| Analytics | `sales-analytics.controller.ts` | `@Roles('admin','operator','viewer')` | ✅ confirmed |
| Insights | `connection.controller.ts` (`GET /connections`) | same | ✅ confirmed |
| Orders | `orders.controller.ts` | same | ✅ confirmed |
| Customers | `customers.controller.ts` | same | ✅ confirmed |
| Fulfilment | `fulfillment-work.controller.ts` | same | ✅ confirmed |
| Invoices | `invoicing.controller.ts` | same | ✅ confirmed |

All six check out exactly. No fabricated guard claims.

### Architecture & mechanism
- ✅ Reuses the exact `LiveNavItem.requiresRole` gate #3108 already built and used for "Pack bench" — no new mechanism invented, no duplicated visibility logic.
- ✅ `RoleValues` widening with `'viewer'` was explicitly pre-planned in the #3108-era docblock ("There is already a known consumer waiting") — this PR is that named consumer, not an unplanned scope creep.
- ✅ Both consumers of `isNavItemVisible` (sidebar via `nav-registry.test.ts`, ⌘K via `command-palette-provider.test.tsx`) get coverage — correctly avoids the "sidebar-only test leaves the documented bypass open" trap the issue explicitly calls out.
- ✅ No `@Roles(...)` decorator touched in `apps/api` — correctly scoped as a UI-affordance fix, not an authorization change. The nav gate hides, the API still enforces.
- ✅ The five `@AnyRole()` entries (Products, Listings, Shipments, Returns) and the already permission-gated Automations are correctly left untouched — matches the issue's own measured table, not the PR review's original mis-framing ("every one 403s") that #3221 itself corrected.

### Tests
- Named-explicitly assertions (`PACKER_VISIBLE_OPERATIONS_ITEMS` / `PACKER_HIDDEN_OPERATIONS_ITEMS`) rather than a count — matches the issue's own stated acceptance criterion reasoning (a count would pass with the wrong five items visible).
- Non-regression coverage for admin/operator/viewer on every gated item, and for the untouched `@AnyRole()` items across all four roles.
- `it.each` used well — no copy-pasted near-duplicate test bodies for the per-role sweep.

### Minor / non-blocking
- `Invoices` was listed as "partial" in the original issue (mixed `@AnyRole()`/`@Roles('admin')` across the invoicing controllers) — this PR resolves that ambiguity by checking the controller directly and confirming the *primary* read (`GET /invoices`) is fully role-gated, with a comment explaining "every other read on the page carries the same guard, so this is not a partial fix." That's exactly the kind of verification the issue's own Assumptions section asked for. Good.
- No migration/schema concerns — pure frontend nav-visibility change, as expected.

**Verdict**: ✅ Approve — every factual claim underlying the fix was independently verified against the actual controller guards and checks out, the mechanism is reused rather than reinvented, and both nav-visibility consumers (sidebar + ⌘K) are covered. No blocking or important issues found.

norbert-kulus-blockydevs:
Checked the review above against the code — every finding was ✅ Approve / confirmed, with the two "minor" notes being verification confirmations rather than action items. Nothing actionable, so no code changes were made.

piotrswierzy (approved):
## Review — `fix(web): gate packer-403ing Operations nav entries behind requiresRole`

### Summary

Small, exact, and correct. The nav gate hides an affordance and grants nothing, no `@Roles(...)` decorator moves, and the API stays the authorization boundary — the right shape for this class of fix, and stated as such in the body.

I verified all six claims in the table against the controllers rather than the issue, since the whole change rests on them:

| Nav item | Controller | Guard |
|---|---|---|
| Analytics | `sales-analytics.controller.ts:110` | `@Roles('admin','operator','viewer')` ✓ |
| Insights | `connection.controller.ts:160` (`GET /`) | ✓ |
| Orders | `orders.controller.ts:277` | ✓ |
| Customers | `customers.controller.ts:69` | ✓ |
| Fulfilment | `fulfillment-work.controller.ts:87` | ✓ |
| Invoices | `invoicing.controller.ts:209` | ✓ |

And the four left ungated hold up too: `products.controller.ts` is `@AnyRole()` throughout (9/0), `listings` and `returns` carry `@AnyRole()` on their reads with `@Roles` on writes. So a packer opening Products or Listings gets a working page, which is the whole reason for not gating them.

---

### Positive observations

**The six `requiresRole` literals are repeated rather than extracted, and that is right.** A `NON_PACKER_ROLES` constant would read tidier and would be wrong: each declaration mirrors one specific controller's decorator, and those six happening to be identical today is a coincidence of six independent facts, not a shared rule. Repeating them keeps each gate checkable against its own controller — which is exactly the check the body says was done. Worth keeping when someone inevitably suggests the constant.

**The stale prophecy in the docblock was discharged, not left standing.** The pre-existing text named this exact work ("There is already a known consumer waiting… the per-item remedy is `requiresRole: ['admin', 'operator', 'viewer']`… which this union cannot yet express — tracked against epic #3104") and the change removes it and records what actually happened. A docblock that predicts a fix and then outlives it is how the next reader concludes the fix never landed.

**Both consumers are asserted, by name.** The body notes the issue calls out a sidebar-only test as leaving the documented bypass open, and the tests cover `nav-registry` and `command-palette-provider` — both of which read the shared `isNavItemVisible`. Naming the five expected items explicitly rather than asserting a count is what makes the test fail usefully when a seventh entry is added.

**Each `requiresRole` carries the reason inline**, naming the route and the guard, so a future reader can re-verify the gate without hunting for the controller. That is what made this review cheap.

---

### Findings

**[IMPORTANT]** — `nav-registry.types.ts:63-65` — the docblock's closing sentence is falsified by the line beneath it

```
 * ... this union is
 * scoped to what a nav gate needs to express, not a mirror of the backend's
 * full role vocabulary.
 */
export const RoleValues = ['admin', 'operator', 'packer', 'viewer'] as const;
```

`UserRoleValues` is `['admin', 'operator', 'viewer', 'packer']`. As of this commit the two contain exactly the same four members — the union *is* a full mirror of the backend's role vocabulary, differing only in order. That sentence was true before the change and is false after it, and it is precisely the sentence that tells the next reader they need not keep the two in sync.

That matters because there is no guard. I checked: `scripts/check-permission-mirror.mjs` mirrors `PermissionValues` only, between `role.types.ts` and `apps/web/src/shared/auth/session.types.ts` — `RoleValues` is covered by nothing, and `session.types.ts` carries no role union of its own, so `nav-registry.types.ts` is now the sole frontend copy. The permission mirror exists for exactly this reason, in its own words: *"the two arrays are copies — and a copy drifts silently."*

The drift is at least fail-safe in direction — a role added backend-only is absent from every `requiresRole` allow-list, so its nav entries hide rather than 403 — but "safe" is not "intended", and hiding Orders from a newly-added role with no signal is its own support ticket.

Two options, either fine:
- Add the `RoleValues` ↔ `UserRoleValues` pair to `check-permission-mirror.mjs`. It already parses `role.types.ts`, so this is a second array pair against the same textual parser, and it would catch the ordering difference too.
- Or, if a guard feels heavy for a four-member union, just correct the sentence: say it is now a full copy, that nothing enforces that, and what a reader should do when the backend gains a fifth role.

**[SUGGESTION]** — the test plan runs `@openlinker/web` lint and type-check but not `pnpm check:invariants`

Nothing in this diff trips a guard today (I checked the mirror script's scope above), so this is not a gap in *this* PR. But it is the step that would catch the finding above once the mirror is extended, and `check-permission-mirror` is one of the guards that reads `role.types.ts` — worth adding to the plan for anything touching a role or permission union.

---

### Final assessment

Correct, minimal, verified at the source rather than from the issue, and tested on both surfaces that consume the gate. Nothing blocks.

One priority note: fix or guard the "not a mirror of the backend's full role vocabulary" sentence — it became untrue in this commit, and it is the sentence that would stop someone keeping the two unions aligned.

Base-branch targeting is right and stated (both dependencies are on `3104-oms-onboarding-ui`, not `main`); worth re-running the six decorator checks after that epic branch rebases, since `@AnyRole()` / `@Roles(...)` on those controllers is the only thing this change depends on.

norbert-kulus-blockydevs:

- **[IMPORTANT]** `nav-registry.types.ts:63-65` docblock sentence ("not a mirror of the backend's full role vocabulary") was false as of this commit, since `RoleValues` and `UserRoleValues` now contain the same 4 members. Fixed: rewrote the sentence to state the union IS now a full copy, that nothing enforces alignment (`check-permission-mirror.mjs` covers `PermissionValues` only, not `RoleValues`), and what to do if a fifth role is added.
- **[SUGGESTION]** Test plan missing `pnpm check:invariants` for role/permission-touching PRs - declined for this PR specifically (reviewer confirmed nothing in this diff trips the mirror script today), noted as general guidance for future PRs rather than a change here.

Pushed as e2f5bffa3.

norbert-kulus-blockydevs:
## Re-review (post piotrswierzy's IMPORTANT finding)

Verified commit \`e2f5bffa3\` against piotrswierzy's finding directly (read the diff, not just the fix-commit's own description).

**[IMPORTANT] resolved.** The docblock's false closing sentence ("not a mirror of the backend's full role vocabulary") is replaced with an accurate statement: \`RoleValues\` is now a full copy of \`UserRoleValues\`, nothing enforces alignment, \`check-permission-mirror.mjs\` doesn't cover this union, and — the part that matters most for a future reader — the failure mode if the two drift is stated explicitly (fails **closed**: a role missing from this array just hides the nav item, never a 403). This is exactly piotrswierzy's option 2 ("if a guard feels heavy for a four-member union, just correct the sentence: say it is now a full copy, that nothing enforces that, and what a reader should do"). Comment-only change, no logic touched — zero regression surface.

**[SUGGESTION] on the test plan** — correctly declined for *this* PR (nothing in the diff trips the mirror script today, per piotrswierzy's own check) and left as general forward guidance rather than forced into scope here. Right call, not a gap.

No newer comment from piotrswierzy since the fix landed — nothing outstanding.

**Verdict**: ✅ Approve.

piotrswierzy (approved):
## Re-review — the finding is fixed, and the replacement says the useful part

Re-read at `eb6d3e487`. The false sentence is gone and what replaced it does more than correct the tense — it states all three things a reader needs:

- that `RoleValues` **is** a full copy of `UserRoleValues` as of this change, differing only in declaration order;
- that **nothing enforces it** — naming `check-permission-mirror.mjs` and its actual scope (`PermissionValues`, between `role.types.ts` and `session.types.ts`) so a reader does not assume the guard next door covers this too;
- and what the failure mode is when it drifts: **fails closed**, the nav item stays hidden, never a 403, *"so the failure mode is silent rather than loud."*

That last clause is the one I'd have written if asked, and it is the honest framing — the direction is safe, and silent-and-safe is still a support ticket rather than a non-event. Taking the documentation option over extending the mirror script is a fair call for a four-member union; what mattered was that the sentence which told the next reader *not* to keep the two aligned stopped saying so.

Everything from the original review stands: six `requiresRole` declarations each verified against their own controller's decorator, the six literals deliberately repeated rather than extracted into a constant that would imply a shared rule where there are six independent facts, the stale prophecy in the docblock discharged rather than left to outlive its own fix, and both consumers of `isNavItemVisible` asserted by name rather than by count.

The base-branch note also stands: this targets `3104-oms-onboarding-ui`, so the six decorator checks are worth re-running after that epic rebases — `@AnyRole()` / `@Roles(...)` on those controllers is the only thing this gate depends on.

**✅ Approve.**

piotrswierzy (changes requested):
## The pack-bench stack merged, and it changed this PR's premise

`main` moved `5d470b97e..fbae25c1b` — the fourteen-PR pack-bench stack landed. This PR is **90 commits behind** and now **conflicts**, in all three of the files it exists to change:

```
CONFLICT (content): apps/web/src/app/nav-registry.ts
CONFLICT (content): apps/web/src/app/nav-registry.types.ts
CONFLICT (content): apps/web/src/app/nav-registry.test.ts
```

A rebase is needed regardless. The reason this is a review comment and not a one-line "please rebase" is that **the fact these comments are reasoning from is no longer true**, so resolving the conflict mechanically would carry a false premise onto `main`.

### `ROLE_PERMISSIONS.packer` is no longer `[]`

This PR's comments say, at several items:

> a `packer` (empty `ROLE_PERMISSIONS` grant, #2413) 403s

On `main` today:

```ts
packer: ['bench:write'],
```

#3439 added `bench:write` and granted it to exactly **admin + operator + packer** — with a guard, `scripts/check-bench-write-roles.mjs`, that keeps that grant identical to the `@Roles` lists on `BenchWorkController` and `BenchParcelController`, failing in both directions.

That matters for one item in particular. This PR writes:

```ts
{ to: '/bench', label: 'Pack bench', requiresRole: ['admin', 'operator', 'packer'] },
```

with a comment saying the item admits all three because no permission can. **A permission that names exactly those three now exists**, is guarded against the routes it mirrors, and `docs/frontend-architecture.md` § Access Control And UI Visibility is explicit about which input the frontend should read:

> The session's `permissions[]` array … is the only authorization input the frontend reads. `session.user.role` is typed `string` and **must never be compared inline** — a typo type-checks and silently evaluates false.

So `/bench` should be `requiresPermission: 'bench:write'`. That is one word of config instead of a role array, it inherits the guard that keeps it aligned with the controllers, and it removes the one item where this PR's role-array mechanism was least avoidable.

### `main`'s own comment on that item is stale in the same way — and this is the PR to fix it

Worth knowing before you resolve, because the conflict will show you both sides and one of them is wrong. `main`'s `/bench` entry now reads:

```ts
// Gated on `orders:write`, held by exactly admin + operator, for the same
// reason as the entry above: the bench's own routes are
// `@Roles('admin', 'operator', 'packer')`, and `ROLE_PERMISSIONS.packer`
// is `[]`, so no permission can name all three.
{ to: '/bench', label: 'Pack bench', requiresPermission: 'orders:write' },
```

Two PRs in one stack: one changed `packer` from `[]` to `['bench:write']`, the other's comment still describes the old state. The premise (`is []`) and the conclusion drawn from it (`no permission can name all three`) are both false on `main` right now. The *behaviour* is defensible — the comment says a packer gets no sidebar at all because `/bench` renders outside `AuthenticatedAppLayout`, so the entry exists for an admin or operator checking the floor, and they hold `orders:write` — but the stated reason for the gate is not the real one any more.

Since this PR is the next thing to touch that line, take the opportunity rather than merging around it.

### What I am NOT asking for

I am **not** asking you to drop `requiresRole` as a mechanism. `NavContribution.requiresRole` is a documented slot (`docs/frontend-architecture.md` § Platform Plugins), widening it from `'admin'` to a `Role[]` is a reasonable change, and several items here genuinely have no permission that expresses their route's `@Roles` list — `/`, `/insights`, `/orders`, `/customers`, `/invoices` are all `@Roles('admin','operator','viewer')` with no `viewer`-inclusive permission to name. Role arrays are the right tool there and the per-item comments citing each controller's decorator list are exactly the right way to justify them.

The narrow ask is that `/bench`, where a purpose-built permission now exists, uses it.

---

## Final assessment

### Summary
Nav entries that 403 for a `packer` get gated so the sidebar stops offering them, via a `requiresRole` widened from a single `'admin'` to a role array, with each item's gate justified against its controller's own `@Roles` list. The mechanism and the per-item reasoning are sound and I have no objection to either. What changed underneath is that the pack-bench stack merged a permission — `bench:write`, guarded against the very controllers this PR cites — which makes one of these items expressible the way the frontend docs prefer, and which falsifies the "`ROLE_PERMISSIONS.packer` is `[]`" premise several comments on both sides rest on.

### Merge readiness
❌ **Request changes** — a rebase is mandatory (three-file conflict), and the premise change is worth acting on while resolving rather than after.

### Priority fixes
1. Rebase onto `main` and resolve the three `nav-registry*` conflicts.
2. Change `/bench` to `requiresPermission: 'bench:write'`, and update its comment — the "no permission can name all three" reasoning is now false on both sides of the conflict.
3. Correct the remaining comments that describe `ROLE_PERMISSIONS.packer` as an empty grant.
4. Re-check `nav-registry.test.ts` after the rebase: both sides changed it, so a mechanical resolution can leave assertions covering neither design.

PR #3324, opened 2026-09-21
fix(sales-documents): warn at pick time instead of converging the destination predicate

### LINKED ISSUE #3232: [BUG] Frontend — converge the "which connection may a routing decision name" predicate across all four surfaces
## Problem / Context

The stack behind #3173 has converged **two of three** answers to "which connection may a routing decision name", and left the third behind. The residue is reachable by an operator who has misconfigured nothing.

**The loop:**

1. The rule composer (`apps/web/src/features/sales-documents/components/sales-document-rule-composer-dialog.tsx:154`) offers candidate destinations through `selectInvoicingCandidates` / `selectFiscalizationCandidates`. Those filter on **capability** (`Invoicing` / `Fiscalization` enabled) and on `status === 'active'`.
2. The operator picks a connection that carries `Invoicing` but has no `config.salesDocument.documentKind`. The rule saves — nothing refuses it.
3. `find-sales-document-destination-warnings.ts` (#3209) then raises an alert on that rule, because its predicate filters on **role** (`status === 'active' && documentKind !== null`).

So one surface invites the rule and the next surface complains about it. Neither is wrong on its own; they answer the same question two ways.

**What already converged:**

- `find-sales-document-destination-warnings.ts` (#3209) and `sales-document-country-defaults.tsx` (#3210) now both read `status === 'active' && documentKind !== null`, and each carries a comment naming the other.
- The remaining gap is therefore narrower than when it was first raised: it is **capability vs role**, not capability vs nothing — the composer does filter status.

**Also on the capability predicate:** `apps/web/src/features/sales-documents/components/sales-document-template-screen.tsx:78-79` uses the same capability pair and has the same question to answer.

**Why an issue rather than a comment:** it is currently named in three places — the #3210 and #3209 review threads and two source comments (`sales-document-country-defaults.tsx`, `find-sales-document-destination-warnings.ts`). Comments are where this kind of thing goes to be forgotten; the reviewer asked for it to be tracked.

## Proposed Solution

Either converge the four surfaces onto one predicate, or make the divergence explicit and safe:

- The composer dialog and the template screen resolve candidates through the same predicate the warning and the country-default select use (`status === 'active' && documentKind !== null`), or
- A stated reason is recorded for why the composer must legitimately offer a wider set (e.g. so an operator can assign a role and a rule in one pass) — in which case the composer must warn at pick time rather than letting the operator discover the gap from the rules-list alert afterwards.

## Classification

**Type**: Frontend
**Layer**: Interface (feature module)
**File(s)**:
- `apps/web/src/features/sales-documents/components/sales-document-rule-composer-dialog.tsx`
- `apps/web/src/features/sales-documents/components/sales-document-template-screen.tsx`
- `apps/web/src/features/sales-documents/lib/find-sales-document-destination-warnings.ts`
- `apps/web/src/features/sales-documents/components/sales-document-country-defaults.tsx`

## Dependencies

- #3173 (epic) — the stack this residue was left behind by.
- #3209 / #3210 — both merged, and both already converged on the `documentKind !== null` predicate this issue extends to the remaining two surfaces.

No blocking dependency outside work already on `main`.

## Assumptions

- The four surfaces named above are the complete set reading this predicate today; a fifth surface discovered during implementation should be folded in rather than left as a fresh instance of the same gap.
- Converging vs. diverging-with-a-warning are both acceptable outcomes per the acceptance criteria — the issue does not mandate one over the other, only that the gap stop being silent.

## Acceptance Criteria

- [ ] The composer dialog and the template screen resolve candidates through the same predicate the warning and the country-default select use, or a stated reason is recorded for why the composer must legitimately offer a wider set (e.g. so an operator can assign a role and a rule in one pass)
- [ ] If they converge: a connection carrying `Invoicing` but no `documentKind` can no longer be picked in the composer without the UI saying what is missing and how to fix it
- [ ] If they deliberately diverge: the composer warns at pick time rather than letting the operator discover it from the rules-list alert afterwards
- [ ] The three source comments that currently name each other are updated to point at whichever single answer survives

norbert-kulus-blockydevs:
## Summary

- `#3232` asked to either converge the four "which connection may a routing decision name" surfaces onto one predicate, or diverge deliberately with a pick-time warning.
- I chose **deliberate divergence**, not convergence. The rule composer and starter-template screen keep their capability-only candidate list (`selectInvoicingCandidates` / `selectFiscalizationCandidates`); the destination-warnings list (#3209) and country-default picker (#3210) keep their stricter `status === 'active' && documentKind !== null` predicate.
- New shared helper `find-sales-document-connection-role-gap.ts` is used by both the composer and the template screen to render a non-blocking pick-time warning when the chosen connection has no role yet, so the gap is visible immediately instead of only discovered later on the destination-warnings list.

## Why divergence, not convergence

I traced the actual runtime routing path (`AutoIssueTriggerService` → `chooseSalesDocumentDecision` → `dispatchRoute`, `libs/core/src/sales-documents/domain/domain-services/`) rather than only the issue text's summary of it:

- A **rule** (and a template's rule) carries its **own** `documentKind`, picked explicitly by the operator in the dialog. `evaluateSalesDocumentRules` uses the rule's own `(documentKind, connectionId)` directly — it never reads the connection's `config.salesDocument.documentKind` role at all.
- `AutoIssueTriggerService.dispatchRoute` resolves the winning connection from the **plain active + capability-enabled** connection list, never from the role-filtered `SalesDocumentRoutingCandidate` pool.
- That role-filtered pool is consulted **only** by the pre-#2170 zero-configuration fallback resolver (`resolveSalesDocumentRouting`), which is reached **only** when the rule engine answers `'no-configuration-for-country'` — i.e. only for a country with **no rules and no country defaults at all** (`chooseSalesDocumentDecision`).

So requiring a role in the composer/template screen would refuse a rule that routes correctly on its own, and would block exactly the workflow the issue's own proposed-solution text names as a legitimate reason for a wider set: *"so an operator can assign a role and a rule in one pass."* This is also empirically backed by the existing test suite — `sales-document-rule-composer-dialog.test.tsx`'s payload-shape test already exercises a role-less connection (`sampleConnection`'s default `config` carries no `salesDocument.documentKind`), which would have broken under a role-required convergence.

The role's real remaining purpose is narrower: the country-default picker needs *some* source for `documentKind` because the operator doesn't pick one there (it's derived from the role), and the destination-warnings/conflict displays both answer a *different* question (can this connection win the pre-#2170 fallback's "primary" slot) that a role-less connection genuinely cannot answer.

## What changed

- **New**: `apps/web/src/features/sales-documents/lib/find-sales-document-connection-role-gap.ts` (+ test) — pure helper: `findSalesDocumentConnectionRoleGap(connectionId, connections)` returns the connection's name when it has no role, else `null`; `describeSalesDocumentConnectionRoleGap` renders the warning sentence.
- **`sales-document-rule-composer-dialog.tsx`**: renders a non-blocking `Alert` under "Document & destination" when the picked connection has no role. Save stays enabled.
- **`sales-document-template-screen.tsx`** (previously untested at the unit level — added `sales-document-template-screen.test.tsx`): renders the same warning per rule slot.
- **`sales-document-country-defaults.tsx`**, **`find-sales-document-destination-warnings.ts`**, **`detect-sales-document-conflict.ts`**: header/inline comments updated to state the resolved divergence and point at the shared helper, replacing the stale "converging the two is a follow-up" note. `detect-sales-document-conflict.ts` was a fifth surface sharing the same predicate, discovered during implementation and folded in per the issue's own instruction rather than left as a fresh instance of the gap.

## Acceptance criteria

- [x] A stated reason is recorded for why the composer/template screen must legitimately offer a wider set (the role-vs-dispatch trace above, recorded in `find-sales-document-connection-role-gap.ts`'s doc comment)
- [x] The composer (and template screen) warn at pick time rather than letting the operator discover it from the rules-list alert afterwards
- [x] The source comments that named this gap are updated to point at the resolution
- [x] Tests added for the new helper, the composer's new warning states, and the template screen (which had none before)
- [x] No architecture boundary violations — no new `shared` → `features` import, cross-feature imports stay within the established `features/connections` pattern already used by `deriveSalesDocumentRows`

## Test plan

- [x] `pnpm --filter @openlinker/web lint` — 0 errors (pre-existing warnings only, none new to touched files beyond the repo-wide `explicit-function-return-type` style warning already present on every other test file)
- [x] `pnpm vitest run src/features/sales-documents` — 25 files / 195 tests pass
- [x] `pnpm vitest run` (full `apps/web` suite) — 575 files / 6208 tests pass
- [x] `pnpm check:invariants` — all pass

Closes #3232


norbert-kulus-blockydevs:
## Tech-lead review

**Summary**: Implements #3232 by choosing deliberate divergence over convergence — the composer/template-screen keep a capability-only candidate list, the destination-warnings/country-defaults/conflict-detector keep the stricter capability+role predicate, and a new pick-time warning closes the "discover it later" gap. I verified the PR's central technical claim by reading the actual routing code rather than trusting the description.

### Verification of the core claim

Traced `chooseSalesDocumentDecision` (`libs/core/src/sales-documents/domain/domain-services/choose-sales-document-decision.ts:74-88`): a rule-engine decision of kind `'route'` (or any kind other than `unresolved`/`'no-configuration-for-country'`) is returned **as-is**, never touching the role-filtered candidate pool — the fallback resolver (`resolveSalesDocumentRoutingFromCandidates`) is reached only when `ruleDecision` is `null` or reports `no-configuration-for-country`. And `evaluate-sales-document-rules.ts:188-189` (and the two other match tiers) pull `documentKind`/`connectionId` straight off the matched rule object — never off the connection's role. Finally, `AutoIssueTriggerService.dispatchRoute` (`auto-issue-trigger.service.ts:623`) resolves the winning connection with `connections.find((c) => c.id === decision.connectionId)` against the **plain** connections list, not the role-filtered `SalesDocumentRoutingCandidate` pool built earlier in the same method (`:430-441`). **The PR's reasoning for choosing divergence is correct, not just plausible.** Requiring a role in the composer would refuse a rule that genuinely routes fine.

The 5th surface it folded in (`detect-sales-document-conflict.ts`) does share the identical `status === 'active' && documentKind !== null` predicate — confirmed by reading the file — so including it wasn't scope creep.

### Non-blocking claim, verified

The new composer test explicitly asserts `expect(within(root).getByTestId('rule-save')).not.toBeDisabled()` alongside the warning rendering — the "non-blocking" claim isn't just asserted in prose, it's pinned by a test that would catch a regression to blocking.

### 🟡 IMPORTANT

**`apps/web/src/features/sales-documents/components/sales-document-rule-composer-dialog.tsx` (~line 566)** — the warning wraps `Alert` in an extra `<div data-testid="rule-connection-role-gap">` instead of using `Alert`'s own `data-testid` prop:

```tsx
// ❌ Current
<div data-testid="rule-connection-role-gap">
  <Alert tone="warning" title="This connection has no role yet">
    <p>{describeSalesDocumentConnectionRoleGap(connectionRoleGap)}</p>
  </Alert>
</div>

// ✅ Per docs/frontend-architecture.md § Test-introspection `data-testid` prop
<Alert data-testid="rule-connection-role-gap" tone="warning" title="This connection has no role yet">
  <p>{describeSalesDocumentConnectionRoleGap(connectionRoleGap)}</p>
</Alert>
```

`Alert` already carries this prop (adopted in the #3268 review specifically so callers don't need a wrapper — `apps/web/src/shared/ui/alert.tsx:28`), and the doc states the rule explicitly: *"forwarded onto the same element a test needs to name — never onto an internal wrapper `<div>`/`<span>`, since that changes how the primitive lays out inside its parent's flex/grid context."* This diff does exactly what the convention exists to prevent — an easy one-line fix, no test changes needed (`getByTestId` still finds the same string, now on the right element).

### Test coverage

Good — `sales-document-template-screen.tsx` genuinely had zero unit coverage before this (only e2e, per the PR description and confirmed by the diff being an entirely new test file), and the new suite covers the accordion, the empty-template case, and all three role-gap states (warns / no-warning-with-role / no-warning-before-pick) for both the composer and the template screen. The pure helper (`find-sales-document-connection-role-gap.ts`) has its own direct unit tests including the "status is not this helper's concern" edge case, which is a good catch — it documents *why* an inactive connection still resolves a name (candidate lists upstream already filter to active).

### Architecture

No violations. The `Connection` type import from `'../../connections'` matches the pre-existing pattern (`derive-sales-document-rows.ts`, `find-sales-document-destination-warnings.ts` already do this, and the `sales-documents` slug is already registered in both `no-restricted-imports` pattern groups per `docs/frontend-architecture.md`'s #2761 note). No new cross-context or `shared → features` edge.

### Documentation

The new helper's docblock is thorough and correctly cites the exact production code path it's asserting a fact about — this is the right shape for a "why we diverged" comment (traces `chooseSalesDocumentDecision`, names the file, explains the two-tier fallback). Good precedent for future "converge or diverge" decisions in this codebase.

**Verdict**: 🔄 **Approve with changes** — one easy, non-blocking `data-testid` placement fix (documented convention violation, not a functional bug); everything else — the load-bearing technical reasoning, the non-blocking guarantee, and the test coverage — checks out against the actual code.

norbert-kulus-blockydevs:
Addressed the review's one IMPORTANT finding in `ab0c7d9fe`:

- **`sales-document-rule-composer-dialog.tsx`**: `data-testid="rule-connection-role-gap"` now lands directly on the `Alert` component (via its documented `data-testid` prop, `apps/web/src/shared/ui/alert.tsx:28`) instead of on a wrapper `<div>`, per `docs/frontend-architecture.md` § Test-introspection `data-testid` prop. No test changes needed — `findByTestId`/`queryByTestId` in `sales-document-rule-composer-dialog.test.tsx` are DOM-depth-agnostic and still resolve.

Quality gate on the updated branch: `type-check` clean, `lint` 0 errors (132 pre-existing warnings only), `vitest run src/features/sales-documents` → 25 files / 195 tests pass.

Separately, filed #3366 for the pre-existing (not introduced by this PR) predicate-duplication note: `detectSalesDocumentConflict` and its two siblings (`find-sales-document-destination-warnings.ts`, `sales-document-country-defaults.tsx`) each independently inline the same `status === 'active' && documentKind !== null` expression — each file's docblock already names the other two, so the duplication is documented but not structurally enforced. Out of scope for this PR; tracked as tech debt in #3366.

One note on process: I could not obtain a fresh GPG signature on the new commit — the local gpg-agent's passphrase cache had expired and the rotating unlock passphrase isn't something I have access to in this session. The commit does carry the DCO `Signed-off-by` trailer (`-s`) as required, just not a GPG signature. Flagging so it can be re-signed (`git commit --amend -S` + force-push) if the merge requires it.

piotrswierzy (approved):
## Tech Lead review — `fix(sales-documents): warn at pick time instead of converging the destination predicate`

The best outcome an issue like this can have: it was filed as "converge two divergent predicates", and the investigation concluded the divergence is **correct**, recorded why, and shipped the thing that was actually missing — visibility. That is a harder and more valuable answer than doing what the ticket said.

### ✅ I verified the load-bearing claim

Everything rests on this sentence in `find-sales-document-connection-role-gap.ts`:

> *`AutoIssueTriggerService.dispatchRoute` resolves the winning connection from the plain active + capability-enabled connection list, never from the role-filtered `SalesDocumentRoutingCandidate` pool that role feeds.*

Checked it. `dispatchRoute` (line 613) takes `connections` and does a plain `.find(candidate => candidate.id === decision.connectionId)` — no role filter anywhere in it. And the fallback precedence matches `docs/architecture-overview.md` §14: only `'no-configuration-for-country'` falls through to `resolveSalesDocumentRouting`, which is the one consumer of the role-filtered pool. So a country with a matching rule genuinely never consults the connection's role, and requiring one in the composer would have **refused a rule that routes perfectly well**.

That's the finding that turns "residual gap" into "deliberate divergence", and it's worth having written down where the next person will reach for the convergence again.

### ✅ Four surfaces, one question each, now stated

The new module's docblock is the useful artefact: it enumerates all four callers, says which predicate each uses, and — critically — says *what different question each is answering*:

- country-defaults needs a role because the operator never picks a `documentKind` there (it's derived **from** the role)
- destination-warnings / conflict answer the pre-#2170 fallback's own eligibility
- composer / template carry their own `documentKind` and route without the role

Adding the matching cross-references back into `detect-sales-document-conflict.ts`, `find-sales-document-destination-warnings.ts` and `sales-document-country-defaults.tsx` — so each file says "not converged with X, see Y" — is what stops this being re-litigated from whichever file someone opens first.

### ✅ Non-blocking, at pick time

`expect(within(root).getByTestId('rule-save')).not.toBeDisabled()` is the right assertion to include: the whole argument is that a role-less pick routes correctly, so blocking the save would contradict the rationale. Warning at pick time rather than letting the operator meet it later as a "cannot route" alert on the rule they just saved is exactly the contradiction #3232 was filed about.

Reusing `deriveSalesDocumentRows`'s coercion rather than re-reading `config.salesDocument.documentKind` inline — so this helper and the country-default picker cannot disagree about the same connection — is the right call for a fourth reader of that config path.

### ✅ Tests

The helper gets five unit cases including the two null-returning edges (unknown id, nothing picked), both surfaces get the three-case treatment, and `sales-document-template-screen.tsx` gains its **first unit coverage** (previously e2e only), which is a real bonus beyond the issue's scope.

### 🟢 SUGGESTION — the helper derives every row to look up one connection

```ts
const connection = connections.find((c) => c.id === connectionId);
if (connection === undefined) return null;
const row = deriveSalesDocumentRows(connections).find((r) => r.connectionId === connectionId);
```

Two full passes plus a complete `deriveSalesDocumentRows` for a single-connection question — and in `sales-document-template-screen.tsx` it's called **inside the `.map()` over template rules**, so it's O(rules × connections) derives per render.

Harmless at today's connection counts, and correctness is unaffected. But hoisting one `deriveSalesDocumentRows(connections)` above the map and passing the derived rows in (or having the helper take a pre-derived row) would be cheaper and would read better next to the comment about reusing that coercion.

### 🟢 SUGGESTION

`style={{ marginTop: 'var(--space-1)' }}` inline on the template-screen warning — the token is right, a class would be tidier. Same note I left on the fiscal-receipt detail page.

### Notes

- `features/sales-documents` isn't a `check-ui-vocabulary.mjs` scan root, so the warning sentence living in `lib/*.ts` rather than a `*.copy.ts` bypasses no gate. Worth knowing if that folder is ever added as a root.
- CI green across all nine.

**Verdict:** ✅ Approve.

norbert-kulus-blockydevs:
Addressed both suggestions from piotrswierzy's review (the review was an approve, no blockers):

- Perf suggestion (helper re-derives `deriveSalesDocumentRows` on every call, called inside a `.map()` in `sales-document-template-screen.tsx`, O(rules x connections) per render): fixed. Split `findSalesDocumentConnectionRoleGap` into a public single-derive wrapper (kept as-is for the composer dialog, which checks one connection per render) plus `findSalesDocumentConnectionRoleGapFromRows`, which takes a pre-derived `rows` list. The template screen now hoists one `deriveSalesDocumentRows(connections)` above the map and passes it into the row-based variant. Added unit tests for the new variant.
- Style suggestion (inline `style={{ marginTop: 'var(--space-1)' }}` on the role-gap warning paragraph): fixed. Replaced with a scoped `.rule-card__role-gap` CSS class in `index.css`.
- "Notes" section (documentation-only observation about `check-ui-vocabulary.mjs` scan roots): no action needed, informational only.

norbert-kulus-blockydevs:
## Re-review (verifying piotrswierzy's follow-up)

Confirmed piotrswierzy's formal review (2026-09-22T09:10) was already ✅ **Approve** with two 🟢 SUGGESTIONS — no blockers, no inline comments. Both suggestions are genuinely fixed in the current diff, verified by reading the live diff (not trusting the follow-up comment's claim):

- **Perf**: `findSalesDocumentConnectionRoleGap` split into the single-connection wrapper (kept for the composer, one connection per render) + `findSalesDocumentConnectionRoleGapFromRows(id, connections, rows)`, which takes a pre-derived `rows` list. `sales-document-template-screen.tsx` now hoists one `deriveSalesDocumentRows(connections)` above the `.map()` instead of re-deriving per rule — confirmed at lines 373/388. New unit tests cover the row-based variant (`findSalesDocumentConnectionRoleGapFromRows` describe block).
- **Style**: the inline `style={{ marginTop: 'var(--space-1)' }}` is gone, replaced by `.rule-card__role-gap` in `index.css` — confirmed present in both the component and the stylesheet.

Also re-verified our own earlier IMPORTANT finding (Alert wrapped in a div instead of using its own `data-testid`) is still fixed and wasn't reverted by the follow-up commits — `data-testid="rule-connection-role-gap"` lands directly on `Alert`, confirmed at line 167 of the current diff.

**On #3366**: confirmed it's a linked follow-up issue only, not a code change inside this PR — correctly out of scope, as the follow-up comment states. No drift there.

CI: all 9 checks green (Build, Docker Smoke, Integration, Lint, PHP Unit, Test, Test(web), Type Check, changed-paths).

**Verdict**: ✅ **Approve** — nothing outstanding. Ready to merge as far as this review is concerned.

Issue #2016, opened 2026-08-07
[BUG] Frontend — RateLimitSection can't revert config.rateLimit back to the adapter default


jakubretajczykBD:
## Problem / Context

`EditConnectionForm`'s `RateLimitSection` lets an operator set `config.rateLimit` (Requests/min, Max concurrent), but once set, **clearing both fields and saving does not revert the connection to the adapter's default policy** — the stale limit stays in effect forever.

Root cause is in `EditConnectionForm.tsx`'s submit handler (~line 567-582):

```ts
const input = toUpdateConnectionInput(values);
const fresh = await apiClient.connections.getById(connection.id);
await updateConnection.mutateAsync({
  connectionId: connection.id,
  input: { ...input, config: { ...fresh.config, ...input.config } },
});
```

- When both rate-limit fields are cleared, `mergeStructuredIntoConfig`'s `rateLimit` clause (`edit-connection.schema.ts:668-685`) does `delete next.rateLimit` — so `input.config` correctly omits the key.
- `fresh.config` is a **live refetch of the connection right before submit**, added so a concurrent sibling write from a plugin's own `CredentialsPanel` isn't clobbered by the form's stale load-time snapshot. `fresh.config` still has the old `rateLimit`, since nothing has been saved yet.
- `{ ...fresh.config, ...input.config }` is a shallow spread. A spread can only **override** a key that is present in the right-hand object — it cannot express "this key must be absent." Because the cleared key is *absent* from `input.config`, the spread falls through to `fresh.config`'s value and the deletion never reaches the server.

Confirmed live: for an Allegro connection with `config.rateLimit: { requestsPerMinute: 10, maxConcurrent: 11 }`, clearing both Requests/min and Max concurrent fields and saving still sends the old `rateLimit` object in the `PATCH /connections/:id` payload.

Adapters resolve their effective policy as `connection.config?.rateLimit ?? resolvedAdapterDefault` (`HttpTransportFactory`, `RateLimitStatusService`) — so "revert to default" matters differently per platform: Allegro ships no `defaultRateLimit` (truly unlimited), while PrestaShop/WooCommerce ship a non-null default (e.g. 60 req/min). Either way, an operator who sets an explicit override and later wants to go back to whatever the adapter would apply on its own currently cannot — the override is stuck.

## Proposed Solution

[...]

PR #3326, opened 2026-09-21
feat(connections): add stock-location override field to connection form

### LINKED ISSUE #3207: [TASK] Frontend — connection form field for the stock-location override

norbert-kulus-blockydevs:
## Summary

Implements the frontend half of #3206: a third group in `StockAndPricingSection` for the per-connection `config.stockLocationOverride` key.

- Location picker sourced from `GET /inventory/locations` (all statuses, so a retired location an operator previously selected still renders with its name and an "— inactive" suffix rather than disappearing).
- The group renders only once at least one `inventory_locations` row exists (gated by the same query that supplies the picker options — no second network request).
- Copy states explicitly that this is an operator **assertion**, never a detection.
- Clearing the selection writes an explicit `null` (never deletes the config key), matching the `stockSafetyBuffer` / `pricingRule` / `rateLimit` convention (#2016).
- A backend 400 (unknown or retired location id) surfaces as a field-level form error via `form.setError('stockLocationOverride', ...)` rather than a generic toast.

## Dependency check

Verified #3206 (backend: `stockLocationOverride` config key, `readStockLocationOverride`, `ConnectionService` validation) is already merged to `main` via PR #3208 before starting.

## Not touched, with reasons

- `stock-and-pricing-preview.ts` — `stockLocationOverride` has no computed worked example to mirror (unlike `stockSafetyBuffer` / `pricingRule`), so the mirrored-preview file needs no change.
- Options are **not** filtered to active-only, deliberately: an all-statuses list naturally exercises the retired-location 400 path end-to-end, and there's no need to track a "frozen at open time" active set the way `LocationDialog` does.

## Test plan

- [x] `pnpm --filter @openlinker/web lint` — 0 errors (pre-existing warnings only, none in touched files)
- [x] `pnpm check:invariants` — all mirror/gate scripts pass
- [x] New test suite (`stock-location override (#3206/#3207)`, 9 cases) + all pre-existing `StockAndPricingSection` tests pass (15/15 total)
- [ ] Manual verification against a live backend (not done in this session — reviewer should confirm the 400→field-error path against a real `ConnectionService.update` rejection)

Closes #3207


norbert-kulus-blockydevs:
## Tech-lead review

**Summary**: Adds the third `StockAndPricingSection` group for `config.stockLocationOverride` (#3206's frontend half). Follows the file's established conventions closely — explicit-`null`-on-clear (#2016), gating on `hasAnyLocation` before rendering the group, all-statuses location listing so a retired-but-stored value still renders by name, and cross-feature access to `inventory` correctly through its public barrel (`../../inventory`, per `docs/frontend-architecture.md § Feature Public Surface`). Verified the backend message shape (`ConnectionService.validateStockLocationOverride`, `apps/api/src/integrations/application/services/connection.service.ts:491/496`) actually contains the literal substring `stockLocationOverride`, so the `error.message.includes('stockLocationOverride')` field-routing check in `onSubmit` is correct, not a guess.

### IMPORTANT

**[IMPORTANT]** — `apps/web/src/features/connections/components/EditConnectionForm.tsx` (`onSubmit`, ~line 748)
> The stated goal is that a `stockLocationOverride` 400 "surfaces as a field-level error rather than the generic ... alert." It does surface as a field-level error — but it does **not** replace the generic one. `updateConnection.error` is set by `useUpdateConnectionMutation` (a bare `useMutation`) automatically whenever `mutateAsync` rejects, independent of the local `try/catch`. The render always includes `{updateConnection.error ? <Alert tone="error" title="Unable to update connection">{updateConnection.error.message}</Alert> : null}` (line ~778) unconditionally alongside the form. So on the location-error path, the operator sees the SAME message twice: once in the top-of-form `Alert`, once under the `Location` select via `form.setError`. That's a real, verifiable double-render, not a style nit — it undercuts the PR's own stated intent. The new test suite doesn't catch it because it drives `StockAndPricingSection` in isolation with a hand-built `errors` prop (`ErroredHarness`), never through a full `EditConnectionForm` render with a rejecting mutation.
>
> Fix: call `updateConnection.reset()` (or otherwise suppress `updateConnection.error` for this one path) alongside `form.setError`, so only the field-level error renders when the 400 names this field.

### SUGGESTION

**[SUGGESTION]** — `apps/web/src/features/connections/components/EditConnectionForm.tsx` (comment at `onSubmit`, ~line 750)
> The comment says the field-level-error mapping is "matching how other StockAndPricingSection fields already do." Grepped the pre-PR file: there is no prior `form.setError(...)` call anywhere in `EditConnectionForm.tsx` — every other field (`stockPolicy`, `pricingRule`, `rateLimit`, etc.) falls through to the generic `updateConnection.error` banner today. This PR is actually the FIRST field-level API-error mapping in this component, not a precedent-follow. Worth correcting the comment so a future reader doesn't go looking for a sibling example that doesn't exist — and worth noting explicitly as a new pattern being introduced, since the IMPORTANT finding above is a direct consequence of it being new (nothing else in the file had to solve the "don't also show the generic banner" problem before).

### Positive observations

- `stockLocationOverrideFormSchema` deliberately stays a loose non-empty-string check and defers existence/active-status validation to the server (`ConnectionService.validateStockLocationOverride`) — correctly avoids duplicating backend validation as a source of truth (`docs/frontend-architecture.md § App Boundary`).
- `mergeStructuredIntoConfig`'s trim-and-null-on-empty handling mirrors `readStockLocationOverride`'s own trim exactly, so a value round-tripped through the form never differs from what the core reader sees — correctly reasoned in the inline comment.
- The retired-location rendering (`{location.status !== 'active' ? ' — inactive' : ''}`) and the `hasAnyLocation` gate are both directly tested (`renders a retired location...`, `does not render the group when no inventory_locations row exists`) — good, these are exactly the two states most likely to silently regress.
- `LOCATION_OPTIONS_PAGE_SIZE = 200` with an explicit one-line justification ("an install with hundreds of warehouses is not this control's shape") rather than an unexplained magic number.

### Test coverage
9 new cases covering: hidden-when-no-locations, closed-by-default, hydrate-open-with-stored-value, retired-location-still-renders, assertion-copy-present, select+resync, clear+resync, field-level-error-render, disabled-when-unparseable. Good breadth at the `StockAndPricingSection` level. The one gap is exactly the IMPORTANT finding above — no test drives the double-alert scenario through `EditConnectionForm` itself.

**Verdict**: 🔄 **Approve with changes** — the IMPORTANT finding is a real UX regression (duplicate error message) directly contradicting the PR's stated intent, and should be fixed before merge; it's a small, contained fix (`updateConnection.reset()` on the field-matched branch). The SUGGESTION is a comment-accuracy fix, non-blocking.

norbert-kulus-blockydevs:
Addressed both review findings in d05a9ea:

**IMPORTANT (dual-render banner)** — `apps/web/src/features/connections/components/EditConnectionForm.tsx`, `onSubmit`'s `catch` block: added `updateConnection.reset()` right after `form.setError('stockLocationOverride', ...)`. `updateConnection.error` is set automatically by `useMutation` on rejection, independent of the local `try/catch`, so without clearing it the generic "Unable to update connection" `Alert` kept rendering the same message alongside the new field-level error under the `Location` select. `reset()` clears the mutation's own error state so only the field-level error renders.

**SUGGESTION (false precedent in comment)** — same file, same block: reworded the comment. It previously claimed the field-level mapping was "matching how other StockAndPricingSection fields already do," which grepping the pre-PR file disproves (no prior `form.setError(...)` call exists anywhere in `EditConnectionForm.tsx`). The comment now states plainly that this is the first field-level API-error mapping in this component.

Also added a regression test in `EditConnectionForm.test.tsx` ("shows the stockLocationOverride API error only as a field-level error, not also as the generic banner") that renders the full form with a rejecting `connections.update` mock carrying a 400 naming `stockLocationOverride`, and asserts the generic banner text is absent while the field-level error still renders. Verified it fails (red) with `updateConnection.reset()` reverted and passes (green) with the fix applied. `pnpm --filter @openlinker/web lint` and `type-check` are clean (pre-existing warnings only, 0 errors).

piotrswierzy (approved):
## Review — `feat(web): per-connection stock-location override on the connection form`

### Summary

This is the operator surface #3206 shipped without: the per-connection `config.stockLocationOverride` existed, was read by `MasterInventorySyncService`, and could only be set by hand-editing the raw config JSON. The change adds a third toggle group to `StockAndPricingSection`, a flat schema field, a `syncStockLocationOverrideToJson` clone of `syncRateLimitToJson`, the `mergeStructuredIntoConfig` clause, and the first field-level API-error mapping in this component. The shape is a faithful copy of the two sibling groups — same ordering trap, same `!configIsParseable` early return, same explicit-`null`-not-delete convention, same open/close sync on both edges — which is the right instinct on a mega-form with five writers of one config key.

The copy is the strongest part: "This is an **assertion**, not a detection" plus the named consequence ("setting it would misattribute stock to the wrong location for fulfilment") is exactly the #2610 rule that a knob must state its effect rather than only take a value, and both sentences are asserted by a test rather than left to drift.

Two IMPORTANT items below; neither loses data nor crosses a boundary, so this is mergeable once they are weighed.

---

### Positive observations

**`UNKNOWN_LOCATION_FORM_ERRORS` hoisted to a stable reference** — `stock-and-pricing-section.test.tsx`. `useForm({ errors })` re-syncs on that object's identity, so a fresh literal per render loops forever *with no thrown error, no console output, and no "Maximum update depth exceeded"* — it reads as CI going quiet until the job timeout kills it. Diagnosing that and writing the mechanism down at the point of temptation, rather than just hoisting the constant, is the right treatment.

**`updateConnection.reset()` with its reason stated in place** — `EditConnectionForm.tsx`. `useMutation` sets its own error on rejection independently of this `catch`, so without the reset the same sentence renders a third time in the generic banner. The comment says that rather than leaving the next reader to wonder why a mutation is being reset inside a handler that already returned.

**The merge clause trims to match the core reader.** `readStockLocationOverride` trims before the sync uses the value, and `validateStockLocationOverride` compares trimmed on both sides — so a value round-tripping through this form can never differ from the id the sync actually asks for. Getting that wrong would have made every unrelated save re-validate (and potentially refuse) a stored id.

---

### Findings

**[IMPORTANT]** — `apps/web/src/features/connections/components/stock-and-pricing-section.tsx` (`hasAnyLocation` / the `<Select>` block)

The group's visibility and the picker's selected value are both derived from one page of locations, and three reachable states make the surface assert the opposite of the truth for a connection that *has* an override:

1. **A failed read renders identically to "no locations exist."** The docblock names this ("Absent while loading, on error, and when nothing exists") and justifies it against #2229's ceiling-readout convention — but #2229's absent field means *the adapter declared nothing*, whereas here the absence is an unread fact. The connection whose override matters most is the one whose override silently disappears from the form.

2. **A stored id absent from the fetched page falls back to the empty placeholder.** `<Select value={stockLocationOverride}>` with no matching `<option>` resolves to `''`, so the picker reads `Select a location` while the checkbox beside it is checked — a self-contradicting pair, and the *specific* location is unrecoverable from this surface. Two ways in: a location deleted after the override was set, which `architecture-overview.md § Inventory (#3206)` calls out as a *deliberately unenforced standing degraded state* rather than an edge case ("`deleteLocation`'s referential refusal counts `inventory_items` rows and cannot see a config key"); and an install past `LOCATION_OPTIONS_PAGE_SIZE = 200`, where the constant's own comment concedes the shape but the failure is silent rather than degraded.

Nothing is destroyed — `configText` keeps the key because `syncStockLocationOverrideToJson` only runs from this section's own handlers — so this is a display problem, not a data one. The remedy is the same one #3318 is landing for the eparagony select, one PR over: render the group whenever the form value is non-empty regardless of the query's outcome, and when the stored id is not in `locations`, emit a sentinel option carrying that id so the current state is named and re-picking a real location is a genuine change.

**[IMPORTANT]** — `apps/web/src/features/connections/components/EditConnectionForm.tsx:862`

`StockAndPricingSection` is mounted for **every** connection ("generic, platform-neutral: rendered for every connection"), and the new group inherits that with no capability gate. `validateStockLocationOverride` (`connection.service.ts:463`) checks only that the id names an existing, active location — it never asks whether the connection is an `InventoryMaster` — so the override can be set on a KSeF, carrier or marketplace-only connection, persist happily, and be read by nothing. That is the configuration-that-decides-nothing shape #2407 exists to refuse, and this file already holds the precedent for a per-population gate in `pricingRuleManagedElsewhere` / `needsMasterCatalog`.

Gating on the connection's enabled `InventoryMaster` capability also removes the second cost: `useInventoryLocationsQuery` carries no `enabled` and no `staleTime`, so every connection-edit page load now issues one uncached `GET /inventory/locations?limit=200` and hydrates up to 200 rows to decide a boolean, on connections that can never use the answer.

The server-side permissiveness is #3206's and not this PR's — but this PR is the first thing that *offers* the knob, and the offer is where an operator is standing.

**[SUGGESTION]** — `EditConnectionForm.tsx` (the `ApiError` arm)

`error.message.includes('stockLocationOverride')` does catch all three messages the service raises today (unknown location, retired location, not-a-non-empty-string), which is the right breadth. But it is a match on prose, and the spec hand-writes the message (`const fieldErrorMessage = 'config.stockLocationOverride names an unknown location: …'`) rather than importing it — so rewording any of those three reverts this to the generic banner with nothing failing. A shared constant, or the machine-readable-code treatment `ORDER_ALREADY_ON_HOLD` / `HOLD_ALREADY_RELEASED` got in #2341, would make the link structural. Worth doing now while there is exactly one such mapping in the component.

**[SUGGESTION]** — the retired-location option

A non-active location is listed *and selectable*, while the server refuses a transition onto one with a 400 naming the remedy. Listing it is correct and the docblock argues it well (a stored value must keep its real name). Selecting it is a refusal the operator could have been spared: `disabled` on every non-active option except the currently-stored value keeps the naming property and removes the reachable 400.

**[SUGGESTION]** — `EditConnectionForm.test.tsx`

`expect(fieldErrors.length).toBeGreaterThan(0)` is legitimate here — the message genuinely renders twice (field + `submitCount > 0` summary), so `findByText` would throw, and the assertion actually under test is the banner's absence, which is asserted separately and exactly. But the count is deterministic, so `toHaveLength(2)` would additionally fail loudly if the summary and the field ever stopped agreeing, which is the sort of drift a `> 0` swallows.

---

### Final assessment

Nothing here is a boundary violation, a missing migration, or a data-loss path, and the group behaves correctly on the happy path an operator will actually walk. The two IMPORTANT items are about what the surface says when the locations read or the stored id disagrees with the picker's one page — worth a follow-up in the same wave, especially the sentinel option, since #3318 is establishing that exact pattern next door this week.

Priority, if you take them:
1. Render the group whenever a value is stored, and carry an unknown stored id as a sentinel option rather than falling back to the empty placeholder.
2. Gate the group on `InventoryMaster` — both to stop offering an override that decides nothing, and to stop the locations read firing on every connection.
3. Anchor the error mapping to a shared constant or code.

norbert-kulus-blockydevs:

- [IMPORTANT] Group/Select visibility disagreeing with a failed/absent locations read: fixed. The group now renders whenever the form's `stockLocationOverride` value is non-empty, independent of the read's outcome (loading, error, or a stored id absent from the fetched page). An unresolved stored id renders as an explicit sentinel option ("Currently set to: ... (not in this list)") instead of silently resolving to the empty placeholder.
- [IMPORTANT] No InventoryMaster capability gate: fixed. `EditConnectionForm` now passes `stockLocationOverrideCapable={connection.enabledCapabilities.includes('InventoryMaster')}`; the group (and its `GET /inventory/locations` read, via a new `enabled` option on `useInventoryLocationsQuery`) is skipped entirely for a connection that can't use the answer.
- [SUGGESTION] `error.message.includes(...)` prose matching: fixed. Backend now carries a machine-readable `error: 'STOCK_LOCATION_OVERRIDE_INVALID'` code on all three `validateStockLocationOverride` refusals (the `ORDER_ALREADY_ON_HOLD` precedent); the frontend mapping branches on that code via a new `stock-location-override-error.ts` mirror.
- [SUGGESTION] Retired-location option selectable: fixed. Every non-active option is now `disabled` except whichever one is the currently stored value.
- [SUGGESTION] `toBeGreaterThan(0)` on the field-error count: fixed to `toHaveLength(2)`.

Added tests: capability gate (no render + no request), sentinel option for an unknown stored id, group rendering with zero fetched locations, `useInventoryLocationsQuery`'s new `enabled` option, and a backend spec pinning the error code shape.

norbert-kulus-blockydevs:
## Re-review (post piotrswierzy's round)

Pulled the current diff and full comment history, not relying on the earlier review's memory.

**Both my own IMPORTANT finding and piotrswierzy's two IMPORTANT findings are genuinely resolved, verified against the actual current code:**

1. **Dual error banner** (my finding, commit `d05a9ea`) — `updateConnection.reset()` now runs alongside `form.setError('stockLocationOverride', ...)` in `EditConnectionForm.onSubmit`'s catch block. Confirmed the generic `updateConnection.error` alert and the field-level error can no longer both render for the same rejection — verified by reading the new regression test (`EditConnectionForm.test.tsx`, "shows the stockLocationOverride API error only as a field-level error"), which asserts `queryByText('Unable to update connection')` is absent.

2. **Group/select visibility disagreeing with a failed locations read** (piotrswierzy) — `shouldRenderLocationGroup = stockLocationOverrideCapable && (hasAnyLocation || stockLocationOverride !== '')` now renders the group whenever a value is stored, independent of the read's outcome (loading/error/absent-from-page). A `storedOverrideIsUnknown` sentinel `<option>` names an unresolved stored id explicitly rather than silently falling to the empty placeholder — the exact failure mode piotrswierzy flagged (checkbox reading "on" while the picker read "nothing chosen"). Verified against `stock-and-pricing-section.tsx`'s current logic, not just the PR description's claim.

3. **No `InventoryMaster` capability gate** (piotrswierzy) — `EditConnectionForm` now passes `stockLocationOverrideCapable={connection.enabledCapabilities.includes('InventoryMaster')}`, threaded into `useInventoryLocationsQuery`'s new `enabled` option so the `GET /inventory/locations` read doesn't fire at all on an incapable connection. Confirmed the hook change (`use-inventory-locations-query.ts`) and its own test ("does not fire the request when `enabled` is false").

4. **Message-substring matching → machine-readable code** — `STOCK_LOCATION_OVERRIDE_INVALID_ERROR_CODE` now travels in the response body's `error` field on all three backend refusal sites, and the frontend mapping (`stock-location-override-error.ts`) branches on that code, not on `error.message.includes(...)`. I checked whether this new BE/FE string mirror has an automated drift guard (the pattern this codebase usually enforces via `scripts/check-*-mirror.mjs`) and found it doesn't — but neither does the PR's own cited precedent (`ORDER_ALREADY_ON_HOLD`/`HOLD_ALREADY_RELEASED`, `orders.controller.ts`), confirmed by grep. This PR is consistent with existing practice, not introducing a new gap — not flagging it.

5. **Retired-location re-selection, error-count assertion** — both fixed as described (disabled on non-active options except the currently-stored one; `toHaveLength(2)` with a documented reason — the RHF error-summary echo is legitimate, not a bug).

### No new findings

Read the full current diff (backend service + spec, three FE components, the new error-mirror file, the hook change) end to end. Architecture boundaries, naming, test coverage, and the app-boundary rule (server stays the only place existence/active-status is actually checked; FE schema stays a loose string check) all hold. The capability gate is UX-only by design — the backend does not additionally refuse `stockLocationOverride` on a non-`InventoryMaster` connection, which is correct: the value is simply inert there (never read by `MasterInventorySyncService`), consistent with this repo's "the form is not the only way in" convention for advisory config, and the raw-JSON editor remains the escape hatch regardless.

**Verdict**: ✅ **Approve** — both review rounds' findings are genuinely fixed, not just claimed fixed; no new issues found on a fresh read of the current state.

Issue #3325, opened 2026-09-21
[TECH-DEBT] Price-changes queue — replace CHIP_COUNTS_LIMIT with a real counts-by-connection endpoint


norbert-kulus-blockydevs:
## Problem / Context

Split out of #3237 per that issue's own Assumptions ("Item 4 can be split off again if the endpoint turns out to need its own backend discussion").

`apps/web/src/features/price-changes/components/price-changes-queue-table.tsx` bounds the per-connection chip-count/eligibility read to `CHIP_COUNTS_LIMIT = 200` (#3164 approval) — an unfiltered `usePriceChangesQuery({ limit: 200 })` read whose `items` are then bucketed client-side by `destinationConnectionId`. Past that ceiling, a connection chip renders a confidently-wrong undercount for the long tail. The `All` chip is unaffected (it reads the authoritative server-side `total`), which is why this was a suggestion rather than a blocker.

## Proposed Solution

A real counts-by-connection endpoint (e.g. `GET /price-changes/counts` returning `{ connectionId, count }[]`, computed server-side with a `GROUP BY`) that the queue's connection filter-bar chips read instead of the bounded client-side aggregation. This removes `CHIP_COUNTS_LIMIT` and the `unfilteredQuery` read entirely.

## Classification

**Type**: Backend (new endpoint) + Frontend (chip-count wiring)
**Layer**: Interface (HTTP) + Application service
**File(s)**:
- `apps/web/src/features/price-changes/components/price-changes-queue-table.tsx` (remove `CHIP_COUNTS_LIMIT`, `unfilteredQuery`, `connectionCounts`)
- Price-changes backend service/controller (new counts-by-connection read)

## Acceptance Criteria

- [ ] A server-side counts-by-connection read exists and is exact (not bounded by a page-size ceiling)
- [ ] `CHIP_COUNTS_LIMIT` and the unfiltered `usePriceChangesQuery` read are removed from the queue table
- [ ] Connection chips render the real count regardless of how many open episodes exist across the install
- [ ] Tests added or updated for non-trivial logic

norbert-kulus-blockydevs:
🤖 claimed for work by branch `3325-price-change-counts-endpoint` at 2026-09-22T08:45:04Z

Issue #2352, opened 2026-08-22
[TASK] CORE — inert-state reason unions and persisted reason columns


piotrswierzy:
> Part of the OMS programme backlog (**Wave 2**, stream S2, size M). Backlog slug: `W2-15`. Sibling `W2-*` slugs resolve via the Wave-2 epic's checklist. Design of record: Wave-2 product spec §4.2 (PR #2318); the #2100 persisted-reason shape.

## Problem / Context

Spec §4.2 enumerates eight inert states. #2100 established the shape they must take: a **persisted** reason, level-triggered (including back to `null`), with an attention-worthy/routine split so a default answer never becomes a red number on a healthy install.

## Proposed Solution

Reason unions in the `fulfillment-authority` leaf covering the eight states (A1-U, A2-A, A3-X, UF-L, RS-S, A5-A, RB-L, OR-P), plus `AttentionReasonValues` (the aggregate-worthy subset) and coercion guards — the `sales-documents` reason-vocabulary shape, reused wholesale.

Persistence follows the owning row: connection-scoped states on the connection, order-scoped states on `order_records` (level-triggered, excluded from `toOrm`, single writer), return-scoped states on the return. Values this build does not recognise render neutrally and are **never counted** — an older reason must not become an uncountable red number after a downgrade.

Wave-2-reachable states are A1-U, A5-A, RS-S, RB-L, OR-P; A2-A / A3-X / UF-L ship **declared but never written** pending Wave 3, exactly as #2100 shipped `missing-required-tax-id`.

## Classification

**Type**: CORE
**Layer**: Domain + Infrastructure (**migration-bearing**; synthetic-sequential timestamp per `docs/migrations.md` rule 3)
**File(s)**: `libs/core/src/fulfillment-authority/domain/types/`; `libs/core/src/orders/infrastructure/persistence/entities/order-record.orm-entity.ts`; the returns and connections entities; new migration

## Dependencies

- #2304 (`W1a-1`); #2327 (`W1c-1` — the returns entity to hang OR-P on)

## Assumptions

- Declaring three unreachable values is cheaper and safer than growing the union later under a Wave-3 deadline; the #2100 precedent is explicit about this.

## Acceptance Criteria

- [ ] All eight values declared; the five reachable ones are written by their owning service
- [ ] Reason columns are level-triggered — fixing the condition clears the value (test)
- [ ] Reason columns are excluded from every `toOrm` that could round-trip them (regression test)
[...]

PR #3327, opened 2026-09-21
feat(web): give DataTable per-row identity, migrate the price-changes queue onto it

### LINKED ISSUE #3237: [TECH-DEBT] Frontend — give DataTable per-row identity, then migrate the price-changes queue onto it

norbert-kulus-blockydevs:
## Summary

- `DataTable` gains two purely-additive props — `rowClassName` and `rowAttributes` — so a caller can hang a per-row class list and stable DOM attributes (`id`/`data-testid`/`data-state`/…) on the desktop `<tr>` without the primitive hard-coding them. `rowClassName`'s result is **appended** to the primitive's own computed `data-table__row …` classes (never replacing them); `rowAttributes` is spread onto the `<tr>` with `key`/`className`/`onClick`/`style` reserved and dropped, so a caller cannot silently disable row identity/navigation the primitive owns. Every existing `DataTable` caller is unchanged — no prop added, no behaviour change.
- `price-changes-queue-table.tsx` (the "Price changes" tab) is migrated onto `DataTable` + `DataTableCardView`, retiring its hand-rolled `<table>` and #3223's own CSS-grid mobile block in favour of the shared primitive's responsive card layout.
- `apps/web/src/index.css` loses ~270 lines: the queue's own `.table-wrap`/`.table-scroll` desktop markup styling (now handled by `.data-table`'s generic rules) and the entire `#3223` mobile-card `@media` block (now handled by `.data-table__card*`). What remains is app-specific cell **content** styling (`.cell-product`, `.price-compare`, `.delta-chip`, `.connection-tag`, `.status-note`, `.refresh-note`, `.row-actions`, `.filter-bar*`) plus the row-**state** classes (`is-resolved`/`is-group-start`/`is-grouped`/`is-flagged`), rewritten to target `.data-table__row` instead of a hand-rolled `<tr>`.
- The obsolete stylesheet-text test `price-changes-queue-card-view.test.ts` (which asserted the now-deleted #3223 CSS block) is removed.

Preserved verbatim (each covered by a test):
- write gating (`useWriteAccess('listings:write', demoMode)` + `ReadOnlyLock`) on the action cell, bulk bar, and both checkboxes
- grouping — computed from the whole page (not neighbour adjacency), and the exact "Also changes here" continuation content shared between the desktop `product` column and the mobile card `title` slot
- the muted resolved-row and flagged (needs-refresh) row treatments
- select-all comparing the **visible/selectable** intersection
- URL-namespaced filters (`queueConn`/`dir`/`big`)
- the four `queue-*` container `data-state` handles and the per-row `id`/`data-testid`/`data-row-id`/`data-state` handles
- chip counts that never render `0` for an absent or failed read

## Item 4 (counts-by-connection endpoint)

Per the issue's own Assumptions ("item 4 can be split off again if the endpoint turns out to need its own backend discussion"), the `CHIP_COUNTS_LIMIT` removal is split into a follow-up: openlinker-project/openlinker#3325. Items 1–3 ship together here.

## Dependency

#3223 (mobile card view on the existing markup) is merged to `main` via PR #3238 (`Closes #3223`), so this PR replaces that markup with the primitive's own card layout rather than building on top of it, as the issue anticipated.

## Test plan

- [x] `pnpm --filter web type-check` — clean
- [x] `pnpm --filter web lint` — 0 errors (pre-existing warning baseline unchanged)
- [x] `pnpm --filter web test` — 572 files / 6194 passed, 1 skipped (pre-existing)
- [x] `pnpm check:invariants` — all checks green, including `check-css-structure` (balanced blocks) and `check-cross-context-imports`
- [x] New `DataTable` unit tests: `rowClassName` appends rather than replaces; `rowAttributes` spreads with reserved keys dropped; every prop-omitted caller renders unchanged
- [x] New price-changes queue tests: row identity attributes/classes (grouping/resolved/flagged), select-all-visible semantics, URL-namespaced connection filter, and a mobile-card-view smoke test (`mockMobileViewport()`) asserting no `<table>` renders and the same row actions/dialogs are wired

Closes #3237


norbert-kulus-blockydevs:
## Tech-lead review

**Summary**: Two-part change — (1) `DataTable` gains `rowClassName`/`rowAttributes` as a purely additive per-row identity hook, and (2) `price-changes-queue-table.tsx` migrates its hand-rolled `<table>` onto that primitive plus `DataTableCardView`, retiring ~270 lines of duplicated desktop/mobile CSS. Verified the diff directly (not just the PR description's claims) against the specific concerns raised for this review.

### 1. Are the two new `DataTable` props genuinely additive?

✅ Yes, verified in the diff itself:
- `rowClassName?.(row) ?? ''` is appended as a fourth entry in the existing `[...]` .filter(Boolean).join(' ')` chain — it cannot replace the primitive's own `data-table__row`/`--expandable`/`--linked`/`--expanded` classes, only add to them.
- `rowAttributes` and `rowClassName` both default to `undefined`, so an omitted-prop caller produces byte-identical output to before — confirmed by the new `renders every existing caller unchanged when rowClassName/rowAttributes are omitted` test, which asserts the row's `class` attribute is exactly `'data-table__row'` with no stray `id`/`data-testid`.
- No existing `DataTable` call site in the diff had a prop added — only `price-changes-queue-table.tsx` supplies the new props.

### 2. Does `rowAttributes` correctly reject reserved keys?

✅ Yes. `sanitizeRowAttributes` filters against `RESERVED_ROW_ATTRIBUTE_KEYS = new Set(['key', 'className', 'onClick', 'style'])` and **drops** (not overrides) those keys before spreading — so a `{...extraRowAttributes}` spread after the primitive's own `className`/`onClick`/`style`/`key` on the `<tr>` cannot desync from them. The test explicitly supplies all four reserved keys with poison values (`'should-not-apply'`) and asserts they never reach the DOM, while the primitive's own `data-table__row--linked` class and presumably-live `onClick` navigation survive. This is exactly the shape the prop's own docblock promises, and the promise is tested, not just documented.

One minor observation, not blocking: `key` in `RESERVED_ROW_ATTRIBUTE_KEYS` is dead weight for the sanitizer's actual behavior — `key` is a React-reserved prop that never appears in `Object.entries()`/JSX spread in the first place (React strips it before it reaches props), so including it in the reserved set is defensive documentation rather than something the sanitizer function can ever actually catch. Harmless, and arguably good defensive doc, but worth knowing it's not doing real work.

### 3. Are the "review-won behaviours" actually preserved, and tested rather than just claimed?

Checked each one against the diff:

| Behaviour | Preserved? | Evidence |
|---|---|---|
| Write gating (`useWriteAccess`/`ReadOnlyLock`) | ✅ unchanged | The gating JSX (`write.visible ? ... : null`, `ReadOnlyLock` wrapping) is moved verbatim into the new `columns`/`cardView` closures — not rewritten. Pre-existing write-gating tests (outside this diff's hunks) continue to exercise the same logic. |
| Grouping from `groupFirstIndex` | ✅ preserved, **correctly adapted** | `groupFirstIndex: Map<string, number>` (array-index-keyed) became `groupStartIds: Set<string>` (item-id-keyed) — a genuine and necessary change, not a regression: `rowClassName`/`rowAttributes` receive the row object, not its array position, so an index-keyed map can't be consulted from those callbacks. The docblock explains this precisely. Covered by a new test asserting `is-group-start`/`is-grouped`/continuation-content correctly split across three items sharing a group key. |
| Select-all-visible semantics | ✅ untouched + tested | `allSelected`/`selectableIds`/`toggleSelectAll` logic is **not present in the diff at all** — confirmed by grep, it lives outside the touched hunks. A new test explicitly asserts a resolved (non-selectable) row's checkbox is rendered-but-disabled and never checked by "select all". |
| URL-namespaced filters (`queueConn`/`dir`/`big`) | ✅ untouched + tested | New test asserts clicking a connection chip drives `list` with `connectionId: 'dest-1'` via the URL param plumbing, unchanged from before. |
| Four `queue-*` container handles + per-row `id`/`data-testid`/`data-row-id`/`data-state` | ✅ preserved | Per-row handles now come from `rowAttributesFor()`, byte-identical values to the old inline `<tr id=… data-testid=… data-row-id=… data-state=…>`. Tested directly (`document.getElementById('price-change-row-ep-1')` assertions). |
| Chip counts never render `0` for absent/failed reads | Not touched by this diff, not re-verified here — outside the scope of what changed. | — |
| Mobile card view has no `columns` fallback | ✅ correctly handled | `cardView.actions` carries `ActionCell` explicitly (Accept/Edit/Keep would otherwise be invisible below the breakpoint per the issue's own warning) — new mobile smoke test asserts `row-accept`/`row-edit`/`row-ignore` render and that clicking Accept opens the real dialog, not a decorative stand-in. |

### Minor / non-blocking

- **`SUGGESTION`** — `rowClassNameFor`/`rowAttributesFor`/`renderProductIdentity`/`renderSelectCheckbox` are declared as plain function declarations inside the component body, recreated on every render (no `useCallback`/`useMemo`). Given `columns` and `cardView` are also freshly-constructed object/array literals on every render already (pre-existing pattern, not introduced here), this doesn't change the component's re-render cost profile — flagging only because `DataTable` presumably re-runs its own `rowAttributes?.(row)` call per row per render, same as before with inline JSX. Not worth blocking on.
- Item 4 (`CHIP_COUNTS_LIMIT`/counts-by-connection endpoint) is correctly split into #3325 per the issue's own stated Assumption that this was acceptable — the comment pointing at it is in place.

### Documentation Gaps

None — `docs/frontend-architecture.md` § UI Library Policy already documents `DataTable` as the `@tanstack/react-table` wrapper; this PR extends its documented props table implicitly via the inline docblocks, which is consistent with how other `DataTable` props are documented in-file rather than in the architecture doc.

### Test Coverage & Quality

Strong. Arrange-Act-Assert followed throughout, descriptive test names, edge cases covered (resolved-but-rendered-disabled checkbox, group continuation vs. group start, flagged+non-grouped combination). The deleted `price-changes-queue-card-view.test.ts` correctly removed rather than left stale (it asserted CSS text for markup that no longer exists).

**Verdict**: ✅ **Approve** — the two stated risk areas (additive-only props, reserved-key sanitization) both hold up under direct diff inspection, not just the PR description's claims, and every "review-won behaviour" the issue called out is either unchanged in the diff or explicitly re-tested against the new primitive.

norbert-kulus-blockydevs:

Commit [\`18d4160f9\`](https://github.com/openlinker-project/openlinker/commit/18d4160f9c0a5ec660c9c3cf9f1195d97c5fb7bf) fixes both items from the review.

**1. \`RESERVED_ROW_ATTRIBUTE_KEYS\` including \`key\`** — the review flagged this as harmless-but-misleading dead documentation, on the reasoning that React strips \`key\` before it reaches spread props. **That reasoning was wrong, verified empirically**: \`sanitizeRowAttributes\` receives a plain object from the caller's \`rowAttributes(row)\`, and if an un-sanitized \`key\` survives into the spread onto \`<tr>\`, React 19 still *honours* it for reconciliation (only warns, doesn't ignore it) — silently stealing a row's own identity across re-renders (proven with a row-swap test: removing \`key\` from the reserved set made a swapped-in row inherit the previous row's checkbox state). Corrected the comment to state the real mechanism and added a regression test (\`data-table.test.tsx\`) that fails without the guard.

**2. Mobile card has no visual counterpart to the desktop \`is-resolved\`/\`is-flagged\` row tint** — confirmed. Rather than widening the \`DataTable\` primitive (which explicitly documents \`rowClassName\`/\`rowAttributes\` as desktop-\`<tr>\`-only by design), added a small \`StatusBadge\` to the existing \`cardView.meta\` slot in \`price-changes-queue-table.tsx\`, driven by the same \`item.resolvedAt\`/\`item.needsRefresh\` state the desktop tint already reads. No primitive change, no new state.

Tests: 63/63 passing (\`data-table.test.tsx\` + \`price-changes-queue-table.test.tsx\`), lint/type-check clean.

piotrswierzy (approved):
## Review — `feat(web): give DataTable per-row identity, migrate the price-changes queue onto it`

### Summary

Two additive props on the shared primitive, and one feature's hand-rolled `<table>` plus its bespoke mobile `@media` block retired onto it — net ~235 lines of CSS deleted and a stylesheet-text test deleted with them. The migration is careful in the places migrations usually aren't: every removed CSS rule is scoped under `.price-changes-queue` (I checked the full removed-selector set — nothing global, nothing shared), every `DataTableCardView` slot the queue needs already existed so the claim "costs no primitive change" holds, and `cardView.actions` is supplied, which matters because that slot's own docblock warns that "a row action that lives only in an actions COLUMN simply does not exist below the mobile breakpoint."

I'd approve this on the `key` reservation alone. More on that below.

---

### Positive observations

**The `key` sanitization is load-bearing, and the test proves it is.** This is the part worth copying elsewhere. `key` never renders as a DOM attribute, so the natural assertion — `expect(row).not.toHaveAttribute('key')` — passes whether or not `RESERVED_ROW_ATTRIBUTE_KEYS` contains it. The follow-up test instead drives the *actual* hazard: React still honours a `key` found in a spread props object (warning only), so a constant poisoned `key` makes two different rows one reconciliation node, and the test observes that as native uncontrolled checkbox state surviving a row swap it must not survive. A test that would pass against no implementation is not weaker evidence, it is none — and this one genuinely fails against the unsanitized version. The constant's docblock also says the behaviour was *verified empirically against this project's React 19* rather than asserted from memory, which is the right standard for a claim like that.

**`rowClassName` appends rather than replaces, and the test names why.** `expect(flaggedRow).toHaveClass('data-table__row--linked')` alongside the caller's own class is the assertion that stops a future refactor turning "append" into "override" — it pins the primitive's state class surviving, not just the caller's class arriving.

**`expect(row).toHaveAttribute('class', 'data-table__row')` on the props-omitted path** is an exact-string assertion rather than a `toHaveClass`, so a stray always-on class added later fails here rather than silently landing on every existing caller. That is the right strictness for a "nothing changes for anyone else" claim.

**The CSS rewrite reasons about specificity instead of hoping.** The comment explaining why `tr.is-resolved td { background }` collapses to one `tr`-level declaration — `<td>` carries no background of its own in the primitive, only `.data-table__sticky-col` does, and this table doesn't freeze columns — is exactly the fact a reader needs to know is *true* before trusting the collapse. Likewise the group accent inverting from "group-start ADDS a border" to "continuation SUPPRESSES the primitive's own `border-top`" is the correct adaptation rather than a fight with the primitive's defaults.

**The mobile state gap is named and handled honestly.** `DataTableCardView` has no per-row class hook, so a card cannot reproduce the `<tr>` tint; rather than adding one to the primitive, `cardMetaFor` puts a badge in the existing `meta` slot, and the docblock justifies it as a *passive-scanning aid on top of an already-present fact* — `ActionCell` states resolved / needs-refresh as text on both layouts. That is the right call: the tint was never the only carrier of that state, so the card is not losing information.

**The stylesheet-text test was deleted, not rewritten.** `price-changes-queue-card-view.test.ts` asserted the existence of CSS text; replacing it with a render-level mobile smoke test that asserts no `<table>` renders and the same actions/dialogs are wired is a strict upgrade — the old shape could only ever assert that a string was present in a file.

---

### Findings

**[SUGGESTION]** — `price-changes-queue-table.tsx` — three encodings of one row state, kept in agreement by hand

`rowStateFor(item)` (the `data-state` attribute), `rowClassNameFor(item)` (`is-resolved` / `is-flagged`) and `cardMetaFor(item)` (the mobile badge) each re-derive the same row state from `item.resolvedAt` / `item.needsRefresh`, in three places, with three slightly different expressions of the same precedence — `item.needsRefresh && !item.resolvedAt` in one, an early `if (item.resolvedAt) return …` in another. They agree today; nothing asserts they will, and the desktop tint and the mobile badge disagreeing is precisely the drift the mobile counterpart exists to prevent.

`rowStateFor` already computes the answer. Deriving the class list and the badge from *it* rather than from the item makes the three impossible to diverge — the "one table, two readers" shape `AUTHORITY_ATTENTION_REASON_DESCRIPTORS` (#2352) uses, applied one layer down.

**[SUGGESTION]** — `data-table.tsx:sanitizeRowAttributes` — a dropped reserved key is silent

A caller writing `rowAttributes={() => ({ className: 'x' })}` gets nothing and no signal; the class simply never appears, and the reason lives only in the prop's docblock. The drop is the right behaviour — the alternative (throwing from a render path) is worse, and merging would defeat the point — but a `console.warn` guarded on dev would turn a puzzling absence into a one-line explanation at the moment it happens. Cheap, and this is a shared primitive whose next caller will not have read the docblock.

**[SUGGESTION]** — `rowAttributes` reserves four keys; the primitive arguably owns a fifth

`key` / `className` / `onClick` / `style` are reserved because the primitive owns identity, computed classes, row navigation and virtualized height. `Record<string, string>` already blocks the sharpest remaining edge — an event handler can't be passed as a string without React rejecting it — so this is a small point. But `tabIndex` is spreadable as a string and, on a table that sets `rowHref`, would let a caller silently change the row's keyboard reachability. Either reserve it or say in the docblock that keyboard affordances live on the in-cell `<a>` (which is where they are today, so nothing is broken — it's the next reader I'd worry about).

---

### Final assessment

Both new props are genuinely additive and proved so; the reserved-key handling is the kind of detail that is normally discovered a year later as an unreproducible stale-row bug, and here it is reasoned about, verified against the actual React version, and pinned by a test that can fail. The CSS deletion is safe by construction — everything removed was scoped under the feature's own root — and the migration keeps write gating, grouping content, select-all semantics and the row handles that E2E depends on.

The suggestions are all follow-up-sized; none of them blocks. If you take one, take the first — collapsing the three row-state derivations onto `rowStateFor` is a few lines and removes the only place in this change where two layouts can quietly start disagreeing.

Splitting item 4 out to #3325 was the right call, and saying so in the body with the issue's own assumption quoted is what makes it reviewable as a split rather than a gap.

norbert-kulus-blockydevs:
Addressed piotrswierzy's review (APPROVED, 0 blockers, 0 important, 3 suggestions - all addressed in commit ad0994571).

- rowStateFor/rowClassNameFor/cardMetaFor three-way state derivation drift - fixed: rowClassNameFor and cardMetaFor now derive from rowStateFor's own result (via a new isRowStateResolved helper) instead of independently re-reading item.resolvedAt/item.needsRefresh, so the desktop tint, mobile badge and data-state attribute cannot disagree. rowStateFor also gained a proper RowState union return type.
- sanitizeRowAttributes silently dropping a reserved key - fixed: added a dev-only console.warn (guarded by import.meta.env.DEV, matching the existing expandable+virtualize warning in the same file) naming the dropped key.
- tabIndex as a possible fifth reserved key - verified the <tr> never sets tabIndex itself (keyboard reachability lives on the in-cell rowHref anchor) and documented that explicitly in the rowAttributes JSDoc rather than reserving the key, per the reviewer's own lighter suggested option.

Note: an earlier `gh api` call against this PR's reviews returned a stale/cached result (a review body for an unrelated PR, #3363). A second fresh call returned the correct, single APPROVED review quoted above; the stale one was not acted on.

PR #3328, opened 2026-09-21
feat(listings): resolve a destination's real currency for price-change detection

### LINKED ISSUE #3203: [TASK] Backend — resolve a destination connection's real currency for price-change detection
## Problem / Context

`readConnectionCurrency` (`libs/core/src/listings/domain/types/price-change-block.types.ts`, #3143/ADR-072) reads `Connection.config.currency` to decide whether a destination's currency matches the source's, for the price-change automatic-apply bypass. On the repo's current topology this key is written by exactly one surface in the whole tree — the PrestaShop (source/master) setup form — and read by exactly one consumer that stamps the SOURCE product's currency. No destination form (Allegro, Erli, WooCommerce, or a generic connection editor) ever writes it.

The consequence: on the common PrestaShop → Allegro/Erli/WooCommerce topology, every destination resolves `null` → `'destination-currency-unknown'` → the automatic-apply bypass can never fire, no matter how the operator configures price-sync mode. The conservative posture (treat unknown as unknown, never as "known to match") is correct and intentional (ADR-061 / #2599 precedent) — what's missing is a real way to resolve a destination's actual currency.

## Proposed Solution

Resolve a destination's currency from a source OL can trust, instead of a shared, source-only config key:

- An adapter-declared value (e.g. the destination marketplace/shop account's configured currency, where the platform's API exposes one), analogous to other adapter-declared facts in the tree (`DescriptionFormat`, `ResolveConcurrencyCeiling`).
- Or an explicit, destination-scoped config field, set on the destination's own connection form rather than borrowed from the source's.

## Classification

**Type**: CORE (`listings`, `identifier-mapping`), possibly per-adapter
**Layer**: Domain / Application, possibly Infrastructure (adapter)
**File(s)**:
- `libs/core/src/listings/domain/types/price-change-block.types.ts` (`readConnectionCurrency`)

## Dependencies

- #3139 (epic: opt-in recurring price propagation with review queue)

## Assumptions

- The resolution mechanism (adapter-declared vs. destination-scoped config) is a design decision left open for the implementer to make and record — either satisfies the acceptance criteria below.
- The conservative "unknown ≠ known-to-match" fallback must survive whichever mechanism is chosen; this issue closes the coverage gap, not the safety rule.

## Acceptance Criteria

- [ ] A destination's currency is resolvable through at least one of the two proposed mechanisms (adapter-declared value or destination-scoped config field), with the chosen approach documented in the implementing PR
- [ ] `readConnectionCurrency`'s callers consume the new resolution path so a destination on the common PrestaShop → Allegro/Erli/WooCommerce topology whose currency genuinely matches the source no longer returns `'destination-currency-unknown'`
- [ ] A destination with no resolvable currency still returns `'destination-currency-unknown'` — never a guessed or defaulted match

norbert-kulus-blockydevs:
## Summary

`readConnectionCurrency` (`libs/core/src/listings/domain/types/price-change-block.types.ts`, #3143/ADR-072) reads `Connection.config.currency` to decide whether a destination's currency matches the source's, for the price-change automatic-apply bypass. That key is written by exactly one surface in the tree — the PrestaShop (source) setup form — and read by exactly one other consumer that stamps the *source* product's currency. No destination form (Allegro, Erli, WooCommerce, generic connection editor) ever writes it, so on the common PrestaShop → Allegro/Erli/WooCommerce topology every destination resolved `null` → `'destination-currency-unknown'`, permanently blocking the automatic-apply bypass no matter how price-sync mode was configured.

This closes the gap by resolving a destination's currency from a source OL can trust, instead of only the shared, source-only config key.

## Which mechanism, and why

The issue leaves the resolution mechanism open — an adapter-declared value, or a destination-scoped config field. **I chose the adapter-declared value**, mirroring the existing `DescriptionFormat` (ADR-046) / `ResolveConcurrencyCeiling` (#2229) precedent: a pure, synchronous fact the adapter declares, probed defensively (never trusted from the static type), with a graceful fallback when nothing is declared.

This fit the two named marketplace destinations exactly: Allegro's marketplace is already treated as PL-first in this codebase (`DEFAULT_ALLEGRO_TAX_COUNTRY = 'PL'` in the offer-manager adapter, used for tax-settings writes), and Erli is PLN-only with no currency field on the wire at all (`ERLI_CURRENCY = 'PLN'`, already used for its create-offer body). Both are therefore fixed, already-known facts the adapter can declare with **zero I/O and zero new configuration** — no new form field, no migration, no operator action needed to unblock the automatic-apply bypass on those two destinations.

WooCommerce's real store currency would need a live API read to resolve reliably (no adapter in this tree reads it today), which is out of scope for this issue — a WooCommerce destination keeps relying on the pre-existing `Connection.config.currency` fallback (operator-settable), unchanged from before.

## What changed

**New (CORE, `libs/core/src/listings`):**
- `domain/ports/capabilities/offer-currency-declarer.capability.ts` — optional `OfferCurrencyDeclarer` sub-capability of `OfferManagerPort` (`getDestinationCurrency(): string | null`), guard `isOfferCurrencyDeclarer`.
- `domain/ports/capabilities/shop-currency-declarer.capability.ts` — the shop-side sibling, `ShopCurrencyDeclarer` on `ShopProductManagerPort`, guard `isShopCurrencyDeclarer`. Kept optional (unlike `getDescriptionFormat`, which is required on the base port) because no in-tree shop adapter can answer it purely/synchronously today.
- `application/services/destination-currency-resolution.ts` — pure resolver functions (`resolveOfferDestinationCurrency` / `resolveShopDestinationCurrency`), mirroring `description-format-resolution.ts`.
- `application/services/destination-currency-resolution.service.ts` (+ `.interface.ts`) — `DestinationCurrencyResolutionService`, probing `OfferManager` then `ProductPublisher` (the `DescriptionFormatReadService` two-probe order), resolved via `IIntegrationsService.getCapabilityAdapter`.
- New DI token `DESTINATION_CURRENCY_RESOLUTION_SERVICE_TOKEN`, wired into `ListingsModule`, exported from the `@openlinker/core/listings` barrel.

**Changed (CORE):**
- `PriceChangeDetectionService.detectForDestination` — resolves the destination currency via the new service first, falling back to `readConnectionCurrency(connection.config)`, with a short-TTL per-connection cache (`currencyCache`, mirrors the existing `connectionCache`) so a 100-product/3-variant sweep page doesn't re-resolve a capability adapter once per changed variant.
- `PriceChangesService` — the review-queue's `destinationCurrency` display field (`toQueueItem`) now uses the same batched resolution (`fetchDestinationCurrencies`), so the reviewer-facing currency and the persisted `blockReason` never disagree.
- `readConnectionCurrency`'s docblock updated to describe its new role as the *fallback* source rather than the (previously inert) primary one.
- `docs/capabilities.md` / `docs/architecture-overview.md` — added the two new sub-capabilities to the capability reference per the "update this file in the same PR" contract, and adapter-coverage notes for Allegro/Erli.

**Adapters:**
- `AllegroOfferManagerAdapter implements ... OfferCurrencyDeclarer` → `getDestinationCurrency() { return 'PLN'; }`.
- `ErliOfferManagerAdapter implements ... OfferCurrencyDeclarer` → `getDestinationCurrency() { return ERLI_CURRENCY; }`.

## Acceptance criteria

- [x] A destination's currency is resolvable through an adapter-declared value (documented above)
- [x] `readConnectionCurrency`'s callers (`PriceChangeDetectionService`, `PriceChangesService`) consume the new resolution path — Allegro/Erli destinations on the PrestaShop → Allegro/Erli topology now resolve `'PLN'` and no longer block on `'destination-currency-unknown'` when the source currency also matches
- [x] A destination with no resolvable currency (adapter declares nothing, no `config.currency`) still returns `'destination-currency-unknown'` — never a guessed/defaulted match
- [x] Tests added/updated (capability guards, pure resolvers, the resolution service, both consuming services, both adapters)
- [x] No architecture boundary violations — verified with `check-service-interfaces.mjs` and `check-cross-context-imports.mjs`

## Test plan

- [x] `pnpm --filter @openlinker/core build/lint/type-check/test` — 6808/6808 unit tests pass
- [x] `pnpm --filter @openlinker/integrations-allegro build/lint/type-check/test` — 707/707 pass
- [x] `pnpm --filter @openlinker/integrations-erli build/lint/type-check/test` — 504/504 pass
- [x] `node scripts/check-service-interfaces.mjs` / `node scripts/check-cross-context-imports.mjs` — pass
- [x] New/updated specs cover: both capability guards, the pure resolver functions, `DestinationCurrencyResolutionService`'s probe order and fallback, `PriceChangeDetectionService`'s adapter-declared-over-config precedence + per-connection caching, `PriceChangesService`'s `destinationCurrency` resolution precedence in the review-queue list read

Closes #3203


norbert-kulus-blockydevs:
## Tech-lead review

**Summary**: Closes the destination-currency gap for the price-change automatic-apply bypass by adding two new optional sub-capabilities (`OfferCurrencyDeclarer` on `OfferManagerPort`, `ShopCurrencyDeclarer` on `ShopProductManagerPort`), a `DestinationCurrencyResolutionService` that probes them, and wiring both callers of `readConnectionCurrency` (`PriceChangeDetectionService`, `PriceChangesService`) to try the adapter-declared value first with a fallback to the pre-existing config key. Well-scoped and follows established codebase precedent closely rather than inventing a new pattern.

### Architecture & capability-port correctness
- ✅ **Probe-not-trust (ADR-046) correctly applied.** Both `isOfferCurrencyDeclarer` / `isShopCurrencyDeclarer` guards test only `typeof adapter.getDestinationCurrency === 'function'` at runtime — never trusting the static `implements` declaration alone. This is the right shape; a guard that only checked the TS type would break for an out-of-tree plugin compiled against an older `libs/core`.
- ✅ **`DestinationCurrencyResolutionService.tryResolve`'s broad `catch` is intentional and precedented**, mirroring `DescriptionFormatReadService.tryResolve` — a capability-not-supported, disabled-connection, or credential failure all collapse to "not this kind of destination," which is correct for a best-effort resolution feeding a fallback chain rather than a primary read path.
- ✅ **Fallback to `config.currency` is genuinely preserved** for WooCommerce and any adapter declaring neither capability — verified in `price-change-detection.service.ts`'s `getDestinationCurrencyCached` (`declared ?? readConnectionCurrency(connection.config)`) and the equivalent in `price-changes.service.ts`'s `fetchDestinationCurrencies`. `null ?? null` correctly falls through to `null`, never a guessed value.
- ✅ **Conservative "unknown ≠ known-to-match" posture is preserved.** Nothing in the new code paths defaults to a currency when neither source resolves; `resolvePriceChangeBlockReason` still receives `null` in that case and reads it as `'destination-currency-unknown'`, matching the ADR-061/#2599 pattern this PR explicitly cites and restates rather than reinvents.
- ✅ **Two-probe order (`OfferManager` then `ProductPublisher`) correctly mirrors `DescriptionFormatReadService`'s established order** rather than inventing a new discovery strategy — the docblock states the reasoning (a connection resolves to at most one publishing capability in practice) rather than leaving it implicit.
- ✅ **Docs actually updated, and the arithmetic checks out**: `docs/capabilities.md` — `OfferManagerPort` 20→21, `ShopProductManagerPort` 4→5, plus adapter-coverage notes for Allegro/Erli/WooCommerce; `docs/architecture-overview.md`'s sub-capability count 47→49 (matches the +2 delta). This satisfies the "update this file in the same PR" contract those docs state.
- ✅ **DI wiring follows the existing pattern exactly** — Symbol token named `{CONTEXT}_{INTERFACE}_TOKEN` = `DESTINATION_CURRENCY_RESOLUTION_SERVICE_TOKEN`, `Symbol('IDestinationCurrencyResolutionService')`, bound via `useExisting` in `ListingsModule`, exported from the module's `exports:` array and the `@openlinker/core/listings` barrel — no repository port or ORM entity crosses the barrel, so the cross-context contract stays clean.

### Minor / non-blocking
- `PriceChangeDetectionService.getDestinationCurrencyCached` and `PriceChangesService.fetchDestinationCurrencies` both wrap `resolveForConnection(...)` in an *additional* `.catch(() => null)` on top of `tryResolve`'s own broad catch inside the service — meaning `resolveForConnection` can essentially never throw already. Harmless belt-and-suspenders, but worth knowing it's redundant defense rather than covering a real second failure mode.
- The two adapters (`AllegroOfferManagerAdapter`, `ErliOfferManagerAdapter`) both hardcode `'PLN'` as a fixed value rather than reading a per-connection config — consistent with the PR's own stated reasoning (`DEFAULT_ALLEGRO_TAX_COUNTRY`, `ERLI_CURRENCY` precedents already baked into each adapter) and correctly scoped to "zero I/O, zero new config" per the issue's own acceptance criteria, but worth flagging for anyone extending Allegro support to non-PL storefronts later — the docblocks on both `getDestinationCurrency()` implementations already say so.

### Tests
Good coverage at every new seam: capability guard tests for both new sub-capabilities (presence/absence/non-function), pure resolver function tests, the resolution service's probe order + fallback, both consuming services' precedence (adapter-declared over config) and caching/batching behavior, and both adapters' fixed-value declarations. Matches the "unit test for non-trivial logic" bar from `docs/engineering-standards.md`.

**Verdict**: ✅ Approve — correctly scoped to the issue's stated boundaries, no architecture-boundary violations, conservative fallback semantics genuinely preserved, docs updated as claimed.

norbert-kulus-blockydevs:

Commit [\`2487bcba9\`](https://github.com/openlinker-project/openlinker/commit/2487bcba9adfcab3efb97896d71044747e052764) fixes both items from the review.

**1. Redundant double-catch** — \`PriceChangeDetectionService.getDestinationCurrencyCached\` and \`PriceChangesService.fetchDestinationCurrencies\` both wrapped \`DestinationCurrencyResolutionService.resolveForConnection(...)\` in an extra \`.catch(() => null)\`, on top of \`resolveForConnection\`'s own internal broad catch — so it could never actually reject. Removed the redundant catch at both call sites, left a one-line comment explaining why it's unneeded.

**2. Hardcoded \`'PLN'\`** — \`AllegroOfferManagerAdapter.getDestinationCurrency()\`'s docblock now states explicitly that \`this.environment\` (sandbox/production) is an unrelated axis carrying no country signal, plus a note to revisit once a connection can declare a non-PL Allegro storefront. Erli's method left untouched — its wire format has no currency field at all, a structural platform fact rather than a narrowing assumption.

Quality gate: \`@openlinker/core\` (440 suites / 6808 tests), \`integrations-allegro\`, \`integrations-erli\` all green.

piotrswierzy (changes requested):
## Review — `feat(listings): resolve a destination's real currency for price-change detection`

### Summary

The diagnosis is right and well told: `readConnectionCurrency` read a key only the PrestaShop *source* form ever wrote, so every destination resolved `null` → `'destination-currency-unknown'` → the automatic-apply bypass was unreachable on the repo's default topology. The mechanism chosen is the right one, and built the right way — `DestinationCurrencyResolutionService` discovers the destination kind by probing rather than switching on `platformType`, probes for the *method* rather than trusting the type, collapses every failure mode to `null`, keeps the config key as a live fallback, and never guesses. `fetchDestinationCurrencies` batches by distinct connection so a queue page costs one resolution per destination rather than one per row, and both caches state why they exist. `docs/capabilities.md` updated in the same PR, per its own contract.

What I can't approve as it stands is the **precedence**, which combines with the fixed Allegro value to make a wrong answer both unreachable-to-correct and consequential — the thing this value gates is a price published with no human in the loop.

---

### Findings

**[BLOCKING]** — `price-change-detection.service.ts:getDestinationCurrencyCached` / `price-changes.service.ts:fetchDestinationCurrencies` — the adapter's fixed guess overrides the operator's stated fact

Both sites resolve `declared ?? readConnectionCurrency(connection.config)`. Meanwhile the rewritten `readConnectionCurrency` docblock tells the operator the opposite:

> an operator may still set `config.currency` by hand on ANY connection — source or destination — to unblock it

That escape hatch is unreachable on exactly the two adapters that declare, because a non-null declaration short-circuits before the config key is ever read. And the direction matters: the adapter's value is a fixed literal, the operator's is a statement about their own account, and the adapter wins.

Concretely: `AllegroOfferManagerAdapter.getDestinationCurrency()` returns `'PLN'` unconditionally. A seller on `allegro-cz` / `allegro-sk` / `allegro-hu` — topologies this repo already enumerates by name (`allegro-order-source.adapter.ts:1121`, with its own comment about Czech delivery methods) — settles in CZK/EUR/HUF. With a PLN source they now resolve `PLN === PLN` → no block → **automatic apply**, and there is no value they can set anywhere in the product to correct it, because the one documented remedy is overridden. Before this PR they were blocked as `destination-currency-unknown`, which was wrong but safe.

The repo's own precedent on this exact question runs the other way. #2229: *"The operator's `config.rateLimit.maxConcurrent` clamps the ceiling DOWNWARD only… a safety valve on the operator's own quota."* Operator-narrows-adapter, never adapter-overrides-operator.

Inverting to `readConnectionCurrency(connection.config) ?? declared` costs one line, keeps the whole acceptance criterion (a PL Allegro/Erli destination with no config key still resolves `'PLN'` and unblocks), and turns the .cz/.sk/.hu hazard into a one-field remedy an operator can actually apply.

**[IMPORTANT]** — `allegro-offer-manager.adapter.ts:getDestinationCurrency` — declared and enforced are two different answers from one adapter

The PR cites ADR-046 / #2229 as its precedent, and those precedents have a property this one does not. #2229's whole point: *"Reported === enforced structurally: one pure function with two callers… the raw constant is unexported so a third call site cannot reopen that gap."*

Here the adapter declares a fixed `'PLN'` while its two write paths send `currency: cmd.price.currency` (line 2165) and `currency: cmd.fields.price.currency` (line 1638) — i.e. it publishes whatever currency the command carries, which is the *source's*. Nothing in the adapter pins the two together, and nothing can, because one is an assumption about the account and the other is a passthrough.

**Erli is the counter-example and it is exemplary**: `getDestinationCurrency()` returns `ERLI_CURRENCY`, the same constant the create-offer body already sends. Declaration and behaviour are one value; drift is not expressible. That is the shape the precedent actually describes.

The docblock is honest about the .cz/.sk/.hu gap and says "Revisit this once a connection can declare a non-PL Allegro storefront", which I'd normally take as sufficient. I don't here, only because the value gates an *automatic* price apply rather than a warning — a wrong ceiling costs throughput, a wrong currency costs money at a price nobody reviewed. With the precedence inverted above this becomes tolerable, since the operator regains a remedy; worth also saying in the docblock that the declared value describes the assumed *account*, not what the adapter writes.

**[SUGGESTION]** — `destination-currency-resolution.service.ts:tryResolve` — discovery should be manifest-first

`getCapabilityAdapter` resolves credentials and constructs an adapter. For a WooCommerce destination this runs twice per resolution (`OfferManager` — which WC does implement, base-port-only — returns no declarer, then `ProductPublisher`), and on the `PriceChangesService` path there is no cache between requests, so it repeats on every queue read. #2229 states the rule: *"Discovery is manifest-first — building a capability adapter resolves credentials, so `supportedCapabilities` is checked before construction."* `resolveAdapterMetadata` is a metadata-only lookup that would skip both constructions for a connection advertising neither declarer, and works on a disabled connection where `getAdapter` throws.

**[SUGGESTION]** — `ShopCurrencyDeclarer` ships with no implementer, and its synchronous signature is what prevents the one adapter that could implement it from doing so

ADR-048 decision 1 is the standing rule here (*"forbids an interface with no implementer"*), and the PR's own reason — WooCommerce cannot answer purely and synchronously — is the interesting part rather than a footnote. **PR #3317, open right now, builds exactly this value**: a WooCommerce store-currency read off `/settings/general`, with a shared cached read proven by `expect(settingsCalls).toBe(1)` and a `null` that means "declares none", never a default. That is the real answer to "what currency does this shop settle in", and `ShopCurrencyDeclarer`'s `string | null` return type is the only thing that makes it unusable here.

Either type the shop-side declarer `Promise<string | null>` so #3317's reader can satisfy it (the offer-side one can stay sync — Erli and Allegro answer from constants), or drop the shop half from this PR and land it with the adapter that implements it. The marketplace half alone delivers every acceptance criterion as written.

**[SUGGESTION]** — `resolvePriceChangeBlockReason` compares raw strings, and this PR makes that comparison live for the first time

`destinationCurrency !== sourceCurrency` with no trim and no case fold. That never mattered while the destination side was permanently `null`. Now it does: the source side is normalised (the PrestaShop schema trims + uppercases, and it is a `<select>`), but the fallback `readConnectionCurrency` returns `typeof raw === 'string' && raw.length > 0 ? raw : null` verbatim — and the docblock explicitly invites setting it *by hand*, which means the raw JSON editor, curl or MCP, where nothing normalises. A destination hand-set to `pln` then reports `currency-mismatch` between PLN and PLN.

The failure direction is safe (it blocks), so this is a usability cost rather than a money one — but it reads as a bug in OpenLinker, and the remedy is the `inventory_locations.code` rule: normalise in exactly one place, here the comparison.

---

### Positive observations

**The fallback is preserved rather than replaced.** It would have been easy to treat the config key as legacy and delete it; keeping it live is what makes WooCommerce no worse than before and gives every undeclared destination a route.

**`null` never collapses into a favourable match**, and the reasoning is restated at each new site rather than assumed — `resolveForConnection` returning `null`, the resolver functions returning `null`, and both consumers reading `null` as `'destination-currency-unknown'`. The acceptance criterion "never a guessed/defaulted match" is genuinely held.

**Both caches justify themselves in place** — `currencyCache` names the cost it avoids (`detectForDestination` runs per changed variant, and resolution constructs an adapter) and shares `connectionCache`'s TTL rather than inventing a second number. And both call sites document *why* there is no `.catch()` (the probe already collapses every failure to `null`) instead of leaving a bare unguarded `await` for the next reader to worry about.

**The stale docblock was rewritten, not appended to.** `readConnectionCurrency`'s old text asserted the guard was inert and pointed at #3203 as the open follow-up; it now describes its actual new role as the fallback. Leaving that paragraph standing is the ordinary failure mode and it was avoided.

---

### Final assessment

**Merge readiness: ❌ Request changes**

One inverted line of precedence stands between this and an approve. The service, the guards, the fallback discipline and the caching are all right, and the Erli declaration is a model of the pattern the PR set out to follow. What blocks is that the Allegro declaration is a fixed assumption which — under the current `declared ?? config` order — overrides the operator's own statement about their account, on a value that decides whether a price is published without review.

Priority fixes:

1. `readConnectionCurrency(connection.config) ?? declared`, at both consuming sites — operator narrows adapter, per #2229.
2. Say in the Allegro docblock that the declared value describes the assumed account, not what the adapter writes (it sends `cmd.price.currency`).
3. Manifest-first before `getCapabilityAdapter` in `tryResolve`.
4. Either make `ShopCurrencyDeclarer` async so #3317's WooCommerce reader can implement it, or split it out.
5. Normalise the currency comparison in one place now that it is reachable.

norbert-kulus-blockydevs:
Addressed piotrswierzy's review feedback (commit a47ccc9d5):

- **BLOCKING** (precedence inverted) - `readConnectionCurrency(connection.config)` now wins over the adapter-declared value in both `PriceChangeDetectionService.getDestinationCurrencyCached` and `PriceChangesService.fetchDestinationCurrencies`. The operator's own config now always has the last word over a fixed adapter guess (e.g. Allegro's PL-first `'PLN'` on an allegro-cz/-sk/-hu seller).
- **IMPORTANT** (docblock updated) - `AllegroOfferManagerAdapter.getDestinationCurrency` now states explicitly that the declared value describes the assumed account, not what the adapter writes (`updateOfferQuantity`/`updateOfferFields` pass through `cmd.price.currency` verbatim, with no pinning between the two).
- **SUGGESTION** (implemented) - `DestinationCurrencyResolutionService` is now manifest-first: it calls `resolveAdapterMetadata` before `getCapabilityAdapter`, so an unsupported/disabled capability is filtered out without constructing an adapter or resolving credentials.
- **SUGGESTION** (implemented) - `ShopCurrencyDeclarer.getDestinationCurrency` is now `Promise<string | null>` (async), diverging deliberately from the sync `OfferCurrencyDeclarer`, so a future WooCommerce store-settings reader can implement it without a breaking signature change.
- **SUGGESTION** (implemented) - `resolvePriceChangeBlockReason` now normalises (trim + uppercase) both sides of the currency comparison, since the destination side can arrive from an operator-typed `config.currency` value with no normalisation applied anywhere else.

Tests updated for the new precedence, the manifest-first probe, and the async shop declarer. `docs/capabilities.md` updated to describe the new precedence and the async signature.

norbert-kulus-blockydevs:
@piotrswierzy Re-confirming the fixes from commit [`a47ccc9d5`](https://github.com/openlinker-project/openlinker/commit/a47ccc9d51e0e75f4025770d8da1af3558727989) address every finding from your review, and the full quality gate is green on that commit:

1. **[BLOCKING] Precedence inverted** — `readConnectionCurrency(connection.config) ?? declared` now applies at both `PriceChangeDetectionService.getDestinationCurrencyCached` and `PriceChangesService.fetchDestinationCurrencies`. The operator's own `config.currency` always wins over a fixed adapter assumption (Allegro's PL-first `'PLN'`), so a seller on `allegro-cz`/`-sk`/`-hu` has a remedy: set `config.currency` by hand and it now overrides the guess, exactly as #2229's `rateLimit.maxConcurrent` clamps downward only.
2. **[IMPORTANT] Docblock clarified** — `AllegroOfferManagerAdapter.getDestinationCurrency` now states explicitly that the declared value describes the *assumed account*, not what the adapter *writes* (`updateOfferQuantity`/`updateOfferFields` pass through `cmd.price.currency` verbatim, with nothing pinning the two together).
3. **[SUGGESTION] Manifest-first discovery** — `DestinationCurrencyResolutionService.tryResolve` now calls `resolveAdapterMetadata` (metadata-only, no credentials resolved, no adapter constructed) before `getCapabilityAdapter`, and additionally checks `connection.enabledCapabilities` — so a WooCommerce connection (or any connection not declaring/enabling the capability) is filtered out without constructing an adapter.
4. **[SUGGESTION] `ShopCurrencyDeclarer` made async** — `getDestinationCurrency(): Promise<string | null>`, deliberately diverging from the sync `OfferCurrencyDeclarer`, so a future WooCommerce store-settings reader (e.g. #3317) can implement it without a breaking signature change. Left in this PR rather than split out, since it's dead weight until an implementer lands but costs nothing to carry.
5. **[SUGGESTION] Currency comparison normalised** — `resolvePriceChangeBlockReason` now trims + uppercases both sides before comparing, since the destination side can arrive from an operator-typed `config.currency` value (raw JSON editor / curl / MCP) with no normalisation applied anywhere upstream.

**Full quality gate, re-run on the current HEAD (`a47ccc9d5`):**
- `pnpm --filter @openlinker/core test` — 440 suites / 6812 tests, all pass
- `pnpm --filter @openlinker/worker type-check` — clean
- `pnpm check:invariants` — all ~50 invariant scripts pass (including `check-cross-context-imports`, `check-service-interfaces`, `check-core-capability-mirror`)

Nothing further to change — ready for re-review.

norbert-kulus-blockydevs:
## Re-review (post-@piotrswierzy fixes, HEAD `a47ccc9d5`)

Verified against the current diff, not just the claim comments — all 5 of @piotrswierzy's findings genuinely landed:

1. **[BLOCKING] Precedence inverted** — confirmed fixed. Both `PriceChangeDetectionService.getDestinationCurrencyCached` and `PriceChangesService.fetchDestinationCurrencies` now read `readConnectionCurrency(connection.config) ?? declared` — operator config wins, adapter-declared value is the fallback. This is the correct direction: a fixed adapter guess (Allegro's PL-first `'PLN'`) must never silently override an operator's own statement about their account with no remedy.
2. **[IMPORTANT] Allegro docblock** — confirmed present: *"This describes the assumed ACCOUNT, not what this adapter WRITES"* is in `allegro-offer-manager.adapter.ts`'s `getDestinationCurrency()` docblock.
3. **[SUGGESTION] Manifest-first discovery** — confirmed: `resolveAdapterMetadata` is called before `getCapabilityAdapter`, plus an `enabledCapabilities` check, with test coverage for both the not-enabled and the `resolveAdapterMetadata`-itself-fails paths.
4. **[SUGGESTION] Async `ShopCurrencyDeclarer`** — confirmed: `getDestinationCurrency(): Promise<string | null>`, deliberately diverging from the sync `OfferCurrencyDeclarer`, with rationale in `docs/capabilities.md`.
5. **[SUGGESTION] Currency normalization** — confirmed: `currency.trim().toUpperCase()` in `resolvePriceChangeBlockReason`.

No regression from the fix commits — the earlier `redundant double-catch` fix still holds (both call sites correctly have no `.catch()`, with a comment explaining `resolveForConnection` never rejects, and a test proving it returns `null` even when `resolveAdapterMetadata` itself throws).

CI: all 10 checks green on current HEAD.

**Verdict**: ✅ Approve. Nothing outstanding.

piotrswierzy (approved):

Re-read at `a47ccc9d5`. The blocking finding is fixed:

```ts
const currency = readConnectionCurrency(connection.config) ?? declared;
```

Operator narrows adapter, per #2229's own precedent — and the escape hatch `readConnectionCurrency`'s docblock advertises ("an operator may still set `config.currency` by hand on ANY connection … to unblock it") is now actually reachable on the two adapters that declare. A seller on `allegro-cz` / `-sk` / `-hu` can correct the value, which matters because the wrong answer there is the one that *unblocks* an automatic price apply rather than blocking it.

The acceptance criterion survives intact: a PL Allegro or Erli destination with no config key still resolves `'PLN'` from the declaration and no longer blocks on `'destination-currency-unknown'`.

Everything I liked about the original still stands — probe-not-trust on both guards, `null` never collapsing into a favourable match and the rule restated at each new site, the config fallback kept live so WooCommerce is no worse than before, both caches justifying themselves in place, and `readConnectionCurrency`'s stale docblock rewritten rather than appended to.

The three remaining suggestions are all follow-up-sized and none blocks: manifest-first discovery before `getCapabilityAdapter` in `tryResolve`, the sync `ShopCurrencyDeclarer` signature that keeps #3317's WooCommerce reader from implementing it, and normalising the now-reachable currency comparison in one place.

**✅ Approve.**

PR #3330, opened 2026-09-21
docs(mockups): pack bench redesign + new Assign Packing Work screen

### LINKED ISSUE #3329: [PRODUCT-DESIGN] Frontend — Pack bench redesign + new "Assign Packing Work" operator screen (mockups)
## Problem / Context

The real pack-bench screen (`/bench`, Wave 3b — #2413/#2416/#2418/#2421/#2422) has no dedicated
UX/copy review since it shipped, and the product has no screen at all for a supervisor to
pre-assign a parcel to a specific packer ahead of time — every packer can currently claim any
unassigned `FulfillmentWork`, with no `assignedToUserId` anywhere in the schema (only
`packedByUserId`, stamped after the fact).

Two interactive HTML mockups were built this session against the real design system
(`apps/web/src/index.css` tokens, IBM Plex Sans/Mono, signal-orange accent) and grounded against
real shipped components (`bench-parcel.copy.ts`, `bench-work-row.tsx`, `bench-documents.tsx`)
wherever a claim could be checked against the actual codebase:

1. **Pack Bench redesign** (packer-facing, existing screen) — https://claude.ai/artifact/G518HJoZsGvdpovAoyXbg6
2. **Assign Packing Work** (new operator-facing screen, no existing equivalent) — https://claude.ai/artifact/CGRGG51AHPfThJWBM2ystg

Both artifacts are currently shared at the organisation level — an owner needs to widen sharing
via the Share menu if broader distribution is wanted.

The Pack Bench mockup was checked against named competitors (BaseLinker, Apilo, IdoSell,
Sellasist, ShipStation, ShipHero, Extensiv, Linnworks, Cin7 Core) across ten research/audit passes
— process, data, layout, UX, four direct point-by-point competitor comparisons, and two
independent fresh-eyes UI/UX/copy audits with no prior context. Several real, reproduced bugs were
found and already fixed in the mockup itself: an over-scan false-positive, a "scanning paused"
banner that didn't actually pause scanning, a stale scan-confirmation badge, and an identity chip
that didn't clear on lockout.

## Proposed Solution

Commit both mockups under `docs/plans/mockups/` per `docs/frontend-architecture.md` § UX Mockups,
each keeping its own bottom section (below a divider, so it doesn't interfere with screenshots)
documenting per-round what's grounded-real vs. proposed-new, with citations — this is itself
review material, not just the mockup. This issue is the design proposal; no implementation is
requested here.

Two things need a decision before any implementation ticket can be opened:
- Whether the Pack Bench copy/UX changes are adopted wholesale, partially, or rejected.
- Whether "Assign Packing Work" is adopted at all — it requires a schema decision
  (`assignedToUserId` on `FulfillmentWork`) that has not been made.

## Classification

**Type**: Frontend
**Layer**: Interface (feature module + route, not yet built)

## Dependencies

- None to commit the mockups.
- Implementing "Assign Packing Work" depends on a schema decision for `FulfillmentWork.assignedToUserId` (not scoped here).

## Assumptions

- This issue does not request code changes to `apps/web/src/features/bench` or any schema change — it is a design proposal for review.
- "Adopted" review outcomes get split into their own implementation tickets later, each citing this issue and the specific mockup `data-state` values it implements.

## Acceptance Criteria

- [ ] Both mockup HTML files committed under `docs/plans/mockups/`
- [ ] Each mockup's own honesty-notes section (grounded-real vs. proposed-new, with citations) is preserved
- [ ] Issue reviewed and a decision recorded (adopt / partially adopt / reject) before any implementation ticket references it

norbert-kulus-blockydevs:
## Summary
- Adds two interactive HTML mockups under `docs/plans/mockups/`, per `docs/frontend-architecture.md` § UX Mockups:
  - `pack-bench-redesign.html` — a UX/copy redesign of the real pack-bench screen (Wave 3b), checked against BaseLinker, Apilo, IdoSell, Sellasist, ShipStation, ShipHero, Extensiv, Linnworks and Cin7 Core across ten research/audit passes; several real, reproduced bugs were found and are already fixed in the mockup.
  - `assign-packing-work.html` — a new operator-facing screen (pre-assign a parcel to a specific packer) with no existing equivalent; would need a schema decision (`assignedToUserId` on `FulfillmentWork`) before implementation.
- Both files keep their own bottom section (below a divider) documenting per-round what's grounded-real vs. proposed-new, with citations — review material, not just the mockup.
- Live, interactive Artifact versions (shared at the org level today):
  - Pack Bench: https://claude.ai/artifact/G518HJoZsGvdpovAoyXbg6
  - Assign Packing Work: https://claude.ai/artifact/CGRGG51AHPfThJWBM2ystg

No code changes — this is a design proposal for review, not an implementation PR.

Closes #3329

## Test plan
- [x] Both mockups open standalone (self-contained HTML/CSS/JS, no build step)
- [x] Every interactive control in both mockups verified via Playwright this session (scan/confirm/undo, location grouping, hold states, idle-lock, keyboard shortcuts, copy buttons)
- [ ] Reviewer records adopt / partially adopt / reject decision on #3329 before any implementation ticket cites this PR

https://claude.ai/code/session_01We5SKx2oRv1TVrDPRCq6mj


piotrswierzy (approved):
## Review — `docs(mockups): pack bench redesign + new Assign Packing Work screen`

Docs-only, two self-contained HTML files, no code. Reviewed against `docs/frontend-architecture.md § UX Mockups`, and I verified the load-bearing claims rather than taking them from the body.

### Decision (the test-plan item this PR is waiting on)

**Adopt `pack-bench-redesign.html`. Partially adopt `assign-packing-work.html`** — as a design direction, not as a spec ready to be cited by an implementation ticket, for the two reasons below.

---

### What I checked

**Every design token resolves.** I extracted every `var(--…)` from both files and diffed against the custom properties declared in `apps/web/src/index.css`: **zero** unresolved in either. 57 distinct tokens in the pack bench, 53 in the assign screen, all real. That is the "use real design tokens so the mockup reads as 'this app'" rule genuinely honoured rather than approximated — the usual failure is a plausible-looking invented token name, and there isn't one here.

**The schema claim is accurate.** `pack-bench-redesign.html` opens its assignment section with *"Checked first: the real schema has no `assignedToUserId` on a `FulfillmentWork` today — only `packedByUserId`, stamped AFTER the fact."* Confirmed: `fulfillment-work.orm-entity.ts:313` declares `packedByUserId`, there is no `assignedToUserId` anywhere in `libs/core/src/fulfillment`, and `assignedConnectionId` (line 152) is the ADR-054 holder axis rather than a person. Checking before proposing, and saying so in the file, is the right habit for a mockup that proposes a schema change.

**The state switcher is real and wired.** Six named states (`working`, `ready`, `unlabelled`, `hold`, `empty`, `locked`) plus four independent toggles (concurrent picker, simulated scan, idle-lock warning, connectivity), all driven from one `setState`, with `aria-pressed` maintained on the pills.

---

### Findings

**[IMPORTANT]** — `assign-packing-work.html` carries no state attribute of any kind

The rule: *"Every distinct UI state the mockup demonstrates … gets a stable `data-state` (or equivalent) attribute value"*, and the rule immediately after it: *"Backend and frontend implementation tickets for the feature must cite the specific mockup file path **and `data-state` value(s)** they implement."*

This file has zero `data-state`, zero `data-*` state equivalents, and no switcher — I checked for dialogs, modals and empty states too and there are none, so it genuinely demonstrates one screen. That makes the missing switcher defensible; it does not make the missing *name* defensible, because a ticket for this screen currently has nothing to cite and would have to fall back on a free-text description of the screen — which is the one thing the rule exists to forbid.

One `data-state="assign"` on the root is enough to close it. If the screen really does have variants worth building (an empty queue, a conflict when two supervisors assign the same parcel — both plausible for this flow), they want names and buttons too.

**[IMPORTANT]** — the body's "several real, reproduced bugs were found" reads as product defects; the file says they were mockup-draft defects

The PR body: *"checked against BaseLinker, Apilo, IdoSell … several real, reproduced bugs were found and are already fixed in the mockup."* A reviewer reads that as "defects in the shipped pack bench, corrected in the redesign" — which would mean issues need filing, because a fix that exists only in a mockup ships to nobody.

The mockup's own round-by-round section says otherwise. Round 3: *"the earlier draft showed a [control] … as a button and once as a separate tiny 'mark manually' link. Fixed to one control"*, and *"Also fixed: a table row that quietly contradicted its own quantity and status (a real bug, not a design choice)"* — both describing earlier drafts of this file, not shipped UI.

Worth one sentence in the body making that explicit. And if any of the ten research/audit passes *did* reproduce something against the real `/pack` screen, those need issues of their own rather than a silent correction in a design document.

**[SUGGESTION]** — the six switcher states are addressed through three different attribute vocabularies

`working` / `ready` / `unlabelled` / `hold` are `data-state` values toggled by `hidden`. `empty` is `[data-bench-body="empty"]`. `locked` is a `lockOverlay` toggled via `data-open`, layered over whichever inner state was active (with `stateBeforeLock` preserved — a nice touch, and the right behaviour).

That is within the rule's letter, which says "`data-state` (or equivalent)". It sits awkwardly against the rule's stated purpose — *"so there is exactly one unambiguous way to find the state"* — since a ticket author looking at a flat list of six pills has to discover which of three attributes names theirs, and `locked` in particular is not a sibling of the other five but a layer over them.

Either give all six one vocabulary, or add two lines to the file's own analysis section mapping pill → attribute → value. The latter costs nothing and is what a ticket will actually copy from.

**[SUGGESTION — for the #3329 decision, not this PR]** — `assignedToUserId` is not a free column

The proposal is coherent: `assignedConnectionId` names the *holder* (ADR-054's unit of assignment), and a person-level axis within an OL-executed holder is a genuinely different question that nothing currently answers. But `fulfillment_works` is a table the repository file header governs with a per-column writer table naming five writers plus five insert-only columns, and there is no `save(work)` at all — every mutation is a narrow conditional UPDATE. #2728 declined a `shippedAt` column on that table in exactly those terms: *"a sixth writer on a five-writer table, a second source of truth."*

So the decision on #3329 should carry that cost explicitly: who writes `assignedToUserId`, under what guard, and how it interacts with the existing claim path (today *"every packer can currently claim any unassigned parcel"*, as the mockup correctly notes). Not an objection — a pre-assignment axis plausibly deserves a sixth writer where a derivable timestamp did not — but it is the question the implementation ticket will be stuck on if the design decision doesn't answer it.

---

### Positive observations

**The round-by-round analysis below the divider is the most useful part of the PR.** Separating grounded-real from proposed-new with citations, and recording what two independent cold-read audits converged on, turns the file into review material rather than a picture. It is also what let me tell the mockup-draft bugs from product bugs above — a file that only showed the final design would have left the body's claim unfalsifiable.

**The restraint on hover.** *"every tooltip-only hint (connectivity, undo scope) moved onto the screen as visible text, since this is a touch kiosk with no hover"* is the correct call for this surface and matches the style guide's own pointer-vs-width reasoning, and it was reached from the device the screen runs on rather than from a rule.

**Naming the rail rows by order reference rather than internal work id.** Small, and exactly the class of thing that only surfaces when someone clicks through the states cold.

---

### Final assessment

It is documentation, it is self-contained, it follows the convention, every token is real and the one schema claim it rests on is true. Nothing blocks merging it as the durable repo copy the convention asks for.

Before an implementation ticket cites either file:

1. Give `assign-packing-work.html` at least one state name — a ticket cannot satisfy the citation rule against it today.
2. Clarify in the body (or the file) whether any of the "real, reproduced bugs" were shipped-product defects; if so, file them.
3. Answer the `assignedToUserId` writer question on #3329, since `fulfillment_works` has a recent precedent for declining a new column on writer-count grounds.
4. Map the six switcher pills to their three attribute vocabularies, so tickets and E2E specs cite one thing.

norbert-kulus-blockydevs:
Addressed piotrswierzy's review feedback (single review, no separate inline comments):

- **[IMPORTANT] assign-packing-work.html has no data-state** - fixed. Added `data-state="assign"` on the root `.shell` element so an implementation ticket has something to cite, per the mockup convention. No variant states were added since the file genuinely shows one screen; the switcher's absence was already correct.
- **[IMPORTANT] "several real, reproduced bugs" reads as shipped-product defects** - fixed, in the file itself rather than the PR body. `gh pr edit` / `gh api PATCH` on the PR body both failed in this session (GraphQL Projects-classic error on the former, blocked as an external-system write by the harness on the latter), so I added an explicit closing sentence to pack-bench-redesign.html's Round 3 section stating both named "real bugs" are mockup-draft mistakes caught by the audits, never reproduced against the shipped `/pack` screen - no product issue is being filed. If you'd rather have the PR body itself corrected, happy to do that once I have PATCH access, or feel free to edit it directly.
- **[SUGGESTION] map the six switcher pills to their three attribute vocabularies** - fixed. Added a two-line note directly under the state switcher in pack-bench-redesign.html.
- **[SUGGESTION] assignedToUserId is not a free column** - not changed here; you flagged this explicitly as "for the #3329 decision, not this PR." No file in this PR proposes the column's implementation, so there's nothing to fix on the PR itself - the cost you named (fifth/sixth writer on `fulfillment_works`) is recorded as the open question for whoever answers #3329's adopt/reject decision.

Pushed as 44b27e16.

piotrswierzy (approved):
## Re-review — all three findings addressed

Re-read at `932995e27`.

**`assign-packing-work.html` carries a `data-state` now.** It had none at all, so an implementation ticket had nothing to cite and the convention's "reference these values by name — never by a free-text description of *the screen where X happens*" was unenforceable for that file. One attribute on the root, and the single-screen mockup is citable like its sibling.

**The three-vocabulary mapping note is the right fix and better than renaming.** `pack-bench-redesign.html`'s six pills really do drive three different attributes — `main[data-state]`, `[data-bench-body]`, and `#lockOverlay[data-open]` — and the note says so in three lines directly under the switcher, including the part a ticket author would otherwise get wrong: the lock overlay *layers over* whichever `[data-state]` was active rather than replacing it. Collapsing them into one attribute would have been a bigger change to a file that is already the spec for five open PRs; documenting the mapping where the switcher is costs nothing and answers the question at the point it is asked.

**The "real bugs" ambiguity is closed explicitly, and the closing sentence does more than disambiguate.** It separates the two classes on the page — draft defects in *this file*, caught by the audits, versus the items tagged `new`, which are proposed changes to the shipped product — and says no product issue is filed for the former, with the reason. That is the distinction a reader arriving from the PR body could not make, and it now cannot be misread in either direction.

Everything from the original review stands. The mockups remain the reference for the whole #3401 wave, so the `data-state` coverage and the mapping note are load-bearing for every ticket citing them — including #3362 one level up, which had the same class of gap and has since fixed it.

**✅ Approve.**

norbert-kulus-blockydevs:
## Nothing outstanding here

Your two IMPORTANTs and the SUGGESTION were closed in the round that produced
the second approval, and nothing since has touched the mockups.

Commenting anyway so this PR is not the one with no reply — silence on a PR in
a stack reads as overlooked rather than done.

---

### One thing that affects this whole stack, raised here on every PR

`#3386` is `mergeable_state: dirty` and the three PRs above it read `unknown`,
because GitHub cannot evaluate past a base that does not merge. Measured with
`git merge-tree`: merging `3343-e2e-assign-packing-work` into
`3344-bench-apw-docs` conflicts in **19 files**, among them
`bench-parcel.service.ts`, `fulfillment-work.repository.ts`,
`fulfillment-work.types.ts`, `bench-work-eligibility.ts` and ADR-074 itself.

The cause is that the lower half of the stack was force-pushed with review
fixes after the upper half was cut: `3415` is missing **23 commits** that exist
on `3343`, and `3331`'s ADR commit `1e0077f35` is not in its ancestry at all.
That is also the mechanical explanation for #3360's blocking finding — the ADR
and the code contradict each other because the amended ADR and the amended code
are on branches the top of the stack never picked up.

I have not resolved those 19 files. They are other people's review fixes across
four contexts, and guessing at them is how a merge silently reverts a fix. It
needs a deliberate call on shape — rebase the top four onto the fixed lower
stack, or land the bottom ten first and rebase once onto `main` — and I would
rather ask than pick.

norbert-kulus-blockydevs:
## Heads-up: the four PRs above this one were rebased

No change to this branch — it is untouched, and its findings stay as reviewed.

The upper four (`#3386`, `#3435`, `#3438`, `#3439`) turned out to be a parallel
lineage rather than stale copies: their merge base with `3343` was the first
commit of the whole stack, so they carried their own pre-review versions of the
work this branch reviewed. That is what made `#3386` unmergeable and what
produced #3360's ADR-versus-code contradiction. They have been replayed onto
`3343`'s reviewed lineage, dropping the duplicated pre-review commits, so the
fixes made on this branch are now the only copy in the stack.

The stack now merges end to end: all fourteen PRs report `mergeable: true`, and
`#3386` is no longer `dirty`. `behind` on the lower ten is the ordinary stacked
state and clears as each base advances.

**Gate after the rebase**, because a clean replay is not a result: `pnpm lint`
and `pnpm type-check` clean across the repo, and the 37-state live sweep green
against a stack rebuilt and redeployed from the rebased head.

norbert-kulus-blockydevs:
## Stack status - this PR is unchanged

Pushed `0648b608` to the stack tip. Nothing in this PR's own diff changed; this
is a status note so the thread is complete.

**What that commit contains**, none of it touching this PR's content:

- the ADR-074 contradiction #3360 blocked on - the ADR moves to the standalone
  `setSelfServeEligible` (folding it would force a read-then-act on the PATCH
  path), and the residual it protected against is closed three ways: the reset
  moves into `clearAssignment`'s own statement, the `false` direction is guarded
  on an assignee existing, and `CHK_fulfillment_works_exclusive_needs_packer`
  backs both at the table.
- the ADR-074 filter-parity rule from #3439's review, given a compile-time guard
  (verified red) plus a spec on the argument forward.
- one CI failure: an int-spec asserting `packer` holds no permissions, which
  #3424's `bench:write` made false.

## The stack lands as ONE merge

Confirming the question raised on #3386, #3438 and #3439: **this stack merges as
a single merge, not bottom-up one PR at a time.** That closes all three
placement findings at once, which is what they were raised together for:

| Finding | Raised on | Why it closes |
|---|---|---|
| the two user-guide PNGs live three PRs above the markdown that references them | #3386 | no interval in which the guide is on `main` without its images |
| the ADR-062 scope amendment lives one PR above the reversal it records | #3438 | ADR-062 and `buyerNameMasked` land together |
| the `1894 -> 1899` migration rename | #3439 | already resolved separately - #3435 now introduces the file as `1899` |

No content moves between PRs, because the window each finding describes does not
open.

piotrswierzy (approved):
Re-approving at `26740d6`. Own diff against `main` is content-identical to `b708f08` (zero `+`/`-` content-line delta) — this is the stack bottom being rebased onto current `main`, which puts the whole chain on top of #3320/#3328/#3383 and closes the "behind main" note I left further up the stack.

PR #3332, opened 2026-09-21
feat(eparagony): issue corrective invoices

### LINKED ISSUE #3193: feat(eparagony): issue corrective invoices

norbert-kulus-blockydevs:
## Summary

Implements `CorrectionIssuer` on `EparagonyInvoicingAdapter` over the vendor's `eCorrectiveInvoice` document kind.

- A correction composes from the caller-assembled `originalDocument` snapshot: `correctedMetadata` links to the original BY INVOICE NUMBER, `correctingMetadata` states the post-correction totals and requires `merchantTIN`/`merchantName`/`merchantAddress` (unlike a plain invoice, where those are optional), and `metadata.invoiceNumber` is the correction's own new number.
- `issueCorrection` shares the create/poll/result machinery with `issueInvoice`, reusing the tax-rate policy and a newly-extracted `composeLineSet` arithmetic core now shared by both document kinds.
- The registration key is namespaced `'correction'` apart from `'invoice'`, so an idempotency-key-less correction on the same order derives a distinct documentToken/transactionToken pair.
- `CorrectionIssuer` added to the manifest's `supportedCapabilities` as an advertised-without-dispatch sub-capability, narrowed from the dispatched `Invoicing` adapter — the same pattern `RegulatoryStatusReader`/`FiscalRegistrationLocator` already use.
- Reuses `readSellerEntityAddress` (the #3274 fix already on `main`) and refuses on an incomplete seller address rather than rendering `"undefined undefined"` onto a fiscal document.

Part of epic #3173.

## Test plan

- [x] `pnpm --filter @openlinker/integrations-eparagony lint` — 0 errors
- [x] `pnpm --filter @openlinker/integrations-eparagony type-check` — clean
- [x] `pnpm --filter @openlinker/core type-check` — clean
- [x] `pnpm --filter @openlinker/integrations-eparagony test` — 346/346 pass (15 suites), including a new `eparagony-corrective-invoice.mapper.spec.ts` (32 cases) and 10 new `issueCorrection` cases in `eparagony-invoicing.adapter.spec.ts`
- [ ] CI: full `pnpm test` / `pnpm test:integration`

Closes #3193


piotrswierzy (approved):
## Review — `feat(eparagony): issue corrective invoices`

### Summary

This reverses a deliberate non-implementation the right way round: #3192 refused `CorrectionIssuer` because the vendor's `eCorrectiveInvoice` needs a before/after metadata pair the adapter could not compose, and #3193 composes it and *then* advertises the name. That ordering is the #980 rule held rather than merely cited.

The composition is careful in the places a fiscal document has to be. Every failure is an `EparagonyConfigException` → `failureMode: 'rejected'` → nothing crosses the boundary, and the list of things it refuses is the right list: no `originalDocument` snapshot, a currency that isn't the settlement currency, a missing seller TIN / name / address, a missing original document number, a malformed `issueDate`, an empty line set. Refusing a correction with no snapshot rather than rebuilding from the order's current state is the same call #2374 made on the returns side, and for the same reason.

---

### Positive observations

**The plugin file's "NOT declared, and each for a stated reason" list had its `CorrectionIssuer` entry removed, not left standing.** That is the whole discipline: a reasons list that outlives its reason stops being documentation and becomes a false claim. The surrounding prose was updated from "two sub-capabilities" to three in all four places it appears, including the `@IsIn(CoreCapabilityValues)` paragraph. (The two repo-level docs did not get the same treatment — see the finding below.)

**`applyCorrectionLines` gets four separate things right.** It refuses an `originalLineNumber` outside `[1, n]` and refuses the same position named twice, with the consequence stated (*"either would silently misattribute a delta to the wrong line, or drop it"*) — and `NrWierszaFa`-style positional linkage is exactly where a silent misattribution becomes an unauditable entry on a filed document. It never mutates the caller's array, with the reason given (*"core persists that same array"*). It carries an unnamed line through unchanged because `correctingMetadata` states the *full* post-correction totals, so every original line must exist in the "after" state. And it uses `??` rather than `||`, so a correction zeroing a line (`newQuantity: 0`) applies instead of silently falling back to the original quantity.

**The buyer's tax number travels verbatim off the issued snapshot** (`readNonEmpty(originalDocument.buyer.taxId?.value)`) and is never re-read from the order, with ADR-073 decision 5 and #1297 named in place. A correction restating a tax number the original document did not carry is precisely what the frozen snapshot exists to prevent.

**`metadata.invoiceNumber` is the correction's own number, never the original's**, with the original confined to `correctedMetadata` — and the comment says why (*"This adapter is not a `DocumentNumberConsumer`, so today the vendor generates it"*), which tells the next reader what would change if it became one.

**`composeLineSet` was extracted rather than duplicated**, and a correction runs it twice — once over the original lines, once over the corrected set — so the two summaries cannot be computed by different arithmetic. Reusing `readSellerEntityAddress` (#3274) rather than re-deriving the seller block means the `"undefined undefined"` class of defect cannot reappear on the new path.

---

### Findings

**[IMPORTANT]** — `eparagony-invoicing.adapter.ts:resolveRegistrationKey` — the kind namespace applies only to the fallback key

```ts
const supplied = idempotencyKey?.trim() ?? '';
return supplied.length > 0 ? supplied : `${kind}:${this.connectionId}:${orderId}`;
```

The method's own docblock states the hazard precisely: without the namespace, a correction derives the same `documentToken` / `transactionToken` the original invoice did, *"and the vendor — which dedupes on that token — would answer the correction with the original document"*. `issueCorrection`'s docblock adds the consequence: *"OpenLinker would record a correction that was never issued."*

That reasoning applies just as much to a supplied key, and nothing upstream makes one distinct. `IssueCorrectionRequestDto.idempotencyKey` is an optional caller-supplied string (`issue-correction-request.dto.ts:110`), and core passes it straight through (`invoice.service.ts:1053`, `idempotencyKey: cmd.idempotencyKey ?? null`) — so an integration reusing one key per order across the issue and the correction endpoint collides, and the failure is silent: a 2xx, an `InvoiceRecord` persisted as a correction, and the original document behind the token.

Namespacing unconditionally — `${kind}:${supplied}` — costs nothing that matters. A genuine retry of the same correction still derives the same token and still replays, so idempotency is preserved; the cross-kind collision simply stops being expressible. Given what it produces, I'd rather it not depend on every future caller choosing distinct keys.

**[IMPORTANT]** — `docs/architecture-overview.md` and `docs/capabilities.md` now state the opposite of the code

Neither file is in this PR, and both assert the non-implementation in their own words:

- `architecture-overview.md:366`: *"**`CorrectionIssuer` is deliberately not advertised**, because the vendor models a correction as its own `eCorrectiveInvoice` document with a before/after metadata pair the adapter does not compose; it refuses a correction pre-call, and advertising the name would promise a document the plugin cannot produce (the #980 rule…)"*
- `capabilities.md:187`: *"…does NOT implement `CorrectionIssuer`: the vendor models a correction as its own…"*, in the adapter-coverage lines at 179-187 that list which provider implements what.

`capabilities.md` carries the "update this file in the same PR" contract for exactly this (#3328's body cites it while doing so for two new sub-capabilities), and its coverage block is the file a reader consults to answer "can eparagony correct?". The plugin file got this right; the docs of record should say the same thing. `capabilities.md:168`'s `CorrectionIssuer` row needs nothing — it's the coverage prose below it, plus the `architecture-overview.md` § 14 bullet.

**[SUGGESTION]** — `issueCorrection` stamps `documentType` with no guard

`assertIssuableDocument`'s docblock makes the case against a mismatched type better than I could: *"two documents disagreeing about what was issued, one of them transmitted to a tax authority."* The new path deliberately has no counterpart, on the grounds that *"the caller has already chosen the correction capability, so `documentType` describes the document being produced rather than selecting one this adapter may not compose."*

That is true of *selection* and not of *recording*: `buildIssueResult` writes `params.documentType ?? defaultDocumentType` straight onto the `InvoiceRecord`, and core does not constrain it either (`invoice.service.ts:1051` only defaults). So a caller passing `documentType: 'invoice'` on a correction command records a plain invoice for an `eCorrectiveInvoice` — the same disagreement, arrived at from the other direction. Refusing a non-correction type here would apply the principle the sibling guard already states.

**[SUGGESTION]** — `supportedCapabilities` ordering

`'CorrectionIssuer'` is appended after an otherwise-alphabetical list (`Fiscalization`, `FiscalRegistrationLocator`, `Invoicing`, `RegulatoryStatusReader`). Cosmetic, but the array is short enough that the ordering reads as intentional and now isn't.

---

### Final assessment

**Merge readiness: 🔄 Approve with changes**

The correction composition, the refusal set and the shared create/poll machinery are all right, and the manifest change follows the repo's own capability-declaration discipline exactly. Nothing here crosses a boundary, and the failure mode for every composition problem is the safe one.

Priority fixes:

1. Namespace the registration key unconditionally (`${kind}:${supplied}`) — a supplied key shared between the issue and the correction currently collides on the token the vendor dedupes by, which is the failure the namespace was introduced to prevent.
2. Update `docs/capabilities.md`'s eparagony coverage lines and the `architecture-overview.md` § 14 bullet — both currently state that this adapter does not and should not implement `CorrectionIssuer`.
3. Optionally, refuse a non-correction `documentType` in `issueCorrection`, for the reason `assertIssuableDocument` already gives.

norbert-kulus-blockydevs:
piotrswierzy's review on this PR (state: CHANGES_REQUESTED) is titled "feat(bench): add the C keyboard shortcut (scope narrowed after auditing the real feature)" and its entire content (useScannerInput, the C shortcut, the pack-bench mockup) is about the pack-bench feature. That work lives on a separate branch/PR (#3363, `3339-pack-bench-redesign-fe`) and touches files under `apps/web/src/features/bench` and `apps/web/src/pages/bench` — none of which are part of this PR's diff (which only touches `libs/integrations/eparagony/**`).

No inline comments and no issue comments from piotrswierzy exist on this PR. Since the only review present does not address any code in this PR's actual diff, no changes were made here — applying that review's findings (scanner-burst guard, mockup annotation, unreachable-state feedback) would mean editing files this PR doesn't touch. That review should be re-posted on #3363 if it isn't already there.

